PR #2841, opened 2023-05-10
`bun add` from root folder when call within workspaces

alexlamsl:
- ignore invalid `package.json` from parent directories
- `echo` new lines from `make` targets correctly


Jarred-Sumner:
LGTm

Jarred-Sumner (approved):

Issue #2794, opened 2023-05-03
browserGlobal.fetch.bind is not a function


dilutedev:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

Errors while running application with [Node-Fetch](https://www.npmjs.com/package/node-fetch) as a dependency

### What is the expected behavior?

App running on port ...

### What do you see instead?

3 |     return (mod && mod.__esModule) ? mod : { "default": mod };
4 | };
5 | // istanbul ignore file
6 | var node_fetch_1 = __importDefault(require("node-fetch"));
7 | var browserGlobal = typeof window !== 'undefined' ? window : typeof self !== 'undefined' ? self : null; // self is the global in web workers
8 | module.exports = !browserGlobal ? node_fetch_1.default : browserGlobal.fetch.bind(browserGlobal);
                                                             ^
TypeError: browserGlobal.fetch.bind is not a function. (In 'browserGlobal.fetch.bind(browserGlobal)', 'browserGlobal.fetch.bind' is undefined)

SuperKruse:
[ky](https://github.com/sindresorhus/ky) returns almost the same error

```
119 |             // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
120 |             prefixUrl: String(options.prefixUrl || ''),
121 |             retry: normalizeRetryOptions(options.retry),
122 |             throwHttpErrors: options.throwHttpErrors !== false,
123 |             timeout: typeof options.timeout === 'undefined' ? 10000 : options.timeout,
124 |             fetch: options.fetch ?? globalThis.fetch.bind(globalThis),
                                        ^
TypeError: globalThis.fetch.bind is not a function. (In 'globalThis.fetch.bind(globalThis)', 'globalThis.fetch.bind' is undefined)
```

PR #2842, opened 2023-05-10
fix(fetch) fix fetch inheritance

cirospaciari:
Fix https://github.com/oven-sh/bun/issues/2794

### DIFF
--- a/src/bun.js/webcore/response.zig
+++ b/src/bun.js/webcore/response.zig
@@ -609,17 +609,23 @@ pub const Fetch = struct {
         break :brk errors;
     };
 
-    pub const Class = NewClass(
-        void,
-        .{ .name = "fetch" },
-        .{
-            .call = .{
-                .rfn = Fetch.call,
-                .ts = d.ts{},
-            },
-        },
-        .{},
-    );
+    // pub const Class = NewClass(
+    //     void,
+    //     .{ .name = "fetch" },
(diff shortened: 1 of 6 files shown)


src/bun.js/webcore/response.zig
```diff
@@ -929,6 +935,31 @@ pub const Fetch = struct {
         }
     };
 
+    pub fn jsFunction(
+        globalThis: *JSC.JSGlobalObject,
+        callframe: *JSC.CallFrame,
+    ) callconv(.C) js.JSObjectRef {
+        const arguments = callframe.arguments(3);
+        var args = @ptrCast([*]const JSC.C.JSValueRef, callframe.argumentsPtr())[0..arguments.len];
```
cirospaciari: @Jarred-Sumner  not sure if this is the better approach here, `Bun.fetch` still uses `Fetch.call`, and `globalThis.fetch` calls `Fetch.jsFunction` that also wraps `Fetch.call`, can I attach `Fetch.jsFunction` into `Bun.fetch` into ZigGlobalObject?

cirospaciari: solved at https://github.com/oven-sh/bun/pull/2842/commits/506cb9f307cfe5d8fdc1a03a54b815756f9f50ab

Jarred-Sumner:
nice

PR #2844, opened 2023-05-10
add tests for ws.close

cirospaciari:


Jarred-Sumner:
now the question is

are the tests on macOS timing out due to a github actions issue or due to the test timing out

cirospaciari:
> now the question is
> 
> are the tests on macOS timing out due to a github actions issue or due to the test timing out

I will add a manual timeout on ws.close test, so if is the test, will popup as fail

Jarred-Sumner:
Thank you

Issue #2819, opened 2023-05-08
`fetch` does not honor `redirect: 'error'`


spiffytech:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Linux 6.2.14-300.fc38.x86_64 x86_64 unknown

### What steps can reproduce the bug?

`await fetch('https://httpbin.org/redirect/1', {redirect: 'error'});`

### What is the expected behavior?

The promise should reject because the endpoint returns a `302` redirect.

### What do you see instead?

Bun follows the redirect, and the promise resolves / rejects based on fetching the new `location`.

PR #2845, opened 2023-05-10
feat(fetch) add redirect: 'error' support

cirospaciari:
Fix: #2819

### DIFF
--- a/src/bun.js/webcore/response.zig
+++ b/src/bun.js/webcore/response.zig
@@ -554,7 +554,6 @@ pub const Fetch = struct {
     pub const fetch_error_no_args = "fetch() expects a string but received no arguments.";
     pub const fetch_error_blank_url = "fetch() URL must not be a blank string.";
     pub const fetch_error_unexpected_body = "fetch() request with GET/HEAD/OPTIONS method cannot have body.";
-    pub const fetch_error_redirect = "fetch() redirect must be one of: \"follow\", \"error\", \"manual\".";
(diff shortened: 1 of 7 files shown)


src/bun.js/webcore/response.zig
```diff
@@ -1030,8 +1031,11 @@ pub const Fetch = struct {
                     }
 
                     if (options.get(ctx, "redirect")) |redirect_value| {
```
Jarred-Sumner:
```suggestion
                    if (options.getOptionalEnum(ctx, "redirect", FetchRedirect)) |redirect_value| {
```

Then, in `FetchRedirect`:

```zig
pub const Map = bun.enumMap(FetchRedirect, &.{ .{ "manual",  .manual}, .{"redirect", .redirect"}, .{"error", .error}});
```

then we can rename the enum values to their lower case version (except error would be `@"error"`)

This will handle printing a human-readable error message for us when the enum string value is incorrect (rather than silently ignoring it)

Jarred-Sumner: enumMap might be wrong btw, can just use ComptimeStringMap

cirospaciari: fix here https://github.com/oven-sh/bun/commit/616bfbeef5a7877b327eacfe67ae5b65039d53c4

Jarred-Sumner (approved):
LGTM once tests finish

Issue #2675, opened 2023-04-17
dns.lookup does not respect options


Ehesp:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Darwin 22.2.0 arm64 arm

### What steps can reproduce the bug?

The following code:

```
options.all = true;
dns.lookup('example.com', options, (err, addresses) =>
  console.log('addresses: %j', addresses));
```

Is expected to return 

```
// addresses: [{"address":"2606:2800:220:1:248:1893:25c8:1946","family":6}] 
```

Taken from: https://nodejs.org/api/dns.html#dnslookuphostname-options-callback

### What do you see instead?

The result of `addresses` returns a single IP address (not an array, nor containing the object of `{ address, family }`:

```
options.all = true;
dns.lookup('example.com', options, (err, addresses) =>
  console.log('addresses: %j', addresses));
```

Returns

```
// addresses: 0.0.0.0
```

PR #2846, opened 2023-05-11
feat(dns) add support for { all: true } option on dns.lookup

cirospaciari:
Fix https://github.com/oven-sh/bun/issues/2675


Jarred-Sumner:
something is very strange with CI
image

cirospaciari:
> something is very strange with CI
> image
> 

Really weird

Issue #2674, opened 2023-04-17
`Bun.write` fails to write to stdout with long strings


savioxavier:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 unknown

### What steps can reproduce the bug?

Write a long string to `Bun.stdout` using `Bun.write()`

> It will fail at string lengths greater than 262143 (see Additional Information below)

If the string is small enough, the command is successfully executed
```js
import { write, stdout } from 'bun'

const str = "hello"
await write(stdout, str)
```

However, if the string is large (for example a huge text or code file), the command fails
```js
import { write, stdout, file } from 'bun'

const str = await file("path/to/huge/file").text() // for extracting the text
await write(stdout, str)
```

The same can be achieved even if we use a huge string as well:
```js
import { write, stdout } from 'bun'

const str = "a".repeat(300000)
await write(stdout, str)
```

### What is the expected behavior?

The text should be displayed in the console properly, instead of the `ENOENT` error.

### What do you see instead?

```
ENOENT: No such file or directory
syscall: "open"
  errno: -2
```

#### Reason

- Although this works fine with `console.log()`, I wanted to use `Bun.write()` because it writes to stdout _way_ faster than `console.log()`, which is really helpful while outputting a really long string
- Although I could have directly used the `file()`'s Blob output, I required it to be converted to a String format so that I could modify the text (eg: string replaces)

#### Experiment

I wanted to see at what string length would `Bun.write()` fail, so I made a small script as follows:

![image](https://user-images.githubusercontent.com/38729705/232490117-07b68a5b-b011-43dc-b5e3-9628acb9a9bc.png)

With my previous tests on reading file contents using `Bun.write()`, I figured the max length was somewhere in the range 250k < x < 300k, and with the help of some trial and error, I was able to find out the failing length.

The bash script runs the JS file repeatedly until it _does not fail_. When the JS file runs successfully, the bash script exits and greps the "len" field. I used a reverse loop instead of starting from length 0 as it would have taken an unnaturally long time to fully execute.

Executing the script shows us that it fails at string length _greater than_ 262143

I should also note that this worked perfectly fine in v0.2 and has not been working since. I probably should have reported this earlier, but I really didn't have a chance to experiment in the past few months.

Jarred-Sumner:
As a temporary.workaround: you could use `console.write`

savioxavier:
[...]

PR #2850, opened 2023-05-11
fix large stdout output

cirospaciari:
Fix: #2674

### DIFF
--- a/test/js/bun/io/bun-write.test.js
+++ b/test/js/bun/io/bun-write.test.js
@@ -290,3 +290,8 @@ it.skip("Bun.write('output.html', HTMLRewriter.transform(Bun.file)))", async don
   expect(await Bun.file(outpath).text()).toBe("<div><blink>it worked!</blink></div>");
   done();
 });
+
+it.skip("#2674", async () => {
+  const str = "a".repeat(300000);
+  expect(await Bun.write(Bun.stdout, str)).toBe(300000);
+});
--- a/test/js/bun/io/bun-write.test.js
+++ b/test/js/bun/io/bun-write.test.js
@@ -1,7 +1,7 @@
 import fs from "fs";
 import { it, expect, describe } from "bun:test";
 import path from "path";
-import { gcTick, withoutAggressiveGC, bunExe } from "harness";
+import { gcTick, withoutAggressiveGC, bunExe, bunEnv } from "harness";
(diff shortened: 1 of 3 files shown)


test/js/bun/io/bun-write.test.js
```diff
@@ -290,3 +290,8 @@ it.skip("Bun.write('output.html', HTMLRewriter.transform(Bun.file)))", async don
   expect(await Bun.file(outpath).text()).toBe("<div><blink>it worked!</blink></div>");
   done();
 });
+
+it("#2674", async () => {
```
cirospaciari, line 294: @Jarred-Sumner this will pollute the console **out,** but I added it so we can test on CI/CD at least once before merging, after this I will add `it.skip`

Jarred-Sumner: Can we run this in a separate script that is spawnSync’d?

Jarred-Sumner: Then we could either redirect to a file or read the buffer in memory in the pipe or both

cirospaciari: sounds good I will do that

cirospaciari: done in https://github.com/oven-sh/bun/pull/2850/commits/0dcd90d693e4be86eacc373d4d7ecdd1fbfef2cc

test/js/bun/io/bun-write.test.js
```diff
@@ -290,3 +290,19 @@ it.skip("Bun.write('output.html', HTMLRewriter.transform(Bun.file)))", async don
...
+  const file = path.join(import.meta.dir, "big-stdout.js");
+
+  const { stderr, stdout, exitCode } = Bun.spawnSync({
+    cmd: [bunExe(), "run", file],
+    stderr: "pipe",
```
Jarred-Sumner, line 300: this should use `env: bunEnv`

cirospaciari: added https://github.com/oven-sh/bun/pull/2850/commits/8c0f6001a3890e1397d0687c38fc3783db319c6f

Jarred-Sumner (approved):

Jarred-Sumner:
LGTM

lets merge once tests run

Issue #2568, opened 2023-04-05
`mkdtemp` crashes if parent directory doesn't exist


paperclover:
### What version of Bun is running?

0.5.9 [canary]

### What platform is your computer?

Linux 6.2.2-arch1-1 x86_64

### What steps can reproduce the bug?

```
import { mkdtempSync } from "node:fs";
console.log(mkdtempSync("/tmp/hello/world"));
```
`rm /tmp/hello -rf`
`bun file.js`

### What is the expected behavior?

thrown error
![image](https://user-images.githubusercontent.com/24465214/230213361-b833072f-9f24-4cf3-b807-d69f51813615.png)

### What do you see instead?

seg fault

PR #2851, opened 2023-05-11
fix(fs) mkdtemp and mkdtempSync errors

cirospaciari:
Fix: #2568


src/bun.js/fs.exports.js
```diff
@@ -157,9 +157,15 @@ export var access = function access(...args) {
 function callbackify(fsFunction, args) {
   try {
     const result = fsFunction.apply(fs, args.slice(0, args.length - 1));
-    queueMicrotask(() => args[args.length - 1](null, result));
+    const callback = args[args.length - 1];
```
Jarred-Sumner, line 160: nice

Issue #1443, opened 2022-11-01
No error is thrown when port is taken


Electroid:
```js
import { serve } from "bun";

serve({
  port: 4040,
  fetch(request) {
    return new Response("Hi");
  }
});

serve({
  port: 4040,
  fetch(request) {
    return new Response("Hi (again)");
  }
});
```

I would expect an error to be thrown on the 2nd `serve`. There is also no error when the port is taken by a non-Bun process. Tested on MacOS Silicon, while trying to debug #247

Jarred-Sumner:
This is more complicated than it seems

SO_REUSEPORT is a builtin way to do load balancing, but it only really works properly on Linux

We should probably disable SO_REUSEPORT when development mode is enabled

Electroid:
I think it's a potential footgun to not have this throw, even though `SO_REUSEPORT` is sometimes useful in certain deployments. Maybe that's a flag you have to opt-in to?

```js
export default {
  port: 3000,
  portReuse: true,
  // ...
}
```

Electroid:
This was fixed.

calebeaires:
@Electroid, when you mentioned that "we will modify the default behavior to indicate an error in the event of a port conflict," are you implying that the reusePort feature is being considered for deprecation?

I have found the reusePort feature to be highly beneficial, as it serves as a viable alternative to PM2 cluster mode!

khaled-0:
This serves as a built in load balancer for me. Should it?
Seems like an unintentional feature to me.

Jarred-Sumner:
@calebeaires @khaled-0 we are not deprecating `reusePort: true`, do not worry.

Originally, `reusePort: true` did not exist, it was implicitly the default behavior (i.e. no option), which means we did not throw an error when the port was already in use. This is very confusing in development, but in production this feature is important. So we added `reusePort: true` to make this behavior explicit - to allow the production usecase without causing confusion in development.

PR #2852, opened 2023-05-11
Implement `Bun.serve({fd})` and `Bun.listen({fd})` for socket-activated deployments

shivak:
**Overview**. This PR implements `Bun.serve({fd})` and `Bun.listen({fd})`, which runs the HTTP (or TCP) server on a socket that has are already been bound and listened. The overall idea is to have a separate process `bind()` and `listen()` to a socket, and pass it to the Bun process. This is used in production deployments for:
- efficiency: servers can be started on demand in, e.g., large multi-tenant edge setups ☺
- security: a privileged process can bind port < 1024 and then pass the socket to an unprivileged server. Also, the socket can be passed into a restricted network namespace wherein the server is run. This allows the server to respond to requests, but blocks all other network activity.
- reliability: socket activation can implement failover.

This feature is available in Go, Node, etc but not Deno (https://github.com/denoland/deno/issues/6529). Here, it's mildly tricky to implement, since it involves changes to two layers of low-level platform bindings - see the companion PRs at the bottom.

**Usage**. Typical users (on Linux/systemd) would call `Bun.serve({fd: process.env.LISTEN_FDS})`. On Mac/launchd, the fd is obtained from [launch_activate_socket](https://www.manpagez.com/man/3/launch_activate_socket/). On Windows, it could ostensibly be transferred by [WSADuplicateSocket](https://learn.microsoft.com/en-us/windows/win32/winsock/shared-sockets-2).

**Workaround**. socat or systemd-socket-proxyd. These aren't ideal for efficiency, security, or reliability.

**Notes**.
1. The Zig and C++ code correctly handle the socket type, which is usually just an fd, but isn't on Windows. The Javascript API is simple and oblivious, expecting positive ints to be passed.

2. This introduces an asymmetry between `listen()` and `connect()`, since now you can `listen()` to an fd, but you cannot `connect()` to one. Hence the somewhat awkward `FdOrUnixOrHost` type, which I don't think is worth fretting over.

3. Testing requires some Zig code to perform the `bind()` and `listen()` typically done out-of-process.

4. I copied the library changes to `_libusockets.h` and `libuwsockets.cpp`. These seem to differ in minor (unnecessary?) ways from the underlying libraries, for reasons I don't currently understand.

5. I haven't tried running this on Windows. Also, as @Jarred-Sumner [mentioned](https://github.com/uNetworking/uSockets/pull/192), some sanitization of fd might be useful. Also, I've never actually written any Zig code, so maybe double-check things.

**Building**. For politeness, I didn't redirect any submodules. Wait for the following PRs to be merged, or use my feature branches.
* https://github.com/Jarred-Sumner/uSockets/pull/3
* https://github.com/Jarred-Sumner/uWebSockets/pull/10

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -634,6 +634,9 @@ boringssl-debug: boringssl-build-debug boringssl-copy
 compile-ffi-test:
 	clang $(OPTIMIZATION_LEVEL) -shared -undefined dynamic_lookup -o /tmp/bun-ffi-test.dylib -fPIC ./test/js/bun/ffi/ffi-test.c
 
+.PHONY: compile-direct-fd-test
+compile-direct-fd-test:
+	zig build-lib ./test/js/bun/net/direct-fd-test.zig -dynamic -OReleaseFast -femit-bin=/tmp/libdirect-fd-test
 sqlite:
 
 
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -1513,6 +1513,18 @@ declare module "bun" {
     hostname?: string;
 
     /**
+     * Instead of binding and listening to a hostname and port, the server
+     * can operate off a socket which has already been bound and listened
+     * by a separate process. This enables "socket activated" deployments.
+     *
+     * @example
+     * ```js
+     * process.env.LISTEN_FDS // Use fd passed by systemd socket activation
+     * ```
+     */
+     fd?: string | number;
+
+    /**
      * What URI should be used to make {@link Request.url} absolute?
      *
      * By default, looks at {@link hostname}, {@link port}, and whether or not SSL is enabled to generate one
@@ -3030,8 +3042,9 @@ declare module "bun" {
 
   interface TCPSocketListenOptions<Data = undefined>
     extends SocketOptions<Data> {
-    hostname: string;
-    port: number;
+    fd?: number;
+    hostname?: string;
+    port?: number;
     tls?: TLSOptions;
   }
 
@@ -3075,6 +3088,7 @@ declare module "bun" {
    * @param options.data The per-instance data context
    * @param options.hostname The hostname to connect to
    * @param options.port The port to connect to
+   * @param options.fd The bound socket to attach to
    * @param options.tls The TLS configuration object
    * @param options.unix The unix socket to connect to
    *
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -219,6 +219,7 @@ const Handlers = struct {
 pub const SocketConfig = struct {
     hostname_or_unix: JSC.ZigString.Slice,
     port: ?u16 = null,
+    fd: ?uws.socket_t = null,
     ssl: ?JSC.API.ServerConfig.SSLConfig = null,
     handlers: Handlers,
     default_data: JSC.JSValue = .zero,
@@ -229,6 +230,7 @@ pub const SocketConfig = struct {
         globalObject: *JSC.JSGlobalObject,
         exception: JSC.C.ExceptionRef,
     ) ?SocketConfig {
+        var fd: ?uws.socket_t = null;
         var hostname_or_unix: JSC.ZigString.Slice = JSC.ZigString.Slice.empty;
         var port: ?u16 = null;
         var exclusive = false;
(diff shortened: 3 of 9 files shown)


shivak:
Aside: this is probably the safest way to deploy with good multicore performance. The key is SO_REUSEPORT, which Bun uses by default. But there are downsides to making each Bun worker handle its own socket binding:

- security: each worker is unprivileged, and has to keep the port open for additional workers. But then any user process can interfere in the pool.
- usability: https://github.com/oven-sh/bun/issues/1443

With `fd`, the privileged monitor gets sockets and passes them to only legitimate workers. With multiple workers, a monitor process is anyways necessary, since crashed workers need to be detected and restarted.

Jarred-Sumner:
Hey, this PR is great and really appreciate the work you put in and I
intend to merge it. Trying to ship the bundler so I haven’t had a chance to
review in depth yet but @cirospaciari can you take a look?

cirospaciari:
> 1. The Zig and C++ code correctly handle the socket type, which is usually just an fd, but isn't on Windows. The Javascript API is simple and oblivious, expecting positive ints to be passed.

Probably we will need an abstraction here when Windows support for Runtime release. 

> 2. This introduces an asymmetry between `listen()` and `connect()`, since now you can `listen()` to an fd, but you cannot `connect()` to one. Hence the somewhat awkward `FdOrUnixOrHost` type, which I don't think is worth fretting over.

The most common use case should be `listen()` we can work on `connect()` later if requested.

> 4. I copied the library changes to `_libusockets.h` and `libuwsockets.cpp`. These seem to differ in minor (unnecessary?) ways from the underlying libraries, for reasons I don't currently understand.

This API diverged slightly from the current C API of WebSockets because the C API itself was not done yet when Bun started to use it, and Bun required some changes to achieve some features (like sendfile) and add fixes. So `uws.zig` bindings actually use `libuwsockets.cpp` not the `capi/libuwebsockets.cpp`

> 5. I haven't tried running this on Windows. Also, as @Jarred-Sumner [mentioned](https://github.com/uNetworking/uSockets/pull/192), some sanitization of fd might be useful. Also, I've never actually written any Zig code, so maybe double-check things.

I'm working on Windows support, this can be added later.
 
> **Building**. For politeness, I didn't redirect any submodules. Wait for the following PRs to be merged, or use my feature branches.
> 
> * [Add support for already-listening sockets Jarred-Sumner/uSockets#3](https://github.com/Jarred-Sumner/uSockets/pull/3)
> * [Add support for already-listening sockets Jarred-Sumner/uWebSockets#10](https://github.com/Jarred-Sumner/uWebSockets/pull/10)

I take a look at `uSockets` and `uWebSockets` PR's and looks to be great. Probably we will need more testing like listening to the same socket twice on the same loop.

LGTM, thanks @shivak

shivak:
This PR touches some critical codepaths (creating any kind of server) and the underlying libs, so it is a good candidate to merge before a stable release. In the next few days, I'll resolve the conflicts that have arisen.

infrahead:
@Jarred-Sumner Is the plan for this to be merged still?  I'm hoping to migrate my Node.js web services to Bun, but using systemd sockets is how I architect things for zero-downtime deployments.  Alternatively I can look into SO_REUSEPORT which I know works with uWS, which on modern Linux kernels might also be part of a solution for zero-downtime setup, but before I look into it was curious on when might get fd support.

karimfromjordan:
I would also like to express my interest in this feature. In Node.js this isn't an issue.

e3dio:
This PR is needed for SystemD Socket Activation,  @shivak explained some benefits:

> Efficiency: servers can be started on demand in, e.g., large multi-tenant edge setups
> 
> Security: the socket can be passed into a restricted network namespace (container) where the server is run. This allows the server to respond to requests, but blocks all other network activity 
>
> Reliability: socket activation can implement failover

I would add **Performance**:

Rootless containers on Linux use a network layer (slirp4netns or pasta) which significantly impacts performance, passing a SystemD socket bypasses this and allows native networking performance inside the secure rootless container:

[benchmark (native/rootfull vs rootless)](https://github.com/containers/podman/discussions/22559) 

Since SystemD + containers is a very common deployment setup it makes sense to prioritize this feature:

```js
// use the socket passed by systemd

const server = Bun.serve({ fd });
```

Jarred-Sumner:
I would like for us to do this, but I think it'll need a new PR due to bitrot

Issue #1500, opened 2022-11-13
segfault with `Bun.spawnSync` with stdout: ArrayBufferView 


paperclover:
### What version of Bun is running?

0.2.2 (also on main branch as of creating the issue)

### What platform is your computer?

`Linux 6.0.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 03 Nov 2022 18:01:58 +0000 x86_64 x86_64`

### What steps can reproduce the bug?

Types say that `SpawnOptions.Readable` can be an `ArrayBufferView`

`echo "hello world" > test.txt`

```ts
const view: ArrayBufferView = new Uint8Array(100);
const proc = Bun.spawnSync({
  cmd: ["cat"],
  stdin: Bun.file("test.txt"),
  stdout: view,
  stderr: "inherit",
});
```

### How often does it reproduce? Is there a required condition?

every time.

### What is the expected behavior?

honestly, i'm not sure what i'm expecting to happen, maybe write the contents to the buffer. in the async spawn case which doesnt crash, **nothing is written to the buffer**, that might be another bug.

### What do you see instead?

Segfault at 0x8563000000000000

Electroid:
@robobun 

```js
import { writeFileSync } from "node:fs";
import { file } from "bun";

writeFileSync("./test.txt", "hello world", "utf-8");
const stdout = new Uint8Array(100);
const process = Bun.spawnSync({
  cmd: ["cat"],
  stdin: file("test.txt"),
  stdout,
  stderr: "inherit",
});
console.log(stdout);
```

robobun:
@Electroid here you go!
```tsx

```
<details>
  <summary>Code</summary>
  
```tsx
import { writeFileSync } from "node:fs";
import { file } from "bun";

writeFileSync("./test.txt", "hello world", "utf-8");
const stdout = new Uint8Array(100);
const process = Bun.spawnSync({
  cmd: ["cat"],
  stdin: file("test.txt"),
  stdout,
  stderr: "inherit",
});
console.log(stdout);
```
</details>

_Ran using the latest build of Bun, an all-in-one JavaScript runtime._

PR #2854, opened 2023-05-11
avoid unreacheable and fix tests to not hang

cirospaciari:


src/bun.js/api/bun/subprocess.zig
```diff
@@ -201,8 +201,8 @@ pub const Subprocess = struct {
                     };
                 },
                 .path => Readable{ .ignore = {} },
-                .blob, .fd => Readable{ .fd = @intCast(bun.FileDescriptor, fd) },
-                else => unreachable,
+                // TODO: array_buffer must be implemented yet but this will allow to proceed without faulting like in v0.5.9
```
cirospaciari, line 204: Related to https://github.com/oven-sh/bun/issues/1500

test/js/node/child_process/child_process-node.test.js
```diff
@@ -225,44 +225,40 @@ describe("ChildProcess spawn bad stdio", () => {
...
-        exitDone();
+        if (typeof expectData === "string") {
+          strictEqual(data?.trim(), expectData);
+        } else {
+          expect(data).toBeTruthy();
+        }
+        setTimeout(closeDone, 2);
       } catch (err) {
-        exitDone(err);
+        setTimeout(() => closeDone(err), 2);
```
cirospaciari, line 258: just a workaround to avoid errno: -32 when exiting the main process sooner, until I figure out how to solve this on macOS (only happens on tests, was unable to reproduce with `bun run`).

Jarred-Sumner: it wouldn't reproduce with `bun run` because it would only happen with a pipe

`bun run` is your shell's TTY

Jarred-Sumner: to reproduce it with bun run you would need to pipe it somewhere

PR #2859, opened 2023-05-12
`root` bundle option

dylan-conway:


src/bundler/bundle_v2.zig
```diff
@@ -3840,7 +3841,15 @@ const LinkerContext = struct {
                 const pathname = Fs.PathName.init(this.graph.entry_points.items(.output_path)[chunk.entry_point.entry_point_id].slice());
                 chunk.template.placeholder.name = pathname.base;
                 chunk.template.placeholder.ext = "js";
-                chunk.template.placeholder.dir = pathname.dir;
+
+                var dir = std.fs.cwd().openDir(pathname.dir, .{}) catch |err| {
```
Jarred-Sumner, line 3845: 1) Do we need to do this? Opening directories is expensive. If the original file was a symlink, it would produce unexpected results - it would become the realpath instead of the input path. Instead, IMO, we should join the input path with the root path and choose based on that.
2) If we do need to do this, don't use `openDir` because it asks for more permissions than necessary on Linux. Instead, we should add a version of `openFileForPath` that adds the `O.DIRECTORY` flag.

 https://github.com/oven-sh/bun/blob/a008e49a40565c1481b4216bfa9d60016de8b568/src/bun.zig#L1501

src/bun.js/api/JSBundler.zig
```diff
@@ -151,6 +154,33 @@ pub const JSBundler = struct {
                 return error.JSException;
             }
 
+            {
+                const path: ZigString.Slice = brk: {
+                    if (try config.getOptional(globalThis, "root", ZigString.Slice)) |slice| {
```
Jarred-Sumner, line 159: What's the difference between `root` and `dir`? I think we should leave a comment explaining this because everyone will forget

src/bun.js/api/JSBundler.zig
```diff
@@ -151,6 +154,33 @@ pub const JSBundler = struct {
...
+                    if (entry_points.len == 1) {
+                        break :brk ZigString.Slice.fromUTF8NeverFree(std.fs.path.dirname(entry_points[0]) orelse ".");
+                    }
+
+                    break :brk ZigString.Slice.fromUTF8NeverFree(resolve_path.getIfExistsLongestCommonPath(entry_points) orelse ".");
+                };
+
+                defer path.deinit();
+
+                var dir = std.fs.cwd().openDir(path.slice(), .{}) catch |err| {
```
Jarred-Sumner, line 174: Ditto for this call to openDir, should use the more specific function adding the O_PATH flag.

Jarred-Sumner (commented):
close

src/bundler/bundle_v2.zig
Jarred-Sumner: re: symlink, another option is to pass the flag to not follow symlink path - so it would open the symlink instead of the real path

additionally, we should have a hash table here that checks for duplicate dirname

Jarred-Sumner:
ill finish it

PR #2860, opened 2023-05-12
fix(fetch:proxy) return correct status when permission is dennied

cirospaciari:


Jarred-Sumner (approved):
LGTM once tests on macOS run

PR #2861, opened 2023-05-12
Allow reading config from within plugins, and partially implement esbuild `initialOptions`

paperclover:
Fixes some esbuild plugins and gives bun-specific plugins full ability to edit the BuildConfig object.

`.initialOptions` is intentionally undocumented. we do not want people using this, but rather just be available as a compatibility layer.


src/bun.js/builtins/cpp/BundlerPluginBuiltins.cpp
```diff
@@ -327,11 +327,27 @@ const char* const s_bundlerPluginRunSetupFunctionCode =
...
+    "    config,\n" \
     "    onDispose,\n" \
     "    onEnd,\n" \
     "    onLoad,\n" \
     "    onResolve,\n" \
     "    onStart,\n" \
+    "    //\n" \
+    "    initialOptions: {\n" \
+    "      bundle: true,\n" \
+    "      entryPoints: config.entrypoints ?? config.entryPoints,\n" \
```
Jarred-Sumner: if the user messes this up, they'll get a weird error in the plugin instead of a more helpful error saying this field is missing

i suggest `?? []` at the end

Jarred-Sumner (approved):
LGTM

src/bun.js/builtins/cpp/BundlerPluginBuiltins.cpp
paperclover: image
i know what you mean but that error doesn't happen. ill add it anyways since it should definitely exist.

PR #2866, opened 2023-05-12
fix(spawn) add Uint8Array support for stdout

cirospaciari:
Fix: #1500

### DIFF
--- a/test/js/bun/spawn/spawn.test.ts
+++ b/test/js/bun/spawn/spawn.test.ts
@@ -162,6 +162,38 @@ for (let [gcTick, label] of [
         }
       });
 
+      it("ArrayBuffer works as stdout", () => {
+        gcTick();
+        const stdout_buffer = new Uint8Array(11);
+        const { stdout } = spawnSync(["echo", "hello world"], {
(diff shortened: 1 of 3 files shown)


test/js/bun/spawn/spawn.test.ts
```diff
@@ -162,6 +162,38 @@ for (let [gcTick, label] of [
...
+        });
+        gcTick();
+        const text = new TextDecoder().decode(stdout);
+        const text2 = new TextDecoder().decode(stdout_buffer);
+        expect(text.trim()).toBe("hello world");
+        expect(text2.trim()).toBe("hello world");
+        gcTick();
+      });
+
+      it("Uint8Array works as stdout when is smaller than output", () => {
```
Jarred-Sumner, line 181: what happens when the output is (1) larger or (2) exactly sized? those cases need tests

cirospaciari: here smaller, larger, or exactly sized https://github.com/oven-sh/bun/pull/2866/commits/64d536fd05ed8a81aa6bfeab2b590b57fb279906

Issue #2722, opened 2023-04-23
Bun can’t run module with follow-redirects / axios bundled (`contentful-management` package)


controversial:
tl;dr: basic usage of the `contentful-export` package throws an error from inside `axios` internals.

### What version of Bun is running?

0.6.0

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

test.ts:
```ts
import contentfulExport from 'contentful-export';

const contentModel = await contentfulExport({
  spaceId: 'abc',
  managementToken: 'abc',
});

console.log(contentModel);
```
(Note these are garbage values for `spaceId` and `managementToken`, but the same issue occurs with real values)

Run `bun test.ts`.

### What is the expected behavior?

The script runs and prints the final result (or, in the case of garbage data as above, reports a 404 for “the resource could not be found”).

### What do you see instead?

```
❯ bun test.ts
[0.17ms] ".env"
  ✔ Initialize client (1s)
  ❯ Fetching data from space
    ✖ Connecting to space (1s)
      → TypeError: socket.setTimeout is not a function. (In 'socket.setTimeout(m
secs)', 'socket.setTimeout' is undefined)…
      Fetching content types data
      Fetching tags data
      Fetching editor interfaces data
      Fetching content entries data
      Fetching assets data
      Fetching locales data
      Fetching webhooks data
      Fetching roles data
    Download assets
    Write export log file


The following 1 errors and 0 warnings occurred:

15:56:09 - Error: TypeError: socket.setTimeout is not a function. (In 'socket.setTimeout(msecs)', 'socket.setTimeout' is undefined)
```

### Additional information

`contentful-export` requires `contentful-management`, which distributes a webpack bundle called `contentful-management.node.js`.

As far as I can tell, this `socket.setTimeout` error comes from inside code bundled for the [`follow-redirects`](https://www.npmjs.com/package/follow-redirects) package, which is used internally by `axios`.

Jarred-Sumner:
ah we are missing `socket.setTimeout` somewhere

controversial:
I can no longer reproduce this issue on latest canary

PR #2494, opened 2023-03-27
feat(socket.io) compatibilty with workarounds

cirospaciari:
This draft can make this example of socket.io run:

```javascript
const html = `
<!DOCTYPE html>
<html>
<head>
  <title>Socket.IO chat</title>
  <style>
    body { margin: 0; padding-bottom: 3rem; font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif; }

    #form { background: rgba(0, 0, 0, 0.15); padding: 0.25rem; position: fixed; bottom: 0; left: 0; right: 0; display: flex; height: 3rem; box-sizing: border-box; backdrop-filter: blur(10px); }
    #input { border: none; padding: 0 1rem; flex-grow: 1; border-radius: 2rem; margin: 0.25rem; }
    #input:focus { outline: none; }
    #form > button { background: #333; border: none; padding: 0 1rem; margin: 0.25rem; border-radius: 3px; outline: none; color: #fff; }

    #messages { list-style-type: none; margin: 0; padding: 0; }
    #messages > li { padding: 0.5rem 1rem; }
    #messages > li:nth-child(odd) { background: #efefef; }
  </style>
</head>
<body>
  <ul id="messages"></ul>
  <form id="form" action="">
    <input id="input" autocomplete="off" /><button>Send</button>
  </form>
  <script src="https://cdn.socket.io/4.5.4/socket.io.min.js"></script>
<script>
var socket = io();

var messages = document.getElementById('messages');
var form = document.getElementById('form');
var input = document.getElementById('input');

form.addEventListener('submit', function(e) {
  e.preventDefault();
  if (input.value) {
    socket.emit('chat message', input.value);
    input.value = '';
  }
});

socket.on('chat message', function(msg) {
  var item = document.createElement('li');
  item.textContent = msg;
  messages.appendChild(item);
  window.scrollTo(0, document.body.scrollHeight);
});
</script>
</body>
</html>

`;
const app = require('express')();
const http = require("http").createServer(app);
const io = require('socket.io')(http, {
  wsEngine:  require("./node_modules/ws/index.js").Server 
});
const port = process.env.PORT || 3000;
app.get("/", (req, res)=> res.send(html));
io.on('connection', (socket) => {
  socket.on('chat message', msg => {
    io.emit('chat message', msg);
  });
});

http.listen(port, () => {
  console.log(`Socket.IO server running at http://localhost:${port}/`);
});
```

Pending:
[...]

PR #2868, opened 2023-05-12
proper stub for http socket

cirospaciari:
Today we have a stub for Socket on `http` module, so axios sometimes errors trying to use `setTimeout` like in #2722  and it also includes a step further for `socket.io` like in https://github.com/oven-sh/bun/pull/2494/files, not sure if we should stub or extends `Duplex`

### DIFF
--- a/src/bun.js/http.exports.js
+++ b/src/bun.js/http.exports.js
@@ -1,5 +1,5 @@
 const { EventEmitter } = import.meta.require("node:events");
-const { Readable, Writable } = import.meta.require("node:stream");
+const { Readable, Writable, Duplex } = import.meta.require("node:stream");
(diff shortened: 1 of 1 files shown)


src/bun.js/http.exports.js
```diff
@@ -40,16 +40,85 @@ var _globalAgent;
...
+
+  setTimeout(timeout, callback) {
     return this;
   }
-  removeListener() {}
-  removeAllListeners() {}
+
+  unref() {}
+
+  _write(chunk, encoding, callback) {}
```
Jarred-Sumner, line 121: is this return type correct

cirospaciari: Yes on our `node:net` this does not return anything

PR #2869, opened 2023-05-12
windows path compatibility

paperclover:
`export-*-from.ts` -> `export-star-from.ts`

makes it so the repo is clonable on windows, though note we do not build on windows yet. opens the door for others to at least mess around.


cirospaciari (approved):
LGTM

PR #2872, opened 2023-05-13
handle printing missing expressions and add tests

dylan-conway:
we could add an `ArrayHole` ast node, but some testing showed we have all cases covered in the printer except for `if/if else/else`


src/js_printer.zig
```diff
@@ -4936,10 +4936,19 @@ fn NewPrinter(
...
                     }
+
+                    if (s.yes.isMissingExpr()) {
+                        p.printSemicolonIfNeeded();
+                    }
                 },
             }
 
             if (s.no) |no_block| {
+                if (no_block.isMissingExpr() and p.options.minify_syntax) {
```
Jarred-Sumner: are you sure we can't fix this in the parser so that this never happens in the first place?

dylan-conway: moved the fix to the parser

src/js_parser.zig
```diff
@@ -18015,8 +18016,12 @@ fn NewParser_(
 
                     data.value = p.visitExpr(data.value);
 
+                    if (should_trim_primitive and data.value.isPrimitiveLiteral()) {
+                        return;
+                    }
+
                     // simplify unused
-                    data.value = SideEffects.simpifyUnusedExpr(p, data.value) orelse data.value.toEmpty();
+                    data.value = SideEffects.simpifyUnusedExpr(p, data.value) orelse return;
```
Jarred-Sumner, line 18024: we should always simplify unused im pretty sure

src/js_parser.zig
```diff
@@ -18015,8 +18016,12 @@ fn NewParser_(
 
                     data.value = p.visitExpr(data.value);
 
+                    if (should_trim_primitive and data.value.isPrimitiveLiteral()) {
```
Jarred-Sumner, line 18019: this condition will never be true

src/js_parser.zig
```diff
@@ -19551,6 +19588,21 @@ fn NewParser_(
         }
 
         fn visitSingleStmt(p: *P, stmt: Stmt, kind: StmtsKind) Stmt {
+            if (stmt.data == .s_block) {
+                var new_stmt = stmt;
+                p.pushScopeForVisitPass(.block, stmt.loc) catch unreachable;
+                var stmts = ListManaged(Stmt).initCapacity(p.allocator, stmt.data.s_block.stmts.len) catch unreachable;
+                stmts.appendSlice(stmt.data.s_block.stmts) catch unreachable;
```
Jarred-Sumner, line 19595: im confused what this is doing? flattening the block?

src/js_parser.zig
Jarred-Sumner: oh it would if data.value was simplified

src/js_parser.zig
dylan-conway: we append it if it can be simplified, or we return and don't append the stmt. this was actually the code that replaced empty blocks with `e_missing` and made empty if/while/for statements look like this after building:
```js
if (r)
    ;
else if (r)
    ;
```
this is fine, but doesn't work when minifying syntax and whitespace later on

Jarred-Sumner: oh got it cool

src/js_parser.zig
dylan-conway: yeah, this is for flattening blocks following for/if/while statements

Jarred-Sumner:
thank you

PR #2874, opened 2023-05-13
feat(Timer.refresh) add refresh support on Timer

cirospaciari:
This is part of `socket.io` support


src/bun.js/api/bun.zig
```diff
@@ -2990,13 +2990,111 @@ pub const Timer = struct {
         id: i32 = -1,
         kind: Timeout.Kind = .setTimeout,
         ref_count: u16 = 1,
+        // we need this information because we can refresh it after it has ended
+        arguments: JSC.Strong = .{},
+        callback: JSC.Strong = .{},
```
Jarred-Sumner: Instead of storing these as `JSC.Strong` - a separate heap allocation that must be tracked by the GC separately

We can track it with the GC using `values` in `node.classes.ts`:

image


You would set the `arguments` via `TimerObject.argumentsValueSetCached(timerJSValue, globalThis, value)`

Jarred-Sumner: An example:

https://github.com/oven-sh/bun/blob/44be22ca98ca0cea5eba1f804c0aca4b86784114/src/bun.js/test/jest.zig#L843

cirospaciari: Cool feature good to know about it here is the update https://github.com/oven-sh/bun/pull/2874/commits/34d88a061fa95c625887e6b7b6ee8345e184ee96

PR #2885, opened 2023-05-15
Fix running commonjs modules as entry point

paperclover:
fixes `bun run` with the following
```ts
this;
console.log("Hello World");
```
or more specifically this generated snippet
```ts
export default __cJS2ESM(() => console.log("Hello World"))
```

### DIFF
--- a/src/bundler/entry_points.zig
+++ b/src/bundler/entry_points.zig
@@ -192,7 +192,7 @@ pub const ServerEntryPoint = struct {
                     \\var cjs = start?.default;
                     \\if (cjs && typeof cjs ===  'function' && cjsSymbol in cjs) {{
                     \\  // if you module.exports = (class {{}}), don't call it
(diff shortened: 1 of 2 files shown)


src/bundler/entry_points.zig
```diff
@@ -192,7 +192,7 @@ pub const ServerEntryPoint = struct {
                     \\var cjs = start?.default;
                     \\if (cjs && typeof cjs ===  'function' && cjsSymbol in cjs) {{
                     \\  // if you module.exports = (class {{}}), don't call it
-                    \\  entryNamespace = ("prototype" in cjs) ? cjs : cjs();
+                    \\  entryNamespace = import.meta.primordials.isConstructor(cjs.constructor) ? cjs : cjs();
```
Jarred-Sumner: is it `.isConstructor` on the constructor function or on the type itself?

paperclover:
i could not get it working with isConstructor. it thinks a standard function is a constructor, which i guess is true. but not a class constructor.

PR #2888, opened 2023-05-15
Add plugins for yaml & RSC

colinhacks:
This can be merged but `JSBundler` hasn't yet been updated to expect the new `bun-plugin-server-components` structure

Implements `bun-plugin-server-components` and `bun-plugin-yaml`. 

Adds directories (unimplemented) for:
- `bun-plugin-lightningcss`
- `bun-plugin-css`
- `bun-plugin-mdx`


Jarred-Sumner:
yay

TommasoAmici:
@colinhacks I'm interested in taking a stab at `bun-plugin-lightningcss` (I was actually about to start one from scratch until I realized there's empty directories in here), is there any documentation as to what behavior is expected for this plugin?

EDIT: I opened a draft here https://github.com/oven-sh/bun/pull/3759

PR #2890, opened 2023-05-16
change the generated `debugId` suffix

paperclover:
`BBBBBBBBBBBBBBBB` (b is for bun) -> `64756e2164756e21` (hex encoded "bun!bun!")

the reason we have a constant string at all is because debug Ids must be 32 hex characters (aka a UUID)


Jarred-Sumner:
Thank you

PR #2904, opened 2023-05-16
fix asset naming output paths

dylan-conway:
fixes:
```
naming/AssetNamingMkdir
naming/AssetNamingDir
```


src/bundler/bundle_v2.zig
```diff
@@ -8901,6 +8901,11 @@ const LinkerContext = struct {
 
         const root_path = c.resolver.opts.output_dir;
 
+        if (root_path.len == 0 and c.parse_graph.additional_output_files.items.len > 0) {
```
Jarred-Sumner: Did you check if this works with `--compile`? We technically can support that without an output directory

dylan-conway: added an option for `--compile`

PR #2913, opened 2023-05-17
Fix streams breaking on reverted EventEmitter / Make Discord.js work

### LINKED ISSUE #2077: Support `discord.js`
We've long had requests for `discord.js` support. It doesn't seem too far away now that we've got `undici.Agent` and a workaround for the `worker_threads` requirement.

Currently it looks like we need the following to finish support for `discord.js`:
- [x] `undici.request` (tracking in #1987)
- [x] `response.arrayBuffer()` hangs when called more than once (no tracking yet)

### LINKED ISSUE #2707: `node:crypto` createHash or `node:stream`, stream isn't readable

0.6.0

_No response_

```ts
  it("stream (async)", done => {
    const hash = crypto.createHash("sha256");
    hash.on("readable", () => {
      const data = hash.read();
      if (data) {
        expect(data.toString("hex")).toBe("6a2da20943931e9834fc12cfe5bb47bbd9ae43489a30726962b576f4e3993e50");
        done();
      }
    });
    hash.write("some data to hash");
    hash.end();
  });
```

_No response_

### What do you see instead?

infinite hang

### Additional information

caused by #2604 i think

paperclover:
Closes #2707, Closes #2077

fixes some of the things that were reverted before

- crypto improvement with lazy init stream
- js implementation for node events
- cjs entrypoints

in addition to this, it makes discord.js work

- ws changes
- undici and builtins bug fix

note for the first one, it does cause slower runtime with

```
const hasher = createHash(alg);
hasher.write(data);
hasher.end();
hasher.read();
```

and we could totally optimize this simple use case too but i did not do that.

there were two fixes to this, and i went with (2)

image

1 = current
2 = lazy init + old event emitter from 0.6 and prior
3 = lazy init + js event emitter + pure js streams


src/bundler/entry_points.zig
```diff
@@ -192,7 +192,7 @@ pub const ServerEntryPoint = struct {
                     \\var cjs = start?.default;
                     \\if (cjs && typeof cjs ===  'function' && cjsSymbol in cjs) {{
                     \\  // if you module.exports = (class {{}}), don't call it
-                    \\  entryNamespace = ("prototype" in cjs) ? cjs : cjs();
+                    \\  entryNamespace = import.meta.primordials.isCallable(cjs) ? cjs() : cjs;
```
Jarred-Sumner: this shouldn't be in here

src/bun.js/events.exports.js
```diff
@@ -0,0 +1,464 @@
...
+  return events[type]?.length ?? 0;
+};
+
+EventEmitter.prototype.eventNames = function eventNames() {
+  return this._eventsCount > 0 ? Reflect.ownKeys(this._events) : [];
+};
+
+EventEmitter.prototype[kCapture] = false;
+
+function once(emitter, type, { signal } = {}) {
```
Jarred-Sumner: we can't set a default argument in the function this way

people will pass `null` or something truthy and then get a very strange error

src/bun.js/events.exports.js
```diff
@@ -0,0 +1,464 @@
...
+      reject(new AbortError(undefined, { cause: signal?.reason }));
+    }
+    if (signal != null) {
+      eventTargetAgnosticAddListener(signal, "abort", abortListener, { once: true });
+    }
+  });
+}
+EventEmitter.once = once;
+
+function on(emitter, type, { signal, close, highWatermark = Number.MAX_SAFE_INTEGER, lowWatermark = 1 } = {}) {
```
Jarred-Sumner: ditto

src/bun.js/events.exports.js
```diff
@@ -0,0 +1,464 @@
...
+
+  this._maxListeners ??= undefined;
+  if (
+    (this[kCapture] = opts?.captureRejections ? Boolean(opts?.captureRejections) : EventEmitter.prototype[kCapture])
+  ) {
+    this.emit = emitWithRejectionCapture;
+  }
+}
+
+EventEmitter.prototype._events = undefined;
```
Jarred-Sumner: instead of 

```js
EventEmitter.prototype
```

People might mess with it at runtime

Instead, lets assign it to a local variable and use that one

```js
const EventEmitterPrototype = EventEmitter.prototype;
```

PR #2920, opened 2023-05-17
do not fail bundles on warnings

paperclover:
This makes it so we do not success=false or status code 1 if we have warnings and no errors.

**context from discord**:

image


Jarred-Sumner:
Thank you

Issue #2922, opened 2023-05-17
mongodb UNABLE_TO_GET_ISSUER_CERT_LOCALLY with 0.6.1


RUSHt:
### What version of Bun is running?

0.6.1

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

This works OK in 0.5.9, but fails when bun upgrade > 0.6.1.  Reverted with npm i bun@0.5.9 and works again.

```
        console.log('connect')
        const _DB = 'mongodb+srv://'+DBuser+':'+DBpw+Cluster+DBdefault+'?retryWrites=true&w=majority';
        const client = new MongoClient(_DB);
        console.log('got client')
        await client.connect();
        console.log('got connect')
```

Output on fail is:-

```
connect
got client
94 |         if (message instanceof Error) {
95 |             super(message.message);
96 |             this.cause = message;
97 |         }
98 |         else {
99 |             super(message);
                ^
error: UNABLE_TO_GET_ISSUER_CERT_LOCALLY
 code: "undefined"

      at new MongoError (/Users/<user>/Dev/<app>/node_modules/mongodb/lib/error.js:99:12)
      at new MongoSystemError (/Users/<user>/Dev/<app>/node_modules/mongodb/lib/error.js:589:12)
      at new MongoServerSelectionError (/Users/<user>/Dev/<app>/node_modules/mongodb/lib/error.js:611:8)
      at /Users/<user>/Dev/<app>/node_modules/mongodb/lib/sdam/topology.js:278:37
```

### What is the expected behavior?

If it worked I'd expect to see:-

```
connect
got client
got connect
```

### What do you see instead?

```
connect
got client
94 |         if (message instanceof Error) {
95 |             super(message.message);
96 |             this.cause = message;
97 |         }
98 |         else {
99 |             super(message);
                ^
error: UNABLE_TO_GET_ISSUER_CERT_LOCALLY
 code: "undefined"

      at new MongoError (/Users/<user>/Dev/<app>/node_modules/mongodb/lib/error.js:99:12)
      at new MongoSystemError (/Users/<user>/Dev/<app>/node_modules/mongodb/lib/error.js:589:12)
      at new MongoServerSelectionError (/Users/<user>/Dev/<app>/node_modules/mongodb/lib/error.js:611:8)
      at /Users/<user>/Dev/<app>/node_modules/mongodb/lib/sdam/topology.js:278:37
```

### Additional information

[...]

Issue #2737, opened 2023-04-25
`socket.io` silently fails when connecting to wss endpoints


MattJustMatt:
### What version of Bun is running?

0.6.0 (canary)

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

<img width="808" alt="image" src="https://user-images.githubusercontent.com/5265661/234271620-d8107679-985e-4687-955f-b17b73b7b84e.png">

NGINX (in front of the SIO server) responds with "400 The plain HTTP request was sent to HTTPS port"

When forcing transport to websockets (no HTTP long polling/XHR):
<img width="1903" alt="image" src="https://user-images.githubusercontent.com/5265661/234273218-ceee8876-23a8-406e-a7eb-81b453b69320.png">

The connection gets stuck in the opening ready state. 

With #2494 non TLS (ws://) endpoints work well (both in XHR+WS transports), but wss:// does not. 

Thanks for looking!

MattJustMatt:
With further testing, ws:// only works with long polling. Forcing the transport to 'websocket' gets stuck in 'opening' state, same as wss://

Issue #2485, opened 2023-03-25
Support `socket.io`


ThatOneBro:
This issue is for tracking progress towards getting `socket.io` working in Bun.

TODO:
* [ ] Investigate why it's not working

fucksophie:
Investigating, there is a very interesting bug I ran into.
```js
    let socketIo = io("https://supplyyourownserver.com");
    
    setInterval( () =>{
      console.log(socketIo.io.engine.readyState)
    }, 10);

    socketIo.on("connect", () => {
        console.log("connected")
    });

    socketIo.on("disconnect", () => {
      console.log("disconnect")
    })
```
This exact code never fires `connect` or `disconnect`, but the readyState for the eio engine switches between opening and closed. 

I think this might be due to the `ws` NPM module clashing with Bun's `WebSocket` implementation.

privatenumber:
It's erroring because [`socket.io` dependency `engine.io` relies on `ws.Server`](https://unpkg.com/browse/engine.io@6.4.1/build/server.js#L28:~:text=wsEngine%3A%20ws_1.Server,29), which Bun doesn't export: https://github.com/oven-sh/bun/blob/main/src/bun.js/ws.exports.js

I can correct the exports so it at least has a verbose error message.

fifididi:
Moved to bun recently, and I am facing a bug that I wasn't facing when using node.

socket.io-client@1.4.5 (I have to use this version, out of my control)
ws@^8.8.1

Here's how the client is initialized: 
```js
const socket = io("https://server.com", { origins: "*:*", });
```

And here is the relevant debug information:
<details> <summary> debug.log </summary>

```
end {
  "objectMode": false,
  "highWaterMark": 16384,
  "finalCalled": false,
  "needDrain": false,
  "ending": false,
  "ended": false,
  "finished": false,
  "destroyed": false,
  "decodeStrings": true,
  "defaultEncoding": "utf8",
  "writing": false,
  "corked": 0,
  "sync": true,
  "bufferProcessing": false,
  "onwrite": [Function: onwrite],
  "writecb": null,
  "writelen": 0,
  "afterWriteTickInfo": null,
  "buffered": [],
  "bufferedIndex": 0,
  "allBuffers": true,
  "allNoop": true,
  "pendingcb": 0,
  "constructed": true,
  "prefinished": false,
  "errorEmitted": false,
  "emitClose": true,
  "autoDestroy": true,
  "errored": null,
  "closed": false,
[...]

Issue #2355, opened 2023-03-09
Timeout: Cannot connect to remote mongodb via a IP Address.


ltcaosj:
### What version of Bun is running?

0.5.7

### What platform is your computer?

Mac M1

### What steps can reproduce the bug?

This code does not work.
import { MongoClient } from "mongodb";

const client = new MongoClient("mongodb://192.168.1.200:27017");
await client.connect();
const db = client.db("test");
const collection = db.collection("test");
await collection.insertOne({ hello: "world" });
const result = await collection.findOne({ hello: "world" });

console.log(JSON.stringify(result));

### What is the expected behavior?

Can connect to remote mongodb succesfully.

### Additional information

Same code,  but if I connect to mongodb at localhost - it's working fine.

import { MongoClient } from "mongodb";

const client = new MongoClient("mongodb://localhost:27017");
await client.connect();
const db = client.db("test");
const collection = db.collection("test");
await collection.insertOne({ hello: "world" });
const result = await collection.findOne({ hello: "world" });

console.log(JSON.stringify(result));

Angusross111:
Im also getting the same problem. It seems it exits quietly at the first await. eg `await client.connect();`. 
Looking in the code I think it fails here: `await promisify(callback => topology.connect(options, callback))();` in `mongo_client.ts`
I tested it with bun `v0.5.6`, `v0.5.7`, and `v0.5.8-canary`. Also tested it with mongodb `v5.1`, `v5.0` and `v4.14`

swagisays:
got same issue

PR #2934, opened 2023-05-17
fix(tls.connect) fix SNI on tls sockets and also servername (mongodb)

cirospaciari:
Depends on https://github.com/Jarred-Sumner/uSockets/pull/4 and https://github.com/Jarred-Sumner/uWebSockets/pull/11

Fix: #2922 
Fix: #2355 
Fix: #1991 
Fix: #2737
Fix: #2485
Fix: #313 (fastify will at least start after this)
Fix: #2955


Jarred-Sumner:
lets merge once macOS tests run

Jarred-Sumner:
macOS x64 test is timing out

src/bun.js/api/bun/socket.zig
```diff
@@ -1651,6 +1674,11 @@ fn NewSocket(comptime ssl: bool) type {
             if (!this.socket.isClosed()) {
                 this.socket.close(0, null);
             }
+            if (this.connection != null) {
```
Jarred-Sumner: nitpick:


```suggestion
            if (this.connection) |connection| {
                 this.connection = null;
                 connection.deinit();
            }
```

src/bun.js/ws.exports.js
```diff
@@ -6,19 +6,51 @@ import EventEmitter from "events";
...
+  }
   on(event, callback) {
     if (event === "message") {
-      var handler = ({ data }) => {
+      var handler = ({ message }) => {
         try {
-          callback(data);
+          if (typeof message === "string") {
+            if (this.#binaryType === "arraybuffer") {
+              const encoder = new TextEncoder();
```
Jarred-Sumner: can avoid this overhead by re-using the `TextEncoder` instance at the top

Jarred-Sumner: actually

we can set `binaryType` on the WebSocket itself.

Jarred-Sumner: right?

cirospaciari: We only support `arraybuffer` rn on the `WebSocket.cpp`, so the idea was add this wrapper and follow up with native

src/bun.js/api/server.zig
```diff
@@ -91,6 +91,58 @@ const SendfileContext = struct {
...
+                        JSC.throwInvalidArguments(std.fmt.comptimePrint("Unable to access {s} path", .{fieldname}), .{}, global, exception);
+                        return .{ .data = null };
+                    }
+                    const cwd = std.process.getCwdAlloc(bun.default_allocator) catch {
+                        JSC.throwInvalidArguments(std.fmt.comptimePrint("Unable to access {s} path", .{fieldname}), .{}, global, exception);
+                        return .{ .data = null };
+                    };
+                    defer bun.default_allocator.free(cwd);
+
+                    const fullpath = std.fs.path.resolve(bun.default_allocator, &.{ cwd, path }) catch {
```
Jarred-Sumner: `std.fs.path.resolve` doesn't actually go through the filesystem so it doesn't do as intended

Jarred-Sumner: instead, use `bun.getFdPath`

src/bun.js/api/server.zig
```diff
@@ -91,6 +91,58 @@ const SendfileContext = struct {
...
+                if (pathlike == .path) {
+                    const path = pathlike.path.slice();
+
+                    const path_z = bun.default_allocator.dupeZ(u8, path) catch unreachable;
+                    defer bun.default_allocator.destroy(path_z);
+                    if (std.os.system.access(path_z, std.os.F_OK) != 0) {
+                        JSC.throwInvalidArguments(std.fmt.comptimePrint("Unable to access {s} path", .{fieldname}), .{}, global, exception);
+                        return .{ .data = null };
+                    }
+                    const cwd = std.process.getCwdAlloc(bun.default_allocator) catch {
```
Jarred-Sumner: this is unnecessary, we can use `Fs.FilesSystem.instance.top_level_dir`

src/bun.js/api/server.zig
```diff
@@ -91,6 +91,58 @@ const SendfileContext = struct {
...
 const linux = std.os.linux;
 
+const BlobFileContentResult = struct {
+    data: [*c]const u8,
+
+    const log = Output.scoped(.BlobFileContentResult, false);
+    fn init(comptime fieldname: []const u8, js_obj: JSC.JSValue, global: *JSC.JSGlobalObject, exception: JSC.C.ExceptionRef) ?BlobFileContentResult {
+        if (JSC.WebCore.Body.Value.fromJS(global, js_obj)) |body| {
+            if (body == .Blob and body.Blob.store != null and body.Blob.store.?.data == .file) {
+                const pathlike = body.Blob.store.?.data.file.pathlike;
```
Jarred-Sumner: This entire function should be removed and replaced with a version of `readFile` easier to call from Zig,  like `writeFileWithPathBuffer`.

https://github.com/oven-sh/bun/blob/1973975ce0f56cea236697fcab36d8e2a211ac70/src/bun.js/node/node_fs.zig#L3290-L3428

src/bun.js/api/server.zig
Jarred-Sumner: ignore this, see above

src/bun.js/api/server.zig
```diff
@@ -91,6 +91,58 @@ const SendfileContext = struct {
 const DateTime = bun.DateTime;
 const linux = std.os.linux;
 
+const BlobFileContentResult = struct {
+    data: [*c]const u8,
```
Jarred-Sumner: don't use a C pointer, it's ambiguous and crashy

use `:0` or just `[*]`

src/bun.js/api/server.zig
cirospaciari: done on https://github.com/oven-sh/bun/pull/2934/commits/5d2830353e60d42f869b4b0e8571c9f763b473f3 fixed throw on https://github.com/oven-sh/bun/pull/2934/commits/52daa77765591ed061eefb2bea48463eb82ff30b

src/bun.js/node/node_fs.zig
```diff
@@ -3414,10 +3455,20 @@ pub const NodeFS = struct {
...
+                        if (comptime string_type == .default) {
+                            break :brk .{
+                                .result = .{
+                                    .string = buf.items,
+                                },
+                            };
+                        } else {
+                            break :brk .{
+                                .result = .{
+                                    .null_terminated = buf.toOwnedSliceSentinel(0) catch unreachable,
```
Jarred-Sumner, line 3468: It’s fine for now but keep in mind that .toOwnedSlice potentially reallocates the array

Jarred-Sumner:
Thank you

PR #2937, opened 2023-05-18
use raw template contents for tagged templates

### LINKED ISSUE #622: `String.raw` produces incorrect output
### Version

0.1.3

### Platform

Linux malus 5.17.15-1-MANJARO #1 SMP PREEMPT Wed Jun 15 07:09:31 UTC 2022 x86_64 GNU/Linux

### What steps will reproduce the bug?

Create a file with following contents:
```js
const regex = new RegExp(String.raw`(sig|signature|AWSAccessKeyId|password|passcode)=(\?P<secret>[a-z0-9%/+]{16,})`, 'gi');
console.log(regex.test('asdf'));
```
Run it with `bun run file.js`

It happens every time, the code snippet is from snowflake sdk and it looks like the `=(\?` is the problem.

It should be able to parse the regex (altough it seems wierd, my guess is that the last part is supposed to be something like `(?<secret>`, but node parses it correctly and it works), same as `node file.js`:
```console
$ node file.js
true
```

### What do you see instead?

```console
$ bun run test.ts
1 | const regex = new RegExp(String.raw`(sig|signature|AWSAccessKeyId|password|passcode)=(?P<secret>[a-z0-9%/+]{16,})`, "gi");
                 ^
 SyntaxError: Invalid regular expression: unrecognized character after (?
      at ./test.ts:1:14
```

### Additional information

_No response_

dylan-conway:
fixes #622

### DIFF
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -2359,11 +2359,13 @@ pub const E = struct {
         value: ExprNodeIndex,
         tail_loc: logger.Loc,
         tail: E.String,
+        tail_raw: ?string = null,
     };
 
     pub const Template = struct {
         tag: ?ExprNodeIndex = null,
         head: E.String,
+        head_raw: ?string = null,
         parts: []TemplatePart = &([_]TemplatePart{}),
 
         /// "`a${'b'}c`" => "`abc`"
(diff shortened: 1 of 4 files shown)


Jarred-Sumner (approved):

Jarred-Sumner:
Is there a case where there’s a tag but it’s not cooked? I wonder we can get true of the optional on the tag too

PR #2961, opened 2023-05-19
Implement `test.todo`

blackmann:
Progress for #1825


src/cli/test_command.zig
```diff
@@ -567,10 +596,8 @@ pub const TestCommand = struct {
                 Output.prettyError(" {d:5>} expect() calls\n", .{reporter.summary.expectations});
             }
 
-            Output.prettyError("Ran {d} tests across {d} files ", .{
-                reporter.summary.fail + reporter.summary.pass,
-                test_files.len,
-            });
+            const total_tests = reporter.summary.fail + reporter.summary.pass + reporter.summary.skip + reporter.summary.todo;
+            Output.prettyError("Ran {d} tests across {d} files. <d>{d} total<r> ", .{ reporter.summary.fail + reporter.summary.pass, test_files.len, total_tests });
```
blackmann, line 612: Added total count

paperclover:
this looks great. i love how nicely it formats in my terminal.

Can you add a feature to this that isn't specifically defined by jest?

```ts
test.todo("test", () => {
  apiThatIsBrokenOrNotImplemented()
})
```
This would have the semantics of "TODO: implement the code we are testing" instead of "TODO: implement this test".

and it will run this test and if it fails, it logs a todo notice like we already have. but if the test passes, it logs a test fail informing that the test has been implemented. so it's kind of like `test.failing` wit different logging.

this specific feature is something I've wanted for a while to use for our bundler tests, as we have many failing and skipped tests, when we are aware of their failure, but if any of them start passing, we want to know so we can add them to the list of expected passes.

blackmann:
image

This is the logs for different cases of `.todo()`.

@paperdave

paperclover:
looks awesome. i love it.

Jarred-Sumner (approved):
LGTM

thank you

PR #2962, opened 2023-05-19
Make `node-fetch` polyfill better

Jarred-Sumner:
`node-fetch` has the following exports:
```js
  AbortError: [class AbortError extends FetchBaseError],
  Blob: [class Blob],
  FetchError: [class FetchError extends FetchBaseError],
  File: [class File extends Blob],
  FormData: [class FormData],
  Headers: [class Headers extends URLSearchParams],
  Request: [class Request extends Body],
  Response: [class Response extends Body],
  blobFrom: [Function: blobFrom],
  blobFromSync: [Function: blobFromSync],
  default: [AsyncFunction: fetch],
  fileFrom: [Function: fileFrom],
  fileFromSync: [Function: fileFromSync],
  isRedirect: [Function: isRedirect]
```

We were only exporting `fetch`. note that this polyfill is also run in browsers which is why it has the `in` checks.

### DIFF
--- a/src/node-fallbacks/node-fetch.js
+++ b/src/node-fallbacks/node-fetch.js
@@ -1,8 +1,88 @@
-var fetchHandler = globalThis.fetch;
+var { Headers, Request, Response, Blob, File = Blob, FormData, fetch: realFetch } = globalThis;
 
 if ("Bun" in globalThis) {
-  fetchHandler = Bun.fetch;
+  realFetch = Bun.fetch;
 }
 
-export default fetchHandler;
-export { fetchHandler as fetch };
+function fetch(url, opts) {
+  return realFetch(url, opts);
+}
+
+class AbortError extends DOMException {
+  constructor(message) {
+    super(message, "AbortError");
+  }
+}
+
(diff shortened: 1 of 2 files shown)


src/node-fallbacks/node-fetch.js
```diff
@@ -1,8 +1,88 @@
...
 if ("Bun" in globalThis) {
-  fetchHandler = Bun.fetch;
+  realFetch = Bun.fetch;
 }
 
-export default fetchHandler;
-export { fetchHandler as fetch };
+function fetch(url, opts) {
+  return realFetch(url, opts);
+}
```
paperclover, line 13: is there a reason why were wrapping `fetch` like this. before we used to just export the function as is, so node-fetch fetch === fetch

Jarred-Sumner: `require("node-fetch")` returns the `default` export which means we need to repeat the ESM exports onto it. We don't want to copy that onto the global `fetch` object, so we wrap it. 

This makes it works similar to before node-fetch went esm-only and broke a bunch of packages

PR #2963, opened 2023-05-19
resolve rope string for multipart template

### LINKED ISSUE #2960: Weird const variable change

0.6.2

Darwin 22.4.0 arm64 arm

Just run this code

```ts
const CONST_VALUE = "CONST_VALUE"; 

function blaBla(a) {
  const { propertyName } = a;
  const condition = `${propertyName}.${CONST_VALUE}AA.WHAT`; // this somehow change the `CONST_VALUE`
  return condition;
}

console.log(CONST_VALUE);
console.log(CONST_VALUE === "CONST_VALUE"); // return false :D
```

Will return result for constant value: `CONST_VALUEAA.WHAT`. Weird. 

On previous version before 0.6 it works as expected.

Also hint: 
if I change ` const { propertyName } = a;` to `propertyName = something` it works
or if I remove `propertyName` variable from condition - it works.

Const value cannot be changed :p

### What do you see instead?

_No response_

### Additional information

_No response_

dylan-conway:
fixes #2960

### DIFF
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -2428,8 +2428,7 @@ pub const E = struct {
                 if (part.value.data == .e_string and part.tail.cooked.isUTF8() and part.value.data.e_string.isUTF8()) {
                     if (parts.items.len == 0) {
                         if (part.value.data.e_string.len() > 0) {
-                            part.value = part.value.clone(allocator) catch unreachable;
-                            head.data.e_string.push(part.value.data.e_string);
+                            head.data.e_string.push(Expr.init(E.String, part.value.data.e_string.clone(allocator) catch unreachable, part.value.loc).data.e_string);
                         }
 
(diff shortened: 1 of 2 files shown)


src/js_ast.zig
```diff
@@ -2447,6 +2447,7 @@ pub const E = struct {
                             }
 
                             if (part.tail.cooked.len() > 0) {
+                                prev_part.tail.cooked.resolveRopeIfNeeded(allocator);
```
Jarred-Sumner: oh this re-allocates the string on every push

lets not do that

instead:

src/js_ast.zig
```diff
@@ -2447,6 +2447,7 @@ pub const E = struct {
                             }
```
Jarred-Sumner, line 2446: can we clone the string up here, just the E.String

PR #2999, opened 2023-05-22
Write out builtins with TypeScript + Minify them

paperclover:


Jarred-Sumner:
cc @constellation incase you find this interesting - this build step lets us use TypeScript & Prettier in JS builtins so its easier to find/see errors in our editor, replacing the `@` builtin syntax with `$` (substituting it out in the final output). It also runs bun's minifier on the code to make it faster to parse at runtime (because there is less source code to iterate through). Also makes the builtins generator step 0.8s for us instead of ~6 seconds

PR #3008, opened 2023-05-23
[node:http] 443 should default to https when no protocol is informed

cirospaciari:


Jarred-Sumner:
there’s supposed to be a default https agent and a default non-https agent and the default https agent should have a port set for 443 

I think that’s a better fix but only marginally so merging anyway

PR #3036, opened 2023-05-23
Bundle and minify `.exports.js` files.

### LINKED ISSUE #2351: Enable `linting` and typechecking via `Typescript` for `Node builtins`
Currently we have no automatic linting or typechecking done on the Node builtins (as noted in #2346) as they currently are not transpiled and use Bun's custom `import.meta.require` CJS imports, which are not automatically picked up by any linters (yet). This is mostly because the builtins are currently embedded as is in the binary. We also have linting disabled for builtins. We really need to change that because linters are nice

We should probably migrate the builtins to `.ts` files, and transpile the builtins and embed them from a `build` directory at build time.

We can start by just migrating imports to ESM syntax and enabling linting to catch obvious syntax errors which would be a great improvement over the status quo.

paperclover:
Closes #2351 

## Hardcoded Modules

All of our `.exports.js` files have been moved to `src/js/`, sorted into three folders:

- `./node` contains all `node:*` modules
- `./bun` contains all `bun:*` modules
- `./thirdparty` contains npm modules we replace like `ws`

The naming convention here is to take the import name and replace `/` with `.`, like `fs.promises.js` and `child_process.js`. We can also use TypeScript now.

Whenever these files are changed, you will need to run `make hardcoded`. CI is unable to run `bun` right now, so you need to make sure you run and commit new changes. After rebuilding, you should be able to run `bun-debug` without rebuilding and changes will be present (excluding `fs/promises` and `bun:ffi`)

For doing different things based on the platform/arch, you can directly use `process.platform`:
```ts
export const suffix =
  process.platform === "darwin"
    ? "dylib"
    : process.platform === "linux"
    ? "so"
    : process.platform === "win32"
    ? "dll"
    : $bundleError("Unsupported platform");
```
Which will compile and minify to include the single string. If `$bundleError` is caught in the build, it fails with the message at bundle time (like zig `@compileError`)

For not implemented modules, we can use `throwNotImplemented` which is defined in `shared.ts`

```ts
export function doSomething() {
  throwNotImplemented("doSomething", 1234);
  // "NotImplementedError: doSomething is not yet implemented in Bun. Track the status & thumbs up the issue: https://github.com/oven-sh/bun/issues/1234"
}
```

For now that helper function and any others get bundled into each module that uses it, but if we have enough of these, later it may make sense to have a secret bun global or internal module that deduplicates this code.

## Builtins

Builtins are moved into `src/js/builtins`, and otherwise function the same as before.


paperclover:
Also, this might be something we don't want to do, but I am not minifying anything in debug mode. This means there are two builds being done (the dev one is gitignore'd), and the idea here is to help with error messages being more useful to us, but that might not be the case. I'm going to go ahead right now and enable minifySyntax for the dev build but not identifiers or whitespace.

Jarred-Sumner:
We probably need to do compressed sourcemaps + source contents so that stack traces in our JS builtins produce useful output. We have a partial implementation of this for —compile but it is incomplete. Also need to see what the bundle size impact is after that, ie if compressed source contents & sourcemaps + minified < input.

Jarred-Sumner:
@paperdave can you rebase and then let's merge it for 0.6.5?

Issue #2484, opened 2023-03-25
issue with process.env being inlined wrong


paperclover:
### What version of Bun is running?

0.5.9 [canary]

### What platform is your computer?

Linux 6.2.2-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

```ts
console.log(process.env.TEST);
console.log(process.env[String("TEST")]);
```

Run like
```bash
TEST="\\\"" bun bug.ts
```

### What is the expected behavior?

both lines to say `\"`

### What do you see instead?

```
"
\"
```

PR #3037, opened 2023-05-23
Load `.env.test`, set NODE_ENV=test in `bun test`, load `.env.{test,production,development}.local`

Jarred-Sumner:
This also reintroduces accidentally removed process.env.NODE_ENV inlining by default in v0.6

```js

test("load .env", () => {
  console.log((process.env.NODE_ENV = "production") === process.env.NODE_ENV);
});
```

That means the following will log `false`.

While `process.env.NODE_ENV` is updated at runtime to `"production"`, at build time, `process.env.NODE_ENV` when not being assigned or deleted is inlined into the string literal `"test"`. This can be changed by explicitly passing the `NODE_ENV` variable.

### DIFF
--- a/src/bun_js.zig
+++ b/src/bun_js.zig
@@ -211,6 +211,8 @@ pub const Run = struct {
             }
         }
 
+        vm.bundler.env.loadTracy();
+
         var callback = OpaqueWrap(Run, Run.start);
         vm.global.vm().holdAPILock(&run, callback);
     }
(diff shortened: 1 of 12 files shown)


paperclover:
I'm adding tests for this on another branch, is it okay if I push those on this branch?

summary of that is most of it works but when you use multiple of these files, they don't override in the same way next.js docs you sent me define it as. mostly with `NODE_ENV=test`

- in `bun test`
  - `.env.development.local` should not be considered
  - `.env.development` should not be considered
  - `.env.local` should not be considered

- doing `NODE_ENV=test bun file.js`
  - `.env.test` isn't loaded, but `process.env.NODE_ENV` reflects `test`

I also added some test cases from existing issues. mainly #2484 was actually fixed in canary but broken now (I think because you turned off the inlining transform?)

Jarred-Sumner:
push to this branch

PR #3040, opened 2023-05-23
Implement `bun test --timeout`

Electroid:
You can now change the default per-test timeout in `bun test` with the `--timeout` flag.

```sh
bun test --timeout 10
```
```
Timeout: test "test" timed out after 10ms
✗ test [18.24ms]
```

However, this does not change the timeout when `test(fn, timeout)` is used, such as:
```ts
import { sleep } from "bun";

test("test", async () => {
  await sleep(10);
}, 5);
```

### DIFF
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -35,8 +35,6 @@ const getAllocator = @import("../base.zig").getAllocator;
 const JSPrivateDataPtr = @import("../base.zig").JSPrivateDataPtr;
 const GetJSPrivateData = @import("../base.zig").GetJSPrivateData;
 
-const default_timeout = std.time.ms_per_min * 5;
-
 const ZigString = JSC.ZigString;
 const JSInternalPromise = JSC.JSInternalPromise;
 const JSPromise = JSC.JSPromise;
@@ -360,8 +358,6 @@ pub const TestRunner = struct {
     only: bool = false,
     last_file: u64 = 0,
 
-    timeout_seconds: f64 = 5.0,
-
     allocator: std.mem.Allocator,
     callback: *Callback = undefined,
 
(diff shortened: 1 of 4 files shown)


src/bun.js/test/jest.zig
```diff
@@ -360,8 +358,6 @@ pub const TestRunner = struct {
     only: bool = false,
     last_file: u64 = 0,
 
-    timeout_seconds: f64 = 5.0,
```
Electroid, line 363: This was unused.

src/bun.js/test/jest.zig
```diff
@@ -35,8 +35,6 @@ const getAllocator = @import("../base.zig").getAllocator;
 const JSPrivateDataPtr = @import("../base.zig").JSPrivateDataPtr;
 const GetJSPrivateData = @import("../base.zig").GetJSPrivateData;
 
-const default_timeout = std.time.ms_per_min * 5;
```
Electroid, line 38: Changed from `std.time.ms_per_min` to `std.time.ms_per_s`. This appeared to be a bug, since the default should be 5 seconds, not 5 minutes.

Jarred-Sumner (approved):
LGTM!

Jarred-Sumner (approved):
LGTM!

Electroid:
```
PASS - test/cli/test/bun-test.test.ts
  bun test v0.6.3 (9382cf05)
  
  bun-test.test.ts:
  ✓ bun test > --timeout > must provide a number timeout [7.81ms]
  ✓ bun test > --timeout > must provide non-negative timeout [4.90ms]
  ✓ bun test > --timeout > timeout can be set to 1ms [19.64ms]
  ✓ bun test > --timeout > timeout should default to 5000ms [5024.59ms]
  
   4 pass
   0 fail
   4 expect() calls
  Ran 4 tests across 1 files. 4 total [5.10s]
```

Issue #2711, opened 2023-04-21
Segfault with server.fetch in testing


simylein:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

```

uh-oh: Not implemented yet!!
bun will crash now 😭😭😭

----- bun meta -----
Bun v0.5.9 (f3ab445c) macOS Silicon 22.4.0
TestCommand: dotenv 
Elapsed: 36ms | User: 1819ms | Sys: 410ms
RSS: 42.81MB | Peak: 42.81MB | Commit: 34.62MB | Faults: 3399
----- bun meta -----

0   0x102286788 WTFGetBacktrace
1   ??? Bun__crashReportDumpStackTrace
2   ??? src.report.fatal
3   ??? Zig::GlobalObject::moduleLoaderFetch(JSC::JSGlobalObject*, JSC::JSModuleLoader*, JSC::JSValue, JSC::JSValue, JSC::JSValue)
4   ??? JSC::moduleLoaderFetch(JSC::JSGlobalObject*, JSC::CallFrame*)
5   ???
6   ???
7   ???
8   ???
9   ???
10  ???
11  ???
12  ??? llint_entry
13  ???
14  ??? llint_entry
15  ??? llint_entry
16  ???
17  ??? llint_entry
18  ??? vmEntryToJavaScript
19  ??? JSC::Interpreter::executeCall(JSC::JSGlobalObject*, JSC::JSObject*, JSC::CallData const&, JSC::JSValue, JSC::ArgList const&)
20  ??? JSC__JSModuleLoader__loadAndEvaluateModule

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord


```

### Additional information

Does server.fetch call dotenv somewhere because I am never calling it myself. I have not yet managed to make a minimal reproduction repository.

simylein:
I managed to debug a little more and noticed in the console:

<img width="806" alt="Screenshot 2023-05-04 at 22 11 32" src="https://user-images.githubusercontent.com/67859356/236318673-df5166eb-29f4-4de7-ae19-4f3fd9c98875.png">

Jarred-Sumner:
is the code using a `with` statement? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/with

simylein:
I searched my whole repository and did not find any `with` statements in the code directly. 
[...]

PR #3051, opened 2023-05-24
[server.fetch] call when using Request object

cirospaciari:
fix: https://github.com/oven-sh/bun/issues/2711

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -4653,10 +4653,9 @@ pub fn NewServer(comptime ssl_enabled_: bool, comptime debug_mode_: bool) type {
             var args = JSC.Node.ArgumentsSlice.from(ctx.bunVM(), arguments);
             defer args.deinit();
 
-            var url: URL = undefined;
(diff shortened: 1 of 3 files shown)


src/bun.js/api/server.zig
```diff
@@ -4704,24 +4703,22 @@ pub fn NewServer(comptime ssl_enabled_: bool, comptime debug_mode_: bool) type {
...
 
-            if (existing_request == null) {
                 existing_request = Request{
                     .url = url.href,
                     .headers = headers,
                     .body = JSC.WebCore.InitRequestBodyValue(body) catch unreachable,
                     .method = method,
                 };
+            } else if (first_arg.as(Request)) |request_| {
+                existing_request = request_.*;
```
Jarred-Sumner, line 4714: i feel like there are more bugs lurking in here due to not cloning the request like with `.clone`

but i think its better than status quo

Jarred-Sumner:
these failing websocket-server.test.ts are new

PR #3055, opened 2023-05-24
Support `NO_COLOR` environment variable

Electroid:
## Changes
- [`NO_COLOR`](https://no-color.org/) now is checked
- [`FORCE_COLOR`](https://github.com/nodejs/node/pull/37477) now matches the same logic as Node.js
- `DISABLE_BUN_ANALYTICS` was removed and replaced with a more standard [`DO_NOT_TRACK`](https://do-not-track.dev/)
- Change the docs to show a table of supported environment variables that Bun checks (note: this is still missing a lot of things, but will need to be updated in another PR)

## Notes
- We cannot properly test when `NO_COLOR` is not defined, since we would need to fake a TTY, and because `FORCE_COLOR` will always override `NO_COLOR`. Those tests are marked as `todo`.

### DIFF
--- a/src/output.zig
+++ b/src/output.zig
@@ -107,7 +107,11 @@ pub const Source = struct {
         // - "1", "true", or "" to indicate 16-color support
         // - "2" to indicate 256-color support
         // - "3" to indicate 16 million-color support
-        return force_color.len == 0 or strings.eqlComptime(force_color, "1") or strings.eqlComptime(force_color, "true") or strings.eqlComptime(force_color, "2") or strings.eqlComptime(force_color, "3");
+        return force_color.len == 0 or
+            strings.eqlComptime(force_color, "1") or
+            strings.eqlComptime(force_color, "true") or
+            strings.eqlComptime(force_color, "2") or
+            strings.eqlComptime(force_color, "3");
     }
 
     pub fn isColorTerminal() bool {
--- a/src/output.zig
+++ b/src/output.zig
@@ -95,29 +95,28 @@ pub const Source = struct {
     }
 
     pub fn isNoColor() bool {
-        return bun.getenvZ("NO_COLOR") != null;
+        const no_color = bun.getenvZ("NO_COLOR") orelse return false;
+        // https://no-color.org/
+        // "when present and not an empty string (regardless of its value)"
+        return no_color.len != 0;
     }
 
-    pub fn isForceColor() ?bool {
-        const force_color_str = bun.getenvZ("FORCE_COLOR") orelse return null;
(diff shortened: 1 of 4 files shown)


src/output.zig
```diff
@@ -128,27 +131,26 @@ pub const Source = struct {
...
-                    enable_ansi_colors_stderr = is_color_terminal orelse isColorTerminal();
-                } else if (isForceColor()) |val| {
-                    enable_ansi_colors_stderr = val;
-                } else {
-                    enable_ansi_colors_stderr = false;
                 }
 
-                enable_ansi_colors = enable_ansi_colors_stderr or enable_ansi_colors_stdout;
+                enable_ansi_colors_stdout = enable_color orelse is_stdout_tty;
+                enable_ansi_colors_stderr = enable_color orelse is_stderr_tty;
```
Jarred-Sumner, line 152: I don't think this is correct

TTYs do not all support colors

Electroid: If the tty does not support color, `enable_color` is set to `false`.

Jarred-Sumner: TTYs declare their support for color, the default is false

Electroid: yes, that's what this check does?

https://github.com/oven-sh/bun/pull/3055/files#diff-f2619d0e89ac22f8e6016642173178fea360dd0ea1bddf72f1b9ea078e1db9c9R137-R139

Jarred-Sumner: if `enable_color` is `null`, then it will be enabled if `is_stderr_tty` is true

`orelse` in zig is `or` but only for `null`, like `??` in JS

Electroid: > if enable_color is null, then it will be enabled if is_stderr_tty is true

That's intended

```zig
var enable_color: ?bool = null;
if (isForceColor()) {
  enable_color = true;
} else if (isNoColor() or !isColorTerminal()) {
  enable_color = false;
}
const is_stdout_tty = _source.stream.isTty();
```

When a terminal does not support colors...
*  `enable_color` starts off as `null`
* `isForceColor()` evaluated to `false`, continue
* `isNoColor()` evaluated to `false`, continue
* `isColorTerminal()` evaluated to `false`, but the statement is negated, so it evaluated to `true`, step
  * `enable_color = false;` is evaluated

In this case, `enable_color` is a tri-state, it is only `null` when there is no `NO_COLOR` or `FORCE_COLOR` variable _and_ the terminal accepts color. Let me know if that makes sense.

Would `force_color` be a better variable name? I get the sense maybe it's confusing.

PR #3062, opened 2023-05-25
docs: add a bit more info on Elysia

capaj:


Jarred-Sumner:
Thank you

capaj:
ooops, I will open a fix for this. Obviously markdown rendering is not the same in docs as here on github:

image

PR #3071, opened 2023-05-25
Small improvements to `bun test` 

Electroid:
## Changelog

c09fcb2380ba5d4958341daceb7e2e6a44e870da The skip icon is now `»` instead of `-` (it's nicer).

Screen_Shot_2023-05-25_at_11 21 05_AM

---

1d7b76b2fb2a1c683eb0ae93aa66af37beddea85 The title of each test file is now its path instead of its filename.

| Before | After |
| - | - |
| Screen Shot 2023-05-25 at 2 23 37 PM | Screen Shot 2023-05-25 at 2 20 16 PM |

---

4926e6a8dc617c04a751f9e2bc7adafc6614614e When `bun test` is run in Github Actions, logs are grouped by test file using [workflow commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines). Detection is done by checking that the `GITHUB_ACTIONS` environment variable is set to `true`.

https://github.com/oven-sh/bun/assets/3238291/b0f84105-44e2-46ec-99bf-4f615f478f95

---

d1c658e0be8e974b434edb403f6e9ab92ab3d251 When an error occurs in a test, an annotation is emitted.

https://github.com/oven-sh/bun/assets/3238291/6ef956c9-2c82-4b28-860b-cce305799f35


src/output.zig
```diff
@@ -121,12 +121,20 @@ pub const Source = struct {
...
+        return false;
+    }
+
     pub fn set(_source: *Source) void {
         source = _source.*;
 
         source_set = true;
         if (!stdout_stream_set) {
             stdout_stream_set = true;
+            is_github_action = isGithubAction();
```
Jarred-Sumner: this is annoying but can we set this only in test_command.zig

Jarred-Sumner: otherwise LGTM

Jarred-Sumner:
This looks good though there are new test failures

Jarred-Sumner:
test.runIf

interesting

Jarred-Sumner:
maybe `test.if`?

simylein:
If you are already on it can we implement that test failures log last because the constant scrolling up the logs to find details about failing tests is rather annoying. I remember having seen an issue about this but cannot find it at the moment.

PR #3074, opened 2023-05-25
[eventsource] SSE Client

cirospaciari:
Supported features:
- [x] retry, id, data, event fields
- [x] chunked, content-length or no size hint
- [x] auto retry
- [x] HTTP / 1.1
- [x] TLS support

Future features that we need to add on the future:
- [ ] decompression with gzip/deflate
- [ ] HTTP / 2.0


cirospaciari:
~~`test/js/bun/eventsource/eventsource.test.js` on `bun-darwin-x64-baseline`  should be fixed by now, waiting the tests to run to see~~ fixed

Jarred-Sumner:
Probably needs a rebase but imo let's ship it

cirospaciari:
> Probably needs a rebase but imo let's ship it

Updated, seems like js/out is auto-generated with hardcoded + make builtins both are updated

Jarred-Sumner:
I think the dev stuff was not meant to be committed and should be git ignored, will fix

.gitignore
```diff
@@ -121,3 +121,5 @@ cold-jsc-start
 cold-jsc-start.d
 
 /test.ts
+
+src/js/out/modules_dev
```
Electroid, line 125: @paperdave should we do this?

PR #3081, opened 2023-05-26
Pretty formatter for `Headers` & `URLSearchParams`

Jarred-Sumner:


test/js/web/html/FormData.test.ts
```diff
@@ -335,9 +335,29 @@ describe("FormData", () => {
     formData.append("foo", new Blob(["bar"]));
     formData.append("bar", "baz");
     formData.append("boop", Bun.file("missing"));
+    console.log(new Headers());
```
Electroid: Extra log statement

PR #3097, opened 2023-05-28
Implement process.env.npm_lifecycle_event

### LINKED ISSUE #2972: implement process.env.npm_lifecycle_event

0.6.2

Darwin 22.5.0 arm64 arm

run the following code with `bun start:dev`

// package.json
```
{
  "scripts": {
    "start:dev": "bun main.ts"
  }
}
```

// main.ts
```ts
console.log(process.env.npm_lifecycle_event);
```

it should log `start:dev`

### What do you see instead?

it logs `undefined`

### Additional information

This is probably not used very often but we should provide a similar (or the same) way to access the currently running lifecycle script

v3nntrnx:
This PR implements process.env.npm_lifecycle_event.

npm_lifecycle_event is an environment variable provided by the the script runner runtime during the execution of scripts. When running scripts defined in a project's package.json file using the npm run command, npm_lifecycle_event stores the name of the currently executing npm lifecycle event.

For example:
// package.json
```
{
  "scripts": {
    "start:dev": "bun a.js"
  }
}
```
// a.js
`console.log(process.env.npm_lifecycle_event);`

When running a.js using `bun run start:dev` it should output "start:dev", in current version of bun it just returns undefined.

Fixes #2972

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -960,6 +960,7 @@ pub const RunCommand = struct {
         var ORIGINAL_PATH: string = "";
         var this_bundler: bundler.Bundler = undefined;
         var root_dir_info = try configureEnvForRun(ctx, &this_bundler, null, &ORIGINAL_PATH, log_errors, force_using_bun);
+        this_bundler.env.map.putDefault("npm_lifecycle_event", script_name_to_search) catch unreachable;
         if (root_dir_info.enclosing_package_json) |package_json| {
             if (package_json.scripts) |scripts| {
                 switch (script_name_to_search.len) {
--- a/test/cli/run/env.test.ts
+++ b/test/cli/run/env.test.ts
@@ -1,47 +1,5 @@
 import { describe, expect, test } from "bun:test";
-import os from "os";
-import fs from "fs";
-import path from "path";
-import { bunEnv, bunExe } from "harness";
-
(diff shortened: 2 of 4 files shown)


v3nntrnx:
Just implemented a test, works on my machine.

Jarred-Sumner:
Thanks for this 

Looks good except the utility function should be moved into a separate file so that the other files tests are not run

v3nntrnx:
> Thanks for this 
> 
> Looks good except the utility function should be moved into a separate file so that the other files tests are not run 

Sure, I'll quickly do it now. In this PR right?

Jarred-Sumner:
We will squash & merge so the specific commit doesn’t matter, but yes in this PR

Note that the test runner will report a bunch of unrelated failures. We need to mark these tests as TODO

v3nntrnx:
Done, I moved all script runs and stuff to the harness file.

Now it runs only the tests it needs.

Jarred-Sumner (approved):
LGTM

Let’s let the tests run to make sure the specific files changes pass in CI. There will be other test failures but don’t worry about those.

Jarred-Sumner:
Thank you

v3nntrnx:
> Thank you

Thank you too, my first serious PR lol ❤️

PR #3101, opened 2023-05-28
[node:net] fix createConnection options passing

### LINKED ISSUE #3094: `net.createConnection(path)` param format not supported

0.6.4

macOS

Not working

```
const socket = net.createConnection(sockname);
```

Working

```
const socket = net.createConnection({path: sockname});
```

https://nodejs.org/api/net.html#netcreateconnectionpath-connectlistener

---

As used here: https://github.com/facebook/watchman/blob/main/watchman/node/index.js#L114

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

cirospaciari:
fix #3094

### DIFF
--- a/src/bun.js/net.exports.js
+++ b/src/bun.js/net.exports.js
@@ -532,7 +532,7 @@ const Socket = (function (InternalSocket) {
 
 function createConnection(port, host, connectListener) {
   if (typeof port === "object") {
-    // port is option pass Socket options and let connection handle connect options
+    // port is option pass Socket options and let connect handle connection options
     return new Socket(port).connect(port, host, connectListener);
   }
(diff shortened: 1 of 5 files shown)


src/bun.js/net.exports.js
```diff
@@ -531,18 +531,12 @@ const Socket = (function (InternalSocket) {
 );
 
 function createConnection(port, host, connectListener) {
-  if (typeof host == "function") {
-    connectListener = host;
-    host = undefined;
+  if (typeof port === "object") {
```
Jarred-Sumner, line 534: this should probably respect arguments.length but it's better than status quo

Jarred-Sumner:
Thank you

also for fixing prettier

PR #3118, opened 2023-05-30
feat(fs/copyFile): implement fallback for EXDEV

### LINKED ISSUE #1675: copyFile: Cross-device link error
EDIT: This also breaks create-vite, which seems to be related to the /tmp folder mentioned below.

0.4.0

Linux 6.1.0-gentoo x86_64 AMD Ryzen 5 5600X 6-Core Processor

Attempting to use `fs.promises.copyFile` across btrfs fileystems.

The file to be succesfully copied into the target folder.

### What do you see instead?

EXDEV: Cross-device link
   path: "/home/theo/.bun/bin/bun"
syscall: "copy_file_range"
  errno: -18

### Additional information

See https://github.com/oven-sh/bun/blob/main/src/bun.js/node/node_fs.zig#L2621.

The following code works fine.
```
export const copyFile = async (input: string, output: string) => {
  await Bun.write(output, await Bun.file(input).arrayBuffer());
};
```

If you were to remove the arrayBuffer call it has the same issue as earlier.

v3nntrnx:
Fixes #1675 

I also have to mention that it already handles the truncate stuff and the permissions.


src/bun.js/node/node_fs.zig
```diff
@@ -2633,14 +2633,30 @@ pub const NodeFS = struct {
...
+
+                                        var slice = buffer[0..bytes_read];
+
+                                        written = switch (Syscall.write(dest_fd, slice)) {
+                                            .result => |result| result,
+                                            .err => |_err| return Maybe(Return.CopyFile){ .err = _err.withPath(src) },
+                                        };
+                                    },
+                                    else => return err,
+                                }
```
xhyrom: Can we reduce this duplicate code?

```zig
var written = linux.copy_file_range(src_fd, &off_in_copy, dest_fd, &off_out_copy, std.mem.page_size, 0);

if (ret.errnoSysP(written, .copy_file_range, dest)) |err| {
    switch (err.getErrno()) {
        .XDEV => {
            var buffer: [4096]u8 = undefined;

            var bytes_read = switch (Syscall.read(src_fd, &buffer)) {
                .result => |result| result,
                .err => |_err| return Maybe(Return.CopyFile){ .err = _err.withPath(src) },
            };

            if (bytes_read == 0) break;

            var slice = buffer[0..bytes_read];

            written = switch (Syscall.write(dest_fd, slice)) {
                .result => |result| result,
                .err => |_err| return Maybe(Return.CopyFile){ .err = _err.withPath(src) },
            };
        },
        else => return err,
    }
}
// wrote zero bytes means EOF
if (written == 0) break;
wrote +|= written;
```

https://github.com/TiranexDev/bun/blob/92474f3302454fd3a280a7b3c0a1695ffc33a251/src/bun.js/node/node_fs.zig#L2633-L2698

xhyrom (changes requested):

src/bun.js/node/node_fs.zig
v3nntrnx: Sure, I just want to see if Jaared will like the code first.

xhyrom: cc @Jarred-Sumner

v3nntrnx: I'll see what I can do meanwhile

v3nntrnx: Just pushed it

v3nntrnx:
Tried my best, hope it's better now

v3nntrnx:
to confirm it actually works i asked theo paris to test it (the initial author of the issue)
link to exe <https://ufile.io/k1bq9cwg> (Linux, x86_64) (I downloaded and tested it, should work)

cc @theoparis

v3nntrnx:
VM tests: (also tried on a large image & empty text file)
image
image

PR #2628, opened 2023-04-11
Add `bun docs`

rmorey:
Adds `bun docs` which opens https://bun.sh/docs. 

Copied entirely from the implementation of `bun discord`

PR #3133, opened 2023-05-31
refactor(cli)!: remove discord command

xhyrom:
Like with https://github.com/oven-sh/bun/pull/2628, we should consider removing `discord` command as its unnecessary.
If you have bun installed, you've probably been to the site and had the option to connect to the discord server directly on the site.
image

In my opinion, we shouldn't add such subcommands to the CLI that do "useless" things. Better to do some other command that return also web, discord and all other medias. This is just my opinion, I would be happy if more of us would vote on this change.

At this stage of bun this is not a big change and it is a breaking change but I think it will affect minimum people/projects or even nobody. We need bun to get to some stage before 1.0 where making breaking changes will then be possible, but not as easily as we can (can't) afford to do now.


v3nntrnx (commented):
LGTM, I agree too. "bun discord" command is really useless, I never used it myself or saw anyone else use it.

paperclover:
i wonder if putting the link in the help text as well as the link to the docs might be neat

xhyrom:
> i wonder if putting the link in the help text as well as the link to the docs might be neat

Good point

Jarred-Sumner:
I think it’s worth keeping because it helps people find the Discord server. It doesn’t cost much to keep it around either. On Linux it doesn’t work as well but on macOS it opens the discord in your browser

xhyrom:
> I think it’s worth keeping because it helps people find the Discord server. It doesn’t cost much to keep it around either. On Linux it doesn’t work as well but on macOS it opens the discord in your browser

I think what says paperdave is the best solution - put website, docs, discord into help instead of custom command

v3nntrnx:
> > I think it’s worth keeping because it helps people find the Discord server. It doesn’t cost much to keep it around either. On Linux it doesn’t work as well but on macOS it opens the discord in your browser
> 
> I think what says paperdave is the best solution - put website, docs, discord into help instead of custom command

Not a bad way too

Electroid:
For now, we'll keep the command. It is nice for discoverability as we grow the community.

xhyrom:
> For now, we'll keep the command. It is nice for discoverability as we grow the community.

Well, I think it's going to be even harder to remove then. The links in the helper would be quite sufficient, I don't even think that's necessary.  Also, it doesn't even work correctly on Linux. Of course, the decision is up to the Core team.

Issue #3126, opened 2023-05-31
`path.parse('/foo/bar').root` returns `""` instead of `"/"`


vjpr:
### What version of Bun is running?

0.6.5

### What platform is your computer?

Darwin 22.2.0 arm64 arm

### What steps can reproduce the bug?

```
const {root} = path.parse('/tmp')
console.log(root)
```

```
bun index.js
node index.js
```

### What is the expected behavior?

```
"/"
```

### What do you see instead?

```
""
```

### Additional information

This breaks Sindre's `find-up` package.

vjpr:
Source:

https://github.com/oven-sh/bun/blob/557aac6a34bcd06bba98060f67bb7295d6c462d9/src/bun.js/node/types.zig#L1784

Issue #2338, opened 2023-03-07
`path.parse` includes file extension in the `name` field


errilaz:
### What version of Bun is running?

0.5.7

### What platform is your computer?

Linux 5.19.0-35-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
const path = require("path")
console.log(path.parse("foo.txt").name)
```

### What is the expected behavior?

`foo`

### What do you see instead?

`foo.txt`

Ruk33:
Hello, I'd like to work on this issue if that's ok. Thanks!

PR #3134, opened 2023-05-31
fix(path) fix parse behavior

cirospaciari:
Fix: https://github.com/oven-sh/bun/issues/3126
Fix: https://github.com/oven-sh/bun/issues/2338


Jarred-Sumner:
Thank you

Issue #600, opened 2022-07-11
Support SvelteKit


niiwade:
### What is the problem this feature will solve?

We need to have Svelte to be included in the project creation step to make it easy to using Bun to create Sveltekit apps. 

### What is the feature you are proposing to solve the problem?

We need `bun create svelte ./app` to make it easy to create a new svelte project

FinnRG:
@niiwade Would you like to create a PR for this?

niiwade:
yes sure

swyxio:
note that there is a community adapter here you can maybe crib from https://github.com/gornostay25/svelte-adapter-bun

niiwade:
> note that there is a community adapter here you can maybe crib from https://github.com/gornostay25/svelte-adapter-bun

Thank you @sw-yx , am checking it out

ryanrussell:
Any good news on SvelteKit + bun runtime?

This would be a great framework to get first class support for in bun. 

![image](https://user-images.githubusercontent.com/523300/181996608-a4476a06-f01a-4227-9acf-d95edbe21aef.png)

paperclover:
Status on SvelteKit:

- It depends on `worker_threads`, though not much of it. Just `parentPort` and `new Worker`, and passing `env` data.
- Since we do not export anything in `worker_threads`, it causes vite to crash loading the config file.

kansson:
SvelteKit is now working with `bun upgrade --canary` until `v0.7.2` is released 🚀
EDIT: There is still some work to be done see https://github.com/oven-sh/bun/issues/600#issuecomment-1664814473.

gitblit:
This was a nice surprise. Thanks for commenting @hanssonduck . Looks like there are some minor console tty issues on Vite and some HMR issues too, although fair to say those are not SvelteKit directly. On a starter app built from the Svelte bootstrapping command, I have success. On a more complex app, I have hanging requests. Not sure why yet. But overall getting closer!!

<img width="604" alt="image" src="https://github.com/oven-sh/bun/assets/868217/c9f3c84e-63bc-4cac-b598-03bd9091bc83">

paperclover:
Worker threads got it so importing sveltekit stopped throwing, which gets us past the vite config load step.

But

We're not fully there yet, unfortunately. I've spotted some reliability issues with actions and `.formData()`, where the promise never resolved. Though small reproductions dont actually recreate the issue. Something subtle with our streams implementation.

[...]

Issue #250, opened 2022-07-06
Vite


brillout:
Vite is on [Bun's roadmap](https://github.com/Jarred-Sumner/bun/issues/159), which is awesome.

I'm opening this issue to track progress on Vite support.

A lot of Vite folks would gladly replace Node.js with Bun :-).

### Edit

Shameless plug: myself and @cyco130 are the authors of:
 - [HatTip](https://github.com/hattipjs/hattip) - a modern Express.js alternative: Fetch API / WinterCG standards + works anywhere (Node.js Edge, Deno, and Bun &mdash; of course 🙂).
 - [vite-plugin-ssr](https://vite-plugin-ssr.com/) - Like Next.js / Nuxt but as do-one-thing-do-it-well Vite plugin. (The production runtime works with Bun.)

bluwy:
Tried Bun with Vite this weekend at https://github.com/bluwy/bun-vite-ts-test. Noted down some issues so far in the readme, with a patch to fix any fixable things so far. Currently stucked but I'll be watching Bun's releases and keep it up to date.

brillout:
It's happening: https://twitter.com/jarredsumner/status/1551643885998247938.

Jarred-Sumner:
As of Bun v0.1.11 (released today), here is the current status of the list of items mentioned in https://github.com/bluwy/bun-vite-ts-test 

> - [ ] http, http2, https not implemented.

The server parts of `http` are partially implemented - enough for `express` to work. `http2` and `https` aren't implemented at all yet

 - [x] `child_process` not implemented. Blocks esbuild.

> - [x] Regex negative lookbehind not supported in JSC. PITA to refactor.

This one is tricky. It's probably the biggest JSC-specific compatibility issue I've seen. We may need to have some WASM or ffi thing inserted by the transpiler to workaround it until https://bugs.webkit.org/show_bug.cgi?id=174931 is resolved, or attempt to implement it ourselves

- [x] When `const require = createRequire()` exists, bun treats the file as CJS, transpiling imports to `require()`, but because `const require` is declared late, it gives ` ReferenceError: Cannot access uninitialized variable`. Can be workaround by avoiding `require` and rename as `_require`. https://github.com/oven-sh/bun/issues/453

> - [ ] Dynamic `import()` doesn't execute file, as if a no-op. 

[...]

PR #3140, opened 2023-05-31
random fixes that help vite/sveltekit

paperclover:
**Status** — hello worlds work, but real applications including the sveltekit demo do not fully work

This PR makes progress towards #250 and #600, fixing small random bugs in places to help vite and sveltekit run within bun's runtime:

- fs: `existsSync(null)` and other invalid paths must not throw
- resolver: `import "file:///..."`
- ~~`Path.isAbsoluteString` supports utf16 strings~~
- unrelated, make `Bun.readableStreamToArrayBuffer` sync if possible.

this also (temporarily) contains changes from #2913 which I think help out vite

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -1739,25 +1739,11 @@ pub const Arguments = struct {
         }
     };
     pub const Exists = struct {
-        path: PathLike,
+        path: ?PathLike,
 
         pub fn fromJS(ctx: JSC.C.JSContextRef, arguments: *ArgumentsSlice, exception: JSC.C.ExceptionRef) ?Exists {
-            const path = PathLike.fromJS(ctx, arguments, exception) orelse {
-                if (exception.* == null) {
-                    JSC.throwInvalidArguments(
-                        "path must be a string or buffer",
-                        .{},
-                        ctx,
-                        exception,
-                    );
-                }
-                return null;
-            };
-
-            if (exception.* != null) return null;
-
             return Exists{
-                .path = path,
+                .path = PathLike.fromJS(ctx, arguments, exception),
             };
         }
     };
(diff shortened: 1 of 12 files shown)


src/resolver/resolver.zig
```diff
@@ -1129,6 +1129,43 @@ pub const Resolver = struct {
             return .{ .not_found = {} };
         }
 
+        if (strings.startsWith(import_path, "file:///")) {
```
Jarred-Sumner: `startsWith` returns true for empty strings and the right-hand side is comptime-known which means we can make it faster through more specific codegen


```suggestion
        if (strings.hasPrefixComptime(import_path, "file:///")) {
```

src/resolver/resolver.zig
```diff
@@ -1129,6 +1129,43 @@ pub const Resolver = struct {
...
+                return .{
+                    .success = Result{
+                        .path_pair = .{ .primary = Path.init(import_path) },
+                        .is_external = true,
+                    },
+                };
+            }
+
+            // TODO: this should only load the file itself
+            if (r.loadAsFileOrDirectory(path, kind)) |entry| {
```
Jarred-Sumner: isn't there a `.loadAsFile`?

Jarred-Sumner: The code path here should be similar to the one that applies to absolute paths

src/bun.js/node/types.zig
```diff
@@ -1672,7 +1672,7 @@ pub const Path = struct {
         }
     }
     fn isAbsoluteString(path: JSC.ZigString, windows: bool) bool {
-        if (!windows) return path.len > 0 and path.slice()[0] == '/';
+        if (!windows) return path.len > 0 and path.ptr[0] == '/';
```
Jarred-Sumner: this is a segfault

we should never read from `ptr` in JSC.ZigString directly

i'll make it more clear in the code that we shoulnd't do that

Jarred-Sumner: both are actually segfaults, i'll fix

src/bun.js/node/node_fs.zig
```diff
@@ -1739,25 +1739,11 @@ pub const Arguments = struct {
...
-                    );
-                }
-                return null;
-            };
-
-            if (exception.* != null) return null;
-
             return Exists{
-                .path = path,
+                .path = PathLike.fromJS(ctx, arguments, exception),
```
Jarred-Sumner, line 1746: Why is it okay for this to be optional?

Jarred-Sumner: ah

it should just immediately return false

paperclover: yeah, this fixes something within sveltekit's config stuff. I forget exactly what it was.

src/resolver/resolver.zig
```diff
@@ -1129,6 +1129,41 @@ pub const Resolver = struct {
             return .{ .not_found = {} };
         }
 
+        if (strings.hasPrefixComptime(import_path, "file:///")) {
```
Jarred-Sumner, line 1132: This doesn't handle space characters or anything percent-encoded 

but it is better than status quo

src/bun.js/node/node_fs.zig
Jarred-Sumner, line 1746: I don't like this fix, it should be handled earlier in the function instead of these extra null checks. 

But it does fix the issue. Please add a TODO note to address this.

PR #3141, opened 2023-05-31
fix base/filename internally

cirospaciari:


Jarred-Sumner:
lets merge once some tests run

PR #3155, opened 2023-06-01
refactor: use main repository for upgrading

xhyrom:
I kept commands in Makefile because few more releases should be also released to the old repository but we should migrate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ SHELL := $(shell which bash) # Use bash syntax to be consistent
 
 OS_NAME := $(shell uname -s | tr '[:upper:]' '[:lower:]')
 ARCH_NAME_RAW := $(shell uname -m)
+# TODO: remove this in upcomming next few versions
 BUN_AUTO_UPDATER_REPO = Jarred-Sumner/bun-releases-for-updater
 
 CMAKE_CXX_COMPILER_LAUNCHER_FLAG :=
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -2,6 +2,7 @@ SHELL := $(shell which bash) # Use bash syntax to be consistent
 
 OS_NAME := $(shell uname -s | tr '[:upper:]' '[:lower:]')
 ARCH_NAME_RAW := $(shell uname -m)
+# TODO: remove this in upcomming next versions
```
Electroid: We won't remove this for a very, very long time; if ever. Instead, can we add a comment that mentions that we publish to that repo for backwards compatibility?

docs/cli/bun-upgrade.md
```diff
@@ -2,7 +2,7 @@ To upgrade Bun, run `bun upgrade`.
 
 It automatically downloads the latest version of Bun and overwrites the currently-running version.
 
-This works by checking the latest version of Bun in [bun-releases-for-updater](https://github.com/Jarred-Sumner/bun-releases-for-updater/releases) and unzipping it using the system-provided `unzip` library (so that Gatekeeper works on macOS)
+This works by checking the latest version of Bun in [bun-releases-for-updater](https://github.com/oven-sh/bun/releases) and unzipping it using the system-provided `unzip` library (so that Gatekeeper works on macOS)
```
Electroid:
```suggestion
This works by checking the latest release of Bun from GitHub [releases](https://github.com/oven-sh/bun/releases) and unzipping it using the system-provided `unzip` library (so that Gatekeeper works on macOS)
```

Makefile
xhyrom: Sure

xhyrom:
@Electroid done

Electroid:
We're actually going to keep the status-quo for now. Reasoning: this allows us to understand how many people upgrade vs. direct download. This may seem like a small thing, but it is an important metric for us to track as we grow Bun.

PR #3156, opened 2023-06-01
ci: dont run builds in forks

xhyrom:


Electroid:
Makes sense not to run deploys, for the website, but why not builds?

xhyrom:
> Makes sense not to run deploys, for the website, but why not builds?

Because you're using your custom runners that fork doesn't have so it's stucked.

Example: https://github.com/xhyrom-forks/bun/actions/workflows/bun-linux-build.yml

Electroid (approved):
Thanks

PR #3159, opened 2023-06-01
record cjs export name if it is assigned

### LINKED ISSUE #3152: Unexpected identifier 'root'

0.6.6

Darwin 22.5.0 x86_64 i386

```js
import { uniq } from "lodash-es";

console.log(uniq([1, 1, 2, 3]));
```

on our example, we need to see in the console `[ 1, 2, 3 ]`

### What do you see instead?

```console
SyntaxError: Unexpected identifier 'root'. import call expects one or two arguments.
```

### Additional information

* on bun 0.6.4 this works correctly
* this issue occurred on version 0.6.5 (with the release of the new bundler)
* note: the `lodash` version for example is working on the current version of bun, but not the `lodash-es` version

dylan-conway:
fixes #3152 

line 5 in `isBuffer.js` in lodash-es should not record `nodeType` as a cjs export because nothing is assigned to the property:
```js
var freeExports = typeof exports == 'object' && exports && !exports.nodeType && exports;
```

### DIFF
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -17656,7 +17656,7 @@ fn NewParser_(
                                     },
                                     name_loc,
                                 );
-                            } else if (p.options.features.commonjs_at_runtime) {
+                            } else if (p.options.features.commonjs_at_runtime and identifier_opts.assign_target != .none) {
                                 // Record this CommonJS export name for use later.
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
nice

Jarred-Sumner:
lets merge once tests run

PR #3162, opened 2023-06-01
[napi] fix thread safe function callback

cirospaciari:


Jarred-Sumner:
we should add a test for this later

PR #3166, opened 2023-06-02
Fix github annotation when using inspect()

Electroid:
Fixes a bug where `inspect()` would emit a Github annotation.


Jarred-Sumner:
LGTM once tests run

PR #3176, opened 2023-06-02
[napi] some fixes for prisma

cirospaciari:


src/napi/js_native_api.h
```diff
@@ -238,6 +238,8 @@ NAPI_EXTERN napi_status napi_unwrap(napi_env env, napi_value js_object,
                                     void **result);
 NAPI_EXTERN napi_status napi_remove_wrap(napi_env env, napi_value js_object,
                                          void **result);
+NAPI_EXTERN napi_status napi_create_object(napi_env env,
```
Jarred-Sumner, line 241: Did they add this symbol recently?

Jarred-Sumner: It most likely already is defined.

src/napi/napi.zig
```diff
@@ -653,6 +648,7 @@ pub extern fn napi_define_class(
 pub extern fn napi_wrap(env: napi_env, js_object: napi_value, native_object: ?*anyopaque, finalize_cb: napi_finalize, finalize_hint: ?*anyopaque, result: [*]*Ref) napi_status;
 pub extern fn napi_unwrap(env: napi_env, js_object: napi_value, result: [*]*anyopaque) napi_status;
 pub extern fn napi_remove_wrap(env: napi_env, js_object: napi_value, result: [*]*anyopaque) napi_status;
+pub extern fn napi_create_object(env: napi_env, result: *napi_value) napi_status;
```
Jarred-Sumner, line 652: Will need to add this under doNotOptimizeAway at the bottom of this file and to symbols.txt so that macOS sees it

src/napi/js_native_api.h
cirospaciari: just moved from Zig to C++ so we can use the `NapiPrototype`, this way `napi_create_object` will work with `napi_wrap/unwrap`

src/napi/napi.zig
cirospaciari: is already added

Jarred-Sumner:
we need to find a way to improve our test coverage for NAPI 

but lets merge this

PR #3181, opened 2023-06-03
fix readableStreamToArrayBuffer

paperclover:


test/js/bun/util/readablestreamtoarraybuffer.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  let counter = 0;
+  // @ts-ignore
+  Promise.prototype.then = (...args) => {
+    counter++;
+    return _then.apply(this, args);
+  };
+  try {
+    const result = await Bun.readableStreamToArrayBuffer(
+      new ReadableStream({
+        async start(controller) {
```
Jarred-Sumner, line 15: nitpick: `start` is executed immediately ,synchronously so this test does not cover promises

it would need to go under pull

Jarred-Sumner (approved):

PR #3182, opened 2023-06-03
fixes with hardcoded modules

paperclover:
this separates detect-libc into two files, one per platform and a similar thing for ffi

also removes from GitHub the `modules_dev` folder which isn't something I ever intended to commit.

`make dev` will run `make hardcoded` which is partially needed because the dev version of these files is not committed and it feels better to run it on every build in case you forget to run the other command.

this does make it so that you need to have bun installed to build bun, which is fine for most cases but would be an issue if trying to build on other platforms

this also todo's out a failing test


Jarred-Sumner (approved):

Jarred-Sumner:
test/js/third_party/body-parser/express-body-parser-test.test.ts

that is sus

PR #898, opened 2022-07-28
chore: better build workflow

xhyrom:
> **Warning**
> Haven't tested, can't test it

Every `matrix.settings` element will run as new job

Added `bun-build.yml` that builds macos + linux

Not sure if it works

PR #3185, opened 2023-06-03
ci: better structure, build workflow

xhyrom:
Have 100 of different workflows for macos and Linux is bad approach. We can make it in two (Mac and Linux) and readable.

First PR: https://github.com/oven-sh/bun/pull/898

This is not done, don't close or merge it.

> Warning
> I can't ly test this, because custom runners but I'll try to do what I can

### DIFF
--- a/.github/workflows/bun-mac-x64-baseline.yml
+++ b/.github/workflows/bun-mac-x64-baseline.yml
@@ -11,7 +11,7 @@ env:
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


Electroid:
Hey, thanks for trying to work on this! I'd recommend forking bun, commenting out the custom runners, then trying your changes. Feel free to open a PR when the changes are ready.

PR #3186, opened 2023-06-03
Add test case for bundler

paperclover:
Test case from #2918 isolated. This was the underlying issue that caused express to fail


Jarred-Sumner:
Thank you

PR #3188, opened 2023-06-03
make hardcoded -> make esm + document needing bun to build bun

paperclover:
renames the hardcoded modules script to `make esm`, and then documents the fact you need bun to build bun as we do not explicitly state this


Jarred-Sumner:
thank you

PR #3196, opened 2023-06-03
child_process fork function implementation

haouarihk:
This merge request adds an implementation of the fork function, based on the code found in the Node.js repository. It also includes the necessary missing functions. Please note that this implementation has not been tested yet.


haouarihk:
you're welcome to decline it, if it doesn't work.

Jarred-Sumner:
Thank you for the PR, but we are not going to accept this since it has not been tested and there are no tests. If you’d like to contribute to Bun, please add automated tests. After you add automated tests (ideally ported from Node) feel free to submit this PR again 

I’m also not sure it should work this way - fork() is a system call we would ideally be able to leverage to reduce start time.

haouarihk:
Thanks for the note

PR #3197, opened 2023-06-03
[tests] prisma tests

cirospaciari:
using  `PRISMA_POSTGRES_DATABASE_URL` and `PRISMA_MONGODB_DATABASE_URL` for configuration if not configured test will auto skip with a warning


test/package.json
```diff
@@ -19,7 +19,9 @@
     "undici": "^5.20.0",
     "socket.io": "^4.6.1",
     "socket.io-client": "^4.6.1",
-    "supertest": "^6.1.6"
+    "supertest": "^6.1.6",
```
Electroid, line 22: We should probably change `js/third_party` to be a mono repo, but we don't need to make that change in this PR.

Jarred-Sumner:
Thank you

PR #3198, opened 2023-06-03
Add bun-can-run to ecosystem nav

emastho:
https://github.com/emastho/bun-can-run

I'm actively working on keeping the list fresh, following discord and issues, I'd love to get some eyes on this, so people don't even need to ask if something is working or not


xhyrom (approved):
LGTM nut I know something similar was already here (but there were no docs) and it was argued that such things are cool but weird to put officially since bun has (and will) support everything.

emastho:
> 

Well, once Bun supports most of the things we will be able to just remove it from the docs, but right now this is still a thing I needed for example when I first started using Bun. I think its more optimal to do this, than for every single new Bun user to redigg their own way to our ecosystem. 😄

PR #3204, opened 2023-06-04
Implement `Bun.password` and `Bun.passwordSync`

Jarred-Sumner:
This adds password hashing and verifying functions to the `Bun` global and import. `Bun.password` runs each task in bun's thread pool and `passwordSync` runs on the same thread.

`bcrypt` and `argon2` is supported. `argon2` is the default.

```js
import {password, passwordSync} from "bun";

// uses argon2
const hash = await password.hash("hello world");

// reads the algorithm from the previous hash
const verify = await password.verify("hello world", hash);
console.log(verify); // true
```

```js
import {password, passwordSync} from "bun";

// uses bcrypt
const hash = await password.hash("hello world", "brypt");

// reads the algorithm from the previous hash, but you can specify it as well
const verify = await password.verify("hello world", hash);
console.log(verify); // true
```

cc @jedisct1, this uses `std.crypto.pwhash.bcrypt` and `std.crypto.pwhash.argon2` from Zig's standard library. Would love if you could look over this PR. There will be a couple unrelated test failures as CI isn't green right now

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -1013,6 +1013,174 @@ declare module "bun" {
     //     };
   }
 
+  type PasswordAlgorithm =
+    | "bcrypt"
+    | "argon2"
+    | "argon2id"
+    | "argon2d"
+    | "argon2i";
+
+  /**
+   * Hash and verify passwords using argon2 or bcrypt. The default is argon2.
+   * Password hashing functions are necessarily slow, and this object will
+   * automatically run in a worker thread.
+   *
+   * The underlying implementation of these functions are provided by the Zig
+   * Standard Library. Thanks to @jedisct1 and other Zig constributors for their
+   * work on this.
+   *
+   * ### Example with argon2
+   *
+   * ```ts
+   * import {password} from "bun";
+   *
+   * const hash = await password.hash("hello world");
+   * const verify = await password.verify("hello world", hash);
+   * console.log(verify); // true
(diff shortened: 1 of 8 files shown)


jedisct1:
Hi Jarred,

This is a nice and useful addition to Bun!

A few remarks, though.

`argon2` is not a specific algorithm. It's a family of algorithms. In that diff, it looks like it's more or less of an alias for `argon2id`, but that feels confusing.

It may be better to explicitly write `argon2id` everywhere, and not have `argon2` that doesn't refer to a specific algorithm.

Now, about the parameters. This is always the tricky thing about password hashing: it's very hardware- and application- dependent. That PR uses reasonable settings for interactive protocols. But for applications that can afford more CPU cycles and more memory, that may not be enough. So, maybe these parameters should be accessible to Bun applications.

Regarding `bcrypt`: good job for having paid attention to the 72 bytes limit! SHA-512 is a better fit (it's not as efficient for GPU attackers) than SHA-256 and ideally, it should be keyed. https://github.com/ziglang/zig/pull/15955 is going to take care of this automatically for you.

Jarred-Sumner:
Thanks for reviewing @jedisct1. I've added `memoryCost` and `timeCost` options for argon2, removed `"argon2"`, exposed `rounds` for `"bcrypt"`, and switched from SHA-256 to SHA-512 (though not keyed, but only because we don't have the BoringSSL binding for this setup yet)

PR #3213, opened 2023-06-05
allow `v` flag in regexp literal

### LINKED ISSUE #3212: Support new ECMAScript RegExp `v` flag
Ref: https://github.com/tc39/proposal-regexp-v-flag

JavaScriptCore has already implemented the new flag on runtime:
- https://github.com/WebKit/WebKit/pull/10984

I have verified Bun's current WebKit version used already contains the linked PR commit.

However Bun itself currently doesn't support the flag at parsing level on regex literals such as `/foo/v` (but can be "tricked" to accept it with all forms of string code/regex generation, `eval`, `Function`, `new RegExp(..., 'v')`, etc and it will run fine at runtime as predicted)

dylan-conway:
closes #3212

### DIFF
--- a/src/js_lexer.zig
+++ b/src/js_lexer.zig
@@ -2070,7 +2070,7 @@ fn NewLexer_(
                         lexer.step();
 
                         var has_set_flags_start = false;
-                        const flag_characters = "dgimsuyv";
+                        const flag_characters = "dgimsuvy";
                         const min_flag = comptime std.mem.min(u8, flag_characters);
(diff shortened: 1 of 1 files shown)


Jarred-Sumner (commented):
should be alphabetical, the order

en0ent1ty:
Awesome fast resolution, although one small thing, this seems to be missing the scenario of conflict between the `u` and `v` flags, they are not supposed to be allowed together. (It will still runtime error by JSC anyway of course, but I imagine Bun would like to keep its parser on par with the spec)

PR #3219, opened 2023-06-06
[breaking][bun:sqlite] `.values()` returns `[]` instead of `null`

Jarred-Sumner:
Not sure if we'll merge this yet

Previously, the following returned `null` due to no matching rows:
```js
db.query("SELECT * FROM foo WHERE id > 9999").values();
// null
```

After this PR, it returns `[]`:
```js
db.query("SELECT * FROM foo WHERE id > 9999").values();
// []
```

This is consistent with `.all()` and what people usually expect for functions that return an array of results - to always return an array and not change between an array and `null`.


simylein:
I would argue for the [] over null because this is the behaviour I am used to from various libraries

PR #3230, opened 2023-06-07
Add `resolveDir` to `onResolve.args`

colinhacks:


test/bundler/bundler_plugin.test.ts
```diff
@@ -830,4 +830,32 @@ describe("bundler", () => {
...
+        "index.ts": /* ts */ `
+          import "./foo.magic";
+        `,
+      },
+      entryPoints: ["./index.ts"],
+      plugins(build) {
+        build.onResolve({ "filter": /.magic$/ }, args => {
+          console.log({ root, resolveDir: args.resolveDir });
+          expect(args.resolveDir).toBeDefined();
+          expect(args.resolveDir!.replace("/private", "")).toEqual(root);
```
Jarred-Sumner: @paperdave root should be a resolved path. tmpdir is prefixed with /private on macOS

colinhacks: Fixed

Issue #2614, opened 2023-04-10
`bun remove` removes trailing newline from package.json


jakeboone02:
### What version of Bun is running?

`bun -v` says `0.6.0` but `npm list --global bun` says `0.5.10-canary.20230407.1+c8e09f5`

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

1. Generate a package.json with a trailing newline (LF) character.
2. `bun add` a dependency
3. `bun remove` a dependency

### What is the expected behavior?

The presence or absence of a trailing newline should remain consistent after `bun remove <pkg>`.

`bun add` does not exhibit the same problem.

### What do you see instead?

Transcript from my WSL2 zsh terminal:

```
~/misc-----------------------------------------------------------------------
$ mkdir temp
~/misc-----------------------------------------------------------------------
$ cd temp
~/misc/temp------------------------------------------------------------------
$ git init
Initialized empty Git repository in /home/jaboone/misc/temp/.git/
~/misc/temp--------------------------------------------------------------main
$ bun add cowsay                                                              <<<<< Add package
bun add v0.6.0 (373248ce)

 installed cowsay@1.5.0 with binaries:
  - cowsay
  - cowthink


 33 packages installed [2.03s]
~/misc/temp----------------------------------------------------------main[3?]
$ echo "" >> package.json                                                     <<<<< Add trailing newline
~/misc/temp----------------------------------------------------------main[3?]
$ git add package.json
~/misc/temp-------------------------------------------------------main[1+|2?]
$ bun remove cowsay                                                           <<<<< Remove package
bun remove v0.6.0 (373248ce)
[2.00ms] done
No packages! Deleted empty lockfile
~/misc/temp----------------------------------------------------main[1+|1!|1?]
$ git --no-pager diff package.json
diff --git a/package.json b/package.json
index 5743747..9e26dfe 100644
--- a/package.json
+++ b/package.json
@@ -1 +1 @@
-{ "dependencies": { "cowsay": "^1.5.0" } }
+{}
\ No newline at end of file                                                   <<<<< PROBLEM
~/misc/temp----------------------------------------------------main[1+|1!|1?]
$ echo "" >> package.json                                                     <<<<< Add trailing newline
~/misc/temp----------------------------------------------------main[1+|1!|1?]
$ git --no-pager diff package.json
diff --git a/package.json b/package.json
index 5743747..0967ef4 100644
[...]

PR #3231, opened 2023-06-07
Fix to retain a newline after removing a package

ytakhs:
This will fix https://github.com/oven-sh/bun/issues/2614

Also I added some tests for `bun remove` command.


Electroid:
Thanks for contributing and adding these tests! Queued your PR to run the actions test suite

Electroid (approved):

PR #3235, opened 2023-06-07
[install] fix life cycle scripts execution

alexlamsl:
- scripts from dependencies did not run during re-installation
- `bun.lockb` now contains data on life cycle scripts
- handle prior lockfile format by reading missing info from every `package.json`

### DIFF
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -63,22 +63,22 @@ const Dependency = @import("./dependency.zig");
 const Behavior = Dependency.Behavior;
 const FolderResolution = @import("./resolvers/folder_resolver.zig").FolderResolution;
 const Install = @import("./install.zig");
+const Aligner = Install.Aligner;
+const alignment_bytes_to_repeat_buffer = Install.alignment_bytes_to_repeat_buffer;
 const PackageManager = Install.PackageManager;
+const DependencyID = Install.DependencyID;
 const ExternalSlice = Install.ExternalSlice;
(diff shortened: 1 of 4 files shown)


src/install/lockfile.zig
```diff
@@ -1704,8 +1704,11 @@ pub const FormatVersion = enum(u32) {
     // bun v0.1.7+
     // This change added tarball URLs to npm-resolved packages
     v2,
+    // bun v0.6.8+
+    // Store life cycle scripts of resolved packages
+    v3,
     _,
-    pub const current = FormatVersion.v2;
+    pub const current = FormatVersion.v3;
```
Jarred-Sumner: Couldn't we do an in-place migration here to add the extra field and then a marker at the very end of the file that denotes whether this field will exist? As-is, this will automatically invalidate all existing lockfiles in Bun which shouldn't strictly be necessary

alexlamsl: except those older lockfiles would cause dependency life cycle scripts to not execute, so keeping them around will keep this bug alive in the wild long after this fix

alexlamsl: Done in https://github.com/oven-sh/bun/pull/3235/commits/f0cd071a508cd2e02832ff50096a9ef8ff699887 &minus; handle existing `bun.lockb` by obtaining life cycle scripts information from each package's `package.json`

Jarred-Sumner (approved):
LGTM

PR #3243, opened 2023-06-08
Add `pluginData` in bundler plugins

colinhacks:
Currently compiles but segfaults. It gets the data out from `onResolve` but segfaults when passing into `onLoad`

Not really sure how to do the memory management here


src/js/builtins/BundlerPlugin.ts
```diff
@@ -320,7 +320,7 @@ export function runOnLoadPlugins(this: BundlerPlugin, internalID, path, namespac
           path,
           namespace,
           // suffix
-          // pluginData
+          pluginData,
```
terracatta, line 323: @colinhacks Is this in the right order? I notice everywhere else pluginData is passed as the last argument?

I ask because I am highly interested in this PR and not sure if this was the cause of your segfaulting.

PR #3245, opened 2023-06-08
docs: add missing right parenthesis in example code

jamesliu96:
add missing right parenthesis in example code in `HTTP server` section


Electroid (approved):
Thanks!

PR #3249, opened 2023-06-08
[feat] fs.watch

cirospaciari:


src/js/out/modules/node/fs.js
```diff
@@ -1,3 +1,4 @@
+var {EventEmitter } = import.meta.require("node:stream");
```
Jarred-Sumner, line 1: This is not correct and I'm not surprised it works

cirospaciari:
@Jarred-Sumner 

Today once initialized we never deinit this loop, but we can deinit it when no watchers are active https://github.com/oven-sh/bun/blob/a448078534811b8e2ecba1df1309e4101ab12a88/src/bun.js/node/fs_events.zig#L541-L563

This implementation is also using babylist to keep track of active stuff and we reuse free slots, not the optimal way but should be fine when using 1-100 watchers https://github.com/oven-sh/bun/blob/a448078534811b8e2ecba1df1309e4101ab12a88/src/bun.js/node/fs_events.zig#L501-L508

This is the main logic https://github.com/oven-sh/bun/blob/a448078534811b8e2ecba1df1309e4101ab12a88/src/bun.js/node/fs_events.zig#L355-L406
and we use 1 thread for all watchers and just filter by path like uv does, watcher.zig is using 1 thread per watcher and maybe we can refactor it to do the same as fs_events, relative path logic is also much simpler on fs_events loop.

Jarred-Sumner:
🎉

src/js/out/modules/node/fs.js
L3P3: You mean the formatting?

PR #3252, opened 2023-06-08
Implement mocks in bun:test

Jarred-Sumner:
This implements:
- `expect(fn).toHaveBeenCalled()`
- `expect(fn).toHaveBeenCalledTimes(1)`

```ts
import { describe, expect, it, mock } from "bun:test";

describe("mocks", () => {
  it("are callable", () => {
    const fn = mock(() => 42);
    expect(fn()).toBe(42);
    expect(fn.mock.calls.length).toBe(1);
    expect(fn.mock.calls[0].length).toBe(0);
    expect(fn).toHaveBeenCalled();
    expect(fn).toHaveBeenCalledTimes(1);
  });
});

```

This leaves `toHaveBeenReturned` & friends to a subsequent PR. 

mockImplementationOnce() doesn't work right yet

### DIFF
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -697,8 +697,16 @@ pub const Jest = struct {
             Expect.getConstructor(globalObject),
         );
 
+        module.put(
+            globalObject,
+            ZigString.static("mock"),
+            JSMockFunction__createObject(globalObject),
+        );
+
         return module;
     }
+
+    extern fn JSMockFunction__createObject(*JSC.JSGlobalObject) JSC.JSValue;
 
     extern fn Bun__Jest__testPreloadObject(*JSC.JSGlobalObject) JSC.JSValue;
     extern fn Bun__Jest__testModuleObject(*JSC.JSGlobalObject) JSC.JSValue;
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -599,7 +599,7 @@ pub const Jest = struct {
     pub fn Bun__Jest__createTestModuleObject(globalObject: *JSC.JSGlobalObject) callconv(.C) JSC.JSValue {
         JSC.markBinding(@src());
 
-        const module = JSC.JSValue.createEmptyObject(globalObject, 7);
+        const module = JSC.JSValue.createEmptyObject(globalObject, 11);
 
         const test_fn = JSC.NewFunction(globalObject, ZigString.static("test"), 2, TestScope.call, false);
         module.put(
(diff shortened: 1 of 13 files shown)


src/bun.js/test/jest.zig
```diff
@@ -3821,7 +3829,92 @@ pub const Expect = struct {
...
+        if (not) {
+            const signature = comptime getSignature("toHaveBeenCalled", "<green>expected<r>", true);
+            const fmt = signature ++ "\n\nExpected: not <green>{any}<r>\n";
+            if (Output.enable_ansi_colors) {
+                globalObject.throw(Output.prettyFmt(fmt, true), .{calls.toFmt(globalObject, &formatter)});
+                return .zero;
+            }
+            globalObject.throw(Output.prettyFmt(fmt, false), .{calls.toFmt(globalObject, &formatter)});
+            return .zero;
+        }
```
dylan-conway, line 3887: the not not case needs to be handled here and in `toHaveBeenCalledTimes`

src/bun.js/test/jest.zig
```diff
@@ -3821,7 +3829,92 @@ pub const Expect = struct {
...
+
+        if (arguments.len < 1 or !arguments[0].isAnyInt()) {
+            globalObject.throwInvalidArguments("toHaveBeenCalledTimes() requires 1 integer argument", .{});
+            return .zero;
+        }
+
+        const times = arguments[0].coerce(i32, globalObject);
+
+        var pass = @intCast(i32, calls.getLength(globalObject)) >= times;
```
dylan-conway: should this be `==`?

Jarred-Sumner: a function that has been called 5 times has also been called 4 times

i didn't test if Jest behaves this way though

dylan-conway: just checked, looks like they test for the exact number of times

src/bun.js/test/jest.zig
```diff
@@ -3821,7 +3829,101 @@ pub const Expect = struct {
...
+        const value: JSValue = JSC.Jest.Expect.capturedValueGetCached(thisValue) orelse {
+            globalObject.throw("Internal consistency error: the expect(value) was garbage collected but it should not have been!", .{});
+            return .zero;
+        };
+
+        const calls = JSMockFunction__getCalls(value);
+        active_test_expectation_counter.actual += 1;
+
+        if (calls == .zero or !calls.jsType().isArray()) {
+            globalObject.throw("Expected value must be a mock function: {}", .{value});
```
Electroid, line 3857: The `if` check and error message don't seem to match up?

src/bun.js/test/jest.zig
```diff
@@ -3821,7 +3829,101 @@ pub const Expect = struct {
...
+            const fmt = signature ++ "\n\nExpected <green>{any}<r>\n";
+            if (Output.enable_ansi_colors) {
+                globalObject.throw(Output.prettyFmt(fmt, true), .{calls.toFmt(globalObject, &formatter)});
+                return .zero;
+            }
+            globalObject.throw(Output.prettyFmt(fmt, false), .{calls.toFmt(globalObject, &formatter)});
+            return .zero;
+        }
+
+        unreachable;
```
Electroid, line 3945: Can be removed

test/js/bun/test/mock-test.test.ts
```diff
@@ -0,0 +1,39 @@
+import { test, mock, expect } from "bun:test";
```
Electroid: looks good, would add test cases for:
* async fn that resolves
* async fn that rejects
* fn that returns already resolved promise
* fn called with multiple args
* fn that calls another mock fn

src/bun.js/test/jest.zig
Jarred-Sumner: that is how it detects if it's a mocked callback

PR #3253, opened 2023-06-09
feat(bun/test): Implement "bail" option for "bun test"

### LINKED ISSUE #3044: bun:test - option to abort the whole test run if one test fails

Exit the test suite immediately upon N number of failing test suite

```ts
test('0', () => {
  expect(true).toBeTruthy()
})
test('1', () => {
  expect(false).toBeTruthy() // this fails
})
test('2', () => {
  // this test still runs even the previous one failed
})
```

```ts

// new option (like @playwright/test)
test.configure({ maxFailures: 1 })

// or like jest: https://jestjs.io/docs/cli#--bailn

test('0', () => {
  expect(true).toBeTruthy()
})
test('1', () => {
  expect(false).toBeTruthy() // this fails
})
test('2', () => {
  // this test should not run if the previous one fails
})
```

_No response_

v3nntrnx:
https://jestjs.io/docs/configuration#bail-number--boolean

Fixes #3044

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -217,7 +217,7 @@ pub const Arguments = struct {
         clap.parseParam("--rerun-each <NUMBER>            Re-run each test file <NUMBER> times, helps catch certain bugs") catch unreachable,
         clap.parseParam("--only                           Only run tests that are marked with \"test.only()\"") catch unreachable,
         clap.parseParam("--todo                           Include tests that are marked with \"test.todo()\"") catch unreachable,
-        clap.parseParam("--bail <NUMBER>                  Exit the test suite after <NUMBER> failures") catch unreachable,
+        clap.parseParam("--bail <NUMBER>?                 Exit the test suite after <NUMBER> failures. If you do not specify a number, it defaults to 1.") catch unreachable,
     };
 
     const build_params_public = public_params ++ build_only_params;
@@ -395,6 +395,8 @@ pub const Arguments = struct {
                         Output.prettyErrorln("--bail expects a number greater than 0", .{});
                         Global.exit(1);
                     }
+                } else {
+                    ctx.test_options.bail = 1;
                 }
             }
             if (args.option("--rerun-each")) |repeat_count| {
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -390,6 +390,11 @@ pub const Arguments = struct {
                         Output.prettyErrorln("--bail expects a number: {s}", .{@errorName(e)});
                         Global.exit(1);
                     };
+
+                    if (ctx.test_options.bail == 0) {
+                        Output.prettyErrorln("--bail expects a number greater than 0", .{});
+                        Global.exit(1);
+                    }
                 }
             }
             if (args.option("--rerun-each")) |repeat_count| {
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -217,6 +217,7 @@ pub const Arguments = struct {
         clap.parseParam("--rerun-each <NUMBER>            Re-run each test file <NUMBER> times, helps catch certain bugs") catch unreachable,
         clap.parseParam("--only                           Only run tests that are marked with \"test.only()\"") catch unreachable,
         clap.parseParam("--todo                           Include tests that are marked with \"test.todo()\"") catch unreachable,
+        clap.parseParam("--bail <NUMBER>                  Exit the test suite after <NUMBER> failures") catch unreachable,
     };
 
     const build_params_public = public_params ++ build_only_params;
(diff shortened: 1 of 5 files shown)


src/cli/test_command.zig
```diff
@@ -200,6 +201,14 @@ pub const CommandLineReporter = struct {
         this.summary.fail += 1;
         this.summary.expectations += expectations;
         this.jest.tests.items(.status)[id] = TestRunner.Test.Status.fail;
+
+        if (this.jest.bail == this.summary.fail) {
```
Electroid, line 205: We should reuse the same code for printing the existing summary here so we don't have 2 different formats. I like the "bail out" message, so we can keep that if it's a bail.

src/cli.zig
```diff
@@ -383,7 +384,14 @@ pub const Arguments = struct {
                     };
                 }
             }
-            ctx.test_options.update_snapshots = args.flag("--update-snapshots");
+            if (args.option("--bail")) |bail| {
+                if (bail.len > 0) {
+                    ctx.test_options.bail = std.fmt.parseInt(u32, bail, 10) catch |e| {
+                        Output.prettyErrorln("--bail expects a number: {s}", .{@errorName(e)});
```
Electroid, line 390: Should add a check that `bail > 0`, since `bail = 0` doesn't make sense

TiranexDev: Oh yeah was a little typo

v3nntrnx: Wait, it checks for length of the parameter, so if it doesn't exist

TiranexDev: other numbers do it too

src/cli/test_command.zig
```diff
@@ -81,6 +81,7 @@ pub const CommandLineReporter = struct {
     summary: Summary = Summary{},
     prev_file: u64 = 0,
     repeat_count: u32 = 1,
+    start_time: i128 = 0,
```
Jarred-Sumner: start_time shouldn't need to be stored here, it's a global var

src/cli.zig
v3nntrnx: oh, i just realized by what you mean. sure i'll do it

src/cli.zig
```diff
@@ -383,7 +384,19 @@ pub const Arguments = struct {
...
+                        Global.exit(1);
+                    };
+
+                    if (ctx.test_options.bail == 0) {
+                        Output.prettyErrorln("--bail expects a number greater than 0", .{});
+                        Global.exit(1);
+                    }
+                }
```
Electroid, line 400: missing an `else` case, when you do `bun test --bail`, that implies a bail of `1`. so this should set `ctx.test_options.bail = 1;`

src/cli/test_command.zig
```diff
@@ -246,6 +252,19 @@ pub const CommandLineReporter = struct {
         this.summary.expectations += expectations;
         this.jest.tests.items(.status)[id] = TestRunner.Test.Status.todo;
     }
+
+    pub fn printSummary(this: *CommandLineReporter, summary: Summary, start_time: i128, total_tests: ?u32) void {
```
Electroid: Shouldn't this also print all the skipped/todo tests? also give the summary of 0 pass, 0 fail, etc.? That's what I would expect

Electroid (changes requested):

src/cli/test_command.zig
v3nntrnx: well the problem is the total_tests is not very total until all of the tests are run, so when it bails out we don't know amount of all tests

v3nntrnx: or did you mean the runned_tests variable?

v3nntrnx: oh i got you

Electroid: Since `start_time` is always `bun.start_time`, this doesn't need to be a function argument.

src/cli/test_command.zig
```diff
@@ -80,7 +80,7 @@ pub const CommandLineReporter = struct {
     last_dot: u32 = 0,
     summary: Summary = Summary{},
     prev_file: u64 = 0,
-    repeat_count: u32 = 1,
+    repeat_count: u32 = 0,
```
Electroid: Is this a drive-by bug fix? Can we make sure it doesn't break `--rerun-each`?

v3nntrnx: Idk how that happened, I swear-

paperclover (approved):
looks good, gonna let tests run. some non-related test fails will happen

paperclover (approved):
yeah

Jarred-Sumner (approved):

PR #3256, opened 2023-06-09
fix TS5096

huseeiin:


colinhacks:
Can you elaborate on the issue you're seeing? Are you seeing this in your application code?

huseeiin:
> Can you elaborate on the issue you're seeing? Are you seeing this in your application code?

vscode: Option 'allowImportingTsExtensions' can only be used when either 'noEmit' or 'emitDeclarationOnly' is set.

PR #3257, opened 2023-06-09
Improvements to `WebSocket` and `ServerWebSocket`

Electroid:
```js
import { serve } from "bun";

serve({
  fetch(request, server) {
    if (server.upgrade(request)) {
      return;
    }
    return new Response();
  },
  websocket: {
    open(ws) {
      ws.ping();
    },
    ping(ws, data) {
      ws.pong(data);
    },
    pong(ws, data) {
      ws.terminate();
    }
  }
});
```

# New stuff
* Default `binaryType` in `WebSocket` is now "nodebuffer" to match the server **[breaking]**
* `ws.terminate()` - abruptly closes the connection
* `ws.ping(data?)` - sends a ping
* `ws.pong(data?)` - sends a pong
# Bug fixes
* `ws.close(1000, "non-ascii")` was not encoded properly
* `hostname` and `pathname` were not encoded properly on WebSocket upgrade
* `idleTimeout` could crash if between `(0, 8]` due to uws limitation


src/bun.js/bindings/webcore/WebSocket.cpp
```diff
@@ -650,6 +650,211 @@ ExceptionOr<void> WebSocket::close(std::optional<unsigned short> optionalCode, c
...
+
+    // No exception is raised if the connection was once established but has subsequently been closed.
+    if (m_state == CLOSING || m_state == CLOSED) {
+        size_t payloadSize = message.length();
+        m_bufferedAmountAfterClose = saturateAdd(m_bufferedAmountAfterClose, payloadSize);
+        m_bufferedAmountAfterClose = saturateAdd(m_bufferedAmountAfterClose, getFramingOverhead(payloadSize));
+        return {};
+    }
+
+    // TODO: this->sendWebSocketString(message, Opcode::Ping);
```
Electroid: These are all stubbed out for now, mostly because there are framing bugs in `WebSocket` client that need to be resolved.

paperclover:
is the binary type default on the global WebSocket client? because that kind of breaking change would break the web websocket spec- ideally those should match the browser

Jarred-Sumner:
Bun already doesn’t support Blob for binaryType

Note that websocket messages say whether binary or text, it’s not a
decision we make

src/deps/uws.zig
```diff
@@ -884,17 +869,15 @@ pub const AnyWebSocket = union(enum) {
...
+    pub inline fn getBufferedAmount(this: AnyWebSocket) u32 {
+        return uws_ws_get_buffered_amount(this.is_ssl(), this.raw());
     }
 
-    pub fn getRemoteAddress(this: AnyWebSocket, buf: []u8) []u8 {
+    pub inline fn getRemoteAddress(this: AnyWebSocket, buf: []u8) []u8 {
         return switch (this) {
-            .ssl => this.ssl.getRemoteAddress(buf),
-            .tcp => this.tcp.getRemoteAddress(buf),
+            .ssl => return this.ssl.getRemoteAddress(buf),
```
Jarred-Sumner: this return is redundant

src/deps/uws.zig
```diff
@@ -784,95 +784,80 @@ extern fn us_socket_is_closed(ssl: i32, s: ?*Socket) i32;
...
-        return switch (this) {
-            .ssl => uws_ws_unsubscribe(1, this.raw(), topic.ptr, topic.len),
-            .tcp => uws_ws_unsubscribe(0, this.raw(), topic.ptr, topic.len),
-        };
+
+    pub inline fn unsubscribe(this: AnyWebSocket, topic: []const u8) bool {
+        return uws_ws_unsubscribe(this.is_ssl(), this.raw(), topic.ptr, topic.len);
     }
-    pub fn isSubscribed(this: AnyWebSocket, topic: []const u8) bool {
-        return switch (this) {
```
Jarred-Sumner: these patterns were actually deliberate

wanted to tell LLVM that these are separate functions

Jarred-Sumner: but i didn't inspect the LLVM IR to see if it actually did the monomorphization

src/js/thirdparty/ws.js
```diff
@@ -67,14 +65,18 @@ class BunWebSocket extends EventEmitter {
         this.emit("message", this.#fragments ? [encoded] : encoded, isBinary);
       }
     });
+    ws.addEventListener("ping", ({ data }) => {
```
Jarred-Sumner: this should only be added if the user starts listening for ping and pong?

src/js/thirdparty/ws.js
```diff
@@ -184,16 +186,11 @@ class BunWebSocket extends EventEmitter {
 
     if (typeof data === "number") data = data.toString();
 
-    // deviation: we don't support ping
-    emitWarning("ping()", "ws.WebSocket.ping() is not implemented in bun");
+    this.#ws.ping(data);
```
Jarred-Sumner: does this need a try/catch to call the callback with an error? will it ever throw?

src/js/thirdparty/ws.js
Electroid: It's already somewhat lazy. We don't generate the `ping` or `pong` event unless there is a listener in `WebSocket.cpp`. Should it be lazy here too?

test/js/web/websocket/websocket-client.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    type: ArrayBuffer,
+  },
+] as const;
+
+let servers: Subprocess[] = [];
+let clients: WebSocket[] = [];
+
+afterEach(() => {
+  for (const client of clients) {
+    client.terminate();
```
Jarred-Sumner, line 60: what is terminate? we use "shutdown" elsewhere?

test/js/web/websocket/websocket-client.test.ts
```diff
@@ -0,0 +1,273 @@
...
+        .catch(done);
+    },
+    { timeout: timeout ?? 1000 },
+  );
+}
+
+async function listen(): Promise<URL> {
+  const { pathname } = new URL("./websocket-server-echo.mjs", import.meta.url);
+  const server = spawn({
+    cmd: [nodeExe() ?? bunEnv(), pathname],
```
Jarred-Sumner: Why is this using node instead of bun?

test/js/web/websocket/websocket-client.test.ts
```diff
@@ -0,0 +1,273 @@
...
+  {
+    label: "arraybuffer",
+    type: ArrayBuffer,
+  },
+] as const;
+
+let servers: Subprocess[] = [];
+let clients: WebSocket[] = [];
+
+afterEach(() => {
```
Jarred-Sumner: This afterEach will definitely cause tests to hang due to Subprocess or WebSocket keeping the event loop alive

test/js/web/websocket/websocket-client.test.ts
Electroid: That's what the `ws` library does.

test/js/web/websocket/websocket-client.test.ts
```diff
@@ -0,0 +1,273 @@
...
+  const server = spawn({
+    cmd: [nodeExe() ?? bunEnv(), pathname],
+    cwd: import.meta.dir,
+    env: bunEnv,
+    stderr: "ignore",
+    stdout: "pipe",
+  });
+  servers.push(server);
+  for await (const chunk of server.stdout) {
+    const text = new TextDecoder().decode(chunk);
```
Jarred-Sumner, line 272: Won't this be flaky if the process writes to stdout multiple times?

test/js/web/websocket/websocket-server-echo.mjs
```diff
@@ -0,0 +1,94 @@
...
+const server = createServer();
+const wss = new WebSocketServer({
+  perMessageDeflate: false,
+  noServer: true,
+});
+
+server.on("listening", () => {
+  const { address, port, family } = server.address();
+  const { href } = new URL(family === "IPv6" ? `ws://[${address}]:${port}` : `ws://${address}:${port}`);
+  console.log(href);
```
Jarred-Sumner, line 13: oh it only writes here?

Jarred-Sumner: thats probably okay then

src/deps/uws.zig
```diff
@@ -784,95 +784,80 @@ extern fn us_socket_is_closed(ssl: i32, s: ?*Socket) i32;
...
             .ssl => this.ssl.as(Type),
             .tcp => this.tcp.as(Type),
         };
     }
 
-    pub fn close(this: AnyWebSocket) void {
-        const ssl_flag = @intFromBool(this == .ssl);
-        return uws_ws_close(ssl_flag, this.raw());
+    pub inline fn is_ssl(this: AnyWebSocket) u1 {
+        return if (comptime @hasDecl(@TypeOf(this), "ssl")) 1 else 0;
```
Jarred-Sumner: this should fail to compile

Jarred-Sumner: oh actually

it just always returns 1

Jarred-Sumner: so this will segfault for tcp connections most likely

Electroid: I removed all the changes to `uws.zig`, as it wasn't necessary for this PR.

src/js/out/modules/thirdparty/ws.js
```diff
@@ -59,51 +66,83 @@ class BunWebSocket extends EventEmitter {
   #paused = !1;
   #fragments = !1;
   #binaryType = "nodebuffer";
-  readyState = BunWebSocket.CONNECTING;
```
Jarred-Sumner, line 62: `readyState` is an expected attribute here. https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/readyState

Jarred-Sumner: Though we could make it a getter which just returns from the underlying websocket.

src/http/websocket_http_client.zig
```diff
@@ -96,7 +101,7 @@ fn buildRequestBody(
             "{any}" ++
             "{any}" ++
             "\r\n",
-        .{ pathname_, host_, pico_headers, extra_headers },
+        .{ pathname_.slice(), host_.slice(), pico_headers, extra_headers },
```
Jarred-Sumner, line 104: the previous version of this would be the same btw

the formatter converts it to UTF-8.

Jarred-Sumner: If you want to keep this change, it should be `{s}` instead of `{any}`. `{any}` is unspecified for strings

src/http/websocket_http_client.zig
```diff
@@ -1002,16 +1008,15 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
                         out.didReceiveText(true, &outstring);
                     }
                 },
-                .Binary => {
+                .Binary, .Ping, .Pong => {
                     JSC.markBinding(@src());
-                    out.didReceiveBytes(data_.ptr, data_.len);
+                    out.didReceiveBytes(data_.ptr, data_.len, @as(u8, @intFromEnum(kind)));
```
Jarred-Sumner, line 1013: you should be able to send enums without the `@intFromEnum` so long as `Opcode` has a known size

src/http/websocket_http_client.zig
```diff
@@ -1419,16 +1431,20 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
             var mask_buf: *[4]u8 = final_body_bytes[2..6];
             std.mem.writeIntSliceBig(u16, final_body_bytes[6..8], code);
 
+            var reason = JSC.ZigString.Empty;
             if (body) |data| {
-                if (body_len > 0) @memcpy(final_body_bytes[8..][0..body_len], data[0..body_len]);
+                if (body_len > 0) {
+                    reason = JSC.ZigString.init(data[0..body_len]);
```
Jarred-Sumner: I don't think this is correct? It should use the copy from final_body_bytes instead of data. data is temporary. This will segfault if it's not cloned.

Jarred-Sumner: I suggest making this a BunString and then doing `bun.String.create(data[0..body_len])`, assuming reason is always a string.

src/bun.js/bindings/webcore/WebSocket.cpp
```diff
@@ -1225,19 +1427,35 @@ extern "C" void WebSocket__didConnect(WebCore::WebSocket* webSocket, us_socket_t
...
-    webSocket->didReceiveBinaryData({ bytes, len });
+    auto opcode = static_cast<WebCore::WebSocket::Opcode>(op);
+    switch (opcode) {
+    case WebCore::WebSocket::Opcode::Binary:
+        webSocket->didReceiveBinaryData("message"_s, { bytes, len });
+        break;
+    case WebCore::WebSocket::Opcode::Ping:
+        webSocket->didReceiveBinaryData("ping"_s, { bytes, len });
+        break;
+    case WebCore::WebSocket::Opcode::Pong:
```
Jarred-Sumner, line 1455: This might add a warning about missing default switch case

src/bun.js/bindings/webcore/WebSocket.cpp
```diff
@@ -1051,158 +1253,158 @@ void WebSocket::didFailWithErrorCode(int32_t code)
     // invalid_response
     case 1: {
         static NeverDestroyed<String> message = MAKE_STATIC_STRING_IMPL("Invalid response");
-        didReceiveMessageError(1002, message);
+        didReceiveClose(false, 1002, message);
```
Jarred-Sumner: I would feel a little better if this was an enum instead of a boolean because false and true are very ambiguous here.

src/bun.js/bindings/webcore/WebSocket.cpp
```diff
@@ -650,6 +648,211 @@ ExceptionOr<void> WebSocket::close(std::optional<unsigned short> optionalCode, c
...
+        return Exception { InvalidStateError };
+
+    if (m_state == CLOSING || m_state == CLOSED) {
+        unsigned payloadSize = arrayBufferView.byteLength();
+        m_bufferedAmountAfterClose = saturateAdd(m_bufferedAmountAfterClose, payloadSize);
+        m_bufferedAmountAfterClose = saturateAdd(m_bufferedAmountAfterClose, getFramingOverhead(payloadSize));
+        return {};
+    }
+
+    auto buffer = arrayBufferView.unsharedBuffer().get();
```
Jarred-Sumner, line 766: I'm not sure if this is correct. What happens if you pass a `new SharedArrayBuffer(length)` or a `new Uint8Array(sab)`?

Should it be `arrayBufferView.impl()`?

src/js/out/modules/thirdparty/ws.js
Electroid: That's what I did

src/http/websocket_http_client.zig
Electroid: Previous version would segfault

src/bun.js/bindings/webcore/WebSocket.cpp
Electroid: That method doesn't exist, also copied from the existing `send`

Jarred-Sumner (approved):
LGTM after tests run

Electroid:
Looks like there's a flaky test with the websocket server, will take a look. (it's probably uws related)

PR #3260, opened 2023-06-09
`toMatchObject` and some asymmetric matchers

dylan-conway:
This adds `toMatchObject` and asymmetric matchers:
 - `expect.anything()`
 - `expect.stringContaining()`
 - `expect.stringMatching()`

The inverse versions, `expect.not.stringContaining()` and `expect.not.stringMatching()` are not implemented yet


src/bun.js/bindings/bindings.cpp
```diff
@@ -723,6 +723,165 @@ bool Bun__deepEquals(JSC__JSGlobalObject* globalObject, JSValue v1, JSValue v2,
...
+        JSObject* constructorObject = constructorValue.getObject();
+
+        if (constructorObject->hasInstance(globalObject, otherProp)) {
+            return AsymmetricMatcherResult::PASS;
+        }
+
+        // check for basic types
+        VM& vm = globalObject->vm();
+        ZigString name = {};
+        JSC__JSValue__getNameProperty(JSValue::encode(constructorValue), globalObject, &name);
```
Jarred-Sumner, line 751: We don't need to look up the constructor name

We can rely on the `JSType`

- `type() == JSType::NumberObject`
- `type() == JSType::StringObject`
- `type() == JSType::DerivedStringObject`

This also doesn't account for DerivedArray (class extends Array)

Jarred-Sumner: look at the code for util/types for examples of this

src/bun.js/bindings/bindings.cpp
```diff
@@ -723,6 +723,165 @@ bool Bun__deepEquals(JSC__JSGlobalObject* globalObject, JSValue v1, JSValue v2,
...
+        JSCell* subsetPropCell = subsetProp.asCell();
+        JSCell* propCell = prop.asCell();
+
+        if (subsetProp.isCell() and subsetPropCell->type() == JSC::JSType(JSDOMWrapperType)) {
+            switch (matchAsymmetricMatcher(globalObject, subsetPropCell, prop, throwScope)) {
+            case AsymmetricMatcherResult::FAIL:
+                return false;
+            case AsymmetricMatcherResult::PASS:
+                if (replacePropsWithAsymmetricMatchers) {
+                    obj->putDirect(vm, subsetProps[i], subsetProp);
```
Jarred-Sumner, line 857: I think putDirect fails if the object is big enough, like if it has no inline storage left

but not sure

src/bun.js/bindings/bindings.cpp
```diff
@@ -723,6 +723,165 @@ bool Bun__deepEquals(JSC__JSGlobalObject* globalObject, JSValue v1, JSValue v2,
...
+    VM& vm = globalObject->vm();
+    JSObject* obj = objValue.getObject();
+    JSObject* subsetObj = subsetValue.getObject();
+
+    PropertyNameArray subsetProps(vm, PropertyNameMode::StringsAndSymbols, PrivateSymbolMode::Include);
+    subsetObj->getPropertyNames(globalObject, subsetProps, DontEnumPropertiesMode::Exclude);
+
+    // arrays should match exactly
+    if (isArray(globalObject, objValue) && isArray(globalObject, subsetValue)) {
+        if (obj->getArrayLength() != subsetObj->getArrayLength()) {
```
Jarred-Sumner, line 827: if they're both arrays this should probably not go through getPropertyNames and instead have a separate path for that? Since it will let us loop through keys faster?

dylan-conway: I think it needs to go through getPropertyNames when it's an array for situations like this:
```js
const f = Symbol.for("foo");

var a1 = [1];
a1[f] = 99;
expect(a1).toMatchObject([1]);
```
This should fail. Also the code in this commit doesn't fail for this case, will be updated soon

Jarred-Sumner: i think the Structure stores whether or not it has dictionary keys like that

can you add a TODO comment that says we should add a fast path for:
- two "simple" objects (using ->forEachProperty in both)
- two "simple" arrays

Similar to what we have in deepEquals

dylan-conway: sounds good

src/bun.js/bindings/bindings.cpp
dylan-conway: I don't think we can use `type()` here because the values are not cell types. do you know if it's possible to get the jstype of primitives like number or boolean?

test/js/bun/test/test-test.test.ts
```diff
@@ -2522,6 +2522,215 @@ test("toBeOdd()", () => {
...
+      constructor(str) {
+        super(str);
+      }
+    }
+    expect({ a: "hello world" }).toMatchObject({ a: expect.stringContaining("wor") });
+    expect({ a: "hello world" }).not.toMatchObject({ a: expect.stringContaining("wol") });
+    expect({ a: "hello String" }).toMatchObject({ a: expect.stringContaining(new String("Str")) });
+    expect({ a: "hello String" }).not.toMatchObject({ a: expect.stringContaining(new String("Strs")) });
+    expect({ a: "hello derived String" }).toMatchObject({ a: expect.stringContaining(new DString("riv")) });
+    expect({ a: "hello derived String" }).not.toMatchObject({ a: expect.stringContaining(new DString("rivd")) });
```
Jarred-Sumner, line 2625: great tests

PR #3268, opened 2023-06-10
feat(bun/test): Implement "toSatisfy" & "toIncludeRepeated"

v3nntrnx:
_toSatisfy:_

`expect(value).toSatisfy(function)`, the function passed to "toSatisfy" function is called with the value in expect, so it's almost the same as calling:

`expect(function(value)).toBeTrue();`
More about it: https://vitest.dev/api/expect.html#tosatisfy

What I already have done for toSatisfy:

- [x] Types
- [x] Testing
- [x] Implement toSatisfy itself

---

_toIncludeRepeated:_

More about it: https://jest-extended.jestcommunity.dev/docs/matchers/string/#toincluderepeatedsubstring-times

What I already have done for toIncludeRepeated:

- [x] Types
- [x] Testing
- [x] Implement toIncludeRepeated itself

<sub>_Huge thanks to @inKryption for helping me with toIncludeRepeated() <3_</sub>


TiranexDev:
I'll upload the bindings and stuff when someone reviewed my code and it's ready to be merged.

src/bun.js/test/jest.zig
```diff
@@ -3592,6 +3592,147 @@ pub const Expect = struct {
...
+
+        if (!count.isNumber()) {
+            globalThis.throw("toIncludeRepeated() requires the second argument to be a number", .{});
+            return .zero;
+        }
+
+        const countAsNum = std.fmt.parseInt(u32, count.toString(globalThis).toSlice(globalThis, default_allocator).slice(), 10) catch {
+            globalThis.throw("toIncludeRepeated() requires the second argument to be a number", .{});
+            return .zero;
+        };
```
Electroid: There is a more efficient way to do this, that doesn't require allocation. Instead of checking `isNumber()` in the code before this highlight, you can use `isAnyInt()`.

src/bun.js/test/jest.zig
```diff
@@ -3592,6 +3592,147 @@ pub const Expect = struct {
...
+
+        if (not) {
+            const expected_line = "Expected to not include: <green>{any}<r> <green>{any}<r> times\n";
+            const received_line = "Received: <red>{any}<r>\n";
+            const fmt = comptime getSignature("toIncludeRepeated", "<green>expected<r>", true) ++ "\n\n" ++ expected_line ++ received_line;
+            globalThis.throwPretty(fmt, .{ substring_fmt, times_fmt, expect_string_fmt });
+            return .zero;
+        }
+
+        const expected_line = "Expected to include: <green>{any}<r> <green>{any}<r> times \n";
```
Electroid: How about this:

* If times = 0: `Expected to not include: {}`
* If times = 1: `Expected to include: {}`
* If times > 1: `Expected to include: {} ({} times)`
* If not and times = 0: `Expected to include: {}`
* If not and times = 1: `Expected not to include: {}`
* If not and times > 1: `Expected not to include: {} ({} times)`

src/bun.js/test/jest.zig
```diff
@@ -3592,6 +3592,147 @@ pub const Expect = struct {
...
+            globalThis.throw("toSatisfy() must be called in a test", .{});
+            return .zero;
+        }
+
+        active_test_expectation_counter.actual += 1;
+
+        const right = arguments[0];
+        right.ensureStillAlive();
+
+        if (!right.jsType().isFunction()) {
```
Electroid:
```suggestion
        if (!right.isCallable(globalThis)) {
```

src/bun.js/test/jest.zig
```diff
@@ -3592,6 +3592,147 @@ pub const Expect = struct {
...
+        }
+
+        if (this.scope.tests.items.len <= this.test_id) {
+            globalThis.throw("toSatisfy() must be called in a test", .{});
+            return .zero;
+        }
+
+        active_test_expectation_counter.actual += 1;
+
+        const right = arguments[0];
```
Electroid: "right" and "left" aren't really descriptive. the expect value can be "value" and the argument can be "predicate"

src/bun.js/test/jest.zig
```diff
@@ -3592,6 +3592,147 @@ pub const Expect = struct {
...
+        }
+
+        const left = Expect.capturedValueGetCached(thisValue) orelse {
+            globalThis.throw("Internal consistency error: the expect(value) was garbage collected but it should not have been!", .{});
+            return .zero;
+        };
+        left.ensureStillAlive();
+
+        const not = this.op.contains(.not);
+        var pass = right.call(globalThis, &.{left}).toBoolean();
```
Electroid: We should print the result of the satisfy function, for example:

```zig
var result = predicate.call(globalThis, &{value});
if (result.toError()) |err| {
  // Predicate threw an error: {}
} else {
  // Received: value
}
var pass = (result.isBoolean() and result.toBoolean()) != not;
```

test/js/bun/test/expect.test.ts
```diff
@@ -354,6 +354,33 @@ describe("expect()", () => {
     expect("bob").not.toInclude("alice");
   });
 
+  test("toIncludeRepeated()", () => {
```
Electroid, line 357: Should add test cases for:
* n = 0
* n = 0 with not
* value is empty string
* include is empty string
* both value and include are empty strings
* emoji strings

Electroid (changes requested):
Thanks for working on this, left a few comments for review.

src/bun.js/test/jest.zig
```diff
@@ -3592,6 +3592,202 @@ pub const Expect = struct {
...
+
+        const count = arguments[1];
+        count.ensureStillAlive();
+
+        if (!count.isAnyInt()) {
+            globalThis.throw("toIncludeRepeated() requires the second argument to be a number", .{});
+            return .zero;
+        }
+
+        const countAsNum = std.fmt.parseInt(u32, count.toString(globalThis).toSlice(globalThis, default_allocator).slice(), 10) catch {
```
Jarred-Sumner: We shouldn't use the zig std lib function for this. Instead, I think we can use .toNumber? One of the functions in JSValue will call `[Symbol.toPrimitive()]` and pass it a `"number"` which would do the coercions equivalent to `Number(string)` in JS.

v3nntrnx: I‘ll be on it when I come home

v3nntrnx:
I'll open a new one, my git has some troubles

PR #3270, opened 2023-06-10
Fix wording and code example in docs for Bun.sleepSync

paperluigis:


Jarred-Sumner:
Thank you for this

The current example in the docs is wrong, not just because it shows Bun.sleep in the `Bun.sleepSync` section, but also because Bun.sleepSync doesn't support a `Date` object. I suggest instead deleting the paragraph with the example entirely

paperluigis:
that is done

Jarred-Sumner:
Thank you

PR #3275, opened 2023-06-10
Implement `mock().mock.lastCall`

Electroid:


Jarred-Sumner:
Instead of creating a JSCustomGetterSetter for every instance of the mock object, we can create a single prototype object and make the Structure use that. Then only one CustomGetterSetter is needed. That should reduce memory usage, at a slight cost to getter being called

paperclover:
implemented in #3304 instead

PR #3276, opened 2023-06-10
Swap Deno.serve() arguments

lpinca:
The `options` argument comes first.


lpinca:
See https://deno.land/api@v1.34.2?s=Deno.serve&unstable=.

Electroid (approved):
Thanks!

Issue #3203, opened 2023-06-04
[bug]: bun install args are order-dependent


ImLunaHey:
### What steps can reproduce the bug?

Start with a new dir with no `package.json`
```
➜  ct-logs bun install --save certstream

error: MissingPackageJSON
bun could not find a package.json file.
➜  ct-logs bun install certstream --save
bun add v0.6.7 (59d7c47e)

 installed certstream@1.1.1

[6.00ms] done
```

### What is the expected behavior?

The order of the args not to matter.

### What do you see instead?

```
error: MissingPackageJSON
bun could not find a package.json file.
```

### Additional information

https://canary.discord.com/channels/876711213126520882/876711213126520885/1114778392412291132

PR #3277, opened 2023-06-11
add --save argument to install

kvakil:
When bun encounters a long opt argument which it doesn't understand, it immediately stops parsing all remaining arguments (see src/deps/zig-clap/clap/streaming.zig:96). This leads to some weird behavior:

1. We currently don't define `--save` as a possible argument to `bun install`. This meant that `bun install --save X` would be incorrectly parsed as `bun install`.
2. On the other hand, `bun install X --save` would be parsed as `bun install X`. But that would be redirected to `bun add` and parsed as `bun add X`, which works fine.

To fix this, define `--save` as a possible argument for `bun install`. This in turn requires some refactoring because we used to use the presence of `--save` in the argument parser definition to determine if the subcommand should default to modifying package.json or not. Instead we introduce a Subcommand enum and clean that logic up, along with some similar subcommand-specific logic.

A future diff may fix the long opt behavior, assuming I'm not missing a good reason for its existence. :)

Fixes: https://github.com/oven-sh/bun/issues/3203

### DIFF
--- a/src/cli/package_manager_command.zig
+++ b/src/cli/package_manager_command.zig
@@ -51,7 +51,7 @@ pub const PackageManagerCommand = struct {
         @memcpy(&lockfile_buffer, lockfile_.ptr, lockfile_.len);
         lockfile_buffer[lockfile_.len] = 0;
         var lockfile = lockfile_buffer[0..lockfile_.len :0];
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
LGTM

PR #3278, opened 2023-06-11
doc: minor fixes

kvakil:
* `bun link` only modifies `package.json` when `--save` is passed.
* clang-15 is a separate package from llvm-15.


Electroid (approved):
Thanks!

Issue #1562, opened 2022-11-30
FileNotFound when filename is not English


5d-jh:
### What version of Bun is running?

0.2.2

### What platform is your computer?

Darwin 21.6.0 Darwin Kernel Version 21.6.0: Thu Sep 29 20:13:56 PDT 2022; root:xnu-8020.240.7~1/RELEASE_ARM64_T6000 arm64 arm

### What steps can reproduce the bug?

1. Create a file/directory that isn't english.
2. `bun run ./<filename_that_is_not_english>.ts`

### What do you see instead?

```
error: FileNotFound reading "/Users/.../src/ì±ëª¨ë.ts"
```

PR #3279, opened 2023-06-11
Support using `WTF::StringImpl` from Zig

Jarred-Sumner:
This enables zero-copy sharing of strings between JavaScript/C++ and Zig

For now, this is only enabled for module resolution but it fixes https://github.com/oven-sh/bun/issues/1562


cirospaciari (approved):

cirospaciari:
LGTM

PR #3282, opened 2023-06-11
bump typescript and bun types

breno-alves:
update typescript and bun-types version in init command.

image


breno-alves:
closing as it got fixed in https://github.com/oven-sh/bun/pull/3305.

colinhacks:
Thanks @breno-alves. As you saw we decided to just stick with `latest` for bun types. As for TypeScript, we rely on `allowImportingTsExtensions` which was introduced in 5.0. Beyond that, we don't really care which version ppl use so `^5.0.0` is the least restrictive semver range. A fresh install with `^5.0.0` will install 5.1.x anyway. Thanks!

PR #3286, opened 2023-06-12
typo in readline

paperclover:
this variable isn't used or defined. doesn't seem to unblock any issues directly but it does fix a thrown referenceerror when using readline


Jarred-Sumner:
Thank you

PR #3288, opened 2023-06-12
[install] support trustedDependencies

### LINKED ISSUE #2073: Support `trustedDependencies` in `package.json`
There are npm packages, like `esbuild`, that need `postinstall` support to install binaries or run other scripts. While we could allow any package to run scripts, that wouldn't be the safest thing to do. We also can't _not_ support it at all, because these packages would simply not work.

We're thinking of adding a new `trustedDependencies` property to the `package.json`, which would require developers to explicitly define an allowlist of packages that are allowed to run scripts.

```json
{
  "private": true,
  "dependencies": {
    "esbuild": "^0.17.8"
  },
  "trustedDependencies": [
    "esbuild"
  ]
}
```

This would not extend to child dependencies, so if package `foo` depends on `bar`, and `foo` is on the allowlist, `bar` is _not_ on the allowlist, unless it is explicitly added.

alexlamsl:
closes #2073

### DIFF
--- a/src/install/extract_tarball.zig
+++ b/src/install/extract_tarball.zig
@@ -365,7 +365,7 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
     if (switch (this.resolution.tag) {
         // TODO remove extracted files not matching any globs under "files"
         .github, .local_tarball, .remote_tarball => true,
-        else => strings.indexEqualAny(this.package_manager.lockfile.trusted_dependencies.items, name) != null,
+        else => this.package_manager.lockfile.trusted_dependencies.contains(@truncate(u32, Semver.String.Builder.stringHash(name))),
     }) {
         const json_file = final_dir.openFileZ("package.json", .{ .mode = .read_only }) catch |err| {
             this.package_manager.log.addErrorFmt(
--- a/src/install/extract_tarball.zig
+++ b/src/install/extract_tarball.zig
@@ -157,13 +157,14 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
     var tmpdir = this.temp_dir;
     var tmpname_buf: [256]u8 = undefined;
     const name = this.name.slice();
-
-    var basename = this.name.slice();
-    if (basename[0] == '@') {
-        if (std.mem.indexOfScalar(u8, basename, '/')) |i| {
-            basename = basename[i + 1 ..];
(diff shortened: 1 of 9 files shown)


Electroid:
Note: we're gonna rename it to `trustedDependencies`

src/install/extract_tarball.zig
```diff
@@ -157,13 +157,14 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
...
     var tmpname_buf: [256]u8 = undefined;
     const name = this.name.slice();
-
-    var basename = this.name.slice();
-    if (basename[0] == '@') {
-        if (std.mem.indexOfScalar(u8, basename, '/')) |i| {
-            basename = basename[i + 1 ..];
+    const basename = brk: {
+        if (name[0] == '@') {
+            if (std.mem.indexOfScalar(u8, name, '/')) |i| {
```
Jarred-Sumner: nit: `strings.indexOfChar(name, '/')` is faster (though it shouldn't matter here)

src/install/lockfile.zig
```diff
@@ -3069,6 +3073,37 @@ pub const Package = extern struct {
             }
         }
 
+        if (comptime features.privileged_dependencies) {
+            if (json.asProperty("privilegedDependencies")) |q| {
+                switch (q.expr.data) {
+                    .e_array => |arr| {
+                        try lockfile.privileged_dependencies.ensureUnusedCapacity(allocator, arr.items.len);
+                        for (arr.slice()) |item| {
```
Jarred-Sumner, line 3124: i'm sorry this is a really annoying suggestion but

could we count the total length here and then use StringBuilder? These might become a long list

src/install/extract_tarball.zig
alexlamsl: Replaced all usages of `std.mem.indexOfScalar(u8, ...)` and `std.mem.lastIndexOfScalar(u8, ...)` within `src/install`

src/install/lockfile.zig
alexlamsl: Dropped strings altogether &minus; store and compares hashes instead 😉

arcanis:
FYI Yarn already has a syntax in the `package.json` for this; docs [here](https://yarnpkg.com/configuration/manifest#dependenciesMeta.built):

```json
{
  "dependenciesMeta": {
    "some-dep": {
      "built": true
    }
  }
}
```

PR #3290, opened 2023-06-12
workaround quote escape issues for `bun run`

### LINKED ISSUE #53: Spaces and quotes aren't handled correctly by bun run
package.json:
```json
{
  "scripts": {
    "argv": "node -e 'console.log(process.argv)'"
  }
}
```

Screenshot 2021-11-05 at 1 55 05 AM

alexlamsl:
fixes #53

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -192,6 +192,24 @@ pub const RunCommand = struct {
 
                     delimiter = 0;
                 },
+                // TODO: handle escape sequences properly
+                // https://github.com/oven-sh/bun/issues/53
+                '\\' => {
+                    delimiter = 0;
+
+                    if (entry_i + 1 < script.len) {
+                        switch (script[entry_i + 1]) {
+                            '"', '\'' => {
+                                entry_i += 1;
+                                continue;
+                            },
+                            '\\' => {
+                                entry_i += 1;
+                            },
+                            else => {},
+                        }
+                    }
+                },
                 else => {
                     delimiter = 0;
                 },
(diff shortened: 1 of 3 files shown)


src/cli/run_command.zig
```diff
@@ -192,6 +192,22 @@ pub const RunCommand = struct {
 
                     delimiter = 0;
                 },
+                // TODO: handle escape sequences properly
+                // https://github.com/oven-sh/bun/issues/53
+                '\\' => {
+                    delimiter = 0;
+
+                    switch (script[entry_i + 1]) {
```
Jarred-Sumner: does this need a bounds check for entry_i + 1 < script.len

Jarred-Sumner: `"foo": "bar\\\"`

test/harness.ts
```diff
@@ -81,7 +81,7 @@ export function hideFromStackTrace(block: CallableFunction) {
 }
 
 export function tempDirWithFiles(basename: string, files: Record<string, string>) {
-  const dir = fs.mkdtempSync(path.join(os.tmpdir(), basename + "_"));
+  const dir = fs.mkdtempSync(path.join(fs.realpathSync(os.tmpdir()), basename + "_"));
```
Electroid, line 84: Looks like there are some new `test-test.test.ts` failures, might be from this change?

alexlamsl: 1. tests on `bun-darwin-aarch64` weren't even running before due to `sudo`, which was fixed recently: #3254
2. `test-test.test.ts` is also failing in a separate PR: https://github.com/oven-sh/bun/actions/runs/5248738253/jobs/9480987433?pr=3288#step:6:7

src/cli/run_command.zig
alexlamsl: Oops &minus; while that's not valid JavaScript/JSON, we shouldn't segfault either... 😓

Jarred-Sumner:
Thank you

Issue #2916, opened 2023-05-17
`process.argv[0]` is always "bun" in the executable produced by `bun build --compile`


rickymohk:
### What version of Bun is running?

0.6.1

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

Given `index.js`
```js
console.log(process.argv);
```
Run 
```bash
bun build ./index.js --compile --outfile mycli
./mycli
```
prints
```
[ "bun", "compiled://root/mycli" ]
```

### What is the expected behavior?

`process.argv[0]` should be "path/to/mycli"

### What do you see instead?

`process.argv[0]` is always "bun"

### Additional information

I known `Bun.argv` exists and it represents the command line arguments exactly. But it is not as useful because of some inconsistency in some situation. Specifically when running with the `bun` command, one can run `bun index.js <param>` or `bun run index.js <param>`. As a result `<param>` may come in either `Bun.argv[2]` or `Bung.argv[3]`, while being in `Bung.argv[1]` when running `./mycli <param>` after compile. In contrast, using `process.argv` always have `<param>` starting from `process.argv[2]`. This makes me prefer `process.argv` over `Bun.argv`.

Jarred-Sumner:
The behavior with `process.argv` is intentional because existing scripts do `process.argv.slice(2)`, so continuing to make that work is least likely to break code

TBH, I didn't think about it as much for `Bun.argv`. Maybe we should do the same there.

rickymohk:
> The behavior with `process.argv` is intentional because existing scripts do `process.argv.slice(2)`, so continuing to make that work is least likely to break code
> 
> TBH, I didn't think about it as much for `Bun.argv`. Maybe we should do the same there.

I don't get it. What does `process.argv.slice(2)` have to do with `process.argv[0]` being `bun`? Shouldn't `process.argv[0]` always be the executing file? I know `pkg` does that.

ckhatton-transreport:
Yeah I totally agree - `Bun.argv` should mirror the behaviour of `process.argv` (if `Bun.argv` is to be kept). I am going to switch to using `process.argv` now that I know of the risk of a wrong argument appearing in `Bun.argv`.

I actually stumbled across a TypeScript error when using `Bun.argv`, which I will report in a bit.

`Bun.argv.slice(3)`

> Property 'argv' does not exist on type 'typeof import("bun")'.

PR #3302, opened 2023-06-13
Add typedefs for `Bun.argv`

### LINKED ISSUE #3293: TypeScript error when using `Bun.argv`

0.6.5

Darwin 22.5.0 arm64 arm

1. Create a new directory
2. `bun init`
4. Add this code to `index.ts`:
```js
const rawArgs: string[] = Bun.argv.slice(3);
```
5. In VS Code hover your mouse over `argv`.

There should be no TypeScript error.

### What do you see instead?

> Property 'argv' does not exist on type 'typeof import("bun")'.

I would have a guess that the TypeScript definition is missing for `argv` (string array) within the module, even though `argv` still functions.

### Additional information

Mentioned in issue #2916.

paperclover:
Closes #3293

`Bun.argv` is weird, it's the raw arguments. Not 100% sure if the current behavior is what we even want, but maybe there is some value to being able to access the raw arguments.

Maybe later we should have a nice api like `Bun.args` which is equivalent of `process.argv.slice(2)`; that's what i assume the main usecase of process.argv is.

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -51,6 +51,10 @@ declare module "bun" {
    *
    */
   export const env: Env;
+  /**
+   * The raw arguments passed to the process, including flags passed to Bun. If you want to easily read flags passed to your script, consider using `process.argv` instead.
+   */
+  export const argv: string[];
   export const origin: string;
 
   /**


Jarred-Sumner:
lets just make it the same as process.argv

I don't think anyone wants to use the current version of it. It's just confusing.

ckhatton-transreport:
Thank you for fixing this 🙏

> Bun.argv is weird, it's the raw arguments

And the first index is always "bun" as mentioned in #2916.

PR #3304, opened 2023-06-14
finish implementing JSMockFunction prototype + lots of expect test organizing

paperclover:
- `expect()` does not throw anymore, and is equivalent to `expect(undefined)`
- implement all of JSMockFunction.prototype
- test test orginization + making it run in jest


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "all-the-package-names": "^2.0.833",
```
Jarred-Sumner: is this intentional

Jarred-Sumner:
@paperdave wanna fix the things causing the test failures? 

It's _probably_ here

https://github.com/oven-sh/bun/blob/171ba6327d995bdde473e1b723f449b3e88d1c14/src/bun.js/bindings/JSMockFunction.cpp#L336-L353

package.json
paperclover: no it was not let me undo that.

src/bun.js/test/jest.zig
```diff
@@ -698,22 +698,22 @@ pub const Jest = struct {
...
 
         const jest = JSValue.createEmptyObject(globalObject, 3);
-        jest.put(globalObject, ZigString.static("fn"), mock_object);
+        jest.put(globalObject, ZigString.static("fn"), mock_fn);
         jest.put(globalObject, ZigString.static("spyOn"), spyOn);
         jest.put(globalObject, ZigString.static("restoreAllMocks"), restoreAllMocks);
         module.put(globalObject, ZigString.static("jest"), jest);
         module.put(globalObject, ZigString.static("spyOn"), spyOn);
 
+        const vi = JSValue.createEmptyObject(globalObject, 1);
```
Jarred-Sumner, line 713: heh

src/bun.js/bindings/JSMockFunction.cpp
```diff
@@ -295,6 +345,15 @@ class JSMockFunction : public JSC::InternalFunction {
         }
         return val;
     }
+    JSArray* getInvocationCallOrder() const
+    {
+        JSArray* val = invocationCallOrder.get();
```
Jarred-Sumner, line 351: we could avoid this by saying `if (invocationCallOrder)` which will check if the WriteBarrier has never been written to

but this is also fine

src/bun.js/bindings/JSMockFunction.cpp
```diff
@@ -911,14 +1021,9 @@ JSC_DEFINE_HOST_FUNCTION(jsMockFunctionGetMockName, (JSC::JSGlobalObject * globa
...
-    JSValue implValue = thisObject->implementation.get();
-    if (!implValue) {
-        implValue = jsUndefined();
-    }
-
-    if (auto* impl = jsDynamicCast<JSMockImplementation*>(implValue)) {
+    if (auto* impl = tryJSDynamicCast<JSMockImplementation*>(thisObject->implementation.get())) {
         if (impl->kind == JSMockImplementation::Kind::Call) {
-            JSObject* object = impl->internalFields[0].get().asCell()->getObject();
+            JSObject* object = impl->underlyingValue.get().asCell()->getObject();
```
Jarred-Sumner: I think this is fine, but it might be good to check that underlyingValue.get() is not `0`

src/bun.js/bindings/JSMockFunction.cpp
```diff
@@ -1192,35 +1217,195 @@ JSC_DEFINE_HOST_FUNCTION(jsMockFunctionMockRejectedValue, (JSC::JSGlobalObject *
...
+
+    MarkedArgumentBuffer args;
+    NakedPtr<Exception> exception;
+    JSValue returnValue = call(globalObject, callback, callData, jsUndefined(), args, exception);
+
+    if (auto promise = tryJSDynamicCast<JSC::JSPromise*>(returnValue)) {
+        auto capability = JSC::JSPromise::createNewPromiseCapability(globalObject, globalObject->promiseConstructor());
+
+        JSFunction* func = JSC::JSFunction::create(vm, globalObject, 2,
+            String(), jsMockFunctionWithImplementationCleanup, ImplementationVisibility::Public);
```
Jarred-Sumner: This function can be cached on the JSGlobalObject so that it doesn't need to be re-created each invocation

Jarred-Sumner: LazyProperty<JSGlobalObject, JSFunction>

src/bun.js/bindings/JSMockFunction.cpp
```diff
@@ -1192,35 +1217,195 @@ JSC_DEFINE_HOST_FUNCTION(jsMockFunctionMockRejectedValue, (JSC::JSGlobalObject *
...
+
+    return JSC::JSValue::encode(jsUndefined());
+}
+} // namespace Bun
+
+namespace JSC {
+
+template<unsigned passedNumberOfInternalFields>
+template<typename Visitor>
+void JSInternalFieldObjectImpl<passedNumberOfInternalFields>::visitChildrenImpl(JSCell* cell, Visitor& visitor)
```
Jarred-Sumner, line 1409: yeah we really need to copy this header from JSC so that this doesn't need to be manually copy-pasted

src/bun.js/bindings/ZigGlobalObject.h
```diff
@@ -464,6 +465,7 @@ class GlobalObject : public JSC::JSGlobalObject {
      * those callbacks will eventually never be called anymore. But it'll work the first time!
      */
     LazyProperty<JSGlobalObject, JSC::Structure> m_pendingVirtualModuleResultStructure;
+    LazyProperty<JSGlobalObject, JSC::Structure> m_mockWithImplementationCleanupDataStructure;
```
Jarred-Sumner: Can you put this on the Mock struct like the other ones for mocks?

Jarred-Sumner:
This looks great. Only thing I think should change is adding the LazyProperty for spyOn onto `Bun::JSMockModule` and same for the other structure that was added

src/bun.js/bindings/JSMockFunction.cpp
paperclover: no there is a way you can push a non-object Call (this is allowed because jest "allows" non-functions but throws when you try calling them). good catch


edit: wait no that's not what i'm referring to. it's still a good idea to check.

Jarred-Sumner:
🎉

PR #3310, opened 2023-06-14
Make `Bun.argv` the same as `process.argv`

### LINKED ISSUE #3307: Bun.argv => argv  with Chinese  is not supported

0.6.9

ubuntu x64

```js
const arg = Bun.argv
console.log(arg)
```

bun test.js 中国
```js
['bun', 'test.js', '中国']
```

### What do you see instead?

```js
[ "bun", "test.mjs", "ä¸­å›½" ]
```

### Additional information

_No response_

Jarred-Sumner:
This makes `process.argv` and `Bun.argv` return the same thing. Also makes `process.argv` faster because it becomes the value 

This fixes #3307

### DIFF
--- a/src/bun.js/api/bun.zig
+++ b/src/bun.js/api/bun.zig
@@ -440,7 +440,7 @@ pub fn getAssetPrefix(
     _: js.JSStringRef,
     _: js.ExceptionRef,
 ) js.JSValueRef {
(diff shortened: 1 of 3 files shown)


ckhatton-transreport:
Great, it will also fix #2916.

PR #3314, opened 2023-06-14
[Bun.serve] fix `Bun.serve` argument check

cirospaciari:
This adds the check into `tls` option (for sockets and Bun.serve) and also fails if no object/argument is passed into `Bun.serve()`


Jarred-Sumner:
Thank you

PR #3316, opened 2023-06-14
feat(bun/test): Impl. "toBeArray", "toBeArrayOfSize" & "toBeTypeOf"

v3nntrnx:
*toBeArray*: Determines whether a value is an array.

- [x] Implement [toBeArray](https://jest-extended.jestcommunity.dev/docs/matchers/array/#tobearray)
    - [x] Function itself
    - [x] Types
    - [x] Tests
<hr>
 
 *toBeArrayOfSize*: Determinies whether a value is an array with a specific size (length)
- [x] Implement [toBeArrayOfSize](https://jest-extended.jestcommunity.dev/docs/matchers/array/#tobearrayofsize) 
    - [x] Function itself
    - [x] Types
    - [x] Tests
    
<hr>    

*toBeTypeOf*: `expect(1).toBeTypeOf('number')`, same as `expect(typeof 1).toBe('number')`
- [x] Implement [toBeTypeOf](https://vitest.dev/api/expect.html#tobetypeof)
    - [x] Function itself
    - [x] Types
    - [x] Tests

### DIFF
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -3006,6 +3006,94 @@ pub const Expect = struct {
         return .zero;
     }
 
+    pub fn toBeArray(this: *Expect, globalThis: *JSGlobalObject, callFrame: *CallFrame) callconv(.C) JSValue {
+        defer this.postMatch(globalThis);
+
+        const thisValue = callFrame.this();
+        const value = Expect.capturedValueGetCached(thisValue) orelse {
+            globalThis.throw("Internal consistency error: the expect(value) was garbage collected but it should not have been!", .{});
+            return .zero;
+        };
+        value.ensureStillAlive();
+
+        if (this.scope.tests.items.len <= this.test_id) {
+            globalThis.throw("toBeArray() must be called in a test", .{});
+            return .zero;
+        }
+
+        active_test_expectation_counter.actual += 1;
+
+        const not = this.op.contains(.not);
+        const pass = value.jsType().isArray() != not;
+
+        if (pass) return thisValue;
+
+        var formatter = JSC.ZigConsoleClient.Formatter{ .globalThis = globalThis, .quote_strings = true };
+        const received = value.toFmt(globalThis, &formatter);
+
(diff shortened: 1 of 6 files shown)


src/bun.js/test/jest.zig
```diff
@@ -3312,6 +3400,99 @@ pub const Expect = struct {
...
+        const value = Expect.capturedValueGetCached(thisValue) orelse {
+            globalThis.throw("Internal consistency error: the expect(value) was garbage collected but it should not have been!", .{});
+            return .zero;
+        };
+        value.ensureStillAlive();
+
+        const expected = arguments[0];
+        expected.ensureStillAlive();
+
+        const expectedAsStr = expected.toString(globalThis).toSlice(globalThis, default_allocator).slice();
```
Jarred-Sumner, line 3429: we should use a ComptimeStringMap for each `typeof` name and then `expected.toBunString(globalThis).inMap(JSTypeOf) orelse { ...`

src/bun.js/test/jest.zig
```diff
@@ -3312,6 +3400,99 @@ pub const Expect = struct {
...
+        active_test_expectation_counter.actual += 1;
+
+        if (!expected.isString()) {
+            globalThis.throwInvalidArguments("toBeTypeOf() requires a string argument", .{});
+            return .zero;
+        }
+
+        if (!std.mem.eql(u8, expectedAsStr, "function") and
```
Jarred-Sumner, line 3437: We generally avoid using `std.mem.eql` because its pretty slow. It goes one character at a time.

Instead, we can use `strings.eqlComptime`, however the `ComptimeStringMap` would make this code unnecessary

Jarred-Sumner:
This looks good. There are tests and types. My comments are very nitpicky, so I'm going to merge but please address them in a follow-up PR.

v3nntrnx:
> This looks good. There are tests and types. My comments are very nitpicky, so I'm going to merge but please address them in a follow-up PR.

Got it

PR #3318, opened 2023-06-14
`expect().resolves` and `expect().rejects`

Electroid:
Adds support for:

```js

test("resolves and rejects", () => {
  expect(Promise.resolve()).resolves.toBeUndefined();
  expect(Promise.reject(new Error()).rejects.toBeInstanceOf(Error);
});
```


Electroid:
Also, you'll want to review this PR per-commit, since the first commit is just moving code from 1 file to several.

Jarred-Sumner:
@Electroid we want mark the Promise as handled. Promises internally hold a state flag. Promise.reject() goes into the rejection queue immediately, but then at the end of the next tick, when promise rejections are drained, we check if the rejected Promise was handled. If handled, it is not reported.

You'll need to read JSC's code to figure out exactly where, but it will ultimately be an `|=` on this flag:

https://github.com/oven-sh/bun/blob/669e7898b093bdfc621c168337f8099525770fc6/src/bun.js/bindings/bindings.cpp#L2451

src/bun.js/test/expect.zig
```diff
@@ -0,0 +1,3151 @@
...
+            return null;
+        }
+
+        const value = Expect.capturedValueGetCached(thisValue) orelse {
+            globalThis.throw("Internal error: the expect(value) was garbage collected but it should not have been!", .{});
+            return null;
+        };
+        value.ensureStillAlive();
+
+        if (this.resolves) |resolves| {
```
Jarred-Sumner: The main question here is if `value` will need to be wrapped in a Promise as well. 

The pattern is 

```js
await expect(fetch(url)).resolves.toBeInstanceOf(Response)
```


Sometimes, people do this:
```js
try {
   expect(123).toBe(456);
} catch (e) {

} 
```

So if they do that with a Promise and call `.then`, `.catch`, `.finally` etc, their code will throw an error.

Issue #3295, opened 2023-06-13
vue ssr support


Years-Enron:
### What version of Bun is running?

0.6.8

### What platform is your computer?

Linux 6.3.5-200.fc38.x86_64 x86_64 unknown

### What steps can reproduce the bug?

```
git clone https://github.com/vuejs/vue-hackernews-2.0.git
cd vue-hackernews-2.0
bun install
bun run server.js
```

Jarred-Sumner:
Looks like https://github.com/oven-sh/bun/pull/3294 is not fully fixed.

It's the same pattern. After importing a massive number of files, we get a crash while GC is visiting.

birkskyum:
With 0.8 I get:

```sh
➜ bun run server.js
server started at localhost:8080
164 |     rawEmitters.push(rawEmitter);
165 |     container = FsWatchFileInstances[fullPath] = {
166 |       listeners: listeners,
167 |       rawEmitters: rawEmitters,
168 |       options: options,
169 |       watcher: fs.watchFile(fullPath, options, function(curr, prev) {
                   ^
TypeError: undefined is not a function (near '...fs.watchFile...')
      at setFsWatchFileListener (/Users/admin/repos/vue-hackernews-2.0/node_modules/chokidar/lib/nodefs-handler.js:169:15)
      at /Users/admin/repos/vue-hackernews-2.0/node_modules/chokidar/lib/nodefs-handler.js:223:13
      at /Users/admin/repos/vue-hackernews-2.0/node_modules/chokidar/lib/nodefs-handler.js:255:15
      at /Users/admin/repos/vue-hackernews-2.0/node_modules/chokidar/lib/nodefs-handler.js:473:15
```

- #3812

birkskyum:
The server is starting as expected now.

PR #3320, opened 2023-06-15
Tweak CommonJS output

Jarred-Sumner:
This doesn't 100% fix the situation. We still crash when ~tens of thousands of CommonJS modules have been loaded from an ES module, an exception occurs in the top-level, and then garbage collector runs. That is the immediate cause of  #3295, however it is throwing in that case because fs.watch is not implemented and Chokidar needs it. 

Testing this is not easy. 

My current process is:
- Try to load Elysia (which uses TypeBox which does interesting things with CommonJS)
- Run the bench snippet
- import `discord.js`, `vyme`, and socket-io
- Run the code in #3295 

Currently, the latter one throws, hangs, or segfaults but the others succeed.

The better fix here I think, which I'll work on next, is splitting up how we load CommonJS into two phases:

## ESM <> CJS

This would look similar to how we do it now.

## CJS <> CJS

This would skip going through the module loader when a CJS module loads another CJS module. This would be slightly faster, but more usefully here, it should be better able to deal with all the garbage collector stuff by not entering/exiting from JavaScript quite so much. From there, we can lazily materialize CommonJS modules into ES module exports when imported/loaded from ESM

### DIFF
--- a/bench/snippets/module-exports-putter.cjs
+++ b/bench/snippets/module-exports-putter.cjs
@@ -1,0 +1,65 @@
+// This is a stress test of some internals in How Bun does the module.exports assignment.
+// If it crashes or throws then this fails
+import("./runner.mjs").then(({ bench, run }) => {
+  bench("Object.defineProperty(module, 'exports', { get() { return 42; } })", () => {
+    Object.defineProperty(module, "exports", {
+      get() {
+        return 42;
+      },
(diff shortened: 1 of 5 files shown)


Jarred-Sumner:
This is flaky, but it is better than status quo. Will follow up with a better iteration before the next release.

PR #3326, opened 2023-06-15
Implement `process.report`

en0ent1ty:
Primarily implements `process.report.getReport()` and `process.report.writeReport()`

Not all fields of the returned report object are implemented, many are stubbed, some temporarily, others may be forever, due to being too closely attached to Node or V8 specific internals, but the most frequently used fields by libraries such as `detect-libc` are implemented, which are the ones that matter. (Namely `header.glibcVersion{Runtime,Compiler}` and `sharedObjects`)

The non-function properties on process.report used for configuring reports behavior are stubbed with their default values to be read but cannot be modified as their underlying functionality isn't implemented.

Implementing `getReport()` unblocks the official `detect-libc` to work reliably in Bun, so Bun's internal `detect-libc` polyfill, which was also broken and inaccurate, is also removed by this PR.

Types and tests included.


test/js/node/process/report/report-getreport.test.js
```diff
@@ -0,0 +1,36 @@
+import helper from './helper.js';
+import assert from 'assert';
+
+{
```
Jarred-Sumner, line 4: these tests won't run 

can these be wrapped in a `test` scope?

Jarred-Sumner: oh, interesting they will run

src/bun.js/bindings/Process.cpp
```diff
@@ -10,6 +12,8 @@
 #include "ImportMetaObject.h"
 #include <sys/stat.h>
 #include "ZigConsoleClient.h"
+#include <gnu/libc-version.h>
```
Jarred-Sumner, line 15: is this linux only? if so, it should be wrapped in an #ifdef

src/bun.js/bindings/Process.cpp
```diff
@@ -451,6 +512,154 @@ JSC_DEFINE_CUSTOM_SETTER(Process_setterRelease,
...
+
+    callbackData->sharedObjs->push(globalObject, jsString(globalObject->vm(), makeAtomString(info->dlpi_name)));
+    return 0;
+};
+JSC_DEFINE_HOST_FUNCTION(Process_functionGetReport,
+    (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
+{
+    auto& vm = globalObject->vm();
+
+    std::chrono::time_point now = std::chrono::system_clock::now();
```
Jarred-Sumner, line 532: Instead of pulling in this c++ standard library function which is not currently used afaict (and which will be another symbol to load dynamically)

We can use a WebKit function for this

I believe it's `WTF::MonotonicTime::now()`

src/bun.js/bindings/Process.cpp
```diff
@@ -451,6 +512,154 @@ JSC_DEFINE_CUSTOM_SETTER(Process_setterRelease,
...
+    header->putDirect(vm, Identifier::fromString(vm, "wordSize"_s), jsNumber(64));
+    header->putDirect(vm, Identifier::fromString(vm, "arch"_s), jsString(vm, makeAtomString(PROCESS_ARCH)));
+    header->putDirect(vm, Identifier::fromString(vm, "platform"_s), jsString(vm, makeAtomString(PROCESS_PLATFORM)));
+    header->putDirect(vm, Identifier::fromString(vm, "componentVersions"_s), componentVersions);
+    header->putDirect(vm, Identifier::fromString(vm, "release"_s), JSC::JSValue::decode(Process_getterRelease(globalObject, JSValue::encode(jsUndefined()), nullptr)));
+    header->putDirect(vm, Identifier::fromString(vm, "osName"_s), jsEmptyString(vm));    // TODO: os.type()
+    header->putDirect(vm, Identifier::fromString(vm, "osRelease"_s), jsEmptyString(vm)); // TODO: os.release()
+    header->putDirect(vm, Identifier::fromString(vm, "osVersion"_s), jsEmptyString(vm)); // TODO: os.version()
+    header->putDirect(vm, Identifier::fromString(vm, "osMachine"_s), jsEmptyString(vm)); // TODO: os.machine()
+    header->putDirect(vm, Identifier::fromString(vm, "cpus"_s), JSC::constructEmptyArray(globalObject, nullptr)); // TODO: os.cpus()
```
Jarred-Sumner, line 564: these fields from os are implemented in zig currently, have a look at node_os.zig

Jarred-Sumner:
Superseeded by https://github.com/oven-sh/bun/pull/7171

PR #3330, opened 2023-06-15
Docs for DOM testing and FileSink

colinhacks:
Re: DOM testing, this just has the basics. Followup guides for different component libs coming later.

### DIFF
--- a/docs/api/websockets.md
+++ b/docs/api/websockets.md
@@ -199,7 +199,9 @@ const server = Bun.serve<{ username: string }>({
       console.log(`upgrade!`);
       const username = getUsernameFromReq(req);
       const success = server.upgrade(req, { data: { username } });
-      return success ? undefined : new Response("WebSocket upgrade error", { status: 400 });
+      return success
+        ? undefined
+        : new Response("WebSocket upgrade error", { status: 400 });
     }
 
     return new Response("Hello world");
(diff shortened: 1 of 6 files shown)


docs/api/websockets.md
```diff
@@ -173,42 +175,59 @@ Bun.serve<WebSocketData>({
...
+
+```ts#browser.js
+const socket = new WebSocket("ws://localhost:3000/chat");
+
+socket.addEventListener("message", event => {
+  console.log(event.data);
+})
+```
+
+The cookies that are currently set on the page will be sent with the WebSocket upgrade request and available on `req.headers` in the `fetch` handler. Parse these cookies to determine the identity of the connecting user and set the value of `data` accordingly.
```
Jarred-Sumner: Maybe add a section header here because its very commonly asked and people might miss this

### Identifying Users

docs/api/websockets.md
```diff
@@ -150,11 +150,13 @@ type WebSocketData = {
 // TypeScript: specify the type of `data`
 Bun.serve<WebSocketData>({
   fetch(req, server) {
+    const cookies = parseCookies(req.headers.get("Cookie"));
```
Jarred-Sumner, line 121: maybe we should do a parseCookies fn

docs/test/dom.md
```diff
@@ -0,0 +1,63 @@
...
+  import {test, expect} from 'bun:test';
+
+  test('dom test', () => {
+    document.body.innerHTML = `<button>My button</button>`;
+    const button = document.querySelector('button');
+    expect(button?.innerText).toEqual('My button');
+  });
+```
+
+Let's run this test with `bun test`:
```
Jarred-Sumner, line 54: We should do a react testing library example here too. They need to call `cleanup` in an `afterEach` and re-register GlobalRegistrator.

docs/api/websockets.md
colinhacks: 100% agree

docs/test/dom.md
colinhacks: I think we should add this only once `expect.extend` is working. Virtually all RTL tests rely on the extended RTL expect matchers (even the basic examples in their docs).

Jarred-Sumner: yeah makes sense

PR #3343, opened 2023-06-17
Set the publishToSelf option to true

lpinca:
All messages sent by the clients are expected to be received by all clients. Without the `publishToSelf` option a message is not sent back to the sender and the benchmark hangs.


Electroid (approved):
Thanks

PR #3347, opened 2023-06-17
revamp dotEnv parser

### LINKED ISSUE #411: issues with `.env` with spaces and then the lack of a trailing newline
env files sometimes chop off the last character off when there is a space in the value, but not followed by a newline. here are my findings. pay attention to the trailing new line.

---

case 1: the bug
```
VARNAME=A B
```
VARNAME is `"A "`

---

case 2: a new line fixing the bug
```
VARNAME=A B

```
VARNAME is `"A B"`

---

case 3 and 4: bug not present without a space in the value
```
VARNAME=ABC
```
```
VARNAME=ABC

```
VARNAME is `"ABC"` in both cases

### LINKED ISSUE #2823: .env can't interpret ${letter} in a value

0.6.0 latest canary of writing this

Linux 5.17.0-1021-oem x86_64 x86_64

1. make a `.env` with the contents: ```
A="a$t"
```
2. make a `a.ts` with the contents: ```
console.log(process.env.A)
```
3. `bun run a.ts` => prints a and nothing else

print a$t

### What do you see instead?

```
$ bun run a.ts
[0.07ms] ".env"
a
```

### Additional information

_No response_

### LINKED ISSUE #3042: `.env` prevents usage of `bun install` and `bun add`

0.6.3

I currently have a `.env` file with the following data:
```
PORT="7000"
USERNAME="test1"
PASSWORD="test2"
```
While the file is present I can't run `bun install` and `bun add`.

I expected to be able to run `bun install` and `bun add` and the package would be added to my project.

### What do you see instead?

Segmentation fault

### Additional information

_No response_

### LINKED ISSUE #3380: Segmentation fault - OpenAI API and Ubuntu

0.6.10

Linux 5.15.0-73-generic x86_64 x86_64

1. Create some project `bun init -y`
2. Add openai with `bun add openai`
3. Create `.env` file with credentials [0]
4. Create `index.ts` [1]
5. Run `bun index.ts`

[0]:
```env
OPENAI_API_KEY=KEY_HERE
OPENAI_ORG_ID=KEY_HERE
```

[1]:
```ts
import { Configuration, OpenAIApi } from "openai"
import { storeLatency } from "./googleFormsDatabase"

const configuration = new Configuration({
  apiKey: process.env.OPENAI_API_KEY,
  organization: process.env.OPENAI_ORG_ID,
})

const openai = new OpenAIApi(configuration)
```

File runs with no output.

### What do you see instead?

```
root@main:~/bun_crash_test# bun index.ts
[0.02ms] ".env"
Segmentation fault (core dumped)
```

### Additional information

This issue only seems to occur on my Ubuntu machine. macOS can run this same example just fine.

alexlamsl:
- fixes `strings.indexOfAny()`
- fixes OOB array access

fixes #411 
fixes #2823
fixes #3042
fixes #3380

### DIFF
--- a/src/bun.js/api/bun.zig
+++ b/src/bun.js/api/bun.zig
@@ -4935,11 +4935,11 @@ pub const EnvironmentVariables = struct {
     pub fn getEnvNames(globalObject: *JSC.JSGlobalObject, names: []ZigString) usize {
         var vm = globalObject.bunVM();
         const keys = vm.bundler.env.map.map.keys();
-        const max = @min(names.len, keys.len);
-        for (keys[0..max], 0..) |key, i| {
-            names[i] = ZigString.initUTF8(key);
+        const len = @min(names.len, keys.len);
+        for (keys[0..len], names[0..len]) |key, *name| {
+            name.* = ZigString.initUTF8(key);
         }
-        return keys.len;
+        return len;
     }
     pub fn getEnvValue(globalObject: *JSC.JSGlobalObject, name: ZigString) ?ZigString {
         var vm = globalObject.bunVM();
--- a/src/env_loader.zig
+++ b/src/env_loader.zig
@@ -17,384 +17,6 @@ const Fs = @import("./fs.zig");
 const URL = @import("./url.zig").URL;
 const Api = @import("./api/schema.zig").Api;
 const which = @import("./which.zig").which;
-const Variable = struct {
-    key: string,
-    value: string,
-    has_nested_value: bool = false,
-};
-
-// i don't expect anyone to actually use the escape line feed character
-const escLineFeed = 0x0C;
-// arbitrary character that is invalid in a real text file
-const implicitQuoteCharacter = 8;
-
-// you get 4k. I hope you don't need more than that.
-threadlocal var temporary_nested_value_buffer: [4096]u8 = undefined;
-
-pub const Lexer = struct {
-    source: *const logger.Source,
-    iter: CodepointIterator,
-    cursor: CodepointIterator.Cursor = CodepointIterator.Cursor{},
-    _codepoint: CodePoint = 0,
-    current: usize = 0,
-    last_non_space: usize = 0,
-    prev_non_space: usize = 0,
-    start: usize = 0,
-    end: usize = 0,
-    has_nested_value: bool = false,
-    has_newline_before: bool = true,
-    was_quoted: bool = false,
-
-    pub inline fn codepoint(this: *const Lexer) CodePoint {
-        return this.cursor.c;
-    }
-
-    pub inline fn step(this: *Lexer) void {
-        const ended = !this.iter.next(&this.cursor);
-        if (ended) this.cursor.c = -1;
-        this.current = this.cursor.i + @as(usize, @intFromBool(ended));
-    }
-
-    pub fn eatNestedValue(
-        _: *Lexer,
-        comptime ContextType: type,
-        ctx: *ContextType,
-        comptime Writer: type,
-        writer: Writer,
-        variable: Variable,
-        comptime getter: fn (ctx: *const ContextType, key: string) ?string,
-    ) !void {
-        var i: usize = 0;
-        var last_flush: usize = 0;
-
-        top: while (i < variable.value.len) {
-            switch (variable.value[i]) {
-                '$' => {
-                    i += 1;
-                    const start = i;
-
-                    const curly_braces_offset = @as(usize, @intFromBool(variable.value[i] == '{'));
-                    i += curly_braces_offset;
-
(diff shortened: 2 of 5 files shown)


src/env_loader.zig
```diff
@@ -909,58 +534,251 @@ pub const Loader = struct {
...
-        source: *const logger.Source,
+const Parser = struct {
+    pos: usize = 0,
+    src: string,
+
+    const whitespace_chars = "\t\x0B\x0C \xA0\n\r";
+
+    fn skipLine(this: *Parser) void {
+        var i = this.pos;
+        while (i < this.src.len) : (i += 1) {
```
Jarred-Sumner: Can this use strings.indexOfNewlineOrNonASCII?

src/env_loader.zig
```diff
@@ -909,58 +534,251 @@ pub const Loader = struct {
...
+                    return;
+                },
+                else => {},
+            }
+        }
+        this.pos = i;
+    }
+
+    fn skipWhitespaces(this: *Parser) void {
+        var i = this.pos;
```
Jarred-Sumner, line 554: Do we need to care about non-ascii whitespace characters?

Jarred-Sumner:
Nice. This mostly looks good. Can we look at a perf comparison? Maybe just paste a file with the same text copy pasted a thousand times and then paste timings of .env before & after 

Performance matters for this because it’s used everywhere in Bun. So it can’t become a bottleneck. It impacts buns install, bun run, the runtime, basically everything

src/env_loader.zig
Jarred-Sumner: (I ask because it uses SIMD)

src/env_loader.zig
alexlamsl: Do you mean `\u{2028}` etc.?

I don't believe anybody relies on that, but we can always add support for it if a use case in the wild pops up.

src/env_loader.zig
alexlamsl: fixed `strings.indexOfAny()`, employed SIMD there and converted this function to call that

src/bun.js/api/bun.zig
```diff
@@ -4935,11 +4935,11 @@ pub const EnvironmentVariables = struct {
     pub fn getEnvNames(globalObject: *JSC.JSGlobalObject, names: []ZigString) usize {
         var vm = globalObject.bunVM();
         const keys = vm.bundler.env.map.map.keys();
-        const max = @min(names.len, keys.len);
-        for (keys[0..max], 0..) |key, i| {
+        const len = @min(names.len, keys.len);
+        for (keys[0..len], 0..) |key, i| {
```
Jarred-Sumner:
```suggestion
        for (keys[0..len], names[0..len]) |key, *name| {
             name.* = ZigString.initUTF8(key);
```

src/string_immutable.zig
```diff
@@ -39,10 +39,40 @@ pub fn toUTF16Literal(comptime str: []const u8) []const u16 {
...
+
+            remaining = remaining[ascii_vector_size..];
+        }
+
+        if (comptime Environment.allow_assert) std.debug.assert(remaining.len < ascii_vector_size);
+    }
+
+    for (remaining, 0..) |c, i| {
+        if (strings.indexOfChar(str, c) != null) {
+            return @intCast(OptionalUsize, i + slice.len - remaining.len);
```
Jarred-Sumner, line 75: there's technically a pointer artihmetic thing we can do here

but pointer arithmetic is an easy way to segfault

test/cli/run/env.test.ts
```diff
@@ -194,28 +194,104 @@ describe("dotenv priority", () => {
...
+test(".env export assign", () => {
+  const dir = tempDirWithFiles("dotenv-export", {
+    ".env": "export FOO = foo\nexport = bar",
+    "index.ts": "console.log(process.env.FOO, process.env.export);",
+  });
+  const { stdout } = bunRun(`${dir}/index.ts`);
+  expect(stdout).toBe("foo bar");
+});
+
+test(".env value expansion", () => {
```
Jarred-Sumner, line 215: yay tests

src/string_immutable.zig
alexlamsl: They do that in `indexOfNewlineOrNonASCIIOrANSI()` etc. &minus; however they also do `slice.len - remaining.len` a few lines above (`first` cannot be a pointer after all), so I opt for consistency/readability here.

src/bun.js/api/bun.zig
alexlamsl: Well spotted, thanks!

Jarred-Sumner:
Thank you for this

PR #3357, opened 2023-06-19
`.randomInt()` support

### LINKED ISSUE #3355: node:crypto `randomInt` support

0.6.3

Darwin 22.5.0 arm64 arm

```ts
import { randomInt } from 'node:crypto';

const result = await randomInt(40, 120);

console.log(result);
```

_No response_

### What do you see instead?

```sh
~/Workspace/temporary/bun-random-int via ⬢ 18
➜ bun random-int.ts
1 | import { randomInt } from 'node:crypto';
2 |
3 | const result = await randomInt(40, 120);
                         ^
TypeError: randomInt is not a function. (In 'randomInt(40, 120)', 'randomInt' is undefined)
      at /Users/silas/Workspace/temporary/bun-random-int/random-int.ts:3:21
```

### Additional information

[nodejs.org/api/crypto#random-int](https://nodejs.org/api/crypto.html#cryptorandomintmin-max-callback)

lenovouser:
Closes #3355


Jarred-Sumner:
Thank you for this

I think this would be better implemented in native code, but a JS implementation is better than no implementation

It looks like the numbers must be 2^48, which is greater than max unsigned 32 bit integer. That means this will need to be a Float64Array instead of a Uint32Array. I suggest instead having a single Float64Array above the function scope, a `DataView` reading that backing ArrayBuffer, and then to get the value, zero out the last 16 bits and read the value as a Float64? 

It might end up being easier to implement in Zig/C++. That'll look something like this:

```zig
pub fn randomInt(globalObject: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSC.JSValue {
  const arguments_ = callframe.arguments(2);
  const arguments = arguments_.ptr[0..2];
  
  var min: u52 = 0;
  var max: u52 = std.mhat.maxInt(u52);
   
  if (arguments.len > 0) {
     const min_value = arguments[0];
     const max_value = arguments[1];
     if (max_value.isNumber()) max = max_value.to(u52);
     if (min_value.isNumber()) min = min_value.to(u52);
  }
  
  return JSValue.jsNumber(std.crypto.random.intRangeAtMost(u52, min, max));
}
```

And it would ideally go next to the other crypto functions in webcore.zig (which is technically misnamed)

lenovouser:
@Jarred-Sumner never worked with Zig before, but I can try to implement your suggestion and add some more tests for it, how do I run these locally by the way? Also just saw that the function also takes a callback function so I will try to implement that too.

paperclover:
heads up, node-fallbacks/* isn't used by the runtime, the actual `node:crypto` module is in `src/js/node/crypto.js`. reason for this difference is one is used for bundling and the other is used for running with bun (and features native code).

probably should keep the additions to fallback, but the test fails due to that.

also for extra context, `src/js/node/crypto.js` was originally just a bundled version of browserify crypto but many parts of it have been modified to use native calls via the `Bun` global or the secret `globalThis[Symbol.for("Bun.lazy")]` function

lenovouser:
@Jarred-Sumner @paperdave I tried getting this to work for a few hours, but I am just running into error after error produced by the Zig compiler that I am not capable enough to understand them.

The current commit I pushed using the `callframe` implementation fails with:

```
src/bun.js/base.zig:2485:20: error: expected 2 argument(s), found 3
            return @call(.auto, @field(Container, functionName), .{
                   ^~~~~
referenced by:
    Export: src/bun.js/base.zig:2497:25
    comptime_0: src/bun.js/base.zig:2705:46

make: *** [dev-obj] Error 2
```

I had another implementation working a few hours before that used `JSC.wrapWithHasContainer`. If that is also fine with you we can just use it, will push shortly.

Jarred-Sumner:
This looks good

Jarred-Sumner (approved):

simylein:
Can we merge this?

lenovouser:
I rebased and updated the generated files, should be ready to merge now 👍🏻

Issue #2883, opened 2023-05-15
TypeError: callSite.getFileName is not a function. (In 'callSite.getFileName()', 'callSite.getFileName' is undefined)


denimbasu:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

Steps to reproduce the bug:
1. Init new typescript project with `bun init`.
2. Install packages `@solana/web3.js` and `@solana/spl-token`.
3. Import the packages on index.ts
```
import * as token from "@solana/spl-token";
import * as web3 from "@solana/web3.js";
```
4. Write following code to establish connection.
```
const connection = new web3.Connection(
	web3.clusterApiUrl("devnet"),
	"confirmed"
);
```

### What is the expected behavior?

The expected behaviour is code runs without any errors. 

### What do you see instead?

```
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
4743 | /**
4744 |  * Get call site location as array.
4745 |  */
4746 | 
4747 | function callSiteLocation (callSite) {
4748 |   var file = callSite.getFileName() || '<anonymous>';
                 ^
TypeError: callSite.getFileName is not a function. (In 'callSite.getFileName()', 'callSite.getFileName' is undefined)
      at callSiteLocation (/home/denimcodes/Projects/sol/scripts/close-accounts/node_modules/@solana/web3.js/lib/index.esm.js:4748:13)
      at depd (/home/denimcodes/Projects/sol/scripts/close-accounts/node_modules/@solana/web3.js/lib/index.esm.js:4607:13)
      at /home/denimcodes/Projects/sol/scripts/close-accounts/node_modules/@solana/web3.js/lib/index.esm.js:5035:18
```

mk-pmb:
@Jarred-Sumner 
Should I have the fix in bun v1.0.12? (&larr; the version the install command got me a few minutes ago.) Because that's the error that currently blocks me from using bun for my app. Seems to be about body-parser and express.

PR #3359, opened 2023-06-19
implement more of V8's stack trace API

kvakil:
- fix source map positions for getLineNumber / getColumnNumber
- fix return value getting coerced to a string
- implement CallFrame.p.toString
- add tests for getFunction, getThis, isConstructor, isNative, toString, getLineNumber, getColumnNumber

still not implemented:
- isPromiseAll/getPromiseIndex
- getEvalOrigin
- getScriptHash
- getPosition
- getEnclosingColumnNumber/getEnclosingLineNumber
- isAsync
- accessing Error.stack should call prepareStackTrace

still broken:
- isEval: often returns false when it should return true
- isToplevel: often returns true when it should return false

Refs: https://v8.dev/docs/stack-trace-api
Refs: v8/src/objects/call-site-info.cc
Fixes: https://github.com/oven-sh/bun/issues/2883

### DIFF
--- a/src/bun.js/bindings/CallSite.cpp
+++ b/src/bun.js/bindings/CallSite.cpp
@@ -86,6 +86,42 @@ void CallSite::visitChildrenImpl(JSCell* cell, Visitor& visitor)
     visitor.append(thisCallSite->m_sourceURL);
 }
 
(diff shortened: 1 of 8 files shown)


Jarred-Sumner:
Ran this locally and `depd` works, nice.

Issue #1776, opened 2023-01-12
[install] Support for `--frozen-lockfile`


ArnaudBarre:
### What is the problem this feature would solve?

Fail on CI on inconsistent package.json/lockfile

### What is the feature you are proposing to solve the problem?

Add `--frozen-lockfile` to CLI like other package manager.

Nice to have: pnpm set it to true by default on CI: https://pnpm.io/cli/install#--frozen-lockfile

### What alternatives have you considered?

Run a git command to fail if there are git diff on the lockfile

Jarred-Sumner:
Try `--production`

Jarred-Sumner:
Though maybe adding the flag `--frozen-lockfile` would be good for compatibility

ArnaudBarre:
If `CI=true bun install` fails like `bun install --production` I think this is also ok (not the case on v0.4.0)

PR #3365, opened 2023-06-20
Add support for install with --frozen-lockfile

tiagotex:
I have broken deployments a few times because running `bun install` on CI server does not raise if the lockfile is updated, but it will raise when running with `--production` flag during deployment.

Attempted to resolve this by adding `--frozen-lockfile` flag to match the API of other package managers (https://pnpm.io/cli/install#--frozen-lockfile) as it was briefly mentioned in this issue: https://github.com/oven-sh/bun/issues/1776#issuecomment-1380973259 

In this PR I also added the flag to the schema, but I'm wondering if that is necessary/desired. Let me know if you think it should only be a CLI option. 

Thank you for providing such an excellent tool! 💚

### DIFF
--- a/completions/bun.bash
+++ b/completions/bun.bash
@@ -92,10 +92,10 @@ _bun_completions() {
     PACKAGE_OPTIONS[REMOVE_OPTIONS_LONG]="";
     PACKAGE_OPTIONS[REMOVE_OPTIONS_SHORT]="";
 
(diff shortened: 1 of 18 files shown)


Jarred-Sumner:
This looks good, thank you for doing this and I’m sorry the deploys were broken. 

It isn’t necessary to add to the schemas, confusingly the BunInstall struct isn’t actually used and that code should be deleted (along with api/demo) 

The only thing missing are tests.

tiagotex:
No worries, bun has saved us a lot of time 🍻

To clarify, do you want me to revert the changes to schemas/api/demo?

I will try to add some tests.

tiagotex:
Added tests for CI option and config.

Jarred-Sumner:
Thank you for this

PR #3367, opened 2023-06-21
enable asymmetric matchers in `expect.toEqual`, `expect.toStrictEqual`, and `expect.toHaveProperty`

dylan-conway:
also includes a few `deepEquals` bug fixes for `Dates`, `Strings`, `Sets`, and `Maps`


Jarred-Sumner:
nice

PR #3372, opened 2023-06-21
Update module resolution order docs

aplr:
If I'm reading and interpreting [these lines](https://github.com/oven-sh/bun/blob/69f558db8e0eba637febdc62c6ba752444b54401/src/options.zig#L605-L627) correct, the module resolution order [in the current docs](https://bun.sh/docs/runtime/modules#resolution) is outdated.

This change will fix the order.


Jarred-Sumner:
Yes that is correct, thank you.

PR #3373, opened 2023-06-21
⏪  ESM condition loop order

### LINKED ISSUE #3371: Module resolution issue: axios fails with "Adapter 'http' is not available in the build"

0.6.9

Darwin 22.5.0 arm64 arm

This is actually quite easy to reproduce. Run the following lines in your shell and you'll see the error:

```
mkdir bunVsAxios && cd bunVsAxios
bun init -y
bun install resend
cat <<EOT > index.ts
import resend from "resend";

const client = new resend.Resend("re_xxx");

await client.sendEmail({
  from: "hello@example.net",
  to: "world@example.net",
  subject: "Hello World!",
  text: "Hello World!",
});
EOT
bun run index.ts
```

It's a fresh bun project with [resend-node](https://github.com/resendlabs/resend-node) installed, which uses axios internally. A simple API call triggers the error.

Here's the code for `index.ts` itself in case you don't want to copy&paste the script above:

```
import resend from "resend";

const client = new resend.Resend("re_xxx");

await client.sendEmail({
  from: "hello@example.net",
  to: "world@example.net",
  subject: "Hello World!",
  text: "Hello World!",
});
```

Request is sent with axios without adapter errors.

### What do you see instead?

```
error: Adapter 'http' is not available in the build
      at /playground/bunVsAxios/node_modules/axios/dist/browser/axios.cjs:2444:12
      at /playground/bunVsAxios/node_modules/axios/dist/browser/axios.cjs:2499:18
      at request (/playground/bunVsAxios/node_modules/axios/dist/browser/axios.cjs:2859:16)
      at httpMethod (/playground/bunVsAxios/node_modules/axios/dist/browser/axios.cjs:2898:13)
      at wrap (/playground/bunVsAxios/node_modules/axios/dist/browser/axios.cjs:6:11)
      at /playground/bunVsAxios/node_modules/resend/build/src/resend.js:66:29
      at /playground/bunVsAxios/node_modules/resend/build/src/resend.js:8:8
      at /playground/bunVsAxiosx/node_modules/resend/build/src/resend.js:4:11
      at /playground/bunVsAxios/node_modules/resend/build/src/resend.js:122:39
      at /playground/bunVsAxios/node_modules/resend/build/src/resend.js:8:8
      at /playground/bunVsAxios/node_modules/resend/build/src/resend.js:4:11
      at /playground/bunVsAxios/index.ts:5:18
```

### Additional information

When installing and using axios directly, everything works as expected, and bun's module resolution logic apparently resolves to the correct axios export.

However, when installing a library which uses axios internally (i.e. [resend-node](https://github.com/resendlabs/resend-node) in my case), the request fails with the following error:

```
Adapter 'http' is not available in the build
```

That was weird to me initially, because resolution was fixed [some time ago](https://github.com/oven-sh/bun/pull/2673/files#diff-b5f3aea99b4d6bca1f40b379bc33eaf925dcb2816b9e74ad94a7dc638990917b), specifically mentioning axios.

However, I found that bun is actually resolving to the browser version of axios, as it can be seen in the stack trace:

```
/playground/bunVsAxios/node_modules/axios/dist/browser/axios.cjs:2444:12
```

However, I'd expect to see bun resolving to the node version of axios instead.

Jarred-Sumner:
Fixes #3371 (probably, still waiting to compile)


src/resolver/package_json.zig
```diff
@@ -1769,8 +1769,12 @@ pub const ESModule = struct {
...
                 const slice = object.list.slice();
                 const keys = slice.items(.key);
-                for (keys, 0..) |key, i| {
-                    if (r.conditions.contains(key)) {
+
+                // This looping order is really sublte and important.
+                // https://github.com/oven-sh/bun/issues/3371
+                // https://github.com/oven-sh/bun/pull/2673
+                for (r.conditions.keys()) |key| {
+                    if (strings.indexEqualAny(keys, key)) |i| {
```
Jarred-Sumner, line 1777: cc @Andarist for no reason other than I CC'd you the last time on a commit like this

Andarist: not a problem :p if you give me TLDR I can try to review this 😉

PR #3376, opened 2023-06-22
Add vi.spyOn and clean up some mock function binding calls

paperclover:
also made it so you can run mock-fn.test.js with vitest but tbh it doesn't really work the way i wanted it to.

did run to see that the things that matter pass (theres some subtle differences in how vi.fn handles name and so on)


src/bun.js/test/jest.zig
```diff
@@ -698,31 +698,32 @@ pub const Jest = struct {
             Expect.getConstructor(globalObject),
         );
 
-        const mock_fn = JSMockFunction__createObject(globalObject);
-        const spyOn = JSC.NewFunction(globalObject, ZigString.static("spyOn"), 2, JSMock__spyOn, false);
-        const restoreAllMocks = JSC.NewFunction(globalObject, ZigString.static("restoreAllMocks"), 2, jsFunctionResetSpies, false);
-        module.put(globalObject, ZigString.static("mock"), mock_fn);
+        const mockFn = JSC.NewFunction(globalObject, ZigString.static("fn"), 1, JSMock__jsMockFn, false);
```
Jarred-Sumner, line 701: doesn't this remove all the static functions?

Jarred-Sumner: ah

test/cli/run/cjs-repro/c.js
```diff
@@ -0,0 +1,8 @@
+import EE from "node:events";
+
+console.log([EE]);
+
+var __commonJS = (cb, mod) => function () {};
```
Jarred-Sumner: FWIW, I think this CommonJS function is not used and therefore has no impact

paperclover: oh oops i never meant to commit this file

PR #3379, opened 2023-06-22
Rewrite Bun's runtime CommonJS loader

Jarred-Sumner:
- This fixes the main crash people have been running into. ~There still is one more crash to address.~
- Builtins now have a `SourceOrigin` set to the `builtin://` URL protocol. This isn't user-facing. 
- `import.meta.primordials` is replaced with `globalThis[Symbol.for("Bun.lazy")("primordials")`. This function returns undefined if the caller source origin is not `builtin://`
- `depd` is no longer a hardcoded module
- The code that re-exports CommonJS module into the ESM registry is now only run when the CommonJS module has been imported or needs to exist in 
- `import.meta.require` is no longer used in builtin ESM modules. It is now a regular import
- `node:url` is no longer minified in the source and it uses a tweaked, vendored version of @ljharb's module. The main tweaks are using Bun's `fileURLToPath`, `pathToFileURL` functions and using `new URLSearchParams(query).toJSON()` to parse the the query string instead of a separate module

The cause of the GC crash was due to:
- Incorrect JSC::Structure use (setting inline capacity seems to cause property offsets to be wrong in non-trivial prototypes)
- Using `functionSpace` instead of `plainObjectSpace` for the prototypes

~The remaining crash repros like this:~ **fixed!**
```ts
import path from 'path';
import esbuild from 'esbuild';
console.log(path, esbuild) // mark them as used
```

This crash happens because `esbuild` is evaluated at link time. `esbuild` uses `path`. `path` is evaluated after `esbuild` since it is already in the ESM registry

Note that the following does not crash (esbuild loading before path)
```ts
import esbuild from 'esbuild';
import path from 'path';
console.log(path, esbuild) // mark them as used
```

We could make the builtins use `import.meta.require` when loaded from user code, but that would only solve this issue for the builtins (and only when statically known). I do not think it would fix it for cjs -> ESM -> cjs.

Our options here are:
- Load CommonJS & ESM modules twice
- Do some hoisting thing that probably won't work
- Rely on static analysis to pick up CommonJS export names, evaluate CommonJS at the same time as ES Modules (what Node.js does)
- Evaluate CommonJS immediately, before link time


src/bun.js/bindings/ModuleLoader.cpp
```diff
@@ -350,6 +355,105 @@ extern "C" void Bun__onFulfillAsyncModule(
...
+        case SyntheticModuleType::StringDecoder: {
+            return synthetic(generateStringDecoderSourceCode);
+        }
+        default: {
+            return jsNumber(-1);
+        }
+        }
+    }
+
+    // if (JSC::JSValue virtualModuleResult = JSValue::decode(Bun__runVirtualModule(globalObject, specifier))) {
```
Jarred-Sumner, line 422: regression: loaders don't work for CommonJS here

we can address this though, that feature is sort of already broken

src/bun.js/bindings/ZigSourceProvider.cpp
```diff
@@ -58,17 +73,17 @@ Ref<SourceProvider> SourceProvider::create(Zig::GlobalObject* globalObject, Reso
             }
         }
     }
-    auto stringImpl = Bun::toWTFString(resolvedSource.source_code);
+    auto stringImpl = Bun::toWTFString(resolvedSource.source_code).isolatedCopy();
```
Jarred-Sumner: This clone is unnecessary

src/js/builtins/BunBuiltinNames.h
```diff
@@ -265,3 +273,9 @@ class BunBuiltinNames {
 };
 
 } // namespace WebCore
+
+#ifdef ORIGINAL_ASSERT_ENABLED
```
Jarred-Sumner, line 278: This wasn't mean to be added

src/js/builtins/ImportMetaObject.ts
```diff
@@ -119,17 +118,14 @@ export function internalRequire(this: ImportMetaObject, resolved) {
...
-    return cached;
+    return cached.exports;
   }
 
   // TODO: remove this hardcoding
   if (last5 === ".json") {
     var fs = (globalThis[Symbol.for("_fs")] ||= Bun.fs());
     var exports = JSON.parse(fs.readFileSync(resolved, "utf8"));
-    $requireMap.$set(resolved, exports);
+    $requireMap.$set(resolved, { exports });
```
Jarred-Sumner: we should make these use the proper CommonJS module object

src/js/builtins/ImportMetaObject.ts
```diff
@@ -249,7 +245,7 @@ export function createRequireCache() {
 
 $sloppy;
 export function require(this: ImportMetaObject, name) {
-  var from = this?.path ?? arguments.callee.path;
+  var from = this?.path ?? arguments.callee?.path;
```
Jarred-Sumner, line 217: the ? is probably unnecessary

src/js_parser.zig
```diff
@@ -21226,14 +21226,72 @@ fn NewParser_(
...
+                    bind_args[0] = this_module;
+                    const get_require = p.newExpr(
+                        E.Dot{
+                            .name = "require",
+                            .target = this_module,
+                            .name_loc = logger.Loc.Empty,
+                        },
+                        logger.Loc.Empty,
+                    );
+
```
Jarred-Sumner, line 21240: now that the cause of the GC crash is clear we could probably move .bind() back to C++

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -59,204 +59,494 @@
...
 #include "ZigSourceProvider.h"
+#include "JavaScriptCore/FunctionPrototype.h"
+#include "CommonJSModuleRecord.h"
+#include <JavaScriptCore/JSModuleNamespaceObject.h>
+#include <JavaScriptCore/JSSourceCode.h>
+#include <JavaScriptCore/LazyPropertyInlines.h>
+
+namespace JSC {
+// this is not correct
+class JSModuleEnvironment : public JSScope {
```
dylan-conway: this can be removed?

PR #3384, opened 2023-06-23
fix tests for `expect()`

alexlamsl:


test/js/bun/test/expect.test.js
```diff
@@ -2743,7 +2836,6 @@ describe("expect()", () => {
     expect([]).toBeArrayOfSize(0);
     expect(new Array()).toBeArrayOfSize(0);
     expect([1, 2, 3, "🫓"]).toBeArrayOfSize(4);
-    expect((new Array() < string) | (number > (1, 2, 3, "🫓"))).toBeArrayOfSize(4);
```
alexlamsl, line 2746: @paperdave I couldn't figure out what this test is supposed to look like... 🤔

paperclover: copied from test-test.test.ts. lol *what* the hell is that.

oh
this is supposed to be the TS code 
```ts
expect((new Array<string|number>(1, 2, 3, "🫓"))).toBeArrayOfSize(4)
```

i pasted it in a js file and it got reformatted as such

```ts
expect((new Array(1, 2, 3, "🫓"))).toBeArrayOfSize(4)
```

alexlamsl: Ah that makes sense &minus; many thanks 👍

paperclover (approved):
cool. i do want to note there was a label function somewhere that i didn't copy paste in. we could have just used that.

also is 30 minutes really needed vs 10?

alexlamsl:
> there was a label function somewhere that i didn't copy paste in

Would you mind pointing me to the original TS source?

Keeping in sync with it would be a good idea anyway.

> also is 30 minutes really needed vs 10?

`bun-darwin-aarch64` is currently taking longer to run than previously &minus; it is not a consequence of this PR, however it is very difficult to test if this works if the tests themselves got cancelled... 😖

PR #3389, opened 2023-06-23
docs: add troubleshooting section to installation page

no-stack-dub-sack:
Hi, when I tried to install Bun on WSL today, I got:

```
error: unzip is required to install bun (see: https://github.com/oven-sh/bun#unzip-is-required)
```
Which I quickly found out this documentation does not exist at this location anymore. It looks like the main branch has already fixed this, but the [docs page it points to](https://bun.sh/docs/installation) contains no information about this error. It looks like the error is described on the [troubleshooting ](https://github.com/oven-sh/bun/blob/main/docs/troubleshooting.md) page, however, that page is commented out in nav.ts [here](https://github.com/oven-sh/bun/blob/217501e180eadd1999f30733e0f13580cd1f0abf/docs/nav.ts#L318).

This PR adds a troubleshooting section to the installation page containing the info about unzip which has been taken verbatim from the troubleshooting page. This seemed more innocuous than bringing back the troubleshooting page altogether as I assume it was commented out for a reason at some point. 

This also updates the link in install.sh to bring users directly to the troubleshooting section.

Please feel free to close this out if this is not a desired change. Also happy to tweak this if something else is needed. Thanks!

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -143,3 +143,29 @@ To write the completions to a custom location:
 $ bun completions > path-to-file      # write to file
 $ bun completions /path/to/directory  # write into directory
 ```
+
+## Troubleshooting
+
+### Unzip is required
+
+Unzip is required to install Bun on Linux. You can use one of the following commands to install `unzip`:
+
+{% codetabs %}
+
+```sh#Debian/Ubuntu/Mint
+$ sudo apt install unzip
+```
+
+```sh#RedHat/CentOS/Fedora
+$ sudo dnf install unzip
+```
+
+```sh#Arch/Manjaro
+$ sudo pacman -S unzip
+```
+
+```sh#OpenSUSE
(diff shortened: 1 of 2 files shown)


paperclover (changes requested):
can you use the codetabs component in the docs? earlier in the file is an example of it.

no-stack-dub-sack:
Sure, will take care of this tomorrow. Thanks!

no-stack-dub-sack:
@paperdave updated. let me know if this looks good or if you want me to pull in any more info from the original troubleshooting section.

colinhacks:
LGTM

colinhacks:
I don't think we should have a whole Troubleshooting section here, it gives the impression that installing Bun is hard and error-prone. I added a note at the top alongside the kernel version callout instead.

blowsie:
This url is no longer valid

PR #3390, opened 2023-06-24
Add `Bun.EventStream`

### LINKED ISSUE #2443: Support Server Sent Events

Implementing Server-Sent Events from Buns HTTP server is seemingly not possible at the moment, as returning a ReadableStream as the data of a Response object sends no response data (headers or body content) until the stream is closed.

Being able to return a ReadableStream object as the data source of a HTTP Response, having headers be instantly sent to the clients, and data be sent to the client as it is enqueued. I imagine the API would look something like this:

```ts
const msg = new TextEncoder().encode("data: hello\r\n");

export default {
  async fetch(req) {
    let timerId: Timer | undefined;

    const body = new ReadableStream({
      start(controller) {
        timerId = setInterval(() => {
          controller.enqueue(msg);
        }, 1000);
      },
      cancel() {
        if (timerId) {
          clearInterval(Number(timerId));
        }
      },
    });

    return new Response(body, {
      headers: {
        "Content-Type": "text/event-stream",
      },
    });
  },
};
```

This is based off how Deno works, a Deno example can found here: https://dash.deno.com/playground/server-sent-events

I think the Deno way best fits Bun as they both use the Web Standard Response object for HTTP Responses. And Bun also already allows returning streams in some cases, such as when using the [file API](https://bun.sh/docs/api/http#streaming-files).

However another method could be to implement something specific just for return streams to replicate Nodes `res.write()` method.

### LINKED ISSUE #2663: SSE with `serve` documentation missing

Documentation is missing

### What is the issue?

I would love to use `serve` for SSE but cannot find any documentation about it apart from [this](https://github.com/oven-sh/bun/issues/2091) issue. If `serve` is ready for handling SSE with the help of `ReadableStream` there should be documentation about it. 

### Where did you find it?

_No response_

paperclover:
This makes it easier to use bun for Server Side Events, via node:http or Bun.serve.

Fixes #2443
Relates to #2663

Here is a demo on how to use the `Bun.EventStream` API.

```ts
import { EventStream, serve } from "bun";

serve({
  port: 3000,
  fetch(req) {
    let interval;
    return new Response(
      new EventStream({
        start(stream) {
          stream.send("Hello, World!");

          interval = setInterval(() => {
            stream.send("event_name", { json_data: Bun.nanoseconds() });
          }, 1000);
        },
        cancel() {
          clearInterval(interval);
        }
      }),
    );    
  },
});
```

You can also construct `EventStream` without any options and call `.send` on it. `EventStream` extends `ReadableStream` so it may be passed as a response body. `cancel()` is called when the stream ends.


paperclover:
btw open to different names for `.send` if you think that is a poor naming choice, though i do like the two variants of it and how it automatically handles json data

Jarred-Sumner:
I'm not 100% sure this is a bug

When using Web Streams, you're supposed to use the `cancel` function for handling cancellation. Not the request's AbortSignal

Jarred-Sumner:
(it might be `"error"`, I can't quite remember off the top of my head)

paperclover:
> I'm not 100% sure this is a bug
> 
> When using Web Streams, you're supposed to use the `cancel` function for handling cancellation. Not the request's AbortSignal

the `cancel` function doesn't get called either. that's a bug. EventStream gives a cancel function as well, i just didn't show it.

~~i feel like both ways of watching for a terminated request/response should work.~~ actually i see why it shouldn't work. i kind of want to play around to make sure i understand it right though.

src/bun.js/api/server.zig
```diff
@@ -2106,7 +2106,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
 
             this.setAbortHandler();
             streamLog("is in progress, but did not return a Promise. Finalizing request context", .{});
-            this.finalize();
+            this.finalizeWithoutDeinit();
```
Jarred-Sumner: Can you check if this leaks the memory?

It's possible we should create a JSCell for RequestContext and then just rely on GC to manage the lifetime of the request instead of trying to be smart and cover all the edgecases

paperclover: ill definitely check for leaks from this. but we cannot free it at this point because onAbort can still get called.

paperdave:
> I've been doing SSE with readable stream till now.

yeah, this was to add a specific api to make SSE easier but in reality this code should just be extracted into a userspace library, or manually used with ReadableStream as you are

Issue #3383, opened 2023-06-22
pathlib-js fails with "Object is not a constructor"


vedantroy:
### What version of Bun is running?

0.6.9

### What platform is your computer?

Linux 5.16.10-051610-generic x86_64 x86_64

### What steps can reproduce the bug?

Install this: `pathlib-js`.

Run this:
```
import Path from "pathlib-js"

console.log(new Path("foo"))
```

### What is the expected behavior?

It works. (It works with node / tsx).

### What do you see instead?

1 | 
2 | console.log(new Path("foo"));
               ^
TypeError: Object is not a constructor (evaluating 'new Path("foo")')
      at /home/vedantroy/Desktop/llms/gensite/foo.ts:3:12

PR #3393, opened 2023-06-24
Runtime support for `__esModule` annotations

Jarred-Sumner:
This interprets the `__esModule` annotation at runtime:
- If `default` and the `__esModule` property is defined on `module.exports`, then the `default` ESM export
becomes the value of `module.exports.default` (like Babel), unless the enclosing package.json's `"type"` field is set to `"module"` (like esbuild)
- If  the `__esModule` annotation is not defined on `module.exports`, then we
set the `default` ESM export to `module.exports` or if `"package.json"'s `"type"` field is set to `"module"` (like Node.js)

Note that this interpretation is slightly different than esbuild and other tooling:
-  We do not disable this based on the caller's file extension. In a JS runtime, there is only one ModuleNamespaceObject per ES module
-  We ignore the value of the `__esModule` annotation. We only look for the existence of the `__esModule` property on `module.exports`. This is for performance reasons (bun loops through the `module.exports` objects at runtime). 

Relevant links:
- https://andrewbranch.github.io/interop-test
- https://stackoverflow.com/questions/50943704/whats-the-purpose-of-object-definepropertyexports-esmodule-value
- https://github.com/nodejs/node/issues/40891
- https://github.com/evanw/bundler-esm-cjs-tests
- https://github.com/evanw/esbuild/issues/1591
   

Fixes https://github.com/oven-sh/bun/issues/3383

cc @andrewbranch @colinhacks @paperdave


guybedford:
Note that "other tooling" here also includes Node.js itself. If you really thing the ecosystem should move in this direction, can we at least propose a new flag to Node.js `--es-module-interop` or something to not just diverge by default? Otherwise Node.js and Bun will always be incompatible?

guybedford:
Or at least figure out how to distinguish Node.js usage from browser style interop?

Jarred-Sumner:
> Note that "other tooling" here also includes Node.js itself. If you really thing the ecosystem should move in this direction, can we at least propose a new flag to Node.js `--es-module-interop` or something to not just diverge by default? Otherwise Node.js and Bun will always be incompatible?

It doesn’t diverge by default, Node.js-like is what this keeps unless the CommonJS module has __esModule exported. The other difference is __esModule is excluded from the ESM namespace export (we could skip doing that though)

Where I feel the least good about this PR is not paying attention to .mjs or `”type”: “module”` as opt-outs. 

I think supporting __esModule overall will make more packages work the way people expect.

Jarred-Sumner:
I have updated it to respect `"type": "module"`.

guybedford:
Completely understood supporting current bundler behaviours is needed, this sounds like it might be the right kind of compromise as Webpack worked out.

Issue #3341, opened 2023-06-17
is `puppeteer` fully working: a simple test throws a bug?


charnould:
### What version of Bun is running?

0.6.9

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

Hello,
Thanks for making `Bun`
I'm a newbie with `Puppeteer` but I try to use it to run my E2E tests.
According to `Bun` blog,  `puppeteer` [should work](https://bun.sh/blog/bun-v0.6.7#use-puppeteer-in-bun) since 0.6.7.

However, running this simple test:

```
import { beforeEach, afterEach, describe, afterAll, expect, it } from 'bun:test'
import puppeteer from 'puppeteer'

describe('A test', async () => {

  const browser = await puppeteer.launch({ defaultViewport: null, headless: false })

  const page = await browser.newPage()

  beforeEach(async () => await page.goto('http://localhost:3000'))
  afterEach(async () => await page.close())
  afterAll(async () => await browser.close())

  it('should learn puppeteer', async () => {

    console.log(page.url())                  // it works (log: http://localhost:3000/)

    const source = await page.content();
    console.log(source)                     // it works

    page
      .click('input[type="submit"]')
      .then(() => console.log('got it'))
      .catch((e) => console.log(e))        // it throws

    console.log('this line is logged')     // this line is logged
    expect(1).toBe(1)                      // this test pass
  })
})
```

...throws an error (see below)

**Nota:** My webpage (http://localhost:3000) do include somewhere: `<input type="submit" value="Login">`

Do you have any clue? And is it `Bun`-related?
Thanks

### What is the expected behavior?

No bug should be return (?)

### What do you see instead?

```
34 |     /**
35 |      * @internal
36 |      */
37 |     constructor(message) {
38 |         super(message);
39 |         this.name = this.constructor.name;
                ^
TargetCloseError: Protocol error (Runtime.callFunctionOn): Target closed
 code: "undefined"

      at /Users/charnould/GitHub/malro/node_modules/puppeteer-core/lib/esm/puppeteer/common/Errors.js:39:13
```

### Additional information

[Stack Overflow issue](https://stackoverflow.com/questions/76496160/puppeteer-why-is-there-a-protocol-error-runtime-callfunctionon-target-close)

charnould:
Thanks! 👍 
[...]

PR #3394, opened 2023-06-24
Reliability bugfix for `WebSocket`

Jarred-Sumner:
This fixes https://github.com/oven-sh/bun/issues/3341 (though I can't say with confidence it fully works, just the specific cause was due to under-estimating latin1 length)

The specific fix here is:
1) Count UTF-8 byte length for latin1 characters instead of assuming latin1 is ASCI
1) Fix a bug with counting Latin1 characters caused by the SIMD code


cirospaciari:
LGTM let's wait for the  CI and add a fail test case for #3341 to avoid regression.

PR #3396, opened 2023-06-24
[fs.watch] fix reference/deinit

cirospaciari:


src/bun.js/node/node_fs_watcher.zig
```diff
@@ -540,241 +562,202 @@ pub const FSWatcher = struct {
...
+    }
 
-        pub fn emitAbort(this: *JSObject, err: JSC.JSValue) void {
-            if (this.closed) return;
-            defer this.close(true);
-
-            err.ensureStillAlive();
+    pub fn emitAbort(this: *FSWatcher, err: JSC.JSValue) void {
+        if (this.isClosed()) return;
+        defer this.close(true);
```
Jarred-Sumner: need to keep the value alive for the duration of this function call. It could finalize while the listener is called.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -540,241 +562,202 @@ pub const FSWatcher = struct {
...
-                        _ = listener.callWithGlobalThis(
-                            globalThis,
-                            &args,
-                        );
-                    }
-                }
+    }
+    pub fn emitError(this: *FSWatcher, err: string) void {
+        if (this.isClosed()) return;
+        defer this.close(true);
```
Jarred-Sumner: need to keep the value alive for the duration of this function call. It could finalize while the listener is called.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -540,241 +562,202 @@ pub const FSWatcher = struct {
...
+        _ = this.task_count.fetchAdd(1, .Monotonic);
+        return true;
+    }
 
-        pub fn hasRef(this: *JSObject, _: *JSC.JSGlobalObject, _: *JSC.CallFrame) callconv(.C) JSC.JSValue {
-            return JSC.JSValue.jsBoolean(this.persistent);
-        }
+    pub fn hasPendingActivity(this: *FSWatcher) callconv(.C) bool {
+        @fence(.Acquire);
+        return this.task_count.load(.Acquire) > 0 or !this.closed.load(.Acquire);
```
Jarred-Sumner: This has a race condition. task_count could become greater than zero after this.closed is loaded, or vice versa

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -540,241 +562,202 @@ pub const FSWatcher = struct {
...
-                            JSC.ZigString.static(eventType).toValue(globalThis),
-                            filename,
-                        };
-                        _ = listener.callWithGlobalThis(
-                            globalThis,
-                            &args,
-                        );
+    pub fn emit(this: *FSWatcher, file_name: string, comptime eventType: string) void {
+        if (this.js_this != .zero) {
+            if (FSWatcher.listenerGetCached(this.js_this)) |listener| {
```
Jarred-Sumner: need to keep the value alive for the duration of this function call. It could finalize while the listener is called.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -540,241 +562,202 @@ pub const FSWatcher = struct {
...
-                .js_this = .zero,
-                .encoding = encoding,
-                .closed = false,
-            };
-            const instance = obj.getFSWatcher();
-
-            if (persistent) {
-                obj.poll_ref.ref(instance.ctx);
-            }
+        this.js_this = FSWatcher.toJS(this, this.globalThis);
```
Jarred-Sumner: This will need to be assigned to the stack so that the GC can see it.


```suggestion
        const js_this = FSWatcher.toJS(this, this.globalThis);
        js_this.ensureStillAlive();
        this.js_this = js_this;
```

src/bun.js/bindings/JSSink.cpp
```diff
@@ -1,6 +1,6 @@
 
 // AUTO-GENERATED FILE. DO NOT EDIT.
-// Generated by 'make generate-sink' at 2023-06-14T21:38:04.394Z
```
Jarred-Sumner, line 3: did you mean to update webkit (above)?

src/bun.js/node/node_fs_watcher.zig
cirospaciari: if user call `.close()` but we still have pending tasks I just wait and deinit after it, if is not closed and have no tasks we need to keep it alive, so will only close when user call `.close()` and we have no tasks

src/bun.js/bindings/JSSink.cpp
cirospaciari: I just run `make codegen` to update the generated classes

src/bun.js/node/node_fs_watcher.zig
Jarred-Sumner: The `finalize` function will be called (on the main thread) when `hasPendingActivity` (called on a different thread) returns false. It will be called only once per instance. Instead of having two atomics, why not have one for pending activity specifically?

Jarred-Sumner: The race condition is when task_count becomes > 0 at the same time as the developer calls `.close()` from JS

Jarred-Sumner: It is possible for that to happen because task_count is incremented on a different thread

cirospaciari: > The `finalize` function will be called (on the main thread) when `hasPendingActivity` (called on a different thread) returns false. It will be called only once per instance. Instead of having two atomics, why not have one for pending activity specifically?

We can have multiple tasks pending, I can just start task_count with 1 and close will just decrement it) so will be only one atomic, and I rename it back to ref_count. I update this.

cirospaciari: > The `finalize` function will be called (on the main thread) when `hasPendingActivity` (called on a different thread) returns false. It will be called only once per instance. Instead of having two atomics, why not have one for pending activity specifically?

We can have multiple pending tasks, if user calls close and I just decrement the ref_count more pending tasks can be created (because ref_count > 0), if I use `closed` flag I can stop the creation of new tasks and end the process sooner.

src/bun.js/node/node_fs_watcher.zig
cirospaciari: fixed into https://github.com/oven-sh/bun/pull/3396/commits/feb30ca7197f17b39c0057dc6a0c455b2964ff89

src/bun.js/node/node_fs_watcher.zig
cirospaciari: fixed into https://github.com/oven-sh/bun/pull/3396/commits/feb30ca7197f17b39c0057dc6a0c455b2964ff89

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -603,12 +608,15 @@ pub const FSWatcher = struct {
...
-        if (this.isClosed()) return;
+        if (this.closed) return;
         defer this.close(true);
 
         err.ensureStillAlive();
         if (this.js_this != .zero) {
-            if (FSWatcher.listenerGetCached(this.js_this)) |listener| {
+            const js_this = this.js_this;
+            js_this.ensureStillAlive();
+            if (FSWatcher.listenerGetCached(js_this)) |listener| {
```
Jarred-Sumner, line 618: what happens if the user calls `.close()` in the callback and there are no file changes?

cirospaciari: the user will call close and its already closed when the code call close again, and the unref will only happen after both all on main thread.

Jarred-Sumner:
lets merge once the x64 tests finish up

Jarred-Sumner:
🎉

PR #3400, opened 2023-06-25
Speculative fix for regression causing hang with `install --production`

Jarred-Sumner:


Jarred-Sumner:
No longer speculative, it fixes it

Issue #3395, opened 2023-06-24
fs.createWriteStream truncating files


armenfilipetyan:
### What version of Bun is running?

0.6.9

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

bun's implementation of `fs.createWriteStream()` does not seem to append to the file as expected.

Take for example this script:

```
import fs from 'fs';

const writer = fs.createWriteStream('testing.txt', { flags: 'a' });
writer.write('write something\n');
writer.end();
```

```
node -r esm log.js
bun run log.js
```

Run it once, both look the same. Run it twice or more and you'll see that the append flag does not work.

### What is the expected behavior?

Running with nodejs returns multiple lines, once for each run. Bun truncates and produces the output like it's only using the 'write' flag and truncates the previous content.

Hanaasagi:
The issue can also be reproduced on Linux. It looks like that the `flag` in the syscall is incorrect.

![2023-06-25_15-56](https://github.com/oven-sh/bun/assets/9482395/e2e65e7d-db4e-4ef3-94ec-40d982a273e0)

armenfilipetyan:
Thank you!

PR #3402, opened 2023-06-25
Fix the parameters of  WriteStream constructor.

Hanaasagi:
Close: #3395


Jarred-Sumner (approved):
LGTM

Thank you

test/js/node/fs/fs.test.ts
```diff
@@ -1078,6 +1078,16 @@ describe("createWriteStream", () => {
     stream.write("first line\n");
     stream.end();
 
+    await new Promise((resolve, reject) => {
```
Jarred-Sumner, line 1081: test might still fail because we commit the generated ESM modules since the CI doesn't depend on bun to run

Jarred-Sumner:
this looks right i'm going to just merge it

PR #3405, opened 2023-06-25
Support reading embedded files in compiled executables

Jarred-Sumner:
This adds support for reading files embedded in executables created via `bun build --compile`.

The following snippet works now:

```bash
bun build --compile ./hello.ts
```

```ts
// hello.ts
import index from "./file.html";
import pic from "./favicon.png";
import { readFileSync } from "node:fs";
import { file, serve } from "bun";

serve({
  fetch(req) {
    const { pathname } = new URL(req.url);
    if (pathname === "/") {
      return new Response(file(index));
    }

    if (pathname === "/favicon.png") {
      return new Response(file(pic));
    }

    return new Response("Not found", { status: 404 });
  },
});

// Works with readFileSync() as well.
console.log("index.html is", readFileSync(index).byteLength, "bytes");
```

This adds support for `Bun.file(path)` and `readFileSync(path)`. it doesn't add support for the rest of the `node:fs` functions.

Confusingly, outside of `bun build --compile`, the snippet above doesn't work without setting `.html` and `.png` to be a `file` loader:
```bash
bun --loader=.html:file --loader=.png:file ./hello.ts
```

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -3446,31 +3446,32 @@ pub const NodeFS = struct {
                         if (this.vm) |vm| {
                             if (vm.standalone_module_graph) |graph| {
                                 if (graph.find(path)) |file| {
-                                    return switch (args.encoding) {
-                                        .buffer => .{
+                                    if (args.encoding == .buffer) {
+                                        return .{
                                             .result = .{
-                                                .buffer = Buffer.fromBytes(bun.default_allocator.dupe(u8, file.contents) catch @panic("out of memory"), bun.default_allocator, .Uint8Array),
+                                                .buffer = Buffer.fromBytes(
+                                                    bun.default_allocator.dupe(u8, file.contents) catch @panic("out of memory"),
+                                                    bun.default_allocator,
+                                                    .Uint8Array,
+                                                ),
                                             },
-                                        },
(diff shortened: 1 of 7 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -3442,6 +3443,34 @@ pub const NodeFS = struct {
...
+                                                    },
+                                                };
+                                            } else {
+                                                break :brk2 .{
+                                                    .result = .{
+                                                        .null_terminated = bun.default_allocator.dupeZ(u8, file.contents) catch @panic("out of memory"),
+                                                    },
+                                                };
+                                            }
+                                        },
```
alexlamsl: would the following look better?
```zig
else => if (comptime string_type == .default)
    .{
        .result = .{
            .string = bun.default_allocator.dupe(u8, file.contents) catch @panic("out of memory"),
        },
    }
else
    .{
        .result = .{
            .null_terminated = bun.default_allocator.dupeZ(u8, file.contents) catch @panic("out of memory"),
        },
    },

Jarred-Sumner: nested switch used to have bugs, but maybe they fixed it

alexlamsl (approved):
more `bun.String` goodness 👍

PR #3419, opened 2023-06-26
Implement writev & readv

Jarred-Sumner:


cirospaciari (commented):
overall looks good, we can add more tests later, some edge cases:
What happens if we go over the max safe integer (over u52 or i53)?
we are casting usize/isize to i32 this can overflow or panics? is i32 the native limit? node/libuv uses i64 here
on linux writev dont work when fd is on append mode, should we check for this? (nodejs don't check)

Jarred-Sumner:
> overall looks good, we can add more tests later, some edge cases: What happens if we go over the max safe integer (over u52 or i53)? we are casting usize/isize to i32 this can overflow or panics? is i32 the native limit? libuv uses i64 here on linux writev dont work when fd is on append mode, should we check for this?

i32 here is for the list of buffers, not the write length. The libc API needs an i32. The max number of iovec is much smaller, and the system call will throw an error if it goes over.

I'm not sure re: append-only mode. If node doesn't do it we probably shouldn't either

cirospaciari (approved):
I see we used isize for the position I miss read this one, node uses size_t but is silent converted to i32 parameter on libuv parameter too (and earlier uses uint32_t to iterate over it), which will not be a problem.

LGTM

PR #3422, opened 2023-06-27
Fix spying on something twice

### LINKED ISSUE #3412: Calling `bun:test` `spyOn` a second time causes a crash with `SIGSEGV`

0.6.9

Linux dmx 6.3.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 21 Jun 2023 20:46:20 +0000 x86_64 GNU/Linux

1. Create a file `index.test.ts`:
```ts
import { spyOn, test } from 'bun:test';

test('runs OK', () => {
  spyOn(console, 'log');
});

test('triggers SIGSEGV', () => {
  spyOn(console, 'log');
});
```

2. Run `bun test`

Calling `spyOn` more than once does not cause a segmentation fault. Both tests should pass and `bun test` should exit cleanly.

### What do you see instead?

```
~/P/_/bun-test-sigsegv ❱ bun test
bun test v0.6.10 (76626ac5)

index.test.ts:
✓ runs OK [0.58ms]
fish: Job 1, 'bun test' terminated by signal SIGSEGV (Address boundary error)
```

### Additional information

Only causes the crash when `spyOn` is called a second time with _the same_ arguments.

Calling `mockRestore()` etc. on the spy/mock does not change the result.

Both bun stable `v0.6.9 (2ab797ff)` and current canary `v0.6.10 (76626ac5)` exhibit the same behavior.

Bun `v0.6.8 (f18423cf)` does not have this issue, which leads me to think it could be related to commit <https://github.com/oven-sh/bun/commit/dbb2416542ee391fac5a11ba56090bf946117b9d>.

Possibly related:
- https://github.com/oven-sh/bun/issues/3315
- https://github.com/oven-sh/bun/issues/3200

paperclover:
Closes #3412

### DIFF
--- a/src/bun.js/bindings/JSMockFunction.cpp
+++ b/src/bun.js/bindings/JSMockFunction.cpp
@@ -569,15 +569,19 @@ extern "C" EncodedJSValue JSMock__jsSpyOn(JSC::JSGlobalObject* lexicalGlobalObje
 
     // easymode: regular property or missing property
     if (!hasValue || slot.isValue()) {
+        JSValue value = jsUndefined();
+        if (hasValue) {
+            value = slot.getValue(globalObject, propertyKey);
+            if (jsDynamicCast<JSMockFunction*>(value)) {
+                return JSValue::encode(value);
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you

PR #3438, opened 2023-06-28
Fix .randomInt() export

lenovouser:
Missed something. Now this works:

```ts
import { randomInt } from "node:crypto";

randomInt();
```

Whereas before only this worked:

```ts
import crypto from "node:crypto";

crypto.randomInt();
```


simylein:
Thank you for fixing this 🙂

Jarred-Sumner:
Thank you

PR #3447, opened 2023-06-28
[jest] fix lifecycle hook execution order

alexlamsl:
Ref: https://jestjs.io/docs/setup-teardown#order-of-execution

### DIFF
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -493,8 +493,7 @@ pub const Jest = struct {
         var filepath = Fs.FileSystem.instance.filename_store.append([]const u8, slice) catch unreachable;
 
         var scope = runner_.getOrPutFile(filepath);
-        DescribeScope.active = scope;
-        DescribeScope.module = scope;
+        scope.push();
 
         return Bun__Jest__testModuleObject(ctx).asObjectRef();
     }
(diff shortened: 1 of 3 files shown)


src/bun.js/test/jest.zig
```diff
@@ -493,8 +493,7 @@ pub const Jest = struct {
         var filepath = Fs.FileSystem.instance.filename_store.append([]const u8, slice) catch unreachable;
 
         var scope = runner_.getOrPutFile(filepath);
-        DescribeScope.active = scope;
-        DescribeScope.module = scope;
+        scope.push();
```
Jarred-Sumner, line 496: Two questions:
- This continues to work with `--preload` adding global before{all,Each} and after{all,Each} which are cross-module, right?
- Cross-module before{all,Each} and after{all,Each} continue to not run after the file has finished executing, right?

test/js/bun/test/test-test.test.ts
```diff
@@ -540,18 +540,18 @@ beforeEach: #2
...
-afterEach: one describe scope
 afterAll: one describe scope
 beforeEach: #1
 beforeEach: #2
 beforeEach: TEST-FILE
 -- the top-level test --
+afterEach: TEST-FILE
 afterEach: #1
 afterEach: #2
-afterEach: TEST-FILE
```
alexlamsl, line 554: @Jarred-Sumner 

> This continues to work with --preload adding global before{all,Each} and after{all,Each} which are cross-module, right?

Should be covered by this existing test case &minus; I've adjusted the output which I believed to be correct now, but PTAL.

src/bun.js/test/jest.zig
alexlamsl: > Cross-module before{all,Each} and after{all,Each} continue to not run after the file has finished executing, right?

Do you mean if `bun test` is run across multiple files, `beforeAll` etc. from the first file won't get run when processing subsequent files?

If so, I can confirm that is indeed the case via stepping through with the debugger.

PR #3448, opened 2023-06-29
use main field over module for runtime

### LINKED ISSUE #3434: vuejs: render ssr

0.6.11

Darwin 22.5.0 x86_64 i386

> Official docs: https://vuejs.org/guide/scaling-up/ssr.html#rendering-an-app

```ts
// this runs in Node.js on the server.
import { createSSRApp } from "vue";
// Vue's server-rendering API is exposed under `vue/server-renderer`.
import { renderToString } from "vue/server-renderer";

const app = createSSRApp({
    data: () => ({ count: 1 }),
    template: `<button @click="count++">{{ count }}</button>`,
});

renderToString(app).then((html) => {
    console.log(html);
});

```

```html
<button>1</button>
```

### What do you see instead?

```html
<!---->
```

### Additional information

_No response_

dylan-conway:
fixes #3434


src/resolver/resolver.zig
```diff
@@ -3408,7 +3408,10 @@ pub const Resolver = struct {
                             // with this same path. The goal of this code is to avoid having
                             // both the "module" file and the "main" file in the bundle at the
                             // same time.
-                            if (kind != ast.ImportKind.require) {
+                            //
+                            // Additionally, if this is for the runtime, use the "main" field.
+                            // If it doesn't exist, the "module" field will be used.
+                            if (!r.opts.runtime and kind != ast.ImportKind.require) {
```
Jarred-Sumner: Since this is only used by the Resolver, maybe we could make this a flag inside Resolver like `enable_module_field`?

Jarred-Sumner: and then default it to true but set it to false in VirtualMachine.init

dylan-conway: good idea, updated

src/bun.js/javascript.zig
```diff
@@ -898,6 +898,7 @@ pub const VirtualMachine = struct {
 
         vm.bundler.macro_context = null;
         vm.bundler.resolver.store_fd = store_fd;
```
Jarred-Sumner, line 901: What about the initStandaloneModuleGraph function?

PR #3452, opened 2023-06-29
Fixes `node:http` and `node:stream` so `ytdl-core` works.

### LINKED ISSUE #3336: ytld-core: segmentation fault (core dumped)

0.6.1 (also tested on canary, same problem)

Linux 6.3.6-arch1-1 x86_64 unknown

Paste the following code snippet into a new file:
```js
import ytdl from 'ytdl-core';
const info = await ytdl.getInfo('https://www.youtube.com/watch?v=dQw4w9WgXcQ');
console.log(info);
```

Run your file using `bun start`

Expected behavior is information about the video provided in a JSON format printed in the console.

### What do you see instead?

A segmentation fault error.
`[1]    27760 segmentation fault (core dumped)  bun run tst.js`

### Additional information

_No response_

paperclover:
Fixes #3336 
```ts
import ytdl from "ytdl-core";
const info = await ytdl.getInfo("https://www.youtube.com/watch?v=dQw4w9WgXcQ");
console.log(info);
```

This fixes a lot of subtle things in bun:

- In JSReadableState, check if `highWaterMark` is a number. Would segfault if exists but not an integer number.
- Cleanup `node:http`
- `node:https`'s `request` and `get` will set the protocol to https properly now.
- reading a body with `request`/`get` now uses a `ReadableStream` under the hood, as the old implementation didn't seem to work properly (might be a bug relating to `Readable.fromWeb`). this approach is simpler and also would be a little faster.
- Add types for `readMany`
- Fix a crash with `createServer` and you abort a request in a specific way.
- Fix `emitClose` being false by default in some cases, causing `"close"` not be emitted.


src/js/node/http.ts
```diff
@@ -649,52 +652,43 @@ export class IncomingMessage extends Readable {
...
       }
 
-      // this can be closed by the time we get here if enough data was synchronously available
-      this.#bodyStream &&
-        this.#bodyStream.on("end", () => {
-          this.#closeBodyStream();
-        });
-    } else {
-      // this.#bodyStream.read(size);
+      (async () => {
```
Jarred-Sumner: we should move this to an internal function so that we avoid creating a new function each time read is called

Jarred-Sumner: `#consumeReader(reader) { ...` or something

src/js/node/http.ts
```diff
@@ -649,52 +652,43 @@ export class IncomingMessage extends Readable {
...
+        while (true) {
+          var { done, value, size } = await reader.readMany();
```
Jarred-Sumner: I think we might need to handle `reader.closed` or one of those states

paperclover:
could be merged, however if a request cancels halfway through, the "error" event is never emitted, and neither is end or close.

i'm not sure if there's a great way to pick up on this. abortsignal? cancel? need to check that zig actually has this done right

src/js/node/http.ts
```diff
@@ -483,8 +483,8 @@ export class Server extends EventEmitter {
       if (tls) {
         this.serverName = tls.serverName || host || "localhost";
       }
-      this.#server = Bun.serve({
-        tls,
+      this.#server = Bun.serve<any>({
+        // tls,
```
Jarred-Sumner: Won’t this break TLS?

src/js/node/http.ts
```diff
@@ -649,64 +652,51 @@ export class IncomingMessage extends Readable {
...
+  async #consumeStream(reader: ReadableStreamDefaultReader) {
+    while (true) {
+      var { done, value } = await reader.readMany();
+      if (this.#aborted) return;
+      if (done) {
+        this.push(null);
+        this.destroy();
+        break;
+      }
+      for (var v of value) {
```
Jarred-Sumner, line 641: This does ignore backpressure on the stream but it’s probably okay for now

src/js/node/http.ts
```diff
@@ -649,64 +652,51 @@ export class IncomingMessage extends Readable {
...
 
-      // this can be closed by the time we get here if enough data was synchronously available
-      this.#bodyStream &&
-        this.#bodyStream.on("end", () => {
-          this.#closeBodyStream();
-        });
-    } else {
-      // this.#bodyStream.read(size);
+      this.#consumeStream(reader);
+
```
Jarred-Sumner: Shouldn’t this be assigned before consuming? The stream can start sending chunks synchronously which push() can then cause events to be emitted which could make this re-entrant and cause unexpected behavior?

src/js/node/http.ts
```diff
@@ -649,64 +652,51 @@ export class IncomingMessage extends Readable {
...
+      this.#bodyStream = reader;
     }
   }
 
   get aborted() {
     return this.#aborted;
   }
 
-  abort() {
```
Jarred-Sumner, line 653: Is this not a public function in node?

src/js/node/http.ts
paperclover: oh i meant to ask you about this but the types dont define tls as a valid property here. i forgot to properly map it to the top level properties that do exist.

unless we support this but untyped, i didnt check.

src/js/node/http.ts
```diff
@@ -1328,8 +1320,8 @@ export class ClientRequest extends OutgoingMessage {
       } else {
         protocol = defaultAgent.protocol || "http:";
       }
-      this.#protocol = protocol;
     }
+    this.#protocol = protocol;
```
Jarred-Sumner, line 1299: Can you verify that redirects from https -> http and https -> http still work? I don’t know if we have a test for that.

src/js/node/https.ts
```diff
@@ -0,0 +1,65 @@
...
+  Server,
+  METHODS,
+  STATUS_CODES,
+  createServer,
+  ServerResponse,
+  IncomingMessage,
+  request,
+  get,
+  maxHeaderSize,
+  // validateHeaderName,
```
Jarred-Sumner: We should add these validate functions because it’ll break code to not have them and they’re small

src/js/node/http.ts
Jarred-Sumner: The types are not the source of truth. Node.js behavior is the source of truth.

src/js/node/http.ts
paperclover: its not but i should double check both cases you can get this class in node

PR #3461, opened 2023-06-30
[jest] fix lifecycle hook execution order

alexlamsl:


src/bun.js/test/jest.zig
```diff
@@ -751,17 +750,23 @@ pub const DescribeScope = struct {
...
-        new.parent = DescribeScope.active;
+        if (comptime is_bindgen) return;
+        if (new.parent) |scope| {
+            if (comptime Environment.allow_assert) {
+                std.debug.assert(DescribeScope.active != new);
+                std.debug.assert(scope == DescribeScope.active);
+            }
+        } else if (DescribeScope.active) |scope| {
+            // calling Bun.jest() within (already active) module
+            if (scope.parent != null) return;
```
Jarred-Sumner, line 761: What happens if you call `Bun.jest(import.meta.path)` from another file, use that same file repeatedly? They may not use a new instance of Bun.jest() each call. Basically, `expect` and whatnot needs to be thought of as globals even though they are not globals.

alexlamsl: Here we explicitly avoid any child `describe()` instances to be replaced by locally created "root" instance when `Bun.jest(path)` is called.

Issue #3085, opened 2023-05-26
Fileupload error using Express/Multer (busboy) 


sp90:
### What version of Bun is running?

0.6.4 (canary)

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

Made a repo to simply reproduce the error:

1. `git clone https://github.com/sp90/bun-multer-busboy`
2. `bun run index.ts`
3.  Visit `localhost:3000`
4. Upload file
5. See the error in the terminal

### What is the expected behavior?

In this given example we should get the uploaded file meta data through a console.log

### What do you see instead?

```
Example app listening on port 3000
58 |             ++this.byteCount;
59 |             const code = chunk[pos];
60 |             if (TOKEN[code] !== 1) {
61 |               if (code !== 58/* ':' */)
62 |                 return -1;
63 |               this.name += chunk.latin1Slice(start, pos);
                               ^
TypeError: Expected string
      at push (/Users/simon/Documents/dev/multer-bun/node_modules/busboy/lib/types/multipart.js:63:27)
      at /Users/simon/Documents/dev/multer-bun/node_modules/busboy/lib/types/multipart.js:394:22
      at /Users/simon/Documents/dev/multer-bun/node_modules/streamsearch/lib/sbmh.js:248:4
      at push (/Users/simon/Documents/dev/multer-bun/node_modules/streamsearch/lib/sbmh.js:104:15)
      at _write (/Users/simon/Documents/dev/multer-bun/node_modules/busboy/lib/types/multipart.js:567:4)
      at writeOrBuffer (node:stream:3614:8)
      at node:stream:3574:13
      at ondata (node:stream:2987:20)
      at node:stream:2913:8
```

sp90:
This also would close #2026

sp90:
This issue persists in version `0.6.10`

Hanaasagi:
`base64Slice` and `asciiSlice` also have the same issue. A reproducible code example is as follows:

```JavaScript
const buffer = require("node:buffer");

const buf = buffer.Buffer.from("hello world", "latin1");

console.log(buf.asciiSlice(1, 5));
console.log(buf.base64Slice(1, 5));
```

--- 

I'm attempting to resolve this problem, and it appears that making a modification like the following will be work:

```diff
diff --git a/src/js/builtins/JSBufferPrototype.ts b/src/js/builtins/JSBufferPrototype.ts
index 97b25b9b..f1c84596 100644
[...]

PR #3467, opened 2023-06-30
[WIP]Fix calling `Buffer.toString` with `(offset, length, encoding)`

Hanaasagi:
Close: https://github.com/oven-sh/bun/issues/3085

- Fix calling `Buffer.toString` with `(offset, length, encoding)`
- Fix `utf8Slice`, `latin1Slice`, ...


src/js/builtins/JSBufferPrototype.ts
```diff
@@ -427,29 +427,29 @@ export function hexWrite(this: BufferExt, text, offset, length) {
   return this.write(text, offset, length, "hex");
 }
 
-export function utf8Slice(this: BufferExt, offset, length) {
-  return this.toString(offset, length, "utf8");
+export function utf8Slice(this: BufferExt, start, end) {
+  return this.toString("utf8", start, end);
```
Hanaasagi, line 431: `start` and `end` are used instead of `offset` and `length` to maintain compatibility with nodejs.

test/js/node/buffer.test.js
```diff
@@ -2353,6 +2353,85 @@ it("Buffer.byteLength()", () => {
...
+
+it("Buffer.ucs2Slice()", () => {
+  const buf = Buffer.from("あいうえお", "ucs2");
+
+  expect(buf.ucs2Slice()).toStrictEqual("あいうえお");
+  expect(buf.ucs2Slice(2)).toStrictEqual("いうえお");
+  expect(buf.ucs2Slice(2, 6)).toStrictEqual("いう");
+});
+
+it("Buffer.base64Slice()", () => {
```
Hanaasagi, line 2419: I get a segmentfault in the base64-related tests on my local machine. I don't know what happened here.

Hanaasagi: Attempting to debug a coredump file using gdb, the arguments passed to `WTF__toBase64URLStringValue` looks like strange.



Updated:

I build the latest main branch, also find this problem. `toString('base64)` will segment fault. But if i use the canary release bun, not find this problem.

Jarred-Sumner: I vaguely remember @cirospaciari running into strange issues with .toBase64URL as well... I think we should switch to using the SIMD base64 library that Node.js also uses. Our current implementation uses a mix of WebKit's internal `atob` function and Zig's standard library. I would guess the cause of this is the Zig standard library usage, but it's hard to say without digging deeper

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1436,43 +1436,56 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunction_toStringBody(JSC::JS
...
+
+            int32_t istart = arg2.toInt32(lexicalGlobalObject);
+            if (istart < 0) {
+                throwTypeError(lexicalGlobalObject, scope, "Start must be a positive integer"_s);
+                return JSC::JSValue::encode(jsUndefined());
+            }
+            offset = static_cast<uint32_t>(istart);
+            length = (length > offset) ? (length - offset) : 0;
+    } else {
+        int32_t ioffset = arg1.toInt32(lexicalGlobalObject);
```
Jarred-Sumner, line 1472: `toInt32` can throw, we should technically check for exceptions here (we did not before)

test/js/node/buffer.test.js
```diff
@@ -2353,6 +2353,85 @@ it("Buffer.byteLength()", () => {
...
+  expect(buf.toString(100, 200, "utf8")).toStrictEqual("");
+  expect(buf.toString(100, 50, "utf8")).toStrictEqual("");
+});
+
+it("Buffer.asciiSlice())", () => {
+  const buf = Buffer.from("0123456789", "ascii");
+
+  expect(buf.asciiSlice()).toStrictEqual("0123456789");
+  expect(buf.asciiSlice(3)).toStrictEqual("3456789");
+  expect(buf.asciiSlice(3, 4)).toStrictEqual("3");
```
Jarred-Sumner, line 2384: These tests are failing

Jarred-Sumner: oh its failing because we need to commit regenerated builtins

Jarred-Sumner:
Manually tested, looks good.

Thank you for this!

test/js/node/buffer.test.js
Hanaasagi: Thank you. I had mistakenly thought that the generated code was not required for commit.

Jarred-Sumner:
Reading through node's implementation, do they use this interface? I don't see it

image

Hanaasagi:
@Jarred-Sumner 
As I mentioned in issue https://github.com/oven-sh/bun/issues/3085, Node.js does not support the `offset` and `length` parameters. If you are certain that Bun does not rely on them internally, it would be better to remove them(personal opinion).

PR #3475, opened 2023-06-30
fix "readable" event being emitted incorrectly

paperclover:
Fixes `bun --bun x prisma generate`

The issue is if you do createReadStream on a large file, the `readable` event was emitted extra times after the stream was destroyed because the readablestate destroyed value was never set. This made prisma's hash function call .digest multiple times.


Jarred-Sumner:
The child process test failures need to be fixed before this PR can move forward

paperclover:
the process.nextTick push null thing is very weird and to me feels like a workaround, but it makes it work.

PR #3488, opened 2023-07-02
Allow zero length WebSocket client & server messages

Jarred-Sumner:


src/http/websocket_http_client.zig
```diff
@@ -1201,17 +1203,16 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
                         if (data.len == 0) break;
                     },
                     .need_body => {
-                        if (receive_body_remain == 0 and data.len > 0) {
```
Jarred-Sumner, line 1204: @cirospaciari this is the part I'm least sure about

Does it look right to you?

cirospaciari: hmm when `receive_body_remain == 0 and data.len > 0` I think we should terminate and allow `data.len == 0 and receive_body_remain >= 0`, We can write some test to force this condition and check the behavior of other clients too

Jarred-Sumner: `receive_body_remain == 0` means there is no body for the payload

Jarred-Sumner: maybe we should handle that separately though

PR #3496, opened 2023-07-02
fix zero length env var values

paperclover:
changes the lazy evaluated process.env.value getter to allow returning zero length strings. fixes this code:

```ts
var dotenv = require("dotenv");
var dotenvExpand = require("dotenv-expand");

var myEnv = dotenv.config();
dotenvExpand.expand(myEnv);

console.log(process.env);
```
with this dotenv
```
VALUE=''
```


Jarred-Sumner:
LGTM, thanks

PR #3510, opened 2023-07-04
[install] fix run-time module loading

### LINKED ISSUE #3507: Auto-install fails to run twice

0.6.12

Linux 6.3.6-arch1-1 x86_64 unknown

Create a new folder, write test.ts, run `bun run test.ts` twice
Code:
```
import { say } from "cowsay";

console.log(say({ text: "grazing in the browser" }));
```

Both steps execute and display:
```
< grazing in the browser >
 ------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

### What do you see instead?

First run work as expected, second show an error

```
▶ bun run test.ts                                                    
  🔍 require-main-filename [1/1]  ________________________
< grazing in the browser >
 ------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

/git/buntest     
▶ bun run test.ts


error: Unexpected pending import in "/home/rfiszel/.bun/install/cache/cowsay@1.5.0/lib/balloon.js". To automatically install npm packages with Bun, please use an import statement instead of require() or dynamic import().
This error can also happen if dependencies import packages which are not referenced anywhere. Worst case, run `bun install` and opt-out of the node_modules folder until we come up with a better way to handle this error.
```

### Additional information

_No response_

alexlamsl:
- fix version buffer confusion
- improve workaround to handle cached modules

fixes #3507

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -1742,55 +1742,51 @@ pub const PackageManager = struct {
     pub fn enqueueDependencyToRoot(
         this: *PackageManager,
         name: []const u8,
-        version_buf: []const u8,
         version: *const Dependency.Version,
+        version_buf: []const u8,
         behavior: Dependency.Behavior,
     ) DependencyToEnqueue {
-        const str_buf = this.lockfile.buffers.string_bytes.items;
-        for (this.lockfile.buffers.dependencies.items, 0..) |dependency, dependency_id| {
-            if (!strings.eqlLong(dependency.name.slice(str_buf), name, true)) continue;
-            if (!dependency.version.eql(version, str_buf, version_buf)) continue;
-            return switch (this.lockfile.buffers.resolutions.items[dependency_id]) {
-                invalid_package_id => .{
-                    .pending = @truncate(DependencyID, dependency_id),
-                },
-                else => |resolution_id| .{
-                    .resolution = .{
-                        .resolution = this.lockfile.packages.items(.resolution)[resolution_id],
-                        .package_id = resolution_id,
-                    },
-                },
+        const dep_id = @truncate(DependencyID, brk: {
+            const str_buf = this.lockfile.buffers.string_bytes.items;
+            for (this.lockfile.buffers.dependencies.items, 0..) |dep, id| {
+                if (!strings.eqlLong(dep.name.slice(str_buf), name, true)) continue;
+                if (!dep.version.eql(version, str_buf, version_buf)) continue;
+                break :brk id;
+            }
+
+            var builder = this.lockfile.stringBuilder();
+            const dummy = Dependency{
+                .name = String.init(name, name),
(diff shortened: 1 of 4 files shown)


src/install/install.zig
```diff
@@ -1742,55 +1742,51 @@ pub const PackageManager = struct {
...
+            const dep = dummy.cloneWithDifferentBuffers(name, version_buf, @TypeOf(&builder), &builder) catch unreachable;
+            builder.clamp();
+            const index = this.lockfile.buffers.dependencies.items.len;
+            this.lockfile.buffers.dependencies.append(this.allocator, dep) catch unreachable;
+            this.lockfile.buffers.resolutions.append(this.allocator, invalid_package_id) catch unreachable;
+            if (comptime Environment.allow_assert) std.debug.assert(this.lockfile.buffers.dependencies.items.len == this.lockfile.buffers.resolutions.items.len);
+            break :brk index;
+        });
+
+        if (this.lockfile.buffers.resolutions.items[dep_id] == invalid_package_id) {
```
Jarred-Sumner, line 1777: What happens when the dependency is pending install?

src/install/install.zig
```diff
@@ -1742,55 +1742,51 @@ pub const PackageManager = struct {
...
-            failRootResolution,
-        ) catch |err| {
-            return .{ .failure = err };
-        };
+            builder.allocate() catch |err| return .{ .failure = err };
+
+            const dep = dummy.cloneWithDifferentBuffers(name, version_buf, @TypeOf(&builder), &builder) catch unreachable;
+            builder.clamp();
+            const index = this.lockfile.buffers.dependencies.items.len;
+            this.lockfile.buffers.dependencies.append(this.allocator, dep) catch unreachable;
```
Jarred-Sumner, line 1771: this always leaks the dependency right?

alexlamsl: It follows the allocation lifecycle of the `Lockfile` instance.

src/install/install.zig
alexlamsl: Even within `bun install` we would call `enqueueDependencyWithMainAndSuccessFn()` on the same dependency multiple times, e.g. various processing stages for `.npm`, so we should be safe here.

Jarred-Sumner: yes but we wouldn't append to the list of dependencies if it already existed

alexlamsl: And we don't - we look for and reference an existing dependency here:
https://github.com/oven-sh/bun/pull/3510/files#diff-2d4db027e059280840de7dfe46dda1752467e66c1c3014369ed701b026cd06fcR1751

After that search fails (which also means there can't be pending since it doesn't exist before), then we create a new one.

Jarred-Sumner:
Thank you

PR #3513, opened 2023-07-04
[tls] fix servername

cirospaciari:


Jarred-Sumner:
In the future we can use .toBunString() and then `server_name` and then store the `ZigString.Slice` which would lazily become a UTF-8 string if necessary or otherwise skip the duplication

PR #3516, opened 2023-07-04
Bunch of fixes

### LINKED ISSUE #3497: `utf-8-validate` doesn't work with `bun install` due to not running postinstall script

0.6.12

Linux 5.15.0-76-generic x86_64 x86_64

I have found this issue while trying to use the package 

https://www.npmjs.com/package/utf-8-validate

After installing the package (with bun install utf-8-validate), just copy the example code there
in a file validate.js 

``` js
'use strict';

const isValidUTF8 = require('utf-8-validate');

const buf = Buffer.from([0xf0, 0x90, 0x80, 0x80]);

console.log(isValidUTF8(buf));
// => true
```
and run it with bun validate.js

Under node  (tested with version16.20-1) it prints "true" as documented in that page.

### What do you see instead?

7 | console.log(isValidUTF8(buf));
               ^
TypeError: isValidUTF8 is not a function. (In 'isValidUTF8(buf)', 'isValidUTF8' is an instance of Object)

### Additional information

Adding a console.log(isValidUTF8)
 shows it is an anonymous function

[Function (anonymous)]

Electroid:
[`utf-8-validate`](https://www.npmjs.com/package/utf-8-validate) is a popular package to validate utf-8, mostly because `ws` [recommends](https://github.com/websockets/ws#opt-in-for-performance) that developers use it.

Node.js has since introduced the [`isUtf8`](https://nodejs.org/api/buffer.html#bufferisutf8input) function on `node:buffer`. This PR makes utf-8-validate use the built-in function, to avoid potential bugs and issues with the package


Jarred-Sumner:
This looks good but the version of utf-8-validate should not be ^range, it should be exact (as well as types). That’s why the package-json-lint test is failing

Jarred-Sumner:
Also we don’t actually need to have that package in dependencies since it will not be used

PR #3531, opened 2023-07-05
fix callbacks on release version

cirospaciari:


src/bun.js/api/bun/socket.zig
```diff
@@ -2114,11 +2113,17 @@ fn NewSocket(comptime ssl: bool) type {
...
                 .onHandshake = this.handlers.onHandshake,
                 .binary_type = this.handlers.binary_type,
+                .is_server = this.handlers.is_server,
             };
-
-            raw_handlers_ptr.* = cloned_handlers;
-            raw_handlers_ptr.is_server = this.handlers.is_server;
-            raw_handlers_ptr.protect();
+            this.handlers.onOpen = .zero;
+            this.handlers.onClose = .zero;
```
Jarred-Sumner, line 2119: is the idea here that we detach the previous handlers?

src/bun.js/api/bun/socket.zig
```diff
@@ -1972,7 +1972,11 @@ fn NewSocket(comptime ssl: bool) type {
             _: *JSC.CallFrame,
         ) callconv(.C) JSValue {
             JSC.markBinding(@src());
-            this.socket.open(!this.handlers.is_server);
+            if (comptime ssl) {
+                if (!this.detached) {
+                    this.socket.open(!this.handlers.is_server);
```
Jarred-Sumner, line 1977: I don't 100% understand the purpose of this function to begin with

src/bun.js/api/bun/socket.zig
cirospaciari: yes, this function basically detach the current socket and create 2 new ones, one raw and one TLS

src/bun.js/api/bun/socket.zig
cirospaciari: this basically starts TLS handshake, in a control manner after we deal with JS stuff internal socket attach

PR #3534, opened 2023-07-05
Fix detect-libc

Electroid:


Electroid:
Hmm, looks like it didn't work on Linux.

PR #3538, opened 2023-07-06
don't unwrap react below version `18.0.0`

### LINKED ISSUE #2951: _assign is not a function

0.6.2

Darwin 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 20:59:28 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T6000 arm64 arm

on a large React project - using the following command (with --splitting):

 bun build  --outdir out ./src/index.js --target browser --format esm --splitting  --jsx-runtime classic

Getting _assign is not a function
image

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

dylan-conway:
fixes #2951

### DIFF
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -2279,7 +2279,7 @@ pub const ParseTask = struct {
     known_target: ?options.Target = null,
     module_type: options.ModuleType = .unknown,
     ctx: *BundleV2,
-    package_version: ?string = null,
+    package_version: string = "",
 
     /// Used by generated client components
     presolved_source_indices: []const Index.Int = &.{},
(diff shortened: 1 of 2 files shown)


src/bundler/bundle_v2.zig
```diff
@@ -2279,6 +2279,7 @@ pub const ParseTask = struct {
     known_target: ?options.Target = null,
     module_type: options.ModuleType = .unknown,
     ctx: *BundleV2,
+    package_version: ?string = null,
```
Jarred-Sumner: this is 32 bytes, 16 bytes for string and due to alignment, an extra 16 bytes

we can do a regular string and just make it empty by default instead of null

src/js_parser.zig
```diff
@@ -2629,6 +2629,7 @@ pub const Parser = struct {
 
         tree_shaking: bool = false,
         bundle: bool = false,
+        package_version: ?string = null,
```
Jarred-Sumner, line 2632: same here

Issue #267, opened 2022-07-06
No brotli in zlib


bmansfie:
#### This issue has been fixed as of Bun v1.1.8 in https://bun.sh/blog/bun-v1.1.8

Original issue below

----

The zlib implementation available doesn't support brotli.   It's also significantly different than the current node.js structure of the library.

Jarred-Sumner:
Bun doesn't currently include a brotli implementation, but it really should

Kyza:
The Node API of `zlib` isn't very well structured or documented. A better version API being included in Bun with the Node version of it just for compatibility would be great.

The helper function definition [here](https://nodejs.org/docs/latest/api/zlib.html#zlibbrotlicompresssyncbuffer-options) shows the [options](https://nodejs.org/docs/latest/api/zlib.html#class-brotlioptions), but then has a `params` object inside of that which has the keys as constants from `zlib.constants`. And there are a *lot* of constants defined there.

Lutymane:
Any progress for Brotli support?

marcospassos:
This is the only thing that prevents us from using Bun :/

bmansfie:
We're in the same boat, this is the last thing lacking on the compatibility list.

jefer94:
@bmansfie do you work for Bun? Because this should be good news for me, I can't even try to make the binding by myself because event the @axios team didn't say anything that could help me to solve this issue, I'm so disappointed about how both teams has been managing this issue during two years

If you don't be a staff member of Bun, should someone tell me what's the requirements to be met? Or if you have a lib chosen to be bound to Bun now, will be great to write these stuffs in an issue, otherwise I should to deal with it manually

hker9527:
> @bmansfie do you work for Bun? Because this should be good news for me, I can't even try to make the binding by myself because event the @axios team didn't say anything that could help me to solve this issue, I'm so disappointed about how both teams has been managing this issue during two years
> 
> If you don't be a staff member of Bun, should someone tell me what's the requirements to be met? Or if you have a lib chosen to be bound to Bun now, will be great to write these stuffs in an issue, otherwise I should to deal with it manually

[This line](https://github.com/axios/axios/blob/b3be36585884ba1e237fdd0eacf55f678aefc396/lib/adapters/http.js#L38) determines whether the current environment supports brotli:
```js
const isBrotliSupported = utils.isFunction(zlib.createBrotliDecompress);
```

[...]

PR #3541, opened 2023-07-06
Add brotli support

Jarred-Sumner:
Doesn't work yet. Just the code to get it to compile, link, and the zig translate-c

- [x] Link
- [ ] Expose on the Bun global
- [ ] Expose in node:zlib
- [ ] Support in fetch() (probably won't include this initially)

Note: this uses the latest HEAD commit instead of the last recently released version because Brotli seems to not have had a release in several years but the HEAD commit has several performance improvements.


qweered:
I believe it fixes #267 ?

marc-barry:
Make sure you don't make the same [mistake Deno made](https://github.com/denoland/deno/issues/19737) which was addressed in https://github.com/denoland/deno/pull/19758. I tried to port https://github.com/marc-barry/deno-spa-proxy to Bun but it is a no-go without Brotli support.

PR #3542, opened 2023-07-06
Fixes #3129

### LINKED ISSUE #3129: Readonly on global

0.6.5

Darwin 22.4.0 arm64 arm

1. Goto - https://github.com/sp90/reproduce-bun-readonly-global
2. Clone
3. Run `bun run dist/reproduce-bun-readonly-global/server/main.js`

I have precompiled the code but there are a guide in the repo if the code recompiled should be done

Should start the server

### What do you see instead?

Screenshot_2023-05-30_at_10 22 48

### Additional information

Its an angular precompiled frontend, that is served using their universal addition, which normally works on node, and that implementation uses express under the hood

arturovt:
This marks properties as `DOMException`, `Event`, `CustomEvent`, and others, as writable properties. They are writable in other environments and may patched by third-party libraries (which are responsible for error handling, etc.). Given the following code which is valid:
```js
const originalImpl = globalThis.CustomEvent;
globalThis.CustomEvent = function(...args) {
  // Do some tracking stuff...
  return new originalImpl(...args);
};
```


Jarred-Sumner:
Thank you for this PR, but this segfaults at runtime. 

The setters need to be defined, the WriteBarrier needs to be added, the WriteBarrier needs to be visited, and the getter needs to be updated. 

This also needs a test. 

Look at WEBCORE_GENERATED_* in the same file.

Jarred-Sumner:
This looks good. Let's let the tests run. There will be some unrelated test failures

Only thing to keep in mind is that we might still not call overriden constructors for internally-generated events. For example, WebSocket `MessageEvent` will be the unmodified original

arturovt:
@Jarred-Sumner appreciate your help as a first-time contributor!

Regarding these changes I may also go quickly through the spec to see what other global properties are considered as readonly and not writable.

Just a quick note since I had to go through the `development` page to setup the local build I'm facing different errors especially on Ubuntu (20 and 22 versions).

One of the errors when running `make setup` is `#include <span> not found`, tho I have all of the deps installed as mentioned in the guide.

I've run `make setup` successfully on my Mac, but running `make dev` attempts to override files under `src/js/out/modules/node/*.js` (e.g., `assert.js`):

Снимок экрана 2023-07-07 в 17 39 28

Electroid:
Thanks for working on this @arturovt! I started the test suite.

Also, the reasoning behind the extra diffs with `.js` files is due to a bug with Bun's bundler where symbols can be non-deterministic (which we'll fix). You can ignore these diffs for now

arturovt:
@Electroid thank you. Please ping me back if anything should be updated.

PR #3543, opened 2023-07-06
refactor tls upgrade

cirospaciari:
This refactors the wrapTLS into upgradeTLS and removes the internal open call that doesn't make sense anymore. Also fixed the "connection" event behavior for this case.


Jarred-Sumner:
Thank you

PR #3551, opened 2023-07-07
Update types, partially fix `typecheck`

colinhacks:
Fixes `tsc` when run from root

Still broken when running from inside `test` but that's a bigger project


src/js/builtins/StreamInternals.ts
```diff
@@ -184,7 +184,7 @@ export function createFIFO() {
       this._capacityMask = (this._capacityMask << 1) | 1;
     }
 
-    shrinkArray() {
+    _shrinkArray() {
```
Jarred-Sumner, line 187: nice catch

test/package.json
```diff
@@ -8,6 +8,7 @@
     "@swc/core": "1.3.38",
     "@types/react": "18.0.28",
     "@types/react-dom": "18.0.11",
+    "@types/supertest": "^2.0.12",
```
Jarred-Sumner: the CI thing can be fixed by removing the ^

PR #3558, opened 2023-07-07
Fix potential coercion issue

Jarred-Sumner:


Jarred-Sumner:
> test/node_modules/pg-protocol/dist/inbound-parser.test.d.ts

Issue #3562, opened 2023-07-07
Decoding an invalid UTF-8 sequence will result in a coredump (`node:string_decoder`)


Hanaasagi:
### What version of Bun is running?

0.6.13

### What platform is your computer?

Linux 6.4.1-arch2-1 x86_64 unknown

### What steps can reproduce the bug?

Minimal reproducing code.

```JavaScript
const StringDecoder = require("string_decoder").StringDecoder;

const decoder = new StringDecoder("utf-8");
decoder.write(Buffer.from("B9", "hex"));
```

### What is the expected behavior?

No coredump occurs even if the input is invalid.

### What do you see instead?

```
[1]    1031621 segmentation fault (core dumped)  bun run t.js
```

Jarred-Sumner:
Looks like it's `string_decoder`, line 134

https://github.com/oven-sh/bun/blob/768d1f82be979d85eb0e286fd93606f15b2f155f/src/bun.js/bindings/JSStringDecoder.cpp#L126-L138

<img width="1160" alt="image" src="https://github.com/oven-sh/bun/assets/709451/ec8ef478-d6d2-4aac-b26a-387b4a5e86ea">

Hanaasagi:
yes, `--j` is overflow here. I will try to open a pull request for this. :)

PR #3563, opened 2023-07-07
fix decoding invalid UTF-8 input

Hanaasagi:
Close: #3562

- Add boundary check to `utf8CheckIncomplete` function.


Jarred-Sumner:
Thank you

PR #3579, opened 2023-07-09
Implement process.memoryUsage()

Jarred-Sumner:


Electroid:
Nice, though this is missing the overload for [`process.memoryUsage.rss()`](https://nodejs.org/api/process.html#processmemoryusagerss)

PR #3581, opened 2023-07-09
Implement nearly all of `process` object methods and properties

Jarred-Sumner:
This adds:

- [x] _debugEnd (stub)
- [x] _debugProcess (stub)
- [x] _fatalException (stub)
- [x] _getActiveHandles (always returns an empty array)
- [x] _getActiveRequests (always returns an empty array)
- [x] _linkedBinding (returns undefined)
- [x] _preload_modules (empty object)
- [x] _rawDebug (returns undefined)
- [x] _startProfilerIdleNotifier (returns undefined)
- [x] _stopProfilerIdleNotifier (returns undefined)
- [x] _tickCallback (returns undefined)
- [x] allowedNodeEnvironmentFlags (returns empty Set)
- [x] assert
- [x] debugPort (validates input, but does nothing)
- [x] features (fake object mimicking node)
- [x] getActiveResourcesInfo (returns empty array)
- [x] getegid
- [x] geteuid
- [x] getgid
- [x] getgroups
- [x] getuid
- [x] moduleLoadList (returns empty object)
- [x] openStdin (calls process.stdin.resume() and returns process.stdin)
- [x] reallyExit
- [x] setSourceMapsEnable (returns undefined)

This also makes `process` use a HashTable for the properties and removes some of the not-so-great hacks we were doing to make these values lazier. It should slightly improve startup time. 

This also makes `require("process")` directly just return globalThis.process, which makes `require("process")` 7ms faster.


sirenkovladd:
Sorry, but maybe it's better to fix `process._preload_modules` to an empty array (rather than an object)?
https://github.com/fastify/fastify-autoload/blob/v5.7.1/index.js#L13

I can't use fastify because of this

Issue #3575, opened 2023-07-08
randomUUID() returns invalid and not-v4 UUID in Bun (thus failing Zod-validation)


charnould:
### What version of Bun is running?

0.6.13

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

Run this simple script:

```
import { randomUUID } from "node:crypto";
import { version, validate } from "uuid";
import { z } from "zod";

export const uuid_schema = z.string().uuid();

const a_crypto_uuid = randomUUID();

console.log("For ", a_crypto_uuid);
console.log("version    : ", version(a_crypto_uuid));
console.log("Validation : ", validate(a_crypto_uuid));

uuid_schema.parse(a_crypto_uuid);
```

Sometimes an error is thrown like below.

<img width="721" alt="Capture d’écran 2023-07-08 à 10 24 07" src="https://github.com/oven-sh/bun/assets/7513363/a9ab8e57-ee6b-44ea-ad87-7b2dcfe33dac">

Moreover, sometimes `UUID` version seems to be (according to `UUID` lib): `v2`, `v5` whereas `randomUUID` should return `v4`. The same script never fails with `node` runtime (`node index.js`) and always returns `v4`

### What is the expected behavior?

An always succeeding script

### What do you see instead?

See image.

Jarred-Sumner:
I think the bug is right here

https://github.com/oven-sh/bun/blob/5846ad00edddd89127873bed8df911b48c5bc792/src/bun.js/uuid.zig#L17-L20

charnould:
@Jarred-Sumner I'm unfortunately not a `zig`-guy. Hope you can fix it. BTW, congrats for the road toward `1.0`!

Hanaasagi:
If no one else is currently working on it,  I think I can fix it.

PR #3583, opened 2023-07-09
fix metadata bits of uuid (`randomUUID()`)

Hanaasagi:
Close: #3575
- Fix the metadata fields in the UUID.

https://github.com/oven-sh/bun/blob/59570fe237f91dd04ce8f37779902cffa4352010/src/bun.js/webcore.zig#L690-L701

This bug occurs because the `bytes` field of the UUID is directly filled without proper initialization (Line 697).

### DIFF
--- a/src/bun.js/rare_data.zig
+++ b/src/bun.js/rare_data.zig
@@ -57,7 +57,7 @@ pub fn nextUUID(this: *RareData) UUID {
     this.entropy_cache.?.fill();
 
     const bytes = this.entropy_cache.?.get();
-    return UUID.initWith(bytes);
+    return UUID.initWith(&bytes);
 }
 
 pub fn entropySlice(this: *RareData, len: usize) []u8 {
--- a/src/bun.js/rare_data.zig
+++ b/src/bun.js/rare_data.zig
@@ -9,6 +9,7 @@ const std = @import("std");
 const BoringSSL = @import("root").bun.BoringSSL;
 const bun = @import("root").bun;
 const WebSocketClientMask = @import("../http/websocket_http_client.zig").Mask;
+const UUID = @import("./uuid.zig");
 
 boring_ssl_engine: ?*BoringSSL.ENGINE = null,
 editor_context: EditorContext = EditorContext{},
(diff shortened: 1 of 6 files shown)


src/bun.js/uuid.zig
```diff
@@ -18,6 +18,16 @@ pub fn init() UUID {
     uuid.bytes[6] = (uuid.bytes[6] & 0x0f) | 0x40;
     // Variant 1
     uuid.bytes[8] = (uuid.bytes[8] & 0x3f) | 0x80;
+
+    return uuid;
+}
+
+pub fn initWith(bytes: [16]u8) UUID {
```
Jarred-Sumner: This should be a pointer because it’s copied twice unnecessarily 
```suggestion
pub fn initWith(bytes: *const [16]u8) UUID {
```

src/bun.js/uuid.zig
```diff
@@ -18,6 +18,16 @@ pub fn init() UUID {
...
+    var uuid = UUID{ .bytes = bytes };
```
Jarred-Sumner:
```suggestion
    var uuid = UUID{ .bytes = bytes.* };
```

src/bun.js/rare_data.zig
```diff
@@ -47,13 +48,16 @@ pub fn filePolls(this: *RareData, vm: *JSC.VirtualMachine) *JSC.FilePoll.HiveArr
...
     if (this.entropy_cache == null) {
         this.entropy_cache = default_allocator.create(EntropyCache) catch unreachable;
         this.entropy_cache.?.init();
     }
 
-    return this.entropy_cache.?.get();
+    this.entropy_cache.?.fill();
+
+    const bytes = this.entropy_cache.?.get();
+    return UUID.initWith(bytes);
```
Jarred-Sumner:
```suggestion
    return UUID.initWith(&bytes);
```

Jarred-Sumner:
Thank you, please change it to a pointer otherwise LGTM

src/bun.js/uuid.zig
Hanaasagi: Fixed.

Jarred-Sumner:
Thanks

Issue #3585, opened 2023-07-09
express-session Set-Cookie not occurring with Bun


ketiddies:
### What version of Bun is running?

0.6.13

### What platform is your computer?

Darwin 22.2.0 arm64 arm

### What steps can reproduce the bug?

Run this simple express http server with `bun run index.js` and then with `node index.js`
```
const express = require('express')
const session = require('express-session')

const app = express()
app.use(
  session({
   name: `bun-cookie-test`,
    secret: 'blah-secret',
    resave: false,
    saveUninitialized: true,
    cookie: {
      secure: false,
    },
  }),
)
// Simple route to test session
app.get('/', (req, res) => {
  if (req.session) {
    if (req.session.views) {
      req.session.views++
      res.setHeader('Content-Type', 'text/html')
      res.write('<p>views: ' + req.session.views + '</p>')
      res.end()
    } else {
      req.session.views = 1
      res.end('Welcome to the session demo. Refresh the page!')
    }
  }
})
app.listen(8080)
```

### What is the expected behavior?

You see a `Set-Cookie` header in the response.

Also, you would be able to refresh the page and see the view count rendered in the html

### What do you see instead?

No `Set-Cookie` header with bun.

Yes `Set-Cookie` header with node.

ketiddies:
Did a little digging into the dependencies...

Maybe related to incomplete `node:crypto`?
https://github.com/tj/node-cookie-signature/blob/master/index.js#L19

This is the last line I see in debug logs
https://github.com/expressjs/session/blob/master/index.js#L477

I'll leave this dump here. Not sure if it's coming from bun or somewhere else
run with `DEBUG=express-session node index.js` to reproduce
```
  express-session no SID sent, generating session +2s
resume
read - n = 0
need readable false
length less than watermark true
state.constructed? false
reading, ended or constructing false
n @ fromList 0
ret @ read null
state.length while ret = null 0
read - n = undefined
need readable true
length less than watermark true
state.constructed? false
reading, ended or constructing false
n @ fromList 0
ret @ read null
state.length while ret = null 0
read - n = 0
need readable true
length less than watermark true
state.constructed? false
reading, ended or constructing false
n @ fromList 0
ret @ read null
state.length while ret = null 0
read - n = 0
need readable true
length less than watermark true
do read
readableAddChunk null
n @ fromList 0
ret @ read null
state.length while ret = null 0
read - n = undefined
read: calling endReadable if length 0 -- length, state.ended 0 true
[...]

PR #3593, opened 2023-07-10
Call `writeHead` before sending headers

Hanaasagi:
Close: #3585


test/js/node/http/node-http.test.ts
```diff
@@ -146,6 +146,29 @@ describe("node:http", () => {
...
+              return function writeHead() {
+                if (!fired) {
+                  fired = true;
+                  listener.call(this);
+                }
+                return prevWriteHead.apply(this, arguments);
+              };
+            }
+
+            function addPoweredBy() {
```
Electroid, line 161: This function is never called.

src/js/node/http.ts
```diff
@@ -1058,6 +1061,7 @@ export class ServerResponse extends Writable {
       var data = this.#firstWrite || "";
       this.#firstWrite = undefined;
       this.#finished = true;
+      this._implicitHeader();
```
Jarred-Sumner: I suspect this will have a meaningful negative performance impact because it means that the headers are flushed early

Since this is for compatibility with packages that override `writeHead`, instead we could check if `this.writeHead !== originalWriteHead` and only then call it

test/js/node/http/node-http.test.ts
Hanaasagi: Emm, I found that the locally generated esm files should also be included in the pr.

src/js/node/http.ts
Hanaasagi: Fixed, but there are a few things you need to be aware of. In Node.js, I found that each call to `write_` checks whether the `header` has been initialized and then calls the `_implicitHeader` function to generate `header`. In Bun, this logic is not relied, at least `_implicitHeader` and `writeHead` are not automatically called. I'm not sure if this implementation difference would affect other libraries.




Related code:

- https://github.com/nodejs/node/blob/41f70568a46efceb59bdb00a727b4fe8d858cb2b/lib/_http_server.js#L337-L341
- https://github.com/nodejs/node/blob/41f70568a46efceb59bdb00a727b4fe8d858cb2b/lib/_http_outgoing.js#L914-L920


If this fix is not what you expected, feel free to close this.

Jarred-Sumner:
Hey i'm going to make a couple small tweaks to your code and then merge it

PR #3596, opened 2023-07-10
[tls] General compatibility improvements

cirospaciari:
- [x] _start() added (calls .connect, not documented but is a hack way to start socket after upgrading it) 
- [x] event `secure` (not documented but used for some frameworks)
- [x] tls.checkServerIdentity and new TLSocket({ checkServerIdentity })
- [x] TLSSocket.getPeerFinished (BoringSSL does not support this for TLSv1.3+)
- [x] TLSSocket.getFinished (BoringSSL does not support this for TLSv1.3+)
- [x] TLSSocket.getProtocol
- [x] TLSSocket.getSharedSigalgs
- [x] TLSSocket.isSessionReused
- [x] TLSSocket.exportKeyingMaterial
- [x] TLSSocket.setMaxSendFragment
- [x] TLSSocket.getPeerCertificate() / TLSSocket.getPeerCertificate(true)
- [x] TLSSocket.getCertificate
- [x] TLSSocket.enableTrace (mocked)
- [x] TLSSocket.disableRenegotiation
- [x] TLSSocket.getCipher
- [x] TLSSocket.getEphemeralKeyInfo (BorringSSL does not support this, so we will return empty object if we cant get a private key)
- [x] TLSSocket.getTLSTicket (This is client, we still need to implement tickets, setTicketKeys and getTicketKeys on server)
- [x] TLSSocket.getSession 
- [x] TLSSocket.setSession and tls.connect({ session })

Will be implemented on another PR:
- [ ] ~~disconnect raw data stream after upgrading socket to TLS~~ this needs more investigation 
- [ ] TLSSocket.renegotiate
- [ ] TLSSocket.getPeerCertificate(false)
- [ ] TLSSocket.getX509Certificate
- [ ] TLSSocket,getX509PeerCertificate
- [ ] ~~TLSSocket.createSecurePair~~ (deprecated)
- [ ] event `keylog`
- [ ] event `OCSPResponse`
- [ ] event `session` (not implemented yet, needs usockets adjustments on SSL_CTX_get_ex_data and SSL_CTX_set_ex_data)
- [ ] TLSServer.setTicketsKeys
- [ ] TLSServer.getTicketKeys

Missing options:
- [ ] sigalgs
- [ ] minVersion, maxVersion
- [ ] pskCallback
- [ ] minDHSize
- [ ] highWaterMark 
- [ ] onread
- [ ] pfx 
- [ ] privateKeyIdentifier
- [ ] privateKeyEngine
- [ ] honorCipherOrder
- [ ] ecdhCurve
- [ ] crl
- [ ] clientCertEngine
- [ ] ticketKeys
- [ ] sessionIdContext
- [ ] sessionTimeout


src/bun.js/api/bun/socket.zig
```diff
@@ -56,6 +58,79 @@ const BoringSSL = bun.BoringSSL;
...
+                written += 1;
+                @memcpy(output_buf[written..][0..reason.len], reason);
+                written += reason.len;
+            }
+        }
+    }
+
+    if (written > 0) {
+        var message = output_buf[0..written];
+        zig_str = ZigString.init(std.fmt.allocPrint(bun.default_allocator, "OpenSSL {s}", .{message}) catch unreachable);
```
Jarred-Sumner, line 112: This should use `bun.String.fmt` I think

src/bun.js/api/bun/socket.zig
```diff
@@ -1880,6 +1954,104 @@ fn NewSocket(comptime ssl: bool) type {
...
+        ) callconv(.C) JSValue {
+            if (comptime ssl == false) {
+                return JSValue.jsUndefined();
+            }
+
+            if (this.detached) {
+                return JSValue.jsUndefined();
+            }
+
+            var ssl_ptr: *BoringSSL.SSL = @ptrCast(*BoringSSL.SSL, this.socket.getNativeHandle());
```
Jarred-Sumner: maybe we should extract this into a helper like `this.socket.ssl()`

src/bun.js/api/bun/socket.zig
```diff
@@ -1880,6 +1954,104 @@ fn NewSocket(comptime ssl: bool) type {
...
+            const session = BoringSSL.SSL_get_session(ssl_ptr) orelse return JSValue.jsUndefined();
+            var ticket: [*c]const u8 = undefined;
+            var length: usize = 0;
+            BoringSSL.SSL_SESSION_get0_ticket(session, @ptrCast([*c][*c]const u8, &ticket), &length);
```
Jarred-Sumner, line 1975: What is the lifetime of this value? Do we need to free it? Note that `.createBuffer` will clone the contents of the ticket, which means if this is allocated then `ticket` must be freed at the end of this scope.

src/bun.js/api/bun/socket.zig
```diff
@@ -1910,7 +2082,495 @@ fn NewSocket(comptime ssl: bool) type {
...
+            }
+            const ssl_ptr = @ptrCast(*BoringSSL.SSL, this.socket.getNativeHandle());
+
+            const nsig = BoringSSL.SSL_get_shared_sigalgs(ssl_ptr, 0, null, null, null, null, null);
+
+            const array = JSC.JSValue.createEmptyArray(globalObject, @intCast(usize, nsig));
+
+            for (0..@intCast(usize, nsig)) |i| {
+                var hash_nid: c_int = 0;
+                var sign_nid: c_int = 0;
```
Jarred-Sumner, line 2376: i haven't finished reading the PR yet but i hope there are tests for all this stuff because it will be really hard to know if we broke any of this in like 2 months

src/bun.js/api/bun/socket.zig
cirospaciari: basically from the docs`The pointer is only valid while the connection is in use.` the cloning is intentional.

src/bun.js/api/bun/x509.zig
```diff
@@ -0,0 +1,553 @@
...
+            }
+        } else {
+            const prop_name = JSC.ZigString.fromUTF8(name_slice);
+            result.put(globalObject, &prop_name, JSC.ZigString.fromUTF8(value_slice).toValueGC(globalObject));
+        }
+    }
+    return result;
+}
+
+inline fn isSafeAltName(name: [*c]const u8, length: usize, utf8: bool) bool {
```
Jarred-Sumner: why is this a c pointer and not a slice?

src/bun.js/api/bun/socket.zig
cirospaciari: I wrote tests but `getSharedSigalgs` needs `sigalgs` option so we can do more reliable tests on it.

src/bun.js/api/bun/x509.zig
```diff
@@ -0,0 +1,553 @@
...
+                        return false;
+                    }
+                }
+            },
+        }
+    }
+    return true;
+}
+
+inline fn printAltName(out: *BoringSSL.BIO, name: [*]const u8, length: usize, utf8: bool, safe_prefix: ?[*]const u8) void {
```
Jarred-Sumner: these should be slices

src/bun.js/api/bun/x509.zig
```diff
@@ -0,0 +1,553 @@
...
+}
+
+inline fn printUTF8AltName(out: *BoringSSL.BIO, name: *BoringSSL.ASN1_UTF8STRING, safe_prefix: ?[*]const u8) void {
+    printAltName(out, bun.cast([*]const u8, name.data), @intCast(usize, name.length), true, safe_prefix);
+}
+
+pub const kX509NameFlagsRFC2253WithinUtf8JSON = BoringSSL.XN_FLAG_RFC2253 & ~BoringSSL.ASN1_STRFLGS_ESC_MSB & ~BoringSSL.ASN1_STRFLGS_ESC_CTRL;
+
+const NID_ms_upn = 649;
+const NID_id_on_SmtpUTF8Mailbox = 1208;
```
Jarred-Sumner: can you include comments where all these magic numbers came from?

src/bun.js/api/bun/x509.zig
cirospaciari: basically because I translated some C++ -> Zig, i will refactor this one.

src/bun.js/api/bun/x509.zig
```diff
@@ -0,0 +1,553 @@
...
+
+// This function emulates the behavior of i2v_GENERAL_NAME in a safer and less
+// ambiguous way. "othername:" entries use the GENERAL_NAME_print format.
+fn x509PrintGeneralName(out: *BoringSSL.BIO, name: *BoringSSL.GENERAL_NAME) bool {
+    if (name.name_type == .GEN_DNS) {
+        _ = BoringSSL.BIO_write(out, "DNS:", 4);
+        // Note that the preferred name syntax (see RFCs 5280 and 1034) with
+        // wildcards is a subset of what we consider "safe", so spec-compliant DNS
+        // names will never need to be escaped.
+        printLatin1AltName(out, name.d.dNSName, null);
```
Jarred-Sumner, line 174: you sure this is always latin1? i feel like we will run into weird bugs in this code; it's safer to write as UTF-8 and then convert if different

src/bun.js/bindings/bindings.cpp
```diff
@@ -2190,6 +2190,14 @@ CPP_DECL void JSC__JSValue__putIndex(JSC__JSValue JSValue0, JSC__JSGlobalObject*
     array->putDirectIndex(arg1, arg2, value2);
 }
 
+CPP_DECL void JSC__JSValue__push(JSC__JSValue JSValue0, JSC__JSGlobalObject* arg1, JSC__JSValue JSValue3)
+{
+    JSC::JSValue value = JSC::JSValue::decode(JSValue0);
+    JSC::JSValue value2 = JSC::JSValue::decode(JSValue3);
+    JSC::JSArray* array = JSC::jsCast<JSC::JSArray*>(value);
+    array->pushInline(arg1, value2);
```
Jarred-Sumner: this should just be `push` i think

test/js/node/tls/node-tls-server.test.ts
```diff
@@ -340,6 +341,69 @@ describe("tls.createServer events", () => {
...
+            O: "Internet Widgits Pty Ltd",
+          });
+
+          expect(cert.ca).toBeTrue();
+          expect(cert.bits).toBeNumber();
+          expect(typeof cert.modulus).toBe("string");
+          expect(typeof cert.exponent).toBe("string");
+          expect(cert.pubkey).toBeInstanceOf(Buffer);
+          expect(typeof cert.valid_from).toBe("string");
+          expect(typeof cert.valid_to).toBe("string");
```
Jarred-Sumner: Can these tests assert a specific value and check its the same in Node & Bun? Would be easy for that code that encodes as either latin1 or utf-8 to have some subtle bug we never catch because we aren't checking the specific values

src/bun.js/api/bun/socket.zig
Jarred-Sumner: can you add that as a comment there so we know that later?

src/bun.js/api/bun/socket.zig
```diff
@@ -1880,6 +1954,104 @@ fn NewSocket(comptime ssl: bool) type {
...
+            const session = BoringSSL.SSL_get_session(ssl_ptr) orelse return JSValue.jsUndefined();
+            const size = BoringSSL.i2d_SSL_SESSION(session, null);
+            if (size <= 0) {
+                return JSValue.jsUndefined();
+            }
+            var buffer = bun.default_allocator.alloc(u8, @intCast(usize, size)) catch unreachable;
+            defer bun.default_allocator.free(buffer);
+            var tmp = @ptrCast([*c]u8, buffer.ptr);
+            _ = BoringSSL.i2d_SSL_SESSION(session, &tmp);
+            const result = JSC.ArrayBuffer.createBuffer(globalObject, buffer);
```
Jarred-Sumner: we could skip this extra allocation, just create the buffer immediately, call `.asArrayBuffer(globalObject).?` on the vale, then write to the array buffer instead of the clone

src/bun.js/api/bun/socket.zig
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

src/bun.js/api/bun/x509.zig
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

src/bun.js/api/bun/x509.zig
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

src/bun.js/bindings/bindings.cpp
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

test/js/node/tls/node-tls-server.test.ts
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

src/bun.js/api/bun/socket.zig
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

src/bun.js/api/bun/socket.zig
cirospaciari: fix in https://github.com/oven-sh/bun/commit/f2769584f56981d0c362c9224b4ab96d428f2b36#diff-d051905c3e784ffd4b91d0b20f99d7db1a12594b761e3651522925fed42d5bbd

src/bun.js/api/bun/socket.zig
```diff
@@ -2044,12 +2045,13 @@ fn NewSocket(comptime ssl: bool) type {
...
-            var tmp = @ptrCast([*c]u8, buffer.ptr);
-            _ = BoringSSL.i2d_SSL_SESSION(session, &tmp);
-            const result = JSC.ArrayBuffer.createBuffer(globalObject, buffer);
-            return result;
+
+            const buffer_size = @intCast(usize, size);
+            var buffer = JSValue.createBufferFromLength(globalObject, buffer_size);
+            var buffer_ptr = @ptrCast([*c]u8, buffer.asArrayBuffer(globalObject).?.ptr);
+
+            _ = BoringSSL.i2d_SSL_SESSION(session, &buffer_ptr);
```
Jarred-Sumner: should we assert

src/bun.js/api/bun/x509.zig
cirospaciari: always latin1, othername can be utf8 and we do the checking, `ASN1_IA5STRING` is always latin1 and `ASN1_UTF8STRING` is always UTF8

src/bun.js/api/bun/socket.zig
cirospaciari: hmm is better to assert this, I will update, pretty unlikely to give us a different result than the pre-flight

src/bun.js/api/bun/x509.zig
cirospaciari: fix on https://github.com/oven-sh/bun/commit/927f54269bdf489d21be4a8a6b89b448fdf48184

src/bun.js/api/bun/socket.zig
cirospaciari: fix on https://github.com/oven-sh/bun/pull/3596/commits/a0e4a35a1697dadbdf04f0606bd2b37f9b450415

Issue #3584, opened 2023-07-09
Implement `getCurves` in `node:crypto`


venkatd:
### What version of Bun is running?

0.6.13

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

Create a `curves.js` file with the following code:

```js
const { getCurves } = require("crypto");

if (getCurves().includes("prime256v1")) {
  console.log("yay");
}
```

### What is the expected behavior?

It should print yay

### What do you see instead?

```
$ bun run curves.js
1 | const { getCurves } = require("crypto");
2 | 
3 | if (getCurves().includes("prime256v1")) {
       ^
TypeError: getCurves is not a function. (In 'getCurves()', 'getCurves' is undefined)
      at /Users/venkat/code/tressel/curves.js:3:4
      at globalThis (/Users/venkat/code/tressel/curves.js:4:19)
```

### Additional information

This is one of the blockers to deploying a server in production. We use the `firebase-admin` library which makes use of node crypto functions.

dror-weiss:
I'm getting this error while trying to use a library that has [openpgp](https://github.com/openpgpjs/openpgpjs) as a dependency.

venkatd:
Wow, that was fast 🤯

PR #3598, opened 2023-07-10
feat(crypto) add getCurves

cirospaciari:
Fix https://github.com/oven-sh/bun/issues/3584


paperclover:
can you add this to the file in node-fallbacks, and then update the docs?

paperclover (approved):
nice

PR #3599, opened 2023-07-10
docs(sqlite): database prepare types reflect runtime

### LINKED ISSUE #3080: typings of `database.prepare()` are wrong

Documentation is incorrect

### What is the issue?

the jsdoc of `database.prepare()` tells me there are two arguments but `bun-types` tells me it only accepts one. The runtime however is accepting the passing of params as a second argument. 
Screenshot 2023-05-26 at 13 04 59
Screenshot 2023-05-26 at 13 05 08

### Where did you find it?

_No response_

simylein:
This should fix #3080. 
In the database prepare function I added optional param types.
Feel free to correct if you feel like there are better types to reflect the current runtime behavior.

### DIFF
--- a/packages/bun-types/sqlite.d.ts
+++ b/packages/bun-types/sqlite.d.ts
@@ -235,7 +235,7 @@ declare module "bun:sqlite" {
       ParamsType extends SQLQueryBindings | SQLQueryBindings[],
     >(
       sqlQuery: string,
-      params?: SQLQueryBindings[]
+      params?: ParamsType
     ): Statement<
       ReturnType,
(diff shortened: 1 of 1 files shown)


simylein:
Also I am not sure if the type should be `SQLQueryBindings[]` or rather `SQLQueryBindings[] | SQLQueryBindings`.
Maybe someone can clarify what the runtime actually expects/wants.

Electroid (approved):
I changed `SQLQueryBindings` to `ParamsType`, otherwise lgtm.

PR #3631, opened 2023-07-14
Update docs/types for process

colinhacks:


docs/runtime/nodejs-apis.md
```diff
@@ -510,7 +510,7 @@ The table below lists all globals implemented by Node.js and Bun's current compa
 
 - {% anchor id="node_process" %} [`process`](https://nodejs.org/api/process.html) {% /anchor %}
 - 🟡
-- Missing `process.allowedNodeEnvironmentFlags` `process.channel()` `process.connected` `process.constrainedMemory()` `process.cpuUsage()` `process.debugPort` `process.disconnect()` `process.{get|set}ActiveResourcesInfo()` `process.{get|set}{uid|gid|egid|euid|groups}()` `process.hasUncaughtExceptionCaptureCallback` `process.initGroups()` `process.kill()` `process.listenerCount` `process.memoryUsage()` `process.report` `process.resourceUsage()` `process.setSourceMapsEnabled()` `process.send()`.
+- Missing `process.allowedNodeEnvironmentFlags` `process.channel()` `process.connected` `process.constrainedMemory()` `process.disconnect()` `process.{get|set}ActiveResourcesInfo()` `process.set{uid|gid|egid|euid|groups}()` `process.hasUncaughtExceptionCaptureCallback` `process.initGroups()` `process.report` `process.resourceUsage()` `process.setSourceMapsEnabled()` `process.send()`.
```
Jarred-Sumner: `process.allowedNodeEnvironmentFlags` is stubbed as an empty object
`process.setSourceMapsEnabled()` is stubbed
`process.getActiveResourcesInfo()` is stubbed
`process.connected` is only defined in a child_process.fork() and that might also apply to process.disconnect and process.send but unsure
`process.get egid|euid|groups|` are all defined
process.set egid|euid|groups are only sometimes defined in node

packages/bun-types/globals.d.ts
```diff
@@ -75,6 +75,34 @@ interface ArrayConstructor {
   ): Promise<Array<T>>;
 }
 
+type UncaughtExceptionOrigin = "uncaughtException" | "unhandledRejection";
```
Jarred-Sumner, line 78: We have not implemented `unhandledRejection` or `uncaughtException`

packages/bun-types/globals.d.ts
```diff
@@ -377,15 +407,28 @@ interface Process {
...
   dlopen(module: { exports: any }, filename: string, flags?: number): void;
   stdin: import("stream").Duplex & { isTTY: boolean };
   stdout: import("stream").Writable & { isTTY: boolean };
   stderr: import("stream").Writable & { isTTY: boolean };
 
+  /**
+   *
+   * @deprecated This is deprecated; use the "node:assert" module instead.
+   */
+  assert(value: unknown, message?: string | Error): asserts value;
```
Jarred-Sumner, line 430: TIL `asserts`

packages/bun-types/globals.d.ts
colinhacks: This isn't actually used. I added it so later when we implement it we can just uncomment this line further down

```ts
// on(event: "uncaughtException", listener: UncaughtExceptionListener): this;
```

PR #3645, opened 2023-07-16
Implement Workers

Jarred-Sumner:
This adds `Worker` support to Bun.

```js

test("worker", done => {
  const worker = new Worker(new URL("worker-fixture.js", import.meta.url).href, {
    bun: {
//      smol: true,
    },
  });
  worker.postMessage("hello");
  worker.onerror = e => {
    console.log(e);
    worker.terminate();
  };
  worker.ref();
  worker.onmessage = e => {
    console.log(e.data);
    worker.unref();
    done();
  };
});

```

It's currently something in-between the web API and Node.js `worker_threads`. 

Like Node.js:
- `Worker` has a `ref` and `unref` function
- By default, workers will terminate when they finish executing code. To keep a worker alive indefinitely, .ref() must be called.
- `process.exit()` inside a Worker terminates the worker instead of the process

Like the web:
- `Worker` is a global that inherits from `EventTarget`
- `Worker` accepts a `name`. This becomes the posix thread name.
- `worker.postMessage` works and `self.onmessage`, `self.postMessage`, `self.addEventListener("message", cb)`, etc works

Technically neither, but sort of like the web:
- `Worker` emits a `close` event on exit
- `Worker` emits an `open` event after the initial code has been evaluated

`Worker` also supports a `bun: { smol: true}` flag which reduces memory usage at the cost of running garbage collection more often.

What's missing:
- [ ] We do not yet handle lifetimes for async data correctly yet (e.g. `fetch` response that shows up later for a terminated worker will currently not be freed). We will need to do something about task cancellation. 
- [ ] Need to write tests.
- [ ] `BroadcastChannel`, `MessagePort` and friends
- [x] console.log needs a mutex. Outside of debug builds, interleaved output is extremely possible


lgarron:
I wanted to say thanks again for landing this! With the fix from https://github.com/oven-sh/bun/issues/3669, `bun` stable now supports *highly* non-trivial code out of the box:

```shell
git clone https://github.com/cubing/cubing.js && cd cubing.js
bun install
bun run src/bin/scramble.ts -- 333
```

I'm thoroughly impressed, because I have never found a fast, direct way to run TypeScript code with web workers until now. 🤩

PR #3658, opened 2023-07-17
watch new files in directory

### LINKED ISSUE #3657: new file is not being watched in `fs.watch`

0.6.14

Linux 6.4.3-arch1-1 x86_64

**Step 1**

Create `testDir`.

```Bash
mkdir testDir
```

**Step 2**

Run this file.

```JavaScript
const fs = require("fs");

const dir = "testDir";

fs.watch(dir, (eventType, filename) => {
  console.log(eventType, filename);
});

```

**Step 3**

Execute following commands in other console.

```Bash
echo hello >> testDir/test
echo hello >> testDir/test
echo hello >> testDir/test
```

Received multiple `change` events.

```
rename test
change test
change test
change test
```

### What do you see instead?

Only received one event.

```
rename test
```

### Additional information

_No response_

Hanaasagi:
Close #3657.

- Add newly created files in the directory to the watch list.


src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +404,32 @@ pub const FSWatcher = struct {
...
+                            };
+
+                            // Add new files or directories to the watch list.
+                            if (!is_watched) {
+                                this.file_paths.push(bun.default_allocator, full_path) catch unreachable;
+                                const fs_info = fdFromAbsolutePathZ(full_path) catch null;
+                                if (fs_info) |info| {
+                                    if (info.is_file) {
+                                        this.default_watcher.?.addFileAssumeLocked(info.fd, full_path, file_hash, options.Loader.file, 0, null, false) catch unreachable;
+                                        current_task.append(relative_path, .change, .destroy);
```
Hanaasagi: Add this because when a new file is created, two events (`rename`, `change`) should be triggered.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +404,32 @@ pub const FSWatcher = struct {
 
                             current_task.append(relative_path, event_type, .destroy);
 
+                            const full_path = bun.default_allocator.dupeZ(u8, path_slice) catch unreachable;
```
Jarred-Sumner: The file will often already be watched. That means this allocation will usually be unnecessary which hurts performance. Instead, this should be copied into a temporary buffer of size `bun.MAX_PATH_BYTES + 1` and only cloned if needed. That will also fix the memory leak this code creates where `full_path` is not freed

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +404,32 @@ pub const FSWatcher = struct {
...
+                            const is_watched = blk: {
+                                for (this.file_paths.slice()) |path| {
```
Jarred-Sumner: This will get very expensive fast. It sounds like we should have a hash table here.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +404,32 @@ pub const FSWatcher = struct {
...
+                                        break :blk true;
+                                    }
+                                }
+                                break :blk false;
+                            };
+
+                            // Add new files or directories to the watch list.
+                            if (!is_watched) {
+                                this.file_paths.push(bun.default_allocator, full_path) catch unreachable;
+                                const fs_info = fdFromAbsolutePathZ(full_path) catch null;
```
Jarred-Sumner, line 458: We definitely don't have an existing file descriptor for this file path, right? This is a very expensive system call so we should be sure to only do it if definitely necessary

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +404,32 @@ pub const FSWatcher = struct {
...
+                                if (fs_info) |info| {
+                                    if (info.is_file) {
+                                        this.default_watcher.?.addFileAssumeLocked(info.fd, full_path, file_hash, options.Loader.file, 0, null, false) catch unreachable;
+                                        current_task.append(relative_path, .change, .destroy);
+                                    } else {
+                                        if (this.recursive) {
```
Jarred-Sumner, line 464: What happens if it's a symlink? Will this attempt to watch symlinks? It should specifically check that the file is a directory and only run this code in that case (if fs_info doesn't track that it should be added)

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +404,32 @@ pub const FSWatcher = struct {
 
                             current_task.append(relative_path, event_type, .destroy);
 
+                            const full_path = bun.default_allocator.dupeZ(u8, path_slice) catch unreachable;
+                            const is_watched = blk: {
+                                for (this.file_paths.slice()) |path| {
+                                    if (std.mem.eql(u8, path, full_path)) {
```
Jarred-Sumner: When comparing if two strings are eql in Bun, use `strings.eql` (we might switch it later to use eqlLong)

src/bun.js/node/node_fs_watcher.zig
Jarred-Sumner: Does the lifetime of `relative_path` work for this?

Hanaasagi: `relative_path` is created by `bun.default_allocator.dupe`.

Updated:

Fixed. It seems that a new `relative_path` needs to be cloned here because the prev one might get freed.

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: If it is a symbolic link, `is_file` will be `true`, and this branch will not be executed.

https://github.com/oven-sh/bun/blob/13b5d9d4de4e9fe897d373a6ed7fdf2c9884c71f/src/bun.js/node/node_fs_watcher.zig#L793-L821

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: Okay, I will fix it.

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: This branch is executed when the event's kind is `.directory`. I think this occurs only when files are created, renamed, or deleted in a directory. Perhaps I misunderstood, but in most cases, this branch is only executed once.

https://github.com/oven-sh/bun/blob/13b5d9d4de4e9fe897d373a6ed7fdf2c9884c71f/src/bun.js/node/node_fs_watcher.zig#L366

Hanaasagi: I tried resuing `_on_file_update_path_buf`. However, when a new file is created, it needs a heap-allocated `file_path`. I deferred this operation to `addFile` by using the `copy_file_path` parameter.

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: Should I change the data structure to ArrayHashMap?

https://ziglang.org/documentation/master/std/#A;std:ArrayHashMap

test/js/node/watch/fs.watch.test.ts
```diff
@@ -97,6 +102,66 @@ describe("fs.watch", () => {
...
+
+    const interval = repeat(() => {
+      // Use this event to test if the watcher has been started.
+      if (!ready) {
+        fs.writeFileSync(path.join(root, "new-file.txt"), "hello");
+        return;
+      }
+      const fd = fs.openSync(path.join(root, "new-file2.txt"), "a");
+      fs.writeSync(fd, "hello");
+      fs.writeSync(fd, "world");
```
Hanaasagi: I think I know why this test sometimes fails. Counting the number of events itself is unreliable, sometimes events get merged.

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: Should I change the data structure to ArrayHashMap?

https://ziglang.org/documentation/master/std/#A;std:ArrayHashMap

cirospaciari: We should probably use `std.hash.Wyhash` and `std.HashMap` take a look on https://github.com/oven-sh/bun/blob/main/src/bun.zig#L826-L828

test/js/node/watch/fs.watch.test.ts
cirospaciari: Yeah `repeat` was used on another test to be more reliable

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: Done.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -403,6 +405,31 @@ pub const FSWatcher = struct {
...
+                            }
+                            // Add new files or directories to the watch list.
+                            if (!is_watched) {
+                                const full_path_clone = bun.default_allocator.dupe(u8, full_path) catch unreachable;
+                                this.file_paths.put(full_path_clone, {}) catch unreachable;
+                                const fs_info = fdFromAbsolutePathZ(full_path) catch null;
+                                if (fs_info) |info| {
+                                    if (info.is_file) {
+                                        this.default_watcher.?.addFileAssumeLocked(info.fd, full_path_clone, file_hash, options.Loader.file, 0, null, false) catch unreachable;
+                                        const relative_path_clone = bun.default_allocator.dupe(u8, relative_path) catch unreachable;
```
cirospaciari: If we save the file paths and their relative paths into the HashMap we can avoid recreating the relative and dupe it here, and just do current_task with `.none` so we only clone it once and don't free it when the task is consumed

Hanaasagi: What was the original purpose of the `file_paths` variable? I haven't seen any place in the code where the values from `file_paths` are read (except for this pr). Is this s used to store some path variables that need to be freed by this struct? Should I define `file_paths` as a structure with keys as `full_path` and values as `relative_path` to better represent the association between paths?


Thanks.

cirospaciari: The original purpose is to keep it alive so the watcher.zig can use it safely (it expect that this paths are alive), but at this point we can use it to avoid the extra dupes. key as full path and values as relative sounds good for this case.

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -357,9 +367,26 @@ pub const FSWatcher = struct {
                             if (this.verbose)
                                 Output.prettyErrorln("<r><d>File changed: {s}<r>", .{relative_slice});
 
-                            const relative_path = bun.default_allocator.dupe(u8, relative_slice) catch unreachable;
-
-                            current_task.append(relative_path, event_type, .destroy);
+                            if (this.file_paths.getEntry(file_path)) |*entry| {
```
Hanaasagi, line 370: I also change this branch (`kind` == `.file`). Now it won't copy the new variable for each event.

src/bun.js/node/node_fs_watcher.zig
Hanaasagi: Fixed.

cirospaciari:
I'll wait for the tests, and will do a deeper look at them, but LGTM

EDIT: @Hanaasagi can you update/rebase this branch?

Hanaasagi:
@cirospaciari Done.

test/js/node/watch/fs.watch.test.ts
```diff
@@ -97,6 +102,67 @@ describe("fs.watch", () => {
...
+      }
+      count++;
+      try {
+        if (count == 1) {
+          ready = true;
+        } else if (count == 2) {
+          expect(event).toBe("rename");
+          expect(filename).toBe("new-file.txt");
+        } else if (count == 3) {
+          expect(event).toBe("change");
```
Hanaasagi, line 129: It seems that on darwin platform, `onFSEventUpdate` is being executed instead of `onFileUpdate`. Unfortunately, I don't have a Mac laptop at the moment to test this 🤦. Feel sorry about this.

https://github.com/oven-sh/bun/blob/2dc90f39081131f8994dbf687f131cfa2105ad25/src/bun.js/node/node_fs_watcher.zig#L271-L288

cirospaciari: No worries if needed I fix macOS after this

SkylerLipthay:
What's the post-mortem on this PR? Is there another path forward for addressing #3657? I'd like to help.

PR #3661, opened 2023-07-17
feat(bun/run): Impl. npm_package_config

### LINKED ISSUE #73: Add support for npm_package_config_* variables
To minimise repetition in package.json scripts, it'd be useful to support the `$npm_package_config_` prefix enabling:

```json
{
  "name": "whatever",
  "config": {
    "buildCommand": "echo env TS_NODE_PROJECT=tsconfig.common.json some-bundler"
  },
  "scripts": {
    "build:dev": "$npm_package_config_buildCommand --mode development",
    "build:prod": "$npm_package_config_buildCommand --mode production"
  }
}
```

For `npm run build:dev`, `yarn run build:dev` and `pnpm run build:dev` the result is:
```
env TS_NODE_PROJECT=tsconfig.common.json some-bundler --mode development
```

For `bun`, the output is currently 😢 
```
bun run build:dev
$ $npm_package_config_buildCommand --mode development
/usr/local/bin/bash: line 1: --mode: command not found
Script error "build:dev" exited with 127 status
```

v3nntrnx:
Closes #73

<sub>Thanks to @InKryption for helping me <3</sub>


src/cli/run_command.zig
```diff
@@ -241,6 +243,19 @@ pub const RunCommand = struct {
 
         try replacePackageManagerRun(&copy_script, script);
 
+        if (package_json) |pkg_json| {
+            var i: usize = 0;
+            const prefix = "$npm_package_config_";
+            while (std.mem.indexOfPos(u8, copy_script.items, i, prefix)) |start| {
```
Jarred-Sumner, line 274: this is inlining all the package.json config environment variables from the shell command? it won't be very reliable this way, but i'm not sure why this is necessary

src/resolver/package_json.zig
```diff
@@ -667,6 +669,22 @@ pub const PackageJSON = struct {
             }
         }
 
+        if (json.asProperty("config")) |npm_pkg_cfg| {
```
Jarred-Sumner, line 672: This should not be exposed in the runtime, it should have a flag to only expose it in `bun run`. Similar to what we do with the `"bin"` field

src/resolver/package_json.zig
```diff
@@ -667,6 +669,22 @@ pub const PackageJSON = struct {
...
+            switch (npm_pkg_cfg.expr.data) {
+                .e_object => |obj| {
+                    for (obj.properties.slice()) |*prop| {
+                        const key = prop.key.?.asString(allocator) orelse continue;
+                        const value = prop.value.?.asString(allocator) orelse continue;
```
Jarred-Sumner, line 677: I'm pretty sure it's supposed to work with arbitrary objects, like `config: { foo: { bar: 1 } }` becomes `$npm_config_foo_bar=1`

src/resolver/package_json.zig
Jarred-Sumner: (This way we don't copy all these config objects for things which will never use it, like the bundler)

src/resolver/package_json.zig
Jarred-Sumner: That being said, it'd probably be fine to skip objects for the first version of this since copying them at all would strictly be an improvement over status quo

v3nntrnx:
TODO:
- [x] not inline the values
- [x] expose it only in bun run
- [x] support objects (optional)

TiranexDev:
I have to add support for numbers

v3nntrnx:
important TODO, currently " " (with space) is being interpreted as "" in bun. it is supposed to be like " " in node.js.
also we need to add objects support, but that can be added in the second version as jarred said

[Discord Message with the " " bug](https://discord.com/channels/876711213126520882/903845031230713857/1130937689059889314)

After a bit of investigating and debugging. I found that in zig, the string is correct. but not when comparing it in a js file.
image

it also happens with ' foo ', it just turns 'foo', after digging even more, it turns out the bun executable strips it. the env is set correctly, because if you run bun run dev and dev launched node, it works correctly. so we might need to address this issue in a different pr.

UPDATE: JULY 20, I just filed an issue on github (#3689) and when talking with someone on the discord server, they supposedly already found the fix, so it might be fixed.

UPDATE: JULY 23, seems to be fixed

v3nntrnx:
I want to see what happens with arrays in node, I'll check it tomorrow

v3nntrnx:
image
image
image

interesting... but I think for this version i'll drop the support for arrays, very complicated

v3nntrnx:
uh--- having troubles..

PR #3662, opened 2023-07-17
[install] handle duplicated workspace declarations gracefully

### LINKED ISSUE #3644: Can't depend on workspace packages

0.6.14

Darwin 21.6.0 arm64 arm

Create `package.json` with:

```
{
  "name": "bun-workspaces",
  "module": "index.ts",
  "type": "module",
  "workspaces": [
    "packages/b"
  ],
  "dependencies": {
    "b": "workspace:*"
  }
}
```

Then `packages/b/package.json` with:

```
{
  "name": "b",
  "version": "1.0.0",
  "module": "index.ts",
  "type": "module"
}
```

Running `bun i` in the root of the project should work, and I should have `node_modules/b` symlinked to `packages/b`

### What do you see instead?

Fails with the error:

```
bun install v0.6.14 (b5665739)


error: Workspace name "b" already exists
{
^
/Users/sergio/src/experiments/bun-workspaces/package.json:1:1 0
```

### Additional information

It works if I remove the dependency on `b` in `package.json`. But it goes against what other package managers do, and agains the docs in https://bun.sh/docs/install/workspaces

Also, it doesn't matter if I depend on `b: "*"`, `b: "1.0.0"` or any other variant, I always get the same error regarding duplicated workspace.

alexlamsl:
fixes #3644

### DIFF
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -2533,7 +2533,7 @@ pub const Package = extern struct {
                 dependency_version.value.workspace = path;
             } else {
                 const workspace = dependency_version.value.workspace.slice(buf);
-                const path = string_builder.append(
+                var path = string_builder.append(
                     String,
                     if (strings.eqlComptime(workspace, "*")) "*" else Path.relative(
                         FileSystem.instance.top_level_dir,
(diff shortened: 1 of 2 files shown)


src/install/lockfile.zig
```diff
@@ -2547,13 +2547,32 @@ pub const Package = extern struct {
...
+                defer dependency_version.value.workspace = path;
                 var workspace_entry = try lockfile.workspace_paths.getOrPut(allocator, @truncate(u32, external_name.hash));
                 if (workspace_entry.found_existing) {
-                    log.addErrorFmt(&source, logger.Loc.Empty, allocator, "Workspace name \"{s}\" already exists", .{
-                        external_name.slice(buf),
-                    }) catch {};
-                    return error.InstallFailed;
+                    const old_path = workspace_entry.value_ptr.*;
+
+                    if (strings.eqlComptime(workspace, "*")) {
```
Jarred-Sumner, line 2555: Should `*` be special cased? That's my main question here. Or should we just ignore duplicate dependency errors for workspace packages that are also in dependencies.

alexlamsl: I see `workspace:*` as "whatever you can find under `workspaces: [...]` that fits my bill, use that"

We are essentially ignoring it, as we `return null`, i.e. dropped the dependency instead of creating a new entry under `package.dependencies`

PR #3664, opened 2023-07-18
package json `main` field extension order

### LINKED ISSUE #3659: BUG: error: Cannot use GraphQLScalarType "String" from another module or realm is thrown when trying to use graphql-jit to compileQuery

v0.6.14 (b5665739)

Ubuntu

1. Check out https://github.com/capaj/bun-graphql-sample-server
2. bun install
3. bun run server.ts

No error is thrown

### What do you see instead?

```
error: Cannot use GraphQLScalarType "String" from another module or realm.

Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
      at /home/capaj/oss/bun-graphql/node_modules/graphql/jsutils/instanceOf.js:43:56
,
      extensions: undefined,
      stack: "Error: Cannot use GraphQLScalarType \"String\" from another module or realm.\n\nEnsure that there is only one instance of \"graphql\" in the node_modules\ndirectory. If different versions of \"graphql\" are the dependencies of other\nrelied on modules, use \"resolutions\" to ensure only one version is installed.\n\nhttps://yarnpkg.com/en/docs/selective-version-resolutions\n\nDuplicate \"graphql\" modules cannot be used at the same time since different\nversions may have different capabilities and behavior. The data from one\nversion used in the function from another could produce confusing and\nspurious results.\n    at instanceOf (/home/capaj/oss/bun-graphql/node_modules/graphql/jsutils/instanceOf.js:42:56)\n    at isScalarType (/home/capaj/oss/bun-graphql/node_modules/graphql/type/definition.js:116:60)\n    at isLeafType (/home/capaj/oss/bun-graphql/node_modules/graphql/type/definition.js:279:46)\n    at compileType (/home/capaj/oss/bun-graphql/node_modules/graphql-jit/dist/execution.js:370:102)\n    at compileDeferredField (/home/capaj/oss/bun-graphql/node_modules/graphql-jit/dist/execution.js:278:17)\n    at <anonymous> (/home/capaj/oss/bun-graphql/node_modules/graphql-jit/dist/execution.js:256:53)\n    at forEach (native)\n    at compileDeferredFields (/home/capaj/oss/bun-graphql/node_modules/graphql-jit/dist/execution.js:254:16)\n    at compileOperation (/home/capaj/oss/bun-graphql/node_modules/graphql-jit/dist/execution.js:229:16)\n    at compileQuery (/home/capaj/oss/bun-graphql/node_modules/graphql-jit/dist/execution.js:84:27)",
      name: "GraphQLError",
      toString: [Function: toString],
      toJSON: [Function: toJSON]
    }
  ]
}
```

### Additional information

this is know problem even on node.js which is easily solvable by adding 
```
  "resolutions": {
    "graphql": "^16.7.1",
  }
```

seems like bun package manager does not take it into account.

dylan-conway:
fixes #3659

### DIFF
--- a/src/options.zig
+++ b/src/options.zig
@@ -1417,6 +1417,7 @@ pub const BundleOptions = struct {
     public_path: []const u8 = "",
     extension_order: []const string = &Defaults.ExtensionOrder,
     esm_extension_order: []const string = &Defaults.ModuleExtensionOrder,
+    main_field_extension_order: []const string = &Defaults.MainFieldExtensionOrder,
     out_extensions: bun.StringHashMap(string),
     import_path_format: ImportPathFormat = ImportPathFormat.relative,
     framework: ?Framework = null,
@@ -1553,6 +1554,16 @@ pub const BundleOptions = struct {
             ".js",
             ".mjs",
             ".mts",
+            ".json",
+        };
+
+        pub const MainFieldExtensionOrder = [_]string{
+            ".js",
+            ".cjs",
+            ".cts",
+            ".tsx",
+            ".ts",
+            ".jsx",
             ".json",
         };
 
--- a/src/resolver/resolver.zig
+++ b/src/resolver/resolver.zig
@@ -3378,7 +3378,13 @@ pub const Resolver = struct {
                         continue;
                     };
 
-                    var _result = r.loadFromMainField(path, dir_info, field_rel_path, key, extension_order) orelse continue;
+                    var _result = r.loadFromMainField(
+                        path,
+                        dir_info,
+                        field_rel_path,
+                        key,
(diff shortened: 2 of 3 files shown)


capaj:
lovely, this should fix it for good. I've been using this exact resolution order in vitest without any issue at least a year now on multiple projects.

PR #3678, opened 2023-07-19
Better error for workspace dependency not found

Jarred-Sumner:
This might make sense to not merge after @alexlamsl's changes but we shall see

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -2601,7 +2601,9 @@ pub const PackageManager = struct {
             },
             .workspace => {
                 // relative to cwd
-                const res = FolderResolution.getOrPut(.{ .relative = .workspace }, version, this.lockfile.str(&version.value.workspace), this);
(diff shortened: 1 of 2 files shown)


src/install/install.zig
```diff
@@ -2601,7 +2601,9 @@ pub const PackageManager = struct {
             },
             .workspace => {
                 // relative to cwd
-                const res = FolderResolution.getOrPut(.{ .relative = .workspace }, version, this.lockfile.str(&version.value.workspace), this);
+                const workspace_path: *const String = this.lockfile.workspace_paths.getPtr(@truncate(String.Builder.stringHash(this.lockfile.str(&version.value.workspace)))) orelse &version.value.workspace;
```
alexlamsl, line 2604: This might end up getting changed due to `workspace_paths` storing versioned paths of the same name &minus; but we can just merge this now regardless.

alexlamsl (approved):
LGTM &minus; that `test/cli/install/bunx.test.ts` failure seems unrelated.

Issue #3572, opened 2023-07-08
TextDecoder does not respect fatal


PotentialStyx:
### What version of Bun is running?

0.6.14

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 unknown

### What steps can reproduce the bug?

When creating a TextDecoder it can be configured to throw an error when it encounters invalid data like so:
```ts
const decoder = new TextDecoder("utf-8", { fatal: true });
```

However bun does not respect this, as seen with this code:
```ts
const decoder = new TextDecoder("utf-8", { fatal: true });
console.log(decoder.decode(new Uint8Array([195])));
```

### What is the expected behavior?

The code should error when run, yet bun doesn't throw an error, and just returns a string.

In chrome the same code gives the following error:
```
VM28:2 Uncaught TypeError: Failed to execute 'decode' on 'TextDecoder': The encoded data was not valid.
    at <anonymous>:2:21
(anonymous) @ VM28:2
```

And in firefox:
```
Uncaught TypeError: TextDecoder.decode: Decoding failed.
    <anonymous> debugger eval code:1
```

### What do you see instead?

Bun instead outputs:
```ascii
�
```

PR #3692, opened 2023-07-19
Pass constructor arguments to TextDecoder

Parzival-3141:
The constructor now actually sets TextDecoder properties using the options parameter. This lets users enable the `fatal` parameter, closing #3572.

### DIFF
--- a/src/bun.js/webcore/encoding.zig
+++ b/src/bun.js/webcore/encoding.zig
@@ -676,25 +676,56 @@ pub const TextDecoder = struct {
         globalThis: *JSC.JSGlobalObject,
         callframe: *JSC.CallFrame,
     ) callconv(.C) ?*TextDecoder {
-        var args_ = callframe.arguments(1);
+        var args_ = callframe.arguments(2);
         var arguments: []const JSC.JSValue = args_.ptr[0..args_.len];
 
-        var encoding = EncodingLabel.@"UTF-8";
-        if (arguments.len > 0) {
-            if (!arguments[0].isString()) {
-                globalThis.throwInvalidArguments("TextDecoder(encoding) label is invalid", .{});
+        var decoder = getAllocator(globalThis).create(TextDecoder) catch unreachable;
+        decoder.* = TextDecoder{}; // Is this necessary? Does allocation initialize variables?
+
+        var ctor_failed = false; // using errdefer would be much nicer but would require another fn
+        defer if (ctor_failed) getAllocator(globalThis).destroy(decoder);
+
+        if (arguments.len == 0) {
+            return decoder;
+        }
+
+        if (!arguments[0].isString()) {
+            globalThis.throwInvalidArguments("TextDecoder(encoding) label is invalid", .{});
+            ctor_failed = true;
+            return null;
+        }
+
+        // encoding
+        {
+            var str = arguments[0].toSlice(globalThis, default_allocator);
+            defer if (str.isAllocated()) str.deinit();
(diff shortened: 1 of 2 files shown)


src/bun.js/webcore/encoding.zig
```diff
@@ -676,25 +676,56 @@ pub const TextDecoder = struct {
...
+            const options = arguments[1];
+
+            if (!options.isObject()) {
+                globalThis.throwInvalidArguments("TextDecoder(options) is invalid", .{});
+                ctor_failed = true;
+                return null;
+            }
+
+            if (options.get(globalThis, "fatal")) |fatal| {
+                decoder.fatal = fatal.asBoolean();
```
Jarred-Sumner: What happens if you pass a `Symbol` or a value which is not a boolean here and in `ignoreBOM`?

src/bun.js/webcore/encoding.zig
```diff
@@ -676,25 +676,56 @@ pub const TextDecoder = struct {
...
+        var args_ = callframe.arguments(2);
         var arguments: []const JSC.JSValue = args_.ptr[0..args_.len];
 
-        var encoding = EncodingLabel.@"UTF-8";
-        if (arguments.len > 0) {
-            if (!arguments[0].isString()) {
-                globalThis.throwInvalidArguments("TextDecoder(encoding) label is invalid", .{});
-                return null;
-            }
+        var decoder = getAllocator(globalThis).create(TextDecoder) catch unreachable;
```
Jarred-Sumner: If we validate arguments and then allocate, we can avoid deallocation

It would also be okay to create a stack-local instance of TextDecoder and then allocate and assign it at the end

Jarred-Sumner (changes requested):
Thank you. This is close, but not ready to merge yet

Parzival-3141:
Two things that I noticed while testing:
1. `TextDecoder.decode` throws a generic `Error` instead of `TypeError` as specified in the WHATWG standard. Is this intentional or an oversight?
2. The getter for `TextDecoder.ignoreBOM` seems to be broken and will always return `undefined`. I'm not sure why this happens

paperclover:
point 1 is definitely an oversight, or maybe there wasnt a binding for `throwTypeError` at the time.

the hint for point 2 is in `encoding.classes.ts`

Parzival-3141:
@paperdave hm, I can't seem to get my changes in `encoding.classes.ts` to work. Running `make codegen` fails as well:
```sh
~/bun$ make codegen
bun src/bun.js/scripts/generate-jssink.js
/usr/bin/clang-format-15 -i  src/bun.js/bindings/JSSink.cpp  src/bun.js/bindings/JSSink.h
/home/ubuntu/bun/src/bun.js/WebKit/Source/JavaScriptCore/create_hash_table src/bun.js/bindings/JSSink.cpp > src/bun.js/bindings/JSSinkLookupTable.h
/usr/bin/bash: line 1: /home/ubuntu/bun/src/bun.js/WebKit/Source/JavaScriptCore/create_hash_table: No such file or directory
make: *** [Makefile:1474: generate-sink] Error 127
```

paperclover:
We should copy that script into our repo, but you can use `make generate-classes` instead

Parzival-3141:
@paperdave no dice unfortunately. `make generate-classes` runs fine but `ignoreBOM` is still undefined. Is there something else I'm missing?

Jarred-Sumner:
Thank you

PR #3718, opened 2023-07-21
Update `development.md`

arturovt:


paperclover:
We don't use GCC when compiling bun. The instructions there should refer to earlier in the document where LLVM and clang 15 is specifically requested. Bun should not be compiled on anything other than that.

arturovt:
> We don't use GCC when compiling bun. The instructions there should refer to earlier in the document where LLVM and clang 15 is specifically requested. Bun should not be compiled on anything other than that.

Hey. I haven’t said “you” are using GCC explicitly when compiling bun. I said clang is linking against the standard library (which is provided by gcc) by default in Linux. That’s not happening e.g. on Mac

paperclover:
I understand.

> While Clang may link against the `libc++` library

I missed that.

docs/project/development.md
```diff
@@ -260,6 +260,37 @@ $ git checkout <hash>
 
 ## Troubleshooting
 
+### 'span' file not found
+
+On Linux systems, the Clang compiler typically uses the `libstdc++` C++ standard library by default. `libstdc++` is the default C++ Standard Library implementation provided by the GNU Compiler Collection (GCC). While Clang may link against the `libc++` library, this requires explicitly providing the `-stdlib` flag when running Clang.
+
+Bun relies on C++20 features like `std::span`, which are not available in GCC versions lower than 11. GCC 10 doesn't have all of the C++20 features implemented. As a result, running `make setup` may fail with the following error:
+
+```bash
```
paperclover: can you remove `bash` from here, since it's not bash script.

(the markdown renderer will format this weirdly)

docs/project/development.md
```diff
@@ -260,6 +260,37 @@ $ git checkout <hash>
...
+# Now run `apt install` again
+$ sudo apt install gcc-11 g++-11
+```
+
+Now, we need to set GCC 11 as the default compiler:
+
+```bash
+$ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 100
+$ sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-11 100
+```
```
paperclover, line 294: this is probably fine since i don't think this issue happens on any other distros. maybe can just clarify that these instructions are for ubuntu.

paperclover (changes requested):
minor edits requested. looks good overall

docs/project/development.md
arturovt: I added `on Ubuntu` in the heading and a warning that implies to use the provided commands on Ubuntu

paperclover (approved):
thanks

PR #3721, opened 2023-07-21
clean tables before testing

cirospaciari:


Jarred-Sumner:
now we just hope two tests don't run at the same time 😆

PR #3732, opened 2023-07-21
Merge import.meta.require and require to be the same thing

### LINKED ISSUE #3724: Vite dev is failing on 0.7

0.7.0

Darwin 22.4.0 arm64 arm

` bun --bun vite dev
`
returns 
`failed to load config from /Users/mohammedkalouti/Dev/S9OSS/InfolinkUi/vite.config.ts
error when starting dev server:
Error: config must export or return an object.
    at <anonymous> (/Users/mohammedkalouti/Dev/S9OSS/InfolinkUi/node_modules/vite/dist/node/chunks/dep-a8e37fae.js:66014:15)
error: "vite" exited with code 1 (SIGHUP)
`

this is the vite config `import {defineConfig} from 'vite'
import react from '@vitejs/plugin-react-swc'
import {posix} from "path";

const config = defineConfig({
    plugins: [react()],
    build: {
        outDir: "build",
    },
    server: {
        port: 3000,

    },
    resolve: {
        alias: [
            {find: "@/src", replacement: posix.resolve("/", "src")},
            {find: "src", replacement: posix.resolve("/", "/src")}
        ]
    },
})

export default config`

Vite dev runs 

### What do you see instead?

import {defineConfig} from 'vite'
import react from '@vitejs/plugin-react-swc'
import {posix} from "path";

const config = defineConfig({
    plugins: [react()],
    build: {
        outDir: "build",
    },
    server: {
        port: 3000,

    },
    resolve: {
        alias: [
            {find: "@/src", replacement: posix.resolve("/", "src")},
            {find: "src", replacement: posix.resolve("/", "/src")}
        ]
    },
})

export default config

### Additional information

_No response_

paperclover:
We used to implement `require` twice. once for `import.meta.require` and then once in the actual `require` implementation. This is just old code from before the CJS rewrite.

`import.meta.require` and `createRequire` now create an empty `JSCommonJSModule` and return it's `require` function.

Fixes #3724, because the old require implementation would not append `__esModule` to certain case.

### DIFF
--- a/src/bun.js/bindings/CommonJSModuleRecord.cpp
+++ b/src/bun.js/bindings/CommonJSModuleRecord.cpp
@@ -164,98 +164,74 @@ JSC_DEFINE_HOST_FUNCTION(requireResolvePathsFunction, (JSGlobalObject * globalOb
     return JSValue::encode(JSC::constructEmptyArray(globalObject, nullptr, 0));
 }
 
+JSC_DEFINE_CUSTOM_GETTER(jsRequireCacheGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
+{
+    Zig::GlobalObject* thisObject = jsCast<Zig::GlobalObject*>(globalObject);
+    return JSValue::encode(thisObject->lazyRequireCacheObject());
+}
+
+JSC_DEFINE_CUSTOM_SETTER(jsRequireCacheSetter,
+    (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue,
+        JSC::EncodedJSValue value, JSC::PropertyName propertyName))
+{
+    JSObject* thisObject = jsDynamicCast<JSObject*>(JSValue::decode(thisValue));
+    if (!thisObject)
+        return false;
+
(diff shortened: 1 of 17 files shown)


src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -969,4 +938,40 @@ std::optional<JSC::SourceCode> createCommonJSModule(
...
+        moduleObject,
+        ArgList(), 1, jsString(vm, String("require"_s)));
+
+    JSFunction* resolveFunction = JSC::JSBoundFunction::create(vm,
+        globalObject,
+        globalObject->requireResolveFunctionUnbound(),
+        moduleObject,
+        ArgList(), 1, jsString(vm, String("require"_s)));
+
+    requireFunction->putDirect(vm, builtinNames.resolvePublicName(), resolveFunction, 0);
```
Jarred-Sumner: PropertyAttribute::Function | 0

Jarred-Sumner:
Looks like this breaks plugins because plugins for "real" CommonJS aren't actually implemented yet (that's why plugins + svelte tests are failing) 

I think that will need to be addressed before this PR can move forward.

PR #3733, opened 2023-07-22
fix #3716

### LINKED ISSUE #3716: Bun.env truncates .env string values containing `#` when run as an npm script

0.6.14

Darwin 22.5.0 x86_64 i386

Create a directory containing
.env
```
TEST_VAR='ABC#DEF'
```

reproFile.js
```
console.log(Bun.env.TEST_VAR);
```

package.json
```
{
  "name": "bun-env-bug-repro",
  "scripts": {
    "reproScript": "bun run reproFile.js"
  }
}
```

cd into new directory
run `bun run reproFile.js` - the terminal output is correct `ABC#DEF`
run `bun reproScript` - the terminal output is incorrect `ABC` (truncates `#` and rest of string)

The output of `bun reproScript` and `bun run reproFile.js` is `ABC#DEF` in both cases

### What do you see instead?

`bun run reproFile.js` outputs `ABC#DEF`
`bun reproScript` outputs `ABC` and truncates the `#` and the rest of the string

### Additional information

The same bug appears if the reproFile is `.ts`
`bun run reproScript` has the same bug as just `bun reproScript`

Not sure whether that's any help

dylan-conway:
fixes #3716

### DIFF
--- a/src/env_loader.zig
+++ b/src/env_loader.zig
@@ -329,13 +329,24 @@ pub const Loader = struct {
     pub fn loadProcess(this: *Loader) void {
         if (this.did_load_process) return;
 
-        // This is a little weird because it's evidently stored line-by-line
-        var source = logger.Source.initPathString("process.env", "");
-
         this.map.map.ensureTotalCapacity(std.os.environ.len) catch unreachable;
-        for (std.os.environ) |env| {
-            source.contents = bun.span(env);
-            Parser.parse(&source, this.allocator, this.map, true, true);
+        for (std.os.environ) |_env| {
+            var env = bun.span(_env);
+            if (strings.indexOfChar(env, '=')) |i| {
+                var key = env[0..i];
+                var value = env[i + 1 ..];
+                if (key.len > 0) {
+                    if (value.len > 0) {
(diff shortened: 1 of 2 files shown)


src/env_loader.zig
```diff
@@ -329,13 +329,24 @@ pub const Loader = struct {
...
-        // This is a little weird because it's evidently stored line-by-line
-        var source = logger.Source.initPathString("process.env", "");
-
         this.map.map.ensureTotalCapacity(std.os.environ.len) catch unreachable;
-        for (std.os.environ) |env| {
-            source.contents = bun.span(env);
-            Parser.parse(&source, this.allocator, this.map, true, true);
+        for (std.os.environ) |_env| {
+            var env = bun.span(_env);
+            if (strings.indexOfChar(env, '=')) |i| {
```
Jarred-Sumner, line 335: wonder what happens if an env key has an `=` in it

Jarred-Sumner: according to Stack Overflow, this is correct

https://stackoverflow.com/questions/41109789/is-the-equals-sign-character-allowed-in-the-value-of-an-environment-value-in

Issue #3740, opened 2023-07-22
fs.writeFile mode is not Node.js compatible


gutenye:
### What version of Bun is running?

0.6.14

### What steps can reproduce the bug?

```
import fs from 'fs'
fs.writeFileSync('a.txt', 'content', { mode: 33188 })
```

### What is the expected behavior?

Works fine on Node

### What do you see instead?

TypeError: Invalid mode: must be an octal number
 code: "ERR_INVALID_ARG_TYPE"

Hanaasagi:
It looks like the upper limit of the value is constrained to `0o777` in bun, which is causing this issue.

https://github.com/oven-sh/bun/blob/636cec03e10ab487b1df3057aaab60b4d2b02c99/src/bun.js/node/types.zig#L1008-L1011

In Node.js, this limit is a u32.


https://github.com/nodejs/node/blob/9dd574c9e232515aa894795b8a2522f53024b6f7/lib/internal/validators.js#L69-L80

Jarred-Sumner:
ah, we probably should wrap it

PR #3747, opened 2023-07-22
Fix `writeFileSync` when the mode is greater than 0o777.

Hanaasagi:
Close: #3740

The octal representation of `33188` is `0o100644`, and it appears that this value is from `stat.st_mode`.

Ref: https://man7.org/linux/man-pages/man7/inode.7.html


Jarred-Sumner:
Thank you

PR #3749, opened 2023-07-22
Display the version installed after installing

### LINKED ISSUE #3720: Add bun version into CLI output of the install script

image

to know which bun version was installed

you can run `bun -v` to check, but this would be a nice inclusion

MohammedWaheeb:
This was requested by a user on twitter 

Closes #3720


Electroid:
Instead of making a request to the Github API, it can just output the result of `bun -v`

MohammedWaheeb:
@Electroid done

paperclover:
Closes #3720

(i know typing this in the main comment will make it close but idk if i type it in a comment here)

MohammedWaheeb:
@paperdave is there anything i need to do so it can be merged

birkskyum:
can we make a way to print the commit hash a version is build from?

When I use `bun -v` i just see the generic:

```
➜ bun -v
0.7.4
```


**An example of a more informative cli is **rustc**:**

``` zig
➜ rustc -V
rustc 1.70.0 (90c541806 2023-05-31) (built from a source tarball)
```

paperclover:
> can we make a way to print the commit hash a version is build from?
> 
> When I use `bun -v` i just see the generic:
> 
> ```
> ➜ bun -v
> 0.7.4
> ```
> 
> **An example of a more informative cli is rustc:**
> 
> ```zig
> ➜ rustc -V
> rustc 1.70.0 (90c541806 2023-05-31) (built from a source tarball)
> ```

#4027 added `bun --revision` which prints a string like

```
0.7.4-debug+d6cf039ccd0e3fd36f57564c816e68fef1403189
```

(we cannot change -v because bun upgrade relies on it and would break older bun upgrades)

MohammedWaheeb:
@paperdave should i add it next to or under the version

PR #3754, opened 2023-07-23
[install] improve workspace substitution of npm dependencies

alexlamsl:
- respect semver ranges
- allow double definition of package within `workspaces` & `dependencies` as long as versioning is compatible


Jarred-Sumner:
Thank you

PR #3759, opened 2023-07-23
wip: bun-plugin-lightningcss

TommasoAmici:
work started here https://github.com/oven-sh/bun/pull/2888

One thing I'm not sure about is having both a `bun-plugin-lightningcss` and a `bun-plugin-css`. 

The fact we're using Lightningcss is just an implementation detail.

I think these two plugins should be merged into `bun-plugin-css`


Jarred-Sumner:
Closing due to bitrot

PR #3761, opened 2023-07-23
Fixes #172

### LINKED ISSUE #172: `NapiClass` constructor cannot be subclassed
Filing so I remember to fix this

This currently impacts `@resvg/resvg-js`

```
module.exports.Resvg = class Resvg extends _Resvg {
  constructor(svg, options) {
     super(svg, JSON.stringify(options))
```

The problem in Bun is this code:
```c++
   NapiClass* napi = jsDynamicCast<NapiClass*>(newTarget);
    if (UNLIKELY(!napi)) {
        JSC::throwVMError(globalObject, scope, JSC::createTypeError(globalObject, "NapiClass constructor called on an object that is not a NapiClass"_s));
        return JSC::JSValue::encode(JSC::jsUndefined());
    }

```

That `!napi` is being returned because `newTarget` is a subclass of NapiClass.

Maybe the prototype needs to be checked instead? I'm not sure.

arturovt:
I've firstly checked the comment left by Jarred in the issue, but seems like `inherits` is never truthy (I was using `@resvg/resvg-js`). I was only able to get a valid pointer to `NapiClass` going through the prototype chain. I don't know if the code in this PR is valid, but I was able to make `@resvg/resvg-js` functioning.

P.S. where should be tests located in?

Issue: #172

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -1245,7 +1245,21 @@ JSC_DEFINE_HOST_FUNCTION(NapiClass_ConstructorFunction,
 
     JSObject* newTarget = asObject(callFrame->newTarget());
 
-    NapiClass* napi = jsDynamicCast<NapiClass*>(newTarget);
+    NapiClass* napi = nullptr;
+    if (newTarget->inherits<NapiClass>()) {
+        napi = jsCast<NapiClass*>(newTarget);
+    } else {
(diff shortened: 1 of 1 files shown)


paperclover:
This definitely needs tests. They can go anywhere within `test/js`, probably could make an napi tests folder like `test/js/node/napi`

paperclover:
This was fixed in #4220

PR #3770, opened 2023-07-23
Set error code `EADDRINUSE` in case conflict port, closes #3767

### LINKED ISSUE #3767: bun fails to increment vite's port

0.7.1

ubuntu 22.04

bun --bun vite

```
Forced re-optimization of dependencies
Port 5000 is in use, trying another one...

  VITE v4.4.6  ready in 1662 ms

  ➜  Local:   http://localhost:5001/
  ➜  Network: use --host to expose
  ➜  press h to show help
```

### What do you see instead?

```
Forced re-optimization of dependencies
358 |     try {
359 |       const tls = this.#tls;
360 |       if (tls)
361 |         this.serverName = tls.serverName || host || "localhost";
362 | 
363 |       this.#server = Bun.serve({
                        ^
error: Failed to start server. Is port 5000 in use?
      at listen (node:http:363:21)

  VITE v4.4.6  ready in 1512 ms

  ➜  Network: use --host to expose
  ➜  press h to show help
```

### Additional information

_No response_

hyuzea:


Jarred-Sumner:
thank you for this, but it should be done in the native code instead of in node:http

yus-ham:
> thank you for this, but it should be done in the native code instead of in node:http

Indeed I also prefer to use native code. but until now I've tried I still can't have running bun codebase.
Does bun will also use same API with node for error codes?

shivak:
@Jarred-Sumner what is the desired behavior? I think the clearest approach would be to just turn off `SO_REUSEPORT` when Bun listens. In development, that flag is a foot gun. In production with load balancing, a parent watchdog process should be used to manage the port and workers. 

I can just turn off the flag as part of https://github.com/oven-sh/bun/pull/2852 .

PR #3780, opened 2023-07-24
bun:sqlite gets 10% faster

Jarred-Sumner:
By being a little smarter about how JSC APIs are used, bun:sqlite gets 10% faster at returning results from sqlite

image

cc @AlexBlokh

### DIFF
--- a/src/bun.js/bindings/sqlite/JSSQLStatement.cpp
+++ b/src/bun.js/bindings/sqlite/JSSQLStatement.cpp
@@ -235,76 +235,75 @@ static void initializeColumnNames(JSC::JSGlobalObject* lexicalGlobalObject, JSSQ
     if (count == 0)
         return;
 
-    // if (count > 63) {
-    JSC::ObjectInitializationScope initializationScope(vm);
+    if (count > 63) {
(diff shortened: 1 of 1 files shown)


src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -1058,50 +1076,76 @@ static inline JSC::JSValue constructResultObject(JSC::JSGlobalObject* lexicalGlo
...
-            case SQLITE3_TEXT: {
-                size_t len = sqlite3_column_bytes(stmt, i);
-                const unsigned char* text = len > 0 ? sqlite3_column_text(stmt, i) : nullptr;
+        });
 
-                if (len > 64) {
-                    result->putDirectOffset(vm, i, JSC::JSValue::decode(Bun__encoding__toStringUTF8(text, len, lexicalGlobalObject)));
-                    continue;
-                }
+        result = JSC::constructEmptyObject(vm, structure);
```
Jarred-Sumner, line 1139: btw @devongovett this is an example of how to populate an object really fast from native code

Issue #3782, opened 2023-07-24
`dirname` returns wrong result When the path is the root


Hanaasagi:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Linux 6.4.4-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const path = require("path");
const assert = require("assert");
assert.strictEqual(path.posix.dirname("/"), "/");

```

### What is the expected behavior?

No `AssertionError``.

### What do you see instead?

```
AssertionError: Expected values to be strictly equal:

'.' !== '/'

 code: "ERR_ASSERTION"

```

### Additional information

I think this bug is related to https://github.com/oven-sh/bun/issues/2676

Here, the returned value should be determined by more context details instead of returning a `.`.

https://github.com/oven-sh/bun/blob/967ccb5d50b3ce163bb98fd125c56b039743de90/src/bun.js/node/types.zig#L1699-L1702

Hanaasagi:
After running the `dirname` related test files in Nodejs, I found some other incompatibilities.

```JavaScript
assert.strictEqual(path.posix.dirname("////"), "/");
assert.strictEqual(path.posix.dirname("//a"), "//");
assert.strictEqual(path.win32.dirname("c:\\"), "c:\\");
assert.strictEqual(path.win32.dirname("\\"), "\\");

assert.strictEqual(path.win32.dirname("c:"), "c:");
assert.strictEqual(path.win32.dirname("c:foo"), "c:");
assert.strictEqual(path.win32.dirname("c:foo\\"), "c:");
assert.strictEqual(path.win32.dirname("\\\\unc\\share"), "\\\\unc\\share");
assert.strictEqual(path.win32.dirname("/"), "/");
assert.strictEqual(path.win32.dirname("////"), "/");
```


Ref: https://github.com/nodejs/node/blob/main/test/parallel/test-path-dirname.js

Jarred-Sumner:
I’m surprised we still have these bugs

Hanaasagi:
I think i can fix this.

Jarred-Sumner:
Please do

PR #3785, opened 2023-07-24
fix some cases in the `dirname` function.

Hanaasagi:
Close: #3782

### DIFF
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -1682,6 +1682,87 @@ pub const Path = struct {
 
         return JSC.ZigString.init(out).withEncoding().toValueGC(globalThis);
     }
+
+    fn dirnameWindows(path: []const u8) []const u8 {
+        if (path.len == 0)
+            return ".";
+
+        const root_slice = std.fs.path.diskDesignatorWindows(path);
+        if (path.len == root_slice.len)
+            return root_slice;
+
+        const have_root_slash = path.len > root_slice.len and (path[root_slice.len] == '/' or path[root_slice.len] == '\\');
+
+        var end_index: usize = path.len - 1;
+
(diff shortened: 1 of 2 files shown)


src/bun.js/node/types.zig
```diff
@@ -1682,6 +1682,87 @@ pub const Path = struct {
 
         return JSC.ZigString.init(out).withEncoding().toValueGC(globalThis);
     }
+
+    fn dirnameWindows(path: []const u8) []const u8 {
```
Hanaasagi, line 1686: If still use `std.fs.path.dirnameWindows`, it may cause the calling side to re-parse the path in some cases to handle `null` scenarios.

Code borrowed and modified from https://github.com/ziglang/zig/blob/e8fa19960243a99cef5d9af94c3db52173787a4e/lib/std/fs/path.zig#L796-L828

src/bun.js/node/types.zig
```diff
@@ -1682,6 +1682,87 @@ pub const Path = struct {
...
+        return path[0..end_index];
+    }
+
+    fn dirnamePosix(path: []const u8) []const u8 {
+        if (path.len == 0)
+            return ".";
+
+        var end_index: usize = path.len - 1;
+
+        while (path[end_index] == '/') {
```
Jarred-Sumner, line 1735: this needs a bounds check

Jarred-Sumner:
i _think_ this is fine, but I would feel about 50% better about this code if we had some tests for non-ascii input. It's iterating byte-by-byte which means it ignores code-units/codepoints

Hanaasagi:
@Jarred-Sumner You're right, I found that `dirname` didn't handle the encoding correctly before. Please see https://github.com/oven-sh/bun/pull/3799

PR #3796, opened 2023-07-25
fix rope string push

### LINKED ISSUE #3751: SyntaxError for valid conditional assignment (regular expression mixed with comments)

0.7.0

Linux 6.2.0-25-generic x86_64 x86_64

The following valid code snippet is from a fork of Google's Closure library and causes the following error when run with Bun:

``` typescript
let SELECTOR_REGEX_ =
    // Don't even evaluate it on older browsers (IE8 and IE9), it throws a
    // syntax error and we don't use it anyway.
    !(userAgent.IE && document.documentMode < 10) ?
    new RegExp(
        '\\s*' +              // Discard initial space
            '([^\\s\'",]+' +  // Beginning of the match. Anything but a comma,
                              // spaces or a string delimiter. This is the only
                              // non-optional component of the regex.
            '[^\'",]*' +      // Spaces are fine afterwards (e.g. "a > b").
            ('(' +  // A series of optional strings with matching delimiters
                    // that can contain anything, and optional non-quoted text
                    // without commas.
             '(\'([^\'\\r\\n\\f\\\\]|\\\\[^])*\')|' +  // Optional single-quoted
                                                       // string.
             '("([^"\\r\\n\\f\\\\]|\\\\[^])*")|' +     // Optional double-quoted
                                                       // string.
             '[^\'",]' +  // Optional non-string content.
             ')*') +      // String and non-string
                          // content can come in any
                          // order.
            ')',          // End of the match.
        'g') :
    null;
```

Bun throws a `SyntaxEception` when parsing this code. I'm not sure if the ternary operator (incl. parenthesis) or the inline comments inside the regex declaration (or the combination of both) causes the problem.

No SyntaxException should be thrown as this is valid code.

### What do you see instead?

The error thrown by Bun is:

``` 
34 |  */
35 | let SELECTOR_REGEX_ =
36 |     // Don't even evaluate it on older browsers (IE8 and IE9), it throws a
37 |     // syntax error and we don't use it anyway.
38 |     !(userAgent.IE && document.documentMode < 10) ?
39 |     new RegExp(
        ^
SyntaxError: Invalid regular expression: missing )
      at /home/daniel/CQSE/teamscale-git/webapp/node_modules/.pnpm/ts-closure-library@2022.502.5/node_modules/ts-closure-library/lib/html/sanitizer/csssanitizer.js:39:4
```

### Additional information

_No response_

dylan-conway:
fixes #3751

### DIFF
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -2163,7 +2163,9 @@ pub const E = struct {
                 this.next = other;
                 this.end = other;
             } else {
-                this.end.?.next = other;
+                var next = this.end.?;
+                while (next.next != null) next = next.next.?;
+                next.next = other;
                 this.end = other;
             }
         }
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -2163,9 +2163,9 @@ pub const E = struct {
                 this.next = other;
                 this.end = other;
             } else {
-                var next = this.end.?;
-                while (next.next != null) next = next.next.?;
-                next.next = other;
+                var end = this.end.?;
+                while (end.next != null) end = end.end.?;
+                end.next = other;
                 this.end = other;
             }
         }
--- a/test/transpiler/transpiler.test.js
+++ b/test/transpiler/transpiler.test.js
@@ -2807,6 +2807,18 @@ console.log(foo, array);
       expectPrinted("'a' + 1", '"a" + 1');
       expectPrinted("x * 'a' + 'b'", 'x * "a" + "b"');
 
+      // rope string push another rope string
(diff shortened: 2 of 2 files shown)


src/js_ast.zig
```diff
@@ -2163,7 +2163,9 @@ pub const E = struct {
                 this.next = other;
                 this.end = other;
             } else {
-                this.end.?.next = other;
+                var next = this.end.?;
+                while (next.next != null) next = next.next.?;
```
Jarred-Sumner: this should just be next.end right?

dylan-conway: yeah that's right. confusingly it still works using next

PR #3797, opened 2023-07-25
Start time performance improvements to build tools

Jarred-Sumner:
- Buffer the file instead of calling read() one byte at a time...
- Use faster method for equality checks
- Cache the result since your CPUs will not change without restarting your computer

### DIFF
--- a/src/bun.js/node/path_watcher.zig
+++ b/src/bun.js/node/path_watcher.zig
@@ -367,7 +367,7 @@ pub const PathWatcherManager = struct {
                     const path_ = path.path;
                     this.main_watcher.remove(path.hash);
                     _ = this.file_paths.remove(path_);
-                    bun.default_allocator.destroy(path_);
+                    bun.default_allocator.free(path_);
                 }
             }
         }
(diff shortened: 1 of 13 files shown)


src/bun.js/node/path_watcher.zig
```diff
@@ -0,0 +1,560 @@
...
+        this.mutex.lock();
+        defer this.mutex.unlock();
+
+        if (this.watcher_count == this.watchers.len) {
+            this.watcher_count += 1;
+            this.watchers.push(bun.default_allocator, watcher) catch unreachable;
+        } else {
+            var watchers = this.watchers.slice();
+            for (watchers, 0..) |w, i| {
+                if (w == null) {
```
Jarred-Sumner, line 359: nit: *w can be used and then no need for a counter

test/js/node/watch/fs.watch.test.ts
```diff
@@ -373,6 +373,11 @@ describe("fs.watch", () => {
     });
     expect(promise).resolves.toBe("change");
   });
+
+  test("immediately closing works correctly", async () => {
+    for (let i = 0; i < 100; i++) fs.watch(testDir, { persistent: true }).close();
+    for (let i = 0; i < 100; i++) fs.watch(testDir, { persistent: false }).close();
```
Jarred-Sumner, line 379: We should also have these tests for files too.

PR #3815, opened 2023-07-26
Implement `fs.watchFile()`

### LINKED ISSUE #3812: Implement fs.watchFile
### Needed for Rollup support

When running Rollup, the missing fs.watchFile breaks.

➜ bun --bun run build-dev
$ rollup --configPlugin @rollup/plugin-typescript -c --environment BUILD:dev
[!] (plugin typescript) TypeError: _fs.watchFile is not a function. (In '_fs.watchFile(fileName, { persistent: !0, interval: pollingInterval }, fileChanged)', '_fs.watchFile' is undefined)

Implement fs.watchFile ([support status](https://bun.sh/docs/runtime/nodejs-apis#node_fs))

_No response_

Electroid:
Adds support for `fs.watchFile()`, will add tests.

Closes #3812

### DIFF
--- a/src/bun.js/node/node_fs_watcher.zig
+++ b/src/bun.js/node/node_fs_watcher.zig
@@ -364,11 +364,6 @@ pub const FSWatcher = struct {
                     }
                 },
                 .directory => {
-                    // macOS should use FSEvents for directories
-                    if (comptime Environment.isMac) {
-                        @panic("Unexpected directory watch");
-                    }
-
                     const affected = event.names(changed_files);
 
                     for (affected) |changed_name_| {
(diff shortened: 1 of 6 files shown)


src/bun.js/node/node_fs_watcher.zig
```diff
@@ -364,11 +364,6 @@ pub const FSWatcher = struct {
                     }
                 },
                 .directory => {
-                    // macOS should use FSEvents for directories
```
Electroid, line 367: @cirospaciari I removed this and everything seemed to work ok

cirospaciari: It's OK to remove, maybe add a debug assert, I just added this to see if I did not forget any place, MacOS should use FSEvents for directories.

Jarred-Sumner:
fs.watchFile is supposed to work by calling stat() repeatedly, avoiding strangeness & inconsistentcies that come from platform & filesystem-specific watchers at a performance & latency cost. Not wrapping fs.watch. In Node, it wraps a `uv_file_poll_t` which is effectively `setInterval` + `stat` but in C.  

The underlying implementation will need to be redone

PR #3845, opened 2023-07-27
Remove no longer needed flag from benchmark documentation

lpinca:
`Deno.serve()` was stabilized in version 1.35.0.


Jarred-Sumner:
Thank you

PR #3846, opened 2023-07-27
Resolve watch directories outside main thread + async iterator and symlink fixes

cirospaciari:


src/bun.js/node/fs_events.zig
```diff
@@ -107,6 +107,12 @@ var fsevents_mutex: Mutex = Mutex.init();
 var fsevents_default_loop_mutex: Mutex = Mutex.init();
 var fsevents_default_loop: ?*FSEventsLoop = null;
 
+fn symlinkPath(path: string, buffer: *[bun.MAX_PATH_BYTES + 1]u8) ?string {
```
Jarred-Sumner: symlinks can have multiple levels

zig's std library here is not reliable due to usage of `catch unreachable`

Do we have a file descriptor for this or nah? if we do, we can use the code path for getting the path from the file

I think it's also worth looking through FSEvents to see if they expose a way to tell you this information in their events

src/bun.js/node/path_watcher.zig
```diff
@@ -111,6 +116,13 @@ pub const PathWatcherManager = struct {
             .vm = vm,
             .watcher_count = 0,
             .mutex = Mutex.init(),
+            // we schedule directory search to run on a thread pool because we dont want to block the main thread
+            // threads are only spawned when some task is schedule
+            .thread_pool = ThreadPool.init(.{
```
Jarred-Sumner: don't use a dedicated thread pool for this

just use JSC.WorkPool

src/bun.js/node/fs_events.zig
cirospaciari: we have the fd because we open it to see if its a directory, FSEvents already resolve the path for me, I only need the dest path if its a symlink so I can check against the watcher. I will get from the fd before passing to FSEvents

src/bun.js/node/path_watcher.zig
```diff
@@ -168,6 +181,8 @@ pub const PathWatcherManager = struct {
 
                         for (watchers) |w| {
                             if (w) |watcher| {
+                                if (watcher.fsevents_watcher != null) continue;
```
Jarred-Sumner: this is a macOS only thing right? could we add a `if (comptime Environment.isMac)` here?

cirospaciari: right i forgot that my bad

src/bun.js/node/fs_events.zig
Jarred-Sumner: we want the resolved symlink path, right?  if yes, use the macOS function for getting the fully resolved symlink path  instead of readlink(). readlink() is only one hop. you could have symlink -> symlink -> dir and this would only catch the 2nd symlink.

src/bun.js/node/fs_events.zig
```diff
@@ -352,27 +358,46 @@ pub const FSEventsLoop = struct {
...
+                if (handle.symlink_path == null) {
+                    handle.symlink_path = symlinkPath(handle.path, &buffer);
+                    if (handle.symlink_path) |path| {
+                        handle.symlink_path = bun.default_allocator.dupe(u8, path) catch |err| {
+                            handle.emit(@errorName(err), false, .@"error");
+                            handle.flush();
+                            continue;
+                        };
+                    } else {
+                        handle.symlink_path = "";
```
Jarred-Sumner: did you mean to make symlink_path both nullable and empty? can we just make it empty and treat that as null?

cirospaciari: i meant to use null to check if I already tried to read the symlink (with can fail if is not a sym link) if is not this turns empty and I stop trying, but I will remove this logic FSEvents will already have the resolved path at this point.

src/bun.js/node/fs_events.zig
cirospaciari: `std.os.getFdPath` did the trick also add a test for this in https://github.com/oven-sh/bun/pull/3846/commits/d2ed85ec96c43fd0be74e60c7dc998d37a0d1b0e

src/bun.js/node/path_watcher.zig
cirospaciari: Fixed on https://github.com/oven-sh/bun/pull/3846/commits/b3995906c528ae6b40cb211c59ed29bec5e2b5ba

Jarred-Sumner:
Let's merge after macOS tests run

src/js/node/fs.promises.ts
```diff
@@ -38,32 +38,51 @@ export function watch(
...
+                closed = true;
+                break;
+              }
+              if (event.eventType === "error") {
+                closed = true;
+                throw event.filename;
+              }
+              return { value: event, done: false };
+            }
+            await new Promise((resolve: Function) => (nextEventResolve = resolve));
```
Jarred-Sumner: This can be `Promise.withResolvers` and that will skip creating the extra temporary function

src/js/node/fs.promises.ts
```diff
@@ -38,32 +38,51 @@ export function watch(
...
-        while (events.length) {
-          let event = events.shift() as Event;
-          if (event.eventType === "close") {
-            closed = true;
-            break;
+      return {
+        async next() {
+          while (!closed) {
+            while (events.length) {
+              let event = events.shift() as Event;
```
Jarred-Sumner: This will be a lot of garbage 

We can use `createFIFO()` for this from `import.meta.primordials`

cirospaciari: fix on https://github.com/oven-sh/bun/pull/3846/commits/897082a4fddabe926c3a5a6661709fc45c8ebd80

src/js/node/fs.promises.ts
cirospaciari: fix on https://github.com/oven-sh/bun/pull/3846/commits/897082a4fddabe926c3a5a6661709fc45c8ebd80

Jarred-Sumner:
yay

PR #3847, opened 2023-07-27
Add ecosystem guides

colinhacks:


docs/guides/ecosystem/elysia.md
```diff
@@ -0,0 +1,31 @@
+---
+name: Use Elysia
```
Jarred-Sumner: This title should be more actionable. Assume nobody knows what Elysia is

Maybe something like

> Building type-safe full-stack apps with Elysia & Bun

docs/guides/ecosystem/express.md
```diff
@@ -0,0 +1,40 @@
+---
+name: Use Express
```
Jarred-Sumner: ditto for this title and others

Issue #3771, opened 2023-07-23
Support Nuxt (Vite framework for Vue)


birkskyum:
Nuxt.js [GitHub](https://github.com/nuxt/nuxt) - [Website ](https://nuxt.com/)

### What version of Bun is running?

0.7.0

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

* `bun x nuxi@latest init my-nuxt-app`
* select `bun` as package manager
* `cd my-nuxt-app`
* `bun --bun run dev`

### What is the expected behavior?

```sh
➜ bun --bun run dev
Nuxi 3.6.5                                                                                                                                                 10:06:22 PM
Nuxt 3.6.5 with Nitro 2.5.2                                                                                                                                10:06:22 PM
[get-port] Unable to find an available port (tried 3000). Using alternative port: 3001                                                                     10:06:23 PM
                                                                                                                                                           10:06:23 PM
  > Local:    http://localhost:3001/ 
  > Network:  http://192.168.0.115:3001/
  > Network:  http://[2a02:a317:e440:2c80:1463:d84f:69c5:ac27]:3001/
  > Network:  http://[2a02:a317:e440:2c80:6c04:d94d:e2fb:8671]:3001/

✔ Nuxt Devtools is enabled v0.7.1 (experimental)                                                                                                          10:06:23 PM
ℹ Vite client warmed up in 591ms                                                                                                                          10:06:24 PM
✔ Nitro built in 249 ms     
```

### What do you see instead?

```sh
➜ bun --bun run dev
$ nuxt dev
223 |     throw err;
224 |   return ret.stdout;
225 | }
226 | function fork() {
227 | 
228 |   throw new Error("Not implemented");
           ^
error: Not implemented
      at fork (node:child_process:228:8)
      at start (/Users/admin/repos/my-nuxt/node_modules/nuxi/dist/cli-wrapper.mjs:27:16)
      at startSubprocess (/Users/admin/repos/my-nuxt/node_modules/nuxi/dist/cli-wrapper.mjs:23:2)
      at /Users/admin/repos/my-nuxt/node_modules/nuxi/dist/cli-wrapper.mjs:7:2
error: script "dev" exited with code 1 (SIGHUP)
```

### Additional information

Related:

- #3793 
- https://github.com/oven-sh/bun/issues/3760
- https://github.com/oven-sh/bun/pull/4090
- #4098 
- https://github.com/oven-sh/bun/pull/4208
[...]

Issue #3620, opened 2023-07-12
Implement `child_process.fork()`


Electroid:
Spawns another process of Bun with IPC.

https://nodejs.org/api/child_process.html#child_processforkmodulepath-args-options

simylein:
Can we close this as it is now implemented?

sirenkovladd:
I think, probably not

https://bun.sh/blog/bun-v0.7.1#child-process-fork-is-partially-implemented
> ... IPC support is not implemented yet and will be added in a future release.

PR #3851, opened 2023-07-28
add fork to child_process

sirenkovladd:
Copy from https://github.com/nodejs/node/blob/v20.5.0/lib/child_process.js#L101-L173
and corrected for `bun`

### DIFF
--- a/src/js/out/modules/node/child_process.js
+++ b/src/js/out/modules/node/child_process.js
@@ -224,10 +224,47 @@ function execSync(command, options) {
     throw err;
   return ret.stdout;
 }
-function fork() {
-  throw new Error("Not implemented");
-}
-var convertToValidSignal = function(signal) {
+var stdioStringToArray = function(stdio, channel) {
+  const options = [];
+  switch (console.log("stdioStringToArray", stdio, channel, options, typeof options), stdio) {
+    case "ignore":
+    case "overlapped":
+    case "pipe":
+      ArrayPrototypePush.call(options, stdio, stdio, stdio);
+      break;
+    case "inherit":
+      ArrayPrototypePush.call(options, 0, 1, 2);
+      break;
+    default:
+      throw new ERR_INVALID_ARG_VALUE("stdio", stdio);
+  }
+  if (channel)
+    ArrayPrototypePush.call(options, channel);
+  return options;
+}, fork = function(modulePath, args = [], options) {
+  modulePath = getValidatedPath(modulePath, "modulePath");
+  let execArgv;
+  if (args == null)
+    args = [];
+  else if (typeof args === "object" && !ArrayIsArray(args))
+    options = args, args = [];
+  else
+    validateArray(args, "args");
+  if (options != null)
+    validateObject(options, "options");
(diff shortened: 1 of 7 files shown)


Jarred-Sumner:
did you try it and see if it works?

sirenkovladd:
Fixes https://github.com/oven-sh/bun/issues/3771 https://github.com/oven-sh/bun/issues/3620

sirenkovladd:
> did you try it and see if it works?

Yes, for `dev` of course)
currently generating for release, may take some time

Jarred-Sumner:
How does it communicate with the parent process and support sending data? The native code parts don't support additional file descriptors I don't think. Or is IPC less commonly used and this is more used as a way to run a file?

src/js/out/modules/node/child_process.js
```diff
@@ -705,7 +745,6 @@ var child_process_default = {
 export {
   spawnSync,
   spawn,
-  fork,
```
birkskyum: How come the named export is removed?

sirenkovladd: I lost the export on the function, i fixed it

Jarred-Sumner:
Can you include automated tests? These can be copied from Node. See https://github.com/oven-sh/bun/blob/main/test/js/node/child_process/child_process-node.test.js#L21  for an example

sirenkovladd:
Ok, actually I found the blocks for this PR, it's the `process.send` method

Without it I can't send messages between processes

But I think this PR can be accepted and we can make a issue on solve `process.send` / `process.on('message')`

sirenkovladd:
### What does this PR do?

Implement method `fork` in `child_process`

the fork method also needs methods for handling messages (process.send, process.on('message'), ChildProcess.send) to be fully compatible with node

- [X] Code changes

- [X] I wrote automated tests
- [X] I tested in my project which used `fork` from `node`

(No rule to make target 'js')



- [X] The types already existed

- [X] I made sure that specific versions of dependencies are used instead of ranged or tagged versions

test/js/node/child_process/child_process-node.test.js
```diff
@@ -483,3 +492,261 @@ describe("child_process double pipe", () => {
...
+  describe.todo("no-shell", () => {
+    // https://github.com/nodejs/node/blob/v20.5.0/test/parallel/test-child-process-fork-no-shell.js
+  });
+  describe.todo("ref", () => {
+    // https://github.com/nodejs/node/blob/v20.5.0/test/parallel/test-child-process-fork-ref.js
+  });
+  describe.todo("stdio", () => {
+    // https://github.com/nodejs/node/blob/v20.5.0/test/parallel/test-child-process-fork-stdio.js
+  });
+  describe.only("fork", () => {
```
Jarred-Sumner: It should not be using .only, instead .skip

Jarred-Sumner:
This looks good, minus the usage of describe.only

Jarred-Sumner (approved):

test/js/node/child_process/child_process-node.test.js
sirenkovladd: Fix it, thank you for finding this

Jarred-Sumner:
The test for message should just be removed entirely since none of that works and a test which immediately calls done should be skipped since it does nothing and it won’t be entirely clear to future readers why the test exists or what it was supposed to do 

That being said, this is an improvement over the status quo so thank you and I’ll merge it now

Issue #3668, opened 2023-07-18
Missing last byte when serving static files


RobotoSkunk:
### What version of Bun is running?

0.6.14

### What platform is your computer?

Linux 6.3.0-1-amd64 x86_64 unknown

### What steps can reproduce the bug?

1. Install `express` and `compression` packages.
2. Configure compression to work with express.
3. Serve static files with express.
4. Add some example files to the static directory.
5. Start the server.
6. Try loading any static file.

### What is the expected behavior?

The whole file is being sent across HTTP.

![image](https://github.com/oven-sh/bun/assets/49355293/f7faba06-432f-4659-be23-7083d53f3ffb)

### What do you see instead?

The last byte (or bytes) is missing in the file, so, things like WebP, JPG, BMP and other files are affected.

![Screenshot_20230726_190405](https://github.com/oven-sh/bun/assets/49355293/ac79c6f9-6d0d-4bd5-881b-6488b8112008)

### Additional information

I have a repository prepared to test Bun (https://github.com/RobotoSkunk/personal-website/).

If you're going to use it, run it and check out the `/assets/img/*` URL path (`public/assets/img/*` in the repo), there are prepared files to test.

This does not happen with Node.js

Hanaasagi:
After some debugging, I believe the minimal reproducing code for this bug is as follows:

### Step 1 create a test file

```Bash
echo -n 12345 > test.txt
```

### Step 2 minimal http static server

index.js

```JavaScript
var http = require("http");
var send = require("send");

var server = http.createServer(function onRequest(req, res) {
  send(req, "./test.txt").pipe(res);
});

server.listen(8080);

```

*NOTE: express utilizes the server-static package to handle static routes, and this package depends on the send package.*

![2023-07-27_23-12](https://github.com/oven-sh/bun/assets/9482395/59d42d30-21f3-4a47-9c9f-331fd60a6476)

The content-length here is also calculated incorrectly (should be 5 not 4).


----

Updated:

I set the debugging flag for this file https://github.com/oven-sh/bun/blob/main/src/js/node/stream.js and found some interesting logs.


![2023-07-28_03-07](https://github.com/oven-sh/bun/assets/9482395/71ffdc59-4231-4c5e-b0d6-f81010ef2cac)

Some callstack


https://github.com/oven-sh/bun/blob/704ee133923a4e28bd51bc95b56eb54d50424c17/src/js/node/stream.js#L5359-L5368


[...]

PR #3853, opened 2023-07-28
fix the chunk boundary (`node:stream:createReadStream`)

Hanaasagi:
Close: #3668

## What does this PR do?

- [x] Code changes

Test Code:

```JavaScript

const fs = require("node:fs");
const s = require("node:stream");

const path = "./test.txt";
fs.writeFileSync(path, "12345");

const stream = fs.createReadStream(path, { start: 0, end: 4 });

const writable = new s.Writable({
  write(chunk, encoding, callback) {
    console.log(chunk);
  },
});

stream.pipe(writable);

```

### Checklist

If Zig files changed:

- [ ] I or my editor ran `zig fmt` on the changed files

- [ ] I made sure that specific versions of dependencies are used instead of ranged or tagged versions

If functions were added to exports.zig or bindings.zig

- [ ] I ran `make headers` to regenerate the C header file

If \*.classes.ts files were added or changed:

- [ ] I ran `make codegen` to regenerate the C++ and Zig code

If a new builtin ESM/CJS module was added:

- [ ] I added a test that imports the module
- [ ] I added a test that require() the module

### DIFF
--- a/src/js/node/fs.js
+++ b/src/js/node/fs.js
@@ -540,14 +540,14 @@ ReadStream = (function (InternalReadStream) {
             chunk = chunk.slice(-n);
             var [_, ...rest] = arguments;
             this.pos = this.bytesRead;
-            if (this.end && this.bytesRead >= this.end) {
+            if (this.end !== undefined && this.bytesRead > this.end) {
               chunk = chunk.slice(0, this.end - this.start + 1);
             }
             return super.push(chunk, ...rest);
           }
           var end = this.end;
           // This is multi-chunk read case where we go passed the end of the what we want to read in the last chunk
-          if (end && this.bytesRead >= end) {
+          if (end !== undefined && this.bytesRead > end) {
             chunk = chunk.slice(0, end - currPos + 1);
             var [_, ...rest] = arguments;
             this.pos = this.bytesRead;
--- a/src/js/node/fs.js
+++ b/src/js/node/fs.js
@@ -541,14 +541,14 @@ ReadStream = (function (InternalReadStream) {
             var [_, ...rest] = arguments;
             this.pos = this.bytesRead;
             if (this.end && this.bytesRead >= this.end) {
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
i'm surprised there's an off-by-one error here but it also is consistent with the bug reports show.

you're sure it's not the += thing in http.write?

Hanaasagi:
I need some time to test this issue, but at least this is not related to `http.write`. You can run this code, and compare the output with node.js.

```JavaScript
const fs = require("node:fs");
const s = require("node:stream");

const path = "./test.txt";
fs.writeFileSync(path, "12345");

const stream = fs.createReadStream(path, { start: 0, end: 4 });

const writable = new s.Writable({
  write(chunk, encoding, callback) {
    console.log(chunk);
  },
});

stream.pipe(writable);
```

This won't execute any code related to http, just used `stream.pip`.

Hanaasagi:
Hi, here's an explanation of why the `createReadStream` should be fixed.

As mentioned in issue #3668, the `server-static` package of `express` depends on `send`, which uses `createReadStream` to create read streams.

https://github.com/pillarjs/send/blob/b69cbb3dc4c09c37917d08a4c13fcd1bac97ade5/index.js#L691-L704
https://github.com/pillarjs/send/blob/b69cbb3dc4c09c37917d08a4c13fcd1bac97ade5/index.js#L777-L793


According to the Node.js documentation, `end` refers to the offset, not the size.

Link: https://nodejs.org/api/fs.html#fscreatereadstreampath-options

> options can include start and end values to read a range of bytes from the file instead of the entire file. Both start and end are inclusive and start counting at 0, allowed values are in the [0, [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)] range.

Jarred-Sumner:
Thank you

Issue #3607, opened 2023-07-11
express.static returns error when serving an empty file


RobotoSkunk:
### What version of Bun is running?

0.6.13

### What platform is your computer?

Linux 6.3.0-1-amd64 x86_64 unknown

### What steps can reproduce the bug?

1. Create a new project and install express.
2. Configure express to serve static files with `express.static`.
3. Create an empty file (CSS, JS or whatever).
4. Run the project.
5. Load the file with an HTTP request.

### What is the expected behavior?

Express returning an empty HTTP response.

### What do you see instead?

The next error:
```
TypeError: |this| is not a Promise
    at then (native)
    at #internalRead (node:fs:333:13)
    at <anonymous> (node:stream:2:36318)
    at maybeReadMore (native)
    at performMicrotask (native)
```

### Additional information

I'm testing bun with this project: https://github.com/RobotoSkunk/personal-website/

You can easily reproduce this bug with `public/assets/css/stylesheet.css` file, writing content in it and leaving it empty like in the steps to reproduce the bug.

This bug is not that important because static files should not be empty, but it can be confusing for newcomers.

Hanaasagi:
Testing with bun 0.7.0, the initial requests are processed correctly, but subsequent identical requests result in errors.

![2023-07-29_00-00](https://github.com/oven-sh/bun/assets/9482395/401487e3-fa60-4562-9cc2-ad99c4d39b19)

- bun 0.7.0
- Linux 6.4.4-arch1-1 x86_64 unknown


Update:

When using the latest debug build, there might be a coredump generated.

![2023-07-29_23-23](https://github.com/oven-sh/bun/assets/9482395/46da26bc-8d35-44c0-bafd-c215ea6c62b5)

GDB stack trace

![2023-07-29_23-23_1](https://github.com/oven-sh/bun/assets/9482395/6583dd05-d296-437b-bafe-60bc31a46488)

PR #3882, opened 2023-07-29
Fix coredump when reading an empty file(`node:stream:createReadStream`)

Hanaasagi:
Close: #3607

### What does this PR do?

Fix coredump when reading an empty file.

- [x] Code changes

### Something helpful

Please check the call stack information for the coredump in the issue #3607 .

When reading a empty file by `readStream`:

https://github.com/oven-sh/bun/blob/681be10294c19b8ce402ec44df5cd6554e2c86c0/src/bun.js/webcore/streams.zig#L4435-L4438

`readable_file.start(&blob.data.file);` will call

https://github.com/oven-sh/bun/blob/681be10294c19b8ce402ec44df5cd6554e2c86c0/src/bun.js/webcore/streams.zig#L4026-L4036

And the result could be `.empty` (`stat.size` will be 0 when file is empty).

```zig
 if (result != .ready) { 
     return result; 
 } 
```

This would cause an early return, preventing `this.lazy_readable` from transitioning from `.bloc` to `.readable`.

I wrote automated tests

### DIFF
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -4433,7 +4433,7 @@ pub const FileReader = struct {
                     var readable_file = File{ .loop = this.globalThis().bunVM().eventLoop() };
 
                     const result = readable_file.start(&blob.data.file);
-                    if (result != .ready) {
+                    if (result != .ready and result != .empty) {
                         return result;
                     }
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (approved):
LGTM, thank you 

let’s run the tests and then merge

PR #3890, opened 2023-07-30
docs(bun/install): Mention {pre|post}prepare

### LINKED ISSUE #3874: It doesn't mention it will run `prepare` lifecycle script.

Documentation is missing

### What is the issue?

In https://bun.sh/docs/cli/install, it says "Run your project's {pre|post}install scripts at the appropriate time."

However, it fails to mention it will _also_ run `prepare` ([source](https://github.com/oven-sh/bun/blob/53cc4df191aa823c3a5c5a2a377701439a2d3f89/src/install/install.zig#L7946))

### Where did you find it?

https://bun.sh/docs/cli/install

v3nntrnx:
### What does this PR do?

Mention that `bun install` will also run `{pre|post}prepare` script. (As mentioned [here](https://github.com/oven-sh/bun/blob/53cc4df191aa823c3a5c5a2a377701439a2d3f89/src/install/install.zig#L7946))
Closes #3874

### DIFF
--- a/docs/cli/install.md
+++ b/docs/cli/install.md
@@ -40,7 +40,7 @@ On Linux, `bun install` tends to install packages 20-100x faster than `npm insta
 Running `bun install` will:
 
 - **Install** all `dependencies`, `devDependencies`, and `optionalDependencies`. Bun does not install `peerDependencies` by default.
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you

Issue #3896, opened 2023-07-30
Could not import WASI


sirenkovladd:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Linux 5.19.0-50-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
import * as wasi from "wasi";

console.log(wasi)
```

### What is the expected behavior?

```sh
❯ node wasi.js
[Module: null prototype] {
  WASI: [class WASI],
  default: { WASI: [class WASI] }
}
```

### What do you see instead?

```sh
❯ bun wasi.js 
1 | var nodeFsConstants = constants, __getOwnPropNames = Object.getOwnPropertyNames, __commonJS = (cb, mod) => function __require2() {
                         ^
ReferenceError: Can't find variable: constants
      at node:wasi:1:22
```

### Additional information

src/js/node/wasi.js
```js
// Hardcoded module "node:wasi"
// HUGE thanks to:
// - @williamstein and https://github.com/sagemathinc/cowasm/tree/main/core/wasi-js
// - @syrusakbary for wasmer-js https://github.com/wasmerio/wasmer-js
// - Gus Caplan for node-wasi https://github.com/devsnek/node-wasi
//
// Eventually we will implement this in native code, but this is just a quick hack to get WASI working.

/** **/
// constants is injected into the top of this file
const nodeFsConstants = constants;

var __getOwnPropNames = Object.getOwnPropertyNames;
```

https://bun.sh/docs/runtime/nodejs-apis#node_wasi

> [node:wasi](https://nodejs.org/api/wasi.html) 	🟡	Partially implemented.

but I can't even just import

paperclover:
this is going to be fixed in the next version of bun

PR #3897, opened 2023-07-30
fix node:wasi

sirenkovladd:
### What does this PR do?

Fix https://github.com/oven-sh/bun/issues/3896
add log [exports](https://github.com/sirenkovladd/bun/blob/08904eeeb84ca7ac0944b7e1f243bc99bf85a070/test/exports/bun-exports.bun-v0.7.2.json)

First, I tried adding `constants_string` in [module_loader.zig](src/bun.js/module_loader.zig) for `node:wasi` as in `node:fs/promises`, but that added the `constants` to the export in `wasi`

In my opinion, it would be better to just import `constant` from `fs`

### Documentation or TypeScript types
~Code changes~

~I ran `make js` and committed the transpiled changes~
I or my editor ran Prettier on the changed files (or I ran `bun fmt`)
I included a test for the new code, or an existing test covers it (test exports)


paperclover:
thanks for this pr, but i fixed this in my earlier pr for this that makes constants a native global, like `$processBindingContants.fs`

PR #3905, opened 2023-07-31
Update stric.md

aquapi:
### What does this PR do?

Just a quick PR to update Stric


colinhacks:
This guide got rewritten and moved under `docs/guides` since this PR. Sorry about that. Feel free to open another PR against the new guide. That said, I think the inclusion of `macro` is confusing, as its a feature that is not self-explanatory and relatively advanced. Best to stick with the most basic example, then link out to he docs directly for more interesting/advanced features.

Issue #2477, opened 2023-03-24
`request.clone()` -> `Request.prototype.json()` not working from HTTP server


Electroid:
Reported: https://discord.com/channels/876711213126520882/876711213126520885/1088916040387395614

The following snippet:
```js
Bun.serve({
  async fetch(request) {
    const req = request.clone();
    const b = await req.json();
    const a = await request.json();
    return new Response(`Bun!`);
  },
});
```

gives me this error with Bun v0.5.8 (and it worked with v0.5.7)
```
16 | Bun.serve({
17 |   async fetch(request) {
18 |     const req = request.clone();
                    ^
SyntaxError: Unexpected end of JSON input
```

LuisMalhadas:
This issue still exists with 0.5.9

PassiDel:
Still in 0.6.9.

This breaks json validation in hono :confused:  
honojs/middleware#81

otherguy:
Still an issue in 1.0.0

mirismaili:
Gives me the same error on the next line:
```
7 |   async fetch(request) {
8 |     const req = request.clone()
9 |     const b = await req.json()
                            ^
SyntaxError: Unexpected end of JSON input
```

Jarred-Sumner:
Duplicate of #6348

Jarred-Sumner:
This issue is being tracked in #6348 . We will fix it, but please follow along there.

Issue #1381, opened 2022-10-23
`Request.clone()` does not clone or support streaming body


Electroid:
### What version of Bun is running?

0.2.1 (canary)

### What platform is your computer?

Darwin 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 20:57:23 PDT 2022; root:xnu-8020.121.3~2/RELEASE_ARM64_T8110 arm64 arm

### What steps can reproduce the bug?

```ts
export default {
  async fetch(request) {
    console.log(".clone().json()", await request.clone().json());
    console.log(".json()", await request.json());
    return new Response("Ok");
  }
}
```

`curl -d '{}' http://localhost:3000`

### How often does it reproduce? Is there a required condition?

Always. Same with `.text()`.

### What do you see instead?

Terminal output:
```bash
.clone().json() # body is missing
.json() {} # works fine
```

Jarred-Sumner:
Need to do .tee() on ReadableStream. It clones it for non-streaming bodies (same for Response)

Electroid:
You can use `.tee()` but this should still work, see example [here](https://cloudflareworkers.com/#2333e8d02dc95fdc8518f898d772b00d:https://tutorial.cloudflareworkers.com/). (send a POST request with "{}" body)

Jarred-Sumner:
Sorry - I meant Bun needs to internally call tee(). It is a bug that this doesn't work right now

JustGreg:
is there any progress on this issue?

> https://github.com/oven-sh/bun/issues/2477
> https://github.com/honojs/middleware/issues/81#issuecomment-1509457656
> https://github.com/oven-sh/bun/pull/1693

beorn:
This is causing all kinds of weird (and hard-to-track down) problems for people that do `new Request(req)` type stuff.  Would really appreciate it being prioritized :-|

PR #1693, opened 2023-01-01
feat(web api): support request.clone()

zhiyuang:
Support request.clone

fix https://github.com/oven-sh/bun/issues/1381

PR #3908, opened 2023-07-31
Support streams in request.clone() & response.clone()

Jarred-Sumner:
### What does this PR do?

This introduces support for `Request` and `Response` bodies to be cloned when the body is streamed.

```js
const server = Bun.serve({
  port: 3000,
  hostname: "localhost",
  error(e) {
    console.error(e);
  },
  async fetch(req) {
    if (req.method === "POST") {
      try {
        const cloned = req.clone();
        console.log(await cloned.text());

        return new Response(await req.text());
      } catch (e) {
        console.log(e);
      }
    }
    return new Response("Hello, world! " + req.headers.get("huge-number"));
  },
});
```

This also updates our .tee() function to match https://github.com/oven-sh/WebKit/commit/221bafcd1af8f91209af9effef02b941b69d584d

Fixes https://github.com/oven-sh/bun/issues/2477
Fixes https://github.com/oven-sh/bun/issues/1381
Fixes https://github.com/oven-sh/bun/issues/2935

 I wrote automated tests


Jarred-Sumner:
The tests from #1693 highlighted a memory or GC issue that needs to be addressed. can't merge this PR until then. In the meantime though, I'll pull out the other two separate fixes since those are much safer

Jarred-Sumner:
Superseded by https://github.com/oven-sh/bun/pull/12473

Issue #485, opened 2022-07-09
`bun create` to be compatible with `npm create`


gi-dellav:
The feature that I'm proposing is the support for `bun create` to create project using NPM packages.

For example, with npm this `npm create vite@latest` can be used to create a project using the latest version of vite, while bun doesn't support npm for his `create` command.

I realize that this project is still in its infancy, but I wanted to say to every developer of the project that your work is truly amazing.

birkskyum:
`npm create` is a simple alias for `npx create-`, so if you write `npx create-vite@latest` it's the same - and that you can do with bun already:

`bun x create-vite@latest`, or for now to make sure you use bun `bun --bun x create-vite@latest`

  `bun create` is not a simple alias like that, and does it's on thing entirely... would arguably be a lot less confusion if it was a simple alias in the same way that npm is, and that the current `bun create` is renamed to something else that isn't overloaded like `bun generate` or `new`, `make`, `example`, `begin`. Since the files comes from the `examples` folder, maybe `example` is the most obvious.
  
  @Jarred-Sumner , if the aim is npm compat, maybe this is worth considering before the 1.0 locks down the APIs. Many many docs use the `npm create` shorthand, and therefore people will try to run `bun create` and get unexpected outcome
  
Alternatively, a new create script for these examples could be made on par with all the other ones, so that it would be executed with `bun x create-bun-example`, and due to the alias it would then be `bun create bun-example`, and then select i.e. react in the cli.

PR #3910, opened 2023-07-31
Implement `bun start`

Jarred-Sumner:
### What does this PR do?

I don't know if we should ship this

This makes `bun start` work like `bun .`, except if there is a `server.js` or `server.ts` file, it will load that instead.

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -893,7 +893,7 @@ pub const RunCommand = struct {
         }
 
         if (log_errors or force_using_bun) {
-            if (script_name_to_search.len > 0) {
+            if (script_name_to_search.len > 0 and !strings.eqlComptime(script_name_to_search, "start")) {
                 possibly_open_with_bun_js: {
                     const ext = std.fs.path.extension(script_name_to_search);
                     var has_loader = false;
@@ -998,6 +998,7 @@ pub const RunCommand = struct {
         var this_bundler: bundler.Bundler = undefined;
         var root_dir_info = try configureEnvForRun(ctx, &this_bundler, null, &ORIGINAL_PATH, log_errors, force_using_bun);
         this_bundler.env.map.put("npm_lifecycle_event", script_name_to_search) catch unreachable;
+
         if (root_dir_info.enclosing_package_json) |package_json| {
             if (package_json.scripts) |scripts| {
                 switch (script_name_to_search.len) {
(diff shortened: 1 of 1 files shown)


paperclover (approved):
cool so long as `bun start` can still run package.json start script, along with `bun start args to start script` also does that.

remorses:
I don’t like adding 1 more command just to mimic npm, I have never seen someone using npm start to run server.js, most people also define a script with that name, I think it’s a legacy command no one uses anymore

birkskyum:
It's starting to become confusing with all these overloaded names of standard package.json scripts/node conventions, like [bun dev](https://github.com/oven-sh/bun/issues/4166#issue-1851743731), [bun start](https://github.com/oven-sh/bun/pull/3910#issuecomment-1697661387), `bun test`, `bun build`, not to mention the overloaded [bun create](https://github.com/oven-sh/bun/issues/485), because these names are so canonical many people will assume they map directly to `bun run __` or `bun x create-__`, but they do something very different. If the overloads could be kept to a minimum I'm sure it would be appreciated.

birkskyum:
- `bun build` "_bundle a project_", then maybe it should be called `bun bundle` which isn't overloaded
- `bun start` is longer then `bun .`, so it's really not needed. Also `bun` is both `node` and `npm`, and where `node start` is like `node .`, the `npm start` is like `npm run start`, so nobody moving from node/npm can intuitively tell what will happen when executing `bun start`, and not having it is thus a less ambiguous approach.
- `bun create` could be something like `generate`, `new`, `make`, `example`, `begin`.
Screenshot 2023-08-29 at 14 14 55

Mouvedia:
The shorthand `npm start` should be matched by `bun start` which would alias `bun run start`.
**`npm start` and `npm test` are the most used shorthand scripts by far.**

If you do add `start` _for that purpose_ remember to add its less common counterparts, `restart` and `stop`, and their `pre/post` scripts.

n7trd:
I agree with @birkskyum that it can become confusing.

I'd much rather run scripts with:
`bun run <script>` 
This approach makes the intend clear and obviates the need to remember any conflicting script names correlated to bun commands. It would be a lot clearer without reading advices like:

> If there is a name conflict between a package.json script and a built-in bun command (install, dev, upgrade, etc.) Bun's built-in command takes precedence. In this case, use the more explicit bun run command to execute your package script.

The confusion between `bun build` and `bun run build` is worth addressing. In my opinion, `bun bundle` aptly describes the action and is relatively less commonly used in package scripts, which works as an advantage.

Overall, I appreciate the ability to run (node) code without making changes to it. However, I don't believe this applies to the the CLI for me. They are distinct toolkits, and tweaking a few files to achieve your end goal doesn't seem too onerous. That's why I'm not entirely convinced by the bunx, which somewhat imitate node and tend to overlap with "bun create". But that's another matter altogether.

Electroid:
We're not going to implement this now, we may revisit it later.

Issue #3912, opened 2023-07-31
If I try to import a node module (`node:*` ) that doesn't exist, I don't get the error


sirenkovladd:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Linux 5.19.0-50-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
const net = require('node:net1');
// or
import net from 'node:net1';

console.log(net, typeof net === 'function' && net());
```

### What is the expected behavior?

```sh
❯ node index.js
node:internal/modules/cjs/loader:916
      throw new ERR_UNKNOWN_BUILTIN_MODULE(request);
      ^

Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:net1
    at new NodeError (node:internal/errors:405:5)
    at Module._load (node:internal/modules/cjs/loader:916:13)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at Object.<anonymous> (/home/sirenko/programming/bun-test/index.js:1:13)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12) {
  code: 'ERR_UNKNOWN_BUILTIN_MODULE'
}
```

### What do you see instead?

```sh
❯ bun index.js
[Function] {}
```

PR #3913, opened 2023-07-31
throw error if node module does not exist

sirenkovladd:
### What does this PR do?

An error will occur when you try to import a built-in module that does not exist
Fixes https://github.com/oven-sh/bun/issues/3912

I or my editor ran Prettier on the changed files (or I ran `bun fmt`)
I included a test for the new code, or an existing test covers it

### Zig files changed:

I or my editor ran `zig fmt` on the changed files
I included a test for the new code, or an existing test covers it

### DIFF
--- a/test/js/node/missing-module.test.js
+++ b/test/js/node/missing-module.test.js
@@ -1,0 +1,7 @@
+import { expect, test } from "bun:test";
+
+test("not implemented yet module masquerades as undefined and throws an error", () => {
+  const missingModule = "node:missing" + "";
+  expect(() => require(missingModule)).toThrow(/^Cannot find package "node:missing" from "/);
+  expect(() => import(missingModule)).toThrow(/^Cannot find package "node:missing" from "/);
+});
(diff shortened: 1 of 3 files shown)


v3nntrnx:
lgtm

test/js/node/missing-module.test.js
```diff
@@ -0,0 +1,7 @@
+import { expect, test } from "bun:test";
+
+test("not implemented yet module masquerades as undefined and throws an error", () => {
+  const missingModule = "node:missing" + "";
```
Jarred-Sumner, line 4: surprising that transpiler does't inline this

test/js/node/missing-module.test.js
```diff
@@ -0,0 +1,7 @@
+import { expect, test } from "bun:test";
+
+test("not implemented yet module masquerades as undefined and throws an error", () => {
```
Jarred-Sumner, line 3: it doesn't masqsuerade as undefined, masquerade as undefined is a function which is callable but typeof returns `"undefined"`

Jarred-Sumner:
Thank you. The test is mislabeled but it's better than status quo so I merged it

test/js/node/missing-module.test.js
sirenkovladd: I did concat because in `import(missingModule)` without concat I get an error when running the script test

PR #3915, opened 2023-08-01
prefer `module` in `exports` field

### LINKED ISSUE #3901: bun is loading the ESM version instead of the CommonJS version

0.7.1

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 unknown

```ts
import { StoreRegistry } from "@sapphire/pieces";

const store = new StoreRegistry();
await store.load();
```

It must be not show the error like node does

### What do you see instead?

image

### Additional information

_No response_

dylan-conway:
### What does this PR do?

This pr will ensure the `module` field in `exports` does not resolve before `require`, `import`, or `default` for runtime imports.

I wrote tests to run the same code bundled and unbundled and compared the output to make sure the correct files were imported.

fixes #3901

### DIFF
--- a/src/resolver/package_json.zig
+++ b/src/resolver/package_json.zig
@@ -1248,6 +1248,7 @@ pub const ESModule = struct {
     conditions: ConditionsMap,
     allocator: std.mem.Allocator,
     module_type: *options.ModuleType = undefined,
+    prefer_module_field_in_exports: bool,
 
     pub const Resolution = struct {
         status: Status = Status.Undefined,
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
Could/should we just ignore the module field at runtime?

Jarred-Sumner:
Do we need to make the size of PackageJSON larger? We never really free those so any size change will impact minimum memory usage

dylan-conway:
> Could/should we just ignore the module field at runtime?

I think it makes sense to ignore because that's what node does

Issue #3116, opened 2023-05-30
Http upstream encoding problem


wycers:
### What version of Bun is running?

0.6.5_debug

### What platform is your computer?

Linux 6.2.15-200.fc37.x86_64 x86_64 x86_64

### What steps can reproduce the bug?

1. generate a file filled with hex > 0x80
2. upload this file with formdata/multipart

### What is the expected behavior?

The server should receive a same file

### What do you see instead?

The server will receive a file, but hex in this file are replaced with `0xef 0xbf 0xbd`

wycers:
And this issue may be related to #3085 and #2026.

I have dived into this bug for a little bit. I believe the problem is located at `src/bun.js/http.exports.js`. Maybe I can take up this and make a pr.

wycers:
The root cause is [src/bun.js/http.exports.js#L1191](https://github.com/oven-sh/bun/blob/756e15f6e283fd7e21fe91e060d63440ce5c72de/src/bun.js/http.exports.js#L1191), the add operation implicitly calls Buffer.toString() and turns invalid `utf-8` bytes into `U+FFFD(�)`.

sp90:
@Wycers any news still seems to be an issue in the canary version of bun :)

Hanaasagi:
Hi, I'm trying to reproduce this problem.

> The root cause is [src/bun.js/http.exports.js#L1191](https://github.com/oven-sh/bun/blob/756e15f6e283fd7e21fe91e060d63440ce5c72de/src/bun.js/http.exports.js#L1191), the add operation implicitly calls Buffer.toString() and turns invalid utf-8 bytes into U+FFFD(�).

Based on your previous debugging, I guess you are referring to an issue on the client side. However, I couldn't reproduce locally, or perhaps I misunderstood something or there might be some problem with the steps.


1. Create binary file.
```
$ echo -n -e '\x80\x81\x82\x83' > byteFile
```

2. Client Code

```JavaScript
const fs = require("fs");
const http = require("http");

function uploadFile(filepath, url) {
  const fileStream = fs.createReadStream(filepath);

  const options = {
    method: "POST",
    headers: {
      "Content-Type": "multipart/form-data",
      "Content-Length": fs.statSync(filepath).size,
    },
  };

  const req = http.request(url, options, res => {
    res.on("data", chunk => {
      console.log(`BODY: ${chunk}`);
    });
    res.on("end", () => { });
  });

  req.on("error", e => {
    console.log("error", e);
  });

[...]

PR #3916, opened 2023-08-01
Fix encoding problem when uploading a binary file (`http.ClientRequest.write`)

Hanaasagi:
Close: #3116

### What does this PR do?

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -1214,7 +1214,7 @@ export class ClientRequest extends OutgoingMessage {
   #path;
   #socketPath;
 
-  #body: Buffer | null = null;
+  #bodyChunks: Buffer[] | null = null;
   #fetchRequest;
   #signal: AbortSignal | null = null;
   [kAbortController]: AbortController | null = null;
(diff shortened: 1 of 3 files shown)


src/js/node/http.ts
```diff
@@ -1249,18 +1249,18 @@ export class ClientRequest extends OutgoingMessage {
       callback();
       return;
     }
-    this.#body = body + chunk;
+    this.#body = Buffer.concat([body, chunk]);
```
Hanaasagi: I see that the `chunk` has already been converted to the `Buffer` type here.

https://github.com/oven-sh/bun/blob/7a8f57c4e527f7fc84b85a4757723bb97540ebc3/src/js/node/stream.js#L3573-L3589

Jarred-Sumner: This will make node:http a lot slower.

Can we avoid doing this when the input is a string or properly-encoded string? Maybe we can use BufferList for this?

Hanaasagi: Fixed.

src/js/out/modules/node/http.js
```diff
@@ -834,27 +834,25 @@ class ClientRequest extends OutgoingMessage {
...
-    this.#body = body + chunks.join(), callback();
+    this.#bodyChunks.push(...chunks), callback();
   }
   _final(callback) {
     if (this.#finished = !0, this[kAbortController] = new AbortController, this[kAbortController].signal.addEventListener("abort", () => {
       this[kClearTimeout]();
     }), this.#signal?.aborted)
       this[kAbortController].abort();
-    var method = this.#method, body = this.#body;
+    var method = this.#method, body = Buffer.concat(this.#bodyChunks || []);
```
Jarred-Sumner, line 855: nit: can we add a check for if the number of chunks was exactly 1 and if so, use the existing hcunk instead of creating a new one?

paperclover:
landed in 3939

src/js/out/modules/node/http.js
Hanaasagi: Fixed in https://github.com/oven-sh/bun/pull/3946

Issue #3911, opened 2023-07-31
Private key signing errors; private key string is prepended the letter 'n' on every line


chenglou:
### What version of Bun is running?

0.7.2 (canary)

### What platform is your computer?

macOS

### What steps can reproduce the bug?

See below. I'm kinda positive there's some line breaking escaping issue going on here

### What is the expected behavior?

Successful key signing

### What do you see instead?

<img width="107" alt="Screenshot 2023-07-31 at 08 05 22" src="https://github.com/oven-sh/bun/assets/1909539/782b1f9d-ef67-4a17-b5e0-2c9baf5b5a34">

Stack trace:

```
- error Error: error:0908F066:PEM routines:get_header_and_data:bad end line
    at Sign.sign (node:internal/crypto/sig:131:29)
    at Object.sign (/Users/chenglou/github/gallery/node_modules/jwa/index.js:153:45)
    at Object.jwsSign [as sign] (/Users/chenglou/github/gallery/node_modules/jws/lib/sign-stream.js:32:24)
    at GoogleToken.requestToken (/Users/chenglou/github/gallery/node_modules/gtoken/build/src/index.js:232:31)
    at GoogleToken.getTokenAsyncInner (/Users/chenglou/github/gallery/node_modules/gtoken/build/src/index.js:166:21)
    at GoogleToken.getTokenAsync (/Users/chenglou/github/gallery/node_modules/gtoken/build/src/index.js:145:55)
    at GoogleToken.getToken (/Users/chenglou/github/gallery/node_modules/gtoken/build/src/index.js:97:21)
    at JWT.refreshTokenNoCache (/Users/chenglou/github/gallery/node_modules/google-auth-library/build/src/auth/jwtclient.js:172:36)
    at JWT.refreshToken (/Users/chenglou/github/gallery/node_modules/google-auth-library/build/src/auth/oauth2client.js:153:24)
    at JWT.getRequestMetadataAsync (/Users/chenglou/github/gallery/node_modules/google-auth-library/build/src/auth/oauth2client.js:298:28)
    at JWT.getRequestMetadataAsync (/Users/chenglou/github/gallery/node_modules/google-auth-library/build/src/auth/jwtclient.js:94:26)
    at JWT.getRequestHeaders (/Users/chenglou/github/gallery/node_modules/google-auth-library/build/src/auth/oauth2client.js:261:37)
[...]

PR #3918, opened 2023-08-01
Fix expanding on .env files

cirospaciari:
### What does this PR do?

- [x] Code changes

I wrote automated tests

Fix: https://github.com/oven-sh/bun/issues/3911

### DIFF
--- a/test/cli/run/env.test.ts
+++ b/test/cli/run/env.test.ts
@@ -338,3 +338,12 @@ test(".env in a folder doesn't throw an error", () => {
   const { stdout } = bunRun(`${dir}/index.ts`);
   expect(stdout).toBe("hey");
 });
+
+test("#3911", () => {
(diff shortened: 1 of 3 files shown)


test/cli/run/env.test.ts
```diff
@@ -338,3 +338,12 @@ test(".env in a folder doesn't throw an error", () => {
   const { stdout } = bunRun(`${dir}/index.ts`);
   expect(stdout).toBe("hey");
 });
+
+test("#3911", () => {
```
Jarred-Sumner, line 359: Can you add a test for a bounds check here? That if it’s \\n it doesn’t go out of bounds

cirospaciari: sure

Jarred-Sumner:
Thank you

PR #3920, opened 2023-08-01
Fix path_watcher

cirospaciari:
### What does this PR do?

- [x] Code changes

Existing tests

### DIFF
--- a/src/bun.js/node/path_watcher.zig
+++ b/src/bun.js/node/path_watcher.zig
@@ -25,7 +25,7 @@ pub const PathWatcherManager = struct {
     const options = @import("../../options.zig");
     pub const Watcher = GenericWatcher.NewWatcher(*PathWatcherManager);
     const log = Output.scoped(.PathWatcherManager, false);
-    main_watcher: *Watcher,
+    main_watcher: ?*Watcher,
 
     watchers: bun.BabyList(?*PathWatcher) = .{},
     watcher_count: u32 = 0,
@@ -104,22 +104,23 @@ pub const PathWatcherManager = struct {
             return err;
         };
         errdefer watchers.deinitWithAllocator(bun.default_allocator);
+        const main_watcher = try Watcher.init(
+            this,
+            vm.bundler.fs,
+            bun.default_allocator,
+        );
         var manager = PathWatcherManager{
             .file_paths = bun.StringHashMap(PathInfo).init(bun.default_allocator),
             .current_fd_task = bun.FDHashMap(*DirectoryRegisterTask).init(bun.default_allocator),
             .watchers = watchers,
-            .main_watcher = try Watcher.init(
-                this,
-                vm.bundler.fs,
-                bun.default_allocator,
-            ),
+            .main_watcher = main_watcher,
             .vm = vm,
             .watcher_count = 0,
             .mutex = Mutex.init(),
         };
 
         this.* = manager;
-        try this.main_watcher.start();
+        try main_watcher.start();
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner (changes requested):
This does not fix it? The `finalized` field is set to true in the same scope  as the struct is destroyed, which means all cases the struct  is used after finalized is a use after free. 

https://github.com/oven-sh/bun/blob/044a6facdb3dd624b65960fd390de24980df4989/src/bun.js/node/path_watcher.zig#L816

cirospaciari:
> finalized

It's only used when `pending_directories > 0` (on another thread) finalized just say that the owner is gone, when the last task ends/cancel it will call deinit again (but not emit/flush).

Jarred-Sumner:
It's not threadsafe though. The finalized boolean would need to be loaded and stored atomically with a @fence to ensure memory ordering. But even that's not safe because the other thread could be accessing it after deinit

cirospaciari:
> It's not threadsafe though. The finalized boolean would need to be loaded and stored atomically with a @Fence to ensure memory ordering. But even that's not safe because the other thread could be accessing it after deinit

I used mutex to guarantee access to `finalized` and `pending_directories` but I found another mistake let me finish it, and you are right about memory ordering

Jarred-Sumner:
Not all usages are locked by a mutex. If it were, it would hang until after it’s freed and then segfault

src/bun.js/node/path_watcher.zig
```diff
@@ -431,19 +458,20 @@ pub const PathWatcherManager = struct {
...
-                    watcher.mutex.unlock();
-                    manager._decrementPathRef(entry_path_z);
-                    return err;
-                };
-                watcher.mutex.unlock();
+                {
+                    watcher.mutex.lock();
+                    defer watcher.mutex.unlock();
+                    watcher.file_paths.push(bun.default_allocator, child_path.path) catch |err| {
+                        manager._decrementPathRef(entry_path_z);
```
Jarred-Sumner, line 465: dead lock

src/bun.js/node/path_watcher.zig
```diff
@@ -338,46 +365,57 @@ pub const PathWatcherManager = struct {
...
+            var routine: *DirectoryRegisterTask = undefined;
+            {
+                manager.mutex.lock();
+                defer manager.mutex.unlock();
+
+                // use the same thread for the same fd to avoid race conditions
+                if (manager.current_fd_task.getEntry(path.fd)) |entry| {
+                    routine = entry.value_ptr.*;
+
+                    if (watcher.refPendingDirectory()) {
```
Jarred-Sumner, line 380: dead lock

src/bun.js/node/path_watcher.zig
```diff
@@ -338,46 +365,57 @@ pub const PathWatcherManager = struct {
...
+                routine.* = DirectoryRegisterTask{
+                    .manager = manager,
+                    .path = path,
+                    .watcher_list = bun.BabyList(*PathWatcher).initCapacity(bun.default_allocator, 1) catch |err| {
+                        bun.default_allocator.destroy(routine);
+                        return err;
+                    },
+                };
+                errdefer routine.deinit();
+                if (watcher.refPendingDirectory()) {
```
Jarred-Sumner, line 401: dead lock

src/bun.js/node/path_watcher.zig
```diff
@@ -338,46 +365,57 @@ pub const PathWatcherManager = struct {
...
+                    };
+                } else {
+                    return error.UnexpectedFailure;
+                }
+                manager.current_fd_task.put(path.fd, routine) catch |err| {
+                    watcher.unrefPendingDirectory();
                     return err;
                 };
+            }
+            if (manager.refPendingTask()) {
```
Jarred-Sumner, line 414: dead lock

src/bun.js/node/path_watcher.zig
cirospaciari: manager uses a different lock from watcher

src/bun.js/node/path_watcher.zig
cirospaciari: unlock with defer after last context exits

Jarred-Sumner:
Thank you

Issue #3862, opened 2023-07-28
*Regression* Segmentation fault when using `Bun.spawn` and `fetch`


GingerAdonis:
### What version of Bun is running?

- Latest affected and tested version: `v0.7.1`
- Affected releases: `>= v0.6.10`
- Unaffected releases: `<= v0.6.9`

### What platform is your computer?

Linux 6.3.12-200.fc38.x86_64 x86_64

### What steps can reproduce the bug?

**Make sure _docker_ is available to execute.**

⚠️⚠️ The following script will create files in the current working directory and create docker containers! ⚠️⚠️

Run the following script:
```ts
import { hostname } from 'node:os';
import { sleep } from 'bun';

const filesCount = 15;
console.log(`This script will create ${filesCount} bin.gz files and call a webhook when done`);

for (let i = 0; i < filesCount; i++) {
  const filename = `archive${i}.bin.gz`;

  const proc = Bun.spawn(
    [
      'bash',
      '-c',
      `docker run ubuntu head -c 25M /dev/urandom | gzip > ${filename}`
    ],
    {
      stderr: 'inherit',
      stdout: 'inherit'
    }
  );

  const exitCode = await proc.exited;
  if (exitCode !== 0) {
    throw new Error('Process exited with non-zero exit code');
  }

  console.log(`✅ Created file ${filename}`);
  await sleep(200);
}

console.log('Created all files. Waiting 2 seconds before sending webhook...');
await sleep(2000);
console.log('Sending webhook...');

{
  const data = {
    content: `Random content @ ${new Date().toLocaleString('en-GB')}`,
    username: `User @ ${hostname()}`
  };

  const result = await fetch('https://httpbin.org/post', {
    body: JSON.stringify(data),
    headers: {
      // eslint-disable-next-line @typescript-eslint/naming-convention
      'Content-Type': 'application/json'
    },
    method: 'POST'
  });

  console.log('Webhook result', result);
}

```

### What is the expected behavior?

Script successfully executes.

### What do you see instead?

A segfault. :sad

```
This script will create 15 bin.gz files and call a webhook when done
✅ Created file archive0.bin.gz
✅ Created file archive1.bin.gz
✅ Created file archive2.bin.gz
✅ Created file archive3.bin.gz
✅ Created file archive4.bin.gz
✅ Created file archive5.bin.gz
✅ Created file archive6.bin.gz
✅ Created file archive7.bin.gz
✅ Created file archive8.bin.gz
[...]

PR #3929, opened 2023-08-02
fix(fetch) clean declare structs

cirospaciari:
### What does this PR do?

- [x] Code changes

Run https://github.com/oven-sh/bun/issues/3862 test on release build with `Bun.gc(true)` before fetching

Fix: https://github.com/oven-sh/bun/issues/3862


Jarred-Sumner:
Very strange that this fixes anything at all

PR #3945, opened 2023-08-03
[install] fix stale `bun.lockb` on workspaces

alexlamsl:
- handle workspaces under `--production` correctly
- fix `Makefile`

### What does this PR do?

- [x] Code changes

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -2385,25 +2385,35 @@ pub const Package = extern struct {
         };
 
         pub fn generate(
-            _: Allocator,
+            allocator: Allocator,
+            log: *logger.Log,
             from_lockfile: *Lockfile,
             to_lockfile: *Lockfile,
             from: *Lockfile.Package,
             to: *Lockfile.Package,
-            mapping: []PackageID,
+            id_mapping: ?[]PackageID,
         ) !Summary {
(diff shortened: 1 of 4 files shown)


alexlamsl:
not sure if these are intentional or just leftovers of a refactoring exercise (`js` vs. `esm`):
https://github.com/oven-sh/bun/blob/928816bef0945e9655c115767c719d6865ff138c/Makefile#L557-L562

src/install/lockfile.zig
```diff
@@ -2385,37 +2385,83 @@ pub const Package = extern struct {
...
+                                const bytes = try file.readToEndAlloc(allocator, std.math.maxInt(usize));
+                                defer allocator.free(bytes);
+                                const source = logger.Source.initPathString(path, bytes);
+
+                                var workspace = Package{};
+                                try workspace.parseMain(to_lockfile, allocator, log, source, Features.workspace);
+
+                                var from_pkg = from_lockfile.packages.get(from_resolutions[i]);
+                                const diff = try generate(
+                                    allocator,
```
Jarred-Sumner, line 2446: at some point we should look into doing this in parallel, on large workspaces this would be a significant perf hit

Issue #3948, opened 2023-08-03
Encoding problem in `path.extname`.


Hanaasagi:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Linux 6.4.7-arch1-2 x86_64 unknown

### What steps can reproduce the bug?

```
const path = require("path");
const res = path.posix.extname("hello.🔥");
console.log(res);
```

### What is the expected behavior?

```
.🔥
```

### What do you see instead?

```
.ð¥
```

### Additional information

Mojo lang use the 🔥 as the file extension.

Ref: https://docs.modular.com/mojo/programming-manual.html

PR #3949, opened 2023-08-03
Fix the value encoding in `path.extname`.

Hanaasagi:
Close: #3948

### What does this PR do?

- Fix the encoding issue in the return value of `path.extname`.

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
Thank you

Issue #3924, opened 2023-08-02
path.relative edge case


paperclover:
### What steps can reproduce the bug?

`path.relative('/webpack-hot-middleware', '/webpack/buildin/module.js')`

<img width="413" alt="image" src="https://github.com/oven-sh/bun/assets/24465214/2417d4fd-8f6d-4228-abe1-2f55efc2acf7">

PR #3952, opened 2023-08-03
Fix edge case in `path.relative`.

Hanaasagi:
Close: #3924

### What does this PR do?

- `path.relative("/webpack-hot-middleware", "/webpack/buildin/module.js")` should return `../webpack/buildin/module.js`

### Some explanations.

Assuming the input is as follows:

- `input[0]` => `/webpack-hot-middleware`
- `input[1]` => `/webpack/buildin/module.js`

https://github.com/oven-sh/bun/blob/928816bef0945e9655c115767c719d6865ff138c/src/resolver/resolve_path.zig#L198-L207

The longest common substring is `/webpack`, So the index is `8` (point to `-` or `/`).

But here 

https://github.com/oven-sh/bun/blob/928816bef0945e9655c115767c719d6865ff138c/src/resolver/resolve_path.zig#L297-L303

we will get `/webpack/` here, a wrong directory.

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
Thank you! Let's see the tests run if anything interesting happens. Changes to path resolution often have a lot of unexpected side effects, though nowadays we have much better test coverage so I expect any issues to come up in tests

PR #3966, opened 2023-08-04
Remove native events from streams

paperclover:
### What does this PR do?

Removes `$lazy("events")` and `$lazy("bun:stream")`

- [X] Code changes


paperclover:
there is a failing test with http that is unclear what is going on.

PR #3967, opened 2023-08-04
fix macro string escaping

### LINKED ISSUE #3957: macro serialization broken for when returning escaped string

0.7.1

Darwin 22.5.0 arm64 arm

see the screenshot for a complete repro

image

it should not fail (it works if I remove escaping) and I think it should not even un-escape the saved result/ast - or at least it should definitely not do so for `\\`

### What do you see instead?

`bun run` fails

### Additional information

@dylan-conway

dylan-conway:
### What does this PR do?

strings passed as arguments to macros and strings returned from macros are escaped correctly

added macro tests for escape characters

fixes #3957

### DIFF
--- a/src/js_lexer.zig
+++ b/src/js_lexer.zig
@@ -75,6 +75,19 @@ pub const JSONOptions = struct {
     was_originally_macro: bool = false,
 };
 
+pub fn decodeUTF8(bytes: string, allocator: std.mem.Allocator) ![]const u16 {
+    var log = logger.Log.init(allocator);
+    defer log.deinit();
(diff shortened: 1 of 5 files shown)


src/js_lexer.zig
```diff
@@ -75,6 +75,19 @@ pub const JSONOptions = struct {
     was_originally_macro: bool = false,
 };
 
+pub fn decodeUTF8(bytes: string, allocator: std.mem.Allocator) ![]const u16 {
```
Jarred-Sumner, line 78: `decodeWTF8`

PR #3971, opened 2023-08-04
Buffer.copy should ignore out-of-range sourceEnd

### LINKED ISSUE #3947: `Buffer.copy` return value is wrong when `sourceEnd` is greater than actual length.

0.7.0

Linux 6.4.7-arch1-2 x86_64 unknown

```JavaScript
const assert = require("assert");

const src = Buffer.allocUnsafe(8);
const target = Buffer.allocUnsafe(16);

const copied = src.copy(target, 0, 0, src.length + 1); // +1 here
assert.strictEqual(src.length, 8);
assert.strictEqual(target.length, 16);
assert.strictEqual(copied, src.length);
```

No assertion error.

### What do you see instead?

```
AssertionError: Expected values to be strictly equal:

9 !== 8

 code: "ERR_ASSERTION"
```

### Additional information

This should return the actual number of elements copied. `src` only has 8 elements.

yifeiw3:
### What does this PR do?

fixes #3947 

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/bindings/JSBuffer.cpp
+++ b/src/bun.js/bindings/JSBuffer.cpp
@@ -949,7 +949,8 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunction_copyBody(JSC::JSGlob
     }
 
     targetStart = std::min(targetStart, targetEnd);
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you!

PR #3973, opened 2023-08-04
FFI typo

paperclover:
image

I cant believe tests did not catch this. This is my biggest mistake of the century.

I wrote a new test for specifically making sure this never happens again.


Jarred-Sumner:
dave can you fix prettier

paperclover:
i swear i turned on save on format. seems my vscode config is sufficiently broken

PR #3974, opened 2023-08-04
[install] store resolved workspace path in lockfile

### LINKED ISSUE #3685: Fresh installation in a workspace fails with missing linked packages

v0.6.15 (17903570)

Darwin 21.6.0 arm64 arm

Create a file `package.json` with:
```
{
  "name": "bun-workspaces",
  "module": "index.ts",
  "type": "module",
  "workspaces": [
    "packages/a"
  ],
  "dependencies": {
    "a": "workspace:*"
  }
}
```

Create a second file `packages/a/package.json` with:

```
{
  "name": "a",
  "version": "1.0.0"
}
```

Then run `bun install` in the root. This will correctly create `node_modules` with a linked package `a`, and a `bun.lockb` file.

Then simulate a fresh install: remove `node_modules` and run `bun install` again (`rm -fr **/node_modules; bun install`).

When `bun.lockb` is committed to the repo and the project is a monorepo with linked workspaces, running `bun install` should just work.

### What do you see instead?

Instead, the second time you run it (i.e. with `bun.lockb` present but deleted `node_modules`) you get:

```
bun install v0.6.15 (17903570)


error: package "a" is not linked

To install a linked package:
   bun link my-pkg-name-from-package-json

Tip: the package name is from package.json, which can differ from the folder name.
```

### Additional information

image

### LINKED ISSUE #3958: Error installing packages from workspaces when using workspace protocol

0.7.1

Darwin 21.6.0 arm64 arm

Create a `package.json` with

```
{
  "name": "bun-workspaces",
  "module": "index.ts",
  "type": "module",
  "workspaces": [
    "packages/a"
  ],
  "dependencies": {
    "a": "workspace:*"
  }
}
```

And a `packages/a/package.json` with:

```
{
  "name": "a",
  "version": "1.0.0"
}
```

Then run `bun install`.

It installs without errors. `node_modules/a` is linked to `packages/a`.

### What do you see instead?

Got an error:

```
bun install v0.7.1 (53cc4df1)


error: workspace dependency "a" not found

Searched in "./*"

Workspace documentation: https://bun.sh/docs/install/workspaces
```

### Additional information

Looks related to the `"workspace:*"` dependency, because depending on just `"*"` seems to work.

alexlamsl:
- resolve cwd consistently
- speed up package diff
- update `test/bun.lockb`

fixes #3685
fixes #3958

### What does this PR do?

- [x] Code changes

I wrote automated tests

If Zig files changed:


src/install/lockfile.zig
```diff
@@ -2651,20 +2648,22 @@ pub const Package = extern struct {
                             ),
                         ),
                     );
-                    defer dependency_version.value.workspace = path;
+                    defer {
+                        dependency_version.literal = path;
+                        dependency_version.value.workspace = path;
```
Jarred-Sumner: This is not the absolute path right

This is relative to the root package.json?

Could we have an assertion here that asserts its not an absolute path

Jarred-Sumner (approved):
LGTM after adding assertion

src/install/lockfile.zig
alexlamsl: > This is relative to the root package.json?

Yes, from the line directly above this:
https://github.com/oven-sh/bun/blob/76dee9312ba5201583ba9f8232b96397fa490808/src/install/lockfile.zig#L2640-L2653

> Could we have an assertion here that asserts its not an absolute path

Does checking `path[0] != '/'` cover this or is there a special function for that?

Jarred-Sumner: `std.debug.assert(!std.fs.path.isAbsolute(path)`

Issue #3158, opened 2023-06-01
Support test coverage reporters in `bun test`


jakeboone02:
### What is the problem this feature would solve?

Existing test libraries like Jest and Vitest support coverage reporters like istanbul and c8. `bun test` merely runs the tests, but it does not collect coverage information that could be provided to a reporter.

### What is the feature you are proposing to solve the problem?

Add the ability to specify a reporter (possibly include a default reporter with Bun) and collect coverage information with a `--coverage` flag or something similar.

### What alternatives have you considered?

The only alternative is to stick with Jest/Vitest/etc when coverage reporting is needed.

TiBianMod:
+1

Jarred-Sumner:
Bun v0.7.3 now includes a default code coverage reporter.

To use it:
```bash
bun test --coverage
```

It will output something like:

```ts
-------------|---------|---------|-------------------
File         | % Funcs | % Lines | Uncovered Line #s
-------------|---------|---------|-------------------
All files    |   38.89 |   42.11 |
 index-0.ts  |   33.33 |   36.84 | 10-15,19-24
 index-1.ts  |   33.33 |   36.84 | 10-15,19-24
 index-10.ts |   33.33 |   36.84 | 10-15,19-24
 index-2.ts  |   33.33 |   36.84 | 10-15,19-24
 index-3.ts  |   33.33 |   36.84 | 10-15,19-24
 index-4.ts  |   33.33 |   36.84 | 10-15,19-24
 index-5.ts  |   33.33 |   36.84 | 10-15,19-24
 index-6.ts  |   33.33 |   36.84 | 10-15,19-24
 index-7.ts  |   33.33 |   36.84 | 10-15,19-24
 index-8.ts  |   33.33 |   36.84 | 10-15,19-24
 index-9.ts  |   33.33 |   36.84 | 10-15,19-24
 index.ts    |  100.00 |  100.00 |
-------------|---------|---------|-------------------
```

ImLunaHey:
@Jarred-Sumner is there an issue we can follow for support for html, json, etc. reporters for that?

Jarred-Sumner:
@ImLunaHey there is an issue for lcov #4015 which could be used to generate an HTML reporter. But feel free to file one for HTML specifically

idfunctor:
@Jarred-Sumner 
[...]

Issue #2311, opened 2023-03-05
code coverage


nitsanavni:
### What is the problem this feature would solve?

code coverage is useful for many things

### What is the feature you are proposing to solve the problem?

could we have something similar to [c8](https://github.com/bcoe/c8) or [istanbul](https://istanbul.js.org/) for example?

Jarred-Sumner:
yeah, we need to figure out a way to do this. JavaScriptCore _probably_ has something for this, but I haven't seen it yet.

JUSTIVE:
think this PR is ok to close now

Electroid:
We added initial code coverage in #3975. To support other formats, we're tracking support here:
* #4015 
* #6715 
* #6716

maxmilton:
Issue #4346 seems to be about CLI configurable thresholds rather than supporting other output providers. Perhaps we need to open a new issue?

Electroid:
> Issue #4346 seems to be about CLI configurable thresholds rather than supporting other output providers. Perhaps we need to open a new issue?

I linked the wrong issue, I updated my comment.

PR #3975, opened 2023-08-04
Code coverage for `bun test`

Jarred-Sumner:
### What does this PR do?

This is an extremely WIP PR that lets you see what code ran in `bun test`. 

This adds a new flag `--coverage` to `bun test` which enables JavaScriptCore's `JSC::ControlFlowProfiler` which tracks the positions (byte offsets) of code that was executed and how many times.

The hard part here is not the code coverage, but is source-mapping from these byte offsets to line numbers in the input source file. 

I see a couple approaches to solving this:
1) Store byte ranges in sourcemaps, in addition to lines and columns (expensive!)
2) JSC::SourceProviderCache maybe??

Too manually. We need some integration tests for this.


Jarred-Sumner:
Fixes #2311 
Fixes #3158

Issue #3954, opened 2023-08-03
bun —hot clear screen option


simylein:
### What is the problem this feature would solve?

Hot reloading an http server with `bun —hot` multiple times leads to a lot of output in the terminal. 

### What is the feature you are proposing to solve the problem?

Similar to how `bun —watch` clears the screen on file change it could be done as well for `bun —hot`

### What alternatives have you considered?

Running `console.clear()` beforehand

Jarred-Sumner:
do you think it should be the default?

simylein:
That would be awesome and match the `--watch` behaviour

simylein:
This is now available on canary and will be in the next release.

comunidadio:
> That would be awesome and match the `--watch` behaviour

Currently clearing the screen is not the default behaviour for bun --watch actually.
https://github.com/oven-sh/bun/issues/3951

simylein:
Yes in the test context currently not but I think this will be adjusted in the future. 
For normal file watching and hot reloading it is. 

https://github.com/oven-sh/bun/assets/67859356/5afcb7c7-ef65-4532-b5f5-34e3ea75c688


https://github.com/oven-sh/bun/assets/67859356/8c72a79d-59c4-4ceb-807d-eb90741e6a2d

PR #3976, opened 2023-08-04
feat(hot-clear-screen): clear terminal on hot reload

simylein:
### feat(hot-clear-screen): clear terminal on hot reload

this clears the screen on file change when running with the `--hot` flag which is similar to how `--watch` behaves currently.
this will address #3954.

- [x] Code changes

I tested it locally

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -626,6 +626,12 @@ pub const VirtualMachine = struct {
             bun.reloadProcess(bun.default_allocator, !strings.eqlComptime(this.bundler.env.map.get("BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD") orelse "0", "true"));
         }
 
+        if (!strings.eqlComptime(this.bundler.env.map.get("BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD") orelse "0", "true")) {
+            Output.flush();
+            Output.disableBuffering();
+            Output.resetTerminalAll();
+        }
+
         this.global.reload();
         this.pending_internal_promise = this.reloadEntryPoint(this.main) catch @panic("Failed to reload");
     }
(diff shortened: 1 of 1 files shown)


src/bun.js/javascript.zig
```diff
@@ -1793,6 +1793,13 @@ pub const VirtualMachine = struct {
         );
         this.eventLoop().ensureWaker();
 
+        // Clear the terminal
+        if (!strings.eqlComptime(this.bundler.env.map.get("BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD") orelse "0", "true")) {
```
simylein: I kinda copied this from the `bun.reloadProcess()` function in https://github.com/oven-sh/bun/blob/main/src/bun.zig#L1308.

Jarred-Sumner: This will cause it to clear the terminal every time bun starts, whenever a Worker starts, and whenever a macro is called


Instead, put it here: https://github.com/oven-sh/bun/blob/8fccc1d27dd4b43833367b47973c65d1a44ce563/src/bun.js/javascript.zig#L629

simylein: Thank you 🙂

src/bun.js/javascript.zig
```diff
@@ -626,6 +626,12 @@ pub const VirtualMachine = struct {
             bun.reloadProcess(bun.default_allocator, !strings.eqlComptime(this.bundler.env.map.get("BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD") orelse "0", "true"));
         }
 
+        if (!strings.eqlComptime(this.bundler.env.map.get("BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD") orelse "0", "true")) {
+            Output.flush();
+            Output.disableBuffering();
+            Output.resetTerminalAll();
```
Jarred-Sumner, line 632: we should re-enable buffering after this line

otherwise LGTM

simylein: Will do, should this also be done at https://github.com/oven-sh/bun/blob/main/src/bun.zig#L1311?

Jarred-Sumner: unnnecessary bevcause the process will reload

Jarred-Sumner:
Thank you!

src/bun.js/javascript.zig
simylein: thank you for the explanation 🙂

mentalisttraceur: > Will do, should this also be done at https://github.com/oven-sh/bun/blob/main/src/bun.zig#L1311?

For anyone else reading from the future, the permalink for the above was https://github.com/oven-sh/bun/blob/76dee9312ba5201583ba9f8232b96397fa490808/src/bun.zig#L1311

Issue #3981, opened 2023-08-04
`dns.getServers` is not implemented.


Hanaasagi:
### What version of Bun is running?

0.7.2

### What platform is your computer?

Linux 6.4.7-arch1-2 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const dns = require("node:dns");
console.log(dns.getServers());

```

### What is the expected behavior?

Print local dns config.

### What do you see instead?

```
TypeError: dns.getServers is not a function. (In 'dns.getServers()', 'dns.getServers' is undefined)

```

PR #3982, opened 2023-08-04
feat: impl `dns.getServers`

Hanaasagi:
Close: #3981

### What does this PR do?

- Implement the `dns.getServers` function

Reference:

- https://nodejs.org/api/dns.html#dnsgetservers
- https://c-ares.org/ares_get_servers_ports.html

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
nice

test/js/node/dns/node-dns.test.js
```diff
@@ -190,3 +192,34 @@ test("dns.lookup (localhost)", done => {
     done(err);
   });
 });
+
+test("dns.getServers", done => {
+  function parseResolvConf() {
+    let servers = [];
+    try {
+      const content = fs.readFileSync("/etc/resolv.conf", "utf-8");
```
Jarred-Sumner, line 200: I think this test will only work on linux 

can use test.skipIf(process.platform === “darwin”)

Jarred-Sumner (approved):
Thank you. This looks good. Let’s let the tests run. 

Only part I’m not sure about is the buffer length. If any of the numbers are not 100% correct in all the cases it could cause a crash or undefined behavior and it’d be hard to reproduce. 

Assuming the current length isn’t huge (like 64 bytes probably), I suggest adding an extra 512 bytes to it just for wiggle room

Hanaasagi:
About the buffer length, I checked the source code of `ares_inet_ntop`. It checks the `size` parameter we pass in, so overflow should not occur. However, I think I should handle the case where the return value is `NULL`.

https://github.com/c-ares/c-ares/blob/bad62225b7f6b278b92e8e85a255600b629ef517/src/lib/inet_ntop.c#L58-L95

And `std.fmt.bufPrint` will also check buf length.


https://github.com/ziglang/zig/blob/fc6e5756848cd627b69756853937ce22500e62cc/lib/std/fmt.zig#L1996
https://github.com/ziglang/zig/blob/fc6e5756848cd627b69756853937ce22500e62cc/lib/std/io/fixed_buffer_stream.zig#L63

Jarred-Sumner:
> About the buffer length, I checked the source code of `ares_inet_ntop`. It checks the `size` parameter we pass in, so overflow should not occur. However, I think I should handle the case where the return value is `NULL`.
> 
> https://github.com/c-ares/c-ares/blob/bad62225b7f6b278b92e8e85a255600b629ef517/src/lib/inet_ntop.c#L58-L95
> 
> And `std.fmt.bufPrint` will also check buf length.
> 
> https://github.com/ziglang/zig/blob/fc6e5756848cd627b69756853937ce22500e62cc/lib/std/fmt.zig#L1996 https://github.com/ziglang/zig/blob/fc6e5756848cd627b69756853937ce22500e62cc/lib/std/io/fixed_buffer_stream.zig#L63

👍

PR #3998, opened 2023-08-05
Implement --test-name-pattern

### LINKED ISSUE #3990: Support jest's `--testNamePattern`/`-t`

Running a single test or a few tests from a large test file is not easy. Having this flag makes it a little easier to get extensions/programatic execution of tests.

Implement this flag exactly as is:
https://jestjs.io/docs/cli#--testnamepatternregex

Though maybe for programatic cases, we have some line-number based filtering method so that it can more accurately tell the runner which tests to skip.

Commenting out/removing the tests.

---

I swear we already had an issue for this open but i couldnt find it.

paperclover:
### What does this PR do?

Closes #3990

image

- [X] Code changes


Jarred-Sumner:
@paperdave looks like the changes to JSON strings broke some tests

Issue #3444, opened 2023-06-28
Bun init fails when entry point does not exists


gtrabanco:
### What version of Bun is running?

0.6.11

### What platform is your computer?

Darwin 22.5.0 x86_64 i386

### What steps can reproduce the bug?

```
$ bun init
```

Just type a different entry point that does not exists different than default one. I typed `./src/index.ts`

### What is the expected behavior?

I expect to create the same project as default one but creating the file in the entry point if does not exists or simply ignore that entry point does not exists.

### What do you see instead?

```
$ bun init
bun init helps you get started with a minimal project and tries to guess sensible defaults. Press ^C anytime to quit

package name (buncg):
entry point (index.ts): ./src/index.ts

Done! A package.json file was saved in the current directory.

error: FileNotFound
bun could not find a file, and the code that produces this error is missing a better error.

----- bun meta -----
Bun v0.6.11 (0de5bb22) macOS x64 22.5.0
Unknown:
Elapsed: 7868ms | User: 5ms | Sys: 5ms
RSS: 4.57MB | Peak: 4.57MB | Commit: 33.55MB | Faults: 0
----- bun meta -----
```

Hanaasagi:
It seems that the reason for this error is the lack of logic to automatically create the non-existent directory (`./src`). A temporary workaround is `mkdir src` before `bun init`

https://github.com/oven-sh/bun/blob/0de5bb22af427dcad57731d7063f8678b13265e2/src/cli/init_command.zig#L339-L345

The value of `fields.entry_point` is `./src/index.ts`.

PR #4002, opened 2023-08-05
Fix(cli/init): support subpath entrypoint.

Hanaasagi:
Close: #3444

### What does this PR do?

- Support subpath entrypoint, e.g. `./src/index.ts`

- [x] Code changes

test locally with

- `./src/index.ts`
- `src/index.ts`

### DIFF
--- a/src/cli/init_command.zig
+++ b/src/cli/init_command.zig
@@ -337,7 +337,14 @@ pub const InitCommand = struct {
         }
 
         if (fields.entry_point.len > 0 and !exists(fields.entry_point)) {
-            var entry = try std.fs.cwd().createFile(fields.entry_point, .{ .truncate = true });
+            const cwd = std.fs.cwd();
+            if (std.fs.path.dirname(fields.entry_point)) |dirname| {
+                if (!(dirname.len == 0 and dirname[0] == '.')) {
+                    cwd.makePath(dirname) catch {};
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


src/cli/init_command.zig
```diff
@@ -337,7 +337,14 @@ pub const InitCommand = struct {
         }
 
         if (fields.entry_point.len > 0 and !exists(fields.entry_point)) {
-            var entry = try std.fs.cwd().createFile(fields.entry_point, .{ .truncate = true });
+            const cwd = std.fs.cwd();
+            if (std.fs.path.dirname(fields.entry_point)) |dirname| {
+                if (!(dirname.len == 0 and dirname[0] == '.')) {
```
Jarred-Sumner: I think you meant 

```suggestion
                if (!(strings.eqlComptime(dirname, “.”) or strings.eqlComptime(dirname, “./))) {
```

dirname will return null if it has no separator so checking for .len == 0 will never be reached

Jarred-Sumner (approved):

src/cli/init_command.zig
Hanaasagi: ah, a typo.

Jarred-Sumner:
Thank you

src/cli/init_command.zig
Hanaasagi: Fixed. `dirname` will not return `"./"`, comparing to `"."` is enough here.

PR #4028, opened 2023-08-06
[install] handle `bun add` of existing `peerDependencies` correctly

alexlamsl:
### What does this PR do?

Previously `bun add pkg` on the following `package.json`:
```json
{
  "peerDependencies": {
    "pkg": "^1.2.3"
  }
}
```
will result in failure, i.e.:
```json
{
  "peerDependencies": {
    "pkg": ""
  }
}
```
without installing `pkg`.

- [x] Code changes


Jarred-Sumner:
Thank you

Issue #3937, opened 2023-08-03
`TypeError: undefined is not an object` when using `require.resolve`


Electroid:
This is a blocker for Next.js.

```js
const resolve = process.env.NEXT_MINIMAL ? __non_webpack_require__.resolve : require.resolve;
```

```js
bun --bun run dev
$ next dev
79 | addHookAliases([
80 |     // Use `require.resolve` explicitly to make them statically analyzable
81 |     // styled-jsx needs to be resolved as the external dependency.
82 |     [
83 |         "styled-jsx",
84 |         dirname(resolve("styled-jsx/package.json"))
                   ^
TypeError: undefined is not an object
      at /node_modules/next/dist/server/require-hook.js:84:16
      at globalThis (/node_modules/next/dist/server/require-hook.js:115:46)
      at /node_modules/next/dist/build/worker.js:12:0
      at globalThis (/node_modules/next/dist/build/worker.js:14:82)
error: script "dev" exited with code 1 (SIGHUP)
```

PR #4030, opened 2023-08-07
Bind require.resolve()

Jarred-Sumner:
### What does this PR do?

This fixes code like the following:
```js
var r = require.resolve;
r("foo");
```

Previously, it would throw a strange error about `undefined is not an object`.

Our CommonJS function wrapper now looks like this:

```js
(function (module, exports, require, __dirname, __filename) {
}).call(
  $_BunCommonJSModule_$.module.exports,
  $_BunCommonJSModule_$.module,
  $_BunCommonJSModule_$.module.exports,
  (($_BunCommonJSModule_$.module.require =
    $_BunCommonJSModule_$.module.require.bind($_BunCommonJSModule_$.module)),
  ($_BunCommonJSModule_$.module.require.path = $_BunCommonJSModule_$.module.id),
  ($_BunCommonJSModule_$.module.require.resolve =
    $_BunCommonJSModule_$.module.require.resolve.bind(
      $_BunCommonJSModule_$.module.id
    )),
  $_BunCommonJSModule_$.module.require),
  $_BunCommonJSModule_$.__dirname,
  $_BunCommonJSModule_$.__filename
);
```

Updated our require-cache fixture to also use require.resolve from a `var`

### DIFF
--- a/src/js/builtins/Module.ts
+++ b/src/js/builtins/Module.ts
@@ -81,8 +81,8 @@ export function require(this: CommonJSModuleRecord, id: string) {
   return mod.exports;
 }
 
-export function requireResolve(this: CommonJSModuleRecord, id: string) {
-  return $resolveSync(id, this.path, false);
+export function requireResolve(this: string | { path: string }, id: string) {
+  return $resolveSync(id, typeof this === "string" ? this : this?.path, false);
 }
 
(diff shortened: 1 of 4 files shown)


paperclover:
This will close https://github.com/oven-sh/bun/issues/3937

Issue #4041, opened 2023-08-07
`path.normalize` should return `.` when argument is empty string.


Hanaasagi:
### What version of Bun is running?

0.7.2

### What platform is your computer?

Linux 6.4.7-arch1-2 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const assert = require("node:assert");
const path = require("node:path");

assert.strictEqual(path.normalize(""), ".");

```

### What is the expected behavior?

No error.

### What do you see instead?

```
AssertionError: Expected values to be strictly equal:

'' !== '.'

 code: "ERR_ASSERTION"

```

### Additional information

From Nodejs doc: https://nodejs.org/api/path.html#pathnormalizepath

> If the path is a zero-length string, '.' is returned, representing the current working directory.

PR #4042, opened 2023-08-07
Fix `path.normalize` edge case.

Hanaasagi:
Close: #4041

### What does this PR do?

`path.normalize("")`: If the path is a zero-length string, '.' is returned, representing the current working directory.

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
Thank you

PR #4043, opened 2023-08-07
WASM test analyzer

Jarred-Sumner:
### What does this PR do?

This does three things
1) Get's Bun's transpiler to successfully run in WASM
2) Creates a small internal abstraction for using bun's transpiler to analyze parsed ASTs
3) Uses the above abstraction to export a `getTests` function to WASM which returns an array of `test` and `describe` calls in a file along with the name of the tests. It will include tests with `test.only`, but ignore any other kinds of tests. IT skips if the test does not have a string argument.

@jhmaster2000 this is probably an okay starting point for other cases where you might want to directly use Bun in `node-bun`

Some notes:
- WASM build is 1.3 MB with release fast and with release small, it's 800 KB.
- .tsx only (which also includes .js and .ts)

API example in `test/node.mjs` in `packages/bun-wasm`:

```ts
import { init, getTests } from "../";

const buf = (process.argv.length > 2 ? readFileSync(process.argv.at(-1)) : "") || readFileSync(import.meta.url);
await init(new URL("../bun.wasm", import.meta.url));

console.log(getTests(buf));
```

If we pass that to [./test/js/node/buffer.test.js](https://github.com/oven-sh/bun/blob/jarred/wasm-test-analyzer/test/js/node/buffer.test.js#L2555)

```js
[
  { name: 'isAscii', byteOffset: 258, kind: 'test' },
  { name: 'isUtf8', byteOffset: 625, kind: 'test' },
  { name: 'Buffer global is settable', byteOffset: 1166, kind: 'test' },
  { name: 'length overflow', byteOffset: 1448, kind: 'test' },
  { name: 'truncate input values', byteOffset: 1681, kind: 'test' },
  { name: 'Buffer.allocUnsafe()', byteOffset: 1957, kind: 'test' },
  { name: 'Buffer.from()', byteOffset: 2064, kind: 'test' },
  { name: 'offset properties', byteOffset: 2154, kind: 'test' },
  {
    name: 'creating a Buffer from a Uint32Array',
    byteOffset: 2312,
    kind: 'test'
  },
  {
    name: 'creating a Buffer from a Uint32Array (old constructor)',
    byteOffset: 2530,
    kind: 'test'
  },
  { name: 'invalid encoding', byteOffset: 2757, kind: 'test' },
  { name: 'create 0-length buffers', byteOffset: 3159, kind: 'test' },
  {
    name: 'write() beyond end of buffer',
    byteOffset: 3436,
    kind: 'test'
  },
  ...
]
```

Note that to add exported functions, you have to:

- Append to  `-sEXPORTED_FUNCTIONS` in the Makefile
- `export fn ...` in Zig
- Reference the function in a `comptime` scope
- Reference the function in `emcc_main` 

Communication between WASM and JS is handled via peechy, using the schema. The heap size is a hardcoded limit you can pass on initialization

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -703,32 +703,32 @@ dev-build-obj-wasm:
 
 .PHONY: dev-wasm
 dev-wasm: dev-build-obj-wasm
-	emcc -sEXPORTED_FUNCTIONS="['_bun_free', '_cycleStart', '_cycleEnd', '_bun_malloc', '_scan', '_transform', '_init']" \
-		-g -s ERROR_ON_UNDEFINED_SYMBOLS=0  -DNDEBUG  \
+	emcc -sEXPORTED_FUNCTIONS="['_bun_free', '_cycleStart', '_cycleEnd', '_bun_malloc', '_scan', '_transform', '_init', '_getTests']" \
+		-g2 -s ERROR_ON_UNDEFINED_SYMBOLS=0  -DNDEBUG  \
 		$(BUN_DEPS_DIR)/libmimalloc.a.wasm  \
-		packages/debug-bun-freestanding-wasm32/bun-wasm.o $(OPTIMIZATION_LEVEL) --no-entry --allow-undefined  -s ASSERTIONS=0  -s ALLOW_MEMORY_GROWTH=1 -s WASM_BIGINT=1  \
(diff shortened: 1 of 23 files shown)


jhmaster2000:
Nice thanks, will look into integrating the polyfills with this later this week, there's a few smaller things I want to get out of the way first.

PR #4046, opened 2023-08-07
implement `bun update`

alexlamsl:
- analogous to `npm update`
- `bun update <name>` to refresh specified package under `package.json`
- `bun update` to refresh all package to latest versions

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1092,11 +1092,11 @@ pub const Command = struct {
                 for (args_iter.buf) |arg| {
                     const span = std.mem.span(arg);
                     if (span.len > 0 and (strings.eqlComptime(span, "-g") or strings.eqlComptime(span, "--global"))) {
-                        break :brk Command.Tag.AddCommand;
+                        break :brk .AddCommand;
                     }
                 }
 
-                break :brk Command.Tag.InstallCommand;
+                break :brk .InstallCommand;
             },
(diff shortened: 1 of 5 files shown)


Jarred-Sumner:
Can you add a test with dependencies of dependencies that get updated? 

From reading the code, I'm a little confused why this works when the dependency is not passed in the list. Where does it choose to update the package? 

I'm also wondering how it deals with version tags like "latest", if those get updated. And if "*" is chosen.

alexlamsl:
> From reading the code, I'm a little confused why this works when the dependency is not passed in the list. Where does it choose to update the package?

I'm assuming you meant `bun update` without any package names &minus; in which case `PackageManager.package_json_updates` would indeed an empty list, but since we've set `.to_update = true` we can differentiate this and mark all existing packages to be updated.

> I'm also wondering how it deals with version tags like "latest", if those get updated. And if "*" is chosen.

Same way as it would be handled by `bun install` without an existing lockfile, which is covered here IIRC:
https://github.com/oven-sh/bun/blob/f2f227720b3ffe1797a0a4e500e9a9a639167dc6/test/cli/install/bun-install.test.ts#L1228-L1288

> Can you add a test with dependencies of dependencies that get updated?

Will do 👌

Jarred-Sumner:
We need to update zsh/fish/bash completions to add the `update` command.

Jarred-Sumner:
Thank you

PR #4047, opened 2023-08-07
feat(bun:test) add support for test.each() and describe.each()

jecquas:
### What does this PR do?

add support for [test.each()](https://jestjs.io/docs/api#testeachtablename-fn-timeout) and [describe.each()](https://jestjs.io/docs/api#describeeachtablename-fn-timeout)

- [x] Code changes

I wrote automated tests and ran bun test

image


src/bun.js/test/jest.zig
```diff
@@ -540,7 +550,12 @@ pub const Jest = struct {
 pub const TestScope = struct {
     label: string = "",
     parent: *DescribeScope,
-    callback: JSC.JSValue,
+
+    func: JSC.JSValue,
+    func_arg: []JSC.JSValue,
```
Jarred-Sumner, line 561: in zig, slices store a len so we don't need to store both arg_size and len

src/bun.js/test/jest.zig
```diff
@@ -631,17 +650,22 @@ pub const TestScope = struct {
...
         if (comptime is_bindgen) return undefined;
         var vm = VirtualMachine.get();
-        const callback = this.callback;
+        const func = this.func;
         Jest.runner.?.did_pending_test_fail = false;
         defer {
-            callback.unprotect();
-            this.callback = .zero;
+            var idx: u32 = 0;
+            while (idx < this.func_arg_size) {
```
Jarred-Sumner: I think you can do
```suggestion
            for (this.func_args) |arg| {
               arg.unprotect();
             }
```

src/bun.js/test/jest.zig
```diff
@@ -1705,3 +1748,180 @@ pub fn printGithubAnnotation(exception: *JSC.ZigException) void {
...
+            if (has_callback_function) {
+                arg_size += 1;
+            }
+
+            var function_args = allocator.alloc(JSC.JSValue, arg_size) catch @panic("can't create function_args");
+            var idx: u32 = 0;
+
+            if (item_is_array) {
+                // Spread array as args
+                const item_length = item.getLength(globalThis);
```
Jarred-Sumner: This should use `JSC.JSArrayIterator` to iterate through the array.

src/bun.js/test/jest.zig
```diff
@@ -1705,3 +1748,180 @@ pub fn printGithubAnnotation(exception: *JSC.ZigException) void {
...
+                while (idx < item_length) : (idx += 1) {
+                    var arg = item.getIndex(globalThis, idx);
```
Jarred-Sumner: Every call to `getIndex` can potentially throw. If `arg` == `.zero` then it threw. If we do not handle exceptions, Bun will segfault.

Also, use `const` instead of `var` whenever possible

src/bun.js/test/jest.zig
```diff
@@ -1705,3 +1748,180 @@ pub fn printGithubAnnotation(exception: *JSC.ZigException) void {
...
+
+        if (array.isUndefinedOrNull() or !array.jsType().isArray()) {
+            return .zero;
+        }
+
+        var iter = JSC.JSArrayIterator.init(array, globalThis);
+
+        while (iter.next()) |item| {
+            // TODO: node:util.format() the label
+            const label = if (description == .zero)
```
Jarred-Sumner: description will not be zero in this case, it would be `.undefined`

src/bun.js/test/jest.zig
```diff
@@ -1705,3 +1748,180 @@ pub fn printGithubAnnotation(exception: *JSC.ZigException) void {
...
+    const arguments = callframe.arguments(3);
+    const args = arguments.ptr[0..arguments.len];
+
+    if (args.len < 2) {
+        globalThis.throwPretty("{s} a description and callback function", .{signature});
+        return .zero;
+    }
+
+    var description = args[0];
+    var function = args[1];
```
Jarred-Sumner, line 1793: where does it check if `function` is a function? You can use `.isCallable` for that.

Jarred-Sumner:
Thank you for this, it's a good start

jecquas:
changes made
still get a sigsegv when running all the tests, but happens on main as well (#4062)

src/bun.js/test/jest.zig
```diff
@@ -1737,3 +1772,187 @@ pub fn printGithubAnnotation(exception: *JSC.ZigException) void {
     Output.printError("\n", .{});
     Output.flush();
 }
+
+pub const EachData = struct { strong: JSC.Strong, is_test: bool };
+
+fn eachBind(
```
Jarred-Sumner, line 1778: This should really be two different functions or a single function with a comptime argument and then `is_test` becomes unnecessary

but that shouldn't block this from shipping

src/bun.js/test/jest.zig
```diff
@@ -1737,3 +1772,187 @@ pub fn printGithubAnnotation(exception: *JSC.ZigException) void {
...
+
+            var function_args = allocator.alloc(JSC.JSValue, arg_size) catch @panic("can't create function_args");
+            var idx: u32 = 0;
+
+            if (item_is_array) {
+                // Spread array as args
+                var item_iter = item.arrayIterator(globalThis);
+                while (item_iter.next()) |array_item| {
+                    if (array_item == .zero) {
+                        allocator.free(function_args);
```
Jarred-Sumner, line 1878: this leaks memory because existing function args will not be unprotected 

but lets handle that a little later, this is very much an edgecase as-is

Jarred-Sumner:
Thank you for this. Looks good but we should add the formatters too. I think a simplified version would be fine. Probably using the console.log formatter to handle %s, and other common ones.

PR #4048, opened 2023-08-07
fix iterating headers with `set-cookie`

dylan-conway:
### What does this PR do?

before:
```js
const headers = new Headers();
headers.append("Set-Cookie", "cookies");
headers.append("Content-Type", "application/json");

console.log([...headers]);
// output: [ [ "content-type", "application/json" ], [ "content-type", "application/json" ] ]
```
after:
```js
// output: [ [ "content-type", "application/json" ], [ "set-cookie", "cookies" ] ]
```

updated and added test for headers with `set-cookie`

### DIFF
--- a/src/bun.js/bindings/webcore/FetchHeaders.cpp
+++ b/src/bun.js/bindings/webcore/FetchHeaders.cpp
@@ -227,6 +227,8 @@ ExceptionOr<bool> FetchHeaders::has(const String& name) const
     return m_headers.contains(name);
 }
 
+static NeverDestroyed<const String> setCookieLowercaseString(MAKE_STATIC_STRING_IMPL("set-cookie"));
+
 ExceptionOr<void> FetchHeaders::set(const String& name, const String& value)
 {
     String normalizedValue = value.trim(isHTTPSpace);
@@ -237,7 +239,12 @@ ExceptionOr<void> FetchHeaders::set(const String& name, const String& value)
         return {};
 
     ++m_updateCounter;
-    m_headers.set(name, normalizedValue);
+    if (equalIgnoringASCIICase(name, setCookieLowercaseString)) {
+        m_headers.getSetCookieHeaders().clear();
+        m_headers.getSetCookieHeaders().append(normalizedValue);
+    } else {
+        m_headers.set(name, normalizedValue);
+    }
 
     if (m_guard == FetchHeaders::Guard::RequestNoCors)
         removePrivilegedNoCORSRequestHeaders(m_headers);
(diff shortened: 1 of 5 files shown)


paperclover:
~~This will close #600 :)~~ I mispoke, it will only close it for `bun --bun run dev`

There is a bug in building that still has to get investigated.

src/bun.js/bindings/webcore/FetchHeaders.cpp
```diff
@@ -261,30 +268,56 @@ void FetchHeaders::filterAndFill(const HTTPHeaderMap& headers, Guard guard)
...
 {
     if (m_keys.isEmpty() || m_updateCounter != m_headers->m_updateCounter) {
+        bool hasSetCookie = !m_headers->getSetCookieHeaders().isEmpty();
         m_keys.resize(0);
-        m_keys.reserveCapacity(m_headers->m_headers.size());
+        m_keys.reserveCapacity(m_headers->m_headers.size() + (hasSetCookie ? 1 : 0));
         for (auto& header : m_headers->m_headers)
             m_keys.uncheckedAppend(header.asciiLowerCaseName());
-        std::sort(m_keys.begin(), m_keys.end(), WTF::codePointCompareLessThan);
+        if (hasSetCookie)
```
Jarred-Sumner, line 274: Do we need to pay this performance overhead here? Wouldn't it be fine if we put the Set-Cookie header either first or last?

src/bun.js/bindings/webcore/FetchHeaders.cpp
```diff
@@ -237,7 +239,12 @@ ExceptionOr<void> FetchHeaders::set(const String& name, const String& value)
         return {};
 
     ++m_updateCounter;
-    m_headers.set(name, normalizedValue);
+    if (equalIgnoringASCIICase(name, setCookieLowercaseString)) {
```
Jarred-Sumner: Please see the code in `HTTPHeaderMap::set`

This is unnecessary duplicate work that will make the Headers object slower

https://github.com/oven-sh/bun/blob/0b183beb51367004795d8a431eb06bb2fa4f8250/src/bun.js/bindings/webcore/HTTPHeaderMap.cpp#L112-L115

We can use the header name from here ^

Jarred-Sumner (changes requested):
These changes should be moved to HTTPHeaderMap to avoid duplicate work

src/bun.js/bindings/webcore/FetchHeaders.cpp
```diff
@@ -227,6 +227,8 @@ ExceptionOr<bool> FetchHeaders::has(const String& name) const
     return m_headers.contains(name);
 }
 
+static NeverDestroyed<const String> setCookieLowercaseString(MAKE_STATIC_STRING_IMPL("set-cookie"));
```
Jarred-Sumner: this should already exist in HTTPHeaderNames

src/bun.js/bindings/webcore/FetchHeaders.cpp
dylan-conway: we do, this is part of the fetch spec https://fetch.spec.whatwg.org/#convert-header-names-to-a-sorted-lowercase-set

Jarred-Sumner: It was also expected in the spec for keys to be unique which isn’t true for set-cookies 

I think we should just always put it at the end and call it a day. Each call to next() shouldn’t slow it down

src/bun.js/bindings/webcore/HTTPHeaderMap.cpp
```diff
@@ -236,15 +236,7 @@ String HTTPHeaderMap::get(HTTPHeaderName name) const
...
-        auto cookieName = extractCookieName(value);
-        size_t length = m_setCookieHeaders.size();
-        const auto& cookies = m_setCookieHeaders.data();
-        for (size_t i = 0; i < length; ++i) {
-            if (extractCookieName(cookies[i]) == cookieName) {
-                m_setCookieHeaders[i] = value;
-                return;
-            }
-        }
+        m_setCookieHeaders.clear();
```
Jarred-Sumner, line 239: Will this work if you call headers.append? It doesn’t

dylan-conway: Yeah this will work. headers.append will go through HTTPHeaderMap::add. Set will clear the existing set-cookie values.

src/bun.js/bindings/webcore/FetchHeaders.cpp
dylan-conway: updated to append after the sort

src/bun.js/bindings/webcore/FetchHeaders.cpp
```diff
@@ -261,30 +262,58 @@ void FetchHeaders::filterAndFill(const HTTPHeaderMap& headers, Guard guard)
     }
 }
 
-static NeverDestroyed<const String> setCookieLowercaseString(MAKE_STATIC_STRING_IMPL("set-cookie"));
+static const WTF::String setCookieLowercaseString = WTF::staticHeaderNames[static_cast<uint8_t>(HTTPHeaderName::SetCookie)];
+
+static bool compareIteratorKeys(const String& a, const String& b)
```
Jarred-Sumner: this can be removed since we no longer useit here

src/bun.js/bindings/webcore/FetchHeaders.cpp
```diff
@@ -261,30 +262,58 @@ void FetchHeaders::filterAndFill(const HTTPHeaderMap& headers, Guard guard)
     }
 }
 
-static NeverDestroyed<const String> setCookieLowercaseString(MAKE_STATIC_STRING_IMPL("set-cookie"));
+static const WTF::String setCookieLowercaseString = WTF::staticHeaderNames[static_cast<uint8_t>(HTTPHeaderName::SetCookie)];
```
Jarred-Sumner: can we move this to be inside the usage so it doesn't impact startup time?

src/bun.js/bindings/webcore/FetchHeaders.cpp
dylan-conway: updated

src/bun.js/bindings/webcore/FetchHeaders.cpp
dylan-conway: updated

Jarred-Sumner:
Thank you

PR #4052, opened 2023-08-08
Add `env` option for `node:worker_threads`

paperclover:
### What does this PR do?

- Types: Fix Worker and Node Worker being the same type; they are different.
- Implement the `env` option for workers.
- In some stubbed parts of worker_threads, throw better errors.
- .terminate calls unref properly
- ref true is default

Makes sveltekit work

- [x] Code changes

new automated tests


paperclover:
ready for rewview

PR #4054, opened 2023-08-08
Fix `Bun.hash` functions

jhmaster2000:
### What does this PR do?

The `Bun.hash` hashing functions have been both receiving and returning inaccurate data, this PR fixes these:
* The 64-bit hashes cannot be represented by plain JS numbers and were being truncated, yielding incorrect hashes, they now return bigints.
* The 64-bit hashes use 64-bit seeds, but the seeds were being truncated to 32 bits, resulting in the upper seed range being impossible to achieve.
* All hashes were ignoring bigint seed inputs, they now accept bigints alongside numbers, even 32-bit hashes, which will just truncate the input to 32 bits if needed.
* All hashes were returning signed values while hashes are not meant to be negative, they now return proper unsigned values.

Additionally, the hash `murmur32v2` was undocumented, so it is now. (I have already updated `bun-types` for these changes on the bun-polyfills branch so I'm not including them here to avoid a future merge conflict)

- [x] Code changes

The existing `Bun.hash` tests have been updated to reflect the correct expected behavior with the fixes.

### DIFF
--- a/docs/api/hashing.md
+++ b/docs/api/hashing.md
@@ -77,7 +77,7 @@ The standard `Bun.hash` functions uses [Wyhash](https://github.com/wangyi-fudan/
 
 ```ts
 Bun.hash("some data here");
-// 976213160445840
+// 11562320457524636935n
 ```
 
 The input can be a string, `TypedArray`, `DataView`, `ArrayBuffer`, or `SharedArrayBuffer`.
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
looks good, don't forget to update the types so that they report bigint instead of number where appropriate

Jarred-Sumner (approved):

jhmaster2000:
> looks good, don't forget to update the types so that they report bigint instead of number where appropriate

As I said I'm already updating the types on `bun-polyfills` branch so doing it here would create a merge conflict, I will make the initial PR for that branch tomorrow so this is sorted.

Jarred-Sumner:
Thank you

PR #4083, opened 2023-08-09
feat(bun/run): Impl. npm_package_config (fwup)

### LINKED ISSUE #73: Add support for npm_package_config_* variables

v3nntrnx:
Reopening: #3661 (Reason: troubles with git)

---

### What does this PR do?

This implements the "config" option in package.json which declares env variables.
Closes #73

- [x] Code changes

I wrote automated tests

### Note:
arrays are not implemented here


src/resolver/package_json.zig
```diff
@@ -1021,6 +1043,60 @@ pub const PackageJSON = struct {
...
+                            package_json.npm_cfg_map.put(lkey, "") catch unreachable;
+                        },
+                        // https://github.com/oven-sh/bun/pull/3661#issuecomment-1657227789
+                        .e_array => {
+                            r.log.addWarning(json_source, prop.value.?.loc, "Arrays in package.json config is not implemented in Bun. If you depend on this, please file an issue.") catch unreachable;
+                            continue;
+                        },
+                        // TODO: Add support for arrays, https://github.com/oven-sh/bun/pull/3661#issuecomment-1657227789
+                        else => {
+                            r.log.addWarning(json_source, prop.value.?.loc, "Values of 'config' must be either a boolean, number, string, or object.") catch unreachable;
```
paperclover: let's be consistant with ourself, double quotes

paperclover (approved):
looks good, assuming this follows exactly what `npm` / `yarn` / `pnpm` does with config. minus the array stuff.

v3nntrnx:
> looks good, assuming this follows exactly what `npm` / `yarn` / `pnpm` does with config. minus the array stuff.

I spent so much time debugging this stuff, so I am pretty sure in myself

v3nntrnx:
Just made a test, if the config is "port": 8080 and there is a duplicate (e.g. "port":  7070) it will take the second value (in node and current implementation here), might optimize that to use a single call.

Jarred-Sumner (changes requested):
1) package.json `"config"` objects need to only be parsed for `bun install` lifecycle scripts or `bun run`. Not when resolving package.json for the module resolver in the runtime or bundler or transpiler. There is a flag that looks at the `"bin"` field and that flag could be re-used here for these `"config"` objects. 

2) Why is the `npm_cfg_map` actually necessary if it's only used for putting into the env map?

3) If the npm_cfg_map is necessary, it should be using a stackFallbackAllocator + an bun.ArenaAllocator allocator for all keys. As-is, all the calls to `strings.concat` leak memory. And the keys and the values also leak memory. The implementation of `bun.StringMap` already duplicates the values and has an option to duplicate the keys:

https://github.com/oven-sh/bun/blob/main/src/bun.zig#L1395-L1461

4) Can `runPackageScript` be called on multiple package.json? I think the answer is yes due to workspaces in `bun install`. And if so, the changes to `this_bundler.env` must be isolated per package.json, which means directly writing to the env map won't work (you would have config keys from an entirely different package.json)

PR #4092, opened 2023-08-09
bun test: format description of test.each

jecquas:
### What does this PR do?

Add description formatting to test.each and describe.each callbacks

https://jestjs.io/docs/api#testeachtablename-fn-timeout

Generate unique test titles by positionally injecting parameters with [printf formatting](https://nodejs.org/api/util.html#util_util_format_format_args):

- [ ] %p - [pretty-format](https://www.npmjs.com/package/pretty-format).
- [x] %s- String.
- [x] %d- Number.
- [x] %i - Integer.
- [x] %f - Floating point value.
- [x] %j - JSON.
- [x] %o - Object.
- [x] %# - Index of the test case.
- [x] %% - single percent sign ('%'). This does not consume an argument.

- [x] Code changes

I updated automated tests

image


test/js/bun/test/jest-each.test.ts
```diff
@@ -11,10 +11,10 @@ describe("jest-each", () => {
     expect(it.each).toBeTypeOf("function");
     expect(it.each([])).toBeTypeOf("function");
   });
-  it.each(NUMBERS)("add two numbers", (a, b, e) => {
+  it.each(NUMBERS)("%i + %i = %i", (a, b, e) => {
```
Electroid, line 14: There should be tests in `test/cli/test` that verifies that these labels are formatted properly, from stdout. Otherwise, it can only be verified from a human.

Jarred-Sumner:
Thank you!

PR #4107, opened 2023-08-10
fs.zig: create temp files with 0o700, not 0o007

adtac:
### What does this PR do?

Fix the permission on temporary files created by bun. I'm not sure why this was ever IRWXO.

- [x] Code changes

This is a one character change.


Jarred-Sumner:
Thank you

PR #4114, opened 2023-08-11
Fix worker event loop ref/unref + leak

paperclover:
### What does this PR do?

Fixes worker tests but it makes these changes to WebWorker.

**was reimplemented a little on aug 11**:

- A worker's event loop impacts the main process' event loop. Meaning if you setTimeout in a worker/listen for messages, it keeps main alive.
- Worker is by default `.ref()`, but workers will exit if their event loop dies.
- If you do `addEventListener("message")` on a message port or the global, this refs the event loop.
- If you do `worker.unref()` this no longer makes the worker's event loop affect the caller's. This is what esbuild depends on.
- We were not freeing WebWorker struct. This was because its a little complex. ~~I solved it by assigning vm to null when the vm is done with, then the while(true) loop for the worker (the source of the UAF) would call real deinit.~~ It's not really that complex with the changes to the worker event loop. `.terminate` sets a flag to tell the worker to exit the event loop, and the deinit is called from within `spin()` on the Worker's thread.

I'd like a close review to make sure i did everything right. I triple checked my uses of event loop refs and memory usage in the worker zig file.

- [X] Code changes

I ran unit tests. For testing large scale worker event loop stuff.

```ts
import { Worker, isMainThread, workerData } from "worker_threads";

if (isMainThread) {
    const w = new Worker("/Users/dave/code/bun/test/js/web/worker-demo.mjs", { ref: false, workerData: i });
    let recieved = false;
    w.on("message", msg => {
      if (msg === "initial message") {
        recieved = true;
      } else {
        console.log("WHAT?", msg);
      }
    });
    setTimeout(() => {
      if (!recieved) {
        w.terminate();
        console.log("worker", i, "did not respond");
      }
    }, 5000 + 500);
  }
} else {
  setTimeout(() => {
    self.postMessage("initial message");
    console.log("workerid=", workerData);
  }, Math.random() * 5000 + 500);
}
```

Pipe this to a file to see that exactly 1000 lines are output.


src/bun.js/module_loader.zig
```diff
@@ -2021,11 +2021,11 @@ pub const ModuleLoader = struct {
         } else if (HardcodedModule.Map.getWithEql(specifier, bun.String.eqlComptime)) |hardcoded| {
             switch (hardcoded) {
                 .@"bun:main" => {
-                    defer {
-                        if (jsc_vm.worker) |worker| {
-                            worker.queueInitialTask();
-                        }
-                    }
+                    // defer {
```
Jarred-Sumner: is this intentional?

src/bun.js/web_worker.zig
```diff
@@ -55,17 +50,10 @@ pub const WebWorker = struct {
...
         JSC.markBinding(@src());
 
-        if (this.vm == null) {
-            return !this.requested_terminate;
-        }
-
-        if (!this.allowed_to_exit) {
-            return true;
+        if (this.vm) |vm| {
+            return vm.eventLoop().tasks.count > 0 or vm.active_tasks > 0 or vm.uws_event_loop.?.active > 0;
```
Jarred-Sumner: I think this method might not actually be called

but if it is, this code isn't threadsafe

src/bun.js/web_worker.zig
```diff
@@ -20,11 +20,9 @@ pub const WebWorker = struct {
     arena: bun.MimallocArena = undefined,
     name: [:0]const u8 = "Worker",
     cpp_worker: *anyopaque,
-    allowed_to_exit: bool = false,
     mini: bool = false,
-    parent_poll_ref: JSC.PollRef = .{},
-    initial_poll_ref: JSC.PollRef = .{},
-    did_send_initial_task: bool = false,
+    user_poll_ref: JSC.PollRef = .{},
+    eventloop_poll_ref: JSC.PollRef = .{},
```
Jarred-Sumner: `parent_poll_ref`? both of these are the event loop

src/bun.js/web_worker.zig
```diff
@@ -303,30 +276,29 @@ pub const WebWorker = struct {
...
-                    vm.eventLoop().autoTickActive();
-                }
-
-                if (!this.allowed_to_exit) {
-                    this.flushLogs();
-                    vm.eventLoop().tickPossiblyForever();
-                    continue;
+        while (true) {
+            while (vm.eventLoop().tasks.count > 0 or vm.active_tasks > 0 or vm.uws_event_loop.?.active > 0) {
+                this.eventloop_poll_ref.refConcurrently(this.parent);
```
Jarred-Sumner: this should not be here

src/bun.js/web_worker.zig
```diff
@@ -303,30 +276,29 @@ pub const WebWorker = struct {
...
+            if (this.vm == null) {
+                this.deinit();
+                return;
             }
 
             this.flushLogs();
-            this.onTerminate();
+            this.eventloop_poll_ref.unrefConcurrently(this.parent);
+            vm.eventLoop().tickPossiblyForever();
+            this.parent.eventLoop().wakeup();
```
Jarred-Sumner: there is a race condition here where the parent event loop is unref'd

Jarred-Sumner: for the case where the parent event loop is not the main thread

paperclover: oh i did not think about workers starting other workers

src/bun.js/web_worker.zig
paperclover: this was just refactoring the left without allowed_to_exit; so the original code wouldn't be threadsafe either?

src/bun.js/web_worker.zig
paperclover: i'll rename it back but the idea is one is the user's .ref/.unref, the other one is doing ref only if the worker's event loop has tasks. they both ref on the parent.

Jarred-Sumner: that makes sense, the name should be clearer though

src/bun.js/module_loader.zig
paperclover: worker.queueInitialTask was removing an eventloop ref way too early. since the ref was now based on the worker's event loop, the initial task to unref isn't needed i think. did not mean to leave as commented code.

src/bun.js/web_worker.zig
paperclover: i meant to have this at the end of the while(true) (didnt push that commit because was looking into other bugs)

Jarred-Sumner: It should not call .refConcurrently() on every tick of the event loop? It should only need to call it once while there are pending tasks.

src/bun.js/base.zig
```diff
@@ -2122,6 +2122,11 @@ pub export fn MarkedArrayBuffer_deallocator(bytes_: *anyopaque, _: *anyopaque) v
     // zig's memory allocator interface won't work here
     // mimalloc knows the size of things
     // but we don't
+    if (comptime Environment.allow_assert) {
```
paperclover: [existing bug reproduced by sveltekit build. this assertion makes it fail consistantly]

src/bun.js/module_loader.zig
paperclover: code here was removed instead of commented

test/js/third_party/esbuild/esbuild-test.js
```diff
@@ -82,5 +82,3 @@ import { build, buildSync, transform, transformSync } from "esbuild";
     throw new Error("Test failed.");
   }
 }
-
-process.exit(0);
```
paperclover, line 86: probably was not needed, but a reason to keep this is that the process should exit since there are no ref'd polls.

Jarred-Sumner (approved):

paperclover:
bug with `terminate()` but usable

Douile:
This PR adds documentation about compiling a specific version of ZLS that works better which is after the [Install zig](https://bun.sh/docs/project/development#install-zig) part of the docs. But the version of zig installed by the install zig section (0.11.0-dev.4004+a57608217) is too old to compile the specific version of ZLS (v0.11.0-dev.4332+43b830415 is required). Perhaps its worth adding something that mentions this if this is intended.

https://github.com/oven-sh/bun/commit/78defe7a87226b5b10766e24fae458a62811dab2#diff-f0c08b47e6be8e8b574a71f652c3c2c66f49d157b392f3091c58e8585d411010R183-R202

PR #4126, opened 2023-08-12
`bun-wasm` fixes & improvements

jhmaster2000:
### What does this PR do?

* Fix WASM build on Linux by unhardcoding mimalloc filename in Makefile
* Fix most TS errors in `bun-wasm`, add missing types, stronger types
* Add scripts to document the process of building `bun-wasm`
* Improve usability of `bun-wasm` as a proper package:
  * `init()` can now be called with no args and will resolve to the default location of next to the built index.js
    * The url argument still exists to be used if overriding the default wasm file location is needed.
    * To use the default location and pass the other args such as heap size, `null` can be given as the url.
  * Split `index.d.ts` into `index.d.cts` and `index.d.mts`, these are exact copies of each other but are necessary for optimal editor intellisense compatibility, as it tries to map the output `index.mjs` and `index.cjs` files to definition files of matching extension.

There's also some more isolated bun-polyfills progress as I branched this branch from the bun-polyfills branch by accident but that should be fine. (no bun-types changes included this time)

Manual testing based on the `bun-wasm/test/node.mjs` file. (`bun-wasm` is not part of the test suite yet)


Jarred-Sumner:
nice

PR #4127, opened 2023-08-12
fetch(stream) add stream support for compressed and uncompressed data

cirospaciari:
### What does this PR do?

- [x] Code changes

I wrote automated tests + existing tests


misctools/fetch.zig
```diff
@@ -187,7 +187,7 @@ pub fn main() anyerror!void {
     var ctx = try default_allocator.create(HTTP.HTTPChannelContext);
     ctx.* = .{
         .channel = channel,
-        .http = try HTTP.AsyncHTTP.init(default_allocator, args.method, args.url, args.headers, args.headers_buf, response_body_string, args.body, 0, HTTP.FetchRedirect.follow),
+        .http = try HTTP.AsyncHTTP.init(default_allocator, args.method, args.url, args.headers, args.headers_buf, response_body_string, args.body, 0, HTTP.FetchRedirect.follow, false),
```
Jarred-Sumner: can we make this an options struct

hard to know what `false` here means

cirospaciari: Sure, maybe adding the min buffer size for the streaming too.

src/deps/uws.zig
```diff
@@ -200,6 +200,24 @@ pub fn NewSocketHandler(comptime is_ssl: bool) type {
                 this.socket,
             ).?;
         }
+
+        pub fn pausePoll(this: ThisSocket) i32 {
+            const p = bun.cast(*Poll, this.socket);
```
Jarred-Sumner: I suggest adding a `.poll` fn to `ThisSocket` instead

janwilmake:
GREAT WORK! Looking forward to this one.

src/bun.js/unbounded_queue.zig
```diff
@@ -73,6 +73,8 @@ pub fn UnboundedQueue(comptime T: type, comptime next_field: meta.FieldEnum(T))
         }
 
         pub fn pop(self: *Self) ?*T {
+            if (self.isEmpty()) return null;
```
Jarred-Sumner: What happens if `count` is incremented after `isEmpty()` returns true?

Why isn't it sufficient for `front` to be null and then return?

src/bun.js/webcore/body.zig
```diff
@@ -376,6 +377,14 @@ pub const Body = struct {
                 .Blob => this.Blob.size,
                 .InternalBlob => @as(Blob.SizeType, @truncate(this.InternalBlob.sliceConst().len)),
                 .WTFStringImpl => @as(Blob.SizeType, @truncate(this.WTFStringImpl.utf8ByteLength())),
+                .Locked => {
+                    if (this.Locked.readable) |readable| {
```
Jarred-Sumner: I suggest moving this into a helper fn incase we use it in more places

src/bun.js/webcore/response.zig
```diff
@@ -974,9 +1139,24 @@ pub const Fetch = struct {
...
+            task.body_size = result.body_size;
+
+            const success = result.isSuccess();
+            task.response_buffer = result.body.?.*;
+
+            if (success) {
+                _ = task.scheduled_response_buffer.write(task.response_buffer.list.items) catch @panic("OOM");
+            }
+
+            if (!task.has_schedule_callback.load(.Acquire)) {
```
Jarred-Sumner: This is a race condition. The value can change between this load and the store below it. Instead, you want to use one of the compare and exchange ones.

src/bun.js/unbounded_queue.zig
cirospaciari: I think I can revert this, this only evades the problem but is still race, for some reason we are getting an assertion error on `assert(@atomicRmw(usize, &self.count, .Sub, 1, .Monotonic) >= 1);`

image

Jarred-Sumner: Maybe we should use `popBatch`

src/bun.js/webcore/response.zig
cirospaciari: https://github.com/oven-sh/bun/pull/4127/commits/76b4e7bfdb810d77e2e8e36213563b00d77edb6f changed for compare and swap

src/bun.js/webcore/body.zig
cirospaciari: done in https://github.com/oven-sh/bun/pull/4127/commits/76b4e7bfdb810d77e2e8e36213563b00d77edb6f

src/bun.js/webcore/body.zig
```diff
@@ -371,11 +372,22 @@ pub const Body = struct {
...
+
+            return this.Locked.size_hint;
+        }
+
         pub fn size(this: *const Value) Blob.SizeType {
             return switch (this.*) {
                 .Blob => this.Blob.size,
                 .InternalBlob => @as(Blob.SizeType, @truncate(this.InternalBlob.sliceConst().len)),
                 .WTFStringImpl => @as(Blob.SizeType, @truncate(this.WTFStringImpl.utf8ByteLength())),
+                .Locked => {
```
Jarred-Sumner:
```suggestion
                .Locked => this._lcokedSize()
```

src/bun.js/webcore/response.zig
```diff
@@ -619,23 +621,32 @@ pub const Fetch = struct {
         javascript_vm: *VirtualMachine = undefined,
         global_this: *JSGlobalObject = undefined,
         request_body: HTTPRequestBody = undefined,
+        // buffer being used by AsyncHTTP
```
Jarred-Sumner: use `///` because `//` doesn't show up in autocomplete

src/bun.js/webcore/response.zig
```diff
@@ -708,9 +720,100 @@ pub const Fetch = struct {
...
+                                        .temporary_and_done = bun.ByteList.initConst(chunk),
+                                    },
+                                    bun.default_allocator,
+                                );
+                            }
+
+                            return;
+                        }
+                    } else {
+                        this.response.?.body.value.Locked.size_hint = @as(u52, @intCast(this.body_size));
```
Jarred-Sumner: is the pointer aliasing here intentional?

src/bun.js/webcore/response.zig
```diff
@@ -708,9 +720,100 @@ pub const Fetch = struct {
...
+                }
+            }
+
+            if (!success) {
+                const err = this.onReject();
+                globalThis.throwValue(err);
+                return;
+            }
+
+            if (this.response) |response| {
```
Jarred-Sumner: What is keeping this response alive? How can we be sure that the garbage collector doesn't free the Response object in the middle of streaming?

Jarred-Sumner:
I think this needs about two dozen more tests

src/bun.js/webcore/response.zig
cirospaciari: fixed https://github.com/oven-sh/bun/pull/4127/commits/4676cfe8c877ca99e2fb4f2598afb51b51ceaadc

src/bun.js/webcore/response.zig
cirospaciari: fixed https://github.com/oven-sh/bun/pull/4127/commits/4676cfe8c877ca99e2fb4f2598afb51b51ceaadc

src/bun.js/webcore/response.zig
cirospaciari: fixed https://github.com/oven-sh/bun/pull/4127/commits/4676cfe8c877ca99e2fb4f2598afb51b51ceaadc

src/bun.js/webcore/body.zig
cirospaciari: done in https://github.com/oven-sh/bun/pull/4127/commits/4676cfe8c877ca99e2fb4f2598afb51b51ceaadc

Jarred-Sumner:
looks like macOS build fails

Jarred-Sumner:
@cirospaciari this still needs more tests. 

there should be tests for:
- large sizes
- small sizes
- matrix of [large chunk, large chunk, small chunk, small chunk, empty chunk] x [large chunk, large chunk, small chunk, small chunk, empty chunk]
- incomplete chunk of data from zlib
- strings which become invalid UTF-8
- several simultaneous bodies streaming
- returning a Response in Bun.serve() which comes from a streaming fetch request `Bun.serve({async fetch(req) { return new Response(await fetch("https://big-slow-api.com"), {headers: req.headers, method: req.method})); });`
- What happens when you do `await Bun.write("/tmp/hey.txt", await fetch("https://big-slow-api.com"))`?
- What happens when you use `pipeThrough()` with a streaming body and send it to some other JS library, pipeThrough() that, then return it as the Response body in Bun.serve()'s fetch() callback?

src/bun.js/webcore/body.zig
```diff
@@ -370,11 +371,20 @@ pub const Body = struct {
             }
         }
 
+        fn _lockedSize(this: *const Value) Blob.SizeType {
+            if (this.Locked.readable) |readable| {
+                return readable.ptr.Bytes.size_hint;
```
Jarred-Sumner: This needs to check if `Bytes` is the active tag or it will segfault.

src/bun.js/webcore/body.zig
```diff
@@ -370,11 +371,20 @@ pub const Body = struct {
             }
         }
 
+        fn _lockedSize(this: *const Value) Blob.SizeType {
```
Jarred-Sumner: Why is this a method on `Value` instead of `Locked`?  Why does it have an `_` in the name (it's a private method, that is unnecessary)

src/bun.js/webcore/body.zig
```diff
@@ -370,11 +371,20 @@ pub const Body = struct {
...
+
+            return this.Locked.size_hint;
+        }
+
         pub fn size(this: *const Value) Blob.SizeType {
             return switch (this.*) {
                 .Blob => this.Blob.size,
                 .InternalBlob => @as(Blob.SizeType, @truncate(this.InternalBlob.sliceConst().len)),
                 .WTFStringImpl => @as(Blob.SizeType, @truncate(this.WTFStringImpl.utf8ByteLength())),
+                .Locked => this._lockedSize(),
```
Jarred-Sumner: Can you add a comment explaining why this is safe? It's not clear from reading this if `_lockedSize` returns the currently buffered size (which may not be the entire size in the case of chunked transfer encoding) or if it returns the whole size. It's also unclear what happens when the size is not known.

This method has to be accurate or else the HTTP response will potentially hang/timeout. So if we only return the estimated size, it's incorrect.

src/bun.js/webcore/response.zig
```diff
@@ -618,23 +620,32 @@ pub const Fetch = struct {
...
         response_buffer: MutableString = undefined,
+        /// buffer used to stream response to JS
+        scheduled_response_buffer: MutableString = undefined,
+        /// response strong ref
+        response: JSC.Strong = .{},
         request_headers: Headers = Headers{ .allocator = undefined },
         promise: JSC.JSPromise.Strong,
         concurrent_task: JSC.ConcurrentTask = .{},
         poll_ref: JSC.PollRef = .{},
+        body_size: usize = 0,
```
Jarred-Sumner: This variable name should explicitly make it clear whether it's the total body size or the current, buffered, or received body size

src/bun.js/webcore/body.zig
Jarred-Sumner: i think i suggested the name, sorry

src/bun.js/bindings/bindings.cpp
```diff
@@ -1181,7 +1181,7 @@ WebCore::FetchHeaders* WebCore__FetchHeaders__createFromPicoHeaders_(const void*
 
         for (size_t j = 0; j < end; j++) {
             PicoHTTPHeader header = pico_headers.ptr[j];
-            if (header.value.len == 0)
+            if (header.value.len == 0 || header.name.len == 0)
```
Jarred-Sumner, line 1184: interesting

src/bun.js/webcore/blob.zig
```diff
@@ -3794,6 +3794,10 @@ pub const Blob = struct {
                         } else {
                             return build.blob.dupe();
                         }
+                    } else if (current.toSliceOrNull(global)) |sliced| {
+                        if (sliced.allocator.get()) |allocator| {
+                            return Blob.initWithAllASCII(bun.constStrToU8(sliced.slice()), allocator, global, false);
```
Jarred-Sumner, line 3799: This should be cloned. There are too many cases where we assume Blob is always owned by mimalloc, but in this case it might not be.

Jarred-Sumner: also if you do .arrayBuffer() on a Response it will transfer the buffer which would mean that the string is now mutable which will break an unknown number of things.

src/bun.js/webcore/body.zig
```diff
@@ -385,15 +400,18 @@ pub const Body = struct {
...
                 .WTFStringImpl => @as(Blob.SizeType, @truncate(this.WTFStringImpl.byteSlice().len)),
+                .Locked => this.Locked.sizeHint(),
                 // .InlineBlob => @truncate(Blob.SizeType, this.InlineBlob.sliceConst().len),
                 else => 0,
             };
         }
 
         pub fn estimatedSize(this: *const Value) usize {
             return switch (this.*) {
+                .Blob => this.Blob.size,
```
Jarred-Sumner: This is not correct. `Blob.size` defaults to a max value which is an absurdly large number to indicate that it is not a valid number. This would need to call `resolveSize` which would be too expensive to call here because it potentially does `stat()` on a file in the filesystem.

src/bun.js/webcore/response.zig
```diff
@@ -784,25 +922,78 @@ pub const Fetch = struct {
...
+                };
+            }
+
+            return .{
+                .estimated_size = size_hint,
             };
+        }
+
+        fn getSizeHint(this: *FetchTasklet) u52 {
+            if (this.body_size == .content_length) {
```
Jarred-Sumner: it would be more idiomatic Zig for this to be a switch statement

src/bun.js/webcore/response.zig
```diff
@@ -784,25 +922,78 @@ pub const Fetch = struct {
...
+                    },
+                };
+            }
+
+            return .{
+                .estimated_size = size_hint,
             };
+        }
+
+        fn getSizeHint(this: *FetchTasklet) u52 {
```
Jarred-Sumner: Can you use `Blob.SizeType` instead of directly using `u52`? I think that makes it more clear that it is consistent

src/bun.js/webcore/response.zig
```diff
@@ -784,25 +922,78 @@ pub const Fetch = struct {
...
+            if (this.body_size == .content_length) {
+                return @as(u52, @intCast(this.body_size.content_length));
```
Jarred-Sumner: This should be truncate and the `@as` is unnecessary

src/bun.js/webcore/response.zig
```diff
@@ -830,7 +1022,10 @@ pub const Fetch = struct {
             const allocator = bun.default_allocator;
             var response = allocator.create(Response) catch unreachable;
             response.* = this.toResponse(allocator);
-            return Response.makeMaybePooled(@as(js.JSContextRef, @ptrCast(this.global_this)), response);
+            const response_js = Response.makeMaybePooled(@as(js.JSContextRef, @ptrCast(this.global_this)), response);
```
Jarred-Sumner: the `@ptrCast` here does nothing btw

src/bun.js/webcore/streams.zig
```diff
@@ -4064,17 +4101,21 @@ pub const File = struct {
             this.concurrent.read = @as(Blob.SizeType, @truncate(result catch |err| {
                 if (@hasField(HTTPClient.NetworkThread.Completion, "result")) {
                     this.pending.result = .{
-                        .err = Syscall.Error{
-                            .errno = @as(Syscall.Error.Int, @intCast(-completion.result)),
-                            .syscall = .read,
+                        .err = .{
```
Jarred-Sumner, line 4114: err.err is kinda weird

cirospaciari: yeah will rename to `sys_err`

src/bun.js/webcore/blob.zig
cirospaciari: weird `toSliceOrNull` dont clone it? 
https://github.com/oven-sh/bun/blob/main/src/bun.js/bindings/bindings.zig#L4193-L4198

test/js/web/fetch/fetch.stream.test.ts
```diff
@@ -0,0 +1,1129 @@
...
+            const { done, value } = (await reader?.read()) as ReadableStreamDefaultReadResult<any>;
+            if (value) {
+              buffer = Buffer.concat([buffer, value]);
+            }
+            if (done) {
+              break;
+            }
+          }
+          gcTick(false);
+          expect(buffer.byteLength).toBe(content.byteLength);
```
Jarred-Sumner: This only checks the length. It needs to check the value too. `expect(buffer).toEqual(content)`

Jarred-Sumner:
Thank you for writing the tests. I think we can ship this

test/js/web/fetch/fetch.stream.test.ts
cirospaciari: Will add it

Jarred-Sumner (approved):
any idea why the fsevents test is failing?

src/bun.js/event_loop.zig
```diff
@@ -545,8 +545,7 @@ pub const EventLoop = struct {
                 },
                 .FetchTasklet => {
                     var fetch_task: *Fetch.FetchTasklet = task.get(Fetch.FetchTasklet).?;
-                    fetch_task.onDone();
-                    fetch_task.deinit();
+                    fetch_task.onProgressUpdate();
```
Jarred-Sumner, line 548: when does the task itself get freed?

cirospaciari:
> any idea why the fsevents test is failing?

no idea, this poped up after the last rebase

src/bun.js/webcore/blob.zig
Jarred-Sumner: toSlice does not if it's an ASCII latin1 string. toOwnedSlice clones it.

src/bun.js/event_loop.zig
cirospaciari: it self `deinit` after receiving the `has_more == false` after `clearData` calls

src/bun.js/webcore/streams.zig
```diff
@@ -525,9 +525,19 @@ pub const StreamResult = union(Tag) {
     into_array: IntoArray,
     into_array_and_done: IntoArray,
     pending: *Pending,
-    err: Syscall.Error,
+
+    err: union(Err) {
```
Jarred-Sumner: maybe do

```
Error: Syscall.Error
JSValue: JSC.JSValue
```

That JSvalue version also needs to be kept alive somehow.

src/bun.js/event_loop.zig
Jarred-Sumner: but the Tasklet part. The thing which is queued to the tasks queue.

cirospaciari: > but the Tasklet part. The thing which is queued to the tasks queue.

Oh I see dummy mistake, let me fix it ASAP

src/bun.js/webcore/blob.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d

src/bun.js/webcore/response.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d

src/bun.js/webcore/response.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d

src/bun.js/webcore/response.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d

src/bun.js/webcore/response.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d

src/bun.js/webcore/body.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d

src/bun.js/event_loop.zig
cirospaciari: Actually only one task is allocated, and on the last callback it is freed. We use `task.javascript_vm.eventLoop().enqueueTaskConcurrent(task.concurrent_task.from(task, .manual_deinit));` so we use the same pointer to the task, not need to be freed, on each callback only in the last one the deinits are:

https://github.com/oven-sh/bun/blob/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d/src/bun.js/webcore/response.zig#L735

https://github.com/oven-sh/bun/blob/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d/src/bun.js/webcore/response.zig#L855

https://github.com/oven-sh/bun/blob/ef6989f4118ee842adf88f83c9d2b3ba64f04f0d/src/bun.js/webcore/response.zig#L863

test/js/web/fetch/fetch.stream.test.ts
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/4127/commits/70d8e6941743d8ae072946846f015751201d1601

src/bun.js/webcore/streams.zig
cirospaciari: Fixed on https://github.com/oven-sh/bun/pull/4127/commits/4d07c988db7ed2a783567aa8fb07dec1bfd4dda3

Jarred-Sumner:
🥳 🥳 🥳

PR #4146, opened 2023-08-13
Fix issue with test names containing back ticks (" ` ")

ian-weir:
### What does this PR do?

Fixes an issue where any test names that included back ticks (" ` ") were not processed correctly and would result in misreported failures.

- [X] Code changes

I wrote automated tests, also tested it with a dummy project.

[ N/A? ] I ran `make js` and committed the transpiled changes

[ N/A ] I included a test for the new code, or an existing test covers it

 [ N/A ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed


src/bun.js/test/expect.zig
```diff
@@ -220,6 +220,16 @@ pub const Expect = struct {
 
     pub fn getSnapshotName(this: *Expect, allocator: std.mem.Allocator, hint: string) ![]const u8 {
         const test_name = this.scope.tests.items[this.test_id].label;
+        var name_builder = MutableString.initEmpty(allocator);
+
+        for (test_name) |value| {
```
Jarred-Sumner, line 225: 1) Since this is an edgecase, let's avoid the allocation unless we know for sure there's a backtick quote character:
```zig
var escaped_name = test_name;
{
  if (strings.indexOfChar(test_name, '`')) |start_i| {
      ...
  }
}
```

2) Instead of copying the string byte-by-byte, which is slow, let's find the index of the next backtick character, copy everything before that index, and then copy the escape + backtick character. And then don't forget to append at the end to handle any leftover after the backtick character.

src/bun.js/test/expect.zig
```diff
@@ -220,6 +220,16 @@ pub const Expect = struct {
 
     pub fn getSnapshotName(this: *Expect, allocator: std.mem.Allocator, hint: string) ![]const u8 {
         const test_name = this.scope.tests.items[this.test_id].label;
+        var name_builder = MutableString.initEmpty(allocator);
```
Jarred-Sumner, line 223: We can stick the `defer name_builder.deinit()` here to be more careful that it never leaks

test/js/bun/test/snapshot-tests/snapshots/__snapshots__/snapshot.test.ts.snap
```diff
@@ -368,3 +368,11 @@ exports[`most types: testing 7 2`] = `9`;
 exports[`most types: testing 7 3`] = `8`;
 
 exports[`most types: undefined 1`] = `undefined`;
+
+exports[`should work with expect.anything() 1`] = `
```
Jarred-Sumner, line 372: oh if we change these to be UTF16 strings, i think it will properly escape them including handling other cases like unicode escape sequences and octal literals and the like

src/bun.js/test/expect.zig
ian-weir: For point 1, why would we use shadowing for the if block?

test/js/bun/test/snapshot-tests/snapshots/__snapshots__/snapshot.test.ts.snap
ian-weir: Cool, I'll give that a shot

marvinside:
There seems to be issues not only for escaping the Test name, but also the snapshot value.

See: 
* https://github.com/oven-sh/bun/issues/4967
* https://github.com/oven-sh/bun/issues/5272

Maybe this can be fixed in the same PR

marvinside:
After having a look at `src/bun.js/test/expect.zig`, it seems to me this file is not the part that needs fixing, but rather the jest part which writes the snapshot file.

The Snapshot Name itself may contain any chars, but at the part of writing to disk in the snapshot format it needs the escaping. Escaping here is too early in my opinion.

Same for the snapshot value.

(Also I have never worked with Zig, just having a look here)

Jarred-Sumner:
Closing due to bitrot

PR #4157, opened 2023-08-15
Fix event loop issue with `Bun.connect`

### LINKED ISSUE #4108: `async.series` doesn't keep process alive

0.7.3

Darwin 22.6.0 arm64 arm

Here's a cloneable repo to make things easier: https://github.com/nwalters512/bun-pg-fs-reproduction. Prerequisites and reproduction steps can be found in the `README.md`.

When running the unmodified `index.ts` file, I would expect to see the following output printed:

```
connecting pool...
connected pool
succeeded!
```

### What do you see instead?

Instead, only the following is printed:

```
connecting pool...
```

### Additional information

After uncommenting the `fs.readFile` line, things work as expected and the expected output is printed:

```
connecting pool...
connected pool
succeeded!
```

Note that moving things out of `async.series` and to the root of the file also makes things behave as expected. That is, both `async.series` and `fs.readFile` seem to be necessary for things to break.

paperclover:
### What does this PR do?

Before, the event loop is not reffed between the time `Bun.connect` is called and the `open` callback is emitted. We are relying on the `await` to keep the event loop on.

This fixes #4108 as well as this minimal repro.

```ts
(async () => {
  await Bun.sleep(10);
  console.log("connecting pool...");
  await Bun.connect({
    hostname: "localhost",
    port: 5432,
    socket: { data() {} },
  });
  console.log("connected pool");
})();
```

Though it makes me think if we'll need more cases for keeping the event loop alive, as we should not have to rely on top-level await to carry us. Later we should investigate if `Bun.build` and any other promise-returning API has this sort of issue.

- [X] Code changes

I verified the current tests work as well as the issue did not reproduce.

### DIFF
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -999,7 +999,6 @@ pub const Listener = struct {
                 .protos = null,
                 .server_name = null,
             };
-            tcp.poll_ref.ref(handlers.vm);
 
             TCPSocket.dataSetCached(tcp.getThisValue(globalObject), globalObject, default_data);
 
@@ -1012,6 +1011,7 @@ pub const Listener = struct {
                 exception.* = ZigString.static("Failed to connect").toErrorInstance(globalObject).asObjectRef();
                 return .zero;
             };
+            tcp.poll_ref.ref(handlers.vm);
 
             return promise_value;
         }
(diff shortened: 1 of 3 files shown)


Jarred-Sumner (changes requested):
Please include a test that previously did not pass

Jarred-Sumner (approved):

Issue #3844, opened 2023-07-27
Bun overrides "ws" package even when --target=node


jackson-bean:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Microsoft Windows NT 10.0.22621.0 x64 | Using WSL

### What steps can reproduce the bug?

- `bun init`

```typescript
// index.ts
import { WebSocket } from "ws";

console.log(WebSocket);
```

- `bun build --target=node --outdir=dist index.ts`

```javascript
// dist/index.js
import {WebSocket} from "ws";
console.log(WebSocket);
```

### What is the expected behavior?

Bun should bundle the original/node "ws" package since --target=node.

### What do you see instead?

The bundle contains the import without the "ws" package bundled with it.

jackson-bean:
Trying to get around this by using a plugin in the meantime. I don't know if it's just that I don't know much about making plugins, but when I try to build with the plugin, no output files are generated and no errors are thrown. And yes, "outdir" is specified. Don't know if this is an issue with me or bun.

Jarred-Sumner:
confirming that it is a bug that it is using bun's module list to mark as external when --target=node instead of using a different list for node

paperclover:
resolver.zig line 592 is where we are using .Aliases, which contains ws and all the bun modules. should we just make a separate comptime list for this?

paperclover:
this is also a problem for `@vercel/fetch` or any other module we override at runtime.

Hanaasagi:
trying to work on this

PR #4163, opened 2023-08-15
Fix(bundler): use different alias mappings based on the target.

Hanaasagi:
### What does this PR do?

Close: #3844

1. When bundling the code, use different alias mappings based on the target.

- [x] Code changes

I wrote automated tests.


src/bun.js/module_loader.zig
```diff
@@ -2335,165 +2335,208 @@ pub const HardcodedModule = enum {
...
-            .{ "undici", .{ .path = "undici" } },
-            .{ "utf-8-validate", .{ .path = "utf-8-validate" } },
-            .{ "ws", .{ .path = "ws" } },
-            .{ "ws/lib/websocket", .{ .path = "ws" } },
-        },
-    );
+
+    pub const Aliases = struct {
+        // Used by both Bun and Node.
+        const CommonAliases = bun.ComptimeStringMap(
```
Jarred-Sumner: can we make this two separate ComptimeStringMap with all the values for each? That way there's only one hash table lookup for each target, instead of two. The `++` operator should work to combine two comptime-known arrays of tuples

Hanaasagi: yes, fixed.

Jarred-Sumner:
Thank you. Sorry I took awhile to merge this

PR #4176, opened 2023-08-16
add "bun up" alias of "bun update"

hyuzea:
### What does this PR do?

- [x] Saves next four characters
- [x] Code changes


v3nntrnx:
In my opinion it is a nice idea and the code LGTM. Though please note im not someone from the core team, i'm just saying my opinion.

colinhacks:
Not sure about this.

- It would be nice to avoid the alias hell of the `npm` CLI. Better to have one right way to do things, by and large.
- This is extra ambiguous with update vs upgrade. It'll already be hard for people to remember which is which and this is another layer of potential confusion.

Jarred-Sumner:
I agree with @colinhacks, I don't think this shorthand is worth preventing people from being able to use a script from package.json called "up" via `bun up`

PR #4186, opened 2023-08-17
Improved support for `debug-adapter-protocol`

Electroid:
### What does this PR do?

Made various changes and improvements to how the `debug-adapter-protocol` works.


package.json
```diff
@@ -9,6 +9,7 @@
     "prettier": "^2.4.1",
     "react": "next",
     "react-dom": "next",
+    "source-map-js": "^1.0.2",
```
Jarred-Sumner, line 12: this should go in the debug adapter package.json, not the root one

packages/bun-debug-adapter-protocol/debugger/sourcemap.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const defaultSourceMap = new NoopSourceMap();
+
+export function SourceMap(url?: string): SourceMap {
+  if (!url || !url.startsWith("data:")) {
+    return defaultSourceMap;
+  }
+  try {
+    const [_, base64] = url.split(",", 2);
+    const decoded = Buffer.from(base64, "base64").toString("utf8");
```
Jarred-Sumner: you _should_ be able to use ascii in this case because valid base64 strings are ascii

PR #4190, opened 2023-08-17
replace bunx to bun x

AbdallahAskar1:
### What does this PR do?

- [ ] Code changes


colinhacks:
Why? We use `bunx` primarily throughout the rest of the docs (this is intentional)

Jarred-Sumner:
I think we need to fix a bug with symlinking bunx so that people don’t run into issues with it not existing, but that’s not a documentation problem. 

Thank you for this, but we prefer to keep it as bunx and not bun x

colinhacks:
Seems symlinking fails in codespaces, might be useful for reproduction: https://github.com/microsoft/playwright/pull/26454#discussion_r1293220002

Issue #4178, opened 2023-08-16
TypeError: undefined is not an object (evaluating 'buf.length')


colinhacks:
### What version of Bun is running?

latest canary

### What platform is your computer?

macos

### What steps can reproduce the bug?

This bug breaks the following commands.
```
bunx --bun create-solid
# and
bunx --bun nuxi init
``` 

Seems to be an issue with `fs-minipass`. Minimal repro:

```ts
const fsm = require('fs-minipass');
const readStream = new fsm.ReadStream('file.txt');
```

Error:

```
3830 |   [_handleChunk] (br, buf) {
3831 |     let ret = false;
3832 |     // no effect if infinite
3833 |     this[_remain] -= br;
3834 |     if (br > 0)
3835 |       ret = super.write(br < buf.length ? buf.slice(0, br) : buf);
                                 ^
TypeError: undefined is not an object (evaluating 'buf.length')
      at /Users/colinmcd94/Documents/bun/fun/node_modules/fs-minipass/lib/index.js:141:29
      at /Users/colinmcd94/Documents/bun/fun/node_modules/fs-minipass/lib/index.js:117:15
```

Hanaasagi:
fs-minipass use `fs.read(fd, buffer, offset, length, position, callback)`


https://github.com/npm/fs-minipass/blob/8348d32797eadf1bad05fae1d8ba2af3da53cd44/lib/index.js#L108-L110

Minimal reproducible code without third-party dependencies.


```JavaScript
const fs = require("node:fs");
const assert = require("node:assert");

const filePath = "file.txt";
fs.writeFileSync(filePath, "hello");

const buffer = Buffer.alloc(16);
const offset = 0;
const length = buffer.length;
const position = 0;

fs.open(filePath, "r", (err, fd) => {
  fs.read(fd, buffer, offset, length, position, (err, bytesRead, buffer) => {
    assert.notEqual(buffer, undefined);
  });
});

```

Ref: https://nodejs.org/api/fs.html#fsreadfd-buffer-offset-length-position-callback

PR #4191, opened 2023-08-17
Fix(node:fs): add `buffer` parameter in `fs.read` callback.

Hanaasagi:
Close: #4178

### What does this PR do?

1. Add `buffer` parameter in `fs.read` callback. Ref: https://nodejs.org/api/fs.html#fsreadfd-buffer-offset-length-position-callback

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
Thank you for this. We should move fs.read to be actually async, but this is better than the current state so it's worth merging.

PR #4192, opened 2023-08-17
correct guide's bunfig example option

huseeiin:
### What does this PR do?

Changed example bunfig.toml to use "install.lockfile" instead of "install"

- [ ] Code changes


colinhacks:
Thanks!

PR #4193, opened 2023-08-17
refactor: move HTMLRewriter to c++ bindings

bru02:
### What does this PR do?

- [x] Code changes

Using [this](https://gist.github.com/bru02/e3ee5a79c6c9ab20bc47a846c25e0483) real-wordish HTMLRewriter on ~59k records.

Here are the bench results:
Screenshot 2023-08-17 at 18 07 35


simylein:
This looks awesome 🙂

Jarred-Sumner:
Haven’t read code yet but exciting benchmark results, especially considering the faster one is the debug build of bun.

src/bun.js/api/html_rewriter.classes.ts
```diff
@@ -0,0 +1,263 @@
...
+        getter: "getName",
+        setter: "setName",
+      },
+    },
+  }),
+  define({
+    name: "AttributeIterator",
+    construct: true,
+    finalize: true,
+    JSType: "0b11101110",
```
Jarred-Sumner, line 166: There's a JSC function for creating a `{next, value}` object but we don't have a binding for it right now. This current implementation will be pretty slow due to not re-using the underlying JSC::Structure.

We should add a binding to this function: https://github.com/oven-sh/WebKit/blob/main/Source/JavaScriptCore/runtime/IteratorOperations.cpp#L147

src/bun.js/api/html_rewriter.zig
```diff
@@ -1117,85 +1036,64 @@ pub const TextChunk = struct {
...
-                .rfn = wrap(DocEnd, "append"),
-            },
-        },
-        .{},
-    );
+    pub usingnamespace JSC.Codegen.JSDocEnd;
 
-    fn contentHandler(this: *DocEnd, comptime Callback: (fn (*LOLHTML.DocEnd, []const u8, bool) LOLHTML.Error!void), thisObject: js.JSObjectRef, globalObject: *JSGlobalObject, content: ZigString, contentOptions: ?ContentOptions) JSValue {
+    fn contentHandler(this: *DocEnd, comptime Callback: (fn (*LOLHTML.DocEnd, []const u8, bool) LOLHTML.Error!void), callFrame: *JSC.CallFrame, globalObject: *JSGlobalObject, content: ZigString, contentOptions: ?ContentOptions) JSValue {
         if (this.doc_end == null)
```
Jarred-Sumner, line 1097: one thing to be aware of is that `callFrame` is a pointer to a stack frame and things can get moved around as the stack frame changes a little 

So we should prefer to reference all variables from the call frame at the beginning of the callframe, and not pass it around to non-inline functions

In this case, it's super cheap to get the `this` value. It's a pointer lookup without other side effects

Jarred-Sumner:
This is a clear improvement over the current state, removes longstanding technical debt and improves performance. There are a couple small things to address but those can be addressed in a follow-up PR. 

Thank you. Merged.

PR #4194, opened 2023-08-17
Add `util.inspect.custom` support to `util.inspect/Bun.inspect/console.log`

paperclover:
### What does this PR do?

This merges `util.inspect` and `Bun.inspect`, as well as gives the console formatter support for calling `util.inspect.custom` functions.

**Breaking change**: `Bun.inspect` instead of taking variadic arguments, it will function exactly like `util.inspect`.

- Code changes

- [X] I wrote automated tests
- [X] I added TypeScript types for the new methods, getters, or setters


Jarred-Sumner:
Thank you

PR #4218, opened 2023-08-19
Fix bundler inlining `string.length` incorrectly.

### LINKED ISSUE #4217: Length of a new line sequence differs in node and bun

0.7.3

Darwin 21.5.0 x86_64 i386

1. Create the following file as `newline-length.js`:
```javascript
console.log('line 1\nline 2'.length);
```
2. Run with node (outputs `13`)
3. Run with bun (outputs `14`)

Command line:
```shell
$ echo "console.log('line 1\nline 2'.length)" > newline-length.js
$ node newline-length.js
13
$ bun run newline-length.js
14
```

The length provided is not consistent with how strings are sliced as demonstrated here:
```javascript
console.log('line 1\n\n\nline 2'.slice('line 1\n\n\n'.length))
```

Node output: `line 2`
Bun output: `e 2`

```shell
$ echo "console.log('line 1\n\n\nline 2'.slice('line 1\n\n\n'.length))" > newline-slice.js
$ node newline-slice.js
line 2
$ bun run newline-slice.js
e 2
```

I would expect the length of a string containing `\n` to be the same as in node (`13`).

### What do you see instead?

It gives the length of `\n` as 2, rather than 1.

### Additional information

The library I'm using (https://github.com/futurGH/ts-to-jsdoc/blob/main/index.ts#L345) is slicing strings based on the length of another string. As the length of that string is different in node vs bun, it leads to further complications.
While I intend to fork and provide an alternative fix for this library, my expectation would be for node and bun to provide the same output for something as simple as the length of a string.

paperclover:
### What does this PR do?

Closes #4217

When the JS lexer sees `"\n"`, it marks this as utf8 / fast path, but the string `javascriptLenth` will incorrectly take the fast path and return the byte length instead of the escaped length.

During testing i noticed issues `dedent` was causing, so I rewrote that as a one off util.

- [X] Code changes

Added this case to the bundler tests


Jarred-Sumner:
This is not mergeable - note the failing tests.

Issue #4195, opened 2023-08-17
`--outfile` doesn't work with directory names


exoRift:
### What version of Bun is running?

0.7.4 (canary)

### What platform is your computer?

Linux 4.4.0-22621-Microsoft x86_64 x86_64

### What steps can reproduce the bug?

`bun build index.ts --compile --outfile foo/bar`

### What is the expected behavior?

Compilation of the code to a file called `bar` in the `foo` directory

### What do you see instead?

The directory is created, but then Bun throws a `FileNotFound` error.

Hanaasagi:
**Some debugging information**

From the error log:

```
error: failed to rename /home/kumiko/zig/tmp/test_4195/.f638a7bc19b07c22bun-build to foo/bar.js: FileNotFound
```

And the location where the exception is thrown in the code (Line 447)


https://github.com/oven-sh/bun/blob/360acf5a80d7e0191d13b87226c8122a633f701f/src/standalone_bun.zig#L437-L454


The value of variable `outfile` is `foo/bar.js`. `moveFileZWithHandle` function calls the `renameatZ` function, which results in the destination path being `foo/foo/bar.js`. (`to_dir` is the fd of `foo/` and `destination` is `foo/bar.js`)

https://github.com/oven-sh/bun/blob/360acf5a80d7e0191d13b87226c8122a633f701f/src/c.zig#L115-L126


I think this is the reason for this error.

PR #4226, opened 2023-08-20
Fix(bundler): allow generating exe file in nested path.

Hanaasagi:
### What does this PR do?

1. Fix `--outfile` doesn't work with nested path.  Close: #4195

- [x] Code changes

I wrote automated tests 

### DIFF
--- a/src/c.zig
+++ b/src/c.zig
@@ -113,10 +113,11 @@ pub fn moveFileZ(from_dir: std.os.fd_t, filename: [*:0]const u8, to_dir: std.os.
 }
 
 pub fn moveFileZWithHandle(from_handle: std.os.fd_t, from_dir: std.os.fd_t, filename: [*:0]const u8, to_dir: std.os.fd_t, destination: [*:0]const u8) !void {
+    _ = from_handle;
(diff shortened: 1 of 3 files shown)


src/c.zig
```diff
@@ -113,10 +113,11 @@ pub fn moveFileZ(from_dir: std.os.fd_t, filename: [*:0]const u8, to_dir: std.os.
 }
 
 pub fn moveFileZWithHandle(from_handle: std.os.fd_t, from_dir: std.os.fd_t, filename: [*:0]const u8, to_dir: std.os.fd_t, destination: [*:0]const u8) !void {
+    _ = from_handle;
```
Hanaasagi: In this scenario, `from_handle` is the file descriptor obtained from the `inject`, and it does not have read permissions.

https://github.com/oven-sh/bun/blob/3a9a6c63ac58564d8bae126a3edfc368ca4be671/src/standalone_bun.zig#L401

Should I directly modify the `inject` function to add read permissions?

Jarred-Sumner: Yes

Hanaasagi: Done.

paperclover (approved):

paperclover:
Thank you

Issue #4255, opened 2023-08-22
Ask users to use `bun --revision` instead of `bun -v` when submitting a bug in the issue template


vjpr:
https://github.com/oven-sh/bun/blob/52802a4c556d7498d75d00e4fe4e45633f4283e8/.github/ISSUE_TEMPLATE/2-bug-report.yml?plain=1#L17

xhyrom:
gonna do this

PR #4256, opened 2023-08-22
Ask for `bun --revision` instead `bun -v`

xhyrom:
### What does this PR do?
Fixes https://github.com/oven-sh/bun/issues/4255

Ask for `bun --revision` instead `bun -v` in bug issue template

- [ ] Code changes


xhyrom:
How about mentioning that you should use canary buns?

PR #4265, opened 2023-08-23
Fix ffi type

codehz:
### What does this PR do?

- [ ] Code changes

I used its type in https://github.com/codehz/bun_python/blob/%F0%9F%A5%9F/src/types.ts and verified it should works better than current types.

It's just two small changes:
1. Add Readonly so we can use `as const` to forced type narrow (when use with dedicated SYMBOLS table like that https://github.com/codehz/bun_python/blob/%F0%9F%A5%9F/src/symbols.ts#L277)
2. Split FFITypeToType to FFITypeToArgsType and FFITypeToReturnsType, it more precisely reflects the actual type than before.

### DIFF
--- a/packages/bun-types/ffi.d.ts
+++ b/packages/bun-types/ffi.d.ts
@@ -449,36 +449,7 @@ declare module "bun:ffi" {
 
   export type FFITypeOrString =
     | FFIType
-    | "char"
-    | "int8_t"
-    | "i8"
-    | "uint8_t"
-    | "u8"
-    | "int16_t"
-    | "i16"
-    | "uint16_t"
-    | "u16"
-    | "int32_t"
-    | "i32"
-    | "int"
-    | "uint32_t"
-    | "u32"
-    | "int64_t"
-    | "i64"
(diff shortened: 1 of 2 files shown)


paperclover:
Looks good overall, but the type tests are failing. Can you fix those / Add more.

Also, I haven't played around with these typedefs myself, but does using `Readonly` like you are still allow usage without `as const`. Can you make sure the example on https://bun.sh/docs/api/ffi properly typechecks.

codehz:
> but does using `Readonly` like you are still allow usage without `as const`.

Readonly doesn't prevent you to use mutable type, e.g. `const a: readonly number[] = []` pass the typecheck


> Can you fix those / Add more.

I will fix those soon

codehz:
@paperdave now it should working, and several tests are added

paperclover:
in ~30min im gonna manually git checkout and triple check

lgtm

paperclover (approved):

paperclover:
Thank you

PR #4266, opened 2023-08-23
fix(install): fix crash when installing package that uses loose semver pre-release

paperclover:
### What does this PR do?

These need tests. Fixes a segfault when doing `bun install grunt-contrib-yuidoc` without it cached.

I havent fully gone through how bun install tests work, and it's definitely a bit quirky to test because the first

- [X] Code changes

i verified it manually but i will write tests soon

### DIFF
--- a/src/install/semver.zig
+++ b/src/install/semver.zig
@@ -1151,6 +1151,37 @@ pub const Version = extern struct {
 
         return std.fmt.parseInt(u32, bytes[0..byte_i], 10) catch 0;
     }
+
+    /// Without doing a full semver parse, will this string parse to contain a pre-release tag?
+    /// Does not check for a valid semver, return value is only valid if the input is a valid semver.
+    ///
+    /// This is used by PackageManifest.parse() to classify strings, and determines how
+    /// much memory to allocate for package version entries in two lists.
+    ///
+    /// If all npm packages were strict semvers, this test would only do "contains('-')", but some
+    /// packages expect a version like "1.0.0rc7" which wouldn't work here. A mismatch between this
+    /// pre-prediction function and the actual semver parser will end up causing a segfault due to
+    /// incorrect memory allocation later on.
+    ///
+    /// Strings that contain a pre-release tag:
+    ///     "1.0.0-hello" (there is a dash)
+    ///     "1.0.0-0" (there is a dash)
+    ///     "3.7.5rc" (there is letters A-Z or a-z)
+    /// Strings that do not contain a pre-release tag:
+    ///     "1.0.0" (no dash or letters)
+    ///     "1.0.0+hello" (a plus is a post-release)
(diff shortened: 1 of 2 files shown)


src/install/semver.zig
```diff
@@ -1151,6 +1151,37 @@ pub const Version = extern struct {
...
+    ///     "1.0.0-0" (there is a dash)
+    ///     "3.7.5rc" (there is letters A-Z or a-z)
+    /// Strings that do not contain a pre-release tag:
+    ///     "1.0.0" (no dash or letters)
+    ///     "1.0.0+hello" (a plus is a post-release)
+    pub fn stringHasPrereleaseTag(input: []const u8) bool {
+        var i: usize = 0;
+        while (i < input.len) {
+            switch (input[i]) {
+                '-', 'A'...'Z', 'a'...'z' => return true,
```
Jarred-Sumner, line 1177: i think `_` is another one

paperclover: according to npm semver, it doesn't seem to be a valid character to start the pre-release tag.

Jarred-Sumner: yeah but what if packages do it

src/install/semver.zig
```diff
@@ -1151,6 +1151,37 @@ pub const Version = extern struct {
...
+    /// incorrect memory allocation later on.
+    ///
+    /// Strings that contain a pre-release tag:
+    ///     "1.0.0-hello" (there is a dash)
+    ///     "1.0.0-0" (there is a dash)
+    ///     "3.7.5rc" (there is letters A-Z or a-z)
+    /// Strings that do not contain a pre-release tag:
+    ///     "1.0.0" (no dash or letters)
+    ///     "1.0.0+hello" (a plus is a post-release)
+    pub fn stringHasPrereleaseTag(input: []const u8) bool {
```
alexlamsl, line 1173: I'm not sure if making up a separate parsing function is the best approach here, as:
- `Semver.Version.parse()` isn't that slow (and we are doing one anyway, see below)
- any further compatibility work will only cause `parsed_version.version.tag.hasPre()` to deviate from this
- does not address the check over here (note it requires handling build, not just pre): https://github.com/oven-sh/bun/blob/ad326b77342dd3d8585a30b7da803d32f9c11fe2/src/install/npm.zig#L1031-L1037

src/install/semver.zig
alexlamsl: to be fair we aren't handling `_` in `Semver.Version.parse()` either:
https://github.com/oven-sh/bun/blob/ad326b77342dd3d8585a30b7da803d32f9c11fe2/src/install/semver.zig#L1070-L1074

all the more reasons we should just extend and use `parse()` consistently instead of writing a new function though

PR #4281, opened 2023-08-24
Add Debugger docs and a couple guides

colinhacks:
### What does this PR do?

- [ ] Code changes


gtrabanco:
I know this is close. But perhaps this is useful to anyone.

To check if it is a runtime or which runtime it is I made a gist:
- https://gist.github.com/gtrabanco/38cb21f716253f75b9709521068498b1

PR #4282, opened 2023-08-24
🚰  drain the microtasks more

Jarred-Sumner:
### What does this PR do?

Tasks from uSockets need to also drain the microtask queue or they will be unnecessarily delayed and cost a bunch of memory.

Existing tests

### DIFF
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1392,6 +1392,8 @@ pub const Subprocess = struct {
             JSValue.zero;
         subprocess.this_jsvalue = out;
 
+        var send_exit_notification = false;
+
         if (comptime !is_sync) {
             var poll = JSC.FilePoll.init(jsc_vm, pidfd, .{}, Subprocess, subprocess);
             subprocess.poll_ref = poll;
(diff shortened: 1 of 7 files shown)


src/bun.js/api/bun/subprocess.zig
```diff
@@ -1406,13 +1408,20 @@ pub const Subprocess = struct {
                         @panic("This shouldn't happen");
                     }
 
-                    // process has already exited
-                    // https://cs.github.com/libuv/libuv/blob/b00d1bd225b602570baee82a6152eaa823a84fa6/src/unix/process.c#L1007
-                    subprocess.onExitNotification();
+                    send_exit_notification = true;
```
Jarred-Sumner, line 1411: @dylan-conway i'm wondering and hoping this partially addresses the reliability issues with Bun.spawn

src/bun.js/api/bun/subprocess.zig
```diff
@@ -1406,13 +1408,20 @@ pub const Subprocess = struct {
...
+                    lazy = false;
                 },
             }
         }
 
+        defer {
+            if (send_exit_notification) {
```
L3P3, line 1418: Why not doing this in the above if block?

PR #4293, opened 2023-08-24
docs: improved ESM and built-in modules section

huseeiin:
### What does this PR do?

- Saying "ESM Modules" doesn't make sense. It translates to "ECMAScript Modules Modules" changed it to "ESM" and "ECMAScript Modules"
- "Builtins JS" ???
- "built-in modules" has far more Google search results than "builtins"
- "Node.js modules" doesn't clarify that the doc is referring to Node.js *built-in* modules

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,13 +47,13 @@ TODO: document this (see [`bindings.zig`](src/bun.js/bindings/bindings.zig) and 
 
 Copy from examples like `Subprocess` or `Response`.
 
-### ESM Modules and Builtins JS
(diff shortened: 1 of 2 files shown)


paperclover:
I edited this section last, and I agree it's confusing, but the edits here aren't 100% accurate either, as none of the code in src/js is technically an ES Module. It's a mix of isolated functions and a custom module format.

Definitely a step in the right direction though. I think the entire section should be named `Bun/Node Built-in Modules` and I kind of want to tidy some of the explanation myself to be more technically correct (context: I did most of the tooling work for `src/js`).

Issue #4298, opened 2023-08-24
`http.server` crashes after exception in the dev build.


Hanaasagi:
### What version of Bun is running?

0.8.0_debug

### What platform is your computer?

Linux 6.4.11-arch2-1 x86_64 unknown

### What steps can reproduce the bug?

server.js

```JavaScript
import { createServer, request } from "node:http";

const server = createServer(function(req, res) {
  // NOTE: url is not imported !!!
  // we need a exception here to reproduce.
  let option = url.parse(req.url);
});

server.listen(1087);
console.log("Listening on port 1087");
```

Run

```
curl 127.0.0.1:1087
```

### What is the expected behavior?

No crash.

### What do you see instead?

Crash

### coredump stack

```
#0  0x00007f8e6568e83c in ?? () from /usr/lib/libc.so.6
[Current thread is 1 (Thread 0x7f8e65930400 (LWP 722865))]
(gdb) bt
#0  0x00007f8e6568e83c in ?? () from /usr/lib/libc.so.6
#1  0x00007f8e6563e668 in raise () from /usr/lib/libc.so.6
#2  0x00007f8e65626542 in abort () from /usr/lib/libc.so.6
#3  0x000055bbc411f9c9 in os.abort () at /home/kumiko/zig/zig-linux-x86_64-0.11.0-dev.4006+bf827d0b5/lib/std/os.zig:617
#4  0x000055bbc411b729 in debug.panicImpl (trace=0x0, first_trace_addr=..., msg=...)
    at /home/kumiko/zig/zig-linux-x86_64-0.11.0-dev.4006+bf827d0b5/lib/std/debug.zig:356
#5  0x000055bbc411a0cb in builtin.default_panic ()
    at /home/kumiko/zig/zig-linux-x86_64-0.11.0-dev.4006+bf827d0b5/lib/std/builtin.zig:814
#6  src.panic_handler.NewPanicHandler((function 'default_panic')).handle_panic () at /home/kumiko/zig/bun/src/panic_handler.zig:49
#7  src.main.panic (msg=..., error_return_trace=0x0, addr=...) at /home/kumiko/zig/bun/src/main.zig:9
#8  0x000055bbc411bce7 in debug.assert (ok=false)
    at /home/kumiko/zig/zig-linux-x86_64-0.11.0-dev.4006+bf827d0b5/lib/std/debug.zig:258
[...]

PR #4300, opened 2023-08-24
Make the server not crash if an error occurs in the dev build.

Hanaasagi:
### What does this PR do?

- Make the server not crash if an error occurs in the dev build.

Close: #4298 

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
I see. This started happening because we defer deinitializing.

PR #4302, opened 2023-08-24
[install] fix crash when installing package that uses loose semver pre-release

alexlamsl:
- also fix parsing of `1.2.3pre+build`

fixes #4266

### What does this PR do?

- [x] Code changes

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/install/npm.zig
+++ b/src/install/npm.zig
@@ -846,8 +846,16 @@ pub const PackageManifest = struct {
                 const versions = versions_q.expr.data.e_object.properties.slice();
                 for (versions) |prop| {
                     const version_name = prop.key.?.asString(allocator) orelse continue;
+                    const sliced_version = SlicedString.init(version_name, version_name);
+                    const parsed_version = Semver.Version.parse(sliced_version);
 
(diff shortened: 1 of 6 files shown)


src/install/npm.zig
```diff
@@ -846,8 +846,16 @@ pub const PackageManifest = struct {
                 const versions = versions_q.expr.data.e_object.properties.slice();
                 for (versions) |prop| {
                     const version_name = prop.key.?.asString(allocator) orelse continue;
+                    const sliced_version = SlicedString.init(version_name, version_name);
+                    const parsed_version = Semver.Version.parse(sliced_version, allocator);
```
Jarred-Sumner: This leaks some memory. We should probably use an arena in a few places here

alexlamsl: Actually, that `allocator` in `parse()` is unused historical artifact &minus; updated PR to trim it.

(OTOH, `Semver.Query.parse()` does require memory allocation &minus; however its return type comes with `.deinit()` so we are covered there)

Jarred-Sumner:
thanks

PR #4320, opened 2023-08-25
docs: fix http simple example log statement

karlbohlmark:
I just removed the incorrect escape of the port number in the simple http server example:

Screenshot 2023-08-25 at 11 34 48

Feel free to close and fix as part of another change if that is preferred.


xhyrom (approved):
👍

colinhacks:
Thanks!

PR #4331, opened 2023-08-25
make pending_response and metdata life cycle more clear and make fetch more reliable

cirospaciari:
### What does this PR do?

- [x] Code changes

There are tests


src/zlib.zig
```diff
@@ -429,7 +429,7 @@ pub const ZlibReaderArrayList = struct {
 
     pub fn alloc(ctx: *anyopaque, items: uInt, len: uInt) callconv(.C) *anyopaque {
         var this = @as(*ZlibReader, @ptrCast(@alignCast(ctx)));
-        const buf = this.allocator.alloc(u8, items * len) catch unreachable;
+        const buf = this.arena.allocator().alloc(u8, items * len) catch unreachable;
```
Jarred-Sumner: We shouldn’t need to use the arena here. Zlib shouldn’t leak memory. It likely means we are missing a function call. 

arenas are really bad for performance when used inside arrays because zig’s arena allocator only frees at the end. So we are keeping a duplicate copy of the entire array on every allocation

cirospaciari: changed to use mimalloc directly, but reporting back to HTTP client is still pending.

src/zlib.zig
```diff
@@ -909,7 +899,7 @@ pub const ZlibCompressorArrayList = struct {
             @sizeOf(zStream_struct),
         )) {
             ReturnCode.Ok => {
-                try zlib_reader.list.ensureTotalCapacityPrecise(allocator, deflateBound(&zlib_reader.zlib, input.len));
+                try zlib_reader.list.ensureTotalCapacityPrecise(list_allocator, deflateBound(&zlib_reader.zlib, input.len));
```
Jarred-Sumner, line 902: this is harmless right now because list_allocator and allocator are the same, but it could randomly break in the future

Issue #4299, opened 2023-08-24
dns.promises.reverse not implemented


aclelland:
### What version of Bun is running?

0.8

### What platform is your computer?

Linux 5.11.0-1022-aws x86_64 x86_64

### What steps can reproduce the bug?
```
let go = async()=>{
        let dns = require("node:dns");
        let hostnames = await dns.promises.reverse("1.1.1.1");
        console.log(hostnames);
};
go();
```
### What is the expected behavior?

Running in node.js the script will print
>[ 'one.one.one.one' ]

### What do you see instead?

>undefined

### Additional information

This issue seems a little different from some of the other DNS issues which were reported ( https://github.com/oven-sh/bun/issues/3433 , https://github.com/oven-sh/bun/issues/3981) since I don't get an exception, instead I just get `undefined` returned. 

Thanks,
Alan

Hanaasagi:
Yes, it's not implemented. It seems like we just need to wrap the `ares_gethostbyaddr`. Maybe I can work on it this weekend.

Ref: https://c-ares.org/ares_gethostbyaddr.html

PR #4332, opened 2023-08-25
feat(node:dns): implement `dns.reverse`.

Hanaasagi:
### What does this PR do?

Implement `dns.reverse`, Close: #4299 

See also: https://c-ares.org/ares_gethostbyname.html

- [x] Code changes

I wrote automated tests.


Jarred-Sumner (approved):
LGTM. thank you

PR #4336, opened 2023-08-25
fix(node:net): emit close event on connection error

### LINKED ISSUE #4316: `net.Socket` does not emit `error` if cannot connect to server

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

Darwin 22.1.0 arm64 arm

```ts

import net from 'node:net';

const getAvailablePort = (portToTry: number) =>
	new Promise<'available' | 'unavailable'>((resolve) => {
		let status: 'available' | 'unavailable' = 'unavailable';

		const host = '127.0.0.1';
		const socket = new net.Socket();

		socket.on('connect', () => {
			status = 'unavailable';
			socket.destroy();
		});

		socket.setTimeout(3000);
		socket.on('timeout', () => {
			status = 'unavailable';
			socket.destroy();
			resolve(status);
		});

		socket.on('error', () => {
			status = 'available';
		});

		socket.on('close', () => resolve(status));

		socket.connect(portToTry, host);
	});

console.log(await getAvailablePort(3000));
```

```
bun get-port.ts
```

Print `"available"`

### What do you see instead?

Hanging process, no output

### Additional information

In Node.js, this works. `error` event is emitted and therefore we know that this port is available to listen to.

Maybe a second bug: The "timeout" also never fires.

paperclover:
### What does this PR do?

Fixes #4316

We add a handlers object with connectError, but also the connect promise rejects with the same error, leading to two error events, so I took out the handler and only use the promise rejection (we'd have to call .catch regardless to prevent unhandled rejection).

We also need to call the close event on the socket.

- [ ] Code changes


Jarred-Sumner:
needs a test

paperclover:
there is a test now

Jarred-Sumner (approved):

PR #4340, opened 2023-08-26
feat(node:fs): add `cp`/`cpSync`/`promises.cp` + async `copyFile`

### LINKED ISSUE #1519: `fs.cp`, `fs.cpSync` and `fsPromises.cp` are missing

0.2.3

`Linux 5.14.0-70.26.1.el9_0.x86_64 #1 SMP PREEMPT Thu Sep 22 11:36:42 UTC 2022 x86_64 x86_64`

_No response_

_No response_

_No response_

### What do you see instead?

'cp' is undefined
'cpSync' is undefined

### Additional information

_No response_

### LINKED ISSUE #3778: Implement `cp` for `node:fs/promises`

I've tried completely replacing `node` with `bun` for a [fairly large project](https://github.com/cubing/cubing.js). There were a few incompatibilities that were easy to work around (as well as a general issue running Playwright), but the main thing that's not easy to fix within the project itself was a dependency using `cp` from `fs/promises`: https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options

The function is marked as experimental, but has been available without a flag for two years.

Implement it? 🤓

I can work around this, since I control the build dependency. But I thought I'd report that this was the main thing blocking us from a trivial switchover to `bun` for most of our project!

paperclover:
### What does this PR do?

- Implements `cpSync`/`cp` partially in native code. If you pass `filter` flag or another that isnt supported in the native impl, it falls back to the JS version which is a lazy loaded version of what node.js does.
  - On MacOS, this uses `clonefile`, achieving extremely high performance. 
- Implement async version of `copyFile`, this improves performance when doing many parallel copies, which makes the async fallback faster.
- Async `fs.cp()`

Closes #3778
Closes #1519

- [X] Code changes

Added tests.


src/js/node/fs.js
```diff
@@ -1065,10 +1071,59 @@ Object.defineProperties(fs, {
...
+  throw new TypeError(`Expected ${name} to be a number or Date`);
+}
+
 export default {
+  Dirent,
+  FSWatcher,
+  ReadStream,
+  Stats,
+  WriteStream,
+  _toUnixTimestamp,
```
Jarred-Sumner, line 1128: TIL

src/js/node/fs.promises.ts
```diff
@@ -89,11 +89,37 @@ function watch(
...
+  if (typeof options !== "object") {
+    throw new TypeError("options must be an object");
+  }
+  let mode = options.mode;
+  if (
+    options.dereference ||
+    options.filter ||
+    options.preserveTimestamps ||
+    options.verbatimSymlinks ||
+    !options.force ||
```
Jarred-Sumner: `force` or `errorOnExist` will usually not be set, which means the fast path will be the unusual path.

Maybe `options` should be combined to have the default values set to distinguish between when options was set to an empty object/missing some properties versus when the developer explicitly passed `force: false` or `errorOnExist: false`.

paperclover:
At this point:
- `.cpSync` and `.cp` work on MacOS. I need to test linux out still, I think it will fail the symlink tests.
- I do not fully understand the different atomic modes, so there may be a possible race condition.
- write more tests? It seems to be stable on mac.

There are `// TODO` comments that mark my own notes on the code, should be removed before merge.

src/js/node/fs.js
```diff
@@ -426,6 +435,7 @@ ReadStream = (function (InternalReadStream) {
         tempThis.path = tempThis.file = tempThis[readStreamPathOrFdSymbol] = pathOrFd;
       } else if (typeof pathOrFd === "number") {
         pathOrFd |= 0;
+        console.log("pathOrFd", pathOrFd);
```
Jarred-Sumner: extra comment

src/bun.js/node/node_fs.zig
```diff
@@ -419,6 +419,268 @@ pub const AsyncReadFileTask = struct {
...
+        }
+
+        this.deinit();
+    }
+
+    pub fn deinit(this: *AsyncCpSingleFileTask) void {
+        bun.default_allocator.destroy(this);
+
+        // There is only one path buffer for both paths. 2 extra bytes are the nulls at the end of each
+        bun.default_allocator.free(this.src.ptr[0 .. this.src.len + this.dest.len + 2]);
```
Jarred-Sumner, line 676: UAF

src/bun.js/node/node_fs.zig
```diff
@@ -419,6 +419,268 @@ pub const AsyncReadFileTask = struct {
...
+                    this.cp_task.finishConcurrently(result);
+                    this.deinit();
+                    return;
+                },
+                .result => {},
+            }
+        }
+
+        // TODO: the atomics are not right here
+        _ = @atomicRmw(usize, &this.cp_task.subtasks_completed, .Add, 1, .Monotonic);
```
Jarred-Sumner: one std.atomic.Atomic(usize) for remaining tasks which acts as a reference count would not be racy

birkskyum:
Related to
- https://github.com/oven-sh/bun/issues/3778
- https://github.com/oven-sh/bun/issues/1519

paperclover:
this took way too long to complete

Jarred-Sumner:
fs tests are failing on all 4 builds 

seems like a pattern. Can you fix that?

Jarred-Sumner:
Thank you

PR #4344, opened 2023-08-26
Automatically hot reload Bun.serve()

Jarred-Sumner:
### What does this PR do?

This makes it so the following hot reloads with zero configuration:

```js
Bun.serve({
  fetch(req) {
     return new Response("!!!");
  }
});
```

Previously, you had to use `export default { fetch(req) { ... }`  instead to support hot reloading.

This is only enabled when `bun --hot` itself is enabled. Even so, it can be disabled by setting `id` to `null` in `Bun.serve`

This also adds an `id` property to `Bun.serve()` which is initially just used for hot reloads, but in the future it could be used in stack traces or other logging.

Just manually for now

### DIFF
--- a/docs/runtime/hot.md
+++ b/docs/runtime/hot.md
@@ -102,54 +102,24 @@ Traditional file watchers like `nodemon` restart the entire process, so HTTP ser
 Bun provides the following simplified API for implementing HTTP servers. Refer to [API > HTTP](/docs/api/http) for full details.
 
 ```ts#server.ts
-import {type Serve} from "bun";
+import {serve} from "bun";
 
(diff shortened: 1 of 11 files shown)


samnoonabrar (commented):
Looks good

samnoonabrar (commented):
The coding convention seems fantastic! 

Regarding tests, 
The following tests failed:
6 files with test failures on bun-darwin-aarch64
1 files with test failures on linux-x64-baseline
7 files with test failures on bun-darwin-x64-baseline

And, the test failures on linux-x64 have been resolved

Overall, it is a good PR.

Issue #3951, opened 2023-08-03
bun test --watch should have an option to clear screen before tests run again


comunidadio:
### What is the problem this feature would solve?

Avoid cluttered/confusing showing old and current test(s) output together.

### What is the feature you are proposing to solve the problem?

--watch should clear the screen before restarting the tests, either by default or with a new --clear-screen option.

Fyi, Deno does clear the screen by default for "deno test" and provides a --no-clear-screen option to disable the behavior.

paperclover:
related to #3954 

i think `--no-clear-screen` is a good flag, for this and for --hot/--watch.

simylein:
also relates to #3491

technoplato:
Got a super hack here in the mean time until the awesome bun team adds this: https://github.com/oven-sh/bun/issues/3491#issuecomment-2149479491

gkiely:
Here's my current workaround

Basic clear:
```sh
# bunfig.toml
preload = ["./bun-setup.ts"]
```

```ts
// bun-setup.ts
import { $ } from 'bun';
await $`clear && printf '\033[3J'`;
```

---
More complicated clear (same formatting as bun test):

```sh
# test.sh
if [ -z "$1" ] ; then
  echo "No argument provided"
  exit 0
fi

# Matches against the start of the file name
FILES=$(find src -type f -name "*.test.ts" -o -name "*.test.tsx" | xargs -n 1 basename | grep -i "^$1")
FILES_WITH_PATH=$(find src -type f -name "*.test.ts" -o -name "*.test.tsx" | grep -i "$FILES")

# Check if any files returned
if [ -z "$FILES" ] ; then
  echo "No files found matching: $1"
  exit 0
fi

# Set process.env
export FILES=$FILES
export FILES_WITH_PATH=$FILES_WITH_PATH

# Run test watch
if [ "$WATCH" = true ] ; then
  if [ "$ONLY" = true ] ; then
    bun test --watch --only $FILES
  else
    bun test --watch $FILES
  fi
else
  if [ "$ONLY" = true ] ; then
    bun test --only $FILES
  else
    bun test $FILES
  fi
fi
```

package.json
```json
{
  "scripts": {
    "test-watch": "WATCH=true ./test.sh",
    "test-watch-only": "WATCH=true ONLY=true ./test.sh",
  }
}
```

```sh
# bunfig.toml
preload = ["./bun-setup.ts"]
```

```ts
// bun-setup.ts
const green = '\x1b[32m';
[...]

PR #4352, opened 2023-08-26
feat(test-clear-screen): clear when in watch mode

simylein:
### What does this PR do?

This will partially address #3951 by clearing the screen if `bun test` is in `--watch` mode.

- [x] Code changes

I ran it locally with `bun test` and `bun --watch test`

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -2735,7 +2735,7 @@ pub fn NewHotReloader(comptime Ctx: type, comptime EventLoopType: type, comptime
                     return;
 
                 if (comptime reload_immediately) {
-                    bun.reloadProcess(bun.default_allocator, Output.enable_ansi_colors);
+                    bun.reloadProcess(bun.default_allocator, true);
                     unreachable;
                 }
 
--- a/src/cli/test_command.zig
+++ b/src/cli/test_command.zig
@@ -861,6 +861,12 @@ pub const TestCommand = struct {
             files: []const PathString,
             allocator: std.mem.Allocator,
             pub fn begin(this: *@This()) void {
+                if (this.vm.hot_reload == .watch) {
+                    Output.flush();
+                    Output.disableBuffering();
+                    Output.resetTerminalAll();
+                    Output.enableBuffering();
+                }
                 var reporter = this.reporter;
                 var vm = this.vm;
                 var files = this.files;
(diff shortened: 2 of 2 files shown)


simylein:
I have two questions:

1) is this the right place to do the screen clearing?
2) do we need `Output.enableBuffering();` since we are in `--watch` mode and not `--hot` mode and the process will restart anyway?

src/cli/test_command.zig
```diff
@@ -861,6 +861,12 @@ pub const TestCommand = struct {
             files: []const PathString,
             allocator: std.mem.Allocator,
             pub fn begin(this: *@This()) void {
+                if (this.vm.hot_reload == .watch) {
```
Jarred-Sumner: This isn't the right place for it

Here is the right place:

https://github.com/oven-sh/bun/blob/efe987e8d12e824dde840b56cbb704feabe26ed1/src/bun.zig#L1292-L1296

We should pass `true` to `clear_terminal`.

src/bun.js/javascript.zig
```diff
@@ -2735,7 +2735,7 @@ pub fn NewHotReloader(comptime Ctx: type, comptime EventLoopType: type, comptime
                     return;
 
                 if (comptime reload_immediately) {
-                    bun.reloadProcess(bun.default_allocator, Output.enable_ansi_colors);
+                    bun.reloadProcess(bun.default_allocator, true);
```
Jarred-Sumner, line 2738: The problem is what if the terminal doesn’t support ansi colors? Instead of clearing the screen, it will write some strange-looking characters. 

I think the issue is that Output.enable_ansi_colors is returning false for something it shouldn’t be returning false. What’s the case where the terminal isn’t cleared with —watch?

simylein: `bun --watch test`, see #3951

Jarred-Sumner:
its definitely clearing the screen on my machine

that means the issue is not with the code to clear the screen

Screen Recording 2023-08-28 at 5 24 26 AM


the issue is with `Output.enable_ansi_colors` 

must be set to false on your machine for some reason. A bug in here somewhere:

https://github.com/oven-sh/bun/blob/ebfaa682f761aac3d0c3c6d6ec1bbf4831d86ea4/src/output.zig#L136-L154

simylein:
You are correct, it starts working as soon as I set the env var `FORCE_COLOR`.

Shouldn't it be

```zig
 if (isForceColor() or isColorTerminal()) { 
     enable_color = true; 
 } else if (isNoColor() or !isColorTerminal()) { 
     enable_color = false; 
 } 
```

instead of

```zig
 if (isForceColor()) { 
     enable_color = true; 
 } else if (isNoColor() or !isColorTerminal()) { 
     enable_color = false; 
 } 
```

or am I missing something here?

PR #4354, opened 2023-08-26
just kernel32 things

Jarred-Sumner:
### What does this PR do?

batch of changes that should have no impact 

it doesn't. just slightly gets us closer, hopefully without breaking anything


src/bun.js/node/syscall.zig
```diff
@@ -15,13 +15,14 @@ const fd_t = bun.FileDescriptor;
 const C = @import("root").bun.C;
 const linux = os.linux;
 const Maybe = JSC.Maybe;
+const kernel32 = windows.kernel32;
```
PodaruDragos: it's happening !!!! 👯

xhyrom:
:eyes: :tada:

src/io/io_windows.zig
```diff
@@ -0,0 +1,1268 @@
...
+        // and uses the NTFS/Windows epoch of 1601-01-01 instead of Unix Epoch 1970-01-01.
+        const epoch_adjust = std.time.epoch.windows * (std.time.ns_per_s / 100);
+        return (@as(i64, @bitCast(ft64)) + epoch_adjust) * 100;
+    }
+
+    pub fn tick(_: *Self) void {}
+};
+pub const system = os.windows;
+
+pub const Waker = struct {
```
kprotty, line 96: If this lives unrelated to the IO instance, you could use a Mutex+Condition.

If it's mean to live with an IO, it could reserve a completion_key from it and reuse the IO's iocp handle for posts.

src/io/io_windows.zig
```diff
@@ -0,0 +1,1268 @@
...
+        var attributes: os.windows.DWORD = 0;
+        // attributes |= os.windows.FILE_FLAG_NO_BUFFERING;
+        // attributes |= os.windows.FILE_FLAG_WRITE_THROUGH;
+
+        // This is critical as we rely on O_DSYNC for fsync() whenever we write to the file:
+        // assert((attributes & os.windows.FILE_FLAG_WRITE_THROUGH) > 0);
+
+        // TODO: Add ReadFileEx/WriteFileEx support.
+        // Not currently needed for O_DIRECT disk IO.
+        attributes |= os.windows.FILE_FLAG_OVERLAPPED;
```
kprotty: see above. This version uses pread/pwrite and doesnt register to IOCP for files.

src/io/io_windows.zig
Jarred-Sumner: thanks for the pointer, i'll do the latter. this is meant to allow one thread to wakeup another thread that is otherwise waiting on I/O. on macOS, we use a machport and on linux, we use an eventfd. 

example: https://github.com/oven-sh/bun/blob/main/src/io/io_darwin.zig#L507-L578

src/io/io_windows.zig
Jarred-Sumner: i'm going to change this part a lot. need to also incorporate usockets. since windows has IOCP, I think it can do almost everything on the main thread

Issue #4358, opened 2023-08-26
Invalid address from http.Server


sirenkovladd:
### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Linux 6.2.0-26-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
const server = require('http').createServer();
server.listen(3000, 'localhost', () => {
  console.log(JSON.stringify(server.address()))
  process.exit(0);
});
```

### What is the expected behavior?

```sh
❯ node bug.js 
{"address":"127.0.0.1","family":"IPv4","port":3000}
```

### What do you see instead?

```sh
❯ bun bug.js
{"address":"localhost","family":"IPv4","port":3000}
```

### Additional information

(http.Server instanceof net.Server)

https://github.com/nodejs/node/blob/v20.5.1/lib/net.js#L2004

https://github.com/nodejs/node/blob/v20.5.1/lib/net.js#L2062-L2073

sirenkovladd:
@Jarred-Sumner, is it possible to apply this PR because it blocks the use of `fastify-cli`?

```
330 |       family: isIPv6(address) ? "IPv6" : "IPv4",
331 |       port: this.#server.port
332 |     };
333 |   }
334 | 
335 |   listen(port, host, backlog, onListen) {
                   ^
error: Failed to start server. Is port 3000 in use?
      at listen (node:http:335:16)
      at <cut>/node_modules/fastify/lib/server.js:234:6
      at <cut>/node_modules/fastify/lib/server.js:189:10
      at node:dns:11:70
```

gabriiels:
https://github.com/oven-sh/bun/issues/1641#issuecomment-1708651616

sirenkovladd:
So it's a workaround

s1nistr4:
Yeah, fastify doesn't work with bun yet.

7f8ddd:
> Yeah, fastify doesn't work with bun yet.

Bun is working with Fastify on the canary release, including with TypeORM. Only issues I've had is related to NAPI libraries, such as node-canvas/canvas. This is with NestJS.
```
bun upgrade --canary
// TS:
bun add -d bun-types@canary
```

Jarred-Sumner:
This bug was fixed in Bun 1.0

```js
/tmp/checky
❯ bun such-work.js
{"address":"::1","family":"IPv6","port":3000}

/tmp/checky
❯ node such-work.js
[...]

PR #4359, opened 2023-08-26
lookup in http.listen

sirenkovladd:
### What does this PR do?

Fixes: https://github.com/oven-sh/bun/issues/4358
This adds dns.lookup when calling http.Server.listen

(http.Server instanceof net.Server)
https://github.com/nodejs/node/blob/v20.5.1/lib/net.js#L2062-L2073

- [X] Code changes


paperclover (approved):
looks good. i wish this dns lookup wasnt needed for every node:http server, but it seems this is the behavior node.js does

test/js/node/http/node-http.test.ts
```diff
@@ -823,4 +826,41 @@ describe("node:http", () => {
...
+      if (err) {
+        return closeAndFail("dns lookup failed");
+      }
+
+      server.listen(
+        0,
+        host,
+        mustCall(() => {
+          const address = server.address() as AddressInfo;
+          expect(address.address).toStrictEqual(ip);
```
Jarred-Sumner, line 920: this test is failing

Jarred-Sumner (changes requested):
lets merge once test failures are fixed

sirenkovladd:
@Jarred-Sumner @paperdave 
This binary contains no changes to http.ts [bun-linux-x64](https://github.com/oven-sh/bun/actions/runs/5986240007#:~:text=Artifacts,-Produced%20during%20runtime)
I have change in `src/js/out/InternalModuleRegistryConstants.h` but this is the out file
I really would not like to push changes from this file, as it will be impossible to review

paperclover:
the internal module registry constants is a generated file based on bundling. ci currently isnt able to generate this file, so we commit them locally. that's probably the reason for the test failure

sirenkovladd:
Ok, should I push this file? Are you planning to fix this?

paperclover:
yes, can you push that generated file?

sirenkovladd:
@Jarred-Sumner can you retest?

test/js/node/http/node-http.test.ts
sirenkovladd: https://github.com/oven-sh/bun/actions/runs/6002838901/job/16325640437#step:5:6184

sirenkovladd:
@Jarred-Sumner

sirenkovladd:
@Jarred-Sumner @paperdave

PR #4394, opened 2023-08-29
fix(node-fetch): use stream.Readable instead of web streams

paperclover:
### What does this PR do?

- [X] Code changes

New tests


src/js/thirdparty/node-fetch.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (body instanceof Readable) {
+      for await (const chunk of body) {
+        chunks.push(chunk);
+      }
+      init = { ...init, body: new Blob(chunks) };
+    }
+  }
+
+  const response = await nativeFetch(url, init);
+  Object.setPrototypeOf(response, Response.prototype);
```
paperclover, line 39: how slow is this prototype switch? should this be a subclass that implements all the getters or something else?

A good thing about this is it makes the wrapper very minimal in terms of # of js code.

Jarred-Sumner: pretty bad but not as bad as the network call

src/js/node/stream.js
```diff
@@ -3327,7 +3327,7 @@ var require_readable = __commonJS({
 
         streamReadable.pause();
 
-        const cleanup = finished(streamReadable, error => {
+        const cleanup = eos(streamReadable, error => {
```
Jarred-Sumner, line 3330: How intentional is this on a scale of 1 to 10

Jarred-Sumner (approved):

src/js/node/stream.js
paperclover: i looked at the place we copied this from
```
import { eos as finished } from ...
```

(there is no variable named `finished` in this scope)

so 10

paperclover:
rebased, letting tests run.

PR #4395, opened 2023-08-29
feat(compat): support `got`

### LINKED ISSUE #3150: the Got library not work with Bun

0.6.5

Darwin 22.1.0 arm64 arm

import got from 'got';

let response = await got('https://bun.sh/');
console.log(response.statusCode);

200

### What do you see instead?

 7 | const ClientRequest = require('./client-request.js');
 8 | const IncomingMessage = require('./incoming-message.js');
 9 | const auto = require('./auto.js');
10 | const {
11 |    HttpOverHttp2,
12 |    HttpsOverHttp2
    ^
TypeError: undefined is not an object (evaluating 'new tls.TLSSocket(new stream.PassThrough)._handle._parentWrap') 

### Additional information

_No response_

paperclover:
### What does this PR do?

I got `got` working.

Closes #3150

- Allow setting `req.statusMessage` (it is a no-op, which is fine for supporting got, as they do `= req.statusMessage ?? http.STATUS_CODES[req.status]` which is already what we set it to.
- Fix `newListeners` in native EE to emit before the listener is added.

- [X] Code changes

Tests for both individual fixes.


Electroid:
@paperdave We should add a basic test(s) that include `got` so we know if there's a regression in the future

sirenkovladd:
I think, also need tests with https

sirenkovladd:
```ts
test("https gzip", async () => {
    const stream = await got('https://bun.sh/', {
      headers: {
        'Accept-Encoding': 'gzip',
      }
    });
    expect(stream.statusCode).toBe(200);
  });
```

paperclover:
the https test does not work and blocks this. looking into it

src/http_client_async.zig
```diff
@@ -2621,6 +2621,8 @@ pub fn onData(this: *HTTPClient, comptime is_ssl: bool, incoming_data: []const u
             if (this.state.content_encoding_i < response.headers.len and !this.state.did_set_content_encoding) {
                 // if it compressed with this header, it is no longer because we will decompress it
                 var mutable_headers = std.ArrayListUnmanaged(picohttp.Header){ .items = response.headers, .capacity = response.headers.len };
+                // we remove the content encoding header
+                _ = mutable_headers.swapRemove(this.state.content_encoding_i);
```
Jarred-Sumner: This will change the order that the headers are passed to the server, which is technically observable

src/http_client_async.zig
```diff
@@ -2621,6 +2621,8 @@ pub fn onData(this: *HTTPClient, comptime is_ssl: bool, incoming_data: []const u
             if (this.state.content_encoding_i < response.headers.len and !this.state.did_set_content_encoding) {
                 // if it compressed with this header, it is no longer because we will decompress it
                 var mutable_headers = std.ArrayListUnmanaged(picohttp.Header){ .items = response.headers, .capacity = response.headers.len };
+                // we remove the content encoding header
+                _ = mutable_headers.orderedRemove(this.state.content_encoding_i);
```
liz3, line 2625: we can't remove the header here since it might be used by the user, see: https://github.com/oven-sh/bun/pull/5743

cirospaciari: If we keep the header we should not decode the body, because the user or libraries will see the header and try to decompress (like got) if the user wants the raw output (compressed) the user can use `decompress: false` options

f7zorn:
Any update on this? I know it's a bad question but got alone has 8.000 dependents.

prabhu:
Latest error with got 14.3.0

```
767 |             const cacheRequest = cacheableStore.get(options.cache)(options, async (response) => {
768 |                 response._readableState.autoDestroy = false;
769 |                 if (request) {
770 |                     const fix = () => {
771 |                         if (response.req) {
772 |                             response.complete = response.req.res.complete;
                                                      ^
TypeError: undefined is not an object (evaluating 'response.req.res.complete')
      at fix (/home/runner/work/cdxgen/cdxgen/node_modules/got/dist/source/core/index.js:772:49)
      at /home/runner/work/cdxgen/cdxgen/node_modules/got/dist/source/core/index.js:776:21
      at /home/runner/work/cdxgen/cdxgen/node_modules/got/dist/source/core/index.js:767:84
      at /home/runner/work/cdxgen/cdxgen/node_modules/cacheable-request/dist/index.js:148:25
      at handler (/home/runner/work/cdxgen/cdxgen/node_modules/cacheable-request/dist/index.js:77:40)
      at node:http:930:191
```

panga:
Are there any update on this?

PR #4409, opened 2023-08-30
Align `process.nextTick` execution order with Node

### LINKED ISSUE #4252: Unpredictable behavior using fastify server encapsulation

0.7.3

Darwin 22.5.0 arm64 arm

Unpredictable behavior using `fastify` sever with multiple levels of context encapsulation. I the below example every call to `await server.register()` method creates a copy of the `server` and passes it down to the plugin, thus creating an encapsulated instance per `plugin`. The `prefix` option allows us to assign a URL prefix for the plugin routes.

```typescript
import fastify, { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';

await start();

async function start() {
  const server = fastify();

  await server.register(plugin, { prefix: '/en' });
  await server.register(plugin, { prefix: '/ja' });
  await server.register(plugin, { prefix: '/fr' });
  
  console.log(server.printRoutes());
}

async function plugin(server: FastifyInstance) {
  server.get('/', async function handler(request: FastifyRequest, reply: FastifyReply) {
    return reply.send({ ok: true })
  });
}
```

Only if I add a `wait(1000)` somewhere between the plugin registration methods, the output of `console.log(server.printRoutes());` is as expected.

```ts
...
async function start() {
  const server = fastify();

  await server.register(plugin, { prefix: '/en' });
  await server.register(plugin, { prefix: '/ja' });

  // added some wait
  await wait(1000);
  await server.register(plugin, { prefix: '/fr' });
  
  console.log(server.printRoutes());
}
...

// added wait util
async function wait(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}
```

The expected output of the `server.printRoutes()` should be

```sh
bun-fastify $ bun index.ts
└── /
    ├── en (GET, HEAD)
    │   └── / (GET, HEAD)
    ├── ja (GET, HEAD)
    │   └── / (GET, HEAD)
    └── fr (GET, HEAD)
        └── / (GET, HEAD)
```

### What do you see instead?

The actual output, and plugin prefix levels, is somewhat unexpected, erroneous. The last plugin with prefix `/fr` was passed not the top level server instance, but the one that was passed to the plugin with the prefix `/ja`. This also means that plugin `/fr` has become a child of plugin `/ja`.

```sh
bun-fastify $ bun index.ts
└── /
    ├── en (GET, HEAD)
    │   └── / (GET, HEAD)
    └── ja (GET, HEAD)
        └── / (GET, HEAD)
            └── fr (GET, HEAD)
                └── / (GET, HEAD)
```

### Additional information

It is clear that, if we add an explicit delay with `wait` the behavior is as expected. Might this be an issue of object cloning via `Object.create(original)` or an `async/await` problem? 

Related `fastify` source code can be found somewhere [here](https://github.com/fastify/fastify/blob/main/lib/pluginOverride.js#L37) where instance cloning is happening.

Jarred-Sumner:
### What does this PR do?

In Node, `process.nextTick` runs before microtasks are drained. Previously, Bun was treating `process.nextTick` as the same queue as the microtask queue. This led to subtle bugs.

Fixes #4252, probably others

The code is loosely based on https://github.com/nodejs/node/blob/main/lib/internal/process/task_queues.js. The main difference is that our implementation is lazier and we don't emit on unhandledexception callbacks currently.

A couple tests. Have to avoid referencing the process object inside outer function to avoid causing it to infinite loop. Using the `$createFIFO` builtin creates too large of an array in stress tests, causing an infinite loop. That's why it uses this fixed buffer abstraction from node. 

I think we will need to make one small tweak to support AsyncLocalStorage, but @paperdave will know more about that than me


src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -479,6 +480,15 @@ extern "C" JSC__JSGlobalObject* Zig__GlobalObject__create(void* console_client,
 
     JSC::gcProtect(globalObject);
 
+    vm.setOnEachMicrotaskTick([globalObject](JSC::VM& vm) -> void {
+        if (auto nextTickQueue = globalObject->m_nextTickQueue.get()) {
+            vm.setOnEachMicrotaskTick(nullptr);
```
paperclover: why is the handler cleared here? it doesn't seem to ever be restored.

also we now need to coordinate this with the async_hooks `setOnEachMicrotaskTick` overwriting this handler temporarily.

I will push some tests that hit these edge cases

paperclover (changes requested):
I've been checking with node and this implementation is incorrect, `nextTick` runs after `queueMicrotask` as of node 20. Hopefully switching this around isn't too hard, but I don't think `onEachMicrotaskTick` will do the trick.

```ts
let resolve;
let promise = new Promise(res => {
  resolve = res;
});

const order = [];
let nextTickI = 0;
let microtaskI = 0;
let remaining = 20;
let runs = [];
  queueMicrotask(() => {
    runs.push(queueMicrotask);
    order.push("queueMicrotask " + microtaskI++);
    if (--remaining === 0) resolve(order);
  });
  process.nextTick(() => {
    runs.push(process.nextTick);
    order.push("process.nextTick " + nextTickI++);
    if (--remaining === 0) resolve(order);
  });
}

await promise;

console.log(order);
```
yields:

image

I'm going to update the tests to reflect this, as well as use jest as a baseline to ensure that they pass on node too.

paperclover:
Actually, I'm incorrect. Having this not be at the top level in node causes the order to be as you implemented it. going off my prev script,

```diff
 let resolve;
 let promise = new Promise(res => {
   resolve = res;
 });

 const order = [];
 let nextTickI = 0;
 let microtaskI = 0;
 let remaining = 20;
 let runs = [];
 for (let i = 0; i < 10; i++) {
+  process.nextTick(() => {
     queueMicrotask(() => {
       runs.push(queueMicrotask);
       order.push("queueMicrotask " + microtaskI++);
       if (--remaining === 0) resolve(order);
     });
     process.nextTick(() => {
       runs.push(process.nextTick);
       order.push("process.nextTick " + nextTickI++);
       if (--remaining === 0) resolve(order);
     });
+  });
 }

 await promise;

console.log(order);
```

image

I'm sure you checked this page: https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick#phases-overview

paperclover (approved):
merge if tests pass

Jarred-Sumner:
Tests do not pass, we need to address the exit code issue and possibly this readline promises test failure

Jarred-Sumner:
Let’s merge after tests run (you’ll need to fix the conflicts)

PR #4410, opened 2023-08-30
feat: Windows + CMake Build System

Jarred-Sumner:
### What does this PR do?

don't have time before bun 1.0 to learn enough to get good at windows I/O 

so we will use libuv on Windows

this is the prep code, does nothing but hopefully it's mergeable and then we can implement `FilePoll`, `PollRef` and then various functions.

it doesn't work


src/bun.js/bindings/ActiveDOMCallback.cpp
```diff
@@ -28,7 +28,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "config.h"
+#include "webcore/config.h"
```
dylan-conway, line 31: this can be reverted?

src/bun.js/bindings/ModuleLoader.cpp
```diff
@@ -80,14 +77,14 @@ generateInternalModuleSourceCode(JSC::JSGlobalObject* globalObject, InternalModu
         if (!object) {
             return;
         }
-        RETURN_IF_EXCEPTION(throwScope, {});
+        RETURN_IF_EXCEPTION(throwScope, void());
```
dylan-conway, line 81: {} doesn't coerce to jsUndefined anymore?

Jarred-Sumner: the function needs to return void

src/bun.js/bindings/headers-handwritten.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#ifndef HEADERS_HANDWRITTEN
```
dylan-conway:
```suggestion
#define HEADERS_HANDWRITTEN
```

src/bun.js/bindings/musl-memmem.h
```diff
@@ -0,0 +1 @@
+extern "C" void *memmem(const void *h0, size_t k, const void *n0, size_t l);
```
dylan-conway: we might need header guards here

Jarred-Sumner:
cc @bnoordhuis, Bun will soon be using libuv on Windows

src/async/windows_event_loop.zig
```diff
@@ -0,0 +1,340 @@
...
+        this.handle.unref();
+    }
+
+    /// From another thread, Prevent a poll from keeping the process alive.
+    pub fn unrefConcurrently(this: *KeepAlive, vm: *JSC.VirtualMachine) void {
+        _ = vm;
+        if (this.status != .active)
+            return;
+        this.status = .inactive;
+        this.handle.unref();
```
bnoordhuis, line 79: Is this called from a different thread than the one that calls uv_run()? If so, that's not safe, you can only call uv_async_send() from another thread, not uv_ref() or uv_unref().

Jarred-Sumner: thank you

LeonardoZivieri:
I've created an fast installation using wsl, running the command above in powershell. And you can see it in #5751 pull request

```powershell
$BunInstallFolder = "$env:ProgramData/bun" # An installation folder. You can change it
mkdir $BunInstallFolder
cd $BunInstallFolder
Write-Output "@ECHO OFF`nwsl bash -ic `"bun %*`"" | out-file "bun.bat" -Encoding ASCII
$env:Path += ";$pwd"
wsl bash -ic "curl -fsSL https://bun.sh/install | bash"
```

This script simple create a folder, add a script that call bun inside wsl, add the folder to path variable to run in future, and install bun inside wsl

Seblor:
Sorry for the out of topic, but I must say that last commit cracked me up.

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2032,7 +2037,7 @@ pub const DNSResolver = struct {
                     .code = bun.String.static(err.code()),
                     .message = bun.String.static(err.label()),
                 };
-
+                defer ip_slice.deinit();
```
Jarred-Sumner, line 2040: TODO: this is sus

src/cli.zig
```diff
@@ -145,7 +145,7 @@ pub const Arguments = struct {
         var paths = [_]string{ cwd, filename };
         const outpath = try std.fs.path.resolve(allocator, &paths);
         defer allocator.free(outpath);
-        var file = try std.fs.openFileAbsolute(outpath, std.fs.File.OpenFlags{ .mode = .read_only });
+        var file = try bun.openFileZ(&try std.os.toPosixPath(outpath), std.fs.File.OpenFlags{ .mode = .read_only });
```
Jarred-Sumner, line 148: TODO: this shouldn't be posix

src/install/install.zig
```diff
@@ -1575,10 +1652,10 @@ const PackageInstall = struct {
                     return result;
                 } else |err| {
                     switch (err) {
-                        error.NotSameFileSystem => {
-                            supported_method = .copyfile;
-                            supported_method_to_use = .copyfile;
-                        },
+                        // error.ENXIO => {
```
Jarred-Sumner, line 1655: TODO:!

src/install/install.zig
```diff
@@ -1636,8 +1713,11 @@ const TaskCallbackContext = union(Tag) {
 
 const TaskCallbackList = std.ArrayListUnmanaged(TaskCallbackContext);
 const TaskDependencyQueue = std.HashMapUnmanaged(u64, TaskCallbackList, IdentityContext(u64), 80);
-const TaskChannel = sync.Channel(Task, .{ .Static = 4096 });
-const NetworkChannel = sync.Channel(*NetworkTask, .{ .Static = 8192 });
+
+// Windows seems to stack overflow in debug builds due to the size of these allocations.
+const TaskChannel = sync.Channel(Task, .{ .Static = 4096 / (if (Environment.isWindows) 16 else 1) });
```
Jarred-Sumner, line 1718: todo: remove this diff

src/install/install.zig
```diff
@@ -1783,13 +1863,14 @@ pub const PackageManager = struct {
...
         _ = this.wait_count.fetchAdd(1, .Monotonic);
-        this.waiter.wake() catch {};
+        this.waiter.wake();
     }
 
     pub fn sleep(this: *PackageManager) void {
         if (this.wait_count.swap(0, .Monotonic) > 0) return;
         Output.flush();
-        _ = this.waiter.wait() catch 0;
+        bun.Mimalloc.mi_collect(false);
```
Jarred-Sumner: collect is bad

Issue #4398, opened 2023-08-30
Used Caret Ranges, but installed the wrong version


xihale:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Linux 6.1.44-1-MANJARO x86_64 unknown

### What steps can reproduce the bug?

First, init a new project, and add it to the `package.json`  
```json
{
  "Dependencies": {
    "hexo": "^6.3.0",
  }
}
```
Then run `bun i`  
Then run `head -n 4 node_modules/hexo/package.json`, as you can see, the version is "7.0.0-rc2"(or higher)

### What is the expected behavior?

reference the https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004, It should manager version likes this, otherwise there will be breaking errors in dev.

### What do you see instead?

When I use this `package.json`
```json
{
  "Dependencies": {
    "hexo": "^6.3.0",
  }
}
```
It installed the breaking changed version "7"
```json
"version": "7.0.0-rc2",
```

Hanaasagi:
I think this is related to comparing two semver versions. Semantically, `7.0.0-rc2` is considered less than `7.0.0`, but npm has its own additional rules https://github.com/npm/node-semver#prerelease-tags. 

> For example, the range >1.2.3-alpha.3 would be allowed to match the version 1.2.3-alpha.7, but it would not be satisfied by 3.4.5-alpha.9, even though 3.4.5-alpha.9 is technically "greater than" 1.2.3-alpha.3 according to the SemVer sort rules. The version range only accepts prerelease tags on the 1.2.3 version. The version 3.4.5 would satisfy the range, because it does not have a prerelease flag, and 3.4.5 is greater than 1.2.3-alpha.7.



![2023-08-30_19-09](https://github.com/oven-sh/bun/assets/9482395/36c5fea9-9173-4e44-973b-543ad1aa5a84)

xihale:
> I think this is related to comparing two semver versions. Semantically, `7.0.0-rc2` is considered less than `7.0.0`, but npm has its own additional rules https://github.com/npm/node-semver#prerelease-tags.
> 
[...]

PR #4412, opened 2023-08-30
Fix(semver): fix matching with pre-release version

Hanaasagi:
### What does this PR do?

Before this, we believed that the version range `[6.3.0, 7.0.0]` included version `7.0.0-rc-2`. However, in npm, if a version with a pre-release tag is not explicitly specified in `package.json`, it is not allowed to match pre-release versions. This pr adds additional restrictions, close close https://github.com/oven-sh/bun/issues/4398

Ref: https://github.com/npm/node-semver#prerelease-tags

I wrote automated tests

### DIFF
--- a/src/install/semver.zig
+++ b/src/install/semver.zig
@@ -1266,7 +1266,7 @@ pub const Range = struct {
             return lhs.op == rhs.op and lhs.version.eql(rhs.version);
         }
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (approved):
LGTM, thank you

Issue #4408, opened 2023-08-30
HTMLRewriter Issue in Bun >=0.8.0


gtrabanco:
### What version of Bun is running?

`0.8.2+f24ca3900400bb41faefb42f426651d5d7e7c2e6`

### What platform is your computer?

`Darwin 21.6.0 x86_64 i386`

### What steps can reproduce the bug?

Execute any valid HTMLRewriter:

```ts
const url = 'https://midu.dev';
const res = await fetch(url);
const titles: string[] = [];

const rw = new HTMLRewriter();
rw.on('h3', {
  element: (element: HTMLRewriterTypes.Element) => {
    // console.log('Inside element');
  },
  text: ({ text }: HTMLRewriterTypes.Text) => {
    const newsTitle = text.trim();
    if (newsTitle.length > 0) {
      titles.push(newsTitle);
    }
  }
})

rw.transform(res);

console.log(titles)
```

### What is the expected behavior?

The expected behaviour is to output:

```shell
[ "Últimos artículos", "CURSOS de Programación Full Stack gratuitos",
  "Artículos por temas", "Libros", "Último vídeo en YouTube",
  "Último Podcast", "¡Suscríbete a la newsletter!"
]
```

### What do you see instead?

No output because it not process the HTML.

### Additional information / Workaround

There is a workaround that makes this works, you should replace the line with the transform for this one:

```ts
rw.transform(new Response(await res.text())); // This works
```

With that, you will see the expected output.

The code I am using in production and same code that is failing with Bun >= `0.8.x` works as expected with Bun `0.7.3`.

Jarred-Sumner:
Thanks for the reproduction. Confirming that I'm able to reproduce this issue when running the script above enough times. It's likely a bug with fetch() body streaming (which was added in v0.8)

<img width="552" alt="image" src="https://github.com/oven-sh/bun/assets/709451/d39534b2-2211-495e-a9ef-e55551fada04">

gtrabanco:
This is still failing with revision `0.8.2+f73f77d0dec9006f5f42ff401fb32f9eaefa749a`. Not sure if it should be resolved but seems like it should due the #4418 was applied in commit `2d80f94edafe09329b027424b32908632694553d` which is behind.

@cirospaciari

gtrabanco:
I had to do a this workaround to rewrite a response with `HTMLRewriter` which means that this is happening also without doing `fetch`:
[...]

PR #4418, opened 2023-08-30
fix(HTMLRewriter) buffer response before transform

cirospaciari:
### What does this PR do?
Fixes https://github.com/oven-sh/bun/issues/4408

- [x] Code changes

I wrote automated tests


src/bun.js/api/html_rewriter.zig
```diff
@@ -409,59 +407,75 @@ pub const HTMLRewriter = struct {
...
+                } else if (sink.response.body.value == .Locked and @intFromPtr(sink.response.body.value.Locked.task) == @intFromPtr(sink) and
+                    sink.response.body.value.Locked.promise != null)
+                {
+                    sink.response.body.value.Locked.onReceiveValue = null;
+                    sink.response.body.value.Locked.task = null;
+                }
+                if (is_async) {
+                    sink.response.body.value.toErrorInstance(err, sink.global);
+                } else {
+                    sink.tmp_sync_error = throwLOLHTMLError(sink.global);
```
Jarred-Sumner: this will segfault 

we cannot hold this value in the heap without the GC visiting it

src/bun.js/webcore/body.zig
```diff
@@ -1258,3 +1263,322 @@ pub fn BodyMixin(comptime Type: type) type {
...
+    const log = bun.Output.scoped(.BodyValueBufferer, false);
+
+    const ArrayBufferSink = JSC.WebCore.ArrayBufferSink;
+    const Callback = *const fn (ctx: *anyopaque, bytes: []const u8, err: ?JSC.JSValue, is_async: bool) void;
+
+    ctx: *anyopaque,
+    onFinishedBuffering: Callback,
+
+    js_sink: ?*ArrayBufferSink.JSSink = null,
+    byte_stream: ?*JSC.WebCore.ByteStream = null,
```
Jarred-Sumner, line 1277: it's not always a ByteStream? Is this specific to fetch() and only fetch()?

src/bun.js/webcore/streams.zig
```diff
@@ -52,6 +52,59 @@ pub const ReadableStream = struct {
     value: JSValue,
     ptr: Source,
 
+    pub const Strong = struct {
+        held: JSC.Strong = .{},
+        self: ?ReadableStream = null,
```
Jarred-Sumner: JSC.Strong has the inner value here, it's not a good idea to have both fields

test/js/workerd/html-rewriter.test.js
```diff
@@ -427,3 +427,46 @@ it("#3520", async () => {
     ["šž", "🕵🏻"],
   ]);
 });
+
+it("works with fetch", async () => {
```
Jarred-Sumner: this needs a lot more tests before we can be confident it works

src/bun.js/webcore/body.zig
cirospaciari: js_sink for `.Direct` and `.Javascript` + byte_stream for `.Bytes`, this was inspired by `Bun.serve` code, should not be specific to only `fetch`, `.Blob, .File` should be handled by `toBlobIfPossible` if the comment is true

src/bun.js/api/html_rewriter.zig
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/28eafc84e0d67e910bd9af656426aa3cdfba5c12 and https://github.com/oven-sh/bun/pull/4418/commits/e4ae61ac86219e56a766cd85712bd3d1749f7f4c

src/bun.js/webcore/streams.zig
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/28eafc84e0d67e910bd9af656426aa3cdfba5c12 and https://github.com/oven-sh/bun/pull/4418/commits/e4ae61ac86219e56a766cd85712bd3d1749f7f4c

src/bun.js/webcore/streams.zig
```diff
@@ -54,53 +54,52 @@ pub const ReadableStream = struct {
...
                 .held = JSC.Strong.create(this.value, globalThis),
             };
         }
 
         pub fn get(this: *Strong) ?ReadableStream {
-            if (this.held.has()) {
-                return this.self;
+            if (this.globalThis) |globalThis| {
+                if (this.held.get()) |value| {
+                    return ReadableStream.fromJS(value, globalThis);
```
cirospaciari, line 81: this did not work as expected for some reason

cirospaciari: Looks like server does `stream.detach` so thats why was not working, we cannot detach anymore

src/bun.js/api/html_rewriter.zig
```diff
@@ -344,7 +340,9 @@ pub const HTMLRewriter = struct {
         rewriter: *LOLHTML.HTMLRewriter,
         context: LOLHTMLContext,
         response: *Response,
-        input: JSC.WebCore.AnyBlob = undefined,
+        bodyValueBufferer: ?JSC.WebCore.BodyValueBufferer = null,
+        tmp_sync_error: ?JSC.JSValue = null,
```
Jarred-Sumner, line 344: this tmp_sync_error is low quality. you can do better (example: a union(enum) instead of 2 optionals and an error) 

but you can do that in a follow-up

src/bun.js/webcore/streams.zig
```diff
@@ -3068,12 +3122,24 @@ pub fn ReadableStreamSource(
...
+            if (this.ref_count == 0 or this.deinited) {
+                return 0;
+            }
+
+            this.ref_count -= 1;
+            if (this.ref_count == 0) {
+                this.deinited = true;
+                deinit_fn(&this.context);
+            }
+            return this.ref_count;
```
Jarred-Sumner, line 3144: UAF

src/bun.js/webcore/streams.zig
```diff
@@ -52,6 +52,48 @@ pub const ReadableStream = struct {
...
+            if (this.globalThis) |globalThis| {
+                if (this.held.get()) |value| {
+                    return ReadableStream.fromJS(value, globalThis);
+                }
+            }
+            return null;
+        }
+
+        pub fn deinit(this: *Strong) void {
+            if (this.get()) |readable| {
```
Jarred-Sumner, line 88: This leaks memory in the case where the Strong's ptr exists but the value has been cleared from it. We can always call `this.held.deinit()` at the end.

src/bun.js/webcore/body.zig
```diff
@@ -1258,3 +1263,345 @@ pub fn BodyMixin(comptime Type: type) type {
...
+        const locked = &value.Locked;
+        if (locked.readable) |stream_| {
+            const stream: JSC.WebCore.ReadableStream = stream_;
+            stream.value.ensureStillAlive();
+            // keep the stream alive until we're done with it
+            sink.readable_stream_ref = try JSC.WebCore.ReadableStream.Strong.init(stream, sink.global);
+
+            value.* = .{ .Used = {} };
+
+            if (stream.isLocked(sink.global)) {
```
Jarred-Sumner, line 1515: This leaks memory when the stream is already locked. 

sink.readable_stream_ref needs to only be created after used.

src/bun.js/webcore/body.zig
```diff
@@ -1258,3 +1263,345 @@ pub fn BodyMixin(comptime Type: type) type {
...
+            buffer_stream,
+            @as(**anyopaque, @ptrCast(&signal.ptr)),
+        );
+
+        assignment_result.ensureStillAlive();
+
+        // assert that it was updated
+        std.debug.assert(!signal.isDead());
+
+        if (assignment_result.toError() != null) {
```
Jarred-Sumner: Use `.isError()` because `toError` consumes it which potentially creates a single-use object

src/bun.js/webcore/body.zig
```diff
@@ -1258,3 +1263,345 @@ pub fn BodyMixin(comptime Type: type) type {
...
+                }
+            },
+        }
+    }
+    fn onStreamPipe(sink: *@This(), stream: JSC.WebCore.StreamResult, allocator: std.mem.Allocator) void {
+        var stream_needs_deinit = stream == .owned or stream == .owned_and_done;
+
+        defer {
+            if (stream_needs_deinit) {
+                if (stream.isDone()) {
```
Jarred-Sumner: we probably should add a check for which union type it is just to be sure

cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/7cd66913e79f7c27d13efeb30801dfcbbabe45e6

src/bun.js/webcore/body.zig
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/7cd66913e79f7c27d13efeb30801dfcbbabe45e6

src/bun.js/webcore/body.zig
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/7cd66913e79f7c27d13efeb30801dfcbbabe45e6

src/bun.js/webcore/streams.zig
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/7cd66913e79f7c27d13efeb30801dfcbbabe45e6

src/bun.js/webcore/streams.zig
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/4418/commits/7cd66913e79f7c27d13efeb30801dfcbbabe45e6

PR #4421, opened 2023-08-31
`bun install` correctly join dependency URLs

### LINKED ISSUE #3899: `bun install` uses incorrect url with private registry

0.7.1

_No response_

Setup bunfig.toml for a private registry: 
`[install]
registry = {url="https://artifactory.xxx.yyy/artifactory/api/npm/my-npm/",token="..."}
`
run bun install

Console logs show 
`GET https://artifactory.xxx.yyy/artifactory/api/npm/my-npm/@types/eslint-visitor-keys - 200
`
(or nothing)
packages are installed

### What do you see instead?

Console logs show 
`GET https://artifactory.xxx.yyy//artifactory/api/npm/my-npm/@types/eslint-visitor-keys - 500
`

packages are not installed

### Additional information

Private registry has a self signed cert, but shouldn't affect this behaviour.

Parzival-3141:
### What does this PR do?
Closes #3899
Registry and package URLs are now joined using `WTF`, instead of blindly format printing.
Errors are now reported when enqueuing dependencies as well.

- [x] Code changes

I wrote automated tests.


src/install/install.zig
```diff
@@ -3432,7 +3398,18 @@ pub const PackageManager = struct {
                 i,
                 &dependency,
                 resolution,
-            ) catch {};
+            ) catch |err| {
+                this.log.addZigErrorWithNote(
```
Jarred-Sumner: This will cause failing optional dependencies to fail the build and it should not 

Instead, I suggest setting level to warn if the dependency.behavior is optional or peer dependency

Jarred-Sumner:
Can you remove the changes to builtins?

Parzival-3141:
I rebased and removed the builtin changes. Just needs to be updated with`main`.

Jarred-Sumner:
Thank you

Issue #4430, opened 2023-08-31
improve http.ServerResponse


sirenkovladd:
### What version of Bun is running?

0.8.2+9334fbe9b4facdfa9057a25e88e998f1fd2e346a

### What platform is your computer?

Linux 6.2.0-26-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
const http = require('http');
const res = new http.ServerResponse({});
console.log(res.constructor.name);
```

### What is the expected behavior?

```sh
❯ node test-bun.js
ServerResponse
```

### What do you see instead?

```sh
❯ bun test-bun.js 
621 | }
622 | var OriginalWriteHeadFn, OriginalImplicitHeadFn;
623 | 
624 | class ServerResponse extends Writable {
625 | 
626 |   constructor({ req, reply }) {
                              ^
TypeError: undefined is not an object
      at new ServerResponse (node:http:626:26)
      at <path>/test-bun.js:4:12
      at globalThis (<path>/test-bun.js:5:32)
```

### Additional information

https://github.com/fastify/light-my-request/blob/v5.10.0/lib/response.js#L10

btmnk:
I would really love to see these changes in a release. If I'm not mistaken this currently also affects @fastify/websocket.
Fastify works great with bun so far but the websocket integration breaks with the node:http implementation in bun (Same error as stated above).

I wonder if there is a workaround I can use until this gets released? (Except for using node again :D)

grefrit:
@btmnk Did you find workaround for fastify/websocket? It seems that assignSocket not implemented error is the only thing that stops me from using Bun

btmnk:
> @btmnk Did you find workaround for fastify/websocket? It seems that assignSocket not implemented error is the only thing that stops me from using Bun

Unfortunately not yet.. I'm using node again for the project until this gets sorted.

(I didn't try much though, so there might be a workaround but the project wasn't that important so I didn't invest much time)

PR #4431, opened 2023-08-31
improve http.ServerResponse

sirenkovladd:
### What does this PR do?

fix: https://github.com/oven-sh/bun/issues/4430
more compatible http

add ServerResponse.assignSocket
change ServerResponse.constructor 

- [ ] Code changes

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -35,6 +35,10 @@ const validateHeaderValue = (name, value) => {
   }
 };
 
+function ERR_HTTP_SOCKET_ASSIGNED() {
+  return new Error(`ServerResponse has an already assigned socket`);
+}
+
 // Cheaper to duplicate this than to import it from node:net
 function isIPv6(input) {
   const v4Seg = "(?:[0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])";
@@ -540,7 +544,7 @@ class Server extends EventEmitter {
 
           const upgrade = http_req.headers.upgrade;
 
-          const http_res = new ResponseClass({ reply, req: http_req });
+          const http_res = new ResponseClass(http_req, reply);
 
           http_req.socket[kInternalSocketData] = [_server, http_res, req];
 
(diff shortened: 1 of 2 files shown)


src/js/node/http.ts
```diff
@@ -953,11 +957,43 @@ class OutgoingMessage extends Writable {
...
+  }
+}
+
 let OriginalWriteHeadFn, OriginalImplicitHeadFn;
-class ServerResponse extends Writable {
+class ServerResponse extends OutgoingMessage {
   declare _writableState: any;
 
-  constructor({ req, reply }) {
+  constructor(req, reply) {
```
paperclover, line 1066: My only worry is that in Node, this `reply` is an object of `options` (though i couldnt find any docs on it). I think this should be fine though.

paperclover (approved):

sirenkovladd:
Any updates?

grefrit:
Any updates on this? assignSocket not implemented error breaks fastify/websocket

paperclover (changes requested):
sorry, i did review this a while ago and did approve. but this back in my inbox i really dont think we should merge changing the signature of ServerResposne/OutgoingMessage, thing is neither before or after is the correct signature, which is problematic

i think the other changes are reasonable though. if you can undo the constructor signature change we should merge this

paperclover:
also, since this alot of the build system changed, `src/js/out` was removed

sirenkovladd:
> ...i really dont think we should merge changing the signature of ServerResposne/OutgoingMessage, thing is neither before or after is the correct signature, which is problematic

Are you talking about `class ServerResponse extends Writable` -> `class ServerResponse extends OutgoingMessage`?
Or about `ServerResponse.assignSocket`?

src/js/node/http.ts
paperclover, line 1066: this is what im referring to. the other changes i think are ok

sirenkovladd: But this is the main problem that this PR solves
many people face this problem, also many examples are related
the problem is that now ServerResponse is not compatible with node
https://github.com/nodejs/node/blob/v21.3.0/lib/_http_server.js#L195

sirenkovladd: I think this change is necessary
if such a change is a concern, I can make this parameters more compatible with the current version of bun

sirenkovladd: https://github.com/nodejs/node/blob/v21.3.0/lib/_http_server.js#L195

paperclover: ah i see. this is okay then.

paperclover (approved):

sirenkovladd:
Is there anything left to do for the merge?

PR #4435, opened 2023-08-31
fix(runtime): fix dns_resolver crash

paperclover:
### What does this PR do?

- [X] Code changes

I did a release build locally as the crash does not happen in debug builds, and then ran `bun test dns`


paperclover:
yay, no more c pointers

Jarred-Sumner (changes requested):
The test continues to crash on macOS arm64

Hanaasagi:
@paperdave  

Sorry my bad 🙇

The `buf` is `undefined`, and will no be filled with with `0xaa` in release build. I use the slice `buf[1..]` to format `ip`. The value of `buf[0]` may be 0.

https://github.com/oven-sh/bun/blob/59edbe645ca568aa6438f8fa53c1a567cf1159c0/src/bun.js/api/bun/dns_resolver.zig#L2344-L2351


the `size` will be 0 here

https://github.com/oven-sh/bun/blob/59edbe645ca568aa6438f8fa53c1a567cf1159c0/src/bun.js/api/bun/dns_resolver.zig#L2368

---- 

git patch

```diff
diff --git a/src/bun.js/api/bun/dns_resolver.zig b/src/bun.js/api/bun/dns_resolver.zig
index 7364c2ffa..fab0e675c 100644
--- a/src/bun.js/api/bun/dns_resolver.zig
+++ b/src/bun.js/api/bun/dns_resolver.zig
@@ -2342,6 +2342,7 @@ pub const DNSResolver = struct {
             // Port range is 6 bytes (e.g., ':65535').
             // Null terminator '\0' uses 1 byte.
             var buf: [INET6_ADDRSTRLEN + 2 + 6 + 1]u8 = undefined;
+
             const family = cur.*.family;
 
             const ip = if (family == std.os.AF.INET6) blk: {
@@ -2365,7 +2366,7 @@ pub const DNSResolver = struct {
                 port = IANA_DNS_PORT;
             }
 
-            const size = bun.len(bun.cast([*:0]u8, &buf));
+            const size = bun.len(bun.cast([*:0]u8, buf[1..])) + 1;
             if (port == IANA_DNS_PORT) {
                 values.putIndex(globalThis, i, JSC.ZigString.init(buf[1..size]).withEncoding().toValueGC(globalThis));
             } else {

```

paperclover:
good catch. i was just about to start debugging this after spending way too much time fixing the failing fs test here.

paperclover:
crash in the test runner. narrowed to.
```
test("dns.promises.reverse", async () => {
  await dns.promises.reverse("8.8.8.8");
  await dns.promises.reverse("1.1.1.1");
  Bun.gc(true);
  await dns.promises.reverse("2606:4700:4700::1111");
});
```

been trying to figure out where

Hanaasagi:
`hostent` should not be freed.

https://github.com/oven-sh/bun/blob/40d749b480a00bbc63beaac16625c529709f1cd6/src/bun.js/api/bun/dns_resolver.zig#L1501


https://c-ares.org/ares_gethostbyaddr.html

> On successful completion of the query, the callback argument hostent points to a struct hostent containing the name of the host returned by the query. The callback need not and should not attempt to free the memory pointed to by hostent; the ares library will free it when the callback returns. If the query did not complete successfully, hostent will be NULL.

Issue #1678, opened 2022-12-29
Follow XDG Base Directory Specification


jtrv:
### What is the problem this feature would solve?

The `.bun/` directory unnecessarily pollute users' `$HOME` directory.
Related: #696, #965

### What is the feature you are proposing to solve the problem?

The [XDGBDS](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) defines where these files should be placed with environment variables to avoid polluting the `$HOME` directory.

From the XDGBDS:

> The XDG Base Directory Specification is based on the following concepts:
>
>   - There is a single base directory relative to which user-specific data files should be written. This directory is defined by the environment variable `$XDG_DATA_HOME`.
>
>   - There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable `$XDG_CONFIG_HOME`.
>
>   - There is a single base directory relative to which user-specific state data should be written. This directory is defined by the environment variable `$XDG_STATE_HOME`.
>
>   - There is a single base directory relative to which user-specific executable files may be written.
>
>   - There is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable `$XDG_CACHE_HOME`.
>
> User-specific executable files may be stored in `$HOME/.local/bin`.

### Implementing XDGBDS should work as follows:

_IF_ a corresponding XDG environment variable exists _THEN_ use it's location according to the spec.

_ELSE_ If the corresponding XDG_* env var is not defined, the program must fallback to the specified defaults:

env | fallback
-- | --
XDG_DATA_HOME | ~/.local/share
XDG_CONFIG_HOME | ~/.config
XDG_STATE_HOME | ~/.local/state
XDG_CACHE_HOME | ~/.cache

So the default/fallback directory layout mapping should be:

current | xdg compliant
-- | --
~/.bun/bin/bun | ~/.local/bin/bun
~/.bun/install/cache | ~/.cache/bun
~/.bun/install/global/node_modules | ~/.local/lib/node_modules
~/.bun/install/global/package.json | ~/.config/bun/package.json
~/.bun/install/global/bun.lockb | ~/.config/bun/bun.lockb

In addition, globally installed modules should go in `~/.local/lib/node_modules`, with binaries symlinked under `~/.local/bin` (which is what npm does).

### What alternatives have you considered?

We can also fallback to the `.bun` directory for backwards compatibility but as others have mentioned below this is not ideal and not part of the XDGBDS. 

[...]

PR #4438, opened 2023-08-31
Use `$XDG_CACHE_HOME/bun` instead of `$XDG_CACHE_HOME/.bun` (when the var is set).

lgarron:
This addresses the comment at https://github.com/oven-sh/bun/issues/1678#issuecomment-1668660285

> … I have yet to see see any other program use a `.` prefix for folders
> *inside* the XDG convention folders. This may cause the folder to be
> unexpectedly hidden when viewing it in a file explorer app or running
> `ls` without `-a`. (That is, someone may expect `~/.cache` to be
> hidden — but that all folders are visible once they are viewing that
> hidden folder.)

The only previous workaround would have been to set `$BUN_INSTALL`, but
that affects more than the cache directory (sort of defeating the point
of `$XDG_CACHE_HOME` in the first place).

Since this:

- only affects the subset of users who have explictly
opted into `$XDG_CACHE_HOME`, and
- only affects cache data, which is meant to be safe to abandon/lose at
  any moment,

… it shouldn't be harmful to leave the old directory lying around and
allow a new one to be created. But it would also be possible to (re)move the
existing directory with an additional change.

### What does this PR do?

- [x] Code changes

This is an adaptation of existing code patterns — modifying a string in two places and forking an `if` statement into two copies.
I'm unable to build `bun` on my system at the moment, but this is as safe a change as I could author.

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -4469,7 +4469,7 @@ pub const PackageManager = struct {
                 return try std.fs.cwd().makeOpenPathIterable(path, .{});
             }
 
-            if (orelse bun.getenvZ("HOME")) |home_dir| {
+            if (bun.getenvZ("HOME")) |home_dir| {
                 var buf: [bun.MAX_PATH_BYTES]u8 = undefined;
                 var parts = [_]string{
                     ".bun",
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -3674,7 +3674,7 @@ pub const PackageManager = struct {
         }
 
         if (env.map.get("XDG_CACHE_HOME")) |dir| {
-            var parts = [_]string{ dir, ".bun/", "install/", "cache/" };
+            var parts = [_]string{ dir, "bun/", "install/", "cache/" };
             return CacheDir{ .path = Fs.FileSystem.instance.abs(&parts), .is_node_modules = false };
         }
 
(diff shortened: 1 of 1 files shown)


src/install/install.zig
```diff
@@ -4459,7 +4459,17 @@ pub const PackageManager = struct {
...
+                var buf: [bun.MAX_PATH_BYTES]u8 = undefined;
+                var parts = [_]string{
+                    "bun",
+                    "bin",
+                };
+                var path = Path.joinAbsStringBuf(home_dir, &buf, &parts, .auto);
+                return try std.fs.cwd().makeOpenPathIterable(path, .{});
+            }
+
+            if (orelse bun.getenvZ("HOME")) |home_dir| {
```
booniepepper:
```suggestion
            if (bun.getenvZ("HOME")) |home_dir| {
```

src/install/install.zig
```diff
@@ -4459,7 +4459,17 @@ pub const PackageManager = struct {
...
                 var buf: [bun.MAX_PATH_BYTES]u8 = undefined;
                 var parts = [_]string{
                     ".bun",
```
booniepepper, line 4475: I think I can't suggest the exact change(?) since it goes outside the diff context, but this should be:

```zig
var parts = [_]string{ ".cache", "bun", "bin" };
```

lgarron: This would change the cache to be outside of the default bun dir for users who don't have `$XDG_CACHE_HOME` set (which was outside the scope of my PR goals).

I'd be in favor of that, but is that desirable to the `bun` team?

lgarron:
Notwithstanding the discussion at https://github.com/oven-sh/bun/issues/1678, I think everything in this PR is still relevant.

booniepepper:
FWIW it looks good to me. I think it can go farther but that can be a different PR

lgarron:
I'd still be rather glad to see this land in some form.

PR #4439, opened 2023-09-01
fix(runtime): `fs.cp` edge cases

### LINKED ISSUE #4436: Recursive `cp` fails when the destination has a trailing slash
Glad to see https://github.com/oven-sh/bun/pull/4340 in canary! I've run into the following issue with the initial implementation:

0.8.2+9334fbe9b4facdfa9057a25e88e998f1fd2e346a

macOS Ventura 13.5.1

```js
import { mkdir, cp, writeFile } from "node:fs/promises";

// Set up
await mkdir("./a/", { recursive: true });
writeFile("./a/foo.txt", "hello world");
await mkdir("./b/", { recursive: true });

// Copy from `./a/` to `./b/c/`
await cp("./a/", "./b/c/", { recursive: true });
```

A folder at `./b/c/` is created, containing `foo.txt`.

(This is what `node` does. Note that both paths have trailing slashes, so they unambiguously refer to folders and not files.)

### What do you see instead?

```
   path: "./b/"
 syscall: "clonefile"
   errno: -2
```

### Additional information

For a "real-world" case, try:

```shell
git clone https://github.com/lgarron/barely-a-dev-server && cd barely-a-dev-server
git checkout 79f70b086365a54b3b7fe7f09e3cc34dcd96cf17
bun run ./test/build.js
```

On [this line](https://github.com/lgarron/barely-a-dev-server/blob/923c708a0f2a0c17cd5ec5d5a99aade28d12ac0b/src/index.js#L28) it should be copying the contents of `./test/root/` to `dist/test/root/` but instead it errors with:

```
   path: "dist/test/ro"
 syscall: "clonefile"
   errno: -2
```

Note that:

- The error shows a truncated path of `dist/test/ro`
- Even if `dist/test/root` exists, it is *deleted* by `bun`(!) instead of having files copied to it.

paperclover:
### What does this PR do?

Closes #4436 

- Error messages did not always contain the proper path.
- Fix a crash during one of the errors.
- `clonefile` path can `ENOENT` if there the destination directory doesnt exist.

- [X] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -523,6 +523,9 @@ pub const AsyncCpTask = struct {
     /// The maintask thread starts this at 1 and decrements it at the end, to avoid the promise being resolved while new tasks may be added.
     subtask_count: std.atomic.Atomic(usize),
 
+    /// This contains the error buffer, maybe a better approach is to only allocate this when the error happens
+    node_fs: NodeFS,
+
     pub fn create(
         globalObject: *JSC.JSGlobalObject,
         cp_args: Arguments.Cp,
@@ -539,6 +542,7 @@ pub const AsyncCpTask = struct {
             .tracker = JSC.AsyncTaskTracker.init(vm),
             .arena = arena,
             .subtask_count = .{ .value = 1 },
+            .node_fs = NodeFS{},
         };
         task.ref.ref(vm);
         task.args.src.toThreadSafe();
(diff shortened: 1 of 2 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -523,6 +523,9 @@ pub const AsyncCpTask = struct {
     /// The maintask thread starts this at 1 and decrements it at the end, to avoid the promise being resolved while new tasks may be added.
     subtask_count: std.atomic.Atomic(usize),
 
+    /// This contains the error buffer, maybe a better approach is to only allocate this when the error happens
+    node_fs: NodeFS,
+
```
paperclover: im not super happy about this, but this is the simplest path though

Jarred-Sumner:
Thank you

PR #4441, opened 2023-09-01
only set initial debugger breakpoint once

dylan-conway:
### What does this PR do?

- [ ] Code changes

tested manually that `--hot` and the test runner only set the breakpoint once.


Jarred-Sumner:
Thank you

PR #4449, opened 2023-09-02
`bun install` support private GitHub dependencies using oauth2

### LINKED ISSUE #3555: Support for private GitHub modules using GitHub oauth2 fine-grained personal access tokens in bun install

Installing private modules using GitHub oauth2 fine-grained personal access tokens (PAT):

```
"dependencies": {
    "@biz/module": "https://oauth2:github_pat_11...w@github.com/Business/module.git#master"
}
```

Installation of modules behind GitHub oauth2 fine-grained PAT is currently not supported with `bun install`.

https://github.blog/2022-10-18-introducing-fine-grained-personal-access-tokens-for-github/

```
error: package "@biz/module" not found registry.npmjs.org/@biz/module 404
```

```
error: @biz/module@https://oauth2:github_pat_11...w@github.com/Business/module.git#master failed to resolve
```

The feature is to support installation of private modules behind GitHub oauth2 fine-grained PAT.

There are none, AFAIK.

Parzival-3141:
### What does this PR do?
Closes #3555
Adds support for dependencies using a OAuth2 GitHub URL. E.g.
```json
"dependencies": {
    "@biz/module": "https://oauth2:your_token_here@github.com/user/repo.git#master"
}
```
It's more of a workaround than I'd like, as dependency parsing and request header generation is mostly hardcoded for npm style workflows. This makes it hard to modify how dependencies are fetched through `bun install`.

The included test requires fetching a private GitHub repo using a real OAuth2 token.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/.github/workflows/bun-linux-build.yml
+++ b/.github/workflows/bun-linux-build.yml
@@ -192,6 +192,7 @@ jobs:
       - id: test
         name: Test (node runner)
         env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SMTP_SENDGRID_SENDER: ${{ secrets.SMTP_SENDGRID_SENDER }}
           TLS_MONGODB_DATABASE_URL: ${{ secrets.TLS_MONGODB_DATABASE_URL }}
           TLS_POSTGRES_DATABASE_URL: ${{ secrets.TLS_POSTGRES_DATABASE_URL }}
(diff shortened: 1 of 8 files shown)


Jarred-Sumner:
This increases the size of the repository struct, so we will need to do another, more tedious approach to avoid invalidating every user of bun install's lockfile

Instead, we can keep the tag as something like github_oauth2 and then possibly add oauth2 tokens to the end of the lockfile?

Parzival-3141:
After further thought, it seems insecure to allow storing the token inside the `package.json` file. Authentication should be a system-level thing, not in a file that gets committed. I propose we allow passing an `OAUTH2_TOKEN` environment variable, either through `package.json` or automatically detected by `bun install`. This could also be generalized to any form of authentication, although it'd probably have to use the former approach.

Issue #4198, opened 2023-08-18
`unlinkSync` throws EPERM when the target file doesn't exist


riywo:
### What version of Bun is running?

0.7.3

### What platform is your computer?

Linux 5.10.16.3-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

`foo.js`:

```js
import { unlinkSync } from "node:fs";

unlinkSync("aaaa");
```

```
$ ls
foo.js

$ bun foo.js   
1 | import { unlinkSync } from "node:fs";
2 | 
3 | unlinkSync("aaaa");
   ^
EPERM: Operation not permitted
   path: "aaaa"
 syscall: "unlink"
   errno: -1

      at /foo/foo.js:3:0
```

### What is the expected behavior?

Node.js returns `ENOENT`, so `ENOENT` is expected:

```
$ node foo.js  
node:fs:1899
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, unlink 'aaaa'
    at unlinkSync (node:fs:1899:3)
    at file:///foo/foo.js:3:1
    at ModuleJob.run (node:internal/modules/esm/module_job:192:25)
    at async DefaultModuleLoader.import (node:internal/modules/esm/loader:228:24)
    at async loadESM (node:internal/process/esm_loader:40:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12) {
  errno: -2,
  syscall: 'unlink',
  code: 'ENOENT',
  path: 'aaaa'
}

Node.js v20.5.1
```

### What do you see instead?

`EPERM` like above:

```
EPERM: Operation not permitted
   path: "aaaa"
 syscall: "unlink"
   errno: -1

```

### Additional information

Probably due to this difference, `eslint` doesn't work with Bun runtime:

```
$ strace -f bun eslint
...
[pid 14449] unlink("/foo/.eslintcache") = -1 ENOENT (No such file or directory)
[pid 14449] write(2, "\33[0m\33[31m", 9) = 9
[pid 14449] write(2, "\nOops! Something went wrong! :(\n"..., 544
Oops! Something went wrong! :(

ESLint: 8.47.0

EPERM: Operation not permitted
    at unlinkSync (native)
    at executeOnFiles (/foo/node_modules/eslint/lib/cli-engine/cli-engine.js:785:14)
    at <anonymous> (/foo/node_modules/eslint/lib/eslint/eslint.js:551:8)
[...]

PR #4461, opened 2023-09-03
fix(syscall): fix handling syscall errno

Hanaasagi:
### What does this PR do?

Close: #4198 

- [x] Code changes

I wrote automated tests

### Why `unlinkSync` throws `EPERM`

https://github.com/oven-sh/bun/blob/5c43744bce358f132a63d5002baffd7647d47a76/src/sys.zig#L912-L921

`unlinkSync` uses the `sys.unlink` which behaves differently at compile time based on whether we link to `libc` or not.

https://ziglang.org/documentation/master/std/#A;std:os.system

If we link to `libc`, we use the `unlink` function from the C library. From `man 2 unlink`, we should read the `errno`

> RETURN VALUE
       On success, zero is returned.  On error, -1 is returned, and errno is set to indicate the error.

https://github.com/oven-sh/bun/blob/5c43744bce358f132a63d5002baffd7647d47a76/src/sys.zig#L247-L264

`-1` will be converted to `1` (`EPERM`), so we get a `Operation not permitted`.

### Why `open` work correctly before?

https://github.com/oven-sh/bun/blob/5c43744bce358f132a63d5002baffd7647d47a76/src/sys.zig#L419-L434

`Syscall.system.openat` is https://github.com/ziglang/zig/blob/62f727eedb2b8d8b8f922d52a0ff4218a92e6cff/lib/std/os/linux.zig#L750. This is not `libc` function, just a linux platform syscall.

### Why `unlinkSync` work on darwin?

On darwin, libc is linked, and the value directly from `std.os.errno` is used in `getErrno`.

### DIFF
--- a/src/sys.zig
+++ b/src/sys.zig
@@ -257,8 +257,8 @@ pub fn getErrno(rc: anytype) bun.C.E {
     const Type = @TypeOf(rc);
 
     return switch (Type) {
-        comptime_int, usize => return std.os.linux.getErrno(@as(usize, rc)),
-        i32, c_int, isize => return std.os.errno(rc),
+        comptime_int, usize => std.os.linux.getErrno(@as(usize, rc)),
+        i32, c_int, isize => std.os.errno(rc),
         else => @compileError("Not implemented yet for type " ++ @typeName(Type)),
     };
 }
--- a/src/sys.zig
+++ b/src/sys.zig
@@ -256,16 +256,11 @@ pub fn getErrno(rc: anytype) bun.C.E {
     if (comptime use_libc) return std.os.errno(rc);
     const Type = @TypeOf(rc);
 
-    switch (Type) {
+    return switch (Type) {
         comptime_int, usize => return std.os.linux.getErrno(@as(usize, rc)),
-        i32, c_int, isize => {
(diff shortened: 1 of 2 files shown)


src/sys.zig
```diff
@@ -257,8 +257,8 @@ pub fn getErrno(rc: anytype) bun.C.E {
     const Type = @TypeOf(rc);
 
     return switch (Type) {
-        comptime_int, usize => std.os.linux.getErrno(@as(usize, rc)),
-        i32, c_int, isize => std.os.linux.getErrno(@as(usize, @bitCast(@as(isize, rc)))),
+        comptime_int, usize => return std.os.linux.getErrno(@as(usize, rc)),
```
Hanaasagi: In Zig, the return values of syscalls are all of type `usize`. `libc` use the `c_int`.

https://github.com/ziglang/zig/blob/master/lib/std/os/linux/x86_64.zig
https://github.com/ziglang/zig/blob/master/lib/std/os/linux/arm64.zig

Jarred-Sumner:
oh this is great

thank you

Jarred-Sumner (approved):

PR #4467, opened 2023-09-03
feat(runtime): Implement `fs.watchFile`

paperclover:
### What does this PR do?

Implements `fs.watchFile` and `fs.unwatchFile`, as well as the `StatWatcher` class it returns. This code is not optimal in terms of performance but it does implement the API as correctly as possible.

This implementation isn't very optimal but it doesn't block the JS thread very much.

- [X] Code changes

I did mostly manual testing and wrote some unit tests. After 1.0 we can revisit writing more tests that hit more edge cases.

I used debug logging to verify `deinit` is called when needed.


birkskyum:
Closes #3812

src/bun.js/node/node_fs_stat_watcher.zig
```diff
@@ -0,0 +1,508 @@
...
+                prev.next = c.next;
+                curr = c.next;
+                if (c.js_this == .zero) {
+                    c.deinit();
+                } else {
+                    c.used_by_scheduler_thread = false;
+                }
+                continue;
+            }
+            if (now.since(c.last_check) > (@as(u64, @intCast(c.interval)) * 1_000_000 - 500)) {
```
Jarred-Sumner: integer overflow

src/bun.js/node/node_fs_stat_watcher.zig
```diff
@@ -0,0 +1,508 @@
...
+        return this;
+    }
+
+    pub fn append(this: *StatWatcherScheduler, watcher: *StatWatcher) void {
+        log("append new watcher {s}", .{watcher.path});
+        std.debug.assert(watcher.closed == false);
+        std.debug.assert(watcher.next == null);
+
+        this.mutex.lock();
+        defer this.mutex.unlock();
```
Jarred-Sumner: this mutex does nothing

theres no other user of it

src/bun.js/node/node_fs.zig
```diff
@@ -5245,6 +5245,25 @@ pub const NodeFS = struct {
...
+            args.global_this.throwValue((JSC.SystemError{
+                .message = bun.String.init(buf),
+                .path = bun.String.init(args.path.slice()),
+            }).toErrorInstance(args.global_this));
+            return Maybe(Return.Watch){ .result = JSC.JSValue.undefined };
+        };
+        return Maybe(Return.Watch){ .result = watcher };
+    }
     pub fn unwatchFile(_: *NodeFS, _: Arguments.UnwatchFile, comptime _: Flavor) Maybe(Return.UnwatchFile) {
         return Maybe(Return.UnwatchFile).todo;
```
Jarred-Sumner, line 5268: lol

src/bun.js/node/node_fs_stat_watcher.zig
```diff
@@ -0,0 +1,508 @@
...
+        var this = timer.ext(StatWatcherScheduler).?;
+        JSC.WorkPool.schedule(&this.task);
+    }
+
+    pub fn workPoolCallback(task: *JSC.WorkPoolTask) void {
+        var this: *StatWatcherScheduler = @fieldParentPtr(StatWatcherScheduler, "task", task);
+        // Instant.now will not fail on our target platforms.
+        const now = std.time.Instant.now() catch unreachable;
+
+        var prev: *StatWatcher = this.head.?;
```
Jarred-Sumner: since you used a linked list, this can be made race-free by swapping this.head with null

and at the end of the loop, swapping it back and appending the new head to the tail

Jarred-Sumner (changes requested):
A couple comments

Jarred-Sumner:
This looks fine but there should really be a test that watches 1000 files and unwatches 1000. It should test that it doesn’t crash. We should also test duplicate files are handled correctly. I imagine we could port some of node’s tests

PR #4475, opened 2023-09-04
[WIP] [install] allow parallel execution of `postinstall` scripts

alexlamsl:
- fixes spurious exit code 42 from `spawn()`

### What does this PR do?

TODO: suppress log output from `postinstall` scripts instead of spewing all over the console &minus; only display logs from failed scripts to aid debugging.

Making this PR now because during development I have discovered the root cause for the `exitCode=42` bug, so seems like a good point to check this in.

- [x] Code changes

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -41,6 +41,7 @@ const NpmArgs = struct {
 const yarn_commands: []u64 = @import("./list-of-yarn-commands.zig").all_yarn_commands;
 
 const ShellCompletions = @import("./shell_completions.zig");
+const PosixSpawn = @import("../bun.js/api/bun/spawn.zig").PosixSpawn;
 
 pub const RunCommand = struct {
     const shells_to_search = &[_]string{
@@ -49,7 +50,7 @@ pub const RunCommand = struct {
         "zsh",
     };
 
-    pub fn findShell(PATH: string, cwd: string) ?string {
+    pub fn findShell(PATH: string, cwd: string) ?stringZ {
         if (comptime Environment.isWindows) {
             return "C:\\Windows\\System32\\cmd.exe";
         }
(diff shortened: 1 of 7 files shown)


src/cli/run_command.zig
```diff
@@ -265,57 +310,73 @@ pub const RunCommand = struct {
...
-                        Output.flush();
-                    }
-
-                    Global.exit(code);
-                }
-            },
-            .Signal => |signal| {
-                Output.prettyErrorln("<r><red>error<r><d>:<r> script <b>\"{s}\"<r> exited with {any}<r>", .{ name, bun.SignalCode.from(signal) });
+    pub fn waitForPackageScript(name: string, pid: std.os.pid_t, is_sync: bool) bool {
+        switch (PosixSpawn.waitpid(pid, if (is_sync) 0 else std.os.W.NOHANG)) {
```
Jarred-Sumner: this needs a `while (true)` when `is_sync` incase signals are sent but there's no action necessary

src/cli/run_command.zig
```diff
@@ -225,26 +226,70 @@ pub const RunCommand = struct {
...
+        }
+
+        var attr = try PosixSpawn.Attr.init();
+        defer attr.deinit();
+        try attr.set(@intCast(flags));
+        try attr.resetSignals();
+
+        var actions = try PosixSpawn.Actions.init();
+        defer actions.deinit();
+        try actions.inherit(0);
```
Jarred-Sumner, line 248: I don't think we can ship this with `inherit`. It's going to be too noisy.

alexlamsl: sounds reasonable &minus; will work on this now

src/cli/run_command.zig
alexlamsl: done in https://github.com/oven-sh/bun/pull/4475/commits/9bafac63900647655a48bf0be4c2bea8e71fdc2e

paperclover:
was merged to #5077

Issue #4377, opened 2023-08-28
bunfig [loaders] overrides are ignored


SirbyAlive:
### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Linux 6.3.13_1 x86_64 unknown

### What steps can reproduce the bug?

Reproduction repo: https://github.com/SirbyAlive/bun-bunfig-loaders-are-ignored

According to the [**runtime** documentation](https://bun.sh/docs/runtime/configuration#runtime),

```toml
[loaders]
".toml" = "text" 
```
should override the default loaders (or provide a mean to support extra file types), but it is ignored. Meanwhile the flag `--loader .toml:text` does the expected behaviour.

Current output of loading a toml file is:

```bash
$ bun run index.ts
{
  loaders: {
    ".toml": "text"
  }
}
$ bun run --loader .toml:text index.ts
[loaders]
".toml" = "text"
```

### What is the expected behavior?

The expected behaviour is that the tuples `".ext" = "loader"` provided in `[loaders]`  of `bunfig.toml` are honoured.

Hanaasagi:
There's a typo in doc, it should be "loader" instead of "loaders"

https://github.com/oven-sh/bun/blob/18767906db0dd29b2d898f84a023d403c3084d6e/src/bunfig.zig#L686


```toml
[loader]
".toml" = "text"

```

PR #4476, opened 2023-09-04
docs typo: `loaders` -> `loader`

Hanaasagi:
### What does this PR do?

It should be `loader` instead of `loaders`. Close: #4377

https://github.com/oven-sh/bun/blob/18767906db0dd29b2d898f84a023d403c3084d6e/src/bunfig.zig#L686

In command line, we also use the `--loader`, no `--loaders`.

- [ ] Code changes


Jarred-Sumner:
thank you

Issue #4478, opened 2023-09-04
ZlibError with fetch


jmmoser:
### What version of Bun is running?

0.8.2+18767906db0dd29b2d898f84a023d403c3084d6e

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```javascript
// test.js
const req = await fetch('https://api.open-meteo.com/v1/forecast?latitude=44&longitude=-88&hourly=temperature_2m', {
  verbose: true,
});
console.log(await req.json());
```

`bun run test.js`

### What is the expected behavior?

The response from the api is logged to console. `node test.js` works as expected.

### What do you see instead?

```
Request: GET /v1/forecast?latitude=44&longitude=-88&hourly=temperature_2m
        Connection: keep-alive
        User-Agent: Bun/0.8.2
        Accept: */*
        Host: api.open-meteo.com
        Accept-Encoding: gzip, deflate

Request: GET /v1/forecast?latitude=44&longitude=-88&hourly=temperature_2m
        Connection: keep-alive
        User-Agent: Bun/0.8.2
        Accept: */*
        Host: api.open-meteo.com
        Accept-Encoding: gzip, deflate

Request: GET /v1/forecast?latitude=44&longitude=-88&hourly=temperature_2m
        Connection: keep-alive
        User-Agent: Bun/0.8.2
        Accept: */*
        Host: api.open-meteo.com
        Accept-Encoding: gzip, deflate

Request: GET /v1/forecast?latitude=44&longitude=-88&hourly=temperature_2m
        Connection: keep-alive
        User-Agent: Bun/0.8.2
        Accept: */*
        Host: api.open-meteo.com
        Accept-Encoding: gzip, deflate

Response: < 200 OK
<       Date: Mon, 04 Sep 2023 16:02:38 GMT
<       Content-Type: application/json; charset=utf-8
<       Transfer-Encoding: chunked
<       Connection: keep-alive
<       Content-Encoding: deflate

Zlib error: ZlibError
ZlibError: fetch() failed. For more information, pass `verbose: true` in the second argument to fetch()
 path: "https://api.open-meteo.com/v1/forecast?latitude=44&longitude=-88&hourly=temperature_2m"
```

### Additional information

Adding header `Accept-Encoding: identity` mentioned [here](https://github.com/oven-sh/bun/issues/686#issuecomment-1301468824) works.

PR #4483, opened 2023-09-04
fix zlib deflate on fetch

cirospaciari:
### What does this PR do?

Fix https://github.com/oven-sh/bun/issues/4478

- [x] Code changes

 I wrote automated tests


Jarred-Sumner:
If I return a response body with where the first byte is 120 - does that mean it will automatically assume it’s DELFATE now and then decompress?

cirospaciari:
> If I return a response body with where the first byte is 120 - does that mean it will automatically assume it’s DELFATE now and then decompress?

this function is only called when `this.encoding` is `Encoding.deflate` or `Encoding.gzip`
`this.encoding` must be `Encoding.deflate`, so if is deflate we can check the header.

Jarred-Sumner:
Thank you

Issue #4474, opened 2023-09-04
Support `undici` ProxyAgent


Hebilicious:
## Workaround

There's a fetch workaround available.

<img width="415" alt="image" src="https://github.com/oven-sh/bun/assets/13395944/ecfc6ef9-7a6c-4d31-ac78-0a771fcbc86d">

### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

The following code works with node, but doesn't work with bun : 

Undici ProxyAgent : 

```ts
import { ProxyAgent, type RequestInit as UndiciRequestInit, fetch as uFetch } from "undici"

export const fetchWithProxyUndici = async (url: string, options: UndiciRequestInit, proxy: URL) => {
  const dispatcher = new ProxyAgent({
    uri: proxy.host,
    token: `Basic ${Buffer.from(`${proxy.username}:${proxy.password}`).toString("base64")}`
  })
  return uFetch(url, { ...options, dispatcher })
}
```

Something like this should work but it doesn't  : 

```ts
import http from "node:http"
import https from "node:https"
export async function fetchWithNodeProxy(url: string, options: Record<string, any>, proxy: URL) {
  return new Promise<{ text: () => Promise<string>; headers: Headers; status?: number; ok: boolean }>((resolve, reject) => {
    const urlParsed = new URL(url)
    const headers = options.headers || {}
    headers["Proxy-Authorization"] = `Basic ${Buffer.from(`${proxy.username}:${proxy.password}`).toString("base64")}`
    const requestOptions: http.RequestOptions = {
      hostname: proxy.hostname,
      port: proxy.port,
      method: "CONNECT",
      path: `${urlParsed.host}:443`,
      headers
    }
    http.request(requestOptions)
      .on("connect", (res, socket) => {
        console.log("RES !", res.statusCode)
        delete headers["Proxy-Authorization"]
        if (res.statusCode === 200) {
          const agent = new https.Agent({ socket })
          const req = https.get({
            host: urlParsed.hostname,
            path: urlParsed.pathname,
            agent,
          }, (response) => {
            const chunks = [] as Uint8Array[]
            response.on("data", (chunk) => {
              chunks.push(chunk)
            })
            response.on("end", () => {
              console.log("Succesful response !")
              resolve({
                ok: true,
                headers: new Headers(response.headers as any),
[...]

PR #4489, opened 2023-09-05
feat(undici): add `ProxyAgent` for `undici`

### LINKED ISSUE #4457: `ProxyAgent` in `"undici"` doesn't work

0.8.2+5288178c865cd9dcb500b00a65197bf2641d4e5c

Darwin 22.4.0 arm64 arm

hi, I got some problem with undici

I've already get undici installed by bun
```
╰>>> bun add undici
bun add v0.8.2 (5288178c)

 installed undici@5.23.0
```

and here is the sample code 
```ts
import { ProxyAgent } from "undici";

console.log(ProxyAgent.name);

```

but bun reports me the export class `ProxyAgent` not found in undici
```
╰>>> bun run index.ts 
SyntaxError: Import named 'ProxyAgent' not found in module 'undici'.
```

ts-node can run this piece of code properly, print the name of the imported class `ProxyAgent`

```
╰>>> ts-node-esm index.ts 
ProxyAgent
```

### What do you see instead?

```
╰>>> bun run index.ts 
SyntaxError: Import named 'ProxyAgent' not found in module 'undici'.
```

### Additional information

_No response_

ThatOneBro:
### What does this PR do?

This PR adds `ProxyAgent` to the `undici` module, as well as converts `undici.js` -> `undici.ts`. 
There are likely a few bug fixes along with this feature but I still need to add more tests.

Should close #4457 and partially addresses #4474 with the caveat that I don't think TLS is supported yet 

- [x] Code changes

TODO: 
- [ ] Write some tests for proxies


src/js/thirdparty/undici.ts
```diff
@@ -102,36 +101,37 @@ class BodyReadable extends ReadableFromWeb {
 // *   upgrade?: boolean | string | null;
 // *   blocking?: boolean;
 
+type RequestOptions = {
+  dispatcher?: Dispatcher;
+  method: HttpMethod | string;
+  signal?: AbortSignal | e.EventEmitter | null;
+  maxRedirections?: number;
+  body?: string | Buffer | Uint8Array | s.Readable | null | FormData;
+  headers?: IncomingHttpHeaders | string[] | null;
```
Hebilicious, line 110:
```suggestion
  headers?: IncomingHttpHeaders | null;
```

src/js/thirdparty/undici.ts
```diff
@@ -200,21 +202,39 @@ async function request(
...
 
+  let proxy: string | undefined;
+  if (dispatcher && dispatcher instanceof ProxyAgent) {
+    // get proxy uri
+    proxy = dispatcher[kBunUndiciProxyAgentUri];
+    // if proxyAgent.token is defined, put a proxy auth header with token in the headers
+    const token = dispatcher[kBunUndiciProxyAgentToken];
+    if (token) {
+      if (inputHeaders) {
+        if (Array.isArray(inputHeaders)) {
```
Hebilicious, line 213: How can headers be `string[]` ? 🤔

src/js/thirdparty/undici.ts
```diff
@@ -265,6 +285,63 @@ function Undici() {
...
+    if (uri.startsWith("https://")) throw new TypeError("proxy over TLS not implemented");
+    if (!uri.startsWith("http://")) throw new Error("uri must be a valid URL string with the protocol `http://`");
+
+    let token;
+    if (options) {
+      if (typeof options !== "object") throw new TypeError("options must be an object");
+      // Auth token can either be passed via `token` or `auth` prop, though `auth` is now deprecated
+      const { token: _token, auth } = options;
+      token = _token ?? auth;
+      if (token && (typeof token !== "string" || !token.startsWith("Basic ")))
```
Hebilicious, line 317: looking at undici implementation https://github.com/nodejs/undici/blob/83cb522ae0157a19d149d72c7d03d46e34510d0a/lib/proxy-agent.js#L72
`auth` should always need `Basic`, maybe error before and say to use token ?

src/js/thirdparty/undici.ts
```diff
@@ -265,6 +285,63 @@ function Undici() {
...
+  token?: string;
+  headers?: IncomingHttpHeaders;
+};
+
+class ProxyAgent extends Dispatcher {
+  #uri: string;
+  #token?: string;
+  // #headers?: Dict<string | string[]>;
+
+  constructor(uri: string, options: ProxyAgentOptions) {
```
Hebilicious, line 304: `ProxyAgent` should also accept an object as a single argument

src/js/thirdparty/undici.ts
```diff
@@ -265,6 +285,63 @@ function Undici() {
...
+
+  get [kBunUndiciProxyAgentUri]() {
+    return this.#uri;
+  }
+
+  get [kBunUndiciProxyAgentToken]() {
+    return this.#token;
+  }
+
+  // get [kBunUndiciProxyAgentHeaders]() {
```
Hebilicious, line 334: If think headers should be merged with the incoming headers, but if we can't maybe throw a notImplemented error/warning ?

Hebilicious (commented):
Thank you for implementing this ❤️ 
I left a few comments here and there.
I'm also thinking it could be nice to emit a warning/have some TSdoc that inform the user about `fetch(url, { proxy })` when using `fetch(url, { dispatcher })` ?

src/js/thirdparty/undici.ts
ThatOneBro: Good question, this is definitely wrong. I think it's actually supposed to be a `string[][]` or `[string, string][]` more precisely. This when headers are provided as an array of tuples of `["header-name", "header-value"]`. I'm going to fix this type and some other things later.

src/js/thirdparty/undici.ts
ThatOneBro: 100% I missed that, good catch

Zamion101:
Any update on this PR? We really need undici Proxies for sending requests.

Jarred-Sumner:
Closing due to bitrot

PR #4490, opened 2023-09-05
remove `chmod` usage with `--bun` flag

### LINKED ISSUE #4143: --bun flag adds +300 ms

0.7.4+78defe7a87226b5b10766e24fae458a62811dab2

Darwin 22.5.0 arm64 arm

* `bun init`
* add this to package.json

```
  "scripts": {
    "dev": "echo \"test\""
  }
```

Run `bun --bun run dev` (387 ms)
Run `npm run dev` (225 ms)
Run `bun run dev` (59 ms)

The way I logged the runtime was:

```sh
start_ms=$(ruby -e 'puts (Time.now.to_f * 1000).to_i')
bun run dev      
end_ms=$(ruby -e 'puts (Time.now.to_f * 1000).to_i')
elapsed_ms=$((end_ms - start_ms))
echo "$elapsed_ms ms passed"
```

I didn't expect --bun to add this much runtime

### What do you see instead?

> 300 ms added

### Additional information

I didn't expect --bun to add this much runtime. I've seen indications ([link](https://github.com/withastro/astro/pull/7944#issuecomment-1666084805)) that --bun likely won't be needed from v1, what does that exactly mean? That it's added always under the hood, adding this overhead, or will the overhead go away then?

dylan-conway:
### What does this PR do?
fixes #4143 

- [ ] Code changes

tested manually

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -453,8 +453,6 @@ pub const RunCommand = struct {
                         continue;
                     };
                 }
-                if (comptime Environment.isPosix)
-                    _ = bun.C.chmod(bun_node_dir ++ "/node", 0o777);
                 break;
             }
         }


Jarred-Sumner:
We don’t need to chmod because it’s a symlink?

dylan-conway:
> We don’t need to chmod because it’s a symlink?

yep

Issue #4426, opened 2023-08-31
Audit every usage of the `undefined` keyword in our Zig code


Jarred-Sumner:
`undefined` is dangerous. `undefined` should only be used when the value is about to be entirely replaced by another value. When used to avoid the cost of zeroing buffers, we need to be very careful of bounds checks.

ghiscoding:
@robobun that is a lot of labels 🤣

robobun:
There is no Zig code left in the repository to audit. #30412 moved Bun to Rust, and #32621 removed the remaining `.zig` sources.

Closing this as obsolete. A concern about uninitialized memory in the Rust code belongs in a new issue against the current sources. Reopen this one if it still applies.

PR #4492, opened 2023-09-05
less undefined

CGQAQ:
### What does this PR do?

- [x] Code changes

This PR includes:
- any kind of ptr that is initialized to `undefined` changes to `foo: ?ptr = null`, as `.?` will panic in `safemode` instead of `UB`. Plus [an optional pointer is guaranteed to be the same size as a pointer. The null of the optional is guaranteed to be address 0](https://ziglang.org/documentation/master/#Optional-Pointers)
- any fields that don't need to init to undefined, `= undefined` is removed

`.?` should be optimized away in non-safe mode, so these changes should not impact performance, just make use of these nullable pointers more explicit

Closes: #4426


src/bun.zig
```diff
@@ -534,6 +534,15 @@ pub fn clone(val: anytype, allocator: std.mem.Allocator) !@TypeOf(val) {
...
+/// The undefind is unspecified value, cant be compared to anything include itself.
+/// In Debug mode,
+/// Zig writes 0xaa bytes to undefined memory.
+/// This is to catch bugs early, and to help detect use of undefined memory in a debugger.
+///   However, this behavior is only an implementation feature, not a language semantic,
+///   so it is not guaranteed to be *observable* to code.
+/// ref: https://ziglang.org/documentation/master/#undefined
 pub fn assertDefined(val: anytype) void {
     if (comptime !Environment.allow_assert) return;
     const Type = @TypeOf(val);
```
CGQAQ, line 548: We should use the optional type when we need this

CGQAQ:
@dylan-conway Please take a look. Sorry for so many changes in a single PR, but they're pretty straightforward.

CGQAQ:
image

image

oh, seems like I caught a bug already

Jarred-Sumner:
I think this is an interesting start but we shouldn't replace all usages of undefined with optional. We can do this more gradually. It'll be too difficult to confidently merge in this current state. When they're not pointers, marking things as optional can increase the size and alignment of structs which can in of itself cause new bugs

CGQAQ:
> I think this is an interesting start but we shouldn't replace all usages of undefined with optional. We can do this more gradually. It'll be too difficult to confidently merge in this current state. When they're not pointers, marking things as optional can increase the size and alignment of structs which can in of itself cause new bugs

This PR only changes *ptr fields in the struct that default initialized* to undefined to optional ptr, as [an optional pointer is guaranteed to be the same size as a pointer. The null of the optional is guaranteed to be address 0](https://ziglang.org/documentation/master/#Optional-Pointers), and only few non ptr type changes, so it should not cause any new bugs

PR #4493, opened 2023-09-05
Complete rework of the majority of `node:util`, primarily `util.inspect`

jhmaster2000:
### What does this PR do?

Removed the majority of the very buggy `node:util` browserify implementation (to the extent of functions of the module `util.inspect` depends on) in favor of a fresh implementation based on [node-inspect-extracted](https://github.com/hildjj/node-inspect-extracted) with Bun-specific patches and optimizations.

Extras:
* Fix `util.types.isArrayBuffer` to not return true for SharedArrayBuffers
* Add missing primordial functions to `node:tty`

This new implementation is still incomplete and has some remaining TODOs, but most are minimal and it's already many times better than the original.

- [x] Code changes

I adapted Node's automated tests for `node:util` into Bun, under the `node-inspect-tests` folder.


src/js/thirdparty/utl/primordials.js
```diff
@@ -0,0 +1,214 @@
...
+
+const copyProps = (src, dest) => {
+  Array.prototype.forEach.call(Reflect.ownKeys(src), key => {
+    if (!Reflect.getOwnPropertyDescriptor(dest, key)) {
+      Reflect.defineProperty(dest, key, Reflect.getOwnPropertyDescriptor(src, key));
+    }
+  });
+};
+
+const makeSafe = (unsafe, safe) => {
```
Jarred-Sumner: Let's not use any of this library 

We can figure out how to have JSC give us these functions instead of a JS-based primordials implementation

paperclover (approved):

paperclover:
Prettier CI failed not due to formatting but due to the inability to update the comment.

lenovouser:
I am not entirely sure, but I think this commit broke the `inspect()` function completely.

If you take a simple file like this and execute it:

```ts
import { inspect } from 'node:util';

console.log(inspect({ a: 1, b: '2', c: true }));
```

I get the following error:

```
27 | }, makeSafe = (unsafe, safe) => {
28 |   if (Symbol.iterator in unsafe.prototype) {
29 |     const dummy = new unsafe;
30 |     let next;
31 |
32 |     ArrayPrototypeForEach(Reflect.ownKeys(unsafe.prototype), (key) => {
                                                                                                                             ^
TypeError: true is not an Object.
      at internal:primordials:32:122
      at forEach (:1:20)
      at internal:primordials:28:28
      at internal:primordials:1:20
      at internal:util/inspect:1:20
      at node:util:1:20
```

I am using `v1.0.8` on ARM macOS

jhmaster2000:
@lenovouser I'm unable to reproduce on either release 1.0.8 nor current canary with WSL2, it prints out `{ a: 1, b: '2', c: true }` as normal for me.

Is that truly your entire file and which exact command are you using to run the file? Also, are you on one of the "special" builds like -profile or -baseline?

lenovouser:
@jhmaster2000 you are right.

I found the actual issue - this is caused by importing `core-js`, which was used to polyfilll things in node. The weird thing is this only happens if you import it in another file that then imports the file with inspect. I am not sure why it does break `inspect()` in the first place though, but in case it interests you here is the code:

**`test-a.ts`**

```ts
import 'core-js';

import './test-b';
```

**`test-b.ts`**

```ts
import { inspect } from 'node:util';

console.log(inspect({ a: 1, b: '2', c: true }));
```

You can then just run `bun test-a.ts`

jhmaster2000:
@lenovouser with that exact setup I still cannot reproduce your error on Bun 1.0.11, which admittedly came out 2 days ago so it's impossible to have been the one you tried on, could you retry with that Bun version now? If you still reproduce it then, what exact (not the version range) core-js version is installed?

lenovouser:
@jhmaster2000 I still do have this issue, I don't really know why, but now it happens only after the 2nd execution of the script and from then on forever.

Screenshot 2023-11-10 at 17 07 58

lenovouser:
**`system_profiler SPSoftwareDataType SPHardwareDataType`**
```
Software:

    System Software Overview:

      System Version: macOS 13.6.1 (22G313)
      Kernel Version: Darwin 22.6.0
      Boot Volume: Macintosh HD
      Boot Mode: Normal
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 1 day, 5 hours, 10 minutes

Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: Mac14,6
      System Firmware Version: 10151.41.12
      OS Loader Version: 8422.141.2.700.1
```

**`uname -a`**


```
Darwin Silas-MBP 22.6.0 Darwin Kernel Version 22.6.0: Wed Oct  4 21:26:55 PDT 2023; root:xnu-8796.141.3.701.17~4/RELEASE_ARM64_T6020 arm64
```

**`bun -v`**

```
1.0.11
```

Let me know whether there is anything else I can provide to help you out.

jhmaster2000:
Oh huh, I didn't expect that so I had only tested it once, indeed running it a second time now I can reproduce it, how odd 🤔 My first thought was some sort of weird core-js one-time initialization code somehow but reinstalling didn't seem to change anything, so that's probably not it, clearing Bun caches also didn't work, will investigate further.

lenovouser:
Yeah I tried it once, and was like "huh, it's gone, nice". Then I tried a 2nd time just to make sure and it appeared again 😄 No clue what causes this, probably a really weird and interesting bug.

jhmaster2000:
@lenovouser I've figured it out, its caused by the fact `import 'core-js';` tries to import *literally everything*, including unstable TC39 ES proposals, and it's adding some polyfills from this proposal: [tc39/proposal-collection-methods](https://github.com/tc39/proposal-collection-methods), which unexpectedly break some of `node:util`'s internals due to their addition violating certain assumptions about the Map prototype, in part because core-js' polyfills are returning incorrect values, but its not something we should rely on either way, so I'm making both a fix on Bun side to not rely on these faulty assumptions anymore and making a PR to core-js to fix their polyfills.

In the meantime though, you *really shouldn't* be using `import 'core-js';`, any of the safer/less-aggressive imports like `import 'core-js/actual';` resolves the problem for now, as they won't try to pull in unstable polyfills, see the [core-js README Usage section](https://github.com/zloirock/core-js#commonjs-api) for details on choosing the right import to use.

jhmaster2000:
@lenovouser I still recommend you read above and optimize your core-js import, but the `import 'core-js';` one will work now on the next Bun release (1.0.13, release TBA) or the earliest next canary shortly.

PR #4503, opened 2023-09-05
fix(fetch) always use readable stream if it is available

### LINKED ISSUE #4500: Fetch: Empty request on read stream

0.8.2+f73f77d0dec9006f5f42ff401fb32f9eaefa749a

1. Run server with node:
```
var http = require('http');
const { createReadStream } = require('fs');
const { pipeline } = require('stream')

const errorHandler = (err) => err && console.log(err.message);

var server = http.createServer(function (req, res) {
    res.writeHead(200, {'Content-Type': 'text/plain'});
    pipeline(createReadStream(__filename), res, errorHandler)
}).listen(8099);
```
2. Run client with bun:
```
async function newFetch() {
    const response = await fetch("http://localhost:8099", {verbose: true});
    const hasBody = response.body && true;
    if (hasBody) {
        let res = await response.text()
        console.log(res.length, response.status);
        return res;
    }
    throw new Error("No body");
}

async function waitError() {
    while (true) {
        let len = await newFetch();
        if(!len) {
            return;
        }
        //return;
        await new Promise(resolve => setTimeout(resolve, 50));
    }
}

waitError();
```

I expected the app to run indefinitely, similar to how Node.js does

### What do you see instead?

After some time zero length returned
```
Request: GET /
        Connection: keep-alive
        User-Agent: Bun/0.8.2
        Accept: */*
        Host: localhost:8099
        Accept-Encoding: gzip, deflate

Response: < 200 OK
<       Content-Type: text/plain
<       Date: Tue, 05 Sep 2023 13:13:51 GMT
<       Connection: keep-alive
<       Keep-Alive: timeout=5
<       Transfer-Encoding: chunked

458 200
Request: GET /
        Connection: keep-alive
        User-Agent: Bun/0.8.2
        Accept: */*
        Host: localhost:8099
        Accept-Encoding: gzip, deflate

Response: < 200 OK
<       Content-Type: text/plain
<       Date: Tue, 05 Sep 2023 13:13:51 GMT
<       Connection: keep-alive
<       Keep-Alive: timeout=5
<       Transfer-Encoding: chunked

0 200
```

### Additional information

I tried it on Ubuntu 20.04 (5.4.0-122-generic, 0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366), and it worked better initially. However, after some time, it breaks with a zero-length output. On WSL, it breaks after 2 requests, whereas on Linux, it lasts longer but eventually breaks as well. 
It worked as expected with Bun server and Bun client.

cirospaciari:
### What does this PR do?

Fixes #4500

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/webcore/response.zig
+++ b/src/bun.js/webcore/response.zig
@@ -792,6 +792,36 @@ pub const Fetch = struct {
                 return;
             }
 
+            if (this.readable_stream_ref.get()) |readable| {
+                if (readable.ptr == .Bytes) {
+                    readable.ptr.Bytes.size_hint = this.getSizeHint();
+                    // body can be marked as used but we still need to pipe the data
+                    var scheduled_response_buffer = this.scheduled_response_buffer.list;
+
+                    const chunk = scheduled_response_buffer.items;
+
+                    if (this.result.has_more) {
+                        readable.ptr.Bytes.onData(
+                            .{
+                                .temporary = bun.ByteList.initConst(chunk),
+                            },
+                            bun.default_allocator,
+                        );
+
+                        // clean for reuse later
+                        this.scheduled_response_buffer.reset();
+                    } else {
+                        readable.ptr.Bytes.onData(
+                            .{
+                                .temporary_and_done = bun.ByteList.initConst(chunk),
+                            },
+                            bun.default_allocator,
+                        );
+                    }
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you

PR #4514, opened 2023-09-06
feat(fetch) rejectUnauthorized and checkServerIdentity

cirospaciari:
### What does this PR do?

should reject any cert error (default: true)
```js
const result = await fetch("https://bun.sh", {
  tls: { rejectUnauthorized: true },
}).then((res) => res.text());
```

should NOT reject any cert error
```js
const result = await fetch("https://bun.sh", {
  tls: { rejectUnauthorized: false },
}).then((res) => res.text());
```

should be able to pass a custom `checkServerIdentity` to check self-sign certs
```js
const result = await fetch("https://mydomain.test", {
  tls: {
    checkServerIdentity: (hostname, cert) => {
      console.log(hostname, cert);
      if (hostname !== "mydomain.test")
        throw new Error("Only mydomain.test is allowed to have cert errors");
    },
  },
}).then((res) => res.text());

```

- [x] Code changes

I wrote automated tests + existing tests


src/bun.js/bindings/JSSinkLookupTable.h
```diff
@@ -119,11 +119,11 @@ static const struct HashTableValue JSFileSinkPrototypeTableValues[7] = {
 static const struct HashTable JSFileSinkPrototypeTable =
     { 7, 15, false, nullptr, JSFileSinkPrototypeTableValues, JSFileSinkPrototypeTableIndex };
 
+} // namespace JSC
 
+#include "Lookup.h"
```
Jarred-Sumner: something went wrong here, the imports should've been stripped

cirospaciari: weird, looks like `make codegen` did not complete (probably because I hit some key) runned again and its fixed https://github.com/oven-sh/bun/pull/4514/commits/a53872033a8233246360ccfef5c1ae022f74a6a4

src/http_client_async.zig
```diff
@@ -772,6 +864,184 @@ pub const HTTPThread = struct {
...
+    return outIP[0..size];
+}
+
+fn ipv4ToString(ipv4Bytes: *[4]u8, outIP: *[INET6_ADDRSTRLEN + 1]u8) ![]const u8 {
+    var ipv4String: [16]u8 = undefined;
+    return try canonicalizeIP(try std.fmt.bufPrint(&ipv4String, "{}.{}.{}.{}", .{ ipv4Bytes[0], ipv4Bytes[1], ipv4Bytes[2], ipv4Bytes[3] }), outIP);
+}
+
+const hexChars = "0123456789ABCDEF";
+fn ipv6ToString(ipv6Bytes: *[16]u8, outIP: *[INET6_ADDRSTRLEN + 1]u8) ![]const u8 {
```
Jarred-Sumner: i'm a little worried about edgecases that lead to security issues in this code

can we rely more on c-ares for this?

src/http_client_async.zig
```diff
@@ -772,6 +864,184 @@ pub const HTTPThread = struct {
...
+        if (BoringSSL.SSL_get_peer_cert_chain(ssl_ptr)) |cert_chain| {
+            if (BoringSSL.sk_X509_value(cert_chain, 0)) |x509| {
+
+                // check if we need to report the error (probably to `checkServerIdentity` was informed from JS side)
+                // this is the slow path
+                if (client.signals.get(.cert_errors)) {
+                    // clone the relevant data
+                    const cert_size = BoringSSL.i2d_X509(x509, null);
+                    var cert = bun.default_allocator.alloc(u8, @intCast(cert_size)) catch @panic("OOM");
+                    var buffer_ptr = @as([*c]u8, cert.ptr);
```
Jarred-Sumner: we should avoid creating any cases to `[*c]u8` in our code

src/http_client_async.zig
```diff
@@ -397,9 +413,85 @@ fn NewHTTPContext(comptime ssl: bool) type {
...
+                ptr: *anyopaque,
+                socket: HTTPSocket,
+                success: i32,
+                ssl_error: uws.us_bun_verify_error_t,
+            ) void {
+                const authorized = if (success == 1) true else false;
+
+                const handshake_error = HTTPCertError{
+                    .error_no = ssl_error.error_no,
+                    .code = if (ssl_error.code == null) "" else ssl_error.code[0..bun.len(ssl_error.code) :0],
```
Jarred-Sumner, line 445: is this memory static?

cirospaciari: this is inside BoringSSL ctx, but we cloned if we need outside `onCertError` because its the recomended

src/http_client_async.zig
cirospaciari: I will take a look and change the translated.zig instead of using this convertions.

src/http_client_async.zig
cirospaciari: sure need to check c-ares manual

src/js/node/tls.js
```diff
@@ -197,7 +197,7 @@ function checkServerIdentity(hostname, cert) {
     let error = new Error(`ERR_TLS_CERT_ALTNAME_INVALID: Hostname/IP does not match certificate's altnames: ${reason}`);
     error.name = "ERR_TLS_CERT_ALTNAME_INVALID";
     error.reason = reason;
-    error.host = host;
+    error.host = hostname;
```
cirospaciari, line 200: this is actually a bugfix

Jarred-Sumner (approved):
needs tests but otherwise LGTM

Jarred-Sumner:
test failures indicate something is wrong

src/http_client_async.zig
cirospaciari: fixed

src/http_client_async.zig
cirospaciari: fixed

Jarred-Sumner:
thank you

Issue #4516, opened 2023-09-06
The "bun ." call does not work in docker


doroved:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

**CMD ["bun", "."]** - doesn't want to work, error in the container logs in the screenshot.

```dockerfile
FROM oven/bun:latest
WORKDIR /app
COPY package.json bun.lockb ./
RUN bun i
COPY . .

CMD ["bun", "."]
```

![image](https://github.com/oven-sh/bun/assets/30116541/7cba52fa-1c0a-48d9-accf-964cc5aa516e)

**CMD [ "bun", "run", "./src/index.ts" ]** - this option works.

Hanaasagi:
I think this bug is related to how paths are handled when executing `bun .`. If change` WORKDIR /app` to `WORKDIR /app/test`, it does not throw this error.

```Dockerfile
FROM oven/bun:latest
WORKDIR /app/test
COPY package.json bun.lockb ./
RUN bun i
COPY . .

CMD ["bun", "."]

```

----
Update:


This bug is due to `PathName`, if the argument `_path` is `/app/`. The `dir` will be a empty string.


https://github.com/oven-sh/bun/blob/daaac7792cc348030d64a33087f1a41b3a3822cf/src/fs.zig#L1398-L1443

Then if we call `dirWithTrailingSlash`, we will get `./`.

https://github.com/oven-sh/bun/blob/daaac7792cc348030d64a33087f1a41b3a3822cf/src/fs.zig#L1385-L1396

PR #4523, opened 2023-09-06
Fix(fs): Fix the `dir` field parsing in `PathName`.

Hanaasagi:
### What does this PR do?

Close: #4516

If we passed a path like `/app/` into the `init` function of `PathName`, the parsed `dir` would be an empty string.

https://github.com/oven-sh/bun/blob/daaac7792cc348030d64a33087f1a41b3a3822cf/src/fs.zig#L1405-L1419

In above, we stop at first non-trailing slash, `i` could be `0` here, so `dir` is empty. 

https://github.com/oven-sh/bun/blob/daaac7792cc348030d64a33087f1a41b3a3822cf/src/fs.zig#L1385-L1396

`dirWithTrailingSlash` will return `./`.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/fs.zig
+++ b/src/fs.zig
@@ -1405,10 +1405,19 @@ pub const PathName = struct {
         var _i = strings.lastIndexOfChar(path, '/');
         while (_i) |i| {
             // Stop if we found a non-trailing slash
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
thanks for this

lets see what the tests say

PR #4528, opened 2023-09-07
`bun install` fix Git SSH URLs

### LINKED ISSUE #4068: bun install with Gitlab repository in dependencies not working

0.7.3

Linux 6.4.7-arch1-1 x86_64

1. `bun init`
2. Add Gitlab dependency like this, in `package.json`:
```
"dependencies": {
    "package": "git+ssh://git@gitlab.example.com:foobar.git"
}
```
3. `bun install`

Gitlab dependency should be installed without errors.

### What do you see instead?

```
bun install v0.7.3 (a9b3d583)
  🔍 Resolving [1/1] 
error: InstallFailed cloning repository for "<package-name>"
error: "git clone" for "<package-name>" failed
error: "git clone" for "<package-name>" failed
```

### Additional information

1. HTTPS protocol works, SSH doesn't.
`git+https://gitlab.example.com/foobar.git` ✅ 
`git+ssh://git@gitlab.example.com:foobar.git` ❌ 

2. `npm install` works just fine.

3. I would assume this might be an issue on Github as well, haven't tested though.

Parzival-3141:
### What does this PR do?

Closes #4068

`npm` allows SSH and SCP style URLs to have a `:` between the host and path, but `git` will not accept URLs like that. So we need to do a small conversion between the two.
E.g.  `git+ssh://git@example.com:path/to/repo` -> `git+ssh://git@example.com/path/to/repo`

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/install/dependency.zig
+++ b/src/install/dependency.zig
@@ -521,16 +521,6 @@ pub const Version = struct {
                         return .npm;
                     }
                 },
-                // ssh://user@example.com/user/repo
-                // ssh://user@example.com:user/repo
-                // ssh://example.com/user/repo
-                // ssh://example.com:user/repo
-                's' => {
-                    if (strings.hasPrefixComptime(dependency, "ssh://")) {
-                        return .git;
-                    }
-                },
-
                 // v1.2.3
                 // verilog
                 // verilog.tar.gz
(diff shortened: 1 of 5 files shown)


Parzival-3141:
`filesink.test.ts`,`postgres.test.ts`, and `prisma.test.ts` are the only tests that are blocking this PR. I think the rest are issues with `macOS-x64` builds in general.

h2oearth:
I'm looking forward to this PR: –). We can not use bun in some repos :-(, because of the `git+ssh://git@github.com` issue

timeturnback:
how was this going guys

Jarred-Sumner:
This was addressed some time ago

PR #4531, opened 2023-09-07
Example for fibonacci in TCO

imcotton:
### What does this PR do?

Two things come in my mind when talking about WebKit as web developer, the first being out-of-box TCO, thus adding example for fib 10_000.


Electroid:
Hi there, thanks for submitting this pull request. I think we already have enough examples in the repo.

PR #4538, opened 2023-09-07
fix(cli): final touches for 1.0

### LINKED ISSUE #4465: Reserve future command keywords
There are some future commands that Bun will introduce, but not for 1.0. To avoid breakage in the future, we will reverse these keywords.

For example, if `bun deploy` is a command and a package.json has a "deploy" script, you would need to run `bun run deploy` to avoid a clash with `bun deploy`.

List of npm keywords:
* config
* publish/unpublish
* prune
* version
* whoami
* login/logout

List of other keywords:
* user/account/org
* auth
* token
* info
* deploy
* cloud

paperclover:
Closes #4465

the pr is named bun_init but i actually gave up and stashed my bun init specific changes


src/cli.zig
```diff
@@ -1445,7 +1499,7 @@ pub const Command = struct {
                         var slice = std.mem.trim(u8, bun.asByteSlice(remainder[remainder_i]), " \t\n;");
                         if (slice.len > 0 and !strings.hasPrefixComptime(slice, "--")) {
                             if (positional_i == 0) {
-                                template_name_start = remainder_i;
+                                template_name_start = remainder_i + 2;
```
Jarred-Sumner, line 1502: why 2?

paperclover: earlier: `args[2..]`

PR #4541, opened 2023-09-07
Update Svelte Kit Docs

e253:
I added some guidance about how to build for production. Still WIP since I would like to add a more complete deployment guide.

### What does this PR do?

- [ ] Code changes

### DIFF
--- a/docs/guides/ecosystem/sveltekit.md
+++ b/docs/guides/ecosystem/sveltekit.md
@@ -63,3 +63,58 @@ Visit [http://localhost:5173](http://localhost:5173/) in a browser to see the te
 ---
 
 If you edit and save `src/routes/+page.svelte`, you should see your changes hot-reloaded in the browser.
+
+---
+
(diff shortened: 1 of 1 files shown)


happysalada:
hey just a question as an outsider, bun's aim to replace vite right ?
Right now vite is still needed but in the future if there is better integration, we should be able to remove it?
(I couldn't find the answer to this anywhere, so I'm asking here. No worries if you are busy though).

e253:
Not sure either. I think that bun *could* replace vite in sveltekit. At the moment, you can use bun to install deps and run the project in production.

colinhacks:
Thanks!

elvisisvan:
> hey just a question as an outsider, bun's aim to replace vite right ? Right now vite is still needed but in the future if there is better integration, we should be able to remove it? (I couldn't find the answer to this anywhere, so I'm asking here. No worries if you are busy though).

what? i thought bun is to replace deno?

PR #4544, opened 2023-09-07
Make the link to application clickable in the terminal

brablc:
### What does this PR do?

Terminals like iTerm require valid URL to make them clickable.  Adding a space to make link clickable in terminal. 

Similar problem is on https://bun.sh/, where protocol is missing.

Cosmetic change. Tested on macOS with iTerm.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -39,14 +39,14 @@ const server = Bun.serve({
   },
 });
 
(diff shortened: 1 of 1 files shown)


v3nntrnx:
LGTM

Electroid (approved):
Thanks!

PR #4585, opened 2023-09-08
Fix `Bun.CryptoHasher` missing argument in docs

samualtnorman:
### What does this PR do?

Add in the missing arguments for `new Bun.CryptoHasher()`. Running the examples in the docs before the fix gives you `TypeError: Expected an algorithm name as an argument`.

- [ ] Code changes

You can't have bugs when didn't write any code.


Electroid (approved):
Thanks!

PR #4600, opened 2023-09-08
Move "Importing CJS from CJS" back out of low-level details section

dfabulich:
### What does this PR do?

Move "Importing CJS from CJS" back out of low-level details section

This section appears to have been accidentally moved into the low-level details section in 5424ea3403df9cd4672290865f12b6f5b01cf2d0.

This fixes the example in the low-level details section, because the "Importing CJS from CJS" section is not the example code that the low-level details section is intended to provide.

- [ ] Code changes


Electroid (approved):
Thanks! nice find

PR #4623, opened 2023-09-08
`bunx` doesn't work with the install of version 1.0.0

dysbulic:
### What does this PR do?

Switches the `bunx` command to `bun x` because `bunx` isn't shipped with v1.0.0.

Ran it and the project was created.

### DIFF
--- a/docs/guides/ecosystem/sveltekit.md
+++ b/docs/guides/ecosystem/sveltekit.md
@@ -2,10 +2,10 @@
 name: Build an app with SvelteKit and Bun
 ---
 
-Use `bunx` to scaffold your app with the `create-svelte` CLI. Answer the prompts to select a template and set up your development environment.
+Use `bun x` to scaffold your app with the `create-svelte` CLI. Answer the prompts to select a template and set up your development environment.
 
 ```sh
-$ bunx create-svelte my-app
+$ bun x create-svelte my-app
 ┌  Welcome to SvelteKit!
 │
 ◇  Which Svelte app template?


ghost:
Isn't `bunx` is an alias for `bun x`?

hastebrot:
Checked it and `bunx` is installed in `~/.bun/bin/bunx` on macOS.

Is there maybe a different setup on windows, until support for it is feature-complete?

As far as I understand `bunx` is just a ~~copy~~ symbolic link of `bun` that detects that its filename is `bunx`.

dysbulic:
> Checked it and `bunx` is installed in `~/.bun/bin/bunx` on macOS.

@hastebrot, you did this with v1.0.0? I'm looking at [https://bun.sh/install](//bun.sh/install) & it is just pulling a zipfile from [releases](//github.com/oven-sh/bun/releases/) & expanding it. [bun-darwin-x64.zip](//github.com/oven-sh/bun/releases/download/bun-v1.0.0/bun-darwin-x64.zip) only contains the `bun` executable as does [bun-linux-x64.zip](//github.com/oven-sh/bun/releases/download/bun-v1.0.0/bun-linux-x64.zip) which is what's getting installed for me in Ubuntu.

hastebrot:
@dysbulic you're correct. The zip file does not contain the link for `bunx` and I also could not find the line that creates the link in the install script.

```
❯ zipinfo -1 bun-darwin-x64.zip
bun-darwin-x64/
bun-darwin-x64/bun

❯ xh GET https://bun.sh/install --body | grep ln | wc -l
0
```

strangely, when I reinstall it, `bunx` is created.

```
❯ rm -r ~/.bun
❯ curl -fsSL https://bun.sh/install | bash
❯ ls -1 ~/.bun/bin/
bun*
bunx@
```

dysbulic:
@hastebrot, that is bizarre. I know of no mechanism whereby the link would be created unless it's done from within the `bun` executable for some reason, which a perusal of the repo shows no evedince of.

In any case, I can tell you `bunx` isn't created in Ubuntu. Since `bunx`'s used so much in the documentation, an issue should likely be created.

Issue #4549, opened 2023-09-07
Issue: Sharp fails when installing with bun


gtrabanco:
### What version of Bun is running?

1.0.0+5b9d8b87c41410091a7c602b895ad9ffda17819c

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

## Install Sharp

```shell
$ mkdir my-project && cd $_
$ bun init
$ bun install
$ bun add --exact sharp
$ mv index.ts index.js
```

## Code

```js
import sharp from "sharp";

const shield = sharp("shield.png");
const { width, height } = await shield.metadata();

console.log(await shield.metadata());
```

### What is the expected behavior?

Output next content:

```shell
{
  format: "png",
  width: 415,
  height: 415,
  space: "srgb",
  channels: 4,
  depth: "uchar",
  density: 72,
  isProgressive: false,
  hasProfile: false,
  hasAlpha: true,
  orientation: 1,
  exif: Buffer(80) [ 77, 77, 0, 42, 0, 0, 0, 8, 0, 2, 1, 18, 0, 3, 0, 0, 0, 1, 0, 1, 0, 0, 135, 105, 0, 4, 0, 0, 0, 1, 0, 0, 0, 38, 0, 0, 0, 0, 0, 3, 160, 1, 0, 3, 0, 0, 0, 1, 0, 1, 0, 0, 160, 2, 0, 4, 0, 0, 0, 1, 0, 0, 1, 159, 160, 3, 0, 4, 0, 0, 0, 1, 0, 0, 1, 159, 0, 0, 0, 0 ],
[...]

PR #4629, opened 2023-09-08
Test trustedDependencies better

birkskyum:
trustedDependencies are only tested for a clean install without node_modules or lockfile. I built out the test suite to narrow down #4549. This PR extends the test suite, but doesn't contain a fix, so it's expected to fail.

I added this function as a test harness:

```zig
async function installPackages(shouldSaveLockfile: boolean, shouldRunLifeCycleScripts: boolean)
```

It allow me to test a story of deletion of bun.lockd and node_modules:

```ts
await installPackages(true, true); // 1. This line is all the previous tests covered
await installPackages(false, false); // 2. Reinstalling - don't save lockfile, don't run lifecycles
await rm(join(package_dir, "bun.lockb"));
await installPackages(true, true); // 3. No lockfile -> Save lockfile, Run lifecycles.
await rm(join(package_dir, "bun.lockb"));
await rm(join(package_dir, "node_modules"), { recursive: true });
await installPackages(true, true); // 4. Clean install, but where a cache exists
await rm(join(package_dir, "node_modules"), { recursive: true });
await installPackages(false, true); // 5. No node_modules -> Should run lifecycles
```

When I use bun i have to delete BOTH bun.lockb and node_modules to run life cycles (similar to above install 3. and 5.), but when I run these tests only the install 5 is breaking which is surprising to me, but it's still an improvement.

### DIFF
--- a/test/cli/install/bun-install.test.ts
+++ b/test/cli/install/bun-install.test.ts
@@ -5117,39 +5117,67 @@ it("should handle trustedDependencies", async () => {
   });
   await writeFile(join(package_dir, "moo", "package.json"), moo_package);
   await writeFile(join(package_dir, "moo", "echo.js"), "console.log(`moo|${process.argv[2]}|${import.meta.dir}`);");
-  const { stdout, stderr, exited } = spawn({
-    cmd: [bunExe(), "install"],
-    cwd: package_dir,
-    stdout: null,
-    stdin: "pipe",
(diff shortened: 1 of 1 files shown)


birkskyum:
Think the trustedDependencies in a saved lockfile aren't loaded correctly from disk.

birkskyum:
- #5472

PR #4647, opened 2023-09-09
fix(docs): Fix the text that `bun run --bun` is the same as `bun`

DuGlaser:
### What does this PR do?

Hello, I've just started using bun!
I was a bit puzzled by the different behavior between `bun` and `bun run` when dealing with files that have a shebang (the documentation states that the two commands should behave identically).
When I use the `--bun` option, the behavior becomes the same as expected. I think this might be a better approach. What do you think? (I've just started using bun, so I could be wrong).
P.S. This message was machine-translated, so I apologize if it's hard to understand.

- [ ] Code changes

### DIFF
--- a/docs/cli/run.md
+++ b/docs/cli/run.md
@@ -26,7 +26,7 @@ $ bun run index.ts
 $ bun run index.tsx
 ```
 
-The "naked" `bun` command is equivalent to `bun run`.
+The "naked" `bun` command is equivalent to `bun run --bun`.
 
 ```bash
 $ bun index.tsx
--- a/docs/cli/run.md
+++ b/docs/cli/run.md
@@ -26,10 +26,11 @@ $ bun run index.ts
 $ bun run index.tsx
 ```
 
-The "naked" `bun` command is equivalent to `bun run --bun`.
+Alternatively, you can omit the `run` keyword and use the "naked" command; it behaves identically.
 
 ```bash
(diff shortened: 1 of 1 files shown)


jakeboone02:
Per the documentation, `bun` and `bun run` _are_ equivalent when running `package.json` scripts.

The `--bun` flag is useful for the package runner `bunx`. The `--bun` flag forces Bun to use its own runtime instead of Node when a script's shebang specifies Node (`#!/usr/bin/env node`). Relevant docs here: https://bun.sh/docs/cli/bunx#shebangs.

(Important note: `--bun` must come before the executable name.)

DuGlaser:
Yes, the documentation states that `bun` and `bun run` should behave the same. However, they behave differently when dealing with files that have a shebang.

For example, consider the following:

```
$ cat index.js
#!/usr/bin/env node

console.log('Hello, world!')

$ bun run index.js
error: missing script "index.js"

$ bun index.js
Hello, world!
```

Because of this, I thought it might be more accurate to write that `bun` and `bun --bun run` behave the same, rather than saying bun and bun run are identical.

jakeboone02:
Oh I see. I think this is a bug rather than a documentation issue. I filed #4850 to track.

colinhacks:
Definitely a bug, so we'll leave the doc unchanged (more or less). I took this opportunity to improve the wording a bit.

Issue #4509, opened 2023-09-05
fs.promises.stat doesn't throw ENOENT with empty string


coratgerl:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

fs.promises.stat should return ENOENT with path equal to "".

**This code with bun :**
```javascript
import { promises } from 'fs'

const test = async () => {
	console.log(await promises.stat(''))
}

test()
```
![image](https://github.com/oven-sh/bun/assets/73360179/4e5742da-bee9-4ade-964b-fdf5fb844cfd)

**Same code with node :**
<img width="468" alt="image" src="https://github.com/oven-sh/bun/assets/73360179/ea3cd9d9-6a35-4f4b-bb13-863368db85dd">

### What is the expected behavior?

Return ENOENT with empty string

### Additional information

This bug block the use of the library mongodb-memory-server usefull to test a lot of applications.

birkskyum:
same with latest version - 0.8.2+bc2b55fdeee3001b07252d0f250671ea1876a3ed

jdmarshall:
There are a bunch of other places where bun doesn't issue ENOENT where nodejs does.

I've already seen 4 PRs for github projects that are 'fixing' the module so it works with Bun. How many modules touch the filesystem? It must be thousands. 

I just convinced someone to reject a PR on the basis that BunJS is wrong, not the node module, on the premise that I/We don't want to see 10,000 PRs to 'correct' file access sanity checking code across the entire Javascript and Typescript ecosystem. 

https://github.com/node-config/node-config/pull/744

How about we don't invite a proliferation of logic that looks like 

```
if (e2.code !== 'ENOENT' && e2.code !== 'EBADF') {
```

by externalizing compatibility issues onto a 10+ year old ecosystem, full of projects that haven't accepted a PR in 3 years?

I can see from other bug reports and PRs here that Bun is capable of issuing ENOENT, it just doesn't do it in the same places Node does. I don't think that's a feature. I think it's an unnecessary compatibility issue, and perhaps speaks to Bun needing E2E tests that consist entirely of running unit and integration tests for known OSS projects, and ensuring they are green on Bun.

When you break compatibility it should be for something important, like architectural improvements, not error codes.

Electroid:
[...]

PR #4693, opened 2023-09-09
fix: node compatibility with empty path string (stat, exist, ...)

coratgerl:
### What does this PR do?

Closes: #4509 

This PR fix the node compatibility with the path string which need to accept empty string. Then the syscall return the good result (ENOENT for `stat`, false for `exist`, etc...). See the posix stat documentation (https://man7.org/linux/man-pages/man2/fstat.2.html or the extract in the screen below):

image

- [x] Code changes

With tests 


coratgerl:
Fix https://github.com/oven-sh/bun/issues/4509

Jarred-Sumner:
Thank you

PR #4712, opened 2023-09-09
Fix incorrect remote port in net shim

### LINKED ISSUE #4660: `node:net` client sockets remoteAddress property return the local address instead of the remote address

1.0

Linux x86_64

Bun returns the local server address instead of the remote address.

```
import net from 'net'

const server = net.createServer((socket) => {
    console.log('New connection from ' + socket.remoteAddress + ':' + socket.remotePort);
});

server.listen(9292, '0.0.0.0', () => {
    console.log('TCP server listening on 0.0.0.0:9292');
});
```

```
$ bun main.js 
TCP server listening on 0.0.0.0:9292
New connection from <server address>:9292
```

```
$ node main.js 
TCP server listening on 0.0.0.0:9292
New connection from <client address>:37862
```

The connecting client's address

### What do you see instead?

The local server address

### Additional information

_No response_

SuperAuguste:
### What does this PR do?

Closes #4660

- [x] Code changes

Using repro code in the aforementioned issue.

Before:
Screenshot 2023-09-09 at 1 34 06 PM

After:
Screenshot 2023-09-09 at 1 32 57 PM


src/bun.js/api/bun/socket.zig
```diff
@@ -1677,6 +1677,17 @@ fn NewSocket(comptime ssl: bool) type {
             return ZigString.init(text).toValueGC(globalThis);
         }
 
+        pub fn getRemotePort(
+            this: *This,
+            _: *JSC.JSGlobalObject,
+        ) callconv(.C) JSValue {
+            if (this.detached) {
+                return JSValue.jsUndefined();
+            }
```
gruckion, line 1686: What happens if I don't define a port?

src/bun.js/bindings/generated_classes.zig
```diff
@@ -6327,6 +6327,9 @@ pub const JSTCPSocket = struct {
         if (@TypeOf(TCPSocket.getRemoteAddress) != GetterType)
             @compileLog("Expected TCPSocket.getRemoteAddress to be a getter");
 
+        if (@TypeOf(TCPSocket.getRemotePort) != GetterType)
+            @compileLog("Expected TCPSocket.getRemotePort to be a getter");
+
```
gruckion, line 6332: Do the other conditions have test coverage? If so likely will need it for this too.

Jarred-Sumner: This is generated code

gruckion:
Nice work 👍

src/bun.js/api/bun/socket.zig
SuperAuguste: Hi. As far as I know this is a non-issue as this API is not directly exposed to end users and I only invoke it in a place where it guaranteed to exist, and, as you can see from that guard clause, it will return undefined if the socket is detached for some reason. :)

paperclover:
looks good

could you write a test for this? probably good enough to just check the port is not the local port.

also `remotePort` is technically exposed in the public api; can you add typedefs in `bun.d.ts`?

PR #4756, opened 2023-09-10
docs/run: Reference reserved keywords

tordans:
Reference reserved keywords to clarify that bun run is not always the same as bun run <file>

Ideally, this reference would link to a list of reserved keywords. Is there such a list in the docs? I could only find https://github.com/oven-sh/bun/issues/4465.


v3nntrnx (commented):
There is a typo in "execpt".
Other than that, LGTM

Issue #4789, opened 2023-09-10
`node:path.relative` edge case


Hanaasagi:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 6.4.12-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const path = require("node:path");
console.log(path.relative("/app/node_modules/pkg", "../static"));
```

### What is the expected behavior?

```
>>​ node i.cjs
../../../static
```

### What do you see instead?

```
>>​ bun run i.cjs
../../static

```

`app` is my process cwd.

Hanaasagi:
testcases
```JavaScript
const assert = require("assert");
const path = require("path");

const cwd = process.cwd();
const cwdParent = path.dirname(cwd);
const parentIsRoot = cwdParent === "/";

const cases = [
  [
    "/app/node_modules/pkg",
    "../static",
    `../../..${parentIsRoot ? "" : cwdParent}/static`,
  ],
  [
    "/app/node_modules/pkg",
    "../../static",
    `../../..${parentIsRoot ? "" : path.dirname(cwdParent)}/static`,
  ],
  ["/app", "../static", `..${parentIsRoot ? "" : cwdParent}/static`],
  ["/app", "../".repeat(64) + "static", "../static"],
  [".", "../static", "../static"],
  ["/", "../static", parentIsRoot ? "static" : `${cwdParent}/static`.slice(1)],
];

cases.forEach(([from, to, expected]) => {
  console.log("test", from, to);
  const result = path.relative(from, to);
  assert.equal(result, expected);
});

console.log("All test cases passed!");

```

Jarred-Sumner:
Can we port all of node's tests for this? So that it stops being a problem going forward

Hanaasagi:
@Jarred-Sumner 

I compared the test cases for `path.relative` in node and bun, and I found that Bun currently has more test cases than node. 🤦 Maybe we need a fuzz test.

node.js: https://github.com/nodejs/node/blob/afea87ee14cd47e70049cbd741ffccb8223f95ca/test/parallel/test-path-relative.js#L4
bun: https://github.com/oven-sh/bun/blob/afcbed218c425872648b7c0aa2ce331ad3b0d612/test/js/node/path/path.test.js#L458

ottomated:
This will close #4616 :)

PR #4811, opened 2023-09-10
fix(path): Fix edge case in `path.relative`

Hanaasagi:
### What does this PR do?

Close: #4789

In Bun 0.1, this code will log an empty string, but node will log `..`.

```JavaScript
const path = require("path");

console.log(path.relative("../", "../../"));
```

This is because the parameter `allow_above_root` we passed is `false`.

https://github.com/oven-sh/bun/blob/d48ff53e4ec4cb49172e0e96ca9890446f971a0e/src/resolver/resolve_path.zig#L479-L490

In `normalizeStringGeneric` function, if `allow_above_root` is false, `../` will be converted to empty string. 

https://github.com/oven-sh/bun/blob/d48ff53e4ec4cb49172e0e96ca9890446f971a0e/src/resolver/resolve_path.zig#L535-L544

This also affects paths like `../static`, which will be converted into `static`. So in Bun, the result for the following code is `true`. Actually, it's comparing `static` and `static`.

```JavaScript
console.log(path.relative("../static", "../../static") == "");
```

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/resolver/resolve_path.zig
+++ b/src/resolver/resolve_path.zig
@@ -471,7 +471,7 @@ pub fn relativePlatform(from: []const u8, to: []const u8, comptime platform: Pla
         Fs.FileSystem.instance.top_level_dir,
         &relative_from_buf,
         &[_][]const u8{
-            normalizeStringBuf(from, relative_from_buf[1..], false, platform, true),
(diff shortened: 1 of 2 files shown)


dylan-conway (approved):
LGTM

PR #4858, opened 2023-09-10
Remove the case-insensitive bit from the docs.

farcaller:
Bun actually throws FileNotFound.

### What does this PR do?

This fixes the docs oversight, because they don't match the behavior of bun:

```sh
$ echo 'export default "hello, world";' > hello.js
$ echo -e 'import val from "./hello";\nconsole.log(val);' > index.ts
$ bun index.ts
hello, world
$ echo -e 'import val from "./HELLO";\nconsole.log(val);' > index.ts
$ bun index.ts

error: FileNotFound reading "/root/HELLO.js"

$ bun --version
1.0.0
```

- [ ] Code changes

I ran it against the current release.


Jarred-Sumner:
This is unintentionally platform-specific. We have code for warning against incorrectly-cased imports

farcaller:
@Jarred-Sumner should I open a bug instead? It's just that what the code does and what the docs say is distinctly opposite.

I suppose it will be platform-specific because the code seems to rely on the filesystem being case-insensitive. Which isn't much of a bun feature at that point. But maybe I'm overlooking something?

PR #4870, opened 2023-09-10
Fix punctuation

matthewyu01:
### What does this PR do?

- [ ] Code changes

it's => its
lets => let's
were => we're


Electroid (approved):
Thanks!

PR #4875, opened 2023-09-11
Fix docs for workspace dependencies

bndkt:
Referencing other packages in a monorepo requires the workspace protocol (workspace:*).

### What does this PR do?

- [ ] Code changes


braaar:
This change should be made in the [docs](https://github.com/oven-sh/bun/blob/main/docs/install/workspaces.md) as well, not just the guide

PR #4881, opened 2023-09-11
Update simple.md

TwanLuttik:
I have removed the / before ${server.port} because it its incorrect and I replaced the localhost to ${server.hostname}


colinhacks:
Thanks!

PR #4882, opened 2023-09-11
declare global in watch-mode/HTTP Servers

### LINKED ISSUE #4884: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.

Example code is not working

### What is the issue?

I get `Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.` error when using example from `Runtime/Watch Mode/HTTP Servers`

### Where did you find it?

https://bun.sh/docs/runtime/hot#http-servers

FreePhoenix888:
### What does this PR do?

declare global because typescript can show error `Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.`
It fixes #4884

- [ ] Code changes

### DIFF
--- a/docs/runtime/hot.md
+++ b/docs/runtime/hot.md
@@ -104,6 +104,10 @@ Bun provides the following simplified API for implementing HTTP servers. Refer t
 ```ts#server.ts
 import {serve} from "bun";
 
+declare global {
+  var count: number | undefined;
+}
+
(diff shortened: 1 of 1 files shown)


FreePhoenix888:
Hi, @Jarred-Sumner. Why have you closed it?

PR #4907, opened 2023-09-11
Fix missing script error when running empty file

JeremyFunk:
### What does this PR do?

- [x] Remove a line that would break out of the block handling code execution

- [x] I ran `zig fmt` on the changed files
- [x] Automated tests


test/cli/run/empty-file.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { join } from "path";
+
+it("should execute empty scripts", () => {
+  const { stdout, exitCode } = Bun.spawnSync({
+    cmd: [bunExe(), "run", "--bun", join(import.meta.dir, "empty-file.js")],
+    env: bunEnv,
+    stderr: "inherit",
+  });
+
+  expect(stdout.toString().trim().endsWith("--pass--")).toBe(true);
```
Parzival-3141, line 12: You should be checking if `stderr` is empty here instead. Use `"pipe"` instead of `"inherit"`.

Parzival-3141 (changes requested):

paperclover:
Thanks for this fix! It will be merged with corrected tests in #5025

PR #4910, opened 2023-09-11
Make detect-libc package match the upstream api

fehnomenal:
### What does this PR do?

The version API of `detect-libc` was implemented in reverse (see https://www.npmjs.com/package/detect-libc). This causes the installation of sharp to fail.

- [x] Code changes


Electroid (approved):
Thanks!

Issue #2137, opened 2023-02-23
Alias `bun uninstall` to `bun remove`


Electroid:
`bun uninstall` = `bun remove`

Electroid:
Done in https://github.com/oven-sh/bun/commit/9024125d41751895496048e30e3fdc30019ef4ad

plutoniumm:
hey it seems this can go one step more & we'll have full compatibility with npm
`bun un` can also be same as `npm un` with just another alias

PR #4917, opened 2023-09-11
Add 'un' to package remove

plutoniumm:
### What does this PR do?
Adds an alias for `bun un` for `bun remove` to achieve full compatibility with the npm API, this is an extension of #2137, since it was a minor change i did not consider opening an issue and have dropped a comment on #2137

- [x] Code changes

Compiled it locally and created tried to run it as needed

```sh
un_test) cat package.json | jq
# {
#   "dependencies": {
#     "vite": "^4.4.9"
#   }
# }
```

Seemed to run ok on macOS with no issues
```sh
un_test) ../packages/debug-bun-darwin-aarch64/bun-debug un vite
[SYS] read(3, 4096) = 4096 (0.062ms)
[SYS] close(3)
bun remove v1.0.0_debug (589d236e)
[16.00ms] done
No packages! Deleted empty lockfile
```

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1158,6 +1158,7 @@ pub const Command = struct {
             RootCommandMatcher.case("r"),
             RootCommandMatcher.case("remove"),
             RootCommandMatcher.case("rm"),
+            RootCommandMatcher.case("un"),
             RootCommandMatcher.case("uninstall"),
             => .RemoveCommand,
 


src/cli.zig
```diff
@@ -1158,6 +1158,7 @@ pub const Command = struct {
             RootCommandMatcher.case("r"),
             RootCommandMatcher.case("remove"),
             RootCommandMatcher.case("rm"),
+            RootCommandMatcher.case("un"),
```
staabm, line 1161: Is "npm un" really a separate command or does npm just autocomplete the "un" to "uninstall" since no other command with the same prefix exists?

If it autocompletes, it would mean also "npm uni" or "npm unin" etc. are supported?

plutoniumm: hey its a seperate command marked as a proper alias [npm docs](https://docs.npmjs.com/cli/v7/commands/npm-uninstall), it works the same in both npm (& pnpm, the equivalent drop in replacement which does symlinking)

jbergstroem: > npm uninstall [<@scope>/]<pkg>...
> aliases: unlink, remove, rm, r, un

..perhaps add `unlink` while at it?

plutoniumm: I think bun unlink already exists that’s why I didn’t want to touch it

that may have broken a few things

colinhacks:
I'm not sure how I feel about supporting all of npm's crazy aliases. It's not currently our goal to be compatible with every `npm` command, and `bun un` isn't very aesthetic imo. It's also a bit ambiguous with `bun unlink`.

Screenshot 2023-09-13 at 10 13 09 PM

plutoniumm:
> I'm not sure how I feel about supporting all of npm's crazy aliases. It's not currently our goal to be compatible with every `npm` command, and `bun un` isn't very aesthetic imo. It's also a bit ambiguous with `bun unlink`

this is fully fair game, the reason to why I thought 'un' should exist as opposed to all the other aliases is that its fairly commonly used in both npm and pnpm ecosystems , It is also worth nothing that yarn does not have 'un'

ps i can see how its unaesthetic :p, i just went for it being convenient to type

Electroid:
Going to close, reasoning same as Colin's.

PR #4926, opened 2023-09-11
Naked `bun start` rather than `bun run start` in docs home

diogo405:
### What does this PR do?

- [ ] Code changes


Electroid:
While both are correct, this is more about visibility, so people know `bun run` is a top-level command.

PR #4975, opened 2023-09-11
Simplify getting Set of extentions

Connormiha:
### What does this PR do?
I removed needless array wrapper. `new Set` already has array as argument. No need to create additional wrapper.

- [x] Code changes

Yes


Electroid:
Thanks!

PR #4987, opened 2023-09-11
Update quickstart.md

pablotrianda:
Remove req to avoid warning interpreter for: `req is declared but its value never read.` message

### What does this PR do?

- [ ] Code changes


colinhacks:
Thanks for the contribution, though I think it's worth keeping just as an indication of how the API works

PR #4997, opened 2023-09-11
Update simple.md

tomredman:
Remove errant slash preventing the correct console log

Old:
image

Fixed:
image

### What does this PR do?

Fixes the websocket quick start guide output from `Listening on localhost:${server.port}` to `Listening on localhost:3000`

- [ ] Code changes


Electroid (approved):
Nice catch, thanks for making this better.

PR #5002, opened 2023-09-11
Adding proper bash quote escaping on install.sh

lmmfranco:
### What does this PR do?

This PR adds proper quote escaping to the "bash" section of the install script when recommending manual change to the bash profile file.

Current output:
```
bun was installed successfully to ~/.bun/bin/bun 

Manually add the directory to ~/.bash_profile (or similar):
  export BUN_INSTALL="$HOME/.bun" 
  export PATH=$BUN_INSTALL/bin:$PATH 

To get started, run: 

  bun --help 
``` 

(You can see the second statement is not quoted as the previous one)

- [x] Code changes

- Manually ran the install script and verified that the output was correct


lmmfranco:
As for the reasoning for the quotes, it prevents issues when paths have spaces.
Eg:
```sh
$ export FOO=aa bb cc dd
$ echo $FOO
aa
```

paperclover (approved):

PR #5061, opened 2023-09-12
file.exists() needs to be awaited to get the value

amt8u:
### Updated code for file exists() function

- [ ] Code changes


Jarred-Sumner:
Thank you

PR #5069, opened 2023-09-12
update dev build instruction for arch

mi4uu:
### What does this PR do?

clang and llvm on arch install v16, update to use v15 and update env.

- [x] Documentation for building bun
- [ ] Code changes

make setup was failing, now it produces working packages/debug-bun-linux-x64/bun-debug


Electroid (approved):
Thanks for making this better!

PR #5070, opened 2023-09-12
Bun.file replaced with file

bagherani:
Because the `file` has already been imported, there is no need to use `Bun.file`

### DIFF
--- a/docs/api/http.md
+++ b/docs/api/http.md
@@ -216,7 +216,7 @@ import { serve, file } from "bun";
 
 serve({
   fetch(req) {
-    return new Response(Bun.file("./hello.txt"));
+    return new Response(file("./hello.txt"));
   },
 });
 ```


docs/api/http.md
```diff
@@ -216,7 +216,7 @@ import { serve, file } from "bun";
 
 serve({
   fetch(req) {
-    return new Response(Bun.file("./hello.txt"));
+    return new Response(file("./hello.txt"));
```
mquandalle, line 219: I find it clearer to consistently keep `Bun.serve` and `Bun.file` in the documentation

paperclover: @colinhacks what do you think? I agree with that we should make all the examples on here use the full `Bun.serve`/`Bun.file`

colinhacks:
We've updated the docs to use the `Bun` global basically everywhere - let's stick with that. 

Good catch @bagherani

PR #5104, opened 2023-09-12
[bun install] Add `-E` as alias of `--exact`

jonahsnider:
### What does this PR do?

`bun install` and `bun add` both support the `--exact` flag for pinning a dependency's version.
This adds a shorter `-E` flag as an alias, something that [npm](https://docs.npmjs.com/cli/v10/commands/npm-install#save-exact:~:text=additional%2C%20optional%20flags%3A-,%2DE%2C%20%2D%2Dsave%2Dexact,-%3A%20Saved%20dependencies%20will), [Yarn](https://yarnpkg.com/cli/add#options-E%2C-exact), and [pnpm](https://pnpm.io/cli/add#--save-exact--e) also support.

  - There are a few examples of how to use `--exact` where it may be helpful to have examples for `-E` as well
  - I think it's fine to leave them as-is though
- [x] Code changes

I've tested locally and confirmed that the flag works as expected:

screenshot of `bun install zod -E` working locally

There are existing tests for `--exact`. If it'd be useful, those could be duplicated to do the same assertions but with `-E` instead of `--exact. The only thing that'd test would be the CLI arg parsing though, not sure if that's useful.

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -5783,7 +5783,7 @@ pub const PackageManager = struct {
         clap.parseParam("-d, --dev                 Add dependency to \"devDependencies\"") catch unreachable,
         clap.parseParam("-D, --development") catch unreachable,
         clap.parseParam("--optional                        Add dependency to \"optionalDependencies\"") catch unreachable,
-        clap.parseParam("--exact                      Add the exact version instead of the ^range") catch unreachable,
+        clap.parseParam("-E, --exact                  Add the exact version instead of the ^range") catch unreachable,
         clap.parseParam("<POS> ...                         ") catch unreachable,
     };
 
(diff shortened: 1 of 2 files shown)


Electroid:
Thanks for adding this! Mind also adding a test, you can copy-paste from this:
https://github.com/oven-sh/bun/blob/92e95c86dd100f167fb4cf8da1db202b5211d2c1/test/cli/install/bun-add.test.ts#L304

jonahsnider:
Added a test for `-E`

BrennanColberg:
strongly support this, my default `[manager] install -DE` works on npm/pnpm/yarn but not here

v3nntrnx:
LGTM

Electroid (approved):

PR #5109, opened 2023-09-12
feat(nodejs): implement `os.availableParallelism`

### LINKED ISSUE #4986: node:os availableParallelism() not implemented

Darwin 22.5.0 arm64 arm

The [`availableParallelism()`](https://nodejs.org/api/os.html#osavailableparallelism) method seems to be missing from the node:os module.

#### test.js
```javascript
import { availableParallelism } from 'node:os';

console.log(availableParallelism());
```

`node test.js` -> "8"

`bun test.js` -> "SyntaxError: Import named 'availableParallelism' not found in module 'os'."

I checked the [compatibility overview page ](https://bun.sh/docs/runtime/nodejs-apis) and node:os is marked as fully implemented, so I guess this is a bug.

The function should return the available number of threads/parallelism of the system.

### What do you see instead?

SyntaxError: Import named 'availableParallelism' not found in module 'os'.

### Additional information

_No response_

WingLim:
### What does this PR do?

Resolves #4986

- [x] Code changes

- [x] Add test in [os.test.js](https://github.com/WingLim/bun/commit/2b92bcb7e743bb2b501ce85c7802254552ba2d3a#diff-d6b42a80df08c1f8ca95728d1b2166de3388a2c8a0d3085a53177a1c9e7f6154)


ghost:
`std.Thread.getCpuCount`?

src/bun.js/node/node_os.zig
```diff
@@ -48,6 +49,24 @@ pub const Os = struct {
...
+            if (rc < 1) {
+                return JSC.JSValue.jsNumberFromUint64(1);
+            }
+            return JSC.JSValue.jsNumberFromUint64(rc);
+        } else |_| {
+            const err = JSC.SystemError{
+                .message = bun.String.static("Failed to get availableParallelism information"),
+                .code = bun.String.static(@as(string, @tagName(JSC.Node.ErrorCode.ERR_SYSTEM_ERROR))),
+            };
+            globalThis.vm().throwError(globalThis, err.toErrorInstance(globalThis));
```
paperclover: I don't think we should throw here. I couldn't find code in node.js that allows this to throw.

paperclover (changes requested):

src/bun.js/node/node_os.zig
WingLim: You are right, deleted.

Jarred-Sumner:
This can be done in JS. `navigator.hardwareConcurrency` exposes this information.

Jarred-Sumner:
Thank you

PR #5126, opened 2023-09-12
fix(node/fetch): Make data URL fetch consistent with node

### LINKED ISSUE #5073: Fetch data URL text decoding works erratically depending on the presence of certain characters

Darwin 22.6.0 arm64 arm

```
> fetch("data:application/json,%EF%BB%BF").then(r => r.text())  // ok
Promise { '' }
> fetch("data:application/json,%EF%BB%BF%7B%7D").then(r => r.text())  // ok
Promise { '{}' }
> fetch("data:application/json,%EF%BB%BFhello").then(r => r.text()) // ok
Promise { 'hello' }
> fetch("data:application/json,%EF%BB%BF{}").then(r => r.text())  // weird
Promise { '%EF%BB%BF{}' }
> fetch("data:application/json,{%7B%7D").then(r => r.text()) // weird
Promise { '{%7B%7D' }
```

The "weird" results should probably decode the URL-escaped sequences.

At least `|`, `{`, `}`, `#` in the URL trigger this behavior; `,` and `/` don't. I didn't check exhaustively.

### Additional information

Found via #4876.

davidmhewitt:
### What does this PR do?
Fixes #5073 

Implements a non-strict mode of the percent encoding decoder that instead of immediately throwing an error and using the original data when there is non-URL safe character in a data URL, it passes the non-URL safe characters through and only replaces the percent encoded elements.

`test.mjs`:
```
console.log(await fetch("data:application/json,{%7B%7D}").then((r) => r.text()));
console.log(await fetch("data:application/json,{%%7B%7D%%}%%").then((r) => r.text()));
console.log(await fetch("data:,%20%😀").then((r) => r.text()));
```

Node output:
```
{{}}
{%{}%%}%%
 %😀
```

Bun before this commit:
```
{%7B%7D}
{%%7B%7D%%}%%
%20%😀
```

After this commit, the output from Bun matches Node.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/resolver/data_url.zig
+++ b/src/resolver/data_url.zig
@@ -39,18 +39,24 @@ pub const PercentEncoding = struct {
         };
     }
 
-    /// decode path if it is percent encoded
+    /// decode path if it is percent encoded, returns EncodeError if URL unsafe characters are present and not percent encoded
     pub fn decode(allocator: Allocator, path: []const u8) EncodeError!?[]u8 {
+        return _decode(allocator, path, true);
+    }
+
+    /// Replaces percent encoded entities within `path` without throwing an error if other URL unsafe characters are present
+    pub fn decodeUnstrict(allocator: Allocator, path: []const u8) EncodeError!?[]u8 {
+        return _decode(allocator, path, false);
+    }
+
+    fn _decode(allocator: Allocator, path: []const u8, strict: bool) EncodeError!?[]u8 {
         var ret: ?[]u8 = null;
         errdefer if (ret) |some| allocator.free(some);
         var ret_index: usize = 0;
         var i: usize = 0;
 
         while (i < path.len) : (i += 1) {
-            if (path[i] == '%') {
-                if (!isPchar(path[i..])) {
-                    return error.InvalidCharacter;
-                }
+            if (path[i] == '%' and path[i..].len >= 3 and isHex(path[i + 1]) and isHex(path[i + 2])) {
                 if (ret == null) {
                     ret = try allocator.alloc(u8, path.len);
                     bun.copy(u8, ret.?, path[0..i]);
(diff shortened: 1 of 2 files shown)


src/resolver/data_url.zig
```diff
@@ -39,18 +39,24 @@ pub const PercentEncoding = struct {
...
     pub fn decode(allocator: Allocator, path: []const u8) EncodeError!?[]u8 {
+        return _decode(allocator, path, true);
+    }
+
+    /// Replaces percent encoded entities within `path` without throwing an error if other URL unsafe characters are present
+    pub fn decodeUnstrict(allocator: Allocator, path: []const u8) EncodeError!?[]u8 {
+        return _decode(allocator, path, false);
+    }
+
+    fn _decode(allocator: Allocator, path: []const u8, strict: bool) EncodeError!?[]u8 {
```
Jarred-Sumner, line 52: we can make `strict` a `comptime` argument and then it maybe would produce better code 

but its fine to keep it also

Jarred-Sumner (approved):
LGTM. Thank you for this. Let's merge after tests run

akx:
Thanks for picking up my issue 👍

davidmhewitt:
Looking at the GitHub actions failures here, it looks like it may be a permissions thing with my account (as an outside contributor), rather than actual test failures. It seems to be the action being unable to comment on the PR with results that caused the overall failure.

It looks like the failed tests for linux-x64 were:
  - `test/js/node/dns/node-dns.test.js`
  - `test/js/third_party/postgres/postgres.test.ts`
  - `test/js/third_party/prisma/prisma.test.ts`

Unsure of the cause for the DNS test failure, but I imagine the other two are related to my GitHub account again not having permission to access the env variables for the Database secrets for those tests.

PR #5127, opened 2023-09-12
Updated Lambda readme

tsndr:
### What does this PR do?

- [ ] Code changes

### DIFF
--- a/packages/bun-lambda/README.md
+++ b/packages/bun-lambda/README.md
@@ -8,7 +8,7 @@ First, you will need to deploy the layer to your AWS account. Clone this reposit
 
 ```sh
 git clone git@github.com:oven-sh/bun.git
-cd packages/bun-lambda
+cd bun/packages/bun-lambda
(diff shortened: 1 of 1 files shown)


packages/bun-lambda/README.md
```diff
@@ -8,7 +8,7 @@ First, you will need to deploy the layer to your AWS account. Clone this reposit
 
 ```sh
 git clone git@github.com:oven-sh/bun.git
-cd packages/bun-lambda
+cd bun/packages/bun-lambda
```
xhyrom, line 11: why?

tsndr: because `git clone` will create a `bun` folder and that way you could just copy and paste the whole block to your terminal.

xhyrom: i didn'ŧ mention that, that's true

Electroid (approved):
Thanks!

PR #5143, opened 2023-09-12
[Docs] Use git's `--global` flag for lockfile diffs instead of manually editing config files.

Southpaw1496:
### What does this PR do?

This PR changes the method for enabling Bun lockfile diffs in git. Instead of manually modifying config files, the new method uses the git CLI's `--global` flag to change the global `.gitattributes` file. This should be easier, safer, and more cross-platform than manually modifying the config files. I also clarified that the "local" git-attributes file refers to a per-repository setting, while the global `.gitattributes` file refers to a system-wide setting.

- [ ] Code changes


Jarred-Sumner:
Thank you

Issue #4843, opened 2023-09-10
Bun Lambda Layer does not properly handle API Gateway V1 events


tekkeon:
### What steps can reproduce the bug?

1. Follow the steps in the Lambda Layer README to publish the layer and create a Lambda function.
2. Connect the Lambda Function with the Bun Lambda Layer to a REST API Gateway (V1).
3. Test the endpoint.

### What is the expected behavior?

The Lambda function should properly handle the request.

### What do you see instead?

Errors in the Lambda Layer trying to construct the request.

### Additional information

I've found these specific issues with the Lambda Layer `runtime.ts` code that, when changed, fixes the issues with V1 events:
1. `isHttpEventV1` assumes the event has a field called `version` and a value of `1.0`. In fact, only V2 events have a `version` with value `2.0` while V1 events simply omit the version field altogether.
2. `formatHttpEventV1` inserts headers to a Headers instance both from the `event.headers` field and the `event.multiValueHeaders` and then uses the Headers instance to get the Host and X-Forwarded-Proto header. The problem is that API gateway actually duplicates these values across headers and multiValueHeaders, probably as a convenience thing, so when you retrieve Host from Headers, you end up with something like `example.org, example.org` when you just want `example.org`.
3. If Host header is not defined, then `response.text()` is returned. I'm not sure why exactly this is, but it breaks the testing experience in API Gateway unless you explicitly include a Host header.

~~Also, while trying to debug all of this, I noticed that trying to `console.log` multiple values (e.g. `console.log('Message', object)`) results in only logging the first value. I see that the `runtime.ts` puts a wrapper around `console.log` and I'm guessing that `Bun.inspect` is perhaps the culprit here.~~ Looks like there is a PR fixing this issue here: https://github.com/oven-sh/bun/pull/4781

tekkeon:
I'd be happy to take this issue and can probably get a fix out in the next couple days, but wanted to have the issue open before submitting a PR.

tekkeon:
I see this PR: https://github.com/oven-sh/bun/pull/4781 already fixes the last comment I had about console.log only printing the first argument, so I'll just focus on the other four issues.

Electroid:
Fixed in #5161, thanks @mkossoris! 🎉

PR #5161, opened 2023-09-12
fix(bun-lambda) Fix API Gateway V1 events and expand on Lambda documentation

tekkeon:
### What does this PR do?

As described in this issue I opened (https://github.com/oven-sh/bun/issues/4843), the Bun Lambda Layer does not currently work properly with API Gateway V1 (REST API) events. There were a few small bugs that contributed to this:
> 1. `isHttpEventV1` assumes the event has a field called `version` and a value of `1.0`. In fact, only V2 events have a `version` with value `2.0` while V1 events simply omit the version field altogether.
> 2. `formatHttpEventV1` inserts headers to a Headers instance both from the `event.headers` field and the `event.multiValueHeaders` and then uses the Headers instance to get the Host and X-Forwarded-Proto header. The problem is that API gateway actually duplicates these values across headers and multiValueHeaders, probably as a convenience thing, so when you retrieve Host from Headers, you end up with something like `example.org, example.org` when you just want `example.org`.
> 3. If Host header is not defined, then `response.text()` is returned. I'm not sure why exactly this is, but it breaks the testing experience in API Gateway unless you explicitly include a Host header.

Regarding `isHttpEventV1` and `isHttpEventV2`, I followed the same logic used by the popular `serverless-express` library that they use for determining the incoming event type, hopefully hardening these two event identification helpers in case someone wants to use Bun with a different AWS service than API Gateway: https://github.dev/vendia/serverless-express/tree/mainline/src.

Regarding the Host header and `response.text()`, it was unclear exactly what they were trying to do, but from what I could tell, it seemed like the code was meant to return regular text when the event was not an API Gateway event. That fixed the issue where testing in the API Gateway console did not work or using the API from a non-browser environment was not guaranteed to work.

In addition to these bug fixes, the Lambda Layer README.md instructions were lacking with no specific configuration instructions, so I've reorganized some of the documentation and added more instruction to help users more quickly get acquainted with using Bun on Lambda.

- [x] Code changes

No existing test setup was included for the Lambda Layer, so I didn't add that myself as it seemed out of scope for a bug fix. I validated that it works through manual testing. Specifically, I deployed the Lambda Layer, created a Lambda function that uses the published Layer, attached it to a V1 REST API, and validated it did not work. Then, I updated the Lambda layer with the changes in this PR and validated that it does work, including printing the incoming request and body to ensure the values were as expected.


paul-uz:
Thank you this is super useful information and is exactly what I was looking for!

Electroid (approved):
Thank you, this looks great!

paul-uz:
@mkossoris looks like the documentation is a bit wonky.

>  This allows you to also run your Lambda function as a local Bun server with `bun run <handler-name>.ts`

This does not work at all.

Create the following file, `index.ts`

```typescript
export default {
  async fetch() {
    console.log('hello world')
  },
};
```

Now run it via `bun run index.ts` and nothing happens.

PR #5167, opened 2023-09-12
decode regex if needed

### LINKED ISSUE #5132: replace string with regular expression got wrong output

Darwin 22.5.0 arm64 arm

when try to match a regular expression with these emoji and replace it with string
```
let str = "🔴11 54 / 10,000";
str = str.replace(/[🔵🔴,]+/g, '');
console.log(str);
```

the outputs should be:
```11 54 / 10000```
instead of
```🔴  / 0000```

### What do you see instead?

_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by dylan-conway**
### What does this PR do?
fixes #5132 

- [ ] Code changes

added test and tested manually.

### DIFF
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -2538,7 +2538,10 @@ pub const E = struct {
     };
 
     pub const RegExp = struct {
-        value: string,
+        data: union(enum) {
+            raw: string,
+            decoded: std.ArrayList(u16),
+        },
 
         // This exists for JavaScript bindings
         // The RegExp constructor expects flags as a second argument.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/js_ast.zig [hunk 1]
```diff
@@ -2538,7 +2538,10 @@ pub const E = struct {
     };
 
     pub const RegExp = struct {
-        value: string,
+        data: union(enum) {
+            raw: string,
+            decoded: std.ArrayList(u16),
```
Jarred-Sumner: `bun.BabyList(u16)` so we don't increase the size much

PR #5200, opened 2023-09-13
fix(node:dns): fix crash and compatibility issues with `resolve`

Hanaasagi:
#### 1. long string could lead to a segfault.

Segmentfault code: 

```JavaScript
const dns = require("node:dns");

dns.resolveCname("a".repeat(3000), (err, results) => {});
```

`null` should not be used here as it can lead to a crash. (Line 530)

https://github.com/oven-sh/bun/blob/3b2c0941e4b418dcf2bfbc5117da76338f45a90e/src/deps/c_ares.zig#L526-L541

#### 2. Fix a compatibility issue (`resolveNs` and `resolveSoa`)

```JavaScript
const dns = require("node:dns");
for (const f of [dns.resolveNs, dns.resolveSoa]) {
  f("", (err, results) => {
    console.log(f, results);
  });
}

// nodejs output
// [Function: bound queryNs] [
//   'h.root-servers.net',
//   'j.root-servers.net',
//   'k.root-servers.net',
//   'd.root-servers.net',
//   'b.root-servers.net',
//   'f.root-servers.net',
//   'l.root-servers.net',
//   'e.root-servers.net',
//   'i.root-servers.net',
//   'a.root-servers.net',
//   'm.root-servers.net',
//   'g.root-servers.net',
//   'c.root-servers.net'
// ]
// [Function: bound querySoa] {
//   nsname: 'a.root-servers.net',
//   hostmaster: 'nstld.verisign-grs.com',
//   serial: 2023091202,
//   refresh: 1800,
//   retry: 900,
//   expire: 604800,
//   minttl: 86400
// }
```

bun doesn't allow empty strings to be passed into these two functions.

---

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/deps/c_ares.zig
+++ b/src/deps/c_ares.zig
@@ -524,11 +524,12 @@ pub const Channel = opaque {
     }
 
     pub fn resolve(this: *Channel, name: []const u8, comptime lookup_name: []const u8, comptime Type: type, ctx: *Type, comptime cares_type: type, comptime callback: cares_type.Callback(Type)) void {
+        if (name.len >= 1023 or (name.len == 0 and !(bun.strings.eqlComptime(lookup_name, "ns") or bun.strings.eqlComptime(lookup_name, "soa")))) {
+            return cares_type.callbackWrapper(lookup_name, Type, callback).?(ctx, ARES_EBADNAME, 0, null, 0);
+        }
+
         var name_buf: [1024]u8 = undefined;
-        const name_ptr: ?[*:0]const u8 = brk: {
-            if (name.len == 0 or name.len >= 1023) {
-                break :brk null;
-            }
+        const name_ptr: [*:0]const u8 = brk: {
             const len = @min(name.len, name_buf.len - 1);
             @memcpy(name_buf[0..len], name[0..len]);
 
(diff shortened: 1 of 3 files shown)


src/deps/c_ares.zig
```diff
@@ -524,11 +524,12 @@ pub const Channel = opaque {
     }
 
     pub fn resolve(this: *Channel, name: []const u8, comptime lookup_name: []const u8, comptime Type: type, ctx: *Type, comptime cares_type: type, comptime callback: cares_type.Callback(Type)) void {
+        if (name.len >= 1023 or (name.len == 0 and !(bun.strings.eqlComptime(lookup_name, "ns") or bun.strings.eqlComptime(lookup_name, "soa")))) {
+            return cares_type.callbackWrapper(lookup_name, Type, callback).?(ctx, ARES_EBADNAME, 0, null, 0);
```
Hanaasagi, line 528: > ARES_EBADNAME The query name name could not be encoded as a domain name, either because it contained a zero-length label or because it contained a label of more than 63 characters. 

Ref: https://c-ares.org/ares_query.html

src/deps/c_ares.zig
```diff
@@ -524,11 +524,12 @@ pub const Channel = opaque {
...
+        }
+
         var name_buf: [1024]u8 = undefined;
```
Hanaasagi, line 531: maybe a smaller size here

Jarred-Sumner:
Thank you

PR #5225, opened 2023-09-13
modules documentation didn't have correct import example

miccou:
### What does this PR do?

- [ ] Code changes

I noticed that the [modules documentation](https://bun.sh/docs/runtime/modules) doesn't include import example code in the imports section. 

image

This PR replaces the screenshotted code block with:
```ts
import { foo } from "./foo"; // extensions are optional
import bar from "./bar.ts";
import { stuff } from "./my-commonjs.cjs";
```


colinhacks:
Oops. Thanks!

PR #5233, opened 2023-09-13
fix(runtime): require cache should not include unevaluated ESM modules.

### LINKED ISSUE #4593: Implement process.binding("stream_wrap")

1.0.0

Linux 6.4.12-arch1-1 x86_64 unknown

Clone this repo

https://github.com/zone-eu/zone-mta-template

Install dependencies and run

bun index.js

_No response_

### What do you see instead?

TypeError: process.binding("stream_wrap") is not implemented in Bun. If that breaks something, please file an issue and include a reproducible code sample.
      at binding (:1:20)
      at /run/media/zone-mta-template/node_modules/handle-thing/lib/handle.js:17:11
      at globalThis (/run/media/zone-mta-template/node_modules/handle-thing/lib/handle.js:262:9)
      at require (:1:20)
      at /run/media/zone-mta-template/node_modules/spdy/lib/spdy/handle.js:4:4
      at globalThis (/run/media/zone-mta-template/node_modules/spdy/lib/spdy/handle.js:246:3)
      at require (:1:20)
      at /run/media/zone-mta-template/node_modules/spdy/lib/spdy.js:6:5

### Additional information

_No response_

### LINKED ISSUE #4880: require.cache value undefined

Darwin 22.5.0 arm64 arm

```
git clone git@github.com:opennetwork/logistics.git ./opennetwork-logistics
cd ./opennetwork-logistics
git checkout 300bb0e
yarn # bun install, bcrypt_lib.node seems to not work on macOS, but not relevant, bun install might work for you
node example.js # no error
THROW_ON_BUG=1 node example.js # no error
bun example.js # no error
THROW_ON_BUG=1 bun example.js # this throws an error
``` 

TLDR import a bunch of modules (key count on `require.cache` is 1776 in this case), one or two of them might end up with `undefined` values with bun

For me I see `sharp` and `extract` showing up as `undefined` values:

```
sharp undefined
extract undefined
```

I do see sharp itself, the full file name for `sharp/lib/index.js` does come through. 

```
node_modules/sharp/lib/index.js has exports
```

That if a key is in `require.cache` that it is a value, and not undefined

https://github.com/nodejs/node/blob/b3fc9173ed0a562e71ea9bb2d719c27a92857c76/lib/internal/modules/cjs/loader.js#L931 

https://github.com/nodejs/node/blob/b3fc9173ed0a562e71ea9bb2d719c27a92857c76/lib/internal/modules/cjs/loader.js#L921C18-L921C30 

If an error is thrown, [`delete Module._cache[filename];`](https://github.com/nodejs/node/blob/b3fc9173ed0a562e71ea9bb2d719c27a92857c76/lib/internal/modules/cjs/loader.js#L942C7-L942C38) is used  

For node the values do not come through as `undefined` 

### What do you see instead?

Some values for `require.cache` _might_ be `undefined` for bun

### Additional information

I found this here https://github.com/opennetwork/logistics/actions/runs/6141617299/job/16662152259#step:11:398 

When looking into it, this line from fastify expects `require.cache[key]` to always be a `Module`

https://github.com/fastify/fastify/blob/eaf7b7b7442ef81c370d6714859626aa141977c5/lib/pluginUtils.js#L34 

Local on my mac this had ran, but maybe the difference in env I have setup for CI, something triggers this `getPluginName` path. 

Fastify could be updated to use `require.cache[key]?.exports` here, but I would expect any prior usage of `require.cache` assumed the quirks of node and have been implemented like this.

paperclover:
### What does this PR do?

Fixes #4593
Fixes #4880

Fastify does this thing where they iterate over `require.cache` during plugin registration

```ts
  // let's see if this is a file, and in that case use that
  // this is common for plugins
  const cache = require.cache
  // cache is undefined inside SEA
  if (cache) {
    const keys = Object.keys(cache)

    for (let i = 0; i < keys.length; i++) {
      const key = keys[i]
      if (cache[key].exports === func) { // ERROR HERE: undefined is not an object
        return key
      }
    }
  }
```

Not sure how it gets inserted into the ESM registry, but `msgpack-extract` somehow does this; it's an unevaluated module.

I don't think unevaluated modules should appear in the require cache at all, period. This PR filters those out.

- [X] Code changes

I wrote automated tests


Jarred-Sumner (approved):

BlusteryS:
image
Still not working with `bun build` and `bun run`

PR #5234, opened 2023-09-13
Correct the configuration file names.

nathanhammond:
Bun stores configuration in these three places:

- `$XDG_CONFIG_HOME/.bunfig.toml`
- `$HOME/.bunfig.toml`
- `${pwd}/bunfig.toml`

The use of `bun.toml` is a documentation error.


Jarred-Sumner:
Thanks. We were originally going to switch to `bun.json` before the release and then didn't

PR #5248, opened 2023-09-13
Add informative message on `bun create react`

colinhacks:
### What does this PR do?

Print informative message on `bun create react`. This command is in a lot of tutorials with good SEO so we need a fix here.

```bash
$ bun-debug create react
The "react" template has been deprecated.
It is recommended to use "react-app" or "vite" instead.

To create a project using Create React App, run

  bun create react-app

To create a React project using Vite, run

  bun create vite

Then select "React" from the list of frameworks.
```

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1510,6 +1510,28 @@ pub const Command = struct {
 
                 const template_name = positionals[0];
 
+                // if template_name is "react"
+                // print message telling user to use "bun create vite" instead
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
LGTM

jt3k:
I believe that the react app should not be popularized because it is not intended for anything other than quickly running code examples, creating one-off prototypes and training. On the react website it was previously offered for installation, but now they offer to run vitе

PR #5281, opened 2023-09-13
chore(docs): include missing links to Node.js APIs

styfle:
### What does this PR do?

This PR adds some missing links to make it easy to jump to another section.


Electroid (approved):
Thanks!

PR #5285, opened 2023-09-13
doc(guides): update sveltekit guide 

mroyme:
### What does this PR do?

Use `bun create svelte` instead of `bunx create-svelte`. 

This is to bring the documentation at par with the SvelteKit official doc to [create a project](https://kit.svelte.dev/docs/creating-a-project).

- [ ] Code changes


colinhacks:
Thanks!

PR #5294, opened 2023-09-13
docs(guide): fix expect assertion example in guide for `spyOn`

winghouchan:
### What does this PR do?

Fixes example with `spyOn` and assertions. The example failed because the spied function would be called once but the expectation asserted 0 calls.

- [ ] Code changes

1. Create test file with example.
2. Run `bun test`.

Example test should now pass.


Electroid (approved):
Nice catch, thanks!

PR #5311, opened 2023-09-13
add uninstall instructions

browner12:
It's always good to have uninstall instructions as well as install, and I couldn't find them anywhere else on the site.

IMO, this gives users a little more confidence to try new tools out, as they know it's easily reversible.

I'm not familiar with how to uninstall with Docker, so would appreciate some help there.

We could also possibly add a note for "bash" on how to remove Bun from the `$PATH`.

### What does this PR do?

- [ ] Code changes

NA

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -149,3 +149,27 @@ To write the completions to a custom location:
 $ bun completions > path-to-file      # write to file
 $ bun completions /path/to/directory  # write into directory
 ```
+
+## Uninstall
+
+If you need to remove Bun from your system, use the following commands.
+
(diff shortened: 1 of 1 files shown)


Electroid (approved):
Thanks!

warcayac:
You should update your instruction to [uninstall](https://bun.sh/docs/installation#uninstall) **bun** on Linux, because it is located at `~/.reflex` directory, so I suppose I need to delete that directory, moreover there are changes to do in `~/.bashrc` or `~/.zshrc` but they are not mentioned.

Issue #4893, opened 2023-09-11
Unterminated string literal 


Btaralte:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

```
 const SMS = {
     TOKEN: "[TOKEN]",
     TEMPLATES: {
         FORGOT_PASSWORD: "Your forgot password link!\
		\nLINK\
		\n \
		\nRegards, \n ",
     }
 }; 
```
Due to this code bun throws this errror.

```

error: Unterminated string literal

         FORGOT_PASSWORD: "Your forgot password link!\
```
nodejs is able to run this.

Jarred-Sumner:
It looks like not all the text was copy pasted correctly because the snippet provided in the github issue runs successfully on my machine

Btaralte:
Yes.Sorry for not copying the whole file.I also just run this snippet and it run succesfully,but what i didn't understand is when i comment out this part ,the code run succesfully.Btw this is the full code

```
 const STATUS_MSG = {
     ERROR: {
         DATA_NOT_FOUND: {
             statusCode: 401,
             type: "DATA_NOT_FOUND",
             message: "Result not data"
         },
         TOKEN_ALREADY_EXPIRED: {
             statusCode: 401,
             message: "Your login session expired!",
             type: "TOKEN_ALREADY_EXPIRED"
         },
         DB_ERROR: {
             statusCode: 400,
             message: "DB Error : ",
             type: "DB_ERROR"
         },
         INVALID_ID: {
             statusCode: 400,
             message: "Invalid Id Provided : ",
             type: "INVALID_ID"
         },
         APP_ERROR: {
             statusCode: 400,
             message: "Application Error",
             type: "APP_ERROR"
         },
         ADDRESS_NOT_FOUND: {
             statusCode: 400,
             message: "Address not found",
             type: "ADDRESS_NOT_FOUND"
         },
         APP_VERSION_ERROR: {
             statusCode: 400,
             message: "One of the latest version or updated version value must be present",
             type: "APP_VERSION_ERROR"
         },
         INVALID_TOKEN: {
             statusCode: 401,
             message: "Invalid token provided",
             type: "INVALID_TOKEN"
         },
         LINK_USED: {
             statusCode: 401,
             message: "Link is already used once.",
             type: "LINK_USED"
         },
[...]

PR #5318, opened 2023-09-13
Fix bug with multiline string in CRLF terminated files (#4893)

tikotzky:
### What does this PR do?

correctly handles escaped CRLF line endings in multiline strings

fixes https://github.com/oven-sh/bun/issues/4893

- [x] Code changes

I manually tested against three files
- terminated with CRLF
- terminated with CR
- terminated with LF

Pre change it errored on CRLF and passed on the other 2
Post change it passes on all 3

I also added an automated test for the failing CRLF scenario.

### DIFF
--- a/src/js_lexer.zig
+++ b/src/js_lexer.zig
@@ -642,7 +642,7 @@ fn NewLexer_(
                         lexer.step();
 
                         // Handle Windows CRLF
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (approved):

Jarred-Sumner:
Thank you

tikotzky:
Thank you for the quick review and merge!

Issue #5319, opened 2023-09-13
Bun cannot handle a large stdout


codymikol:
### What version of Bun is running?

1.0.1

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Found this while building an auditing report tool that prints our a large JSON blob, but you can reproduce this very easily in bun with this script that prints out an "=" n times

This will be fine

```javascript
import {exec} from "child_process";

exec("printf '=%.0s' {1..262144}")
```

If we increase n by one, bun will crash

```javascript
import {exec} from "child_process";

exec("printf '=%.0s' {1..262145}")
```

Probably some kind of memory allocation problem 🤷 

### What is the expected behavior?

It just returns the stdout

### What do you see instead?

```
TypeError: second argument to Function.prototype.apply must be an Array-like object
      at onChildStdoutEncoded (node:child_process:122:96)
      at addChunk (node:stream:1940:42)
      at readableAddChunk (node:stream:1894:58)
      at node:stream:13:342
      at node:stream:101:102
      at processTicksAndRejections (:1:2602)
```

### Additional information

Thanks for all your hard work 🤠

codymikol:
~~Has something to do with this event handler that's registered when buffers are supposed to have infinite size~~

```javascript
      child.stdout.on(
          "data",
/
          function onUnlimitedSizeBufferedData(chunk) {
            ArrayPrototypePush.call(_stdout, chunk);
          },
      );
```

codymikol:
`ArrayPrototypePush.call(_stdout, StringPrototypeSlice.apply(chunk, 0, truncatedLen));`
 
 truncatedLen here is actually NaN

codymikol:
I actually don't think that is the issue either, because it seems to be NaN in both cases where the string is too long / not too long. 

Latest suspicion is that this is busted

StringPrototypeSlice.apply(chunk, 0, truncatedLen)

I can't seem to get the result of that to log and it would explain the function failing.

codymikol:
I got it, PR coming soon

codymikol:
Kinda lost on these tests, bun test doesn't seem to be evaluating my changes in child_process, is there a more correct way to run these tests?

Aviortheking:
Hey, any news on this ? I have the same problem currently

codymikol:
It's fixed in my PR, whenever they have time to merge it should be resolved

codymikol:
@Electroid is this good to merge?

[...]

PR #5324, opened 2023-09-14
fix(child_process): handle stdout exceeds max buffer size

codymikol:
* resolve an issue where the calculated encodedStdoutLen would always be resolved to undefined due to never being initialized

* resolve an improper usage of 'apply' on StringPrototypeSlice where the arguments passed to apply are required to be passed as an array rather than varargs.

Fixes: #5319

### What does this PR do?

When stdout is too large, bun will crash due to both an uninitialized value "encodedStdoutLen" which always results in NaN and also an improper usage of `apply` on StringPrototypeSlice which would always result in an exception being thrown.

- [x] Code changes

The failing test case I provided in the attached issue now works as expected.


codymikol:
Not sure about committing the transpiled changes, is that supposed to be the InternalModuleRegistryConstants.h ? I don't see any other transpiled files.

dylan-conway:
> Not sure about committing the transpiled changes, is that supposed to be the InternalModuleRegistryConstants.h ? I don't see any other transpiled files.

Yes, for this change `InternalModuleRegistryConstants.h` would need to be committed

codymikol:
✅

dylan-conway:
could you also add a test for this in `child_process-node.test.js`?

codymikol:
doing it now actually

codymikol:
Kinda lost on these tests, bun test doesn't seem to be evaluating my changes in child_process, is there a more correct way to run these tests?

codymikol:
scratch that, I'm a dummy

codymikol:
I feel pretty good about the tests now, this should be ready to review

codymikol:
rebased and formatted

Electroid:
Oops, I accidentally broke this while testing, will fixup.

codymikol:
Are you still interested in this fix?

Electroid:
If you could force push up your changes that would be great, we can adapt it. My bad for messing up the diff.

codymikol:
:heavy_check_mark:

codymikol:
its cool, don't sweat it

PR #5333, opened 2023-09-14
Remove `--save` from helptext

colinhacks:
### What does this PR do?

Currently `--save` is listed as a flag for `bun install|add|remove` even though it's a no-op.

This changes it to only appear in `link` and `unlink` where it does something

The existence of `--save` in the helptext also implied that Bun wouldn't update package.json by default which causes confusion

Per https://github.com/oven-sh/bun/pull/3277 there is a bug where Bun will stop parsing when it encounters an unrecognized long flag, so `bun install --save zod` will behave like `bun install`. This is true for all manner of unimplemented `npm` flags, and adding `--save` to support this use case isn't worth the confusion. Flags like `--save-dev` or more common anyway. 

Later we should refactor the args parsing to its easier to ignore unrecognized long flags and have more control over helptext, etc.

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -5763,7 +5763,6 @@ pub const PackageManager = struct {
         clap.parseParam("-y, --yarn                        Write a yarn.lock file (yarn v1)") catch unreachable,
         clap.parseParam("-p, --production                  Don't install devDependencies") catch unreachable,
         clap.parseParam("--no-save                         Don't save a lockfile") catch unreachable,
-        clap.parseParam("--save                            Save to package.json") catch unreachable,
         clap.parseParam("--dry-run                         Don't install anything") catch unreachable,
         clap.parseParam("--lockfile <PATH>                  Store & load a lockfile at a specific filepath") catch unreachable,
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
The existence of `--save` adds the existence of `--no-save` and `--no-save` should work

colinhacks:
> The existence of --save adds the existence of --no-save and --no-save should work

Not totally clear what you're advocating but this doesn't affect `--no-save`. Also `--save` and `--no-save` are totally unrelated.  `--save` relates to updating package.json whereas `--no-save` relates to bun.lockb

ghiscoding:
> > The existence of --save adds the existence of --no-save and --no-save should work
> 
> Not totally clear what you're advocating but this doesn't affect `--no-save`. Also `--save` and `--no-save` are totally unrelated. `--save` relates to updating package.json whereas `--no-save` relates to bun.lockb

@colinhacks I think that he meant that typically in a CLI, if you pass the `--no-` prefix, it will explicitly assign the xyz option to `false`, for example see this [yargs boolean(key)](https://yargs.js.org/docs/#api-reference-argv) docs. So in this case you could pass `--save: true | false` or `--no-save` which is the equivalent of `save: false`

PR #5335, opened 2023-09-14
Remove the ability to configure lockfile location.

### LINKED ISSUE #5259: install.lockfile.path is unvalidated
Passing in absolute paths (or constructed paths) to `install.lockfile.path` or `install.lockfile.savePath` in `.bunfig.toml` can create cross-install contention.

These configuration values should likely be constrained to:
- Non-traversing.
- Unix separators.
- And be lexically cleaned.

Either that, or they can simply be removed. It's unnecessary complexity that no other package manager implements.

***

1. Set up a global config with `path` set absolutely.
```
~$ cat $HOME/.bunfig.toml
[install.lockfile]
path = "/Users/nathanhammond/Downloads/global.lockb"%
```

2. First `bun` project.
```
~$ cd repos/triage/test-bun 
~/repos/triage/test-bun$ cat package.json
{
  "dependencies": {
    "left-pad": "1.3.0"
  }
}%
```

3. See what happened.
```
~/repos/triage/test-bun$ bun install
bun install v1.0.0 (822a00c4)
 + left-pad@1.3.0

 1 packages installed [8.00ms]
```

4. Human readable lockfile:
```
~/repos/triage/test-bun$ bun /Users/nathanhammond/Downloads/global.lockb
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1
# bun ./bun.lockb --hash: 1741E594B1196BC9-6aa7802de12757c0-9DB326781358C6E7-66300595bc92c8b7


left-pad@1.3.0:
  version "1.3.0"
  resolved "https://registry.npmjs.org/left-pad/-/left-pad-1.3.0.tgz"
  integrity sha512-XI5MPzVNApjAyhQzphX8BkmKsKUxD4LdyK24iZeQGinBN9yTQT3bFlCBy/aVx2HrNcqQGsdot8ghrjyrvMCoEA==
```

5. A second project with the same global config!
```
~/repos/triage/test-bun$ cd ../other-bun
~/repos/triage/other-bun$ cat package.json
{
  "dependencies": {
    "lodash": "4.17.21"
  }
}%
~/repos/triage/other-bun$ bun install
bun install v1.0.0 (822a00c4)

 1 packages removed [8.00ms]
```

6. And now the lockfile is for the _other_ project.
```
~/repos/triage/other-bun$ bun /Users/nathanhammond/Downloads/global.lockb
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1
# bun ./bun.lockb --hash: B92B5CF6073A72D6-3f03e4023393ac70-DB2890571E14F614-880995460e34a0b3


lodash@4.17.21:
  version "4.17.21"
  resolved "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz"
  integrity sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==
~/repos/triage/other-bun$ 
```

nathanhammond:
1. Enabling configuration of lockfile location at the global config level means that the lockfile isn't guaranteed consistent across a team. Currently, the only "safe" (read: assured consistent) way to use `bun` with multiple contributor teams requires setting that value in the local `bunfig.toml`.

2. Splitting the two values from each other enables a configuration where you can’t read after writing.

3. `env` priority is "correct" for `BUN_CONFIG_LOCKFILE_SAVE_PATH` but I don't feel like it should exist. It isn't used in the test suite, makes shared configuration impossible, and appears [semi-arbitrarily added](https://github.com/oven-sh/bun/commit/a765b13f526f5d379c7d2535f496fecd9a202add) two years ago in December 2021.

4. `--lockfile` introduces the same problem, and since it is per-execution, is _extremely_ difficult to ensure consistency. It also gets removed.

***

Approach notes:
- This does not change the peechy schema because I'm ~certain that isn't possible.
- I don't know what I don't know in Zig.
- I _think_ there are some other places that can be cleaned up because `lockfile_path` is now a static string.

***

Closes #5259

### DIFF
--- a/completions/bun.bash
+++ b/completions/bun.bash
@@ -92,12 +92,12 @@ _bun_completions() {
     PACKAGE_OPTIONS[REMOVE_OPTIONS_LONG]="";
     PACKAGE_OPTIONS[REMOVE_OPTIONS_SHORT]="";
 
-    PACKAGE_OPTIONS[SHARED_OPTIONS_LONG]="--config --yarn --production --frozen-lockfile --no-save --dry-run --lockfile --force --cache-dir --no-cache --silent --verbose --global --cwd --backend --link-native-bins --help";
+    PACKAGE_OPTIONS[SHARED_OPTIONS_LONG]="--config --yarn --production --frozen-lockfile --no-save --dry-run --force --cache-dir --no-cache --silent --verbose --global --cwd --backend --link-native-bins --help";
     PACKAGE_OPTIONS[SHARED_OPTIONS_SHORT]="-c -y -p -f -g";
 
-    PM_OPTIONS[LONG_OPTIONS]="--config --yarn --production --frozen-lockfile --no-save --dry-run --lockfile --force --cache-dir --no-cache --silent --verbose --no-progress --no-summary --no-verify --ignore-scripts --global --cwd --backend --link-native-bins --help"
+    PM_OPTIONS[LONG_OPTIONS]="--config --yarn --production --frozen-lockfile --no-save --dry-run --force --cache-dir --no-cache --silent --verbose --no-progress --no-summary --no-verify --ignore-scripts --global --cwd --backend --link-native-bins --help"
     PM_OPTIONS[SHORT_OPTIONS]="-c -y -p -f -g"
-     
+
     local cur_word="${COMP_WORDS[${COMP_CWORD}]}";
     local prev="${COMP_WORDS[$(( COMP_CWORD - 1 ))]}";
 
--- a/completions/bun.zsh
+++ b/completions/bun.zsh
@@ -55,7 +55,6 @@ _bun() {
                 '--no-save[]' \
                 '--dry-run[Don'"'"'t install anything]' \
                 '--force[Always request the latest versions from the registry & reinstall all dependenices]' \
-                '--lockfile[Store & load a lockfile at a specific filepath]:lockfile' \
                 '--cache-dir[Store & load cached data from a specific directory path]:cache-dir' \
                 '--no-cache[Ignore manifest cache entirely]' \
                 '--silent[Don'"'"'t output anything]' \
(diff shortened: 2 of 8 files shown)


Jarred-Sumner:
This makes sense. I think we should do it. I don't know if people are using this flag though

nathanhammond:
> This makes sense. I think we should do it. I don't know if people are using this flag though

Amusingly, if any users are using any of these features after this PR they end up with _precisely_ the same experience that a separate contributor trying to work on their repository would have had.

nathanhammond:
I omitted the alternate proposal because I don't think it provides value and introduces complexity, but, for completeness, this is also a valid approach:

Same:
- Remove `--lockfile`.
- Remove `BUN_CONFIG_LOCKFILE_SAVE_PATH`.
- Remove `install.lockfile.savePath`.
- Remove global config for `install.lockfile.path`.

Different:
_Allow_ local config (`bunfig.toml`) for `install.lockfile.path` with the following conditions:
1. It is lexically cleaned.
2. It does not traverse above `pwd`.
3. It uses `/` as separator, even on Windows.

(Turborepo calls this an `AnchoredUnixPath`.)

Further, during restoration, `path.resolve` or whatever it is on Zig must be prefixed by the root path. (You can't insert a symlink/Junction into the path and jump out of the directory.)

***

The difference in code change between the two approaches are extremely limited and it introduces a lot of unnecessary complexity into ecosystem and the mental model.

- `.bunfig.toml` and `bunfig.toml` have distinct sets of available options.
- You have to teach users what is valid. (What is an `AnchoredUnixPath`?)
- All tooling for `bun` has to implement parsing for `bunfig.toml`, increasing friction for integration.

nathanhammond:
(Test failures appear to be unrelated?)

PR #5346, opened 2023-09-14
In `bunx`, always get latest version when @latest is explicitly passed

Jarred-Sumner:
### What does this PR do?

Previously, `bunx create-t3-app@latest` would use the `/tmp/create-t3-app@latest...`  folder and rely on the OS tmpdir cache removal mechanism to know when to re-download. That's not soon enough. 

This makes it so the following:
```bash
bunx create-t3-app@latest
```

will always check versions from the registry, and possibly re-install it. 

It does not change the scenario when you do `bunx create-t3-app` by itself. That will still be cached for 1-3 days.

Manual. Would be nice to have a test for this.

### DIFF
--- a/src/cli/bunx_command.zig
+++ b/src/cli/bunx_command.zig
@@ -279,65 +279,67 @@ pub const BunxCommand = struct {
 
         const passthrough = passthrough_list.items;
 
-        // Similar to "npx":
-        //
-        //  1. Try the bin in the current node_modules and then we try the bin in the global cache
-        if (bun.which(
-            &path_buf,
-            PATH_FOR_BIN_DIRS,
-            this_bundler.fs.top_level_dir,
-            initial_bin_name,
-        ) orelse bun.which(
-            &path_buf,
-            bunx_cache_dir,
-            this_bundler.fs.top_level_dir,
-            absolute_in_cache_dir,
-        )) |destination| {
-            const out = bun.asByteSlice(destination);
-            _ = try Run.runBinary(
-                ctx,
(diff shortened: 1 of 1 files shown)


src/cli/bunx_command.zig
```diff
@@ -353,22 +355,28 @@ pub const BunxCommand = struct {
...
+            try std.fmt.allocPrint(
                 ctx.allocator,
                 "{s}@{s}",
                 .{
                     update_request.name,
                     display_version,
                 },
             ),
+            // disable the manifest cache when a tag is specified
+            // so that @latest is fetched from the registry
```
t3dotgg, line 374: IMO you should still cache if it's @{specific-version}, just not for tagged releases

Jarred-Sumner: `dist_tag` refers to non-ascii npm tagged versions, like `@1.0.0` would be cached but `@latest` or `@next` would not be cached

mroyme: Couldn't we fetch the version tag for `@latest` and use that for cache? I'm not sure if we can do something similar in `bun`:
```
# ~
; npm view create-t3-app dist-tags.latest --json
"7.20.2"
```

Jarred-Sumner: this is basically like 

package.json:
```
{"dependencies": {"create-t3-app": "latest"}}
```

Then, running:
```
bun update
```

`--no-cache` says "ignore Cache-Control headers from npm when deciding whether to check for new versions of dependencies"

0masn (commented):
lgtm

PR #5387, opened 2023-09-14
feat(encoding): TextDecoder support undefined

WingLim:
### What does this PR do?

Closes #5211

- [x] Code changes

### How did you verify your code works?

- [x] Add test

PR #5390, opened 2023-09-14
Allow undefined as encoding in TextDecoder

### LINKED ISSUE #5211: TextDecoder incompatibility with Nodejs when explicitly passing undefined

1.0.0

Darwin 22.4.0 arm64 arm

The following code will throw a TypeError in Bun but would work just fine in Nodejs.
in Bun 1.0.0
```js
new TextDecoder(undefined) // or pass any undefined variable
/** 
 * throws 
 * TypeError: TextDecoder(encoding) label is invalid
 * code: "ERR_INVALID_ARG_TYPE"
 * */
```
in NodeJS v20
```js
new TextDecoder(undefined) // or pass any undefined variable
/**
* TextDecoder { encoding: 'utf-8', fatal: false, ignoreBOM: false }
* */
```
image

Similar output to NodeJS

### What do you see instead?

```
TypeError: TextDecoder(encoding) label is invalid
 code: "ERR_INVALID_ARG_TYPE"
```

### Additional information

I came across this while I was migrating an existing codebase test suite from nodeJS + Jest to Bun test runner.
it's a small codebase and almost everything worked out of the box 🚀  
except with [MSW's interceptor ](https://github.com/mswjs/interceptors/blob/45df4685cb31d2f66c37fcd60b0f53359d838c98/src/utils/bufferUtils.ts#L7-L10)

https://github.com/mswjs/interceptors/blob/45df4685cb31d2f66c37fcd60b0f53359d838c98/src/utils/bufferUtils.ts#L8

This works in NodeJS without any issues

JibranKalia:
Resolves #5211

### What does this PR do?

This matches behavior in Node where undefined as first argument is considered utf-8

- [X] Code changes

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/bun.js/webcore/encoding.zig
+++ b/src/bun.js/webcore/encoding.zig
@@ -702,6 +702,8 @@ pub const TextDecoder = struct {
                     globalThis.throwInvalidArguments("Unsupported encoding label \"{s}\"", .{str.slice()});
                     return null;
                 }
+            } else if (arguments[0].isUndefined()) {
+                // undefined is used to mean "use the default encoding"
             } else {
                 globalThis.throwInvalidArguments("TextDecoder(encoding) label is invalid", .{});
                 return null;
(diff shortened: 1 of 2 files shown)


a613:
similar to https://github.com/oven-sh/bun/pull/5387

Electroid:
Thanks for working on this! As it turns out, we got a similar PR just a few minutes before this one (what are the odds?!) so we'll close this in favor of that one instead: #5387 

Thank you again!

PR #5419, opened 2023-09-14
fix warnings during bun run publish-layer

nangchan:
When running `bun run publish-layer` the following warnings displayed:

```sh
$ bun scripts/publish-layer.ts
101 |     async load() {
102 |         this.type = this.options.type || 'core';
103 |         this.tag = this.options.tag;
104 |         const root = await findRoot(this.options.name, this.options.root);
105 |         if (!root)
106 |             throw new Error(`could not find package.json with ${(0, util_1.inspect)(this.options)}`);
                      ^
warn: could not find package.json with { type: 'dev',
```

### What does this PR do?

Added missing package @oclif/plugin-plugins to resolved warning

- [x] Code changes

- [x] I ran `bun run publish-layer` and verified layer was published

### DIFF
--- a/packages/bun-lambda/package.json
+++ b/packages/bun-lambda/package.json
@@ -2,6 +2,7 @@
   "name": "bun-lambda",
   "private": true,
   "devDependencies": {
+    "@oclif/plugin-plugins": "^3.5.0",
     "bun-types": "^0.7.0",
     "jszip": "^3.10.1",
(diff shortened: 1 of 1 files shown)


Electroid:
Thanks for fixing this!

nangchan:
> Thanks for fixing this!

Thanks for the merge.

PR #5447, opened 2023-09-15
add warning to Ensure correct placement of the '--watch' flag

a4amado:
### What does this PR do?

 
update docs with a warning to Ensure correct placement of the '--watch' flag in the running command.
 

a4amado:
@Jarred-Sumner

docs/cli/run.md
```diff
@@ -60,10 +60,16 @@ $ bun index.js
 
 To run a file in watch mode, use the `--watch` flag.
 
+> `--watch` won't work if you place it anywhere in the command other than directly after 'bun'.
```
Jarred-Sumner: Thank you for this

It should be a notice instead 

People might not read it closely and copy paste the wrong command

a4amado: Done.

PR #2369, opened 2023-03-11
feat(docker): replace alpine/glibc with distroless

jbergstroem:
The current Dockerfile uses Alpine with a set of glibc libraries. This is not sustainable for a few reasons (see #1567). Since container size is important, we switch to distroless which allows a minimal host but based on debian 11.

Additionally, add an entrypoint that points to bun.

This PR increases the uncompressed container size by ~2MB.

Sidenote: google has aarch64 versions which makes it possible to release this for additional architectures.

---

Todo

- [ ] more testing, I don't use bun much in images
- [ ] fix CI build/label error
- [ ] add opencontainer labels
- [ ] explore more architectures than amd64

PR #2126, opened 2023-02-22
Fix glibc error in latest alpine

WebReflection:
There's an override error with latest alpine and various versions of glibc (including 2.35): https://github.com/sgerrand/alpine-pkg-glibc/issues/185

This MR proposes a workaround (tested on fly.io - it works) so that it's possible to have bun usable with minimal linux dependencies.

PR #5454, opened 2023-09-15
alpine-with-glic-that-works

pilishen:
### What does this PR do?

based on the original alpine image, which cannot run bun successfully , 

if we add `gcompat` into it , bun can run smoothly in most cases, we can happily use bun to boost vite dev performance, such as : `bun --bun x vite --host`

but in some edge cases, especially lightningcss related scenarioes, we still need further add `glic-bin` into this container. 
FYI, lightningcss will be used heavily in next version of tailwindcss -- tailwindcss 3.4. And since vite 4.4, there's a experimental feature using lightningcss as its css moudule.  

```
RUN /usr/glibc-compat/bin/ldd /usr/local/bin/bun
```
BTW, this line can be omitted , just to make sure there are no errors in loading bun related libraries


pilishen:
related #2126 #2369

pilishen:
if without `glic-bin`, bun with lightningcss will trigger error like this:
```
Cannot find module '../lightningcss.linux-x64-gnu.node'
```
related [Cannot find module '../lightningcss.linux-x64-gnu.node'](https://github.com/parcel-bundler/lightningcss/issues/567)

[Cannot find module '../lightningcss.linux-x64-gnu.node'](https://github.com/parcel-bundler/lightningcss/issues/279)

Issue #5420, opened 2023-09-14
The ES modules aren't always asynchronous.


DevinJohw:
### What is the type of issue?

Documentation is confusing

### What is the issue?

I see something in the Bun documentation like this:
"The biggest difference between CommonJS and ES Modules is that CommonJS modules are synchronous, while ES Modules are asynchronous. ". The static ES modules are synchronous, only the modules imported by`import()` function are asynchronous. This sentence is kind of confusing.

### Where did you find it?

This is the URL: https://bun.sh/docs/runtime/modules

h2210316651:
Created a PR to update the Docs [#5456](https://github.com/oven-sh/bun/pull/5456)

DevinJohw:
Awesome!

MonsterDeveloper:
Seems like an issue is resolved. Should this be closed then?

PR #5456, opened 2023-09-15
Updated modules.md to address issue #5420

h2210316651:
Updated docs to address `The ES modules aren't always asynchronous. #5420` Issue. 
- Included an elaborate explanation citing the difference between CommonJS and ES Modules, 
- Added a summary


colinhacks:
Love it, thanks!

Issue #5314, opened 2023-09-13
Build, OnResolve() doesn't support pending promises yet


gungun974:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Linux 6.3.12-200.fc38.x86_64 x86_64 unknown (Fedora 38)

### What steps can reproduce the bug?

I try to create a custom loader that take an image and do some transformation on it.
For that I decided to use `sharp` to do the work. (I will don't bother to tell `sharp` installation is not easy, that not the subject of this issue).

After successfully install sharp I decide to test if with Bun I can change the path of a file that is loaded.
I manage to do that with this tiny example :
```ts
import { type BunPlugin } from 'bun';
import sharp from 'sharp';

function ImageLoaderPlugin(): BunPlugin {
  return {
    name: 'ImageLoader',
    async setup(build) {
      build.onResolve({ filter: /\.(png)/ }, _ => {
        return {
          path: '/tmp/test.jpg',
        };
      });
    },
  };
}

export default ImageLoaderPlugin;
```
Here every `png` return the same file, and when I build the file is properly copy into the build folder.

Now I edit this code to take an image and try to return a transformed version with `sharp`
```ts
import { type BunPlugin } from 'bun';
import sharp from 'sharp';

function ImageLoaderPlugin(): BunPlugin {
  return {
    name: 'ImageLoader',
    async setup(build) {
      build.onResolve({ filter: /\.(png)/ }, async _ => {
        const image = sharp('/a/path/to/a/real/file/img1.png');

        await image.toFile('/tmp/test.jpg');

        return {
          path: '/tmp/test.jpg',
        };
      });
    },
  };
}

export default ImageLoaderPlugin;
```
And here I get this error when I try to run my plugin in Bun runtime
```
onResolve() doesn't support pending promises yet
```
The things I expect is of course that Bun runtime loader will wait like he does with `build.onLoad` but it appear it does not.
Is there any technical reason for now I can do the same thing ?

Just for the note my program works when using `Bun.build`, it's only the runtime who can't perform async operation in `build.onResolve`

So maybe my issue is not really a bug but more a missing feature but you known, it better to write down things and repeats than never talk about it and forgot it for many years later.

Anyway thanks to all the Bun team for this incredible 1.0 release ^^

PR #5477, opened 2023-09-15
Add support for async plugin `build.onResolve` at runtime

gungun974:
### What does this PR do?

This PR addresses my issue outlined in #5314.

The problem with #5314 is that the Bun runtime does not allow plugins to handle asynchronous imports.
I personally need this feature, so I decided to create this PR to implement it for all the Bun users.
I also planed after this PR to create a Bun plugin capable of importing image and do some image operation before building and live.

Although I'm not really familiar with Zig, I have knowledge in C/C++ and Rust and I know how computer memory can been slow some time. So I try the best to not create any useless variable.
Please Verify I didn't anything is not idiomatic to Zig and will hurt de performance of Bun.

- [x] Code changes

- [x] I have written some tests for the bundler for verify if it's still works.
- [x] I have written some tests for the runtime plugin to verify if I didn't broke the none async `builder.onResolve` and if the async `build.onResolve` work as expected
- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be ***(But since I'm not a zig expert so please verify)***

### DIFF
--- a/test/bundler/bundler_plugin.test.ts
+++ b/test/bundler/bundler_plugin.test.ts
@@ -28,12 +28,25 @@ describe("bundler", () => {
   };
 
   itBundled("plugin/Resolve", {
-    todo: true,
     files: resolveFixture,
     // The bundler testing api has a shorthand where the plugins array can be
     // the `setup` function of one plugin.
     plugins(builder) {
       builder.onResolve({ filter: /\.magic$/ }, args => {
+        return {
+          path: path.resolve(path.dirname(args.importer), args.path.replace(/\.magic$/, ".ts")),
+        };
+      });
+    },
+    run: {
+      stdout: "foo",
+    },
+  });
+  itBundled("plugin/ResolveAsync", {
+    files: resolveFixture,
+    plugins(builder) {
+      builder.onResolve({ filter: /\.magic$/ }, async args => {
+        await new Promise(resolve => setTimeout(resolve, 10));
         return {
           path: path.resolve(path.dirname(args.importer), args.path.replace(/\.magic$/, ".ts")),
         };
--- a/test/bundler/bundler_plugin.test.ts
+++ b/test/bundler/bundler_plugin.test.ts
@@ -46,7 +46,7 @@ describe("bundler", () => {
     files: resolveFixture,
     plugins(builder) {
       builder.onResolve({ filter: /\.magic$/ }, async args => {
-        await new Promise(resolve => setTimeout(resolve, 10));
+        await Bun.sleep(10);
         return {
           path: path.resolve(path.dirname(args.importer), args.path.replace(/\.magic$/, ".ts")),
         };
--- a/test/js/bun/plugin/plugins.test.ts
+++ b/test/js/bun/plugin/plugins.test.ts
@@ -114,7 +114,7 @@ plugin({
   name: "async resolve",
   setup(builder) {
     builder.onResolve({ filter: /original/, namespace: "async-resolve" }, async () => {
-      await new Promise(resolve => setTimeout(resolve, 10));
+      await Bun.sleep(10);
       return {
         path: "changed",
         namespace: "async-resolve",
--- a/test/js/bun/plugin/plugins.test.ts
+++ b/test/js/bun/plugin/plugins.test.ts
@@ -279,7 +279,7 @@ describe("dynamic import", () => {
   });
 
   it("should change resolved import when onResolve is async", async () => {
-    const result = require("async-resolve:original");
+    const result = await import("async-resolve:original");
     expect(result.default).toBe("changed");
   });
 
--- a/test/js/bun/plugin/plugins.test.ts
+++ b/test/js/bun/plugin/plugins.test.ts
@@ -96,6 +96,39 @@ plugin({
 });
 
 plugin({
+  name: "resolve",
+  setup(builder) {
+    builder.onResolve({ filter: /original/, namespace: "resolve" }, () => ({
+      path: "changed",
+      namespace: "resolve",
+    }));
+
+    builder.onLoad({ filter: /changed/, namespace: "resolve" }, () => ({
+      contents: `export default "changed";`,
+      loader: "js",
+    }));
+  },
+});
+
+plugin({
+  name: "async resolve",
+  setup(builder) {
+    builder.onResolve({ filter: /original/, namespace: "async-resolve" }, async () => {
+      await new Promise(resolve => setTimeout(resolve, 10));
+      return {
+        path: "changed",
+        namespace: "async-resolve",
+      };
+    });
+
+    builder.onLoad({ filter: /changed/, namespace: "async-resolve" }, () => ({
+      contents: `export default "changed";`,
+      loader: "js",
+    }));
+  },
+});
+
+plugin({
   name: "async onLoad",
   setup(builder) {
     globalThis.asyncOnLoad = "";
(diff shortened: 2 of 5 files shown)


gungun974:
I may have managed to add unit tests in the end but I didn't understand how I could run them on my machine.

When I use Bun test, I notice that my tests are done with my installed version of bun only.

In short I may be the worst person to made a PR on this repository but hey, you know it's the effort that counts and the willingness ^^

gungun974:
Like said, I made this plugin, the [Bun Image Transform](https://github.com/gungun974/bun-image-transform) to easily convert and transform image on import.

test/bundler/bundler_plugin.test.ts
```diff
@@ -43,6 +42,20 @@ describe("bundler", () => {
       stdout: "foo",
     },
   });
+  itBundled("plugin/ResolveAsync", {
+    files: resolveFixture,
+    plugins(builder) {
+      builder.onResolve({ filter: /\.magic$/ }, async args => {
+        await new Promise(resolve => setTimeout(resolve, 10));
```
Jarred-Sumner:
```suggestion
        await Bun.sleep(10);
```

test/js/bun/plugin/plugins.test.ts
```diff
@@ -95,6 +95,39 @@ plugin({
...
+      loader: "js",
+    }));
+  },
+});
+
+plugin({
+  name: "async resolve",
+  setup(builder) {
+    builder.onResolve({ filter: /original/, namespace: "async-resolve" }, async () => {
+      await new Promise(resolve => setTimeout(resolve, 10));
```
Jarred-Sumner:
```suggestion
      await Bun.sleep(10);
```

Jarred-Sumner:
Thank you for this

It's technically possible to do this -- but permitting it is a massive performance footgun. Sharp should not be executed during module resolution. It should only be executed onLoad, which fully supports async/await. Multiple files can be loaded simultaneously. Module resolution is (unfortunately) synchronous. If the API allows this, people will get confused (as shown in this PR) 

I'm not confident this makes sense for Bun to support.

gungun974:
> Thank you for this
> 
> It's technically possible to do this -- but permitting it is a massive performance footgun. Sharp should not be executed during module resolution. It should only be executed onLoad, which fully supports async/await. Multiple files can be loaded simultaneously. Module resolution is (unfortunately) synchronous. If the API allows this, people will get confused (as shown in this PR)
> 
> I'm not confident this makes sense for Bun to support.

I was unaware of this difference between the `onResolve` and `onBuild` APIs of plugins. 
The initial reason I went down this implementation route was that I found it strange that Bun could bundle an asynchronous `onResolve` and that `esbuild` could do the same for its `onResolve.` 
However, I can understand if `onResolve` is 100% synchronous and blocking, that this could pose a problem.

Afterwards, I had three reasons for thinking that this PR could be good.

The first reason is that technically what my plugin aims to do is not to load a file with its content, but to indicate to Bun that the import should be done with another file. That's why `onResolve` seemed more coherent to me.

The second reason is the infamous issue with async functions. Once you find yourself with an async function, you can't make it synchronous even if you want to. There's no magical AsyncToSync function.

The third reason is that `onLoad` simply does not support passing a path to the pre-existing `file"` loader. This results in Bun never copying the transformed assets into the build folder when bundling the project, which is very inconvenient in my case.

In any case, here are the initial three reasons why I went with this implementation. 
If this is behavior that Bun should never have at runtime, I can understand. 
On the other hand, that would mean that the phrase `onResolve() doesn't support pending promises yet` should be changed, and that differences could exist between Build and Runtime modes, which in my opinion could confuse other people.

 thank you for looking at my PR @Jarred-Sumner

gungun974:
Of course, just one last thing regarding the use of Sharp in `onResolve`. Obviously, it's a heavy operation, which is why in my plugin I don't regenerate an image that has already been generated.

It's my responsibility (the plugin maker), not really Bun's, to ensure that I'm not running a 15-second process between each Import. (I'm exaggerating with the time, but you get the point.)

gungun974:
# The Discovery
So here some news about the development of my Plugin.

Since you mentioned performance issues, I looked into whether it would be possible to add the missing file loader to Bun from the start. 

So I was working in the source code and starting to write unit tests. While writing these tests, I wondered if I could create an `onLoad` that returns an import function that would import as Bun already does with file. 

It turns out this is possible and actually ***mimics the behavior of `onResolve`, but async,*** with a slightly longer process.

So in this branch of my plugin I made some experiment between the version that used this patch and my workaround with bun 1.0.2 https://github.com/gungun974/bun-image-transform/tree/asyncOnResolveWorkaround

# The benchmarks

For each test, I based myself on this command to run the tests 1000 times and see the performance difference between the PR and directly using `onLoad`, which returns an import.
```bash
bun test --rerun-each 1000
```

## Here the results of Bun 1.0.2 with my work around :
```
 14985 pass
 0 fail
 25974 expect() calls
Ran 14985 tests across 1000 files. [8.18s]
```

## Here the result of Bun 1.0.3 with my patch and the async thing :
```
[...]
tests/plugin.test.ts: (run #108)
[...]
zsh: IOT instruction (core dumped)  ~/bun test --rerun-each 1000
```
If I rerun :
```
[...]
tests/plugin.test.ts: (run #30)
[...]
zsh: segmentation fault (core dumped) 
```
Again :
```
[...]
tests/plugin.test.ts: (run #78)
[...]
GC scan found corrupt object: structureID is zero!
GC type: Eden
Object at: 0x7fa5d1eb5260
Structure ID: 0 (0x7f8e00000000)
Object contents: 0x0000000000000000 0x0000000000000000
Is marked: true
Is newly allocated: true
Block: 0x7fa5d1eb4000
Live:YES, Empty:no, Allocated:no, CanAllocateButNotEmpty:no, Destructible:no, Eden:YES, Unswept:no, MarkingNotEmpty:YES, MarkingRetired:no
Is marked raw: true
Marking version: 2
Heap marking version: 2
Is newly allocated raw: true
Newly allocated version: 128
Heap newly allocated version: 128
zsh: IOT instruction (core dumped) 
```
# So what ?

As you can see, the results for my patch are not great. While normally no one should ever do this with my plugin, the fact that one could theoretically crash Bun is simply not acceptable.

Even though I personally find it odd that we only have async `onResolve` in Bun.build, adding async in runtime with my patch would be far too unstable.

I could try to find ways to speed this up, but as you've seen, it's not worth it. In reality, no one would ever need an async `onResolve`.

However, the reason I'm here today is because in ticket #5314, the error message was `onResolve() doesn't support pending promises yet`. 
This could lead others, like me, to be confused in the future, thinking that Bun plans to implement this feature one day, when in fact that will never be the case.

Also, the reason I did all this was because nowhere in the documentation was it written that it was possible to simply use the following method (see below) to change the loaded file.
```ts
build.onLoad({ filter: /&bunimg$/ }, async (args) => {
    const newPath = await someAwaitThing(args.path);
    
    return {
        contents: `
            export {default} from ${JSON.stringify(newPath)}
        `,
        loader: "js",
    };
});
```
This forces one to think that if they want to load a file, they must go through `onResolve`, and if they can't handle asynchronous behavior, then it's simply impossible. However, as I was able to show you, that's not the case.

# What to do so ?

I propose two modifications that go against the original direction of this PR:

- Change the error message for asynchronous `onResolve` to `onResolve() with pending promises is unstable in Bun Runtime.` this way, no one else would look to modify it **like I did**.

- Secondly, make changes in the `runtime/plugins` section of the documentation so that:
  - It's clearly understood that for performance reasons, `onResolve` cannot be asynchronous, unlike `onLoad`.
  - It's reminded to plugin creators that it's possible to mimic the behavior of asynchronous `onResolve` and the file loader, with a simple JS rexport as I showed you earlier.

These changes could prevent future confusion and guide developers more effectively.

In any case, I'm grateful for your comment and the magic of the 12-hour time difference, which allowed me to conduct this in-depth study of the issue and propose an alternative that will likely suit both you and me, and also benefit the Bun community.

The only question to ask you is whether I should close this PR and open another PR with these fixes, or can I save time and just change all my commits and the title? This would avoid having to recontextualize the whole history.

In short, @Jarred-Sumner, I know this was long to write, but it's just that I want to give my best for this young and incredible community that is Bun. You are doing truly amazing work, and that's why, even if my contribution will be smaller, we will all move forward together!

I await your opinion and whether I should create another PR now. ^^

PR #5485, opened 2023-09-15
perf(node:events): optimize `emit(...)` function

### LINKED ISSUE #5136: `node:readline`: listening to 'line' event of `readline.Interface` slows down code execution

Darwin 22.5.0 arm64 arm

I noticed that attaching a callback function to the ['line'](https://nodejs.org/api/readline.html#event-line) event of the [`readline.Interface`](https://nodejs.org/api/readline.html#class-readlineinterface) slows down the code execution in my case by around 25%. Even if the callback function has an empty body. Node.js shows no performance slowdown in the same example.

#### test.js
```javascript
import EventEmitter from 'node:events';
import { createReadStream } from 'node:fs';
import { performance } from 'node:perf_hooks';
import { createInterface } from 'node:readline';

let tStart, tdiff;

// warmup run to avoid any startup issues or buffering etc
const lineReader0 = createInterface({
    input: createReadStream('lichess_db_standard_rated_2013-12.pgn')
});
await EventEmitter.once(lineReader0, 'close');

// ======================
//        TEST 1
// ======================
console.log('TEST 1: line reader without callback');
tStart = performance.now();

// init line reader
const lineReader1 = createInterface({
    input: createReadStream('lichess_db_standard_rated_2013-12.pgn')
});

// wait for line reader to finish
await EventEmitter.once(lineReader1, 'close');
tdiff = Math.round(performance.now() - tStart) / 1000;
console.log(`--> Read all lines in ${tdiff}s.\n`);

// ======================
//        TEST 2
// ======================
console.log('TEST 2: line reader with empty callback');
tStart = performance.now();

// init line reader
const lineReader2 = createInterface({
    input: createReadStream('lichess_db_standard_rated_2013-12.pgn')
});

// empty on 'line' callback
lineReader2.on('line', (_) => {});

// wait for line reader to finish
await EventEmitter.once(lineReader2, 'close');
tdiff = Math.round(performance.now() - tStart) / 1000;
console.log(`--> Read all lines in ${tdiff}s.\n`);

// ======================
//        TEST 3
// ======================
console.log('TEST 3: line reader with callback function counting the lines');
tStart = performance.now();

// init line reader
const lineReader3 = createInterface({
  input: createReadStream('lichess_db_standard_rated_2013-12.pgn')
});

let lineCount = 0;
// on 'line' callback: count lines
lineReader3.on('line', (_) => {
    lineCount += 1;
});

// wait for line reader to finish
await EventEmitter.once(lineReader3, 'close');
tdiff = Math.round(performance.now() - tStart) / 1000;
console.log(
    `--> Read ${lineCount} lines in ${tdiff}s (${Math.round(
        lineCount / tdiff
    )} lines/s).\n`
);

```

#### run with bun
```
>bun test.js
TEST 1: line reader without callback
--> Read all lines in 0.278s.

TEST 2: line reader with empty callback
--> Read all lines in 0.352s.

TEST 3: line reader with callback function counting the lines
--> Read 2989876 lines in 0.356s (8398528 lines/s).
```

#### run with node.js v20.6.1
```
>node test.js
TEST 1: line reader without callback
--> Read all lines in 0.279s.

TEST 2: line reader with empty callback
--> Read all lines in 0.28s.

TEST 3: line reader with callback function counting the lines
--> Read 2989876 lines in 0.28s (10678129 lines/s).
```

It should be reproducible with any large enough file, but if needed I can upload my example file as well.

Since `readline.Interface` inherits from `EventEmitter` it could also be a general issue with the `EventEmitter` class.

If node.js is the benchmark to beat, listening to the 'line' event should not slow down execution.

### What do you see instead?

A slowdown of around 25% simply by defining a callback.

### Additional information

Not sure if performance issues are already in your scope at this point in time, but I tried running my library [chessalyzer.js](https://github.com/YSchroe/chessalyzer.js) with bun and noticed that it is significantly slower compared to running it with node.js. The library is used to parse chess games from (large) PGN files and basically replays the games to extract some KPIs from them. It follows three main steps:

1. Read-in the PGN file asynchronously via the node:readline module
2. If enough lines have been read in pass the data to a node:worker_threads.Worker
3. In there parse the PGN (regex, string compare etc.) and extract the game stats. Then pass the data back to the main thread.

For the first step I already noticed some performance differences, described in this issue. Is fixing performance issues in your scope yet? Then I will have a look at the other two steps as well.

yschroe:
### What does this PR do?

Optimizes how `EventEmitterPrototype.emit(...)` triggers the callback functions, especially noticable with the 'line' event of readline.Interface.

- Brings performance for emits with only one argument for `...args` to Node.js level
- Emits with multiple args should have improved performance as well

Fixes #5136 
_(Please also see my comments there, I think there is a different underlying issue regarding array cloning/spreading which is just worked around with this pull request)_

- [X] Code changes

I did manual tests and also ran the tests in test/js/node/events


yschroe:
Unsure about the "...committed the transpiled changes" part. I have 4 changed files in my src/js/out folder, do I need to commit them all? WebCoreJSBuiltins.{cpp||d.ts|h} and InternalModuleRegistryConstants.h.

Jarred-Sumner:
Do you have an accompanying benchmark? Can you paste the before and after results for the benchmarks in bench/emitter?

https://github.com/oven-sh/bun/blob/main/bench/emitter/microbench.mjs#L1

yschroe:
- Missed `rawListeners()` -> added. 
- Also the tests for `rawListeners()` were not testing the correct function before -> fixed

Will add benchmark in a sec...

yschroe:
#### microbench.mjs (before)
```
cpu: AMD Ryzen 5 2600X Six-Core Processor
runtime: bun 1.0.1 (x64-linux)

benchmark                       time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------ -----------------------------
single emit: bun             99.03 ns/iter  (82.22 ns … 435.76 ns)  91.19 ns 318.49 ns 334.33 ns
on x 10_000 (handler): bun   10.37 ms/iter    (4.34 ms … 16.64 ms)  12.74 ms  16.64 ms  16.64 ms
```

#### microbench.mjs (after)
```
cpu: AMD Ryzen 5 2600X Six-Core Processor
runtime: bun 1.0.1_debug (x64-linux)

benchmark                       time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------ -----------------------------
single emit: bun             47.67 ns/iter  (36.96 ns … 371.77 ns)  40.03 ns 242.92 ns 278.15 ns
on x 10_000 (handler): bun    3.25 ms/iter   (954.47 µs … 5.99 ms)   4.28 ms   5.77 ms   5.99 ms
```

src/js/node/events.js
```diff
@@ -92,9 +92,11 @@ const emitWithoutRejectionCapture = function emit(type, ...args) {
...
-
-  for (var handler of [...handlers]) {
-    handler.apply(this, args);
+  // iterate in reverse so the handlers array does not need to be cloned
+  for (let i = handlers.length - 1; i >= 0; --i) {
+    // for performance reasons Function.call(...) is used if only one arg needs to be passed
+    if (args.length === 1) handlers[i].call(this, args[0]);
+    else handlers[i].apply(this, args);
   }
   return true;
```
jdalton, line 120: The big perf win is the `.call` when args are 0 or 1. It can be done to 0, 1, 2 (or however a common call is) before going to `apply`

src/js/node/events.js
```diff
@@ -92,9 +92,11 @@ const emitWithoutRejectionCapture = function emit(type, ...args) {
   if (events === undefined) return false;
   var handlers = events[type];
   if (handlers === undefined) return false;
-
-  for (var handler of [...handlers]) {
```
jdalton, line 96: The use of `for-of` and all the spreading and stuff isn't great for JS perf hot spots.

src/js/node/events.js
```diff
@@ -132,7 +138,9 @@ EventEmitterPrototype.addListener = function addListener(type, fn) {
     events[type] = [fn];
     this._eventsCount++;
   } else {
-    handlers.push(fn);
+    // for performance reasons the handlers are stored in reverse
+    // -> adding = unshifting
+    handlers.unshift(fn);
```
jdalton: `unshift` is slow... Look like either forward or backward there is a `unshift` use.

src/js/node/events.js
```diff
@@ -247,15 +257,17 @@ EventEmitterPrototype.listeners = function listeners(type) {
   if (!events) return [];
   var handlers = events[type];
   if (!handlers) return [];
-  return handlers.map(x => x.listener ?? x);
+  // reverse array since handlers are stored in reverse internally
+  return handlers.reverse().map(x => x.listener ?? x);
```
jdalton: Use of `.map` and things tends to be slow when stacked on other ES5 array methods. Individually less so. But still keeping things bare bones may be better here.

src/js/node/events.js
```diff
@@ -247,15 +257,17 @@ EventEmitterPrototype.listeners = function listeners(type) {
...
 };
 
 EventEmitterPrototype.rawListeners = function rawListeners(type) {
   var { _events } = this;
   if (!_events) return [];
   var handlers = _events[type];
   if (!handlers) return [];
-  return handlers.slice();
+  // reverse array since handlers are stored in reverse internally
+  return handlers.reverse().slice();
```
jdalton: It's worth looking at Node's micro opts in array cloning here:
https://github.com/nodejs/node/blob/main/lib/events.js#L883-L894

src/js/node/events.js
```diff
@@ -92,9 +92,11 @@ const emitWithoutRejectionCapture = function emit(type, ...args) {
   if (events === undefined) return false;
   var handlers = events[type];
   if (handlers === undefined) return false;
-
-  for (var handler of [...handlers]) {
-    handler.apply(this, args);
+  // iterate in reverse so the handlers array does not need to be cloned
+  for (let i = handlers.length - 1; i >= 0; --i) {
```
jdalton: ☝️ Node has prependListener and addListener so cloning would need to take place as the listeners can be modified while iterating.

src/js/node/events.js
```diff
@@ -107,8 +109,12 @@ const emitWithRejectionCapture = function emit(type, ...args) {
...
   var handlers = events[type];
   if (handlers === undefined) return false;
-  for (var handler of [...handlers]) {
-    var result = handler.apply(this, args);
+  // iterate in reverse so the handlers array does not need to be cloned
+  for (let i = handlers.length - 1; i >= 0; --i) {
+    let result;
+    // for performance reasons Function.call(...) is used if only one arg needs to be passed
+    if (args.length === 1) result = handlers[i].call(this, args[0]);
+    else result = handlers[i].apply(this, args);
```
jdalton: It would be wonderful if event loops didn't have the `this` binding. I experimented over the years with trying to detect if it was needed but the detection was always too heavy. Essentially arrow-function or bound listeners don't need the `.call` and could be executed straight up.

jdalton: ... oh wait. bound functions and arrow functions don't have a `.prototype` property.... 
```js
var a = () => {}
a.prototype
// undefined
'prototype' in a
// false
var a = function() {}.bind();
a.prototype
// undefined
'prototype' in a
// false
```
Maybe a little sniff there could turn it into a non-call invocation.

src/js/node/events.js
yschroe: Ah, you mean the case where the called callback may actually add new listeners to the event? You are right, this might mess up this logic. So back to array cloning it is. 😕 I can also add a new test for this since this PR passes all tests atm.
(At work atm, will check your other comments later 🙂)

yschroe:
Thanks for the review @jdalton!

I now reverted back to cloning the handler array, since my version did not handle the case in which a new callback is added via a callback correctly.

Still due to two changes there is still a performance improvement vs. bun v1.0.2.

1. If only one handler is attached, I still don't clone the array. Should be no problem since arr.length is 1 so we don't care about changes in length after handling the first and only callback.
2. Use `Function.call()` instead of `Function.apply()` whenever possible.

#### Performance
```
bun v1.0.2_debug                     time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------------- -----------------------------
single emit (one handler): bun     42.9 ns/iter  (32.83 ns … 450.39 ns)  34.58 ns 250.83 ns 267.58 ns
single emit (two handlers): bun   69.18 ns/iter  (53.71 ns … 351.81 ns)  63.47 ns  295.4 ns 322.01 ns
on x 10_000 (one handler): bun      5.2 ms/iter    (1.63 ms … 10.01 ms)   6.87 ms   9.61 ms  10.01 ms
on x 10_000 (two handlers): bun     7.7 ms/iter    (2.68 ms … 13.66 ms)   10.3 ms  13.66 ms  13.66 ms


bun v1.0.2                           time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------------- -----------------------------
single emit (one handler): bun     95.4 ns/iter   (79.21 ns … 434.1 ns)  84.69 ns  349.3 ns 358.03 ns
single emit (two handlers): bun  136.63 ns/iter (121.44 ns … 394.66 ns) 127.53 ns 359.12 ns 373.09 ns
on x 10_000 (one handler): bun    10.17 ms/iter    (4.29 ms … 16.98 ms)  13.19 ms  16.98 ms  16.98 ms
on x 10_000 (two handlers): bun   22.22 ms/iter    (11.4 ms … 34.03 ms)  26.96 ms  34.03 ms  34.03 ms
```

Also I've added two new testcases to catch this one use-case in which my old implementation would fail.

src/js/node/events.js
yschroe: I found that there was actually no measurable difference between `handler()` and `handler.call(this)` so I just kept the `.call` version for now. Getting rid of `.apply` made a huge difference though.

src/js/node/events.js
yschroe: Yes `.apply()` is super slow for some reason. I now use `.call()` in the cases for 0, 1, 2 and 3 args and `.apply()` else.

Here is a quick bench for the case with `args.length = 1`:
```
handler.apply()                      time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------------- -----------------------------
single emit (one handler): bun    69.65 ns/iter  (57.65 ns … 337.51 ns)  61.58 ns  299.2 ns 306.25 ns
single emit (two handlers): bun  124.88 ns/iter (107.02 ns … 377.68 ns)  117.1 ns 343.13 ns 355.99 ns
on x 10_000 (one handler): bun    10.01 ms/iter     (3.9 ms … 17.88 ms)  12.75 ms  17.88 ms  17.88 ms
on x 10_000 (two handlers): bun   15.82 ms/iter    (7.33 ms … 24.57 ms)  19.88 ms  24.57 ms  24.57 ms


handler.call()                       time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------------- -----------------------------
single emit (one handler): bun     42.9 ns/iter  (32.83 ns … 450.39 ns)  34.58 ns 250.83 ns 267.58 ns
single emit (two handlers): bun   69.18 ns/iter  (53.71 ns … 351.81 ns)  63.47 ns  295.4 ns 322.01 ns
on x 10_000 (one handler): bun      5.2 ms/iter    (1.63 ms … 10.01 ms)   6.87 ms   9.61 ms  10.01 ms
on x 10_000 (two handlers): bun     7.7 ms/iter    (2.68 ms … 13.66 ms)   10.3 ms  13.66 ms  13.66 ms
```

Node.js seems to always use `apply()` though with no performance loss, as far as I can tell. Could be an issue in JavaScriptCore? Or where is apply implemented?

jdalton: It's been an issue, `call` vs `apply`, in the past with engines so it's definitely possible.

src/js/node/events.js
jdalton: Perfect! Thank you for digging into that.

src/js/node/events.js
```diff
@@ -92,9 +92,28 @@ const emitWithoutRejectionCapture = function emit(type, ...args) {
   if (events === undefined) return false;
   var handlers = events[type];
   if (handlers === undefined) return false;
-
-  for (var handler of [...handlers]) {
-    handler.apply(this, args);
+  // clone handlers array if necessary since handlers can be added/removed during the loop
+  const maybeClonedHandlers = handlers.length > 1 ? handlers.slice() : handlers;
+  for (let i = 0; i < maybeClonedHandlers.length; i++) {
```
jdalton:
```suggestion
  for (let i = 0, { length } = maybeClonedHandlers; i < length; i++) {
```
Engines can inline this access but we can help them out. It's low level but a good habit to be part of the repeated low-level loop scaffolding

src/js/node/events.js
```diff
@@ -92,9 +92,28 @@ const emitWithoutRejectionCapture = function emit(type, ...args) {
...
+      case 2:
+        handler.call(this, args[0], args[1]);
+        break;
+      case 3:
+        handler.call(this, args[0], args[1], args[2]);
+        break;
+      default:
+        handler.apply(this, args);
+        break;
+    }
```
jdalton, line 118: ☝️ Do we know what common arguments are passed to event handlers? So we can intuit possible counts to unroll to.

src/js/node/events.js
```diff
@@ -92,9 +92,28 @@ const emitWithoutRejectionCapture = function emit(type, ...args) {
   if (events === undefined) return false;
   var handlers = events[type];
   if (handlers === undefined) return false;
-
-  for (var handler of [...handlers]) {
-    handler.apply(this, args);
+  // clone handlers array if necessary since handlers can be added/removed during the loop
+  const maybeClonedHandlers = handlers.length > 1 ? handlers.slice() : handlers;
```
jdalton, line 98: Instead of `.slice` what about using `arrayClone`:
https://github.com/nodejs/node/blob/18e00a577d7405ca0e6c26113675c230677342c2/lib/events.js#L883-L894

yschroe: I tried that as well but it did not really improve performance. So I just used `slice()` to keep it simple.

#### arrayClone()
```
benchmark                            time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------------- -----------------------------
single emit (one handler): bun    46.64 ns/iter  (32.96 ns … 590.09 ns)  38.89 ns 280.46 ns 309.31 ns
single emit (two handlers): bun   71.91 ns/iter  (46.66 ns … 423.27 ns)  68.05 ns  287.8 ns 338.52 ns
on x 10_000 (one handler): bun     5.75 ms/iter    (2.01 ms … 12.03 ms)   7.27 ms  11.81 ms  12.03 ms
on x 10_000 (two handlers): bun    8.48 ms/iter    (3.11 ms … 15.64 ms)   11.2 ms  15.64 ms  15.64 ms
```

jdalton: Ah cool! What if you tested the number of handlers arrayClone optimizes for, so 2-6 (I don't know why 1 is skipped). And then when it breaks over 6 to say the unoptimized 7 handler count.

src/js/node/events.js
yschroe: I had a quick look through buns codebase and the cases with 0 or 1 arguments for `emit(...)` are by far the most common cases. Some emits did emit 2 args, but I did not find any case with 3 args yet. So from my point of view handling the cases 0-3 should probably cover 99% of all use-cases.

src/js/node/events.js
yschroe: Case 1 is skipped in Node.js because they actually do not store the handlers as an array, if only one handler is attached. But this leads to some ugly `typeof x === 'function'` checks 😵.
https://github.com/nodejs/node/blob/main/lib/events.js#L513-L537

jdalton: If Bun's approach is faster or on par and we don't have to unroll as far that's a Win for Bun :)

paperclover:
this is really cool to see optimized.

> Unsure about the "...committed the transpiled changes" part. I have 4 changed files in my src/js/out folder, do I need to commit them all? WebCoreJSBuiltins.{cpp||d.ts|h} and InternalModuleRegistryConstants.h.

Only `InternalModuleRegistryConstants.h` should have changed. Please commit that one please.

src/js/node/events.js
paperclover: > Case 1 is skipped in Node.js because they actually do not store the handlers as an array, if only one handler is attached. But this leads to some ugly `typeof x === 'function' checks 😵.
> https://github.com/nodejs/node/blob/main/lib/events.js#L513-L537

when i did the initial version of `events.js` i found that JSC was faster when using a single length array vs the function-or-array approach. tbh we should add a comment about this choice

i forgot if i even tried benching the `arrayClone` function.

src/js/node/events.js
paperclover: i just want to add, and it may not perform well but it is worth trying: `call(x, ...args)` instead of apply. i'm sure they both are jit-d the same but just throwing out there.

test/js/node/events/event-emitter.test.ts
```diff
@@ -288,18 +357,26 @@ describe("EventEmitter", () => {
...
+    myEmitter.once("foo", fn3);
+    expect(myEmitter.listeners("foo")).toEqual([fn, fn3]);
   });
 
   test("rawListeners", () => {
     const myEmitter = new EventEmitter();
     const fn = () => {};
     myEmitter.on("foo", fn);
-    expect(myEmitter.listeners("foo")).toEqual([fn]);
+    expect(myEmitter.rawListeners("foo")).toEqual([fn]);
```
paperclover, line 369: good catch

src/js/node/events.js
yschroe: I just tried that and spreading the args into `.call` was way slower than using `.apply`. Not sure what is causing that exactly.

#### For a test with 4 args in the emit
```
handler.call(this, ...args)        time (avg)             (min … max)       p75       p99      p995
--------------------------------------------------------------------- -----------------------------
single emit (5 handlers): bun  274.39 ns/iter (204.39 ns … 634.24 ns) 321.03 ns 587.95 ns 634.24 ns
single emit (6 handlers): bun  313.08 ns/iter  (240.1 ns … 568.57 ns) 414.08 ns 551.74 ns 568.57 ns
single emit (7 handlers): bun  362.51 ns/iter (274.87 ns … 593.25 ns)  453.3 ns 572.99 ns 593.25 ns
single emit (8 handlers): bun  387.85 ns/iter (305.97 ns … 637.27 ns) 496.41 ns  587.6 ns 637.27 ns
on x 10_000 (5 handlers): bun   54.83 ms/iter   (40.74 ms … 70.82 ms)  65.17 ms  70.82 ms  70.82 ms
on x 10_000 (6 handlers): bun   64.24 ms/iter    (40.2 ms … 90.42 ms)  78.42 ms  90.42 ms  90.42 ms
on x 10_000 (7 handlers): bun   73.53 ms/iter   (45.6 ms … 101.21 ms)  86.55 ms 101.21 ms 101.21 ms
on x 10_000 (8 handlers): bun   84.14 ms/iter  (52.45 ms … 109.17 ms) 100.06 ms 109.17 ms 109.17 ms

handler.apply(this, args)          time (avg)             (min … max)       p75       p99      p995
--------------------------------------------------------------------- -----------------------------
single emit (5 handlers): bun  144.27 ns/iter (119.34 ns … 617.92 ns) 131.89 ns 376.44 ns 401.73 ns
single emit (6 handlers): bun  183.05 ns/iter (163.72 ns … 441.73 ns) 173.29 ns 405.11 ns 423.69 ns
single emit (7 handlers): bun  198.07 ns/iter (173.14 ns … 451.08 ns)  193.1 ns 408.01 ns 422.04 ns
single emit (8 handlers): bun  216.75 ns/iter (193.62 ns … 470.83 ns) 203.98 ns 448.03 ns 450.76 ns
on x 10_000 (5 handlers): bun   37.06 ms/iter   (20.85 ms … 52.62 ms)  45.94 ms  52.62 ms  52.62 ms
on x 10_000 (6 handlers): bun   43.45 ms/iter    (24.4 ms … 61.58 ms)  52.84 ms  61.58 ms  61.58 ms
on x 10_000 (7 handlers): bun   47.95 ms/iter   (29.53 ms … 67.53 ms)  58.41 ms  67.53 ms  67.53 ms
on x 10_000 (8 handlers): bun   54.75 ms/iter   (33.73 ms … 74.82 ms)  65.29 ms  74.82 ms  74.82 ms
```

yschroe:
> this is really cool to see optimized.
> 
> > Unsure about the "...committed the transpiled changes" part. I have 4 changed files in my src/js/out folder, do I need to commit them all? WebCoreJSBuiltins.{cpp||d.ts|h} and InternalModuleRegistryConstants.h.
> 
> Only `InternalModuleRegistryConstants.h` should have changed. Please commit that one please.

After the initial `make setup` I had all those files marked as changed in the source control tab in VS code (or after any of the numerous `make js`, can't fully remember). I have commited `InternalModuleRegistryConstants.h` now.

Also, as another benchmark, here is a real-world example using the readline module to read a file with ~10 million rows and counting the rows in the callback:
```
bun v1.0.2
line reader with callback function counting the lines
--> Read 10405422 lines in 2.518s (4132415 lines/s).

bun v1.0.2_debug
line reader with callback function counting the lines
--> Read 10405422 lines in 2.071s (5024347 lines/s).
```

yschroe:
Anything left to do here?

cyfung1031:
Might I ask, why have to create a copy of `handlers`? (`maybeClonedHandlers`)

`.forEach` is not an option?

yschroe:
> Might I ask, why have to create a copy of `handlers`? (`maybeClonedHandlers`)
> 
> `.forEach` is not an option?

It's because the handlers array could be modified while looping over it, if a handler is removed/added via a callback. You can check out [this discussion](https://github.com/oven-sh/bun/pull/5485#discussion_r1328285276) above.

jdalton:
@yschroe ~~LGTM.~~

We have some encapsulation issues creeping in, in the JS side of things but that could be handled in a separate PR.

For example the listeners is a user-land array so we can't trust that `.slice` isn't monkey-patched. Node avoids this using things like `ReflectApply(ArrayProtoSlice, handlers, [])`.

Use of `ArrayPrototypeSlice` is great but use of `ArrayPrototypeSlice.call` is not as it lends to monkey patching on `Function.prototype.call`. We use do something like:
```js
const ReflectApply = Reflect.apply;
const ArrayPrototypeSlice = Array.prototype.slice;

// later
ReflectApply(ArrayPrototypeSlice, theArray, []);
```

src/js/node/events.js
jdalton, line 98:
```suggestion
  const clonedHandlers = handlers.slice();
```
The user can always modify the handlers. We should always slice.

jdalton:
Spoke too soon on the LGTM. Minor nits left.

jdalton:
Use of `ArrayPrototypeSlice` is great but use of `ArrayPrototypeSlice.call` is not as it lends to monkey patching on `Function.prototype.call`. We use do something like:
```js
const ReflectApply = Reflect.apply;
const ArrayPrototypeSlice = Array.prototype.slice;

// later
ReflectApply(ArrayPrototypeSlice, theArray, []);
```

src/js/node/events.js
yschroe: In which case would this be an issue? If we have just one handler attached and this handler is called during the for loop (over the array with length 1), it can either
1) remove itself from the `handlers` array, e.g. if it was attached via .once(). -> No problem since no following handlers need to be called which would be affected by the change in array length.
2) or add additional handlers to the `handlers` array -> also not an issue in my opinion, since the newly added handlers are not called due to the for loop of length 1.
3) or modify itself somehow -> ~same as for 2) I would assume those modifications should only be considered the next time the emit is triggered.~ Actually, using `slice` here would result in no difference since this creates a shallow copy only anyways. So the only difference between `slice` and using the original array is that we need to be aware of array length changes in the latter case which I think are no problem.

Am I missing something?

jdalton: Really good point @yschroe! Can you add a code comment above it explaining that for future "us" to know about that.

yschroe:
> @yschroe ~LGTM.~
> 
> We have some encapsulation issues creeping in, in the JS side of things but that could be handled in a separate PR.
> 
> For example the listeners is a user-land array so we can't trust that `.slice` isn't monkey-patched. Node avoids this using things like `ReflectApply(ArrayProtoSlice, handlers, [])`.
> 
> Use of `ArrayPrototypeSlice` is great but use of `ArrayPrototypeSlice.call` is not as it lends to monkey patching on `Function.prototype.call`. We use do something like:
> 
> ```js
> const ReflectApply = Reflect.apply;
> const ArrayPrototypeSlice = Array.prototype.slice;
> 
> // later
> ReflectApply(ArrayPrototypeSlice, theArray, []);
> ```

I think those changes should be done in a separate PR, as you suggested, as this PR is getting quite long already and `slice` was already used before.
And I never worked on this kind of robustness changes before so I have no experience in what functions need to be "safe-guarded" and how to do it exactly.

Appreciate the feedback nonetheless. :)

src/js/node/events.js
yschroe: > Really good point @yschroe! Can you add a code comment above it explaining that for future "us" to know about that.

Sure! Is it clear this way?
```
// clone handlers array if necessary since handlers can be added/removed during the loop.
// cloning can be skipped in the case of exactly one attached handler since array length
// changes have no side-effects in a for-loop of length 1
```

jdalton: Yes, thank you!
> // cloning the handlers array is necessary since handlers can be added/removed during the loop.
// cloning can be skipped in the case of exactly one attached handler since array length
// changes have no side-effects in a for-loop of length <= 1

jdalton:
✅

Electroid:
Thanks @yschroe and @jdalton for working together on this, tests pass and this is good to go.

PR #5493, opened 2023-09-15
Remove dead code due to change in format

spullara:
At some point it was decided that bun would only format using KB, MB, etc. vs also using KiB, MiB, etc. The code that enabled it was never removed.

### What does this PR do?
Removes dead code.

There were no changes to execution.


v3nntrnx:
LGTM

Issue #5508, opened 2023-09-15
Building zls failed by following the document


shinichy:
### What version of Bun is running?

1.0.2

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

Building zls failed by following the development document https://bun.sh/docs/project/development#zls.

```
git clone https://github.com/zigtools/zls
cd zls
git checkout f91ff831f4959efcb7e648dba4f0132c296d26c0
zig build
```

### What is the expected behavior?

zig build succeeds.

### What do you see instead?

```
❯ zig build
zig build-exe zls Debug native: error: the following command failed with 1 compilation errors:
/Users/shinichi/.bun/install/global/node_modules/@oven/zig-darwin-x64/zig build-exe /Users/shinichi/ghq/github.com/zigtools/zls/src/main.zig --cache-dir /Users/shinichi/ghq/github.com/zigtools/zls/zig-cache --global-cache-dir /Users/shinichi/.cache/zig --name zls --mod binned_allocator::/Users/shinichi/.cache/zig/p/1220363c7e27b2d3f39de6ff6e90f9537a0634199860fea237a55ddb1e1717f5d6a5/binned_allocator.zig --mod diffz::/Users/shinichi/.cache/zig/p/122089a8247a693cad53beb161bde6c30f71376cd4298798d45b32740c3581405864/DiffMatchPatch.zig --mod known-folders::/Users/shinichi/.cache/zig/p/122048992ca58a78318b6eba4f65c692564be5af3b30fbef50cd4abeda981b2e7fa5/known-folders.zig --mod build_options::/Users/shinichi/ghq/github.com/zigtools/zls/zig-cache/c/151f041d97bdd9419b1a223a71b32ddf/options.zig --deps build_options,known-folders,diffz,binned_allocator --listen=-
Build Summary: 1/4 steps succeeded; 1 failed (disable with --summary none)
install transitive failure
└─ install zls transitive failure
   └─ zig build-exe zls Debug native 1 errors
src/DocumentStore.zig:989:23: error: no field or member function named 'deinit' in 'zig.system.NativePaths'
    defer native_paths.deinit();
          ~~~~~~~~~~~~^~~~~~~
/Users/shinichi/.bun/install/global/node_modules/@oven/zig-darwin-x64/lib/std/zig/system/NativePaths.zig:1:1: note: struct declared here
const std = @import("../../std.zig");
^~~~~
referenced by:
[...]

PR #5510, opened 2023-09-15
Fix ZLS commit hash in the document

shinichy:
### What does this PR do?

Update the commit hash of `zls` in the document. This fixes https://github.com/oven-sh/bun/issues/5508.

- [ ] Code changes


Jarred-Sumner:
we can remove the zls download stuff from the docs

we just use latest now

shinichy:
You're right. The zls downloaded by the extension automatically works well. I've just removed the zls section from the doc.

Electroid (approved):

PR #5513, opened 2023-09-15
docs(development): typo which would lead to wrong llvm installation

sum117:
### What does this PR do?

- [ ] Code changes

Installed the correct version by changing the number.


paperclover:
thanks. we upgraded to llvm 16 some time ago and did not update this instruction

PR #5521, opened 2023-09-16
fix(bundler): Add a space before minified require

### LINKED ISSUE #5501: minify `require` for `target=bun` output error

Darwin 21.3.0 arm64 arm

```ts
/// foo.ts
export function load(ident: string) {
	return require(ident);
}
```

```sh
$  bun build foo.ts --minify --target=bun
```

minified syntax that's valid

```js
// @bun
function b(a){return import.meta.require(a)}export{b as load};
```

### What do you see instead?

note the `returnimport` (no space)

```js
// @bun
function b(a){returnimport.meta.require(a)}export{b as load};
```

### Additional information

_No response_

davidmhewitt:
Fixes #5501

Ensures `require` and `resolve` statements are treated as an identifier with a space before them, so you don't end up with `returnrequire` in the minified output.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/js_printer.zig
+++ b/src/js_printer.zig
@@ -2179,6 +2179,7 @@ fn NewPrinter(
                     }
                 },
                 .e_require_call_target => {
+                    p.printSpaceBeforeIdentifier();
                     p.addSourceMapping(expr.loc);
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you

Issue #4851, opened 2023-09-10
HTMLRewriter with `fetch` crashes the process


AndersCan:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.5.0 x86_64 i386

### What steps can reproduce the bug?

A HTMLRewriter that returns a `fetch` will fail with `Error: Cork buffer must not be held across event loop iterations!`

```ts
const rewriter = new HTMLRewriter();

rewriter.on("foo", {
  element(el) {
    return fetch("https://bun.sh/").then(() => console.log("Will never run"))
  },
});


const server = Bun.serve({
  fetch(req: Request) {
    return rewriter.transform(new Response(`
        <h1>Heading 1</h1>
       <foo></foo>
    `, {
      headers: {"content-type": "text/html"}
    }));
  },
});

console.log(`Listening on http://localhost:${server.port}`);
```

### Additional information
Crash report:
```
uh-oh: A C++ exception occurred
bun will crash now 😭😭😭

----- bun meta -----
Bun v1.0.0 (822a00c4) macOS x64 22.5.0
RunCommand: 
Elapsed: 966ms | User: 34ms | Sys: 15ms
RSS: 18.29MB | Peak: 18.29MB | Commit: 34.61MB | Faults: 0
----- bun meta -----

0   0x107a63850 WTFGetBacktrace

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord
```

gtrabanco:
Is this related to #4418 ? Have you tried with the workaround?

PR #5525, opened 2023-09-16
fix(corking) uncork if needed

cirospaciari:
### What does this PR do?
Fix: [4851](https://github.com/oven-sh/bun/issues/4851)

- [x] Code changes

### DIFF
--- a/packages/bun-uws/src/Loop.h
+++ b/packages/bun-uws/src/Loop.h
@@ -58,12 +58,6 @@ private:
         for (auto &p : loopData->postHandlers) {
             p.second((Loop *) loop);
         }
-
-        /* After every event loop iteration, we must not hold the cork buffer */
-        if (loopData->corkedSocket) {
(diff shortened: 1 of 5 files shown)


packages/bun-uws/src/Loop.h
```diff
@@ -58,12 +58,6 @@ struct Loop {
         for (auto &p : loopData->postHandlers) {
             p.second((Loop *) loop);
         }
-
```
cirospaciari, line 61: Mental notes: another solution is uncorking here but is unlikely to be a problem because we always use `HttpResponse *cork(MoveOnlyFunction<void()> &&handler)` on HttpResponse and `void cork(MoveOnlyFunction<void()> &&handler)` on WebSocket.

`this.global.bunVM().waitForPromise(promise);` can trigger this condition.

Jarred-Sumner:
nice

Issue #5471, opened 2023-09-15
ignoreBOM property not initialized on TextDecoder


milkmuffin:
### What version of Bun is running?

1.0.1

### What platform is your computer?

Linux d31459b29751 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 GNU/Linux

### What steps can reproduce the bug?

```js
new TextDecoder('utf-8').ignoreBOM;
new TextDecoder('utf-8', { ignoreBOM: false}).ignoreBOM;
new TextDecoder('utf-8', { ignoreBOM: true}).ignoreBOM;
`ignoreBOM` in new TextDecoder();
```

### What is the expected behavior?

`ignoreBOM` property should be `false`

### What do you see instead?

`ignoreBOM` property does not exist (and thus evaluates as undefined), regardless of how it's initialized.

Timelord65:
Should be related to https://github.com/oven-sh/bun/issues/5211

ashutosh887:
@milkmuffin @WingLim 
Let me try this issue... Please assign

Electroid:
@ashutosh887 Feel free to submit a pull request, you don't need to be assigned to work on it.

ashutosh887:
Is this issue accepted? @Electroid

Electroid:
Yes, but also, there is already a pending pull request in #5550. So if you would like to contribute, I'd pick another issue.

PR #5550, opened 2023-09-16
feat(encoding): support BOM detection

WingLim:
### What does this PR do?

Support BOM detection and `TextEncoder.ignoreBOM`.

Closes https://github.com/oven-sh/bun/issues/5471

- [x] Code changes

- [x] Add test

### DIFF
--- a/src/bun.js/webcore/encoding.zig
+++ b/src/bun.js/webcore/encoding.zig
@@ -559,6 +559,13 @@ pub const TextDecoder = struct {
                     remainder = remainder[1..];
                     continue;
                 },
+                // BOM handling
+                0xFEFF => {
+                    buffer.ensureTotalCapacity(allocator, 1) catch unreachable;
+                    buffer.items.ptr[buffer.items.len] = remainder[0];
+                    buffer.items.len += 1;
+                    remainder = remainder[1..];
+                },
 
                 // Is this an unpaired low surrogate or four-digit hex escape?
                 else => {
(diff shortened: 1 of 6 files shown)


test/js/web/encoding/text-decoder.test.js
```diff
@@ -250,7 +250,7 @@ describe("TextDecoder", () => {
   it("constructor should set values", () => {
     const decoder = new TextDecoder("utf-8", { fatal: true, ignoreBOM: false });
     expect(decoder.fatal).toBe(true);
-    // expect(decoder.ignoreBOM).toBe(false); // currently the getter for ignoreBOM doesn't work and always returns undefined
+    expect(decoder.ignoreBOM).toBe(false);
```
Jarred-Sumner, line 253: we haven't actually implemented BOM detection

WingLim: Oh, okay, let me try to implement the BOM detection in this PR, I will convert this to draft for now.

src/bun.js/webcore/encoding.zig
```diff
@@ -559,6 +559,13 @@ pub const TextDecoder = struct {
                     remainder = remainder[1..];
                     continue;
                 },
+                // BOM handling
+                0xFEFF => {
```
Jarred-Sumner, line 563: This should only be in there for the first character right? This code runs on every character

WingLim: We handle the first character of BOM in `decodeSlice`.

In unicode `0xFEFF` is a special, we should not rewrite it to `0xFFFD` in else block, just keep it. This is to be consistent with node.

Jarred-Sumner:
Thank you

PR #5572, opened 2023-09-16
feat(test): Implement `arrayContaining`

WingLim:
### What does this PR do?

Implement `arrayContaining`, releated to https://github.com/oven-sh/bun/issues/1825
Furthermore, the algorithm to caclute array containing is `O(m * n)`,  there should have optimisation for it.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -253,7 +253,9 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
 
                     for (unsigned n = 0; n < otherLength; n++) {
                         JSValue otherValue = otherArray->getIndex(globalObject, n);
-                        if (JSValue::strictEqual(globalObject, expectedValue, otherValue)) {
+                        ThrowScope scope = DECLARE_THROW_SCOPE(globalObject->vm());
+                        Vector<std::pair<JSValue, JSValue>, 16> stack;
+                        if (Bun__deepEquals<false, true>(globalObject, expectedValue, otherValue, stack, &scope, true)) {
                             found = true;
                             break;
                         }
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -248,11 +248,11 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
 
                 // O(m*n) but works for now
                 for (unsigned m = 0; m < expectedLength; m++) {
-                    JSValue expectedValue = expectedArray->get(globalObject, m);
+                    JSValue expectedValue = expectedArray->getIndex(globalObject, m);
                     bool found = false;
 
                     for (unsigned n = 0; n < otherLength; n++) {
-                        JSValue otherValue = otherArray->get(globalObject, n);
+                        JSValue otherValue = otherArray->getIndex(globalObject, n);
                         if (JSValue::strictEqual(globalObject, expectedValue, otherValue)) {
                             found = true;
                             break;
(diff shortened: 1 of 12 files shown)


src/bun.js/bindings/bindings.cpp
```diff
@@ -229,6 +229,45 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
             }
         }
 
+        return AsymmetricMatcherResult::FAIL;
+    } else if (auto* expectArrayContaining = jsDynamicCast<JSExpectArrayContaining*>(matcherPropCell)) {
+        JSValue expectedArrayValue = expectArrayContaining->m_arrayValue.get();
+
+        if (otherProp.isCell() && otherProp.asCell()->type() == ArrayType) {
```
Jarred-Sumner: There are a few other types of arrays this could be, like `DerivedArray` (a subclass of Array)

Instead, for checking if a JSValue is an array, we should use `JSC::isJSArray`

src/bun.js/bindings/bindings.cpp
```diff
@@ -229,6 +229,45 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
...
+                unsigned otherLength = otherArray->length();
+
+                // A empty array is all array's subset
+                if (expectedLength == 0) {
+                    return AsymmetricMatcherResult::PASS;
+                }
+
+                // O(m*n) but works for now
+                for (unsigned m = 0; m < expectedLength; m++) {
+                    JSValue expectedValue = expectedArray->get(globalObject, m);
```
Jarred-Sumner: Instead of calling `get`, we should call `getIndex`

```suggestion
                    JSValue expectedValue = expectedArray->getIndex(globalObject, m);
```

src/bun.js/bindings/bindings.cpp
Jarred-Sumner: ah turns out that method actually does the same thing

we should still use it though incase JSC makes a change in the future

src/bun.js/bindings/bindings.cpp
```diff
@@ -229,6 +229,45 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
...
+                    bool found = false;
+
+                    for (unsigned n = 0; n < otherLength; n++) {
+                        JSValue otherValue = otherArray->get(globalObject, n);
```
Jarred-Sumner:
```suggestion
                        JSValue otherValue = otherArray->getIndex(globalObject, n);
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -229,6 +229,45 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
...
+                        if (JSValue::strictEqual(globalObject, expectedValue, otherValue)) {
```
Jarred-Sumner: Are you sure it's `strictEqual` and not loose/deep equal?

Jarred-Sumner:
Thank you for this. Some comments but it's close to mergeable

src/bun.js/bindings/bindings.cpp
WingLim: I check jest code, that should use deep equal, fix it already.

Jarred-Sumner: The other thing: arrays can have holes in them. This needs to check if `otherValue.isEmpty()` (it is represented as `0`, which is separate from `undefined` or `null`)

Jarred-Sumner:
thank you!

PR #5593, opened 2023-09-16
Docs: path aliases fix

e253:
### What does this PR do?

Change string:string mapping in the bun docs to match the expected string:array[string] mapping.

See: https://www.typescriptlang.org/tsconfig#paths

- [ ] Code changes


Jarred-Sumner:
Thanks

PR #5595, opened 2023-09-16
feat(console.log): Print anonymous when class name is unknown

JibranKalia:
### What does this PR do?

This matches the functionality in Node. when printing a class that doesn't have a name.

Previously this code:
```js

console.log(class { });

``` 

Would print:
```
[class]
```

Updated to print matching with Node:
```
[class (anonymous)]
```

- [X] Code changes


Jarred-Sumner (approved):
LGTM. Thank you

PR #5600, opened 2023-09-17
Update README.md for older cpu arch solution

Adamkaram:
here is just hint for older cpu to fix dump Illegal instruction (core dumped) issue

### What does this PR do?

- [ ] Code changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ cp -r misc /usr/local/bin/
 ```
 ```2
 echo "alias bun='sde -chip-check-disable -- bun'" >> ~/.bashrc
-
+```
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
We aren't going to recommend people to use a CPU emulator, as that is a 1000x slowdown. Instead, we can reduce the minimum CPU requirement from nehalem to something a little older.

Adamkaram:
> We aren't going to recommend people to use a CPU emulator, as that is a 1000x slowdown. Instead, we can reduce the minimum CPU requirement from nehalem to something a little older.

older than core 2 duo ?

Issue #5604, opened 2023-09-17
`performance.now()` returns `NaN` when called extremely fast


jack-weilage:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49
1.0.3+0a318ecca14ab1e71aa60ae8eadbd16343b3a3ea

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64
Linux 6.2.0-32-generic x86_64 x86_64

### What steps can reproduce the bug?

1. Create a file with a loop checking `performance.now()`

```ts
let i = 0
while (true) {
  const now = performance.now()

  if (Number.isNaN(now)) {
    console.log(i)
  }

  i++
}
```

2. Run with `bun run file.js`

### What is the expected behavior?

The loop should never log anything, as `performance.now()` should never return `NaN`.

When run in the browser or via `node bug.js`, the loop continues forever.

### What do you see instead?

```bash
bun run bug.js
932599
2521622
8352525
9482475
12140024
16258548
16422489
16584413
18517983
19007226
20344855
20887794
26493524
27206784
28011873
28498102
29000819
31453228
37471626
^C
```

The loop logs once every ~1m calls on average, with a max of ~10m calls and min of ~100k calls.

### Additional information

This bug ruins benchmarks which use `performance.now()`.

Jarred-Sumner:
Likely a DOMJIT issue, potentially with FTL.

pan93412:
> Likely a DOMJIT issue, potentially with FTL.

Unreproducible in Safari Technology Preview, so I suppose that this issue is not from WebKit.

![CleanShot 2023-09-17 at 18 33 45@2x](https://github.com/oven-sh/bun/assets/28441561/28cf74e9-ae2d-4a44-86e4-e68b01b3f256)

I added the `std::cout` debug statement in `functionPerformanceNowBody()`:

![CleanShot 2023-09-17 at 18 35 14@2x](https://github.com/oven-sh/bun/assets/28441561/816d6e77-04d8-4052-8340-d0dc0aab295a)

And I found that it only triggers when the `time` from `Bun__readOriginTimer(global->bunVM())` is divisible by `1000000`:

<details>
<summary>Code for test</summary>

```js
for (let i = 0; i < 11451400; i++) {
    const now = performance.now();
    if (Number.isNaN(now)) {
      console.log("Triggered")
    }
}
```

</details>

[...]

PR #5615, opened 2023-09-17
Workaround #5604

Jarred-Sumner:
### What does this PR do?

This works around #5604.

Disable DOMJIT for performance.now(). Makes performance.now() a little slower

There is a test


dylan-conway:
I think we should use `jsDoubleNumber` in `functionPerformanceNowBody` and keep DOMJIT enabled
https://github.com/oven-sh/bun/issues/5604#issuecomment-1722544211

PR #5616, opened 2023-09-17
fix(test): when tests run with --only the nested describe blocks `.on…

igorshapiro:
### What does this PR do?

Today adding the `.only` specifier to a nested `describe` block will not run it, even if launched with the `--only` switch.
This PR fixes this behavior (the following code now works):

```ts
import { test, describe } from "bun:test";
describe("outer", () => {
  describe.only("inner (nested)", () => {
    test("test", () => {
      console.error("reachable");
    })
  })
})
```

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -1610,7 +1610,7 @@ inline fn createScope(
             .parent = parent,
             .file_id = parent.file_id,
             .tag = if (parent.is_skip) parent.tag else tag,
-            .is_skip = is_skip or parent.is_skip,
+            .is_skip = is_skip and parent.is_skip,
         };
 
         return scope.run(globalThis, function, &.{});
--- a/src/bun.js/test/jest.zig
+++ b/src/bun.js/test/jest.zig
@@ -793,14 +793,25 @@ pub const DescribeScope = struct {
     current_test_id: TestRunner.Test.ID = 0,
     value: JSValue = .zero,
     done: bool = false,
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you for this

Looks like test-test is failing

src/bun.js/test/jest.zig
```diff
@@ -1610,7 +1610,7 @@ inline fn createScope(
             .parent = parent,
             .file_id = parent.file_id,
             .tag = if (parent.is_skip) parent.tag else tag,
-            .is_skip = is_skip or parent.is_skip,
+            .is_skip = is_skip and parent.is_skip,
```
dylan-conway: This will break skip tests because if the parent scope is marked as skip, the child scope should also be skipped. These are the test failures we are seeing in `test-test.test.ts`

igorshapiro: Makes sense. I thought to refactor it to 2 separate flags - `is_skip` and `is_only` that will be aggregated throughout the traversal. And then we can be more precise in the decisions. 
Because the `is_skip` flag alone is not sufficient to handle this logic. Sounds 🆗  to you?

igorshapiro: You're right. Refactored the code to avoid the complexity around maintaining `is_skip` in `--only` and non-`--only` modes. 
The skip tests now pass

Electroid:
Thanks for working on this! Running the test suite one more time

PR #5620, opened 2023-09-17
Added react installation to react.md

jt3k:
Added installation line to react guide

I just paste the commands from the manual after a clean install of bun

without installing react i got an error
```
error: Cannot find module "react/jsx-dev-runtime" from "/Users/jt3k/tmp/bun/react.tsx"
```


colinhacks:
Thanks! I took the opportunity to flesh out this guide a little bit as well.

PR #5628, opened 2023-09-17
Added .DS_Store to gitignore-for-init

Cilooth:
### What does this PR do?

This PR adds .DS_Store files, which are automatically generated by macOS's Finder application, to the default .gitignore generated by the bun init command to prevent them from being inadvertently committed.

- [x] Code changes

This is a simple update to the gitignore-for-init file to comply with OSS best-practices.

### DIFF
--- a/src/cli/gitignore-for-init
+++ b/src/cli/gitignore-for-init
@@ -170,3 +170,7 @@ dist
 
 # IntelliJ based IDEs
 .idea
+
+# Finder (MacOS) folder config
+.DS_Store
+


gtrabanco:
User decisions (like system or editor) should be in local `~/.gitignore_global` and project decisions (like which is the build dir) in project `.gitignore`.

Source:
- https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

uncenter:
But you can't really expect every new Git user to have that set up... much better for everyone to include it by default.

Bun itself has `.DS_Store` in the `.gitignore`:
https://github.com/oven-sh/bun/blob/639a12f59fc5342360c341bc9858244c107cc30e/.gitignore#L1

gtrabanco:
> But you can't really expect every new Git user to have that set up... much better for everyone to include it by default.
> 
> Bun itself has `.DS_Store` in the `.gitignore`:
> 
> https://github.com/oven-sh/bun/blob/639a12f59fc5342360c341bc9858244c107cc30e/.gitignore#L1

You are not dealing with users in Github you are dealing with programmers which should learn how to use right their tools. Of course you can't expect every dev to know everything but they should be open to learn how to use the tools right.

I do not expect a building worker to use all the tools right in the first day even with experience but if at some point he has to use a machine he should learn how to use it right.

And thats why all code is reviewed and we have discussions here, to explain, teach and correct mistakes like files that should not be in a PR.

Of course at the end is up to whoever is in charge to approve a PR if he/she approve with mistakes or won't, I mean files that should not be there in this case.

Jarred-Sumner:
Thank you



> User decisions (like system or editor) should be in local `~/.gitignore_global` and project decisions (like which is the build dir) in project `.gitignore`.
> 
> Source:
> 
> * https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

This would be nice, but it's also harmless to include `.DS_Store` in the default `.gitignore` file.

gtrabanco:
> Thank you
> 
> > User decisions (like system or editor) should be in local `~/.gitignore_global` and project decisions (like which is the build dir) in project `.gitignore`.
> > Source:
> > 
> > * https://gist.github.com/subfuzion/db7f57fff2fb6998a16c
> 
> This would be nice, but it's also harmless to include `.DS_Store` in the default `.gitignore` file.

Yes totally agree and in this project is more useful to have it. But I hope if any came here to this PR and read this learn something as I have learned before in a PR. =)

It is useful and you can have a files that you use in your projects like a `playground.ts` to do some checks when you have doubts (as example of useful use case).

Cheers!

PR #5660, opened 2023-09-18
Remove hardcoded references to zig in Makefile

xbjfk:
### What does this PR do?
Instead of zig being hardcoded in some places in the Makefile, i made them use the ZIG variable. 

- [X] Code changes

I observed `make run-all-unit-tests` works with ZIG= set but zig not in path.


Jarred-Sumner:
Thank you

Issue #4589, opened 2023-09-08
bun install fails if private registry does not ends with /


carneiror:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```
[install]
registry = "https://<private-registry>/api/npm"
```

### What do you see instead?

```
> bun install
error: package "rxjs" not found <private-registry>/api/rxjs 404
```

### Additional information

If I change my registry config to `https://<private-registry>/api/npm/` everything works as expected. Should be a minor issue when concating a URL with pathname.

Issue #5368, opened 2023-09-14
Bun install doesn't work with a registry url that don't end with `/`


atombarel:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Linux 5.15.90.3-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

- create a bunfig.toml with a registry URL without a slash as last character
- run command bun install

### What is the expected behavior?

- all packages get installed 

### What do you see instead?

- all packages result in a 404 error

### Additional information

Maybe related to a past bug #2462

Hanaasagi:
Related to https://github.com/oven-sh/bun/issues/4589

Jarred-Sumner:
Fixed by @Hanaasagi

PR #5662, opened 2023-09-18
fix(config): support for registry url without trailing slash

Hanaasagi:
### What does this PR do?

Close: #4589, #5368

https://github.com/oven-sh/bun/blob/57e38e831222add628b25e72cbcaff84c71c4a4c/src/install/install.zig#L254-L259

If the registry url does not end with a `/`, it will result in the wrong url when joining. For example, joining `http://example.com/aaa` and `package` results in `http://example.com/package` instead of `http://example.com/aaa/package`

This PR fixes this by adding a trailing slash during the config parsing to ensure consistency.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bunfig.zig
+++ b/src/bunfig.zig
@@ -70,7 +70,7 @@ pub const Bunfig = struct {
                         registry.password = url.password;
                         registry.url = try std.fmt.allocPrint(this.allocator, "{s}://{s}/{s}/", .{ url.displayProtocol(), url.displayHostname(), std.mem.trim(u8, url.pathname, "/") });
                     } else {
-                        if (std.mem.endsWith(u8, url.href, "/")) {
+                        if (strings.hasSuffixComptime(url.href, "/")) {
                             registry.url = url.href;
                         } else {
                             registry.url = try std.fmt.allocPrint(this.allocator, "{s}/", .{url.href});
--- a/src/bunfig.zig
+++ b/src/bunfig.zig
@@ -80,7 +80,7 @@ pub const Bunfig = struct {
                 .e_object => |obj| {
                     if (obj.get("url")) |url| {
                         try this.expect(url, .e_string);
-                        if (std.mem.endsWith(u8, url.data.e_string.data, "/")) {
+                        if (strings.hasSuffixComptime(url.data.e_string.data, "/")) {
                             registry.url = url.data.e_string.data;
                         } else {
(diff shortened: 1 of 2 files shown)


src/bunfig.zig
```diff
@@ -64,19 +64,27 @@ pub const Bunfig = struct {
...
-                        registry.url = try std.fmt.allocPrint(this.allocator, "{s}://{s}/{s}", .{ url.displayProtocol(), url.displayHostname(), std.mem.trimLeft(u8, url.pathname, "/") });
+                        registry.url = try std.fmt.allocPrint(this.allocator, "{s}://{s}/{s}/", .{ url.displayProtocol(), url.displayHostname(), std.mem.trim(u8, url.pathname, "/") });
                     } else if (url.username.len > 0 and url.password.len > 0) {
                         registry.username = url.username;
                         registry.password = url.password;
-                        registry.url = try std.fmt.allocPrint(this.allocator, "{s}://{s}/{s}", .{ url.displayProtocol(), url.displayHostname(), std.mem.trimLeft(u8, url.pathname, "/") });
+                        registry.url = try std.fmt.allocPrint(this.allocator, "{s}://{s}/{s}/", .{ url.displayProtocol(), url.displayHostname(), std.mem.trim(u8, url.pathname, "/") });
                     } else {
-                        registry.url = url.href;
+                        if (std.mem.endsWith(u8, url.href, "/")) {
```
Jarred-Sumner:
```suggestion
                        if (strings.hasSuffixComptime(url.href, "/")) {
```

src/bunfig.zig
```diff
@@ -64,19 +64,27 @@ pub const Bunfig = struct {
...
+                        } else {
+                            registry.url = try std.fmt.allocPrint(this.allocator, "{s}/", .{url.href});
+                        }
                     }
                 },
                 .e_object => |obj| {
                     if (obj.get("url")) |url| {
                         try this.expect(url, .e_string);
-                        registry.url = url.data.e_string.data;
+                        if (std.mem.endsWith(u8, url.data.e_string.data, "/")) {
```
Jarred-Sumner:
```suggestion
                        if (strings.hasSuffixComptime(url.data.e_string.data, "/")) {
```

Jarred-Sumner:
Thank you

PR #5679, opened 2023-09-18
fix: provide empty string to 0 length process environment variables

liz3:
### What does this PR do?

For process loaded env vars, its a bug to give them the literal value '""' if the provided length is 0.
I am not entirely sure if this is expected behaviour for DotEnv Files, which is why i left the constant unmodified and opted to only modify the process based value, to which is should apply.

If the current behaviour is also faulty for dotenv, changing the `empty_string_value` might be a better approach

- [x] Code changes

Compiled/ran a test example to verify the behaviour, can add tests if needed.


src/env_loader.zig
```diff
@@ -361,7 +361,7 @@ pub const Loader = struct {
                     if (value.len > 0) {
                         this.map.put(key, value) catch unreachable;
                     } else {
-                        this.map.put(key, empty_string_value) catch unreachable;
+                        this.map.put(key, "") catch unreachable;
```
paperclover: if it's just going to be an empty string, we no longer need the if branch, as `value` will be an empty string if it's zero length.

also, can you check if `empty_string_value` is used anywhere else, and if not, remove it too

paperclover (changes requested):
thanks for the one line fix, though we can make the code in this function a little simpler if we arent going to use an empty string.

also can you add a test case to `env.test.ts`, probably copying the existing test `.env with zero length strings`... somehow that test already asserts this behavior but in a different case.

src/env_loader.zig
liz3: `empty_string_value` is used in quite a few places

liz3:
> thanks for the one line fix, though we can make the code in this function a little simpler if we arent going to use an empty string.
> 
> also can you add a test case to `env.test.ts`, probably copying the existing test `.env with zero length strings`... somehow that test already asserts this behavior but in a different case.

I added a test for the process based case, i want to clarify again that im unsure whether its expected behaviour for DotEnv files to return a quoted string or they too should behave like this.
If thats the case we could change the `empty_string_value` constant to BE a empty string literal rather then only for this case.

paperclover:
> I want to clarify again that im unsure whether its expected behaviour for DotEnv files to return a quoted string or they too should behave like this.

Seems in node.js they will not quote the environment variable.

image

I don't know why we ever did this, tried to find any examples where this is done.

> If thats the case we could change the empty_string_value constant to BE a empty string literal rather then only for this case.

We should just remove that constant. the only usages are done in `if len == 0` checks, besides this following one:

image

for this case let's put `""` inline, which will translate to an empty js string.

liz3:
> > I want to clarify again that im unsure whether its expected behaviour for DotEnv files to return a quoted string or they too should behave like this.
> 
> Seems in node.js they will not quote the environment variable.
> image
> 
> I don't know why we ever did this, tried to find any examples where this is done.
> 
> > If thats the case we could change the empty_string_value constant to BE a empty string literal rather then only for this case.
> 
> We should just remove that constant. the only usages are done in `if len == 0` checks, besides this following one:
> image
> 
> for this case let's put `""` inline, which will translate to an empty js string.

Changes are implemented, can confirm that the behaviour is consistent now!
image

paperclover (approved):
waiting for tests to run now

liz3:
Format check on `env.test.ts` failed, committed the correction.

liz3:
(the failing of prettier in unrelated to the code now)

PR #5684, opened 2023-09-18
fix: remove unneeded branch in toJSONWithBytes

liz3:
### What does this PR do?
This branch resulted in the same statement independent of the conditions value so having the branch is not needed

- [ ] Code changes


Jarred-Sumner:
Thank you

Issue #5683, opened 2023-09-18
Support disable telemetry through `bunfig.toml`


opiation:
### What is the problem this feature would solve?

Currently, the means of disabling telemetry involve add a `DO_NOT_TRACK=1` environment variable wherever bun is used. This is somewhat inflexible requiring users to either commit `.env` to version control or explicitly add the env var to their personal setup to opt out rather than requiring an explicit opt-in.

### What is the feature you are proposing to solve the problem?

Add some telemetry toggling setting to `bunfig.toml` so it can be version controlled without exposing other env vars. Also, disable telemetry by default or emit a warning on install / first use that some telemetry is being reported.

### What alternatives have you considered?

- forking `bun` to disable telemetry by default
  This is too heavyweight and unnecessarily divisive at this time
- submitting a PR to allow this configurability
  This is possible but might be a lot of effort as I'm not familiar with Zig nor any of the other source code for Bun
- documenting in every project's `README` that they should add this variable to disable telemetry before using bun
  This is fine but error prone as many contributors tend not to read such details in docs
- committing some `.env` with `DO_NOT_TRACK=1` to version control and moving sensitive vars to `.env.local` or something like that
  This is feasible but undesirable since Bun would currently be the only reason to require this change for many projects I contribute to.

opiation:
Thanks for implementing this so quickly!

PR #5690, opened 2023-09-18
feat: switch disableTelemetry to bunfig

coratgerl:
### What does this PR do?
Closes: https://github.com/oven-sh/bun/issues/5683

This PR switch ```disableTelemetry```to bunfig instead of .env.

- [x] Code changes

Manual tests

### DIFF
--- a/docs/runtime/env.md
+++ b/docs/runtime/env.md
@@ -75,3 +75,10 @@ These environment variables are read by Bun and configure aspects of its behavio
 
 - `FORCE_COLOR`
 - If `FORCE_COLOR=1`, then ANSI color output is force enabled, even if `NO_COLOR` is set.
+
+---
+
+- `DO_NOT_TRACK`
+- If `DO_NOT_TRACK=1`, then analytics are [disabled](https://do-not-track.dev/). Bun records bundle timings (so we can answer with data, "is Bun getting faster?") and feature usage (e.g., "are people actually using macros?"). The request body size is about 60 bytes, so it's not a lot of data. Equivalent of `telemetry=false` in bunfig.
+
+{% /table %}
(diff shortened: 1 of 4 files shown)


liz3:
Maybe keeping the option to set it via a environment variable would be a better approach, so that both can be used, since while in a config it makes the most logical sense, having the ability to set it via a environment variable as a second option is a more "complete" approach.

Maybe there will be scenarios where it's supposed to be disabled in a configuration but a certain direct execution(shell) wants it enabled, arguably thats a edge case but the possibility does exist.

docs/runtime/env.md
```diff
@@ -75,10 +75,3 @@ These environment variables are read by Bun and configure aspects of its behavio
 
 - `FORCE_COLOR`
 - If `FORCE_COLOR=1`, then ANSI color output is force enabled, even if `NO_COLOR` is set.
-
----
-
-- `DO_NOT_TRACK`
```
Electroid, line 81: We should keep environment variable.

src/bunfig.zig
```diff
@@ -207,6 +207,11 @@ pub const Bunfig = struct {
                 if (json.get("preload")) |expr| {
                     try this.loadPreload(allocator, expr);
                 }
+
+                if (json.get("disableTelemetry")) |expr| {
```
Electroid: Instead of inverting the logic, we can just do `telemetry = false`

docs/runtime/env.md
coratgerl: Ok thank you, I will revert this :)

src/bunfig.zig
coratgerl: Thank you for your feedback, I will fix this :)

Electroid (approved):
Thanks!

PR #5718, opened 2023-09-19
Implement util.MIMEType and util.MIMEParams

gaurishhs:
### What does this PR do?

This implements util.MIMEType and util.MIMEParams

- [x] Code changes

### DIFF
--- a/completions/bun.fish
+++ b/completions/bun.fish
@@ -32,132 +32,153 @@ function __fish__get_bun_bun_js_files
 	string split ' ' (bun getcompletes j)
 end
 
-function bun_fish_is_nth_token --description 'Test if current token is on Nth place' --argument-names n
-	set -l tokens (commandline -poc)
-	set -l tokens (string replace -r --filter '^([^-].*)' '$1' -- $tokens)
-	test (count $tokens) -eq "$n"
-end
-
-function __bun_command_count --argument-names n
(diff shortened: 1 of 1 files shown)


Xachiey (commented):
Maybe you shouldn't change the style of the codebase. 
Just focus on modifying the functionality.
:D

Jarred-Sumner:
Thank you for this 

util needs to be implemented more in native code. The amount of JavaScript in Bun is growing too much, which costs start time and execution time. Bun already has an internal implementation of something very much like util.MIMEType. It's missing util.MIMEParams. If you want to attempt a Zig implementation of this, have a look in src/http/mime_type.zig. There are also WebKit implementaitons of Content-Type header parsing which might be helpful here.

gaurishhs:
@Jarred-Sumner i understand that, do you have any examples of any built in modules which are completely in zig rn? 

EDIT: i saw the dns implementation, so i should probably implement it in zig in (src/bun.js) and then declare a property util on the Bun global ( `Bun.util`) and then call each zig function in js? there should be something to directly import zig in js files something similar to what hyperimport did but that was a bun plugin..

gaurishhs:
> Maybe you shouldn't change the style of the codebase. Just focus on modifying the functionality. :D

i didn't do the formatting changes they are probably due to `bun fmt` which i ran as said when opening the PR

Issue #5668, opened 2023-09-18
`content-encoding` is removed from response


jimmywarting:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

```js
fetch('https://httpbin.org/gzip').then(async r => {
  const json = await r.json()
  console.log('accept-encoding', new Headers(json.headers).get('accept-encoding'))
  console.log('response', r.headers.get('content-encoding'))
})
```

### What is the expected behavior?

expected `content-encoding` to be `gzip`

### What do you see instead?

`null`

liz3:
This peeked my interest but i have yet to find where this is done(where the header would be removed from the map), will tinker with this tmrw, seams like another good issue to get to know the codebase!

jimmywarting:
I think this can have something to do with CORS request, 
browser do this too if you fetch something from another origin. 

I think the solution could be to treat any request as if it was **not** a CORS request

liz3:
> I think this can have something to do with CORS request, browser do this too if you fetch something from another origin.
> 
> I think the solution could be to treat any request as if it was **not** a CORS request

A server side process does not have anything within the concept of a "origin" to begin with since theres no origin website.
Iirc there is cors policies for headers like that but they should not apply on server side processes, cors is by definition a browser concept.

My theory is that the header is not copied somewhere but im going to research this now.

liz3:
https://github.com/oven-sh/bun/pull/5743

PR #5743, opened 2023-09-19
fix: don't remove content-encoding header from header table

liz3:
### What does this PR do?
It seams the 'content-encoding' header was being removed(if present), to run code conditionally, that had the unwanted side effect of the header not being present in the list returned to the caller. This instead uses a bool to keep track of the state regarding the state whether the header was prcoessed or not.
Maybe this has some other side effects but tests(fetch.test.ts) seams to not have any failures.

Fixes https://github.com/oven-sh/bun/issues/5668

- [x] Code changes

I wrote automated tests


jimmywarting:
If #5648 has been resolved first then i would have suggested an other approach 😉

jimmywarting:
lgtm

Electroid:
Thanks for investigating and fixing this, lgtm!

PR #5751, opened 2023-09-19
Windows install with wsl

LeonardoZivieri:
With a simple bat file, you can configure the command bun use directly wsl to run your env

### What does this PR do?

In docummentation, I put  an example of how you can set your bun to run on Windows with wsl, creating a bat file, adding the current folder to path and installing bun inside wsl.

- [ ] Code changes

Installed on my personal computer

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -48,6 +48,15 @@ The test runner, package manager, and bundler are still under development. The f
 - `bun link/unlink`
 - `bun build`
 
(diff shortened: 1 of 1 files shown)


paperclover:
bun on wsl's `/mnt/c` is slow and this encourages using it (by setting it to the windows path) instead of fully within wsl.

i don't think this is a pattern we should recommend.

PR #5770, opened 2023-09-19
bundler: default import_prefix to `./` if file_path does not start with a relative path signifier

### LINKED ISSUE #5700: Bun.build: `./` is omitted at the start of some imports.

```
git clone https://github.com/hex2f/bun-bugs
cd bun-bugs/5700
bun run index.ts
```
Then look at the imports in `out/src/a.js`

```js
import {
greet
} from "./greet.js";
```

### What do you see instead?

```js
import {
greet
} from "greet.js";
```

### Additional information

_No response_

hex2f:
### What does this PR do?

When generating import paths, the bundler now defaults `import_path` to `./` if no override is provided and if `bun.path.relative(from_chunk_dir, file_path)` does not result in a path with a relative path signifier (`../`). This fixes #5700.

- [ ] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be (*i am not sure how to do this properly*)

### DIFF
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -11002,11 +11002,7 @@ pub const Chunk = struct {
                                     else => unreachable,
                                 };
 
-                                const pre_norm_file_path = if (from_chunk_dir.len == 0) file_path else bun.path.relative(from_chunk_dir, file_path);
-                                const cheap_normalizer = cheapPrefixNormalizer(if (import_prefix.len == 0 and !strings.hasPrefixComptime(pre_norm_file_path, "../"))
-                                    "./"
-                                else
-                                    import_prefix, pre_norm_file_path);
+                                const cheap_normalizer = cheapPrefixNormalizer(import_prefix, if (from_chunk_dir.len == 0) file_path else bun.path.relative(from_chunk_dir, file_path));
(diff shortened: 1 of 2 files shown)


paperclover:
can you remove the git submodule you added, thanks.

paperclover:
`cheapPrefixNormalizer` is used in some other places, so those code paths would break even with this patch. I think the fix to this is changing that function to ensure the path starts with `./` or `../`

there seems to be some attempt to do this already in that function but it doesn't work.

i think theres alot more wrong with how we're doing this prefix stuff; `--public-path` doesn't work as it is already.

hex2f:
> `cheapPrefixNormalizer` is used in some other places, so those code paths would break even with this patch. I think the fix to this is changing that function to ensure the path starts with `./` or `../`
> 
> there seems to be some attempt to do this already in that function but it doesn't work.
> 
> i think theres alot more wrong with how we're doing this prefix stuff; `--public-path` doesn't work as it is already.

i've moved it to `cheapPrefixNormalizer` instead and all tests pass.

i don't know about `--public-path` as a cli option, but it seems to work when using `Bun.build()`

Jarred-Sumner:
Closing due to bitrot

Issue #5756, opened 2023-09-19
`bun:sqlite`: segfault when calling `.loadExtension()` when `sqlite3_load_extension` is not available


asg017:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 21.4.0 x86_64 i386

### What steps can reproduce the bug?

My `package.json`: 

```json
{
  "name": "bun-sqlite",
  "module": "index.ts",
  "type": "module",
  "devDependencies": {
    "bun-types": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  },
  "dependencies": {
    "sqlite-regex": "^0.2.3"
  }
}
```

In `index.ts`:

```ts
import * as sqlite_regex from "sqlite-regex";
import { Database } from "bun:sqlite";

const db = new Database(":memory:");
db.loadExtension(sqlite_regex.getLoadablePath());

const query = db.query("select regex_version();");
console.log(query.get());
```

### What is the expected behavior?

It should not segfault, and instead throw an error like "extension loading isn't supported on this SQLite build" or something. 

### What do you see instead?

```
$ bun run index.ts 
Segmentation fault: 11
```

### Additional information

This happens because the default sqlite3 library on MacOS doesn't support extension loading, so [this part](https://github.com/oven-sh/bun/blob/8677ae9fb154dea49939dd396fdd1363959f96de/src/bun.js/bindings/sqlite/JSSQLStatement.cpp#L701C14-L701C36) segfaults. I could solve it by with `Database.setCustomSQLite` like so:

```ts
Database.setCustomSQLite('/usr/local/opt/sqlite/lib/libsqlite3.dylib');
```

However, I still think it shouldn't segfault when extension loading isn't available, and instead throw a JS error. 

Maybe [`sqlite3_compileoption_used()`](https://www.sqlite.org/capi3ref.html#sqlite3_compileoption_get) can be used with [`SQLITE_OMIT_LOAD_EXTENSION`](https://www.sqlite.org/compile.html#omit_load_extension) to check if extension loading is possible, before attempting?

PR #5773, opened 2023-09-19
fix: add check to sqlite extension loading logic

liz3:
### What does this PR do?
The sqlite3 logic to dynamically load extensions can be disabled to save space, which macos does by default.
sqlite3 provides a api to check for these compile time settings at runtime, we can use that to throw a js error rather then crashing. It is worth noting though that the api to check for these settings at runtime itself can be disabled through SQLITE_OMIT_COMPILEOPTION_DIAGS but this seams to be a edge case.

Fixes https://github.com/oven-sh/bun/issues/5756

- [x] Code changes

Its not completely straight forward to add automated testing for this as it requires a custom build of sqlite3, assuming macOS version of sqlite3 will always disable extension loading is a bad approach.

Furthermore i work on arch even though i have a mac(i might test this later this week), i used this to test for the presence of flags outside of `SQLITE_OMIT_LOAD_EXTENSION` which worked as expected. So testing was done.


Jarred-Sumner (approved):
LGTM. Thank you

PR #5780, opened 2023-09-20
Show when a newer version is available in the install summary

Jarred-Sumner:
### What does this PR do?

Adds `(v${version} available)` text to the install summary:

image

This does mean reading and opening the manifest cache for each newly-installed package. To minimize perf impact, this feature is disabled in CI environments. It will only show when newer versions are available when that package was newly installed (not every time). 

Manual

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -1841,3 +1841,26 @@ pub fn fdi32(fd_: anytype) i32 {
 }
 
 pub const OSPathSlice = if (Environment.isWindows) [:0]const u16 else [:0]const u8;
+pub const LazyBoolValue = enum {
+    unknown,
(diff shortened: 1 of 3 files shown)


staabm:
would it make sense to show it only in interactive CLI sessions?

jumoog:
> would it make sense to show it only in interactive CLI sessions?

`To minimize perf impact, this feature is disabled in CI environments.`

staabm:
there is a difference between detecting a CI environment and detecting a interactive CLI session.

when I use bun as part of a script which I run locally in a cron deamon it also doesn't make sense to show this version information.
it only makes sense when bun was triggered by a real human

Issue #5635, opened 2023-09-17
Incorrect Handling of Relative URLs in Fetch HTTP Redirects


eskalera:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

I have encountered an issue with the handling of HTTP redirects involving relative URLs in the bun.sh runtime. According to the HTTP/1.1 specifications, when a server responds with a redirect containing a relative URL in the `Location` header, the HTTP client should interpret that relative URL with respect to the original request URL.

To reproduce this issue, I set up a simple Express.js server that responds to a GET request with a 302 redirect to a relative URL. Here is the server code:

```javascript
const express = require('express');
const app = express();
const port = 3000;

app.get('/redirect', (req, res) => {
  res.redirect(302, 'relative/path');
});

app.get('/relative/path', (req, res) => {
  res.send('You have been redirected to a relative path!');
});

app.listen(port, () => {
  console.log(`Server running at http://localhost:${port}/`);
});
```
And this is the client code which throws different results if executed with nodejs or bun.sh

```javascript
const resp = await fetch("http://localhost:3000/redirect", {
    method: "GET",
    verbose: true
});
const respText = await resp.text()
console.log(respText)
```

I found that when I make a request to `http://localhost:3000/redirect` using bun.sh, the client does not follow the redirect correctly. Instead of being redirected to `http://localhost:3000/relative/path` (as would be expected according to the HTTP/1.1 specifications), the client is attempting to access `http://localhost:3000relative/path`.

However, when I modify the server to respond with a redirect to `/relative/path` (noting the leading slash), bun.sh follows the redirect correctly.

### What is the expected behavior?

The formed redirect url should be `{host}/{relative/path}`. Therefore the output in my example should be
```
You have been redirected to a relative path!
```

### What do you see instead?

The formed redirect url is `{host}{relative/path}`. Therefore the current output is
```
ConnectionRefused: Unable to connect. Is the computer able to access the url?
 path: "http://localhost:3000/redirect"
```

### Additional information

This shouldn't be relevant to the case but I've tested it with a Bun server as well to test server redirects.

```javascript
const server = Bun.serve({
    port: 3000,
    fetch(req) {
        console.log(req.url);
        const url = URL(req.url);
        if (url.pathname === "/redirect") return Response.redirect("relative/path", 302);
[...]

PR #5781, opened 2023-09-20
fix(fetch): fix redirect in relative path location.

Hanaasagi:
### What does this PR do?

Close: #5635 

relative path location does not work in bun.

```JavaScript
import * as http from "node:http";
import * as assert from "node:assert";

const testCases = [
  ["/a/b", "/c", "/c"],
  ["/a/b", "c", "/a/c"],
  ["/a/b", "/c/d", "/c/d"],
  ["/a/b", "c/d", "/a/c/d"],
  ["/a/b", "../c", "/c"],
  ["/a/b", "../c/d", "/c/d"],
  ["/a/b", "../../../c", "/c"],
  // slash
  ["/a/b/", "/c", "/c"],
  ["/a/b/", "c", "/a/b/c"],
  ["/a/b/", "/c/d", "/c/d"],
  ["/a/b/", "c/d", "/a/b/c/d"],
  ["/a/b/", "../c", "/a/c"],
  ["/a/b/", "../c/d", "/a/c/d"],
  ["/a/b/", "../../../c", "/c"],
];

for (const [pathname, location, expected] of testCases) {
  const server = http.createServer((req, res) => {
    if (req.url === pathname) {
      res.writeHead(302, { Location: location });
      return res.end();
    } else if (req.url === expected) {
      res.writeHead(200, { "Content-Type": "text/plain" });
      return res.end("OK");
    }
    res.writeHead(404);
    res.end("Not Found");
  });

  server.listen(0, () => {
    const port = server.address().port;
    fetch(`http://127.0.0.1:${port}${pathname}`)
      .then((response) => response.text())
      .then((data) => {
        console.log("request", pathname, "=>", expected, data);
        assert.strictEqual(data, "OK");
        server.close();
      })
      .catch((error) => {
        assert.strictEqual(error, null);
        server.close();
      });
  });
}

```

- [x] Code changes

I wrote automated tests


src/http_client_async.zig
```diff
@@ -3480,21 +3480,16 @@ pub fn handleResponseMetadata(
...
-                            ) catch return error.RedirectURLTooLong);
-                        } else {
-                            this.url = URL.parse(std.fmt.bufPrint(
-                                &url_buf.data,
-                                "{s}://{s}:{d}{s}",
-                                .{ original_url.displayProtocol(), original_url.displayHostname(), port, location },
-                            ) catch return error.RedirectURLTooLong);
-                        }
+
+                        const new_url_ = bun.JSC.URL.join(
```
Hanaasagi, line 3485: Not sure using `JSC.URL.join`, maybe there's another function can be used instead.

Jarred-Sumner:
Are you sure this is broken in the http client and not something to do with node:http? There were tests for this I’m pretty sure

Hanaasagi:
Yes, I found this issue when making requests to a Server (which is written in another language I guessed). The server returns a **relative path** in `Location` header, which is not common. However, it seems that this is allowed. https://stackoverflow.com/questions/8250259/is-a-302-redirect-to-relative-url-valid-or-invalid


This issue can also be reproduced using the following code:


*server.js*

```JavaScript
import * as http from "node:http";

const server = http.createServer((req, res) => {
  if (req.url === "/aaa/bbb") {
    res.writeHead(302, { Location: "ccc/ddd" });
    return res.end();
  } else if (req.url === "/aaa/ccc/ddd") {
    res.writeHead(200, { "Content-Type": "text/plain" });
    return res.end("OK");
  }
  res.writeHead(404);
  res.end("Not Found");
});

server.listen(3000);

```

*client.js*

```JavaScript
async function main() {
  const resp = await fetch("http://127.0.0.1:3000/aaa/bbb", { verbose: true });
  console.log(await resp.text());
}

main();

```

Run `server.js` using node, and then execute `client.js` separately using both node and bun.

src/http_client_async.zig
Jarred-Sumner: URL.join is probably the right move here. But it will need to handle the case where the URL is invalid

src/http_client_async.zig
```diff
@@ -3480,21 +3480,16 @@ pub fn handleResponseMetadata(
...
+                            bun.String.fromUTF8(original_url.href),
+                            bun.String.fromUTF8(location),
+                        );
+                        defer new_url_.deref();
+
+                        const new_url = new_url_.byteSlice();
```
Jarred-Sumner: URL.join returns a latin1 or utf-16 encoded string. It must be converted to UTF8 using the usual methods

test/js/web/fetch/fetch.test.ts
```diff
@@ -1596,3 +1596,44 @@ it("same-origin status code 302 should not strip headers", async () => {
...
+  // slash
+  ["/a/b/", "/c", "/c"],
+  ["/a/b/", "c", "/a/b/c"],
+  ["/a/b/", "/c/d", "/c/d"],
+  ["/a/b/", "c/d", "/a/b/c/d"],
+  ["/a/b/", "../c", "/a/c"],
+  ["/a/b/", "../c/d", "/a/c/d"],
+  ["/a/b/", "../../../c", "/c"],
+])("should handle relative location in the redirect ('%s', '%s'), issue#5635", async (pathname, location, expected) => {
+  const server = Bun.serve({
```
Jarred-Sumner, line 1661: This test will be kind of slow/unreliable due to occupying a unique port per call 

Instead, can we create a single Bun.serve() instance, call `.reload` and then one .stop(true) inside an afterAll?

src/http_client_async.zig
Hanaasagi: Fixed.

test/js/web/fetch/fetch.test.ts
Hanaasagi: Fixed.

src/http_client_async.zig
```diff
@@ -3479,22 +3479,17 @@ pub fn handleResponseMetadata(
...
-                            ) catch return error.RedirectURLTooLong);
-                        }
+
+                        const new_url_ = bun.JSC.URL.join(
+                            bun.String.fromUTF8(original_url.href),
+                            bun.String.fromUTF8(location),
+                        );
+                        defer new_url_.deref();
+
+                        const new_url = new_url_.toUTF8(fba.allocator());
```
Hanaasagi, line 3494: `toUTF8` does not return error. If the buffer size is not enough, is it safe?

Jarred-Sumner: Not really, we assume memory allocations won't fail in that method. We should just use global allocator I think

`defer new_url.deinit()`

Hanaasagi: I added a size check to it, hoping it'll be useful.

```zig
                        if (new_url_.utf8ByteLength() > url_buf.data.len) {
                            return error.RedirectURLTooLong;
                        }
```

Jarred-Sumner: I'll merge this but `.toUTF8` might not always clone the string into the buffer, so we need to make sure it does that

PR #5788, opened 2023-09-20
Treat `undefined` value as empty in expect.toThrow

Brooooooklyn:
Fix: https://github.com/napi-rs/napi-rs/blob/main/examples/napi/__tests__/test.framework.js#L16-L19


Jarred-Sumner (approved):
LGTM. Thank you

PR #5799, opened 2023-09-20
fix(doc): correct `server.reload`

Hanaasagi:
### What does this PR do?

`server.update` to `server.reload`

```
TypeError: server.update is not a function. (In 'server.update({
  async fetch(request) {
    return new Response("Hello World v2");
  }
})', 'server.update' is undefined)

```

- [ ] Code changes


Jarred-Sumner:
Thanks

Issue #4742, opened 2023-09-09
parent-module not running on bun


noraj:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 6.4.12-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

- bun x create-docusaurus@latest my-website classic
- cd my-website
- bun install
- bun run docusaurus

### What is the expected behavior?

docusaurus running like on nodejs

### What do you see instead?

```
➜ bun run docusaurus
$ docusaurus
 9 |    }
10 | 
11 |    let seenVal = false;
12 | 
13 |    // Skip the first stack as it's this function
14 |    stacks.shift();
    ^
TypeError: stacks.shift is not a function. (In 'stacks.shift()', 'stacks.shift' is undefined)
      at /tmp/my-website/node_modules/parent-module/index.js:14:1
      at /tmp/my-website/node_modules/import-fresh/index.js:11:20
      at /tmp/my-website/node_modules/@docusaurus/core/lib/server/config.js:36:10
      at processTicksAndRejections (:1:2602)
error: script "docusaurus" exited with code 1 (SIGHUP)
```

### Additional information

> If you run into any bugs with a particular package, please [open an issue](https://bun.sh/issues). Opening issues for compatibility bugs helps us prioritize what to work on next.

Asking for bun support on upstream https://github.com/facebook/docusaurus/issues/9294

Josh-Cena:
Note that the stack ultimately points to `parent-module`. The code looks like:

```js
const callsites = require('callsites');

module.exports = filePath => {
	const stacks = callsites();
// ...
```

`callsites` looks like this:

```js
const callsites = () => {
	const _prepareStackTrace = Error.prepareStackTrace;
	Error.prepareStackTrace = (_, stack) => stack;
	const stack = new Error().stack.slice(1);
	Error.prepareStackTrace = _prepareStackTrace;
	return stack;
};
```

Therefore, it's probably because JSC doesn't have `Error.prepareStackTrace`. Bun would probably need to polyfill these V8 APIs.

The issue title is misleading because it has nothing to do with Docusaurus.

noraj:
This issue may be overlapping with #3426 and #4324

Jarred-Sumner:
Fixed in https://github.com/oven-sh/bun/pull/5802

This will be released in Bun v1.0.3. 

[...]

Issue #4606, opened 2023-09-08
No compatibility between node and bun for `new Error().stack` when I patched `prepareStackTrace`.


baseballyama:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

Run below code with `bun` command and `node` command.

```js
Error.prepareStackTrace = (_, stack) => stack;
const stack = new Error().stack.slice(1);
console.log(stack);
```

### What is the expected behavior?

Print `[ CallSite {} ]`.

### What do you see instead?

Print this.

```txt
rror: 
    at module code (/path/to/a.js:2:17)
```

### Additional information

I tried to run ESLint on bun but It failed due to this different behavior.

jansedlon:
Just adding that this might be the reason why 'pino' doesn't work in bun. It was throwing errors around working with stack traces as well

imZack:
Was trying to replace node with bun v1.0 but this is the the roadblocker.

Voldemat:
Related: https://github.com/oven-sh/bun/issues/4591

mentos1386:
Related: https://github.com/oven-sh/bun/issues/4742

Jarred-Sumner:
Fixed in https://github.com/oven-sh/bun/pull/5802

This will be released in Bun v1.0.3. 

If you don't want to wait until then, an hour after this comment was written you should be able to upgrade to the canary build of Bun:

```bash
bun upgrade --canary
```

Issue #4591, opened 2023-09-08
Error: stacks.shift is not a function. (In 'stacks.shift()', 'stacks.shift' is undefined)


singh-jay:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 6.2.0-32-generic x86_64 x86_64

### What steps can reproduce the bug?

Create a react app using CRA template(webpack 5)

- bunx create-react-app@latest react-test
- cd react-test
- bun --bun run start

### What is the expected behavior?

it should run the webpack dev server as node does.

### What do you see instead?

Failed to compile.

[eslint] Cannot read config file: /home/jaysingh/projects/test/react-test/node_modules/eslint-config-react-app/base.js
Error: stacks.shift is not a function. (In 'stacks.shift()', 'stacks.shift' is undefined)
Referenced from: BaseConfig
ERROR in [eslint] Cannot read config file: /home/jaysingh/projects/test/react-test/node_modules/eslint-config-react-app/base.js
Error: stacks.shift is not a function. (In 'stacks.shift()', 'stacks.shift' is undefined)
Referenced from: BaseConfig

ERROR in   Error: Child compilation failed:
  [eslint] Cannot read config file: /home/jaysingh/projects/test/react-test/node_modules/eslint-config-react-app/base.js
  Error: stacks.shift is not a function. (In 'stacks.shift()', 'stacks.shift' is undefined)
  Referenced from: BaseConfig

1 ERROR in child compilations (Use 'stats.children: true' resp. '--stats-children' for more details)
webpack compiled with 3 errors
Watchpack Error (watcher): Error: AccessDenied
71 | 
72 |             return;
73 |           }
74 | 
75 |           client.isAlive = false;
76 |           client.ping(() => {});
              ^
TypeError: client.ping is not a function. (In 'client.ping(() => {
          })', 'client.ping' is undefined)
      at /home/jaysingh/projects/test/react-test/node_modules/webpack-dev-server/lib/servers/WebsocketServer.js:76:10
      at forEach (:1:20)
      at /home/jaysingh/projects/test/react-test/node_modules/webpack-dev-server/lib/servers/WebsocketServer.js:64:6
65 |         /**
66 |          * @param {ClientConnection} client
67 |          */
68 |         (client) => {
69 |           if (client.isAlive === false) {
70 |             client.terminate();
                ^
[...]

Issue #4324, opened 2023-08-25
Support parent-module (npm package)


birkskyum:
parent-module - [npm](https://www.npmjs.com/package/parent-module) - [github](https://github.com/sindresorhus/parent-module#readme)

### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

Example from https://www.npmjs.com/package/parent-module :

```ts
// bar.js
import parentModule from 'parent-module';

export default function bar() {
	console.log(parentModule());
	//=> '/Users/<myyser>/<path>/foo.js'
};

// foo.js
import bar from './bar.js';

bar();
```

`bun foo.js`

### What is the expected behavior?

That it doesn't break and prints the foo.js path:

`/Users/<myyser>/<path>/foo.js`

### What do you see instead?

```sh
➜ bun foo.js 
2 | 
3 | export default function parentModule(filePath) {
4 |     const stacks = callsites();
5 | 
6 |     if (!filePath) {
7 |             return stacks[2].getFileName();
             ^
TypeError: undefined is not an object (evaluating 'stacks[2].getFileName')
      at parentModule (/Users/admin/repos/bun-parent-module-test/node_modules/parent-module/index.js:7:9)
      at bar (/Users/admin/repos/bun-parent-module-test/bar.ts:4:13)
      at /Users/admin/repos/bun-parent-module-test/foo.ts:3:0
```

### Additional information

Blocks:

- #3426 
- https://github.com/oven-sh/bun/issues/3154

Jarred-Sumner:
Fixed in https://github.com/oven-sh/bun/pull/5802

This will be released in Bun v1.0.3. 

If you don't want to wait until then, an hour after this comment was written you should be able to upgrade to the canary build of Bun:

```bash
bun upgrade --canary
```

PR #5802, opened 2023-09-20
Call `Error.prepareStackTrace` on `new Error().stack`

Jarred-Sumner:
### What does this PR do?

Previously, Bun was only calling `Error.prepareStackTrace` when `Error.captureStackTrace` was called. This broke a lot of packages. 

This makes Bun call `Error.prepareStackTrace` whenever it's defined, and also makes it very slightly faster to check for the existence of `Error.prepareStackTrace.

There are tests.


Jarred-Sumner:
Fixes https://github.com/oven-sh/bun/issues/4606
Fixes https://github.com/oven-sh/bun/issues/4591
Fixes https://github.com/oven-sh/bun/issues/4742
Fixes https://github.com/oven-sh/bun/issues/4324

Jarred-Sumner:
Fixes https://github.com/oven-sh/bun/issues/4280

Issue #5805, opened 2023-09-20
new URL("blob:https://example.com").origin should be https://example.com


fabiancook:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

Create the file `origin.js`:

```javascript
console.log(JSON.stringify(new URL("blob:https://example.com").origin));
```

Run

```shell
node origin.js
bun origin.js
```

### What is the expected behavior?

For `"https://example.com"` to be logged

```shell
$ node origin.js
"https://example.com"
$ bun origin.js
"https://example.com"
```

### What do you see instead?

```shell
$ node origin.js
"https://example.com"
$ bun origin.js
"null"
```

### Additional information

Chrome:

```
"https://example.com"
```

Safari:

```
"https://example.com"
```

Firefox:

```
"https://example.com/"
```

This behaviour is noted here: 

- https://developer.mozilla.org/en-US/docs/Web/API/URL/origin
- https://url.spec.whatwg.org/#origin
- https://url.spec.whatwg.org/#example-43b5cea5 

> The [origin](https://url.spec.whatwg.org/#concept-url-origin) of blob:https://whatwg.org/d0360e2f-caee-469f-9a2f-87d5b0456f6f is the [tuple origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-tuple) ("https", "whatwg.org", null, null).

Electroid:
Thanks for flagging this down, we'll fix this.

liz3:
i wanna look at this, sounds interesting!

PR #5825, opened 2023-09-20
fix: implement correct behaviour for urls with blob: scheme

liz3:
### What does this PR do?
Urls using the blob scheme can have special behaviour if their origin itself is a url when parsing.
This fixes that by parsing the path as a own domain and if valid and the schemes(protocols) are valid returns its origin.
Ive used node.js here a lot to make sure its behaviour is copied 1:1 and enabled the automated tests for it.

Fixes https://github.com/oven-sh/bun/issues/5805

- [x] Code changes


src/bun.js/bindings/URLDecomposition.cpp
```diff
@@ -35,7 +35,14 @@ String URLDecomposition::origin() const
 
     if (fullURL.protocolIsInHTTPFamily() or fullURL.protocolIsInFTPFamily() or fullURL.protocolIs("ws"_s) or fullURL.protocolIs("wss"_s))
         return fullURL.protocolHostAndPort();
-
+    if(fullURL.protocolIsBlob()) {
+        const String& path = fullURL.path().toString();
+        const URL subUrl { URL {}, path };
+        if(subUrl.isValid()) {
```
ghiscoding: hello, I'm not a maintainer or anything, but just noticed both your `if` (this one and the top one) are missing a space and that would probably fail Prettier linting. Also maybe wonder why you have 2 nested `if` when that could probably be just one combined, even though it might be a long condition

liz3: i ran bun fmt, but will fix ty, the seperation for the branches is because they concern different things.
In this case it does make sense to keep them separated

fabiancook:
If the tests return the origin for `"file://"` here for a blob with file... is this _different_ from normal behaviour of bun where `"null"` is normally returned for a file URL?

https://discord.com/channels/876711213126520882/876711213126520885/1154023498977640561 

This is how I originally found this bug

```javascript
new URL("file:///var/1").origin
```
Chrome:

```
"file://"
```

Safari:

```
"file://"
```

Firefox:

```
"null"
```

Node:

```
"null"
```

Bun:

```
"null"
```

Looking at the spec, it seems both are right if the origin is opaque/internal in any way. 

I can create a separate ticket for this, but I felt like it wasn't a bug and is allowed by the spec, but it should be consistent either way. 

https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin

```
$ node
Welcome to Node.js v20.6.1.
Type ".help" for more information.
> new URL("file:///var/1").origin
'null'
> new URL("blob:file:///var/1").origin
'null'
```

liz3:
> If the tests return the origin for `"file://"` here for a blob with file... is this _different_ from normal behaviour of bun where `"null"` is normally returned for a file URL?
> 
> https://discord.com/channels/876711213126520882/876711213126520885/1154023498977640561
> 
> This is how I originally found this bug
> 
> ```js
> new URL("file:///var/1").origin
> ```
> 
> Chrome:
> 
> ```
> "file://"
> ```
> 
> Safari:
> 
> ```
> "file://"
> ```
> 
> Firefox:
> 
> ```
> "null"
> ```
> 
> Node:
> 
> ```
> "null"
> ```
> 
> Bun:
> 
> ```
> "null"
> ```
> 
> Looking at the spec, it seems both are right if the origin is opaque/internal in any way.
> 
> I can create a separate ticket for this, but I felt like it wasn't a bug and is allowed by the spec, but it should be consistent either way.
> 
> https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin

`file://` is not the same as `blob:file://...` since the root scheme is blob for one and file for the other.
So there opening another ticket might be worth since thats not concerning `blob` scheme based urls.

But you mentioning `new URL("blob:file:///var/1").origin` peeked my interest as with the changes from this pr this does not return `"null"` it returns `"file://"`. why? because node.js on my laptop did that behaviour.

image

This is because on my laptop i have node `v18.17.1` while on my tower its `v18.15.0` so node changed its handling inbetween those versions.

liz3:
@fabiancook if this is in scope for the maintainers, i can handle the file scheme too, implementing it does not seam hard.
Otherwise file scheme based urls might need their own ticket/pr, which i would prefer as it keeps concerns seperated.
Again the reason i do handle `blob:file://....`  here is as its a subset of `blob` scheme based urls.

But i think mantaining consistency with how node implements it it more important then imo then to diviate from its impl even if it means being inconistent in this case.

cyfung1031:
The current origin for blob url in Bun's `URL` is not the same as Chrome & Node.js

Not only `blob:file://` but also `blob:https://` ... etc


```js
console.log(new URL('blob:https://github.com/d8500472-f516-47fc-91f3-10077983d1fc'))
```

## chrome

Screen Shot 2023-09-27 at 21 01 17




## node.js
```js

URL {
  href:
   'blob:https://github.com/d8500472-f516-47fc-91f3-10077983d1fc',
  origin: 'https://github.com',
  protocol: 'blob:',
  username: '',
  password: '',
  host: '',
  hostname: '',
  port: '',
  pathname: 'https://github.com/d8500472-f516-47fc-91f3-10077983d1fc',
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }
```


### bun
Screen Shot 2023-09-27 at 21 01 36

liz3:
@cyfung1031 your comment is what this pull requests resolves, so is there something i missed?
Here is the url you gave printed with this branch:
image
Origin: `https://github.com`

cyfung1031:
> @cyfung1031 your comment is what this pull requests resolves, so is there something i missed? Here is the url you gave printed with this branch: image Origin: `https://github.com`

Oh, I was wrong. Sorry.

liz3:
I believe non of the failures are related to the PR.

Jarred-Sumner:
This looks good. Thank you. Sorry for taking too long to review

liz3:
> This looks good. Thank you. Sorry for taking too long to review

No worries thanks for getting to it.

PR #5827, opened 2023-09-20
feat(runtime): align import.meta.resolve with node.js's implementation

### LINKED ISSUE #2472: Implement synchronous `import.meta.resolve(…)`

It is common to write apps and library code in JS that references relative file paths, such as:

- Dynamic JS entry files (in particular, for web workers/service workers/shared workers)
- WASM
- CSS, JSON, arbitrary resources, …

`import.meta.resolve(…)` provides a standard way to refer to a relative resource, and also supports bare names and absolute paths. It is already supported in all major browsers and [`deno`](https://deno.land/manual@v1.31.1/runtime/import_meta_api#importmetaresolve-example). It would be valuable for `import.meta.resolve(…)` to be available in *all* target environments, including `bun`.

Implement https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta/resolve

As a library author, I can use code that uses `new URL(…, import.meta.url).href` instead. However:

- `import.meta.resolve(…)` is more ergonomic and conveys clear intentions compared to being a "magic pattern", and
- it is not supported by all bundlers, and it's unclear if it ever will be (possibly due to the previous reason).

paperclover:
### What does this PR do?

Closes #2472 

- [ ] Code changes


packages/bun-types/globals.d.ts
```diff
@@ -303,23 +303,23 @@ interface ImportMeta {
...
   // tslint:disable-next-line:unified-signatures
-  resolve(moduleId: string, parent: string): Promise<string>;
+  resolve(moduleId: string, parent: string): string;
 
   /**
    * Resolve a module ID the same as if you imported it
    *
    * The `parent` argument is optional, and defaults to the current module's path.
+   *
+   * Will throw if the module does not exist.
```
jimmywarting: Erm... it should not throw an error if you do:

```js
import.meta.resolve('./this-file-dose-not-exist.md') // or
import.meta.resolve('./this-dir-dose-not-exist')
```
That is not how Deno, browser or NodeJS works right now. 
if the module can't be resolved by looking at some import-map or some package.json
then it should just return the relative string.

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -269,44 +269,92 @@ JSC_DEFINE_HOST_FUNCTION(functionImportMeta__resolve,
...
+    auto scope = DECLARE_THROW_SCOPE(globalObject->vm());
 
-        if (moduleName.isUndefinedOrNull()) {
+    auto thisValue = callFrame->thisValue();
+    auto specifierValue = callFrame->argument(0);
+    // 1. Set specifier to ? ToString(specifier).
+    auto specifier = specifierValue.toWTFString(globalObject);
+    RETURN_IF_EXCEPTION(scope, JSC::JSValue::encode(JSC::JSValue {}));
+
+    // Node.js allows a second argument for parent
```
jimmywarting, line 354: only behind a flag. it's likely that NodeJS will not support a 2nd argument for `import.meta.resolve` as they want to align with Deno and Browser that dose not support a 2nd argument.

packages/bun-types/globals.d.ts
paperclover: this documentation is for `resolveSync` not `resolve`

src/bun.js/bindings/ImportMetaObject.cpp
pi0: I think as of v20.11.1, `import.meta.resolve` supports second argument  (also for what is worth, if it is not supported, tooling support would be hell considering a module cannot resolve on behalf of another)

paperclover: it is still behind a flag: `--experimental-import-meta-resolve`

since we do not have node's feature flags, i think this is ok to just have, but it can just not be present in the typescript definitions

paperclover:
posix test failrues

test/bundler/bundler_compile.test.ts
test/js/bun/plugin/plugins.test.ts
test/transpiler/macro-test.test.ts

window test failure

test\js\bun\resolve\import-meta-resolve.test.mjs

paperclover:
test\bundler\bundler_compile.test.ts panics on windows

PR #5838, opened 2023-09-20
Improve types for `test.each`, `describe.each`

colinhacks:
Per https://discord.com/channels/876711213126520882/1153238257853800468/1153238257853800468

### DIFF
--- a/packages/bun-types/bun-test.d.ts
+++ b/packages/bun-types/bun-test.d.ts
@@ -14,6 +14,8 @@
  * ```
  */
 
+type id<T> = T;
+type flatten<T> = id<{ [k in keyof T]: T[k] }>;
 declare module "bun:test" {
   type AnyFunction = (...args: any) => any;
   /**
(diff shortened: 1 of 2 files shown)


paperclover (approved):

packages/bun-types/bun-test.d.ts
```diff
@@ -174,18 +174,18 @@ declare module "bun:test" {
...
+    each<T extends Readonly<Readonly<[any, ...any[]]>[]>>(
+      table: T,
     ): (
       label: string,
-      fn: (...args: T) => void | Promise<unknown>,
+      fn: (...args: Readonly<T>[number]) => void | Promise<unknown>,
       options?: number | TestOptions,
     ) => void;
-    each<T>(
+    each<T extends Array<any>>(
```
mProjectsCode, line 184: This breaks support for `each` with an array of objects. Now only an array of arrays is supported.

packages/bun-types/bun-test.d.ts
```diff
@@ -419,18 +419,18 @@ declare module "bun:test" {
...
       table: ReadonlyArray<T>,
     ): (
       label: string,
-      fn: (arg: T, done: (err?: unknown) => void) => void | Promise<unknown>,
```
mProjectsCode, line 433: What happened to the `done` callback?

packages/bun-types/bun-test.d.ts
```diff
@@ -419,18 +419,18 @@ declare module "bun:test" {
...
+    each<T extends Readonly<Readonly<[any, ...any[]]>[]>>(
+      table: T,
     ): (
       label: string,
-      fn: (...args: T) => void | Promise<unknown>,
+      fn: (...args: Readonly<T>[number]) => void | Promise<unknown>,
       options?: number | TestOptions,
     ) => void;
-    each<T>(
+    each<T extends Array<any>>(
```
mProjectsCode, line 429: Same thing as above

mProjectsCode (commented):
This is a good change, but it breaks support for `each` with an array of objects and the `done` callback was lost along the way.

mProjectsCode:
@colinhacks @paperdave I would suggest taking another look at this.

packages/bun-types/bun-test.d.ts
colinhacks: The previous typings that included `done` were just completely wrong. There was no guarantee that `done` would be the second argument to this. It's basically impossible to include `done` in the types in the general case here, if we want any sort of inference on the argument to `.each()`.

PR #5839, opened 2023-09-20
fix(console.log): change default depth from 8 to 2

JibranKalia:
### What does this PR do?

Changes the default depth to match Node. [Source in Node](https://github.com/nodejs/node/blob/480ab8c3a40451d5ea459dd35b4039679b26e195/doc/api/console.md?plain=1#L285)

- [X] Code changes


Jarred-Sumner:
Thank you

PR #5873, opened 2023-09-21
chore(git): ignore .envrc file

igorshapiro:
### What does this PR do?

Using `.envrc` files allows [automatically loading](https://direnv.net/) custom env vars during development.
This file might contain tokens, keys or any other developer-workflow-specific config that we don't want to push to the repo.

This PR adds .envrc to .gitignore


Jarred-Sumner:
Thanks

Issue #5867, opened 2023-09-21
Express believes `protocol=http` when HTTPS request


Foorack:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Linux 6.2.0-33-generic x86_64 x86_64

### What steps can reproduce the bug?

I made a tiny single-file repo which consistently reproduces the bug on multiple machines.

### Setup:

1. git clone https://github.com/Foorack/bun-proto-test
2. cd bun-proto-test
3. bun install

### Verify it works with Node

`node index.js`

Go to `https://localhost:4343` and see it says `https`.

> OK : https : Node

### Verify it is broken in Bun

`bun run index.js`

Go to `https://localhost:4343` and see it says `http`.

> OK : http : Bun

### What is the expected behavior?

It is supposed to print "https" when running with Bun, becasue the request is HTTPS. It works in Node.

### What do you see instead?

I see "http" (req.protocol = http)

### Additional information

This causes `express-session` to not send secure cookies, which completely breaks logins in all systems. Had to revert to Node. 

This is the last bug blocking migration to Bun. :slightly_smiling_face:

liz3:
on this rn

Foorack:
@liz3  I did more digging and I believe this is because bun's implementation of `net.Socket` does not implement `socket.encrypted`.

Node: `console.log(req.socket.encrypted)` = true
Bun: `console.log(req.socket.encrypted)` = undefined

liz3:
correct

Foorack:
@liz3  Thanks for looking into this. I went back into the archives, and this seems to be a magic undocumented bool Node sets on TLS sockets. Major libraries like ExpressJS then depend on this undocumented bool, and has been for the last 10 years.

Breifly removed and reintroduced: https://github.com/nodejs/node-v0.x-archive/pull/6737/files

Discussion: https://github.com/nodejs/node-v0.x-archive/issues/6735

Related to https://github.com/expressjs/express/issues/1864

PR #5878, opened 2023-09-21
fix: correctly pass the encrypted(bool) property on "Socket" for express.js

liz3:
### What does this PR do?

Express relies on this for setting the requests protocol.
Since this is on a dummy object, a property is simply set with the actual value.
Which seams okay as a workaround.

Fixes: https://github.com/oven-sh/bun/issues/5867

- [X] Code changes


src/js/node/http.ts
```diff
@@ -155,6 +155,10 @@ var FakeSocket = class Socket extends Duplex {
 
   _final(callback) {}
 
+  get encrypted() {
```
paperclover: instead of a getter, you should just set the `encrypted` property directly on it

paperclover (changes requested):

src/js/node/http.ts
liz3: 👍

Jarred-Sumner:
Thank you

PR #5882, opened 2023-09-22
Hide env elapsed time when loglevel=info

colinhacks:
Disables `[0.05ms] ".env"` messages when logLevel is info


Jarred-Sumner:
I think there's a test for this somewhere

Jarred-Sumner:
looks good tho

Issue #5460, opened 2023-09-15
undefined is not an object - if dynamic library is not found (Bun FFI)


LIMPIX31:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Linux 6.1.51-1-MANJARO x86_64 unknown

### What steps can reproduce the bug?

Take the example from https://bun.sh/docs/api/ffi#usage-bun-ffi and specify a non-existent path to the dynamic library

### What is the expected behavior?

The error message will indicate that the file path does not exist

### What do you see instead?
```
56 |       break;
57 |     }
58 |   }
59 |   return wrap.native = functionToCall, wrap.ptr = functionToCall.ptr, wrap;
60 | 
61 | }, dlopen = function(path, options) {
                                           ^
TypeError: undefined is not an object
      at bun:ffi:61:40
```

### Additional information

`Bun.FFI.dlopen` outputs the message correctly, while `dlopen` imported from `bun:ffi` is confusing

hasanayan:
I have the same problem, , except I am not able to use Bun.FFI.dlopen

EzraEllette:
So I tracked down the problem with `dlopen`. is is in the `object.IsEmptyOrUndefinedOrNull()` call. When I take this out the error messages are clear. #5886 
[ffi.zig](https://github.com/oven-sh/bun/blob/main/src/bun.js/api/ffi.zig#L286-L288)

```zig
        if (object.isEmptyOrUndefinedOrNull() or !object.isObject()) {
            return JSC.toInvalidArguments("Expected an options object with symbol names", .{}, global);
        }
```

EzraEllette:
@paperdave How was this resolved?

PR #5886, opened 2023-09-22
update ffi.zig for dlopen

EzraEllette:
### What does this PR do?

The changes fix a bug where `dlopen` is not working in `bun:ffi`. Without these changes, attempts to use the function will result in an undefined error, and no real errors can propagate.

- [x] Code changes

- [x]  I included a test for the new code, or an existing test covers it


EzraEllette:
https://github.com/oven-sh/bun/issues/5460

paperclover:
these tests already pass in the current canary. the change does not change any logic

EzraEllette:
> these tests already pass in the current canary. the change does not change any logic

Then your canary is broken.

EzraEllette:
The issue it no longer present in 1.0.3

Issue #5898, opened 2023-09-22
bufferutil shouldn't be in cache


qweered:
### What version of Bun is running?

1.0.3

### What steps can reproduce the bug?

```javascript
const I = require('bufferutil')
const cache = require.cache
console.log('bufferutil' in cache, Object.keys(cache).includes('bufferutil'))

// in bun -> true false
// in node -> false false
```

PR #5903, opened 2023-09-22
fix(runtime): exclude unevaluated module in `require.cache`

Hanaasagi:
### What does this PR do?

Close: #5898

#5233 does not filter them in `has`, which causes that `"bufferutil" in require.cache` returns `true`.

- [x] Code changes

I wrote automated tests 


paperclover (approved):
oh wow i forgot about has. good catch.

Jarred-Sumner:
Thank you

Issue #4734, opened 2023-09-09
pm2 with bun


pazer001:
### What is the type of issue?

Documentation is missing

### What is the issue?

I can't find how it is possible to make pm2 work with bun

eliot-akira:
I haven't tried it yet, but I believe the `interpreter` option of PM2 can be used.

As [CLI option](https://pm2.keymetrics.io/docs/tutorials/using-transpilers-with-pm2#execution-interpreter):

```sh
pm2 start --interpreter ~/.bun/bin/bun index.ts
```

..Or in the [configuration file](https://pm2.keymetrics.io/docs/usage/application-declaration/#attributes-available):

```js
module.exports = {
  name: 'app',
  script: 'index.ts',
  interpreter: '~/.bun/bin/bun',
}
```

---

It would great if there was a process manager like PM2 but simpler and designed for Bun.

RicoCelsius:
> ```shell
> pm2 start --interpreter ~/.bun/bin/bun index.ts
> ```

Tried it, does work :)

eliot-akira:
> PM2 but simpler and designed for Bun

Maybe this one:

- PMB: Process Manager for Bun - https://github.com/w-xuefeng/pmb

pazer001:
> I haven't tried it yet, but I believe the `interpreter` option of PM2 can be used.
> 
> As [CLI option](https://pm2.keymetrics.io/docs/tutorials/using-transpilers-with-pm2#execution-interpreter):
> 
> ```shell
> pm2 start --interpreter ~/.bun/bin/bun index.ts
> ```
> 
> ..Or in the [configuration file](https://pm2.keymetrics.io/docs/usage/application-declaration/#attributes-available):
> 
> ```js
> module.exports = {
>   name: 'app',
>   script: 'index.ts',
>   interpreter: '~/.bun/bin/bun',
> }
> ```
> 
> It would great if there was a process manager like PM2 but simpler and designed for Bun.

it didn't work. i mean, I starts the services but some actions does not works. i think its because its not reading the environment variables. this is my config:
``const os = require('os');
const instances = Math.floor(os.cpus().length);

module.exports = {
  apps: [
    {
      name: 'system_settings_backtest1d',
      script: './dist/main.js',
      instances: instances,
      autorestart: true,
      out_file: '/dev/null',
      error_file: '/dev/null',
      interpreter: 'bun',
      env: {
        USE_RANDOM_PORT: true,
[...]

PR #5931, opened 2023-09-22
Docs : Added instructions to run bun apps in daemon (PM2) to address …

h2210316651:
…issue #4734

Added instructions to set bun as pm2 interpreter to extend same functionality as node.js apps.


h2210316651:
Updated against main

richardguerre:
Thank you this is so useful to know.

PR #5937, opened 2023-09-23
docs: Removing versions from arch dependencies

TheCynosure:
### What does this PR do?

Removes versions from arch dependencies as they are not needed. Below are the two packages in Arch x86_64 repository which are named without the versions:

Clang: https://archlinux.org/packages/?name=clang
LLVM: https://archlinux.org/packages/?name=llvm

- [ ] Code changes

Verified the install command on my arch machine.


Jarred-Sumner:
Thanks for this, but the specific versions of these tools are necessary. Bun expects LLVM 16, and when LLVM 17 comes out this would be incorrect.

PR #5939, opened 2023-09-23
Remove bun dev stuff from create_command

colinhacks:
Built & tested locally

```
bun-debug create xixixao/convex-nextjs-template
```

I left things commented in case there are elements of this we need for other reasons


Jarred-Sumner:
Thank you

PR #5950, opened 2023-09-23
Use c-ares function for checking if a string is an IP address

Jarred-Sumner:
### What does this PR do?

c-ares is more well-tested in production so we should opt for using it for these sorts of things

existing tests

### DIFF
--- a/src/string_immutable.zig
+++ b/src/string_immutable.zig
@@ -4676,30 +4676,30 @@ test "eqlCaseInsensitiveASCII" {
 }
 
 pub fn isIPAddress(input: []const u8) bool {
-    if (containsChar(input, ':'))
-        return true;
+    var max_ip_address_buffer: [512]u8 = undefined;
+    if (input.len > max_ip_address_buffer.len) return false;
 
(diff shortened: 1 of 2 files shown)


src/string_immutable.zig
```diff
@@ -4676,30 +4676,30 @@ test "eqlCaseInsensitiveASCII" {
...
-        return true;
-    } else |_| {
-        return false;
-    }
+    var sockaddr: std.os.sockaddr = undefined;
+    @memset(std.mem.asBytes(&sockaddr), 0);
+    @memcpy(max_ip_address_buffer[0..input.len], input);
+    max_ip_address_buffer[input.len] = 0;
+
+    var ip_addr_str: [:0]const u8 = max_ip_address_buffer[0.. :0];
```
Hanaasagi, line 4701: dev build panic here `uh-oh: sentinel mismatch: expected 0, found 1`

```zig
    var ip_addr_str: [:0]const u8 = max_ip_address_buffer[0..input.len :0];

```

Issue #5888, opened 2023-09-22
bun update always change lockb file


CGQAQ:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Linux C20230424168060 5.15.0-79-generic #86-Ubuntu SMP Mon Jul 10 16:07:21 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### What steps can reproduce the bug?

```console
mkdir empty
cd empty
bun add webpack
bun update
cp bun.lockb bun.lockb.old
bun update
vimdiff  <(xxd bun.lockb) <(xxd bun.lockb.old)
```

### What is the expected behavior?

no diff

### What do you see instead?

![image](https://github.com/oven-sh/bun/assets/15936231/45d61046-e640-49d4-ab60-68882afb2e53)

mattbdc:
bun update followed by bun update --save also doesn't touch package.json, I think it should right? npm update followed by npm update --save does.

CGQAQ:
This issue is not related to `package.json` at all, what I'm saying is lockfile

PR #5957, opened 2023-09-23
fix(lockfile): ensure all bytes of union are initialized before serialization.

Hanaasagi:
### What does this PR do?

Close: #5888

When serializing to `bun.lockb`, we directly use `std.mem.sliceAsBytes` to write the memory of the struct.

https://github.com/oven-sh/bun/blob/ec0e931e9f7934f4f1f7617eac2a880d13794d0c/src/install/lockfile.zig#L3757-L3762

In the `Bin` struct, the `value` field is `union` type, and its size depends on the largest member. If we initialize it with `.{.none = {}}`, there will be some uninitialized memory. This can lead to the `bun.lockb` file being different with each `bun update`(no packages updated).

https://github.com/oven-sh/bun/blob/ec0e931e9f7934f4f1f7617eac2a880d13794d0c/src/install/bin.zig#L19-L23

https://github.com/oven-sh/bun/blob/ec0e931e9f7934f4f1f7617eac2a880d13794d0c/src/install/bin.zig#L102-L135

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/install/bin.zig
+++ b/src/install/bin.zig
@@ -20,7 +20,7 @@ pub const Bin = extern struct {
     tag: Tag = Tag.none,
     _padding_tag: [3]u8 = .{0} ** 3,
 
-    value: Value = Value{ .none = {} },
+    value: Value = std.mem.zeroes(Value),
 
     pub fn verify(this: *const Bin, extern_strings: []const ExternalString) void {
         if (comptime !Environment.allow_assert)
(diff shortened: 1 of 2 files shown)


src/install/bin.zig
```diff
@@ -67,36 +67,39 @@ pub const Bin = extern struct {
     }
 
     pub fn clone(this: *const Bin, buf: []const u8, prev_external_strings: []const ExternalString, all_extern_strings: []ExternalString, extern_strings_slice: []ExternalString, comptime StringBuilder: type, builder: StringBuilder) Bin {
-        return switch (this.tag) {
-            .none => Bin{ .tag = .none, .value = .{ .none = {} } },
-            .file => Bin{
-                .tag = .file,
-                .value = .{ .file = builder.append(String, this.value.file.slice(buf)) },
+        var cloned = std.mem.zeroes(Bin);
```
Jarred-Sumner: I think we can be a little safer here about zeroing if we instead create it via `@bitCast`

```suggestion
        const cloned_bytes = .{0} ** @byteSize(Bin);
        var cloned: Bin = @bitCast(cloned_bytes);
```

Jarred-Sumner:
thank you for finding this

i spent a bunch of time trying to figure out what the cause was and did not succeed

src/install/bin.zig
Hanaasagi: need some help, I'm not familiar with `{0} **`. When I changed it to this

```zig
        const cloned_bytes = .{0} ** @sizeOf(Bin);
        var cloned: Bin = @bitCast(cloned_bytes);
```

compiled error:

```
src/install/bin.zig:74:36: error: cannot @bitCast from 'struct{comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0, comptime comptime_int = 0}'; struct does not have a guaranteed in-memory layout
        var cloned: Bin = @bitCast(cloned_bytes);
                                   ^~~~~~~~~~~~

```

And change to 

```zig
        const cloned_bytes: [@sizeOf(Bin)]u8 = .{0} ** @sizeOf(Bin);
        var cloned: Bin = @bitCast(cloned_bytes);
```

got


```
zig build-obj bun-debug Debug native-native-gnu.2.27: error: thread 2127830 panic: TODO implement readFromMemory for more types
Unable to dump stack trace: debug info stripped

```

Jarred-Sumner: ah yeah it's 
```suggestion
const cloned_bytes = [_]u8{0} ** @sizeOf(Bin);
var cloned: Bin = @bitCast(cloned_bytes);
```

Jarred-Sumner: if that continues to fail then i suggest 

```zig
var cloned: Bin = std.mem.zeroes(Bin);
@memset(std.mem.asBytes(&cloned), 0);
```

Hanaasagi: Fixed. `[_]u8{0} ** @sizeOf(Bin);` continues to fail with `TODO implement readFromMemory for more types`. I change it to `@memset`.

PR #5968, opened 2023-09-23
Update installation.md

sonyarianto:
Adding notes for users that currently use canary version and want to  switch back to stable version.

### What does this PR do?

- [ ] Code changes


sonyarianto:
Adding notes for users that using canary version and want to switch back to stable version.

PR #5998, opened 2023-09-24
Add `Module._extensions`

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/ljharb/define-data-property/actions/runs/6165229997/job/17073357298, probably

There is a test

### DIFF
--- a/src/bun.js/modules/NodeModuleModule.h
+++ b/src/bun.js/modules/NodeModuleModule.h
@@ -330,12 +330,17 @@ DEFINE_NATIVE_MODULE(NodeModule) {
     exportNames.append(name);
     exportValues.append(value);
   };
-  exportNames.reserveCapacity(15);
-  exportValues.ensureCapacity(15);
+  exportNames.reserveCapacity(16);
+  exportValues.ensureCapacity(16);
   exportNames.append(vm.propertyNames->defaultKeyword);
   exportValues.append(defaultObject);
 
   put(Identifier::fromString(vm, "Module"_s), defaultObject);
+
+  // Module._extensions === require.extensions
+  put(Identifier::fromString(vm, "_extensions"_s),
+      globalObject->requireFunctionUnbound()->get(
+          globalObject, Identifier::fromString(vm, "extensions"_s)));
 
   defaultObject->putDirectCustomAccessor(
       vm, JSC::Identifier::fromString(vm, "_resolveFilename"_s),
--- a/test/js/node/module/node-module-module.test.js
+++ b/test/js/node/module/node-module-module.test.js
@@ -87,3 +87,10 @@ test("Module.prototype._compile", () => {
   expect(fn).toBe("/file/path/goes/here.js");
   expect(dn).toBe("/file/path/goes");
 });
+
+test("Module._extensions", () => {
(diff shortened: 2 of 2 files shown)


ljharb (commented):
Seems good! Will adding an extension function here work, a la ts-node, babel-node, etc?

Jarred-Sumner:
> Seems good! Will adding an extension function here work, a la ts-node, babel-node, etc?

Nah, these are just stubs. We could add it in the future if it breaks libraries to not do it though. Usually they add them to support TypeScript/JSX which Bun does automatically. But that certainly doesn't cover every usecase

ljharb:
Gotcha - in this case, that will then just break things in a different way - code coverage via istanbul/nyc requires being able to wrap the extension-based CJS loaders that live there, so it's a pretty widespread use case.

Jarred-Sumner:
The underlying ccde coverage API using the V8 API there wouldn't work. We need to provide a code coverage API and then probably add a plugin for nyc. It also needs to be sourcemapped since Bun transpiles every file by default. We have sourcemapped code coverage for `bun test` with `bun test --coverage` but not much of a programmatic API yet.

ljharb:
Not the v8 API; istanbul predates it and the version I use (nyc 10) doesn't rely on it. nyc does support source maps, at least.

Is there no way to allow functions installed on Module._extensions to be called during CJS module load?

ljharb:
on v1.0.4 - v1.0.6, here's the errors now: https://github.com/ljharb/define-data-property/actions/runs/6165229997/job/17544945356
> \> nyc tape 'test/**/*.js'
> npm ERR! p.stdin.end is not a function. (In 'p.stdin.end()', 'p.stdin.end' is Infinity)
> npm ERR! A complete log of this run can be found in: /home/runner/.npm/_logs/2023-10-09T23_40_08_062Z-debug-0.log
> 155 |   var source = null
> 156 |   var ext = path.extname(filePath)
> 157 |   if (typeof Module._extensions[ext] === 'undefined') {
> 158 |     ext = '.js'
> 159 |   }
> 160 |   Module._extensions[ext]({
>       ^
> TypeError: Module._extensions[ext] is not a function. (In 'Module._extensions[ext]({
>       _compile: function(content, filename) {
>         source = content;
>       }
>     }, filePath)', 'Module._extensions[ext]' is true)
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/index.js:160:2
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/index.js:144:15
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/index.js:176:4
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/index.js:244:4
>       at forEach (:1:20)
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/index.js:243:2
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/index.js:174:2
>       at /home/runner/work/define-data-property/define-data-property/node_modules/nyc/bin/nyc.js:42:18
>       at globalThis (/home/runner/work/define-data-property/define-data-property/node_modules/nyc/bin/nyc.js:44:6)

Issue #6007, opened 2023-09-24
Create Project Error


jashok2:
### What version of Bun is running?

1.0.3

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

to run the command:
bun create @quick-start/electron

### What is the expected behavior?

Should create a new electron-vite project

### What do you see instead?

🔍 @quick-startcreate-/electron [1/1]
error: package "@quick-startcreate-/electron" not found registry.npmjs.org/@quick-startcreate-%2felectron 404

I don't know what's going on, but as you can see bun adds "create-" when parsing the created project name string

jashok2:
I also tried using the command "bun create '@quick-start/electron'", wrapping '@quick-start/electron' in single quotes, and got the same error

PR #6013, opened 2023-09-24
Fix create command with template prefixed with @ char #6007

axlEscalada:
### What does this PR do?

fix error when command `bun create <@some-template/template>` is executed, the command is appending a `create-` prefix wrongly.
For example:
`bun create @quick-start/electron`
tries to create with this template: `@quick-startcreate-/electron`

Fixes: https://github.com/oven-sh/bun/issues/6007

- [x] Code changes

Typescript test of `create` command that check the error case with the correct template.
I couldn't add a test in `bunx_command.zig` because it fails when tries to find "root".bun file ([zig issue here](https://github.com/ziglang/zig/issues/17109)). 

image

### DIFF
--- a/src/cli/bunx_command.zig
+++ b/src/cli/bunx_command.zig
@@ -23,7 +23,8 @@ pub const BunxCommand = struct {
 
         var new_str = try allocator.allocSentinel(u8, input.len + prefixLength, 0);
         if (input[0] == '@') {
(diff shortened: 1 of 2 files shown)


paperclover:
thank you

paperclover (approved):

PR #6019, opened 2023-09-24
(console.log) Print className for an object if present

JibranKalia:
### What does this PR do?

This adds the functionality to print Object's class name if it exists matching Node's functionality. 

As well as adding adding tests for printing Maps, Sets and ArrayBuffers.

- [X] Code changes

### DIFF
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -2862,6 +2862,12 @@ pub const ZigConsoleClient = struct {
                         .parent = value,
                     };
 
+                    var class_name = ZigString.Empty;
+                    value.getClassName(this.globalThis, &class_name);
+                    if (class_name.len > 0 and !strings.eqlComptime(class_name.slice(), "Object")) {
+                        writer.print("{} ", .{class_name});
+                    }
+
                     if (this.depth > this.max_depth) {
                         if (this.always_newline_scope or this.goodTimeForANewLine()) {
                             writer.writeAll("\n");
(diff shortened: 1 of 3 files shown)


src/bun.js/bindings/exports.zig
```diff
@@ -2862,6 +2862,12 @@ pub const ZigConsoleClient = struct {
                         .parent = value,
                     };
 
+                    var class_name = ZigString.Empty;
```
JibranKalia, line 2865: Uses ZigString instead of bun.string because `getClassName` expects it.

Jarred-Sumner:
Tests for `URLSearchParams` and `URL` seem to be failing

JibranKalia:
> Tests for `URLSearchParams` and `URL` seem to be failing

Its hard for me to tell exactly what relevant tests failed in the Github actions UI. I ran `bun test url` locally and those tests passed. Do you have any tips?

JibranKalia:
Just rebased on top of main. Hopefully, that's not an issue here. I usually do this at work but I am not sure if its frowned upon here.

JibranKalia:
I think I will have to close this PR. I am not really able to figure out the issue. I got to here:

https://github.com/oven-sh/bun/blob/745b6b94ee56cad24d475799690cc9a89957d15b/src/bun.js/bindings/exports.zig#L1823-L1831

However, I can't understand why `value.getNameProperty` returns `URL` when passed the first child - "href". However, it doesn't do the that for an object instantiated from a Class.

JibranKalia:
Closing this. Will try to figure out a different approach.

PR #6026, opened 2023-09-25
fix latest dev build panic

Hanaasagi:
#### repro

use the latest dev build to run following code.

```JavaScript
const http = require("node:http");

const server = http.createServer((req, res) => {});

server.listen(3000);

```

coredump stack

- [x] Code changes

existing tests


Jarred-Sumner:
Thank you

PR #6028, opened 2023-09-25
Lambda bootstrap smol

vladaman:
### What does this PR do?

Adds extra command line argument for Lambda bun executable. Allows your to flexibly tune bun when needed. In particular we use it with "`--smol`" argument.

Deployed and tested on AWS Lambda. When Env variable is empty it has no effect on the bootstrap.


vladaman:
I learned today that `smol` parameter can be added to `bunconig.yaml` - however this PR allows us to use any parameter regardless and probably is future-proof.

PR #6029, opened 2023-09-25
WIP: separate File and Blob

### LINKED ISSUE #4460: blob to file conversion in formData

0.8.1

Darwin 22.4.0 arm64 arm

```js
var fd = new FormData()

var f1 = new File(['abc'], 'folder/readme.md', { type: 'text/plain' })
var f2 = new Blob(['abc'])

fd.set('f1', f1)
fd.set('f2', f2, 'customName.txt')

console.assert(fd.get('f1') instanceof File, 'adding a file remain as a file')
console.assert(fd.get('f2') instanceof File, 'adding a blob should be converted into a File')

console.assert(fd.get('f2').name === 'customName.txt', 'the name should be customName.txt')
console.assert(fd.get('f1').name === 'folder/readme.md', 'the name should be folder/readme.md')

new Response(fd).formData().then(fd => {
  console.assert(fd.get('f1') instanceof File, 'adding a file remain as a file')
  console.assert(fd.get('f2') instanceof File, 'adding a file remain as a file')

  console.assert(fd.get('f2').name === 'customName.txt', 'the name should be customName.txt')
  console.assert(fd.get('f1').name === 'folder/readme.md', 'the name should be folder/readme.md')
})
```

not throwing any assertion errors

### What do you see instead?

- i got blobs back from `fd.get(x)`
- file names wasn't what i expected.

### Additional information

_No response_

### LINKED ISSUE #5980: `Bun.file` should extend `File` not `Blob`

AaronDewes:
### What does this PR do?

It separates `File` and `Blob`. Not sure if this is a good way to do it, and it is still WIP (Crashes on Startup if a script uses file, but REPL works, but that may be using a stable binary, not sure about how it works yet).

Fixes #4460 and will also later fix #5980.

I'm still learning Zig, so if this PR is useless and I took a completely wrong approach, please let me now.

- [x] Code changes

It's broken right now, but I will add tests once this works.


Jarred-Sumner:
Thank you for starting work on this

Note that `File` is supposed to extend `Blob` and methods defined on `Blob` like `text`, `arrayBuffer` etc are supposed to not be duplicated but instead come from the `Blob` prototype. 

This will involve a change to the code generator to support setting the prototype in the generated code to `JSBlob::info()`   in `JSFile::s_info = ...`

If you want to keep them as separate implementations, then it will need some way to tell what type the pointer is (`TaggedPointerUnion` is one option probably). It's probably simpler to revert back to a combined implementation (though keep the changes removing JSDOMFile etc)

AaronDewes:
> Thank you for starting work on this
> 
> Note that `File` is supposed to extend `Blob` and methods defined on `Blob` like `text`, `arrayBuffer` etc are supposed to not be duplicated but instead come from the `Blob` prototype.
> 
> This will involve a change to the code generator to support setting the prototype in the generated code to `JSBlob::info()` in `JSFile::s_info = ...`
> 
> If you want to keep them as separate implementations, then it will need some way to tell what type the pointer is (`TaggedPointerUnion` is one option probably). It's probably simpler to revert back to a combined implementation (though keep the changes removing JSDOMFile etc)

You're right, I overcomplicated things a lot here. I'll clean that up

jimmywarting:
Keep up the good work! you are doing a great job! 👏 

kind of wish this got solved right now, cuz it's a blocker for me atm:
```js
const file = new File(['foo'], 'dummy.txt', { type: 'text/plain' })
const fd = new FormData()
fd.set('foo', file)
fd.get('foo').name // undefined
```

AaronDewes:
I've made some progress locally, but currently stuck at a very weird issue. I'll push the new version soon, maybe someone else can figure that issue out.

src/env.zig
```diff
@@ -17,7 +17,7 @@ pub const isMac = build_target == .native and @import("builtin").target.os.tag =
 pub const isBrowser = !isWasi and isWasm;
 pub const isWindows = @import("builtin").target.os.tag == .windows;
 pub const isPosix = !isWindows and !isWasm;
-pub const isDebug = std.builtin.Mode.Debug == @import("builtin").mode;
+pub const isDebug = false;
```
AaronDewes, line 20: That was an accident, I'll revert this

src/standalone_bun.zig
```diff
@@ -57,7 +57,8 @@ pub const StandaloneModuleGraph = struct {
                     blob_.content_type_allocated = false;
                 }
 
-                store.data.bytes.stored_name = bun.PathString.init(this.name);
+                // Assigning this.name directly would cause a use-after-free if blob was deinited
+                blob_.fileName = bun.default_allocator.dupeZ(u8, this.name) catch @panic("out of memory");
```
AaronDewes, line 61: Not sure if this is correct, I need to check again why I did this

src/bun.js/bindings/webcore/JSDOMFormData.cpp
```diff
@@ -464,17 +465,23 @@ static inline JSC::EncodedJSValue jsDOMFormDataPrototypeFunction_set2Body(JSC::J
...
+        fileValue = File::fromJSBlob(argument1.value(), &filenameString);
+    } else {
+        throwTypeError(lexicalGlobalObject, throwScope, "Expected argument to be a Blob or File."_s);
+        RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
     }
 
-    if (!blobValue) {
-        throwTypeError(lexicalGlobalObject, throwScope, "Expected argument to be a Blob."_s);
+    if (!fileValue) {
+        throwTypeError(lexicalGlobalObject, throwScope, "This should never be reached!"_s);
```
AaronDewes, line 480: This is the issue I'm currently stuck at: !fileValue causes fileValue to disappear, and I have no idea why. On the line before this, I can access properties of fileValue without a problem, for example get the name. But this causes it to be dropped for some reason.

AaronDewes (commented):
Some notes, mostly for myself. Separation now only happens in CPP & codegen, the Zig implementation is shared. This will be cleaned up further soon.

src/env.zig
jimmywarting: Here is a little nudge
```diff
- pub const isDebug = false;
+ pub const isDebug = std.builtin.Mode.Debug == @import("builtin").mode;
```
Just accept the changes 👇

jimmywarting:
friendly bump

Issue #5974, opened 2023-09-23
`bunx` does not work with `github:user/repo` syntax


cyfung1031:
### What version of Bun is running?

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

### What platform is your computer?

Darwin 21.6.0 arm64 arm

### What steps can reproduce the bug?

`bunx github:cyfung1031/vm-repack-2154 --help`

### What is the expected behavior?

same as `npx github:cyfung1031/vm-repack-2154 --help`
which is clone the project from GitHub, run the bin `vm-repack-2154`

### What do you see instead?

```sh
  🔍 Resolving [1/1] 
error: InstallFailed cloning repository for @github:cyfung1031/vm-repack-215
🔍 Resolving [2/1] 
error: "git clone" for "@github:cyfung1031/vm-repack-2154" failed


error: "git clone" for "@github:cyfung1031/vm-repack-2154" failed
```

PR #6042, opened 2023-09-25
Fix bunx command for github package #5974

axlEscalada:
### What does this PR do?
Fix the `bunx` command when a github repo is passed as an argument. There is a bug where the package name is parsed as an `UpdateRequest` that leaves the name of the package empty and fails when trying to concatenate `name + @ + version`
In bunx_command name and version are checked and if they are equal set the name as the package name, and format it without `@`. Also this fmt package is used to build the path where is installed.

Fixes: https://github.com/oven-sh/bun/issues/5974

- [x] Code changes


src/cli/bunx_command.zig
```diff
@@ -262,22 +262,34 @@ pub const BunxCommand = struct {
         else
             update_request.version.literal.slice(update_request.version_buf);
 
+        var package_fmt: []const u8 = undefined;
```
Jarred-Sumner: `undefined` is dangerous and should be avoided whenever possible

```suggestion
        const package_fmt: []const u8 = brk: {
  if (!strings.eql(update_request.version_buf, update_request.name)) {
            break :brk try std.fmt.allocPrint(
                ctx.allocator,
                "{s}@{s}",
                .{
                    update_request.name,
                    display_version,
                },
            );
        }
        
        break :brk update_request.name;
};
```

Jarred-Sumner:
Thank you for this. One nitpicky comment

src/cli/bunx_command.zig
axlEscalada: fixed

Electroid:
Tested and good-to-go. Thanks!

Issue #6039, opened 2023-09-25
`console.dir` does not respect `depth` option


JacobWeisenburger:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Microsoft Windows NT 10.0.22621.0 x64

### What steps can reproduce the bug?

Bun claims that `node:console` is Fully implemented.
![image](https://github.com/oven-sh/bun/assets/31667350/7a6440c5-761b-4734-99cc-9f257b3bb44b)

However, it seems that this code doesn't do the same thing in Bun as Node.
```ts
console.dir( { 1: { 2: { 3: 3 } } }, { depth: 0 } )
// node: { '1': [Object] }
// bun: { "1": { "2": { "3": 3 } } } { depth: 0 }
```

Is this supposed to do the same thing as Node?

### What is the expected behavior?

`{ '1': [Object] }`

### What do you see instead?

`{ "1": { "2": { "3": 3 } } } { depth: 0 }`

### Additional information

https://bun.sh/docs/runtime/nodejs-apis#node-console
https://nodejs.org/api/console.html#consoledirobj-options

danielo515:
What version of bun has this fixed? I have 1.0.3

PR #6059, opened 2023-09-26
fix: support console.dir options object correctly

liz3:
`console.dir` can be passed a second argument which is a object of options. This implements that logic with the currently supported properties: `depth` and `colors`.
I used node as a reference for implementation details.

Fixes: https://github.com/oven-sh/bun/issues/6039

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -983,22 +983,40 @@ pub const ZigConsoleClient = struct {
         else
             console.writer;
         var writer = buffered_writer.writer();
-
         const Writer = @TypeOf(writer);
-        if (len > 0)
+
+        var print_length = len;
+        var print_options: FormatOptions = .{
+                    .enable_colors = enable_colors,
+                    .add_newline = true,
+                    .flush = true,
+        };
+
+        if (message_type == .Dir and len >= 2) {
+            print_length = 1;
+            var opts = vals[1];
+            if (opts.isObject()) {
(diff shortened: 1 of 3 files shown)


src/bun.js/bindings/exports.zig
```diff
@@ -983,22 +983,40 @@ pub const ZigConsoleClient = struct {
...
+            .enable_colors = enable_colors,
+            .add_newline = true,
+            .flush = true,
+        };
+
+        if (message_type == .Dir and len >= 2) {
+            print_length = 1;
+            var opts = vals[1];
+            if (opts.isObject()) {
+                var depth_prop = opts.get(global, "depth") orelse JSValue.zero;
```
paperclover: instead of `orelse ...`, use `if (opts.get(...)) |depth_prop| {` this way

src/bun.js/bindings/exports.zig
```diff
@@ -983,22 +983,40 @@ pub const ZigConsoleClient = struct {
...
+                if (depth_prop.isNumber())
+                    print_options.max_depth = depth_prop.toU16();
```
paperclover: ~~`toU16` will crash if given a float.~~ actually i am wrong here. can you add a test that passes a float depth, also a negative depth, also `Infinity`.

paperclover (changes requested):

liz3:
@paperdave ive implemented request ed changes

paperclover (approved):
looks good will merge after tests run

paperclover:
thank you

PR #6086, opened 2023-09-26
fix(fetch/server) fix server end of stream, fix fetch not streaming without content-length or chunked encoding, fix case when stream do not return a promise on pull

cirospaciari:
…gth or chunked encoding, fix case when stream do not return a promise on pull

### What does this PR do?

- [x] Code changes

I wrote automated tests + Existing Tests

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -1443,6 +1443,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         }
 
         pub fn endStream(this: *RequestContext, closeConnection: bool) void {
+            ctxLog("endStream", .{});
             if (this.resp) |resp| {
                 if (this.flags.is_waiting_for_request_body) {
                     this.flags.is_waiting_for_request_body = false;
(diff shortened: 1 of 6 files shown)


src/bun.js/api/server.zig
```diff
@@ -2255,7 +2267,12 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
 
             this.setAbortHandler();
             streamLog("is in progress, but did not return a Promise. Finalizing request context", .{});
-            this.finalize();
+            response_stream.sink.onFirstWrite = null;
+            response_stream.sink.ctx = null;
+            response_stream.detach();
+            stream.cancel(globalThis);
```
Jarred-Sumner, line 2273: Does `detach` call cancel?

test/js/web/fetch/body-stream.test.ts
```diff
@@ -13,6 +13,35 @@ var port = 0;
   ];
   const useRequestObjectValues = [true, false];
 
+  test("Should not crash when not returning a promise when stream is in progress", async () => {
```
Jarred-Sumner, line 16: I feel like we need more tests

src/bun.js/api/server.zig
```diff
@@ -2223,6 +2226,14 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                         },
                     }
                     return;
+                } else {
+                    // if is not a promise we treat it as Error
+                    streamLog("returned an error", .{});
+                    if (!this.flags.aborted) resp.clearAborted();
+                    response_stream.detach();
+                    this.sink = null;
+                    response_stream.sink.destroy();
```
Jarred-Sumner, line 2235: I don't feel great about this code

Issue #2875, opened 2023-05-13
`eval` uses global scope instead of local scope


kzc:
### What version of Bun is running?

0.6.0 canary

### What platform is your computer?

macos 10.14

### What steps can reproduce the bug?

given:
```
$ cat e.mjs
var a = "PASS";
function foo(s) {
    eval(`ba${s}()`);
}
function bar() {
    console.log(a);
}
function baz() {
}
export default foo("r");
```
expected:
```
$ node e.mjs 
PASS
```
actual:
```
$ bun run e.mjs
1 | bar()
   ^
ReferenceError: Can't find variable: bar
```

### What is the expected behavior?

No DCE or renaming of any identifiers reachable from `eval`.

### Additional information

Same bug exists in `bun build` and `bun build --minify`.

Jarred-Sumner:
eval is using global scope instead of local scope

but i'm not sure why

kzc:
`eval()` starts resolving identifiers in the local scope from which it is called and then goes up the parent scope chain.  So any identifier that could *potentially* be reached from a given `eval()` cannot be dropped or renamed.

Here's a more illustrative example:

```js
$ cat local-eval.mjs
export default 1; // ES6 module
var a = "MODULE variable";
const foo = arg => console.log("module foo:", arg);
const bar = arg => console.log("module bar:", arg);
function Run(fn, args) {
    var b = "LOCAL variable";
    const bar = arg => console.log("local bar:", arg);
    eval(`${fn}(${args})`);
}
Run("foo", "a");
Run("foo", "b");
Run("bar", "a");
Run("bar", "b");
```
```
$ node local-eval.mjs
module foo: MODULE variable
module foo: LOCAL variable
local bar: MODULE variable
local bar: LOCAL variable
```

MassHesteria:
I was trying to play around with this, but I'm still getting up to speed with zig. The one thing that jumped out to me initially was the `is_direct_eval` member variable in the `Call` struct defined in `js_ast.zig`. From what I can tell that variable is always false which, based on naming, makes me think that we are getting an indirect eval instead of a direct eval. Indirect evals work in the global scope instead of local.

Could be a red herring though. I'm not really seeing where the `is_direct_eval` flag is actually being used relative to JSC which could also be part of the issue.

kzc:
The new issue title "eval uses global scope instead of local scope" doesn't accurately convey all the problems with `eval` handling.

[...]

PR #5932, opened 2023-09-22
`deadCodeElimination` toggle for Bun.Transpiler

jhmaster2000:
### What does this PR do?

Adds a new option for `Bun.Transpiler`, called `deadCodeElimination`, which allows disabling of dead code elimination, which in turn enables proper use of Bun.Transpiler for transpilation of code meant to be top-level evaluated which would otherwise be voided by false-positive DCE due to lack of context to the transpiler. (examples: code passed to `eval()` or in a REPL)

The option is a boolean and is enabled by default for backwards compatibility given the transpiler has been doing DCE by default so far.

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

I wrote automated tests

- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [x] I or my editor ran `zig fmt` on the changed files
- [x] I included a test for the new code, or an existing test covers it
- [ ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed

- [x] I added TypeScript types for the new methods, getters, or setters

PR #6091, opened 2023-09-27
fix change in code behavior with `eval` after building (#2875)

Kyza:
### What does this PR do?

Stops Bun from [changing direct `eval` calls to indirect `(0, eval)` calls](https://github.com/oven-sh/bun/issues/2875), eliminating the differing behavior between what is written and what gets run.

- [x] Code changes

I wrote automated tests.


kzc:
This PR appears to be a partial fix for #2875 but doesn't address the `eval` problems related to `--minify-identifiers` and `--minify-syntax` as detailed in https://github.com/oven-sh/bun/issues/2875#issuecomment-1733029655.

The manual `deadCodeElimination = false` option setting introduced in #5932 might help `eval` in #6091 work with `--minify-syntax` (enabled by default in `bun`?) but I can't seem to get it working from the command line interface.  I don't *think* it helps `eval` work with `--minify-identifiers` but I could be mistaken.

Kyza:
> ... but doesn't address the `eval` problems related to `--minify-identifiers` and `--minify-syntax` as detailed in [#2875 (comment)](https://github.com/oven-sh/bun/issues/2875#issuecomment-1733029655).

I never intended this PR to cover that specific problem since the only experience I have with Bun's codebase is *this PR*. I can, however, at least attempt to resolve it here as well since it's related.

I'll post an update after I give it a shot.

Jarred-Sumner:
no longer an issue

PR #6095, opened 2023-09-27
Get Next.js Pages Router to work

paperclover:
### What does this PR do?

also does asyncLocalStorage.bind also does tty fixes for inquirer also adds an assert that will fail when using next.js but its only a dev one

- [ ] Code changes

because i said so


Jarred-Sumner (approved):

Jarred-Sumner:
-webkit needs to not change
-launch.json changes are prob unintentional

Jarred-Sumner:
@paperdave 

test failures:
  - `test/integration/next/default-pages-dir/test/dev-server.test.ts`
  - `test/js/node/process/process-stdio.test.ts`
  - `test/js/node/stream/node-stream.test.js`

Jarred-Sumner (changes requested):
Can’t merge until tests pass

PR #6090, opened 2023-09-27
fix: Docker - Include `bunx` symlink in distroless variant

polarathene:
### What does this PR do?

Seems [my suggestion](https://github.com/oven-sh/bun/issues/5269#issuecomment-1724856820) was missed. Keeps consistent with other images by including `bunx` symlink.

I saw the merged PR and [commented about making this change](https://github.com/oven-sh/bun/pull/5771/files#r1337887524). I figured I might as well raise a PR for it while I'm at it 😅 

### How did you verify your code works?

Comparision:

```console
# Resolved with single COPY, symlink works:
$ docker save local-bun-build | gzip -c | wc -c | numfmt --to iec
40M

# The duplicate issue that previously existed:
$ docker save oven/bun | gzip -c | wc -c | numfmt --to iec
97M

# Resolved vs original image size:
$ docker images
REPOSITORY                  TAG                IMAGE ID       CREATED          SIZE
local-bun-build             latest             020acf83835e   7 minutes ago    110MB
oven/bun                    latest             647ebb4444b7   3 days ago       271MB
```

Full information: https://github.com/oven-sh/bun/issues/5269#issuecomment-1724856820

Issue #5269, opened 2023-09-13
Docker image consists of two large identical layers for `bun` and `bunx`


marvinruder:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```
docker pull oven/bun:1.0.1

# For detailed analysis:
mkdir bun && cd bun
docker save oven/bun | tar -xz

# Then have a look at the largest files: ./blobs/sha256/{f96a...,7ce0..., 4704...} (all are gzipped tarballs)
```

### What is the expected behavior?

The image is approximately 65 MB large and contains only two larger layers (and some more smaller ones): one is the debian-slim base image (around 30 MB), the other contains the `bun` binary (30–35 MB based on architecture).

### What do you see instead?

The image is 95–100 MB large (depending on architecture).

[Docker Hub](https://hub.docker.com/r/oven/bun/tags):

| DIGEST | OS/ARCH | COMPRESSED SIZE |
| - | - | - |
| 456d511a28e6 | linux/amd64 | 100.52 MB |
| 7190627f958b | linux/arm64 | 94.72 MB |

[Docker Hub Image Details](https://hub.docker.com/layers/oven/bun/latest/images/sha256-456d511a28e648b028180b5c3b0f21b878626c0520647611c543ed075a974b89?context=explore):

| Step | Command | Size |
| - | - | - |
| 1 | `ADD file ... in /` | 29.96 MB |
| 2 | `CMD ["bash"]` | 0 B |
| 3 | `RUN /bin/sh -c groupadd bun` | 3.98 KB |
| 4 | `COPY docker-entrypoint.sh /usr/local/bin # buildkit` | 294 B |
| 5 | `COPY /usr/local/bin/bun /usr/local/bin # buildkit` | 35.28 MB |
| 6 | `COPY /usr/local/bin/bunx /usr/local/bin # buildkit` | 35.28 MB |
| 7 | `WORKDIR /home/bun/app` | 140 B |
| 8 | `ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]` | 0 B |
| 9 | `CMD ["/usr/local/bin/bun"]` | 0 B |

Output during pull:

```
Using default tag: latest
bdf51429e651: Download complete 
7190627f958b: Download complete 
056088840a56: Download complete 
[...]

PR #6100, opened 2023-09-27
fix: Docker - Apply workaround with `RUN` to symlink `bunx`

polarathene:
### What does this PR do?

In the [previous attempt to resolve this](https://github.com/oven-sh/bun/pull/6090), I mistakenly missed a `/` for the destination and did not properly test `bunx` actually ran.

This was resolved, but [doubled the size again](https://github.com/oven-sh/bun/issues/5269#issuecomment-1736887598) as Docker is resolving the symlink when copying 😞 

I [remembered a technique I proposed for `testssl.sh`](https://github.com/drwetter/testssl.sh/pull/2344/commits/1392987f01b748021cda4e98cbaf8e151c2f9a32#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R69-R84) when revising their `Dockerfile`. Even though it was only required for building the images, the [maintainer decision was the features used were too new](https://github.com/drwetter/testssl.sh/issues/2299#issuecomment-1479094496):
- `RUN --mount` requires a [Docker Engine release from 2022H2 (with BuildKit)](https://github.com/drwetter/testssl.sh/issues/2299#issuecomment-1479136828).
- HereDoc (_[introduced support in 2021H2](https://github.com/drwetter/testssl.sh/issues/2299#issuecomment-1478863039)_).

---

Using [`RUN --mount`](https://docs.docker.com/engine/reference/builder/#run---mounttypebind), we can run the command with the build stage files overlayed for the `ln` and `which` commands.

**NOTE:** The mounted `/bin` is a symlink to `/usr/bin`, both seem required to work correctly.

Runs with the [same output now as the `bun x` command](https://github.com/oven-sh/bun/issues/5269#issuecomment-1736916694):

```console
$ docker buildx build -t bun-small .

$ docker run --rm -it --entrypoint /usr/local/bin/bunx bun-small eslint --version
error: Failed to run "eslint" due to error FileNotFound
```

But is still smaller:
```console
$ docker images
REPOSITORY   TAG          IMAGE ID       CREATED         SIZE
bun-small    latest       387d3d1f0c9f   3 minutes ago   110MB
oven/bun     distroless   f649763a2ca5   5 hours ago     205MB

$ docker save bun-small | gzip -c | wc -c | numfmt --to iec
40M
```

### DIFF
--- a/dockerhub/distroless/Dockerfile
+++ b/dockerhub/distroless/Dockerfile
@@ -52,16 +52,18 @@ RUN apt-get update -qq \
     && mv "bun-linux-$build/bun" /usr/local/bin/bun \
     && rm -f "bun-linux-$build.zip" SHASUMS256.txt.asc SHASUMS256.txt \
     && chmod +x /usr/local/bin/bun \
-    && ln -s /usr/local/bin/bun /usr/local/bin/bunx \
     && which bun \
-    && which bunx \
     && bun --version
 
 FROM gcr.io/distroless/base-nossl-debian11
 
-# List of sources to destination (final path):
-COPY --from=build \
-  /usr/local/bin/bun /usr/local/bin/bunx \
-  /usr/local/bin/
+COPY --from=build /usr/local/bin/bun /usr/local/bin/
+
+# Temporarily use the `build`-stage image binaries to create a symlink:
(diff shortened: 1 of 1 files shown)


Electroid:
Awesome, thanks for digging into this and for the fix.

PR #6112, opened 2023-09-27
fix(bun install): Handle vercel and github tarball path dependencies

booniepepper:
package.json dependency entries like:

```
"when": "https://github.com/cujojs/when/tarball/1.0.2"
```

and

```
"@vercel/turbopack-ecmascript-runtime": "https://gitpkg-fork.vercel.sh/vercel/turbo/crates/turbopack-ecmascript-runtime/js?turbopack-230922.2"
```

are both links to tarballs. In the github case, this is a URL pattern, and in the vercel case, this seems to be everything served from their server.

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/install/dependency.zig
+++ b/src/install/dependency.zig
@@ -233,7 +233,7 @@ pub inline fn isGitHubTarballPath(dependency: string) bool {
 
     while (parts.next()) |part| {
         n_parts += 1;
-        if (n_parts == 3 and !part.eql("tarball")) {
+        if (n_parts == 3 and !strings.eql(part, "tarball")) {
             return false;
         } else if (n_parts > 4) {
             return false;
--- a/src/install/dependency.zig
+++ b/src/install/dependency.zig
@@ -223,11 +223,11 @@ pub inline fn isGitHubRepoPath(dependency: string) bool {
 
 // Github allows for the following format of URL:
 // https://github.com/<org>/<repo>/tarball/<ref>
-// This is a legacy (but still supported) method of retrieiving a tarball of an
+// This is a legacy (but still supported) method of retrieving a tarball of an
 // entire source tree at some git reference. (branch, tag, etc)
 // The function expects "https://github.com/" is already stripped from the string
 pub inline fn isGitHubTarballPath(dependency: string) bool {
-    var parts = dependency.split("/");
+    var parts = strings.split(dependency, "/");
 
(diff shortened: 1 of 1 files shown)


src/install/dependency.zig
```diff
@@ -221,6 +221,28 @@ pub inline fn isGitHubRepoPath(dependency: string) bool {
     return hash_index != dependency.len - 1 and first_slash_index > 0 and first_slash_index != dependency.len - 1;
 }
 
+// Github allows for the following format of URL:
+// https://github.com/<org>/<repo>/tarball/<ref>
+// This is a legacy (but still supported) method of retrieiving a tarball of an
+// entire source tree at some git reference. (branch, tag, etc)
+// The function expects "https://github.com/" is already stripped from the string
+pub inline fn isGitHubTarballPath(dependency: string) bool {
+    var parts = dependency.split("/");
```
booniepepper:
```suggestion
    var parts = strings.split(dependency, "/");
```

src/install/dependency.zig
```diff
@@ -221,6 +221,28 @@ pub inline fn isGitHubRepoPath(dependency: string) bool {
     return hash_index != dependency.len - 1 and first_slash_index > 0 and first_slash_index != dependency.len - 1;
 }
 
+// Github allows for the following format of URL:
+// https://github.com/<org>/<repo>/tarball/<ref>
+// This is a legacy (but still supported) method of retrieiving a tarball of an
```
booniepepper:
```suggestion
// This is a legacy (but still supported) method of retrieving a tarball of an
```

src/install/dependency.zig
```diff
@@ -221,6 +221,28 @@ pub inline fn isGitHubRepoPath(dependency: string) bool {
...
+// entire source tree at some git reference. (branch, tag, etc)
+// The function expects "https://github.com/" is already stripped from the string
+pub inline fn isGitHubTarballPath(dependency: string) bool {
+    var parts = strings.split(dependency, "/");
+
+    var n_parts: usize = 0;
+
+    while (parts.next()) |part| {
+        n_parts += 1;
+        if (n_parts == 3 and !part.eql("tarball")) {
```
booniepepper:
```suggestion
        if (n_parts == 3 and !strings.eql(part, "tarball")) {
```

Issue #5812, opened 2023-09-20
feat: add local tgz via `bun add local.tgz`


OrbisK:
### What is the problem this feature would solve?

Currently you can't install local tgz packages (at least not that I know of).

### What is the feature you are proposing to solve the problem?

`bun add example.tgz`
`bun add ./example.tgz`

### What alternatives have you considered?

Current behavior:
`error: package "example.tgz" not found registry.npmjs.org/example.tgz 404`

Electroid:
Fixed by #6118

PR #6118, opened 2023-09-27
Add local tarball install #5812

axlEscalada:
…all it, test for local tarball

### What does this PR do?
- Check if dependency is a tarball before check if it is an npm package
- Alloc `url` to avoid losing tiny string when task is read.

Related to: https://github.com/oven-sh/bun/issues/5812

- [x] Code changes


test/cli/install/bun-add.test.ts
```diff
@@ -1627,3 +1627,53 @@ it("should add dependency alongside peerDependencies", async () => {
...
+    JSON.stringify({
+      name: "foo",
+      version: "0.0.1",
+    }),
+  );
+  const tarball = "baz-0.0.3.tgz"
+  await copyFile(tarball, join(package_dir, tarball));
+  const { stdout, stderr, exited } = spawn({
+    // cmd: ["pwd"],
+    cmd: ["/Users/axel.escalada/bun-dev/packages/debug-bun-darwin-aarch64/bun-debug", "add", tarball],
```
Jarred-Sumner:
```suggestion
    cmd: [bunExe(), "add", tarball],
```

Jarred-Sumner (approved):
Thank you

Jarred-Sumner:
looks like the file path to the tarball is not quite right
```
     path: "baz-0.0.3.tgz"
   syscall: "stat"
     errno: -2
  
  ✗ should add local tarball dependency [0.56ms]
  ```

axlEscalada:
> looks like the file path to the tarball is not quite right
> 
> ```
>   ENOENT: No such file or directory
>      path: "baz-0.0.3.tgz"
>    syscall: "stat"
>      errno: -2
>   
>   ✗ should add local tarball dependency [0.56ms]
> ```

Ok, I was running the test inside of the test folder and it was trying to copy the file using a relative path. Now the test uses an absolute path to copy the tarball file, so the test passes from anywhere.

Jarred-Sumner:
Thank you

PR #6122, opened 2023-09-27
fix(bun install): Handle vercel and github tarball path dependencies

booniepepper:
### What does this PR do?

- fix(bun install): Handle vercel and github tarball path dependencies

Follows a similar approach (in this context) of [checking if a dependency is a tarball before checking github](https://github.com/pnpm/pnpm/blob/main/resolving/default-resolver/src/index.ts#L30-L35), and [filtering out repositories based on number of parts](https://github.com/pnpm/pnpm/blob/main/resolving/tarball-resolver/src/index.ts#L10)

- [x] Code changes

### DIFF
--- a/src/install/dependency.zig
+++ b/src/install/dependency.zig
@@ -224,8 +224,8 @@ pub inline fn isGitHubRepoPath(dependency: string) bool {
 // Github allows for the following format of URL:
 // https://github.com/<org>/<repo>/tarball/<ref>
 // This is a legacy (but still supported) method of retrieving a tarball of an
-// entire source tree at some git reference. (branch, tag, etc)
-// The function expects "https://github.com/" is already stripped from the string
+// entire source tree at some git reference. (ref = branch, tag, etc. Note: branch
+// can have arbitrary number of slashes)
 pub inline fn isGitHubTarballPath(dependency: string) bool {
     var parts = strings.split(dependency, "/");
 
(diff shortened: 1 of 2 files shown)


booniepepper:
Plan to add tests for these cases too, based on these two real-world dependencies:

```
"@vercel/turbopack-ecmascript-runtime": "https://gitpkg-fork.vercel.sh/vercel/turbo/crates/turbopack-ecmascript-runtime/js?turbopack-230922.2",
"when": "https://github.com/cujojs/when/tarball/1.0.2"
```

Both should be handled as tarballs

booniepepper:
There's another test that could be added based on a real world dependency:

```
"@vercel/turbopack-ecmascript-runtime": "https://gitpkg-fork.vercel.sh/vercel/turbo/crates/turbopack-ecmascript-runtime/js?turbopack-230922.2"
```

but I'm observing that the HTTP request for this can be long (more than seconds)

dylan-conway (approved):
LGTM. I'll let the tests run and add the other test after

dylan-conway:
Thanks!

booniepepper:
Thanks too that was a blast!

booniepepper:
> LGTM. I'll let the tests run and add the other test after

I added it here https://github.com/oven-sh/bun/pull/6190

PR #6128, opened 2023-09-28
packages:bun-types:add-type-parameter-to-expect

Voldemat:
### What does this PR do?

This adds type parameter to expect, so Expect structure could infer input type parameter and warn user if he tries for example compare string and number.

- [ ] Code changes


Voldemat:
Hi, when you will have time to review pull request?

Electroid:
Thanks for working on this!

PR #6151, opened 2023-09-28
fix Response.statusText

### LINKED ISSUE #5753: Response.statusText is empty
```js
console.log(
    new Response("foo", {
        statusText: "ok",
        status: 200
    }).statusText
);
```

Appears empty, seems like a bug?

_Originally reported on Discord: [`Ask Inkeep`](https://discord.com/channels/876711213126520882/1153742219480662068/1153742220948680834)_

toshok:
### What does this PR do?

Restructures `Response`/`Body` a little, moving `Body.Init` under `Response`.  Also adds `statusText` handling to `Response.Init` (which fixes the linked bug).  This had a few other knock-on effects, like moving `@"200"`/`@"404"` out of `Body` (again to `Response`.)

fixes #5753

- [x] Code changes

I wrote a some tests to handle various cases.


src/bun.js/bindings/bindings.cpp
```diff
@@ -4114,6 +4115,9 @@ static JSC::Identifier builtinNameMap(JSC::JSGlobalObject* globalObject, unsigne
     case BuiltinNamesMap::headers: {
         return clientData->builtinNames().headersPublicName();
     }
+    case BuiltinNamesMap::statusText: {
+        return clientData->builtinNames().statusTextPublicName();
+    }
```
toshok, line 4128: all of the bindings{.cpp/.zig} changes were very much dogscience.  are there tests or things I should worry about?

toshok:
a lot of the code motion was due to me trying to make things clearer for myself while I fixed the bug.  I could open a different PR that only addresses the bug and save the code motion for after, if you'd prefer.  lemme know.

src/bun.js/webcore/response.zig
```diff
@@ -172,7 +178,7 @@ pub const Response = struct {
     }
 
     pub fn isOK(this: *const Response) bool {
-        return this.body.init.status_code == 304 or (this.body.init.status_code >= 200 and this.body.init.status_code <= 299);
+        return this.init.status_code == 304 or (this.init.status_code >= 200 and this.init.status_code <= 299);
```
Electroid: Unrelated to your changes, but might as well fix it. `response.ok` is only suppose to be true for `[200,300)`. This was a mistake when implemented initially.

CONTRIBUTING.md
```diff
@@ -10,7 +10,7 @@ Today (February 2023), Bun's codebase has five distinct parts:
 
 - JavaScript, JSX, & TypeScript transpiler, module resolver, and related code
 - JavaScript runtime ([`src/bun.js/`](src/bun.js/))
-- JavaScript runtime bindings ([`src/bun.zig/bindings/**/*.cpp`](src/bun.zig/bindings/))
```
Jarred-Sumner: nice

src/bun.js/bindings/bindings.cpp
Jarred-Sumner: this is right, thanks

context here: `Identifier::fromString` is super slow. It frequently shows up in profiles. It has a lock, it clones a string and hashes it each call

Jarred-Sumner (approved):
Looks good. I'm pretty confident we can merge it, but lets let the tests run. There will be a handful of unrelated test failures, mostly due to permissions but also because CI isn't currently green (especially on macOS)

Jarred-Sumner:
Looks like there are some new test failures

toshok:
☝️ rebased off main to clear up the conflict.  incoming two commits fix the test failures (`js/bun/http` passes now) and addresses the suggested change above.

src/bun.js/webcore/response.zig
toshok: done

Jarred-Sumner:
Don't worry about the `cp.test` failure, fixed in 54d25b8f6b0f30b07146162a15d5f500fd45b2ee

Jarred-Sumner:
`inspect.test` needs to be updated

paperclover:
image

Ordering of these fields changed (inspect.test.js). I think it's okay if these tests are updated.

Electroid:
Thanks for working on this!

PR #6153, opened 2023-09-28
Fix minimum kernel version in docs

klatka:
### What does this PR do?

- [ ] Code changes

### DIFF
--- a/docs/cli/install.md
+++ b/docs/cli/install.md
@@ -9,7 +9,7 @@ The `bun` CLI contains a Node.js-compatible package manager designed to be a dra
 {% /callout %}
 
 {% details summary="For Linux users" %}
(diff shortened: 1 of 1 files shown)


v3nntrnx:
LGTM

colinhacks:
I think the current version is more correct, unless there are some show-stopping errors when using a pre-5.6 version?

klatka:
> I think the current version is more correct, unless there are some show-stopping errors when using a pre-5.6 version?

We cannot say the minimum version is 5.1 and at the same time 5.1 - 5.5 **should** work.

paperclover (approved):

paperclover (changes requested):
i think this should be worded as "recommended" because saying minimum is something then saying something below it also works is a contradiction.

paperclover (approved):
@colinhacks wdyt?

colinhacks:
Went with "recommended minimum" lol

Issue #6144, opened 2023-09-28
fetch() request with GET/HEAD/OPTIONS method cannot have body


dimonnwc3:
### What version of Bun is running?

1.0.4+387f1260c9dc0cea667b44ec0152fff0cd4def25

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

using https://github.com/sindresorhus/ky library, which is using fetch under the hood

Following code throws an error:

```
const res = await ky.post(baseURL("/api/v1/countries"), {
      json: { email: "aaaaaaa", name: "a" },
})
```

### What is the expected behavior?

it should not throw an error and send request

### What do you see instead?

```
TypeError: fetch() request with GET/HEAD/OPTIONS method cannot have body.
code: "ERR_INVALID_ARG_VALUE"
```

### Additional information

related issue: https://github.com/sindresorhus/ky/issues/527

liz3:
Interesting i want to look a this

liz3:
@dimonnwc3 This seams to be a wanted behaviour, executing a get request with a body in node produces the same error.
As a result i would assume its part of the spec for fetch.
I don't think this is inherently a issue since even though the HTTP 1.1 standard can have body contents on any request method it is very untypical to send GET/HEAD/OPTIONS with a body.

dimonnwc3:
@liz3 it supposed to be a POST request

here is a one line simplification

```
import ky from "ky"

await ky
  .post("https://httpbin.org/anything", { json: { hello: "world" } })
  .json()

```

this code works in node, but doesn't in bun

here is also test for node, which passes: https://github.com/sindresorhus/ky/issues/527#issuecomment-1739038827

liz3:
Il check again thanks

vinyll:
I am having this issue despite I just upgraded to Bun 1.0.4:
```
TypeError: fetch() request with GET/HEAD/OPTIONS method cannot have body.
```

Is this coming with the next release or maybe should I compile somehow the current _main_ branch?

liz3:
> I am having this issue despite I just upgraded to Bun 1.0.4:
> 
> ```
> TypeError: fetch() request with GET/HEAD/OPTIONS method cannot have body.
> ```
> 
> Is this coming with the next release or maybe should I compile somehow the current _main_ branch?

can you post example code, il have a look at it then

vinyll:
Here's how to reproduce:

```bash
[...]

PR #6154, opened 2023-09-28
fix: don't set default request method when creating a Request from another

liz3:
In the case of creating a Request with the parameters `(Request, object)`, there was a bug that method and headers are set from the default created by the init rather then the already present value from the request param. This is because for a to me unknown reason the order in which the parameters are processed is reversed.

This fixes that by adding a check which stops the defaults from being set, unless they are explicitly passed, but only in this specific case.

Fixes: https://github.com/oven-sh/bun/issues/6144

### What does this PR do?

- [x] Code changes


liz3:
is there anything i need to do before this can be merged?

Jarred-Sumner:
Looks good. Thanks for following up

PR #6167, opened 2023-09-29
Implement virtual module support in `Bun.plugin`

Jarred-Sumner:
### What does this PR do?

Define a virtual module that can be imported or required 

Define a plugin:
```js
import { plugin } from "bun";
plugin({
  name: "i am virtual!",
  setup(builder) {
    builder.module("my-virtual-module-async", async () => {
      // check
      await Bun.sleep(1);
      return {
        exports: {
          hello: "world",
        },
        loader: "object",
      };
    });

    builder.module("my-virtual-module-sync", () => {
      return {
        exports: {
          hello: "world",
        },
        loader: "object",
      };
    });
  },
});
```

Example usage:
```ts
  it("modules are overridable", async () => {
    // @ts-expect-error
    let { hello, there } = await import("my-virtual-module-sync");
    expect(there).toBeUndefined();
    expect(hello).toBe("world");

    Bun.plugin({
      setup(builder) {
        builder.module("my-virtual-module-sync", () => ({
          exports: {
            there: true,
          },
          loader: "object",
        }));
      },
    });

    {
      const { there, hello } = require("my-virtual-module-sync");
      expect(there).toBe(true);
      expect(hello).toBeUndefined();
    }

    Bun.plugin({
      setup(builder) {
        builder.module("my-virtual-module-sync", () => ({
          exports: {
            yo: true,
          },
          loader: "object",
        }));
      },
    });

    {
      // @ts-expect-error
      const { there, hello, yo } = await import("my-virtual-module-sync");
      expect(yo).toBe(true);
      expect(hello).toBeUndefined();
      expect(there).toBeUndefined();
    }
  });
```

Tests


jimmywarting:
Erm... this is not how i would go about creating dynamic modules.

This is not how browser works... it feels too Bun.js specific (where as i prefer cross compatible code that can run anywhere)

```js
var code = 'export const now = Date.now'
var blob = new Blob([code], { type: 'text/javascript' })
var url = URL.createObjectURL(blob)
var { now } = await import(url)

now() // 1696002828764
```

src/bun.js/bindings/BunPlugin.cpp
```diff
@@ -86,6 +88,76 @@ static EncodedJSValue jsFunctionAppendOnLoadPluginBody(JSC::JSGlobalObject* glob
...
+
+    String moduleId = moduleIdValue.toWTFString(globalObject);
+    RETURN_IF_EXCEPTION(scope, encodedJSValue());
+
+    if (moduleId.isEmpty()) {
+        throwException(globalObject, scope, createError(globalObject, "virtual module cannot be blank"_s));
+        return JSValue::encode(jsUndefined());
+    }
+
+    if (Bun::isBuiltinModule(moduleId)) {
```
dylan-conway, line 122: what happens if `moduleId` is `bun:main` or `bun:test`?

dylan-conway (approved):

PR #6185, opened 2023-09-29
fix(runtime): followup for `server.requestIP`

paperclover:
### What does this PR do?

Mainly handles Unix Socket cases. In node:http, `socket()` should return {} on unix socket. In `server.requestIp` this returns null.

- [X] Code changes

Adds a test for unix sockets.


src/deps/libuwsockets.cpp
```diff
@@ -4,6 +4,7 @@
 #include <bun-uws/src/AsyncSocket.h>
 #include <bun-usockets/src/internal/internal.h>
 #include <string_view>
+#include <c-ares/include/ares.h>
```
Jarred-Sumner:
```suggestion
#include <ares.h>
```

PR #6186, opened 2023-09-29
Static frequency for os.cpus(); affinity and cgroup support for os.availableParallelism()

ruihe774:
### What does this PR do?

In this PR:
- `os.cpus()` gets static CPU frequency from `/proc/cpuinfo` as a fallback if dynamic CPU frequency from `/sys/devices/system/cpu/cpu{}/cpufreq/scaling_cur_freq` is not available, which is a common scenario in VM or container. The static CPU frequency is compatible with Node.js.
- `os.availableParallelism()` now recognize CPU affinity mask got from `sched_getaffinity()` and cgroup v2 CPU quota got from `cpu.max` controller. This is compatible with Node.js and Rust. 

- [x] Code changes

Existing tests covered `os.cpus()` and `os.availableParallelism()`. I verified the features added in this PR by manual testing. Because testing this PR involves system-wide or process-wide setting, it is difficult to write automatic test. If you have a solution please tell me :)

JavaScript/TypeScript modules or builtins changed:

Zig files changed:


paperclover:
i think we should update `navigator.hardwareConcurrency` to match this behavior, right?

ruihe774:
> i think we should update `navigator.hardwareConcurrency` to match this behavior, right?

I don't know the semantics of `navigator.hardwareConcurrency` in this scenario. It seems that Web spec does not define what to do with affinity or cgroup. My opinion is to keep it as is.

ruihe774:
I’ve tried Firefox and Deno. Firefox does not regard affinity or cgroup, while Deno does.

Jarred-Sumner:
Closing due to bitrot

PR #6190, opened 2023-09-30
test(bun install): cover http(s) non-github case

booniepepper:
### What does this PR do?

Adds a test that was missing in https://github.com/oven-sh/bun/pull/6122

- [ ] Code changes
- [x] Just a test case

Ran locally


booniepepper:
FYI @dylan-conway

Issue #4561, opened 2023-09-08
Package manager doesn't works on linux or is too slow


iddar:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Linux 6.4.12-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

run `bun install` or `bunx bun-repl`

![image](https://github.com/oven-sh/bun/assets/199103/f65eba44-2b83-42fb-8988-a72177027e04)

### What is the expected behavior?

works 

### What do you see instead?

include bun crash file output after run ` bun create elysia app` same output for `bun install`
```
error: ConnectionRefused

----- bun meta -----
Bun v0.8.1 (16b4bf34) Linux x64 #1 SMP PREEMPT_DYNAMIC Thu, 24 Aug 2023 00:38:14 +0000
CreateCommand: 
Elapsed: 129772ms | User: 0ms | Sys: 9ms
RSS: 34.61MB | Peak: 20.76MB | Commit: 34.61MB | Faults: 0
----- bun meta -----

0   0x55ee96152bbb
1   ???
2   ???
3   ???
4   ???
5   ???
6   ???

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord

```

### Additional information

I already tried doing a clean installation and changing my dns and nothing worked, with other package managers it works fine.

My internet provider is Starlink and I use zsh.

If I can help in any way I am available.

Thanks for such a great project.

Hanaasagi:
Related to #4507.

Crash after `ConnectionRefused` when installing package or reading manifest. A simple way to reproduce this crash is

```Bash
http_proxy=http://127.0.0.1:1087 https_proxy=http://127.0.0.1:1087 bun create elysia app
```

This proxy doesn't exist, and the port is not being listened to by any program.

Issue #4066, opened 2023-08-08
bun install takes extremely long time


alyiev:
### What platform is your computer?

Darwin 22.6.0 arm64 arm

### How did you attempt to install or upgrade?

bun install

### What do you see instead?

bun install
bun install v0.7.3 (a9b3d583)
 + @babel/core@7.22.10
 + @babel/plugin-proposal-export-namespace-from@7.18.9
 + @tamagui/babel-plugin@1.50.2
 + @types/lodash-es@4.17.8
 + @types/ramda@0.28.25
 + @types/react@18.2.19
 + @types/uuid@9.0.2
 + babel-plugin-transform-inline-environment-variables@0.4.4
 + eslint@8.46.0
 + eslint-config-airbnb-typescript-prettier@5.0.0
 + eslint-import-resolver-alias@1.1.2
 + eslint-plugin-import@2.28.0
 + eslint-plugin-prettier@5.0.0
 + expo-dev-client@2.4.6
 + prettier@3.0.1
 + react-devtools@4.28.0
 + typescript@5.1.6
 + @expo-google-fonts/andika@0.2.3
 + @expo/vector-icons@13.0.0
 + @reduxjs/toolkit@1.9.5
 + @tamagui/animations-moti@1.50.2
 + @tamagui/animations-react-native@1.50.2
 + @tamagui/config@1.50.2
 + @tamagui/shorthands@1.50.2
 + @tamagui/themes@1.50.2
 + expo@49.0.6
 + expo-constants@14.4.2
 + expo-font@11.4.0
 + expo-linking@5.0.2
 + expo-location@16.1.0
 + expo-random@13.2.0
 + expo-router@2.0.0
 + expo-splash-screen@0.20.5
 + expo-status-bar@1.6.0
 + immer@10.0.2
 + lodash-es@4.17.21
 + ramda@0.28.0
 + react@18.2.0
 + react-dom@18.2.0
 + react-native@0.72.3
 + react-native-gesture-handler@2.12.1
 + react-native-get-random-values@1.8.0
 + react-native-maps@1.7.1
 + react-native-mmkv@2.10.1
 + react-native-reanimated@3.3.0
 + react-native-safe-area-context@4.6.3
 + react-native-screens@3.22.1
 + react-native-url-polyfill@1.3.0
 + react-native-web@0.19.7
 + react-redux@8.1.2
 + redux-persist@6.0.0
 + tamagui@1.50.2
 + zustand@4.4.1

[...]

PR #6192, opened 2023-09-30
Fix hang in `bun install`

Jarred-Sumner:
### What does this PR do?

- Fixes a hang in the HTTP client that could occur in `bun install` on installs with manifest cached
- Fixes an infinite loop that could occur when

Fixes https://github.com/oven-sh/bun/issues/5831
Fixes https://github.com/oven-sh/bun/issues/4066
Fixes https://github.com/oven-sh/bun/issues/4561 (msot likely)

Tests

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -1906,3 +1906,35 @@ pub inline fn serializableInto(comptime T: type, init: anytype) T {
 
     return result.*;
 }
+
+/// Like std.fs.Dir.makePath except instead of infinite looping on dangling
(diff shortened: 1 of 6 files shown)


imran-khani:
it still hang in installing nextjs and take too long to install packages

mxvsh:
it still hangs when i run `bun init`

prabhatgpal:
It just hung for me. Not sure if this is resolved
I'm on Mac M1

PR #6207, opened 2023-10-01
fix(node:process): fix return value of `process.kill`

Hanaasagi:
### What does this PR do?

`process.kill` should return `true` on success, not `undefined`. It affected some code used to test for the existence of a process. e.g.
 
```JavaScript
function processExist(pid) {
  try {
    return kill(pid, 0);
  } catch {
    return false;
  }
}

console.log(processExist(process.pid));
console.log(processExist(234234));
```

Ref: https://github.com/nodejs/node/blob/092fb9f541ce8cc07289b5a69eb93892445739f5/lib/internal/process/per_thread.js#L211-L237

- [x] Code changes

I wrote automated tests

### DIFF
--- a/packages/bun-types/globals.d.ts
+++ b/packages/bun-types/globals.d.ts
@@ -681,7 +681,7 @@ interface Process {
    */
   setSourceMapsEnabled(enabled: boolean): void;
 
-  kill(pid: number, signal?: string | number): void;
+  kill(pid: number, signal?: string | number): true;
 
   on(event: "beforeExit", listener: BeforeExitListener): this;
   // on(event: "disconnect", listener: DisconnectListener): this;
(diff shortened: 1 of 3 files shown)


packages/bun-types/globals.d.ts
```diff
@@ -681,7 +681,7 @@ interface Process {
    */
   setSourceMapsEnabled(enabled: boolean): void;
 
-  kill(pid: number, signal?: string | number): void;
+  kill(pid: number, signal?: string | number): true;
```
Hanaasagi, line 684: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/00ee3b9fac18dff2549f4dd0fe02b66e7a56b3b9/types/node/process.d.ts#L1034

Jarred-Sumner (approved):
LGTM, thanks

PR #6208, opened 2023-10-01
String response for lambda function - no need to stringify string

vladaman:
### What does this PR do?

Lambda string response doesn't need to be stringified. This is especially important when we want to respond to services such as Congito, which require JSON output.

- [x] Code changes

Added type check for string:

```javascript
body: response === null ? null : (typeof response === 'string' ? response : JSON.stringify(response)),
```


Jarred-Sumner:
Thank you

Issue #5930, opened 2023-09-22
Change default `fetch` timeout to be longer


di-danielngo:
### What is the problem this feature would solve?

Currently `fetch` will automatically timeout after 30 secs, which is not really ideal for certain time consuming services. Even chrome has a default of 5 minutes.

### What is the feature you are proposing to solve the problem?

Is there any way to set the timeout duration in fetch option?

### What alternatives have you considered?

There are some hacky ways in the node discussions, however if Bun has a canon api it would be so much better.

Jarred-Sumner:
Fixed by #6217

morpig:
is it possible to set timeout seconds from fetch function? #6217 PR seems to be hardcoded from bun's side.

in my case, need a 5-10 seconds fetch timeout for payments-related stuffs.

Issue #4799, opened 2023-09-10
bun install fails on slow registries


carneiror:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```
➜  bun-test time bun install axe-core
bun add v1.0.0 (822a00c4)
  🔍 axe-core [6/6]
error: Timeout downloading package manifest axe-core
error: axe-core@ failed to resolve
bun install axe-core  0.05s user 0.14s system 0% cpu 6:00.05 total
```

### What is the expected behavior?

```
➜  bun-test time npm install axe-core
npm install axe-core  4.31s user 0.53s system 3% cpu 2:20.91 total
```

### Additional information

This registry and the package in question are extremely slow (18mb and take for than 1min to download). Bun do 6 retries, 1 minute each and fails every time.

I tried set higher values for [fetch-retry-maxtimeout](https://docs.npmjs.com/cli/v10/using-npm/config#fetch-retry-maxtimeout), [fetch-retry-mintimeout](https://docs.npmjs.com/cli/v10/using-npm/config#fetch-retry-mintimeout) and [fetch-timeout](https://docs.npmjs.com/cli/v10/using-npm/config#fetch-timeout) without any success.

This is an issue with my current private registry but should be also an issue for people with really bad connections.

tomconder:
+1 my use case is similar: timeout fetching axe-core from a slow private registry.

Jarred-Sumner:
Fixed by #6217

PR #6217, opened 2023-10-01
Set `fetch` timeout to 5 minutes

Jarred-Sumner:
### What does this PR do?

- Set the default timeout for `fetch` to 5 minutes instead of 60 seconds

Existing tests

### DIFF
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -771,7 +771,7 @@ pub const Listener = struct {
             Socket.dataSetCached(this_socket.getThisValue(globalObject), globalObject, default_data);
         }
         socket.ext(**anyopaque).?.* = bun.cast(**anyopaque, this_socket);
-        socket.timeout(120000);
+        socket.setTimeout(120000);
     }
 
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
Fixes https://github.com/oven-sh/bun/issues/4799
Fixes https://github.com/oven-sh/bun/issues/5930

remorses:
Why do you need a default timeout?

Jarred-Sumner:
> Why do you need a default timeout?

Processes generally shouldn't hang forever. The timeout is not an absolute number, it is since the last time a message was received. You can disable the timeout for `fetch` with `timeout: false`

empz:
> > Why do you need a default timeout?
> 
> Processes generally shouldn't hang forever. The timeout is not an absolute number, it is since the last time a message was received. You can disable the timeout for `fetch` with `timeout: false`

Sorry to post here, but where do you set this? 

```ts
Bun.fetch(url, {
  method: "POST",
  body: someBody,
  timeout: false // Typescript doesn't like
}
```
image


Is it a types problem and the argument actually works??

Issue #3944, opened 2023-08-03
`os.freemem()` node compatibility


Hanaasagi:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Linux 6.4.7-arch1-2 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const os = require("node:os");

const free_memory = os.freemem();
const free_mem_in_gb = Math.floor(free_memory / 1024 / 1024 / 1024);

console.log("Free memory: " + free_mem_in_gb + "GB");

```

### What do you see instead?

![2023-08-03_20-41](https://github.com/oven-sh/bun/assets/9482395/bd47021d-1279-4ca6-a8ee-e3c9ed3af70c)

### Additional information

- node version: v18.16.0

It seems that bun uses the `free` field, while node uses the `available` field.

pcapp:
I am taking a look at this. Based on [PR 4012](https://github.com/oven-sh/bun/pull/4012) commentary and my own digging, I see Node.js using libuv to do the the following:
1. Attempting to grab the total memory from /proc/meminfo
2. Falling back to a sysinfo call and using data from the returned struct to calculate the total memory.

PR #6244, opened 2023-10-02
Read /proc/meminfo first.

pcapp:
### What does this PR do?

This commit makes node:os freemem on Linux behave like Node.js. 

Node.js uses libuv's `uv_get_free_memory`, which first calls `uv__read_proc_meminfo` to read the available memory from /proc/meminfo. If reading /proc/meminfo fails, then the libuv implementation falls back to a `sysinfo` call. This commit emulates this behavior.

This fixes [#3944](https://github.com/oven-sh/bun/issues/3944).

- [x] Code changes

It is possible to mock /proc/meminfo, but this requires elevated priviledges. One can create the desired /proc/meminfo values and store them in a file. 

```
#!/bin/bash
unshare --mount bash -c '
  mount --bind ./cooked_memavailable.txt /proc/meminfo
  free -k
'
```

### DIFF
--- a/src/linux_c.zig
+++ b/src/linux_c.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const bun = @import("root").bun;
+const heap_allocator = bun.default_allocator;
 pub const SystemErrno = enum(u8) {
     SUCCESS = 0,
     EPERM = 1,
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Closing due to bitrot + lack of test coverage

PR #6247, opened 2023-10-02
Fix `bun install` reading Github API from wrong environment variable

Electroid:
### What does this PR do?

`bun install` would previously read `GITHUB_API_DOMAIN` to check for Github's API. This environment variable does not exist in Github Actions, it is instead `GITHUB_API_URL`.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -2146,7 +2146,7 @@ pub const PackageManager = struct {
             this.allocator,
             "{s}/repos/{s}/{s}{s}tarball/{s}",
             .{
-                github_api_url,
+                strings.withoutTrailingSlash(github_api_url),
                 owner,
                 repo,
                 // repo might be empty if dep is https://github.com/... style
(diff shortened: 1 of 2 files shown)


src/install/install.zig
```diff
@@ -2144,9 +2144,9 @@ pub const PackageManager = struct {
 
         return std.fmt.allocPrint(
             this.allocator,
-            "https://{s}/repos/{s}/{s}{s}tarball/{s}",
+            "{s}/repos/{s}/{s}{s}tarball/{s}",
             .{
-                github_api_domain,
+                github_api_url,
```
Jarred-Sumner: incase they add a trailing slash to the end of the URL:

```suggestion
                strings.withoutTrailingSlash(github_api_url),
```

Jarred-Sumner: also technically needs to handle if they add a query param but hopefully they just don't do that

PR #6248, opened 2023-10-03
Support `bun run --if-present`

### LINKED ISSUE #5670: Support `--if-present` argument with `bun run`

When building standardized CI workflows, we set some standard commands that are typical to find in projects.
If the script doesn't exists, bun fails, which breaks those workflows if a project doesn't have that script/command.

We typically use `pnpm run --if-present <command>` on workflows today.

This will only run the command if the script is available in the package.

- We can probably build some script that skips steps in our workflow based on package.json.
- Add all scripts defined in a workflow, e.g. `exit 0`

Electroid:
### What does this PR do?

```sh
bun run doesnotexist # => exits with 1
bun run --if-present doesnotexist # => exits with 0
```

Closes #5670

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -1160,6 +1160,11 @@ pub const RunCommand = struct {
             }
         }
 
+        // If the script is not present and --if-present is passed, exit with 0
+        if (ctx.runtime_options.if_present) {
+            Global.exit(0);
+        }
+
         if (comptime log_errors) {
             Output.prettyError("<r><red>error<r><d>:<r> missing script \"<b>{s}<r>\"\n", .{script_name_to_search});
             Global.exit(1);
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -1160,9 +1160,8 @@ pub const RunCommand = struct {
             }
         }
 
-        // If the script is not present and --if-present is passed, exit with 0
         if (ctx.runtime_options.if_present) {
-            Global.exit(0);
+            return true;
         }
 
         if (comptime log_errors) {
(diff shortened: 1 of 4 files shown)


src/cli/run_command.zig
```diff
@@ -1160,6 +1160,11 @@ pub const RunCommand = struct {
             }
         }
 
+        // If the script is not present and --if-present is passed, exit with 0
+        if (ctx.runtime_options.if_present) {
```
Jarred-Sumner, line 1163: This should only happen when `comptime log_errors` to account for both `bun run --if-present` and `bun --if-present`

test/cli/run/if-present.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    env: bunEnv,
+    stdout: "pipe",
+    stderr: "pipe",
+  });
+  expect(stdout.toString()).toBeEmpty();
+  expect(stderr.toString()).toBeEmpty();
+  expect(exitCode).toBe(0);
+});
+
+it("should not error using `bun --if-present`", () => {
```
Jarred-Sumner:
```suggestion
it("should not error using `bun run --if-present`", () => {
```

test/cli/run/if-present.test.ts
```diff
@@ -0,0 +1,26 @@
+import { it, expect } from "bun:test";
+import { bunEnv, bunExe } from "harness";
+
+it("should not error using `bun --if-present`", () => {
```
Jarred-Sumner: this needs the positive test too:
- `bun --if-present works`
- `bun run --if-present works`
- `bun --if-present ./works.ts`
- `bun run --if-present ./works.ts`

PR #6255, opened 2023-10-03
Add license information to package.json

mathiasrw:
### What does this PR do?

Add MIT as license information in the package.json of `bun-types`

No code changed

### Assumtions

`bun-types` is distributed under MIT just like the rest of the Bun project as indicated here: https://bun.sh/docs/project/licensing

### Background

I use bun for bundling and testing, so I depend on bun-types. But without the license information in the package.json I get this FOSSA warning

image


Electroid:
Thanks!

PR #6258, opened 2023-10-03
Store workspace package versions

dylan-conway:
### What does this PR do?

- [ ] Code changes


test/cli/install/bun-install.test.ts
```diff
@@ -5550,6 +5550,85 @@ it("should handle installing packages from inside a workspace with `*`", async (
   await access(join(package_dir, "bun.lockb"));
 });
 
+it("should handle installing packages from inside a workspace with without prefix", async () => {
```
staabm:
```suggestion
it("should handle installing packages from inside a workspace without prefix", async () => {
```

dylan-conway: thanks

Issue #6252, opened 2023-10-03
The offset of Blob.slice.slice is incorrect


chawson:
### What version of Bun is running?

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

### What platform is your computer?

Linux 6.5.0-5-generic x86_64 x86_64

### What steps can reproduce the bug?

```ts
//@file index.js

import {readFileSync} from 'fs'

const filePath = '/video.mp4'
const blob = new Blob([readFileSync(filePath)])

// The following line of code does not meet expectations
const b = await blob.slice(744595, 754372).slice(0, 8).arrayBuffer()

console.log(Array.from(new Uint8Array(b)).join(','))
```

Executing `node index.js` outputs `0,0,0,108,109,118,104,100` as expected
Executing `bun index.js` output `0,0,0,32,102,116,121,112` is not as expected

### What is the expected behavior?

`0,0,0,108,109,118,104,100`

### What do you see instead?

`0,0,0,32,102,116,121,112`

The above result output is actually the data at offset 0 bytes, not the data at offset 744595 bytes.

### Additional information

Modify the above code and the following code will have the same problem when obtaining Blob.
```ts
const blob = Bun.file(filePath)
```
[video.mp4 download address](https://github.com/lokenetwork/FFmpeg-Principle/raw/main/videos/walking-dead.mp4)

Hanaasagi:
https://github.com/oven-sh/bun/blob/47651f321ae5eb82686da5d759e9b1b5b12340ad/src/bun.js/webcore/blob.zig#L3010-L3012

The `offset` of prev slice is not been added here 🧐 maybe `+=`.

PR #6259, opened 2023-10-03
fix(blob): Add the current offset to the new blob slice

Hanaasagi:
### What does this PR do?

Close: #6252

```JavaScript
const blob = new Blob(["0123456789"]);
console.log(await blob.slice(2, 8).slice(0, 3).text());
// Output:
// node: 234
// bun: 012
```

`offset` is overwritten by `relativeStart`, causing `blob.slice().slice()` to always use the parameters from the last `slice` call. This Add the current `offset` to the new blob slice.

https://github.com/oven-sh/bun/blob/47651f321ae5eb82686da5d759e9b1b5b12340ad/src/bun.js/webcore/blob.zig#L3010-L3012

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -3003,12 +3003,13 @@ pub const Blob = struct {
             }
         }
 
+        const offset = this.offset +| @as(SizeType, @intCast(relativeStart));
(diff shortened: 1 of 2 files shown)


Electroid:
Tests pass, thanks!

PR #6282, opened 2023-10-04
Update Node.js compat docs

colinhacks:
### What does this PR do?

- [ ] Code changes

### DIFF
--- a/docs/runtime/nodejs-apis.md
+++ b/docs/runtime/nodejs-apis.md
@@ -6,7 +6,7 @@ This page is updated regularly to reflect compatibility status of the latest ver
 
 ### [`node:assert`](https://nodejs.org/api/assert.html)
 
-🟢 Fully implemented.
+🟡 Missing `doesNotMatch`
 
 ### [`node:async_hooks`](https://nodejs.org/api/async_hooks.html)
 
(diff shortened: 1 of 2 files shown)


docs/runtime/nodejs-apis.md
```diff
@@ -42,19 +44,19 @@ This page is updated regularly to reflect compatibility status of the latest ver
...
 
 ### [`node:domain`](https://nodejs.org/api/domain.html)
 
-🟢 Fully implemented.
+🟡 Missing `Domain`, `active`
 
 ### [`node:events`](https://nodejs.org/api/events.html)
 
-🟡 Missing `require('node:events').on`.
+🟡 Missing `on` `addAbortListener` `getMaxListeners`
```
paperclover, line 55: til addAbortListener. also surprising the top level getMaxListeners was missed.

paperclover:
node:module is missing a ton of stuff too. maybe it should also be clarified that monkey-patching that module may not work.

for events, `addAbortListener` was added in v20. i think we should keep a better eye on new Node.js APIs, and also put the version of node that this page is for.

docs/runtime/nodejs-apis.md
```diff
@@ -74,11 +76,11 @@ This page is updated regularly to reflect compatibility status of the latest ver
 
 ### [`node:module`](https://nodejs.org/api/module.html)
 
-🟢 Fully implemented.
+🟢 Missing `runMain` `syncBuiltinESMExports`, `Module#load()`. Attempts to override or patch the module cache will fail.
```
birkskyum, line 79: This circle should be Yellow

PR #6290, opened 2023-10-04
fix: use latest version of Bun for setup

### LINKED ISSUE #6287: bun-types tag latest mixed with canary

1.0.4

Darwin 23.0.0 arm64 arm

image

If we set the version to the latest, then we did not install what we expected, we installed the canary instead.

version latest should get the latest, not the canary version.

### What do you see instead?

I got the canary version instead.

### Additional information

I suspect this commit:

- https://github.com/oven-sh/bun/commit/237a5ada4750ff255781d91e004e5e9570f031a2

We already have:

https://github.com/oven-sh/bun/blob/df9461ecc58e2882afa34cdf59d9887d9242b5f9/.github/workflows/bun-release-types-canary.yml#L1-L54C22

Why do we need bun-types published in 2 workflows?

image

image

ImBIOS:
### What does this PR do?

Closes #6287

- [ ] Code changes


Electroid:
This was fixed.

ImBIOS:
image

I don't think so

Issue #6297, opened 2023-10-04
Update `preload` entry documentation


Pandapip1:
### What is the problem this feature would solve?

~~Currently, to use plugins, you have to have a custom build script instead of being able to use `bun build` or `bun run`.~~

EDIT: Never mind. This already exists. The documentation is just bad. Updating the issue.

It's not obvious how to pass a plugin list to `bun build` and `bun run`, even though it is documented: https://bun.sh/docs/runtime/bunfig#preload

### What is the feature you are proposing to solve the problem?

~~I would like to be able to set a plugin list in `bunfig.toml`. I would also like a CLI option to override those entries.~~

The documentation should be updated to specify that this is a list of plugins to be loaded, not just 'helful for loading plugins'.

PR #6298, opened 2023-10-04
Update `preload` documentation to make applicability to plugins more obvious

Pandapip1:
### What does this PR do?

This updates the bunfig documentation to make it obvious that this is where plugins should be passed. It might seem obvious with the current documentation. I promise you, it's not.

- [X] Documentation or TypeScript types
- [ ] Code changes

Fixes https://github.com/oven-sh/bun/issues/6297


Jarred-Sumner:
Thank you

Issue #6284, opened 2023-10-04
process.exitCode setter not aligned with Node.js


panva:
### What version of Bun is running?

1.0.4+745b6b94ee56cad24d475799690cc9a89957d15b

### What steps can reproduce the bug?

```
process.exitCode = 255
```

### What is the expected behavior?

```sh
➜ node -e 'process.exitCode = 255; process.exit()'
➜ echo $?                                         
255
```

### What do you see instead?

```
bun repl
Welcome to Bun v1.0.4
Type ".help" for more information.
[!] Please note that the REPL implementation is still experimental!
    Don't consider it to be representative of the stability or behavior of Bun overall.
> process.exitCode = 255
1 | 
2 | process.exitCode = 255;
   ^
RangeError: exitCode must be between 0 and 127
```

### Additional information

I've encountered this running a test suite using `mocha` which, upon test suite failure, sets the exit code to 255.

PR #6303, opened 2023-10-04
fix: support uint8 exit code range

liz3:
The exit code support is between 0-255 and not only in the signed positive range (0-127).
Node.js does not seam to throw on a bigger integer and just wraps around, but throwing a error is a good approach and makes the behaviour more defined.

This allows the range to be 0-255.

Fixes: https://github.com/oven-sh/bun/issues/6284

### What does this PR do?

- [x] Code changes


Jarred-Sumner (approved):
Thank you

paperclover:
i remember discovering the wrap around behavior when doing process.exit, i hope no one actually relies on it. i think it's fine to throw here, but technically we should be wrapping.

paperclover (approved):

liz3:
can this be merged?

PR #6319, opened 2023-10-05
Installing unzip.

AbdullahOB:
### What does this PR do?

This change is for Developers documentation, so it informs the developers that they need to install unzip before installing bun.

- [ ] Code changes

### DIFF
--- a/docs/project/development.md
+++ b/docs/project/development.md
@@ -7,6 +7,7 @@ Before starting, you will need to already have a release build of Bun installed,
 {% codetabs %}
 
 ```bash#Native
+$ sudo apt-get install zip unzip -y
 $ curl -fsSL https://bun.sh/install | bash # for macOS, Linux, and WSL
 ```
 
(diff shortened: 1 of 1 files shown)


v3nntrnx:
It says native but apt is only for debian

AbdullahOB:
I've tried it on WSL and it didn't run the install before I installed the unzip package for wsl.

v3nntrnx:
> I've tried it on WSL and it didn't run the install before I installed the unzip package for wsl.

What I am trying to say is, `apt`/`apt-get` is not available on MacOS and other Linux distributions than Debian. It says "native" one line prior

AbdullahOB:
> > I've tried it on WSL and it didn't run the install before I installed the unzip package for wsl.
> 
> What I am trying to say is, `apt`/`apt-get` is not available on MacOS and other Linux distributions than Debian. It says "native" one line prior

Yes I understand, thank you for clarifiying I will perform edits on the commit.

AbdullahOB:
I've added for both CentOS/RHEL and Debian/Ubuntu, i think this would be more clear.

Electroid:
We want to make these install commands concise, will keep as-is for now.

PR #6324, opened 2023-10-05
add optional symbol minification to bundler (resolves #2797)

### LINKED ISSUE #2797: Add optional symbol minification to `bun build --minify`

The [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) built-in takes an optional string description as a parameter. This string is purely for debugging purposes and does not effect the identity or uniqueness of the symbol itself.

```js
// example.js
const mySymbolA = Symbol('k_thisIsMySymbolA');
const mySymbolB = Symbol('k_thisIsMySymbolB');
const mySymbolC = Symbol('k_thisIsMySymbolC');
```

```bash
$ bun build --minify --outdir=out example.js
```

```js
// out/example.js 
var c=Symbol("k_thisIsMySymbolA"),y=Symbol("k_thisIsMySymbolB"),I=Symbol("k_thisIsMySymbolC");
```

During minification, on large codebases that make use of symbols, dropping the string parameter can reduce the size of the final bundle.

```js
const mySymbolA = Symbol('k_thisIsMySymbolA');
const mySymbolB = Symbol('k_thisIsMySymbolB');
const mySymbolC = Symbol('k_thisIsMySymbolC');
```

```bash
$ bun build --minify --minify-symbols --outdir=out example.js
```

```js
// out/example.js
var c=Symbol(),y=Symbol(),I=Symbol();
```

Using Vue.js as a example of an existing codebase, the optimization saves 1kb~ (763 bytes). The result will vary depending on the usage of Symbols, for example the Three.js project does not include a single Symbol in the final build, so does not benefit at all.

While the impact of this is relatively small, the implementation should be fairly trivial and can be optional as with the other minification options.

[I implemented the same optimization into Terser](https://github.com/terser/terser/pull/582) under the `unsafe_symbols` flag. We did so as disabled-by-default because you *can* use the description of a symbol via `.toString()`, which can affect runtime if used. I see no reason not to have this as an optional flag though.

_No response_

Kruithne:
### What does this PR do?

Extends the bundler minification to include a `symbols` option, which when enabled removes arguments provided to `Symbol()`.  

```ts
// Before
const my_symbol = Symbol('foo_bar');

// After
const my_symbol = Symbol();
```

Both `Symbol.for(...)` and `Symbol.iterator` are left untouched. In 99% of cases, the description provided to `Symbol()` is only used for debugging purposes and therefore makes sense to remove these for production. In the rare event that someone is relying on symbol inspection, this can simply be disabled.

[I implemented the same optimization into Terser](https://github.com/terser/terser/pull/582) under the `unsafe_symbols` flag.

> [!NOTE]
> As a note, this is my first time writing anything in Zig and my first time touching the Bun codebase. Please let me know if there's any mistakes or anything I should have done differently.

#2797

In addition to adding an automated test `minify/SymbolMinification` to `bundler_minify.test.ts`, I've also compiled Bun locally and tested the functionality of the new feature.


src/js/builtins/BundlerPlugin.ts
```diff
@@ -174,6 +174,7 @@ export function runSetupFunction(this: BundlerPlugin, setup: Setup, config: Buil
       minifyIdentifiers: config.minify === true || (config.minify as MinifyObj)?.identifiers,
       minifyWhitespace: config.minify === true || (config.minify as MinifyObj)?.whitespace,
       minifySyntax: config.minify === true || (config.minify as MinifyObj)?.syntax,
+      minifySymbols: config.minify === true || (config.minify as MinifyObj)?.symbols,
```
paperclover: minifySymbols isn't an option on esbuild's config, we dont need this

docs/bundler/vs-esbuild.md
```diff
@@ -333,6 +333,11 @@ In Bun's CLI, simple boolean flags like `--minify` do not accept an argument. Ot
 
 ---
 
+- `--minify-symbols`
```
paperclover: esbuild does not have --minify-symbols

paperclover:
edge cases to try
```ts
capture(Symbol({})) // minify
capture(Symbol(`${1}`)) // minify
capture(Symbol(1, 2)) // minify

capture(Symbol(a)) // don't minify
capture(Symbol(b()) // don't minify
capture(Symbol(`${b()}`)) // don't minify

{
  const Symbol = 1;
  capture(Symbol("hello")) // don't minify
}
```

docs/bundler/vs-esbuild.md
Jarred-Sumner: yeah, let's not have this as a flag. if `--minify-identifiers` is enabled, so should this be

Kruithne: > yeah, let's not have this as a flag. if `--minify-identifiers` is enabled, so should this be

Do you want it entirely merged with identifiers so '{ minify: { identifiers: true } }' as well as '--minify-identifiers'?

I separated it out originally since it might be considered breaking for people who inspect symbols for any reason. Not a concern?

paperclover (changes requested):
Needs to ensure that side effects don't get optimized away before this can be merged

Kruithne:
Since moving the functionality behind the identifier flag, it appears to have stopped working. With limited understanding of Zig/Bun, I'm unsure why.

PR #6346, opened 2023-10-06
[types] allow onLoad plugin callbacks to return undefined

silversquirl:
Returning `undefined` simply falls through to the next plugin, or to the default loader.
This behaviour is defined by esbuild, and supported by Bun, but the types don't reflect it properly. This PR fixes that.


Jarred-Sumner:
Thank you

PR #6350, opened 2023-10-07
Fix memory leak in fetch()

Jarred-Sumner:
### What does this PR do?

We were not calling inflateEnd, which caused fetch() requests with gzip/deflate to leak memory

hmasd

Memory usage is more stable now


joulev:
This is amazing. As someone who is using Bun to run a long-running process in production, this is really needed. Props to the team 💯

ralyodio:
good.

PR #6352, opened 2023-10-07
feat(install): automatically migrate package-lock.json to bun.lockb

paperclover:
### What does this PR do?

If you do not have a `bun.lockb`, but a `package-lock.json` exists, this PR reads that data and generates a bun lockfile in memory, and uses that for the install.

I think the migration can be made faster, but people will ideally only run this once per project, so it is fine.

image

If a `yarn.lock` or `pnpm-lock.yaml` exists, a warning is printed

image

All of the above only happens if no `bun.lockb` is present. so this has no perf impact on existing projects.

You can also run `bun pm migrate` to run just the migrator code path, useful for testing.

There are some situations where this migration is not 100% accurate to what npm installs, which is due to bugs and limitations/optimizations of Bun's package hoisting logic.


src/install/install.zig
```diff
@@ -7647,8 +7658,10 @@ pub const PackageManager = struct {
             )
         else
             .{ .not_found = {} };
+
         var root = Lockfile.Package{};
-        var needs_new_lockfile = load_lockfile_result != .ok or (load_lockfile_result.ok.buffers.dependencies.items.len == 0 and manager.package_json_updates.len > 0);
+        var needs_new_lockfile = load_lockfile_result != .ok or load_lockfile_result.ok.buffers.dependencies.items.len == 0 and manager.package_json_updates.len > 0;
```
paperclover: is this right

src/install/lockfile.zig
```diff
@@ -1875,18 +1904,28 @@ pub const Package = extern struct {
...
+    /// so we can access it faster.
+    ///
+    /// Each index in this array corresponds to the same index in dependencies.
+    /// Each value in this array corresponds to the resolved package ID for that dependency.
+    ///
+    /// So this is how you say "what package ID for lodash does this package actually resolve to?"
+    ///
+    /// By default, the underlying buffer is filled with "invalid_id" to indicate this package ID
+    /// was not resolved
+    resolutions: PackageIDSlice = .{},
```
paperclover, line 1935: lets not move the order of these fields

src/install/migration.zig
```diff
@@ -0,0 +1,964 @@
...
+            Output.printElapsed(@as(f64, @floatFromInt(timer.read())) / std.time.ns_per_ms);
+            Output.prettyError(" ", .{});
+            Output.prettyErrorln("<d>migrated lockfile from <r><green>package-lock.json<r>", .{});
+            Output.flush();
+        }
+
+        return lockfile;
+    }
+
+    unsupported: {
```
paperclover: ✂️

src/install/lockfile.zig
paperclover: damn i didnt realize i did that

src/install/install.zig
paperclover: probably not. lol

paperclover:
can u add a test case for `bun add`

paperclover:
cool

PR #6358, opened 2023-10-07
Remove `bunx --bun` from Vite guide

jonkoops:
The Vite guide references running Vite commands using `bunx --bun vite`. This should no longer be required since Bun 1.0 (according to documentation on the same page), so we can fall back to `bun run` instead.

### What does this PR do?

This replaces the instructions for running Vite with `bunx  --bun` with 'regular' `bun run` commands.

- [ ] Code changes

### DIFF
--- a/docs/guides/ecosystem/vite.md
+++ b/docs/guides/ecosystem/vite.md
@@ -28,42 +28,10 @@ bun install
 
 ---
 
-Start the development server with the `vite` CLI using `bunx`.
-
-The `--bun` flag tells Bun to run Vite's CLI using `bun` instead of `node`; by default Bun respects Vite's `#!/usr/bin/env node` [shebang line](<https://en.wikipedia.org/wiki/Shebang_(Unix)>). After Bun 1.0 this flag will no longer be necessary.
-
-```bash
-bunx --bun vite
-```
-
----
-
-To simplify this command, update the `"dev"` script in `package.json` to the following.
-
-```json-diff#package.json
-  "scripts": {
--   "dev": "vite",
-+   "dev": "bunx --bun vite",
-    "build": "vite build",
-    "serve": "vite preview"
-  },
-  // ...
-```
-
----
-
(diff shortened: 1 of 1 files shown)


birkskyum:
--bun is still needed to force usage of Bun in many cases - The default behaviour of using Node.js wasn't changed at the 1.0 release despite it being the initial plan. IIRC there has been a PR merged into vite that helps it recognize bun, but this documentation still serves a purpose, until the switch is flipped to default the --bun behaviour under the hood.

jonkoops:
Thanks for the info @birkskyum. Would you like me to keep this PR open in the mean time? I can also create a new PR that more accurately reflects the current status.

> IIRC there has been a PR merged into vite that helps it recognize bun

I am not sure if correct, but I assume you are referring to https://github.com/vitejs/vite/pull/14379. If that is the case, that has now landed in [a recent release](https://github.com/vitejs/vite/blob/v4.4.11/packages/vite/CHANGELOG.md#4410-2023-10-03).

> until the switch is flipped to default the --bun behaviour under the hood.

So this will have to be done in Bun, and not in Vite, correct? Is there a place where this progress can be tracked?

birkskyum:
I don't think it can be tracked anywhere, but yes, it's a change that will be in Bun when things generally don't break.

birkskyum:
Regarding if this should be kept changed: cc @colinhacks

yukulele:
If that request isn't acceptable, this one should be : https://github.com/oven-sh/bun/pull/6399

jonkoops:
> If that request isn't acceptable, this one should be : #6399

Yeah, I think it's worth removing the 'after 1.0', since it no longer is correct.

colinhacks:
As mentioned elsewhere we ultimately decided to respect shebangs by default with `bun run` so `--bun` *is* still necessary post-1.0. Thanks everyone, sorry for any confusion there.

Issue #6336, opened 2023-10-06
Copying a file displays an incorrect error, using the source path instead of the destination path if the destination directory doesn't exist


Vanilagy:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

With `baz.txt` existing, run the following:
```ts
await Bun.write('./foo/bar.txt', Bun.file('./baz.txt'));
```
This operation errors because the folder `foo` doesn't exist.

### What is the expected behavior?

It should tell me that the output directory does not exist.

### What do you see instead?

```
ENOENT: No such file or directory
   path: "./baz.txt"
 syscall: "open"
   errno: -2
```

Hanaasagi:
Additionally, when the second argument is a string, there is no `path` in the error message. It seems that `path` field might be using the value from the second.

```JavaScript
await Bun.write("./foo/bar.txt", "");

// Output
// 1 | await Bun.write("./foo/bar.txt", "");
//          ^
// ENOENT: No such file or directory
//  syscall: "open"
//    errno: -2


```

Vanilagy:
slay

PR #6370, opened 2023-10-08
fix(error): correct the `path` field in syscall error message.

Hanaasagi:
### What does this PR do?

This PR fixes the `path` field in error messages within `Bun.write`. Close: #6336

#### 1
```JavaScript
// 1. `./baz.txt` exists.
// 2. `foo/` doesn't exist.
await Bun.write('./foo/bar.txt', Bun.file('./baz.txt'));

// Output:
// ENOENT: No such file or directory
//    path: "./baz.txt"
//  syscall: "open"
//    errno: -2

```

https://github.com/oven-sh/bun/blob/35109160ca5d439116bedeb3302ec3745e2895d5/src/bun.js/webcore/blob.zig#L2292

https://github.com/oven-sh/bun/blob/35109160ca5d439116bedeb3302ec3745e2895d5/src/bun.js/webcore/blob.zig#L2197-L2200

The `path` field in the error message will use the value from the second parameter. This happens because in the `doOpenFile` function, we do not assign the `path` field to the error, the error context is incomplete. And when rejecting, it defaults to the source's path.

----

#### 2
```JavaScript
await Bun.write("./foo/bar.txt", "");

// Output
// 1 | await Bun.write("./foo/bar.txt", "");
//          ^
// ENOENT: No such file or directory
//  syscall: "open"
//    errno: -2

```

In the `writeStringToFileFast` function, the `path` field is not assigned.

https://github.com/oven-sh/bun/blob/35109160ca5d439116bedeb3302ec3745e2895d5/src/bun.js/webcore/blob.zig#L1043C1-L1043C1

- [x] Code changes

I wrote automated tests.

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -1040,7 +1040,10 @@ pub const Blob = struct {
                     break :brk result;
                 },
                 .err => |err| {
-                    return JSC.JSPromise.rejectedPromiseValue(globalThis, err.toJSC(globalThis));
+                    return JSC.JSPromise.rejectedPromiseValue(
+                        globalThis,
+                        err.withPath(pathlike.path.slice()).toJSC(globalThis),
+                    );
                 },
             }
             unreachable;
@@ -1081,7 +1084,10 @@ pub const Blob = struct {
                             return .zero;
                         }
 
-                        return JSC.JSPromise.rejectedPromiseValue(globalThis, err.toJSC(globalThis));
+                        return JSC.JSPromise.rejectedPromiseValue(
+                            globalThis,
+                            err.withPath(pathlike.path.slice()).toJSC(globalThis),
+                        );
                     },
                 }
             }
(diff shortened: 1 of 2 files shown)


src/bun.js/webcore/blob.zig
```diff
@@ -1110,7 +1116,10 @@ pub const Blob = struct {
                     break :brk result;
                 },
                 .err => |err| {
-                    return JSC.JSPromise.rejectedPromiseValue(globalThis, err.toJSC(globalThis));
+                    return JSC.JSPromise.rejectedPromiseValue(
```
Jarred-Sumner, line 1121: These need to check if it's a file descriptor first. This will crash if pathlike's active union member is `fd`

Hanaasagi: Fixed.

Jarred-Sumner:
Thank you. This looks good.

PR #6376, opened 2023-10-09
Fix lifecycle scripts not running on reinstallation

### LINKED ISSUE #5472: `bun install` ignores trustedDependencies when bun.lockb is present

1.0.1+92e95c86dd100f167fb4cf8da1db202b5211d2c1

Darwin 23.0.0 arm64 arm

1. Create the following `package.json`:
    ```json
    {
      "dependencies": {
        "@prisma/client": "^5.3.1"
      },
      "trustedDependencies": ["@prisma/client"]
    }
    ```
    note `@prisma/client` is just an arbitrary package with a postinstall script; you can replace with any other package that runs a `postinstall`.
2. `bun install` 

    Observe that `bun.lockb` and `node_modules` are created.
    Observe also that the postinstall script runs (see output like `prisma:warn The postinstall script ...`; the warning is expected in the absence of a `schema.prisma file`)

4. `rm -rf node_modules && bun install`
    
    Observe that the postinstall script for `@prisma/client` does *not* run—there is no visible output in the terminal on a “second install” once `bun.lockb` is present.

5. `rm -rf node_modules bun.lockb && bun install`
    
    Observe that when neither `node_modules` *nor* `bun.lockb` is present, the postinstall script once again runs for `trustedDependencies`

Install scripts for `trustedDependencies` should run for subsequent installs even once `bun.lockb` is present.

### What do you see instead?

Install scripts for `trustedDependencies` only run on the first install (before `bun.lockb` is created).
When a new user clones a repository and runs `bun install`, install scripts do not run, since `bun.lockb` is committed but `node_modules` is not.

### Additional information

Some discussion of this bug in the comments following https://github.com/oven-sh/bun/issues/4549#issuecomment-1711602815

Opening a separate issue as #4549 was originally about a slightly different issue (ie the need to add packages like `sharp` to `trustedDependencies`)

Arden144:
### What does this PR do?

This PR adds serialization and deserialization of the trustedDependencies field from the package.json field in the lockfile. I added an 8 byte flag to the lockfile to indicate if the field is present or not.

Please let me know if you'd like this change implemented differently and I'd be happy to make changes.

This PR closes #5472 

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
This might not strictly be necessary since we do read this information from the package.json

but I think it's also okay to store it

Looks good, lets let the tests run. There will be one failure in bun-install due to an existing bug.

dylan-conway:
Thank you

Issue #6012, opened 2023-09-24
FormData video file does not have type on Windows 10 WSL2


wfl-junior:
### What version of Bun is running?

1.0.3

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

I created a small server for uploading files, copying from the example at https://bun.sh/guides/http/file-uploads

Here is the code:

```ts
const server = Bun.serve({
  port: 4000,
  async fetch(request) {
    const url = new URL(request.url);

    if (url.pathname === "/upload") {
      const formdata = await request.formData();
      const file = formdata.get("files") as Blob;
      if (!file) throw new Error("Must upload a file.");

      console.log({ type: file.type });

      // await Bun.write('profilePicture.png', file);
      return new Response("Success");
    }

    return new Response("Not Found", { status: 404 });
  },
});
```

And make a multipart form request to http://localhost:4000 sending `files` key with a video type file.

### What is the expected behavior?

To see the mimetype of the video file.

### What do you see instead?

Console logs:

```ts
{
  type: ""
}
{
  type: "image/jpeg"
}
{
  type: ""
}
{
  type: ""
}
{
  type: ""
}
{
  type: "image/jpeg"
}
{
  type: "image/png"
}
{
  type: "image/png"
}
{
  type: ""
}
{
  type: ""
}
{
  type: ""
}
{
  type: ""
}
```

The empty ones are videos, it works fine with images.

### Additional information

I am running Bun on WSL2 Windows 10 (Microsoft Windows NT 10.0.19045.0 x64).
Requesting via Insomnia with multipart form body.
I have tried requesting via a simple html form with multipart/form-data enctype and same issue.

jimmywarting:
How dose the upload part look like? the same as the html page with file inputs?
try have a look at input.files to see what type they have.

it's not the `request.formData()` fault that it's not getting any type, it's probably due to not sending any...

Electroid:
Could you share the code or script that you used to upload the form?

wfl-junior:
> How dose the upload part look like? the same as the html page with file inputs? try have a look at input.files to see what type they have.
> 
> it's not the `request.formData()` fault that it's not getting any type, it's probably due to not sending any...

I have tried both on a normal html form, I checked it sends the video as `video/mp4` and uploading from Insomnia.

wfl-junior:
> Could you share the code or script that you used to upload the form?

```tsx
[...]

PR #6380, opened 2023-10-09
fix: form data content type

yadav-saurabh:
fix for #6012 

```
std.fs.path.extension(filename_str); 
// return extension name with a dot(.) the ex: somefile.pdf -> .pdf
```

```
bun.HTTP.MimeType.byExtensionNoDefault()
// expects extension name without any dot(.)  
```

verified the code runs on my local build

image


Jarred-Sumner (approved):
LGTM

Jarred-Sumner:
Thank you

PR #6390, opened 2023-10-10
fix(AbortSignal/fetch) fix AbortSignal.timeout, fetch lock behavior and fetch errors

### LINKED ISSUE #6366: Bun.serve hang with CPU 100% load when AbortSignal.timeout is used

1.0.4

Darwin 23.0.0 arm64 arm

Start a simple script, and send request few times
```js
Bun.serve({
    port: 3000,
    async fetch() {
        const signal = AbortSignal.timeout(1)
        signal.throwIfAborted()
        return await fetch('https://google.com', { signal })
    }
})
```

`AbortSignal.timeout` should work as expected and do not hang entire server.

### What do you see instead?

After 3-4 requests CPU load of bun become 100% all the time, until you kill the process.

### Additional information

_No response_

cirospaciari:
### What does this PR do?

This fix fetch behavior on error cases, fix HTTP destroying too early causing missing parts/events not to be emitted, fix AbortSignal.timeout, fix proper closing/ending off fetch when aborted, and also fix the state of readable streams to be used not only empty.

Fix #6366

- [x] Code changes

I wrote automated tests


packages/bun-usockets/src/eventing/epoll_kqueue.c
```diff
@@ -438,15 +442,19 @@ void us_timer_set(struct us_timer_t *t, void (*cb)(struct us_timer_t *t), int ms
...
     struct us_internal_callback_t *internal_cb = (struct us_internal_callback_t *) timer;
 
     struct kevent64_s event;
     EV_SET64(&event, (uint64_t) (void*) internal_cb, EVFILT_TIMER, EV_DELETE, 0, 0, (uint64_t)internal_cb, 0, 0);
     kevent64(internal_cb->loop->fd, &event, 1, NULL, 0, 0, NULL);
 
     /* (regular) sockets are the only polls which are not freed immediately */
-    us_poll_free((struct us_poll_t *) timer, internal_cb->loop);
+    if(fallthrough){
+        us_free(timer);
```
Jarred-Sumner, line 454: isn't there a scenario where:
1) Timer is scheduled to run in the next tick
2) Tick happens, but before timer happens, the timer is cancelled
3) Timer is freed
4) Callback for timer runs after the timer is freed, which becomes a use-after-free

cirospaciari: today we call `us_poll_free` with just do:
```c 
void us_poll_free(struct us_poll_t *p, struct us_loop_t *loop) {
    loop->num_polls--;
    us_free(p);
}
```
so is safe to assume that we are not breaking any behavior only avoiding `loop->num_polls--;` because on `fallthrough: true` we dont increment num_pools, But will check this scenario.

PR #6395, opened 2023-10-10
CLI flags

colinhacks:
### What does this PR do?

- Add hand-written helptext for each top-level command
- Warn on unrecognized flags instead of aborting early
- Organize CLI flags better
- Do not warn on unrecognized flags when argv[0] is node (`--bun`)


src/cli.zig
```diff
@@ -526,10 +501,42 @@ pub const Arguments = struct {
 
         ctx.passthrough = args.remaining();
 
-        opts.no_summary = args.flag("--no-summary");
+        // opts.no_summary = args.flag("--no-summary");
```
colinhacks: Used in `bun bun` only

src/cli.zig
```diff
@@ -526,10 +501,42 @@ pub const Arguments = struct {
...
+                if (options.GlobalCache.Map.get(enum_value)) |result| {
+                    ctx.debug.global_cache = result;
+                    // -i, --install
+                } else if (enum_value.len == 0) {
+                    ctx.debug.global_cache = options.GlobalCache.force;
+                } else {
+                    Output.prettyErrorln("Invalid value for --install: \"{s}\". Must be either \"auto\", \"fallback\", \"force\", or \"disable\"\n", .{enum_value});
+                    Global.exit(1);
+                }
+            }
```
colinhacks: These flags are now only parsed for commands that use the runtime

paperclover:
0. `bun` alone should not print "failed to parse command"

1. This command:
`bun --help                     Show all supported flags and commands`
does not print the flags

2. do you think `bun run` where it prints your existing scripts, should it have a hint for the --help flags, should colors be different here

3. many of these have leading whitespace on the first line, like `bun install`

4. these commands dont have accessible help menus, which would be a expectation if you mention "print helptext for a specific command". also now realizing maybe that should say "help text"
  - init

5. bun create still prints a header before the usage, can be removed for consistancy

6. `bun add` alone should equal what `bun add --help`

7. some of these do not have a trailing newline. some shells will print their prompt on the same line
image
see the arrow there, on bash
image


whitespace nitpicks:
image
image

src/cli.zig
```diff
@@ -511,13 +485,14 @@ pub const Arguments = struct {
...
 
-        if (args.option("--origin")) |origin| {
-            opts.origin = origin;
-        }
+        // if (args.option("--origin")) |origin| {
+        //     opts.origin = origin;
+        // }
+
+        // if (args.option("--port")) |port_str| {
+        //     opts.port = std.fmt.parseInt(u16, port_str, 10) catch return error.InvalidPort;
```
paperclover: do we not use these for anything?

src/cli/package_manager_command.zig
```diff
@@ -44,7 +44,7 @@ const ByName = struct {
 };
 
 pub const PackageManagerCommand = struct {
-    pub fn printHelp(_: std.mem.Allocator) void {}
+    // pub fn printHelp(_: std.mem.Allocator) void {}
```
paperclover: delete instead of comment

paperclover (changes requested):
code looks great but need to address the giant comment i left already

src/cli.zig
colinhacks: Not that I can tell. Seems like they're vestigial from `bun dev`. 

I'd also argue they *shouldn't* do anything...seems too application-level to be exposed as a CLI flag.

paperclover: imo if these are used for things that got ✂️'d, then we should completely remove it from the code, instead of being a dead comment.

colinhacks:
Addressed all of Dave's feedback

- [x] bun alone should not print "failed to parse command"
- [x] `bun --help` command does not print the flags
- [x] print detected package.json commands in `bun run --help`
- [x] fix leading whitespace on the first line
- [x] add trailing newline where forgotten
- [x] add `bun init --help`
- [x] remove old `bun create` header
- [x] `bun add` (no arg) should be equivalent to `bun add --help`

paperclover (approved):

src/cli.zig
```diff
@@ -1082,8 +1083,7 @@ pub const Command = struct {
 
             if (comptime Command.Tag.uses_global_options.get(command)) {
                 ctx.args = try Arguments.parse(allocator, &ctx, command);
-            }
-
+            } else {}
```
Jarred-Sumner:
```suggestion
  }
```

src/deps/zig-clap/clap.zig
```diff
@@ -208,27 +209,36 @@ pub const Diagnostic = struct {
...
         switch (err) {
-            error.DoesntTakeValue => try stream.print("The argument '{s}{s}' does not take a value\n", .{ a.prefix, a.name }),
-            error.MissingValue => try stream.print("The argument '{s}{s}' requires a value but none was supplied\n", .{ a.prefix, a.name }),
-            error.InvalidArgument => if (a.prefix.len > 0 and a.name.len > 0)
-                try stream.print("Invalid argument '{s}{s}'\n", .{ a.prefix, a.name })
-            else
-                try stream.print("Failed to parse argument due to unexpected single dash\n", .{}),
-            else => try stream.print("Error while parsing arguments: {s}\n", .{@errorName(err)}),
+            error.DoesntTakeValue => {
+                Output.pretty("<red>error<r><d>:<r> The argument '{s}' does not take a value.\n", .{name});
```
Jarred-Sumner: Should these be `stdout` or `stderr`? This is stdout. Output.prettyError is stderr.

src/deps/zig-clap/clap/streaming.zig
```diff
@@ -1,6 +1,7 @@
 const builtin = @import("builtin");
 const clap = @import("../clap.zig");
 const std = @import("std");
+const Output = @import("../../../output.zig");
```
Jarred-Sumner:
```suggestion
const Output = @import("root").bun.Output;
```

Issue #4627, opened 2023-09-08
[Linux ARM64] fs.mkdirSync siliently fails


Xiot:
### What version of Bun is running?

1.0.0

### What platform is your computer?

Linux 5.15.49-linuxkit-pr aarch64 unknown

### What steps can reproduce the bug?

This was using the `oven/bun` docker image

The setup that I currently have is a little complicated, but the gist of it is that it makes a bunch of folders in quick succession.

This works fine on my M2 laptop

This is mainly debugging code that I added, to try to figure out what is going on.
But the `mkdirp.before` and `mkdirp.after` lines will both show that the folder (`ret`) doesn't exist, even though I attempted to create it twice.  Neither instance throws an `Error` either
```javascript
console.log('mkdirp.before', ret, fs.existsSync(ret))
if (!fs.existsSync(ret)) fs.mkdirSync(ret, {recursive: true})
if (!fs.existsSync(ret)) fs.mkdirSync(ret, {recursive: true})
console.log('mkdirp.after', ret, fs.existsSync(ret))
```

It's a little weirder because I can re-run my script multiple times, and it will always 'fail' at the same place.  If I create the folder manually, my script will continue for a little bit, then start to fail (and continue to fail) at a different place.

### What is the expected behavior?

The folder is either successfully created, or an `Error` is thrown

### What do you see instead?

The call completes, however the directory doesn't actually exist.

### Additional information

This is the basics of the script.
It reads in coverage data, and creates a html report

```javascript
import libReport from 'istanbul-lib-report'
import reports from 'istanbul-reports'
import libCoverage from 'istanbul-lib-coverage'

const coverage = libCoverage.createCoverageMap()
map.merge(coverageJson)

const ctx = libReport.createContext({
  dir: reportPath,
  coverageMap: coverage,
})
reports.create('html', { subdir: 'html' }).execute(ctx)
```

Xiot:
I'm not sure what actually changed, but seems to all work now.  
I re-created the image (for the 20th time) and it now looks like it is working.

Sorry for the noise

dlorenc:
I just ran into this too on Docker (on an M2 mac)

Whobeu:
I found what may be a related issue. For use with Node.js, I wrap mkdirSync in a try/catch block since it will throw an error if the directory already exists. In Bun, no such error is thrown. The unmodified code still works as expected in Node.js with the catch block catching the failure.

Xiot:
[...]

Issue #5943, opened 2023-09-23
Puppeteer not working on Docker


JawadAhbab:
### What version of Bun is running?

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Chromium is pre-downloaded in the image in this location `/usr/bin/chromium`

```ts
const browser = await puppeteer.launch({
    headless: 'new',
    executablePath: `/usr/bin/chromium`,
    args: ['--no-sandbox', '--disable-setuid-sandbox', '--disable-dev-shm-usage'],
})
```

### What do you see instead?

Getting this error
```
$ bun run src/index.ts
71 |             return (0, util_js_1.emplace)(this.searchCache, from, search);
72 |         }
73 |         return search();
74 |     }
75 |     #readConfiguration(filepath) {
76 |         const contents = fs_1.default.readFileSync(filepath, 'utf8');
                             ^
EBADF: Bad file descriptor
 syscall: "fstat"
   errno: -9

      at #readConfiguration (/home/bun/app/node_modules/puppeteer/node_modules/cosmiconfig/dist/ExplorerSync.js:76:25)
      at search (/home/bun/app/node_modules/puppeteer/node_modules/cosmiconfig/dist/ExplorerSync.js:44:39)
      at emplace (/home/bun/app/node_modules/puppeteer/node_modules/cosmiconfig/dist/util.js:12:19)
      at getInternalOptions (/home/bun/app/node_modules/puppeteer/node_modules/cosmiconfig/dist/index.js:74:36)
      at cosmiconfigSync (/home/bun/app/node_modules/puppeteer/node_modules/cosmiconfig/dist/index.js:183:28)
      at getConfiguration (/home/bun/app/node_modules/puppeteer/lib/esm/puppeteer/getConfiguration.js:35:19)
      at /home/bun/app/node_modules/puppeteer/lib/esm/puppeteer/puppeteer.js:19:22
      at loadCJS2ESM (:1:20)
      at requireESM (:1:20)
```

sandersemmel:
Same issue here:

```
Error message:

69 |             return (0, util_js_1.emplace)(this.searchCache, from, search);
70 |         }
71 |         return search();
72 |     }
73 |     #readConfiguration(filepath) {
74 |         const contents = fs_1.default.readFileSync(filepath, 'utf8');
                             ^
EBADF: Bad file descriptor
[...]

Issue #5935, opened 2023-09-22
Error with HMR in NestJS app | Error: spawn EBADF


diavrank:
### What version of Bun is running?

1.0.3+67defd95afb8f19259b2bde9be10e1a76a8137b5

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. Create a Nestjs app
2. Run the app: `NODE_ENV=local nest start --watch --exec "bun run"`

Note: The environment variable shouldn't not affect to your app. 

3. Do a change in a file (e.g. add a `console.log('hello world') `)
4. Wait like 5 seconds
5. Verify that the app is stopped with an error.

### What is the expected behavior?

To restart the server correctly with the new changes

### What do you see instead?

```shell
[5:01:20 PM] File change detected. Starting incremental compilation...

DeprecationWarning: 'getMutableClone' has been deprecated since v4.0.0. Use an appropriate `factory.update...` method instead, use `setCommentRange` or `setSourceMapRange`, and avoid setting `parent`.
DeprecationWarning: 'createLiteral' has been deprecated since v4.0.0. Use `factory.createStringLiteral`, `factory.createStringLiteralFromNode`, `factory.createNumericLiteral`, `factory.createBigIntLiteral`, `factory.createTrue`, `factory.createFalse`, or the factory supplied by your transformation context instead.
node:internal/child_process:413
    throw errnoException(err, 'spawn');
    ^

Error: spawn EBADF
    at ChildProcess.spawn (node:internal/child_process:413:11)
    at spawn (node:child_process:757:9)
    at /Users/myuser/Projects/nestjs/api/node_modules/tree-kill/index.js:33:18
    at buildProcessTree (/Users/myuser/Projects/nestjs/api/node_modules/tree-kill/index.js:90:14)
    at module.exports (/Users/myuser/Projects/nestjs/api/node_modules/tree-kill/index.js:32:9)
    at /Users/myuser/Projects/nestjs/api/node_modules/@nestjs/cli/actions/start.action.js:72:17
    at Object.onWatchStatusChange (/Users/myuser/Projects/nestjs/api/node_modules/@nestjs/cli/lib/compiler/watch-compiler.js:64:17)
    at /Users/myuser/Projects/nestjs/api/node_modules/typescript/lib/typescript.js:126329:117
    at emitFilesAndReportErrors (/Users/myuser/Projects/nestjs/api/node_modules/typescript/lib/typescript.js:126148:13)
[...]

Issue #5306, opened 2023-09-13
Get cdktf-cli package working


ozyman42:
### What version of Bun is running?

1.0.2

### What platform is your computer?

Darwin 22.6.0 arm64 arm

#### From the host OS
docker run -it --rm --entrypoint bash oven/bun:1.0.2

#### From within the container
export PATH=$PATH:/root/.bun/bin
export BUN_INSTALL=/root/.bun
bun i -g cdktf-cli
bunx cdktf

### What is the expected behavior?

It prints the CLI options

### What do you see instead?

```
see https://github.com/jprichardson/node-fs-extra/issues/269
```
then a bunch of minified JS then
```
EBADF: Bad file descriptor
 syscall: "fstat"
   errno: -9
      at aC (/root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf.js:53:32710)
      at /root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf.js:53:33219
      at /root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf.js:1:272
      at /root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf.js:54:1931
      at globalThis (/root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf.js:77:0)
      at require (:1:20)
      at /root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf:2:0
      at globalThis (/root/.bun/install/global/node_modules/cdktf-cli/bundle/bin/cdktf:2:0)
error: "cdktf" exited with code 1 (SIGHUP)
```

### Additional information

I've seen different errors depending on the host OS I'm using. Sometimes it's this error, sometimes I see 

```
error: SystemResources

----- bun meta -----
Bun v1.0.0 (822a00c4) Linux arm64 #1 SMP PREEMPT Thu May 25 07:27:39 UTC 2023
BunxCommand:
Elapsed: 5ms | User: 2ms | Sys: 4ms
RSS: 33.55MB | Peak: 11.87MB | Commit: 33.55MB | Faults: 0
----- bun meta -----

0   0xaaaac9e1f478
1   ???
2   ???
3   ???
4   ???
5   ???

Crash report saved to:
  ~/.bun-crash/v1.0.0-1694633037924.crash

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord
```

[...]

PR #6398, opened 2023-10-10
fix syscall error handling

### LINKED ISSUE #4723: Bun & Docker issue

Darwin 22.6.0 arm64 arm

Create a new SvelteKit app as outlined in the docs: https://bun.sh/guides/ecosystem/sveltekit

Create a `Dockerfile` like so:

```docker
FROM oven/bun:1.0

WORKDIR /app

COPY package.json .
COPY bun.lockb .

RUN bun install

COPY . .

# start app
CMD ["bun", "run", "dev"]
```

Create a `docker-compose.yml` for convenience:

```yaml
version: '3.8'

services:
  ui:
    build:
      context: .
      dockerfile: Dockerfile
    ulimits:
      memlock:
        soft: -1
        hard: -1
    volumes:
      - './:/app'
      - '/app/node_modules'
    ports:
      - '5173:5173'
``` 

Start the environment with `docker-compose up`

Running in a docker container shouldn't produce errors, just like when you run it on the host machine. tldr, it works fine on the host but fails in a Dockerized environment with file descriptor issues.

### What do you see instead?

```bash
docker-compose up
[+] Running 1/1
 ⠿ Container my-app-ui-1  Recreated                                                                         0.5s
Attaching to my-app-ui-1
my-app-ui-1  | $ vite dev --host
my-app-ui-1  | 
my-app-ui-1  | Failed to load PostCSS config: Failed to load PostCSS config (searchPath: /app): [EBADF] Bad file descriptor
my-app-ui-1  | undefined
my-app-ui-1  |    code: "EBADF"
my-app-ui-1  |  syscall: "fstat"
my-app-ui-1  |    errno: -9
my-app-ui-1  | 
my-app-ui-1  | Failed to load PostCSS config: Failed to load PostCSS config (searchPath: /app): [EBADF] Bad file descriptor
my-app-ui-1  | undefined
my-app-ui-1  |    code: "EBADF"
my-app-ui-1  |  syscall: "fstat"
my-app-ui-1  |    errno: -9
my-app-ui-1  | 
my-app-ui-1  | Failed to load PostCSS config: Failed to load PostCSS config (searchPath: /app): [EBADF] Bad file descriptor
my-app-ui-1  | undefined
my-app-ui-1  |    code: "EBADF"
my-app-ui-1  |  syscall: "fstat"
my-app-ui-1  |    errno: -9
my-app-ui-1  | 
my-app-ui-1  | Failed to load PostCSS config: Failed to load PostCSS config (searchPath: /app): [EBADF] Bad file descriptor
my-app-ui-1  | undefined
my-app-ui-1  |    code: "EBADF"
my-app-ui-1  |  syscall: "fstat"
my-app-ui-1  |    errno: -9
my-app-ui-1  | 
my-app-ui-1  | 
my-app-ui-1  |   VITE v4.4.9  ready in 711 ms
my-app-ui-1  | 
my-app-ui-1  |   ➜  Local:   http://localhost:5173/
my-app-ui-1  |   ➜  Network: http://172.20.0.2:5173/
```

It appears to be have file descriptor errors trying to load a PostCSS config. I'm investigating where this comes from, but not familiar with SvelteKit/Vite generation of those files.   

### Additional information

_No response_

tobiaskohlbau:
Hello,

first of all I want to express that I'm new to bun. I've tried to built an SvelteKit application and ran into the same issue as #4723. Afterwards I jumped down deep into that rabbit hole :-D 

Fixes #4723

Possible related to some extend to one or more of #4671 #5935 #5943 #4896 #5306 #1446 

### What does this PR do?

The error handling between calling zig syscalls and using libc is different. Within zigs implementation, and therefore it's clone within bun, the return value of a syscall is of type usize. Libc implemenations use a signed integer namely c_int. As soon as libc is linked every response code of a syscall is interpreted via the libc errno function, but not every syscall is made through the libc interface. Only a few syscalls on aarch64 are used from libc every other one is called through zig itself. This could lead to errors beeing misinterpreted as success, cause its strictly positive and therefore not an error for the libc implementation. If an open syscall internally fails but is interpreted as a success the following write syscall would operate on a wrong file descriptor resulting in error bad file descriptor (EBADF).

I've built a release version of bun and verified that the expected behaviour is observed.
Before that I've built a reproducer to force the bad behaviour.

1. Create a _index.ts_ with the below content.
2. Create a directory called test `mkdir test`
3. Run bun via `bun index.ts`
4. This should result in ISDIR error, cause the file gets opened an the syscall fails with _ERRISDIR_ but instead the open succeeds and the stat afterwards gets an _EBADF_.

# TypeScript reproducer

```TypeScript
import fs from "fs";
fs.writeFileSync(`test`, `hello world`);
```

The error from the second syscall is correctly interpreted as this is run on aarch64 and bun uses libc stat on aarch64.

https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/sys.zig#L31-L50

## Before
```
# ./packages/bun-linux-aarch64/bun index.ts
1 | import fs from "fs";
2 | fs.writeFileSync(`test`, `hello world`);
   ^
EBADF: Bad file descriptor
    fd: "-21"
 syscall: "write"
   errno: -9

      at /home/user/tmp/bun/index.ts:2:0
```

## After
```
# ./packages/bun-linux-aarch64/bun index.ts
1 | import fs from "fs";
2 | fs.writeFileSync(`test`, `hello world`);
   ^
EISDIR: Is a directory
   path: "test"
 syscall: "open"
   errno: -21

      at /home/user/tmp/bun/index.ts:2:0
```

# Zig reproducer

```Zig
const std = @import("std");
const os = std.os;
const linux = os.linux;
const system = linux;
const builtin = @import("builtin");

pub fn main() !void {
    std.log.info("libc {}", .{builtin.link_libc});
    const rc_libc = system.openat(@as(system.fd_t, @intCast(-100)), "test", 321, 438);
    switch (os.errno(rc_libc)) {
        .SUCCESS => std.log.info("success", .{}),
        .ISDIR => std.log.info("is directory", .{}),
        else => return,
    }

    const rc_linux = system.openat(@as(system.fd_t, @intCast(-100)), "test", 321, 438);
    switch (linux.getErrno(rc_linux)) {
        .SUCCESS => std.log.info("success", .{}),
        .ISDIR => std.log.info("is directory", .{}),
        else => return,
    }
}
```

## Output after building with _zig build-exe main.zig_
```
info: libc false
info: is directory
info: is directory
```

## Output after building with _zig build-exe main.zig --library c_
```
info: libc true
info: success
info: is directory
```


Jarred-Sumner:
Thank you very much for this and for the write-up

I suspect this change should only be applied on Linux aarch64 and not on Linux x64. But there should be some extra test failures in a few places if that's true (there are also a handful of test failures in CI right now, but these ones should have to do with error messages)

tobiaskohlbau:
> Thank you very much for this and for the write-up
> 
> I suspect this change should only be applied on Linux aarch64 and not on Linux x64. But there should be some extra test failures in a few places if that's true (there are also a handful of test failures in CI right now, but these ones should have to do with error messages)

As far as I understand shouldn't the following type switch handle both sys call interfaces and it's response types, either on aarch64 or x64 (usize vs c_int). That's my reason to only remove it instead of checking the architecture at comptime. I'm new to zig too, but is the type assertion (`@TypeOf`) comptime evaluated?

https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/sys.zig#L259-L262


Thanks for your response, please let me know if any changes are required.

Hanaasagi:
I think the meaning of `use_libc` is somewhat ambiguous. It only represents that some syscalls use functions from libc, such as `stat` on aarch64. However, we use this boolean value in a more generic `getErrno` function to handle the return values of every syscall.

https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/sys.zig#L23-L24

https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/sys.zig#L247-L256


`openatOSPath` is implemented directly through syscall on Linux but via `openat` on macOS.

https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/sys.zig#L398-L420

In the `getErrno` function, the `use_libc` variable differs across these three platforms:

- Linux x86_64: `use_libc = false`
- macOS: `use_libc = true`
- aarch64: `use_libc = true`

Both aarch64 and macOS have `use_libc` set to true, but aarch64 uses syscall, while macOS uses `openat`. This leads to correct behavior on macOS but incorrect behavior on aarch64.

tobiaskohlbau:
What you're stating is correct, I've tried to say the same within the commit message itself 😅 Looks like it was not as clear.

tobiaskohlbau:
@Electroid I've seen you've merged the main into this. I'm unsure what to do from here, please let me know if the PR itself needs some update. I'm more than willing to get this over the finish line.

tarruda:
This will also close #4627 and #5820

This is an integer cast error which is deep into the error handling of multiple fs functions and will likely fix many Linux/arm64 issues. Would be a good idea to enable CI tests for Linux/arm64 after this is merged

Jarred-Sumner:
Closing in favor of https://github.com/oven-sh/bun/pull/6942

PR #6406, opened 2023-10-11
bunfig.toml GitHub api option

dylan-conway:
### What does this PR do?
adds a field to bunfig.toml to enable/disable using the github api for installing github dependencies

error message example:
Screenshot 2023-10-10 at 6 49 53 PM

- [ ] Code changes

tested manually


Electroid:
What if we fallback to git if it fails?

PR #6420, opened 2023-10-11
Update development.md adding zig to zsh path environment

samihenrique:
### What does this PR do?

The command bun install -g @oven/zig doesn't register ZIG path in my ZSH environment, adding the simple instructions above whe fix them adding all bun/bin in my zig environment

export PATH="$PATH:$HOME/.bun/bin"

- [ ] Code changes

### DIFF
--- a/docs/project/development.md
+++ b/docs/project/development.md
@@ -133,6 +133,12 @@ $ bun install -g @oven/zig
 $ zigup 0.12.0-dev.163+6780a6bbf
 ```
 
+If you are utilizing ZSH, it may be necessary to add the Bun binary directory to your PATH environment variable. To achieve this, append the following line to your .zshrc file:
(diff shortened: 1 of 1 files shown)


colinhacks:
These lines get added automatically with you install with the `curl` command: https://github.com/oven-sh/bun/blob/main/src/cli/install.sh#L205-L213

Sounds like this didn't work for you. Any ideas why?

paperclover:
if `$HOME/.bun/bin` wasnt in your path, how did you run bun? I think this should already be configured when installing bun, if merged this should be at the top with bun itself.

Electroid:
Going to close this, since this is potentially a bug with how bun was installed. If you're able to repro this, please open another issue and we'll take a look.

PR #6435, opened 2023-10-12
feat(install): support npm overrides/yarn resolutions, one level deep only

paperclover:
### What does this PR do?

overrides and resolutions. yippee

- [ ] Code changes


ghiscoding:
This is obviously a feature, so I wonder why was this released as `1.0.6` and not `1.1.0`, it doesn't look like semver is followed in this project, am I right?

Issue #6413, opened 2023-10-11
`server.address()` does not handle unix sockets


Hanaasagi:
### What version of Bun is running?

1.0.5+1bf28e0d77a8b2261befbdb708cefd03e0126960

### What platform is your computer?

Linux 6.5.6-arch2-1 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const http = require("node:http");

const server = http.createServer((req, res) => {});

setTimeout(() => {
  console.log(server.address());
  server.close();
}, 1000);
server.listen("./socket");

```

### What is the expected behavior?

node.js output:

```
./socket
```

### What do you see instead?

bun output:
```
{
  address: "",
  family: "IPv4",
  port: undefined
}

```

Hanaasagi:
I also found that when not specifying a `hostname` in `listen`, it results in getting the IP as `localhost`. It should be `127.0.0.1`.

```JavaScript
server.listen(3000);

// {
//   address: "localhost",
//   family: "IPv4",
//   port: 3000
// }
```

I think we should add some additional checks here and avoid using the `hostname` field. Address and hostname are different. Trying to fix this.

https://github.com/oven-sh/bun/blob/1bf28e0d77a8b2261befbdb708cefd03e0126960/src/js/node/http.ts#L448-L457

Issue #5850, opened 2023-09-21
ipaddr: the address has neither IPv6 nor IPv4 format


krassx:
### What version of Bun is running?

1.0.2

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. Create new Angular 16.x web application and try starting it with bun instead of node
2. During the spawning the local web server, the unhandled exception is raised with the reference for more details in the local file
3. Open the file and see the stack trace there.

### What is the expected behavior?

Local web server should start with no issues.

### What do you see instead?

Local web server launch fails with exception.

### Additional information

The stack trace for the exception is the following:

```
[error] Error: ipaddr: the address has neither IPv6 nor IPv4 format
    at <anonymous> (/Users/build/Projects/viewer/node_modules/ipaddr.js/lib/ipaddr.js:934:11)
    at logStatus (/Users/build/Projects/viewer/node_modules/webpack-dev-server/lib/Server.js:2883:31)
    at <anonymous> (/Users/build/Projects/viewer/node_modules/webpack-dev-server/lib/Server.js:3393:19)
    at processTicksAndRejections (:55:39)
```

(Btw, another issue here is that the line numbers are incorrect and point to unrelated lines, though they are somewhat close to the actual ones).

After digging into it a bit, it seems that bun and node return different values HTTP/HTTPS server address() method call.

Here is the code snippet from `Server.js` where the original data is fetched
```javascript
const { address, port } =
        /** @type {import("net").AddressInfo} */
        (
          /** @type {import("http").Server} */
          (this.server).address()
        );
```

It is later passed to the `ipaddr` in the following way:

```javascript
const parsedIP = ipaddr.parse(address);
```

So, when this code is run in node, address contains "127.0.0.1", whereas bun returns "localhost". `ipaddr` does not expect address to be aliased this way and fails.

JohnDDuncanIII:
`bun --bun run dev` fails on our company repo for the same reason

![Screenshot 2023-10-04 at 12 25 11 AM](https://github.com/oven-sh/bun/assets/9257785/3c800328-7660-46fe-9c04-fd2c57d24065)

PR #6442, opened 2023-10-12
fix(node:http): fix `server.address()`

Hanaasagi:
### What does this PR do?

- Close: https://github.com/oven-sh/bun/issues/6413
- Close: https://github.com/oven-sh/bun/issues/5850

Previously, when obtaining the address, we used the `hostname` field, which could result in getting the string `localhost`. This PR implemented the `address()` function in zig side.

https://github.com/oven-sh/bun/blob/1bf28e0d77a8b2261befbdb708cefd03e0126960/src/js/node/http.ts#L448-L457

- [x] Code changes

I wrote automated tests


src/bun.js/api/server.classes.ts
```diff
@@ -45,6 +45,10 @@ function generate(name) {
         getter: "getHostname",
         cache: true,
       },
+      address: {
+        fn: "getAddress",
```
Jarred-Sumner: This should be a getter with `cache: true`

src/bun.js/api/server.zig
```diff
@@ -4834,6 +4834,13 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
         temporary_url_buffer: std.ArrayListUnmanaged(u8) = .{},
 
         cached_hostname: bun.String = bun.String.empty,
+        cached_localAddr: struct {
```
Jarred-Sumner: Using a getter, we won't need to cache it here

Instead, the `JSValue` will only be created once

src/bun.js/api/server.zig
```diff
@@ -5297,6 +5304,44 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
...
 
+        pub fn getAddress(this: *ThisServer, globalThis: *JSGlobalObject, _: *JSC.CallFrame) callconv(.C) JSC.JSValue {
+            switch (this.config.address) {
+                .unix => |unix| {
+                    var value = bun.String.create(bun.sliceTo(@constCast(unix), 0));
+                    defer value.deref();
+                    return value.toJS(globalThis);
+                },
+                .tcp => {
+                    if (this.cached_localAddr.address.isEmpty()) {
```
Jarred-Sumner: This should reuse the `JSC::Structure` object which I think @paperdave had done in earlier code for `requestIp` 

As-is, this will be pretty slow to create these objects if it's called on each request

Jarred-Sumner: oh I'm misunderstanding, this is per server

it's probably fine, though a getter + cache: true is still a good idea

src/bun.js/api/server.zig
```diff
@@ -5297,6 +5304,44 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
...
+                                    .address = bun.String.create(buf[0..@as(usize, @intCast(len))]),
+                                    .is_v4 = is_v4,
+                                };
+                            }
+                        }
+                    }
+                    if (this.cached_localAddr.address.isEmpty()) {
+                        return JSValue.jsNull();
+                    }
+                    var result = JSC.JSValue.createEmptyObject(globalThis, 3);
```
paperclover: if you look at the requestIP, specficically `JSSocketAddress.cpp` code there is already a binding and structure to create this object, but as jarred said the performance of this isn't that much of a concern

Hanaasagi: Ok, I reuse the `JSSocketAddress` class.

src/bun.js/api/server.zig
Hanaasagi: Fixed. Now using a getter and `cache: true`.

test/js/web/fetch/fetch.stream.test.ts
```diff
@@ -213,7 +213,12 @@ describe("fetch() with streaming", () => {
           .listen(0);
 
         const address = server.address() as AddressInfo;
-        const url = `http://${address.address}:${address.port}`;
+        let url;
+        if (address.family == "IPv4") {
+          url = `http://${address.address}:${address.port}`;
+        } else {
+          url = `http://[${address.address}]:${address.port}`;
```
Hanaasagi, line 220: IPv6 need to be enclosed in brackets. The previous tests passed because `localhost` was returned.

src/deps/uws.zig
```diff
@@ -306,6 +306,59 @@ pub fn NewSocketHandler(comptime is_ssl: bool) type {
...
+        /// Get the local address of a socket in text format.
+        ///
+        /// # Arguments
+        /// - `buf`: A buffer to store the text address data.
+        /// - `length`: A pointer to an integer representing the length of the buf.
+        /// - `is_ipv6: A pointer to an boolean representing whether address is IPv6.
+        ///
+        /// # Returns
+        /// This function returns void, and updated `buf` and `length` inplace.
+        pub fn localAddressText(this: ThisSocket, buf: [*]u8, length: *i32, is_ipv6: *bool) void {
```
paperclover: i wish this code was 'safer'. i think the handling of memory is ok here after reading it very carefully.

maybe it should return a slice instead of mutating a length pointer? this way it would be harder for the caller to access the undefined memory in `buf`.

Hanaasagi: Now it returns a slice on success.
```zig
        pub fn localAddressText(this: ThisSocket, buf: []u8, is_ipv6: *bool) ?[]const u8 {

```

paperclover (approved):

PR #6447, opened 2023-10-12
add gitpod enviroment for quick setup 

Abhinavpatel00:
### What does this PR do?

adds an instant dev enviroment configuration for gitpod


Electroid:
Thanks for submitting this change. For now, I think we'll stick with devcontainer. It's going to be complicated to have many different dev environments that we need to maintain, and will likely become out-of-date.

PR #6451, opened 2023-10-12
Adds systemd guide to run a bun application as a daemon

RaresAil:
### What does this PR do?

Adds a guide for using `systemd` to run a bun application as a daemon

### DIFF
--- a/docs/guides/ecosystem/systemd.md
+++ b/docs/guides/ecosystem/systemd.md
@@ -19,7 +19,7 @@ $ touch my-app.service
 
 ---
 
-Here is a typical service file that runs an application on system start. You can use this as a template for your own service.
+Here is a typical service file that runs an application on system start. You can use this as a template for your own service. Refer to the [systemd documentation](https://www.freedesktop.org/software/systemd/man/systemd.service.html) for more information on each setting.
 
 ```ini
 [Unit]
@@ -83,7 +83,7 @@ $ sudo systemctl status my-app
 
 ---
 
-To update the service, edit the contents of the service file then reload the daemon.
+To update the service, edit the contents of the service file, then reload the daemon.
 
 ```bash
 sudo systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


zjhendryx:
@RaresAil One quick change I would suggest is, you have "User=root" in your systemd file, when going by your guide you likely want that to be the value of YOUR_USER, as it doesn't need to run as root. So it's best to avoid a possible security situation, especially since you have both bun and the application installed in /home/YOUR_USER following your guide.

RaresAil:
You are right, I will modify that, in copy pasted from my current case because i run the app as root cause i was lazy to give permission to my user to can run the app on port 443 😂

RaresAil:
@zjhendryx i added that and explain how to allow the non-root user to use the privileged ports

colinhacks:
Made some fairly heavy edits. I don't think there's much value in re-iterating systemd documentation here—better to provide a minimal guide that is easy to modify and follow.

docs/guides/ecosystem/systemd.md
```diff
@@ -0,0 +1,103 @@
...
+# usually you'll use 'simple'
+# one of https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=
+Type=simple
+# which user to use when starting the app
+User=YOUR_USER
+# path to your application's root directory
+WorkingDirectory=/home/YOUR_USER/path/to/my-app
+# the command to start the app
+# requires absolute paths
+ExecStart=/home/YOUR_USER/.bun/bin/bun run index.ts
```
colinhacks, line 43: I updated this to just `index.ts`. Ideally no build step would be necessary with Bun. 

I also didn't understand why you had an absolute path to `dist/index.js` before - we've already set a working directory so that shouldn't be necessary right?

RaresAil: You are right, also thank you for the simplification of the guide

colinhacks:
Thanks @RaresAil, this was excellent!

karimfromjordan:
> @RaresAil One quick change I would suggest is, you have "User=root" in your systemd file, when going by your guide you likely want that to be the value of YOUR_USER, as it doesn't need to run as root. So it's best to avoid a possible security situation, especially since you have both bun and the application installed in /home/YOUR_USER following your guide.

Small suggestions, `DynamicUser=yes` would be even better. systemd will then dynamically allocate a user for you and also enable `PrivateTemp`,  `ProtectHome` and `ProtectSystem` automatically.

PR #6454, opened 2023-10-12
docs: fix typo and add a direct troubleshooting for `installation.md`

cyfung1031:
### What does this PR do?

1. Fix typo in the zsh code tab
2. arrange the order to align with the code tabs
3. add a direct troubleshooting to add the installation path based on the shell name.

- [ ] Code changes

N/A

cc: @colinhacks

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -89,7 +89,7 @@ f you've installed Bun but are seeing a `command not found` error, you may have 
 
 ```sh
 $ echo $SHELL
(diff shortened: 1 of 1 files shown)


colinhacks:
Thanks for the PR. I updated the ordering in a separate commit. I think including a complex-looking bash script here is a little overwhelming for users who may not be familiar with that, and its important to keep this page as approachable as possible, so I'd prefer to stick with the simplified instructions in this case. Thanks again.

PR #6466, opened 2023-10-12
Add failing test for writing binary values to a Buffer

aardvarkk:
### What does this PR do?

- [x] Code changes

It is a failing test I believe should pass. This test passes if the value written is in the range `[0x00, 0x7F]` but fails if the value written is in the range `[0x80, 0xFF]`. I filed [this bug report](https://github.com/oven-sh/bun/issues/6467) that links to this PR.


Electroid:
Thanks! I have a PR coming that will fix this.

Issue #4765, opened 2023-09-10
[Web APIs] - Request clone causes body data to be lost


medz:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```ts
Bun.serve({
    fetch(request) {
        const data1 = request.clone().text();
        const data2 = request.clone().text();
        console.log(data1, data2)// Output "" empty string.
        // ...
    }
});
```

### What is the expected behavior?

https://developer.mozilla.org/en-US/docs/Web/API/Request/clone

The request object's stream should be copied correctly

### What do you see instead?

No errors, but wrong behavior.

ceopaludetto:
This can be related also https://github.com/jacob-ebey/bun-remix/issues/3. When I tried to run a simple remix app using bun as a server, every version that I tried greater than or equal 0.8.0 doesn't work. So I'm stucked into 0.7.3 right now

ceopaludetto:
A minimal remix repro https://codesandbox.io/p/sandbox/suspicious-cloud-7fxrwk

hjonasson:
I got here through @ceopaludetto 's comment on `bun-remix`. If it helps, I built the example from his reference issue in both Node and Bun and found that this repros, body is preserved in Node and not Bun.

ceopaludetto:
Seems to be fixed on Bun 1.0.2

medz:
> Seems to be fixed on Bun 1.0.2

No

liz3:
Ive worked on this and hve found the solution but the pr isnt quite ready yet, give me a day and there should be a pr for this.

Electroid:
Duplicate of #6348

PR #6468, opened 2023-10-12
Handle Request/Response cloning with pending(.Locked) body.

liz3:
### What does this PR do?
When cloning a `Request` or `Response` before its body was settled(resolved) it would simply ignore the body and provide a empty one, this came from the fact that the `await fetch(..)` promise is resolved before the entire body is received in cases of bigger requests.

This changes that by introducing a ref back to the original response for clones AND making the original response(to which the FetchTasklet is also referenced) keep track of its clones, no matter the levels of clones in order to be able to resolve their bodies too when the response completes.

**Note**: This introduces a two way memory reference(ptrs) between the requests, i used a ArrayList of ptrs as well as a optional ptr for the original_response, the links are correctly unrefed when each response is finalized but maybe theres a more idiomatic way to archive this in zig rather then this two way ref approach.

Fixes: https://github.com/oven-sh/bun/issues/4765
Fixes: https://github.com/oven-sh/bun/issues/6969

- [x] Code changes

## To test this
the interval is not actually needed its to keep the process alive when testing deallocation of the responses,
```js
const f = async () => {
    const res = await fetch("https://github.com/");
    console.log("resolved");
    const clone = res.clone();
    const text1 = await clone.clone().text();
    const text2 = await clone.text();
    const orig = await res.text();
    console.log(text1.length, text2.length, orig.length, text2 === text1)

}
await f()
setInterval(() => {}, 500)
```

To test the serve part:
```js
const handle_req = async(request) => {

    console.log("awaiting first")
    const data1 = await request.clone().clone().text();
    console.log("awaiting second")
    const data2 = await request.clone().text();
    console.log(data1, data2)// Output "" empty string.
    return new Response("hello world");
}
const r = Bun.serve({
    async fetch(request) {
       return handle_req(request) 
    }
});

```


Jarred-Sumner:
@cirospaciari can you review this?

src/bun.js/webcore/response.zig
```diff
@@ -843,7 +883,30 @@ pub const Fetch = struct {
                                 var scheduled_response_buffer = this.scheduled_response_buffer.list;
 
                                 const chunk = scheduled_response_buffer.items;
-
+                                for (response.response_clones.?.items) |other_request| {
```
cirospaciari: What happens in the case we get the `readable_stream_ref`?

cirospaciari (changes requested):
Overall looks good, we should add tests for cloning a clone, clone + streaming, clone a clone in the middle of streaming, and test HTML Rewriter with cloned response etc.
If we have the `readable_stream_ref` stored on fetch probably will not reach the code path for updating all response clones if I read the code correctly.

src/bun.js/webcore/response.zig
liz3: will look at it

liz3: @cirospaciari readable_stream_ref seams to not be part of the response object
, can you specify how we should handle this?

cirospaciari: `readable_stream_ref` is created in the `Response` when we call `response.body.getReader()` after this we basically only update the stream because body would be used, in this case we still need to check and update `response_clones` but we do an early return  here https://github.com/oven-sh/bun/blob/3202486e55370827e9fe957422b4a8b81ee8d9cf/src/bun.js/webcore/response.zig#L991-L1019 and we also probably check the fail case here https://github.com/oven-sh/bun/blob/3202486e55370827e9fe957422b4a8b81ee8d9cf/src/bun.js/webcore/response.zig#L965-L989

We can still call `this.response.get()` because we do not discard this reference even when we get a `readable_stream_ref` in `onReadableStreamAvailable`, streams are a little confusing right now

liz3:
image

@Jarred-Sumner @cirospaciari i think this is ready for final review now, ive added a test suite which should cover most cases.
Theres some behaviour changes especially for streaming requests but thats needed for correct cloning support, let me know if you have any questions, reviews, feedback.

Jarred-Sumner:
I wonder if `req.clone` should go through ReadableStream `tee` instead? It will be slower but a lot less likely to cause crashes/instability than this variation

liz3:
> I wonder if `req.clone` should go through ReadableStream `tee` instead? It will be slower but a lot less likely to cause crashes/instability than this variation

Readablestream tee? Can you elaborate

test/js/web/fetch/body-cloning.test.ts
```diff
@@ -0,0 +1,556 @@
...
+    const reader1 = res.body.getReader();
+
+    const merge = (a, b) => {
+      const n = new Uint8Array(a.length + b.length);
+      n.set(a, 0);
+      n.set(b, a.length);
+      return n;
+    };
+
+    const readerToArray = async reader => {
```
Jarred-Sumner: `Bun.readableStreamToArray`

test/js/web/fetch/body-cloning.test.ts
```diff
@@ -0,0 +1,556 @@
...
+    });
+    const res = await requestServer();
+    const stream = res.body.getReader();
+    expect(() => {
+      res.clone();
+    }).toThrow();
+  });
+  it("reader only does not hold data", async () => {
+    const data = "this is a very very very very long response".repeat(100000);
+    const dataArray = new Uint8Array(Buffer.from(data));
```
Jarred-Sumner: `Buffer` is already a `Uint8Array`

liz3:
@Jarred-Sumner passthrough test works too

test/js/web/fetch/body-cloning.test.ts
```diff
@@ -0,0 +1,737 @@
...
+        "\r\n",
+      ].join("\r\n"),
+    );
+    await sleep(500);
+    for (const char of data.split("")) {
+      socket.write(`${char}`);
+      await sleep(150);
+    }
+    socket.write("\r\n\r\n");
+    socket.flush();
```
Jarred-Sumner: this doesn't check that the contents were actually written

liz3: @Jarred-Sumner why would you need to check that if the test is performed on the server side, but sure il add it

liz3: if the data is received on the server it HAS to have been written

src/bun.js/api/server.zig
```diff
@@ -1235,6 +1279,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         req: *uws.Request,
         signal: ?*JSC.WebCore.AbortSignal = null,
         method: HTTP.Method,
+        request_clones: ?ArrayList(*Request) = null,
```
Jarred-Sumner: are you sure it's okay for this to be an `?ArrayList` and not a `?*ArrayList`? Is there any chance we might push to it and then have a stale reference? 

if so, can you leave a comment? i also don't think this needs to be a n`?` type, as that makes it go from 96 bytes to possibly 128 bytes. instead, we could check if .cap == 0

Jarred-Sumner: how can you be confident of the lifetime of the `Request` pointer while it's inside this array? what prevents it from being garbage collected?

src/bun.js/api/server.zig
```diff
@@ -1618,10 +1685,19 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                 }
             }
 
+            if (this.request_clones) |clones| {
```
Jarred-Sumner: this feels suspicious

src/bun.js/api/server.zig
```diff
@@ -1681,6 +1757,23 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
+                        if (body.value == .Locked) {
+                            // the promise is pending
+                            if (body.value.Locked.action != .none or body.value.Locked.promise != null) {
+                                this.pending_promises_for_abort += 1;
+                            } else if (body.value.Locked.readable != null) {
+                                body.value.Locked.readable.?.abort(this.server.globalThis);
+                                body.value.Locked.readable = null;
+                                any_js_calls = true;
+                            }
+                            body.value.toErrorInstance(JSC.toTypeError(.ABORT_ERR, "Request aborted", .{}, this.server.globalThis), this.server.globalThis);
```
Jarred-Sumner, line 1770: how do we ensure this Request does not leak?

src/bun.js/api/server.zig
liz3: @Jarred-Sumner https://github.com/oven-sh/bun/pull/6468/files#diff-e9484fda309708f80db81a125281dd0b9fdadaf3d2400ae8e78a2b8fc14d85ebR1336-R1345 and https://github.com/oven-sh/bun/pull/6468/files#diff-79f37ca42929ed1662aa69e017b6445a2cb38aa0a9e6c989a85e403eb32c07c9R287-R288

src/bun.js/webcore/request.zig
```diff
@@ -72,6 +72,8 @@ pub const Request = struct {
     request_context: JSC.API.AnyRequestContext = JSC.API.AnyRequestContext.Null,
     https: bool = false,
     upgrader: ?*anyopaque = null,
+    js_ref: ?JSC.JSValue = null,
```
Jarred-Sumner: we should not need to store the JSValue in here. If we did need it, we would never need to store it as an `?JSValue` because `.zero` is never visible to JS and therefore can serve as the optional value.

src/bun.js/webcore/request.zig
```diff
@@ -281,7 +283,9 @@ pub const Request = struct {
             headers.deref();
             this.headers = null;
         }
-
+        this.maybeUnprotect();
```
Jarred-Sumner: we should not need to use protect/unprotect in here. We could use a JSC.Weak and keep it alive that way. But protect/unprotect is very slow and we should not need it.

src/bun.js/webcore/request.zig
```diff
@@ -728,8 +732,32 @@ pub const Request = struct {
         globalThis: *JSC.JSGlobalObject,
         _: *JSC.CallFrame,
     ) callconv(.C) JSC.JSValue {
+        if (this.body.value == .Used or (this.body.value == .Locked and (this.body.value.Locked.promise != null or this.body.value.Locked.readable != null))) {
```
Jarred-Sumner, line 732: We probably do need this check in here regardless

src/bun.js/webcore/request.zig
```diff
@@ -728,8 +732,32 @@ pub const Request = struct {
...
+            const err = JSC.toTypeError(.ERR_INVALID_ARG_VALUE, "Request.clone: body already used", .{}, globalThis);
+            globalThis.throwValue(err);
+            return .zero;
+        }
         var cloned = this.clone(getAllocator(globalThis), globalThis);
-        return cloned.toJS(globalThis);
+        var js_value = cloned.toJS(globalThis);
+        if (cloned.has_parent_ref) {
+            cloned.js_ref = js_value;
+            js_value.protect();
```
Jarred-Sumner: as mentioned earlier, we should not use protect here

src/bun.js/webcore/response.zig
```diff
@@ -64,9 +65,12 @@ pub const Response = struct {
 
     allocator: std.mem.Allocator,
     body: Body,
+    original_response: ?*Response = null,
+    response_clones: ?ArrayList(*Response) = null,
```
Jarred-Sumner: it should be part of Body, not Response.

src/bun.js/webcore/response.zig
```diff
@@ -64,9 +65,12 @@ pub const Response = struct {
...
     init: Init,
     url: bun.String = bun.String.empty,
     redirected: bool = false,
+    js_value: ?JSC.JSValue = null,
```
Jarred-Sumner: use JSC.Weak, not this

Jarred-Sumner:
Please see this document from WebKit about memory management and lifetimes: https://github.com/WebKit/WebKit/blob/main/Introduction.md#active-dom-objects

We can use a WeakHandle when a request/response body is cloned to ensure it is alive for the necessary duration to dispatch messages to each clone'd body as needed. protect and unprotect will cause crashes and otherwise be very difficult to maintain. We should avoid using it in the vast majority of cases

test/js/web/fetch/body-cloning.test.ts
Jarred-Sumner: `socket.write` returns the number of bytes written. `socket.flush` does too. if the number of bytes written is not the expected number, the test will be flaky (for reasons unrelated to whether or not the code being tested works)

liz3:
@Jarred-Sumner could you look over this when you have a free minute 🙂

liz3:
Was there any test failure related to this, i checked and didn't see one

danthegoodman1:
really need this in, @Jarred-Sumner friendly ping :)

reaperkrim:
Hi guys, any chance on fixing this issue in near future (is there even any plan to fix it this year)? This issue makes Bun unusable for caching big server responses.

@Jarred-Sumner any chances?

Jarred-Sumner:
Superseeded by https://github.com/oven-sh/bun/pull/12473

PR #6471, opened 2023-10-12
Upgrade zig to 0.12.0-dev.888+130227491

paperclover:
### What does this PR do?

- bump version
- Workaround https://github.com/ziglang/zig/issues/16197 by using explicit `@as`
- Changes to std library and `@fabs` -> `@abs`
- typo toJSReponse -> toJSRe**s**ponse

zig builds on M2 go from ~55 second --> ~50 second 🎉


paperclover:
Tracked down the compiler crash in linux, same issue as linked :/

Build should pass now, hopefully tests do too

Issue #6456, opened 2023-10-12
url.pathToFileURL ignores cwd


jeron-diovis:
### What version of Bun is running?

1.0.5+1bf28e0d77a8b2261befbdb708cefd03e0126960

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

Save following snippet as `.mjs` file and execute it with Node and Bun:
```js
import { pathToFileURL } from 'node:url'
console.log(pathToFileURL('foo.js'))
```

### What is the expected behavior?

Output produced by Node:

```sh
~/dev/bun-test » node demo.mjs

URL {
  href: 'file:///Users/jeron_diovis/dev/bun-test/foo.js',
  origin: 'null',
  protocol: 'file:',
  username: '',
  password: '',
  host: '',
  hostname: '',
  port: '',
  pathname: '/Users/jeron_diovis/dev/bun-test/foo.js',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

`href` contains an absolute path to a current working directory plus given filename.

### What do you see instead?

Output produced by Bun:

```sh
~/dev/bun-test » bun demo.mjs

URL {
  href: "file:///foo.js",
  origin: "null",
  protocol: "file:",
  username: "",
  password: "",
  host: "",
  hostname: "",
  port: "",
  pathname: "/foo.js",
  hash: "",
  search: "",
  searchParams: {},
  toJSON: [Function: toJSON],
  toString: [Function: toString]
}
```

`href` contains only a filename, ignoring current working directory.

### Additional information

I'm facing this trying to run my existing project with Bun instead of Node. One of my dependencies calls `pathToFileURL` to generate an import path – and thus app crashes because obviously can't resolve a module.

PR #6472, opened 2023-10-12
fix: correctly prepend system path to relative urls on `pathToFileURL`

liz3:
The `pathToFileURL` was not handling the case that when a relative link is passed, it should convert it to a absolute url and prepend the top_folder, or otherwise described, resolve the path before creating the url.

This does that by calling into buns Path api and using its `resolve` function.

Theres the detail that i used preprocessor directives to determine whether we are on windows because there does not seam to be a easy way to get that at runtime in the `BunObject.cpp` file.

Fixes: https://github.com/oven-sh/bun/issues/6456

### What does this PR do?

- [x] Code changes


src/bun.js/bindings/BunObject.cpp
```diff
@@ -506,7 +506,13 @@ JSC_DEFINE_HOST_FUNCTION(functionPathToFileURL, (JSC::JSGlobalObject * lexicalGl
     auto& globalObject = *reinterpret_cast<Zig::GlobalObject*>(lexicalGlobalObject);
     auto& vm = globalObject.vm();
     auto throwScope = DECLARE_THROW_SCOPE(vm);
-    auto path = JSC::JSValue::encode(callFrame->argument(0));
+    auto path = JSC::JSValue::encode(callFrame->uncheckedArgument(0));
```
paperclover: uncheckedArgument will crash if the argument is not provided. since this is user exposed we cannot assume this

src/bun.js/bindings/BunObject.cpp
```diff
@@ -506,7 +506,13 @@ JSC_DEFINE_HOST_FUNCTION(functionPathToFileURL, (JSC::JSGlobalObject * lexicalGl
...
+    // i dont know a runtime way to check this, help
```
paperclover: TIL `#if OS(WINDOWS)`. we can use that here

paperclover (changes requested):

liz3:
@paperdave addressed reviews!

paperclover:
what's the reason it has to be a runtime check, i think inlining the boolean true or false is fine

liz3:
> what's the reason it has to be a runtime check, i think inlining the boolean true or false is fine

@paperdave there is no inherit reason for it to be runtime checked, but when i was implementing the fix i first thought about implementing this myself but then when counterchecking this would require some more complex logic thought it might be more worth using existing logic, i.e the path module, the c++ binding for the path module sets a [internal property](https://github.com/oven-sh/bun/blob/main/src/bun.js/bindings/Path.cpp#L44-L45) on the global object which is runtime defined and so i didn't want to break that pattern but was unable to find a quick way of accessing it which made a preprecessor directive more reasonable.

paperclover (approved):

src/bun.js/bindings/BunObject.cpp
```diff
@@ -507,6 +507,12 @@ JSC_DEFINE_HOST_FUNCTION(functionPathToFileURL, (JSC::JSGlobalObject * lexicalGl
     auto& vm = globalObject.vm();
     auto throwScope = DECLARE_THROW_SCOPE(vm);
     auto path = JSC::JSValue::encode(callFrame->argument(0));
+    // NOTE(Liz3): this is supposed to be replaced with a runtime call at some point maybe.
+#if OS(WINDOWS)
+    path = Bun__Path__resolve(lexicalGlobalObject, true, reinterpret_cast<JSC__JSValue*>(&path), 1);
+#else
+    path = Bun__Path__resolve(lexicalGlobalObject, false, reinterpret_cast<JSC__JSValue*>(&path), 1);
+#endif
```
```suggestion
  RETURN_IF_EXCEPTION(throwScope, {});
```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -507,7 +507,13 @@ JSC_DEFINE_HOST_FUNCTION(functionPathToFileURL, (JSC::JSGlobalObject * lexicalGl
...
     auto throwScope = DECLARE_THROW_SCOPE(vm);
     auto path = JSC::JSValue::encode(callFrame->argument(0));
-
+    // NOTE(Liz3): this is supposed to be replaced with a runtime call at some point maybe.
+#if OS(WINDOWS)
+    path = Bun__Path__resolve(lexicalGlobalObject, true, reinterpret_cast<JSC__JSValue*>(&path), 1);
+#else
+    path = Bun__Path__resolve(lexicalGlobalObject, false, reinterpret_cast<JSC__JSValue*>(&path), 1);
+#endif
+  RETURN_IF_EXCEPTION(throwScope, {});
```
```suggestion
   RETURN_IF_EXCEPTION(throwScope, {});
```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -507,7 +507,13 @@ JSC_DEFINE_HOST_FUNCTION(functionPathToFileURL, (JSC::JSGlobalObject * lexicalGl
...
     auto throwScope = DECLARE_THROW_SCOPE(vm);
     auto path = JSC::JSValue::encode(callFrame->argument(0));
-
+    // NOTE(Liz3): this is supposed to be replaced with a runtime call at some point maybe.
+#if OS(WINDOWS)
+    path = Bun__Path__resolve(lexicalGlobalObject, true, reinterpret_cast<JSC__JSValue*>(&path), 1);
+#else
+    path = Bun__Path__resolve(lexicalGlobalObject, false, reinterpret_cast<JSC__JSValue*>(&path), 1);
+#endif
+   RETURN_IF_EXCEPTION(throwScope, {});
```
```suggestion
    RETURN_IF_EXCEPTION(throwScope, {});
```

PR #6475, opened 2023-10-13
fix(runtime): improve IPC reliability + organization pass on that code

### LINKED ISSUE #6394: Fix reliability issue in IPC implementation
All calls to `socket.write` must check if it successfully wrote the entire contents and if not, it must enqueue to an internal buffer and then drain the internal buffer when `onWritable` is called.

This code fails to handle when `socket.write` returns 0 or otherwise doesn't write the entire contents:
https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/bun.js/ipc.zig#L114

This code is similarly wrong, for four reasons:
https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/bun.js/bindings/Serialization.cpp#L34-L38

1) It needs to check if the entire contents has successfully been written and if not, enqueue the remaining
2) It is treating the socket as a file descriptor which is correct in this case but it really should go through usockets 
3) `write` is not the best system call to use when writing data to a socket. It should use `sendto` so that we can ensure it is always non-blocking. 
4) It should not be 3 separate `write` calls. It should be a single `sendto` call. 

`onWritable` needs to be implemented here:
https://github.com/oven-sh/bun/blob/6301778a589254e2c3c0d95f768fce303f528b03/src/bun.js/ipc.zig#L218-L221

In `onWritable`, we need to call `socket.write` on all the pending data

paperclover:
### What does this PR do?

Fixes #6394, this also does a pass over the code I wrote and improves it a bit.

Results of the organizing this means we have this new binding:

```zig
const serialized = value.serialize(globalThis) orelse return false;
defer serialized.deinit();

writer.write(serialized.data); // .data is a []const u8
```

Have to test more but this may finally fix Next.js/Nuxt IPC issues.


src/js/out/InternalModuleRegistryConstants.h
```diff
@@ -419,7 +419,7 @@ static constexpr ASCIILiteral NodeStreamPromisesCode = "(function (){\"use stric
...
+static constexpr ASCIILiteral NodeStreamWebCode = "(function (){\"use strict\";// src/js/out/tmp/node/stream.web.ts\nvar $;\n$ = {\n  ReadableStream: @ReadableStream,\n  ReadableStreamDefaultReader: @ReadableStreamDefaultReader,\n  ReadableStreamBYOBReader: @ReadableStreamBYOBReader,\n  ReadableStreamBYOBRequest: @ReadableStreamBYOBRequest,\n  ReadableByteStreamController: @ReadableByteStreamController,\n  ReadableStreamDefaultController: @ReadableStreamDefaultController,\n  TransformStream: @TransformStream,\n  TransformStreamDefaultController: @TransformStreamDefaultController,\n  WritableStream: @WritableStream,\n  WritableStreamDefaultWriter: @WritableStreamDefaultWriter,\n  WritableStreamDefaultController: @WritableStreamDefaultController,\n  ByteLengthQueuingStrategy,\n  CountQueuingStrategy,\n  TextEncoderStream: class {\n  },\n  TextDecoderStream: class {\n  },\n  CompressionStream: @undefined,\n  DecompressionStream: @undefined\n};\nreturn $})\n"_s;
```
paperclover: don't do this

PR #6478, opened 2023-10-13
Guide to containerize a bun application using Docker

RaresAil:
### What does this PR do?

Adds basic documentation on how to containerize  a bun application using docker

### DIFF
--- a/docs/guides/ecosystem/docker.md
+++ b/docs/guides/ecosystem/docker.md
@@ -1,0 +1,109 @@
+---
+name: Containerize a bun application using Docker
+---
+
+[Docker](https://www.docker.com) is a set of platform as a service products that use OS-level virtualization to deliver software in packages called containers. The service has both free and premium tiers.
+
+---
+
+For this first you need to install Docker Desktop to can build docker images, this can be downloaded from their [official web application](https://www.docker.com/products/docker-desktop/)
+
+---
+
+Now that you have docker, we can start writing the `Dockerfile`, this file should be in the root of your application, a basic `Dockerfile` that uses a build stage can be this, if you don't need a build stage you can remove it from the file.
+
+```dockerfile#Dockerfile
+# * Here we use the official image of bun which already has bun installed, you can specify the exact version
+# * Like 1.0.6-alpine or always get the latest version with alpine e.g. FROM oven/bun:alpine as base
+# * Getting always the latest is not such a great idea as it might introduce breaking changes
+# * You have 2 more options which are 1-alpine and 1.0-aline, which will use the version 1.x.x (Latest minor version) or 1.0.x (Latest revision version)
+FROM oven/bun:1.0.6-alpine as base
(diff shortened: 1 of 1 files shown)


docs/guides/ecosystem/docker.md
```diff
@@ -0,0 +1,109 @@
...
+---
+
+Now that you have docker, we can start writing the `Dockerfile`, this file should be in the root of your application, a basic `Dockerfile` that uses a build stage can be this, if you don't need a build stage you can remove it from the file.
+
+```dockerfile#Dockerfile
+# * Here we use the official image of bun which already has bun installed, you can specify the exact version
+# * Like 1.0.6-alpine or always get the latest version with alpine e.g. FROM oven/bun:alpine as base
+# * Getting always the latest is not such a great idea as it might introduce breaking changes
+# * You have 2 more options which are 1-alpine and 1.0-aline, which will use the version 1.x.x (Latest minor version) or 1.0.x (Latest revision version)
+FROM oven/bun:1.0.6-alpine as base
```
Electroid: I don't think alpine should be the default in the guide. Alpine is technically hacky, since we don't support MUSL.

RaresAil: Would debian be better? I usually choose alpine cause it has a smaller size

RaresAil: I added debian @Electroid

colinhacks:
Thanks again @RaresAil! I stripped it down a bit, as before. Great stuff.

RaresAil:
> Thanks again @RaresAil! I stripped it down a bit, as before. Great stuff.

Thanks 😄

abrahamsaanchez:
I think the optional test and build steps can fail, for example:

If I'm using `faker.js` as a devDependency (because I only use it in my tests to generate fake data), this DockerFile will try to run the tests without the devDependencies installed and will fail.

I think the tests step can be removed or do in another stage with the devDependencies to avoid this type of failures.

RaresAil:
> I think the optional test and build steps can fail, for example:
> 
> If I'm using `faker.js` as a devDependency (because I only use it in my tests to generate fake data), this DockerFile will try to run the tests without the devDependencies installed and will fail.
> 
> I think the tests step can be removed or do in another stage with the devDependencies to avoid this type of failures.

Did you used the exact code? As for tests is uses the node modules with the dev dependencies

abrahamsaanchez:
> > I think the optional test and build steps can fail, for example:
> > 
> > If I'm using `faker.js` as a devDependency (because I only use it in my tests to generate fake data), this DockerFile will try to run the tests without the devDependencies installed and will fail.
> > 
> > I think the tests step can be removed or do in another stage with the devDependencies to avoid this type of failures.
> 
> Did you used the exact code? As for tests is uses the node modules with the dev dependencies 

You're right, I didn't see that the test step is using all the dependencies. My bad 🤦🏼‍♂️.

Issue #6121, opened 2023-09-27
Hanging on `client.end()` with `pg`


colinhacks:
### What platform is your computer?

macos

### What steps can reproduce the bug?

```
bun add pg
touch .env
# set DATABASE_URL to connection string of postgres DB (`postgres://localhost:5432/test_db_name`)
```

```ts
import { Client } from "pg";

const client = new Client({
  connectionString: process.env.DATABASE_URL,
});

await client.connect();
await client.query(
  `CREATE TABLE IF NOT EXISTS persons (id serial PRIMARY KEY, name text NOT NULL);`
);
await client.query(`INSERT INTO persons (name) VALUES ('John');`);

try {
  const res = await client.query("SELECT * from persons");
  console.log(res.rows); // prints data
} catch (err) {
  console.error(err);
} finally {
  // hangs here
  await client.end();
  console.log("end");
}
```

Issue #6071, opened 2023-09-26
Hanging when querying to table using lib `pg` npm package


bozkayasalihx:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

its hanging when querying to selection query on table, its not problem  about postgres , i've check it its working. 
but i dont know its just keep hanging

Electroid:
Hi, could you provide some code so we can try to reproduce this?

colinhacks:
I can't replicate this. @bozkayasalihx can you share a reproduction schema & script?

bozkayasalihx:
Ok, i tried last night to check it out if problem persist but that time I changed to TLS configuration off and it worked. I assume problem caused by my side. Thanks for the great support

gggdomi:
This is not a perfect repro, but something quite suspicious and maybe related is happening for me.

If this is verified and all "big" queries done with bun + pg are impacted, it may be a bit critical? (if it's a regression)

Bun version: `1.0.6`
pg version: `8.11.3`
OS: `macOS 13.4 (M1 Pro)`

With this pretty simple code:
```js
// query.js
import pg from 'pg'

const DATABASE_URL = "postgres://user:pass@host:5432/dbname"

const client = new pg.Client({
    connectionString: DATABASE_URL,
    ssl: { rejectUnauthorized: false },
});

await client.connect();
await client.query(`SELECT id FROM some_table LIMIT 1`); // small queries work just fine, you need to retrieve 10-100k rows to crash
console.log('🟢');
const res = await client.query(`SELECT id FROM some_table LIMIT 100000`); // <--- bun stuck here
console.log('🟢', res.rows.length);
```

`node query.js` properly exits with all the right outputs after ~100ms

`bun run query.js` slowly exhausts (~2min) all my 32GB memory and never reaches the second console.log.  
Sometimes I get a short `[1]    15055 segmentation fault  bun run query.js`. 

Sometimes a much longer stack trace:
```
 8 |   var setup = () => {
 9 |     queue = function createQueue() {
10 | 
11 |       class FixedCircularBuffer {
12 | 
13 |         constructor() {
                                         ^
RangeError: Out of memory
      at new FixedCircularBuffer (:13:38)
      at push (:41:24)
      at nextTick (:93:38)
[...]

Issue #5902, opened 2023-09-22
Unable to acquire sequelize connection


dereke:
### What version of Bun is running?

1.0.2

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Save this code to 'sequelize.js' (install sequelize)

```js
import { QueryTypes, Sequelize } from "sequelize";

const sequelize = new Sequelize(null, null, null, {dialect: 'postgres'});
console.log(await sequelize.query('select 1;', {type: QueryTypes.SELECT}))
await sequelize.close()
```

Run it with `DEBUG=sequelize* bun sequelize.js`
The process will hang and the query will not run.

The output will look like this:
```
  sequelize:pool pool created with max/min: 5/0, no replication +0ms
  sequelize:connection:pg connection acquired +0ms
```

Run the same code with node:
`DEBUG=sequelize* node sequelize.js`

```js
  sequelize:pool pool created with max/min: 5/0, no replication +0ms
  sequelize:connection:pg connection acquired +0ms
  sequelize:connection:pg connection acquired +22ms
  sequelize:pool connection acquired +77ms
  sequelize:sql:pg Executing (default): select 1; +0ms
Executing (default): select 1;
  sequelize:sql:pg Executed (default): select 1; +2ms
  sequelize:pool connection released +5ms
[ { '?column?': 1 } ]
  sequelize:pool connection drain due to process exit +1ms
  sequelize:pool connection destroy +2ms
``

### What is the expected behavior?

The code should connect to the database, run the query and close the connection.

### What do you see instead?

The process hangs.

### Additional information

_No response_

scarabcoder:
Running into this same issue. 
As far as I can tell, it seems to be an issue with disconnecting from the Postgres server, when Sequelize boots up it does a Postgres version check, then disconnects, but the disconnect callback from the version check query is never being called, so it just hangs.

scarabcoder:
A bit more debugging, it appears this event is not being emitted when Stream#end is called: https://nodejs.org/api/stream.html#event-close

This appears to be an issue in the pg library, not Sequelize itself. 

Here's an excerpt of the code from node_modules/pg/lib/connection.js, the connect() function to open a new connection to Postgres server:

```JavaScript
connect(port, host) {
    var self = this

    this._connecting = true
    this.stream.setNoDelay(true)
    this.stream.connect(port, host)

    this.stream.once('connect', function () {
      if (self._keepAlive) {
[...]

Issue #4982, opened 2023-09-11
bun server crashes when returning very specific limit of data


japhex:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

We previously had a node server running tsx which was fine, we changed it to bun and we're getting 1 very specific and strange scenario.

We use **sequelize** to connect to a **postgres DB** on **Azure**, when we query any tables and produce over around **246** rows of results, the server hangs instantly. Up to that number we receive results within 30ms...

- if we run the generated sequelize query directly against the DB it works absolutely fine in 0.03s.
- if we run the server against a local postgres DB its also absolutely fine.

Which points to the Azure DB, but like I said we don't have any issues at all running with node.

libraries and versions:

@apollo/server: ^4.9.0
express: ^4.18.2
pg: ^8.11.3
pg-hstore: ^2.3.4
sequelize: ^6.33.0

### What is the expected behavior?

Expected behaviour is for responses containing more than 246 records to be returned like normal.

### What do you see instead?

No error or console output at all, the SQL statement gets printed as it is started and then the server completely hangs until being force stopped.

### Additional information

I can't think of anything, but would like to say in advance thank you **very** much for your time, we're super excited about using bun, especially in production so really hoping we can get to the bottom of this.

japhex:
Similarly if I reduce the complexity of the query and increase the LIMIT on the SELECT clause, it also works but again, only up until a certain point. I can then query for 300 records, but 350 still hangs.

It's very strange, its almost like the results are streaming into memory and past a certain point the bun server just hangs...BUT again, if I run this on a local postgres DB not the Azure one, its works absolutely fine.

japhex:
Also - sorry I did think of some info that may be relevant! 

Its an existing express server with an Apollo Server implementation, too. Just in case that makes any difference, I have thought about re-producing this in hono, but as its been fine with a local database instance, have been reluctant!

Zwimber:
You really haven't provided any steps to reproduce this bug

japhex:
> You really haven't provided any steps to reproduce this bug

I didn't really know what to put other than:

- Create a Burstable Azure Cosmos Postrgres DB with 2 vCores, 4GB RAM & 64GB of storage
- Create a table with 5 rows: id, string, string, date, date
- Fille the table with 7,270 rows of data
[...]

PR #6487, opened 2023-10-13
fix(net/tls) fix pg hang on end + hanging on query

cirospaciari:
### What does this PR do?

Fix: #6071
Fix: #6121
Fix: #4982
Fix: #5902 
Fix: #1623 

pg should execute big queries and exit with this code:
```js
const client = new pg.Client({
     connectionString: DATABASE_URL,
     ssl: { rejectUnauthorized: false },
});
await client.connect();
const res = await client.query(`SELECT * FROM users LIMIT 1000`);
console.log(res.rows);
await client.end()
```
sequelize should work and exit with this code:
```js
const sequelize = new Sequelize(DATABASE_URL, {dialect: 'postgres'});
console.log(await sequelize.query('SELECT * FROM users LIMIT 1000;', {type: QueryTypes.SELECT}))
await sequelize.close()
```

knex should work and exit now with this code:
```js

import Knex from "knex";

const knex = Knex({
  client: "pg",
  connection: DATABASE_URL,
});

const { rows } = await knex.raw("`SELECT * FROM users LIMIT 1000");
console.log(rows);
await knex.destroy();
```

- [x] Code changes

I wrote automated test + existing tests


src/js/node/stream.js
```diff
@@ -2608,7 +2608,8 @@ var require_readable = __commonJS({
     function addChunk(stream, state, chunk, addToFront) {
       $debug("adding chunk", stream.__id);
       $debug("chunk", chunk.toString(), stream.__id);
-      if (state.flowing && state.length === 0 && !state.sync && stream.listenerCount("data") > 0) {
+      const has_listenners = stream.listenerCount("data") > 0;
```
Jarred-Sumner: This code should very much mirror what Node.js does.

Is this what Node.js does?

src/js/node/net.js
```diff
@@ -466,6 +477,7 @@ const Socket = (function (InternalSocket) {
               // replace socket
               connection[bunSocketInternal] = raw;
               raw.timeout(raw.timeout);
+              this.on("end", this.#closeRawConnection);
```
Jarred-Sumner: `on` or `once`?

src/js/node/net.js
```diff
@@ -491,6 +503,7 @@ const Socket = (function (InternalSocket) {
                 // replace socket
                 connection[bunSocketInternal] = raw;
                 raw.timeout(raw.timeout);
+                this.on("end", this.#closeRawConnection);
```
Jarred-Sumner: `on` or `once`?

test/js/third_party/postgres/postgres.test.ts
```diff
@@ -19,6 +19,33 @@ describe("pg", () => {
...
+        expect(res.rows.length).toBeGreaterThanOrEqual(300);
+        await client.end();
+        return "success";
+      } catch (e) {
+        return e;
+      }
+    }
+
+    async function timeout() {
+      await Bun.sleep(5000);
```
Jarred-Sumner: This will always execute after the test has completed 

Which is not intended I think

src/js/node/stream.js
cirospaciari: let me check that

test/js/third_party/postgres/postgres.test.ts
cirospaciari: Yeah, will just rely on the test already implemented timeout:
https://github.com/oven-sh/bun/pull/6487/commits/a96898d20397ead25587bcdace5d09b02ab04077

src/js/node/net.js
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/6487/commits/a96898d20397ead25587bcdace5d09b02ab04077

src/js/node/net.js
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/6487/commits/a96898d20397ead25587bcdace5d09b02ab04077

src/js/node/stream.js
cirospaciari: Reverted it back in https://github.com/oven-sh/bun/pull/6487/commits/b68c074aa49d5f6ec9c9659bf9791d01f5c6bea3

src/js/node/net.js
```diff
@@ -537,6 +552,7 @@ const Socket = (function (InternalSocket) {
     _final(callback) {
       this[bunSocketInternal]?.end();
       callback();
+      process.nextTick(closeNT, this);
```
Jarred-Sumner, line 553: Is there a scenario where `"close"` can be emitted twice due to the use of `nextTick`?

Jarred-Sumner:
Thank you.

PR #6508, opened 2023-10-14
(console.log) Print className for an object if present

JibranKalia:
### What does this PR do?

This adds the functionality to print Object's class name if it exists matching Node's functionality.

This is my re-attempt from my earlier PR of https://github.com/oven-sh/bun/pull/6019

- [X] Code changes


src/bun.js/bindings/exports.zig
```diff
@@ -1839,6 +1832,26 @@ pub const ZigConsoleClient = struct {
...
+                    value.getPrototype(globalThis).getNameProperty(globalThis, &name_str);
+                    if (name_str.len > 0 and !strings.eqlComptime(name_str.slice(), "Object")) {
+                        return name_str;
+                    }
+
+                    value.getClassName(globalThis, &name_str);
+                    if (name_str.len > 0 and !strings.eqlComptime(name_str.slice(), "Object")) {
+                        return name_str;
+                    }
+                    return ZigString.init("");
```
JibranKalia: Not sure if this is the best way. I have a check `if (name_str.len > 0) {` that I would like to fail if we reach this case.

paperclover: use ZigString.static here

src/bun.js/bindings/exports.zig
```diff
@@ -1839,6 +1832,26 @@ pub const ZigConsoleClient = struct {
                     this.formatter.writeIndent(Writer, this.writer) catch {};
                 }
 
+                fn getObjectName(this: *@This(), globalThis: *JSC.JSGlobalObject, value: JSValue) ZigString {
```
paperclover: remove unused parameter `this`. it will make it so you call this fucntion just as `getObjectName(...)` and not `this.getObjectName(...)`

src/bun.js/bindings/exports.zig
```diff
@@ -1839,6 +1832,26 @@ pub const ZigConsoleClient = struct {
...
+                    _ = this;
+                    var name_str = ZigString.init("");
```
paperclover: i notice the code for getClassName seems to call `getNameProperty`.

is there a case getPrototype().getNameProperty() is actually needed, or calling getNameProperty before getClassName is the correct effect?

(you can find the source for these two in bindings.cpp, search for `JSC__JSValue__getClassName`)

src/bun.js/bindings/exports.zig
paperclover: actually these functions are nearly identical. init is fine here.

paperclover (changes requested):
looks great overall! minor things to be addressed is all.

src/bun.js/bindings/exports.zig
JibranKalia: > is there a case getPrototype().getNameProperty() is actually needed

I did a fair amount of testing to come up with an example where this would be necessary. But the standard `getClassName` seems to cover everything.

Lets run the full test suite and see if I missed anything.

src/bun.js/bindings/exports.zig
JibranKalia: Good point. Done.

JibranKalia:
@paperdave can you approve the test run. Just want to make sure my hypothesis about only needing once className is correct.

paperclover (approved):

paperclover:
Sorry for late review. this looks alright assuming it works. there are conflicts since last merged. Can you do a merge conflict locally `git merge origin/main`, and then make sure the tests pass locally?

JibranKalia:
Jarred's change in https://github.com/oven-sh/bun/pull/6612 also fixes this as well. Closing this PR. Thank you for reviewing though.

almmiko:
Hey @JibranKalia #6612 fixes the missing function name. The issue with the missing class name still exists

```sh
❯ bun-debug -v
1.0.9-debug
```

```js
// tmp/a.js
class Foo {}
console.log(new Foo());
```

```sh
❯ bun run /tmp/a.js
{}
```

JibranKalia:
@almmiko Thank you for pointing this out. I realized I was using an incorrect bun-debug. Will fix and reopen this MR.

JibranKalia:
@paperdave I have resolved the conflicts. @jaredpalmer in https://github.com/oven-sh/bun/pull/6612 switched the method call from `getNameProperty` to `getName`. However, with as much testing as I could do If found just calling `getClassName` on the values resulted with the we needed including function names. Furthermore, the `value.getPrototype(globalThis).getName(globalThis);` check seems redundant as well.

paperclover:
image
This bundler test needs to be updated (the test is now wrong)

paperclover (changes requested):
good but one failing test

JibranKalia:
@paperdave have updated the test: https://github.com/oven-sh/bun/pull/6508/commits/5c7453dd2c25a8b1243883d8849ffd5a045f8c19

paperclover (approved):
awesome

JibranKalia:
For some reason, the tests were still failing. I just find the GitHub actions output really hard to read.

paperclover:
when you fix tests, please run them locally before submitting.

paperclover:
thank you

src/bun.js/bindings/exports.zig
```diff
@@ -1973,6 +1965,15 @@ pub const ZigConsoleClient = struct {
             };
         }
 
+        fn getObjectName(globalThis: *JSC.JSGlobalObject, value: JSValue) ZigString {
+            var name_str = ZigString.init("");
+            value.getClassName(globalThis, &name_str);
+            if (name_str.len > 0 and !strings.eqlComptime(name_str.slice(), "Object")) {
```
Jarred-Sumner, line 1971: @paperdave this code has a mistake in it (which was likely copied from elsewhere)

It's asserting that the class name is always ascii.  It could be UTF-16. 

Instead, we can do `.eqlComptime` on the string itself.

JibranKalia: My bad. I copied this from https://github.com/oven-sh/bun/blob/68324daf7852343d895e415062f5ad659b511bee/src/bun.js/bindings/exports.zig#L1828.

I didn't replicate your update because I assumed incorrectly that it was the difference between ZigString vs BunString

JibranKalia:
> when you fix tests, please run them locally before submitting.

Will do in future PRs.

Issue #6507, opened 2023-10-14
Bun.file ignores type option


ghosTedBuddy:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Darwin 23.0.0 x86_64 i386

### What steps can reproduce the bug?

i am uploading a file and writing it into the temp directory without an extension (because "reasons"). for that i am loading the file ref based on [the docs](https://bun.sh/docs/api/file-io#writing-files-bun-write:~:text=%2C%20options%3F%3A-,%7B%20type%3F%3A%20string%20%7D,-)%3A%20BunFile) with the mime type from the formData

```
let tempFilePath = `/tmp/${Date.now().toString(36)}`;
console.log(`\n\n###\nwrite file to: ${tempFilePath}`);
await Bun.write(tempFilePath, upload);
const tempFile = Bun.file(tempFilePath, { type: upload.type });

console.log({
	tempFile,
	tempFileType: tempFile.type,
	ResponseContentType: new Response(tempFile).headers.get(
		'Content-Type'
	),
	uploadType: upload.type,
});
```

### What is the expected behavior?

i would assume the mimeType of the Blob is used, when referencing the file. kind of "overwriting" the interpreted mimeType. which would result in an output like this:

```
{
  tempFile: FileRef ("/tmp/lnq4ejul") {
    type: "text/markdown"
  },
  tempFileType: "text/markdown",
  ResponseContentType: "text/markdown",
  uploadType: "text/markdown"
}
```

### What do you see instead?

```
{
  tempFile: FileRef ("/tmp/lnq4ejul") {
    type: "application/octet-stream"
  },
  tempFileType: "application/octet-stream",
  ResponseContentType: "application/octet-stream",
  uploadType: "text/markdown"
}
```

### Additional information

i tried it with images, zip archives, markdown files and svg file. everything results in the same behaviour. the only way to make it work is by explicitly extract the file ending from the blob's name attribute and add it to the file path.

so either the option isn't really used or respected or i am misunderstanding the documentation or its a bug.

PR #6512, opened 2023-10-15
fix: support custom file type in Bun.file

liz3:
In the docs it seamed to suggest this is something supported, it seamed to be only supported in JSDOMFiles or blob. 

This Adds the 2 properties `type` and `lastModified` to be supported on `Bun.file`

Fixes: https://github.com/oven-sh/bun/issues/6507

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -1316,7 +1316,7 @@ pub const Blob = struct {
 
         var blob = Blob.findOrCreateFileFromPath(path, globalObject);
 
-        if (arguments.len == 2) {
+        if (arguments.len >= 2) {
             const opts = arguments[1];
 
             if (opts.isObject()) {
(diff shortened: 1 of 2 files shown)


src/bun.js/webcore/blob.zig
```diff
@@ -1314,7 +1314,39 @@ pub const Blob = struct {
...
+                    inner: {
+                        if (file_type.isString()) {
+                            var allocator = bun.default_allocator;
+                            var str = file_type.toSlice(globalObject, bun.default_allocator);
+                            defer str.deinit();
+                            var slice = str.slice();
+                            if (!strings.isAllASCII(slice)) {
+                                break :inner;
+                            }
+                            if (vm.mimeType(str.slice())) |entry| {
```
Jarred-Sumner, line 1334: if the mimeType exists in the hash table, we avoid cloning the string

https://github.com/oven-sh/bun/blob/a87aa2fafe05c358455173ea65877be5eefcbd63/src/bun.js/webcore/blob.zig#L3011-L3018

I suggest copying from the above pattern instead

src/bun.js/webcore/blob.zig
```diff
@@ -1314,7 +1314,39 @@ pub const Blob = struct {
             return .undefined;
         };
 
-        const blob = Blob.findOrCreateFileFromPath(path, globalObject);
+        var blob = Blob.findOrCreateFileFromPath(path, globalObject);
+
+        if (arguments.len == 2) {
```
Jarred-Sumner:
```suggestion
        if (arguments.len >= 2) {
```

We should ignore extra arguments rather than require an exact count of arguments (since that's more common in JS)

liz3: Fixed in https://github.com/oven-sh/bun/pull/6512/commits/775a31a2617aeae121824926e59c7e25fe733c2c

Jarred-Sumner:
Thank you

Jarred-Sumner:
Thank you

PR #6518, opened 2023-10-15
docs: More accurate decription of import.meta.dir

Creative-Difficulty:
### What does this PR do?

More accurate description of import.meta.dir

- [ ] Code changes

### DIFF
--- a/docs/api/import-meta.md
+++ b/docs/api/import-meta.md
@@ -19,7 +19,7 @@ import.meta.resolveSync("zod")
 ---
 
 - `import.meta.dir`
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
I'm confused how this is more accurate? `import.meta.dir` is not tied to the current working directory of the process. `process.cwd()` does that.

Creative-Difficulty:
> I'm confused how this is more accurate? `import.meta.dir` is not tied to the current working directory of the process. `process.cwd()` does that.

But it is exactly the same when actually tested

paperclover:
they are not the same when tested:

image

PR #6521, opened 2023-10-15
fix(node:worker_threads): ensure threadId property is exposed on worker_threads instance

jerome-benoit:
### What does this PR do?

Add threadId property missing on `Worker` instance from `node:worker_threads` module.

- [x] Code changes


src/js/node/worker_threads.ts
```diff
@@ -224,6 +224,10 @@ class Worker extends EventEmitter {
     this.#worker.addEventListener("open", this.#onOpen.bind(this));
   }
 
+  threadId() {
+    return _threadId;
```
Jarred-Sumner: Isn't this not the `Worker`'s threadId but the parent's? I think you want `this.#worker.threadId`

jerome-benoit: Indeed, and it must be a proper getter.

Electroid (approved):
Thanks for working on this!

PR #6525, opened 2023-10-15
Made `bun init` completely a drop-in replacement for `npm init`.

tr1ckydev:
### What does this PR do?

Adds all missing fields for `bun init` that are available in `npm init`, making it now a completely drop-in replacement for `npm init`.

- [x] Code changes

Screenshot_20231016_025649

Screenshot_20231016_025522


Jarred-Sumner:
Thank you for this.  I'm not really sure all the fields `npm init` asks for are worth having

tr1ckydev:
Whenever I had to create a bun package, I always had to use npm init to generate the template with the git repository and then run bun init again to setup bun-types and tsconfig. This resolves that.

If it's too much, we can also have a flag which gives us all the fields to setup instead of making it a default behavior. What do you think?

paperclover:
\- i dont think `test command` is worth having, ideally for bun it is always `bun test`
\- i think the default license should be MIT as it's the most common
\- i wonder if we can detect an existing git repository and set the repository field to the `origin` remote if so
\- we should make typescript a dev dependency not a peer dependency, but maybe i am wrong about this.

tr1ckydev:
> \- i dont think `test command` is worth having, ideally for bun it is always `bun test`

Then we can put something else there, just for the sake that the scripts object is created with a default value, makes it easier for the user to just edit it.

> \- i think the default license should be MIT as it's the most common

Agree. Will change that.

> \- i wonder if we can detect an existing git repository and set the repository field to the `origin` remote if so

We can through spawning a git command in the repository folder to get the origin url. Not sure if there are better ways to do that.

> \- we should make typescript a dev dependency not a peer dependency, but maybe i am wrong about this.

That's been a default behavior always. It would be better if addressed in a separate PR.

Issue #6492, opened 2023-10-13
toStrictEqual differs in jest and bun


pablo-mayrgundter:
### What version of Bun is running?

1.0.6

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

describe('parseUrl', () => {
  it('bun bug', () => {
    const urlStr = 'http://google.com'
    const url1 = new URL(urlStr)
    const url2 = new URL(urlStr)
    expect(url1).toEqual(url2)
    expect(url1).toStrictEqual(url2)
  })
})

### What is the expected behavior?

jest passes for both cases

### What do you see instead?

bun fails strict check and probably shouldn't.  output

```
src/urls.test.js:
 9 |   it('bun bug', () => {
10 |     const urlStr = 'http://google.com'
11 |     const url1 = new URL(urlStr)
12 |     const url2 = new URL(urlStr)
13 |     expect(url1).toEqual(url2)
14 |     expect(url1).toStrictEqual(url2)
        ^
error: expect(received).toStrictEqual(expected)

  {}

- Expected  - 0
+ Received  + 0

      at .../src/urls.test.js:14:4
```

PR #4105, opened 2023-08-10
Fixes #4089

Jarred-Sumner:
### What does this PR do?

Fixes #4089 

### How did you verify your code works?

Automated tests

Issue #4089, opened 2023-08-09
`deepEquals` returns `true` for different URLs.


tr1ckydev:
### What version of Bun is running?

0.7.4

### What platform is your computer?

Linux 6.4.6-1-cachyos x86_64 unknown

### What steps can reproduce the bug?

```ts
console.log(Bun.deepEquals(new URL("https://bun.sh/"), new URL("https://deno.com/")));
```

### What is the expected behavior?

console should log `false`

### What do you see instead?

console logs `true`

paperclover:
so...
<img width="673" alt="image" src="https://github.com/oven-sh/bun/assets/24465214/2911918d-fd9b-4dee-908a-ac6c6c00e747">

tr1ckydev:
I see, but it also shouldn't happen right?

PR #6528, opened 2023-10-15
fix(jest): fix toStrictEqual on same URLs

JoaoAlisson:
### What does this PR do?

This fix toStrictEqual on compare two same URLs.

This line on **Bun__deepEquals**, always return false if the URLs are equal on strict.
`return url1->wrapped().href() != url2->wrapped().href();`

On toStrictEqual should continue checking additional properties like toEqual.

- [X] Code changes

Fixes: #6492 
Correlated issue: https://github.com/oven-sh/bun/issues/4089
Correlated fix: https://github.com/oven-sh/bun/pull/4105


Electroid:
Thanks for your contribution!

Issue #5315, opened 2023-09-13
bun incorrectly reports "port in use" inside docker container


sartak:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

inside container: `Linux 5.4.0-144-generic x86_64 unknown`, on host: `Linux 5.4.0-144-generic x86_64 x86_64`

### What steps can reproduce the bug?

1. `sudo docker run -it debian:sid-slim`
    * (Everything that follows is inside the container)
2. `apt update && apt install -y curl unzip`
3. `curl -fsSL https://bun.sh/install | bash`
4. `source /root/.bashrc`
5. `bunx detect-port-alt 3000`
6. `apt install -y npm`
7. `npx -y detect-port-alt 3000`

### What is the expected behavior?

`bunx detect-port-alt 3000`, like `npx detect-port-alt 3000`, correctly reports that port 3000 is available.

### What do you see instead?

```
root@d08a0ebb9212:/# bunx detect-port-alt 3000
port 3000 was occupied
get available port 52021
```

```
root@d08a0ebb9212:/# npx -y detect-port-alt 3000
get available port 3000
```

### Additional information

This is blocking me from running a vanilla `create-react-app` project using bun.

It's important to try step 5 above (`bunx detect-port-alt 3000`) before step 6 (installing node) because if node is available, it will be used instead of bun (giving misleading results):

```
root@d08a0ebb9212:/# echo '#!/bin/echo' > `which node`
root@d08a0ebb9212:/# bunx detect-port-alt 3000
/usr/bin/node /tmp/detect-port-alt@latest--bunx/node_modules/.bin/detect 3000
```

sirenkovladd:
https://github.com/node-modules/detect-port/blob/master/lib/detect-port.js#L114

duplicate of https://github.com/oven-sh/bun/issues/4341

sirenkovladd:
less repro

index.js
```js
const net = require('net');

const server = new net.Server();

server.on('error', err => {
  console.log(err);
  server.close();
});

server.listen(3000, 'localhost', () => {
  port = server.address();
  server.close();
  console.log(port);
});
```

```sh
docker run -it -v ./index.js:/tmp/index.js -w /tmp debian:sid-slim
apt update && apt install -y curl unzip
curl -fsSL https://bun.sh/install | bash
~/.bun/bin/bun index.js
```
```
[...]

PR #6533, opened 2023-10-16
fix(serve): When IPv6 is not enabled, attempt to bind to IPv4 address under the same hostname.

Hanaasagi:
### What does this PR do?

Close: #5315. 

Please take a look at this https://github.com/oven-sh/bun/issues/5315#issuecomment-1719291283 first. About the IPv6 hosts, see the discussion in https://github.com/moby/moby/issues/35954. 

Previously, uWebSockets would iterate the results of `getaddrinfo` in order of IPv6 and IPv4 priorities, and if it failed, it would return an error immediately. This PR will attempt to bind to the IPv4 address under the same hostname when IPv6 fails (if the hostname is bound to both IPv6 and IPv4).

- [x] Code changes

manual testing.

#### Test step 

This issue can also be reproduced on the host machine.

- index.js

```JavaScript
const net = require("net");

const server = new net.Server();

server.on("listening", () => {
  console.log("listen success");
  server.close();
});

server.on("error", err => {
  console.log(err);
  server.close();
});

server.listen(3000, "localhost");
```

Add following lines to your `/etc/hosts`.

```
127.0.0.1 localhost
::1     localhost ip6-localhost ip6-loopback
```

Then disable ipv6 on loopback interface.

```Bash
sudo sysctl net.ipv6.conf.lo.disable_ipv6=1
```

Run `index.js`, Bun 1.0.6 will print 

```
error: Failed to listen at localhost
 code: "EADDRNOTAVAIL"
```

Use this pr's patch will print "listen success".


packages/bun-usockets/src/bsd.c
```diff
@@ -446,24 +446,49 @@ LIBUS_SOCKET_DESCRIPTOR bsd_create_listen_socket(const char *host, int port, int
...
 
-    if (listenFd == LIBUS_SOCKET_ERROR) {
-        freeaddrinfo(result);
-        return LIBUS_SOCKET_ERROR;
-    }
+    freeaddrinfo(result);
+    return LIBUS_SOCKET_ERROR;
+}
+
+inline LIBUS_SOCKET_DESCRIPTOR _bsd_create_listen_socket(
```
dylan-conway: this function needs to be declared before `bsd_create_listen_socket`

packages/bun-usockets/src/bsd.c
```diff
@@ -446,24 +446,49 @@ LIBUS_SOCKET_DESCRIPTOR bsd_create_listen_socket(const char *host, int port, int
 
     LIBUS_SOCKET_DESCRIPTOR listenFd = LIBUS_SOCKET_ERROR;
     struct addrinfo *listenAddr;
-    for (struct addrinfo *a = result; a && listenFd == LIBUS_SOCKET_ERROR; a = a->ai_next) {
+    for (struct addrinfo *a = result; a != NULL; a = a->ai_next) {
         if (a->ai_family == AF_INET6) {
             listenFd = bsd_create_socket(a->ai_family, a->ai_socktype, a->ai_protocol);
             listenAddr = a;
+
+            if (_bsd_create_listen_socket(listenFd, listenAddr, port, options) != LIBUS_SOCKET_ERROR) {
```
dylan-conway: should we also have a check for `listenFd != LIBUS_SOCKET_ERROR`?

dylan-conway: before calling the function?

packages/bun-usockets/src/bsd.c
Hanaasagi: Fixed.

packages/bun-usockets/src/bsd.c
Hanaasagi: Yes, fixed.

packages/bun-usockets/src/bsd.c
```diff
@@ -446,61 +488,47 @@ LIBUS_SOCKET_DESCRIPTOR bsd_create_listen_socket(const char *host, int port, int
...
     LIBUS_SOCKET_DESCRIPTOR listenFd = LIBUS_SOCKET_ERROR;
     struct addrinfo *listenAddr;
-    for (struct addrinfo *a = result; a && listenFd == LIBUS_SOCKET_ERROR; a = a->ai_next) {
+    for (struct addrinfo *a = result; a != NULL; a = a->ai_next) {
         if (a->ai_family == AF_INET6) {
             listenFd = bsd_create_socket(a->ai_family, a->ai_socktype, a->ai_protocol);
             listenAddr = a;
+
+            if (listenFd == LIBUS_SOCKET_ERROR) {
+                break;
```
dylan-conway: I think this break and lines 505, 517, and 525 should be `continue` so we can try more entries in `result`

dylan-conway:
LGTM after last comment is addressed 👍

packages/bun-usockets/src/bsd.c
Hanaasagi: done

dylan-conway:
We are setting up ci to run tests for this pr, will merge once we get it working

dylan-conway:
Thanks!

PR #6538, opened 2023-10-16
Development -> Contributing

colinhacks:


paperclover:
can you make the old page redirect?

also, edit the makefile help message, and any other references we have to this page link.

paperclover:
oh it was merged already

PR #6550, opened 2023-10-17
perf(bun-types): remove needless some call

Connormiha:
### What does this PR do?

I just simplified the filter for `index.d.ts` a little.

- [x] Code changes


Electroid:
Thanks!

PR #6551, opened 2023-10-17
fix: set user agent header

thomas9911:
### What does this PR do?

This add the agent header that some repos expect to be there

- [x] Code changes

I tried to build bun locally however I couldnt get it to build (I read the getting started)

Is this something that you guys want? (if so I can look further into adding tests ect)


Electroid:
Which repositories expect a `User-Agent` to be present? Do they cause an error?

thomas9911:
In our project we use https://www.npmjs.com/package/docxtemplater which just install normally via npm. However you can buy separate plugins these are hosted on their own repo. If you install it via their repo with npm/yarn/pnpm it works because those set the user-agent header to `npm/?`. Bun install does not do this and then the endpoint returns an error.

Sadly I cant share the url because it contains the api-key in it :/ (I also didnt find another project that has this restriction)

PR #6558, opened 2023-10-17
docs: fix ws.publish

aralroca:
In this example there is no server variable in the context, and here it makes more sense to use ws.publish. It is explained below that once the serve is done, the server.publish can be used.

### What does this PR do?

- [ ] Code changes


Electroid:
Thanks!

PR #6564, opened 2023-10-18
Fix websocket upgrade

Electroid:
### What does this PR do?

This fixes an issue where the `port` in the `Host` header was not defined on WebSocket upgrade. This also removes 2 cache headers that are not needed.

- [x] Code changes


src/http/websocket_http_client.zig
```diff
@@ -86,22 +89,25 @@ fn buildRequestBody(
         host_.deinit();
     }
 
+    const host_fmt = ZigURL.HostFormatter{
+        .is_https = is_https,
+        .host = host_.slice(),
```
dylan-conway, line 94: I would change this and line 110 to use `byteSlice()` to handle the utf16 case to be safe

src/http/websocket_http_client.zig
```diff
@@ -86,22 +89,25 @@ fn buildRequestBody(
...
+        .port = port,
+    };
     const headers_ = static_headers[0 .. 1 + @as(usize, @intFromBool(client_protocol.len > 0))];
     const pico_headers = PicoHTTP.Headers{ .headers = headers_ };
 
     return try std.fmt.allocPrint(
         allocator,
         "GET {s} HTTP/1.1\r\n" ++
-            "Host: {s}\r\n" ++
-            "Pragma: no-cache\r\n" ++
```
Jarred-Sumner, line 96: why get rid of these?

Electroid: Other clients don't send them, they don't affect servers because the request is upgrade (non-cacheable), and it could conflict with user-set headers.

cirospaciari: > Other clients don't send them, they don't affect servers because the request is upgrade (non-cacheable), and it could conflict with user-set headers. 

Host header is a must for 1.1 but not for 1.0, we should keep at least Host

```
A client MUST include a Host header field in all HTTP/1.1 request messages
```
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

Electroid: `Host` is included? Check the full diff

cirospaciari: > `Host` is included? Check the full diff

Oh sorry is included in fact.

src/http/websocket_http_client.zig
```diff
@@ -86,22 +89,25 @@ fn buildRequestBody(
...
-            "Cache-Control: no-cache\r\n" ++
+            "Host: {any}\r\n" ++
             "Connection: Upgrade\r\n" ++
             "Upgrade: websocket\r\n" ++
             "Sec-WebSocket-Version: 13\r\n" ++
             "{s}" ++
             "{s}" ++
             "\r\n",
-        .{ pathname_.slice(), host_.slice(), pico_headers, extra_headers },
+        .{ pathname_.byteSlice(), host_fmt, pico_headers, extra_headers },
```
Jarred-Sumner: `byteSlice` is not correct. It should convert to UTF8 first. `.toSlice()` and then `.slice` on that.

Jarred-Sumner:
Thank you

Issue #6570, opened 2023-10-18
Concating buffers with the length of undefined (/null/etc) results in an empty buffer


Darker-Ink:
### What version of Bun is running?

1.0.6

### What platform is your computer?

Microsoft Windows NT 10.0.22621.0 x64

### What steps can reproduce the bug?

When you try to concat an array of buffers together with the length being undefined in bun, Bun returns a empty Buffer, while on node when you do the same thing it concats the buffers together.

```js
const buffers =
[
    Buffer.from([0, 1, 1, 0]),
    Buffer.from([1, 0, 0, 1]),
];

const concated = Buffer.concat(buffers, undefined);

const combinedBuffers = Buffer.from([0, 1, 1, 0, 1, 0, 0, 1]);

console.log(`Concated With Undefined`, concated);
console.log(`Combined Buffers`, combinedBuffers);
console.log(`Do they match?`, concated.equals(combinedBuffers) ? 'Yes' : 'No', concated.equals(combinedBuffers));
```

![image](https://github.com/oven-sh/bun/assets/45852757/6816f959-bd53-4791-8ccb-2bce7c2e3404)

### What is the expected behavior?

The buffers are all concated together

### What do you see instead?

Instead, Bun returns an empty buffer

### Additional information

Original SS ![image](https://github.com/oven-sh/bun/assets/45852757/df561483-8872-47f1-9054-3efabee139f9)

Issue #3639, opened 2023-07-15
Inconsistent behavior of the `totalLength` parameter in `Buffer.concat` compared to Node


Hanaasagi:
### What version of Bun is running?

0.6.14

### What platform is your computer?

Linux 6.4.3-arch1-1 x86_64

### What steps can reproduce the bug?

```JavaScript
const assert = require("node:assert");

const b1 = Buffer.from("hello");
const b2 = Buffer.from("world");

const res = Buffer.concat([b1, b2], 100);

assert.equal(res.length, 100);

```

### What is the expected behavior?

No exceptions occurred, same as in Node.js.

### What do you see instead?

```
AssertionError: 10 == 100
 code: "ERR_ASSERTION"

```

### Additional information

From: https://nodejs.org/api/buffer.html#static-method-bufferconcatlist-totallength

According to the documentation, the `totalLength` parameter should be the length of the resulting Buffer. If the combined length of the Buffers in the list exceeds `totalLength`, the result is truncated to `totalLength`. When the combined length is less than `totalLength`, it should be filled  with 0.

Hanaasagi:
https://github.com/oven-sh/bun/blob/e546e4064c5f22e60559f1fd544928fe3ffe260f/src/bun.js/bindings/JSBuffer.cpp#L744-L748

Maybe line 746 could directly override the `byteLength` with the `totalLength` argument.

```C++
byteLength = byteLengthValue.toTypedArrayIndex(lexicalGlobalObject, "totalLength must be a valid number"_s);
```

PR #6574, opened 2023-10-18
fix(node:buffer): fix the behavior of `totalLength` in `Buffer.concat`

Hanaasagi:
### What does this PR do?

Close: #6570
Close: #3639

- [x] Code changes

I wrote automated tests.

### DIFF
--- a/src/bun.js/bindings/JSBuffer.cpp
+++ b/src/bun.js/bindings/JSBuffer.cpp
@@ -743,9 +743,22 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
     }
 
     if (callFrame->argumentCount() > 1) {
-        auto byteLengthValue = callFrame->uncheckedArgument(1);
-        byteLength = std::min(byteLength, byteLengthValue.toTypedArrayIndex(lexicalGlobalObject, "totalLength must be a valid number"_s));
-        RETURN_IF_EXCEPTION(throwScope, {});
+        auto totalLengthValue = callFrame->uncheckedArgument(1);
+        if (!totalLengthValue.isUndefined()) {
+            if (UNLIKELY(!totalLengthValue.isNumber())) {
+                throwTypeError(lexicalGlobalObject, throwScope, "totalLength must be a valid number"_s);
+                return JSValue::encode(jsUndefined());
+            }
+
+            auto totalLength = totalLengthValue.toInt32(lexicalGlobalObject);
+            RETURN_IF_EXCEPTION(throwScope, {});
+
+            if (UNLIKELY(totalLength < 0)) {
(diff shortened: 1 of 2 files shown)


src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -743,9 +743,22 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
...
-        byteLength = std::min(byteLength, byteLengthValue.toTypedArrayIndex(lexicalGlobalObject, "totalLength must be a valid number"_s));
-        RETURN_IF_EXCEPTION(throwScope, {});
+        auto totalLengthValue = callFrame->uncheckedArgument(1);
+        if (!totalLengthValue.isUndefined()) {
+            if (UNLIKELY(!totalLengthValue.isNumber())) {
+                throwTypeError(lexicalGlobalObject, throwScope, "totalLength must be a valid number"_s);
+                return JSValue::encode(jsUndefined());
+            }
+
+            auto totalLength = totalLengthValue.toInt32(lexicalGlobalObject);
```
Jarred-Sumner: Int32? Are you sure it shouldn’t be an Int52? Since you might want a buffer larger than 2 GB?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -743,9 +743,22 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
     }
 
     if (callFrame->argumentCount() > 1) {
```
Jarred-Sumner: We could simplify this a little and say something like:

```
JSValue lengthValue = callframe->argument(1)
if (lengthValue.isUndefined()) {
  // …
```

IIRC, argument() will return undefined if it’s outside the bounds of argumentCount()

src/bun.js/bindings/JSBuffer.cpp
Jarred-Sumner: Note that toTypedArrayIndex returns a size_t

Hanaasagi: 😱 fixed.

Hanaasagi: This file uses int32 as the type for variable such as `length`, `offset` in many places.

example:

```JavaScript
console.log(Buffer.alloc(4294967296).length);
// node: 4294967296
// bun: 0
```

https://github.com/oven-sh/bun/blob/2f10398c74b771bce738d636429135a679f52bdb/src/bun.js/bindings/JSBuffer.cpp#L422

Jarred-Sumner: ah we should fix tghat

Jarred-Sumner:
Thank you

PR #6579, opened 2023-10-18
fix: online docs moved

mountainash:
### What does this PR do?

Fixes a broken link.

https://bun.sh/docs/install/utilities was going to a 404 (a default Vercel styled 404), I've changed this to https://bun.sh/docs/cli/pm (note that the Help Docs are not complete - there's actually more commands like `bun pm migrate` which are in the Docs).

No, but it's a very simple change (see commit diff).


colinhacks:
Thanks, good catch. I added a redirect as well.

> not that the Help Docs are not complete

We're putting the final touches on [this PR](https://github.com/oven-sh/bun/pull/6395) which will flesh out the docs more

PR #6587, opened 2023-10-18
fix: change `--no-scripts` to `--ignore-scripts`

owlcode:
### What does this PR do?

Change `--no-scripts` to `--ignore-scripts` inside docs


Electroid (approved):
Thanks!

Issue #271, opened 2022-07-06
CA support for SSL connection to NPM registries


Trebh:
In a corporate environment this is often mandatory for connection to private NPM registries.

.npmrc config reference:
https://docs.npmjs.com/cli/v8/using-npm/config#ca

When acting as a NPM replacement, it should be possible to configure Bun with this option

bswan0002:
\+ cafile support please :)

viceice:
Bun should support OpenSSL `SSL_CERT_FILE` environment variables like NodeJS (with `--use-openssl-ca` option) to allow configuring custom root ca certificates via ENV

bswan0002:
> Bun should support OpenSSL `SSL_CERT_FILE` environment variables like NodeJS (with `--use-openssl-ca` option) to allow configuring custom root ca certificates via ENV

Nope, Bun PM not respecting those ENV variables when running `bun install`.

viceice:
I know, that's why I said it should. 😉

tutturen:
The lack of this option makes it impossible to start using bun in our organization, because we can't install packages from our private registry.

dengelma:
Are there any updates on this issue?

BeGj:
Can anyone try this workaround for bun install?

set env `NODE_TLS_REJECT_UNAUTHORIZED=0`. It got supported after [ this merge request ](https://github.com/oven-sh/bun/pull/4829) and it kind of works like a temporary workaround when using JOSE and fetching JWKS public certificate. 

Might not be good enough temp workaround for prod, but it enables you to do some internal testing at least.

agrison:
Waiting also for a solution to configure the `cafile` from within the `.bunfile.toml` config file, but in the meantime `NODE_TLS_REJECT_UNAUTHORIZED=0 bun install` is working fine.

admmasters:
> Waiting also for a solution to configure the `cafile` from within the `.bunfile.toml` config file, but in the meantime `NODE_TLS_REJECT_UNAUTHORIZED=0 bun install` is working fine.

This doesn't seem to be working for our private registry.

```
 package "@types/express" not found <private-artifactory-url>/artifactory/api/npm/@types%2fexpress 404
```

ryanc16:
This alone is a deal breaker for adoption in corporate/enterprise environments.

lirc571:
Related to #7124

[...]

PR #6588, opened 2023-10-18
Adds macOS Keychain certs to default CA store

ansg191:
### What does this PR do?

Adds support to `bun-usockets` to load trusted certificates from the native platform certificate store into the default CA store. This PR implements this for the macOS Keychain. Windows Certificate Stores could be added in the future.

Partially relevant: #271 

Note: All added code is usockets internal. Also change does not update `rootCertificates` in `internals/tls`.

Implementation based on [rustls-native-certs](https://github.com/rustls/rustls-native-certs/blob/main/src/macos.rs).

- [x] Code changes

- [x] Tested locally
	- Not sure how to since it's platform specific

### DIFF
--- a/packages/bun-usockets/src/crypto/openssl.c
+++ b/packages/bun-usockets/src/crypto/openssl.c
@@ -45,8 +45,97 @@ void *sni_find(void *sni, const char *hostname);
 
 #if defined(__APPLE__) && defined(__MACH__)
 
-#include <CoreFoundation/CoreFoundation.h>
-#include <Security/Security.h>
+#include <dlfcn.h>
+
+// CoreFoundation Types, Enums & Macros
+#define CFSTR(cStr)  cf_funcs.__CFStringMakeConstantString("" cStr "")
+
+typedef int32_t OSStatus;
+typedef unsigned char Boolean;
+typedef signed long CFIndex;
+typedef void *CFArrayRef;
+typedef void *CFMutableArrayRef;
+typedef void *CFDataRef;
+typedef void *CFStringRef;
+typedef void *CFDictionaryRef;
+typedef void *CFNumberRef;
+
+typedef enum : CFIndex {
+    kCFCompareLessThan = -1L,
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
I think this is a great idea, but I think we should dynamically load Core Foundation and the Security framework instead of passing it as a linker flag. Loading more frameworks at start time impacts start time. So we should instead use dlopen() and dlsym() to lazily load the symbols as needed.

ansg191:
@Jarred-Sumner Just pushed a commit to do that.

The Core Foundation & Security Frameworks will be dynamically loaded on the first request to `us_get_default_ca_store`.

packages/bun-usockets/src/crypto/openssl.c
```diff
@@ -647,6 +949,7 @@ X509_STORE* us_get_default_ca_store() {
     }
     
     us_internal_init_root_certs();
+    us_internal_init_native_certs();
```
Jarred-Sumner, line 952: If we do this, we should just use the native certs and not both right? So on macOS, it would only load native certs and on Linux, it would only load root certs. 

I think that we also should see if there's any perf difference in load time as I remember a 30ms perf hit a long time ago when loading native certs using a non-OpenSSL/BoringSSL library and loading from the system trust store.

ansg191: I just did some profiling, and the results are not good. The 3 `SecTrustSettingsCopyCertificates` requests to get all the certificates took 70ms total & the `SecTrustSettingsCopyTrustSettings` on each certificate to determine whether the system trusts the cert took 78ms total. As of 1.0.6, bun only takes 6ms to load certificates.

That's really disappointing. I was not expecting these APIs to be so slow.

I think a better approach would be to follow what Deno did in denoland/deno#11491. On all platforms, default to embedded Mozilla certs, and have some flag/env variable/config to enable the use of native certs for those who want/need it and are willing to take the first request performance hit.

Issue #5659, opened 2023-09-18
appears as if gc not freeing values created via node-api & napi-rs


universalmind303:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

originally opened in nodejs-polars, but seems to behave a bit differently in bun. 

```js
const pl = require("nodejs-polars");

function runBuggyCode(entries) {
  let df = pl.DataFrame(entries);
}

async function main() {
  const data = Array(50)
    .fill(null)
    .map((_, i) =>
      Array(10000)
        .fill(0)
        .map((_, ii) => i * 100 + ii)
    )

  let count = 0;
  let peakMemUsage = 0;
  while (count < 500) {
    const start = Date.now();
    runBuggyCode(data);
    console.log("process time:", Date.now() - start, "ms");
    peakMemUsage = Math.max(process.memoryUsage().rss, peakMemUsage);
    console.log(
      "peak rss mem usage:",
      Math.round(peakMemUsage / 1024 ** 2),
      "MB"
    );
    console.log("run:", count++);
    await new Promise((res) => setTimeout(res, 100));
  }
}


  void main().catch((err) => console.error(err));
```

### What is the expected behavior?

memory usage doesn't keep growing

### What do you see instead?

memory grows on every iteration

### Additional information

Nodejs: peak rss mem usage: 207 MB
Bun: peak rss mem usage: 2111 MB

mem stats
```
heap stats:    peak      total      freed    current       unit      count   
  reserved:   32.0 MiB   32.0 MiB      0       32.0 MiB                        
 committed:   32.0 MiB   32.0 MiB   20.0 MiB   11.9 MiB                        
     reset:      0          0          0          0                            ok
   touched:   64.2 KiB   64.2 KiB  458.6 KiB -394.3 KiB                        ok
  segments:      1          1          0          1                            not all freed!
-abandoned:      0          0          0          0                            ok
   -cached:      0          0          0          0                            ok
     pages:      0          0         22        -22                            ok
-abandoned:      0          0          0          0                            ok
 -extended:      0    
 -noretire:      0    
     mmaps:      0    
   commits:      0    
   threads:      0          0          0          0                            ok
  searches:     0.0 avg
numa nodes:       1
   elapsed:      59.317 s
[...]

PR #6598, opened 2023-10-19
fix(napi): incorrect refCount with napi_wrap()

alangecker:
## TLDR
updates `napi_wrap()` so that it it's refCount starts with 0 instead of 1 to match the behavior of node.js

## Details
while trying to get [`node-usb`](https://github.com/node-usb/node-usb) running with bun, it always failed because close() is only allowed when there are no open references.

https://github.com/node-usb/node-usb/blob/7e0182df8cfc684f6ec0cc59691c79731bdb459f/src/node_usb.h#L39-L41
```c++
    inline void ref(){ refs_ = Ref();}
    inline void unref(){ refs_ = Unref();}
    inline bool canClose(){return refs_ == 0;}
```

`Ref()` and `Unref()` are both called once, with node.js resulting in `refs_ == 0` (which is expected), but with bun `refs_ == 1`.

I've made this small script to reproduce the bug:
https://github.com/alangecker/bun-ref-bug/blob/main/binding.cc
```
run with bun 1.0.6:
 - refcount: 2 (expected: 1)
run with node 20.8.1:
 - refcount: 1 (expected: 1)
```

during a long debugging journey I found out, that buns `NapiRef::ref()` is also just called once (as expected), but within `napi_wrap()` the `NapiRef` gets initialized already with the refCount set to 1

https://github.com/oven-sh/bun/blob/378385ba60900be7bd797923c219ec489101f2f5/src/bun.js/bindings/napi.cpp#L669
After changing it to `new NapiRef(globalObject, 0)` it got the expected behavior / same as with node.js and node-usb works. as far as I understand it, a `NapiRef` with refCount=0 should then be weak instead of strong, which is why I have changed this too.

### What does this PR do?

- [x] Code changes

- I've ran the `make test-all` which fails with the same set of tests as without the change^^
- I've manually tested some modules using the napi API

## unclear to me
1. should the Reference really be changed to be weak?
2. does this change have any other implications / breaking something?

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -666,8 +666,8 @@ extern "C" napi_status napi_wrap(napi_env env,
 
     auto clientData = WebCore::clientData(vm);
 
-    auto* ref = new NapiRef(globalObject, 1);
-    ref->strongRef.set(globalObject->vm(), value.getObject());
+    auto* ref = new NapiRef(globalObject, 0);
+    ref->weakValueRef.setObject(value.getObject(), weakValueHandleOwner(), ref);
 
     if (finalize_cb) {
         ref->finalizer.finalize_cb = finalize_cb;


Jarred-Sumner:
This looks great, but I'm worried about crashes and our napi test coverage is rather terrible

Jarred-Sumner:
This seems to help a little bit with https://github.com/oven-sh/bun/issues/5659

`bun --smol` (main):
```
peak rss mem usage: 460 MB
run: 100
```

`bun-debug --smol` (this branch):

```
peak rss mem usage: 352 MB
run: 100
```

Jarred-Sumner:
@cirospaciari can you manually test in any napi projects that come to mind?

cirospaciari (approved):
I did not find any crashes after this change, and following the doc https://nodejs.org/api/n-api.html#napi_wrap
```
The optional returned reference is initially a weak reference, meaning it has a reference count of 0.
```

Tested Prisma, DuckDB and other packages that I found around searching for ` Napi::ObjectWrap<`

Jarred-Sumner:
thank you

PR #6607, opened 2023-10-19
fix(install): dont replace git urls when already present

liz3:
### What does this PR do?
When adding a git url as package which was already added, it would replace the existing with a faulty entry using a key with the url, this is because there was a step where both the entries would be present before resolving so the new entry did not have the actual name yet leading to preemtively adding it.
When the name then got resolved the logic matched but the key wasnt updated so it replaced the correct one.

This prevents the addititon to begin with by checking if the version matches any values.

- [x] Code changes

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -5221,6 +5221,18 @@ pub const PackageManager = struct {
                                         }
                                     }
                                     break;
+                                } else {
+                                    if (request.version.tag == .github or request.version.tag == .git) {
+                                        for (query.expr.data.e_object.properties.slice()) |item| {
+                                            if (item.value) |v| {
+                                                const url = request.version.literal.slice(request.version_buf);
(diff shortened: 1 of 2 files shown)


Electroid:
Mind adding a test for this? see `test/cli/install`

liz3:
@Electroid https://github.com/oven-sh/bun/pull/6607/commits/0fe0c85f93afa32b7c9d120b29a621e81b757033

src/install/install.zig
```diff
@@ -5221,6 +5221,19 @@ pub const PackageManager = struct {
                                         }
                                     }
                                     break;
+                                } else {
+                                    if (request.version.tag == .github or request.version.tag == .git) {
```
Jarred-Sumner, line 5225: do we need to do this for `gitlab` and any other cases?

Jarred-Sumner:
Thank you

PR #6620, opened 2023-10-20
Update quickstart.md - removed the part for editing `compilerOptions`…

PaulaBurgheleaGithub:
… in `tsconfig.json`

The line is already added to `compilerOptions` in `tsconfig.json` so there is no need to edit the file.

    "types": [
      "bun-types" // add Bun global
    ]

This was already added when the project initialized it seems.

### What does this PR do?

- [ ] Code changes

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -48,16 +48,6 @@ If you're using TypeScript, you may see a type error on the `Bun` global. To fix
 $ bun add -d bun-types
 ```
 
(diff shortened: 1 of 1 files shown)


colinhacks:
I added this relatively recently because *so* many people try Bun and don't know how to install types properly. Though you're right, this isn't necessary if you're following this guide as-is with `bun init`. I added this to a collapsible section instead. 
Thanks!

Issue #4569, opened 2023-09-08
[Jest] Support `%p` in test name formatting


Hanaasagi:
### What is the problem this feature would solve?

sometimes, we use `it.each` to run the same test suites with different data. When the data is a list of functions, I wish to be able to format them in the test name. Jest supports `%p` for formatting, but Bun does not support it currently.

Example code

```JavaScript
function A() { }

describe("test", () => {
  it.each([A])("test %p", async f => {
    f();
  });
});

```

Expect

```
✓ test > test [Function A] [0.07ms]
```

Bun

```
✓ test > test p [0.07ms]
```

### What is the feature you are proposing to solve the problem?

Use the `%p`. See https://jestjs.io/docs/api#1-describeeachtablename-fn-timeout

Electroid:
I implemented this, but it looks like your commit might be better.

PR #6627, opened 2023-10-21
feat(jest): add %p in label

Electroid:
Close: #4569

### What does this PR do?

- [x] Code changes


Hanaasagi:
Jest's `%p` outputs on one line, while the current `JestPrettyFormat` in Bun defaults to multi-line output. There's a difference here. At the time, I'm not sure about whether to add a new paramenter to `JestPrettyFormat` (e.g. `multiline = false`), so I didn't open this PR before.

Jarred-Sumner:
Added in https://github.com/oven-sh/bun/blob/caeea11706d899a6b6124d3ebec7eef214184492/src/bun.js/test/jest.zig#L1989

PR #6634, opened 2023-10-21
bun-types: Make expect parameter mandatory

Voldemat:
### What does this PR do?

Fixes type inference of expect function with undefined values.
```typescript
const a: string | undefined = "some string"
const expectedValue: string | undefined = "another thing"
expect(a).toBe(expectedValue) // Type 'undefined' is not assignable to type 'string'.ts(2345)
```
expect function will return type Expect<string>, but not Expect<string | undefined> as it should.

- [ ] Code changes


paperclover:
it needs to be optional because of `expect().fail()` which is a narrow use case but still, a use case...

Voldemat:
> it needs to be optional because of `expect().fail()` which is a narrow use case but still, a use case...

Is it used just to fail a test? If thats the case, wouldn\`t it be more reasonable to make a global function fail, which will do the same thing? Using the fail on expect doesn\`t make much sense, as we don\`t expect anything. We just wanna fail a test.

ghiscoding:
@Voldemat see jest-extended [`expect.fail(message)`](https://jest-extended.jestcommunity.dev/docs/matchers/fail)

Voldemat:
> @Voldemat see jest-extended [`expect.fail(message)`](https://jest-extended.jestcommunity.dev/docs/matchers/fail)

Yeah, I looked it up. But my previous point is still standing, it doesn\`t make much sense. I think this feature can be easily deprecated and eventually removed in favor of global "fail" function.

mProjectsCode:
I think the underlying problem is a typescript bug. It incorrectly removes undefined from the inferred parameter.
I created a [typescript playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAECiAeAHApsALgHgCoD5oG8AoASDQHsAhZACmSVTQC5osBKZgNzIEsATAwiQC+hEYQBmAVwB26bmWnQ6KdNmgBeaDIDW0sgHdpOamHSSwIAPzM2zBCsy4BxAE7I0kl4unJ9ceqq41KwA3KKCUrJo8ooA5sjSAILBzBBoLtzSsdAAPlrSvMjimcj8RK7unooyhcU+vGFiygzU8UnBrAB05FTUAORoyGl9oYTN6K0JyaxdPTQ1RSW8oUA) that showcases this bug and i am going to open an issue on the TS repo.

mProjectsCode:
The TS people say that this is not a bug, but I think I found a fix for this. I will open a new PR soon.

PR #6651, opened 2023-10-22
fix(node:buffer): fix `Buffer.write` stuck

Hanaasagi:
### What does this PR do?
Run following code, bun will not exit.
```JavaScript
const buffer = Buffer.alloc(5);
buffer.write("\xF0\x9F\x98\xAD\xFF");
console.log(buffer);

// node.js output
// <Buffer c3 b0 c2 9f 00>
```

The `while` loop at line 2071 copies ASCII slice to buffer, and next byte is greater than `0x7f`. UTF-8 encoding requires two bytes for characters between 0x80 and 0xff. Line 2082 only checks `buf.len >= 2`. If `buf.len == 1` it will cause a dead loop.

https://github.com/oven-sh/bun/blob/4cdaabd4336866adb7aae9ab4cbb41ad4b4a8f2a/src/string_immutable.zig#L2071-L2089

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/string_immutable.zig
+++ b/src/string_immutable.zig
@@ -2079,12 +2079,16 @@ pub fn copyLatin1IntoUTF8StopOnNonASCII(buf_: []u8, comptime Type: type, latin1_
             }
         }
 
-        if (latin1.len > 0 and buf.len >= 2) {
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (approved):
Thank you

Issue #4223, opened 2023-08-20
extra method shown in console.log


ImLunaHey:
### What version of Bun is running?

0.7.4

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

### What steps can reproduce the bug?

```ts
class ImageSettings {
    a = 123;

    toJSON() {
        return { a: '456' };
    }
}

const imageSettings = new ImageSettings();
console.log(imageSettings);
console.log(JSON.stringify(imageSettings));
```

```bash
$ node src/poc.js
ImageSettings { a: 123 }
{"a":"456"}
```

```bash
$ bun run src/poc.js
[0.09ms] ".env"
{
  a: 123,
  toJSON: [Function: toJSON]
}
{"a":"456"}
```

### What is the expected behavior?

`toJSON: [Function: toJSON]` not to be logged.

### What do you see instead?

`toJSON: [Function: toJSON]` included in the logs.

ArslanYM:
Can I work on this?

HenryZhang-ZHY:
reference #2966

osallak:
Hello maintainers,

I'm interested in working on this issue. Since it's already open, could you please consider assigning it to me? I believe I can contribute to its resolution and would like to take it on.

Looking forward to your response.

Best regards,
Oussama Sallak

robobun:
Same bug as #1713: console.log / Bun.inspect prints properties found on the object's prototype chain (here the class's toJSON method, there the Object.create proto's data key) as if they were own properties, because the formatter's forEachProperty walk in bindings.cpp climbs the prototype chain. Both still reproduce on Bun 1.4.0 and the open fix PR #36595 lists both issues, so the work is being tracked in #1713.

Duplicate of #1713

Please add a 👍 or any extra details over there so it all stays in one place. If this is actually a different problem, say so here and we'll reopen.

PR #6654, opened 2023-10-22
fix(console) ignore class methods in console.log output

almmiko:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/4223


src/bun.js/bindings/exports.zig
```diff
@@ -1852,6 +1852,10 @@ pub const ZigConsoleClient = struct {
                     if (key.eqlComptime("constructor")) return;
                     if (key.eqlComptime("call")) return;
 
+                    if (value.jsType().isFunction()) {
```
Jarred-Sumner: What happens if it's a function defined on the instance but not on the prototype? 

```js
class FooBar {
  constructor() {
    this.abc = () => {};
  }
  method() {
    return 42;
  }
}
```

In node, this logs
```js
FooBar { abc: [Function (anonymous)] }
```

Jarred-Sumner (changes requested):
This change likely needs to be done in `forEachProperty` C++ binding and not in the Zig code as we do not have the information to know whether the properties are the prototype's properties or the instance.

image

almmiko:
Hey @Jarred-Sumner thanks for highlighting the possible place where the change should be made.

jdalton (approved):

Jarred-Sumner:
Closing due to bitrot

Issue #4767, opened 2023-09-10
The `bun create` with a local template doesn't work.


yskszk63:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 6.4.12-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

Run `bun create` with a local template.

### What is the expected behavior?

A project is created from a local template.

### What do you see instead?

```bash
$ cat /tmp/tmp.XpmD9AznVC/package.json
{ "name": "tmp.XpmD9AznVC" }
$ bun create /tmp/tmp.XpmD9AznVC
Fetching package.json... 

error: "/tmp/tmp.XpmD9AznVC" was not found. Here are templates you can use:

  bun create apollo-server tmp.XpmD9AznVC

  bun create bun-bakery tmp.XpmD9AznVC

  bun create discord-interactions tmp.XpmD9AznVC

  bun create elysia tmp.XpmD9AznVC

  bun create elysia-buchta tmp.XpmD9AznVC

  bun create hono tmp.XpmD9AznVC

  bun create kingworld tmp.XpmD9AznVC

  bun create react tmp.XpmD9AznVC

  bun create react-ssr tmp.XpmD9AznVC

  bun create stric tmp.XpmD9AznVC

  bun create svelte-kit tmp.XpmD9AznVC

  bun create websi tmp.XpmD9AznVC

$
```

KalZyyrk:
i have the same problem
i try with the the directory .bun-create and the environment variable BUN_CREATE_DIR
with the path, nothing work

hassanallaham:
i have the same problem:
### What version of Bun is running?
```
$ bun -v
1.0.1
```

### What platform is your computer?
Ubuntu 20.04.6 LTS 64-bit

### What steps can reproduce the bug?
I have the needed path $HOME/.bun-create in my file system and there is a template created according to the docs:
```
$ cat $HOME/.bun-create/my-bun-template/package.json
{
    "name": "my-bun-template",
    "version": "0.0.1",
    "main": "index.js",
    "dependencies": {
        "cross-env": "*",
        "express": "*",
        "morgan": "*"
    },
    "bun-create": {
        "preinstall": "echo 'Installing...'",
        "postinstall": [
            "echo 'Done!'",
            "echo 'Go ahead'"
        ],
        "start": "bun run echo 'Hello world!'"
    }
}

```
I have the needed env variable:
```
[...]

PR #6670, opened 2023-10-23
fix cli create from local template

pierre-cm:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/4767.

This PR moves the parsing of the 'create' command information ahead of the command execution. This change allows us to ascertain whether the command refers to a local folder template before deciding whether to execute the command using bunx or not. If the command references a local folder, bunx should not be used.

- [x] Code changes

- [x] JSValue used outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed


colinhacks:
tested locally, works great. thanks pierre!

PR #6672, opened 2023-10-24
fix(fetch)

cirospaciari:
### What does this PR do?

- [x] Code changes

I wrote automated tests


src/js/node/http.ts
```diff
@@ -446,8 +448,21 @@ class Server extends EventEmitter {
   }
 
   address() {
-    if (!this.#server) return null;
-    return this.#server.address;
+    const address = this.#server?.address;
```
Jarred-Sumner: was this intentional

test/js/node/http/node-http.test.ts
```diff
@@ -1079,3 +1081,38 @@ describe("server.address should be valid IP", () => {
     });
   });
 });
+
+it("should not accept untrusted certificates", async () => {
```
Jarred-Sumner, line 1085: can you add more tests and copy some of node's tests? they should cover non-node:http usage too. Can you also add tests for WebSocket cert verification?

src/js/node/http.ts
cirospaciari: was not intentional https://github.com/oven-sh/bun/commit/01a7f28a6bfc072434a9641784d97dd18882e761#diff-91e6e16a6a18c42ea259bb91cd801b3707bbbf4f8c8acb48cc3df2152c02d1bd sorry

liz3:
I think at some point in the future it would be good to check why we get a different error then node but this solves it. 👍

Jarred-Sumner:
those failing tests are interesting

test/js/node/tls/fetch-tls-cert.test.ts
```diff
@@ -0,0 +1,360 @@
...
+    server: {
+      key: server.key,
+      cert: server.cert,
+      ca: client.ca,
+      requestCert: true,
+    },
+  });
+});
+
+it.todo("complete cert chains sent to peer, but without requesting client's cert.", async () => {
```
cirospaciari, line 73: a lot of this is `todo` because we still need to add these options to `Bun.serve` and/or `fetch`

test/js/node/http/node-http.test.ts
cirospaciari: Added, waiting tests for run

packages/bun-usockets/src/crypto/openssl.c
```diff
@@ -44,1637 +44,1892 @@ void* sni_find(void* sni, const char* hostname);
...
-    return length;
+  loop_ssl_data->ssl_read_input_offset += length;
+  loop_ssl_data->ssl_read_input_length -= length;
+  return length;
 }
 
-struct us_internal_ssl_socket_t* ssl_on_open(struct us_internal_ssl_socket_t* s, int is_client, char* ip, int ip_length)
-{
+struct us_internal_ssl_socket_t *ssl_on_open(struct us_internal_ssl_socket_t *s,
+                                             int is_client, char *ip,
```
Jarred-Sumner, line 177: are these formatting choices intentional?

cirospaciari: was clang-format-16 auto formating not intentional

src/js/node/tls.js
```diff
@@ -30,15 +30,19 @@ function parseCertString() {
...
 }
 
+const rejectUnauthorizedDefault =
+  process.env.NODE_TLS_REJECT_UNAUTHORIZED !== "0" && process.env.NODE_TLS_REJECT_UNAUTHORIZED !== "false";
 function isValidTLSArray(obj) {
   if (typeof obj === "string" || isTypedArray(obj) || obj instanceof ArrayBuffer || obj instanceof Blob) return true;
   if (Array.isArray(obj)) {
     for (var i = 0; i < obj.length; i++) {
-      if (typeof obj !== "string" && !isTypedArray(obj) && !(obj instanceof ArrayBuffer) && !(obj instanceof Blob))
+      const item = obj[i];
```
Jarred-Sumner, line 39: where is the test that makes sure we don't make this mistake again?

packages/bun-uws/src/HttpContext.h
```diff
@@ -64,6 +64,33 @@ struct HttpContext {
 
     /* Init the HttpContext by registering libusockets event handlers */
     HttpContext<SSL> *init() {
+        
+        if(SSL) {
```
Jarred-Sumner, line 68: why wasn't this necessary before?

cirospaciari: Basically BoringSSL will try to auto handshake the TLS connect, but only if enough data was ended, normally HTTP requests have enough data, but the handshake rejection was not working properly before for uWS, this will add same behavior that we expect with fetch and tls minus the checkServerIdentity part, we can add it too if we customize more, but I would suggest doing this in another PR.

src/boringssl.zig
```diff
@@ -74,6 +77,136 @@ export fn OPENSSL_memory_get_size(ptr: ?*const anyopaque) usize {
...
+    // use the null-terminated size to return the string
+    const size = bun.len(bun.cast([*:0]u8, outIP));
+    return outIP[0..size];
+}
+
+pub fn checkX509ServerIdentity(
+    x509: *boring.X509,
+    hostname: []const u8,
+) bool {
+    // we check with native code if the cert is valid
```
Jarred-Sumner, line 127: ngl this code feels a little unsafe

src/js/node/tls.js
cirospaciari: some here https://github.com/oven-sh/bun/blob/05bc3b68bb5b3832674cd7b626da94a86d54d7bc/test/js/node/tls/node-tls-cert.test.ts#L164-L215

We already have tests that uses Buffer or string too, but I can provide additional tests.

src/bun.js/bindings/webcore/JSWebSocket.cpp
```diff
@@ -230,10 +230,25 @@ static inline JSC::EncodedJSValue constructJSWebSocket3(JSGlobalObject* lexicalG
                 RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
             }
         }
+
+        if (JSValue tlsOptionsValue = options->getIfPropertyExists(globalObject, PropertyName(Identifier::fromString(vm, "tls"_s)))) {
+            if (!tlsOptionsValue.isUndefinedOrNull() && tlsOptionsValue.isObject()) {
+                if (JSC::JSObject* tlsOptions = tlsOptionsValue.getObject()) {
+
+                    if (JSValue rejectUnauthorizedValue = tlsOptions->getIfPropertyExists(globalObject, PropertyName(Identifier::fromString(vm, "rejectUnauthorized"_s)))) {
```
Jarred-Sumner, line 238: we _technically_ need to check for exceptions here incase the developer does something insane like

```js
new WebSocket(url, {tls: { get rejectUnauthorized() { throw Symbol("123"); } })
```

src/http/websocket_http_client.zig
```diff
@@ -949,15 +980,29 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
 
         pub fn handleHandshake(this: *WebSocket, socket: Socket, success: i32, ssl_error: uws.us_bun_verify_error_t) void {
             JSC.markBinding(@src());
-            _ = socket;
-            _ = ssl_error;
-            JSC.markBinding(@src());
+            const authorized = if (success == 1) true else false;
```
Jarred-Sumner, line 983:
```suggestion
            const authorized = success == 1;
```

src/bun.js/webcore/response.zig
```diff
@@ -1280,6 +1280,73 @@ pub const Fetch = struct {
                     error.FailedToOpenSocket => bun.String.static("Was there a typo in the url or port?"),
                     error.TooManyRedirects => bun.String.static("The response redirected too many times. For more information, pass `verbose: true` in the second argument to fetch()"),
                     error.ConnectionRefused => bun.String.static("Unable to connect. Is the computer able to access the url?"),
+
+                    error.UNABLE_TO_GET_ISSUER_CERT => bun.String.static("unable to get issuer certificate"),
```
Jarred-Sumner, line 1284: I have a feeling this code and other parts of this file would be a lot simpler if `us_bun_verify_error_t` had some more methods defined on it.

Jarred-Sumner:
I think we should merge this, but I think the code is really messy which means it will be harder to maintain/understand in the future and more prone to bugs.

Issue #6669, opened 2023-10-23
hang / infinite loop setTimeout(0) in a loop


scottbessler:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

run this code:

```typescript

export type AsyncLoopable<T> = AsyncIterable<T | Promise<T>> | Iterable<T | Promise<T>>;
export type AsyncWrappable<T> = null | undefined | AsyncLoopable<T>;

function delay(delayMs: number): Promise<void> {
  return new Promise((resolve) => setTimeout(resolve, delayMs));
}

async function asyncToArray<T>(source: AsyncWrappable<T>) {
  if (!source) {
    return [];
  }
  const result: T[] = [];
  for await (const el of source) {
    result.push(el);
  }
  return result;
}

async function* asyncGenerator<T>(queue: T[], done: boolean[]): AsyncGenerator<T, void, unknown> {
  let index = 0;
  while (true) {
    if (queue.length > index) {
      console.log('yield queue', index);
      yield queue[index++];
    } else if (done[0]) {
      return;
    } else {
      await new Promise((resolve) => setTimeout(resolve, 0)); // Wait for new items in the queue
    }
  }
}

/** turn 1 asynciterable into 2 separately iterable clones */
function asyncFork<T>(source: AsyncWrappable<T>): [AsyncIterable<T>, AsyncIterable<T>] {
  if (!source) {
    return [asyncGenerator([], [true]), asyncGenerator([], [true])];
  }
  const queue: T[] = [];
  const done = [false];
  (async () => {
    for await (const item of source) {
      console.log('queue push', item);
      queue.push(item);
    }
    done[0] = true;
  })();
  return [asyncGenerator(queue, done), asyncGenerator(queue, done)];
}

async function* createAsyncIterable<T>(els: T[]) {
  for (const el of els) {
    await delay(1);
    yield el;
  }
}

const iter = createAsyncIterable([1, 2, 3]);
const [clone1, clone2] = asyncFork(iter);

(async () => {
  console.log('hi');
  console.log(await asyncToArray(clone2));
  console.log(await asyncToArray(clone1));
})();


```

### What is the expected behavior?

when I run this with swc-node (or run the transpiled output of tsc) I get:
```

> swc-node foo.ts
hi
queue push 1
yield queue 0
queue push 2
yield queue 1
queue push 3
[...]

Issue #6547, opened 2023-10-17
`setTimeout` causes hang


orlandoxu:
### What version of Bun is running?

1.0.6

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

```JavaScript
// Execute the following test code, the program will never print, it seems that the program is permanently blocked
// if create 10000, it cost 18ms，but if 100*10000，never print the cost time.
// nodejs cost 62ms AND 784ms

const { randomInt } = require("crypto");

console.time("setTimeout");

// start 100 * 10000 setTimeout
let num = 0;
for (let i = 0; i < 100 * 10000; i++) {
  setTimeout(() => {
    num = num + 1;
    // console.log('setTimeout');
  }, 50 + randomInt(2000));
}

// print the cost time
console.timeEnd("setTimeout");

```

### What is the expected behavior?

The program can run normally instead of getting stuck。

### What do you see instead?

The program is stuck

### Additional information

![image](https://github.com/oven-sh/bun/assets/8287723/6585a677-a4ca-4754-b060-4a2216baf726)
![image](https://github.com/oven-sh/bun/assets/8287723/27bcc225-05a4-40fc-9de5-3f4b841998de)

orlandoxu:
![image](https://github.com/oven-sh/bun/assets/8287723/4c54d516-d6da-4aad-a32a-ec4d852bc8c5)

NOTICE: safari cost 1132ms

Issue #6253, opened 2023-10-03
setTimeout blocking in while loop


BSN4:
### What version of Bun is running?

1.0.4+ea956e58e9bac03b30d56ff74070603080913023

### What platform is your computer?

M1 mac

### What steps can reproduce the bug?

```js
            while (true) {
               // blocking code

                await new Promise((resolve) => setTimeout(resolve, 0))
            }
```

### What is the expected behavior?

in node.js same code will be running non-blocking

### What do you see instead?

blocking i/o

### Additional information

as workaround `await new Promise((resolve) => setTimeout(resolve, 1)` works

jumoog:
#3333

Issue #3333, opened 2023-06-16
setImmediate / setTimeout with delay 0 seems blocking the main thread


andrasat:
### What version of Bun is running?

0.6.9

### What platform is your computer?

Linux 6.1.31-2-MANJARO x86_64 unknown

### What steps can reproduce the bug?

So ive been playing with setImmediate and setTimeout in this [repo](https://github.com/andrasat/test-async-js), and try make a simple server with endpoints:

- `/ping`
- `/blocking`
- `/not-blocking`

`/blocking` and `/not-blocking` has the same cpu intensive task, with difference that `/blocking` is sync, and `/not-blocking` is async.

Steps:
- run server
```bash
bun start:bun // bun

bun start:node // node
npm run start:node // node
```
- hit `/ping` every second with this command
```bash
while true; do date; curl 127.0.0.1:3030/ping; sleep 1; echo; done;
```
- in another terminal, hit `/not-blocking`
```bash
curl 127.0.0.1:3030/not-blocking
```

In node, they are working as expected, when `/not-blocking` is still running, i can still receive response from `/ping` endpoint. But in bun, `/not-blocking` is blocking and `/ping` response is returned after the task is finished.

### What is the expected behavior?

- hit `/not-blocking` first, then `/ping`
- `/ping` should return `{ "data": "PONG" }` immediately without waiting for `/not-blocking` to finish

### What do you see instead?

- hit `/not-blocking` first, then `/ping`
- `/ping` is waiting for `/not-blocking` finish before returning `{ "data": "PONG" }`

simylein:
For me this actually segfaults at the moment with 
```
curl: (52) Empty reply from server
``` 
and
```
BUN:GET:/ping:2023-07-15T18:21:44.126Z
BUN:GET:/not-blocking:2023-07-15T18:21:44.298Z
Segmentation fault: 11
```

I am running bun `0.6.14` on `Darwin 22.5.0 arm64 arm`

andrasat:
UPDATED: 
@simylein  I got segfaulted to in my m1 pro lol, `Darwin 22.5.0 arm64 arm`, and bun version `0.6.14`, even in `/blocking` endpoint, it seems there is some problem with crypto module?

-- Screenshot --
<img width="1450" alt="Screenshot 2023-07-16 at 18 17 05" src="https://github.com/oven-sh/bun/assets/13641073/0e00696d-eff5-478c-a23e-cc5a8442b6b9">

[...]

PR #6674, opened 2023-10-24
Fix `setTimeout(() => {}, 0)` and align `setImmediate` behavior with Node.js

Jarred-Sumner:
### What does this PR do?

This makes `setTimeout` have a minimum duration of `1`, which aligns the behavior with Node and browsers.

This also makes `setImmediate` behave more similarly to Node. There is now an `immediate_task_queue` and a `next_immediate_task_queue`. `setImmediate` enqueues to the `next_immediate_task_queue` and `immediate_task_queue` is swapped with `next_immediate_task_queue` before each tick.

Fixes https://github.com/oven-sh/bun/issues/6669
Fixes https://github.com/oven-sh/bun/issues/6547
Fixes https://github.com/oven-sh/bun/issues/6253
Fixes https://github.com/oven-sh/bun/issues/2410
Fixes https://github.com/oven-sh/bun/issues/3333

Updated some tests and added a couple more tests.

### DIFF
--- a/src/bun.js/api/bun.zig
+++ b/src/bun.js/api/bun.zig
@@ -3630,7 +3630,6 @@ pub const Timer = struct {
         var map = vm.timer.maps.get(kind);
 
         // setImmediate(foo)
-        // setTimeout(foo, 0)
         if (kind == .setTimeout and interval == 0) {
             var cb: CallbackJob = .{
                 .callback = JSC.Strong.create(callback, globalThis),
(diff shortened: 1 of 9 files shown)


src/bun.js/api/bun.zig
```diff
@@ -3693,6 +3692,31 @@ pub const Timer = struct {
...
+    ) callconv(.C) JSValue {
+        JSC.markBinding(@src());
+        const id = globalThis.bunVM().timer.last_id;
+        globalThis.bunVM().timer.last_id +%= 1;
+
+        const interval: i32 = 0;
+
+        const wrappedCallback = callback.withAsyncContextIfNeeded(globalThis);
+
+        Timer.set(id, globalThis, wrappedCallback, interval, arguments, false) catch
```
dylan-conway, line 3708: This will use the `setTimeout` map in `Timer.set`. Maybe we should change this to take the timer kind instead of a boolean

PR #6679, opened 2023-10-24
add cache and main

rohanmayya:
### What does this PR do?

Added remaining types to Require's interface, cache and main. 

Previously, in my own project, I could not do:
`delete require.cache[filePath];`

to clear my import cache, because cache was not a recognized type on require.

- [ ] Code changes


Electroid (approved):
thanks!

Issue #6597, opened 2023-10-19
SQLite Database.query.get returns wrong result for a join


frostburn:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Linux 6.2.0-34-generic x86_64 x86_64

### What steps can reproduce the bug?

```typescript
import {Database} from 'bun:sqlite';

const db = new Database();

db.query(
  'CREATE TABLE Users (Id INTEGER PRIMARY KEY, Name VARCHAR(255));'
).run();

db.query(
  'CREATE TABLE Cars (Id INTEGER PRIMARY KEY, Driver INTEGER, FOREIGN KEY (Driver) REFERENCES Users(Id))'
).run();

db.query('INSERT INTO Users (Id, Name) VALUES (1, "Alice");').run();
db.query('INSERT INTO Cars (Id, Driver) VALUES (1, 1);').run();

const car = db.query('SELECT * FROM Cars JOIN Users ON Driver=Users.Id').get();

console.log(car);

db.close();
```

### What is the expected behavior?

```bash
{
  Id: 1,
  Driver: 1,
  "Users.Id": 1,
  "Users.Name": "Alice"
}
```

### What do you see instead?

```bash
{
  Id: 1,
  Driver: 1,
  Name: 1
}
```

### Additional information

To be specific, the issue is that the `Name` attribute gets populated from the wrong column. It doesn't really have to be called `"Users.Name"` as long as the right data is there.

matsumoto0912:
It seems that the issue still persists, hope this can be fixed.

jonnybrooks:
Yeah JOINs currently don't work for me at all due to this bug 😢

Jarred-Sumner:
Previously:
```js
{
  Id: 1,
  Driver: 1,
  Name: 1,
}
```

New in Bun v1.1.4: 
```js
{
  Driver: 1,
  Id: 1,
  Name: "Alice",
}
```

We do not include the duplicated column in the result, which matches better-sqlite3's behavior:
```js
❯ node sqlite3.cjs
{ Id: 1, Driver: 1, Name: 'Alice' }
```

PR #6690, opened 2023-10-24
fix(sqlite): fix column offset for join result

Hanaasagi:
### What does this PR do?

Close: #6597
Close: https://github.com/oven-sh/bun/issues/7147

When iterating the columns from `sqlite3_column_name`, we will get 

```
Id
Driver
Id
Name
```

It will contain duplicated column names But `PropertyNameArray` will not store them. The offset is mismatched.

|  `PropertyNameArray` | value from `sqlite3_column_value`  |
|---|---|
| Id  | 1 |
| Driver  | 1 |
| Name  | 1 |
|   | `Alice` |

This PR includes two kinds of fixes, I'm not sure which one to use.

- https://github.com/oven-sh/bun/commit/d56ed431b7c136c4e0ac2c3e48b8c8c1eda6389b  adds a bitmap `columnOffsets` to track the offsets corresponding to the names stored in `PropertyNameArray`. It will skip duplicated column name. If there are many columns, hashset might be more efficient. 

```
    Id: 1,
    Driver: 1,
    Name: "Alice",
```

- https://github.com/oven-sh/bun/commit/b77e6acea40dceeb5d236a94be5ac8ff56016591 just concat the table name and column name.

```
    Id: 1,
    Driver: 1,
    Users_Id: 1,
    Name: "Alice",
```

- [x] Code changes

I wrote automated tests.

### DIFF
--- a/src/bun.js/bindings/sqlite/JSSQLStatement.cpp
+++ b/src/bun.js/bindings/sqlite/JSSQLStatement.cpp
@@ -199,6 +199,8 @@ public:
     VersionSqlite3* version_db;
     uint64_t version;
     bool hasExecuted = false;
+    // default setting for SQLITE_MAX_COLUMN is 2000, Ref: https://sqlite.org/limits.html
+    uint16_t columnOffsets = 0;
     std::unique_ptr<PropertyNameArray> columnNames;
     mutable JSC::WriteBarrier<JSC::JSObject> _prototype;
     mutable JSC::WriteBarrier<JSC::Structure> _structure;
@@ -225,6 +227,7 @@ static void initializeColumnNames(JSC::JSGlobalObject* lexicalGlobalObject, JSSQ
             castedThis->columnNames->vm(),
             castedThis->columnNames->propertyNameMode(),
             castedThis->columnNames->privateSymbolMode()));
+        castedThis->columnOffsets = 0;
     }
     castedThis->update_version();
 
@@ -262,7 +265,15 @@ static void initializeColumnNames(JSC::JSGlobalObject* lexicalGlobalObject, JSSQ
                 break;
             }
 
-            columnNames->add(Identifier::fromString(vm, WTF::String::fromUTF8(name, len)));
+            int preCount = columnNames->size();
+            columnNames->add(
+                Identifier::fromString(vm, WTF::String::fromUTF8(name, len))
+            );
+            int curCount = columnNames->size();
+
+            if (preCount != curCount) {
+                castedThis->columnOffsets |= (1u << i);
+            }
         }
 
         if (LIKELY(!anyHoles)) {
@@ -283,6 +294,7 @@ static void initializeColumnNames(JSC::JSGlobalObject* lexicalGlobalObject, JSSQ
                 castedThis->columnNames->vm(),
                 castedThis->columnNames->propertyNameMode(),
                 castedThis->columnNames->privateSymbolMode()));
+            castedThis->columnOffsets = 0;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -265,7 +265,17 @@ static void initializeColumnNames(JSC::JSGlobalObject* lexicalGlobalObject, JSSQ
...
-            columnNames->add(Identifier::fromString(vm, WTF::String::fromUTF8(name, len)));
+            int preCount = columnNames->size();
+            columnNames->add(
+                Identifier::fromString(vm, WTF::String::fromUTF8(name, len))
+            );
+            int curCount = columnNames->size();
+
+            if (preCount == curCount) {
+                const char* tableName = sqlite3_column_table_name(stmt, i);
+                WTF::String fullName = WTF::String::fromUTF8(tableName, strlen(tableName)) + "_" + name;
```
Jarred-Sumner:
```suggestion
                WTF::String fullName = makeString(tableName, "_"_s, name);
```

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -265,7 +265,17 @@ static void initializeColumnNames(JSC::JSGlobalObject* lexicalGlobalObject, JSSQ
...
             }
 
-            columnNames->add(Identifier::fromString(vm, WTF::String::fromUTF8(name, len)));
+            int preCount = columnNames->size();
+            columnNames->add(
+                Identifier::fromString(vm, WTF::String::fromUTF8(name, len))
+            );
+            int curCount = columnNames->size();
+
+            if (preCount == curCount) {
```
Jarred-Sumner:
```suggestion
            // When joining multiple tables, the same column names can appear multiple times
            // columnNames de-dupes property names internally
            // We can't have two properties with the same name, so in that case we join the table name with the original name when there's a duplicate
            if (preCount == curCount) {
```

Hanaasagi:
Btw, better-sqlite3 returns the last value of duplicate fields in above case.

Test code
```JavaScript
const db = require("better-sqlite3")();

db.prepare(
  "CREATE TABLE Users (Id INTEGER PRIMARY KEY, Name VARCHAR(255), CreatedAt TIMESTAMP)",
).run();
db.prepare(
  "CREATE TABLE Cars (Id INTEGER PRIMARY KEY, Driver INTEGER, CreatedAt TIMESTAMP, FOREIGN KEY (Driver) REFERENCES Users(Id))",
).run();
db.prepare("INSERT INTO Users (Id, Name, CreatedAt) VALUES (?, ?, ?);").run(
  1,
  "Alice",
  "2022-01-01",
);
db.prepare("INSERT INTO Cars (Id, Driver, CreatedAt) VALUES (?, ?, ?);").run(
  2,
  1,
  "2023-01-01",
);

const car = db
  .prepare("SELECT * FROM Cars JOIN Users ON Driver=Users.Id")
  .get();

console.log(car);

db.close();

```

Output:

```
{ Id: 1, Driver: 1, CreatedAt: '2022-01-01', Name: 'Alice' }
// `CreatedAt` is from table `Users`, not `Cars`.
```


Should we align this behavior with that of better-sqlite3 ?

Jarred-Sumner:
I think using the last one is the most consistent with other database clients. I vaguely remember ActiveRecord being similar.

Hanaasagi:
Close https://github.com/oven-sh/bun/issues/7147

Hanaasagi:
@Jarred-Sumner  hi, any comments for this?

Hanaasagi:
no response for a long time, closed.

Issue #6583, opened 2023-10-18
Bun will not close socket if receive not understandable request in Bun.serve


rinsuki:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

Run some small http server:

```typescript
Bun.serve({
    fetch(req) {
        console.log("!")
        return new Response("hello")
    },
    port: 9999,
})
```

and access as a https server, like `curl https://localhost:9999`

### What is the expected behavior?

If received bytes don't looks like HTTP request, (send some error and)? close sockets.

for example, Node.js will close socket with `400 Bad Request` response.
![image](https://github.com/oven-sh/bun/assets/6533808/4d1a30ca-84f6-44bc-8396-4772f1396ba2)

### What do you see instead?

Bun just ignore and socket will hangs.

```
$ curl -v https://localhost:9999
*   Trying 127.0.0.1:9999...
* Connected to localhost (127.0.0.1) port 9999 (#0)
* ALPN: offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
^C
```

LukasKastern:
I tested this with my change from #8187. The request times out after ~10 seconds for me.
So it at least doesn't hang forever anymore. 

But the socket is still not immediately closed when sending an invalid header.

PR #6693, opened 2023-10-24
fix: don't hang when request is invalid

liz3:
UWS would not correctly handle a invalid request and hang because uws would not correctly report a failed header line following the http protocol.

This fixes that by returning a error if headers where not parsed correctly.

Fixes: https://github.com/oven-sh/bun/issues/6583

### What does this PR do?

- [x] Code changes


Jarred-Sumner:
What happens if you open a connection and then don't send any data for 1s? What about keep-alive connections which may keep the connection open for awhile but not send any data? I worry about not having sufficient test coverage to make a change like this right now (at least without running uWS' tests) 

I think a better & more limited change is to only do this when receiving an invalid header.  I know for example we don't immediately close the connection in that case and we probably should.

liz3:
@Jarred-Sumner this is on the parsing side though, if you dont send data you never reach that handler.
This does trigger only if the data is invalid from what i saw.

Jarred-Sumner:
I think this needs a lot more tests before we can comfortably move forward.

uNetworkingAB:
If you get a _valid_ request in two TCP chunks, this PR would break, as the first consume would consume 0. So essentially this PR completely breaks everything that is not received in 1 single TCP chunk, which is close to say that it completely breaks the server.

liz3:
> If you get a _valid_ request in two TCP chunks, this PR would break, as the first consume would consume 0. So essentially this PR completely breaks everything that is not received in 1 single TCP chunk, which is close to say that it completely breaks the server.

What's a suitable approach then to detect incorrect requests like a https request to a http server or any non HTTP 1.0/1.1 conforming tcp chunk/frame?

liz3:
@uNetworkingAB the `consumeRequestLine` method does correctly return 0(https://github.com/oven-sh/bun/blob/0e5d02a1fe1aa5ca5d56244014adfc4d5442db28/packages/bun-uws/src/HttpParser.h#L385), but the caller `fenceConsumePostPadded` doesn't interpret that as a error, maybe peeking into the requestline could be a better approach.

uNetworkingAB:
If you look at (upstream) uWS you can find HTTP_ERROR_505_HTTP_VERSION_NOT_SUPPORTED where consumeRequestLine is called. So upstream uWS will break on invalid request line but Bun's fork still just leaves it hanging for up to 10 seconds before it considers it closed. Both solutions are valid and secure, but upstream will fail quicker (and with reason)

uNetworkingAB:
Upstream has HttpErrors.h where you have immediate errors like so:

/* Possible errors from http parsing */
enum HttpError {
    HTTP_ERROR_505_HTTP_VERSION_NOT_SUPPORTED = 1,
    HTTP_ERROR_431_REQUEST_HEADER_FIELDS_TOO_LARGE = 2,
    HTTP_ERROR_400_BAD_REQUEST = 3,
    HTTP_ERROR_404_FILE_NOT_FOUND = 4
};

For all those errors, Bun will hang for 10 seconds while upstream will immediately return the proper status code and error

Jarred-Sumner:
> If you look at (upstream) uWS you can find HTTP_ERROR_505_HTTP_VERSION_NOT_SUPPORTED where consumeRequestLine is called. So upstream uWS will break on invalid request line but Bun's fork still just leaves it hanging for up to 10 seconds before it considers it closed. Both solutions are valid and secure, but upstream will fail quicker (and with reason)

Oh, I wonder if that change was when we made it allow HTTP 1.0  so that nginx default configuration would work. We should do what upstream does (though continue to allow HTTP 1.0 because configuring nginx is hard for a lot of people)

uNetworkingAB:
This is the release when uWS started being more explicit on error reporting: https://github.com/uNetworking/uWebSockets/releases/tag/v20.44.0 it's especially useful for NGINX since it will make it explicit that HTTP 1.0 is not supported

liz3:
I've ported the upstream changes to our version in this PR and also added a lot more tests.
Thanks again for the correction of the wrong implementation prior.

cc @Jarred-Sumner @uNetworkingAB

Jarred-Sumner:
- I don't think it makes sense for us to use the error pages
- We must continue to support HTTP/1.0 or at least not error in the vast majority of cases
- For invalid requests, we should add some callback that runs

liz3:
this does not remove 1.0 support.
> For invalid requests, we should add some callback that runs

i do agree with that but that a bigger scope to implement since we are within uws here but ye i think that would be good

liz3:
@Jarred-Sumner ive added a handler for invalid requests to serve, in https://github.com/oven-sh/bun/pull/6693/commits/ac02ce5497819ea2e4cfc463e56a31cb1da56b64
It does not allow the user to return a specific response but i don't think thats needed when the incoming request is damaged, also it would be hard because idk if theres a response in the scope and we are not in the request handler but server scope, but it does provide the broken request data we got so the user can debug it.

ive also added the new handler to the tests

uNetworkingAB:
Because you can't know what headers, URL or even treat the request as valid, at most I can see a use for lambdas that generate some response page like so:

app.setErrorHandler(function that takes error code and returns a response). Or even just call it app.error(f).

Those default handlers would just return the default error pages (modified for Bun). I don't see any reason for allowing anything more advanced than a simple callback like this. Possibly it could be returning some structured ErrorPage object instead of a raw string but that's really stretching the feature as far as I can think.

Allowing the full use of Response.writeHeader, Response.writeStatus, write, end, onWritable and so on could be more elegant but it really makes no sense since you are dealing with a broken "connection" that really cannot be considered working.

And since nothing really can be retrieved from the request (you can't know cookies, or anything user related) it makes no sense to support async handling because there is nothing to look up on a per user basis.

So TLDR it really just boild down to the ability to log the error, and to dynamically change the look of the response.

PR #6697, opened 2023-10-25
fix(runtime): remove env variable limit

### LINKED ISSUE #6694: Bun.env limited to 768?

v1.0.7

Linux 5.10.196-183.736.amzn2.x86_64

Load up a lot of environment variables. In my case, I'm running this in a oven/bun docker image inside of a large K8s cluster. This K8s cluster is populating lots of environment variables.

```sh
echo "console.log(Object.keys(Bun.env).length)" > test.ts
bun run test.ts
# -> 768
env | wc -l
# -> 952
```

I expect Bun.env to see the same env variables
I mainly noticed it because of the 4 env variables I was having k8s assign to my pod, only 1 of them was recognized inside Bun, but all 4 were in the `env | grep MY_ENV` command.

### What do you see instead?

Bun doesn't seem to load past 768 Env vars. Perhaps on of my variables is breaking the parsing somehow, but 768 does seem like a somewhat round number (3 * 256)

### Additional information

https://discord.com/channels/876711213126520882/1166504028771987556/1166504028771987556

paperclover:
### What does this PR do?

Previously, we had a 768 env var limit, only because it wasn't super clear how to get the data from zig to c++. Using pointers to the zig bytes, this is done using the minimal amount of pointer work and only one string copy per key.

Fixes #6694

### DIFF
--- a/src/bun.js/api/bun.zig
+++ b/src/bun.js/api/bun.zig
@@ -4414,8 +4414,16 @@ pub const FFIObject = struct {
 /// Also, you can't iterate over process.env normally since it only exists at build-time otherwise
 // This is aliased to Bun.env
 pub const EnvironmentVariables = struct {
-    pub export fn Bun__getEnvNames(globalObject: *JSC.JSGlobalObject, names: [*]ZigString, max: usize) usize {
-        return getEnvNames(globalObject, names[0..max]);
+    pub export fn Bun__getEnvCount(globalObject: *JSC.JSGlobalObject, ptr: *[*][]const u8) usize {
+        const bunVM = globalObject.bunVM();
+        ptr.* = bunVM.bundler.env.map.map.keys().ptr;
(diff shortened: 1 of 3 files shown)


src/bun.js/api/bun.zig
```diff
@@ -4414,8 +4414,16 @@ pub const FFIObject = struct {
 /// Also, you can't iterate over process.env normally since it only exists at build-time otherwise
 // This is aliased to Bun.env
 pub const EnvironmentVariables = struct {
-    pub export fn Bun__getEnvNames(globalObject: *JSC.JSGlobalObject, names: [*]ZigString, max: usize) usize {
-        return getEnvNames(globalObject, names[0..max]);
```
dylan-conway, line 4418: `getEnvNames` function should be deleted

dylan-conway (approved):
LGTM

PR #6702, opened 2023-10-25
Update Drizzle docs

dankochetov:
### What does this PR do?

Update code examples in Drizzle guide

- [ ] Code changes

1. You shouldn't specify the `BunSQLiteDatabase` type directly on the db, it'll prevent the relational queries from working if they are configured. The type should be inferred from the `drizzle()` call.
2. `db.get/all/run/values` and `migrate` API is sync for `bun:sqlite`.
3. The query return type was specified incorrectly.


colinhacks:
Apologies, I'd made these fixes before seeing this PR. Thanks Dan!

PR #6703, opened 2023-10-25
fix drizzle guide

jasperkelder:
### What does this PR do?

- [ ] Code changes


colinhacks:
Good catch!

Issue #6713, opened 2023-10-25
Reading from `process.stdin` doesn't work (hangs forever)


acheronfail:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Linux 6.1.59-1-lts x86_64 unknown

### What steps can reproduce the bug?

`bun` hangs forever when reading `process.stdin` - and is completely unable to read `process.stdin` in some cases.

I setup a reproducible repository here: https://github.com/acheronfail/bun-stdin/

Here is what doesn't work:

```js
// NOTE: this doesn't work - it hangs forever
function readProcessStdin() {
  return new Promise((resolve, reject) => {
    const chunks = [];
    process.stdin.on('error', (err) => reject(err));
    process.stdin.on('close', () => resolve(Buffer.concat(chunks).toString('utf-8')));
    process.stdin.on('data', (chunk) => chunks.push(chunk));
  });
}
```

Here is a workaround that does appear to work:

```js
// NOTE: this does work
async function readBunStdin() {
  const chunks = [];
  for await (const chunk of Bun.stdin.stream()) {
    chunks.push(chunk);
  }

  return Buffer.concat(chunks).toString('utf-8');
}
```

### What is the expected behavior?

I would expect that `process.stdin` would behave the same as it does with `node`.

The fact that it doesn't means I can't just use `bun` as a replacement for `node`, since `bun`'s `process.stdin` implementation doesn't appear to work properly. :cry: 

### What do you see instead?

`process.stdin` hangs forever!

Using the repository to reproduce:

```bash
# this uses process.stdin:
$ node -e 'console.log("X".repeat(10_000))' | bun ./index.mjs process
reading stdin... # hangs forever here
```

acheronfail:
Oh! On further investigation, I think `bun` is not emitting the `close` event on `process.stdin`! It does however, emit the `end` event.

Compare these:

```js
// NOTE: this doesn't work - it hangs forever
function readProcessStdin() {
  return new Promise((resolve, reject) => {
    const chunks = [];
    process.stdin.on('error', (err) => reject(err));
    process.stdin.on('close', () => resolve(Buffer.concat(chunks).toString('utf-8')));
    process.stdin.on('data', (chunk) => chunks.push(chunk));
  });
}
```

```js
// NOTE: this WORKS
function readProcessStdin() {
  return new Promise((resolve, reject) => {
    const chunks = [];
[...]

PR #6720, opened 2023-10-26
fix: emit close on stdin stream end

liz3:
We where not emitting the `close` event when done reading stdin data. It took me way too long to find this but i kept searching in zig/cpp code of the stream implementations...

Fixes: https://github.com/oven-sh/bun/issues/6713

### What does this PR do?

- [x] Code changes


Jarred-Sumner:
does this align with what node does? does node emit `close` before or after `end`? are they emitted on the same tick? or is there a `nextTick` between them? is it possible to emit `close` between `end` (or vice versa) and is it possible for node to end up emitting the event twice (do we need to handle that?)

liz3:
@Jarred-Sumner 
image

Jarred-Sumner:
i suggest also testing with inquirer

liz3:
@Jarred-Sumner it was good you made me aware again because with this change `close` was being emited twice because of the internals `close` handler, meant making the logic more complex because we obviously don't want to emit `close` twice...
Node only sends `end` when the data is fully consumed, i tried getting `end` without close in node but didnt.
end/close are still emitted on the same tick but we only emit close once, through stream.destroy().

Il play around with inquirer.

Jarred-Sumner:
is destroy correct there? I thought destroy can only be done once. What happens if you try to read from stdin twice this way?

liz3:
> is destroy correct there? I thought destroy can only be done once. What happens if you try to read from stdin twice this way?

thats for what the boolean is though, il test what happens if you try to read again.
The destroy is responsible for actually emitting the `close` event.

liz3:
@Jarred-Sumner it just resolves to null. `process.stdin.closed` and `process.stdin.destroyed` are true
image

src/js/builtins/ProcessObjectInternals.ts
```diff
@@ -145,7 +147,11 @@ export function getStdinStream(fd) {
         }
       } else {
         stream.emit("end");
-        stream.pause();
+        if (!stream_destroyed) {
+          stream_destroyed = true;
+          stream.destroy();
+          unref();
+        }
```
jerome-benoit, line 154: Why don't you emit 'close' to avoid to duplicate code and add boolean variable?

liz3: because of the close handler below that.

Jarred-Sumner:
Thank you

src/js/builtins/ProcessObjectInternals.ts
jerome-benoit: Why destroy() on a stream is emitting the 'close' event? 
Semantically, destroying is orthogonal to emitting an event. The destroy() caller is responsible of emitting event accordingly given the context.

Issue #5803, opened 2023-09-20
pidfd_open(2) system call is not supported


garder500:
### What version of Bun is running?

1.0.2

### What platform is your computer?

Linux 4.4.0-19041-Microsoft x86_64 unknown

### What steps can reproduce the bug?

bun create next-app

### What is the expected behavior?

The project is succefuly created

### What do you see instead?

Aborting installation.
```
Unexpected error. Please report it as a bug:
 610 |   }
611 |   get stdio() {
612 |     return this.#stdioObject ??= this.#createStdioObject();
613 |   }
614 | 
615 |   spawn(options) {
                 ^
error: "pidfd_open(2)" system call is not supported by your Linux kernel
To fix this error, either:
- Upgrade your Linux kernel to a newer version (current: 4.4.0-wn frame descri)
- Ensure the seccomp filter allows "pidfd_open"
      at spawn (node:child_process:615:14)
      at node:child_process:2:40
      at spawn (/tmp/create-next-app@latest--bunx/node_modules/create-next-app/dist/index.js:8:83666)
      at /tmp/create-next-app@latest--bunx/node_modules/create-next-app/dist/index.js:74:33775
      at new Promise (:1:20)
      at install (/tmp/create-next-app@latest--bunx/node_modules/create-next-app/dist/index.js:74:33158)
      at /tmp/create-next-app@latest--bunx/node_modules/create-next-app/dist/index.js:74:38669
      at processTicksAndRejections (:55:76)
```

### Additional information

Version WSL : 1.2.5.0
Version du noyau : 5.15.90.1
Version WSLg : 1.0.51
Version MSRDC : 1.2.3770
Version direct3D : 1.608.2-61064218
Version de DXCore : 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp
Version de Windows : 10.0.19045.3448

Electroid:
As the error suggests, you'll want to see if you can upgrade your Linux Kernel. Since you're using WSL, you can run: `wsl --update`

niudai:
running into the same error when I'm using debian 10

arno-fukuda:
Same here with Debian 10.
Using Google Cloud Compute Engine and can't upgrade beyond 4.19

edit: Ended up having to change to a different image with Debian 12.

garder500:
> As the error suggests, you'll want to see if you can upgrade your Linux Kernel. Since you're using WSL, you can run: `wsl --update`

[...]

PR #6724, opened 2023-10-26
Fix running `Bun.spawn` on Vercel and GCP

Jarred-Sumner:
### What does this PR do?

`pidfd_open` is not implemented by gVisor and by older Linux kernels. This implements a workaround where we spawn a thread, enqueue concurrent tasks and call `waitpid` on each `Subprocess`'s `pid`.  This workaround is very loosely based on libuv's default behavior. 

Fixes https://github.com/oven-sh/bun/issues/5803

Need to manually test that it does indeed fix the issue, but it should

I tested manually + added a hidden environment variable to let us manually enable this for testing purposes

### DIFF
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1808,11 +1808,11 @@ pub const Subprocess = struct {
                         switch (this.watch()) {
                             .result => {},
                             .err => |err| {
-                                if (err.getErrno() == .SRCH) {
-                                    waitpid_result = PosixSpawn.waitpid(pid, if (sync) 0 else std.os.W.NOHANG);
-                                    continue;
-                                } else {
-                                    // @panic("This shouldn't happen");
+                                if (!WaiterThread.shouldUseWaiterThread()) {
+                                    if (err.getErrno() == .SRCH) {
+                                        waitpid_result = PosixSpawn.waitpid(pid, if (sync) 0 else std.os.W.NOHANG);
+                                        continue;
+                                    }
                                 }
                             },
                         }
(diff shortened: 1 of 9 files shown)


src/output.zig
```diff
@@ -476,11 +476,23 @@ pub fn scoped(comptime tag: @Type(.EnumLiteral), comptime disabled: bool) _log_f
             defer lock.unlock();
 
             if (Output.enable_ansi_colors_stderr) {
-                out.print(comptime prettyFmt("<r><d>[" ++ @tagName(tag) ++ "]<r> " ++ fmt, true), args) catch unreachable;
-                buffered_writer.flush() catch unreachable;
+                out.print(comptime prettyFmt("<r><d>[" ++ @tagName(tag) ++ "]<r> " ++ fmt, true), args) catch {
```
Jarred-Sumner, line 479: this was to silence debug logs from throwing "PipeError"

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2072,4 +2127,135 @@ pub const Subprocess = struct {
...
+        concurrent_queue: Queue = .{},
+        queue: std.ArrayList(*Subprocess) = std.ArrayList(*Subprocess).init(bun.default_allocator),
+        started: std.atomic.Atomic(u32) = std.atomic.Atomic(u32).init(0),
+
+        pub fn setShouldUseWaiterThread() void {
+            should_use_waiter_thread = true;
+        }
+
+        pub fn shouldUseWaiterThread() bool {
+            return @atomicLoad(bool, &should_use_waiter_thread, .Monotonic);
```
paperclover, line 2189: should be fine but why not use std.atomic and/or atomic store when setting

Issue #6668, opened 2023-10-23
Bun.serve `requestIP` not working as expected


itohatweb:
### What version of Bun is running?

1.0.7

### What platform is your computer?

Linux 6.5.4-76060504-generic x86_64 x86_64

### What steps can reproduce the bug?

1. Write a simple http server
```ts
const server = Bun.serve({
    hostname: "0.0.0.0",
    port: 3000,
    development: false,
    async fetch(request) {
        console.log("IP:", this.requestIP(request));
        return new Response("Ok!");
    },
});

console.log(`Listening on 0.0.0.0: ${server.port}`);
```
2. Send a request to the server, from a different device (eg. open it on phone through a browser)
3. You get a strange IP Address
```json
{
  "address": "4294967232.4294967208.68.57",
  "family": "IPv4",
  "port": 40174
}
```

### What is the expected behavior?

The IP address should look like `192.168.68.57`.

### What do you see instead?

The IP address looks like `4294967232.4294967208.68.57`.

### Additional information

I don't know much about Buns inner workings, but the wrong numbers look suspiciously close to the `u32` max (`4294967295`) so this error might be related to an overflow.

Edit:
I've now tested this on a server too, and there the first 2 numbers are correct, and the last 2 numbers have this overflow problem. (`90.100.4294967212.4294967245`)

Electroid:
Thanks for flagging this, we'll fix it.

PR #6725, opened 2023-10-26
fix(requestIP): fix type cast from binary to text format for IPv4.

Hanaasagi:
### What does this PR do?

Close: #6668

- [x] Code changes

Manual testing: Use another device to request the server and check the request IP.

test code:
```JavaScript
const server = Bun.serve({
    hostname: "0.0.0.0",
    port: 3000,
    development: false,
    async fetch(request) {
        console.log("IP:", this.requestIP(request));
        return new Response("Ok!");
    },
});

console.log(`Listening on 0.0.0.0: ${server.port}`);
```

### DIFF
--- a/src/deps/libuwsockets.cpp
+++ b/src/deps/libuwsockets.cpp
@@ -1492,7 +1492,6 @@ extern "C"
                    LIBUS_SOCKET_READABLE | LIBUS_SOCKET_WRITABLE);
   }
 
- 
   void uws_res_override_write_offset(int ssl, uws_res_t *res, uintmax_t offset)
   {
     if (ssl)
@@ -1504,7 +1503,7 @@ extern "C"
       uwsRes->setWriteOffset(offset); //TODO: when updated to master this will bechanged to overrideWriteOffset
     }
   }
-  
+
   void uws_res_cork(int ssl, uws_res_t *res, void *ctx,
                     void (*corker)(void *ctx))
   {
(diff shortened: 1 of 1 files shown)


paperclover:
wow, `static_cast<unsigned char>` is not what i expected.

Hanaasagi:
@paperdave  Sorry, I didn't understand. Should I change it to this?


```c++
      unsigned char *ub = (unsigned char *) b;
      length = snprintf(b, 64, "%u.%u.%u.%u", ub[0], ub[1], ub[2], ub[3]);

```

Just like what's done in `AsyncSocket::addressAsText`.

https://github.com/oven-sh/bun/blob/2b8aae05d1e95b85c7869f15567a29162106110c/packages/bun-uws/src/AsyncSocket.h#L219-L223

Or using `ares_inet_ntop AF_INET`?

cirospaciari:
> @paperdave Sorry, I didn't understand. Should I change it to this?
> 
> ```c++
>       unsigned char *ub = (unsigned char *) b;
>       length = snprintf(b, 64, "%u.%u.%u.%u", ub[0], ub[1], ub[2], ub[3]);
> ```
> 
> Just like what's done in `AsyncSocket::addressAsText`.
> 
> https://github.com/oven-sh/bun/blob/2b8aae05d1e95b85c7869f15567a29162106110c/packages/bun-uws/src/AsyncSocket.h#L219-L223
> 
> Or using `ares_inet_ntop AF_INET`?

Thinking in edge cases probably `ares_inet_ntop` is safer

Hanaasagi:
Fixed. But indeed, I think there isn't much difference. The root cause of this bug is that we need to convert `char` to `unsigned char` first and then format it as `%u`. `ares_inet_ntop` does it the same way.

https://github.com/c-ares/c-ares/blob/0e7a5dee0fbb04080750cf6eabbe89d8bae87faa/src/lib/inet_ntop.c#L81-L93

cirospaciari (approved):

cirospaciari:
Thank you

PR #6732, opened 2023-10-26
test(worker): refactor worker tests

jerome-benoit:
### What does this PR do?

+ constify variables
+ typing
+ untangle web worker from worker_threads tests
+ cleanup variables namespace
+ move worker related tests into workers directory

no functional changes on tests

ran impacted tests


jerome-benoit:
@Electroid: May I ask for your review on that PR? 
The repo setup does not allow me to ask for a review via the intended field. 

I really would like to have the worker tests refactoring one merged. I have other fixes to `worker_threads` code stacked on top of that PR that I would like to push upstream.  

I also have a question regarding design discussions: where do they usually happen? 
I need inputs by people having deep bun internals knowledge to tackle properly some of the issues encountered with `worker_threads`.

test/js/web/worker-fixture-env.js
```diff
@@ -1,12 +0,0 @@
-import worker_threads from "worker_threads";
-
-if (worker_threads.isMainThread) throw new Error("worker_threads.isMainThread is wrong");
```
paperclover, line 3: this check was removed

test/js/node/worker_threads/worker.js
```diff
@@ -1,6 +1,8 @@
-import { parentPort, workerData } from "worker_threads";
+import { isMainThread, parentPort, workerData } from "worker_threads";
 
-parentPort.on("message", m => {
+if (isMainThread) throw new Error("worker_threads.isMainThread is wrong");
+
+parentPort?.on("message", m => {
```
paperclover: `?.` -> `.`

we want an error to be thrown if this property is undefined.

paperclover (changes requested):
thanks for cleaning up the mess that is these worker tests.

two nitpicks i would like fixed before merging, see review comments.

also, if you would like to add this, could you add an extra test that checks if a `.tsx` file works. theyre supposed to but re-reading these tests there isnt any coverage of that. no worries if not.

test/js/web/worker-fixture-env.js
jerome-benoit: You mean the test has been removed upstream? 
I've used `git mv` twice on few files and dunno if it's a git bug or expected behavior but the link to the original file and its changes is broken for these files.
I will inspect upstream changes, address review comments then ask for review in a few days.

paperclover: it was deleted here where i commented and not re-added in `test/js/web/workers/worker-fixture-env.js` where it was moved.

test/js/node/worker_threads/worker.js
jerome-benoit: TS complains about 'parentPort can be null'. So I've added a null check + error throwing in `node:worker_threads` related workers.

test/js/web/worker-fixture-env.js
jerome-benoit: The check has been moved to `node:worker_threads` tests worker implementation because `isMainThread` is not part of the web worker API.

Tests entanglement between worker_threads and web worker is not complete yet.

I will do another PR to move worker_threads tests and workers code to node:worker_threads test directory.

jerome-benoit:
> also, if you would like to add this, could you add an extra test that checks if a `.tsx` file works. theyre supposed to but re-reading these tests there isnt any coverage of that. no worries if not.

As part of another PR then.

+ one PR to finish the worker_threads and web worker test code untanglement
+ one PR to add a test for `.tsx` worker file handling for web worker

test/js/node/worker_threads/worker.js
paperclover: ah i see

test/js/web/worker-fixture-env.js
paperclover: i guess this makes sense, but it is something bun allows, but i don't think we explicilty document this behavior anywhere.

paperclover (approved):
looks good. before i run the tests i am going to push a commit fixing the submodule. will merge these changes after tests run

PR #6739, opened 2023-10-27
[docs] fix ws/pubsub; closed ws don't publish to topic

### LINKED ISSUE #6691: `websocket` not publishing to topic after connection close.

1.0.7

Linux 5.10.102.1-microsoft-standard-WSL2 x86_64 x86_64

just create a simple websocket chat application as described in the documentation.
https://bun.sh/docs/api/websockets#pub-sub

```ts
close(ws) {
      const msg = `${ws.data.username} has left the chat`;
      ws.unsubscribe("the-group-chat");
      ws.publish("the-group-chat", msg);  // 🔴 this message should get published to the topic and other subscribers should get this
}
```

### What do you see instead?

The message doesn't get published. 😕
I ran some tests and also checked the return value of the publish method.
seems like it returns a number
  - (3+ message length) : when it successfully publishes the message and 
  - 0: if the message is not published.
 
Message is not published if there is only one user subscribing to the topic and it is the sending user himself, which is OK.
but I tried with 2 and 3 users connected, and that publish line always returned 0, and the message in the close callback never gets published, so no one got to know that some user disconnected from chat.
 

### Additional information

_No response_

RohitKaushal7:
### What does this PR do?

fix #6691 - fix docs in `close` callback of a websocket connection. replaced `ws.publish` with `server.publish`

- [ ] Code changes


Jarred-Sumner:
Thank you

PR #6745, opened 2023-10-27
docs: fix `setup.sh` path in `contributing.md`

RaisinTen:
### What does this PR do?

It was added in https://github.com/oven-sh/bun/pull/4410. The `src` directory is not a part of the file path.


paperclover:
thanks

do note i dont think that script works yet. from that pr changed alot of stuff and i havent fully tested repo setup on a fully fresh clone

PR #6746, opened 2023-10-27
fix: `ninja` invocation error in `scripts/setup.sh`

RaisinTen:
### What does this PR do?

Running the script ended with this error:

```sh
$ bash ./scripts/setup.sh
...
ninja: error: loading 'build.ninja': No such file or directory
```

It happens because the `build.ninja` file is inside the `build` directory, not the current directory, so `ninja` is not able to detect the presence of that file and run the builds.

The error started happening in https://github.com/oven-sh/bun/pull/4410, where this script was introduced.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -8,4 +8,4 @@ bun i
 
 mkdir build
 cmake -B build -S . -DCMAKE_BUILD_TYPE=Debug -G Ninja
-ninja
+ninja -C build


paperclover:
i dont think this fixes the entire setup script.

RaisinTen:
With this change, it does invoke ninja successfully but the ninja process fails later on with errors like:
```zig
src/runtime.zig:200:46: error: unable to open './runtime.out.js': FileNotFound
    pub const ProdSourceContent = @embedFile("./runtime.out.js");
                                             ^~~~~~~~~~~~~~~~~~
```
So I thought that this was a step in the right direction.

paperclover:
id rather merge a fully working script. this is something i'll get working and document tomorrow / the weekend.

i think you need to run `make runtime_js fallback_decoder bun_error node-fallbacks` but ideally we shouldn't rely on the old makefile, so i also dont want to just have the makefile stuff just thrown into this script.

RaisinTen:
Alright, sounds good!

PR #6752, opened 2023-10-27
dx: Fix setup script and contributing docs

paperclover:
This branch is two things

- Fixes the setup script for Mac+Linux
- Gets some of the CI for windows working. not done yet so it's not set as active yet

image


Jarred-Sumner:
I think the line removing -fPIC is probably a mistake. But otherwise it looks good

Jarred-Sumner:
Thank you

Issue #6704, opened 2023-10-25
`bun build` segfaults when using string interpolation in macro arguments.


Dev-Siri:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Linux 6.2.0-35-generic x86_64 x86_64

### What steps can reproduce the bug?

Create a file like `main.ts` and import a macro

```ts
import { macro } from "./macro" assert { type: "macro" };

const llo = "llo";

// This needs to be a template literal that interpolates a string
console.log(macro(`He${llo}`));
```

The macro is simple:

```ts
export function macro(text: string) {
  return text + " :D";
}
```

Then build it

```sh
$ bun build ./main.ts
```

### What is the expected behavior?

I'd expect this to just run the macro and keep the string interpolation as-is or maybe it will interpolate it in the argument before the macro even runs.
And if its not allowed, (because macros run at build time and string interpolation is done at runtime) then I'd expect to see a friendly error-message rather than it crashing with a segfault.

### What do you see instead?

```sh
[1]    11703 segmentation fault  bun build ./main.ts
```

Electroid:
Related #3830

Electroid:
Closed by #6756

Issue #6014, opened 2023-09-24
Segfault when using const value as argument for macro function.


tobilinz:
### What version of Bun is running?

1.0.3

### What platform is your computer?

Linux 6.5.3-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

1. Create a new bun project using `bun init -y`
2. Paste the following code into `index.ts`:
```
import { myMacro } from './other' with { type: 'macro' };
const test = 'Hello World ';
console.log(myMacro(test));
```
5. Paste the following code into a new file called `other.ts`:
```
export function myMacro(value: string): string {
  return value + "and Bun!";
}
```
7. Run `bun build ./index.ts`

### What is the expected behavior?

Console output:
```
// index.ts
console.log("Hello World and Bun!");
```

### What do you see instead?

Console output:
`[1]    27447 segmentation fault (core dumped)  bun build ./index.ts`

### Additional information

This bug also occurs when importing a const from another file  or importing  the content of a .txt* file and using it as input for a macro function.

* .wgsl files also cause this behavior. Other file types have not been tested.

tobilinz:
I accidentally closed it.

maxmilton:
Possibly related: https://github.com/oven-sh/bun/issues/3832 and https://github.com/oven-sh/bun/issues/3830#issuecomment-1664829699

Electroid:
Closed by #6756

maxmilton:
As mentioned in https://github.com/oven-sh/bun/issues/3830 the segfault is now correctly handled, however there's still an issue with handling certain values in macros. Not sure this should be closed yet.

Issue #3830, opened 2023-07-26
Macro still crashes with SIGSEGV when non-ASCII character in string template literal with interpolation


maxmilton:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Linux 6.4.6-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

1. Create file `macro.ts`:
```ts
export function fn(str) {
  return str;
}
```

2. Create file `index.test.ts`:
```ts
import { expect, test } from 'bun:test';
import { fn } from './macro' assert { type: 'macro' };

test('works', () => {
  expect(fn(`A${''}`)).toBe('A');
});

test('segfaults', () => {
  expect(fn(`©${''}`)).toBe('©');
});
```

3. Run `bun test`.

### What is the expected behavior?

Template literals are processed correctly as Unicode when passed as macro arguments.
Bun process runs without a segmentation fault.

### What do you see instead?

```
~/P/_/bun-macro-sigsegv ❱ bun test
bun test v0.7.0 (aa1ad7f0)

index.test.ts:
fish: Job 1, 'bun test' terminated by signal SIGSEGV (Address boundary error)
```

### Additional information

This is a continuation of https://github.com/oven-sh/bun/issues/3641

Now that bun v0.7.0 was released, it looks like the original bug has not been fixed; a bun macro with arg string template literal with non-ASCII characters + interpolation still crash with a segfault.

Same result in both bun stable `v0.7.0 (aa1ad7f0)` and canary `v0.7.1 (f3200ac0)`.

Jarred-Sumner:
This is a slightly different case. For the conversion to happen successfully, minification must be enabled so that the string gets simplified. The crash is happening due to incorrect error handling logic when throwing the error. Haven't narrowed down why yet.

maxmilton:
When experimenting with workarounds, I found this is also an issue with plain strings when there is string concatenation:

```ts
import { expect, test } from 'bun:test';
import { fn } from './macro' assert { type: 'macro' };

test('works 1', () => {
  expect(fn('A')).toBe('A');
});

test('works 2', () => {
  expect(fn('©')).toBe('©');
});

test('works 3', () => {
  expect(fn('A' + '')).toBe('A');
});

test('segfaults', () => {
  expect(fn('©' + '')).toBe('©');
});
```

This aligns with Jarred's note about the string needing to be simplified and still confirms it only happens for non-ASCII characters.

[...]

PR #6756, opened 2023-10-27
fix: Macro segmentation faults

I-A-S:
### What does this PR do?

- [x] Code changes

Fixes the occurrence of segmentation faults under certain macro inputs as indicated by issues #3830, #6014 and #6704. Bun will now display the relevant error instead of crashing with a segmentation fault.


src/js_ast.zig
```diff
@@ -7278,8 +7278,9 @@ pub const Macro = struct {
 
             exception_holder = Zig.ZigException.Holder.init();
             var js_args: []JSC.JSValue = &.{};
+            var js_processed_args_len: usize = 0;
             defer {
-                for (js_args[0 .. js_args.len - @as(usize, @intFromBool(!javascript_object.isEmpty()))]) |arg| {
+                for (js_args[0 .. js_processed_args_len - @as(usize, @intFromBool(!javascript_object.isEmpty()))]) |arg| {
```
Jarred-Sumner:
```suggestion
                for (js_args[0 .. js_processed_args_len -| @as(usize, @intFromBool(!javascript_object.isEmpty()))]) |arg| {
```

Jarred-Sumner:
A test would be great. I think we should merge this regardless, but it'd be easy for this to regress in the future without a test.

I-A-S:
> A test would be great. I think we should merge this regardless, but it'd be easy for this to regress in the future without a test.

Agreed I'll add a test.

I-A-S:
Added a test

Jarred-Sumner:
Thank you

PR #6764, opened 2023-10-28
build: move runtime js and fallback decoder builds to cmake

I-A-S:
### What does this PR do?

This PR move runtime js and fallback decoder builds from makefile to cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,10 +27,12 @@ endif()
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
     set(DEBUG ON)
     set(ZIG_OPTIMIZE "Debug")
+    set(NODE_ENV "development")
     set(bun "bun-debug")
 elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
     set(DEBUG OFF)
     set(ZIG_OPTIMIZE "ReleaseFast")
+    set(NODE_ENV "production")
     set(bun "bun-profile")
 endif()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -651,14 +653,59 @@ list(APPEND BUN_RAW_SOURCES "${BUN_WORKDIR}/codegen/WebCoreJSBuiltins.cpp")
...
+    COMMAND "${ESBUILD}" "--define:process.env.NODE_ENV=\"${NODE_ENV}\"" "--target=esnext" "--bundle" "--format=iife" "--platform=browser" "--global-name=BUN_RUNTIME" "--minify" "--external:/bun:*" "${BUN_SRC}/runtime/index.ts" "--outfile=${BUN_SRC}/runtime.out.js"
+    COMMAND cat ${BUN_SRC}/runtime.footer.js >> ${BUN_SRC}/runtime.out.js
+)
+add_custom_command(
+    OUTPUT "${BUN_SRC}/runtime.out.refresh.js"
+    COMMAND "${ESBUILD}" "--define:process.env.NODE_ENV=\"${NODE_ENV}\"" "--target=esnext" "--bundle" "--format=iife" "--platform=browser" "--global-name=BUN_RUNTIME" "--minify" "--external:/bun:*" "${BUN_SRC}/runtime/index-with-refresh.ts" "--outfile=${BUN_SRC}/runtime.out.refresh.js"
+    COMMAND cat ${BUN_SRC}/runtime.footer.with-refresh.js >> ${BUN_SRC}/runtime.out.refresh.js
+)
+add_custom_command(
+    OUTPUT "${BUN_SRC}/runtime.nodebun.pre.out.js"
```
paperclover, line 670: is this right?

CMakeLists.txt
```diff
@@ -651,14 +653,59 @@ list(APPEND BUN_RAW_SOURCES "${BUN_WORKDIR}/codegen/WebCoreJSBuiltins.cpp")
...
 # --- Runtime.js ---
-if(NOT NO_CODEGEN)
-    add_custom_command(
-        OUTPUT "src/fallback.out.js"
-        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
-        COMMAND "${ESBUILD}" "--target=esnext" "--bundle" "src/fallback.ts" "--format=iife" "--platform=browser" "--minify" "--outfile=src/fallback.out.js"
-        DEPENDS "src/fallback.ts"
-    )
-endif()
+#if(NOT NO_CODEGEN)
```
paperclover: remove the commented code since no longer in use

CMakeLists.txt
```diff
@@ -651,14 +653,59 @@ list(APPEND BUN_RAW_SOURCES "${BUN_WORKDIR}/codegen/WebCoreJSBuiltins.cpp")
...
+#        WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+#        COMMAND "${ESBUILD}" "--target=esnext" "--bundle" "src/fallback.ts" "--format=iife" "--platform=browser" "--minify" "--outfile=src/fallback.out.js"
+#        DEPENDS "src/fallback.ts"
+#    )
+#endif()
+
+# --- Runtime JS ---
+add_custom_command(
+    OUTPUT "${BUN_SRC}/runtime.out.js"
+    COMMAND "${ESBUILD}" "--define:process.env.NODE_ENV=\"${NODE_ENV}\"" "--target=esnext" "--bundle" "--format=iife" "--platform=browser" "--global-name=BUN_RUNTIME" "--minify" "--external:/bun:*" "${BUN_SRC}/runtime/index.ts" "--outfile=${BUN_SRC}/runtime.out.js"
```
paperclover, line 661: can you set node_env to production always for this. the worry is because they live in source you might get an incorrect build if you have debug and release builds.

it would make sense to take these out of src/, but ci currently expects the exact output paths and fixing all that might be a bit tricky.

paperclover (changes requested):
overall great. thanks

CMakeLists.txt
I-A-S: Yes this is an optimization. Current make file goes like this

build runtime.node.pre.out.js
cat runtime.node.pre.out.js runtime.footer.node.js > runtime.node.out.js

build runtime.bun.pre.out.js
cat runtime.bun.pre.out.js runtime.footer.bun.js > runtime.bun.out.js

build instructions for both runtime.node.pre.out.js and runtime.bun.pre.out.js are identical. So instead of building it twice we build it once and call the file runtime.nodebun.pre.out.js
and use it in the cats

CMakeLists.txt
I-A-S: Ah all good currently it's set at the top of the file depending on the build. Debug builds development and release builds production, I'll change this to always production

Issue #6767, opened 2023-10-28
Error in setup.sh to make my first build


TaduJR:
### What version of Bun is running?

1.0.7

### What platform is your computer?

Linux 6.1.21.2-microsoft-standard-WSL2+ x86_64 x86_64

### What steps can reproduce the bug?

bash ./scripts/setup.sh command to make my first build

### What is the expected behavior?

Build without error

### What do you see instead?

-- Downloading WebKit
-- Using static sqlite3
-- Configuring done
-- Generating done
-- Build files have been written to: /home/tadujr/bun/build
ninja: error: loading 'build.ninja': No such file or directory

paperclover:
can you post the full log of CMake?

I-A-S:
This is a little bug in the setup.sh

For now run `ninja -C build` after setup.sh exits with `ninja: error: loading 'build.ninja': No such file or directory`

Hanaasagi:
Same for me, full log 
```
>>​ bash ./scripts/setup.sh
+ git submodule update --init --recursive --progress --depth=1 --checkout src/deps/picohttpparser src/deps/mimalloc src/deps/zlib src/deps/libarchive src/deps/boringssl src/deps/lol-html src/deps/tinycc src/deps/c-ares src/deps/zstd src/deps/base64
[...]

PR #6769, opened 2023-10-28
fix: hotfix for currently broken first build

I-A-S:
### What does this PR do?

Hotfix for the broken first build. Fixes a few minor faults that's currently preventing the builds including issue #6767

### DIFF
--- a/scripts/build-mimalloc.sh
+++ b/scripts/build-mimalloc.sh
@@ -25,4 +25,4 @@ cmake "${CMAKE_FLAGS[@]}" . \
 
 ninja
 
-cp libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.a
+cp libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.o
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -6,6 +6,8 @@ cd ../
 
 bun i
 
+make all-js
+
 mkdir build
 cmake -B build -S . -DCMAKE_BUILD_TYPE=Debug -G Ninja
-ninja
+ninja -C build


scripts/setup.sh
```diff
@@ -6,6 +6,8 @@ cd ../
 
 bun i
 
+make all-js
```
jerome-benoit, line 9: I think the goal is to not rely on make anymore. 
There's no cmake equivalent?

scripts/build-mimalloc.sh
```diff
@@ -25,4 +25,4 @@ cmake "${CMAKE_FLAGS[@]}" . \
 
 ninja
 
-cp libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.a
+cp libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.o
```
jerome-benoit, line 28: A static library file is quite different from an object file ... 
The script should: 
+ generate an object file
+ the object file must be copied to the target directory 

if the object file is expected to be reused

Or

+ the current archive file must be properly handled by the buildsystem

I-A-S: Hey jerome,

Thank you for the comment, I understand that an object file is different from a static library. In this context .o is not meant to represent an object file rather a static library. I understand that this can cause confusion and is not ideal but unfortunately due how the CMake file is written for now, this is the solution. Note this is a hotfix. Look at the cmake file at

`
if(WIN32)
        target_link_libraries(${bun} PRIVATE "${BUN_DEPS_OUT_DIR}/mimalloc-static.lib")
    elseif(APPLE)
        # https://github.com/microsoft/mimalloc/issues/512
        # Linking mimalloc via object file on macOS x64 can cause heap corruption
        target_link_libraries(${bun} PRIVATE "${BUN_DEPS_OUT_DIR}/libmimalloc.a")
    else()
        target_link_libraries(${bun} PRIVATE "${BUN_DEPS_OUT_DIR}/libmimalloc.o")
    endif()
`

jerome-benoit, line 28:
```suggestion
ar -x libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.o
```

scripts/setup.sh
I-A-S: Hey again jerome,

If you look at the pull request list you'll see I already have opened a PR moving runtime js and fallback decoder to CMake. Dave is working on adding bun_error and node-fallbacks to the scripts. The use of make command here is meant to be a hotfix till those changes are integrated, otherwise contributors won't be able to build the project till those changes.

Thank you

scripts/build-mimalloc.sh
jerome-benoit: Given the copied CMakefile content, the current fix will not work in all cases because object files can also be generated: 
+ `[ -f libmimalloc.a ] && cp ...`
+ `[ -f libmimalloc.o ] && cp ...`

I-A-S: As I mentioned in my previous comment, this is not the intended behavior and adding this suggestion will break the build

I-A-S: It is my understanding that object files are not the sole/main output due to the `DMI_BUILD_STATIC=ON` flag and since we are cping from the build directory this shouldn't be a problem. However a second cp command as `cp libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.a` would have helped the mac builds. Eitherway this PR is moot now due to the new script that was just pushed

jerome-benoit: Fair enough. Let's test the new build system then.

PR #6778, opened 2023-10-28
feat(console): add printer for mapiterator

liz3:
MapIterator was not supported in printing, libraries like discordjs make big use of maps and so i think supporting them would be a good idea.

### What does this PR do?

- [x] Code changes

- Manual test

### DIFF
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -1314,6 +1314,7 @@ pub const ZigConsoleClient = struct {
             Error,
             TypedArray,
             Map,
+            MapIterator,
             Set,
             BigInt,
             Symbol,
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
This looks good, but needs a test. You can use `Bun.inspect` to write the test.

src/bun.js/bindings/exports.zig
```diff
@@ -1532,6 +1534,7 @@ pub const ZigConsoleClient = struct {
                         JSValue.JSType.BooleanObject => .Boolean,
                         JSValue.JSType.JSFunction => .Function,
                         JSValue.JSType.JSWeakMap, JSValue.JSType.JSMap => .Map,
+                        JSValue.JSType.JSMapIterator => .MapIterator,
```
Jarred-Sumner, line 1539: Do we need to worry about `SetIterator`?

liz3: il double check

liz3:
@Jarred-Sumner i added a printer for `SetIterator` and added tests :)

Jarred-Sumner (approved):
Thank you

Issue #6762, opened 2023-10-27
Assigning to `process.env` does not overwrite value


michaelhays:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Linux 6.2.0-35-generic x86_64 x86_64

### What steps can reproduce the bug?

Create a `.env` file:

```
FOO=first
```

Create a `run.js` file:

```js
console.log(process.env.FOO)
process.env.FOO = 'second'
console.log(process.env.FOO)
```

### What is the expected behavior?

`bun run.js` should output:

```
first
second
```

### What do you see instead?

`bun run.js` outputs:

```
first
first
```

### Additional information

Testing with Node.js 20.9.0, running `node --env-file=.env run.js` has the correct output:

```
first
second
```

Hanaasagi:
And, delete env also doesn't work as expected.
```JavaScript
console.log(process.env.FOO);
delete process.env.FOO;
console.log(process.env.FOO);
```

This should be `undefined` after `delete`.

link https://github.com/oven-sh/bun/issues/4383

liz3:
i would look at this tmrw unless someone does it before me just lmk here so we dont end up doing duplicated work 😅

Hanaasagi:
Please go ahead with it; I haven't started this. @liz3 :tada:

simylein:
This was working at some point, seems to be a regression. #2696

michaelhays:
Fixed in Bun 1.0.15

Issue #4383, opened 2023-08-29
`delete process.env.XXX` not work


matsumoto0912:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Linux

### What steps can reproduce the bug?

```
console.log(process.env.NODE_ENV);
delete process.env.NODE_ENV;
console.log(process.env.NODE_ENV);
```

### What is the expected behavior?

`undefined` after deleted.

### What do you see instead?

not deleted.

### Additional information

https://nodejs.org/api/process.html#processenv

![2023-08-29_11-23](https://github.com/oven-sh/bun/assets/135786106/2a75ba76-cd2d-4ea9-baca-40791c9d2dd7)

michaelhays:
This is fixed in Bun 1.0.15

PR #6779, opened 2023-10-29
fix: don't inline process.env calls in runtime 

liz3:
### What does this PR do?
This caused issues with reassigning or deleting env vars as the inlining into a static
value meant that the changes would not be reflected at the runtime, while maybe it would be
somehow possible to do static analyisis whether a value is changed that is massively out of scope.
This disables the inlining of environment variables at transpile time.

Fixes: https://github.com/oven-sh/bun/issues/6762
Fixes: https://github.com/oven-sh/bun/issues/4383

- [x] Code changes


Jarred-Sumner (changes requested):
This breaks loading .env files for:
- The runtime in general?
- `bun run <package-json-script>`
- `bun install <lifecycle-package-json-script>`

liz3:
@Jarred-Sumner apologize for that i assumed the behaviour was only for inlining, i refactored it to be a seperate setting

michaelhays:
Hey there, anything left to do here? I'm happy to test on my machine if there's a way I can do that

michaelhays:
Sorry to push, but is there anything I can do to help this get merged?

My project has local scripts which prompt the environment, and then load our environment variables into `process.env` based on that response, so having this work in Bun would improve our workflow a ton!

```ts
const environment = await promptEnvironment()

process.env = {
  ...process.env,
  ...environmentGetEnvVars(environment),
}

const user = await prisma.user.findUniqueOrThrow({ where: ... })
```

liz3:
> Sorry to push, but is there anything I can do to help this get merged?
> 
> My project has local scripts which prompt the environment, and then load our environment variables into `process.env` based on that response, so having this work in Bun would improve our workflow a ton!
> 
> ```ts
> const environment = await promptEnvironment()
> 
> process.env = {
>   ...process.env,
>   ...environmentGetEnvVars(environment),
> }
> 
> const user = await prisma.user.findUniqueOrThrow({ where: ... })
> ```

It seams for regular env variables(not NODE_ENV) this has been resolved, i suspect by https://github.com/oven-sh/bun/pull/7365, at least i can't reproduce the issue anymore.
The reason this didn't get merged is that there is a issue with numerical var keys i.e `123`or `345623` with the underlying engine, you should'nt notice that though with release builds since it only crashes in debug builds.

Should you be in a case where this still happens(upgrade bun maybe if possible), you can work around this by tricking the transpiler:
```js
const getEnv = (k) => process.env[k];

console.log(getEnv("FOO"))
 process.env["FOO"] = "NEW-VALUE";
console.log(getEnv("FOO"))
```
Liz

michaelhays:
Right you are! Upgrading to Bun 1.0.15 fixed this for me :)

Thanks so much @liz3!

PR #6787, opened 2023-10-29
Update contributing.md

rossampro:
Adding ruby as a dev dependency to Arch list

### What does this PR do?

- [ ] Code changes


rossampro:
Making this PR as it seems that `ruby` is now a dependency to run `bun setup` when trying to contribute. Have only added for Arch so far. Can update for the other distros

paperclover:
thanks

PR #6788, opened 2023-10-29
docs(contributing) add ruby dependency

gamedevsam:
Don't know why ruby is required to build bun, but without it `bun setup` fails.

### What does this PR do?

Improves contributing doc.

N / A


paperclover:
thanks

PR #6789, opened 2023-10-30
feat(parseArgs): add parseArgs to node:util module

gamedevsam:
### What does this PR do?

This PR adds the `parseArgs` function into `node:util` module. I was annoyed that I couldn't run any of my CLI scripts with bun because I relied on that function, and so here it is.

The code for the parseArgs implementation is from the @pkgjs/parseargs package, not node (the node version imports lots of internal modules that made it tricky to bring into Bun). The implementation was extracted from this commit: https://github.com/pkgjs/parseargs/commit/67749086a6cb216a4896283de200b1f36646afdd

I tested the built binary with the code on my own repo that leveraged parseArgs. It's a simple package, so there isn't much room for edge cases.

I tried, but got this error:

```
samuel@SAMUEL-DESKTOP:~/WSL_Dev/Personal/bun$ make js
NODE_ENV=production bun src/js/_codegen/index.ts
error: module not found "src/js/_codegen/index.ts"
make: *** [Makefile:552: js] Error 1
```


src/js/internal/util/parse_args/index.js
```diff
@@ -0,0 +1,404 @@
+// TODO: Should these be centralized? They are duplicated lots of times in the `js` folder.
+var __getOwnPropNames = Object.getOwnPropertyNames;
```
gamedevsam, line 2: I feel like these common helper vars should really be centralized and imported in the internal modules, there's a lot of these functions defined in the various JS scripts in there.

gamedevsam:
Tests are still WIP.

src/js/internal/util/parse_args/index.js
gamedevsam: @Jarred-Sumner do you have any opinions / insight here? Was there a reason there were not put in a utility file / shared?

gamedevsam:
Damn migrating tests to Bun test is a gigantic pain...

test/js/node/util/parse_args/utils.test.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const tape = require('tape');
```
Electroid: Thank you for porting all these tests. Could we use `bun:test` instead of `tape`?

gamedevsam: That was my plan initially but I stalled out because it's a significant pain to manually re-write the tests due to different functionality and interfaces of the APIs. I'm not sure when I'll have time & motivation to complete the work, so either someone picks up this branch and helps to re-write the tests, or we merge the work without including the tests, or this will just become a stale PR and someone else with more time and motivation will re-do all the work to bring in parseArgs into bun.

otgerrogla: I'm interested in this feature, I can help porting the tests if you want

paperclover:
i dont know if this is an ideal approach to vendor in a multi-thousand line npm package. this feels like something that should be done in native code.

regardless, if this is the approach used, this solution must use less `.js` files as the number of modules in src/js does add unneeded cost to InternalModuleRegistry.

test/js/node/util/parse_args/utils.test.js
otgerrogla: Finished porting the tests, they are on this commit on my fork: https://github.com/otgerrogla/bun/commit/955495de5a3f9d8e7fac44015da6592bbf171145

However I had to expose some internal functions of parseArgs so they can be accessed by the tests, probably there's a better way to do that.

gamedevsam: I will spend some time analyzing your conversion and will merge it into my PR, thanks for taking this on!

gamedevsam:
> i dont know if this is an ideal approach to vendor in a multi-thousand line npm package. this feels like something that should be done in native code.
> 
> regardless, if this is the approach used, this solution must use less `.js` files as the number of modules in src/js does add unneeded cost to InternalModuleRegistry.

You're welcome to open a PR and re-implement this module in Native code. I agree it would be better, especially if it's a perfect port including all the tests. But this isn't something I was interested in doing, mostly because it's been a long time since I wrote a significant amount of native code (and my native experience begins and ends with C++, not Zig).

My plan with this PR was to bring in working, well tested code, and plug a compatibility hole with Node and ship code that works and is well tested. If this approach is against the "bun philosophy" because it's not fast enough, then feel free to close this PR. I'm interested in speeding up my NPM CLIs with bun, but I'm not going to lose any sleep if I have to keep using Node because bun is missing `parseArgs`.

test/js/node/util/parse_args/utils.test.js
gamedevsam: @otgerrogla I've opened a PR with your changes, but the change to the exported object bothers me, more info: https://github.com/gamedevsam/bun/pull/1/files#r1391944352

I'll spend some time to see if this can be avoided.

gamedevsam: Here's the PR that removes the internal functions from the exported module: https://github.com/otgerrogla/bun/pull/1/files

I have already merged it with my branch, so this PR should be good to go!

gamedevsam:
This is now fully tested and ready for merge.

gamedevsam:
@otgerrogla do you mind pulling my changes and validating everything still works on your end? I ran into a build error when running `bun run setup` today and need to debug my zig installation (this was the error in case you somehow magically know how to fix it: `error: root struct of file 'mem' has no member named 'readIntLittle'`).

paperclover (changes requested):
this needs to be done in native code, for the following reasons

- each file in src/js contributes to startup time and there really should not be more files than truly needed. this adds 5 modules for a single function.
- the amount of code for one function is a ton and i think this will slow down the already large `node:util` module
- as is, this code will trigger assertion failure in BuiltinExecutable for using identifiers `apply` and `call`

**note**: the reason `util.inspect` *does* use this approach is sort of a 'todo: do this in native', but it is more acceptable as 1:1 with node printing is not as trivial to implement in native code, this function is. i should add more comments in these files so future contributors do not follow this pattern.

also, all tests must use `bun:test`

gamedevsam:
Someone with more skills and interest in writing native code can revive this if they want. I will not even attempt it, godspeed.

otgerrogla:
Implemented parseArgs in native code at PR #7310

gamedevsam:
@otgerrogla I am humbled by your skills and persistence. If you don't mind me asking, I would like to poke your brain with a few questions:

1. Are you paid by oven.sh or just doing this for fun (if not, maybe [you should consider it](https://oven.sh/))?
2. How did you learn Zig and how to integrate it with JS in the bun codebase? Just by reading existing code, reading guides, anything else?

I'm very impressed by your skills, rebuilding a JS module in Zig in a week is definitely very impressive to me. Any advice you have for a JS  / C++ coder looking at Zig is appreciated.

otgerrogla:
Thanks, I actually did this as hobby because I thought it would be a good way to both learn Zig, and the inner workings of a JS engine (and learn I did!). I mostly went on by searching how stuff was already made in the existing codebase, with some help of the official Zig docs (as sadly the Zig community is not yet that big).

PR #3814, opened 2023-07-26
Rewrite built-in modules to use CommonJS over ESM

paperclover:
Closes #3773
Closes #3863 
Closes
Fixes immediate segfault of #3746 but astro isnt working fully yet

- This changes all `node:*` and `bun:*` modules to use CommonJS modules. Since the bundler does not emit CommonJS code, this is done with a custom pre/postprocessor step.
  - `require` is intercepted at build time and inlined to `$requireId(...)` where a static number is inlined.
  - Introduce `InternalModuleRegistry` in C++. This is an internal fields object of about 50.
- All builtin modules now actually use builtins syntax. Meaning we can access private globals like `$isArray` and so on.
- `make builtins` & `make esm` are now one build process: `make js`. It is also faster.
- Removes some hacks we had
  - `Symbol.for("CommonJS")`
  - `$lazy("primordials")`
  - in the future we can remove `globalThis[Symbol.for("Bun.lazy")]` and make `$lazy` real, or take a diff approach to this.
- Changes how native c++ modules are formatted to be a bit easier for us to add more if we want.
- Removes `bun:events_native`. it is now `$lazy('events')`, we should fully remove this later.
- Moves bun:jsc into a native module
- process.binding('constants') into native code, add `ProcessBindingConstants` and `$processBindingConstants`
- a few specific strings to process.binding() will throw issues with github issue links.

PR #6790, opened 2023-10-30
Fix memory leak in `require`

Jarred-Sumner:
### What does this PR do?

It makes `require("discord.js")` use 30% less memory

- This fixes a memory leak in `require` that regressed in #3814
- This adds a small performance optimization for the JS parser when there are many scopes created in a file

```js
❯ bun discordy.js
[90.22ms] Discord.js
Memory 55.1875

❯ bun-1.0.6 discordy.js
[92.39ms] Discord.js
Memory 73.5625
```

This adds a test to make sure we don't regress again.


src/bun.js/bindings/ZigSourceProvider.cpp
```diff
@@ -153,6 +156,9 @@ void SourceProvider::cacheBytecode(const BytecodeCacheGenerator& generator)
     if (update)
         m_cachedBytecode->addGlobalUpdate(*update);
 }
+SourceProvider::~SourceProvider()
+{
```
liz3, line 164: is the remove of the two calls in the destructor intentional?

Jarred-Sumner: The two calls do nothing so it's just confusing

PR #6800, opened 2023-10-30
Fix libmimalloc linking on Linux

tarruda:
The CMakeLists.txt is linking against `libmimalloc.o` which doesn't exist.


gamedevsam:
Bless your soul.

jerome-benoit:
Please remove the OS conditions and ensure the `.a` archive will be picked up in all cases.

paperclover:
this isnt the correct solution. we're linking the o file on linux for a reason. the issue was adding this line to the dependency build scripts:

https://github.com/oven-sh/bun/blob/36984a8773610e02a5f1ce237f4225667488722b/scripts/build-mimalloc.sh#L29

Issue #6281, opened 2023-10-04
Segfault on some constructors when used without `new`


colinhacks:
### What steps can reproduce the bug?

```ts
import { File } from "buffer";
File();
```

```ts
import {Script} from "vm";
Script();
```

Jarred-Sumner:
This needs to become a function that throws a TypeError "Class constructor File cannot be used without new"

https://github.com/oven-sh/bun/blob/main/src/bun.js/bindings/JSDOMFile.cpp#L20

krk:
`File` fixed in https://github.com/oven-sh/bun/pull/5881 with a typo in the message.

PR #6809, opened 2023-10-30
Fix 6281

krk:
### What does this PR do?

Throw an error when `Script` is called without new, fix typo in error for `File` without new.

Fix https://github.com/oven-sh/bun/issues/6281

- [X] Code changes

I wrote automated tests


Jarred-Sumner:
Thank you

PR #6811, opened 2023-10-31
chore: build system improvements

paperclover:
### What does this PR do?

This addresses build system issues.

- changes to use latest bun in ci.
- fix mimalloc linking
- enable verbose on ci builds


scripts/all-dependencies.sh
```diff
@@ -56,8 +56,8 @@ dep boringssl libcrypto.a libssl.a libdecrepit.a
 dep cares libcares.a
 dep libarchive libarchive.a
 dep lolhtml liblolhtml.a
-dep mimalloc-debug libmimalloc-debug.a
-dep mimalloc libmimalloc.a
+dep mimalloc-debug libmimalloc-debug.a libmimalloc-debug.o
+dep mimalloc libmimalloc.a libmimalloc.o
```
jerome-benoit, line 60: What is the rationale for using `.o` here if its usage is in a static binary?

scripts/build-mimalloc-debug.sh
```diff
@@ -27,3 +27,4 @@ cmake "${CMAKE_FLAGS[@]}" . \
 ninja
 
 cp libmimalloc-debug.a $BUN_DEPS_OUT_DIR/libmimalloc-debug.a
+cp CMakeFiles/mimalloc-obj.dir/src/static.c.o $BUN_DEPS_OUT_DIR/libmimalloc-debug.o
```
jerome-benoit, line 30: Why picking `.o` directly in the CMakeFiles intermediate files? Their locations and names is likely to change. 

The change to the deps list should generate it directly as a final build artefact.

scripts/all-dependencies.sh
paperclover: It supposedly allows mimalloc to override `malloc` easier, but I'm not sure. this logic is ported from the old Makefile.

```make
_MIMALLOC_FILE = libmimalloc.o
_MIMALLOC_INPUT_PATH = CMakeFiles/mimalloc-obj.dir/src/static.c.o
```
```make
ifeq ($(OS_NAME),darwin)
  _MIMALLOC_OBJECT_FILE = 0
  _MIMALLOC_FILE = libmimalloc.a
  _MIMALLOC_INPUT_PATH = libmimalloc.a
  _MIMALLOC_LINK = -lmimalloc
endif
```

Not sure if this claim is still relevant, or any performance differences exist, but it is easier to keep the old system going.

scripts/build-mimalloc-debug.sh
paperclover: (see other comment)

jerome-benoit: > (see other comment)

It would be better if the mimalloc build artefacts is a `.o` file, then ar is  used to create the matching `.a`. The current behaviour is enforced by the mimalloc buildsystem behaviour at outputting only archive file and a need to use to `.o` file for an undocumented reason.

.github/workflows/bun-mac-aarch64.yml
```diff
@@ -6,6 +6,7 @@ concurrency:
 
 env:
   LLVM_VERSION: 16
+  BUN_DOWNLOAD_URL_BASE: https://pub-5e11e972747a44bf9aaf9394f185a982.r2.dev/releases/latest
```
Jarred-Sumner, line 9: @Electroid can you make this a public URL?

Jarred-Sumner: ignore me it is

PR #6814, opened 2023-10-31
peer dependency and semver prerelease bug fixes

dylan-conway:
### What does this PR do?
- Ensures peer dependencies do not install multiple versions.
- Installs existing packages instead the maximum version for a range. Previously this would install `bar@1.0.1` and the highest version for `bar@^1.0.1`. Now it will only install `bar@1.0.1`.
```
root
├── bar@1.0.1
└── foo@1.0.0
    └── bar@^1.0.1
```
- Orders prerelease tags by numbers separated by `.` (`1.0.0-canary.0.0.0.1` is less than `1.0.0-canary.0.0.0.2`)

- [ ] Code changes

manually tested


Jarred-Sumner:
Thank you

PR #6819, opened 2023-10-31
workflows(bun-mac-x64.yml): Fix typo "bun-obj-darwin-x6"

Marukome0743:
### What does this PR do?

Change `bun-obj-darwin-x6` to `bun-obj-darwin-x64` at line 110 in [bun-mac-x64.yml](https://github.com/oven-sh/bun/blob/main/.github/workflows/bun-mac-x64.yml).

- [ ] Code changes


paperclover:
interesting that we dont actually use that variable. still, cant believe i left that typo. maybe we should just remove the unused variables.

Electroid (approved):

PR #6857, opened 2023-11-02
Adds X509 Common Name to checkServerIdentity

ansg191:
### What does this PR do?

Adds support in `checkX509ServerIdentity` for hostnames & IP addresses in the common name of a X509 certificate subject if **and only if** there is no SAN in the certificate.

- [x] Code changes

  - [x] Tested locally
  - [ ] Unable to run `zig build test` per #6791 
  - [ ] Unable to test using fixtures in `test/` dir until #6621


Jarred-Sumner:
Closing due to bitrot, also not sure about this API

PR #6861, opened 2023-11-02
Add Bun.Glob

zackradisic:
### What does this PR do?

Still WIP

This adds a `Glob` class to bun which implements glob pattern matching on filesystem/strings.

Basic example:
```typescript
import { Glob } from "bun";

const glob = new Glob("**/*.ts");
const entries = await glob.match({ cwd: "src" });

console.log("Entries", entries);
```

This adds

- [x] Code changes

Still need to do:
- [x] symlink support
- [x] more comprehensive tests (see blow)
- [x] return async iterator for `Glob.scan()`

I wrote automated tests:
- [x] Basic tests for glob matching/walking
- [x] Copy tests from fast-glob
- [x] Weird edge case tests (invalid surrogate pairs, filenames with latin1 characters, etc.)
- [x] Stress tests

If Zig files changed:

- [x] I made sure that specific versions of dependencies are used instead of ranged or tagged versions

If \*.classes.ts files were added or changed:


src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,381 @@
...
+is_ascii: bool,
+
+const MatchOpts = struct {
+    cwd: ?BunString,
+    dot: bool,
+    absolute: bool,
+
+    fn fromJS(globalThis: *JSGlobalObject, arguments: *ArgumentsSlice, comptime fnName: []const u8, arena: *Arena) ?MatchOpts {
+        const optsObj: JSValue = arguments.nextEat() orelse return null;
+        if (!optsObj.isObject()) {
```
Jarred-Sumner, line 53: Probably should ignore if the value is `undefined`

src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,381 @@
...
+            return null;
+        }
+
+        var out: MatchOpts = .{
+            .cwd = null,
+            .dot = false,
+            .absolute = false,
+        };
+
+        if (optsObj.get(globalThis, "absolute")) |absoluteVal| parse_absolute: {
```
Jarred-Sumner: Same here, can use `getTruthy` which ignores null or undefined

src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,381 @@
...
+    return BunString.toJSArray(globalThis, globWalk.matchedPaths.items[0..]);
+}
+
+fn makeGlobWalker(
+    this: *Glob,
+    globalThis: *JSGlobalObject,
+    arguments: *ArgumentsSlice,
+    comptime fnName: []const u8,
+    alloc: Allocator,
+    arena_: Arena,
```
Jarred-Sumner: The pointer to the arena will no longer be valid outside this scope. This is a very confusing thing about the arena allocator in Zig’s standard library.

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+    }
+};
+
+pub const GlobWalker = struct {
+    pub const Result = Maybe(void);
+
+    arena: Arena = undefined,
+
+    /// not owned by this struct
+    pattern: []const u8 = undefined,
```
Jarred-Sumner: We have to be very careful whenever `undefined` is used 

Can this instead be an empty string?

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
+const std = @import("std");
+const math = std.math;
+const mem = std.mem;
+const BunString = @import("./bun.zig").String;
+const expect = std.testing.expect;
+const Allocator = std.mem.Allocator;
+const ArrayList = std.ArrayListUnmanaged;
+const ArrayListManaged = std.ArrayList;
+const DirIterator = @import("./bun.js/node/dir_iterator.zig");
+const Syscall = @import("./sys.zig");
```
Jarred-Sumner: This should really be encoded in a linter, but can you use namespaces instead of relative imports when possible? This makes it easier for us to move files around in the future and sometimes also with autocomplete 

`bun.sys` in this case

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+        this: *GlobWalker,
+        err: Syscall.Error,
+        path_buf: [:0]const u8,
+    ) Syscall.Error {
+        @memcpy(this.pathBuf[0 .. path_buf.len + 1], @as([]const u8, @ptrCast(path_buf[0 .. path_buf.len + 1])));
+        return err.withPath(this.pathBuf[0 .. path_buf.len + 1]);
+    }
+
+    pub fn walk(this: *GlobWalker) !Maybe(void) {
+        if (this.patternComponents.items.len == 0) return .{ .result = undefined };
```
Jarred-Sumner: Why are we returning undefined? Lets use an empty string or something which indicates it is empty instead of undefined. Undefined is undefined behavior in release builds

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+
+        while (this.workbuf.items.len > 0) {
+            const work_item: WorkItem = this.workbuf.pop();
+            switch (try this.handleDirEntries(&work_item, &path_buf, false)) {
+                .err => |err| return .{ .err = err },
+                else => {},
+            }
+        }
+
+        return .{ .result = undefined };
```
Jarred-Sumner: Same here

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+                    this.patternComponents.items[component_idx + 1].syntax_hint == .Double) : (component_idx += 1)
+                {}
+            }
+            break :component_idx component_idx;
+        };
+
+        var pattern = this.patternComponents.items[component_idx];
+        // On windows we convert the codepoint buffer upfront so no need for
+        // these checks
+        if (comptime !isWindows) {
```
Jarred-Sumner: Can you add an assert for this on Windows?

src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,381 @@
...
+                if (ResolvePath.Platform.auto.isAbsolute(cwd_str_raw.slice())) {
+                    const cwd_str = cwd_str_raw.clone(arena.allocator()) catch {
+                        globalThis.throwOutOfMemory();
+                        return null;
+                    };
+                    break :cwd_str cwd_str.ptr[0..cwd_str.len];
+                }
+
+                var path_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
+                var path_buf2: [bun.MAX_PATH_BYTES]u8 = undefined;
```
Jarred-Sumner: I suggest making this `bun.MAX_PATH_BYTES*2` just in case

src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,381 @@
...
+                const cwd = std.os.getcwd(&path_buf) catch {
```
Jarred-Sumner: Use `bun.sys.getcwd` instead because `getcwd` behaves strangely on Windows iirc

src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    return glob;
+}
+
+pub fn finalize(
+    this: *Glob,
+) callconv(.C) void {
+    const alloc = JSC.VirtualMachine.get().allocator;
+    alloc.free(this.pattern);
```
Jarred-Sumner, line 386: What keeps the `pattern` alive if the `Glob` object is freed before the `Promise` for an async task resolves successfully? You might need to use `hasPendingActivity` and make it return `true` when promises have not been fulfilled yet.

Jarred-Sumner: Something like this should reproduce it:

```zig
import {Glob} from 'bun';
const promise = (async () => () {
  const glob = new Glob("**/*.zig")
  return glob.match("/");
})();
Bun.gc(true);
await promise;
```

bench/glob/match.js
```diff
@@ -0,0 +1,125 @@
+import { run, bench, group } from "mitata";
+import fg from "fast-glob";
+import { Glob, Transpiler } from "bun";
```
Jarred-Sumner: maybe we should have a lowercase `glob` API to avoid the constructor

```js
import {glob} from 'bun';
glob.match("pattern", "folder')
```

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+                needles[i] = @splat(tok);
+            }
+            break :needles needles;
+        };
+
+        var i: usize = 0;
+        while (i + 16 <= pattern.len) : (i += 16) {
+            const haystack: @Vector(16, u8) = pattern[i..][0..16].*;
+            inline for (needles) |needle| {
+                if (std.simd.firstTrue(needle == haystack) != null) return true;
```
Jarred-Sumner: nice

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+                        }
+                        continue;
+                    };
+
+                    const subdir_parts: []const []const u8 = &[_][]const u8{
+                        dir_path[0..dir_path.len],
+                        entry_name,
+                    };
+
+                    const subdir_entry_name = try std.fs.path.join(this.arena.allocator(), subdir_parts);
```
Jarred-Sumner: Can we use bun's file path join logic instead of `std.fs.path`? I don't generally trust the zig standard library for string-parsing related code in it's current level of stability. It would be unsurprising if it didn't work with Windows' UTF-16 ill-formed surrogates or otherwise handle unicode correctly.

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+    }
+
+    fn matchPatternImpl(
+        this: *GlobWalker,
+        pattern_component: *Component,
+        filepath: []const u8,
+    ) bool {
+        // FIXME: handle utf-16?
+        if (!this.dot and GlobWalker.startsWithDot(filepath)) return false;
+        switch (pattern_component.syntax_hint) {
```
Jarred-Sumner: Small style thing: `switch` can be an expression

```suggestion
return switch (pattern_component.syntax_hint) {
            .Double, .Single => true,
            .WildcardFilepath => if (comptime !isWindows)
                matchWildcardFilepath(this.pattern[pattern_component.start .. pattern_component.start + pattern_component.len], filepath)
            else
                this.matchPatternSlow(pattern_component, filepath),
            .Literal => if (comptime !isWindows)
                matchWildcardLiteral(this.pattern[pattern_component.start .. pattern_component.start + pattern_component.len], filepath)
            else
                this.matchPatternSlow(pattern_component, filepath),
            else => this.matchPatternSlow(pattern_component, filepath),
};
```

src/glob.zig
```diff
@@ -0,0 +1,2555 @@
...
+        this: *GlobWalker,
+        entry_name: []const u8,
+        component_idx: u32,
+        is_last: bool,
+        pattern: *Component,
+        next_pattern: ?*Component,
+    ) bool {
+        // Handle case a)
+        if (!is_last) return pattern.syntax_hint == .Double and
+            component_idx + 1 == this.patternComponents.items.len - 1 and
```
Jarred-Sumner:
```suggestion
            component_idx + 1 == this.patternComponents.items.len -| 1 and
```

src/bun.js/api/glob.zig
zackradisic: What is the best way to handle this, because the GlobWalker wants to own an `Arena`. Should I pass an `*Arena` to this function and then dereference it to give it to the `GlobWalker`?

Or should `GlobWalker` take an `*Arena`, and only when it is done allocating it can dereference it and store it?

src/glob.zig
zackradisic: I changed it to `u0` is that okay?

The-King-of-Toasters:
FWIW, I've updated https://github.com/The-King-of-Toasters/globlin with your syntax changes (which I should've done a while ago). Just some questions:

- Why strip out the micromatch test suite?
- What is the purpose of `preprocess_glob`?

Also a shoutout in the release notes would be nice ;).

zackradisic:
> * Why strip out the micromatch test suite?
> * What is the purpose of `preprocess_glob`?



Zig tests are not working in bun right now, and I made some slight changes that caused all the tests to fail to compile, so I commented them out for now, but I'm going to port them to Typescript in globMatch.test.ts

To be honest I can't exactly remember what I had in mind for `preprocess_glob()`, I think the idea was to compile a glob pattern into tokens (e.g. `**` becomes `Globstar`) or something

> Also a shoutout in the release notes would be nice ;).

For sure!

test/js/bun/glob/scan.test.ts
```diff
@@ -0,0 +1,321 @@
...
+  test("invalid surrogate pairs", async () => {
+    const pattern = `**/*.{md,\uD83D\uD800}`;
+    const cwd = "test/js/bun/glob";
+
+    const glob = new Glob(pattern);
+    const entries = await Array.fromAsync(glob.scan({ cwd }));
+
+    expect(entries.sort()).toEqual(
+      [
+        "fixtures/file.md",
```
Electroid, line 147: For the fixtures, instead of having them committed to source, I would recommend importing `createTmpDir` (not exact signature) from `"harness"`. This makes a bunch of temp files in a temp directory.

zackradisic: Removed the fixtures from git also the tests need to scan the `test/js/bun/glob` dir so I made a function to create a temp `fixtures` dir there

test/js/bun/glob/stress.test.ts
```diff
@@ -0,0 +1,34 @@
...
+];
+
+test("Glob.scan stress test", async () => {
+  await Promise.all(
+    Array(1000)
+      .fill(null)
+      .map(() =>
+        Array.fromAsync(new Glob("src/**/*.zig").scan()).then(results => {
+          const set = new Set(results);
+          return set.size == paths.length && paths.every(path => set.has(path));
```
Jarred-Sumner, line 31: I am confused why it would be matching `*.zig` for a list of `.md` and `.json` files.

test/js/bun/glob/match.test.ts
```diff
@@ -0,0 +1,1501 @@
...
+    glob = new Glob("hello/*/friends");
+    expect(glob.match("hello/\uD800\uD800/friends")).toBeTrue();
+
+    glob = new Glob("*.{js,\uD83D\u0027}");
+    expect(glob.match("runtime.node.pre.out.ts")).toBeFalse();
+    expect(glob.match("runtime.node.pre.out.js")).toBeTrue();
+  });
+
+  /**
+   * These tests are ported from micromatch, glob-match, globlin
```
Jarred-Sumner, line 136: can you put their licenses at the top of this file?

src/bun.js/api/glob.zig
```diff
@@ -0,0 +1,455 @@
...
+    defer arguments.deinit();
+
+    var arena = std.heap.ArenaAllocator.init(alloc);
+    var globWalker = this.makeGlobWalker(globalThis, &arguments, "scan", alloc, &arena) orelse {
+        arena.deinit();
+        return .undefined;
+    };
+
+    updateHasPendingActivityFlag(&this.has_pending_activity, true);
+    var task = WalkTask.create(globalThis, alloc, globWalker, &this.has_pending_activity) catch {
```
Jarred-Sumner, line 422: what if there are multiple simultaneous walkers? should `has_pending_activity` be a number that gets incremented/decremented?

src/js/builtins/Glob.ts
```diff
@@ -0,0 +1,21 @@
+interface Glob {
+  __scan(opts);
```
Jarred-Sumner: this is not great, people might try to use it as an API

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+            .cursor = CodepointIterator.Cursor{
+                .i = this.cursor.i + i,
+                .c = @truncate(nextCp),
+                .width = 1,
+            },
+        };
+    }
+};
+
+var dummyu32 = [_]u32{ 105, 102, 32, 121, 111, 117, 32, 115, 101, 101, 32, 116, 104, 105, 115, 44, 32, 122, 97, 99, 107, 32, 105, 115, 32, 97, 110, 32, 105, 100, 105, 111, 116 };
```
Jarred-Sumner: what is this? why is it a var?

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+    /// If the pattern contains "./" or "../"
+    has_relative_components: bool = false,
+
+    patternComponents: ArrayList(Component) = .{},
+    matchedPaths: ArrayList(BunString) = .{},
+    i: u32 = 0,
+
+    dot: bool = false,
+    absolute: bool = false,
+    cwd: BunString = undefined,
```
Jarred-Sumner: can this be `BunString.empty` instead? `undefined` has caused strange issues many times
```suggestion
    cwd: BunString = BunString.empty,
```

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+    follow_symlinks: bool = false,
+    error_on_broken_symlinks: bool = false,
+    only_files: bool = true,
+
+    root_cwd_slice: []const u8 = undefined,
```
Jarred-Sumner: can this be `""` instead? `undefined` has caused strange issues many times
```suggestion
    root_cwd_slice: []const u8 = "",
```

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+            dot,
+            absolute,
+            follow_symlinks,
+            error_on_broken_symlinks,
+            only_files,
+        );
+        return .{ .result = globWalker };
+    }
+
+    pub fn convertUtf8ToCodepoints(codepoints: []u32, pattern: []const u8) void {
```
Jarred-Sumner: technically on Windows, this can be literally anything unfortunately.

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+    ) Syscall.Error {
+        @memcpy(this.pathBuf[0 .. path_buf.len + 1], @as([]const u8, @ptrCast(path_buf[0 .. path_buf.len + 1])));
+        return err.withPath(this.pathBuf[0 .. path_buf.len + 1]);
+    }
+
+    pub fn walk(this: *GlobWalker) !Maybe(u0) {
+        if (this.patternComponents.items.len == 0) return .{ .result = 0 };
+        var path_buf: [bun.MAX_PATH_BYTES]u8 = std.mem.zeroes([bun.MAX_PATH_BYTES]u8);
+
+        const root_path = this.cwd.toZigString().slice();
```
Jarred-Sumner: this is going to leak memory or not be thread safe (must call `.toThreadSafe` before `BunString` can be accessed on another thread)

you should copy the directory as a string IMO

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+            }
+        }
+
+        dir_path.len = len;
+
+        return component_idx;
+    }
+
+    // NOTE you must check that the pattern at `idx` has `syntax_hint == .Double` first
+    fn collapseSuccessiveDoubleWildcards(this: *GlobWalker, idx: u32) u32 {
```
Jarred-Sumner: nice

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+
+        const flags = std.os.O.DIRECTORY | std.os.O.RDONLY;
+        const fd = switch (Syscall.openat(cwd_fd, dir_path, flags, 0)) {
+            .err => |err| return .{
+                .err = this.handleSysErrWithPath(err, dir_path),
+            },
+            .result => |fd_| fd_,
+        };
+        defer {
+            // _ = Syscall.close(fd);
```
Jarred-Sumner:
```suggestion

```

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+        switch (try this.handleDirEntriesImpl(fd, component_idx, dir_path)) {
+            .err => |err| return .{ .err = err },
+            .result => return .{ .result = 0 },
+        }
+    }
+
+    fn handleDirEntriesImpl(this: *GlobWalker, fd: bun.FileDescriptor, component_idx: u32, dir_path: [:0]const u8) !Maybe(u0) {
+        const only_files = this.only_files;
+        const dot = this.dot;
+        _ = dot;
```
Jarred-Sumner: is this intentional?

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+                return true;
+        }
+
+        return false;
+    }
+
+    /// NOTE This doesn't check that there is leading dot, use `hasLeadingDot()` to do that
+    fn removeLeadingDot(filepath: []const u8, comptime allow_non_utf8: bool) []const u8 {
+        if (comptime bun.Environment.isWindows and allow_non_utf8) {
+            return filepath[4..];
```
Jarred-Sumner: can you assert that there is a leading dot?

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+        } else {
+            return filepath[2..];
```
Jarred-Sumner: can you assert that there is a leading dot?

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+                switch (c) {
+                    '*', '[', '{', '?', '!' => return true,
+                    else => {},
+                }
+            }
+            return false;
+        }
+
+        const syntax_tokens = comptime [_]u8{ '*', '[', '{', '?', '!' };
+        const needles: [syntax_tokens.len]@Vector(16, u8) = needles: {
```
Jarred-Sumner: can this be `comptime`?
```suggestion
        const needles: [syntax_tokens.len]@Vector(16, u8) = comptime needles: {
```

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+        const iter = CodepointIterator.init(pattern);
+        var cursor = CodepointIterator.Cursor{};
+
+        var cp_len: u32 = 0;
+        var prevIsBackslash = false;
+        while (iter.next(&cursor)) : (cp_len += 1) {
+            const c = cursor.c;
+
+            const cu8: u8 = @truncate(c);
+            _ = cu8;
```
Jarred-Sumner: did you mean to keep this?
```suggestion

```

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+    var negated = false;
+    while (state.glob_index < glob.len and glob[state.glob_index] == '!') {
+        negated = !negated;
+        state.glob_index += 1;
+    }
+
+    while (state.glob_index < glob.len or state.path_index.cursor.i < path.len) {
+        if (state.glob_index < glob.len) {
+            const gu8: u8 = @truncate(glob[state.glob_index]);
+            _ = gu8;
```
Jarred-Sumner: did you mean to keep this?
```suggestion

```

src/glob.zig
```diff
@@ -0,0 +1,3021 @@
...
+            const pu8: u8 = @truncate(state.path_index.cursor.c);
+            _ = pu8;
```
Jarred-Sumner: did you mean to keep this?
```suggestion

```

src/js/builtins/Glob.ts
Electroid: We should make this private somehow, people _will_ use it as an API.

zackradisic: We could probably make it private with the `#` syntax so we have `#__scan` which is only accessible within the class

Jarred-Sumner: We can use private symbols for this. `vm.commonIdentifiers.match` and `vm.commonIdentifiers.test` I think will work.

test/js/bun/glob/stress.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  "test/js/bun/glob/fixtures/second/nested/directory/file.md",
+  "test/js/bun/glob/fixtures/third/library/b/book.md",
+  "test/js/bun/glob/fixtures/third/library/a/book.md",
+  "test/js/bun/glob/fixtures/first/file.md",
+  "test/js/bun/glob/fixtures/first/nested/file.md",
+  "test/js/bun/glob/fixtures/first/nested/directory/file.md",
+  "test/js/bun/glob/fixtures/first/nested/directory/file.json",
+];
+
+tempFixturesDir();
```
Jarred-Sumner: This code should not be run in the top-level of the file incase the file is loaded but the test is skipped. Instead, it should be within a `beforeAll` callback:

```suggestion
beforeAll(() => {
  tempFixturesDir();
 });
```

test/js/bun/glob/stress.test.ts
```diff
@@ -0,0 +1,39 @@
+import { expect, test, describe } from "bun:test";
```
Jarred-Sumner:
```suggestion
import { expect, test, describe, beforeAll } from "bun:test";
```

test/js/bun/glob/stress.test.ts
```diff
@@ -0,0 +1,39 @@
...
+          const set = new Set(results);
+          return set.size == paths.length && paths.every(path => set.has(path));
+        }),
+      ),
+  );
+});
+
+test("Glob.match stress test", () => {
+  for (let i = 0; i < 10000; i++) {
+    new Glob("src/**/*.zig").match("src/cli/package_manager_command.zig");
```
Jarred-Sumner:
```suggestion
    if (!new Glob("src/**/*.zig").match("src/cli/package_manager_command.zig")) {
       throw new Error("test failed on run " + i);
    }
```

src/glob.zig
```diff
@@ -0,0 +1,3032 @@
...
+        }
+
+        return .{ .result = 0 };
+    }
+
+    pub fn handleSymlink(
+        this: *GlobWalker,
+        work_item: *const WorkItem,
+        scratch_path_buf: *[bun.MAX_PATH_BYTES]u8,
+    ) !Maybe(u0) {
```
Jarred-Sumner: why is this a `Maybe(u0)` rather than a `Maybe(void)`? Usually `u0` is useful for things which we might want to think of like a number but disable it

src/glob.zig
```diff
@@ -0,0 +1,3032 @@
...
+            if (comptime bun.Environment.isPosix) {
+                const m = stat_result.mode & std.os.S.IFMT;
+                switch (m) {
+                    std.os.S.IFDIR => break :kind .directory,
+                    std.os.S.IFLNK => break :kind .sym_link,
+                    std.os.S.IFREG => break :kind .file,
+                    else => {},
+                }
+            } else if (comptime bun.Environment.isWindows) {
+                @panic("TODO!");
```
Jarred-Sumner:
```suggestion
     return bun.todo(@src(), .{.result = {}});
```

src/glob.zig
```diff
@@ -0,0 +1,3032 @@
...
+        if (encountered_dot_dot) {
+            const flags = std.os.O.DIRECTORY | std.os.O.RDONLY;
+            const fd = switch (Syscall.openat(cwd_fd, dir_path, flags, 0)) {
+                .err => |err| return .{
+                    .err = this.handleSysErrWithPath(err, dir_path),
+                },
+                .result => |fd_| fd_,
+            };
+            defer {
+                _ = Syscall.closeAllowingStdoutAndStderr(fd);
```
Jarred-Sumner: This can just be `close`, it should never be stdout or stderr. If it is, that is a bug.
```suggestion
                _ = Syscall.close(fd);
```

src/glob.zig
```diff
@@ -0,0 +1,3032 @@
...
+
+        const flags = std.os.O.DIRECTORY | std.os.O.RDONLY;
+        const fd = switch (Syscall.openat(cwd_fd, dir_path, flags, 0)) {
+            .err => |err| return .{
+                .err = this.handleSysErrWithPath(err, dir_path),
+            },
+            .result => |fd_| fd_,
+        };
+        defer {
+            _ = Syscall.closeAllowingStdoutAndStderr(fd);
```
Jarred-Sumner:
```suggestion
            _ = Syscall.close(fd);
```

test/js/bun/glob/stress.test.ts
```diff
@@ -1,26 +1,26 @@
...
+  path.join(import.meta.dir, "fixtures/first/nested/directory/file.json"),
 ];
 
 beforeAll(() => {
   tempFixturesDir();
 });
 
 test("Glob.scan stress test", async () => {
-  const cwd = "test/js/bun/glob";
+  const cwd = import.meta.dir;
```
Jarred-Sumner, line 23: @zackradisic the cwd is the `test` folder in the process, not the `bun` directory. Did you manually edit bunfig.toml to make it not the `test` directory?

Jarred-Sumner:
@zackradisic the next steps here to merge this are:

1) Add a memory leak test for sync & async which runs in a separate process, runs several scans of probably the same directory and waits for each one to complete one at a time. It should run `Bun.gc(true)` before and after, and compare `process.memoryUsage.rss()`. It should be executed with `--smol`. The memory usage gap should not be more than around 100 MB or so.
2) Fix the Arena allocator issue 
3) Fix the Windows build

I did some cleanup work to make the tests pass, fix a memory leak, do the zig upgrade

src/install/lockfile.zig
```diff
@@ -3448,23 +3532,23 @@ pub const Package = extern struct {
...
+                // if (!strings.endsWithComptime(without_trailing_slash, "/*") and !strings.eqlComptime(without_trailing_slash, "*")) {
+                //     log.addError(source, item.loc,
+                //         \\TODO glob star * in the middle of a path. For now, try something like "packages/*", at the end of the path.
+                //     ) catch {};
+                //     continue;
+                // }
+
+                glob_workspace_paths.append(input_path) catch unreachable;
                 continue;
             } else if (strings.containsAny(input_path, "!{}[]")) {
```
dylan-conway: can we append to `glob_workspace_paths` for this case too?

dylan-conway: I think each `input_path` should be appended, because each should be a valid glob pattern

src/install/lockfile.zig
```diff
@@ -3533,93 +3617,88 @@ pub const Package = extern struct {
             });
         }
 
-        if (asterisked_workspace_paths.items.len > 0) {
+        if (glob_workspace_paths.items.len > 0) {
```
dylan-conway: if each is appended, this could go inside the `arr.slice()` loop

src/install/lockfile.zig
```diff
@@ -3368,6 +3376,70 @@ pub const Package = extern struct {
         version: ?[]const u8 = null,
     };
 
+    fn processWorkspaceNameGlob(
```
dylan-conway: unused?

src/install/lockfile.zig
```diff
@@ -3368,6 +3376,70 @@ pub const Package = extern struct {
...
+        defer walker.deinit();
+
+        var iterator: GlobWalker.Iterator = .{
+            .walker = &walker,
+        };
+        switch (try iterator.init()) {
+            .err => |err| {
+                _ = err;
+                // FIXME handle errors gracefully
+                @panic("Ooops");
```
Jarred-Sumner: This does need to be done before we can ship.

src/install/lockfile.zig
```diff
@@ -3368,6 +3376,70 @@ pub const Package = extern struct {
...
+            },
+            else => {},
+        }
+        defer iterator.deinit();
+
+        while (switch (try iterator.next()) {
+            .err => |err| {
+                _ = err;
+                // FIXME handle errors gracefully
+                @panic("Ooops");
```
Jarred-Sumner: This does need to be done before we can ship.

src/install/lockfile.zig
```diff
@@ -3368,6 +3376,70 @@ pub const Package = extern struct {
...
+            },
+            .result => |r| r,
+        }) |matched_path| {
+            _ = matched_path;
```
Jarred-Sumner: Why isn't this returned?

src/install/lockfile.zig
```diff
@@ -3533,93 +3617,88 @@ pub const Package = extern struct {
...
+                    false,
+                    false,
+                    false,
+                    false,
+                    false,
+                )) {
+                    .err => |err| {
+                        _ = err;
+                        // FIXME handle errors gracefully
+                        @panic("Ooops");
```
Jarred-Sumner: This does need to be handled

src/resolver/resolve_path.zig
```diff
@@ -436,16 +436,16 @@ pub fn relativeNormalized(from: []const u8, to: []const u8, comptime platform: P
 pub fn dirname(str: []const u8, comptime platform: Platform) []const u8 {
     switch (comptime platform.resolve()) {
         .loose => {
-            const separator = lastIndexOfSeparatorLoose(str);
-            return str[0 .. separator + 1];
+            const separator = lastIndexOfSeparatorLoose(str) orelse return "";
+            return str[0..separator];
```
Jarred-Sumner, line 440: Did you check this doesn't break anything?

Jarred-Sumner:
🎉

Issue #6827, opened 2023-10-31
Bun write from empty Response doesn't create any file.


tr1ckydev:
### What version of Bun is running?

1.0.8+732650d6a43d235d36a9bf8717766e012f445461

### What platform is your computer?

Linux 6.5.9-1-cachyos x86_64 unknown

### What steps can reproduce the bug?

```ts
Bun.write("ok.txt", new Response(""));
```

### What is the expected behavior?

An empty file `ok.txt` will be created in the current directory.

### What do you see instead?

No file is created.

### Additional information

This doesn't happen on passing empty string.
```ts
Bun.write("ok.txt", "");
```
This generates the empty file successfully.

RiskyMH:
Also if the file exists, then it does make content ""

PR #6872, opened 2023-11-03
create empty file when Bun.write with empty Response

YutaUra:
### What does this PR do?

fix https://github.com/oven-sh/bun/issues/6827

When `Bun.write("path-to-file", new Response(""))`, it has been modified to generate an empty file.

- [x] Code changes

I wrote automated tests

I am new to Bun and zig so I am not sure if I am handling it properly. If I need free, I would be very grateful if you could tell me how to do it.


src/bun.js/webcore/blob.zig
```diff
@@ -708,14 +708,43 @@ pub const Blob = struct {
     ) JSC.JSValue {
         const destination_type = std.meta.activeTag(destination_blob.store.?.data);
 
-        // Writing an empty string to a file is a no-op
         if (source_blob.store == null) {
-            destination_blob.detach();
-            return JSC.JSPromise.resolvedPromiseValue(ctx.ptr(), JSC.JSValue.jsNumber(0));
+            return switch (destination_type) {
+                .file => {
+                    // when writing empty blob to file, we just create an empty file
```
Jarred-Sumner, line 714: Since this is creating an empty file, we only need to do `openat(FD_CWD, path, O_TRUNC | O_CREAT)`

dylan-conway: we could probably reuse `writeStringToFileFast` here, similar to how it is used earlier in `Blob.writeFile`

YutaUra: Hi. Thanks for the advice, it's very helpful!
I refactored it to use `writeStringToFileFast`! What do you think?

Jarred-Sumner:
This was fixed some time ago

Issue #6804, opened 2023-10-30
fs.utimesSync sets the time in seconds (so it seems)


maor-benami:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

### What steps can reproduce the bug?

```javascript
fs.utimesSync(pathname, atime, mtime)
console.log(fs.statSync(pathname).mtime.valueOf())
```

result always ends with "000" with no milliseconds

### What is the expected behavior?

mtime should be set precisely

### What do you see instead?

tested it also with node v18.17.1 and i get the correct results

vitalspace:
I don't know if it is a reliable alternative, but for now it seems that ctime does return milliseconds. 

console.log(fs.statSync(filePath).ctime);

![Screenshot_1](https://github.com/oven-sh/bun/assets/29004070/9d4b931f-37d0-4ca4-af3d-d6c8bdda37df)

PR #6876, opened 2023-11-03
Fix TimeLike implementation to preserve milliseconds in fs.utimesSync.

YutaUra:
### What does this PR do?

fix https://github.com/oven-sh/bun/issues/6804

The millisecond portion of the data was lost in fs.utimesSync.

The TimeLike implementation only kept seconds. For example, Node.js [`process.hrtime.bigint()`](https://nodejs.org/dist/latest-v20.x/docs/api/process.html#processhrtimebigint) returns the current time in nanoseconds, so I modified it to keep the data in two parts, the seconds part and the nanoseconds part, for future implementation.

- [x] Code changes

I wrote automated tests

  - maybe yes

I am new to Bun and zig so I am not sure if I am handling it properly. If I need free, I would be very grateful if you could tell me how to do it.

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -890,7 +890,7 @@ pub const Arguments = struct {
                 return null;
             };
 
-            const atime = JSC.Node.timeLikeFromJS(ctx, arguments.next() orelse {
+            const atime = JSC.Node.TimeLike.fromJS(ctx, arguments.next() orelse {
                 if (exception.* == null) {
                     JSC.throwInvalidArguments(
                         "atime is required",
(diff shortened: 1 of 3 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -4924,14 +4924,12 @@ pub const NodeFS = struct {
 
         var times = [2]std.c.timeval{
             .{
-                .tv_sec = args.mtime,
-                // TODO: is this correct?
-                .tv_usec = 0,
+                .tv_sec = args.mtime.seconds,
+                .tv_usec = args.mtime.nanos * 1000,
```
dylan-conway, line 4928: This could overflow very easily. Maybe change `nanos` to an `isize` and truncate?

YutaUra: @dylan-conway 
Thanks for the advice. You are exactly right. I am now re-implementing it in a way that uses i64, but it is taking a bit of time to fix.

Issue #4517, opened 2023-09-06
[HTTP server] req.signal.onabort not called when HTTP request is closed


A-D-E-A:
### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Microsoft Windows NT 10.0.25931.0 x64 > Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 unknown

### What steps can reproduce the bug?

The `onabort` handler (or `addEventListener("abort", ...)`) set on the request signal is not called when ending a request to which a response has been sent, unless we manually check the `aborted` property.

Run the server below.

```ts
Bun.serve({
	port: 2000,
	fetch(req) {
		const { readable, writable } = new TransformStream();
		const encoder = new TextEncoder();
		const writer = writable.getWriter();

		setInterval(() => {
			const bytes = encoder.encode("Hello!\n");
			const bytesWritten = bytes.length;
			writer.write(bytes);
			const bytes_needed = 4096 - (bytesWritten % 4096);
			const BACKSPACE = 8;
			writer.write(new Uint8Array(bytes_needed).fill(BACKSPACE));
		}, 1000);

		req.signal.onabort = () => {
			console.log("Aborted!");
		};

		return new Response(readable, {
			headers: {
				"Content-Type": "text/event-stream",
			},
		});
	},
});
```

In another console,
```sh
curl http://localhost:2000
```

### What is the expected behavior?

Bun terminal
```sh
> bun run index.ts
# 2.5 seconds later >
Aborted!
```

Curl terminal
```sh
> curl http://localhost:2000
Hello!
Hello!
# After 2.5 seconds, cancel
^C
>
```

### What do you see instead?

Bun terminal
```sh
> bun run index.ts
# Blank, nothing is logged
```

Curl terminal
```sh
> curl http://localhost:2000
Hello!
Hello!
# After 2.5 seconds, cancel
^C
>
```

### Additional information

It works if you cancel the request before it has been sent:
```ts
return new Promise((resolve) => {
	setTimeout(() => {
		resolve(new Response(readable, {
			headers: { "Content-Type": "text/event-stream" }
		});
	, 1000);
});
```
(cancel the request before 1sec and the `onabort` handler is called)

It also works if you read the `aborted` property.
```ts
req.signal.onabort = () => console.log("Aborted!");
setInterval(() => req.signal.aborted /* do nothing */, 2000);
```
[...]

PR #6897, opened 2023-11-04
fix: correctly call abort handler when request is aborted

liz3:
We where not protecting the signal js object on the request when it was requested, leading to the garbage collection of it, preventing handlers like `on("abort")` and `signal.onabort = ...` from being executed on the JS side.
This fixes that by storing the js value and unprotecting it when required.

Fixes: https://github.com/oven-sh/bun/issues/4517

### What does this PR do?

- [x] Code changes


Jarred-Sumner (changes requested):
Why is this going through the JSValue and adding it to the Request struct (which also already exists as signalCachedValue) instead of using the AbortSignal which exists on the Request struct already?

PR #6900, opened 2023-11-04
Add a new argument to the bash installer to be able to skip editing the users rc files (#6901)

AlexAegis:
### What does this PR do?

Skips installing completions if BUN_INSTALL is already defined, assuming that it's already done

~~The current [install script](https://github.com/oven-sh/bun/blob/main/src/cli/install.sh) always edits my `~/.zshrc` file, which I don't want it to do as I prefer to set completions/environment variables/etc myself.~~

~~This PR adds a new cli option to `install.sh` (curl -fsSL https://bun.sh/install | bash)
`--skip-rc`. It just skips the section at the end where it would install completions/edit rc files.~~

~~The existing check that disallowed adding more than 2 options was moved to the argument interpreter, and only limits positional arguments, not flags. Examples of this behavior:~~

```sh
src/cli/install.sh foo bar # okay
src/cli/install.sh foo bar zed # same error as before
src/cli/install.sh --skip-rc foo bar # okay
src/cli/install.sh -vq --skip-rc --another-thing foo bar # okay
src/cli/install.sh -vq foo bar zed # still an error
```

> ~~Note that the argument parser errors for unknown parameters so the bottom two examples would error out on that note first. But this would be the behavior if they would be defined.~~

### Prior art

~~This is used in other similar install scripts for example in [jabba](https://github.com/shyiko/jabba/blob/master/install.sh) a Java version manager.~~

~~The argument parser was taken from [a shell script I wrote](https://github.com/AlexAegis/pont/blob/master/pont.sh#L683) where the aim was compatibility across Linux/BSD/macOS so using getopt(s) was a no-go. Not to mentions both getopt and getopts has their own problems that prompted me to write a parser myself.~~

~~I tried it out. I didn't find any pre-existing test for this install script.
I used a similar format in which this script will be used in (piped it into bash too, besides just running the script)~~

```sh
cat src/cli/install.sh | bash -s -- --skip-rc
```

### DIFF
--- a/src/cli/install.sh
+++ b/src/cli/install.sh
@@ -150,14 +150,6 @@ tildify() {
 
 success "bun was installed successfully to $Bold_Green$(tildify "$exe")"
 
-if command -v bun >/dev/null; then
-    # Install completions, but we don't care if it fails
-    IS_BUN_AUTO_UPDATE=true $exe completions &>/dev/null || :
-
-    echo "Run 'bun --help' to get started"
-    exit
-fi
-
 refresh_command=''
 
 tilde_bin_dir=$(tildify "$bin_dir")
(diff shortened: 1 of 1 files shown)


AlexAegis:
I recommend hiding whitespace changes in the diff, I had to put a chunk of the script in a function. (And sorry for the close/reopen, I think that was github, I was messing around rebasing my fork)

paperclover (changes requested):
putting an entire fully-featured argument parser just to parse one flag is overkill (imo the wrong approach). this file needs to be small and readable.

in my opinion we should simply detect if the `BUN_INSTALL` environment variable was set then assume that shell configuration was already done. we can also add `--skip-rc` on top of that but i think your use case would be solved with auto-detection

AlexAegis:
@paperdave I added your suggested changes

I also removed the other "bun --help" echo that was only there because of an early exit, now the function that installs the completions just returns early, and the same "bun --help" message is used that was already there at the end of the script

> putting an entire fully-featured argument parser just to parse one flag is overkill (imo the wrong approach).

I respect this, that's why I simplified the PR, but my stance on this is that while for one arg it's indeed overkill, it opens up the door for other flags that might be hard to implement without some groundwork. You can already see signs of this being needed from the way errors are handled around having too many args, if you'd try to introduce an optional flag, that kind of check will no longer be useful.

src/cli/install.sh
```diff
@@ -169,129 +161,141 @@ fi
...
+        ;;
+    *)
+        echo 'Manually add the directory to ~/.bashrc (or similar):'
+        info_bold "  export $install_env=$quoted_install_dir"
+        info_bold "  export PATH=\"$bin_env:\$PATH\""
+        ;;
+    esac
+}
+
+if [ -z ${!install_env} ]; then
```
paperclover, line 296: this syntax is valid? shouldn't this be `if [ -z "${install_env}" ]`?

paperclover (commented):
this is much better. i'll do some of my own testing to ensure this script works as expected later.

src/cli/install.sh
AlexAegis: This is an indirect variable reference, I'm checking if BUN_INSTALL is defined or not. `install_env` is always defined as it's defined in this script with a default.

I could use `BUN_INSTALL` directly, but it was only mentioned once, when reading the default value for `install_env`, and I didn't dare write a variables value twice.

AlexAegis: I didn't add double quotes as it's referring to an env variables name and that should be a continuous string, not expected to split

PR #6908, opened 2023-11-05
Fix libuv initialization code

Jarred-Sumner:
### What does this PR do?

On windows, we were creating 2 uv loops instead of 1

Did not


paperclover:
Build does not compile on Windows

paperclover:
merged into in #6938

PR #6913, opened 2023-11-05
Fix build on arch and fix `bun x` to execute local binaries from root package.json

tr1ckydev:
### What does this PR do?

This allows `bun x` command to search first for local binaries mentioned in the root package.json and execute them if found similar to npx, which wasn't possible earlier.

For example we have a package.json in our root dir
```json
{
  "name": "cli",
  "module": "index.ts",
  "type": "module",
  "bin": "./index.ts",
  "devDependencies": {
    "bun-types": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  }
}
```
Previously `bun x cli` would not execute this binary but `npx cli` worked. This PR allows `bun x cli` to work as expected and to execute local binaries from root package.json.

Also fixes build not working on arch distros having different names.

- [x] Code changes

Manual tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,7 +214,7 @@ endif()
 
 set(DEFAULT_USE_STATIC_LIBATOMIC ON)
 if(UNIX AND NOT APPLE)
-    execute_process(COMMAND cat /etc/os-release COMMAND head -n1 OUTPUT_VARIABLE LINUX_DISTRO)
+    execute_process(COMMAND cat /usr/lib/os-release COMMAND head -n1 OUTPUT_VARIABLE LINUX_DISTRO)
     if(${LINUX_DISTRO} STREQUAL "NAME=\"Arch Linux\"\n")
         set(DEFAULT_USE_STATIC_LIBATOMIC OFF)
     endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -214,7 +214,7 @@ endif()
 
 set(DEFAULT_USE_STATIC_LIBATOMIC ON)
 if(UNIX AND NOT APPLE)
-    execute_process(COMMAND cat /etc/os-release COMMAND head -n1 OUTPUT_VARIABLE LINUX_DISTRO)
+    execute_process(COMMAND cat /usr/lib/os-release COMMAND head -n1 OUTPUT_VARIABLE LINUX_DISTRO)
```
Hanaasagi, line 217: > Also fixes build not working on arch distros having different names.

Which distribution has this issue?

afak, `/etc/os-release` has a higher priority.

> The file /etc/os-release takes precedence over /usr/lib/os-release. Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if it is missing. 

Link: https://man.archlinux.org/man/os-release.5.en

tr1ckydev: I have CachyOS Linux currently which doesn't return Arch Linux for the previous one. Hence the former change.

PR #6938, opened 2023-11-06
windows: more windows stuff

paperclover:
### What does this PR do?

this message is out of date but:

- Ban `bun install` for the time being.
  - It is now a runtime panic to initWithCLI or invoke the installation code.
  - Auto-Install is set to never be used.
- Do not treat absolute file paths as package names.
- `bun upgrade`: do not show status bar for Expand-Archive cmdlet. This is done because the archive extraction is very quick on most hardware, and it is unpleasant to see a 10 frame flash of blue on the top 5 lines of your terminal.
- Make Bundler Env Map case insensitive.
- Reimplemented `which` for windows.
  - `bun.sys.isExecutableFilePath` may not be called when compiling to windows due to different semantics of executable file paths.
- Implement some FS functions for windows (untested, but readFileSync works)


src/cli/run_command.zig
```diff
@@ -363,12 +363,21 @@ pub const RunCommand = struct {
...
                         }
                     }
                 }
             }
             Output.prettyErrorln("<r><red>error<r>: Failed to run \"<b>{s}<r>\" due to error <b>{s}<r>", .{ std.fs.path.basename(executable), @errorName(err) });
+            if (@errorReturnTrace()) |trace| {
+                std.debug.dumpStackTrace(trace.*);
+            }
+            if (Environment.isDebug) {
+                Output.prettyErrorln("Full path: <b>{s}<r>", .{executable});
```
Jarred-Sumner: can you use `log` here so it can be disabled

src/resolver/resolver.zig
```diff
@@ -49,7 +49,11 @@ const DotEnv = @import("../env_loader.zig");
 
 pub fn isPackagePath(path: string) bool {
```
Jarred-Sumner, line 49: This should be split into two functions

```js
isPackagePath
isPackagePathNotAbsolute
```

and the version used in resolver.zig should be `isPackagePathNotAbsolute`

To avoid the duplicate work of checking if its an absolute path twice

src/which.zig
```diff
@@ -44,8 +52,76 @@ pub fn which(buf: *[bun.MAX_PATH_BYTES]u8, path: []const u8, cwd: []const u8, bi
...
+        if (bun.strings.eqlComptimeCheckLenWithType(u8, file_ext, ext, false)) return true;
+    }
+    return false;
+}
+
+/// This is the windows version of `which`.
+/// It operates on wide strings, and resolves .exe / .cmd extensions.
+/// It is similar to Get-Command in powershell.
+pub fn whichW(buf: *[bun.MAX_PATH_BYTES / 2]u16, path: []const u8, cwd: []const u8, bin: []const u8) ?[:0]const u16 {
+    _ = cwd;
```
Jarred-Sumner, line 77: This should be joined with the `bin` as the first check so that `./node_modules/.bin/prettier` works to become `C:\Users\dave\app\node_modules\.bin\prettier`

src/windows.zig
```diff
@@ -58,7 +58,7 @@ pub usingnamespace ntdll;
 pub const user32 = windows.user32;
 pub const advapi32 = windows.advapi32;
 
-pub const INVALID_FILE_ATTRIBUTES = -1;
+pub const INVALID_FILE_ATTRIBUTES: u32 = std.math.maxInt(u32);
```
Jarred-Sumner, line 61: is this right

src/string_immutable.zig
```diff
@@ -4948,3 +4947,31 @@ pub fn concatIfNeeded(
     }
     std.debug.assert(remain.len == 0);
 }
+
+pub fn convertUTF8toUTF16InBuffer(
+    buf: []u16,
+    input: []const u8,
+) []const u16 {
+    if (!Environment.enableSIMD) {
```
Jarred-Sumner: ✂️ these functions always support simd

src/string_immutable.zig
```diff
@@ -4948,3 +4947,31 @@ pub fn concatIfNeeded(
...
+        .success => return buf[0..result.count],
+        else => @panic("TODO: handle error in convertUTF8toUTF16"),
+    }
+}
+
+pub fn convertUTF16toUTF8InBuffer(
+    buf: []u8,
+    input: []const u16,
+) ![]const u8 {
+    if (!Environment.enableSIMD) {
```
Jarred-Sumner: ✂️ these functions always support simd

src/install/install.zig
```diff
@@ -6710,6 +6713,10 @@ pub const PackageManager = struct {
         comptime op: Lockfile.Package.Diff.Op,
         comptime subcommand: Subcommand,
     ) !void {
+        if (Environment.isWindows and !Environment.isDebug) {
+            @panic("Bun install does not support Windows");
```
```suggestion
            @panic("Windows support for bun install is not implemented yet");
```

src/install/install.zig
```diff
@@ -5673,6 +5673,9 @@ pub const PackageManager = struct {
         cli: CommandLineArguments,
         comptime subcommand: Subcommand,
     ) !*PackageManager {
+        if (Environment.isWindows and !Environment.isDebug) {
+            @panic("Bun install does not support Windows");
```
```suggestion
            @panic("Windows support for bun install is not implemented yet");
```

src/bun.zig
```diff
@@ -882,6 +882,46 @@ pub const StringArrayHashMapContext = struct {
     };
 };
 
+pub const CaseInsensitiveASCIIStringContext = struct {
+    pub fn hash(_: @This(), str_: []const u8) u32 {
+        var str = str_;
+        var wyhash = std.hash.Wyhash.init(0);
+        var buf: [1024]u8 = undefined;
```
Jarred-Sumner: std.hash.Wyhash.init is slower than `.hash`, maybe worth a check here

CMakeSettings.json
```diff
@@ -0,0 +1,15 @@
+﻿{
```
Jarred-Sumner: what is this?

src/__global.zig
```diff
@@ -143,6 +143,9 @@ pub fn notimpl() noreturn {
 // Make sure we always print any leftover
 pub fn crash() noreturn {
     @setCold(true);
+    if (Environment.isDebug) {
+        @panic("Global.crash()");
```
Jarred-Sumner: I think you want `@breakpoint` and not `@panic`.
```suggestion
        @breakpoint();
```

CMakeSettings.json
paperclover: probably can remove. it is used by MS Visual Studio's built-in CMake support.

src/__global.zig
paperclover: This was added to print a stack trace when a debug build is invoked without a debugger. This is a pattern i'm using in some places to make certain errors panic in debug, as tracing the message to the source code takes time and being able to click and see where the message is coming from is more productive. In a release, crash still acts as this "quiet panic"

- panic will also break on the callsite
- breakpoint will not print this info but break immediatly, or otherwise exit without a printed trace

src/windows.zig
paperclover: Yes, the return value is a u32 as per ms docs, but using -1 is out of range so the comparison results in comptime false.

fun fact, it's an intended feature of zig for this mistake to not be a compiler error... https://github.com/ziglang/zig/issues/17853

src/deps/uws.zig
```diff
@@ -2413,6 +2413,12 @@ pub const UVLoop = extern struct {
     pre: *uv.uv_prepare_t,
     check: *uv.uv_check_t,
 
+    pub fn init() *UVLoop {
+        return uws_get_loop_with_native(bun.windows.libuv.Loop.get());
```
Jarred-Sumner, line 2437: is libuv's default loop threadlocal or global? If it's threadlocal, this is fine. if it's not, we need to change it

Jarred-Sumner:
serve test is segfaulting

src/__global.zig
Jarred-Sumner: I still dont' think this should be done. It'll mess up tests and differs a lot from release build behavior. Can you add an environment variable that is disabled by default?

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -514,42 +556,42 @@ void ImportMetaObject::finishCreation(VM& vm)
...
         ImportMetaObject* meta = jsCast<ImportMetaObject*>(init.owner);
-        WTF::URL url = meta->url.startsWith('/') ? WTF::URL::fileURLWithFileSystemPath(meta->url) : WTF::URL(meta->url);
+        WTF::URL url = isAbsolutePath(meta->url) ? WTF::URL::fileURLWithFileSystemPath(meta->url) : WTF::URL(meta->url);
 
         init.set(jsString(init.vm, url.string()));
     });
     this->dirProperty.initLater([](const JSC::LazyProperty<JSC::JSObject, JSC::JSString>::Initializer& init) {
         ImportMetaObject* meta = jsCast<ImportMetaObject*>(init.owner);
 
-        WTF::URL url = meta->url.startsWith('/') ? WTF::URL::fileURLWithFileSystemPath(meta->url) : WTF::URL(meta->url);
```
Jarred-Sumner, line 524: are you sure

src/bun.js/javascript.zig
```diff
@@ -2870,7 +2893,7 @@ pub const VirtualMachine = struct {
 
         if (show.fd) {
             if (show.syscall) {
-                try writer.writeAll("   ");
+                try writer.writeAll("     ");
```
Jarred-Sumner: I doubt this is right

src/bun.js/node/dir_iterator.zig
```diff
@@ -18,305 +19,330 @@ const mem = std.mem;
...
+                                    },
+                                };
+                            }
 
-                    if (rc != .SUCCESS) {
-                        if ((bun.windows.Win32Error.fromNTStatus(rc).toSystemErrno())) |errno| {
                             return .{
                                 .err = .{
-                                    .errno = @truncate(@intFromEnum(errno)),
+                                    .errno = @intFromEnum(bun.C.SystemErrno.EUNKNOWN),
```
Jarred-Sumner, line 233: Why aren't we converting the errno?

Jarred-Sumner: oh we are above okay

src/bun.js/webcore/streams.zig
```diff
@@ -4104,7 +4101,8 @@ pub const FIFO = struct {
         this: *FIFO,
         buf: []u8,
     ) ReadResult {
-        switch (Syscall.read(this.fd, buf)) {
+        std.debug.print("DO READ\n", .{});
```
Jarred-Sumner:
```suggestion

```

src/bun.js/webcore/streams.zig
```diff
@@ -4104,7 +4101,8 @@ pub const FIFO = struct {
...
+        switch (Syscall.sys_uv.read(this.fd, buf)) {
```
Jarred-Sumner: did you mean to keep this here?

src/bun.js/webcore/streams.zig
```diff
@@ -4244,27 +4242,27 @@ pub const File = struct {
 
         if (file.pathlike != .path and !(file.is_atty orelse false)) {
             if (comptime Environment.isWindows) {
-                bun.todo(@src(), {});
```
Jarred-Sumner, line 4246: why remove `bun.todo`? It logs where it happened

src/cli/init_command.zig
```diff
@@ -41,6 +41,9 @@ pub const InitCommand = struct {
 
         const input = try std.io.getStdIn().reader().readUntilDelimiterAlloc(alloc, '\n', 1024);
         if (input.len > 0) {
+            if (strings.endsWithChar(input, '\r')) {
```
Jarred-Sumner: what if the only character is a `\r`?

Jarred-Sumner:
did you mean to update boringssl, libarchive, c-ares, and the rest? i don't think you did.

src/bun.js/webcore/streams.zig
paperclover: it, and many others, were converted to `@panic("TODO on Windows")`, which gives a full trace and stops execution, which is more desired in this case because by ignoring the error you get into weird state like the zero length buffer. not as relevant now as the todo log is no longer disabled by default, but i like having these full-stop panic todos in these places.

src/__global.zig
paperclover: that is reasonable.

src/bun.js/bindings/ImportMetaObject.cpp
paperclover: Absolute paths on windows do not always start with `/`

the replaced method `isAbsolutePath` calls `.startsWith('/')` on unix and does the proper check on windows.

paperclover:
i did mean to upgrade these dependencies, but that could be the instability observed and i think i should downgrade them back; we can upgrade our dependencies separately.

src/bun.js/webcore/streams.zig
paperclover: no i was trying to figure out where reading was happening because the todolog was disabled by default

src/cli/init_command.zig
paperclover: then it will be incorrectly an empty slice. will fix

src/bun.js/javascript.zig
paperclover: latest bun on linux:

image

i realize we dont actually attach fd to alot of our errors, i just added it to macos so beforehand this isnt shown but now it is, and at the right alignment

image

actually, better question, should fd be placed at the end as syscall and errno are more relevant

src/bun.js/api/bun/subprocess.zig
```diff
@@ -24,7 +26,7 @@ pub const Subprocess = struct {
     pub usingnamespace JSC.Codegen.JSSubprocess;
     const default_max_buffer_size = 1024 * 1024 * 4;
 
-    pid: std.os.pid_t,
+    pid: if(Environment.isWindows) c_int else std.os.pid_t,
     // on macOS, this is nothing
```
dylan-conway, line 31: should this be `uv.pid_t` for windows?

src/bun.js/bindings/ProcessBindingConstants.cpp
```diff
@@ -671,8 +675,11 @@ static JSValue processBindingConstantsGetFs(VM& vm, JSObject* bindingObject)
 #ifdef O_EXCL
     object->putDirect(vm, PropertyName(Identifier::fromString(vm, "O_EXCL"_s)), jsNumber(O_EXCL));
 #endif
+#if OS(WINDOWS)
+    object->putDirect(vm, PropertyName(Identifier::fromString(vm, "UV_FS_O_FILEMAP"_s)), jsNumber(536870912));
```
dylan-conway, line 679: `node_fs_constants:143` needs this change too

src/bun.js/node/node_fs.zig
```diff
@@ -3990,7 +4046,7 @@ pub const NodeFS = struct {
     pub fn chown(this: *NodeFS, args: Arguments.Chown, comptime flavor: Flavor) Maybe(Return.Chown) {
         _ = flavor;
         if (comptime Environment.isWindows) {
-            return Maybe(Return.Fchmod).todo;
+            return Syscall.chown(args.path.sliceWithWinNormalizerCWDZ(&this.sync_error_buf), args.uid, args.gid);
```
dylan-conway: Maybe `sliceWithWinNormalizerCWDZ` and `sliceZ` should become one function that checks `isWindows`

src/bun.js/node/types.zig
```diff
@@ -2271,30 +2293,95 @@ pub const Path = struct {
...
-        const path_name = Fs.NodeJSPathName.init(
-            path,
-            if (win32) std.fs.path.sep_windows else std.fs.path.sep_posix,
-        );
-        var dir = JSC.ZigString.init(path_name.dir);
-        const is_absolute = (win32 and dir.len > 0 and isZigStringAbsoluteWindows(dir)) or (!win32 and path.len > 0 and path[0] == '/');
+
+        const is_absolute =
+            switch (win32) {
+            true => path.len > 0 and std.fs.path.isAbsoluteWindows(path),
```
dylan-conway: `path.len > 0` is checked in both functions

src/cli/init_command.zig
```diff
@@ -39,7 +39,10 @@ pub const InitCommand = struct {
 
         Output.flush();
 
-        const input = try std.io.getStdIn().reader().readUntilDelimiterAlloc(alloc, '\n', 1024);
+        var input = try std.io.getStdIn().reader().readUntilDelimiterAlloc(alloc, '\n', 1024);
```
dylan-conway, line 42: it would be nice if there was a readline like method

src/resolver/resolve_path.zig
```diff
@@ -334,24 +273,67 @@ threadlocal var relative_to_common_path_buf: [4096]u8 = undefined;
...
-    const last_common_separator = strings.lastIndexOfChar(_common_path, separator) orelse 0;
+    const shortest = @min(normalized_from.len, normalized_to.len);
 
     if (shortest == common_path.len) {
-        if (normalized_to.len > normalized_from.len) {
+        if (normalized_to.len >= normalized_from.len) {
             if (common_path.len == 0) {
+                if (platform == .windows and
+                    normalized_to.len > 3 and
+                    normalized_to[normalized_to.len - 1] == separator)
```
dylan-conway, line 332:
```suggestion
    const common_path = if (win_root_len) |root_len|
```

src/resolver/resolve_path.zig
```diff
@@ -581,8 +716,9 @@ pub const Platform = enum {
         return switch (comptime platform) {
             .auto => (comptime platform.resolve()).isAbsolute(path),
             .posix => path.len > 0 and path[0] == '/',
-            .windows => std.fs.path.isAbsoluteWindows(path),
-            .loose => isAbsolute(.posix, path) or isAbsolute(.windows, path),
+            .windows,
+            .loose,
```
dylan-conway, line 720: windows applies to both?

src/main.zig
```diff
@@ -29,6 +29,9 @@ pub fn main() void {
         bun.win32.STDOUT_FD = bun.toFD(std.io.getStdOut().handle);
         bun.win32.STDERR_FD = bun.toFD(std.io.getStdErr().handle);
         bun.win32.STDIN_FD = bun.toFD(std.io.getStdIn().handle);
+
+        // This fixes printing unicode characters
+        _ = std.os.windows.kernel32.SetConsoleOutputCP(65001);
```
dylan-conway, line 34: is there a constant in the standard library we can use?

paperclover: image
only in c++ :(

src/resolver/resolve_path.zig
paperclover: yes. you can think of isAbsoluteWindows as a superset of isAbsolutePosix

src/c.zig
```diff
@@ -105,29 +105,31 @@ pub fn lstat_absolute(path: [:0]const u8) !Stat {
...
-    switch (bun.sys.renameat(from_dir, filename, to_dir, destination)) {
+    switch (bun.sys.renameat(bun.toFD(from_dir), filename, bun.toFD(to_dir), destination)) {
         .err => |err| {
             // allow over-writing an empty directory
             if (err.getErrno() == .ISDIR) {
-                _ = bun.sys.rmdirat(to_dir, destination.ptr);
+                _ = bun.sys.rmdirat(bun.toFD(to_dir), destination.ptr);
 
-                try (bun.sys.renameat(from_dir, filename, to_dir, destination).unwrap());
+                try (bun.sys.renameat(bun.toFD(from_dir), filename, bun.toFD(from_dir), destination).unwrap());
```
dylan-conway: is the third arg supposed to be `bun.toFD(to_dir)`?

src/sys.zig
```diff
@@ -404,7 +415,13 @@ pub fn openDirAtWindows(
     );
 
     if (comptime Environment.allow_assert) {
-        log("NtCreateFile({d}, {}, ) = {d} (dir) = {d}", .{ dirFd, bun.strings.fmtUTF16(path), rc, @intFromPtr(fd) });
+        log("NtCreateFile({d}, {}) = {d} (dir) = {d}", .{ dirFd, bun.strings.fmtUTF16(path), rc, @intFromPtr(fd) });
+    }
+
+    if (Environment.isDebug) {
+        if (rc == std.os.windows.NTSTATUS.OBJECT_PATH_SYNTAX_BAD) {
```
dylan-conway: `std.debug.assert(rc != std.os.windows.NTSTATUS.OBJECT_PATH_SYNTAX_BAD)`?

src/sys.zig
```diff
@@ -517,7 +539,13 @@ pub fn openatWindows(dirfD: bun.FileDescriptor, path_: []const u16, flags: bun.M
         );
 
         if (comptime Environment.allow_assert) {
-            log("NtCreateFile({d}, {}) = {d} (file) = {d}", .{ dirfD, bun.strings.fmtUTF16(path), rc, @intFromPtr(result) });
+            log("NtCreateFile({d}, {}) = {d} (file) = {d}", .{ dirfd, bun.strings.fmtUTF16(path), rc, @intFromPtr(result) });
+        }
+
+        if (Environment.isDebug) {
+            if (rc == std.os.windows.NTSTATUS.OBJECT_PATH_SYNTAX_BAD) {
```
dylan-conway: `std.debug.assert(rc != std.os.windows.NTSTATUS.OBJECT_PATH_SYNTAX_BAD)`?

dylan-conway (approved):

Issue #6915, opened 2023-11-05
v1.0.8+ breaks scoped package import in Docker image


4lch4:
### What version of Bun is running?

v.1.0.8

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

- Create a simple "hello world" app with bun.
- Add the @4lch4/backpack dependency.
  - `bun add @4lch4/backpack`
- Import the package in `index.ts`.
  - `import { logger } from '@4lch4/backpack'`
- Add a Docker file that uses `FROM oven/bun:latest` at the start.
- Build & run the Docker image.
  - `docker build -t 4lch4/bun-test:latest .`
  - `docker run 4lch4/bun-test:latest`

This should give you an error that it can't find the package `@4lch4/backpack`. If you change the `FROM oven/bun:latest` to `FROM oven/bun:1.0.7`, everything works.

### What is the expected behavior?

It should output the hello world message to the console.

### What do you see instead?

The console says that the package could not be found.

### Additional information

This only happens within Docker. When I use the latest version locally with something like `bun run src/index.ts`, it'll work just fine. But as soon as I do the same thing within a Docker image, it can't find my package.

Here are links to the package from above, in case it helps:

- https://github.com/4lch4/Backpack
- https://www.npmjs.com/package/@4lch4/backpack

4lch4:
I created [the following demo repo][0] to help make it easier to reproduce the issue:

https://github.com/4lch4/Bun-Docker-Bug

[0]: https://github.com/4lch4/Bun-Docker-Bug

4lch4:
Thanks for fixing this so quickly!

Issue #6666, opened 2023-10-23
Issue with Vite: Script "build" exited with code 1


janswist:
### What version of Bun is running?

latest Docker one (can't output it)

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

`bunx --bun vite build` inside Dockerfile

### What is the expected behavior?

To build. Outside Docker it all builds properly.

### What do you see instead?

#0 0.257 $ bunx --bun vite build && cp src/assets/_headers dist/_headers                                                                 
#0 0.269  Resolving dependencies                                                                                                         
#0 5.297  Resolved, downloaded and extracted [76]                                                                                        
#0 5.298  Saved lockfile
#0 5.552 failed to load config from /code/vite.config.js
#0 5.552 error during build:
#0 5.552 undefined
#0 5.568 error: script "build" exited with code 1 (SIGHUP)

benglewis:
@Jarred-Sumner Unfortunately, I am seeing the same thing, but in reverse.

The Docker image builds fine on my local M3 Pro MacBook Pro (even though I use `platform=linux/amd64` in the `FROM` section of the `Dockerfile`, so the image being built is using Rosetta, but when I try to build the exact same image on GitHub in GitHub Actions, it fails during the Vite build. I made sure that both systems are using the latest Docker version (26.1.1) and I fixed my `bun` base image version to the latest too (specifically `oven/bun:1.1.8-alpine`). In terms of other context that may help to reproduce/debug the issue, I am using SvelteKit which inside is using Vite with the `@sveltejs/adapter-static` adapter.

My issue also seems related to [10811](https://github.com/oven-sh/bun/issues/10811)

Should I open a separate ticket and try to provide a simple codebase that reproduces the issue?

PR #6942, opened 2023-11-07
Probably fix linux arm64 issues

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/6915
Fixes https://github.com/oven-sh/bun/issues/6666
Fixes https://github.com/oven-sh/bun/issues/6918
Fixes https://github.com/oven-sh/bun/issues/5943 (probably)

Supersedes https://github.com/oven-sh/bun/pull/6398

~Need to test if this works as expected~ It does

### DIFF
--- a/src/sys.zig
+++ b/src/sys.zig
@@ -281,7 +281,7 @@ pub fn getErrno(rc: anytype) bun.C.E {
         return bun.C.E.UNKNOWN;
     }
 
-    if (comptime use_libc) return std.os.errno(rc);
+    if (comptime Environment.isMac) return std.os.errno(rc);
     const Type = @TypeOf(rc);
 
     return switch (Type) {


src/sys.zig
```diff
@@ -281,7 +281,7 @@ pub fn getErrno(rc: anytype) bun.C.E {
         return bun.C.E.UNKNOWN;
     }
 
-    if (comptime use_libc) return std.os.errno(rc);
+    if (comptime Environment.isMac) return std.os.errno(rc);
```
tarruda, line 284: Note that before this patch, Mac was also falling under the `@TypeOf` check below, which means it was calling `std.os.linux.getErrno` to properly handle `usize` rc.

Does `std.os.errno` have a different implementation on Mac than on Linux?  If not,and you pass an unsigned integer, it might crash because of the `integer cast truncated bits` assertion (https://github.com/oven-sh/bun/pull/6706#issue-1961666697). 

Note that all of the linked issues are Linux/arm64 issues, and only happen on Mac because Docker on Apple silicon runs under  a Linux/arm64 VM. So in the best scenario this condition is not doing anything, in the worst case it is moving the bug to Mac.

jakkunight:
1.0.10 Fixed the bug for me. I tried again today and it seems to work correctly as expected.

PR #6991, opened 2023-11-08
Fix fetch response skipping body if no content-length (#6932)

### LINKED ISSUE #6932: fetch() returns empty body when Response has no Content-Length or Transfer-Encoding header

1.0.9+98f20170a

Darwin 21.6.0 x86_64 i386

*fetchtest.js:*
```js
const path = (globalThis.Bun ?? process).argv[2] ?? '';
const res = await fetch(`http://127.0.0.1:8081${path}`);
const body = await res.text();
console.log(JSON.stringify(body));
```

*webserver.py:* (Sample Python3 http server that doesn't add any header automatically; couldn't manage to set it up in Bun)
```python
from http.server import BaseHTTPRequestHandler, HTTPServer

content = "Hello world"

class MyServer(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header("Content-Type", "text/plain")
        if self.path == "/a":
            self.send_header("Transfer-Encoding", "chunked")
            self.end_headers()
            self.wfile.write(bytes("%x\r\n%s\r\n0\r\n\r\n" % (len(content), content), "utf-8"))
        elif self.path == "/b":
            self.send_header("Content-Length", str(len(bytes(content, "utf-8"))))
            self.end_headers()
            self.wfile.write(bytes(content, "utf-8"))
        else:
            self.end_headers()
            self.wfile.write(bytes(content, "utf-8"))

webserver = HTTPServer(("localhost", 8081), MyServer)
webserver.serve_forever()
```

To reproduce the failing case, start the webserver (`python3 webserver.py`) in a shell and then run `bun fetchtest.js` in another shell.

The stdout output should be `"Hello world"` string in all cases.

### What do you see instead?

When the HTTP Response has no `Content-Length` header, and it is not chunked (has no `Transfer-Encoding: chunked` header), the stdout output is the empty string `""`.

### Additional information

The example programs can repro these 3 cases:
- `bun testfetch.mjs /a`: Content-Length header is set-> 🟢 works ok
- `bun testfetch.mjs /b`: content is chunked -> 🟢 works ok
- `bun testfetch.mjs`: neither header is set ->  🔴  body text is the empty string

The same program when run on NodeJS (`node testfetch.mjs`), prints the correct output in all the cases.

Another side effect of this bug seems to be that `await res.json()` wrongly returns `null` for the failing case when ran on Bun.

According to [RFC 7230 section 3.3.3](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.3) the absence of `Content-Length` and `Content-Transfer` is a valid response under HTTP/1.1. The relevant paragraphs:
> "1. Any response ... with a 1xx (Informational), 204 (No Content), or 304 (Not Modified) status code is always terminated by the first empty line after the header fields, regardless of the header fields present in the message, and thus cannot contain a message body."
> "7. Otherwise, this is a response message without a declared message body length, so the message body length is determined by the number of octets received prior to the server closing the connection."

otgerrogla:
### What does this PR do?

Fixes the case when `fetch()` Response has no `Content-Length` header (and is not chunked) returning an empty body, instead of returning all bytes until the TCP socket is closed by the server, which is the expected behavior for HTTP/1.1 according to [RFC 7230 section 3.3.3](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.3) bullet point 7.

Fixes #6932 and adds tests for the fix, and also for any potential regression on edge cases.

- [x] Code changes

I wrote automated tests

Also ran these tests (sligthly tweaked so they work) on NodeJS, to ensure the behavior is consistent, and all passed on NodeJS v21.1.0.

### DIFF
--- a/test/js/web/fetch/fetch.test.ts
+++ b/test/js/web/fetch/fetch.test.ts
@@ -1841,6 +1841,13 @@ describe("http/1.1 response body length", () => {
             );
           } else if (text.startsWith("GET /empty")) {
             socket.end("HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n");
+          } else if (text.startsWith("GET /keepalive/bad")) {
+            const resp = "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nConnection: keep-alive\r\n\r\nHello, World!";
+            socket.end(`${resp}${resp}`);
+          } else if (text.startsWith("GET /keepalive")) {
+            const resp =
+              "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nConnection: keep-alive\r\nContent-Length: 13\r\n\r\nHello, World!";
+            socket.end(`${resp}${resp}`);
           } else {
             socket.end(`HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nContent-Length: 13\r\n\r\nHello, World!`);
           }
@@ -1869,6 +1876,22 @@ describe("http/1.1 response body length", () => {
       expect(response.status).toBe(200);
       expect(response.json<unknown>()).resolves.toEqual({ "hello": "World" });
     });
+
+    it("should disable keep-alive", async () => {
+      // according to http/1.1 spec, the keep-alive persistence behavior should be disabled when
+      // "Content-Length" header is not set (and response is not chunked)
+      // therefore the response text for this test should contain
+      // the 1st http response body + the full 2nd http response as text
+      const response = await fetch(`http://${getHost()}/keepalive/bad`);
+      expect(response.status).toBe(200);
+      expect(response.text()).resolves.toHaveLength(95);
+    });
+  });
+
+  it("should support keep-alive", async () => {
+    const response = await fetch(`http://${getHost()}/keepalive`);
+    expect(response.status).toBe(200);
+    expect(response.text()).resolves.toBe("Hello, World!");
   });
 
   it("should support transfer-encoding: chunked", async () => {
--- a/test/js/web/fetch/fetch.test.ts
+++ b/test/js/web/fetch/fetch.test.ts
@@ -1,4 +1,4 @@
-import { AnyFunction, serve, ServeOptions, Server, sleep } from "bun";
+import { AnyFunction, serve, ServeOptions, Server, sleep, TCPSocketListener } from "bun";
 import { afterAll, afterEach, beforeAll, describe, expect, it, beforeEach } from "bun:test";
 import { chmodSync, mkdtempSync, readFileSync, realpathSync, rmSync, writeFileSync } from "fs";
 import { mkfifo } from "mkfifo";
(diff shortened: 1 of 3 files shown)


test/js/web/fetch/fetch.test.ts
```diff
@@ -1816,3 +1816,84 @@ it("304 not modified with 0 content-length does not cause a request timeout", as
...
+                `HTTP/1.1 ${status} STATUS MESSAGE\r\nContent-Type: text/plain\r\nContent-Length: 13\r\n\r\nHello, World!`,
+              );
+            }
+          }
+        }
+      });
+    });
+
+    const { promise: start, resolve } = Promise.withResolvers();
+    server.listen(8080, resolve);
```
Electroid: To prevent flaky tests, we should make this change, so it picks a random, open port:
```suggestion
    server.listen(0, resolve);
```

test/js/web/fetch/fetch.test.ts
```diff
@@ -1816,3 +1816,84 @@ it("304 not modified with 0 content-length does not cause a request timeout", as
...
+    await start;
+  });
+  afterAll(() => {
+    server?.close?.();
+  });
+
+  const getHost = () => `127.0.0.1:${(server!.address() as net.AddressInfo).port}`;
```
Electroid: Instead of hard-coded `127.0.0.1`, use the hostname from the `AddressInfo`.

test/js/web/fetch/fetch.test.ts
```diff
@@ -1816,3 +1816,84 @@ it("304 not modified with 0 content-length does not cause a request timeout", as
...
+      const response = await fetch(`http://${getHost()}/json`);
+      expect(response.status).toBe(200);
+      expect(await response.json<unknown>()).toEqual({ "hello": "World" });
+    });
+  });
+
+  it("should support transfer-encoding: chunked", async () => {
+    const response = await fetch(`http://${getHost()}/chunked`);
+    expect(response.status).toBe(200);
+    expect(await response.text()).toBe("Hello, World!");
```
Electroid: `expect(await ...).toX()` can be changed to `expect(...).resolves.toX()`

test/js/web/fetch/fetch.test.ts
```diff
@@ -1816,3 +1816,84 @@ it("304 not modified with 0 content-length does not cause a request timeout", as
   expect(await response.arrayBuffer()).toHaveLength(0);
   server.stop(true);
 });
+
+describe("http/1.1 response body length", () => {
+  // issue #6932 (support response without Content-Length and Transfer-Encoding) + some regression tests
+
+  let server: net.Server | undefined;
+  beforeAll(async () => {
+    server = net.createServer(socket => {
```
Electroid: Could we change this to use `Bun.listen()` instead of `net.createServer()`? The former will be faster, and it's consistent with the rest of the tests.

See line 1798 for an example.

test/js/web/fetch/fetch.test.ts
```diff
@@ -1816,3 +1816,84 @@ it("304 not modified with 0 content-length does not cause a request timeout", as
...
+      socket.on("data", data => {
+        const lines = data.toString().split("\r\n");
```
Electroid: This for-loop is confusing to read, since the logic will essentially terminate on the first iteration. Could we change it to something like:

```js
const text = data.toString();
if (text.startsWith("GET /text") {
  // ...
} else if (text.startsWith("GET /json") {
  // ...
}

Electroid (changes requested):
Thanks for working on this! Overall looks good, I did make some suggestions about the tests. I'll still run the test suite

otgerrogla:
Just saw these changes break tests that check that status 304 without `Content-Length` doesn't trigger a timeout.

According to the RFC:
> 1. Any response to a HEAD request and any response with a 1xx (Informational), 204 (No Content), or 304 (Not Modified) status code is always terminated by the first empty line after the header fields, regardless of the header fields present in the message, and thus cannot contain a message body or trailer section
Therefore 

So the code needs to be tweaked in order that when received these specific status codes, it assumes a Content-Length of 0. Will work on it, and implement the changes to the tests stated above (thanks!)

Electroid:
In `test/js/bun/http/serve.test.ts`, the test named "should response with HTTP status code (101)" is failing. But it looks like the test is wrong, given the bug this issue is fixing. It's checking that a response with a status of `101` has a body, but that is not allowed, given the changes.

Electroid (approved):

Jarred-Sumner (changes requested):
If content length is not specified, then we must ensure the connection will not be added to the keep-alive list

Jarred-Sumner:
Looks great. Thank you

PR #6994, opened 2023-11-08
Update Dockerfile: Change `BUN_VERSION="bun-v1.0.7"` to `BUN_VERSION="bun-v1.0.11"`

JamBalaya56562:
#### Update `BUN_VERSION="bun-v1.0.11"`
My opinion is `BUN_VERSION` ought to be **correspond with latest release version** at the same time in Dockerfile.

https://github.com/oven-sh/bun/blob/03298eb8dd96620ba0639997a3ba6f4c8f2c2d1d/Dockerfile#L19
<br>
--
`BUN_VERSION` argument will be a part of Download URL.

https://github.com/oven-sh/bun/blob/03298eb8dd96620ba0639997a3ba6f4c8f2c2d1d/Dockerfile#L20
https://github.com/oven-sh/bun/blob/03298eb8dd96620ba0639997a3ba6f4c8f2c2d1d/Dockerfile#L118
<br>
--
#### Use `latest` instead of `BUN_VERSION`
BTW, there is the other way to use `latest` instead of `BUN_VERSION` argument like below.

https://github.com/oven-sh/bun/blob/03298eb8dd96620ba0639997a3ba6f4c8f2c2d1d/.github/workflows/bun-mac-x64.yml#L9
I would like to check which is better.

- [ ] Code changes


Electroid:
This is intentional for now. That `Dockerfile` is used to _build_ Bun from scratch, not to _use_ Bun. If you want to use Bun, check out our official Docker images: https://hub.docker.com/r/oven/bun

Issue #6595, opened 2023-10-19
Do not pass script flags to lifecycle scripts


gonzaloriestra:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

- Create a package.json with the following content:
```
{
    "scripts": {
      "premyscript": "cat package.json",
      "myscript": "bun"
  }
}
```
- `bun myscript -v`

### What is the expected behavior?

It should run `cat package.json` and then `bun -v`.

### What do you see instead?

It tries to run `cat package.json -v`

### Additional information

It happens the same when adding a `postmyscript`. The flag should only be passed to the main script, not to the pre or post ones. Other package managers don't do that.

PR #6997, opened 2023-11-08
don't pass script flags to pre- or post- scripts

nektro:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/6595

- [x] Code changes

I wrote automated tests

- [ ] I made sure that specific versions of dependencies are used instead of ranged or tagged versions
- [ ] I ran `make headers` to regenerate the C header file
- [ ] I ran `make codegen` to regenerate the C++ and Zig code
- [ ] I added a test that imports the module
- [ ] I added a test that require() the module


dylan-conway (approved):
Thanks!

Issue #6996, opened 2023-11-08
util.formatWithOptions type issue


brettgoulder:
### What version of Bun is running?

1.0.11

### What platform is your computer?

MacOS

### What steps can reproduce the bug?

It seems util.formatWithOptions was added here: https://github.com/oven-sh/bun/issues/3760

However, its type does not appear in the VSCode editor, resulting in the Typescript error: `Property 'formatWithOptions' does not exist on type 'typeof import("util")'.`

I'm pretty sure it's because this is commented out: https://github.com/oven-sh/bun/blob/main/packages/bun-types/util.d.ts#L129-L130

Have any thoughts?

paperclover:
an older version of bun did not have this function. when i implemented it, i did not uncomment the types.

do you want to open a pr that uncomments this?

brettgoulder:
@paperdave yep, I can. I didn't quite understand the comment regarding the polyfill though. Is there something that I need to do other than uncomment this?

paperclover:
nope. the "but is not defined in the polyfill" refers to before i implemented it in the polyfill.

brettgoulder:
okay great, got a PR up for your review

PR #7002, opened 2023-11-09
Fix util.formatWithOptions typing.

brettgoulder:
### What does this PR do?

Uncomments the Typescript type for util.formatWithOptions. When formatWithOptions was [first implemented](https://github.com/oven-sh/bun/pull/4090), its types were not uncommented. 

Referenced in this issue: https://github.com/oven-sh/bun/issues/6996


paperclover:
thanks

PR #7009, opened 2023-11-09
fix(ffi): fix ffi with large `uint32_t` values

### LINKED ISSUE #7007: FFIType.u32 misbehavior

1.0.10

Linux 4.19.289-1-MANJARO x86_64 unknown

Topic FFI. 
We have an given C Struct **Color**, that has a size of 32Bit = 4byte. In my case it's on raylib. 
```C
typedef struct Color {
    unsigned char r;        // Color red value
    unsigned char g;        // Color green value
    unsigned char b;        // Color blue value
    unsigned char a;        // Color alpha value
} Color;
```

Some methods, for example `void DrawRectangle(int posX, int posY, int width, int height, Color color);` demands to get this struct passed by value. (Other functions using color as byValue parameter also had been tested with the same result)

FFI dlopen parameters: 
```DrawRectangle: {args:[FFIType.i32, FFIType.i32, FFIType.i32, FFIType.i32, FFIType.u32]},``` 

The Plan: bypass the 'issue' of not beeing able to pass structs byValue (at least I dont know any way) with passing an u32 value. 

Function for preparing data:
```js
function rgba(r,g,b,a) {
  const arr = new Uint8Array([a,b,g,r]);
  return new DataView(arr.buffer).getUint32(0);
}
``` 

```js
var color = undefined;

//1. Draws invisible magenta rect
color = rgba(255, 0, 255, 0);
DrawRectangle(100,100, 100, 100, color);
console.log(color.toString(16)); // = 00FF00FF

//2. Draws 50% transparent magenta rect
color = rgba(255, 0, 255, 127);
DrawRectangle(100,100, 100, 100, color);
console.log(color.toString(16)); // = 7FFF00FF

//3. Draws magenta rect 
color = rgba(255, 0, 255, 255);
DrawRectangle(100,100, 100, 100, color);
console.log(color.toString(16)); // = FFFF00FF
```

### What do you see instead?

In case 3 we do se an about 25% visible, blue square. 
image

This is archived as soon as the most significant bit of alpha, which is also the most significant bit of the whole u32, gets set. So in a range of 128-255.

In that case also the function of all the other bytes becomes random from 'does nothing at all' up to 'gets black or blue'.
As Example: An entire solid blue can be archived by using `255, 255, 0, 255` - thats make no sense at all. 

### Additional information

Even thought the console outputs show the expected values, I also computed them manually using bitmask. The outcome stay the same.

Since the kinda random, but deterministic behavior starts with the most significant bit, I could just guess that even thought u32 is used, at some point an signed value is beeing expected. 

It would be way better to not have to use u32 at all, to just pass an buffer as argument in the first place. But so far only pointers seems to be supported. 

Im new to bun. So I dont know where in the sourcecode this could be adressed. But I would like to look deeper into this, if someone could give me the right place to look at.

paperclover:
### What does this PR do?

Fixes #7007 

This fixes the following C code
```c
unsigned int return_u32(unsigned int x) {
  printf("%x\n", x);
  return x;
}
```

And the given JS code
```js
import { dlopen } from "bun:ffi";

const dl = dlopen("wtf.dylib", {
  return_u32: {
    args: ["uint32_t"],
    returns: "uint32_t",
  },
});

console.log(dl.symbols.return_u32(0x7fffffff).toFixed(16));
console.log(dl.symbols.return_u32(0x80000000).toFixed(16));
console.log(dl.symbols.return_u32(0x80000001).toFixed(16));
console.log(dl.symbols.return_u32(0x80000002).toFixed(16));
```

bun is before, bd is after
image

To get this fix, I am abusing JSValue's int32 tag and the js bitwise or operator to efficiently pass uint32 through to C. Here is the comment left in the source code:

> This cast with `|0` looks incorrect as it converts 0xffffffff into -1, but this misinterpretation of the integer is taken advantage of by a second misinterpretation of the bytes in the C binding The bitwise operator | forces a conversion to int32_t, but it will wrap to negative numbers when going above >0x7fffffff.
>
> What this `|0` operatation also *seems to do* (citation needed) is convert the internal representation of `JSC::JSValue` to ALWAYS use Int32Tag, which is important as `JSValue::asInt32()` can only handle
> this encoding to properly deserialize this as an int32.
>
> *tldr jsc internals*: JSValue represents int32 as a tag value, then the int32 bytes. and all other integers are as tagged 64-bit floats.
>
> The trick to fixing the bug: after using |0 to misinterpret and force the integer into Int32Tag,
> when passing the value to the C ffi code, misinterpret it again, resulting in the correct uint32_t.
>
> To do this in native code, there is a spot in zig where `uint32_t` just prints `int32_t`.

A separate issue existed for going back from C to JS, which is handled by implementing a dedicated uint32_t -> JSValue, as this must encode values above MAX_U32 as a double

Manually tested FFI and added tests to `ffi.test.ts`. I spotted existing crashes from this test when you enable FFI.

### DIFF
--- a/src/bun.js/api/FFI.h
+++ b/src/bun.js/api/FFI.h
@@ -197,7 +197,7 @@ static EncodedJSValue INT32_TO_JSVALUE(int32_t val) {
 
 static EncodedJSValue UINT32_TO_JSVALUE(uint32_t val) {
   EncodedJSValue res;
-  if(val < MAX_INT32) {
+  if(val <= MAX_INT32) {
     res.asInt64 = NumberTag | val;
     return res;
   } else {
--- a/src/bun.js/api/FFI.h
+++ b/src/bun.js/api/FFI.h
@@ -195,8 +195,18 @@ static EncodedJSValue INT32_TO_JSVALUE(int32_t val) {
    return res;
 }
 
-
-
+static EncodedJSValue UINT32_TO_JSVALUE(uint32_t val) {
+  EncodedJSValue res;
+  if(val < MAX_INT32) {
+    res.asInt64 = NumberTag | val;
+    return res;
+  } else {
+    EncodedJSValue res;
+    res.asDouble = val;
+    res.asInt64 += DoubleEncodeOffset;
+    return res;
+  }
+}
 
 static EncodedJSValue FLOAT_TO_JSVALUE(float val) {
   return DOUBLE_TO_JSVALUE((double)val);
--- a/src/bun.js/api/ffi.zig
+++ b/src/bun.js/api/ffi.zig
@@ -1097,12 +1097,11 @@ pub const FFI = struct {
                     try writer.writeAll(", ");
                 }
                 first = false;
-                try arg.typename(writer);
+                try arg.paramTypename(writer);
                 try writer.print(" arg{d}", .{i});
             }
             try writer.writeAll(
                 \\);
-                \\
                 \\
                 \\/* ---- Your Wrapper Function ---- */
                 \\ZIG_REPR_TYPE JSFunctionCall(void* JS_GLOBAL_OBJECT, void* callFrame) {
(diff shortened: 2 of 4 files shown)


Jarred-Sumner:
interesting/clever

thanks for writing a comment, this will be hard to understand in 6 months

PR #7011, opened 2023-11-09
fix(node-fetch) Set-Cookie key on node-fetch.raw() should be set-cookie instead of Set-Cookie to match nodejs behavior

cirospaciari:
### What does this PR do?

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
It really expects `raw` to always be lowercased? The name `raw` implies it was as it was received by the HTTP server.

cirospaciari:
> It really expects `raw` to always be lowercased? The name `raw` implies it was as it was received by the HTTP server.

Yes, I was a little surprised too, so I tested with:

```js
const server = net
  .createServer((socket) => {
    socket.write("HTTP/1.1 200 OK\r\n");
    socket.write("Set-Cookie: session=12345\r\n");
    socket.write("Set-Cookie: session2=12345\r\n");
    socket.write("\r\n");
    socket.write("ok");
    socket.end();
  })
  .listen(8080);
```

```js
    import fetch from "node-fetch";
    fetch("http://127.0.0.1:8080", {
      headers: {
        "User-Agent": "real-ua/1.0.0",
      },
    }).then(async (r) => {
      console.log(r.headers.raw());
    });
```

Nodejs will always return lowercase
image

Issue #6330, opened 2023-10-05
Support `expect.extend()`


tinchox5:
### What is the problem this feature would solve?

I'm working on a CSS UI library and the best way I found for testing it is comparing image snapshots. Code snapshots are not useful in my case.

### What is the feature you are proposing to solve the problem?

Implement image snapshots diff feature.

### What alternatives have you considered?

On  Node I use jest with   "jest-image-snapshot" (a image snapshot extension) and puppeteer. That solves my requirements.

Electroid:
We can support this once Bun implements `expect.extend()`

etiennejcharles:
I think this might be a duplicate of #3621 - follow here instead @Electroid - I might be wrong too 🤷🏿‍♂️

Electroid:
Duplicate of #3621

PR #7017, opened 2023-11-09
Add integration tests for Vite, Astro, and Nuxt

Electroid:
This adds some integration tests for Vite, Astro, and Nuxt. We should also add pupeteer tests, but this is blocked by #6330


Jarred-Sumner:
Closing for now but let's revisit in the near future

PR #7018, opened 2023-11-10
add nix flake

rupurt:
### What does this PR do?

Adds a nix flake with an overlay to override the version if it hasn't been added as a package yet. This was extracted from my own repo https://github.com/rupurt/bun-nix

```
nix develop -c $SHELL
bun --version
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,111 @@
+{
+  lib,
+  stdenvNoCC,
+  fetchurl,
+  autoPatchelfHook,
+  unzip,
+  installShellFiles,
+  openssl,
+  writeShellScript,
+  curl,
+  jq,
+  common-updater-scripts,
+}: {specialArgs ? {}}: let
+  shas = {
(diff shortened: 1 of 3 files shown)


paperclover:
what the use case in having a nix flake on our repository instead of using [the version on nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/web/bun/default.nix). to me, this just adds files that will likely be left unmaintained and go out of date.

Eveeifyeve:
> what the use case in having a nix flake on our repository instead of using [the version on nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/web/bun/default.nix). to me, this just adds files that will likely be left unmaintained and go out of date.

I might add a follow up that keeps the flake updated for the devshell via the [update-flake-lock ](https://github.com/DeterminateSystems/update-flake-lock) action.

PR #7019, opened 2023-11-10
Fix toBeNaN's misleading JSDoc comment

joseph082:
### What does this PR do?

This fixes `toBeNaN`'s JSDoc comment, which currently says that it accepts any value coerced to `NaN` and that this is the same as `Number.isNaN`. This is misleading because `Number.isNaN` doesn't coerce the value. Only `isNaN` coerces.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#description

The [zig source code](https://github.com/oven-sh/bun/blob/046d1975192316cb78886269c70c1c292a04a733/src/bun.js/test/expect.zig#L734) shows that the function acts like `Number.isNaN` and only accepts `NaN`, since pass is only true when the value is a number and doesn't equal itself.

- [ ] Code changes

I ran this test file with Bun and Jest to check.

```js
import { expect, test } from 'bun:test';

test('toBeNaN', () => {
  const NaNs = [NaN, Number.NaN, 0 / 0];
  for (const n of NaNs) {
    expect(n).toBeNaN();
    expect(Number.isNaN(n)).toBe(true);
  }

  const coercedNaNs = ['NaN', undefined, {}, 'blabla', 'notanumber'];
  for (const n of coercedNaNs) {
    expect(n).not.toBeNaN();
    expect(Number.isNaN(n)).toBe(false);
    expect(isNaN(n)).toBe(true);
  }

  const notNaNs = [37, true, null, '37', '37.37', '', ' ', Infinity];
  for (const n of notNaNs) {
    expect(n).not.toBeNaN();
    expect(Number.isNaN(n)).toBe(false);
    expect(isNaN(n)).toBe(false);
  }
});
```

### DIFF
--- a/packages/bun-types/bun-test.d.ts
+++ b/packages/bun-types/bun-test.d.ts
@@ -767,14 +767,14 @@ declare module "bun:test" {
      */
     toBeNull(): void;
     /**
-     * Asserts that a value can be coerced to `NaN`.
+     * Asserts that a value is `NaN`.
      *
      * Same as using `Number.isNaN()`.
      *
      * @example
      * expect(NaN).toBeNaN();
-     * expect(Infinity).toBeNaN();
-     * expect("notanumber").toBeNaN();
+     * expect(Infinity).toBeNaN(); // fail
+     * expect("notanumber").toBeNaN(); // fail
      */
     toBeNaN(): void;
     /**


Electroid (approved):
Thanks!

PR #7027, opened 2023-11-10
Lots of stuff

Jarred-Sumner:
### What does this PR do?

- Use the debug build of JSC by default in development
- Use the debug build of mimalloc in development
- Add a build of Bun in CI which uses the debug build of JSC and ReleaseSafe build of Zig (linux only for now)
- Enable assertions in ReleaseSafe
- Disable spurious mimalloc warnings so that CI doesn't fail
- Fix assertion failure in `process.mainModule` (add support for `BuiltinAccessor` to `create_static_hash_tables`)
- Fix potential crash when loading a lot of modules in the test runner due to not ref'ing the initially-imported String
- Fix assertion failure in `onerror`
- Fix assertion failure when creating a `Worker` (x2)
- Fix assertion failure in coerceToInt64
- Fix assertion failure in `Headers.prototype.getAll`
- Fix assertion failure in BroadcastChannel
- Fix assertion failures in bun:sqlite
- Fix assertion failures in debugger
- Fix assertion failures in JSPromise::reject
- Fix assertion failure in Bun.serve() when a request is aborted
- Fix crash in debuger
- Fix assertion failure in module.require && module.require.resolve
- Fixes [#7067](https://github.com/oven-sh/bun/issues/7067)
- Fix assertion failure in ServerWebSocket.close
- Upgrade to the latest Zig
- [napi] Make promises cheaper
- Fix assertion failure in N-API with subclasses
- Fix assertion failure in `bun:jsc`'s `memoryUsage()`
- Preserve directives in the transpiler (such as `"use client"`)
- Workaround assertion faliures in ServerWebSocket.sendText & ServerWebSocket.sendBinary (due to (ptr, bool))

Locally, but lets see if CI works


scripts/build-mimalloc.sh
```diff
@@ -27,3 +27,26 @@ ninja
 
 cp libmimalloc.a $BUN_DEPS_OUT_DIR/libmimalloc.a
 cp CMakeFiles/mimalloc-obj.dir/src/static.c.o $BUN_DEPS_OUT_DIR/libmimalloc.o
+
+rm -rf CMakeCache* CMakeFiles
```
paperclover: if we have this, please rm build-mimalloc-debug.sh

paperclover:
looks good, but i do not think the entire suite of tests will pass with assertions on. still probably best we do this change

src/env.zig
```diff
@@ -24,7 +24,7 @@ pub const isLinux = @import("builtin").target.os.tag == .linux;
 pub const isAarch64 = @import("builtin").target.cpu.arch.isAARCH64();
 pub const isX86 = @import("builtin").target.cpu.arch.isX86();
 pub const isX64 = @import("builtin").target.cpu.arch == .x86_64;
-pub const allow_assert = isDebug or isTest;
+pub const allow_assert = isDebug or isTest or std.builtin.Mode.ReleaseSafe == @import("builtin").mode;
```
paperclover, line 27: this is fine but `std.debug.runtime_safety` is a boolean for this type of thing

CMakeLists.txt
```diff
@@ -362,6 +372,7 @@ if(NOT WEBKIT_DIR)
     set(ASSERT_ENABLED "0")
 
     if(USE_DEBUG_JSC)
+        add_compile_definitions("BUN_DEBUG=1")
```
paperclover, line 375: this is likely incorrect. i was under the impression BUN_DEBUG is only when using bun-debug, not release with assertions. for if we are using JSC debug we have ENABLE_ASSERT.

if you want to keep this, this must change InternalModuleRegistry to use a different mechanism, as in `#if BUN_DEBUG` will use a filesystem loading mechanism that is not portable to other machines.

paperclover: You should change

```
static void maybeAddCodeCoverage(JSC::VM& vm, const JSC::SourceCode& code)
#ifdef BUN_DEBUG
```
to `#if ENABLE_ASSERT`

and so on in other places you want to enable extra assertion-related code.

paperclover (changes requested):
overall good.

- one comment that requires attention
- kinda wish arm mac had an assertions build but it's ok.
- are zig error return traces enabled? i think they are. that would be cool.

CMakeLists.txt
paperclover, line 375:
```diff
-add_compile_definitions("BUN_DEBUG=1")
```

paperclover (changes requested):
need to remove definition for `BUN_DEBUG` in CMakeLists.txt or else the JSC debug build + release build will not be portable.

CMakeLists.txt
paperclover: idk how to do the suggestion thing

paperclover (approved):

SeedyROM:
Would this PR happen to affect this issue? https://github.com/oven-sh/bun/issues/7063

Seems like something was weird with the WTFStrings at GC.

docs/project/contributing.md
paperclover, line 1: can you update the date in this file:

```md
{% callout %}
We last updated Zig on **November 13th, 2023**
{% /callout %}
```

src/bun.js/bindings/BunString.cpp
```diff
@@ -467,3 +468,20 @@ size_t BunString::utf8ByteLength(const WTF::String& str)
         return simdutf::utf8_length_from_utf16(reinterpret_cast<const char16_t*>(str.characters16()), static_cast<size_t>(str.length()));
     }
 }
+
+WTF::String BunString::toWTFString() const
```
paperclover, line 481: can you remove `Bun::toWTFString(const BunString& bunString)`, and also move the assertion that was defined in that variant of the function to this?

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -198,14 +209,14 @@ Structure* RequireFunctionPrototype::createStructure(
     JSC::VM& vm,
     JSC::JSGlobalObject* globalObject)
 {
-    return Structure::create(vm, globalObject, globalObject->functionPrototype(), TypeInfo(JSFunctionType, StructureFlags), info());
+    return Structure::create(vm, globalObject, globalObject->functionPrototype(), TypeInfo(ObjectType, StructureFlags), info());
```
paperclover, line 212: i thought function prototypes need to be of function type but if it works it works. my mistake must be that it should not have been done on the prototype but just the actual structure.

src/bun.js/bindings/webcore/JSFetchHeaders.cpp
```diff
@@ -605,7 +612,7 @@ JSC::JSValue getInternalProperties(JSC::VM& vm, JSGlobalObject* lexicalGlobalObj
         obj = constructEmptyObject(lexicalGlobalObject);
         RETURN_IF_EXCEPTION(throwScope, {});
         RELEASE_AND_RETURN(throwScope, obj);
-    } else if (size < 64) {
+    } else if (size < JSFinalObject::maxInlineCapacity) {
```
paperclover: <=

src/bun.js/bindings/webcore/JSURLSearchParams.cpp
```diff
@@ -404,7 +404,7 @@ JSC::JSValue getInternalProperties(JSC::VM& vm, JSC::JSGlobalObject* lexicalGlob
     auto throwScope = DECLARE_THROW_SCOPE(vm);
 
     JSObject* obj;
-    if (impl.size() + 1 < 64) {
+    if (impl.size() + 1 < JSFinalObject::maxInlineCapacity) {
```
paperclover: <=

paperclover (commented):
this is certainly a lot of stuff. no major concerns

Issue #1265, opened 2022-09-21
Support `--env-file` to import custom `.env` files


danosiersonos:
### What is the problem this feature will solve?

So for a lot of users/companies, they have multiple environments to deal with different datasets, connections, API, integrations, etc. Right now bun supports .env, .env.local, and .env.development (unless I'm mistaken).

It would be nice if bun could support more than just these 4, and support custom names as well. Staging, performance, testing, etc.

### What is the feature you are proposing to solve the problem?

I think that this could be solved multiple ways, likely.

I think you could support some type of map/fallback structure in the bunfig.toml.
Base the name of the .env file you load from a BUN_ENV, or to be more compatible with node, use NODE_ENV.

### What alternatives have you considered?

To force everyone to only ever have 3 env files ¯\_(ツ)_/¯

I'm no expert, I just thought this would be an excellent addition

Jarred-Sumner:
The only reason Bun hardcodes it right now was due to prioritization constraints when implementing it - not philosophical 

A good question is how best to specify which .env files to load. An environment variable could work, with some constraint that the environment variable comes from the process instead of another .env 

A CLI flag would also make sense 

A bunfig option would sort of make sense except there's only one bunfig right now

danosiersonos:
I assumed it was a prioritization thing, not a philosophical one 🙌. But always good to have something on the board rather than possibly lose track of it.

CLI flag would be good too.

Does the one bunfig prevent you from using it? Worried about it becoming massive?

joshuat:
@Jarred-Sumner I think the defaults in place make a lot of sense, but I've always really liked the API from [node-foreman](https://github.com/strongloop/node-foreman/blob/782cf090d4917ff137e9980a36803b93df818b96/nf.js#L20) for .env loading. You're able to supply a comma-separated list of filenames which load in order similar to how Bun loads the current files in order.

bertyhell:
An cli option would be nice. Something similar to the latest nodejs version:
```
node --env-file=env-folder/.env.production index.js
```

https://netbasal.com/node-js-v20-6-0-introducing-built-in-env-file-support-fb1b1de4f1fa

This would provide a good migration path for people using dotenv:
```
dotenv -e ./env-folder/.env.production node index.js 
```

eg:
```
bun --env-file=env-folder/.env.production index.ts
```

webfussel:
I'd like to push this issue.
[...]

PR #7032, opened 2023-11-10
feat(cli) implement --env-file arg for bun run to load custom env files

otgerrogla:
### What does this PR do?

Implements #1265: adds support for the `--env-file` argument for `bun run`, to load custom dotenv files (`.env`).

Specifically, when this argument is passed, it disables the default behavior of looking for `.env`, `.env.production`, `.env.production.local`, etc.; and instead only loads the specified files. In case of specifying more than one file, the latter one always has higher priority and overrides values from previous ones (regardless of which syntax is used).

It supports the following syntaxes:
- `--env-file=.env1`
- `--env-file .env1`
- `--env-file .env1 --env-file=.env2` (load multiple files by passing multiple instances of the arg)
- `--env-file .env1,.env2` (load multiple files as a comma-separated list of files)

These syntaxes provide compatibility with the [NodeJS --env-file arg](https://nodejs.org/api/cli.html#--env-fileconfig) (implemented on v20.6.0), along with other popular toolings.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/completions/bun.bash
+++ b/completions/bun.bash
@@ -84,7 +84,7 @@ _bun_completions() {
 
     local SUBCOMMANDS="dev bun create run install add remove upgrade completions discord help init pm x";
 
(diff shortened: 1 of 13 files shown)


Jarred-Sumner:
This looks great. 

Thank you.

PR #6064, opened 2023-09-26
WIP: Add implementation for `server.url`

GabrielDougherty:
### What does this PR do?

This adds a new `url` property on `server` which returns a URL object.  That way, the URL object can be passed around instead of the individual fields `server.hostname` and `server.port`.

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

I wrote automated tests

If JavaScript/TypeScript modules or builtins changed:

- [x] I ran `make js` and committed the transpiled changes
- [x] I or my editor ran Prettier on the changed files (or I ran `bun fmt`)
- [x] I included a test for the new code, or an existing test covers it

 If Zig files changed: **(TODO)**

- [ ] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [x] I or my editor ran `zig fmt` on the changed files
- [ ] I included a test for the new code, or an existing test covers it
- [ ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed

 If new methods, getters, or setters were added to a publicly exposed class:

- [x] I added TypeScript types for the new methods, getters, or setters

 If \*.classes.ts files were added or changed:

- [x] I ran `make codegen` to regenerate the C++ and Zig code

PR #7043, opened 2023-11-11
Implement `server.url`

### LINKED ISSUE #5758: Implement `server.url`
It's a common pattern to see this:

```js
const server = Bun.serve(...);
const result = await fetch(`http://${server.hostname}:${server.port}/`);
```

But that's cumbersome, and is invalid for IPv6 addresses. Instead, Bun should provide a built-in API for the URL.

```ts
interface Server {
  get url(): URL;
}
```

Electroid:
### What does this PR do?

This adds support for `server.url` from `Bun.serve()`.

```js
const server = Bun.serve({
  fetch() {
    return new Response();
  }
});

console.log(server.url instanceof URL); // true
console.log(server.url.protocol); // "http:"
console.log(server.url.hostname); // "localhost"
console.log(server.url.port); // 3000
```

For unix sockets, it also works.

```js
const server = Bun.serve({
  unix: "/tmp/bun.sock",
  fetch() {
    return new Response();
  }
});

console.log(server.url instanceof URL); // true
console.log(server.url.protocol); // "unix:"
console.log(server.url.pathname); // "/tmp/bun.sock"
```

- [x] Code changes

Closes #5758, #6064


Jarred-Sumner:
yay

PR #7055, opened 2023-11-11
feat: adds remove comments option to transpiler

pierre-cm:
### What does this PR do?

This impements the [removeComments](https://www.typescriptlang.org/tsconfig#removeComments) tsconfig option for for bun's transpiler.

It is set to `true` by default to maintain current behavior. However, when explicitly set to `false`, it would retain comments in the transformed JavaScript code.

- [x] Code changes

If Zig files changed:


Jarred-Sumner:
Thank you for this. It's close to shippable. Note that we never implemented `removeMultilineCommentIndent` so the indentation in a multiline comment might be preserved which would be confusing when the code is re-printed or used with `--minify-syntax`

pierre-cm:
Thank you for your feedback. I've added an implementation for `removeMultilineCommentIndent`. My approach is straightforward, as I wasn't certain if there were specific cases to consider. I've made the assumption that comments are simply excluded when the code is minified.
If there are any specific requirements or scenarios I should account for, please let me know so I can adjust the implementation accordingly

PR #7064, opened 2023-11-12
Fixed mistake in Guide "Stream File as an HTTP Reponse"

GColoy:
### What does this PR do?
The dots in Front of the relative Paths were missing, so it always gave an error that the directory or file doesn't exist.

- [ ] Code changes


Electroid:
This could go either way, because `Bun.file` supports both relative and absolute paths. Going to keep as is.

PR #7070, opened 2023-11-12
nodejs compat into a compact list-like page

jcbhmr:
### What does this PR do?

- [ ] Code changes

this is a pretty opinionated thing. i think it looks better and is more readable when you dont need to scroll so much with all the blank space from the header margins. but that's me. feel free to close this pr if this doesn't look good to you 👍

preview gh markdown viewer https://github.com/jcbhmr/bun/blob/patch-1/docs/runtime/nodejs-apis.md
vs live docs page https://bun.sh/docs/runtime/nodejs-apis

oh and pls make sure i got all the text right and make sure i didn't mix anything up lol

### DIFF
--- a/docs/runtime/nodejs-apis.md
+++ b/docs/runtime/nodejs-apis.md
@@ -4,448 +4,227 @@ This page is updated regularly to reflect compatibility status of the latest ver
 
 ## Built-in modules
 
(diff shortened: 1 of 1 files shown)


Electroid:
I think we'll keep it as-is for now, but appreciate your time to make this PR!

PR #7072, opened 2023-11-12
Handle whitespace in no_proxy environment variable

### LINKED ISSUE #6339: `NO_PROXY` environment variable does not work with spaces

Newest version as of yesterday (2023-10-05) installed via NPM in GitLab Runner

Docker: Slightly modified Amazonlinux

Have `NO_PROXY` set with multiple domains, some include a space after the comma.

According to https://github.com/oven-sh/bun/issues/1440, there is [this blog post from GitLab](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/) explaining how `HTTPS_PROXY`, `HTTP_PROXY` and `NO_PROXY` should work. (I know, just a recommendation by gitlab)

Quote from the Blog Post:
> ## Standardizing no_proxy
> Knowing the least common denominator can help avoid issues if these definitions are copied for different Web clients. But should no_proxy and the other proxy settings have a documented standard rather than an ad hoc convention? The list below may serve as a starting point for a proposal:
> 1. [...]
> 2. Use comma-separated hostname:port values.
>    * Each value may include optional whitespace.
>
> [...]

However, Bun install will not accept `NO_PROXY` in the format `test.com, test2.com`

The issues comes from https://github.com/oven-sh/bun/blob/main/src/env_loader.zig#L129
```zig
var no_proxy_list = std.mem.split(u8, no_proxy_text, ",");
```

I would suggest to trim each entry in `no_proxy_list`.

Bun should allow spaces in `NO_PROXY`.

### What do you see instead?

Bun ignores some domains as ` domain.com` does not match `domain.com` (note the space)

### Additional information

_No response_

dottedmag:
### What does this PR do?

De facto standard format of no_proxy enivronment variable allows whitespace around the host names. Make it work.

Closes #6339

Zig unit tests are completely bitrotted (I have tried to make them work, but after an hour or so chasing `@import`s that do not work and disabling tests that do not compile anymore I gave up) , so I had to test it manually.

### DIFF
--- a/src/env_loader.zig
+++ b/src/env_loader.zig
@@ -140,6 +140,7 @@ pub const Loader = struct {
         }
 
         // NO_PROXY filter
+        // See the syntax at https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/
         if (http_proxy != null) {
             if (this.map.get("no_proxy") orelse this.map.get("NO_PROXY")) |no_proxy_text| {
                 if (no_proxy_text.len == 0 or strings.eqlComptime(no_proxy_text, "\"\"") or strings.eqlComptime(no_proxy_text, "''")) {
@@ -149,7 +150,7 @@ pub const Loader = struct {
                 var no_proxy_list = std.mem.split(u8, no_proxy_text, ",");
                 var next = no_proxy_list.next();
                 while (next != null) {
-                    var host = next.?;
+                    var host = strings.trim(next.?, &strings.whitespace_chars);
                     if (strings.eql(host, "*")) {
                         return null;
                     }
--- a/src/string_immutable.zig
+++ b/src/string_immutable.zig
@@ -4367,10 +4367,14 @@ pub fn trim(slice: anytype, comptime values_to_strip: []const u8) @TypeOf(slice)
     return slice[begin..end];
 }
 
+pub const whitespace_chars = [_]u8{ ' ', '\t', '\n', '\r', std.ascii.control_code.vt, std.ascii.control_code.ff };
+
 pub fn lengthOfLeadingWhitespaceASCII(slice: string) usize {
     for (slice) |*c| {
         switch (c.*) {
-            ' ', '\t', '\n', '\r', std.ascii.control_code.vt, std.ascii.control_code.ff => {},
+            whitespace: {
+                inline for (whitespace_chars) |wc| break :whitespace wc;
+            } => {},
             else => {
                 return @intFromPtr(c) - @intFromPtr(slice.ptr);
             },


src/env_loader.zig
```diff
@@ -149,7 +150,7 @@ pub const Loader = struct {
                 var no_proxy_list = std.mem.split(u8, no_proxy_text, ",");
                 var next = no_proxy_list.next();
                 while (next != null) {
-                    var host = next.?;
+                    var host = std.mem.trim(u8, next.?, &std.ascii.whitespace);
```
dylan-conway: instead of using `std.ascii.whitespace` and `std.mem.trim`, can you use `strings.trim` with the whitespace characters from `lengthOfLeadingWhitespaceASCII`? probably best to move them into a pub const in `strings_immutable.zig`

dottedmag: I have replaced `std.mem.trim` with `strings.trim`.

I have also tried to extract the constant, and found out that the whitespace characters from `lengthOfLeadingWhitespaceASCII` are identical to `std.ascii.whitespace`. `lengthOfLeadingWhitespaceASCII` uses its own set because it spells them out in `case`.

So this will create a constant that's the same as one in `std.ascii` and `lengthOfLeadingWhitespaceASCII` won't be able to reuse it. Should I do it, or is it OK to keep the code as is?

dylan-conway: We can reuse the constant. Here's a little comptime magic to spread the items from `whitespace_chars` into the switch statement.
```zig
pub const whitespace_chars = [_]u8{ ' ', '\t', '\n', '\r', std.ascii.control_code.vt, std.ascii.control_code.ff };

pub fn lengthOfLeadingWhitespaceASCII(slice: string) usize {
    for (slice) |*c| {
        switch (c.*) {
            whitespace: {
                inline for (whitespace_chars) |wc| break :whitespace wc;
            } => {},
            else => {
                return @intFromPtr(c) - @intFromPtr(slice.ptr);
            },
        }
    }

    return slice.len;
}
```
Sorry this is annoying, but we want to make sure this array of whitespace chars doesn't change when upgrading zig

dottedmag: > Here's a little comptime magic to spread the items from whitespace_chars into the switch statement.

Thanks, this is neat. And the assembly is pretty similar as well.

> Sorry this is annoying, but we want to make sure this array of whitespace chars doesn't change when upgrading zig

Not annoyng at all. I have forgotten that Zig does not guarantee stability of its stdlib yet.

There are several uses of `std.ascii.is{Whitespace,Control,Alphanumeric}` functions in the codebase, should they also be pulled into Bun proper?

dylan-conway: for now this is great

dylan-conway (approved):

dylan-conway:
Thanks!

Issue #4923, opened 2023-09-11
Bun lockfile broken permissions in devcontainer


dmaasland:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

Using these files, create a devcontainer in VS Code:

**`Dockerfile`**
```Dockerfile
FROM mcr.microsoft.com/devcontainers/base:bookworm

ARG BUN_INSTALL=/usr/local

RUN : \
    && curl -fsSL https://bun.sh/install | bash
```

**`devcontainer.json`**
```json
{
  "name": "Debian",
  "build": {
    "dockerfile": "Dockerfile"
  },
  "features": {
    "ghcr.io/devcontainers/features/common-utils:2": {}
  }
}
```

Run these commands (I chose the demo app with typescript). As per: https://bun.sh/guides/ecosystem/sveltekit
```bash
bunx create-svelte my-app
cd my-app
bun install
```

It will error out with:
```bash
bun install v1.0.0 (822a00c4)
  🔒 Saving lockfile... error: failed to save lockfile: AccessDenied
```

The lockfile is created with these permissions:
```bash
vscode ➜ /workspaces/bun_test/my-app $ ls -la
ls: cannot access 'bun.lockb': No such file or directory
total 28
drwxr-xr-x 12 vscode vscode 384 Sep 11 11:06 .
drwxr-xr-x  4 vscode vscode 128 Sep 11 11:05 ..
-?????????  ? ?      ?        ?            ? bun.lockb
-rw-r--r--  1 vscode vscode 148 Sep 11 11:05 .gitignore
-rw-r--r--  1 vscode vscode  43 Sep 11 11:05 .npmrc
```

### What is the expected behavior?

A working install of SvelteKit

### What do you see instead?

An error and weird file permissions.

### Additional information

I am not entirely sure this is a Bun issue or a Docker issue, but this is the first time I've come across it so figured I might as well report it should anyone else run into the same issue. Some more version info:

- Docker version 24.0.2, build cb74dfc
- VS Code Version: 1.81.1

ghost:
Same here. Same issue opened two weeks ago has not been addressed: #4314

kkirkfield:
[...]

PR #2588, opened 2023-04-07
Rewrite Developer Docs, Improve DX for new users, moving away from Devcontainer

paperclover:
The dev container isn't a great experience for working on bun (biggest limit is you can't easily take bun-debug outside of it), and setting up a development environment on the host computer isn't very difficult, it's just not documented. This PR aims to fix that, Instead of 30-90 minutes on figuring out how to set up zig and docker, just ~5 commands and ~10-30 minutes of network and build time.

It only covers these three common situations, which I think we should expand to some more
- MacOS
- Ubuntu
- Arch Linux

But this is much better than before, where we pushed the Docker Devcontainer instead.

Documentation TODO
- [x] Move all other information in old that I omitted
- [ ] Explain purpose for common make commands
- [x] IDE Stuff
  - [ ] ZLS, solved already by https://github.com/oven-sh/bun/pull/2434
  - [x] C++ Config
- [ ] Explain that `-j` shouldn't be used except on bindings. Note on why make dev runs on one core. etc.
- [x] rename for consistency; URL should match sidebar `/project/development`, readme should match too.
- [x] Explain that npm can be used if bun is not available
- [x] Make note of the devcontainer; it works, but meh. (or should it be removed completely)
- [x] How to build release

Makefile changes:
- `-l:libatomic.a` -> `-latomic` to support Arch Linux only shipping with a `.so` file of this.
- [x] Hide `which` errors, since this confused me initially and made me think I had to install other packages when in reality the Makefile used a different binary that did exist.
- [x] Use `node_modules/esbuild` instead of global (Ubuntu 20)
- [X] Use `python3` when `python` and `python2` don't exist. (Ubuntu 20)
- [x] Add `make setup` which is a shortcut for vendor, bindings, etc
- [X] Add `make regenerate-bindings` which is a shortcut for `make clean-bindings builtins && make bindings -j (cpu count)`

Other ideas
- Move to using `bun i @oven/zig` in the actual repo? it's probably too soon for that in this PR. once that's been done, if decided, then the zig section of the document can be almost entirely removed.

PR #1439, opened 2022-11-01
Fix $PATH on dev container

jpaquim:
$PATH was pointing to `/build/packages/*` instead of `/build/bun/packages/*` where they actually get built.

The `code /build/README.md` in the post-create command was also causing some trouble, I replaced it with a generic cat instead.

Issue #743, opened 2022-07-16
devcontainer build error


jmhungdev:
### Version

`@vscode/dev-container-cli 0.241.3.`

### Platform

`Darwin 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020.121.3~4/RELEASE_X86_64 x86_64`

### What steps will reproduce the bug?

Following this guide: https://github.com/oven-sh/bun#vscode-dev-container-linux 
up to the step of `devcontainer build`

### What do you see instead?
```
[+] Building 213.3s (14/16)                                                     
 => [internal] load .dockerignore                                          0.0s
 => => transferring context: 35B                                           0.0s
 => [internal] load metadata for docker.io/bunbunbunbun/bun-base-with-zig  1.0s
 => [auth] bunbunbunbun/bun-base-with-zig-and-webkit:pull token for regis  0.0s
 => ERROR importing cache manifest from bun.devcontainer:latest            1.4s
 => [internal] load build context                                          0.0s
 => => transferring context: 491B                                          0.0s
 => [bun.devcontainer 1/9] FROM docker.io/bunbunbunbun/bun-base-with-zig-  0.0s
 => [auth] library/bun.devcontainer:pull token for registry-1.docker.io    0.0s
 => CACHED [bun.devcontainer 2/9] RUN apt-get -y update && update-alterna  0.0s
 => CACHED [bun.devcontainer 3/9] COPY .devcontainer/workspace.code-works  0.0s
 => CACHED [bun.devcontainer 4/9] COPY .devcontainer/zls.json /build/work  0.0s
 => CACHED [bun.devcontainer 5/9] COPY .devcontainer/limits.conf /etc/sec  0.0s
 => CACHED [bun.devcontainer 6/9] COPY .devcontainer/scripts/ /scripts/    0.0s
 => CACHED [bun.devcontainer 7/9] COPY .devcontainer/scripts/getting-star  0.0s
 => [bun.devcontainer 8/9] RUN mkdir -p /home/ubuntu/.bun /home/ubuntu/  210.7s
```

```
#13 232.1 error: the following build command failed with exit code 9:
#13 232.1 /home/ubuntu/zls/zig-cache/o/7f1cfb4bad3d0bdd8da0f89db0bd1c8f/build /build/zig/zig /home/ubuntu/zls /home/ubuntu/zls/zig-cache /root/.cache/zig -Drelease-fast
------
[...]

PR #7083, opened 2023-11-13
Add 'Bun Development Environement' devcontainer

dionhut:
### What does this PR do?

Added VSCode DevContainer to automate setup of Bun development environment.

TLDR;
- Added a `bun-devcontainer` image Dockerfile + dependency build setup scripts
- Added .devcontainer VSCode config
- Image based on known good VSCode base devcontainer image designed for devcontainers
- build and node_modules folders are mounted volumes which survive container rebuilds and build performance.
- Pre-built published image can save developers building their own development environment.
- Bun project can provide a default set of VSCode extensions which are automatically installed in the devcontainer.

First off, I'm a huge fan of the project! I initially wanted to help contribute by looking into fixing the devcontainer lockfile issue - https://github.com/oven-sh/bun/issues/4923
I started out seeing I could build the **bun** repo in a devcontainer cause I didn't want to contaminate my existing macbook development environment.  I use isolated devcontainers where I can.

It's my view that the value of being able to contribute to the bun project itself or developing with bun javascript projects in isolation with other dev environments on a dev's work or personal machine and not have to worry about setup of extensive deps that can interfere with other projects is a must.

Also devcontainers are great at allowing a project to capture infra as code for the development environment in source control and each contributor could rely on working with the same environment every time an issue is reported or giving confidence PRs are at least testing with a consistent dev env.  Hopefully I don't have to sell it much more :)

#### Tested building docker image
Built and tested using Macbook M1

From **bun** repo folder
`cd ~/code/bun`

Build the image - Ideally this image would be built via CI or something and published most likely the same way the oven/bun image is.  This probably goes with out saying, but contributor developers wouldn't need to build the `bun-devcontainer` image.
`cd .devcontainer && ./build-devcontainer.sh`

#### Tested opening devcontainer + bun project in VSCode

`code .`

Open project in the devcontainer
`Command-Shift-P` - `>dev Containers: Reopen in Container`

You should see this message after devcontainer is ready for work.
Screenshot 2023-11-12 at 8 06 27 PM

Now build **bun** as normal.   I found I had to run `scripts/clean-dependencies` to clean out all the other failed attempts to get the build environment right first.
`bun setup`

**And Bob's your uncle!**
Screenshot 2023-11-12 at 2 54 25 PM

#### Testing running bun-debug binary and debugging in VSCode with LLVM extension
Screenshot 2023-11-12 at 2 57 38 PM

#### TODO: Building the multiple images for each supported target architectures.  I might need help with this I'm not familiar with this project and build pipeline.

Looking forward to any feedback or if this is something the project wants to incorporate.  This change is definitely not a triumph or anything but a nice to have even if it only lives on in my fork.  Thanks.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,27 @@
+FROM mcr.microsoft.com/vscode/devcontainers/base:bullseye
+
+ARG BUN_VERSION="1.0.11"
+ARG NODE_VERSION="20"
+ARG LLVM_VERSION="16"
+ARG ZIG_VERSION="0.12.0-dev.1297+a9e66ed73"
+ARG CMAKE_VERSION="3.27.7"
+ARG TARGETARCH
+
+ENV TARGETARCH=${TARGETARCH}
+ENV NODE_VERSION=${NODE_VERSION}
+ENV LLVM_VERSION=${LLVM_VERSION}
+ENV ZIG_VERSION=${ZIG_VERSION}
+ENV CMAKE_VERSION=${CMAKE_VERSION}
+
+USER vscode
+
+ADD scripts/ /tmp/build-scripts
+
+RUN /tmp/build-scripts/setup-first.sh 
+RUN /tmp/build-scripts/setup-llvm.sh
+RUN /tmp/build-scripts/setup-nodejs.sh
+RUN /tmp/build-scripts/setup-cmake.sh
+RUN /tmp/build-scripts/setup-zig.sh
+
+RUN /tmp/build-scripts/setup-bun.sh
+RUN /tmp/build-scripts/setup-rust.sh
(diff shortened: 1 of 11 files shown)


staabm:
Can this image be used for github codespaces?

Jarred-Sumner:
Thank you for this. 

We [previously](https://github.com/oven-sh/bun/pull/2588) had a dev container (#743 #1680 #1439 #472 #933), but since the core team doesn't use dev containers, they frequently stopped working. The iteration cycle time is better building outside Docker, and we put a lot of effort into trying to make Bun easy to build via CMake. 

We don't have the bandwidth to maintain a devcontainer ourselves right now, but you're obviously welcome to do that in a separate repository if you'd like

dionhut:
> Can this image be used for github codespaces?

Great question. I don't see why not but I haven't tried it.  I guess I would say developing in a bun consumer javascript project environment in the cloud seems more compelling to me rather than working on the bun platform code itself.   And BTW currently bun install is broken (fails to save lockfile) devcontainers.  Cheers.

PR #7092, opened 2023-11-13
fix: made fs.readSync compatible with node version 21.1.0

### LINKED ISSUE #6832: fs.readSync not compatible with node

1.0.7

Linux 6.1.21.2-microsoft-standard-WSL2-maple x86_64 unknown

Compare

```shell
printf a | node -e 'let buf = Buffer.alloc(1); fs.readSync(0, buf, 0, 1); console.log(buf)'
// <Buffer 61>
```

and

```shell
printf a | bun repl --eval 'let buf = Buffer.alloc(1); fs.readSync(0, buf, 0, 1); console.log(buf)'
// TypeError: length and position are required
// code: "ERR_INVALID_ARG_TYPE"
```

`fs.readSync(0, buf, 0, 1)` should work as `position` is optional: https://nodejs.org/api/fs.html#fsreadsyncfd-buffer-offset-length-position

### What do you see instead?

image

### Additional information

_No response_

nithinkjoy-tech:
### What does this PR do?
Fixes #6832 

- [X] Code changes

- [X] I included a test for the new code and an existing test covers it


src/bun.js/node/node_fs.zig
```diff
@@ -2141,25 +2141,23 @@ pub const Arguments = struct {
...
                     }
-                    if (arguments.remaining[0].isNumber() or arguments.remaining[0].isBigInt())
-                        args.length = arguments.remaining[0].to(u52);
 
-                    if (args.length == 0) {
+                    args.offset = current.to(u53);
+                    if (args.offset > args.buffer.buffer.len) {
                         JSC.throwInvalidArguments(
-                            "length must be greater than 0",
+                            "The value of \"offset\" should be lesser than buffer length",
```
Jarred-Sumner: are you sure that node's behavior is to throw an error in this case and not clamp it?
```suggestion
                            "The value of \"offset\" should be less than the buffer length",
```

Jarred-Sumner:
Most of these tests seem to already pass in the canary build of Bun. I suggest copying more tests from Node

src/bun.js/node/node_fs.zig
nithinkjoy-tech: Thank you for the review.

In node, for example:
If buffer length is 1000 and offset is 1100 and  length is 100 it would give error ```The value of "length" is out of range. It must be <= -100. Received 100.``` But it won't accept length less than 0.

Should we make this behave same as node?

nithinkjoy-tech:
> Most of these tests seem to already pass in the canary build of Bun. I suggest copying more tests from Node

Sure. Will do that.

src/bun.js/node/node_fs.zig
nithinkjoy-tech: In the latest commit I have made it same as node.

nithinkjoy-tech:
@Jarred-Sumner Please review this PR.

nithinkjoy-tech:
@paperdave Please review this PR

Issue #4667, opened 2023-09-09
import.meta.env support


beepsoft:
### What is the problem this feature would solve?

Introducing access to `.env` variables via `import.meta.env` would make bun more compatible with existing tools (eg. Vite) and would help migrating to bun.

### What is the feature you are proposing to solve the problem?

It would be great to have support for accessing `.env` variables via `import.meta.env` as in Vite: https://vitejs.dev/guide/env-and-mode.html

taranek:
@Jarred-Sumner Any idea when this might happen? We (+ probably many other devs) can't migrate our tests to bun due to this issue

MarkLyck:
@Jarred-Sumner Also having this issue, a lot of my components import something from my config file which includes:

```
const NODE_ENV = import.meta.env.NODE_ENV
```

resulting in an error when running tests:

```
TypeError: undefined is not an object (evaluating 'import.meta.env.NODE_ENV')
```

sciolist:
As a workaround (worked for me at least) you should be able to create a file, say, fix-import-meta-env.js with the content `import.meta.constructor.prototype.env = process.env` then run your scripts with `bun -r ./fix-import-meta-env.js my-script.ts`.

Not something you'd want to keep in the long term. Would be nice to get it built in!

Jarred-Sumner:
Runtime support for import.meta.env is merged, thanks to @otgerrogla. Not yet supported at build-time

mnpenner:
I don't know when this started happening, but `bun run --bun build` works but `bun run build` doesn't. I thought it worked without `--bun` before but it isn't now. (Revision `1.1.25-canary.28+996847bca`)

Jarred-Sumner:
Seeing as `import.meta.env` is implemented at runtime, i'm going to close this. For now, we're keeping `process.env` as the only build-time inlined environment variable object.

PR #7094, opened 2023-11-13
Add support import.meta.env

otgerrogla:
### What does this PR do?

Implements #4667. Adds support for `import.meta.env` as an alias to `process.env` or `Bun.env`, to improve compatibility with `vite` and other tools.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/bindings/ImportMetaObject.cpp
+++ b/src/bun.js/bindings/ImportMetaObject.cpp
@@ -403,6 +403,11 @@ JSC_DEFINE_CUSTOM_GETTER(jsImportMetaObjectGetter_require, (JSGlobalObject * glo
 
     return JSValue::encode(thisObject->requireProperty.getInitializedOnMainThread(thisObject));
 }
+JSC_DEFINE_CUSTOM_GETTER(jsImportMetaObjectGetter_env, (JSGlobalObject * jsGlobalObject, JSC::EncodedJSValue thisValue, PropertyName propertyName))
+{
+    auto* globalObject = jsCast<Zig::GlobalObject*>(jsGlobalObject);
+    return JSValue::encode(globalObject->m_processEnvObject.getInitializedOnMainThread(globalObject));
+}
 
 static const HashTableValue ImportMetaObjectPrototypeValues[] = {
     { "resolve"_s, static_cast<unsigned>(JSC::PropertyAttribute::Function | PropertyAttribute::DontDelete), NoIntrinsic, { HashTableValue::NativeFunctionType, functionImportMeta__resolve, 0 } },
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
I wonder if we need to do this at build time too.

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -412,6 +417,7 @@ static const HashTableValue ImportMetaObjectPrototypeValues[] = {
     { "file"_s, static_cast<unsigned>(JSC::PropertyAttribute::ReadOnly | JSC::PropertyAttribute::CustomAccessor | PropertyAttribute::DontDelete), NoIntrinsic, { HashTableValue::GetterSetterType, jsImportMetaObjectGetter_file, 0 } },
     { "path"_s, static_cast<unsigned>(JSC::PropertyAttribute::ReadOnly | JSC::PropertyAttribute::CustomAccessor | PropertyAttribute::DontDelete), NoIntrinsic, { HashTableValue::GetterSetterType, jsImportMetaObjectGetter_path, 0 } },
     { "require"_s, static_cast<unsigned>(JSC::PropertyAttribute::ReadOnly | JSC::PropertyAttribute::CustomAccessor | PropertyAttribute::DontDelete), NoIntrinsic, { HashTableValue::GetterSetterType, jsImportMetaObjectGetter_require, 0 } },
+    { "env"_s, static_cast<unsigned>(JSC::PropertyAttribute::ReadOnly | JSC::PropertyAttribute::CustomAccessor | PropertyAttribute::DontDelete), NoIntrinsic, { HashTableValue::GetterSetterType, jsImportMetaObjectGetter_env, 0 } },
```
Jarred-Sumner, line 420: Is it read-only in Vite?

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -403,6 +403,11 @@ JSC_DEFINE_CUSTOM_GETTER(jsImportMetaObjectGetter_require, (JSGlobalObject * glo
 
     return JSValue::encode(thisObject->requireProperty.getInitializedOnMainThread(thisObject));
 }
+JSC_DEFINE_CUSTOM_GETTER(jsImportMetaObjectGetter_env, (JSGlobalObject * jsGlobalObject, JSC::EncodedJSValue thisValue, PropertyName propertyName))
+{
```
Jarred-Sumner, line 407: Can you add a test that runs this in node:vm? Just to make sure this jsCast doesn’t crash when the global object is not a Zig::GlobalObject

Jarred-Sumner:
Thank you for this. Couple questions, but it’s merge-able so I’m merging it.

src/bun.js/bindings/ImportMetaObject.cpp
otgerrogla: Tried to run this test, but the parser complains with `SyntaxError: import.meta is only valid inside modules.`, and I couldn't find any way to force it to run as an ESM, other than using node:vm modules which are not yet implemented.

src/bun.js/bindings/ImportMetaObject.cpp
otgerrogla: The property of the env map/object itself (`Object.getOwnPropertyDescriptor(import.meta, 'env')`) I'd say so.
The values inside of the map can be writen to, I just checked it

src/bun.js/bindings/ImportMetaObject.cpp
Jarred-Sumner: you could pass the import.meta.env object to the node:vm instance

PR #7099, opened 2023-11-14
Support import.meta.env on build time

otgerrogla:
### What does this PR do?

Implements support for build-time replacement of accesses to `import.meta.env` (see issue #4667, note that #7094 did only implement support for runtime).

Currently only exposes the env variables that either start with `BUN_` or  `VITE_`, and all others appear as if they were not defined, following [vite's rationale](https://vitejs.dev/guide/env-and-mode.html#env-files):
> To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code.

---

- [x] Code changes

I wrote automated tests

### DIFF
--- a/test/bundler/esbuild/default.test.ts
+++ b/test/bundler/esbuild/default.test.ts
@@ -2786,6 +2786,37 @@ describe("bundler", () => {
       bunArgs: ["--define", 'import.meta.url="url_here"', "--define", 'import.meta.path="path_here"'],
     },
   });
+  itBundled("default/ImportMetaEnv", {
+    files: {
+      "/entry.js":
+        "console.log([" +
+        "import.meta.env.BUN_A," + // basic access
+        "import.meta['env'].BUN_A," + // access the env object with ['']
+        "import.meta['env']['BUN_A']," + // access both with ['']
+        "import.meta.env['BUN_A']," + // access env with ['']
+        'import.meta.env["BUN_A"],' + // access env with [""]
+        "import.meta.env[`BUN_A`]," + // access env with [``]
+        "import.meta.env.BUN_A.length," + // access a subproperty
+        "import.meta.env['BUN_A'].length," +
+        "import.meta.env.BUN_A['length']," +
+        "import.meta.env['BUN_A']['length']," +
+        "import.meta.env.BUN_B," +
+        "import.meta.env.BUN_C," + // access undefined env, should return undefined
+        "import.meta.env.BUN_C?.length," + // access undefined env subproperty
+        "import.meta.env.SECRET," + // access an env which should be hidden and act as if didn't exist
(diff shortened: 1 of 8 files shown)


codehz:
what's about those variables?

> Vite exposes env variables on the special import.meta.env object. Some built-in variables are available in all cases:
>
> `import.meta.env.MODE`: {string} the [mode](https://vitejs.dev/guide/env-and-mode.html#modes) the app is running in.
>
> `import.meta.env.BASE_URL`: {string} the base url the app is being served from. This is determined by the [base config option](https://vitejs.dev/config/shared-options#base).
>
> `import.meta.env.PROD`: {boolean} whether the app is running in production.
>
> `import.meta.env.DEV`: {boolean} whether the app is running in development (always the opposite of import.meta.env.PROD)
>
> `import.meta.env.SSR`: {boolean} whether the app is running in the [server](https://vitejs.dev/guide/ssr#conditional-logic).

can it have default value?
like `PROD = Bun.env.NODE_ENV === 'production'`, `DEV = Bun.env.NODE_ENV === 'development'`, `SSR = target !== 'browser'`

Jarred-Sumner:
The environment variables should be copied into the `DefineData` like what we currently do with `process.env` (which we shouldn't really be doing)

otgerrogla:
Wouldn't this be slower? As in having to do extra string comparisons for almost all identifiers etc.

codehz:
also you cannot actually define for something like `import.meta.env?.MODE` which is widely used in code bases

https://github.com/search?q=import.meta.env?.mode&type=code

hyuzea:
maybe its ok to expose all variables to `--target` other than `browser`

otgerrogla:
Or perhaps disable the replacement of `import.meta.env` for targets other than browser?

test/bundler/esbuild/default.test.ts
```diff
@@ -2786,6 +2786,37 @@ describe("bundler", () => {
       bunArgs: ["--define", 'import.meta.url="url_here"', "--define", 'import.meta.path="path_here"'],
     },
   });
+  itBundled("default/ImportMetaEnv", {
```
paperclover, line 2789: can you add a second test that uses `target: 'bun'`. i believe by default that target will inline all environment variables. i'm not sure what should be followed with import.meta.env, but i would like a test for it.

paperclover:
> also you cannot actually define for something like `import.meta.env?.MODE` which is widely used in code bases
> 
> https://github.com/search?q=import.meta.env?.mode&type=code

i remember when writing the bundler tests initially, i noticed --define does not handle some of these cases, this could be considered a different bug than getting `import.meta.env` to work.

Jarred-Sumner:
Closing due to bitrot

PR #7107, opened 2023-11-14
fix types for bun test expect

mProjectsCode:
### What does this PR do?

This PR builds on https://github.com/oven-sh/bun/pull/6634 and fixes the types for `expect` by overloading the `expect` function. This way of fixing it keeps `expect().fail()` possible. 

I added a new test for this but I couldn't figure out how to run the tests.

There is an extra file that changed in this PR because I ran prettier.

- [ ] Code changes


mProjectsCode:
any update on this?

packages/bun-types/bun-test.d.ts
```diff
@@ -511,6 +511,8 @@ declare module "bun:test" {
    * @param actual the actual value
    */
   export const expect: {
+    <T = undefined>(): Expect<T>;
+    <T = unknown>(actual: T): Expect<T>;
     <T = unknown>(actual?: T): Expect<T>;
     any: (
```
jdalton, line 517: Jest expect:
https://github.com/jestjs/jest/blob/8c78a08c6e75c1a0c4447aa8a0c61ad9e395e16f/packages/expect/src/types.ts

DefTyped:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/351120329b25603d3f2005d9be04725268578d80/types/jest/v16/index.d.ts#L14C39-L14C50

Chai:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chai/index.d.ts#L87

Issue #7031, opened 2023-11-10
Express.js middlewares are reused when creating a second Express app


1cedsoda:
### What version of Bun is running?

1.0.11+f7f6233ea

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

```javascript
// bun version 1.0.11+f7f6233ea
// node version v18.18.2

import express from "express";

async function newServer() {
  const app = express();
  const x = Math.random();
  app.use((req, res, next) => {
    console.log("middleware x", x);
    res.send("Hello World");
  });
  return new Promise((resolve, reject) => {
    const listener = app.listen(9999, () => {
      resolve(listener);
    });
  });
}

async function main() {
  const listener = await newServer();
  await fetch("http://localhost:9999");
  await listener.close();

  const listener2 = await newServer();
  await fetch("http://localhost:9999");
  await listener2.close();
}

main();
```

Running `bun x.js`
```bash
middleware x 0.6306528523654233
middleware x 0.6306528523654233
```
Expected: numbers to be different
Got: numbers are same

Running `node x.js`
```bash 
middleware x 0.63468082718277
middleware x 0.16123682051243216
```
Expected: numbers to be different
Got:  numbers are different

### What is the expected behavior?

Bun: numbers should be different (middlewares are redeclared)
Express: numbers should be different (middlewares are redeclared)

### What do you see instead?

Bun: numbers are the same (middlewares are reused)
Express: numbers should be different (middlewares are redeclared

### Findings
In Bun the express middlewares seem to be reused.

### Additional information

node version `v18.18.2`
express version `4.18.2`

Electroid:
This is likely a difference in behaviour for when listening to the same port twice. We will fix it.

1cedsoda:
Can confirm, it only happens if the same port is used multiple times.

liz3:
I can confirm that calling `server.stop(true)` instead of `server.stop()` here: https://github.com/oven-sh/bun/blob/main/src/js/node/http.ts#L446
does fix the issue, @Electroid is this something which would be okay to do?

Electroid:
> I can confirm that calling `server.stop(true)` instead of `server.stop()` here: https://github.com/oven-sh/bun/blob/main/src/js/node/http.ts#L446 does fix the issue, @Electroid is this something which would be okay to do?

Yep, that should be a good fix.

1cedsoda:
Still appears in `1.1.0+5903a6141`

[...]

Issue #6632, opened 2023-10-21
Bun.serve `server.stop()` doesn't stop server when same port is reused/reuses handlers


liz3:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

darwin

### What steps can reproduce the bug?

Run this script through bun once with the port line static(staying 3000) and once with the incrementing `..c++`.
```js
const sleep = (t) => new Promise((r) => setTimeout(r, t));

const fetchServer = async (server, opts = null) => {
    return fetch(`http://${server.hostname}:${server.port}/`, opts);
};

const tests = [
    {
        opts: {
            exclusive: true,
            async fetch(request) {
                await sleep(2000);
                // throw new Error("this is my error");
                return new Response("first request returning a normal string response");
            },
            error(err){
                return new Response(err.message, {
                    status: 500
                })
            }
        },
        async run(server) {
            const res = await fetchServer(server);
            console.log(res.status)
            const t = await res.text();
            console.log("body", t.length, t.substr(0, 200))
        },
    },
    {
        opts: {
            exclusive: true,
            async fetch(request) {
                await sleep(2000);
                throw new Error("this is the second request returning 204 through an error");
            },
            error(err){
                return new Response(err.message, {
                    status: 204
                })
            }
        },
        async run(server) {
            const res = await fetchServer(server);
            console.log(res.status)
            const t = await res.text();
            console.log("body", t.length, t.substr(0, 200))

        },
    }
];


let c = 0;
for (const test of tests) {
    const port = 3000
    // const port = 3000 + (c++)
    try {
        const server = await Bun.serve({...test.opts, port});
        console.log(server)
        await test.run(server);
        server.stop();
        await sleep(500);
    } catch (err) {
        console.log("uncaught test error", err);
    }
    console.log("====")
}
```

### What is the expected behavior?

It should stop the server entirely through `server.stop()` and restart on the next iteration,
but instead the second time `Bun.serve` gets called it seams to reuse the old instance.

### What do you see instead?

Bun reuses the existing instance of the server including its handlers if the port is the same.

```
HTTPServer {
  address: {
    address: "::",
    family: "IPv6",
    port: 3000
  },
  development: false,
  fetch: [Function: fetch],
  hostname: "localhost",
  id: "",
  pendingRequests: 0,
  pendingWebSockets: 0,
  port: 3000,
  protocol: "http",
  publish: [Function: publish],
[...]

PR #7126, opened 2023-11-15
fix: force close Bun.serve in node http listener

liz3:
This forces the app close within the Bun.serve listener for node http, if we don't do this, the issue appears where if you instantly recreate a listener on the same port, its hitting the old server causing weird issues.

Fixes: https://github.com/oven-sh/bun/issues/7031
Related: https://github.com/oven-sh/bun/issues/6632

### What does this PR do?

- [x] Code changes

### DIFF
--- a/test/js/third_party/body-parser/express-body-parser-test.test.ts
+++ b/test/js/third_party/body-parser/express-body-parser-test.test.ts
@@ -23,7 +23,7 @@ test("iconv works", () => {
 
 // https://github.com/oven-sh/bun/issues/7031
 test("doesn't reuse listeners", async () => {
-  const PORT = 9999;
+  let PORT = 0;
   const newServer = async () => {
     const app = express();
     const x = Math.random();
@@ -38,12 +38,14 @@ test("doesn't reuse listeners", async () => {
     });
   };
   const listener = await newServer();
-  let r = await fetch("http://localhost:9999");
+  const addr = listener.address();
+  PORT = addr.port;
+  let r = await fetch(`http://localhost:${PORT}`);
   const text1 = await r.text();
   await listener.close();
 
   const listener2 = await newServer();
-  r = await fetch("http://localhost:9999");
+  r = await fetch(`http://localhost:${PORT}`);
   const text2 = await r.text();
   await listener2.close();
 
--- a/test/js/third_party/body-parser/express-body-parser-test.test.ts
+++ b/test/js/third_party/body-parser/express-body-parser-test.test.ts
@@ -21,6 +21,35 @@ test("iconv works", () => {
   expect(iconv.encodingExists("us-ascii")).toBe(true);
 });
 
+// https://github.com/oven-sh/bun/issues/7031
+test("doesn't reuse listeners", async () => {
+  const PORT = 9999;
+  const newServer = async () => {
+    const app = express();
+    const x = Math.random();
+    app.use((req, res, next) => {
+      res.send(x.toString());
+      lastX = x;
+    });
+    return new Promise((resolve, reject) => {
+      const listener = app.listen(PORT, () => {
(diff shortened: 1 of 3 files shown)


test/js/third_party/body-parser/express-body-parser-test.test.ts
```diff
@@ -21,6 +21,35 @@ test("iconv works", () => {
   expect(iconv.encodingExists("us-ascii")).toBe(true);
 });
 
+// https://github.com/oven-sh/bun/issues/7031
+test("doesn't reuse listeners", async () => {
+  const PORT = 9999;
```
Electroid: Instead of a hard-coded port, use port `0` to create the first server. Then use the `.address()` getter to get the port, then use that to listen the 2nd time.

liz3: https://github.com/oven-sh/bun/pull/7126/commits/975dbb1b88654fc4bbf54cb9131da1c210244715

Jarred-Sumner:
Does Node halt all currently in-flight requests?

liz3:
> Does Node halt all currently in-flight requests?

no doesn't seam like they do. fetching while the handler closes does not seam to abort the request

Jarred-Sumner:
server.stop(true) halts in-flight http requests, so this wouldn't work

liz3:
@Jarred-Sumner this diff works too, the issue isn't that there are pending requests but that the gc did not collect the server yet, is that a approach which would be okay? 
image

Jarred-Sumner:
I don't think that's the issue either. I think the issue is because we defer closing the server until the next tick of the event loop, which means if you call it twice, it will only close the previous one later. It doesn't explain why the handlers are pointing to the wrong server though.

liz3:
> I don't think that's the issue either. I think the issue is because we defer closing the server until the next tick of the event loop, which means if you call it twice, it will only close the previous one later. It doesn't explain why the handlers are pointing to the wrong server though.

mh okay, makes sense but we can keep that in mind as a workaround since it does solve the issue. But maybe then the gc would visit the already deinited server.
Strictly speaking i think the server.stop and creation of the second server do happen on the same tick since internally we don't await the serve but from my issue awaiting it does not fix it.

schettn:
> server.stop(true) halts in-flight http requests, so this wouldn't work


Hey,

using `.stop(true)` works for me. But I did not find any documentation about this. Is there some; is this even a proper public API?

```ts
 let serve = async () => {
    const app = await loadApp(filePath)

    if (server) {
      server.stop(true)
      server = null
    }

    if (!server) {
      // @ts-ignore
      server = Bun.serve({...app, port: options.port})

      if (server) {
        // With color
        console.log(
          '\x1b[32m%s\x1b[0m',
          `Listening on http://${server.hostname}:${server.port}`
        )
      }
    }
  }
```

liz3:
> > server.stop(true) halts in-flight http requests, so this wouldn't work
> 
> Hey,
> 
> using `.stop(true)` works for me. But I did not find any documentation about this. Is there some; is this even a proper public API?
> 
> ```ts
>  let serve = async () => {
>     const app = await loadApp(filePath)
> 
>     if (server) {
>       server.stop(true)
>       server = null
>     }
> 
>     if (!server) {
>       // @ts-ignore
>       server = Bun.serve({...app, port: options.port})
> 
>       if (server) {
>         // With color
>         console.log(
>           '\x1b[32m%s\x1b[0m',
>           `Listening on http://${server.hostname}:${server.port}`
>         )
>       }
>     }
>   }
> ```

Yes you can use `stop(true)`, it does interrupt all in flight requests though and isn't the definite solution for the issue but if it solves your issue for now you can use it.

1cedsoda:
Any updates?

PR #7133, opened 2023-11-15
MINOR : Update gitignore-for-init removing duplicate .cache pattern

Pierre-Mike:
'.cache' pattern is defined more than once

### What does this PR do?

`Warning:(137, 1) '.cache' pattern is defined more than once`

removing this error.


Electroid (approved):
Thanks!

PR #7141, opened 2023-11-15
fix(tests+sockets) make some tests less flasky and postgres fixes

cirospaciari:
### What does this PR do?

- [x] Code changes
- [x] More reliable tests 

Existing tests

### DIFF
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -200,12 +200,14 @@ const Handlers = struct {
 
     pub fn resolvePromise(this: *Handlers, value: JSValue) void {
         const promise = this.promise.trySwap() orelse return;
-        promise.asAnyPromise().?.resolve(this.globalObject, value);
+        const anyPromise = promise.asAnyPromise() orelse return;
+        anyPromise.resolve(this.globalObject, value);
     }
 
     pub fn rejectPromise(this: *Handlers, value: JSValue) bool {
         const promise = this.promise.trySwap() orelse return false;
-        promise.asAnyPromise().?.reject(this.globalObject, value);
+        const anyPromise = promise.asAnyPromise() orelse return false;
+        anyPromise.reject(this.globalObject, value);
         return true;
     }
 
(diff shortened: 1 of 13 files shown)


test/js/web/fetch/fetch-gzip.test.ts
```diff
@@ -18,7 +18,7 @@ it("fetch() with a buffered gzip response works (one chunk)", async () => {
   });
   gcTick(true);
 
-  const res = await fetch(`http://${server.hostname}:${server.port}`, { verbose: true });
+  const res = await fetch(server.url, { verbose: true });
```
Jarred-Sumner, line 21: There’s no way this makes the test more reliable 

Easier to write though

Electroid: That test will fail if the `hostname` is an IPv6 address (in its old form)

cirospaciari: sometimes the server get a IPV6 and the concat will be a invalid url because will need `[::1]` not only `::1`

Electroid: That's one of the big reasons for `server.url` to exist. We should audit all usages of `server.hostname` + `server.port` in our tests and convert them to `server.url`

cirospaciari: > That's one of the big reasons for `server.url` to exist. We should audit all usages of `server.hostname` + `server.port` in our tests and convert them to `server.url`

Im doing that, will push soon

src/bun.js/api/bun/socket.zig
```diff
@@ -1495,6 +1497,7 @@ fn NewSocket(comptime ssl: bool) type {
         pub fn onClose(this: *This, socket: Socket, err: c_int, _: ?*anyopaque) void {
             JSC.markBinding(@src());
             log("onClose", .{});
+            if (this.detached) return;
```
Jarred-Sumner: shouldn't we still call the close callback here? `detached` refers to the TCP socket being connected, not the handlers

Jarred-Sumner: i.e. if you call `.close`, it should run the `close` callback

cirospaciari: Yeah you are right let me remove this

Fixed on https://github.com/oven-sh/bun/pull/7141/commits/ec626e19b3b457f88548421d6624343a5bc04e5e

Jarred-Sumner:
Closing due to bitrot

PR #7145, opened 2023-11-15
Fix IncomingMessage class to properly handle pending chunks

zenshidev:
### What does this PR do?

There is an issue with IncomingMessage class, where if you `.pipe()` its result to other stream, it hangs indefinitely when number of chunks is > 3.
I have added a test in `node-http.test.ts` which demonstrates this issue.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/cli/gitignore-for-init
+++ b/src/cli/gitignore-for-init
@@ -10,6 +10,10 @@ yarn-error.log*
 lerna-debug.log*
 .pnpm-debug.log*
 
+# Caches 
+
+.cache
+
 # Diagnostic reports (https://nodejs.org/api/report.html)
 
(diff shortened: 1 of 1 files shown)


test/js/node/http/node-http.test.ts
```diff
@@ -1548,6 +1546,61 @@ it("IncomingMessage with a RequestLike object", () => {
   ]);
 });
 
+describe("IncomingMessage", () => {
+  export class Collector extends Writable {
```
Jarred-Sumner: This is not valid syntax

zenshidev: Right! Fixed :)

PR #7150, opened 2023-11-16
fix typo in help menu

rupurt:
### What does this PR do?

Fix typo in help menu `more more` -> `or more`


Jarred-Sumner:
Thank you

Issue #7153, opened 2023-11-16
Server from http is not created correctly


HK-SHAO:
### What version of Bun is running?

1.0.11+f7f6233ea

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

`index.ts`

```ts
console.log("Hello via Bun!");

import { Server } from "http";

console.log(`----------------------`);
console.log(`const server1 = new Server({ key: "123" });`);
const server1 = new Server({ key: "123" });
console.log(`----------------------`);
console.log(`const server2 = new Server({ key: ["123"] });`);
const server2 = new Server({ key: ["123"] });
```

```bash
bun index.ts
```

### What is the expected behavior?

Create two servers normally.

```log
Hello via Bun!
----------------------
const server1 = new Server({ key: "123" });
----------------------
const server2 = new Server({ key: ["123"] });
```

### What do you see instead?

```log
Hello via Bun!
----------------------
const server1 = new Server({ key: "123" });
----------------------
const server2 = new Server({ key: ["123"] });
281 |   #tls;
282 |   #is_tls = !1;
283 |   listening = !1;
284 |   serverName;
285 | 
286 |   constructor(options, callback) {
                             ^
TypeError: key argument must be an string, Buffer, TypedArray, BunFile or an array containing string, Buffer, TypedArray or BunFile
      at new Server (node:http:286:26)
      at index.ts:10:16
```

### Additional information

Seems to have checked TLSArray incorrectly.

PR #7154, opened 2023-11-16
Fix the TypeError of `isValidTLSArray` for http

HK-SHAO:
fix: https://github.com/oven-sh/bun/issues/7153

### What does this PR do?

- [x] Code changes

Code from: https://github.com/oven-sh/bun/blob/85c997513d2b05acab21e758512954273e3a589a/src/js/node/tls.js#L35

It should be consistent with the one in: https://github.com/oven-sh/bun/blob/85c997513d2b05acab21e758512954273e3a589a/src/js/node/http.ts#L95

`index.ts`

```ts
console.log("Hello via Bun!");

import { Server } from "http";
import { isTypedArray } from "util/types";

function isValidTLSArray1(obj) {
    if (typeof obj === "string" || isTypedArray(obj) || obj instanceof ArrayBuffer || obj instanceof Blob) return true;
    if (Array.isArray(obj)) {
        for (var i = 0; i < obj.length; i++) {
            if (typeof obj !== "string" && !isTypedArray(obj) && !(obj instanceof ArrayBuffer) && !(obj instanceof Blob))
                return false;
        }
        return true;
    }
}

function isValidTLSArray2(obj) {
    if (typeof obj === "string" || isTypedArray(obj) || obj instanceof ArrayBuffer || obj instanceof Blob) return true;
    if (Array.isArray(obj)) {
        for (var i = 0; i < obj.length; i++) {
            const item = obj[i];
            if (typeof item !== "string" && !isTypedArray(item) && !(item instanceof ArrayBuffer) && !(item instanceof Blob))
                return false;
        }
        return true;
    }
    return false;
}
console.log(`----------------------`);
console.log(`isValidTLSArray1("")`, isValidTLSArray1(""));
console.log(`isValidTLSArray1([])`, isValidTLSArray1([]));
console.log(`isValidTLSArray1([""])`, isValidTLSArray1([""]));
console.log(`isValidTLSArray2("")`, isValidTLSArray2(""));
console.log(`isValidTLSArray2([])`, isValidTLSArray2([]));
console.log(`isValidTLSArray2([""])`, isValidTLSArray2([""]));
console.log(`----------------------`);
console.log(`const server1 = new Server({ key: "123" });`);
const server1 = new Server({ key: "123" });
console.log(`----------------------`);
console.log(`const server2 = new Server({ key: ["123"] });`);
const server2 = new Server({ key: ["123"] });
```

```log
Hello via Bun!
----------------------
isValidTLSArray1("") true
isValidTLSArray1([]) true
isValidTLSArray1([""]) false
isValidTLSArray2("") true
isValidTLSArray2([]) true
isValidTLSArray2([""]) true
----------------------
const server1 = new Server({ key: "123" });
----------------------
const server2 = new Server({ key: ["123"] });
281 |   #tls;
282 |   #is_tls = !1;
283 |   listening = !1;
284 |   serverName;
285 | 
286 |   constructor(options, callback) {
                             ^
TypeError: key argument must be an string, Buffer, TypedArray, BunFile or an array containing string, Buffer, TypedArray or BunFile
      at new Server (node:http:286:26)
      at index.ts:41:16
```

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -96,11 +96,13 @@ function isValidTLSArray(obj) {
   if (typeof obj === "string" || isTypedArray(obj) || obj instanceof ArrayBuffer || obj instanceof Blob) return true;
   if (Array.isArray(obj)) {
     for (var i = 0; i < obj.length; i++) {
-      if (typeof obj !== "string" && !isTypedArray(obj) && !(obj instanceof ArrayBuffer) && !(obj instanceof Blob))
+      const item = obj[i];
+      if (typeof item !== "string" && !isTypedArray(item) && !(item instanceof ArrayBuffer) && !(item instanceof Blob))
         return false;
     }
     return true;
   }
+  return false;
 }
 
 class ERR_INVALID_ARG_TYPE extends TypeError {


Jarred-Sumner:
Thank you

Issue #7159, opened 2023-11-16
Incorrect return type for `Bun.serve({ fetch })`


Didas-git:
### What version of Bun is running?

1.0.12+85c997513

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

```ts
const server = Bun.serve({
    fetch(req, server) {
        server.upgrade(req);
    }
})
```

### What is the expected behavior?

No type errors

### What do you see instead?

```
Type '(this: Server, req: Request, server: Server) => void' is not assignable to type '((this: Server, request: Request, server: Server) => Response | Promise<Response>) | ((this: Server, request: Request, server: Server) => Response | Promise<...>) | ((this: Server, request: Request, server: Server) => Response | ... 1 more ... | undefined) | ((this: Server, request: Request, server: Server) => Respo...'.
  Type '(this: Server, req: Request, server: Server) => void' is not assignable to type '(this: Server, request: Request, server: Server) => Response | Promise<Response>'.
    Type 'void' is not assignable to type 'Response | Promise<Response>'.
```

### Additional information

If this is intended then all the docs and examples need to be updated and should mention this

PR #7160, opened 2023-11-16
Change `Bun.serve({ fetch })` types to allow `void` when using websockets

Didas-git:
### What does this PR do?

Since #7159 recieved the `typescript` tag i assumed that this was not intentional so i changed the fetch types so you can return whatever you want/not return.

- [ ] Code changes

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -1915,7 +1915,7 @@ declare module "bun" {
       this: Server,
       request: Request,
       server: Server,
-    ): any | Promise<any>;
+    ): Response | Promise<Response>;
   }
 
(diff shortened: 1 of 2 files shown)


paperclover (changes requested):
this is not correct. the fetch function should stay as returning Response, but we should also allow `void` only if websocket handlers are provided

paperclover:
this looks much better. i should've mentioned in the first review but also can you add a type test inside of `serve.test-d.ts`

Didas-git:
> this looks much better. i should've mentioned in the first review but also can you add a type test inside of `serve.test-d.ts`

Added a simple test, tell me if there is something else.

paperclover (approved):

PR #7171, opened 2023-11-17
Implement process.report.getReport()

Jarred-Sumner:
### What does this PR do?

Implements `process.report.getReport()`

There is a test. It's not very good though.

### DIFF
--- a/src/bun.js/bindings/wtf-bindings.cpp
+++ b/src/bun.js/bindings/wtf-bindings.cpp
@@ -304,5 +304,6 @@ size_t toISOString(JSC::VM& vm, double date, char in[64])
         return 0;
 
     return charactersWritten;
+}
 
 }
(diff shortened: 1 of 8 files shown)


src/bun.js/bindings/wtf-bindings.cpp
```diff
@@ -273,4 +273,36 @@ String base64URLEncodeToString(Vector<uint8_t> data)
...
+
+    // Maximum amount of space we need in buffer: 7 (max. digits in year) + 2 * 5 (2 characters each for month, day, hour, minute, second) + 4 (. + 3 digits for milliseconds)
+    // 6 for formatting and one for null termination = 28. We add one extra character to allow us to force null termination.
+    char buffer[28];
+    // If the year is outside the bounds of 0 and 9999 inclusive we want to use the extended year format (ES 15.9.1.15.1).
+    int ms = static_cast<int>(fmod(date, msPerSecond));
+    if (ms < 0)
+        ms += msPerSecond;
+
+    int charactersWritten;
```
liz3, line 294: this can just be unsigned i think?

PR #7172, opened 2023-11-17
fix(process): enable SIGKILL and SIGSTOP mappings

antongolub:
### What does this PR do?

This PR enables _fast_ child process killing (execa, zx scripts) via `SIGKILL`

```bash
p.kill("SIGKILL")
...
RangeError: Unknown signal name
      at ~/zx-extra/node_modules/zx/build/core.js:251:16
      at processTicksAndRejections (:61:7
```

I've tweaked a test:
```bash
bun run build
bun-debug test ./test/js/node/process/process.test.js
```


antongolub:
@dylan-conway,

As I can see, you've introduced this behaviour. Why the mapping was disabled?

Jarred-Sumner:
I think I misread `SIGKILL` as `SIGILL` (when we discussed IRL)

Jarred-Sumner (approved):
LGTM

Electroid:
SIGSTOP is also commented out.

antongolub:
@Jarred-Sumner

Should we enable `SIGSTOP` too?

dylan-conway (changes requested):
`signalNameToNumberMap` is also used for `process.on(<signal name>, ...);` and it's not possible to set handlers on SIGKILL and SIGSTOP. We can uncomment both of them but will need to add something like this in `onDidChangeListeners` on line 804 and 752:
```cpp
if (auto signalNumber = signalNameToNumberMap->get(eventName.string())) {
            if (signalNumber != SIGKILL && signalNumber != SIGSTOP) {
...
```

dylan-conway (approved):
LGTM. will merge after ci runs

dylan-conway:
Thank you

Issue #7165, opened 2023-11-17
fs does not write base64 strings to files correctly


KhalfaniW:
### What version of Bun is running?

1.0.12

### What platform is your computer?

Linux 6.2.0-36-generic x86_64 x86_64   

### What steps can reproduce the bug?

writing file using base64 works on node to create an image from a base64 string

```js
import fs from "fs";                                                                                                                                                                          
const base64data = fs.readFileSync(`./image.jpg`, "base64");                                                                                                                                   
fs.writeFileSync(`./new_image.jpg`,base64data, "base64");                                                                                                                                     
```

### What is the expected behavior?

original file type:
```
$ file image.jpg
image.jpg: JPEG image data, ....
```
expect new file to have same type 
```
$ file new_image.jpg
new_image.jpg: JPEG image data, ....
```

### What do you see instead?

```
$ file new_image.jpg
new_image.jpg: data
```

PR #7174, opened 2023-11-17
fix(fs.writeFile): correctly decode when encoding is base64

liz3:
We where not decoding base64 into the binary form when the passed encoding was base64 leading to a incorrect file content.

Fixes: https://github.com/oven-sh/bun/issues/7165

### What does this PR do?

- [x] Code changes


src/bun.js/node/node_fs.zig
```diff
@@ -4450,13 +4451,27 @@ pub const NodeFS = struct {
             .fd => |_fd| _fd,
         };
 
+        var buf = args.data.slice();
+        var written: usize = 0;
```
jdalton, line 4455:
```suggestion
        var written: usize = 0;
        var isBase64 = args.encoding == .base64;
```
For reuse?

PR #7175, opened 2023-11-17
Fix `napi_get_value_string_utf8` to match node

### LINKED ISSUE #6949: NAPI: napi_get_value_string_utf8 does not behave like nodejs

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

Darwin 23.0.0 arm64 arm

Reproduction repo: https://github.com/oguimbal/bun-napi-bug

```bash
git clone git@github.com:oguimbal/bun-napi-bug.git
cd bun-napi-bug
yarn
yarn build && node index.js && bun index.js
```

Should log the same thing as node:

```
Copied: 1
Value str: b
encoding 1
```

### What do you see instead?

```
Copied: 2
Value str: bu
encoding 3
```

### Additional information

This is a simple reproduction of the bug described in my original issue #6784 (which I closed).

To my understanding, the `bufsize` argument of `napi_get_value_string_utf8()` is not handled well in Bun. I suppose that Node takes into account the null char string terminator character in this bufsize, while Bun does not ?

oguimbal:
closes #6949

### What does this PR do?

As described in #6949, the `napi_get_value_string_utf8` function does not match Nodejs NAPI behaviour.

This PR:
- fixes this issue
- adds unit testing to check that both node & bun behave the same

I believe NAPI is not unit tested at all.
In this PR, I'm proposing a way to test coherence between bun's NAPI & node's.

It consists of a NAPI node/bun application (in `test/napi/napi-app`) that can be called by unit tests.

When tests run, it:
1) Builds the given api application 
2) Runs this application in node (with flags that depends on the test)
3) Runs this application in bun  (using the `./build/bun-debug` executable)
4) Compares both outputs to ensure they are the same.

This is obviously a very opiniated way of testing things, so I'd totaly understand if this not OK.

As a side-note, I did not check wether if this is OK with your build pipeline, as it requires both `node` and `yarn` to be available in path.

I ran those tests locally using `bun run build && bun test test/napi`.

Before fixes:

image

After fixes:
image


test/napi/napi.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    it("copies auto len", () => {
+      const result = checkSameOutput("test_napi_get_value_string_utf8_with_buffer", ["abcdef", 424242]);
+      expect(result).toEndWith("str:");
+    });
+  });
+});
+
+function checkSameOutput(test: string, args: any[]) {
+  const nodeResult = runOn("node", test, args).trim();
+  let bunResult = runOn(join(__dirname, "../../build/bun-debug"), test, args);
```
Jarred-Sumner, line 50:
```suggestion
  let bunResult = runOn(bunExe(), test, args);
```

test/napi/napi.test.ts
```diff
@@ -0,0 +1,64 @@
+import { it, expect, test, beforeAll, describe } from "bun:test";
```
Jarred-Sumner, line 1: If you add this import, it will make this a little easier:
```suggestion
import { it, expect, test, beforeAll, describe } from "bun:test";
import {bunExe, bunEnv} from "harness";
```

test/napi/napi.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  let bunResult = runOn(join(__dirname, "../../build/bun-debug"), test, args);
+  // remove all debug logs
+  bunResult = bunResult.replaceAll(/^\[\w+\].+$/gm, "").trim();
+  expect(bunResult).toBe(nodeResult);
+  return nodeResult;
+}
+
+function runOn(executable: string, test: string, args: any[]) {
+  const exec = spawnSync({
+    cmd: [executable, join(__dirname, "napi-app/main.js"), test, JSON.stringify(args)],
```
Jarred-Sumner, line 59:
```suggestion
    cmd: [executable, join(__dirname, "napi-app/main.js"), test, JSON.stringify(args)],
    env: bunEnv,
```

test/napi/napi.test.ts
oguimbal: If I do this locally, tests fail again as if it was run on the current bun version... is `bunExe()` supposed to refer to the version of bun built by the pipeline ?

Jarred-Sumner: hm. yes, is `bun-debug` available in `$PATH`?

oguimbal: yes, but it's available as `bug-debug`, not `bun`. Thus, these tests only pass locally if I run them via `bun run build && bun-debug test test/napi`  (with a lot of noisy logs, obviously).

If that's the executable that runs on your test pipeline instead of the current version of bun, then fine, lets commit that 🙃

Jarred-Sumner: The tests pass locally with `process.execPath` in use and they will not pass in CI without that.

Jarred-Sumner:
Thank you for this. I'm going to merge it and then get it running in CI a follow-up PR.

oguimbal:
Thanks !

PR #7183, opened 2023-11-17
Output verbose logs when running in Github Action with debug mode

Electroid:
### What does this PR do?

When `bun install` or `bun create` is run in a Github Action with debug mode, it will automatically show logs as if it was run using `--verbose`.

Screenshot 2023-11-17 at 9 17 00 AM

- [x] Code changes

Manually.


Jarred-Sumner:
Thank you

PR #7188, opened 2023-11-17
Update docs of TCP sockets to use IP address in hostname

PranjalAgni:
### What does this PR do?

Updating the TCP sockets docs with hostname to IP address

- [ ] Code changes


Electroid:
This is a bug in Bun, we should fix the bug.

Issue #7187, opened 2023-11-17
error message with Bun.serve for EACCESS is shown as EADDRINUSE


paperclover:
### What platform is your computer?

Linux

### What steps can reproduce the bug?

```
Bun.serve({ fetch(){}, port: 1003 });
```
 

### What is the expected behavior?

```
EACCES: permission denied 0.0.0.0:1003
```

### What do you see instead?

```
EADDRINUSE: Failed to start server. Is port 1003 in use?
```

### Additional information

i think this is an easy fix. grep for this string to see where listen failures are handled

`noinline fn onListenFailed(this: *ThisServer) void {`

grep this line for where i think we start listening
`this.app.listenWithConfig(*ThisServer, this, onListen, .{`

bun.serve is defined in src/bun.js/api/bun.zig by
`pub fn serve(`

liz3:
I don't actually think that this is very easy, because we have no indication of why the socket creation failed, i spend some time looking into this and this goes deep into usockets. the zig side only gets: "we failed to allocate the socket".

Implementing this requires passing up the error which actually occurs.

paperclover:
my bad then, removed gfi tag

PR #7191, opened 2023-11-17
fix(Bun.serve): return EACCESS when we don't have perms

liz3:
The error reported to js land when listening fails was always the same, this adds a second one for EACCESS when we are not the super user.

Fixes: https://github.com/oven-sh/bun/issues/7187

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -14,6 +14,7 @@ const IdentityContext = @import("../../identity_context.zig").IdentityContext;
 const Fs = @import("../../fs.zig");
 const Resolver = @import("../../resolver/resolver.zig");
 const ast = @import("../../import_record.zig");
+const Sys = @import("../../sys.zig");
 
 const MacroEntryPoint = bun.bundler.MacroEntryPoint;
 const logger = @import("root").bun.logger;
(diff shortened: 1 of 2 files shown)


src/bun.js/api/server.zig
```diff
@@ -5601,11 +5602,25 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
             if (error_instance == .zero) {
                 switch (this.config.address) {
                     .tcp => |tcp| {
-                        error_instance = (JSC.SystemError{
-                            .message = bun.String.init(std.fmt.bufPrint(&output_buf, "Failed to start server. Is port {d} in use?", .{tcp.port}) catch "Failed to start server"),
-                            .code = bun.String.static("EADDRINUSE"),
-                            .syscall = bun.String.static("listen"),
-                        }).toErrorInstance(this.globalThis);
+                        // this value is hard set by the underlying system, but it might be wrong for windows?
```
paperclover: windows does not have EACCESS on listening

i dont think mac does either

paperclover:
nice

src/bun.js/api/server.zig
liz3: il check

liz3: yeeee its not passing on mac

liz3:
@paperdave should be gud now

paperclover (approved):

PR #7213, opened 2023-11-19
Fixed markdown

aurorarissime:
### What does this PR do?

- [ ] Code changes

I increased a little the size of the image to make it a bit more visible


paperclover:
Why was the callout component removed? It should be using that so it renders properly

aurorarissime:
On the page it didn't rendered using callouts, that why I changed it. Where should it render? I might just have been confused

Jarred-Sumner:
Thank you, but this is deliberate. These tags are used by our static site generator. This is used in https://bun.sh/docs/cli/install and it renders successfully there.

PR #7217, opened 2023-11-19
Micro-optimization: omit instead of delete headers

Jarred-Sumner:
### What does this PR do?

`Bun.serve` bans certain headers from being sent, since they will essentially always be wrong

To ban those headers, previously we would simply delete them from the `Headers` object. This is observable - it shouldn't do that. But it also means potentially resizing the common headers array

This should in theory make it slightly faster to serve requests but honestly I didn't benchmark 

There are tests for this


Jarred-Sumner:
This seems to not have any performance impact

src/bun.js/api/server.zig
```diff
@@ -1812,16 +1812,28 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
             this: *RequestContext,
             headers: *JSC.FetchHeaders,
         ) void {
             ctxLog("writeHeaders", .{});
-            headers.fastRemove(.ContentLength);
-            headers.fastRemove(.TransferEncoding);
-            if (!ssl_enabled) headers.fastRemove(.StrictTransportSecurity);
             if (this.resp) |resp| {
-                headers.toUWSResponse(ssl_enabled, resp);
+                headers.toUWSResponse(ssl_enabled, resp, &headers_to_exclude);
```
jdalton, line 1836: How is fastRemove fast?

Jarred-Sumner: It uses headers from this list, which skip storing the original name and skip some of the normalization normally necessary 

https://github.com/zackradisic/bun/blob/main/src/bun.js/bindings/webcore/HTTPHeaderNames.h#L36

PR #7218, opened 2023-11-19
WIP: Bun-BSD - FreeBSD port

LowByteFox:
### What does this PR do?

Adds ability to compile bun under FreeBSD

- [x] Code changes

I have managed to compile all bun dependencies and bun itself and was able to run bun which gave me help screen but that's all, IO needs to be fixed so bun can read file properly
image

I wrote `freebsd_c.zig` for C on FreeBSD, `io_freebsd.zig` for async IO on Freebsd using `kevent`, fixed type on flags in `sys.zig` to match right OS specific type, it gave me compile issues when it was `Bun.Mode` as flags on FreeBSD are `i32` and not `u16` ( `std.os.mode_t` )<br>

I have added patch for mimalloc so it would successfully compile on FreeBSD<br>
I have patched couple of other files so bun's zig code compiles<br>

### TODO
- [ ] Make bun run `Hello World`


LowByteFox:
Ignore the `Unverified` badge, I don't have my gpg key to sing commits on the VM

LowByteFox:
If someone wants to give it a try and maybe resolve the issue. I'll provide you with everything

Guide [here](https://gist.github.com/LowByteFox/a4e86c64aa3eaae6f79efb293faefe74)

LowByteFox:
I fixed the issue
image

LowByteFox:
Closed so only I have to deal with future issues and everything and so it doesn't disrupt oven flow

LowByteFox:
If someone wants FreeBSD native bun, you can always point to me

Electroid:
Thank you for working on this, looks like it definitely took some work! We have our hands full for a while with Windows, but I'm sure they'll be a time when we revisit FreeBSD.

dch:
@LowByteFox if you're interested in having this (unofficial) fork included in FreeBSD ports tree, I'd be happy to take a look

LowByteFox:
Thanks, however it has to work first, compile is kind of not enough

lin72h:
Is there anyone working on FreeBSD ports?

alsosprachben:
> @alsosprachben did you find the system? I do have a FreeBSD installed, maybe I could help.

I installed it on my laptop, and got Codex ported to FreeBSD, attempted to continue the old port, but too much drifted. Attempted to port the head, but ran into a zig crash in the FreeBSD port of Zig, which is version 14. It looks like 15 is available, but Bun may not yet work for that, so I paused for now. It is difficult to share the current progress due to the amount of components changed. I did get building working to the point where zig crashes, so it is theoretically not that far away. Codex figured out by itself to use User events instead of Mach ports in porting the kqueue from Mac to FreeBSD. I was prepared to tell it to do that, so that is really impressive. I did need to guide it in the order of work that made the best sense, because it has a tendency to leave dependencies for later, which seems nuts, especially while porting. The dependencies are the crux, always.

I can get the Codex port upstream, and tell you how I installed it, and you can try that way yourself, or I can try to try to get you my environment.

yurivict:
Can somebody summarize what is preventing Bun from building on FreeBSD?

PR #7223, opened 2023-11-19
Cleanup how build errors are displayed slightly

Jarred-Sumner:
### What does this PR do?

Two things:

#### Formatting tweaks

- Removes extra newlines
- Fixes `logger.rangeOfIdentifier` to include the complete length of the identifier (usage of `defer` broke that)
- Making the selected portion colored is too intense IMO for longer lines

#####  Before:
image

#####  After:
image

#### Build-time error for unsupported ESM<>CommonJS features 

You can use `require` and `import` in the same file in Bun, but you cannot use `import` and `module.exports` in the same file. You also cannot use `exports`, `module`, `with` or top-level return.

Currently, we rely on JavaScriptCore's errors for these things, but they're really hard to read and don't give us an easy way to show a stack trace pointing to which line exactly caused it. 

This moves those errors to build-time.

#####  Input:
```js
import "abc";

with (Bun) {
  write("/tmp/file.txt", escapeHTML(file("/tmp/file.html").text()));
  console.log({
    contents: file("/tmp/file.txt").text(),
  });
}
```

#####  Before (note how it says "import call"):

image

#####  After:
image

##### Input:
```js
import _ from "lodash";
module.exports = _.constant(42);
```

##### Before:
image

##### After:
image

I wrote a few tests

### DIFF
--- a/src/logger.zig
+++ b/src/logger.zig
@@ -257,41 +257,23 @@ pub const Data = struct {
         to: anytype,
         kind: Kind,
         comptime enable_ansi_colors: bool,
-        comptime is_note: bool,
     ) !void {
         if (this.text.len == 0) return;
 
         const message_color = switch (kind) {
             .err => comptime Output.color_map.get("b").?,
-            .note => comptime Output.color_map.get("cyan").? ++ Output.color_map.get("d").?,
+            .note => comptime Output.color_map.get("blue").?,
             else => comptime Output.color_map.get("d").? ++ Output.color_map.get("b").?,
         };
 
         const color_name: string = switch (kind) {
             .err => comptime Output.color_map.get("red").?,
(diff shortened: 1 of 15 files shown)


src/logger.zig
```diff
@@ -257,95 +257,106 @@ pub const Data = struct {
...
                             if (is_colored) {
                                 try to.writeAll("\x1b[0m");
                             }
 
-                            try to.writeAll(rest_of_line[end_of_segment..]);
+                            var after = rest_of_line[middle_segment.len..];
+                            if (after.len > 40) {
+                                after = after[0..40];
+                            }
+                            try to.writeAll(after);
```
dylan-conway, line 348: do we also need a color reset here?

dylan-conway: oh, I don't think we do because it would've reset on line 341

PR #7226, opened 2023-11-19
 Fix IncomingMessage class to properly handle pending chunks

zenshidev:
### What does this PR do?

There is an issue with IncomingMessage class, where if you `.pipe()` its result to other stream, it hangs indefinitely when number of chunks is > 3.
I have added a test ("should allow piping multiple chunks to collector stream") in `node-http.test.ts` which demonstrates this issue. Other tests have been added for safety, just to make sure other behaviour still works.

This pattern is used in AWS SDK v3, or in other words, it fixes a bug where AWS SDK v3 requests occasionally hang indefinitely, where response is chunked a certain way. In my tests it affects less than 0.1% of all requests, buuut nonetheless, its extremely annoying bug.

Closed previous PR by accident, sorry for spam

- [x] Code changes

I wrote automated tests

### DIFF
--- a/test/js/node/http/node-http.test.ts
+++ b/test/js/node/http/node-http.test.ts
@@ -171,6 +171,14 @@ describe("node:http", () => {
   });
 
   describe("request", () => {
+    class Collector extends Writable {
+      public bufferedBytes: Uint8Array[] = [];
+
+      _write(chunk: Uint8Array, encoding: string, callback: () => void) {
+        this.bufferedBytes.push(chunk);
+        callback();
+      }
+    }
     function runTest(done: Function, callback: (server: Server, port: number, done: (err?: Error) => void) => void) {
       var timer;
       var server = createServer((req, res) => {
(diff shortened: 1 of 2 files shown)


test/js/node/http/node-http.test.ts
```diff
@@ -744,6 +750,119 @@ describe("node:http", () => {
...
+                const start = i * chunkSize;
+                controller.enqueue(testData.subarray(start, start + chunkSize));
+              }
+
+              controller.close();
+            },
+          }),
+          {
+            headers: {
+              "content-length": testData.length.toString(),
```
Jarred-Sumner, line 792: This content-length header will be completely ignored by the server

zenshidev: Hmmm i dont know why that would be an issue? From what I read the code content-length is only used to decide whether body should be parsed or not.

Reading body should be the same for requests and responses.

zenshidev:
Fixed in v1.0.30

PR #7234, opened 2023-11-20
fix(buildsystem): test for make presence in the buildsystem

jerome-benoit:
Trivial code change to the bun buildsystem testing for make presence


Electroid (approved):
Thanks!

PR #7235, opened 2023-11-20
refactor, perf: cleanup worker_threads TS module implementation

jerome-benoit:
### What does this PR do?

+ worker_threads TS module code cleanup: style, type
+ worker_threads test code cleanup
+ micro-optimization to webkit MessagePort event handling code

- [x] Code changes


src/js/node/worker_threads.ts
```diff
@@ -77,11 +75,16 @@ function injectFakeEmitter(Class) {
   };
 
   function EventClass(eventName) {
-    if (eventName === "error" || eventName === "messageerror") {
-      return ErrorEvent;
-    }
+    switch (eventName) {
```
paperclover, line 70: is this change really necessary?

tbh this whole file is just style changes.

src/bun.js/bindings/webcore/MessagePort.cpp
```diff
@@ -430,8 +430,7 @@ Ref<MessagePort> MessagePort::entangle(ScriptExecutionContext& context, Transfer
 bool MessagePort::addEventListener(const AtomString& eventType, Ref<EventListener>&& listener, const AddEventListenerOptions& options)
 {
     if (eventType == eventNames().messageEvent) {
-        if (listener->isAttribute())
-            start();
+        start();
```
paperclover: this change fixes the node:worker_threads module but it breaks cases using the raw `Worker` class.

paperclover (changes requested):

src/js/node/worker_threads.ts
jerome-benoit: As part of the bug fix, they are not necessary but harmless. They just make the coding style consistent in that file with the rest of the JS code in bun. 
They can be carved out that PR and be part of another code cleanup PR.

src/bun.js/bindings/webcore/MessagePort.cpp
jerome-benoit: How exactly? Is there a code sample that I can test? 

The message port is started in any case at message event listener registration, so I need more to understand the issue with that change.

jerome-benoit: The current webkit code for MessagePort is designed to follow the Web API and can't handle by default event listener stacked registration on attribute

So, 
+ Revert the fix
+ Change the PR purpose: 
  - worker_threads TS module code cleanup
  - micro-optimization to webkit MessagePort event handling code  

I do not see why code cleanup should be lost 

A proper fix could be: 
+ Make a dedicated TS proxy class for worker_threads MessagePort and MessageChannel following/reusing the node EventEmitter API
+ Reuse it as parentPort in worker_threads module

The performance would be better if the implementation is part of the webkit glue. It's possible, the webkit EventEmitter class already support stacked listener registration 

In any case, that's the purpose of another PR.

PR #7238, opened 2023-11-20
fix(worker_threads): do not terminate worker at parentPort closing

jerome-benoit:
### What does this PR do?

Until a proper way to implement parent port in node:worker_threads is done, avoid to terminate the worker by invoking process.exit() in parentPort close() and make it a no-op. 

- [x] Code changes


src/js/node/worker_threads.ts
```diff
@@ -135,9 +135,7 @@ function fakeParentPort() {
   });
 
   Object.defineProperty(fake, "close", {
-    value() {
-      return process.exit(0);
-    },
+    value() {},
```
Electroid, line 139: Could you add a todo comment here, so we know to revisit this?

jerome-benoit: I've added a TODO comment at the top of the `fakeParentPort()` function. You can change its content if it does not fit.

src/js/node/worker_threads.ts
```diff
@@ -135,9 +136,7 @@ function fakeParentPort() {
   });
 
   Object.defineProperty(fake, "close", {
-    value() {
-      return process.exit(0);
```
Jarred-Sumner, line 139: `process.exit` in a `Worker` does not crash the process. It terminates the Worker and sets the exit code accordingly, but does not exit

jerome-benoit: The worker termination, clean or not, is not the expected behavior at worker_threads parent port closing.

paperclover: only observable behavior i can find in closing the parent port is simply being unable to send more messages from it.

we can make our emulation better by simply having an internal flag `let closed = false` and have close set it to true, and prevent messages. but tbh this is good enough.

paperclover (approved):

src/js/node/worker_threads.ts
jerome-benoit: I think the parent port emulation in node:worker_threads design needs thoughts. And I unfortunately does not enough knowledge on bun internals to propose a scalable message port implementation (unlike the node.js one) and a complete parent port emulation for node:worker_threads compatibility bun module. At least for now.

paperclover:
merging because it is better than what bun does right now but it isnt perfect emulation

PR #7257, opened 2023-11-22
fix issue with snapshots in bun test

Jarred-Sumner:
### What does this PR do?

fix an issue causing `bun test` to mark snapshots as newly added

This isn't done yet

### DIFF
--- a/src/bun.js/test/expect.zig
+++ b/src/bun.js/test/expect.zig
@@ -221,44 +221,38 @@ pub const Expect = struct {
     pub fn getSnapshotName(this: *Expect, allocator: std.mem.Allocator, hint: string) ![]const u8 {
         const test_name = this.scope.tests.items[this.test_id].label;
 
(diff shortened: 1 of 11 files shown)


src/bun.js/test/expect.zig
```diff
@@ -221,44 +221,38 @@ pub const Expect = struct {
...
-            bun.copy(u8, buf[index..], test_name);
-        }
-        // copy describe scopes in reverse order
-        curr_scope = this.scope;
         while (curr_scope) |scope| {
             if (scope.label.len > 0) {
-                index -= scope.label.len + 1;
-                bun.copy(u8, buf[index..], scope.label);
-                buf[index + scope.label.len] = ' ';
+                builder.write(scope.label);
```
dylan-conway, line 244: this will copy describe scope names in child -> parent order. It needs to go parent -> child

Issue #5426, opened 2023-09-15
SyntaxError: Indirectly exported binding name 'AnyArena' is not found.


dpeek:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

```TypeScript
// test.ts
const capnp = require('capnp-ts');
console.log(capnp);
```

```bash
bun test.ts
```

### What is the expected behavior?

No error

### What do you see instead?

```
SyntaxError: Indirectly exported binding name 'AnyArena' is not found.
      at loadCJS2ESM (:1:20)
      at requireESM (:1:20)
      at require (:1:20)
      at .../node_modules/capnp-ts/src/index.js:7:4
      at globalThis (.../node_modules/capnp-ts/src/index.js:52:106)
      at require (:1:20)
      at /Users/dpeek/code/estii/packages/api/test.ts:1:6
      at globalThis (.../packages/api/test.ts:2:17)
```

### Additional information

Seems related to #1852, although I was unable to create a minimal repro.

Digging into the library, `AnyArena` is exported as a type that is a union of two classes exported from other modules.

```TypeScript
import { MultiSegmentArena } from "./multi-segment-arena";
import { SingleSegmentArena } from "./single-segment-arena";

export type AnyArena = MultiSegmentArena | SingleSegmentArena;
```

See https://github.com/jdiaz5513/capnp-ts/blob/108850ece76b33755d552f78eac736055ec8f1b4/packages/capnp-ts/src/serialization/arena/any-arena.ts

zlooun:
Same ussue

M-jerez:
Similar error here

Bellow re-export does not work
```ts
export {SerializedTypes} from '@deepkit/type';
```

Workaround:
```ts
import {SerializedTypes} from '@deepkit/type';
export type {SerializedTypes};
```

similar to #74

dpeek:
I did a bit more digging into this today. It seems like the issue stems from `capnp-ts` shipping TypeScript source along side compiled JS and type definitions.

There's no mention of `AnyArena` in the JS source (it's a type export) which means bun is parsing the TS source instead.

The initial `require('capnp-ts')` loads `node_modules/capnp-ts/src/index.js` which does `require("./serialization")`

[...]

PR #7259, opened 2023-11-22
Load `*.{m,c}ts{x}` last in node_modules

Jarred-Sumner:
### What does this PR do?

Bun currently loads `.ts` and `.tsx` files first. This usually works great in local development. This doesn't always work in `node_modules` because some projects will need special tsconfig.json options which we do not support to work:
- ts transformer plugins
- using `import {Foo}` rather than `import {type Foo}` or `export {Foo}` rather than `export {type Foo}`

This fixes that by only choosing `.{m,c}ts{x}` first when not in node_modules

~I'm marking this as a draft because the logic currently only applies to the main/index file and it needs to also apply to the extension order. We can't merge this until that's fixed.~ Fixed

Fixes https://github.com/oven-sh/bun/issues/5426

There are tests

### DIFF
--- a/src/resolver/dir_info.zig
+++ b/src/resolver/dir_info.zig
@@ -67,7 +67,6 @@ pub const Flags = enum {
     has_node_modules,
 
     inside_node_modules,
-    has_set_inside_node_modules,
 
     pub const Set = std.enums.EnumSet(Flags);
 };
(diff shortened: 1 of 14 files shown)


src/resolver/dir_info.zig
```diff
@@ -55,12 +55,20 @@ pub inline fn isNodeModules(this: *const DirInfo) bool {
...
+}
+
 pub const Flags = enum {
     /// This directory is a node_modules directory
     is_node_modules,
     /// This directory has a node_modules subdirectory
     has_node_modules,
 
+    inside_node_modules,
+    has_set_inside_node_modules,
```
dylan-conway: remove this one?

dylan-conway (approved):

PR #7270, opened 2023-11-22
Use stderr for info about how to open inspector

mathiasrw:
### What does this PR do?

It will print instructions from `--inspect` to `stderr`(instead of `stdout`) to make it easier to debug / inspect scripts that are meant to output large amounts of data. 

----

At the moment `--inspect` info is part of stdout so you can: 

- use `bun --inspect main.js` to get the data but then clutter your CLI with the full output of the script OR 
- use`bun --inspect main.js > out.json` and deal with "fishing" the details out of the out.json file that looks like this:

image

This PR will make sure to print to stderr so you can `bun --inspect main.js > out.json`, get the info in the cli AND have your output sent to the file.

- [x] Code changes


src/js/internal/debugger.ts
```diff
@@ -20,12 +20,12 @@ export default function (
...
-    console.log(dim("--------------------- Bun Inspector ---------------------"), reset());
-    console.log(`Listening:\n  ${dim(href)}`);
+    console.error(dim("--------------------- Bun Inspector ---------------------"), reset());
+    console.error(`Listening:\n  ${dim(href)}`);
     if (protocol.includes("ws")) {
-      console.log(`Inspect in browser:\n  ${link(`https://debug.bun.sh/#${host}${pathname}`)}`);
+      console.error(`Inspect in browser:\n  ${link(`https://debug.bun.sh/#${host}${pathname}`)}`);
     }
-    console.log(dim("--------------------- Bun Inspector ---------------------"), reset());
+    console.error(dim("--------------------- Bun Inspector ---------------------"), reset());
```
Jarred-Sumner: console.error turns text red

this should be Bun.stderr.writer() and then writer.write()

mathiasrw: Ahh, good point! Let me fix that.

mathiasrw: Hmm. From your last comment I was of the impression that you asked me to do something like 

```
const writer = Bun.stderr.writer();
writer.write('foobar\n')
```

But nothing gets printed with this code. Not sure if its a bug or I cant see the logic. 

In any case, I ended up implementing the same result using

    Bun.write(Bun.stderr, "foobar\n")

Let me know if you feel like it needs more tweaking to be up to standard.

paperclover (approved):

PR #7271, opened 2023-11-22
UDP support

### LINKED ISSUE #1630: Implement dgram

### LINKED ISSUE #6306: Implement `$var` interpolation in `test.each`
Per https://jestjs.io/docs/api#1-testeachtablename-fn-timeout

---

```ts

test.each([
  { a: 1, b: 1, expected: 2 },
  { a: 1, b: 2, expected: 3 },
  { a: 2, b: 1, expected: 3 },
])(".add($a, $b)", ({ a, b, expected }) => {
  expect(a + b).toBe(expected);
});

// Expected:
// .add(1, 1)

// Actual:
// .add($a, $b)
```

_Originally reported on Discord: [`$variable does not work in test.each`](https://discord.com/channels/876711213126520882/1159178170898071722/1159178170898071722)_

Electroid:
## Issues
* Closes #1630
* Closes #6306
* Closes #6792
   * This builds upon the work from @tarruda, who will be credited as a co-author if this PR is merged.

## TODO
- [x] Diff needs to be cleaned up, since files in `packages/bun-usockets` were re-formatted.
- [x] Added support for `$variable` in `test.each`.
- [x] Fixed issue where `Bun.spawn` would remove empty arguments from `cmd` .
- [ ] Update documentation for `node:dgram`
- [ ] Update documentation for Bun's UDP API
- [ ] Decide on final Bun API (still work-in-progress)

I wrote automated tests.


Jarred-Sumner:
this doesn't implement `dgram` yet

packages/bun-usockets/src/internal/eventing/epoll_kqueue.h
```diff
@@ -64,8 +64,8 @@ struct us_loop_t {
 
 struct us_poll_t {
     alignas(LIBUS_EXT_ALIGNMENT) struct {
-        signed int fd : 28; // we could have this unsigned if we wanted to, -1 should never be used
-        unsigned int poll_type : 4;
+        signed int fd : 27; // we could have this unsigned if we wanted to, -1 should never be used
```
Jarred-Sumner, line 67: 33 million file descriptors 

seems ok

packages/bun-usockets/src/internal/internal.h
```diff
@@ -123,6 +124,21 @@ struct us_wrapped_socket_context_t {
...
+    alignas(LIBUS_EXT_ALIGNMENT) struct us_poll_t p;
+    struct us_udp_packet_buffer_t *receive_buf;
+    void (*on_data)(struct us_udp_socket_t *, struct us_udp_packet_buffer_t *, int);
+    void (*on_drain)(struct us_udp_socket_t *);
+    void *user;
+    struct us_loop_t *loop;
+    /* An UDP socket can only ever be bound to one single port regardless of how
+     * many interfaces it may listen to. Therefore we cache the port after creation
+     * and use it to build a proper and full sockaddr_in or sockaddr_in6 for every received packet */
+    int port;
```
Jarred-Sumner:
```suggestion
    uint16_t port;
```

packages/bun-usockets/src/udp.c
```diff
@@ -64,45 +66,67 @@ struct us_udp_packet_buffer_t *us_create_udp_packet_buffer() {
...
+}
 
-    udp->data_cb(s, udp->receive_buf, packets);
+void us_udp_socket_close(struct us_udp_socket_t *s) {
+  struct us_poll_t *p = (struct us_poll_t *) s;
+//   us_poll_stop(p, s->loop);
+//   bsd_close_socket(us_poll_fd(p));
+    s->closed = 1;
+    // change to writable poll to trigger cleanup on next loop cycle
+    us_poll_change(p, s->loop, LIBUS_SOCKET_WRITABLE);
```
Jarred-Sumner: what happens if the socket isn't writable until many past ticks? we would be leaving the packet buffer allocated (expensive) and not actually closing the socket, leaking the file descriptor.

IMO, this should:
1) shutdown() the socket so that no more data can be read/written
2) free the packet buffer
3) have a `us_udp_socket* next_to_close` pointer (or similar) and a `us_udp_socket* udp_socket_to_close_head` pointer on the `us_loop_t` struct. then in `us_loop_integrate`, it would check if there are any to close and close them there

Jarred-Sumner: then, `us_udp_socket_close` could check if it's already closed by seeing if the pointer is not null

packages/bun-usockets/src/udp.c
```diff
@@ -44,11 +42,15 @@ int us_udp_packet_buffer_payload_length(struct us_udp_packet_buffer_t *buf, int
 
 // what should we return? number of sent datagrams?
 int us_udp_socket_send(struct us_udp_socket_t *s, struct us_udp_packet_buffer_t *buf, int num) {
+    if (num == 0) return 0;
```
Jarred-Sumner, line 44: we should move this check into our code instead IMO

packages/bun-usockets/src/loop.c
```diff
@@ -397,6 +397,36 @@ void us_internal_dispatch_ready_poll(struct us_poll_t *p, int error, int events)
             }
             break;
         }
+        case POLL_TYPE_UDP: {
+            struct us_udp_socket_t *u = (struct us_udp_socket_t *) p;
+            // do cleanup
+            if (u->closed) {
```
Jarred-Sumner, line 411: this in theory could lose readable data (see above, there was a similar bug before)

packages/bun-usockets/src/loop.c
```diff
@@ -397,6 +397,36 @@ void us_internal_dispatch_ready_poll(struct us_poll_t *p, int error, int events)
...
+                break;
+            }
+
+            if (events & LIBUS_SOCKET_WRITABLE && !error) {
+                u->on_drain(u);
+                if (u->closed) {
+                    break;
+                }
+                // TODO handle socket close in callback
+                us_poll_change(&u->p, u->loop, us_poll_events(&u->p) & LIBUS_SOCKET_READABLE);
```
Jarred-Sumner, line 422: why do we call us_poll_change? readable/writable for sockets shouldn't be one shot so we shouldn't need to mark it as readable/writable each time

packages/bun-usockets/src/bsd.c
```diff
@@ -54,55 +54,46 @@ struct us_internal_udp_packet_buffer {
...
     /* Let's just use sendto here */
     /* Winsock does not have sendmsg, while macOS has, however, we simply use sendto since both macOS and Winsock has it.
-     * Besides, you should use Linux either way to get best performance with the sendmmsg */
+     * Besides, you should use Linux either way to get best performance with sendmmsg */
 
 
     // while we do not get error, send next
 
-    for (int i = 0; i < LIBUS_UDP_MAX_NUM; i++) {
+    for (int i = 0; i < vlen; i++) {
```
Jarred-Sumner:
```suggestion
    for (size_t i = 0; i < vlen; i++) {
```

packages/bun-usockets/src/bsd.c
```diff
@@ -54,55 +54,46 @@ struct us_internal_udp_packet_buffer {
...
             // if we fail then we need to buffer up, no that's not our problem
             // we do need to register poll out though and have a callback for it
             return i;
         }
-
-        //printf("sendto: %d\n", ret);
     }
 
-    return LIBUS_UDP_MAX_NUM; // one message
+    return vlen; // one message
```
Jarred-Sumner:
```suggestion
    return vlen; // number of messages sent
```

packages/bun-usockets/src/bsd.c
```diff
@@ -54,55 +54,46 @@ struct us_internal_udp_packet_buffer {
...
-    for (int i = 0; i < LIBUS_UDP_MAX_NUM; i++) {
+    for (int i = 0; i < vlen; i++) {
         // need to support ipv6 addresses also!
-        int ret = sendto(fd, packet_buffer->buf[i], packet_buffer->len[i], flags, (struct sockaddr *)&packet_buffer->addr[i], sizeof(struct sockaddr_in));
+        int ret = 0;
+        switch (packet_buffer->addr[i].ss_family) {
+            case AF_INET: {
+                struct sockaddr *addr = (struct sockaddr *) &packet_buffer->addr[i];
+                socklen_t len = sizeof(struct sockaddr_in);
+                ret = sendto(fd, packet_buffer->buf[i], packet_buffer->len[i], flags, addr, len);
```
Jarred-Sumner: this needs to handle `EINTR` for each of these function calls

packages/bun-usockets/src/bsd.c
```diff
@@ -54,55 +54,46 @@ struct us_internal_udp_packet_buffer {
...
     /* Winsock does not have sendmsg, while macOS has, however, we simply use sendto since both macOS and Winsock has it.
-     * Besides, you should use Linux either way to get best performance with the sendmmsg */
+     * Besides, you should use Linux either way to get best performance with sendmmsg */
 
 
     // while we do not get error, send next
 
-    for (int i = 0; i < LIBUS_UDP_MAX_NUM; i++) {
+    for (int i = 0; i < vlen; i++) {
         // need to support ipv6 addresses also!
```
Jarred-Sumner: we do handle ipv6 here now right?
```suggestion

```

packages/bun-usockets/src/bsd.c
```diff
@@ -54,55 +54,46 @@ struct us_internal_udp_packet_buffer {
...
-        int ret = sendto(fd, packet_buffer->buf[i], packet_buffer->len[i], flags, (struct sockaddr *)&packet_buffer->addr[i], sizeof(struct sockaddr_in));
+        int ret = 0;
```
Jarred-Sumner: We can copy the `addr` to a local variable here since we look it up multiple times
```suggestion
        int ret = 0;
        const struct sockaddr* addr = packet_buffer->addr[i];
```

packages/bun-usockets/src/bsd.c
```diff
@@ -54,55 +54,46 @@ struct us_internal_udp_packet_buffer {
...
-
-#elif defined(_WIN32)
-
+#if defined(_WIN32) || defined(__APPLE__)
     struct us_internal_udp_packet_buffer *packet_buffer = (struct us_internal_udp_packet_buffer *) msgvec;
 
     /* Let's just use sendto here */
     /* Winsock does not have sendmsg, while macOS has, however, we simply use sendto since both macOS and Winsock has it.
-     * Besides, you should use Linux either way to get best performance with the sendmmsg */
+     * Besides, you should use Linux either way to get best performance with sendmmsg */
```
Jarred-Sumner: can you add the macos `sendmsg_x`? it probably will be not much different than `sendmmsg` below

packages/bun-usockets/src/udp.c
```diff
@@ -44,11 +42,15 @@ int us_udp_packet_buffer_payload_length(struct us_udp_packet_buffer_t *buf, int
...
     int fd = us_poll_fd((struct us_poll_t *) s);
 
-    // we need to poll out if we failed
-
-    return bsd_sendmmsg(fd, buf, num, 0);
+    int sent = bsd_sendmmsg(fd, buf, num, 0);
+    if (0 <= sent && sent < num) {
+        // if we couldn't send all packets, register a writable event so we can call the drain callback
+        us_poll_change((struct us_poll_t *) s, s->loop, LIBUS_SOCKET_READABLE | LIBUS_SOCKET_WRITABLE);
```
Jarred-Sumner: like with TCP sockets, we should just keep it polled imo

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+const JSGlobalObject = JSC.JSGlobalObject;
+const JSValue = JSC.JSValue;
+
+const log = Output.scoped(.UdpSocket, false);
+
+const INET6_ADDRSTRLEN = if (bun.Environment.isWindows) 65 else 46;
+
+extern fn ntohs(nshort: u16) u16;
+extern fn htonl(hlong: u32) u32;
+extern fn htons(hshort: u16) u16;
```
Jarred-Sumner, line 20: i think we have wrappers for these elsewhere

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+        if (hostname == null or port == 0) {
+            continue;
+        }
+
+        const slice = buf.getPayload(i);
+
+        const loop = udpSocket.vm.eventLoop();
+        loop.enter();
+        defer loop.exit();
+        udpSocket.js_refcount += 1;
```
Jarred-Sumner: this feels strange imo

also if we are using has_pending_activity, then this needs to be done threadsafely which means using a `std.atomic.Value` and calling fetchadd/fetchsub

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+            globalThis.throwInvalidArguments("Expected 1 argument, got {}", .{arguments.len});
+            return .zero;
+        }
+
+        const arg0 = arguments.ptr[0];
+
+        var iter = arg0.arrayIterator(globalThis);
+        var i: u16 = 0;
+        var vals: [3]JSValue = .{ .zero, .zero, .zero };
+        while (iter.next()) |val| : (i += 1) {
```
Jarred-Sumner, line 409: shouldn't we check if the length of the array is divisible by 3 and then if it's not throw an error? 

then, we can simplify this code into either sendManyOneAddress or sendMany, and get the next 3 vals, fill the array, etc

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+    ) ?JSValue {
+        if (index >= 1024) {
+            globalThis.throwInvalidArguments("Too many packets to send, maximum is 1024", .{});
+            return .zero;
+        }
+        const arg0 = data_val;
+        const payload = init: {
+            if (arg0.asArrayBuffer(globalThis)) |arrayBuffer| {
+                break :init arrayBuffer.slice();
+            } else if (bun.String.tryFromJS(arg0, globalThis)) |value| {
```
Jarred-Sumner: This will call `.toString`, which means it will not throw if you pass it `{}` or `new Response` etc

We should probably instead check if `isString` and only then use tryFromJS

also, `tryFromJS` will only return null when it would otherwise have thrown or allocation failure (while joining rope strings). This would likely only happen when using something which throws in the `toString` function, like calling `Symbol("asd").toString()` - in which case, this would segfault because throwing an error inside an exception calls throw with a 0 JSValue

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+            connect_info.port,
+            length == 16,
+        );
+    }
+
+    pub fn getBinaryType(
+        this: *This,
+        globalThis: *JSGlobalObject,
+    ) callconv(.C) JSValue {
+        return switch (this.config.binary_type) {
```
Jarred-Sumner, line 674: maybe we should add a `label` method to `BinaryType` since i think we've implemented this before

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+        };
+    }
+
+    pub fn finalize(this: *This) callconv(.C) void {
+        log("Finalize {*}", .{this});
+        this.deinit();
+    }
+
+    pub fn deinit(this: *This) void {
+        if (!this.closed) {
```
Jarred-Sumner: this feels sketchy

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+            this.socket.close();
```
Jarred-Sumner: can `close` call `on_close` callbacks?

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,714 @@
...
+        }
+
+        this.poll_ref.unref(this.vm);
+
+        this.send_buf.destroy();
+
+        this.config.deinit();
+
+        default_allocator.destroy(this);
```
Jarred-Sumner: please make this use `bun.NewRefCounted` instead and then call `this.destroy` + `UDPSocket.new`. That will make it easier for us to track memory leaks with this later on

src/bun.js/test/jest.zig
```diff
@@ -1789,6 +1789,37 @@ fn formatLabel(globalThis: *JSC.JSGlobalObject, label: string, function_args: []
                     // ignore unrecognized fmt
                 },
             }
+        } else if (char == '$' and (idx + 1 < label.len) and !(args_idx >= function_args.len)) {
```
Jarred-Sumner: oh nice is this the `$0`, `$1` thing?

src/bun.js/test/jest.zig
```diff
@@ -1789,6 +1789,37 @@ fn formatLabel(globalThis: *JSC.JSGlobalObject, label: string, function_args: []
...
+                const remaining = label[idx + 1 ..];
+                const dot_idx = strings.indexOfChar(remaining, '.') orelse remaining.len;
+                const space_idx = strings.indexOfChar(remaining, ' ') orelse remaining.len;
+
+                if (dot_idx < space_idx) {
+                    const property = remaining[0..dot_idx];
+                    if (property.len == 0) break;
+
+                    idx += property.len;
+                    if (arg.get(globalThis, property)) |value| {
```
Jarred-Sumner: i hate to be a downer but this will need an exception check, something we don't do nearly enough but is very plausible in this case

`if (globalThis.hasException())` and then if yes return .zero probably

src/deps/uws.zig
```diff
@@ -1300,8 +1300,8 @@ pub const Poll = opaque {
         pub const write_flag = if (Environment.isLinux) std.os.linux.EPOLL.OUT else 2;
     };
 
-    pub fn deinit(self: *Poll) void {
-        us_poll_free(self);
+    pub fn deinit(self: *Poll, loop: *Loop) void {
+        us_poll_free(self, loop);
```
Jarred-Sumner, line 1306: do we need to pass the loop through?

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+  return port | 0;
+}
+hideFromStack(validatePort);
+
+function validateFunction(value, name) {
+  if (typeof value !== "function") throw new ERR_INVALID_ARG_TYPE(name, "Function", value);
+}
+hideFromStack(validateFunction);
+
+const BIND_STATE_UNBOUND = 0;
```
Jarred-Sumner: can you move these to the top of the file?

the bundler will be smart enough to inline all the primitive constants if they're at the top

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+
+  state.connectState = CONNECT_STATE_CONNECTING;
+  if (state.bindState === BIND_STATE_UNBOUND) this.bind({ port: 0, exclusive: true }, null);
+
+  if (state.bindState !== BIND_STATE_BOUND) {
+    enqueue(this, FunctionPrototypeBind(_connect, this, port, address, callback));
+    return;
+  }
+
+  ReflectApply(_connect, this, [port, address, callback]);
```
Jarred-Sumner: Can we use .$apply instead of Reflect.apply? Reflect.apply is probably slower

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+  try {
+    if (list.length === 1) {
+      if (port) {
+        socket.send(list[0], port, ip);
+      } else {
+        socket.send(list[0]);
+      }
+      sent = 1;
+    } else {
+      const packets = [];
```
Jarred-Sumner: can we set the length on this array? `$newArrayWithSize(size)` i think is the builtin name

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+          packets.push(buf);
+        }
+      }
+      sent += socket.sendMany(packets);
+    }
+  } catch (e) {
+    err = e;
+  }
+  // TODO check if this makes sense
+  if (callback) {
```
Jarred-Sumner, line 691: the callback should only be called after drain has drained them all? or is that not how this works?

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+  SymbolAsyncDispose,
+  SymbolDispose,
+  StringPrototypeTrim,
+  NumberIsNaN,
+} = require("internal/primordials");
+
+const { promisify } = require("node:util");
+const EventEmitter = require("node:events");
+
+const dc = require("node:diagnostics_channel");
```
Jarred-Sumner: can these be lazier? we only need to load node:diagnostics_channel when the socket is about to be created right? tons of packages will just require a ton of node core modules and then never use them or only use them infrequently-called codepaths, so we should avoid eager initialization as much as possible

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+  defaultTriggerAsyncIdScope(this[async_id_symbol], process.nextTick, socketCloseNT, this);
+
+  return this;
+};
+
+Socket.prototype[SymbolAsyncDispose] = async function () {
+  if (!this[kStateSymbol].handle) {
+    return;
+  }
+  console.log(this, this.close, promisify(this.close));
```
Jarred-Sumner:
```suggestion

```

src/js/node/dgram.ts
```diff
@@ -1,23 +1,1082 @@
...
+  return promisify(this.close).$call(this);
```
Jarred-Sumner: do we need promisify at all?

Jarred-Sumner (commented):
Some comments but a good start

packages/bun-usockets/src/bsd.c
gvilums: yes we do, the comment was stale

packages/bun-usockets/src/udp.c
gvilums: keeping it polled results in a `drain` notification being delivered on every tick of the event loop, which isn't really helpful

gvilums: also, we don't expect to poll for writable often, because most uses will only send one packet anyway, which will always fit into the write buffer

src/js/node/dgram.ts
gvilums: that's what node was doing, and from my reading of the code it makes sense - we want to return a promise from the async function that resolves when this.close is done.
I updated it to just return a new promise that resolves in the close callback

test/js/bun/udp/udp.test.ts
```diff
@@ -0,0 +1,488 @@
...
+          // handle unreliable transmission in UDP
+          function sendRec() {
+            if (!client.closed) {
+              validateSend(client.send(data, server.port, "127.0.0.1"));
+              setTimeout(sendRec, 10);
+            }
+          }
+          sendRec();
+        },
+        { timeout: 100000000 },
```
Electroid: this timeout is too long

test/js/bun/udp/udp.test.ts
```diff
@@ -0,0 +1,488 @@
...
+    expect(res).toBeGreaterThanOrEqual(0);
+    expect(res).toBeLessThanOrEqual(count);
+  };
+
+  for (const { binaryType, type } of dataTypes) {
+    for (const { label, data, bytes } of dataCases) {
+      test(
+        `send ${label} (${binaryType || "undefined"})`,
+        async done => {
+          const client = await udpSocket({});
```
Electroid: could we make it so you don't need to do `({})`

test/js/bun/udp/udp.test.ts
```diff
@@ -0,0 +1,488 @@
...
+        }
+        sendRec();
+      });
+    }
+  }
+});
+
+describe("createSocket()", () => {
+  test("connect", done => {
+    const PORT = 12345;
```
Electroid: this will cause flaky tests, could we use `0`?

test/js/bun/udp/udp.test.ts
```diff
@@ -0,0 +1,488 @@
+import { udpSocket } from "bun";
```
Electroid: This is going to be annoying, but can we break up the tests into 2 places:

`test/js/bun/udp` and `test/js/node/dgram`

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,751 @@
...
+
+        const hostname = brk: {
+            if (options.getTruthy(globalThis, "hostname")) |value| {
+                if (!value.isString()) {
+                    globalThis.throwInvalidArguments("Expected \"hostname\" to be a string", .{});
+                    return null;
+                }
+                const str = value.toBunString(globalThis);
+                defer str.deref();
+                break :brk str.toOwnedSliceZ(default_allocator) catch bun.outOfMemory();
```
Jarred-Sumner, line 138: These will leak on error

Instead, I suggest:
1) Move things which allocate to the end of this function
2) Assign them directly to `This` instead of a local variable 
3) `defer if (globalThis.hasPendingException()) config.deinit()`
4) `config.deinit` should call `unprotect` on the handlers to ensure we do not leak the callbacks

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,751 @@
...
+
+    pub usingnamespace JSC.Codegen.JSUDPSocket;
+
+    pub fn constructor(globalThis: *JSGlobalObject, _: *CallFrame) callconv(.C) ?*This {
+        globalThis.throw("Cannot construct UDPSocket", .{});
+        return null;
+    }
+
+    pub fn hasPendingActivity(this: *This) callconv(.C) bool {
+        return this.js_refcount > 0;
```
Jarred-Sumner:
```suggestion
        return this.js_refcount.fetch(.Monotonic) > 0;
```

test/js/bun/udp/udp.test.ts
gvilums: the problem is that this test is supposed to test for the fact that binding a port work. not sure how to get around this without getting rid of the test

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,751 @@
...
+            onData,
+            onDrain,
+            config.hostname,
+            config.port,
+            this,
+        )) |socket| {
+            this.socket = socket;
+        } else {
+            globalThis.throw("Failed to bind socket", .{});
+            return .zero;
```
Jarred-Sumner, line 326: Leaks:
- `this`
- `config`

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,751 @@
...
+        }
+
+        if (config.connect) |connect| {
+            const ret = this.socket.connect(connect.address, connect.port);
+            if (ret != 0) {
+                if (JSC.Maybe(void).errnoSys(ret, .connect)) |err| {
+                    globalThis.throwValue(err.toJS(globalThis));
```
Jarred-Sumner, line 333: Leaks:
- `this`
- `config`

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,751 @@
...
+        return .undefined;
+    }
+
+    pub fn close(
+        this: *This,
+        globalThis: *JSGlobalObject,
+        _: *CallFrame,
+    ) callconv(.C) JSValue {
+        if (this.closed) {
+            globalThis.throw("Socket is already closed", .{});
```
Jarred-Sumner: Should this throw or return undefined? I think it should not be an error to call .close on a closed socket

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -0,0 +1,751 @@
...
+        this.doClose();
+
+        return .undefined;
+    }
+
+    fn doClose(
+        this: *This,
+    ) void {
+        this.closed = true;
+        this.config.unprotect();
```
Jarred-Sumner: Can the socket be closed from usockets due to an error? Can the `onClose` callback still be called? 

I don't think it's safe to clean up the socket handlers until after usockets calls the on close callback

src/bun.js/api/bun/udp_socket.zig
Jarred-Sumner: i don't think it should be possible to enter this state

defaultkavy:
Thank you very much for completing this🙏

benjamingwynn:
For my own reference, advanced dgram methods are covered in issue #10381

RadhiRasho:
This might just be me, but it would seem like the bun-types changes never made it out/weren't bundled correctly. I've tried every way I can to get the latest and I'm on the latest version of Bun: 
image

I've also set the version of the @types/bun package to be the latest and cleared the pm cache a few times using `bun pm cache rm`.

This is all in a simple `bun init` project and is easily reproducible.

RadhiRasho:
> This might just be me, but it would seem like the bun-types changes never made it out/weren't bundled correctly. I've tried every way I can to get the latest and I'm on the latest version of Bun: image
> 
> I've also set the version of the @types/bun package to be the latest and cleared the pm cache a few times using `bun pm cache rm`.
> 
> This is all in a simple `bun init` project and is easily reproducible.

Ignore, seems to have been resolved

PR #7277, opened 2023-11-22
feat(test): `toHaveBeenCalledWith` and `toHaveBeenLastCalledWith`

james-elicx:
### What does this PR do?

- Implements jest matchers `toHaveBeenCalledWith` and `toHaveBeenLastCalledWith`.
- Updates that they're supported in the docs.

I would note that the formatted expected and received output when a test fails is not properly aligned with what jest does because, well, I'm not sure how you would have liked that to be approached. These matchers are somewhat unique in that they allow an unlimited number of arguments that they try to match against the arguments in a call.

I'm still relatively new to zig, so I wasn't sure how to approach the trying to print the array of call arguments like jest does on test failures in this codebase. I don't view that as a blocker though, it's just a nice to have for failures that could be done down the road perhaps, or taken up by one you guys if you want it in this PR(?)

If this PR goes well, I'll probably try some of the other unimplemented matchers :).

- [x] Code changes

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/bun.js/test/expect.zig
+++ b/src/bun.js/test/expect.zig
@@ -3493,8 +3493,137 @@ pub const Expect = struct {
         return .zero;
     }
 
-    pub const toHaveBeenCalledWith = notImplementedJSCFn;
-    pub const toHaveBeenLastCalledWith = notImplementedJSCFn;
+    pub fn toHaveBeenCalledWith(this: *Expect, globalObject: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSC.JSValue {
+        JSC.markBinding(@src());
+
+        const thisValue = callframe.this();
+        const arguments_ = callframe.argumentsPtr()[0..callframe.argumentsCount()];
+        const arguments: []const JSValue = arguments_.ptr[0..arguments_.len];
+        defer this.postMatch(globalObject);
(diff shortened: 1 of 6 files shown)


src/bun.js/test/jest.classes.ts
```diff
@@ -146,11 +146,9 @@ export default [
       },
       toHaveBeenCalledWith: {
         fn: "toHaveBeenCalledWith",
-        length: 1,
       },
       toHaveBeenLastCalledWith: {
         fn: "toHaveBeenLastCalledWith",
-        length: 1,
       },
```
james-elicx, line 152: Since they can have an unlimited number of arguments, it didn't seem like the length property would apply here. (i'm assuming it refers to fixed length)

src/bun.js/test/expect.zig
```diff
@@ -3493,8 +3493,137 @@ pub const Expect = struct {
...
+            globalObject.throw("Expected value must be a mock function: {}", .{value});
+            return .zero;
+        }
+
+        var pass = false;
+
+        if (calls.getLength(globalObject) > 0) {
+            var itr = calls.arrayIterator(globalObject);
+            while (itr.next()) |callItem| {
+                if (callItem == .zero or !callItem.jsType().isArray()) {
```
Jarred-Sumner, line 3519: This could be a holey array element I think

Jarred-Sumner:
Looks good. Thank you.

PR #7279, opened 2023-11-22
Support `glob.scan(string)`

Jarred-Sumner:
### What does this PR do?

This adds a single string argument as an overload of `glob.scan`

Tests


zackradisic (approved):
lgtm!

PR #7283, opened 2023-11-23
Delete `bun_dev_http_server.zig`

Jarred-Sumner:
### What does this PR do?

We removed `bun dev`, but we left a lot of the code in there. This deletes more of it.

It compiled


src/http.zig
```diff
@@ -3138,16 +3138,19 @@ fn handleResponseBodyChunkedEncodingFromMultiplePackets(
         buffer.list.items.ptr + (buffer.list.items.len -| incoming_data.len),
         &bytes_decoded,
     );
+    if (comptime Environment.allow_assert) {
```
Jarred-Sumner, line 3141: oops

paperclover (approved):
yay

Issue #7233, opened 2023-11-20
Migrating package-lock.json with overrides produces bun.lockb without overrrides


hborchardt:
### What version of Bun is running?

1.0.13+f5bf67bd1

### What platform is your computer?

Linux 5.15.0-47-generic x86_64 unknown

### What steps can reproduce the bug?

Hi, I want to use bun in our CI to install dependencies based on an existing package-lock.json. My package.json contains overrides, which leads to the following problem:

- Have a package.json with overrides:
```json
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@swagger-api/apidom-parser-adapter-asyncapi-json-2": "0.81.0"
  },
  "overrides": {
    "@swagger-api/apidom-parser-adapter-json": "npm:empty-npm-package@1.0.0"
  }
}
```
- Run the following:
  - `npm install` to generate `package-lock.json`
  - `bun pm migrate` to generate `bun.lockb` from `package-lock.json`
  - `bun install --frozen-lockfile` to reinstall dependencies
- Fails with `error: lockfile had changes, but lockfile is frozen`

### What is the expected behavior?

No error, as the package.json did not change.

### What do you see instead?

`error: lockfile had changes, but lockfile is frozen`

### Additional information

With some debugging I identified that the problem is that during migration, the Lockfile's overrides field is not set. So the following code path is triggered, leading to a difference in lockfiles:
https://github.com/oven-sh/bun/blob/778bad9dfd1b093cfd58a41a44ca9106828a9e0f/src/install/lockfile.zig#L2896-L2897
This is probably because the information is not explicitly contained in the package-lock.json file, and the migration only considers the package-lock.json, not the package.json.
So my ideas would be:
- Set `Lockfile.overrides` based on `package.json`
- Somehow deduce overrides from the `package-lock.json`. E.g. in my case, the `package-lock.json` contains the following:
```json
"node_modules/@swagger-api/apidom-parser-adapter-json": {
      "name": "empty-npm-package",
      "version": "1.0.0",
      "resolved": "https://registry.npmjs.org/empty-npm-package/-/empty-npm-package-1.0.0.tgz",
[...]

Issue #6608, opened 2023-10-19
Support nested "resolutions" / "overrides"


soneymathew:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

in package.json add nested resolutions

example from a package.json that failed bun install
```json
 "resolutions" : {
          "**/@myscope/pkg1/**/@anotherscope/pkg2": "187.8.3",
        "**/@myscope/pkg3/**/@emotion/react": "11.10.6",
        "**/@myscope/pkg4/**/@myscope/pkg5": "3.0.3",
        "**/jest-environment-enzyme/jest-environment-jsdom": "^28.0.0",
        "@myscope/pkg6/**/@anotherscope/pkg7": "^10.6.0",
        "@types/webpack-dev-server/**/webpack": "^4.44.1",
        "apollo/moment": "^2.29.4"
 }
```

### What is the expected behavior?

bun install should succeed without errors

### What do you see instead?

I see
```
warn: Bun currently does not support nested "resolutions"
```

rostero1:
This would be helpful in mitigating findings from security scans, where the immediate dependency may not have the fix, but I know I can safely bump the nested dependency.

ivanabrkic:
+1

gerardmarquinarubio:
+1

Currently `pdfjs-dist` can't be used with bun because it [doesn't support `node_module_register`](https://github.com/oven-sh/bun/issues/5835) and therefore `canvas`. This could be solved if we `bun add canvas@npm:@napi-rs/canvas` and:
```package.json
"overrides": {
  "pdfjs-dist": {
      "canvas": "$canvas"
    }
}
```

I'm sure many other packages would benefit from nested overrides, especially in cases where there is a dependency that's not yet supported by bun but that has other dependencies that mimic the api but don't depend on node-specific functionality.

For others with a similar problem for now I had to:
1. `bun add canvas@npm:@napi-rs/canvas`
2. `rsync -a --delete node_modules/canvas/ node_modules/pdfjs-dist/node_modules/canvas`

I added the second command to the "postinstall" script to make it something reproducible.

sainjay:
+1

warn: Bun currently does not support nested "overrides"

jebarpg:
Is there any work being done on this yet?

Jarred-Sumner:
[...]

PR #7285, opened 2023-11-23
Copy overrides from package.json during migration

hborchardt:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/7233
During migration of a `package-lock.json` to `bun.lockb`, this also parses the `package.json` file to get the necessary `overrides` section and include it in the bun Lockfile. While writing the test, I also found a missing counting of the package name string length.

- [x] Code changes

I wrote automated tests

- [ ] ~JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed~

This is my first endeavour into zig, so I hope I got it right :) if not, feel free to criticise.

### DIFF
--- a/src/install/migration.zig
+++ b/src/install/migration.zig
@@ -952,6 +952,45 @@ pub fn migrateNPMLockfile(this: *Lockfile, allocator: Allocator, log: *logger.Lo
     this.buffers.resolutions.items.len = (@intFromPtr(resolutions_buf.ptr) - @intFromPtr(this.buffers.resolutions.items.ptr)) / @sizeOf(Install.PackageID);
     this.buffers.dependencies.items.len = this.buffers.resolutions.items.len;
 
+    // consult the corresponding package.json file for overrides, if possible
+    const overrides_copied = copied: {
+        const package_json_dir_path = path[0..(path.len - "package-lock.json".len)];
+        const package_json_path_buf = try this.allocator.alloc(u8, path.len - "package-lock.json".len + "package.json".len);
+        defer this.allocator.free(package_json_path_buf);
+        @memcpy(package_json_path_buf[0..package_json_dir_path.len], package_json_dir_path);
+        package_json_path_buf[package_json_dir_path.len..package_json_path_buf.len][0.."package.json".len].* = "package.json".*;
+
+        var json_file = bun.openFile(package_json_path_buf, .{ .mode = .read_only }) catch break :copied false;
+        defer json_file.close();
+        const json_stat_size = try json_file.getEndPos();
+        const json_buf = try this.allocator.alloc(u8, json_stat_size + 64);
+        defer this.allocator.free(json_buf);
+        const json_len = try json_file.preadAll(json_buf, 0);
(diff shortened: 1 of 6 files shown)


src/install/migration.zig
```diff
@@ -937,6 +942,45 @@ pub fn migrateNPMLockfile(this: *Lockfile, allocator: Allocator, log: *logger.Lo
...
 
+    // consult the corresponding package.json file for overrides, if possible
+    var overrides_copied = copied: {
+        var package_json_dir_path = path[0..(path.len - "package-lock.json".len)];
+        const package_json_path_buf = try this.allocator.alloc(u8, path.len - "package-lock.json".len + "package.json".len);
+        defer this.allocator.free(package_json_path_buf);
+        @memcpy(package_json_path_buf[0..package_json_dir_path.len], package_json_dir_path);
+        package_json_path_buf[package_json_dir_path.len..package_json_path_buf.len][0.."package.json".len].* = "package.json".*;
+
+        var json_file = bun.openFile(package_json_path_buf, .{ .mode = .read_only }) catch break :copied false;
```
Jarred-Sumner, line 968: Thank you for this, but we shouldn't read package-lock.json twice. `json` is already accessible in this function, which contains the package-lock.json file.

src/install/migration.zig
```diff
@@ -270,6 +270,11 @@ pub fn migrateNPMLockfile(this: *Lockfile, allocator: Allocator, log: *logger.Lo
             }
         }
 
+        if (pkg.get("name")) |pkg_name| {
```
Jarred-Sumner, line 276: The package's name is already counted:

https://github.com/oven-sh/bun/blob/3c707357426d218400ca6cbaff665c061627f83d/src/install/migration.zig#L143

Jarred-Sumner (changes requested):
This will need some changes before it can be merged

src/install/migration.zig
hborchardt: This is loading the _package.json_ file, to temporarily generate a bun Lockfile from it, because the package-lock.json does not contain enough information about overrides.

src/install/migration.zig
hborchardt: That line counts the root package's name, but this change is about an installed (dependency) package's overridden name, see for example https://github.com/oven-sh/bun/pull/7285/files#diff-ad9059f9fe54fff89c5cfc27830b3ee35e7a8dca6881878059c027e9fe2c32c5R32 which is fetched here:
https://github.com/oven-sh/bun/blob/3c707357426d218400ca6cbaff665c061627f83d/src/install/migration.zig#L417-L418
and appended here:
https://github.com/oven-sh/bun/blob/3c707357426d218400ca6cbaff665c061627f83d/src/install/migration.zig#L434

hborchardt:
@Jarred-Sumner I believe that no changes are necessary, please reconsider. Thanks!

birkskyum:
Related
- https://github.com/oven-sh/bun/issues/6608

PR #7287, opened 2023-11-23
Set NODE_ENV to production in release step

annervisser:
### What does this PR do?

Not only the build, but also the final release should run with NODE_ENV set to production

- [ ] Code changes

### DIFF
--- a/docs/guides/ecosystem/docker.md
+++ b/docs/guides/ecosystem/docker.md
@@ -43,6 +43,7 @@ RUN bun run build
 
 # copy production dependencies and source code into final image
 FROM base AS release
+ENV NODE_ENV=production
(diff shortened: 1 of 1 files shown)


paperclover:
do environment variables inherit? maybe this node env should be set in the base image or even within oven-sh/bun if not already.

annervisser:
> do environment variables inherit? maybe this node env should be set in the base image or even within oven-sh/bun if not already.

They do inherit, so it could be set in the base layer instead.

It's not currently set in the `oven/bun` image

Setting it in oven/bun might be undesirable: 
 1. Bun running in the docker container would always grab the .env.production values (according to the [docs](https://bun.sh/docs/runtime/env#setting-environment-variables))
 2. This will mean a lot of libraries will not give debug logs etc. by default.

PR #7292, opened 2023-11-23
fix(node:fs): fix `fs.promises.opendir` not having a `path` property

### LINKED ISSUE #4995: `path` property is missing in returned object for `fs.promises.opendir`

Darwin 23.0.0 arm64 arm

```
import { opendir } from 'fs/promises'

const dir = await opendir('.');

console.log(dir.path);
```

It should log the loaded path for the directory (nodejs seems to just point to whatever is passed into opendir)

### What do you see instead?

`undefined`

### Additional information

_No response_

samfundev:
### What does this PR do?

Adds the path property to the Dir class returned by fs.promises.opendir. Closes #4995

- [x] Code changes

### DIFF
--- a/src/js/node/fs.promises.ts
+++ b/src/js/node/fs.promises.ts
@@ -103,8 +103,13 @@ function cp(src, dest, options) {
 // This is currently stubbed for Next.js support.
 class Dir {
   #entries: Dirent[];
-  constructor(e: Dirent[]) {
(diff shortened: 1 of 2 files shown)


paperclover (approved):
thanks, we really need to put this into native code sometime. but this patch fixes the missing functionality.

PR #7294, opened 2023-11-24
Feat: add jsx spread children support

rhyzx:
#### Example
```js
const a = <h1>{...[123]}</h1>
```
##### Before: 
```sh
error: Unexpected ...
```
##### After:
```js
var a = jsx_dev_runtime.jsxDEV("h1", {
  children: [...[123]]
}, undefined, true, undefined, this);
```

This feature is already supported by most transpilers, eg:
[Typescript](https://www.typescriptlang.org/play?target=99&jsx=5#code/MYewdgzgLgBAhjAvDAPACwIwD4DeA6AgbQwCYBmAXQF8UB6TLAKCA)
[Esbuild](https://esbuild.github.io/try/#dAAwLjE5LjcAewogIGxvYWRlcjogJ2pzeCcsCiAganN4OiAnYXV0b21hdGljJywKfQA8ZGl2PnsuLi5bXX08L2Rpdj4)
[SWC](https://play.swc.rs/?version=1.3.100-nightly-20231124.1&code=H4sIAAAAAAAAA0vOzysuUUhUsFWwSckss6vW09OLjq210QdxAHOBudocAAAA&config=H4sIAAAAAAAAA1WOMQrDMAxF95zCaO5QPPY2wijFIbaDJENL8N0rO2lJN%2F33kPT3yTlYJMDD7TZa2JCF%2BJeNyDsrvowAhYQSOG4Kt69dpCvlSoO0Q8BaipCJGVehkyljlrlwul5nwqAX0FHNGlPfBqxaEmoMcOr290WRn6Sjmvi796NWm9oHTYlQ69cAAAA%3D)

- [x] Code changes

I am willing to add test code for this feature, but I need some help.
As I understand it, the test code should be placed in `test/snippets`, `test/snapshots`.
However, the `scripts/browser.js` that tests them does not seem to be able to execute properly.

### DIFF
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -14392,14 +14392,18 @@ fn NewParser_(
                         // Use Next() instead of NextJSXElementChild() here since the next token is an expression
                         try p.lexer.next();
 
-                        // The "..." here is ignored (it's used to signal an array type in TypeScript)
-                        if (p.lexer.token == .t_dot_dot_dot and is_typescript_enabled) {
+                        const is_spread = p.lexer.token == .t_dot_dot_dot;
+                        if (is_spread) {
                             try p.lexer.next();
                         }
 
                         // The expression is optional, and may be absent
                         if (p.lexer.token != .t_close_brace) {
-                            try children.append(try p.parseExpr(.lowest));
+                            var item = try p.parseExpr(.lowest);
+                            if (is_spread) {
+                                item = p.newExpr(E.Spread{ .value = item }, loc);
+                            }
+                            try children.append(item);
                         }
 
                         // Use ExpectJSXElementChild() so we parse child strings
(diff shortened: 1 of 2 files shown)


paperclover (changes requested):
can you add a test to transpiler.test.ts?

paperclover:
also, i wonder if we can detect an array literal spread, to emit `[1, 2, 3]` instead of `[...[1, 2, 3]]`

rhyzx:
> can you add a test to transpiler.test.ts?

DONE

rhyzx:
> also, i wonder if we can detect an array literal spread, to emit `[1, 2, 3]` instead of `[...[1, 2, 3]]`

Yes, the optimizer should do this, and actually Bun has already implemented this.

```js
// in.js
console.log([...[1, 2, 3]]);
console.log(
  jsxDEV(
    "div",
    {
      children: [...[1, 2, 3]],
    },
    undefined,
    true,
    undefined,
    this,
  ),
);
console.log(<div>{...[1, 2, 3]}</div>);

// bun-debug build test.js --minify
// ...
console.log([1,2,3]);
console.log(jsxDEV("div",{children:[1,2,3]},void 0,!0,void 0,OJ));
console.log(VJ.jsxDEV("div",{children:[...[1,2,3]]},void 0,!0,void 0,this))})
// ...
```

However, strangely, the code transformed through JSX does not execute this optimization.🤔

If I have time, I will attempt to fix this issue next.

src/js_parser.zig
```diff
@@ -15087,28 +15091,23 @@ fn NewParser_(
...
+                                    const is_static_jsx = e_.children.len > 1 or (e_.children.len == 1 and e_.children.ptr[0].data == .e_spread);
+
+                                    if (is_static_jsx) {
+                                        props.append(allocator, G.Property{
+                                            .key = children_key,
+                                            .value = p.newExpr(E.Array{
+                                                .items = e_.children,
+                                                .is_single_line = e_.children.len < 2,
+                                            }, e_.close_tag_loc),
+                                        }) catch unreachable;
```
paperclover, line 15105: i know the old code had this but can you change `catch unreachable` to `catch bun.outOfMemory()`. the old code will behave weirdly in out of memory, where calling this new function will properly crash.

paperclover (changes requested):
looks good, i would like to get the `unreachable` removed.

paperclover:
these test cases would be nice to have but dont pass right now. i dont really know how hard adding this optimization in would be.
```
expect(bun.transformSync("export var foo = <div>{...[1,2,3]}</div>")).toBe(
  `export var foo = jsxDEV("div", {
  children: [1, 2, 3]
}, undefined, true, undefined, this);
`,
);
expect(bun.transformSync("export var foo = <div>a{...[1,2,3]}b</div>")).toBe(
  `export var foo = jsxDEV("div", {
  children: [
    "a",
    1,
    2,
    3,
    "b"
  ]
}, undefined, true, undefined, this);
`,
);
```

src/js_parser.zig
rhyzx: DONE

rhyzx:
> these test cases would be nice to have but dont pass right now. i dont really know how hard adding this optimization in would be.
> 
> ```
> expect(bun.transformSync("export var foo = <div>{...[1,2,3]}</div>")).toBe(
>   `export var foo = jsxDEV("div", {
>   children: [1, 2, 3]
> }, undefined, true, undefined, this);
> `,
> );
> expect(bun.transformSync("export var foo = <div>a{...[1,2,3]}b</div>")).toBe(
>   `export var foo = jsxDEV("div", {
>   children: [
>     "a",
>     1,
>     2,
>     3,
>     "b"
>   ]
> }, undefined, true, undefined, this);
> `,
> );
> ```

I've observed that esbuild also doesn't optimize spread children in jsx, [see]( https://esbuild.github.io/try/#dAAwLjE5LjcAewogIGxvYWRlcjogJ2pzeCcsCiAganN4OiAnYXV0b21hdGljJywKICBtaW5pZnk6IHRydWUKfQBjb25zb2xlLmxvZyg8ZGl2PnsuLi5bMSwyXX08L2Rpdj4pCmNvbnNvbGUubG9nKFsuLi5bMSwyXV0pCmNvbnNvbGUubG9nKFsxLCAuLi5bMiwgM10sIDRdKQ)

Implementing this optimization may not be as easy as it seems. I think we should open a new issue to discuss it.

paperclover (approved):

paperclover:
alright. looks good otherwise. will merge after tests run.

paperclover:
thank you

PR #7296, opened 2023-11-24
Implement `recursive` option in `fs.readdir`

Jarred-Sumner:
### What does this PR do?

This implements support for `readdir(path, {recursive: true})`

```js
import {readdir} from 'fs/promises';
const results = await readdir(__dirname, { recursive: true });
console.log(results) // ["a.js", "b/c.js" ...
```

There are tests.

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -593,8 +593,17 @@ pub const AsyncReaddirRecursiveTask = struct {
         this.result_list.deinit();
         var batch = this.result_list_queue.popBatch();
         var iter = batch.iterator();
+        var to_destroy: ?*ResultListEntry = null;
+
         while (iter.next()) |val| {
             val.value.deinit();
+            if (to_destroy) |dest| {
+                bun.default_allocator.destroy(dest);
+            }
+            to_destroy = val;
+        }
+        if (to_destroy) |dest| {
+            bun.default_allocator.destroy(dest);
         }
         this.result_list_count.store(0, .Monotonic);
     }
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -346,7 +346,7 @@ pub const AsyncReaddirRecursiveTask = struct {
     /// All the subtasks will use this fd to open files
     root_fd: FileDescriptor = bun.invalid_fd,
 
-    ///
+    /// This isued when joining the file paths for error messages
     root_path: PathString = PathString.empty,
 
     pending_err: ?Syscall.Error = null,
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -102,6 +102,8 @@ pub const Async = struct {
 
     pub const cp = AsyncCpTask;
 
+    pub const readdir_recursive = AsyncReaddirRecursiveTask;
+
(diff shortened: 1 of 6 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -317,6 +319,349 @@ pub const AsyncCpTask = struct {
...
 
+pub const AsyncReaddirRecursiveTask = struct {
+    promise: JSC.JSPromise.Strong,
+    args: Arguments.Readdir,
+    globalObject: *JSC.JSGlobalObject,
+    task: JSC.WorkPoolTask = .{ .callback = &workPoolCallback },
+    ref: bun.Async.KeepAlive = .{},
+    tracker: JSC.AsyncTaskTracker,
+
+    // It's not 100% clear this one is necessary
```
ishank-s, line 330: hi @Jarred-Sumner  , wanted to know your opinion on comments like these?
are they cool for all code bases or you add these cause you're the owner. 
Thanks for your great work!

src/bun.zig
```diff
@@ -2052,3 +2052,7 @@ pub fn exitThread() noreturn {
         @panic("Unsupported platform");
     }
 }
+
+pub fn outOfMemory() noreturn {
```
paperclover, line 2056: thank you for using this pattern.

src/bun.js/node/node_fs.zig
```diff
@@ -317,6 +319,349 @@ pub const AsyncCpTask = struct {
...
+        if (result.items.len > 0) {
+            const Field = comptime switch (ResultType) {
+                bun.String => .files,
+                Dirent => .with_file_types,
+                Buffer => .buffers,
+                else => unreachable,
+            };
+            var list = bun.default_allocator.create(ResultListEntry) catch bun.outOfMemory();
+            errdefer {
+                bun.default_allocator.destroy(list);
```
paperclover, line 525: nitpick: is this code ever run?

> Zig has `errdefer`, which evaluates the deferred expression on block exit path if and only if the function returned with an error from the block.

(there is no `return error` case)

src/bun.js/node/node_fs.zig
```diff
@@ -4208,19 +4542,335 @@ pub const NodeFS = struct {
...
+                    return .{
+                        .err = err,
+                    };
+                },
+                .result => .{ .result = @unionInit(Return.Readdir, file_type, entries.items) },
+            };
+        }
+
+        if (comptime recursive) {
+            @panic("This code path should never be reached. It should only go through readdirWithEntriesRecursiveAsync.");
```
paperclover, line 4856: can this be `@compileError`?

paperclover (approved):
no meaningful comments, this looks good

src/bun.js/node/node_fs.zig
Jarred-Sumner: I tried. But no, unfortunately

NReilingh:
Doesn't this forget to update the type definitions? My LSP is complaining when I pass `{ recursive: true }` on 1.0.15. I took a look at https://github.com/oven-sh/bun/blob/main/packages/bun-types/fs.d.ts and expect to see the `recursive` option alongside `withFileTypes`; only `withFileTypes` is present on the `readdir`/`readdirSync` calls.

PR #7305, opened 2023-11-24
dx: automatically install and update zig

paperclover:
### What does this PR do?

this makes it much harder for someone to use the wrong version of zig when compiling. by default `bun setup` will auto-install the version specified in `build.zig`. this also makes cmake configure_depends on build.zig, which will make it so auto-updates happen here.

unless you manually specify -DZIG_COMPILER, you cant really mess this up

also if you run `zig build`, you get

```
Build darwin-aarch64 v11.0.0 - v13.6.1 (apple_m1)
Zig v0.12.0-dev.1685+994e19164

error: To build Bun from source, please use `bun run setup` instead of `zig build`"

If you want to build the zig code only, run:
  'zig build obj -Dgenerated-code=./build/codegen [...opts]'

For more info, see https://bun.sh/docs/project/contributing
```

also for new zig upgrades: `bash scripts/update-zig.sh master`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,7 +320,7 @@ function(validate_zig validator_result_var item)
     endif()
 endfunction()
 
-if (WIN32 OR ZIG_COMPILER OR CI)
+if (WIN32 OR ZIG_COMPILER)
     if(ZIG_COMPILER STREQUAL "system")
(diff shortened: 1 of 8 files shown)


paperclover:
oh lol all the builds fail

PR #7309, opened 2023-11-25
Add `import.meta.env` to `bun-types`

RiskyMH:
### What does this PR do?

`import.meta.env` was missing from bun types, so I added.

### DIFF
--- a/packages/bun-types/globals.d.ts
+++ b/packages/bun-types/globals.d.ts
@@ -306,6 +306,14 @@ interface ImportMeta {
    */
   readonly file: string;
   /**
+   * The environment variables of the process
+   * 
+   * ```
+   * import.meta.env === process.env 
+   * ```
+   */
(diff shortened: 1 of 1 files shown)


packages/bun-types/globals.d.ts
```diff
@@ -305,6 +305,14 @@ interface ImportMeta {
    * Filename of the source file
    */
   readonly file: string;
+  /**
+   * The environment variables of the process
+   * 
+   * ```
```
Didas-git: The extension is missing, it should be ```ts for better readability

Jarred-Sumner:
Thank you

Jarred-Sumner:
We should make sure these don't somehow conflict with Vite types

jagu-sayan:
@cirospaciari Look like it's not supported by the new package `@types/bun`

PR #7310, opened 2023-11-25
feat(parseArgs): Native implementation of parseArgs from node:util

otgerrogla:
### What does this PR do?

Native zig implementation of the Node `parseArgs` function from the `node:util` package.
Includes the corresponding tests from Node v21, ported to bun:test.

I also added this function to the bun global object, exposed at `Bun.parseArgs`. Will remove it if thats's an issue, but I'm in the opinion that parseArgs is very useful to write quick scripts without needing dependencies, and I'd like to see it get first-class support on Bun, as in Node it was kind of hidden away and not that well-known.

- [x] Code changes

I wrote automated tests (ported)
Both in js for the whole function, and some ported to zig for unit testing.

### DIFF
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -260,12 +260,12 @@ pub const ZigString = extern struct {
         return this.slice()[0] == char;
     }
 
-    pub fn substringWithLen(this: ZigString, offset: usize, len: usize) ZigString {
+    pub fn substringWithLen(this: ZigString, start_index: usize, end_index: usize) ZigString {
         if (this.is16Bit()) {
-            return ZigString.from16Slice(this.utf16SliceAligned()[@min(this.len, offset)..len]);
+            return ZigString.from16Slice(this.utf16SliceAligned()[start_index..end_index]);
         }
 
-        var out = ZigString.init(this.slice()[@min(this.len, offset)..len]);
+        var out = ZigString.init(this.slice()[start_index..end_index]);
         if (this.isUTF8()) {
             out.markUTF8();
         }
@@ -277,8 +277,8 @@ pub const ZigString = extern struct {
         return out;
     }
 
-    pub fn substring(this: ZigString, offset: usize) ZigString {
-        return this.substringWithLen(offset, this.len);
+    pub fn substring(this: ZigString, start_index: usize) ZigString {
+        return this.substringWithLen(@min(this.len, start_index), this.len);
     }
 
     pub fn maxUTF8ByteLength(this: ZigString) usize {
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -3557,10 +3557,12 @@ pub const JSValue = enum(JSValueReprInt) {
 
     extern fn JSC__JSValue__constructEmptyObject(globalObject: *JSGlobalObject, prototype: [*c]JSC.JSObject, len: usize) JSValue;
 
+    /// Creates a new empty object with the specified prototype, or no prototype if null
     pub fn constructEmptyObject(global: *JSGlobalObject, prototype: ?*JSC.JSObject, len: usize) JSValue {
         return JSC__JSValue__constructEmptyObject(global, prototype, len);
     }
 
+    /// Creates a new empty object, with Object as its prototype
     pub fn createEmptyObject(global: *JSGlobalObject, len: usize) JSValue {
         return cppFn("createEmptyObject", .{ global, len });
     }
--- a/src/bun.js/node/util/parse_args.zig
+++ b/src/bun.js/node/util/parse_args.zig
@@ -285,8 +285,7 @@ fn parseOptionDefinitions(globalThis: *JSGlobalObject, options_obj: JSValue, opt
 
         if (obj.getOwn(globalThis, "multiple")) |multiple_value| {
             if (!multiple_value.isUndefined()) {
-                try validateBoolean(globalThis, multiple_value, "options.{s}.multiple", .{option.long_name});
-                option.multiple = multiple_value.toBoolean();
+                option.multiple = try validateBoolean(globalThis, multiple_value, "options.{s}.multiple", .{option.long_name});
             }
         }
 
@@ -295,16 +294,16 @@ fn parseOptionDefinitions(globalThis: *JSGlobalObject, options_obj: JSValue, opt
                 switch (option.type) {
                     .string => {
                         if (option.multiple) {
-                            try validateStringArray(globalThis, default_value, "options.{s}.default", .{option.long_name});
+                            _ = try validateStringArray(globalThis, default_value, "options.{s}.default", .{option.long_name});
                         } else {
                             try validateString(globalThis, default_value, "options.{s}.default", .{option.long_name});
                         }
                     },
                     .boolean => {
                         if (option.multiple) {
-                            try validateBooleanArray(globalThis, default_value, "options.{s}.default", .{option.long_name});
+                            _ = try validateBooleanArray(globalThis, default_value, "options.{s}.default", .{option.long_name});
                         } else {
-                            try validateBoolean(globalThis, default_value, "options.{s}.default", .{option.long_name});
+                            _ = try validateBoolean(globalThis, default_value, "options.{s}.default", .{option.long_name});
                         }
                     },
                 }
@@ -587,17 +586,14 @@ pub fn parseArgs(globalThis: *JSGlobalObject, config_obj: JSValue) !JSValue {
     const config_return_tokens: JSValue = (if (config) |c| c.getOwn(globalThis, "tokens") else null) orelse JSValue.jsBoolean(false);
     const config_options_obj: ?JSValue = if (config) |c| c.getOwn(globalThis, "options") else null;
 
-    try validateBoolean(globalThis, config_strict, "strict", .{});
-    const strict = config_strict.toBoolean();
+    const strict = try validateBoolean(globalThis, config_strict, "strict", .{});
 
     var allow_positionals = !strict;
     if (config_allow_positionals) |config_allow_positionals_value| {
-        try validateBoolean(globalThis, config_allow_positionals_value, "allowPositionals", .{});
-        allow_positionals = config_allow_positionals_value.toBoolean();
+        allow_positionals = try validateBoolean(globalThis, config_allow_positionals_value, "allowPositionals", .{});
     }
 
-    try validateBoolean(globalThis, config_return_tokens, "tokens", .{});
-    const return_tokens = config_return_tokens.toBoolean();
+    const return_tokens = try validateBoolean(globalThis, config_return_tokens, "tokens", .{});
 
     // Phase 0.C: Parse the options definitions
 
@@ -625,6 +621,7 @@ pub fn parseArgs(globalThis: *JSGlobalObject, config_obj: JSValue) !JSValue {
     //
     log("Phase 2: parse options from tokens (tokens.len={d})", .{tokens.items.len});
 
+    // note that "values" needs to have a null prototype instead of Object, to avoid issues such as "values.toString"` being defined
     var result_values = JSValue.constructEmptyObject(globalThis, null, 0);
     var result_positionals = JSC.JSValue.createEmptyArray(globalThis, 0);
     var result_positionals_len: u32 = 0;
--- a/src/bun.js/node/util/parse_args.zig
+++ b/src/bun.js/node/util/parse_args.zig
@@ -1,0 +1,687 @@
+const std = @import("std");
+const bun = @import("root").bun;
+const string = bun.string;
+const strings = bun.strings;
+const String = bun.String;
+const JSC = bun.JSC;
+const JSValue = JSC.JSValue;
+const JSGlobalObject = JSC.JSGlobalObject;
+const ZigString = JSC.ZigString;
+
+const validators = @import("./validators.zig");
+const validateArray = validators.validateArray;
+const validateBoolean = validators.validateBoolean;
+const validateBooleanArray = validators.validateBooleanArray;
+const validateObject = validators.validateObject;
+const validateString = validators.validateString;
+const validateStringArray = validators.validateStringArray;
+const validateStringEnum = validators.validateStringEnum;
+
+const utils = @import("./parse_args_utils.zig");
+const OptionValueType = utils.OptionValueType;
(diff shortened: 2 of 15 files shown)


test/js/node/util/parse_args/default-args.test.mjs
```diff
@@ -0,0 +1,80 @@
...
+  ])('running "bun %s"', async (argline, valuesKeys, positionals, execArgv) => {
+    const result = await spawnBun(...argline.split(/\s+/));
+    let output;
+    expect(() => (output = JSON.parse(result.stdout))).not.toThrow();
+    expect(Object.keys(output?.values ?? {}).sort()).toEqual(valuesKeys.sort());
+    expect(output?.positionals).toEqual(positionals);
+    if (execArgv) {
+      expect(output?.execArgv).toEqual(execArgv);
+    }
+  });
```
gamedevsam, line 79: This hurts my brain, impressive and weird!

src/bun.js/node/util/validators.zig
```diff
@@ -0,0 +1,233 @@
...
+pub fn validateStringEnum(comptime T: type, globalThis: *JSGlobalObject, value: JSValue, comptime name_fmt: string, name_args: anytype) !T {
+    const str = value.toBunString(globalThis);
+    inline for (@typeInfo(T).Enum.fields) |enum_field| {
+        if (str.eqlComptime(enum_field.name))
+            return @field(T, enum_field.name);
+    }
+
+    const values_info = comptime blk: {
+        var out: []const u8 = "";
+        for (@typeInfo(T).Enum.fields, 0..) |enum_field, i| {
```
gamedevsam, line 242: What the heck is going on here? Any docs I can read up to understand this?

src/bun.js/node/util/parse_args.zig
```diff
@@ -0,0 +1,687 @@
...
+            }
+        }
+    }
+
+    //
+    // Phase 4: build the resulting object: `{ values: [...], positionals: [...], tokens?: [...] }`
+    //
+    log("Phase 4: Build result object", .{});
+
+    var result = JSValue.createEmptyObject(globalThis, if (return_tokens) 3 else 2);
```
gamedevsam, line 676: What does this do? ` if (return_tokens) 3 else 2`

gamedevsam: Looks like the number of properties in the object?

gamedevsam (commented):
I can't really follow this code closely, as I'm a Zig noob, but the code is well commented, and is tested, I am excited to see this land in the mainline and I can get back to trying bun in my repos. Phenomenal work.

src/bun.js/bindings/bindings.cpp
```diff
@@ -3172,6 +3189,33 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsImpl(JSC__JSValue JSValue0,
...
+{
+    VM& vm = globalObject->vm();
+    JSValue value = JSC::JSValue::decode(JSValue0);
+    auto identifier = JSC::Identifier::fromString(vm, propertyName->toWTFString(BunString::ZeroCopy));
+    auto property = JSC::PropertyName(identifier);
+    PropertySlot slot(value, PropertySlot::InternalMethodType::GetOwnProperty);
+    if (value.getOwnPropertySlot(globalObject, property, slot)) {
+        return JSValue::encode(slot.getValue(globalObject, property));
+    }
+    return JSValue::encode({});
```
gamedevsam: I don't know how this is valid C++, what does `{}` return, isn't that a block?

src/bun.js/node/util/parse_args.zig
otgerrogla: Yeah, afaik it preallocates the number or properties in the object, it's an internal JavaScriptCore engine optimization

src/bun.js/bindings/bindings.cpp
otgerrogla: It's an empty initialization list, that invokes the default initializer for a struct/class fields (https://en.cppreference.com/w/cpp/language/aggregate_initialization)

In this case it returns an "empty" (invalid) JSValue, which the zig code checks for

src/bun.js/node/util/validators.zig
otgerrogla: It's a compile-time snippet that concats the enum fields names into a string, to show in the error message,  using compile-time reflection. Made it based on the official zig docs such as [this](https://ziglang.org/documentation/master/#enum:~:text=//%20%40typeInfo%20tells%20us%20the%20field%20count%20and%20the%20fields%20names%3A)

src/bun.js/bindings/bindings.cpp
gamedevsam: Looks like C++ has changed a bit since I last used it extensively, my JS focus has made me rusty in native code. Thanks for the explanations, much appreciated!

src/bun.js/bindings/bindings.zig
```diff
@@ -3542,6 +3542,18 @@ pub const JSValue = enum(JSValueReprInt) {
         return this.jsType();
     }
 
+    extern fn JSC__jsTypeStringForValue(globalObject: *JSGlobalObject, value: JSValue) *JSC.JSString;
+
+    pub fn jsTypeString(this: JSValue, globalObject: *JSGlobalObject) *JSC.JSString {
+        return JSC__jsTypeStringForValue(globalObject, this);
+    }
+
+    extern fn JSC__JSValue__constructEmptyObject(globalObject: *JSGlobalObject, prototype: [*c]JSC.JSObject, len: usize) JSValue;
```
paperclover, line 3558: this is covered by `createEmptyObject` and is only used once. should remove

src/bun.js/node/util/parse_args.zig
```diff
@@ -0,0 +1,687 @@
...
+const ParseArgsError = error{ParseError};
+
+/// Represents a slice of a JSValue array
+const ArgsSlice = struct {
+    array: JSValue,
+    start: usize,
+    end: usize,
+};
+
+const TokenKind = enum { positional, option, @"option-terminator" };
```
paperclover, line 38: why not option_terminator?

src/bun.js/node/util/parse_args.zig
```diff
@@ -0,0 +1,687 @@
...
+/// - option (along with value, if any)
+/// - positional
+/// - option-terminator
+fn tokenizeArgs(globalThis: *JSGlobalObject, args: ArgsSlice, options: []const OptionDefinition, tokens: *std.ArrayList(Token)) !void {
+    var index: i32 = -1;
+    var group_count: i32 = 0;
+
+    // build a queue of args to process, because new args can be inserted during the processing
+    var queue_allocator = std.heap.stackFallback(32 * @sizeOf(JSValue), globalThis.allocator());
+    var queue = try std.ArrayList(JSValue).initCapacity(queue_allocator.get(), args.end - args.start);
```
paperclover, line 336: does this need to be ArrayList(JSValue). i dont really like that we are allocating extra js strings just to take the string value out of it.

src/bun.js/node/util/parse_args.zig
```diff
@@ -0,0 +1,687 @@
...
+    defer tokens.deinit();
+
+    try tokenizeArgs(globalThis, args, option_defs.items, &tokens);
+
+    //
+    // Phase 2: process tokens into parsed option values and positionals
+    //
+    log("Phase 2: parse options from tokens (tokens.len={d})", .{tokens.items.len});
+
+    var result_values = JSValue.constructEmptyObject(globalThis, null, 0);
```
paperclover, line 625: use `createEmptyObject`

src/bun.js/node/util/parse_args_utils.zig
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+    return long_option_index;
+}
+
+//
+// TESTS
+//
+
+var no_options: []const OptionDefinition = &[_]OptionDefinition{};
```
paperclover, line 90: note: we don't run the zig tests in ci. it's nice to have, but i sense these may get out of date.

src/bun.js/node/util/validators.zig
```diff
@@ -0,0 +1,233 @@
...
+    globalThis.vm().throwError(globalThis, err);
+    return error.InvalidArgument;
+}
+
+/// -(2^53 - 1)
+pub const NUMBER__MIN_SAFE_INTEGER: i64 = -9007199254740991;
+/// (2^53 – 1)
+pub const NUMBER__MAX_SAFE_INTEGER: i64 = 9007199254740991;
+
+pub fn validateInteger(globalThis: *JSGlobalObject, value: JSValue, comptime name_fmt: string, name_args: anytype, min_value: ?i64, max_value: ?i64) !void {
```
paperclover: i like this pattern. i wonder if making it return the validated data would be a nicer pattern.

paperclover (commented):
this is overall very exciting.

src/bun.js/node/util/parse_args.zig
otgerrogla: I actually had that but changed it because it needs to be an object with null prototype instead of the Object prototype. Node even had [some tests](https://github.com/oven-sh/bun/blob/61d5f5b3f684dfc4a031dd5602ff477e32194726/test/js/node/util/parse_args/parse-args.test.mjs#L438) for this which were failing otherwise:

src/bun.js/bindings/bindings.zig
otgerrogla: I needed this to create an object with null prototype, `createEmptyObject` creates it with the `Object` prototype

src/bun.js/node/util/parse_args.zig
otgerrogla: I feel it nicer to create the output token array, then I can just do `@tagName` on it:
```js
var val = String.static(@as(string, @tagName(token_generic))).toJSConst(globalThis);
```

src/bun.js/node/util/parse_args.zig
otgerrogla: Actually for most part no JSValue are allocated, they are just the input values. The only case is when a short option group is found (e.g. `-abc`) where it needs to create "virtual" arguments and push them into the queue.

I decided to use an array with JSValues, so I can reuse input values for the output as much as possible, to avoid doing: `input JSValue -> String -> JSValue for the output`

src/bun.js/bindings/bindings.zig
paperclover: ah, did not realize the zig `null` turned this into a null prototype. maybe the method name should simply say it's going to have a null prototype, like `createObjectWithNullProto`, but that is a bit verbose. i just suspect `createEmptyObject` and `constructEmptyObject` are going to be very confusing together in the same namespace.

src/bun.js/node/util/parse_args_utils.zig
otgerrogla: Yeah, they were more for my benefit to unit-test the functions ported to zig for correctness. Anyways I don't think the arg token classification is gonna change much

src/bun.js/bindings/bindings.zig
otgerrogla: I decided to go with the same name as the JavaScriptCore equivalent in C++, and leave it generic enough so it can be also used to create objects with an specific prototype different than Object (it's something I'm using for other wip stuff).

I'd rather leave it as it and just add doc comments to specify the differences, but I will change if needed

src/bun.js/node/util/validators.zig
otgerrogla: done!

otgerrogla:
(rebased + fix merge conflict)

src/bun.js/node/util/parse_args.zig
```diff
@@ -0,0 +1,684 @@
...
+        switch (token_generic) {
+            .option => |token| {
+                obj.put(globalThis, ZigString.static("index"), JSValue.jsNumberFromInt32(token.index));
+                obj.put(globalThis, ZigString.static("name"), token.name);
+                obj.put(globalThis, ZigString.static("rawName"), try token.makeRawNameJSValue(globalThis));
+
+                // only for boolean options, it is "undefined"
+                obj.put(globalThis, ZigString.static("value"), token.value);
+                obj.put(globalThis, ZigString.static("inlineValue"), if (token.value.isUndefined()) JSValue.undefined else JSValue.jsBoolean(token.inline_value));
+            },
```
dylan-conway, line 544:
```suggestion
                if (!token.value.isUndefined()) {
```
was something like this meant to be added here?

dylan-conway (approved):
This is great. Going to merge and add a few changes:
- remove `Bun.parseArgs`
- change `constructEmptyObject` to `createEmptyObjectWithNullPrototype`

Jarred-Sumner:
This code will segfault if the garbage collector is run when there are more than 32 arguments.

https://github.com/oven-sh/bun/blob/3963d2de64195b96bcdec4ac4fbc81134806bad5/src/bun.js/node/util/parse_args.zig#L335-L340

I am reverting this PR.

otgerrogla:
@Jarred-Sumner The issue is only for new JSValue that are created dynamically (like when splitting `-ab` into `-a`, `-b`), right?
If it's only reusing existing JSValues from the parseArgs args from callstack, it would be okay to temporarily store them in the heap?

otgerrogla:
By the way, I found two other issues:
- Some string-related memory leaks (due to misunderstanding how String works)
- Panics when options are numbers. Looks like calling `JSValue.put` with a key which is a numeric string triggers an assertion failed:
```
ASSERTION FAILED: !parseIndex(propertyName)
/Users/runner/work/WebKit/WebKit/Source/JavaScriptCore/runtime/JSObjectInlines.h(369) : ASCIILiteral JSC::JSObject::putDirectInternal(VM &, PropertyName, JSValue, unsigned int, PutPropertySlot &)
1   0x104125719 WTFCrash
2   0x10525a01b WTFCrashWithInfo(int, char const*, char const*, int)
3   0x105269b75 WTF::ASCIILiteral JSC::JSObject::putDirectInternal<(JSC::JSObject::PutMode)1>(JSC::VM&, JSC::PropertyName, JSC::JSValue, unsigned int, JSC::PutPropertySlot&)
4   0x10528e527 JSC::JSObject::putDirect(JSC::VM&, JSC::PropertyName, JSC::JSValue, unsigned int)
5   0x103bf41a5 JSC__JSValue__put
```

Working on fixes

PR #7312, opened 2023-11-25
Replace unmodified variable declarations with const

hustLer2k:
### What does this PR do?

There are a lot of declarations using var instead of const that are not mutated. Since Zig's build 0.12 (0.12.0-dev.1664+8ca4a5240) it is considered an error that prevents compilation.

This PR replaces unmodified variable declarations with const.

- [✔️] Code changes

- [✔️ ] I included a test for the new code, or an existing test covers it

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -33,7 +33,7 @@ const color_map = std.ComptimeStringMap([]const u8, .{
 });
 
 fn addInternalPackages(b: *Build, step: *CompileStep, _: std.mem.Allocator, _: []const u8, target: anytype) !void {
-    var io: *Module = brk: {
+    const io: *Module = brk: {
         if (target.isDarwin()) {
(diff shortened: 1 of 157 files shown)


paperclover:
there is already a pr open that does this. and neither of them actually upgrade the version of zig to ensure that this issue doesn't happen again

paperclover:
going to close in favor of the other pr/until we upgrade zig for real

hustLer2k:
send a link pls

hustLer2k:
is there already a zig upgrade PR in progress or i should take on the challenge? i've already resolved some of the other errors that prevent compilation

paperclover:
there is not. i closed the other pr because it is equivilent to this.

i kind of want to wait until after the next release of bun to do the upgrade.

Issue #7261, opened 2023-11-22
Error when parsing keys with hyphen in some JSON files


Azarattum:
### What version of Bun is running?

1.0.13+f5bf67bd1

### What platform is your computer?

Linux 5.15.0-101.103.2.1.el8uek.x86_64 x86_64 x86_64

### What steps can reproduce the bug?

When importing/requiring some JSON files that contain hyphens in the keys, Bun can throw a syntax error. Currently can reproduce with files named `tsconfig.json`. Maybe parsing of `tsconfig` is somehow different?

```sh
echo '{ "key-with-hyphen": true }' > tsconfig.json
bun -e 'require("./tsconfig.json")'
```

Results in:
```
SyntaxError: Unexpected token '-'. Expected '}' to end an export list.
      at /home/tsconfig.json:1:2
```

With files named other than `tsconfig.json` everything works as expected:
```sh
echo '{ "key-with-hyphen": true }' > demo.json
bun -e 'require("./demo.json")'
```

### What is the expected behavior?

All JSON files should be treated and parsed the same.

### What do you see instead?

`tsconfig.json` files throw when threre are hyphens in keys.

### Additional information

The issue was discovered with `tsconfig.json` containing a key for `ts-node` configuration (which includes a hyphen).

Jarred-Sumner:
Interesting

DontBreakAlex:
So it turns out that JSON config files are handled differently
https://github.com/oven-sh/bun/blob/22818e6806cfff66804a74bdbd71f90d195f8183/src/bun.js/module_loader.zig#L1574-L1585 https://github.com/oven-sh/bun/blob/22818e6806cfff66804a74bdbd71f90d195f8183/src/fs.zig#L1608-L1615
They go through are a special parser
https://github.com/oven-sh/bun/blob/22818e6806cfff66804a74bdbd71f90d195f8183/src/json_parser.zig#L885
and are rewritten as modules, thus producing a syntax error in this case:
```TS
var key_with_hyphen = !0

export {
    key_with_hyphen as key-with-hyphen
};

export default { "key-with-hyphen": key_with_hyphen };
```
@Jarred-Sumner why are these files rewritten like this ? Is it for JSON5 compatibility ?

DontBreakAlex:
According to MDN [quoting the alias should work](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export), but I get the same error rewriting it like this:
```JS
var key_with_hyphen = !0

export {
[...]

PR #7316, opened 2023-11-25
Fix error when parsing keys with hyphen in some JSON files

DontBreakAlex:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/7261 introduced by https://github.com/oven-sh/bun/pull/4783.
The solution is to quote export aliases when required so that 

```JS
var key_with_hyphen = !0

export {
    key_with_hyphen as key-with-hyphen
};

export default { "key-with-hyphen": key_with_hyphen };
```
becomes
```JS
var key_with_hyphen = !0

export {
    key_with_hyphen as "key-with-hyphen"
};

export default { "key-with-hyphen": key_with_hyphen };
```

Relevant: https://github.com/tc39/ecma262/pull/2154 and https://bugs.webkit.org/show_bug.cgi?id=217576

I wrote automated tests

I did not add allocations
- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test 07261.test.ts`)

### DIFF
--- a/src/bundler.zig
+++ b/src/bundler.zig
@@ -1435,18 +1435,12 @@ pub const Bundler = struct {
                                     }, prop.key.?.loc),
                                     .value = prop.value.?,
                                 };
-                                var maybe_quoted = name;
-                                if (strings.containsChar(name, '-')) {
-                                    var quote_buf = MutableString.init(allocator, name.len + 2) catch return null;
-                                    quote_buf = js_printer.quoteForJSON(name, quote_buf, false) catch return null;
-                                    maybe_quoted = quote_buf.list.items;
-                                }
                                 export_clauses[i] = js_ast.ClauseItem{
                                     .name = .{
                                         .ref = ref,
                                         .loc = prop.key.?.loc,
                                     },
-                                    .alias = maybe_quoted,
+                                    .alias = name,
                                     .alias_loc = prop.key.?.loc,
                                 };
                                 prop.value = js_ast.Expr.initIdentifier(ref, prop.value.?.loc);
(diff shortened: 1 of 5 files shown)


src/bundler.zig
```diff
@@ -1435,12 +1435,18 @@ pub const Bundler = struct {
                                     }, prop.key.?.loc),
                                     .value = prop.value.?,
                                 };
+                                var maybe_quoted = name;
+                                if (strings.containsChar(name, '-')) {
```
DontBreakAlex: Is there already a function that checks if a string is a valid variable name or do I need to write one ?

src/bundler.zig
```diff
@@ -1435,12 +1435,18 @@ pub const Bundler = struct {
                                     }, prop.key.?.loc),
                                     .value = prop.value.?,
                                 };
+                                var maybe_quoted = name;
```
Jarred-Sumner:
```suggestion
                                var maybe_quoted = MutableString.ensureValidIdentifier(name, allocator) catch return null;
```

src/bundler.zig
Jarred-Sumner: yeah, `MutableString.ensureValidIdentifier`. It will check if it's a valid identifier, and create a new one with the non-valid characters replaced by `_` if it is not valid

src/bundler.zig
DontBreakAlex: We already use this function to create the name of temporary variable, and it replaces the hyphens with underscores. With your suggestion the code becomes this:
```JS
var key_with_hyphen = !0;

export {
  key_with_hyphen
};
export default { "key-with-hyphen": key_with_hyphen };
```

DontBreakAlex:
So I figured It was not the bundler's job to handle quoting stuff. I instead modified to printer to quote aliases if they are not a valid identifier. This removes the need for allocation.

I think this PR is ready to be merged.

paperclover (approved):

paperclover:
thank you

Issue #3621, opened 2023-07-12
Implement `expect.extend()`


Electroid:
https://jestjs.io/docs/expect#expectextendmatchers

```js
import { expect } from "bun:test";

function toBeWithinRange(actual, floor, ceiling) {
  if (
    typeof actual !== 'number' ||
    typeof floor !== 'number' ||
    typeof ceiling !== 'number'
  ) {
    throw new Error('These must be of type number!');
  }

  const pass = actual >= floor && actual <= ceiling;
  if (pass) {
    return {
      message: () =>
        `expected ${this.utils.printReceived(
          actual,
        )} not to be within range ${this.utils.printExpected(
          `${floor} - ${ceiling}`,
        )}`,
      pass: true,
    };
  } else {
    return {
      message: () =>
        `expected ${this.utils.printReceived(
          actual,
        )} to be within range ${this.utils.printExpected(
          `${floor} - ${ceiling}`,
        )}`,
      pass: false,
    };
  }
}

expect.extend({
  toBeWithinRange,
});
```

v3nntrnx:
I can try working on it!

UPD: after working on expect.extend, I realized that it was too hard to me. sorry for the time waste, but I just can't.

mgasner:
fwiw, this is quite important if you're adopting bun (as we would like to do) with a significant base of existing tests that use the react testing library

Electroid:
> I can try working on it!
> 
> UPD: after working on expect.extend, I realized that it was too hard to me. sorry for the time waste, but I just can't.

It's all good, the Bun team is going to take this one.

mgasner:
> fwiw, this is quite important if you're adopting bun (as we would like to do) with a significant base of existing tests that use the react testing library

still!

GuilhermeMGBR:
With support to extensions with `extends` any edge case that would take a while to be prioritized and added to bun could be implemented internally by the team wishing to adopt bun on their workflow.

fatlard1993:
Not a perfect fix/workaround by any means, but for those looking for *something* before official support is added:

You can always use the standard expect assertions to use your matchers, Here's an example using @testing-library/jest-dom/matchers:
```
import { toBeVisible } from '@testing-library/jest-dom/matchers';
...
expect(toBeVisible(getByRole(container, 'combobox')).pass).toBe(true);
```
VS
```
import * as matchers from '@testing-library/jest-dom/matchers';
...
expect.extend(matchers);
...
[...]

PR #7319, opened 2023-11-25
feat(test): expect.extend and asymmetric matcher improvements

otgerrogla:
### What does this PR do?

This PR implements the following to `bun test`:
 - `expect.extend({ ... })` support to register custom matchers (#3621)
    - Note: support is still not 100%, it is missing some helpers and variables in the matcher context (i.e. the `this` and `this.utils` inside custom matcher functions)
    - Support for declaring typescript types for custom matchers through declaration merging (same way as jest)
    - Added related tests from Jest
 - `expect.not` support, to negate of asymmetric matchers (e.g. `expect.not.stringContaining('bun')`)
 - Add new asymmetric matchers from Jest (including related tests):
   - `expect.closeTo`
   - `expect.objectContaining`
 - ~Fixes for existing asymmetric matchers:~
   - ~`expect.arrayContaining([])` (with empty array) should match anything, even numbers (that's what jest does)~
 - Adds `expect.resolvedTo` / `expect.rejectedTo`, to support for asynchronous asymmetric matchers (e.g. `expect.resolvedTo.stringContaining('bun')`). This is not in jest, although it existed as a feature proposal: https://github.com/jestjs/jest/issues/10556
 - Some refactoring of the `bun:test` typescript types to accomodate custom matchers, the expect.not, etc.
 - Tests:
   - Tests for all new features and fixes above at `expect.test.js` and `expect-extend.test.js`
   - Improved interop compatibility to run these test files as-is on bun, jest, and vitest
   - Added type-checking to these test files, so they can be used as type-tests for types of `bun:test` (although not automated as of now, just for visual testing)

Implementation notes:
 - ~It adds some capabilities to `generate-classes.ts` (in order to support custom matchers):~
   - ~Option to definine a `getOwnPropertySlot` callback, to dynamically resolve properties at native level.~
   - ~Option to define an `apply` callback, executed when the class instance is called as a function (the class itself, not the constructor).~
 - ~Uses `getOwnPropertySlot` to dynamically resolve custom matchers to improve performance (jest actually populates a new object with all matchers each time)~
 - The custom matchers are currently stored in a singleton object in the ZigGlobalObject

- [X] Code changes

I wrote automated tests (some created + some ported from jest)


Jarred-Sumner:
Thank you very much for this. 

`getOwnPropertySlot` is a neat approach but the downside is it prevents inline-caching, which means that all property accesses will be slow on the `expect` instances (something like 10x - 100x slower). This would have a meaningfully negative performance impact, even when no custom matchers are in use.

There's another way that won't have a performance impact when not in use and will probably be less code. We can directly mutate the prototype used by the `Expect` object. We can define the custom matchers as functions on the prototype. For asymmetric matchers, we can `.get` on that prototype. 

I think it'd also be good to find a way to avoid making the flags a `JSValue`

otgerrogla:
> There's another way that won't have a performance impact when not in use and will probably be less code. We can directly mutate the prototype used by the `Expect` object. We can define the custom matchers as functions on the prototype. For asymmetric matchers, we can `.get` on that prototype.

It's not so direct, because when calling e.g. `expect.myCustomMatcher()` it doesn't directly call the user-provided js matcher function, but calls a native function that wraps the call and applies some logic (parsing and validating the user-provided function return value, etc.)

I guess this can be achieved by using bound native functions. Any other/better idea?

otgerrogla:
Actually bound functions won't work, because you lose the reference to the "this" and can't access the matcher flags

otgerrogla:
Doesn't work with bound functions either, because you lose the reference to "this" and can't retrieve the flags

src/bun.js/bindings/bindings.cpp
```diff
@@ -232,20 +269,24 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
...
-                JSArray* otherArray = jsDynamicCast<JSArray*>(otherProp);
+        if (JSC::isArray(globalObject, expectedArrayValue)) {
+            JSArray* expectedArray = jsDynamicCast<JSArray*>(expectedArrayValue);
+            unsigned expectedLength = expectedArray->length();
 
-                unsigned expectedLength = expectedArray->length();
-                unsigned otherLength = otherArray->length();
+            if (expectedLength == 0) {
+                // jest makes it so an empty array matches anything,
+                // even a primitive like a string or a number
```
Electroid: What's an example of this? We have deviated a handfull of times from Jest where something was _obviously_ wrong. (aka. potential to break is outweighed by catching likely errored code)

src/bun.js/bindings/bindings.zig
```diff
@@ -5872,3 +5927,30 @@ pub const ScriptExecutionStatus = enum(i32) {
...
+// ECMA 262-3 8.6.1
+// Property attributes
+pub const PropertyAttribute = struct {
+    pub const ReadOnly: PropertyAttributes = 1 << 1; // property can be only read, not written
+    pub const DontEnum: PropertyAttributes = 1 << 2; // property doesn't appear in (for .. in ..)
+    pub const DontDelete: PropertyAttributes = 1 << 3; // property can't be deleted
+    pub const Accessor: PropertyAttributes = 1 << 4; // property is a getter/setter
+    pub const CustomAccessor: PropertyAttributes = 1 << 5;
+    pub const CustomValue: PropertyAttributes = 1 << 6;
+};
```
Electroid: This would be a good candidate to use Zig's [packed struct](https://devlog.hexops.com/2022/packed-structs-in-zig/#zig-packed-structs).

src/bun.js/test/expect.zig
```diff
@@ -57,8 +69,30 @@ pub const Expect = struct {
...
+        pub fn toBitset(this: Flags) BitsetType {
+            return (if (this.not) FLAG_NOT else 0) | (if (this.promise == .resolves) FLAG_RESOLVES else 0) | (if (this.promise == .rejects) FLAG_REJECTS else 0);
+        }
+
+        pub fn fromBitset(bitset: BitsetType) Flags {
+            return Flags{
+                .not = (bitset & FLAG_NOT) != 0,
+                .promise = if ((bitset & FLAG_RESOLVES) != 0) .resolves else if ((bitset & FLAG_REJECTS) != 0) .rejects else .none,
+            };
+        }
```
Electroid: Same comment, good candidate for packed structs.

src/bun.js/test/expect.zig
```diff
@@ -3776,6 +4101,98 @@ pub const ExpectStringMatching = struct {
...
+            globalObject.throwPretty("<d>expect.<r>closeTo<d>(number, <r>precision?<d>)<r>\n\nPrecision must be a number or undefined", .{});
+            return .zero;
+        }
+
+        if (Jest.runner.?.pending_test == null) {
+            const err = globalObject.createErrorInstance("expect.closeTo() must be called in a test", .{});
+            err.put(globalObject, ZigString.static("name"), ZigString.init("TestNotRunningError").toValueGC(globalObject));
+            globalObject.throwValue(err);
+            return .zero;
+        }
```
Electroid, line 4325: We should fix this, this technically isn't right. Asymmetric matchers should be able to be defined outside the context of a test. For example:

```js
test.each([
  {
    actual: { a: 1, b: 2 },
    // this is not called in the test scope, but in global scope
    expected: { a: expect.closeTo(1), b: expect.closeTo(2) },
  }
])("%s", ({ actual, expected }) => {
  expect(actual).toEqual(expected);
});
```

I think we can just remove these checks for the matchers.

Electroid (commented):
Very exciting changes. Someone else on the team will do a more through review, but left a few comments

src/bun.js/bindings/bindings.cpp
otgerrogla: I added some tests to `expect.test.js`, and ran them on Jest where they pass
```js
    test("ArrayContaining matches", () => {
      expect(expect.arrayContaining([])).toEqual(42);
      expect(expect.arrayContaining([])).toEqual("jest");
      expect(expect.arrayContaining(["foo"])).toEqual(["foo"]);
      expect(expect.arrayContaining(["foo"])).toEqual(["foo", "bar"]);
      expect(expect.arrayContaining([])).toEqual({});
    });
```
Although I agree this is a very weird behavior

src/bun.js/test/expect.zig
otgerrogla: I put it in because I saw the existing asymmetric matchers had this check, although I agree this is a good point

src/bun.js/bindings/bindings.zig
otgerrogla: I though about it, but wanted to mimic the C++ definition as close as possible.
Anyways I'm gonna remove this part

src/bun.js/test/expect.zig
otgerrogla: Indeed, this makes more sense here
Thanks

src/bun.js/test/expect.zig
otgerrogla: Should I remove it from all asymmetric matchers?

Electroid: Yep

otgerrogla:
@Jarred-Sumner I replaced the usage of `getOwnPropertySlot`, so now when `expect.extend` is called, it mutates the prototypes/constructor by adding new `JSValue` instances of `JSFunction` which point to a same native function, but each has a javascript-level property `fn` with the user-provided matcher function so it can known which custom matcher to use. That's the best way I could came up with to do it while still allowing the matchers to access the flags.

I also moved the flags from JSValues to being fields in the zig structs themselves.

src/bun.js/test/expect.zig
```diff
@@ -3707,10 +3803,197 @@ pub const Expect = struct {
...
+            globalObject.throwValue(err);
+            return false;
+        };
+        if (flags.not) pass = !pass;
+        if (pass or silent) return pass;
+        // handle failure
+        const message_fn = matcher_result.get(globalObject, "message") orelse JSValue.undefined;
+        if (message_fn.isCallable(globalObject.vm())) {
+            var message_result = message_fn.callWithGlobalThis(globalObject, &[_]JSValue{});
+            if (message_result.isAnyError()) {
```
Jarred-Sumner:
```suggestion
            if (message_result.toError(globalObject)) |err| {
                globalObject.bunVM().onUnhandledError(globalObject, err);
                return false;
```

src/bun.js/test/expect.zig
```diff
@@ -3707,10 +3803,197 @@ pub const Expect = struct {
...
+        value = Expect.processPromise(expect.flags, globalObject, value, "[[custom matcher]]", "<green>expected<r>", false) orelse return .zero;
+        value.ensureStillAlive();
+
+        active_test_expectation_counter.actual += 1;
+
+        // prepare the args array
+        const args_ptr = callFrame.argumentsPtr();
+        const args_count = callFrame.argumentsCount();
+        var allocator = std.heap.stackFallback(8 * @sizeOf(JSValue), globalObject.allocator());
+        var matcher_args = std.ArrayList(JSValue).initCapacity(allocator.get(), args_count + 1) catch {
```
Jarred-Sumner, line 4050: Can you add a test that calls a custom matcher 10,000 times in a loop?

This current implementation will likely cause GC issues & crashes. JSC's GC stack scanning is limited. I kind of think we might need to implement at least part of this in C++ to avoid that situation. 

In C++, I would use a `ArgsList` here

src/bun.js/test/expect.zig
```diff
@@ -3948,10 +4459,382 @@ pub const ExpectArrayContaining = struct {
...
+        if (ExpectStatic.fromJS(thisValue)) |expect_static| {
+            flags = expect_static.flags;
+        } else {
+            // if it's not an ExpectStatic instance, it was called from the Expect constructor, so assume default flags
+            flags = .{};
+        }
+
+        // retrieve the user-provided matcher function
+        var func: JSValue = callFrame.callee();
+        var matcher_fn = func.get(globalObject, Expect.custom_matcher_fn_property) orelse unreachable;
```
Jarred-Sumner: We can use a `JSC::JSBoundFunction` here I think. JSBoundFunction can preserve the `this` and any additional arguments. `JSBoundFunction` gets inlined in the JIT, so it has very little overhead.

If we cannot use a JSBoundFunction, then the backup would be to store the custom matcher as a `WriteBarrier` on the instance itself. Then we can avoid the (observable) property access. This can be done by subclassing `JSFunction`. See `JSFFIFunction` for an example of that.

otgerrogla: I tried with a bound function but got no way to recover the Expect/ExpectStatic instance to retrieve the flags as you need to specify the `this` on creation (i.e. in the expect.extends execution).

I will do the second way, I thought about subclassing `JSFunction` but didn't know how to go about it, thanks for the tip!

Jarred-Sumner: Another option would be to make custom matchers a `CustomGetterSetter` which returns the JSFunction subclass.

It would work like [EventListenerMap in WebCore](https://github.com/oven-sh/WebKit/blob/0aa1f6dfc9b48be1e89b8f709fd44e3c88feaf33/Source/WebCore/dom/EventListenerMap.h#L81-L89):


```c++
template<typename Visitor>
void EventListenerMap::visitJSEventListeners(Visitor& visitor)
{
    Locker locker { m_lock };
    for (auto& entry : m_entries) {
        for (auto& eventListener : entry.second)
            eventListener->callback().visitJSFunction(visitor);
    }
}
```

The functions would lazily be created on each instance on access. Then, cached and visited within [`visitAdditionalChildren`](https://github.com/oven-sh/WebKit/blob/0aa1f6dfc9b48be1e89b8f709fd44e3c88feaf33/Source/WebCore/bindings/js/JSWorkerGlobalScopeCustom.cpp#L40-L54) on each instance:
```c++
template<typename Visitor>
void JSWorkerGlobalScope::visitAdditionalChildren(Visitor& visitor)
{
    if (auto* location = wrapped().optionalLocation())
        addWebCoreOpaqueRoot(visitor, *location);
    if (auto* navigator = wrapped().optionalNavigator())
        addWebCoreOpaqueRoot(visitor, *navigator);
    ScriptExecutionContext& context = wrapped();
    addWebCoreOpaqueRoot(visitor, context);
    
    // Normally JSEventTargetCustom.cpp's JSEventTarget::visitAdditionalChildren() would call this. But
    // even though WorkerGlobalScope is an EventTarget, JSWorkerGlobalScope does not subclass
    // JSEventTarget, so we need to do this here.
    wrapped().visitJSEventListeners(visitor);
}
```

src/bun.js/test/expect.zig
otgerrogla: Run this test locally several times with 1M calls and didn't crash, I think this is safe as the JSValue arguments are already "pinned" (although I might be wrong):
- For symmetric matchers:
  - First arg (the expected value): stored/owned on the `Expect` instance
  - Rest: pinned on the callFrame
- For asymmetric matchers:
  - First arg (the expected value): comes from a callFrame up above (deepEquals or whatever call)
  - Rest: stored/owned in a JSArray at the `ExpectCustomAsymmetricMatcher` instance
 
Also what difference would using an `ArgList` make? As I got it you still have to allocate an array to put the JSValues, as you need to prepend the expected value for the call

src/bun.js/bindings/bindings.cpp
Electroid: Let's change those to fail, so `expect.arrayContaining` fails when not an array. Same with the other asymmetric matchers like `expect.objectContaining`. This just defies common sense, if code breaks from these paths, it was already broken.

src/bun.js/test/expect.zig
otgerrogla: Did it with the JSFunction subclassing approach, seems to work fine!
This way you can access both the the expect instance to get the flags (`callframe.this()`) and recover the user-provided matcher function from `calframe.callee()` which is the new subclass instance

test/js/bun/test/expect-extend.test.js
```diff
@@ -0,0 +1,213 @@
...
+      default: undefined,
+    }),
+  ).toThrow(/*'expect.extend: `default` is not a valid matcher. Must be a function, is "undefined"'*/);
+  expect(() =>
+    expect.extend({
+      // @ts-expect-error
+      default: 42,
+    }),
+  ).toThrow(/*'expect.extend: `default` is not a valid matcher. Must be a function, is "number"'*/);
+  expect(() =>
```
dylan-conway, line 200: were these meant to be commented? they still pass uncommented

src/bun.js/bindings/bindings.cpp
```diff
@@ -129,23 +129,51 @@ using namespace JSC;
...
 enum class AsymmetricMatcherResult : uint8_t {
     PASS,
     FAIL,
     NOT_MATCHER,
 };
 
-AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSCell* matcherPropCell, JSValue otherProp, ThrowScope* throwScope)
+template<class T>
+bool readFlagsAndProcessPromise(JSValue& instanceValue, T* instance, ExpectFlags& flags, JSGlobalObject* globalObject, JSValue& value)
+{
```
dylan-conway, line 149: is the template necessary? `instance` doesn't seem to be used

src/bun.js/bindings/bindings.cpp
```diff
@@ -272,11 +309,74 @@ AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSC
...
     }
 
     return AsymmetricMatcherResult::NOT_MATCHER;
 }
 
+AsymmetricMatcherResult matchAsymmetricMatcher(JSGlobalObject* globalObject, JSValue matcherProp, JSValue otherProp, ThrowScope* throwScope)
+{
+    ExpectFlags flags = ExpectFlags();
+    AsymmetricMatcherResult result = matchAsymmetricMatcherAndGetFlags(globalObject, matcherProp, otherProp, throwScope, flags);
+    if (result != AsymmetricMatcherResult::NOT_MATCHER && (flags & FLAG_NOT)) {
```
dylan-conway, line 373: In jest both `expect.not.any` and `expect.not.anything` are not valid asymmetric matchers. We will need to throw for both of them.
Screenshot 2023-11-28 at 4 14 24 PM

src/bun.js/test/pretty_format.zig
```diff
@@ -1986,4 +1951,141 @@ pub const JestPrettyFormat = struct {
...
+            } else {
+                this.addForNewLine("Anything".len);
+                writer.writeAll("Anything");
+            }
+        } else if (value.as(expect.ExpectAny)) |matcher| {
+            const constructor_value = expect.ExpectAny.constructorValueGetCached(value) orelse return true;
+
+            printAsymmetricMatcherPromisePrefix(matcher.flags, this, writer);
+            if (matcher.flags.not) {
+                this.addForNewLine("Any<".len);
```
dylan-conway: `"NotAny<"` and `"Any<"` need to swap places

src/bun.js/bindings/bindings.cpp
otgerrogla: true! it was a leftover of when the flags where stored as JSValue

src/bun.js/test/pretty_format.zig
otgerrogla: good catch!

src/bun.js/bindings/bindings.cpp
otgerrogla: I know, but I feel like we can support them "for free" as a feature, as it's not gonna break any existing jest test. They might be useful, e.g.:
- match a null or undefined with `expect.not.anything()`
- match something that's not a Date with `expect.not.any(Date)`

dylan-conway: I agree, let's keep them. `expect.not.any(Class)` seems very useful

test/js/bun/test/expect-extend.test.js
otgerrogla: forgot about them, thanks

otgerrogla:
Pushed some more changes because the error messages and error handling for the custom matchers were not working properly. Also added support for async matchers (custom matchers that return a promise), and added tests for these cases.

src/bun.js/bindings/bindings.cpp
```diff
@@ -1685,6 +1805,19 @@ void JSC__JSFunction__optimizeSoon(JSC__JSValue JSValue0)
     JSC::optimizeNextInvocation(value);
 }
 
+bool JSC__JSFunction__getSourceCode(JSC__JSValue JSValue0, ZigString* outSourceCode)
+{
+    JSC::JSValue value = JSC::JSValue::decode(JSValue0);
+    if (JSC::JSFunction* func = jsDynamicCast<JSC::JSFunction*>(value)) {
+        auto* sourceCode = func->sourceCode();
+        if (sourceCode != nullptr) { // native functions have no source code
+            *outSourceCode = Zig::toZigString(sourceCode->view());
```
Jarred-Sumner, line 1814: This will become invalid memory

It should be a BunString.

src/bun.js/bindings/bindings.cpp
```diff
@@ -1685,6 +1805,19 @@ void JSC__JSFunction__optimizeSoon(JSC__JSValue JSValue0)
     JSC::optimizeNextInvocation(value);
 }
 
+bool JSC__JSFunction__getSourceCode(JSC__JSValue JSValue0, ZigString* outSourceCode)
```
Jarred-Sumner, line 1808: Do we need to do this? `toString` on a function returns the source code for the function

Jarred-Sumner:
Thank you for very much. I'm going to merge this and then make some small tweaks.

PR #7320, opened 2023-11-26
feat(test): `toHaveBeenNthCalledWith` + improve some fail messages

james-elicx:
### What does this PR do?

- Implement jest matcher `toHaveBeenNthCalledWith`.
- Updates that it's supported in the docs.
- Improve fail messages for `toHaveBeenCalled` + `toHaveBeenCalledTimes` so that they provide more info that jest failures provide.

.

- [x] Code changes

I wrote automated tests

If Zig files changed:


Jarred-Sumner (approved):
Thank you

src/bun.js/test/expect.zig
```diff
@@ -3389,8 +3379,8 @@ pub const Expect = struct {
             return .zero;
         }
 
-        if (arguments.len < 1 or !arguments[0].isAnyInt()) {
```
Electroid, line 3392: Does `isAnyInt` fail for `0`?

james-elicx: No. The reason I changed it from `isAnyInt` to `isUInt32AsAnyInt` was so that it would fail for negative numbers since this matcher is only supposed to accept an argument that is `>=0`. Jest throws a matcher error for the arg type saying that it expects a non-negative argument, so this aligns the behaviour with how it works in Jest.

Issue #7262, opened 2023-11-22
Use trailing commas when printing JS objects & arrays


ArnaudBarre:
### What is the problem this feature would solve?

When copy pasting something from the console to somewhere else, it's easier to use multi cursor or regex on it.
Plus that's the standard in code (default in prettier 3), so it will look more familiar.

Actual
```
{
  a: 1,
  b: 2
}
```
Expected
```
{
  a: 1,
  b: 2,
}
```

### What is the feature you are proposing to solve the problem?

See title

Jarred-Sumner:
I like this suggestion

imdevtej:
Can i work on this issue ?
@ArnaudBarre

ArnaudBarre:
See with the team, I have no time to learn zig for now 😁

Electroid:
Closed by #7334

PR #7334, opened 2023-11-27
Use trailing commas when printing multiline JS objects

hustLer2k:
### What does this PR do?

This PR implements https://github.com/oven-sh/bun/issues/7262

- [✓ ] Code changes

- [ ✓] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [✓ ] I included a test for the new code, or an existing test covers it
- [✓ ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed
- [ ✓] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test-file-name.test`)


Jarred-Sumner:
Thank you for this. There probably will be some tests to update. I'm not totally confident this will cover every case (such as when there is only one object to print)

hustLer2k:
Could you suggest where the related tests are located?

paperclover (changes requested):
looks great. there are more test failures. all of these are going to be fixing the test to match the new output. after that this can be merged.

one test failure:

image

paperclover:
list of failing test files to check for, some of these might just be known flaky tests:
  
  - `test/js/bun/test/test-test.test.ts`
  - `test/js/bun/util/inspect.test.js`
  - `test/js/node/util/bun-inspect.test.ts`
  - `test/js/node/util/custom-inspect.test.js`
  - `test/js/web/fetch/headers.test.ts`
  - `test/js/web/html/URLSearchParams.test.ts`
  - `test/js/web/url/url.test.ts`
  - `test/transpiler/property.test.ts`

paperclover (approved):
great.

PR #7335, opened 2023-11-27
feat: syntax highlight stack traces + overall visual improvements

Jarred-Sumner:
### What does this PR do?

This fixes a number of bugs where Bun would not log a stack trace or it would not log source lines for a stack trace. This also makes `SyntaxError` report a line number which is usually more accurate but honestly not always. There remains a pre-existing off-by-one error somewhere in the code which is very hard to figure out where. 

We now parse `error.stack` as a string when `Vector<JSC::StackFrame>` on `JSC::ErrorInstance` is empty.  This implements a V8 error.stack property parser in C++, which is then fed back into Zig. This means that error.stack modified by userland code will be logged to the console.

We now syntax highlight source code previews using a very simple (and not super accurate) syntax highlighter.  Our source code previewer splits it up line by line. 

Here's what some of `http.ts` looks like in my terminal:

image

This really needs some syntax highlighter tests. I ran it on `checker.ts` and confirmed it does not infinite loop or otherwise look bad in that case. There's some strangeness where it will highlight typescript keywords as though they were keywords when, in context, they are not keywords - they are variable names. This can't really be solved unless we were to fully parse the input, but fully parsing the input would be both much more complicated and slower. Another approach is using a TreeSitter-like syntax highlighter, but I think it's kind of overkill for our usecase. I think this approach is a good compromise between fast, simple, and works well enough.


src/bun.zig
```diff
@@ -34,6 +34,465 @@ pub const path = @import("./resolver/resolve_path.zig");
 pub const fmt = struct {
     pub usingnamespace std.fmt;
 
+    pub fn fmtJavaScript(text: []const u8, enable_ansi_colors: bool) QuickAndDirtyJavaScriptSyntaxHighlighter {
```
paperclover, line 37: why was the syntax highlighter put here in bun.zig/

src/bun.js/bindings/bindings.cpp
```diff
@@ -3613,6 +3614,133 @@ static void populateStackFrame(JSC::VM& vm, ZigStackTrace* trace, const JSC::Sta
         is_top ? trace->source_lines_numbers : nullptr,
         is_top ? trace->source_lines_to_collect : 0, &frame->position);
 }
+
+class V8StackTraceIterator {
```
paperclover, line 3626: imo this should belong in it's own file

src/bun.js/bindings/bindings.cpp
```diff
@@ -3613,6 +3614,133 @@ static void populateStackFrame(JSC::VM& vm, ZigStackTrace* trace, const JSC::Sta
...
+        if (functionName == "<anonymous>"_s) {
+            functionName = StringView();
+        }
+
+        if (functionName == "global code"_s) {
+            functionName = StringView();
+            frame.isGlobalCode = true;
+        }
+
+        if (functionName.length() > 4 && functionName.startsWith("new "_s)) {
```
paperclover: this length check is not needed

image

src/bun.js/javascript.zig
```diff
@@ -2716,17 +2744,18 @@ pub const VirtualMachine = struct {
...
 
-            const int_size = std.fmt.count("{d}", .{source.line});
+            const int_size = std.fmt.count("{d}", .{display_line});
             const pad = max_line_number_pad - int_size;
             last_pad = pad;
             try writer.writeByteNTimes(' ', pad);
 
             try writer.print(
-                comptime Output.prettyFmt("<r><d>{d} | <r>{s}\n", allow_ansi_color),
+                comptime Output.prettyFmt("<r><d>{d} | <r>{any}\n", allow_ansi_color),
```
paperclover: avoid `{any}` as it will find a way to format anything, using `{}` is a little safer

paperclover (approved):
yay

paperclover:
merge after tests

paperclover:
image

PR #7344, opened 2023-11-28
Glob fixes + optimization

zackradisic:
### What does this PR do?

1. Fix crashes in glob related to handling sys errors
2. Add tests to test codepaths where glob handles errors
3. Makes glob faster by avoiding unnecessary stat syscall when handling symlinks

- [x] Code changes


src/glob.zig
```diff
@@ -304,93 +305,68 @@ pub fn GlobWalker_(
...
+                                    this.iter_state = .get_next;
+                                    const maybe_dir_fd: ?bun.FileDescriptor = switch (Syscall.openat(this.cwd_fd, symlink_full_path_z, std.os.O.DIRECTORY | std.os.O.RDONLY, 0)) {
+                                        .err => |err| brk: {
+                                            if (@as(usize, @intCast(err.errno)) == @as(usize, @intFromEnum(bun.C.E.NOTDIR))) {
+                                                break :brk null;
+                                            }
+                                            if (this.walker.error_on_broken_symlinks) return .{ .err = this.walker.handleSysErrWithPath(err, symlink_full_path_z) };
+                                            // Broken symlink, but if `only_files` is false we still want to append
+                                            // it to the matched paths
+                                            if (!this.walker.only_files) {
```
Jarred-Sumner, line 375: why open it immediately instead of enqueuing the path to be opened? doesn't this risk:
1) `EMFILE` (too many open file descriptors)
2) stack overflow if the number of directories is large enough?

zackradisic: The way the code works it should only ever have the cwd and current open directory file descriptors open, but just to make sure I added some checks in debug mode to assert the # of open file descriptors is <= 4, it never goes above 2

This code won't run into stack overflow because the queue is heap allocated

src/glob.zig
```diff
@@ -193,6 +196,7 @@ pub fn GlobWalker_(
                     .err => |err| return .{ .err = this.walker.handleSysErrWithPath(err, @ptrCast(path_buf[0 .. root_path.len + 1])) },
                     .result => |fd| fd,
                 };
+                this.fds_open += 1;
```
dylan-conway:
```suggestion
if (bun.Environment.allow_assert) {
                this.fds_open += 1;
}
```

dylan-conway (approved):

Issue #7352, opened 2023-11-28
family is undefined when listening at localhost (net:server)


Hanaasagi:
### What version of Bun is running?

1.0.14+d8be3e51b

### What platform is your computer?

Linux 6.6.2-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

```JavaScript
const net = require("node:net");
const server = net
  .createServer(socket => {
    socket.end("goodbye\n");
  })
  .on("error", err => {
    throw err;
  });

server.listen(0, "localhost", () => {
  console.log("opened server on", server.address());
});

```

### What is the expected behavior?

```
opened server on { address: '127.0.0.1', family: 'IPv4', port: 38529 }

```

### What do you see instead?

```
opened server on {
  port: 35349,
  address: "localhost",
  family: undefined
}

```

### Additional information

https://github.com/oven-sh/bun/blob/f5115835da965921b5f4e59a85149217835df65b/src/js/node/net.js#L768-L769

https://nodejs.org/api/net.html#class-netsocketaddress

PR #7353, opened 2023-11-28
fix(node:net): fix local address of `net.server`

Hanaasagi:
### What does this PR do?

Close https://github.com/oven-sh/bun/issues/7352

- [x] Code changes

 wrote automated tests

### DIFF
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -887,6 +887,38 @@ pub const Listener = struct {
         return JSValue.jsNumber(this.connection.host.port);
     }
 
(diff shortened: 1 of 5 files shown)


Hanaasagi:
There is a problem with test case. It is actually bound to `::`, not `0.0.0.0`. see https://github.com/oven-sh/bun/issues/7355

Hanaasagi:
I really need `server.address()` to work, but it has been blocking for a long time.

nektro:
sorry we didnt get to this!
soon to be fixed by https://github.com/oven-sh/bun/pull/18771

PR #7364, opened 2023-11-29
Fix scripts/download-zig.sh

joeyw:
### What does this PR do?

Ubuntu 22.04 x86_64 fresh install and bun build attempt using latest contributing guide.

`bun setup` was failing with:

image

Resulting from `scripts/download-zig.sh`

```bash
$ scripts/download-zig.sh

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

After removing the gzip flag, the following error then resulted

```bash
$ uname -i 
x86_64
$ scripts/download-zig.sh
$ .cache/zig/zig
bash: .cache/zig/zig: cannot execute binary file: Exec format error
```

```bash
$ rm -rf .cache
$ scripts/download-zig.sh
$ .cache/zig/zig version
0.12.0-dev.1604+caae40c21
```

`bun setup`
image

### DIFF
--- a/scripts/download-zig.sh
+++ b/scripts/download-zig.sh
@@ -79,7 +79,7 @@ fi
 
 rm -rf "${extract_at}"
 mkdir -p "${extract_at}"
-tar -xzf "${dest}" -C "${extract_at}" --strip-components=1
+tar -xf "${dest}" -C "${extract_at}" --strip-components=1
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you

Issue #7355, opened 2023-11-28
When `0.0.0.0` is specified, server is bound `::` (`net:server`)


Hanaasagi:
### What version of Bun is running?

1.0.14+d8be3e51b

### What platform is your computer?

Linux 6.6.2-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

x.js
```JavaScript
const net = require("node:net");
const server = net
  .createServer(socket => {
    socket.end("goodbye\n");
  })
  .on("error", err => {
    throw err;
  });

server.listen(0, "0.0.0.0", () => {
  console.log("opened server on", server.address());
});

```

```
>>​ sudo strace -ebind ~/.bun/bin/bun run x.js
[sudo] password for kumiko:
bind(13, {sa_family=AF_NETLINK, nl_pid=0, nl_groups=00000000}, 12) = 0
bind(13, {sa_family=AF_INET6, sin6_port=htons(0), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::", &sin6_addr), sin6_scope_id=0}, 28) = 0


```

### What is the expected behavior?

Binding `0.0.0.0`, like node

```
>>​ sudo strace -ebind node x.js
bind(24, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("0.0.0.0")}, 16) = 0
```

### What do you see instead?

Binding `::`

### Additional information

https://github.com/oven-sh/bun/blob/f5115835da965921b5f4e59a85149217835df65b/src/bun.js/api/bun/socket.zig#L22-L28

Electroid:
Presumably, this is also an issue for `Bun.serve`?

Hanaasagi:
> Presumably, this is also an issue for Bun.serve?

No, `Bun.serve` works as expected.


```JavaScript
Bun.serve({
  hostname: "0.0.0.0",
  port: 8080,
  fetch(request) {
    console.log(request.url);
  },
});

```

```
>>​ sudo strace -ebind ~/.bun/bin/bun run server.js
bind(13, {sa_family=AF_INET, sin_port=htons(8080), sin_addr=inet_addr("0.0.0.0")}, 16) = 0

```

PR #7366, opened 2023-11-29
fix(socket): fix `0.0.0.0` binding address

Hanaasagi:
### What does this PR do?

Close: #7355

- [x] Code changes

existed tests.


Hanaasagi:
Related to https://github.com/oven-sh/bun/pull/3758. I'm not sure why `0.0.0.0` is being changed to `null`; this causes even the use of `0.0.0.0` to bind to IPv6 `::`.

paperclover (changes requested):
can you add a test specifically listening on 0.0.0.0 and then connecting over ipv6 and ensuring that the connect fails.

(or do you think that test would not be reliable?)

Hanaasagi:
> can you add a test specifically listening on 0.0.0.0 and then connecting over ipv6 and ensuring that the connect fails.

Local tested this on Linux, and it works fine. Waiting the drawin ci results.

paperclover (approved):

paperclover:
image

it seems some tests fail. the server in this case is listening on 0.0.0.0, but the connection is unclear what the default is. im not sure if this is a bad test (does node default to ipv6 in `new Socket`)

Hanaasagi:
Good catch. I found that both Linux and Darwin failed in CI. But it worked fine on my local machine (Arch Linux). Need some time to investigate, I convert this to draft.

Hanaasagi:
made a simple action test for this, when the server listens on 0.0.0.0, client connected without `hostname`, node will also use `::1`.

https://github.com/Hanaasagi/bun-ci-7366/actions/runs/7192177485/job/19588118516

paperclover (approved):

PR #7368, opened 2023-11-29
Fix imports in run-tests example

mostlyinvisible:
### What does this PR do?

Adds `describe` to the import statement in an example script for running tests. Without this, the example does not run properly and returns an error.


paperclover (approved):

paperclover:
thanks

Issue #7034, opened 2023-11-10
Bunx: support specifying package to install


davidlj95:
### What is the problem this feature would solve?

If the binary to run is not named same as the package, you won't be able to run it with `bunx`. There are some binaries included in packages that don't have same name as package. 

For instance, [Renovate]: a tool to update dependencies. Their package provides [two binaries](https://github.com/renovatebot/renovate/blob/35.159.7/package.json#L6-L7):

1. `renovate` (same as package name) to run the dependencies update bot
2. `renovate-config-validator`: to validate your config

[Renovate]: https://www.mend.io/renovate/

You can run the bot with `bunx renovate`. But you can't run the `renovate-config-validator` using `bunx` right now. 

Other package managers offer the `-p` CLI option to specify which package to install, then the first argument is the binary to run. For instance, following previous example:

- **NPM offers [`npm exec --package`](https://docs.npmjs.com/cli/v10/commands/npm-exec)/[`npx -p|--package`](https://docs.npmjs.com/cli/v10/commands/npx)**
  We can run the script with:
  - `npm exec --package renovate renovate-config-validator`
  - `npx --package renovate renovate-config-validator`
- **Yarn (berry) offers [`yarn dlx -p|--package`](https://yarnpkg.com/cli/dlx#options)**
  We can run the script with:
  - `yarn dlx -p renovate renovate-config-validator`
 

### What is the feature you are proposing to solve the problem?

Would be nice to have also some `-p` or `--package` to enable this use case :) 

### What alternatives have you considered?

Installing the dependency, then use `bun run` with path to script in `node_modules` directory. But the great of `bunx` is to just install and run for dependencies that won't be needed often. Like validating the configuration of your dependencies update tool.

Until then, other package managers can be used for this purpose

telekid:
I'll take a look at this when I have a bit.

SergheiGurgurov:
this feature would be really useful to spin up angular projects.
installing angular globally is not a good approach if u work on multiple projects.

currently the workflow for a local installation with bun looks like this
```bash
bun install @angular/cli
# > delete package.json and lockfile cause the conflict with the ng-new command
bunx ng new my-project --directory .
```

the npx counterpart looks like this

```bash
npx -p @angular/cli ng new my-project
```

davidlj95:
> this feature would be really useful to spin up angular projects. installing angular globally is not a good approach if u work on multiple projects.
> 
@SergheiGurgurov in this case, there's a solution. You can run

```
[...]

PR #7369, opened 2023-11-29
bunx: specify package with --package

telekid:
### What does this PR do?
This PR adds a `--package` flag to `bunx`. I've tried to keep both the syntax and semantics of `--package` as close to NPM's implementation as possible.

Relevant docs:
 - https://docs.npmjs.com/cli/v10/commands/npm-exec
 - https://docs.npmjs.com/cli/v10/commands/npx

**Note for reviewers**

In earlier versions of this code, no distinction was drawn between the package being run and the binary being invoked. The binary name was derived directly from the package. This PR separates those two concepts.

While reviewing, it's helpful to think of there being two separate modes of operation: implicit package and explicit package. In implicit package mode, the package name is assumed to be the same as the first command argument. In explicit mode, the package name is specified via the `--package` flag. 

Tests that are currently failing appear to be doing so because I moved argument parsing to within `bunx_command.zig`. This is causing `bun x` to capture flags like `--help` that were previously passed along directly to the invoked command. I believe this new behavior is actually more consistent with how `node-exec` behaves, with that caveat that I haven't been able to test the difference between `bunx` and `bun x` because I'm not quite sure how to make dev builds of the former yet. Note that this is a breaking change which is probably avoidable. I'm not sure what Bun's stance on breaking changes is. 

**TL;DR:**
`bun x cowsay --help` previously passed `--help` to `cowsay` command, but now passes `--help` to `bun` command. I need input from the bun team regarding the preferred approach here.

For more discussion, see [npx docs](https://docs.npmjs.com/cli/v10/commands/npx#npx-vs-npm-exec).

**TODO**

 - [ ] `.ReplCommand` relies on `.BunxCommand`. Should I be testing anything specific?
 - [ ] test `--bun`. (Currently it looks like the incorrect bin is used with this flag.)
 - [ ] test `bun x tsc main.ts`

**To test**
 - [x] `bun x cowsay hello # this is existing shortcut behavior` [here](https://github.com/oven-sh/bun/pull/7369/files#diff-8223dcf25255bd59c8022233507cdc78a6cef9a6d9d985f7ee3172ab469289f4R32)
 - [ ] `bun x -- cowsay hello`
 - [x] `bun x --package=cowsay -- cowsay hello` [here](https://github.com/oven-sh/bun/pull/7369/files#diff-8223dcf25255bd59c8022233507cdc78a6cef9a6d9d985f7ee3172ab469289f4R251)
 - [x] `bun x --package=cowsay -- cowthink hello` [here](https://github.com/oven-sh/bun/pull/7369/files#diff-8223dcf25255bd59c8022233507cdc78a6cef9a6d9d985f7ee3172ab469289f4R270)
 - [x] `bun x --package=cowsay -- cowthink --version` [here](https://github.com/oven-sh/bun/pull/7369/files#diff-8223dcf25255bd59c8022233507cdc78a6cef9a6d9d985f7ee3172ab469289f4R288)
 - [x] `bun x --package=cowsay hello` errors [here](https://github.com/oven-sh/bun/pull/7369/files#diff-8223dcf25255bd59c8022233507cdc78a6cef9a6d9d985f7ee3172ab469289f4R311)
 - [ ] Test that everything behaves as expected with `bunx` command, too

Closes https://github.com/oven-sh/bun/issues/7034.

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -182,11 +182,6 @@ pub const Arguments = struct {
     };
     pub const run_params = run_only_params ++ runtime_params_ ++ transpiler_params_ ++ base_params_;
 
-    const bunx_commands = [_]ParamType{
-        clap.parseParam("--silent                          Don't print the script command") catch unreachable,
-        clap.parseParam("-b, --bun                         Force a script or package to use Bun's runtime instead of Node.js (via symlinking node)") catch unreachable,
-    };
-
     const build_only_params = [_]ParamType{
         clap.parseParam("--compile                        Generate a standalone Bun executable containing your bundled code") catch unreachable,
         clap.parseParam("--watch                          Automatically restart the process on file change") catch unreachable,
@@ -1343,12 +1338,14 @@ pub const Command = struct {
                 if (comptime bun.fast_debug_build_mode and bun.fast_debug_build_cmd != .BunxCommand) unreachable;
                 const ctx = try Command.Context.create(allocator, log, .BunxCommand);
 
+                // TODO: This probably isn't right anymore
                 try BunxCommand.exec(ctx, bun.argv()[1..]);
                 return;
             },
             .ReplCommand => {
                 // TODO: Put this in native code.
                 if (comptime bun.fast_debug_build_mode and bun.fast_debug_build_cmd != .BunxCommand) unreachable;
+                // Why does .ReplCommand use .BunxCommand?
                 var ctx = try Command.Context.create(allocator, log, .BunxCommand);
                 ctx.debug.run_in_bun = true; // force the same version of bun used. fixes bun-debug for example
                 var args = bun.argv()[1..];
@@ -1894,12 +1891,14 @@ pub const Command = struct {
                         \\Execute an npm package executable (CLI), automatically installing into a global shared cache if not installed in node_modules.
                         \\
                         \\Flags:
-                        \\  <cyan>--bun<r>      Force the command to run with Bun instead of Node.js
+                        \\  <cyan>--bun<r>          Force the command to run with Bun instead of Node.js
+                        \\  <cyan>--package<r>      Specify package to install
                         \\
                         \\Examples<d>:<r>
                         \\  <b>bunx prisma migrate<r>
                         \\  <b>bunx prettier foo.js<r>
                         \\  <b>bunx<r> <cyan>--bun<r> <b>vite dev foo.js<r>
+                        \\  <b>bunx<r> <cyan>--package<r> <b>prettier prettier foo.js<r>
                         \\
                     , .{});
                 },
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1885,18 +1885,9 @@ pub const Command = struct {
 
                 Command.Tag.BunxCommand => {
                     Output.prettyErrorln(
-                        \\<b>Usage: bunx <r><cyan>[...flags]<r> \<package\><d>[@version] [...flags and arguments]<r>
                         \\Execute an npm package executable (CLI), automatically installing into a global shared cache if not installed in node_modules.
                         \\
-                        \\Flags:
-                        \\  <cyan>--bun<r>          Force the command to run with Bun instead of Node.js
-                        \\  <cyan>--package<r>      Specify package to install
-                        \\
-                        \\Examples<d>:<r>
-                        \\  <b>bunx prisma migrate<r>
-                        \\  <b>bunx prettier foo.js<r>
-                        \\  <b>bunx<r> <cyan>--bun<r> <b>vite dev foo.js<r>
-                        \\  <b>bunx<r> <cyan>--package<r> <b>prettier prettier foo.js<r>
+                        \\Run <cyan>bunx --help<r> for additional usage information.
                         \\
                     , .{});
                 },
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1338,14 +1338,12 @@ pub const Command = struct {
                 if (comptime bun.fast_debug_build_mode and bun.fast_debug_build_cmd != .BunxCommand) unreachable;
                 const ctx = try Command.Context.create(allocator, log, .BunxCommand);
 
-                // TODO: This probably isn't right anymore
                 try BunxCommand.exec(ctx, bun.argv()[1..]);
                 return;
             },
             .ReplCommand => {
                 // TODO: Put this in native code.
                 if (comptime bun.fast_debug_build_mode and bun.fast_debug_build_cmd != .BunxCommand) unreachable;
-                // Why does .ReplCommand use .BunxCommand?
                 var ctx = try Command.Context.create(allocator, log, .BunxCommand);
                 ctx.debug.run_in_bun = true; // force the same version of bun used. fixes bun-debug for example
                 var args = bun.argv()[1..];
--- a/src/cli/bunx_command.zig
+++ b/src/cli/bunx_command.zig
@@ -226,7 +226,7 @@ pub const BunxCommand = struct {
                 \\
             ;
 
-            Output.pretty(text, .{ package_name_for_update_request[0], package_name_for_update_request[0] });
+            Output.prettyErrorln(text, .{ package_name_for_update_request[0], package_name_for_update_request[0] });
             Global.exit(1);
         }
 
--- a/src/cli/bunx_command.zig
+++ b/src/cli/bunx_command.zig
@@ -153,7 +153,7 @@ pub const BunxCommand = struct {
         const cli = try CommandLineArguments.parse(ctx.allocator);
 
         var requests_buf = bun.PackageManager.UpdateRequest.Array.init(0) catch unreachable;
-        var run_in_bun = ctx.debug.run_in_bun;
+        var run_in_bun = ctx.debug.run_in_bun or cli.run_in_bun;
 
         var passthrough_list = try std.ArrayList(string).initCapacity(ctx.allocator, argv.len);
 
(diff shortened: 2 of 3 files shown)


src/cli.zig
```diff
@@ -1343,12 +1338,14 @@ pub const Command = struct {
                 if (comptime bun.fast_debug_build_mode and bun.fast_debug_build_cmd != .BunxCommand) unreachable;
                 const ctx = try Command.Context.create(allocator, log, .BunxCommand);
 
+                // TODO: This probably isn't right anymore
```
telekid: rm

src/cli.zig
```diff
@@ -182,11 +182,6 @@ pub const Arguments = struct {
     };
     pub const run_params = run_only_params ++ runtime_params_ ++ transpiler_params_ ++ base_params_;
 
-    const bunx_commands = [_]ParamType{
```
telekid, line 185: This constant was never actually referenced anywhere, and in this PR I move `bunx` CLI parsing to the `bunx_command` file.

src/cli.zig
```diff
@@ -1894,12 +1891,14 @@ pub const Command = struct {
...
                         \\Flags:
-                        \\  <cyan>--bun<r>      Force the command to run with Bun instead of Node.js
+                        \\  <cyan>--bun<r>          Force the command to run with Bun instead of Node.js
+                        \\  <cyan>--package<r>      Specify package to install
                         \\
                         \\Examples<d>:<r>
                         \\  <b>bunx prisma migrate<r>
                         \\  <b>bunx prettier foo.js<r>
                         \\  <b>bunx<r> <cyan>--bun<r> <b>vite dev foo.js<r>
+                        \\  <b>bunx<r> <cyan>--package<r> <b>prettier prettier foo.js<r>
```
telekid: TODO: incorrect example; update

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,101 @@ pub const BunxCommand = struct {
...
+                    try passthrough_list.append(positional);
+                }
             }
         }
 
+        if (cli.package.len != 0 and bin_name[0].len == 0) {
+            const text =
+                \\ <r><red>error<r><d>:<r> --package specified without indicating command<r>
+                \\ Perhaps you meant the following?:
+                \\ bunx --package {s} <b>-- {s}<r>
```
telekid, line 225: Imperfect suggested action for two reasons:

1. It doesn't include arguments
2. The user may have run `bun x` rather than `bunx`

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,101 @@ pub const BunxCommand = struct {
...
 
                 if (positional.len == 0) continue;
 
-                if (positional[0] != '-') {
-                    if (!found_subcommand_name) {
-                        found_subcommand_name = true;
-                        if (positional.len == 1 and positional[0] == 'x')
-                            continue;
-                    }
+                if (positional.len == 2 and positional[0] == '-' and positional[1] == '-') {
```
telekid, line 200: This logic should be lifted into `clap`.

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,101 @@ pub const BunxCommand = struct {
...
+                    parsing_passthroughs = true;
+                    continue;
                 }
 
-                if (!run_in_bun and !found_subcommand_name) {
-                    if (strings.eqlComptime(positional, "--bun")) {
-                        run_in_bun = true;
-                        continue;
-                    }
+                if (!parsing_passthroughs) {
```
telekid, line 205: Note to self: is this redundant?

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,101 @@ pub const BunxCommand = struct {
...
                 }
 
-                try passthrough_list.append(positional);
+                if (captured_bin_name and parsing_passthroughs) {
+                    try passthrough_list.append(positional);
+                }
             }
         }
 
+        if (cli.package.len != 0 and bin_name[0].len == 0) {
```
telekid, line 221: Specifying `--package` without separating `bunx` flags from `bin` with `--` is ambiguous and confusing, so it seems better to forbid that for now.

src/cli/bunx_command.zig
```diff
@@ -470,4 +522,77 @@ pub const BunxCommand = struct {
...
+
+        // TODO: Finish text
+        const intro_text =
+            \\<b>bunx<r>
+            \\<b>Usage<r>: <b><green>bun install<r> <cyan>[flags]<r> [...\<pkg\>]
+            \\<b>Alias: <b>bun i<r>
+            \\  Install the dependencies listed in package.json
+        ;
+
+        const outro_text =
```
telekid, line 525: TODO: Write this text

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
 
     pub fn exec(ctx_: bun.CLI.Command.Context, argv: [][*:0]const u8) !void {
         var ctx = ctx_;
+
+        const cli = try CommandLineArguments.parse(ctx.allocator);
+
         var requests_buf = bun.PackageManager.UpdateRequest.Array.init(0) catch unreachable;
-        var run_in_bun = ctx.debug.run_in_bun;
+        var run_in_bun = ctx.debug.run_in_bun or cli.run_in_bun;
```
telekid, line 156: Is this logic correct?

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
...
         var passthrough_list = try std.ArrayList(string).initCapacity(ctx.allocator, argv.len);
+
+        var bin_name = [1]string{""};
+
         var package_name_for_update_request = [1]string{""};
-        {
-            var found_subcommand_name = false;
+
+        if (cli.package.len == 0) {
+            {
```
telekid, line 165: self: new fn "parsePackageSpecifiedArgs"

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
...
+                        package_name_for_update_request[0] = positional;
+                        bin_name[0] = positional;
+                        continue;
+                    }
+
+                    try passthrough_list.append(positional);
+                }
+            }
+        } else {
+            package_name_for_update_request[0] = cli.package;
```
telekid, line 191: self: new fn "parsePackageUnspecifiedArgs"

src/cli/bunx_command.zig
```diff
@@ -220,15 +265,6 @@ pub const BunxCommand = struct {
...
 
-        const initial_bin_name = if (strings.eqlComptime(update_request.name, "typescript"))
-            "tsc"
-        else if (update_request.version.tag == .github)
-            update_request.version.value.github.repo.slice(update_request.version_buf)
-        else if (strings.lastIndexOfChar(update_request.name, '/')) |index|
-            update_request.name[index + 1 ..]
-        else
-            update_request.name;
-
```
telekid, line 231: Reviewer: I understand the `tsc` special case and believe I have accounted for that, but I don't understand the `(update_request.version.tag == .github)` and `(strings.lastIndexOfChar(update_request.name, '/'))`. 

Specifically, it isn't clear to me why the `bin` name (not that `package` name) would ever be other than the value contained with `argv`.

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
...
 
         var passthrough_list = try std.ArrayList(string).initCapacity(ctx.allocator, argv.len);
+
+        var bin_name = [1]string{""};
+
         var package_name_for_update_request = [1]string{""};
-        {
-            var found_subcommand_name = false;
+
+        if (cli.package.len == 0) {
```
telekid, line 164: The logic in this conditional is confusing. It should really be broken up into something closer to this:

```
if explict_package_mode:
  package_name, bin_name, passthrough_args = parse_argv_explicitly(...)
else
  package_name, bin_name, passthrough_args = parse_argv_implicitly(...)
```

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
...
                 }
 
-                try passthrough_list.append(positional);
+                if (captured_bin_name and parsing_passthroughs) {
+                    try passthrough_list.append(positional);
+                }
             }
         }
 
+        if (cli.package.len != 0 and bin_name[0].len == 0) {
```
telekid, line 221: TODO: Add a test for this exit

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
...
+            const text =
+                \\ <r><red>error<r><d>:<r> --package specified without indicating command<r>
+                \\ Perhaps you meant the following?:
+                \\ bunx --package {s} <b>-- {s}<r>
+                \\
+            ;
+
+            Output.pretty(text, .{ package_name_for_update_request[0], package_name_for_update_request[0] });
+            Global.exit(1);
```
telekid, line 230: Is exit(1) appropriate here?

src/cli/bunx_command.zig
```diff
@@ -148,43 +149,87 @@ pub const BunxCommand = struct {
...
+
+        const cli = try CommandLineArguments.parse(ctx.allocator);
+
         var requests_buf = bun.PackageManager.UpdateRequest.Array.init(0) catch unreachable;
-        var run_in_bun = ctx.debug.run_in_bun;
+        var run_in_bun = ctx.debug.run_in_bun or cli.run_in_bun;
 
         var passthrough_list = try std.ArrayList(string).initCapacity(ctx.allocator, argv.len);
+
+        var bin_name = [1]string{""};
```
telekid, line 160: Is this okay? Seems weird but the strategy was already in use for `package_name_for_update_request`.

PR #7372, opened 2023-11-29
Update file-io.md: BunFile.stream() doesn't return promise

NReilingh:
Based on testing, `.stream()` returns a `ReadableStream` directly instead of a `Promise`.

### What does this PR do?

Fixes the type signature reference for `BunFile` in the docs.

- [ ] Code changes


Electroid:
Thanks for fixing this, we missed it.

PR #7388, opened 2023-12-01
docs: update misleading `clonefile` information

### LINKED ISSUE #5973: Global Cache docs are harmful to macOS adoption due to incorrect statement about `clonefile` using more disk space

Documentation is incorrect

### What is the issue?

It is incorrect, and **_harmful to the adoption of Bun_**, to say that `This benefit does not extend to macOS, which uses clonefile for performance reasons.` in the global cache documentation.

On APFS drives cloning a file just points the new file at the blocks for the old file, so **no extra space is used on your drive**. Only when one of the two files change in some way are the 'dirty' blocks copied and the other blocks are still shared ([COW in computer storage](https://en.wikipedia.org/wiki/Copy-on-write#In_computer_storage)).

It is actively harmful to say the files are duplicated because some people see this as a blocking issue to using Bun, thinking that they'll end up wasting lots of space on their modules in their projects. The person in [this YT video](https://youtu.be/dQkv5C-Lfkw?si=tRUoiBs54Y63f87b&t=500) says he won't use Bun on macOS only because of this issue, and the video has over 114 thousand views as I write this.

First, the behavior should be verified to ensure that cloning the files is using APFS to just make clones. If that isn't the case, a new issue should be made to fix that. Bun should be using APFS clones, no exceptions.
Second, the documentation should be updated to say that on APFS drives (which is the vast majority, only old machines using hard drives _might_ be on HFS still) `clonefile` is used, but the files still don't take up more space, and _only_ if they are still on HFS (which again, is very unlikely today) will this mean they are taking up more space.

(I would do the verification step for you right now, but I won't have the time at the. moment unfortunately.)

This blog post has some great info about this:
https://eclecticlight.co/2020/04/14/copy-move-and-clone-files-in-apfs-a-primer/#:~:text=APFS%20clone%20files%20(%27copy%20on,same%20data%20as%20the%20original.

### Where did you find it?

The documentation:
https://bun.sh/docs/install/cache#:~:text=it%20uses%20clonefile%20.-,Saving%20disk%20space,disk%20space%20dedicated%20to%20node_modules%20.

Again, the highly viewed video that mentions this:
https://youtu.be/dQkv5C-Lfkw?si=tRUoiBs54Y63f87b&t=500

ImBIOS:
### What does this PR do?

Close #5973 

- [ ] Code changes

_null_


paperclover:
im not 100% sure about this because while clonefile files do not take more space on the disk, they do contribute to the total disk space used (as in, the file is counted twice)

paperclover:
i think instead of talking about HFS (filesystem that most do not use anymore, and im unsure if Bun even works on these older mac os systems), we should say that `clonefile` is copy-on-write, meaning it will not ocupy disk space, but it will count towards your drive's limit.

we can also say the copy-on-write behavior is useful because if something attempts to patch `node_modules/*`, it is impossible to affect other installations. but this is less important than clarifying the disk space issue.

ImBIOS:
@paperdave updated and ready to review again.

PR #7391, opened 2023-12-01
Fixes #4876

### LINKED ISSUE #4876: res.json() throw JSON Parse error

1.0.0

Darwin 21.6.0 x86_64 i386

run below code with bun will throw `SyntaxError: JSON Parse error: Unrecognized token ''`

```
const res = await fetch("https://www.1823.gov.hk/common/ical/en.json");
const data = await res.json();
console.log(data);
```

log the response json

### What do you see instead?

`SyntaxError: JSON Parse error: Unrecognized token ''`

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #4876 

There are tests

### DIFF
--- a/src/js/builtins/ReadableStreamInternals.ts
+++ b/src/js/builtins/ReadableStreamInternals.ts
@@ -1686,7 +1686,7 @@ export function readableStreamIntoArray(stream) {
   return processManyResult(manyResult);
 }
 
-function withoutUTF8BOM(result) {
+export function withoutUTF8BOM(result) {
   if (result.charCodeAt(0) === 0xfeff) {
     return result.slice(1);
   }
--- a/src/js/builtins/ReadableStreamInternals.ts
+++ b/src/js/builtins/ReadableStreamInternals.ts
@@ -1070,11 +1070,15 @@ export function createTextStream(highWaterMark) {
       }
 
       if (hasString && !hasBuffer) {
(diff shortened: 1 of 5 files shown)


src/js/builtins/ReadableStreamInternals.ts
```diff
@@ -1070,11 +1070,15 @@ export function createTextStream(highWaterMark) {
       }
 
       if (hasString && !hasBuffer) {
+        if (rope.charCodeAt(0) === 0xfeff) {
+          rope = rope.slice(1);
+        }
+
         return rope;
```
akx, line 1077: Should this just call `withoutUTF8BOM`?

src/js/builtins/ReadableStreamInternals.ts
```diff
@@ -1089,12 +1093,16 @@ export function createTextStream(highWaterMark) {
       }
       array.length = 0;
       if (rope.length > 0) {
+        if (rope.charCodeAt(0) === 0xfeff) {
+          rope = rope.slice(1);
+        }
```
akx, line 1098: Ditto, could also be a `withoutUTF8BOM` call?

src/js/builtins/ReadableStreamInternals.ts
TheDirigible: But then the source would be readable.  This fix doesn't even have a comment explaining what it's doing.

Issue #7396, opened 2023-12-01
Illegal instruction (core dumped)


skyfallwastaken:
### What version of Bun is running?

v1.0.14 - I cannot run the command as that also crashes

### What platform is your computer?

Linux 6.5.6-76060506-generic x86_64 x86_64

### What steps can reproduce the bug?

Run any Bun command

### What is the expected behavior?

Bun to fully function

### What do you see instead?

Illegal instruction (core dumped)

### Additional information

- CPU: i7-2600
- Installed via pkgx

skyfallwastaken:
Update: This is an issue w/ how pkgx packages Bun - sorry!

Electroid:
What was the issue and how did you fix it? We want to make sure others don’t run into this.

skyfallwastaken:
@Electroid The issue's above, and what fixed it was installing Bun using the official install script instead of using pkgx.

PR #7399, opened 2023-12-01
Added the command to run bun using pkgx installation.md

michaelessiet:
### What does this PR do?

This PR adds the command to run and install bun via [pkgx](https://pkgx.sh) to the installation docs.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -31,6 +31,10 @@ $ docker pull oven/bun
 $ docker run --rm --init --ulimit memlock=-1:-1 oven/bun
 ```
 
+```bash#pkgx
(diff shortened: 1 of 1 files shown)


Electroid:
It seems there are issues when using pkgx, so I don't think we will recommend it at this time. Also, we want to avoid information overload by having too many install methods.

https://github.com/oven-sh/bun/issues/7396#issuecomment-1836723213

skyfallwastaken:
@Electroid If you guys would fix the pkgx install process for Bun, that'd be great!

Bun's install process is defined here: https://github.com/pkgxdev/pantry/blob/main/projects/bun.sh/package.yml

PR #7402, opened 2023-12-01
Fix `Location` header not being cloned from `Response.redirect()`

### LINKED ISSUE #7397: Memory-stomping in Response.redirect() implementation??

1.0.14+d8be3e51b

Darwin 23.1.0 arm64 arm

Run the following code, note the output of the "Location" header.

```
let test = new URL('http://example1.com');
test.hostname = 'example2.com';
console.log(Response.redirect(test.href));
```

I would expect to see a sane Response object.

### What do you see instead?

Note the ```"location": "°\u001F%Rß\u0004\u0000\u0000xample2.com/"``` some memory is getting stomped.

```
Response (0 KB) {
  ok: false,
  url: "",
  status: 302,
  statusText: "",
  headers: Headers {
    "location": "°\u001F%Rß\u0004\u0000\u0000xample2.com/"
  },
  redirected: false,
  bodyUsed: false
}
```

### Additional information

This could be pretty serious if redirects are constructed in response to user input.

Electroid:
### What does this PR do?

Fixes #7397

The value from `Response.redirect` was being cloned, but then `deinit`'d. The problem was  the slice was also added to the headers, which meant the header value was garbage.

- [x] Code changes

I wrote a regression test.

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -1456,7 +1456,7 @@ void WebCore__FetchHeaders__put_(WebCore__FetchHeaders* headers, const ZigString
 {
     auto throwScope = DECLARE_THROW_SCOPE(global->vm());
     WebCore::propagateException(*global, throwScope,
-        headers->set(Zig::toString(*arg1), Zig::toString(*arg2)));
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
I think the code appending it is wrong, not the .deinit(). This would leak memory if it was a UTF-16 string converted into UTF/8

PR #7405, opened 2023-12-02
Implement Base64 encoding in fs.readFile and fs.readFileSync

### LINKED ISSUE #4845: fs.readFileSync with base64 encoding incompatible with Node.js

Create file `b64.js` with contents
```
console.log(require('fs').readFileSync('b64.js', 'base64'))
```
then run `bun b64.js` and `node b64.js`

Bun should write out the same base64 string as Node.js:
```
Y29uc29sZS5sb2cocmVxdWlyZSgnZnMnKS5yZWFkRmlsZVN5bmMoJ2I2NC5qcycsICdiYXNlNjQnKSkK
```

### What do you see instead?

Bun writes out the raw contents of the file, without base64-encoding them:
```
console.log(require('fs').readFileSync('b64.js', 'base64'))
```

### Additional information

This breaks `require('pdfmake')`

replydev:
### What does this PR do?

This PR implements base64 encoding for `fs.readFile` and `fs.readFileSync` API.
Example usage:
```
const base64Encoded = fs.readFileSync("filename.txt", "base64");
```
Should fix #4845 and a [bug](https://github.com/bpampuch/pdfmake/issues/2648) with [pdfmake](https://github.com/bpampuch/pdfmake) library.

1. I have tried steps from this issue: #4845 and after the changes it works.
2. I have ran the compiled artifact in my personal project which uses `pdfmake` library and now the problem is solved.

I wrote automated tests

Zig files changed:


src/bun.js/node/node_fs.zig
```diff
@@ -5055,12 +5056,32 @@ pub const NodeFS = struct {
...
+            }
+        }
+
         return switch (args.encoding) {
             .buffer => .{
                 .result = .{
                     .buffer = Buffer.fromBytes(buf.items, bun.default_allocator, .Uint8Array),
                 },
             },
+            .base64 => .{
```
paperclover: instead of doing this check three times, let's do it once, inside of the `.base64 =>` block. in zig you can use the pattern

```
.base64 => buffer: {
  const something = ...
  
  break :buffer something;
}
```
this will avoid use of the `undefined` keyword, as well as checking .encoding three times

src/bun.js/node/node_fs.zig
```diff
@@ -5055,12 +5056,32 @@ pub const NodeFS = struct {
...
+        // Handle Base64 encoding. Should fix:
+        // https://github.com/oven-sh/bun/issues/4845
+        // https://github.com/bpampuch/pdfmake/issues/2648
+        var base64_buffer = if (args.encoding == .base64)
+            bun.default_allocator.alloc(u8, Base64.encodeLen(buf.items)) catch @panic("oom")
+        else
+            undefined;
+
+        if (args.encoding == .base64) {
+            const result = Base64.encode(base64_buffer, buf.items);
```
paperclover: does `buf.items` get freed?

paperclover (changes requested):
see review comments.

also this doenst have to be done for this pr, but what about `writeFileSync` and all of the other encodings.

we should at least open an issue for these if it isnt fixed here.

also, jarred suggested that we should not allocate the entire buffer twice for large files, as this would mean using ~2x the file size of memory to load and decode the file.

liz3:
https://github.com/oven-sh/bun/pull/7174 🥴 Adds this for the write function.

src/bun.js/node/node_fs.zig
replydev: You are right, i'm a bit new with Zig syntax, i will improve :+1: .
I have committed a refactor.

src/bun.js/node/node_fs.zig
replydev: Oops, deallocated now.

replydev:
@paperdave I've committed some refactors. Regarding write functions, as himself said, have been implemented by @liz3 (whose I got some inspirations).

I'll try to find a new way to prevent double allocation before base64 conversion. Any thoughts welcome.

replydev:
@paperdave Hello again. I pushed a draft with a new approach on memory allocation. All the tests passes. Is that good for you?

src/bun.js/node/node_fs.zig
```diff
@@ -5094,6 +5084,38 @@ pub const NodeFS = struct {
...
+        const encoded_chunk_size = comptime Base64.encodeLenFromSize(chunk_size);
+        var offset: usize = 0;
+
+        // Declare buffer to store encoded Base64 Chunk over every chunked input
+        const temp_encoding_buffer = bun.default_allocator.alloc(u8, encoded_chunk_size) catch |err| return err;
+        defer bun.default_allocator.free(temp_encoding_buffer);
+
+        while (true) {
+            const remaining_bytes = buf.items.len - offset;
+            const chunk_to_be_encoded = if (remaining_bytes <= chunk_size) {
```
paperclover: the `if` expression here is confusing at first to read because one branch of the if is `noreturn` while the other is `[]const u8`. can you rewrite this to just the if statement with the early return followed by `const chunk_to_be_encoded = buf.items[offset..(offset + chunk_size)];`

src/bun.js/node/node_fs.zig
```diff
@@ -5094,6 +5084,38 @@ pub const NodeFS = struct {
...
+    /// Memory is allocated as needed and not at once time, preventing double allocations for input and output.
+    /// Returns an owned slice of the converted input.
+    fn base64Conversion(buf: *std.ArrayList(u8)) anyerror![]u8 {
+        // Chunk size should be a multiple of 3 bytes to be able to split base64 calculation
+        const chunk_size = 16383;
+        const encoded_chunk_size = comptime Base64.encodeLenFromSize(chunk_size);
+        var offset: usize = 0;
+
+        // Declare buffer to store encoded Base64 Chunk over every chunked input
+        const temp_encoding_buffer = bun.default_allocator.alloc(u8, encoded_chunk_size) catch |err| return err;
```
paperclover: i think we can get away with a single stack allocated buffer to reduce allocations

```
var encode_buf: [chunk_size]u8 = undefined;
```

src/bun.js/node/node_fs.zig
```diff
@@ -5094,6 +5084,38 @@ pub const NodeFS = struct {
...
+        defer bun.default_allocator.free(temp_encoding_buffer);
+
+        while (true) {
+            const remaining_bytes = buf.items.len - offset;
+            const chunk_to_be_encoded = if (remaining_bytes <= chunk_size) {
+                const final_buffer = bun.default_allocator.alloc(u8, Base64.encodeLenFromSize(remaining_bytes)) catch |err| return err;
```
paperclover: instead of this second allocation, reuse the existing buffer

src/bun.js/node/node_fs.zig
```diff
@@ -5094,6 +5084,38 @@ pub const NodeFS = struct {
...
+                defer bun.default_allocator.free(final_buffer);
+
+                _ = Base64.encode(final_buffer, buf.items[offset..]);
```
paperclover: technically .encode() returns the slice of the given buffer which it encoded to, i think in your case it is always going to be the same length as the input, but it would be better to not use a discard and:

```
const encoded = Base64.encode(&temp_encoding_buffer, buf.items[offset..])
```

src/bun.js/node/node_fs.zig
```diff
@@ -5094,6 +5084,38 @@ pub const NodeFS = struct {
...
+                buf.replaceRange(offset, remaining_bytes, final_buffer) catch |err| return err;
+
+                return buf.toOwnedSlice() catch |err| return err;
+            } else buf.items[offset..(offset + chunk_size)];
+
+            _ = Base64.encode(temp_encoding_buffer, chunk_to_be_encoded);
```
paperclover: same here

paperclover (changes requested):
this is much much better. i left some review comments on how heap allocations can be unused. also, the pattern of `x catch |err| return err| can be rewritten as `try x`

src/bun.js/node/node_fs.zig
replydev: As far as i can see, the encode method return the length of the encoded buffer, which is not the same as the input one (for 3 byte input we get 4 byte output).

So i think we can ignore the output value as the encoded result is set in the passed slice.

src/bun.js/node/node_fs.zig
replydev: Same as previous reply

src/bun.js/node/node_fs.zig
replydev: Totally agree. Sorry but i'm not so proficient in Zig Lang yet.

src/bun.js/node/node_fs.zig
replydev: Agree, extracted the if statement

src/bun.js/node/node_fs.zig
replydev: Done :+1:

replydev:
Hello @paperdave, any new feedback on this? Just two little points to discuss.

paperclover (approved):

replydev:
@paperdave Could you please approve the workflow run? I checked the latest failure but it seems they are not related to my implementations. What do you think about? Thanks

replydev:
@paperdave It seems that this feature has already been merged in this [commit](https://github.com/oven-sh/bun/commit/a119e8d63604be667c4184c33a7bbca1fa299612#diff-17148eb18dfeb9b3be6300551b847a139ca10a79193038a77fe2e1b7cef3e2eaR948) by the main developer. Feel free to close this if it is not useful anymore.

Issue #7302, opened 2023-11-24
Npm package aliases do not write to yarn lock file correctly with --yarn


yharaskrik:
### What version of Bun is running?

1.0.14

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

In the `package.json` of a repo add a dependency of:

`"othertypescript": "npm:typescript@4.5.5"` (version and package don't really matter)

Run `bun install --yarn`

In the generated lockfile the section for the `typescript` package will look like:

```
"typescript@^3 || ^4", typescript@^4.4.4, typescript@npm:typescript@4.5.5:
```

The alias part (the last in the csv list) should be wrapped in quotes

```
"typescript@^3 || ^4", typescript@^4.4.4, "typescript@npm:typescript@4.5.5":
```

### What is the expected behavior?

The yarn lock file should be printed correctly

### What do you see instead?

The lockfile is printed with incorrect formatting in the case of npm aliases

yharaskrik:
I have a fix for this, just figuring out how to write tests

ArnaudBarre:
@yharaskrik This should be marked as fixed?

yharaskrik:
> @yharaskrik This should be marked as fixed?

Yup!

PR #7409, opened 2023-12-02
fix(lockfile): print yarn lockfile aliases deps with quotes

yharaskrik:
### What does this PR do?

Fixes yarn lockfile printing to wrap aliases dependencies in quotes

https://github.com/oven-sh/bun/issues/7302

- [x] Code changes

Just need to get bun building so I can test locally

### DIFF
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -1230,7 +1230,7 @@ pub const Printer = struct {
                             needs_comma = false;
                         }
                         const version_name = dependency_version.literal.slice(string_buf);
-                        const needs_quote = always_needs_quote or std.mem.indexOfAny(u8, version_name, " |\t-/!") != null or std.mem.startsWith(u8, version_name, "npm:");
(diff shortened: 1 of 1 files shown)


src/install/lockfile.zig
```diff
@@ -1230,7 +1230,7 @@ pub const Printer = struct {
                             needs_comma = false;
                         }
                         const version_name = dependency_version.literal.slice(string_buf);
-                        const needs_quote = always_needs_quote or std.mem.indexOfAny(u8, version_name, " |\t-/!") != null;
+                        const needs_quote = always_needs_quote or std.mem.indexOfAny(u8, version_name, " |\t-/!") != null or std.mem.startsWith(u8, version_name, "npm:");
```
Jarred-Sumner:
```suggestion
                        const needs_quote = always_needs_quote or std.mem.indexOfAny(u8, version_name, " |\t-/!") != null or strings.hasPrefixComptime(version_name, "npm:");
```

Jarred-Sumner:
Thank you for this. Please add a test in a follow-up PR, but I'm going to go ahead and merge it

PR #7414, opened 2023-12-02
Handle stack overflow in binary expressions in JS Parser

Jarred-Sumner:
### What does this PR do?

This addresses the potential for stack overflow in long chains of binary expressions, mostly based on esbuild's implementation.

Both manual tests and an automated test

### DIFF
--- a/src/bun.js/RuntimeTranspilerCache.zig
+++ b/src/bun.js/RuntimeTranspilerCache.zig
@@ -7,6 +7,9 @@ const expected_version = 1;
 const debug = Output.scoped(.cache, false);
 const MINIMUM_CACHE_SIZE = 50 * 1024;
 
+// When making parser changes, it gets extremely confusing.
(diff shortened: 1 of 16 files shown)


src/bun.js/RuntimeTranspilerCache.zig
```diff
@@ -352,6 +371,10 @@ pub const RuntimeTranspilerCache = struct {
     fn reallyGetCacheDir(
         buf: *[bun.MAX_PATH_BYTES]u8,
     ) [:0]const u8 {
+        if (comptime bun.Environment.allow_assert) {
+            restore_from_cache_in_debug_build = bun.getenvZ("BUN_RESTORE_FROM_CACHE") != null;
```
dylan-conway: maybe `BUN_DEBUG_RESTORE_FROM_CACHE`?

PR #7431, opened 2023-12-03
Use `new.target` instead of `instanceof` to detect if its a constructor call

Jarred-Sumner:
### What does this PR do?

Replaces usages of 

```js
if (!(this instanceof MyClass)) return new MyClass(...args)
```
 with 
```js
if (!new.target) return new MyClass(...args)
``` 

It should be faster, though I have yet to benchmark it.

See also https://github.com/nodejs/node/pull/51026

Existing tests should cover it

### DIFF
--- a/src/bun.js/bindings/JSStringDecoder.cpp
+++ b/src/bun.js/bindings/JSStringDecoder.cpp
@@ -12,6 +12,7 @@
 #include <JavaScriptCore/FunctionPrototype.h>
 #include <JavaScriptCore/LazyClassStructure.h>
 #include <JavaScriptCore/LazyClassStructureInlines.h>
+#include <JavaScriptCore/VMTrapsInlines.h>
 
 namespace WebCore {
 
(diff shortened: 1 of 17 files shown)


paperclover:
stupid question but are you certain that all these classes have at least one test that calls the constructor without new?

paperclover:
in the future we need to not put multiple things into the same pr. i know that is hypocritical of me considering #6938 having like 50 changes, but still.

i'm skeptical of the streams changes and would honestly want that as a separate pr. it feels like something we're going to end up breaking and have to revert, but you have valuable changes to the syntax highlighter and napi.

paperclover (approved):
ive moved the changes not related to new.target to another pr.

changes to `node:stream` must be done carefully. after looking over it a few times it looks alright, but i still wonder if there is anything that can break with using `!new.target`

i think this should only be merged after we manually verify that Next.js, Express, Fastify, Webpack, and some other packages that may abuse stream mechanics to verify they still work.

paperclover (changes requested):
(see prev comment)

Issue #7028, opened 2023-11-10
Publishing lambda layer fails by "--zip-file must be a zip file with the fileb:// prefix" error


smt7174:
### What version of Bun is running?

1.0.8 or later(version 1.0.7 or earlier, this error does not occur)

### What platform is your computer?

Linux 5.4.72-microsoft-standard-WSL2 x86_64 x86_64 and Ubuntu 18

### What steps can reproduce the bug?

1. install bun by ```curl -fsSL https://bun.sh/install | bash``` command 
2. git clone this repository by ```git clone https://github.com/oven-sh/bun.git``` or ```git clone git@github.com:oven-sh/bun.git``` command
3. change current directory to bun/packages/bun-lambda by ```cd bun/packages/bun-lambda``` command
4. install packages by ```bun install``` command.
5. publish layer to your aws account by ```bun run publish-layer```

- 2 - 5 are tha same processes written at ```bun/packages/bun-lambda/README.md```
- aws cli v2 and profiles (```aws configure```) is set correctly. (version 1.0.7 or earlier, publishing lambda layer finishes correctly.)

### What is the expected behavior?

publishing lambda layer finishes correctly (with no errors) like below.

```
$ bun scripts/publish-layer.ts
Downloading... https://bun.sh/download/latest/linux/aarch64?avx2=true
Extracting...
Saving... bun-lambda-layer.zip
Saved
Publishing...
Published <your lambda layer ARN>
Done
```

### What do you see instead?

publishing lambda layer fails by "--zip-file must be a zip file with the fileb:// prefix" error. (error code: 252)

```
bun scripts/publish-layer.ts
Downloading... https://bun.sh/download/latest/linux/aarch64?avx2=true
Extracting...
Saving... bun-lambda-layer.zip
Saved
Publishing...
12 |     const result = stdout.toString("utf-8").trim();
13 |     if (status === 0) {
14 |       return result;
15 |     }
16 |     const reason = stderr.toString("utf-8").trim() || result;
17 |     throw new Error(`aws ${args.join(" ")} exited with ${status}: ${reason}`);
              ^
[...]

PR #7446, opened 2023-12-03
Fix for building lambda runtime zip files

asilvas:
### What does this PR do?

Resolves https://github.com/oven-sh/bun/issues/7028

Building the zip file required by the lambda runtime was broken and result in a corrupted file that could not be opened. This prevented the publish script from working. It's not clear what broke but it may have exposed a bug in bun.

- [ ] ~~Documentation or TypeScript types (it's okay to leave the rest blank in this case)~~
- [x] Code changes

Verified I could build and deploy new Bun-based lambda runtimes.


Electroid:
Thanks, going to merge this even though this is an issue with Bun. I'll keep that issue open so we can figure it out.

PR #7460, opened 2023-12-04
Fix parseArgs issues

otgerrogla:
### What does this PR do?

See https://github.com/oven-sh/bun/pull/7310#issuecomment-1837135393 and below

Fixes the following issues with the parseArgs implementation:
- Potential GC crashes due to some JSValue not being in the stack (specifically when large number of args is processed)
- Some memory leaks for strings in corner cases
- A panic when options were numeric strings like `-1` (this happened because calling `JSValue.put` with a key which is a numeric string triggers an assertion failed)
- Support for unicode args, imitating  the behavior from nodejs where utf8 chars encoded in 1 or 2 bytes are supported as short-options.

Rewrote some a large portion of the code. Originally I tried to replicate the phases in the NodeJS implementation by using temporary arrays. Now, I got rid of both the arraylist for the args processing queue, as well as the arraylist of tokens; both which could potentially exist in the heap. So tokens are now emitted and processed immediately without needed intermediate arraylists. The only remaining arraylist is the one for the parsed options received from the config.

Also, to avoid issues with JSValues and GC as well as to reduce the creation of unneeded JSValues, the Token struct fields now can be refs to either a JSValue that exists in the input (on either the `args` array or `options` entries), or a BunString. So no JSValues are allocated except when necessary to build the final output object, in which case they are put in js objects/arrays that reside on the stack.

Also reenables the parseArg function

- [X] Code changes

 I wrote automated tests for the latter two, and a stress test with parsing many args

- [x] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed __(now for sure! -_-)__

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -3030,14 +3030,15 @@ void JSC__JSValue__put(JSC__JSValue JSValue0, JSC__JSGlobalObject* arg1, const Z
 extern "C" void JSC__JSValue__putMayBeIndex(JSC__JSValue target, JSC__JSGlobalObject* globalObject, const BunString* key, JSC__JSValue value)
 {
     JSC::VM& vm = globalObject->vm();
+    ThrowScope scope = DECLARE_THROW_SCOPE(vm);
 
     WTF::String keyStr = key->toWTFString();
     JSC::Identifier identifier = JSC::Identifier::fromString(vm, keyStr);
 
     JSC::JSObject* object = JSC::JSValue::decode(target).asCell()->getObject();
     object->putDirectMayBeIndex(globalObject, JSC::PropertyName(identifier), JSC::JSValue::decode(value));
+    RETURN_IF_EXCEPTION(scope, void());
 }
-
 
 bool JSC__JSValue__isClass(JSC__JSValue JSValue0, JSC__JSGlobalObject* arg1)
 {
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -3027,6 +3027,18 @@ void JSC__JSValue__put(JSC__JSValue JSValue0, JSC__JSGlobalObject* arg1, const Z
     object->putDirect(arg1->vm(), Zig::toIdentifier(*arg2, arg1), JSC::JSValue::decode(JSValue3));
 }
 
+extern "C" void JSC__JSValue__putMayBeIndex(JSC__JSValue target, JSC__JSGlobalObject* globalObject, const BunString* key, JSC__JSValue value)
+{
+    JSC::VM& vm = globalObject->vm();
+
+    WTF::String keyStr = key->toWTFString();
+    JSC::Identifier identifier = JSC::Identifier::fromString(vm, keyStr);
+
+    JSC::JSObject* object = JSC::JSValue::decode(target).asCell()->getObject();
+    object->putDirectMayBeIndex(globalObject, JSC::PropertyName(identifier), JSC::JSValue::decode(value));
+}
+
+
 bool JSC__JSValue__isClass(JSC__JSValue JSValue0, JSC__JSGlobalObject* arg1)
 {
     JSValue value = JSValue::decode(JSValue0);
--- a/src/bun.js/node/util/parse_args.zig
+++ b/src/bun.js/node/util/parse_args.zig
@@ -31,68 +31,122 @@ const ParseArgsError = error{ParseError};
 /// Represents a slice of a JSValue array
 const ArgsSlice = struct {
     array: JSValue,
-    start: usize,
-    end: usize,
+    start: u32,
+    end: u32,
+
+    pub inline fn get(this: ArgsSlice, globalThis: *JSGlobalObject, i: i32) JSValue {
+        return this.array.getIndex(globalThis, this.start + @as(u32, @intCast(i)));
+    }
 };
 
-const TokenKind = enum { positional, option, @"option-terminator" };
+/// Helper ref to either a JSValue or a String,
+/// used in order to avoid creating unneeded JSValue as much as possible
+const ValueRef = union(Tag) {
+    jsvalue: JSValue,
+    bunstr: String,
+
+    const Tag = enum { jsvalue, bunstr };
+
+    pub fn asBunString(this: ValueRef, globalObject: *JSGlobalObject) bun.String {
+        return switch (this) {
+            .jsvalue => |str| str.toBunString(globalObject),
+            .bunstr => |str| return str,
+        };
+    }
+
+    pub fn asJSValue(this: ValueRef, globalObject: *JSGlobalObject) JSValue {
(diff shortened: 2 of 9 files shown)


src/string.zig
```diff
@@ -757,28 +770,12 @@ pub const String = extern struct {
...
-        }
-    }
-
-    pub fn indexOfCharU8(this: String, chr: u8) ?usize {
-        return indexOfChar(this, @intCast(chr));
+    pub fn indexOfAsciiChar(this: String, chr: u8) ?usize {
+        std.debug.assert(chr < 128);
+        return switch (this.isUTF16()) {
+            true => std.mem.indexOfScalar(u16, this.utf16(), @intCast(chr)),
+            false => std.mem.indexOfScalar(u8, this.byteSlice(), chr),
```
paperclover: for `u8` please use `bun.indexOfCharUsize`

src/string.zig
```diff
@@ -757,28 +770,12 @@ pub const String = extern struct {
...
-            else => return null,
-        }
-    }
-
-    pub fn indexOfCharU8(this: String, chr: u8) ?usize {
-        return indexOfChar(this, @intCast(chr));
+    pub fn indexOfAsciiChar(this: String, chr: u8) ?usize {
+        std.debug.assert(chr < 128);
+        return switch (this.isUTF16()) {
+            true => std.mem.indexOfScalar(u16, this.utf16(), @intCast(chr)),
```
paperclover, line 776: i like the change from truncate -> intCast :)

src/bun.js/node/util/parse_args.zig
```diff
@@ -31,68 +31,122 @@ const ParseArgsError = error{ParseError};
 /// Represents a slice of a JSValue array
 const ArgsSlice = struct {
     array: JSValue,
-    start: usize,
-    end: usize,
+    start: u32,
+    end: u32,
+
+    pub inline fn get(this: ArgsSlice, globalThis: *JSGlobalObject, i: i32) JSValue {
```
paperclover: why `i32` if it is only casted to `u32`. it seems usages can be changed to use unsigned as well

src/bun.js/node/util/parse_args.zig
```diff
@@ -623,50 +717,25 @@ pub fn parseArgsImpl(globalThis: *JSGlobalObject, config_obj: JSValue) !JSValue
...
-                    return error.ParseError;
-                }
-                result_positionals.putIndex(globalThis, result_positionals_len, token.value);
-                result_positionals_len += 1;
-            },
-            else => {},
-        }
-    }
+        // note that "values" needs to have a null prototype instead of Object, to avoid issues such as "values.toString"` being defined
+        .values = JSValue.createEmptyObjectWithNullPrototype(globalThis),
```
paperclover: whenever JS cells are allocated, it potentially may run the garbage collector. JSC attempts to scan the stack for JSValues but not certain if this will always work.

to verify there are no gc issues, can you add a test that runs parseArgs on a very large input ~1000 times, calling `Bun.gc(true)` between each run?

src/bun.js/node/util/parse_args.zig
```diff
@@ -461,99 +517,137 @@ fn tokenizeArgs(globalThis: *JSGlobalObject, args: ArgsSlice, options: []const O
...
+                var val = String.static(@as(string, @tagName(token_generic))).toJSConst(globalThis);
+                this.kinds_jsvalues[kind_idx] = val;
+                break :kindval val;
+            };
+
+            var obj = JSValue.createEmptyObject(globalThis, num_properties);
+            obj.put(globalThis, ZigString.static("kind"), kind_jsvalue);
+            switch (token_generic) {
+                .option => |token| {
+                    obj.put(globalThis, ZigString.static("index"), JSValue.jsNumberFromInt32(token.index));
```
paperclover: this is fine but we do have `jsNumber()` which works from any type

paperclover (changes requested):
it is very nice to see the introduction of `ValueRef` (i was really hoping that would've been done on the first one)

check my review comments for what should be changed

src/bun.js/node/util/parse_args.zig
otgerrogla: didn't know this, would be it better to call `protect()` on the three `result` sub-objects/arrays until just before returning it?

src/bun.js/node/util/parse_args.zig
otgerrogla: right, it was a leftover from before the new changes, but it indeed does not make any sense now

src/bun.js/node/util/parse_args.zig
paperclover: use the JSC.Strong wrapper struct (i might be wrong on the name)

again im not 100% sure if it is needed. doing a stress test on the function would be a good way to find out.

otgerrogla: I have run a test of 100 args parsed 1000 times, several times, and so far didn't crash or anything

paperclover:
would like to see the `std.mem.indexOfScalar` change otherwise i think this pr is good now.

otgerrogla:
ah yes sorry, did the change but forgot to stage the file

paperclover (approved):

src/bun.js/bindings/bindings.cpp
```diff
@@ -3027,6 +3027,18 @@ void JSC__JSValue__put(JSC__JSValue JSValue0, JSC__JSGlobalObject* arg1, const Z
...
 
+extern "C" void JSC__JSValue__putMayBeIndex(JSC__JSValue target, JSC__JSGlobalObject* globalObject, const BunString* key, JSC__JSValue value)
+{
+    JSC::VM& vm = globalObject->vm();
+
+    WTF::String keyStr = key->toWTFString();
+    JSC::Identifier identifier = JSC::Identifier::fromString(vm, keyStr);
+
+    JSC::JSObject* object = JSC::JSValue::decode(target).asCell()->getObject();
+    object->putDirectMayBeIndex(globalObject, JSC::PropertyName(identifier), JSC::JSValue::decode(value));
```
dylan-conway, line 3039: this could throw if the object is a proxy object. can you add exception checking?

dylan-conway (changes requested):
one requested change for `putMayBeIndex`. Looks good to me afterwards

src/bun.js/bindings/bindings.cpp
otgerrogla: You mean a `DECLARE_THROW_SCOPE` / `RETURN_IF_EXCEPTION`? Or something else?
Sorry, I'm not really sure how exception handling works

dylan-conway: Yep, `DECLARE_THROW_SCOPE` and `RETURN_IF_EXCEPTION` with the scope and the return value. You can find examples of this pattern in `bindings.cpp` if you search for `getIfPropertyExists`

dylan-conway (approved):

PR #7464, opened 2023-12-05
fix: jsx key after spread

### LINKED ISSUE #7328: Handling of key and spread in JSX not aligned with SWC and esbuild

1.0.14+d8be3e51b

Darwin 22.6.0 arm64 arm

Code: (a.tsx)
```tsx
const obj = {};
console.log(<div {...obj} key="after" />, <div key="before" {...obj} />);
```

Output closer to esbuild and swc, or at least not crashing on parse (discovered while running scanImports on a codebase).
Also the createElement fallback should probably use a named import so that it solves #7278

The issue is that the spec is not clear about that: https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#unresolved-questions

Did you find some other information that would instead shows that bun is doing the right thing?
Tagging @sebmarkbage if by chance we can have some official updates on deprecation of keys in JSX

### What do you see instead?

esbuild: `esbuild a.tsx`
```js
import { jsx } from "react/jsx-runtime";
import { createElement } from "react";
const obj = {};
console.log(/* @__PURE__ */ createElement("div", { ...obj, key: "after" }), /* @__PURE__ */ jsx("div", { ...obj }, "before"));
```
SWC: [playground](https://play.swc.rs/?version=1.3.94-nightly-20231016.1&code=H4sIAAAAAAAAA0vOzysuUchPylKwVaiuteZKBvLzc1L1cvLTNWxSMssUqvX09IDStQrZqZW2SolpJalFSgr6djoKYFmwYFJqWn5RqhJCqb6dpjUXAI9U7q1aAAAA&config=H4sIAAAAAAAAA1VQuw7CMAzc%2BYrKMwMwMPAPfIQVXAhqHrJdRIX67yRtmrZLEt%2Bd7xz%2FDk0DbzFwa37pmYqILMS1TogMXvGbENAhkhi2UeG4sCqZUu5phsb5AmX00gZ2G6sEM6HRHTSBLZO8dj4L1Xu1jnI69hocqjWwUzzoQ12IjrwWg8qO5VVnQn5SVgHJ5XQ5FyPoQpAc0WInJR6c9bYdtnswwcU0puyFWYr%2B2dX%2BdQvTAS48%2BoktC85LnEe4wipa0qozWLkvndOnxj%2F4fyl9rAEAAA%3D%3D)
```js
import { jsxDEV as _jsxDEV } from "react/jsx-dev-runtime";
import { createElement as _createElement } from "react";
const obj = {};
console.log(/*#__PURE__*/ _createElement("div", {
    ...obj,
    key: "after",
    __source: {
        fileName: "input.tsx",
        lineNumber: 2,
        columnNumber: 13
    },
    __self: this
}), /*#__PURE__*/ _jsxDEV("div", {
    ...obj
}, "before", false, {
    fileName: "input.tsx",
    lineNumber: 2,
    columnNumber: 43
}, this));
```
Babel: [playground](https://babeljs.io/repl#?browsers=&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=MYewdgzgLgBCBGArGBeGBvAvgbgFCkhABsBTAOiJAHMAKAHgBMBLANwzI4UUxgGsSAnigBEAQwBmUEgCdhMAPQA-ADQxGrPoJHwS4kNJJz0HMlx5KAlHiA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=react&prettier=false&targets=&version=7.23.4&externalPlugins=&assumptions=%7B%7D)
```js
import { createElement as _createElement } from "react";
import { jsx as _jsx } from "react/jsx-runtime";
const obj = {};
console.log( /*#__PURE__*/_createElement("div", {
  ...obj,
  key: "after"
}), /*#__PURE__*/_jsx("div", {
  ...obj
}, "before"));
```
tsc: `tsc --jsx react-jsx a.tsx`
```js
var jsx_runtime_1 = require("react/jsx-runtime");
var react_1 = require("react");
var obj = {};
console.log((0, react_1.createElement)("div", __assign({}, obj, { key: "after" })), (0, jsx_runtime_1.jsx)("div", __assign({}, obj), "before"));
```
bun: `bun build a.tsx -e react`
```js
// a.tsx
import {
jsxDEV
} from "react/jsx-dev-runtime";
var obj = {};
console.log(jsxDEV("div", {
  ...obj
}, "after", false, undefined, this), React.createElement("div", {
  key: "before",
  ...obj
}));
2 | } key="after" />, <div key="before" {...obj} />);
                                            ^
warn: "key" prop before a {...spread} is deprecated in JSX. Falling back to classic runtime.
```

### Additional information

_No response_

rhyzx:
### What does this PR do?

Fix #7328.
Following the implementation of [TypeScript](https://www.typescriptlang.org/play?target=99&jsx=4#code/MYewdgziA2CmB00QHMAUAeAJgSwG4AIBveEkAIwCsBffAa1gE8BeAIgEMAzAF1gCcX8AegB8AGnxY8dRqzKwOIXrAHFSlGiICUQA), if jsx with key comes after spread props, it should fallback to `createElement`.

### Example:
`bun build --outdir=dist --external=react example.js`
```js
console.log(<div {...obj} key="after" />, <div key="before" {...obj} />)
```

**Before**
`warn: "key" prop before a {...spread} is deprecated in JSX. Falling back to classic runtime.`
```js
import {
jsxDEV
} from "react/jsx-dev-runtime";
console.log(jsxDEV("div", {
  ...obj
}, "after", false, undefined, this), React.createElement("div", {
  key: "before",
  ...obj
}));
```

**After**
```js
import {
jsxDEV
} from "react/jsx-dev-runtime";
import {
createElement
} from "react";
console.log(createElement("div", {
  ...obj,
  key: "after"
}), jsxDEV("div", {
  ...obj
}, "before", false, undefined, this));
```

- [x] Code changes

Test included, [see](https://github.com/oven-sh/bun/pull/7464/files#diff-fa88bc0d245c46b2dbd67a0132a3646fdf248b6b9fd4beefbffcfc27ff56d74bR1259-R1264)

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -1078,6 +1078,11 @@ declare module "bun" {
      * it in some cases. Do your own benchmarks!
      */
     inline?: boolean;
+
+    /**
+     * @default "warn"
+     */
+    logLevel?: "verbose" | "debug" | "info" | "warn" | "error";
   }
 
   /**
--- a/src/bun.js/api/JSTranspiler.zig
+++ b/src/bun.js/api/JSTranspiler.zig
@@ -113,6 +113,7 @@ pub const TransformTask = struct {
             .loader = loader,
             .replace_exports = transpiler.transpiler_options.runtime.replace_exports,
         };
+        transform_task.log.level = transpiler.transpiler_options.log.level;
         transform_task.bundler = transpiler.bundler;
         transform_task.bundler.linker.resolver = &transform_task.bundler.resolver;
 
@@ -318,7 +319,6 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
         .transform = default_transform_options,
         .log = logger.Log.init(allocator),
     };
-    transpiler.log.level = .warn;
 
     if (!object.isObject()) {
         JSC.throwInvalidArguments("Expected an object", .{}, globalObject, exception);
@@ -723,6 +723,18 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
         transpiler.runtime.replace_exports = replacements;
     }
 
+    if (object.get(globalThis, "logLevel")) |logLevel| {
+        const logLevelString = logLevel.toBunString(globalThis);
+        defer logLevelString.deref();
+        const key = logLevelString.toOwnedSlice(allocator) catch bun.outOfMemory();
+        if (logger.Log.Level.Map.get(key)) |level| {
+            transpiler.log.level = level;
+        } else {
+            JSC.throwInvalidArguments("logLevel must be one of \"verbose\", \"debug\", \"info\", \"warn\", or \"error\"", .{}, globalObject, exception);
+            return transpiler;
+        }
+    }
+
     transpiler.tree_shaking = tree_shaking orelse false;
     transpiler.trim_unused_imports = trim_unused_imports orelse transpiler.tree_shaking;
 
@@ -1125,6 +1137,7 @@ pub fn transformSync(
     this.bundler.setAllocator(arena.allocator());
     this.bundler.macro_context = null;
     var log = logger.Log.init(arena.backingAllocator());
+    log.level = this.transpiler_options.log.level;
     this.bundler.setLog(&log);
 
     defer {
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -115,15 +115,13 @@ const JSXImport = enum {
         jsxs: ?LocRef = null,
         Fragment: ?LocRef = null,
         createElement: ?LocRef = null,
-        factory_name: []const u8 = "React.createElement",
-        fragment_name: []const u8 = "Fragment",
 
         pub fn get(this: *const Symbols, name: []const u8) ?Ref {
             if (strings.eqlComptime(name, "jsx")) return if (this.jsx) |jsx| jsx.ref.? else null;
             if (strings.eqlComptime(name, "jsxDEV")) return if (this.jsxDEV) |jsx| jsx.ref.? else null;
             if (strings.eqlComptime(name, "jsxs")) return if (this.jsxs) |jsxs| jsxs.ref.? else null;
-            if (strings.eql(name, this.fragment_name)) return if (this.Fragment) |Fragment| Fragment.ref.? else null;
-            if (strings.eql(name, this.factory_name)) return if (this.createElement) |createElement| createElement.ref.? else null;
+            if (strings.eqlComptime(name, "Fragment")) return if (this.Fragment) |Fragment| Fragment.ref.? else null;
+            if (strings.eqlComptime(name, "createElement")) return if (this.createElement) |createElement| createElement.ref.? else null;
             return null;
         }
 
@@ -157,11 +155,14 @@ const JSXImport = enum {
             }
 
             if (this.Fragment != null) {
-                buf[i] = this.fragment_name;
+                buf[i] = "Fragment";
                 i += 1;
             }
 
             return buf[0..i];
+        }
+        pub fn sourceImportNames(this: *const Symbols) []const string {
+            return if (this.createElement != null) &[_]string{"createElement"} else &[_]string{};
         }
     };
 };
@@ -4019,6 +4020,19 @@ pub const Parser = struct {
                     false,
                 ) catch unreachable;
             }
+
+            const source_import_names = p.jsx_imports.sourceImportNames();
+            if (source_import_names.len > 0) {
+                p.generateImportStmt(
+                    p.options.jsx.package_name,
+                    source_import_names,
+                    &before,
+                    &p.jsx_imports,
+                    null,
+                    "",
+                    false,
+                ) catch unreachable;
+            }
         }
 
         var parts_slice: []js_ast.Part = &([_]js_ast.Part{});
@@ -6363,14 +6377,6 @@ fn NewParser_(
             //  "Foo.Bar.createElement" becomes:
             //      import { Bar } from 'foo';
             //      Usages become Bar.createElement
-
-            if (p.options.jsx.runtime == .classic) {
-                if (p.options.jsx.fragment.len > 0)
-                    p.jsx_imports.fragment_name = p.options.jsx.fragment[if (p.options.jsx.fragment.len > 1) 1 else 0];
-            }
-
-            if (p.options.jsx.factory.len > 0)
-                p.jsx_imports.factory_name = p.options.jsx.factory[if (p.options.jsx.factory.len > 1) 1 else 0];
 
             switch (comptime jsx_transform_type) {
                 .react => {
@@ -14735,7 +14741,7 @@ fn NewParser_(
 
                 var spread_loc: logger.Loc = logger.Loc.Empty;
                 var props = ListManaged(G.Property).init(p.allocator);
-                var spread_prop_i: i32 = -1;
+                var first_spread_prop_i: i32 = -1;
                 var i: i32 = 0;
                 parse_attributes: while (true) {
                     switch (p.lexer.token) {
@@ -14785,7 +14791,7 @@ fn NewParser_(
                                     try p.lexer.next();
                                     can_be_inlined = false;
 
-                                    spread_prop_i = i;
+                                    if (first_spread_prop_i == -1) first_spread_prop_i = i;
                                     spread_loc = p.lexer.loc();
                                     try props.append(G.Property{ .value = try p.parseExpr(.comma), .kind = .spread });
                                 },
@@ -14850,13 +14856,13 @@ fn NewParser_(
                     }
                 }
 
-                const is_key_before_rest = key_prop_i > -1 and spread_prop_i > key_prop_i;
-                flags.setPresent(.is_key_before_rest, is_key_before_rest);
-                if (is_key_before_rest and p.options.jsx.runtime == .automatic and !p.has_classic_runtime_warned) {
-                    try p.log.addWarning(p.source, spread_loc, "\"key\" prop before a {...spread} is deprecated in JSX. Falling back to classic runtime.");
+                const is_key_after_spread = key_prop_i > -1 and first_spread_prop_i > -1 and key_prop_i > first_spread_prop_i;
+                flags.setPresent(.is_key_after_spread, is_key_after_spread);
+                properties = G.Property.List.fromList(props);
+                if (is_key_after_spread and p.options.jsx.runtime == .automatic and !p.has_classic_runtime_warned) {
+                    try p.log.addWarning(p.source, spread_loc, "\"key\" prop after a {...spread} is deprecated in JSX. Falling back to classic runtime.");
                     p.has_classic_runtime_warned = true;
                 }
-                properties = G.Property.List.fromList(props);
             }
 
             // People sometimes try to use the output of "JSON.stringify()" as a JSX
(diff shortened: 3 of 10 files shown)


Jarred-Sumner:
Thank you for this. This looks good - one question: if  the developer uses a jsx comment pragma (` // @jsx jsx`), what happens? Can you add a test for that case?

test/bundler/bun-build-api.test.ts
```diff
@@ -250,19 +250,6 @@ describe("Bun.build", () => {
     expect(x.logs[0].position).toEqual(null);
   });
 
-  test("warnings do not fail a build", async () => {
```
paperclover: why was this test removed? it will hide a potential regression if this warning or syntax is a failing build.

we should still print the warning because `import { createElement } from "react";` *is still* the classic JSX runtime and the warning points out a probably mistake in the user's code.

also, this warning seems to only be printed once per build

test/transpiler/transpiler.test.js
```diff
@@ -1251,6 +1251,17 @@ export default <>hi</>
...
+
+    // key after spread props
+    // https://github.com/oven-sh/bun/issues/7328
+    expect(bun.transformSync(`console.log(<div {...obj} key="after" />, <div key="before" {...obj} />);`)).toBe(
+      `console.log(createElement(\"div\", {\n  ...obj,\n  key: \"after\"\n}), jsxDEV(\"div\", {\n  ...obj\n}, \"before\", false, undefined, this));
+`,
+    );
+    expect(bun.transformSync(`console.log(<div {...obj} key="after" {...obj2} />);`)).toBe(
+      `console.log(createElement(\"div\", {\n  ...obj,\n  key: \"after\",\n  ...obj2\n}));
 `,
```
paperclover, line 1266: as jarred said, can we get a test with the `// @jsx` comment

paperclover (changes requested):

paperclover:
Also, can you increment the number in `RuntimeTranspilerCache.zig`

```diff
 // ** Update the version number when any breaking changes are made to the cache format or to the JS parser **
-const expected_version = 1;
+const expected_version = 2;
```

rhyzx:
> Thank you for this. This looks good - one question: if the developer uses a jsx comment pragma (` // @jsx jsx`), what happens? Can you add a test for that case?

It should not affect the results if jsx setting is `react-jsx`;
Test added

rhyzx:
> Also, can you increment the number in `RuntimeTranspilerCache.zig`
> 
> ```diff
>  // ** Update the version number when any breaking changes are made to the cache format or to the JS parser **
> -const expected_version = 1;
> +const expected_version = 2;
> ```

DONE

src/js_parser.zig
```diff
@@ -14850,12 +14856,8 @@ fn NewParser_(
                     }
                 }
 
-                const is_key_before_rest = key_prop_i > -1 and spread_prop_i > key_prop_i;
-                flags.setPresent(.is_key_before_rest, is_key_before_rest);
-                if (is_key_before_rest and p.options.jsx.runtime == .automatic and !p.has_classic_runtime_warned) {
```
paperclover, line 14855: i still want the revival of this flag. after that should be good to merge

paperclover: not flag, i meant to say "warning"

test/bundler/bun-build-api.test.ts
rhyzx: Put the key before/after spread props is a valid approach, and I suggest not warning for it(ts/esbuild/swc also do not warn it).

rhyzx: Additionally, this warning will break transpiling, eg.

```js
const transpiler = new Bun.Transpiler({ loader: "jsx" });
try {
  const result = await transpiler.transform(`console.log(<div {...obj} key="after" />, <div key="before" {...obj} />)`);
  console.log(result);
} catch (e) {
  console.log("err", e);
}
```

```
err 1 | console.log(<div {...obj} key="after" />, <div key="before" {...obj} />)
                                                                    ^
warn: "key" prop before a {...spread} is deprecated in JSX. Falling back to classic runtime.
   at input.jsx:1:65
```

paperclover: didnt see this. the warning is valuable, but `Bun.Transpiler` shouldnt throw on it. really any warning shouldnt throw. jarred suggested we make the default logLevel for transpiler be `err` instead of `warn`, but it seems this isnt too easily configurable.

easy fix would just be to make it never throw on warnings in JSTranspiler.zig

```diff
-if ((this.bundler.log.warnings + this.bundler.log.errors) > 0) {
+if (this.bundler.log.errors > 0) {
   globalThis.throwValue(this.bundler.log.toJS(globalThis, globalThis.allocator(), "Parse error"));
   return .zero;
 }
```

maybe this behavior is desired sometimes and we should allow passing a logLevel property to bun.Transpiler

rhyzx: I see, I will add the warning back.
Let me clarify my understanding to avoid confusion: We should issue a warning when the key appears **after** spread props, not **before**, is that correct?

paperclover: warn before, because {...obj} can override the key property, which is not allowed in react-jsx as the key is a separate argument from the props (check the bundled output and you will see)

rhyzx: Emmm… this is a bit confusing for me. You see, the `createElement` downgrade is for the key **after** spread props, but should we issue a warning for the non-downgraded key **before** spread props? What should the content of this warning message be?

rhyzx: BTW, `react-jsx` can handle key before spread perfectly, cause the key is in another param [see](https://www.typescriptlang.org/play?target=99&jsx=4#code/MYewdgziA2CmB00QHMAUAeAJgSwG4AIBveEkAIwCsBffAa1gE8BeAIgEMAzAF1gCcX8AegB8AGnxY8dRqzKwOIXrAHFSlGiICUQA)

paperclover: oh wait. i am completely wrong. it is swapped, and that is what the PR fixes. we should warn in the case that causes the classic runtime to be used.

paperclover: i double checked esbuild and it behaves how you have modified it.

so before = no warn, automatic jsx runtime
after = warn, fallback to classic

and this fixes the output to match tsc and esbuild

rhyzx: No confusion for me now, I've added the warning back.

paperclover (approved):
awesome

paperclover (changes requested):
image
aw, it still throws

rhyzx:
> image aw, it still throws

Yes, i leave the error level parts untouched. Should I change that as well?

rhyzx:
Give me a few minutes, I will submit the the error level parts.

paperclover (approved):

rhyzx:
Hold on. I've replaced all `this.bundler.log.warnings + this.bundler.log.errors) > 0` in `JSTranspiler.zig`, but it still throws. I need to investigate further.

rhyzx:
Had to add the configurable error level 😂 (otherwise a large number of test cases that catching warnings would fail.

All tests(transpiler/bundler) should pass now.

src/bun.js/api/JSTranspiler.zig
```diff
@@ -722,6 +722,10 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
         transpiler.runtime.replace_exports = replacements;
     }
 
+    if (object.get(globalThis, "throwOnWarning")) |throwOnWarning| {
+        transpiler.log.level = if (throwOnWarning.toBoolean()) .warn else .err;
```
paperclover: should we make this `logLevel: string`? @Jarred-Sumner @jhmaster2000  

(option to disable warning in Bun.Transpiler)

paperclover (changes requested):
can you change `throwOnWarning` to `logLevel`? it would take a string "error"/"warn"/"info"/"debug"/"verbose". parsing this would be using .toBunString() on the JSValue and then using `logger.Log.Level.Map.get()`

and then it defaults to `"warn"` for backwards compatibility.

rhyzx:
DONE
(not familiar with the manual memory management parts, hope I have done it right

src/bun.js/api/JSTranspiler.zig
```diff
@@ -723,6 +723,15 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
         transpiler.runtime.replace_exports = replacements;
     }
 
+    if (object.get(globalThis, "logLevel")) |logLevel| {
+        const logLevelString = logLevel.toBunString(globalThis);
+        defer logLevelString.deref();
+        const key = logLevelString.toOwnedSlice(allocator) catch unreachable;
```
paperclover: `catch unreachable` shouldnt be used on memory allocations because they are possible (i know this is an existing pattern and you just copied code, but i want to make sure we stop using this pattern). this will invoke undefined behavior, when really we should crash the build.

to ensure this happens, use `catch bun.outOfMemory()` here.

src/bun.js/api/JSTranspiler.zig
```diff
@@ -723,6 +723,15 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
...
+        if (logger.Log.Level.Map.get(key)) |level| {
+            transpiler.log.level = level;
+        }
```
paperclover, line 735: add
```zig
else {
  JSC.throwInvalidArguments("logLevel must be one of \"verbose\", \"debug\", \"info\", \"warn\", or \"error\"", .{}, globalObject, exception);
  return transpiler;
}
```

packages/bun-types/bun.d.ts
```diff
@@ -1078,6 +1078,11 @@ declare module "bun" {
      * it in some cases. Do your own benchmarks!
      */
     inline?: boolean;
+
+    /**
+     * @default "warn"
+     */
+    logLevel?: string;
```
paperclover:
```
logLevel?: "verbose" | "debug" | "info" | "warn" | "error";
```

paperclover (changes requested):
almost there, just some nitpicks

src/bun.js/api/JSTranspiler.zig
rhyzx: DONE

src/bun.js/api/JSTranspiler.zig
rhyzx: That makes sense, thank you for your suggestions.

src/bun.js/api/JSTranspiler.zig
```diff
@@ -723,6 +723,18 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
         transpiler.runtime.replace_exports = replacements;
     }
 
+    if (object.get(globalThis, "logLevel")) |logLevel| {
+        const logLevelString = logLevel.toBunString(globalThis);
+        defer logLevelString.deref();
+        const key = logLevelString.toOwnedSlice(allocator) catch bun.outOfMemory();
+        if (logger.Log.Level.Map.get(key)) |level| {
```
Jarred-Sumner, line 730: We don't need to clone the string here. I'm going to merge this and then fix that

PR #7466, opened 2023-12-05
test: Add tests for Web Intl functionality

ImBIOS:
### What does this PR do?

Adding Test

- [x] Code changes

I ran `bun test web-intl` and everything is green.


Jarred-Sumner:
Thank you for this 

This is caused by us using a different version of libicu on Linux versus macOS. We could use the same version. It would at least be consistent across platforms for the same version of Bun. But it would mean that the version is always about 2 years out of date, since on macOS we must use the system-provided version (or else Bun will become 90 MB) and that means targeting the lowest available version. It is a reasonable thing to ask, but note that SSR hydration warnings would continue to happen between different browser versions

ImBIOS:
@Jarred-Sumner need help to do chores?

ImBIOS:
@Jarred-Sumner if this is a good idea, feel free to review or merge it, if not, it's ok to close it :)

PR #7475, opened 2023-12-05
perf: make napi_create_object 2.5x faster

paperclover:
### What does this PR do?

Cherry-picked unrelated changes from #7431 

https://twitter.com/jarredsumner/status/1731836157606940851

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -59,6 +59,26 @@
 using namespace JSC;
 using namespace Zig;
 
+#define NAPI_VERBOSE 0
+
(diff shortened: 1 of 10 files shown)


src/bun.js/bindings/napi.cpp
```diff
@@ -59,6 +59,26 @@
 using namespace JSC;
 using namespace Zig;
 
+#define NAPI_VERBOSE 0
```
paperclover, line 62: why is this always set to 0 and always disabled on windows?

paperclover:
Cannot be merged, string_decoder no longer works after the proposed changes.

PR #7476, opened 2023-12-05
increase default thread stack size

dylan-conway:
### What does this PR do?

- [ ] Code changes


paperclover (approved):
lol want to add a test that runs `bun build` on the prettier typescript plugin that fails. i think we have it installed in the root node_modules

Issue #7428, opened 2023-12-03
Node 100x faster writing to AWS S3 (PUT's) than Bun v1.0.15 (in same region)


asilvas:
### What version of Bun is running?

1.0.15+b3bdf22eb

### What platform is your computer?

Linux 6.1.0-13-cloud-amd64 x86_64 unknown

### What steps can reproduce the bug?

```js
import { S3Client, PutObjectCommand, GetObjectCommand } from '@aws-sdk/client-s3';

const BUCKETS = ['myS3ExpressBucket', 'myS3StandardBucket'];
const CONCURRENCY = 10;
const TIME_PER_TEST_MS = 10_000;

const s3 = new S3Client({
  region: process.env.AWS_REGION || 'us-east-1',
});

for (const Bucket of BUCKETS) {
  await testBucket(Bucket);
}

async function testBucket(Bucket) {
  console.log(`Write testing ${Bucket}...`);
  const writes = await testWrites(Bucket);
  console.log(`Read testing ${Bucket}...`);
  const reads = await testReads(Bucket);

  console.log(`Write rate: ${writes.rate.toFixed(1)} ops/sec, latency: ${writes.latency.toFixed(1)} ms/op`);
  console.log(`Read rate: ${reads.rate.toFixed(1)} ops/sec, latency: ${reads.latency.toFixed(1)} ms/op`);
}

async function testWrites(Bucket) {
  const start = Date.now();
  let count = 0;
  while (Date.now() - start < TIME_PER_TEST_MS) {
    await Promise.all(Array.from({ length: CONCURRENCY }, async (_, i) => {
      await s3.send(new PutObjectCommand({
        Bucket,
        Key: `rand/test-${i}.txt`,
        Body: 'Hello World!',
      }));
    }));
    count += CONCURRENCY;
  }

  const elapsed = Date.now() - start;
  const rate = count / (elapsed / 1000);
  const latency = (elapsed / count) * 1000;

  return { count, rate, latency };
}

async function testReads(Bucket) {
  const start = Date.now();
  let count = 0;
  while (Date.now() - start < TIME_PER_TEST_MS) {
    await Promise.all(Array.from({ length: CONCURRENCY }, async (_, i) => {
      await s3.send(new GetObjectCommand({
        Bucket,
        Key: `rand/test-${i}.txt`,
      }));
    }));
    count += CONCURRENCY;
  }

  const elapsed = Date.now() - start;
  const rate = count / (elapsed / 1000);
[...]

PR #7480, opened 2023-12-06
fix(fetch) continue event

cirospaciari:
### What does this PR do?

Fix: #7428

- [x] Code changes

Manual tested using s3 and also added a test

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -1027,6 +1027,11 @@ class OutgoingMessage extends Writable {
   }
 }
 
+function emitContinueNT(self) {
+  if (!self._closed) {
+    self.emit("continue");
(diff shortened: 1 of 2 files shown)


asilvas:
Was hoping this was going to make 1.0.16 release since this might resolve the AWS SDK client issues. What's the blocker?

Jarred-Sumner:
The blocker is that this approach doesn’t actually detect the continue event. It fakes it. It will emit the continue event even when there was no continue sent. 

The change needs to be done in the http client to emit a progress event when it receives continue and then propane that through to JS

cirospaciari:
To resolve this we need duplex streaming on fetch first, so closing in favor of a true solution.

PR #7483, opened 2023-12-06
feat(test): run any test file by passing it as a path

paperclover:
### What does this PR do?

https://fxtwitter.com/thepaperdave/status/1732249590886551705 

Also absolute paths.

I wrote tests for my tests


Jarred-Sumner:
thank you

Issue #4415, opened 2023-08-30
http.ServerResponse: Cannot call a class constructor without |new|


sirenkovladd:
### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Linux 6.2.0-26-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
const inject = require('light-my-request')

const dispatch = function (req, res) {
  const reply = 'Hello World'
  res.writeHead(200, { 'Content-Type': 'text/plain', 'Content-Length': reply.length })
  res.end(reply)
}

inject(dispatch, { method: 'get', url: '/' }, (err, res) => {
  console.log(res.payload)
});
```

short repro
```js
const http = require('http')
function Response(req) {
  http.ServerResponse.call(this, {});
}
console.log(new Response().constructor.name)
```

### What is the expected behavior?

```js
❯ node bun-http.ServerResponse.js
Hello World
```

less code
```js
❯ node bun-http.ServerResponse.js
Response
```

### What do you see instead?

```sh
❯ bun bun-http.ServerResponse.js 
621 | }
622 | var OriginalWriteHeadFn, OriginalImplicitHeadFn;
623 | 
624 | class ServerResponse extends Writable {
625 | 
626 |   constructor({ req, reply }) {
                              ^
TypeError: Cannot call a class constructor without |new|
      at ServerResponse (node:http:626:26)
      at new Response (<path>/bun-http.ServerResponse.js:15:2)
      at <path>/bun-http.ServerResponse.js:17:12
      at globalThis (<path>/bun-http.ServerResponse.js:17:43)
```

### Additional information

`fastify` library
https://github.com/fastify/light-my-request/blob/master/lib/response.js#L10

https://github.com/nodejs/node/blob/main/lib/_http_server.js#L195
https://github.com/oven-sh/bun/blob/bun-v0.8.1/src/js/node/http.ts#L952

birkskyum:
This ServerResponse error seem similar to the ClientRequest error here:
- https://github.com/oven-sh/bun/issues/2297

shaunmolloy:
Using bun v1.0.1 and a similar issue occurs using [snowflake-sdk](https://github.com/snowflakedb/snowflake-connector-nodejs), from the following code:

```ts
import * as snowflake from 'snowflake-sdk';

let connection: snowflake.Connection;

export async function getSnowflake(): Promise<snowflake.Connection> {
[...]

PR #7485, opened 2023-12-06
http convert es5 class

sirenkovladd:
### What does this PR do?

Convert http.OutgoingMessage and http.ServerResponse into es5 syntax
Fix: https://github.com/oven-sh/bun/issues/4415

- [X] Code changes

- [X] I ran `bun fmt`


src/js/node/http.ts
```diff
@@ -1060,246 +1072,247 @@ function onServerResponseClose() {
 }
 
 let OriginalWriteHeadFn, OriginalImplicitHeadFn;
-class ServerResponse extends OutgoingMessage {
-  declare _writableState: any;
+const privateAttribute = Symbol("privateAttribute");
+function getPrivate(obj, key) {
```
paperclover: instead of this, can you (unfortunately) use a separate symbol for each private property. this way we can avoid the extra object

maybe its worth looking into `$putByIdDirectPrivate` also. not sure if it would be meaningful to use it over symbols.

paperclover (changes requested):
see comments

src/js/node/http.ts
sirenkovladd: I'm getting an segmentation fault when using $putByIdDirectPrivate and $getByIdDirectPrivate
```
[RequestContext] create (src.bun.js.api.server.NewRequestContext(false,true,src.bun.js.api.server.NewServer(ZigGeneratedClasses.JSDebugHTTPServer,false,true))@20000610100)
ASSERTION FAILED: symbol
/Users/jarred/actions-runner/_work/WebKit/WebKit/Source/JavaScriptCore/bytecompiler/NodesCodegen.cpp(1726) : RegisterID *JSC::BytecodeIntrinsicNode::emit_intrinsic_putByIdDirectPrivate(BytecodeGenerator &, RegisterID *)
1   0x10695377c WTFCrash
2   0x106390220 Bun::Process_functionDlopen(JSC::JSGlobalObject*, JSC::CallFrame*)
3   0x106e5e01c JSC::BytecodeIntrinsicNode::emit_intrinsic_putByIdDirectPrivate(JSC::BytecodeGenerator&, JSC::RegisterID*)
4   0x106e79660 JSC::ExprStatementNode::emitBytecode(JSC::BytecodeGenerator&, JSC::RegisterID*)
5   0x106e79480 JSC::SourceElements::emitBytecode(JSC::BytecodeGenerator&, JSC::RegisterID*)
6   0x106e793d0 JSC::BlockNode::emitBytecode(JSC::BytecodeGenerator&, JSC::RegisterID*)
7   0x106e79480 JSC::SourceElements::emitBytecode(JSC::BytecodeGenerator&, JSC::RegisterID*)
8   0x106e80254 JSC::FunctionNode::emitBytecode(JSC::BytecodeGenerator&, JSC::RegisterID*)
9   0x106e205f8 JSC::BytecodeGenerator::generate(unsigned int&)
```

paperclover: not a segfault but an assertion failure. but i guess i am incorrect in my understanding of that function. 

we should do the separate symbols approach.

src/js/node/http.ts
```diff
@@ -879,153 +879,169 @@ function write_(msg, chunk, encoding, callback, fromEnd) {
...
 
-  #fakeSocket;
-  #timeoutTimer?: Timer;
-  [kAbortController]: AbortController | null = null;
+Object.setPrototypeOf((OutgoingMessage.prototype = {}), Writable.prototype);
+Object.setPrototypeOf(OutgoingMessage, Writable);
 
-  // Express "compress" package uses this
-  _implicitHeader() {}
+OutgoingMessage.prototype._implicitHeader = function () {};
```
paperclover, line 899: please preserve comment
`// Express "compress" package uses this`

src/js/node/http.ts
```diff
@@ -879,153 +879,169 @@ function write_(msg, chunk, encoding, callback, fromEnd) {
...
-  }
+  return this;
+};
 
-  hasHeader(name) {
-    if (!this.#headers) return false;
-    return this.#headers.has(name);
-  }
+Object.defineProperty(OutgoingMessage.prototype, "headers", {
+  get: function () {
```
paperclover, line 980: comment `// For compat with IncomingRequest`

src/js/node/http.ts
```diff
@@ -879,153 +879,169 @@ function write_(msg, chunk, encoding, callback, fromEnd) {
...
-  _implicitHeader() {}
+OutgoingMessage.prototype._implicitHeader = function () {};
 
-  // For compat with IncomingRequest
-  get headers() {
-    if (!this.#headers) return kEmptyObject;
-    return this.#headers.toJSON();
-  }
+OutgoingMessage.prototype.appendHeader = function (name, value) {
+  this[headersSymbol] = this[headersSymbol] ?? new Headers();
```
paperclover: can you change this back to what it was before. this is an extra property access

```ts
var headers = this[headersSymbol] ??= new Headers();
headers.append(name, value);
```

src/js/node/http.ts
```diff
@@ -879,153 +879,169 @@ function write_(msg, chunk, encoding, callback, fromEnd) {
...
-  set socket(val) {
-    this.#fakeSocket = val;
-  }
+OutgoingMessage.prototype.addTrailers = function (headers) {
+  throw new Error("not implemented");
+};
 
-  get connection() {
-    return this.socket;
+OutgoingMessage.prototype[kClearTimeout] = function () {
```
paperclover: since kClearTimeout is a private symbol, this method shouldnt be accessible from outside this file. we can rewrite this to just a function

```ts
function outgoingMessageClearTimeout(self)
```

src/js/node/http.ts
```diff
@@ -1060,246 +1076,242 @@ function onServerResponseClose() {
...
-  _removedConnection = false;
-  _removedContLen = false;
-  _hasBody = true;
-  #deferred: (() => void) | undefined = undefined;
-  #finished = false;
-  // Express "compress" package uses this
-  _implicitHeader() {
-    // @ts-ignore
-    this.writeHead(this.statusCode);
+ServerResponse.prototype._implicitHeader = function () {
```
paperclover, line 1103: preserve comment
`// Express "compress" package uses this`

paperclover (changes requested):
this looks great though i have some really stupid nitpicks. after that should be good to merge

src/js/node/http.ts
sirenkovladd: I don't see this method being used, can I remove it?

paperclover (approved):
lets see how tests go

PR #7489, opened 2023-12-06
fix(upgrade): print a better message on NixOS

paperclover:
### What does this PR do?

even if i do not enjoy using NixOS, i am quite the enjoyer of better error messages


Jarred-Sumner:
did you manually test this somehow?

paperclover:
haven't manually tested but i have seen reports of this situation and my understanding of nixos should make this be the case.

PR #7498, opened 2023-12-07
fix(create): allow `bun create @scope` -> `bunx @scope/create`

paperclover:
### What does this PR do?

This is an edge case we didn't handle.

Added tests.


Jarred-Sumner:
Thank you

PR #7506, opened 2023-12-07
fix: `stream` on sliced `BunFile`

### LINKED ISSUE #7057: `stream` on sliced `Bunfile` doesn't work

1.0.11

_No response_

#### Create the following files

`./file.txt`

```txt
12345
```

`./index.ts`

```ts
const file = Bun.file('./file.txt').slice(0, 2);

const content = await(new Response(file.stream())).text();

console.log(content);
console.log(file.size);
```

#### Run

```bash
bun run index.ts
```

Console output:

```bash
12
2
```

### What do you see instead?

Console output:

```bash
12345
2
```

### Additional information

_No response_

veera-sivarajan:
### What does this PR do?

Stores sliced offsets inside `FileReader` for `.stream()` to use. Fixes #7057 

- [X] Code changes

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -74,6 +74,7 @@ pub const Blob = struct {
 
     size: SizeType = 0,
     offset: SizeType = 0,
+    was_sliced: bool = false,
     /// When set, the blob will be freed on finalization callbacks
     /// If the blob is contained in Response or Request, this must be null
     allocator: ?std.mem.Allocator = null,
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
Thank you for this. The tests don’t hit the code path unfortunately. Bun implements fast paths for converting streams into native representations when using Response or Request. The chunks will need to be read manually

veera-sivarajan:
Thanks for the quick review.

Yeah, I see that tests fail when checking streams on Response or Request, but I'm not sure if I understand this correctly. Should I reimplement BunFile's `.stream()` to read the chunks manually?

Issue #7451, opened 2023-12-04
Add `-h` as an alias of `--help` to CLI commands


RiskyMH:
# What version of Bun is running?

1.0.15+b3bdf22eb

# What platform is your computer?

Linux 5.15.133.1-microsoft-standard-WSL2 x86_64 x86_64

# What is the issue?

Most commands have no help menu command  using `-h` and only `--help`, for example:
* bun install
* bun add
* bun init
* bun create
* bun upgrade
* bun pm
* bun link 
* bun unlink

And some failed color highlighting
* `bun run -h` in command description such as  `--config`

# What is the expected behavior?

For all commands to a have a basic help menu to at least describe their purpose and for color highlighting to work and not just show the color codes.

# What do you see instead?

`bun run -h` showing the color highlighting codes instead of actioning them
![image](https://github.com/oven-sh/bun/assets/56214343/21d28343-c043-4ef3-926b-43a3fbf7a26c)

`bun init -h` just runs the command and ignores help flag
![image](https://github.com/oven-sh/bun/assets/56214343/8a0dc1d4-c3f1-4ca8-aa2d-25161674c3b2)

`bun link -h` says the flag is invalid
![image](https://github.com/oven-sh/bun/assets/56214343/000fa1cc-2249-4e8d-be03-638738f61b07)

`bun add -h` also throws invalid flag, but I feel like it can just ignore the flag and show the help (like when ran basic `bun add`)

# Additional information

I feel like the help commands not existing is ok, but I would like it more if I could get minor information. And the failed color is something that should be fixed.

Jarred-Sumner:
This is not accurate, the flag is `--help`. We should add an alias for `-h` though

RiskyMH:
ohh your right, im really sorry about this

I saw some worked with -h and assumed everything... but yeah i would like alias for -h and also fix `bun run`'s help issue (should i make a different issue for this?)

RiskyMH:
Looks like #7519 added this!

PR #7519, opened 2023-12-07
Add -h as alias to --help

hustLer2k:
### What does this PR do?

Implements https://github.com/oven-sh/bun/issues/7451.

- [x] Code changes

_**Existing tests pass and manual testing.**_


src/cli.zig
```diff
@@ -368,7 +368,7 @@ pub const Arguments = struct {
             Global.exit(1);
         };
 
-        const print_help = args.flag("--help");
+        const print_help = args.flag("--help") or args.flag("-h");
```
paperclover: did you check if the `or args.flag("-h")` is needed. i think clap is able to do this implicitly once you speicfy the -h alias  in the param list

hustLer2k: it's actually not needed

src/cli.zig
```diff
@@ -1515,7 +1515,7 @@ pub const Command = struct {
                 const print_help = brk: {
                     for (bun.argv()) |arg_| {
                         const arg = bun.span(arg_);
-                        if (strings.eqlComptime(arg, "--help")) {
+                        if (strings.eqlComptime(arg, "--help") or strings.eqlComptime(arg, "-h")) {
```
paperclover: unneeded

paperclover (changes requested):

src/cli/init_command.zig
```diff
@@ -90,7 +90,7 @@ pub const InitCommand = struct {
         const print_help = brk: {
             for (argv) |arg_| {
                 const arg = bun.span(arg_);
-                if (strings.eqlComptime(arg, "--help")) {
+                if (strings.eqlComptime(arg, "--help") or strings.eqlComptime(arg, "-h")) {
```
paperclover: unneeded x2

paperclover:
looks great besides hte unneeded ones

paperclover (approved):
looks great. will merge after tests

PR #7542, opened 2023-12-09
Unconditionally reset stack on deepEqual exit

dfaio:
### What does this PR do?

deepEqual maintains a stack for cycle checking. The bug was that while checking an array, the prior array elements would not be removed from the stack before checking later array elements.

Fix by unconditionally resetting the stack on deepEqual exit.

- [ ] Code changes

Added a test to expect.tests that reproduces the issue. The [obj,obj] array thought that there was a self reference between the "first" array element and the "zeroth" array element.


paperclover (approved):
nice

dylan-conway (approved):

PR #7543, opened 2023-12-09
Improve bun-types handling of JSON types & some extras

Didas-git:
Auto formatter had some fun with braces

### What does this PR do?

- [ ] Code changes

### Notes

This has "breaking" changes.
We should not blindly follow the bad practices of node and should instead enforce better types on JSON like we do for other interfaces.

I decided to also apply this to `Array.isArray` as well for the same reason.

On the other hand i also decided to fix the issue where `Array.filter(Boolean)` doesnt properly reflect the types in typescript (this is just a QoL improvement and could be removed from this pr)

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -207,10 +207,10 @@ declare module "bun" {
      *
      * @param {string} command The name of the executable or script
      * @param {string} options.PATH Overrides the PATH environment variable
-     * @param {string} options.cwd Limits the search to a particular directory in which to searc
+     * @param {string} options.cwd Limits the search to a particular directory in which to search
      *
      */
-    parse(input: string): object;
+    parse(input: string): Record<string, unknown>;
   }
   export const TOML: TOML;
 
@@ -494,7 +494,7 @@ declare module "bun" {
    * @param stream The stream to consume.
    * @returns A promise that resolves with the concatenated chunks as a {@link String}.
    */
-  export function readableStreamToJSON(stream: ReadableStream): Promise<any>;
+  export function readableStreamToJSON(stream: ReadableStream): Promise<string>;
 
   /**
    * Consume all data from a {@link ReadableStream} until it closes or errors.
@@ -776,7 +776,7 @@ declare module "bun" {
     unref(): void;
   }
 
-  export interface FileBlob extends BunFile {}
+  export interface FileBlob extends BunFile { }
   /**
    * [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) powered by the fastest system calls available for operating on files.
    *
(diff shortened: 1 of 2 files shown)


packages/bun-types/bun.d.ts
```diff
@@ -207,10 +207,10 @@ declare module "bun" {
      *
      * @param {string} command The name of the executable or script
      * @param {string} options.PATH Overrides the PATH environment variable
-     * @param {string} options.cwd Limits the search to a particular directory in which to searc
+     * @param {string} options.cwd Limits the search to a particular directory in which to search
      *
      */
-    parse(input: string): object;
+    parse(input: string): Record<string, unknown>;
```
paperclover, line 213: should match `json<TJSONReturnType = any>(): Promise<TJSONReturnType>;`

packages/bun-types/globals.d.ts
```diff
@@ -984,7 +984,7 @@ declare interface Blob {
    * This first decodes the data from UTF-8, then parses it as JSON.
    *
    */
-  json<TJSONReturnType = any>(): Promise<TJSONReturnType>;
+  json(): Promise<unknown>;
```
paperclover, line 987: this was prior behavior but was changed to the generic because it makes every developer's experience worse.

packages/bun-types/globals.d.ts
```diff
@@ -3876,3 +3876,30 @@ interface Navigator {
 }
 
 declare var navigator: Navigator;
+
```
paperclover, line 3879: this is built in to typescript?

paperclover (changes requested):
any -> unknown makes everyone's development experience worse

packages/bun-types/globals.d.ts
```diff
@@ -3876,3 +3876,30 @@ interface Navigator {
...
+   * @param text A valid JSON string.
+   * @param reviver A function that transforms the results. This function is called for each member of the object.
+   * If a member contains nested objects, the nested objects are transformed before the parent object is.
+  */
+  parse(text: string, reviver?: (this: void, key: string, value: any) => any): unknown;
+}
+
+// Allow `<Array>.filter(Boolean)` to properly reflect
+interface Array<T> {
+  filter(predicate: BooleanConstructor, thisArg?: any): T[];
```
paperclover, line 3896: oh i didnt realize what this is doing. @colinhacks 👀 what do you think about adding this override?

packages/bun-types/globals.d.ts
jakub791: > this was prior behavior but was changed to the generic because it makes every developer's experience worse.

The generic was just a hidden type assertion.

jakub791:
> any -> unknown makes everyone's development experience worse

If you want `any`, you can be explicit and do that type assertion in your code instead. But honestly you likey should properly vaildate the data, which this change would encourage.

PR #7550, opened 2023-12-09
fix: better error message for bun install fs permission errors

paperclover:
i got bored

Three things
- Remove `prettyWarn`. It was a misleading function that only logged information during debug mode.
- Add some Output helpers: `Output.warn`, `Output.note`, and `Output.err`. These do what you actually expect them to, they print the level along with the message itself. I had some fun writing `err`
- Print better warnings when we hit AccessDenied/EACCES during bun install.

image

image


Jarred-Sumner:
This visually looks great, but those test failures need to be addressed before this PR can move forward

paperclover:
image

ah yes. this shouldnt print for this case because all is our outlier that isnt an flag but is

Jarred-Sumner:
conflicts

PR #7565, opened 2023-12-10
fix(console): console format use whatwg standard

### LINKED ISSUE #7401: %f not correctly coercing object

1.0.14+d8be3e51b

```js
class Frac {
  constructor(num, den) {
    this.num = num
    this.den = den
  }
  [Symbol.toPrimitive]() {
    return this.num / this.den
  }
}
console.log("%f", new Frac(1, 2))
console.log("%f", 0.5)
```
```sh
bun testconsole.js
```

```sh
jcbhmr@PIG-2016:~/Documents$ bun testconsole.js
0.5
0.5
```

### What do you see instead?

```sh
jcbhmr@PIG-2016:~/Documents$ bun testconsole.js
[Number (Frac): 0.5]
0.5
```

### Additional information

Chrome:
```js
class Frac {
  constructor(num, den) {
    this.num = num
    this.den = den
  }
  [Symbol.toPrimitive]() {
    return this.num / this.den
  }
}
console.log("%f", new Frac(1, 2))
console.log("%f", 0.5)
VM47:10 0.5
VM47:11 0.5
```

Firefox:
```js
class Frac {
  constructor(num, den) {
    this.num = num
    this.den = den
  }…
0.500000 [debugger eval code:10:9](chrome://devtools/content/webconsole/debugger%20eval%20code)
0.500000
```
https://bugzilla.mozilla.org/show_bug.cgi?id=1846606

should follow this: https://console.spec.whatwg.org/#formatter

WingLim:
### What does this PR do?

Standard link: https://console.spec.whatwg.org/#formatter

Resolves #7401

1. Print NaN when value is a Symbol.
2. Not print the class name when format is float

- [x] Code changes

- [x] add test

### DIFF
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -2130,7 +2130,7 @@ pub const ZigConsoleClient = struct {
                     }
                 },
                 .Integer => {
-                    if (value.isSymbol()) {
+                    if (value.isSymbol() or value.isUndefined()) {
                         this.addForNewLine("NaN".len);
                         writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
                         return;
@@ -2160,7 +2160,7 @@ pub const ZigConsoleClient = struct {
                     writer.print(comptime Output.prettyFmt("<r><yellow>{s}n<r>", enable_ansi_colors), .{out_str});
                 },
                 .Double => {
-                    if (value.isSymbol()) {
+                    if (value.isSymbol() or value.isUndefined()) {
                         this.addForNewLine("NaN".len);
                         writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
                         return;
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -2166,30 +2166,6 @@ pub const ZigConsoleClient = struct {
                         return;
                     }
 
-                    // if (value.isCell()) {
-                    //     var number_name = ZigString.Empty;
-                    //     value.getClassName(this.globalThis, &number_name);
-
-                    //     var number_value = ZigString.Empty;
-                    //     value.toZigString(&number_value, this.globalThis);
-
-                    //     if (!strings.eqlComptime(number_name.slice(), "Number")) {
-                    //         this.addForNewLine(number_name.len + number_value.len + "[Number ():]".len);
-                    //         writer.print(comptime Output.prettyFmt("<r><yellow>[Number ({s}): {s}]<r>", enable_ansi_colors), .{
-                    //             number_name,
-                    //             number_value,
-                    //         });
-                    //         return;
-                    //     }
-
-                    //     this.addForNewLine(number_name.len + number_value.len + 4);
-                    //     writer.print(comptime Output.prettyFmt("<r><yellow>[{s}: {s}]<r>", enable_ansi_colors), .{
-                    //         number_name,
-                    //         number_value,
-                    //     });
-                    //     return;
-                    // }
-
                     const num = value.coerce(f64, this.globalThis);
 
                     if (std.math.isPositiveInf(num)) {
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -2130,6 +2130,12 @@ pub const ZigConsoleClient = struct {
                     }
                 },
                 .Integer => {
+                    if (value.isSymbol()) {
+                        this.addForNewLine("NaN".len);
+                        writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
+                        return;
+                    }
+
                     const int = value.coerce(i64, this.globalThis);
                     if (int < std.math.maxInt(u32)) {
                         var i = int;
@@ -2154,6 +2160,12 @@ pub const ZigConsoleClient = struct {
                     writer.print(comptime Output.prettyFmt("<r><yellow>{s}n<r>", enable_ansi_colors), .{out_str});
                 },
                 .Double => {
+                    if (value.isSymbol()) {
+                        this.addForNewLine("NaN".len);
+                        writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
+                        return;
+                    }
+
                     if (value.isCell()) {
                         var number_name = ZigString.Empty;
                         value.getClassName(this.globalThis, &number_name);
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -2166,31 +2166,31 @@ pub const ZigConsoleClient = struct {
                         return;
                     }
 
-                    if (value.isCell()) {
-                        var number_name = ZigString.Empty;
-                        value.getClassName(this.globalThis, &number_name);
+                    // if (value.isCell()) {
+                    //     var number_name = ZigString.Empty;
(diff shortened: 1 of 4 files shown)


test/js/node/console/console.test.ts
```diff
@@ -88,3 +88,30 @@ test("console._stderr", () => {
     configurable: true,
   });
 });
+
+describe("console formatter", () => {
```
paperclover: this `Console` object uses the `node:console` polyfill which is [confusingly] not the ZigConsoleClient struct. can you move these to `console-log.js` and `console-log.expected.txt`

src/bun.js/bindings/exports.zig
```diff
@@ -2154,31 +2160,37 @@ pub const ZigConsoleClient = struct {
...
+                    //     this.addForNewLine(number_name.len + number_value.len + 4);
+                    //     writer.print(comptime Output.prettyFmt("<r><yellow>[{s}: {s}]<r>", enable_ansi_colors), .{
+                    //         number_name,
+                    //         number_value,
+                    //     });
+                    //     return;
+                    // }
 
-                    const num = value.asNumber();
+                    const num = value.coerce(f64, this.globalThis);
```
paperclover, line 2169: what if this throws

paperclover (changes requested):
can you fix the comments, but also add a bunch of other test cases, like weird nonsense and see if it lines up exactly with the spec. also stuff with `{ [toPrimative]: () => { throw 1 } }`

src/bun.js/bindings/exports.zig
WingLim: After I dive in the code, it seems coerce won't throw error for `f64`, it will return NaN if error occurs.

Because of this, when we run this code: 

```js
console.log("%f", {
    [Symbol.toPrimitive]() {
        throw 1
      },
});
```

it will print NaN, won't print the error stack as same as node.

results:
```
# bun result before this PR:

[Number (Object): ]

# bun result after this PR:

NaN

# node result:

test.js:3
        throw 1
        ^
1
(Use `node --trace-uncaught ...` to show where the exception was thrown)

Node.js v21.2.0

```

See the code here:
https://github.com/oven-sh/bun/blob/5029fc8564f2f2dceef4d83ef4bd7f9ecc17ac72/src/bun.js/bindings/bindings.cpp#L3516-L3527

I have no idea why bun did this.

paperclover: ah i see. `NaN` is better than a crash

WingLim: So, should we keep it? If so, I will add a test for `throw 1` and the result to `NaN`.

paperclover: yeah that sounsd good

test/js/web/console/console-log.js
```diff
@@ -170,3 +170,21 @@ console.log(hole([1, 2, 3], 2));
...
+  [Symbol.toPrimitive]() {
+    return "abc";
+  },
+});
+console.log("%f", Symbol(0.1));
+
+console.log("%d", 1);
+console.log("%d", Symbol(1));
+console.log("%i", 1);
+console.log("%i", Symbol(1));
```
paperclover, line 198: oh, we should also check for new Number(1), "1", "one", 0.5 for %i/%d, toPrimative handling here, objects/functions, etc

src/bun.js/bindings/exports.zig
```diff
@@ -2130,6 +2130,12 @@ pub const ZigConsoleClient = struct {
                     }
                 },
                 .Integer => {
+                    if (value.isSymbol() or value.isUndefined()) {
+                        this.addForNewLine("NaN".len);
+                        writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
+                        return;
+                    }
+
                     const int = value.coerce(i64, this.globalThis);
```
WingLim, line 2139: I got a problem here and need some help, coerce to `i64` won't  return `NaN`.
So if we try: `console.log("%d", "one")`, it will print `0`.

Two problem here:

1. coerceToInt64

When we pass a string like `one`, it will finally call `value.toInt32`.

https://github.com/oven-sh/bun/blob/c2cd7402a03f8dfd10c9f77facf108e626e2ac83/src/bun.js/bindings/bindings.cpp#L3539-L3556

2. [value.toInt32](https://github.com/oven-sh/WebKit/blob/7cd84abfa787e4b96b27c3ef1e28a4eb1aa49aa1/Source/JavaScriptCore/runtime/JSCJSValueInlines.h#L45-L52)

Then, `toNumber` will return a `NaN`, but it will be convert to 0 in the `JSC::toInt32`.

```cpp
ALWAYS_INLINE int32_t JSValue::toInt32(JSGlobalObject* globalObject) const
{
    if (isInt32())
        return asInt32();

    double d = toNumber(globalObject);
    return JSC::toInt32(d);
}
```

I am not familiar of cpp, how can I return a NaN in a int type? Or can we return a maxInt to represent the NaN?
As far as I know, NaN will be and only will be a float.

@paperdave

WingLim: Or I have a simple solution for it:

```suggestion
                    const int = value.coerce(i64, this.globalThis);
                    if (value.isString() and int == 0) {
                        var str = ZigString.Empty;
                        value.toZigString(&str, this.globalThis);

                        if (!str.eql(ZigString.init("0"))) {
                            this.addForNewLine("NaN".len);
                            writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
                            return;
                        }
                    }
```

paperclover: we should follow the spec, which unfortunately says to call `parseInt`. here is what i am thinking after messing with this for a little bit

```zig
// https://console.spec.whatwg.org/#formatter

// 1. If Type(current) is Symbol, let converted be NaN
// 2. Otherwise, let converted be the result of Call(%parseInt%, undefined, current, 10)
const int: i64 = brk: {
    if (value.isInt32()) {
        // Already an int
        break :brk value.asInt32();
    }
    double_convert: if (value.isNumber() or !value.isSymbol()) {
        var value = value.coerceToDouble();

        if (!std.math.isFinite(value)) {
            // for NaN and the string Infinity and -Infinity, parseInt returns NaN
            break :double_convert;
        }

        // simulate parseInt, which converts the argument to a string and
        // then back to a number, without converting it to a string
        const max_before_e_notation = 1000000000000000000000;
        const min_before_e_notation = 0.000001;
        if (value > 0) {
            if (value >= max_before_e_notation) {
                // toString prints 0.000+e0, which parseInt will
                var int: i64 = @floor(value);
                while (value > 10) value = @divFloor(int, 10);
                break :brk 0;
            } else if (value <= min_before_e_notation) {
                // todo
            }
        } else if (value <= -max_before_e_notation) {
            // todo
        } else if (value >= min_before_e_notation) {
            // todo
        }

        // 0000.00 float, floor would be the same as parseInt
        break :brk @floor(value);
    }
    // for NaN and the string Infinity and -Infinity, parseInt returns NaN
    this.addForNewLine("NaN".len);
    writer.print(comptime Output.prettyFmt("<r><yellow>NaN<r>", enable_ansi_colors), .{});
};
```
not fully done but the goal is to keep this as fast as possible

i wonder if we should split .Integer into two types, .ConvertToInteger for the `%i` specifier, and then .Integer for the "yeah, this is always an integer"

paperclover (changes requested):
left a comment

src/bun.js/bindings/exports.zig
paperclover: this might need a separate case for strings where we just call `parseInt` on the string

paperclover:
Closing in favor of my fix as a part of 8733

PR #7568, opened 2023-12-10
bun-types: add recursive option to readdir/Sync

NReilingh:
### What does this PR do?

This corrects the type definitions for `fs.readdir`, `fs.readdirSync`, and `fs.promises.readdir` to include the `recursive` option, which support was added for in #7296 (1.0.15).

- [ ] Code changes

I basically just made this consistent with https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts since that appears to have been the model for these type definitions.

I built the `bun-types` package and imported it into my other project which uses `recursive`; verifying that the LSP was no longer complaining about the invalid type signature.

### DIFF
--- a/packages/bun-types/fs.d.ts
+++ b/packages/bun-types/fs.d.ts
@@ -1721,6 +1721,7 @@ declare module "fs" {
       | {
           encoding: BufferEncoding | null;
           withFileTypes?: false | undefined;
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you

PR #7578, opened 2023-12-11
Bun polyfills

jhmaster2000:
### What does this PR do?

Initial finished version of the polyfills' implementations, does not include bundler integration, will be doing that on a separate PR for organization, enough changes on this one as-is.

Includes full adaptation of the Bun JS APIs test suites to work with the polyfills' own test runner, re-using the existing tests rather than needing new ones, naturally this is not yet hooked up to CI, can be ran locally from within the `packages/bun-polyfills` directory with the command `bun run test-all`.

Introduces the `BUN_POLYFILLS_TEST_RUNNER` environment variable to the test suite for detecting whether it's running on the polyfills' test runner (set to `1`) or native `bun test` (unset).


jhmaster2000:
All these remaining test failures seem unrelated to the changes in this PR, given they're present in other recent PRs too.

packages/bun-polyfills/src/global/index.ts
```diff
@@ -1,9 +1,17 @@
...
 import './process.js';
+import './htmlrewriter.js';
 import os from 'node:os';
 
-//? NodeJS Blob doesn't implement Blob.json(), so we need to polyfill it.
+//? Yet another globalThis alias, because why not right?
+Object.defineProperty(globalThis, 'self', {
+    get() { return globalThis; }, set(_) { },
+    enumerable: true, configurable: true,
+});
```
jdalton, line 12: ☝️ This is more a browser space thing. Adding `self` is not a polyfill for node so I'd keep this as a separate PR feature.

packages/bun-polyfills/src/global/index.ts
```diff
@@ -12,13 +20,43 @@ Blob.prototype.json = async function json<T>(this: Blob): Promise<T> {
...
+        }).formData() as Promise<FormData>;
+    } else if (this.type === 'application/x-www-form-urlencoded') {
+        return readableStreamToFormData(this.stream());
+    } else {
+        throw new TypeError('Blob type is not well-formed multipart/form-data or application/x-www-form-urlencoded');
+    }
+};
+Reflect.set(Blob.prototype, 'readable', undefined /*satisfies BunFile['readable']*/);
+Reflect.set(Blob.prototype, 'lastModified', -1 satisfies BunFile['lastModified']);
+Reflect.set(Blob.prototype, 'exists', (async function exists() {
```
jdalton, line 38: Reflect.set is just the verbose way of a direct property assignment here.

packages/bun-polyfills/src/global/index.ts
jdalton: Is there a ifdef kind of way to add only for browser polyfilling? What we have should probably be split out into node, browser, common.

jhmaster2000: At least at the time of having written that code, Bun had `self` defined on runtime, which means server-side Bun code could use it which is why that's there.

packages/bun-polyfills/src/global/index.ts
jhmaster2000: Reflect.set avoids triggering typescript about non-existent or type-mismatched fields, and also gracefully fails silent instead of throwing if the properties have been previously tampered with somehow (e.g. an earlier loaded package) in a way that causes the assignment to fail like making it non-writable.

Issue #7579, opened 2023-12-11
Missing "webcrypto" in "node:crypto".


baboon-king:
### What version of Bun is running?

v1.0.16

### What platform is your computer?

MacOS M1

### What steps can reproduce the bug?

1. 
```ts
import { webcrypto } from "node:crypto";

const encoder = new TextEncoder();
const data = encoder.encode("hello!");

const hashBuffer = await webcrypto.subtle.digest("SHA-256", data);
```
2. 
get Type Error

Type Error: `Module '"node:crypto"' has no exported member 'webcrypto'.`

![image](https://github.com/oven-sh/bun/assets/63645337/b19d1e68-fd87-4b1e-bf50-2a0e3427eea5)

### What is the expected behavior?

```ts
import { webcrypto } from "node:crypto";

const encoder = new TextEncoder();
const data = encoder.encode("hello!");

const hashBuffer = await webcrypto.subtle.digest("SHA-256", data);

```

Type Error: No Error

### What do you see instead?

Module '"node:crypto"' has no exported member 'webcrypto'.ts(2305)
![image](https://github.com/oven-sh/bun/assets/63645337/8d22150e-32a0-406a-9c9f-5d620a820640)

### Additional information

At runtime, "webcrypto" is present in "node:crypto", so now it's just missing the type definition

PR #7580, opened 2023-12-11
fix(types):  add "webcrypto" to "node:crypto"

baboon-king:
### What does this PR do?

add "webcrypto" to "node:crypto" of TypeScript types

close: #7579 

- [x] Code changes

only change TypeScript types


packages/bun-types/crypto.d.ts
```diff
@@ -3913,6 +3913,13 @@ declare module "crypto" {
     candidate: LargeNumberLike,
     options?: CheckPrimeOptions,
   ): boolean;
+  /**
+   * An implementation of the Web Crypto API standard.
+   *
+   * See the {@link https://nodejs.org/docs/latest/api/webcrypto.html Web Crypto API documentation} for details.
+   * @since v15.0.0
```
KilianB: The version flag is pointing to a nodejs version. In my opinion this should be omitted as it will lead to confusion.

baboon-king: You are right, but this "node:xx" is "Node.js compatibility". The link here points to "Node", so maybe there's nothing wrong with that.
Of course, if it were you, you'd do this

--------
Or are you referring to "@since v15.0.0" as redundant?

baboon-king: see latest commit. 🥳

Issue #6982, opened 2023-11-08
Allow `MAX_HEADERS` to be configurable.


TomHennessy:
### What is the problem this feature would solve?
Currently a system my team and I would like to migrate over is blocked by headers not being passed through. This is because there's a hard limit of 50 set here:
https://github.com/oven-sh/bun/blob/131b96bbba3c2fc02b418e36a1ec0f9f7801171d/packages/bun-uws/src/HttpParser.h#L49C1-L49C1

Allowing this value to be configurable would allow us to increase it as we've audited the headers we're sending and we could get down only to around 70.

### What is the feature you are proposing to solve the problem?
An optional config option that gives control over this constant.

### What alternatives have you considered?
 - Upping the `MAX_HEADERS` value. This doesn't seem like a future proof solution.
 - We attempted to get down to below 50 headers but as it's a very large system it doesn't look possible without a major refactor.

Electroid:
We should probably increase the default value, and adding some way to configure this makes sense.

TomHennessy:
> We should probably increase the default value, and adding some way to configure this makes sense.

That'd be great! Would you like me to create a PR for it? If so, what value sounds reasonable to increase it to?

I'll leave the configurability to you unless you're clear on how you'd like it done and could point me in that direction.

morpig:
@TomHennessy i wonder what response/return data was sent from bun when max_headers are reached?

trying to debug a similar issue, in my case it was "connection terminated"

PR #7586, opened 2023-12-11
Upped the `MAX_HEADERS` value

TomHennessy:
### What does this PR do?

This ups the `MAX_HEADERS` value from `50` to `256`. 

This should be configurable but as a stop-gap I've upped the value so it doesn't cause some headers to drop off in large projects.

We have a large project that requires > 70 headers and currently this maximum value is a blocker to our migration.

I've left room for future-proofing reasons but long term it would be ideal to have this be configurable.

- [x] Code changes

### DIFF
--- a/packages/bun-uws/src/HttpParser.h
+++ b/packages/bun-uws/src/HttpParser.h
@@ -46,7 +46,7 @@ namespace uWS
         friend struct HttpParser;
 
     private:
-        const static int MAX_HEADERS = 50;
+        const static int MAX_HEADERS = 256;
         struct Header
         {
             std::string_view key, value;


TomHennessy:
Linked issue: https://github.com/oven-sh/bun/issues/6982

Jarred-Sumner:
There's a couple more places that'll need to change.

In bindings.cpp, we assume a max of 50 headers. 

In http.zig, the client assumes some maximum of headers as well 

Also we should add a test that checks what happens if the client and server both send and receive 100 headers, to fully check it works

paperclover (changes requested):
(see jarred's comments)

TomHennessy:
Thanks both, I'll take a look when I can and get you something.

If it were to be configurable how would you want it to be structured?

PR #7588, opened 2023-12-12
fix(install): fix semver sorting with prereleases

dylan-conway:
### What does this PR do?
fixes a bug when ordering prereleases. parts of the prerelease would be compared as strings when they shouldn't.

- [ ] Code changes

added tests


paperclover (approved):

paperclover:
image
hmm

PR #7612, opened 2023-12-13
fix(resolver): wildcard tsconfig path not including suffix

james-elicx:
### What does this PR do?

- Fixes bug where wildcard tsconfig paths would fail to resolve due to the suffix not being added.

The error thrown that it fixes:

```
error: Cannot find module "@faasjs/bar" from ".../test/js/bun/resolve/resolve-test.js"
```

Feel free to make changes to this branch if there is something you want to modify :)

- [x] Code changes

I wrote automated tests

If Zig files changed:

### DIFF
--- a/src/resolver/resolver.zig
+++ b/src/resolver/resolver.zig
@@ -2846,6 +2846,12 @@ pub const Resolver = struct {
                 const total_length: ?u32 = strings.indexOfChar(original_path, '*');
                 var prefix_parts = [_]string{ abs_base_url, original_path[0 .. total_length orelse original_path.len] };
 
+                // Concatenate the matched text with the suffix from the wildcard path
+                const suffix = std.mem.trimLeft(u8, original_path[total_length orelse original_path.len ..], "*");
+                var matched_text_with_suffix = r.allocator.alloc(u8, matched_text.len + suffix.len) catch unreachable;
+                defer r.allocator.free(matched_text_with_suffix);
+                bun.concat(u8, matched_text_with_suffix, &.{ matched_text, suffix });
+
                 // 1. Normalize the base path
                 // so that "/Users/foo/project/", "../components/*" => "/Users/foo/components/""
                 var prefix = r.fs.absBuf(&prefix_parts, bufs(.tsconfig_match_full_buf2));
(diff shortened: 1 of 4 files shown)


src/resolver/resolver.zig
```diff
@@ -2846,6 +2846,12 @@ pub const Resolver = struct {
                 const total_length: ?u32 = strings.indexOfChar(original_path, '*');
                 var prefix_parts = [_]string{ abs_base_url, original_path[0 .. total_length orelse original_path.len] };
 
+                // Concatenate the matched text with the suffix from the wildcard path
+                const suffix = std.mem.trimLeft(u8, original_path[total_length orelse original_path.len ..], "*");
+                var matched_text_with_suffix = r.allocator.alloc(u8, matched_text.len + suffix.len) catch unreachable;
+                defer r.allocator.free(matched_text_with_suffix);
+                bun.concat(u8, matched_text_with_suffix, &.{ matched_text, suffix });
```
james-elicx: Concatenating these here is necessary as if it is done via the parts array, it would have a backslash in instances where one is not expected (e.g. when the suffix is a file extension like .js).

Not sure if this is the ideal way to concat them though.

src/resolver/resolver.zig
```diff
@@ -2846,6 +2846,12 @@ pub const Resolver = struct {
                 const total_length: ?u32 = strings.indexOfChar(original_path, '*');
                 var prefix_parts = [_]string{ abs_base_url, original_path[0 .. total_length orelse original_path.len] };
 
+                // Concatenate the matched text with the suffix from the wildcard path
+                const suffix = std.mem.trimLeft(u8, original_path[total_length orelse original_path.len ..], "*");
+                var matched_text_with_suffix = r.allocator.alloc(u8, matched_text.len + suffix.len) catch unreachable;
```
Jarred-Sumner: 1) Please do this without dynamically allocating temporary memory, possibly adding another temporary buffer to the threadlocal list used elsewhere in this file

2) Do we need to do this on each call? Can we avoid doing it in certain cases?

james-elicx: Since the threadlocal buffers are a fixed size, wouldn't using one of them mean that when it's joining the other strings together at the end, it would be trying to join a buffer that is larger than its contents and try to access indexes that don't exist?

Good point, this only needs to happen when `total_length != null` since that is the only time this value is consumed.

Jarred-Sumner: the size of the buffers we use are the max allowed filepath length. we could add 2x to that. but filesystems by default limit the path length to be around 1024 bytes (4096 if linux)

james-elicx: I just pushed a new version using a buffer in the threadlocal list. Is this more like what you had in mind?

Jarred-Sumner:
Thank you

PR #7614, opened 2023-12-13
fix: do not inline `process.env` during `bun run` and `bun test`

### LINKED ISSUE #7609: You still cant override `process.env.NODE_ENV` at runtime
```
process.env.NODE_ENV = "production";
console.log(process.env.NODE_ENV);
```
This was initially a feature, but considered a bug since probably the start of the runtime. but process.env is automatically inlined, so this doesnt work.

paperclover:
### What does this PR do?

No longer inline env vars in bun run and bun test. this fixes #7609.

In bun test, explicitly setting NODE_ENV applies.

In bun test, dynamically reading NODE_ENV reads test if an explicit node_env is set

Removes check for adding `BUN_HMR_ENABLED` as this variable is not documented anywhere and never was functional due to a typo.
```
console.log(process.env.BUN_HMR_ENABLED);process.exit(0)
```

many new tests for different configs. bundler is tested by existing

### DIFF
--- a/src/cli/test_command.zig
+++ b/src/cli/test_command.zig
@@ -569,6 +569,7 @@ pub const TestCommand = struct {
         var env_loader = brk: {
             var map = try ctx.allocator.create(DotEnv.Map);
             map.* = DotEnv.Map.init(ctx.allocator);
+            try map.put("NODE_ENV", "test");
 
             var loader = try ctx.allocator.create(DotEnv.Loader);
             loader.* = DotEnv.Loader.init(map, ctx.allocator);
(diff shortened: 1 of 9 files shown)


src/cli/test_command.zig
```diff
@@ -640,6 +639,14 @@ pub const TestCommand = struct {
         vm.bundler.options.rewrite_jest_for_tests = true;
         vm.bundler.options.env.behavior = .load_all_without_inlining;
 
+        const node_env_entry = try env_loader.map.getOrPutWithoutValue("NODE_ENV");
```
Jarred-Sumner, line 642: key must also be cloned I think but only if it didn’t previously exist

Issue #7605, opened 2023-12-12
bun install --ignore-scripts should take an optional list of packages


mrdomino:
### What is the problem this feature would solve?

I recently had my build broken by `protobufjs` trying to run `node scripts/postinstall` in its `postinstall` target. (Incidentally, this happened on upgrading to the latest `oven/bun:alpine` from one that was about three weeks old. If that is a bug, I can report it separately.)

I was able to retrieve a working build by passing `--ignore-scripts` in my Dockerfile, but this blanket-disables all scripts for all packages, whereas I only want to disable the one that was broken.

### What is the feature you are proposing to solve the problem?

Passing `--ignore-scripts=protobufjs` or even `--ignore-scripts=protobufjs:postinstall` would let me more selectively target just the script I don't want to run, without surprising behavior if other packages need other scripts to work.

paperclover:
a tangentally related idea, cc @dylan-conway, what if `trustedDependencies` could be used as a negation like `["-protobufjs"]` to ignore their script.

though looking into this a bit closer.

- protobufjs' postinstall script doesnt do anything helpful (it prints warnings but most package managers do not show script logs unless they fail)
- a bug in bun causes this script to fail

ricardojmendez:
I can confirm the report - installing works with bun's docker image up until 1.0.16, but broke in 1.0.17.

What workaround are you using, @mrdomino? Staying on an older image, or something else?

ricardojmendez:
This issue persists on 1.0.18+36c316a24 - how it manifests with protobufjs is that it ends up stuck on the post-install script, instead of aborting like it did on 1.0.17.

![Screenshot 2023-12-16 at 09 40 54](https://github.com/oven-sh/bun/assets/31263/7429e05b-6929-4880-abbd-9f2619f4903f)

Jarred-Sumner:
If you set `”trustedDependencies”: []` in package.json, it will prevent any dependencies scripts from loading. That should do most of what you want.

ricardojmendez:
> If you set `”trustedDependencies”: []` in package.json, it will prevent any dependencies scripts from loading. That should do most of what you want.

Thanks @Jarred-Sumner - that does not help with the protobufjs install issue, though.

mrdomino:
> I can confirm the report - installing works with bun's docker image up until 1.0.16, but broke in 1.0.17.
> 
> What workaround are you using, @mrdomino? Staying on an older image, or something else?

[...]

Issue #7504, opened 2023-12-07
using `/tmp/bun-node` for overriding node.js is incorrect


paperclover:
### What steps can reproduce the bug?

If you at any point ever install `bun` in one place, run `bun --bun`, and then install it somewhere else, the second bun will use the old symlink and point to the first installation (which may be deleted)

This is rare but there are cases this can happen:
- if you are on a machine where two people have bun installed to `~/.bun/bin/bun`
- you download any release manually and run it from your downloads folder
- you use any other package manager to install bun that does not put it in `~/.bun/bin` (if you use `sudo yarn i -g`, this can end up with `/usr/local/bin/bun`)
- otherwise tampering with `/tmp/bun-node` (note: /tmp is usually mutable by any user)

a user reported this on nixos, where the binary is located at a unique path, it is pointing to a version as old as 0.6.2, though this issue isn't specific to nix.

![image](https://github.com/oven-sh/bun/assets/24465214/d596ba29-a81d-4cc9-a889-725a4a8adf62)

### What is the expected behavior?

i think a solution would be to move this into either XDG/Mac cache dir or into `.bun` somewhere, and/or use some hash of argv0, like `bun-node-AJVZHD` so that multiple installs point properly without having to fully validate them every time.

paperclover:
New test shows this bug in our own CI, yet another reason `/tmp/bun-node` was a bad idea.

<img width="815" alt="image" src="https://github.com/oven-sh/bun/assets/24465214/8dd159ec-fd9a-494f-be8f-3f4640f268e9">

LucaRed:
This bug is still present in v1.1.34.

Example:
1. User **foo** installs bun in their `~/.bun` folder.
2. User **foo** launches some command through bun.
3. A temporary folder such as `/tmp/bun-node-5e5e7c60f` is created, inside of which there are two symlinks pointing to excecutables in `/home/foo/.bun/bin`
4. User **bar** tries to launch a command (e.g. `vite build`) from their installation of bun. The following error appears.
```
/usr/bin/env: ‘node’: Permission denied
error: "vite" exited with code 126
```

This happens because **bar**'s installation of bun tries to use the same temporary folder, but it can't access those symlinks.

A solution would be to use a random or user-scoped temporary folder.

PR #7615, opened 2023-12-13
fix(node): always use .RunCommand when binary is named `node`

### LINKED ISSUE #7610: bun vs node cli prevents some lifecycle scripts from working

_No response_

_No response_

Related to #7605 

```
mkdir scripts
touch scripts/hello.js
```

```
bun --bun run node scripts/hello
```

it should resolve and run the script since node can

### What do you see instead?

error: script not found "scripts/hello"

### Additional information

cc @dylan-conway

paperclover:
### What does this PR do?

This makes `bun --bun node` always use the run command. Fixes #7610.

image

Related issue #7605, but this just fixes the underlying case that prompted wanted having a feature to disable a single script. I think its fine to keep that open as there are legitimate reasons to disallow a single script, and that probably can come with a more robust `trustedDependencies` field.

Also, this tweaks some error messages to be more consistent with each other.

image

also addresses #7504 by using `/tmp/bun-node-<git sha>` however this is not fully closing the issue as it does not actually fix the issue it describes, but simply makes it harder to hit.

I wrote tests

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -15,10 +15,11 @@ const JSC = bun.JSC;
 const WaiterThread = JSC.Subprocess.WaiterThread;
 
 const lex = bun.js_lexer;
-const logger = @import("root").bun.logger;
-const clap = @import("root").bun.clap;
-const cli = @import("../cli.zig");
-const Arguments = cli.Arguments;
+const logger = bun.logger;
+const clap = bun.clap;
+const CLI = bun.CLI;
+const Arguments = CLI.Arguments;
+const Command = CLI.Command;
 
 const options = @import("../options.zig");
 const js_parser = bun.js_parser;
@@ -31,7 +32,6 @@ const sync = @import("../sync.zig");
 const Api = @import("../api/schema.zig").Api;
 const resolve_path = @import("../resolver/resolve_path.zig");
 const configureTransformOptionsForBun = @import("../bun.js/config.zig").configureTransformOptionsForBun;
-const Command = cli.Command;
 const bundler = bun.bundler;
 
 const DotEnv = @import("../env_loader.zig");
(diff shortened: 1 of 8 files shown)


src/cli/run_command.zig
```diff
@@ -17,7 +17,8 @@ const WaiterThread = JSC.Subprocess.WaiterThread;
 const lex = bun.js_lexer;
 const logger = @import("root").bun.logger;
 const clap = @import("root").bun.clap;
-const Arguments = @import("../cli.zig").Arguments;
+const cli = @import("../cli.zig");
```
Jarred-Sumner:
```suggestion
const cli = bun.CLI;
```

src/cli/run_command.zig
```diff
@@ -1731,6 +1732,43 @@ pub const RunCommand = struct {
 
         return false;
     }
+
+    pub fn execAsIfNode(ctx: Command.Context) !void {
+        std.debug.assert(cli.pretend_to_be_node);
+
+        if (ctx.positionals.len == 0) {
```
Jarred-Sumner, line 1757: we could technically make this open `bunx bun-repl`

src/cli/run_command.zig
```diff
@@ -1731,6 +1732,43 @@ pub const RunCommand = struct {
...
+        // var win_resolver = resolve_path.PosixToWinNormalizer{};
+
+        const filename = ctx.positionals[0];
+
+        const normalized_filename = if (std.fs.path.isAbsolute(filename))
+            // TODO(@paperdave): merge windows branch
+            // try win_resolver.resolveCWD("/dev/bun/test/etc.js");
+            filename
+        else brk: {
+            const cwd = try bun.getcwd(&path_buf);
```
Jarred-Sumner, line 1772: we haven't already loaded the cwd at this point?

paperclover: i dont think so. code copied lightly from the existing code path

PR #7628, opened 2023-12-13
Fix incorrect pipe reading code for lifecycle scripts, add basic ioctl_ficlone support

Jarred-Sumner:
### What does this PR do?

Previously, the code for reading the output from a lifecycle script could have a blocking read and would only read up to 32 bytes at a time initially (much too little). This fixes that, and adds a small abstraction we can use for more things later when reading from a FIFO without blocking. 

This also adds `ioctl_ficlone` which is the linux version of `clonefile`. It's supported by brtfs and overlayfs, and not much else. It's completely untested. These changes don't necessarily need to batch together, but I learned that `copy_file_range` is unsupported by gVisor which means it will frequently not work in CI environments. So if we can use `ioctl_ficlone` that's an improvement at least.

Ran the existing tests


src/io/PipeReader.zig
```diff
@@ -0,0 +1,117 @@
...
+/// Read a blocking pipe without blocking the current thread.
+pub fn PipeReader(
+    comptime This: type,
+    // Originally this was the comptime vtable struct like the below
+    // But that caused a Zig compiler segfault as of 0.12.0-dev.1604+caae40c21
+    comptime getFd: fn (*This) bun.FileDescriptor,
+    comptime getBuffer: fn (*This) *std.ArrayList(u8),
+    comptime onReadChunk: ?fn (*This, chunk: []u8) void,
+    comptime onRegisterPoll: ?fn (*This) void,
+    comptime onDone: fn (*This, []u8) void,
```
zackradisic: I feel like this should be renamed to `registerPoll` and `done`

The `on` prefix makes me think its just a callback thats called after the poll is registered, or it is done, but it looks like these functions are the ones actually registering the poll and closing/cleaning this up

PR #7648, opened 2023-12-14
chore(build): include `src/*/*/*.zig`

Hanaasagi:
### What does this PR do?

When running `bun run build`, `src/bun.js/test/expect.zig` is not included in the compilation.

local test


Jarred-Sumner:
Thank you

PR #7649, opened 2023-12-14
Node compat: Add `path` member to `fs.Dirent`

### LINKED ISSUE #7358: Node compat: `fs.Dirent` does not have `path` member

1.0.12+85c997513

Darwin 22.6.0 arm64 arm

```javascript
// repro.js
const fs = require('fs');

const result = fs.readdirSync('/private/tmp', { withFileTypes: true });

console.log(result[0].path, result[0].name);
```

```
% node repro.js
/private/tmp someTempFile
```

### What do you see instead?

```
% bun repro.js
undefined someTempFile
```

### Additional information

`fs.Dirent` is documented here: https://nodejs.org/api/fs.html#class-fsdirent

This type is returned from methods on `fs.Dir`, as well as `fs.readdir()` and `fs.readdirsync()` when `withFileTypes` option is `true`.

NReilingh:
### What does this PR do?

This changes the output of `fs.readdir({ withFileTypes: true })` (and all of its variants) to be consistent with node, such that the `Dirent.name` property is the path basename and the `Dirent.path` property is the path dirname. Fixes #7358. This is a "breaking" change since the `Dirent.name` property currently contains the full relative path for `{ recursive: true }`.

- [x] Code changes

I wrote a script at `test/dirent.js` to manually compare the behavior of node and bun for a mock set of folders, subfolders, and files addressed by different paths. I'm new to this repo and to Zig so I know this isn't idiomatic, but it's the best I could come up with so far.

I'm coming in here with very little systems programming experience, so I did my best to follow any patterns I could find that appeared to be relevant to my changes. For example, I added `item.path.deref()` anywhere I saw `item.name.deref()`.

`@since` parameter on the TS type needs to be populated if/when this is scheduled to a release.

None of my changes seem to affect the behavior of `bun repl`.


test/dirent.js
```diff
@@ -0,0 +1,38 @@
+// This script tests the contents of fs.Dirent for all of the methods that return it.
+// Compare running with node vs. running with Bun for compatibility.
+
+const fs = require('fs');
```
Jarred-Sumner: Please add automated tests

This is not an automated test. It won't run in CI.

Have a look at files with `.test.js` or `.test.ts` for an exmaple

NReilingh: Ah, I now see the `readdir` tests in `test/js/node/fs/fs.test.ts` -- is it preferred to add these tests to that same file?

paperclover: that sounds like a good place for the test

paperclover (changes requested):
overall looks fine

however `path` is deprecated in node and they intend to replace it with `parentPath`

https://github.com/nodejs/node/pull/50976
https://github.com/nodejs/node/pull/51020

we should add both properties, where one is an alias for the other.

test/dirent.js
NReilingh: Added some passing tests to test/js/node/fs/fs.test.ts. Ideally I would also test relative paths (i.e. `readdir('../..')`) for behavior consistent with node, but I'm not sure if it's worthwhile to do so, or what a good approach would be for setting the test's working directory.

NReilingh:
@paperdave Ah, didn't know about that Node.js deprecation. I made a change so that both `path` and `parentPath` work.

packages/bun-types/fs.d.ts
```diff
@@ -176,6 +176,17 @@ declare module "fs" {
      * @since v0.0.67
      */
     name: string;
+    /**
+     * The base path that this `fs.Dirent` object refers to.
+     * Deprecated in favor of `parentPath`.
+     * @since v??????
```
paperclover, line 182: `since ????` should be removed

paperclover (approved):
i can update the comment in the code that needs to be changed. looks to go me otherwise

packages/bun-types/fs.d.ts
NReilingh: Seems like this entire file gets deleted to resolve the merge conflict -- if I'm understanding things correctly, type definitions are now in https://github.com/DefinitelyTyped/DefinitelyTyped per #7670.

Jarred-Sumner (changes requested):
I don't think this should've been approved

We should use a rope string for the `path` getter. The `path` getter will rarely be called. As-is, this will be a significant memory usage increase to `readdir`, which matters when calling it on a large directory tree

To create a rope string, we can use `jsRopeString` in C++. We would need to add a binding to `BunString.cpp` that creates a rope string from two different `BunString` pointers and increment the reference count for the `path` string.

NReilingh:
Closing because a different implementation was merged: https://github.com/oven-sh/bun/pull/11135

PR #7658, opened 2023-12-14
Add 'listening' boolean property to 'net:Server'

### LINKED ISSUE #7657: server.listening is not implemented

1.0.18+36c316a24

Darwin 23.2.0 arm64 arm

```typescript
import { createServer } from 'node:net';

const server = createServer();
await new Promise((resolve) => {
    server.listen(0, "localhost", 1, () => { resolve(undefined) });
});

// The `listening` property is not defined (i.e. `undefined` here).
expect(server.listening).toBeTruthy();
```

When the server is up, `listening` should be _true_; before, `listening` should be _false_ – never `undefined`.

### What do you see instead?

_No response_

### Additional information

See https://github.com/nodejs/node/commit/5ef9989bd6f980bccae8a186c93b974f51791a4d.

malthe:
### What does this PR do?

Added `listening` flag to `net:Server`, see https://nodejs.org/api/net.html#serverlistening.

This fixes #7657.

- [x] Code changes

I added a test.

### DIFF
--- a/src/js/node/net.js
+++ b/src/js/node/net.js
@@ -706,7 +706,6 @@ class Server extends EventEmitter {
   [bunSocketServerConnections] = 0;
   [bunSocketServerOptions];
   maxConnections = 0;
-  listening = false;
 
   constructor(options, connectionListener) {
     super();
@@ -727,6 +726,10 @@ class Server extends EventEmitter {
     this[bunSocketServerOptions] = options;
   }
 
+  get listening() {
+    return !!this.#server;
+  }
+
   ref() {
     this.#server?.ref();
     return this;
@@ -743,7 +746,6 @@ class Server extends EventEmitter {
       this.#server = null;
       this[bunSocketServerConnections] = 0;
       this.emit("close");
-      this.listening = false;
       if (typeof callback === "function") {
         callback();
       }
@@ -940,7 +942,6 @@ function emitListeningNextTick(self, onListen) {
       self.emit("error", err);
     }
   }
-  self.listening = true;
   self.emit("listening");
 }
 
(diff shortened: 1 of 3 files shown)


packages/bun-types/net.d.ts
```diff
@@ -528,6 +528,7 @@ declare module "net" {
    * @since v0.1.90
    */
   class Server extends EventEmitter {
+    listening: boolean;
```
cirospaciari: on nodejs this is a getter, so you can do something like:
```suggestion
    readonly listening: boolean;
```

cirospaciari (changes requested):
Thanks for your contributions, if you change it to a getter and all tests pass I will approve and merge it.

src/js/node/net.js
```diff
@@ -706,6 +706,7 @@ class Server extends EventEmitter {
   [bunSocketServerConnections] = 0;
   [bunSocketServerOptions];
   maxConnections = 0;
+  listening = false;
```
cirospaciari: on nodejs this is a getter, so you can do something like:
```suggestion
  get listening() {
      return !!this.#server;
  }
```

packages/bun-types/net.d.ts
malthe: This is now defined in `/@types/node/net.d.ts` where it's actually not marked as readonly, but that could be an upstream contribution then.

(I have since opened up https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67978.)

malthe:
@cirospaciari should be good now, I have changed to a getter.

cirospaciari (approved):
Thank you

PR #7666, opened 2023-12-14
regression fix(env): define a default NODE_ENV

### LINKED ISSUE #7663: process.env.NODE_ENV is undefined in 1.0.18

1.0.18+36c316a24

Darwin 23.2.0 arm64 arm

Create an `index.ts` with the following contents:

```ts
console.log(process.env.NODE_ENV)
```

Run it against Bun 1.0.17 and Bun 1.0.18.

The script should output `development`:

```shell
$ ./bun-1.0.17 index.ts
development
```

### What do you see instead?

It outputs `undefined`:

```shell
$ ./bun-1.0.17 index.ts
development
$ ./bun-1.0.18 index.ts
undefined
```

### Additional information

This feels pretty critical. Lots of web apps have checks for `process.env.NODE_ENV === "production"` and failing to meet those checks will likely cause lots of issues in apps running on Bun. For example:

```tsx
if (process.env.NODE_ENV === "production") {
  sendEmail()
} else {
  logToConsole() // this gets triggered with this bug
}
```

paperclover:
### What does this PR do?

In 1.0.18 we removed NODE_ENV inlining, which fixes many things. But if the system doesnt define a default NODE_ENV (like basically all the time) we need to set this string to development. we now do that.

Fixes #7663, caused by my code in #7614. sorry for the regression

- [ ] Code changes


paperclover:
ah i changed the webkit submodule. its okay. this will be resolved in windows branch soon:tm:

cirospaciari (approved):
LGTM

Issue #5134, opened 2023-09-12
bun-types causes Typescript to ignore compilerOptions.lib


CraigglesO:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

Work on a project that utilized typescript with a lib properly like so:

```ts
{
  "compilerOptions": {
    "target": "esnext",
    "lib": [
      "DOM",
      "DOM.Iterable",
      "ESNext",
      "WebWorker"
    ],
...
}
```

Install bun-types as a dependency or dev-dependency:

```sh
bun i -D bun-types
```

### What is the expected behavior?

My library depends upon the DOM and DOM.Iterable at a minimum. No errors occur and tsc will properly build types / the codebase

### What do you see instead?

<img width="964" alt="Screenshot 2023-09-12 at 1 44 58 PM" src="https://github.com/oven-sh/bun/assets/5249140/0a45454b-598b-4469-a5af-f4985a0913ab">

Webpack refuses to build because DOM library doesn't exist.

As soon as I remove `bun-types`, remove `node-modules`, and reinstall `bun i` everything works agian.

This is a private repository, so if you want me to go out searching for a repo you can perfectly duplicate I can. Figured you might have an idea as to what is causing this without me doing a bunch of extra work.

bjon:
We have the same problem. Only thing that stops us from using `bun test`.

colinhacks:
This is an unfortunately consequence of adding `"types": ["bun-types"]` in tsconfig.

Per https://bun.sh/docs/test/dom, the fix is to add this line at the top of a file somewhere in your project:

```ts
/// <reference lib="dom" />
``` 

You can replace "dom" with any `lib` you want to be added into your project. We're working on better solutions but currently this is the best way.

CraigglesO:
Yep that works.
Added a .d.ts file like so:
```ts
/// <reference lib="dom" />
/// <reference lib="dom.iterable" />
/// <reference lib="esnext" />
/// <reference lib="webworker" />
```

Happy enough with this solution for now since bun has such a massive speedup. Appreciate it. Will leave this open as a bug so that I can properly track.

johnschoeman:
Ran into this issue as well and found an alternate solution of removing this line:

```
/// <reference no-default-lib="true" />
```

from `node_modules/bun-types/types.d.ts`

[...]

PR #7670, opened 2023-12-15
Better types

colinhacks:
Merge in conjunction with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67827

---

### What does this PR do?

- Add `@types/node` as a dependency to `bun-types`. 
- Remove `no-default-lib`. 
- Make type declarations compatible with `lib.dom.d.ts` and `@types/node`

Possible breakages:

- This puts a *lot* of intermediate types into a namespace called `BunJS`. If a users depends on one of these (say, `BlobPart`) to be declared in the global scope, they will need to re-write it to `BunJS.BlobPart`
- If someone's code depends on the old declarations and isn't compatible with the official `@types/node` declarations, they may see TS compiler errors
- Without `no-default-lib`, anyone with `lib: ["dom"]` in their tsconfig.json will now see the DOM types reflected in their project. Currently bun-types is set up to *defer* to lib.dom.d.ts, so those declarations for, say, `Response` or `ReadableStream` effectively override Bun's.
- By extension, Bun overrides the declarations in `@types/node` for the following globals, but in most cases the declarations are identical.
   - Blob
   - Event
   - EventTarget
   - RequestInit
   - Request
   - ResponseInit
   - Response
   - FormData
   - Headers
   - File
   - performance
   - URL
   - URLSearchParams
   - TextDecoder
   - TextEncoder
   - BroadcastChannel
   - MessageChannel
   - MessagePort


paperclover:
> BunJS.ResponseInit

Is this different from lib.dom.d.ts' `RequestInit` which is in the global scope?

Also, I don't really like *that* as the type namespace, because I feel like we dont want people to call the runtime "BunJS" as some already do, and this would push that further.

paperclover:
Fixes #5134 and surely many more

colinhacks:
> Is this different from lib.dom.d.ts' `RequestInit` which is in the global scope?


Oops bad example. `ResponseInit` is an interface and we declare it top-level. I updated to `BlobPart`...a union `type` that can't go top-level because it would conflict with `lib.dom.d.ts`. You can't have multiple `type` declarations with the same name but you can have multiple `interfaces` with the same name, as long as they're compatible.

> Also, I don't really like _that_ as the type namespace, because I feel like we dont want people to call the runtime "BunJS" as some already do, and this would push that further.

It can't be `Bun` because we `declare var Bun` already, and a `namespace` is a runtime construct. Open to other options here but nothing springs to mind.

dsonet:
> > Is this different from lib.dom.d.ts' `RequestInit` which is in the global scope?
> 
> Oops bad example. `ResponseInit` is an interface and we declare it top-level. I updated to `BlobPart`...a union `type` that can't go top-level because it would conflict with `lib.dom.d.ts`. You can't have multiple `type` declarations with the same name but you can have multiple `interfaces` with the same name, as long as they're compatible.
> 
> > Also, I don't really like _that_ as the type namespace, because I feel like we dont want people to call the runtime "BunJS" as some already do, and this would push that further.
> 
> It can't be `Bun` because we `declare var Bun` already, and a `namespace` is a runtime construct. Open to other options here but nothing springs to mind.

What about just `bun`? Easier to type and recognize.

paperclover:
Is there a reason it cant be imported from `bun` how it already is being done.

But also, i'm trying to figure out why namespaces can't be used. this snippet seems to declare all the types properly, but i may be missing something

image

if it's that `namespace Namespace` defines an empty object, that is fine, because the `Bun` global *does* exist as an object.

colinhacks:
> But also, i'm trying to figure out why namespaces can't be used. this snippet seems to declare all the types properly, but i may be missing something

Ah, I'd been getting "Duplicate identifier" errors but they only show up when the namespace includes runtime constructs (makes sense).  Previously I'd been using `var` and `class` inside the `BunJS` namespace but the current approach is type-only so `Bun` will work after all. Cool.

> Is there a reason it cant be imported from bun how it already is being done.

It's kind of annoying to get these types out of the `module "bun"` when they're needed to declare globals. For instance:

```ts
declare module "bun" {
  // see the top of global.d.ts to see why this is conditional
  type _Response = "some conditional type..."
}

// doesn't work, must be a fully qualified name
interface Response extends import("bun")._Response;

// this breaks everything (can't have top-level imports)
import type {_Response} from "bun" 
interface Response extends _Response {};

// only way I'm aware of to make this work is "re-exporting" with a namespace
namespace Bun {
  type _Response = import("bun")._Response;
}
interface Response extends Bun.Response {}
```

I also think it's a decent goal to avoid a bunch of internal types being exported from the `"bun"` module. They'll show up in autocompletion when people are, say, writing out function parameter types. I think that's not ideal because there's a lot, they look intimidating, some are prefixed with `_`, it's rarely desirable. Putting them in a global namespace is cleaner imo.

Issue #7189, opened 2023-11-17
mock.module() does not work on node_modules, especially with default exports


JeffreyATW:
### What version of Bun is running?

1.0.12+85c997513

### What platform is your computer?

Darwin 23.1.0 x86_64 i386

### What steps can reproduce the bug?

package.json
```
{
  "dependencies": {
    "@reduxjs/toolkit": "^1.9.2",
    "@sendgrid/mail": "^7.7.0"
  }
}
```

mail.test.ts
```
import sgMailOrig from "@sendgrid/mail";
import { createNextState } from "@reduxjs/toolkit";
import { expect, mock, test } from "bun:test";

test("mocks a node module with a default export", async () => {
  mock.module("@sendgrid/mail", () => ({
    default: {
      send: 'foo',
    },
  }));
  const sgMailAgain = await import("@sendgrid/mail");
  expect(sgMailAgain.send).toBe('foo'); // fails
  expect(sgMailAgain.default).toBe(undefined); // fails
  expect(sgMailAgain.default.send).toBe(undefined); // fails
  expect(sgMailOrig.default).toBe(undefined); // passes
  expect(sgMailOrig.send).toBe('foo'); // fails
});

test("mocks a node module with named exports", async () => {
  mock.module("@reduxjs/toolkit", () => ({
    createNextState: 'foo',
  }));
  const { createNextState: createNextStateAgain } = await import("@reduxjs/toolkit");
  expect(createNextStateAgain).toBe('foo'); // passes
  expect(createNextState).toBe('foo'); // fails
});

```

### What is the expected behavior?

I should be able to mock the default export for a node module and immediately access its mock value in both original and re-imports.

I should be able to mock a named export for a node module and immediately access its mock value in both original and re-imports.

### What do you see instead?

I cannot mock the default export of a node module at all.

I can only mock the named export of a node module after re-importing it; the original import remains the same (which makes it useless for testing).

### Additional information

Follow-up of #6884.

nfarina:
I have noticed that module mocking does work but only for dynamic imports not already imported statically. If you remove your `import sgMailOrig from "@sendgrid/mail";` at the top, I'll bet it will work.

It looks like their internal tests only check this case as well (from original implementation in #6685):

  https://github.com/oven-sh/bun/pull/6685/files#diff-33d1285e4f8f34d2d01657e87aebd087f56f332997173a9b63bc0750874d8000R60

[...]

PR #7673, opened 2023-12-15
fix: Resolve specifiers passed to `mock.module`

### LINKED ISSUE #7677: Module mocking not working as expected

1.18.0

Darwin 23.0.0 arm64 arm

When using mock.module, I expect the function to be mocked, however nothing seems to happen.

**The function:**
```tsx
import { v4 } from "uuid";

export const getRandomNumber = () => {
  return v4();
};
```

**The test:**
```tsx
import { describe, expect, mock, test } from "bun:test";
import { getRandomNumber } from "./getRandomNumber";

describe("Given the getRandomFunction is called", () => {
  test("Then the length of the number should match 36", () => {
    const value = getRandomNumber();
    expect(value).toHaveLength(36);
  });

  test("Then the value should equal the mocked value", () => {
    mock.module("uuid", () => ({ v4: "mocked-uuid" }));
    const value = getRandomNumber();
    expect(value).toEqual("mocked-uuid");
  });
});
```

**Reproducible repo:**
https://github.com/jacobarvidsson/bun-mocking

The test should pass.

### What do you see instead?

The module is never mocked and generates a new uuid every test run.

### Additional information
Screenshot 2023-12-15 at 10 21 45

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/7189
Fixes #7677 

Previously, mock.module would assume the specifier was already resolved.

For example: `mock.module("./foo", () => ({ bar: 42 }))` would only load successfully if `foo` lived in the path `./dir/foo` (i.e. it did not automatically infer the file extension or otherwise go through module resolution).

This was very confusing because, if you happened to mock a module before it was ever imported, then it would behave as expected because module resolution for mocks happens before the rest of the module resolution code.

Updated tests which newly failed

### DIFF
--- a/src/bun.js/bindings/BunPlugin.cpp
+++ b/src/bun.js/bindings/BunPlugin.cpp
@@ -549,10 +549,14 @@ extern "C" EncodedJSValue JSMock__jsModuleMock(JSC::JSGlobalObject* lexicalGloba
     } else {
         JSC::SourceOrigin sourceOrigin = callframe->callerSourceOrigin(vm);
         const URL& url = sourceOrigin.url();
-        if (url.protocolIsFile()) {
+        if (url.isValid() && url.protocolIsFile()) {
             auto fromString = url.fileSystemPath();
             BunString from = Bun::toString(fromString);
+            auto catchScope = DECLARE_CATCH_SCOPE(vm);
             auto result = JSValue::decode(Bun__resolveSyncWithSource(globalObject, JSValue::encode(specifierString), &from, true));
+            if (catchScope.exception()) {
+                catchScope.clearException();
+            }
 
             if (result && result.isString()) {
                 auto* specifierStr = result.toString(globalObject);
@@ -560,9 +564,20 @@ extern "C" EncodedJSValue JSMock__jsModuleMock(JSC::JSGlobalObject* lexicalGloba
                     specifierString = specifierStr;
                     specifier = specifierString->value(globalObject);
                 }
-            } else {
-                // If module resolution fails, we leave the specifier as-is
-                // You might not have the module installed in the first place
+            } else if (specifier.startsWith("./"_s) || specifier.startsWith("../"_s) || specifier.startsWith("."_s)) {
+                // If module resolution fails, we try to resolve it relative to the current file
+                auto relativeURL = URL(url, specifier);
+
+                if (relativeURL.isValid()) {
+                    globalObject->onLoadPlugins.mustDoExpensiveRelativeLookup = true;
+
+                    if (relativeURL.protocolIsFile())
+                        specifier = relativeURL.fileSystemPath();
+                    else
+                        specifier = relativeURL.string();
+
+                    specifierString = jsString(vm, specifier);
+                }
             }
 
             RETURN_IF_EXCEPTION(scope, {});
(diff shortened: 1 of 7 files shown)


src/bun.js/bindings/BunPlugin.cpp
```diff
@@ -541,16 +542,30 @@ extern "C" EncodedJSValue JSMock__jsModuleMock(JSC::JSGlobalObject* lexicalGloba
         return {};
     }
 
-    if (specifier.startsWith("./"_s) || specifier.startsWith("../"_s) || specifier == "."_s) {
+    if (specifier.startsWith("file:"_s)) {
+        URL url = URL(URL(), specifier);
```
paperclover:
```
URL url(specifier)
if (!url.isValid()) {
  throw
}
```

src/bun.js/bindings/BunPlugin.cpp
```diff
@@ -541,16 +542,30 @@ extern "C" EncodedJSValue JSMock__jsModuleMock(JSC::JSGlobalObject* lexicalGloba
...
+                    specifierString = specifierStr;
+                    specifier = specifierString->value(globalObject);
+                }
+            } else {
+                // If module resolution fails, we leave the specifier as-is
+                // You might not have the module installed in the first place
+            }
+
+            RETURN_IF_EXCEPTION(scope, {});
         }
```
paperclover, line 592:
```
else {
  // throw runtime exception, or RELEASE_ASSERT_NOT_REACHED()
}
```

i am surprised this is not a compiler warning

paperclover (changes requested):
need to handle some edge cases i pointed out but also:

- `mock.module("./non-exist")` should now make a fake file "non-exist" exist that can be imported. similarly `mock.module('react')` should work without react installed

src/bun.js/bindings/BunPlugin.cpp
paperclover: NVM

Jarred-Sumner:
updated

src/bun.js/bindings/BunPlugin.cpp
```diff
@@ -541,16 +542,45 @@ extern "C" EncodedJSValue JSMock__jsModuleMock(JSC::JSGlobalObject* lexicalGloba
         return {};
     }
 
-    if (specifier.startsWith("./"_s) || specifier.startsWith("../"_s) || specifier == "."_s) {
+    if (specifier.startsWith("file:"_s)) {
+        URL url = URL(URL(), specifier);
```
paperclover: this is still an issue as this can be an invalid file url, and calling `fileSystemPath()` will not valid.

`specifier` is taken as `specifierString` which is taken from `callframe->argument(0)` which can be any string

src/bun.js/bindings/BunPlugin.cpp
```diff
@@ -738,6 +768,26 @@ EncodedJSValue BunPlugin::OnLoad::run(JSC::JSGlobalObject* globalObject, BunStri
...
+std::optional<String> BunPlugin::OnLoad::resolveVirtualModule(const String& path, const String& from)
+{
+    ASSERT(virtualModules);
+
+    if (this->mustDoExpensiveRelativeLookup) {
+        String joinedPath = path;
+
+        if (path.startsWith("./"_s) || path.startsWith("../"_s) || path.startsWith("."_s)) {
+            auto url = WTF::URL::fileURLWithFileSystemPath(from);
+            if (url.isValid()) {
```
paperclover: is this ever false? (if so, it should be added as an assertion failure in WebKit)

src/bun.js/bindings/BunPlugin.cpp
```diff
@@ -738,6 +768,26 @@ EncodedJSValue BunPlugin::OnLoad::run(JSC::JSGlobalObject* globalObject, BunStri
...
 }
 
+std::optional<String> BunPlugin::OnLoad::resolveVirtualModule(const String& path, const String& from)
+{
+    ASSERT(virtualModules);
+
+    if (this->mustDoExpensiveRelativeLookup) {
+        String joinedPath = path;
+
+        if (path.startsWith("./"_s) || path.startsWith("../"_s) || path.startsWith("."_s)) {
```
paperclover: i don't think `path.startsWith(".")` is correct. Is this supposed to be `path == "."_s`. if not, then this entire condition boils down to `if(path.startsWith("."_s))`

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -277,49 +286,52 @@ extern "C" JSC::EncodedJSValue functionImportMeta__resolveSync(JSC::JSGlobalObje
     return result;
 }
 
-extern "C" JSC::EncodedJSValue functionImportMeta__resolveSyncPrivate(JSC::JSGlobalObject* globalObject, JSC::CallFrame* callFrame)
+extern "C" JSC::EncodedJSValue functionImportMeta__resolveSyncPrivate(JSC::JSGlobalObject* lexicalGlobalObject, JSC::CallFrame* callFrame)
```
paperclover, line 289: good renaming

paperclover (changes requested):
notes are mostly due to me spending so much time looking at how `WTF::URL` works, but very minor things.

i assume this code does work

src/bun.js/bindings/BunPlugin.cpp
Jarred-Sumner: It should be `./` || `..` because you could theoretically import files prefixed with `.`

PR #7674, opened 2023-12-15
fix: change how exit codes and signals are printed

### LINKED ISSUE #7425: Better error message when test script not available

1.0.15+b3bdf22eb

Darwin 22.6.0 arm64 arm

`bun run test` with a package.json without a test script

error: missing script "test"

### What do you see instead?

```
arnaud@Mac eslint-plugin-react-refresh % bun run test
error: "test" exited with code 1 (SIGHUP)
arnaud@Mac eslint-plugin-react-refresh % bun run non-existing
error: missing script "non-existing"
```

### Additional information

Could be niced also to differentiate script not found and package.json not found

paperclover:
### What does this PR do?

Fixes #7425

This PR was initially to add a better error message, as `bun run test` will run `/bin/test`. So bun test runs a binary named `test` and it exits with code 1, so we print this extra note. You can silence this note by running `bun /bin/test` or running `/bin/test` directly.

image

This also removes the signal name from non-signal based exits, as it isn't accurate.

image

The message here makes it look like I hit an Illegal Instruction, which is inaccurate.

~~For signals this will print it with a description of the code similar to how fish does it.~~ I removed this printing for the case of bun run, but when LifecycleScriptSubprocess runs into it, it will print the signal code like this still. 

image

When a signal code is raised, we now forward that signal code over:
image

- [ ] Code changes

I wrote automated tests


src/js/internal/debugger.ts
```diff
@@ -25,7 +25,7 @@ export default function (
     if (protocol.includes("ws")) {
       Bun.write(Bun.stderr, `Inspect in browser:\n  ${link(`https://debug.bun.sh/#${host}${pathname}`)}\n`);
     }
-    Bun.write(Bun.stderr, dim("--------------------- Bun Inspector ---------------------")+ reset() + "\n");
+    Bun.write(Bun.stderr, dim("--------------------- Bun Inspector ---------------------") + reset() + "\n");
```
Jarred-Sumner: This works, but it is missing an `await`

src/__global.zig
```diff
@@ -93,7 +93,24 @@ pub fn exit(code: u8) noreturn {
...
+        const act = std.os.Sigaction{
+            .handler = .{ .sigaction = @ptrCast(@alignCast(std.os.SIG.DFL)) },
+            .mask = 0,
+            .flags = 0,
+        };
+        std.os.sigaction(@intCast(sig), &act, null) catch {};
+    }
+    // TODO(@paperdave): report a bug that this intcast shouldnt be needed. signals are i32 not u32
+    // after that is fixed we can make this function take i32
+    _ = std.c.raise(@intCast(sig));
```
Jarred-Sumner, line 112: Why are we raising and then aborting?

aborting causes an exit code of 1.

paperclover: this is to do *something* in the case the process is not dead after `raise` is called. since `raise` in zig is not `noreturn` and it returns something, this implies there is some possibility for it to not end the process.

> aborting causes an exit code of 1

this is not true.

image

please see the [manpage on abort(3)](https://man7.org/linux/man-pages/man3/abort.3.html)

Jarred-Sumner:
Looks like Linux build is failing

PR #7675, opened 2023-12-15
docs: Add Tunnelmole as an open source alternative to ngrok and provide examples for both

robbie-cahill:
This PR adds [Tunnelmole](https://github.com/robbie-cahill/tunnelmole-client) as an open source alternative to ngrok for your Discord bot example. I've also added examples for both Tunnelmole and ngrok.

Tunnelmole is a FOSS tunnelling solution with a growing community. The client and server are both open source and its possible to configure the client to use your own self hosted server, or just leave it to use the [tunnelmole.com](https://tunnelmole.com) service, which is the default.

Example:
```
➜  ~ tmole 8000
http://bvdo5f-ip-49-183-170-144.tunnelmole.net is forwarding to localhost:8000
https://bvdo5f-ip-49-183-170-144.tunnelmole.net is forwarding to localhost:8000

### DIFF
--- a/docs/dev/discord.md
+++ b/docs/dev/discord.md
@@ -1,6 +1,6 @@
 ## Creating a Discord bot with Bun
 
-Discord bots perform actions in response to _application commands_. There are 3 types of commands accessible in different interfaces: the chat input, a message's context menu (top-right menu or right-clicking in a message), and a user's context menu (right-clicking on a user).
+Discord bots perform actions in response to _application commands_. There are three types of commands accessible in different interfaces: the chat input, a message's context menu (top-right menu or right-clicking in a message), and a user's context menu (right-clicking on a user).
 
 To get started you can use the interactions template:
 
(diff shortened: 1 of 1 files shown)


paperclover:
this doc page isn't public. and imo should be moved into the [guides](https://bun.sh/guides) section.

robbie-cahill:
@paperdave sounds like a good idea. Where can it be committed to move it to the guides section?

paperclover:
in `docs/guides/*`. unfortunately the side code is not public at the moment so previewing these is not going to be fun. The two-column thing is weird in those files.

note that `bun create` is no longer powered by builtin templates, so `bun create discord-interactions my-interactions-bot` doesn't work. you will have to rewrite the entire guide page.

also, these guides are intended to be short and to the point, and I think that you have too much details on using the tunneling tool than spent actually showing how to do the thing. (in this current draft, the tunneling instructions are half of the entire page). i think just briefly mentioning the popular tunneling services by link and including `ngrok http 1337 # or 'tmole 1337'` within the list of commands to run (more people are familiar with ngrok, so it probably should be displayed first), as opposed to having entire sections for two different ways to do the same thing.

Jarred-Sumner:
I think it's smart to put your project in more docs & guides but it doesn't make sense for us here in this guide

PR #7686, opened 2023-12-15
Use EdgeQL `for` expression for bulk insert in seed.ts

scotttrinh:
### What does this PR do?

Perhaps this was a JavaScript for loop before to make it a little easier to grok for non-EdgeDB users. However, hopefully this is easy enough to understand that people can still follow the example even if they do not know EdgeQL.

Also found some minor schema issue based on the type shown later.

- [ ] Code changes


Jarred-Sumner:
thank you

PR #7688, opened 2023-12-15
Fix file descriptor leak in `readdir`

vlechemin:
### What does this PR do?

Closes file descriptor after `Syscall.open`

- [x] Code changes

- [x] I checked with `lsof` that the file descriptor is closed
- [x] I ran `bun test test/js/node/fs/fs.test.ts` and they pass

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -4972,6 +4972,8 @@ pub const NodeFS = struct {
             .result => |fd_| fd_,
         };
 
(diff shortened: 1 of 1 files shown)


paperclover (approved):
i wonder if we can write an automated test for this.

i think a fd leak was attempted in `spawn-streaming-stdin.test.ts` but im not 100% confident the approach there works.

otherwise looks okay.

Jarred-Sumner:
let's merge, but we need to add a test that checks the count of the file descriptors do not go up

PR #7690, opened 2023-12-16
feat: Make using `await` inside a non-async function have a helpful error message

Jarred-Sumner:
### What does this PR do?

Comparison:

image

Also:
- Adds `notes` as a getter on`BuildMessage`
- Frees the `notes` slice in `deinit`

There are tests. Most of the parser changes were based on what esbuild does.

### DIFF
--- a/src/bun.js/BuildMessage.zig
+++ b/src/bun.js/BuildMessage.zig
@@ -27,6 +27,24 @@ pub const BuildMessage = struct {
         return null;
     }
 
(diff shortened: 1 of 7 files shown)


paperclover:
i broke `test/napi/napi.test.ts` on the windows branch.

this seems to be flaky with test timeout. cc @dylan-conway
image

paperclover (approved):
great

paperclover (changes requested):
can we fix this test case:
```
() => {
  await 1;
};
```

Jarred-Sumner:
~no~ let's fix in a follow-up

PR #7695, opened 2023-12-16
fix: random improvements to `bun.String`

paperclover:
This PR fixes issues relating to empty strings:

- running a file that was empty would trip a ref count assertion
- running a file that was empty would trip a null pointer assertion
- runtime transpiler cache crashes when loading an empty file

In the process I made some changes to how bun.String works, mainly regarding empty strings.

## Changes to `String.createUninitialized`

`String.createUninitialized` now has an explicit assertion for zero length strings, and it's documentation now reads:

> Allocate memory for a WTF::String of a given length and encoding, and
> return the string and a mutable slice for that string.
>
> This is not allowed on zero-length strings, in this case you should
> check earlier and use String.empty in that case.

An example of using it now. Instead of using `@constCast` the character slice is returned alongside the string.

```zig
if (len == 0) return String.empty.toJS();

const string, var chars = String.createUninitialized(.latin1, len);
defer string.deref();

fillWithData(chars);

return string.toJS();
```

The allocation failure in `String.createUninitialized` has always caused an application, which is why this no longer returns an optional.

## `createAtom` -> `createAtomASCII`

`createAtom` is more strict, and also renamed to createAtomASCII. It has more assertions, but less runtime checks. It should only be used with developer hardcoded strings (as all of our current usages)

## `tryCreateAtom` (theoretically) doesn't crash on failure

Haven't confirmed this but the code for `.tryCreateAtom` called JSC's makeAtomString instead of tryMakeAtomString. the non-try one quite literally calls `CRASH()` if the result is null.

So the zig binding no now uses the correct function.

## Removed `bun.String.toJSConst`

Use `bun.String.toJS` instead.

## `-O0`

I set the build to compile with `-O0` because it makes debugging much easier at not much debug runtime performance.

## `String.init` and `String.static`

I spent some time seeing if I could make String.static be forced for all static strings, but I didn't really get far in this.

Maybe in the future I can figure out how to make this an actual `MAKE_STATIC_STRING_IMPL`


src/bun.js/bindings/BunString.cpp
```diff
@@ -508,15 +506,12 @@ WTF::String BunString::toWTFString(ZeroCopyTag) const
...
-
-    if (this->tag == BunStringTag::WTFStringImpl) {
-        ASSERT(this->impl.wtf->refCount() > 0);
+    } else if (this->tag == BunStringTag::WTFStringImpl) {
+        ASSERT(this->impl.wtf->refCount() > 0 && !this->impl.wtf->isEmpty());
         return WTF::String(this->impl.wtf);
     }
 
-    return WTF::String();
+    return WTF::String(MAKE_STATIC_STRING_IMPL(""));
```
Jarred-Sumner:
```suggestion
    return WTF::String();
```

Jarred-Sumner:
This looks good overall. Just some conflicts to address.

Jarred-Sumner (approved):

Jarred-Sumner:
Test failures look non-flaky

Jarred-Sumner:
image

src/bun.js/bindings/BunString.cpp
```diff
@@ -86,15 +92,13 @@ JSC::JSValue toJS(JSC::JSGlobalObject* globalObject, BunString bunString)
...
-JSC::JSValue toJS(JSC::JSGlobalObject* globalObject, BunString bunString, size_t length)
-{
-#if BUN_DEBUG
-    if (bunString.tag == BunStringTag::WTFStringImpl) {
-        RELEASE_ASSERT(bunString.impl.wtf->refCount() > 0);
-    }
-#endif
-    return jsSubstring(globalObject, jsUndefined(), bunString.toWTFString(BunString::ZeroCopy), 0, length);
-}
+// JSC::JSValue toJS(JSC::JSGlobalObject* globalObject, BunString bunString, size_t length)
```
dylan-conway: should this commented code be deleted?

src/bun.js/RuntimeTranspilerCache.zig
```diff
@@ -421,28 +421,23 @@ pub const RuntimeTranspilerCache = struct {
...
-        buf: *[bun.MAX_PATH_BYTES]u8,
-    ) [:0]const u8 {
-        if (is_disabled) return "";
-
-        if (!runtime_transpiler_cache_loaded) {
-            runtime_transpiler_cache_loaded = true;
-            runtime_transpiler_cache = @constCast(reallyGetCacheDir(&runtime_transpiler_cache_static_buffer));
-            if (runtime_transpiler_cache.len == 0) {
+    fn getCacheDir(buf: *[bun.MAX_PATH_BYTES]u8) ![:0]const u8 {
+        if (is_disabled) return error.Disabled;
```
dylan-conway: this should return `error.CacheDisabled` so we don't need to check for len == 0 where we call `try getCacheFilePath`

src/bun.js/RuntimeTranspilerCache.zig
```diff
@@ -421,28 +421,23 @@ pub const RuntimeTranspilerCache = struct {
...
+        const path = runtime_transpiler_cache orelse path: {
+            const path = reallyGetCacheDir(&runtime_transpiler_cache_static_buffer);
+            if (path.len == 0) {
                 is_disabled = true;
-                return "";
+                return error.Disabled;
```
dylan-conway: same here

dylan-conway (approved):
Looks good after `error.CacheDisabled` change

PR #7697, opened 2023-12-16
perf: make `Bun.peek` 90x faster on pending promises, 1.5x on resolved

paperclover:
image


Jarred-Sumner (approved):
Awesome. You should tweet about it 

in the next version of Bun 

Bun.peek(promise) gets 90x faster

Bun.peek() lets you view a Promise’s wrapped value

PR #7702, opened 2023-12-16
fix: text imports should escape backslashes correctly (#3449)

### LINKED ISSUE #3449: text loader parses backslashes

0.6.11

Darwin 22.4.0 x86_64 i386

I was able to reproduce using `bun test`, or logging both moons to the console — https://github.com/PineappleRind/bun-import-txt 

imo the text loader should act like readFileSync and leave escapes as-is

### What do you see instead?

Having an escaped space in a txt file (`\ `) results in the backslash being removed after importing (or it seems so, console.log doesn't show them)

`bun test`

image

the screenshot shows that moonFromFS() has `\\` though I think that's just `bun test`, the original file doesn't

### Additional information

My current workaround is to escape the backslash (`\\`) in the txt file itself

sportshead:
### What does this PR do?

Text imports, which previously were just put into a JS template string, now have backslashes escaped before being outputted to a JS string.

Fixes #3449

- [x] Code changes

^ not sure about this one. It's my first time using Zig, and I've done my best to follow best practices re allocating memory. Might have accidentally introduced a memory leak though.


src/bundler.zig
```diff
@@ -1536,8 +1536,12 @@ pub const Bundler = struct {
             },
             // TODO: use lazy export AST
             .text => {
+                const size = std.mem.replacementSize(u8, source.contents, "\\", "\\\\");
```
Jarred-Sumner, line 1561: do we need to worry about freeing source.contents? I think the answer is no because they should be allocated in the threadlocal heap. But I'm not 100% sure.

sportshead: Might be a memory leak, since running `bun-debug test regression` with this change results in a failed assertion at the end of the tests:
```
...
 23 pass
 0 fail
 snapshots: +1 added
 83 expect() calls
Ran 23 tests across 18 files. [2.63s]
mimalloc: assertion failed: at "/Users/user/projects/bun/src/deps/mimalloc/src/alloc.c":246, mi_page_usable_size_of
  assertion: "ok"
Abort trap: 6
```

Using the previous commit and running regression test (with the new tests checked out) results in:
```
...
1 tests failed:
✗ importing .txt files should not parse backslashes [22.03ms]

 22 pass
 1 fail
 snapshots: +1 added
 82 expect() calls
Ran 23 tests across 18 files. [2.67s]
``` with no failed assertions.

However the weird thing is that running `bun-debug run 03449-fixture.ts` with these changes and `bun-debug test 03449.test.ts` doesn't cause an assertion fail, and executes properly.

Jarred-Sumner: This means the allocator in use should be  bun.default_allocator

sportshead: @Jarred-Sumner I've updated the PR with deallocation through the `allocator` local var. I tried using `bun.default_allocator`, but that still caused a mimalloc assertion to fail, but using `allocator` allows the test to pass without any failed assertions.

sportshead:
Closed by #10456

Issue #1936, opened 2023-01-30
Install script should use POSIX and not bash


coolaj86:
### What is the problem this feature would solve?

Bun would be universally installable on any POSIX system for which a binary exists.

Well... actually would need a POSIX .tar as well, but... one thing at a time. :) \
(and although `.gz` isn't POSIX, it seems to be more widely adopted than POSIX `.Z`)

**Context:** I'm adding `bun` to [Webi](https://webinstall.dev) (<https://beta.webinstall.dev/bun>). We do POSIX because over the years enough people have complained about bash scripts breaking stuff.

### What is the feature you are proposing to solve the problem?

Running `shellcheck` and `shfmt` in POSIX mode (i.e. change the shebange to `#!/bin/sh` on `install.sh`

### What alternatives have you considered?

N/A

P.S. I'm drafting this up right now. I don't know where install.sh lives, but I'll post below in probably 10 minutes or so.

P.P.S. Taking a little longer than I thought - there's A LOT of fairly complex bashisms here. 

P.P.P.S. There's very inconsistent use of `$x` vs `${x}` vs `"$x"` vs `"${x}"`, so I'm curious to whether or not directories with spaces in their names are currently allowed.

## Where is install.sh?

I didn't see a repo that seemed like a good place to do a PR.

## PR Process

Trying to look at all of these changes in aggregate is pretty much impossible, but I can break them down into feature-by-feature commits such that each checkmarked change is very easy to digest and understand its correctness.

(this, of course, is just a _Works On My Machine™_ Proof-of-Concept)

coolaj86:
# POSIX-compliant install

- Changes
- Proof-of-Concept
- Diff

## Changes

- [x] replace `echo` with `printf`
- [x] replace bash `[[` with POSIX `[`
- [x] replace bash inline replace with `sed` (I _think_ I got it right)
- [x] replace bash inline `if` pattern matching with `case`
- [x] replace known-length bash arrays with multiple variables (same lines of code, btw)
- [x] replace unknown-length bash arrays with newline separated via
   ```sh
   my_ifs="${IFS}"
   IFS='
   '
   ```
- [x] replace `local` variables
- [ ] what is `|| :`? (search fail on that one, of course)
- [ ] where is replication of `pipefail` required?

## Works On My Machine™ Proof-of-Concept

(actual `diff` below)

```sh
#!/bin/sh
#shellcheck disable=SC2059
set -e
set -u

[...]

PR #7707, opened 2023-12-17
Dedicated install.sh for Alpine 

alxivnov:
### What does this PR do?

This adds a dedicated Alpine `install-alpine.sh`.
Original `install.sh` has some bash specific script features which prevent its modification, like:
1. `install_dir=${!install_env:-$HOME/.bun}` - sh does not support substituting install_env with BUN_INSTALL and then testing if it is set,
2. `commands=("", "")` - sh does not support arrays.

Also sh evaluates full script (even branches that would not execute), so 2nd problem can't be rewritten w/o copy-pasting commands into for loops.
So, dedicated script is required)

I verified that it works on Alpine containers locally with and without $1 set to version and $2 to set to debug-info.

`wget -O - https://raw.githubusercontent.com/alxivnov/bun/alpine/src/cli/install-alpine.sh | sh` this command would not require to install `curl`. In fact, nothing else is required. `unzip` package is present in the Alpine images I tested.


paperclover (changes requested):
instead of creating a second install script, is it not possible to make the install.sh posix compliant. (this is being tracked by issue #1936, so changes to work towards it are welcome)

PR #7711, opened 2023-12-18
fix: Tweaks to syntax highlighter

Jarred-Sumner:
### What does this PR do?

Top: After
Bottom: Before
image

Manual

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -457,6 +457,7 @@ pub const fmt = struct {
                         '}', '{' => {
                             // support potentially highlighting "from" in an import statement
                             if (!((prev_keyword orelse Keyword.@"continue") == .import)) {
+                            if ((prev_keyword orelse Keyword.@"continue") != .import) {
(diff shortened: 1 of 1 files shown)


src/bun.zig
```diff
@@ -452,9 +454,18 @@ pub const fmt = struct {
                             try writer.writeAll(remain[0..i]);
                             remain = remain[i..];
                         },
-                        '}', '[', ']', '{' => {
+                        '}', '{' => {
+                            // support potentially highlighting "from" in an import statement
+                            if (!((prev_keyword orelse Keyword.@"continue") == .import)) {
```
paperclover, line 459:
```suggestion
                            if (!((prev_keyword orelse Keyword.@"continue") == .import)) {
                            if ((prev_keyword orelse Keyword.@"continue") != .import) {
```

paperclover (changes requested):
awesome, can you simplify the one if statement i commented on

PR #7715, opened 2023-12-18
perf: More efficient `setTimeout` and `setInterval`

Jarred-Sumner:
### What does this PR do?

This makes timers much more efficient by using a timer heap instead of a timerfd

On Linux, this is about an 8x improvement on the below benchmark which schedules 20,000 timers simultaneously. On macOS, it's 5.6x.

image

We are not yet faster than Node on this. I think there is 1 more kind of coalescing to do. 

Benchmark:
```js
const maxPerBatch = 1024 * 20;
const total = 1_000_000;

var ran = 0;
const noop = function () {
  ran++;

  if (ran > maxPerBatch - 1 && ran % maxPerBatch === 0) {
    if (ran >= total) {
      console.log(`Executed ${ran} timers in`, performance.now(), "ms");
      return;
    }

    for (let i = 0; i < maxPerBatch; i++) {
      setTimeout(noop, 0);
    }
  }
};

for (let j = 0; j < maxPerBatch; j++) {
  setTimeout(noop, 0);
}
```

Existing tests

### DIFF
--- a/src/bun.js/api/bun.zig
+++ b/src/bun.js/api/bun.zig
@@ -3517,6 +3517,9 @@ pub const Timer = struct {
                 .callback = &onRequest,
             },
             interval: i32 = -1,
+            concurrent_task: JSC.ConcurrentTask = undefined,
+
+            pub const Pool = bun.HiveArray(TimerReference, 1024).Fallback;
 
             fn onRequest(req: *bun.io.Request) bun.io.Action {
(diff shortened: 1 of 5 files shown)


Jarred-Sumner:
cc @mitchellh, this uses a modified version of your timer heap from libxev

paperclover (approved):
you should include the `setTimeout.js` benchmark used

Issue #7689, opened 2023-12-15
The performance of the `bun install` has been degraded in 1.0.17 in github actions


tonypinkevych:
### What version of Bun is running?

1.0.17

### What platform is your computer?

ubuntu-latest

### What steps can reproduce the bug?

Create simple github action and just add `bun install` step.

```yml
name: Simple bun install reproducable

on:
  push:
    branches:
      - main

jobs:
  validate:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3

      - uses: oven-sh/setup-bun@v1
        with:
          bun-version: '1.0.17' # if you set here `1.0.16` then everything works fine

      - name: Install dependencies
        run: bun install
```

### What is the expected behavior?

It installs in 15-20 seconds, just like before. Find video below.

https://github.com/oven-sh/bun/assets/13913581/ffa21cfa-4b5e-40ba-9be7-b86295ab776e

### What do you see instead?

It holds for 90+ seconds on `bun install v1.0.17` row. Find video below.

https://github.com/oven-sh/bun/assets/13913581/6764fec8-37e4-4ada-8f81-16b76db09cfc

### Additional information

Perhaps the problem is related to this update: https://bun.sh/blog/bun-v1.0.17#bun-install-now-runs-lifecycle-scripts-for-the-top-500-npm-packages. But it worked fine before that.
I hope you can find a solution to make it fast, but still run the necessary scripts.

Hope this helps.

paperclover:
what packages are you installing? this isnt enough to reproduce the issue.

dylan-conway:
If `re2` is being installed this could be related to #7597

tonypinkevych:
@paperdave Yes, you're right, sorry.
@dylan-conway No, I don't use re2. This may have something to do with my research below. 
Hope this helps.

I debugged all the dependencies and found the one that caused the degradation.

[Firebase-tools](https://www.npmjs.com/package/firebase-tools) is very slow to install in the bun `v1.0.17`.

It's `package.json` that reproduces the problem in the github actions:

```json
{
  "name": "bun-test",
  "scripts": {},
  "dependencies": {},
  "devDependencies": {
    "firebase-tools": "^13.0.2"
  }
}
```

**Installation time with the provided package.json:**
[...]

Issue #7597, opened 2023-12-12
Bun i freezes and/or is much slower than npm i, and errors on `re2`


brianmhunt:
### What version of Bun is running?

1.0.17

### What platform is your computer?

Codespace (Linux 6.2.0-1016-azure x86_64 unknown)

### What steps can reproduce the bug?

Running `bun i` with package.json like this (and probably some subset of this) on Linux / Github Codespace.

```json
{
  "name": "bun-test",
  "devDependencies": {
    "@firebase/testing": "^0.20.11",
    "@sentry/browser": "^7.64.0",
    "@types/big.js": "^6.1.6",
    "@types/docusign-esign": "^5.19.0",
    "@types/file-saver": "^2.0.5",
    "@types/jest": "^29.5.1",
    "@types/jss": "^10.0.0",
    "@types/knockout": "^3.4.72",
    "@types/webappsec-credential-management": "^0.6.4",
    "@typescript-eslint/eslint-plugin": "^5.59.6",
    "@typescript-eslint/parser": "^5.59.6",
    "currency-symbol-map": "^5.1.0",
    "esbuild": "^0.17.19",
    "eslint": "^8.40.0",
    "eslint-plugin-import": "^2.27.5",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-react": "^7.32.2",
    "firebase-tools": "^12.6",
    "full-icu": "^1.5.0",
    "jest": "^29.5.0",
    "klaw-sync": "^6.0.0",
    "sinon": "^15.0.4",
    "typescript": "^5.0.4",
    "xunit-viewer": "^10.3.0"
  },
  "dependencies": {
    "@sentry/integrations": "^7.64.0",
    "@stripe/stripe-js": "^1.53.0",
    "@tko/build.reference": "^4.0.0-beta1.5",
    "@types/lodash": "^4.14.194",
    "@types/wicg-file-system-access": "^2020.9.6",
    "big-integer": "^1.6.51",
    "big.js": "^6.2.1",
    "bignumber.js": "^9.1.1",
    "cbor-x": "^1.5.3",
    "chart.js": "^4.3.0",
    "cors": "^2.8.5",
    "dagre": "^0.8.5",
    "date-fns": "^2.30",
    "eslint-plugin-promise": "^6.1.1",
    "express": "^4.18.2",
    "fast-json-patch": "^3.1.1",
    "fast-json-stable-stringify": "^2.1.0",
    "fflate": "^0.7.4",
[...]

PR #7719, opened 2023-12-18
feat: print the longest postinstall if it took more than 500ms

paperclover:
### What does this PR do?

This will help identify the root cause of issues like #7689 and #7597 where a postinstall takes a very long time.

image

the warning is printed above the footer of "x pkgs installed" but below the (potentially very long) list of packages, so that it is seen.

i also remove a empty line present in all bun installs
image
now:
image

- [ ] Code changes


Jarred-Sumner:
Looks good but tests need to be updated.

dylan-conway (changes requested):
Update tests then this is good to go

paperclover:
ok to merge

PR #7725, opened 2023-12-19
fix(windows): store global std{in,out,err} a little differently

paperclover:
### What does this PR do?

This also cleans up alot of the code in the `main()` function.

havent tested if it builds on linux and mac

this resolves calling Bun.stdout.writer() / Bun.file().writer() causing a file descriptor casting issue but the FileSink does not actually work as expected.

### DIFF
--- a/src/__global.zig
+++ b/src/__global.zig
@@ -62,28 +62,28 @@ else
 /// Like std.Thread.setName but auto-truncate, and only works on the current thread.
 /// The u8 -> u16 conversion is ascii-only
 pub fn setThreadName(name: [:0]const u8) void {
-    var truncated_name = switch (bun.OSPathSlice) {
-        [:0]const u8 => if (name.len > std.Thread.max_name_len) brk: {
-            var truncated_name: [std.Thread.max_name_len + 1]u8 = undefined;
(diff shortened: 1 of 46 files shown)


src/__global.zig
```diff
@@ -59,16 +59,61 @@ else if (Environment.isAarch64)
...
-    if (Environment.isLinux) {
-        _ = std.os.prctl(.SET_NAME, .{@intFromPtr(name.ptr)}) catch 0;
-    } else if (Environment.isMac) {
-        _ = std.c.pthread_setname_np(name);
+/// Like std.Thread.setName but auto-truncate, and only works on the current thread.
+/// The u8 -> u16 conversion is ascii-only
+pub fn setThreadName(name: [:0]const u8) void {
+    var truncated_name = switch (bun.OSPathSlice) {
+        [:0]const u8 => if (name.len > std.Thread.max_name_len) brk: {
+            var truncated_name: [std.Thread.max_name_len + 1]u8 = undefined;
```
dylan-conway: this buffer is stack allocated in this block. is it safe to use it later in this function?

paperclover:
too far out of date

PR #7733, opened 2023-12-19
Make large stdio in Bun.spawnSync efficient on Linux

Jarred-Sumner:
### What does this PR do?

This makes spawnSync efficient on Linux when the output or input is large amounts of data. It leverages an in-memory file descriptor to avoid cloning and buffering via pipes

image

Microbenchmark:
```js
const { spawnSync } = require("node:child_process");
const runs = 5;

console.time("cat 1 GB file x " + runs);
for (let i = 0; i < runs; i++) {
  const { stdout } = spawnSync("cat", ["1gig.file"], {
    maxBuffer: Infinity,
    stdout: "pipe",
    encoding: "buffer",
  });
  // access the file somehow
  if (stdout.byteLength < 1024 * 1024 * 1024) {
    throw new Error("Expected huge file");
  }

}
console.timeEnd("cat 1 GB file x " + runs);
console.log("RSS", (process.memoryUsage.rss() / 1024 / 1024) | 0, "MB");
```

Write a 1 gigabyte file:
```js
const buffer = Buffer.allocUnsafe(1024 * 1024 * 1024);

for (let i= 0; i < 256; i++) {
    buffer[i] = i;
}

// fill the buffer with values from 0-255 
for (let i = 0; i < buffer.length; i+= 255) {
    buffer.copyWithin(i, 0, 255);
}

await Bun.write("./2mb.file", buffer);

```

### DIFF
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -196,6 +196,7 @@ pub const Subprocess = struct {
 
     const Readable = union(enum) {
         fd: bun.FileDescriptor,
+        memfd: bun.FileDescriptor,
 
         pipe: Pipe,
         inherit: void,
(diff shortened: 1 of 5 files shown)


src/bun.js/api/bun/subprocess.zig
```diff
@@ -1831,6 +1867,8 @@ pub const Subprocess = struct {
             }
         }
 
+        should_close_memfd = false;
```
dylan-conway, line 1870: this variable seems to only be used once. Replace it with `Environment.isLinux`?

Jarred-Sumner: it's used for errors, to ensure we close the memfd by default unless there's an error

PR #7746, opened 2023-12-19
fix: renaming `strings. endsWithChar` to be more consistant + fix crash in `bun init`

paperclover:
### What does this PR do?

I have totally written some more incorrect code assuming this `endsWithChar` does not include an empty string. imo this is a useless function as startsWith doesnt work this way, JS spec startsWith/endsWith doesnt work like this, and even *webkit's WTF::String* does not work like this.

image

i rename this to a more annoying function name:

image

i dont know if we need it, but to retain compatibility i have moved most usages of the old function to this new name

- [x] Code changes

i wrote a test using `bun init` on piped \n\n\n\n\n\n input. it will segfault on canary due to unchecked access on invalid memory.


dylan-conway (approved):
Nice

PR #7748, opened 2023-12-20
feat: Bun shell

zackradisic:
### What does this PR do?

> Note that this is currently still a WIP, but opening this now to make reviewing easier because the changes are approaching 15k lines. I'm sorry.

This adds a cross-platform bash-like shell scripting language to Bun. Scripts are runnable through a [zx](https://github.com/google/zx) inspired API that allows for embedding of JS expressions into the shell command (similar to embedding JS expressions in JSX):

```typescript
import { $ } from 'bun'

const output = new Uint8Array(1 << 20);
// Redirect the output of `ls` to output
await $`ls > ${output}`
```

Bun shell scripts will also be runnable in package.json scripts, or through `bun run`, and files with the appropriate shebang.

## Shell features
- [x] Supports ascii and unicode
- [x] Pipelining `cmd1 | cmd2`
- [x] Conditional control flow `cmd1 && cmd2`
  - [x] &&
  - [x] ||
- [x] Redirections (`cmd > lmao.txt`)
  - [x] stdout/stderr (`foo 1> lmao.txt` or `foo 2> lmao.txt`)
  - [ ] stdin 
- [x] Shell variables
  - [x] Assigning `FOO=bar`
  - [x] Variable expansion `echo $FOO`
  - [x] Local variables (only used by shell for expansion)
  - [x] Exported (used in env)
- [x] Command substitution (`echo $(cmd with args)`)
- [x] Brace expansion `echo {a,b,c}{d,e,f}`
  - [x] Nested braces `echo {a,b,{c,d}}{e,f}`
- [x] Glob expansion
- [x] zx inspired JS api
  - [ ] Redirecting to JS objects
    - [x] Array buffer
    - [ ] Subprocesses?
    - [ ] Other shell instances?
   - [ ] JS objects in a pipeline? (e.g. `${someBuffer} | cmd1 | cmd2`)
  - [x] Custom thenable to lazily execute scripts
- [ ] Runnable through `bun run`
- [ ] Runnable through package.json script
- [ ] Shebang

## Supported builtin commands
- [x] `export`: set environment variables for the shell session to be used by subprocesses
- [x] `cd`: change the current working directory for the shell
  - [x] `cd <arg>`
  - [x] `cd -` (go to previous directory)
- [x] `echo`: write text to stdout
- [x] `pwd`: print previous directory
- [x] `which`: print absolute paths of stuff
- [x] `rm`: delete stuff
  - [x] recursively delete directories concurrently
  - [x] `-v`: verbose mode, prints out what is deleted
  - [ ] `-i`: interactive mode, ask user for confirmation
- [x] `mv`: move/rename stuff
  - [x] move files and directories
  - [x] verbose mode
  - [ ] support moving across different filesystems
  - [ ] interactive mode
- [x] `ls`: list stuff in directories
  - [x] recursive mode
  - [x] `-a/-A`: show all or show everything except `.` and `..`
  - [ ] all the other flags
- [ ] cp
- [ ] mkdir
- [ ] cat
- [ ] sleep

## Other stuff todo
- [ ] Fix error handling in some places
- [ ] Builtin commands need to write UTF-16 text to stdout/stderr on windows I think
- [x] Write more tests 
- [x] Write tests for memory leaks
- [ ] JS template arguments are currently `.protect()`ed and `.unprotect()`ed, but this is slow need a better way to do it
- [x] Shell escape template argument strings 
- [ ] Would be cool to support command substitution like this:
   ```typescript
   const firstCmd = $`ls`
   await $`mv $(${firstCmd}) new_dir/`
   ```

### Lexer/parser
The `src/shell.zig` file contains the lexer and parser for the shell language. This file also contains the v1 blocking and synchronous interpreter which isn't really used right now but might be used for a sync API.

### Shell subprocess
The `src/shell/subproc.zig` file contains a modified version of the `Subprocess` struct for the shell. Originally, it was supposed to share some code with the original `Subprocess`, (which was inside `src/shell/util.zig`) but it became a bit too complicated to abstract it for both because they had different requirements so now they are separate things and `src/shell/util.zig` should probable be deleted.

### Interpreter
The `src/shell/interpreter.zig` file contains the interpreter rewritten to be non-blocking. It also contains the implementation for all the builtin shell commands.

The interpreter works by walking the AST and creating state machine structs for each node. 

For example, for something like `cmd1 && cmd2` it would create a state machine for the binary expression, storing some state like whether its executing the left hand side or the right hand side. It will also create state machines for its children, `cmd1` and `cmd2` which need their own state for all the execution state of managing the subprocess.

When we need to do work that might block like writing to a file descriptor, we schedule the poll and stop doing work in the interpreter, basically suspending execution. Then when the poll is ready it will call a function on whichever state machine it was associated with and resume execution

### Memory management
The lexer/parser allocate tokens/ast nodes into an arena. The AST and arena are both handed to the Interpreter.

The interpreter has to heap allocate state machines, so it uses an allocator different from the AST arena to do that.

Builtin commands, when executed, need to store their `argv` and some environment variables, those are allocated into an arena. The builtin has access to that arena for the duration of its execution if it wants to use it for quick scratch allocations. For example, when the builtin fails it needs to write the error message to stderr so it might use the arena to allocate a formatted string


paperclover:
legendary

src/shell/braces.zig
```diff
@@ -0,0 +1,805 @@
...
+                return v;
+            }
+            return null;
+        }
+    };
+}
+
+/// This may have false positives but it is fast
+pub fn fastDetect(src: []const u8) bool {
+    const Quote = enum { single, double };
```
paperclover, line 101: unused type?

src/shell/braces.zig
```diff
@@ -0,0 +1,805 @@
...
+        for (src) |char| {
+            switch (char) {
+                '{' => {
+                    has_open = true;
+                },
+                '}' => {
+                    has_close = true;
+                },
+            }
+            if (has_close and has_close) return true;
```
paperclover: `has_close and has_close`? is this supposed to be has_close and has_open?

or, what if this was just in the switch itself (twice, one per) checking for the other boolean?

src/bun.js/webcore/streams.zig
```diff
@@ -3761,6 +3761,10 @@ pub const ReadResult = union(enum) {
                     StreamResult{ .owned = bun.ByteList.init(slice) }
                 else if (done)
                     StreamResult{ .into_array_and_done = .{ .len = @as(Blob.SizeType, @truncate(slice.len)), .value = view } }
+                else if (view == .zero and done)
```
paperclover: probably not an issue, but you have this == .zero check twice, it could be a nested if?

src/bun.js/node/types.zig
```diff
@@ -98,6 +98,19 @@ pub fn Maybe(comptime ResultType: type) type {
             };
         }
 
+        pub inline fn initErr(e: Syscall.Error) Maybe(ReturnType) {
```
paperclover, line 82: is initErr and initResult worth it over typing .{ .err = e } directly? its fine but ive always just inlined it (though would be legendary if the decl literals proposal was implemented)

src/shell/braces.zig
```diff
@@ -0,0 +1,805 @@
...
+        }
+        if (std.simd.firstTrue(needles[1] == haystack)) {
+            has_close = true;
+        }
+        if (has_open and has_close) return true;
+    }
+
+    if (i < src.len) {
+        for (src) |char| {
+            switch (char) {
```
Jarred-Sumner, line 141: do you need to check for escapes

src/shell/braces.zig
```diff
@@ -0,0 +1,805 @@
...
+    pub fn init(tokens: []const Token, alloc: Allocator) Parser {
+        return .{
+            .tokens = tokens,
+            .alloc = alloc,
+            .errors = std.ArrayList(Error).init(alloc),
+        };
+    }
+
+    pub fn parse(self: *Parser) !AST.Group {
+        var group_alloc_ = std.heap.stackFallback(@sizeOf(AST.Atom), self.alloc);
```
Jarred-Sumner, line 350: This memory is invalid. We cannot use stackFallback and return values from the array without cloning them to the heap.

src/shell/braces.zig
```diff
@@ -0,0 +1,805 @@
...
+        }
+
+        return self.tokens[self.current + n];
+    }
+
+    fn prev(self: *Parser) Token {
+        return self.tokens[self.current - 1];
+    }
+
+    fn add_error(self: *Parser, comptime fmt: []const u8, args: anytype) !void {
```
Jarred-Sumner, line 479: we're not totally consistent about this, but generally in Zig we should do camelCase function names and snake_cased field names

src/bun.js/bindings/BunObject+exports.h
```diff
@@ -64,6 +65,9 @@
     macro(spawnSync) \
     macro(which) \
     macro(write) \
+    macro($) \
+    macro(shellParse) \
```
paperclover, line 69: are these functions for debugging / internal testing or are these meant to be more public API?

src/shell/braces.zig
```diff
@@ -0,0 +1,805 @@
...
+
+    if (bun.Environment.allow_assert) {
+        for (table.items[0..], 0..) |variant, kdjsd| {
+            _ = kdjsd;
+            std.debug.assert(variant.start != 0 and variant.end != 0);
+        }
+    }
+}
+
+const Encoding = @import("./shell.zig").StringEncoding;
```
Jarred-Sumner, line 594: It would be reasonable to add `bun.shell` and use that instead of re-declaring the import a bunch of times in this file

paperclover (commented):
havent read all the code, some comments on some files.

src/shell/braces.zig
Jarred-Sumner: There are footguns with `@import`: case insensitive filesystems don't work consistently, if we move files around it's tedious to rewrite all the references since that tooling in Zig doesn't exist yet, etc

src/shell/delete_tree.zig
```diff
@@ -0,0 +1,41 @@
+//! Modified version of `std.fs.deleteTree`:
+//! - nonsense instances of `unreachable` removed
+//! - uses Bun's DirIterator
+//! - can pass a Context which allows you to inspect which files/directories have been deleted (needed for shell's implementation of rm with verbose flag)
+const std = @import("std");
+const os = std.os;
+
+const bun = @import("root").bun;
+const DirIterator = @import("../bun.js/node/dir_iterator.zig");
+const Maybe = @import("../bun.js/node/types.zig").Maybe;
```
Jarred-Sumner, line 10: Prefer using `bun.` instead of `@import`, `bun.Maybe` I think works here.

src/shell/delete_tree.zig
```diff
@@ -0,0 +1,41 @@
...
+const Syscall = @import("../sys.zig");
```
Jarred-Sumner, line 11:
```suggestion
const Syscall = bun.sys;
```

src/shell/interpreter.zig
```diff
@@ -0,0 +1,5967 @@
...
+const os = std.os;
+const builtin = @import("builtin");
+const Arena = std.heap.ArenaAllocator;
+const Allocator = std.mem.Allocator;
+const ArrayList = std.ArrayList;
+const JSC = bun.JSC;
+const JSValue = bun.JSC.JSValue;
+const JSPromise = bun.JSC.JSPromise;
+const JSGlobalObject = bun.JSC.JSGlobalObject;
+const which = @import("../which.zig").which;
```
Jarred-Sumner, line 30:
```suggestion
const which = bun.which;
```

src/shell/interpreter.zig
```diff
@@ -0,0 +1,5967 @@
...
+
+const Pipe = [2]bun.FileDescriptor;
+const shell = @import("./shell.zig");
+const Token = shell.Token;
+const ShellError = shell.ShellError;
+const ast = shell.AST;
+
+const GlobWalker = @import("../glob.zig").GlobWalker_(null, true);
+
+pub fn OOM(e: anyerror) noreturn {
```
Jarred-Sumner, line 58:
```suggestion
pub const OOM = bun.outOfMemory;
```

Jarred-Sumner:
Haven't finished reading yet, but I suggest making `$` return a Promise subclass object which keeps an internal reference to the arguments and to the un-executed script. Then, `await` would call `.then` and at that point shell execution would begin.

src/sys.zig
```diff
@@ -662,6 +662,15 @@ pub fn close(fd: bun.FileDescriptor) ?Syscall.Error {
     return bun.FDImpl.decode(fd).close();
 }
 
+pub fn close2(fd: bun.FileDescriptor) ?Syscall.Error {
```
paperclover, line 678: can this be renamed to `closeIgnoreStdio`, and then moved into FDImpl following the logic in `pub fn close` there. though i wonder if the original guard on this should be ignoring stdin too.

with my changes in one branch this snippet will not compile on windows

liz3:
This is a good thing to have build in, giving a shell esque experience.
But it is flawed to implement a subset of the core tools, why?

It might sound appealing to have a cross platform implementations of certain commands but there are a ton of pitfalls.
* most of the commands have a HUGE amount of options which a user would probably expect to work.
* GNU coreutils(used by most linux distros) have more options then the posix version, which are going to be supported?
* What if a user alias'ed a command like ls?
* what about other tools: wc, grep, tail, head, find, less, touch, date, awk, sed....? the list is endless

it makes more sense to use the existing implementation always unless not present, then having a small subset might be useful, but defaulting to a built in(maybe not complete) tool vs decades old established tools might become a big hurdle when people try to port shell scripts to bun shell.

Even though thats a critque i still think this is a great idea - the above, i personally use zx on dotfiles, having it built in the runtime would be cool.

johnpyp:
Could this go under `bun:shell` instead of the main namespace? Generally, I think it would be useful to have a more clear separation of builtin modules.

src/shell/subproc.zig
```diff
@@ -0,0 +1,3999 @@
...
+                return .{ .err = globalThis.throwTODO("spawn() is not yet implemented on Windows") };
+            }
+            var arena = @import("root").bun.ArenaAllocator.init(bun.default_allocator);
+            defer arena.deinit();
+
+            var spawn_args = spawn_args_;
+
+            var out_watchfd: if (Environment.isLinux) ?std.os.fd_t else ?i32 = null;
+
+            const subprocess = switch (spawnMaybeSyncImpl(
```
Jarred-Sumner, line 1262: What is this supposed to do? I'm confused.

Jarred-Sumner:
When running in a debugger, seeing some EBADF:
image

src/result.zig
```diff
@@ -0,0 +1,11 @@
+pub fn Result(comptime T: type, comptime E: type) type {
```
Jarred-Sumner, line 1: Maybe we should combine Maybe and Result and then Maybe is Result with a Mixin applied?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+
+/// Environment strings need to be copied a lot
+/// So we make them reference counted
+///
+/// But sometimes we use strings that are statically allocated, or are allocated
+/// with a predetermined lifetime (e.g. strings in the AST). In that case we
+/// don't want to incur the cost of heap allocating them and refcounting them
+///
+/// So environment strings can be ref counted or borrowed slices
+pub const EnvStr = packed struct {
```
Jarred-Sumner, line 205: nice

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    .kind = .subshell_inherit,
+                    ._buffered_stdout = stdout,
+                    ._buffered_stderr = stderr,
+                    .shell_env = this.shell_env.clone(),
+                    .cmd_local_env = EnvMap.init(allocator),
+                    .export_env = this.export_env.clone(),
+
+                    .__prev_cwd = this.__prev_cwd.clone() catch bun.outOfMemory(),
+                    .__cwd = this.__cwd.clone() catch bun.outOfMemory(),
+                    // TODO probably need to use os.dup here
```
Jarred-Sumner, line 636: you do! up there

did you mean to keep this?

```suggestion

```

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            }
+        }
+
+        fn errored(this: *ThisInterpreter, the_error: ShellError) void {
+            _ = the_error; // autofix
+
+            if (comptime EventLoopKind == .js) {
+                // defer this.deinit();
+                // this.promise.reject(this.global, the_error.toJSC(this.global));
+                _ = this.resolve.call(this.resolve, &[_]JSValue{JSValue.jsNumberFromChar(1)});
```
Jarred-Sumner: does this leak?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    0,
+                )) {
+                    .result => |fd| fd,
+                    .err => |err| {
+                        return Maybe(void).initErr(err);
+                    },
+                };
+                _ = Syscall.close2(this.cwd_fd);
+
+                this.__prev_cwd.clearRetainingCapacity();
```
Jarred-Sumner, line 704: what do these underscores mean?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                jsobj.unprotect();
+            }
+            this.arena.deinit();
+            this.allocator.destroy(this);
+        }
+
+        pub fn setResolve(this: *ThisInterpreter, globalThis: *JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSC.JSValue {
+            _ = globalThis;
+            const value = callframe.argument(0);
+            this.resolve = value;
```
Jarred-Sumner: What is keeping this JSValue alive?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            return .undefined;
+        }
+
+        pub fn setReject(this: *ThisInterpreter, globalThis: *JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSC.JSValue {
+            _ = globalThis;
+            const value = callframe.argument(0);
+            this.reject = value;
```
Jarred-Sumner: What is keeping this JSValue alive?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            // PATH = "";
+
+            while (object_iter.next()) |key| {
+                const keyslice = key.toOwnedSlice(bun.default_allocator) catch bun.outOfMemory();
+                var value = object_iter.value;
+                if (value == .undefined) continue;
+
+                const value_str = value.getZigString(globalThis);
+                const slice = value_str.toOwnedSlice(bun.default_allocator) catch bun.outOfMemory();
+                const keyref = EnvStr.initRefCounted(keyslice);
```
Jarred-Sumner, line 1216: I suggest adding a `EnvStr.fromJS` function

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                done,
+                err,
+            },
+            child_state: union(enum) {
+                idle,
+                cmd_subst: struct {
+                    cmd: *Script,
+                    quoted: bool = false,
+                },
+                // TODO
```
Jarred-Sumner, line 1330: Is this still a todo?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            out: Result,
+            out_idx: u32,
+
+            const ParentPtr = StatePtrUnion(.{
+                Cmd,
+                Assigns,
+            });
+
+            const ChildPtr = StatePtrUnion(.{
+                // Cmd,
```
Jarred-Sumner, line 1345: What dcoes this comment mean?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    .parent = parent,
+
+                    .word_idx = 0,
+                    .state = .normal,
+                    .child_state = .idle,
+                    .out = out_result,
+                    .out_idx = 0,
+                    .current_out = std.ArrayList(u8).init(interpreter.allocator),
+                };
+                // var expansion = interpreter.allocator.create(Expansion) catch bun.outOfMemory();
```
Jarred-Sumner, line 1421:
```suggestion

```

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                            } else {
+                                this.state = .done;
+                            }
+                        },
+                        .glob => {
+                            this.transitionToGlobState();
+                            // yield
+                            return;
+                        },
+                        .done, .err => unreachable,
```
Jarred-Sumner, line 1528: Is this unreachable truly unreachable? Can it be a panic instead?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                }
+
+                return false;
+            }
+
+            /// Remove a set of values from the beginning and end of a slice.
+            pub fn trim(slice: []u8, values_to_strip: []const u8) []u8 {
+                var begin: usize = 0;
+                var end: usize = slice.len;
+                while (begin < end and std.mem.indexOfScalar(u8, values_to_strip, slice[begin]) != null) : (begin += 1) {}
```
Jarred-Sumner, line 1621: why this instead of std.mem.trim?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            pub fn pushCurrentOut(this: *Expansion) void {
+                if (this.current_out.items.len == 0) return;
+                if (this.current_out.items[this.current_out.items.len - 1] != 0) this.current_out.append(0) catch bun.outOfMemory();
+                this.pushResult(&this.current_out);
+                this.current_out = std.ArrayList(u8).init(this.base.interpreter.allocator);
+            }
+
+            pub fn pushResult(this: *Expansion, buf: *std.ArrayList(u8)) void {
+                this.out.pushResult(buf);
+                // if (comptime bun.Environment.allow_assert) {
```
Jarred-Sumner, line 1832: what do these comments mean? if the code should be deleted, can you delete it?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            }
+
+            pub fn appendSlice(this: *Expansion, buf: *std.ArrayList(u8), slice: []const u8) void {
+                _ = this;
+                buf.appendSlice(slice) catch bun.outOfMemory();
+            }
+
+            pub fn pushResultSlice(this: *Expansion, buf: [:0]const u8) void {
+                this.out.pushResultSlice(buf);
+                // if (comptime bun.Environment.allow_assert) {
```
Jarred-Sumner, line 1811: what do these comments mean? if the code should be deleted, can you delete it?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                this.parent.childDone(this, 0);
+            }
+
+            pub fn childDone(this: *Pipeline, child: ChildPtr, exit_code: u8) void {
+                if (comptime bun.Environment.allow_assert) {
+                    std.debug.assert(this.cmds.?.len > 0);
+                }
+
+                const idx = brk: {
+                    const ptr_value: u64 = @bitCast(child.ptr.repr);
```
Jarred-Sumner, line 2703: did you mean to keep this?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+        };
+
+        pub const Pipeline = struct {
+            base: State,
+            node: *const ast.Pipeline,
+            /// Based on precedence rules pipeline can only be child of a stmt or
+            /// conditional
+            parent: ParentPtr,
+            exited_count: u32,
+            cmds: ?[]CmdOrResult,
```
Jarred-Sumner, line 2518: This shouldn't be optional. This should be an empty slice.

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            pipes: ?[]Pipe,
```
Jarred-Sumner, line 2519: This shouldn't be optional. This should be an empty slice.

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                        }
+                    }
+                    this.state = .done;
+                    this.parent.childDone(this, last_exit_code);
+                    return;
+                }
+            }
+
+            pub fn deinit(this: *Pipeline) void {
+                // If commands was zero then we didn't allocate anything
```
Jarred-Sumner, line 2738: what about the pipeline?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    pub fn closed(this: *BufferedIoState) bool {
+                        return this.state == .closed;
+                    }
+                };
+
+                fn deinit(this: *BufferedIoClosed, jsc_vm_allocator: Allocator) void {
+                    if (this.stdin) |*io| {
+                        _ = io; // autofix
+
+                        // io.deinit(jsc_vm_allocator);
```
Jarred-Sumner, line 2864: is this supposed to do nothing?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                                    cmd.base.shell.buffered_stderr().append(bun.default_allocator, readable.pipe.buffer.internal_buffer.slice()) catch bun.outOfMemory();
+                                }
+
+                                stderr.state = .{ .closed = readable.pipe.buffer.internal_buffer };
+                                io.stderr.pipe.buffer.internal_buffer = .{};
+                            }
+                        },
+                        .stdin => {
+                            this.stdin = true;
+                            // if (this.stdin) |*stdin| {
```
Jarred-Sumner, line 2912: Is this supposed to be commented out? Or should it be deleted?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+
+                const HandleIOWrite = struct {
+                    fn run(cmd: *Cmd, bufw: BufferedWriter) void {
+                        cmd.state = .{ .waiting_write_err = bufw };
+                        cmd.state.waiting_write_err.writeIfPossible(false);
+                    }
+                };
+                _ = this.base.shell.writeFailingError(buf, this, HandleIOWrite.run);
+
+                // switch (this.base.shell.io.stderr) {
```
Jarred-Sumner, line 2963: Is this supposed to be commented out?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+            }
+
+            pub fn init(
+                interpreter: *ThisInterpreter,
+                shell_state: *ShellState,
+                node: *const ast.Cmd,
+                parent: ParentPtr,
+                io: IO,
+            ) *Cmd {
+                var cmd = interpreter.allocator.create(Cmd) catch |err| {
```
Jarred-Sumner, line 2995:
```suggestion
                var cmd = interpreter.allocator.create(Cmd) catch bun.outOfMemory()
```

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                            return;
+                        },
+                        .waiting_write_err => {
+                            return;
+                        },
+                        .exec => {
+                            // yield execution to subproc/builtin
+                            return;
+                        },
+                        .done, .err => unreachable,
```
Jarred-Sumner, line 3101: Should this be unreachable? Or just panic?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    .err => |e| {
+                        global_handle.get().actuallyThrow(e);
+                        return;
+                    },
+                };
+                subproc.ref();
+                this.spawn_arena_freed = true;
+                arena.deinit();
+
+                // if (this.cmd.stdout == .pipe and this.cmd.stdout.pipe == .buffer) {
```
Jarred-Sumner, line 3375: We don't need to watch the pipe?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                }
+                // if (this.exit_code != null) {
+                //     if (this.cmd) |cmd| {
+                //         _ = cmd.tryKill(9);
+                //         cmd.unref(true);
+                //         cmd.deinit();
+                //     }
+                // }
+
+                // if (this.cmd) |cmd| {
```
Jarred-Sumner, line 3457: should these comments be deleted?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                this.exit_code = exit_code;
+
+                const has_finished = this.hasFinished();
+                if (has_finished) {
+                    this.state = .done;
+                    this.next();
+                    return;
+                    // this.parent.childDone(this, exit_code);
+                }
+                // } else {
```
Jarred-Sumner, line 3436: should these comments be deleted?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                }
+
+                pub fn usageString(this: Kind) []const u8 {
+                    return switch (this) {
+                        .@"export" => "",
+                        .cd => "",
+                        .echo => "",
+                        .pwd => "",
+                        .which => "",
+                        .rm => "usage: rm [-f | -i] [-dIPRrvWx] file ...\n       unlink [--] file\n",
```
Jarred-Sumner, line 3584: colors?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+
+            pub fn deinit(this: *Builtin) void {
+                this.callImpl(void, "deinit", .{});
+
+                _ = Syscall.close(this.cwd);
+                this.stdout.deinit();
+                this.stderr.deinit();
+                this.stdin.deinit();
+
+                // this.arena.deinit();
```
Jarred-Sumner, line 3987: Should this be deleted?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    .result => {},
+                }
+
+                return Maybe(void).success;
+            }
+
+            pub fn deinit(this: *Builtin) void {
+                this.callImpl(void, "deinit", .{});
+
+                _ = Syscall.close(this.cwd);
```
Jarred-Sumner: Is this file descriptor always valid?

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    /// End each output line with NUL, not newline
+                    end_with_nul: bool = false,
+
+                    /// `-1`
+                    /// List one file per line
+                    one_file_per_line: bool = false,
+
+                    /// `--help`
+                    /// Display help and exit
+                    show_help: bool = false,
```
Jarred-Sumner, line 5378: nice

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+
+                    /// From the man pages of `mv`:
+                    /// ```txt
+                    /// As the rename(2) call does not work across file systems, mv uses cp(1) and rm(1) to accomplish the move.  The effect is equivalent to:
+                    ///     rm -f destination_path && \
+                    ///     cp -pRP source_file destination && \
+                    ///     rm -rf source_file
+                    /// ```
+                    fn moveAcrossFilesystems(this: *@This(), src: [:0]const u8, dest: [:0]const u8) void {
+                        _ = this;
```
Jarred-Sumner, line 5713: maybe you can use `bun.C.copyFileZSlowWithHandle`

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+                }
+
+                break :brk;
+            }
+
+            switch (file) {
+                .atom => {
+                    // FIXME TODO expand atom
+                    // if expands to multiple atoms, throw "ambiguous redirect" error
+                    @panic("FIXME TODO redirect builtin");
```
Jarred-Sumner: this should not panic

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+                        if (cmd.redirect.stdout) {
+                            stdout = builtinio;
+                        }
+
+                        if (cmd.redirect.stderr) {
+                            stderr = builtinio;
+                        }
+                    } else if (self.jsobjs[file.jsbuf.idx].as(JSC.WebCore.Blob)) |blob| {
+                        _ = blob;
+                        @panic("FIXME TODO HANDLE BLOB");
```
Jarred-Sumner: please change this panic to an error

```js
globalThis.throwTODO("TODO: handle blob")
```

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+                    } else {
+                        @panic("FIXME TODO Unhandled");
```
Jarred-Sumner: please change this to throw an error

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+            // `export` allows arguments to be empty, which just skips it.
+            // For example `export $FOO` will do nothing
+            if (arg_len == 0) {
+                continue;
+            }
+            const arg: [:0]const u8 = arg_cstr[0..arg_len :0];
+            const idx = std.mem.indexOfScalar(u8, arg[0..arg.len], '=') orelse {
+                if (!isValidVarName(arg)) {
+                    try bltn.write_err(&bltn.stderr, .@"export", "`{s}`: not a valid identifier", .{arg});
+                    self.globalThis.throwInvalidArguments("Invalid variable name: \"{s}\"", .{arg});
```
Jarred-Sumner: does this leak memory?

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+            self.cwd[0..self.cwd.len],
+            dot,
+            absolute,
+            follow_symlinks,
+            error_on_broken_symlinks,
+            only_files,
+        )) {
+            .result => {},
+            .err => {
+                @panic("FIXME TODO handle error properly");
```
Jarred-Sumner: please change this to a runtime error

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+
+        return 0;
+    }
+
+    /// FIXME TODO error code is not u8 this will cause @intCast truncate bits panic
+    fn rmDir(this: *Rm, absolute_path: [:0]const u8, path: []const u8) !u8 {
+        log("[rm] delete: {s}", .{absolute_path});
+        // FIXME TODO handle this better
+        try this.verboseDelete(path);
+        const result = std.os.system.rmdir(absolute_path);
```
Jarred-Sumner: please don't use`std.os` 

please use `bun.sys.rmdir`

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+        };
+    }
+
+    fn verboseDelete(this: *Rm, path: []const u8) !void {
+        if (this.opts.verbose) {
+            try this.bltn.write_fmt(&this.bltn.stdout, "{s}\n", .{path});
+        }
+    }
+
+    fn isRoot(this: *Rm, path: []const u8) bool {
```
Jarred-Sumner: This function can be:

```zig
return std.fs.path.dirname(path) == null;
```

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+        }
+
+        if (path.len > bun.MAX_PATH_BYTES) {
+            try this.bltn.write_err(&this.bltn.stderr, .rm, "{s}: File name too long", .{path});
+            return 1;
+        }
+
+        const absolute_path = this.preparePath(path);
+
+        if (this.isRoot(absolute_path) and this.opts.preserve_root) return 0;
```
Jarred-Sumner: Can you delete all of this code and just use JSC.Node.NodeFS? Why add another implementation of `fs.rm`?

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
+const bun = @import("root").bun;
```
Jarred-Sumner: This file has a lot of dead code in it.

src/shell/interpreter_sync.zig
```diff
@@ -0,0 +1,1832 @@
...
+            comptime fmt: []const u8,
+            args: anytype,
+        ) !void {
+            switch (io.*) {
+                .fd => |fd| {
+                    var stack_alloc = std.heap.stackFallback(256, this.args.allocator);
+                    const len = std.fmt.count(fmt, args);
+                    const buf = try stack_alloc.get().alloc(u8, len);
+                    _ = try std.fmt.bufPrint(buf, fmt, args);
+                    _ = try std.os.write(fd, buf);
```
Jarred-Sumner: Do not use std.os

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                    parse_opts: struct {
+                        args_slice: []const [*:0]const u8,
+                        idx: u32 = 0,
+                        state: union(enum) {
+                            normal,
+                            wait_write_err: BufferedWriter,
+                        } = .normal,
+                    },
+                    exec: struct {
+
```
Jarred-Sumner, line 6928: Please delete all of this code and use the node fs rm implementation instead.

src/shell/interpreter.zig
```diff
@@ -0,0 +1,7766 @@
...
+                            &this.io,
+                            false,
+                        );
+                        if (coro_result == .yield) return;
+
+                        if (comptime bun.Environment.allow_assert) {
+                            std.debug.assert(this.exec == .bltn);
+                        }
+
+                        log("WTF: {s}", .{@tagName(this.exec)});
```
Jarred-Sumner, line 3223: Did you mean to keep this?

src/bun.js/api/bun.zig
```diff
@@ -525,6 +527,54 @@ pub fn shell(
...
+        return .undefined;
+    }
+
+    const jsval = callframe.argument(0);
+    const bunstr = jsval.toBunString(globalThis);
+    defer bunstr.deref();
+
+    var outbuf = std.ArrayList(u8).init(bun.default_allocator);
+    defer outbuf.deinit();
+    if (bunstr.isUTF8()) {
```
Jarred-Sumner, line 545: this wil always be false.

`isUTF8` only works for ZigString.

you want `.toSlice(bun.default_allocator)` and then check if it is allocated.

malezjaa (commented):
Looks good to me

Jarred-Sumner:
🎉

PR #7749, opened 2023-12-20
Brotli decompression using JS implementation from Google (#267)

vlechemin:
### What does this PR do?

It implements brotli decompression (#267) in JS using:
https://github.com/google/brotli/blob/master/js/decode.js

The file size is pretty big, I'm not sure of the impact on performance. If the impact is too high for those not using brotli, we could remove the dictionary, it is supposed to work anyway (https://en.wikipedia.org/wiki/Brotli#cite_note-14).

Pushing decoded `chunks` in the `transform` method does not work but I didnt investigate why because this version should be equivalent speed wise. The test case is pretty big so I did not include it.

Thanks to @paperdave for his help on Discord.

- [X] Code changes

- [X] I ran my tests locally and they pass (`bun-debug test test/js/node/zlib/zlib.test.js`)


Jarred-Sumner:
Thank you for this. I really appreciate the time you spent on this PR, but I have some bad news.

This is something that's very performance sensitive, which means it must be implemented in native code. Ideally using the official library.

Issue #7567, opened 2023-12-10
Holey arrays appearing incorrect in console.log


mostik:
### What version of Bun is running?

1.0.16+800fb1290

### What platform is your computer?

Linux 6.1.64-1-MANJARO x86_64

### What steps can reproduce the bug?

```js
let a = [1, , ,2]

console.log(a)
```

### What is the expected behavior?

```js
[ 1, 2 x empty items, 2 ]
```

### What do you see instead?

```js
[ 12 x empty items, 2 ]
```

paperclover:
~~Fixed in 1.0.16~~ nevermind

kaioduarte:
Seems fixed on 1.0.20

Electroid:
Closed by #7751

PR #7751, opened 2023-12-20
Handle holey arrays console log edge case

ayymart:
Closes #7567

", " was only being written when appropriate before "empty item", and not "N x empty items".

### What does this PR do?

This fixes the aforementioned bug, and adds a line to the corresponding JS test.

- [x] Code changes

If Zig files changed:

### DIFF
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -2348,19 +2348,18 @@ pub const ZigConsoleClient = struct {
                             }
 
                             if (empty_start) |empty| {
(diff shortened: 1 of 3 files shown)


ayymart:
Hi Oven devs! Lorena pointed me at #7567.

I had to `brew install icu4c`, which wasn't in the contributor guide.

dylan-conway (approved):
Looks great

dylan-conway:
Seems like we need to add `icu4c` to the list of required dependencies in our docs

PR #7752, opened 2023-12-20
Fixes #7736

### LINKED ISSUE #7736: Large maps that are equal aren't comparing equal

1.0.18+36c316a24

Darwin 23.1.0 arm64 arm

```ts
//
// NOTE:
// - At <= 912 elements, the test always passes
// - At > 912 elements, the test starts failing significantly more (though not _always_)
//
const MAP_SIZE = 913;

function* genpairs() {
  for (let i = 0; i < MAP_SIZE; i++) {
    yield ["k" + i, 1] as const;
  }
}

test("bug", () => {
  const x = new Map(genpairs());
  const y = new Map(genpairs());
  expect(x).toEqual(y);
});
```

No matter how large the maps are that are being compared, I'd expect this statement to not fail as long as their contents are equal.

### What do you see instead?

The test starts failing most of the time when comparing maps larger than 912 elements.

### Additional information

Here is a repo with a clean/minimal replication:

https://github.com/nvie/clean-project/tree/map-equals-bug (not `main` branch)

Jarred-Sumner:
### What does this PR do?

- Fixes #7736
- Makes `expect(map1).toEqual(map2)` something like 50x faster

Regression tests

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -294,7 +294,8 @@ AsymmetricMatcherResult matchAsymmetricMatcherAndGetFlags(JSGlobalObject* global
                         JSValue otherValue = otherArray->getIndex(globalObject, n);
                         ThrowScope scope = DECLARE_THROW_SCOPE(globalObject->vm());
                         Vector<std::pair<JSValue, JSValue>, 16> stack;
-                        if (Bun__deepEquals<false, true>(globalObject, expectedValue, otherValue, stack, &scope, true)) {
+                        MarkedArgumentBuffer gcBuffer;
+                        if (Bun__deepEquals<false, true>(globalObject, expectedValue, otherValue, gcBuffer, stack, &scope, true)) {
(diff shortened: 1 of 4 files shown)


src/bun.js/bindings/bindings.cpp
```diff
@@ -524,8 +533,59 @@ bool Bun__deepEquals(JSC__JSGlobalObject* globalObject, JSValue v1, JSValue v2,
...
+                    if (bucket2ptr && (*bucket2ptr)->deleted()) {
+                        bucket2ptr = nullptr;
+                    }
+
+                    if (!bucket2ptr) {
+                        auto findDeepEqualKey = [&]() -> bool {
+                            auto* bucket = set2->head();
+                            while (bucket) {
+                                if (!bucket->deleted()) {
+                                    auto key2 = bucket->key();
```
dylan-conway, line 557: RETURN_IF_EXCEPTION(*scope, false);

dylan-conway (approved):

PR #7758, opened 2023-12-21
ci: run windows tests and also run them concurrently

paperclover:
The `bun-internal-test` runner is available under `bun run test` in our repository. It runs tests concurrently and buffers output into a report file `test-report.md`

https://github.com/oven-sh/bun/assets/24465214/b58f43b6-e357-4c06-8a92-14ad0b26d482

This is not limited to windows, but is primarily helpful for windows.

When testing the windows port, files can be marked as "known failure" with a comment
```
// @known-failing-on-windows: 2 failing
```

where the text after the comma is the failing reason.

On windows, you can run `bun run update-known-failing` to take the previous test runner's result and use it to update files' comments so that they are no longer marked as regressions.

This also disables MacOS Baseline testing CI until those test failures are fixed. Their output is too noisy and are what I blame for recent PR merges that ignore failing CI, breaking Windows compilation.


packages/bun-internal-test/src/runner.node.mjs
```diff
@@ -1,41 +1,53 @@
...
+let force_ram_size = Number(BigInt(nativeMemory) >> BigInt(2)) + "";
+if (!(Number.isSafeInteger(force_ram_size_input) && force_ram_size_input > 0)) {
+  force_ram_size = force_ram_size_input + "";
 }
 
 const cwd = resolve(fileURLToPath(import.meta.url), "../../../../");
 process.chdir(cwd);
 
-const isAction = !!process.env["GITHUB_ACTION"];
+const ci = !!process.env["GITHUB_ACTION"];
```
dylan-conway:
```suggestion
const ci = !!process.env["GITHUB_ACTIONS"];
```

packages/bun-internal-test/src/runner.node.mjs
```diff
@@ -50,103 +62,368 @@ function* findTests(dir, query) {
...
       timeout: 1000 * 60 * 3,
       env: {
         ...process.env,
         FORCE_COLOR: "1",
         BUN_GARBAGE_COLLECTOR_LEVEL: "1",
-        BUN_JSC_forceRAMSize,
+        BUN_JSC_forceRAMSize: force_ram_size,
         BUN_RUNTIME_TRANSPILER_CACHE_PATH: "0",
         // reproduce CI results locally
         GITHUB_ACTION: process.env.GITHUB_ACTION ?? "true",
```
dylan-conway:
```suggestion
        GITHUB_ACTIONS: process.env.GITHUB_ACTIONS ?? "true",
```

test/cli/install/__snapshots__/semver.test.ts.snap
dylan-conway, line 1: the snapshot bug was triggered and made this file way larger than it should be. If this happens with every run we should probably fix it before this merges. I think jarred has a branch with a fix

test/js/bun/util/__snapshots__/inspect-error.test.js.snap
dylan-conway, line 1: snapshot bug also happened here

test/regression/issue/__snapshots__/03830.test.ts.snap
dylan-conway, line 1: not sure if this is a bug or not, but I don't think this file should be new

dylan-conway (changes requested):
Needs snapshot files truncated

PR #7759, opened 2023-12-21
fix MacOS build when OpenSSL is also installed

zack466:
### What does this PR do?

I was building bun for the first time, and I came across an issue building on MacOS. I'm on Sonoma 14.0 using MacPorts.
I was hit with a bunch of build errors when running `bun run setup` that looked like the following:
```
...
In file included from /Users/zack4/bun/src/bun.js/bindings/KeyObject.cpp:40:
In file included from /Users/zack4/bun/src/deps/boringssl/include/openssl/mem.h:60:
/Users/zack4/bun/src/deps/boringssl/include/openssl/base.h:343:13: error: typedef redefinition with different types ('int' vs 'struct crypto_threadid_st')
typedef int CRYPTO_THREADID;
            ^
/opt/local/include/openssl/crypto.h:309:3: note: previous definition is here
} CRYPTO_THREADID;
  ^
In file included from /Users/zack4/bun/src/bun.js/bindings/KeyObject.cpp:40:
In file included from /Users/zack4/bun/src/deps/boringssl/include/openssl/mem.h:60:
/Users/zack4/bun/src/deps/boringssl/include/openssl/base.h:347:29: error: typedef redefinition with different types ('struct asn1_null_st' vs 'int')
typedef struct asn1_null_st ASN1_NULL;
                            ^
/opt/local/include/openssl/types.h:74:13: note: previous definition is here
typedef int ASN1_NULL;
            ^
...
```
during the final stage in which it actually built the bun executable.

I tracked the error down to the fact that when `KeyObject.cpp` included `openssl`, it seems to have ended up including both `boringssl` headers and `openssl` headers (since `openssl` is installed on my system), and I assume these two libraries are completely incompatible.

The reason that this even happened seems to be that the `CMakeLists.txt` file adds `/opt/local/include` (where library headers are installed by MacPorts) to the list of included directories while it was adding `libiconv` to the build (see the diff). The openssl headers are also in this directory, so I think that's what led to the conflicts. I removed the line in which this happens, and build succeeded. It doesn't seem like `iconv.h` is included directly by any files in the bun source code, so I guess it's fine?

I'm not sure if the error is specific to MacPorts or would also happen with Homebrew if OpenSSL is installed system-wide. Anyway, I figured that it wasn't worth opening an issue for a one-line change, so I just opened this pull request. Let me know what you think.

Thanks, Zack


paperclover (approved):

jdalton:
Let's see how it goes. I'm on OSX so let's see :)

CMakeLists.txt
```diff
@@ -1055,7 +1055,6 @@ if(APPLE)
     target_link_libraries(${bun} PRIVATE "${ICU4C_DIR}/lib/libicudata.a")
     target_link_libraries(${bun} PRIVATE "${ICU4C_DIR}/lib/libicui18n.a")
     target_link_libraries(${bun} PRIVATE "${ICU4C_DIR}/lib/libicuuc.a")
-    include_directories(${ICU4C_DIR}/include)
```
jdalton, line 1089:
```suggestion
    include(${ICU4C_DIR}/include/iconv.h)
```

jdalton: On second thought maybe the code suggestion of the specific header to include... we can do this for other areas using include_directories too.

zack466: Unfortunately, I don't think CMake allows you to include just a single header file? I'm pretty sure the CMake `include` command is used to include CMake code from a different file, not add a header file to the build. Making the change and then running `bun run build` gives an error like:
```
CMake Error at /opt/local/include/iconv.h:1:
  Parse error.  Expected a command name, got unquoted argument with text
  "/*".
Call Stack (most recent call first):
  CMakeLists.txt:1089 (include)
```
since a header file is obviously not valid CMake code. Based on some googling, I think that the only thing we can do is control is the order in which directories get included (by actually reordering the `include_directories` or using BEFORE).

paperclover: we shouldnt do this change. `include` is fine

Issue #7685, opened 2023-12-15
node-addon-api cannot receive more than 6 args


kkocdko:
### What version of Bun is running?

1.0.18+36c316a24

### What platform is your computer?

Fedora 39 Linux 6.6.4-200.fc39.x86_64

### What steps can reproduce the bug?

The minimal reproducable example [here](https://github.com/kkocdko/utils4linux/tree/master/napibun):

```cpp
#define NODE_ADDON_API_DISABLE_DEPRECATED
#include <napi.h>

#include <cassert>

Napi::Value Foo(const Napi::CallbackInfo& info) {
  Napi::Env napiEnv(info.Env());
  Napi::HandleScope scope(napiEnv);
  #define napi_assert(expr) {if(!expr){Napi::Error::New(napiEnv, #expr).ThrowAsJavaScriptException();}}
  napi_assert(info[0].IsNumber());
  napi_assert(info[1].IsNumber());
  napi_assert(info[2].IsNumber());
  napi_assert(info[3].IsNumber());
  napi_assert(info[4].IsNumber());
  napi_assert(info[5].IsNumber());
  napi_assert(info[6].IsNumber());
  napi_assert(info[7].IsNumber());
  return Napi::String::New(napiEnv, "success");
}

Napi::Object Init(Napi::Env env, Napi::Object exports) {
  exports.Set(Napi::String::New(env, "foo"),
              Napi::Function::New(env, Foo));
  return exports;
}

NODE_API_MODULE(addon, Init)
```

```js
const native = require("./dist/native.node");
const args = [...Array(20).keys()];
console.log(JSON.stringify(args));
console.log(native.foo(...args));
```

Test in [GitHub Actions here](https://github.com/kkocdko/utils4linux/actions/runs/7225228142/job/19688194811).

### What is the expected behavior?

Like Node.js.

### What do you see instead?

The 7th argument `info[6].IsNumber()` assert failed.

```
> bun
[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19]
1 | const native = require("./dist/native.node");
2 | const args = [...Array(20).keys()];
3 | console.log(JSON.stringify(args));
4 | console.log(native.foo(...args));
                ^
error: info[7].IsNumber()
      at /home/runner/work/utils4linux/utils4linux/napibun/main.js:4:13

> node
[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19]
success
```

### Additional information

[...]

Issue #6902, opened 2023-11-04
Bcrypt Compare Segmentation Fault


ansgarlichter:
### What version of Bun is running?

1.0.7

### What platform is your computer?

Microsoft Windows NT 10.0.22621.0 x64 using WSL Ubuntu 22.10

### What steps can reproduce the bug?
I am using a Nest.js Project and I am trying to migrate to Bun as runtime. The projects uses currently bcrypt to hash the passwords and compare the hashes during the validation of the login process or during registering.

The project can be found here: [nestjs-api](https://github.com/AnsgarLichter/nestjs-api/tree/migrate-to-bun).

Clone the project, execute ``bun i``, run the server with ``bun run src/main.ts``. Now send a post request to the endpoint to register an user (localhost:3000/api/v1/auth/register) with the following body:
```json
{
    "name": "test2",
    "username": "test2",
    "password": "test123",
    "roles": "admin",
    "email": "test2@tester.de",
    "isAccountDisabled": false
}
```
Now 

### What is the expected behavior?

The bcrypt functions to compare and hash a password should work without problems.

### What do you see instead?

Everytime bcrypt.compare is called the server stops with the message "Segmentation Fault":
const match = await compare(pass, user.password);
![image](https://github.com/oven-sh/bun/assets/43957906/66a6d1cc-d811-4e9d-8ce2-7ebd51f3de4b)

The same happens registering a new user. Every time the password is hashed the server crashes with "Segmentation Fault"
user.password = await hash(input.password, 10);

### Additional information

I know I can use the functions from Bun to hash a password (see [docs](https://bun.sh/guides/util/hash-a-password)) Shouldn't bcrypt work for compatibility reasons with current Node.js projects to be a drop-in replacement? Or is this just not possible because bcrypt is a node binding written in C++?

Jarred-Sumner:
Duplicate of https://github.com/oven-sh/bun/issues/3201

Issue #5848, opened 2023-09-21
NAPI Argon2 Segfault


JoeyEamigh:
### What version of Bun is running?

1.0.3 (main branch ac19c7c)

### What platform is your computer?

Linux 6.5.3-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

1. `git clone https://github.com/ranisalt/node-argon2.git`
2. `bun install`
3. `bun run --bun mocha test/test.js`

### What is the expected behavior?

No segfault!

### What do you see instead?

```
[napi] napi_get_typedarray_info
[napi] napi_get_typedarray_info
[napi] napi_get_buffer_info
[napi] napi_get_typedarray_info
[napi] napi_get_typedarray_info
[napi] napi_get_buffer_info
[napi] napi_coerce_to_number
[napi] napi_get_value_uint32
[napi] napi_coerce_to_number
[napi] napi_get_value_uint32
[napi] napi_coerce_to_number
[napi] napi_get_value_uint32
[napi] napi_coerce_to_number
[napi] napi_get_value_uint32
[napi] napi_coerce_to_number
[napi] napi_get_value_uint32
[napi] napi_coerce_to_number
[napi] napi_get_value_int32
[napi] napi_create_string_latin1
[napi] napi_create_async_work
[napi] napi_queue_async_work
[Loop] ref
[napi] napi_get_undefined
[Loop] ref
[napi] napi_open_handle_scope
[napi] napi_open_handle_scope
[napi] napi_get_undefined
[napi] napi_create_buffer_copy: 32
[napi] napi_get_typedarray_info
[napi] napi_get_typedarray_info
[napi] napi_open_escapable_handle_scope
[napi] napi_get_undefined
[napi] napi_escape_handle
[napi] napi_close_escapable_handle_scope
[napi] napi_close_handle_scope
[napi] napi_close_handle_scope
[napi] napi_delete_async_work
[Loop] unref
[Loop] unref
[uws] Timer.deinit()
      ✔ hash with argon2i
error: "mocha" exited with code 11 (SIGSEGV)
```

### Additional information

I am not sure the best way to debug zig as I have never used it before, but I am happy to be of any assistance I can!

JoeyEamigh:
I am not sure if this is very important as one can just use the hashing functions built into the bun runtime, but it may be worth looking at in case there is something bigger going on here!

v3nntrnx:
Duplicate of #2550?

JoeyEamigh:
[...]

Issue #3201, opened 2023-06-04
bcrypt async functions segfault


Jarred-Sumner:
Repro:
```js
mkdir bcrypty
bun init -y 
npm add bcrypt # because node-gyp
```

Code:
```js
const bcrypt = require("bcrypt");
const saltRounds = 10;
const myPlaintextPassword = "s0//P4$$w0rD";

bcrypt.hash(myPlaintextPassword, saltRounds, function (err, hash) {
  console.log(hash);
});

```

<img width="605" alt="image" src="https://github.com/oven-sh/bun/assets/709451/3270913f-a008-403a-b9ea-781f81438077">

<details>

<summary>debug logs</summary>

```js
❯ bun-debug index.ts
[SYS] read(3, 4096) = 4096 (0.047ms)
[SYS] close(3)
[fs] openat(11, /Users/jarred/Build/bcrypty/package.json) = 12
[SYS] close(12)
[fs] openat(11, /Users/jarred/Build/bcrypty/tsconfig.json) = 12
[SYS] close(12)
[SYS] close(5)
[SYS] close(6)
[SYS] close(9)
[SYS] close(10)
[SYS] close(11)
[fs] openat(0, /Users/jarred/Build/bcrypty/index.ts) = 9
[fs] openat(11, /Users/jarred/Build/bcrypty/node_modules/bcrypt/package.json) = 12
[SYS] close(12)
[SYS] close(10)
[SYS] close(11)
[fs] openat(0, /Users/jarred/Build/bcrypty/node_modules/bcrypt/bcrypt.js) = 10
[fs] openat(12, /Users/jarred/Build/bcrypty/node_modules/@mapbox/node-pre-gyp/package.json) = 13
[SYS] close(13)
[SYS] close(11)
[SYS] close(12)
[SYS] close(11)
[fs] openat(0, /Users/jarred/Build/bcrypty/node_modules/@mapbox/node-pre-gyp/lib/node-pre-gyp.js) = 11
[SYS] close(12)
[fs] openat(12, /Users/jarred/Build/bcrypty/node_modules/nopt/package.json) = 13
[SYS] close(13)
[SYS] close(12)
[SYS] close(12)
[fs] openat(12, /Users/jarred/Build/bcrypty/node_modules/npmlog/package.json) = 13
[SYS] close(13)
[SYS] close(12)
[...]

PR #7765, opened 2023-12-21
NAPI fixes

Jarred-Sumner:
### What does this PR do?

- Fixes https://github.com/oven-sh/bun/issues/5848
- Fixes https://github.com/oven-sh/bun/issues/3201
- Fixes https://github.com/oven-sh/bun/issues/6902
- Fixes https://github.com/oven-sh/bun/issues/299
- Fixes https://github.com/oven-sh/bun/issues/2550
- Fixes https://github.com/oven-sh/bun/issues/7685

Also adds some of the missing aliases in expect, like `expect(...).toThrowError`

Ran tests in argon2 and bcrypt

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -1687,12 +1687,7 @@ JSC_DEFINE_HOST_FUNCTION(NapiClass_ConstructorFunction,
 
     auto result = napi->constructor()(globalObject, reinterpret_cast<JSC::CallFrame*>(NAPICallFrame::toNapiCallbackInfo(frame)));
     RETURN_IF_EXCEPTION(scope, {});
-
-    if (!result) {
-        return JSValue::encode(frame.thisValue());
-    }
-
-    RELEASE_AND_RETURN(scope, result);
+    RELEASE_AND_RETURN(scope, frame.thisValue());
 }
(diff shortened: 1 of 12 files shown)


src/bun.js/bindings/napi.cpp
```diff
@@ -1502,65 +1677,17 @@ JSC_DEFINE_HOST_FUNCTION(NapiClass_ConstructorFunction,
...
-
-    if (count > 6) {
-        for (size_t i = 6; i < count; i++) {
-            args.append(callFrame->uncheckedArgument(i));
-        }
-    }
+    MarkedArgumentBufferWithSize<12> args;
+    size_t argc = callFrame->argumentCount() + 1;
+    args.fill(vm, argc, [&](auto* slot) {
+        memcpy(slot, callFrame->addressOfArgumentsStart() - 1, sizeof(JSC::JSValue) * argc);
```
cirospaciari: we do the -1 offset to copy this into slot?

Jarred-Sumner: yeah i should write a comment here for this

its the `this` value's slot

cirospaciari (approved):
LGTM do we already have a `bcrypt` third-party test? would be to have one

PR #7766, opened 2023-12-21
Fixes #7740: Change "Duplicate dependency" to warn instead of error

### LINKED ISSUE #7740: Change "Duplicate dependency" to be a warning instead of error

1.0.18

Darwin 23.2.0 arm64 arm

I am trying to replace all `npm install --no-save` calls with `bun install --no-save` and it fails for a project because the dev declared it in `dependencies` and `devDependencies`

```
{
    "devDependencies": {
        "vuex": "3.6.2"
    },
    "dependencies": {
        "vuex": "3.6.2"
    }
}
```

With `npm` it runs. I know the dev made a mistake here. But maybe this should be a warning. I have to annoy the team and wait for a release. I think making it to a warning could save people time. 

Runs like `npm` but shows a warning that the package is defined in both.

### What do you see instead?

```
❯ bun i
bun install v1.0.18 (36c316a2)
3 |         "vuex": "3.6.2"
            ^
error: Duplicate dependency: "vuex" specified in package.json
    at /Users/shyim/Downloads/foo/package.json:3:9

6 |         "vuex": "3.6.2"
                    ^
note: "vuex" originally specified here
   at /Users/shyim/Downloads/foo/package.json:6:17
```

### Additional information

_No response_

knightspore:
### What does this PR do?

- Fixes #7740 
- Add `addRangeWarningFmtWithNotes()` to `src/logger.zig`

- Added a regression test covering this behaviour (`test/regression/issue/07740.test.ts`)
- Ran all regression tests to identify potential conflicts (none spotted)

### DIFF
--- a/src/logger.zig
+++ b/src/logger.zig
@@ -1115,6 +1115,16 @@ pub const Log = struct {
         });
     }
 
+    pub fn addRangeWarningFmtWithNotes(log: *Log, source: ?*const Source, r: Range, allocator: std.mem.Allocator, notes: []Data, comptime fmt: string, args: anytype) !void {
+        @setCold(true);
+        log.errors += 1;
+        try log.addMsg(.{
+            .kind = .warn,
+            .data = try rangeData(source, r, allocPrint(allocator, fmt, args) catch unreachable).cloneLineText(log.clone_line_text, log.msgs.allocator),
+            .notes = notes,
+        });
+    }
+
     pub fn addRangeErrorFmtWithNote(
         log: *Log,
         source: ?*const Source,
--- a/src/logger.zig
+++ b/src/logger.zig
@@ -1117,7 +1117,7 @@ pub const Log = struct {
 
     pub fn addRangeWarningFmtWithNotes(log: *Log, source: ?*const Source, r: Range, allocator: std.mem.Allocator, notes: []Data, comptime fmt: string, args: anytype) !void {
         @setCold(true);
-        log.errors += 1;
+        log.warnings += 1;
         try log.addMsg(.{
             .kind = .warn,
             .data = try rangeData(source, r, allocPrint(allocator, fmt, args) catch unreachable).cloneLineText(log.clone_line_text, log.msgs.allocator),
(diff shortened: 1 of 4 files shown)


knightspore:
I may need a second pair of eyes on:

>  I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be

Since I'm unfamiliar with Zig and don't want to provide a false confirmation.

knightspore:
I'm also not sure if it's preferable that I added `addRangeWarningFmtWithNotes` (to provide similar usage to `addRangeErrorFmtWithNotes`):
https://github.com/oven-sh/bun/blob/3fed31573cfbbd5dcd5401b78bf72e59f36abbf0/src/logger.zig#L1118-L1126

Or if I should instead be using the existing `addRangeWarningFmtWithNote`:
https://github.com/oven-sh/bun/blob/3fed31573cfbbd5dcd5401b78bf72e59f36abbf0/src/logger.zig#L1093-L1116

src/logger.zig
```diff
@@ -1115,6 +1115,16 @@ pub const Log = struct {
         });
     }
 
+    pub fn addRangeWarningFmtWithNotes(log: *Log, source: ?*const Source, r: Range, allocator: std.mem.Allocator, notes: []Data, comptime fmt: string, args: anytype) !void {
+        @setCold(true);
+        log.errors += 1;
```
Jarred-Sumner:
```suggestion
        log.warnings += 1;
```

test/regression/issue/07740.test.ts
```diff
@@ -0,0 +1,29 @@
+import { bunEnv } from "harness";
+import { bunExe } from "harness";
+import { tempDirWithFiles } from "harness"
+
+it("duplicate dependencies should warn instead of error", () => {
+    const package_json = JSON.stringify({
+        devDependencies: {
+            vuex: "3.6.2"
```
Jarred-Sumner: Let's use a simpler package to avoid increasing test times
```suggestion
            "empty-package-for-bun-test-runner": "1.0.0"
```

test/regression/issue/07740.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        },
+        dependencies: {
+            vuex: "3.6.2"
```
Jarred-Sumner: Let's use a simpler package to avoid increasing test times
```suggestion
            "empty-package-for-bun-test-runner": "1.0.0"
```

Jarred-Sumner:
Thank you

PR #7772, opened 2023-12-21
feat: add `toBeEmptyObject`

coratgerl:
### What does this PR do?

Add `toBeEmptyObject` from https://github.com/oven-sh/bun/issues/1825

- [x] Code changes

- I wrote automated tests 


src/bun.js/test/expect.zig
```diff
@@ -113,6 +113,31 @@ pub const Expect = struct {
         return received ++ matcher_name ++ "<d>(<r>" ++ args ++ "<d>)<r>";
     }
 
+    pub fn isObjectEmpty(globalThis: *JSGlobalObject, value: JSValue) bool {
```
dylan-conway: This should be moved to `bindings.zig` as another method for `JSValue`. also jest-extended checks the length of `Object.keys` to determine if the object is empty. Can you add some more tests for non-enumerable properties and symbols?

dylan-conway (changes requested):
Needs some code moved and a few more tests

src/bun.js/bindings/bindings.cpp
```diff
@@ -2455,6 +2455,48 @@ JSC__JSValue JSC__JSValue__fromEntries(JSC__JSGlobalObject* globalObject, ZigStr
...
 }
 
+JSC__JSValue JSC__JSValue__keys(JSC__JSGlobalObject* globalObject, JSC__JSValue objectValue) {
+    JSC::VM& vm = globalObject->vm();
+
+    JSC::JSObject* object = JSC::JSValue::decode(objectValue).getObject();
+    JSC::JSType type = object->type();
+
+    switch(type){
+        // Map and Set always return empty array for Object.keys
```
dylan-conway: This isn't always true, a property added to a map/set after construction will appear in the array
Screenshot 2024-01-01 at 8 32 13 AM

src/bun.js/bindings/bindings.zig
```diff
@@ -4157,6 +4164,16 @@ pub const JSValue = enum(JSValueReprInt) {
     pub inline fn isObject(this: JSValue) bool {
         return this.isCell() and this.jsType().isObject();
     }
+    pub fn isObjectEmpty(this: JSValue, globalObject: *JSGlobalObject) bool {
+        const type_of_value = this.jsType();
+        // https://github.com/jestjs/jest/blob/main/packages/jest-get-type/src/index.ts#L26
+        // Map and Set are not considered as object in jest-extended
+        if (type_of_value.isMap() or type_of_value.isSet()) {
```
dylan-conway: This also applies to `Date` and `RegExp`

src/bun.js/bindings/bindings.cpp
```diff
@@ -2455,6 +2455,48 @@ JSC__JSValue JSC__JSValue__fromEntries(JSC__JSGlobalObject* globalObject, ZigStr
     return JSC::JSValue::encode(object);
 }
 
+JSC__JSValue JSC__JSValue__keys(JSC__JSGlobalObject* globalObject, JSC__JSValue objectValue) {
```
dylan-conway, line 2477: I like this approach, but instead of a custom implementation of `Object.keys`, we should use the one available in JSC. You can use `JSC::ownPropertyKeys` and the contents of the function should look almost identical to https://github.com/oven-sh/WebKit/blob/b4de09f41b83e9e5c0e43ef414f1aee5968b6f7c/Source/JavaScriptCore/runtime/ObjectConstructor.cpp#L269, except the argument is passed in rather than coming from a callframe

coratgerl: Nice, thanks this is better !

src/bun.js/bindings/bindings.cpp
coratgerl: You are right 👍

src/bun.js/bindings/bindings.cpp
```diff
@@ -2455,6 +2455,17 @@ JSC__JSValue JSC__JSValue__fromEntries(JSC__JSGlobalObject* globalObject, ZigStr
...
+    JSC::VM& vm = globalObject->vm();
+
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    JSC::JSObject* object = JSC::JSValue::decode(objectValue).getObject();
```
dylan-conway:
```suggestion
    JSC::JSObject* object = JSC::JSValue::decode(objectValue).toObject(globalObject);
```
`toObject` will coerce to an object, which is needed for strings in this situation

dylan-conway (approved):
Looks good. Merge conflicts need to be fixed then it will be ready to merge

coratgerl:
It goods thanks you for the time for review 👍

dylan-conway:
Thank you

PR #7782, opened 2023-12-22
whitelist lightningcss-cli

o-az:
lightningcss is a popular css library with over 0.5 million weekly downloads. It also offers a cli tool that requires postinstall script to run.

https://www.npmjs.com/package/lightningcss
https://www.npmjs.com/package/lightningcss-cli?activeTab=code

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
Thank you. cc @devongovett, `bun install` will run the postinstall script for `lightningcss-cli` so it should work as expected

PR #7784, opened 2023-12-22
Fixes #7767: Updated `bun upgrade` to return error when used with cmd arguments

### LINKED ISSUE #7767: Provide better DX when someone confuses `bun update` with `bun upgrade`

`bun update` and `bun upgrade` are often confused (and I'm a victim too: #6422). The fact the industry didn't decide which command does what doesn't help either. For example, `brew` uses `update` and `upgrade` exactly the opposite to how Bun uses it.

There's one thing Brew does very well. When you confuse the commands, Brew is actually super helpful:

```
wmaj@MacBook-Air % brew update node
Error: This command updates brew itself, and does not take formula names.
Use `brew upgrade node` instead.
```

while Bun isn't:

```
wmaj@MacBook-Air % bun upgrade is-even
Congrats! You're already on the latest version of Bun (which is v1.0.18)
```

_No response_

BrookJeynes:
### What does this PR do?

Updated the `upgrade` command to check for command-line arguments and return an error if so providing a suggested `update` command.

fixes #7767 

```
❯ bun upgrade bun-types --dev
error: this command updates bun itself, and does not take package names.
Use `bun update bun-types --dev` instead.
```

- [x] Code changes

I wrote automated tests

If Zig files changed:


test/cli/install/bun-upgrade.test.ts
```diff
@@ -0,0 +1,112 @@
+import { spawn, spawnSync } from "bun";
```
paperclover: for all `bun upgrade` tests that actually invoke the upgrader, you need to copy the bun binary somewhere else then run it, or else bun will override itself, and you'll be testing the latest release.

src/cli.zig
```diff
@@ -1180,7 +1180,27 @@ pub const Command = struct {
             RootCommandMatcher.case("init") => .InitCommand,
             RootCommandMatcher.case("build"), RootCommandMatcher.case("bun") => .BuildCommand,
             RootCommandMatcher.case("discord") => .DiscordCommand,
-            RootCommandMatcher.case("upgrade") => .UpgradeCommand,
+            RootCommandMatcher.case("upgrade") => brk: {
+                if (args_iter.buf.len > 2) {
```
paperclover: is this the best place to put this, can we do this check in upgrade_command.zig?

src/cli.zig
```diff
@@ -1180,7 +1180,27 @@ pub const Command = struct {
...
+                    for (args_iter.buf[2..]) |arg| {
+                        const span = std.mem.span(arg);
+                        if (!strings.contains(span, "--")) {
+                            Output.prettyError(
+                                \\<r><red>error<r><d>:<r> this command updates bun itself, and does not take package names.
+                                \\Use `bun update
+                            , .{});
```
paperclover:
```suggestion
                            Output.prettyError(
                                \\<r><red>error<r><d>:<r> This command updates Bun itself, and does not take package names.
                                \\<blue>note<r><d>:<r> Use `bun update
                            , .{});
```

paperclover (changes requested):
I love this feature. I remember discussion before about confusion between "update" and "upgrade" but this feature is nice solution to the ambiguity.

there is a critical mistake in the test files, so i will delay running ci until that is fixed

src/cli.zig
BrookJeynes: Hey @paperdave, I believe this commit (https://github.com/oven-sh/bun/pull/7784/commits/5279915ba160472fbedea4fcd239138574527ecb) resolves this. Please let me know if not :)

test/cli/install/bun-upgrade.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  );
+  cpSync(bunExe(), run_dir);
+});
+afterEach(async () => {
+  await rm(run_dir, { force: true, recursive: true });
+});
+
+it("two invalid arguments, should display error message and suggest command", async () => {
+  const { stderr, exitCode } = spawn({
+    cmd: [bunExe(), "upgrade", "bun-types", "--dev"],
```
paperclover: bunExe() is the absolute path to the bin binary so this doesnt work, you need to pass the path to the copied executable

test/cli/install/bun-upgrade.test.ts
BrookJeynes: Hey @paperdave, I believe this commit (https://github.com/oven-sh/bun/pull/7784/commits/e0e76c436aaf5825947ae85a5661b1f097342d51) resolves this. Please let me know if not :)

test/cli/install/bun-upgrade.test.ts
BrookJeynes: fixed via commits https://github.com/oven-sh/bun/pull/7784/commits/e0e76c436aaf5825947ae85a5661b1f097342d51 and https://github.com/oven-sh/bun/pull/7784/commits/5284fbc0fc16fc5d380b7235bc4e776e1d5db4b6

paperclover (approved):
i apologize for long delay in review. this looks alright.

PR #7794, opened 2023-12-22
Don't print the cd part of the create command's get started comment if the destination for create was `.`

VesterDe:
### What does this PR do?

The current message for getting started always includes a `cd` part, even when the directory you made the project into is `.`. It feels like it might make more sense not to show it in that case?

This is what I mean:

image

- [x] Code changes

I did not write automated tests. I apologise. I was thinking about it and didn't know what to call the command with, maybe a minimal local template, made just for the test?


Jarred-Sumner:
This code doesn't work

PR #7813, opened 2023-12-24
Update init template & TS documentation

### LINKED ISSUE #7803: bun init doesn't use `@types/bun`

1.0.19+906f86d6f

_No response_

`bun init`

`@types/bun` in dev deps

### What do you see instead?

`bun-types` in dev deps

### Additional information

My expectation comes from: https://bun.sh/blog/bun-v1.0.19#:~:text=or%20you%20can%20ue%20bun%20init%3A

ArnaudBarre:
Fixes #7803

I've done a similar update to the TSConfig close to one I've done for Vite:
- Remove flags enabled by default (`forceConsistentCasingInFileNames`, `allowSyntheticDefaultImports` with `bundler` mode)
- Remove `composite` because this not needed for one file
- Remove `downlevelIteration` as the compilation is done by Bun
- Group flags by topic so this is easier to read
- Use the case of the TS doc (`esnext` -> `ESNext`)

Then the things that are up to decide by the core team:
- Enable `verbatimModuleSyntax` flag which allows to warn when some things are imported as type only which avoid transpiler bug like https://twitter.com/jarredsumner/status/1728790622536962448 (even if catches less things than the TS-ESLint lint rule) 
- Enable `noUnusedLocals`, `noUnusedParameters` & `noFallthroughCasesInSwitch` which for me is always good
- Enable `useUnknownInCatchVariables`, which is also enabled in Vite templates
- Enable `noPropertyAccessFromIndexSignature`, which is a stronger one I discovered recently and really important to make you aware when you're dealing with a loosely typed object. I did not enable in Vite because people are used to `styles.button` instead of `styles["button"]` for CSS Modules. I've personnaly working with it and it's nice when writing code outside of quick scripting.

### DIFF
--- a/docs/guides/runtime/typescript.md
+++ b/docs/guides/runtime/typescript.md
@@ -2,31 +2,10 @@
 name: Install TypeScript declarations for Bun
 ---
 
-To install TypeScript definitions for Bun's built-in APIs in your project, install `bun-types`.
+To install TypeScript definitions for Bun's built-in APIs in your project, install `@types/bun`.
 
 ```sh
-$ bun add -d bun-types # dev dependency
-```
-
----
-
-Then include `"bun-types"` in the `compilerOptions.types` in your `tsconfig.json`:
-
-```json-diff
-  {
-    "compilerOptions": {
-+     "types": ["bun-types"]
-    }
-  }
-```
-
----
-
-Unfortunately, setting a value for `"types"` means that TypeScript will ignore other global type definitions, including `lib: ["dom"]`. If you need to add DOM types into your project, add the following [triple-slash directives](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) at the top of any TypeScript file in your project.
-
-```ts
-/// <reference lib="dom" />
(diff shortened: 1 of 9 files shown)


Jarred-Sumner:
I think this is overall great. The only ones I would skip are:
- `noPropertyAccessFromIndexSignature` because it might annoy some people 
- `forceConsistentCasingInFileNames` is good for catching errors that might not happen in macOS (case insensitive) but do happen on Linux (case sensitive)

Jarred-Sumner:
I'm going to merge and the make those changes

ArnaudBarre:
For info `forceConsistentCasingInFileNames` default to true since v5 that's why I removed it otherwise I agree this is a must have

baboon-king:
The current version of "Bun" is 1.0.20, but the version of the dependencies in "@types/bun" is fixed at 1.0.18.
In this case user can only wait for "@types/bun" to be updated version.

ArnaudBarre:
Some version of Bun can contains only bug fixes/speed improvements without any new API so it's fine to have types that are few versions below the runtime one

PR #7817, opened 2023-12-24
Update installation.md

huseeiin:
npm states that their name should be written as "npm" in lowercase rather than "NPM" or "Npm" etc.

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
Thank you

PR #7824, opened 2023-12-25
15% faster crypto.randomUUID

### LINKED ISSUE #7811: crypto.randomUUID runs faster on node than bun

bun slower than node in generating randomUUID

not sure 🤔 

Code: 

```const Benchmark = require("benchmark");
const crypto = require("crypto");

// Your two functions to compare
function function1() {
  const code = "some data to hash";

  const hash = crypto.createHash("sha256").update(code).digest("hex");
}

function function2() {
  const hash = crypto.randomUUID();
}

function function3() {
     const a = crypto.randomBytes(Math.ceil(12 / 2)).toString('hex').slice(0, 12);
}   

// Create a new benchmark suite
const suite = new Benchmark.Suite();

// Add tests to the suite
suite
  .add("Function 1", function () {
    function1();
  })
  .add("Function 2", function () {
    function2();
  })
  .add("Function 3", function () {
    function3();
  })
  // Add listeners for the 'cycle' and 'complete' events
  .on("cycle", function (event) {
    console.log(String(event.target));
  })
  .on("complete", function () {
    console.log("Fastest is " + this.filter("fastest").map("name"));
  })
  // Run the benchmark
  .run({ async: true });
```

image

Jarred-Sumner:
### What does this PR do?

Fixes #7811

image

There are existing tests

### DIFF
--- a/src/bun.js/webcore.zig
+++ b/src/bun.js/webcore.zig
@@ -662,7 +662,7 @@ pub const Crypto = struct {
         defer str.deref();
 
         // randomUUID must have been called already many times before this kicks
-        // in so we can skip the rare_data poitner check.
+        // in so we can skip the rare_data pointer check.
         const uuid = globalThis.bunVM().rare_data.?.nextUUID();
 
         uuid.print(bytes[0..36]);
--- a/src/bun.js/webcore.zig
+++ b/src/bun.js/webcore.zig
@@ -624,11 +624,13 @@ pub const Crypto = struct {
         globalThis: *JSC.JSGlobalObject,
         _: *JSC.CallFrame,
     ) callconv(.C) JSC.JSValue {
-        var out: [36]u8 = undefined;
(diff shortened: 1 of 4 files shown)


src/bun.js/webcore.zig
```diff
@@ -656,11 +658,15 @@ pub const Crypto = struct {
         _: *Crypto,
         globalThis: *JSC.JSGlobalObject,
     ) callconv(.C) JSC.JSValue {
-        var out: [36]u8 = undefined;
-        const uuid = globalThis.bunVM().rareData().nextUUID();
+        const str, var bytes = bun.String.createUninitialized(.latin1, 36);
+        defer str.deref();
+
+        // randomUUID must have been called already many times before this kicks
+        // in so we can skip the rare_data poitner check.
```
staabm:
```suggestion
        // in so we can skip the rare_data pointer check.
```

PR #7831, opened 2023-12-25
Allow custom glob patterns for bun test

zackradisic:
### What does this PR do?

- [x] Code changes

This allows specifying a custom glob pattern to configure which files to scan for tests by adding the `--include` flag and the `test.include` property in bunfig.toml

Since we use the word "filter" to refer to the positional argument, I named this option "include" to avoid confusion and it is also what Vitest uses so it will be familiar to users of that

This feature works in conjunction with positional arguments. For example:
```bash
bun test --include "tests/**/*.ts" foo bar
```
Which will only run tests in .ts files inside `test/` or one of its subdirectories, and only if the filenames include `foo` or `bar` in the name

I wrote automated tests

Zig files changed:

### DIFF
--- a/src/cli/test_command.zig
+++ b/src/cli/test_command.zig
@@ -499,13 +499,14 @@ const Scanner = struct {
     }
 
     pub fn doesAbsolutePathMatchFilter(this: *Scanner, name: string) bool {
-        if (this.filter_names.len == 0) return true;
+        return TestCommand.doesAbsolutePathMatchFilter(this.filter_names, name);
+        // if (this.filter_names.len == 0) return true;
 
-        for (this.filter_names) |filter_name| {
-            if (strings.startsWith(name, filter_name)) return true;
-        }
+        // for (this.filter_names) |filter_name| {
+        //     if (strings.startsWith(name, filter_name)) return true;
+        // }
 
-        return false;
+        // return false;
     }
 
     pub fn doesPathMatchFilter(this: *Scanner, name: string) bool {
@@ -577,6 +578,26 @@ pub const TestCommand = struct {
         enabled: bool = false,
         fail_on_low_coverage: bool = false,
     };
+
+    pub fn doesAbsolutePathMatchFilter(filter_names: []const []const u8, name_: string) bool {
+        if (filter_names.len == 0) return true;
+
+        for (filter_names) |filter_name| {
+            if (strings.startsWith(name_, filter_name)) return true;
+        }
+
+        return false;
+    }
+
+    pub fn doesPathMatchFilter(filter_names: []const []const u8, name_: string) bool {
+        if (filter_names.len == 0) return true;
+
+        for (filter_names) |filter_name| {
+            if (strings.contains(name_, filter_name)) return true;
+        }
+
+        return false;
+    }
 
     pub fn exec(ctx: Command.Context) !void {
         if (comptime is_bindgen) unreachable;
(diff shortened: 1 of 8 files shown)


src/cli/test_command.zig
```diff
@@ -499,13 +499,14 @@ const Scanner = struct {
...
 
     pub fn doesAbsolutePathMatchFilter(this: *Scanner, name: string) bool {
-        if (this.filter_names.len == 0) return true;
+        return TestCommand.doesAbsolutePathMatchFilter(this.filter_names, name);
+        // if (this.filter_names.len == 0) return true;
 
-        for (this.filter_names) |filter_name| {
-            if (strings.startsWith(name, filter_name)) return true;
-        }
+        // for (this.filter_names) |filter_name| {
```
Jarred-Sumner, line 505: did you mean to keep this code in here commented out?

src/cli/test_command.zig
```diff
@@ -718,6 +739,70 @@ pub const TestCommand = struct {
             // Treat arguments as filters and scan the codebase
             const filter_names = if (ctx.positionals.len == 0) &[0][]const u8{} else ctx.positionals[1..];
 
+            if (ctx.test_options.include) |include_glob_pattern| {
+                const glob = @import("../glob.zig");
+                const GlobWalker = glob.GlobWalker_(glob.ignoreNodeModules, true);
```
Jarred-Sumner, line 744: it also should ignore `.git`, and I think a few other folders?

src/cli/test_command.zig
```diff
@@ -578,6 +579,26 @@ pub const TestCommand = struct {
         fail_on_low_coverage: bool = false,
     };
 
+    pub fn doesAbsolutePathMatchFilter(filter_names: []const []const u8, name_: string) bool {
+        if (filter_names.len == 0) return true;
+
+        for (filter_names) |filter_name| {
+            if (strings.startsWith(name_, filter_name)) return true;
```
Jarred-Sumner, line 586: `hasPrefix` should be used here, not `startsWith`. `startsWith` includes empty strings.

```suggestion
            if (strings.hasPrefix(name_, filter_name)) return true;
```

src/cli/test_command.zig
```diff
@@ -718,6 +739,70 @@ pub const TestCommand = struct {
             // Treat arguments as filters and scan the codebase
             const filter_names = if (ctx.positionals.len == 0) &[0][]const u8{} else ctx.positionals[1..];
 
+            if (ctx.test_options.include) |include_glob_pattern| {
+                const glob = @import("../glob.zig");
```
Jarred-Sumner, line 743: can you use `bun.Glob` or something which is not an import path? if we move files around later, this will be something that needs to be updated

test/cli/test/bun-test.test.ts
```diff
@@ -6,6 +6,53 @@ import { describe, test, expect } from "bun:test";
...
+        contents: /* typescript */ `
+        import { test, expect } from "bun:test"
+        test("test #4", () => {
+          expect(true).toBe(true);
+        });
+        `,
+      },
+    ]);
+    const stderr = runTest({
+      args: ["--include", "**/*.lol.{ts,js}"],
```
Jarred-Sumner, line 49: This test should include files which are in `node_modules` and also in folders. It should also include files which are symlinks to other tests, just to see what happens

Jarred-Sumner:
@zackradisic i'm closing this because it has bitrotted, but we still should do this if we have not already

PR #7832, opened 2023-12-26
Whitelist Railway CLI

o-az:
https://railway.app

Railway CLI is used to deploy Railway services. The CLI npm package requires running postinstall scripts.

https://github.com/railwayapp/cli/blob/9de89415f000316edf98017885dd08e684f284fb/package.json#L17

### What does this PR do?

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #7837, opened 2023-12-26
chore: sort default-trusted-dependencies.txt

ThatOneCalculator:
### What does this PR do?

- [ ] Code changes
- [x] Config changes

Ref: https://github.com/oven-sh/bun/pull/7832#pullrequestreview-1795984616 cc @paperdave


paperclover:
sorry for the delay in review. there is a merge conflict.

src/install/default-trusted-dependencies.txt
```diff
@@ -73,6 +73,10 @@
 @pulumi/awsx
 @pulumi/command
 @pulumi/kubernetes
+@pulumi/postgresql
```
paperclover: why was @pulumi/postgresql and these others added? i checked postgresql and it does not have an install script

paperclover (changes requested):
see comment

src/install/default-trusted-dependencies.txt
paperclover: i think this was removed in another pr, but the merge conflict left this back in. can you remove

PR #7838, opened 2023-12-26
adding purescript and spago to default-trusted-dependencies.txt

rtxanson:
### What does this PR do?

This adds two packages to default-trusted-dependencies.txt. These are two important packages for purescript development, `purescript` is the language (including a binary file that must be installed), and `spago` is the purescript package manager (also including a binary file that must be installed). 

Took a very loooong time for me to realize that trusted dependencies were the problem due to lack of obvious error feedback in the bun install process, so hoping to save some people time with this.


Jarred-Sumner:
Thank you

PR #7839, opened 2023-12-26
Implement Brotli support in `fetch()`

Jarred-Sumner:
### What does this PR do?

This adds support for brotli in `fetch()` (and technically `bun install`, for the manifest requests)

It does not expose runtime APIs to use Brotli just yet (so this also means it doesn't work in node:http since node:http has to disable compression first).

To avoid increasing Bun's binary size, this adds support for two versions of Brotli:
- `libbrotlidec`, which is the standard Google brotli implementation. It is dynamically loaded. 
- `Compression.framework`, which ships with macOS >= 10.11 and includes a Brotli implementation. It is also dynamically loaded

This also adds a demo CLI tool that uses Compression.framework into misctools. Mostly I wanted to see if the performance of the Google Brotli implementation was different and it wasn't different.

Will update PR with tests shortly.


src/deps/brotli_decoder.zig
```diff
@@ -0,0 +1,322 @@
...
+    const brotli_decoder_metadata_chunk_func = ?*const fn (?*anyopaque, [*]const u8, usize) callconv(.C) void;
+
+    var brotli_handle: ?*anyopaque = null;
+
+    fn loadBrotli() ?*anyopaque {
+        if (brotli_handle != null) {
+            return brotli_handle;
+        }
+
+        brotli_handle = bun.C.dlopen("brotlidec", 1) orelse brk: {
```
cirospaciari, line 40: by default on debian `libbrotli-dev` is not installed, I got a debug warning `debug warn: failed to load Brotli`, we should document how to enable `brotli`, and probably expose `canUseBrotli` to check if it was loaded successfully at runtime (for release builds) in the future (or something similar).

src/http.zig
```diff
@@ -1194,6 +1205,130 @@ pub const CertificateInfo = struct {
...
+                                &body_out_str.list,
+                                body_out_str.allocator,
+                                .{},
+                            ),
+                        };
+                    }
+
+                    return;
+                },
+                else => unreachable,
```
cirospaciari:
```suggestion
                else => @panic("Invalid encoding. This code should not be reachable"),
```

src/http.zig
```diff
@@ -1194,6 +1205,130 @@ pub const CertificateInfo = struct {
...
+
+                reader.input = buffer;
+                reader.total_in = @as(u32, @truncate(buffer.len));
+
+                const initial = body_out_str.list.items.len;
+
+                reader.list = body_out_str.list;
+                reader.total_out = @truncate(initial);
+            },
+            else => unreachable,
```
cirospaciari:
```suggestion
            else => @panic("Invalid encoding. This code should not be reachable"),
```

cirospaciari (commented):
Overall, it looks good, I saw that we have fetch tests failing, which looks related to redirect.

Jarred-Sumner:
I’m merging this, but will probably add brotli as a static dependency because I think it’s likely too unreliable to dynamically link it

PR #7840, opened 2023-12-26
Add `nuxt` to trusted dependencies whitelist

L422Y:
[Nuxt](https://nuxt.com) is a popular open source web development framework, which supports `bun` as a default package manager.

Nuxt requires running of a `postinstall` task to generate initial types, during installation and after upgrades, if this doesn't get run after an upgrade of existing projects, things can appear broken.

https://www.npmjs.com/package/nuxt


Jarred-Sumner:
Thank you

PR #7856, opened 2023-12-27
Fixes #7849

### LINKED ISSUE #7849: `bun test`: API `describe(fn, ...)` does not register test properly

1.0.20+09d51486e

Darwin 23.0.0 arm64 arm

with this in `fn.test.ts`:
```typescript
function fn(n: number) { 
  return n + 1
}

describe(fn, () => {
  test("zero", () => {
    expect(fn(0)).toBe(1)
  })
})
```
Running `bun test` will not recognize this as a valid suite of tests; it will find the `fn.test.ts` file but return "0 pass", "0 fail". If you change this to `describe("fn"...`, it will recognize and mark the test as passing.

Using the function reference in `describe`, as in `describe(fn, ...)`, should register a test with the name of the function, as in Jest.
Screenshot 2023-12-26 at 11 37 44

### What do you see instead?

The test is not registered.

### Additional information

This is important for backwards compatibility with existing-codebase Jest tests.

Jarred-Sumner:
### What does this PR do?

Fixes #7849

There is a test


src/bun.js/test/jest.zig
```diff
@@ -1548,6 +1548,16 @@ inline fn createScope(
     var function = if (args.len > 1) args[1] else .zero;
     var options = if (args.len > 2) args[2] else .zero;
 
+    // https://github.com/oven-sh/bun/issues/7849
+    if (!description.isEmptyOrUndefinedOrNull() and description.isCallable(globalThis.vm())) {
```
jdalton, line 1552: Do you need both the isEmptyOrUndefinedOrNull() in combo with the isCallable() or can it work with just the isCallable?

src/bun.js/test/jest.zig
```diff
@@ -1548,6 +1548,16 @@ inline fn createScope(
...
+        // TODO: do we need an exception check here?
+        if (description.get(globalThis, "name")) |name| {
```
jdalton, line 1554: A function name in JS could throw if defined to.

PR #7861, opened 2023-12-27
Fixes #7001

### LINKED ISSUE #7001: Segmentation fault and Request body stream locking

1.0.11

Darwin 22.6.0 x86_64 i386

There seem to be two problems 1) the body ReadableStream is not being locked when reading it and 2) trying to read the stream afterwards fails with segmentation fault.

```ts
const req = new Request('https://example.com/', {
  body: 'test',
  method: 'POST',
});

// getting array buffer of the body to calculate body hash (for request signing for example)
const ab = await new Response(req.body).arrayBuffer();

// should be true, but is false in bun
console.log('Locked:', req.body?.locked);

// send request
await fetch(req);
```

Just for correctness, the right way of getting array buffer is using `req.clone().body` (which works fine in bun); Discovered this segmentation fault problem by mistake.

1) req.body.locked should be true
2) should not fail with segmentation fault; should throw error (stream locked) just like other JS environments

### What do you see instead?

_No response_

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

This fixes the immediate cause of #7001. It doesn't fix it in every case (such as streams)

1) Makes it so `ReadableStream.prototype.isLocked` returns true if our hidden native pointer value is set to `-1`. 
2) We check if the `ReadableStream` is locked, including for Blob values

There are some tests

@cirospaciari please review this closely because it's very easy to subtly break things with this code and I'm worried something will break due to this change


src/bun.js/webcore/streams.zig
```diff
@@ -96,10 +97,21 @@ pub const ReadableStream = struct {
...
+            this.value.unprotect();
+            this.* = .{ .ptr = .{ .Invalid = {} }, .value = .zero };
+        }
+    }
+
     pub fn toAnyBlob(
         stream: *ReadableStream,
         globalThis: *JSC.JSGlobalObject,
     ) ?JSC.WebCore.AnyBlob {
+        stream.reloadTag(globalThis);
```
Jarred-Sumner, line 117: This is the main cause 

We internally store the `ReadableStream` on the `PendingValue`, and the pointer becomes stale. It is a use-after-free where the `ptr` has been detached, but the existing `PendingValue` hasn't been updated to reflect that. To address this, we reload the tag.

src/bun.js/webcore/streams.zig
```diff
@@ -147,14 +159,19 @@ pub const ReadableStream = struct {
 
     pub fn cancel(this: *const ReadableStream, globalThis: *JSGlobalObject) void {
         JSC.markBinding(@src());
-        this.value.unprotect();
         ReadableStream__cancel(this.value, globalThis);
+        this.value.unprotect();
```
Jarred-Sumner, line 167: Moving this after out of caution that the value gets GC'd immediately after its unprotected

src/js/builtins/ReadableStreamInternals.ts
```diff
@@ -1391,7 +1391,10 @@ export function readableStreamDefaultControllerClose(controller) {
 }
 
 export function readableStreamClose(stream) {
-  $assert($getByIdDirectPrivate(stream, "state") === $streamReadable);
+  $assert(
```
Jarred-Sumner, line 1394: This fixes the debug assertion failure in the direct readable stream tests.

src/bun.js/webcore/response.zig
```diff
@@ -1911,7 +1911,23 @@ pub const Fetch = struct {
                     }
                 } else {
                     method = request.method;
+
+                    if (request.body.value == .Locked) {
+                        if (request.body.value.Locked.readable) |stream| {
+                            if (stream.isDisturbed(globalThis)) {
```
cirospaciari, line 1917: should we also check isDisturbed in other places we call `useAsAnyBlob`? like: https://github.com/oven-sh/bun/blob/99aef8c217c783d5a23ed5911dca77678b6f1bde/src/bun.js/webcore/response.zig#L1996 https://github.com/oven-sh/bun/blob/99aef8c217c783d5a23ed5911dca77678b6f1bde/src/bun.js/webcore/response.zig#L1779

src/js/node/stream.js
```diff
@@ -5279,7 +5279,7 @@ function createNativeStreamReadable(nativeType, Readable) {
 
       var ptr = this.#bunNativePtr;
       $debug("ptr @ NativeReadable._read", ptr, this.__id);
-      if (ptr === 0) {
+      if (ptr === 0 || ptr === -1) {
```
cirospaciari, line 5272: maybe adding some comments for what -1 and 0 mean (Locked/Detached) so we can remember next time

test/regression/issue/07740.test.ts
```diff
@@ -0,0 +1,29 @@
+import { bunEnv } from "harness";
+import { bunExe } from "harness";
+import { tempDirWithFiles } from "harness";
+
+it("duplicate dependencies should warn instead of error", () => {
```
cirospaciari: this test fails on linux

cirospaciari (changes requested):
tests failures

PR #7866, opened 2023-12-27
feat: implement console.table

### LINKED ISSUE #802: console.table

bun `console.table` doesn't display tabular data in a table

https://nodejs.org/api/console.html#consoletabletabulardata-properties

```
> console.table([{a:1},{a:2}])
┌─────────┬───┐
│ (index) │ a │
├─────────┼───┤
│    0    │ 1 │
│    1    │ 2 │
└─────────┴───┘
```

It also works in deno

_No response_

otgerrogla:
### What does this PR do?

Closes #802

Implements `console.table` in native zig code.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/bindings/exports.zig
+++ b/src/bun.js/bindings/exports.zig
@@ -991,6 +991,23 @@ pub const ZigConsoleClient = struct {
             .flush = true,
         };
 
+        if (message_type == .Table and len >= 1) {
+            // if value is not an object/array/iterable, don't print a table and just print it
+            var tabular_data = vals[0];
+            if (tabular_data.isObject()) {
+                const properties = if (len >= 2 and vals[1].jsType().isArray()) vals[1] else JSValue.undefined;
+                var table_printer = TablePrinter.init(
+                    global,
+                    level,
+                    tabular_data,
+                    properties,
+                );
+                table_printer.printTable(writer) catch return;
+                buffered_writer.flush() catch {};
+                return;
+            }
+        }
+
         if (message_type == .Dir and len >= 2) {
             print_length = 1;
             var opts = vals[1];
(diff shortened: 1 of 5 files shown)


src/bun.js/bindings/exports.zig
```diff
@@ -1031,6 +1048,343 @@ pub const ZigConsoleClient = struct {
...
+                        }
+                    }
+                } else {
+                    var cols_iter = JSC.JSPropertyIterator(.{
+                        .skip_empty_name = false,
+                        .include_value = true,
+                    }).init(this.globalObject, row_value.asObjectRef());
+                    defer cols_iter.deinit();
+
+                    while (cols_iter.next()) |col_key| {
```
Jarred-Sumner, line 1156: sidenote: in a different PR, we should change PropertyIterator to use `bun.String` because the memory here is pretty unsafe (its referencing the latin1/utf16 character pointer which may be freed at any time later)

Jarred-Sumner: this is a problem elsewhere, not here. and in this case, it's pretty much fine because the values will be alive for the scope of this function call anyway

src/bun.js/bindings/exports.zig
```diff
@@ -1031,6 +1048,343 @@ pub const ZigConsoleClient = struct {
...
+                    try writer.writeByteNTimes(' ', col.width + 2 * PADDING);
+                } else {
+                    const len: u32 = this.getWidthForValue(value);
+
+                    const pad_l = (col.width - len) >> 1;
+                    const pad_r = col.width - len - pad_l;
+                    try writer.writeByteNTimes(' ', pad_l + PADDING);
+                    const tag = ZigConsoleClient.Formatter.Tag.get(value, this.globalObject);
+                    switch (Output.enable_ansi_colors) {
+                        inline else => |enable_ansi_colors| {
```
Jarred-Sumner: IMO, it would be better to do this as a comptime argument to this function and then we only need to check it once instead of each column

but that's a small thing

src/bun.js/bindings/exports.zig
```diff
@@ -1031,6 +1048,343 @@ pub const ZigConsoleClient = struct {
...
+
+            // if the "properties" arg was provided, pre-populate the columns
+            if (!this.properties.isUndefined()) {
+                var properties_iter = JSC.JSArrayIterator.init(this.properties, globalObject);
+                while (properties_iter.next()) |value| {
+                    var property_name = value.toStringOrNull(globalObject) orelse {
+                        return error.JSError;
+                    };
+                    try columns.append(.{
+                        .name = String.init(property_name.getZigString(globalObject)),
```
Jarred-Sumner:
```suggestion
                        .name = property_name.toBunString(globalObject),
```

test/js/bun/console/console-table.test.ts
```diff
@@ -0,0 +1,211 @@
...
+└─────────┴─────────────────────────────────┘
+`,
+      },
+    ],
+  ])("expected output for: %s", (label, { args, output }) => {
+    const { stdout } = spawnSync({
+      cmd: [bunExe(), `${import.meta.dir}/console-table-run.ts`, args.toString()],
+      stdout: "pipe",
+      stderr: "inherit",
+      env: {
```
Jarred-Sumner: you can use `bunEnv` here and it will remove debug logs and colors

Jarred-Sumner:
This looks very close to mergeable. Just a few comments

Jarred-Sumner:
Looks like the console.log test is failing. I wonder if it's related to the single_line change?

otgerrogla:
Fixed, test is now working.
Looks like a git auto-merge did mess up at some point. I checked there are no other instances of this.

Jarred-Sumner:
Awesome!

Issue #7239, opened 2023-11-20
`assert.deepStrictEqual` ignores `message` property of Error objects


davidchambers:
### What version of Bun is running?

1.0.13

### What platform is your computer?

Darwin 21.6.0 arm64 arm

### What steps can reproduce the bug?

```console
$ bun repl
Welcome to Bun v1.0.13
Type ".help" for more information.
[!] Please note that the REPL implementation is still experimental!
    Don't consider it to be representative of the stability or behavior of Bun overall.
> assert.deepStrictEqual(Error('foo'), Error('bar'))
undefined
```

### What is the expected behavior?

```console
$ node
Welcome to Node.js v20.0.0.
Type ".help" for more information.
> assert.deepStrictEqual(Error('foo'), Error('bar'))
Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ [Error: foo]
- [Error: bar]
          ^
    at REPL1:1:8
    at Script.runInThisContext (node:vm:122:12)
    at REPLServer.defaultEval (node:repl:570:29)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:900:10)
    at REPLServer.emit (node:events:523:35)
    at REPLServer.emit (node:domain:489:12)
    at [_onLine] [as _onLine] (node:internal/readline/interface:415:12)
    at [_line] [as _line] (node:internal/readline/interface:886:18) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: [Error: foo],
  expected: [Error: bar],
  operator: 'deepStrictEqual'
}
```

### Additional information

I have discovered one other problem with `assert.deepStrictEqual`; I won't mention it here as it may have a different cause.

PR #7867, opened 2023-12-27
Compare error name and message on `Bun.deepEquals` and `assert.deepStrictEqual`

kaioduarte:
### What does this PR do?

Currently, Bun doesn't handle comparisons among error instances, it only checks if the types are different.

- [x] Code changes

If C++* files changed:

### Sanity check
| `Bun.deepEquals` | Bun's `assert.deepStrictEqual` |
| --- | --- |
|image|image|

#### Node for comparison
image

Closes https://github.com/oven-sh/bun/issues/7239

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -830,6 +830,24 @@ bool Bun__deepEquals(JSC__JSGlobalObject* globalObject, JSValue v1, JSValue v2, 
 
         return false;
     }
+    case ErrorInstanceType: {
+        if (c2Type != ErrorInstanceType) {
+            return false;
+        }
+
+        if (JSC::ErrorInstance* left = jsDynamicCast<JSC::ErrorInstance*>(v1)) {
+            JSC::ErrorInstance* right = jsDynamicCast<JSC::ErrorInstance*>(v2);
+
+            if (UNLIKELY(!right)) {
+                return false;
+            }
+
+            return (
+                left->sanitizedNameString(globalObject) == right->sanitizedNameString(globalObject) &&
+                left->sanitizedMessageString(globalObject) == right->sanitizedMessageString(globalObject)
+            );
+        }
+    }
     case Int8ArrayType:
     case Uint8ArrayType:
     case Uint8ClampedArrayType:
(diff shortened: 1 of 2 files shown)


kaioduarte:
The `assert.deepStrictEqual` snippet I shared is not working on REPL, any idea why? I thought it was something with the debug build, but it doesn't work on the release build as well.

src/bun.js/bindings/bindings.cpp
```diff
@@ -830,6 +830,24 @@ bool Bun__deepEquals(JSC__JSGlobalObject* globalObject, JSValue v1, JSValue v2,
...
+            JSC::ErrorInstance* right = jsDynamicCast<JSC::ErrorInstance*>(v2);
+
+            if (UNLIKELY(!right)) {
+                return false;
+            }
+
+            return (
+                left->sanitizedNameString(globalObject) == right->sanitizedNameString(globalObject) &&
+                left->sanitizedMessageString(globalObject) == right->sanitizedMessageString(globalObject)
+            );
```
kaioduarte, line 848: This is based on Node's implementation:

https://github.com/nodejs/node/blob/5e6448310d228aab139d1a5cb62ef5b771ca2c4b/lib/internal/util/comparisons.js#L235-L242

test/js/bun/test/test-test.test.ts
```diff
@@ -161,6 +161,10 @@ test("testing Bun.deepEquals() using isEqual()", () => {
   expect(Infinity).toEqual(1 / 0);
   expect(-Infinity).toEqual(-Infinity);
   expect(-Infinity).toEqual(-1 / 0);
+
+  expect(Error("foo")).toEqual(Error("foo"));
+  expect(Error("foo")).not.toEqual(Error("bar"));
+  expect(Error("foo")).not.toEqual("foo");
```
Jarred-Sumner, line 167: What happens if the `Error` is a subclass of `Error`? 

```suggestion
  class CustomError extends Error { constructor(message) { super(message); } };
  expect(new CustomError("foo")).not.toEqual(new CustomError("bar"));
  expect(new CustomError("foo")).toEqual(new CustomError("foo"));
```

Does jest behave similar here?

kaioduarte: Yes, it behaves similarly on Jest. I'll add that to the tests.

image

Jarred-Sumner:
Thank you

PR #7870, opened 2023-12-27
scripts/setup.sh: don't check for zig since it installs zig later

nektro:
zig installation added in #7305

needed this change during my environment setup


Electroid (approved):

Electroid:
🎉

PR #7871, opened 2023-12-27
script/setup.sh: don't hardcode search for pkg-config

nektro:
documented [1] and [2] and [3] and necessary to use alternate implementations.

needed this for my environment setup.

[1]: https://linux.die.net/man/1/pkg-config
[2]: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html
[3]: https://github.com/pkgconf/pkgconf


paperclover:
nice

paperclover (approved):

Issue #7734, opened 2023-12-19
console.log omits `call` property


peetzweg:
### What version of Bun is running?

1.0.18+36c316a24

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

```sh
bun repl -e "console.log({call:1, not_call:2})"
```

prints `not_call` only, leaving out the object key `call`.
```js
{
  not_call: 2,
}
```

### What is the expected behavior?

Should print every key in given object. Not sure if this issue is only limited to a key called `call` only.

Should print
```js
{ 
  call: 1, 
  not_call: 2 
}
```

comapre with `node -e "console.log({call:1, not_call:2})"`

``js
{ call: 1, not_call: 2 }
``

### What do you see instead?

```js
{
  not_call: 2,
}
```

paperclover:
<img width="423" alt="image" src="https://github.com/oven-sh/bun/assets/24465214/45248180-e227-4b8f-a520-5ab42a29eb47">

peetzweg:
https://github.com/oven-sh/bun/blob/a623e7c7026554496abf5ad23b6fdd4129b3a773/src/bun.js/test/pretty_format.zig#L793-L795

Any specific reason for leaving out these keys?

Seems to work just fine in node.
```sh
node -e "console.log({call:1, constructor:2})"
{ call: 1, constructor: 2 }
```

No too big of a hassle to use a different key name, however it left me perplexed a bit.

paperclover:
> Any specific reason for leaving out these keys?

it's to filter out stuff on class constructors and function objects, but i think this is completely incorrect to do this, and if these keys were to be filtered out, it should be filtered a different method.

paperclover:
ive noticed a few things with how our property iterator works, and eventually ill open a pr to fix this an a bunch of other things, but if you or anyone wants to submit a patch to remove these two that would be okay.

PR #7874, opened 2023-12-28
inspect: print 'call' property in objects

nektro:
functions have separate code path now

related but different: #7734


Jarred-Sumner:
Thank you

PR #7875, opened 2023-12-28
bun.js bindings: tidy with inline range

nektro:
makes switch body more clear, feature was not in previous version of zig


Jarred-Sumner:
Thank you

PR #7876, opened 2023-12-28
bun.js bindings: remove redundant as-bitcast-as

nektro:
source and destination type are the same, this was doing nothing


Jarred-Sumner:
Thank you

Issue #7872, opened 2023-12-28
Support passing a Blob to Bun.CryptoHasher


rosszurowski:
### What is the problem this feature would solve?

I've been using `Bun.CryptoHasher` to get the SHA-256 of a multipart uploaded file. The files I'm dealing with are quite large (~20mb to ~100mb), so I've been using `blob.stream()` to avoid duplicating the whole set of data, but according to [this discord thread](https://discord.com/channels/876711213126520882/1189349384190697562/1189712458936492072), this duplicates data.

If `Bun.CryptoHasher` took a blob, I could do this more simply.

### What is the feature you are proposing to solve the problem?

Something along the lines of:

```ts
const hasher = new Bun.CryptoHasher("sha256");
hasher.update(blob);
// or alternatively,
hasher.fromBlob(blob);
```

### What alternatives have you considered?

Using `blob.stream()` (which apparently duplicates data), converting to a `Uint8Array` (which would duplicate the data).

Electroid:
This makes sense, we would want to support a way to do this without copying required.

PR #7878, opened 2023-12-28
Support `Blob` in `CryptoHasher`

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/7872

It does not add support for `Bun.file` (`File`, like in `FormData` though should work fine) 

Tests


erikschul:
@Jarred-Sumner This is a weird PR. It doesn't add support, which is the title. It adds tests to ensure that errors are thrown. This is confusing.

Issue #7786, opened 2023-12-22
expect(obj).toMatchSnapshot(undefined, 'a') doesn't work


xlc:
### What version of Bun is running?

1.0.17+5e60861c3

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

```ts
import { test, expect } from 'bun:test'

test('it should work', () => {
  const obj = { a: 1 }
  expect(obj).toMatchSnapshot(undefined, 'a')
})
```

and `bun test`

For reference, this is the signature and docs from bun-types

```ts
    /**
     * Asserts that a value matches the most recent snapshot.
     *
     * @example
     * expect([1, 2, 3]).toMatchSnapshot();
     * expect({ a: 1, b: 2 }).toMatchSnapshot({ a: 1 });
     * expect({ c: new Date() }).toMatchSnapshot({ c: expect.any(Date) });
     *
     * @param propertyMatchers Object containing properties to match against the value.
     * @param hint Hint used to identify the snapshot in the snapshot file.
     */
    toMatchSnapshot(propertyMatchers?: object, hint?: string): void;
```

so I am expecting `propertyMatchers` can be undefined but instead it throws error `Matcher error: Expected properties must be an object`

```
bun test v1.0.17 (5e60861c)

tests/bun.test.ts:
1 | import { test, expect } from 'bun:test'
2 | 
3 | test('it should work', () => {
4 |   const obj = { a: 1 }
5 |   expect(obj).toMatchSnapshot(undefined, 'a')
      ^
error: expect(received).toMatchSnapshot(properties, hint)

Matcher error: Expected properties must be an object

      at /Users/xiliangchen/projects/acala/e2e-tests/tests/bun.test.ts:5:3
✗ it should work [2.38ms]

 0 pass
 1 fail
 1 expect() calls
Ran 1 tests across 1 files. [9.00ms]
```

xlc:
looks like I can do `expect(obj).toMatchSnapshot('a' as any)`, so the issue is the signature isn't correct

PR #7880, opened 2023-12-28
fix toMatchSnapshot signature

xlc:
### What does this PR do?

closes https://github.com/oven-sh/bun/issues/7786

For reference, this is the def in jest https://github.com/jestjs/jest/blob/6604b37930d734d6ca9f69cc764785c3e862d63c/packages/jest-snapshot/src/types.ts#L39-L52

and this is the zig code https://github.com/oven-sh/bun/blob/c195926c190dd7b424e17da4b783b8475d95b347/src/bun.js/test/expect.zig#L2192-L2217

so that it supports signature with only hint message but ts defs doesn't reflect that

- [ ] Code changes

Just eyeball it.

I can add some tests if you can let me know which file to add it.

### DIFF
--- a/packages/bun-types/test.d.ts
+++ b/packages/bun-types/test.d.ts
@@ -1145,6 +1145,14 @@ declare module "bun:test" {
      * Asserts that a value matches the most recent snapshot.
      *
      * @example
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you

PR #7894, opened 2023-12-28
Fix Example Syntax for Plugins

### LINKED ISSUE #7893: Example Syntax for Plugin Documentation is Unclear (missing plugin())

Documentation is missing, Documentation is incorrect, Documentation is confusing, Example code is not working

### What is the issue?

Bun's current documentation for Runtime Plugins shows using this code for a plugin:

```js
import { plugin, type BunPlugin } from "bun";

const myPlugin: BunPlugin = {
  name: "Custom loader",
  setup(build) {
    // implementation
  },
};
```

However, the code never actually registers `myPlugin` as a plugin:

```js
plugin(myPlugin);
```

### Where did you find it?

[Bun.sh > Runtime Plugins](https://bun.sh/docs/runtime/plugins)

ghost:
### What does this PR do?

This PR fixes #7893 by making the following changes:
- updated docs/runtime/plugins.md to use plugin()
- update docs/runtime/plugins.md to use "loaded" instead of "registered" for clarity

- [X] Documentation or TypeScript types
- [ ] Code changes

### DIFF
--- a/docs/runtime/plugins.md
+++ b/docs/runtime/plugins.md
@@ -15,9 +15,11 @@ const myPlugin: BunPlugin = {
     // implementation
   },
 };
+
+plugin(myPlugin);
 ```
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you

Issue #7555, opened 2023-12-10
Curl made requests to Bun.serve and received the message: 'Leftovers after chunking: 7u bytes'.


drag0n-app:
### What version of Bun is running?

1.0.15+b3bdf22eb

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

```ts
Bun.serve({
    port: 1000, 
    async fetch(req) {
        const res = await fetch('https://bun.sh/');
        return new Response(res.body);
    },
});
```

```
curl -v --raw --url http://127.0.0.1:1000
```

### What is the expected behavior?

The correct response ends with:
```
...
</body>
</html>
0

* Connection #0 to host 127.0.0.1 left intact
```

But I randomly got:
```
...
</body>
* Connection #0 to host 127.0.0.1 left intact
</html>% 
```
or
```
...
</body>
</html>
0


0

* Leftovers after chunking:  7u bytes
* Connection #0 to host 127.0.0.1 left intact
```

### What do you see instead?

```
curl -v --raw --url http://127.0.0.1:1000
*   Trying 127.0.0.1:1000...
* Connected to 127.0.0.1 (127.0.0.1) port 1000 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:1000
> User-Agent: curl/8.1.2
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Sun, 10 Dec 2023 02:32:03 GMT
< Transfer-Encoding: chunked
<
19e1

  <!DOCTYPE html>
<html lang="en">
...
</body>
</html>
0


0

* Leftovers after chunking:  7u bytes
* Connection #0 to host 127.0.0.1 left intact
```

drag0n-app:
Who's going to check this out?

staabm:
the same people which take care of the other open ~1900 issues. If you want to see it fixed in the short term, work on the problem yourself

Jarred-Sumner:
@cirospaciari want to take a look?

PR #7897, opened 2023-12-28
fix(bun.serve) avoid double ending chunk

cirospaciari:
### What does this PR do?

Fix: [#7555](https://github.com/oven-sh/bun/issues/7555)

- [x] Code changes


packages/bun-uws/src/HttpResponse.h
```diff
@@ -43,561 +43,624 @@ static const char *HTTP_200_OK = "200 OK";
...
-
-        if (!(httpResponseData->state & HttpResponseData<SSL>::HTTP_WRITE_CALLED)) {
-            /* Write mark on first call to write */
-            writeMark();
-
-            writeHeader("Transfer-Encoding", "chunked");
-            httpResponseData->state |= HttpResponseData<SSL>::HTTP_WRITE_CALLED;
-        }
+    /*  this will be sent always when state is HTTP_WRITE_CALLED inside internalEnd, so no need to write the terminating 0 chunk here
+     Super::write("\r\n0\r\n\r\n", 7); */
```
cirospaciari: the only change is that we comment on this part because we guarantee that state is always HTTP_WRITE_CALLED

packages/bun-uws/src/HttpResponse.h
```diff
@@ -43,561 +43,624 @@ static const char *HTTP_200_OK = "200 OK";
...
-             * This check also serves to limit writing the header only once. */
-            if ((httpResponseData->state & HttpResponseData<SSL>::HTTP_CONNECTION_CLOSE) == 0) {
-                writeHeader("Connection", "close");
+        /* Ignoring optional for now */
+        Super::write(data.data(), (int)data.length());
+      }
+
+      /* Terminating 0 chunk */
+      Super::write("\r\n0\r\n\r\n", 7);
+
```
cirospaciari: here is the place that will always send the end chunk when chunk-encoded

Jarred-Sumner:
for some reason it's not running tests

Jarred-Sumner:
Thank you

Issue #7892, opened 2023-12-28
Bun's `ws` mock is missing readonly instance members


mmtftr:
### What version of Bun is running?

1.0.8+2a405f691e80725fe0b97b93afd3b8cfed13fa5f

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

There is a missing instance member in the Bun mocked WebSocket. I found this bug due to a problem I had when I was using trpc with bun.

https://github.com/trpc/trpc/blob/4e101e3af7ce3ee92eb5c3f7d167d2a800e73a55/packages/server/src/adapters/ws.ts#L162

This line depends on this instance member being present. I checked @types/ws and it does indeed declare this to be an instance member in addition to being a static member.

```ts
import {Server} from 'ws';

  const wss = new ws.Server({
    host: "127.0.0.1",
    port: 3000,
  });

wss.on("connection", (client) => {
  console.log(client.OPEN); // undefined in Bun, 1 in "ws"
});
```

### What is the expected behavior?

I expected the instance members to be present.

### What do you see instead?

They are not.

### Additional information

My current workaround for trpc is to do this:

```ts
  wss.once('connection', (client) => {
    // @ts-ignore
    client.__proto__.OPEN = 1;
  });
```

vortex73:
I would like to work on this issue. Please elaborate on this.

PR #7898, opened 2023-12-28
fix: add ws properties to BunWebSocketMocked prototype

kaioduarte:
### What does this PR do?

Injects some properties to the `BunWebSocketMocked` prototype. Currently, these values are injected, [as in the original ws module](https://github.com/websockets/ws/blob/d343a0cf7bba29a4e14217cb010446bec8fdf444/lib/websocket.js#L499-L569), only for the `BunWebSocket` class, and not for the WebSocket instances from `WebSocketServer` or `Server`.

I'm only adding the properties to the prototype because the class is not exposed.

- [x] Code changes

Closes https://github.com/oven-sh/bun/issues/7892


Jarred-Sumner:
Thank you

PR #7903, opened 2023-12-29
cli: resolve TODO filing upstream issue

nektro:
filed the issue here https://github.com/ziglang/zig/issues/18397

however this is not a blocker for us because moving the `&` to the left would make the result a pointer to stack memory instead of comptime memory and introduce a bug


Jarred-Sumner:
thanks

PR #7905, opened 2023-12-29
misc: use compileError instead of unreachable for comptime-known switches

nektro:
this ensures the value stays comptime-known and prevents any stray `unreachable` in these cases from reaching runtime by accident as the code changes


Jarred-Sumner:
Thanks

PR #7907, opened 2023-12-29
tidy: bun.zig: remove switch on literal

nektro:
this code was previously copied from elsewhere but then never cleaned up when the flag was removed


Jarred-Sumner:
Thanks

PR #7822, opened 2023-12-25
 feat(server): add server backlog parameter 

Hanaasagi:
### What does this PR do?

This pull request adds the `backlog` parameter to both `Bun.listen` and `Bun.serve`. Additionally, it addresses the issue where the `backlog` parameter of the` server.listen` function on the Node.js side was not taking effect. The default value is `512`.

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

passed existing tests. And manually tested.

##### `Bun.serve`

```JavaScript
Bun.serve({
  hostname: "127.0.0.1",
  port: 8080,
  backlog: 1024,
  fetch(request) {
    console.log(request.url);
  },
});
```

##### `Bun.listen`

```JavaScript
Bun.listen({
  hostname: "127.0.0.1",
  port: 8080,
  backlog: 1024,
  socket: {
    data(socket, data) { },
  },
});
```

##### `http.createServer`

```JavaScript
const http = require("http");

const server = http.createServer((req, res) => {});
server.listen(8080, "127.0.0.1", 1024, () => {
  console.log("opened server on", server.address());
});

```

##### `net.createServer`

```JavaScript
const net = require("net");
const server = net.createServer();

server.listen(8080, "127.0.0.1", 1024, () => {
  console.log("opened server on", server.address());
});
```

Run the above code and see `send-q` via `ss` command

```
>> ss -tuln | grep 8080
tcp   LISTEN 0      2048       127.0.0.1:8080       0.0.0.0:*
```

Ref

https://www.linuxjournal.com/files/linuxjournal.com/linuxjournal/articles/023/2333/2333s2.html

PR #7908, opened 2023-12-29
Revert " feat(server): add server backlog parameter "

Jarred-Sumner:
Reverts oven-sh/bun#7822

The tests fail


Hanaasagi:
Are there any other failed tests besides this one?

PR #7909, opened 2023-12-29
Try to make fs.watch tests less flaky

Jarred-Sumner:
### What does this PR do?

- This makes `task_count` an `atomic.Value` to be extra careful when getting/setting the value
- Makes `close` happen in a microtask so that the watcher is already detached by the time the `close` event actually fires
- uses `bun.new` to more easily track allocations

This probably doesn't make the tests less flaky, but its worth an attempt

Ran the existing tests for this locally


KatSick:
I'm not sure what makes tests flaky, but it may be an issue with `fs.watch` itself. See https://github.com/oven-sh/bun/issues/7919

cirospaciari (approved):
LGTM `bun.new` is a nice addition

PR #7911, opened 2023-12-29
Make next-build test less flaky

Jarred-Sumner:
### What does this PR do?

This test was failing because a different number of spaces were being set in bun vs node 

Tests


paperclover (changes requested):
it fails on linux. hashes of the files are the same, but the file sizes are off by one. perhaps we are emitting a newline somewhere?

jdalton (dismissed):

jdalton:
> it fails on linux. hashes of the files are the same, but the file sizes are off by one. perhaps we are emitting a newline somewhere?

The code change in this PR is for when the test is already failing. Maybe the tweak should be `replace(/\s+/g, '')`

PR #7958, opened 2024-01-03
node:child_process: support defining extra pipes

nektro:
adds support in `require("node:child_process").spawn` for specifying extra file descriptors that are pipes to be sent to the child process. also fixes other small discrepancies we had with the node api in this area.


src/bun.js/api/bun.classes.ts
```diff
@@ -107,10 +107,12 @@ export default [
       signalCode: {
         getter: "getSignalCode",
       },
-
       exited: {
         getter: "getExited",
       },
+      pipe_fds: {
```
Jarred-Sumner: lets not add this getter to the api

instead, lets add a `stdio` getter which returns an array of `[this.stdin, this.stdout, this.stderr, ...rest]`

src/bun.js/api/bun/subprocess.zig
```diff
@@ -1797,6 +1829,26 @@ pub const Subprocess = struct {
...
         ) catch |err| return globalThis.handleError(err, "in configuring child stderr");
 
+        for (stdio_pipes.slice()) |*item| {
+            const maybe = blk: {
+                var fds: [2]c_int = undefined;
+                const rc = std.os.system.socketpair(os.AF.UNIX, os.SOCK.STREAM, 0, &fds);
+                switch (std.os.system.getErrno(rc)) {
+                    .SUCCESS => {},
+                    .AFNOSUPPORT => break :blk error.AddressFamilyNotSupported,
+                    .FAULT => unreachable,
```
Jarred-Sumner: error.Fault or error.Unexpected

src/js/node/stream.js
```diff
@@ -80,7 +78,7 @@ var require_primordials = __commonJS({
   "node_modules/readable-stream/lib/ours/primordials.js"(exports, module) {
     "use strict";
     module.exports = {
-      ArrayIsArray(self) {
+      $isJSArray(self) {
```
Jarred-Sumner: this is not valid

PR #7962, opened 2024-01-03
feat(worker_threads) support argv and execArgv options

### LINKED ISSUE #4130: worker_threads.Worker missing `argv` and `execArgv` option
https://nodejs.org/api/worker_threads.html#new-workerfilename-options

`argv` modifies the inner `process.argv`. we can copy how `env` is handled but slightly different semantics. should be an easy issue.

`execArgv` has more semantics in node.js, since those options can control the runtime. but we dont implement these flags. what we will do for this issue is simply make it visible in `process.execArgv`, so basically the same as `process.argv` but i think this is an override instead of an append.

these two keys can be added to the global worker, and then the warnings from `src/js/node/worker_threads.ts` can be removed.

otgerrogla:
### What does this PR do?

Adds support for arguments to NodeJS `worker_threads` and web Worker constructor options:
- `argv`: set the additional arguments that worker threads see at `process.argv`/`Bun.argv` after `argv[0]` (executable path) and `argv[1]` (main script path).
- `execArgv`: bun does not use it, but added for compatibility. When the option is set, workers threads see the given string array values at `process.execArgv`, otherwise the value of this defaults to the one of the main thread.

Also fixes a small bug where worker threads always had `"bun"` as 3rd arg (`argv[2]`). Now when `argv` option is not specified, threads only see `argv[0]` and `argv[1]` in `process.argv`.

Closes #4130

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
This looks great. Thank you for this. Let's let the tests run and merge if the worker tests pass

Jarred-Sumner:
I made the macOS CI runners bigger to hopefully fix the CI issue

PR #7966, opened 2024-01-03
fix(console.table): fixes for character width breaking the layout

### LINKED ISSUE #7955: `console.table()` doesn't handle emoji widths properly

1.0.21+837cbd60d

Linux 6.6.9-native_amd-xanmod1-1 x86_64 unknown

1. `console.table([{"test": "✅"}, {"test": "✅"}, {"test": "✅"}, {"test": "✅"}])`
2. `console.table([{"test 1": "✅"}, {"test 2": "✅"}, {"test 3": "✅"}, {"test 4": "✅"}])`

Non-broken tables

### What do you see instead?

1.
image

2.
image

### Additional information

Terminal is kitty with the Cartograph CF font

otgerrogla:
### What does this PR do?

Contains 3 fixes:
- Replaces the algorithm for determining which unicode chars are full-width, by an auto-generated switch based on [official Unicode data](https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt). Specifically, characters tagged by unicode as `'W'` (wide) and `'F'` (full-width) are set to have width 2, the rest to 1.
    - Fixes several cases of emojis such as ✅ breaking the layout (the algorithm was outdated)
    - Fixes #7955
- Adds zero-width space (formerly known as BOM) as character with width 0
- Fixes bug where if an object-row did not have value for a given column (`obj.get()` returned `JSValue.empty`), an extra space was appended which broke the table layout.

- [x] Code changes

Fixed the snapshot test for the third issue. For the other two, wrote a test but it cannot be automated because there's no way to measure the actual printed width. Visual output: (the test prints about 2000 of the most common unicode codepoints, was run on a macOS)

image


test/js/bun/console/console-table.test.ts
```diff
@@ -157,3 +157,24 @@ test("console.table json fixture", () => {
   expect(actualOutput).toMatchSnapshot();
   console.log(actualOutput);
 });
+
+test.skip("console.table character widths", () => {
+  // note: this test cannot be automated because cannot test printed witdhs consistently.
```
Electroid, line 162: Not sure I understand this, why can't we take a snapshot of the stdout?

ThatOneCalculator: Also, comment should be "widths" not "witdhs"

otgerrogla: The unicode characters that are printed will be always output the same bytes, so a snapshot would not help.
What the test should look at is how they are actually visually printed in a terminal/tty, but there's no easy way to do that, even more if it has to support running on CI without actual visual terminals

otgerrogla: Even so, anyways how it actually ends up printed is gonna depend on the font configured by the user, the fallback fonts in the OS, etc.
So the safest bet is to rely the Unicode specs on width

Jarred-Sumner:
Thank you!

PR #7976, opened 2024-01-03
Add brotli test

Jarred-Sumner:
### What does this PR do?

Fixes https://discord.com/channels/876711213126520882/1192197434827345930/1192197434827345930

There is a test


test/js/web/fetch/fetch.brotli.test.ts
```diff
@@ -0,0 +1,25 @@
+import { test, expect } from "bun:test";
+
+test("fetch brotli response works", async () => {
+  const [firstText, secondText, { headers }] = await Promise.all([
+    fetch("https://bun.sh/logo.svg", {
```
Electroid, line 5: Can we use `example.com` instead?

Jarred-Sumner: no, example.com doesn't return with brotli

PR #7982, opened 2024-01-04
fix: properly transform http method to uppercase

### LINKED ISSUE #7250: [Regression] Update from 1.0.9 to 1.0.10 break fastify cors preflight

1.0.13

Macos 16.6 (Darwin 22.6.0 arm64 arm)

Before upgrading bun to 1.0.13, It was on 1.0.7 and was working smoothly.

After upgrade, the fastify cors preflight OPTIONS requests are broken, and return the following : 
```json
{
  "statusCode": 404,
  "code": "FST_ERR_NOT_FOUND",
  "error": "Not Found",
  "message": "Not Found"
}
```

Please note than it works with nodejs, as well as previous version of bun, so I'm quite confident that its a regression caused by bun upgrade. Reproduction sample : 

Package.json setup : 
```json
...
"scripts": {
    "start": "bun --watch main.js"
},
"dependencies": {
    "fastify": "^4.0.0",
    "@fastify/cors": "^8.0.0"
},
...
```

main.js
```js
// Importing Fastify and CORS plugin
import Fastify from 'fastify'
import cors from "@fastify/cors"

// Create a Fastify instance
const fastify = Fastify()

// Register CORS plugin
fastify.register(cors, { 
  origin: true
})

// Declare a route
fastify.get('/', async (request, reply) => {
  return { message: 'Hello World' }
})

// Run the server
const start = async () => {
  try {
    await fastify.listen({ port: 3000 })
    console.log(`Server listening on ${fastify.server.address().port}`)
  } catch (err) {
    fastify.log.error(err)
    process.exit(1)
  }
}

start()
``` 

`bun run start`

OPTIONS requests working as expected

### What do you see instead?

See above

### Additional information

_No response_

asomethings:
### What does this PR do?

Transforming HTTP method names to uppercase inside `assignHeadersFromUWebSockets` uses wrong HTTP Method length which causes `PURGE` / `OPTIONS` methods in lowercase.

I assume this will fix #7250

- [x] Code changes

I wrote automated tests

- [x] I ran my tests locally and they pass (`bun-debug test test/js/node/http/node-http.test.ts`)

### DIFF
--- a/src/bun.js/bindings/NodeHTTP.cpp
+++ b/src/bun.js/bindings/NodeHTTP.cpp
@@ -154,6 +154,10 @@ static EncodedJSValue assignHeadersFromUWebSockets(uWS::HttpRequest* request, JS
                 methodString = "FETCH"_s;
                 break;
             }
+            if (methodView == std::string_view("purge", 6)) {
+                methodString = "PURGE"_s;
+                break;
+            }
 
             break;
         }
@@ -161,10 +165,6 @@ static EncodedJSValue assignHeadersFromUWebSockets(uWS::HttpRequest* request, JS
         case 6: {
             if (methodView == std::string_view("delete", 6)) {
                 methodString = "DELETE"_s;
-                break;
-            }
-            if (methodView == std::string_view("purge", 6)) {
-                methodString = "PURGE"_s;
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you

PR #7991, opened 2024-01-05
fix(install): incremental support for windows

dylan-conway:
### What does this PR do?
This branch has a few fixes for bun install on windows. The following should be working:
- libarchive
- installing with copy file
- installing with hardlinks
- packages with binaries
- installing from install cache

Support for lifecycle scripts on windows will be added in another pr

tested manually


src/bun.js/node/node_fs.zig
```diff
@@ -4273,15 +4273,15 @@ pub const NodeFS = struct {
         };
     }
 
-    pub fn _isSep(char: std.meta.Child(bun.OSPathSlice)) bool {
+    pub fn _isSep(char: bun.OSPathChar) bool {
```
paperclover, line 4276: yay thank you for using the new types

src/bun.zig
```diff
@@ -32,799 +32,9 @@ pub const ComptimeStringMap = @import("./comptime_string_map.zig").ComptimeStrin
...
-        return std.fmt.formatBuf(fbs.getWritten(), opts, writer);
-    }
-
-    /// Return a Formatter for number of nanoseconds according to its magnitude:
-    /// [#y][#w][#d][#h][#m]#[.###][n|u|m]s
-    pub fn fmtDurationOneDecimal(ns: u64) std.fmt.Formatter(formatDurationOneDecimal) {
-        return .{ .data = FormatDurationData{ .ns = ns } };
-    }
-};
+pub const fmt = @import("./fmt.zig");
```
paperclover, line 37: thank you

src/bun.zig
```diff
@@ -2788,6 +2014,174 @@ pub inline fn pathLiteral(comptime literal: anytype) *const [literal.len:0]u8 {
...
+            else => .posix,
+        }, T).init(path_);
+    }
+};
+
+pub const Dirname = struct {
+    /// copy/paste of `std.fs.path.dirname` and related functions and modified to support u16 slices.
+    /// inside `Dirname` scope to make deleting later easier.
+    /// TODO(dylan-conway) delete `Dirname`
+    pub fn dirname(comptime T: type, path_: []const T) ?[]const T {
```
paperclover, line 2131: is there a reason we have this instead of our own implementation***s*** of dirname:

- resolver.zig's Dirname.dirname
- resolve_path.zig's dirname

src/install/bin.zig
```diff
@@ -352,6 +355,184 @@ pub const Bin = extern struct {
             setPermissions(this.package_installed_node_modules, dest_path);
         }
 
+        const cmd_contents_begin =
```
paperclover: is there actually nothing we can do about this? ive always felt this is really big but at the same time, it is probably some crazy compatibility things

src/install/bin.zig
```diff
@@ -458,7 +634,15 @@ pub const Bin = extern struct {
...
 
-                    this.setSimlinkAndPermissions(target_path, dest_path);
+                    if (comptime Environment.isWindows) {
+                        var dest_path_buf: bun.PathBuffer = undefined;
+                        @memcpy(dest_path_buf[0..dest_path.len], dest_path);
+                        this.createCmdFile(target_path, &dest_path_buf, dest_path);
+                        this.createPs1File(target_path, &dest_path_buf, dest_path);
+                        this.createShFile(target_path, dest_path);
+                    } else {
+                        this.setSimlinkAndPermissions(target_path, dest_path);
```
paperclover: simlink? symlink?

src/install/bin.zig
```diff
@@ -458,7 +634,15 @@ pub const Bin = extern struct {
                     from_remain[0] = 0;
                     const dest_path: [:0]u8 = target_buf[0 .. @intFromPtr(from_remain.ptr) - @intFromPtr(&target_buf) :0];
 
-                    this.setSimlinkAndPermissions(target_path, dest_path);
+                    if (comptime Environment.isWindows) {
+                        var dest_path_buf: bun.PathBuffer = undefined;
+                        @memcpy(dest_path_buf[0..dest_path.len], dest_path);
+                        this.createCmdFile(target_path, &dest_path_buf, dest_path);
+                        this.createPs1File(target_path, &dest_path_buf, dest_path);
+                        this.createShFile(target_path, dest_path);
```
paperclover: i guess its okay if the answer is yes but do we really need all three of these.

(also note: powershell is VERY slow at startup time, like an order of magnitude more so)

src/install/install.zig
```diff
@@ -1349,26 +1396,20 @@ const PackageInstall = struct {
...
                                 if (entry.path.len > to_copy_into1.len or entry.path.len > to_copy_into2.len) {
                                     return error.NameTooLong;
                                 }
 
-                                // TODO: this copy shouldn't be necessary in the first place.
-                                strings.copyU8IntoU16(
-                                    to_copy_into1,
-                                    entry.path,
-                                );
+                                @memcpy(to_copy_into1[0..entry.path.len], entry.path);
```
paperclover, line 1408: nice

src/install/lockfile.zig
```diff
@@ -1658,15 +1658,13 @@ pub fn saveToDisk(this: *Lockfile, filename: stringZ) void {
...
-    if (comptime Environment.isWindows) {
-        // TODO: make this executable
-        @panic("TODO on Windows");
-    } else {
-        _ = C.fchmod(
-            tmpfile.fd,
-            // chmod 777
-            0o0000010 | 0o0000100 | 0o0000001 | 0o0001000 | 0o0000040 | 0o0000004 | 0o0000002 | 0o0000400 | 0o0000200 | 0o0000020,
-        );
+    switch (bun.sys.fchmod(tmpfile.fd, 0o0000010 | 0o0000100 | 0o0000001 | 0o0001000 | 0o0000040 | 0o0000004 | 0o0000002 | 0o0000400 | 0o0000200 | 0o0000020)) {
```
paperclover: please add comment from before
`// chmod 777`
maybe we ought to just inline it `0o1777`

src/libarchive/libarchive.zig
```diff
@@ -510,7 +506,8 @@ pub const Archive = struct {
                         }
                     }
 
-                    var tokenizer = std.mem.tokenize(u8, bun.asByteSlice(pathname), std.fs.path.sep_str);
+                    // TODO(dylan-conway): I think this should only be '/'
+                    var tokenizer = std.mem.tokenizeAny(u8, bun.asByteSlice(pathname), "/\\");
```
paperclover: you're probably right since this is parsing of (cross platform) tarball files.
we shouldnt do tokenizeAny on posix though, i think it is a little slower.

src/libarchive/libarchive.zig
```diff
@@ -616,24 +617,26 @@ pub const Archive = struct {
...
                                 }
                                 // archive_read_data_into_fd reads in chunks of 1 MB
                                 // #define	MAX_WRITE	(1024 * 1024)
-                                if (size > 1_000_000) {
-                                    C.preallocate_file(
-                                        file.handle,
-                                        0,
-                                        entry_size,
-                                    ) catch {};
+                                if (comptime Environment.isLinux) {
```
paperclover, line 630: do we need the if check. these functions are defined in both darwin_c and windows_c but are empty.

src/output.zig
```diff
@@ -186,7 +186,7 @@ pub var stderr_descriptor_type = OutputStreamDescriptor.unknown;
 pub var stdout_descriptor_type = OutputStreamDescriptor.unknown;
 
 pub inline fn isEmojiEnabled() bool {
-    return enable_ansi_colors and !Environment.isWindows;
+    return enable_ansi_colors;
```
paperclover, line 189: oh lol. thats cool, now emojis work. we might have to do something later if conhost doesnt support emojis (or version checks to when conhost didn't)

src/string_immutable.zig
```diff
@@ -3442,6 +3295,15 @@ pub fn isValidUTF8(slice: []const u8) bool {
 }
 
 pub fn isAllASCII(slice: []const u8) bool {
+    if (@inComptime()) {
```
paperclover, line 3302: NICE

src/walker_skippable.zig
```diff
@@ -36,73 +42,83 @@ pub fn next(self: *Walker) !?WalkerEntry {
...
-                {
-                    errdefer new_dir.close();
-                    try self.stack.append(StackItem{
-                        .iter = new_dir.iterate(),
-                        .dirname_len = self.name_buffer.items.len,
-                    });
-                    top = &self.stack.items[self.stack.items.len - 1];
+        switch (top.iter.next()) {
+            .err => {
+                // TODO(dylan-conway):
```
paperclover: have to handle this

src/windows_c.zig
```diff
@@ -1195,6 +1195,46 @@ pub const E = enum(u16) {
     UV_EUNATCH = -uv.UV_EUNATCH,
 };
 
+pub const S = struct {
```
paperclover, line 1198: can you change uses of `bun.windows.libuv.S` to here. and then also fallback to the real `S` in case that `S` is different.

paperclover (changes requested):
very very very exciting and great work so far

i left notes

src/install/bin.zig
dylan-conway: lol

src/deps/zlib.win32.zig
```diff
@@ -10,7 +10,7 @@ const voidpf = *anyopaque;
 const Bytef = [*]u8;
 const Byte = u8;
 const uInt = c_uint;
-const uLong = c_ulong;
+const uLong = u64;
```
Jarred-Sumner, line 13: is this right? cc @cirospaciari

src/install/bin.zig
```diff
@@ -352,6 +355,193 @@ pub const Bin = extern struct {
...
+                    .err => |err| {
+                        this.err = @errorFromInt(err.errno);
+                        return;
+                    },
+                }
+            }
+        }
+
+        const ps1_contents_fmt =
+            \\#!/usr/bin/env pwsh
```
Jarred-Sumner: @zackradisic wdyt about using bun shell here

src/deps/zlib.win32.zig
dylan-conway: without this change zlib errors with `InvalidArgument` because internally it checks for expected size of structs
Screenshot 2024-01-10 at 2 13 41 PM

src/libarchive/libarchive.zig
dylan-conway: Since they aren't doing anything on windows or macos, might as well remove the function call and `if (size < ...)` at comptime

src/bun.zig
dylan-conway: keeping this because it was easy to add u16 slice support, and there seems to be small differences between this dirname and resolve_path.zig/resolver.zig dirname (returning the root instead of null is the main one I'm seeing). Still, this should be removed in the near future

src/install/bin.zig
dylan-conway: copying npm/pnpm behavior. By default all are created, but should look into removing these with bun shell

src/install/bin.zig
zackradisic: Would need to support if statements (e.g. `if [ -x "$basedir/node" ]; then...`)

Would need $0 so `basedir=$(dirname echo $0)` works

And also would be nice to have `dirname` implemented in bun shell. Maybe we can have a custom command for getting the basedir itself

Issue #7971, opened 2024-01-03
segfault in `bun c -h`


benwaffle:
### What version of Bun is running?

1.0.21+837cbd60d

### What platform is your computer?

Linux 6.6.5-zen1-1-zen x86_64 unknown

### What steps can reproduce the bug?

1. Run `bun c -h`
2. Segfault

### What is the expected behavior?

No segfault

### What do you see instead?

```
❯ bun c -h
[1]    1676198 segmentation fault (core dumped)  bun c -h
```

gyf304:
`uname -a`: `Linux --- 5.15.0-89-generic #99-Ubuntu SMP Mon Oct 30 20:42:41 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
Can reproduce `1.0.21 (837cbd60)` with trace
```
(gdb) r c -h
Starting program: /home/---/.bun/bin/bun c -h
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x0000555558023510 in ?? ()
(gdb) bt
#0  0x0000555558023510 in ?? ()
#1  0x0000555557de8549 in ?? ()
#2  0x0000555557de5d1e in ?? ()
#3  0x00007ffff7db7d90 in __libc_start_call_main (main=main@entry=0x555557de59f0, argc=argc@entry=3, argv=argv@entry=0x7fffffffde38)
    at ../sysdeps/nptl/libc_start_call_main.h:58
#4  0x00007ffff7db7e40 in __libc_start_main_impl (main=0x555557de59f0, argc=3, argv=0x7fffffffde38, init=<optimized out>, 
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffde28) at ../csu/libc-start.c:392
#5  0x0000555557de592a in ?? ()
```

`bun repl` also crashes with SEGFAULT, with backtrace
```
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=11, threadid=140737350570816) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=11, threadid=140737350570816) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737350570816, signo=signo@entry=11) at ./nptl/pthread_kill.c:89
[...]

PR #7995, opened 2024-01-05
fix: crash in `bun create -h` #7971

coratgerl:
### What does this PR do?

Fixes: #7971 
Refactor the create command to match with the architecture of the other command.

- [x] Code changes

I wrote automated tests 


Jarred-Sumner:
It should print the help menu, not a new error

coratgerl:
Indeed, I made the change to display the help menu.

- For this I made a refactor to have the same architecture as for the other command (the file is more clear now)
- I also fixed a test that I don't understand how it passed before, indeed, on the screen below we can see the template is create under the `bun-create` folder(in main branch), this is the expected behavior but the test checked if file exist in `test-template`and not in bun-create as you can see on the second screen :  
```javascript 
const bunCreateDir = join(x_dir, "bun-create");
  const testTemplate = "test-template";

  await mkdir(`${bunCreateDir}/${testTemplate}`, { recursive: true });
  const { exited } = spawn({
    cmd: [bunExe(), "create", testTemplate],
    stdout: null,
    stdin: "pipe",
    stderr: "pipe",
    cwd: x_dir,
    env: { ...env, BUN_CREATE_DIR: bunCreateDir },
  });

  await exited;

  const dirStat = await stat(`${x_duir}/${testTemplate}`);
  expect(dirStat.isDirectory()).toBe(true);
```

 
image

src/cli.zig
```diff
@@ -1988,27 +2022,31 @@ pub const Command = struct {
...
+                        \\<b>Usage<r>: <b><green>bun create<r> <cyan>\<template\><r> [\<destination\>]
+                        \\<b>Alias: <b>bun c<r>
+                        \\Create a new bun project using a create-\<template\> npm package, GitHub repo, or a local template
                     ;
 
                     const outro_text =
-                        \\If given a GitHub repository name, Bun will download it and use it as a template,
-                        \\otherwise it will run <b><magenta>bunx create-\<template\><r> with the given arguments.
+                        \\<b>Examples:<r>
+                        \\  <b>From npm :<r>
```
paperclover:
```suggestion
                        \\  <d>From npm<r>
```

src/cli.zig
```diff
@@ -1988,27 +2022,31 @@ pub const Command = struct {
...
+                        \\  <b><green>bun create remix<r>
+                        \\
+                        \\  <b>From GitHub :<r>
+                        \\  <b><green>bun create github.com/\<user\>/\<repo\><r>
```
paperclover:
```suggestion
                        \\  <d>From GitHub<r>
                        \\  <b><green>bun create \<user\>/\<repo\><r>
```
maybe we can pick an actual example, idk

paperclover (changes requested):

src/cli.zig
coratgerl: I think we can keep the basic example of the doc as you put in your suggestion. The link to the full documentation is in the bottom of the man command :)

src/cli.zig
```diff
@@ -1988,27 +2022,31 @@ pub const Command = struct {
...
+                        \\<b>Examples:<r>
+                        \\  <d>From npm<r>
+                        \\  <b><green>bun create remix<r>
+                        \\
+                        \\  <d>From GitHub<r>
+                        \\  <b><green>bun create \<user\>/\<repo\><r>
                         \\
-                        \\Learn more about creating new projects: <magenta>https://bun.sh/docs/cli/bun-create<r>
+                        \\  <d>From local template<r>
+                        \\  <b><green>bun create $HOME/.bun-create/\<name\><r>
```
paperclover, line 2088: wait does this work?

coratgerl: I took this elements from the documentation but I don't tested. If the documentation is correct I suppose yes it's work

paperclover (approved):
sorry for late review. this only really needed to change one thing but the rest seems ok.

coratgerl:
Thanks for your feedback, the conflict is fixed @paperdave

Issue #7385, opened 2023-11-30
cbor@9.0.1 decoding not working as expected


alleneubank:
### What version of Bun is running?

1.0.14+d8be3e51b

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

Execute the following code using `bun run`

```js
import cbor from 'cbor'

// const m = new Map()
// m.set(1, 2)
// console.log(cbor.encode(m).toString('hex')) // <Buffer a1 01 02>

const map = new Map()
map.set(1, 2)
map.set(3, -7)
map.set(-1, 1)
map.set(
  -2,
  Buffer.from(
    'a9 60 7c 40 66 b1 3a f4 54 f6 41 11 bf 51 55 dd 7c e1 60 25 1b 54 dc 58 b6 52 24 f7 11 38 b1 45'.replace(
      / /g,
      ''
    ),
    'hex'
  )
)
map.set(
  -3,
  Buffer.from(
    '08 9d 70 4a 02 e2 41 2f d9 3a bb b8 69 3c 95 a7 d2 c5 51 f1 02 6a ff ec f1 b2 1e b9 ce 27 84 ad'.replace(
      / /g,
      ''
    ),
    'hex'
  )
)

const encoded = cbor.encode(map)

console.log(cbor.decodeAllSync(encoded))
```

### What is the expected behavior?

When running with `ts-node` it decodes the cbor encoding without issue.

```
bunx ts-node  ./scratch/cbor-testing.js
[
  Map(5) {
    1 => 2,
    3 => -7,
    -1 => 1,
    -2 => <Buffer a9 60 7c 40 66 b1 3a f4 54 f6 41 11 bf 51 55 dd 7c e1 60 25 1b 54 dc 58 b6 52 24 f7 11 38 b1 45>,
    -3 => <Buffer 08 9d 70 4a 02 e2 41 2f d9 3a bb b8 69 3c 95 a7 d2 c5 51 f1 02 6a ff ec f1 b2 1e b9 ce 27 84 ad>
  }
]
```

### What do you see instead?

It fails with error:
```
268 | 
269 |       while (!state.done) {
270 |         const b = s.read(state.value)
271 | 
272 |         if ((b == null) || (b.length !== state.value)) {
273 |           throw new Error('Insufficient data')     

error: Insufficient data
      at decodeAllSync (/Users/0xbigboss/0xsend/sendapp/node_modules/cbor/lib/decoder.js:273:16)
[...]

PR #8007, opened 2024-01-05
Fix multiple partial consume from BufferList

hborchardt:
### What does this PR do?

The JSUint8Array::possiblySharedBuffer() returns the backing array,
not taking into account the byteOffset that indicates the start of
the data in the backing array. This means that when creating an array
with the same backing array, the current byteOffset needs to be added
to the start of the new slice.

This led to consume() returning the same data when repeatedly consuming
small numbers of bytes from the BufferList.

This PR fixes the problem in https://github.com/oven-sh/bun/issues/7385

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
thank you for fixing this

Jarred-Sumner (approved):

Issue #6785, opened 2023-10-29
(docs) setup error: 'ruby' is missing


gamedevsam:
### What is the type of issue?

Documentation issue.

### What is the issue?

I attempted to follow the contributing guide, but it errors out with this error:

```
samuel@SAMUEL-DESKTOP:~/WSL_Dev/Personal/bun$ bun setup
$ ./scripts/setup.sh
setup error: 'ruby' is missing
error: script "setup" exited with code 1 (SIGHUP)
```

### Where did you find it?

https://bun.sh/docs/project/contributing

Hanaasagi:
Hi, I see that doc has been updated.
- https://github.com/oven-sh/bun/pull/6787
- https://github.com/oven-sh/bun/pull/6788

PR #8012, opened 2024-01-06
Update contributing.md

karmabadger:
Adding ruby to dependencies installation for macOS and Fedora

### What does this PR do?

- [ ] Code changes


karmabadger:
as per https://github.com/oven-sh/bun/issues/6785 and https://github.com/oven-sh/bun/pull/6788
ruby was missing for dependencies to install for both macOS and Fedora. Somehow, it was added for both arch and ubuntu/debian but not the rest. 
This PR should fix it.

PR #8013, opened 2024-01-06
Fix using two const result=... in same script

guest271314:
```
  entrypoints: ["./index.ts"],
});

for (const result of result.outputs) {
  // Can be consumed as blobs
  await result.text();

  // Bun will set Content-Type and Etag headers
  new Response(result);

  // Can be written manually, but you should use `outdir` in this case.
  Bun.write(path.join("out", result.path), result);
}
```

throws

```
                         ^
ReferenceError: Cannot access uninitialized variable.
```
### What does this PR do?

Changes duplicate `const result =...`

- [ X] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

Ran the code in terminal.

### DIFF
--- a/docs/bundler/index.md
+++ b/docs/bundler/index.md
@@ -276,15 +276,15 @@ const result = await Bun.build({
   entrypoints: ["./index.ts"],
 });
 
-for (const result of result.outputs) {
+for (const res of result.outputs) {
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you

PR #8022, opened 2024-01-06
all: make bun.FileDescriptor a unique type

nektro:
ensure better type safety by making `bun.FileDescriptor` a unique type. by being an enum we can not only ensure unintentional c_int's don't get passed as FD's while keeping the same abi size and behavior, but also add convenience methods for interacting with them.

### DIFF
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1813,6 +1813,7 @@ pub const Subprocess = struct {
             }
         }
 
+        // TODO: move pipe2 to bun.sys so it can return [2]bun.FileDesriptor
         const stdin_pipe = if (stdio[0].isPiped()) os.pipe2(0) catch |err| {
             globalThis.throw("failed to create stdin pipe: {s}", .{@errorName(err)});
             return .zero;
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -2658,12 +2658,11 @@ pub const Subprocess = struct {
 
             return true;
         } else if (value.isNumber()) {
-            const fd_ = value.toInt64();
-            if (fd_ < 0) {
+            const fd = value.asFileDescriptor();
+            if (fd.int() < 0) {
                 globalThis.throwInvalidArguments("file descriptor must be a positive integer", .{});
                 return false;
             }
-            const fd = bun.toFD(fd_);
 
             switch (bun.FDTag.get(fd)) {
                 .stdin => {
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1646,9 +1646,9 @@ pub const Subprocess = struct {
             const stderr_pipe = if (stdio[2].isPiped()) stdio[2].makeUVPipe(globalThis) orelse return .zero else undefined;
 
             var uv_stdio = [3]uv.uv_stdio_container_s{
-                stdio[0].setUpChildIoUvSpawn(bun.posix.STDIN_FD, stdin_pipe[0]) catch |err| return globalThis.handleError(err, "in setting up uv_process stdin"),
-                stdio[1].setUpChildIoUvSpawn(bun.posix.STDOUT_FD, stdout_pipe[1]) catch |err| return globalThis.handleError(err, "in setting up uv_process stdout"),
-                stdio[2].setUpChildIoUvSpawn(bun.posix.STDERR_FD, stderr_pipe[1]) catch |err| return globalThis.handleError(err, "in setting up uv_process stderr"),
+                stdio[0].setUpChildIoUvSpawn(0, stdin_pipe[0]) catch |err| return globalThis.handleError(err, "in setting up uv_process stdin"),
+                stdio[1].setUpChildIoUvSpawn(1, stdout_pipe[1]) catch |err| return globalThis.handleError(err, "in setting up uv_process stdout"),
+                stdio[2].setUpChildIoUvSpawn(2, stderr_pipe[1]) catch |err| return globalThis.handleError(err, "in setting up uv_process stderr"),
             };
 
             var cwd_resolver = bun.path.PosixToWinNormalizer{};
@@ -1682,13 +1682,13 @@ pub const Subprocess = struct {
                 .globalThis = globalThis,
                 .pid = subprocess.pid,
                 .pidfd = 0,
-                .stdin = Writable.init(stdio[0], stdin_pipe[1], globalThis) catch {
+                .stdin = Writable.init(stdio[0], bun.toFD(stdin_pipe[1]), globalThis) catch {
                     globalThis.throwOutOfMemory();
                     return .zero;
                 },
                 // stdout and stderr only uses allocator and default_max_buffer_size if they are pipes and not a array buffer
-                .stdout = Readable.init(stdio[1], stdout_pipe[0], jsc_vm.allocator, default_max_buffer_size),
-                .stderr = Readable.init(stdio[2], stderr_pipe[0], jsc_vm.allocator, default_max_buffer_size),
+                .stdout = Readable.init(stdio[1], bun.toFD(stdout_pipe[0]), jsc_vm.allocator, default_max_buffer_size),
+                .stderr = Readable.init(stdio[2], bun.toFD(stderr_pipe[0]), jsc_vm.allocator, default_max_buffer_size),
                 .on_exit_callback = if (on_exit_callback != .zero) JSC.Strong.create(on_exit_callback, globalThis) else .{},
 
                 .ipc_mode = ipc_mode,
@@ -2662,7 +2662,6 @@ pub const Subprocess = struct {
                 globalThis.throwInvalidArguments("file descriptor must be a positive integer", .{});
                 return false;
             }
-
             const fd = bun.toFD(fd_);
 
             switch (bun.FDTag.get(fd)) {
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1575,12 +1575,12 @@ pub const Subprocess = struct {
                     }
 
                     if (args.get(globalThis, "stderr")) |value| {
-                        if (!extractStdio(globalThis, 1, value, &stdio[1]))
+                        if (!extractStdio(globalThis, 2, value, &stdio[2]))
                             return .zero;
                     }
 
                     if (args.get(globalThis, "stdout")) |value| {
-                        if (!extractStdio(globalThis, 2, value, &stdio[2]))
+                        if (!extractStdio(globalThis, 1, value, &stdio[1]))
                             return .zero;
                     }
                 }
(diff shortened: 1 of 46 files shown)


src/bun.js/bindings/bindings.zig
```diff
@@ -4856,6 +4859,11 @@ pub const JSValue = enum(JSValueReprInt) {
         return FFI.JSVALUE_TO_INT32(.{ .asJSValue = this });
     }
 
+    pub fn asFileDescriptor(this: JSValue) bun.FileDescriptor {
+        std.debug.assert(this.isNumber());
+        return bun.FDImpl.fromUV(this.asInt32()).encode();
+    }
```
nektro: is this lossy on windows? I see there's also the method `value.toInt64()`

src/bun.js/bindings/bindings.zig
```diff
@@ -3873,6 +3873,9 @@ pub const JSValue = enum(JSValueReprInt) {
                 if (comptime Number == std.os.fd_t) {
                     return jsNumber(bun.toFD(number));
                 }
+                if (Number == bun.FileDescriptor) {
+                    return jsNumber(number.int());
+                }
```
nektro, line 3878: oh I wonder if I should've used cast here

src/bun.js/node/node_fs.zig
```diff
@@ -6463,7 +6463,7 @@ pub const NodeFS = struct {
             .result => {},
         }
 
-        const dir = std.fs.Dir{ .fd = fd };
+        const dir = std.fs.Dir{ .fd = fd.int() };
```
nektro: this can be `const dir = fd.asDir();`

src/bun.js/node/path_watcher.zig
```diff
@@ -707,7 +707,7 @@ pub const PathWatcher = struct {
         if (comptime Environment.isMac) {
             if (!path.is_file) {
                 var buffer: [bun.MAX_PATH_BYTES]u8 = undefined;
-                const resolved_path_temp = std.os.getFdPath(path.fd, &buffer) catch |err| {
+                const resolved_path_temp = std.os.getFdPath(path.fd.int(), &buffer) catch |err| {
```
nektro: since this is in a macOS branch it'll be fine but ideally this'd use cast

src/bun.js/webcore/blob.zig
```diff
@@ -3367,7 +3367,7 @@ pub const Blob = struct {
                         this.max_length > bun.C.preallocate_length and
                         this.max_length != Blob.max_size)
                     {
-                        bun.C.preallocate_file(this.destination_fd, 0, this.max_length) catch {};
+                        bun.C.preallocate_file(this.destination_fd.int(), 0, this.max_length) catch {};
```
nektro: cast

src/bun.js/webcore/blob.zig
```diff
@@ -3420,7 +3420,7 @@ pub const Blob = struct {
                         return;
                     };
                     if (stat.size != 0 and @as(SizeType, @intCast(stat.size)) > this.max_length) {
-                        _ = darwin.ftruncate(this.destination_fd, @as(std.os.off_t, @intCast(this.max_length)));
+                        _ = darwin.ftruncate(this.destination_fd.int(), @as(std.os.off_t, @intCast(this.max_length)));
```
nektro: cast

src/bun.js/webcore/streams.zig
```diff
@@ -4236,9 +4236,9 @@ pub const File = struct {
         };
 
         if (comptime Environment.isPosix) {
-            if ((file.is_atty orelse false) or (fd < 3 and std.os.isatty(fd))) {
+            if ((file.is_atty orelse false) or (fd.int() < 3 and std.os.isatty(fd.int()))) {
```
nektro: isatty cast

src/bun.js/webcore/streams.zig
```diff
@@ -4236,9 +4236,9 @@ pub const File = struct {
...
                 var termios = std.mem.zeroes(std.os.termios);
-                _ = std.c.tcgetattr(fd, &termios);
+                _ = std.c.tcgetattr(fd.int(), &termios);
```
nektro: cast

src/bun.zig
```diff
@@ -2577,9 +2619,9 @@ pub const Stat = if (Environment.isWindows) windows.libuv.uv_stat_t else std.os.
 
 pub const posix = struct {
     // we use these on windows for crt/uv stuff, and std.os does not define them, hence the if
-    pub const STDIN_FD = if (Environment.isPosix) std.os.STDIN_FILENO else 0;
-    pub const STDOUT_FD = if (Environment.isPosix) std.os.STDOUT_FILENO else 1;
-    pub const STDERR_FD = if (Environment.isPosix) std.os.STDERR_FILENO else 2;
+    pub const STDIN_FD = toFD(if (Environment.isPosix) std.os.STDIN_FILENO else 0);
+    pub const STDOUT_FD = toFD(if (Environment.isPosix) std.os.STDOUT_FILENO else 1);
+    pub const STDERR_FD = toFD(if (Environment.isPosix) std.os.STDERR_FILENO else 2);
```
nektro: this if statement feels redundant. is this older code that can be changed to constants?

src/bun.js/api/bun/subprocess.zig
```diff
@@ -1848,6 +1848,7 @@ pub const Subprocess = struct {
 
         for (stdio_pipes.items) |*item| {
             const maybe = blk: {
+                // TODO: move this to bun.sys so it can return [2]bun.FileDesriptor
```
nektro, line 1852: will do in a followup pr

src/bun.js/bindings/bindings.zig
Jarred-Sumner: It's not lossy, but `isNumber` is both decimal and int32. `asInt32` is only int32. This would produce unexpected results when the value is a double. Instead, this should either use `coerce` or `toInt32`

src/bun.zig
```diff
@@ -838,6 +839,41 @@ else if (Environment.isWindows)
 else
     std.os.fd_t;
 
+pub const FileDescriptor = enum(FileDescriptorInt) {
```
paperclover, line 842: If we're going to make `FileDescriptor` not an integer type, I feel like it would be possible to use `FDImpl` directly for everything (rename that to FileDescriptor), instead of creating a new wrapper.

I had tried that initially but I did not feel like altering every usage of FileDescriptor to make this work.

Note at the time of writing FDImpl, I had ensured that ReleaseFast was able to optimize every line of code away in that struct on posix systems, as it was simply a different interpretation of `i32`.

src/bun.zig
```diff
@@ -1418,7 +1454,8 @@ pub const FDHashMapContext = struct {
         // a file descriptor is i32 on linux, u64 on windows
         // the goal here is to do zero work and widen the 32 bit type to 64
         // this should compile error if FileDescriptor somehow is larger than 64 bits.
-        return @as(std.meta.Int(.unsigned, @bitSizeOf(FileDescriptor)), @bitCast(fd));
+        if (@bitSizeOf(FileDescriptorInt) > 64) @compileError("no bueno");
```
paperclover: i had this suggestion initially:
```suggestion
        comptime std.debug.assert(@bitSizeOf(FileDescriptor) <= 64);
```


but i think the `@intCast` now does work of converting a i32 bits -> u64 bits, where previously the code would just pad 32 zero bits the bit representation of the thing.

i cant imagine the performance difference matters, but i was very intentional on how the cast was written before.

src/bun.zig
```diff
@@ -2463,13 +2500,18 @@ pub inline fn toFD(fd: anytype) FileDescriptor {
             FDImpl.UV => FDImpl.fromUV(fd),
             FileDescriptor => FDImpl.decode(fd),
             // TODO: remove u32
-            u32, i32 => FDImpl.fromUV(@as(FDImpl.UV, @intCast(fd))),
+            u32, i32, comptime_int, usize, i64 => FDImpl.fromUV(@as(FDImpl.UV, @intCast(fd))),
```
paperclover: these additional allowed types is likely a mistake (`@intCast` on file descriptor representation is always incorrect)

src/bun.zig
```diff
@@ -2463,13 +2500,18 @@ pub inline fn toFD(fd: anytype) FileDescriptor {
...
             else => @compileError("toFD() does not support type \"" ++ @typeName(T) ++ "\""),
         }).encode();
     } else {
         // TODO: remove intCast. we should not be casting u32 -> i32
         // even though file descriptors are always positive, linux/mac repesents them as signed integers
-        return @intCast(fd);
+        return switch (T) {
+            FileDescriptor => fd, // TODO: remove the toFD call from these places and make this a @compileError
```
paperclover, line 2521: in the windows path this was allowed due to some generic cast situations
i dont know if this todo comment will be achiveable

src/bun.zig
paperclover: i wonder if there is any target where STDIN_FILENO is not 0. but i am pretty sure these constants are not defined when compiling to windows.

src/bundler.zig
```diff
@@ -1027,7 +1027,7 @@ pub const Bundler = struct {
 
                     if (bundler.fs.fs.needToCloseFiles()) {
                         file.close();
-                        file_op.fd = 0;
+                        file_op.fd = .zero;
```
paperclover, line 1030: i havent looked over every usage of .zero, but i suspect most of them can become `invalid_fd`

src/fs.zig
```diff
@@ -83,14 +83,14 @@ pub const FileSystem = struct {
...
-    pub var max_fd: FileDescriptorType = 0;
+    pub var max_fd: FileDescriptorType = .zero;
 
     pub inline fn setMaxFd(fd: anytype) void {
         if (!FeatureFlags.store_file_descriptors) {
             return;
         }
 
-        max_fd = @max(fd, max_fd);
+        max_fd = @enumFromInt(@max(fd, max_fd.int()));
```
paperclover, line 93: treating an FD like a number (max fd) case will (and has not ever) work on windows, because on windows is it not a number; there is not a great notion of "the highest fd".

thanks for uncovering this. i dont think it needs to be addressed yet because we dont set `store_file_descriptors` on windows.

src/install/bin.zig
```diff
@@ -578,7 +579,7 @@ pub const Bin = extern struct {
                 dest_buf[0.."../".len].* = "../".*;
                 remain = dest_buf["../".len..];
             } else {
-                if (this.global_bin_dir.fd >= bun.invalid_fd) {
+                if (this.global_bin_dir.fd >= bun.invalid_fd.int()) {
```
paperclover, line 582: \>= with `invalid_fd` was and is wrong and can break on windows.

paperclover (commented):
this feels very similar to what i tried to do when writing FDImpl initially, i just didn't want to pass the entire codebase to use a container type for file descriptors. it is cool to see but i dont think this should be merged yet

src/bun.zig
nektro: enums are also ints in codegen and I can consolidate them later into one type in a followup but decided it would be better to hold off and not do it in this one

PR #8023, opened 2024-01-07
Fixes #5178

### LINKED ISSUE #5178: `--compile` TypeError: Argument must be a file URL on a new Sveltekit

1.0.1

Darwin 21.6.0 arm64 arm, Mac M1

**Steps to reproduce**

Same steps as: https://bun.sh/guides/ecosystem/sveltekit

```
npm create svelte@latest my-app
cd my-app
bun install
bun add svelte-adapter-bun

# update adapter to: import adapter from 'svelte-adapter-bun'
vi svelte.config.js

bun run build
bun build build/index.js --compile --outfile binary
./binary
```

**Error**

```
╰─$ ./binary
17771 |   const protocol = protocol_header && headers.get(protocol_header) || "http";
17772 |   const host = headers.get(host_header);
17773 |   return `${protocol}://${host}`;
17774 | };
17775 |
17776 | var __dirname2 = path.dirname(fileURLToPath(new URL(import.meta.url)));
                                 ^
TypeError: Argument must be a file URL
      at compiled://root/binary:17776:30
```

**Notes**

Every step of build passes as ok, `bun run dev` and `bun build/index.js` work fine.
Just that the compiled bun binary cannot start the Sveltekit app and throws the error above.

Should start server on `localhost:3000`

### What do you see instead?

_No response_

### Additional information

**Same issues** - previously closed

https://github.com/oven-sh/bun/issues/1069
https://github.com/oven-sh/bun/issues/1070

**Other similar issues**

https://github.com/oven-sh/bun/issues/4592

Jarred-Sumner:
### What does this PR do?

Fixes #5178

This changes the internally-generated path to files compiled with `bun build --compile` from `compiled://root/` to `/$bunfs/root`. This does effectively ban importing files in `/$bunfs/`, but that seems like a better tradeoff than expecting all software that use `file:` URLs to also check for `compiled:`. The `$` in `$bunfs` makes it less likely to conflict in practice  because it's not valid on Windows (iirc) and if you type that in a terminal, it will think its an environment variable. 

Tests

### DIFF
--- a/src/cli/build_command.zig
+++ b/src/cli/build_command.zig
@@ -109,8 +109,8 @@ pub const BuildCommand = struct {
 
             // We never want to hit the filesystem for these files
             // This "compiled" protocol is specially handled by the module resolver.
-            this_bundler.options.public_path = "compiled://root/";
-            this_bundler.resolver.opts.public_path = "compiled://root/";
+            this_bundler.options.public_path = "/$bunfs/root/";
+            this_bundler.resolver.opts.public_path = "/$bunfs/root/";
 
             if (outfile.len == 0) {
                 outfile = std.fs.path.basename(this_bundler.options.entry_points[0]);
--- a/src/cli/build_command.zig
+++ b/src/cli/build_command.zig
@@ -108,7 +108,12 @@ pub const BuildCommand = struct {
             }
 
             // We never want to hit the filesystem for these files
-            // This "compiled" protocol is specially handled by the module resolver.
+            // We use the `/$bunfs/` prefix to indicate that it's a virtual path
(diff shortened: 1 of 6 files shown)


src/cli/build_command.zig
```diff
@@ -109,8 +109,8 @@ pub const BuildCommand = struct {
 
             // We never want to hit the filesystem for these files
             // This "compiled" protocol is specially handled by the module resolver.
```
paperclover: the comment here is out of date

paperclover (approved):
you should touch up the comment but looks good otherwise

PR #8025, opened 2024-01-07
Fixes #8010

### LINKED ISSUE #8010: Bun install is failing in vercel (Turbo Repo)

1.0.21

vercel

Deploy any project in vercel using bun install (bun.lockb included)

The installation should be finished

### What do you see instead?

Error: Cannot find module '/vercel/path0/install.js'

image

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #8010

This makes it so we always regenerate the lockfile whenever the meta hash changes. Previously, we would skip this work in certain cases, but that makes bun install less deterministic when we upgrade versions. 

In our site repo (~680 packages to install), it doesn't seem to have a meaningful performance impact. Maybe +1ms on an M1. 

I think this would have a slight perf impact in very large workspaces, but probably not more than a few milliseconds at most, which we can afford right now.

 

Let's see if CI gets mad

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -7688,8 +7688,21 @@ pub const PackageManager = struct {
             }
 
             const trees = this.lockfile.buffers.trees.items;
-            this.tree_install_counts[tree_id] += 1;
-            if (this.tree_install_counts[tree_id] >= trees[tree_id].dependencies.len) {
+            const current_count = this.tree_install_counts[tree_id];
+            const max = trees[tree_id].dependencies.len;
+
+            if (current_count == std.math.maxInt(usize)) {
+                if (comptime Environment.allow_assert)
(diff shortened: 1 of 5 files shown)


src/install/install.zig
```diff
@@ -7688,8 +7688,21 @@ pub const PackageManager = struct {
             }
 
             const trees = this.lockfile.buffers.trees.items;
-            this.tree_install_counts[tree_id] += 1;
-            if (this.tree_install_counts[tree_id] >= trees[tree_id].dependencies.len) {
+            const current_count = this.tree_install_counts[tree_id];
+            const max = trees[tree_id].dependencies.len;
+
+            if (current_count == std.math.maxInt(usize)) {
```
Jarred-Sumner, line 7694: @dylan-conway the previous assertion for this was failing in the `test/` folder in bun's repo when `rm -rf ~/.bun/install/cache`

I suspect the count included workspace scripts which may have not completed yet

PR #8028, opened 2024-01-07
5% - 10% faster `new Headers(object)` & `new URLSearchParams(object)`

Jarred-Sumner:
### What does this PR do?

This makes JDOMConvertRecord faster by using `structure()->forEachProperty`. `structure()->forEachProperty` can only be used when it's a regular object that cannot observe getters or has property.

This makes `new Headers` & `new URLSearchParams` faster when the input is an object. It does nothing for arrays and very little for empty objects.

image

Existing tests + added one which was probably unnecessary

### DIFF
--- a/src/bun.js/bindings/webcore/JSDOMConvertRecord.h
+++ b/src/bun.js/bindings/webcore/JSDOMConvertRecord.h
@@ -98,6 +98,71 @@ private:
         ReturnType result;
         HashMap<KeyType, size_t> resultMap;
 
+        bool canUseFastPath = false;
+        JSC::Structure* structure = nullptr;
+
+        switch (object->type()) {
+        case ObjectType:
+        case FinalObjectType: {
+            structure = object->structure();
+            canUseFastPath = structure->canPerformFastPropertyEnumeration() && !structure->hasNonReifiedStaticProperties();
+
+            if (canUseFastPath) {
+                JSValue prototype = object->getPrototypeDirect();
+
+                if (!(prototype == lexicalGlobalObject.objectPrototype() || prototype.structureOrNull() == lexicalGlobalObject.nullPrototypeObjectStructure())) {
+                    canUseFastPath = false;
+                }
+            }
+            break;
+        }
+        default: {
+            break;
+        }
+        }
+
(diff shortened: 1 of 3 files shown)


src/bun.js/bindings/webcore/JSDOMConvertRecord.h
```diff
@@ -98,6 +98,71 @@ template<typename K, typename V> struct Converter<IDLRecord<K, V>> : DefaultConv
...
+        switch (object->type()) {
+        case ObjectType:
+        case FinalObjectType: {
+            structure = object->structure();
+            canUseFastPath = structure->canPerformFastPropertyEnumeration() && !structure->hasNonReifiedStaticProperties();
+
+            if (canUseFastPath) {
+                JSValue prototype = object->getPrototypeDirect();
+
+                if (!(prototype == lexicalGlobalObject.objectPrototype() || prototype.structureOrNull() == lexicalGlobalObject.nullPrototypeObjectStructure())) {
```
Jarred-Sumner, line 113: I feel like there's a better way to do this, but I'm not sure what it is

TiBianMod:
can you please bench this @Jarred-Sumner 

```ts
// setup
class Something {
    #headers = new Headers({
        "Content-Type": "text/plain; charset=utf-8",
        "Header-X": "Value of header",
    });

    reset() {
        this.#headers = new Headers();
    }
}

const something = new Something();
```

```ts
// bench the reset method
something.reset();
```

> Node 21.5.0
CleanShot 2024-01-07 at 23 55 28

> Bun 1.0.21
CleanShot 2024-01-07 at 23 57 53

paperclover (approved):

paperclover:
i verify this is an issue but i suspect this patch will not improve or worsen the situation
```ts
import { bench, run } from "mitata";

class Something {
  headers = new Headers({
    "Content-Type": "text/plain; charset=utf-8",
    "Header-X": "Value of header",
  });

  reset() {
    this.headers = new Headers();
  }
}

const something = new Something();

bench("reset", () => {
  something.reset();
});

bench("new Headers", () => {
  return new Headers();
});

await run();
```

image

PR #8037, opened 2024-01-08
Tweaks to lifecycle script enqueue

Jarred-Sumner:
### What does this PR do?

This hopefully helps with #8010, but I'm doubtful it fixes the issue

This makes it so that the current working directory passed to `readDirInfo` has a trailing slash, which is necessary. 

This also:
- Removes an extra call to `try bun.getFdPath`
- Adds more logging for lifecycle scripts in debug builds
- Gives us a way to force the usage of the /proc/ workaround in debug builds
- Closes the json file descriptor sooner

Manually

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -1836,7 +1836,7 @@ var needs_proc_self_workaround: bool = false;
 // necessary on linux because other platforms don't have an optional
 // /proc/self/fd
 fn getFdPathViaCWD(fd: std.os.fd_t, buf: *[@This().MAX_PATH_BYTES]u8) ![]u8 {
-    const prev_fd = try std.os.openatZ(std.fs.cwd().fd, ".", 0, 0);
+    const prev_fd = try std.os.openatZ(std.fs.cwd().fd, ".", std.os.O.DIRECTORY, 0);
     var needs_chdir = false;
     defer {
         if (needs_chdir) std.os.fchdir(prev_fd) catch unreachable;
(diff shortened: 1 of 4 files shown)


src/bun.zig
```diff
@@ -1836,7 +1836,7 @@ var needs_proc_self_workaround: bool = false;
 // necessary on linux because other platforms don't have an optional
 // /proc/self/fd
 fn getFdPathViaCWD(fd: std.os.fd_t, buf: *[@This().MAX_PATH_BYTES]u8) ![]u8 {
-    const prev_fd = try std.os.openatZ(std.fs.cwd().fd, ".", 0, 0);
+    const prev_fd = try std.os.openatZ(std.fs.cwd().fd, ".", std.os.O.DIRECTORY, 0);
```
dylan-conway, line 1839: This will fail if the fd is for a file, but it would fail before this change too when fchdir is called

Jarred-Sumner: `std.fs.cwd().fd` is never a file

but the fchdir call below - yeah. that fails

src/install/install.zig
```diff
@@ -8554,15 +8556,19 @@ pub const PackageManager = struct {
         // no need to download packages you've already installed!!
         var skip_verify_installed_version_number = false;
         const cwd = std.fs.cwd();
-        const node_modules_folder = cwd.openDir("node_modules", .{}) catch brk: {
+        const node_modules_folder = cwd.openDir("node_modules", .{
```
dylan-conway: maybe we should change this and the others to `bun.openDir`

dylan-conway (approved):

PR #8038, opened 2024-01-08
fix: #7932

### LINKED ISSUE #7932: `peerDependencies` not resolving properly

1.0.20+09d51486e

Install a a package that requires a peerDependency to work, also install said peerDependency.
Afterwards, try to update the peerDependency.

If you are using typescript you will find lots of type collisions because of missmatched versions, if you observe further you will see that bun has installed 2 versions of the dependency instead of just having the one you updated

I expect bun to just have 1 version of the dependency

### What do you see instead?

Bun installs multiple versions

### Additional information

_No response_

eriklangille:
### What does this PR do?

Fixes #7932 

When updating a package that is used as a peerDependency in one of the installed packages, then old package would be kept as a module within the child package's node_modules. With this fix, the child package uses the root package, as expected.

- [x] Code changes

I wrote automated tests

- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test/cli/install/registry/bun-install-registry.test.ts`)


dylan-conway:
This is great. This also fixes a todo test in `bun-install-registry.test.ts`. If you search for the test `it should install in such a way that two identical packages with different peer dependencies are different instances`, remove the `todo`, and change `expect(out).toBe("true\ntrue\ntrue\n");` to `expect(out).toBe("true\ntrue\nfalse\n");`, it should pass. Testing manually with yarn and npm shows that the last boolean in the string should actually be false.

dylan-conway (approved):

eriklangille:
> This is great. This also fixes a todo test in `bun-install-registry.test.ts`. If you search for the test `it should install in such a way that two identical packages with different peer dependencies are different instances`, remove the `todo`, and change `expect(out).toBe("true\ntrue\ntrue\n");` to `expect(out).toBe("true\ntrue\nfalse\n");`, it should pass. Testing manually with yarn and npm shows that the last boolean in the string should actually be false.

Done.

dylan-conway:
Thank you

PR #8057, opened 2024-01-09
fix(bundler): improve external `require` transpilation on all targets

### LINKED ISSUE #6168: `bun build --target node` injects incorrect `require` function

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

Darwin 21.6.0 x86_64 i386

```
edison@EMC3 localtunnel % bun build app.ts --outdir . --target node 

  ./app.js  92.64 KB

[57ms] bundle 43 modules
edison@EMC3 localtunnel % node -v
v20.8.0
edison@EMC3 localtunnel % node app.js                               
file:///Users/edison/Documents/webserver/git/localtunnel/app.js:19
  return import.meta.require(id);
                     ^

TypeError: (intermediate value).require is not a function
    at __require (file:///Users/edison/Documents/webserver/git/localtunnel/app.js:19:22)
    at file:///Users/edison/Documents/webserver/git/localtunnel/app.js:2622:19
    at file:///Users/edison/Documents/webserver/git/localtunnel/app.js:17:45
    at file:///Users/edison/Documents/webserver/git/localtunnel/app.js:2744:16
    at file:///Users/edison/Documents/webserver/git/localtunnel/app.js:17:45
    at file:///Users/edison/Documents/webserver/git/localtunnel/app.js:2758:34
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)

Node.js v20.8.0
```

```
edison@EMC3 localtunnel % bun app.js
Ingresa el puerto de tu servidor localhost: 
```

### What do you see instead?

```
file:///Users/edison/Documents/webserver/git/localtunnel/app.js:19
  return import.meta.require(id);
                     ^

TypeError: (intermediate value).require is not a function
```

### Additional information

```javascript
var __create = Object.create;
var __defProp = Object.defineProperty;
var __getProtoOf = Object.getPrototypeOf;
var __getOwnPropNames = Object.getOwnPropertyNames;
var __hasOwnProp = Object.prototype.hasOwnProperty;
var __toESM = (mod, isNodeMode, target) => {
  target = mod != null ? __create(__getProtoOf(mod)) : {};
  const to = isNodeMode || !mod || !mod.__esModule ? __defProp(target, "default", { value: mod, enumerable: true }) : target;
  for (let key of __getOwnPropNames(mod))
    if (!__hasOwnProp.call(to, key))
      __defProp(to, key, {
        get: () => mod[key],
        enumerable: true
      });
  return to;
};
var __require = (id) => {
  return import.meta.require(id);
};
```

### LINKED ISSUE #7326: bun build issue on require

1.0.14+d8be3e51b

MacOs

running on my application:
bun build --target=bun ./src/app/index.ts --outfile=./dist/index.js

Screen Shot 2023-11-26 at 12 23 00 PM
Screen Shot 2023-11-26 at 12 24 01 PM

on this line:
  var requireFunction = typeof __webpack_require__ === "function" || typeof __non_webpack_require__ === "function" ? __non_webpack_require__ : require;
  
need to be:
  var requireFunction = typeof __webpack_require__ === "function" || typeof __non_webpack_require__ === "function" ? __non_webpack_require__ : import.meta.require;
  
  obs: import.meta.require to works, application tuning after change the line:
Screen Shot 2023-11-26 at 12 27 18 PM

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #7583: `bun build --target browser` includes non-browser code `return import.meta.require(id);`

1.0.16+800fb1290

Darwin 23.1.0 arm64 arm

I have made a minimal repo here: https://github.com/iamacup/bun-require-repo

It looks like whatever is triggering this function to get 'bundled' is in a package called @slate-serializers/html, but i can't narrow it down more than that.

Checkout the repo or bundle this code:

```
import { slateToHtml } from "@slate-serializers/html";

slateToHtml([])

console.log("Hello via Bun!");
```

with `bun build ./index.ts --outdir ./dist --target browser`

Output bundled code should run in browser

### What do you see instead?

A __require function (which is never called or used) is generated which includes `imports.` which can not be parsed and results in a JS error: `Uncaught SyntaxError: import.meta may only appear in a module`
```
var __require = (id) => {
  return import.meta.require(id);
};

```

### Additional information

You can workaround this issue by, after the build step (or on file change if combined with `watch`) commenting out the line `return import.meta.require(id);` as it is a no-op anyway, never called.

there is a related issue but not targeted at browser here: https://github.com/oven-sh/bun/issues/6168

paperclover:
### What does this PR do?

This fixes the old import.meta.require bug, and also addresses every incorrect bundling situation related to detecting and emitting external requires. this also fixes some unused variables that used to get emitted.

the big impact of this is going to fix almost everyone's usage of `--target=node` including `import.meta.require`:

in:
```
console.log(require(String('react')));
```

out:
```diff
+import {createRequire} from "node:module";
 var __commonJS = (cb, mod) => () => (mod || cb((mod = { exports: {} }).exports, mod), mod.exports);
+var __require = createRequire(import.meta.url);
-var __require = id => import.meta.require(id);

 // index.cjs
 var require_scratchpad = __commonJS(() => {
-  console.log(require(String("react")));
+  console.log(__require(String("react")));
 });
 export default require_scratchpad();
```

**specific changelog**
- feat(bundler): implement `__require` for `--target=node` and `--target=browser`
- fix(bundler): properly emit the require ref for `e_require_call_target`, `e_require_resolve_call_target`, etc
- fix(bundler): do not include unused __toESM shim on a known dynamic import that is marked external
- fix(bundler): rewrite browser polyfill for `node:tty`. it is 1/4th the size
- internal: rename dynamic_require -> use_import_meta_require
- fix(bundler): `let y = require` properly attaches to the require ref
- fix(bundler): transposed require resolve is now set as import type `require_resolve`, fixing dce of `__require` in `require.resolve('h')` with `--target=bun`
- fix(bundler): transposed require resolve with ternary of non-strings works now
- fix(bundler): only call `generateRuntimeSymbolImportAndUse` when it actually is going to create the symbol (ignore when count=0)
 
Fixes #7326
Fixes #6168
Fixes #7583

- [x] Fix renaming `require` to `__require` (existing bug)
- [x] regressions
- [X] lots of changes to how require is handled, fixing many subtle and existing bugs
- [ ] New Tests

alot of things fixed reduce code output (the __require helper is not used for target=bun):
image

i suspect that this code will be a hair faster at visiting call expressions / etc


paperclover:
image
new regressions because it printing `__require` (and not the definition) instead of `import.meta.require`

src/js_printer.zig
```diff
@@ -2029,14 +2029,12 @@ fn NewPrinter(
...
-                    } else {
-                        p.print("require");
-                    }
+                if (p.options.require_ref) |ref| {
+                    p.printSymbol(ref);
+                } else if (p.options.module_type == .esm and is_bun_platform) {
+                    p.print("import.meta.require");
                 } else {
-                    p.printSymbol(p.options.require_ref.?);
+                    p.print("require");
```
Jarred-Sumner, line 2036: this is why it broke btw

paperclover:
i got side tracked and fixed a bug with node:tty while bundling:
it treeshakes about a billion times nicer
image
image

paperclover:
note: the rationale for removing the internal `rewrite_require_resolve` flag is because it always leads to inaccurate codegen. esbuild does no such thing either

the intent of a bundled require.resolve is always to query the exact resolve path.
this is what esbuild does and they even have a warning if you dont mark such paths as external:

image

bundling for `require.resolve('./hello.js')`:

- in runtime: absolute path
- in bundle before: inline string `"./hello.js"`
- in bundle now: dynamic call to `import.meta.resolveSync("./hello.js")` or `__require.resolve`

src/bundler/bundle_v2.zig
```diff
@@ -2309,27 +2310,72 @@ pub const ParseTask = struct {
...
+        const code = @embedFile("../runtime.js") ++ switch (target) {
+            .bun, .bun_macro =>
+            \\export var __require = /* @__PURE__ */ import.meta.require;
+            \\
+            ,
+            .node =>
+            \\import { createRequire } from "node:module";
+            \\export var __require = /* @__PURE__ */ createRequire(import.meta.url);
+            ,
+            // Copied from esbuild's runtime.go:
```
jdalton, line 2337: Do we reference ESBuilds MIT license in say a thirdparty directory? It's MIT we should include it somewhere.

src/js_parser.zig
```diff
@@ -1554,7 +1554,7 @@ const StaticSymbolName = struct {
 
         pub const __name = NewStaticSymbol("__name");
         pub const __toModule = NewStaticSymbol("__toModule");
-        pub const __require = NewStaticSymbol("require");
+        pub const __require = NewStaticSymbol("__require");
```
jdalton: ☝️ Do these symbols get unique'd if say a user code has an identifier named `__require`?

src/runtime.zig
```diff
@@ -313,7 +313,7 @@ pub const Runtime = struct {
 
         /// Use `import.meta.require()` instead of require()?
         /// This is only supported in Bun.
-        dynamic_require: bool = false,
+        use_import_meta_require: bool = false,
```
dylan-conway, line 316: I like this rename

src/bundler/bundle_v2.zig
paperclover: there is a very small note here. https://bun.sh/docs/project/licensing

the entire file is reimplemented off of esbuild, perhaps this could be noted.

src/js_parser.zig
paperclover: yes

src/bundler/bundle_v2.zig
jdalton: Okay that's a start. I think having a folder to put these licenses (copies of the derivative works licenses will be good too, basically a bunch of MIT license files from the projects we derive from).

src/js_parser.zig
```diff
@@ -4936,13 +4933,41 @@ fn NewParser_(
...
+            std.debug.assert(arg.data == .e_string);
+
+            // Ignore calls to import() if the control flow is provably dead here.
+            // We don't want to spend time scanning the required files if they will
+            // never be used.
+            if (p.is_control_flow_dead) {
+                return p.newExpr(E.Null{}, arg.loc);
+            }
+
+            const import_record_index = p.addImportRecord(.require_resolve, arg.loc, arg.data.e_string.string(p.allocator) catch unreachable);
```
dylan-conway, line 4955: was this always supposed to be `require_resolve`?

src/js_parser.zig
```diff
@@ -5066,8 +5096,6 @@ fn NewParser_(
                     );
                 },
             }
-
-            return arg;
```
dylan-conway, line 5070: nice

src/js_parser.zig
```diff
@@ -4982,7 +5007,6 @@ fn NewParser_(
                             .import_record_id = import_record_index,
                         }) catch unreachable;
                         p.import_items_for_namespace.put(p.allocator, namespace_ref, ImportItemForNamespaceMap.init(p.allocator)) catch unreachable;
-                        p.ignoreUsage(p.require_ref);
```
dylan-conway, line 4985: why is this and the other ignoreUsage not necessary anymore?

dylan-conway (commented):
Had a couple questions. Overall looks good, want to see test results from the most recent commit

src/js_parser.zig
paperclover: i changed alot of how require_ref/__require was tracked.

i'm almost certain that the usage count of the require_ref `the unbound require` global, doesn't actually matter, but the usage count of the runtime.js `__require` symbol does. that's what `p.recordUsageOfRuntimeRequire();` is doing, and it is only called when we are 100% certain the ast node is a runtime require call.

in this path we are returning either namespace_ref (not a runtime require), or e_require_string, which *may* be a runtime require. and in either case an import record has been created.

at the end, we check for these and mark them as needed (see `var runtime_require_uses`)

i want to add tests to bundle each possible way to call require separately to ensure nothing is mistaken. i spent an absurd amount of time dealing with DCE of the __require helper.

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -340,6 +340,8 @@ extern "C" JSC::EncodedJSValue functionImportMeta__resolveSyncPrivate(JSC::JSGlo
                 auto bunStr = Bun::toString(parentIdStr);
                 args.append(jsBoolean(Bun__isBunMain(lexicalGlobalObject, &bunStr)));
 
+                auto thisValue = callFrame.thisValue();
```
Jarred-Sumner: y tho

paperclover: mistakenly included, it is already removed.

src/resolver/resolver.zig
```diff
@@ -833,6 +823,25 @@ pub const Resolver = struct {
...
+                bun.JSC.HardcodedModule.Aliases.has(import_path, r.opts.target))
+            {
+                return .{
+                    .success = Result{
+                        .import_kind = kind,
+                        .path_pair = PathPair{
+                            .primary = Path.init(import_path),
+                        },
+                        .is_external = true,
+                        .module_type = .cjs,
```
Jarred-Sumner, line 838: This should be either `.esm` or `.cjs` depending on what the input kind was

```suggestion
                        .module_type = switch (kind) {
                            .require, .require_resolve => .cjs,
                            else => .esm
                        },
```

paperclover:
failing tests

Creative-Difficulty:
🎉

mathiasrw:
Nice work everyone.

Issue #8032, opened 2024-01-07
`bun remove org/repo` (github) throws an error


Jarred-Sumner:
bun:

```js
❯ bun remove mishoo/UglifyJS
bun remove v1.0.21 (703e8bd2)

error: "" is not in a package.json file
```

npm:
```js
❯ npm remove mishoo/UglifyJS

changed 1 package, and audited 2 packages in 2s

found 0 vulnerabilities
```

PR #8091, opened 2024-01-11
fix(bun-install): gracefully handle missing package removal

kaioduarte:
### What does this PR do?

Handles gracefully the removal of missing packages from the package.json, this not only applies to `org/repo` packages but to any missing package.

This matches the NPM behavior, which doesn't error out in such cases.

- [x] Code changes

Manually tested too.
| Bun | NPM |
| --- | --- |
| image | image |

Closes https://github.com/oven-sh/bun/issues/8032

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -7482,11 +7482,6 @@ pub const PackageManager = struct {
                     }
                 }
 
-                if (!any_changes) {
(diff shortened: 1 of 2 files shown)


paperclover:
could you write a simple unit test for this?

kaioduarte:
@paperdave done, I missed the existing tests.

paperclover (approved):

Jarred-Sumner:
Thank you

PR #8108, opened 2024-01-11
Fix loose comparisons in `Bun.semver.order`, also fix #8040

### LINKED ISSUE #8040: `bun install` not respecting version ranges

1.0.21+437faa944

Linux 6.2.0-1018-azure x86_64 unknown

1. Add resolution to package.json: 
```json
"resolutions": {
  "vue": ">=3.3.0-beta.1 <3.4.0-beta.3"
}
```
2. run `bun add vue`

Expected to see at highest vue@3.4.0-beta.3.

### What do you see instead?

Vue latest is installed, `3.4.5`

### Additional information

I'm running on Bun canary.

Electroid:
### What does this PR do?

Previously, `Bun.semver.order("4.x", "4.0")` would return `0`, now it returns `1`. This was because we were rounding down to 0 when comparing versions.

This change only affects the `semver.order` API, and does not touch `bun install` version ordering.

---

Fixes: #8040

We were double-counting `-` when calculating the length of a semver. This would cause a compound semver to be incorrectly parsed after a tag was present.

This will affect version order, as this was a logical error.

```js
import { semver } from "bun";

test("semver with multiple tags work properly", () => {
  expect(semver.satisfies("3.4.5", ">=3.3.0-beta.1 <3.4.0-beta.3")).toBeFalse();
});
```

Before: true (incorrect)
After: false (correct)

- [x] Code changes

Automated tests.


test/cli/install/semver.test.ts
```diff
@@ -105,6 +105,24 @@ describe("Bun.semver.order()", () => {
...
+  // not supported by semver, but supported by Bun
+  test.each([
+    ["0", "0.0"],
+    ["1", "1.0"],
+    ["1.2", "1.2.0"],
+    ["1.x", "1.0.x"],
+    ["1.x.x", "1.0.x"],
+    ["2.x", "1.x"],
+    ["2.x", "2.1"],
+    ["2", "1"],
```
dylan-conway, line 117: this also applies to wildcards like `*`, `2.*`, `2.1.*`? Could you add a couple tests for those as well?

Jarred-Sumner:
linux build is failing

test/cli/install/semver.test.ts
```diff
@@ -683,6 +704,10 @@ describe("Bun.semver.satisfies()", () => {
       [">=1.0.0 <1.1.0-pre", "1.1.0-pre"],
 
       ["== 1.0.0 || foo", "2.0.0"],
+
+      // https://github.com/oven-sh/bun/issues/8040
+      [">=3.3.0-beta.1 <3.4.0-beta.3", "3.4.5"],
+      ["^3.3.0-beta.1", "3.4.5"],
```
dylan-conway: the last test in this list should be moved to the `range includes` test

dylan-conway (approved):
Looks good. Ready to merge when tests are fixed

PR #8110, opened 2024-01-11
Fix error when null/undefined is passed to stdio in spawn/spawnSync

### LINKED ISSUE #8095: bun 1.0.22 breaks child_process spawn() / spawnSync() - Invalid stdio option "null"

1.0.22

Linux 5.15.0-67-generic x86_64 x86_64 - ubuntu 22.04

-

-

### What do you see instead?

```js
@yusham:00-cup/bun-play> cat ./subprocess.js 
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: ./subprocess.js
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ import cp from 'child_process';
   2   │ 
   3   │ 
   4   │ cp.spawnSync('ls', ['-lagh'], {stdio: [null, 'inherit', null]})
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@yusham:00-cup/bun-play> bun ./subprocess.js 
1 | import cp from 'child_process';
2 | 
3 | 
4 | cp.spawnSync('ls', ['-lagh'], {stdio: [null, 'inherit', null]})
                        ^
error: Invalid stdio option "null"
      at node:child_process:357:47
      at map (:1:21)
      at node:child_process:173:47
      at /data/www/00-cup/bun-play/subprocess.js:4:1
@yusham:00-cup/bun-play> 
```

### Additional information

_No response_

Electroid:
### What does this PR do?

Fixes #8095 

- Fixed `spawn/Sync({ stdio: [null/undefined] })` being an error
- Fixed `spawn/Sync().stdout/error` being undefined instead of null (match Node.js)
- Fixed `spawn().stdio` incorrectly being an object, instead of an array

- [x] Code changes

Added regression tests.

### DIFF
--- a/src/js/node/child_process.js
+++ b/src/js/node/child_process.js
@@ -1334,7 +1334,7 @@ const nodeToBunLookup = {
 function nodeToBun(item, index) {
   // If not defined, use the default.
   // For stdin/stdout/stderr, it's pipe. For others, it's ignore.
-  if (item === undefined || item === null) {
+  if (item == null) {
     return index > 3 ? "ignore" : "pipe";
   }
   // If inherit and we are referencing stdin/stdout/stderr index,
--- a/src/js/node/child_process.js
+++ b/src/js/node/child_process.js
@@ -571,18 +571,22 @@ function spawnSync(file, args, options) {
     }
   }
 
-  const { stdout, stderr, success, exitCode } = Bun.spawnSync({
+  const {
+    stdout = null,
+    stderr = null,
+    success,
+    exitCode,
+  } = Bun.spawnSync({
     cmd: options.args,
     env: options.env || undefined,
     cwd: options.cwd || undefined,
-    stdin: bunStdio[0],
-    stdout: bunStdio[1],
-    stderr: bunStdio[2],
+    stdio: bunStdio,
   });
 
   const result = {
     signal: null,
     status: exitCode,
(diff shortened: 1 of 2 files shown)


src/js/node/child_process.js
```diff
@@ -1327,16 +1331,22 @@ const nodeToBunLookup = {
   ipc: "ipc",
 };
 
-function nodeToBun(item) {
+function nodeToBun(item, index) {
+  // If not defined, use the default.
+  // For stdin/stdout/stderr, it's pipe. For others, it's ignore.
+  if (item === undefined || item === null) {
```
Jarred-Sumner:
```suggestion
  if (item == null) {
```

Jarred-Sumner:
I think we could also allow `null` or `undefined` in the Bun.spawn implementation itself fwiw

Electroid:
> I think we could also allow `null` or `undefined` in the Bun.spawn implementation itself fwiw

I noticed there was a bug with this, but didn't get to it. Looks like `null` to `Bun.spawn` is not ignoring, but using the default instead.

PR #8114, opened 2024-01-12
windows: support `bun:ffi`

paperclover:
- [x] Compile TCC for win32
- [x] Get FFI working

iw ill do these in a followup
- [ ] ~~Fix a Bun.write issue preventing tests from being run~~
- [ ] ~~Get the test file passing~~


paperclover:
This attempts to but doesnt fix Bun.write

src/bun.js/webcore/blob.zig
```diff
@@ -1192,18 +1192,23 @@ pub const Blob = struct {
             var file_path: [bun.MAX_PATH_BYTES]u8 = undefined;
             switch (bun.sys.open(
                 pathlike.path.sliceZ(&file_path),
-                // we deliberately don't use O_TRUNC here
-                // it's a perf optimization
-                std.os.O.WRONLY | std.os.O.CREAT | std.os.O.NONBLOCK,
+                if (!Environment.isWindows)
+                    // we deliberately don't use O_TRUNC here
+                    // it's a perf optimization
+                    std.os.O.WRONLY | std.os.O.CREAT | std.os.O.NONBLOCK
```
Jarred-Sumner, line 1198: i guess because if it was not a file it would be different?

.vscode/settings.json
```diff
@@ -29,7 +29,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "zig.zls.enableInlayHints": false,
-  "zig.path": "${workspaceFolder}/.cache/zig/zig",
+  "zig.path": "${workspaceFolder}/.cache/zig/zig.exe",
```
Jarred-Sumner: are we going to make it alias zig.exe to zig on posix

PR #8115, opened 2024-01-12
Added Dependabot to automatically upgrade dependencies.

amirfarzamnia:
### What does this PR do?

This pull request will introduce a new file named .github/dependabot.yml. This file will integrate a bot into the GitHub repository, enabling automatic upgrades for the utilized packages.

- [x] Code changes

The file has been verified using GitHub.


xhyrom:
I don't think this will work. Dependabot doesn't support bun https://github.com/dependabot/dependabot-core/issues/6528

ghiscoding:
on the other hand, one would have better chance with Renovate https://docs.renovatebot.com/modules/manager/bun/

paperclover:
thanks for the contribution, but i do not think this bot should be added. this doesn't cover all of the dependencies. it just gets the root of package.json, which is honestly the dependencies that dont really matter what version is being used.

the submodule dependencies are what matter more, but it almost always makes more sense to do these upgrades manually.

PR #8118, opened 2024-01-12
Some Windows tweaks

Jarred-Sumner:
### What does this PR do?

- Only normalize libuv error codes on errors from libuv
- Adds a misctools/ script for different ways of reading the contents of a directory, to see what the best-performing one is
- Fixes a performance issue where the module resolution cache was not in use, leading to roughly 2x the system calls to start JavaScript
- Fixes an unnecessary `@memcpy` in the directory iterator and increasing the struct size without needing to (cc @dylan-conway)
- Switches to using bun's DirIterator in more places, so that we use the FileDirectoryInformation class info instead of FileBothInformation which is slightly faster
- Fixes reading tsconfig.json `"extends"` files using stdin as a file descriptor (cc @nektro) 
- Switches to using bun.sys when reading files instead of zig's std.os. This is likely to break something, lets see what tests fail

Let's see which tests fail


paperclover:
awesome

packages/bun-usockets/src/loop.c
```diff
@@ -342,7 +342,7 @@ void us_internal_dispatch_ready_poll(struct us_poll_t *p, int error, int events)
 
                     if (length > 0) {
                         s = s->context->on_data(s, loop->data.recv_buf + LIBUS_RECV_BUFFER_PADDING, length);
-                        
+                        #ifndef WIN32
```
paperclover, line 346: why is this disabled on windows? is it because libuv does callback reads? please add a comment so we don't forget.

paperclover (approved):

PR #8119, opened 2024-01-12
fix(docs): prisma not installed as dev dependency

aayushbtw:
Fixed simple docs error.
Added flag `-d` to `bun add prisma` (now `bun add -d prisma`)


Jarred-Sumner:
Thank you

Issue #4555, opened 2023-09-07
Unintuitive error on invalid tls configuration


Airtune:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

```js
Bun.serve({
  fetch(req, server) {
    return undefined;
  },
  tls: {
    key: Bun.file('./key.pm'),
    cert: Bun.file('./cert.pem')
  }
});
```

### What is the expected behavior?

I expected either an error for the BunFile not existing or for the tls configuration being wrong since I had a typo: `key.pm` instead of the correct `key.pem`

### What do you see instead?
```js
1 | Bun.serve({
  fetch(req, server) {
    return;
  },
  tls: {
    key: Bun.file("./key.pm"),
    cert: Bun.file("./cert.pem")
  }
});
   ^
TypeError: Expected fetch() to be a function
 code: "ERR_INVALID_ARG_TYPE"

index.ts:1:0
```

Airtune:
I upgraded to bun v1.0.0 and I'm still seeing the same error

josiaho:
Same result on 1.0.3

tonyea:
Dang. Got the same issue as I was about to launch a production microservice. Going to have to route HTTPS via a containerized nginx for now.

josiaho:
@tonyea Not sure if it helps in the meantime, but I have letsencrypt/certbot SSL working like this:

    const fs = require('fs');
    const https = require('https');
    const express = require('express');
    
    const app = express();
    
    app.get('/', (req, res) => {
      return res.send('Hello, world!');
    });
    
    https.createServer( {
      cert: fs.readFileSync('server.crt'),
      key: fs.readFileSync('server.key')
    }, app).listen(443);

tonyea:
> const https = require('https');

Good point. Could just use express in the meantime without switching to a reverse proxy or abandoning bun

tonyea:
Actually ended up using an nginx container and reverse proxied tls

mollthecoder:
Having this issue on v1.0.14

ghost:
I think found a solution!
You have to put the tls bit before the fetch bit, I had the same issue and this worked for me!
Your code would look like this:
```
Bun.serve({
  tls: {
    key: Bun.file('./key.pm'),
    cert: Bun.file('./cert.pem')
  },
  fetch(req, server) {
    return undefined;
  }
});
```
[...]

PR #8120, opened 2024-01-12
Changed documentation to help users avoid bug

ghost:
### What does this PR do?

This pull request changes the documentation for the tls part of the Bun.serve function.
This is because of an error that comes up when the fetch() is placed *before* the tls object:
```
TypeError: Expected fetch() to be a function
 code: "ERR_INVALID_ARG_TYPE"
```
I changed the documentation to show users to put the tls part before the fetch() so that this error would not show up

- [ ] Code changes

I used the preview function for the .md file

### DIFF
--- a/docs/api/http.md
+++ b/docs/api/http.md
@@ -110,14 +110,14 @@ Bun supports TLS out of the box, powered by [BoringSSL](https://boringssl.google
 
 ```ts-diff
   Bun.serve({
-    fetch(req) {
-      return new Response("Hello!!!");
-    },
-
 +   tls: {
 +     key: Bun.file("./key.pem"),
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
This error should definitely not happen. Are you sure it is happening?

ghost:
> This error should definitely not happen. Are you sure it is happening?

Yes it is happening, it is raised here: https://github.com/oven-sh/bun/issues/4555

paperclover:
maybe we could add this line _**temporarily**_. it should be noted as a workaround to a known bug.

really we should find out why this bug occurs.

paperclover:
this bug was fixed

Issue #8098, opened 2024-01-11
Difference in `url.parse` between Node and Bun


silvio2402:
### What version of Bun is running?

1.0.22+b400b36ca

### What platform is your computer?

Linux 5.15.90.4-microsoft-standard-WSL2 x86_64 unknown

### What steps can reproduce the bug?

Parse an URL consisting of only the scheme in Bun:
```js
const url = require("url");

console.log(url.parse("http://"));
```

### What is the expected behavior?

I expect to be able to parse the scheme like in Node where the output is this:
```txt
Url {
  protocol: 'http:',
  slashes: true,
  auth: null,
  host: '',
  port: null,
  hostname: '',
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: 'http://'
}
```

### What do you see instead?

I get a TypeError:
```txt
1 | const url = require("url");
2 | 
3 | console.log(url.parse("http://"));
                ^
TypeError: "http://" cannot be parsed as a URL.
      at node:url:56:61
      at node:url:4:45
      at /workspaces/pfadimh-site/test.js:3:13
```

### Additional information

I am aware that the `url.parse` function is deprecated, however it would be beneficial if it behaved equally as in Node for backwards-compatibility.

Electroid:
It's recommended to use `new URL`, but regardless this is a bug we'll fix.

paperclover:
for someone looking to fix this, it is an issue in the node:url polyfill, located at `src/js/node/url.js`.

PR #8126, opened 2024-01-12
fix(node:url): handle url with missing host on `url.parse`

kaioduarte:
### What does this PR do?
Aligns the deprecated `url.parse` function with Node's implementation.

- [x] Code changes

Closes https://github.com/oven-sh/bun/issues/8098

### DIFF
--- a/src/js/node/url.js
+++ b/src/js/node/url.js
@@ -23,7 +23,6 @@
 
 "use strict";
 
-let toASCII;
 const { URL, URLSearchParams } = globalThis;
 
 function Url() {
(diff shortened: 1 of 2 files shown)


src/js/node/url.js
```diff
@@ -315,7 +316,7 @@ Url.prototype.parse = function (url, parseQueryString, slashesDenoteHost) {
        * have non-ASCII characters, i.e. it doesn't matter if
        * you call it with a domain that already is ASCII-only.
        */
-      this.hostname = new URL("http://" + this.hostname).hostname;
+      this.hostname = toASCII(this.hostname);
```
kaioduarte: Reference
https://github.com/nodejs/node/blob/c25878d370c6d96abb382d9ee10d1838da4ec29d/lib/url.js#L400-L404

Jarred-Sumner: can this be lazily loaded? something like

```ts
// at the top of the file:
let toASCII;
```

```suggestion
      this.hostname = (toASCII ??= require("node:punycode").toASCII)(this.hostname);
```

kaioduarte: Done! Thanks for reviewing.

paperclover (changes requested):
Is it possible to implement this fix simply for accounting for the empty hostname.

```ts
if (this.hostname)
  this.hostname = new URL('http://' + this.hostname).hostname
```

it appears we used the new URL approach because it may have been faster.

did not benchmark it myself.

i dont feel so comfortable about loading punycode at all for `node:url`, even if node.js themselves do it.

this also assumes punycode decoding actually happens with `new URL`.

if you could test these and update the code to what is the most efficient, that would be great. otherwise we can use this patch (with jarred's suggestion)

kaioduarte:
Indeed the `new URL` version is marginally faster.

image

paperclover (approved):
i added some extra tests just to be certain.

paperclover:
thank you

PR #8128, opened 2024-01-12
fix(webkit:performance): correctly track size of measures map

gvilums:
### What does this PR do?

Fix an issue in `PerformanceUserTiming.cpp` that would lead to `m_measureCounter` to get out of sync with the size of `m_measuresMap`

The tests now pass with debug assertions.


dylan-conway (approved):
nice

PR #8129, opened 2024-01-13
tidy: remove if(false) dead block

nektro:


src/bun.js/node/node_fs.zig
```diff
@@ -5609,65 +5609,6 @@ pub const NodeFS = struct {
 
             return Maybe(Return.Rm).success;
         }
-
-        if (false) {
```
dylan-conway, line 5613: This should be isWindows

dylan-conway: True branch works above 👍

dylan-conway (approved):

PR #8131, opened 2024-01-13
remove while loops where a for would be more efficient

nektro:
for loops over iterables elides indexing safety checks since its known to be able to succeed


src/js_parser.zig
```diff
@@ -8685,11 +8683,9 @@ fn NewParser_(
                     item_refs.putAssumeCapacity(name, name_loc.*);
                 }
             }
-            var i: usize = 0;
             var end: usize = 0;
 
-            while (i < stmt.items.len) : (i += 1) {
-                var item: js_ast.ClauseItem = stmt.items[i];
+            for (stmt.items) |*item| {
```
Jarred-Sumner: this

src/js_parser.zig
```diff
@@ -8721,7 +8717,7 @@ fn NewParser_(
                 }
 
                 item_refs.putAssumeCapacity(item.alias, item.name);
-                stmt.items[end] = item;
+                stmt.items[end] = item.*;
```
Jarred-Sumner: it has to be a stack copy

Jarred-Sumner:
Thank you

PR #8134, opened 2024-01-13
allow StringOrTinyString to hold 31 bytes

nektro:
small number of strings will no longer have to use the fallback


paperclover:
@dylan-conway StringOrTinyString is not used in the lockfile right?

dylan-conway:
> @dylan-conway StringOrTinyString is not used in the lockfile right?

it's used for network tasks and linking, doesn't look like it's used for lockfile

paperclover (approved):
swag

PR #8138, opened 2024-01-13
tidy: use std.mem.zeroes instead of loop or memset

nektro:
this moves the memset to comptime

### DIFF
--- a/src/bun.js/api/bun.zig
+++ b/src/bun.js/api/bun.zig
@@ -3300,7 +3300,6 @@ pub const Timer = struct {
             if (result.isAnyError()) {
                 vm.onUnhandledError(globalThis, result);
                 this.deinit();
-                //
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
Can you mark these comptime? Sometimes when we explicitly memset 0, it’s because we need to make sure the padding/alignment bytes are also 0 for serialization reasons

src/bun.js/api/bun.zig
```diff
@@ -3308,6 +3300,7 @@ pub const Timer = struct {
             if (result.isAnyError()) {
                 vm.onUnhandledError(globalThis, result);
                 this.deinit();
```
Jarred-Sumner: Empty comment

src/bun.js/api/bun.zig
```diff
@@ -3308,6 +3300,7 @@ pub const Timer = struct {
...
+                //
```
Jarred-Sumner:
```suggestion

```

Jarred-Sumner:
Thank you

PR #8141, opened 2024-01-13
make node:fs/promises.open return a FileHandle and expose fdatasync

nektro:
previously `fs.promises.open` was returning a `Number` and giving the fd directly.

this upgrade should be entirely transparent since it was also upgraded to allow the other fs/promises functions to accept a FileHandle in their `fd` options fields


src/bun.js/node/types.zig
```diff
@@ -979,8 +979,14 @@ pub const ArgumentsSlice = struct {
 pub fn fileDescriptorFromJS(ctx: JSC.C.JSContextRef, value: JSC.JSValue, exception: JSC.C.ExceptionRef) ?bun.FileDescriptor {
     return if (bun.FDImpl.fromJSValidated(value, ctx, exception) catch null) |fd|
         fd.encode()
-    else
-        null;
+    else if (value.isObject()) { // FileHandle
```
Jarred-Sumner: sorry, this is going to cause a performance issue in node:fs and we cannot do it this way

Jarred-Sumner (changes requested):
We should use a FileHandle class implemented in native code so that the `fd` property can be read without a JavaScript getter being accessed.

Jarred-Sumner (changes requested):
fs tests are failing

PR #8142, opened 2024-01-13
fix: match node's `path.parse` on dotfiles.

### LINKED ISSUE #8090: path.parse inconsistency between `Node` and `Bun` for dot files

1.0.22+b400b36ca

Darwin 23.2.0 arm64 arm

Type `bun repl` in one Terminal window, type `node` in another.
Paste `path.parse('.prettierrc')` into both.
Compare the results. Bun sets it as the ext, while Node does not.
Screenshot 2024-01-10 at 3 58 18 PM

To Match Node

### What do you see instead?

`.prettierrc` is put down as the extension also. This breaks the loader https://github.com/prettier/prettier/blob/45ad4668ebc133621c7f94e678ce399cab318068/src/config/prettier-config/load-config.js#L15

Also found a StackOverFlow where someone ran into the same issue. https://stackoverflow.com/a/77699726/458642

### Additional information

I first discovered this using `prettier.resolveConfig`. I'm working on some code that generates code based on JSON files, stuff like errors or status codes wanting in a common format. I just renamed my config file as a workaround since that works thankfully.

Part of me feels like it detecting it as an extension itself might be reasonable, so I'm surprised they didn't just add `.prettierrc` as an extension to the Prettifier loader.

But also since Bun aims to be Node compatible wonder if any other tools would break upon the same issue. Maybe the best way to address this would be treating files starting with a dot the same way as Node since if it starts with a dot, it's the file name itself and not actually the extension technically might be a way to look at it.

Aarav-Juneja:
### What does this PR do?

- [x] Code changes

My new test started passing and no other tests started failing.

I am a new developer to open source. Please tell me if anything is wrong and I need to fix anything.

### DIFF
--- a/src/fs.zig
+++ b/src/fs.zig
@@ -1392,7 +1392,7 @@ pub const NodeJSPathName = struct {
             if (strings.lastIndexOfChar(filename, '.')) |dot| {
                 if (dot > 0) {
                     filename = filename[0..dot];
-                    ext = filename[dot..];
+                    ext = base[dot..];
                 }
             }
         }
--- a/src/fs.zig
+++ b/src/fs.zig
@@ -1390,9 +1390,10 @@ pub const NodeJSPathName = struct {
         if (filename.len > 1) {
             // Strip off the extension
             if (strings.lastIndexOfChar(filename, '.')) |dot| {
-                ext = filename[dot..];
-                if (dot > 0)
+                if (dot > 0) {
                     filename = filename[0..dot];
+                    ext = filename[dot..];
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


paperclover (approved):

paperclover:
Fixes #8090

paperclover (changes requested):
there are failing tests

https://github.com/oven-sh/bun/actions/runs/7509482444?pr=8142

path.test.js is the one of note.

Aarav-Juneja:
I have fixed the issue and it now passes tests.
Screenshot 2024-01-14 at 12 33 52 PM

paperclover (approved):

PR #8143, opened 2024-01-13
fix(install): add warning and skip patched packages

dylan-conway:
### What does this PR do?
adds a warning for patched packages in overrides/resolutions and skips adding it to the map. This prevents hanging when using the patched package version string to resolve the package.
Screenshot 2024-01-12 at 5 43 29 PM

tested on the jest repo


Jarred-Sumner:
nice

PR #8147, opened 2024-01-13
For build scripts, symlink `zig.exe` to `zig` on posix

Jarred-Sumner:
### What does this PR do?

Maybe this will workaround the zls issue 

checked the symlink is created

### DIFF
--- a/scripts/download-zig.sh
+++ b/scripts/download-zig.sh
@@ -61,6 +61,7 @@ update_repo_if_needed() {
     printf "Zig was updated to ${zig_version}. Please commit new files."
 
     # symlink extracted zig to  extracted zig.exe
+    # TODO: Workaround for https://github.com/ziglang/vscode-zig/issues/164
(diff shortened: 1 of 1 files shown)


paperclover:
See https://github.com/ziglang/vscode-zig/issues/164 for context.

I think this will work as a workaround. Clever solution

scripts/download-zig.sh
```diff
@@ -59,6 +59,11 @@ update_repo_if_needed() {
     done
 
     printf "Zig was updated to ${zig_version}. Please commit new files."
+
+    # symlink extracted zig to  extracted zig.exe
```
paperclover, line 63:
```suggestion
    # symlink extracted zig to  extracted zig.exe
    # TODO: Workaround for https://github.com/ziglang/vscode-zig/issues/164
```

paperclover (changes requested):

PR #8151, opened 2024-01-13
feat: Implement `using` keyword (explicit resource management)

### LINKED ISSUE #4448: Support TypeScript@5.2 "using" declarations

"ES Explicit Resource Management" is stage 3. It works similarly to `defer` in Go or `using` in C#. Allows to define a clean up method which is run before the `return` of a function.

```ts
// From TS Blog Post
function loggy(id: string): Disposable {
    console.log(`Creating ${id}`);

    return {
        [Symbol.dispose]() {
            console.log(`Disposing ${id}`);
        }
    }
}

function func() {
    using a = loggy("a");
    using b = loggy("b");
    {
        using c = loggy("c");
        using d = loggy("d");
    }
    using e = loggy("e");
    return;

    // Unreachable.
    // Never created, never disposed.
    using f = loggy("f");
}

func();
// Creating a
// Creating b
// Creating c
// Creating d
// Disposing d
// Disposing c
// Creating e
// Disposing e
// Disposing b
// Disposing a
```

It's been implemented on TypeScript as of version 5.2+ as well as in esbuild v0.18.7. It'd be great to have this in Bun.js.

1. [**using** Declarations and Explicit Resource Management](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management)
2. [Esbuild v0.18.7](https://github.com/evanw/esbuild/releases/tag/v0.18.7)
3. [TC39 - ECMAScript Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management)

Opening this issue mostly to track when this will be implemented in Bun.js. It currently doesn't recognize the keyword `using`.

_No response_

paperclover:
i got bored and implemented this.

Closes #4448

PR is not yet complete.

- [X] Parse `using`
- [X] Parse `using await`
- [x] Lower using syntax to try/catch
- [x] Upgrade WebKit to include `Symbol.dispose`/`Symbol.asyncDispose` (see https://github.com/oven-sh/WebKit/pull/37)
- [x] Upgrade WebKit to include `SuppressedError`
- [x] Test `Symbol.dispose` exists
- [x] Test lowering works in runtime and --target=bun
- [x] Test lowering works in --target=browser/node
- [x] Fix symbol collision bug
- [x] Fix --minify-syntax

Note: this does not implement `DisposableStack` or `AsyncDisposableStack` or the iterable prototype using. I will do this in a follow up.


erictheswift:
@paperdave
Trying/adopting `using`s stopped me from using bun because of lack of native support.
Having thought of dropping `using` usages (I like how it sounds :D), I just discovered you're implementing `using` support in bun — thank you, that sounds quite nice!

Please let me know if/when aarch64 build artifacts are available for early testing ([latest build fails](https://github.com/oven-sh/bun/actions/runs/7511899477/job/20451995261#step:5:2270))

src/bundler.zig
```diff
@@ -1821,16 +1821,17 @@ pub const Bundler = struct {
...
-                options.OutputFile.initBuf(
-                    runtime.Runtime.sourceContent(false),
-                    bun.default_allocator,
-                    Linker.runtime_source_path,
-                    .js,
-                    null,
-                    null,
-                ),
-            );
+            @panic("that's crazy");
```
Jarred-Sumner:
```suggestion

```

src/js_parser.zig
```diff
@@ -21512,6 +21794,312 @@ fn NewParser_(
             p.log.addRangeError(p.source, logger.Range{ .loc = comma_after_spread, .len = 1 }, "Unexpected \",\" after rest pattern") catch unreachable;
         }
 
+        /// When not transpiling we dont use the renamer, so our solution is to generate really
```
Jarred-Sumner, line 21797:
```suggestion
        /// When not bundling we dont use the renamer, so our solution is to generate really
```

PR #8162, opened 2024-01-14
Use one isosubspace for all generated constructors

Jarred-Sumner:
### What does this PR do?

Constructors in our code generator do not have any fields in C++ and do not have destructors, so we do not need to create a unique isosubspace for each one. We can instead create a single isosubspace and use one for all of them. This is also what Safari does in their code generator.

Manual


aXXo-dev:
(❁´◡`❁)

src/codegen/generate-classes.ts
```diff
@@ -1927,7 +1939,7 @@ void GlobalObject::visitGeneratedLazyClasses(GlobalObject *thisObject, Visitor&
 const ZIG_GENERATED_CLASSES_HEADER = `
 /// Generated code! To regenerate, run:
 ///
-///    make codegen
+///    bun run build
 ///
 /// This file is generated by:
 ///  1. \`bun src/bun.js/scripts/generate-classes.ts\`
```
paperclover, line 1945: since we're editing this comment, we should fix the entire comment

```suggestion
///  1. \`bun src/codegen/generate-classes.ts\`
```

paperclover (approved):
looks good.

i wonder if we could further reduce the number of subspaces in the future, considering many classes just have a m_ctx pointer, where it exists in the zig space.

Jarred-Sumner:
There are security benefits to having separate isosubspaces, but since there are no Zig pointers associated with constructors it doesn't really matter there

PR #8166, opened 2024-01-14
internal: remove secret hidden internals and introduce new way to call native code from js

paperclover:
### What does this PR do?

*ooooh magic javascript preprocessor shenanigans.*

This PR removes `$lazy`, and replaces it with two alternatives: `$cpp` and `$zig`. These both take a filename and a function name, and they will call the function defined with no arguments.

- [X] Preprocessor Transform
- [X] Codegen
- [X] New impl of `@lazy`

**Examples:**

image

image

With this done, I will also be removing the following definitions

- [X] `globalThis[Symbol.for("Bun.lazy")]`
- [x] `Bun.fs`
- [x] `Bun._Path`
- [x] `Bun._Os()`

The API `Bun.DO_NOT_USE_OR_YOU_WILL_BE_FIRED_mimalloc_dump` has been renamed to `Bun.unsafe.dumpMimalloc()`. While left undocumented, this will hopefully reduce the number of people fired from bun.

If internals are required for unit testing, two mechanisms will be provided. In order for either to work, the environment variables `BUN_GARBAGE_COLLECTOR_LEVEL` and `BUN_FEATURE_FLAG_INTERNAL_FOR_TESTING` must be set, or you must be running in a debug build.

~~From there, runtime versions of `$zig` and `$cpp` will exist on the global object. These work the same as the bundled ones, but only map to symbols that have a used binding in the `src/js`. Perhaps we can add a way to add additional bindings that aren't used in `src/js` but able to be exposed for testing, such as the syntax highlighter.~~

From there, you can use `bun:internal-for-testing`, which currently contains three declarations:

image

Extra Tasks

- [X] Internal Typescript Definitions and Documentation
- [X] Merge bundle-functions and bundle-modules once and for all
- [X] Finish porting old code to new API


Jarred-Sumner:
Exciting

We can keep DO_NOT_USE_OR_YOU_WILL_BE_FIRED_mimalloc_dump imo

src/bun.js/api/bun.zig
```diff
@@ -935,7 +935,8 @@ fn fs(globalObject: *JSC.JSGlobalObject, _: *JSC.CallFrame) callconv(.C) JSC.JSV
 }
 
 fn _Os(globalObject: *JSC.JSGlobalObject, _: *JSC.CallFrame) callconv(.C) JSC.JSValue {
-    return Node.Os.create(globalObject);
+    _ = globalObject;
+    @panic("lose");
```
Jarred-Sumner: lol

Jarred-Sumner:
v fancy

src/bun.js/bindings/JS2Native.cpp
```diff
@@ -0,0 +1,77 @@
...
+      return ByteStream__JSReadableStreamSource__load(lexicalGlobalObject);
+    }
+  }
+
+#if BUN_DEBUG
+  // in release, it is most likely that a negative int will be hit,
+  // and a segfault will happen instead of this message.
+  //
+  // that is ok considering we do not expose this function to the public
+  CRASH_WITH_INFO("Invalid call to @native. If you aren't calling this directly then bug @paperdave as they made a mistake in the code generator");
```
jdalton: I like the this error message :)

src/bun.js/bindings/NodeVMScript.cpp
```diff
@@ -595,4 +595,28 @@ void NodeVMScript::destroy(JSCell* cell)
...
+        vm, JSC::PropertyName(JSC::Identifier::fromString(vm, "isContext"_s)),
+        JSC::JSFunction::create(vm, globalObject, 0, "isContext"_s, vmModule_isContext, ImplementationVisibility::Public), 0);
+    obj->putDirect(
+        vm, JSC::PropertyName(JSC::Identifier::fromString(vm, "runInNewContext"_s)),
+        JSC::JSFunction::create(vm, globalObject, 0, "runInNewContext"_s, vmModuleRunInNewContext, ImplementationVisibility::Public), 0);
+    obj->putDirect(
+        vm, JSC::PropertyName(JSC::Identifier::fromString(vm, "runInThisContext"_s)),
+        JSC::JSFunction::create(vm, globalObject, 0, "runInThisContext"_s, vmModuleRunInThisContext, ImplementationVisibility::Public), 0);
+    return obj;
+}
```
jdalton, line 622: Nice :)

src/bun.js/bindings/Path.cpp
```diff
@@ -184,27 +185,30 @@ static JSC::JSObject* createPath(JSGlobalObject* globalThis, bool isWindows)
...
 
 } // namespace Zig
 
-extern JSC__JSValue Bun__Path__create(JSC::JSGlobalObject* globalObject, bool isWindows)
+namespace Bun {
+    
+JSC::JSValue createNodePathBinding(Zig::GlobalObject* globalObject)
 {
     JSC::VM& vm = globalObject->vm();
+    auto binding = constructEmptyArray(globalObject, nullptr, 2);
```
jdalton, line 185: Would this be better:
```
 binding = JSC::JSArray::tryCreateUninitializedRestricted(
            globalObject,
            globalObject->arrayStructureForIndexingTypeDuringAllocation(JSC::ArrayWithContiguous),
            2);
```

src/bun.js/bindings/Path.cpp
```diff
@@ -184,27 +185,30 @@ static JSC::JSObject* createPath(JSGlobalObject* globalThis, bool isWindows)
...
+    binding->putByIndexInline(
+        globalObject,
+        (unsigned)0,
+        Zig::createPath(globalObject, false),
+        false
+    );
+    binding->putByIndexInline(
```
jdalton, line 191: What's the diff between putByIndexInline vs putDirectIndex?

src/bun.js/bindings/ProcessBindingTTYWrap.cpp
```diff
@@ -394,12 +394,12 @@ class TTYWrapConstructor final : public JSC::InternalFunction {
...
 const ClassInfo TTYWrapConstructor::s_info = { "TTY"_s, &Base::s_info, nullptr, nullptr, CREATE_METHOD_TABLE(TTYWrapConstructor) };
 
-JSValue createBunTTYFunctions(JSC::JSGlobalObject* globalObject)
+JSValue createBunTTYFunctions(Zig::GlobalObject* globalObject)
 {
     auto& vm = globalObject->vm();
     auto* obj = constructEmptyObject(globalObject);
 
-    obj->putDirect(vm, PropertyName(Identifier::fromString(vm, "isTTY"_s)), JSFunction::create(vm, globalObject, 0, "isatty"_s, Zig::jsFunctionTty_isatty, ImplementationVisibility::Public), 0);
+    obj->putDirect(vm, PropertyName(Identifier::fromString(vm, "isatty"_s)), JSFunction::create(vm, globalObject, 0, "isatty"_s, Zig::jsFunctionTty_isatty, ImplementationVisibility::Public), 0);
```
jdalton, line 499: ☝️ This fix a different bug?

src/bun.js/bindings/bindings.zig
```diff
@@ -5596,6 +5621,18 @@ pub const JSArray = struct {
...
     }
+
+    /// Must be called with a tuple of JSValue
+    pub inline fn createComptime(
+        global: *JSGlobalObject,
+        items: anytype,
+    ) JSValue {
+        const a = JSValue.createEmptyArray(global, comptime items.len);
+        inline for (items, 0..) |item, i| {
+            a.putIndex(global, i, item);
```
jdalton: would putDirectIndex be better?

src/js/node/readline.js
```diff
@@ -27,6 +27,8 @@
 // ----------------------------------------------------------------------------
 const EventEmitter = require("node:events");
 const { StringDecoder } = require("node:string_decoder");
+const internalGetStringWidth = $newZigFunction("string.zig", "String.jsGetStringWidth", 1);
```
jdalton, line 30: Neeeeaaattt

src/bun.js/bindings/Path.cpp
```diff
@@ -184,27 +185,30 @@ static JSC::JSObject* createPath(JSGlobalObject* globalThis, bool isWindows)
...
-        path->putDirect(vm, clientData->builtinNames().delimiterPublicName(),
-            JSC::jsString(vm, WTF::String(";"_s)), 0);
-    } else {
-        path->putDirect(vm, clientData->builtinNames().sepPublicName(),
-            JSC::jsString(vm, WTF::String("/"_s)), 0);
-        path->putDirect(vm, clientData->builtinNames().delimiterPublicName(),
-            JSC::jsString(vm, WTF::String(":"_s)), 0);
-    }
-
     return path;
```
jdalton, line 176: Can we use the added JSC_zig function helpers here to cleanup the creation of the zig to c++ methods?

src/bun.js/bindings/ProcessBindingTTYWrap.cpp
paperclover: no it just matches the name we actually assign this function into at the binding

image

src/bun.js/bindings/Path.cpp
paperclover: i inlined these constant strings right into the js code

src/bun.js/bindings/Path.cpp
paperclover: image

reading this comment, tbh no.

src/bun.js/bindings/Path.cpp
paperclover: i looked at both source codes and theyre very similar. in this case they both call `setIndexQuickly`

src/bun.js/bindings/bindings.zig
paperclover: i dont think we have a binding for it, i might be wrong.

src/bun.js/api/bun/h2_frame_parser.zig
```diff
@@ -2526,3 +2521,11 @@ pub const H2FrameParser = struct {
         this.deinit();
     }
 };
+
+pub fn createNodeHttp2Binding(global: *JSC.JSGlobalObject) JSC.JSValue {
+    return JSC.JSArray.createComptime(global, .{
```
jdalton: Is this `JSC.JSArray.createComptime` being called not-at-comptime?

paperclover: this should get a clearer name

it is inferring the length at comptime and doing an `inline for` over it.

src/bun.js/bindings/Path.cpp
jdalton: eeee. that's some disclaimer. We use `JSC::JSArray::tryCreateUninitializedRestricted(` in places, I wonder if we should rethink it.

src/bun.js/api/bun/h2_frame_parser.zig
jdalton: (learning): From a zig POV how can it comptime the `items.len` when `items` is an `anytype` not known at comptime. Is it because `JSC.JSArray.createComptime` is designated as `inline` and so inlined at comptime and because `items` are fixed length literals?

paperclover: it is seen as a fixed-size array or tuple in this usage, so it knows there are two items. it doesn't know the values of this at compile time, but it knows the count.

src/bun.js/bindings/JS2Native.cpp
```diff
@@ -0,0 +1,81 @@
...
+    JSC::JSValue target = callFrame->uncheckedArgument(0);
+
+#if BUN_DEBUG
+    ASSERT_WITH_MESSAGE(target.isInt32(), "In call to $lazy: expected Int32, got %s", target.toWTFString(lexicalGlobalObject).utf8().data());
+#endif
+
+    int id = target.asInt32();
+    if (LIKELY(id < 0)) {
+#if BUN_DEBUG
+        ASSERT_WITH_MESSAGE("In call to $lazy: expected int in range, got %d. This is a bug in JS2Native code generator.", id);
```
dylan-conway: this assert is always true. is there a minimum it should compare with?

paperclover: yes

src/bun.js/bindings/bindings.cpp
```diff
@@ -5433,6 +5433,30 @@ extern "C" EncodedJSValue ExpectStatic__getPrototype(JSC::JSGlobalObject* global
...
 
+extern "C" EncodedJSValue JSFunction__createFromZig(
+    JSC::JSGlobalObject* global,
+    BunString fn_name,
+    NativeFunction implementation,
+    unsigned arg_count,
+    ImplementationVisibility implementation_visibility,
+    Intrinsic intrinsic,
+    NativeFunction constructorOrNull,
+    const DOMJIT::Signature* domjit)
```
Jarred-Sumner: Let's omit this, as it is a constexpr property which cannot be correctly passed through Zig

src/bun.js/bindings/bindings.zig
```diff
@@ -5736,6 +5761,25 @@ pub const JSArray = struct {
...
+        for (items, 0..) |item, i| {
+            a.putIndex(global, i, item);
+        }
+        return a;
+    }
+
+    /// Must be called with a tuple of JSValue
+    pub inline fn createInline(global: *JSGlobalObject, items: anytype) JSValue {
+        // TODO(@paperdave): make this use JSArray::constructArray and initializeIndex instead
+        const a = JSValue.createEmptyArray(global, comptime items.len);
```
Jarred-Sumner: Instead of this, let's use `JSC::constructArray`: https://github.com/oven-sh/bun/blob/a82d61093522ac33b9755b0ac82bde1f77d75f56/src/bun.js/bindings/webcore/JSDOMConvertSequences.h#L393


It has an overload which accepts a JSC::ArgList which is a pointer and length to JSValue's.

src/bun.js/bindings/bindings.zig
```diff
@@ -5736,6 +5761,25 @@ pub const JSArray = struct {
     pub fn from(globalThis: *JSGlobalObject, arguments: []const JSC.JSValue) JSValue {
         return JSC.JSValue.c(JSC.C.JSObjectMakeArray(globalThis, arguments.len, @as(?[*]const JSC.C.JSObjectRef, @ptrCast(arguments.ptr)), null));
     }
+
+    pub fn create(global: *JSGlobalObject, items: []JSValue) JSValue {
+        // TODO(@paperdave): make this use JSArray::constructArray and initializeIndex instead
+        const a = JSValue.createEmptyArray(global, items.len);
+        for (items, 0..) |item, i| {
```
Jarred-Sumner: Let's also delete this one

Jarred-Sumner:
merge conflicts

paperclover:
this PR is marked breaking as it removes some hidden apis. we will merge this before 1.1 but before the last 1.0.x

paperclover:
test/js/bun/wasm/wasi.test.js test i think is real

Issue #5489, opened 2023-09-15
Bun build failing: thread 2298039 panic: Internal error: missing asset file


robobun:
Hi all, having an issue with Bun build:

```
uh-oh: Internal error: missing asset file
bun will crash now 😭😭😭

----- bun meta -----
Bun v1.0.1 (31aec4eb) macOS Silicon 22.3.0
BuildCommand: 
Elapsed: 15ms | User: 26ms | Sys: 20ms
RSS: 42.02MB | Peak: 42.02MB | Commit: 89.31MB | Faults: 17
----- bun meta -----

0   0x1064b4d24 WTFGetBacktrace
1   ??? Bun__crashReportDumpStackTrace

Crash report saved to:
  ~/.bun/.bun-crash/v1.0.1-1694796321166.crash

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord

thread 2298039 panic: Internal error: missing asset file
```

Definitely something to do with what's included in my build, but the error doesn't provide much context.

_Originally reported on Discord: [`Bun build failing: thread 2298039 panic: Internal error: missing asset file`](https://discord.com/channels/876711213126520882/1152284651663143064/1152284651663143064)_

chrisdmacrae:
FYI this was caused by using chokidar in the app

manutzsong:
` bun build --target=bun --external canvas ./src/index.ts --outdir ./dist`

I use this command and it build successful. However, still failed to other error.

```
1 | __dirname
   ^
ReferenceError: Can't find variable: __dirname

```

danestves:
> FYI this was caused by using chokidar in the app

I can say that facing the same issue, If I remove chokidar from the code, it runs smoothly

zp1ke:
I have winston dependencies, and I am facing the same issue. Is there any solution? I don't think removing or changing dependencies is a solution.

Micnubinub:
Getting the same thing in 1.0.4

```
uh-oh: Internal error: missing asset file
bun will crash now 😭😭😭

----- bun meta -----
Bun v1.0.4 (745b6b94) macOS Silicon 23.0.0
BuildCommand: dotenv external 
Elapsed: 171ms | User: 391ms | Sys: 122ms
RSS: 0.48GB | Peak: 0.48GB | Commit: 0.67GB | Faults: 38
----- bun meta -----

0   0x103d5cd50 WTFGetBacktrace
1   ??? Bun__crashReportDumpStackTrace

...

[...]

Issue #3903, opened 2023-07-31
compile: missing asset file (hard to debug)


janwilmake:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Darwin 21.6.0 arm64 arm

### What steps can reproduce the bug?

compile a big codebase, for some reason, it gives "missing asset file" for me. 

### What is the expected behavior?

I get a stack trace or something that would tell me what exactly is missing and where

### What do you see instead?

Just "missing asset file" is not good enough for me to solve the problem

### Additional information

 ```
bun-server bun run build
$ bun build ./src/server.ts --compile --outfile server

uh-oh: Internal error: missing asset file
bun will crash now 😭😭😭

----- bun meta -----
Bun v0.7.1 (53cc4df1) macOS Silicon 21.6.0
BuildCommand: 
Elapsed: 137ms | User: 287ms | Sys: 95ms
RSS: 0.34GB | Peak: 0.34GB | Commit: 0.41GB | Faults: 0
----- bun meta -----

0   0x102042b4c WTFGetBacktrace
1   ??? Bun__crashReportDumpStackTrace

Crash report saved to:
  ~/.bun/.bun-crash/v0.7.1-1690787568648.crash

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord

thread 1236859 panic: Internal error: missing asset file
error: script "build" exited with code 6 (SIGABRT)
```

Jarred-Sumner:
Are any of the packages using .node files?  It's likely caused by that

Jarred-Sumner:
(but yes the error is terrible and needs to be made clearer)

janwilmake:
Yes there are multiple packages I use that have .node files, such as `@swc/core`, `uiohook-napi` and `tiktoken-node`. I certainly need `@swc/core` (https://swc.rs) so wondering if this will become possible 

Edit: removed unused packages `uiohook-napi` and `tiktoken-node`... Still getting the same. Maybe the `.node` file in `@swc/core`

Jarred-Sumner:
yeah unfortunately `bun build --compile` and `bun build` lacks . node file support right now. `bun build` is entirely optional though, you can use bun without the build step

CM-IV:
I'm having the same issue using a library made with the `age` Rust crate via `napi-rs`.

Electroid:
Related: #6567

Issue #3653, opened 2023-07-16
`@napi-rs/canvas` does not bundle/compile


jkbz64:
### What version of Bun is running?

0.6.14

### What platform is your computer?

Linux 6.3.0-asahi-13-1-edge-ARCH aarch64 unknown

### What steps can reproduce the bug?

Trying to use `bun build --target bun` or `bun build --target bun --compile` on a script that imports/uses  `@napi-rs/canvas` library  crashes.

### What is the expected behavior?

It should bundle/create a standalone executable.

### What do you see instead?

```
uh-oh: Internal error: missing asset file
bun will crash now 😭😭😭

----- bun meta -----
Bun v0.6.14 (b5665739) Linux arm64 #2 SMP PREEMPT_DYNAMIC Thu, 13 Jul 2023 00:24:08 +0000
BuildCommand: 
Elapsed: 16ms | User: 2ms | Sys: 21ms
RSS: 100.78MB | Peak: 71.83MB | Commit: 100.78MB | Faults: 50
----- bun meta -----

0   0xaaab000effe8

Crash report saved to:
  ~/.bun/.bun-crash/v0.6.14-1689546595701.crash

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord

thread 1984 panic: Internal error: missing asset file
[1]    1984 IOT instruction (core dumped)  bun build --target bun --compile index.js
```

birkskyum:
@jkbz64 , can you provide a reproduction of this?

jkbz64:
Just like I said in the steps to reproduce, importing `@napi-rs/canvas` is enough for it to crash, tested in on various versions between the issue one, there was no difference (in some versions like here I get more bun meta entries).

```js
// index.js
import "@napi-rs/canvas";
// Also tried with import { createCanvas } from and similar
```

Logs from the v0.8.1

```
uh-oh: Internal error: missing asset file
bun will crash now 😭😭😭

----- bun meta -----
Bun v0.8.1 (16b4bf34) Linux arm64 #2 SMP PREEMPT_DYNAMIC Mon, 21 Aug 2023 05:46:52 +0000
BuildCommand: 
Elapsed: 15ms | User: 6ms | Sys: 22ms
RSS: 101.79MB | Peak: 70.25MB | Commit: 101.79MB | Faults: 0
----- bun meta -----

0   0xaaaabf34f2b8
1   ???
2   ???
3   ???
4   ???
5   ???
6   ???
7   ???
[...]

PR #8178, opened 2024-01-15
feat: Support `import db from './my.db' with {type: "sqlite"}`

Jarred-Sumner:
### What does this PR do?

This implements `import` support for SQLite databases.

```js
import db from './my.db' with {type: "bun"};

db.query("SELECT * from users LIMIT 1");
```

This also implements support for embedding SQLite databases with `bun build --compile`.

```js
import db from './my.db' with {type: "bun", embed: "true"};

db.query("SELECT * from users LIMIT 1;");
```

 Also, fixed bundling .node files 

Fixes https://github.com/oven-sh/bun/issues/5489
Fixes https://github.com/oven-sh/bun/issues/3653
Fixes https://github.com/oven-sh/bun/issues/3903

Tests


src/bun.js/module_loader.zig
```diff
@@ -2277,6 +2327,26 @@ pub const ModuleLoader = struct {
             const specifier_utf8 = specifier.toUTF8(bun.default_allocator);
             defer specifier_utf8.deinit();
             if (graph.files.get(specifier_utf8.slice())) |file| {
+                if (file.loader == .sqlite or file.loader == .sqlite_embedded) {
+                    const code =
+                        \\/* Generated code */
+                        \\import {Database} from 'bun:sqlite';
+                        \\import {readFileSync} from 'node:fs';
+                        \\export const db = new Database(readFileSync(import.meta.path));
```
Electroid, line 2335: Is this suppose to be the same for embeded and non-embeded?

Electroid (approved):
Nice

src/bun.js/module_loader.zig
Jarred-Sumner: yeah, readFileSync will do the right thing

codehz:
maybe also add support for open flags?
like use SQLITE_OPEN_URI (so we can use it as named shared in-memory database (file:xxxx?mode=memory&cache=shared), which can be shared between macro and build script)
or just mapping more argument to attribute
```
import db from "test" with { type: "sqlite", mode: "memory", cache: "shared" }
```
mode=memory = [SQLITE_OPEN_MEMORY](https://www.sqlite.org/c3ref/c_open_autoproxy.html)
cache=shared = [SQLITE_OPEN_SHAREDCACHE](https://www.sqlite.org/c3ref/c_open_autoproxy.html)
etc

Jarred-Sumner:
> maybe also add support for open flags? like use SQLITE_OPEN_URI (so we can use it as named shared in-memory database (file:xxxx?mode=memory&cache=shared), which can be shared between macro and build script) or just mapping more argument to attribute
> 
> ```
> import db from "test" with { type: "sqlite", mode: "memory", cache: "shared" }
> ```
> 
> mode=memory = [SQLITE_OPEN_MEMORY](https://www.sqlite.org/c3ref/c_open_autoproxy.html) cache=shared = [SQLITE_OPEN_SHAREDCACHE](https://www.sqlite.org/c3ref/c_open_autoproxy.html) etc

Maybe in the future, but not right now

src/bundler/bundle_v2.zig
```diff
@@ -2501,6 +2526,116 @@ pub const ParseTask = struct {
                 }, Logger.Loc{ .start = 0 });
                 return JSAst.init((try js_parser.newLazyExportAST(allocator, bundler.options.define, opts, log, root, &source, "")).?);
             },
+
+            .sqlite_embedded, .sqlite => {
+                if (!bundler.options.target.isBun()) {
+                    log.addError(
+                        null,
+                        Logger.Loc.Empty,
```
paperclover, line 2534: this not having a location is lame.

image

someone will eventually run into this error, where they want to find and remove the import.

paperclover (changes requested):
awesome. not yet ready to ship imo.

- `bun build` without --compile will emit the same import statement with the `with` clause. this is fine but i was expecting the wrapper to be bundled in. (no action needed on this point)
- `embed: "true"` does not work outside of --compile. this should either be an error, or the generated code should just be a top level import statement like it is without embed
- `To use the "sqlite" loader, set target to "bun"` this message has no error location and will confuse people. i hope it is possible to fix this, but if not i will open an issue about it.

Jarred-Sumner:
> embed: "true" does not work outside of --compile. this should either be an error, or the generated code should just be a top level import statement like it is without embed

It does work outside of `--compile`. It is copied into the executable and loaded at runtime. It becomes a no-op at runtime. There is a test for this.

> bun build without --compile will emit the same import statement with the with clause. this is fine but i was expecting the wrapper to be bundled in. (no action needed on this point)

We rely on loading it at runtime for this

src/bundler/bundle_v2.zig
Jarred-Sumner: could add the error earlier i think

PR #8185, opened 2024-01-15
feat(cli): --filter flag

### LINKED ISSUE #2232: Support `bun run --workspace`

0.5.7

Darwin 22.3.0 arm64 arm

In a monorepo with npm workspaces, scripts defined in a specific package's `package.json` can be run with 

```sh
npm run my-script --workspace packages/my-package
```

Bun should allow me to run the package's script via the `--workspace` arg.

A possible workaround is:

```sh
(cd packages/my-package && bun run my-script)
```

### What do you see instead?

error: missing script "my-script"

### Additional information

_No response_

### LINKED ISSUE #5207: Support `bun run --workspaces`

Run a package.json script in all workspaces (if it is defined)

---

_No response_

gvilums:
### What does this PR do?

Adds support for the `--filter <PATTERN>` argument for `bun run`. See the [pnpm docs](https://pnpm.io/filtering#--filter-package_name) for details.

Note that currently only `--filter <name_pattern>` and `--filter ./<dir_pattern>` are supported. Selecting dependencies or dependents of packages is not yet possible.

Resolves #2232
Resolves #5207 

Tests in `test/cli/run/filter-workspace.test.ts`.


src/cli/run_command.zig
```diff
@@ -1010,8 +1009,61 @@ pub const RunCommand = struct {
...
+            _ = try exec(ctx, bin_dirs_only, true);
+            return;
+        }
+
+        const fsinstance = try bun.fs.FileSystem.init(null);
+        const olddir = fsinstance.top_level_dir;
+        defer {
+            // change back to the original directory once we're done
+            fsinstance.top_level_dir = olddir;
+            std.os.chdir(olddir) catch |err| {
```
dylan-conway:
```suggestion
            switch (bun.sys.chdir(olddir)) {
                .err => |err| {
                    ...
                },
                .success => {},
            }
```

src/cli/run_command.zig
```diff
@@ -1010,8 +1009,61 @@ pub const RunCommand = struct {
...
+
+        if (workspace_paths.items.len == 0) {
+            Output.prettyErrorln("<r><red>error<r>: No packages matched the filter", .{});
+            Global.exit(1);
+        }
+
+        var ok = true;
+        for (workspace_paths.items) |path| {
+            Output.prettyErrorln("<d><b>In <r><yellow><d>{s}<r>:", .{path});
+            std.os.chdir(path) catch |err| {
```
dylan-conway: `bun.sys.chdir` here too

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,183 @@
...
+    defer {
+        bun.JSAst.Expr.Data.Store.reset();
+        bun.JSAst.Stmt.Data.Store.reset();
+    }
+
+    var workdir = workdir_;
+
+    while (true) : (workdir = std.fs.path.dirname(workdir) orelse break) {
+        const parent_trimmed = strings.withoutTrailingSlash(workdir);
+        var buf2: [bun.MAX_PATH_BYTES + 1]u8 = undefined;
```
dylan-conway:
```suggestion
        var buf2: bun.PathBuffer = undefined;
```
is the `+ 1` needed? null byte should be included

src/cli.zig
```diff
@@ -402,20 +409,26 @@ pub const Arguments = struct {
...
-                    Output.prettyErrorln("error resolving --cwd: {s}", .{@errorName(err)});
-                    Global.exit(1);
-                };
-                break :brk try allocator.dupe(u8, out);
+            var outbuf: [bun.MAX_PATH_BYTES]u8 = undefined;
+            const out = std.os.realpath(cwd_, &outbuf) catch |err| {
+                Output.prettyErrorln("error resolving --cwd: {s}", .{@errorName(err)});
+                Global.exit(1);
+            };
+            std.os.chdir(out) catch |err| {
```
dylan-conway: also `bun.sys.chdir`

test/cli/run/filter-workspace.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  let cmd = [bunExe(), "run"];
+  if (pattern instanceof Array) {
+    for (let p of pattern) {
+      cmd.push("--filter", p);
+    }
+  } else {
+    cmd.push("--filter", pattern);
+  }
+  cmd.push("present");
+  console.log(cmd);
```
dylan-conway: these console logs can be removed?

dylan-conway (commented):
Looks good so far. One thing that stands out as a difference from pnpm: `pnpm run --filter="*" debug` will run all `debug` scripts including one from the root package

src/cli/filter_arg.zig
gvilums: updated to use PathBuffer

test/cli/run/filter-workspace.test.ts
gvilums: removed

gvilums:
Just pushed a commit that includes the root package in script resolution

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,195 @@
...
+        // add the root package to the workspace map, too
+        const entry = Package.WorkspaceMap.Entry{ .name = try allocator.dupe(u8, name), .version = null, .name_loc = bun.logger.Loc.Empty };
+        try workspace_map.insert(try allocator.dupe(u8, parent_trimmed), entry);
+
+        return;
+    }
+
+    // if we were not able to find a workspace root, try globbing for package.json files
+
+    var walker = Glob.BunGlobWalker{};
```
dylan-conway: this glob walker should ignore `node_modules` and other directories that pnpm ignores if they exist. I believe this is done through the `ignore_filter_fn`

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,195 @@
...
+        const json_source = bun.logger.Source.initPathString(path, json_buf[0..json_len]);
+
+        var parser = try json_parser.PackageJSONVersionChecker.init(allocator, &json_source, log);
+        _ = try parser.parseExpr();
+        if (!parser.has_found_name) {
+            continue;
+        }
+        const entry = Package.WorkspaceMap.Entry{ .name = try allocator.dupe(u8, parser.found_name), .version = null, .name_loc = bun.logger.Loc.Empty };
+        const dirpath = std.fs.path.dirname(path) orelse continue;
+        try workspace_map.insert(try allocator.dupe(u8, dirpath), entry);
```
dylan-conway: package jsons found here may or may not actually be workspace packages. If the script isn't found in the `package.json` we shouldn't print an error
Screenshot 2024-01-17 at 4 06 37 PM

Jarred-Sumner:
Running `bun-debug run --filter="*" echo hi` crashes

image

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,195 @@
...
+            prop.loc,
+            null,
+        ) catch |err| {
+            return err;
+        };
+
+        // add the root package to the workspace map, too
+        const entry = Package.WorkspaceMap.Entry{ .name = try allocator.dupe(u8, name), .version = null, .name_loc = bun.logger.Loc.Empty };
+        try workspace_map.insert(try allocator.dupe(u8, parent_trimmed), entry);
+
```
Jarred-Sumner, line 98: This duplicates memory twice.

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,195 @@
...
+const string = bun.string;
+const Output = bun.Output;
+const Global = bun.Global;
+const strings = bun.strings;
+const json_parser = bun.JSON;
+const Glob = @import("../glob.zig");
+
+const Package = @import("../install/lockfile.zig").Package;
+
+fn findWorkspaceMembers(allocator: std.mem.Allocator, log: *bun.logger.Log, workspace_map: *Package.WorkspaceMap, workdir_: []const u8) !void {
```
Jarred-Sumner: In large directory trees, this approach takes seconds to show any output. We cannot ship it without being an iterator.

Jarred-Sumner (changes requested):
4 issues before this can be shipped.

1) This approach is not fast enough. It needs to be an iterator so that the developer gets immediate feedback instead of waiting for potentially their entire directory tree to parse every package.json file.

2) There is a resource leak. After enough scripts are run, spawning any process fails with `SystemResources` which usually means a file descriptor leak.

3) it scans node_modules. It should not scan node_modules, .git, or any `.` directories.

4) We need tests for what happens when you run multiple JavaScript/TypeScript files in different projects same process (`bun run --filter="*" index.js`). Currently, Bun assumes only one JavaScript file every executes for a given thread. That assumption is not true when using these code paths.

5) It would be best to use the module resolution cache so that when running a JavaScript file later, the subsequent calls to getdents64() get cached. This will involve a small change to the GlobWalker.

src/cli/filter_arg.zig
gvilums: The `--if-present` flag accomplishes this, but maybe this should be the default, just like for pnpm.

src/cli/filter_arg.zig
gvilums: Does it? From my understanding we dupe the name (part of the hash map value) in the first line, and duple the path (hash map key) in the second line.

Jarred-Sumner: Right here:

https://github.com/oven-sh/bun/blob/52894121ea2436481b97b5318d1ce64176b18c0e/src/install/lockfile.zig#L3562-L3573

src/cli/filter_arg.zig
gvilums: understood, will get on that 👍

src/cli/filter_arg.zig
gvilums: fixed in latest commit

src/cli/run_command.zig
```diff
@@ -1142,6 +1147,8 @@ pub const RunCommand = struct {
                 .result => {},
             }
             fsinstance.top_level_dir = path;
```
Jarred-Sumner: only this one should be necessary

gvilums:
> 4 issues before this can be shipped.
> 
> 1. This approach is not fast enough. It needs to be an iterator so that the developer gets immediate feedback instead of waiting for potentially their entire directory tree to parse every package.json file.

This is now implemented in the `PackageFilterIterator ` in `filter_args.zig`
> 2. There is a resource leak. After enough scripts are run, spawning any process fails with `SystemResources` which usually means a file descriptor leak.

I'm able to reproduce this, but running `lsof -p <pid>` to list the currently open file descriptors shows only a handful at any given time. Not sure what's causing this issue. 

Interestingly, when running a large number of js files, one fd seems to be leaked per file that's executed. Not sure where those are coming 
(EDIT: there was a preexisting fd leak in run_command.zig that is now fixed. Error when just running `echo hi` persists).
(EDIT 2: Error when running `echo hi` is fixed)

> 3. it scans node_modules. It should not scan node_modules, .git, or any `.` directories.

Implemented by passing an ignore function to the `GlobWalker`
> 4. We need tests for what happens when you run multiple JavaScript/TypeScript files in different projects same process (`bun run --filter="*" index.js`). Currently, Bun assumes only one JavaScript file every executes for a given thread. That assumption is not true when using these code paths.

We now `fork` before executing any JS. This of course doesn't work on windows, and more generally we should decide how to handle this. Options include:
- Figure out the correct command line arguments and spawn a new copy of the bun process. Slowest, but would provide complete isolation. 
- Continue using fork and use something like `RtlCloneUserProcess` on windows. Caveat is that these are not "official" windows APIs and their stability is unknown.
- Work on isolating VM instances and run everything in the same process. Fastest, but also potentially risky?
> 5. It would be best to use the module resolution cache so that when running a JavaScript file later, the subsequent calls to getdents64() get cached. This will involve a small change to the GlobWalker.

Implemented by abstracting how the `GlobWalker` does filesystem access.

gvilums:
After investigating 2. more, it seems like the error lies somewhere in child_process.spawnAndWait. Specifically, the error that's happening is that the child fails between being forked and calling execv, and signals this error to the parent. 

EDIT:
After switching the implementation to posix_spawn, the specific error that comes up seems to be E2BIG, which has a description of "The number of bytes in the new process's argument list is larger than the system-imposed limit.  This limit is specified by the sysctl(3) MIB variable KERN_ARGMAX." On my machine, this error starts occurring after spawning roughly 3000 processes. I'm not aware of any resource leaks that could cause this (after all, each process is spawned with an argument list that's basically identical, and each process has its own argument list), but I propose that we leave this issue be for now, as it is an extreme edge case

EDIT 2:
So I was able to solve the mystery: The culprit was an ever-growing path variable, which, each time an attempt at executing a package was made, was extended with new items. This lead to it growing unmanageably large after a few thousand executions. The fix was to reset PATH after every execution.

src/cli/run_command.zig
gvilums: The `cwd` member does seem to be used in the `kind` function [here](https://github.com/oven-sh/bun/blob/26a720defe3a0962f8b1062734b221b838613d4a/src/fs.zig#L1252-L1270), but that's the only spot. The other places in the `fs` implementation just use `std.fs.cwd`. So I think to ensure that the `kind` of a file is correctly resolved this should be updated (but maybe I'm wrong on that, in general the `cwd` handling is kind of a mess)

MickL:
Does this also work with `bun add`? It is very annoying that I always need to go in the directory. pnpn has `pnp add --filter workspace`

src/glob.zig
```diff
@@ -624,6 +784,18 @@ pub fn GlobWalker_(
             only_files: bool,
         ) !Maybe(void) {
             errdefer arena.deinit();
+            var cwd: []const u8 = undefined;
```
dylan-conway: Is this `cwd` supposed to be used with `initWithCwd` on line 788?

src/deps/tinycc
dylan-conway, line 1: It looks like there isn't a diff, was this change intentional?

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,302 @@
...
+            .{ .mode = .read_only },
+        ) catch continue;
+        defer json_file.close();
+
+        const json_stat_size = try json_file.getEndPos();
+        const json_buf = try allocator.alloc(u8, json_stat_size + 64);
+        defer allocator.free(json_buf);
+        const json_len = try json_file.preadAll(json_buf, 0);
+        const json_source = bun.logger.Source.initPathString(json_path, json_buf[0..json_len]);
+        const json = try json_parser.ParseJSONUTF8(&json_source, log, allocator);
```
Jarred-Sumner, line 69: We are parsing the package.json files multiple times here since it may happen again later 

I'm hesitant to block the PR on this but we should figure out a way to cache this to avoid the overhead 

Also, we can simplify this code a bit:

`bun.sys.File.toSource(json_path, allocator);`

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,302 @@
...
+        // TODO fixed buffer allocator with fallback?
+        var self = FilterSet{ .allocator = allocator, .filters = try allocator.alloc(Pattern, filters.len) };
+        for (0.., filters) |idx, filter_utf8_| {
+            var filter_utf8 = filter_utf8_;
+            const is_path = filter_utf8.len > 0 and filter_utf8[0] == '.';
+            if (is_path) {
+                const parts = [_]string{filter_utf8};
+                filter_utf8 = bun.path.joinAbsStringBuf(cwd, &buf, &parts, .auto);
+            }
+            var filter_utf32 = std.ArrayListUnmanaged(u32){};
```
Jarred-Sumner: feels like this will be very slow but i guess the filters won't be long so it should be fine

src/cli/run_command.zig
```diff
@@ -348,13 +349,23 @@ pub const RunCommand = struct {
...
+                }
+            }
+            allocator.free(envp);
+        }
+
         const spawn_result = switch ((bun.spawnSync(&.{
             .argv = &argv,
             .argv0 = shell_bin.ptr,
 
             // TODO: remember to free this when we add --filter or --concurrent
```
Jarred-Sumner:
```suggestion
            // TODO: remember to free this when we add --concurrent
```

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,302 @@
...
+fn globIgnoreFn(val: []const u8) bool {
+    if (val.len == 0) {
+        return false;
+    }
+    // skip hidden directories
+    if (val[0] == '.') {
+        return true;
+    }
+    // skip node_modules
+    if (strings.eqlComptime(val, "node_modules")) {
```
dylan-conway: we could skip `CMakeFiles` as well, but it's less likely to exist so probably not necessary

src/cli/filter_arg.zig
```diff
@@ -0,0 +1,302 @@
...
+                    @memcpy(pattern[pattern_expr.data.len..size], "/package.json");
+
+                    // @memcpy(pattern[0..parent_trimmed.len], parent_trimmed);
+                    // pattern[parent_trimmed.len] = '/';
+                    // @memcpy(pattern[parent_trimmed.len + 1 .. parent_trimmed.len + 1 + pattern_expr.data.len], pattern_expr.data);
+                    // @memcpy(pattern[parent_trimmed.len + 1 + pattern_expr.data.len .. size], "/package.json");
+                    try out_patterns.append(pattern);
+                },
+                else => {
+                    // TODO log error and fail
```
dylan-conway: We need an error message here

src/cli/run_command.zig
```diff
@@ -348,13 +349,23 @@ pub const RunCommand = struct {
             Output.flush();
         }
 
+        const envp = try env.map.createNullDelimitedEnvMap(allocator);
```
paperclover: on a different branch i had a version of this function that did one large allocation so that the defer free would be a single free.

code is https://github.com/oven-sh/bun/pull/9642 and might be cleaner to use.

it was put on pause because of change of focus to get 1.1 out.

src/cli/run_command.zig
paperclover: envp is freed. this entire comment can be removed now.

src/cli/bunx_command.zig
```diff
@@ -525,8 +525,7 @@ pub const BunxCommand = struct {
                     passthrough,
                     null,
                 );
-                // runBinary is noreturn
-                @compileError("unreachable");
+                Global.exit(if (res.notFailure()) 0 else 1);
```
paperclover: this is expected to exit with the code the subprocess does,
and if the subprocess exits with a signal code instead of an exit code (SIGSEGV, SIGILL), we should raise the same one. the code here will only exit with 1.

i hope there are already tests for this but you can try to `bunx node -e 'process.abort()'`

Jarred-Sumner (commented):
this needs more tests:
- what happens when a permissions error occurs when reading package.json or globbing? 
- what happens when a package.json is malformed?
- we should have a test that bun returns the appropriate exit code when a script run via --filter exits with a non-zero exit code
- we should look at tests from pnpm or yarn or npm to copy from
- a test that checks script & package names with non-ascii characters works

src/cli/run_command.zig
```diff
@@ -1280,13 +1290,144 @@ pub const RunCommand = struct {
         Output.flush();
     }
 
+    pub const ExecResult = union(enum) {
```
paperclover: can you add `signal: u32` here, this will make this similar to `std.process.Child.Term`. stopped and unknown are not as important.

and then i recommend adding some function in here to exit with the execresult, which will call exitWide for the codes, but use `raiseIgnoringPanicHandler` for signals. (posix only)

paperclover (changes requested):

dylan-conway:
> this needs more tests:
> 
> * what happens when a package.json is malformed?

Missing name and empty name are two tests we should have. Pnpm seems to not allow filtering with `--filter ""` so we can exit early when that pattern is given

Jarred-Sumner:
> Missing name and empty name are two tests we should have. 

It's reading package.jsons to check the name/version in this code, so we need to test that here too

src/deps/tinycc
gvilums: nope, not intended

src/glob.zig
gvilums: I think this is leftover code, I removed the cwd stuff

src/cli/run_command.zig
gvilums: Done

src/cli/bunx_command.zig
gvilums: Makes sense, I updated the ExecResult to include an `exit` method which does the right thing

src/cli/filter_arg.zig
gvilums: Added

src/cli/filter_arg.zig
Jarred-Sumner: we should also skip`.git`

Jarred-Sumner: but that can be added later

beorn:
Perhaps also support the '-F' shorthand?

kevinmarrec:
@gvilums @Jarred-Sumner @dylan-conway 

Dependency order is not respected as said in the docs since this PR
https://github.com/oven-sh/bun/issues/12203
https://github.com/oven-sh/bun/issues/13239

I may have never worked, or there has been a breaking chance since.

PR #8187, opened 2024-01-15
fix: idleTimeout in websockets  

### LINKED ISSUE #1397: idleTimeout not working in WebSocket

0.2.1

Darwin 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:35 PDT 2022;

When I add the property `idleTimeout` to 8 for example, the connection keeps connected after not sending a message in more than 8 seconds.

_No response_

The server to close the connection after not sending a message in the timeout.

### What do you see instead?

The connection keeps connected.

### Additional information

_No response_

LukasKastern:
### What does this PR do?

Before this change `us_loop_integrate` was invoked on each call to `us_loop_run_bun_tick`. 
Which caused the timer used for socket timeouts to never be invoked.

I changed the code to only integrate the timer if it hasn't been before.

Fixes #1397

- [x] Code changes

The minimum idle timeout is 8 seconds which is a bit long for a unit test. 

I only tried the change locally instead of adding a test.
```
const idle_timeout_seconds = 8;
const allowed_timeout_deviation = 2;

let server = undefined;

let promise = new Promise((resolve, reject) => {
  let timeout = undefined;

  server = Bun.serve({
    port: 0,
    fetch(request, server) {
      const url = new URL(request.url);
      if (url.pathname === "/ws") {
        if (server.upgrade(request)) {
          return;
        }
        return new Response("Upgrade failed :(", { status: 500 });
      }
      return new Response("Bun");
    },
    websocket: {
      idleTimeout: idle_timeout_seconds,
      sendPings: false,
      open() {
        timeout = setTimeout(() => {
          reject.call();
        }, (idle_timeout_seconds + allowed_timeout_deviation) * 1000);
      },
      close() {
        clearTimeout(timeout);
        resolve.call();
      }
    },
  });
});

const ws = new WebSocket(`ws://${server.hostname}:${server.port}/ws`)

await promise
server.stop(true);

console.log("Success")
```

### DIFF
--- a/packages/bun-usockets/src/eventing/epoll_kqueue.c
+++ b/packages/bun-usockets/src/eventing/epoll_kqueue.c
@@ -183,7 +183,13 @@ void us_loop_run_bun_tick(struct us_loop_t *loop, int64_t timeoutMs, void* tickC
     if (loop->num_polls == 0)
         return;
 
-    us_loop_integrate(loop);
+    struct us_internal_callback_t *timer_callback = (struct us_internal_callback_t*)loop->data.sweep_timer;
+
+    // Only integrate the loop if we haven't already.
+    // Otherwise we will keep restarting the timer.
+    if(!timer_callback->cb) {
+        us_loop_integrate(loop);
+    }
 
     if (tickCallbackContext) {
         bun_on_tick_before(tickCallbackContext);


Jarred-Sumner:
This looks great. I'm scared to see what tests will fail but those tests probably should fail

Jarred-Sumner:
Thank you

Lan-Hekary:
Hello @LukasKastern and @Jarred-Sumner,
This PR fixed it indeed for Linux and it's Still Working fine in v1.1.3 on Linux ARM64.
But The Bug is still there in Bun for Windows x64 v1.1.3.
Websocket Server properties sendPings and idleTimeout does not have an effect at all.

PR #8193, opened 2024-01-16
Add support for custom Module._extensions

jdalton:
### What does this PR do?

Add support for custom Module._extensions

- [ ] Code changes

 I added cjs unit tests.

### DIFF
--- a/src/bun.js/bindings/CommonJSModuleRecord.cpp
+++ b/src/bun.js/bindings/CommonJSModuleRecord.cpp
@@ -68,11 +68,13 @@
 #include <JavaScriptCore/HeapAnalyzer.h>
 
 extern "C" bool Bun__isBunMain(JSC::JSGlobalObject* global, const BunString*);
+extern "C" JSC__JSValue Bun__Path__basename(JSC__JSGlobalObject* arg0, bool arg1, JSC__JSValue* arg2, uint16_t arg3);
+extern "C" JSC__JSValue Bun__Path__dirname(JSC__JSGlobalObject* arg0, bool arg1, JSC__JSValue* arg2, uint16_t arg3);
 
 namespace Bun {
 using namespace JSC;
 
-JSC_DECLARE_HOST_FUNCTION(jsFunctionRequireCommonJS);
+JSC_DECLARE_HOST_FUNCTION(jsFunctionRequirePrivate);
 
 static bool canPerformFastEnumeration(Structure* s)
 {
@@ -103,19 +105,23 @@ static bool evaluateCommonJSModuleOnce(JSC::VM& vm, Zig::GlobalObject* globalObj
         return false;
     }
 
+    auto clientData = WebCore::clientData(vm);
+    auto builtinNames = clientData->builtinNames();
+    auto globalString = globalObject->commonStrings().resolveString(globalObject);
+
     JSFunction* resolveFunction = JSC::JSBoundFunction::create(vm,
         globalObject,
         globalObject->requireResolveFunctionUnbound(),
         moduleObject->id(),
-        ArgList(), 1, globalObject->commonStrings().resolveString(globalObject));
+        ArgList(), 1, globalString);
     JSFunction* requireFunction = JSC::JSBoundFunction::create(vm,
         globalObject,
         globalObject->requireFunctionUnbound(),
         moduleObject,
-        ArgList(), 1, globalObject->commonStrings().requireString(globalObject));
+        ArgList(), 1, globalString);
     requireFunction->putDirect(vm, vm.propertyNames->resolve, resolveFunction, 0);
-    moduleObject->putDirect(vm, WebCore::clientData(vm)->builtinNames().requirePublicName(), requireFunction, 0);
 
+    moduleObject->putDirect(vm, builtinNames.requirePublicName(), requireFunction, 0);
     moduleObject->hasEvaluated = true;
 
     // This will return 0 if there was a syntax error or an allocation failure
@@ -148,11 +154,11 @@ static bool evaluateCommonJSModuleOnce(JSC::VM& vm, Zig::GlobalObject* globalObj
     return exception.get() == nullptr;
 }
 
-JSC_DEFINE_HOST_FUNCTION(jsFunctionLoadModule, (JSGlobalObject * lexicalGlobalObject, CallFrame* callframe))
+JSC_DEFINE_HOST_FUNCTION(jsFunctionEvaluateCommonJSModule, (JSGlobalObject * lexicalGlobalObject, CallFrame* callFrame))
 {
     auto* globalObject = jsCast<Zig::GlobalObject*>(lexicalGlobalObject);
     auto throwScope = DECLARE_THROW_SCOPE(globalObject->vm());
-    JSCommonJSModule* moduleObject = jsDynamicCast<JSCommonJSModule*>(callframe->argument(0));
+    JSCommonJSModule* moduleObject = jsDynamicCast<JSCommonJSModule*>(callFrame->argument(0));
     if (!moduleObject) {
         RELEASE_AND_RETURN(throwScope, JSValue::encode(jsBoolean(true)));
     }
@@ -184,18 +190,18 @@ JSC_DEFINE_HOST_FUNCTION(jsFunctionLoadModule, (JSGlobalObject * lexicalGlobalOb
     RELEASE_AND_RETURN(throwScope, JSValue::encode(jsBoolean(true)));
 }
 
-JSC_DEFINE_HOST_FUNCTION(requireResolvePathsFunction, (JSGlobalObject * globalObject, CallFrame* callframe))
+JSC_DEFINE_HOST_FUNCTION(jsFunctionRequireResolvePaths, (JSGlobalObject * globalObject, CallFrame* callFrame))
 {
     return JSValue::encode(JSC::constructEmptyArray(globalObject, nullptr, 0));
 }
 
-JSC_DEFINE_CUSTOM_GETTER(jsRequireCacheGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
+JSC_DEFINE_CUSTOM_GETTER(requireCacheGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
 {
     Zig::GlobalObject* thisObject = jsCast<Zig::GlobalObject*>(globalObject);
     return JSValue::encode(thisObject->lazyRequireCacheObject());
 }
 
-JSC_DEFINE_CUSTOM_SETTER(jsRequireCacheSetter,
+JSC_DEFINE_CUSTOM_SETTER(requireCacheSetter,
     (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue,
         JSC::EncodedJSValue value, JSC::PropertyName propertyName))
 {
@@ -208,11 +214,11 @@ JSC_DEFINE_CUSTOM_SETTER(jsRequireCacheSetter,
 }
 
 static const HashTableValue RequireResolveFunctionPrototypeValues[] = {
-    { "paths"_s, static_cast<unsigned>(JSC::PropertyAttribute::Function), NoIntrinsic, { HashTableValue::NativeFunctionType, requireResolvePathsFunction, 1 } },
+    { "paths"_s, static_cast<unsigned>(JSC::PropertyAttribute::Function), NoIntrinsic, { HashTableValue::NativeFunctionType, jsFunctionRequireResolvePaths, 1 } },
 };
 
 static const HashTableValue RequireFunctionPrototypeValues[] = {
-    { "cache"_s, static_cast<unsigned>(JSC::PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, jsRequireCacheGetter, jsRequireCacheSetter } },
+    { "cache"_s, static_cast<unsigned>(JSC::PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, requireCacheGetter, requireCacheSetter } },
 };
 
 Structure* RequireFunctionPrototype::createStructure(
@@ -236,7 +242,6 @@ Structure* RequireResolveFunctionPrototype::createStructure(
 RequireResolveFunctionPrototype* RequireResolveFunctionPrototype::create(JSC::JSGlobalObject* globalObject)
 {
     auto& vm = globalObject->vm();
-
     auto* structure = RequireResolveFunctionPrototype::createStructure(vm, globalObject);
     RequireResolveFunctionPrototype* prototype = new (NotNull, JSC::allocateCell<RequireResolveFunctionPrototype>(vm)) RequireResolveFunctionPrototype(vm, structure);
     prototype->finishCreation(vm);
@@ -247,12 +252,9 @@ RequireFunctionPrototype* RequireFunctionPrototype::create(
     JSC::JSGlobalObject* globalObject)
 {
     auto& vm = globalObject->vm();
-
     auto* structure = RequireFunctionPrototype::createStructure(vm, globalObject);
     RequireFunctionPrototype* prototype = new (NotNull, JSC::allocateCell<RequireFunctionPrototype>(vm)) RequireFunctionPrototype(vm, structure);
     prototype->finishCreation(vm);
-
-    prototype->putDirect(vm, builtinNames(vm).resolvePublicName(), jsCast<Zig::GlobalObject*>(globalObject)->requireResolveFunctionUnbound(), 0);
 
     return prototype;
 }
@@ -263,26 +265,34 @@ void RequireFunctionPrototype::finishCreation(JSC::VM& vm)
     ASSERT(inherits(info()));
 
     reifyStaticProperties(vm, info(), RequireFunctionPrototypeValues, *this);
+
+    auto clientData = WebCore::clientData(vm);
+    auto* globalObject = reinterpret_cast<Zig::GlobalObject*>(this->globalObject());
+    auto builtinNames = clientData->builtinNames();
+
     JSC::JSFunction* requireDotMainFunction = JSFunction::create(
         vm,
         moduleMainCodeGenerator(vm),
-        globalObject()->globalScope());
+        globalObject->globalScope());
 
     this->putDirectAccessor(
-        globalObject(),
+        globalObject,
         JSC::Identifier::fromString(vm, "main"_s),
-        JSC::GetterSetter::create(vm, globalObject(), requireDotMainFunction, requireDotMainFunction),
+        JSC::GetterSetter::create(vm, globalObject, requireDotMainFunction, requireDotMainFunction),
         PropertyAttribute::Accessor | PropertyAttribute::ReadOnly | 0);
 
-    auto extensions = constructEmptyObject(globalObject());
+    this->putDirect(vm, builtinNames.resolvePublicName(), globalObject->requireResolveFunctionUnbound(), 0);
+
+    auto extensions = constructEmptyObject(vm, globalObject->nullPrototypeObjectStructure());
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".js"_s), jsBoolean(true), 0);
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".json"_s), jsBoolean(true), 0);
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".node"_s), jsBoolean(true), 0);
-
+    extensions->putDirect(vm, builtinNames.originalStructureIDPrivateName(), jsNumber(0), 0);
+    extensions->putDirect(vm, builtinNames.originalStructureIDPrivateName(), jsNumber(extensions->structureID().bits()), 0);
     this->putDirect(vm, JSC::Identifier::fromString(vm, "extensions"_s), extensions, 0);
 }
 
-JSC_DEFINE_CUSTOM_GETTER(getterFilename, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
+JSC_DEFINE_CUSTOM_GETTER(filenameGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
 {
     JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
     if (UNLIKELY(!thisObject)) {
@@ -290,7 +300,20 @@ JSC_DEFINE_CUSTOM_GETTER(getterFilename, (JSC::JSGlobalObject * globalObject, JS
     }
     return JSValue::encode(thisObject->m_filename.get());
 }
-JSC_DEFINE_CUSTOM_GETTER(getterId, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
+
+JSC_DEFINE_CUSTOM_SETTER(filenameSetter,
+    (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue,
+        JSC::EncodedJSValue value, JSC::PropertyName propertyName))
+{
+    JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
+    if (!thisObject)
+        return false;
+
+    thisObject->m_filename.set(globalObject->vm(), thisObject, JSValue::decode(value).toString(globalObject));
+    return true;
+}
+
+JSC_DEFINE_CUSTOM_GETTER(idGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
 {
     JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
     if (UNLIKELY(!thisObject)) {
@@ -299,16 +322,42 @@ JSC_DEFINE_CUSTOM_GETTER(getterId, (JSC::JSGlobalObject * globalObject, JSC::Enc
     return JSValue::encode(thisObject->m_id.get());
 }
 
-JSC_DEFINE_CUSTOM_GETTER(getterPath, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
+JSC_DEFINE_CUSTOM_SETTER(idSetter,
+    (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue,
+        JSC::EncodedJSValue value, JSC::PropertyName propertyName))
+{
+    JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
+    if (!thisObject)
+        return false;
+
+    thisObject->m_id.set(globalObject->vm(), thisObject, JSValue::decode(value).toString(globalObject));
+    return true;
+}
+
+JSC_DEFINE_CUSTOM_GETTER(loadedGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
 {
     JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
     if (UNLIKELY(!thisObject)) {
         return JSValue::encode(jsUndefined());
     }
-    return JSValue::encode(thisObject->m_id.get());
+
+    return JSValue::encode(jsBoolean(thisObject->hasEvaluated));
 }
 
-JSC_DEFINE_CUSTOM_GETTER(getterParent, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
+JSC_DEFINE_CUSTOM_SETTER(loadedSetter,
+    (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue,
+        JSC::EncodedJSValue value, JSC::PropertyName propertyName))
+{
+    JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
+    if (!thisObject)
+        return false;
+
+    thisObject->hasEvaluated = JSValue::decode(value).toBoolean(globalObject);
+
+    return true;
+}
+
+JSC_DEFINE_CUSTOM_GETTER(parentGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
 {
     JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
     if (UNLIKELY(!thisObject)) {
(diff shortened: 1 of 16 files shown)


src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -246,13 +263,10 @@ RequireFunctionPrototype* RequireFunctionPrototype::create(
     JSC::JSGlobalObject* globalObject)
 {
     auto& vm = globalObject->vm();
-
     auto* structure = RequireFunctionPrototype::createStructure(vm, globalObject);
     RequireFunctionPrototype* prototype = new (NotNull, JSC::allocateCell<RequireFunctionPrototype>(vm)) RequireFunctionPrototype(vm, structure);
     prototype->finishCreation(vm);
 
-    prototype->putDirect(vm, builtinNames(vm).resolvePublicName(), jsCast<Zig::GlobalObject*>(globalObject)->requireResolveFunctionUnbound(), 0);
-
```
jdalton, line 256: ☝️ Moved into `prototype->putDirect` into `prototype->finishCreation(vm)`

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -262,34 +276,60 @@ void RequireFunctionPrototype::finishCreation(JSC::VM& vm)
...
+
+    auto clientData = WebCore::clientData(vm);
+    auto globalObject = this->globalObject();
+    auto builtinNames = clientData->builtinNames();
+
     JSC::JSFunction* requireDotMainFunction = JSFunction::create(
         vm,
         moduleMainCodeGenerator(vm),
-        globalObject()->globalScope());
+        globalObject->globalScope());
```
jdalton, line 276: ☝️ Instead of `globalObject()` I referenced `this->globalObject()`. Are they identical?

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -262,34 +276,60 @@ void RequireFunctionPrototype::finishCreation(JSC::VM& vm)
...
+    auto extensions = constructEmptyObject(vm, globalObject->nullPrototypeObjectStructure());
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".js"_s), jsBoolean(true), 0);
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".json"_s), jsBoolean(true), 0);
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".node"_s), jsBoolean(true), 0);
+    extensions->putDirectCustomAccessor(
+        vm,
+        builtinNames.structureChangedPrivateName(),
+        CustomGetterSetter::create(vm, structureChangedGetter, nullptr),
+        PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly | 0);
+    extensions->putDirect(vm, builtinNames.originalStructureIDPrivateName(), jsNumber(extensions->structureID().bits()));
```
jdalton: ☝️ Captured original `extensions->structureID()`

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -298,16 +338,42 @@ JSC_DEFINE_CUSTOM_GETTER(getterId, (JSC::JSGlobalObject * globalObject, JSC::Enc
...
+    JSCommonJSModule* thisObject = jsDynamicCast<JSCommonJSModule*>(JSValue::decode(thisValue));
+    if (!thisObject)
+        return false;
+
+    thisObject->hasEvaluated = JSValue::decode(value).toBoolean(globalObject);
+
+    return true;
+}
+
+JSC_DEFINE_CUSTOM_GETTER(parentGetter, (JSC::JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::PropertyName))
```
jdalton, line 360: ☝️ Naming consistency and colocating thingGetter, thingSetter.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -437,20 +467,20 @@ static JSValue createChildren(VM& vm, JSObject* object)
     return constructEmptyArray(object->globalObject(), nullptr, 0);
 }
 
-JSC_DEFINE_HOST_FUNCTION(functionCommonJSModuleRecord_compile, (JSGlobalObject * globalObject, CallFrame* callframe))
+JSC_DEFINE_HOST_FUNCTION(jsFunctionCommonJSModuleRecord_compile, (JSGlobalObject * globalObject, CallFrame* callFrame))
```
jdalton, line 454: ☝️ Naming consistency.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -494,15 +524,87 @@ JSC_DEFINE_HOST_FUNCTION(functionCommonJSModuleRecord_compile, (JSGlobalObject *
     return JSValue::encode(jsUndefined());
 }
 
+extern "C" JSC::EncodedJSValue jsFunctionResolveSyncPrivate(JSC::JSGlobalObject* lexicalGlobalObject, JSC::CallFrame* callFrame)
```
jdalton, line 515: ☝️ Moved from ImportMetaObject to CommonJSModuleRecord because this is used in more contexts than the ImportMetaObject

src/js/builtins.d.ts
```diff
@@ -407,6 +407,8 @@ declare function $streamErrored(): TODO;
 declare function $streamReadable(): TODO;
 declare function $streamWaiting(): TODO;
 declare function $streamWritable(): TODO;
+declare function $stringIndexOfInternal(searchString: string, position?: number): TODO;
+declare function $stringSubstring(indexStart: number, indexEnd?: number): TODO;
```
paperclover: nice finds. do these return number and string respectively. could add that.

this file was mostly a stub based on what i thought existed. i should clean it up to be exaclty what is available and fill out the types

src/js/builtins/Module.ts
```diff
@@ -34,45 +38,70 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
     return $internalRequire(id);
   }
-
   // To handle import/export cycles, we need to create a module object and put
   // it into the map before we import it.
   const mod = $createCommonJSModule(id, {}, false, this);
   $requireMap.$set(id, mod);
 
+  // @ts-ignore - TS doesn't know each function is isolated.
+  const Module = require("node:module");
```
paperclover: this call is going to make every module load slower. we need to find a better place to do this.

src/bun.js/bindings/CommonJSModuleRecord.cpp
paperclover: identical

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -494,15 +524,87 @@ JSC_DEFINE_HOST_FUNCTION(functionCommonJSModuleRecord_compile, (JSGlobalObject *
...
+                auto bunStr = Bun::toString(parentIdStr);
+
+                MarkedArgumentBuffer args;
+                args.append(moduleName);
+                args.append(parentModuleObject);
+                args.append(jsBoolean(Bun__isBunMain(globalObject, &bunStr)));
+
+                auto ModuleIdStr = jsString(vm, WTF::String("module"_s));
+                auto* ModuleModuleObject = jsCast<Bun::JSCommonJSModule*>(requireMap->get(globalObject, ModuleIdStr));
+                auto ModuleExportsObject = ModuleModuleObject->exportsObject();
```
jdalton: ☝️ Made the the `thisObject` be the `Module` object

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -494,15 +524,87 @@ JSC_DEFINE_HOST_FUNCTION(functionCommonJSModuleRecord_compile, (JSGlobalObject *
...
+                auto parentIdStr = parentID.toWTFString(globalObject);
+                auto bunStr = Bun::toString(parentIdStr);
+
+                MarkedArgumentBuffer args;
+                args.append(moduleName);
+                args.append(parentModuleObject);
+                args.append(jsBoolean(Bun__isBunMain(globalObject, &bunStr)));
+
+                auto ModuleIdStr = jsString(vm, WTF::String("module"_s));
+                auto* ModuleModuleObject = jsCast<Bun::JSCommonJSModule*>(requireMap->get(globalObject, ModuleIdStr));
```
jdalton: Using `jsCast` because the Module should exist.

src/js/builtins/Module.ts
```diff
@@ -34,45 +38,70 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
+  const { _extensions } = Module;
+  if (_extensions && _extensions.$structureChanged) {
```
paperclover: this is going to be very slow. we should not use a custom dynamic getter.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -494,15 +524,87 @@ JSC_DEFINE_HOST_FUNCTION(functionCommonJSModuleRecord_compile, (JSGlobalObject *
...
-    { "paths"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, getterPaths, setterPaths } },
+    { "filename"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, filenameGetter, filenameSetter } },
+    { "id"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, idGetter, idSetter } },
+    { "loaded"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, loadedGetter, loadedSetter } },
+    { "parent"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor | PropertyAttribute::DontEnum), NoIntrinsic, { HashTableValue::GetterSetterType, parentGetter, parentSetter } },
+    { "path"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, pathGetter, pathSetter } },
+    { "paths"_s, static_cast<unsigned>(PropertyAttribute::CustomAccessor), NoIntrinsic, { HashTableValue::GetterSetterType, pathsGetter, pathsSetter } },
```
jdalton, line 594: ☝️ Name juggle

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -836,8 +937,6 @@ void JSCommonJSModule::toSyntheticSource(JSC::JSGlobalObject* globalObject,
     JSC::MarkedArgumentBuffer& exportValues)
 {
     auto result = this->exportsObject();
-
-    auto& vm = globalObject->vm();
     populateESMExports(globalObject, result, exportNames, exportValues, this->ignoreESModuleAnnotation);
```
jdalton, line 929: ☝️ Unused `vm`.

src/js/builtins/Module.ts
```diff
@@ -34,45 +38,70 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
   // Module._compile are actually important for compatibility.
   //
   // Note: we do not need to wrap this in a try/catch, if it throws the C++ code will
   // clear the module from the map.
-  //
-  var out = this.$require(id, mod);
-
-  // -1 means we need to lookup the module from the ESM registry.
-  if (out === -1) {
+  if (this.$require(id, mod) === -1) {
```
paperclover: this is good cleanup

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -597,16 +698,16 @@ JSCommonJSModule* JSCommonJSModule::create(
     return cell;
 }
 
-JSC_DEFINE_HOST_FUNCTION(jsFunctionCreateCommonJSModule, (JSGlobalObject * globalObject, CallFrame* callframe))
+JSC_DEFINE_HOST_FUNCTION(jsFunctionCreateCommonJSModule, (JSGlobalObject * globalObject, CallFrame* callFrame))
```
jdalton, line 688: `callframe` -> `callFrame` arg name consistency

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -889,23 +988,23 @@ const JSC::ClassInfo JSCommonJSModule::s_info = { "Module"_s, &Base::s_info, nul
 const JSC::ClassInfo RequireResolveFunctionPrototype::s_info = { "resolve"_s, &Base::s_info, nullptr, nullptr, CREATE_METHOD_TABLE(RequireResolveFunctionPrototype) };
 const JSC::ClassInfo RequireFunctionPrototype::s_info = { "require"_s, &Base::s_info, nullptr, nullptr, CREATE_METHOD_TABLE(RequireFunctionPrototype) };
 
-JSC_DEFINE_HOST_FUNCTION(jsFunctionRequireCommonJS, (JSGlobalObject * lexicalGlobalObject, CallFrame* callframe))
+JSC_DEFINE_HOST_FUNCTION(jsFunctionRequireCommonJS, (JSGlobalObject * lexicalGlobalObject, CallFrame* callFrame))
```
jdalton: `callframe` -> `callFrame` arg name consistency

paperclover (dismissed):
im worried this is going to make CommonJS a bajillion times slower.

src/bun.js/bindings/CommonJSModuleRecord.h
```diff
@@ -106,12 +108,6 @@ class JSCommonJSModule final : public JSC::JSDestructibleObject {
     }
 };
 
-JSCommonJSModule* createCommonJSModuleWithoutRunning(
-    Zig::GlobalObject* globalObject,
-    Ref<Zig::SourceProvider> sourceProvider,
-    const WTF::String& sourceURL,
-    ResolvedSource source);
-
 JSC::Structure* createCommonJSModuleStructure(
```
jdalton, line 110: ☝️ Unused thing

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3695,21 +3695,19 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
...
-    putDirectNativeFunction(vm, this, builtinNames.resolveSyncPrivateName(), 1, functionImportMeta__resolveSyncPrivate, ImplementationVisibility::Public, NoIntrinsic, PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly);
-    putDirectNativeFunction(vm, this, builtinNames.createInternalModuleByIdPrivateName(), 1, InternalModuleRegistry::jsCreateInternalModuleById, ImplementationVisibility::Public, NoIntrinsic, PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly);
+    putDirectBuiltinFunction(vm, this, builtinNames.requireESMPrivateName(), importMetaObjectRequireESMCodeGenerator(vm), PropertyAttribute::Builtin | PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly);
+    putDirectBuiltinFunction(vm, this, builtinNames.requireNativeModulePrivateName(), moduleRequireNativeModuleCodeGenerator(vm), PropertyAttribute::Builtin | PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly);
+    putDirectNativeFunction(vm, this, builtinNames.resolveSyncPrivateName(), 1, jsFunctionResolveSyncPrivate, ImplementationVisibility::Public, NoIntrinsic, PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly);
```
jdalton, line 3679: ☝️ Order nits for easy locating

src/codegen/replacements.ts
```diff
@@ -41,17 +42,15 @@ export const globalsToPrefix = [
   "ReadableStreamBYOBRequest",
   "ReadableStreamDefaultController",
   "ReadableStreamDefaultReader",
+  "RegExp",
+  "String",
   "TransformStream",
   "TransformStreamDefaultController",
   "Uint8Array",
-  "String",
-  "Buffer",
```
jdalton, line 48: Alphanumeric ordered and removal of duplicate "Buffer" entry.

src/js/builtins.d.ts
jdalton: ☝️ Using more WebKit primordial equivs.

src/js/builtins/Module.ts
```diff
@@ -12,8 +12,12 @@ export function require(this: CommonJSModuleRecord, id: string) {
...
 $visibility = "Private";
 export function overridableRequire(this: CommonJSModuleRecord, id: string) {
-  const existing = $requireMap.$get(id) || $requireMap.$get((id = $resolveSync(id, this.path, false)));
-  if (existing) {
+  let existing = $requireMap.$get(id);
+  if (existing === undefined) {
+    id = $resolveSync(id, this.path, false);
+    existing = $requireMap.$get(id);
+  }
+  if (existing && !existing.loaded) {
```
jdalton: ☝️ Checking if `existing.loaded` is not true before continuing down this path

src/js/builtins/Module.ts
```diff
@@ -34,45 +38,70 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
   // clear the module from the map.
-  //
-  var out = this.$require(id, mod);
-
-  // -1 means we need to lookup the module from the ESM registry.
-  if (out === -1) {
+  if (this.$require(id, mod) === -1) {
+    // -1 means we need to lookup the module from the ESM registry.
     try {
-      out = $requireESM(id);
```
jdalton, line 66: Removed unused variable assignment to `out`.

src/js/builtins/Module.ts
```diff
@@ -34,45 +38,70 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
-    if (esm?.evaluated && (esm.state ?? 0) >= $ModuleReady) {
+    if (esm && esm.evaluated && (esm.state ?? 0) >= $ModuleReady) {
       const namespace = Loader.getModuleNamespaceObject(esm!.module);
-      return (mod.exports =
-        // if they choose a module
-        namespace.__esModule ? namespace : Object.create(namespace, { __esModule: { value: true } }));
+      const exports = namespace.__esModule 
+        // If they choose a module.
+        ? namespace 
+        : $Object.$create(namespace, { __esModule: { value: true } });
```
jdalton, line 81: ☝️ used JSC `$Object.$create`

src/js/builtins/Module.ts
```diff
@@ -85,7 +114,7 @@ export function requireResolve(this: string | { path: string }, id: string) {
 $visibility = "Private";
 export function requireNativeModule(id: string) {
   let esm = Loader.registry.$get(id);
-  if (esm?.evaluated && (esm.state ?? 0) >= $ModuleReady) {
+  if (esm && esm.evaluated && (esm.state ?? 0) >= $ModuleReady) {
```
jdalton, line 98: ☝️ Simplified conditional but it's mostly habit.

test/js/node/module/resolveFilenameOverwrite.cjs
```diff
@@ -1,15 +1,26 @@
...
+    expect(request.endsWith("💔")).toBe(true);
+    assertions++;
+    expect(parent.filename).toBe(path.join(__dirname, "./resolveFilenameOverwrite.cjs"));
+    assertions++;
+    expect(isMain).toBe(true);
+    assertions++;
+    expect(this).toBe(Module);
+    assertions++;
+    return path.join(__dirname, "./resolveFilenameOverwrite-fixture.cjs");
+  };
```
jdalton, line 19: ☝️ Added assertion count check, `isMain` check, and `thisObject` check.

test/js/node/module/resolveFilenameOverwrite.cjs
```diff
@@ -1,15 +1,26 @@
...
-Module._resolveFilename = (specifier, parent, isMain) => {
-  eql(specifier.endsWith("💔"), true);
-  eql(parent.filename, path.join(__dirname, "./resolveFilenameOverwrite.cjs"));
-  return path.join(__dirname, "./resolveFilenameOverwrite-fixture.cjs");
-};
-eql(require("overwriting _resolveFilename broke 💔"), "winner");
-Module._resolveFilename = original;
+const { expect, test } = require("bun:test");
+const Module = require("node:module");
+const path = require("node:path");
```
jdalton, line 4: Use `bun:test`.

test/js/node/module/modulePrototypeOverwrite.cjs
```diff
@@ -1,17 +1,19 @@
-// This behavior is required for Next.js to work
-const eql = require("assert").deepStrictEqual;
-const Module = require("module");
+const { expect, test } = require("bun:test");
+const Module = require("node:module");
```
jdalton, line 2: Use `bun:test`.

test/js/node/module/moduleExtensionsChange.cjs
```diff
@@ -0,0 +1,22 @@
...
+
+  const changedResult = require("./moduleExtensionsChange-fixture.cjs");
+  expect(changedResult).toBe("winner");
+
+  delete Module._cache[require.resolve("./moduleExtensionsChange-fixture.cjs")];
+  Module._extensions['.js'] = true;
+
+  const revertedResult = require("./moduleExtensionsChange-fixture.cjs");
+  expect(revertedResult).toBe("original");
+});
```
jdalton, line 51: ☝️ Test for `Module._extensions` change and revert.

src/js/builtins/Module.ts
jdalton: > this call is going to make every module load slower. we need to find a better place to do this.

My understanding is that `const Module = require("node:module")` is hoisted (or cached in some way through magic transformation) 🤞 

~~How does this magic work?~~

Updated with a local cache to avoid the internal builtin module load.

src/js/builtins/Module.ts
jdalton: > this is going to be very slow. we should not use a custom dynamic getter.

Is this more expensive than say any other JS related thing like `propertyIsEnumerable`cor `String#slice`? If it's comparable it may be okay.

src/js/builtins.d.ts
jdalton: > nice finds. do these return number and string respectively. could add that.

yes :) they return identical to the builtins. I will update their definitions.

src/js/builtins/Module.ts
jdalton: Measuring with `require('@babel/core')` showed no noticeable change in variance in the debug build (though I didn't stress test it). @paperdave Is there a scenario you're interested in?

src/js/builtins/Module.ts
jdalton: Perf concern covered https://github.com/oven-sh/bun/pull/8193#discussion_r1452960872.

src/js/builtins/Module.ts
jdalton: @paperdave Is there a scenario you're interested in?

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -91,80 +91,6 @@ namespace Zig {
 using namespace JSC;
 using namespace WebCore;
 
-static JSC::EncodedJSValue functionRequireResolve(JSC::JSGlobalObject* globalObject, JSC::CallFrame* callFrame, const WTF::String& fromStr)
-{
```
jdalton, line 95: ☝️ Unused method

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -183,33 +109,20 @@ Zig::ImportMetaObject* Zig::ImportMetaObject::create(JSC::JSGlobalObject* global
...
-    JSValue thisValue = callFrame->thisValue();
-    WTF::String fromStr;
-
-    if (thisValue.isString()) {
-        fromStr = thisValue.toWTFString(globalObject);
-    }
-
-    return functionRequireResolve(globalObject, callFrame, fromStr);
-}
```
jdalton, line 197: ☝️ Unused method

src/bun.js/bindings/ImportMetaObject.cpp
```diff
@@ -218,198 +131,132 @@ extern "C" JSC::EncodedJSValue functionImportMeta__resolveSync(JSC::JSGlobalObje
...
-    }
-
-    scope.release();
-    return result;
-}
-
-extern "C" bool Bun__isBunMain(JSC::JSGlobalObject* global, const BunString*);
-
-extern "C" JSC::EncodedJSValue functionImportMeta__resolveSyncPrivate(JSC::JSGlobalObject* lexicalGlobalObject, JSC::CallFrame* callFrame)
-{
```
jdalton, line 294: ☝️ Moved to CommonJSModuleRecord.

src/bun.js/bindings/ImportMetaObject.h
```diff
@@ -24,9 +21,6 @@ class ImportMetaObject final : public JSC::JSNonFinalObject {
     using Base = JSC::JSNonFinalObject;
 
     static ImportMetaObject* create(JSC::VM& vm, JSC::JSGlobalObject* globalObject, JSC::Structure* structure, const WTF::String& url);
-
-    static JSObject* createRequireFunction(VM& vm, JSGlobalObject* lexicalGlobalObject, const WTF::String& pathString);
-
```
jdalton, line 29: ☝️ Didn't exist.

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3825,13 +3823,13 @@ void GlobalObject::visitChildrenImpl(JSCell* cell, Visitor& visitor)
     thisObject->m_builtinInternalFunctions.visit(visitor);
     thisObject->m_commonStrings.visit<Visitor>(visitor);
     visitor.append(thisObject->m_assignToStream);
+    visitor.append(thisObject->m_nodeModuleOverriddenResolveFilename);
```
jdalton, line 3794: ☝️ If the structure trick is cheap perf wise it may be a better approach than 1 time trigger.

src/js/builtins/Module.ts
```diff
@@ -34,53 +43,83 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
     $argumentCount(),
-    // the object containing a "type" attribute, if they passed one
+    // The object containing a "type" attribute, if they passed one
     // maybe this will be "paths" in the future too.
-    arguments[1],
-  );
-
-  // -1 means we need to lookup the module from the ESM registry.
-  if (out === -1) {
+    type,
```
jdalton: ☝️ Uses a param instead of materializing the `arguments` object. To better take advantage of the `@argumentCount` use.

src/js/builtins/Module.ts
```diff
@@ -34,53 +47,84 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
...
   $requireMap.$set(id, mod);
 
+  let { Module } = cache;
+  if (Module === undefined) {
+    // @ts-ignore - TS doesn't know each function is isolated.
+    Module = require("node:module");
+    cache.Module = Module;
+  }
+  const { _extensions } = Module;
+  if (_extensions && _extensions.$structureChanged) {
```
Jarred-Sumner: Instead of adding a private property, could we do this in `this.$require`?

We could add the StructureID as a field on the Zig::GlobalObject at the time of creation (with some way to mark it as unknown because it was never in use)

Then we can avoid `cache` + these property lookups in JS + pulling in `basename`

For figuring out the file extension from C++, we can do `.findLast('.')` I think on the `WTF::WTFString`

jdalton: Yap. Doable!

src/js/builtins/Module.ts
jdalton: Moving to c++ helper.

src/js/builtins/Module.ts
jdalton: I tried setting a global value based on the structure of the extensions object but it always being different than the real one. Not sure why creating the object in an initLater and getting its structure was any different than within the require function creation. Setting it on the real extensions object though does seem to work.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -263,34 +264,55 @@ void RequireFunctionPrototype::finishCreation(JSC::VM& vm)
...
-    auto extensions = constructEmptyObject(globalObject());
+    this->putDirect(vm, builtinNames.resolvePublicName(), globalObject->requireResolveFunctionUnbound(), 0);
+
+    auto extensions = constructEmptyObject(vm, globalObject->nullPrototypeObjectStructure());
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".js"_s), jsBoolean(true), 0);
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".json"_s), jsBoolean(true), 0);
     extensions->putDirect(vm, JSC::Identifier::fromString(vm, ".node"_s), jsBoolean(true), 0);
-
+    extensions->putDirect(vm, builtinNames.originalStructureIDPrivateName(), jsNumber(0), 0);
+    extensions->putDirect(vm, builtinNames.originalStructureIDPrivateName(), jsNumber(extensions->structureID().bits()), 0);
```
jdalton, line 291: ☝️ Secret sauce the double put. Once to get the shape (structure transition), the other to set the value.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -890,23 +972,85 @@ const JSC::ClassInfo JSCommonJSModule::s_info = { "Module"_s, &Base::s_info, nul
...
+                WTF::Vector<JSC::EncodedJSValue, 1> basenameArgs;
+                basenameArgs.reserveInitialCapacity(1);
+                basenameArgs.unsafeAppendWithoutCapacityCheck(JSC::JSValue::encode(specifierValue));
+
+#if OS(WINDOWS)
+                auto name = JSValue::decode(Bun__Path__basename(globalObject, true, reinterpret_cast<JSC__JSValue*>(basenameArgs.data()), 1)).toString(globalObject)->getString(globalObject);
+#else
+                auto name = JSValue::decode(Bun__Path__basename(globalObject, false, reinterpret_cast<JSC__JSValue*>(basenameArgs.data()), 1)).toString(globalObject)->getString(globalObject);
+#endif
+                size_t index = 0;
```
jdalton, line 1016: ☝️ Calling a zig function extern woo 🎉

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -890,23 +972,83 @@ const JSC::ClassInfo JSCommonJSModule::s_info = { "Module"_s, &Base::s_info, nul
...
+                }
+                if (UNLIKELY(extHandler)) {
+                    JSC::CallData callData = JSC::getCallData(extHandler);
+                    MarkedArgumentBuffer args;
+                    args.append(moduleObject); // module
+                    args.append(specifierValue); // id
+                    // Call Module._extensions[ext](module, id)
+                    JSC::call(globalObject, extHandler, callData, extensionsObject, args);
+                    return JSValue::encode(moduleObject->exportsObject());
+                }
```
jdalton, line 1048: ☝️ Converted the Node JS logic into C++, very similar.

src/bun.js/modules/NodeModuleModule.h
```diff
@@ -112,18 +114,29 @@ JSC_DEFINE_HOST_FUNCTION(jsFunctionNodeModuleModuleConstructor,
...
 
   auto scope = DECLARE_THROW_SCOPE(vm);
+
   if (idValue.isString()) {
-    idString = idValue.toString(globalObject);
+    id = idValue.toString(globalObject);
     RETURN_IF_EXCEPTION(scope, JSC::JSValue::encode(JSC::jsUndefined()));
-
-    auto index = idString->tryGetValue().reverseFind('/', idString->length());
-
```
jdalton, line 120: ☝️ Assuming forward slash is not great... Windows supports BOTH forward and backslashes (https://github.com/nodejs/node/blob/eb4432c12c673b91190b6f921ee3ca864f442386/lib/path.js#L53). By using our zig method implementations (ex: [here](https://github.com/oven-sh/bun/pull/8193/files#diff-cbb1481fc1d6d8dde6440ac67b658aea7de72d55bda67909090f058aacd9214dR126)) we can rely on [zig's standard lib](https://github.com/ziglang/zig/blob/master/lib/std/fs/path.zig#L968-L985) to handle this.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -438,52 +451,56 @@ static JSValue createChildren(VM& vm, JSObject* object)
...
 
+    auto filename = JSC::jsStringWithCache(vm, filenameWTF);
+    WTF::Vector<JSC::EncodedJSValue, 1> dirnameArgs;
+    dirnameArgs.reserveInitialCapacity(1);
+    dirnameArgs.unsafeAppendWithoutCapacityCheck(JSValue::encode(filenameValue));
+#if OS(WINDOWS)
+    auto dirname = JSValue::decode(Bun__Path__dirname(globalObject, true, reinterpret_cast<JSC__JSValue*>(dirnameArgs.data()), 1)).toString(globalObject);
+#else
+    auto dirname = JSValue::decode(Bun__Path__dirname(globalObject, false, reinterpret_cast<JSC__JSValue*>(dirnameArgs.data()), 1)).toString(globalObject);
+#endif
```
jdalton, line 480: ☝️ zig dirname for better Windows path support.

src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -598,39 +685,41 @@ JSCommonJSModule* JSCommonJSModule::create(
...
-        dirname = JSC::jsSubstring(globalObject, requireMapKey, 0, index);
-    }
+    WTF::Vector<JSC::EncodedJSValue, 1> dirnameArgs;
+    dirnameArgs.reserveInitialCapacity(1);
+    dirnameArgs.unsafeAppendWithoutCapacityCheck(JSValue::encode(id));
+#if OS(WINDOWS)
+    auto dirname = JSValue::decode(Bun__Path__dirname(globalObject, true, reinterpret_cast<JSC__JSValue*>(dirnameArgs.data()), 1)).toString(globalObject);
+#else
+    auto dirname = JSValue::decode(Bun__Path__dirname(globalObject, false, reinterpret_cast<JSC__JSValue*>(dirnameArgs.data()), 1)).toString(globalObject);
+#endif
```
jdalton, line 718: ☝️ zig dirname for better Windows path support.

PR #8194, opened 2024-01-16
windows: make it a static binary and ship debug symbols

paperclover:
does not work yet. i also converted the CMakeLists file to LF so sorry for the full-file merge conflict.

here is where im stuck:

image

If i include ucrt, i get more linking errors. I assume I am compiling LibUV wrong.


vildantursic:
I have stupid question 😄 ... temp solution ... would it be possible to add required vc_redist (e.g. `vc_runtime140.dll` and `msvcp140.dll`) files in public dir and load them in memory and that way avoid issues on windows when it require those files?
I tried it but idk how to "load" them during runtime 🤷🏻

PR #8195, opened 2024-01-16
refactor(types): Improve `bun-types`

Didas-git:
This PR aims to improve `bun-types` both for users and maintainers.

## The problem

Since 1.0.19 a lot of issues related to the types have emerged and some people are stuck with older versions of the types for this reason, this issues started to appear because of the huge changes and usage of `@types/node` under the hood.

Some of the changes made include a lot of duplication and unecessary hacks for the sake of declaration merging, this is a problem for maintainers because changing something you need to do it twice and for users this causes issues because the hacks that were applied do not work in all typescript versions.

### Declaration merging

To fix the declaration merging issues the types would require massive changes, while this changes **would not** break current user code, they would be a massive PR to change all the hacks to proper declaration merging and following the new ts standards, i am willing and want to work on this but i would need green light from the maitainers since huge prs are hard to review.

### Stricter/better types

Something that is requested quite a lot by more advanced typescript users is more strict and improved types.
While there was an attempt at this, it was refused due to breaking user code, with that in mind, by using the opportunity of making the changes above listed, we could also introduce `bun-types/strict`.
This subpackage would allow us to add **opt in** stricter types without massive changes, for the initial idea all it would be is 1 file with less than 70 lines just for the main changes people want.

### Not bundling (Impl)

This is how im opening this pr because its the smallest and simplest change.

While people might think that bundling types is the way to go because there is less files there is something important we must keep in mind.
We are not sending type definitions to browsers.

Type definitions are loaded by tss (typescript server) and stay in memory while needed/referenced by the project, the problem with bundling is that typescript is required to load everything even if you just need simple types.

### Chunking

Another advantage of not bundling is that we can break our type definitions into smaller chunks, doing this would make maintaining the types way easier specially when it comes to the `globals` file.

## Note

I want to know the maintainers opinion before moving forward with this pr, i dont want to spend time on a pr of this size for it to just get red lighted afterwards.


Electroid:
Hi @Didas-git, thanks for working on this and writing up your thoughts!

> Declaration merging

I think this is a good idea. Right now there is a lot of weirdness/brokenness around types we add additional properties to, for instance, `WebSocket` and `Headers`. You have the green light.

> Stricter/better types

This is a cool idea. I think we should do it, but maybe after we get declaration merging, merged?

> Not bundling

This seems fine. I'll merge these changes into the work I'm doing in: #8189

Didas-git:
Thanks for answering, i will start working on this asap.
I will leave this draft for all the declaration merging fixes and afterwards open a new one to discuss implementing stricter types in more details.

As for removing bundling.
I decided to add a `build` script so you dont have to manually change the version everytime, but it could be removed in favor of just manually bumping it like we do for bun, i will leave that for you guys to decide.

TiBianMod:
> ### Chunking
> Another advantage of not bundling is that we can break our type definitions into smaller chunks, doing this would make maintaining the types way easier specially when it comes to the `globals` file.

now `bun-types` exports a single `d.ts` file `types.d.ts`, this approach makes my IDE `WebStorm` very laggy.

Please keep in mind, if we can do something about this issue.

> I will suggest something like this, but you know better.

```ts
/// <reference types="bun-types/sqlite.d.ts" />
/// <reference types="bun-types/tty.d.ts" />
/// <reference types="bun-types/bun.d.ts" />
/// <reference types="bun-types/assert.d.ts" />

// ...etc
```

Didas-git:
> > ### Chunking
> > Another advantage of not bundling is that we can break our type definitions into smaller chunks, doing this would make maintaining the types way easier specially when it comes to the `globals` file.
> 
> now `bun-types` exports a single `d.ts` file `types.d.ts`, this approach makes my IDE `WebStorm` very laggy.
> 
> Please keep in mind, if we can do something about this issue.
> 
> > I will suggest something like this, but you know better.
> 
> ```ts
> /// <reference types="bun-types/sqlite.d.ts" />
> /// <reference types="bun-types/tty.d.ts" />
> /// <reference types="bun-types/bun.d.ts" />
> /// <reference types="bun-types/assert.d.ts" />
> 
> // ...etc
> ```

This is part of `bundling` which is already removed in this pr, chunking would only affect the `globals` file because the others are already in chunks.

packages/bun-types/package.json
```diff
@@ -1,25 +1,38 @@
 {
   "name": "bun-types",
-  "repository": "https://github.com/oven-sh/bun",
   "license": "MIT",
+  "main": "",
```
paperclover, line 4: is main being "" intentional?

paperclover (commented):
re: bundling

i was skeptical because i havent seen any proof of this being more performant, but if it makes webstorm not "very laggy" then that is great.

assuming ci and everything still function with this, i think we can ship just the bundling bit for canary and see how it goes.

i do want someone to verify that this package, if published, would properly install and work in ts language server. i know you're a huge ts nerd so i doubt that mistake will be here, but it is important to triple check this.

packages/bun-types/package.json
Didas-git: Yes, this is to show an error to people trying to import the package in js instead of just using it for the types

Didas-git:
> i do want someone to verify that this package, if published, would properly install and work in ts language server. i know you're a huge ts nerd so i doubt that mistake will be here, but it is important to triple check this.

I will do a lot of testing on my end before marking the PR as ready for review, but also having other people test it is good since there might be some version related bugs since som older typescript versions are not as nice with declaration merging

Didas-git:
@Electroid there are some things i would like to discuss.
- There is a type that is exported called `AnyFunction` would you agree on just removing this alias and typing `(...args: any[]) => any` in the places that use (there arent many) it would help the api be more explicit and would have less useless exports.
- Regarding `bun:ffi` there is a type called `Pointer`, im not sure it is what you guys want.

`type Pointer = number & {}` just means it is a non nullable number which in this case is the same as just `number`
This means this is assignable to anything that accepts a number and any number can be assigned to this type.

What i think you guys want instead is:
`type Pointer = number & { [unique symbol]: "pointer" }` or even `type Pointer = number & { pointer: null }`
With the above examples or any iteration of them you can constrain the functions to only get exactly this type, which means the user needs to get the number from a function or cast to this type explicitly.

paperclover:
> `& { [unique symbol]: "pointer" }`

I like this.

I think we should remove `AnyFunction`

Didas-git:
@Electroid @paperdave Regarding `__filename` and `__dirname`.
Do we really need to redefine them **and** mark them as deprecated?
Typescript already handles the types for them depending if you are using `cjs` or `mjs` (and the settings that make those changes as well)

Electroid:
> @Electroid @paperdave Regarding `__filename` and `__dirname`. Do we really need to redefine them **and** mark them as deprecated? Typescript already handles the types for them depending if you are using `cjs` or `mjs` (and the settings that make those changes as well)

In Bun, those work in both mjs and cjs, so it may make sense to keep the overrides.

Didas-git:
> In Bun, those work in both mjs and cjs, so it may make sense to keep the overrides.

I am aware, but, usually you want people to use the newer alternatives that will also work cross runtimes, and since it has been marked on the types as deprecated for a while i was wondering if its rly needed.

Electroid:
> > In Bun, those work in both mjs and cjs, so it may make sense to keep the overrides.
> 
> I am aware, but, usually you want people to use the newer alternatives that will also work cross runtimes, and since it has been marked on the types as deprecated for a while i was wondering if its rly needed.

Fair enough, I think we can remove it then.

Didas-git:
About the proposed chunking idea:

I probably will scrap that idea, i though it could be separated in chunks due to the amount of declarations in there but turns out it just has like 30 of the same declaration over and over instead of doing it all in one, so instead of chuncking i will be merging them all and fixing all the duplication and incorrectly exposed types (the ones prefixed with `_`)

Didas-git:
@Electroid @paperdave I think the PR is ready to start reviewing.
Its worth noting that the type tests do not work, and its not because of my changes they are broken in `main` and all the other branches as well.

packages/bun-types/wasm.d.ts
```diff
@@ -0,0 +1,291 @@
+export {};
```
Electroid, line 1: Is all of this new? Or did we already have a version of this?

packages/bun-types/test.d.ts
```diff
@@ -13,20 +13,14 @@
...
-  export interface Mock<T extends AnyFunction>
-    extends JestMock.MockInstance<T> {
-    (...args: Parameters<T>): ReturnType<T>;
-  }
-  type _Mock<T extends AnyFunction> = Mock<T>;
+  export type Mock<T extends (...args: any[]) => any> = JestMock.Mock<T>;
 
   export const mock: {
-    <T extends AnyFunction>(Function: T): Mock<T>;
+    <T extends (...args: any[]) => any>(Function: T): Mock<T>;
```
Electroid, line 23: This comment is in regards to all usage of `(...args: any[])`. Can we use `(...args: unknown[])` instead? Especially if we are not returning the types anywhere.

packages/bun-types/wasm.d.ts
Didas-git: This was just extracted from `globals.d.ts` into its own file, even tho this file still declares globals.
Perhaps naming it `global.wasm.d.ts` would be more descriptive.

packages/bun-types/test.d.ts
Didas-git: We cannot in the node apis and globals, since `bun:test` is a bun api we could change it, but, it might affect some specific users already using it.

packages/bun-types/bun.d.ts
```diff
@@ -2304,7 +2306,7 @@ declare module "bun" {
     unix: string;
     tls?: TLSOptions;
   }
-  interface Errorlike extends Error {
+  interface ErrorLike extends Error {
```
Electroid, line 2309: I agree with this rename, but let's make sure we maintain an alias somewhere that is deprecated.
This goes for anything we remove, like `AnyFunction`. We should not fully remove.

Can we add a `deprecated.d.ts` file?

Electroid: Basically, we do not want people to upgrade, then have to make big changes just for a rename. It should just work.

Didas-git: Oh right, that was just changed because of cSpell marking it as an error... we can yes add an alias and mark it as deprecated for now.

Aboud adding a file for deprecated features... it might yes be a good idea since i have seen some other stuff also marked as deprecated

packages/bun-types/package.json
```diff
@@ -1,25 +1,37 @@
...
+  "files": [
+    "*.d.ts"
+  ],
+  "homepage": "https://bun.sh",
   "dependencies": {
-    "@types/node": "*",
-    "@types/ws": "*",
-    "undici-types": "~5.26.4"
+    "@types/node": "~20.11.3",
+    "@types/ws": "~8.5.10"
```
Electroid, line 18: I believe the rationale behind not pinning a version of `node`/`ws` was to reduce the dependency conflicts that people have. I'm inclined to think we should keep it that way.

packages/bun-types/test.d.ts
Electroid: Makes sense, if that's the case we can just keep `any` as-is.

packages/bun-types/wasm.d.ts
Electroid: The current name is fine, just checking.

packages/bun-types/package.json
Didas-git: My explanation for this is on #8189

Electroid: You may have mistyped, mind linking the comment?

Didas-git: I did, but i will quote it here again

>The dependencies should not be *, major version changes can cause types to break so we should stick with the latest rn while accepting minor and patch versions.
Major version updates should be done manually to ensure types dont break.

packages/bun-types/package.json
```diff
@@ -1,25 +1,37 @@
 {
   "name": "bun-types",
-  "repository": "https://github.com/oven-sh/bun",
```
jdalton, line 3: Does repo need to go?

packages/bun-types/package.json
```diff
@@ -1,25 +1,37 @@
...
+    "@types/node": "~20.11.3",
+    "@types/ws": "~8.5.10"
   },
   "devDependencies": {
     "@definitelytyped/dtslint": "^0.0.199",
     "@definitelytyped/eslint-plugin": "^0.0.197",
     "prettier": "^2.4.1",
     "typescript": "^5.0.2"
   },
-  "private": true,
```
jdalton, line 16: ☝️ Exposing this (just confirming).

packages/bun-types/package.json
Didas-git: Its not getting deleted, it just got moved down a bit and its using the object form of it for better redirects in npm

packages/bun-types/package.json
Didas-git: Yes, because this is the package.json that will be published to npm.
Before with bundling the script used to generate a new package.json on the dist folder, now we will just use the one already in the repository.

Electroid (approved):
Thanks so much for working on this, much better than the status-quo.

Electroid:
@Didas-git Looks like CI for types is failing due to TS issues, mind taking a look?

packages/bun-types/package.json
```diff
@@ -1,25 +1,37 @@
 {
   "name": "bun-types",
```
jdalton, line 2: @Electroid @paperdave et all. If we are publishing this can we use the `@bun` scoped package? I notice we own the package name `bun` which I believe allows us to create the scope too. This prevents a bunch of `bun-` stuff.

Electroid: @types/bun essentially aliases to bun-types, so most people won’t actually install it directly.

We could change it, but it’s already so widely used

Didas-git:
> @Didas-git Looks like CI for types is failing due to TS issues, mind taking a look?

They have been broken for a long time apparently.
I commented about this some [days ago](https://github.com/oven-sh/bun/pull/8195#issuecomment-1902609213): 
>Its worth noting that the type tests do not work, and its not because of my changes they are broken in main and all the other branches as well.

packages/bun-types/package.json
Didas-git: `bun-types` was the default for a long time, while i do think changing to `@bun/types` could be a good idea we would need to have a deprecation notice on `bun-types` and update `@types/bun`

jdalton: > @types/bun essentially aliases to bun-types, so most people won’t actually install it directly.
> 
> We could change it, but it’s already so widely used

Ahh okay. The legacy way aliases to `@types/bun`. 
Thanks for the added insight @Electroid and @Didas-git !

Electroid (approved):

Electroid:
We'll have to fix our CI later, thanks for working on this. Big improvement.

Issue #8102, opened 2024-01-11
A minor internal bug in the node:fs ReadStream#constructor


Edgar-P-yan:
### What version of Bun is running?

main branch

### What platform is your computer?

all

### What steps can reproduce the bug?

Found this when was reading through the sources, looks like a bug.
Pass an encoding like `'utf-8'` instead of a parameters object as the second argument to `fs.createReadStream('file/path', 'utf-8')`, by design it is allowed. This will get passed down to the `ReadStream` class.
At this line:
https://github.com/oven-sh/bun/blob/0bfe3760f6dd542603aab239bb31034171bb7191/src/js/node/fs.js#L600
the `options` gets deconstructed into an object when calling `super`:
```js
super(ptr, {
  ...options,
  encoding,
  autoDestroy,
  autoClose,
  emitClose,
  highWaterMark,
});
```

In case if `options` is a string representing the encoding (which is by the protocol) the resulting object will look something like this:
```js
{
  0: 'u',
  1: 't',
  2: 'f',
  3: '-',
  4: '8',
  encoding: 'utf-8',
  autoDestroy: ...
  ...
}
```
I thing this does not break anything, but still clutters internal object shapes.

Jarred-Sumner:
Fixed by @keepdying in #8212

PR #8212, opened 2024-01-16
fix(node:fs): handle passing string options in readstream

yemreck:
### What does this PR do?

A small PR to fix string deconstruction seen in #8102 by adding an one liner type check

- [X] Code changes


Jarred-Sumner:
Thank you. 

We should do a similar thing for this line - `typeof options !== "string" ? {} : options`

https://github.com/oven-sh/bun/pull/8212/files#diff-f8c2a15789dde99a70fe05cdf728bdddd5c395aea42f3635133c5796412ebf62R544

Issue #8153, opened 2024-01-13
Bun hangs if console.log is called while inspecting a Set or Map


chocolateboy:
### What version of Bun is running?

1.0.22

### What platform is your computer?

Linux x86_64 (Arch)

### What steps can reproduce the bug?

```javascript
class MyMap extends Map {
    get size () {
        console.log('inside size')
        return 0
    }
}

const map = new MyMap()

console.log('size:', map.size) // OK
console.log('map:', map)       // XXX hangs
```

### What is the expected behavior?

```
inside size
size: 0
inside size
map: Map {}
```

### What do you see instead?

```
inside size
size: 0
// hangs
```

### Additional information

Same issue with Set.

Same issue if `console.debug` or `console.dir` are called, but not `console.warn`, `console.error`, or `process.stdout.write`.

No issue if `console.warn` is used to dump the Map/Set:

```javascript
const map = new MyMap()

console.log('size:', map.size) // OK
console.warn('map:', map)      // OK
```

Jarred-Sumner:
I believe this is caused by a mutex

paperclover:
you can cause this by logging inside of a custom inspect function too

PR #8213, opened 2024-01-16
fix: Hang when recursively logging via ConsoleObject.messageWithTypeAndLevel 

LukasKastern:
### What does this PR do?

Avoid deadlock caused by recursive invocations of `messageWithTypeAndLevel`.
This is done by tracking whether the locks are already held with a threadlocal counter per mutex. 

Fixes: #8153

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/ConsoleObject.zig
+++ b/src/bun.js/ConsoleObject.zig
@@ -72,6 +72,9 @@ pub const MessageType = enum(u32) {
 var stderr_mutex: bun.Lock = bun.Lock.init();
 var stdout_mutex: bun.Lock = bun.Lock.init();
 
+threadlocal var stderr_lock_count: u16 = 0;
+threadlocal var stdout_lock_count: u16 = 0;
+
 /// https://console.spec.whatwg.org/#formatter
 pub fn messageWithTypeAndLevel(
     //console_: ConsoleObject.Type,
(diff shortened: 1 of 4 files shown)


src/bun.js/ConsoleObject.zig
```diff
@@ -72,6 +72,9 @@ pub const MessageType = enum(u32) {
 var stderr_mutex: bun.Lock = bun.Lock.init();
 var stdout_mutex: bun.Lock = bun.Lock.init();
 
+threadlocal var stderr_lock_count: u16 = 0;
```
Jarred-Sumner, line 75: This can be a boolean and then a stack-local boolean and then there's no risk of integer overflow

LukasKastern: Can it be?
I haven't tried it out but you should be able to get multiple levels of recursion.
For example when adding another MyMap to https://github.com/oven-sh/bun/pull/8213/files#diff-aa4ea7eac69e2fd1b42ca19ac25fbc56508a271a55a0ac6d584c5ff5278bd905R10 and printing that as well.

We can bump the u16 to something else. I felt pretty comfortable with 65k logs but who knows lol.

LukasKastern: Sorry for the edits. Was confused for a sec. I reverted to the original message.

Jarred-Sumner (approved):
Thank you for this, lets merge after tests run

Jarred-Sumner:
Thank you

PR #8216, opened 2024-01-16
Fix decimal -> integer conversion bug in process.exit & process.exitCode

Jarred-Sumner:
### What does this PR do?

Use `isAnyInt` instead of `isInt32` when reading the exit code for process.exit and process.exitCode. This fixes spurious "The "code" argument must be an integer` errors when using the Prisma CLI.

Test


paperclover (approved):
good catch

PR #8221, opened 2024-01-17
fix: bom stuff

paperclover:
pro tip: do not merge PRs that do not have tests

note to self: do not open PRs marked as "ready to review" without tests and also forget to mention in the notes that there are not tests


src/bun.js/webcore/blob.zig
```diff
@@ -4488,27 +4482,45 @@ pub const Blob = struct {
             // we don't need to clone
             .clone => {
                 this.store.?.ref();
+
+                if (bom == .utf16_le) {
```
Jarred-Sumner, line 4549: memory leak right here

src/bun.js/webcore/blob.zig
```diff
@@ -4488,27 +4482,45 @@ pub const Blob = struct {
...
+                }
                 // we don't need to worry about UTF-8 BOM in this case because the store owns the memory.
                 return ZigString.init(buf).external(global, this.store.?, Store.external);
             },
             .transfer => {
                 const store = this.store.?;
                 std.debug.assert(store.data == .bytes);
                 this.transfer();
+
+                if (bom == .utf16_le) {
```
Jarred-Sumner, line 4562: memory leak right here

src/bun.js/webcore/blob.zig
```diff
@@ -4488,27 +4482,45 @@ pub const Blob = struct {
...
                 // we don't need to worry about UTF-8 BOM in this case because the store owns the memory.
                 return ZigString.init(buf).external(global, store, Store.external);
             },
             // strings are immutable
             // sharing isn't really a thing
             .share => {
                 this.store.?.ref();
-                // we don't need to worry about UTF-8 BOM in this case because the store owns the memory.s
+
+                if (bom == .utf16_le) {
```
Jarred-Sumner, line 4575: memory leak right here

src/bun.js/webcore/blob.zig
```diff
@@ -4518,6 +4530,12 @@ pub const Blob = struct {
                     return out.toJS(global);
                 }
 
+                if (bom == .utf16_le) {
```
Jarred-Sumner, line 4597: and memory leak right here

src/bun.js/webcore/blob.zig
Jarred-Sumner: move this above this.store.?.ref() to fix the leak

src/bun.js/webcore/blob.zig
Jarred-Sumner: call this.detach() before this.transfer() to fix the leak

src/bun.js/webcore/blob.zig
```diff
@@ -4518,6 +4531,12 @@ pub const Blob = struct {
...
+                    var out = bun.String.createUTF16(bun.reinterpretSlice(u16, buf));
+                    defer out.deref();
```
Jarred-Sumner, line 4599: this still leaks!

Jarred-Sumner (changes requested):
1 more leak

src/bun.js/webcore/blob.zig
jdalton: Learning moment for me. How is this a leak and why does moving it above fix it?

PR #8223, opened 2024-01-17
windows: implement os.hostname

paperclover:
Instead of using libuv i just call the windows method directly and save a utf8 encode step


Jarred-Sumner:
Thank you

PR #8224, opened 2024-01-17
`bun run`: "package.json scripts" at bottom

RiskyMH:
### What does this PR do?

This makes the `bun run` command show the scripts last so that its the easiest to see scripts that are available. This also improves the errors of not having `package.json` by still showing the examples and docs but at end warn user (because you can still use `bun run` just on a file).

This also removes an unnessary usage of `switch` as @paperdave suggested.

- [x] Code changes

I tested the command with locally built version.


Jarred-Sumner:
Looks great. Thank you

paperclover (approved):
lstm

Issue #3786, opened 2023-07-24
Support Nx


birkskyum:
Nx - [Website](https://nx.dev/) - [GitHub](https://github.com/nrwl/nx)

### What version of Bun is running?

0.7.0

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

* `bun x create-nx-workspace`

```zig
NX   Let's create a new workspace [https://nx.dev/getting-started/intro]

✔ Where would you like to create your workspace? · test2
✔ Which stack do you want to use? · react
✔ What framework would you like to use? · none
✔ Standalone project or integrated monorepo? · standalone
✔ Which bundler would you like to use? · vite
✔ Default stylesheet format · css
✔ Enable distributed caching to make your CI faster · No
```
* `bun i`
* `bun --bun run start`

### What is the expected behavior?

```zig

➜ npm run start

> @test2/source@0.0.0 start
> nx serve


> nx run test2:serve:development

  ➜  Local:   http://localhost:4200/


```

### What do you see instead?

```zig
➜ bun --bun run start
$ nx serve
perf_hooks_1.performance.mark is not a function. (In 'perf_hooks_1.performance.mark("init-local")', 'perf_hooks_1.performance.mark' is undefined)
error: script "start" exited with code 1 (SIGHUP)
```

<details>
<summary>Debug log</summary>

```zig
[CPP] ZigString__toExternalValue
[JSC] deref (src/string.zig:95)
[JSC] deref (src/string.zig:95)
[CPP] JSGlobalObject__bunVM
[CPP] JSGlobalObject__bunVM
[CPP] JSValue__isBoolean
[CPP] JSValue__jsType
[CPP] JSGlobalObject__vm
[CPP] JSValue__isCallable
[CPP] JSValue__toZigString
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
perf_hooks_1.performance.mark is not a function. (In 'perf_hooks_1.performance.mark("init-local")', 'perf_hooks_1.performance.mark' is undefined)
[CPP] JSGlobalObject__bunVM
[CPP] JSGlobalObject__bunVM
[JSC] dispatchOnExit (src/bun.js/javascript.zig:353)
error: script "start" exited with code 1 (SIGHUP)
```
</details>

### Additional information

[...]

PR #8237, opened 2024-01-17
node:fs: allow WriteStream._write to be overwritten

nektro:
Related #3786 gets `nx build` working


Jarred-Sumner (changes requested):
process-stdio test is failing

PR #7692, opened 2023-12-16
fix(expect): bug in string handling in toContain and return undefined in all matchers

DontBreakAlex:
### What does this PR do?

This changes toContains to pass when the needle is an empty string. [ref](https://tc39.es/ecma262/2023/#sec-stringindexof).
This also changes all matchers to return undefined on success instead of returning expect.

fixes #7144 

### How did you verify your code works?

I wrote automated tests

PR #8270, opened 2024-01-18
fix(bun:test): `toContain` and `toContainEqual` string fix

dylan-conway:
### What does this PR do?
Matches jest behavior with `toContain(Equal)` and strings.

Partly a follow up to https://github.com/oven-sh/bun/pull/7692, fixes two tests in `expect.test.js`.

added tests and ran them with bun and jest


src/bun.js/bindings/bindings.zig
```diff
@@ -4183,6 +4191,27 @@ pub const JSValue = enum(JSValueReprInt) {
...
+
+    /// Returns true if
+    /// - `new String("123")`
+    /// - `class DerivedString extends String; new DerivedString("123")`
+    pub inline fn isStringObjectLike(this: JSValue) bool {
+        if (!this.isCell()) {
+            return false;
+        }
+
+        return jsType(this).isStringObject() or jsType(this).isDerivedStringObject();
```
Jarred-Sumner: This should be in `JSType` and then this should call `this.jsType().isStringObjectLike()`

Jarred-Sumner: because this calls `jsType(this)` twice when it definitely doesn't need to

src/bun.js/test/expect.zig
```diff
@@ -961,12 +963,30 @@ pub const Expect = struct {
...
+                const expected_string = expected.toString(globalObject).toSlice(globalObject, default_allocator);
+                defer expected_string.deinit();
+
+                // jest does not have a `typeof === "string"` check for `toContainEqual`.
+                // it immediately spreads the value into an array.
+
+                var expected_codepoint_cursor = strings.CodepointIterator.Cursor{};
+                var expected_iter = strings.CodepointIterator.init(expected_string.slice());
+                _ = expected_iter.next(&expected_codepoint_cursor);
+
```
Jarred-Sumner, line 981: we're iterating over it character by character after already converting to UTF-8?

can't we use strings.indexOf?

dylan-conway: good point

src/bun.js/bindings/bindings.zig
dylan-conway: updated

Jarred-Sumner (approved):

PR #8291, opened 2024-01-20
Add PrivateRecursive type annotations

jdalton:
### What does this PR do?

~~Add alwaysInline and PrivateRecursive annotations~~

Add PrivateRecursive type annotations

-->

- [ ] Code changes

### DIFF
--- a/src/js/builtins.d.ts
+++ b/src/js/builtins.d.ts
@@ -23,7 +23,7 @@ declare var $overriddenName: string;
 /** ??? */
 declare var $linkTimeConstant: never;
 /** Assign to this directly above a function declaration (like a decorator) to set visibility */
-declare var $visibility: "Public" | "Private";
+declare var $visibility: "Public" | "Private" | "PrivateRecursive";
 /** ??? */
 declare var $nakedConstructor: never;
 /** Assign to this directly above a function declaration (like a decorator) to set intrinsic */


src/js/builtins/Module.ts
```diff
@@ -3,14 +3,16 @@ export function main() {
   return $requireMap.$get(Bun.main);
 }
 
-$visibility = "Private";
+$visibility = "PrivateRecursive";
+$alwaysInline;
```
jdalton: Adds `@alwaysInline` Like JSC does for `Array#filter`:
https://github.com/WebKit/WebKit/commit/6cb56ee3051f43ee8030b9c34b4ea035a2effa6a

src/js/builtins/ImportMetaObject.ts
```diff
@@ -1,6 +1,6 @@
 type ImportMetaObject = Partial<ImportMeta>;
 
-$visibility = "Private";
+$visibility = "PrivateRecursive";
 export function loadCJS2ESM(this: ImportMetaObject, resolvedSpecifier: string) {
```
jdalton: Protects call stack of internals recursively. (see https://github.com/WebKit/WebKit/commit/37f814d56d6446ea011af18ee65b6644c1127b4e).

jdalton:
This PR is mainly to introduce these, as I found them cool. Y'all may have a better idea of where to add them and see if it makes a difference.

src/js/builtins/ImportMetaObject.ts
```diff
@@ -109,7 +109,7 @@ export function requireESM(this: ImportMetaObject, resolved) {
   return exports;
 }
 
-$visibility = "Private";
+$visibility = "PrivateRecursive";
 export function internalRequire(this: ImportMetaObject, id) {
   var cached = $requireMap.$get(id);
```
jdalton: Depending on how much these restrict the stack we may not want to do this. It's mostly trying them on.

src/js/builtins/Module.ts
paperclover: i wonder if this actually does anything. this function is almost always `.bind()`ed to the module record.

src/js/builtins/Module.ts
```diff
@@ -3,14 +3,16 @@ export function main() {
...
 export function require(this: CommonJSModuleRecord, id: string) {
   return $tailCallForwardArguments($overridableRequire, this);
 }
 
 // overridableRequire can be overridden by setting `Module.prototype.require`
 $overriddenName = "require";
-$visibility = "Private";
+$visibility = "PrivateRecursive";
+$alwaysInline;
```
paperclover: similarly on this one, i wonder if this actually ever gets inlined, considering it is a part of the `tailCallForwardArguments`.

paperclover (commented):
i dont think making any of these functions PrivateRecursive is going to have any specific effect. but i dont think it is of harm to do it either.

so long as stack traces of a required module dont get hidden (`require()` calls into the user code, so perhaps that gets recursively marked), then it's okay.

but i'm not really sold on this change needing to happen at all. adding the typedef is good tho, we should add that at the very minimum.

src/js/builtins/Module.ts
jdalton: Ya, this PR was mainly a "see something, say something" moment. I think knowing we have the ability to specify `$awaysInline` is powerful. Not likely the place for it though you are right.

src/js/builtins/Module.ts
jdalton: My concern is that this would hide things we DON'T want hidden too.

jdalton:
@paperdave 
> but i'm not really sold on this change needing to happen at all. adding the typedef is good tho, we should add that at the very minimum.

Same. I noticed in WebKit's code they reach for `PrivateRecursive` now more than `Private` but our usage/milage may vary from WebKit core.

jdalton:
Updated to just the d.ts type 😸

PR #8295, opened 2024-01-20
fix Bun.connect(localhost) overly preferring ipv6

nektro:
makes more progress towards passing the https://github.com/sidorares/node-mysql2 test suite

one of the diffs is from changing a test that was added in a fix for https://github.com/oven-sh/bun/issues/7355. below is proof the actual issue has not regressed.

image

image

### DIFF
--- a/packages/bun-usockets/src/bsd.c
+++ b/packages/bun-usockets/src/bsd.c
@@ -24,6 +24,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <arpa/inet.h>
 
 #ifndef _WIN32
 //#define _GNU_SOURCE
@@ -751,6 +752,7 @@ static int bsd_do_connect(struct addrinfo *rp, int *fd)
 
 LIBUS_SOCKET_DESCRIPTOR bsd_create_connect_socket(const char *host, int port, const char *source_host, int options) {
     struct addrinfo hints, *result;
+    char buf[16];
     memset(&hints, 0, sizeof(struct addrinfo));
     hints.ai_family = AF_UNSPEC;
     hints.ai_socktype = SOCK_STREAM;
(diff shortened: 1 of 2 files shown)


test/js/node/net/node-net-server.test.ts
```diff
@@ -282,9 +282,7 @@ describe("net.createServer listen", () => {
         } catch (e) {
           err = e as Error;
         }
-        expect(err).not.toBeNull();
-        expect(err!.message).toBe("Failed to connect");
-        expect(err!.name).toBe("ECONNREFUSED");
+        expect(err).toBeNull();
```
Hanaasagi: If the server is listening on `0.0.0.0`, and the client connects using `::`, should this be successful?


node.js server and client

```JavaScript
const net = require("net");

const server = net.createServer(socket => {
  console.log("Client connected");

  socket.on("data", data => {
    console.log(`Received data from client: ${data}`);

    socket.write(data);
  });

  socket.on("end", () => {
    console.log("Client disconnected");
  });
});

server.listen(8000, "0.0.0.0", () => {
  console.log("Server listening on", server.address());
});

const client = net.createConnection(
  {
    host: "::",
    // host: "127.0.0.1",
    port: 8000,
  },
  () => {
    console.log("Connected to server successfully.");

    client.write("Hello, server!");

    client.end();
  },
);

client.on("data", data => {
  console.log(`Received data from server: ${data}`);
});

client.on("end", () => {
  console.log("Disconnected from server");
});

client.on("error", err => {
  console.error(`Error connecting to server: ${err.message}`);
});

```

Python client

```Python
import socket

server_address = ("::", 8000)
# server_address = ("0.0.0.0", 8000)

with socket.create_connection(server_address) as client:
    print(f"Connected to server {server_address}")

    message = "Hello, server!"
    client.sendall(message.encode("utf-8"))
    print(f"Sent data to server: {message}")

    data = client.recv(1024)
    print(f'Received data from server: {data.decode("utf-8")}')

```

7f8ddd:
I believe this was (and may still be) an issue that makes the debugger not work. It'll say the port is in use when it's not. There's an issue somewhere about it already. I personally just gave up on trying, as the workaround involved editing the hosts file to remove the :: entries.

PR #8299, opened 2024-01-20
(docs) Fixed color in node compat

bjon:
`node:module` was green, when it should be yellow.

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
Thank you

Issue #8280, opened 2024-01-19
error: cannot write multiple output files without an output directory


andyjessop:
### What version of Bun is running?

1.0.23+83f2432da

### What platform is your computer?

Darwin 23.2.0 x86_64 i386

### What steps can reproduce the bug?

```ts
async function main() {
  const result = await Bun.build({
    entrypoints: ['./src/entry.ts'],
    naming: '[dir]/[name].[hash].[ext]',
    splitting: true,
  });

  console.log(result);
}
```
Gives:
```bash
{
  outputs: [],
  success: false,
  logs: [
    error: cannot write multiple output files without an output directory
  ],
}
```
Adding an `outDir` produces the following results:
<img width="270" alt="Screenshot 2024-01-19 at 14 53 44" src="https://github.com/oven-sh/bun/assets/1304307/7083dad9-e2b8-4279-b64f-78576c8bebf1">

### What is the expected behavior?

I would expect that not having an `outDir` will produce exactly the same output, just not written to disk.

### What do you see instead?

The error above

Jarred-Sumner:
Can you try `write: false`? I think that exists

andyjessop:
I get the same result:
```
{
  outputs: [],
  success: false,
  logs: [
    error: cannot write multiple output files without an output directory
  ],
}
```

Full build object for reference:
```
{
  config: {
    entrypoints: [ "./src/entry.ts" ],
    naming: "[dir]/[name].[hash].[ext]",
    plugins: [
      [Object ...]
    ],
    splitting: true,
    write: false,
  },
  onDispose: [Function],
  onEnd: [Function],
  onLoad: [Function: onLoad],
  onResolve: [Function: onResolve],
  onStart: [Function],
  resolve: [Function],
  module: [Function],
  initialOptions: {
    entrypoints: [ "./src/entry.ts" ],
    naming: "[dir]/[name].[hash].[ext]",
    plugins: [
      [Object ...]
    ],
    splitting: true,
    write: false,
    bundle: true,
    entryPoints: [ "./src/entry.ts" ],
    minify: false,
    minifyIdentifiers: undefined,
    minifyWhitespace: undefined,
    minifySyntax: undefined,
    outbase: undefined,
    platform: undefined,
  },
  esbuild: {},
}
```

aboqasem:
The test is there but commented out:

[...]

PR #8315, opened 2024-01-21
test(bun-build-api): improve tests

aboqasem:
### What does this PR do?

- [ ] Code changes


packages/bun-types/bun.d.ts
```diff
@@ -1265,7 +1265,7 @@ declare module "bun" {
   type ModuleFormat = "esm"; // later: "cjs", "iife"
 
   interface BuildConfig {
-    entrypoints: string[]; // list of file path
+    entrypoints: [string, ...string[]]; // list of file path
```
paperclover: this type change may break some uses of the api

insight from @Didas-git:

image

imo we shouldn't do this change.

test/bundler/bun-build-api.test.ts
```diff
@@ -168,36 +154,53 @@ describe("Bun.build", () => {
...
 
-  // test("BuildArtifact properties splitting", async () => {
-  //   Bun.gc(true);
-  //   const x = await Bun.build({
-  //     entrypoints: [join(import.meta.dir, "./fixtures/trivial/index.js")],
-  //     splitting: true,
-  //   });
-  //   expect(x.outputs).toHaveLength(2);
-  //   const [indexBlob, chunkBlob] = x.outputs;
+  test("BuildArtifact properties splitting", async () => {
```
paperclover, line 155: lol why were these commented out.

paperclover (changes requested):
looks good but skeptical about the typedef change

packages/bun-types/bun.d.ts
Didas-git: To explain what i meant in the last line:
When it comes to this tuple types tss will do a lot of instanciations wich is pretty much useless if you are not making use of the type to begin with.

_the white theme hurts me tho_

aboqasem: makes sense, thanks for the review! i've reverted the type change

paperclover (approved):

test/bundler/bun-build-api.test.ts
aboqasem: there's even one more haha: https://github.com/oven-sh/bun/issues/8280#issuecomment-1902430096, i'm just not sure what's the expected there

PR #8317, opened 2024-01-21
fix(transpiler): loose equality folding bugfix

### LINKED ISSUE #8311: Bun returns "" == 0 as false

1.0.23+83f2432da

Darwin 23.2.0 arm64 arm

```javascript
// test.js
const a = "" == 0; 
const b = 0 == ""; 

console.log(`a is ${a}`);
console.log(`b is ${b}`);
```

```shell
❯ bun test.js
```

```shell
a is true
b is true
```

### What do you see instead?

```shell
a is false
b is true
```

### Additional information

_No response_

dylan-conway:
### What does this PR do?
fixes edge case involving loose equality of a string and a number or two bigints

```
"" == 0;
"-0" == 0;
1234n == 5678n;
```
before: false, false, true
after: true, true, false

fixes #8311 

added tests

### DIFF
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -5266,7 +5266,7 @@ pub const Expr = struct {
                                 if (r.value == 0 or r.value == 1) {
                                     equality.ok = true;
                                     equality.equal = if (r.value == 0)
-                                        l.eqlComptime("0")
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
awesome

Issue #8264, opened 2024-01-18
bun crash `mimalloc: error: mi_free_size: invalid (unaligned) pointer: 0x2000c166903`


Hanaasagi:
### What version of Bun is running?

1.0.23+83f2432da

### What platform is your computer?

Linux 6.6.10-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

Clone this repo https://github.com/wasm-fmt/biome_fmt and build.

```
git clone https://github.com/wasm-fmt/biome_fmt.git
curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
./crates/biome_fmt/scripts/build.sh
cd crates/biome_fmt/pkg && npm pack
```

then

```
bun test test_bun
```

### What is the expected behavior?

no crash

### What do you see instead?

```
mimalloc: error: mi_free_size: invalid (unaligned) pointer: 0x2000c166903
[SYS] read(526, 65536) = 0 (0.002ms)
mimalloc: assertion failed: at "/home/kumiko/zig/bun/src/deps/mimalloc/include/mimalloc/internal.h":436, _mi_segment_page_of
  assertion: "diff > 0 && diff <= (ptrdiff_t)MI_SEGMENT_SIZE"
[fs] close(526)
[Blob] ReadFile.onFinish() = immediately
[1]    104581 IOT instruction (core dumped)  ~/zig/bun/build/bun-debug test test_bun

```

### Additional information

coredump stack

```
(lldb) target create "/home/kumiko/zig/bun/build/bun-debug" --core "/var/tmp/coredump-6U5rUP"
Core file '/var/tmp/coredump-6U5rUP' (x86_64) was loaded.
(lldb) bt
* thread #1, name = 'bun-debug', stop reason = signal SIGABRT
  * frame #0: 0x00007f0a1edf383c libc.so.6`___lldb_unnamed_symbol3601 + 268
    frame #1: 0x00007f0a1eda3668 libc.so.6`raise + 24
    frame #2: 0x00007f0a1ed8b542 libc.so.6`abort + 353
    frame #3: 0x0000559b005c6183 bun-debug`_mi_assert_fail(assertion=<unavailable>, fname=<unavailable>, line=436, func=<unavailable>) at options.c:373:3
    frame #4: 0x0000559b005bc8ec bun-debug`_mi_segment_page_of(segment=0x0000000000000000, p=0x000002000c166903) at internal.h:436:3
[...]

PR #8322, opened 2024-01-21
fix: fix the invalid free pointer in toStringWithBytes.

Hanaasagi:
### What does this PR do?

Close: #8264

`buf` is from `const bom, const buf = strings.BOM.detectAndSplit(raw_bytes);`. The likely cause is that when calling `bun.default_allocator.free(buf);`, the `buf` pointer's value is not a valid and properly aligned pointer.

- [x] Code changes

Re-run the tests from https://github.com/wasm-fmt/biome_fmt, and no crash.

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -4474,7 +4474,7 @@ pub const Blob = struct {
                 }
 
                 if (lifetime == .temporary) {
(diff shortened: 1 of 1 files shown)


dylan-conway (approved):

dylan-conway:
Nice, thank you

PR #8327, opened 2024-01-21
Introduce `Bun.stringWidth`

### LINKED ISSUE #8155: `console.table()` doesn't handle colored text properly

1.0.22+b400b36ca

Linux 6.6.10-arch1-1 x86_64 unknown

when highlighting text using [chalk](https://www.npmjs.com/package/chalk) console.table widths are not displayed correctly

```ts
console.table([
    {
        // chalk.green("ALL")(
        Method: "\u001B[32mALL\u001B[39m",
        // chalk.yellow("/api/auth/*")(
        Path: "\u001B[33m/api/auth/*\u001B[39m",
    },
    {
        Method: "\u001B[32mGET\u001B[39m",
        Path: "\u001B[33m/api/user/me\u001B[39m",
    },
    {
        Method: "\u001B[32mPOST\u001B[39m",
        Path: "\u001B[33m/api/web3/challenge\u001B[39m",
    },
    {
        Method: "\u001B[32mPOST\u001B[39m",
        Path: "\u001B[33m/api/web3/verify\u001B[39m",
    },
    {
        Method: "\u001B[32mGET\u001B[39m",
        Path: "\u001B[33m/src/*\u001B[39m",
    },
    {
        Method: "\u001B[32mGET\u001B[39m",
        Path: "\u001B[33m/*\u001B[39m",
    },
]);
```
image

borders to all line up and zero width characters aren't affecting string length

with and without chalk/escape characters
image

### What do you see instead?

borders are extra wide, but content becomes even narrower than before when using chalk

### Additional information

I'm using green/yellow chalk, but likely affects many/all of these terminal color escape seqences

possible duplicate or related to https://github.com/oven-sh/bun/issues/7955 however i am running a newer version and this isn't emojis so could be seperate

Jarred-Sumner:
### What does this PR do?

This fixes an issue where console.table would not align text properly when the input contains ANSII escape sequences.

This also introduces `Bun.stringWidth`, which has nearly the same API as the extremely popular [`string-width`](https://www.npmjs.com/package/string-width) npm package (> 100 million weekly downloads) except this implementation is roughly 490x faster compared to running string-width in node.

Fixes #8155 

```zig
❯ bun string-width.mjs
cpu: Apple M1 Max
runtime: bun 1.0.24 (arm64-darwin)

benchmark                                    time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------------- -----------------------------
npm/string-width (ansi + emoji + ascii)   10.68 µs/iter   (8.17 µs … 832.46 µs)   9.92 µs  16.83 µs  23.13 µs
npm/string-width (ansi + emoji)            8.67 µs/iter     (7.88 µs … 9.59 µs)   8.92 µs   9.59 µs   9.59 µs
npm/string-width (ansi + ascii)             9.9 µs/iter   (7.75 µs … 734.63 µs)   8.75 µs   14.5 µs  28.04 µs
Bun.stringWidth (ansi + emoji + ascii)    82.29 ns/iter   (78.89 ns … 88.89 ns)  83.53 ns  86.75 ns  87.52 ns
Bun.stringWidth (ansi + emoji)            82.48 ns/iter   (76.63 ns … 94.76 ns)  84.05 ns  87.02 ns   88.3 ns
Bun.stringWidth (ansi + ascii)            24.71 ns/iter   (20.75 ns … 35.67 ns)  25.37 ns  26.84 ns  27.58 ns


❯ node string-width.mjs
cpu: Apple M1 Max
runtime: node v21.4.0 (arm64-darwin)

benchmark                                    time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------------- -----------------------------
npm/string-width (ansi + emoji + ascii)   13.85 µs/iter  (11.54 µs … 770.92 µs)  12.71 µs  18.29 µs  22.58 µs
npm/string-width (ansi + emoji)             9.5 µs/iter     (8.94 µs … 9.98 µs)   9.66 µs   9.98 µs   9.98 µs
npm/string-width (ansi + ascii)           12.31 µs/iter     (9.96 µs … 2.28 ms)  11.25 µs   15.5 µs  19.54 µs
```

There are some tests


Jarred-Sumner:
image

tiptenbrink:
This is not a trivial problem, I would be careful taking on the maintenance burden. Looking at the implementation, I think the most important things are correct (also very nice fast implementation), although I'm not entirely sure how a character knows to be width zero after following a zero-width join. I'm sure you've looked at a lot of resources already, but https://www.jeffquast.com/post/ucs-detect-test-results/ is a good article on the difficulty of the problem. It includes some specific examples that I recommend adding to the tests. Of course there is also the problem of width changing over time for Unicode versions, so what is actually displayed might differ by device.

Jarred-Sumner:
> This is not a trivial problem, I would be careful taking on the maintenance burden. Looking at the implementation, I think the most important things are correct (also very nice fast implementation), although I'm not entirely sure how a character knows to be width zero after following a zero-width join. I'm sure you've looked at a lot of resources already, but https://www.jeffquast.com/post/ucs-detect-test-results/ is a good article on the difficulty of the problem. It includes some specific examples that I recommend adding to the tests. Of course there is also the problem of width changing over time for Unicode versions, so what is actually displayed might differ by device.

We're mostly aiming to match the behavior of the `string-width` library and/or node's `readline`. @paperdave has a branch that makes it more directly use ICU's tables instead of our copied version. The ZWJ issue will need to be fixed and I had a branch that does it but it really should use icu's functions instead of our copied version.

Jarred-Sumner:
Merging because I'd like to get this in the release and it feels pretty safe to do so.

Issue #8316, opened 2024-01-21
console.log ignores enumerable:false


sroussey:
### What version of Bun is running?

1.0.24+6fa35839c

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

In file test.js:
```js
const test = {
  a: 1,
  b: 2,
};

console.log(test);

Object.defineProperty(test, "b", { enumerable: false });

console.log(test);
```

### What is the expected behavior?

❯ node test.js
{ a: 1, b: 2 }
{ a: 1 }

### What do you see instead?

❯ bun test.js
{
  a: 1,
  b: 2,
}
{
  a: 1,
  b: 2,
}

sroussey:
So, I made a change that fixes it, but it changes bindings.cpp and I have no idea what other ramifications it might have.

sroussey:
https://github.com/oven-sh/bun/compare/main...sroussey:bun:main

PR #8337, opened 2024-01-21
Don't iterate over non-enumerable

sroussey:
### What does this PR do?

This changes object property iteration to skip all that have enumeration false. This could have unintended side effects that are wide ranging depending on how the function is used.

- [x] Code changes

Tests not written yet, thus currently a draft.

Related: https://github.com/oven-sh/bun/issues/8316

### DIFF
--- a/src/cli/gitignore-for-init
+++ b/src/cli/gitignore-for-init
@@ -10,7 +10,7 @@ yarn-error.log*
 lerna-debug.log*
 .pnpm-debug.log*
 
-# Caches 
+# Caches
 
 .cache
 


src/bun.js/bindings/bindings.cpp
```diff
@@ -4795,7 +4795,11 @@ void JSC__JSValue__forEachProperty(JSC__JSValue JSValue0, JSC__JSGlobalObject* g
         bool anyHits = false;
         JSC::JSObject* objectToUse = prototypeObject.getObject();
         structure->forEachProperty(vm, [&](const PropertyTableEntry& entry) -> bool {
-            if ((entry.attributes() & (PropertyAttribute::Function)) == 0 && (entry.attributes() & (PropertyAttribute::Builtin)) != 0) {
+            int attributes = entry.attributes();
```
Jarred-Sumner:
```suggestion
            const int attributes = entry.attributes();
```

Jarred-Sumner:
There will be a handful of test failures

Jarred-Sumner:
Would be worth looking at the console-log test failures

sroussey:
I see why it failed. But the more I look the less qualified I feel to fix the enumerable issue.

For example, what is `fast` and why is it there?

In the fast area, when it sees `underscoreProto` it short-circuits but later in the while loop it only does that if it is also marked at non-enumerable.

And that check for PropertyAttribute::DontEnum happens above another check for it later. Why would it not be down in the lower `if` PropertyAttribute::DontEnum area? Are the side effects in the code in-between problematic?

And what is the point of the `prototypeCount`?

_[edit: never mind the questions above, i have a better idea of what's going on now]_

sroussey:
Also, it looks like `JSC__JSValue__forEachPropertyOrdered` has more different with it than `JSC__JSValue__forEachProperty` that goes beyond ordering.

jdalton:
Related to https://github.com/oven-sh/bun/pull/6654

sroussey:
I guess my test should move to the main console.log test. Now that I look at it, it is under bun folder which I now see how bun specific tests, which this is not.

sroussey:
I guess I can't get away from coding on console.log. Another decade... lol

sroussey:
BTW: any reason `@actions/core` is not installed? It is required by the tests.

PR #8341, opened 2024-01-22
docs: remove unncessary use of `async`

DaleSeo:
### What does this PR do?

The `async` keyword is unnecessary and could be misleading because there is no `await` calls in the test functions.

- [ ] Code changes


paperclover:
wonder if we should have an example using async *and* await since it is supported

DaleSeo:
@paperdave I found some examples using `﻿async` on https://bun.sh/docs/test/writing#basic-usage. I have no opinions but I guess this is sufficient?

PR #8364, opened 2024-01-22
Include Mise in the installation methods

pepicrft:
### What does this PR do?

Bun can be installed using [Mise](https://github.com/jdx/mise), so I'm adding it to the list of options to install Bun in Linux and macOS environments. For context, Mise is an environment manager that can install and activate tool versions. The installation instructions are codified in plugins. In the case of Bun, that's provided by [this plugin](https://github.com/cometkim/asdf-bun). 

- [ ] Code changes

There have yet to be code changes.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -35,6 +35,10 @@ $ docker run --rm --init --ulimit memlock=-1:-1 oven/bun
 $ proto install bun
 ```
 
+```bash#Mise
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you for this, but we haven't heard enough requests from users about this for it to make sense to include in the readme

PR #8368, opened 2024-01-22
Allow `statusCode` and `statusMessage` to be changed on IncomingMessage

Electroid:
### What does this PR do?

`ngrok` is changing the `statusCode` and `statusMessage` on incoming requests, which fixes the following issue:

```js
1212 |         this[kStopReading] = true;
1213 |         if (!(error instanceof RequestError)) {
1214 |             error = new RequestError(error.message, error, this);
1215 |         }
1216 |         const typedError = error;
1217 |         const { response } = typedError;
                    ^
RequestError: Attempted to assign to readonly property.
 code: "ERR_GOT_REQUEST_ERROR"
```

Unfortunately, `ngrok` still doesn't work. (it hangs instead of error) But I figured we should get this fix out anyway.

- [x] Code changes

There are tests.


src/js/node/http.ts
```diff
@@ -771,13 +773,19 @@ Object.defineProperty(IncomingMessage.prototype, "connection", {
...
 });
 
 Object.defineProperty(IncomingMessage.prototype, "statusMessage", {
   get() {
-    return STATUS_CODES[this[reqSymbol].status];
+    return this[statusMessageSymbol] ?? STATUS_CODES[this[reqSymbol].status];
+  },
+  set(value) {
+    this[statusMessageSymbol] = value;
   },
```
jdalton, line 789: For context:

Node http2 uses a symbol called `'state'`:
https://github.com/nodejs/node/blob/main/lib/internal/http2/compat.js#L71

Related to IncomingMessage however Node uses a data property (not accessor):
https://github.com/nodejs/node/blob/main/lib/_http_incoming.js#L88

Jarred-Sumner: yeah, we should just set the property on the object and not use a getter/setter

svemat01:
Any plans for this PR or is it ready to merge?

PR #8377, opened 2024-01-23
Fix eqlComptimeCheckLenWithKnownType

gvilums:
### What does this PR do?

This PR fixes the function `eqlComptimeCheckLenWithKnownType`, which performed invalid comparisons for types other than u8.

The fix is to specialize the function just for u8 and have the generic function cast its slices to u8 before calling the specialized version.

Furthermore this PR simplifies the implementation.

Also, some tests in env.test.ts are adapted for windows. (this is how the bug was discovered in the first place)


Jarred-Sumner (approved):

Jarred-Sumner:
Let's merge after some tests run

PR #8378, opened 2024-01-23
Update default-trusted-dependencies.txt

balazsorban44:
### What does this PR do?

`bunx auth` should not fail for [`auth`](https://npm.im/auth). Currently, it throws
> error: Failed to run "auth" due to error InvalidExe

Which I traced down to this file. Let me know if this is right! There is no postinstall scripts, see https://www.npmjs.com/package/auth?activeTab=code

This works fine with `pnpx auth`

- [x] Code changes

### DIFF
--- a/src/install/default-trusted-dependencies.txt
+++ b/src/install/default-trusted-dependencies.txt
@@ -109,6 +109,7 @@ appium-windows-driver
 applicationinsights-native-metrics
 argon2
 autorest
+auth
 aws-crt
 azure-functions-core-tools
 azure-streamanalytics-cicd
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Trusted dependencies are for lifecycle scripts in package.json and do nothing for packages without lifecycle scripts

This sounds like a different bug, like maybe we are not marketing the executable file as executable (chmod +x)

ndom91:
Anecdotally, on my machine `bunx auth` does download the package to `/tmp/auth@latest--bunx` and the `node_modules` there contain a `.bin/auth` which has the executable bit set and is symlinked to `../auth/index.js` (which does exist) :thinking:

Jarred-Sumner:
Ah, this is a different bug

There's no shebang at the top of the file

and we currently assume files without extensions in the filename have some kind of shebang

image

balazsorban44:
Thanks for the :fire: reply! Adding a shebang fixed it! `bunx auth` :relieved: :heavy_check_mark:

Issue #8381, opened 2024-01-23
Bun.argv not correct with `bun -e` 


linux-china:
### What version of Bun is running?

1.0.25

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

Run with `bun -e 'console.log(Bun.argv)' first second third`, and output as following

```
[ "/Users/xxx/.bun/bin/bun", "/Users/xxx/demo/[eval]",
  "second", "third"
]
```

and `first` param is ignored.

Issue #2900, opened 2023-05-16
`process.argv[1]` contains the real path when running on a symlink


czuniga9:
### What version of Bun is running?

0.6.1

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

When using a symlink, the content of argv[1] in bun is the file the symlink points to.

```bash
cd /tmp
echo "import { argv } from 'process';console.log(argv[1]);" > foo.mjs
ln -s foo.mjs bar.mjs
bun bar.mjs
```

### What is the expected behavior?

Bun should return the path to the symlink, this is what node does:

`/tmp/bar.mjs`

### What do you see instead?

Bun resolves the symlink and returns the path to the real file:

`/tmp/foo.mjs`

### Additional information

This is useful to have multiple commands in the same file and decide which to run based on argv (see for example busybox).

PR #8385, opened 2024-01-23
some fixes for node:process.argv

nektro:
Fixes https://github.com/oven-sh/bun/issues/8381
Fixes https://github.com/oven-sh/bun/issues/2900

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -512,10 +512,11 @@ pub const Arguments = struct {
         opts.extension_order = args.options("--extension-order");
 
         {
-            var list = std.ArrayList(string).init(allocator);
-            errdefer list.deinit();
-            for (args.positionals()) |item| try list.append(item);
-            for (args.remaining()) |item| try list.append(item);
+            const positionals = args.positionals();
+            const remaining = args.remaining();
+            var list = try std.ArrayList(string).initCapacity(allocator, positionals.len + remaining.len);
+            list.appendSliceAssumeCapacity(positionals);
+            list.appendSliceAssumeCapacity(remaining);
             ctx.passthrough = list.items;
         }
 
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -511,7 +511,13 @@ pub const Arguments = struct {
         opts.env_files = args.options("--env-file");
         opts.extension_order = args.options("--extension-order");
 
-        ctx.passthrough = args.remaining();
+        {
+            var list = std.ArrayList(string).init(allocator);
+            errdefer list.deinit();
+            for (args.positionals()) |item| try list.append(item);
+            for (args.remaining()) |item| try list.append(item);
+            ctx.passthrough = list.items;
+        }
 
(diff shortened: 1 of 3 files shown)


src/cli.zig
```diff
@@ -511,7 +511,13 @@ pub const Arguments = struct {
         opts.env_files = args.options("--env-file");
         opts.extension_order = args.options("--extension-order");
 
-        ctx.passthrough = args.remaining();
+        {
+            var list = std.ArrayList(string).init(allocator);
```
Jarred-Sumner: There's a very common performance mistake in this code.

1. Creating an array of known length
2. Appending to it item-by-item

This means it will resize potentially multiple times.
and it will append to it only one at a time, which is much slower than a single memcpy.

Instead, we can do

```suggestion
            const positional = args.positionals();
            const remaining = args.remaining();
            var list = try std.ArrayList(string).initCapacity(allocator, positional.len + remaining.len);
            list.appendSliceAssumeCapacity(positional);
            list.appendSliceAssumeCapacity(remaining);
            ctx.passthrough = list.items;
```

test/cli/run/run-eval.test.ts
```diff
@@ -40,4 +40,12 @@ describe("bun -e", () => {
     expect(stderr.toString("utf8")).toInclude('"hi" as 2');
     expect(stderr.toString("utf8")).toInclude("Unexpected throw");
   });
+
+  test("positional argv is not mangled", async () => {
```
Jarred-Sumner, line 44: I'm skeptical this will fix the issue without leaking other parameters into it

but we shall see

Jarred-Sumner (changes requested):
Perf thing + lets see what tests say

Jarred-Sumner:
Also I wonder about that code in run_command.zig which looks at positional arguments.

Jarred-Sumner (changes requested):
Test failures

src/cli.zig
```diff
@@ -511,7 +511,14 @@ pub const Arguments = struct {
...
 
-        ctx.passthrough = args.remaining();
+        {
+            const positionals = args.positionals();
+            const remaining = args.remaining();
+            var list = try std.ArrayList(string).initCapacity(allocator, positionals.len + remaining.len);
+            list.appendSliceAssumeCapacity(positionals);
+            list.appendSliceAssumeCapacity(remaining);
+            ctx.passthrough = list.items;
+        }
```
jdalton, line 521: What's the block `{`.... `}` for here?

jdalton:
This is related to changes to https://github.com/oven-sh/bun/pull/8458 which can be applied to both Windows and other (if switching how argv is gotten)

src/cli.zig
nektro: zig has strict no-shadowing rules so i did the local block here to prevent `list` from leaking into the main scope since its not needed once assigned to `ctx.passthrough`

PR #8386, opened 2024-01-23
fix to enqueue bundler tasks in correct event loop

gvilums:
### What does this PR do?

Change the event loop used to enqueue bundler tasks from the raw libuv loop to the uws loop

- [ ] Code changes


Jarred-Sumner:
build is failing

PR #8391, opened 2024-01-23
Delete some dead code

Jarred-Sumner:
### What does this PR do?

✂️


jdalton:
Helpful! I was digging into the File/Blob difference in slice() yesterday early afternoon and was sniffing around all that stuff.

PR #8402, opened 2024-01-23
fix(windows): a bunch of install test fixes

dylan-conway:
ci


src/install/install.zig
```diff
@@ -1910,6 +1910,7 @@ const Waiter = struct {
 // 2.
 pub const PackageManager = struct {
     cache_directory_: ?std.fs.Dir = null,
+    cache_directory_path: string = "",
```
Jarred-Sumner, line 1915: Please add a comment to remove this once we move away from using std.ChildProccess for `git clone`.

test/cli/install/registry/bun-install-registry.test.ts
```diff
@@ -14,25 +14,25 @@ var testCounter: number = 0;
...
-        done();
-      }
-    });
-  });
-});
-
-afterAll(() => {
-  verdaccioServer.kill();
-});
+// beforeAll(async () => {
```
Jarred-Sumner: was disabling verdaccio (and subsequently, these tests) intentional?

test/harness.ts
```diff
@@ -249,3 +250,5 @@ expect.extend({
     };
   },
 });
+
+export const pathSep = platform() === "win32" ? "\\" : "/";
```
Jarred-Sumner:
```suggestion
export const {sep as pathSep} from 'node:path'; 
```

test/cli/install/registry/bun-install-registry.test.ts
dylan-conway: Not intentional, for testing locally

PR #8405, opened 2024-01-23
feat(windows): implement node.exe symlink via the `--bun` flag

paperclover:
this implements --bun for windows.

it places the file at %TEMP%\bun-node-<sha>\node.exe.
it is a hard link to the current executable.

this took much longer than it should and im very mad at string_immutable.zig, the MSCPP debugger, and the unhelpful error message created by the `unreachable` keyword.

uses of std.os are safe because they are given a valid nt object path

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1148,10 +1148,15 @@ pub const Command = struct {
         }
     };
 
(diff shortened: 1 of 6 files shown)


Jarred-Sumner:
yay

PR #8424, opened 2024-01-23
Revise the installation script to align with idiomatic PowerShell

danielniccoli:
### What does this PR do?

This revises the installation script to align with idiomatic PowerShell conventions for better readability and adherence to community standards.

The revision includes several improvements:

- Move the default installation directory on Windows from `%USERPROFILE%\.bun` to `%LOCALAPPDATA%\Programs\bun`. This was changed to adhere to Microsoft OS guidelines, specifically:  
  
  grafik
  https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=net-8.0
- Add a parameter `-InstallToUserProfile` that installs bun to the user profile.
- Add a BUN_HOME environment variable that can be updated easily.
- Add a parameter `-Release` to quickly choose a release channel (cannot be used together with `-Version`).
- Add a parameter `-Version` to manually choose a release version (cannot be used together with `-Release`).
- Add supprot for the `-Verbose` parameter.
- Adds PowerShell 7 as a requirement, drops PowerShell 5 support (I acknowledge this may be controversial, but then again, a developer should have PS7 installed nowadays 🙂)
- Writes output to the correct output streams.

Examples:

### PowerShell 7
```ps
PS> . './install.ps1'
PS> . './install.ps1' -Release Canary
PS> . './install.ps1' -Version 1.0.0
PS> . './install.ps1' -InstallToUserProfile -ForceBaseline -Verbose
PS> iex (iwr https://bun.sh/install.ps1).content
```
### From other processes (i.e., cmd)

The following is the common code scenario to invoke a script after downloading it. This requires that the script is offered as `text/plain`. Alternatively, `iwr` can be used, if the script can only be sent by the server as `text/octet-stream`, but it's less common and `iwr` has a different use-case altogether.
```
pwsh -c "&{iex (iwr https://bun.sh/install.ps1).content}"
```

- [x] Code changes

Manual testing.

### DIFF
--- a/packages/bun-types/globals.d.ts
+++ b/packages/bun-types/globals.d.ts
@@ -1677,11 +1677,45 @@ declare global {
   }
 
   interface ArrayConstructor {
+    /**
+     * Create an array from an iterable or async iterable object.
+     * Values from the iterable are awaited.
+     * 
+     * ```ts
+     * await Array.fromAsync([1]); // [1]
+     * await Array.fromAsync([Promise.resolve(1)]); // [1]
+     * await Array.fromAsync((async function*() { yield 1 })()); // [1]
+     * ```
+     * 
+     * @param arrayLike - The iterable or async iterable to convert to an array.
+     * @returns A {@link Promise} whose fulfillment is a new {@link Array} instance containing the values from the iterator.
+     */
     fromAsync<T>(
-      asyncItems: AsyncIterable<T> | Iterable<T> | ArrayLike<T>,
-      mapfn?: (value: any, index: number) => any,
+      arrayLike: AsyncIterable<T> | Iterable<T> | ArrayLike<T>,
+    ): Promise<Awaited<T>[]>;
+
+    /**
+     * Create an array from an iterable or async iterable object.
+     * Values from the iterable are awaited. Results of the map function are also awaited.
+     * 
+     * ```ts
+     * await Array.fromAsync([1]); // [1]
+     * await Array.fromAsync([Promise.resolve(1)]); // [1]
+     * await Array.fromAsync((async function*() { yield 1 })()); // [1]
+     * await Array.fromAsync([1], (n) => n + 1); // [2]
+     * await Array.fromAsync([1], (n) => Promise.resolve(n + 1)); // [2]
+     * ```
+     * 
+     * @param arrayLike - The iterable or async iterable to convert to an array.
+     * @param mapFn - A mapper function that transforms each element of `arrayLike` after awaiting them.
+     * @param thisArg - The `this` to which `mapFn` is bound.
+     * @returns A {@link Promise} whose fulfillment is a new {@link Array} instance containing the values from the iterator.
+     */
+    fromAsync<T, U>(
+      arrayLike: AsyncIterable<T> | Iterable<T> | ArrayLike<T>,
+      mapFn?: (value: T, index: number) => U,
       thisArg?: any,
-    ): Promise<T[]>;
+    ): Promise<Awaited<U>[]>;
   }
 
   interface ConsoleOptions {
--- a/packages/bun-types/test/array.test.ts
+++ b/packages/bun-types/test/array.test.ts
@@ -1,3 +1,5 @@
+import { expectType } from "./utilities.test";
+
 async function* listReleases() {
   for (let page = 1; ; page++) {
     const response = await fetch(
@@ -14,5 +16,19 @@ async function* listReleases() {
 }
 
 await Array.fromAsync(listReleases());
+
+// Tests from issue #8484
+// https://github.com/oven-sh/bun/issues/8484
+async function* naturals() {
+  for (let i = 0; i < 10; i++) {
+    yield i;
+  }
+}
+
+const test1 = await Array.fromAsync(naturals(), (n) => Promise.resolve(`${n}`));
+expectType<string[]>(test1);
+
+const test2 = await Array.fromAsync([Promise.resolve(1), Promise.resolve(2)]);
+expectType<number[]>(test2);
 
 export {};


src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
-  Write-Output "${C_GREEN}Bun ${DisplayVersion} was installed successfully!${C_RESET}"
-  Write-Output "The binary is located at ${BunBin}\bun.exe`n"
+.OUTPUTS
+None.
 
-  Write-Warning "Bun for Windows is currently experimental.`nFor a more stable experience, please install Bun within WSL:`nhttps://bun.sh/docs/installation`n"
+.LINK
+https://bun.sh/docs/installation
+#>
+#Requires -Version 7
```
paperclover: This will not work on powershell 5. the original script was designed to work on that. Windows 10 Home does not ship with PowerShell 7, so it will not work on a fresh machine.

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
-    Write-Warning "Could not create a hard link for bunx, falling back to a cmd script`n"
-    Set-Content -Path "${BunBin}\bunx.cmd" -Value "@%~dp0bun.exe x %*"
-  }
+A script to install bun.
+
+.DESCRIPTION
+
+This script installs bun. By default, the installation directory is %LOCALAPPDATA%\Programs\bun.
+The installation directory can be customized by specifying the custom installation path through the %BUN_INSTALL% environment variable.
+To install bun to %USERPROFILE%\.bun, simply use the -InstallToUserProfile switch.
```
paperclover: currently the code refers to `%USERPROFILE%\.bun` for where to put data like installation cache and other things. This matches how bun behaves on linux and MacOS.

if this is to be changed in the installation script, it must also be changed in the code so that package management is all in the same `$Env:BUN_INSTALL`

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+#region Prologue
+if ($PSCmdlet.ParameterSetName -eq "Version") {
+    Write-Error "Parameter -Version is not yet supported." -ErrorAction Stop
+}
+
+Write-Warning "Bun for Windows is currently experimental.`nFor a more stable experience, please install Bun within WSL:`nhttps://bun.sh/docs/installation"
+
+trap {
+    if ($installationPath) {
+        Write-Verbose "Installation failed. Cleaning up temporary files, installation directory and PATH variable."
```
paperclover: i like this

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+            ";" + 
+            [Environment]::GetEnvironmentVariable("Path", "User")
+        )
+    }    
+    break
+}
+
+#region Requirements Check
+# Ensure this script is run on Windows and meets the minimum Windows version requirement
+if ($env:OS -eq "Windows_NT") {
```
paperclover: i like this

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+"@
+$kernel32 = Add-Type -MemberDefinition $MethodDefinition -Name kernel32 -Namespace Win32 -PassThru -ErrorAction Stop
+$avx2Support = [bool]($kernel32::GetEnabledXStateFeatures() -band 0b0100)
+if ($avx2Support -and $ForceBaseline) {
+    # We should protect the user from installing a variant with limited functionality, although it is not needed.
+    Write-Warning "You chose to install the baseline variant although your CPU supports the AVX2 instruction set."
+    Write-Warning "The baseline variant has limited functionality."
+    Write-Warning "If you later want to install the fully featured variant, run this script again without the `-ForceBaseline` parameter."
+} elseif (!$avx2Support -and !$ForceBaseline) {
+    # We should not automatically install a version with limited functionality without the user knowing that.
```
paperclover: this should must set the force baseline flag

the downloaded binary in this code path will always fail to verify due to STATUS_ILLEGAL_INSTRUCTION

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+    Write-Error "Bun for Windows is only available for Windows x64." -ErrorAction Stop
+}
+
+# Check for AVX2 support
+$MethodDefinition = @"
+[DllImport("kernel32.dll")]
+public static extern long GetEnabledXStateFeatures();
+"@
+$kernel32 = Add-Type -MemberDefinition $MethodDefinition -Name kernel32 -Namespace Win32 -PassThru -ErrorAction Stop
+$avx2Support = [bool]($kernel32::GetEnabledXStateFeatures() -band 0b0100)
```
paperclover: this code actually doesnt work. which was already the case.
can you swap out for this approach which i've already verified to work a bit better than this check

```
(Add-Type -MemberDefinition '[DllImport("kernel32.dll")] public static extern bool IsProcessorFeaturePresent(int ProcessorFeature);'-Name 'Kernel32'-Namespace 'Win32'-PassThru)::IsProcessorFeaturePresent(40)
```
Taken from https://github.com/oven-sh/bun/issues/43#issuecomment-1837763857

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+if ($avx2Support -and $ForceBaseline) {
+    # We should protect the user from installing a variant with limited functionality, although it is not needed.
```
paperclover: "limited functionality" is incorrect. this binary is just slower because it cannot take advantage of the AVX2 instructions that this binary does not ship with.

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+}
+
+if ($PSCmdlet.ParameterSetName -eq "Release") {
+    $url = "https://github.com/oven-sh/bun/releases/download/$($Release.ToLower())/$file"
+} elseif ($PSCmdlet.ParameterSetName -eq "Version") {
+    $url = "https://github.com/oven-sh/bun/releases/download/bun-v$Version/$file"
+}
+
+try {
+    Invoke-WebRequest $url -OutFile $installer
```
paperclover: IWR is much slower than curl.exe on Powershell 5. it makes the installation experience worse.

src/cli/install.ps1
```diff
@@ -1,176 +1,220 @@
...
+}
+#endregion
+
+#region Installation Verification
+$p = Start-Process $installationPath\bun.exe -ArgumentList "--revision" -NoNewWindow -Wait -PassThru -RedirectStandardOutput NUL
+switch ($p.ExitCode) {
+    0 { continue }
+    # STATUS_ILLEGAL_INSTRUCTION
+    1073741795 {
+        Write-Error ("Validation failed. Bun is not compatible with your CPU. This should have been detected before downloading.`n" +
```
paperclover: imo this should report the error but then re-run the installer with the proper download.

paperclover (changes requested):
big things

- this needs to work on powershell 5. specifically on a clean install of windows 10 1809 (msvcredist is required right now but this will change eventually)
- baseline cpus must default to the baseline build or else they will be unusable
- please switch back to `curl.exe`. it is faster than invoke-webrequest in powershell 5.

paperclover:
oh also IWR shouldn't be used at all because it is not guarenteed to work in powershell 5. this is an absurd edge case but someone will run into it and complain. IRM should be used instead

demo of installing on a testing laptop of mine a few weeks back:

image

src/cli/install.ps1
danielniccoli: I cannot reproduce this

grafik

danielniccoli:
> oh also IWR shouldn't be used at all because it is not guarenteed to work in powershell 5. this is an absurd edge case but someone will run into it and complain. IRM should be used instead
> 
> demo of installing on a testing laptop of mine a few weeks back:
> image

This error only comes um if you don't use `-UseBasicParsing`, which is the default on PS7.

danielniccoli:
I closed this pull request. I'm not comfortable writing an installation script for old PS5, that includes a non-standard installation directory and uses non-standard commands like curl.

PR #8432, opened 2024-01-23
fix windows open flags in some cases

gvilums:
### What does this PR do?

This PR fixes the flags passed to uv_fs_open to be correct on windows. It also makes file truncation work on windows. Fixes issues with `Bun.write`

We should look at the situation more broadly to clean up the way we handle flags around the codebase and make sure that flags are only used in contexts where they actually make sense.

One idea is to introduce a custom `OpenFlags` type which "does the right thing" on each platform and when passed to different functions (such as native apis or libuv) converts to do the right operation.

### DIFF
--- a/src/sys_uv.zig
+++ b/src/sys_uv.zig
@@ -37,7 +37,23 @@ pub const mkdirOSPath = bun.sys.mkdirOSPath;
 
 // Note: `req = undefined; req.deinit()` has a saftey-check in a debug build
 
-pub fn open(file_path: [:0]const u8, flags: bun.Mode, perm: bun.Mode) Maybe(bun.FileDescriptor) {
+// convert linux flags to windows flags
+pub fn normalizeOpenFlags(flags: bun.Mode) c_int {
+    if (comptime Environment.isWindows) {
+        var win_flags: c_int = 0;
+        if (flags & std.os.O.RDONLY != 0) win_flags |= 0o0;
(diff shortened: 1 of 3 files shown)


src/sys_uv.zig
```diff
@@ -37,7 +37,23 @@ pub const mkdirOSPath = bun.sys.mkdirOSPath;
 
 // Note: `req = undefined; req.deinit()` has a saftey-check in a debug build
 
-pub fn open(file_path: [:0]const u8, flags: bun.Mode, perm: bun.Mode) Maybe(bun.FileDescriptor) {
+// convert linux flags to windows flags
+pub fn normalizeOpenFlags(flags: bun.Mode) c_int {
+    if (comptime Environment.isWindows) {
+        var win_flags: c_int = 0;
+        if (flags & std.os.O.RDONLY != 0) win_flags |= 0o0;
```
paperclover, line 44: can you document where these flags are from or use `std.os.windows.`* flags.

paperclover (changes requested):

src/sys_uv.zig
Jarred-Sumner: I have this same code in another PR lol

gvilums: @Jarred-Sumner do you want to merge your PR then and close this one? What do you think?

PR #8437, opened 2024-01-24
Clarifies bun shebang in bunx documentation

FireSquid6:
### What does this PR do?

This PR gives the user information that they can use `#!/usr/bin/env bun` to run scripts with bun. This may feel obvious to some, but I felt it was useful to clarify.


docs/cli/bunx.md
```diff
@@ -63,6 +63,8 @@ $ bunx --bun my-cli # good
 $ bunx my-cli --bun # bad
 ```
 
+Of course, you can also use the `#!/usr/bin/env bun` to execute scripts with bun.
```
Jarred-Sumner: To force Bun to always be used with a script, use a shebang.

```ts
#!/usr/bin/env bun
```

FireSquid6: Think that may be a better way of saying this

Issue #8430, opened 2024-01-23
bun:ffi corrupts some double values passed as arguments


jjrv:
### What version of Bun is running?

1.0.25+a8ff7be64

### What platform is your computer?

Linux 6.2.0-39-generic x86_64 x86_64

### What steps can reproduce the bug?

```
import { dlopen, FFIType } from "bun:ffi";

const { printf } = dlopen("/usr/lib/x86_64-linux-gnu/libc.so.6", {
  printf: { args: [FFIType.cstring, FFIType.f64] }
}).symbols;

printf(new TextEncoder().encode("%.17f\n"), 4);
```

### What is the expected behavior?

Should print:
4.00000000000000000

### What do you see instead?

3.99999999999999956

### Additional information

Adding and removing 0.00000000000001 here:
https://github.com/oven-sh/bun/blob/864a726249115d5d146a54af85d29dc04f093d48/src/js/bun/ffi.ts#L251-L263
is not safe.

There isn't any single floating point value other than 0 that you could replace it with.

Also, if you replace 4 with `-1n` it will print `1`. That's a different bug in the same API function, it's always calling `Math.abs` for no reason.

Jarred-Sumner:
I think we could get away with 0 here

PR #8445, opened 2024-01-24
Fix floating point argument conversion in bun:ffi

jjrv:
### What does this PR do?

Don't try to convert JS values to floating point when calling a function through FFI that expects a double. Instead, handle integer and floating point values in the generated C wrapper code.

See #8430

- [X] Code changes

### DIFF
--- a/test/js/bun/ffi/ffi.test.js
+++ b/test/js/bun/ffi/ffi.test.js
@@ -601,7 +601,7 @@ function ffiRunner(fast) {
           expect(bigint ? BigInt(fn(0n)) : fn(0)).toBe(bigint ? 0n : 0);
 
           for (let i = min; i <= max; i += inc) {
-            // expect(bigint ? BigInt(fn(i)) : fn(i)).toBe(i);
+            expect(bigint ? BigInt(fn(i)) : fn(i)).toBe(i);
           }
         });
       }
--- a/test/js/bun/ffi/ffi.test.js
+++ b/test/js/bun/ffi/ffi.test.js
@@ -377,7 +377,7 @@ function ffiRunner(fast) {
         getDeallocatorBuffer,
       },
       close,
-    } = dlopen("/tmp/bun-ffi-test.dylib", types);
+    } = dlopen("/tmp/bun-ffi-test." + suffix, types);
     it("primitives", () => {
       Bun.gc(true);
       expect(returns_true()).toBe(true);
@@ -455,6 +455,8 @@ function ffiRunner(fast) {
 
       expect(add_float(2.4, 2.8)).toBe(Math.fround(5.2));
       expect(add_double(4.2, 0.1)).toBe(4.3);
+      expect(add_double(4, 4)).toBe(8);
+      expect(add_double(4.5, 4.5)).toBe(9);
       expect(add_int8_t(1, 1)).toBe(2);
       expect(add_int16_t(1, 1)).toBe(2);
       expect(add_int32_t(1, 1)).toBe(2);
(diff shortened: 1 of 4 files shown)


jjrv:
Still need to figure out why the other two tests are broken on my machine. However they're already broken before the code change in this PR. Based on the required `Makefile` changes, the tests may not have been run on Linux before?

jjrv:
Still need to figure out why the other two tests are broken on my machine. However they're already broken before the code change in this PR. Based on the required `Makefile` changes, the tests may not have been run on Linux before?

jjrv:
Still need to figure out why the other two tests are broken on my machine. However they're already broken before the code change in this PR. Based on the required `Makefile` changes, the tests may not have been run on Linux before?

test/js/bun/ffi/ffi.test.js
```diff
@@ -599,7 +601,7 @@ function ffiRunner(fast) {
           expect(bigint ? BigInt(fn(0n)) : fn(0)).toBe(bigint ? 0n : 0);
 
           for (let i = min; i <= max; i += inc) {
-            expect(bigint ? BigInt(fn(i)) : fn(i)).toBe(i);
+            // expect(bigint ? BigInt(fn(i)) : fn(i)).toBe(i);
```
jjrv: This was oddly already failing:

```
Expected: 2147483648
Received: -2147483648

run ffi > FFI runner (fast int) > integer identities work for all possible values > uint32_t
```

paperclover:
one thing jarred mentioned to me about about the approach of doing the check in C is 'how does it impact performance?'

could you do a benchmark in a release build?

test/js/bun/ffi/ffi.test.js
paperclover: this was already broken. i had a pr to fix a different issue with uint32 but it causes another failure. which is funny because this is the test i wrote that is now failing because of that pr.

i do a stupid trick for uint32 and the sign bit is the edge case

jjrv:
> one thing jarred mentioned to me about about the approach of doing the check in C is 'how does it impact performance?'

I added a branch in autogenerated C code and removed a branch from the JS function calling it. It shouldn't make it worse. Is there an existing benchmark or is there reasonable doubt that branches in C are worse?

Also, I *can* make the C branchless using bit twiddling hacks, but doubt it's worth it. That *would* need benchmarking.

test/js/bun/ffi/ffi.test.js
```diff
@@ -605,6 +607,27 @@ function ffiRunner(fast) {
       }
     });
 
+    describe("Floating point identities for all valid exponents", () => {
+      const cases = [
+        // { type: "float", min: -149, max: 128, fn: identity_float },
```
jjrv, line 612: Okayyy, it also corrupts floats so this test fails.

jjrv:
Hmm, with the latest test it just crashes when there's too many `expect` calls :/

test/js/bun/ffi/ffi.test.js
jjrv: I think I fixed it now, but the whole test has become unstable and I suspect it's not because of the number conversions but because it's making too many calls to either FFI or `expect` and something goes wrong

jjrv:
Well, I'm giving up for now due to lack of time. Can't get the tests to run, even though I'm pretty sure the double handling would have been improved. GDB reports:

```
✓ run ffi > FFI runner (fast int) > integer identities work for all possible values > uint64_t [568.95ms]
[jest] test("primitives")

Thread 1 "bun-debug" received signal SIGSEGV, Segmentation fault.
0x0000020000905953 in ?? ()
(gdb) bt
#0  0x0000020000905953 in ?? ()
#1  0x00007fffad408038 in ?? ()
#2  0x00007ffffffe4c30 in ?? ()
#3  0x000055555f005876 in js_trampoline_op_call ()
#4  0x0000000000000000 in ?? ()
```
Looks like an FFI call jumps to a null pointer? Since it always happens after a specific number of tests, maybe something lived in a buffer that got reclaimed? Anyway it'd take some work to find the cause.

Issue #8360, opened 2024-01-22
[windows] Fix panic in `spawn` test


Jarred-Sumner:
The following test panics:
```
bun-debug test test\js\bun\spawn\spawn.test.ts
```

```ts
bun test v1.0.25-canary.7 (b3b5835b)


✓ gcTick > spawnSync > as an array [46.65ms]
✓ gcTick > spawnSync > Uint8Array works as stdin [34.98ms]
✓ gcTick > spawnSync > check exit code [213.12ms]

uh-oh: reached unreachable code
bun will crash now 😭😭😭

----- bun meta -----
Bun v1.0.25-canary.7 (b3b5835b) Windows x64 
TestCommand: bunfig 
Elapsed: 349ms | User: 484ms | Sys: 62ms
RSS: 93.75MB | Peak: 93.78MB | Commit: 0.14GB | Faults: 23258
----- bun meta -----

Search GitHub issues https://bun.sh/issues or join in #windows channel in https://bun.sh/discord

thread 5568 panic: reached unreachable code
???:?:?: 0x7ff7f5569e47 in ??? (???)
???:?:?: 0x7ff7f597f2d9 in ??? (???)
???:?:?: 0x7ff7f5675080 in ??? (???)
???:?:?: 0x2499b36115d in ??? (???)
???:?:?: 0x2499b36111f in ??? (???)
```

Issue #8357, opened 2024-01-22
[windows] Make `$` actually work on Windows


Jarred-Sumner:
@zackradisic I suggest merging more of the code from `Subprocess` and `Subproc` into one to reduce the maintenance burden of having two spawn implementations.

@cirospaciari recently got Bun.spawn working with pipes so I suggest talking to him to do this

Reproduction:
```sh
bun-debug test shell
```

Failing test:
```js
bun test v1.0.25-canary.7 (b3b5835b)




38 |     env(newEnv) {
39 |       if (this.#throwIfRunning(), typeof newEnv === "undefined")
40 |         newEnv = defaultEnv;
41 |       return this.#core.setEnv(newEnv), this;
42 |     }
43 |     #run() {
                     ^
TODOError: spawn() is not yet implemented on Windows
      at #run (:43:17)
      at then (:83:17)
✗ hello world!.repeat(9000) > $(cat < Blob).text() [4.31ms]
✓ hello world!.repeat(9000) > $(cat hello > Blob).text() fails [5.49ms]


38 |     env(newEnv) {
39 |       if (this.#throwIfRunning(), typeof newEnv === "undefined")
40 |         newEnv = defaultEnv;
41 |       return this.#core.setEnv(newEnv), this;
42 |     }
43 |     #run() {
                     ^
TODOError: spawn() is not yet implemented on Windows
      at #run (:43:17)
      at then (:83:17)
✗ hello world!.repeat(9000) > $(cat < Buffer).text() [3.52ms]


38 |     env(newEnv) {
39 |       if (this.#throwIfRunning(), typeof newEnv === "undefined")
40 |         newEnv = defaultEnv;
41 |       return this.#core.setEnv(newEnv), this;
42 |     }
43 |     #run() {
                     ^
TODOError: spawn() is not yet implemented on Windows
      at #run (:43:17)
      at then (:83:17)
✗ hello world!.repeat(9000) > $(cat hello > Buffer).text() passes [3.53ms]


38 |     env(newEnv) {
39 |       if (this.#throwIfRunning(), typeof newEnv === "undefined")
40 |         newEnv = defaultEnv;
41 |       return this.#core.setEnv(newEnv), this;
42 |     }
43 |     #run() {
                     ^
TODOError: spawn() is not yet implemented on Windows
      at #run (:43:17)
      at then (:83:17)
✗ hello world!.repeat(9000) > $(cat < Uint8Array).text() [3.55ms]


38 |     env(newEnv) {
[...]

PR #8456, opened 2024-01-24
wip use wrapper for managing process 

Jarred-Sumner:
### What does this PR do?

This adds `bun.spawn.Process` which is a reusable way to monitor spawned processes

Fixes: #8588
Fixes: #9211
Fixes: #8896
Fixes: #8917
Fixes: #8360
Fixes: #8357
Fixes: #8539


src/deps/uws.zig
```diff
@@ -2554,6 +2554,9 @@ pub const UVLoop = extern struct {
         this.uv_loop.dec();
     }
 
+    pub const ref = inc;
```
paperclover, line 2578: is this for some sort of anytype interface magic thing? maybe we should just rename inc and dec to ref and unref

src/bun.zig
```diff
@@ -915,6 +915,13 @@ pub const SignalCode = enum(u8) {
         return null;
     }
 
+    pub fn toExitCode(value: SignalCode) ?u8 {
```
paperclover, line 1023: please add a comment saying that this is only to be used in Shell scripting, where this is the observed behavior of signals. interpreting a signal code as an exit code is almost always incorrect.

src/deps/libuv.zig
```diff
@@ -1344,14 +1344,50 @@ pub const struct_uv_process_s = extern struct {
...
-    exit_cb: ?*const fn ([*c]struct_uv_process_s, i64, c_int) callconv(.C) void,
+    exit_cb: ?*const fn ([*c]uv_process, i64, c_int) callconv(.C) void,
     pid: c_int,
     exit_req: struct_uv_process_exit_s,
     unused: ?*anyopaque,
     exit_signal: c_int,
     wait_handle: HANDLE,
     process_handle: HANDLE,
     exit_cb_pending: u8,
+
```
paperclover: these functions shouldn't need to alignCast

but i am glad we have more of these wrappers now.
can we not use `HandleMixin` for some of these directly?

src/bun.js/api/bun/process.zig
```diff
@@ -0,0 +1,793 @@
+const bun = @import("root").bun;
```
paperclover, line 1: later it would be nice to merge this file with spawn.zig.PosixSpawn, which usingnamespaces this file.

this file should be named spawn.zig and bun.zig should just do `const spawn = @import("spawn.zig")`. the current way that `bun.spawn` points to this file is much more convoluted than it needs to be.

src/bun.js/api/bun/process.zig
```diff
@@ -0,0 +1,793 @@
...
+
+    pub fn call(this: *const ProcessExitHandler, comptime ProcessType: type, process: *ProcessType, status: Status, rusage: *const Rusage) void {
+        if (this.ptr.isNull()) {
+            return;
+        }
+
+        switch (this.ptr.tag()) {
+            .Subprocess => {
+                if (comptime ProcessType != Process)
+                    unreachable;
```
paperclover: this feels reachable. did you mean
`comptime std.debug.assert(ProcessType == Process)`?

Better question: Why is this type parameter even here? I don't think it is needed

src/bun.js/api/bun/process.zig
```diff
@@ -0,0 +1,793 @@
...
+                const subprocess = this.ptr.as(Subprocess);
+                subprocess.onProcessExit(process, status, rusage);
+            },
+            .LifecycleScriptSubprocess => {
```
paperclover, line 109: it would be interesting if this type of tagged pointer switch could be comptime-ified. i dont think we should do it. but it is something i think about

inline else => |t| { t.as(...).onProcessExit(...) }

paperclover (changes requested):

src/bun.js/bindings/bindings.zig
```diff
@@ -4553,6 +4556,12 @@ pub const JSValue = enum(JSValueReprInt) {
     }
 
     pub fn get(this: JSValue, global: *JSGlobalObject, property: []const u8) ?JSValue {
+        if (comptime bun.Environment.isDebug) {
+            if (bun.ComptimeEnumMap(BuiltinName).has(property)) {
```
dylan-conway: nice

src/which.zig
```diff
@@ -15,6 +15,10 @@ fn isValid(buf: *bun.PathBuffer, segment: []const u8, bin: []const u8) ?u16 {
 // Like /usr/bin/which but without needing to exec a child process
 // Remember to resolve the symlink if necessary
 pub fn which(buf: *bun.PathBuffer, path: []const u8, cwd: []const u8, bin: []const u8) ?[:0]const u8 {
+    if (bin.len == 0) return null;
```
Jarred-Sumner: this diff seems to be incorrect

Jarred-Sumner:
yay

PR #8457, opened 2024-01-24
Reduce dirname/basename/unc methods and rely more on zig standard lib

jdalton:
### What does this PR do?

Bun has dirname, basemame, and handling of UNC namespaced paths (windows) handled in more than a couple places. This leads to not understanding well how paths are being handled (let alone throwing in forking behavior based on platform).
This PR attempts to reduce the places we perform such logic and also leans on Zig's standard lib a bit more. This is step in the direction of consolidation to keep PR scope small and reviewable. Additional to follow.

The file with a good amount of changes is [src/resolver/resolve_path.zig](https://github.com/oven-sh/bun/pull/8457/files#diff-638e3740f72e87644023d9ddf956f545404687b3dec8cec3e705d3f2e05bda75) (which github ui collapses).

Zig `std` apis used
 * `std.fs.path.basename`
 * `std.fs.path.basenamePosix`
 * `std.fs.path.basenameWindows`
 * `std.fs.path.dirname`
 * `std.fs.path.dirnamePosix`
 * `std.fs.path.dirnameWindows`
 * `std.fs.path.extension`
 * `std.fs.path.isAbsolute`
 * `std.fs.path.isAbsolutePosix`
 * `std.fs.path.isAbsoluteWindows`
 * `std.fs.path.windowsParsePath`

Others considered
  * `std.fs.path.ComponentIterator`
  * `std.os.windows.getNamespacePrefix`


build.zig
```diff
@@ -234,7 +234,8 @@ pub fn build_(b: *Build) !void {
     const outfile_maybe = b.option([]const u8, "output-file", "target to install to");
 
     if (outfile_maybe) |outfile| {
-        output_dir = try pathRel(b.allocator, b.install_prefix, std.fs.path.dirname(outfile) orelse "");
+        const to_path = std.fs.path.dirname(outfile) orelse &([_]u8{});
+        output_dir = try pathRel(b.allocator, b.install_prefix, to_path);
```
jdalton, line 238: The `orelse` change from `""` to `&[_]u8{}` more closely matches the empty result done in zig:
https://github.com/ziglang/zig/blob/92211135f1424aaca0de131cfe3646248730b1ca/lib/std/fs/path.zig#L982C20-L982C28

This also avoids a gotcha (not related to an empty string but still) where there was a Bun comment that strings like `"/"` because it affects a buffer where something like `path[0..1]` does not.

src/bun.js/api/ffi.zig
```diff
@@ -711,7 +712,7 @@ pub const FFI = struct {
         const FFI_HEADER: string = @embedFile("./FFI.h");
         pub inline fn ffiHeader() string {
             if (comptime Environment.isDebug) {
-                const dirpath = comptime bun.Environment.base_path ++ (bun.Dirname.dirname(u8, @src().file) orelse "");
+                const dirpath = comptime bun.Environment.base_path ++ (std.fs.path.dirname(@src().file).?);
```
jdalton, line 715: ☝️ Instead of `bun.Dirname.dirname` we use `std.fs.path`. Other places use `.?` when working with `@src().file` so I use it here too. This avoids compilation complaints about the possible `null` case of `std.fs.path.dirname`.

src/bun.js/api/filesystem_router.zig
```diff
@@ -88,11 +89,11 @@ pub const FileSystemRouter = struct {
             if (!(root_dir_path_.len == 0 or strings.eqlComptime(root_dir_path_.slice(), "."))) {
                 // resolve relative path if needed
                 const path = root_dir_path_.slice();
-                if (bun.path.Platform.isAbsolute(.auto, path)) {
+                if (path_handler.Platform.auto.isAbsolute(path)) {
```
jdalton, line 92: ☝️ Consistent access syntax (as follows other places in the code) for `path_handler.Platform.auto.isAbsolute(path)` instead of passing the platform enum into the `isAbsolute(.auto,...` method.

src/bun.js/node/win_watcher.zig
```diff
@@ -57,7 +57,7 @@ pub const PathWatcherManager = struct {
                     .is_file = true,
                     .path = cloned_path,
                     // if is really a file we need to get the dirname
-                    .dirname = Path.dirname(cloned_path, .windows),
+                    .dirname = std.fs.path.dirname(cloned_path),
```
jdalton, line 60: ☝️ Uses std lib dirname. It will detect the Windows platform out of the box.

src/bun.js/node/win_watcher.zig
```diff
@@ -276,12 +276,12 @@ pub const PathWatcher = struct {
         // we need the absolute path to get the file info
         var buf: [bun.MAX_PATH_BYTES + 1]u8 = undefined;
         var parts = [_]string{path};
-        const cwd = Path.dirname(this.path.path, .windows);
+        const cwd = std.fs.path.dirname(this.path.path);
         @memcpy(buf[0..cwd.len], cwd);
```
jdalton, line 280: ☝️ std lib detects Windows out of the box.

src/cli/create_command.zig
```diff
@@ -485,7 +485,7 @@ pub const CreateCommand = struct {
 
                             const createFile = if (comptime Environment.isWindows) std.fs.Dir.createFileW else std.fs.Dir.createFile;
                             var outfile = createFile(destination_dir_, entry.path, .{}) catch brk: {
-                                if (bun.Dirname.dirname(bun.OSPathChar, entry.path)) |entry_dirname| {
+                                if (std.fs.path.dirname(entry.path)) |entry_dirname| {
```
jdalton, line 488: ☝️ This uses `std.fs.path` instead of `bun.Dirname.dirname`. Here there is also a bug that `bun.Dirname.dirname` accepts 1 argument, but here we were passing 2. The usage here suggests at one time this was like [here in zig](https://github.com/ziglang/zig/blob/92211135f1424aaca0de131cfe3646248730b1ca/lib/std/fs/path.zig#L245) with a `comptime T`. However, now [it's not like that now](https://github.com/oven-sh/bun/blob/3447bc2ed3bd68b12729355cc5c132ddd5bc3503/src/resolver/resolver.zig#L4064) like [bun.MakePath.makePath](https://github.com/oven-sh/bun/blob/3447bc2ed3bd68b12729355cc5c132ddd5bc3503/src/bun.zig#L2033).

src/fs.zig
```diff
@@ -124,7 +124,7 @@ pub const FileSystem = struct {
 
         // Ensure there's a trailing separator in the top level directory
         // This makes path resolution more reliable
-        if (!bun.path.isSepAny(top_level_dir[top_level_dir.len - 1])) {
+        if (!std.fs.path.isSep(top_level_dir[top_level_dir.len - 1])) {
```
jdalton, line 127: ☝️ We can rely on the std lib

src/install/install.zig
```diff
@@ -1251,7 +1251,7 @@ pub const PackageInstall = struct {
                     const createFile = if (comptime Environment.isWindows) std.fs.Dir.createFileW else std.fs.Dir.createFile;
 
                     var outfile = createFile(destination_dir_, entry.path, .{}) catch brk: {
-                        if (bun.Dirname.dirname(bun.OSPathChar, entry.path)) |entry_dirname| {
+                        if (std.fs.path.dirname(entry.path)) |entry_dirname| {
```
jdalton, line 1254: ☝️ Use std lib over incorrect `bun.Dirname.dirname`.

src/report.zig
```diff
@@ -72,7 +73,7 @@ pub const CrashReportWriter = struct {
             .{ base_dir, Global.package_json_version, @as(u64, @intCast(@max(std.time.milliTimestamp(), 0))) },
         ) catch return;
 
-        if (bun.path.nextDirname(file_path)) |dirname| {
+        if (std.fs.path.dirname(file_path)) |dirname| {
```
jdalton, line 76: I believe `bun.path.nextDirname` was a one-off way of just doing `dirname` but am unclear. Need a second look 🙏

src/resolver/resolver.zig
```diff
@@ -2542,25 +2542,20 @@ pub const Resolver = struct {
         var path = dir_info_uncached_path_buf[0.._path.len];
 
         bufs(.dir_entry_paths_to_resolve)[0] = (DirEntryResolveQueueItem{ .result = top_result, .unsafe_path = path, .safe_path = "" });
-        var top = Dirname.dirname(path);
+        var top = std.fs.path.dirname(path) orelse &([_]u8{});
```
jdalton, line 2545: ☝️ Instead of `Dirname.dirname` using `std.fs.path.dirname` with a consistent result (avoids the `null`).

src/resolver/resolver.zig
```diff
@@ -2542,25 +2542,20 @@ pub const Resolver = struct {
...
             .hash = 0,
             .status = .not_found,
         };
-        const root_path = if (comptime Environment.isWindows)
-            ResolvePath.windowsFilesystemRoot(path)
-        else
-            // we cannot just use "/"
-            // we will write to the buffer past the ptr len so it must be a non-const buffer
-            path[0..1];
+        const root_path = path_handler.getRoot(path);
```
jdalton, line 2552: ☝️ See https://github.com/oven-sh/bun/pull/8457/files#r1465345955. 
> // we cannot just use "/"
  // we will write to the buffer past the ptr len so it must be a non-const buffer
  
Which is why I avoid string consts like `""` or `"/"` for the `null` case alternative to `std.fs.path.dirname`.

src/resolver/resolver.zig
```diff
@@ -2542,25 +2542,20 @@ pub const Resolver = struct {
...
         var rfs: *Fs.FileSystem.RealFS = &r.fs.fs;
 
         rfs.entries_mutex.lock();
         defer rfs.entries_mutex.unlock();
 
-        while (!strings.eql(top, root_path)) : (top = Dirname.dirname(top)) {
+        while (!strings.eql(top, root_path)) : (top = std.fs.path.dirname(top) orelse &([_]u8{})) {
             const result = try r.dir_cache.getOrPut(top);
```
jdalton, line 2559: ☝️ Swap out `Dirname.dirname` with `std.fs.path.dirname`.

src/resolver/resolver.zig
```diff
@@ -3618,7 +3613,7 @@ pub const Resolver = struct {
             }
         }
 
-        const dir_path = Dirname.dirname(path);
+        const dir_path = std.fs.path.dirname(path) orelse &([_]u8{});
```
jdalton, line 3616: ☝️ Swap out `Dirname.dirname` with `std.fs.path.dirname`.

src/resolver/resolver.zig
```diff
@@ -4011,8 +4006,8 @@ pub const Resolver = struct {
                     try parent_configs.append(tsconfig_json);
                     var current = tsconfig_json;
                     while (current.extends.len > 0) {
-                        const ts_dir_name = Dirname.dirname(current.abs_path);
-                        const abs_path = ResolvePath.joinAbsStringBuf(ts_dir_name, bufs(.tsconfig_path_abs), &[_]string{ ts_dir_name, current.extends }, .auto);
+                        const ts_dir_name = std.fs.path.dirname(current.abs_path) orelse &([_]u8{});
+                        const abs_path = path_handler.joinAbsStringBuf(ts_dir_name, bufs(.tsconfig_path_abs), &[_]string{ ts_dir_name, current.extends }, .auto);
```
jdalton, line 4010: ☝️ Swap out `Dirname.dirname` with `std.fs.path.dirname`.

src/runtime.zig
```diff
@@ -117,7 +117,7 @@ pub const Fallback = struct {
 
     pub inline fn scriptContent() string {
         if (comptime Environment.isDebug) {
-            const dirpath = comptime bun.Environment.base_path ++ (bun.Dirname.dirname(u8, @src().file) orelse "");
+            const dirpath = comptime bun.Environment.base_path ++ (std.fs.path.dirname(@src().file).?);
```
jdalton, line 120: ☝️ Swap out `Dirname.dirname` with `std.fs.path.dirname`. Previously `Dirname.dirname` never returned `null` so the `orelse` was superfluous. Now we don't need it because we are dealing with `@src().file` which we `.?` in all other places.

src/shell/interpreter.zig
```diff
@@ -5665,17 +5666,17 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
 
                         const path_in_dir = std.fs.path.joinZ(fixed_alloc.allocator(), &[_][]const u8{
                             "./",
-                            ResolvePath.basename(src),
+                            std.fs.path.basename(src),
```
jdalton, line 5669: ☝️ Use `std.fs.path.basename` over `resolve_path`'s variant.

src/shell/interpreter.zig
```diff
@@ -5665,17 +5666,17 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
...
                             return false;
                         };
 
                         switch (Syscall.renameat(this.cwd, src, this.target_fd.?, path_in_dir)) {
                             .err => |e| {
-                                const target_path = ResolvePath.joinZ(&[_][]const u8{
+                                const target_path = path_handler.joinZ(&[_][]const u8{
                                     this.target,
-                                    ResolvePath.basename(src),
+                                    std.fs.path.basename(src),
```
jdalton, line 5679: ☝️ Use `std.fs.path.basename` over `resolve_path`'s variant.

src/shell/interpreter.zig
```diff
@@ -6285,10 +6286,10 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
 
                                                     for (filepath_args) |filepath| {
                                                         const path = filepath[0..bun.len(filepath)];
-                                                        const resolved_path = if (ResolvePath.Platform.auto.isAbsolute(path)) path else bun.path.join(&[_][]const u8{ cwd, path }, .auto);
+                                                        const resolved_path = if (path_handler.Platform.auto.isAbsolute(path)) path else path_handler.join(&[_][]const u8{ cwd, path }, .auto);
                                                         const is_root = brk: {
-                                                            const normalized = bun.path.normalizeString(resolved_path, false, .auto);
-                                                            const dirname = ResolvePath.dirname(normalized, .auto);
+                                                            const normalized = path_handler.normalizeString(resolved_path, false, .auto);
+                                                            const dirname = path_handler.Platform.auto.dirname(normalized) orelse &([_]u8{});
```
jdalton, line 6292: ☝️ Use consistent way of platform mode (`.auto`) specific method.

src/string_immutable.zig
```diff
@@ -1651,18 +1655,18 @@ pub fn toWPathNormalized(wbuf: []u16, utf8: []const u8) [:0]const u16 {
...
 
-    if (bun.strings.containsChar(utf8, '/')) {
+    if (bun.strings.containsChar(utf8, sep_posix)) {
         @memcpy(renormalized[0..utf8.len], utf8);
         for (renormalized[0..utf8.len]) |*c| {
-            if (c.* == '/') {
-                c.* = '\\';
+            if (c.* == sep_posix) {
+                c.* = sep_windows;
             }
```
jdalton, line 1663: ☝️ Rely on platform references for things like `'/'` and `\` for meaning/context.

src/string_immutable.zig
```diff
@@ -1651,18 +1655,18 @@ pub fn toWPathNormalized(wbuf: []u16, utf8: []const u8) [:0]const u16 {
...
         }
         path_to_use = renormalized[0..utf8.len];
     }
 
     // is there a trailing slash? Let's remove it before converting to UTF-16
-    if (path_to_use.len > 3 and bun.path.isSepAny(path_to_use[path_to_use.len - 1])) {
+    if (path_to_use.len > 3 and std.fs.path.isSep(path_to_use[path_to_use.len - 1])) {
```
jdalton, line 1669: The `std.fs.path.isSep` works already for Windows.

src/resolver/resolver.zig
```diff
@@ -4060,43 +4055,6 @@ pub const Resolver = struct {
...
-
-        if (end_index == 0 and bun.path.isSepAny(path[0]))
-            return path[0..1];
-
-        if (end_index == 0)
-            return root;
-
-        return path[0 .. end_index + 1];
-    }
-};
```
jdalton, line 4098: Removed the `Dirname.dirname` helper which seems to deviate [on the `null` case of `std.fs.path.dirname`](https://github.com/ziglang/zig/blob/92211135f1424aaca0de131cfe3646248730b1ca/lib/std/fs/path.zig#L805-L871).

src/fs.zig
```diff
@@ -1488,61 +1440,15 @@ pub const PathName = struct {
...
+        const base = std.fs.path.basename(_path);
+        const filename = if (dir.len > 0) _path[dir.len + 1 ..] else _path;
         return PathName{
             .dir = dir,
             .base = base,
-            .ext = ext,
-            .filename = if (dir.len > 0) _path[dir.len + 1 ..] else _path,
+            .ext = std.fs.path.extension(base),
+            .filename = filename,
         };
```
jdalton, line 1452: ☝️ The deleted code above looked to be doing what the standard zig lib does. So I replaced it with the std lib to reduce gotchas. I kept the deviation in behavior though. I need a second look here though 🙏 .

src/fs.zig
```diff
@@ -1352,63 +1352,15 @@ pub const NodeJSPathName = struct {
...
-                }
-            }
+        if (filename.len > 1 and ext.len > 0) {
+            filename = filename[0 .. filename.len - ext.len];
         }
-
-        if (is_absolute) {
-            dir = &([_]u8{});
-        }
-
```
jdalton, line 1411: ☝️ The deleted code above looked to be doing what the standard zig lib does. So I replaced it with the std lib to reduce gotchas. I kept the deviation in behavior though. I need a second look here though 🙏 .

src/resolver/resolve_path.zig
```diff
@@ -167,10 +191,10 @@ pub fn longestCommonPathGeneric(input: []const []const u8, comptime platform: Pl
         inline 2, 3, 4, 5, 6, 7, 8 => |n| {
             // If volume IDs do not match on windows, we can't have a common path
             if (platform == .windows) {
-                const first_root = windowsFilesystemRoot(input[0]);
+                const first_root = getRootWindows(input[0]);
                 comptime var i = 1;
                 inline while (i < n) : (i += 1) {
-                    const root = windowsFilesystemRoot(input[i]);
+                    const root = getRootWindows(input[i]);
```
jdalton, line 197: Created a `getRootWindows` and `getRootPosix` form to generalize as `getRoot`.

src/resolver/resolve_path.zig
```diff
@@ -435,24 +459,6 @@ pub fn relativeNormalized(from: []const u8, to: []const u8, comptime platform: P
...
-            return str[0..separator];
-        },
-        .windows => {
-            const separator = lastIndexOfSeparatorWindows(str) orelse return std.fs.path.diskDesignatorWindows(str);
-            return str[0..separator];
-        },
-        else => @compileError("unreachable"),
-    }
-}
-
```
jdalton, line 455: ☝️ Deleted this `dirname` variation in favor of `std.fs.path.dirname` as this was very bare bones and unlike other variants.

src/resolver/resolve_path.zig
```diff
@@ -496,68 +502,43 @@ pub fn relativeAlloc(allocator: std.mem.Allocator, from: []const u8, to: []const
...
-                    return .{ path.len, idx + 3 };
-                }
-            }
-        }
+    const parsed = std.fs.path.windowsParsePath(path);
+    if (parsed.kind == WindowsPath.Kind.None) {
+        return .{ 0, 0 };
     }
-    return .{ 0, 0 };
+    return .{ parsed.disk_designator.len, 0 };
```
jdalton, line 514: ☝️ Deleted code that the platform covers but need a second look to make sure all is well 🙏

src/resolver/resolve_path.zig
```diff
@@ -496,68 +502,43 @@ pub fn relativeAlloc(allocator: std.mem.Allocator, from: []const u8, to: []const
...
 }
 
 pub fn windowsVolumeName(path: []const u8) []const u8 {
     return path[0..@call(.always_inline, windowsVolumeNameLen, .{path})[0]];
 }
 
+pub fn getRootPosix(path: []const u8) []const u8 {
+    return path[0..1];
```
jdalton, line 522: ☝️ The compiler complained about an unused var or I would have done `sep_str_posix[0 ..1]`

src/resolver/resolve_path.zig
```diff
@@ -496,68 +502,43 @@ pub fn relativeAlloc(allocator: std.mem.Allocator, from: []const u8, to: []const
...
-        if ('a' <= c and c <= 'z' or 'A' <= c and c <= 'Z') {
-            return path[0..3];
-        }
+pub fn getRootWindows(path: []const u8) []const u8 {
+    const parsed = std.fs.path.windowsParsePath(path);
+    const disk_designator = parsed.disk_designator;
+    switch (parsed.kind) {
+        WindowsPath.Kind.None => return path[0..0],
+        WindowsPath.Kind.Drive => return if (path.len > 3) path[0 .. disk_designator.len + 1] else path[0..0],
+        WindowsPath.Kind.NetworkShare => return if (path.len > 4) path[0 .. disk_designator.len + 1] else path[0..0],
```
jdalton, line 532: ☝️ Specifically the `NetworkShare` case needs looking at. I think it may need to be closer to the `strings.indexOfAny(path[3..], "/\\")` thing before. Second look needed :pray: for sure.

src/resolver/resolve_path.zig
```diff
@@ -726,111 +707,126 @@ pub const Platform = enum {
...
+    }
+
     pub fn isAbsolute(comptime platform: Platform, path: []const u8) bool {
         return switch (comptime platform) {
             .auto => (comptime platform.resolve()).isAbsolute(path),
-            .posix => path.len > 0 and path[0] == '/',
-            .windows,
-            .loose,
-            => std.fs.path.isAbsoluteWindows(path),
+            .posix => isAbsolutePosix(path),
```
jdalton, line 731: ☝️ Leverage std lib for `isAbsolutePosix`

src/resolver/resolve_path.zig
```diff
@@ -726,111 +707,126 @@ pub const Platform = enum {
...
-    pub fn getSeparatorFunc(comptime _platform: Platform) IsSeparatorFunc {
-        switch (comptime _platform.resolve()) {
-            .auto => comptime unreachable,
-            .loose => {
-                return isSepAny;
+    pub fn getIsSepFunc(comptime platform: Platform) IsSeparatorFunc {
+        comptime switch (platform.resolve()) {
+            .auto => unreachable,
+            inline .posix, .windows => {
+                return std_isStep;
```
jdalton: Leverage std lib `isSep`.

src/resolver/resolve_path.zig
```diff
@@ -726,111 +707,126 @@ pub const Platform = enum {
...
-            },
-        }
+        };
     }
 
-    pub fn getLastSeparatorFunc(comptime _platform: Platform) LastSeparatorFunction {
-        switch (comptime _platform.resolve()) {
-            .auto => comptime unreachable,
+    pub fn getLastIndexOfSepFunc(comptime platform: Platform) LastSeparatorFunction {
+        comptime switch (platform.resolve()) {
```
jdalton, line 773: ☝️ comptime the whole switch.

src/resolver/resolve_path.zig
```diff
@@ -726,111 +707,126 @@ pub const Platform = enum {
...
-    pub fn getSeparatorFunc(comptime _platform: Platform) IsSeparatorFunc {
-        switch (comptime _platform.resolve()) {
-            .auto => comptime unreachable,
-            .loose => {
-                return isSepAny;
+    pub fn getIsSepFunc(comptime platform: Platform) IsSeparatorFunc {
+        comptime switch (platform.resolve()) {
+            .auto => unreachable,
+            inline .posix, .windows => {
+                return std_isSep;
```
jdalton, line 764: Levage std lib `isSep` and `inline`.

src/resolver/resolve_path.zig
```diff
@@ -726,111 +707,126 @@ pub const Platform = enum {
...
+                if (path[0] == sep_windows)
                     return 0;
 
                 if (path.len < 3)
                     return null;
 
                 // C:\
                 // C:/
-                if (path[0] >= 'A' and path[0] <= 'Z' and path[1] == ':') {
-                    if (path[2] == '/')
```
jdalton, line 831: ☝️ Remove drive letter check as [zig's standard lib](https://github.com/ziglang/zig/blob/993a83081a975464d1201597cf6f4cb7f6735284/lib/std/fs/path.zig#L245) doesn't do this either.

src/resolver/resolve_path.zig
```diff
@@ -907,14 +898,13 @@ pub fn normalizeStringBuf(str: []const u8, buf: []u8, comptime allow_above_root:
             );
         },
         .posix => {
-            return normalizeStringLooseBuf(
+            return normalizeStringPosix(
```
jdalton, line 901: ☝️ Added `normalizeStringPosix` variant for the `.posix` enum.

src/resolver/resolve_path.zig
```diff
@@ -1260,39 +1250,28 @@ fn _joinAbsStringBufWindows(
...
-}
-
-pub fn isSepWin32(char: u8) bool {
-    return char == std.fs.path.sep_windows;
-}
+pub const std_isSep = std.fs.path.isSep;
 
 pub fn isSepAny(char: u8) bool {
-    return @call(.always_inline, isSepPosix, .{char}) or @call(.always_inline, isSepWin32, .{char});
+    return char == sep_posix or char == sep_windows;
```
jdalton, line 1256: ☝️ inlined this instead of calling out to 2 methods.

src/resolver/resolve_path.zig
```diff
@@ -1260,39 +1250,28 @@ fn _joinAbsStringBufWindows(
...
+    return std.mem.lastIndexOfScalar(u8, path, sep_index_of_posix);
 }
 
-pub fn lastIndexOfNonSeparatorPosix(slice: []const u8) ?u32 {
-    var i: usize = slice.len;
-    while (i != 0) : (i -= 1) {
-        if (slice[i] != std.fs.path.sep_posix) {
-            return @as(u32, @intCast(i));
-        }
-    }
```
jdalton, line 1289: ☝️ `lastIndexOfNonSeparatorPosix` was not used ✂️

src/resolver/resolve_path.zig
```diff
@@ -1728,72 +1724,6 @@ test "longestCommonPath" {
...
-    var start_index: usize = end_index;
-    end_index += 1;
-    while (!isSepAny(path[start_index])) {
-        if (start_index == 0)
-            return path[0..end_index];
-        start_index -= 1;
-    }
-
-    return path[start_index + 1 .. end_index];
-}
```
jdalton, line 1750: ☝️ Removed `basename` generic from `resolve_path` in favor of `std.fs.path.basename`.

src/resolver/resolve_path.zig
```diff
@@ -1728,72 +1724,6 @@ test "longestCommonPath" {
...
-
-    if (end_index == 0 and isSepAny(path[0]))
-        return path[0..1];
-
-    if (end_index == 0)
-        return if (root_prefix.len > 0) root_prefix else null;
-
-    return path[0 .. end_index + 1];
-}
-
```
jdalton, line 1796: ☝️ Removed `nextDirname` from `resolve_path` in favor of `std.fs.path.dirname` but need a second set of 👀  on this 🙏

src/fs.zig
paperclover: this function is different

ours:
image

theirs:
image

the difference: we detect backslash on posix and normalize it. im not 100% sure if this is needed in this situation, but for many things we need to handle backslashes on posix.

Jarred-Sumner:
I'm skeptical this makes sense to do. Paths in Bun are necessarily more lenient than in most CLI applications using Zig. Module resolution's path handling needs are very loose. In many cases, we have to mimick the behavior of `node:path` (and also implement it). 

In general, I think we should use less of the zig standard library over time as it has proven to be a maintenance burden when upgrading Zig. Zig's standard library frequently has breaking changes and uses `unreachable` in system calls (less relevant for this PR)

Also, there are 278 test failures

src/fs.zig
jdalton: @paperdave 
> this function is different

Correct. The idea is that in this case we are in code that is executing on a given platform without branching by param for other platforms so using the platform specific `isSep` is okay. This means we don't have to handle both cases and on windows it will handle both back and forward slashes.

jdalton:
@Jarred-Sumner
> Module resolution's path handling needs are very loose. In many cases, we have to mimick the behavior of `node:path` (and also implement it).

Being wiggly is great. Mimicking Node is important so things like `types.zig` path methods make sense. A possibility is moving them to where `resolve_path.zig` has its flavors to merge them. 

> In general, I think we should use less of the zig standard library over time as it has proven to be a maintenance burden when upgrading Zig. Zig's standard library frequently has breaking changes and uses `unreachable` in system calls (less relevant for this PR)

Interesting! With zig being younger it may be that we need to lean on our own initially. Ideally as there are more zig users they will get more 👀 and usage and be hardened more than the projects (like us) that rely on it.

> Also, there are 278 test failures

🙈

jdalton:
I can reference this in future PR work but the exact form in this PR isn't needed.

PR #8458, opened 2024-01-24
fix argv parsing on windows

### LINKED ISSUE #8331: Windows: Non ASCII characters are replaced with '?' in command line arguments

1.0.25-canary.16+a8ff7be64

Execute this command in any terminal emulator that supports emoji - 
```
bun -e "console.log('🐇')"
```

🐇

### What do you see instead?

??

### Additional information

# Reason
Bun uses [std.os.argv](https://ziglang.org/documentation/master/std/#A;std:os.argv) to get command line arguments. This value is "Populated by startup code before main()" according to the docs, and this startup code does not correctly handle non ASCII characters and literally replaces them with question marks (? 0x3f), so what bun sees is `-e console.log("??")` and therefore ?? gets printed.

https://github.com/oven-sh/bun/blob/353f724a9cd12d1749e974e92bbc47d9138a6601/src/bun.zig#L1847-L1858

# Possible Solutions
1. Use `std.process.argsAlloc`. It handles utf-8 correctly and uses  [`GetCommandLineW`](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinew) internally
2. Use a combination of  [`GetCommandLineW`](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinew) and [`GetCommandLineToArgvW`](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw)

### LINKED ISSUE #8412: [windows] non-ascii file path entry point test failure
```ts
bun-debug test test\js\bun\resolve\non-english-import.test.js
```

```ts
bun test v1.0.25-canary.20 (40aff0d4)




15 |   for (let variation of [latin1Chars + "-latin1-prefix.js", "latin1-suffix-" + latin1Chars + ".js"]) {
16 |     const inputPath = join(prefix, variation);
17 |     try {
18 |       mkdirSync(prefix, { recursive: true });
19 |     } catch (e) {}
20 |     await Bun.write(inputPath, `console.log(42);`);
               ^
error: UV_ENOENT
   errno: -4058
 syscall: "open"
   path: "C:\Users\RUNNER~1\AppData\Local\Temp\bun-test-non-english-import-latin1\ÇÇÇÇ-latin1-prefix.js"

      at D:\a\bun\bun\test\js\bun\resolve\non-english-import.test.js:20:11
      at D:\a\bun\bun\test\js\bun\resolve\non-english-import.test.js:10:28
✗ latin1 entry point [2.84ms]


39 |   for (let variation of [utf16Chars + "-utf16-prefix.js", "utf16-suffix-" + utf16Chars + ".js"]) {
40 |     const inputPath = join(prefix, variation);
41 |     try {
42 |       mkdirSync(prefix, { recursive: true });
43 |     } catch (e) {}
44 |     await Bun.write(inputPath, `console.log(42);`);
               ^
error: UV_ENOENT
   errno: -4058
 syscall: "open"
   path: "C:\Users\RUNNER~1\AppData\Local\Temp\bun-test-non-english-import-u16\😀😀😀😀-utf16-prefix.js"

      at D:\a\bun\bun\test\js\bun\resolve\non-english-import.test.js:44:11
      at D:\a\bun\bun\test\js\bun\resolve\non-english-import.test.js:34:27
✗ utf16 entry point [2.52ms]


75 |   try {
76 |     mkdirSync(prefix, { recursive: true });
77 |   } catch (e) {}
78 | 
79 |   const entryCode = imports.map(i => `import "${i}";`).join("\n");
80 |   await Bun.write(inputPath, entryCode);
             ^
error: UV_ENOENT
   errno: -4058
 syscall: "open"
   path: "C:\Users\RUNNER~1\AppData\Local\Temp\bun-test-non-english-import-imports\entry.js"

      at D:\a\bun\bun\test\js\bun\resolve\non-english-import.test.js:80:9
      at D:\a\bun\bun\test\js\bun\resolve\non-english-import.test.js:58:32
✗ latin1 & utf16 imports [2.20ms]



 0 pass
 3 fail
Ran 3 tests across 1 files. [58.00ms]
```

gvilums:
### What does this PR do?
This PR fixes argv parsing on windows. Previously, we assumed UTF-8 encoding on windows argv, but this is wrong: argv can be encoded in any arbitrary code page. The correct thing to do is to use `GetCommandLineW` to get the full command line in UTF-16, convert it to UTF-8, and then parse the arguments. This is what the zig stdlib does internally.

Another option is to first parse the arguments using CommandLineToArgvW and then convert the individual arguments to UTF-8, but that might be slower.

Additionally this PR fixes a test in `non-english-import.test.js` to correctly escape windows file paths.

Fixes #8331
Fixes #8412


src/bun.zig
```diff
@@ -1840,12 +1837,23 @@ pub const win32 = struct {
...
 
-    pub inline fn argv() [][*:0]u8 {
-        return std.os.argv;
+    var _argv: [][*:0]u8 = &[_][*:0]u8{};
+
+    pub fn initArgv(allocator: std.mem.Allocator) !void {
+        var iter = try std.process.ArgIterator.initWithAllocator(allocator);
+        defer iter.deinit();
+        // const cmd_line_w = std.os.windows.kernel32.GetCommandLineW();
+        // var iter = try ArgIteratorWindows.init(allocator, cmd_line_w);
```
jdalton: ☝️ debug left in?

src/bun.zig
```diff
@@ -1840,12 +1837,23 @@ pub const win32 = struct {
...
+        defer iter.deinit();
```
jdalton: Do we need two `defer iter.deinit();` calls (line 1844 and 1847)?

jdalton:
Zig now states that `std.os.argv` is not supported on Windows:
https://github.com/ziglang/zig/commit/7b938767bb18535a870d0460c9f4d9e3d93ab053 and recommends using `std.process.argsAlloc` ([zig source](https://github.com/ziglang/zig/blob/dcaf43674e35372e1d28ab12c4c4ff9af9f3d646/lib/std/process.zig#L1060-L1095)). Note the comment `/// Caller must call argsFree on result.`

Usage example: 
https://github.com/ziglang/zig/blob/dcaf43674e35372e1d28ab12c4c4ff9af9f3d646/test/standalone/load_dynamic_library/main.zig#L6-L7:
```
const args = try std.process.argsAlloc(gpa.allocator());
defer std.process.argsFree(gpa.allocator(), args);
```

🗒️ Some neat case where `freeArgs` isn't needed:
https://github.com/ziglang/zig/commit/1bcf674a4390ee238bc63477d3db545dbf7f66dd

jdalton:
Using `std.process.argsAlloc` Zig avoids forking logic for Posix and Windows and uses just that for getting args.

src/bun.zig
gvilums: removed in latest commit

src/bun.zig
gvilums: this was a leftover, removed in latest commit 👍

gvilums:
I agree that we should just use std.process.argsAlloc, but right now there are many places that depend on `bun.argv()` to return a `[][*:0]const u8`, and `argsAlloc` returns a `[][:0]const u8`. So I think it makes sense to have the separate logic for now and overhaul our argument handling at a later point, especially because it's very isolated to just the initialization of argv

jdalton:
@gvilums 
> I agree that we should just use std.process.argsAlloc, but right now there are many places that depend on `bun.argv()` to return a `[][*:0]const u8`, and `argsAlloc` returns a `[][:0]const u8`. So I think it makes sense to have the separate logic for now and overhaul our argument handling at a later point, especially because it's very isolated to just the initialization of argv

Good news! Via the `zig` discord. I got help with the conversion. Example:

```zig
const std = @import("std");

fn processArgs(argv: [][*:0]u8) void {
    _ = argv; // autofix
}

test {
    // const args = std.os.argv;
    const alloc = std.testing.allocator;
    const args = try std.process.argsAlloc(alloc);
    const ptrslice = try alloc.alloc([*:0]u8, args.len);
    for (0..args.len) |i| ptrslice[i] = args[i].ptr;
    processArgs(ptrslice);
}
```
This way it can be done once cross-platform and no forking.

gvilums:
I just went ahead and updated all uses of bun.argv() to deal with a slice of slices instead of using null-terminated pointers. This cleans the code up quite a bit as it allows us to get rid of a lot of `std.mem.span` calls. The code should now be fully cross-platform, and no special-casing for windows is necessary.

jdalton:
Wow @gvilums! 🚀 🎸 !

Do we need to call 
```zig
defer std.process.argsFree(allocator, _argv);
```
in

https://github.com/oven-sh/bun/pull/8458/files#diff-52c167d9c0701a90829218516da76eefd4a0a05a2c57dd216df4599105c921e9R1820-R1822

Or does the compiler say it's not needed?

gvilums:
I think it's reasonable to not call argsFree and just leak the memory, otherwise we have to remember to call argsFree on program exit

Jarred-Sumner:
Thank you

PR #8464, opened 2024-01-24
Introduce `bun-types/strict`

Didas-git:
This is a follow up to #8195 

This PR introduces **opt-in** stricter types for both bun and typescript in general.
This PR is a new version of #7543 and heavily inspired by @mattpocock's work with `ts-reset`.

### Using

All this pr simply does is introduce a new export, to use it users simply need to update their `tsconfig`.

```diff
"types": [
-  "bun-types"
+  "bun-types/strict"
]
```

### Note

This does not introduce any changes for anyone using the types, its completely opt-in for the users who want a stricter/safer/better environment to work with.

Tagging @Electroid and @paperdave since this is a followup pr.


paperclover (approved):
this is interesting.

PR #8471, opened 2024-01-25
fix(windows): transpiler cache and other test fixes

dylan-conway:
### What does this PR do?
transpiler cache, `process.umask`, `bun.file.exists`, and `process.args` through `bun <script>` working on windows

existing tests and ci


src/bun.js/webcore/blob.zig
```diff
@@ -2894,14 +2894,9 @@ pub const Blob = struct {
...
 
-        if (comptime Environment.isWindows) {
-            this.globalThis.throwTODO("exists is not implemented on Windows");
-            return JSValue.jsUndefined();
-        }
-
         // We say regular files and pipes exist.
         // This is mostly meant for "Can we use this in new Response(file)?"
-        return JSValue.jsBoolean(bun.isRegularFile(store.data.file.mode) or std.os.S.ISFIFO(store.data.file.mode));
+        return JSValue.jsBoolean(bun.isRegularFile(store.data.file.mode) or bun.C.S.ISFIFO(store.data.file.mode));
```
Jarred-Sumner, line 2905: small change this fails on linux if the c version of ISFIFO differs from the linux but we will see

src/linux_c.zig
```diff
@@ -576,6 +576,7 @@ pub const IFF_LOOPBACK = net_c.IFF_LOOPBACK;
 
 pub const Mode = u32;
 pub const E = std.os.E;
+pub const S = std.os.S;
```
Jarred-Sumner, line 579: ah nice

Jarred-Sumner (approved):
LGTM

src/bun.js/RuntimeTranspilerCache.zig
```diff
@@ -246,7 +251,11 @@ pub const RuntimeTranspilerCache = struct {
                 }
             }
 
-            try tmpfile.finish(destination_path.sliceAssumeZ());
+            if (comptime bun.Environment.isWindows) {
+                try tmpfile.finish(@as([:0]const u8, @ptrCast(std.fs.path.basename(destination_path.slice()))));
```
jdalton: Learning bit for me: To convert to the sentinel-terminated pointer we use `@as` combo'd with `@ptrCast`. Can you describe what they do via compiler magic? Side note: My understanding is that sentinel-terminated pointer values are good for interfacing with C++ stuff.

src/bun.js/RuntimeTranspilerCache.zig
```diff
@@ -246,7 +251,11 @@ pub const RuntimeTranspilerCache = struct {
...
+            } else {
+                try tmpfile.finish(destination_path.sliceAssumeZ());
```
jdalton: I see `Z` postfixed things a lot (I don't have a complete grasp on what they are). In this case why fork based on platform. Could the same thing being done in the Windows branch work for Posix?

src/bun.js/RuntimeTranspilerCache.zig
```diff
@@ -539,7 +548,7 @@ pub const RuntimeTranspilerCache = struct {
         const cache_dir_fd = brk: {
             if (std.fs.path.dirname(cache_file_path)) |dirname| {
                 const dir = try std.fs.cwd().makeOpenPath(dirname, .{ .access_sub_paths = true });
-                break :brk bun.toFD(dir.fd);
+                break :brk bun.toLibUVOwnedFD(dir.fd);
```
jdalton, line 548: ✨ ☝️ In other places we platform check before using `toLibUVOwnedFD`... However the non-Windows branch of it just calls `toFD` anyways so this is great. Less platform forking in surrounding logic.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -30,8 +30,11 @@
 #include <sys/utsname.h>
 #else
 #include <uv.h>
+// https://github.com/nodejs/node/blob/62ca05017d758e254eda2dfaf9678d2038a9235b/src/node_process_methods.cc#L28
```
jdalton: The comment url for Umask is not next to the line for defining umask. Could this be moved closer with added comment for context of the link.

src/bun.js/RuntimeTranspilerCache.zig
dylan-conway: This `@ptrCast` is telling the compiler this slice has the correct sentinel at `destination_path[destination_path.len]`, and skips the runtime safefy checks it would normally have in debug builds if it was created like `destination_path[0.. : 0]`. Also this made me realize there's a bug here if `basename` trims from the right side

src/bun.js/RuntimeTranspilerCache.zig
dylan-conway: I'll leave a comment, but the bug this is fixing is a difference in `renameat` on windows when the directory is the same and the destination path is absolute.

src/bun.js/bindings/BunProcess.cpp
dylan-conway: updated

src/bun.js/RuntimeTranspilerCache.zig
dylan-conway: actually I think I will get rid of the windows branch

PR #8473, opened 2024-01-25
Disable dead code elimination when code coverage is enabled

Jarred-Sumner:
### What does this PR do?

This should hopefully make source mappings more accurate when --coverage is in use.

- [ ] Code changes


paperclover (approved):
lstm

Note we have many source-mapping issues. This is treating a symptom instead of the underlying cause. A later project will have to be a full audit of the parser and printer to ensure source mappings are emitted at the correct locations.

Issue #8446, opened 2024-01-24
Bun stucks when I use the HTTP_X_SCHEME header


sirenkovladd:
### What version of Bun is running?

1.0.25+a8ff7be64

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

```js
const { createServer, request } = require('http');

const server = createServer((req, res) => {
  console.log('http', req.url, JSON.stringify(req.headers));
  res.end('ok');
});

server.listen({ port: 0 }, () => {
  const { port } = server.address();
  console.log('port', port);

  const req = request({
    port,
    method: 'POST',
    headers: {
      connection: 'close',
      http_x_scheme: 'http',
    },
  });

  req.on('error', (err) => {
    console.log('err', err);
  });

  req.on('response', (res) => {
    console.log('recieve');
    res.on('data', (chunk) => {
      console.log('data', chunk.toString());
    });
    res.on('end', () => {
      server.close();
    });
  });

  console.log('send');
  req.end();
});
```

### What is the expected behavior?

```sh
❯ node tt.js 
port 52503
send
http / {"connection":"close","http_x_scheme":"http","host":"localhost:52503","content-length":"0"}
recieve
data ok
```

### What do you see instead?

```sh
❯ bun tt.js 
port 52588
send
err ConnectionClosed: The socket connection was closed unexpectedly. For more information, pass `verbose: true` in the second argument to fetch()
 path: "http://localhost:52588/"



```

sirenkovladd:
The error is precisely in uws

when calling `curl -X POST localhost:63453 -H "qwe:qwe"` I see that https://github.com/oven-sh/bun/blob/bun-v1.0.25/packages/bun-uws/src/HttpContext.h#L473 is called

and with `curl -X POST localhost:63453 -H "http_x_scheme:http"` no longer

Maybe @Jarred-Sumner  will have more ideas

sirenkovladd:
Fixed in https://github.com/oven-sh/bun/pull/8830

Issue #3580, opened 2023-07-09
Bun.serve not answers when HTTP header name contains underscore


krutoo:
### What version of Bun is running?

0.6.13

### What platform is your computer?

Darwin 22.1.0 arm64 arm

### What steps can reproduce the bug?

1. create `index.ts` with content:
```ts
Bun.serve({
  port: 1234,
  fetch() {
    return new Response('Hello, world!');
  },
});
```
2. `bun run index.ts`
3. Make http request (for example via Postman) with header `foo-bar_baz: 1.2.3`

### What is the expected behavior?

postman get an answer `Hello, world!`

### What do you see instead?

Postman cannot receive response - it freezes on stage "Sending request..."

When i remove this custom header it works perfectly

### Additional information

According to this resource:
https://www.rfc-editor.org/rfc/rfc9110.html#fields.registry

header names should contains only 

    - alphanumeric characters
    - "."
    - "-"

Maybe it is reason why Bun.serve() not anwsers on requests with headers with underscore

uNetworkingAB:
You aren't even quoting properly. The link you posted says this:

> it SHOULD be restricted to just letters, digits, and hyphen ('-') characters, with the first character being a lette

This is not even close to what you claim it to be. Headers are alphanum and hyphen. Underscore is not allowed, as per the spec recommendation "SHOULD".

krutoo:
@uNetworkingAB but I didn't claim anything =)

uNetworkingAB:
Mr.

Your messages are self-contradictory and also contradicts the chapter of the RFC you linked. You clearly claimed that the RFC allows "dots" while the written word in the very RFC you linked, explicitly does not. You aren't logically making any sense.

Jarred-Sumner:
If we were to add support for this in uWebSockets, I think it'd mean changing these lines:

- https://github.com/Jarred-Sumner/uWebSockets/blob/master/src/HttpParser.h#L256-L264
- https://github.com/Jarred-Sumner/uWebSockets/blob/master/src/HttpParser.h#L233-L240

cc @kwhitley

nickpalmer:
I'm not concerned about it accepting the header or not, but about it hanging if you pass such a header. That is a potential Denial Of Service vector.

If the header isn't valid the server should maybe emit a 409 return code instead of hanging. Alternatively it could just close the connection, but a 409 would at least give consumers an indication something is wrong.

[...]

PR #8475, opened 2024-01-25
fix header with underscore character

sirenkovladd:
### What does this PR do?
Fix response header with underscore character
Fixed: https://github.com/oven-sh/bun/issues/8446

When bun receive request header with underscore request is stuck

- [X] Code changes

### DIFF
--- a/packages/bun-uws/src/HttpParser.h
+++ b/packages/bun-uws/src/HttpParser.h
@@ -227,12 +227,6 @@ namespace uWS
             return unsignedIntegerValue;
         }
 
-        static inline bool isFieldNameFirstByte(unsigned char x)
-        {
-            return ((x > '@') & (x < '[')) |
-                   ((x > 96) & (x < '{'));
-        }
-
         /* RFC 9110 16.3.1 Field Name Registry (TLDR; alnum + hyphen is allowed)
          * [...] It MUST conform to the field-name syntax defined in Section 5.1,
          * and it SHOULD be restricted to just letters, digits,
@@ -241,7 +235,8 @@ namespace uWS
         {
             return (x == '-') |
                    ((x > '/') & (x < ':')) |
-                   isFieldNameFirstByte(x) |
+                   ((x > '@') & (x < '[')) |
+                   ((x > 96) & (x < '{')) |
                    (x == '_') |
                    (x == '.');
         }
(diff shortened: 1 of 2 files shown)


sirenkovladd:
Fixed: https://github.com/oven-sh/bun/issues/3580

packages/bun-uws/src/HttpParser.h
```diff
@@ -235,42 +241,25 @@ namespace uWS
...
-        }
-
         static inline void *consumeFieldName(char *p)
         {
             for (; true; p += 8)
             {
                 uint64_t word;
                 memcpy(&word, p, sizeof(uint64_t));
-                if (notFieldNameWord(word))
+                if (isFieldNameFirstByte(*(unsigned char *)p))
```
Jarred-Sumner: This is too many changes. The only change in the parser code should be to check for `_` and `.`

Jarred-Sumner (changes requested):
One comment

packages/bun-uws/src/HttpParser.h
sirenkovladd: Yes, but per request @paperdave [link](https://discord.com/channels/876711213126520882/888839314056839309/1199998725997723748) i also added first character check for letter only

If it's extra, I can return it

paperclover: i was under the assumption the parser was already this way but it wasnt, i change my mind on this. sorry

paperclover: lets do just the added `|` to keep this as minimal as possible to reduce breaking something

sirenkovladd: will it be enough if i change to

```js
static inline bool notFieldNameWord(uint64_t x)
        {
            return hasLess(x, '9') |
                   hasBetween(x, '9', 'A') |
                   hasBetween(x, 'Z', 'a') |
                   hasMore(x, 'z');
        }
```

sirenkovladd: This will be a small enough change to not break anything, but enough to allow only letters

also a couple of cpu instructions less)
however, in my unprofessional opinion, my version (without hasLess, hasBetween, hasMore) has even fewer cpu instructions

sirenkovladd: @Jarred-Sumner @paperdave

sirenkovladd: Done it

rumblefrog:
Any update on merging & publishing this? This is a rather critical bug that's been opened for a while.

sirenkovladd:
@Jarred-Sumner @paperdave

paperclover:
> This is a rather critical bug that's been opened for a while.

we are instead going with a fix done upstream into usockets itself (which is actually much more permissive than this one was i think)

sirenkovladd:
Fixed in https://github.com/oven-sh/bun/pull/8830

PR #8485, opened 2024-01-25
windows: fix bun plugin

paperclover:
we no longer treat filename "C:\\hello" as namespcae C file path \hello. this change only impacts windows


test/js/bun/plugin/muh.ts
```diff
@@ -0,0 +1,24 @@
+import { plugin } from "bun";
```
Jarred-Sumner, line 1: Where is `muh.ts` loaded?

src/js/builtins/BundlerPlugin.ts
```diff
@@ -253,7 +253,9 @@ export function runOnResolvePlugins(this: BundlerPlugin, specifier, inputNamespa
                 throw new TypeError('onResolve plugin "path" must be absolute when the namespace is "file"');
               }
             } else {
-              // TODO: Windows
+              if (require("path").isAbsolute(path) === false || path.includes("..")) {
```
Jarred-Sumner:
```suggestion
              if (require("node:path").isAbsolute(path) === false || path.includes("..")) {
```

Jarred-Sumner:
what about UNC paths

test/js/bun/plugin/muh.ts
paperclover: testing garbage file to be deleted

paperclover:
> what about UNC paths

a unc path does not have a colon

\\\\server\share\path

PR #8486, opened 2024-01-25
Fix type declaration for `Array.fromAsync`

### LINKED ISSUE #8484: [Types] `Array.fromAsync` mistyped

1.0.25+a8ff7be64

Use the method `Array.fromAsync` in some TS file with 

```ts
async function* naturals() {
		yield i;
	}
}

const test1 = await Array.fromAsync(
	naturals(),
	(n) => Promise.resolve(`${n}`)
);

const test2 = await Array.fromAsync(
	[Promise.resolve(1), Promise.resolve(2)]
);
```

Now, use the language server to see the types of `test1` and `test2`.

The type of `test1` should be `string[]`, and the type of `test2` should be `number[]`.

### What do you see instead?

The type of `test1` is `number[]` and the type of `test2` is `Promise<number>[]`.

image
(screenshot from vscode with inlay hints enabled)

### Additional information

The implementation works, the only thing that needs fixing is the type declaration.

It is declared in `globals.d.ts` (in bun-types) like so:
```ts
interface ArrayConstructor {
  fromAsync<T>(
    asyncItems: AsyncIterable<T> | Iterable<T> | ArrayLike<T>,
    mapfn?: (value: any, index: number) => any,
    thisArg?: any,
  ): Promise<T[]>;
}
```

I think it should be like that instead:
```ts
interface ArrayConstructor {
  fromAsync<T>(
    asyncItems: AsyncIterable<T> | Iterable<T> | ArrayLike<T>,
  ): Promise<Awaited<T>[]>;
  fromAsync<T, U>(
    asyncItems: AsyncIterable<T> | Iterable<T> | ArrayLike<T>,
    mapFn?: (value: T, index: number) => U,
    thisArg?: any,
  ): Promise<Awaited<U>[]>;
}
```

For more information about this method, read the mdn doc https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync.

May I open a PR to fix the declaration and add tests?

A-D-E-A:
### What does this PR do?

Fixes the type declaration for `Array.fromAsync` and doc comments.
Fixes #8484.

Also, the parameter names were changed to match the mdn documentation (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync).

- [ ] Code changes

I added type tests in `array.test.ts` in the `bun-types` package.


jdalton:
I take for granted these in VSCode. Isn't there a built-in definition for JS in general or do they usually roll them into the Node definitions? I feel like for language level things we should defer to whatever the common one/used-thing is.

A-D-E-A:
Since neither TypeScript nor Node implement `Array.fromAsync`, I searched for the `core-js` implementation. The definitely typed package does not define the function but defines `Array.from` which is heavily related.

The new type is basically that one, but with the addition of `AsyncIterator<T>` to the first parameter, and the return type is in a `Promise`.

Electroid:
> Since neither TypeScript nor Node implement `Array.fromAsync`, I searched for the `core-js` implementation. The definitely typed package does not define the function but defines `Array.from` which is heavily related.
> 
> The new type is basically that one, but with the addition of `AsyncIterator<T>` to the first parameter, and the return type is in a `Promise`.

@A-D-E-A It is relatively new: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

Electroid:
Occasionally, we implement new JavaScript APIs before they appear in TS/node defaults. In these cases, we need to define them.

jdalton:
Thanks for digging in all and @A-D-E-A for adding a test case too!

PR #8509, opened 2024-01-26
[windows] nodefs

gvilums:
### What does this PR do?
This PR implements missing functionality from nodefs for windows.
A big change is the update to `KeepAlive`: Instead of using libuv's `async` handles, we simply change the reference count directly on the loop. This means that we don't have to worry about the lifetime of the memory where the `async` handle resides.


src/sys.zig
```diff
@@ -349,42 +349,48 @@ pub fn getErrno(rc: anytype) bun.C.E {
...
 const O = std.os.O;
 const w = std.os.windows;
 
+fn normalizePathWindows(
+    dir_fd: bun.FileDescriptor,
+    path: []const u8,
+    buf: *bun.WPathBuffer,
+) Maybe([:0]const u16) {
+    // TODO this check can be more sophisticated: paths like C:\bun\hello.txt are fine, even though they contain a dot
+    if (!bun.strings.containsChar(path, '.')) {
```
Jarred-Sumner: I think this needs to be more sophisticated before we can realistically ship this PR. This PR will add too many  system calls.

```suggestion
    if (bun.strings.containsRelativePath(path)) {
```

Something like this:
```zig
pub fn  containsRelativePath(input: []const u8) bool {
   const dot = strings.indexOfCharUsize(input, '.') orelse return false;
   // "."
   // "/."
   // ".."
   if (dot == 0 or dot == input.len - 1)
     return true;
 // - "/foo\bar/../"
 // - "/foo\bar/."
 // - "/../"
 // - "/foo/bar/baz/../"
  if (isAnySep(input[dot - 1]) 
     return true;

     return false;
}
```

PR #7912, opened 2023-12-29
feat(build): adjust arch linux auto detect to include manjaro

knightspore:
### What does this PR do?

Following on progress from issue #2664 and corresponding PR #6835, I noticed `bun setup` / `bun run build` were failing on a Manjaro machine for the same reason that Manjaro uses the same suffix as Arch for `libatomic`.

Currently, `CMakeLists.txt` looks for a `NAME` property in `/etc/os-release` matching `"Arch Linux"`. 

This PR introduces changes to the code to include matches for `"(Arch|Manjaro) Linux"`.
 
- [X] Code changes

### How did you verify your code works?

Re-built on Arch & Manjaro, read code in `CMakeLists.txt`.

PR #8510, opened 2024-01-26
feat(build): adjust arch/manjaro linux auto detect to include artix

BeyondMagic:
### What does this PR do?

Following on progress from PR #7912, in which is noted  that `bun setup`/`bun run runb` fails on a Arch Linux distributions for libatomic.

Currently, CMakeLists.txt looks for a NAME property in /etc/os-release matching "Arch Linux" or "Manjaro Linux".

This PR introduces changes to the code to include matches for also Artix Linux..

- [x] Code changes.

Re-built on Arch & Manjaro in VMs, read code in CMakeLists.txt.

### Note:

Maybe on the future we could match this without specifying Linux distribution before it turns into a big mess.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,7 +233,7 @@ endif()
 if(UNIX AND NOT APPLE)
     execute_process(COMMAND cat /etc/os-release COMMAND head -n1 OUTPUT_VARIABLE LINUX_DISTRO)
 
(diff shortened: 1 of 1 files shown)


paperclover:
i wish this didnt have to be a continuously growing allowlist

but i remember it wasnt trivial to detect static linking possiblity in CMake, so i will let this slice.

thank you

PR #8511, opened 2024-01-27
Improve `.vscode` extensions, tasks, and configuration

Electroid:
### What does this PR do?

- Removed unused prettier stuff
- Added more configurations to `launch.json`, including:
  - `--inspect` that auto-opens the inspector in the browser
  - More support for Windows
  - Removes older configurations that were unused 
- Organized `extensions.json` and added more recommended extensions
- Added to `tasks.json` for building Bun


.vscode/settings.json
```diff
@@ -1,66 +1,107 @@
...
-  // We are using biome instead of prettier.
-  "prettier.enable": false,
-
-  "[json]": {
-    "editor.defaultFormatter": "biomejs.biome"
-  },
-  "[jsonc]": {
-    "editor.defaultFormatter": "biomejs.biome"
-  },
+  // "zig.path": "${workspaceFolder}/.cache/zig/zig.exe",
```
paperclover: why is this commented out?

.vscode/settings.json
```diff
@@ -87,33 +128,6 @@
...
-  "[cpp]": {
-    "editor.defaultFormatter": "xaver.clang-format"
-  },
-  "[h]": {
-    "editor.defaultFormatter": "xaver.clang-format"
-  },
-  "[c]": {
-    "editor.defaultFormatter": "xaver.clang-format"
-  },
   "files.associations": {
```
paperclover, line 132: if youre fixing this file, i think almost everything in `files.associations` is useless and should be removed if not needed

scripts/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+ninja -C build
```
paperclover: this script has no purpose and will confuse people in it's current state

- see `package.json`'s `build` script.
- does not ensure the current working directory (see [every other shell script])
- does not ensure the build setup is actually set up

.vscode/extensions.json
```diff
@@ -1,10 +1,35 @@
 {
   "recommendations": [
+    // Zig
     "ziglang.vscode-zig",
-    "biomejs.biome",
+
+    // C/C++
+    "ms-vscode.cpptools",
```
paperclover: why is ms-vscode.cpptools and ms-vscode.cpptools-extension-pack here. is `clangd` not a replacement for the both of these.

paperclover (changes requested):
see above comments

paperclover:
also, in my opinion, adding more things to `launch.json` makes the debugging experience worse. if this file was revamped, it should have been code-generated as a part of the automatic setup process; this way we can include only the windows actions on windows, and the posix actions on posix.

similar feelings to the recommended extensions but to a lesser extent. LLDB is not needed on Windows as we don't use it there. this would lead to installing random extensions without reason (extensions add up and hurt development machine performance, but it's probably fine for this)

scripts/build.sh
Electroid: It's purpose is to mirror `build.ps1`

```json
    {
      "type": "process",
      "label": "Build Bun",
      "dependsOn": ["Setup Environment"],
      "command": "scripts/build.sh",
      "windows": {
        "command": "scripts/build.ps1"
      },
    }
```

Electroid: Going to remove for now though

Issue #8197, opened 2024-01-16
windows cannot run scripts if the environment contains invalid UTF-8


djkloop:
### What version of Bun is running?

1.0.23-canary.48+407f42718

### What platform is your computer?

![image](https://github.com/oven-sh/bun/assets/21050914/5cb3a664-8621-4fa4-864f-95156f275a29)

### What steps can reproduce the bug?

 ```sh
 "scripts": {
    "serve": "node --max_old_space_size=10096 node_modules/@vue/cli-service/bin/vue-cli-service.js serve"
}
```
![image](https://github.com/oven-sh/bun/assets/21050914/f1118ed5-5b5f-4e64-9011-77303c35e422)

### What is the expected behavior?

normal operation

### What do you see instead?

error

**Comment by Jarred-Sumner** [comment]
strange

**Comment by paperclover** [comment]
i couldn't reproduce this easily with the package.json script alone.

**Comment by kafaa-dev** [comment]
getting this error too on version `1.0.25-canary.16+a8ff7be64`, I'm using the baseline build

**Comment by djkloop** [comment]
![image](https://github.com/oven-sh/bun/assets/21050914/e3dcaa83-75e8-4baa-a0c9-11f717a29b5d)


```sh
# https://nuxt.com/docs/getting-started/installation
# The default command for creating a new project using nuxt is also not working
```

paperclover:
funny how this happens to starship prompt users

PR #8521, opened 2024-01-27
windows: some random things

### LINKED ISSUE #8045: windows: bunx / bun create is not implemented
Hello , I have this problem when I create a project next js
bun create-next-app or bunx create-next-app@Latest

_Originally reported on Discord: [`problem when creating Next js project with bun`](https://discord.com/channels/876711213126520882/1193952433186033694/1193952433186033694)_

paperclover:
- Fixes #8045
- Furthers #8197, but invalid utf8 in the env will omit those variables. left a todo comment.
- Wrote the glue to allow Bun Shell in package.json scripts. It is not enabled by default but only if you use --no-native-shell. Once zacks shell branch is merge we will flip to --native-shell to opt out of this behavior.
- Fix running bin files that are three or less characters.


paperclover:
it does not compile on posix, ow

Issue #7418, opened 2023-12-02
Zstandard is in the source code, but does not exist any doc about it


jefer94:
### What is the type of issue?

Documentation is missing

### What is the issue?

There have Zstandard referenced in many files, but it's not documented yet
![image](https://github.com/oven-sh/bun/assets/3018142/b8f5d247-d234-4268-b518-6d5afc40f9df)

This should be useful to optimize the deployments in the frontend 

### Where did you find it?

https://bun.sh/guides
https://bun.sh/docs

Jarred-Sumner:
indeed, it is only currently used with `bun build --compile --sourcemap` to compress the sourcemaps

jefer94:
Should be great to have a module to use the same implementation within Javascript, and it should be a replacement to Brotli until this is implemented, the good part of use Brotli or Zstandard if that I have implemented both within a cache in Redis, but maybe I should look for another compression algorithm, all packages related to them looks abandoned, whatever currently Zstd is experimental in Blink browsers yet

L3P3:
I would have prefered namespaces but as we already have `Bun.gzipSync`, there could just as well be a `Bun.zstdSync`. Maybe this will be my first contribution, never worked with zig however. :sweat_smile:
Even though it could be handy, I would not overestimate its usage. In productive environments, I highly recommend using nginx as a reverse proxy and bun behind it. Or are there benchmarks speaking against it?

bompus:
I saw mention of `Bun.zstdSync`, so figured I would mentiond that node has a [PR](https://github.com/nodejs/node/pull/52100) pending to add this, and it'd be fantastic to have it in Bun natively, rather than needing to import a 3rd-party npm n-api implementation for this functionality.

If the compressor/decompressor logic is already in source, what would it take to expose it to javascript-land?

L3P3:
Not much, mostly wrapper code, tests (!) and updating documentation.
But I think before we should clarify if we keep n compression algorithms directly in Bun namespace or do it better for newly added ones.

jefer94:
Right now `zstd` is supported for the browsers, I am not affected by this because I use Next in Vercel, and Vercel uses Node

L3P3:
zstd is now also in node:zlib, so it will be usable anyways.

Jarred-Sumner:
Zstd decompression and compression is exposed as of [Bun v1.2.17](https://bun.sh/blog/bun-v1.2.17#zstandard-zstd-support-in-node-zlib)

PR #8528, opened 2024-01-28
feat: add support for zstd

theoparis:
### What does this PR do?

This allows custom npm registries and git dependencies to take advantage of compression formats such as zstd.
In BunOS this means large packages can be downloaded using bun with a tar.zst extension and a custom npm registry. An example of this is LLVM (4gb uncompressed).

I'm not sure that this is the right approach (in terms of removing the manual zlib decompression) but since libarchive handles the compression it works fine.

- [x] Code changes

I created an minimal package.json and ran `bun add -d @types/bun` and it still works properly. Additionally, running a copy of https://github.com/bunos2/registry/commit/4a7e9491a9dd90530d0112d1fc696003cec205f2 and running `bun add @bruh/bruh` with a bunfig.toml of the following works fine as well:
```toml
[install.scopes]
"@bruh" = "http://localhost:6969"
```
Adding `@bruh/bruh` should add glibc to `node_modules/@bruh/bruh` with a lib/ directory/ containg libc.a.

## TODO

- [x] Add zstd to http.zig
- [ ] Update http.zig to add zstd to Transfer/Content-Encoding
- [x] Add zstd to libarchive.zig
- [ ] Add tests
- [ ] Add zstd to the dependency build scripts (change the order for libarchive support)

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -3315,6 +3315,49 @@ declare module "bun" {
      */
     static readonly byteLength: 32;
   }
+  
+  /** Compression options for `Bun.deflateSync` and `Bun.gzipSync` */
+  interface ZstdCompressionOptions {
+    /**
+     * The compression level to use. Must be between `-1` and `22`.
+     * - A value of `-1` uses the default compression level (Currently `3`)
+     * - A value of `0` gives no compression
+     * - A value of `1` gives least compression, fastest speed
+     * - A value of `19` gives best compression, slowest speed
+     * - A value of `20`-`22` gives the maxium compression, greater memory usage
+     */
+    level?: -1 | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20 | 21 | 22;
+
+    /**
+     * The base 2 logarithm of the window size (the size of the history buffer).
+     *
+     * Larger values of this parameter result in better compression at the expense of memory usage.
+     */
+    windowLog?:
+      | 10
+      | 11
+      | 12
+      | 13
+      | 14
+      | 15
+      | 16
+      | 17
+      | 18
+      | 19
+      | 20
+      | 21
+      | 22
+      | 23
+      | 24
+      | 25
+      | 26
+      | 27
+      | 28
+      | 29
+      | 30
+      | 31;
+    /**
+  }
 
   /** Compression options for `Bun.deflateSync` and `Bun.gzipSync` */
   interface ZlibCompressionOptions {
@@ -3412,6 +3455,20 @@ declare module "bun" {
    * @returns The output buffer with the decompressed data
    */
   function gunzipSync(data: Uint8Array | string | ArrayBuffer): Uint8Array;
+
+  /**
+   * Decompresses a chunk of data with the `zstd` algorithm.
+   * @param data The buffer of data to decompress
+   * @returns The output buffer with the decompressed data
+   */
+  function zstdDecompressSync(data: Uint8Array, options?: ZstdCompressionOptions): Uint8Array;
+
+  /**
+   * Decompresses a chunk of data with the `zstd` algorithm.
+   * @param data The buffer of data to decompress
+   * @returns The output buffer with the decompressed data
+   */
+  function zstdCompressSync(data: Uint8Array, options?: ZstdCompressionOptions): Uint8Array;
 
   type Target =
     /**
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -3356,7 +3356,6 @@ declare module "bun" {
       | 29
       | 30
       | 31;
-    /**
   }
 
   /** Compression options for `Bun.deflateSync` and `Bun.gzipSync` */
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -3315,7 +3315,7 @@ declare module "bun" {
      */
     static readonly byteLength: 32;
   }
-  
+
   /** Compression options for `Bun.deflateSync` and `Bun.gzipSync` */
   interface ZstdCompressionOptions {
     /**
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -17,6 +17,8 @@ pub const BunObject = struct {
     pub const generateHeapSnapshot = Bun.generateHeapSnapshot;
     pub const gunzipSync = JSC.wrapStaticMethod(JSZlib, "gunzipSync", true);
     pub const gzipSync = JSC.wrapStaticMethod(JSZlib, "gzipSync", true);
+    pub const zstdDecompressSync = JSC.wrapStaticMethod(JSZstd, "zstdDecompressSync", true);
+    pub const zstdCompressSync = JSC.wrapStaticMethod(JSZstd, "zstdCompressSync", true);
     pub const indexOfLine = Bun.indexOfLine;
     pub const inflateSync = JSC.wrapStaticMethod(JSZlib, "inflateSync", true);
     pub const jest = @import("../test/jest.zig").Jest.call;
@@ -130,6 +132,8 @@ pub const BunObject = struct {
         @export(BunObject.generateHeapSnapshot, .{ .name = callbackName("generateHeapSnapshot") });
         @export(BunObject.gunzipSync, .{ .name = callbackName("gunzipSync") });
         @export(BunObject.gzipSync, .{ .name = callbackName("gzipSync") });
+        @export(BunObject.zstdDecompressSync, .{ .name = callbackName("zstdDecompressSync") });
+        @export(BunObject.zstdCompressSync, .{ .name = callbackName("zstdCompressSync") });
         @export(BunObject.indexOfLine, .{ .name = callbackName("indexOfLine") });
         @export(BunObject.inflateSync, .{ .name = callbackName("inflateSync") });
         @export(BunObject.jest, .{ .name = callbackName("jest") });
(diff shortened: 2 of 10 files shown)


Jarred-Sumner:
This is cool, though I think it'd be better to support zstd in `http.zig` instead

What is the binary size impact of this change? Does it cause Bun to dynamically link any new libraries? We want to avoid linking new libraries into Bun and we don't want to increase the binary size further

theoparis:
The dynamic libraries should depend on what the libarchive configure command in scripts/build-libarchive.{sh,ps1} contains. So I'm pretty sure it would just be zlib and zstd if libarchive is built with the official zlib and zstd versions supported by Bun.

I forgot to add `--with-zlib` and `--with-zstd` to the command though. These would point to the zstd and zlib path from the other build scripts in scripts/. The build order would be need to be changed so that it builds  zlib and zstd before libarchive in all-dependencies.

paperclover:
> Does it cause Bun to dynamically link any new libraries?

We already link zstd statically. I cannot imagine this patch to increase the binary too much, but it would be nice to have stats on this.

jdalton:
Chromium is getting zstd: (transitioned from experiment to ship)
https://twitter.com/intenttoship/status/1757200615707488656

src/http.zig
```diff
@@ -1601,8 +1625,8 @@ const connection_closing_header = picohttp.Header{ .name = "Connection", .value
 const accept_header = picohttp.Header{ .name = "Accept", .value = "*/*" };
 
 const accept_encoding_no_compression = "identity";
-const accept_encoding_compression = "gzip, deflate, br";
-const accept_encoding_compression_no_brotli = "gzip, deflate";
+const accept_encoding_compression = "gzip, deflate, ztd, br";
```
paperclover: is the ordering of this relevant for anything?

Jarred-Sumner:
yes, it says "which is the preferred encoding?"

src/install/extract_tarball.zig
```diff
@@ -203,22 +203,7 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
         }
 
         const Archive = @import("../libarchive/libarchive.zig").Archive;
-        const Zlib = @import("../zlib.zig");
-        var zlib_pool = Npm.Registry.BodyPool.get(default_allocator);
-        zlib_pool.data.reset();
-        defer Npm.Registry.BodyPool.release(zlib_pool);
 
-        var zlib_entry = try Zlib.ZlibReaderArrayList.init(tgz_bytes, &zlib_pool.data.list, default_allocator);
```
theoparis, line 226: I removed this because libarchive already handles the compression, and it allows me to add zstd to libarchive.zig.

paperclover:
> What is the binary size impact of this change?

confirmed quite small

src/http.zig
jdalton:
```suggestion
const accept_encoding_compression = "gzip, deflate, zstd, br";
```

paperclover:
Note: this fixes #7418, potentially others.

packages/bun-types/bun.d.ts
```diff
@@ -3304,6 +3304,48 @@ declare module "bun" {
...
 
+  /** Compression options for `Bun.deflateSync` and `Bun.gzipSync` */
+  interface ZstdCompressionOptions {
+    /**
+     * The compression level to use. Must be between `-1` and `22`.
+     * - A value of `-1` uses the default compression level (Currently `3`)
+     * - A value of `0` gives no compression
+     * - A value of `1` gives least compression, fastest speed
+     * - A value of `19` gives best compression, slowest speed
+     * - A value of `20`-`22` gives the maxium compression, greater memory usage
```
argosphil, line 3327: nit: "maximum"

packages/bun-types/bun.d.ts
```diff
@@ -3401,6 +3443,20 @@ declare module "bun" {
...
+   * Decompresses a chunk of data with the `zstd` algorithm.
+   * @param data The buffer of data to decompress
+   * @returns The output buffer with the decompressed data
+   */
+  function zstdDecompressSync(data: Uint8Array, options?: ZstdCompressionOptions): Uint8Array;
+
+  /**
+   * Decompresses a chunk of data with the `zstd` algorithm.
+   * @param data The buffer of data to decompress
+   * @returns The output buffer with the decompressed data
```
argosphil, line 3468: nit: "decompress" -> "compress" in all three lines.

src/bun.js/api/BunObject.zig
```diff
@@ -5297,6 +5302,100 @@ comptime {
...
+            if (options_val.isObject()) {
+                if (options_val.get(globalThis, "windowLog")) |window| {
+                    opts.window_log_max = window.coerce(i32, globalThis);
+                }
+            }
+        }
+
+        const compressed = buffer.slice();
+        const allocator = JSC.VirtualMachine.get().allocator;
+        var list = std.ArrayListUnmanaged(u8).initCapacity(allocator, if (compressed.len > 512) compressed.len else 32) catch unreachable;
```
argosphil, line 5143: nit: why the two different limits? It seems to me that compressed.len between 32 and 512 is a potentially common case and we should use compressed.len as capacity then rather than 32.

src/bun.js/api/BunObject.zig
```diff
@@ -5297,6 +5302,100 @@ comptime {
...
+
+                if (options_val.get(globalThis, "level")) |level| {
+                    opts.compression_level = level.coerce(i32, globalThis);
+                }
+            }
+        }
+
+        const compressed = buffer.slice();
+        const allocator = JSC.VirtualMachine.get().allocator;
+        var list = std.ArrayListUnmanaged(u8).initCapacity(allocator, if (compressed.len > 512) compressed.len else 32) catch unreachable;
```
argosphil, line 5185: nit: as above

src/bun.js/api/BunObject.zig
```diff
@@ -5297,6 +5302,100 @@ comptime {
...
+                    opts.window_log_max = window.coerce(i32, globalThis);
+                }
+
+                if (options_val.get(globalThis, "level")) |level| {
+                    opts.compression_level = level.coerce(i32, globalThis);
+                }
+            }
+        }
+
+        const compressed = buffer.slice();
```
argosphil, line 5183: we shouldn't call this "compressed", since it's not.

src/bun.js/bindings/BunObject.cpp
```diff
@@ -594,6 +594,8 @@ JSC_DEFINE_HOST_FUNCTION(functionHashCode,
     getImportedStyles                              BunObject_callback_getImportedStyles                                DontEnum|DontDelete|Function 1
     gunzipSync                                     BunObject_callback_gunzipSync                                       DontDelete|Function 1
     gzipSync                                       BunObject_callback_gzipSync                                         DontDelete|Function 1
+    zstdCompressSync                                       BunObject_callback_zstdCompressSync                                         DontDelete|Function 1
+    zstdDecompressSync                                       BunObject_callback_zstdDecompressSync                                         DontDelete|Function 1
```
argosphil, line 563: nit: whitespace here is weird

src/http.zig
```diff
@@ -1301,6 +1315,14 @@ const Decompressor = union(enum) {
                 reader.list = body_out_str.list;
                 reader.total_out = @truncate(initial);
             },
+            .zstd => |reader| {
+                reader.input = buffer;
+                reader.total_in = @as(u32, @truncate(buffer.len));
```
argosphil, line 1335: bug: this should be `reader.total_in = 0`.

src/zstd.zig
```diff
@@ -0,0 +1,254 @@
...
+
+            // https://github.com/google/brotli/blob/fef82ea10435abb1500b615b1b2c6175d429ec6c/go/cbrotli/reader.go#L15-L27
+            const result = c.ZSTD_compressStream(
+                this.zstd,
+                &out_buffer,
+                &in_buffer,
+            );
+
+            const bytes_written = unused_capacity.len -| unused_capacity.len;
+            const bytes_read = next_in.len -| next_in.len;
```
argosphil, line 228: bug: this needs to be `bytes_written = out_buffer.pos`, `bytes_read = in_buffer.pos`, I think.

src/zstd.zig
```diff
@@ -0,0 +1,254 @@
...
+                .size = unused_capacity.len,
+                .pos = this.total_out,
+            };
+
+            // https://github.com/google/brotli/blob/fef82ea10435abb1500b615b1b2c6175d429ec6c/go/cbrotli/reader.go#L15-L27
+            const result = c.ZSTD_compressStream(
+                this.zstd,
+                &out_buffer,
+                &in_buffer,
+            );
```
argosphil, line 225: bug: we need to call ZSTD_flushStream when we're done, right?

src/zstd.zig
```diff
@@ -0,0 +1,254 @@
...
+
+            // https://github.com/google/brotli/blob/fef82ea10435abb1500b615b1b2c6175d429ec6c/go/cbrotli/reader.go#L15-L27
+            const result = c.ZSTD_decompressStream(
+                this.zstd,
+                &out_buffer,
+                &in_buffer,
+            );
+
+            const bytes_written = unused_capacity.len -| unused_capacity.len;
+            const bytes_read = next_in.len -| next_in.len;
```
argosphil, line 112: bug: this needs to be `bytes_written = out_buffer.pos`, `bytes_read = in_buffer.pos`, I think.

argosphil (commented):
Sorry for the drive-by review, this is mostly to tell you that I've had to change two minor things to get the test you pasted in Discord to pass, and there are also some nits. Looks great overall though!

theoparis:
I see that you closed this PR. I still have not figured out why it returned an empty buffer 😕  No one from the bun discord was able to help me solve it, so I forgot about the PR entirely....

L3P3:
@theoparis you got corrections in the PR here but have not implemented/pushed them yet.

theoparis:
I actually tried to implement the corrections but it didn't solve the issue I mentioned above so the entire PR wasn't functional.

PR #8537, opened 2024-01-28
docs: remove unnecessary await in lines() example

whygee-dev:
### What does this PR do?

Removes unnecessary await in Bun Shell .lines() example


Jarred-Sumner:
Thank you

PR #8540, opened 2024-01-28
docs: non-null assertion is needed on document.getElementById

DaleSeo:
### What does this PR do?

`document.getElementById` can returns `null`, but `ReactDOM.createRoot` doesn't accept `null`.

- [ ] Code changes

### DIFF
--- a/docs/bundler/index.md
+++ b/docs/bundler/index.md
@@ -43,7 +43,7 @@ Let's build our first bundle. You have the following two files, which implement 
 import * as ReactDOM from 'react-dom/client';
 import {Component} from "./Component"
 
-const root = ReactDOM.createRoot(document.getElementById('root'));
+const root = ReactDOM.createRoot(document.getElementById('root')!);
 root.render(<Component message="Sup!" />)
 ```
 


Didas-git:
Non null assertion is one of the worst things that you can recommend.
Instead the example should be changed to `jsx` and ts developers should figure it out on their own.

DaleSeo:
@Didas-git, I appreciate your feedback. However, I'm not sure if updating the code snippet to use JSX is the best option. One of the selling points of Bun is its ability to run TSX out of the box. I'm under the impression that that's why other examples in the documentation are written in TSX.

Not a big fan of non-null assertion myself, but I believe this is one of the rare valid use cases of it because we can guarantee that there is an element with ﻿id="root" in ﻿index.html for a typical React app. You will see the same approach taken in other React tempaltes such as Vite: https://github.com/vitejs/vite/blob/c3e83bb078e84a8a2d377455801b2a689557763e/packages/create-vite/template-react-ts/src/main.tsx#L6

Please let me know what you think. If you still believe we should change it to JSX. I can make the change. I don't have strong opionons on it. I just want the little type error to be removed from the example code.

Didas-git:
JSX is the best option because there are no types being declared in the snippet.
Non null assertion is also not valid in this case, it can return undefined so you should check for it, as for what others do it doesnt matter, promoting bad practices should not be done no matter if x or y do it.

jcarpe:
FWIW - The paradigm of this entire docs page for `build` uses `.tsx` as input and `.js` as output.

paperclover:
it should not be changed to jsx. 

> check for it,

@Didas-git what is an application supposed to do when it is unable to mount it's UI onto the page? display fallback UI, where? the application is doomed if this element doesnt exist, with or without the assertion.

paperclover (approved):

Didas-git:
> it should not be changed to jsx.
> 
> > check for it,
> 
> @Didas-git what is an application supposed to do when it is unable to mount it's UI onto the page? display fallback UI, where? the application is doomed if this element doesnt exist, with or without the assertion.

The `root` element can be anything, the name is just a convention.
If it returns `null` you can literally write 2 lines of code to create it and then create the react root on it

DaleSeo:
https://github.com/facebook/create-react-app/blob/0a827f69ab0d2ee3871ba9b71350031d8a81b7ae/packages/cra-template-typescript/template/src/index.tsx#L7-L9

I discovered that the TypeScript template of CRA (Create React App) also assumes that `document.getElementById('root')` will never be undefined. This suggests that using type assertion is a common convention in TypeScript when mouting a React component onto an HTML element.

PR #8543, opened 2024-01-28
Update .prettierrc.cjs

huseeiin:
removed props that are set to their default values in prettier

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
we are no longer using prettier, this file should be deleted

PR #8547, opened 2024-01-29
Update vite.md

alissonmdo:
### What does this PR do?

Updates the example on the vite page, since the current examples throws an error when trying to replace node with bun, the solution is replacing `bunx` with `bun --bun x`

The original command `bunx vite` throws:

```
CACError: Unknown option `--bun`
```

- [ ] Code changes

### DIFF
--- a/docs/guides/ecosystem/vite.md
+++ b/docs/guides/ecosystem/vite.md
@@ -28,11 +28,11 @@ bun install
 
 ---
 
-Start the development server with the `vite` CLI using `bunx`.
+Start the development server with the `vite` CLI using `bun x`.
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
This appears to be an issue on your machine. `bunx` is the recommended approach, it is a symlink to `bun`. I'm not sure what `bunx` is installed as on your machine though

alissonmdo:
Yeah, I guess it is a problem here. I saw a previous discussion suggesting this and it worked here (as it did for them), but I was using `asdf` for managing `bun` versions and now I see it is a problem with their bun plugin. Sorry for bothering, I'll bring the discussion to the plugin repo.

Installing via the official script fixes the problem on my machine. Thanks Jarred! Keep up the great work.

PR #8557, opened 2024-01-29
fix(builtin-bundler): do not replace globals when extending classes

paperclover:
### What does this PR do?

Fixes incorrect builtins codegen:

`class X extends Promise {` no longer bundles to `class X extends @Promise {`

### DIFF
--- a/src/codegen/replacements.ts
+++ b/src/codegen/replacements.ts
@@ -54,6 +54,8 @@ export const globalsToPrefix = [
   "undefined",
 ];
 
+replacements.push({ from: new RegExp(`\\bextends\\s+(${globalsToPrefix.join("|")})`, "g"), to: "extends __no_intrinsic__%1" });
+
 // These enums map to $<enum>IdToLabel and $<enum>LabelToId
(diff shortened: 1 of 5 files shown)


src/codegen/replacements.ts
```diff
@@ -54,6 +54,11 @@ export const globalsToPrefix = [
   "undefined",
 ];
 
+replacements.push({
+  from: new RegExp(`\\bextends\\s+(${globalsToPrefix.join("|")})`, "g"),
+  to: "extends __no_intrinsic__%1",
```
jdalton, line 59: Is `__no_intrinsic__` a placeholder to swap things back or does it tie-into plucked references like a stored plucked reference to the `Promise` constructor?

Fun fact, the `Promise` that results from an `await` will not use `window.Promise` in non-Chromium environments and instead use the reference `%Promise%` which is the default value of `window.Promise`. So if you overwrite `window.Promise` it sill uses `%Promise%.

paperclover: its a workaround for the fact `--define Promise=__intrinsic__Promise` is in place.

PR #8560, opened 2024-01-29
get more bundler tests passing on windows

nektro:


src/bundler/bundle_v2.zig
```diff
@@ -11721,3 +11725,10 @@ fn targetFromHashbang(buffer: []const u8) ?options.Target {
 
     return null;
 }
+
+fn platformToUnixInPlace(path_buffer: []u8) void {
+    if (std.fs.path.sep == '/') return;
+    for (path_buffer) |*c| {
```
Jarred-Sumner: please use `strings.indexOfChar` here to do this

it will use SIMD instead of probably scalar code

src/bun.js/node/node_fs.zig
```diff
@@ -5277,9 +5277,30 @@ pub const NodeFS = struct {
         const fd = switch (args.file) {
             .path => brk: {
                 path = args.file.path.sliceZ(pathbuf);
+                bun.path.posixToPlatformInPlace(@constCast(path));
+
+                var is_dirfd_different = false;
+                const dirfd = if (!Environment.isWindows) args.dirfd else blk: {
+                    if (std.mem.startsWith(u8, path, "..\\")) {
```
Jarred-Sumner: This isn’t robust enough + we have similar code in-place. I’m on mobile rn but if you look in strings_immutable, @gvilums added a function for “is this possibly a relative path”?

src/bun.js/node/node_fs.zig
```diff
@@ -5277,9 +5277,30 @@ pub const NodeFS = struct {
...
+                        is_dirfd_different = true;
+                        var buffer = std.mem.zeroes([std.fs.MAX_PATH_BYTES - 1:0]u8);
+                        const dirfd_path_len = bun.windows.GetFinalPathNameByHandleA(args.dirfd.cast(), &buffer, std.fs.MAX_PATH_BYTES, .NORMALIZED);
+                        const dirfd_path = buffer[0..dirfd_path_len];
+                        const parent_path = std.fs.path.dirname(dirfd_path).?;
+                        const newdir = std.fs.cwd().openDir(parent_path, .{}) catch unreachable;
```
Jarred-Sumner: Please don’t use unreachable on a system call.

src/bun.js/node/node_fs.zig
```diff
@@ -5277,9 +5277,30 @@ pub const NodeFS = struct {
         const fd = switch (args.file) {
             .path => brk: {
                 path = args.file.path.sliceZ(pathbuf);
+                bun.path.posixToPlatformInPlace(@constCast(path));
+
+                var is_dirfd_different = false;
+                const dirfd = if (!Environment.isWindows) args.dirfd else blk: {
+                    if (std.mem.startsWith(u8, path, "..\\")) {
+                        is_dirfd_different = true;
+                        var buffer = std.mem.zeroes([std.fs.MAX_PATH_BYTES - 1:0]u8);
```
Jarred-Sumner: This is potentially a 64 KB buffer. We don’t need to zero the memory. So let’s just not.

src/bundler/bundle_v2.zig
```diff
@@ -9027,11 +9025,18 @@ const LinkerContext = struct {
...
+                bun.path.platformToPosixInPlace(rel_path);
                 if ((try path_names_map.getOrPut(rel_path)).found_existing) {
                     try c.log.addErrorFmt(null, Logger.Loc.Empty, bun.default_allocator, "Multiple files share the same output path: {s}", .{rel_path});
                     return error.DuplicateOutputPath;
                 }
-
+                // resolve any /./ and /../ occurrences
+                // use resolvePosix since we asserted above all seps are '/'
+                if (std.mem.indexOf(u8, rel_path, "/./") != null) {
+                    chunk.final_rel_path = std.fs.path.resolvePosix(c.allocator, &.{rel_path}) catch unreachable;
```
Jarred-Sumner: Please use Bun’s path resolve function instead of Zig’s. Ours handles many more cases than theirs and is more well tested.

src/options.zig
```diff
@@ -2588,7 +2588,13 @@ pub const PathTemplate = struct {
     }
 
     pub fn format(self: PathTemplate, comptime _: []const u8, _: std.fmt.FormatOptions, writer: anytype) !void {
-        var remain = self.data;
+        var pathbuf: [std.fs.MAX_PATH_BYTES]u8 = undefined;
+        var remain = pathbuf[0..self.data.len];
+        @memcpy(remain, self.data);
+        for (remain) |*c| {
```
Jarred-Sumner: Can you use your posixToPlatformInPlace function?

src/windows.zig
```diff
@@ -3018,3 +3018,10 @@ pub extern "kernel32" fn GetTempPathW(
     nBufferLength: DWORD, // [in]
     lpBuffer: LPCWSTR, // [out]
 ) DWORD;
+
+pub extern "kernel32" fn GetFinalPathNameByHandleA(
```
Jarred-Sumner: Please don’t add another symbol import for something we already do using the W version. Two reasons why:

- Every symbol we add makes Bun start slower.
- Our UTF16 -> WTF-8 conversion code is faster than the Windows API.

Jarred-Sumner (changes requested):
See comments

src/bun.js/node/node_fs.zig
nektro: in this location they're all relative paths but I'm only fixing the ones that specifically start with upwards directory traversal

src/options.zig
nektro: ah yep, wrote that block before I added it and forgot to come back

src/bun.js/node/node_fs.zig
Jarred-Sumner: What if you did `./foo/../bar`?

Jarred-Sumner: Or is that not going to happen for some reason?

src/bun.js/node/node_fs.zig
```diff
@@ -5277,9 +5277,35 @@ pub const NodeFS = struct {
         const fd = switch (args.file) {
             .path => brk: {
                 path = args.file.path.sliceZ(pathbuf);
```
```suggestion
                // On Windows, we potentially mutate the path in posixToPlatformInPlace
                // We cannot mutate JavaScript strings in-place. That will break many things.
                // So we must always copy the path string on Windows.
                path = args.file.path.sliceZWithForceCopy(pathbuf, Environment.isWindows);
```

paperclover:
it would be nice to double check that the write file function does not fail if given a posix absolute path

`writeFile('/Users/dave/hello.txt', 'alala')` -> should write to C:\\Users\\dave\\hello.txt

src/bundler/bundle_v2.zig
```diff
@@ -9027,11 +9028,20 @@ const LinkerContext = struct {
...
+                bun.path.platformToPosixInPlace(u8, rel_path);
+
                 if ((try path_names_map.getOrPut(rel_path)).found_existing) {
                     try c.log.addErrorFmt(null, Logger.Loc.Empty, bun.default_allocator, "Multiple files share the same output path: {s}", .{rel_path});
                     return error.DuplicateOutputPath;
                 }
-
+                // resolve any /./ and /../ occurrences
+                // use resolvePosix since we asserted above all seps are '/'
+                if (std.mem.indexOf(u8, rel_path, "/./") != null) {
```
```suggestion
  if (Environment.isWIndows) {
                if (std.mem.indexOf(u8, rel_path, "/./") != null) {
```

src/bundler/bundle_v2.zig
```diff
@@ -674,6 +675,7 @@ pub const BundleV2 = struct {
             }
         }
         path.* = try path.dupeAlloc(this.graph.allocator);
+        path.pretty = bun.sliceInBuffer(path.text, path.pretty);
```
Jarred-Sumner, line 679: this really shouldn't be necessary. dupeAlloc() should be doing this already.

src/bun.zig
```diff
@@ -553,7 +554,10 @@ pub inline fn isSliceInBuffer(slice: []const u8, buffer: []const u8) bool {
 }
 
 pub inline fn sliceInBuffer(stable: string, value: string) string {
-    if (std.mem.indexOf(u8, stable, value)) |index| {
+    if (allocators.sliceRange(stable, value)) |start_len| {
+        return stable[start_len[0]..][0..start_len[1]];
```
Jarred-Sumner: `return value;`

PR #8566, opened 2024-01-30
Fixes #8555

### LINKED ISSUE #8555: Bun keeps installing package

1.0.25+a8ff7be64

(Microsoft Windows NT 10.0.19045.0 x64"Microsoft Windows NT 10.0.19045.0 x64)

setup

```sh
mkdir bun-repro && cd bun-repro
bun init -y
bun add eslint-plugin-no-relative-import-paths
```

bug

```sh
~/git/n4bb12/bun-repro $ bun i
bun install v1.0.25 (a8ff7be6)

 + eslint-plugin-no-relative-import-paths@1.5.3

 1 package installed [6.00ms]
```

bun should respond with `Checked X installs across Y packages (no changes) [Z.00ms]`

### What do you see instead?

bun keeps installing this package as if it were new

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Some packages seem to have `v` in the "version" field in package.json. Bun currently assumes the `"version"` string always matches exactly what was in the manifest.  We could make this check looser to also remove leading/trailing ASCII whitespace, but it is unnecessary work to do so. 

I didn't. Should really add a test here.

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -1062,9 +1062,28 @@ pub const PackageInstall = struct {
         _ = package_json_checker.parseExpr() catch return false;
         if (!package_json_checker.has_found_name or !package_json_checker.has_found_version or log.errors > 0) return false;
 
-        // Version is more likely to not match than name, so we check it first.
-        return strings.eql(package_json_checker.found_version, this.package_version) and
-            strings.eql(package_json_checker.found_name, this.package_name);
+        // Check if the version matches
+        if (!strings.eql(package_json_checker.found_version, this.package_version)) {
+            // If it doesn't match, check again but with a leading "v" removed
+            //    v1.0.7
+            //     1.0.7
+            //    ^
+            //    ignore the v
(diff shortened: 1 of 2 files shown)


src/install/install.zig
```diff
@@ -1062,9 +1062,28 @@ pub const PackageInstall = struct {
...
-        // Version is more likely to not match than name, so we check it first.
-        return strings.eql(package_json_checker.found_version, this.package_version) and
-            strings.eql(package_json_checker.found_name, this.package_name);
+        // Check if the version matches
+        if (!strings.eql(package_json_checker.found_version, this.package_version)) {
+            // If it doesn't match, check again but with a leading "v" removed
+            //    v1.0.7
+            //     1.0.7
+            //    ^
+            //    ignore the v
```
dylan-conway: do we also need to ignore `=`?
https://github.com/npm/node-semver/blob/816c7b2cbfcb1986958a290f941eddfd0441139e/internal/re.js#L124

dylan-conway (approved):

PR #8567, opened 2024-01-30
fix(windows): more reliable extracting

paperclover:
Windows cannot move a directory while we have a handle to it open

The zig standard library also aquires file locks.

We also close a file descriptor to only re-open it for renmaing (posix doesn't have this type of "rename open file handle", but windows does, so use it!)

Some other misc changes i did while messing with other parts of codebase.


Jarred-Sumner (changes requested):
does not compile on posix

src/install/extract_tarball.zig
```diff
@@ -432,3 +441,92 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
...
+    switch (rc) {
+        .SUCCESS => return result,
+        .OBJECT_NAME_INVALID => return error.BadPathName,
+        .OBJECT_NAME_NOT_FOUND => return error.FileNotFound,
+        .OBJECT_PATH_NOT_FOUND => return error.FileNotFound,
+        .NOT_A_DIRECTORY => return error.NotDir,
+        // This can happen if the directory has 'List folder contents' permission set to 'Deny'
+        // and the directory is trying to be opened for iteration.
+        .ACCESS_DENIED => return error.AccessDenied,
+        .INVALID_PARAMETER => unreachable,
```
Jarred-Sumner:
```suggestion
        .INVALID_PARAMETER => return error.EINVAL,
```

src/libarchive/libarchive.zig
```diff
@@ -564,7 +562,7 @@ pub const Archive = struct {
                         Kind.sym_link => {
                             const link_target = lib.archive_entry_symlink(entry).?;
                             if (comptime Environment.isWindows) {
-                                @panic("TODO on Windows");
+                                @panic("TODO on Windows: Extracting archives containing symbolic links.");
```
Jarred-Sumner, line 565: We should just support this. npm packages do it all the time.

paperclover: sure. Does NPM/Yarn/PNPM have a fallback for if symbolic links are not allowed?

if so, we need to implement that in addition to trying the symlink.

Jarred-Sumner: probably reparse points

src/cli/create_command.zig
```diff
@@ -2255,18 +2255,30 @@ const GitHandler = struct {
...
+        // Benchmark #1: bun create react3 app --force --no-install
+        //   Time (mean ± σ):     306.7 ms ±   6.1 ms    [User: 31.7 ms, System: 269.8 ms]
+        //   Range (min … max):   299.5 ms … 318.8 ms    10 runs
 
         if (which(&bun_path_buf, PATH, destination, "git")) |git| {
             const git_commands = .{
-                &[_]string{ bun.asByteSlice(git), "init", "--quiet" },
-                &[_]string{ bun.asByteSlice(git), "add", destination, "--ignore-errors" },
-                &[_]string{ bun.asByteSlice(git), "commit", "-am", "Initial commit (via bun create)", "--quiet" },
+                &[_]string{ git, "init", "--quiet" },
```
Jarred-Sumner, line 2279: this is a usecase for the shell!

src/install/extract_tarball.zig
```diff
@@ -195,7 +209,8 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
             return error.InstallFailed;
         };
 
-        defer extract_destination.close();
+        errdefer if (Environment.isWindows) extract_destination.close();
```
Jarred-Sumner, line 212: double close on `errdefer`.

Jarred-Sumner: nvm actually

src/install/install.zig
```diff
@@ -692,7 +692,9 @@ const Task = struct {
                 ) catch |err| {
                     if (comptime Environment.isDebug) {
                         if (@errorReturnTrace()) |trace| {
-                            std.debug.dumpStackTrace(trace.*);
+                            _ = trace; // autofix
```
Jarred-Sumner, line 695: ?

PR #8568, opened 2024-01-30
create zig.exe symlink on every download-zig.sh execution

vinnichase:
Symlink is not created on first setup.

see https://github.com/ziglang/vscode-zig/issues/164
and https://github.com/oven-sh/bun/pull/8147

### What does this PR do?

Moved the symlink creation out of the if statement to ensure it will be run on first clone and on subsequent updates of zig.

Ran the script with empty `.cache` folder and again with zig already installed. `zig.exe` is present.


paperclover:
thanks

PR #8576, opened 2024-01-30
fix(bundler)!: stop inlining process.env in the bundler

### LINKED ISSUE #3835: bun build injects env into bundle when --target=bun

0.7.1

Darwin 21.4.0 arm64 arm

- `bun init`

```dotenv
# .env
TEST_VALUE="hello, world!"
```
```typescript
// index.ts
console.log(process.env.TEST_VALUE);
```

- `bun build --target=bun --outdir=dist index.ts`

```javascript
// dist/index.js
console.log("hello, world!");
```

Unless this is intentional, the output of the bundle should still contain "process.env.TEST_VALUE" as opposed to the value present at bundle time.

### What do you see instead?

Bun injects the .env values present at bundle time into the bundle as hardcoded values.

### Additional information

I think having the option for this unintentional behavior, potentially via a flag, could actually be beneficial as a feature. I'm considering frontend applications where while the concept of environment variables doesn't exist, a level of bundle-time configuration could be useful. Similar behavior can be found in [create-react-app](https://create-react-app.dev/docs/adding-custom-environment-variables/) and [vite](https://vitejs.dev/guide/env-and-mode.html#production-replacement).

paperclover:
This should have never been implemented in the first place. This is a breaking change, but no one should rely on automatic exposing of secret values.

Fixes #3835


paperclover:
need to do tests

paperclover:
Will take an alternate approach.

rubenfiszel:
Any updates on this @paperdave

paperdave:
wait, why didnt this get the followup pr.

PR #8578, opened 2024-01-30
fix(windows): make `process.env` case-insensitive

paperclover:
### What does this PR do?

- `process.env` keys are now case insensitive
- process env values become strings
- bump the version 👁️


src/bun.js/bindings/JSEnvironmentVariableMap.cpp
```diff
@@ -125,19 +135,30 @@ JSValue createEnvironmentVariablesMap(Zig::GlobalObject* globalObject)
...
+#endif
         if (name == TZ) {
             hasTZ = true;
             continue;
         }
         ASSERT(len > 0);
-
-        Identifier identifier = Identifier::fromString(vm, name);
+#if OS(WINDOWS)
+        String idName = name.convertToASCIIUppercase();
```
dylan-conway, line 157: do we need to keep the original casing as well? `Object.keys` on `process.env` will report the original

paperclover: yes due to the proxy ownKeys

test/cli/run/env.test.ts
```diff
@@ -701,8 +701,27 @@ console.log(dynamic().NODE_ENV);
...
-    "package.json": '{"scripts":{"show-env":"env | grep NODE_ENV && exit 1 || true"}}',
+    "package.json": '{"scripts":{"show-env":' + JSON.stringify(getenv) + "}}",
   });
   expect(bunRunAsScript(tmp, "show-env", {}).stdout).toBe("");
 });
+
+const todoOnPosix = process.platform !== "win32" ? test.todo : test;
+todoOnPosix("setting process.env coerces the value to a string", () => {
+  // @ts-expect-error
+  process.env.SET_TO_TRUE = true;
```
dylan-conway, line 715: i wonder if `todoIf` would be useful

dylan-conway (approved):

src/js/builtins/ProcessObjectInternals.ts
```diff
@@ -344,3 +344,47 @@ export function setMainModule(value) {
...
+
+export function windowsEnv(internalEnv: InternalEnvMap, envMapList: Array<string>) {
+  // The use of String(key) here is intentional because Node.js as of v21.5.0 will throw
+  // on symbol keys as it seems they assume the user uses string keys:
+  //
+  // it throws "Cannot convert a Symbol value to a string"
+
+  return new Proxy(internalEnv, {
+    get(_, p) {
+      return typeof p === "string" ? Reflect.get(internalEnv, p.toUpperCase()) : undefined;
```
Jarred-Sumner, line 358: Can you add a `TODO` to use intrinsics instead of `Reflect`? We don't need to use Reflect here at all.

paperclover: no i will just fix it

PR #8579, opened 2024-01-30
windows(fix): `__dirname` is no longer empty.

paperclover:
`__dirname` was previously empty because the path used `\` instead of the expected `/`.

If I ever make a programming language, it will be a syntax error to use the character constant `'/'`

Fixes the `bcrypt` npm module, which seems to just work now.


dylan-conway (commented):
Main concern: will these strings always use the expected platform separator? If it's possible they can have both, we need to check for both

PR #8582, opened 2024-01-31
bun.path.relative: add assert that input is well-formed

nektro:
if asking for a specific semantics and the path does not have the separator as expected this method will return unexpected output. this adds a check for that in safe modes


paperclover (approved):
good idea

PR #8592, opened 2024-01-31
[bun:sqlite] Return changed row count for sqlite insert, update, delete on statement.run()

### LINKED ISSUE #3284: SQLite: return the number of rows changed

It can be useful to determine whether an `UPDATE` or `DELETE` query changed any data, and how much. I.e., did my `WHERE` actually match anything?

`better-sqlite3`'s `.run()` method returns on object that knows number of rows changed. 

`pg`'s `Result` objects behave similarly.

`pg-promise` appears to offer a dedicated `db.result()` method for running a query that discards results but retrieves the count of modified rows.

Adding `RETURNING *` to the relevant queries. Works for the scenarios I've encountered so far, but no telling if it'll always be appropriate, and it makes the DB and Bun do more work than necessary.

### LINKED ISSUE #8284: bun:sqlite: Statement.run returns void instead of object

1.0.21

Darwin 23.2.0 arm64 arm

When running an update statement, there is no way to inspect the number of rows affected by the update. 

```ts

db.run(`
CREATE TABLE bug_demo (
  id INTEGER,
  data TEXT
);
`)
db.run(`insert into bug_demo (id, data) values (1, 'foo')`)
console.log(db.query(`select * from bug_demo`).get()); // => { id: 1, data: "foo"}
const result = db.query(`update bug_demo set data='bar' where id = 1`).run();
console.log(`result: ${result}`); // => result: undefined
console.log(db.query(`select * from bug_demo`).get()); // => { id: 1, data: "bar"}
```

The return value of an update statement should include the number of rows affected.

### What do you see instead?

`.run` returns void. 
https://bun.sh/docs/api/sqlite#reference

### Additional information

better-sqlite3 returns an object from .run
https://github.com/WiseLibs/better-sqlite3/blob/HEAD/docs/api.md#runbindparameters---object

sidesteps:
### What does this PR do?

fixes #3284 fixes #8284. Returns changed row count for sqlite insert, update, delete statements with statement.run()

- [x] Code changes

- [x] I wrote automated tests

### DIFF
--- a/packages/bun-types/sqlite.d.ts
+++ b/packages/bun-types/sqlite.d.ts
@@ -520,21 +520,21 @@ declare module "bun:sqlite" {
     get(...params: ParamsType): ReturnType | null;
 
     /**
-     * Execute the prepared statement. This returns `undefined`.
+     * Execute the prepared statement. Returns the number of rows changed for INSERT, UPDATE or DELETE statements.
      *
      * @param params optional values to bind to the statement. If omitted, the statement is run with the last bound values or no parameters if there are none.
      *
      * @example
      * ```ts
-     * const stmt = db.prepare("UPDATE foo SET bar = ?");
+     * const stmt = db.prepare("INSERT INTO foo (col1) VALUES(?1)");
      * stmt.run("baz");
-     * // => undefined
+     * // => 1
      *
      * stmt.run();
-     * // => undefined
+     * // => 1
      *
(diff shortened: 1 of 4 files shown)


sidesteps:
If PR is accepted, I will update web documentation for statement.run()

Jarred-Sumner:
This was done in Bun v1.1.12 or so

Issue #6526, opened 2023-10-15
`test.only` doesn't stop skip tests that were defined before


wladpaiva:
### What version of Bun is running?

v1.0.6

### What platform is your computer?

Darwin 23.0.0 x86_64 i386

### What steps can reproduce the bug?

`.only` gets even more confused when running inside a `describe` but this is what is basically happening

```ts
import {test} from 'bun:test'

test('should not run', () => {})
test('should not run 2', () => {})
test.only('test.only', () => {}) // runs everything up to here
test('bla bla bla', () => {})
test('bla bla bla bla', () => {})
```

### What is the expected behavior?

Only run the specific test

### What do you see instead?

```sh
bun test v1.0.6 (969da088)

index.test.ts:
✓ should not run [0.01ms]
✓ should not run 2
✓ test.only

 3 pass
 0 fail
Ran 3 tests across 1 files. [9.00ms]
```

Hanaasagi:
Please try `bun test --only`

```
>>​ bun test --only test/6526.test.ts
bun test v1.0.6 (969da088)

test/6526.test.ts:
✓ test.only [0.02ms]

 1 pass
 0 fail
Ran 1 tests across 1 files. [86.00ms]

```

wladpaiva:
@Hanaasagi that kinda makes it work as it should.

But if the idea is to make the `only` work as a `onlyUpUntilHere` then anything that is inside a `describe` gets ignored.

```ts
import { test, describe } from "bun:test";

test("should not run", () => {});
test("should not run 2", () => {});

describe("xxx", () => {
  test("should run then?", () => {});
});

test.only("test.only", () => {}); // runs everything up to here
test("bla bla bla", () => {});
test("bla bla bla bla", () => {});
```

```sh
➜ bun test
bun test v1.0.6 (969da088)

6526.test.ts:
✓ should not run [0.01ms]
✓ should not run 2
✓ test.only

 3 pass
 0 fail
Ran 3 tests across 1 files. [9.00ms]
```

wladpaiva:
<img width="627" alt="Screenshot 2023-10-16 at 08 29 15" src="https://github.com/oven-sh/bun/assets/1843792/ff5aaa64-3a1e-475a-9547-7aedc444d8c7">

Docs also seems misleading

Electroid:
We should update the JSDocs to make this more clear. The reason we do not support previous `.only` tests is for performance reasons (to do so, we need to buffer all tests).

[...]

PR #8600, opened 2024-01-31
Issue-6526 - clarify web docs for the use of `.only()`

jcarpe:
### What does this PR do?

This updates the web docs to clarify the use of `.only()` in tests. Added similar clarification in JSDoc per the reported issue.

Fix: https://github.com/oven-sh/bun/issues/6526


Jarred-Sumner:
Thank you

PR #8604, opened 2024-01-31
add `isntall` alias for install command

sk337:
### What does this PR do?

This adds `isntall` alias for install command

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -1207,6 +1207,7 @@ pub const Command = struct {
 
             RootCommandMatcher.case("i"),
             RootCommandMatcher.case("install"),
+            RootCommandMatcher.case("isntall"),
             => brk: {
                 for (args_iter.buf) |arg| {
                     if (arg.len > 0 and (strings.eqlComptime(arg, "-g") or strings.eqlComptime(arg, "--global"))) {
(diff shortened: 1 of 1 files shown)


paperdave:
i guess this is because of?

image

it might be better to just push people towards shorter aliases like `bun i` instead of adding typo aliases

sk337:
yes and i am tired of rerunning the command whenever i make that typo and dont notice it.

Anush008:
There are also these possible values.

https://github.com/npm/cli/blob/latest/lib%2Futils%2Fcmd-list.js#L120-L128

jdalton:
> Referen

@sk337 Can you align with the list `npm` has ☝️

paperdave:
the more i think about this the more i dont want this. we should add to the script not found message a "note" message "did you mean 'bun install'"

this would make it so someone with a script named istall would run, instead of not running (maybe someome out there has some tallness detector script)

we could then open this up to ALOT of more typos for all commands.

sk337:
> > There are also these possible values.
> > https://github.com/npm/cli/blob/latest/lib%2Futils%2Fcmd-list.js#L120-L128
> 
> @sk337 Can you align with the list `npm` has ☝️

ok

paperdave:
this is far too many. bun does not need to match the npm cli interface. someone with a script named `in` will no longer work. see previous comment

Jarred-Sumner:
I don’t think we should do this. The terminal can provide this feature for you

PR #8608, opened 2024-02-01
fix(windows): fix installing non-ascii paths and make `normalizeBuf` generic

dylan-conway:
### What does this PR do?
Fixes installing packages with non-ascii paths.

Comment in `libarchive.zig`:
```
// TODO:
// Due to path separator replacement and other copies that happen internally, libarchive changes the
// storage type of paths on windows to wide character strings. Using `archive_entry_pathname` or `archive_entry_pathname_utf8`
// on an wide character string will return null if there are non-ascii characters.
// (this can be seen by installing @fastify/send, which has a path "@fastify\send\test\fixtures\snow ☃")
//
// Ideally, we find a way to tell libarchive to not convert the strings to wide characters and also to not
// replace path separators. We can do both of these with our own normalization and utf8/utf16 string conversion code.
```

Also makes `normalizeBuf` and other functions it uses generic, so they can take u8 or u16 slices. In the future we should update more of our string operations to be generic.

manually tested installing `@fastify/send`


src/libarchive/libarchive.zig
```diff
@@ -499,35 +499,56 @@ pub const Archive = struct {
...
+                    // Due to path separator replacement and other copies that happen internally, libarchive changes the
+                    // storage type of paths on windows to wide character strings. Using `archive_entry_pathname` or `archive_entry_pathname_utf8`
+                    // on an wide character string will return null if there are non-ascii characters.
+                    // (this can be seen by installing @fastify/send, which has a path "@fastify\send\test\fixtures\snow ☃")
+                    //
+                    // Ideally, we find a way to tell libarchive to not convert the strings to wide characters and also to not
+                    // replace path separators. We can do both of these with our own normalization and utf8/utf16 string conversion code.
+                    var pathname: bun.OSPathSliceZ = if (comptime Environment.isWindows) brk: {
+                        const normalized = bun.path.normalizeBuf(
+                            u16,
```
paperclover, line 512: yay
generic normalizeBuf

src/libarchive/libarchive.zig
```diff
@@ -678,13 +698,20 @@ pub const Archive = struct {
                                         lib.ARCHIVE_OK => break :possibly_retry,
                                         lib.ARCHIVE_RETRY => {
                                             if (comptime log) {
-                                                Output.prettyErrorln("[libarchive] Error extracting {s}, retry {d} / {d}", .{ pathname_, retries_remaining, 5 });
+                                                Output.prettyErrorln("[libarchive] Error extracting {}, retry {d} / {d}", .{
```
paperclover: consider Output.errGeneric

paperclover (approved):

src/libarchive/libarchive.zig
```diff
@@ -546,15 +567,15 @@ pub const Archive = struct {
...
+                                std.os.mkdiratW(dir_fd, pathname, @as(u32, @intCast(mode))) catch |err| {
                                     if (err == error.PathAlreadyExists or err == error.NotDir) break;
-                                    try bun.makePath(dir, std.fs.path.dirname(slice) orelse return err);
-                                    try std.os.mkdirat(dir_fd, pathname, 0o777);
+                                    try bun.MakePath.makePath(u16, dir, bun.Dirname.dirname(u16, path_slice) orelse return err);
+                                    try std.os.mkdiratW(dir_fd, pathname, 0o777);
                                 };
                             } else {
-                                std.os.mkdiratZ(dir_fd, pathname, @as(u32, @intCast(mode))) catch |err| {
+                                std.os.mkdirat(dir_fd, pathname, @as(u32, @intCast(mode))) catch |err| {
```
Jarred-Sumner: `std.os.mkdiratZ`

PR #8617, opened 2024-02-01
fix: id instead of path

dylan-conway:
### What does this PR do?
fixes some tests


SukkaW (commented):
This should fix the `error: Cannot find module "./xxxx" from "/path/to/node_modules/xxx"` that occurred since an hour ago, right?

dylan-conway:
> This should fix the `error: Cannot find module "./xxxx" from "/path/to/node_modules/xxx"` that occurred since an hour ago, right?

yes, since this commit 2af7cdedc24d3c29baba7e4d8c07e18648a85c95

PR #8626, opened 2024-02-01
fix(debug): get the result for fd debug message before we close it

paperclover:
### What does this PR do?

See https://github.com/ziglang/zig/issues/18756 for more context. This PR fixes the mistake on my side of the bug. We shouldn't run into the crash in the standard library anymore (unless you try printing the mount point manager's fd; something we don't do)

This also attempts to enable FD path logging on Linux and MacOS but it doesnt seem to work as we downcast fds to integers way more often, so the FDImpl formatter basically never hits. But it's possible:tm:


src/fd.zig
```diff
@@ -185,9 +185,9 @@ pub const FDImpl = packed struct {
         if (env.os != .windows or this.kind == .uv) {
             // This branch executes always on linux (uv() is no-op),
             // or on Windows when given a UV file descriptor.
-            const fd = bun.toFD(this.uv());
-            if (fd == bun.STDOUT_FD or fd == bun.STDERR_FD) {
-                log("close({}) SKIPPED", .{this});
+            const fd = this.uv();
+            if (fd == 1 or fd == 2) {
```
paperclover, line 189: `bun.STDOUT_FD` was removed here as it may point to the Windows handle. This is more accurate.

Jarred-Sumner:
Conflict

PR #8631, opened 2024-02-02
fix(windows): initialize uv allocators earlier

paperclover:
We can merge this now, or we can wait for me to finish the worker test fixes.


Jarred-Sumner (changes requested):
doesn't compile on posix

PR #8633, opened 2024-02-02
windows: fix module.paths getter causing a crash

nektro:
ideally I'd like to add a test for this but it returns absolute paths

### DIFF
--- a/src/resolver/resolver.zig
+++ b/src/resolver/resolver.zig
@@ -3278,8 +3278,7 @@ pub const Resolver = struct {
 
         const str = brk: {
             if (std.fs.path.isAbsolute(sliced.slice())) break :brk sliced.slice();
-            const dir_path_buf = bufs(.node_modules_paths_buf);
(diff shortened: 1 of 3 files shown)


src/resolver/resolver.zig
```diff
@@ -3278,8 +3278,7 @@ pub const Resolver = struct {
 
         const str = brk: {
             if (std.fs.path.isAbsolute(sliced.slice())) break :brk sliced.slice();
-            const dir_path_buf = bufs(.node_modules_paths_buf);
-            break :brk r.fs.joinBuf(&[_]string{ r.fs.top_level_dir, sliced.slice() }, dir_path_buf);
+            break :brk bun.path.join(&[_]string{ r.fs.top_level_dir, sliced.slice() }, .auto);
```
Jarred-Sumner: This will cause strange memory errors if the resolver is run recursively. That is why we were using `bufs` here instead of the implicit default buffer that bun.path.join uses.

PR #8646, opened 2024-02-02
Assert we drained the microtask queue since the last time we called into JavaScript

Jarred-Sumner:
### What does this PR do?

This adds an assertion in debug builds that checks we called `drainMicrotasks` after calling into a JavaScript function. 

This adds a helper method `EventLoop.runCallback` for calling into JavaScript functions from the top of the event loop. This will handle exceptions and release weakrefs and anything else we may need to do in the future.

This also fixes cases where we weren't draining microtasks in:
- IPC
- open() callback in Bun.listen/Bun.connect
- `open` and `close` callbacks in Bun.serve() WebSockets
- H2

The panic looks like this:
```zig
uh-oh: 1 JavaScript functions were called outside of the microtask queue without draining microtasks.

Last function name: open

Use EventLoop.runCallback() to run JavaScript functions outside of the microtask queue.

Failing to do this can lead to a large number of microtasks being queued and not being drained, which can lead to a large amount of memory being used and application slowdown.
bun will crash now 😭😭😭
```

It only tracks function names if `BUN_TRACK_LAST_FN_NAME=1`

Manually

### DIFF
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -100,9 +100,11 @@ pub fn WorkTask(comptime Context: type) type {
         // This is a poll because we want it to enter the uSockets loop
         ref: Async.KeepAlive = .{},
 
+        pub usingnamespace bun.New(@This());
+
         pub fn createOnJSThread(allocator: std.mem.Allocator, globalThis: *JSGlobalObject, value: *Context) !*This {
             var vm = globalThis.bunVM();
-            var this = bun.new(This, .{
+            var this = This.new(.{
                 .event_loop = vm.eventLoop(),
                 .ctx = value,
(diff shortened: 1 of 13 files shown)


src/bun.js/event_loop.zig
```diff
@@ -655,6 +662,57 @@ pub const EventLoop = struct {
...
+        log("exit() = {d}", .{count - 1});
+
+        defer this.debug.exit();
+
+        if (count == 1) {
+            this.global.vm().releaseWeakRefs();
+            this.drainMicrotasksWithGlobal(this.global);
+        }
+
+        this.entered_event_loop_count -= 1;
```
dylan-conway, line 713: should it be an assertion that it's never negative?

PR #8649, opened 2024-02-02
Async file copying on windows

gvilums:
### What does this PR do?
This PR implements `fs.promises.cp` on windows, and fixes some errors with symlinks with `fs.cpSync`.

Tests in `cp.test.ts`


src/bun.js/node/node_fs.zig
```diff
@@ -263,10 +263,10 @@ pub const AsyncCpTask = struct {
         vm: *JSC.VirtualMachine,
         arena: bun.ArenaAllocator,
     ) JSC.JSValue {
-        if (comptime Environment.isWindows) {
-            globalObject.throwTODO("fs.promises.cp is not implemented on Windows yet");
-            return .zero;
-        }
+        // if (comptime Environment.isWindows) {
```
paperclover: delete

src/bun.js/node/node_fs.zig
```diff
@@ -6374,14 +6370,55 @@ pub const NodeFS = struct {
...
+                    }
+                    // should be unreachable
+                    return Maybe(Return.CopyFile).todo();
+                }
+                const flags = if (stat_ & windows.FILE_ATTRIBUTE_DIRECTORY != 0)
+                    std.os.windows.SYMBOLIC_LINK_FLAG_DIRECTORY
+                else
+                    0;
+                if (windows.CreateSymbolicLinkW(dest, wbuf[0..len :0], flags) == 0) {
+                    if (Maybe(Return.CopyFile).errnoSysP(0, .copyfile, this.osPathIntoSyncErrorBuf(dest))) |e| {
```
paperclover: i think you can pass the rc of the api call into this function, instead of 0. it will return null on success. this will eliminate the "should be unreachable" thing.

src/bun.js/node/node_fs.zig
```diff
@@ -6390,48 +6427,72 @@ pub const NodeFS = struct {
...
-            task.finishConcurrently(Maybe(Return.Cp).todo());
-            return;
-        }
-
         const args = task.args;
-        var src_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
-        var dest_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
-        const src = args.src.sliceZ(&src_buf);
-        const dest = args.dest.sliceZ(&dest_buf);
+        // var src_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
```
paperclover: these comments can be removed

src/resolver/resolve_path.zig
```diff
@@ -623,6 +623,15 @@ pub fn normalizeStringGeneric(
 ) []u8 {
     return normalizeStringGenericT(u8, path_, buf, allow_above_root, separator, isSeparator, preserve_trailing_slash);
 }
+
+fn separatorAdapter(comptime T: type, func: anytype) fn (T) bool {
```
paperclover, line 639: this is fine but later this would be legendary to make into a generic wrapper for

"wrap any function with type"

src/resolver/resolve_path.zig
```diff
@@ -632,7 +641,34 @@ pub fn normalizeStringGenericT(
...
+        .zero_terminate = false,
+        .add_nt_prefix = false,
+    });
+}
+
+pub fn NormalizeOptions(comptime T: type) type {
+    return struct {
+        allow_above_root: bool = false,
+        separator: T = std.fs.path.sep,
+        isSeparator: fn (T) bool = makeIsSepAnyT(T),
```
paperclover: i wish this could be a comptime field but i suspect it isnt so shrimple to do so is it?

src/sys.zig
```diff
@@ -357,44 +357,22 @@ pub fn getErrno(rc: anytype) bun.C.E {
...
-        }
-        break :brk true;
-    };
-    if (ok) {
-        // no need to normalize, proceed normally
-        return .{
-            .result = bun.strings.toNTPath(buf, path),
-        };
+    if (comptime T != u8 and T != u16) {
+        @compileError("normalizePathWindows only supports u8 and u16 character types");
```
paperclover, line 367: good guard

src/sys.zig
```diff
@@ -562,9 +525,9 @@ pub fn openatWindows(dir: bun.FileDescriptor, path: []const u16, flags: bun.Mode
 ///
 /// In the zig standard library, messing up the input to their equivalent
 /// will trigger `unreachable`. Here there will be a debug log with the path.
-pub fn ntCreateFile(
+pub fn openFileAtWindowsNtPath(
```
paperclover, line 528: this is the best function rename of the month. thanks

src/sys.zig
```diff
@@ -562,9 +525,9 @@ pub fn openatWindows(dir: bun.FileDescriptor, path: []const u16, flags: bun.Mode
...
     dir: bun.FileDescriptor,
-    path_maybe_leading_dot: []const u16,
+    path: [:0]const u16,
```
paperclover: why is this required to be a sentinel? it is converted to `windows.UNICODE_STRING`.

`[]const u16` can be downcasted from `[:0]const u16` without any casts.

src/sys.zig
```diff
@@ -680,6 +643,103 @@ pub fn ntCreateFile(
     }
 }
 
+pub fn openFileAtWindowsT(
```
paperclover, line 646: this is so much nicer.
something i think about and it's fine to not do this now consideirng the size of the pr,

is maybe `access_mask`, `disposition`, etc can be shoved into an options struct similar to what std.os.windows.OpenFile does. there we can hvae the common defaults and also expose other flags (like the share args, but i dont think we'll actually need to customize that part)

src/windows.zig
```diff
@@ -69,6 +69,11 @@ pub const nt_maxpath_prefix = [4]u16{ '\\', '\\', '?', '\\' };
 const std = @import("std");
 pub const HANDLE = win32.HANDLE;
 
+pub extern "kernel32" fn GetFileInformationByHandle(
```
paperclover, line 77: this ... isnt defined anywhere??? damn

For ease of use do you want to add
```
/// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileinformationbyhandle
```

paperclover (changes requested):
legendary

i have some really minor change requests

src/sys.zig
gvilums: Good point, and this was already correct for the directory version of this function. updated.

src/resolver/resolve_path.zig
gvilums: which field do you think should be comptime?

src/bun.js/node/node_fs.zig
gvilums: passing `rc` or `0` doesn't make a difference because `rc` is known to be zero. The main issue is that constructing the error object may fail, for example if we detect that the status code is actually a success.

I've gone ahead and changed these to instead construct errors with a generic `ENOENT` code, so that if these do ever get hit the error is at least somewhat helpful.

src/resolver/resolve_path.zig
gvilums: yeah I agree, could probably live in the bun namespace. Only problem is that zig doesn't support variadics, so it can't be fully generalized 🥲

src/sys.zig
gvilums: yeah I think that would be a good idea. in general I think working with options structs is nice for these more complicated APIs

src/resolver/resolve_path.zig
paperclover: isSeparator. if i remember correctly it was previously a comptime function. i might be misremembering. what is significant is that i am less certain this function will be inlined. in the previous code (if im right and it used comptime function), then it is almost certain that LLVM inlined a function for this `==` thing

gvilums: because the entire struct is only ever used in a comptime context I think this should be basically equivalent to passing a comptime parameter (correct me if I'm wrong on that one)

src/resolver/resolve_path.zig
Jarred-Sumner: see std.meta.ArgsTuple

it’s like variadics but worse

Jarred-Sumner:
awesome

PR #8652, opened 2024-02-03
refactor out zig bug workarounds that are fixed

nektro:


src/logger.zig
```diff
@@ -562,8 +560,6 @@ pub const Msg = struct {
     }
 };
 
-// Do not mark these as packed
```
Jarred-Sumner: Please leave this comment in. This costed me a day.

src/js_ast.zig
```diff
@@ -7371,9 +7371,7 @@ pub const Macro = struct {
                 }
             };
 
-            // TODO: can change back to `return CallData.callWrapper(.{`
-            // when https://github.com/ziglang/zig/issues/16242 is fixed
-            return CallData.callWrapper(CallArgs{
+            return CallData.callWrapper(.{
```
Jarred-Sumner, line 7374: Please manually verify this works as expected. @dylan-conway probably remembers this one well

Jarred-Sumner (changes requested):
These changes look innocuous but I don’t trust our test coverage adequately checks this doesn’t cause regressions

If you want to merge this PR, please manually test each workaround has been fixed

src/js_ast.zig
dylan-conway: To verify this one, you will need to verify the args in `runAsync` are the same as the input passed through `callWrapper`. If I remember correctly, some of the fields in the `CallArgs` struct would become garbage. Probably best way to test is run macros with 0-n args

dylan-conway: I should've added a test with this commit

jdalton:
In my old job we'd call this "trust work". I'm very thankful @nektro for this!

PR #8654, opened 2024-02-03
fix: organize tsconfig

risu729:
### What does this PR do?

This organizes tsconfig for `bun init` and docs.

- Remove `forceConsistentCasingInFileNames` since it is enabled by default. ([docs](https://www.typescriptlang.org/tsconfig#forceConsistentCasingInFileNames))
- Remove `useUnknownInCatchVariables` since it is enabled by default when `strict` is `true`. ([docs](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables))
- Use consistent tsconfig in `bun init` and docs as follows the [document](https://bun.sh/docs/typescript#:~:text=If%20you%20run%20bun%20init%20in%20a%20new%20directory%2C%20this%20tsconfig.json%20will%20be%20generated%20for%20you.).
  > If you run bun init in a new directory, this tsconfig.json will be generated for you.
- Clarify that the stricter flags are not enabled in `tsconfig.json` generated by `bun init`.

- [x] Code changes


docs/guides/runtime/typescript.md
```diff
@@ -29,15 +29,14 @@ Below is the full set of recommended `compilerOptions` for a Bun project. With t
...
-    // Best practices
+    // Linting
     "strict": true,
     "skipLibCheck": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
 
     // Some stricter flags
-    "useUnknownInCatchVariables": true,
```
necmettindev, line 40: still needs `useUnknownInCatchVariables`

docs/typescript.md
```diff
@@ -31,21 +31,20 @@ Bun supports things like top-level await, JSX, and extensioned `.ts` imports, wh
...
-    // Best practices
+    // Linting
     "strict": true,
     "skipLibCheck": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
 
     // Some stricter flags
-    "useUnknownInCatchVariables": true,
```
necmettindev, line 42: and here

docs/guides/runtime/typescript.md
risu729: It is enabled by default with `strict`, so I think using it is meaningless.

src/cli/tsconfig-for-init.json
```diff
@@ -1,22 +1,22 @@
 {
   "compilerOptions": {
+    // Enable latest features
     "lib": ["ESNext"],
     "target": "ESNext",
     "module": "ESNext",
     "moduleDetection": "force",
-    "jsx": "react-jsx",
-    "allowJs": true,
+    "jsx": "react-jsx", // support JSX
```
paperclover: is it not obvious setting the tsconfig option `jsx` would enable jsx support.

i dont think this comment adds any value

src/cli/tsconfig-for-init.json
```diff
@@ -1,22 +1,22 @@
...
+    "allowJs": true, // allow importing `.js` from `.ts`
```
paperclover: i dont think this comment is fully accurate or helpful either.

risu729: I agree with it. I just remained them as is from the previous tsconfig.

src/cli/tsconfig-for-init.json
```diff
@@ -1,22 +1,22 @@
 {
   "compilerOptions": {
+    // Enable latest features
```
jcarpe: Should we remove comments in this context since it's within JSON?

jdalton (dismissed):
@risu729 Please revert any comment changes and md text changes, keeping the removal of unneeded config options.

risu729:
@jdalton Thank you for your review. I reverted the changes.

> If you run `bun init` in a new directory, this `tsconfig.json` will be generated for you.

My little concern is that `tsconfig.json` in docs and it generated by `bun init` are slightly different.

jdalton:
> My little concern is that `tsconfig.json` in docs and it generated by `bun init` are slightly different.

Ah. We should align them. We can bikeshed that in a separate PR.

jdalton:
Thank you @risu729! 
I'll merge this and if you're up for it a follow-up to align docs with our `bun init` reality would be appreciated.

risu729:
Thank you for merging! I'll open a new PR when this is merged.

PR #8657, opened 2024-02-03
Set a default value for `Error.prepareStackTrace` to align with Node

Jarred-Sumner:
### What does this PR do?

When you do `console.log(Error.prepareStackTrace)` Node returns a function which formats the stacktrace

Now, Bun also does that 

This also fixes a bug where you previously could not set `Error.prepareStackTrace` to `undefined`

There is a test

### DIFF
--- a/src/bun.js/bindings/ZigGlobalObject.cpp
+++ b/src/bun.js/bindings/ZigGlobalObject.cpp
@@ -328,6 +328,96 @@ extern "C" Zig::GlobalObject* Bun__getDefaultGlobal();
 // TODO: thread_local for workers
 static bool skipNextComputeErrorInfo = false;
 
+static JSValue formatStackTraceToJSValue(JSC::VM& vm, Zig::GlobalObject* globalObject, JSC::JSGlobalObject* lexicalGlobalObject, JSC::JSObject* errorObject, JSC::JSArray* callSites, JSValue prepareStackTrace)
+{
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    auto* errorConstructor = lexicalGlobalObject->m_errorStructure.constructor(globalObject);
+
+    if (!prepareStackTrace) {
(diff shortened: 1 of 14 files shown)


src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -546,17 +636,21 @@ static String computeErrorInfoWithPrepareStackTrace(JSC::VM& vm, Zig::GlobalObje
...
         }
     }
 
-    globalObject->formatStackTrace(vm, lexicalGlobalObject, errorObject, callSites, prepareStackTrace);
+    JSValue value = formatStackTraceToJSValue(vm, jsDynamicCast<Zig::GlobalObject*>(lexicalGlobalObject), lexicalGlobalObject, errorObject, callSites, prepareStackTrace);
+
+    RETURN_IF_EXCEPTION(scope, String());
+
+    if (value.isString()) {
+        return value.toWTFString(lexicalGlobalObject);
```
Jarred-Sumner, line 652: This is one part i'm not 100% sure about

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -328,6 +328,96 @@ extern "C" Zig::GlobalObject* Bun__getDefaultGlobal();
...
+        }
+    }
+
+    // default formatting
+    size_t framesCount = callSites->length();
+
+    WTF::StringBuilder sb;
+    if (JSC::JSValue errorMessage = errorObject->getIfPropertyExists(lexicalGlobalObject, vm.propertyNames->message)) {
+        auto* str = errorMessage.toString(lexicalGlobalObject);
+        if (str->length() > 0) {
```
Jarred-Sumner, line 353: This fixes a formatting issue

dylan-conway (approved):

Issue #4728, opened 2023-09-09
Extending Bun.serve’s Request object doesn’t work


mzdr:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

I am trying to extend Bun’s Request to have some custom abstraction on my request instance, unfortunately there seem to be an issue with the approach.

```ts
class RichRequest extends Request {}

Bun.serve({
  port: 4444,
  fetch: async (request: Request) => {
    // ❌ doesn’t work. prints empty string.
    return new Response(await new RichRequest(request).text());

    // ✅ works.
    return new Response(await request.text());
  },
});

void fetch("http://localhost:4444", {
  method: "post",
  body: "does not work with RichRequest :(",
})
  .then((response) => response.text())
  .then(console.log);
```

### What is the expected behavior?

It should work like in other runtimes. This code for example runs in the browser and in node (18).

```ts
class RichRequest extends Request {}

await (new RichRequest(new Request('http://localhost/', { method: "post", body: "works" }))).text()
```

### What do you see instead?

I always get an empty string from the extended class, the same for running `.json()` which always fails in `Unexpected end of JSON input`.

mzdr:
@Jarred-Sumner Not sure if this is related to https://github.com/oven-sh/bun/issues/3817.

Seblor:
It seems you are expecting a dependency injection to behave like inheritence.

If your implementation of RichRequest is limited to

    class RichRequest extends Request {}

Then instanciating it with

    new RichRequest(request)

will not copy the data.

However, going all the way with dependency injection, you can implement RichRequest like so:

    class RichRequest extends Request {
      request: Request;
      constructor(request: Request) {
        super(request);
        this.request = request;
      }
      async text() {
        return await this.request.text();
      }
    }

mzdr:
@Seblor Correct me if I’m wrong but according to the [Fetch Standard](https://fetch.spec.whatwg.org/#ref-for-dom-request%E2%91%A0) and the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) docs, passing a request object as constructor argument should:

`Set request to input’s request.` and `A Request object, effectively creating a copy.`.

Don’t get me wrong, your approach is perfectly fine and is actually the workaround I currently do.

Seblor:
[...]

Issue #4161, opened 2023-08-15
TypeError: FormData parse error missing final boundary


janwilmake:
### What version of Bun is running?

0.7.3

### What platform is your computer?

Linux 5.15.0-60-generic x86_64 x86_64

### What steps can reproduce the bug?

I'm using Bun.serve` with the `request.formData` 

I'm using sendgrid to receive emails in bun but Sendgrid is apparently not giving FormData dat you can parse.
Somehow, sometimes it works, sometimes it doesn't.

Before I go more in-depth... maybe you can look at your `request.formData()` or let me know how to use it?

Makes it very difficult to get the files of attachments of emails in sendgrid.

### What is the expected behavior?

It should parse it even without boundary

### What do you see instead?

It throws `TypeError: FormData parse error missing final boundary`

### Additional information

My code:

```
export const tryGetFormData = async (request: Request | undefined) => {
  try {
    return request?.clone().formData();
  } catch (e) {
    return undefined;
  }
};
```

It Throws with Sendgrid incoming mail

janwilmake:
How can I help debug this? Maybe do you need the raw body and headers?

paperclover:
this is just a guess but maybe it is be that `.clone()` and `.formData()` don't work together. can you try this like `request?.formData()`?

janwilmake:
> this is just a guess but maybe it is be that `.clone()` and `.formData()` don't work together. can you try this like `request?.formData()`?

Nah, it seems to be the same. Doesn't clone just create a copy of request so you can do things twice?

EDIT: Weird, you seem to be right. The problem has gone away.

ZiNai:
I use **Bun(1.0.25)** run qwik.js [start docs](https://qwik.builder.io/docs/getting-started/)  .
I click the vote button. expect console log the form inputs, but  triggered this bug.
``` typescript JSX
import { component$ } from '@builder.io/qwik';
import { routeLoader$, Form, routeAction$ } from '@builder.io/qwik-city';
 
export const useJokeVoteAction = routeAction$((props) => {
  // Leave it as an exercise for the reader to implement this.
  console.log('VOTE', props);
});
export const useDadJoke = routeLoader$(async () => {
  const response = await fetch('https://icanhazdadjoke.com/', {
    headers: { Accept: 'application/json' },
  });
  return (await response.json()) as {
    id: string;
    status: number;
    joke: string;
  };
});
 
export default component$(() => {
    const dadJokeSignal = useDadJoke();
[...]

Issue #2644, opened 2023-04-13
FormData missing final boundary


yousefelgoharyx:
### What version of Bun is running?

0.5.9

### What platform is your computer?

WSL - x64

### What steps can reproduce the bug?

```js
async function handler(req: Request) {
	const contentType = req.headers.get("Content-Type")
	if (!contentType) throw new Error("Missing Content-Type");

	if (contentType.startsWith("multipart/form-data")) {
		const fd = await req.formData()
		const body = Object.fromEntries(fd.entries());
		return new Response(JSON.stringify(body))
	}

	throw new Error("Invalid Content-Type");
}
Bun.serve({
	port: 3000,
	async fetch(request) {
		await new Promise((resolve) => setTimeout(resolve, 0));
		const response = await handler(request)
		await new Promise((resolve) => setTimeout(resolve, 0));
		return response;
	}
})
```
Here is the code causing the bug. Try to hit `localhost:3000` many times quickly and the server will throw the following error message: 
`error: FormData missing final boundary`

When removing the setTimeouts everything works correctly. That is what causing the problem.

### What is the expected behavior?

Not to throw an error

### What do you see instead?

The error described above.

Liliandev:
With Astrobuild, i have this error : 

TypeError: FormData parse error missing final boundary

Same thing is i submit with Json

lafkpages:
Same error with SvelteKit and Superform

nanolikestea:
Can confirm the error on Sveltekit

maxa-ondrej:
Getting the same error on SvelteKit with `svelte-adapter-bun` adapter.

Dan1ve:
Same error here

ArthurSegato:
Can confirm the error on Qwik with bun adapter.
```
Ubuntu - 23.10
Bun - 1.0.23
Qwik - 1.4.0
Qwik City - 1.4.0
```

yaikohi:
I am not sure if it is bun or H3, but I am receiving this error as well when using solidstart, which proxies h3s functions.
```error
[h3] [unhandled] 1 | export default "native";
    ^
error: FormData parse error missing final boundary
      at formData (native:1:1)

1 | export default "native";
    ^
TypeError: FormData parse error missing final boundary
      at formData (native:1:1)

```

znycheporuk:
Here is the repo you can clone to reproduce the error (SolidStart)
https://github.com/znycheporuk/bun-formdata

[...]

PR #8668, opened 2024-02-03
fix: request.clone() actually clones the request body.

### LINKED ISSUE #1381: `Request.clone()` does not clone or support streaming body

### LINKED ISSUE #8515: `Request.clone()` does not clone body

1.0.25+a8ff7be64

macOS

clone an incoming request

the body is cloned

### What do you see instead?

blank body

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

This makes request.clone() call .body.tee(), and then re-assigns request.body to the teed ReadableStream. 

Fixes #8515
Fixes #1381

This needs tests.

### DIFF
--- a/src/bun.js/bindings/ZigGlobalObject.cpp
+++ b/src/bun.js/bindings/ZigGlobalObject.cpp
@@ -2284,6 +2284,18 @@ extern "C" void ReadableStream__cancel(JSC__JSValue possibleReadableStream, Zig:
     ReadableStream::cancel(*globalObject, readableStream, exception);
 }
 
+extern "C" void ReadableStream__tee(JSC__JSValue possibleReadableStream, Zig::GlobalObject* globalObject, JSC__JSValue ptr[2])
+{
+    auto* readableStream = jsDynamicCast<JSReadableStream*>(JSC::JSValue::decode(possibleReadableStream));
+    if (UNLIKELY(!readableStream))
+        return;
+
+    if (auto pair = ReadableStream::tee(*globalObject, readableStream)) {
+        ptr[0] = JSC::JSValue::encode(pair->first);
+        ptr[1] = JSC::JSValue::encode(pair->second);
+    }
+}
+
 extern "C" void ReadableStream__detach(JSC__JSValue possibleReadableStream, Zig::GlobalObject* globalObject);
 extern "C" void ReadableStream__detach(JSC__JSValue possibleReadableStream, Zig::GlobalObject* globalObject)
 {
--- a/src/bun.js/bindings/webcore/ReadableStream.cpp
+++ b/src/bun.js/bindings/webcore/ReadableStream.cpp
@@ -24,6 +24,7 @@
  */
 
 #include "config.h"
+#include "JavaScriptCore/JSGlobalObject.h"
 #include "ReadableStream.h"
 
 #include "Exception.h"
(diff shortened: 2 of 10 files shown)


src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -2284,6 +2284,18 @@ extern "C" void ReadableStream__cancel(JSC__JSValue possibleReadableStream, Zig:
     ReadableStream::cancel(*globalObject, readableStream, exception);
 }
 
+extern "C" void ReadableStream__tee(JSC__JSValue possibleReadableStream, Zig::GlobalObject* globalObject, JSC__JSValue ptr[2])
+{
+    auto* readableStream = jsDynamicCast<JSReadableStream*>(JSC::JSValue::decode(possibleReadableStream));
+    if (UNLIKELY(!readableStream))
```
necmettindev, line 2290: just wondering, is there a specific reason why we don't use `LIKELY(readableStream)?`

src/bun.js/bindings/webcore/ReadableStream.cpp
```diff
@@ -125,24 +126,25 @@ void ReadableStream::pipeTo(ReadableStreamSink& sink)
...
     if (!returnedValue)
         return {};
 
-    auto results = Detail::SequenceConverter<IDLInterface<ReadableStream>>::convert(lexicalGlobalObject, *returnedValue);
+    JSValue array = returnedValue.value();
+    auto* arrayObject = jsCast<JSArray*>(array.getObject());
+    auto first = arrayObject->getDirectIndex(&lexicalGlobalObject, 0);
+    auto second = arrayObject->getDirectIndex(&lexicalGlobalObject, 1);
 
-    ASSERT(results.size() == 2);
```
paperclover, line 144: why was this removed

src/bun.js/bindings/ZigGlobalObject.cpp
paperclover: this condition is likely to be true: `!readableStream`

This path only happens if you call this with a |this| value that is not JSReadableStream.

paperclover:
cannot merge because of failing test case. feature works for simpler cases but this may not ship with this memory corruption

image

KilianB:
Would it be worth to also support the copy constructor of `Request` in the same breath? https://developer.mozilla.org/en-US/docs/Web/API/Request/Request Open Bun issue: https://github.com/oven-sh/bun/issues/4728. This would allow to clone the request while adjusting individual fields like the url parameter, fix a bug in elysia (e.g. https://github.com/elysiajs/elysia/issues/332) where mounted routes are not able to access the body.

beorn:
Until this is implemented, perhaps just have `req.clone()` core dump / panic instead of leading to lots of hard-to-debug downstream errors?

billywhizz:
we investigated [an issue](https://github.com/oven-sh/bun/issues/2644) last week with formdata() not working as expected and it seems to be related to the issue addressed by this PR. when instantiating a new Request object and passing an existing request as the first argument to the constructor, we do not clone the body as required by the spec.

This is a list of known issues which are caused by this missing request body clone functionality:

## related issues
- https://github.com/oven-sh/bun/issues/2644
- https://github.com/oven-sh/bun/issues/10789
- https://github.com/oven-sh/bun/issues/4161
- https://github.com/oven-sh/bun/issues/4728

Jarred-Sumner:
Superseded by https://github.com/oven-sh/bun/pull/12473

Issue #4333, opened 2023-08-25
unref() missing on HttpServer


weyert:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Create a file with the following content:

```
require('http').createServer().unref()
```

Gives an error:
```
bun run ./poef.js
1 |
2 |   require("node:http").createServer().unref();
     ^
TypeError: require("node:http").createServer().unref is not a function. (In 'require("node:http").createServer().unref()', 'require("node:http").createServer().unref' is undefined)
      at /Users/weyert/poef.js:1:0
      at globalThis (/Users/weyert/poef.js:1:37)
```

### What is the expected behavior?

I would expect the error not to be thrown

### What do you see instead?

TypeError: require("node:http").createServer().unref is not a function. (In 'require("node:http").createServer().unref()', 'require("node:http").createServer().unref' is undefined)

### Additional information

I was trying out whether OpenTelemetry is working through Bun but I am getting the above error when trying to create an instance of the `PrometheusError` which uses the above `unref()` call. 

Original error I received was:
```
88 |         this._appendTimestamp =
89 |             typeof config.appendTimestamp === 'boolean'
90 |                 ? config.appendTimestamp
91 |                 : PrometheusExporter.DEFAULT_OPTIONS.appendTimestamp;
92 |         // unref to prevent prometheus exporter from holding the process open on exit
93 |         this._server = (0, http_1.createServer)(this._requestHandler).unref();
                ^
TypeError: (0, http_1.createServer)(this._requestHandler).unref is not a function. (In '(0, http_1.createServer)(this._requestHandler).unref()', '(0, http_1.createServer)(this._requestHandler).unref' is undefined)
      at new PrometheusExporter (/Users/weyert/Development/Project/node_modules/.pnpm/@opentelemetry+exporter-prometheus@0.41.2_@opentelemetry+api@1.4.1/node_modules/@opentelemetry/exporter-prometheus/build/src/PrometheusExporter.js:93:13)
      at configureMetricExporter (/Users/weyert/Development/Project/packages/opentelemetry-launcher/dist/launcher.js:356:30)
      at configureOpenTelemetry (/Users/weyert/Development/Project/packages/opentelemetry-launcher/dist/launcher.js:111:4)
      at /Users/weyert/Development/Project/services/service/dist/bootstrap.js:43:12
[...]

Issue #2524, opened 2023-03-31
node:http has missing ref/unref functions


krutoo:
### What version of Bun is running?

0.5.8

### What platform is your computer?

Darwin 22.1.0 arm64 arm

### What steps can reproduce the bug?

Here's an example that shows Bun does not implement the ref and unref functions:
https://github.com/krutoo/bun-opentelemetry

According to @Jarred-Sumner on discord we ned to implement this functions:
https://discord.com/channels/876711213126520882/1091280232025182239/1091376674425274499

### What is the expected behavior?

Bun should implement ref/unref functions of http module

### What do you see instead?

Bun now not implemented ref/unref functions of http module

krutoo:
looks like docs contain incorrect information about node:http module according this issue 
https://bun.sh/docs/runtime/nodejs-apis

JuroOravec:
Came across this while trying to run [Crawlee/Apify](https://crawlee.dev/) (web scraping framework) via Bun (See discussion https://github.com/apify/proxy-chain/issues/521). This and the missing `fs.opendir` are blocking the adoption.

PR #8675, opened 2024-02-04
Implement server.unref() and server.ref()

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/2524
Fixes https://github.com/oven-sh/bun/issues/4333

The following keeps Bun's process alive forever, until `server.stop()` is called:
```js
const server = Bun.serve({
  fetch(req) { return new Response(); }
});
```

What if you want to only keep the process alive while there are active connections, but don't want to stop new connections from being accepted?

Now you can with `server.unref`:

```js
const server = Bun.serve({
  fetch(req) { return new Response(); }
});

await Bun.sleep(100);
server.unref();
```

A few tests

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -351,9 +351,12 @@ var optionsSymbol = Symbol("options");
 var serverSymbol = Symbol("server");
 function Server(options, callback) {
   if (!(this instanceof Server)) return new Server(options, callback);
+  EventEmitter.$call(this);
 
   this.listening = false;
-  EventEmitter.$call(this);
+  this._unref = false;
+  this[serverSymbol] = undefined;
+
   if (typeof options === "function") {
     callback = options;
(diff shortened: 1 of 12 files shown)


src/js/node/http.ts
```diff
@@ -351,9 +351,12 @@ var optionsSymbol = Symbol("options");
 var serverSymbol = Symbol("server");
 function Server(options, callback) {
   if (!(this instanceof Server)) return new Server(options, callback);
+  EventEmitter.$call(this);
 
   this.listening = false;
-  EventEmitter.$call(this);
+  this._unref = false;
```
Jarred-Sumner, line 357: This `_unref` boolean is what Node does

cs01:
Thank you! Looking forward to giving bun another go at work with this fix

Issue #4689, opened 2023-09-09
Implement `watchFiles` and `watchDirs` in plugin API


lucas-jones:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 5.15.0-67-generic x86_64 x86_64

### What steps can reproduce the bug?

When using a plugin `--watch` & `--hot` no longer work

- Calling the function `build.onLoad` will prevent watching & reloading to work on filtered files

## Basic Example
Just transpiles `.ts` into javascript using Bun's transpiler

### Command
`bun --watch ./src/index.ts`

### bunfig.toml
```toml
preload = ["./plugin.ts"]
```

### plugin.ts
```ts
import { BunPlugin, plugin, Transpiler } from "bun";

export default BunPlugin;

plugin(BunPlugin());

export function BunPlugin(): BunPlugin {
	return {
		name: "Test",
		async setup(build) {

			const transpiler = new Transpiler({ loader: "ts" });

			build.onLoad({ filter: /\.(ts)$/ }, async (args) => {
				const typescript = await Bun.file(args.path).text();
				const javascript = transpiler.transformSync(typescript);
				return {
					contents: javascript,
					loader: "js",
				};
			});
		},
	}
};
```

### What is the expected behavior?

Reloading of the application on file changes 

### What do you see instead?

No reloading

alpharder:
This is really frustrating. 
Used `nodemon` as a workaround for now

kirankunigiri:
I saw the latest comment on the PR https://github.com/oven-sh/bun/pull/8679 that adding the `watchedFilePaths` parameter will be added in a separate new PR. I wanted to ask if this issue could also be fixed in that PR because they are so similar. Specifically, allowing any plugin that uses `build.onLoad` to automatically keep the built-in `--watch` behavior (or a general plugin boolean flag to enable it), so all existing bun plugins can also work with `--watch` without needing to update their code.

riepspellz:
Hello, any updates about this ?

mattfysh:
the workaround:

```ts
plugin({
  name: 'graphql',
  setup(build) {
    build.onLoad({ filter: /\.graphql$/ }, async args => {
      const contents = await import(`${args.path}?`, { with: { type: 'text' } })
      return { loader: 'object', exports: contents }
    })
  },
})
```

PR #8679, opened 2024-02-04
Support `watchedFiles` in `Bun.plugin`

Jarred-Sumner:
### What does this PR do?

WIP. Lets onLoad plugins return a file path to add to watch. This makes it easier to do hot reloading with custom loaders & files.

Example code:

```js
import { plugin } from "bun";
import { compile } from "svelte/compiler";

await plugin({
  name: "svelte loader",
  async setup(builder) {
    builder.onLoad({ filter: /\.svelte$/ }, ({ path }) => ({
      contents: compile(readFileSync(path, "utf8"), {
        filename: path,
        generate: "ssr",
      }).js.code,
      loader: "js",
      watchedFilePaths: [path],
      // ^ watch this path
    }));
  },
});
```

If someone else wants to finish this PR, the next steps are roughly:
- Open the file path provided so that we get a file descriptor (for watching on macOS, mostly)
- Call `vm.bun_watcher.addFile`. Set `clone_path` to `true`
- A test which checks that hot-reloading works with e.g. `.svelte` files

I think it'd be fine to only support a single path to ship the initial version. 

It doesn't work yet

### DIFF
--- a/src/bun.js/bindings/ModuleLoader.cpp
+++ b/src/bun.js/bindings/ModuleLoader.cpp
@@ -1,3 +1,5 @@
+#include "JavaScriptCore/JSArray.h"
+#include "JavaScriptCore/JSCast.h"
 #include "root.h"
 #include "headers-handwritten.h"
 
@@ -40,6 +42,15 @@ namespace Bun {
 using namespace JSC;
 using namespace Zig;
 using namespace WebCore;
+extern "C" bool Bun__isAbsolutePath(BunString* path);
+
+enum class HotReload : uint8_t {
+    none = 0,
+    hot = 1,
+    watch = 2
+};
+
+extern "C" HotReload Bun__watchMode;
 
 static OnLoadResult handleOnLoadResultNotPromise(Zig::GlobalObject* globalObject, JSC::JSValue objectValue, bool wasModuleMock = false);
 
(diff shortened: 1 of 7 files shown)


src/bun.js/bindings/ModuleLoader.cpp
```diff
@@ -242,6 +254,46 @@ OnLoadResult handleOnLoadResultNotPromise(Zig::GlobalObject* globalObject, JSC::
     result.value.sourceText.value = JSValue {};
     result.value.sourceText.string = {};
 
+    if (Bun__watchMode != HotReload::none) {
+        if (JSC::JSValue watchedFilePathsValue = object->getIfPropertyExists(globalObject, JSC::Identifier::fromString(vm, "watchedFilePaths"_s))) {
```
Electroid, line 258: Just `watchedPaths`?

paperclover (changes requested):
why is it called `watchedFilePaths`? this should match esbuild and be called `watchFiles`, and if it is easy then `watchDirs` should also be adopted.

see https://esbuild.github.io/plugins/#on-resolve-arguments

paperclover:
This partially implements #4689 which had two duplicates I closed now.

Jarred-Sumner:
This still needs to be addressed, but it will happen in a different PR

PR #8682, opened 2024-02-04
Use bun's bundler to bundle node-fallbacks

Jarred-Sumner:
### What does this PR do?

Currently, the following doesn't work in `bun build`:

```js
import {Buffer} from 'node:buffer';
```

This fixes that

CI


paperclover (changes requested):
Please update `make-old-js.ps1` or else this might stop working on local builds of Windows. ~~Otherwise looks good. CI should just work.~~

There are failing tests in bundler_browser.

paperclover:
I mark this as draft because it regresses bundler_browser tests

jdalton:
Using the fallbacks _(maybe future name bikeshed to "browserify"'ed forms esp if that's the limitation)_ is a good thing _(essentially why they exist)_. ❤️  seeing this come _(when it can)_.

PR #8710, opened 2024-02-05
fix(windows): make fs.symlink normalize forward slashes to backslashes

### LINKED ISSUE #8273: [win][fs] Symlink should use `\` as a path separator

1.0.23

Window 10

```js
import fs from "node:fs/promises";

await fs.mkdir("files/2024", {recursive: true});
await fs.writeFile("files/2024/123.txt", "text");
await fs.symlink("files/2024/123.txt", "file-sym.txt");
console.log(await fs.readlink("file-sym.txt"));
```

```
files\2024\123.txt
```

The symlink works in Explorer.

### What do you see instead?

```
files/2024/123.txt
```
The symlink does not work in Explorer and other programs.

### Additional information

Rel: https://github.com/oven-sh/bun/issues/8272

paperclover:
### What does this PR do?

Fixes #8273

`sliceZWithForcedCopy` now returns a (mutable) []u8 slice instead of a constant one, but only if you enable force copy.

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -5717,8 +5717,16 @@ pub const NodeFS = struct {
         var to_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
 
         if (Environment.isWindows) {
+            const target: [:0]u8 = args.old_path.sliceZWithForceCopy(&this.sync_error_buf, true);
+            for (target) |*c| {
+                if (c.* == '/') {
+                    c.* = '\\';
+                }
+            }
+            // Symlinks targets on windows must contain backslashes. UV will not normalize this for us.
+            // https://github.com/oven-sh/bun/issues/8273
             return Syscall.symlinkUV(
-                args.old_path.sliceZ(&this.sync_error_buf),
+                target,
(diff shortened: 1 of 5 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -5717,8 +5717,16 @@ pub const NodeFS = struct {
         var to_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
 
         if (Environment.isWindows) {
+            const target: [:0]u8 = args.old_path.sliceZWithForceCopy(&this.sync_error_buf, true);
+            // UV does not normalize slashes in symlink targets, but Node does
+            // See https://github.com/oven-sh/bun/issues/8273
+            for (target) |*c| {
+                if (c.* == '/') {
```
Jarred-Sumner, line 5726: i feel like there's a SIMD @Vector thing we can do here

but we can leave that as a TODO

paperclover: i'll add a todo

Jarred-Sumner: sliceWIthZForceCopyWithNormalizedSlashesOnWindows

PR #8717, opened 2024-02-06
tidy: various code cleanups and enhancements

nektro:
fixed some TODOs and cleaned up more along the way


src/string_immutable.zig
```diff
@@ -39,38 +39,19 @@ pub inline fn w(comptime str: []const u8) [:0]const u16 {
     return Static.literal;
 }
 
-pub fn toUTF16Literal(comptime str: []const u8) []const u16 {
-    return comptime literal(u16, str);
-}
-
 pub inline fn literal(comptime T: type, comptime str: string) []const T {
```
paperclover, line 42: i was going to comment about the inComptime thing but it seems that is not needed anymore

i wonder if we should make this function return a sentinel slice since that is what string literals are, or more specifically, pointers to an array terminated by 0

Jarred-Sumner (changes requested):
test failures

src/string_immutable.zig
jdalton: I have found the error helpful.

nektro: the error will still throw but the check was doing redundant work the compiler already does now

src/string_immutable.zig
```diff
@@ -39,38 +39,19 @@ pub inline fn w(comptime str: []const u8) [:0]const u16 {
...
-    for (str, 0..) |c, i| {
-        // TODO(dylan-conway): should we check for non-ascii characters like JSC does with operator""_s
-        output[i] = c;
-    }
-
-    const Static = struct {
-        pub const literal: []const T = output[0..];
+    // literals are already utf-8 by default
+    // make sure we don't get any garbage passed by accident
+    comptime std.debug.assert(std.unicode.utf8ValidateSlice(str));
```
Jarred-Sumner, line 45: this will slow compile-times a little

we could just not do this

src/resolver/resolve_path.zig
```diff
@@ -2011,7 +2009,7 @@ pub fn lastIndexOfSep(path: []const u8) ?usize {
 
 pub fn lastIndexOfSepT(comptime T: type, path: []const T) ?usize {
     if (comptime !bun.Environment.isWindows) {
-        return strings.lastIndexOfCharT(T, path, '/');
+        return std.mem.lastIndexOfScalar(T, path, '/');
```
Jarred-Sumner, line 2012: why change this? they do the same thing right now but we will probably change out this implementation to a libc function later + it insulates us from having to make more code changes when zig std lib randomly updates the signature of this function in 3 months

PR #8732, opened 2024-02-06
Common HTTP server usage guide

jakeg:
### What does this PR do?

- [ ] Code changes

I've found it hard to find documentation for this sort of thing in the past. Thought it might be useful.


gvilums (approved):

gvilums:
Thank you!

docs/guides/http/server.md
```diff
@@ -0,0 +1,46 @@
...
+    const path = new URL(req.url).pathname;
+
+    // respond with text/html
+    if (path === "/") return new Response("Welcome to Bun!");
+
+    // redirect
+    if (path === "/abc") return Response.redirect("/source", 301);
+
+    // send back a file (in this case, *this* file)
+    if (path === "/source") return new Response(Bun.file(import.meta.file));
```
vitch, line 21: Is there a reason this uses import.meta.file rather than import.meta.path?

I ran into an error when I tried using this example code in a subfolder and I'm trying to figure if I need to configure my project differently or if this should be using import.meta.path?

(I know this is just example code and not meant to actually be used as-is but I was as a first step when experimenting with bun and I got tripped up by it so wondering if other people will be too)

((I left this comment on the commit too but then I realised it might be easier to see and discuss on the PR))

gvilums: You're right, using `import.meta.path` would be better, there is no reason to use `import.meta.file` in this case.

vitch: 👍  - submitted #11180 in case it helps...

PR #8741, opened 2024-02-07
Fix ref-counting of Bun.file(0) etc. (#8687)

### LINKED ISSUE #8687: Bun.file(1) FD corruption; GC bug?

1.0.26+c75e768a6

I'm running this script:

```
async function f()
{
    let f;

    for (let i = 0; i < 1000000; i++) {
	f = Bun.file(1);
	console.log(f);
	await Bun.write(f, "hello\n");
	console.log(f);
    }
}

await f();
```

A million repetitions of
```
FileRef (fd: 1) {
  type: "application/octet-stream"
}
hello
FileRef (fd: 1) {
  type: "application/octet-stream"
}
```

which would result in an output file of 7000000 lines.

### What do you see instead?

It's strange enough I have to paste it here:

```
strace -fff bun run ./bunbug2.js  2> strace | ` cat >out
wc -l out
666634 out
strace -fff bun run ./bunbug2.js  2> strace | cat >out
wc -l out
666606 out
strace -fff bun run ./bunbug2.js  2> strace | cat >out
wc -l out
666606 out
```

It's usually `666606` lines, corresponding to 95229 loop iterations running successfully plus one last `FileRef` output like this:

```
FileRef (fd: XXX) {
  type: "application/octet-stream"
}
```

where `XXX` is an unpredictable number corresponding to a 32-bit integer, sometimes negative.

The `cat` part is necessary for reproducibility.

The error message is:
```
3 |     let f;
4 | 
5 |     for (let i = 0; i < 1000000; i++) {
6 | 	f = Bun.file(1);
7 | 	console.log(f);
8 | 	await Bun.write(f, "hello\n");
           ^
EBADF: Bad file descriptor
   errno: -9
 syscall: "write"
      fd: XXX
```
where `XXX` is as above. The `strace` indicates bun is indeed attempting to write to the corrupt file descriptor.

### Additional information

I don't know about `bun` internals, but have some wild speculation:

The strace output makes it seem likely to me that there is a GC running which corrupts (or frees prematurely) the memory area holding the file descriptor. This is somewhat reproducible because the program is deterministic and GC timing should be almost deterministic. My conclusion is this is most likely a GC-related bug.

Experimenting with a FinalizationRegistry confirms that the first GC is happening around the time the bug strikes. With this program:

```
const registry = new FinalizationRegistry((i) => {
    console.log("free", i);
});

async function f()
{
    let f;

    for (let i = 0; i < 1000000; i++) {
	f = Bun.file(1);
	registry.register(f, i);
	console.log(f);
	await Bun.write(f, "hello\n");
	await new Promise(r => r());
	console.log(f);
	await new Promise(r => r());
    }
}

await f()
```

run as `bun run ./bunbug3.js | cat >out`, I see this output:

```
[361456 lines omitted]
FileRef (fd: 1) {
  type: "application/octet-stream"
}
FileRef (fd: 1) {
  type: "application/octet-stream"
}
hello
FileRef (fd: XXX) {
  type: "application/octet-stream"
}
FileRef (fd: XXX) {
  type: "application/octet-stream"
}
free 51636
free 51635
free 51634
free 51633
free 51632
free 51631
free 51630
free 51629
free 51628
free 51627
free 51626
free 51625
free 51624
free 51623
free 51622
free 51621
free 51620
free 51619
free 51618
free 51617
free 51616
free 51615
free 51614
free 51613
free 51612
[51607 lines omitted]
free 4
free 3
free 2
free 1
free 0
```
[edited to fix formatting]

argosphil:
### What does this PR do?

This fixes ref-counting of the Blob.Store objects used for Bun.file(0)/Bun.file(1)/Bun.file(2), which are reused for multiple Blob objects.

The previous code shared one Blob.Store object between several Blobs, which led to those Blob.Store objects being collected and freed prematurely.

See #8687 for further details.

Fixes #8687 

- [x] Code changes

I wrote an automated test, which is included.

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -1553,20 +1553,19 @@ pub const Blob = struct {
                     break :brk copy;
                 },
                 .fd => {
-                    switch (bun.FDTag.get(path_or_fd.fd)) {
-                        .stdin => return Blob.initWithStore(
-                            vm.rareData().stdin(),
+                    const optional_store: ?*Store = switch (bun.FDTag.get(path_or_fd.fd)) {
+                        .stdin => vm.rareData().stdin(),
+                        .stderr => vm.rareData().stderr(),
+                        .stdout => vm.rareData().stdout(),
+                        else => null,
+                    };
+
+                    if (optional_store) |store| {
+                        store.ref();
+                        return Blob.initWithStore(
+                            store,
                             globalThis,
-                        ),
-                        .stderr => return Blob.initWithStore(
-                            vm.rareData().stderr(),
-                            globalThis,
-                        ),
-                        .stdout => return Blob.initWithStore(
-                            vm.rareData().stdout(),
-                            globalThis,
-                        ),
-                        else => {},
+                        );
                     }
                     break :brk path_or_fd.*;
                 },
--- a/test/js/bun/io/bun-write.test.js
+++ b/test/js/bun/io/bun-write.test.js
@@ -310,6 +310,14 @@ it("Bun.write(Bun.stderr, 'new TextEncoder().encode(Bun.write STDERR TEST'))", a
   expect(await Bun.write(Bun.stderr, new TextEncoder().encode("\nBun.write STDERR TEST\n\n"))).toBe(24);
 });
 
+it("Bun.file(0) survives GC", async () => {
+  for (let i = 0; i < 10; i++) {
+    let f = Bun.file(0);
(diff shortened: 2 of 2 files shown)


Jarred-Sumner:
Thank you

argosphil:
> The previous code shared one Blob.Store object between several Blobs, which led to those Blob.Store objects being collected and freed prematurely.

Just to be clear, the new code still does that, it just updates the shared stores' refcounts properly.

PR #8744, opened 2024-02-07
fix(test) http2 memory leak

cirospaciari:
### What does this PR do?

http2 memory test improvement

- [ ] Code changes
- [x] Test changes

Test should be more reliable on CI/CD

### DIFF
--- a/test/js/node/http2/node-http2-memory-leak.js
+++ b/test/js/node/http2/node-http2-memory-leak.js
@@ -31,7 +31,7 @@ const PAYLOAD = Buffer.from("a".repeat(128 * 1024));
 const info = await nodeEchoServer();
 
 async function runGC() {
-  for(let i = 0; i < GC_AGRESSIVENESS; i++) {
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
technically Bun.gc(true) run twice in quick succession is supposed to do nothing

test/js/node/http2/node-http2-memory-leak.js
```diff
@@ -23,11 +23,20 @@ async function nodeEchoServer() {
 }
 // X iterations should be enough to detect a leak
 const ITERATIONS = 50;
+// how many time we should run
+const GC_AGRESSIVENESS = 3;
```
jdalton, line 27:
```suggestion
const GC_RUN_COUNT = 3;
```

jdalton:
I've noticed this issue attempt to be indirectly addressed in other PRs too. Is this PR good enough to push forward now?

Issue #8361, opened 2024-01-22
[windows] Implement fs.watch on Windows


Jarred-Sumner:

Jarred-Sumner:
@cirospaciari fs.watch test is panicking:

```
test\js\node\watch\fs.watch.test.ts:
  ✓ fs.watch > non-persistent watcher should not block the event loop [65.90ms]
  ✓ fs.watch > watcher should close and not block the event loop [73.78ms]
  ✓ fs.watch > unref watcher should not block the event loop [76.06ms]
  ✓ fs.watch > should work with relative files [81.34ms]
  
  Error: 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
        at 
  TimeoutError: The operation timed out.
  ✗ fs.watch > add file/folder to folder [3000.59ms]
  Segmentation fault at address 0x42cf4513004
  ???:?:?: 0x7ff74b38054b in ??? (???)
  ???:?:?: 0x7ff74b37ffaf in ??? (???)
  ???:?:?: 0x7ff74ae5c068 in ??? (???)
  ???:?:?: 0x7ff74aa54b02 in ??? (???)
  ???:?:?: 0x7ff74c57dbda in ??? (???)
  ???:?:?: 0x7ff74a729eb7 in ??? (???)
  ???:?:?: 0x7ff74a78b4bc in ??? (???)
  ???:?:?: 0x7ff74a521492 in ??? (???)
  ???:?:?: 0x7ff74c8d12af in ??? (???)
  ???:?:?: 0x7ffd0b4e4ddf in ??? (KERNEL32.DLL)
  ???:?:?: 0x7ffd0b6bedaa in ??? (ntdll.dll)
  bun test v1.0.25-canary.27 (58260825)
```

Jarred-Sumner:
I'm able to reproduce in a debug build eventually:

call stack:
```js
bun-debug.exe!abort() Line 563 (c:/bun/.cache/zig/lib/std/os.zig:563)
bun-debug.exe!handleSegfaultWindowsExtra(os.windows.EXCEPTION_POINTERS * info, unsigned char msg, []u8 label) Line 2621 (c:/bun/.cache/zig/lib/std/debug.zig:2621)
[...]

PR #8746, opened 2024-02-07
[windows] fix(fs.watch)

cirospaciari:
### What does this PR do?

Fix: https://github.com/oven-sh/bun/issues/8361

- [x] Code changes

Existing Tests


test/js/node/watch/fs.watch.test.ts
```diff
@@ -405,28 +407,28 @@ describe("fs.watch", () => {
     expect(promise).resolves.toBe("change");
   });
 
-  test("should throw if no permission to watch the directory", async () => {
+  skipWindows("should throw if no permission to watch the directory", async () => {
```
gvilums: I feel like we shouldn't introduce another way to skip tests on windows. We have the `isWindows` constant in `harness.ts`, so these could just be `test.skipIf(isWindows)(...)`

test/js/node/watch/fs.watch.test.ts
```diff
@@ -74,7 +76,7 @@ describe("fs.watch", () => {
     watcher.on("change", (event, filename) => {
       count++;
       try {
-        expect(event).toBe("rename");
+        expect(["rename", "change"]).toContain(event);
```
gvilums, line 79: Is this necessary because windows emits multiple events?

src/bun.js/node/win_watcher.zig
```diff
@@ -242,34 +243,38 @@ pub const PathWatcher = struct {
...
+            }
+        }
         defer this.flush();
         // events always use the relative path
         if (events & uv.UV_RENAME != 0) {
             this.emit(path, path_info.hash, timestamp, path_info.is_file, .rename);
-        }
-
-        if (events & uv.UV_CHANGE != 0) {
+        } else if (events & uv.UV_CHANGE != 0) {
```
gvilums: I assume change is to prevent multiple events being emitted for a single file change, right? In that case, why do we need the multiple cases in the test?

gvilums (approved):

test/js/node/watch/fs.watch.test.ts
cirospaciari: actually because we use repeat with can emit multiple events

src/bun.js/node/win_watcher.zig
cirospaciari: I will simplify this

cirospaciari: Here https://github.com/oven-sh/bun/pull/8746/commits/bd7b7a2e59d96e15cd1359ee9738b28338e8fdfd

test/js/node/watch/fs.watch.test.ts
cirospaciari: Agreed fixed here https://github.com/oven-sh/bun/pull/8746/commits/bd7b7a2e59d96e15cd1359ee9738b28338e8fdfd

src/bun.js/node/win_watcher.zig
gvilums: Looks good

PR #8747, opened 2024-02-07
Fix FIFO data loss (#8695)

### LINKED ISSUE #8695: Bun.stdin.stream() loses chunks in async iteration

1.0.26+c75e768a6

In `bash`, run 
```
seq -f '%4095.4090f' 0 1000 | while read; do sleep .1; echo "$REPLY"; done | tee -a log | strace bun run ./bunbug5.js 2> strace
```
where `bunbug5.js` contains this code:

```
let offset = 0;
for await (let chunk of Bun.stdin.stream()) {
    console.log(offset, [...chunk.slice(0, 8)].map(x => String.fromCharCode(x)).join(""),
		offset / parseFloat([...chunk.slice(0, 8)].map(x => String.fromCharCode(x)).join("")));
    offset += chunk.length;
    Bun.gc(true); // optional, see below
    await new Promise(r => setTimeout(r, 100));
}
```

output that looks like this:
```
0    0.000 NaN
4096    1.000 4096
8192    2.000 4096
12288    3.000 4096
16384    4.000 4096
20480    5.000 4096
24576    6.000 4096
28672    7.000 4096
```

(that's what I get if I use `sleep 1` rather than `sleep .1`)

### What do you see instead?

This output:
```
0    0.000 NaN
4096    1.000 4096
8192    3.000 2730.6666666666665
12288    4.000 3072
16384    6.000 2730.6666666666665
20480    7.000 2925.714285714286
```

This indicates that the third 4096-byte line, which starts with "   2.000", has been lost. The same applies to the sixth 4096-byte line, starting with "   5.000". The `strace` and `log` outputs indicate that the line has been read by `bun`:

```
read(14, "   2.000000000000000000000000000"..., 65536) = 4096
...
ioctl(14, FIONREAD, [0])                = 0
read(14, 0x7fa80a420000, 65536)         = -1 EAGAIN (Resource temporarily unavailable)
...
read(14, "   3.000000000000000000000000000"..., 65536) = 4096
...
write(1, "8192    3.000 2730.6666666666665"..., 33) = 33
...
read(14, "   4.000000000000000000000000000"..., 65536) = 4096
```

The ioctl-read pair looks odd (we ask whether there's any data to read, there isn't, so we try to read 64KB anyway), and it happens precisely at the points when data is lost.

### Additional information

The first part of the pipeline generates lines of precisely 4096 characters; the second part reads those and echoes them back at a rate of one per 100 ms; the third part simply saves the log; the fourth part runs the script and saves its strace.

The script reads a chunk (which happens to be a line), and calculates whether the number of bytes read so far corresponds to 4096 times the number the line begins with. That should always be the case after the first iteration. It then delays for 100 ms, which is how long it will take (approximately) for the next line to arrive.

The call to `Bun.gc(true)` isn't strictly necessary.

argosphil:
Under certain circumstances, we would read data from FIFOs without a sensible place to put it, leading to dropped chunks in async iteration over Bun.stdin.stream().

Fixes #8695.

### What does this PR do?

This changes `FIFO.ready` so it doesn't read unconditionally from the FIFO's FD. Instead, we only do so if there is a pending Promise or handler that will receive the data.

- [x] Code changes

~This issue is sensitive to timing and I don't know how to write automated tests for it.~

I wrote a test.

### DIFF
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -4320,6 +4320,8 @@ pub fn NewFIFO(comptime EventLoop: JSC.EventLoopKind) type {
                 }
             }
 
+            if (this.pending.state != .pending) return;
+
             const read_result = this.read(
                 this.buf,
                 // On Linux, we end up calling ioctl() twice if we don't do this
--- a/test/js/bun/io/bun-write.test.js
+++ b/test/js/bun/io/bun-write.test.js
@@ -479,3 +479,18 @@ describe("ENOENT", () => {
     });
   });
 });
+
+it("timed output should work", async () => {
+  const producer_file = path.join(import.meta.dir, "timed-stderr-output.js");
+
+  const producer = Bun.spawn([bunExe(), "run", producer_file], {
+    stderr: "pipe",
+  });
+
+  let text = "";
+  for await (const chunk of producer.stderr) {
+    text += [...chunk].map(x => String.fromCharCode(x)).join("");
+    await new Promise(r => setTimeout(r, 1000));
+  }
+  expect(text).toBe("0\n1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15\n16\n17\n18\n19\n20\n");
+});
--- a/test/js/bun/io/timed-stderr-output.js
+++ b/test/js/bun/io/timed-stderr-output.js
@@ -1,4 +1,4 @@
 for (let i = 0; i <= 20; i++) {
-  console.error(i);
+  await Bun.write(Bun.stderr, i + "\n");
   await new Promise(r => setTimeout(r, 100));
 }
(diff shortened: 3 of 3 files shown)


argosphil:
Added a test which detects #8695 .

argosphil:
Looks like `console.error` adds escape sequences to its output even when it goes into a pipe. It doesn't do that locally, for some reason, but I assume `Bun.write(Bun.stderr, ...)` doesn't inject the escape sequences.

gvilums:
We have a big WIP PR (#8456) that's supposed to fix a lot of the stuff related to reading from pipes and spawning processes, but I just verified that this issue can be reproduced there, too. I think it makes sense to try to fix it there immediately, I will take a look tomorrow.

argosphil:
> We have a big WIP PR (#8456) that's supposed to fix a lot of the stuff related to reading from pipes and spawning processes, but I just verified that this issue can be reproduced there, too. I think it makes sense to try to fix it there immediately, I will take a look tomorrow.

Thanks for checking! I think the test from this PR might be salvageable, but feel free to close it if you prefer, I can open a new PR once the dust has settled.

gvilums:
Yeah the test is good, I'll move it over to the new PR so that we make sure to cover it there

argosphil:
Okay, closing this, then.

PR #8755, opened 2024-02-07
fix glob example to be recursive as stated

jakeg:
### What does this PR do?

- [ ] Code changes

The doc page has a comment saying it will search recursively, but the example didn't search recursively. Now it does.


Jarred-Sumner:
Thank you

PR #8764, opened 2024-02-07
Fix watch attempting to reuse ports that are not yet released by the old process

gvilums:
### What does this PR do?
This PR changes how `--watch` mode reloads work on windows. Previously, we would have one top-level process wait on a job object containing the actual bun process, and exit once all children have exited. The child process would be the one actually spawning a copy and terminating itself.

Now, if the child process wants to reload itself, it exits with a magic exit code that is picked up by the top-level process, which in turn restarts the child.

With the new setup, the process is only restarted once the previous copy has exited. This should fix issues where ports were still in use by the old version of a process when the new copy was spawned.

Manual testing.

### DIFF
--- a/src/__global.zig
+++ b/src/__global.zig
@@ -92,10 +92,14 @@ pub fn runExitCallbacks() void {
 
 /// Flushes stdout and stderr and exits with the given code.
 pub fn exit(code: u8) noreturn {
+    exitWide(@as(u32, code));
+}
+
+pub fn exitWide(code: u32) noreturn {
     runExitCallbacks();
     Output.flush();
     std.mem.doNotOptimizeAway(&Bun__atexit);
-    std.c.exit(code);
+    std.c.exit(@bitCast(code));
 }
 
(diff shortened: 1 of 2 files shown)


src/__global.zig
```diff
@@ -92,10 +92,14 @@ pub fn runExitCallbacks() void {
...
 pub fn exit(code: u8) noreturn {
+    exitWide(@as(u32, code));
+}
+
+pub fn exitWide(code: u32) noreturn {
     runExitCallbacks();
     Output.flush();
     std.mem.doNotOptimizeAway(&Bun__atexit);
-    std.c.exit(code);
+    std.c.exit(@bitCast(code));
```
Jarred-Sumner, line 102: This will fail to compile on Linux, no?

Jarred-Sumner: I think you want `@intCast`

gvilums: std.c.exit takes a `c_int` on all platforms I think

gvilums: or wait, maybe not

PR #8766, opened 2024-02-07
fix a bunch of fetch tests on windows

gvilums:
### What does this PR do?
Updates a bunch of the fetch tests to connect to `localhost` on windows instead of `0.0.0.0` (which is invalid).

Also updates a `chmod` test to be skipped on windows.

The tests now pass.

### DIFF
--- a/test/js/web/fetch/fetch.stream.test.ts
+++ b/test/js/web/fetch/fetch.stream.test.ts
@@ -28,6 +28,9 @@ const bigText = Buffer.from("a".repeat(1 * 1024 * 1024));
 const smallText = Buffer.from("Hello".repeat(16));
 const empty = Buffer.alloc(0);
 
+const listen_host = "0.0.0.0";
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (changes requested):
The bug is that `server.hostname` is wrong, not the tests

Jarred-Sumner:
We shouldn't return a hostname which cannot be connected to

test/js/web/fetch/fetch.stream.test.ts
```diff
@@ -55,7 +58,7 @@ describe("fetch() with streaming", () => {
           },
         });
 
-        const server_url = `http://${server.hostname}:${server.port}`;
+        const server_url = `http://${connect_host}:${server.port}`;
```
cirospaciari, line 61: if it is a `Bun.serve` we can use `server.url.href` so IPv6 will be formated for url

image

gvilums:
Closed in favor of #8770

PR #8769, opened 2024-02-07
Fix async module mocking in bun:test

Jarred-Sumner:
### What does this PR do?

Fix async module mocking in bun:test

There is a test


paperclover (approved):
sure

PR #8772, opened 2024-02-07
Chore - quick note on dependencies for C++

jcarpe:
### What does this PR do?

Adds a small note on C++ standard lib dependencies when setting up bun.

- [ ] Code changes

### DIFF
--- a/docs/project/contributing.md
+++ b/docs/project/contributing.md
@@ -105,7 +105,7 @@ $ export PATH="$PATH:/usr/lib/llvm16/bin"
 
 {% /codetabs %}
 
-> ⚠️ Depending on your system, you may need to install the C++ standard library independently. See the [troubleshooting section](#span-file-not-found-on-ubuntu) for more information.
+> ⚠️ Ubuntu distributions may require installation of the C++ standard library independently. See the [troubleshooting section](#span-file-not-found-on-ubuntu) for more information.
 
 ## Building Bun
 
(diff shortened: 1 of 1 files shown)


paperclover (changes requested):
this is already covered in the troubleshooting section

> `### 'span' file not found on Ubuntu`

im re-reading this and it is a bit more verbose than it honestly needs to be. We should *move* that section up to where you put it, and probably make it a bit shorter so it isnt a hassle for non-ubuntu/debian users.

this way it will be harder to miss the instructions

jcarpe:
@paperdave  Just to clarify - you're saying it is Ubuntu/Deb specific and therefor should be more discreet, but _in_ the contributing page?

Now that I am seeing the troubleshooting, I like the information better but interestingly never actually had the `'span" file not found` issue. It was more explicitly clang not finding the proper standard lib (`libstdc++` in my case).

paperclover:
i think we should move that section up into where you added your note, and trim it down just a bit.

if you think it should be as long as it is, we can also put it in a details block.

```
{% details summary="For Ubuntu/Debian" %}

content

{% /details %}
```

I am 99% certain this situation is specific to the default configuration of Ubuntu.

paperclover:
also want to note that both of these are in the contributing page

jcarpe:
Ultimately I made it more wordy...

I reduced the content to a block quote inline that notes the standard library issue for Ubuntu and anchor linked down to the troubleshooting section. I then _added_ the Clang error that manifests for those who are simply running `bun setup` and getting the noted Clang error.

paperclover:
thank you, i think this is good. ill merge once i double check this renders properly for the site (the doc site itself isnt public yet)

paperclover (approved):

paperclover:
awesome

PR #8773, opened 2024-02-07
fix: crash in release builds

dylan-conway:
### What does this PR do?
Fixes the ~27 test failures on main. Only concern: this seems to be a zig bug and we upgraded zig on Dec 16th. Why are we seeing this now?

ci

### DIFF
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -20005,7 +20005,7 @@ fn NewParser_(
                                 // and unreachable or noreturn else.
                                 else => null,
                             };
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
cc @Vexu incase you find this interesting

paperclover:
I wonder when we upgrade to the latest zig if this can be reverted

Vexu:
> cc @Vexu incase you find this interesting

I do find potential Zig bugs interesting. What kinds of issues was this causing? And if it didn't occur when you upgraded I'd be interested to find out if it goes away the next time.

PR #8778, opened 2024-02-08
Ignore panic in reload

Jarred-Sumner:
### What does this PR do?

Ignore when a panic happens while the process is reloading in another thread


dylan-conway (approved):
Looks good

PR #8784, opened 2024-02-08
[windows] fix(fetch) file:///, keepalive and tests on Windows

cirospaciari:
### What does this PR do?

- [X] Code changes

Existing tests

### DIFF
--- a/src/bun.js/webcore/response.zig
+++ b/src/bun.js/webcore/response.zig
@@ -2167,10 +2167,10 @@ pub const Fetch = struct {
                 }
 
                 var cwd_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
-                const cwd = std.os.getcwd(&cwd_buf) catch |err| {
+                const cwd = if (Environment.isWindows) (std.os.getcwd(&cwd_buf) catch |err| {
                     globalThis.throwError(err, "Failed to resolve file url");
                     return .zero;
-                };
+                }) else globalThis.bunVM().bundler.fs.top_level_dir;
 
                 const fullpath = bun.path.joinAbsStringBuf(
                     cwd,
--- a/src/bun.js/webcore/response.zig
+++ b/src/bun.js/webcore/response.zig
@@ -54,6 +54,7 @@ const Async = bun.Async;
 
 const BoringSSL = bun.BoringSSL;
 const X509 = @import("../api/bun/x509.zig");
+const PosixToWinNormalizer = bun.path.PosixToWinNormalizer;
 
 pub const Response = struct {
     const ResponseMixin = BodyMixin(@This());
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
Which tests does this fix?

cirospaciari:
> Which tests does this fix?

`test\js\web\fetch\fetch.test.ts`

http1.1 keepalive and without content-length or transfer encoding
fetch() file://

src/bun.js/webcore/response.zig
```diff
@@ -2141,24 +2142,55 @@ pub const Fetch = struct {
...
             var stream = std.io.fixedBufferStream(&path_buf2);
-            const url_path_decoded = path_buf2[0 .. PercentEncoding.decode(
+            var url_path_decoded = path_buf2[0 .. PercentEncoding.decode(
                 @TypeOf(&stream.writer()),
                 &stream.writer(),
                 url.path,
             ) catch {
                 globalThis.throwOutOfMemory();
                 return .zero;
             }];
```
zackradisic: It looks this `decode` function returns `error.DecodingError` and not `error.OutOfMemory` so it might be better to capture the error value and do: `globalThis.throwError(err, ...)`

cirospaciari: Yeah makes much more sense fixed here: https://github.com/oven-sh/bun/commit/36ddac0b644b1087e76cac906748fb283a552a7d

src/bun.js/webcore/response.zig
```diff
@@ -2141,24 +2142,55 @@ pub const Fetch = struct {
...
+                        break :brk PosixToWinNormalizer.resolveCWDWithExternalBufZ(&path_buf, url_path_decoded) catch |err| {
+                            globalThis.throwError(err, "Failed to resolve file url");
+                            return .zero;
+                        };
+                    }
+                    break :brk url_path_decoded;
+                }
+
+                var cwd_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
+                const cwd = std.os.getcwd(&cwd_buf) catch |err| {
```
Jarred-Sumner: Is this necessary on posix?

cirospaciari: Fix here: https://github.com/oven-sh/bun/pull/8784/commits/4fd215724bfae7389d190aa041a050c592f77bd2

src/http.zig
```diff
@@ -1556,7 +1554,9 @@ pub fn isKeepAlivePossible(this: *HTTPClient) bool {
         if (this.http_proxy != null and this.url.isHTTPS()) {
             return false;
         }
-        return !this.disable_keepalive;
+
+        //check state
```
Jarred-Sumner, line 1558: what is this comment trying to explain?

cirospaciari: It's pretty unnecessary, but saying that we are checking `this.state.allow_keepalive` (this comments was also present in the removed function I just keep it)

PR #8792, opened 2024-02-08
correctly unblock all signals

### LINKED ISSUE #8786: Server keeps running in background after closing or stopping command.

1.0.26

Linux - Ubuntu

First of all, I apologize for my bad English.

Before I was using version 1.0.26, the server would close itself when I closed the command prompt or stopped it, but after updating the version, the server is still running.

I'm using watch mode, and this happens after the first save.

The example file:

server.ts

```typescript
import { Hono } from 'hono'

const app = new Hono()

app.get('/', () => {
    
    return Response.json({ message: 'Hello, World!' })
})

const server = Bun.serve({
    fetch: app.fetch,
    port: process.env.PORT || 5000
})

console.log(`Server is running in ${server.url.href} address.`)
```

and the command:

```bash
bun --watch server.ts
```

and the error message after the first save:

```bash
5 | app.get('/', () => {
 6 | 
 7 |     return Response.json({ message: 'Hello, World!' })
 8 | })
 9 | 
10 | const server = Bun.serve({
                    ^
EADDRINUSE: Failed to start server. Is port 5000 in use?
 syscall: "listen"
```

and after the error message I have to find for process PID and kill it.

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #8788: Can't terminate bun running in `--watch` mode with Ctrl+C

1.0.26-debug+2f8f77eeb

Linux - Ubuntu

```
Bun.serve({
  fetch() {
    return new Response("Hello world");
  },
});
console.log('listen');
```

run with `bun --watch`, then attempt to stop with Ctrl+C

Bun stops

### What do you see instead?

Nothing happens

### Additional information

This may be the root cause of #8786: People will attempt to end the bun process by terminating the shell, but this leaves bun running in the background, potentially blocking ports.

gvilums:
### What does this PR do?
Previously, the linux code attempting to unblock all signals before a reload was wrong. It accidentally _blocked_ all signals. This PR fixes this to actually _unblock_ all signals.

Fixes #8788 
Fixes #8786 

tests.

### DIFF
--- a/src/bun.js/bindings/c-bindings.cpp
+++ b/src/bun.js/bindings/c-bindings.cpp
@@ -185,7 +185,7 @@ extern "C" void on_before_reload_process_linux()
 
     // reset all signals to default
     sigset_t signal_set;
-    sigfillset(&signal_set);
+    sigemptyset(&signal_set);
     sigprocmask(SIG_SETMASK, &signal_set, nullptr);
 }
 


paperclover (approved):
damn

PR #8796, opened 2024-02-08
fix: flaky test

sroussey:
This often fails in CI (and on my machine) non-deterministically.

The 1ms + the 5000 is likely setting up a race condition in tasks/microtasks. Just changing it to to +10ms instead of +1ms is enough.

- [x] Code changes


Electroid:
Thanks!

PR #8797, opened 2024-02-08
[windows] fix crypto

cirospaciari:
### What does this PR do?

Fix `crypto.key-objects.test.ts`

- [x] Code changes

Existing tests

### DIFF
--- a/src/bun.js/bindings/KeyObject.cpp
+++ b/src/bun.js/bindings/KeyObject.cpp
@@ -128,15 +128,49 @@ struct AsymmetricKeyValueWithDER {
     long der_len;
 };
 
-struct PrivateKeyPassphrase {
+class KeyPassphrase {
+private:
+    WTF::CString pass_holder;
+
+public:
     char* passphrase;
     size_t passphrase_len;
+    bool hasPassphrase;
+    KeyPassphrase(JSValue passphraseJSValue, JSC::JSGlobalObject* globalObject, JSC::ThrowScope& scope)
+    {
+        this->passphrase = nullptr;
+        this->passphrase_len = 0;
+        if (passphraseJSValue.isUndefinedOrNull() || passphraseJSValue.isEmpty()) {
+            this->hasPassphrase = false;
+            return;
+        }
+        this->hasPassphrase = true;
+        if (passphraseJSValue.isString()) {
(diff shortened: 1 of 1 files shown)


src/bun.js/bindings/KeyObject.cpp
```diff
@@ -381,6 +381,7 @@ JSC::EncodedJSValue KeyObject__createPrivateKey(JSC::JSGlobalObject* globalObjec
                         auto value = pass.value();
                         passphrase.passphrase = const_cast<char*>(value.data());
                         passphrase.passphrase_len = value.length();
+                        pass_holder = std::move(value);
```
Jarred-Sumner: This is a use-after-move. 

Maybe PrivateKeyPassphrase should have a wrapper class which owns the WTF::CString?

Jarred-Sumner (changes requested):
A comment

src/bun.js/bindings/KeyObject.cpp
cirospaciari: fixed here: https://github.com/oven-sh/bun/pull/8797/commits/5be7384f3a4651e11858c95893c4529591e3e0c4

src/bun.js/bindings/KeyObject.cpp
```diff
@@ -128,15 +128,49 @@ struct AsymmetricKeyValueWithDER {
...
+        this->passphrase = nullptr;
+        this->passphrase_len = 0;
+        if (passphraseJSValue.isUndefinedOrNull() || passphraseJSValue.isEmpty()) {
+            this->hasPassphrase = false;
+            return;
+        }
+        this->hasPassphrase = true;
+        if (passphraseJSValue.isString()) {
+            auto passphrase_wtfstr = passphraseJSValue.toWTFString(globalObject);
+            RETURN_IF_EXCEPTION(scope, );
```
Jarred-Sumner, line 150:
```suggestion
            RETURN_IF_EXCEPTION(scope, void());
```

Jarred-Sumner (approved):

src/bun.js/bindings/KeyObject.cpp
cirospaciari: Will not compile with void()

PR #8799, opened 2024-02-09
Fix some errors not showing up in GitHub annotations

Electroid:
### What does this PR do?

- Removes dead code that checked logic for preventing `expect()` outside a test
- Changes usages in `jest.zig` and `test_command.zig` to use `vm.onUnhandledError`
- Fixes errors such as timeouts, errors throw in the global scope, and errors thrown in lifecycle hooks from not print a GitHub annotation

Added tests

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -579,8 +579,6 @@ pub const VirtualMachine = struct {
     onUnhandledRejectionCtx: ?*anyopaque = null,
     unhandled_error_counter: usize = 0,
 
-    on_exception: ?*const OnException = null,
-
     modules: ModuleLoader.AsyncModule.Queue = .{},
     aggressive_garbage_collection: GCLevel = GCLevel.none,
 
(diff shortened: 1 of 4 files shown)


src/bun.js/javascript.zig
```diff
@@ -3007,6 +2995,127 @@ pub const VirtualMachine = struct {
...
+            const vm = VirtualMachine.get();
+            const origin = if (vm.is_from_devserver) &vm.origin else null;
+
+            var i: i16 = 0;
+            while (i < frames.len) : (i += 1) {
+                const frame = frames[@as(usize, @intCast(i))];
+                const source_url = frame.source_url.toUTF8(allocator);
+                defer source_url.deinit();
+                const file = bun.path.relative(dir, source_url.slice());
+                const func = frame.function_name.toUTF8(allocator);
```
Jarred-Sumner, line 3077: memory leak

src/bun.js/test/jest.zig
```diff
@@ -1427,10 +1377,12 @@ pub const TestRunnerTask = struct {
         const test_ = this.describe.tests.items[test_id];
         var describe = this.describe;
         describe.tests.items[test_id] = test_;
+
         if (comptime from == .timeout) {
-            Output.prettyErrorln("<r><red>Timeout<r><d>:<r> test <b>{}<r> timed out after {d}ms", .{ bun.fmt.quote(test_.label), test_.timeout_millis });
-            Output.flush();
+            const err = this.globalThis.createErrorInstance("Test {} timed out after {d}ms", .{ bun.fmt.quote(test_.label), test_.timeout_millis });
```
Jarred-Sumner, line 1382: the `name` field should be `TimeoutError` so you can clearly see its a test timeout without thinking much

PR #8801, opened 2024-02-09
fix: allow opening win32 device files in node:fs and bun.file

### LINKED ISSUE #8248: [win][fs] Fails to open Win32 device namespace verbatim prefixed filepaths (`\\.\`, `\\.\UNC\`)

1.0.23

Window 10

```js
import fs from "node:fs/promises";

const filenames = [
    String.raw`c:\temp\test-file.txt`,
    String.raw`\\127.0.0.1\c$\temp\test-file.txt`,
    String.raw`\\LOCALHOST\c$\temp\test-file.txt`,
    String.raw`\\.\c:\temp\test-file.txt`,
    String.raw`\\?\c:\temp\test-file.txt`,
    String.raw`\\.\UNC\LOCALHOST\c$\temp\test-file.txt`,
    String.raw`\\?\UNC\LOCALHOST\c$\temp\test-file.txt`,
    String.raw`\\127.0.0.1\c$\temp\test-file.txt`,
];

for (const filename of filenames) {
    try {
        const stats = await fs.stat(filename);
        console.log(filename, stats.size);
    } catch (e) {
        console.error(filename);
        console.error(e);
    }
}
```

```
c:\temp\test-file.txt 6
\\127.0.0.1\c$\temp\test-file.txt 6      
\\LOCALHOST\c$\temp\test-file.txt 6      
\\.\c:\temp\test-file.txt 6              
\\?\c:\temp\test-file.txt 6              
\\.\UNC\LOCALHOST\c$\temp\test-file.txt 6
\\?\UNC\LOCALHOST\c$\temp\test-file.txt 6
\\127.0.0.1\c$\temp\test-file.txt 6  
```

### What do you see instead?

```
c:\temp\test-file.txt 6
\\127.0.0.1\c$\temp\test-file.txt 6
\\LOCALHOST\c$\temp\test-file.txt 6
\\.\c:\temp\test-file.txt
   errno: -4058
 syscall: "stat"


\\?\c:\temp\test-file.txt 6
\\.\UNC\LOCALHOST\c$\temp\test-file.txt
   errno: -4058
 syscall: "stat"


\\?\UNC\LOCALHOST\c$\temp\test-file.txt 6
\\127.0.0.1\c$\temp\test-file.txt 6
```

### Additional information

It can open paths prefixed with `\\?\`, or `\\?\UNC\`, but not with `\\.\`, or `\\.\UNC\`.
Node.js works OK with them.

https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-device-namespaces
https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#example-ways-to-refer-to-the-same-file
https://doc.rust-lang.org/std/path/enum.Prefix.html#variants

Rel: 
- https://github.com/oven-sh/bun/issues/8246#issuecomment-1897785855

paperclover:
### What does this PR do?

Fixes #8248


test/js/node/fs/fs.windows.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    `\\\\LOCALHOST\\${drive}$\\${file}`,
+    `\\\\.\\${drive}:\\${file}`,
+    `\\\\?\\${drive}:\\${file}`,
+    `\\\\.\\UNC\\LOCALHOST\\${drive}$\\${file}`,
+    `\\\\?\\UNC\\LOCALHOST\\${drive}$\\${file}`,
+    `\\\\127.0.0.1\\${drive}$\\${file}`,
+  ];
+
+  for (const filename of filenames) {
+    test(`Can read '${filename}' with node:fs`, async () => {
```
gvilums: we should just do `test.if(isWindows)(..)` so that we don't have an ever-growing number of ways in which we do platform-specific tests.

src/bun.js/WebKit
gvilums, line 1: Is the submodule update intentional?

paperclover: no it is not

jdalton:
TIL: path stuff inside `src/string_immutable.zig` is another place for path methods. Ideally we'd really locate these in the same place.

test/js/node/fs/fs.test.ts
```diff
@@ -2729,3 +2729,30 @@ it("fs.ReadStream allows functions", () => {
   // @ts-expect-error
   expect(() => new fs.ReadStream(".", {})).not.toThrow();
 });
+
+describe.if(isWindows)("windows path handling", () => {
```
jdalton, line 2733: Does the Bun test runner treat this like the skipped tests in that it displays a message on platforms where skipped (just to know that the test is there but not being run)?

paperclover: yes

paperclover:
> src/string_immutable.zig is another place for path methods. Ideally we'd really locate these in the same place.

imo we should split that file into
- general purpose string stuff
- unicode and utf8/utf16 stuff
- path stuff (combine with other path stuff in other files)

PR #8813, opened 2024-02-09
fix(windows): install bugfixes for workspaces, tarballs, and relative paths

dylan-conway:
### What does this PR do?
lockfiles still need to be tested being used across platforms

existing and new tests


src/cli/bunx_command.zig
```diff
@@ -634,7 +635,7 @@ pub const BunxCommand = struct {
             },
         }
 
-        absolute_in_cache_dir = std.fmt.bufPrint(&absolute_in_cache_dir_buf, bun.pathLiteral("{s}/node_modules/.bin/{s}"), .{ bunx_cache_dir, initial_bin_name }) catch unreachable;
+        absolute_in_cache_dir = std.fmt.bufPrint(&absolute_in_cache_dir_buf, bun.pathLiteral("{s}/node_modules/.bin/{s}{s}"), .{ bunx_cache_dir, initial_bin_name, bin_extension }) catch unreachable;
```
Jarred-Sumner, line 638: this can be ++ right? since its always known at comptime

src/cli/install_command.zig
```diff
@@ -4,6 +4,15 @@ const PackageManager = @import("../install/install.zig").PackageManager;
...
 pub const InstallCommand = struct {
     pub fn exec(ctx: Command.Context) !void {
-        try PackageManager.install(ctx);
+        PackageManager.install(ctx) catch |err| switch (err) {
+            error.InstallFailed,
+            error.InvalidPackageJSON,
+            => {
+                const log = &bun.CLI.Cli.log_;
+                log.printForLogLevel(bun.Output.errorWriter()) catch {};
+                bun.Global.exit(1);
```
Jarred-Sumner, line 13: Global.exit should really print the log level this is unfortunate

src/cli/bunx_command.zig
nektro: yep

src/fmt.zig
```diff
@@ -45,7 +45,7 @@ pub fn formatUTF16Type(comptime Slice: type, slice_: Slice, writer: anytype) !vo
     }
 }
 
-pub fn formatUTF16TypeEscapeBackslashes(comptime Slice: type, slice_: Slice, writer: anytype) !void {
+pub fn formatUTF16TypeWithPathOptions(comptime Slice: type, slice_: Slice, writer: anytype, opts: PathFormatOptions) !void {
```
Jarred-Sumner, line 48: can `opts` be comptime?

src/install/extract_tarball.zig
```diff
@@ -309,7 +309,7 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
     // e.g. @next
     // if it's a namespace package, we need to make sure the @name folder exists
     if (basename.len != name.len and !this.resolution.tag.isGit()) {
-        cache_dir.makeDir(std.mem.trim(u8, name[0 .. name.len - basename.len], "/")) catch {};
+        cache_dir.makePath(std.mem.trim(u8, name[0 .. name.len - basename.len], "/")) catch {};
```
Jarred-Sumner, line 312: is the trailign slash correct

should it be std.fs.path.sep

src/install/lockfile.zig
```diff
@@ -5124,16 +5148,58 @@ const Buffers = struct {
...
-                const remaining = this.dependencies.items;
+                const remaining = buffers.dependencies.items;
+
+                if (comptime Environment.allow_assert) {
+                    for (remaining) |dep| {
+                        switch (dep.version.tag) {
+                            .folder => {
+                                const folder = lockfile.str(&dep.version.value.folder);
+                                if (strings.containsChar(folder, std.fs.path.sep_windows)) {
+                                    std.debug.panic("workspace windows separator: {s}\n", .{folder});
```
Jarred-Sumner, line 5176: nice

test/cli/install/bun-install.test.ts
```diff
@@ -21,6 +20,26 @@ expect.extend({
   toBeWorkspaceLink,
   toBeValidBin,
   toHaveBins,
+  toHaveWorkspaceLink: async function (package_dir: string, [link, real]: [string, string]) {
+    const isWindows = process.platform === "win32";
+    if (!isWindows) {
+      // expect(await readlink(join(package_dir, "node_modules", "Bar"))).toBeWorkspaceLink(join("..", "bar"));
+      // expect(await readlink(join(package_dir, "node_modules", link))).toBeWorkspaceLink(join("..", real));
```
Jarred-Sumner, line 27: are the commented out lines intentional

src/install/install.zig
```diff
@@ -7639,7 +7657,15 @@ pub const PackageManager = struct {
             // add
             // remove
             outer: for (positionals) |positional| {
-                var input = std.mem.trim(u8, positional, " \n\r\t");
+                var input: []u8 = @constCast(std.mem.trim(u8, positional, " \n\r\t"));
+                {
+                    var temp: [2048]u8 = undefined;
+                    const len = std.mem.replace(u8, input, "\\\\", "/", &temp);
```
dylan-conway, line 7663: this should be `input[0..temp.len]` to be safe

src/fmt.zig
nektro: no because of the usage on L106

src/cli/bunx_command.zig
nektro: `bin_extension` is but `initial_bin_name` is not; could `++` to the string literal tho

src/install/install.zig
nektro: what would be the best way to surface an error? ideally if `input.len` > `temp.len` it should fail

src/install/extract_tarball.zig
nektro: yes because its from a normalized package name like `@foo/bar`
trimming basename gives `@foo/`
then trimming / gives `@foo`

PR #8814, opened 2024-02-09
make V8StackTraceIterator frame parser more robust

### LINKED ISSUE #8714: [windows] Fix source paths in error messages, it says `D:` as the file path
image

nektro:
Fixes #8714

fixes the more naive frame line parsing to be more resilient to windows paths and missing numbers

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -100,24 +100,6 @@
 #include "JavaScriptCore/InternalFieldTuple.h"
 #include "wtf/text/StringToIntegerConversion.h"
 
-//
-
-static unsigned StringView_count(WTF::StringView sv, UChar character, unsigned start)
-{
-    auto result = 0;
-    auto start_ = start;
-    while (true) {
-        if (sv.length() == 0)
-            break;
-        auto newfind = sv.find(character, start_);
-        if (newfind == WTF::notFound)
-            break;
-        result += 1;
-        start_ = newfind + 1;
-    }
-    return result;
-}
-
 static WTF::StringView StringView_slice(WTF::StringView sv, unsigned start, unsigned end)
 {
     return sv.substring(start, end - start);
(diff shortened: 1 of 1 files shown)


nektro:
windows traces after this change:

image

src/bun.js/bindings/bindings.cpp
```diff
@@ -4048,37 +4072,79 @@ class V8StackTraceIterator {
             return false;
         }
 
-        auto firstColon = line.find(':', openingParenthese + 1);
+        auto lineInner = StringView_slice(line, openingParenthese + 1, closingParenthese);
+        auto colonCount = StringView_count(lineInner, ':', 0);
```
Jarred-Sumner: This is going to make it a _lot_ slower. Do we really need an exact count? Or can we get away with finding the last 2 colons?

src/bun.js/bindings/bindings.cpp
```diff
@@ -4048,37 +4072,79 @@ class V8StackTraceIterator {
...
-            // at foo (/path/to/file.js:
-            frame.sourceURL = line.substring(openingParenthese + 1, firstColon - openingParenthese - 1);
+        case 1: {
+            // /path/to/file.js:
+            // /path/to/file.js:1
+            // node:child_process
+            // C:\Users\dave\bun\file.js
+
+            auto marker1 = 0;
+            auto marker2 = lineInner.find(':', marker1);
```
Jarred-Sumner, line 4061: First we get a count which gets the index of every colon (on line 4076)

Then we get the first colon again right here

Do we need to duplicate this work?

src/bun.js/bindings/bindings.cpp
```diff
@@ -4048,37 +4072,79 @@ class V8StackTraceIterator {
...
+            // /path/to/file.js:1:
+            // /path/to/file.js:1:2
+            // node:child_process:1:2
+            // C:\Users\dave\bun\file.js:
+            // C:\Users\dave\bun\file.js:1
+            // C:\Users\dave\bun\file.js:1:2
+
+            auto linenoColon = lineInner.find(':', 0);
+            for (size_t i = 0; i < colonCount - 2; i++) {
+                linenoColon = lineInner.find(':', linenoColon + 1);
```
Jarred-Sumner: This finds every colon until the 2nd to last colon

That's a lot of duplicate work

Jarred-Sumner (changes requested):
Please find a way to do this without making the stack trace parser slower

error.stack is used very frequently by HTTP libraries so we have to be careful

src/bun.js/bindings/bindings.cpp
```diff
@@ -100,6 +100,11 @@
 #include "JavaScriptCore/InternalFieldTuple.h"
 #include "wtf/text/StringToIntegerConversion.h"
 
+static WTF::StringView StringView_slice(WTF::StringView sv, unsigned start, unsigned end)
+{
+    return sv.substring(start, end - start);
+}
```
jdalton, line 106: This is a great helper. Do we have a WTF like place for it. (... guessing this may be it)

src/bun.js/bindings/bindings.cpp
```diff
@@ -4048,38 +4054,78 @@ class V8StackTraceIterator {
...
-        auto firstColon = line.find(':', openingParenthese + 1);
+        auto lineInner = StringView_slice(line, openingParenthese + 1, closingParenthese);
 
-        // if there is no colon, that means we only have a filename and no line number
+        {
+            auto marker1 = 0;
+            auto marker2 = lineInner.find(':', marker1);
 
-        //   at foo (native)
-        if (firstColon == WTF::notFound) {
```
jdalton, line 4056: I find the comments that bread crumb where we are in the parsing of the stack trace helpful. The more the merrier. We should also add a note that if the stack trace format changes this is also needs updating. Or... can we comment referencing the V8 code we are mimicking.

src/bun.js/bindings/bindings.cpp
```diff
@@ -4048,38 +4054,78 @@ class V8StackTraceIterator {
...
-            frame.sourceURL = line.substring(openingParenthese + 1, closingParenthese - openingParenthese - 1);
-        } else {
-            // at foo (/path/to/file.js:
-            frame.sourceURL = line.substring(openingParenthese + 1, firstColon - openingParenthese - 1);
-        }
+            if (marker2 == WTF::notFound) {
```
jdalton, line 4063: Can you add a comment somewhere (relevant) that explains why we are doing the search this way (making it more robust). This will help future us when we want to try to optimize smth and we can see oohh ya we do it this way for cross-platform goodness.

jdalton:
Does this unblock a unit test? Is there a test for this robustness (like a test on Windows or something). 
(Indirectly asking if we have a test to cover this and if not to please add one 🥺 )

Jarred-Sumner:
I'm going to merge this but @nektro please address the feedback in a follow-up PR

PR #8824, opened 2024-02-09
fix(crypto) fix KeyObjectkey handling

cirospaciari:
### What does this PR do?

- [x] Code changes

Existing tests

### DIFF
--- a/src/bun.js/bindings/KeyObject.cpp
+++ b/src/bun.js/bindings/KeyObject.cpp
@@ -130,21 +130,64 @@ struct AsymmetricKeyValueWithDER {
 
 class KeyPassphrase {
 private:
-    WTF::CString pass_holder;
+    union {
+        WTF::CStringBuffer* strBuffer;
+        JSUint8Array* array;
+    } u;
+    uint8_t tag;
 
 public:
-    char* passphrase;
-    size_t passphrase_len;
-    bool hasPassphrase;
+    bool hasPassphrase()
+    {
+        return tag > 0;
+    }
+
+    char* data()
+    {
+        switch (tag) {
+        case 1: {
+            if (this->u.strBuffer) {
+                return const_cast<char*>(this->u.strBuffer->data());
+            }
+            return nullptr;
+        }
+        case 2: {
+            if (this->u.array) {
+                return (char*)(this->u.array->vector());
+            }
+            return nullptr;
+        }
+        default:
+            return nullptr;
+        }
+    }
+
+    size_t length()
+    {
+        switch (tag) {
+        case 1: {
+            if (this->u.strBuffer) {
+                return this->u.strBuffer->length();
+            }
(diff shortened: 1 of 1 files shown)


src/bun.js/bindings/KeyObject.cpp
```diff
@@ -130,53 +130,119 @@ struct AsymmetricKeyValueWithDER {
 
 class KeyPassphrase {
 private:
-    WTF::CString pass_holder;
+    union {
+        WTF::CStringBuffer* strBuffer;
+        JSUint8Array* array;
+    } u;
+    uint8_t tag;
```
gvilums: I think this should be an enum with a default value to make it clear what each tag value is referring to, e.g.
```cpp
enum class Tag {
  None,
  StrBuffer,
  Array,
}
```
We could then add an initializer:
`Tag tag = Tag::None;`

src/bun.js/bindings/KeyObject.cpp
```diff
@@ -130,53 +130,119 @@ struct AsymmetricKeyValueWithDER {
...
+    {
+        return tag > 0;
+    }
+
+    char* data()
+    {
+        switch (tag) {
+        case 1: {
+            if (this->u.strBuffer) {
+                return const_cast<char*>(this->u.strBuffer->data());
```
gvilums: Is this `const_cast` safe? If we were to write to it and the underlying data was const we'd introduce UB

src/bun.js/bindings/KeyObject.cpp
```diff
@@ -130,53 +130,119 @@ struct AsymmetricKeyValueWithDER {
...
+            }
+            return nullptr;
+        }
+        case 2: {
+            if (this->u.array) {
+                return (char*)(this->u.array->vector());
```
gvilums: Why a C-style cast here?

gvilums (changes requested):

src/bun.js/bindings/KeyObject.cpp
cirospaciari: Is just a void* casted to char* I think it already was using C style I will change that

src/bun.js/bindings/KeyObject.cpp
cirospaciari: Is safe internally is actually mutable data, and the data is not actually mutate in any manner anyway

Jarred-Sumner:
From this diff, it’s not clear to me why it was crashing before - which means it’s not clear why this fixes it. Which line was the fix?

cirospaciari:
> From this diff, it’s not clear to me why it was crashing before - which means it’s not clear why this fixes it. Which line was the fix?

buffer->ref() and buffer->deref() to keep it alive was the fix (increasing the buffer ref count and decreasing when going out of scope)

PR #8825, opened 2024-02-09
Add charset to text/css mime type

7f8ddd:
Standard: https://www.iana.org/assignments/media-types/text/css
Extra info: https://stackoverflow.com/questions/70643383/which-mime-types-contain-charset-utf-8-directive

tl;dr almost anything that is read as text should include the charset.

### What does this PR do?

Fixes the warning in Edge that says to include utf-8 when Bun returns a stylesheet.

image

Added test to ensure Bun.file("something.css") returns the correct mime type.

- [x] Code changes


paperclover:
can you add a test to ensure bun.file("anything.css") actually returns the intended mimetype

paperclover (approved):

Issue #8498, opened 2024-01-26
zeromq module not working


scottstanfield:
### What version of Bun is running?

1.0.26

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```
bun init
bun add zeromq
echo "require('zeromq')" > index.ts
bun index.ts
```

### What is the expected behavior?

Expecting `zeromq` module to load (and no output). 

### What do you see instead?

```
dyld[19197]: missing symbol called
zsh: killed     bun index.ts
```

### Additional information

Works in node (v21). This appears to be a regression as the same issue was marked closed in #6313 

```
~/code/rs/rtfd/bunbug ❯ node
Welcome to Node.js v21.6.1.
Type ".help" for more information.
> require("zeromq");
{
  capability: [Getter],
  context: [Getter],
  curveKeyPair: [Getter],
  version: [Getter],
  Context: [Getter],
  Socket: [Getter],
  Observer: [Getter],
  Proxy: [Getter],
  Pair: [class Pair extends Socket],
  Publisher: [class Publisher extends Socket],
  Subscriber: [class Subscriber extends Socket],
  Request: [class Request extends Socket],
  Reply: [class Reply extends Socket],
  Dealer: [class Dealer extends Socket],
  Router: [class Router extends Socket],
  Pull: [class Pull extends Socket],
  Push: [class Push extends Socket],
  XPublisher: [class XPublisher extends Socket],
  XSubscriber: [class XSubscriber extends Socket],
  Stream: [class Stream extends Socket]
}
>
```

scottstanfield:
Also tested with bun v1.0.17 (went back to binary chop to a version that works, but no luck yet).

nk1tz:
Can confirm this is still an issue.

bun -v `1.0.25`
`"zeromq": "6.0.0-beta.19"`

```
dyld[28564]: missing symbol called
[1]    28564 killed     bun run index.ts
```

I have a minimal bug reproduction repo here: https://github.com/nk1tz/bun-zeromq-bug-repro which also includes a macOS crash report.

I also included a `docker-compose` way of running which has a different failure mode (I don't know if related) in case that's helpful.

The zeromq install script appears to run successfully during `bun i --verbose`, I don't see any obvious errors there.

grrowl:
~seems to work after `bun upgrade --canary`~

my bad, can install on canary but it doesn't work at runtime

nk1tz:
[...]

PR #8827, opened 2024-02-09
remove zeromq from default-trusted-deps

nk1tz:
### What

Remove `zeromq` from the list of default-trusted-deps.
Bun crashes when running zeromq. 

ex:
```
dyld[28564]: missing symbol called
[1]    28564 killed     bun run index.ts
```

There is an open issue tracking this with all the details: https://github.com/oven-sh/bun/issues/8498

### DIFF
--- a/src/install/default-trusted-dependencies.txt
+++ b/src/install/default-trusted-dependencies.txt
@@ -362,6 +362,5 @@ xxhash
 yarn
 yo
 yorkie
-zeromq
 zlib-sync
 zopflipng-bin


Jarred-Sumner:
trusted dependencies has no relation to the runtime unless node isn't installed

is this causing a crash in `bun install`?

Electroid:
The issue is being tracked here: #8498

nk1tz:
> trusted dependencies has no relation to the runtime unless node isn't installed
> 
> is this causing a crash in `bun install`?

@Jarred-Sumner -> I take your point, I guess I'm asking wether `zeromq` should be in the list if it crashes bun at runtime? 
I'd much rather see the underlying issue resolved, cheers and thank you for all your work 🙏🏻

PR #8828, opened 2024-02-10
fix error message in bun.serve

### LINKED ISSUE #4555: Unintuitive error on invalid tls configuration

paperclover:
### What does this PR do?

Fixes #4555

This is one example why using `JSC.C.ExceptionRef` is a horrible idea. We should not be using these at any point in Bun.

This error in particular has caused many users a lot of confusion because a faulty error was printed.

The real error happens during SSLConfig.inJS (this should be named fromJS)

image

but the exception ref is not set since BunFile does not use exception refs, it uses the real exception system JSC provides.

so the null check fails, we fall down to this getter
image

the getter fails because we have an exception, and so on until `fetch`

image

the getter here fails, and we override the already thrown error with another error.

I've changed this to check for an exception, but that is a band-aid solution that only fixes this case.

We now print the actual first error.
image

To make this harder to hit I've also moved the check after the fetch error to make it impossible to hit the error in this way.

I would add an assertion to ensure we dont double-throw elsewhere but it seems this would make other tests fail (should be investigated one day)

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -5351,3 +5351,8 @@ extern "C" EncodedJSValue ExpectStatic__getPrototype(JSC::JSGlobalObject* global
 {
     return JSValue::encode(reinterpret_cast<Zig::GlobalObject*>(globalObject)->JSExpectStaticPrototype());
 }
+
+extern "C" bool JSGlobalObject__hasException(JSC::JSGlobalObject* globalObject)
+{
+    return DECLARE_CATCH_SCOPE(globalObject->vm()).exception() != 0;
+}
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -861,16 +861,6 @@ pub const ServerConfig = struct {
                 }
             }
 
-            if (arg.getTruthy(global, "tls")) |tls| {
-                if (SSLConfig.inJS(global, tls, exception)) |ssl_config| {
-                    args.ssl_config = ssl_config;
-                }
-
-                if (exception.* != null) {
-                    return args;
-                }
-            }
-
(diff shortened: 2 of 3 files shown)


src/bun.js/bindings/bindings.cpp
```diff
@@ -5351,3 +5351,8 @@ extern "C" EncodedJSValue ExpectStatic__getPrototype(JSC::JSGlobalObject* global
 {
     return JSValue::encode(reinterpret_cast<Zig::GlobalObject*>(globalObject)->JSExpectStaticPrototype());
 }
+
+extern "C" bool JSGlobalObject__hasException(JSC::JSGlobalObject* globalObject)
+{
+    return DECLARE_CATCH_SCOPE(globalObject->vm()).exception() != 0;
```
Jarred-Sumner, line 5357: this leaks memory i'm pretty sure

Jarred-Sumner: it does not leak memory

PR #8829, opened 2024-02-10
Prevent open but inactive http connections from keeping alive http server

gvilums:
### What does this PR do?
Previously, Bun.serve would keep the process alive after stop() if there was a tcp socket that kept the connection alive, even if no HTTP data was being transmitted (keepalive).

This PR makes changes that unref the socket between http requests, allowing the process to exit if the server has been stopped.

Manual testing.

### DIFF
--- a/packages/bun-uws/src/HttpContext.h
+++ b/packages/bun-uws/src/HttpContext.h
@@ -134,6 +134,8 @@ private:
 
         /* Handle HTTP data streams */
         us_socket_context_on_data(SSL, getSocketContext(), [](us_socket_t *s, char *data, int length) {
+            // ref the socket to make sure we process it entirely before it is closed
+            us_socket_ref(s);
 
             // total overhead is about 210k down to 180k
             // ~210k req/sec is the original perf with write in data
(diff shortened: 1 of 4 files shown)


packages/bun-uws/src/HttpContext.h
```diff
@@ -293,6 +295,10 @@ struct HttpContext {
 
             /* We need to uncork in all cases, except for nullptr (closed socket, or upgraded socket) */
             if (returnedSocket != nullptr) {
+                us_socket_t* returnedSocketPtr = (us_socket_t*) returnedSocket; 
+                /* We don't want open sockets to keep the event loop alive between HTTP requests */
+                us_socket_unref(returnedSocketPtr);
+
```
jdalton, line 301: @gvilums Is this something we want to push back upstream to `uNetworking/uWebSockets` and `uNetworking/uSockets`?

Example references:
https://github.com/uNetworking/uWebSockets/blob/master/src/HttpContext.h#L269-L272 
https://github.com/uNetworking/uSockets/blob/master/src/socket.c#L193-L231

gvilums: No, this is specific to how our event loop works: In our http server we keep a separate reference into the loop to keep it alive, which is why we need to unref the socket here to shut the server down. This does not make sense in general though as it would lead to the event loop randomly exiting without the server present to reference it.

PR #8833, opened 2024-02-10
Fix occasional test failure in setTimeout.test.js

argosphil:
### What does this PR do?

This modifies a timer test to call performance.now() before the critical section rather than from within it. It also fixes an incorrect expected value in the test.

- [x] Code changes

The test now succeeds reliably, even running it in a tight loop.

### DIFF
--- a/test/js/web/timers/setTimeout.test.js
+++ b/test/js/web/timers/setTimeout.test.js
@@ -148,9 +148,9 @@ it("Bun.sleep propagates exceptions", async () => {
 });
 
 it("Bun.sleep works with a Date object", async () => {
+  const now = performance.now();
   var ten_ms = new Date();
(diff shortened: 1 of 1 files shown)


argosphil:
#8834 is a similar issue which discusses changing Bun behavior to match the expectation of this test (and also get closer to documented behavior). I think we should apply both changes.

argosphil:
This is superseded by #8859 , which fixes Bun.sleep() so the test matches both @cirospaciari 's expectations and the documenation.

gvilums:
> This is superseded by #8859 , which fixes Bun.sleep() so the test matches both @cirospaciari 's expectations and the documenation.

Does this mean that this PR is good to close?

argosphil:
Closing, we'll probably fix #8834 another way.

jdalton:
Fixed by https://github.com/oven-sh/bun/pull/8950

Issue #8834, opened 2024-02-10
Bun.sleep() rounds incorrectly


argosphil:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Linux 6.1.31-gentoo-dist x86_64 13th Gen Intel(R) Core(TM) i7-1370P

### What steps can reproduce the bug?

Run this code (based on a test from setTimeout.test.js):

```
while (true) {
    let ten_ms = new Date();
    ten_ms.setMilliseconds(ten_ms.getMilliseconds() + 12);
    await Bun.sleep(ten_ms);
    let now = new Date();
    if (!(+ten_ms <= +now)) {
	console.error({ten_ms: +ten_ms, now: +now});
	break;
    }
}
```

### What is the expected behavior?

An infinite loop.

### What do you see instead?

A break, since eventually Bun.sleep() will return a millisecond before the specified instant has been reached.

### Additional information

The fix is simply adding:

```
    millisecondsValue = JSC::jsNumber(ceil(millisecondsValue.asNumber()));
```

in BunObject.cpp:functionBunSleep. This will also make Bun.sleep(.001) sleep for a millisecond, which seems like safe behavior.

See #8833.

jdalton:
Closed by https://github.com/oven-sh/bun/pull/8950

PR #8838, opened 2024-02-10
Fix; make expect(1).toBeGreaterThan(2) throw reliably

argosphil:
### What does this PR do?

This changes expect(...).toBeGreaterThan() to throw an exception when appropriate even when ANSI colors are not in use.

I noticed this while debugging #8833 / #8834 , and it prevented me from writing a test for it so far.

- [x] Code changes

I ran a local test, which falsely succeeded before the change and failed after it.

### DIFF
--- a/src/bun.js/test/expect.zig
+++ b/src/bun.js/test/expect.zig
@@ -1703,6 +1703,7 @@ pub const Expect = struct {
             globalObject.throw(comptime Output.prettyFmt(fmt, true), .{ expected_fmt, value_fmt });
             return .zero;
         }
+	globalObject.throw(comptime Output.prettyFmt(fmt, false), .{ expected_fmt, value_fmt });
         return .zero;
     }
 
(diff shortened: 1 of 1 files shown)


src/bun.js/test/expect.zig
```diff
@@ -1774,6 +1775,7 @@ pub const Expect = struct {
             globalObject.throw(comptime Output.prettyFmt(fmt, true), .{ expected_fmt, value_fmt });
             return .zero;
         }
+	globalObject.throw(comptime Output.prettyFmt(fmt, false), .{ expected_fmt, value_fmt });
```
Electroid: Please make sure to run the Zig formatter. If you use VSCode and our recommended extension it should do it for you.

Also, could we add some tests in `test/js/bun/test` to make sure this doesn't regress?

Electroid: Actually, I think we can pass on this test for now. it's a bit silly to test for this specifically. later we should add overall tests to check output with ansi, ansi + color, color, and none; but that's out of scope for this pr.

argosphil: Thanks for the zig formatter reminder, and sorry for not knowing about it.

I agree about the tests, FWIW. Adding a NO_COLOR option to the test matrix would also work.

Electroid (approved):
Thanks!

PR #8841, opened 2024-02-10
Fix a typo in child_process.js

dhbaird:
### What does this PR do?

Fixes a typo in `stdio` normalization in src/js/node/child_process.js.

- [x] Code changes


Electroid:
Looks good, could we add a test somewhere in `test/js/node` so this doesn't regress?

dhbaird:
Yes, doing now

dhbaird:
Regression test added. Let me know if its too wordy or any other changes you'd like to see. I was considering the possibility that it would be easy to add more test cases. Thanks!

dhbaird:
Actually, there's something specific I wanted to do, but didn't: A very simple statically linked C program like this, for reasons explained in the test comments:

```
int main() { pause(); return 0; }
```

I didn't do this though (not yet, anyways), and wanted to see if you had an opinion on it or could see a better option. Thanks!

Electroid:
Oops, I didn't intend to commit that.

I think the test is slightly over-engineered. If you remove it from the commit, I'll just merge this small fix.

jdalton:
Thank you @dhbaird for taking the time on this. This PR can be closed as a fix has landed via https://github.com/oven-sh/bun/pull/8929. I'm still interested in a test @dhbaird. If a simpler way can be found that would be a great follow-up.

dhbaird:
I can definitely do a simpler test; will do over the weekend. Thanks for the feedback @Electroid and @jdalton!

dhbaird:
Planning ahead, @Electroid / @jdalton within the framework (of bun), (how) can I build the simple C program mentioned above? It would help to eliminate certain race conditions relating to the child process itself opening and closing fds. What I would like to do is spawn that as a child, then use the parent to readdir() on `/proc/PID/fd` and then assert (expect) the fd numbers that get returned. Open to other ideas too. Thanks!

PR #8845, opened 2024-02-10
fix: bun test -t handling

### LINKED ISSUE #8842: bun test --test-name-pattern is broken

1.0.26+c75e768a6

1. `bun test -t '$a' test/cli/test/bun-test.test.ts` (match-nothing regex)
2. `bun test -t '.?' test/cli/test/bun-test.test.ts` (match-all regex)

1. run no tests, not even those inside a .each construct
2. run all tests

### What do you see instead?

1. the tests inside a .each construct are executed
2. the tests inside a .each construct are executed

### Additional information

Working on a PR.

argosphil:
### What does this PR do?

Fixes #8842.

This changes RegularExpression.matches to return true if the match starts at the beginning of the haystack string.

This changes jest.zig to apply the filter regex to tests which make use of the .each() construct, and skip them if appropriate.

- [x] Code changes

~Automated tests require additional changes to bun-test.test.ts which I think should be in a separate PR.~

EDIT: I have now added two basic tests to `bun-test.test.ts`.

- [x] I ran `zig fmt` (thanks again, @Electroid)

### DIFF
--- a/src/bun.js/bindings/RegularExpression.zig
+++ b/src/bun.js/bindings/RegularExpression.zig
@@ -40,7 +40,7 @@ pub const RegularExpression = opaque {
 
     // Simple boolean matcher
     pub inline fn matches(this: *RegularExpression, str: bun.String) bool {
-        return Yarr__RegularExpression__matches(this, str) > 0;
+        return Yarr__RegularExpression__matches(this, str) >= 0;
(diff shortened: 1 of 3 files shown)


Electroid:
Test failure are un-related. Thanks!

Issue #8544, opened 2024-01-28
Have bun.$ reject its promise if exitCode is not zero


mathiasrw:
### What is the problem this feature would solve?

Shell commands via bun.$ are always resolving so error handling needs to happen "manually" by looking at the errorCode in the response. The impact is that I cant use the expected way to deal with errors for async functions. 

It is not so much that error handling cant be done checking the value of errorCode, but it breaks with the philosophy of how things are done in JS.

### What is the feature you are proposing to solve the problem?

I propose changing the behaviour so results from bun.$ rejects when the value of errorCode is not-zero. This means that the below code would return "Command failed" as the binary echoXYZ could not be found. (This currently returns "Command ran OK" with Bun v 1.0.25)

```js
import { $ } from 'bun';

$`echoZYX Hello world`
	.then(o => console.log('Command ran OK: ', o))
	.catch(o => console.error('Command failed: ', o))
```

### What alternatives have you considered?

You can polifil this behaviour by wrapping $ like this

```js
import { $ as _$ } from 'bun';

const $ = async function (...args) {
	const result = await _$(...args);
	if (result.exitCode !== 0) {
		return Promise.reject(result);
	}
	return Promise.resolve(result);
};
```

Electroid:
This would for sure be a breaking change, that said, I’m curious what @zackradisic thinks.

zackradisic:
I think it makes sense to throw on non-zero exit code, and provide something like `.nothrow()` (this is what [zx does](https://google.github.io/zx/process-promise#nothrow)) if you don't want it to throw:

```typescript
import { $ } from 'bun'

// Won't throw on errors
await $`echo hsdfjsldkjf`.nothrow()

// Set the option globally for all shell invocations
$.throws(false)
```

pfgithub:
Right now this is especially a problem for `.text()`

```ts
import {$} from "bun";
await $`ls abc`.text();
```

This just returns `""` with no indication that the command failed and no way to get that information out.

To handle it correctly, `.text()` can't be used at all and instead you have to parse the ShellOutput stdout manually.

rhyek:
migrating to bun shell from execa and this is a pain point with my scripts that have many shell commands where i expect execution to stop and exit on any unexpected error.

rhyek:
this helper function i wrote works for me and maintains the same API:
```typescript
import { $, type Shell } from 'bun';

const $$: Shell = ((...params: Parameters<Shell>) => {
[...]

PR #8846, opened 2024-02-11
Shell changes/fixes

### LINKED ISSUE #8403: await $`ls *.js` returns all files in the current directory

1.0.25+a8ff7be64

Darwin 23.0.0 arm64 arm

`ls` in bun shell gives different results in a trival example vs zsh invocation

```
❯ ls *.js
zsh: no matches found: *.js

❯ cat bun-shell.ts
import { $ } from "bun";

// to stdout:
await $`ls *.js`;

❯ bun bun-shell.ts
index.tsx
bun-shell.ts
bun.lockb
node_modules
package.json
math.test.ts
```

`ls` command should filter to just *.js files

### What do you see instead?

`ls` command lists all files in directory

### Additional information

_No response_

zackradisic:
### What does this PR do?

- Fixes #8403
- Implements #8544 and adds a `ShellError`
- Adds the stdout/stderr utilities (`.text()`, `.blob()`, etc.) to `ShellOutput` and `ShellPromise`
- Uses the `.raw` property to reduce having to put backslashes for shell input string

- [x] Code changes

Zig files changed:

New methods, getters, or setters were added to a publicly exposed class:


src/shell/interpreter.zig
```diff
@@ -1580,17 +1580,36 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
...
 
-                switch (GlobWalker.init(&this.child_state.glob.walker, &arena, pattern, false, false, false, false, false) catch bun.outOfMemory()) {
+                const cwd = this.base.shell.cwd();
+
+                switch (GlobWalker.initWithCwd(
+                    &this.child_state.glob.walker,
+                    &arena,
+                    pattern,
+                    cwd,
+                    false,
```
Jarred-Sumner, line 1606: at some point we should refactor this to an options struct

packages/bun-types/bun.d.ts
```diff
@@ -156,6 +225,16 @@ declare module "bun" {
      * ```
      */
     blob(): Promise<Blob>;
+
+    /**
+     * Configure the shell to not throw an exception on non-zero exit codes.
+     */
+    nothrow(): this;
```
mathiasrw, line 234: Should this be `noThrow` for consistent camel case (like`exitCode`)?

test/js/bun/shell/throw.test.ts
```diff
@@ -0,0 +1,50 @@
+import { $ } from "bun";
+import { beforeAll, describe, test, expect } from "bun:test";
+
+beforeAll(() => {
+  $.nothrow();
```
mathiasrw, line 5: Is the idea that `$.nothrow()` and `$.throws(false)` does the same?

test/js/bun/shell/throw.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      expect("Woops").toBe("Should have thrown");
+    } catch (err) {
+      e = err;
+    }
+    expect(e).toBeDefined();
+  });
+
+  test("disable globally", async () => {
+    $.throws(true);
+    $.nothrow();
```
mathiasrw, line 34: So first we turn throwing on and then off?

packages/bun-types/bun.d.ts
zackradisic: `nothrow()` is from zx's API, but I guess it makes more sense to make it camel case

test/js/bun/shell/throw.test.ts
zackradisic: yep

test/js/bun/shell/throw.test.ts
zackradisic: Not necessary, I'm just doing it here to be extra careful with testing behaviour since `$.nothrow()` is in the `beforeAll`

test/js/bun/shell/throw.test.ts
mathiasrw: Having two ways of doing the same thing will create confusion down the line. 

My input: only be able to set the global behavour using `.throws(Boolean)` while using `.noThrow()` for individual calls. 

(this is of cause not a demand, but a request to reconsider)

Jarred-Sumner (changes requested):
Shell leak tests are failing on Linux

src/shell/interpreter.zig
jdalton: woof

packages/bun-types/bun.d.ts
jdalton: You could keep one as an alias to the other for less hurtles for folks reaching for familiar API

src/js/builtins/shell.ts
```diff
@@ -20,15 +72,24 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
   class ShellPromise extends Promise<ShellOutput> {
     #core: ShellInterpreter;
     #hasRun: boolean = false;
+    #throws: boolean = true;
     // #immediate;
-    constructor(core: ShellInterpreter) {
+    constructor(core: ShellInterpreter, throws: boolean) {
       var resolve, reject;
```
jdalton: `var resolve, reject` -> `let resolve, reject`.

src/js/builtins/shell.ts
```diff
@@ -93,6 +154,16 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
       return this.#quiet();
     }
 
+    nothrow(): this {
+      this.#throws = false;
+      return this;
+    }
+
+    throws(doThrow: boolean | undefined): this {
+      this.#throws = typeof doThrow === "boolean" ? doThrow : false;
```
jdalton:
```suggestion
      this.#throws = !!doThrow;
```
We could just coerce anything to a boolean.

src/js/builtins/shell.ts
```diff
@@ -178,19 +251,30 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
 
       return this;
     }
+    nothrow() {
+      this[throwsSymbol] = false;
+      return this;
+    }
+    throws(doThrow: boolean | undefined) {
+      this[throwsSymbol] = typeof doThrow === "boolean" ? doThrow : false;
+      return this;
```
jdalton, line 260: Same as [above](https://github.com/oven-sh/bun/pull/8846/files#r1488576003)

src/js/builtins/shell.ts
```diff
@@ -178,19 +251,30 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
...
+    }
   }
 
   var BunShell = function BunShell() {
```
jdalton:
```suggestion
  var BunShell = function BunShell(...args) {
```

src/js/builtins/shell.ts
```diff
@@ -199,16 +283,19 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
     }
 
     var Shell = function Shell() {
-      const core = new ShellInterpreter(...arguments);
+      const [first, ...rest] = arguments;
+      if (first.raw === undefined) throw new Error("Please use `$` as a tagged template function: $`cmd arg1 arg2`");
```
jdalton: Error message bikeshed. Since we are using the \` to represent the template literal string at the end of the message using it around `$` seems odd. We could just use $ without wrapping it in \` backticks. Also does using "Please" match our error voice (other messages).

src/js/builtins/shell.ts
```diff
@@ -199,16 +283,19 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
     }
 
     var Shell = function Shell() {
-      const core = new ShellInterpreter(...arguments);
+      const [first, ...rest] = arguments;
```
jdalton: ☝️ `arguments` object nit

src/js/builtins/shell.ts
```diff
@@ -223,6 +310,7 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
 
   BunShell[cwdSymbol] = defaultCwd;
   BunShell[envSymbol] = defaultEnv;
+  BunShell[throwsSymbol] = true;
 
   Object.defineProperties(BunShell, {
     Shell: {
```
jdalton, line 314: JS descriptors will default to `false` for unspecified descriptor properties like `configurable` and `writable` so if it's `configurable: false` you can exclude it and it will be treated as false. Since this is done at bootstrap (I'm guessing) we don't have to worry about prototype injection (or else we'd also want to `__proto__: null` the descriptors as JS reads them as inherited properties (not just own properties).

src/shell/interpreter.zig
```diff
@@ -1580,17 +1583,36 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
...
+                    false,
+                    false,
+                    false,
+                    false,
+                ) catch bun.outOfMemory()) {
                     .result => {},
                     .err => |e| {
-                        global_handle.get().actuallyThrow(bun.shell.ShellErr.newSys(e));
+                        this.state = .{ .err = bun.shell.ShellErr.newSys(e) };
```
jdalton, line 1614: Odd naming nit, we have `ShellErr` here but `ShellError` elsewhere. In other code I've seen things like `SystemError` and `Syscall.Error` (so long for `Error` instead of `Err`.)

src/shell/shell.zig
```diff
@@ -922,7 +943,8 @@ pub const Parser = struct {
                         self.continue_from_subparser(&subparser);
                         if (self.delimits(self.peek())) {
                             _ = self.match(.Delimit);
-                            if (should_break) break;
+                            // if (should_break) break;
```
jdalton: ☝️ Is this debug? (can it be removed instead of commented out)?

test/js/bun/shell/bunshell.test.ts
```diff
@@ -148,8 +153,9 @@ describe("bunshell", () => {
 
     test("escape unicode", async () => {
       const { stdout } = await $`echo \\弟\\気`;
-
-      expect(stdout.toString("utf8")).toEqual(`\弟\気\n`);
+      // expect(stdout.toString("utf8")).toEqual(`\弟\気\n`);
```
jdalton, line 157:
```suggestion
      // TODO: Uncomment and replace after unicode in template tags is supported
      // expect(stdout.toString("utf8")).toEqual(`\弟\気\n`);
```
Adding the TODO comment helps for folks cleaning house (or doing tech debt work)

test/js/bun/shell/throw.test.ts
jdalton: There's inspiration to look at in test runners I wanna say for similar kinds of global vs test level APIs.

packages/bun-types/bun.d.ts
mathiasrw: So `.throws()` would be treated the same as `.noThrow()` (by acting like `.throws(false)`)?

jdalton: > So `.throws()` would be treated the same as `.noThrow()` (by acting like `.throws(false)`)?

I was talking about `nothrow` vs `noThrow`.

paperclover: im not sure how i feel about having this and `throws()`, and then desire to add a third function.

i think we can just say in the doc comment this is equivilent to `throws(false)` and mention the zx api.

src/js/builtins/shell.ts
paperclover: interesting. can verify this behavior is matched in builtin functions:
```ts
#!/usr/bin/env BUN_JSC_useDollarVM=1 bun
{
  const o = {};
  o.__proto__.configurable = true;
};
const c = {};
const def = $vm.createBuiltin(`(function (obj) {
  Object.defineProperty(obj, 'bruh', {
    value: 'bruh',
    writable: true,
  });
  return 1;
})`);
def(c);
console.log(Object.getOwnPropertyDescriptor(c, 'bruh'));
```

I think using __proto__ anywhere it isn't truly needed is ugly, and simply specifying all properties for this is a better idea.

src/shell/shell.zig
```diff
@@ -1778,6 +1800,9 @@ pub fn NewLexer(comptime encoding: StringEncoding) type {
                         switch (char) {
                             '0'...'9' => {
                                 _ = self.eat();
+                                if (count >= 32) {
+                                    return null;
+                                }
                                 buf[count] = @intCast(char);
```
paperclover, line 1805: this wasnt changed in this pr but why is this `@intCast`. i dont actually know the types here but why can this cast fail?

src/shell/shell.zig
```diff
@@ -1778,6 +1800,9 @@ pub fn NewLexer(comptime encoding: StringEncoding) type {
                         switch (char) {
                             '0'...'9' => {
                                 _ = self.eat();
+                                if (count >= 32) {
```
paperclover, line 1802: what is 32. im mostly curious

test/js/bun/shell/parse.test.ts
```diff
@@ -388,6 +388,74 @@ describe("parse shell", () => {
...
+                    "redirect_file": null,
+                  },
+                },
+              },
+            },
+          ],
+        },
+      ],
+    };
+    const result = JSON.parse($.parse`echo $(ls foo) && echo nice`);
```
paperclover, line 455: ~~why is this a public api. is this documented (i hope it is not)~~ it is not.

we should make this hidden with #8166's changes

paperclover (changes requested):
mostly great. john-david has some good comments you should address. i think mainly the `arguments` one. also for that we do have `$argument(n)` (needs an `n` known at comptime, so i dont think it applies). and i mostly just have some questions i wnat answered.

src/js/builtins/shell.ts
jdalton: Defining the descriptor properties _((value + writable for data properties, get + set for accessors) + configurable + enumerable)_ everywhere is an option in those cases (though a little verbose) where it's relevant. It also means if we miss a descriptor property we have a possible issue. The nice thing about a `__proto__: null,` approach is it's forgiving. This kind of vulnerability (prototype injection) effects other APIs too (DOM and JS)
* https://github.com/nodejs/node/pull/43270/files
* https://github.com/nodejs/node/pulls?q=is%3Apr+is%3Aclosed+prototype+pollution
* https://github.com/search?q=repo%3Anodejs%2Fnode+__proto__+null&type=pullrequests

 and is part of a larger issue (for example spreading values into array `[...a]` can be modified by changes to `Array.prototype[Symbol.iterator]` etc). 
 
I've seen folks wrestle with the pretty/ugly bit of the syntax _(I've been there too)_ but at some point security and maintainability > opinions about dangling underscores.

src/shell/shell.zig
zackradisic: The size of the digit buf

src/js/builtins/shell.ts
```diff
@@ -310,20 +310,20 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
...
+  BunShell[throwsSymbol] = false;
 
   Object.defineProperties(BunShell, {
     Shell: {
       value: Shell,
-      configurable: false,
+      // configurable: false,
       enumerable: true,
-      writable: false,
+      // writable: false,
```
Jarred-Sumner: deleting them is fine

src/js/builtins/shell.ts
```diff
@@ -256,13 +256,13 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
...
     throws(doThrow: boolean | undefined) {
-      this[throwsSymbol] = typeof doThrow === "boolean" ? doThrow : false;
+      this[throwsSymbol] = !!doThrow;
       return this;
     }
   }
 
-  var BunShell = function BunShell() {
-    const [first, ...rest] = arguments;
+  var BunShell = function BunShell(...args) {
```
Jarred-Sumner: This creates two arrays, we don't need to do that

src/js/builtins/shell.ts
```diff
@@ -282,9 +282,9 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
       throw new TypeError("Class constructor Shell cannot be invoked without 'new'");
     }
 
-    var Shell = function Shell() {
-      const [first, ...rest] = arguments;
-      if (first.raw === undefined) throw new Error("Please use `$` as a tagged template function: $`cmd arg1 arg2`");
+    var Shell = function Shell(...args) {
```
Jarred-Sumner: ...args causes it to clone the arguments array

and then later, ...rest clones it again

we should use `arguments` more directly or alternatively we could do this destructuring in the function definition like (first, ...rest)

src/js/builtins/shell.ts
```diff
@@ -282,9 +282,9 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
...
+      const [first, ...rest] = args;
+      if (first.raw === undefined) throw new Error("Please use '$' as a tagged template function: $`cmd arg1 arg2`");
```
Jarred-Sumner: another option would be to check if `arguments[0]?.raw` is undefined

src/js/builtins/shell.ts
```diff
@@ -261,9 +261,8 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
     }
   }
 
-  var BunShell = function BunShell(...args) {
-    const [first, ...rest] = args;
-    if (first.raw === undefined) throw new Error("Please use `$` as a tagged template function: $`cmd arg1 arg2`");
+  var BunShell = function BunShell(first, ...rest) {
+    if (first?.raw === undefined) throw new Error("Please use '$' as a tagged template function: $`cmd arg1 arg2`");
     const core = new ShellInterpreter(first.raw, ...rest);
```
jdalton, line 266: Noice!

src/js/builtins/shell.ts
jdalton: `arguments` directly is a likely slower route. The `arguments` object has a lot of baggage with it and potential slowness if touched wrong. WebKit has magic `@` helpers for it though. However, in this case the `first, ...rest` is very nice!

Jarred-Sumner: arguments is a bytecode intrinsic, it won’t materialize an extra object unless necessary  

… means it must always clone 

really the faster way here would be to pass through arguments without the … and then make it an ArgList and reference that. 

You can read JSC code to learn more about this if you search for ClonedArguments

jdalton: > it won’t materialize an extra object unless necessary

That maps to my comment "and potential slowness if touched wrong". Back with my time as the Chakra JS engine performance technical PM at Microsoft we had several gaps in our `arguments` object handling that would slow path us (compared to the competition) which we eventually fixed. Since then I have a weariness of `arguments` (even knowing other engines optimize it in more cases).

I'm speaking generally as maps to my "is a likely slower route" comment. And generally having devs reach for `arguments` should be avoided. Yes, if handled with care it can be okay. But JSC itself "has magic `@` helpers":
https://github.com/WebKit/WebKit/commit/61208896b94b2745450e56a744669977ff1dc0b1#diff-2f6c0607ce64738a74242e2c1419a66b1f9e622f40f36be549f41b4be3cf0fd4

so that they can, in their own JS implemented methods, avoid `arguments`.

As for the `rest` param. While knowing the arguments up front and passing them without needing an array is best, likely followed by magic helper use, of the syntax sugars `rest`, unlike spread, has a secure JS benefit as `rest` will not defer to `Array.prototype[Symbol.iterator]` and so is safe for internal use. The `rest` syntax also has DFG optimizations around it: 
* https://github.com/WebKit/WebKit/commit/babbd952c15c1948a7b1e6a4d63d9d7eb2bb2ae8
* https://github.com/WebKit/WebKit/commit/1908569d19e07989d8d21671723e0c9a5dd1449c
* [WebKit/WebKit@`329af67`](https://github.com/WebKit/WebKit/commit/329af672df18d52add0089e20121c4ac09add9ff#diff-91a5b64badfd992fd984615a03f88c669ac07d10d7671f5052a2fed25335052dR35) _(`rest` participates in the DFGArgumentsEliminationPhase too)_

PR #8858, opened 2024-02-12
fix: distinguish getters and setters in Bun.inspect()

### LINKED ISSUE #8853: JS Setter is written as `Getter` in console.log

1.0.26+c75e768a6

Linux 6.1.66-1-lts x86_64 unknown

```ts
const obj = {
  a: 1,
  get getA() {
    return this.a;
  },
  set setA(n) {
    this.a = n;
  },
};

obj.setA = 2;
console.log(obj.getA);
console.log(obj);
```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

argosphil:
### What does this PR do?

Fixes #8853.

This changes Bun.inspect to print "[Getter]", "[Setter]", or "[Getter/Setter]" for property accessors rather than always defaulting to "[Getter]".

To do so, we need to access the GetterSetter and CustomGetterSetter structures from zig, which requires expanding the bindings somewhat.

I've chosen to duplicate the code paths for GetterSetter and CustomGetterSetter to make it easier for future PRs to change them so we actually obtain references and print the getters and setters, which we can only do in the GetterSetter case.

Finally, two new tests ensure things work, though I'm not sure how to test the CustomGetterSetter case.

- [x] Code changes

I wrote automated tests.


src/bun.js/ConsoleObject.zig
```diff
@@ -921,6 +921,7 @@ pub const Formatter = struct {
         Event,
 
         Getter,
+        CustomGetter,
```
gvilums:
```suggestion
        GetterSetter,
        CustomGetterSetter,
```
If we're making changes it probably makes sense to change these names to be more descriptive

gvilums (commented):
Thanks for the contribution, this looks good overall. One small nit:

src/bun.js/ConsoleObject.zig
argosphil: Agreed, thanks!

paperclover:
a merge conflict but yeah this looks good

argosphil:
> a merge conflict but yeah this looks good

Thanks for letting me know, I hadn't noticed.

gvilums (approved):

PR #8859, opened 2024-02-12
fix: Bun.sleep() should round its argument towards Infinity

### LINKED ISSUE #8834: Bun.sleep() rounds incorrectly

argosphil:
### What does this PR do?

Fixes #8834 .

This changes Bun.sleep() to round its argument towards Infinity. It ensures Bun.sleep(1.5) does not return until at least 2 ms have passed, and Bun.sleep(Date) does not return until Date has been reached.

This assumes C's `double` is the same type as a JavaScript number, but that's true on all modern machines. It's technically possible for this code to require the math library.

- [x] Code changes

I wrote automated tests.

### DIFF
--- a/src/bun.js/bindings/BunObject.cpp
+++ b/src/bun.js/bindings/BunObject.cpp
@@ -362,9 +362,11 @@ JSC_DEFINE_HOST_FUNCTION(functionBunSleep,
         return JSC::JSValue::encode(JSC::JSValue {});
     }
 
+    int32_t countdown = std::ceil(millisecondsValue.asNumber());
+
     Zig::GlobalObject* global = JSC::jsCast<Zig::GlobalObject*>(globalObject);
     JSC::JSPromise* promise = JSC::JSPromise::create(vm, globalObject->promiseStructure());
-    Bun__Timer__setTimeout(globalObject, JSC::JSValue::encode(global->bunSleepThenCallback()), JSC::JSValue::encode(millisecondsValue), JSValue::encode(promise));
+    Bun__Timer__setTimeout(globalObject, JSC::JSValue::encode(global->bunSleepThenCallback()), countdown, JSValue::encode(promise));
     return JSC::JSValue::encode(promise);
 }
 
--- a/src/bun.js/bindings/ZigGlobalObject.cpp
+++ b/src/bun.js/bindings/ZigGlobalObject.cpp
@@ -1338,6 +1338,14 @@ JSC_DEFINE_HOST_FUNCTION(functionSetTimeout,
         return JSC::JSValue::encode(JSC::JSValue {});
     }
 
+    if (UNLIKELY(!num.isNumber())) {
+        auto scope = DECLARE_THROW_SCOPE(globalObject->vm());
+        JSC::throwTypeError(globalObject, scope, "setTimeout expects a number (milliseconds)"_s);
+        return JSC::JSValue::encode(JSC::JSValue {});
+    }
+
+    int32_t countdown = std::ceil(num.asNumber());
+
 #ifdef BUN_DEBUG
     /** View the file name of the JS file that called this function
      * from a debugger */
@@ -1349,7 +1357,7 @@ JSC_DEFINE_HOST_FUNCTION(functionSetTimeout,
     }
 #endif
 
-    return Bun__Timer__setTimeout(globalObject, JSC::JSValue::encode(job), JSC::JSValue::encode(num), JSValue::encode(arguments));
+    return Bun__Timer__setTimeout(globalObject, JSC::JSValue::encode(job), countdown, JSValue::encode(arguments));
 }
 
 JSC_DEFINE_HOST_FUNCTION(functionSetInterval,
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -4405,7 +4405,7 @@ pub const Timer = struct {
     pub fn setTimeout(
         globalThis: *JSGlobalObject,
         callback: JSValue,
-        countdown: JSValue,
+        countdown: i32,
         arguments: JSValue,
     ) callconv(.C) JSValue {
         JSC.markBinding(@src());
@@ -4413,7 +4413,7 @@ pub const Timer = struct {
         globalThis.bunVM().timer.last_id +%= 1;
 
         const interval: i32 = @max(
-            countdown.coerce(i32, globalThis),
+            countdown,
             // It must be 1 at minimum or setTimeout(cb, 0) will seemingly hang
             1,
         );
--- a/src/bun.js/bindings/headers.h
+++ b/src/bun.js/bindings/headers.h
@@ -804,7 +804,7 @@ ZIG_DECL JSC__JSValue Bun__Timer__clearInterval(JSC__JSGlobalObject* arg0, JSC__
 ZIG_DECL JSC__JSValue Bun__Timer__clearTimeout(JSC__JSGlobalObject* arg0, JSC__JSValue JSValue1);
 ZIG_DECL int32_t Bun__Timer__getNextID();
 ZIG_DECL JSC__JSValue Bun__Timer__setInterval(JSC__JSGlobalObject* arg0, JSC__JSValue JSValue1, JSC__JSValue JSValue2, JSC__JSValue JSValue3);
-ZIG_DECL JSC__JSValue Bun__Timer__setTimeout(JSC__JSGlobalObject* arg0, JSC__JSValue JSValue1, JSC__JSValue JSValue2, JSC__JSValue JSValue3);
+ZIG_DECL JSC__JSValue Bun__Timer__setTimeout(JSC__JSGlobalObject* arg0, JSC__JSValue JSValue1, int32_t value2, JSC__JSValue JSValue3);
 
 #endif
 
--- a/test/js/web/timers/setTimeout.test.js
+++ b/test/js/web/timers/setTimeout.test.js
@@ -148,11 +148,25 @@ it("Bun.sleep propagates exceptions", async () => {
 });
 
 it("Bun.sleep works with a Date object", async () => {
+  const now = performance.now();
   var ten_ms = new Date();
   ten_ms.setMilliseconds(ten_ms.getMilliseconds() + 12);
-  const now = performance.now();
   await Bun.sleep(ten_ms);
   expect(performance.now() - now).toBeGreaterThan(11);
+});
+
+it("Bun.sleep rounds towards Infinity", async () => {
+  const now = performance.now();
+  await Bun.sleep(1.5);
+  expect(performance.now() - now).toBeGreaterThan(1.5);
+});
+
+it("Bun.sleep(Date) fulfills after Date", async () => {
+  let ten_ms = new Date();
+  ten_ms.setMilliseconds(ten_ms.getMilliseconds() + 12);
+  await Bun.sleep(ten_ms);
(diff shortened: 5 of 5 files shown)


src/bun.js/bindings/BunObject.cpp
```diff
@@ -362,6 +362,8 @@ JSC_DEFINE_HOST_FUNCTION(functionBunSleep,
         return JSC::JSValue::encode(JSC::JSValue {});
     }
 
+    millisecondsValue = JSC::jsNumber(std::ceil(millisecondsValue.asNumber()));
```
paperclover: this shouldn't have to re-encode a JSValue. though `jsNumber` isn't expensive it is needless as the value will be immediatly converted back to an i32.

let's change the `countdown` argument in BunObject.zig's `Timer.setTimeout` to simply take in an `i32`, and then we can use this logic for rounding here.

paperclover (changes requested):

paperclover:
> This assumes C's double is the same type as a JavaScript number

This assumption is pretty much correct. JSValue is either an i32 or a f64, with different ways to tag these values. the double tag uses a different way to tag NaN (see PureNaN.h), but these conversions are always handled for you.

src/bun.js/bindings/BunObject.cpp
argosphil: should we change `setInterval` and `sleepSync`, too?

paperclover: that makes sense to do as well, not sure if it is strictly necessary to change. do you think this rounding behavior should apply for `setInterval` and `sleepSync`?

argosphil: > that makes sense to do as well, not sure if it is strictly necessary to change. do you think this rounding behavior should apply for `setInterval` and `sleepSync`?

I need to check the specs as to whether correct behavior is allowed here, but I think they should round to the nanosecond (and toward Infinity, but that doesn't matter on real hardware), as should `setTimeout`.

Bun.sleep(Date) is a special case because we're actually hitting the limits of f64 arithmetic, though, so let's fix that one first?

src/bun.js/bindings/BunObject.cpp
```diff
@@ -362,9 +363,11 @@ JSC_DEFINE_HOST_FUNCTION(functionBunSleep,
         return JSC::JSValue::encode(JSC::JSValue {});
     }
 
+    countdown = std::ceil(millisecondsValue.asNumber());
```
paperclover: cant it just be `int countdown =` instead of the declaration and the assignment happening on separate lines?

paperclover (changes requested):
a nitpick

src/bun.js/bindings/BunObject.cpp
argosphil: Absolutely. This isn't Linux, after all :-)

src/bun.js/bindings/BunObject.cpp
```diff
@@ -362,9 +362,11 @@ JSC_DEFINE_HOST_FUNCTION(functionBunSleep,
         return JSC::JSValue::encode(JSC::JSValue {});
     }
 
+    int32_t countdown = std::ceil(millisecondsValue.asNumber());
+
     Zig::GlobalObject* global = JSC::jsCast<Zig::GlobalObject*>(globalObject);
     JSC::JSPromise* promise = JSC::JSPromise::create(vm, globalObject->promiseStructure());
-    Bun__Timer__setTimeout(globalObject, JSC::JSValue::encode(global->bunSleepThenCallback()), JSC::JSValue::encode(millisecondsValue), JSValue::encode(promise));
+    Bun__Timer__setTimeout(globalObject, JSC::JSValue::encode(global->bunSleepThenCallback()), countdown, JSValue::encode(promise));
```
jdalton, line 369: Since we're documenting that sleep is just a promise wrapped setTimeout. We can look to JSC for inspiration if needed: https://github.com/WebKit/WebKit/blob/eeccbd155ef85f55638ba1bb4eacfbfb8d1c22c5/Source/WebCore/page/LocalDOMWindow.cpp#L1832C85-L1832C97

`int`, [`Seconds::fromMilliseconds`](https://github.com/WebKit/WebKit/blob/eeccbd155ef85f55638ba1bb4eacfbfb8d1c22c5/Source/WTF/wtf/Seconds.h#L76). [int DOMTimer::install](https://github.com/WebKit/WebKit/blob/3f6d7f5ff24f92cb8327e9d67a87755d6d3d47a4/Source/WebCore/page/DOMTimer.cpp#L197), etc.

Ideally we keep the contract that it's just `setTimeout` wrapped in a promise so look to however our `setTimeout` is handled:
https://github.com/oven-sh/bun/blob/7405c5a615ff5c7027186652dd1ec7afe8432fb1/src/bun.js/bindings/ZigGlobalObject.cpp#L1352

argosphil: Oh, I'm really sorry I missed that call to Bun__Timer__setTimeout. I'm a bit confused that didn't cause a compile-time error, but I'll fix it and do another search...

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1338,6 +1338,14 @@ JSC_DEFINE_HOST_FUNCTION(functionSetTimeout,
         return JSC::JSValue::encode(JSC::JSValue {});
     }
 
+    if (UNLIKELY(!num.isNumber())) {
```
paperclover, line 1341: you can call setTimeout without a number as the param.

`setTimeout(() => console.log('called'), '1000')`

i forget what `undefined` as the timer value actually ends up doing.

the `JSC__JSValue__coerceToInt32` function is what was used previously by .coerce(i32), so to be safe i think we can just use that.

argosphil: Thanks! I think we should coerce to double, apply ceil(), then cast to int32_t?

jdalton: Instead of getting too funky with ceiling and casting. I think we should narrow down more closely to  what WebKit/Browsers do with setTimeout and simply do that. Via https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value values are coerced to Int32 and even roll-over when too high. I assume using the `JSC__JSValue__coerceToInt32` will do that since it uses [`JSValue::toInt32`](https://github.com/WebKit/WebKit/blob/b9f025ad07776f25423540680b468156460a3039/Source/JavaScriptCore/runtime/JSCJSValueInlines.h#L45-L52) under the hood.

jdalton: I think beyond conforming to the expected known behavior of `setTimeout` any other ceiling or rounding should be skipped unless we change the expectation of `bun.sleep`.

jdalton (changes requested):
At this point I think the PR is more about making sure Bun's implementation of setTimeout is accurate. If we have tweaks needed then this can be morphed into that, Else I don't see much reason to change things.

src/bun.js/bindings/ZigGlobalObject.cpp
argosphil: I'm not sure I fully understand. Do you want to behave in the manner that browsers do, or do you want to skip rounding because everything is done with high precision?

Also, do you want to remove Bun.sleep(Date) or fix it?

Here's a branch that does the minimum fix to get the tests to pass, adds one, and leaves it at that:

https://github.com/oven-sh/bun/compare/main...argosphil:bun:issue-8834-003?expand=1

jdalton: @argosphil Since `Bun.sleep()` is `setTimeout` wrapped in a promise we are dealing with the several factors traditionally. I know that things are different in Bun than a browser because we tie-into event loops and not necessariy the event loop of a browser but just to get the point across traditionally we have to deal with delays being less precise because

* [sometimes delays are throttled](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#reasons_for_delays_longer_than_specified) to 4ms or so
* inactive tabs (min timeout is 1 second or 15minutes on mobile)
* timeouts are placed in a [task queue](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide#tasks) different than the promise which is in a [microtask queue](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide#tasks_vs._microtasks)

This is to say delays of the queued timeout callback can and do happen (esp relying on both the task and microtask queue) so delays are not precise. See details on [browsers event loop](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide/In_depth) and [Node's event loop](https://nodejs.org/en/guides/event-loop-timers-and-nexttick/) (also on [Promises in Node's event loop](https://nodejs.org/en/learn/asynchronous-work/understanding-setimmediate)) which Bun simulates

> Do you want to behave in the manner that browsers do, or do you want to skip rounding because everything is done with high precision?

Do nothing if we behave as passing off to `setTimeout` wrapped in a promise. If we differ from that and our `setTimeout` implementation needs tweaking then that is where we can tweak.

argosphil: @jdalton Thanks for your response!

> @argosphil Since `Bun.sleep()` is `setTimeout` wrapped in a promise

I agree. I proposed to fix the one exception (setTimeout doesn't take a Date object, Bun.sleep does) first, then look into switching all timeouts to high precision.

There is one point where we do and should continue to deviate from the HTML Standard, which is that it mandates a 4 ms delay after five nested calls. Do you think we should implement that delay?

Note that high precision does not violate the HTML Standard. I do not know to what extent it behaves differently from what node does, though.

> Do nothing if we behave as passing off to `setTimeout` wrapped in a promise. If we differ from that and our `setTimeout` implementation needs tweaking then that is where we can tweak.

Yes, you've convinced me of that point, which is why I think I should open a PR for the issue-8834-003 branch, which doesn't touch setTimeout or Bun.sleep(number) in any way.

jdalton: @argosphil 
> There is one point where we do and should continue to deviate from the HTML Standard, which is that it mandates a 4 ms delay after five nested calls. Do you think we should implement that delay?

I'd look to if Node does anything there. Our compat burden is for the Node behavior.

> Yes, you've convinced me of that point, which is why I think I should open a PR for the issue-8834-003 branch, which doesn't touch setTimeout or Bun.sleep(number) in any way.

Awesome! Thank you @argosphil  for digging into this!

jdalton:
Closed by https://github.com/oven-sh/bun/pull/8950.

PR #8861, opened 2024-02-12
fix: use shell-independent globbing in bun run fmt:zig

argosphil:
### What does this PR do?

This fixes the globbing pattern used for `fmt:zig` to work with standard Unix `/bin/sh` implementations. The previous pattern used `**`, which is interpreted as equivalent to `*` by `/bin/sh`.

This also applies zig fmt to a file which was previously generated with the wrong indentation settings.

- [ ] Code changes

I verified this properly formats my broken code from #8858 .


src/bun.js/bindings/GeneratedJS2Native.zig
```diff
@@ -1,24 +1,24 @@
 const JSC = @import("root").bun.JSC;
```
paperclover: hold on im sorry when did this file get committed to main. this is a generated file
and

a generated file from this to-be-merged-but-not-yet pr #8166

i am sorry. this file should be deleted.

argosphil: File? What file? There was never a file here :mage_man:

paperclover:
thanks. one day bun shell will make these kinds of issues go away.

PR #8866, opened 2024-02-12
enable debug logs when BUN_DEBUG_QUIET_LOGS is set to 0

gvilums:
### What does this PR do?
Makes the logs non-quiet if BUN_DEBUG_QUIET_LOGS is set to 0

manual testing.


Jarred-Sumner:
Thanks

PR #8870, opened 2024-02-12
Run zig fmt

zackradisic:
### What does this PR do?

Runs `bun run fmt:zig`


Jarred-Sumner:
I’m surprised we don’t fail CI for not doing this

PR #8871, opened 2024-02-13
scripts/codegen.ps1: update wsl call

nektro:
status quo means u can run into conditions where you get:

```
PS C:\bun> .\scripts\codegen.ps1
The file cannot be accessed by the system.
```

this change allows it to always work and run as expected.


Jarred-Sumner:
_technically_ we should use %PROGRAMFILES% but it doesn't matter

staabm:
in case this scripts is running on the enduser machine, I can confirm that on my german windows version `'%PROGRAMFILES%\WSL\wsl.exe'` works and `'C:\Program Files\WSL\wsl.exe'` doesn't work.

paperclover:
ah wow i didn't realize these files get localized. it probably does matter.

since this is a powershell script it is $env:PROGRAMFILES but potato potato

PR #8875, opened 2024-02-13
feat: make curl install script run the windows installer (if run through cygwin)

AugusDogus:
This updates the *nix install script to run the Windows Powershell install script if it detects it's on Windows.

### DIFF
--- a/src/cli/install.sh
+++ b/src/cli/install.sh
@@ -3,7 +3,7 @@ set -euo pipefail
 
 if [[ ${OS:-} = Windows_NT ]]; then
     powershell -c "irm bun.sh/install.ps1|iex"
-    exit 1
+    exit $?
 fi
 
 # Reset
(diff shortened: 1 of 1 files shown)


src/cli/install.sh
```diff
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 if [[ ${OS:-} = Windows_NT ]]; then
-    echo 'error: Please install bun using Windows Subsystem for Linux'
+    powershell -c "irm bun.sh/install.ps1|iex"
     exit 1
```
paperclover: this exit 1 is not correct for the success case. making it 0 makes it wrong for the failure case. it should mirror what powershell exits with.

paperclover (changes requested):

src/cli/install.sh
AugusDogus: Updated to use `$?`.

paperclover (approved):

PR #8877, opened 2024-02-13
Allow user to uninstall windows version of bun

RiskyMH:
### What does this PR do?

This updates the registry so that the user can go to control panel and uninstall bun. Currently it only should delete all contents inside `.bun` (I tried to delete the root bun folder but powershell wasn't being nice).

Also fixes the install script to have the latest changes.

Registry options here: <https://learn.microsoft.com/en-us/windows/win32/msi/uninstall-registry-key>

```sh
# You can test my changes by running this script:
powershell -c "irm https://raw.githubusercontent.com/The-Fiery-Fire/bun/uninstall-windows/src/cli/install.ps1|iex"
```

- [x] Code changes

I ran my changed install script and tried to uninstall.


src/cli/install.ps1
```diff
@@ -180,4 +176,27 @@ function Install-Bun {
   if(!$hasExistingOther) {
     Write-Output "To get started, restart your terminal/editor, then type `"bun`"`n"
   }
-}
\ No newline at end of file
+
+  $RegistryKey = "HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\Bun"  
+  New-Item -Path $RegistryKey -Force | Out-Null
+  New-ItemProperty -Path $RegistryKey -Name "DisplayName" -Value "Bun" -PropertyType String -Force | Out-Null
+  # New-ItemProperty -Path $RegistryKey -Name "DisplayVersion" -Value $DisplayVersion -PropertyType String -Force | Out-Null
```
RiskyMH, line 196: I commented this out because it doesn't get updated in `bun upgrade` and I didn't want it to be incorrect

paperclover:
i'll have to mess around with this and see if this is something we can reliably do. if not then we can still keep the uninstall script (i specifically want the uninstall flow to always fix the user path env)

PR #8879, opened 2024-02-13
fix(windows): `console-iterator.test.ts` fix

dylan-conway:
### What does this PR do?
calculates estimated byte size of a `Blob` instance when `toJS` is called. This avoids accessing `Blob.Store` after it's gc'd and invalid memory. Also adds the same pattern to `Response` and `Request`.

manually

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -1426,7 +1426,7 @@ pub const Blob = struct {
         return blob_;
     }
 
-    fn calculateEstimatedByteSize(this: *Blob) usize {
+    fn calculateEstimatedByteSize(this: *Blob) void {
         // in-memory size. not the size on disk.
         var size: usize = @sizeOf(Blob);
 
@@ -1444,7 +1444,7 @@ pub const Blob = struct {
             }
         }
 
-        return size + (this.content_type.len * @intFromBool(this.content_type_allocated));
+        this.reported_estimated_size = size + (this.content_type.len * @intFromBool(this.content_type_allocated));
     }
 
     pub fn estimatedSize(this: *Blob) callconv(.C) usize {
@@ -3409,6 +3409,8 @@ pub const Blob = struct {
             },
         }
 
+        blob.calculateEstimatedByteSize();
+
         var blob_ = bun.new(Blob, blob);
         blob_.allocator = allocator;
         return blob_;
@@ -3584,7 +3586,7 @@ pub const Blob = struct {
             std.debug.assert(this.allocator != null);
         }
 
-        this.reported_estimated_size = this.calculateEstimatedByteSize();
+        this.calculateEstimatedByteSize();
         return Blob.toJSUnchecked(globalObject, this);
     }
 
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -86,6 +86,8 @@ pub const Blob = struct {
         closing,
     };
 
+    reported_estimated_size: usize = 0,
+
     size: SizeType = 0,
     offset: SizeType = 0,
     /// When set, the blob will be freed on finalization callbacks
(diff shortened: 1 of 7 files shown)


src/bun.js/webcore/request.zig
```diff
@@ -128,10 +128,16 @@ pub const Request = struct {
     }
 
     pub fn estimatedSize(this: *Request) callconv(.C) usize {
-        return this.reported_estimated_size orelse brk: {
-            this.reported_estimated_size = @as(u63, @truncate(this.body.value.estimatedSize() + this.sizeOfURL() + @sizeOf(Request)));
-            break :brk this.reported_estimated_size.?;
-        };
+        return this.reported_estimated_size;
```
Jarred-Sumner, line 131: This should have an assertion that it is not zero

src/bun.js/webcore/response.zig
```diff
@@ -85,13 +85,19 @@ pub const Response = struct {
...
 
     pub fn estimatedSize(this: *Response) callconv(.C) usize {
-        return this.reported_estimated_size orelse brk: {
-            this.reported_estimated_size = @as(
-                u63,
-                @intCast(this.body.value.estimatedSize() + this.url.byteSlice().len + this.init.status_text.byteSlice().len + @sizeOf(Response)),
-            );
-            break :brk this.reported_estimated_size.?;
-        };
+        return this.reported_estimated_size;
```
Jarred-Sumner, line 88: This should have an assertion that it is not zero

src/bun.js/webcore/request.zig
```diff
@@ -128,10 +128,16 @@ pub const Request = struct {
...
+    }
+
+    pub fn calculateEstimatedSize(this: *Request) usize {
```
Jarred-Sumner: The name of this function should be consistent. `calculcateEstimatedSize` or `calculateEstimatedByteSize`?

src/bun.js/webcore/blob.zig
```diff
@@ -1424,22 +1426,20 @@ pub const Blob = struct {
...
-
-        return 0;
-    }
-
-    pub fn estimatedSize(this: *Blob) callconv(.C) usize {
-        var size = this.estimatedByteSize() + @sizeOf(Blob);
+        var size: usize = @sizeOf(Blob);
+
+        size += if (this.size != Blob.max_size)
+            @intCast(this.size)
```
Jarred-Sumner: We should not report the size of `.file` blobs as the real memory size

This value should only be respected when `this.store` is  .bytes

src/bun.js/webcore/request.zig
Jarred-Sumner: actually we should add that assertion in the code generator

Jarred-Sumner (approved):

src/bun.js/webcore/request.zig
```diff
@@ -128,10 +128,16 @@ pub const Request = struct {
...
-        return this.reported_estimated_size orelse brk: {
-            this.reported_estimated_size = @as(u63, @truncate(this.body.value.estimatedSize() + this.sizeOfURL() + @sizeOf(Request)));
-            break :brk this.reported_estimated_size.?;
-        };
+        return this.reported_estimated_size;
+    }
+
+    pub fn calculateEstimatedByteSize(this: *Request) void {
+        this.reported_estimated_size = this.body.value.estimatedSize() + this.sizeOfURL() + @sizeOf(Request);
+    }
```
jdalton, line 136: (learning): In this context is it always an estimated size and never a known size?

src/codegen/generate-classes.ts
```diff
@@ -1435,6 +1449,7 @@ function generateZig(
     construct,
     finalize,
     noConstructor = false,
+    overridesToJS = false,
     estimatedSize,
```
jdalton, line 1453: ☝️ @dylan-conway What's this bit about? (I'm curious from a how we are generating classes and whether we can leverage extending from base classes or not)

paperclover: check where it is used, you'll find it determines if we generate `pub fn toJS`

this function is the convention for converting something to a JSValue. some functions check if the passed type has this method declared and calls it automatically in some places.

in general the classes generator is done with 
1. a c++ class that hooks into all the webkit apis, and fwiw *is* the class.
2. a zig struct that is constructed along with the c++ class `<1>` contains a pointer to `<2>`
3. a second zig struct which is `usingnamespace`d into `<2>` which provides a default set of methods that let you call back into webkit

dylan-conway: Yes what Dave said. In this situation it was useful to be able to override toJS to add debug assertions and to ensure reported_estimated_size is set

jdalton: Is there a way to check this in another place for the existence of method to avoid added metadata and allow overriding other things if specified like overriding `fromJS` or `getConstructor`.

dylan-conway: To do that we could parse the zig struct implementation for the existence of the methods (`toJS`, `fromJS`, ...), and if any are defined, skip it in the generated code. It's possible but probably too much work for now

PR #8885, opened 2024-02-13
`events.on`: Add compatibility with `readline` module

### LINKED ISSUE #8370: `readline` does not read all line when looping with iterator

1.0.25+a8ff7be64

Linux 6.6.6-76060606-generic x86_64 x86_64

I have a file
example.txt
```txt 
1
2
3
4
5
```

I made 2 versions of the same scripts, one using events, and the other one a forof loop to iterate over the line.

events
```ts
import readline from "node:readline/promises";
import { createReadStream } from "node:fs";

const stream = createReadStream(`./example.txt`, {
  encoding: "utf-8",
});

const rl = readline.createInterface(stream);

rl.on("line", (line) => {
  console.log("line", line);
}) 
```
the output is correct with
```txt
line 1
line 2
line 3
line 4
line 5
```

but with an iterator
```ts
import readline from "node:readline/promises";
import { createReadStream } from "node:fs";

const stream = createReadStream(`./example.txt`, {
  encoding: "utf-8",
});

const rl = readline.createInterface(stream);

for await (const line of rl) {
  console.log("line", line);
}
```

It only output the first line and ends

Both scripts should loop over all the lines, like in the scripts using events.

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #8850: `readline` module yields wrong and incomplete values with a bad performance (compared to NodeJS)

1.0.26+c75e768a6

Linux 6.7.1 x86_64 unknown

Generate `big_file.txt` (1 million lines should be around 25MB, increase if needed):
```
$ cat /dev/urandom | tr -dc '[:lower:]\n' | head -n 1000000 > big_file.txt
```

Create the `test.mjs` script:
```
import fs from 'fs'
import readline from 'readline'

const fileStream = fs.createReadStream('./big_file.txt')
const rl = readline.createInterface({
  input: fileStream,
  crlfDelay: Infinity
})

console.log('Starting...')
for await (const line of rl) {
  console.log(typeof line, line)
}
```

Run script with NodeJS (starts yielding & printing lines immediately):
```
$ node test.mjs
Starting...
string ajqrsoyrabfleslmw
string ggwazolafkjersvirfrmynnucihuywtgjli
string eubcujmhoboaecf
string hokkaxedir
string hngvulerbketqp
string egilek
string bfwowqdnhfmbugppibavbamkbjdbnropmezvarj
string x
string dlfgywzzjmkrvj
[...]
```

Run script with Bun (yields values wrapped in arrays, 3rd line always `undefined`, and fails to print next lines):
```
$ bun test.mjs
Starting...
object [ "ajqrsoyrabfleslmw" ]
object [ "ggwazolafkjersvirfrmynnucihuywtgjli" ]
object [ undefined ]
```

Print the following lines:
```
Starting...
string ajqrsoyrabfleslmw
string ggwazolafkjersvirfrmynnucihuywtgjli
string eubcujmhoboaecf
string hokkaxedir
string hngvulerbketqp
string egilek
string bfwowqdnhfmbugppibavbamkbjdbnropmezvarj
string x
string dlfgywzzjmkrvj
[...]
```

### What do you see instead?

It prints the following lines instead:
```
Starting...
object [ "ajqrsoyrabfleslmw" ]
object [ "ggwazolafkjersvirfrmynnucihuywtgjli" ]
object [ undefined ]
```

### Additional information

Notice that there are 3 bugs being reported here:
1. Bun apparently tries to load entire file into memory before yielding lines
2. Current `readline` implementation yields arrays of strings instead of the string itself
3. Third line is always `[undefined]`, and 4th line and below is never printed

yschroe:
### What does this PR do?

- Implements `kFirstEventParam` in `events.on` to make it compatible with the `for await...of` syntax of the `readline` module.
- Fixes a bug in `events.on` which causes events to be skipped
- Fixes #8370 
- Fixes #8850 (Apart from the "Bun apparently tries to load entire file into memory before yielding lines" part)

- [x] Code changes

image


src/js/node/events.js
```diff
@@ -386,21 +388,20 @@ async function* on(emitter, event, options = {}) {
 
   for (const evName of options?.close || []) {
     emitter.on(evName, () => {
-      emitter.emit(event, undefined);
```
yschroe, line 389: Why is `undefined` emitted when the close event is called? I needed to remove this line after my changes or else I was stuck in an infinite loop after all line have been read-in.

src/js/node/events.js
```diff
@@ -386,21 +388,20 @@ async function* on(emitter, event, options = {}) {
...
       done = true;
     });
   }
 
-  while (!done) {
+  while (!done || !unconsumedEvents.isEmpty()) {
```
yschroe: Added this check to continue yielding the already read-in lines after the close event of `readline` is called. This fixes the issue that not all read-in lines were available in the `for await...of` loop.

Is there a situation in which the queued events should not be yielded and the loop should be stopped immediately after the close event is called?

src/js/node/events.js
```diff
@@ -386,21 +388,20 @@ async function* on(emitter, event, options = {}) {
...
     if (!unconsumedEvents.isEmpty()) {
       yield Promise.$resolve(unconsumedEvents.shift());
-    }
-    if (!unconsumedErrors.isEmpty()) {
+    } else if (!unconsumedErrors.isEmpty()) {
       yield Promise.$reject(unconsumedErrors.shift());
+    } else {
```
yschroe: The other `yield` statements needed to be enclosed by `else` because otherwise when .next() is called on the iterator in the `for await...of` loop, the generator function started at the wrong position.

E.g. following situation:
- Pre: Iterator is created via `events.on`, and shortly after 3 events are emitted at the next tick.
1. `.next()` call: No events or errors are queued, so a new unresolved promise is returned. When the first event is emitted in the next tick, the promise is resolved and `.next()` is called. The other two events are put into `unconsumedEvents`.
2. Call: `unconsumedEvents` does have items, so the first yield statement is called.
3. Call: The execution continues after the first yield statement. Since the other yields were not enclosed by `else` a new `unconsumedPromise` is created and yielded. Even though there is a `unconsumedEvent` available.

This last promise is never resolved because of that. This only happens when 3+ events are emitted, so the test case did not catch that.

Disclaimer: I'm not an expert for generator functions but this was the behaviour I noticed and I hope the explanation makes sense.

src/js/node/events.js
```diff
@@ -386,21 +388,20 @@ async function* on(emitter, event, options = {}) {
...
+      const { promise, reject, resolve } = $newPromiseCapability(Promise);
+      unconsumedPromises.push({ reject, resolve });
+      yield promise;
```
yschroe, line 403: And a second thing about this generator function: To notify the `for await...of` loop that the last item is reached, this item should not be `yield`ed but `return`ed. This is not the case here, yet the execution still stops after the last item. Is this an "ungraceful" exit, since the JS engine notices somehow that the last returned promise can never be resolved?

Again, not an expert on generator functions. 😉

paperclover (approved):

paperclover:
ah wait, this is still a draft. the events.on changes look good so far.

yschroe:
> ah wait, this is still a draft. the events.on changes look good so far.

The change already works for `readline`, just want to do a bit more manual testing and add some automated tests.

Btw, should the `events.on` tests be added to `event-emitter.test.ts` or `events-cjs.test.js`? The current tests are in the cjs file, in the other file there is a commented out section "TODO: extensive events.on tests" though.

src/js/node/events.js
```diff
@@ -356,7 +357,7 @@ function once(emitter, type, options) {
   });
 }
 
-async function* on(emitter, event, options = {}) {
+function on(emitter, event, options = {}) {
```
yschroe, line 360: Needed to change this to a normal function, else events were not stored in the queue before the first `.next()` call.

src/js/node/events.js
```diff
@@ -366,42 +367,56 @@ async function* on(emitter, event, options = {}) {
...
+      }
+      // Else: Add event value to queue so it can be consumed by a future Promise.
+      unconsumedEvents.push(ev);
     }
-    unconsumedEvents.push([ev]);
-  });
+  }
+
+  emitter.on(event, options[kFirstEventParam] ? eventHandler : (...args) => eventHandler(args) );
+
```
yschroe: Needed to move the eventHandler logic out of `emitter.on` to handle kFirstEventParam more efficiently. Alternative would be to always use ...args and return args[0] in the case of kFirstEventParam = true, but I remember that those spread operations were quite performance-heavy.

I kinda dislike nested function definitions though, is there a more elegant way to do it?

yschroe:
While testing I found two situations in which the current bun implementation of `events.on` had a different behaviour than node.js so I needed to do some larger changes.


1. If the event was emitted before `.next()` is called, the event was never stored in the queue and lost:
```javascript
const emitter = new EventEmitter();
const asyncIterator = EventEmitter.on(emitter, "hey");

emitter.emit("hey", 1);

await new Promise((resolve) => setTimeout(resolve, 1000));
const { value } = await asyncIterator.next();
console.log(value) // unreachable in bun, node logs [1]
```

2. Emitting multiple values only returned the first value:
```javascript
const emitter = new EventEmitter();
const asyncIterator = EventEmitter.on(emitter, "hey");

process.nextTick(() => {
  emitter.emit("hey", 1, 2, 3);
})

const { value }  = await asyncIterator.next();
console.log(value) // bun: [1], node: [1, 2, 3]
```

I've added some testcases and hand checked the behaviour in node and bun-debug. Both now behave identical in those tests.

If there is anything I've missed and some features of the old implementation are now missing, let me know. Else this is good to go from my point of view.

paperclover:
when events-cjs.test.js was added, it was only for the first test
```
test("in cjs, events is callable", () => {
  const EventEmitter = require("events");
  new EventEmitter();
});
```
this cannot be in the main test file as it is an es module. and this is specifically testing `require`. though it was also before some major commonjs overhauls. `require` is now the same function as in commonjs when used in an es module, so this distinction is no longer relevant.

i think it is safe to move all of these tests into the larger event-emitter.test.ts file.

src/js/node/events.js
paperclover: i'm not sure theres much to get around this as the nested function is capturing a lot of shared state.

paperclover (changes requested):
Looks good to me. Putting this as request-changes just because I think it would be good to merge the two test files as the distinction between being in a commonjs module isn't what is being tested.

src/js/node/events.js
```diff
@@ -356,52 +357,67 @@ function once(emitter, type, options) {
   });
 }
 
-async function* on(emitter, event, options = {}) {
+function on(emitter, event, options = {}) {
   const signal = options.signal;
-  if (signal?.aborted) throw new AbortError(undefined, { cause: signal?.reason });
```
yschroe, line 361: Removed since this part is only executed on creation of the Iterator and not during iteration itself. That part should already be handled by the `signal?.addEventListener("abort", ...)` part later.

src/js/node/events.js
```diff
@@ -356,52 +357,67 @@ function once(emitter, type, options) {
...
     emitter.emit("error", new AbortError(undefined, { cause: signal?.reason }));
   });
 
+  // If any of the close events is emitted -> remove listeners
+  // and yield only the remaining queued-up values in iterator.
   for (const evName of options?.close || []) {
     emitter.on(evName, () => {
-      emitter.emit(event, undefined);
+      emitter.removeListener(event, eventHandler)
+      emitter.removeListener("error", errorHandler)
```
yschroe: Instead of checking for `done` in the listener body, I think it's cleaner to just remove the listeners.

yschroe:
> Looks good to me. Putting this as request-changes just because I think it would be good to merge the two test files as the distinction between being in a commonjs module isn't what is being tested.

Sounds good! Did merge the two test files and did a few last small modifications to `events.on`.

paperclover (approved):

Jarred-Sumner:
Thank you

PR #8889, opened 2024-02-13
fix(windows): use allocated memory for `\\\\.\\NUL` replacement on windows

dylan-conway:
### What does this PR do?
Allocates `\\\\.\\NUL` because the memory will be freed in `Blob.Store.deinit` when it's gc'd.

If we want to keep this memory static, we could add a flag on the `Store` struct for whether `pathlike.path` is allocated or not, but I'm less certain about the lifetime of the memory if we do this.

tested manually

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -1529,7 +1529,7 @@ pub const Blob = struct {
                     var slice = path_or_fd.path.slice();
 
                     if (Environment.isWindows and bun.strings.eqlComptime(slice, "/dev/null")) {
-                        path_or_fd.path.deinit();
+                        path_or_fd.deinit();
                         path_or_fd.* = .{
                             .path = .{
                                 // this memory is freed with this allocator in `Blob.Store.deinit`
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -1526,12 +1526,17 @@ pub const Blob = struct {
         const path: JSC.Node.PathOrFileDescriptor = brk: {
             switch (path_or_fd.*) {
                 .path => {
-                    const slice = path_or_fd.path.slice();
+                    var slice = path_or_fd.path.slice();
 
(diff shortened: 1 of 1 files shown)


paperclover (approved):
damn

src/bun.js/webcore/blob.zig
```diff
@@ -1526,12 +1526,17 @@ pub const Blob = struct {
...
-                        // in PathLike.deinit does not free anything.
-                        path_or_fd.* = .{ .path = .{ .string = bun.PathString.init("\\\\.\\NUL") } };
+                        path_or_fd.path.deinit();
+                        path_or_fd.* = .{
+                            .path = .{
+                                // this memory is freed with this allocator in `Blob.Store.deinit`
+                                .string = bun.PathString.init(allocator.dupe(u8, "\\\\.\\NUL") catch bun.outOfMemory()),
+                            },
+                        };
+                        slice = path_or_fd.path.slice();
```
jdalton, line 1539: Below https://github.com/oven-sh/bun/pull/8889/files#diff-fccddb3613f2e886d367c9383b53da6fa99925167af2f0fb861dfaa201666261R1546
when doing this kind of pattern the `path_or_fd` is `path_or_fd.deinit();` and not just `path_or_fd.path` being `deinit`.

On https://github.com/oven-sh/bun/pull/8889/files#diff-fccddb3613f2e886d367c9383b53da6fa99925167af2f0fb861dfaa201666261L1552 `deinit()` is not called at all (and maybe should be)?

dylan-conway: I'll change it for consistency, but because we're in the `.path` case of the switch it's guaranteed to be the active tag.

Below, where `.deinit` isn't called, we shouldn't deinit because we take a copy and break out of the scope with it

jdalton: Cool. My thought was that since it's paving the entire union it's a good call to `.deinit` the old. I see via the source that it just defers to `path.deinit` but still a good call since that could change in the future.

PR #8890, opened 2024-02-14
windows: fix bundler reliability and nuke WSL-based codegen scripts

paperclover:
### What does this PR do?

This removes `codegen.ps1` and `cross-compile-codegen.sh`, and makes the build process use `bun.exe` to run the generators, instead of relying on WSL. This will improve iteration time as the old approach nuked `.\build\codegen` every rebuild, while this will support the same few-second js rebuilds that I engineered for Mac and Linux.

To make this possible, a directory invalidation bug in `Bun.build` was fixed. I consider bundle-functions.ts to be a sort of mini stress-test for this. Not sure why it is so unreliable on Windows and not linux, but I do remember some flakiness on Bun.build back when these were written; I had some workarounds:

image

It is probably a good idea to explore if these can be removed now. But not now.

This PR unfortunately took alot of sanity out of me so there are some odd looking changes with certain new debug logs in seemingly random places. Got some more big stability fixes after this that will probably be even harder to debug.


src/bundler.zig
```diff
@@ -1788,7 +1826,7 @@ pub const Bundler = struct {
         var entry_points = try allocator.alloc(_resolver.Result, bundler.options.entry_points.len);
         entry_points = entry_points[0..bundler.enqueueEntryPoints(entry_points, true)];
 
-        if (log.level.atLeast(.debug)) {
+        if (true) {
```
dylan-conway: revert

src/bundler.zig
```diff
@@ -372,16 +372,54 @@ pub const Bundler = struct {
         this.resolver.allocator = allocator;
     }
 
-    pub inline fn resolveEntryPoint(bundler: *Bundler, entry_point: string) anyerror!_resolver.Result {
+    fn _resolveEntryPoint(bundler: *Bundler, entry_point: string) !_resolver.Result {
         return bundler.resolver.resolve(bundler.fs.top_level_dir, entry_point, .entry_point) catch |err| {
-            const has_dot_slash_form = !strings.hasPrefix(entry_point, "./") and brk: {
-                return bundler.resolver.resolve(bundler.fs.top_level_dir, try strings.append(bundler.allocator, "./", entry_point), .entry_point) catch break :brk false;
+            if (!std.fs.path.isAbsolute(entry_point) and !strings.hasPrefix(entry_point, "./")) brk: {
```
dylan-conway, line 377: do we know that paths always have or expect forward slashes at this point? Maybe we add a debug assert just in case

src/bun.zig
```diff
@@ -68,6 +68,7 @@ pub const FileDescriptor = enum(FileDescriptorInt) {
     _,
 
     pub inline fn int(fd: FileDescriptor) FileDescriptorInt {
+        // TODO(@paperdave): make this a compile error to call on windows. every usage is incorrect.
```
jdalton, line 71: Like?
```suggestion
        if (comptime Environment.isWindows) @compileError("Calling FileDescriptor.int() is not allowed on Windows");
```

src/bundler.zig
```diff
@@ -372,16 +372,54 @@ pub const Bundler = struct {
...
 
-            bundler.log.addErrorFmt(null, logger.Loc.Empty, bundler.allocator, "{s} resolving \"{s}\" (entry point)", .{ @errorName(err), entry_point }) catch unreachable;
+            bundler.resolver.bustDirCache(buster_name);
 
-            return err;
+            return _resolveEntryPoint(bundler, entry_point) catch {
+                // return the original error
+                bundler.log.addErrorFmt(null, logger.Loc.Empty, bundler.allocator, "{s} resolving \"{s}\" (entry point)", .{ @errorName(err), entry_point }) catch bun.outOfMemory();
+                // return err;
+                @panic("lose");
```
jdalton, line 421: Is the `@panic(` meant here? If so, 🙏 a more descriptive panic message.

src/codegen/bundle-functions.ts
```diff
@@ -185,7 +185,11 @@ $$capture_start$$(${fn.async ? "async " : ""}${
       define,
       minify: { syntax: true, whitespace: false },
     });
+    console.log(build);
```
jdalton: ☝️ Debug?
```suggestion
```

src/codegen/client-js.ts
```diff
@@ -7,12 +7,12 @@ let $debug_log_enabled = ((env) => (
   || (env.BUN_DEBUG_JS && env.BUN_DEBUG_JS !== '0')
   || (env.BUN_DEBUG_${filepath
     .replace(/^.*?:/, "")
```
jdalton: Fancy name magic! Can you add a comment about the transform above it. Something like
```suggestion
    // Transforms a posix file path from /path/to/file to BUN_DEBUG_PATH_TO_FILE
    // and Windows file path from C:/path/to/file to BUN_DEBUG_PATH_TO_FILE
    .replace(/^.*?:/, "")
```

paperclover: - this comment is not perfectly accurate (does not handle drive letters)
- i don't think it's very unclear what is happening here that two lines of comments are needed

src/codegen/internal-module-registry-scanner.ts
```diff
@@ -7,6 +7,7 @@ export function createInternalModuleRegistry(basedir: string) {
     .flatMap(dir => readdirRecursive(path.join(basedir, dir)))
     .filter(file => file.endsWith(".js") || (file.endsWith(".ts") && !file.endsWith(".d.ts")))
     .map(file => file.slice(basedir.length + 1))
+    .map(x => x.replaceAll("\\", "/"))
```
jdalton, line 10: Can we use the `path.normalize(x).split(path.delimiter).join("/")` method here? Maybe a convoluted case. Thinking more in a `node:path`- way.

src/codegen/internal-module-registry-scanner.ts
```diff
@@ -7,6 +7,7 @@ export function createInternalModuleRegistry(basedir: string) {
     .flatMap(dir => readdirRecursive(path.join(basedir, dir)))
     .filter(file => file.endsWith(".js") || (file.endsWith(".ts") && !file.endsWith(".d.ts")))
```
jdalton, line 8: ☝️ There's a helper internal to Node's path that would simplify the above (gets the longest ext first). Nice to see a case for it externally.

src/fs.zig
```diff
@@ -752,7 +752,7 @@ pub const FileSystem = struct {
         pub fn needToCloseFiles(rfs: *const RealFS) bool {
             // On Windows, we must always close open file handles
             // Windows locks files
-            if (comptime !FeatureFlags.store_file_descriptors) {
+            if (!FeatureFlags.store_file_descriptors or Environment.isWindows) {
```
jdalton: Should we make this `FeatureFlag.store_file_descriptors` defined with`Environment.isWindows` in mind so if referenced in other places it will already be accounted for?

src/codegen/helpers.ts
```diff
@@ -80,3 +80,11 @@ export function writeIfNotChanged(file: string, contents: string) {
...
   }
 }
+
+export function pathToUpperSnakeCase(filepath: string) {
+  return filepath
+    .replace(/^.*?:/, "")
+    .split(/[-_./\\]/g)
+    .join("_")
+    .toUpperCase();
+}
```
jdalton, line 90: ❤️

src/fs.zig
paperclover: jarred decided

i should revert this to how it was before
and actually implement it. the comment is old and incorrect.

src/codegen/internal-module-registry-scanner.ts
paperclover: this feels, overly convoluted

and it will be harder to read.

src/codegen/bundle-functions.ts
paperclover: huh i thought i already removed this

src/codegen/internal-module-registry-scanner.ts
jdalton: Ya, looking for more of a https://github.com/wessberg/crosspath kinda thing to simplify a lot of our normalize to posix stuff.

src/codegen/client-js.ts
jdalton: > * this comment is not perfectly accurate (does not handle drive letters)

I wrote "Something like" 😉 At a glance it looks like (/^.*?:/, '') will trim the drive letter off.

> * i don't think it's very unclear what is happening here that two lines of comments are needed

When dealing with paths in formats of multiple platforms with drive letters/UNC and separator normalization comments on INPUT and OUTPUT can be very helpful.
 
The `pathToUpperSnakeCase` helper that replaced this does help with readability 💪

.github/workflows/bun-windows.yml
```diff
@@ -178,6 +178,11 @@ jobs:
           path: bun-deps
           key: ${{ steps.cache-deps-restore.outputs.cache-primary-key }}
 
+  # TODO(@paperdave): stop relying on this and use bun.exe to build itself.
+  # we cant do that now because there isn't a tagged release to use.
```
jdalton, line 182: Exciting!

PR #8892, opened 2024-02-14
Fix tsconfig path longest prefix matching

james-elicx:
### What does this PR do?

Fixes the logic for matching a tsconfig path so that it will match paths with a longer prefix when the suffix is the same length, as it previously would incorrectly only allow matches when the suffix is greater.

The comment above the if statement suggests that this change is the intended way for the check to work.

- [x] Code changes

I wrote automated tests

If Zig files changed:

- [ ] ~~JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed~~

### DIFF
--- a/test/tsconfig.json
+++ b/test/tsconfig.json
@@ -26,7 +26,8 @@
       "node-harness": ["js/node/harness.ts"],
       "deno:harness": ["js/deno/harness.ts"],
       "foo/bar": ["js/bun/resolve/baz.js"],
-      "@faasjs/*": ["js/bun/resolve/*.js", "js/bun/resolve/*/src/index.js"]
+      "@faasjs/*": ["js/bun/resolve/*.js", "js/bun/resolve/*/src/index.js"],
+      "@faasjs/larger/*": ["js/bun/resolve/*/larger-index.js"]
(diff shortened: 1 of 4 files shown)


paperclover (approved):

jdalton (approved):
Nice. Thank you @james-elicx!

test/tsconfig.json
```diff
@@ -26,7 +26,8 @@
       "node-harness": ["js/node/harness.ts"],
       "deno:harness": ["js/deno/harness.ts"],
       "foo/bar": ["js/bun/resolve/baz.js"],
-      "@faasjs/*": ["js/bun/resolve/*.js", "js/bun/resolve/*/src/index.js"]
+      "@faasjs/*": ["js/bun/resolve/*.js", "js/bun/resolve/*/src/index.js"],
+      "@faasjs/larger/*": ["js/bun/resolve/*/larger-index.js"]
```
jdalton, line 30: :heart: PRs with unit tests!

jdalton:
Test fails seem unrelated. I'll merge. Feel free to revert if needed.

PR #8898, opened 2024-02-14
Release weak refs before responding to websocket messages

Jarred-Sumner:
### What does this PR do?

This might slightly reduce memory usage in the WebSocket client

- We were not releasing weak refs in the `"message"` callback
- We were not draining microtasks in the `"close"` callback
- We were not draining microtasks in the `rejectUnauthorized` (TLS) callback

CI


src/http/websocket_http_client.zig
```diff
@@ -1055,19 +1070,13 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
...
             };
-            if (comptime Environment.isDebug) {
-                this.event_loop.debug.enter();
-            }
+            this.event_loop.enter();
             defer {
-                if (comptime Environment.isDebug) {
-                    this.event_loop.debug.enter();
-                }
+                this.event_loop.enter();
```
jdalton: ☝️ Was this meant as a defer `enter()` and not an `exit()`?

src/http/websocket_http_client.zig
```diff
@@ -1020,6 +1032,9 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
             this.clearData();
             if (this.outgoing_websocket) |ws| {
                 this.outgoing_websocket = null;
+                const loop = JSC.VirtualMachine.get().eventLoop();
+                loop.enter();
+                defer loop.exit();
```
jdalton: (learning) So calling into the event `loop.enter()` and then to `exit()` after the block via `defer` is just to handle `ws.didAbruptClose(ErrorCode.ended);`?

src/http/websocket_http_client.zig
cirospaciari: probably?
```suggestion
                this.event_loop.exit();
```

src/http/websocket_http_client.zig
cirospaciari: Yeah, internally we have a count if is 1 we free weak refs and drain microtasks

cirospaciari (commented):
Should we also do loop.enter/exit in this places, instead of only calling drainMicrotasks?

https://github.com/oven-sh/bun/blob/2d78d835d5aae3849d95b8eba93994b31823fbea/src/http/websocket_http_client.zig#L416-L418

https://github.com/oven-sh/bun/blob/2d78d835d5aae3849d95b8eba93994b31823fbea/src/http/websocket_http_client.zig#L1145-L1150

cirospaciari (approved):

Issue #8886, opened 2024-02-13
`deflateSync` produces binary that `inflateSync` fails to decompress


MarByteBeep:
### What version of Bun is running?

1.0.26

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

```ts
const buffer = new Uint8Array([0x12, 0x01, 0x3, 0x5, 0x5]);

{
    const compressed = Bun.deflateSync(buffer, { level: 9, windowBits: -15 });

    console.log(compressed);

    const decompressed = Bun.inflateSync(compressed);

    // decompressed isn't a Uint8array but an error: invalid stored block lengths
    console.log(decompressed);
}

{
    const compressed = Bun.deflateSync(buffer, { level: 9, windowBits: -15 });

    console.log(compressed);

    const decompressed = Bun.gunzipSync(compressed);

    // decompressed works fine
    console.log(decompressed);
}
```

### What is the expected behavior?

A binary compressed with `deflateSync` should always be decompressible with `inflateSync`. When using the `{ level: 9, windowBits: -15 }` settings, it is not.

However it IS decompressible with `gunzipSync`.

So we end up with the weird situation in which you have to use `gunzipSync` to decompress and `deflateSync` to compress.

### What do you see instead?

`decompressed` is error: invalid stored block lengths

### Additional information

As per discord discussion, might be related to https://github.com/oven-sh/bun/pull/8529

Also I ran into: https://github.com/oven-sh/bun/issues/6401 as well, the docs seem to be incorrect.

argosphil:
You're passing a `windowBits` parameter value of -15.

From the zlib documentation at https://www.zlib.net/manual.html:

> The windowBits parameter is the base two logarithm of the window size (the size of the history buffer). **It should be in the range 8..15 for this version of the library**. Larger values of this parameter result in better compression at the expense of memory usage. The default value is 15 if deflateInit is used instead. 

Internally, a negative value of `windowBits` is used to switch between gzip and zlib compression modes. We should catch negative values in `opts.windowBits` and throw an error, but that's about all we can do.

MarByteBeep:
That then creates a new problem: my original binary is compressed and starts with a`78 DA` header, which is a perfectly valid and common zlib header. I need to be able to recompress said data and end up with that header again. The *only* setting that did that was windowBits -15. How am I able to recreate the original if you simply prevent that value from being passed?

MarByteBeep:
[...]

PR #8901, opened 2024-02-14
fix: throw on gunzipSync/inflateSync error

### LINKED ISSUE #8887: `inflateSync` and `gunzipSync` don't throw but return an Error instead

1.0.26

```ts

{
  const buffer = new Uint8Array([0x12, 0x01, 0x3, 0x5, 0x5]);
  const compressed = Bun.deflateSync(buffer, { level: 9, windowBits: -15 });

  console.log(compressed);

  const decompressed = Bun.inflateSync(compressed);

  // decompressed isn't a Uint8array but an error: invalid stored block lengths
  console.log(decompressed);
}
```

or

```ts
{
  const buffer = new Uint8Array([0x12, 0x01, 0x3, 0x5, 0x5]);
  const decompressed = Bun.inflateSync(buffer);

  // decompressed isn't a Uint8array but an error: Zlib returned an error
  console.log(decompressed);
}
```

Either throw or return `undefined` and add `undefined` as a return type of both functions.

### What do you see instead?

Return type is `Error`, without `Error` being specified as return type

### Additional information

Also related to https://github.com/oven-sh/bun/issues/8886

argosphil:
### What does this PR do?

Fixes #8887 .

This fixes `inflateSync`, `gzipSync`, `deflateSync`, and `gunzipSync` to throw errors rather than return them. 

There's also a drive-by fix to similar error handling in the hashing code. 

- [x] Code changes

I wrote automated tests for the decompression case. I do not know how to trigger a zlib error for the compression case.

### DIFF
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -2079,6 +2079,7 @@ pub const Crypto = struct {
                     .err => {
                         const error_instance = value.toErrorInstance(globalObject);
                         globalObject.throwValue(error_instance);
+                        return .zero;
                     },
                     .hash => |h| {
                         return JSC.ZigString.init(h).toValueGC(globalObject);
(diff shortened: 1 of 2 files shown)


argosphil:
Ideally, I'd like to merge this before any renaming that might result from #8886 .

paperclover (approved):

PR #8904, opened 2024-02-14
feat: More robust and faster shell escaping

zackradisic:
### What does this PR do?

This makes shell escaping faster by changing the `escape()` and `needsEscape()` function to work with any string encoding (it previously would convert the string to UTF-8 and allocate it). The `needsEscape()` implementation for each string encoding is also SIMD-accelerated.

Also, instead of wrapping the string in quotes and adding backslashes to prevent the lexer from interpreting whitespace, quotes, etc. as special syntax, it passes the strings to the lexer to directly turn them into `Text` tokens.

This fixes makes the escaping logic more robust in situations like this:

```typescript
const str = "cookies & creme"
await $`echo "${str}"`
```

Which would before escape to this:
```typescript
await $`echo ""cookies & creme ""`
```

Which is obviously not correct

- [x] Code changes

Zig files changed:


src/bun.js/api/BunObject.zig
```diff
@@ -393,14 +404,25 @@ pub fn shellParse(
     defer arena.deinit();
 
     const template_args = callframe.argumentsPtr()[1..callframe.argumentsCount()];
+    var stack_alloc = std.heap.stackFallback(@sizeOf(bun.String) * 4, arena.allocator());
```
paperclover, line 407: i wonder if a helper function for this abstraction would be ncie
ive seen it a lot "please stack fallback for <n> items but it technically can be infinity so heap in that case"

src/shell/shell.zig
```diff
@@ -15,6 +15,7 @@ const Glob = @import("../glob.zig");
 const ResolvePath = @import("../resolver/resolve_path.zig");
 const DirIterator = @import("../bun.js/node/dir_iterator.zig");
 const CodepointIterator = @import("../string_immutable.zig").PackedCodepointIterator;
+// const CodepointIterator = @import("../string_immutable.zig").CodepointIterator;
```
paperclover: delete this line

src/shell/shell.zig
```diff
@@ -2082,12 +2229,12 @@ const SrcUnicode = struct {
 
     inline fn index(this: *const SrcUnicode) ?IndexValue {
         if (this.cursor.width + this.cursor.i > this.iter.bytes.len) return null;
-        return .{ .char = this.cursor.c, .width = this.cursor.width };
+        return .{ .char = @intCast(this.cursor.c), .width = this.cursor.width };
```
paperclover: what changed that needed this cast?

src/shell/shell.zig
```diff
@@ -2549,10 +2721,11 @@ pub fn handleTemplateValue(
             const last = array.len -| 1;
             var i: u32 = 0;
             while (array.next()) |arr| : (i += 1) {
-                if (!(try handleTemplateValue(globalThis, arr, out_jsobjs, out_script, jsobjref_buf))) return false;
+                if (!(try handleTemplateValue(globalThis, arr, out_jsobjs, out_script, jsstrings, jsobjref_buf))) return false;
                 if (i < last) {
                     const str = bun.String.init(" ");
-                    if (!try appendBunStr(str, out_script, false)) return false;
+                    defer str.deref();
```
paperclover: you can use `String.static` here. it is identical to init with a string literal.

the defer deref is smart but in this case it doesn't really do anything interesting.

src/shell/shell.zig
```diff
@@ -2686,27 +2929,93 @@ pub fn escapeUnicode(str: []const u8, outbuf: *std.ArrayList(u8)) !void {
...
+
+    const non_ascii = bun.strings.firstNonASCII16([]const u16, str) orelse 0;
+    var cp_buf: [4]u8 = undefined;
+
+    var i: usize = 0;
+    loop: while (i < str.len) {
+        const char: u32 = brk: {
+            if (i < non_ascii) {
+                i += 1;
+                break :brk @intCast(str[i]);
```
paperclover: u16 -> u32 does not need intcast

paperclover (changes requested):
would like to see some very very minor touch ups. i also swear there was an issue open for something related to this but i couldn't find it. but great work

src/shell/shell.zig
```diff
@@ -1907,19 +1927,146 @@ pub fn NewLexer(comptime encoding: StringEncoding) type {
...
+                var i: usize = 0;
+                var digit_buf: [32]u8 = undefined;
+                var digit_buf_count: u8 = 0;
+
+                i += literal.len - 1;
+
+                while (i < bytes.len) : (i += 1) {
+                    switch (bytes[i]) {
+                        '0'...'9' => {
+                            if (digit_buf_count >= 32) {
```
Jarred-Sumner: `digit_buf.len` should work btw

src/shell/shell.zig
zackradisic: It doesn't need to be cast, will remove

Issue #4964, opened 2023-09-11
Implement node:http setTimeout


pierophp:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 5.15.90.4-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

I'm trying to use Bun on our service, which uses node:http setTimeout.

Bun is currently throwing a "not implemented" error:

https://github.com/oven-sh/bun/blob/51d3d4382281f789f8175079ed426a63529eb3e7/src/js/node/http.ts#L737C22-L737C37

That is the only issue that I'm having to implement Bun.

Thank you for the great work!

jebarpg:
I just upgraded bun to 1.0.28 and I am now getting this error as well.

stewones:
same here, we just migrated a big workload to Bun. mysteriously the app seems to be working still.

```
2024-02-22T16:15:48.154379302Z NotImplementedError: Socket in HTTP setTimeout is not yet implemented in Bun.

2024-02-22T16:15:48.154389422Z code: "ERR_NOT_IMPLEMENTED"

2024-02-22T16:15:48.154425661Z at new NotImplementedError (internal:shared:19:27)

2024-02-22T16:15:48.154512381Z at internal:shared:9:73

2024-02-22T16:15:48.154645461Z at setTimeout (node:http:332:25)

2024-02-22T16:15:48.154659741Z at destroyOnTimeout (/usr/src/app/node_modules/follow-redirects/index.js:199:5)

2024-02-22T16:15:48.154834020Z at startTimer (/usr/src/app/node_modules/follow-redirects/index.js:213:5)

2024-02-22T16:15:48.154847020Z at /usr/src/app/node_modules/follow-redirects/index.js:244:5

2024-02-22T16:15:48.154879860Z at dispatchHttpRequest (/usr/src/app/node_modules/typesense/node_modules/axios/lib/adapters/http.js:360:7)

2024-02-22T16:15:48.154991020Z at new Promise (:1:21)

2024-02-22T16:15:48.155002820Z at httpAdapter (/usr/src/app/node_modules/typesense/node_modules/axios/lib/adapters/http.js:48:10)

2024-02-22T16:15:48.155183539Z at dispatchRequest (/usr/src/app/node_modules/typesense/node_modules/axios/lib/core/dispatchRequest.js:58:10)

[...]

PR #8911, opened 2024-02-15
Fixes #8909

### LINKED ISSUE #8909: setTimeout on server doesn't work like nodejs

1.0.26+c75e768a6

Darwin 23.2.0 arm64 arm

in node setTimeout return server object and can be chained, but it bun it return nothing:
```
HTTP.createServer(this.handle).setTimeout(t * 1000).listen(p, h)
HTTPS.createServer({cert, key}, this.handle).setTimeout(t * 1000).listen(p, h)
```

_No response_

### What do you see instead?

`TypeError: undefined is not an object (evaluating 'HTTP.createServer(this.handle).setTimeout(t * 1000).listen')`

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

This adds a helper that will warn once when a method that is not implemented but stubbed out is called

This fixes #8909 by turning it into a no-op that still returns `this`

CI

### DIFF
--- a/src/js/internal/shared.ts
+++ b/src/js/internal/shared.ts
@@ -38,7 +38,7 @@ function warnNotImplementedOnce(feature: string, issue?: number) {
   if (warned.has(feature)) {
     return;
   }
-
+  warned.add(feature);
   console.warn(new NotImplementedError(feature, issue));
 }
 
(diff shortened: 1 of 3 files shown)


paperclover (approved):
adding a todo message does not count as implementing a nodejs api. the code will now run but the custom timeouts will not be respected. the issue linked is a duplicate so it is ok. we should leave #4964 open until this feature is actually implemented.

Jarred-Sumner:
yeah fair

PR #8921, opened 2024-02-15
change how `bunx` caches things

### LINKED ISSUE #4981: `bun create` uses the latest locally cached version even when the `@latest` tag is provided

Darwin 21.6.0 arm64 arm

1. Install any `create-thing` package at a specific (non-latest) version
2. Run `bun create thing@latest`

Bun will use the latest cached version instead of checking the registry, which I think is very unexpected.

```
kevinkoste temp % bun install create-vite@4.3.0
bun add v1.0.0 (822a00c4)

 installed create-vite@4.3.0 with binaries:
  - create-vite
  - cva


 1 packages installed [1442.00ms]
kevinkoste temp % bun create vite@latest
✔ Project name: … test-project
✔ Select a framework: › Vanilla
✔ Select a variant: › JavaScript

Scaffolding project in /Users/kevinkoste/workspace/temp/test-project...

Done. Now run:

  cd test-project
  bun install
  bun run dev

kevinkoste test-project % cat test-project/package.json
{
  "name": "test-project",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "vite build",
    "preview": "vite preview"
  },
  "devDependencies": {
    "vite": "^4.3.0"
  }
}
```

For reference, here's my understanding of how the other package managers handle this:
- **npm**: If no version is specified, `npm create` uses the latest globally installed version. If the `@latest` tag is specified, npm fetches the latest version from the registry and uses that.
- **pnpm** and **yarn**: If no version is specified, fetches the latest version from the registry and uses that.

Bun's current behavior is most similar to npm, except Bun does not fetch the latest version from the registry even if you specify it. I'm guessing this is just a bug and would be a quick fix, but I'd like to propose a step futher:

IMO, Bun should (by default) behave like pnpm - if the user does not specify a tag, download and use the latest version from the registry when using the `bun create` shortcut. It would also be lovely to log a "Using create-vite@4.4.1" line similar to what `bun install` produces. This would save a lot of headache and bug reports for authors of `create-*` packages like myself.

### What do you see instead?

_No response_

### Additional information

_No response_

paperclover:
i need to test further on linux and mac more. maybe do some tests.

Fixes #4981, by making dist-tags pass `--force` in addition to `--no-cache`.

secondly, this makes bunx invalidate it's caches after 24 hours, using a stat or fstat call depending on the context of the check.

lastly, to make it harder to have bunx run broken packages, the location of the bunx cache has been moved from `/tmp/{package_name}--bunx/` to `/tmp/bunx-{uid}-{package_name}/`. the reasons for this are outlined in a comment i left. see #7504 for where these concerns come up, but the linked issue is for `bun-node` from --bun, and to be addressed in a different patch.

some other random changes made as well

disallow formatting FDImpl as `{d}`, this is ambiguous and almost certainly a mistake.


src/cli.zig
```diff
@@ -1797,10 +1797,7 @@ pub const Command = struct {
             const file = bun.toLibUVOwnedFD(((brk: {
                 if (std.fs.path.isAbsolute(script_name_to_search)) {
                     var win_resolver = resolve_path.PosixToWinNormalizer{};
-                    var resolved = win_resolver.resolveCWD(script_name_to_search) catch @panic("Could not resolve path");
-                    if (comptime Environment.isWindows) {
-                        resolved = resolve_path.normalizeString(resolved, true, .windows);
-                    }
+                    const resolved = win_resolver.resolveCWD(script_name_to_search) catch @panic("Could not resolve path");
```
dylan-conway: can you check that `bun-debug (run) C:\path\to\..\file.js` works? I think that's why normalizeString was here before

src/cli/bunx_command.zig
```diff
@@ -365,19 +403,64 @@ pub const BunxCommand = struct {
...
+                            if (rc != 0) {
+                                break :is_stale true;
+                            }
+                            break :is_stale std.time.timestamp() - stat.mtime().tv_sec > seconds_cache_valid;
+                        }
+                    };
+
+                    if (is_stale) {
+                        // If delete fails, oh well. Hope installation takes care of it.
+                        // std.fs.deleteDirAbsolute(out) catch {};
```
dylan-conway: is this supposed to be commented?

src/cli/bunx_command.zig
```diff
@@ -402,16 +485,16 @@ pub const BunxCommand = struct {
...
                                 try this_bundler.fs.dirname_store.append(@TypeOf(out), out),
                                 this_bundler.fs.top_level_dir,
                                 this_bundler.env,
                                 passthrough,
                                 null,
                             );
-                            // we are done!
-                            Global.exit(0);
+                            // runBinary is noreturn
+                            comptime unreachable;
```
dylan-conway, line 519: maybe we should give it `!noreturn` return type. It looks like all the switch cases are handled

src/cli/bunx_command.zig
```diff
@@ -498,16 +582,16 @@ pub const BunxCommand = struct {
...
                 try this_bundler.fs.dirname_store.append(@TypeOf(out), out),
                 this_bundler.fs.top_level_dir,
                 this_bundler.env,
                 passthrough,
                 null,
             );
-            // we are done!
-            Global.exit(0);
+            // runBinary is noreturn
+            comptime unreachable;
```
dylan-conway, line 613: yeah let's change the return type to `!noreturn`

dylan-conway: oh, it already is lol

dylan-conway (commented):
Other than the `normalizeString` check, this looks good to me. I like the consistency across platforms with this cache

src/cli.zig
paperclover: Didn't think of that. Makes me think we should normalize this string on all platforms.

Node does path.resolve on the argv[1]:

src/cli/bunx_command.zig
paperclover: ah. i was experimenting with not actually deleting the directory. i'll finish the experiments and check if it's needed or not.

i think it isn't

src/cli/bunx_command.zig
paperclover: the comptime unreachable asserts runBinary is noreturn
and to make this easier to read, so no one can possibly think "oh what happens when this returns".

it could be argued not required. what do you think?

dylan-conway: let's keep it

src/cli.zig
paperclover: im going to revert this and just fix it in a followup

src/cli/bunx_command.zig
```diff
@@ -357,24 +356,61 @@ pub const BunxCommand = struct {
...
+        // Reasoning:
+        // - Prefix with "bunx" to identify the bunx cache, make it easier to "rm -r"
+        //   - Suffix would not work because scoped packages have a "/" in them, and
+        //     before Bun 1.1 this was practically impossible to clear the cache manually.
+        //     It was easier to just remove the entire temp directory.
+        // - Use the uid to prevent conflicts between users. If the paths were the same
+        //   across users, you run into permission conflicts
+        //   - If you set permission to 777, you run into a potential attack vector
+        //     where a user can replace the directory with malicious code.
+        const uid = if (bun.Environment.isPosix) bun.C.getuid();
```
dylan-conway: sad that `getuid` isn't on windows. what if we hash the result from `GetUserName`?

dylan-conway: if we do this code can be the same on each platform, other than the `;`

paperclover: we do not need to because temp dir is (likely) already in a user owned path: `%LOCALAPPDATA%\Temp`. i will clarify that.

dylan-conway (approved):
LGTM after `GetUserName` hash on windows

src/cli/bunx_command.zig
```diff
@@ -365,19 +439,61 @@ pub const BunxCommand = struct {
                 absolute_in_cache_dir,
             )) |destination| {
                 const out = bun.asByteSlice(destination);
-                _ = try Run.runBinary(
+
+                // If this directory was installed by bunx, we want to perform cache invalidation on it
+                // this way running `bunx hello` will update hello automatically to the latest version
+                if (std.mem.startsWith(u8, out, bunx_cache_dir)) {
```
Jarred-Sumner: `strings.hasPrefix`

PR #8932, opened 2024-02-15
Allow for smaller allocations in node:path methods

jdalton:
### What does this PR do?

A follow-up enhancement mentioned by @paperdave to allow for smaller path allocations.

Ran tests locally.

### DIFF
--- a/src/bun.js/bindings/Path.cpp
+++ b/src/bun.js/bindings/Path.cpp
@@ -29,31 +29,21 @@ using namespace JSC;
 
 // clang-format off
 #define DEFINE_CALLBACK_FUNCTION_BODY(ZigFunction) JSC::VM& vm = globalObject->vm(); \
-    auto* thisObject = JSC::jsDynamicCast<JSC::JSFinalObject*>( callFrame->thisValue()); \
-    auto scope = DECLARE_THROW_SCOPE(vm); \
-    if (!thisObject) \
+    auto* thisObject = JSC::jsDynamicCast<JSC::JSFinalObject*>(callFrame->thisValue()); \
+    if (!thisObject) { \
+        auto scope = DECLARE_THROW_SCOPE(vm); \
         return throwVMTypeError(globalObject, scope); \
+    } \
     auto argCount = static_cast<uint16_t>(callFrame->argumentCount()); \
     WTF::Vector<JSC::EncodedJSValue, 16> arguments; \
     arguments.reserveInitialCapacity(argCount); \
-     if (argCount) { \
+    if (argCount) { \
         for (uint16_t i = 0; i < argCount; ++i) { \
             arguments.unsafeAppendWithoutCapacityCheck(JSC::JSValue::encode(callFrame->uncheckedArgument(i))); \
         } \
-     } \
-    auto clientData = WebCore::clientData(vm); \
-    auto isWindows = thisObject->get(globalObject, clientData->builtinNames().isWindowsPrivateName()); \
-    JSC::JSValue result = JSC::JSValue::decode( \
-        ZigFunction(globalObject, isWindows.asBoolean(), reinterpret_cast<JSC__JSValue*>(arguments.data()), argCount) \
-    ); \
-    JSC::JSObject *obj = result.getObject(); \
-    if (UNLIKELY(obj != nullptr && obj->isErrorInstance())) { \
-        scope.throwException(globalObject, obj); \
-        return JSC::JSValue::encode(JSC::jsUndefined()); \
     } \
-    if (UNLIKELY(scope.exception())) \
-        return JSC::JSValue::encode(JSC::jsUndefined()); \
-    return JSC::JSValue::encode(result);
+    auto isWindows = thisObject->get(globalObject, WebCore::clientData(vm)->builtinNames().isWindowsPrivateName()); \
+    return ZigFunction(globalObject, isWindows.asBoolean(), reinterpret_cast<JSC__JSValue*>(arguments.data()), argCount);
 
 // clang-format on
 
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -22,17 +22,16 @@ const Fs = @import("../../fs.zig");
 const IdentityContext = @import("../../identity_context.zig").IdentityContext;
 const JSC = bun.JSC;
 const Mode = bun.Mode;
-const PathBuffer = bun.PathBuffer;
-const PathString = bun.PathString;
 const Shimmer = @import("../bindings/shimmer.zig").Shimmer;
 const Syscall = bun.sys;
 const URL = @import("../../url.zig").URL;
 const Value = std.json.Value;
-const WPathBuffer = bun.WPathBuffer;
+
+const PATH_MIN_WIDE = 4096; // 4 KB
 
 const stack_fallback_size_small = switch (Environment.os) {
-    // Up to 4 KB, instead of MAX_PATH_BYTES which is 96kb on Windows, ouch!
-    .windows => 4096,
+    // Up to 4 KB, instead of MAX_PATH_BYTES which is 96 KB on Windows, ouch!
+    .windows => PATH_MIN_WIDE,
     else => bun.MAX_PATH_BYTES,
 };
 
@@ -579,9 +578,9 @@ pub const Encoding = enum(u8) {
     pub fn encodeWithSize(encoding: Encoding, globalObject: *JSC.JSGlobalObject, comptime size: usize, input: *const [size]u8) JSC.JSValue {
         switch (encoding) {
             .base64 => {
-                var base64: [std.base64.standard.Encoder.calcSize(size)]u8 = undefined;
-                const len = bun.base64.encode(&base64, input);
-                return JSC.ZigString.init(base64[0..len]).toValueGC(globalObject);
+                var buf: [std.base64.standard.Encoder.calcSize(size)]u8 = undefined;
+                const len = bun.base64.encode(&buf, input);
+                return JSC.ZigString.init(buf[0..len]).toValueGC(globalObject);
             },
             .base64url => {
                 var buf: [std.base64.url_safe_no_pad.Encoder.calcSize(size)]u8 = undefined;
@@ -649,7 +648,7 @@ pub const Encoding = enum(u8) {
 };
 
 const PathOrBuffer = union(Tag) {
-    path: PathString,
+    path: bun.PathString,
     buffer: Buffer,
 
     pub const Tag = enum { path, buffer };
@@ -673,7 +672,7 @@ pub fn CallbackTask(comptime Result: type) type {
 }
 
 pub const PathLike = union(enum) {
-    string: PathString,
+    string: bun.PathString,
     buffer: Buffer,
     slice_with_underlying_string: bun.SliceWithUnderlyingString,
     threadsafe_string: bun.SliceWithUnderlyingString,
@@ -730,7 +729,7 @@ pub const PathLike = union(enum) {
         };
     }
 
-    pub fn sliceZWithForceCopy(this: PathLike, buf: *[bun.MAX_PATH_BYTES]u8, comptime force: bool) if (force) [:0]u8 else [:0]const u8 {
+    pub fn sliceZWithForceCopy(this: PathLike, buf: *bun.PathBuffer, comptime force: bool) if (force) [:0]u8 else [:0]const u8 {
         const sliced = this.slice();
 
         if (Environment.isWindows) {
@@ -757,15 +756,15 @@ pub const PathLike = union(enum) {
         return buf[0..sliced.len :0];
     }
 
-    pub inline fn sliceZ(this: PathLike, buf: *PathBuffer) [:0]const u8 {
+    pub inline fn sliceZ(this: PathLike, buf: *bun.PathBuffer) [:0]const u8 {
         return sliceZWithForceCopy(this, buf, false);
     }
 
-    pub inline fn sliceW(this: PathLike, buf: *PathBuffer) [:0]const u16 {
+    pub inline fn sliceW(this: PathLike, buf: *bun.PathBuffer) [:0]const u16 {
         return strings.toWPath(@alignCast(std.mem.bytesAsSlice(u16, buf)), this.slice());
     }
 
-    pub inline fn osPath(this: PathLike, buf: *PathBuffer) bun.OSPathSliceZ {
+    pub inline fn osPath(this: PathLike, buf: *bun.PathBuffer) bun.OSPathSliceZ {
         if (comptime Environment.isWindows) {
             return sliceW(this, buf);
         }
@@ -1927,14 +1926,10 @@ pub const Path = struct {
 
     const StringBuilder = @import("../../string_builder.zig");
 
-    inline fn getPathBufLenT(comptime T: type) usize {
-        return if (T == u16) PATH_MAX_WIDE else MAX_PATH_BYTES;
-    }
-
     /// Based on Node v21.6.1 path.parse:
     /// https://github.com/nodejs/node/blob/6ae20aa63de78294b18d5015481485b7cd8fbb60/lib/path.js#L919
     /// The structs returned by parse methods.
-    fn ParsedPath(comptime T: type) type {
+    fn PathParsed(comptime T: type) type {
         return struct {
             root: []const T = "",
             dir: []const T = "",
@@ -1953,8 +1948,14 @@ pub const Path = struct {
         };
     }
 
-    pub const MAX_PATH_BYTES = bun.MAX_PATH_BYTES;
-    pub const PATH_MAX_WIDE = windows.PATH_MAX_WIDE;
+    pub fn MAX_PATH_SIZE(comptime T: type) usize {
+        return if (T == u16) windows.PATH_MAX_WIDE else bun.MAX_PATH_BYTES;
+    }
+
+    pub fn PATH_SIZE(comptime T: type) usize {
+        return if (T == u16) PATH_MIN_WIDE else bun.MAX_PATH_BYTES;
+    }
+
     pub const shim = Shimmer("Bun", "Path", @This());
     pub const name = "Bun__Path";
     pub const include = "Path.h";
@@ -2021,7 +2022,7 @@ pub const Path = struct {
     }
 
     pub fn getCwdWindowsU8(buf: []u8) MaybeBuf(u8) {
-        const u16Buf: [PATH_MAX_WIDE]u16 = undefined;
+        const u16Buf: bun.WPathBuffer = undefined;
         switch (getCwdWindowsU16(&u16Buf)) {
             .result => |r| {
                 // Handles conversion from UTF-16 to UTF-8 including surrogates ;)
@@ -2062,7 +2063,7 @@ pub const Path = struct {
         if (comptime Environment.isWindows) {
             return getCwdWindowsU16(&buf);
         }
-        const u8Buf: PathBuffer = undefined;
+        const u8Buf: bun.PathBuffer = undefined;
         const result = strings.convertUTF8toUTF16InBuffer(&buf, bun.getcwd(strings.convertUTF16ToUTF8InBuffer(&u8Buf, buf))) catch {
             return MaybeBuf(u16).errnoSys(0, Syscall.Tag.getcwd).?;
         };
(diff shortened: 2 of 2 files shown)


src/bun.js/bindings/Path.cpp
```diff
@@ -43,17 +42,10 @@ using namespace JSC;
...
-    JSC::JSObject *obj = result.getObject(); \
-    if (UNLIKELY(obj != nullptr && obj->isErrorInstance())) { \
-        scope.throwException(globalObject, obj); \
-        return JSC::JSValue::encode(JSC::jsUndefined()); \
-    } \
+    auto result = ZigFunction(globalObject, isWindows.asBoolean(), reinterpret_cast<JSC__JSValue*>(arguments.data()), argCount); \
     if (UNLIKELY(scope.exception())) \
         return JSC::JSValue::encode(JSC::jsUndefined()); \
-    return JSC::JSValue::encode(result);
+    return result;
```
jdalton: ☝️ Tracked down why `return .zero` from path methods would cause a segfault. Now it's safe to `return .zero`.

src/bun.js/node/types.zig
```diff
@@ -3444,21 +3509,20 @@ pub const Path = struct {
                             //   device =
                             //     `\\\\${firstPart}\\${StringPrototypeSlice(path, last, j)}`;
                             //   rootEnd = j;
-                            var tmpBuf1: [getPathBufLenT(T)]T = undefined;
```
jdalton, line 3447: ☝️ Removed need for tmpBuf here.

src/bun.js/bindings/Path.cpp
paperclover: thanks for removing "if this function returns an error, throw it", this pattern is horrible and should never be used. we should always throw errors.

the segfault was a null pointer dereference on the removed `.getObject()`. this will assert in debug mode that the cell is non-zero (0 is classified as a cell type, aka a pointer)

i think we could get away with also removing `if (UNLIKELY(scope.exception()))` too, since if there is an exception thrown, the jsvalue return is ignored, which is where we use zero. crashes due to following this rule are indications of other mistakes in our code.

src/bun.js/node/types.zig
```diff
@@ -2021,7 +2030,7 @@ pub const Path = struct {
     }
 
     pub fn getCwdWindowsU8(buf: []u8) MaybeBuf(u8) {
-        const u16Buf: [PATH_MAX_WIDE]u16 = undefined;
+        const u16Buf: [PathSize(u16).MAX]u16 = undefined;
```
gvilums: Couldn't this just be `bun.WPathBuffer`?

src/bun.js/node/types.zig
```diff
@@ -2301,14 +2310,16 @@ pub const Path = struct {
         if (suffix_ptr) |_suffix_ptr| {
             // Supress exeption in zig. It does globalThis.vm().throwError() in JS land.
             validateString(globalObject, _suffix_ptr, "ext", .{}) catch {
-                return JSC.JSValue.jsUndefined();
+                // Safe to return .zero here without fear of segfault because the result in C
+                // is not used when JSC::ThrowScope scope.exception() returns an Exception pointer.
```
paperclover: this comment is incorrect. to throw javascript exceptions, the exception pointer is set on the vm, with a zero js value returned. when a crash occurs it is a bug in the code using the js value, not checking for an exception first.

in zig this abstraction for this isnt great, especially considering we have an error system in zig but it is not used at all for js exceptions. i want to change this abstraction later

src/bun.js/node/types.zig
```diff
@@ -2848,20 +2865,35 @@ pub const Path = struct {
...
+        if (bufLen > size.MAX) {
             const buf = allocator.alloc(u8, bufLen) catch bun.outOfMemory();
             defer allocator.free(buf);
             return if (isWindows)
                 toJSString(globalObject, formatWindow(pathObject, buf))
             else
                 toJSString(globalObject, formatPosix(pathObject, buf));
         }
-        var buf: PathBuffer = undefined;
+        const buf = switch (bufLen) {
```
paperclover: since bufLen is not known at compile time, all 3 branches are present in the binary, meaning all stack allocations in there are present. so all three of these buffers are present

src/bun.js/node/types.zig
```diff
@@ -2848,20 +2865,35 @@ pub const Path = struct {
...
+            size.MID => brk: {
+                var _buf: [size.MID]u8 = undefined;
+                break :brk &_buf;
+            },
+            size.MIN => brk: {
+                var _buf: [size.MIN]u8 = undefined;
+                break :brk &_buf;
+            },
+            else => unreachable,
+        };
```
gvilums: This is undefined behavior: we are returning a pointer to a buffer that only exists in an inner scope. Unfortunately there is no easy fix for this that keeps the behavior - either just use the `size.MAX` buffer or use a heap fallback with the smaller buffer (I suggest just using the larger buffer).

src/bun.js/node/types.zig
```diff
@@ -3125,23 +3161,52 @@ pub const Path = struct {
...
```
gvilums: See above

src/bun.js/node/types.zig
```diff
@@ -2848,20 +2865,35 @@ pub const Path = struct {
...
             return if (isWindows)
                 toJSString(globalObject, formatWindow(pathObject, buf))
             else
                 toJSString(globalObject, formatPosix(pathObject, buf));
         }
-        var buf: PathBuffer = undefined;
+        const buf = switch (bufLen) {
+            size.MAX => brk: {
+                var _buf: [size.MAX]u8 = undefined;
+                break :brk &_buf;
```
paperclover: i'm not sure if this is actually is an issue but i get weary of keeping a pointer stack memory in a block. you should think of the `var` above and any pointers into it as being invalid once the block containing it ends.

but this whole part is invalid for the other reason

src/bun.js/bindings/Path.cpp
gvilums: Could you elaborate on this change? How are exceptions thrown now? Do we always throw them from the `ZigFunction`?

src/bun.js/node/types.zig
```diff
@@ -3565,25 +3604,40 @@ pub const Path = struct {
...
+            size.MID => brk: {
+                var _buf: [size.MID]u8 = undefined;
+                break :brk &_buf;
+            },
+            size.MIN => brk: {
+                var _buf: [size.MIN]u8 = undefined;
+                break :brk &_buf;
+            },
+            else => unreachable,
+        };
```
gvilums: same issue

gvilums (dismissed):

paperclover (dismissed):
the switch with the runtime-known value is wrong and will use more stack memory.

imo we should reduce the max path length for these functions, i think to 4096 chars of data.
this will cover most paths, and then paths that are longer (on windows this would include the valid paths under 32k, but also the ones over that, and on mac and linux the ones longer than the 4096 of course)

it would be doubly good if this was made into an abstraction we could use across the codebase, and stop relying on `PathBuffer`s everywhere, but i think that can defer until stack space is actually a concern, it does not seem to cause issues using `64kb` buffers all over the place.

src/bun.js/bindings/Path.cpp
jdalton: Ya, the `nullptr` handled via `getObject()` was the culprit. I went back to the return `.zero` pattern but IF we want safe code that doesn't rely on making sure we don't touch it a certain way when returned we could always return `JSC.JSValue.jsUndefined()` which is a valid JSValue result (or at least one that other JSValue methods like).

src/bun.js/node/types.zig
jdalton: The unreachable should turn into a segfault if hit. The code though should never be hit as the bufLen should only be one of those 3 values (`switch (bufLen) {` is only ever MAX,MID,or MIN). If unreachable is scary I could make the MIN and else the same.

src/bun.js/node/types.zig
jdalton: I think your explanation and mine are going in the same direction. 
https://github.com/oven-sh/bun/pull/8932/files#diff-d8f1d097a0cb6d63f73dcef0c76498e64dbd23cccef9530983a1369a95cb4aecR45-R48

The code is safe because we aren't doing things with the `nullptr` like `getObject` and are bailing in the `if (UNLIKELY(scope.exception()))`

I can reference the `nullprt` issue though. Really we should return safe values even when erroring. The comment can be removed when we change the `return .zero` pattern.

src/bun.js/node/types.zig
jdalton: Oh that's super lame. Will fork it.

src/bun.js/node/types.zig
jdalton: I don't think this is an issue but the one above it is totally a biggie.

src/bun.js/bindings/Path.cpp
jdalton: The error object was never actually being returned from zig. So the:
```
if (UNLIKELY(obj != nullptr && obj->isErrorInstance())) { 
  scope.throwException(globalObject, obj); 
  return JSC::JSValue::encode(JSC::jsUndefined()); 
} 
```
was dead code ✂️ . In zig we use the `validateString` helper which will call:
```
const err = JSC.toTypeError(JSC.Node.ErrorCode.ERR_INVALID_ARG_TYPE, fmt, args, globalThis);
globalThis.vm().throwError(globalThis, err);
```
so the exception is already thrown in JS land. We're just trying to leave the function.

src/bun.js/node/types.zig
jdalton: It could be yes. In the code though I sided with consistency of `PathSize()` in places. So instead of thinking "Oh that's a u16 wide path, and Oh we have a special type alias for the size of `windows.PATH_MAX_WIDE`" I went with "Oh CWD uses max path size".  Since size is now a sliding range and may vary. ~~I did miss it being comptimed though~~.  Turns out it was already comptime 💪

jdalton: Turns out it was already comptime 💪

jdalton: Uses bun.WPathBuffer now.

src/bun.js/node/types.zig
jdalton: No more unreachable.

src/bun.js/node/types.zig
jdalton: Reduced the comment and removed `if (UNLIKELY(scope.exception()))`

src/bun.js/node/types.zig
jdalton: > so all three of these buffers are present

Fixed this.

src/bun.js/bindings/Path.cpp
gvilums: that makes sense 👍

jdalton:
@paperdave 
> the switch with the runtime-known value is wrong and will use more stack memory.
> 
> imo we should reduce the max path length for these functions, i think to 4096 chars of data. this will cover most paths, and then paths that are longer (on windows this would include the valid paths under 32k, but also the ones over that, and on mac and linux the ones longer than the 4096 of course)
> 
> it would be doubly good if this was made into an abstraction we could use across the codebase, and stop relying on `PathBuffer`s everywhere, but i think that can defer until stack space is actually a concern, it does not seem to cause issues using `64kb` buffers all over the place.

Fixed.

src/bun.js/node/types.zig
```diff
@@ -2848,20 +2858,14 @@ pub const Path = struct {
...
-        else
-            toJSString(globalObject, formatPosix(pathObject, &buf));
+        return switch (getPathLenT(u8, lenSum)) {
+            bun.MAX_PATH_BYTES => if (isWindows) formatWindowsL(globalObject, pathObject, bun.MAX_PATH_BYTES) else formatPosixL(globalObject, pathObject, bun.MAX_PATH_BYTES),
+            else => |bufLen| brk: {
+                const buf = allocator.alloc(u8, bufLen) catch bun.outOfMemory();
+                defer allocator.free(buf);
+                break :brk if (isWindows) formatWindows(globalObject, pathObject, buf) else formatPosix(globalObject, pathObject, buf);
+            },
+        };
```
gvilums: I don't quite understand the point of this logic here. `getPathLen` returns `size.MIN` for lengths less than `size.MIN`, `size.MAX` for lengths between min and max, and the length itself for larger values. So specifically for lengths that fall between `size.MIN` and `size.MAX` we use a stack allocation (through the `bun.MAX_PATH_BYTES` code path), and otherwise we allocate? I don't quite understand what that achieves versus what we had before (stack buffer with heap fallback).

gvilums (commented):
Changes look good overall, but I still don't understand the logic around `getPathLen`

src/bun.js/node/types.zig
jdalton: @gvilums Sure thing. Clarifying things helps me double check the reasoning too! 

The `getPathLenT(comptime T: type, lenSum: usize) usize` allows for a sliding range of lengths. Initially I had 3, a min, mid, and max. 
```
if (lenSum > size.MID) return @max(lenSum, size.MAX);
if (lenSum > size.MIN) return @max(lenSum, size.MID);
return @max(lenSum, size.MIN);
```

But to keep things simple until necessary on revision went with min and max. The simplified logic is that for u16 we use 4 KB as a path and then for larger allocate (as we'll need to allocate for the larger strings beyond PATH_MAX_WIDE.

This allows us to establish a pattern/abstraction one where we can switch on MIN, MIN, or however we choose to dice it.
```
return switch (getPathLenT(T, lenSum)) {
    size.MIN => if (isWindows) fooWindowsL(T, globalObject, arg1, size.MIN) else fooPosixL(T, globalObject, arg1, size.MIN),
    size.MID => if (isWindows) fooWindowsL(T, globalObject, arg1, size.MID) else fooPosixL(T, globalObject, arg1, size.MID),
    size.MAX => if (isWindows) fooWindowsL(T, globalObject, arg1, size.MID) else fooPosixL(T, globalObject, arg1, size.MAX),
    else => |bufLen| brk: {
        const buf = allocator.alloc(T, bufLen) catch bun.outOfMemory();
        defer allocator.free(buf);
        break :brk if (isWindows) fooWindows(T, globalObject, arg1, buf) else fooPosix(T, globalObject, arg1, buf);
    },
};
  ```

For u8 we have just 1 value specified for min/max or otherwise and that is bun.MAX_PATH_BYTES. So for u8 the len is bun.MAX_PATH_BYTES unless its larger. So this pattern that I'm establishing can hold for it and u16 using the same kind of code subing things out with a `T` when we move the code. Using the `size.MIN` and `size.MAX` we don't care if it's u8 or u16, the same code works for it. Using the variation of the methods `fooWindowsL` and `fooPosixL` allows for methods which stack allocate only (previously as dave mentioned with a switch statement bufs within the comptime stack allocations are made even if the branch is not hit). By separating the stack variation out we allow for it to be dynamic-ish via MID,MIN,MAX, or whatever sliding range, still comptime'd, and allow for using the heap when things get larger than max.

Comment (jdalton) [reply src/bun.js/node/types.zig]: Updated to better show generic-ness

Comment (gvilums) [reply src/bun.js/node/types.zig]: That mostly makes sense to me now. However, in my opinion all of this complexity isn't really warranted. Why wouldn't we just use the stack fallback allocator with a smaller stack buffer? I'm not even sure that it would really provide a measurable performance benefit anyway, because very long path strings rarely occur.

I think the other changes in this PR are very good, but we should simply use a stack fallback allocator approach to allocating the buffers and save ourselves the headache.

Sorry for the delayed response, I'm on vacation right now 😅

Comment (jdalton) [reply src/bun.js/node/types.zig]: @gvilums Done! Now it just has the default smaller size or allocates for larger. Still supporting u8 and u16.

**Comment by jdalton** [comment]
#### 📰 Related path news:

Zig now supports unpaired surrogates in Windows paths https://github.com/ziglang/zig/pull/19005!
It means Bun will need to change `error.InvalidUtf8` to `error.InvalidWtf8`. There's also added methods
* `std.fs.path.fmtAsUtf8Lossy/fmtWtf16LeAsUtf8Lossy`
* `std.unicode.fmtUtf8`

Tracked https://github.com/oven-sh/bun/issues/9122

File: src/bun.js/node/types.zig [hunk 11]
```diff
@@ -2795,27 +2806,54 @@ pub const Path = struct {
...
+            (if (dirLen > 0) dirLen else pathObject.root.len) +
+            (if (baseLen > 0) baseLen else pathObject.name.len + pathObject.ext.len);
+        const size = comptime PathSize(T);
+        return switch (getPathLenT(T, lenSum)) {
+            size.DEFAULT => if (isWindows) formatWindowsJS_L(T, globalObject, pathObject, size.DEFAULT) else formatPosixJS_L(T, globalObject, pathObject, size.DEFAULT),
+            else => |bufLen| brk: {
+                const buf = allocator.alloc(T, bufLen) catch bun.outOfMemory();
+                defer allocator.free(buf);
+                break :brk if (isWindows) formatWindowsJS_T(T, globalObject, pathObject, buf) else formatPosixJS_T(T, globalObject, pathObject, buf);
+            },
```
gvilums: I hope I don't seem pedantic here, but I don't quite see the point of the switch statement. My understanding of the logic:
1. `format` creates a `stackFallback`-allocator and passes it to `formatJS_T`
2. `formatJS_T`computes the `lenSum`. If `lenSum` fits in `size.DEFAULT`, we do the computation on the stack using `formatXJS_L` with a buffer size of `size.DEFAULT`. If it doesn't, we use the allocator to allocate buffers dynamically and then call `formatXJS_T`

My issue is the following: Why are we going through the effort of checking if the size fits in `size.DEFAULT` at all and allocating a separate stack buffer, if we already have the `stackFallback`. I propose we get rid of the `switch` entirely and replace it with the `else` branch.

Same logic applies to all other places where we switch on `getPathLenT`. In general, I think we should just get rid of `getPathLenT` entirely.

gvilums (changes requested):
See my comment that applies to all uses of `getPathLenT`

src/bun.js/node/types.zig
jdalton: Changed the switch to else. It was a hold-over from allowing things to be a sliding range. Kinda pretty but def not needed. The `getPathLenT` does allow for not needing to know the logic of distilling a path buffer length each time, currently just `@max(lenSum, PathSize(T).DEFAULT)`, but still. It is also inlined.

src/bun.js/node/types.zig
```diff
@@ -2795,27 +2806,53 @@ pub const Path = struct {
...
+        // Add one for the possible separator.
+        const lenSum: usize = 1 +
+            (if (dirLen > 0) dirLen else pathObject.root.len) +
+            (if (baseLen > 0) baseLen else pathObject.name.len + pathObject.ext.len);
+        const size = comptime PathSize(T);
+        const bufLen = getPathLenT(T, lenSum);
+        if (bufLen == size.DEFAULT) {
+            return if (isWindows) formatWindowsJS_L(T, globalObject, pathObject, size.DEFAULT) else formatPosixJS_L(T, globalObject, pathObject, size.DEFAULT);
+        }
+        const buf = allocator.alloc(T, bufLen) catch bun.outOfMemory();
```
gvilums: While this is already much simpler than the complex switch statement we had previously, the logic here can still be simplified much further in my opinion.

Currently, what we're doing is we're checking if `bufLen` fits in a buffer of `size.DEFAULT`. If so, we take a path where we have a stack-allocated buffer (through `format*JS_L`). Otherwise we use our allocator to allocate the memory and take that path instead.

But we jave already created a stack fallback allocator (passed through `allocator`) - there's no reason we couldn't just use that instead! In fact, the current version actually wastes stack space in trying to be clever - both the stack fallback allocator created in `format` as well as the stack buffers created in `format*JS_L` reside on the stack at the same time, even when the stack fallback allocator is not used.

My suggestion is to change the code to the following:
```suggestion
        const buf = allocator.alloc(T, lenSum) catch bun.outOfMemory();
```
This would also allow us to get rid of `getPathLen`, `size.DEFAULT`, and the `*JS_L` functions for free.
(same applies to all the other places in the code where this pattern is used)

If you have other opinions on this we should probably hop on a call, I feel like the back and forth of github reviews isn't the most efficient way to do this

gvilums (changes requested):

src/bun.js/bindings/Path.cpp
```diff
@@ -29,31 +29,21 @@ using namespace JSC;
...
-    ); \
-    JSC::JSObject *obj = result.getObject(); \
-    if (UNLIKELY(obj != nullptr && obj->isErrorInstance())) { \
-        scope.throwException(globalObject, obj); \
-        return JSC::JSValue::encode(JSC::jsUndefined()); \
     } \
-    if (UNLIKELY(scope.exception())) \
-        return JSC::JSValue::encode(JSC::jsUndefined()); \
-    return JSC::JSValue::encode(result);
+    auto isWindows = thisObject->get(globalObject, WebCore::clientData(vm)->builtinNames().isWindowsPrivateName()); \
```
Jarred-Sumner, line 45: It was a mistake in my initial implementation to use a private property for this. It should be an internal property on the C++ object or a separate class / classInfo. I didn’t know how to add extra fields to C++ classes without causing allocation failure because I didn’t know about isosubspaces. Not worth fixing right now though

src/bun.js/node/types.zig
```diff
@@ -2301,43 +2312,35 @@ pub const Path = struct {
...
+                suffixZSlice = suffixZStr.toSlice(allocator);
             }
         }
-
-        return if (isWindows)
-            toJSString(globalObject, basenameWindows(pathZSlice.slice(), null))
-        else
-            toJSString(globalObject, basenamePosix(pathZSlice.slice(), null));
+        defer if (suffixZSlice) |_s| _s.deinit();
+        return basenameJS_T(u8, globalObject, isWindows, pathZSlice.slice(), if (suffixZSlice) |_s| _s.slice() else null);
```
Jarred-Sumner, line 2340: Since we have the u8 and u16 versions, couldn’t we use the latin1 or utf16 string pointer directly and then not allocate in the first place?

gvilums: I think we should defer this to a follow-up PR at a later date

jdalton: @Jarred-Sumner Yes! This has come up before in review as a future todo too. We currently aren't using the u16 support that this PR makes way for so I imagine as we step into use things will get further refined.

src/bun.js/node/types.zig
jdalton: True! Nice! (Thanks for patience on that).

jdalton: Fixed. Had to keep some min/max logic to make unit tests happy (no segfaults) since some calculated len estimates were too small to work with within the methods.

gvilums (approved):

PR #8935, opened 2024-02-16
windows: make test/js/web/websocket/websocket-client.test.ts pass

nektro:


test/js/web/websocket/websocket-client.test.ts
```diff
@@ -260,7 +260,7 @@ function test(label: string, fn: (ws: WebSocket, done: (err?: unknown) => void)
 }
 
 async function listen(): Promise<URL> {
-  const { pathname } = new URL("./websocket-server-echo.mjs", import.meta.url);
+  const pathname = path.join(import.meta.dir, "./websocket-server-echo.mjs");
```
jdalton, line 263: ☝️ Is this change working around another bug (like with new URL and import.meta.url on Windows?)

nektro: pulling this out into a `test.js` the previous implementation prints `/C:/bun/websocket-server-echo.mjs`

nektro: and `import.meta.url` is `file:///C:/bun/test.js`

nektro: the old code only happens to work on posix because it uses `/` path separators and has no disk discriminators

Jarred-Sumner: In other words, this should've used `Bun.fileURLToPath` instead of `.pathname`

Jarred-Sumner (approved):

PR #8936, opened 2024-02-16
Add fs.exists[util.promisify.custom]

dmitri-gb:
### What does this PR do?

This adds the missing implementation for `fs.exists[util.promisify.custom]`.

`fs.exists` doesn't follow the error-first-callback convention, so it needs a custom implementation for `util.promisify`. Without it, a promisifed `exists` will always reject (with `true` or `false` as the reason).

[Prisma CLI](https://github.com/prisma/prisma/blob/961906b6ebadc8d0e548d90e8b7858dfd0481fb8/packages/internals/src/cli/getSchema.ts#L9) relies on this.

- [x] Code changes

I wrote automated tests.


paperclover (approved):

src/js/node/fs.js
```diff
@@ -365,6 +365,14 @@ var access = function access(...args) {
...
 var kCustomPromisifiedSymbol = Symbol.for("nodejs.util.promisify.custom");
 
+exists[kCustomPromisifiedSymbol] = async function (path) {
+  try {
+    return await fs.exists(path);
+  } catch (e) {
+    return false;
+  }
+};
+
```
jdalton, line 369: Relavant Node.js code:
https://github.com/nodejs/node/blob/53aed8838cb71142ca1591b655000aac90d91205/lib/fs.js#L276-L281
```js
ObjectDefineProperty(exists, kCustomPromisifiedSymbol, {
  __proto__: null,
  value: function exists(path) { // eslint-disable-line func-name-matching
    return new Promise((resolve) => fs.exists(path, resolve));
  },
});
```
>💡 Might be implemented this way in Node because of its [older perf history with `await`](https://v8.dev/blog/fast-async). 
>Of interest is @Jarred-Sumner's [twitter thread on JSC await perf improvements](https://twitter.com/jarredsumner/status/1561322898588983296).

Node's doesn't suppress the in the promise form because it's already suppressing errors in `fs.exists` (as do we):
https://github.com/nodejs/node/blob/53aed8838cb71142ca1591b655000aac90d91205/lib/fs.js#L265-L273

dmitri-gb: @jdalton are you saying I should remove the try-catch? I only added it because I'm not really sure about the semantics of `Bun.fs().exists` (whether it ever throws/rejects etc) and also because the callback variant defintion [adds a rejection handler](https://github.com/oven-sh/bun/blob/c457699b0af1dec4edee1a22d229c8d722283f1a/src/js/node/fs.js#L155):
```js
fs.exists.$apply(fs, [path]).then(
  existed => callback(existed),
  _ => callback(false),
);
```

jdalton: That bun is handling the rejection and returning `false` aligns with expectations. The try-catch it's wrapped in helps for any synchronous issues so both(sync/async) are covered. No need to wrap it in additional try-catches. I'd just tweak to do as node does. It's a single line and keeps things simple:
```
new Promise((resolve) => fs.exists(path, resolve))
```
in a non async function.

jdalton (approved):

PR #8938, opened 2024-02-16
windows: pass more tests

nektro:


src/bun.zig
```diff
@@ -476,7 +476,24 @@ pub fn isReadable(fd: FileDescriptor) PollFlag {
 pub const PollFlag = enum { ready, not_ready, hup };
 pub fn isWritable(fd: FileDescriptor) PollFlag {
     if (comptime Environment.isWindows) {
-        @panic("TODO on Windows");
+        var polls = [_]std.os.windows.ws2_32.WSAPOLLFD{
```
paperclover, line 479: this is awesome to see, because i did not think this was even possible beforehand.

src/windows.zig
```diff
@@ -3081,3 +3087,63 @@ pub extern "kernel32" fn OpenProcess(
...
+            if (std.os.windows.ws2_32.TransmitFile(wsocket, file_handle, 0, 64 * 1024, overlapped_ptr, null, 0) == 0) {
+                const err = Win32Error.get();
+                if (err != Win32Error.WSA_IO_PENDING) {
+                    this.status = .fail;
+                    return .{ .err = bun.sys.Error.fromCodeInt(err.int(), .sendfile) };
+                }
+            }
+        }
+
+        var sended: u32 = 0;
```
paperclover: the past tense of "send" is "sent"

paperclover (approved):
nice

test/js/web/streams/streams.test.js
```diff
@@ -419,7 +421,7 @@ it("ReadableStream.prototype.values", async () => {
   expect(chunks.join("")).toBe("helloworld");
 });
 
-it("Bun.file() read text from pipe", async () => {
+it.skipIf(isWindows)("Bun.file() read text from pipe", async () => {
```
Jarred-Sumner, line 424: Windows supports named pipes. We shouldn't skip this test.

test/js/workerd/html-rewriter.test.js
```diff
@@ -548,7 +547,7 @@ const request_types = ["/", "/gzip", "/chunked/gzip", "/chunked", "/file", "/fil
     it(`works with ${protocol} fetch using ${url}`, async () => {
       const server = protocol === "http" ? http_server : https_server;
       const server_url = server.url;
-      const res = await fetch(`${server_url}${url}`, { tls: { rejectUnauthorized: false } });
+      const res = await fetch(`${server_url}${url.slice(1)}`, { tls: { rejectUnauthorized: false } });
```
Jarred-Sumner: Is the URL normalization different on Windows versus Posix? I'm confused.

test/cli/run/as-node.test.ts
```diff
@@ -90,11 +89,11 @@ describe("fake node cli", () => {
 
   test("process args work", () => {
     const temp = tempDirWithFiles("fake-node", {
-      "index.js": "console.log(JSON.stringify(process.argv))",
+      "index.js": "console.log(JSON.stringify(process.argv.slice(1)))",
```
Jarred-Sumner, line 92: we do need to test that process.argv[0] is as expected

Jarred-Sumner (changes requested):
Some comments

test/js/workerd/html-rewriter.test.js
nektro: `server.url` makes a string like `http://localhost:8000/`
`url` is a string like `/chunked`
so the url that showed up in the logs was `http://localhost:8000//chunked` with a double slash

test/cli/run/as-node.test.ts
nektro: I included in the commit message why this was done; should i make a separate version of the test for windows+posix?

test/js/workerd/html-rewriter.test.js
cirospaciari: should we just use `server.url.origin` instead of `server.url`? (server.url is a `URL` object)

nektro: that would also work great, I can go with that.

test/js/node/dns/node-dns.test.js
```diff
@@ -230,7 +231,7 @@ test("dns.lookup (localhost)", done => {
   });
 });
 
-test("dns.getServers", done => {
+test.skipIf(isWindows)("dns.getServers", done => {
   function parseResolvConf() {
```
jdalton, line 235: ☝️ We should put a TODO: comment here to get back to it later.

Side note: We should have a `test.todoIf()` like we have a `test.skipIf()`.

test/js/node/stream/node-stream.test.js
```diff
@@ -336,7 +337,7 @@ describe("TTY", () => {
 });
 `;
 
-it("TTY streams", () => {
+it.skipIf(isWindows)("TTY streams", () => {
   mkdirSync(join(tmpdir(), "tty-test"), { recursive: true });
```
jdalton, line 341: ☝️ TODO: comment to get back to it please (unless it's just not a windows thing)

CosmeValera:
You're doing a good job guys

PR #8941, opened 2024-02-16
Support async generator functions in `Response` and `Request` for bodies

Jarred-Sumner:
### What does this PR do?

This makes the following code work:
```js
await new Response(
  async function* stream() {
    yield "hey";
    await Bun.sleep(42);
    yield Buffer.alloc(100);
  }
).arrayBuffer();
```

```js
await new Response({
  async *[Symbol.asyncIterator]() {
    yield "hey";
    await Bun.sleep(42);
    yield Buffer.alloc(100);
  },
}).arrayBuffer();
```

There are tests. The ones for .json() fail. Need to fix that before we can merge

### DIFF
--- a/test/js/bun/http/async-iterator-stream.test.ts
+++ b/test/js/bun/http/async-iterator-stream.test.ts
@@ -1,0 +1,195 @@
+import { describe, expect, test } from "bun:test";
+import { bunExe, bunEnv } from "harness";
+import path from "path";
+
+describe("Streaming body via", () => {
+  test("async generator function", async () => {
+    const server = Bun.serve({
+      port: 0,
+
+      async fetch(req) {
+        return new Response(async function* yo() {
+          yield "Hello, ";
+          await Bun.sleep(30);
+          yield Buffer.from("world!");
+          return "!";
+        });
+      },
+    });
+
+    const res = await fetch(`${server.url}/`);
+    const chunks = [];
+    for await (const chunk of res.body) {
+      chunks.push(chunk);
+    }
+
+    expect(Buffer.concat(chunks).toString()).toBe("Hello, world!!");
+    expect(chunks).toHaveLength(2);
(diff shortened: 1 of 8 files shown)


paperclover (commented):
did not read the test. looks good. i'm gonna give this a go on astro before marking this as approved

test/js/bun/http/async-iterator-stream.test.ts
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  test("async generator function throws an error but continues to send the headers", async () => {
+    const server = Bun.serve({
+      port: 0,
+
+      async fetch(req) {
+        return new Response(
+          async function* () {
+            throw new Error("Oops");
```
Electroid, line 38: Can we either move this test to a subprocess or add a special property that hides the errors from being printed? This makes our GitHub CI spammy with errors

test/js/bun/http/async-iterator-stream.test.ts
```diff
@@ -0,0 +1,195 @@
...
+          {
+            headers: {
+              "X-Hey": "123",
+            },
+          },
+        );
+      },
+    });
+
+    const res = await fetch(`${server.url}/`);
```
Electroid:
```suggestion
    const res = await fetch(server.url);
```

test/js/bun/http/async-iterator-stream.test.ts
Jarred-Sumner: That should be fixed in the GitHub actions printer IMO

paperclover (approved):
nice

test/js/bun/http/async-iterator-stream.test.ts
jdalton: Related to https://github.com/oven-sh/bun/pull/8949

PR #8943, opened 2024-02-16
fix(install): lifecycle script changes

### LINKED ISSUE #8899: bun install doesn't support `file:.` path in `package.json` dependencies

1.0.26+c75e768a6

Linux 6.2.0-39-generic x86_64 x86_64

In a directory with a valid `package.json` file (or if the `.` path in the command is changed to point to ~such a directory~ EDIT: the current package root)

`bun install --save-dev .`
fails.

`npm install --save-dev .`
creates a line in `package.json` and a symlink under `node_modules` that's useful for local development.
With that line in place,
`bun install`
fails.

`bun install` should create a symlink under `node_modules` pointing to the path given as a `file` url in `package.json` dependencies. It should also be able to add such a line when told to install from a local directory containing a `package.json` file, picking the dependency name from that file.

### What do you see instead?

```
bun install v1.0.26 (c75e768a)

error: MissingPackageJSON
Bun could not find a package.json file.
```

### Additional information

_No response_

dylan-conway:
### What does this PR do?

- Breaking change: if `trustedDependencies` exists in package.json, bun install will no longer pull from the default list and will only run scripts from the `trustedDependencies` list. If the list is empty, then no scripts will run. The default trusted dependencies list is only used when `trustedDependencies` does not exist in package.json.

- `--trust`. This flags tells `bun add` to automatically trust the dependency and all it's dependencies, running their lifecycle scripts and adding to `trustedDependencies` in package.json.

- warning for skipped lifecycle scripts
Screenshot 2024-03-06 at 11 26 43 PM

- bugfix for linking root package to itself, fixes #8899 

---

Also included in this pr is `bun pm trusted`. This command without any args will print the current trusted and untrusted dependencies with scripts:
Screenshot 2024-03-07 at 2 25 56 AM

If you provide package names, they will be added to `trustedDependencies` and their scripts will run. In the example above, running `bun pm trust es5-ext` will add this to your package.json:
```
...
trustedDependencies: [
    "es5-ext"
]
...
```

You can also pass `--all` to trust all available untrusted dependencies, and `--default` to list the default trusted dependencies list.

The breaking changes in this pr will be disabled until `v1.1.0`.

new and existing tests


src/install/install.zig
```diff
@@ -3553,6 +3563,134 @@ pub const PackageManager = struct {
...
+            const slice = manager.lockfile.packages.slice();
+            for (slice.items(.meta)) |*meta| {
+                // these are possibly updated later, but need to make sure non are zero
+                meta.setHasInstallScript(false);
+            }
+        }
+
+        return;
+
+        // if (load_lockfile_result == .ok and load_lockfile_result.ok.serializer_result.packages_need_update) {
```
Jarred-Sumner: did you mean to commit this

paperclover: bump

src/install/install.zig
```diff
@@ -9603,6 +10291,16 @@ pub const PackageManager = struct {
         Output.flush();
     }
 
+    fn printSkippedScripts(summary: PackageInstall.Summary) void {
+        const count = summary.packages_with_skipped_scripts_set.count();
+        if (count > 0) {
+            Output.pretty("\n\n<d> Skipped ~{d} script{s}. Run `bun pm trusted` for details.", .{
```
paperclover: is this supposed to be prettyln?

src/js_ast.zig
```diff
@@ -1458,6 +1458,23 @@ pub const E = struct {
 
             return array;
         }
+
+        /// debug.assert the array only contains strings before calling
```
paperclover: can you make this an actual debug assert for this function. or uh, i see you do a runtime check for that... so is it an assertion or? i'd like osmething better than this comment

src/install/lockfile.zig
```diff
@@ -4553,18 +4722,42 @@ pub const Package = extern struct {
 
         man_dir: String = String{},
         integrity: Integrity = Integrity{},
-        _padding_integrity: [3]u8 = .{0} ** 3,
+
+        // 0 - package from old lockfile, needs update
+        // 1 - does not have install scripts
+        // 2 - has install scripts
+        __has_install_script: u8,
```
paperclover: instead of a magic number, can this be an `enum(u8)`. it would also be nice to add this into the comment on line 4709 `when we bump the lockfile version, we should reorder this to:`

also proper documentation comments are `///` but that is less relevant.

src/js_ast.zig
dylan-conway: Moved the debug assert inside this function. Also did the same for `alphabetizeProperties`. Originally I had the runtime check for extra safety

src/install/install.zig
dylan-conway: updated

src/install/lockfile.zig
dylan-conway: switched to enum

test/cli/install/registry/bun-install-registry.test.ts
```diff
@@ -2987,9 +3014,781 @@ for (const forceWaiterThread of [false, true]) {
...
+            trustedDependencies: ["electron"],
+            dependencies: {
+              "electron": "1.0.0",
+            },
+          }),
+        );
+
+        await rm(join(packageDir, "node_modules", "electron", "preinstall.txt"), { force: true });
+
+        // lockfile should save evenn though there are no changes to trustedDependencies due to
```
Jarred-Sumner, line 3657:
```suggestion
        // lockfile should save even though there are no changes to trustedDependencies due to
```

src/install/install.zig
```diff
@@ -8139,7 +8527,13 @@ pub const PackageManager = struct {
                         installer.cache_dir = directory;
                     }
                 },
-                else => return,
+                else => {
+                    if (comptime Environment.allow_assert) {
+                        @panic("bad");
```
Jarred-Sumner, line 8532: bad

src/install/install.zig
```diff
@@ -8204,14 +8598,37 @@ pub const PackageManager = struct {
...
+
+                                    if (this.lockfile.trusted_dependencies == null) this.lockfile.trusted_dependencies = .{};
+                                    this.lockfile.trusted_dependencies.?.put(this.manager.allocator, name_hash, {}) catch bun.outOfMemory();
+                                }
+                            }
+                        }
+
+                        if (resolution.tag != .workspace and !is_trusted and this.lockfile.packages.get(package_id).meta.hasInstallScript()) {
+                            if (comptime log_level.isVerbose()) {
+                                Output.prettyError("Skipped scripts for: {s}@{}\n", .{ alias, resolution.fmt(this.lockfile.buffers.string_bytes.items) });
```
```suggestion
                                Output.prettyError("Blocked scripts for: {s}@{}\n", .{ alias, resolution.fmt(this.lockfile.buffers.string_bytes.items) });
```

Jarred-Sumner (approved):

PR #8944, opened 2024-02-16
fix(sqlite): enable math functions

### LINKED ISSUE #4772: bun:sqlite missing math functions

Linux 5.15.0-67-generic x86_64 x86_64

You can reproduce this problem by running the following file and observing the error. This does not happen with the similarly benchmarked tool `better-sqlite3`

```js
const { Database } = require('bun:sqlite');
// or for testing with better-sqlite3
// const Database = require('better-sqlite3');

db.exec(`
    CREATE TABLE IF NOT EXISTS readings (
        value TEXT NOT NULL
    );
    INSERT INTO readings values 
    (1),
    (2),
    (3),
    (4),
    (5),
    (6),
    (7),
    (8),
    (9),
    (10)
`);

const selectDataStmt = db.prepare(`
    SELECT 
        FLOOR(AVG(value)) as value
    FROM 
        readings
`);
const rows = selectDataStmt.all();

console.log(rows);
```

output should be 

```
[ { value: 5 } ]
```

### What do you see instead?

Observe error output
```txt
$ bun x.js 
185 |     }
186 |     var arg0 = params[0];
187 |     return !isArray(arg0) && (!arg0 || typeof arg0 !== "object" || isTypedArray(arg0)) ? SQL.run(this.#handle, query, params) : SQL.run(this.#handle, query, ...params);
188 |   }
189 | 
190 |   prepare(query, params, flags) {
                     ^
error: no such function: FLOOR
      at prepare (bun:sqlite:190:18)
      at query (bun:sqlite:197:14)
      at /var/www/x.js:22:27
      at globalThis (/var/www/x.js:29:20)
```

### Additional information

`SQLITE_ENABLE_MATH_FUNCTIONS` is not defined in https://github.com/oven-sh/bun/blob/682406c42ceee590c2b90900175f0eb1eaed5637/src/js/bun/sqlite.ts or in https://github.com/oven-sh/bun/raw/main/src/bun.js/bindings/sqlite/sqlite3.c. I believe that this is the root cause of the problem.

ZTL-UwU:
### What does this PR do?

This enables math functions for sqlite by enabling the `-DSQLITE_ENABLE_MATH_FUNCTIONS` flag in the Makefile.

Resolves #4772 

- [x] Code changes


gvilums (approved):

jdalton (approved):
Thank you @ZTL-UwU! Unit tests FTW too!

PR #8945, opened 2024-02-16
Make shell errors slightly better

Jarred-Sumner:
### What does this PR do?

@zackradisic feel free to take over from here


src/js/builtins/shell.ts
```diff
@@ -2,25 +2,34 @@ type ShellInterpreter = any;
...
+      super(`Failed with exit code ${code}`);
       this.#output = output;
-    }
+      this.name = "ShellError";
 
-    get exitCode() {
-      return this.#output.exitCode;
-    }
+      // Maybe we should just print all the properties on the Error instance
+      // instead of speical ones
```
jdalton, line 16: How does Node handle it?

PR #8949, opened 2024-02-16
Do not log SuppressedError to GitHub CI

Electroid:
### What does this PR do?

* Prevents `SuppressedError` from being annotated in GitHub CI

- [x] Code changes

Automated tests.

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -2764,8 +2764,10 @@ pub const VirtualMachine = struct {
         this.had_errors = true;
         defer this.had_errors = prev_had_errors;
 
-        if (allow_side_effects and Output.is_github_action) {
-            defer printGithubAnnotation(exception);
+        if (comptime allow_side_effects) {
+            if (Output.is_github_action) {
+                defer printGithubAnnotation(exception);
+            }
         }
 
         const line_numbers = exception.stack.source_lines_numbers[0..exception.stack.source_lines_len];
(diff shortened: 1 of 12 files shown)


src/bun.js/javascript.zig
```diff
@@ -3078,6 +3084,7 @@ pub const VirtualMachine = struct {
                 defer source_url.deinit();
                 const file = bun.path.relative(dir, source_url.slice());
                 const func = frame.function_name.toUTF8(allocator);
+                defer func.deinit();
```
jdalton, line 3083: Nice catch!

jdalton (approved):

paperclover:
Why? SuppressedError is still an Error. And an uncaught error is still an error.

jdalton:
@Electroid  The intent is to allow devs to suppress errors from Github Actions. 
The use of [SuppressedError](https://github.com/tc39/proposal-explicit-resource-management?tab=readme-ov-file#the-suppressederror-error) is a neat way to do that though seems a little off-of-center from its intended use. Can we take inspiration from that and sniff the error `name` (for something similar) or maybe make it configurable of which error names are not to be annotated.

paperclover (changes requested):
The intent of this PR is make it possible to reduce noise in CIs by making a suppressable error type. that is not what SuppressedError does.

Do not use `SuppressedError` for this purpose. It can be thrown by built-in syntax from the using spec as linked by John-David. **These errors should always be logged in CI and the terminal**.

In my opinion we shouldnt have any error for this, because the fact the error is logged actually a different bug in bun, and this just hides it.

PR #8950, opened 2024-02-16
fix: make sure Bun.sleep(Date) doesn't resolve prematurely

### LINKED ISSUE #8834: Bun.sleep() rounds incorrectly

argosphil:
Fixes #8834.

Replaces #8833, #8859.

This makes Bun.sleep(new Date(number)) fulfill its promise only when Date.now() >= number.

### What does this PR do?

This is a very limited PR which adds a single line of code (plus tests), to ensure that we sleep at least as long as the difference between the current date and the target date. The previous code rounded towards zero in some circumstances, resulting in sleeps that were too short.

- [x] Code changes

I wrote automated tests


test/js/web/timers/setTimeout.test.js
```diff
@@ -157,6 +157,14 @@ it("Bun.sleep works with a Date object", async () => {
   expect(performance.now() - now).toBeGreaterThan(10);
```
paperclover: <i class='graphite__hidden'>[Re: line 155]<br/><br/></i>

can you resolve this test?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/oven-sh/bun/8950?utm_source=unchanged-line-comment">Graphite</a>.</span>

paperclover (changes requested):
i think this is much better than modifying setTimeout.

can you resolve the todo i marked earlier about this issue?

paperclover (approved):
great

jdalton:
Thank you @argosphil!

jdalton (approved):

PR #8953, opened 2024-02-17
docs: use consistent tsconfig for both in docs and `bun init`

risu729:
### What does this PR do?

Follow up of #8654

Fixes inconsistent tsconfig comments in docs vs in `bun init`, except for the stricter flags.

For the stricter flags, I decided to include them also in `bun init` but disable them by default.
This inconsistency is clarified in the docs as follows.
> If you run `bun init` in a new directory, this `tsconfig.json` will be generated for you. (The stricter flags are disabled by default.)

- [x] Code changes


Electroid (approved):
Thanks!

PR #8959, opened 2024-02-17
Regression test for #8929 (typo in child_process.js)

dhbaird:
### What does this PR do?

- [x] Code changes

This is a test case for #8929. Also mentioned in #8841 comments, there is a race condition that I would like to solve too, but at the expense of added complexity (need to compile a program to invoke `pause()`); Working around that by delaying. Feedback welcome. Thanks!

Run w/ and w/o #8929:
```
build/bun-debug test test/js/node/child_process/child_process.test.ts
```

### DIFF
--- a/test/js/node/child_process/child_process.test.ts
+++ b/test/js/node/child_process/child_process.test.ts
@@ -219,7 +219,7 @@ describe("spawn()", () => {
   test.if(isLinux)("should handle null for first 4 positions of stdio", async () => {
     // Run a program that does nothing except pause indefinitely, thus an
     // opportunity to inspect its file descriptors.
-    const proc = spawn("sleep", ["inf"], {stdio: [null, null, null, null]});
+    const proc = spawn("sleep", ["inf"], { stdio: [null, null, null, null] });
     // Delay here because the command (above) opens and closes a bunch of
     // files when starting up, so we have to wait for it to stabilize.  This
     // still is a race condition though, would be nice to eliminate.
--- a/test/js/node/child_process/child_process.test.ts
+++ b/test/js/node/child_process/child_process.test.ts
@@ -1,9 +1,10 @@
 // @known-failing-on-windows: 1 failing
-import { describe, it, expect } from "bun:test";
+import { describe, it, expect, test } from "bun:test";
 import { ChildProcess, spawn, execFile, exec, fork, spawnSync, execFileSync, execSync } from "node:child_process";
 import { tmpdir } from "node:os";
 import { promisify } from "node:util";
-import { bunExe, bunEnv } from "harness";
+import { readdirSync } from "node:fs";
+import { bunExe, bunEnv, isLinux } from "harness";
 import path from "path";
 
 const debug = process.env.DEBUG ? console.log : () => {};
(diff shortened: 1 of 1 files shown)


test/js/node/child_process/child_process.test.ts
```diff
@@ -214,6 +215,21 @@ describe("spawn()", () => {
     const { stdout } = spawnSync("echo", ["hello"], { encoding: "utf8" });
     expect(stdout.trim()).toBe("hello");
   });
+
+  it("should handle null for first 4 positions of stdio", async () => {
```
jdalton: If we split this out into a `test.if(process.platform === 'linux', (...` then we'll know its skipped in other cases (better visibility and awareness of what is and isn't running for a platform).

dhbaird: Done.

test/js/node/child_process/child_process.test.ts
```diff
@@ -214,6 +215,19 @@ describe("spawn()", () => {
...
   });
+
+  test.if(isLinux)("should handle null for first 4 positions of stdio", async () => {
+    // Run a program that does nothing except pause indefinitely, thus an
+    // opportunity to inspect its file descriptors.
+    const proc = spawn("sleep", ["inf"], { stdio: [null, null, null, null] });
+    // Delay here because the command (above) opens and closes a bunch of
+    // files when starting up, so we have to wait for it to stabilize.  This
+    // still is a race condition though, would be nice to eliminate.
+    await Bun.sleep(10);
```
jdalton: I love these comments @dhbaird! Can you expand the one " thus an opportunity to inspect its file descriptors." and explain what we are inspecting them for. To a casual reader getting back "0", "1", "2" may be opaque to the test of supporting 4 positions.

argosphil: Love the comments, too! This looks great.

I think it's sufficient to have a comment here about the existence of the race condition, which I suspect you could hit:
1. running on a single vCPU
2. in QEMU
3. while several other processes are running on the VM
4. with a guest kernel built with HZ=100

If you do want to eliminate it, we can wait for /proc/${pid}/status to include "State:\tS", which would indicate the dynamic loader has done its job and the process is now in sleep state, rather than running or in disk sleep. We could also use "yes" and wait for the first line of output, at which point the loader should have closed its fds, but that relies on a specific implementation of "yes"...

dhbaird: The "yes" proposal seems pretty good to me; A very creative idea. Is there a sensible scenario where it wouldn't work? I'm failing to think of one myself.

argosphil: > The "yes" proposal seems pretty good to me; A very creative idea. Is there a sensible scenario where it wouldn't work? I'm failing to think of one myself.

Well, I'm sure people have rewritten "yes" in a number of programming languages, and some of them might open more fds. I don't think that'd be a sensible thing to do, though... probably not worth worrying about

dhbaird: Updated comment and now using yes. Hopefully it's a net improvement? (Feedback welcome, of course.)

dhbaird: I did also test w/ and w/o the original regression (the typo) to make sure the test is still effective with "yes".

paperclover (approved):

argosphil:
Just a heads-up: #9019 modifies some code which is tested here and changed by #8981. It might be a good idea to merge them one at a time :-)

jdalton (approved):
LGTM

PR #8971, opened 2024-02-17
docs: Fix unclosed code block in the example for WebSocketHandler

eliot-akira:
### What does this PR do?

In the type definition for interface `WebSocketHandler`, the example in the docblock comment was missing the closing ` ``` `.

It resulted in the example looking like this in VS Code:

bun-websockethandler-example

This PR fixes it so it displays correctly.

bun-websockethandler-example-fixed


gvilums (approved):
Thank you!

PR #8981, opened 2024-02-18
Honor 'inherit' in stdio option of Spawn. Fixes #8844.

### LINKED ISSUE #8844: spawnSync() does not honor "inherit" after position 2 in stdio

1.0.26

Linux 5.15.0-46-generic x86_64 x86_64

```
import { spawnSync } from "node:child_process";
import fs from "node:fs";

const fdA = fs.openSync("/dev/null", "r");
const fdB = fs.openSync("/dev/null", "r");
const fdC = fs.openSync("/dev/null", "r");
const fdD = fs.openSync("/dev/null", "r");

const inherits = new Array(Math.max(fdA, fdB, fdC, fdD) + 1).fill(null);
inherits[1] = "inherit";
inherits[fdA] = "inherit";
inherits[fdB] = "inherit";
inherits[fdC] = "inherit";
inherits[fdD] = "inherit";

console.log("fdA =", fdA);
console.log("fdB =", fdB);
console.log("fdC =", fdC);
console.log("fdD =", fdD);
console.log("inherits =", inherits);
console.log("Output of ls /dev/fd:");
const proc = spawnSync("ls", ["/dev/fd"], { stdio: inherits });
```

<...snip...>
Output of ls /dev/fd:
0  1  12  13  14  15  2  3  4

### What do you see instead?

fdA = 12
fdB = 13
fdC = 14
fdD = 15
inherits = [
  null, "inherit", null, null, null, null, null, null, null, null, null, null, "inherit",
  "inherit", "inherit", "inherit"
]
Output of ls /dev/fd:
0  1  2  3  4

### Additional information

`bun.spawnSync()` has same issue.

dhbaird:
### What does this PR do?

Honors the "inherit" stdio options. Fixes #8844.

- [x] Code changes

I wrote automated tests.

  - [ ] ~I copied the structure that was already there `stdio_pipes`. Hopefully it does the right thing, and then so does mine.~
  - [x] Didn't introduce new uses of JSValue

### DIFF
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1977,6 +1977,10 @@ pub const Subprocess = struct {
         var ipc_mode = IPCMode.none;
         var ipc_callback: JSValue = .zero;
         var stdio_pipes: std.ArrayListUnmanaged(Stdio.PipeExtra) = .{};
+        var stdio_inherits: std.ArrayListUnmanaged(i32) = .{};
+        defer stdio_inherits.deinit(bun.default_allocator);
+        var stdio_close: std.ArrayListUnmanaged(i32) = .{};
+        defer stdio_close.deinit(bun.default_allocator);
         var pipes_to_close: std.ArrayListUnmanaged(bun.FileDescriptor) = .{};
         defer {
             for (pipes_to_close.items) |pipe_fd| {
@@ -2144,7 +2148,7 @@ pub const Subprocess = struct {
                             i += 1;
 
                             while (stdio_iter.next()) |value| : (i += 1) {
-                                var new_item: Stdio = undefined;
+                                var new_item: Stdio = .{ .ignore = {} };
                                 if (!extractStdio(globalThis, i, value, &new_item))
                                     return JSC.JSValue.jsUndefined();
                                 switch (new_item) {
@@ -2157,7 +2161,22 @@ pub const Subprocess = struct {
                                             return .zero;
                                         };
                                     },
-                                    else => {},
+                                    .inherit => {
+                                        stdio_inherits.append(bun.default_allocator,
+                                            @intCast(i)
+                                        ) catch {
+                                            globalThis.throwOutOfMemory();
+                                            return .zero;
+                                        };
+                                    },
+                                    else => {
+                                        stdio_close.append(bun.default_allocator,
+                                            @intCast(i)
+                                        ) catch {
+                                            globalThis.throwOutOfMemory();
+                                            return .zero;
+                                        };
+                                    },
                                 }
                             }
                         } else {
@@ -2481,6 +2500,21 @@ pub const Subprocess = struct {
                 _ = std.c.fcntl(fds[0], os.FD_CLOEXEC);
             };
             _ = maybe catch |err| return globalThis.handleError(err, "in configuring child stderr");
+        }
+
+        for (stdio_inherits.items) |fileno| {
+            const maybe = blk: {
+                if (comptime Environment.isMac) {
+                    actions.inherit(bun.toFD(fileno)) catch |err| break :blk err;
+                } else {
+                    actions.dup2(bun.toFD(fileno), bun.toFD(fileno)) catch |err| break :blk err;
+                }
+            };
+            _ = maybe catch |err| return globalThis.handleError(err, "in configuring child stdio");
+        }
+
+        for (stdio_close.items) |fileno| {
+            _ = actions.close(bun.toFD(fileno)) catch |err| return globalThis.handleError(err, "in configuring child stdio");
         }
 
         actions.chdir(cwd) catch |err| return globalThis.handleError(err, "in chdir()");
@@ -3266,6 +3300,7 @@ pub const Subprocess = struct {
         out_stdio: *Stdio,
     ) bool {
         if (value.isEmptyOrUndefinedOrNull()) {
+            out_stdio.* = Stdio{ .ignore = {} };
             return true;
         }
 
--- a/src/bun.js/api/bun/subprocess.zig
+++ b/src/bun.js/api/bun/subprocess.zig
@@ -1978,9 +1978,7 @@ pub const Subprocess = struct {
         var ipc_callback: JSValue = .zero;
         var stdio_pipes: std.ArrayListUnmanaged(Stdio.PipeExtra) = .{};
         var stdio_inherits: std.ArrayListUnmanaged(i32) = .{};
-        defer stdio_inherits.deinit(bun.default_allocator);
         var stdio_close: std.ArrayListUnmanaged(i32) = .{};
-        defer stdio_close.deinit(bun.default_allocator);
         var pipes_to_close: std.ArrayListUnmanaged(bun.FileDescriptor) = .{};
         defer {
             for (pipes_to_close.items) |pipe_fd| {
@@ -2162,7 +2160,7 @@ pub const Subprocess = struct {
                                         };
                                     },
                                     .inherit => {
-                                        stdio_inherits.append(bun.default_allocator,
+                                        stdio_inherits.append(allocator,
                                             @intCast(i)
                                         ) catch {
                                             globalThis.throwOutOfMemory();
@@ -2170,7 +2168,7 @@ pub const Subprocess = struct {
                                         };
                                     },
                                     else => {
-                                        stdio_close.append(bun.default_allocator,
+                                        stdio_close.append(allocator,
                                             @intCast(i)
                                         ) catch {
                                             globalThis.throwOutOfMemory();
--- a/test/js/node/child_process/child_process.test.ts
+++ b/test/js/node/child_process/child_process.test.ts
@@ -1,9 +1,10 @@
 // @known-failing-on-windows: 1 failing
-import { describe, it, expect } from "bun:test";
+import { describe, it, expect, test } from "bun:test";
 import { ChildProcess, spawn, execFile, exec, fork, spawnSync, execFileSync, execSync } from "node:child_process";
 import { tmpdir } from "node:os";
 import { promisify } from "node:util";
-import { bunExe, bunEnv } from "harness";
+import { openSync } from "node:fs";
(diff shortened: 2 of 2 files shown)


argosphil:
I'm confused. I think you're calling `dup2` in the parent process, not the fork()ed child process. Won't that mean that all subsequent spawns inherit the fd?

I just ran this on linux:
```
import {spawnSync} from "bun";
import {openSync} from "node:fs";
const fdA = openSync("/dev/null", "r");
const fdB = openSync("/dev/null", "r");
spawnSync(["ls", "/proc/self/fd"], { stdio: [ "inherit", "inherit", "inherit", "inherit" ]});
{let arr = [ "inherit", "inherit", "inherit" ];
 arr[fdA] = "inherit";
 arr[fdB] = "inherit";
 spawnSync(["ls", "/proc/self/fd"], { stdio: arr});
}
{let arr = [ "inherit", "inherit", "inherit" ];
 arr[fdB] = "inherit";
 spawnSync(["ls", "/proc/self/fd"], { stdio: arr});
}
spawnSync(["ls", "/proc/self/fd"], { stdio: [ "inherit", "inherit", "inherit" ]});
```

and the output was:
```
0  1  2  3  4
0  1  13  15  2  3  4  5  6
0  1  13  15  2  3  4  5  6
0  1  2  3
```

That doesn't look correct to me.

dhbaird:
Yikes, lemme check...

argosphil:
> I'm confused. I think you're calling `dup2` in the parent process, not the fork()ed child process. Won't that mean that all subsequent spawns inherit the fd?

Wait, no, I was misreading things. That part is actually fine, and I don't understand why my test didn't work.

dhbaird:
@argosphil I pushed updated test (commented out) that detects problem you highlighted. Still investigating...

argosphil:
@dhbaird Oops, I meant to link to a tree comparison, but got confused (again!) and opened a PR here:

https://github.com/dhbaird/bun/pull/1

Thanks for working on this, and I'll be happy to run further tests, of course!

dhbaird:
It's looking good (to me, anyways). Going to run it in a few minutes. I think you need to deinit stdio_close (look for `stdio_pipes.deinit`).

argosphil:
> It's looking good (to me, anyways). Going to run it in a few minutes. I think you need to deinit stdio_close (look for `stdio_pipes.deinit`).

Yes, you're right.

dhbaird:
I updated the PR here to include your fix @argosphil (thanks!). I also added the deinit on stdio_close.

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2640,6 +2676,7 @@ pub const Subprocess = struct {
             .stdout = Readable.init(stdio[1], bun.toFD(stdout_pipe[0]), jsc_vm.allocator, default_max_buffer_size),
             .stderr = Readable.init(stdio[2], bun.toFD(stderr_pipe[0]), jsc_vm.allocator, default_max_buffer_size),
             .stdio_pipes = stdio_pipes,
+            .stdio_inherits = stdio_inherits,
             .on_exit_callback = if (on_exit_callback != .zero) JSC.Strong.create(on_exit_callback, globalThis) else .{},
```
argosphil: I think we need to add `.stdio_close = stdio_close` here, as well?

dhbaird: /facepalm, yes, good catch, thanks. Pushed an update for that. (The default initializers on these structs aren't being very helpful in this case.)

argosphil:
I think `stdio_{close,inherits}` might not need to live as long as `stdio_pipes`, so maybe this could be optimized by making them local variables, not part of the Subprocess struct. However, I don't think that would actually be simpler because it adds code paths, and the information is potentially useful for debugging so I'd prefer to keep it in the struct. What do you think?

dhbaird:
Good observation; If I follow (@argosphil) correctly, `stdio_pipes` life must be extended so it can be accessed in `getStdio()`, but no such requirement exists for inherits or closes (that I know of). Is that right? If that's so, then I don't see any problem or added complexity by maintaining `stdio_inherits` and `stdio_close` m as local variables only. (But I also don't see much harm in extending their life either.) Unless someone tells me no, I'll go ahead and make them just local variables (remove them from the Subprocess struct).

dhbaird:
Pushed update per last comment.

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2483,6 +2500,23 @@ pub const Subprocess = struct {
...
+            const maybe = blk: {
+                if (comptime Environment.isMac) {
+                    actions.inherit(bun.toFD(fileno)) catch |err| break :blk err;
+                } else {
+                    actions.dup2(bun.toFD(fileno), bun.toFD(fileno)) catch |err| break :blk err;
+                }
+            };
+            _ = maybe catch |err| return globalThis.handleError(err, "in configuring child stdio");
+        }
+        stdio_inherits.deinit(bun.default_allocator);
```
argosphil: I'd prefer to move the deinit to a `defer` block (like the one around line 1984). That way, an error exit (in, say, line 2511) won't leak memory.

dhbaird: Thanks for that. Pushed a commit with the defers. This raises another question: doesn't stdio_pipes have the same problem? (Fixing it for stdio_pipes is probably out of scope for this PR, but good to have eyes looking at this.)

argosphil: > Thanks for that. Pushed a commit with the defers.

Note that we might want to use `allocator` and skip the deinit for those variables. I'm not really sure about the tradeoffs, to be honest.

> This raises another question: doesn't stdio_pipes have the same problem? (Fixing it for stdio_pipes is probably out of scope for this PR, but good to have eyes looking at this.)

I believe it does, but we should have a separate PR (or separate PRs) to fix the remaining problems with spawnMaybeSync:
1. deinit stdio_pipes on error (but not, of course, on success).
2. remove the "leak?" comment, after ensuring there's no leak left
3. allow empty arguments (rare, but useful, and potentially security-relevant as a maliciously empty string can shift option arguments into non-option slots)
4. set PATH before using it rather than afterwards
5. handle keys containing NUL characters in env and argv, which may otherwise result in very strange behavior
6. test all this

dhbaird: If `allocator` is stack-scoped (w/ an escape hatch to use heap or some other fallback, if a limit is reached; like some of the polymorphic allocators for >=C++17), then I'm for it. I don't know the ( bun/zig ) code well enough to be aware, but it seems like a reasonable expectation.

Regarding the leak, is there a leak risk? I thought the defer should pretty much solve that.

To make the history clean, I propose to squash the commits with you mentioned as a "Co-authored-by:" ... sound okay?

argosphil: > Regarding the leak, is there a leak risk? I thought the defer should pretty much solve that.

I think the code is fine, but there's a comment there that says "leak?", which is confusing.

> To make the history clean, I propose to squash the commits with you mentioned as a "Co-authored-by:" ... sound okay?

Absolutely, feel free to do whatever makes the most sense to you.

dhbaird: Oh right, I see what you are saying about the "leak?" comment now. I pushed a squashed version of the commits. So I guess we're set now?

dhbaird: One moment .... local allocator commit coming soon.

dhbaird: More questions :) What order do the defers execute in? Does the defer deinit of the local allocator (arena) happen before or after the things that use it? Also is there any point to doing a deinit of Arrays that allocate from it since they will (I assume) implicitly go away when the area deinits?

argosphil: > More questions :) What order do the defers execute in? Does the defer deinit of the local allocator (arena) happen before or after the things that use it?

defers execute in reverse/LIFO order (within a scope), so the arena is deinited after all the things that use it are.

> Also is there any point to doing a deinit of Arrays that allocate from it since they will (I assume) implicitly go away when the area deinits?

Certainly no point if the deinit is the last thing that happens before the arena deinit. The only reason I can see is if we allocated so much memory we might want to free it before allocating more, which isn't the case here.

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2157,7 +2161,22 @@ pub const Subprocess = struct {
...
-                                    else => {},
+                                    .inherit => {
+                                        stdio_inherits.append(bun.default_allocator,
+                                            @intCast(i)
+                                        ) catch {
+                                            globalThis.throwOutOfMemory();
+                                            return .zero;
+                                        };
+                                    },
+                                    else => {
```
argosphil, line 2170: ~Maybe use .ignore here explicitly, so if the enum changes there'll be an automatic compile-time error.~

Nevermind, forgot about the .fd possibility.

test/js/node/child_process/child_process.test.ts
```diff
@@ -260,6 +261,34 @@ describe("spawnSync()", () => {
     const { stdout } = spawnSync("echo", ["hello"], { encoding: "utf8" });
     expect(stdout.trim()).toBe("hello");
   });
+
+  test.if(isLinux)("should inherit for fds >= 3", async () => {
```
argosphil: The test as written works only for fds > 3 (fd 3 will be the /dev/fd dirhandle, so always open). However, as bun always opens an epoll fd, there's no risk of `fdA === 3` that I can see.

argosphil:
This looks good to me. If you put "fixes #8844" in the description (not the title) GitHub will link it to that issue and autoclose it once merged, I think.

The next thing to implement would be `stdio: [0, 2, 2]`, right?

dhbaird:
Hi, anything I can do to help move this forward? I think it's ready now (other than rebasing onto latest main). Thanks.

gvilums:
After the discussion in #8844 I assume that this is good to close. If not feel free to re-open!

PR #8995, opened 2024-02-19
Fix napi memory issue

### LINKED ISSUE #7864: Segmentation Fault Crash when using Prisma

1.0.20+09d51486e

- Perform a CPU-intensive task in the same method as where the database insertion can happen
- Use sharp to transform the image, for example, and try to insert some data

The insertion should happen normally.

### What do you see instead?

error: script was terminated by signal SIGSEGV (Address boundary error)
Segmentation fault

### Additional information

We tried to transform an image through `sharp` and store the S3 path to where they were uploaded.
We refactored the functions so that the transformation happens in another method. The transformations and uploads would happen normally, but during Prisma insertion, the script crashes directly.

**This works on 1.0.15, but not since.**

Output from` bunx prisma -v`:

```
prisma                  : 5.7.1
@prisma/client          : 5.7.1
Computed binaryTarget   : debian-openssl-1.1.x
Operating System        : linux
Architecture            : x64
Node.js                 : v18.17.0
Query Engine (Node-API) : libquery-engine 0ca5ccbcfa6bdc81c003cf549abe4269f59c41e5 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Schema Engine           : schema-engine-cli 0ca5ccbcfa6bdc81c003cf549abe4269f59c41e5 (at node_modules/@prisma/engines/schema-engine-debian-openssl-1.1.x)
Schema Wasm             : @prisma/prisma-schema-wasm 5.7.1-1.0ca5ccbcfa6bdc81c003cf549abe4269f59c41e5
Default Engines Hash    : 0ca5ccbcfa6bdc81c003cf549abe4269f59c41e5
Studio                  : 0.495.0
```

camero2734:
### What does this PR do?

Fixes #7864.

(Preface: I'm no expert by far, open to pointers in the right direction to fix this properly)

It seems that there's a memory issue in `napi_create_reference`. A dangling pointer issue can happen in the following scenario:
- We `napi_create_reference` to some value, with a `NapiRef` of `0xdeadbeef`.
- Eventually, something calls `napi_delete_reference`. `0xdeadbeef` gets `delete`d.
- Once again, there's a `napi_create_reference` _to the same JS object_.

The issue is that currently the object stores a `NapiRef` -- but what if this was `delete`d already, as in the scenario above? We assume the `0xdeadbeef` pointer is still valid, but it's a dangling reference to a deleted reference, thus ensues undefined behavior.

-----

Open to suggestions about how to better address this while trying to reuse `NapiRef`s (I assume for performance reasons), but for now I've gone ahead and just deleted the code that reuses a `NapiRef`, and instead _always_ create a `new NapiRef`.

I'll add some tests in a bit, but so far I've verified that my program^ works with this.

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -1151,19 +1151,6 @@ extern "C" napi_status napi_create_reference(napi_env env, napi_value value,
     Zig::GlobalObject* globalObject = toJS(env);
     JSC::VM& vm = globalObject->vm();
 
-    JSObject* jsObject = val.getObject();
-    NapiPrototype* object = jsDynamicCast<NapiPrototype*>(jsObject);
-    if (object && object->napiRef) {
-        *result = toNapi(object->napiRef);
-        return napi_ok;
-    }
-
-    NapiClass* object2 = jsDynamicCast<NapiClass*>(jsObject);
-    if (object2 && object2->napiRef) {
-        *result = toNapi(object2->napiRef);
-        return napi_ok;
-    }
-
     auto* ref = new NapiRef(globalObject, initial_refcount);
     if (initial_refcount > 0) {
         ref->strongRef.set(globalObject->vm(), val);
@@ -1175,12 +1162,6 @@ extern "C" napi_status napi_create_reference(napi_env env, napi_value value,
         } else {
             ref->weakValueRef.setPrimitive(val);
         }
-    }
-
-    if (object) {
-        object->napiRef = ref;
-    } else if (object2) {
-        object2->napiRef = ref;
     }
 
     *result = toNapi(ref);


Jarred-Sumner:
This looks right, and is similar to what Node does: https://github.com/nodejs/node/blob/2f4065250e453da2de2ee9cbb8aaae28cbdb4df5/src/js_native_api_v8.cc#L2626-L2649

Jarred-Sumner:
I'm inclined to merge this once the tests run

camero2734:
> I'm inclined to merge this once the tests run

Should I add any kind of tests to `test/js/third_party/prisma` to cover the case in #7864 or is it good without it?

Jarred-Sumner:
Would be awesome to add tests. They won't pass due to credentials, but that's okay as long as you test it locally. 

You could alternatively add tests to this file: https://github.com/oven-sh/bun/blob/f2cba2eabd7faea0eea3489b75bb7b474505ea97/test/napi/napi-app/main.cpp#L95 which would check that instantiating a napi object and then creating a napi ref to that object separately and deleting that ref would work, but up to you

PR #8998, opened 2024-02-19
Add Google Highway SIMD library

Jarred-Sumner:
### What does this PR do?

Makes stringWidth use the ghostty function

But we cannot merge since the test still fails


src/string_immutable.zig
```diff
@@ -5438,381 +5438,14 @@ pub fn pathContainsNodeModulesFolder(path: []const u8) bool {
...
-    }
-
-    if (isFullWidthCodepointType(T, cp)) {
-        return 2;
-    }
+extern "C" fn ghostty_simd_codepoint_width(cp: u32) u8;
 
-    return 1;
+pub fn visibleCodepointWidthType(comptime T: type, cp: T) usize {
+    return ghostty_simd_codepoint_width(@intCast(cp));
```
nektro, line 5448: this should be removed and `visibleCodepointWidth` should take `u32` instead of `anytype`. `u16`, `u21`, `u32` all coerce to `u32` with no cast necessary.

Issue #8976, opened 2024-02-18
Bun shell: files being emptied when used with `<`


aboqasem:
### What version of Bun is running?

1.0.27+c34bbb2e3

### What platform is your computer?

Darwin 23.2.0 x86_64 i386

### What steps can reproduce the bug?

```ts
import { afterEach, beforeEach, expect, test } from "bun:test";
import child_process from "child_process";
import fs from "fs";
import path from "path";
import { $ } from "bun";

$.throws(true);

let dir1: string;
let dir2: string;
let file1: string;
let file2: string;
let diffPath: string;
let diffFile: ReturnType<typeof Bun.file>;

beforeEach(async () => {
	dir1 = fs.mkdtempSync("dir1");
	dir2 = fs.mkdtempSync("dir2");
	file1 = path.join(dir1, "file");
	file2 = path.join(dir2, "file");
	diffPath = `${file2}.diff`;
	diffFile = Bun.file(diffPath);
	fs.writeFileSync(file1, "file1\n");
	fs.writeFileSync(file2, "file2\n");

	await $`/usr/bin/diff -ruN ${file1} ${file2} > ${diffPath} || true`;
});

afterEach(() => {
	fs.rmSync(dir1, { recursive: true });
	fs.rmSync(dir2, { recursive: true });
});

test("bun shell", async () => {
	const content = await diffFile.text();

	// USING Bun.file OR direct path FAILS
	// await $`/usr/bin/patch -s -p0 < ${diffFile}`;
	await $`/usr/bin/patch -s -p0 < ${diffPath}`;

	// FAILS
	expect(fs.readFileSync(file2, "utf-8")).toEqual(
		fs.readFileSync(file1, "utf-8"),
	);
	// FAILS - FILE IS EMPTY
	expect(content).toEqual(await diffFile.text());
});

test("child_process", async () => {
	const content = await diffFile.text();

	child_process.execSync(`/usr/bin/patch -s -p0 < "${diffPath}"`);

	// WORKS
	expect(fs.readFileSync(file2, "utf-8")).toEqual(
		fs.readFileSync(file1, "utf-8"),
	);
	// WORKS
	expect(content).toEqual(await diffFile.text());
});

```

### What is the expected behavior?

`bun shell` test to succeed.

### What do you see instead?

`bun shell` test fails.

### Additional information

-

aboqasem:
More errors while trying to create a minimal reproducible example:

```ts
describe.only("basic behaviour", () => {
[...]

PR #9000, opened 2024-02-19
Fix file redirect stdin not working

zackradisic:
### What does this PR do?

This fixes an error being returned when redirecting a file to stdin, and it also truncating the file #8976

The issue is that we were using wrong flags for `open(2)` when opening a file descriptor to be redirected to stdin. The error is caused by opening the file with O_WRONLY instead of O_RDONLY, and the file is truncated because it is opened with O_TRUNC

- [x] Code changes

Zig files changed:


paperclover (approved):
good. not in this pr but we shouldn't be using "bun.Mode" for things that arent file modes (permissions)

PR #9002, opened 2024-02-19
feat: use `DiffFormatter` for expect's `toHaveBeenNthCalledWith`

camc314:
warning - this is my first time writing zig, so this might be not be the best way todo it 🙂 

### What does this PR do?

uses `DiffFormatter` to show a richer diff when using `expect(mock).toHaveBeenNthCalledWith()`

Previously:

```
524 |                   const test = mock();
525 |                   test();
526 | 
527 |                   expect(test).toHaveBeenNthCalledWith(1, "heheh", "AAA");
         ^
error: expect(received).toHaveBeenNthCalledWith(expected)

n: 1
Received: []

Number of calls: 1
```

Now:

```
524 |                   const test = mock();
525 |                   test({ foo: "bar" });
526 | 
527 |                   expect(test).toHaveBeenNthCalledWith(1, "heheh", "AAA");
         ^
error: expect(received).toHaveBeenNthCalledWith(expected)

n: 1
Received:   [
+   {
+     foo: "bar",
+   }
-   "heheh",
-   "AAA",
-   empty item
  ]

- Expected  - 3
+ Received  + 3

Number of calls: 1
```

It does this by first convertin the arguments (expected) and recieved to a JSArray and then passing these to the DiffFormatter

- [x] Code changes

I tested it manually - i couldn't see any automated tests (snapshot) for these so didn't write any. If you want me to add tests, let me know

^^ not sure about these especially the first/third


src/bun.js/test/expect.zig
```diff
@@ -4114,20 +4114,27 @@ pub const Expect = struct {
         if (not) pass = !pass;
         if (pass) return .undefined;
 
-        // handle failure
-        var formatter = JSC.ConsoleObject.Formatter{ .globalThis = globalObject, .quote_strings = true };
-        const received_fmt = nthCallValue.toFmt(globalObject, &formatter);
+        const recieved = JSValue.createEmptyArray(globalObject, nthCallValue.getLength(globalObject));
```
coratgerl:
```suggestion
        const received = JSValue.createEmptyArray(globalObject, nthCallValue.getLength(globalObject));
```

camc314: aha thanks

src/bun.js/test/expect.zig
```diff
@@ -4114,20 +4114,27 @@ pub const Expect = struct {
         if (not) pass = !pass;
         if (pass) return .undefined;
 
-        // handle failure
-        var formatter = JSC.ConsoleObject.Formatter{ .globalThis = globalObject, .quote_strings = true };
-        const received_fmt = nthCallValue.toFmt(globalObject, &formatter);
+        const received = JSValue.createEmptyArray(globalObject, nthCallValue.getLength(globalObject));
```
Jarred-Sumner, line 3963: JSC has a way to call `JSC::constructArray()` from a `JSC::ArgList`. `JSC::ArgList` can be constructed from a `JSC::CallFrame`. 

I also wonder if we could add a slice() binding since that's what we want here (not literally calling Array.prototype.slice but an internal function which performs a similar operation)

Maybe we should add a binding for to `JSC.CallFrame` (zig) that calls `JSC::constructArray` instead of doing this?

Electroid: The status quo is very broken, can we follow up on this?

PR #9004, opened 2024-02-20
shell: Allow duplicating output fds (e.g. `2>&1`)

zackradisic:
### What does this PR do?

This PR enables support for the ["Duplicating an Output File Descriptor"](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_07_06) syntax (e.g. `2>&1`)

Only the stdout/stderr fds are supported right now. So you can redirect stderr to stdout `2>&1`, or redirect stdout to stderr `1>&2`

- [x] Code changes

Zig files changed:


src/shell/strsearch.zig
```diff
@@ -0,0 +1,4 @@
+const std = @import("std");
+const bun = @import("root").bun;
+
+pub fn search(str: anytype, needle: anytype) ?usize {}
```
Jarred-Sumner: tihs is dead code

Jarred-Sumner:
Can you add this to the docs to explain its supported? Along with stdin redirection

docs/runtime/shell.md
```diff
@@ -108,6 +108,34 @@ The following JavaScript objects are supported for redirection from:
 - `Bun.file(path)`, `Bun.file(fd)` (reads from the file)
 - `Response` (reads from the body)
 
+### To/From Files
```
Jarred-Sumner: Maybe a section on `Redirects`?

And then

### Redirecting stdout → file

### Redirecting stderr → file

### Redirecting stdin ← file

### Redirecting stdout → stderr

### Redirecting stdout ← stderr

Jarred-Sumner: basically, people might not immediately make the mental leap that  stdout and stderr are a "file" 

Also, we could do buffers in there too.

src/shell/subproc.zig
```diff
@@ -1463,6 +1465,7 @@ pub fn NewShellSubprocess(comptime EventLoopKind: JSC.EventLoopKind, comptime Sh
             spawn_args.stdio[2].setUpChildIoPosixSpawn(
                 &actions,
                 stderr_pipe,
```
Jarred-Sumner, line 1467: stderr_pipe,
stderr_pipe

test/js/bun/shell/bunshell.test.ts
```diff
@@ -661,6 +661,24 @@ describe("deno_task", () => {
     await TestBuilder.command`echo 1 > $EMPTY`.stderr("bun: ambiguous redirect: at `echo`\n").exitCode(1).run();
 
     await TestBuilder.command`echo foo bar > file.txt; cat < file.txt`.ensureTempDir().stdout("foo bar\n").run();
+
+    await TestBuilder.command`BUN_DEBUG_QUIET_LOGS=1 ${BUN} -e ${"console.log('Stdout'); console.error('Stderr')"} 2>&1`
```
Jarred-Sumner, line 665: you can remove BUN_DEBUG_QUIET_LOGS from these, it should do that automatically now

Issue #8093, opened 2024-01-11
Installed package missing vendored node_modules directory


jclem:
### What version of Bun is running?

1.0.22

### What platform is your computer?

Linux 6.5.11-linuxkit aarch64 unknown

### What steps can reproduce the bug?

Install `@tanstack/router-vite-plugin@1.3.3`. The `node_modules` directory that is [vendored by this package](https://www.npmjs.com/package/@tanstack/router-vite-plugin/v/1.3.3?activeTab=code) will be missing. This does not happen on macOS (`Darwin 23.2.0 arm64 arm`).

### What is the expected behavior?

The vendored `node_modules` directory should not be removed, since this package relies on it being present (see line 15 in the package's `build/cjs/packages/router-cli/src/config.js` file).

### What do you see instead?

No `node_modules` directory at the specified path.

eemelipa:
Here's minimal repro https://github.com/eemelipa/bun-issue-8093

Seems like bun install cache correctly has the `build/cjs/node_modules` in place but it isn't in place outside of the cache when installed.

eemelipa:
Seems like running `bun install --backend copyfile` can be used as a workaround as it copies `node_modules` over correctly.

So the problem has something to do with the other backend methods, e.g., `hardlink`

PR #9007, opened 2024-02-20
fix: install vendored node_modules when using hardlinks

eemelipa:
### What does this PR do?

Fixes [issue](https://github.com/oven-sh/bun/issues/8093) where vendored `node_modules` aren't installed with hardlink backend.

Vendored `node_module` is a folder inside an installed package which is called `node_module`. So the library has bundled some of their dependencies into the published package instead of specifying it in their `package.json`. For example

An npm package is called `vendor-baz` and it contains the following files:

```
x index.js
x package.json
x cjs/
x cjs/node_modules/
x cjs/node_modules/foo-dep/
x cjs/node_modules/foo-dep/index.js
```

The contents of `cjs/node_modules/foo-dep/` is a vendored package.

- [X] Code changes

I wrote automated tests


test/regression/issue/08093.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  root_url,
+  setHandler,
+} from "./../../cli/install/dummy.registry.js";
+
+beforeAll(dummyBeforeAll);
+afterAll(dummyAfterAll);
+beforeEach(dummyBeforeEach);
+afterEach(dummyAfterEach);
+
+it("should install vendored node_modules with hardlink", async () => {
```
eemelipa, line 24: Similar testing approach as tests in `test/cli/install/bun-install.test.ts` and more specifically test case `should prefer latest-tagged dependency`

src/install/install.zig
```diff
@@ -1366,7 +1366,7 @@ pub const PackageInstall = struct {
             cached_package_dir,
             this.allocator,
             &[_]bun.OSPathSlice{},
-            &[_]bun.OSPathSlice{bun.OSPathLiteral("node_modules")},
```
dylan-conway, line 1369: TIL you can publish an npm package with a `node_modules` folder as long as it isn't top level

dylan-conway (approved):
Looks good to me. Will wait for CI to run

dylan-conway:
We probably also want to do this for `installWithSymlink`, but it can be addressed in a future pr. I'm guessing there's a good chance top level `node_modules` will actually exist for local packages, so we would need to skip it

dylan-conway:
Thank you

PR #9010, opened 2024-02-20
fix: modify bcrypt to be able to verify long passwords directly

### LINKED ISSUE #9009: surprising Bun.password.verify behavior for "bcrypt" algorithm

1.0.28+705638470

From urretf on Discord:

```
const pass = prompt("Enter Password:");
console.log(`Read: ${pass}`);

console.log(`Testing Sync`);

const hash = Bun.password.hashSync(pass, { algorithm: 'bcrypt', cost: 10 });
console.log(`Hash: ${hash}`);

console.log(`Verify result: ${Bun.password.verifySync(pass, hash)}`);

console.log(`Testing Async`);

const hash2 = await Bun.password.hash(pass, { algorithm: 'bcrypt', cost: 10 });
console.log(`Hash: ${hash2}`);

console.log(`Verify result: ${await Bun.password.verify(pass, hash2)}`);

console.log(`Hash from sync and async were equal: ${hash == hash2}`);
```

Verify result: true for passwords shorter and longer than 72 characters

### What do you see instead?

Verify result: false if the entered password is longer than 72 characters

### Additional information

This is fairly easy to "fix", but should we? There's a test for the old behavior, and what we're implementing isn't, strictly speaking, the `bcrypt` algorithm.

argosphil:
Fixes #9009 

### What does this PR do?

Make the "bcrypt" algorithm (actually a variation of it) easier to use by using analogous code paths for hashing and verification.

This would break compatibility with old code, so maybe it's already too late for it.

- [x] Code changes

I altered the existing automated tests.

### DIFF
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -1952,7 +1952,19 @@ pub const Crypto = struct {
                     return true;
                 },
                 .bcrypt => {
-                    pwhash.bcrypt.strVerify(previous_hash, password, .{ .allocator = allocator }) catch |err| {
+                    var password_to_use = password;
+                    var outbuf: [bun.sha.SHA512.digest]u8 = undefined;
+
+                    // bcrypt silently truncates passwords longer than 72 bytes
+                    // we use SHA512 to hash the password if it's longer than 72 bytes
+                    if (password.len > 72) {
+                        var sha_512 = bun.sha.SHA512.init();
+                        defer sha_512.deinit();
+                        sha_512.update(password);
+                        sha_512.final(&outbuf);
+                        password_to_use = &outbuf;
+                    }
+                    pwhash.bcrypt.strVerify(previous_hash, password_to_use, .{ .allocator = allocator }) catch |err| {
                         if (err == error.PasswordVerificationFailed) {
                             return false;
                         }
--- a/test/js/bun/util/password.test.ts
+++ b/test/js/bun/util/password.test.ts
@@ -174,16 +174,12 @@ describe("verify", () => {
   });
 });
 
-test("bcrypt longer than 72 characters is the SHA-512", async () => {
+test("bcrypt uses the SHA-512 of passwords longer than 72 characters", async () => {
   const boop = Buffer.from("hey".repeat(100));
   const hashed = await password.hash(boop, "bcrypt");
-  expect(await password.verify(Bun.SHA512.hash(boop), hashed, "bcrypt")).toBeTrue();
-});
-
-test("bcrypt shorter than 72 characters is NOT the SHA-512", async () => {
-  const boop = Buffer.from("hey".repeat(3));
-  const hashed = await password.hash(boop, "bcrypt");
-  expect(await password.verify(Bun.SHA512.hash(boop), hashed, "bcrypt")).toBeFalse();
+  expect(await password.verify(boop, hashed, "bcrypt")).toBeTrue();
+  const boop2 = Buffer.from("hey".repeat(24));
+  expect(await password.verify(boop2, hashed, "bcrypt")).toBeFalse();
 });
 
 const defaultAlgorithm = "argon2id";


src/bun.js/api/BunObject.zig
```diff
@@ -1952,7 +1952,19 @@ pub const Crypto = struct {
                     return true;
                 },
                 .bcrypt => {
-                    pwhash.bcrypt.strVerify(previous_hash, password, .{ .allocator = allocator }) catch |err| {
+                    var password_to_use = password;
+                    var outbuf: [bun.sha.SHA512.digest]u8 = undefined;
+
+                    // bcrypt silently truncates passwords longer than 72 bytes
+                    // we use SHA512 to hash the password if it's longer than 72 bytes
+                    if (password.len > 72) {
```
jdalton, line 1960: ☝️ This is good to know! Can the 72 be put in a constant like BCRYPT_MAX_LEN to help in possibly other areas.

jdalton: Do we know how other folks handle this situation (other projects that expose this kind of functionality)?
Some switched to [Argon2](https://en.wikipedia.org/wiki/Argon2):
https://www.monterail.com/blog/more-secure-passwords-bcrypt

argosphil: > ☝️ This is good to know! Can the 72 be put in a constant like BCRYPT_MAX_LEN to help in possibly other areas.

I copied that comment from elsewhere in the same file. The tricky thing about that constant is it's in UTF-8 bytes, not JavaScript String.length.

> Do we know how other folks handle this situation (other projects that expose this kind of functionality)? Some switched to [Argon2](https://en.wikipedia.org/wiki/Argon2): https://www.monterail.com/blog/more-secure-passwords-bcrypt

Dropping "bcrypt" might not be the worst option, particularly since it's broken in the zig library, too (it insists on a `$2b$` prefix even though `$2a$` is also seen in the wild).

jdalton: Instead of working around the limitation I'm leaning towards the suggestion of others in this situation and favor throwing until we can switch to Argon2 or other alternative.

argosphil: > Instead of working around the limitation I'm leaning towards the suggestion of others in this situation and favor throwing until we can switch to Argon2 or other alternative.

I would go one step further: unless it's made explicit by an extra options argument how to handle passwords longer than 72 bytes, print a deprecation warning. The default strategy should be to throw when encountering a long password.

Rationale: 72 is just too low a limit. Assuming our input data has significant entropy in the first 72 bytes is dangerous. For example, I could see people using JSON data (or, uhm, XML) as a "password", and the common prefix of "password" documents may reduce the actual entropy significantly. I believe SHA512-ing our input data is nonstandard, and doing so for long inputs but not short ones seems inconsistent to me. That matters because people may well do some or all of the bcrypt work client-side.

jdalton: Love it @argosphil !

Jarred-Sumner:
Looks great. Thank you

src/bun.js/api/BunObject.zig
jdalton: Related to https://github.com/oven-sh/bun/pull/3204

paperclover: copying what i said in discord:

> bun isnt really the sort of program to print deprecation warnings. we dont have any of these yet. idk if we want to
maybe it's a good idea, but i think we can add the flag first and preserve backwards compatibility, and then we can think about deprecation.
>
> the reason for sha512 here is briefly mentioned in this comment by the author of libsodium.
https://github.com/oven-sh/bun/pull/3204#issuecomment-1575647630

jdalton: Deprecation warnings can be a whole other beast to tackle _(Node's machinery around it is non-trivial)_. It's good to have but if we can float on it for a bit (lower pri) that's a win in my book.

PR #9014, opened 2024-02-20
Change GitHub runners to use Ubicloud

Electroid:
### What does this PR do?

https://twitter.com/jarredsumner/status/1759597721877647644

- [ ] Code changes

### DIFF
--- a/.github/workflows/bun-linux-aarch64.yml
+++ b/.github/workflows/bun-linux-aarch64.yml
@@ -48,7 +48,7 @@ jobs:
             tag: linux-aarch64
             arch: aarch64
             build_arch: arm64
-            runner: linux-arm64
+            runner: ubicloud-standard-16-arm
             build_machine_arch: aarch64
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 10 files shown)


jdalton (approved):
Thank you for digging into this @Electroid! Much needed!

.github/workflows/bun-linux-aarch64.yml
```diff
@@ -48,7 +47,7 @@ jobs:
             tag: linux-aarch64
             arch: aarch64
             build_arch: arm64
-            runner: linux-arm64
+            runner: ubicloud-standard-16-arm
```
staabm, line 50: reading the concurrency limits on https://www.ubicloud.com/docs/github-actions-integration/price-performance#concurrency-limits sounds like you should not have too much `-16` jobs in parallel, as they seem to share a total of 64 CPUs (seems to depend on the pricing plan)

smaller runners also have less RAM -> its a tradeoff

ozgune: Hi there, I'm Ozgun, one of Ubicloud's founders. If you drop us an email at support@ubicloud.com about concurrency limits, we'd be happy to help out.

PR #9017, opened 2024-02-20
shell: Tilde expansion

zackradisic:
### What does this PR do?

This adds basic support for [tilde expansion](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_06_01) in Bun shell. 

"Basic" means that tilde is expanded in two scenarios:

```bash
# lone tilde expands to $HOME
echo ~

# tilde + path expands to $HOME/$PATH
echo ~/Documents
```

Also, this changes how JS strings/objects are substituted into the shell script. It now uses a non-printable char as a special token to denote a substituted JS string/object (it was previously using a tilde).

- [x] Code changes


src/shell/interpreter.zig
```diff
@@ -1530,6 +1530,21 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
...
+                                    const homedir = this.base.shell.getHomedir();
+                                    defer homedir.deref();
+                                    if (this.current_out.items.len > 0) {
+                                        switch (this.current_out.items[0]) {
+                                            '/', '\\' => {
+                                                this.current_out.insertSlice(0, homedir.slice()) catch bun.outOfMemory();
+                                            },
+                                            else => {
+                                                // TODO: Handle username
+                                                this.current_out.insert(0, '~') catch bun.outOfMemory();
```
cirospaciari, line 1542: for username maybe `bun.getUserName` can help

test/js/bun/shell/bunshell.test.ts
```diff
@@ -326,6 +326,20 @@ describe("bunshell", () => {
     expect(stdout.toString()).toEqual(`noice\n`);
   });
 
+  describe("tilde expansion", () => {
```
Electroid, line 329: can we have a test when `HOME` is not defined, empty, or invalid?

paperclover (approved):

test/js/bun/shell/bunshell.test.ts
Jarred-Sumner: also on windows will need to check the User env var

Jarred-Sumner (changes requested):
Comments from ciro and ashcon

src/shell/interpreter.zig
zackradisic: Will do this later as `~username` not commonly used, and also implementation is complicated on windows.

test/js/bun/shell/bunshell.test.ts
zackradisic: Added test for empty or weird values of HOME, and in doing so found a ref counted related leak, so added test for that as well

src/shell/interpreter.zig
```diff
@@ -1676,8 +1692,12 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
                         }
                     },
                     .compound => |cmp| {
-                        for (cmp.atoms[start_word_idx..]) |*simple_atom| {
-                            const is_cmd_subst = this.expandSimpleNoIO(simple_atom, &this.current_out);
+                        const starting_offset: usize = if (this.node.hasTildeExpansion()) brk: {
+                            this.word_idx += 1;
+                            break :brk 1;
+                        } else 0;
+                        for (cmp.atoms[start_word_idx + starting_offset ..]) |*simple_atom| {
```
Jarred-Sumner, line 1699: check that cmp.atoms is > 0

src/shell/interpreter.zig
```diff
@@ -1949,11 +1978,12 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
                 // this.out.array_of_ptr.append(@as([*:0]const u8, @ptrCast(buf.items.ptr))) catch bun.outOfMemory();
             }
 
-            fn expandVar(this: *const Expansion, label: []const u8) EnvStr {
+            fn expandVar(this: *const Expansion, label: []const u8) []const u8 {
                 const value = this.base.shell.shell_env.get(EnvStr.initSlice(label)) orelse brk: {
```
Jarred-Sumner, line 1982: change this to return a slice without incrementing and decrementing the ref count

src/shell/interpreter.zig
```diff
@@ -1617,6 +1631,8 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
                 }
             }
 
+            // fn expandTilde(this: *Expansion, )
```
jdalton, line 1634: Debug code ☝️

src/shell/shell.zig
```diff
@@ -1307,8 +1328,8 @@ pub const LexError = struct {
     /// Allocated with lexer arena
     msg: []const u8,
 };
-pub const LEX_JS_OBJREF_PREFIX = "~__bun_";
-pub const LEX_JS_STRING_PREFIX = "~__bunstr_";
+pub const LEX_JS_OBJREF_PREFIX = &[_]u8{8} ++ "__bun_";
+pub const LEX_JS_STRING_PREFIX = &[_]u8{8} ++ "__bunstr_";
```
jdalton, line 1332: ☝️ (learning): @zackradisic What's the benefit of this bit

**Update:**
Something about using it because it's a non-printable character
> // we use 8 or \b which is a non printable char

Doesn't `\b` clear the console output?

test/js/bun/shell/leak.test.ts
```diff
@@ -97,22 +96,22 @@ describe("fd leak", () => {
                   await ${builder.toString().slice("() =>".length)}.quiet().run()
                 })()
                 Bun.gc(true);
+                Bun.gc(true);
                 const val = process.memoryUsage.rss();
```
jdalton, line 100: ☝️ Related to https://github.com/oven-sh/bun/pull/8744/files

src/shell/shell.zig
zackradisic: We decided to use `\b` because it's hard to create a string with that so there's less of a chance of conflicts

We don't have to worry about `\b` clearing console output because we never print out the final script string. Though if were to create a function to return the final script string we would probably need to do a post-processing step to turn `\b__bun_0` -> `[Object object]` or something

Jarred-Sumner:
mrege conflicts

src/shell/shell.zig
jdalton: And no other character choice can be used (one perhaps without this gotcha?)

zackradisic:
See #10981

PR #9019, opened 2024-02-20
Add `BUN_DEBUG` flag to control where debug logs go

Jarred-Sumner:
### What does this PR do?

This lets you choose where the logs should go which is useful especially for subprocesses / IPC

Manually

### DIFF
--- a/src/output.zig
+++ b/src/output.zig
@@ -486,7 +486,7 @@ pub fn scoped(comptime tag: @Type(.EnumLiteral), comptime disabled: bool) _log_f
 
             if (!out_set) {
                 buffered_writer = .{
-                    .unbuffered_writer = writer(),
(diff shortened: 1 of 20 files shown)


src/output.zig
```diff
@@ -804,6 +804,56 @@ pub inline fn err(error_name: anytype, comptime fmt: []const u8, args: anytype)
     }
 }
 
+fn scopedWriter() std.fs.File.Writer {
+    if (comptime !Environment.isDebug) {
+        unreachable;
```
paperclover: `@compileError("...")`
This will put UB into release builds if hit.

paperclover (changes requested):
please assert at compile time that scopedWriter is not called in a release build.

paperclover (approved):

PR #9022, opened 2024-02-21
improve Bun.stringWidth's algorithm

nektro:
previous one was not parsing out the ansi escape codes at all


packages/bun-types/test.d.ts
```diff
@@ -731,6 +731,8 @@ declare module "bun:test" {
   }
 
   export interface MatchersBuiltin<T = unknown> {
+    [key: string]: any;
```
Jarred-Sumner: can you do this in a separate PR

src/bun.js/bindings/BunObject.cpp
```diff
@@ -675,6 +675,9 @@ class JSBunObject : public JSC::JSNonFinalObject {
         auto structure = createStructure(vm, globalObject, globalObject->objectPrototype());
         auto* object = new (NotNull, JSC::allocateCell<JSBunObject>(vm)) JSBunObject(vm, structure);
         object->finishCreation(vm);
+#ifndef NDEBUG // move this to table above when all the tests in test/js/bun/util/stringWidth.test.ts pass
```
```suggestion
#if BUN_DEBUG
```

src/string_immutable.zig
```diff
@@ -5905,20 +5903,29 @@ pub const visible = struct {
...
+    fn visibleUTF16WidthFn(input: []const u16, exclude_ansi_colors: bool) usize {
         var len: usize = 0;
-        while (bun.strings.firstNonASCII16CheckMin([]const u16, bytes, false)) |i| {
-            len += asciiFn(bytes[0..i]);
-            bytes = bytes[i..];
-
-            const utf8 = utf16CodepointWithFFFD([]const u16, bytes);
-            len += visibleCodepointWidthType(u32, utf8.code_point);
-            bytes = bytes[@min(@as(usize, utf8.len), bytes.len)..];
+        var iter = std.unicode.Utf16LeIterator.init(input);
```
Jarred-Sumner: 1) This is going to be quite a lot slower than the previous implementation. `firstNonASCII16CheckMin` should do the right  thing, then if the u16 char is > 255 then it needs to be decoded as UTF-16. `min` here means is it >= minimum non human readable character
2) We shouldn't use `std.unicode.*` at runtime in Bun's codebase, as it is less tested in production than our own implementations of these things and makes assumptions which are too restrictive for us to depend on

src/string_immutable.zig
```diff
@@ -5905,51 +5886,84 @@ pub const visible = struct {
...
         var len: usize = 0;
-        while (bun.strings.firstNonASCII16CheckMin([]const u16, bytes, false)) |i| {
-            len += asciiFn(bytes[0..i]);
-            bytes = bytes[i..];
 
-            const utf8 = utf16CodepointWithFFFD([]const u16, bytes);
-            len += visibleCodepointWidthType(u32, utf8.code_point);
-            bytes = bytes[@min(@as(usize, utf8.len), bytes.len)..];
+        if (firstNonASCII16([]const u16, input)) |idx| {
+            for (0..idx) |i| {
```
Jarred-Sumner: This should use SIMD. It won't be hard. I can show you how.

src/string_immutable.zig
```diff
@@ -5905,51 +5886,84 @@ pub const visible = struct {
...
+        _ = &input;
         var len: usize = 0;
-        while (bun.strings.firstNonASCII16CheckMin([]const u16, bytes, false)) |i| {
-            len += asciiFn(bytes[0..i]);
-            bytes = bytes[i..];
 
-            const utf8 = utf16CodepointWithFFFD([]const u16, bytes);
-            len += visibleCodepointWidthType(u32, utf8.code_point);
-            bytes = bytes[@min(@as(usize, utf8.len), bytes.len)..];
+        if (firstNonASCII16([]const u16, input)) |idx| {
```
Jarred-Sumner: This should wrap the entire function, not just the leading characters.

src/string_immutable.zig
```diff
@@ -5905,51 +5886,84 @@ pub const visible = struct {
...
+            for (0..idx) |i| {
+                len += visibleLatin1WidthScalar(@intCast(input[i]));
+            }
+            input = input[idx..];
         }
 
-        len += asciiFn(bytes);
-
+        var i: usize = 0;
+        var iter = std.unicode.Utf16LeIterator.init(input);
```
Jarred-Sumner: We should not use std.unicode.Utf16LeIterator. 

This function can be used instead, and the code will get simpler.
```zig
const utf8 = utf16CodepointWithFFFD([]const u16, bytes);
```

src/string_immutable.zig
```diff
@@ -5799,11 +5799,19 @@ pub fn isFullWidthCodepointType(comptime T: type, cp: T) bool {
     };
 }
 
-pub fn visibleCodepointWidth(cp: anytype) u3 {
-    return visibleCodepointWidthType(@TypeOf(cp), cp);
+pub fn visibleCodepointWidth(cp: u21) u3 {
```
Jarred-Sumner: This should either be `u32` or `i32`.

`u21` is much slower

src/string_immutable.zig
```diff
@@ -5799,11 +5799,19 @@ pub fn isFullWidthCodepointType(comptime T: type, cp: T) bool {
...
+    return visibleCodepointWidthType(u21, cp);
+}
+
+pub fn visibleCodepointWidthMaybeEmoji(cp: u21, maybe_emoji: bool) u3 {
```
Jarred-Sumner: This should either be `u32` or `i32`.

`u21` is much slower

src/string_immutable.zig
```diff
@@ -5816,49 +5824,31 @@ pub fn visibleCodepointWidthType(comptime T: type, cp: T) usize {
...
-            // this is a deliberate compiler optimization
-            // it disables auto-vectorizing the "input" for loop.
-            if (!(input.len < simd)) unreachable;
-        }
-
-        for (input) |c| {
-            length += if (c > 0x1f) 1 else 0;
-        }
-
+        for (input) |c| length += visibleLatin1WidthScalar(c);
```
Jarred-Sumner: This should use `@Vector` when the input is >= 16 bytes.

Jarred-Sumner: Depending on what godbolt says, this would probably work. It looks silly to paste it this way instead of using `inline for` but we want to make sure that the compiler knows the entire array is `const`

```zig
const input_end_ptr = input.ptr + input.len - (input.len % 16);
while (input.ptr != input_end_ptr) {
  const input_chunk: [16]u8 = input.ptr[0..16].*;
  const sums = [16]u8{
     visibleLatin1WidthScalar(input_chunk[0]),
     visibleLatin1WidthScalar(input_chunk[1]),
     visibleLatin1WidthScalar(input_chunk[2]),
     visibleLatin1WidthScalar(input_chunk[3]),
     visibleLatin1WidthScalar(input_chunk[4]),
     visibleLatin1WidthScalar(input_chunk[5]),
     visibleLatin1WidthScalar(input_chunk[6]),
     visibleLatin1WidthScalar(input_chunk[7]),
     visibleLatin1WidthScalar(input_chunk[8]),
     visibleLatin1WidthScalar(input_chunk[9]),
     visibleLatin1WidthScalar(input_chunk[10]),
     visibleLatin1WidthScalar(input_chunk[11]),
     visibleLatin1WidthScalar(input_chunk[12]),
     visibleLatin1WidthScalar(input_chunk[13]),
     visibleLatin1WidthScalar(input_chunk[14]),
     visibleLatin1WidthScalar(input_chunk[15]),
  };
  length += @reduce(.Add, sums); 
  input.ptr += 16;
}
input.len %= 16;
```

PR #9027, opened 2024-02-21
fix(ws/client): handle short reads on payload frame length

lithdew:
In the WebSocket specification, control frames may not be fragmented. However, the frame parser should handle fragmented control frames nonetheless. Whether or not the frame parser is given a set of fragmented bytes to parse is subject to the strategy in which the client buffers and coalesces received bytes.

All stages of the frame parser currently support parsing coalesced frames fragmented across multiple TCP segments except for the payload frame length parsing stage.

### What does this PR do?

This PR implements buffering the bytes of a frame's payload length into a client instance so that the websocket client is able to properly parse payload frame lengths despite there being a short read over incoming TCP data.

- [x] Code changes

A test is added to test/js/web/websocket/websocket-client-short-read.test.ts which creates a make-shift WebSocket server that performs short writes over a single WebSocket frame. The test passes with this PR.


argosphil:
Can you provide a hexdump of one of the problematic packets?

cirospaciari (changes requested):
I took a look and tested it using Autobahn and it looks good, but we need a test for this case, can you add a test for this specific case, so we don't regress on that?

src/http/websocket_http_client.zig
```diff
@@ -1320,23 +1320,17 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
...
                             .extended_payload_length_64 => @as(usize, 8),
-                            .extended_payload_length_16 => @as(usize, 2),
+                            .extended_payload_length_16 => @min(@as(usize, 2), data.len),
                             else => unreachable,
                         };
                         // we need to wait for more data
                         if (data.len == 0) return;
 
-                        if (data.len < byte_size) {
-                            this.terminate(ErrorCode.control_frame_is_fragmented);
```
argosphil, line 1330: why is this code being removed? If data.len = 3 but byte_size = 8, bad things will surely happen in line 1337/1331.

lithdew:
Scratch this PR - the issue turns out to be a lack of handling in short reads when parsing the frames payload length. Working on a fix right now and will amend this PR when ready with a test.

argosphil:
> I took a look and tested it using Autobahn

Can you share your test procedure? I'm genuinely curious as to what problem this is trying to fix.

> and it looks good

At this point, all I can say is it looks very confusing.

> but we need a test for this case, can you add a test for this specific case, so we don't regress on that?

We also need to test we don't panic on invalid packets.

cirospaciari:
> > I took a look and tested it using Autobahn
> 
> Can you share your test procedure? I'm genuinely curious as to what problem this is trying to fix.
> 
> > and it looks good
> 
> At this point, all I can say is it looks very confusing.
> 
> > but we need a test for this case, can you add a test for this specific case, so we don't regress on that?
> 
> We also need to test we don't panic on invalid packets.

Look for test\js\web\websocket\autobahn.test.ts

Looks like autobahn never trigger this condition so that why I'm asking for more tests before approving this.

src/http/websocket_http_client.zig
cirospaciari: we will need here a way to buffer the control frame instead of just hiting .need_body

argosphil:
> Scratch this PR - the issue turns out to be a lack of handling in short reads when parsing the frames payload length. Working on a fix right now and will amend this PR when ready with a test.

Thanks for explaining, that makes perfect sense. Please disregard~the~ my other messages, then, I'm sure they won't apply to the amended PR :-)

Edited to clarify.

src/http/websocket_http_client.zig
argosphil: There's a comment stating "Control frames must not be fragmented", but it seems in the OP's case, they sometimes are, and if node's `ws` handles that we should too. Is that the problem?

cirospaciari: Exactly

lithdew:
The PR has been updated with a test that fails prior to this PR being applied. Would appreciate a review 🙏.

test/js/web/websocket/websocket-client-short-read.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    ws.addEventListener("message", (event) => {
+      resolve(event.data.toString("utf-8"));
+    });
+
+    expect(await promise).toEqual(
+      `{"jsonrpc":"2.0","method":"transactionNotification","params":{"result":{"context":{"slot":249602179},"value":{"signature":"2P2xf3mU1vrnYcd1Lich8ELhhXx72oCinMaQXeqjvD7o4eKM5FB3NLahV7DWeQjifbk58KyhBwbwX1hgwgppfvMG","slot":249602179}},"subscription":4028}}`,
+    );
+    ws.close();
+    expect(await closed).toEqual("");
+    server.stop(true);
```
cirospaciari: can you use try/finally so we always stop the server?

src/http/websocket_http_client.zig
```diff
@@ -1323,22 +1326,33 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
...
-                            terminated = true;
+                        // copy available payload length bytes to a buffer held on this client instance
+                        const total_received = @min(byte_size - this.payload_length_frame_len, data.len);
+                        @memcpy(this.payload_length_frame_bytes[this.payload_length_frame_len..][0..total_received], data[0..total_received]);
+                        this.payload_length_frame_len += @intCast(total_received);
+                        data = data[total_received..];
+
+                        // short read on payload length - we need to wait for more data
+                        // whatever bytes were returned from the short read are kept in `payload_length_frame_bytes`
+                        if (this.payload_length_frame_len < byte_size) {
```
cirospaciari, line 1349: should we return in this condition instead of break? like in the `data.len == 0` condition

test/js/web/websocket/websocket-client-short-read.test.ts
```diff
@@ -0,0 +1,101 @@
+import { describe, test, expect } from "bun:test";
+
+const hostname = process.env.HOST || "127.0.0.1";
+const port = parseInt(process.env.PORT || "0");
+
+describe("WebSocket", () => {
```
cirospaciari, line 8: looks like this test is timing out on linux

test/js/web/websocket/websocket-server-echo.mjs
```diff
@@ -90,5 +90,5 @@ process.on("exit", exitCode => {
 });
 
 const hostname = process.env.HOST || "127.0.0.1";
-const port = parseInt(process.env.PORT) || 0;
+const port = parseInt(process.env.PORT || "0");
```
argosphil, line 93: I'm not sure I understand this change. It's not very important though, because the tests shouldn't run with environments which have $PORT set to an invalid string.

test/js/web/websocket/websocket-client-short-read.test.ts
```diff
@@ -0,0 +1,107 @@
+import { TCPSocketListener } from "bun";
+import { describe, test, expect } from "bun:test";
+import { WebSocket } from "ws";
+
+const hostname = process.env.HOST || "127.0.0.1";
+const port = parseInt(process.env.PORT || "0");
```
cirospaciari, line 6: we should always fix the port to 0 and get the dynamic port from `server`

src/http/websocket_http_client.zig
lithdew: Either a break or return should properly cleanup the `onData` handler. To keep styling consistent I changed it to a break.

test/js/web/websocket/websocket-client-short-read.test.ts
lithdew: Done.

test/js/web/websocket/websocket-client-short-read.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      const { promise, resolve } = Promise.withResolvers<string>();
+
+      client = new WebSocket(`ws://${server.hostname}:${server.port}`);
+      client.addEventListener("error", (err) => {
+        throw new Error(err.message);
+      });
+      client.addEventListener("close", (err) => {
+        if (!err.wasClean) {
+          throw new Error(err.reason);
```
cirospaciari, line 87: instead of throw you can use reject from the promise

lithdew: I amended the test to no longer await for a `closed` promise. Let me know if you'd prefer this being added back in, with the promise reporting possible errors on close.

test/js/web/websocket/websocket-client-short-read.test.ts
lithdew: It seems the other WebSocket tests also may take in the `PORT` variable (referring to websocket-client.test.ts). Let me know if it's preferred that `PORT` does not get used though.

cirospaciari (changes requested):
some comments and the test is timeouts on linux 
image

test/js/web/websocket/websocket-client-short-read.test.ts
lithdew: I can't seem to see the CI debug logs. Are you testing it locally? If so, could you share the logs?

lithdew:
> some comments and the test is timeouts on linux

Hm, might you have the `PORT` environment variable set by default by chance? Do you have the debug logs for i.e. syscalls that are ran?

cirospaciari:
> Contributor

I updated and rebuild after your last commits and the test now fails with this error:

image

here are the logs:

```bash
[0m[2m[SYS][0m openat(-100, /home/cirospaciari/test/bun/bunfig.toml) = 3
[0m[2m[SYS][0m openat(9[/home/cirospaciari], package.json) = 10
[0m[2m[fs][0m openat(9[/home/cirospaciari], /home/cirospaciari/package.json) = 10[/home/cirospaciari/package.json]
[0m[2m[fs][0m close(10[/home/cirospaciari/package.json])
[0m[2m[SYS][0m openat(10[/home/cirospaciari/test], package.json) = 11
[0m[2m[fs][0m openat(10[/home/cirospaciari/test], /home/cirospaciari/test/package.json) = 11[/home/cirospaciari/test/package.json]
[0m[2m[fs][0m close(11[/home/cirospaciari/test/package.json])
[0m[2m[SYS][0m openat(11[/home/cirospaciari/test/bun], package.json) = 12
[0m[2m[fs][0m openat(11[/home/cirospaciari/test/bun], /home/cirospaciari/test/bun/package.json) = 12[/home/cirospaciari/test/bun/package.json]
[0m[2m[fs][0m close(12[/home/cirospaciari/test/bun/package.json])
[0m[2m[SYS][0m openat(11[/home/cirospaciari/test/bun], tsconfig.json) = 12
[0m[2m[fs][0m openat(11[/home/cirospaciari/test/bun], /home/cirospaciari/test/bun/tsconfig.json) = 12[/home/cirospaciari/test/bun/tsconfig.json]
[0m[2m[fs][0m close(12[/home/cirospaciari/test/bun/tsconfig.json])
[0m[2m[SYS][0m openat(-100, /home/cirospaciari/test/bun/tsconfig.base.json) = 12
[0m[2m[fs][0m openat([invalid_fd], /home/cirospaciari/test/bun/tsconfig.base.json) = 12[/home/cirospaciari/test/bun/tsconfig.base.json]
[0m[2m[fs][0m close(12[/home/cirospaciari/test/bun/tsconfig.base.json])
[0m[2m[SYS][0m openat(-100, /home/cirospaciari/test/bun/test/js/web/websocket/) = 20
[0m[2m[SYS][0m openat(21[/home/cirospaciari/test/bun/test], package.json) = 22
[0m[2m[fs][0m openat(21[/home/cirospaciari/test/bun/test], /home/cirospaciari/test/bun/test/package.json) = 22[/home/cirospaciari/test/bun/test/package.json]
[0m[2m[fs][0m close(22[/home/cirospaciari/test/bun/test/package.json])
[0m[2m[SYS][0m openat(21[/home/cirospaciari/test/bun/test], tsconfig.json) = 22
[0m[2m[fs][0m openat(21[/home/cirospaciari/test/bun/test], /home/cirospaciari/test/bun/test/tsconfig.json) = 22[/home/cirospaciari/test/bun/test/tsconfig.json]
[0m[2m[fs][0m close(22[/home/cirospaciari/test/bun/test/tsconfig.json])
[0m[2m[SYS][0m openat(-100, /home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts) = 25
[0m[2m[fs][0m openat([invalid_fd], /home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts) = 25[/home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts]
[0m[2m[fs][0m close(25[/home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts])
[0m[2m[jest][0m describe("WebSocket")
[0m[2m[EventLoop][0m enter() = 1
[0m[2m[EventLoop][0m exit() = 1
[0m[2m[jest][0m test("short read on payload length")
[0m[2m[EventLoop][0m enter() = 0
[0m[2m[Listener][0m listen
[0m[2m[Loop][0m ref
[0m[2m[alloc][0m new() = src.http.websocket_http_client.NewHTTPUpgradeClient(false)@200002d3c00
[0m[2m[Loop][0m ref
[0m[2m[uws][0m connect(127.0.0.1, 37061)
[0m[2m[EventLoop][0m exit() = 0
[0m[2m[Listener][0m onCreate
[0m[2m[Socket][0m onOpen ssl: false
[0m[2m[Listener][0m markActive
[0m[2m[WebSocketClient][0m onOpen
[0m[2m[uws][0m us_socket_write(anyopaque@1c, 154) = 154
[0m[2m[Socket][0m onData(154)
[0m[2m[Listener][0m markActive
[0m[2m[EventLoop][0m enter() = 0
[0m[2m[uws][0m us_socket_write(anyopaque@1d, 129) = 129
[0m[2m[Socket][0m write(129, false) = 129
[0m[2m[uws][0m us_socket_write(anyopaque@1d, 3) = 3
[0m[2m[Socket][0m write(3, false) = 3
[0m[2m[Loop][0m ref
[0m[2m[Listener][0m markInactive
[0m[2m[EventLoop][0m exit() = 0
[0m[2m[WebSocketClient][0m onData
[0m[2m[Loop][0m sub 3 - 1 = 2
[0m[2m[WebSocketClient][0m onDidConnect
[0m[2m[EventLoop][0m enter() = 0
[0m[2m[alloc][0m new() = src.http.websocket_http_client.NewWebSocketClient(false)@20000051b80
[0m[2m[Loop][0m ref
[0m[2m[EventLoop][0m exit() = 0
[0m[2m[WebSocketClient][0m onData (need_header)
[0m[2m[WebSocketClient][0m onData (extended_payload_length_16)
[0m[2m[WebSocketClient][0m fail (control_frame_is_fragmented)
[0m[2m[EventLoop][0m enter() = 1
[0m[2m[loop][0m epoll_wait(31[anon_inode:[eventpoll]]) = 1
[0m[2m[SYS][0m read(30[anon_inode:[eventfd]], 8) = 8 (0.017ms)
[0m[2m[SYS][0m openat(-100, /home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts) = 32
[0m[2m[fs][0m openat([invalid_fd], /home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts) = 32[/home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts]
[0m[2m[fs][0m close(32[/home/cirospaciari/test/bun/test/js/web/websocket/websocket-client-short-read.test.ts])
[0m[2m[EventLoop][0m exit() = 1
[0m[2m[Loop][0m sub 3 - 1 = 2
[0m[2m[uws][0m us_socket_close(94728575965088)
[0m[2m[WebSocketClient][0m onClose
[0m[2m[Loop][0m sub 2 - 1 = 1
[0m[2m[uws][0m us_socket_write(anyopaque@1d, 254) = 254
[0m[2m[Socket][0m write(254, false) = 254

```

argosphil (commented):
Looks good to me!

Can you copy your comment from the PR to the source code so it's not lost for posterity, though?

> In the WebSocket specification, control frames may not be fragmented. However, the frame parser should handle fragmented control frames nonetheless. Whether or not the frame parser is given a set of fragmented bytes to parse is subject to the strategy in which the client buffers and coalesces received bytes.

That really makes sense to me.

lithdew:
> I updated and rebuild after your last commits and the test now fails with this error:

Odd. Could you rebuild once more as a sanity check?

There are no bytes written in the test I wrote that should trigger the only condition that would yield the error `control_frame_is_fragmented`.

The bytes written in the test from the server are a proper WebSocket text frame, and the code I amended in the parser does not cause the code to follow back to the case where that error may be reported.

test/js/web/websocket/websocket-server-echo.mjs
lithdew: It's just to have the types make sense as `parseInt` can only accept a `string` whereas `process.env.PORT` by itself is `string | undefined`.

In the case `parseInt` is given `undefined`, it'll yield a `NaN` which will get converted to the number 0 either way with the `||` clause.

lithdew:
> Can you copy your comment from the PR to the source code so it's not lost for posterity, though?

@argosphil Sure, no problem 👍 - added it to the source code before the parser loop in onData().

cirospaciari:
> > I updated and rebuild after your last commits and the test now fails with this error:
> 
> Odd. Could you rebuild once more as a sanity check?
> 
> There are no bytes written in the test I wrote that should trigger the only condition that would yield the error `control_frame_is_fragmented`.
> 
> The bytes written in the test from the server are a proper WebSocket text frame, and the code I amended in the parser does not cause the code to follow back to the case where that error may be reported.

Ok was a cache issue sorry, autobahn and the provided test pass now, will run all websockets tests to check

Update: waiting for the CI to run

argosphil:
I'm running with https://github.com/argosphil/bun/commit/d5c3c9605f19c1cdfdb3501f97b1b5fbf8db0231 this change which splits all data into single-byte chunks. Will let you know once the tests finish.

Do we need to handle fragmented control frames, too?

lithdew:
> I'm running with [argosphil@d5c3c96](https://github.com/argosphil/bun/commit/d5c3c9605f19c1cdfdb3501f97b1b5fbf8db0231) this change which splits all data into single-byte chunks. Will let you know once the tests finish.
> 
> Do we need to handle fragmented control frames, too?

`.need_header`, `.ping`, and subsequently `.pong` handles fragmented control frames properly. I believe this should cover handling all kinds of fragmented control frames properly.

That being said, it is definitely worth it to have an eventual refactor of the frame parser code tbh.

lithdew:
CI appears to be failing on `bun-linux / linux-x64` while installing dependencies needed to run tests. Would anyone be able to restart the CI for it?

cirospaciari:
> CI appears to be failing on `bun-linux / linux-x64` while installing dependencies needed to run tests. Would anyone be able to restart the CI for it?

Restarted

argosphil:
> > I'm running with [argosphil@d5c3c96](https://github.com/argosphil/bun/commit/d5c3c9605f19c1cdfdb3501f97b1b5fbf8db0231) this change which splits all data into single-byte chunks. Will let you know once the tests finish.

So there's a few websocket-related test failures, but it's quite possible I messed up. I'll investigate and I'd propose to merge this pending @cirospaciari 's review, and then to open a new issue for that...

> > Do we need to handle fragmented control frames, too?
> 
> `.need_header`, `.ping`, and subsequently `.pong` handles fragmented control frames properly. I believe this should cover handling all kinds of fragmented control frames properly.
> 
> That being said, it is definitely worth it to have an eventual refactor of the frame parser code tbh.

...and that. What do you think?

cirospaciari:
> > > I'm running with [argosphil@d5c3c96](https://github.com/argosphil/bun/commit/d5c3c9605f19c1cdfdb3501f97b1b5fbf8db0231) this change which splits all data into single-byte chunks. Will let you know once the tests finish.
> 
> So there's a few websocket-related test failures, but it's quite possible I messed up. I'll investigate and I'd propose to merge this pending @cirospaciari 's review, and then to open a new issue for that...
> 
> > > Do we need to handle fragmented control frames, too?
> > 
> > 
> > `.need_header`, `.ping`, and subsequently `.pong` handles fragmented control frames properly. I believe this should cover handling all kinds of fragmented control frames properly.
> > That being said, it is definitely worth it to have an eventual refactor of the frame parser code tbh.
> 
> ...and that. What do you think?

I think we can merge this if the current tests pass and we just do another PR for other fragmented issues.

Jarred-Sumner:
Thank you

PR #9044, opened 2024-02-22
Support `test.fixme`, `test.fixmeIf`, and `test.todoIf`

Electroid:
### What does this PR do?

This adds the same API as playwright's [`test.fixme()`](https://playwright.dev/docs/api/class-test#test-fixme-1) to the test runner. This allows developers to mark a test as flaky or broken, but does not fail the overall test suite.

We will use this internally to get our Github CI to green so we can turn on the merge queue. Others may find it useful.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,7 +13,7 @@
       "request": "launch",
       "name": "bun test [file]",
       "program": "${workspaceFolder}/build/bun-debug",
-      "args": ["test", "${file}"],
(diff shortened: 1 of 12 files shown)


Jarred-Sumner:
Per IRL discussion, I don't think we should do this. Maybe `todoIf`. But we should fix our own test suite instead of making it easier to write flaky tests.

Issue #7829, opened 2023-12-25
`process.env.NODE_ENV` should be `undefined` as default


guyutongxue:
### What version of Bun is running?

1.0.20

### What platform is your computer?

Linux 6.5.0-14-generic x86_64 x86_64

### What steps can reproduce the bug?

```sh
$ cat test.mjs
console.log(process.env.NODE_ENV);
$ bun test.mjs
development
```

### What is the expected behavior?

Should output `undefined`, just like Node.js does:
```sh
$ node test.mjs
undefined
```

### What do you see instead?

Bun produces `development`.

### Additional information

Vite check this value when running `vite build`. If `NODE_ENV` was not set, then `vite build` will automatically set `NODE_ENV` to `production`, then everything goes fine (on Node.js). 

But Bun uses `development` as default `NODE_ENV`, so `vite build` just take this instead of set it to `production`... then things like `import.meta.env.PROD` will be wrong.

As a consequence, I need to run `NODE_ENV=production bunx --bun vite build` instead of simple `bunx --bun vite build` to get the correct result.

TiBianMod:
Related to PR https://github.com/oven-sh/bun/pull/7666

also packages like `dotenv`, will always show `NODE_ENV=development`, no matter what you have on `.env` file

guyutongxue:
OK... I should have checked closed issues; #7663 addresses almost same question -- but in a contrast direction 🫠 

Anyway I tend to believe that undefined `NODE_ENV` is the correct behavior, at least for compatibility to Node.js.

Jarred-Sumner:
`undefined` is the correct behavior. but we unfortunately made it `"development"` early on and changing that is a breaking change.

Electroid:
We are thinking of making this change in Bun v1.1

Electroid:
This change will be announced and made in Bun 1.1

marcospgp:
has this also been fixed for the bundler? I am still seeing it as `development`:

```
// test.js
console.log(process.env.NODE_ENV);
```

```
$ bun build test.js                                                                                                 git:main*
// test.js
console.log("development");
$ bun --version                                                                                                     git:main*
1.1.42
```

vdawg-git:
Same thing for standalone binaries. I still get `"development"`

paperclover:
> has this also been fixed for the bundler? I am still seeing it as

[...]

PR #9049, opened 2024-02-22
Update vite.md

kacperpaczos:
* Added a warning about limited optimization of Vite with Bun at the beginning.
* Changed the section title to a more formal one.
* Separated the installation and navigation steps into individual points.
* Added an explanation of the --bun flag and a note about the shebang line.
* Updated the "dev" script in package.json, adding the --bun flag.
* Introduced an example of script configurations for an application.
* Added a link to the Vite documentation at the end of the document.

### What does this PR do?

- [ ] Code changes

### DIFF
--- a/docs/guides/ecosystem/vite.md
+++ b/docs/guides/ecosystem/vite.md
@@ -3,68 +3,88 @@ name: Build a frontend using Vite and Bun
 ---
 
 {% callout %}
-While Vite currently works with Bun, it has not been heavily optimized, nor has Vite been adapted to use Bun's bundler, module resolver, or transpiler.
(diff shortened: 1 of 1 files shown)


paperclover:
this doesnt render well, and i dont understand how the guides formatter actually works.

image

the site isnt public right now, so i don't know what to do about this.

kacperpaczos:
the site isnt public right now, so i don't know what to do about this.

What do you mean? I found it on the website...
https://bun.sh/guides/ecosystem/vite

guyutongxue:
Notice that `bunx --bun vite build` doesn't do the same thing as `vite build` (running under `node`) because of the inconsistent `NODE_ENV`. You should use `NODE_ENV=production bunx --bun vite build` for now, see #7829.

PR #9066, opened 2024-02-23
vscode: highlight invisible and ambiguous characters

nektro:
makes it much more clear that the hidden characters in this test are intentional and that they exist

image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -81,8 +81,8 @@
   // Markdown
   "[markdown]": {
     "editor.defaultFormatter": "biomejs.biome",
-    "editor.unicodeHighlight.ambiguousCharacters": false,
-    "editor.unicodeHighlight.invisibleCharacters": false,
+    "editor.unicodeHighlight.ambiguousCharacters": true,
+    "editor.unicodeHighlight.invisibleCharacters": true,
     "diffEditor.ignoreTrimWhitespace": false,
     "editor.wordWrap": "on",
     "editor.quickSuggestions": {


sequencerr:
Maybe use `\uXXXX` for these tests?

nektro:
that was just one illustrative example. a bunch recently snuck into our markdown as well and these options would've caught that https://github.com/oven-sh/bun/pull/9067

.vscode/settings.json
```diff
@@ -81,8 +81,8 @@
   // Markdown
   "[markdown]": {
     "editor.defaultFormatter": "biomejs.biome",
-    "editor.unicodeHighlight.ambiguousCharacters": false,
-    "editor.unicodeHighlight.invisibleCharacters": false,
+    "editor.unicodeHighlight.ambiguousCharacters": true,
```
Electroid, line 84: Won't this only affect markdown files?

sequencerr: I guess it will. It's `[markdown]` scope https://code.visualstudio.com/docs/getstarted/settings#_language-specific-editor-settings

sequencerr:
IMHO this setting for cases like Latin/Cyrillic **ambiguous** chars `The character U+0441 "с" could be confused with the ASCII character U+0063 "c", which is more common in source code.`
There is extension for tracking bad chars https://marketplace.visualstudio.com/items?itemName=nhoizey.gremlins

.vscode/settings.json
nektro: good point. i have the setting enabled globally so that might be why i didnt catch that

jdalton (approved):
Improved visibility for the invisible things is a good thing. I've been bit in the past by various invisible characters causing problems on other people's systems _(different languages and encodings)_.

PR #9067, opened 2024-02-23
remove hidden zero-width-space characters from our md files

nektro:


jdalton:
My fav character is the ZWJ. It's a valid identifier character so you can make variable names that are very hard to collide with

PR #9069, opened 2024-02-23
Documentation: `Update installation.md`

lucasmichot:
Simplify homebrew installation instructions

### What does this PR do?

This PR changes the homebrew installation instructions.
It is possible to tap and install in only one command, which is always better than using 2 commands.

First remove bun and untap : `brew remove bun && brew untap oven-sh/bun`
Then install bun again: `brew install oven-sh/bun/bun`


jdalton:
Thank you @lucasmichot!

PR #9076, opened 2024-02-23
Update primordials

jdalton:
### What does this PR do?

In a step to improving our interoperability with Node on fronts like `node:http` we needed to update our primordials (a lower level module used by Node lib modules.

side by side comparison

The image above is comparing Node v20.11.1, Bun main branch, and Bun `jdalton/primordials`. I created a JS file that mimics Node's [`lib/internal/per_context/primordials.js`](https://github.com/nodejs/node/blob/main/lib/internal/per_context/primordials.js) module.
<details>
<summary>benchmark init time (<i>"a"</i>) and 1,000x iterations of each entry</summary>

```js
console.time("a");
//
// The code of lib/internal/per_context/primordials.js
//
console.timeEnd("a");

console.time("Set");
  new Set([1, 2, 3, 4]);
}
console.timeEnd("Set");

console.time("Map");
  new Map([
    [1, 2],
    [3, 4],
  ]);
}
console.timeEnd("Map");

console.time("SafeMap");
  new primordials.SafeMap([
    [1, 2],
    [3, 4],
  ]);
}
console.timeEnd("SafeMap");

console.time("SafeSet");
  new primordials.SafeSet([1, 2, 3, 4]);
}
console.timeEnd("SafeSet");

console.time("makeSafe");
  primordials.makeSafe(Map, class extends Map {});
}
console.timeEnd("makeSafe");

console.time("SafeWeakMap");
  new primordials.SafeWeakMap([
    [{}, {}],
    [{}, {}],
  ]);
}
console.timeEnd("SafeWeakMap");

console.time("SafeWeakSet");
  new primordials.SafeWeakSet([{}, {}, {}, {}]);
}
console.timeEnd("SafeWeakSet");

console.time("SafeWeakRef");
  new primordials.SafeWeakRef({});
}
console.timeEnd("SafeWeakRef");

console.time("hardenRegExp");
  primordials.hardenRegExp(/a/);
}
console.timeEnd("hardenRegExp");

console.log(Reflect.ownKeys(primordials).length);
```
</details>
With the primordials PR we are now creating more primordials (better Node interop), are doing so faster than Node, and on par with Bun's existing code path. I did this by optimizing the creation of the primordials in the following areas:

* unrolling generic iteration
* special pathing uncurrying
* leveraging WebKit JSC `@` helpers
* avoiding known slow paths in JS-land 
_(like `Object.defineProperty`, `Object.defineProperties`, `Object.getOwnPropertyDescriptor`, and friends)_. 

Methods like the internal `primordials.makeSafe` _(that is used nowhere else in Node's code)_ were removed in favor of less generic, more specific approaches. 

I also curated the list of primordials _(documenting which ones have been skipped)_ because Node added a bunch of junk-entries for things like SomeMethodName, SomeMethodLength, and so on.

### DIFF
--- a/src/js/internal/primordials.js
+++ b/src/js/internal/primordials.js
@@ -1,218 +1,1605 @@
-// TODO: Use native code and JSC intrinsics for everything in this file.
-// Do not use this file for new code, many things here will be slow especailly when intrinsics for these operations is available.
-// It is primarily used for `internal/util`
+// This file subclasses and stores the JS built-ins that come from the VM
+// so that Node.js's built-in modules do not need to later look these up from
+// the global proxy, which can be mutated by users.
 
+// Use of primordials have sometimes a dramatic impact on performance, please
+// benchmark all changes made in performance-sensitive areas of the codebase.
+// See: https://github.com/nodejs/node/pull/38248
+
+const primordials = $Object.$create(null);
+
+const {
+  asyncIterator: SymbolAsyncIterator,
+  for: SymbolFor,
+  hasInstance: SymbolHasInstance,
+  isConcatSpreadable: SymbolIsConcatSpreadable,
+  iterator: SymbolIterator,
+  match: SymbolMatch,
+  matchAll: SymbolMatchAll,
+  replace: SymbolReplace,
+  search: SymbolSearch,
+  species: SymbolSpecies,
+  split: SymbolSplit,
+  toPrimitive: SymbolToPrimitive,
+  toStringTag: SymbolToStringTag,
+  unscopables: SymbolUnscopables,
+} = Symbol;
+
+const ObjectDefineProperty = $Object.$defineProperty;
+const { freeze: ObjectFreeze } = $Object;
+
+const defineProps = (object, thruples) => {
+  for (let i = 0, { length } = thruples; i < length; i += 3) {
+    const key = thruples[i];
+    const kind = thruples[i + 1];
+    const value = thruples[i + 2];
+    ObjectDefineProperty(object, key, $createObjectWithoutPrototype(kind, value));
+  }
+  return object;
+};
+
+const finishSafeCtor = SafeCtor => {
+  $setPrototypeDirect.$call(SafeCtor.prototype, null);
+  ObjectFreeze(SafeCtor.prototype);
+  ObjectFreeze(SafeCtor);
+  return SafeCtor;
+};
+
+const defineSafeCtor = (SafeCtor, { ctor, proto }) => {
+  if (ctor) {
+    defineProps(SafeCtor, ctor);
+  }
+  if (proto) {
+    defineProps(SafeCtor.prototype, proto);
+  }
+  return finishSafeCtor(SafeCtor);
+};
+
+/**
+ * Creates a class that can be safely iterated over.
+ *
+ * Because these functions are used by `makeSafe`, which is exposed on the
+ * `primordials` object, it's important to use const references to the
+ * primordials that they use.
+ * @template {Iterable} T
+ * @template {*} TReturn
+ * @template {*} TNext
+ * @param {(self: T) => IterableIterator<T>} factory
+ * @param {(...args: [] | [TNext]) => IteratorResult<T, TReturn>} next
+ * @returns {Iterator<T, TReturn, TNext>}
+ */
 const createSafeIterator = (factory, next) => {
-  class SafeIterator {
-    constructor(iterable) {
-      this._iterator = factory(iterable);
+  return finishSafeCtor(
+    class SafeIterator {
+      #iterator;
+      constructor(iterable) {
+        this.#iterator = factory(iterable);
+      }
+      next() {
+        return next(this.#iterator);
+      }
+      [SymbolIterator]() {
+        return this;
+      }
+    },
+  );
+};
+
+const FinalizationRegistryPrototype = FinalizationRegistry.prototype;
+const {
+  register: FinalizationRegistryPrototypeRegisterRaw,
+  unregister: FinalizationRegistryPrototypeUnregisterRaw,
+  [SymbolToStringTag]: FinalizationRegistryPrototypeSymbolToStringTag,
+} = FinalizationRegistryPrototype;
+
+const MapGetSymbolSpeciesRaw = $Map.__lookupGetter__(SymbolSpecies);
+
+const MapPrototype = $Map.prototype;
+const MapPrototypeClearRaw = MapPrototype.$clear;
+const MapPrototypeDeleteRaw = MapPrototype.$delete;
+const MapPrototypeForEachRaw = MapPrototype.$forEach;
+const MapPrototypeGetRaw = MapPrototype.$get;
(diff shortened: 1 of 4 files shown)


jdalton:
Fixing test fails.

paperclover:
I think the approach of primordials.js is a bad idea in general, which is why jhmasters prior pr was merged with this comment added:

```ts
// TODO: Use native code and JSC intrinsics for everything in this file.
// Do not use this file for new code, many things here will be slow especailly when intrinsics for these operations is available.
// It is primarily used for `internal/util`
```

Even if it is faster than before and faster than node (which is a nice thing to see), I'm of the opinion that none of these methods should be used if possible. If this is merged, I'd like a warning comment of this nature to be 

They also aren't loaded before user code (i imagine this file is slow to load), so it doesn't really achieve the safety guarantee that node primoridals do (which is already not that strong considering so much of the node js library is pollutable despite the primordials). And for the things that *are* safe, these should just be done on a per-case basis using the @ intrinsics, which the builtin bundler will already do for many cases.

jdalton:
@paperdave 
> They also aren't loaded before user code (i imagine this file is slow to load), so it doesn't really achieve the safety guarantee that node primoridals do (which is already not that strong considering so much of the node js library is pollutable despite the primordials).

We should definitely load these first to have such guarantees. Primordials via intrinsics or other mechanisms are important. The reason to have things similar to Node is to ease translation/porting for interoperability because meeting developers and the ecosystem where they are is critical for Bun's success and future. When possible I'm having Node's primordials use JSC private methods and our self hosted code _(native modules written in JS)_ that aren't using primordials should definitely be using the JSC private methods when possible _(my preference is explicit use over magic transpilation)_.

> Even if it is faster than before and faster than node (which is a nice thing to see), I'm of the opinion that none of these methods should be used if possible. If this is merged, I'd like a warning comment of this nature to be

The benefit of these in easing porting of code outweighs the ugly factor or current perf concerns since Node today is saddled with a slower implementation and we can continue to augment with private JSC methods (I've even found private JSC method `@isRegExpObject` is 2x faster than our `isRegExp` from `node:util/types`).

Jarred-Sumner:
Bun’s performance advantage today often comes from technical choices that leverage a deeper understanding & control of system internals. Instead of “what’s the easiest way to make this work given how it works today?”, we lean towards “what’s the most performant way we could make this work reliably?”. Often at overwhelming implementation difficulty (for developers outside of Bun)

The constraints of Node.js/V8 and Bun/JSC are different. Making the interfaces match is very much a goal, but making the implementation match is both a non-goal and in many cases an anti-goal for us to continue to be faster. 

If we really need to have primordials like this, then it’d make more sense to add them to the String, Array, etc prototype in JSC directly as private properties than to make these primordials in a separate function copied. This avoids prototype pollution issues correctly without incurring an extra initialization overhead other than the extra properties added. I’m still not convinced we need all the extra properties in most cases.

jdalton:
@Jarred-Sumner 
> Bun’s performance advantage today often comes from technical choices that leverage a deeper understanding & control of system internals. 

Performance that leads to incorrect results or crashes is not __real__ performance. It's more like a magician’s card trick ✨ . Bun cannot achieve its goal of replacing Node as the ecosystem go-to on perf alone. Devs don’t care what whizbang tech Bun is built on when their stuff doesn’t work. 

>  Instead of “what’s the easiest way to make this work given how it works today?”, we lean towards “what’s the most performant way we could make this work reliably?”. Often at overwhelming implementation difficulty (for developers outside of Bun)

Said another way: Bun is boxing itself in without a lifeline _(the community)_. Implementations _(and duplicate implementations)_ are so complex it's difficult for issues to be root caused, crashes squashed, and issues reliably reproduced – leading to performance that __doesn’t work__. Bun, with an ever growing surface area for crashes and security exploits, is burdened with opaque tech debt that __gets heavier__ with each release.

> The constraints of Node.js/V8 and Bun/JSC are different. Making the interfaces match is very much a goal, but making the implementation match is both a non-goal and in many cases an anti-goal for us to continue to be faster.

Bun is nearly 3 years old and struggling with fundamentals precisely because __it didn’t match implementations__. Behind many methods, especially those self-hosted in JavaScript, there are non-obvious side-effects and unintentionally established cow-paths. The vast majority of Node APIs are not super interesting or performance critical. They __just need to work__. Bun can be faster in very big ways while still aligning with implementations of many APIs. Bun is the newbie trying to gain ground in Node’s ecosystem. It doesn’t have the luxury of choice just yet. 

> If we really need to have primordials like this, then it’d make more sense to add them to the String, Array, etc prototype in JSC directly as private properties than to make these primordials in a separate function copied. This avoids prototype pollution issues correctly without incurring an extra initialization overhead other than the extra properties added.

Yes! This is ideal. However, I have noticed Node leverage some of these in more generic ways _(many JS Array and String methods are generic allowing values to be array-like or coerced to strings)_ which creates obstacles to stumble over _(as I ran into even in this PR)_.

 > I’m still not convinced we need all the extra properties in most cases.

There is room to prune for sure. We are not at the pruning stage yet.

Jarred-Sumner:
> Performance that leads to incorrect results or crashes is not real performance. It's more like a magician’s card trick ✨ . Bun cannot achieve its goal of replacing Node as the ecosystem go-to on perf alone. Devs don’t care what whizbang tech Bun is built on when their stuff doesn’t work.

100%


> Bun is nearly 3 years old and struggling with fundamentals precisely because it didn’t match implementations

IMO, that's not why - it's because I based the implementations of node:http, node:fs, etc on the docs/on the documented behavior instead of the implemented behavior and didn't copy Node's test suite to verify it works correctly - instead manually writing tests which were (and are) not nearly comprehensive enough. Copy-pasting implementations is one way to make it work, but it is not the only way

> Yes! This is ideal. However, I have noticed Node leverage some of these in more generic ways (many JS Array and String methods are generic allowing values to be array-like or coerced to strings) which creates obstacles to stumble over (as I ran into even in this PR).

Makes sense, we can skip extra wrappers and skip worrying about userland overwriting builtins when they're private properties

jdalton:
@Jarred-Sumner 
> Copy-pasting implementations is one way to make it work, but it is not the only way

Yes, though not matching implementation due to basing things off of docs _(something devs are notoriously lax on, me included)_ or otherwise is still not matching implementations 🙃 🏃‍♂️ 🙈 . 

I've found every time I've tried to get clever or tweak implementations a bit it's been far too easy for something to fall through the cracks _(side effects of self-hosted JS implementations and its forgiving APIs are a bummer)_. Copy-pasta / less-creative-direct-porting is not exciting but does give confidence that the code will just work, makes for updating as Node changes easier _(I end up thinking less about the why it does something and more about the steps)_, and more easily gets Node's own tests to pass.

> Makes sense, we can skip extra wrappers and skip worrying about userland overwriting builtins when they're private properties

I think this will be the next PR 😸 _(the private methods are so nice, there is some caution around the ones WebKit provides as they seem to add and remove them at will)_

jdalton:
Moved to https://github.com/oven-sh/bun/pull/9306

PR #9080, opened 2024-02-24
fix debug crash in bun.strings.firstNonASCII16 when input is ascii_u16_vector_size

nektro:
if the input slice length is exactly ascii_u16_vector_size then the `if` would fail and the assert on line 4388 would trip


Jarred-Sumner:
nice

PR #9087, opened 2024-02-24
WebSockets client: handle TCP fragmentation and simplify code

### LINKED ISSUE #9052: WebSockets: TCP fragmentation needs to be handled

1.0.28+705638470

_No response_

This is a follow-up to #9027.

WebSockets are implemented using TCP (or Unix domain sockets). TCP has fragmentation. The WebSockets protocol also has fragmentation. They are distinct and happen at different layers. That confused me.

Our code currently does not handle TCP fragmentation for WebSockets connections. That's not usually a problem because TCP connections tend to send out each WebSockets frame in a new TCP packet. But sometimes it is, so it should be fixed.

This bug can be reproduced by running a WebSockets server which sends out each byte in an individual TCP packet, or simulating one by splitting the data after receiving it.

The expected behavior is that our WebSockets code handles all possible TCP fragmentation types, including single-byte packets for everything.

### What do you see instead?

Using single-byte packets, in the client code, WS pings and pongs currently fail. I haven't tested the server code yet. Fragmentation that happens in the WS payload appears to be handled by the existing code.

### Additional information

I'm testing code locally which fixes the client and will open a PR soon.

argosphil:
Fixes #9052.

### What does this PR do?

This handles TCP fragmentation (TCP bytes being delivered in smaller chunks than they were sent) in the WebSockets client. It's different from WebSockets fragmentation, where one data chunk is split between several WebSockets frames, which may or may not be sent as separate TCP fragments.

- [x] Code changes

I ran the Autobahn test suite. This isn't sufficient as it doesn't appear to split packets into individual octets.

- [ ] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be [not yet!]
- [x] I included a test for the new code, or an existing test covers it [kinda]
- [ ] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test-file-name.test`) [more tests needed]

### DIFF
--- a/src/http/websocket_http_client.zig
+++ b/src/http/websocket_http_client.zig
@@ -936,7 +936,7 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
         globalThis: *JSC.JSGlobalObject,
         poll_ref: Async.KeepAlive = Async.KeepAlive.init(),
 
-        buffered_data: []u8 = &.{},
+        buffered_data: []u8 = "",
         required_data_len: usize = 0,
 
         event_loop: *JSC.EventLoop = undefined,
@@ -983,7 +983,7 @@ pub fn NewWebSocketClient(comptime ssl: bool) type {
             this.poll_ref.unref(this.globalThis.bunVM());
             if (this.buffered_data.len > 0) {
                 bun.default_allocator.free(this.buffered_data);
-                this.buffered_data = this.buffered_data[0..0];
+                this.buffered_data = "";
             }
             this.clearReceiveBuffers(true);
             this.clearSendBuffers(true);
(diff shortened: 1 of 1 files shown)


Jarred-Sumner (commented):
Thank you for this, but I think we should avoid the extra memory allocations when possible. We only need to dynamically allocate when we don't receive enough data, and that can go into a single ArrayList which is reused. Additionally, we can use `bun.copy` to handle copying potentially overlapping data efficiently (without temporary allocations)

Also, when making a slice have length zero, its best to use either `&.{}` or `""` instead of `slice = slice[0..0]` because `slice[0..0]` will preserve the original pointer which makes it easier to turn into a UAF later on.

argosphil:
> Thank you for this, but I think we should avoid the extra memory allocations when possible.

> We only need to dynamically allocate when we don't receive enough data,

That's what the existing PR does. The fast path is fundamentally unchanged (it checks a slice for .len > 0 rather than checking the initial_handler pointer for != null).

However, if we do hit the slow path, there are some unnecessary allocations (we can avoid the extra allocation entirely or do it on the stack). I'll get rid of them.

> and that can go into a single ArrayList which is reused.

I'm not sure I see why an ArrayList would be beneficial.

> Additionally, we can use `bun.copy` to handle copying potentially overlapping data efficiently (without temporary allocations)
> Also, when making a slice have length zero, its best to use either `&.{}` or `""` instead of `slice = slice[0..0]` because `slice[0..0]` will preserve the original pointer which makes it easier to turn into a UAF later on.

Thanks, I'll make those changes!

ETA: changes made. Not using an ArrayList for now, as it doesn't appear to be needed.

lithdew:
Unsure if the intent is to have all the refactors previously discussed for the websocket client to be done in this PR, but I feel the logic here is a bit messy/expensive as the websocket frame parser by itself shouldn't ever need any allocations in any fast/slow path. The only allocations ever needed is in the buffering of data in the `receive_buffer` of the client.

Once I get some freelance work done in roughly a few weeks, I can get to work on implementing an allocation-free standalone version of the frame parser w/ tests in a separate file in the `websockets` dir. This should help fully isolate concerns around allocations away to the client and away from the frame parser itself (and make the implementation extremely efficient IMO).

If you have time and want to get this done ASAP though @argosphil, I can point to some reference code on Discord that an implementation can be based off of as a start.

argosphil:
> Unsure if the intent is to have all the refactors previously discussed for the websocket client to be done in this PR,

It isn't, it's just to have as little code as possible in the rarely-exercised path of frame fragmentation.

> but I feel the logic here is a bit messy/expensive as the websocket frame parser by itself shouldn't ever need any allocations in any fast/slow path. The only allocations ever needed is in the buffering of data in the `receive_buffer` of the client.

What about the initial data, which can have any length and needs to be buffered? I feel it's best to piggyback onto that rather than having yet another slow code path for the (fixed-length, for now, admittedly) control frame fragmentation logic.

> Once I get some freelance work done in roughly a few weeks, I can get to work on implementing an allocation-free standalone version of the frame parser w/ tests in a separate file in the `websockets` dir. This should help fully isolate concerns around allocations away to the client and away from the frame parser itself (and make the implementation extremely efficient IMO).

I think the fragmented control frame case is really rare, while the initial data case is somewhat more common. Handling both through dynamic allocation seems sensible to me, since it's needed in the common case and it makes the rare case use code which is exercised at least once in a while.

As for efficiency, I think the additional icache pressure of having extra code outweighs the benefit of avoiding dynamic allocation in the fragmented control frame case. The initial data case might be another question entirely.

> If you have time and want to get this done ASAP though @argosphil, I can point to some reference code on Discord that an implementation can be based off of as a start.

I'd love that, feel free to message me in whichever way is most convenient to you.

Issue #9086, opened 2024-02-24
error when importing a module


sigmaSd:
### What version of Bun is running?

1.0.26

### What platform is your computer?

linux

### What steps can reproduce the bug?

index.js
```ts
import { Deno } from "@deno/shim-deno"
```

### What is the expected behavior?

no error

### What do you see instead?

error
```
1 | (function (entry, fetcher)
    ^
SyntaxError: Export named 'fdatasync' not found in module 'fs'.
```

### Additional information

works in node

Issue #5122, opened 2023-09-12
TypeError: The "original" argument must be of type Function


ScribeSavant:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Linux 5.15.0-1040-kvm x86_64 x86_64

### What steps can reproduce the bug?

# Code
``` 

// testConvo.js

import { Bot, session } from "grammy";
import { conversations, createConversation } from "@grammyjs/conversations";

async function greeting(conversation, ctx) {
  await ctx.reply("Hi there! What is your name?");
  const { message } = await conversation.wait();
  await ctx.reply(`Welcome to the chat, ${message.text}!`);
}

export const bot = new Bot(Bun.env.BOT_TOKEN);

bot.use(session({
  initial() {
    return {};
  },
}));

// Install the conversations plugin.
bot.use(conversations());
bot.use(createConversation(greeting));
bot.command("start", async (ctx) => {
  await ctx.conversation.enter("greeting");
});
bot.start()
console.log("Bot running")
```

# Steps
1.  `bun add grammy @grammyjs/conversations`
2.  `bun run testConvo.js`

### What is the expected behavior?

Bot running

### What do you see instead?

```
[0.03ms] ".env"
360 |   var keys = Object.keys(add), i = keys.length;
361 |   while (i--)
362 |     origin[keys[i]] = add[keys[i]];
363 |   return origin;
364 | 
365 | }, kCustomPromisifiedSymbol = Symbol.for("util.promisify.custom"), promisify = function promisify2(original) {
                                                                                                                                                                                                       ^
TypeError: The "original" argument must be of type Function
      at promisify2 (node:util:365:196)
      at /home/squirrelpython/codes/ximi-bun/node_modules/@deno/shim-deno/dist/deno/stable/functions/fdatasync.js:7:6
      at globalThis (/home/squirrelpython/codes/ximi-bun/node_modules/@deno/shim-deno/dist/deno/stable/functions/fdatasync.js:8:20)
      at require (:1:20)
      at /home/squirrelpython/codes/ximi-bun/node_modules/@deno/shim-deno/dist/deno/stable/functions.js:42:4
      at globalThis (/home/squirrelpython/codes/ximi-bun/node_modules/@deno/shim-deno/dist/deno/stable/functions.js:174:104)
      at require (:1:20)
[...]

PR #9092, opened 2024-02-24
Add missing fdatasync

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/9086
Fixes https://github.com/oven-sh/bun/issues/5122

There is a test

### DIFF
--- a/src/js/node/fs.js
+++ b/src/js/node/fs.js
@@ -200,6 +200,9 @@ var access = function access(...args) {
   open = function open(...args) {
     callbackify(fs.open, args);
   },
+  fdatasync = function fdatasync(...args) {
+    callbackify(fs.fdatasync, args);
+  },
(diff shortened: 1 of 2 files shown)


src/js/node/fs.js
```diff
@@ -200,6 +200,9 @@ var access = function access(...args) {
   open = function open(...args) {
     callbackify(fs.open, args);
   },
+  fdatasync = function fdatasync(...args) {
+    callbackify(fs.fdatasync, args);
+  },
   read = function read(fd, buffer, offsetOrOptions, length, position, callback) {
```
jdalton, line 206: Looks good. Only nit is the internal src/js/node/fs helper `callbackify` is the same name as `util.callbackify` with different functionality (leads to at a glance confusion)

PR #9097, opened 2024-02-25
Support unix domain sockets in fetch()

Jarred-Sumner:
### What does this PR do?

This adds an option `"unix"` to `fetch` that lets you provide a path to a unix domain socket. I'm not sure about this API in particular, but this needs to exist somehow

Tests in `fetch.unix.test.ts`


argosphil:
Can we keep the naming consistent between `Bun.serve` (which uses `unix`) and `fetch`?

gvilums:
> Can we keep the naming consistent between `Bun.serve` (which uses `unix`) and `fetch`?

changed to use `unix` in `fetch`

src/bun.js/webcore/response.zig
```diff
@@ -798,6 +798,7 @@ pub const Fetch = struct {
             this.request_headers = Headers{ .allocator = undefined };
 
             if (this.http != null) {
+                allocator.free(this.http.?.client.unix_socket_path);
```
cirospaciari: this could be inside of HttpClient deinit? who owns the memory? should we set unix_socket_path to empty string after this and put it inside clearData?

src/bun.js/webcore/response.zig
```diff
@@ -1720,11 +1723,19 @@ pub const Fetch = struct {
...
 
-        var url_proxy_buffer: []const u8 = undefined;
+        var url_proxy_buffer: []const u8 = "";
         var is_file_url = false;
         var reject_unauthorized = script_ctx.bundler.env.getTLSRejectUnauthorized();
         var check_server_identity: JSValue = .zero;
+
+        defer {
+            if (free_memory_reporter) {
+                unix_socket_path.deinit();
```
cirospaciari, line 1738: this condition is right?

src/http.zig
```diff
@@ -747,6 +763,10 @@ pub const HTTPThread = struct {
     }
 
     pub fn connect(this: *@This(), client: *HTTPClient, comptime is_ssl: bool) !NewHTTPContext(is_ssl).HTTPSocket {
+        if (client.unix_socket_path.len > 0) {
+            return try this.context(is_ssl).connectSocket(client, client.unix_socket_path);
+        }
+
         if (client.http_proxy) |url| {
```
cirospaciari, line 770: should we error on Bun__fetch if we try to use http_proxy + unix socket?

test/js/web/fetch/fetch.unix.test.ts
```diff
@@ -0,0 +1,40 @@
...
+});
+
+it("provide body", async () => {
+  const path = startServerUnix({
+    fetch(req) {
+      return new Response(req.body);
+    },
+  });
+  // POST with body
+  for (let i = 0; i < 20; i++) {
```
cirospaciari, line 45: we also should add unix + redirect tests

cirospaciari (changes requested):
some comments

src/http.zig
gvilums: Yeah we probably should

src/bun.js/webcore/response.zig
gvilums: i'll change the name of the variable, from my understanding this is an error cleanup thing

src/http.zig
gvilums: added

src/bun.js/webcore/response.zig
gvilums: moved to http client deinit

test/js/web/fetch/fetch.unix.test.ts
gvilums: added

test/js/web/fetch/fetch.unix.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+      return new Response(null, { status: 404 });
+    },
+  });
+  const path = startServerUnix({
+    fetch(req) {
+      if (req.url.endsWith("/hello")) {
+        return new Response(null, {
+          status: 302,
+          headers: { Location: `http://${server.hostname}:${server.port}/world` },
```
cirospaciari:
```suggestion
          headers: { Location: `${server.url.origin}/world` },
```

cirospaciari (changes requested):
one comment and test/js/node/stream/node-stream.test.js (SIGTERM) on bun-darwin-aarch64

src/http.zig
cirospaciari: we also need a test for this case

Jarred-Sumner:
The node-stream test is a flaky one, I think unrelated to these changes

Jarred-Sumner:
Please add a test for this:
> should we error on Bun__fetch if we try to use http_proxy + unix socket?

PR #9101, opened 2024-02-25
Fix prisma generate by emitting readable event on EoF

### LINKED ISSUE #5320: Bun doesn't run prisma generate or prisma migrate inside docker containers or WSL

# Docker:
```dockerfile
FROM oven/bun AS builder

WORKDIR /tmp/app

COPY package.json .
COPY bun.lockb .

RUN bun install

COPY src ./src
COPY tsconfig.json .

#   RUN bunx prisma:generate

RUN bun run build

FROM oven/bun AS runner

WORKDIR /app

RUN bun install -p

# Refer to the comment above

#   RUN bun add prisma

CMD ["bun", "start"]
```

# WSL:

Install WSL normally...
```powershell
wsl --install
```

Prisma installs normally

### What do you see instead?

```bash
$ bunx prisma migrate deploy
```

### Additional information

_No response_

camero2734:
### What does this PR do?
Fixes #5320

The main reason (there's currently two) that `bunx prisma generate` fails is because of the `getHash` method that listens for the `readable` event on a `ReadStream`:
```ts
// Prisma calls this in its ensureBinariesExist method that's called before every CLI command
// ../fetch-engine/src/getHash.ts
var import_crypto2 = __toESM(require("crypto"));
var import_fs13 = __toESM(require("fs"));
function getHash(filePath) {
  const hash = import_crypto2.default.createHash("sha256");
  const input = import_fs13.default.createReadStream(filePath);
  return new Promise((resolve3) => {
    input.on("readable", () => {
      const data = input.read();
      if (data) {
        hash.update(data);
      } else {
        resolve3(hash.digest("hex"));
      }
    });
  });
}
```

There's a bug in Bun where this event is not called for EoF (i.e. where `read` returns `null`).

As such, the `else` statement is never called, and ends up causing the whole program to exit silently.

-------

This seems to be broken due to a reason related to #4409. We have:
```cpp
// src/bun.js/bindings/JSReadableHelper.cpp
EncodedJSValue emitReadable(JSGlobalObject* lexicalGlobalObject, JSObject* stream, JSReadableState* state)
{
    VM& vm = lexicalGlobalObject->vm();

    state->setBool(JSReadableState::needReadable, false);
    if (!state->getBool(JSReadableState::emittedReadable)) {
        state->setBool(JSReadableState::emittedReadable, true);
        Zig::GlobalObject* globalObject = reinterpret_cast<Zig::GlobalObject*>(lexicalGlobalObject);
        // This is the issue
        globalObject->queueMicrotask(JSValue(globalObject->emitReadableNextTickFunction()), JSValue(stream), JSValue(state), JSValue {}, JSValue {});
    }
}
```

From some testing, it seems this microtask only gets run _after_ the stream has been `destroyed`, at which point the event is not allowed to be emitted anymore.

Similar to #4409, it seems this needs to be called via `process.nextTick` instead so that things run in the correct order.

To fix this, I moved the `onEofChunk` method into JS (`src/js/node/stream.js`) so that I could schedule it with `process.nextTick` instead. With this change, all expected `readable` events are emitted/received, and this fixes prisma CLI commands (almost).

(P.S. Maybe there's a way to add to the `nextTick` queue from C++, but I couldn't find a straightforward way of doing it, so that's why I moved things back to JS)

I added a test to `test/js/node/stream/node-stream.test.js`. It times out in previous versions of bun, but passes with these changes.

I also tested Prisma in a docker container. `bun-patched` is a bun binary built in release mode with these changes.

<details>
<summary>Dockerfile</summary>

```dockerfile
FROM debian:bullseye-slim

# Disable the runtime transpiler cache by default inside Docker containers.
# On ephemeral containers, the cache is not useful
ARG BUN_RUNTIME_TRANSPILER_CACHE_PATH=0
ENV BUN_RUNTIME_TRANSPILER_CACHE_PATH=${BUN_RUNTIME_TRANSPILER_CACHE_PATH}

COPY ./bun-patched /usr/local/bin/bun

RUN groupadd bun \
      --gid 1000 \
    && useradd bun \
      --uid 1000 \
      --gid bun \
      --shell /bin/sh \
      --create-home \
    && ln -s /usr/local/bin/bun /usr/local/bin/bunx \
    && which bun \
    && which bunx \
    && bun --version

WORKDIR /home/bun/app

COPY --link package.json bun.lockb ./

RUN bun install

COPY . .

RUN sed -i '1s/node/bun/' ./node_modules/prisma/build/index.js
RUN bunx prisma generate && echo "Exited" && sleep 1000
# ^ Prints the expected prisma generate success message
```

</details>

<details>
<summary>Relevant output</summary>

```
#13 [8/9] RUN sed -i '1s/node/bun/' ./node_modules/prisma/build/index.js
#13 DONE 0.2s

#14 [9/9] RUN bunx prisma generate && echo "Exited" && sleep 1000
#14 0.912 Prisma schema loaded from prisma/schema.prisma
#14 1.250
#14 1.250 ✔ Generated Prisma Client (v5.10.2) to ./node_modules/@prisma/client in 41ms
#14 1.250
#14 1.250 Start using Prisma Client in Node.js (See: https://pris.ly/d/client)
#14 1.250 ```
#14 1.250 import { PrismaClient } from '@prisma/client'
#14 1.250 const prisma = new PrismaClient()
#14 1.250 ```
#14 1.250 or start using Prisma Client at the edge (See: https://pris.ly/d/accelerate)
#14 1.250 ```
#14 1.250 import { PrismaClient } from '@prisma/client/edge'
#14 1.250 const prisma = new PrismaClient()
#14 1.250 ```
#14 1.250
#14 1.250 See other ways of importing Prisma Client: http://pris.ly/d/importing-client
#14 1.250
#14 1.250 ┌─────────────────────────────────────────────────────────────┐
#14 1.250 │  Deploying your app to serverless or edge functions?        │
#14 1.250 │  Try Prisma Accelerate for connection pooling and caching.  │
#14 1.250 │  https://pris.ly/cli/accelerate                             │
#14 1.250 └─────────────────────────────────────────────────────────────┘
#14 1.250
#14 1.429 Exited
```

</details>

### Further issues
There's still one issue with running `bunx prisma` that this PR does not address. It seems `bunx` does not resolve symlinks correctly. The "executable" (`node_modules/.bin/prisma`) is a symlink to a js file (`node_modules/prisma/build/index.js`) that makes this check:
```ts
if (eval("require.main === module")) {
   // ... actually run
}
```

but this does not return true due to the symlink. I checked `Bun.main` and it returns `node_modules/.bin/prisma`, when I believe it should resolve to the symlinked path (`node_modules/prisma/build/index.js`).  Since these don't match, it fails.

It's still possible to "manually" invoke the CLI without node now though, via:
```bash
bun run --bun ./node_modules/prisma/build/index.js generate
```

which gets around the symlink issue. For whatever reason, it also seems that:
```bash
# Replace #!/usr/bin/env node with #!/usr/bin/env bun
RUN sed -i '1s/node/bun/' ./node_modules/prisma/build/index.js
RUN bunx prisma generate
```

works as well, although I'm not sure why. Perhaps the symlink is only an issue when bun is running in `node` mode.

Without doing one of the two things above, `bunx prisma` commands will still silently fail (though for a different reason)

### DIFF
--- a/src/js/node/stream.js
+++ b/src/js/node/stream.js
@@ -2508,14 +2508,17 @@ var require_readable = __commonJS({
 
     var { addAbortSignal } = require_add_abort_signal();
     var eos = require_end_of_stream();
-    const { maybeReadMore: _maybeReadMore, resume, emitReadable: _emitReadable, onEofChunk } = $lazy("bun:stream");
+    const { maybeReadMore: _maybeReadMore, resume, emitReadable: _emitReadable } = $lazy("bun:stream");
     function maybeReadMore(stream, state) {
       process.nextTick(_maybeReadMore, stream, state);
     }
-    // REVERT ME
     function emitReadable(stream, state) {
       $debug("NativeReadable - emitReadable", stream.__id);
-      _emitReadable(stream, state);
+      state.needReadable = false;
+      if (!state.emittedReadable) {
+        state.emittedReadable = true;
+        process.nextTick(_emitReadable, stream, state);
+      }
     }
     var destroyImpl = require_destroy();
     var {
@@ -2623,6 +2626,24 @@ var require_readable = __commonJS({
       }
       $debug("about to maybereadmore");
       maybeReadMore(stream, state);
+    }
+    function onEofChunk(stream, state) {
+      if (state.ended) return;
+      if (state.decoder) {
+        const chunk = state.decoder.end();
+        if (chunk && chunk.length) {
+          state.buffer.push(chunk);
+          state.length += state.objectMode ? 1 : chunk.length;
+        }
+      }
+      state.ended = true;
+      if (state.sync) {
+        emitReadable(stream, state);
+      } else {
+        state.needReadable = false;
+        state.emittedReadable = true;
+        _emitReadable(stream, state);
+      }
     }
     Readable.prototype.isPaused = function () {
       const state = this._readableState;
--- a/src/js/node/stream.js
+++ b/src/js/node/stream.js
@@ -2629,11 +2629,14 @@ var require_readable = __commonJS({
     }
     function onEofChunk(stream, state) {
       if (state.ended) return;
-      if (state.decoder) {
-        const chunk = state.decoder.end();
-        if (chunk && chunk.length) {
+
+      const decoder = state.decoder;
+      if (decoder) {
+        const chunk = decoder.end();
+        const chunkLength = chunk?.length;
+        if (chunkLength) {
           state.buffer.push(chunk);
-          state.length += state.objectMode ? 1 : chunk.length;
+          state.length += state.objectMode ? 1 : chunkLength;
         }
       }
       state.ended = true;
--- a/src/bun.js/bindings/JSReadableHelper.cpp
+++ b/src/bun.js/bindings/JSReadableHelper.cpp
@@ -175,6 +175,7 @@ EncodedJSValue emitReadable_(JSGlobalObject* lexicalGlobalObject, JSObject* stre
     VM& vm = lexicalGlobalObject->vm();
     auto throwScope = DECLARE_THROW_SCOPE(vm);
     JSValue errored = state->m_errored.get();
+
     if (!state->getBool(JSReadableState::destroyed) && !errored.toBoolean(lexicalGlobalObject) && (state->m_length || state->getBool(JSReadableState::ended))) {
         // stream.emit('readable')
         auto clientData = WebCore::clientData(vm);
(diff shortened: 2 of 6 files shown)


src/js/node/stream.js
```diff
@@ -2624,6 +2627,24 @@ var require_readable = __commonJS({
       $debug("about to maybereadmore");
       maybeReadMore(stream, state);
     }
+    function onEofChunk(stream, state) {
```
Jarred-Sumner, line 2630: In JS, property accesses are slower than variable declarations - so lets only do these once

```suggestion
    function onEofChunk(stream, state) {
     if (state.ended) return;
     const decoder = state.decoder;
      if (decoder) {
        const chunk = decoder.end();
        const chunkLength = chunk?.length;
        if (chunkLength) {
          state.buffer.push(chunk);
          state.length += state.objectMode ? 1 : chunkLength;
        }
      }
      state.ended = true;
      if (state.sync) {
        emitReadable(stream, state);
      } else {
        state.needReadable = false;
        state.emittedReadable = true;
        _emitReadable(stream, state);
      }
    }
```

Jarred-Sumner:
This is great. We really appreciate your PRs. I have one nitpick

Jarred-Sumner:
> There's still one issue with running bunx prisma that this PR does not address. It seems bunx does not resolve symlinks correctly. The "executable" (node_modules/.bin/prisma) is a symlink to a js file (node_modules/prisma/build/index.js) that makes this check:

Addressed in #9105

camero2734:
> > There's still one issue with running bunx prisma that this PR does not address. It seems bunx does not resolve symlinks correctly. The "executable" (node_modules/.bin/prisma) is a symlink to a js file (node_modules/prisma/build/index.js) that makes this check:
> 
> Addressed in #9105 

Awesome!

PR #9105, opened 2024-02-25
Make `Bun.main` the resolved filesystem path

Jarred-Sumner:
### What does this PR do?

This makes `eval('require.main === module')` when the main module is a symlink

Need to write a test


paperclover (approved):
oh this is great. we can now do a bit less work with forcing an absolute file path in the startup cli (it was very silly that we needed to do that)

Issue #8990, opened 2024-02-19
support importing react & react-dom with `react-server` condition


subtleGradient:
### What is the problem this feature would solve?

building a new library for React Server Components

### What is the feature you are proposing to solve the problem?

support runtime `--conditions` flag like node.js for custom conditional exports
https://nodejs.org/api/cli.html#-c-condition---conditionscondition

> You must configure Node.js using the `--conditions react-server` flag
https://github.com/facebook/react/blob/59831c98cffe0edf706238b067928e7cf54d1065/packages/react-server-dom-esm/server.js#L12

### What alternatives have you considered?

Only way I found to import it is to reach into the node_modules directly. But this doesn't work for requires that happen inside.

Importing `react/react.react-server.js` kindof works:

```ts
import type * as ReactType from "react/canary"
// @ts-expect-error -- couldn't figure out how to pass --conditions to bun
const React = (await import("../node_modules/react/react.react-server.js")) as typeof ReactType
if (!("__SECRET_SERVER_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED" in React))
  throw new Error("expected the React build with server internals")
```

Importing `react-dom.react-server.js` has problems:

```ts
import type * as ReactDOMType from "react-dom/canary"
// @ts-expect-error -- couldn't figure out how to pass --conditions to bun
const ReactDOM = (await import("../node_modules/react-dom/react-dom.react-server.js")) as typeof ReactDOMType
// TODO: figure out how to verify that ReactDOM is the server build
// TODO: figure out how to make sure that ReactDOM is not requiring a different React file
```

PROBLEM: `react-dom.react-server.js` has `require('react')`, which then imports the default build of react instead of the react.react-server.js build :(

paperclover:
I think adding this as a runtime + bundler flag would make sense, and not really be that hard either.

for someone wanting to implement this:

`options.zig` has this code, which initializes the conditions list by default based on --target
```zig
opts.conditions = try ESMConditions.init(allocator, Target.DefaultConditions.get(opts.target));
```

some work on this struct can be added to allow appending onto this list after the default conditions list is created.

igorwessel:
@paperdave I tried to find test files for this specific part of "exports"/"imports" but I didn't find anything specific. Can you give me some direction?

Nevermind, I think I found the test.

[...]

PR #9106, opened 2024-02-25
feat: add support --conditions flag

igorwessel:
Closes https://github.com/oven-sh/bun/issues/8990

### What does this PR do?

This adds a new flag --conditions to bun run and build. When set, it will try to resolve custom conditions in package.json exports. Example:

In some NPM package have this package.json with a custom condition:

```json
{
  "name": "package",
  "exports": {
    "some-custom-condition": {
      "import": "./index.mjs",
      "require": "./index.cjs",
      "default": "./index.mjs"
    }
   "default": "./throw-error.mjs"
  }
}
```

We have this JS file importing our "package":
```js
import foo from 'package' // foo is a var with "Hello world!"

console.log(foo);
```

If we run it without passing the condition it should fall into the "throw-error.mjs" fallback because is the default, otherwise should resolves for ESM module or CJS module. Example:
```shell
bun --conditions some-custom-condition ./index.js

// prints "Hello world!"
```

The same behavior should happen in the bundle, resolving to the correct files according to the condition. Example:
```shell
bun build ./index.js --outfile ./build.js --conditions some-custom-condition
```

Our generate code should have the content for resolved custom condition export. Example for build
```js
// node_modules/package/index.js
var $foo = "Hello world!";

// src/index.js
console.log($foo);
```

- [ ] Code changes

I wrote automated tests, and create a simple "package" to test the runtime and bundler behavior.

### DIFF
--- a/src/api/schema.zig
+++ b/src/api/schema.zig
@@ -1755,6 +1755,9 @@ pub const Api = struct {
         /// source_map
         source_map: ?SourceMapMode = null,
 
+        /// conditions
+        conditions: []const []const u8,
+
         pub fn decode(reader: anytype) anyerror!TransformOptions {
             var this = std.mem.zeroes(TransformOptions);
 
@@ -1838,6 +1841,9 @@ pub const Api = struct {
                     },
                     25 => {
                         this.source_map = try reader.readValue(SourceMapMode);
+                    },
+                    26 => {
+                        this.conditions = try reader.readArray([]const u8);
                     },
                     else => {
                         return error.InvalidMessage;
@@ -1948,6 +1954,12 @@ pub const Api = struct {
                 try writer.writeFieldID(25);
                 try writer.writeEnum(source_map);
             }
+
+            if (this.conditions) |conditions| {
+                try writer.writeFieldID(26);
+                try writer.writeArray([]const u8, conditions);
+            }
+
             try writer.endMessage();
         }
     };
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -182,7 +182,7 @@ pub const Arguments = struct {
         clap.parseParam("--prefer-latest                   Use the latest matching versions of packages in the Bun runtime, always checking npm") catch unreachable,
         clap.parseParam("-p, --port <STR>                  Set the default port for Bun.serve") catch unreachable,
         clap.parseParam("-u, --origin <STR>") catch unreachable,
-        clap.parseParam("-c, --conditions <STR>...            Allow pass custom conditions to resolve.") catch unreachable,
+        clap.parseParam("--conditions <STR>...             Allow pass custom conditions to resolve.") catch unreachable,
     };
 
     const auto_only_params = [_]ParamType{
@@ -230,7 +230,7 @@ pub const Arguments = struct {
         clap.parseParam("--minify-whitespace              Minify whitespace") catch unreachable,
         clap.parseParam("--minify-identifiers             Minify identifiers") catch unreachable,
         clap.parseParam("--dump-environment-variables") catch unreachable,
-        clap.parseParam("-c, --conditions <STR>...        Allow pass custom conditions to resolve.") catch unreachable,
+        clap.parseParam("--conditions <STR>...            Allow pass custom conditions to resolve.") catch unreachable,
     };
     pub const build_params = build_only_params ++ transpiler_params_ ++ base_params_;
 
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -182,6 +182,7 @@ pub const Arguments = struct {
         clap.parseParam("--prefer-latest                   Use the latest matching versions of packages in the Bun runtime, always checking npm") catch unreachable,
         clap.parseParam("-p, --port <STR>                  Set the default port for Bun.serve") catch unreachable,
         clap.parseParam("-u, --origin <STR>") catch unreachable,
+        clap.parseParam("-c, --conditions <STR>...            Allow pass custom conditions to resolve.") catch unreachable,
     };
 
     const auto_only_params = [_]ParamType{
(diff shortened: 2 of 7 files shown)


src/api/schema.zig
igorwessel, line 1: I'm not sure here is the best way to set. I feel like this should be somewhere else related to the bundler options than the transform options. But I couldn't identify it. :/

igorwessel, line 1: I'm unable to find a command to "auto-generated" for JS/TS related to schema. Does it have to make manually?

paperclover: this "schema" thing is a holdover from very long ago when bun was just a wasm thing, not a full js runtime/package manager.

the code generator doesnt even work anymore, this is manually edited now. what you did is fine.

src/options.zig
```diff
@@ -1685,6 +1693,10 @@ pub const BundleOptions = struct {
 
         opts.conditions = try ESMConditions.init(allocator, Target.DefaultConditions.get(opts.target));
 
+        if (transform.conditions.len > 0) {
+            opts.conditions.append(transform.conditions) catch unreachable;
```
paperclover:
```suggestion
            opts.conditions.append(transform.conditions) catch bun.outOfMemory();
```
i know we already do this but `unreachable` on a memory allocation is dangerous.

src/options.zig
```diff
@@ -932,6 +932,14 @@ pub const ESMConditions = struct {
             .require = require_condition_map,
         };
     }
+
+    pub fn append(self: *ESMConditions, customConditions: []const string) !void {
+        for (customConditions) |condition| {
+            self.default.putAssumeCapacityNoClobber(condition, {});
```
paperclover, line 942: this will crash if there are not enough items in the list.

use `try self.[list].ensureUnusedCapacity(customConditions.len);`

nit: zig's convention for variable names is snake case, though not followed everywhere. use `custom_conditions`. the variable could also just be called `conditions`

src/options.zig
```diff
@@ -932,6 +932,14 @@ pub const ESMConditions = struct {
             .require = require_condition_map,
         };
     }
+
+    pub fn append(self: *ESMConditions, customConditions: []const string) !void {
```
paperclover: can this be named appendSlice? to not be confused with zig std.ArrayList(T).append, which is a single item append.

src/cli.zig
```diff
@@ -229,6 +230,7 @@ pub const Arguments = struct {
         clap.parseParam("--minify-whitespace              Minify whitespace") catch unreachable,
         clap.parseParam("--minify-identifiers             Minify identifiers") catch unreachable,
         clap.parseParam("--dump-environment-variables") catch unreachable,
+        clap.parseParam("-c, --conditions <STR>...        Allow pass custom conditions to resolve.") catch unreachable,
```
paperclover: i am uncertain on if `-c` should be an alias. not opinionated on this. maybe someone else is.

paperclover (changes requested):
nice to see this added. some notes that must be addressed:

src/cli.zig
igorwessel: I also had doubts but I chose to follow the Node API (https://nodejs.org/api/cli.html#-c-condition---conditionscondition), taking a look at esbuild they don't have aliases. (https://esbuild.github.io/api/#conditions)

paperclover: thanks to jarred for pointing this out. we already have `-c`

image

surprised your code compiled.

paperclover (changes requested):
thanks for fixing the memory mistakes. one more review cycle.

- we should remove the `-c`, as there is already `-c` for config

- it would be nice to have more tests. to be clear i think your code is correct, this is just to prevent regressions.
  - passing two or more conditions (and the ordering of them is what is expected)
  - does it work in `bun build` (it should)

igorwessel:
> thanks for fixing the memory mistakes. one more review cycle.
> 
>     * we should remove the `-c`, as there is already `-c` for config
> 
>     * it would be nice to have more tests. to be clear i think your code is correct, this is just to prevent regressions.
>       
>       * passing two or more conditions (and the ordering of them is what is expected)
>       * does it work in `bun build` (it should)

Do you think there should be more testing for bundler? I saw that there was a test ported from esbuild, so I didn't think much about creating more cases. @paperdave

paperclover (approved):
i think it looks good

src/cli.zig
```diff
@@ -182,6 +182,7 @@ pub const Arguments = struct {
         clap.parseParam("--prefer-latest                   Use the latest matching versions of packages in the Bun runtime, always checking npm") catch unreachable,
         clap.parseParam("-p, --port <STR>                  Set the default port for Bun.serve") catch unreachable,
         clap.parseParam("-u, --origin <STR>") catch unreachable,
+        clap.parseParam("--conditions <STR>...             Allow pass custom conditions to resolve.") catch unreachable,
```
gvilums:
```suggestion
        clap.parseParam("--conditions <STR>...             Pass custom conditions to resolve") catch unreachable,
```

src/cli.zig
```diff
@@ -229,6 +230,7 @@ pub const Arguments = struct {
         clap.parseParam("--minify-whitespace              Minify whitespace") catch unreachable,
         clap.parseParam("--minify-identifiers             Minify identifiers") catch unreachable,
         clap.parseParam("--dump-environment-variables") catch unreachable,
+        clap.parseParam("--conditions <STR>...            Allow pass custom conditions to resolve.") catch unreachable,
```
gvilums:
```suggestion
        clap.parseParam("--conditions <STR>...            Pass custom conditions to resolve") catch unreachable,
```

gvilums (changes requested):
We can merge this once these grammar mistakes are fixed. Thanks for the contribution!

src/cli.zig
igorwessel: thanks!

src/cli.zig
igorwessel: thanks!

gvilums (approved):

PR #9125, opened 2024-02-26
fix(bun:test): fix `toContainKeys` with undefined and null

### LINKED ISSUE #9118: Segfault when using undefined in toContainKeys matcher

1.0.29+a146856d1

Run the following test 

````ts

describe("Base", () => {
  it("contains key segfaults with undefined", () => {
    expect(undefined).toContainKeys(["id"]);
  });
});
````

Test to fail

### What do you see instead?

Test runner crashes with segfault

````
test/index.test.ts:
Segmentation fault
````

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by dylan-conway**
### What does this PR do?
fixes #9118 

added two tests

### CODE REVIEW

File: src/bun.js/bindings/bindings.cpp [hunk 1]
```diff
@@ -2497,9 +2497,13 @@ JSC__JSValue JSC__JSValue__keys(JSC__JSGlobalObject* globalObject, JSC__JSValue
 bool JSC__JSValue__hasOwnProperty(JSC__JSValue jsValue, JSC__JSGlobalObject* globalObject, ZigString key)
 {
     JSC::VM& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
 
     JSC::JSValue value = JSC::JSValue::decode(jsValue);
-    return value.toObject(globalObject)->hasOwnProperty(globalObject, JSC::PropertyName(JSC::Identifier::fromString(vm, Zig::toString(key))));
+    JSObject* obj = value.toObject(globalObject);
```
Jarred-Sumner, line 2503: This potentially coerces, it should be `!isObject` and return false

Jarred-Sumner:
```suggestion
     if (!value.isObject()) return false;
    JSObject* obj = value.getObject();
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -2497,9 +2497,13 @@ JSC__JSValue JSC__JSValue__keys(JSC__JSGlobalObject* globalObject, JSC__JSValue
...
+    RETURN_IF_EXCEPTION(scope, false);
+
+    RELEASE_AND_RETURN(scope, obj->hasOwnProperty(globalObject, JSC::PropertyName(JSC::Identifier::fromString(vm, Zig::toString(key)))));
```
Jarred-Sumner: ->hasOwnProperty can throw when its a proxy and I think its somewhat ambiguous whether it would throw here

src/bun.js/bindings/bindings.cpp
dylan-conway: we need to coerce because given an array or string the indexes should become properties

src/bun.js/bindings/bindings.cpp
dylan-conway: yeah, I think in this case it shouldn't catch the exception because `toContainKey(s)` uses `hasOwnProperty` directly, but other internal uses of this function will probably want to catch it

jdalton: If it was a proxy it would cause `Object.hasOwn` to throw and `.hasOwnProperty('x')` to throw. I think it's fine to blow up here too.. we can do what WebKit does and 

https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/runtime/ObjectConstructor.cpp#L1368-L1377
then goes to:
https://github.com/WebKit/WebKit/blob/e954cf3227f1d0672e1d61645e37400059d29e58/Source/JavaScriptCore/runtime/ObjectPrototype.cpp#L103-L121

Which does capture the ` thisObject->hasOwnProperty` in a throw scope. It also leverages a cache which is neeat.

jdalton: The JS implementation does an odd check,  if the object has a truthy `.hasOwnProperty` (so not a null prototype) and then calls`Object.prototype.hasOwnProperty` so works with a `null` prototype. All instead using `Object.hasOwn`. We could PR that. https://github.com/jest-community/jest-extended/blob/711fdcc54d68c2b2c1992c7cfbdf0d0bd6be0f4d/src/matchers/toContainKeys.js#L1-L6

jdalton: The JS implementation allows `falsey` to pass through meaning they don't allow things like `''` or `0` to go through. I'm guessing their intention is to really only support objects but because JS is wiggly other stuff like strings is doable.

src/bun.js/bindings/bindings.cpp
```diff
@@ -2497,9 +2497,13 @@ JSC__JSValue JSC__JSValue__keys(JSC__JSGlobalObject* globalObject, JSC__JSValue
 bool JSC__JSValue__hasOwnProperty(JSC__JSValue jsValue, JSC__JSGlobalObject* globalObject, ZigString key)
 {
     JSC::VM& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
```
jdalton, line 2500: ~~`hasOwnProperty` should be a separate concern from `toContainKeys`. We should use `hasOwnProperty` or equiv but not make the changes within it directly.~~

Jarred-Sumner: ~I think the jest implementation of `toContainKeys` calls `hasOwnProperty`~

ah you linked to it earlier

src/bun.js/bindings/bindings.cpp
dylan-conway: > Which does capture the ` thisObject->hasOwnProperty` in a throw scope. It also leverages a cache which is neeat.

oh nice, yeah we should use `objectPrototypeHasOwnProperty()` for the cache

dylan-conway: > The JS implementation does an odd check, if the object has a truthy `.hasOwnProperty` (so not a null prototype) and then calls`Object.prototype.hasOwnProperty` so works with a `null` prototype. All instead using `Object.hasOwn`. We could PR that. https://github.com/jest-community/jest-extended/blob/711fdcc54d68c2b2c1992c7cfbdf0d0bd6be0f4d/src/matchers/toContainKeys.js#L1-L6

yep, will update to check for this before `hasOwnProperty` else it will throw when it shouldn't

test/js/bun/test/expect.test.js
```diff
@@ -2421,6 +2424,13 @@ describe("expect()", () => {
     expect({ a: "foo", 1: "test" }).toContainKeys(["a", 1]);
     expect({ a: "foo", b: "bar", c: "baz" }).not.toContainKeys(["a", "b", "e"]);
     expect({ a: "foo", b: "bar", c: "baz" }).not.toContainKeys(["z"]);
+
```
jdalton, line 2427: We should add a test for `'', false` too.

test/js/bun/test/expect.test.js
```diff
@@ -2421,6 +2424,13 @@ describe("expect()", () => {
...
+    expect(() => {
+      expect(undefined).toContainKeys(["id"]);
+    }).toThrow();
+    expect(() => {
+      expect(null).toContainKeys(["id"]);
+    }).toThrow();
```
jdalton: We could detect the actual message as the error could be anything.
```js
let m = '';
try {
    expect(null).toContainKeys(['a', 'b']);
} catch (e) { 
    m = `${e?.message}`;
}
expect(stripAnsi(m)).toMatch(new RegExp(`Received:\\s+${v}`));
```

src/bun.js/bindings/bindings.cpp
jdalton: I see now that our `hadOwnPropertyValue` is really more generic for the JSValue which is less solid around type (ex: can detect isString but not length) so is more forgiving.

jdalton (approved):
If the tests pass it LGTM :)

PR #9127, opened 2024-02-27
Replace `std.unicode.fmtUtf16le` with `bun.fmt.utf16`

Jarred-Sumner:
### What does this PR do?

We should generally avoid using std.unicode.* at runtime in Bun.

This:
- Renames `bun.fmt.formatUTF16` -> `bun.fmt.utf16` to make it easier to find/type
- Replaces usages of `std.unicode.fmtUtf16Le` with `bun.fmt.utf16`

### DIFF
--- a/src/fmt.zig
+++ b/src/fmt.zig
@@ -79,7 +79,7 @@ pub fn formatUTF16TypeEscapeBackslashes(comptime Slice: type, slice_: Slice, wri
     }
 }
 
-pub fn formatUTF16(slice_: []align(1) const u16, writer: anytype) !void {
+pub fn utf16(slice_: []align(1) const u16, writer: anytype) !void {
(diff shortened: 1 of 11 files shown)


src/fmt.zig
```diff
@@ -79,8 +79,8 @@ pub fn formatUTF16TypeEscapeBackslashes(comptime Slice: type, slice_: Slice, wri
     }
 }
 
-pub fn formatUTF16(slice_: []align(1) const u16, writer: anytype) !void {
-    return formatUTF16Type([]align(1) const u16, slice_, writer);
+pub inline fn utf16(slice_: []const u16) FormatUTF16 {
```
dylan-conway, line 82: `fmtUTF16` also exists and is the same thing. might as well replace it

dylan-conway (approved):

PR #9129, opened 2024-02-27
feat: signal handling on windows

paperclover:
### What does this PR do?

you can use on("SIGINT") on windows. works in worker, cleanup, etc etc

this also adds `jsvalue.getClassInfoName()`, which is used to print a better error message when you return a non-object cell type in a macro.


src/bun.js/bindings/BunProcess.cpp
```diff
@@ -655,121 +732,171 @@ static void onDidChangeListeners(EventEmitter& eventEmitter, const Identifier& e
...
     if (!signalToContextIdsMap) {
-        signalToContextIdsMap = new HashMap<int, HashSet<uint32_t>>();
+        signalToContextIdsMap = new HashMap<int, SignalHandleValue>();
     }
 
     if (auto signalNumber = signalNameToNumberMap->get(eventName.string())) {
+#if !OS(WINDOWS)
         if (signalNumber != SIGKILL && signalNumber != SIGSTOP) {
+#else
+        if (signalNumber != SIGKILL) {
```
dylan-conway:
```suggestion
        if (signalNumber != SIGKILL && signalNumber != SIGSTOP) {
```
`SIGSTOP` should be ignored, it's not possible to have a handler for it

dylan-conway: oh wait, this is for windows

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -592,42 +598,113 @@ static void loadSignalNumberMap()
...
 #ifndef SIGINFO
         signalNameToNumberMap->add(signalNames[29], 255);
 #endif
+#ifdef SIGSYS
         signalNameToNumberMap->add(signalNames[30], SIGSYS);
+#endif
     });
 }
 
+extern "C" uv_signal_t* Bun__UVSignalHandle__init(JSC::JSGlobalObject* lexicalGlobalObject, int signalNumber, void (*callback)(uv_signal_t*, int));
```
dylan-conway, line 683: breaking build, needs to be wrapped in `if OS(WINDOWS)`

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -655,121 +732,171 @@ static void onDidChangeListeners(EventEmitter& eventEmitter, const Identifier& e
...
             if (isAdded) {
                 if (!signalToContextIdsMap->contains(signalNumber)) {
-                    HashSet<uint32_t> contextIds;
-                    contextIds.add(contextId);
-                    signalToContextIdsMap->set(signalNumber, contextIds);
-
-                    lock.unlockEarly();
-
+                    SignalHandleValue signal_handle = {
+                        .handle = nullptr,
```
dylan-conway: also needs `#if OS(WINDOWS) #else` between these fields

src/bun.js/bindings/exports.zig
```diff
@@ -926,3 +926,37 @@ comptime {
...
 }
+
+const libuv = bun.windows.libuv;
+
+pub export fn Bun__UVSignalHandle__init(
+    global: *JSC.JSGlobalObject,
+    signal_num: i32,
+    callback: *const fn (sig: *libuv.uv_signal_t, num: c_int) callconv(.C) void,
+) callconv(.C) ?*libuv.uv_signal_t {
+    const signal = bun.new(libuv.uv_signal_t, undefined);
```
dylan-conway: also probably need a `if (comptime !Environment.isWindows)` here

src/bun.js/WebKit
dylan-conway, line 1: is this change intentional?

src/install/install.zig
```diff
@@ -1378,8 +1378,8 @@ pub const PackageInstall = struct {
 
         defer subdir.close();
 
-        var buf: bun.windows.WPathBuffer = undefined;
-        var buf2: bun.windows.WPathBuffer = undefined;
+        var buf: bun.WPathBuffer = undefined;
```
dylan-conway: the buffer is only used on windows, why change from `bun.windows.WPathBuffer`?

src/bun.js/WebKit
paperclover: no

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -592,42 +598,115 @@ static void loadSignalNumberMap()
         signalNameToNumberMap->add(signalNames[1], SIGINT);
         signalNameToNumberMap->add(signalNames[2], SIGQUIT);
         signalNameToNumberMap->add(signalNames[3], SIGILL);
+#ifdef SIGTRAP
         signalNameToNumberMap->add(signalNames[4], SIGTRAP);
```
Jarred-Sumner, line 607: maybe we should use one of those FOR_EACH_MACRO things here?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -655,121 +734,173 @@ static void onDidChangeListeners(EventEmitter& eventEmitter, const Identifier& e
...
 
                     // Clear the sa_mask
                     sigemptyset(&action.sa_mask);
                     sigaddset(&action.sa_mask, signalNumber);
                     action.sa_flags = SA_RESTART;
 
                     sigaction(signalNumber, &action, nullptr);
+#else
+                    // To be more consistant, make sure the loop used is always the main thread's loop
+                    if (eventEmitter.scriptExecutionContext()->isMainThread()) {
```
Jarred-Sumner: Why though? Was there a race condition here?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -655,121 +734,173 @@ static void onDidChangeListeners(EventEmitter& eventEmitter, const Identifier& e
...
+                        signal_handle.handle = Bun__UVSignalHandle__init(
```
Jarred-Sumner, line 853: Can you make this a field on the class which is constructed/destructed, instead of manually initialized and uninitialized? Since we're in C++ we can do that and this code will be duplicated less + safer

src/bun.js/bindings/BunProcess.cpp
paperclover: i'm not 100% sure how this would look, as you cannot nest `#if` and other preprocessor stuff into a macro definition.

src/bun.js/bindings/BunProcess.cpp
paperclover: i need to double check my suspicion is true, as i did not do much testing.

in theory:

- start a worker
- add an event listener on SIGINT
- then, add an event listener on main thread
- kill the worker
- emit a control c
- the uv loop owning the ctrl c was the first context to add it, but since it is now closed, the handle to the signal

if this doesnt reproduce without the threadsafety check, then i'll simplify it.

though, going over this again, it does really feel like this should not use a global singleton, but it would be much more code to fully reinvent the process signal events than to just modify what we have so far.

src/bun.js/bindings/BunProcess.cpp
paperclover: sure.

i tend to avoid using automatic destruction, especially when the lifetime is longer than a single block.

paperclover:
signal pr now segfaults. dont merge

paperclover:
removed cross thread signal handling, since node.js doesnt allow it and it also doesn't make sense. this fixes a possible way to hang.

dylan-conway (approved):
Looks good. One `std.debug.print` needs to be removed

PR #9137, opened 2024-02-28
refactor: remove unnecessary spread

zieka:
### What does this PR do?
removes an unnecessary spread

- [x] Code changes
  - just in misctools

`bun ./misctools/mime.js` ensured output pre and post change were the same

### DIFF
--- a/misctools/mime.js
+++ b/misctools/mime.js
@@ -21,7 +21,7 @@ const withExtensions = [
         return !!json[key]?.extensions?.length;
       })
       .flatMap(mime => {
-        return [...new Set([...json[mime].extensions])].map(ext => {
+        return [...new Set(json[mime].extensions)].map(ext => {
           return [`.{.@"${ext}", all.@"${mime}"}`];
         });
       })


misctools/mime.js
```diff
@@ -21,7 +21,7 @@ const withExtensions = [
         return !!json[key]?.extensions?.length;
       })
       .flatMap(mime => {
-        return [...new Set([...json[mime].extensions])].map(ext => {
```
jdalton, line 24: This is to `_.uniq` the values of `json[mime].extensions` (not present in all datasets but filtered above). It looks like the dataset from `https://raw.githubusercontent.com/jshttp/mime-db/master/db.json` does not contain duplicates anyways so the entire thing could be simplified to:
```js
const withExtensions = [];
{
  const mimes = Object.keys(json);
  for (let i = 0, { length: mimesLen } = mimes; i < mimesLen; i += 1) {
    const mime = mimes[i];
    const exts = json[mime]?.extensions;
    for (let j = 0, extsLen = exts?.length || 0; j < extsLen; j += 1) {
      withExtensions.push(`.{.@"${exts[j]}", all.@"${mime}"}`);
    }
  }
}
```
Which is 11x faster too.
11x faster

zieka: true
I was just removing the unnecessary spread as I just found it annoying; clearly I keep the uniqueness and sorting assuming they were useful.
If uniqueness and sorting are not necessary your suggestion is quite awesome.
Any bun benchmarking utilities you recommend?

jdalton: _RE the screenshot edited out: Microbenmarks get grumpy about closures_

Issue #9119, opened 2024-02-26
Bun.password.needsRehash


nikeee:
### What is the problem this feature would solve?

Determining if a password hash still has all the parameters that are deemed safe.

### What is the feature you are proposing to solve the problem?

`Bun.password` looks a lot like PHP's `password_hash` and `password_verify`. PHP also has `password_needs_rehash` that can be used to build a new hash if the specs changed (for example, because they are now unsafe or their parameters changed):

It works like this:

```php
$password = 'rasmuslerdorf';
$hash = '$2y$10$YCFsG6elYca568hBi2pZ0.3LDL5wjgxct1N8w/oLR/jfHsiQwCqTS';
$algorithm = PASSWORD_BCRYPT;
$options = ['cost' => 12];

// Verify stored hash against plain-text password
if (password_verify($password, $hash)) {
    // Check if either the algorithm or the options have changed
    if (password_needs_rehash($hash, $algorithm, $options)) {
        // If so, create a new hash, and replace the old one
        $newHash = password_hash($password, $algorithm, $options);

        // Update the user record with the $newHash
    }

    // Perform the login.
}
```

So in Bun world, this would be something like this:

```js
const password = 'rasmuslerdorf';
const hash = '$argon2id$v=19$m=65536,t=2,p=1$tFq+9AVr1bfPxQdh6E8DQRhEXg/M/SqYCNu6gVdRRNs$GzJ8PuBi+K+BVojzPfS5mjnC8OpLGtv8KJqF99eP6a4';

const latestHashOptions = {
  algorithm: "argon2id", // "argon2id" | "argon2i" | "argon2d"
  memoryCost: 4, // memory usage in kibibytes
  timeCost: 3, // the number of iterations
};

if (await Bun.password.verify(password, hash)) {
    
    if (Bun.password.needsRehash(hash, latestHashOptions)) {

        // If so, create a new hash, and replace the old one
        const newHash = await Bun.password.hash(password, latestHashOptions);
        // Update the user record with the newHash
    }

    // Perform the login.
}
```

Issue #9078, opened 2024-02-23
Bun.password.verify bcrypt interoperability (accept $2$, $2a$, $2c$, $2x$, $2y$)


mantissa7:
### What version of Bun is running?

1.0.28

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

Bun.password.verify using bcrypt does not work well with hashes created by other libraries. Creating a hash using Bcrypt.Net-Next for example returns false when used as the parameter for `bun.password.verify(password, dotnet_bcrypt_hash)`, however using a hash generated by Bun.password.hash is verified successfully by Bcrypt.Net-Next.

Manually substituting $2a$  for $2b$ fixes the issue. 

```
const password = 'test123'
const hash = '$2a$06$wouz5xNIhZrkWBw7LdMKFu.BBH76dGWsn0IZ06t47nOFac9GIgDB2' // select crypt('test123', gen_salt('bf')) as hash;

const isMatch = await Bun.password.verify(password, hash, 'bcrypt'); 
// false

const fixed_hash = hash.replace(/^\$2a\$/, '$2b$');
const isMatchNow = await Bun.password.verify(password, fixed_hash, 'bcrypt'); 
// true
```

### What is the expected behavior?

Bun.password.verify should return true for a correct password-hash comparison generated by other libraries.

### Additional information

Discord #help reference:
https://discord.com/channels/876711213126520882/1209423413282869249/1209423413282869249

argosphil:
I'm not sure whether there's an exhaustive list of prefixes that we must recognize, or whether it's everything matching `2.?`.

In any case, we can fix this in the Zig code (and maybe it's already been fixed and we can close this after our next Zig update), or work around it in Bun. Which one's better?

sequencerr:
> I'm not sure whether there's an exhaustive list of prefixes that we must recognize, or whether it's everything matching 2.?.

https://en.wikipedia.org/wiki/Bcrypt#Versioning_history

https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt.html#deviations
There is something about `The crypt_blowfish 8-bit bug`

Maybe zig lib just don't support old `bcrypt` versions?

sequencerr:
> does not work well with hashes created by other libraries.

does not work with other versions*

argosphil:
[...]

PR #9139, opened 2024-02-28
bcrypt: explicitly choose what to do when encountering long passwords

argosphil:
### What does this PR do?

A more robust solution for the problem raised in #9009 .

This changes the `hash` and `verify` functions in `Bun.password`, as well as their sync variants, to accept an algorithm specifier of the form

```
{
  algorithm: "bcrypt",
  longPassword: "sha512" | "truncate" | "throw",
}
```

to choose between different variants of the bcrypt algorithm. Bcrypt is limited to 72 bytes of password length. Most implementations silently truncate at that length, which is a security problem. Bun uses a pre-hash sha512 step to avoid the problem, but only for long passwords, which is okay from a security POV but breaks interoperability.

The new option allows choosing between the two variants, and adds a third variant which throws an error upon encountering a password longer than 72 bytes.

Drive-by changes include being able to use an object rather than a string as the third argument to Bun.password.verify[Sync], fixing some typos, and merging `PasswordObject.Algorithm.Value` and `PasswordObject.Algorithm`. As indentation changed, it may be a good idea to exclude whitespace changes when reviewing.

- [x] Code changes

I added tests.

- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be [no memory allocation]
- [ ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed [I believe function args are fine?]

Related issues:
#9078 - we can potentially add another property to accept `$2a$` hashes
#9119 - we can potentially add new properties to specify the minimum acceptable level of hash settings

TODO:
- [ ] I'd like to find a better name than `longPassword`
- [ ] verify this is indeed interoperable with OpenBSD bcrypt, particularly for passwords of byte length 71, 72, 73
- [ ] it would be nice to be able to fix the salt for testing, by adding another option


Jarred-Sumner:
Thank you for this PR , but I'm not convinced the complexity of adding this as an option is warranted. Have people been running into issues with long passwords in bcrypt other than the previous case where it returned false? This feels more like a documentation problem than a code problem

argosphil:
> Thank you for this PR , but I'm not convinced the complexity of adding this as an option is warranted. Have people been running into issues with long passwords in bcrypt other than the previous case where it returned false? This feels more like a documentation problem than a code problem

Thanks for explaining. I don't feel strongly about this one way or the other, to be honest. I would like to point out a few things that ultimately convinced me to open this PR:
- it's only 12 extra lines of code
- it is inaccurate to call something "bcrypt" that is, in fact, a modified variant of it
- the new interface is more consistent - hash and verify take precisely the same algorithm objects/strings
- a number of people seemed to be in favor

People are assuming that our bcrypt is the same as other bcrypts (#9078), and people are using our bcrypt with long passwords (Discord help issue), but not, as far as I know, doing both at the same time.

(@jdalton , @paperdave , @pilcrowOnPaper, @mantissa7 participated in previous discussions, and may have stronger opinions on this, and better knowledge of the subtleties involved, than me)

9Morello:
I appreciate the clearer documentation.
Regarding a better name for `longPassword`, what about `whenPasswordIsTooLong`? It's longer but IMO makes it more explicit what you're doing when you pass `"throw"` or `"truncate"` to it.

pilcrowonpaper:
Would be a breaking change, but might be better to separate bcrypt and bcrypt with hs512?

gvilums:
> > Thank you for this PR , but I'm not convinced the complexity of adding this as an option is warranted. Have people been running into issues with long passwords in bcrypt other than the previous case where it returned false? This feels more like a documentation problem than a code problem
> 
> Thanks for explaining. I don't feel strongly about this one way or the other, to be honest. I would like to point out a few things that ultimately convinced me to open this PR:
> 
> * it's only 12 extra lines of code
> * it is inaccurate to call something "bcrypt" that is, in fact, a modified variant of it
> * the new interface is more consistent - hash and verify take precisely the same algorithm objects/strings
> * a number of people seemed to be in favor
> 
> People are assuming that our bcrypt is the same as other bcrypts (#9078), and people are using our bcrypt with long passwords (Discord help issue), but not, as far as I know, doing both at the same time.
> 
> (@jdalton , @paperdave , @pilcrowOnPaper, @mantissa7 participated in previous discussions, and may have stronger opinions on this, and better knowledge of the subtleties involved, than me)

I think these are good points and I'm in favor of this change. @Jarred-Sumner how do you feel?

packages/bun-types/bun.d.ts
```diff
@@ -1675,7 +1680,8 @@ declare module "bun" {
       /**
        * @default "argon2id"
        *
-       * When using bcrypt, passwords exceeding 72 characters will be SHA256'd before
+       * When using bcrypt, passwords exceeding 72 bytes will be SHA512'd before,
+       * unless algorithm.longPassword is set to "throw" or "truncate".
```
jdalton, line 1692: This feels like a good default as it will still produce valid result and not truncate (silently not be what you want). But I'm torn with it and the truncate behavior since that's the real default.

jdalton:
I like having options ❤️ and this seems like a case folks should be able to be able to be explicit about its handling.

paperclover:
im in favor of the code cleanup and changes overall. there might be a better naming than `longPassword` (i am less happy with `whenPasswordIsTooLong`), but it can work how it is. maybe `longPasswordHandling` or `longInputHandling`, i am not very good at bikeshed.

are there tests for this code that ensure a password generated with an older version of bun is the same. i want to ensure backwards compat here.

cannot merge yet as the zig build is failing.

argosphil:
> im in favor of the code cleanup and changes overall. there might be a better naming than `longPassword` (i am less happy with `whenPasswordIsTooLong`), but it can work how it is. maybe `longPasswordHandling` or `longInputHandling`, i am not very good at bikeshed.

I think "input" is ambiguous as it might refer to the hashed password which is also an input to the `verifySync` function. No objections to `longPasswordHandling`, though. 

> are there tests for this code that ensure a password generated with an older version of bun is the same. i want to ensure backwards compat here.

Unfortunately, the zig API for password hashing doesn't have a "seed" option for reproducible testing. I think that should be added if and when this PR is merged. Adding a few hardcoded testcases wouldn't hurt, though, so I'll look into that.

> cannot merge yet as the zig build is failing.

Will have a look, thanks for the heads-up.

> I like having options ❤️ and this seems like a case folks should be able to be able to be explicit about its handling.

I do think more options make sense in this case, because crypto is subtle. For example, right now it's a bad idea to accept hashed passwords from a source you don't fully trust, as verifying them could be very expensive.

> Would be a breaking change, but might be better to separate bcrypt and bcrypt with hs512?

"bcrypt" vs "buncrypt"? :-)

I don't think that's the right thing to do, to be honest. If we add a new string for every variant of every algorithm we might want to support, things might become pretty unwieldy.

> I appreciate the clearer documentation. Regarding a better name for `longPassword`, what about `whenPasswordIsTooLong`? It's longer but IMO makes it more explicit what you're doing when you pass `"throw"` or `"truncate"` to it.

How do you feel about @paperdave's suggestion?

PR #9140, opened 2024-02-28
resolver: fix debug mode crash in test/bundler/bun-build-api.test.ts

nektro:
path being processed here is normalized here https://github.com/oven-sh/bun/blob/bf3dbda9a2ba25bd1f13e31494626ca7e4279039/src/bun.js/javascript.zig#L1717-L1741 and so will always contain `/` even on windows


Jarred-Sumner:
cc @paperdave

Issue #3216, opened 2023-06-05
Runtime Directory Caching Issue (breaks vite config files)


paperclover:
### What steps can reproduce the bug?

```ts
const file = `${import.meta.dir}/temp.mjs`;

import { existsSync, unlinkSync, writeFileSync } from "fs";

if (existsSync(file)) {
  console.log("temp.mjs cannot exist before running this script");
  unlinkSync(file);
  process.exit(0);
}

writeFileSync(file, "export default 1;");

try {
  const module = await import(file);
  console.log(module.default);
} finally {
  unlinkSync(file);
}
```

### Additional information

Vite blocker

paperclover:
gonna paste this text to make this issue more searchable, this is bug causing vite with config files to crash on startup
```
$ vite dev
failed to load config from vite.config.js
error when starting dev server:
```

paperclover:
the fix doesnt strictly fix this bug, but it does fix the case if you're using a query string in the import path, like `test.mjs?hello`

paperclover:
tbh

i want to fix this for all cases. the file strictly exists and `import()` should be able to catch it

AlbertMarashi:
Related

https://github.com/vitejs/vite/discussions/17851

PR #9144, opened 2024-02-28
fix(windows): fix directory cache regression "expected to end with a trailing slash"

paperclover:
### What does this PR do?

Fixes passing the wrong path type to `bustDirCache` for real this time. In doing so, fixes a lot of other dubious uses of the dir_cache.

I still stand by the aggressive assertion in `bustDirCache`, despite it's recent issues. It, when properly called, is essential for making sure module resolution is reliable when filesystem races happen. This PR calls that assertion in *more* places. What I found out from this is we are very inconsistent in weather there is a trailing slash in our cache keys.

A lot of existing code was uncertain on weather it took a trailing slash or not, and it doesnt seem to be documented. For example: this paradox.

image

To fix this, **I am removing all cases where we keep the trailing slash on the key for `dir_cache`'s hash table**. This was easier than requiring trailing slashes because as fixed old cases, new ones came up that were increasingly harder to get a trailing slash on without copying the entire path into a new buffer.

In favor of making it easy to get a cache key, the normalization will remove it. See `Resolver.assertValidCacheKey` for details on the assertion. I left a very long comment here, as this fix took much longer than it should.

- [X] Added another case to the original test (this issue was a windows only re-regression of #3216) that now passes.
- [X] I built a ReleaseSafe copy of bun before the regression, then used that release copy to build bun again successfully.
- [X] I verified `vite` no longer regresses
- [X] I double checked that every caller of `bustDirCache` is indeed correct.
- [ ] Testing the Mac OS build locally to ensure no regressions of the assertion.
- [ ] Running tests in CI for all platforms to ensure zero regressions across the board.

### DIFF
--- a/packages/bun-internal-test/src/runner.node.mjs
+++ b/packages/bun-internal-test/src/runner.node.mjs
@@ -43,7 +43,7 @@ function defaultConcurrency() {
   // Concurrency causes more flaky tests, only enable it by default on windows
   // See https://github.com/oven-sh/bun/issues/8071
   if (windows) {
-    return Math.floor((cpus().length - 2) / 2);
+    return Math.floor((cpus().length - 2) / 3);
   }
   return 1;
 }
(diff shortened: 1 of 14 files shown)


paperclover:
~~New test does not pass and it is confusing why~~ Resolved

Issue #8693, opened 2024-02-05
stdin.pause/resume not working if there is a pause


jackcannon:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Darwin 23.1.0 x86_64 i386

### What steps can reproduce the bug?

I have made a small demo repo that illustrates the problem: https://github.com/jackcannon/stdin-bug-bun/tree/master

* `bun run example:node` - ✅ Works - runs the script using node
* `bun run example:bun` - ❌ Broken - runs the script using bun
* `bun run example:bun:ignore` - ✅ Works - runs the script using bun WITHOUT pausing between the prompts

The script it runs is: https://github.com/jackcannon/stdin-bug-bun/blob/master/stdin.js

Output (GIF):
![Kapture 2024-02-05 at 16 48 45](https://github.com/oven-sh/bun/assets/1922422/97b25de2-1aa1-402c-bab6-b404db0a9c93)

Output (text):
```
[16:47:51] stdin-bug-bun (master)$ bun --version
1.0.26
[16:48:11] stdin-bug-bun (master)$ bun run example:node
$ node stdin.js
What is your name - a? › jack
Pausing for 1 second...
What is your name - b? › jackb

RESULT: { a: 'jack', b: 'jackb' }
[16:48:23] stdin-bug-bun (master)$ bun run example:bun
$ bun run stdin.js
What is your name - a? › jack
Pausing for 1 second...
What is your name - b? ›

TIMEOUT - exiting...
[16:48:34] stdin-bug-bun (master)$ ackb
-bash: ackb: command not found
[16:48:34] stdin-bug-bun (master)$ bun run example:bun:ignore
$ bun run stdin.js --ignore-pause
What is your name - a? › jack
Skipping pause...
What is your name - b? › jackb

RESULT: {
  a: "jack",
  b: "jackb",
}
[16:48:41] stdin-bug-bun (master)$
```

The script consists of 2 'prompts' with a setTimeout between them. Each prompt does this at the start:
```typescript
      process.stdin.setRawMode(true);
      process.stdin.resume();
      process.stdin.on('data', listenFn);
      process.stdin.setEncoding('utf8');
```

and this when the return/enter button is registered:
```typescript
      process.stdin.setRawMode(false);
      process.stdin.pause();
      process.stdin.off('data', listenFn);
      resolve(result); // resolves the promise - continuing the script
```

In node, this works. Both prompts allow for input to be listened to.
[...]

PR #9155, opened 2024-02-28
Fix issue where process.stdin is ended too early

### LINKED ISSUE #5320: Bun doesn't run prisma generate or prisma migrate inside docker containers or WSL

camero2734:
### What does this PR do?
Fixes #5320 (hopefully the last issue)

The issue this time is kind of similar to the last one (#9101), in that a stream is being ended too early. This one's a bit more straightforward, though I'm not sure if it's the correct behavior.

This issue only occurs with `process.stdin` since it has some custom logic in `src/js/builtins/ProcessObjectInternals.ts`. Turns out there's this line:
```ts
stream._read = function (size) {
  $debug("_read();", reader);
  if (!reader) {
    // TODO: this is wrong
    this.push(null);
  } else if (!shouldUnref) {
    internalRead(this);
  }
};
```

This TODO was, in fact, true. When a `null` is pushed, the stream ends, so this was causing the stream to be ended prematurely if `reader` wasn't defined (which happens if `ref()` isn't called before `_read()` is).

To fix this, I just made it do nothing. Once I did this, however, I also noticed that `end` is emitted twice, so I added a check for whether end was already emitted, similar to the standard code in `stream.js`.

There's a test. You can also more easily test this manually by creating `test.js`:
```js
const { Transform } = require("node:stream");

let totalChunkSize = 0;
const uppercase = new Transform({
  transform(chunk, _encoding, callback) {
    totalChunkSize += chunk.length;
    callback(null, "");
  },
});

process.stdin.pipe(uppercase).pipe(process.stdout);
process.stdin.on("end", () => console.log(totalChunkSize));
```

and then run it with

```bash
printf 'a%.0s' {1..262164} | bun run test.js
```

It should output the correct size (262164). Previously it would output 196608.

I also tested with a large prisma schema -- it failed with the `JSON Parse error: Unterminated string` error before, but now successfully generates

### DIFF
--- a/test/js/third_party/prisma/helper.ts
+++ b/test/js/third_party/prisma/helper.ts
@@ -55,7 +55,7 @@ export function generate(type: string) {
 
   fs.writeFileSync(schema, content);
 
-  const result = Bun.spawnSync([bunExe(), "--bun", "x", "prisma", "generate", "--schema", schema], {
+  const result = Bun.spawnSync([bunExe(), "prisma", "generate", "--schema", schema], {
     cwd,
     env: {
       ...bunEnv,
--- a/test/js/third_party/prisma/helper.ts
+++ b/test/js/third_party/prisma/helper.ts
@@ -55,7 +55,7 @@ export function generate(type: string) {
 
   fs.writeFileSync(schema, content);
 
-  const result = Bun.spawnSync([bunExe(), "prisma", "generate", "--schema", schema], {
+  const result = Bun.spawnSync([bunExe(), "--bun", "x", "prisma", "generate", "--schema", schema], {
     cwd,
     env: {
       ...bunEnv,
--- a/src/js/builtins/ProcessObjectInternals.ts
+++ b/src/js/builtins/ProcessObjectInternals.ts
@@ -132,6 +132,7 @@ export function getStdinStream(fd) {
   const originalOn = stream.on;
 
   let stream_destroyed = false;
+  let stream_endEmitted = false;
   stream.on = function (event, listener) {
     // Streams don't generally required to present any data when only
     // `readable` events are present, i.e. `readableFlowing === false`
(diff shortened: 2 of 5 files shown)


Jarred-Sumner:
nice

Jarred-Sumner:
Can you add a test?

camero2734:
Before:
image

After:
image

test/js/third_party/prisma/helper.ts
```diff
@@ -55,7 +55,7 @@ export function generate(type: string) {
 
   fs.writeFileSync(schema, content);
 
-  const result = Bun.spawnSync([bunExe(), "prisma", "generate", "--schema", schema], {
```
Jarred-Sumner: using `bun x` shouldn't be necessary here

test/js/third_party/prisma/helper.ts
```diff
@@ -55,7 +55,7 @@ export function generate(type: string) {
...
+  const result = Bun.spawnSync([bunExe(), "--bun", "x", "prisma", "generate", "--schema", schema], {
```
Jarred-Sumner:
```suggestion
  const result = Bun.spawnSync([bunExe(), "prisma", "generate", "--schema", schema], {
```

Jarred-Sumner:
Thank you

argosphil:
Can we close #8693 , then?

argosphil:
Hmm. My fix in https://github.com/argosphil/bun/commit/38c60f62ba1b3dc3799caf9b8f9b9f5cc52da2be was a bit more involved, in that I called `internalRead` right away when a reader was added. Sorry again for forgetting about this, I'll do some more testing later to see whether that's still necessary.

test/js/third_party/prisma/helper.ts
camero2734: Yeah, had meant to remove that, thanks!

PR #9164, opened 2024-02-29
windows: pass test/transpiler/transpiler-stack-overflow.test.ts

nektro:
```js
const path = require("path");
console.log(path.resolve("C:\\bun\\test\\transpiler\\", "C:"));
```

```
C:\bun\test\transpiler
```

```js
const path = require("path");
console.log(path.resolve("C:\\bun\\test\\transpiler\\", "C:a"));
```

```
C:\bun\test\transpiler\a
```

### DIFF
--- a/src/sys.zig
+++ b/src/sys.zig
@@ -519,7 +519,7 @@ pub fn normalizePathWindows(
         @compileError("normalizePathWindows only supports u8 and u16 character types");
     }
     var wbuf: if (T == u16) void else bun.WPathBuffer = undefined;
-    const path = if (T == u16) path_ else bun.strings.convertUTF8toUTF16InBuffer(&wbuf, path_);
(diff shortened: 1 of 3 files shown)


src/sys.zig
```diff
@@ -540,6 +540,10 @@ pub fn normalizePathWindows(
         } };
     };
 
+    if (path.len >= 2 and path[0] >= 'A' and path[0] <= 'Z' and path[1] == ':') {
+        path = path[2..];
```
zackradisic, line 544: Don't we have a function for matching the disk drive pattern?

nektro: will change to `bun.path.isDriveLetter(path[0])` once the tests pass

paperclover (approved):

PR #9173, opened 2024-03-01
[windows] Fix `$Path` bug

Jarred-Sumner:
### What does this PR do?

Previously, we'd always check `C\foo` when searching for `foo`, and `C\foo` is not valid

Added tests


dylan-conway (approved):
LGTM

PR #9183, opened 2024-03-01
Quick "fix" for uninitialized memory used by the usockets code.

### LINKED ISSUE #9181: bun-usockets behaves unpredictably: uninitialized memory used

_No response_

_No response_

Run `bun-profile` in a debugger and set a breakpoint on `us_internal_timer_sweep`. Open a socket.

`us_internal_timer_sweep` is called consistently.

### What do you see instead?

`us_internal_timer_sweep` is called sometimes

### Additional information

There are several places in usockets which assume malloc()ed memory is cleared (all pointers set to NULL). This assumption is incorrect.

In this specific case, the problem manifests itself in `epoll_kqueue.c:us_loop_run_bun_tick`, where `us_loop_integrate` is called only if `timer_callback->cb` is NULL. `timer_callback->cb` is in fact uninitialized.

A very similar bug leads to an app crashing after 16 minutes. (Discussion on Discord).

My proposal is to change `us_malloc` to call `calloc` until these issues are fixed upstream. This seems to fix the 16-minute crash.

argosphil:
This simply clears allocated memory, which avoids at least two bugs. There's a slight performance impact, which is why this should eventually be removed again, once all callers have been fixed to initialize their entire data structures.

### What does this PR do?

Fixes #9181 .

This changes the `us_malloc` memory allocation routine to clear the memory it returns. This avoids at least two bugs:

1. `us_internal_timer_sweep` was sometimes never installed as a timer, causing a memory leak
2. `context->on_socket_long_timeout` sometimes contained uninitilaized data (ASCII in our case), causing a crash after 16 minutes when the timeout expired.

- [x] Code changes

I asked the person who'd seen consistent 16-minute crashes to install the change; the crash went away.

No automated tests right now. The right way to test for this automatically is to run CI with valgrind and report any uninitialized memory uses outside of GC, I think.

### DIFF
--- a/packages/bun-usockets/src/libusockets.h
+++ b/packages/bun-usockets/src/libusockets.h
@@ -17,7 +17,7 @@
 // clang-format off
 
 #ifndef us_malloc
-#define us_malloc malloc
+#define us_malloc(x) calloc(1, (x))
 #endif
 
 #ifndef us_realloc


Jarred-Sumner:
let's find out where the memory accesses are uninitialized instead of zeroing everything

argosphil:
Agreed. Closing as superseded by the jarred/zero branch.

PR #9184, opened 2024-03-01
Allow relative paths in cpAsync again

### LINKED ISSUE #9024: EBADF: Bad file descriptor.  errno: -9

1.0.28

Darwin 23.2.0 x86_64 i386 | Linux 5.15.0-94-generic x86_64 x86_64

1. Create bun.config.js

```js
import path from 'path'
import fs from 'fs'

const config = {
  sourcemap: 'external',
  entrypoints: ['app/javascript/application.js'],
  outdir: path.join(process.cwd(), 'app/assets/builds'),
  plugins: [
    {
      name: 'tinyMCE',
      async setup(build) {
        console.log('Start copying files')
        try {
          await fs.promises.rmdir('./public/assets', {
            recursive: true,
            force: true,
          })
          await fs.promises.mkdir('./public/assets/tinymce', {
            recursive: true,
          })
          await fs.promises.cp(
            './node_modules/tinymce',
            './public/assets/tinymce',
            { recursive: true },
          )
          console.log(`Number of plugins: ${build.config.plugins.length}`)
        } catch (err) {
          console.error(err)
        }
      },
    },
  ],
}

const build = async (config) => {
  const result = await Bun.build(config)

  if (!result.success) {
    if (process.argv.includes('--watch')) {
      console.error('Build failed')
      for (const message of result.logs) {
        console.error(message)
      }
      return
    } else {
      throw new AggregateError(result.logs, 'Build failed')
    }
  }
}

const main = async () => {
  await build(config)

  if (process.argv.includes('--watch')) {
    fs.watch(
      path.join(process.cwd(), 'app/javascript'),
      { recursive: true },
      (eventType, filename) => {
        console.log(`File changed: ${filename}. Rebuilding...`)
        build(config)
      },
    )
  } else {
    process.exit(0)
  }
}

main()

```
2. Create package.json

```json
{
  "name": "app",
  "private": "true",
  "scripts": {
    "build": "bun bun.config.js",
    "build:css:compile": "sass ./app/assets/stylesheets/application.sass.scss:./app/assets/builds/application.css --no-source-map --load-path=node_modules",
    "build:css:prefix": "postcss ./app/assets/builds/application.css --use=autoprefixer --output=./app/assets/builds/application.css",
    "build:css": "bun run build:css:compile && bun run build:css:prefix",
    "watch:css": "nodemon --watch ./app/assets/stylesheets/ --ext scss --exec \"bun run build:css\""
  },
  "dependencies": {
    "@hotwired/stimulus": "^3.2.2",
    "@hotwired/turbo-rails": "^7.3.0",
    "@popperjs/core": "^2.11.8",
    "@rails/request.js": "^0.0.9",
    "autoprefixer": "^10.4.16",
    "bootstrap": "^5.3.2",
    "bootstrap-icons": "^1.11.1",
    "flat-tree-builder": "1.5.17",
    "flatpickr": "^4.6.13",
    "nodemon": "^3.0.1",
    "postcss": "^8.4.31",
    "postcss-cli": "^10.1.0",
    "sass": "^1.69.3",
    "shortcut-buttons-flatpickr": "^0.4.0",
    "sortablejs": "^1.15.0",
    "tinymce": "^6.7.0",
    "tinymce-langs": "^1.0.0",
    "tom-select": "^2.2.2"
  }
}
```

3. bun run build

```bash
$ bun bun.config.js
Start copying files
31 |     },
32 |   ],
33 | }
34 |
35 | const build = async (config) => {
36 |   const result = await Bun.build(config)
                            ^
EBADF: Bad file descriptor
   errno: -9
 syscall: "open"
   path: "./node_modules/tinymce"

      at /Users/as/Documents/work/souz/bun.config.js:36:24
      at build (/Users/as/Documents/work/souz/bun.config.js:35:22)
      at /Users/as/Documents/work/souz/bun.config.js:52:9
      at main (/Users/as/Documents/work/souz/bun.config.js:51:14)
      at /Users/as/Documents/work/souz/bun.config.js:68:1
```

Build without errors, for example, bun v.1.0.26

```bash 
#  bun --version
1.0.26

 # bun run build
$ bun bun.config.js
Start copying files
Number of plugins: 1
```

### What do you see instead?

_No response_

### Additional information

This error started from version 1.0.27

argosphil:
### What does this PR do?

Fixes #9024 .

A recent commit in #8649 changed cpAsync to use

```
const fd = switch (Syscall.openatOSPath(bun.invalid_fd, src, open_flags, 0)) {
```

rather than

```
const fd = switch (Syscall.open(src, open_flags, 0)) {
```

The new code doesn't accept relative paths, which caused the issue.

This patch uses the CWD file descriptor instead.

- [x] Code changes

~Will add tests in a bit.~

There is a test.

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -6546,7 +6546,7 @@ pub const NodeFS = struct {
         }
 
         const open_flags = os.O.DIRECTORY | os.O.RDONLY;
-        const fd = switch (Syscall.openatOSPath(bun.invalid_fd, src, open_flags, 0)) {
+        const fd = switch (Syscall.openatOSPath(bun.toFD(std.fs.cwd().fd), src, open_flags, 0)) {
             .err => |err| {
                 task.finishConcurrently(.{ .err = err.withPath(this.osPathIntoSyncErrorBuf(src)) });
                 return false;
--- a/test/js/node/fs/cp.test.ts
+++ b/test/js/node/fs/cp.test.ts
@@ -285,6 +285,27 @@ for (const [name, copy] of impls) {
       assertContent(basename + "/hello/world/a.txt", "a");
       assertContent(basename + "/hello/world/b.txt", "b");
     });
+
+    test("relative paths for directories", async () => {
+      const basename = tempDirWithFiles("cp", {
+        "from/a.txt": "a",
+        "from/b.txt": "b",
+        "from/a.dir": { "c.txt": "c" },
+      });
+
+      const filter = jest.fn((src: string) => true);
+
+      let prev = process.cwd();
+      process.chdir(basename);
+
+      await copy("from", "result", {
+        recursive: true,
+      });
+
+      process.chdir(prev);
+
+      assertContent(basename + "/result/a.dir/c.txt", "c");
+    });
   });
 }
 


gvilums:
I'm not quite sure that this is right - local testing shows that copying works just fine with relative paths. I think the issue is actually a combination of two separate issues:
- when copying to a directory that already exists, we throw an error
- the error we throw has a misleading error message

running
```
console.log(await fs.promises.cp("abc", "def", { recursive: true }));
```
(assuming `abc` exists and contains `foo.txt`, and `def` doesn't exist) works, but if `def` exists it fails with 
```
EBADF: Bad file descriptor
   errno: -9
 syscall: "open"
   path: "abc"
```
This seems to be the error we're seeing. I don't think this is related to relative path processing, but rather the fact that we don't correctly handle the case where the directory already exists.

Node transparently does the copy if the directory already exists.

gvilums:
Disregard the above, I mixed up which use of openatOSPath you meant 😅

argosphil:
> Disregard the above, I mixed up which use of openatOSPath you meant 😅

Thank you, I was confused myself and really shouldn't have opened this without a test. There is one now. Criticism and nitpicking appreciated :-)

Jarred-Sumner:
Thank you

PR #9185, opened 2024-03-01
docs: added fix for linking issue on Ubuntu

benjaminstrasser:
### What does this PR do?

I had problems during `bun setup` on Ubuntu, which were fixed by installing `libstdc++-dev`.
The error message and a short description on the fix were added to the Troubleshooting section of the documentation.
This is useful to other potential contributors, who may encounter a similar error.

Side Note:
I installed a fresh Ubuntu 22.04 and neither of the two errors described in the Troubleshooting section came up.
On a fresh install everything seems to work fine.

- [ ] Code changes

### DIFF
--- a/docs/project/contributing.md
+++ b/docs/project/contributing.md
@@ -283,6 +283,31 @@ $ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 100
 $ sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-11 100
 ```
 
+### lstdc++ linking error on Ubuntu
+
+During the Bun setup process, you might encounter a similar looking linking error:
+
+```
+The C++ compiler
+
+  "/usr/bin/clang++-16"
+
+is not able to compile a simple test program.
+
+...
+
+ld.lld: error: unable to find library -lstdc++
+```
+
+This error indicates that the linker is unable to find the libstdc++ library, which is necessary for compiling C++ applications that depend on it.
+
+To resolve this issue, you may need to install the correct version of the **development** package for libstdc++ that provides the necessary headers and static library files.
+
+```bash
+sudo apt install libstdc++-12-dev
+#sudo apt install libstdc++-{version}-dev
+```
+
 ### libarchive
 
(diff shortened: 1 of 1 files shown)


argosphil:
I think we should also add `clang-16` and `libstdc++-12-dev` to the list of packages to be installed on Debian/Ubuntu, but I'm not sure how that interacts with using the LLVM APT repo, which is what we currently recommend.

What do you think?

benjaminstrasser:
@argosphil 
I think the current setup is fine and works on a fresh Ubuntu 22.04 installation so I would not change it, but I am not incredible familiar with the c/c++ toolchain.

I actually dug a little deeper into the cause of the issue.
After about 30 fresh VMs I am pretty confident the problem is gcc-12, even if gcc-11 is the active apt alternative. This might be an indication of a deeper lying issue with current setup, but I would not even know how start to verify that assumption.

To reproduce follow this scenario:
1. Install fresh Ubunut 22
2. Follow the installation guide
3. `bun setup` succeeds
4. `sudo apt install gcc-12`
5. `bun setup` fails
6. `sudo apt install libstdc++-12-dev`
7. `bun setup` succeeds

#### Some Thoughts

- I am not sure how common this scenario is. My installation is pretty fresh and I somehow managed to encounter it. Might be bad luck.
- I would not update the general guide to include other packages as it generally seems to work
- Maybe update the PR to reflect in the Troubleshooting section that  gcc-12 is a potential issue
- Maybe someone can look into this further to rule out some underlying problems.

What do you think?

paperclover (changes requested):
(edited to add more clarity)

how is this different from the situation that is documented in the section `'span' file not found on Ubuntu`? it really feels like duplication. i think this should be merged into the existing section about ubuntu, renaming it to something that covers both of these. then we can ensure there is a single solution to ensure the correct standard library and compiler toolchain.

as an alternative to this change of documentation, it may be worth actually fixing these issues for all linux distributions and installing LLVM 16 ourselves, since it appears to be very unreliable to do so in some linux distributions (specifically ubuntu).

we already have automatic downloads for the Zig compiler and prebuilt webkit artifacts.

PR #9190, opened 2024-03-01
feat: implement `expect.assertions` and `expect.hasAssertions`

Yash-Singh1:
### What does this PR do?

This PR adds support for `expect.assertions` and `expect.hasAssertions`

Example:

```ts
// Fail
test('example', () => {
  expect.hasAssertions();
})

test('example 2', () => {
  expect.assertions(2);
  expect(2).toBe(2);
})

// Pass
test('example 3', () => {
  expect.hasAssertions();
  expect(2).toBe(2);
})

test('example 4', () => {
  expect.assertions(2);
  expect(2).toBe(2);
  expect(true).toBe(true);
})

```

- [x] Code changes

If Zig files changed:


src/bun.js/test/expect.zig
```diff
@@ -4319,10 +4320,52 @@ pub const Expect = struct {
...
     pub const addSnapshotSerializer = notImplementedStaticFn;
 
+    pub fn hasAssertions(globalObject: *JSGlobalObject, callFrame: *JSC.CallFrame) callconv(.C) JSValue {
+        _ = callFrame;
+        defer globalObject.bunVM().autoGarbageCollect();
+
+        if (Jest.runner.?.pending_test) |pending_test|
+            pending_test.assert_asserts(NeededAssertType.atLeastOne, 0);
+
+        return .zero;
```
Jarred-Sumner: We cannot return `.zero` to JavaScript when there are no exceptions. It will segfault on access.  This should return `.undefined`
```suggestion
        return .undefined;
```

Please add a test that checks `expect().hasAssertions` returns undefined

src/bun.js/test/expect.zig
```diff
@@ -4319,10 +4320,52 @@ pub const Expect = struct {
...
+            globalObject.throw("Expected value must be a non-negative integer: {any}", .{expected.toFmt(globalObject, &fmt)});
+            return .zero;
+        }
+
+        const unsigned_expected_assertions: u32 = @intFromFloat(expected_assertions);
+
+        if (Jest.runner.?.pending_test) |pending_test|
+            pending_test.assert_asserts(NeededAssertType.equalToNeededAsserts, unsigned_expected_assertions);
+
+        return .zero;
```
Jarred-Sumner: This should return `.undefined`
```suggestion
        return .undefined;
```

Jarred-Sumner:
Thank you for this. A couple comments

src/bun.js/test/jest.zig
```diff
@@ -1390,6 +1428,11 @@ pub const TestRunnerTask = struct {
         processTestResult(this, this.globalThis, result, test_, test_id, describe);
     }
 
+    pub fn assert_asserts(this: *TestRunnerTask, assert_type: NeededAssertType, assert_count: u32) void {
+        this.needed_assert_type = assert_type;
```
dylan-conway, line 1432: given a test like this, jest will fail because it expects 0 assertions, but bun will pass because `hasAssertions` is the last to mark `needed_assert_type`. Would need to look at the jest implementation, but I think we need to check both `hasAssertions` and `assertions` after a test finishes
```js
test("debugging", () => {
  expect(1).toBe(1);
  expect.assertions(0);
  expect.hasAssertions();
});
```

Yash-Singh1: Jest keeps the state for `hasAssertions` and `assertions` separate alongside the state for the assertion counts. Should I do something similar by creating `pub var`s in `expect.zig` as Bun does for `active_test_expectation_counter`?:

https://github.com/oven-sh/bun/blob/dcf6f244a4033f5793cf86ce8ea597610ea208aa/src/bun.js/test/expect.zig#L42

PR #9176, opened 2024-03-01
windows: pass test/js/web/timers/setTimeout.test.js

nektro:

PR #9193, opened 2024-03-02
windows: pass test/js/web/websocket/websocket.test.js

nektro:
this struct contains many functions pointers, which when containing a value of uninitialized memory may be `0xCC`, `0xCD`, garbage, or similar. however later libuv and libusockets functions test if those pointers are NULL as a way to see if the particular callback has been added. being not `0x0` the checks would pass and swiftly cause a segfault later on.


nektro:
if windows and macOS fail this branch likely needs the ci fix from https://github.com/oven-sh/bun/pull/9176

nektro:
superseded by https://github.com/oven-sh/bun/pull/9191

PR #9197, opened 2024-03-02
add assertion for generated functions returning `JSValue.zero` without an exception

dylan-conway:
the debug message (not a real test failure):
Screenshot 2024-03-01 at 7 19 03 PM

let's see what CI says


paperclover (changes requested):
this is very nice to see. some of the work on my "delete exceptionref" does something similar to this. just adding it here will be a good ease into that pr (it adds a ton more of these kinds of assertions)

since your message says that toBe can run into this, can you at least fix that case (or is that fixed on a separate pr already?)

dylan-conway:
> since your message says that toBe can run into this, can you at least fix that case (or is that fixed on a separate pr already?)

The `toBe` assertion was forced locally by returning `.zero` on success just to see the debug message work. I'll update the description so it's more clear

PR #9209, opened 2024-03-02
fix: format specifier without characters in between

### LINKED ISSUE #7304: `%s` in Mocha tests?

1.0.14+d8be3e51b

Linux 6.5.0-13-generic x86_64 x86_64

Using [Mocha](https://mochajs.org/), I run `bun run test` and unexpectedly get `%s` in the output.

Not any `%s` since I don't get that when I run `npm run test`.

```
$ npm run test

> example@1.0.0 test
> mocha test/test-*.js



  numbers
    ✔ Integer

```

### What do you see instead?

```
$ bun run test
$ mocha test/test-*.js

%s 
  %s numbers
    ✔ Integer
```

### Additional information

_No response_

### LINKED ISSUE #9208: [BUG] %s Format specifiers without spaces between aren't replaced

1.0.29+a146856d1

Darwin 22.1.0 arm64 arm

1. Run `bun repl`
2. Run the following code:
```ts
console.log('%s%s', 'a', 'b')
```

Output should be:

```text
ab
```

### What do you see instead?

The output is:
```text
a%s b
```

### Additional information

In Node.js this returns `ab` as expected:

```text
> console.log('%s%s', 'a', 'b')
ab
```

This bug breaks the Mocha `spec` reporter, because it depends on `%s%s` for reporting test suite names:

https://github.com/mochajs/mocha/blob/b88978deb3c12f9b95502828f6ac29ebe8be85ef/lib/reporters/spec.js#L54

Yash-Singh1:
### What does this PR do?

Fixes the problem where format specifiers with no characters in between would not be replaced with their values in `console.log` and `util.format`. This was because the character after every format specifier was skipped when scanning.

- [x] Code changes

Wrote automated tests and also tested on a mocha repo

fixes #9208, fixes #7304

### DIFF
--- a/src/bun.js/ConsoleObject.zig
+++ b/src/bun.js/ConsoleObject.zig
@@ -1226,7 +1226,13 @@ pub const Formatter = struct {
         var i: u32 = 0;
         var len: u32 = @as(u32, @truncate(slice.len));
         var any_non_ascii = false;
+        var hit_percent = false;
         while (i < len) : (i += 1) {
+            if (hit_percent) {
+                i = 0;
+                hit_percent = false;
+            }
+
             switch (slice[i]) {
                 '%' => {
                     i += 1;
@@ -1251,6 +1257,7 @@ pub const Formatter = struct {
                     any_non_ascii = false;
                     slice = slice[@min(slice.len, i + 1)..];
                     i = 0;
+                    hit_percent = true;
                     len = @as(u32, @truncate(slice.len));
                     const next_value = this.remaining_values[0];
                     this.remaining_values = this.remaining_values[1..];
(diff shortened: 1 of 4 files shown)


dylan-conway (changes requested):
The new tests in this pr pass with or without the changes to `ConsoleObject.zig`. Could you add some tests to `console-log.test.ts` that fail without this bugfix?

Yash-Singh1:
> The new tests in this pr pass with or without the changes to `ConsoleObject.zig`. Could you add some tests to `console-log.test.ts` that fail without this bugfix?

My bad, I assumed that `util.format` uses the same logic as `console.log` for some reason. The new tests fail originally and pass with the code changes
Screenshot 2024-03-04 at 11 39 20 AM

dylan-conway (approved):
Thank you

PR #9210, opened 2024-03-03
fix: sep state for assertions and hasAssertions

Yash-Singh1:
### What does this PR do?

Makes state for `hasAssertions` and `assertions` separate as per https://github.com/oven-sh/bun/pull/9190#discussion_r1509877965.

- [x] Code changes


dylan-conway (approved):
Looks great, thanks!

PR #9081, opened 2024-02-24
feat: implemented jest.clearAllMocks, fixes #9079

kyr0:
### What does this PR do?

This PR addresses issue #9079, which I opened today. The tracking issue for Jest compatibility, #1825, already marks `jest.clearAllMocks` as completed, but the implementation was missing. I have filled this gap and confirmed that it behaves as specified.

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

I wrote automated tests.

- [x] I included a test for the new code, or existing tests cover it
- [x] I ran my tests locally and they pass (`bun-debug test mock-fn.test`)
- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [x] I included a test for the new code, or an existing test covers it
- [x] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed
- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test mock-fn.test`)
- [x] I added TypeScript types for the new methods, getters, or setters

PR #9217, opened 2024-03-03
Finish jest.clearAllMocks() implementation

Jarred-Sumner:
### What does this PR do?

#9081 added support for `jest.clearAllMocks()` when the mock was created via `spyOn`, but not when the mock was created via `jest.fn` or `mock(() => {})`.

This adds support for those cases as well by having a separate WeakSet for all mocked functions (instead of only spies).

There is a test


Jarred-Sumner:
@dylan-conway i'm going to merge but please take a look when you get a chance

PR #9233, opened 2024-03-04
shell: Improve `ShellError` stacktrace

### LINKED ISSUE #9182: `ShellError` should show original stack trace

1.0.29

Linux aarch64 aarch64

Currently, a failing shell command, like:

```typescript
import { $ } from 'bun'

$.throws(true)

await $`somecommandthatdoesnotexist`
```

Returns:

```text
3 |     return this.toString();
4 |   }
5 | 
6 |   class ShellError extends Error {
7 |     #output;
8 |     constructor(output, code) {
                                  ^
ShellError: Failed with exit code 1
 info: {
  "stderr": "",
  "exitCode": 1,
  "stdout": ""
}
```

With no error message or information, and where the line number of the error is the line number of the Bun source code for the `ShellError`, rather than the TypeScript application source code for the command that threw the error.

Expecting the line number to be the line number of the shell command where it was invoked.

### What do you see instead?

See the "reproduce" step above.

### Additional information

_No response_

zackradisic:
### What does this PR do?

This fixes errors thrown by Bun shell not showing the original stacktrace (Fixes #9182)

This is done by creating an Error object upfront upon creation of the `ShellPromise`.

Creating an Error object shouldn't be too costly

I put breakpoints on the `computeErrorInfo...` functions in `ZigGlobalObject.cpp` to make sure we're not doing the expensive thing of computing the stack trace, didn't seem to hit any of those functions so think we should be good

- [x] Code changes

### DIFF
--- a/src/js/builtins/shell.ts
+++ b/src/js/builtins/shell.ts
@@ -6,17 +6,21 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
     return this.toString();
   }
   class ShellError extends Error {
-    #output: ShellOutput;
-    constructor(output: ShellOutput, code: number) {
-      super(`Failed with exit code ${code}`);
+    #output?: ShellOutput = undefined;
+    constructor() {
+      super("");
+    }
+
+    initialize(output: ShellOutput, code: number) {
+      this.message = `Failed with exit code ${code}`;
       this.#output = output;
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you

PR #9234, opened 2024-03-04
Support `conditions` in Bun.build

Jarred-Sumner:
### What does this PR do?

#9106 added support for `bun --conditions` and `bun build --conditions`, but left out passing `conditions` to `Bun.build` . This adds it

There is a test


zackradisic (approved):
lgtm

PR #9245, opened 2024-03-04
Fix lost constructors in Node.js module classes (fixes #9242)

### LINKED ISSUE #9242: Constructor functions are 'lost' on some Node.js module classes

1.0.29+a146856d1

Darwin 21.6.0 x86_64 i386

Executing the following code (for example in repl):
```JavaScript
let h = require("node:http");
new h.OutgoingMessage().constructor === h.OutgoingMessage // On Node.js this is true on Bun this is false
```

See repl:
image

Constructor should be present in prototype. Since some classes are created using constructor functions on which a new prototype is set, the constructor is 'lost'. It needs to be explicitly (re)added to have it available. Alternative is to create all classes with an actual class definition.

### What do you see instead?

Requesting the constructor answers the first (typically class based) constructor method in the prototype chain. For OutgoingMessage this is the EventEmitter constructor.

### Additional information

A PR to fix this is underway.

ErikOnBike:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/9242

For compatibility with Node.js a number of constructor functions should be present in their respective prototype. Since some classes are created using constructor functions on which a new prototype is set, the constructor is 'lost'. It needs to be explicitly (re)added to have it available. Alternative is to create all classes with an actual class definition (instead of constructor function on which a new prototype is set).

Added tests which show the fix functions. Same tests can be used (without code changes) to verify the code failed before applying the fix. Tests performed locally.

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -431,6 +431,7 @@ function Server(options, callback) {
   return this;
 }
 Object.setPrototypeOf((Server.prototype = {}), EventEmitter.prototype);
+Server.prototype.constructor = Server; // Re-add constructor which got lost when setting prototype
 Object.setPrototypeOf(Server, EventEmitter);
 
 Server.prototype.ref = function () {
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
Thank you

PR #9256, opened 2024-03-05
fix!: remove worker from default conditions

### LINKED ISSUE #9255: Bun includes "worker" in default conditions

1.0.30+1424a196f

Linux 6.6.16-2-MANJARO x86_64 unknown

Bun always includes the "worker" condition, which resolves the "worker" conditional export in package.json.

```json
// lib/package.json
{
  "name": "lib",
  "type": "module",
  "exports": {
    ".": {
      "worker": "./worker.ts",
      "import": "./index.ts"
    }
  },
  "devDependencies": {
    "@types/bun": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  }
}
```

```typescript
// lib/worker.ts

export default import.meta.file;
```

```typescript
// lib/index.ts

export default import.meta.file;
```

```json
// app/package.json
{
  "name": "app",
  "module": "index.ts",
  "type": "module",
  "devDependencies": {
    "@types/bun": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  }
}
```

```typescript
// app/main.ts

import { default as lib } from "lib";

console.log(lib);
```

Output is "worker.ts"

I expect bun runtime not to "resolve" the worker conditional export.

### What do you see instead?

Bun resolves the "worker" conditional export.

### Additional information

Just delete these lines
https://github.com/oven-sh/bun/blob/bd2176ffb0f350ebd932790765bf7cc461761394/src/options.zig#L583
https://github.com/oven-sh/bun/blob/bd2176ffb0f350ebd932790765bf7cc461761394/src/options.zig#L589

ewmb7701:
### What does this PR do?

This removes "worker" from the default conditions for "bun" targets, which fixes #9255.

### DIFF
--- a/src/options.zig
+++ b/src/options.zig
@@ -580,13 +580,11 @@ pub const Target = enum {
         });
         array.set(Target.bun, &[_]string{
             "bun",
-            "worker",
             "node",
         });
         array.set(Target.bun_macro, &[_]string{
             "macro",
             "bun",
-            "worker",
             "node",
         });
 


paperclover:
thanks for making this change. while i think we should do this, i want to delay this from being merged until right before 1.1 is released, as it is a breaking change.

jdalton (approved):

Issue #9235, opened 2024-03-04
mssql/tedious connect ConnectionError "_nodeUrl.default.domainToASCII is not a function."


nxht:
### What version of Bun is running?

1.0.29+a146856d1

### What platform is your computer?

Linux 5.15.133.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

`package.json`: 
```json
{
  "dependencies": { "mssql": "^10.0.2" },
  "devDependencies": {
    "@types/bun": "^1.0.8",
    "@types/mssql": "^9.1.5"
  }
}
```

`mssql` case:
```typescript
import sql from 'mssql';

await sql.connect({ server: 'localhost' });
```

`tedious` case:
```typescript
import { Connection } from 'tedious';

const connection = new Connection({ server: 'localhost' });
connection.connect((err) => {
  if (err) throw err;
});
```

### What is the expected behavior?

Connect normally

### What do you see instead?

`mssql` case:
```
80 |         rejectOnce(err)
81 |         return
82 |       }
83 |       tedious.connect(err => {
84 |         if (err) {
85 |           err = new ConnectionError(err)
                                   ^
ConnectionError: Failed to connect to localhost:1433 - _nodeUrl.default.domainToASCII is not a function. (In '_nodeUrl.default.domainToASCII(host)', '_nodeUrl.default.domainToASCII' is undefined)
 code: "ESOCKET"

      at /home/wsladmin/typescript/test-bun/node_modules/mssql/lib/tedious/connection-pool.js:85:31
      at onConnect (/home/wsladmin/typescript/test-bun/node_modules/tedious/lib/connection.js:1443:12)
      at native:1:1
      at emit (native:1:1)
      at socketError (/home/wsladmin/typescript/test-bun/node_modules/tedious/lib/connection.js:2309:14)
      at /home/wsladmin/typescript/test-bun/node_modules/tedious/lib/connection.js:2030:9
      at processTicksAndRejections (native:1:1)

Exit code: 1
```

`tedious` case:
```
4 |   value: true
5 | });
6 | exports.RequestError = exports.ConnectionError = void 0;
7 | class ConnectionError extends Error {
8 |   constructor(message, code) {
9 |     super(message);
        ^
error: Failed to connect to localhost:1433 - _nodeUrl.default.domainToASCII is not a function. (In '_nodeUrl.default.domainToASCII(host)', '_nodeUrl.default.domainToASCII' is undefined)
[...]

PR #9257, opened 2024-03-05
node:url implement domainToASCII and domainToUnicode

nektro:
```
bun test v1.0.30 (51466ff1)

js/node/url/url-domain-ascii-unicode.test.js:
✓ url.domainToASCII > convert from 'ıíd' to 'xn--d-iga7r' [3.62ms]
✓ url.domainToASCII > convert from 'يٴ' to 'xn--mhb8f' [3.09ms]
✓ url.domainToASCII > convert from 'www.ϧƽəʐ.com' to 'www.xn--cja62apfr6c.com' [2.57ms]
✓ url.domainToASCII > convert from 'новини.com' to 'xn--b1amarcd.com' [2.41ms]
✓ url.domainToASCII > convert from '名がドメイン.com' to 'xn--v8jxj3d1dzdz08w.com' [2.44ms]
✓ url.domainToASCII > convert from 'افغانستا.icom.museum' to 'xn--mgbaal8b0b9b2b.icom.museum' [2.48ms]
✓ url.domainToASCII > convert from 'الجزائر.icom.fake' to 'xn--lgbbat1ad8j.icom.fake' [2.77ms]
✓ url.domainToASCII > convert from 'भारत.org' to 'xn--h2brj9c.org' [2.49ms]
✓ url.domainToUnicode > convert from 'xn--d-iga7r' to 'ıíd' [2.42ms]
✓ url.domainToUnicode > convert from 'xn--mhb8f' to 'يٴ' [2.45ms]
✓ url.domainToUnicode > convert from 'www.xn--cja62apfr6c.com' to 'www.ϧƽəʐ.com' [2.25ms]
✓ url.domainToUnicode > convert from 'xn--b1amarcd.com' to 'новини.com' [2.53ms]
✓ url.domainToUnicode > convert from 'xn--v8jxj3d1dzdz08w.com' to '名がドメイン.com' [2.51ms]
✓ url.domainToUnicode > convert from 'xn--mgbaal8b0b9b2b.icom.museum' to 'افغانستا.icom.museum' [2.17ms]
✓ url.domainToUnicode > convert from 'xn--lgbbat1ad8j.icom.fake' to 'الجزائر.icom.fake' [2.30ms]
✓ url.domainToUnicode > convert from 'xn--h2brj9c.org' to 'भारत.org' [2.20ms]

 16 pass
 0 fail
 16 expect() calls
Ran 16 tests across 1 files. [76.00ms]
Process exited with code 0.
```

Closes: https://github.com/oven-sh/bun/issues/9235


src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,89 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+        return JSC::JSValue::encode(arg0);
+
+    constexpr static int allowedNameToASCIIErrors = UIDNA_ERROR_EMPTY_LABEL | UIDNA_ERROR_LABEL_TOO_LONG | UIDNA_ERROR_DOMAIN_NAME_TOO_LONG | UIDNA_ERROR_LEADING_HYPHEN | UIDNA_ERROR_TRAILING_HYPHEN | UIDNA_ERROR_HYPHEN_3_4;
+    constexpr static size_t hostnameBufferLength = 2048;
+
+    auto encoder = &WTF::URLParser::internationalDomainNameTranscoder();
+    UChar hostnameBuffer[hostnameBufferLength];
+    UErrorCode error = U_ZERO_ERROR;
+    UIDNAInfo processingDetails = UIDNA_INFO_INITIALIZER;
+    int32_t numCharactersConverted = uidna_nameToASCII(encoder, StringView(domain).upconvertedCharacters(), domain.length(), hostnameBuffer, hostnameBufferLength, &processingDetails, &error);
```
Jarred-Sumner: Does `upconvertedCharacters()` always have the same length?

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,89 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+
+    if (U_SUCCESS(error) && !(processingDetails.errors & ~allowedNameToASCIIErrors) && numCharactersConverted) {
+        WTF::Vector<LChar, hostnameBufferLength> ascii;
+        ascii.append(hostnameBuffer, numCharactersConverted);
+        return JSC::JSValue::encode(JSC::jsString(vm, WTF::String(ascii.data(), ascii.size())));
```
Jarred-Sumner:
```suggestion
        return JSC::JSValue::encode(JSC::jsString(vm, WTF::String(hostnameBuffer, numCharactersConverted)));
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,89 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+    }
+    return JSC::JSValue::encode(jsUndefined());
```
Jarred-Sumner, line 1851: Does Node not throw an error?

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,89 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+}
+
+JSC_DEFINE_HOST_FUNCTION(functionDomainToUnicode, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
+{
+    auto& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    if (callFrame->argumentCount() != 1) {
```
Jarred-Sumner: The convention in JavaScript is not to error on additional arguments. The convention is to error if insufficient arguments.

```suggestion
    if (callFrame->argumentCount() < 1) {
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,89 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
     return JSC::JSValue::encode(jsUndefined());
 }
 
+JSC_DEFINE_HOST_FUNCTION(functionDomainToASCII, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
+{
+    auto& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    if (callFrame->argumentCount() != 1) {
```
Jarred-Sumner: The convention in JavaScript is not to error on additional arguments. The convention is to error if insufficient arguments.

```suggestion
    if (callFrame->argumentCount() < 1) {
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,89 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+        throwTypeError(globalObject, scope, "domainToASCII needs 1 argument"_s);
+        return JSC::JSValue::encode(JSC::JSValue {});
+    }
+
+    auto arg0 = callFrame->argument(0);
```
Jarred-Sumner, line 1797: You could remove the argumentCount() check altogether due to this line. argument(0) becomes `undefined` if argumentCount is insufficient.

src/node-fallbacks/url.js
```diff
@@ -741,10 +741,14 @@ export { Url as Url };
 
 export var pathToFileURL;
 export var fileURLToPath;
+export var domainToASCII;
+export var domainToUnicode;
 
 if (lazy) {
   pathToFileURL = lazy("pathToFileURL");
   fileURLToPath = lazy("fileURLToPath");
+  domainToASCII = lazy("domainToASCII");
```
Jarred-Sumner, line 750: lol i don't think this file is loaded anymore

test/exports/bun-exports.bun-v0.6.11.json
```diff
@@ -4452,6 +4452,8 @@
         },
         "URLSearchParams": "function",
         "Url": "function",
+        "domainToASCII": "function",
```
Electroid, line 4455: Is there a reason we're updating this file? (since it's 0.6.11?)

nektro: assumed the version was only when it was originally generated

test/js/node/url/url-domain-ascii-unicode.test.js
```diff
@@ -1,31 +1,31 @@
...
-      const asciiConvertedToUnicode = domainToUnicode(ascii);
-      assert.strictEqual(asciiConvertedToUnicode, domain);
+describe("url.domainToUnicode", () => {
+  for (const [domain, ascii] of pairs) {
+    test(`convert from '${ascii}' to '${domain}'`, () => {
+      const asciiConvertedToUnicode = url.domainToUnicode(ascii);
+      expect(asciiConvertedToUnicode).toEqual(domain);
     });
-  });
+  }
```
jdalton, line 95: Nice! These match Node's tests:
https://github.com/nodejs/node/blob/main/test/parallel/test-url-domain-ascii-unicode.js

jdalton: More tests can be found:
https://github.com/nodejs/node/blob/d96606d9aab35baf9e5c77cf9312145c3ae6086b/test/parallel/test-whatwg-url-custom-domainto.js
and
https://github.com/nodejs/node/blob/d96606d9aab35baf9e5c77cf9312145c3ae6086b/test/fixtures/url-idna.js

docs/runtime/bun-apis.md
```diff
@@ -100,6 +104,26 @@ Click the link in the right column to jump to the associated documentation.
...
+  [`Bun.fileURLToPath()`](/docs/api/utils#bun-fileurltopath)
+  [`Bun.pathToFileURL()`](/docs/api/utils#bun-pathtofileurl)
+  [`Bun.gzipSync()`](/docs/api/utils#bun-gzipsync)
+  [`Bun.gunzipSync()`](/docs/api/utils#bun-gunzipsync)
+  [`Bun.deflateSync()`](/docs/api/utils#bun-deflatesync)
+  [`Bun.inflateSync()`](/docs/api/utils#bun-inflatesync)
+  [`Bun.inspect()`](/docs/api/utils#bun-inspect)
+  [`Bun.nanoseconds()`](/docs/api/utils#bun-nanoseconds)
+  [`Bun.readableStreamTo*()`](/docs/api/utils#bun-readablestreamto)
+  [`Bun.resolveSync()`](/docs/api/utils#bun-resolvesync)
```
jdalton, line 127: ☝️ Is this auto-generated or hand maintained?

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,75 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+{
+    auto& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    auto arg0 = callFrame->argument(0);
+    if (!arg0.isString()) {
+        throwTypeError(globalObject, scope, "the \"domain\" argument is missing or not a string"_s);
+        return JSC::JSValue::encode(jsUndefined());
+    }
+
```
jdalton, line 1806: Node does: https://github.com/nodejs/node/blob/d96606d9aab35baf9e5c77cf9312145c3ae6086b/lib/internal/url.js#L1364-L1378
```
if (arguments.length < 1)
    throw new ERR_MISSING_ARGS('domain');
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,75 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+    UChar hostnameBuffer[hostnameBufferLength];
+    UErrorCode error = U_ZERO_ERROR;
+    UIDNAInfo processingDetails = UIDNA_INFO_INITIALIZER;
+    int32_t numCharactersConverted = uidna_nameToASCII(encoder, StringView(domain).characters16(), domain.length(), hostnameBuffer, hostnameBufferLength, &processingDetails, &error);
+
+    if (U_SUCCESS(error) && !(processingDetails.errors & ~allowedNameToASCIIErrors) && numCharactersConverted) {
+        return JSC::JSValue::encode(JSC::jsString(vm, WTF::String(hostnameBuffer, numCharactersConverted)));
+    }
+    throwTypeError(globalObject, scope, "domainToASCII failed"_s);
+    return JSC::JSValue::encode(jsUndefined());
```
jdalton, line 1851: Instead of throwing Node returns an empty string:
https://github.com/nodejs/node/blob/d96606d9aab35baf9e5c77cf9312145c3ae6086b/src/node_url.cc#L76-L97
For example `url.domainToUnicode('@')` returns `''` and `url.domainToASCII('@')` returns `''`.

jdalton (changes requested):
Notes on arg handling, error handling, and possible added tests (lots of goodies in there).

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,75 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+
+    auto domain = arg0.toWTFString(globalObject);
+    if (domain.isNull())
+        return JSC::JSValue::encode(jsUndefined());
+    if (!domain.is8Bit())
+        return JSC::JSValue::encode(arg0);
+
+    domain.convertTo16Bit();
+
+    constexpr static int allowedNameToUnicodeErrors = UIDNA_ERROR_EMPTY_LABEL | UIDNA_ERROR_LABEL_TOO_LONG | UIDNA_ERROR_DOMAIN_NAME_TOO_LONG | UIDNA_ERROR_LEADING_HYPHEN | UIDNA_ERROR_TRAILING_HYPHEN | UIDNA_ERROR_HYPHEN_3_4;
```
jdalton, line 1906: What are the allowed errors based on (or derived from?)

src/node-fallbacks/url.js
jdalton: @Jarred-Sumner Do you mean the fallback? How does lazy loading Bun bits work in the bundled/browserified scenario?

jdalton:
Node tracks the perf of this indicating it may be something we can compare to:
https://github.com/nodejs/node/blob/d96606d9aab35baf9e5c77cf9312145c3ae6086b/benchmark/url/whatwg-url-idna.js#L3

docs/runtime/bun-apis.md
Jarred-Sumner: hand maintained

src/node-fallbacks/url.js
Jarred-Sumner: Originally, we shared this code between runtime and client-side bundles. This `lazy` came from that. It isn't actually loaded in browser builds. This is dead code.

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,75 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+    UErrorCode error = U_ZERO_ERROR;
+    UIDNAInfo processingDetails = UIDNA_INFO_INITIALIZER;
+
+    int32_t numCharactersConverted = uidna_nameToUnicode(encoder, StringView(domain).characters16(), domain.length(), hostnameBuffer, hostnameBufferLength, &processingDetails, &error);
+
+    if (U_SUCCESS(error) && !(processingDetails.errors & ~allowedNameToUnicodeErrors) && numCharactersConverted) {
+        return JSC::JSValue::encode(JSC::jsString(vm, WTF::String(hostnameBuffer, numCharactersConverted)));
+    }
+    throwTypeError(globalObject, scope, "domainToUnicode failed"_s);
+    return JSC::JSValue::encode(jsUndefined());
```
jdalton, line 1920: I like returning the `jsUndefined()` equiv when throwing in our zig code instead of `.zero` (which creates a nullptr backed JSValue which other JSValue methods get grumpy about). In looking at how WebKit handles this they are a bit all over the place but something I've seen that fits is at least returning the expected type. So in this case `jsEmptyString(vm)` which could be stored higher in the function like https://github.com/WebKit/WebKit/blob/e604d7c70127a9d0abd95421c65a2ca594df690e/Source/JavaScriptCore/runtime/JSCell.cpp#L252 then `JSC::JSValue::encode(emptyString)`.

Jarred-Sumner: It is a bug that we return undefined instead of zero when there's an exception. This used to segfault. They added handling for it, but returning zero is good because it's a value which is unreachable from JavaScript. We cannot do any memory allocations while an exception is pending, they will fail. `jsEmptyString` might still work because the JSCell is not allocated on each call, but that is lucky/brittle

src/bun.js/bindings/ZigGlobalObject.cpp
nektro: WekKit

src/bun.js/bindings/ZigGlobalObject.cpp
nektro: I'd seen that code but wasn't totally sure of its error conditions. I'll try the `'@'` in my version and compare

test/js/node/url/url-domain-ascii-unicode.test.js
nektro: thank you!

test/exports/bun-exports.bun-v0.6.11.json
nektro: cc @Jarred-Sumner should this be left alone/removed in the future/kept and updated/etc ?

src/bun.js/bindings/ZigGlobalObject.cpp
nektro: rn it doesnt throw, it returns back `'@'`

test/js/node/url/url-domain-ascii-unicode.test.js
nektro: adding these cases jumped the result from `16 pass 0 fail` to `111 pass 9 fail`

nektro: fixed the remaining!

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,77 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+    if (!arg0.isString()) {
+        throwTypeError(globalObject, scope, "the \"domain\" argument is missing or not a string"_s);
+        return JSC::JSValue::encode(jsUndefined());
+    }
+
+    auto domain = arg0.toWTFString(globalObject);
+    if (domain.isNull())
+        return JSC::JSValue::encode(jsUndefined());
+    if (!domain.is8Bit())
+        return JSC::JSValue::encode(arg0);
```
Jarred-Sumner, line 1902: This is not correct. A UTF-16 string can have punycode characters.

src/bun.js/bindings/ZigGlobalObject.cpp
nektro: found the list https://url.spec.whatwg.org/#forbidden-host-code-point

src/bun.js/bindings/ZigGlobalObject.cpp
Jarred-Sumner: resolved IRL

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -2053,9 +2141,13 @@ JSC_DEFINE_HOST_FUNCTION(functionLazyLoad,
             return JSC::JSValue::encode(obj);
         }
 
+#if BUN_DEBUG
+        // this should always be unreachable unless in development and adding a new lazy value
```
Jarred-Sumner, line 2210: ASSERT_UNREACHABLE

Jarred-Sumner:
fixes LGTM

src/bun.js/bindings/ZigGlobalObject.cpp
nektro: https://github.com/WebKit/WebKit/blob/978f836eb161e5935c773e7ecaa2c76ac59b0ec2/Source/WTF/wtf/URLParser.cpp#L2966

src/bun.js/bindings/ZigGlobalObject.cpp
jdalton: > We cannot do any memory allocations while an exception is pending, they will fail.

Eeeh, that's a very good point.

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,142 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+        domain.contains(0x003C) || // U+003C (<)
+        domain.contains(0x003E) || // U+003E (>)
+        domain.contains(0x003F) || // U+003F (?)
+        domain.contains(0x0040) || // U+0040 (@)
+        domain.contains(0x005B) || // U+005B ([)
+        domain.contains(0x005C) || // U+005C (\)
+        domain.contains(0x005D) || // U+005D (])
+        domain.contains(0x005E) || // U+005E (^)
+        domain.contains(0x007C) // // U+007C (|).
+    )
```
jdalton, line 1830: ☝️ Very cool! I dig the spec reference and code comments here.

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,142 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
```
jdalton, line 1897: In a future PR we can DRY this out.

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -1779,6 +1784,142 @@ JSC_DEFINE_HOST_FUNCTION(jsReceiveMessageOnPort, (JSGlobalObject * lexicalGlobal
...
+        return JSC::JSValue::encode(jsEmptyString(vm));
+
+    if (domain.containsOnlyASCII())
+        return JSC::JSValue::encode(arg0);
+    if (domain.is8Bit())
+        domain.convertTo16Bit();
+
+    constexpr static int allowedNameToASCIIErrors = UIDNA_ERROR_EMPTY_LABEL | UIDNA_ERROR_LABEL_TOO_LONG | UIDNA_ERROR_DOMAIN_NAME_TOO_LONG | UIDNA_ERROR_LEADING_HYPHEN | UIDNA_ERROR_TRAILING_HYPHEN | UIDNA_ERROR_HYPHEN_3_4;
+    constexpr static size_t hostnameBufferLength = 2048;
```
jdalton, line 1839: 💡 The sitemaps protocol, for notifying Search Engines of URLs at a site, limits the length to 2048.

PR #9265, opened 2024-03-05
Fix running relative/absolute paths when missing package.json scripts

### LINKED ISSUE #9237: `Script not found` error message when running Prisma commands

1.0.30+1424a196f

Linux 6.6.16-linuxkit aarch64 unknown

```
mkdir hello-prisma
cd hello-prisma
bun init -y
bun install prisma
bun prisma init
```

No error message.

### What do you see instead?

```
error: Script not found "/home/bun/app/hello-prisma/node_modules/prisma/build/child"
```

### Additional information

I cannot pinpoint this issue to an exact location in Prisma’s source code, but there is a line in the minified bundle that reads

```js
childPath=path__default.default.join(eval("__dirname"),"child");
```

Changing `child` to `child.js` there removes the error message, but then strangely another newline character is printed after the command exits:

https://github.com/oven-sh/bun/assets/18495294/99567a1f-feb8-4444-8bd2-38c9a5703264

(There are two shell prompts without me pressing any buttons after running the command. The second one appears a moment after the first one)

This is reported in https://github.com/prisma/prisma/issues/23339 as well – not sure if this is an issue with Prisma or with Bun’s module resolution.

I briefly mentioned this issue in #5320 but since the main issues there were resolved, I decided to open another issue specifically for this.

camero2734:
### What does this PR do?
Fixes #9237

The code to attempt to execute relative/absolute files (when missing a file extension) seems to be placed in the wrong place. Currently, it's only executed if a `package.json` exists and it has a `scripts` property.

- Create a new directory with a `package.json` that doesn't contain a `scripts` section.
- Make a file called `hello.js`. Do whatever you want there.
- Try to execute it using `bun run ./hello`
- It will fail :(

It will pass if you just add a `scripts` section to the package.json. It also passes on `node`.

After these changes, it works independent of `package.json`.

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -1270,6 +1270,7 @@ pub const RunCommand = struct {
             return true;
         }
 
+        // Run package.json script
         if (root_dir_info.enclosing_package_json) |package_json| {
             if (package_json.scripts) |scripts| {
                 if (scripts.get(script_name_to_search)) |script_content| {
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
looks good only question is if we need to consider windows file paths here cc @paperdave

paperclover (approved):
this will not introduce regressions so i think it is ok

but i dont see why we are manually doing the work of `!bun.resolver.isPackagePath()` here. using a function like that instead of this manual checking will be a touch more reliable

the resolver (what Run.boot eventually calls) will handle normalization and so forth so i dont see it as a big deal.

PR #9266, opened 2024-03-05
Fixes #9263

### LINKED ISSUE #9263: `URLSearchParams.prototype.{ has, delete }` second argument support

1.0.30

mac

```js
const params = new URLSearchParams('a=1&a=2&b=3');
params.delete('a', 1);
params.delete('b', undefined);
params + ''; // => '', should be 'a=2';
```

_No response_

### What do you see instead?

_No response_

### Additional information

It's supported in all browsers, Node, Deno, etc more than a year.

### LINKED ISSUE #9283: `SuppressedError` should not support cause / options argument

1.0.30

mac

```js
new SuppressedError(1, 2, 3, { cause: 4 }).cause; // => 4, should be undefined
```

_No response_

### What do you see instead?

_No response_

### Additional information

It was removed from the proposal more than a year ago, https://github.com/tc39/proposal-explicit-resource-management/issues/147

Jarred-Sumner:
### What does this PR do?

Fixes #9263

some code copied from webkit

A couple tests

### DIFF
--- a/src/bun.js/bindings/URLSearchParams.cpp
+++ b/src/bun.js/bindings/URLSearchParams.cpp
@@ -86,10 +86,10 @@ String URLSearchParams::get(const String& name) const
     return String();
 }
 
-bool URLSearchParams::has(const String& name) const
+bool URLSearchParams::has(const String& name, const String& value) const
 {
     for (const auto& pair : m_pairs) {
-        if (pair.key == name)
+        if (pair.key == name && (value.isNull() || pair.value == value))
             return true;
     }
     return false;
@@ -101,6 +101,7 @@ void URLSearchParams::sort()
         return WTF::codePointCompareLessThan(a.key, b.key);
     });
     updateURL();
+    needsSorting = false;
 }
 
 void URLSearchParams::set(const String& name, const String& value)
@@ -147,13 +148,11 @@ Vector<String> URLSearchParams::getAll(const String& name) const
     return values;
 }
 
-void URLSearchParams::remove(const String& name)
+void URLSearchParams::remove(const String& name, const String& value)
 {
-    if (!m_pairs.removeAllMatching([&](const auto& pair) {
-            return pair.key == name;
-        })
-        && m_pairs.size() > 0)
-        return;
+    m_pairs.removeAllMatching([&](const auto& pair) {
+        return pair.key == name && (value.isNull() || pair.value == value);
+    });
     updateURL();
     needsSorting = true;
 }
--- a/src/bun.js/bindings/webcore/JSURLSearchParams.cpp
+++ b/src/bun.js/bindings/webcore/JSURLSearchParams.cpp
@@ -57,6 +57,7 @@
 #include <wtf/Vector.h>
 #include <variant>
 #include "GCDefferalContext.h"
+#include "wtf/StdLibExtras.h"
 
 namespace WebCore {
 using namespace JSC;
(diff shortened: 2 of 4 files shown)


src/bun.js/bindings/URLSearchParams.cpp
```diff
@@ -101,6 +101,7 @@ void URLSearchParams::sort()
         return WTF::codePointCompareLessThan(a.key, b.key);
     });
     updateURL();
+    needsSorting = false;
```
dylan-conway, line 104: why does `needsSorting` exist? it doesn't seem to be used

dylan-conway (approved):

src/bun.js/bindings/URLSearchParams.cpp
```diff
@@ -147,13 +148,11 @@ Vector<String> URLSearchParams::getAll(const String& name) const
...
+void URLSearchParams::remove(const String& name, const String& value)
 {
-    if (!m_pairs.removeAllMatching([&](const auto& pair) {
-            return pair.key == name;
-        })
-        && m_pairs.size() > 0)
-        return;
+    m_pairs.removeAllMatching([&](const auto& pair) {
+        return pair.key == name && (value.isNull() || pair.value == value);
+    });
```
jdalton, line 155: Relavent WebKit change: https://github.com/WebKit/WebKit/commit/93a5678f2d4f370aa8f957928f96fbb501e9e168

src/bun.js/bindings/webcore/JSURLSearchParams.cpp
```diff
@@ -278,7 +279,15 @@ static inline JSC::EncodedJSValue jsURLSearchParamsPrototypeFunction_deleteBody(
     EnsureStillAliveScope argument0 = callFrame->uncheckedArgument(0);
     auto name = convert<IDLUSVString>(*lexicalGlobalObject, argument0.value());
     RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
-    RELEASE_AND_RETURN(throwScope, JSValue::encode(toJS<IDLUndefined>(*lexicalGlobalObject, throwScope, [&]() -> decltype(auto) { return impl.remove(WTFMove(name)); })));
+
+    String value;
+    EnsureStillAliveScope argument1 = callFrame->argument(1);
```
jdalton, line 284: Very handy!
https://github.com/WebKit/WebKit/blob/f8b8f96af30ce6909fc2f7ece6c5575072463b7f/Source/JavaScriptCore/runtime/JSCJSValue.h#L720-L723

src/bun.js/bindings/webcore/JSURLSearchParams.cpp
```diff
@@ -278,7 +279,15 @@ static inline JSC::EncodedJSValue jsURLSearchParamsPrototypeFunction_deleteBody(
     EnsureStillAliveScope argument0 = callFrame->uncheckedArgument(0);
     auto name = convert<IDLUSVString>(*lexicalGlobalObject, argument0.value());
     RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
-    RELEASE_AND_RETURN(throwScope, JSValue::encode(toJS<IDLUndefined>(*lexicalGlobalObject, throwScope, [&]() -> decltype(auto) { return impl.remove(WTFMove(name)); })));
+
+    String value;
```
jdalton, line 283: We should do an argsCount() check here before `argument(1)` since the arg count may be 1 and not 2.

Jarred-Sumner: argument() does the check internally

jdalton: Ah that's right, I was looking at the wrong implementation. I thought I remembered something about `uncheckedArgument`: https://github.com/WebKit/WebKit/blob/de89f2977a314e5ddbe49384e35b953061d17f29/Source/JavaScriptCore/interpreter/CallFrame.h#L275-L285

PR #9272, opened 2024-03-06
docs<test>: update test matchers on jest compatibility

Marukome0743:
### What does this PR do?

According to [Bun v1.0.30 Blog](https://bun.sh/blog/bun-v1.0.30#expect-assertions-and-expect-hasassertions-are-now-supported), `expect.assertions()` and `expect.hasAssertions()` have been supported.
That's why I updated a mathers list in the document. 

- [ ] Code changes


Jarred-Sumner:
Thank you

PR #9278, opened 2024-03-06
Minor fixes

argosphil:
I'm working on running valgrind in CI. It's already found a few bugs, including what appears to be one in JSC (https://bugs.webkit.org/show_bug.cgi?id=270415)

Here's the first installment of fixes, though not all of them were valgrind finds. Please feel free to pick and choose.

My guess is that the empty spawn argument one might require discussion, in which case feel free to reject it for now and I'll open a proper issue. If anything else needs discussion, new tests, or docs, please let me know so I can throw it out of this PR and open a new one for that patch.

### What does this PR do?

- [x] Code changes


src/bun.js/api/bun/subprocess.zig
```diff
@@ -2042,11 +2042,6 @@ pub const Subprocess = struct {
                 while (cmds_array.next()) |value| {
                     const arg = value.getZigString(globalThis);
 
-                    // if the string is empty, ignore it, don't add it to the argv
```
Electroid, line 1669: Nice find, I also found this too but it got lost in some other um-merged PR. Mind extracting this out into another PR so it doesn't get lost again?

argosphil: Sure, though ideally #8981 would get merged first :-)

CMakeLists.txt
```diff
@@ -693,6 +693,8 @@ if(NOT NO_CODEGEN)
         "${BUN_SRC}/js/thirdparty/*.ts"
         "${BUN_SRC}/js/internal/*.js"
         "${BUN_SRC}/js/internal/*.ts"
+        "${BUN_SRC}/js/internal/*/*.js"
+        "${BUN_SRC}/js/internal/*/*.ts"
```
jdalton, line 747: Fancy glob support here would be nice (I know it's limited in CMake but this has me dreaming).

Makefile
```diff
@@ -223,7 +223,7 @@ ENABLE_MIMALLOC ?= 1
 
 # https://github.com/microsoft/mimalloc/issues/512
 # Linking mimalloc via object file on macOS x64 can cause heap corruption
-_MIMALLOC_FILE = libmimalloc.o
+_MIMALLOC_FILE = libmimalloc.a
 _MIMALLOC_INPUT_PATH = CMakeFiles/mimalloc-obj.dir/src/static.c.o
```
jdalton, line 227: ☝️ There's been some juggling with .o and .a depending on platform:
https://github.com/oven-sh/bun/pull/6800/files (do want more eyes on this)

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -1351,7 +1351,9 @@ static JSValue constructReportObjectComplete(VM& vm, Zig::GlobalObject* globalOb
 
         {
             char cwd[PATH_MAX] = { 0 };
-            getcwd(cwd, PATH_MAX);
+            char *r = getcwd(cwd, PATH_MAX);
+
+	    assert(r);
```
jdalton, line 1364: If this is mutating `cwd` then I don't think we need to assign it to a variable. We can add a comment that `getcwd` mutates.

src/js_lexer.zig
```diff
@@ -836,6 +836,10 @@ fn NewLexer_(
         }
 
         inline fn nextCodepoint(it: *LexerType) CodePoint {
+            if (it.current >= it.source.contents.len) {
+                it.end = it.current;
+                return -1;
+            }
```
jdalton, line 842: Did valgrind auto fix this? Very cool!

test/js/bun/ffi/ffi.test.js
```diff
@@ -954,6 +954,6 @@ test("can open more than 63 symbols", () => {
   });
 
   expect(Object.keys(lib.symbols).length).toBe(65);
-  expect(lib.symbols.strcasecmp(Buffer.from("ciro"), Buffer.from("CIRO"))).toBe(0);
-  expect(lib.symbols.strlen(Buffer.from("bunbun", "ascii"))).toBe(6n);
+  expect(lib.symbols.strcasecmp(Buffer.from("ciro\0"), Buffer.from("CIRO\0"))).toBe(0);
+  expect(lib.symbols.strlen(Buffer.from("bunbun\0", "ascii"))).toBe(6n);
```
jdalton, line 958: These JS strings don't need to be null terminated.

jdalton:
This is a lot of goodness @argosphil! Thank you.

src/bun.js/bindings/BunProcess.cpp
gvilums: `getcwd` does mutate, but returns `NULL` on error. We could change to `assert(getcwd(cwd, PATH_MAX))`, but I don't see an issue with the explicit version.

test/js/bun/ffi/ffi.test.js
gvilums: They do, because the c functions being called (`strcasecmp` and `strlen`) expect their input strings to be `NUL`-terminated.

argosphil:
> This is a lot of goodness @argosphil! Thank you.

Thank you! But I think I'll make separate PRs for each bug in the future, that way we can discuss things properly.

src/js_lexer.zig
argosphil: Valgrind found it, the fix is mine (and I messed up at first, forgetting the "it.end = it.current;" line). More reason to vet those changes thoroughly.

test/js/bun/ffi/ffi.test.js
jdalton: Ah, this is calling into c-land code. Cool cool.

src/js_lexer.zig
Jarred-Sumner: Which scenario does this happen? We allocate extra padding bytes in the source contents buffer to avoid this extra check in an extremely hot code path.

Makefile
Jarred-Sumner: This distinction should be deliberate (but a comment explaining this would be good). It impacts the order malloc gets loaded in. In some platforms we want to keep malloc usable or it will cause strange crashes at runtime when libc functions call free or malloc internally.

src/bun.zig
```diff
@@ -1137,6 +1137,7 @@ pub const getcwd = std.os.getcwd;
 
 pub fn getcwdAlloc(allocator: std.mem.Allocator) ![]u8 {
     var temp: [MAX_PATH_BYTES]u8 = undefined;
+    @memset(temp[0..], 0);
```
Jarred-Sumner, line 1196: Is this necessary? getcwd expects a fixed size buffer. If it is necessary, we should initialize with `.{0} ** MAX_PATH_BYTES` because it will potentially avoid dirtying the pages

src/string_immutable.zig
```diff
@@ -3904,7 +3904,8 @@ pub fn indexOfNeedsEscape(slice: []const u8) ?u32 {
             const vec: AsciiVector = remaining[0..ascii_vector_size].*;
             const cmp = @as(AsciiVectorU1, @bitCast((vec > max_16_ascii))) | @as(AsciiVectorU1, @bitCast((vec < min_16_ascii))) |
                 @as(AsciiVectorU1, @bitCast(vec == @as(AsciiVector, @splat(@as(u8, '\\'))))) |
-                @as(AsciiVectorU1, @bitCast(vec == @as(AsciiVector, @splat(@as(u8, '"')))));
+                @as(AsciiVectorU1, @bitCast(vec == @as(AsciiVector, @splat(@as(u8, '"'))))) |
```
Jarred-Sumner, line 3907: Instead of adding this extra check, we can make the check above >= 127.

src/toml/toml_lexer.zig
```diff
@@ -139,11 +139,16 @@ pub const Lexer = struct {
     }
 
     inline fn nextCodepointSlice(it: *Lexer) []const u8 {
+        if (it.current >= it.source.contents.len) return "";
```
Jarred-Sumner, line 142: We should add padding bytes instead imo

src/js_lexer.zig
argosphil: > Which scenario does this happen? We allocate extra padding bytes in the source contents buffer to avoid this extra check in an extremely hot code path.

Where does this happen? It's a good idea, both here and in the toml code below.

src/bun.zig
argosphil: No, it's not necessary! Thank you. This should be dependent on being a Valgrind build (as the SIMD code confuses Valgrind).

src/toml/toml_lexer.zig
gvilums: This will probably be branch-predicted perfectly anyway - I suggest we keep this change to ensure correctness.

gvilums:
Hey, I was just looking to move this along so the fixes don't get lost. @argosphil what do you think of the suggestions that @Jarred-Sumner made?

PR #9280, opened 2024-03-06
Fix `child_process.spawn({ timeout })` exiting too early

### LINKED ISSUE #9279: `child_process.spawn` exits instantly if `timeout` is defined
```ts
import { which } from "bun";
import { spawn } from "node:child_process";

test.if(!!which("sleep"))("child_process.spawn({ timeout }) should not exit instantly", async () => {
  const start = performance.now();
  await new Promise<void>((resolve, reject) => {
    const child = spawn("sleep", ["1000"], { timeout: 100 });
    child.on("error", reject);
    child.on("exit", (exitCode, signalCode) => {
      expect(exitCode).toBe(null);
      expect(signalCode).toBe("SIGTERM");
      resolve();
    });
  });
  const end = performance.now();
  expect(end - start).toBeGreaterThanOrEqual(100);
});

```

Electroid:
### What does this PR do?

Fixes #9279

- [x] Code changes

 I wrote automated tests

### DIFF
--- a/src/js/node/child_process.js
+++ b/src/js/node/child_process.js
@@ -163,7 +163,7 @@ function spawn(file, args, options) {
         }
         timeoutId = null;
       }
-    });
+    }, options.timeout);
 
     child.once("exit", () => {
       if (timeoutId) {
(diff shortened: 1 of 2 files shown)


paperclover (approved):

Jarred-Sumner:
While an improvement over status quo, this PR creates a new bug: it doesn't unref() the setTimeout call. This means the timeout keeps the parent process alive.

Also, killing the child_process isn't enough. We have to also destroy stdout and stderr to match Node behavior: https://github.com/nodejs/node/blob/2f4065250e453da2de2ee9cbb8aaae28cbdb4df5/lib/child_process.js#L446-L458

In the future please take a look at the Node implementation when making changes to Node APIs - that goes for both the reviewer and the committer

PR #9288, opened 2024-03-07
install: dont crash in debug mode when a semver has a build tag and multiple hyphens in the pre-release tag

nektro:


dylan-conway (approved):
Looks great!

PR #9289, opened 2024-03-07
fix(install): auto `node-gyp` script bugfix

dylan-conway:
### What does this PR do?
Auto `node-gyp` scripts can be added if `preinstall` and `install` don't exist. Previously we were checking for `install` and `postinstall`.

fixes first bug mentioned here https://github.com/microsoft/node-pty/issues/632#issuecomment-1982243227

existing tests


Jarred-Sumner:
nice

Issue #6630, opened 2023-10-21
NODE_ENV=test does not read .env.test file


purefun:
According to the document: https://github.com/oven-sh/bun/blob/969da088f5db3258a803ec186012e30f992829b4/docs/runtime/env.md?plain=1#L5-L10, Bun should read `.env.test` file, but it does not actually.

How to repreduce:

```sh
$ bun --version
1.0.6
```

```sh
$ cat index.ts
───────┬──────────────────────────────────────────────────
       │ File: index.ts
       │ Size: 47 B
───────┼──────────────────────────────────────────────────
   1   │ console.log(`AWESOME=${process.env.AWESOME}`);
───────┴──────────────────────────────────────────────────
```

```sh
$ cat .env.production
───────┬──────────────────────────────────────────────────
       │ File: .env.production
       │ Size: 19 B
───────┼──────────────────────────────────────────────────
   1   │ AWESOME=production
───────┴──────────────────────────────────────────────────
```

```sh
$  cat .env.test
───────┬──────────────────────────────────────────────────
       │ File: .env.test
       │ Size: 13 B
───────┼──────────────────────────────────────────────────
   1   │ AWESOME=test
───────┴──────────────────────────────────────────────────
```

```sh
$ NODE_ENV=production bun run index.ts
AWESOME=production
```

```sh
$ NODE_ENV=test bun run index.ts
AWESOME=undefined # ❌ should be AWESOME=test
```

Electroid:
Duplicate of #4118

chadacious:
Actually, this doesn't seem to be a duplicate of #4118 . Indeed `bun run index.ts` won't load the `.env.test` file even when `NODE_ENV=test`. The other environments for `.env.development` and `.env.production` work as expected. Perhaps it's because `.env.test` is specifically intended for `bun test` and not `bun run`?

Since there isn't support for `.env.staging` (see #9090 ), I encountered this problem by trying to use `.env.test` in a staging environment but it doesn't work like the others. In fact, in my testing it loads `.env.development` instead.

moshOntong-IT:
what is the status of this issue?

PR #9291, opened 2024-03-07
NODE_ENV=test should load .env.test even when .env.production exists

nektro:
Closes https://github.com/oven-sh/bun/issues/6630

prior to this patch the test's output would be

```
Expected: "test"
Received: "undefined"
```


Jarred-Sumner:
Thank you

Issue #9273, opened 2024-03-06
Crash in `fetch` when receiving invalid redirect Location header


gvilums:
### What version of Bun is running?

1.0.30+1424a196f

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

Save the following to a file and `bun run` it.
```ts
const server = Bun.serve({
  fetch(req) {
    return new Response(null, { status: 302, headers: { Location: 'http://foo:bar/' }})
  }
})

await fetch(`http://${server.hostname}:${server.port}/`);
```

### What is the expected behavior?

Receive some form of error that the redirect URL is invalid

### What do you see instead?

```
❯ bun run test.ts
[1]    66897 segmentation fault  bun run test.ts
```

### Additional information

Stack trace printed by `bun-debug`:

```

uh-oh: reached unreachable code
bun will crash now 😭😭😭
[SYS] openat(-2, /Users/gvilums/.bun/.bun-crash/v1.0.30-debug-1709730350177.crash) = -1

----- bun meta -----
Bun v1.0.30 (e4b2af5b) macOS Silicon 23.3.0
RunCommand: bunfig 
HTTP: 1
Elapsed: 26ms | User: 24ms | Sys: 8ms
RSS: 67.58MB | Peak: 67.58MB | Commit: 34.62MB | Faults: 0
----- bun meta -----

0   0x103c7e220 WTFGetBacktrace
1   ??? Bun__crashReportDumpStackTrace
2   ??? src.report.fatal
3   ??? src.main.panic
4   ??? src.string.String.toOwnedSlice
5   ??? src.http.handleResponseMetadata
6   ??? src.http.onData__anon_323446
7   ??? src.http.NewHTTPContext(false).Handler.onData
8   ??? src.deps.uws.NewSocketHandler(false).configure__anon_240258.SocketHandler.on_data
9   ??? us_internal_dispatch_ready_poll
10  ??? us_loop_run
11  ??? src.deps.uws.PosixLoop.run
12  ??? src.http.HTTPThread.processEvents
13  ??? src.http.HTTPThread.onStart
14  ??? Thread.callFn__anon_125495
15  ??? Thread.PosixThreadImpl.spawn__anon_95741.Instance.entryFn

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord

thread 5458989 panic: reached unreachable code
[...]

PR #9305, opened 2024-03-07
handle invalid URL in Location header for fetch()

nektro:
Closes https://github.com/oven-sh/bun/issues/9273

before:

crash and stack trace from above

after:

```
RedirectURLInvalid: Redirect URL in Location header is invalid.
 path: "http://localhost:3000/"
```


paperclover (approved):
swag

src/bun.js/webcore/response.zig
```diff
@@ -1229,6 +1229,7 @@ pub const Fetch = struct {
                     error.FailedToOpenSocket => bun.String.static("Was there a typo in the url or port?"),
                     error.TooManyRedirects => bun.String.static("The response redirected too many times. For more information, pass `verbose: true` in the second argument to fetch()"),
                     error.ConnectionRefused => bun.String.static("Unable to connect. Is the computer able to access the url?"),
+                    error.RedirectURLInvalid => bun.String.static("Redirect URL in Location header is invalid."),
```
Electroid, line 1232: In the future it would be great to add a test for this

PR #9306, opened 2024-03-07
Add support for remaining private methods

jdalton:
### What does this PR do?

Add support for remaining private methods.
This is a follow-up to https://github.com/oven-sh/WebKit/pull/46.

### DIFF
--- a/src/bun.js/bindings/JSNextTickQueue.cpp
+++ b/src/bun.js/bindings/JSNextTickQueue.cpp
@@ -1,18 +1,19 @@
 #include "root.h"
 
-#include "JavaScriptCore/JSCJSValueInlines.h"
-#include "JavaScriptCore/JSInternalPromise.h"
-#include "JavaScriptCore/LazyPropertyInlines.h"
-#include <JavaScriptCore/Weak.h>
 #include <JavaScriptCore/GetterSetter.h>
+#include <JavaScriptCore/JSCJSValueInlines.h>
+#include <JavaScriptCore/JSGlobalObject.h>
+#include <JavaScriptCore/JSInternalFieldObjectImplInlines.h>
+#include <JavaScriptCore/JSInternalPromise.h>
+#include <JavaScriptCore/LazyPropertyInlines.h>
+#include <JavaScriptCore/Structure.h>
+#include <JavaScriptCore/Weak.h>
 
 #include "JSNextTickQueue.h"
-#include <JavaScriptCore/JSGlobalObject.h>
-#include <JavaScriptCore/Structure.h>
-#include <JavaScriptCore/JSInternalFieldObjectImplInlines.h>
+
+#include "BunClientData.h"
 #include "ExtendedDOMClientIsoSubspaces.h"
 #include "ExtendedDOMIsoSubspaces.h"
-#include "BunClientData.h"
 
 namespace Bun {
 
@@ -84,7 +85,7 @@ void JSNextTickQueue::drain(JSC::VM& vm, JSC::JSGlobalObject* globalObject)
 
     if (!isEmpty()) {
         if (mustResetContext) {
-            globalObject->m_asyncContextData.get()->putInternalField(vm, 0, jsUndefined());
+            globalObject->asyncContextTuple()->putInternalField(vm, 0, jsUndefined());
         }
         auto* drainFn = internalField(2).get().getObject();
 
--- a/src/codegen/replacements.ts
+++ b/src/codegen/replacements.ts
@@ -2,63 +2,137 @@ import { LoaderKeys } from "../api/schema";
 import { sliceSourceCode } from "./builtin-parser";
 import { registerNativeCall } from "./js2native-generator";
 
-// This is a list of extra syntax replacements to do. Kind of like macros
-// These are only run on code itself, not string contents or comments.
-export const replacements: ReplacementRule[] = [
-  { from: /\bthrow new TypeError\b/g, to: "$throwTypeError" },
-  { from: /\bthrow new RangeError\b/g, to: "$throwRangeError" },
-  { from: /\bthrow new OutOfMemoryError\b/g, to: "$throwOutOfMemoryError" },
-  { from: /\bnew TypeError\b/g, to: "$makeTypeError" },
-  { from: /\bexport\s*default/g, to: "$exports =" },
-];
-
-// These rules are run on the entire file, including within strings.
-export const globalReplacements: ReplacementRule[] = [
-  {
-    from: /\bnotImplementedIssue\(\s*([0-9]+)\s*,\s*((?:"[^"]*"|'[^']+'))\s*\)/g,
-    to: "new TypeError(`${$2} is not implemented yet. See https://github.com/oven-sh/bun/issues/$1`)",
-  },
-  {
-    from: /\bnotImplementedIssueFn\(\s*([0-9]+)\s*,\s*((?:"[^"]*"|'[^']+'))\s*\)/g,
-    to: "() => $throwTypeError(`${$2} is not implemented yet. See https://github.com/oven-sh/bun/issues/$1`)",
-  },
-];
-
-// This is a list of globals we should access using @ notation
+// This is a list of globals we should access using @ notation.
 // This prevents a global override attacks.
 // Note that the public `Bun` global is immutable.
 // undefined -> __intrinsic__undefined -> @undefined
 export const globalsToPrefix = [
+  "console",
+  "decodeURI",
+  "decodeURIComponent",
+  "encodeURI",
+  "encodeURIComponent",
+  "escape",
+  "eval",
+  "globalThis",
+  "isFinite",
+  "isNaN",
+  "parseFloat",
+  "parseInt",
+  "undefined",
+  "unescape",
   "AbortSignal",
   "Array",
   "ArrayBuffer",
+  "Atomics",
+  "BigInt",
+  "BigInt64Array",
+  "BigUint64Array",
+  "Boolean",
   "Buffer",
+  "DataView",
+  "Date",
+  "Error",
+  "EvalError",
+  "FinalizationRegistry",
+  "Float32Array",
+  "Float64Array",
+  "Function",
   "Infinity",
+  "Int8Array",
+  "Int16Array",
+  "Int32Array",
+  "Intl",
+  "JSON",
   "Loader",
+  "Math",
   "Promise",
+  "Proxy",
+  "RangeError",
+  "ReferenceError",
+  "Reflect",
   "ReadableByteStreamController",
   "ReadableStream",
   "ReadableStreamBYOBReader",
   "ReadableStreamBYOBRequest",
   "ReadableStreamDefaultController",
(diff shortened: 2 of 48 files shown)


src/codegen/replacements.ts
```diff
@@ -54,6 +54,10 @@ export const globalsToPrefix = [
   "undefined",
 ];
 
+// This is a list of symbol we should access using @ notation.
+// Array.prototype.$$iterator -> Array.prototype.__intrinsic_$iterator -> Array.prototype.@$iterator
```
paperclover: At first, I was thinking this should result in `@@iterator`, so that

```
$$iterator -> __intrinsic____intrinsic__ -> @@iterator
```

but I realized you're probably looking to snapshot the original implementations of the Array.prototype.iterator, right?

I kind of wish that the double dollar sign can convert to double at, as [it seems you can refer to any wellknown symbol this way][1], but with the primordials use case this becomes a bit tricky. I guess it comes down to what looks best in our code. I wonder if special casing the primordial protection version just to `{known_class}.prototype.{known_symbol}` might be better. 

[1]: https://github.com/oven-sh/WebKit/blob/0aa1f6dfc9b48be1e89b8f709fd44e3c88feaf33/Source/JavaScriptCore/builtins/ArrayConstructor.js#L55

was also going to suggest looking at the built-in symbols list, like:
```
const symbolPropsToPrefix = Object.getOwnPropertyNames(Symbol)
  .filter(name => typeof Symbol[name] == 'symbol')
  .map(name => `$${name}`)
```

but the more i think about this, this is more like validation layer than a list to use for replacements (`@@` output will probably not be valid if it doesn't match a well-known symbol)

jdalton: > I kind of wish that the double dollar sign can convert to double at, as [it seems you can refer to any wellknown symbol this way](https://github.com/oven-sh/WebKit/blob/0aa1f6dfc9b48be1e89b8f709fd44e3c88feaf33/Source/JavaScriptCore/builtins/ArrayConstructor.js#L55), but with the primordials use case this becomes a bit tricky. I guess it comes down to what looks best in our code. I wonder if special casing the primordial protection version just to {known_class}.prototype.{known_symbol} might be better.

Yes WebKit's `@@iterator` is a reference to the symbol but not the value. So to snapshot the value I've tweaked it a bit. For our magic reference when one does `.$$iterator` we get `.@$iterator` which is the snapshotted value. We can write `foo[$$iterator]` to get the live value _(translated into `.@@iterator`)_.

> was also going to suggest looking at the built-in symbols list, like:

I want to be explicit here since this is partially added by us to WebKit and something we should update as we go whenever new things are added.

src/codegen/replacements.ts
```diff
@@ -81,95 +128,103 @@ export const enums = {
...
 
 // ------------------------------ //
 
 for (const name in enums) {
   const value = enums[name];
-  if (typeof value !== "object") throw new Error("Invalid enum object " + name + " defined in " + import.meta.file);
-  if (typeof value === null) throw new Error("Invalid enum object " + name + " defined in " + import.meta.file);
-  const keys = Array.isArray(value) ? value : Object.keys(value).filter(k => !k.match(/^[0-9]+$/));
-  define[`$${name}IdToLabel`] = "[" + keys.map(k => `"${k}"`).join(", ") + "]";
-  define[`$${name}LabelToId`] = "{" + keys.map(k => `"${k}": ${keys.indexOf(k)}`).join(", ") + "}";
```
jdalton, line 114: ☝️ Used the `index` param of `map` instead of `keys.indexOf`.

src/codegen/replacements.ts
```diff
@@ -81,95 +128,103 @@ export const enums = {
...
 };
 
 // ------------------------------ //
 
 for (const name in enums) {
   const value = enums[name];
-  if (typeof value !== "object") throw new Error("Invalid enum object " + name + " defined in " + import.meta.file);
-  if (typeof value === null) throw new Error("Invalid enum object " + name + " defined in " + import.meta.file);
-  const keys = Array.isArray(value) ? value : Object.keys(value).filter(k => !k.match(/^[0-9]+$/));
-  define[`$${name}IdToLabel`] = "[" + keys.map(k => `"${k}"`).join(", ") + "]";
```
jdalton, line 113: ☝️ Simplified the join as the keys weren't in need of mapping.

src/codegen/replacements.ts
```diff
@@ -81,95 +128,103 @@ export const enums = {
...
+  "process.platform": JSON.stringify(Bun.env.TARGET_PLATFORM ?? process.platform),
 };
 
 // ------------------------------ //
 
 for (const name in enums) {
   const value = enums[name];
-  if (typeof value !== "object") throw new Error("Invalid enum object " + name + " defined in " + import.meta.file);
-  if (typeof value === null) throw new Error("Invalid enum object " + name + " defined in " + import.meta.file);
-  const keys = Array.isArray(value) ? value : Object.keys(value).filter(k => !k.match(/^[0-9]+$/));
```
jdalton, line 112: `k.match` produces an array or null. In this case we just needed a boolean so `regexp.test` is a better fit.

src/codegen/replacements.ts
```diff
@@ -81,95 +128,103 @@ export const enums = {
...
-  to: string;
+  to: string | Replacer;
   global?: boolean;
 }
 
 /** Applies source code replacements as defined in `replacements` */
 export function applyReplacements(src: string, length: number) {
   let slice = src.slice(0, length);
-  let rest = src.slice(length);
-  slice = slice.replace(/([^a-zA-Z0-9_\$])\$([a-zA-Z0-9_]+\b)/gm, `$1__intrinsic__$2`);
```
jdalton, line 140: ☝️ `a-zA-Z0-9_` is `\w`. _(there's case folding too but not an issue here)_

src/codegen/replacements.ts
```diff
@@ -81,95 +128,103 @@ export const enums = {
...
+    // Replace symbol live property references.
+    slice = slice.replaceAll(`[$$${name}]`, `.__intrinsic____intrinsic__${name}`);
+  }
+  slice = slice.replace(/([^\$\w])\$(\w+\b)/gm, "$1__intrinsic__$2");
   for (const replacement of replacements) {
-    slice = slice.replace(replacement.from, replacement.to.replaceAll("$", "__intrinsic__").replaceAll("%", "$"));
+    slice = slice.replace(replacement.from, replacement.to as Replacer);
   }
-  let match;
-  if ((match = slice.match(/__intrinsic__(debug|assert)$/)) && rest.startsWith("(")) {
```
jdalton: ☝️ Doing the cheap check first (`.startsWith`).

src/codegen/replacements.ts
```diff
@@ -81,95 +128,103 @@ export const enums = {
...
+  // Replace symbol @@ names first.
+  for (const name of symbolPropsToPrefix) {
+    // Replace symbol snapshot property references.
+    slice = slice.replaceAll(`.$$${name}`, `.__intrinsic__$${name}`);
+    // Replace symbol live property references.
+    slice = slice.replaceAll(`[$$${name}]`, `.__intrinsic____intrinsic__${name}`);
+  }
+  slice = slice.replace(/([^\$\w])\$(\w+\b)/gm, "$1__intrinsic__$2");
   for (const replacement of replacements) {
-    slice = slice.replace(replacement.from, replacement.to.replaceAll("$", "__intrinsic__").replaceAll("%", "$"));
```
jdalton, line 142: Move `replacement.to` transforms higher to ensure they are performed once and to improve consistency of transformation between `replacements` and `globalReplacements`.

src/codegen/replacements.ts
```diff
@@ -1,29 +1,7 @@
...
-// These rules are run on the entire file, including within strings.
-export const globalReplacements: ReplacementRule[] = [
-  {
-    from: /\bnotImplementedIssue\(\s*([0-9]+)\s*,\s*((?:"[^"]*"|'[^']+'))\s*\)/g,
-    to: "new TypeError(`${$2} is not implemented yet. See https://github.com/oven-sh/bun/issues/$1`)",
-  },
-  {
-    from: /\bnotImplementedIssueFn\(\s*([0-9]+)\s*,\s*((?:"[^"]*"|'[^']+'))\s*\)/g,
-    to: "() => $throwTypeError(`${$2} is not implemented yet. See https://github.com/oven-sh/bun/issues/$1`)",
-  },
```
jdalton, line 24: Before this PR the `globalReplacements` `.to` was being transformed with`.replaceAll("$", "__intrinsic__")` which would have made the `$2` and `$1` incorrectly transformed because it didn't have the transform that the `replacements` `.to` has for `%1`, `%2`, etc.

src/codegen/replacements.ts
```diff
@@ -1,29 +1,7 @@
...
-  { from: /\bthrow new OutOfMemoryError\b/g, to: "$throwOutOfMemoryError" },
-  { from: /\bnew TypeError\b/g, to: "$makeTypeError" },
-  { from: /\bexport\s*default/g, to: "$exports =" },
-];
-
-// These rules are run on the entire file, including within strings.
-export const globalReplacements: ReplacementRule[] = [
-  {
-    from: /\bnotImplementedIssue\(\s*([0-9]+)\s*,\s*((?:"[^"]*"|'[^']+'))\s*\)/g,
-    to: "new TypeError(`${$2} is not implemented yet. See https://github.com/oven-sh/bun/issues/$1`)",
```
jdalton, line 19: `notImplementedIssue` isn't used in our code ✂️ , where as `notImplementedIssueFn` is.

paperclover (commented):
changes in replacements.ts is great so far. i am happy to see someone else finally add something to the builtins bundler.

src/codegen/replacements.ts
jdalton: Resolved by adding all of the missing things.

src/bun.js/bindings/JSNextTickQueue.cpp
```diff
@@ -84,7 +84,7 @@ void JSNextTickQueue::drain(JSC::VM& vm, JSC::JSGlobalObject* globalObject)
 
     if (!isEmpty()) {
         if (mustResetContext) {
-            globalObject->m_asyncContextData.get()->putInternalField(vm, 0, jsUndefined());
+            globalObject->asyncContextTuple()->putInternalField(vm, 0, jsUndefined());
         }
```
jdalton, line 89: With this PR I've made a distinction between the asyncContextTuple and asyncContextData because in main the current forms are conflated as `asyncContext` which gets confusing (is it the tuple or the data of the tuple). With this we have a clear and consistent distinction.

src/js/builtins/ReadableStreamInternals.ts
```diff
@@ -864,8 +865,8 @@ export function onPullDirectStream(controller) {
     deferFlush = controller._deferFlush;
     controller._deferFlush = controller._deferClose = 0;
 
-    if (asyncContext) {
-      $putInternalField($asyncContext, 0, prev);
```
jdalton, line 868: Fixed as `prev` was not defined in the outer scope.

PR #9309, opened 2024-03-08
fix: bun.serve crash due to ExceptionRef

### LINKED ISSUE #8985: Port in use throws segmentation fault

1.0.3 - 1.0.27

Darwin 23.1.0 arm64 arm

Copied the code from [this example](https://bun.sh/guides/websocket/upgrade) into a ws-server.ts file.

I already have other Bun services running locally on my machine using ports 3000-3010.

```console
% bun ws-server.ts
zsh: segmentation fault  bun ws-server.ts
```

Downgrading to Bun v.1.0.2 shows a more correct error

```console
[0.26ms] ".env"
1 | const server = Bun.serve({
                  ^
error: Failed to start server. Is port 3001 in use?
```

From my testing, it seems that all versions since and including v.1.0.3 shows the segmentation error.

When failing to bind the server to the requested port, I'd expect an error indicating this, as Bun did in v1.0.2.

### What do you see instead?

_No response_

### Additional information

_No response_

paperclover:
image

Fixes #8985

This is a workaround for something fixed by the ExceptionRef PR. Because that pr makes this type of mistake impossible to write, I feel this bug does not need a test case for.

Two exceptions were thrown, leading to a segfault. In the future, this will be covered by

1. An assertion for overlapping errors.
2. All exception-causing functions to return a Zig error type, making it impossible to overlap exceptions like ExceptionRef allowed.


docs/guides/websocket/upgrade.md
```diff
@@ -1,28 +0,0 @@
----
```
paperclover, line 1: This guide is a duplicate of https://bun.sh/guides/websocket/simple

Jarred-Sumner:
great

Issue #9015, opened 2024-02-20
Bun.serve error callback is missing the request


BennyAlex:
### What is the problem this feature would solve?

Accessing any information about the request that leads to the error is not possible, since there is no request parameter for the error(errorLike) callback at the moment.

Users want to show different errors based on the url or user information, or as in my case I want to set the CORS headers dynamically to the request's origin.

### What is the feature you are proposing to solve the problem?

Passing the request to error() callback

### What alternatives have you considered?

idk

paperclover:
the reason we dont have this is because the error callback might not always have the request info. if we were to implement this, i imagine it would be sort of 'best effort' and potentially null / invalid. not sure of the full details

BennyAlex:
ok but I guess in many cases it will be available and then of course we have to check if its there or not but still better than no info at all

PR #9310, opened 2024-03-08
Bun.serve: error: pass Request parameter when available

nektro:
Closes https://github.com/oven-sh/bun/issues/9015


test/js/bun/http/bun-server.test.ts
```diff
@@ -422,6 +423,39 @@ describe("Server", () => {
     expect(stderr).toBeEmpty();
     expect(exitCode).toBe(0);
   });
+
+  test("supports error handling", async () => {
```
Jarred-Sumner, line 438: I think this needs more tests before we can merge it. 

It should have tests for each code path which returns an error. The tests should do something with the Request, like call .headers.get("X-Foo") and make sure it matches "1" or something. We want to trigger the code path which converts from a uWS::Request* into a JSC.WebCore.Request.

I also think it needs tests that check request is null or not null and fail appropriately.

src/bun.js/api/server.zig
```diff
@@ -2945,14 +2960,18 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
             JSC.markBinding(@src());
             if (!this.server.config.onError.isEmpty() and !this.flags.has_called_error_handler) {
                 this.flags.has_called_error_handler = true;
                 const result = this.server.config.onError.callWithThis(
                     this.server.globalThis,
                     this.server.thisObject,
-                    &.{value},
+                    &.{
+                        value,
+                        request_value,
```
Jarred-Sumner, line 2975: Can you add an assertion that `request_value != .zero`? If we pass .zero it will crash with a much weirder error

test/js/bun/http/bun-server.test.ts
```diff
@@ -433,6 +434,74 @@ test("unref keeps process alive for ongoing connections", async () => {
...
+      port: 0,
+      fetch(req) {
+        throw new Error("woops!");
+      },
+      error(error, req) {
+        return new Response(`${req.url}`);
+      },
+    });
+
+    const url = `http://${server.hostname}:${server.port}/`;
```
Electroid, line 498: Use `server.url`

Jarred-Sumner:
great

Issue #1932, opened 2023-01-29
execute code from stdin


Kapsonfire:
### What is the problem this feature would solve?

Executing code from CLI without storing in a file before

### What is the feature you are proposing to solve the problem?

bun -e 'console.log("Hello world")'

### Why -e
e => eval

90dy:
Eval take it from command line options,

I would like something like node, where you can run it that way:
```bash
bun run <<< "console.log('hello world !')"
```

Or maybe with an option
```bash
bun run --stdin <<< "[...]"
```

What do you think about that ?

FreePhoenix888:
It would be nice if this feature would be implemented not only by using CLI but also right in the code
Example:
```
const code = "const result: number = 1+1; result"
const result = Bun.eval(code)
```

Our [organization](https://github.com/deep-foundation/) needs it. Right now we transpile code to js and eval it

FreePhoenix888:
For some situations `bun repl --eval` can be used but note that it does not evaulaute like `eval`. If you want to get some result you have to use console.log for example or write to file.

This command will not write 2 to stdout:
```ts
bun repl --eval "const result: number = 1+1; result"
```

This command will write 2 to stdout:
```ts
bun repl --eval "const result: number = 1+1; console.log(result)"
```

jhmaster2000:
> For some situations `bun repl --eval` can be used but note that it does not evaulaute like `eval`. If you want to get some result you have to use console.log for example or write to file.
> 
> This command will not write 2 to stdout:
> 
> ```ts
> bun repl --eval "const result: number = 1+1; result"
> ```
> 
> This command will write 2 to stdout:
> 
> ```ts
> bun repl --eval "const result: number = 1+1; console.log(result)"
> ```

If you replace --eval with --print on the first one it will show the 2.

FreePhoenix888:
> ```ts
> bun repl --eval "const result: number = 1+1; console.log(result)"
> ```

True!

FreePhoenix888:
Array:
```
freephoenix888@FreePhoenix:~/Programming/bun-test$ bun repl --print "const result: number = 1+1; [result]"
[ 2 ]
```

FreePhoenix888:
Object:
```
freephoenix888@FreePhoenix:~/Programming/bun-test$ bun repl --print "const result: number = 1+1; {result}"
2

freephoenix888@FreePhoenix:~/Programming/bun-test$ bun repl --print "const result: number = 1+1; {result: result}"
2
[...]

PR #9311, opened 2024-03-08
allow bun run to accept js from stdin

nektro:
Closes https://github.com/oven-sh/bun/issues/1932

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -1599,6 +1599,10 @@ pub const VirtualMachine = struct {
                 virtual_source = eval_script;
                 loader = .tsx;
             }
+            if (strings.endsWithComptime(specifier, bun.pathLiteral("/[stdin]"))) {
+                virtual_source = eval_script;
+                loader = .tsx;
+            }
         }
 
         defer jsc_vm.module_loader.resetArena(jsc_vm);
@@ -1670,7 +1674,10 @@ pub const VirtualMachine = struct {
             ret.result = null;
             ret.path = result.path;
             return;
-        } else if (jsc_vm.module_loader.eval_script != null and strings.endsWithComptime(specifier, bun.pathLiteral("/[eval]"))) {
+        } else if (jsc_vm.module_loader.eval_script != null and
+            (strings.endsWithComptime(specifier, bun.pathLiteral("/[eval]")) or
+            strings.endsWithComptime(specifier, bun.pathLiteral("/[stdin]"))))
+        {
             ret.result = null;
             ret.path = specifier;
             return;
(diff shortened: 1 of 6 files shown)


paperclover (approved):

src/cli/run_command.zig
```diff
@@ -1441,6 +1441,41 @@ pub const RunCommand = struct {
...
+            ctx.runtime_options.eval_script = list.items;
+
+            const trigger = bun.pathLiteral("/[stdin]");
+            var entry_point_buf: [bun.MAX_PATH_BYTES + trigger.len]u8 = undefined;
+            const cwd = try std.os.getcwd(&entry_point_buf);
+            @memcpy(entry_point_buf[cwd.len..][0..trigger.len], trigger);
+            const entry_path = entry_point_buf[0 .. cwd.len + trigger.len];
+
+            Run.boot(ctx, ctx.allocator.dupe(u8, entry_path) catch return false) catch |err| {
+                if (Output.enable_ansi_colors) {
```
Jarred-Sumner: lets use this pattern instead

image

Jarred-Sumner:
yay

src/bun.js/javascript.zig
```diff
@@ -1670,7 +1674,10 @@ pub const VirtualMachine = struct {
             ret.result = null;
             ret.path = result.path;
             return;
-        } else if (jsc_vm.module_loader.eval_script != null and strings.endsWithComptime(specifier, bun.pathLiteral("/[eval]"))) {
+        } else if (jsc_vm.module_loader.eval_script != null and
+            (strings.endsWithComptime(specifier, bun.pathLiteral("/[eval]")) or
+            strings.endsWithComptime(specifier, bun.pathLiteral("/[stdin]"))))
```
jdalton, line 1679: I dig the comptime `bun.pathLiteral` helper.
###### _(less pretty alt: `std.fs.path.sep_str ++ "[eval]"`)_


👀 There's an unrelated line just above this that doesn't normalize the `/` like for platform like `bun.pathLiteral`. Maybe a future Win bug?
```
} else if (strings.hasPrefixComptime(specifier, "/bun-vfs/node_modules/")) {
```

src/bun.js/module_loader.zig
```diff
@@ -2141,6 +2141,10 @@ pub const ModuleLoader = struct {
                 virtual_source = eval_script;
                 loader = .tsx;
             }
+            if (strings.endsWithComptime(specifier, bun.pathLiteral("/[stdin]"))) {
```
jdalton, line 2144: ☝️ Since `strings.endsWithComtime` could be only either `/[eval]/` or `/[stdin]` we could follow-up to collapse this into the one above with an `or`

PR #9319, opened 2024-03-08
chore: just a little bit of cleanup

paperclover:


src/shell/interpreter.zig
```diff
@@ -3278,7 +3278,7 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
                         const err = switch (this.state) {
                             .expanding_redirect => this.state.expanding_redirect.expansion.state.err,
                             .expanding_args => this.state.expanding_args.expansion.state.err,
-                            else => @panic("Invalid state"),
+                            else => |t| std.debug.panic("Unexpected state .{s} in Bun shell", .{@tagName(t)}),
```
Jarred-Sumner, line 3281: Output.panic, we should ban std.debug.panic

src/shell/interpreter.zig
```diff
@@ -6732,7 +6732,7 @@ pub fn NewInterpreter(comptime EventLoopKind: JSC.EventLoopKind) type {
                 pub fn onAsyncTaskDone(this: *Rm, task: *ShellRmTask) void {
                     var exec = &this.state.exec;
                     const tasks_done = switch (exec.state) {
-                        .idle => @panic("Invalid state"),
+                        .idle => @panic("Unexpected state .idle in Bun shell 'rm' builtin."),
```
Jarred-Sumner, line 6735: Nice

Jarred-Sumner:
LGTM but @dylan-conway should look it over incase opinions about __has_install_script I feel like that name was deliberate

dylan-conway:
The field should never be used directly was why it had double underscore, but there's also a comment saying this so LGTM

dylan-conway (approved):

Jarred-Sumner:
Maybe we should rename it to @“#has_install_scripts”

dylan-conway:
or we could keep the name `has_install_script` and put `get/set` functions on the enum

PR #9341, opened 2024-03-10
Fix shell `mv` with multiple files

### LINKED ISSUE #9340: Shell `mv` can't move multiple files

1.0.30+1424a196f

```ts
// index.ts
import { $, write } from 'bun';
import { mkdir } from 'node:fs/promises';

await mkdir('foo');
await write('file1', '');
await write('file2', '');
await write('file3', '');

await $`mv file1 file2 file3 foo`;
```

Run the example and check files:
```console
$ bun run index.ts
$ ls
file2 file3 foo index.ts
$ ls foo
file1
```

All specified files should be moved into the target directory.

### What do you see instead?

Only the first file is moved.

Edit: there are actually 3 behaviors:
- If there are between 2 and 5 files: moves only the first file
- If there are between 6 and 9 files: moves only the first file, prints "ls: no such file or directory"
- If there are 10 or more files: moves 1-2 files, then segfaults

fmajestic:
### What does this PR do?

Fixes #9340 

- [x] Code changes

No allocation changes, just a fix of some initialization code.


jdalton:
@fmajestic Thank you for this PR and the unit test! I love that it is less lines of code too.

zackradisic (approved):
Looks good, thanks!

Jarred-Sumner:
@zackradisic i suspect there were merge conflict-related issues causing it to not compile

PR #9345, opened 2024-03-10
Fix bun description

sequencerr:


Electroid (approved):
That's pretty funny. Thanks for fixing

PR #9350, opened 2024-03-10
fix typo in error message

eventualbuddha:
### What does this PR do?

Fixes a typo in an error message. It should be "its" (possessive), not "it's" (contraction of "it is").

- [x] Code changes

- [x] I tried to run zig's test suite and got errors (with `bun`) or never exited (with `bun-debug`)


jdalton:
Thank you @eventualbuddha!

PR #9362, opened 2024-03-11
feat(sqlite): add `fileMustExist` option.

r-rajaneesh:
➕ Added a `fileMustExist` in the database options

### What does this PR do?

This PR adds a check from `better-sqlite3` `fileMustExist` in the options of the constructor.
If `fileMustExist` is set to `true` then it'll not create a file on default else the file is created

- [ ] Code changes

Needed review of this before doing tests.


gvilums:
Thank you for the PR. Do I understand correctly that this new option is the inverse of the `create` option? In that case I don't think we should implement it, as, in my opinion, the convenience of these redundant options is negated by the confusion they cause.

r-rajaneesh:
> Thank you for the PR. Do I understand correctly that this new option is the inverse of the `create` option? In that case I don't think we should implement it, as, in my opinion, the convenience of these redundant options is negated by the confusion they cause.

@gvilums, Jarred and Dave told me to create an issue for this, since `better-sqlite3` has this options and **Bun** does not, even tho it says it implements all the feature in `better-sqlite3`. But I created a PR instead, and `create` was never an option in `better-sqlite3` in the first place.

src/js/bun/sqlite.ts
```diff
@@ -187,6 +187,12 @@ class Database {
     if (typeof options === "object" && options) {
       flags = 0;
 
+      if(!options.fileMustExist) {
+        options.create = true;
+      } else {
+        options.create = false;
+      }
```
Electroid:
```suggestion
      if(options.fileMustExist) {
        options.create = false;
      }
```

Electroid (changes requested):

src/js/bun/sqlite.ts
Jarred-Sumner:
```suggestion
      if(typeof options.fileMustExist === "boolean") {
          options.create = !options.fileMustExist;
      }
```

paperclover: this likely fixes the test failure

Jarred-Sumner:
Closing due to lack of tests

PR #9390, opened 2024-03-13
add @biomejs/biome to trusted dependencies

Arctomachine:
### What does this PR do?
Add @biomejs/biome to default trusted dependencies to run postinstall script

- [ ] Code changes

### DIFF
--- a/src/install/default-trusted-dependencies.txt
+++ b/src/install/default-trusted-dependencies.txt
@@ -14,6 +14,7 @@
 @bazel/rollup
 @bazel/terser
 @bazel/typescript
+@biomejs/biome
 @bufbuild/buf
 @cdktf/node-pty-prebuilt-multiarch
 @ckeditor/ckeditor5-vue


RiskyMH:
After inspecting their postinstall script, I see it does some checks however I dont see it actually helping user as the process shouldnt error: https://github.com/biomejs/biome/blob/main/packages/%40biomejs/biome/scripts/postinstall.js.

paperclover:
thanks for adding biome here, but as RiskyMH said, i dont think we need this in our allowlist

the script in question:

- does not take any system action
- never exits with code >0

therefore, it's affects are invisible to the end user on every package manager, as all modern package managers run all scripts in parallel and only showing output on error.

if there are breakages in biome because of this, we can reconsider, but i will close for now

jdalton:
@Arctomachine See https://twitter.com/jarredsumner/status/1768055758254727218. Next version of bun gets
> `bun install --trust <pkg>` adds to "trustedDependencies"

PR #9397, opened 2024-03-13
Add a helpful error message when the xcode SDK is newer than the current macOS version

Jarred-Sumner:
### What does this PR do?

Adds this warning (pretend instead of "14.4" it said "14.1"). I ran into this and it was very confusing that I had a later version of the macOS SDK than the currently installed version

image

Manually


paperclover (approved):
good.

PR #9398, opened 2024-03-14
WIP Make process.stdout sync on windows

Jarred-Sumner:
### What does this PR do?

- [ ] Code changes


src/bun.js/webcore/streams.zig
```diff
@@ -2891,6 +2892,11 @@ pub const FileSink = struct {
     pub const IOWriter = bun.io.StreamingWriter(@This(), onWrite, onError, onReady, onClose);
     pub const Poll = IOWriter;
 
+    export fn Bun__ForceFileSinkToBeSynchronousOnWindows(globalObject: *JSC.JSGlobalObject, jsvalue: JSC.JSValue) {
```
paperclover: `comptime std.debug.assert(Environment.isWindows)`

this should not be even looked at outside of windows

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -1673,7 +1676,19 @@ static JSValue constructStdioWriteStream(JSC::JSGlobalObject* globalObject, int
...
+    ASSERT_WITH_MESSAGE(JSC::isJSArray(result), "Expected an array from getStdioWriteStream");
+    JSC::JSObject *resultObject = JSC::jsCast<JSC::JSObject*>(result);
+
+#if OS(WINDOWS)
+    // Node.js docs - https://nodejs.org/api/process.html#a-note-on-process-io
+    // Files: synchronous on Windows and POSIX
+    // TTYs (Terminals): asynchronous on Windows, synchronous on POSIX
+    // Pipes (and sockets): synchronous on Windows, asynchronous on POSIX
+    // Synchronous writes avoid problems such as output written with console.log() or console.error() being unexpectedly interleaved, or not written at all if process.exit() is called before an asynchronous write completes. See process.exit() for more information.
+    Bun__ForceFileSinkToBeSynchronousOnWindows(globalThis, JSValue::encode(resultObject->getDirectIndex(1)));
```
paperclover: the comment should explain more, as right now it raises the question 'if a stdio / TTY is said above to be "asynchronous on Windows", why is `ForceFileSinkToBeSynchronousOnWindows` called?'

when i quote other resources while adding onto it to explain descisions on code, i sometimes use a markdown blockquote

```
// Node.js docs - https://nodejs.org/api/process.html#a-note-on-process-io
// > Files: synchronous on Windows and POSIX
// > TTYs (Terminals): asynchronous on Windows, synchronous on POSIX
// > Pipes (and sockets): synchronous on Windows, asynchronous on POSIX
// > Synchronous writes avoid problems such as output written with console.log() or console.error() being unexpectedly interleaved, or not written at all if process.exit() is called before an asynchronous write completes. See process.exit() for more information.
// 
// etc etc Bun makes them syncronous on windows to match posix behavior etc etc issue etc etc
```

src/__global.zig
```diff
@@ -99,6 +99,9 @@ pub fn exitWide(code: u32) noreturn {
     runExitCallbacks();
     Output.flush();
     std.mem.doNotOptimizeAway(&Bun__atexit);
+    if (Environment.isWindows) {
```
jdalton, line 102: comptime?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -1673,7 +1677,19 @@ static JSValue constructStdioWriteStream(JSC::JSGlobalObject* globalObject, int
...
+    ASSERT_WITH_MESSAGE(JSC::isJSArray(result), "Expected an array from getStdioWriteStream");
+    JSC::JSArray* resultObject = JSC::jsCast<JSC::JSArray*>(result);
+
+#if OS(WINDOWS)
+    // Node.js docs - https://nodejs.org/api/process.html#a-note-on-process-io
+    // > Files: synchronous on Windows and POSIX
+    // > TTYs (Terminals): asynchronous on Windows, synchronous on POSIX
+    // > Pipes (and sockets): synchronous on Windows, asynchronous on POSIX
+    // > Synchronous writes avoid problems such as output written with console.log() or console.error() being unexpectedly interleaved, or not written at all if process.exit() is called before an asynchronous write completes. See process.exit() for more information.
+    Bun__ForceFileSinkToBeSynchronousOnWindows(globalThis, JSValue::encode(resultObject->getIndex(globalObject, 1)));
```
jdalton, line 1684: This method name and comments are stroooong 💪

src/bun.js/webcore/blob.zig
```diff
@@ -2961,16 +2962,48 @@ pub const Blob = struct {
...
             var sink = JSC.WebCore.FileSink.init(fd, this.globalThis.bunVM().eventLoop());
 
-            switch (sink.writer.start(fd, false)) {
-                .err => |err| {
-                    globalThis.vm().throwError(globalThis, err.toJSC(globalThis));
-                    sink.deref();
+            if (is_stdout_or_stderr) {
+                switch (sink.writer.startSync(fd, false)) {
+                    .err => |err| {
+                        globalThis.vm().throwError(globalThis, err.toJSC(globalThis));
```
jdalton, line 2990: What's the distinction here between bunVM() and the vm()?

src/bun.js/webcore/streams.zig
```diff
@@ -3029,7 +3037,7 @@ pub const FileSink = struct {
         // TODO: this should be concurrent.
         const fd = switch (switch (options.input_path) {
             .path => |path| bun.sys.openA(path.slice(), options.flags(), options.mode),
-            .fd => |fd_| bun.sys.dupWithFlags(fd_, if (bun.FDTag.get(fd_) == .none) std.os.O.NONBLOCK else 0),
+            .fd => |fd_| bun.sys.dupWithFlags(fd_, if (bun.FDTag.get(fd_) == .none and !this.force_sync_on_windows) std.os.O.NONBLOCK else 0),
```
jdalton, line 3058: Any chance with the `!this.force_sync_on_windows` on 3040 and 3063 that it's the cheaper check?
If so moving it to the front of the `and` chain is the recommendation.

src/bun.js/webcore/streams.zig
```diff
@@ -3375,20 +3402,21 @@ pub const FileReader = struct {
...
+                //     if (comptime Environment.isPosix) {
+                //         if (bun.FDTag.get(file.pathlike.fd) == .none and !(file.is_atty orelse false)) {
+                //             break :brk std.os.O.NONBLOCK;
+                //         }
+                //     }
+
+                //     break :brk 0;
+                // })) {
+                //     .result => |_fd| if (Environment.isWindows) bun.toLibUVOwnedFD(_fd) else _fd,
+                //     .err => |err| {
```
jdalton, line 3434: Dead code ✂️

src/bun.zig
```diff
@@ -2156,7 +2156,17 @@ pub const FDTag = enum {
     stdout,
     pub fn get(fd_: anytype) FDTag {
         const fd = toFD(fd_);
+        const T = @TypeOf(fd_);
         if (comptime Environment.isWindows) {
+            if (@typeInfo(T) == .Int or @typeInfo(T) == .ComptimeInt) {
```
jdalton, line 2175: Would `T` be a `.ComptimeInt` if `get()` was called as part of a `comptime` ?

src/io/PipeWriter.zig
```diff
@@ -1185,6 +1228,9 @@ pub fn WindowsStreamingWriter(
             this.current_payload = this.outgoing;
             this.outgoing = temp;
             switch (pipe) {
+                .sync_file => {
+                    @panic("sync_file pipe write should not be reachable");
```
jdalton, line 1236:
```suggestion
                    @panic("sync_file pipe write should not be reachable - writeLatin1 in PipeWriter.zig");
```
I like the helpfulness of both this and https://github.com/oven-sh/bun/pull/9398/files#diff-dde34f421656d5d1df29476c8c358985d1b22bae2c505c70e0f3d922e0af9fe5R949 but looking for consistency in direction or information

src/io/PipeWriter.zig
```diff
@@ -1233,11 +1279,39 @@ pub fn WindowsStreamingWriter(
...
 
+            if (this.source != null and this.source.? == .sync_file) {
+                defer this.outgoing.reset();
+                var remain = StreamBuffer.writeOrFallback(&this.outgoing, buffer, comptime writeFn) catch {
+                    return .{ .err = bun.sys.Error.oom };
+                };
+                const initial_len = remain.len;
+                const fd = bun.toFD(this.source.?.sync_file.file);
+
+                while (remain.len > 0) {
```
jdalton, line 1294:
```suggestion
                while (true) {
```
both the condition and the one in `.result` do a 0 check which may be okay.

src/io/PipeWriter.zig
```diff
@@ -1246,15 +1320,15 @@ pub fn WindowsStreamingWriter(
         }
 
         pub fn writeUTF16(this: *WindowsWriter, buf: []const u16) WriteResult {
-            return writeInternal(this, buf, StreamBuffer.writeUTF16);
+            return writeInternal(this, buf, &StreamBuffer.writeUTF16);
```
jdalton, line 1327: with this being so tiny is this a candidate for:
`pub inline fn writeUTF16`?

(same with the ones below)

src/io/source.zig
```diff
@@ -8,92 +8,94 @@ pub const Source = union(enum) {
...
     pub fn getHandle(this: Source) *uv.Handle {
-        switch (this) {
-            .pipe => return @ptrCast(this.pipe),
-            .tty => return @ptrCast(this.tty),
-            .file => unreachable,
-        }
+        return switch (this) {
+            .pipe => @ptrCast(this.pipe),
+            .tty => @ptrCast(this.tty),
+            .sync_file, .file => unreachable,
```
jdalton, line 42: Instead of unreachable would returning `null` equiv be a thing? With it `unreachable` it's like a poison method if the Source is .file or sync_file

src/js/builtins/ProcessObjectInternals.ts
```diff
@@ -59,11 +53,10 @@ export function getStdioWriteStream(fd) {
     };
   }
 
-  stream._type = "tty";
   stream._isStdio = true;
   stream.fd = fd;
 
-  return stream;
+  return [stream, stream[require("internal/shared").fileSinkSymbol]];
```
jdalton, line 59: Returning the stream plus a reference to the symbol seems kind of hacky. Instead can we just expose the symbol to  C++ separately.

src/js/node/tty.ts
```diff
@@ -139,17 +136,16 @@ Object.defineProperty(WriteStream, "prototype", {
...
 
     WriteStream.prototype.clearScreenDown = function (cb) {
-      return (readline ??= require("node:readline")).clearScreenDown(this, cb);
+      return require("node:readline").clearScreenDown(this, cb);
     };
 
     WriteStream.prototype.cursorTo = function (x, y, cb) {
-      return (readline ??= require("node:readline")).cursorTo(this, x, y, cb);
+      return require("node:readline").cursorTo(this, x, y, cb);
     };
```
jdalton, line 150: The bits here aren't equiv. At least in the transform step I'm at. We currently don't hoist the `require("node:readline")` -> `@getInternalField(@internalModuleRegistry, 36) || @createInternalModuleById(36)` out.

Issue #8640, opened 2024-02-02
import 'foo%20bar' should follow the URL convention


Fil:
### What version of Bun is running?

1.0.25+a8ff7be64

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

I have a file that I can import from, with a space in its name: `'foo bar.js'`

If I run the script below, node, deno and major browsers will all use this file.

```js
import foobar from "./foo%20bar.js";
console.log({ foobar });
```

### What is the expected behavior?

show the import from `'foo bar.js'`

### What do you see instead?

bun literally imports from a file called `'foo%20bar.js'`.

### Additional information

bun's approach would arguably be the right thing to do if we were starting from scratch, but that ship has sailed, and bun should follow the convention and import from `'foo bar.js'` instead.

Note that `require` follows the same pattern.

Test repo: https://github.com/Fil/importz/

PR #9399, opened 2024-03-14
make `import.meta` behave better when query string is provided

### LINKED ISSUE #8538: [windows] import.meta.url causes Segmentation Fault when import contains a query parameters

~1.0.26-canary.79+5a0ae3c25~
~1.0.26-canary.4+1f0c1c81e~
~1.0.26-canary.63+d0b6218bd~
~1.0.33-canary.10+931e04b01~

**UPD. Fixed in 1.0.36-canary.40+6164fac25**

Microsoft Windows NT 10.0.19044.0 x64

1. Create file **main.ts**
```ts
import './test?param=value'
```
2. Create file **test.ts**
```ts
console.log('file test.ts', import.meta.url)
```
3. Run `bun run main.ts` multiple times

```sh
PS D:\app> bun run main.ts
file test.ts file:///D:/app/test.ts%3Fparam=value
```

### What do you see instead?

Result is not stable, sometimes it works fine
```sh
PS D:\app> bun run main.ts
file test.ts file:///D:/app/test.ts%3Fparam=value
PS D:\app> bun run main.ts
file test.ts file:///D:/app/test.ts%3Fparam=value
PS D:\app> bun run main.ts
Segmentation fault at address 0x2
???:?:?: 0x7ff7e79650e4 in ??? (bun.exe)
???:?:?: 0x7ff7e7964d65 in ??? (bun.exe)
???:?:?: 0x7ff7e791daed in ??? (bun.exe)
???:?:?: 0x7ff7e791e513 in ??? (bun.exe)
???:?:?: 0x7ff7e791c916 in ??? (bun.exe)
???:?:?: 0x1858000115d in ??? (???)
???:?:?: 0x1858000111f in ??? (???)
PS D:\app> bun run main.ts
file test.ts 荸Ȅ胸Ȅ⼃⼯㩃啜愐Ȅ䑜獥瑫䍐Ȅ嬐Ȅ@혊쀣ༀ退ȄȄȄ
PS D:\app> bun run main.ts
Segmentation fault at address 0x8ffe993c4
???:?:?: 0x7ff7e9001f89 in ??? (bun.exe)
???:?:?: 0x7ff7e798116a in ??? (bun.exe)
???:?:?: 0x7ff7e81df51c in ??? (bun.exe)
???:?:?: 0x7ff7e81de991 in ??? (bun.exe)
???:?:?: 0x7ff7e909b77c in ??? (bun.exe)
???:?:?: 0x174fec8af27 in ??? (???)
???:?:?: 0x7ff7e77db2a1 in ??? (bun.exe)
???:?:?: 0x174fec94127 in ??? (???)
```

### Additional information

This code works fine in WSL

paperclover:
### What does this PR do?

Fixes #8538

While investigating this crash, I noticed import.meta is just wrong in this situation

image

So this fixes that issue, which in turn just fixes the Windows crash for me.

What is hard about this situation is that our resolver does not actually handle URLs at all. It believes everything is a file path, which is troubling, because it is impossible to import or run a file with a `?` in it, as it thinks it includes a query string. And using %3F does not work either due to #8640

image

To prevent making potentially large changes to the resolver and module loading code, there is a best effort attempt to make current cases work, as well as a new assertion in `ImportMetaObject::create` to make sure it is not possible to create a bad import meta object.

Added a new test, including `import.meta` in a `.cjs` file.

### DIFF
--- a/src/bun.js/bindings/ImportMetaObject.cpp
+++ b/src/bun.js/bindings/ImportMetaObject.cpp
@@ -1,3 +1,4 @@
+#include "JavaScriptCore/JSCJSValue.h"
 #include "root.h"
 #include "headers.h"
 
@@ -46,6 +47,7 @@
 #include "CommonJSModuleRecord.h"
 #include <JavaScriptCore/JSPromise.h>
 #include "PathInlines.h"
+#include "wtf/text/StringView.h"
 
 namespace Zig {
 using namespace JSC;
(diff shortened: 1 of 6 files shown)


Jarred-Sumner:
> What is hard about this situation is that our resolver does not actually handle URLs at all. It believes everything is a file path, which is troubling, because it is impossible to import or run a file with a ? in it, as it thinks it includes a query string. And using %3F does not work either due to https://github.com/oven-sh/bun/issues/8640

This shouldn't be troubling - this is correct. Module resolution involving tsconfig.json, package.json, or node_modules should continue to be about filesystem paths and not URLs. URLs are unambiguous. Filesystem paths can be ambiguous.

Instead we should just convert URLs to file paths, which we should've been already doing. Query strings should not be visible to the module resolver.

PR #9417, opened 2024-03-14
fix(windows) fix some test failures

cirospaciari:
### What does this PR do?

Should fix test:

- test\js\bun\http\serve.test.ts 
- test\js\bun\http\fetch-file-upload.test.ts
- test\js\bun\http\bun-server.test.ts 

- [x] Code changes

Existing Tests


src/bun.js/webcore/blob/ReadFile.zig
```diff
@@ -723,7 +723,9 @@ pub const ReadFileUV = struct {
     }
 
     fn remainingBuffer(this: *const ReadFileUV) []u8 {
-        return this.buffer.unusedCapacitySlice();
+        var remaining = this.buffer.unusedCapacitySlice();
+        remaining = remaining[0..@min(remaining.len, this.max_length -| this.read_off)];
+        return remaining;
```
jdalton: The saturating subtraction operator is haaandy! ❤️

jdalton:
```suggestion
        return remaining[0..@min(remaining.len, this.max_length -| this.read_off)];
```

src/http.zig
```diff
@@ -122,7 +122,7 @@ pub const Sendfile = struct {
     content_size: usize = 0,
 
     pub fn isEligible(url: bun.URL) bool {
-        return url.isHTTP() and url.href.len > 0 and FeatureFlags.streaming_file_uploads_for_http_client;
+        return url.isHTTP() and url.href.len > 0 and FeatureFlags.streaming_file_uploads_for_http_client and Environment.isPosix;
```
jdalton: Leading with the cheap check first allows to bail out of a chain of `and`s faster.
Here though we can comptime this _(unless the compiler gets grumpy about it already being so)_. If the concern is Windows as the `!Environment.isPosix` we can do `Environment.isWindows` as `pub const isPosix = !isWindows and !isWasm;` also excludes `isWasm` which may be desired here _(if so disregard)_.
```suggestion
        if (comptime Environment.isWindows or 
            comptime !FeatureFlags.FeatureFlags.streaming_file_uploads_for_http_client) return false;
        return url.href.len > 0 and url.isHTTP();
```

test/js/bun/http/serve.test.ts
```diff
@@ -1179,7 +1179,7 @@ it("request body and signal life cycle", async () => {
     expect(true).toBe(true);
     server.stop(true);
   }
-});
+}, 30_000);
```
jdalton, line 1182: I dig numeric separators (though guilty of not using them enough).

src/http.zig
cirospaciari: thanks https://github.com/oven-sh/bun/pull/9417/commits/0598003a36ac83f00854c9bdf2a19ef2c71cd80f added here

src/bun.js/api/server.zig
```diff
@@ -3240,7 +3240,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                                 var prev = body.value.Locked.readable;
                                 body.value.Locked.readable = .{};
                                 readable.value.ensureStillAlive();
-                                prev.deinit();
+                                defer prev.deinit();
```
jdalton, line 3243: In the future some kind of lint could help us with this. It's easy to miss a `defer`.

Jarred-Sumner: @cirospaciari are you sure this fixes anything? readable.value.ensureStillAlive() should make sure its visible to GC


> In the future some kind of lint could help us with this. It's easy to miss a `defer`.

pretty hard to tell when it's necessary

Jarred-Sumner: hm we did do this for fetch as well

PR #9424, opened 2024-03-15
Don't link libicudata?

Jarred-Sumner:
### What does this PR do?

An attempt to address the size regression from Bun v1.0.27 on macOS arm64

CI


jdalton:
I dig that this reduced complexity and may have solved the issue. Very cool.

PR #9426, opened 2024-03-15
`bun pm trust` bugfix

dylan-conway:
### What does this PR do?
ignores file not found errors when checking if a package has scripts or not. this prevents crashing on dependencies that aren't installed.

added a test for `bun pm trust` and `untrusted` with a missing dependency in `node_modules`


test/cli/install/registry/bun-install-registry.test.ts
```diff
@@ -3711,6 +3711,77 @@ for (const forceWaiterThread of [false, true]) {
       expect(await exists(join(packageDir, "postinstall.txt"))).toBeTrue();
     });
 
+    test.only("bun pm trust and untrusted on missing package", async () => {
```
Jarred-Sumner:
```suggestion
    test("bun pm trust and untrusted on missing package", async () => {
```

src/cli/pm_trusted_command.zig
```diff
@@ -263,14 +268,19 @@ pub const TrustCommand = struct {
...
                         pm.lockfile,
                         node_modules_dir,
                         abs_node_modules_path.items,
                         alias,
                         resolution,
-                    )) |scripts_list| {
+                    ) catch |err| {
+                        if (err == error.ENOENT) continue;
+                        return err;
+                    };
```
jdalton, line 281: Is the granularity of the error too wide/large? What I mean is that if one of the scripts fails to have a file does getting the list `getList` fail for all? Should the `error.ENOENT` be done at the [`getList`](https://github.com/oven-sh/bun/blob/e2567512188b3d7e9926b8716767bebac9488619/src/install/lockfile.zig#L2620) level or maybe a new method `.getListOnDisk`

Issue #8939, opened 2024-02-16
implement completions for powershell


paperclover:
It is possible to declare completions: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/register-argumentcompleter?view=powershell-7.4

Probably wont have this for Bun 1.1, but logging the issue so it is not forgotten.

Tagging this as windows as almost all users use powershell on windows, and not on mac and linux, but this is technically an all-platforms situation.

bcdev-com:
A consideration before accepting this.  

As someone who uses PowerShell as their daily driver, if a tool silently adds a thousand lines of script to the end of my profile.ps1 without warning or consent, I'll instantly ban it from all systems I'm responsible for, no matter it's other merits.

Truth is, there's some pretty unpleasant PowerShell used in install.ps1 as well, but I could probably ignore the rest of it.

bcdev-com:
Entirely separate from the previous comment, before installing completions at all, it is important to understand the difference between profile.ps1 and Microsoft.PowerShell_profile.ps1.  It's probably the latter that you would want to offer to at a single line calling your external completion script to, not the former.

paperclover:
> if a tool silently adds a thousand lines of script to the end of my profile.ps1 without warning or consent, I'll instantly ban it from all systems I'm responsible for, no matter it's other merits.

interesting. bun's bash install script currently does this, which some people very much do not like.

> Truth is, there's some pretty unpleasant PowerShell used in install.ps1 as well

i remember having to do many weird things in this script to get the script to work reliably on all testing devices as well as report helpful errors when possible (one example is irm is slow in powershell 5, but curl sometimes fails, and iwr can fail on newly installed systems). im very interested in what specifically you find unpleasant.

bcdev-com:
> interesting. bun's bash install script currently does this, which some people very much do not like.

The best practice version would be something more like the dotnet SDK does.  PowerShell completions for their CLI are handled by invoking the CLI itself from a tiny bit of PowerShell.  Like this:
```powershell
Register-ArgumentCompleter -Native -CommandName dotnet {
    param($commandName, $wordToComplete, $cursorPosition)
    dotnet complete --position $cursorPosition "$wordToComplete" | ForEach-Object {
        [System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
    }
}
```

[...]

PR #9441, opened 2024-03-15
Add PowerShell completions

sequencerr:
### What does this PR do?

Closes: https://github.com/oven-sh/bun/issues/8939

- [ ] Code changes

### DIFF
--- a/completions/bun.ps1
+++ b/completions/bun.ps1
@@ -1,0 +1,1147 @@
+$_bun_subcommands = @{
+  run         = 'Run JavaScript with Bun, a package.json script, or a bin'
+  test        = 'Run unit tests with Bun'
+  x           = 'Install and execute a package bin (bunx)'
+  repl        = 'Start a REPL session with Bun'
+  init        = 'Start-Process an empty Bun project from a blank template'
+  create      = 'Create a new project from a template (bun c)'
+  install     = 'Install dependencies for a package.json (bun i)'
+  add         = 'Add a dependency to package.json (bun a)'
+  remove      = 'Remove a dependency from package.json (bun rm)'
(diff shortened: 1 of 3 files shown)


paperclover:
have not read through in depth but this looks legendary

sequencerr:
<3 thanks. there are still some todos, link unlink completions, replace `get bins` with `bun.exe get completes [a-z]` (most are crashing on win). and zig part is pseudocoded (need help with this)

completions/bun.ps1
```diff
@@ -0,0 +1,1147 @@
...
+}
+
+$_bun_register_completer = {
+  param($wordToComplete, $commandAst, $cursorPosition)
+
+  $param = & $_bun_current_param $commandAst $cursorPosition
+  ((& $_bun_get_completes $commandAst $param) | Where-Object { $_.CompletionText -like "$wordToComplete*" })
+}
+
+Microsoft.PowerShell.Core\Register-ArgumentCompleter -CommandName bun -Native -ScriptBlock $_bun_register_completer
```
sequencerr, line 1147: It's possible to define `bun, bun.exe` at once. Do we need this?

paperclover (changes requested):
I think the approach for this should deviate from what was taken for bash, zsh, and fish. See [this comment](https://github.com/oven-sh/bun/issues/8939#issuecomment-2033044246) in the powershell completions issue.

Instead of a 1.1k line addition to the user's powershell startup profile, this would be better done as a part of `bun getcompletes` and uses what bcdev-com suggests in [this comment](https://github.com/oven-sh/bun/issues/8939#issuecomment-2040644434)

PR #9443, opened 2024-03-15
fix(shell) fix ref counting onReaderDone

cirospaciari:
### What does this PR do?

- [x] Code changes

Existing tests


Jarred-Sumner:
Thank you for this. 

Is there anywhere else in the shell it’s likely we need to do this?

cirospaciari:
> Thank you for this.
> 
> Is there anywhere else in the shell it’s likely we need to do this?

Looking at the code more careful I see that we also call `signalDoneToCmd` in `onReaderError` we should do the same their @zackradisic you probably have more context about the shell we should look at places where `detach` is possible and we are still accessing the `PipeReader` after it.

PR #9444, opened 2024-03-15
fix: return object from `FileHandle#read` and `FileHandle#readv`

eventualbuddha:
### What does this PR do?

Fixes the `FileHandle#read` and `FileHandle#readv` methods from `fs/promises` to return an object with bytes read and buffer(s) instead of just the number of bytes read. This brings these methods in line with the NodeJS implementation. See https://nodejs.org/api/fs.html#filehandlereadoptions.

- [x] Code changes

I wrote automated tests

- [x] I tried to run the automated tests locally, but `bun setup` failed (presumably unrelated to my changes):

```
fatal: transport 'file' not allowed
fatal: Fetched in submodule path 'src/deps/base64', but it did not contain 3a5add8652076612a8407627a42c768736a4263f. Direct fetching of that commit failed.
```

### DIFF
--- a/src/js/node/fs.promises.ts
+++ b/src/js/node/fs.promises.ts
@@ -301,7 +301,8 @@ class FileHandle extends EventEmitter {
 
     try {
       this[kRef]();
-      return real_export.read(fd, buffer, offset, length, position);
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Looks great. Thank you.

PR #9463, opened 2024-03-16
Add shell hanging tests

Jarred-Sumner:
### What does this PR do?

@zackradisic here are tests that are failing in v1.0.31 and canary but not failing in v1.0.30


Jarred-Sumner:
Thank you

PR #9468, opened 2024-03-17
Fix assertion failure when migrating package-lock.json

Jarred-Sumner:
### What does this PR do?

When a package-lock.json is out of sync with the package.json and we attempt to migrate it, there is an integer overflow that can occur when adding up how many packages were added for the summary.

This doesn't seem to be an issue in release fast builds since we use `@truncate`, however it does come up in both debug and windows builds.

This doesn't feel like the best fix. It feels like we shouldn't be in this state in the first place. But i'm not 100% sure.

Fixes the following assertion failure:

```zig
bun install v1.0.32-canary.16 (17631ce6)
[0.41ms] migrated lockfile from package-lock.json

Panic: integer overflow

bun has crashed :'(

----- bun meta -----
Bun v1.0.32-canary.16 (17631ce6) Windows x64
InstallCommand:
Elapsed: 10ms | User: 0ms | Sys: 0ms
RSS: 55.74MB | Peak: 55.74MB | Commit: 90.94MB | Faults: 13770
----- bun meta -----

0   00007FF7F19CB9F8
1   ???
2   ???
3   ???
4   ???
5   ???
6   ???
7   ???
8   ???
9   ???

Search GitHub issues https://bun.sh/issues or join in #windows channel in https://bun.sh/discord

thread 33364 panic: integer overflow
???:?:?: 0x7ff7f050c612 in ??? (bun.exe)
???:?:?: 0x7ff7f0576af0 in ??? (bun.exe)
???:?:?: 0x7ff7f058ed7b in ??? (bun.exe)
???:?:?: 0x7ff7f063b2ca in ??? (bun.exe)
???:?:?: 0x7ff7f038157f in ??? (bun.exe)
???:?:?: 0x7ff7f2e81b23 in ??? (bun.exe)
???:?:?: 0x7ff93a98257c in ??? (KERNEL32.DLL)
???:?:?: 0x7ff93b88aa57 in ??? (ntdll.dll)
```

There is a test

### DIFF
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -729,43 +729,6 @@ pub fn cleanWithLogger(
         try old.preprocessUpdateRequests(updates, exact_versions);
     }
 
-    // Deduplication works like this
-    // Go through *already* resolved package versions
-    // Ask, do any of those versions happen to match a lower version?
-    // If yes, choose that version instead.
-    // Why lower?
-    //
-    // Normally, the problem looks like this:
-    //   Package A: "react@^17"
-    //   Package B: "react@17.0.1
-    //
-    // Now you have two copies of React.
-    // When you really only wanted one.
-    // Since _typically_ the issue is that Semver ranges with "^" or "~" say "choose latest", we end up with latest
-    // if (options.enable.deduplicate_packages) {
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
@dylan-conway please take a look at this when you get a chance

Issue #8556, opened 2024-01-29
Add "--trust" argument to "bun add" command to add the dependency to "trustedDependencies"


pierre-cm:
### What is the problem this feature would solve?

When adding a new dependency via `bun add` command, `postinstall` script is not executed by default. In most cases, `postinstall` script execution is necessary to have the dependency properly installed. 

Currently, the only solution is to manually add the dependency to package.json `trustedDependencies` list before adding it via bun cli. I understand that this is intentional, and I share the security concerns this feature aims to address. However, it can be somewhat inconvenient having to modify the package.json file manually everytime.

### What is the feature you are proposing to solve the problem?

Add a `--trust` flag to `bun add` command that automatically adds the dependency to `trustedDependencies` in package.json before installing the dependency.

paperclover:
im surprised this doesnt have an issue already open. this was something i said was a blocker for trustedDependencies but it seems it was never included as a part of that.

dylan-conway:
added in #8943

PR #9471, opened 2024-03-17
Fixes #9433

### LINKED ISSUE #9433: 1.0.31 bun install fails to resolve packages in devDependencies

1.0.31

Linux 5.15.0-88-generic x86_64 x86_64

- Have package.json with devDependencies, a bun.lockb and no node_modules, like so :
```json
{
  "devDependencies": {
    "@vitejs/plugin-vue": "^5.0.4",
    "eslint": "^8.38.0",
    "eslint-config-standard": "^17.0.0",
    "eslint-plugin-vue": "^9.10.0",
    "vite": "^5.1.6"
  }
}
```
- Then install some of the devDependencies listed but not all, like so : `bun install vite @vitejs/plugin-vue`

Only the packages mentionned should be installed, without error

### What do you see instead?

```
bun add v1.0.31 (e2567512)
error: eslint@^8.38.0 failed to resolve
error: eslint-config-standard@^17.0.0 failed to resolve
error: eslint-plugin-vue@^9.10.0 failed to resolve
```
And a gitlab runner crash, which errors out the pipleline

### Additional information

Happens on my gitlab runner and on my ubuntu VM, does not happen on bun 1.0.30. Have a nice day :smil

Jarred-Sumner:
### What does this PR do?

This PR does a few different things.

#### Verify the metadata hash before installing, save the metadata hash after installing

Fixes #9433

An important part of `--frozen-lockfile` or `--production` is to fail the installation early. It should fail before installing packages. If you're passing one of these flags, you're saying "if there were changes, CI should fail as early as possible".  Ideally, it would fail at the package resolution step.

In #8556, the metadata hash used for checking if the contents of two lockfiles are identical was changed to generate after  installing packages instead of before installing packages. The metadata hash is used to know if a frozen lockfile was changed. This means that `--frozen-lockfile` or `--production` now fail after installing instead of before installing. That's not good, so this PR fixes that by checking the metadata hash before installing and failing the install if it does not match.

However, we may not have all the scripts or a fully normalized state until after the install completes (due to trusted dependencies and scripts and other things). So we want to generate the metadata hash only after installing. We only need to generate the metadata hash when we're going to save the file.

##### Disable saving lockfiles when `--frozen-lockfile` or `--production` is passed

Currently, `bun install --production` will save the lockfile if there were changes unrelated to package resolution. This puts `bun install` in a very weird state which only continues to work if ever subsequent run also passes `--production`.

This PR disables saving the lockfile when `--production` or `--frozen-lockfile` is passed (or equivalent options are specified in `bunfig.toml`)

##### Disable `--trust` and `--frozen-lockfile` or `--production` together

The information necessary for `--trust` is not always known until after installing packages, so if we want `--frozen-lockfile` and `--production` to fail before installing packages, we also should disable `--trust` when used with `--frozen-lockfile` or `--production`. It doesn't make as much sense anyway in a CI environment.

##### Support conflicting `"devDependencies"` and `"dependencies"`

This makes it so the following package.json installs successfully (it chooses `4.17.20`):
```json
{
  "devDependencies": {
    "lodash": "4.17.20"
  },
  "dependencies": {
    "lodash": "4.17.21"
  }
}
```
This matches npm's behavior, preferring `devDependencies` over `dependencies` when both are specified.

Previously, this failed with the following message:
```zig
❯ bun i
bun install v1.0.32 (7c9f0763)
3 |     "lodash": "4.17.20"
        ^
warn: Duplicate dependency: "lodash" specified in package.json
   at /Users/jarred/Desktop/reproo/package.json:3:5

6 |     "lodash": "4.17.21"
                  ^
note: "lodash" originally specified here
   at /Users/jarred/Desktop/reproo/package.json:6:15
error: Package "lodash@4.17.20" has a dependency loop
  Resolution: "lodash@4.17.21"
  Dependency: "lodash@4.17.21"

error: DependencyLoop

----- bun meta -----
Bun v1.0.32 (7c9f0763) macOS Silicon 23.4.0
InstallCommand: bunfig
Elapsed: 8ms | User: 12ms | Sys: 7ms
RSS: 21.81MB | Peak: 21.81MB | Commit: 34.62MB | Faults: 0
----- bun meta -----

0   0x10312ac94
1   ???

Crash report saved to:
  ~/.bun/.bun-crash/v1.0.32-1710669645743.crash

Search GitHub issues https://bun.sh/issues or ask for #help in https://bun.sh/discord
```

Now, it shows this:
```zig
bun install v1.0.32 (895a1045)
3 |     "lodash": "4.17.20"
        ^
warn: Duplicate dependency: "lodash" specified in package.json
   at /Users/jarred/Desktop/reproo/package.json:3:5

6 |     "lodash": "4.17.21"
                  ^
note: "lodash" originally specified here
   at /Users/jarred/Desktop/reproo/package.json:6:15

Checked 1 install across 3 packages (no changes) [7.00ms]
```

There are tests.


Jarred-Sumner:
@dylan-conway please review this when you get a chance

dylan-conway:
One thing: if the package in `dependencies` and `devDependencies` is the same version, it is printed twice in the summary
Screenshot 2024-03-18 at 6 06 55 PM

PR #9472, opened 2024-03-17
Fixes #9430

### LINKED ISSUE #9430: Bun v1.0.31 is taking too long to install packages in Github Actions

v1.0.31

Github Actions

Not clear yet about how to make a correct reproduction. But my CI runs in Github actions are taking a lot of time and are timeouting during `bun install`. My bunfig file:

```toml
[install]
exact = true

[install.lockfile]
save = false
```

Another important info is that I'm using the workspaces feature in package.json:

```
"workspaces": [
  "packages/*"
],
```

So because of this monorepo arrangement, I'm not using lock files to avoid too many branch refreshes due to conflicts in bun.lockb

To have the normal blazing fast bun install times.

### What do you see instead?

Timeouts of 20 minutes in my Github Actions workflows.

### Additional information

I'm using [moonrepo](https://moonrepo.dev/docs) for orchestrating the tasks in the monorepo.

Not sure if this is a clear regression but rolling back to v1.0.30 solves the issue in Github Actions.

Also, using v1.0.31 locally in a Mac (Darwin 23.3.0 arm64 arm) doesn't raise the issue locally.

I'm sending this bug report because others may have the same issue.

Jarred-Sumner:
### What does this PR do?

Fixes #9430

`bun install` has a max number of concurrent lifecycle scripts

Every time we start a script, we incrmeent that number

The bug was was that we never decremented that number :(

Would love your help writing a test for this @dylan-conway

### DIFF
--- a/src/bun.js/api/bun/process.zig
+++ b/src/bun.js/api/bun/process.zig
@@ -543,6 +543,28 @@ pub const Status = union(enum) {
             else => null,
         };
     }
+
+    pub fn format(self: @This(), comptime _: []const u8, _: anytype, writer: anytype) !void {
+        if (self.signalCode()) |signal_code| {
+            if (signal_code.toExitCode()) |code| {
+                try writer.print("code: {d}", .{code});
+                return;
+            }
+        }
+
+        switch (self) {
+            .exited => |exit| {
+                try writer.print("code: {d}", .{exit.code});
+            },
+            .signaled => |signal| {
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
I'm confident it fixes it

https://github.com/Jarred-Sumner/bun-install-bug-repro-9430/actions/runs/8316085891/job/22755142530

PR #9475, opened 2024-03-17
Fixes #9459 (shell mv regression)

### LINKED ISSUE #9459: Regression on Bun.$: `mv` causes `ls: No such file or directory` or `ls: Illegal byte sequence`

1.0.31+e25675121 & 1.0.32+c9a70f168

Darwin 23.2.0 x86_64 i386

```sh
touch tmp.a && bun --print 'await Bun.$`mv tmp.a tmp.b`'; rm -rf tmp.a
# ls: No such file or directory
```

```sh
bun --print 'await Bun.$`touch /tmp/a`; await Bun.$`mv /tmp/a /tmp/b`'
# ls: Illegal byte sequence
```

This was working in `1.0.30`

### What do you see instead?

_No response_

### Additional information

_No response_

zackradisic:
### What does this PR do?

Fixes #9459 and also adds `mv` tests

The target path in a `mv` command was pointing to undefined memory

- [x] Code changes

Zig files changed:

### DIFF
--- a/src/shell/interpreter.zig
+++ b/src/shell/interpreter.zig
@@ -7103,7 +7103,9 @@ pub const Interpreter = struct {
 
                     switch (Syscall.renameat(this.cwd, src, this.cwd, this.target)) {
                         .err => |e| {
-                            this.err = e;
+                            if (e.getErrno() == .NOTDIR) {
+                                this.err = e.withPath(this.target);
+                            } else this.err = e;
                         },
                         else => {},
                     }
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
- failing on windows
- missing tests that use glob
- missing tests that try to move a _lot_ of files
- missing tests that check what happens when strange unicode characters are used

Jarred-Sumner:
but great progress overall, merging

PR #9479, opened 2024-03-17
Development environment documentation for openSUSE Tumbleweed

nellfs:
### What does this PR do?

- [x] Documentation

I made a change in `CMakeLists.txt` to add `openSUSE Tumbleweed` to the list of Linux distributions, I tested this PR on my computer and on a fresh virtual machine to make sure that `bun setup` works. 😊


CMakeLists.txt
```diff
@@ -278,7 +278,7 @@ endif()
 if(UNIX AND NOT APPLE)
     execute_process(COMMAND grep -w "NAME" /etc/os-release OUTPUT_VARIABLE LINUX_DISTRO)
 
-    if(${LINUX_DISTRO} MATCHES "NAME=\"(Arch|Manjaro|Artix) Linux\"\n")
+    if(${LINUX_DISTRO} MATCHES "NAME=\"(Arch|Manjaro|Artix) Linux\"|NAME=\"openSUSE Tumbleweed\"\n")
```
paperclover, line 281: can you change this so that it is just NAME="(all the distros)" instead of an | on the outside, will make this simpler

paperclover (changes requested):
thank you for documenting more distros. see comment about simplifying a regular expression

CMakeLists.txt
nellfs: Unlike other distros, openSUSE does not get the "Linux" name after the distro name

paperclover (approved):

PR #9485, opened 2024-03-18
optimization timers promisify

sirenkovladd:
### What does this PR do?

- [X] Code changes

### DIFF
--- a/src/js/node/util.js
+++ b/src/js/node/util.js
@@ -166,24 +166,26 @@ function defineCustomPromisify(target, callback) {
 {
   const { setTimeout: timeout, setImmediate: immediate, setInterval: interval } = globalThis;
 
+  let timersPromises;
+
   if (timeout) {
     defineCustomPromisify(timeout, function setTimeout(arg1) {
-      const fn = defineCustomPromisify(timeout, require("node:timers/promises").setTimeout);
-      return fn.$apply(this, arguments);
+      timersPromises ||= require("node:timers/promises");
+      return timersPromises.setTimeout.$apply(this, arguments);
     });
   }
 
   if (immediate) {
     defineCustomPromisify(immediate, function setImmediate(arg1) {
-      const fn = defineCustomPromisify(immediate, require("node:timers/promises").setImmediate);
-      return fn.$apply(this, arguments);
+      timersPromises ||= require("node:timers/promises");
+      return timersPromises.setImmediate.$apply(this, arguments);
     });
   }
 
   if (interval) {
     defineCustomPromisify(interval, function setInterval(arg1) {
-      const fn = defineCustomPromisify(interval, require("node:timers/promises").setInterval);
-      return fn.$apply(this, arguments);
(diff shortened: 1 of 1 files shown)


src/js/node/util.js
```diff
@@ -166,24 +166,26 @@ function defineCustomPromisify(target, callback) {
...
   }
 
   if (interval) {
     defineCustomPromisify(interval, function setInterval(arg1) {
-      const fn = defineCustomPromisify(interval, require("node:timers/promises").setInterval);
-      return fn.$apply(this, arguments);
+      if (!timersPromises) timersPromises = require("node:timers/promises");
+      return timersPromises.setInterval.$apply(this, arguments);
     });
   }
```
jdalton, line 190: I like this @sirenkovladd! Can you provide a little microbenchmark result. We have gone back and forth on this pattern and a microbenchmark would help solidify a path.

paperclover: the following optimization pattern should not affect performance. unlike in user code, `require` is inlined to the following code:

```
(@getInternalField(@internalModuleRegistry, 21/*node:events*/) || @createInternalModuleById(21/*node:events*/));
```

(this transform is done in [internal-module-registry-scanner.ts](https://github.com/oven-sh/bun/blob/dave/win/2024-03-11/src/codegen/internal-module-registry-scanner.ts#L50-L52), called in [`sliceSourceCode`](https://github.com/oven-sh/bun/blob/dave/win/2024-03-11/src/codegen/builtin-parser.ts#L96-L104))

the first half of this is already a cache lookup, specifically a native array access into the JSInternalFieldObjectImpl in InternalModuleRegistry.h/cpp. this lookup should be incredibly instant and very JIT friendly. reminder that `getInternalField` is a bytecode intrinsic (see [emit_intrinsic_getInternalField](https://github.com/oven-sh/WebKit/blob/c3712c13dcdc091cfe4c7cb8f2c1fd16472e6f92/Source/JavaScriptCore/bytecompiler/NodesCodegen.cpp#L1538-L1548)). then we check if it is not defined and take the || path which will call into a native function `@createInternalModuleById` (see [jsCreateInternalModuleById](https://github.com/oven-sh/bun/blob/a35961cc55bba597c3b6bdb83574dfe1ba35a72b/src/bun.js/bindings/InternalModuleRegistry.cpp#L152-L165)). the only thing that is slow is the latter call, which is already guarded by a conditional expression.

my stance on this pattern is that it is a unnecessary optimization, and that we should actually reverse any places that we do this.

however ... i have not benchmarked the above, and i'm open to being proved wrong.

but then again, i dont think there is any application that the initialization of `util.promisify` on the timer becomes the slow point of an application.

jdalton:
I created a microbenchmark and could see no noticeable win from this (microbenchmarks hate added scope/closures) and saw no benefits from hoisting here. We can remove the pattern. @sirenkovladd If you open to helping on this front a PR to remove this kind of pattern (if it exists elsewhere) would be helpful.

PR #9496, opened 2024-03-18
docs: api/ffi improvements

nektro:
- console.log the result in the js code so that the user gets visual confirmation it worked
- fix the types in the rust example
- add a c++ example


Jarred-Sumner:
thanks

Issue #8831, opened 2024-02-10
crypto.constants.RSA_PSS_SALTLEN_MAX_SIGN returns `undefined`


saklani:
### What version of Bun is running?

1.0.26

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

## Minimum Reproducible
```js
import crypto from "crypto";

console.log(crypto.constants.RSA_X931_PADDING);
console.log(crypto.constants.RSA_PSS_SALTLEN_DIGEST);
console.log(crypto.constants.RSA_PSS_SALTLEN_MAX_SIGN);
console.log(crypto.constants.RSA_PSS_SALTLEN_AUTO);
```

### What is the expected behavior?

Node returns the following result for the above code.
```zsh 
5
-1
-2
-2
```
`Node v21.6.0`

### What do you see instead?

```zsh 
undefined
undefined
undefined
undefined
```

saklani:
What would be the best place to put these missing constants?

PR #9511, opened 2024-03-19
Define missing crypto.constants defined on Node

saklani:
### What does this PR do?

Some crypto constants are missing that are defined by Node in MacOS, Linux, Windows.
#8831
- [x] Code changes

I am on an older version of MacOS (Darwin 21.6.0 x86_64 i386) so I can't build the latest version of bun, but I built an older version `v1.0.30`.

### Simple Replication across Platforms
Seem to be consistent across platform I tried so should be safe define.
Darwin 21.6.0 x86_64 i386 Node v21.6.0
Microsoft Windows NT 10.0.22631.0 x64 Node v20.11.0
Linux 6.1.77-99.164.amzn2023.x86_64 x86_64 x86_64 Node v18.18.2

#### Script
```js
import crypto from "crypto";

console.log(crypto.constants.RSA_X931_PADDING);
console.log(crypto.constants.RSA_PSS_SALTLEN_DIGEST);
console.log(crypto.constants.RSA_PSS_SALTLEN_MAX_SIGN);
console.log(crypto.constants.RSA_PSS_SALTLEN_AUTO);
```
Node Output
```sh
5
-1
-2
-2
```

Bun Output (v1.0.30) built on Darwin 21.6.0 x86_64 i386
```sh
5
-1
-2
-2
```

### Tests
Waiting for maintainer comment on if this requires a test.


src/bun.js/bindings/ProcessBindingConstants.cpp
```diff
@@ -905,20 +905,20 @@ static JSValue processBindingConstantsGetCrypto(VM& vm, JSObject* bindingObject)
 #ifdef RSA_PKCS1_OAEP_PADDING
     object->putDirect(vm, PropertyName(Identifier::fromString(vm, "RSA_PKCS1_OAEP_PADDING"_s)), jsNumber(RSA_PKCS1_OAEP_PADDING));
 #endif
-#ifdef RSA_X931_PADDING
-    object->putDirect(vm, PropertyName(Identifier::fromString(vm, "RSA_X931_PADDING"_s)), jsNumber(RSA_X931_PADDING));
+#ifndef RSA_X931_PADDING
+    object->putDirect(vm, PropertyName(Identifier::fromString(vm, "RSA_X931_PADDING"_s)), jsNumber(5));
 #endif
```
jdalton, line 912: Thank you @saklani! We still want to use the values if they exist so we shouldn't remove the `#ifdef` but we should add an `#else` and put the default value.

paperclover (changes requested):
see john-david's comment, but i think a better approach to this would be to find where these constants are defined. it is possible this is fixed by adding one missing `#include` at the top of this file.

src/bun.js/bindings/ProcessBindingConstants.cpp
saklani: Done!

saklani:
> see john-david's comment, but i think a better approach to this would be to find where these constants are defined. it is possible this is fixed by adding one missing `#include` at the top of this file.

I was hoping for the same. It would be a cleaner approach. It's not defined unfortunately.
Node's dependency of OpenSSL seems to define it. 

https://github.com/nodejs/node/blob/f4a0a3b04b92596fd2cf40e1d9c5a003e81c77b8/deps/openssl/openssl/include/openssl/rsa.h#L189

And from what I understand bun uses boringSSL instead it seems to not be defined there.

paperclover (approved):

jdalton (approved):

PR #9520, opened 2024-03-20
using bun shell for lifecycle scripts on windows

nektro:
wip; need to test on windows


src/install/lifecycle_script_runner.zig
```diff
@@ -131,13 +131,40 @@ pub const LifecycleScriptSubprocess = struct {
 
         const combined_script: [:0]u8 = copy_script.items[0 .. copy_script.items.len - 1 :0];
 
+        var path_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
+        var path_buf2: [bun.MAX_PATH_BYTES]u8 = undefined;
+        const lifecycle_script_path, var lifecycle_script_file = if (Environment.isWindows or shell_bin == null) blk: {
+            const tempdir = this.manager.getTemporaryDirectory();
+            var lifecycle_script_name = bun.span(try bun.fs.FileSystem.instance.tmpname("lifecycle-script", &path_buf, 12345));
```
Jarred-Sumner, line 138: oh we definitely should not do this

test/harness.ts
```diff
@@ -557,3 +557,21 @@ export function toTOMLString(opts: object) {
   }
   return ret;
 }
+
+const shebang_posix = (program: string) => `#!/usr/bin/env ${program}
+`;
+
+const shebang_windows = (program: string) => `0</* :{
```
Jarred-Sumner, line 563: we should also avoid node-gyp.cmd by us nwo that we have the shell working

PR #9530, opened 2024-03-20
Fixes #9521

### LINKED ISSUE #9521: Segmentation Fault when importing an image file with query parameters

1.0.33

Darwin 23.4.0 arm64 arm

Was not able to create a minimal reproduciable repo for this so I will try to note everything relevant here

- Have an react component as tsx file (e.g. `someComponent.tsx`)
- import an image via an JS import and query parameters
  e.g. `import clippy_webp from '../../../assets/img/Clippy.webp?format=webp&h=25&=25'`
- return an jsx element which includes the js import as a src attribute
  e.g. `return Clippy`
- run a `bun test` with a test that imports the file 

the test succeeds every time (no segmentation fault)

### What do you see instead?

when having one import with query parameters, in 1/10 of the cases I get the following error:
`[1]    6830 segmentation fault  bun test`

When adding more imports with query paremeters, the chance of failure increases until at ~10 imports it pretty much always fails. 

### Additional information

The whole query parameter for image imports is used by [this](https://github.com/JonasKruckenberg/imagetools/tree/main/packages/vite)

I tried really hard to reproduce the issue in a minimal repo but failed. What I tried: 
- just having an image import with query parameters
- trying out different query parameters
- trying out different image types
- referencing the import in function returns
- referencing the import in jsx returns

I also took a look at the bun source code and identify the issue but failed to write a meaningful test to verify my assumptions (also not a zig dev so dont know about string specifics). But I assume it is something with this line copying to much/less
[https://github.com/oven-sh/bun/blob/main/src/bun.js/module_loader.zig#L2084](https://github.com/oven-sh/bun/blob/main/src/bun.js/module_loader.zig#L2084)

Jarred-Sumner:
### What does this PR do?

Fixes #9521

Added tests. The `text` and `file` loader ones triggers the crash. The svelte & .js ones do not trigger the crash, but are good to have nonetheless.

I also manually checked Elysia hot reloading doesn't use a lot more memory suddenly


src/bun.js/bindings/exports.zig
```diff
@@ -184,17 +184,27 @@ pub const ResolvedSource = extern struct {
...
     commonjs_exports_len: u32 = 0,
 
     hash: u32 = 0,
 
     allocator: ?*anyopaque = null,
 
     tag: Tag = Tag.javascript,
+
+    /// This is for source_code
     needs_deref: bool = true,
```
dylan-conway:
```suggestion
    source_code_needs_deref: bool = true,
```

src/bun.js/module_loader.zig
```diff
@@ -312,12 +312,18 @@ pub const RuntimeTranspilerStore = struct {
...
             this.poll_ref.unref(vm);
-            var specifier = if (this.parse_error == null) this.resolved_source.specifier else bun.String.createUTF8(this.path.text);
+
             const referrer = bun.String.createUTF8(this.referrer);
             var log = this.log;
             this.log = logger.Log.init(bun.default_allocator);
             var resolved_source = this.resolved_source;
-            resolved_source.source_url = specifier.dupeRef();
+            const specifier = brk: {
+                if (this.parse_error == null) {
```
dylan-conway: is this intentionally the opposite from before? `if (this.parse_error == null) ...` was previously `this.resolved_source.specifier`

dylan-conway (commented):
One suggestion and one question. Otherwise looks good. Confirmed this also fixes the query string crash we are seeing on windows

src/bun.js/module_loader.zig
Jarred-Sumner: ur right

Issue #8965, opened 2024-02-17
Crash in module loader when cycles exist between ESM & CJS


pythonmcpi:
### What version of Bun is running?

1.0.27+c34bbb2e3 on debian, 1.0.27-canary.2+bc7415453 on win10

### What platform is your computer?

Linux 5.10.0-21-amd64 x86_64 unknown | Microsoft Windows NT 10.0.19045.0 x64

### What steps can reproduce the bug?

Create the following files:

```ts
// 1.ts
import SomeClass from "./3";

import { config } from "./5";

const client = {};

console.log(SomeClass);
client.config = config;
```

```ts
// 3.ts
import somedata = require("./4");

export default class SomeClass {}
```

```js
// 4.js
const config = require(`./5`);

module.exports = {};
```

```ts
// 5.ts
const config_ = {};

type GeneratedConfigType = typeof config_;
export interface Config extends GeneratedConfigType {}

export const config = config_ as Config;
```

Then, run `bun run 1.ts`.

### What is the expected behavior?

For bun to not segfault, and exit quietly after logging `SomeClass` to stdout. Alternatively, for bun to complain that `Property 'config' does not exist on type '{}'` for the `client.config = config` line.

### What do you see instead?

Segmentation fault, on both linux and windows. Windows stack trace:
```
Segmentation fault at address 0x18
???:?:?: 0x7ff7e46b7631 in ??? (bun.exe)
???:?:?: 0x7ff7e4646a4c in ??? (bun.exe)
???:?:?: 0x7ff7e46471e1 in ??? (bun.exe)
???:?:?: 0x7ff7e447534d in ??? (bun.exe)
???:?:?: 0x7ff7e442324e in ??? (bun.exe)
???:?:?: 0x7ff7e440228f in ??? (bun.exe)
???:?:?: 0x7ff7e441f86f in ??? (bun.exe)
???:?:?: 0x7ff7e59c3ae5 in ??? (bun.exe)
???:?:?: 0x7ff7e441f77a in ??? (bun.exe)
???:?:?: 0x7ff7e441dae6 in ??? (bun.exe)
???:?:?: 0x1df8000115d in ??? (???)
???:?:?: 0x1dfc4986cff in ??? (???)
???:?:?: 0x1dfc7657467 in ??? (???)
???:?:?: 0x1dfc77c770f in ??? (???)
???:?:?: 0x1dfc76a2c7f in ??? (???)
???:?:?: 0x7002f4670f in ??? (???)
[...]

Issue #6946, opened 2023-11-07
signal SIGSEGV: invalid address


sirenkovladd:
### What version of Bun is running?

1.0.9+98f20170a

### What platform is your computer?

Linux 6.2.0-36-generic x86_64 x86_64 and Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

t.ts
```ts
import { Logger } from './l';
import './t2'
Logger;
```

t2.js
```js
require('./t3')
```

t3.ts
```ts
import { Logger } from './l';
Logger;
```

l.js
```js
class Logger {}
exports.Logger = Logger;
```

`bun t.ts`

The file extensions must be exactly the same
this error is very rare (1/30 times)
but when more files are involved it is almost always

### What is the expected behavior?

[1]    2219418 segmentation fault (core dumped)  bun t.ts

### What do you see instead?

exit code 0

### Additional information

<details><summary>lldb</summary>
<pre>
(lldb) run
Process 1949467 launched: '/home/vamox/.bun/bin/bun-profile' (x86_64)
Process 1949467 stopped
* thread #1, name = 'bun-profile', stop reason = signal SIGSEGV: invalid address (fault address: 0x1a)
    frame #0: 0x00007fffad43c0b3
->  0x7fffad43c0b3: movq   0x10(%rax), %rax
    0x7fffad43c0b7: movq   %rax, -0x28(%r12)
    0x7fffad43c0bc: movq   %rax, -0x40(%rbp)
    0x7fffad43c0c0: movq   0x30(%rbp), %rax
(lldb) bt all
* thread #1, name = 'bun-profile', stop reason = signal SIGSEGV: invalid address (fault address: 0x1a)
  * frame #0: 0x00007fffad43c0b3
    frame #1: 0x0000555558fab597 bun-profile`llint_op_call + 176
    frame #2: 0x0000555558f8def0 bun-profile`vmEntryToJavaScript + 213
    frame #3: 0x0000555559e32679 bun-profile`JSC::Interpreter::executeCall(JSC::JSObject*, JSC::CallData const&, JSC::JSValue, JSC::ArgList const&) + 1961
    frame #4: 0x000055555a12de84 bun-profile`JSC::profiledCall(JSC::JSGlobalObject*, JSC::ProfilingReason, JSC::JSValue, JSC::CallData const&, JSC::JSValue, JSC::ArgList const&) + 212
[...]

PR #9540, opened 2024-03-21
A few fixes related to CommonJS module loading

Jarred-Sumner:
### What does this PR do?

- Add extra checks to ensure source code is always freed when it should be
- Optimization: precompute the hash upfront for source code transpiled on another thread. This seems
- Make `Bun.gc(true)` free more memory
- Make `CommonJSModule` store the JSC::SourceCode directly instead of using `JSC::JSSourceCode`. Goal is to free it earlier instead of whenever GC decides to free it.
- Fixes https://github.com/oven-sh/bun/issues/6946
- Fixes https://github.com/oven-sh/bun/issues/8965
- Fixes a bug where bun loads & transpiles the same code twice(!)

It doesn't work yet 

It crashes in release builds, but not debug builds

I suspect something fishy with either GC or a C++ use-after-move somewhere. Worst case I revert it back

### DIFF
--- a/src/bun.js/module_loader.zig
+++ b/src/bun.js/module_loader.zig
@@ -86,6 +86,8 @@ const Dependency = @import("../install/dependency.zig");
 const Async = bun.Async;
 const String = bun.String;
 
+const debug = Output.scoped(.ModuleLoader, true);
+
 // Setting BUN_OVERRIDE_MODULE_PATH to the path to the bun repo will make it so modules are loaded
 // from there instead of the ones embedded into the binary.
 // In debug mode, this is set automatically for you, using the path relative to this file.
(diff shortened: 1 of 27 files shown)


src/bun.js/module_loader.zig
```diff
@@ -637,20 +665,31 @@ pub const RuntimeTranspilerStore = struct {
...
-                    break :brk result;
-                },
+                // In a benchmarking loading @babel/standalone 100 times:
+                //
+                // Before ensureHash:
+                // 354.00 ms    4.2%	354.00 ms	 	  WTF::StringImpl::hashSlowCase() const
+                //
+                // After ensureHash:
+                // 506.00 ms    6.1%	506.00 ms	 	  WTF::StringImpl::hashSlowCase() const
+                //
```
paperclover, line 685: is this in the right order

staabm:
> * Optimization: precompute the hash upfront for source code transpiled on another thread. This seems

nit: it seems the PR description is truncated at this point

paperclover (approved):
looks good now.

if you are worried about JS code being wrong, feel free to add more `$assert`s into it (they are tree-shaken)

src/bun.js/module_loader.zig
Jarred-Sumner: image

PR #9550, opened 2024-03-22
fix(windows): fix a few more tests

dylan-conway:
### What does this PR do?
fixes:
- child_process-node.test.js
- napi.test.ts
- hot.test.ts


src/install/lifecycle_script_runner.zig
```diff
@@ -145,15 +151,15 @@ pub const LifecycleScriptSubprocess = struct {
         }
         const spawn_options = bun.spawn.SpawnOptions{
             .stdin = .ignore,
-            .stdout = if (this.manager.options.log_level.isVerbose())
+            .stdout = if (this.manager.options.log_level.isVerbose() or (this.log_level != .silent and this.foreground))
```
Jarred-Sumner: log_level == .silent should be `.ignore`

dylan-conway: updated

src/deps/c_ares.zig
```diff
@@ -473,8 +473,8 @@ pub const Channel = opaque {
         opts.flags = ARES_FLAG_NOCHECKRESP;
         opts.sock_state_cb = &SockStateWrap.onSockState;
         opts.sock_state_cb_data = @as(*anyopaque, @ptrCast(this));
-        opts.timeout = 1000;
-        opts.tries = 3;
+        opts.timeout = -1;
```
dylan-conway, line 476: Defaults taken from node. -1 means c-ares will choose the default (5000ms)

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -1962,6 +1962,9 @@ pub const DNSResolver = struct {
 
     pub fn onDNSPollUv(watcher: [*c]bun.windows.libuv.uv_poll_t, status: c_int, events: c_int) callconv(.C) void {
         const poll = UvDnsPoll.fromPoll(watcher);
+        const vm = poll.parent.vm;
+        vm.eventLoop().enter();
```
dylan-conway, line 1966: Without this enter/exit callbacks like this wouldn't be called. I thought this would fix the dns tests, but it did not.
```js
const dns = require("dns");
dns.resolveNs("", (err, results) => {
  console.log(results);
});
```

test/js/node/child_process/child_process-node.test.js
```diff
@@ -397,38 +398,44 @@ describe("child_process double pipe", () => {
     // const { mustCallAtLeast, mustCall } = createCallCheckCtx(done);
     const mustCallAtLeast = fn => fn;
     const mustCall = fn => fn;
-    let grep, sed, echo;
-    grep = spawn("grep", ["o"], { stdio: ["pipe", "pipe", "pipe"] });
-    sed = spawn("sed", ["s/o/O/"]);
-    echo = spawn("echo", ["hello\nnode\nand\nworld\n"]);
+    let fakeGrep, fakeSed, fakeEcho;
+    fakeGrep = spawn(bunExe(), [
+      "-e",
```
dylan-conway, line 403: Fake grep, sed, and echo with bun because the windows versions work differently.

test/napi/napi.test.ts
```diff
@@ -18,20 +17,7 @@ describe("napi", () => {
     if (!install.success) {
       throw new Error("build failed");
     }
-
-    const build = spawnSync({
```
dylan-conway, line 22: This build step is run during install as a lifecycle script so we can remove it

PR #9557, opened 2024-03-22
Allow 0-length ArrayBuffer & Blob in `Bun.spawn` stdio

Jarred-Sumner:
### What does this PR do?

This makes it so if you pass an empty arraybuffer, typed array, or blob to Bun.spawn it will become /dev/null instead of throwing an error. It is hard to think of a good reason for that to have been an error, and it's easier if it's not

Before:

image

After:

image

Tests


gvilums (approved):

src/bun.js/api/bun/spawn/stdio.zig
```diff
@@ -475,6 +476,12 @@ pub const Stdio = union(enum) {
             return false;
         }
 
+        // Instead of writing an empty blob, lets just make it /dev/null
```
jdalton, line 479: I read this as "lets just take it to the null`. (like take it to the net) 😜

jdalton (approved):
More friendly API FTW.

PR #9571, opened 2024-03-22
Fix bundling for bun not using ascii_only

### LINKED ISSUE #9559: When using bun to build a binary file, the Chinese key in the js object will be garbled.

1.0.33

Linux 5.15.0-86-generic x86_64 x86_64

{C792D954-4A15-4070-A390-D47F11D3FFE1}

build
`bun build ./a.ts --compile --outfile test`
execute
`./test`

Snipaste_2024-03-22_15-29-20

### What do you see instead?

Snipaste_2024-03-22_15-28-37

### Additional information

_No response_

pfgithub:
### What does this PR do?

Fixes #9559

In printWithWriterAndPlatform(), its NewPrinter call was always setting ascii_only to false, but that means when bundling for the bun target, the code is not emitted as ascii only and there will be errors when running because it tries to run as latin1.

This code

```js
console.log(JSON.stringify({我: "a"}));

const 我 = "b";
console.log(JSON.stringify({我}));
```

When built with `bun build --target bun code.js` now emits:

```js
// @bun
// code.js
console.log(JSON.stringify({ "\u{6211}": "a" }));
var \u{6211} = "b";
console.log(JSON.stringify({ "\u{6211}": \u{6211} }));
```

instead of:

```js
// @bun
// code.js
console.log(JSON.stringify({ "我": "a" }));
var 我 = "b";
console.log(JSON.stringify({ "我": 我 }));
```

It also works with `bun build --compile`.

- [x] Code changes

Added a regression test for the issue

It tests `bun build --target bun` and not `bun build --compile` because compile takes a second and they're both the same issue.

If Zig files changed:

- [ ] N/A I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be

### DIFF
--- a/src/js_printer.zig
+++ b/src/js_printer.zig
@@ -6036,7 +6036,8 @@ pub fn printWithWriterAndPlatform(
     comptime generate_source_maps: bool,
 ) PrintResult {
     const PrinterType = NewPrinter(
-        false,
+        // if it's bun, it is also ascii_only
+        is_bun_platform,
         Writer,
         false,
         is_bun_platform,
--- a/test/bundler/bundler_compile.test.ts
+++ b/test/bundler/bundler_compile.test.ts
@@ -220,4 +220,13 @@ describe("bundler", () => {
     },
     run: { stdout: "Hello, world!", setCwd: true },
   });
+  itBundled("compile/Utf8", {
+    compile: true,
+    files: {
+      "/entry.ts": /* js */ `
+        console.log(JSON.stringify({\u{6211}: "\u{6211}"}));
+      `,
+    },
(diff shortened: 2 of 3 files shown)


paperclover (approved):
today i learned `\u{6211}` is valid identifier

Jarred-Sumner:
Thank you for this.

Can you add a test for `bun build --compile` as well? You can see the tests in `bundler_compile` for an example

pfgithub:
Added one in bundler_compile

pfgithub:
I don't think any of the CI failures are my fault

Jarred-Sumner:
Agreed, thank you for this

PR #9578, opened 2024-03-23
use bun shell for lifecycle scripts on windows [v2]

nektro:
wip


src/install/install.zig
```diff
@@ -7017,7 +7017,7 @@ pub const PackageManager = struct {
                     switch (bun.sys.sys_uv.symlinkUV(
                         link_path,
                         dest_path,
-                        bun.windows.libuv.UV_FS_SYMLINK_JUNCTION,
+                        bun.windows.libuv.UV_FS_SYMLINK_DIR,
```
Jarred-Sumner: interesting

src/install/lifecycle_script_runner.zig
```diff
@@ -132,13 +132,40 @@ pub const LifecycleScriptSubprocess = struct {
 
         const combined_script: [:0]u8 = copy_script.items[0 .. copy_script.items.len - 1 :0];
 
+        var path_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
+        var path_buf2: [bun.MAX_PATH_BYTES]u8 = undefined;
+        const lifecycle_script_path, var lifecycle_script_file = if (Environment.isWindows or shell_bin == null) blk: {
+            const tempdir = this.manager.getTemporaryDirectory();
+            var lifecycle_script_name = bun.span(try bun.fs.FileSystem.instance.tmpname("lifecycle-script", &path_buf, 12345));
```
Jarred-Sumner, line 141: why are we saving files to a temporary directory? there are many alternatives which would be more reliable

nektro: it creates a temporary file (a `.bun.sh`) in the global temp directory and then runs it with the bun shell in a child process

Jarred-Sumner: why not pass the script contents as argv? or pass it as stdin? or pass it as an environment variable? any of these would be more reliable + low overhead compared to writing to a temporary file

.github/workflows/bun-mac-x64.yml
```diff
@@ -1,463 +0,0 @@
-name: bun-macOS-x64
```
paperclover: reminder to un-delete before merge

src/bun.zig
```diff
@@ -2816,6 +2816,21 @@ pub fn linuxKernelVersion() Semver.Version {
     return @import("./analytics.zig").GenerateHeader.GeneratePlatform.kernelVersion();
 }
 
+pub fn selfExePath() ![:0]u8 {
+    const memo = struct {
+        var set = false;
+        // this is lame; open issues to fix std soon
+        var value: [4096]u8 = undefined;
```
paperclover, line 2823: max path bytes

src/cli/run_command.zig
```diff
@@ -790,8 +785,16 @@ pub const RunCommand = struct {
...
             else
                 "";
+            @memcpy(target_path_buffer[prefix.len..][len..].ptr, comptime bun.strings.w(dir_name));
             const dir_slice = target_path_buffer[0 .. prefix.len + len + dir_name.len];
 
+            if (Environment.isDebug) {
+                const dir_slice_u8 = std.unicode.utf16leToUtf8Alloc(bun.default_allocator, dir_slice) catch @panic("oom");
+                defer bun.default_allocator.free(dir_slice_u8);
+                std.fs.deleteTreeAbsolute(dir_slice_u8) catch {};
+                std.fs.makeDirAbsolute(dir_slice_u8) catch @panic("huh?");
```
paperclover, line 795: maybe `try` so we get error return trace instead of an opaque stack trace without the error.

src/install/lifecycle_script_runner.zig
```diff
@@ -123,7 +123,7 @@ pub const LifecycleScriptSubprocess = struct {
         this.current_script_index = next_script_index;
         this.has_called_process_exit = false;
 
-        const shell_bin = bun.CLI.RunCommand.findShell(env.get("PATH") orelse "", cwd) orelse return error.MissingShell;
+        const shell_bin = bun.CLI.RunCommand.findShell(env.get("PATH") orelse "", cwd) orelse null;
```
paperclover: in windows, `shell_bin` gets ignored now as bun shell is used.

while computing this is fast (it is a single return with the path to cmd) i dont really like this.

src/install/lifecycle_script_runner.zig
paperclover: im unaware of any way to use bun's shell interpreter via standard in or via argv. `--eval` and `--print` apply to typescript only. so a mechanism would have to be added for this (`--eval-shell`? i dont want to add more flags than needed)

src/cli/run_command.zig
```diff
@@ -817,6 +820,9 @@ pub const RunCommand = struct {
                     }
                 }
             }
+            if (PATH.items.len > 0 and PATH.items[PATH.items.len - 1] != std.fs.path.delimiter) {
```
Jarred-Sumner, line 823: is this supposed to be here

src/which.zig
```diff
@@ -77,7 +82,18 @@ pub fn endsWithExtension(str: []const u8) bool {
 
 /// Check if the WPathBuffer holds a existing file path, checking also for windows extensions variants like .exe, .cmd and .bat (internally used by whichWin)
 fn searchBin(buf: *bun.WPathBuffer, path_size: usize, check_windows_extensions: bool) ?[:0]const u16 {
```
Jarred-Sumner, line 67: this is only used on windows

src/cli/run_command.zig
nektro: ? `@panic` would still produce a stack trace at runtime if this was ever hit

src/cli/run_command.zig
nektro: yes

src/which.zig
```diff
@@ -77,7 +82,18 @@ pub fn endsWithExtension(str: []const u8) bool {
...
-    if (!check_windows_extensions)
+    if (bun.Environment.isWindows) {
+        const haystack = buf[0..path_size :0];
+        for (win_extensionsW) |ext| {
+            if (path_size < 4) continue;
```
Jarred-Sumner: `< 4`?

Jarred-Sumner: if all file extensions in win_extensionsW are at least 4 characters and the path_size < 4 is supposed to check that

then this should use a named constant so that we know what it means

or hasSuffix should be used which would check this

src/cli/run_command.zig
Jarred-Sumner: no, it should not print "huh?"

imagine you're a user

you run `bun install`, it crashes

just says "huh?"

you're frustrated because it crashed, and now you're confused because all it said was "huh?"

Jarred-Sumner: oh this is in Environment.isDebug

its probably fine

src/string_immutable.zig
```diff
@@ -698,6 +698,10 @@ pub inline fn endsWith(self: string, str: string) bool {
     return str.len == 0 or @call(bun.callmod_inline, std.mem.endsWith, .{ u8, self, str });
 }
 
+pub inline fn endsWithGeneric(comptime T: type, self: []const T, str: []const T) bool {
+    return str.len == 0 or @call(bun.callmod_inline, std.mem.endsWith, .{ T, self, str });
```
Jarred-Sumner: eqlComptimeCheckLenWithType
```suggestion
    return str.len == 0 or @call(bun.callmod_inline, eqlComptimeCheckLenWithType, .{ T, self, str, true });
```

src/which.zig
```diff
@@ -15,28 +15,33 @@ fn isValid(buf: *bun.PathBuffer, segment: []const u8, bin: []const u8) ?u16 {
...
-    }
-    if (bin.len == 0) return null;
 
     // handle absolute paths
     if (std.fs.path.isAbsolute(bin)) {
         bun.copy(u8, buf, bin);
         buf[bin.len] = 0;
         const binZ: [:0]u8 = buf[0..bin.len :0];
+        if (bun.Environment.isWindows) {
+            (std.fs.cwd().openFile(bin, .{}) catch return null).close();
```
Jarred-Sumner: this will cause a performance issue and a segfault due to their invalid path format error handling code

instead if we need to use a system call here, this should use `GetFileType`

Jarred-Sumner: actually, `GetFileAttributesW`

src/bun.zig
nektro: this is only for the path to the bun exe itself so I dont forsee this being an issue and the error is handled gracefully in the event this occurs

PR #9598, opened 2024-03-24
Do not skip tests in CI

Jarred-Sumner:
### What does this PR do?

We should not be skipping tests in CI.

CI

### DIFF
--- a/test/js/third_party/mongodb/mongodb.test.ts
+++ b/test/js/third_party/mongodb/mongodb.test.ts
@@ -1,9 +1,10 @@
 import { test, expect, describe } from "bun:test";
+import { isBunCI } from "harness";
 import { MongoClient } from "mongodb";
 
 const CONNECTION_STRING = process.env.TLS_MONGODB_DATABASE_URL;
 
-const it = CONNECTION_STRING ? test : test.skip;
+const it = CONNECTION_STRING ? test : test.skipIf(!isBunCI);
 
 describe("mongodb", () => {
   it("should connect and inpect", async () => {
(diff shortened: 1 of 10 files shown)


test/js/third_party/postgres/postgres.test.ts
```diff
@@ -1,11 +1,12 @@
 import { test, expect, describe } from "bun:test";
+import { isBunCI } from "harness";
 import { Pool, Client } from "pg";
 import { parse } from "pg-connection-string";
 import postgres from "postgres";
 
 const CONNECTION_STRING = process.env.TLS_POSTGRES_DATABASE_URL;
 
-const it = CONNECTION_STRING ? test : test.skip;
+const it = CONNECTION_STRING ? test : test.skipIf(!isBunCI);
```
paperclover, line 9: i agree that ci should run this, but from outside contributors this test will fail as the environment doesnt load. has this been fixed?

test/js/third_party/mongodb/mongodb.test.ts
```diff
@@ -1,9 +1,10 @@
 import { test, expect, describe } from "bun:test";
+import { isBunCI } from "harness";
 import { MongoClient } from "mongodb";
 
 const CONNECTION_STRING = process.env.TLS_MONGODB_DATABASE_URL;
 
-const it = CONNECTION_STRING ? test : test.skip;
+const it = CONNECTION_STRING ? test : test.skipIf(!isBunCI);
```
jdalton, line 8: Neat way to handle `it` aliases as `test`.

PR #9606, opened 2024-03-25
Clarified version pinning in Git imports.

PhoneDroid:
### What does this PR do?

I moved the tag / release pinning example of the Git import 
documentation into a separate block and added another 
example showing the ability to specify a commit Id.

- [ ] Code changes


PhoneDroid:
Waiting for months just to have it rejected without reason ..

PR #9622, opened 2024-03-25
windows: fix argument handling in bin shim

paperclover:
### What does this PR do?

there are some remaining failures in bunx test. it is incredibly flaky there so i need more time to look at


src/cli/run_command.zig
```diff
@@ -1507,6 +1508,7 @@ pub const RunCommand = struct {
 
         // Run absolute/relative path
         if ((script_name_to_search.len > 1 and script_name_to_search[0] == '/') or
```
Jarred-Sumner: don't we have a way to know if a path is absolute?
```suggestion
        if (bun.path.isAbsolute(script_name_to_search))
```

paperclover: i can make this change, but there is a difference between this code and isAbsolute, in that it is checking for a path that is not just the root dir.

this would make `bun run /` never run a script named `/`

why would you name a script this? no idea, but it would no longer be usable. through bun.

test/cli/install/bunx.test.ts
```diff
@@ -95,10 +103,9 @@ it("should output usage if no arguments are passed", async () => {
 });
 
 it("should work for @scoped packages", async () => {
```
dylan-conway, line 105: we should run this test with and without `--bun`, but it can be added later

dylan-conway (approved):
LGTM

PR #9627, opened 2024-03-25
Don't pass kill(0) in console iterator test

Jarred-Sumner:
### What does this PR do?

Don't pass kill(0) in console iterator test

Ran the test


test/js/bun/console/console-iterator.test.ts
```diff
@@ -1,8 +1,10 @@
 import { spawnSync, spawn } from "bun";
 import { describe, expect, it } from "bun:test";
 import { bunEnv, bunExe } from "harness";
+import path from "path";
+const dir = (out: string) => path.join(import.meta.dir, out);
 
-describe("should work for static input", () => {
+it("should work for static input", async () => {
```
jdalton, line 7: describe vs it nit. IMO test should be described with what the batch of "it"'s are doing

test/js/bun/console/console-iterator.test.ts
```diff
@@ -12,20 +14,19 @@ describe("should work for static input", () => {
...
-        env: bunEnv,
-      });
-      expect(stdout.toString()).toBe(input.replaceAll("\n", ""));
+  async function run(input: string) {
+    const { stdout, exited } = spawn({
+      cmd: [bunExe(), dir("console-iterator-run.ts")],
+      stdin: Buffer.from(input),
+      env: bunEnv,
     });
+    expect(await new Response(stdout).text()).toBe(input.replaceAll("\n", ""));
```
jdalton, line 23: In the future this could get wonky on windows if the test file starts to contain `\r\n`.

test/js/bun/console/console-iterator.test.ts
```diff
@@ -70,5 +70,5 @@ it("can use the console iterator more than once", async () => {
   await stdin.end();
 
   expect(await new Response(stdout).text()).toBe('["hello","world"]["another"]');
-  proc.kill(0);
+  proc.kill();
```
jdalton, line 73: What issue was the pid causing?

PR #9628, opened 2024-03-25
Implement `fs.openAsBlob`

Jarred-Sumner:
### What does this PR do?

- Implement `fs.openAsBlob`
- Don't use `fs.readvSync` in `fs.readv` now that we have an actual implementation of `fs.readv`
- Don't use `fs.writevSync` in `fs.writev` now that we have an actual implementation of `fs.writev`
- Add `fs.opendir` stub

There are some tests


paperclover (approved):
good

PR #9631, opened 2024-03-26
feat: Shell `if-else`, conditional expressions, running commands in background

zackradisic:
### What does this PR do?

This adds 3 new features to the shell:
- [if-else](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_09_04_07)  (e.g. `if cond; then stuff; else other stuff; fi`)
- [conditional expressions](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html) (e.g. `[[ -f package.json ]] && echo package.json exists!`)
- Running [commands in the background](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_09_03_02) (e.g. `longrunningcmd &`)

Note that technically commands run in the background should be a separate process with their own PID, but this is not implemented yet

Also, not all conditional expression operators are supported ([here](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html) is a list of all them in bash). For an initial implementation, the following are suppported:
- `-d` is directory
- `-f` file exists
- `-c` is character device
- `-z` string is empty
- `-n` string is non-empty


src/shell/shell.zig
```diff
@@ -342,24 +343,293 @@ pub const AST = struct {
...
+        op: Op,
+        args: ArgList = ArgList.zeroes,
+
+        const ArgList = SmolList(Atom, 2);
+
+        // args: SmolList(1, comptime INLINED_MAX: comptime_int)
+        pub const Op = enum {
+            /// -a file
+            ///   True if file exists.
+            @"-a",
```
dylan-conway, line 402: is `-a` normally a flag? it's already covered by `-e`

zackradisic: Yep, at first I thought it was gpt hallucinating when it generated this struct, but for some reason -a and -e do the same thing according to the gnu bash manual: https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions

dylan-conway: oh interesting

src/bun.js/event_loop.zig
```diff
@@ -883,6 +887,11 @@ pub const EventLoop = struct {
         while (@field(this, queue_name).readItem()) |task| {
             defer counter += 1;
             switch (task.tag()) {
+                @field(Task.Tag, typeBaseName(@typeName(ShellAsync))) => {
+                    var shell_ls_task: *ShellAsync = task.get(ShellAsync).?;
+                    shell_ls_task.runFromMainThread();
+                    // shell_ls_task.deinit();
```
Jarred-Sumner: Please remove this comment

src/bun.js/event_loop.zig
```diff
@@ -893,6 +902,11 @@ pub const EventLoop = struct {
                     shell_ls_task.runFromMainThread();
                     // shell_ls_task.deinit();
                 },
+                @field(Task.Tag, typeBaseName(@typeName(ShellCondExprStatTask))) => {
+                    var shell_ls_task: *ShellCondExprStatTask = task.get(ShellCondExprStatTask).?;
+                    shell_ls_task.task.runFromMainThread();
+                    // shell_ls_task.deinit();
```
Jarred-Sumner: ✂️

src/shell/interpreter.zig
```diff
@@ -2352,6 +2374,8 @@ pub const Interpreter = struct {
                 this.parent.childDone(this, exit_code);
                 return;
             }
+
+            @panic("Looks like you've got a bad case of undefined memory kid");
```
Jarred-Sumner: This is not a helpful error message

src/shell/interpreter.zig
```diff
@@ -3054,6 +3139,591 @@ pub const Interpreter = struct {
...
+
+        pub fn childDone(this: *Async, child_ptr: ChildPtr, exit_code: ExitCode) void {
+            log("{} childDone", .{this});
+            child_ptr.deinit();
+            this.state = .{ .done = exit_code };
+            this.enqueueSelf();
+        }
+
+        pub fn deinit(this: *Async) void {
+            _ = this; // autofix
```
Jarred-Sumner: Doesn’t this leak memory? Don’t you need to call .destroy?

src/shell/interpreter.zig
```diff
@@ -3054,6 +3139,591 @@ pub const Interpreter = struct {
...
+        }
+
+        pub fn actuallyDeinit(this: *Async) void {
+            this.io.deref();
+            bun.destroy(this);
```
Jarred-Sumner, line 3363: This seems very sketchy

Jarred-Sumner: Can you leave a detailed comment explaining why this makes sense?

src/shell/interpreter.zig
```diff
@@ -3054,6 +3139,591 @@ pub const Interpreter = struct {
...
+                            .@"-c" => {
+                                const st: bun.Stat = switch (this.state.stat_complete.stat) {
+                                    .result => |st| st,
+                                    .err => {
+                                        // It seems that bash always gives exit code 1
+                                        this.parent.childDone(this, 1);
+                                        return;
+                                    },
+                                };
+                                this.parent.childDone(this, if (bun.S.ISCHR(@intCast(st.mode))) 0 else 1);
```
Jarred-Sumner, line 3515: Why are we doing special checks for ISCHR? A file, pipe, a socket can all be a TTY and not a character device. You most likely want isatty instead.

src/shell/interpreter.zig
```diff
@@ -3054,6 +3139,591 @@ pub const Interpreter = struct {
...
+
+                        const idx = this.state.exec.stmt_idx;
+                        this.state.exec.stmt_idx += 1;
+                        const stmt = this.state.exec.stmts.getConst(idx);
+                        var newstmt = Stmt.init(this.base.interpreter, this.base.shell, stmt, this, this.io.copy());
+                        newstmt.start();
+                        return;
+                    },
+                    .waiting_write_err => return, // yield execution
+                    .done => std.debug.assert(false),
```
```suggestion
                    .done => @panic(“This code should not be reachable”) 
```

src/shell/shell.zig
```diff
@@ -1084,6 +1666,18 @@ pub const Parser = struct {
         unreachable;
     }
 
+    fn match_if_clausetok(self: *Parser, toktag: IfClauseTok) bool {
+        if (self.peek() == .Text and
+            self.delimits(self.peek_n(1)) and
+            std.mem.eql(u8, self.text(self.peek().Text), @tagName(toktag)))
```
Jarred-Sumner: strings.eql
```suggestion
            strings.eql(self.text(self.peek().Text), @tagName(toktag)))
```

src/shell/shell.zig
```diff
@@ -2743,6 +3479,13 @@ pub const Test = struct {
                 .CmdSubstEnd => return .CmdSubstEnd,
                 .OpenParen => return .OpenParen,
                 .CloseParen => return .CloseParen,
+                .DoubleBracketOpen => return .DoubleBracketOpen,
+                .DoubleBracketClose => return .DoubleBracketClose,
+                // .If => return .If,
```
Jarred-Sumner: What are these comments for? Should you delete this?

Jarred-Sumner (changes requested):
Some comments

Definitely not enough tests. We cannot ship this until there are a lot more tests.

Jarred-Sumner:
I’m also confused where the code for subprocesses or job control is. It’d probably be reasonable to ship if else separately from &

Jarred-Sumner:
@zackradisic there are merge conflicts

src/shell/interpreter.zig
zackradisic: This is for -c conditional expression which checks that the file is a character device

Jarred-Sumner:
Almost ready, but there are test failures that need to be addressed before this PR can be merged.

Tests are failing on Windows:

- test\cli\install\bun-run-bunfig.test.ts 12 failing
- test\cli\install\bun-run.test.ts 4 failing
- test\cli\run\env.test.ts 10 failing
- test\cli\run\if-present.test.ts 1 failing
- test\cli\run\run-process-env.test.ts 2 failing
- test\cli\run\run-quote.test.ts 1 failing
- test\js\bun\shell\bunshell.test.ts 1 failing 

Tests are failing on macOS/Linux:

- [test/cli/install/bun-run-bunfig.test.ts](https://github.com/oven-sh/bun/pull/9631#testcliinstallbun-run-bunfigtestts) 4 failing
- [test/cli/run/env.test.ts](https://github.com/oven-sh/bun/pull/9631#testclirunenvtestts) 7 failing

mrousavy:
Hey - should background commands work already after this PR was merged? When I run a command like this:

```js
await $`"$ANDROID_HOME/emulator/emulator" -avd "${emulatorName}" -netdelay none -netspeed full &`.quiet()
```

I get this bun error:

```diff
- error: Background commands "&" are not supported yet.
-       at BunShell (191:35)
```

src/shell/shell.zig
```diff
@@ -713,6 +992,40 @@ pub const Parser = struct {
...
+                // // background.
+                // //
+                // // So we do a quick check for this kind of syntax here, and
+                // // provide a helpful error message to the user.
+                // if (self.peek() == .DoubleAmpersand) {
+                //     try self.add_error("\"&\" is not allowed on the left-hand side of \"&&\"", .{});
+                //     return ParseError.Unsupported;
+                // }
+
+                // break;
```
mrousavy, line 1027: Ah, I guess the error is expected then. Hm.

PR #9448, opened 2024-03-15
shell: make sure proper .env is loaded in script runner

nektro:
Closes https://github.com/oven-sh/bun/issues/9445

PR #9642, opened 2024-03-26
tidy: clean up code for env loader

### LINKED ISSUE #9445: environment variable loading is broken with 1.0.31

1.0.31

broken for me on both linux and macOS

created a repo for minimum failure case
https://github.com/gopricy/bun-env-bug

run `bun prod` prints "production" as expected in 1.0.30
and prints "development" in 1.0.31

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #9635: NODE_ENV=production in package.json scripts no longer reads from .env.production

1.0.35+940448d6b

Linux 6.5.0-25-generic x86_64 unknown

Create a new bun project.
`bun init`

Change package.json to:
```
{
  "name": "bun-env-issue",
  "module": "index.js",
  "type": "module",
  "scripts": {
    "dev": "NODE_ENV=development bun index.js",
    "prod": "NODE_ENV=production bun index.js"
  },
  "devDependencies": {
    "@types/bun": "latest"
  }
}
```

Chang index.js to
```
console.log("NODE_ENV:", import.meta.env.NODE_ENV)
console.log("ENV_FILE_NAME:", import.meta.env.ENV_FILE_NAME)
```

create .env files 
.env
```
ENV_FILE_NAME=.env
```

.env.development
```
ENV_FILE_NAME=.env.development
```

.env.production
```
ENV_FILE_NAME=.env.production
```

Run 
`bun run prod`

```
$ NODE_ENV=production bun index.js
NODE_ENV: production
ENV_FILE_NAME: .env.production
```

### What do you see instead?

```
$ NODE_ENV=production bun index.js
NODE_ENV: production
ENV_FILE_NAME: .env.development
```

### Additional information

Earlier versions of bun would choose .env.production if NODE_ENV=production was specified inside of scripts in package.json. 
To get the previous behavior in v1.0.35 you need to specify NODE_ENV=production outside of package.json.

This is problematic since a common pattern is to put the different modes inside of package.json scripts. And it breaks previous behaviour.

Recreated the problem including the previous behavior in the following repo where 2 docker images exists one working (bun version 1.0.33, and one broken 1.0.35): [https://github.com/linusback/bun-env-issue](https://github.com/linusback/bun-env-issue)

paperclover:
### What does this PR do?

Fixes #9635
Fixes #9445
Supercedes #9448

When we moved from `std.process` to our own spawn implementation, we moved from `stdEnvMap` to `createNullDelimitedEnvMap` in many places. i would like to highlight one line of the former:

```diff
        while (iter.next()) |entry| {
+           // Allow var from .env.development or .env.production to be loaded again
+           if (!entry.value_ptr.conditional) {
                try env_map.hash_map.put(entry.key_ptr.*, entry.value_ptr.value);
            }
        }
```

Such check is not present in `createNullDelimitedEnvMap`. We started passing a bunch of `.conditional = true` variables to subprocesses.

This PR removes .conditional and changes `HashTableValue` into just `string`. then, instead of loading variables that never get passed to subprocess, let's just not load them. the code reads much nicer.

After adding this check, it works, however this `.conditional = true` flag does not make any sense to me. I detail some more about this issue in [this comment](https://github.com/oven-sh/bun/issues/9635#issuecomment-2021350123) which is worth reading to get an understanding of the broken behavior this fixes.

As for `createNullDelimitedEnvMap`, when i made the change, i also turned the entire operation into a single allocation, instead of many smaller ones. This will make the memory easier to free (it is still a memory leak... maybe i fix)

---

I verify this code works by having many new tests, all of which i verify work on Bun 1.0.33, a commit before the regression.


src/env_loader.zig
```diff
@@ -1011,8 +999,9 @@ const Parser = struct {
         map: *Map,
         comptime override: bool,
         comptime is_process: bool,
-        comptime conditional: bool,
     ) void {
+        Analytics.Features.dotenv = true;
```
dylan-conway, line 1003: this should only be true if it's replacing the dotenv package

src/env_loader.zig
```diff
@@ -1057,61 +1040,92 @@ const Parser = struct {
...
+            var it = env_map.iterator();
+            while (it.next()) |pair| {
+                envp_count += 1;
+                // env line is 'KEY=VALUE\x00'
+                total_bytes += (pair.key_ptr.len + pair.value_ptr.len + "=\x00".len);
+            }
+        }
+        total_bytes += (envp_count + 1) * @sizeOf(?[*:0]u8); // +1 for the null ptr after the pointer list
+
+        const buf = try alloc.alignedAlloc(u8, @alignOf(?[*:0]u8), total_bytes);
```
dylan-conway: why not allocSentinel?

src/env_loader.zig
```diff
@@ -1232,21 +1234,15 @@ pub const Map = struct {
...
-        });
+        _ = try this.map.getOrPutValue(key, value);
     }
 
     pub inline fn getOrPut(this: *Map, key: string, value: string) !void {
-        _ = try this.map.getOrPutValue(key, .{
-            .value = value,
-            .conditional = false,
-        });
+        _ = try this.map.getOrPutValue(key, value);
```
dylan-conway: we should probably remove `getOrPut` or `putDefault` because they're identical

src/cli/run_command.zig
```diff
@@ -876,26 +876,20 @@ pub const RunCommand = struct {
         if (env == null) {
             this_bundler.env.loadProcess();
 
-            if (this_bundler.env.get("NODE_ENV")) |node_env| {
-                if (strings.eqlComptime(node_env, "production")) {
-                    this_bundler.options.production = true;
-                }
-            }
+            if (this_bundler.env.isProduction())
```
dylan-conway, line 881: so we weren't setting production with `BUN_ENV` before? I think this change should be applied to line 1052 as well

Jarred-Sumner:
Nice

src/env_loader.zig
paperclover: this flag was set at every point (like 10 of them)
```
   if (dir.hasComptimeQuery(".env.development")) {
      try this.loadEnvFile(dir_handle, ".env.development", false, true);
-     Analytics.Features.dotenv = true;
   }
```
I move this to the common root that all usages call into.

this was placed at the start of the function so that a crash in dotenv parser would be marked as using dotenv

src/env_loader.zig
paperclover: `buf` is the backing bytes for all environment values as well as the pointers. it is then used as the backing buffer for each string, which is sentinel allocated, and then the list of string pointers, which has it's own sentinel.

a zero on this `buf` is therefore not needed. this is asserted by every [ ... :0] slice in this function.

src/env_loader.zig
dylan-conway: I see, makes sense

src/env_loader.zig
dylan-conway: 👍

src/install/lifecycle_script_runner.zig
```diff
@@ -391,18 +391,19 @@ pub const LifecycleScriptSubprocess = struct {
             this.stderr.deinit();
         }
 
+        this.env.deinit(this.manager.allocator);
```
dylan-conway, line 394: nice

src/shell/subproc.zig
```diff
@@ -845,14 +863,14 @@ pub const ShellSubprocess = struct {
             return .{ .err = .{ .custom = bun.default_allocator.dupe(u8, "out of memory") catch bun.outOfMemory() } };
         };
 
-        spawn_args.env_array.append(allocator, null) catch {
-            return .{ .err = .{ .custom = bun.default_allocator.dupe(u8, "out of memory") catch bun.outOfMemory() } };
-        };
+        // spawn_args.env_array.append(allocator, null) catch {
+        //     return .{ .err = .{ .custom = bun.default_allocator.dupe(u8, "out of memory") catch bun.outOfMemory() } };
```
Jarred-Sumner: this shouldn't be left commented out

either:
- it should be deleted
- it should be un-commented out if that was a mistake

paperclover: thanks for catching. deleted because the null terminator is inserted by the envp construction.

jdalton:
@paperdave I'm glad you caught this. I'm wary of our policy of recreating the wheel of APIs that exist in the platform in part because of things like this that fall through the cracks. I'd rather contribute to fixing the lower level APIs where possible and deviating from them only when reaaally needed.

PR #9643, opened 2024-03-26
Fix `bun install -g` in Dockerfiles

### LINKED ISSUE #8753: Bun global install is not working in a Dockerfile build

1.0.26+c75e768a6

Darwin 23.3.0 arm64 arm

1. Create `Dockerfile` with the content provided here: https://bun.sh/guides/ecosystem/docker
2. Anywhere add these 2 lines:
   ```
   RUN bun install --global cowsay
   RUN cowsay "Bun!"
   ```
3. Run: `docker build --pull -t bun-hello --progress plain .`
4. See error: `cowsay: not found`

Should not error out

### What do you see instead?

_No response_

### Additional information

_No response_

Electroid:
### What does this PR do?

We were not setting the `BUN_INSTALL_BIN` path, so `bun install -g` would not be added to path.

* Closes #8753 

Added docker integrations tests.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -40,7 +40,7 @@
   // C++
   "lldb.verboseLogging": false,
   "cmake.configureOnOpen": false,
-  "C_Cpp.errorSquiggles": "enabled",  
+  "C_Cpp.errorSquiggles": "enabled",
   "[cpp]": {
     "editor.defaultFormatter": "xaver.clang-format"
   },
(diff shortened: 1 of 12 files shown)


Jarred-Sumner (changes requested):
Tests don’t pass

Electroid:
Even though https://github.com/oven-sh/bun/commit/14c23cc429521bd935b2a0dafa36b0966685b0fe is merged, keeping this open so we can fix the integration tests

PR #9647, opened 2024-03-27
fix(docs): correct value for `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD`

trcio:
updates the explanation in the docs - value must be `true`, not `1`
as proven in the implementation:

https://github.com/oven-sh/bun/blob/1675349667109d23984fc60fbcc1c4d0a25e70bf/src/bun.js/javascript.zig#L831
https://github.com/oven-sh/bun/blob/1675349667109d23984fc60fbcc1c4d0a25e70bf/src/bun.js/javascript.zig#L834
https://github.com/oven-sh/bun/blob/1675349667109d23984fc60fbcc1c4d0a25e70bf/src/bun.js/javascript.zig#L3432

### What does this PR do?

- [ ] Code changes

- set `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD=1`
- run `bun --watch`
- change & save a watched file

You will observe that the terminal is cleared

Do the above with `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD=true` and it will work as expected

### DIFF
--- a/docs/runtime/env.md
+++ b/docs/runtime/env.md
@@ -169,7 +169,7 @@ These environment variables are read by Bun and configure aspects of its behavio
 ---
 
 - `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD`
-- If `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD=1`, then `bun --watch` will not clear the console on reload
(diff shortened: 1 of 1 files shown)


paperclover:
instead, what if we made the code check that BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD is not `0`, instead of a check that it is *exactly* `true`? otherwise, this is fine.

Jarred-Sumner:
IMO we should check if the environment variable value is 1 or true

PR #9648, opened 2024-03-27
test: use the same isWindows everywhere

nektro:


Jarred-Sumner:
This needs to be enforced by linter or it will just happen again and be unenforceable

PR #9649, opened 2024-03-27
test: pm: don't delete temporary directories

nektro:


Jarred-Sumner:
I think originally we did this because we ran out of disk space but I think we got bigger CI machines since 

there is a merge conflict but otherwise fine

jdalton (approved):
Looks good to me. Can you provide some context as to the issue this is solving?

nektro:
when debugging test failures it can be nice to inspect the temporary package files but because of these afterEach's the folders were all being deleted. after this patch that's no longer the case

PR #9651, opened 2024-03-27
fix(windows) socket and timers/performance tests

cirospaciari:
### What does this PR do?

- [x] Code changes

Run the tests

### DIFF
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -4053,8 +4053,7 @@ pub const Timer = struct {
                 if (this.cancelled) {
                     _ = uv.uv_timer_stop(&this.timer);
                 }
-                // libuv runs on the same thread
-                return this.runFromJSThread();
+                this.event_loop.enqueueTaskConcurrent(this.concurrent_task.from(this, .manual_deinit));
             }
 
             fn onRequest(req: *bun.io.Request) bun.io.Action {
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -4146,6 +4146,8 @@ pub const Timer = struct {
                 _ = this.scheduled_count.fetchAdd(1, .Monotonic);
                 const ms: usize = @max(interval orelse this.interval, 1);
                 if (Environment.isWindows) {
+                    // we MUST update the timer so we avoid early firing
+                    uv.uv_update_time(uv.Loop.get());
                     if (uv.uv_timer_start(&this.timer, TimerReference.onUVRequest, @intCast(ms), 0) != 0) @panic("unable to start timer");
                     return;
                 }
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -3797,7 +3797,8 @@ pub const Timer = struct {
             if (vm.isInspectorEnabled()) {
                 Debugger.willDispatchAsyncCall(globalThis, .DOMTimer, Timeout.ID.asyncID(.{ .id = this.id, .kind = kind }));
             }
-
+            vm.eventLoop().enter();
+            defer vm.eventLoop().exit();
             const result = callback.callWithGlobalThis(
                 globalThis,
                 args,
@@ -4053,7 +4054,7 @@ pub const Timer = struct {
                 if (this.cancelled) {
                     _ = uv.uv_timer_stop(&this.timer);
                 }
-                this.event_loop.enqueueTaskConcurrent(this.concurrent_task.from(this, .manual_deinit));
+                this.runFromJSThread();
             }
 
             fn onRequest(req: *bun.io.Request) bun.io.Action {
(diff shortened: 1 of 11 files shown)


test/js/bun/net/keep-event-loop-alive.js
```diff
@@ -1,11 +1,11 @@
+const port = process.argv[2] ? parseInt(process.argv[2]) : null;
+
 (async () => {
-  const port = process.argv[2] ? parseInt(process.argv[2]) : null;
-  await Bun.sleep(10);
```
cirospaciari, line 3: adding the `Bun.sleep` causes the process to exit earlier small repro:
````js
(async () => {
  console.log("before");
  await Bun.sleep(10);
  console.log("after");
})();
````

File: src/bun.js/api/BunObject.zig [hunk 2]
```diff
@@ -4053,8 +4053,7 @@ pub const Timer = struct {
                 if (this.cancelled) {
                     _ = uv.uv_timer_stop(&this.timer);
                 }
-                // libuv runs on the same thread
-                return this.runFromJSThread();
+                this.event_loop.enqueueTaskConcurrent(this.concurrent_task.from(this, .manual_deinit));
```
dylan-conway: can we do enter/exit?
```zig
this.event_loop.enter();
defer this.event_loop.exit();
this.runFromJSThread();
```
Because we're already on the thread we can free on I don't think we need to enqueue a task

src/bun.js/api/bun/socket.zig
```diff
@@ -1038,18 +1038,8 @@ pub const Listener = struct {
...
-                handlers.promise.deinit();
-                bun.default_allocator.destroy(tls);
-                const err = JSC.SystemError{
-                    .message = bun.String.static("Failed to connect"),
-                    .syscall = bun.String.static("connect"),
-                    .code = if (port == null) bun.String.static("ENOENT") else bun.String.static("ECONNREFUSED"),
-                };
-                exception.* = err.toErrorInstance(globalObject).asObjectRef();
-                return .zero;
+                tls.handleConnectError(@intFromEnum(bun.C.SystemErrno.ECONNREFUSED));
```
dylan-conway: should we still error instantly with `ENOENT` for unix sockets?

dylan-conway (commented):
Looks good, had two questions

src/bun.js/api/BunObject.zig
cirospaciari: Yeah actually added enter/exit before `callWithGlobalThis` we should always do this even on unix

src/bun.js/api/bun/socket.zig
cirospaciari: Yeah makes sense for a moment I miss read a comment and removed it.

cirospaciari: fixed on https://github.com/oven-sh/bun/pull/9651/commits/073e1ae165e83f1951f55e3d4c396be3d6a77ec2

src/bun.js/api/BunObject.zig
cirospaciari: fixed on https://github.com/oven-sh/bun/pull/9651/commits/073e1ae165e83f1951f55e3d4c396be3d6a77ec2

src/bun.js/api/BunObject.zig
```diff
@@ -4146,6 +4146,8 @@ pub const Timer = struct {
                 _ = this.scheduled_count.fetchAdd(1, .Monotonic);
                 const ms: usize = @max(interval orelse this.interval, 1);
                 if (Environment.isWindows) {
+                    // we MUST update the timer so we avoid early firing
+                    uv.uv_update_time(uv.Loop.get());
```
cirospaciari, line 4150: this should fix setTimeout and performance tests failures

test/js/bun/net/socket.test.ts
```diff
@@ -1,13 +1,13 @@
...
 import { expect, it } from "bun:test";
-import { bunEnv, bunExe, expectMaxObjectTypeCount } from "harness";
+import { bunEnv, bunExe, expectMaxObjectTypeCount, isWindows } from "harness";
 import { connect, fileURLToPath, SocketHandler, spawn } from "bun";
 
 it("should coerce '0' to 0", async () => {
   const listener = Bun.listen({
     // @ts-expect-error
     port: "0",
-    hostname: "localhost",
```
Jarred-Sumner: We should not need to do this 

Please fix the underlying reason for this being necessary instead of applying this bandaid.

Jarred-Sumner (changes requested):
Updating the time makes sense, but replacing localhost with 127.0.0.1 in our tests is not a good enough fix

test/js/bun/net/socket.test.ts
cirospaciari: is not strict necessary but localhost can be ipv6 or ipv4 so I just choose a ipv4 instead of localhost.

Jarred-Sumner: The bug here is that it didn’t pick the right one. This change shouldn’t be necessary

cirospaciari: agreed, just reverted it back and use server.hostname like we do in other tests. Happy Eyeballs should be the fix for ipv6/ipv4.

Jarred-Sumner (approved):
Thank you

PR #9652, opened 2024-03-27
fix(windows) fix performance.test.ts test

cirospaciari:
### What does this PR do?
on Windows libuv can trigger 0-5ms earlier a callback (normally triggers after but it can happen, specially with timers less than 50ms)

- [ ] Code changes
- [x] Test changes


cirospaciari:
fixed on https://github.com/oven-sh/bun/pull/9651

PR #9653, opened 2024-03-27
fix(test): fix crash in `expect.toHaveProperty`

dylan-conway:
### What does this PR do?
Passing `substring(i, j - i)` directly into `Identifier::fromString` would immediately deref the substring after creating the property name causing `getIfPropertyExists` to crash.

Changed this function to use `StringView`.

tested manually

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -3571,7 +3571,7 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsFromPath(JSC__JSValue JSValue0, JS
     JSValue path = JSValue::decode(arg1);
 
     if (path.isString()) {
-        StringView pathString = path.toWTFString(globalObject);
+        String pathString = path.toWTFString(globalObject);
         uint32_t length = pathString.length();
 
         if (length == 0) {
@@ -3640,13 +3640,8 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsFromPath(JSC__JSValue JSValue0, JS
                 jc = pathString.characterAt(j);
             }
 
-            StringView propNameStr = pathString.substring(i, j - i);
-
-            Identifier ident = propNameStr.is8Bit()
-                ? Identifier::fromString(vm, propNameStr.characters8(), propNameStr.length())
-                : Identifier::fromString(vm, propNameStr.characters16(), propNameStr.length());
-
-            PropertyName propName = PropertyName(ident);
+            String propNameStr = pathString.substring(i, j - i);
+            PropertyName propName = PropertyName(Identifier::fromString(vm, propNameStr));
 
             currProp = currProp.toObject(globalObject)->getIfPropertyExists(globalObject, propName);
             RETURN_IF_EXCEPTION(scope, {});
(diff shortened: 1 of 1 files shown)


src/bun.js/bindings/bindings.cpp
```diff
@@ -3571,7 +3571,7 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsFromPath(JSC__JSValue JSValue0, JS
     JSValue path = JSValue::decode(arg1);
 
     if (path.isString()) {
-        String pathString = path.toWTFString(globalObject);
```
Jarred-Sumner: I’m confused how this fixes the issue 

isn’t this a use after free?

shouldn’t we move the String to the top of the function and the String will be kept alive until the function ends?

dylan-conway: The use after free comes from the substring `String` on this line. It's created and passed into `Identifier::fromString`, then destroyed on the same line before `getIfPropertyExists` is called.
```c
PropertyName propName = PropertyName(Identifier::fromString(vm, pathString.substring(i, j - i)));
```
A simple fix for this is just assigning the substring to a variable before creating the identifier
```c
String substring = pathString.substring(i, j - i));
PropertyName propName = PropertyName(Identifier::fromString(vm, substring));
```
The `StringView` wasn't necessary, but seemed like a better choice because we aren't changing the string, only reading from it.

src/bun.js/bindings/bindings.cpp
```diff
@@ -3640,7 +3640,14 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsFromPath(JSC__JSValue JSValue0, JS
                 jc = pathString.characterAt(j);
             }
 
-            PropertyName propName = PropertyName(Identifier::fromString(vm, pathString.substring(i, j - i)));
+            StringView propNameStr = pathString.substring(i, j - i);
```
Jarred-Sumner: I don't think this is a good fix

This should be: 
```suggestion
            String propNameStr = pathString.substring(i, j - i);
            PropertyName propName = Identifier::fromString(vm, propNameStr);
```

Jarred-Sumner: The reason the below works is because Identifier::fromString(vm, const LChar*, unsigned) creates a new AtomString

src/bun.js/bindings/bindings.cpp
```diff
@@ -3571,7 +3571,7 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsFromPath(JSC__JSValue JSValue0, JS
     JSValue path = JSValue::decode(arg1);
 
     if (path.isString()) {
-        String pathString = path.toWTFString(globalObject);
+        StringView pathString = path.toWTFString(globalObject);
```
Jarred-Sumner: This definitely still creates the WTF::String, it just immediately destroys it which means this introduces a different use-after-free. This time, if `path` is a JSRopeString, this would become a use-after-free.

src/bun.js/bindings/bindings.cpp
dylan-conway: Updated to use `String`

Jarred-Sumner (approved):

Issue #9586, opened 2024-03-23
call fetch() is error inWindows 


Cikaros:
### What version of Bun is running?

1.0.35-canary.4+ee5fd51e8

### What platform is your computer?

x64

### What steps can reproduce the bug?

```js
const response = await fetch("https://bun.sh"); // Happens only at https
const html = await response.text(); // HTML string
```

### What is the expected behavior?

Process finished with exit code 0

### What do you see instead?

Syscall: fetch() failed. For more information, pass `verbose: true` in the second argument to fetch()
 path: "https://bun.sh"
 
 Process finished with exit code 1

paperclover:
Once the above PR merges, can you try this again and report back what error code is given? Otherwise, there isn't anything I can do about this.

vikas5914:
@paperdave  Example: 

```
const response = await fetch('https://example.com', {
  proxy: "http://twqhuxwy-rotate:ez1w1miev6ku@p.webshare.io:80/",
  verbose: true
});

console.log(await response.text())
```

Output: 

```
Request: GET /
        Connection: keep-alive
        User-Agent: Bun/1.1.7
        Accept: */*
        Host: example.com
        Accept-Encoding: gzip, deflate, br

Response: < 200 OK

Syscall: Syscall fetching "https://example.com/". For more information, pass `verbose: true` in the second argument to fetch()
 path: "https://example.com/"

Bun v1.1.7 (Windows x64)
```

vikas5914:
its not

vikas5914:
@Jarred-Sumner still having the problem 

```
[fetch] > HTTP/1.1 GET https://example.com/
[fetch] > Connection: keep-alive
[fetch] > User-Agent: Bun/1.1.22
[fetch] > Accept: */*
[fetch] > Host: example.com
[fetch] > Accept-Encoding: gzip, deflate, br

[fetch] < 200 OK

Syscall: Syscall fetching "https://example.com/". For more information, pass `verbose: true` in the second argument to fetch()
 path: "https://example.com/"

Bun v1.1.22-canary.96+df33f2b2a (Windows x64)
```

Jarred-Sumner:
Are you using a proxy? A vpn? Is there anything more you can share about your computer?

vikas5914:
@Jarred-Sumner I am using proxy with fetch.

```
const response = await fetch('https://example.com', {
  proxy: "http://twqhuxwy-rotate:ez1w1miev6ku@p.webshare.io:80/",
[...]

PR #9663, opened 2024-03-27
add a better error message for fetch when it fails with an unknown code

paperclover:
### What does this PR do?

Motivation: #9586 has a useless error message.


src/bun.js/webcore/response.zig
```diff
@@ -1304,7 +1304,8 @@ pub const Fetch = struct {
                     error.STORE_LOOKUP => bun.String.static("Issuer certificate lookup error"),
                     error.NAME_CONSTRAINTS_WITHOUT_SANS => bun.String.static("Issuer has name constraints but leaf has no SANs"),
                     error.UNKKNOW_CERTIFICATE_VERIFICATION_ERROR => bun.String.static("unknown certificate verification error"),
-                    else => bun.String.static("fetch() failed. For more information, pass `verbose: true` in the second argument to fetch()"),
+
+                    else => |e| bun.String.createFormat("fetch() failed with {s}. For more information, pass `verbose: true` in the second argument to fetch()", .{@errorName(e)}) catch bun.outOfMemory(),
```
Jarred-Sumner: maybe something like

```suggestion
                    else => |e| bun.String.createFormat("{s} fetching {}. For more information, pass `verbose: true` in the second argument to fetch()", .{@errorName(e), bun.fmt.quote(this.url) }) catch bun.outOfMemory(),
```

src/string.zig
```diff
@@ -397,6 +397,14 @@ pub const String = extern struct {
         return BunString__fromUTF16(bytes.ptr, bytes.len);
     }
 
+    pub fn createFormat(comptime fmt: []const u8, args: anytype) !String {
+        var sba = std.heap.stackFallback(16384, bun.default_allocator);
```
jdalton, line 401: Is `16384` a constant we can stash somewhere?

paperclover: it is 2 ^ 14 bytes (16 kibibytes). this number is chosen arbitrarily.

src/bun.js/WebKit
gvilums, line 1: is the submodule update intentional?

gvilums (approved):

src/bun.js/WebKit
paperclover: no

PR #9664, opened 2024-03-27
setup must not be async

buhrmi:
when the setup function of a plugin is async, it introduces a race condition. most of the time, the plugin isn't finished loading in time.


buhrmi:
will make new PR with await Bun.plugin

PR #9667, opened 2024-03-28
complex-workspace.test.ts: print stdio output on init failure

nektro:


test/cli/install/migration/complex-workspace.test.ts
```diff
@@ -44,27 +45,36 @@ test("the install succeeds", async () => {
   var subprocess = Bun.spawn([bunExe(), "reset.ts"], {
     env: bunEnv,
     cwd,
-    stdio: ["ignore", "ignore", "ignore"],
+    stdio: ["ignore", "pipe", "pipe"],
```
Jarred-Sumner: this should just be `inherit` no?

nektro: that works too but given they were ignore before I assumed it'd be noise on success

PR #9676, opened 2024-03-28
fix(http2): fix crash in http2 tests

dylan-conway:
### What does this PR do?
Two settings were missing causing an `enumFromInt` call to fail (received 9 when max value is 6). Added the settings.

@cirospaciari can you check that we are using these settings correctly?

manually

### DIFF
--- a/test/js/node/http2/node-echo-server.fixture.js
+++ b/test/js/node/http2/node-echo-server.fixture.js
@@ -17,7 +17,7 @@ server.on("stream", (stream, headers, flags) => {
     "set-cookie": setCookie,
   });
   // errors here are not useful the test should handle on the client side
-  stream.on("error", () => {});
+  stream.on("error", err => console.error(err));
 
   if (headers["x-wait-trailer"]) {
     const response = { headers, flags };
--- a/test/js/node/http2/node-echo-server.fixture.js
+++ b/test/js/node/http2/node-echo-server.fixture.js
@@ -16,6 +16,7 @@ server.on("stream", (stream, headers, flags) => {
     ":status": 200,
     "set-cookie": setCookie,
   });
+  // errors here are not useful the test should handle on the client side
   stream.on("error", () => {});
 
   if (headers["x-wait-trailer"]) {
--- a/test/js/node/http2/node-echo-server.fixture.js
+++ b/test/js/node/http2/node-echo-server.fixture.js
@@ -16,6 +16,7 @@ server.on("stream", (stream, headers, flags) => {
     ":status": 200,
     "set-cookie": setCookie,
   });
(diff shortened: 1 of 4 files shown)


cirospaciari:
Will check  but this settings are not in https://datatracker.ietf.org/doc/html/rfc9113#name-settings don't think we should change the settings frame it self, I found https://datatracker.ietf.org/doc/html/rfc9218 need to read more careful, we should not crash if we receive a invalid setting, we should disconnect and error or ignore instead of using enumFromInt in a not supported settings

Jarred-Sumner:
@cirospaciari when using exhaustive enums in zig, the `else` condition is unreachable if the enum values are not in the list of values. So if the enum is coming from user input and not from lshpack, then it must either be a non-exhaustive enum or it must not be using enums.

test/js/node/http2/node-echo-server.fixture.js
```diff
@@ -16,6 +16,8 @@ server.on("stream", (stream, headers, flags) => {
     ":status": 200,
     "set-cookie": setCookie,
   });
+  // errors here are not useful the test should handle on the client side
+  stream.on("error", () => {});
```
Jarred-Sumner:
```suggestion
  stream.on("error", err => console.error(err));
```

test/js/node/http2/node-http2-memory-leak.js
```diff
@@ -32,8 +32,8 @@ async function runRequests(iterations) {
   for (let j = 0; j < iterations; j++) {
     let client = http2.connect(info.url, { rejectUnauthorized: false });
     let promises = [];
-    // 100 multiplex POST connections per iteration
-    for (let i = 0; i < 100; i++) {
+    // 10 multiplex POST connections per iteration
```
Jarred-Sumner: why not 100?

Jarred-Sumner:
```suggestion
    // 100 multiplex POST connections per iteration
```

test/js/node/http2/node-http2-memory-leak.js
```diff
@@ -32,8 +32,8 @@ async function runRequests(iterations) {
...
+    for (let i = 0; i < 10; i++) {
```
Jarred-Sumner:
```suggestion
    for (let i = 0; i < 100; i++) {
```

test/js/node/http2/node-http2-memory-leak.js
cirospaciari: I think 100 is too much, we can up for 500 iterations instead of 50 connections and would be more useful for detecting leaks than 100 multiplex, also we check for max 1 body leak per connection.

PR #9630, opened 2024-03-25
Fix some shell regressions

zackradisic:
### What does this PR do?

- Fixes a bug where writing above 128kb with the shell might crash (incorrect logic when exceeding shrink threshold)
- Fixes hanging in one the shell/leak.test.ts (For some reason I made the async deinitialization functions deinit the IOWriter/IOReader on a separate thread and I think this caused a race condition leading to the shell hanging)
- Lazily dupes stdin, stdout, stderr
- Enqueues the shell subprocess exit handling to the event loop, to prevent a crash caused when the shell subprocess exits immediately

This fixes some shell regression
- [x] Code changes

### How did you verify your code works?

- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [x] I included a test for the new code, or an existing test covers it
- [x] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed
- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test-file-name.test`)

PR #9684, opened 2024-03-28
Fix shell regressions part 2

zackradisic:
### What does this PR do?

Same as the recently reverted as #9630 except trying to make it not hang CI


test/js/bun/shell/bunshell.test.ts
```diff
@@ -817,6 +816,87 @@ describe("deno_task", () => {
...
       .run();
   });
+
+  test("big_data", async () => {
+    const writerCode = /* ts */ `
+
+    const writer = Bun.stdout.writer();
+    const buf = new Uint8Array(128 * 1024).fill('a'.charCodeAt(0))
+    for (let i = 0; i < 10; i++) {
+      writer.write(buf);
```
Jarred-Sumner, line 826:
```suggestion
      writer.write(buf);
      await writer.flush();
```

test/js/bun/shell/bunshell.test.ts
```diff
@@ -817,6 +816,87 @@ describe("deno_task", () => {
...
+    async () => {
+      const inputCode = /* ts */ `
+    const downArrow = '\\x1b[B';
+    const enterKey = '\\x0D';
+    await Bun.sleep(100)
+    const writer = Bun.stdout.writer();
+    writer.write(downArrow)
+    await Bun.sleep(100)
+    writer.write(enterKey)
+    writer.flush()
```
Jarred-Sumner, line 853:
```suggestion
    await writer.flush()
```

Jarred-Sumner: this shouldn't be necessary actually

PR #9691, opened 2024-03-28
fix!: do not lookup cwd in which

paperclover:
### What does this PR do?

The `which` command does not look up binaries in the current working directory unless cwd is explicitly in `PATH`, but our `Bun.which` function does a lookup in cwd anyways.

While I could not find the source code for `which` to verify this is how `which` works, it is easy to observe this behavior.

Since this is a breaking change in behavior, the `breaking` tag is added here.

### DIFF
--- a/test/js/bun/util/which.test.ts
+++ b/test/js/bun/util/which.test.ts
@@ -1,11 +1,18 @@
 import { test, expect } from "bun:test";
 
-import { which } from "bun";
+import { $, which } from "bun";
 import { rmSync, chmodSync, mkdirSync, realpathSync } from "node:fs";
 import { join, basename } from "node:path";
 import { tmpdir } from "node:os";
-import { cpSync, rmdirSync } from "js/node/fs/export-star-from";
-import { isIntelMacOS, isWindows } from "../../../harness";
+import { rmdirSync } from "js/node/fs/export-star-from";
+import { isIntelMacOS, isWindows, tempDirWithFiles } from "../../../harness";
+
+{
+  const delim = isWindows ? ";" : ":";
+  if (`${delim}${process.env.PATH}${delim}`.includes(`${delim}.${delim}`)) {
+    throw new Error("$PATH includes . which will mess up the test");
+  }
+}
 
 function writeFixture(path: string) {
   var fs = require("fs");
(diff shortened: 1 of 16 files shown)


test/js/bun/util/which.test.ts
```diff
@@ -96,16 +101,28 @@ if (isWindows) {
...
-        }),
-      ).toBe(abs);
     } finally {
       process.chdir(origDir);
       rmSync(basedir, { recursive: true, force: true });
     }
   });
 }
+
+test("which does not look in the current directory", async () => {
```
jdalton:
```suggestion
test("Bun.which does not look in the current directory", async () => {
```

test/js/bun/util/which.test.ts
```diff
@@ -1,11 +1,18 @@
...
 
-import { which } from "bun";
+import { $, which } from "bun";
 import { rmSync, chmodSync, mkdirSync, realpathSync } from "node:fs";
 import { join, basename } from "node:path";
 import { tmpdir } from "node:os";
-import { cpSync, rmdirSync } from "js/node/fs/export-star-from";
-import { isIntelMacOS, isWindows } from "../../../harness";
+import { rmdirSync } from "js/node/fs/export-star-from";
+import { isIntelMacOS, isWindows, tempDirWithFiles } from "../../../harness";
```
jdalton: I think we alias "harness"
```suggestion
import { isIntelMacOS, isWindows, tempDirWithFiles } from "harness";
```

test/js/bun/util/which.test.ts
```diff
@@ -1,11 +1,18 @@
...
+
+{
+  const delim = isWindows ? ";" : ":";
+  if (`${delim}${process.env.PATH}${delim}`.includes(`${delim}.${delim}`)) {
+    throw new Error("$PATH includes . which will mess up the test");
```
jdalton: Can we move this assert into a test block. Could be done as a function `checkPATH()` inside the `test("Bun.which ...` below. That way it doesn't block up the whole file.

paperclover: this check is supposed to block the entire file

jdalton (approved):

jdalton:
This was a lot to pull out. Thank you @paperdave! It's simpler and more aligned. Win!

PR #9692, opened 2024-03-28
Begin to run tests from Node.js repository

Electroid:
### What does this PR do?

- [ ] Code changes

Tests!


test/node/buffer/arraybuffer.js
```diff
@@ -0,0 +1,159 @@
+"use strict";
```
paperclover: are all of these files 1:1 copied from node.js?

what is our plan for syncing this? should it even be pasted and committed into our repo 1:1, this is a lot of text content that is (hopefully) automatically sourced.

Electroid: It's 1:1, aside from prettier formatting. I thought it would be useful to have them checked out, but we could just have them to gitignore and have the script clone them over.

.gitmodules
```diff
@@ -83,3 +83,11 @@ ignore = dirty
 depth = 1
 shallow = true
 fetchRecurseSubmodules = false
+[submodule "test/node/node"]
+	path = test/node/node
+	url = https://github.com/nodejs/node.git
```
jdalton: Can we lock this to a release hash instead of main.

test/node/common/index.js
```diff
@@ -0,0 +1,122 @@
+// https://github.com/nodejs/node/blob/c975384264dc553de62398be814d0c66fc1fc1fb/test/common/index.js
+
```
jdalton, line 2: Is the `assert.js` based on a Node file? If so the commit comment like this addd to that is helpful.

test/node/common/index.js
```diff
@@ -0,0 +1,122 @@
...
+const isWindows = platform === "win32";
+const isSunOS = platform === "sunos";
+const isFreeBSD = platform === "freebsd";
+const isOpenBSD = platform === "openbsd";
+const isLinux = platform === "linux";
+const isOSX = platform === "darwin";
+const isAsan = false;
+const isPi = false;
+const isDumbTerminal = env.TERM === "dumb";
+
```
jdalton, line 22: Thank you! Mimicking this for every one off was becoming non-scalable

test/node/scripts/import-tests.mjs
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      filename = filename.slice(basename(groupPrefix).length + 1);
+    }
+
+    const dst = `${groupMatch}/${filename}`;
+    const dstPath = new URL(dst, testPath);
+
+    const { status, error, stderr, stdout } = spawnSync("bunx", ["prettier", "--stdin", "--parser=babel"], {
+      // HACK: Replace `assert` with `assert.ok` because `mock.module`
+      //       does not preserve the default `assert` function.
```
jdalton: ☝️ Is this a future TODO for us to fix?

.gitmodules
Electroid: It's pinned to a specific commit

test/node/scripts/import-tests.mjs
Electroid: Yes, this is a bug with module mocks.

.gitmodules
nektro: yeah git submodules are always a specific commit

PR #9693, opened 2024-03-28
add server ws.terminate

paperclover:
### What does this PR do?

Missing API found when testing create-react-app

Test


UltimateGG:
When I call ws.terminate() I get an error see #11972

PR #9695, opened 2024-03-28
fix!: default process.env.NODE_ENV to undefined

### LINKED ISSUE #7829: `process.env.NODE_ENV` should be `undefined` as default

paperclover:
### What does this PR do?

Fixes #7829

This is a breaking change.

### DIFF
--- a/src/bun_js.zig
+++ b/src/bun_js.zig
@@ -222,16 +222,6 @@ pub const Run = struct {
             .unspecified => {},
         }
 
-        // Set NODE_ENV to a value if something else hadn't already set it
-        const node_env_entry = try b.env.map.getOrPutWithoutValue("NODE_ENV");
-        if (!node_env_entry.found_existing) {
-            node_env_entry.key_ptr.* = try b.env.allocator.dupe(u8, node_env_entry.key_ptr.*);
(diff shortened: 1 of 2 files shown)


paperclover:
NOTE: lack of major test changes are because the default value of NODE_ENV in the bundler and test runner will NOT change, and we have tests verifying this. only `bun run script.js` has a change in behavior.

jdalton (approved):

PR #9696, opened 2024-03-29
fix(windows `bun upgrade`): skip cleaning up old binary

paperclover:
replacing the binary using `bun upgrade` is problematic, as you cannot delete a binary that is currently running.

my previous solution used a subprocess that outlived the upgrader, but the only way to do this on windows will create a console window for the process. "okay sure, it will flicker in the taskbar but all is fine", i thought.

it seems this window steals user focus for some reason:

image

and

image

yikes. i'm going to remove this behavior for now and if this problem can be solved then we can do so later. the new behavior is equal to what some other software does, for example `deno upgrade`.

### DIFF
--- a/src/cli/upgrade_command.zig
+++ b/src/cli/upgrade_command.zig
@@ -948,46 +948,12 @@ pub const UpgradeCommand = struct {
 
             if (Environment.isWindows) {
                 if (outdated_filename) |to_remove| {
-                    current_executable_buf[target_dir_.len] = '\\';
-                    const delete_old_script = try std.fmt.allocPrint(
-                        ctx.allocator,
-                        // What is this?
-                        // 1. spawns powershell
-                        // 2. waits for all processes with the same path as the current executable to exit (including the current process)
-                        // 3. deletes the old executable
-                        //
-                        // probably possible to hit a race condition, but i think the worst case is simply the file not getting deleted.
-                        //
-                        // in that edge case, the next time you upgrade it will simply override itself, fixing the bug.
-                        //
-                        // -NoNewWindow doesnt work, will keep the parent alive it seems
-                        // -WindowStyle Hidden seems to just do nothing, not sure why.
-                        // Using -WindowStyle Minimized seems to work, but you can spot a powershell icon appear in your taskbar for about ~1 second
-                        //
-                        // Alternative: we could simply do nothing and leave the `.outdated` file.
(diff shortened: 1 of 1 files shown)


src/cli/upgrade_command.zig
```diff
@@ -948,46 +948,12 @@ pub const UpgradeCommand = struct {
...
-                        .argv = &delete_argv,
-                        .cwd = tmpdir_path,
-                        .max_output_bytes = 512,
-                    }) catch {};
+                    // TODO: this file gets left on disk
+                    //
+                    // We should remove it, however we cannot remove an exe that is still running.
+                    // A prior approach was to spawn a subprocess to remove the file, but that
+                    // would open a terminal window, which steals user focus (even if minimized).
+                    _ = to_remove;
```
jdalton, line 956: If we can't remove it can we move it (rename it). If we can rename it then we can on future runs detected the renamed file `<oldfilename>_deleteme` and then remove it.

gvilums: we unfortunately can't rename it either AFAIK

gvilums: It turns out you can rename a running executable. We could follow a strategy as described [here](https://stackoverflow.com/a/7198760)

paperclover: I specifically want to avoid checking for an old exe on every startup of `bun.exe`. I am already following those instructions except for the last one.

jdalton: @gvilums 
>It turns out you can rename a running executable. 

Awww Yiss! From my Windows days I remember hacking around with renaming when it wouldn't let me delete.

@paperdave 

Instead of checking for an old version on startup it can be done on the next `upgrade` call. If it's just a periodic cleanup.

paperclover: > on the next upgrade call

sure, however i don't see too much use here, as the `bun.exe.outdated` file will simply be re-ovewritten. deleting it only really does something in the case you do `bun upgrade` twice in a row.

jdalton: True. There may be other lazy opportunities to perform cleanup. There could even be a once very so many executions try to cleanup or after running for so long try to cleanup (just so it stays out of the hot path).

paperclover:
im going to merge this in interest in having it ready for 1.1. while leaving some data on disk isn't perfectly ideal, it is better than the status quo.

and again i don't feel like this is particularly bad considering other software has taken this route.

PR #9697, opened 2024-03-29
fix bunx-bins verdaccio package

paperclover:
### What does this PR do?

1. remove the python test since not all machines have `python3` (it is not in our devenv setup guide)
2. compile native.exe for baseline cpu. this will fix install registry in windows


nektro:
`windows bin linking shim should work` test is still failing

test/cli/install/registry/packages/bunx-bins/package.json
```diff
@@ -43,7 +44,7 @@
 	"_distfiles": {},
 	"_attachments": {
 		"bunx-bins-1.0.0.tgz": {
-			"shasum": "c23382ce43a04a94035050ac736f04179a8639ff",
+			"shasum": "a77e200b47158585a4bf70ef934202c5fb2b0dd7",
 			"version": "1.0.0"
 		}
```
jdalton, line 49: ☝️ Is this produced from a pack'd npm/bun package? Are there documented steps with how to update this test when needed?

jdalton (approved):

nektro (changes requested):
see comment above, relevant test is still not passing as of writing

test/cli/install/registry/packages/bunx-bins/package.json
paperclover: yes, it can be updated using `npm publish`

PR #9698, opened 2024-03-29
fix: EnvStr.initSlice handles zero length strings more reliably.

### LINKED ISSUE #9481: Bun crash `integer cast truncated bits`

1.0.32-canary.21+5fec71bda

Try run my React ClientApp on Windows and it crash

bun start      
```     
$ react-scripts start

uh-oh: integer cast truncated bits
bun will crash now 😭😭😭

----- bun meta -----
Bun v1.0.32-canary.21 (5fec71bd) Windows x64
AutoCommand: dotenv
Elapsed: 11ms | User: 0ms | Sys: 15ms
RSS: 24.70MB | Peak: 24.70MB | Commit: 68.12MB | Faults: 6174
----- bun meta -----

0   00007FF751A5A9B8
1   ???
2   ???
3   ???
4   ???
5   ???
6   ???
7   ???
8   ???
9   ???

Search GitHub issues https://bun.sh/issues or join in #windows channel in https://bun.sh/discord

thread 24012 panic: integer cast truncated bits
???:?:?: 0x7ff7506b2f51 in ??? (bun.exe)
???:?:?: 0x7ff7506b644c in ??? (bun.exe)
???:?:?: 0x7ff7506c85f6 in ??? (bun.exe)
???:?:?: 0x7ff7506cc7d5 in ??? (bun.exe)
???:?:?: 0x7ff75041157f in ??? (bun.exe)
???:?:?: 0x7ff752f10ae3 in ??? (bun.exe)
???:?:?: 0x7ffd9158257c in ??? (KERNEL32.DLL)
???:?:?: 0x7ffd92bcaa57 in ??? (ntdll.dll)
```

It should start my application.

### What do you see instead?

_No response_

### Additional information

_No response_

paperclover:
### What does this PR do?

Fixes #9481

Zero length strings from the env parser seem to not have a valid address (0xffff...ff). I'm not sure why, but this fixes it by preventing the `@intCast` on the invalid address.

Another route is to use here is `@truncate`, but this needs a complex assertion to ensure the truncation is valid. This route can be taken if desired.

### DIFF
--- a/src/shell/interpreter.zig
+++ b/src/shell/interpreter.zig
@@ -371,6 +371,10 @@ pub const EnvStr = packed struct {
     };
 
     inline fn initSlice(str: []const u8) EnvStr {
+        if (str.len == 0)
+            // Zero length strings may have invalid pointers, leading to a bad integer cast.
+            return .{ .tag = .slice, .ptr = @intCast(@intFromPtr(&"")), .len = 0 };
+
         return .{
             .ptr = @intCast(@intFromPtr(str.ptr)),
             .tag = .slice,
@@ -379,6 +383,11 @@ pub const EnvStr = packed struct {
     }
 
     fn initRefCounted(str: []const u8) EnvStr {
+        if (str.len == 0)
+            // Zero length strings may have invalid pointers, leading to a bad integer cast.
+            // Instead of ref-counting zero lens, just be the same as the above initSlice
+            return .{ .tag = .slice, .ptr = @intCast(@intFromPtr(&"")), .len = 0 };
+
         return .{
             .ptr = @intCast(@intFromPtr(RefCountedStr.init(str))),
             .tag = .refcounted,
--- a/src/shell/interpreter.zig
+++ b/src/shell/interpreter.zig
@@ -358,14 +358,18 @@ pub const IO = struct {
 /// So environment strings can be ref counted or borrowed slices
 pub const EnvStr = packed struct {
     ptr: u48,
-    tag: Tag,
+    tag: Tag = .empty,
     len: usize = 0,
 
     const print = bun.Output.scoped(.EnvStr, true);
 
     const Tag = enum(u16) {
+        /// no value
+        empty,
+
         /// Dealloced by reference counting
         refcounted,
+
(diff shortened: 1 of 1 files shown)


dylan-conway (approved):

src/shell/interpreter.zig
```diff
@@ -371,6 +371,10 @@ pub const EnvStr = packed struct {
     };
 
     inline fn initSlice(str: []const u8) EnvStr {
+        if (str.len == 0)
+            // Zero length strings may have invalid pointers, leading to a bad integer cast.
+            return .{ .tag = .slice, .ptr = @intCast(@intFromPtr(&"")), .len = 0 };
```
dylan-conway: .ptr, this is a ptr to a slice

dylan-conway (changes requested):

src/shell/interpreter.zig
```diff
@@ -379,6 +383,11 @@ pub const EnvStr = packed struct {
     }
 
     fn initRefCounted(str: []const u8) EnvStr {
+        if (str.len == 0)
+            // Zero length strings may have invalid pointers, leading to a bad integer cast.
+            // Instead of ref-counting zero lens, just be the same as the above initSlice
+            return .{ .tag = .slice, .ptr = @intCast(@intFromPtr(&"")), .len = 0 };
```
Jarred-Sumner: This does feel kind of sketchy. Maybe `tag` can be used to indicate a zero-length slice instead? and then `.slice()` can look for this?

zackradisic: I feel like maybe doing @truncate is better, since it's more related to unexpectedly receiving pointers >= 2^48 rather than length being zero.

src/shell/interpreter.zig
```diff
@@ -379,6 +383,11 @@ pub const EnvStr = packed struct {
     }
 
     fn initRefCounted(str: []const u8) EnvStr {
+        if (str.len == 0)
```
zackradisic, line 395: It's not necessary to do the check here because the pointer being casted is the `RefCountedStr` pointer and not the `str.pter`

PR #9701, opened 2024-03-29
Fix Randomly Broken Responses when Sending a File

### LINKED ISSUE #7954: Bun.serve fetch return new Response(Bun.file("test.jpg")); sometimes returns an empty response

1.0.20+09d51486e

Darwin 23.2.0 arm64 arm

`index.js`:
```javascript
Bun.serve({
  async fetch(request) {
    if (request.url !== "http://localhost:3000/") {
      return new Response(Bun.file("test.jpg"));
    }

    let html = "";
    for (let index = 0; index < 100; index++) {
      html += `<img src="/${index}" onerror="console.log(this.src, 'failed to load');" />`;
    }

    return new Response(html, {
      headers: { "Content-Type": "text/html" },
    });
  },
});
```

- Add any `test.jpg` image alongside `index.js`.
- Run `bun .`
- Visit http://localhost:3000/ 
- Open the developer tools Console tab
- Observe several images probably having failed to load

All images have loaded on the first try.

### What do you see instead?

A random amount (between 2-10 % on different reloads) of the images fail to get served:
image

The request is actually a 200 but the response headers are written at the start of the response body instead of into the HTTP stream and then properly delimited it seems. I've also seen seemingly empty response bodies under different circumstances, but I don't have a repro for that now.

Observing the Network tab, for the failed requests, we can confirm the headers are written into the response body:
image

I can reproduce this almost every load in Firefox, every tenth load or so in Safari and seemingly never in Chrome. I guess different browser have different tolerance to the ways the HTTP payload can be (mis?)structured.

### Additional information

I was on the fence about sending this in because at first it seemed like maybe it was a Firefox issue but when I reproduced in Safari I figured this probably is a Bun issue with how the HTTP payload is serialized. If I'm wrong, please accept my apologies.

dl-eric:
### What does this PR do?
- Closes #7954 

This PR fixes an issue where Bun will send an extra CRLF (/r/n) after sending a file for HTTP Servers. This extra CRLF is (1) not defined in the [HTTP protocol](https://www.rfc-editor.org/rfc/rfc7230#section-3.5) and (2) can fail entire responses, depending on how the client handles these extra characters.

- [x] Code changes

Wrote a test that fails without my change and passes with my change. It relies on a large volume of requests to trigger the bug, so I'm not sure how deterministic it is. But at least my machine was able to hit the bug every single test run on both debug and release builds.

Example test failure:
```
Malformed_HTTP_Response: fetch() failed. For more information, pass `verbose: true` in the second argument to fetch()
 path: "http://localhost:48349/"
✗ Response > should consume body correctly > with Bun.file() sendfile with request/response [143.06ms]
```

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -1561,7 +1561,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         }
 
         /// Complete the response without writing a body payload.
-        /// `sendCrlf` if true will send an /r/n at the end of the HTTP response.
+        /// `sendCrlf` if true will send an \r\n at the end of the HTTP response.
         /// If we already wrote a body and are using this function to complete the response, then
         /// `sendCrlf` should be false.
         pub fn endWithoutBody(this: *RequestContext, closeConnection: bool, sendCrlf: bool) void {
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -1560,13 +1560,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             }
         }
 
-        pub fn endWithoutBody(this: *RequestContext, closeConnection: bool) void {
+        /// Complete the response without writing a body payload.
+        /// `sendCrlf` if true will send an /r/n at the end of the HTTP response.
+        /// If we already wrote a body and are using this function to complete the response, then
+        /// `sendCrlf` should be false.
+        pub fn endWithoutBody(this: *RequestContext, closeConnection: bool, sendCrlf: bool) void {
             if (this.resp) |resp| {
                 if (this.flags.is_waiting_for_request_body) {
                     this.flags.is_waiting_for_request_body = false;
                     resp.clearOnData();
                 }
-                resp.endWithoutBody(closeConnection);
+                resp.endWithoutBody(closeConnection, sendCrlf);
                 this.resp = null;
             }
         }
@@ -1866,7 +1870,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         fn cleanupAndFinalizeAfterSendfile(this: *RequestContext) void {
             if (this.resp) |resp| {
                 resp.overrideWriteOffset(this.sendfile.offset);
-                this.endWithoutBody(this.shouldCloseConnection());
+                this.endWithoutBody(this.shouldCloseConnection(), false);
             }
             // use node syscall so that we don't segfault on BADF
             if (this.sendfile.auto_close)
@@ -2879,7 +2883,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                             resp.writeStatus("404 Not Found");
                             this.flags.has_written_status = true;
                         }
-                        this.endWithoutBody(this.shouldCloseConnection());
+                        this.endWithoutBody(this.shouldCloseConnection(), true);
                     },
                     else => {
                         if (!this.flags.has_written_status) {
@@ -5949,7 +5953,7 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
 
                 if (req_len > this.config.max_request_body_size) {
                     resp.writeStatus("413 Request Entity Too Large");
-                    resp.endWithoutBody(true);
+                    resp.endWithoutBody(true, true);
                     this.finalize();
                     return;
                 }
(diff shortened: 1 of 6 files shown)


src/bun.js/api/server.zig
```diff
@@ -1560,13 +1560,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             }
         }
 
-        pub fn endWithoutBody(this: *RequestContext, closeConnection: bool) void {
+        /// Complete the response without writing a body payload.
+        /// `sendCrlf` if true will send an /r/n at the end of the HTTP response.
```
jdalton:
```suggestion
        /// `sendCrlf` if true will send an \r\n at the end of the HTTP response.
```

test/js/web/fetch/fetch.test.ts
```diff
@@ -1259,6 +1259,33 @@ describe("Response", () => {
...
+        },
+      });
+
+      let promises = [];
+
+      // Issue a sufficiently large volume of requests to make sure they all succeed.
+      // This is a repro strategy for #7954. Essentially there was an issue where we
+      // were writing an additional /r/n at the end of responses, which interfered
+      // with adjacent requests on the same connection.
+      for (let i = 0; i < 2000; i++) {
```
jdalton: can this be done with less? While I dig the reasoning and we have done so in the past doing this for all-the-things will be a drag on the test runner times.

jdalton:
Thank you @dl-eric! I dig the detail and unit tests!

test/js/web/fetch/fetch.test.ts
dl-eric: Hey, so here's some data from my machine with an AMD 5800X 8c/16t CPU:

Without my change, I ran the test 10 times with different # of iterations in that for loop, and on different build flavors.

## Release Build:
\# iterations : How many test failures out of 10 runs
200: 0/10 <- bad, test doesn't capture bug
300: 3/10 <- flakey
400: 10/10 <- okay that's promising
500: 10/10 <- should be safe

## Debug Build:
\# iterations : How many test failures out of 10 runs
200: 0/10
300: 10/10
400: 10/10
500: 10/10

Release builds need more iterations to repro the bug, so I suspect the faster the hardware / binary, the less likelihood we hit the issue. I assume the runners are slower than my machine so 500 should be hopefully enough. Here is how fast a successful run takes with the loop set to 500 iterations:

Debug: 1.46s
Release: 44ms

Jarred-Sumner: yeah this feels like something that needs many iterations to reproduce consistently

jdalton: Thank you @dl-eric! 500 is < than 2000 so it's a win! A release build should be faster anyways so we don't need to get into toggling iteration values by build type just yet but very good to know.

src/bun.js/api/server.zig
```diff
@@ -1560,13 +1560,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
+        /// If we already wrote a body and are using this function to complete the response, then
+        /// `sendCrlf` should be false.
+        pub fn endWithoutBody(this: *RequestContext, closeConnection: bool, sendCrlf: bool) void {
             if (this.resp) |resp| {
                 if (this.flags.is_waiting_for_request_body) {
                     this.flags.is_waiting_for_request_body = false;
                     resp.clearOnData();
                 }
-                resp.endWithoutBody(closeConnection);
+                resp.endWithoutBody(closeConnection, sendCrlf);
```
jdalton, line 1573: Not for this PR but a good cleanup would be to have a struct so we could have
`resp.endWithoutBody({ .closeConnection = closeConnection, .sendCrlf = sendCrlf })` to avoid the boolean trap.

jdalton (approved):
LG assuming test run passes.

Jarred-Sumner (changes requested):
Thank you for this

While I think the fix is correct, I think we need to enable TCP_CORK to mimic libc behavior for headers/trailers or else it will be an extra network send (which will also make handling the lifetime of this more tricky). I think we should add a method to `HttpResponse.h` which also looks at the status enum in here:

https://github.com/oven-sh/bun/blob/00b1561a9b3f49692ebd560f279438dbd82b2a8b/packages/bun-uws/src/HttpResponseData.h#L70-L74

It would also be great to try this for both SSL and non-SSL versions of the server, just to ensure we don't end up missing the CRLF when we need to keep it

dl-eric:
Hey @Jarred-Sumner, thanks for your review! To clarify, do you mean we should cork the entire http response (headers + sendfile)? Right now I believe we send the headers corked, but the sendfile() bit is unoptimized. I agree that we should have lifetime hooks in HttpResponse/Data.h to dictate corking/uncorking responses such as these, and I'm interested in tackling the issue in a separate PR. Basically set TCP_CORK- send headers, send file - unset TCP_CORK.

However, I see this PR as addressing correctness (without regression in perf) rather than optimization. Do you mind if I do the TCP_CORK logic for my next PR? I will also publish any perf improvements.

Meanwhile, I've added an SSL flavor to the test - although, I noticed when SSL is enabled, we don't actually go through the sendfile() path, so the test isn't really giving us anything (yet). I'd be happy to add `SSL_sendfile` usage though :)

Jarred-Sumner:
Thank you for this, we ended up fixing this and a number of other bugs in 
- https://github.com/oven-sh/bun/pull/12979
- https://github.com/oven-sh/bun/pull/12978

PR #9703, opened 2024-03-29
memoize all calls to selfExePath

nektro:
breaking out #9578 into multiple PRs in order to better pin down where the CI is going awry

### DIFF
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -4852,8 +4852,7 @@ pub const Process = struct {
     }
 
     pub fn getExecPath(globalObject: *JSC.JSGlobalObject) callconv(.C) JSC.JSValue {
-        var buf: bun.PathBuffer = undefined;
-        const out = std.fs.selfExePath(&buf) catch {
+        const out = bun.selfExePath() catch {
             // if for any reason we are unable to get the executable path, we just return argv[0]
             return getArgv0(globalObject);
         };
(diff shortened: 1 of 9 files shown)


src/bun.js/node/types.zig
```diff
@@ -4935,7 +4934,7 @@ pub const Process = struct {
                 bun.String.static("bun"),
             );
         } else {
-            const exe_path = std.fs.selfExePathAlloc(allocator) catch null;
+            const exe_path = bun.selfExePath() catch null;
```
jdalton, line 4937: Above we handle errors with a pretty error for this. Here we `null`. Is the error likely an out of memory thing where we would do the `bun.outOfMemory()`?

nektro: below this line there's an optional check and provide a fallback without needing to throw an error.

Jarred-Sumner: its pretty unlikely for this to error, the current executable would have to have the permissions changed after it started

src/bun.zig
```diff
@@ -2816,6 +2816,22 @@ pub fn linuxKernelVersion() Semver.Version {
...
+        var value: [4096]u8 = undefined;
+        var len: usize = 0;
+    };
+    if (memo.set) return memo.value[0..memo.len :0];
+    const init = try std.fs.selfExePath(&memo.value);
+    memo.len = init.len;
+    memo.value[memo.len] = 0;
+    memo.set = true;
+    return memo.value[0..memo.len :0];
+}
```
gvilums: Are we sure that `selfExePath` will only be called from one thread?

gvilums (commented):
Looks good overall - see my concern about thread safety. If that's not an issue I think we should merge this.

src/bun.zig
Jarred-Sumner: Fixed in c010351c3d46caba53e247a03a7c1f1840c3cee1

Jarred-Sumner:
I believe the napi tests are new & unrelated.

PR #9704, opened 2024-03-29
shell: windows: make EnvMap case-insensitive

nektro:
breaking out https://github.com/oven-sh/bun/pull/9578 into multiple PRs in order to better pin down where my CI is going awry

### DIFF
--- a/src/shell/interpreter.zig
+++ b/src/shell/interpreter.zig
@@ -510,7 +510,7 @@ pub const EnvMap = struct {
         pub fn hash(self: @This(), s: EnvStr) u32 {
             _ = self;
             if (bun.Environment.isWindows) {
-                 return bun.CaseInsensitiveASCIIStringContext.hash(undefined, s.slice());
+                return bun.CaseInsensitiveASCIIStringContext.hash(undefined, s.slice());
             }
             return std.array_hash_map.hashString(s.slice());
         }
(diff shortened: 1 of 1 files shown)


src/shell/interpreter.zig
```diff
@@ -509,11 +509,17 @@ pub const EnvMap = struct {
     const MapType = std.ArrayHashMap(EnvStr, EnvStr, struct {
         pub fn hash(self: @This(), s: EnvStr) u32 {
             _ = self;
+            if (bun.Environment.isWindows) {
```
jdalton, line 512:
```suggestion
            if (comptime bun.Environment.isWindows) {
```
comptimes?

src/shell/interpreter.zig
```diff
@@ -509,11 +509,17 @@ pub const EnvMap = struct {
...
+                return bun.CaseInsensitiveASCIIStringContext.hash(undefined, s.slice());
+            }
             return std.array_hash_map.hashString(s.slice());
         }
         pub fn eql(self: @This(), a: EnvStr, b: EnvStr, b_index: usize) bool {
             _ = self;
             _ = b_index;
+            if (bun.Environment.isWindows) {
+                return bun.CaseInsensitiveASCIIStringContext.eql(undefined, a.slice(), b.slice(), undefined);
+            }
```
jdalton, line 522: I believe the internals of zig do this kind of thing for their env map logic for windows too

src/shell/interpreter.zig
nektro: already is

gvilums (approved):

Jarred-Sumner:
nice

PR #9705, opened 2024-03-29
shell: add 'exit' builtin command

nektro:
breaking out https://github.com/oven-sh/bun/pull/9578 into multiple PRs in order to better pin down where my CI is going awry

edit: progress on https://github.com/oven-sh/bun/issues/9716


src/shell/interpreter.zig
```diff
@@ -4231,36 +4220,10 @@ pub const Interpreter = struct {
...
-                .ls => {
-                    cmd.exec.bltn.impl = .{
-                        .ls = Ls{
-                            .bltn = &cmd.exec.bltn,
-                        },
-                    };
+                inline else => |tag| {
+                    cmd.exec.bltn.impl = @unionInit(RealImpl, @tagName(tag), .{
+                        .bltn = &cmd.exec.bltn,
+                    });
```
gvilums, line 4232: Honestly not sure how I feel about this, it's definitely more concise than what we had previously but also makes it harder to understand what exactly is going on IMO.

src/shell/interpreter.zig
```diff
@@ -8663,6 +8626,84 @@ pub const Interpreter = struct {
...
+                return Maybe(void).success;
+            }
+
+            pub fn next(this: *Exit) void {
+                while (!(this.state == .err or this.state == .done)) {
+                    switch (this.state) {
+                        .waiting_io => return,
+                        .idle, .done, .err => unreachable,
+                    }
+                }
```
gvilums: why is this a loop? the body either returns immediately or causes UB

gvilums (commented):
I assume the failing tests are just general flakiness

src/shell/interpreter.zig
gvilums:
```suggestion
                if(!(this.state == .err or this.state == .done)) {
                    return;
                }
```

src/shell/interpreter.zig
nektro: could you elaborate on this a bit?

src/shell/interpreter.zig
```diff
@@ -8669,6 +8632,84 @@ pub const Interpreter = struct {
...
+                    this.state = .err;
+                    this.next();
+                    return;
+                }
+                this.state = .done;
+                this.next();
+            }
+
+            pub fn deinit(this: *Exit) void {
+                _ = this;
```
Jarred-Sumner, line 8710: is the memory owned by the parent struct?

src/shell/interpreter.zig
Jarred-Sumner: personally, i like it

it also encourages all the commands have the same interface

src/shell/interpreter.zig
nektro: appears so checking all the other builtins

PR #9714, opened 2024-03-29
Fix bug with PipeWriter on Windows

Jarred-Sumner:
### What does this PR do?

This should fix some tests

CI


cirospaciari (approved):

Jarred-Sumner:
Merging because it fixes the console iterator test

PR #9720, opened 2024-03-29
feat!: shell will now throw on error by default

### LINKED ISSUE #8544: Have bun.$ reject its promise if exitCode is not zero

paperclover:
Fixes #8544

Reverts #9592 and fixes the bug it describes instead of documenting undesired behavior.

It is very common to have bun shell used like:

```ts
await $`do something`;
await $`another`;
await $`and so on`;
```

Where developers are not checking for errors, meaning one error within this code block will be effectively ignored.

The way to do this correctly looks like:

```ts
let result = await $`do something`;
if (result.exitCode !== 0) throw new Error("first command failed");
result = await $`another`;
if (result.exitCode !== 0) throw new Error("second command failed");
let result = await $`and so on`;
if (result.exitCode !== 0) throw new Error("final command failed");
```

We do have `$.throws(true)` to make this work better, but making it the default is a better approach.

Also, a second breaking change, the `bunsh` loader is renamed to just `sh`, and the documentation is fixed from saying it is the `bunshell` loader to the `sh` loader. this aligns with the other loaders like `js`, `tsx`, that mirror file extensions and not arbitrary names.

In addition, this also removes the "Bun Shell is an unstable API still under development." note on the blog. I think the API with this change will be ready to call "good", with the further changes being new features that do not break backwards-compatibility.

+ some code tidy.


zackradisic (changes requested):
Looks good, I think we should also update `.throws(bool)` and `.nothrow()` function to note that the shell throws by default

Something like:

```ts
    /**
     * Configure the shell to not throw an exception on non-zero exit codes.
     * 
     * By default, the shell with throw an exception on commands which return non-zero exit codes.
     */
    nothrow(): this;

    /**
     * Configure whether or not the shell should throw an exception on non-zero exit codes.

     * By default, the shell with throw an exception on commands which return non-zero exit codes.
     */
    throws(shouldThrow: boolean): this;
```

Jarred-Sumner:
The tests need to be updated

Issue #8821, opened 2024-02-09
Breaking changes for Bun v1.1


Electroid:
We do plan to make a few, technically breaking changes to Bun v1.1. We will very clearly communicate this as well.

* #7829
* #8544
* #9255
* #2472
* #4892

No longer planned as part of Bun v1.1, needs more work.
* #9267
* #9243

paperclover:
i think we should revert the `__esModule` thing for the runtime as it is a step that Node.js explicitly does not do, and due to that, most of the babel packages load differently in Bun. This breaks the `astro add` cli, among many other cases.

paperclover:
There isn't an issue to track this but I also want to make `$` accept raw strings instead of 'cooked' strings. I think @zackradisic already has this implemented. It reveals some open issues for transpiler bugs but we will fix that after.

Edit: AFAIK this change was made and already shipped.

mGranTY:
add https://github.com/oven-sh/bun/issues/9243 to the list xD

paperclover:
adding #2472 as well because i have basically all the code written for this i think there was just a failing test

birkskyum:
Time might be ripe soon for pulling this off as well - bun has become quite capable in the past half year:
- https://github.com/oven-sh/bun/issues/4464

ptasheq:
Are you kindly planning soon to implement support for custom TLS certificates (NODE_EXTRA_CA_CERTS or an equivalent)?

PR #9722, opened 2024-03-30
Make shell throw by default

### LINKED ISSUE #8544: Have bun.$ reject its promise if exitCode is not zero

zackradisic:
### What does this PR do?

Makes the shell throw on non-zero exit codes by default

See #8821
Fixes #8544

JavaScript/TypeScript modules or builtins changed:

### DIFF
--- a/src/js/builtins/shell.ts
+++ b/src/js/builtins/shell.ts
@@ -250,7 +250,7 @@ export function createBunShellTemplateFunction(ShellInterpreter) {
   class ShellPrototype {
     [cwdSymbol]: string | undefined;
     [envSymbol]: Record<string, string | undefined> | undefined;
-    [throwsSymbol]: boolean = false;
+    [throwsSymbol]: boolean = true;
 
     env(newEnv: Record<string, string | undefined>) {
       if (typeof newEnv === "undefined" || newEnv === originalDefaultEnv) {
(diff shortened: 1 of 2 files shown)


jdalton:
@zackradisic Looks like @paperdave beat ya to the PR: https://github.com/oven-sh/bun/pull/9720 (more is changed in #9729 👀 )

paperclover:
mv test fails because of nothrow. this is already fixed on my branch of this that i opened a bit before yours (see https://github.com/oven-sh/bun/pull/9720). in hindsight i shouldve told you earlier i was doing this, as ive been working on getting all of the 1.1 breakings done.

going to close in favor of mine because there are a bunch of doc changes on mine, could you review them?

PR #9626, opened 2024-03-25
install: inherit stdout/stderr from root lifecycle scripts

dylan-conway:
### What does this PR do?
Changes root lifecycle scripts to print stdout/stderr. Output will print before the install summary.

Screenshot 2024-03-25 at 1 48 03 PM

### How did you verify your code works?
added tests

PR #9724, opened 2024-03-30
chore!: enable 1.1's breaking changes

### LINKED ISSUE #4892: Installing a package with @latest results in "latest" appearing as the version in package.json

Darwin 22.6.0 x86_64 i386

Install a package with `@latest`:

```
$ bun install axios@latest
bun add v1.0.0 (822a00c4)

 installed axios@1.5.0


 1 packages installed [14.00ms]
```

I expect to get the installed version in package.json, just how npm does it. In this case, Axios 1.5.0 was installed, so I expect to have "^1.5.0" as the version:

```jsonc
{
  // ...
  "dependencies": {
    "axios": "^1.5.0"
  }
}
```

### What do you see instead?

```jsonc
{
  // ...
  "dependencies": {
    "axios": "latest"
  }
}
```

### Additional information

This also happens with `--exact`:

```
bun install axios@latest --exact 
```

paperclover:
This removes `breaking_changes_1_1_0` feature flag and then replaces it with `breaking_changes_1_2`, which currently has nothing.

i had realized i also left the bunx fastpath test as off, so i turn that back on, i think we make sure bunx-bins branch is merged and then merged into here, and then double check sure all tests are OK.

Closes #4892


dylan-conway (commented):
LGTM. We should merge <b>before<r> https://github.com/oven-sh/bun/pull/9626

jdalton (approved):

src/feature_flags.zig
```diff
@@ -164,9 +164,9 @@ pub const disable_auto_js_to_ts_in_node_modules = true;
 
 pub const runtime_transpiler_cache = true;
 
-pub const windows_bunx_fast_path = false; // Disabled to simplify fixing the bunx issue
+pub const windows_bunx_fast_path = true; // Disabled to simplify fixing the bunx issue
```
jdalton: Is the `windows_bunx_fast_path` const needed now? The comment said it's disabled to simplify fixing the bunx issue. If it's enabled does that mean the issue is resolved and there is no need for the flag?

paperclover: true. i kind of want to leave it in if anyone in future needs to disable this behavior.

in reality the bunx issue is not related to bunx at all, but rather is a spawn issue

i will change this comment

Issue #9716, opened 2024-03-29
Implement commands in Bun shell


Electroid:
This is a tracking issue for commands that we should implement as builtins in the Bun shell. Some of these are taken from [GNU coreutils](https://en.wikipedia.org/wiki/List_of_GNU_Core_Utilities_commands).

- [x] `cat`
- [x] `touch`
- [x] `mkdir`
- [x] `cd`
- [x] `echo`
- [x] `pwd`
- [x] `which`
- [x] `rm`
- [x] `mv`
- [x] `ls`
- [x] `exit`
- [x] `true`
- [x] `false`
- [x] `cp`
- [ ] `timeout`
- [x] `yes`
- [x] `basename`
- [ ] `realpath`
- [x] `dirname`
- [ ] `mktemp`
- [ ] `truncate`
- [ ] `head`
- [ ] `tail`
- [ ] `cut`
- [ ] `uniq`
- [x] `seq`
- [ ] `sleep`
- [ ] `tee`
- [ ] `date`
- [ ] `sha1sum`
- [ ] `sha224sum`
- [ ] `sha256sum`
- [ ] `sha384sum`
- [ ] `sha512sum`
- [ ] `md5sum`
- [ ] `sort`
- [ ] `printenv`
- [ ] `env`
- [ ] `whoami`
- [ ] `who`
- [ ] `uptime`

This is not an exhaustive list, we can add commands as we think it's a good idea to implement.

While these will already be generally available on most posix systems, having them in the Shell will also allow them to be readily available on Windows when using Bun Shell.

edimarlnx:
I want to create a bun shell to use instead of bash, zsh, sh, etc.
I started the simple [project](https://github.com/edimarlnx/bun-shell/blob/main/bunsh) to think about how it is possible.

So, I have some questions:
1. Is there a way to register a shell command using Javascript definition?

For example:

```js
$.register('add', (n1,n2) => {
  return n1+n2; 
}) 

$`add 2 5`
```

2. Is there a way to persist changes into shell "session"?

For example: 
When I use the command $`cd DIRNAME`, the work directory of the shell ($) changes to the new directory. 

If these examples are possible, It will do a powerful to make a great shell alternative using Javascript.

thewh1teagle:
It will be cool if `Bun` will support cross platform `unzip` for `zip, 7z` etc, and cross platform `wget` or similar which download a file (`Windows` doesn't have `wget` by default) and using fetch for download is slow.

[...]

PR #9728, opened 2024-03-30
shell: implement 'true' and 'false' builtin commands

nektro:
progress on https://github.com/oven-sh/bun/issues/9716

cross-referenced with:
https://git.savannah.gnu.org/cgit/coreutils.git/tree/src/true.c
https://git.savannah.gnu.org/cgit/coreutils.git/tree/src/false.c


paperclover (approved):
winner

Jarred-Sumner:
nice

PR #9730, opened 2024-03-30
ci: windows: use bun install

nektro:


nektro:
manual run: https://github.com/oven-sh/bun/actions/runs/8487999595

dylan-conway (commented):
I'm pretty sure this test failure is new in `bunx.test.ts`. If it's real we should fix this in bun install before merging
Screenshot 2024-03-30 at 12 15 44 AM

dylan-conway (approved):

PR #9740, opened 2024-03-30
shell: implement $0, $1, argv accessors

nektro:
docs: https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameters


src/shell/shell.zig
```diff
@@ -2229,8 +2226,8 @@ pub fn NewLexer(comptime encoding: StringEncoding) type {
 
                 if (i == 0) {
                     switch (char) {
-                        '=', '0'...'9' => return .{ .start = start, .end = self.j },
-                        'a'...'z', 'A'...'Z', '_' => {},
+                        '=' => return .{ .start = start, .end = self.j },
+                        '0'...'9', 'a'...'z', 'A'...'Z', '_' => {},
```
zackradisic: This is actually not consistent with bash because variable names actually can't start with a digit, 

For example if you were to run a bash repl and then do:
```bash
echo $0foo
```
it prints `bashfoo`, rather than looking up a variable

I recommend making a new token variant (`Positional: u8` or something), having this function return that instead if it sees only digits, and then handling that case in expansion

src/shell/interpreter.zig
```diff
@@ -2152,6 +2151,37 @@ pub const Interpreter = struct {
         }
 
         fn expandVar(this: *const Expansion, label: []const u8) EnvStr {
+            if (bun.strings.isAllASCIIDigit(label)) {
+                var int = std.fmt.parseInt(u32, label, 10) catch return EnvStr.initSlice("");
```
zackradisic: It actually seems like bash only allows 1 digit positionals:

```bash
echo $123456
```

Prints: `23456`

zackradisic: It actually seems like bash only allows 1 digit positionals:

```bash
echo $123456
```

Prints: `23456`

src/shell/interpreter.zig
```diff
@@ -2152,6 +2151,37 @@ pub const Interpreter = struct {
...
+                        const vm = js.virtual_machine;
+                        if (vm.main.len > 0) {
+                            if (int == 0) return EnvStr.initSlice(vm.main);
+                            int -= 1;
+                        }
+
+                        if (vm.worker) |worker| {
+                            if (worker.argv) |argv| {
+                                if (int >= argv.len) return EnvStr.initSlice("");
+                                return EnvStr.initSlice(argv[int].utf8Slice());
```
zackradisic: Are we sure `argv[int]` will always be utf8?

zackradisic (changes requested):

src/shell/interpreter.zig
nektro: what platform are u on? I dont experience this

nektro: image

nektro: same as above

src/shell/interpreter.zig
nektro: not sure, I can try some weirder examples. my first guess would be I dont think so because they're not coming JS

src/shell/interpreter.zig
nektro: `sh` does reproduce what u said

src/shell/shell.zig
nektro: I didn't do this originally because I thought bash supported `$10` for higher argv counts

nektro: but will now

src/shell/interpreter.zig
```diff
@@ -2158,6 +2160,38 @@ pub const Interpreter = struct {
...
+                    const vm = js.virtual_machine;
+                    if (vm.main.len > 0) {
+                        if (int == 0) return vm.main;
+                        int -= 1;
+                    }
+
+                    if (vm.worker) |worker| {
+                        if (worker.argv) |argv| {
+                            if (int >= argv.len) return "";
+                            return argv[int].utf8Slice();
```
Jarred-Sumner: This will segfault at runtime 

`utf8Slice` asserts that it is always a utf8 slice. It might not be.

src/shell/interpreter.zig
```diff
@@ -2158,6 +2179,38 @@ pub const Interpreter = struct {
...
+
+                    if (vm.worker) |worker| {
+                        if (worker.argv) |argv| {
+                            if (int >= argv.len) return "";
+                            return this.base.interpreter.getVmArgsUtf8(argv, int);
+                        }
+                    }
+                    const argv = vm.argv;
+                    if (int >= argv.len) return "";
+                    return argv[int];
```
nektro, line 2228: is `this.base.interpreter.getVmArgsUtf8` not needed on this one?

jdalton:
Related https://twitter.com/jfhbrook/status/1774165832249364906

src/shell/interpreter.zig
gvilums: I guess `getVmArgsUtf8` is a bit of a misnomer - it should be called `getVmWorkerArgsUtf8` (because it's specifically used to get `worker.argv`

Jarred-Sumner:
yay

Issue #7541, opened 2023-12-08
Calling `next` on `console`'s asyncIterable causes its stream to become locked


robobun:
I need to read the first line from stdin, perform some logic, and then iterate over the remaining lines of stdin. The suggested way from the docs (https://bun.sh/docs/api/console) to iterate over stdin lines is using its asyncIterator. But there's no suggested method to get a single line.

I attempted to call the asyncIterator method directly:

```ts
let firstLine = (await console[Symbol.asyncIterator]().next()).value;
// perform logic on firstLine
for await (const line of console) {
  // perform logic on remaining lines
}
```

Defining `firstLine` this way works. However it will fail on `const line of console` with the error `ReadableStream is locked`
It seems that manually retrieving an item from console's asyncIterator causes its stream to become locked, and I'm unsure how to unlock it.

I tried using the `prompt` function with an empty string as the message. But it seems `prompt` will still print a single space character to stdout if called with an empty string as the message. So it's not ideal.

The best working solution I have right now is to write a for loop that I break out of after one iteration.

```ts
for await (const line of console) {
  // perform logic on first line
  break;
}
for await (const line of console) {
  // perform logic on remaining lines
}
```

I'd appreciate any suggestions on how to unlock the ReadableStream, or how to more elegantly handle this use case.

_Originally reported on Discord: [`Calling `next` on `console`'s asyncIterable causes its stream to become locked`](https://discord.com/channels/876711213126520882/1182802275224277042/1182802275224277042)_

paperclover:
`ConsoleObject.ts` line 2 defines the async iterator. i touched this code last and probably made some mistakes. i feel like this is "semi" a good first issue but it requires some knowledge on async iterators and readable streams.

for anyone working on this, note that you have to rebuild bun to see changes in this file apply.

guest271314:
You can do something like this.

```
const decoder = new TextDecoder();
process.stdin.on("readable", () => {
  let input = process.stdin.read();
  const lines = decoder.decode(input).split(/\\n|\\r\\n/);
  const firstLine = lines.shift();
  console.log({ firstLine });
  console.log({ lines });
});
```

I wouldn't mix Node.js Streams implementation with WHATWG Streams. Two different implementations.

Keep in mind ECMA-262 does not define STDIO. JavaScript engines and runtimes implement STDIO differently. From the tests I've done no two JavaScript runtimes implement STDIO the same.

guest271314:
[...]

PR #9754, opened 2024-03-31
console-iterator.test.ts: test using Symbol.asyncIterator

aboqasem:
### What does this PR do?

Adds test for #7541.

- [ ] Code changes

```sh
./build/bun-debug test test/js/bun/console/console-iterator.test.ts -t "Symbol.asyncIterator"
```

fails because:

```sh
15 |
16 | async function readInputUsingSymbolAsyncIterator() {
17 |   const items = [];
18 |   const firstLine = (await console[Symbol.asyncIterator]().next()).value;
19 |   items.push(`FIRST${firstLine}`);
20 |   for await (const line of console) {
                         ^
TypeError: ReadableStream is locked
      at initializeReadableStreamDefaultReader (:1:21)
      at getReader (:1:21)
      at doAsyncGeneratorBodyCall (:1:21)
      at asyncGeneratorResumeNext (:1:21)
      at asyncGeneratorEnqueue (:1:21)
      at /Users/user/dev/bun/test/js/bun/console/console-symbol-iterator-run.ts:20:20
      at asyncFunctionResume (:1:21)
      at promiseReactionJobWithoutPromiseUnwrapAsyncContext (:1:21)
```

### DIFF
--- a/test/js/bun/console/console-symbol-iterator-run.ts
+++ b/test/js/bun/console/console-symbol-iterator-run.ts
@@ -1,0 +1,33 @@
+async function readInputUsingForOf() {
+  const items = [];
+  for await (const line of console) {
+    items.push(`FIRST${line}`);
+    break;
+  }
+  for await (const line of console) {
+    if (line == "break") {
+      break;
+    }
+    items.push(line);
+  }
+  return items;
+}
+
+async function readInputUsingSymbolAsyncIterator() {
+  const items = [];
+  const firstLine = (await console[Symbol.asyncIterator]().next()).value;
+  items.push(`FIRST${firstLine}`);
+  for await (const line of console) {
+    if (line == "break") {
+      break;
+    }
+    items.push(line);
+  }
+  return items;
+}
+
+const a = await readInputUsingForOf();
+console.write(JSON.stringify(a));
+
+const b = await readInputUsingSymbolAsyncIterator();
+console.write(JSON.stringify(b));
--- a/test/js/bun/console/console-iterator.test.ts
+++ b/test/js/bun/console/console-iterator.test.ts
@@ -74,3 +74,19 @@ it("can use the console iterator more than once", async () => {
   expect(await new Response(stdout).text()).toBe('["hello","world"]["another"]');
   proc.kill(0);
 });
+
+// https://github.com/oven-sh/bun/issues/7541
+it("can use console[Symbol.asyncIterator]", async () => {
+  const proc = spawn({
+    cmd: [bunExe(), import.meta.dir + "/" + "console-symbol-iterator-run.ts"],
+    stdin: "pipe",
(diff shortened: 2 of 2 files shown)


gvilums (approved):

test/js/bun/console/console-symbol-iterator-run.ts
```diff
@@ -0,0 +1,33 @@
...
+      break;
+    }
+    items.push(line);
+  }
+  return items;
+}
+
+async function readInputUsingSymbolAsyncIterator() {
+  const items = [];
+  const firstLine = (await console[Symbol.asyncIterator]().next()).value;
```
aboqasem, line 18: btw is this actually an expected usage? I am debugging and I saw that the generator will not be returning so the reader will never be released. So I figured we might first decide whether it's expected to work.

with this diff:

```diff
diff --git a/test/js/bun/console/console-symbol-iterator-run.ts b/test/js/bun/console/console-symbol-iterator-run.ts
index f1c5efbfa..4392831ec 100644
--- a/test/js/bun/console/console-symbol-iterator-run.ts
+++ b/test/js/bun/console/console-symbol-iterator-run.ts
@@ -15,7 +15,9 @@ async function readInputUsingForOf() {
 
 async function readInputUsingSymbolAsyncIterator() {
   const items = [];
-  const firstLine = (await console[Symbol.asyncIterator]().next()).value;
+  const it = console[Symbol.asyncIterator]();
+  const firstLine = (await it.next()).value;
+  await it.return?.();
   items.push(`FIRST${firstLine}`);
   for await (const line of console) {
     if (line == "break") {
```

It will be working as expected (because the `finally` that releases the lock will be executed):

https://github.com/oven-sh/bun/blob/4bbcc39d2f46d0fd9873732f25e3cedd460a408a/src/js/builtins/ConsoleObject.ts#L103-L109

should it work if the user never releases the generator? could be something like:

```diff
diff --git a/src/js/builtins/ConsoleObject.ts b/src/js/builtins/ConsoleObject.ts
index 1e35d8de7..0c417aa75 100644
--- a/src/js/builtins/ConsoleObject.ts
+++ b/src/js/builtins/ConsoleObject.ts
@@ -12,9 +12,14 @@ export function asyncIterator(this: Console) {
   var value: Uint8Array[];
   var value_len: number;
   var pendingChunk: Uint8Array | undefined;
+  var reader: ReadableStreamDefaultReader<Uint8Array>;
 
   async function* ConsoleAsyncIterator() {
-    var reader = stream.getReader();
+    if (reader) {
+      reader.releaseLock();
+    }
+
+    reader = stream.getReader();
     var deferredError;
     try {
       if (i !== -1) {

```

aboqasem: should it work if the user never releases the generator?

PR #9761, opened 2024-03-31
fix: `which` should use `cwd` if given a relative filepath

paperclover:
### What does this PR do?

This corrects a logic mistake in my previous PR to fix `Bun.which`

Doing `which node_modules/.bin/next` should result in the relative path join of cwd and the path, which happens specifically because there are slashes in the path. A path without slashes is treated as a command name, and that is looked up in PATH.

You can verify this new logic is correct by looking at the `which` command's output.


src/bun.js/api/bun/subprocess.zig
```diff
@@ -1642,7 +1642,7 @@ pub const Subprocess = struct {
 
                     if (argv0 == null) {
                         var path_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
-                        const resolved = Which.which(&path_buf, PATH, arg0.slice()) orelse {
+                        const resolved = Which.which(&path_buf, PATH, cwd, arg0.slice()) orelse {
```
dylan-conway, line 1645: should this `cwd` be the one on line 1709?

src/which.zig
```diff
@@ -133,7 +140,20 @@ fn whichWin(buf: *bun.WPathBuffer, path: []const u8, bin: []const u8) ?[:0]const
         return searchBin(buf, bin_utf16.len, check_windows_extensions);
     }
 
-    if (path.len == 0) return null;
+    // check if bin is in cwd
+    if (bun.strings.containsChar(bin, '/')) {
```
dylan-conway: I think these should be backslashes because this is the windows version. Or maybe we should accept both?

src/which.zig
```diff
@@ -133,7 +140,20 @@ fn whichWin(buf: *bun.WPathBuffer, path: []const u8, bin: []const u8) ?[:0]const
...
+        if (searchBinInPath(
+            buf,
+            &path_buf,
+            cwd,
+            bun.strings.withoutPrefixComptime(bin, "./"),
```
dylan-conway, line 149: same here

src/which.zig
```diff
@@ -145,3 +165,14 @@ fn whichWin(buf: *bun.WPathBuffer, path: []const u8, bin: []const u8) ?[:0]const
 
     return null;
 }
+
+test "which" {
```
dylan-conway, line 170: I'm excited to get zig tests working

dylan-conway (commented):
A couple comments

src/which.zig
paperclover: it should be both

src/bun.js/api/bun/subprocess.zig
paperclover: this was from reverting https://github.com/oven-sh/bun/pull/9691. i'll look if this is correct code.

src/which.zig
paperclover: this was from reverting https://github.com/oven-sh/bun/pull/9691. i'll look if this is correct code.

src/which.zig
paperclover: this was from reverting https://github.com/oven-sh/bun/pull/9691. i don't think we'll get the zigtests working

Jarred-Sumner:
which test is failing on windows

PR #9762, opened 2024-03-31
add `bun exec`

zackradisic:
### What does this PR do?

Adds `bun exec` command
- [x] Code changes

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -97,6 +97,7 @@ pub const ShellCompletions = @import("./cli/shell_completions.zig");
 pub const UpdateCommand = @import("./cli/update_command.zig").UpdateCommand;
 pub const UpgradeCommand = @import("./cli/upgrade_command.zig").UpgradeCommand;
 pub const BunxCommand = @import("./cli/bunx_command.zig").BunxCommand;
+pub const ExecCommand = @import("./cli/exec_command.zig").ExecCommand;
 
 pub const Arguments = struct {
     pub fn loader_resolver(in: string) !Api.Loader {
@@ -966,6 +967,7 @@ pub const HelpCommand = struct {
         \\  <b><magenta>test<r>                           Run unit tests with Bun
         \\  <b><magenta>x<r>         <d>{s:<16}<r>     Execute a package binary (CLI), installing if needed <d>(bunx)<r>
         \\  <b><magenta>repl<r>                           Start a REPL session with Bun
+        \\  <b><magenta>exec<r>                           Run a shell script directly with Bun
         \\
         \\  <b><blue>install<r>                        Install dependencies for a package.json <d>(bun i)<r>
         \\  <b><blue>add<r>       <d>{s:<16}<r>     Add a dependency to package.json <d>(bun a)<r>
@@ -1306,6 +1308,8 @@ pub const Command = struct {
 
             RootCommandMatcher.case("run") => .RunCommand,
             RootCommandMatcher.case("help") => .HelpCommand,
+
+            RootCommandMatcher.case("exec") => .ExecCommand,
 
             // These are reserved for future use by Bun, so that someone
             // doing `bun deploy` to run a script doesn't accidentally break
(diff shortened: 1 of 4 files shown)


src/cli/exec_command.zig
```diff
@@ -0,0 +1,42 @@
...
+const MutableString = bun.MutableString;
+const stringZ = bun.stringZ;
+const default_allocator = bun.default_allocator;
+const C = bun.C;
+const std = @import("std");
+const open = @import("../open.zig");
+const Command = bun.CLI.Command;
+
+pub const ExecCommand = struct {
+    const discord_url = "https://bun.sh/discord";
```
Jarred-Sumner:
```suggestion

```

src/cli/exec_command.zig
```diff
@@ -0,0 +1,42 @@
...
+        // this is a hack: make dummy bundler so we can use its `.runEnvLoader()` function to populate environment variables probably should split out the functionality
+        var bundle = try bun.Bundler.init(
+            ctx.allocator,
+            ctx.log,
+            try @import("../bun.js/config.zig").configureTransformOptionsForBunVM(ctx.allocator, ctx.args),
+            null,
+        );
+        try bundle.runEnvLoader(false);
+        const mini = bun.JSC.MiniEventLoop.initGlobal(bundle.env);
+        const code = bun.shell.Interpreter.initAndRunFromSource(mini, "<script>", script) catch |err| {
```
Jarred-Sumner: Instead of `<script>`, it should be the absolute file path of the current working directory with `/[eval]` appended at the end, to match `bun -e`

test/js/bun/shell/exec.test.ts
```diff
@@ -0,0 +1,25 @@
+import * as t from "bun:test";
+import { TestBuilder } from "./test_builder";
+import { bunEnv } from "harness";
+
+const BUN = process.argv0;
+
+t.describe("bun exec", () => {
```
Jarred-Sumner: we don't use `t` anywhere else in the codebase, lets avoid adding this pattern

```suggestion
describe("bun exec", () => {
```

test/js/bun/shell/exec.test.ts
```diff
@@ -0,0 +1,25 @@
+import * as t from "bun:test";
```
Jarred-Sumner:
```suggestion
import {describe} from "bun:test";
```

test/js/bun/shell/exec.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TestBuilder } from "./test_builder";
+import { bunEnv } from "harness";
+
+const BUN = process.argv0;
+
+t.describe("bun exec", () => {
+  TestBuilder.command`${BUN} exec ${"echo hi!"}`.env(bunEnv).stdout("hi!\n").runAsTest("it works");
```
Jarred-Sumner, line 8: Where is the test checking that a script exiting with a non-zero exit code is propagated to the parent process? We should also have a test which prints a large amount of output to stdout or stderr.

Jarred-Sumner:
There is 1 test failure in exec that needs to be addressed and then let's merge this

src/cli.zig
```diff
@@ -2170,6 +2182,21 @@ pub const Command = struct {
...
+                Command.Tag.ExecCommand => {
+                    Output.pretty(
+                        \\<b>Usage: bun exec <r><cyan>\<script\><r>
+                        \\
+                        \\Execute a shell script directly from Bun.
+                        \\
+                        \\<b><red>Note<r>: If executing this from a shell, make sure to escape the string!
+                        \\
+                        \\<b>Examples<d>:<r>
+                        \\  <b>bunx exec "echo hi"<r>
```
Electroid, line 2194: `bunx exec` or `bun exec`?

nektro: should be `bun exec`

Jarred-Sumner:
thank you

PR #9763, opened 2024-04-01
windows: fix `bun pm bin -g` path not added complaining

chawyehsu:
### What does this PR do?

Before this patch `bun pm bin -g` keeps complaining about path not added on Windows.


dylan-conway:
Thank you!

PR #9772, opened 2024-04-01
shell: fix crash in 'ls' and other misc improvements

nektro:


src/shell/interpreter.zig
```diff
@@ -6792,15 +6792,14 @@ pub const Interpreter = struct {
             pub fn start(this: *Pwd) Maybe(void) {
                 const args = this.bltn.argsSlice();
                 if (args.len > 0) {
-                    const msg = "pwd: too many arguments";
+                    const msg = "pwd: too many arguments\n";
```
jdalton, line 6795: This is almost like how console log will add a \n. Is there an equiv lower level one to just auto tack that on (or an opportunity for one if we've reached for that a lot)?

paperclover: yeah, we should fix this later on

Issue #8116, opened 2024-01-12
add winget to install docs


ocweai:
### What is the problem this feature would solve?

- > Package manager

- [brew](https://brew.sh/)
- [winget](https://learn.microsoft.com/zh-cn/windows/package-manager/)

### What is the feature you are proposing to solve the problem?

目前通过 brew 安装的 macos 的 bun 命令行需要手动设置全局变量

### What alternatives have you considered?

crul ...  支持自动设置全平台的命令行全局变量

staabm:
homebrew is already supported, see https://github.com/oven-sh/homebrew-bun

ocweai:
> homebrew is already supported, see https://github.com/oven-sh/homebrew-bun

homebrew 
Global variables bun or bunx will not be automatically set after installation
Causes bun to be undefined after closing the current command line program

YpsilonTM:
I would like to see this happen to. Since winget is getting more and more popular. And I really like it.

sitiom:
- https://github.com/microsoft/winget-pkgs/pull/147222
- https://github.com/microsoft/winget-pkgs/pull/147226
- https://github.com/oven-sh/bun/pull/9770

denelon:
@sitiom after the PRs are merged and updates are automated, it might make sense to update the docs for installing the package over at https://bun.sh/blog/bun-v1.1#getting-started should be updated from:

`> powershell -c "irm bun.sh/install.ps1 | iex"`

to

`winget install Oven-sh.Bun` **or** `winget install Oven-sh.Bun` (depending on processor instruction support).

sitiom:
Yes, of course 🙂 (Can't edit the blog post though):
- https://github.com/oven-sh/bun/pull/9806

CoolPlayLin:
The 1.1.0 full version supports were in
- microsoft/winget-pkgs#147992
- microsoft/winget-pkgs#147991
- microsoft/winget-pkgs#147226
- microsoft/winget-pkgs#147222

sitiom:
Blocker: https://github.com/oven-sh/bun/pull/9806#issuecomment-2033266871

> `# Or if your CPU does not support the AVX2 instruction set. You can check by running this command:`
> 
> This should not be manual. I will look into how this can be solved for `winget` later. If it is not solvable, I don't think we should recommend `winget`'s usage for installing bun.

albertocavalcante:
I believe it would be highly beneficial to have this documented.

[...]

PR #9806, opened 2024-04-01
docs: add winget installation method

matifali:
### What does this PR do?

After https://github.com/microsoft/winget-pkgs/pull/147222, `bun` can be installed via [`winget`](https://learn.microsoft.com/en-us/windows/package-manager/winget/) package manager.


matifali:
As a follow up submitting new releases should be automated with a GitHub action.

docs/installation.md
```diff
@@ -50,6 +50,10 @@ To install, paste this into a terminal:
 powershell -c "irm bun.sh/install.ps1|iex"
 ```
 
+```powershell#winget
+winget install --id Oven-sh.Bun
```
sitiom:
```suggestion
winget install Oven-sh.Bun
# Or if your CPU does not support the AVX2 instruction set. You can check by running this command:
# (Add-Type -MemberDefinition '[DllImport("kernel32.dll")] public static extern bool IsProcessorFeaturePresent(int ProcessorFeature);'-Name 'Kernel32'-Namespace 'Win32'-PassThru)::IsProcessorFeaturePresent(40)
winget install Oven-sh.Bun.Baseline 
```

paperclover:
Before this is added, i want to solve this.

`# Or if your CPU does not support the AVX2 instruction set. You can check by running this command:`

This should not be manual. I will look into how this can be solved for `winget` later. If it is not solvable, I don't think we should recommend `winget`'s usage for installing bun.

sitiom:
> Before this is added, i want to solve this.
> 
> `# Or if your CPU does not support the AVX2 instruction set. You can check by running this command:`
> 
> This should not be manual. I will look into how this can be solved for `winget` later. If it is not solvable, I don't think we should recommend `winget`'s usage for installing bun.

If you made an .exe bootstrapper/installer which handles all the necessary checks (winget does not support `ps1` scripts yet), winget can use that instead. For now, theres a small note post-installation about trying the baseline package if the program doesn't work.

albertocavalcante:
Hi @sitiom, I'm hoping to get this merged, but ack on the issue statement. I've added a comment with a suggestion at https://github.com/oven-sh/bun/issues/8116#issuecomment-2384294695, can you please check it? Thanks!

PR #7882, opened 2023-12-28
Add Scoop installation reference for Windows

sitiom:
[`bun-canary`](https://scoop.sh/#/apps?q=bun-canary&id=94a4ad4b189dee5b4f62aaf863fd2178827f3791) is now available on Scoop via the Versions bucket.

### What does this PR do?

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #9818, opened 2024-04-02
docs: add scoop installation method

sitiom:
### What does this PR do?

Continuation of https://github.com/oven-sh/bun/pull/7882

- [ ] Code changes


paperclover:
As i posted in the #9806 PR

> Before this is added, i want to solve this.
> `# Or if your CPU does not support the AVX2 instruction set. You can check by running this command:`
> 
> This should not be manual. I will look into how this can be solved for `winget` later. If it is not solvable, I don't think we should recommend `winget`'s usage for installing bun.

I'll take a look at how scoop works later and see what can be done.

sitiom:
As mentioned in the discord server a while ago, the check has been already been added to Scoop. But I thought might as well let the users know how to check it themselves 🤔

paperclover:
Is it possible to make so the pre_install script just downloads the correct one?

Jarred-Sumner:
@sitiom thank you for this, but we can't in good faith add this to our docs as a recommended installation method until the `baseline` build detection is automatic

Any ideas how we can make `scoop install bun` choose the baseline build or regular build automatically, without asking users to make a decision and handle a confusing error?

sitiom:
I think the suggestion `Please install 'bun-baseline' instead` is clear. Although if you insist, it's possible for Scoop to download both and then do the check (although that means more download bandwidth), would that be fine?

CC @chawyehsu

Jarred-Sumner:
> I think the suggestion Please install 'bun-baseline' instead is clear.

It's clear to us, but we can't expect that from people trying Bun for the first time

> it's possible for Scoop to download both and then do the check (although that means more download bandwidth), would that be fine?

I think that is a better tradeoff, one option would be to have 3 packages: `bun`, `bun-windows-x64`, `bun-windows-x64-baseline` and then the other 2 are dependencies of `bun` and then a small script would run to detect if they support AVX2 and if so, use the x64 build and otherwise fallback to use the x64-baseline build

chawyehsu:
> I think the suggestion `Please install 'bun-baseline' instead` is clear. Although if you insist, it's possible for Scoop to download both and then do the check (although that means more download bandwidth), would that be fine?
> 
> 
> 
> CC @chawyehsu

I agree with you. And I personally don't like the idea of downloading additional zip on-the-fly via the pre_install script （which will take much unwanted work like validating checksums in an unusual way).

If you think the way Scoop currently provides is not good enough, I'd suggest you bundling bun and bun-baseline into one single zip so we can just download that single zip (or if it's not appropriate then we can download both of the two zips simultaneously) and expose the proper exec for our users based on detected microarch. This takes 2x bandwidth but is succincter and some other packages also do this way (in the way of bundling 32bit and 64bit).

sitiom:
Here you go:
- https://github.com/ScoopInstaller/Main/pull/5680

paperclover:
legendary. will merge this after it lands in scoop.

sitiom:
@paperdave Now merged 🙂

paperclover:
incredible work

Jarred-Sumner:
Thank you

PR #9832, opened 2024-04-02
Truncate source lines in error messages

Jarred-Sumner:
### What does this PR do?

When a minified file throws an exception in Bun, the source lines can be extremely long

This causes two problems:
1) The divot line has a comically large amount of indentation
2) It fills up your terminal with unreadable code

This PR truncates source text to be up to 1024 bytes per line. The divot will no longer appear if the line is > 512 bytes.  The goal is to avoid the above two problems without debugging harder. So we use a probably longer than necessary limit.

These numbers are arbitrary and a better solution overall would be to "scroll" within the lines to the relevant parts. I don't trust our source maps are reliable enough right now to do that well, so this is a compromise until then.

#### After:
image

#### Before:

image

With lots of scrolling:

image

There are a couple snapshot tests


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.22)
 cmake_policy(SET CMP0091 NEW)
 cmake_policy(SET CMP0067 NEW)
 
-set(Bun_VERSION "1.1.0")
+set(Bun_VERSION "1.1.1")
```
jdalton, line 5: I've seen a few diff v1_1_1 things or was that a 1_1_0 thing. This should be in a sep PR though as it's versioning related.

src/bun.js/javascript.zig
```diff
@@ -2864,6 +2864,11 @@ pub const VirtualMachine = struct {
             defer printGithubAnnotation(exception);
         }
 
+        // This is a longer number than necessary because we don't handle this case very well
+        // At the very least, we shouldn't dump 100 KB of minified code into your terminal.
+        const max_line_length_with_divot = 512;
+        const max_line_length = 1024;
```
jdalton, line 2870: I'm fine with less. Usually a line is like 80 to 120 chars.

src/bun.js/javascript.zig
```diff
@@ -2880,13 +2885,27 @@ pub const VirtualMachine = struct {
...
             try writer.writeByteNTimes(' ', pad);
 
-            try writer.print(
-                comptime Output.prettyFmt("<r><d>{d} | <r>{}\n", allow_ansi_color),
-                .{
-                    display_line,
-                    bun.fmt.fmtJavaScript(std.mem.trimRight(u8, std.mem.trim(u8, source.text.slice(), "\n"), "\t "), allow_ansi_color),
-                },
-            );
+            const trimmed = std.mem.trimRight(u8, std.mem.trim(u8, source.text.slice(), "\n"), "\t ");
```
jdalton, line 2888: The snip to `\n` covers `\r\n` too (is my assumption)

src/bun.js/javascript.zig
```diff
@@ -2910,17 +2929,28 @@ pub const VirtualMachine = struct {
...
-                        allow_ansi_color,
-                    ),
-                    .{
-                        bun.fmt.fmtJavaScript(text, allow_ansi_color),
-                    },
-                );
+                const text = trimmed[0..@min(trimmed.len, max_line_length)];
+
+                if (text.len != trimmed.len) {
+                    const fmt = if (comptime allow_ansi_color) "<r><d> | ... truncated <r>\n" else "\n";
```
jdalton, line 2937: Can has ellipsis here? “…” (U+2026). (if it's okay for Windows)

src/bun.js/javascript.zig
```diff
@@ -2931,24 +2961,40 @@ pub const VirtualMachine = struct {
                 try writer.writeByteNTimes(' ', pad);
                 defer source.text.deinit();
                 const text = source.text.slice();
-                const remainder = std.mem.trimRight(u8, std.mem.trim(u8, text, "\n"), "\t ");
+                const trimmed = std.mem.trimRight(u8, std.mem.trim(u8, text, "\n"), "\t ");
```
jdalton, line 2964: This timming is done twice (turn into an inlined function?)

src/bun.js/javascript.zig
```diff
@@ -2931,24 +2961,40 @@ pub const VirtualMachine = struct {
...
-                    ),
-                    .{ display_line, bun.fmt.fmtJavaScript(remainder, allow_ansi_color) },
-                );
+                // TODO: preserve the divot position and possibly use stringWidth() to figure out where to put the divot
+                const clamped = trimmed[0..@min(trimmed.len, max_line_length)];
 
-                if (!top.position.isInvalid()) {
-                    const indent = max_line_number_pad + " | ".len + @as(u64, @intCast(top.position.column_start));
+                if (clamped.len != trimmed.len) {
+                    const fmt = if (comptime allow_ansi_color) "<r><d> | ... truncated <r>\n\n" else "\n\n";
```
jdalton, line 2970: ellipsis?

test/js/bun/util/inspect-error-fixture.min.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * @license React
+ * react.production.min.js
+ *
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+'use strict';var l=Symbol.for("react.element"),n=Symbol.for("react.portal"),p=Symbol.for("react.fragment"),q=Symbol.for("react.strict_mode"),r=Symbol.for("react.profiler"),t=Symbol.for("react.provider"),u=Symbol.for("react.context"),v=Symbol.for("react.forward_ref"),w=Symbol.for("react.suspense"),x=Symbol.for("react.memo"),y=Symbol.for("react.lazy"),z=Symbol.iterator;function A(a){if(null===a||"object"!==typeof a)return null;a=z&&a[z]||a["@@iterator"];return"function"===typeof a?a:null}
```
jdalton, line 10: Given the xz minified/obfuscated payload test stuff in the headlines for ease of future us and transparency could we just minify our own test file via our API.

test/js/bun/util/inspect-error.test.js
```diff
@@ -31,3 +31,31 @@ test("BuildMessage", async () => {
...
+        .replaceAll("\\", "/")
+        .trim(),
+    ).toMatchSnapshot();
+  }
+});
+
+test("Error inside minified file (color) ", () => {
+  try {
+    require("./inspect-error-fixture.min.js");
+    expect.unreachable();
```
jdalton, line 80: Missing documentation for `expect.unreachable()`. (would be nice to know which API we are mimicking (e.g. jest, vitest, etc).
https://github.com/oven-sh/bun/pull/7131

jdalton:
I love these kinds of enhancements ❤️

PR #9901, opened 2024-04-03
Remove documentation references to environment variable inlining because the bundler does not do so

mangs:
### What does this PR do?

In testing our Bun-based lambda while prepping its rollout to production, I realized that despite the documentation saying that environment variable inlining occurs when building, the bundler actually doesn't. So I'm proposing this be removed from the documentation if and until the behavior changes.

I'm creating a follow-up issue to propose the ability to make environment variable inlining configurable. I'll reference this issue in that feature request.

- [ ] Code changes

The code didn't change


paperclover:
thank you

PR #9908, opened 2024-04-04
shell: more builtin commands

nektro:
https://github.com/oven-sh/bun/issues/9716 progress

added yes, seq, dirname, basename.


src/shell/interpreter.zig
```diff
@@ -5174,27 +5157,25 @@ pub const Interpreter = struct {
         }
 
         /// **WARNING** You should make sure that stdout/stderr does not need IO (e.g. `.needsIO(.stderr)` is false before caling `.writeNoIO(.stderr, buf)`)
-        pub fn writeNoIO(this: *Builtin, comptime io_kind: @Type(.EnumLiteral), buf: []const u8) usize {
+        pub fn writeNoIO(this: *Builtin, comptime io_kind: @Type(.EnumLiteral), buf: []const u8) Maybe(usize) {
```
zackradisic, line 5444: Unfortunately, because the previous value was a `usize`, almost every call to (about 34 of them) `writeNoIO()` discarded the value (e.g. `_ = writeNoIO()`)

I think we should revert this change and do it in another PR as it will require a lot of changes to handle the error everywhere

src/shell/interpreter.zig
```diff
@@ -9456,6 +9437,192 @@ pub const Interpreter = struct {
...
+                        this.bltn.done(1);
+                        return Maybe(void).success;
+                    }
+
+                    res = this.print("\n");
+                    if (res == .err) {
+                        this.bltn.done(1);
+                        return Maybe(void).success;
+                    }
+                }
```
zackradisic: I'm pretty sure this will block the main thread forever

Example to demonstrate what I mean
```typescript
import { $ } from 'bun'
await Promise.all([$`yes hello`, $`cat src/js_parser.zig`]);
```

It gets stuck in the while loop and nothing from the `cat` command appears

There's two cases to handle:
1. if stdout/stderr needs IO: enqueue the data and suspend execution
2. if stdout/stderr does not need IO: could still get stuck in an infinite loop, so should probably suspend execution after writing 128kb or something

src/shell/interpreter.zig
```diff
@@ -9456,6 +9437,192 @@ pub const Interpreter = struct {
...
+                    this.bltn.stdout.enqueue(this, msg);
+                    return Maybe(usize).success;
+                }
+                return this.bltn.writeNoIO(.stdout, msg);
+            }
+
+            pub fn onIOWriterChunk(this: *@This(), _: usize, maybe_e: ?JSC.SystemError) void {
+                _ = this;
+                _ = maybe_e;
+                unreachable;
```
zackradisic: This is definitely reachable

src/shell/interpreter.zig
```diff
@@ -9456,6 +9437,192 @@ pub const Interpreter = struct {
...
+            }
+
+            fn do(this: *@This()) Maybe(void) {
+                var current = this.start;
+                var arena = std.heap.ArenaAllocator.init(bun.default_allocator);
+                defer arena.deinit();
+
+                while (if (this.increment > 0) current <= this.end else current >= this.end) : (current += this.increment) {
+                    const str = std.fmt.allocPrint(arena.allocator(), "{d}", .{current}) catch bun.outOfMemory();
+                    defer _ = arena.reset(.retain_capacity);
```
zackradisic, line 9901: What about printing to `this.buf.writer()` instead?

src/shell/interpreter.zig
```diff
@@ -9456,6 +9437,192 @@ pub const Interpreter = struct {
...
+                }
+                return Maybe(void).success;
+            }
+
+            fn print(this: *@This(), msg: string) Maybe(void) {
+                if (this.bltn.stdout.needsIO()) {
+                    this.buf.appendSlice(bun.default_allocator, msg) catch bun.outOfMemory();
+                    return Maybe(void).success;
+                }
+                const res = this.bltn.writeNoIO(.stdout, msg);
```
zackradisic, line 9919: Suffers from same problem as `Yes`, example:

```typescript
await Promise.all([$`seq 9999`, $`cat src/js_parser.zig`]);
```

zackradisic (changes requested):

src/shell/interpreter.zig
nektro: I think its okay if most ignore it, but in `yes` it was imperative so that it knows when to stop

src/shell/interpreter.zig
nektro: this may be a latent bug from elsewhere since `bun-debug exec 'seq 9999'` exits near instantly as do `bun exec 'cat file'`

src/shell/interpreter.zig
nektro: `buf` is only used if `this.bltn.stdout.needsIO()`

src/shell/interpreter.zig
```diff
@@ -4933,25 +4938,8 @@ pub const Interpreter = struct {
...
-                    .echo => "echo",
-                    .pwd => "pwd",
-                    .which => "which",
-                    .rm => "rm",
-                    .mv => "mv",
-                    .ls => "ls",
-                    .exit => "exit",
-                    .true => "true",
-                    .false => "false",
+                    .yes => "yes [expletive]\n",
```
Electroid: heh, is that what it actually says? lol

test/js/bun/shell/commands/yes.test.ts
```diff
@@ -0,0 +1,24 @@
+import { $, ShellOutput } from "bun";
+import { describe, test, expect, beforeEach } from "bun:test";
+
+$.throws(false);
+
+describe("yes", async () => {
```
Electroid, line 6: what a weird command, it just keeps printing the same value over and over again?

Jarred-Sumner: yes

src/shell/interpreter.zig
nektro: indeed 😄 

image

ghiscoding:
it might also be a good idea to update the website ([built-in commands](https://bun.sh/docs/runtime/shell#builtin-commands))  I see only 6 commands as implemented on the website but in reality it's closer to 15 with this PR

PR #9941, opened 2024-04-04
fix typo: missing "v" prefix on installed Bun version (bun upgrade)

Bellisario:
### What does this PR do?

This pull request adds the missing "v" prefix on the installed Bun version on `bun upgrade`.

image1


paperclover:
thanks

PR #9957, opened 2024-04-04
Reference .exists() in File-IO Docs

jdfwarrior:
### What does this PR do?

The docs don't explicitly mention that a Bunfile has an .exists() method on it to check for file existence and it's missing in the type definition as well. The only mention of it is in the Guides area. This just adds it to the type definition and shows an example of it in the docs.


gvilums:
Thank you for the PR!

PR #9958, opened 2024-04-04
Update Window install instructions

IDisposable:
Now that `winget` is so common, might as well give that as an example as it enables later `winget upgrade` to just work.

### What does this PR do?

Updates the install instruction for Windows platform to show an alternative of using [`winget`](https://github.com/microsoft/winget-cli) to install the `Open-sh.Bun` package you're so kindly providing.

Note that I retained the original PowerShell example (and added appropriate comment)

- [ ] Code changes

Tested on my machine using 
```sh
winget install Oven-sh.bun
```

image


paperclover:
Duplicate of #9806, which has unresolved issues.

Issue #9966, opened 2024-04-05
bunx --bun vite stopped working


Garruxx:
### What version of Bun is running?

1.1.1

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

ejecutar bunx --bun vite

### What is the expected behavior?

  VITE v5.2.8  ready in 1038 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help

### What do you see instead?

![image](https://github.com/oven-sh/bun/assets/109706264/18bda550-901c-4463-83d2-fa97081933b0)
```sh
44331 |         sysPath$2.resolve(path, evPath), KEY_LISTENERS, sysPath$2.join(path, evPath)
44332 |       );
44333 |     }
44334 |   };
44335 |   try {
44336 |     return fs$7.watch(path, options, handleEvent);
                   ^
EBUSY: EBUSY watching "C:/Users/sanch/dev/Web/react/xoukan"
 syscall: "watch"
   path: "C:/Users/sanch/dev/Web/react/xoukan"

      at new FSWatcher (node:fs:105:27)
      at watch2 (node:fs:284:199)
      at createFsWatchInstance (C:\Users\sanch\dev\Web\react\xoukan\node_modules\vite\dist\node\chunks\dep-whKeNLxG.js:44336:12)
      at setFsWatchListener (C:\Users\sanch\dev\Web\react\xoukan\node_modules\vite\dist\node\chunks\dep-whKeNLxG.js:44383:15)
      at _watchWithNodeFs (C:\Users\sanch\dev\Web\react\xoukan\node_modules\vite\dist\node\chunks\dep-whKeNLxG.js:44538:14)
      at C:\Users\sanch\dev\Web\react\xoukan\node_modules\vite\dist\node\chunks\dep-whKeNLxG.js:44774:14
error: script "dev" exited with code 1
```

### Additional information

This worked fine in bun 1.1.0

Shenlok:
I'm also seeing this when trying to `bun run --bun dev` on a fresh NextJS (14.1.4) app scaffolded with `create-next-app` (using bun canary `1.1.1+ca1dbb4eb`)

danielj247:
I get the same issue with `bunx --bun vite` - only on Windows machines.

MacOS is behaving as expected.

Jarred-Sumner:
[...]

Issue #9943, opened 2024-04-04
🔒 Saving lockfile... error: failed to open lockfile: EBUSY


JiriMachynka:
### What version of Bun is running?

1.1.1+ca1dbb4eb

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

`bun i` inside a project

### What is the expected behavior?

Bun should install dependencies correctly and properly update its lockfile.

### What do you see instead?

![bun-lockfile-error](https://github.com/oven-sh/bun/assets/65855539/f6c5adfe-2416-45a9-b15e-587f82f57b74)

consp1racy:
also `bun pm migrate`

Jarred-Sumner:
Bun v1.1.2 will ship in less than 12 hours and will fix this

Jarred-Sumner:
The fix is merged

In about an hour, feel free to run `bun upgrade --canary` and give the fix a try

SKumarSpace:
Hello, I ran the `bun upgrade --canary` command and it panicked. 

```sh
$ bun upgrade --canary

uh-oh: attempt to unwrap error
bun will crash now 😭😭😭

----- bun meta -----
Bun v1.1.1 (ca1dbb4e) Windows x64
UpgradeCommand:
Elapsed: 1806ms | User: 0ms | Sys: 0ms
RSS: 95.79MB | Peak: 119.14MB | Commit: 0.15GB | Faults: 35545
----- bun meta -----

0   00007FF77B3BED58
1   ???
2   ???
3   ???
4   ???
5   ???
6   ???
7   ???
8   ???
9   ???

Search GitHub issues https://bun.sh/issues or join in #windows channel in https://bun.sh/discord

thread 116772 panic: attempt to unwrap error
???:?:?: 0x7ff779ffbdd1 in ??? (bun.exe)
???:?:?: 0x7ff779ff9945 in ??? (bun.exe)
???:?:?: 0x7ff779ffc406 in ??? (bun.exe)
???:?:?: 0x7ff77a0090e7 in ??? (bun.exe)
???:?:?: 0x7ff779d31603 in ??? (bun.exe)
???:?:?: 0x7ff77c8b3ec3 in ??? (bun.exe)
???:?:?: 0x7ffc7daf257c in ??? (KERNEL32.DLL)
???:?:?: 0x7ffc7fe0aa47 in ??? (ntdll.dll)
```

lokimckay:
[...]

PR #9972, opened 2024-04-05
Redo file watcher + Fix `EBUSY` when saving lockfile on Windows

Jarred-Sumner:
### What does this PR do?

This PR:
- Fixes the `EBUSY` error when saving the lockfile to disk
- Reimplements `fs.watch` on Windows to be more efficient. It deduplicates by path.

Fixes https://github.com/oven-sh/bun/issues/9980
Fixes https://github.com/oven-sh/bun/issues/9943
Fixes https://github.com/oven-sh/bun/issues/9966

One change is that we now save the temporary lockfile into the current working directory instead of the system temp folder.

The Next.js "dev ssr 100" test will catch this in the future.


src/bun.js/bindings/webcore/EventNames.cpp
```diff
@@ -40,4 +43,35 @@ const EventNames& eventNames()
...
+        case DOMEventName::change:
+            return eventNames().changeEvent;
+        case DOMEventName::error:
+            return eventNames().errorEvent;
+        case DOMEventName::abort:
+            return eventNames().abortEvent;
+        case DOMEventName::close:
+            return eventNames().closeEvent;
+        default:
+            RELEASE_ASSERT_NOT_REACHED();
```
cirospaciari, line 70: do we need `messageerror`, `open`, `message` here? asking because of `DOM_EVENT_NAMES_FOR_EACH` macro

src/bun.js/node/node_fs.zig
```diff
@@ -5862,7 +5862,7 @@ pub const NodeFS = struct {
         std.debug.assert(flavor == .sync);
 
         const watcher = args.createStatWatcher() catch |err| {
-            const buf = std.fmt.allocPrint(bun.default_allocator, "{s} watching {}", .{ @errorName(err), bun.fmt.QuotedFormatter{ .text = args.path.slice() } }) catch unreachable;
+            const buf = std.fmt.allocPrint(bun.default_allocator, "Failed to watch file {}", .{bun.fmt.QuotedFormatter{ .text = args.path.slice() }}) catch unreachable;
```
cirospaciari:
```suggestion
            const buf = std.fmt.allocPrint(bun.default_allocator, "Failed to watch file {}", .{bun.fmt.QuotedFormatter{ .text = args.path.slice() }}) catch bun.outOfMemory();
```

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -150,7 +153,101 @@ pub const FSWatcher = struct {
...
+        pub usingnamespace bun.New(@This());
+
+        file_path: StringOrBytesToDecode = .{ .bytes_to_free = "" },
+        event_type: EventType = .@"error",
+        ctx: *FSWatcher = undefined,
+
+        // To match the API of the posix version
+        count: u0 = 0,
+
+        pub usingnamespace bun.New(@This());
```
cirospaciari, line 181:
```suggestion

```

cirospaciari: already added in line 174

src/bun.js/node/win_watcher.zig
```diff
@@ -165,182 +57,192 @@ pub const PathWatcherManager = struct {
...
+            Output.debugWarn("uvEventCallback called with null data", .{});
+            return;
+        }
+        const this: *PathWatcher = @alignCast(@fieldParentPtr(PathWatcher, "handle", event));
+        if (comptime bun.Environment.isDebug) {
+            std.debug.assert(event.data == @as(?*anyopaque, @ptrCast(this)));
+        }
 
-        const timestamp = std.time.milliTimestamp();
+        const timestamp = event.loop.time;
```
cirospaciari, line 124: this time is only update when the tick starts or when we set some timer like `setTimeout`, I think is fine but if someone is doing block things this time can drift

src/deps/libuv.zig
```diff
@@ -1532,15 +1532,32 @@ pub const struct_uv_fs_event_s = extern struct {
...
-    filew: [*]WCHAR,
-    short_filew: [*]WCHAR,
-    dirw: [*]WCHAR,
+    filew: ?[*]WCHAR = null,
+    short_filew: ?[*]WCHAR = null,
+    dirw: ?[*]WCHAR = null,
     buffer: [*]u8,
+
+    pub fn isDir(this: *const uv_fs_event_t) bool {
+        return this.dirw != null;
```
cirospaciari, line 1546: nice

cirospaciari (changes requested):
LGTM just some comments

src/bun.js/bindings/webcore/EventNames.cpp
Jarred-Sumner: no, this is only used by fs.watch

src/bun.js/node/win_watcher.zig
Jarred-Sumner: yeah i think thats fine

Issue #9985, opened 2024-04-05
Bun.nanoseconds JSDoc seems to be cut off


tomerh2001:
### What is the type of issue?

Documentation is missing

### What is the issue?

The `Bun.nanoseconds` function seems to have it's JSDoc cut off:

```
	/**
	 * Nanoseconds since Bun.js was started as an integer.
	 *
	 * This uses a high-resolution monotonic system timer.
	 *
	 * After 14 weeks of consecutive uptime, this function
	 * wraps
	 */
	function nanoseconds(): number;
``` 

It also hints at a maximum value of 14 weeks, which is incorrect - the maximum value is roughly `2^64 - 1` nanoseconds (which is about 585 years)

PR #9986, opened 2024-04-05
docs: Updated Bun.nanoseconds documentation

tomerh2001:
### What does this PR do?

Updated `Bun.nanoseconds` function documentation.

- [ ] Code changes

Closes https://github.com/oven-sh/bun/issues/9985


tomerh2001:
https://github.com/oven-sh/bun/issues/9985

paperclover (changes requested):
thanks for this pr, you've brought up something subtle but important.

> And is represented as a 64-bit unsigned integer

JavaScript does not have this data type, numbers larger than this will lose precision. This function is intended to wrap at (2^53 - 1) (Number.MAX_SAFE_INTEGER) to ensure that the number is always percise, but after inspecting the code further (BunObject.zig, fn nanoseconds), this code does not wrap.

> unless the counter is artificially manipulated

I don't think this counter can be artificially manipulated. `origin_timer` is shared by the logic for performance.now (which also starts losing precision after 14.8 weeks)

Overall, I think the documentation should be updated to say something along the lines of "this loses precision after 14.8 weeks, due to floating point behavior of numbers above Number.MAX_SAFE_INTEGER." not sure exactly how to word it.

(14.8 is derived from `Number.MAX_SAFE_INTEGER / (1e+9) / 60 / 60 / 24 / 7`)

Jarred-Sumner:
(2^53) nanoseconds = 104.249991 days

tomerh2001:
@paperdave @Jarred-Sumner 
Thanks for the input. I made another attempt at rewording the JSDoc; how about now?

paperclover (approved):
thanks. though now this function is now very documented in comparison to others that are much shorter / no doc comment. heh.

tomerh2001:
> thanks. though now this function is now very documented in comparison to others that are much shorter / no doc comment. heh.
> 
This specifically troubled me since it hinted at a 14 weeks maximum value that wasn't explained

If I encounter similar issues I'll make sure to create additional PRs, also hopefully this'll encourage others to do the same :)

PR #9995, opened 2024-04-05
Fixes #9978

### LINKED ISSUE #9978: Bun install cannot install @tanstack/router-vite-plugin

1.1.1+ca1dbb4eb

Reproduce:
`bun install @tanstack/router-vite-plugin`

To install without any errors.

### What do you see instead?

`🚚 @tanstack/router-vite-plugin... error: ENOENT extracting tarball for @tanstack/router-vite-plugin`

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #9978

manual

### DIFF
--- a/src/libarchive/libarchive.zig
+++ b/src/libarchive/libarchive.zig
@@ -514,6 +514,17 @@ pub const Archive = struct {
                             &w_path_buf,
                             .windows,
                         );
+
+                        // When writing files on Windows, translate the characters to their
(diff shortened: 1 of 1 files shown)


quantizor:
If you wanted to preserve the meaning (just based on a reply to your post on x) you could do something like this

```
| to [pipe]
< to [lt]
> to [gt]
... etc
```

so `foo?` would become `foo[question]` or `foo[q]` perhaps

PR #10069, opened 2024-04-08
Fixes #9952

### LINKED ISSUE #9952: [Windows] Exiting a running bun process with Ctrl + C messes up the cmd command history

1.1.0

```JavaScript
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

(async () => {
  console.log('Running...')
  await sleep(10000)
  console.log('Passed.')
})()
```

While the process hasn't finished executed, if you force exit by pressing `Ctrl + C` and then press the **up arrow** to get the last command that was executed, a unrelated sequence of characters will be provided instead, such as `^[[A`

Video examples:

### What do you see instead?

After exiting a running process and pressing the up arrow, the last executed command should be provided, for those that don't use something like nodemon or watch mode this is extremely important to quickly run the last command again.

Video example:

https://github.com/oven-sh/bun/assets/58047457/0ca27626-06cf-4e3f-a17d-66d552f26ff5

### Additional information

If you continue to press the up arrow you'll constantly get the same string again and again.

image

Jarred-Sumner:
### What does this PR do?

Fixes #9952

`bun i` == before
`bun-debug i` == after

Screen Recording 2024-04-08 at 5 57 53 AM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
 
         # TODO(@paperdave): Remove this
         # it is enabled for the time being to make sure to catch more bugs in the experimental windows builds
-	set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
+        set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
     else()
         if(ZIG_OPTIMIZE STREQUAL "Debug")
             set(bun "bun-debug")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
 
         # TODO(@paperdave): Remove this
         # it is enabled for the time being to make sure to catch more bugs in the experimental windows builds
-	set(DEFAULT_ZIG_OPTIMIZE "ReleaseFast")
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
 
         # TODO(@paperdave): Remove this
         # it is enabled for the time being to make sure to catch more bugs in the experimental windows builds
-        set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
+	set(DEFAULT_ZIG_OPTIMIZE "ReleaseFast")
```
Jarred-Sumner:
```suggestion
	set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
```

CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
 
         # TODO(@paperdave): Remove this
         # it is enabled for the time being to make sure to catch more bugs in the experimental windows builds
-        set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
+	set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
```
Jarred-Sumner:
```suggestion
        set(DEFAULT_ZIG_OPTIMIZE "ReleaseSafe")
```

PR #10091, opened 2024-04-08
fix: null is not an object at readableStreamCancel

### LINKED ISSUE #9542: TypeError: null is not an object at readableStreamCancel

1.0.30+1424a196f

Linux 6.5.0-26-generic x86_64 x86_64

```js
fetch('https://jsonplaceholder.typicode.com/posts')
  .then(res => res.body?.cancel())
  .then(() => {
    console.log('good');
  });
```

Should log "good"

### What do you see instead?

```
1 | (function (stream, reason) {"use strict";
                        ^
TypeError: null is not an object
      at readableStreamCancel (:1:21)
➜  fetch-tools git:(main) ✗ bun test.mjs
1 | (function (stream, reason) {"use strict";
                        ^
TypeError: null is not an object
      at readableStreamCancel (:1:21)
```

### Additional information

_No response_

jdalton:
### What does this PR do?

Fixes #9542 

[x] Added test.

I still need to learn more about the stream lifecycle what cases the `controller` is `null`.

### DIFF
--- a/test/js/web/fetch/response.test.ts
+++ b/test/js/web/fetch/response.test.ts
@@ -7,7 +7,7 @@ test("zero args returns an otherwise empty 200 response", () => {
 });
 
 test("calling cancel() on response body doesn't throw", () => {
-  expect(() => new Response('').body?.cancel()).not.toThrow();
+  expect(() => new Response("").body?.cancel()).not.toThrow();
 });
 
 test("undefined args don't throw", () => {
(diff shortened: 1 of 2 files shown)


test/js/web/fetch/response.test.ts
```diff
@@ -6,6 +6,10 @@ test("zero args returns an otherwise empty 200 response", () => {
   expect(response.statusText).toBe("");
 });
 
+test("calling cancel() on response body doesn't throw", () => {
+  expect(() => new Response("").body?.cancel()).not.toThrow();
+});
+
 test("undefined args don't throw", () => {
```
jerome-benoit, line 13:
```suggestion
test("undefined args doesn't throw", () => {
```

PR #10102, opened 2024-04-09
fix(which/windows): ignore file extension case

### LINKED ISSUE #10101: Bun is case-sensetive to file extensions on Windows

1.1.3+2615dc742

Microsoft Windows NT 10.0.22635.0 x64

Returns an error:

```cmd
bun -e "var spawn = require('child_process').spawn;spawn('notepad.EXE')"
```

Works correctly:

```cmd
bun -e "var spawn = require('child_process').spawn;spawn('notepad.exe')"
```

Happens due to this check:
https://github.com/oven-sh/bun/blob/e30a848c4c7a4aa80c8c15a6b559ea55e4233fc5/src/which.zig#L82

I expect Bun to be case-insensitive to file extensions, so the mentioned command will open notepad

### What do you see instead?

```
TypeError: Executable not found in $PATH: "notepad.EXE"
  code: "ERR_INVALID_ARG_TYPE"

    at spawn (node:child_process : 650:15)
    at node:child_process : 2:41
    at globalThis (C:\Users\Sunset\[eval]: 1:44)
```

### Additional information

Node handles this correctly:
```cmd
node -e "var spawn = require('child_process').spawn;spawn('notepad.EXE')"
```

SunsetTechuila:
### What does this PR do?

This fixes #10101

- [x] Code changes

I compiled the executable and tested it manually


paperclover (changes requested):
can you add a unit test to verify this works?

SunsetTechuila:
Ok, I will

SunsetTechuila:
@paperdave done
but I have no idea how to run this test
don't think this can be done with js/ts test since this is a windows-specific function

gvilums:
We don't currently run zig tests in our test suite. Instead, I suggest adding something like the following to `which.test.ts`:
```typescript
test.if(isWindows)("which is case-insensitive", () => {
  expect(which("cmd.EXE")).toBe("C:\\Windows\\system32\\cmd.exe");
})
```
(feel free to adjust)

SunsetTechuila:
oh, so you have os-specific js/ts tests, i see now

SunsetTechuila:
@gvilums done

Jarred-Sumner:
Thank you

PR #10086, opened 2024-04-08
fix(webkit): backport pipeTo memory leak fix from WebKit

lithdew:
Fixes a memory leak identified by passing an AbortSignal to pipeTo. This PR is one of a series of PR's aimed to resolve [memory leaks while running a Next.js standalone build](https://github.com/oven-sh/bun/issues/9977).

For more details, refer to this WebKit PR [here](https://github.com/WebKit/WebKit/pull/6759).

### What does this PR do?

This PR backports memory leak fixes to do with piping a ReadableStream into a WritableStream with an AbortSignal passed through from the following [WebKit PR](https://github.com/WebKit/WebKit/pull/6759) authored by @youennf.

This PR additionally removes the exported builtin on AbortSignal, `whenSignalAborted`, which is a builtin that is not officially documented.

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

#9977 contains an example benchmark to do with spamming a Node.JS HTTP server that pipes a ReadableStream to http.ServerResponse. On Bun v1.1.3, the benchmark causes memory to leak.

Existing tests were run to ensure that the fixes that are backported to AbortSignal pass.

- [x] I included a test for the new code, or existing tests cover it
- [x] I ran my tests locally and they pass (`bun-debug test test-file-name.test`)

Issue #9977, opened 2024-04-05
Memory leak running Next.js standalone bundle


lithdew:
### What version of Bun is running?

1.1.1+ca1dbb4eb

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

1. Create a new Next.js app.
 
```console
$ bun create next-app --ts --tailwind --app --use-bun --eslint --no-src-dir --import-alias "@/*" test`
```

2. Make modifications to the following files.

`next.config.mjs`
```js
/** @type {import('next').NextConfig} */
const nextConfig = {output: "standalone"};

export default nextConfig;
```

`app/layout.tsx`
```ts
import type { Metadata } from "next";
import { Inter } from "next/font/google";
import "./globals.css";

const inter = Inter({ subsets: ["latin"] });

export const metadata: Metadata = {
  title: "Create Next App",
  description: "Generated by create next app",
};

export const dynamic = "force-dynamic";

export default function RootLayout({
  children,
}: Readonly<{
  children: React.ReactNode;
}>) {
  return (
    <html lang="en">
      <body className={inter.className}>{children}</body>
    </html>
  );
}
```

3. Build Next.js standalone bundle.

```console
$ bun run build
```

4. Run Next.js standalone bundle.

```console
$ bun run .next/standalone/server.js
```

5. Spam your Next.js app with web requests.
```console
$ go install -mod=mod github.com/codesenberg/bombardier
$ bombardier -c 200 -d 1000h http://localhost:3000
```

6. Notice memory usage go up indefinitely.

### What is the expected behavior?

No memory leak. Next.js on Node generally consumes roughly 400MB of memory. Even with assumptions on increased memory usage, I would assume Next.js on Bun should consume no more than 1GB of memory.

### What do you see instead?

<img width="885" alt="image" src="https://github.com/oven-sh/bun/assets/63115601/86fc6483-aedd-43e8-a25b-cc27e0ffdc1c">

### Additional information

Happy to help with debugging/fixing this memory leak - would like to know what tools maintainers are using for debugging these memory leaks.

mmarifat:
Getting the same with latest bun version. Was not facing in bun 1.0.35

lithdew:
> Getting the same with latest bun version. Was not facing in bun 1.0.35

Tested it out with bun 1.0.35+940448d6b and got the same issue.

lithdew:
[...]

PR #10106, opened 2024-04-09
Upgrade AbortSignal & AbortController to latest from WebKit

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/9977
Closes https://github.com/oven-sh/bun/pull/10086

Thank you @lithdew for investigating and most of the fixes. This adds more of the changes we missed from WebKit into Bun like the ability to follow other signals

1st:  after 2,000 requests
2nd: after 10,000 requests
3rd: after 50,000 requests

image

### DIFF
--- a/src/bun.js/bindings/WebCoreOpaqueRoot.h
+++ b/src/bun.js/bindings/WebCoreOpaqueRoot.h
@@ -26,6 +26,8 @@
 
 #pragma once
 
+#include "Node.h"
+
 namespace WebCore {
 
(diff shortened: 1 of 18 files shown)


zackify:
I'm seeing a leaking issue using remix with 

`import { renderToReadableStream } from "react-dom/server.bun";`

let me know if i can help debug in some way. I'm trying to change a few things if I can isolate where it is. 

Memory just grows like crazy as more requests come in.

PR #10124, opened 2024-04-09
feat(auto-corking) enable auto-corking in uWS

cirospaciari:
### What does this PR do?

When streaming this reduces the amount of write operations if we auto uncork by auto cork again when needed

- [X] Code changes

Existing tests and manual performance test

```js
const data = Buffer.alloc(8 * 1024, 'X');
// async iterator
async function* generate() {
  for(let i = 0; i < 10; i++) {
    // simulate something that retrieve 3 chunks of data
    await Bun.sleep(1); 
    yield data;
    yield data;
    yield data;
  }
}
const server = Bun.serve({
  port: 8080,
  fetch(req) {
    const webStream = new ReadableStream({
      async pull(controller) {
        await 1
        const it = generate();
        for await(const chunk of it) {
          controller.enqueue(chunk);
        }
        controller.close();
      } 
    });
    return new Response(webStream);
  },
});
```
image

### DIFF
--- a/scripts/make-old-js.ps1
+++ b/scripts/make-old-js.ps1
@@ -3,7 +3,7 @@ $npm_client = "npm"
 # & ${npm_client} i
 
 $root = Join-Path (Split-Path -Path $MyInvocation.MyCommand.Definition -Parent) "..\"
-$esbuild = Join-Path $root "node_modules\.bin\esbuild.cmd"
+$esbuild = Join-Path $root "node_modules\.bin\esbuild.exe"
 
 $env:NODE_ENV = "production"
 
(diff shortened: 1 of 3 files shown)


scripts/make-old-js.ps1
```diff
@@ -3,7 +3,7 @@ $npm_client = "npm"
 # & ${npm_client} i
 
 $root = Join-Path (Split-Path -Path $MyInvocation.MyCommand.Definition -Parent) "..\"
-$esbuild = Join-Path $root "node_modules\.bin\esbuild.cmd"
```
mGranTY, line 6: reminder to change it back to .cmd

Jarred-Sumner:
im very skeptical of merging this without any new tests

cirospaciari:
> im very skeptical of merging this without any new tests

Will push a test that I used to force the auto cork with some variants

PR #10129, opened 2024-04-09
Re-sync `URL` from WebKit + set `ERR_MISSING_ARGS`

Jarred-Sumner:
### What does this PR do?

- Re-sync `URL` from WebKit, which should 1/2 the size of the JS `URL` object in memory
- Sets `ERR_MISSING_ARGS` everywhere we throw createNotEnoughArguments

5%-ish perf improvement to `new URL`
image

There is a test


paperclover (approved):
good

PR #10130, opened 2024-04-09
stdout and sterr are in err

DaleSeo:
### What does this PR do?

The `output` object is not present in the `catch` clause. `stdout` and `sterr` can be accessed from the `err` object.

- [ ] Code changes


Jarred-Sumner:
Thank you

PR #10161, opened 2024-04-11
Upgrade to LLVM 17, fix linux debug build, upgrade JSC, remove some C API usages

Jarred-Sumner:
### What does this PR do?

- Upgrade from LLVM 16 -> LLVM 17
- Linux debug build works once again
- Upgrade JSC
- Replace C API usage of JSPropertyIterator with a C++ bindings one, which also avoids cloning the property names an extra time
- Adds a `BunString` version of `JSValue.put` which will be faster in most cases since it avoids atomizing the strings again
- Bunch of small changes for strings for the JSC upgrade

This won't compile until the upgraded JSC successfully compiles in CI. Tested locally.


scripts/setup.sh
```diff
@@ -78,6 +78,7 @@ rm -f build/CMakeCache.txt
 cmake -B build -S . \
   -G Ninja \
   -DUSE_DEBUG_JSC=ON \
+  -DWEBKIT_DIR="$(pwd)/src/bun.js/WebKit/WebKitBuild/Debug" \
```
dylan-conway:
```suggestion
```

dylan-conway (approved):

PR #10174, opened 2024-04-11
feat: Add `cp` builtin to shell for Windows

zackradisic:
### What does this PR do?

This adds the `cp` shell builtin to Bun shell

Not all options are supported so it is only enabled on Windows right now


ghiscoding:
I'm assuming this PR should probably reference #9716

src/bun.js/node/node_fs.zig
```diff
@@ -242,10 +242,79 @@ pub const Async = struct {
     }
 };
 
+pub const AsyncCPTaskVtable = struct {
```
gvilums: I'm not sure I understand the point of the vtable here. The set of implementations is closed - either it's in js, or it's in the shell. Wouldn't a simple boolean suffice? The vtable introduces a lot of complexity, and handling checks by comparing pointers to a magic value is bad practice in my opinion.
My suggestion: move the `is_shell` boolean to the `AsyncCpTask`, and just use `if` statements instead of a vtable.

gvilums (changes requested):

src/bun.js/node/node_fs.zig
```diff
@@ -243,9 +243,9 @@ pub const Async = struct {
 };
 
 pub const AsyncCpTask = struct {
```
Jarred-Sumner: Can't we make this a comptime fn and then have `AsyncShellCpTask`?

Jarred-Sumner:
I don't think we can ship this without `cp -r`

Jarred-Sumner (changes requested):
This segfaults when copying directories.

That means:
1) The automated tests are not nearly comprehensive enough
2) You didn't manually test copying directories

Please test your code both manually and through automated means.

```js
PS C:\bun> bun-10174 exec 'cp -R node_modules C:\temp\node_mods'
Segmentation fault at address 0xb0
???:?:?: 0x7ff6b7deebc3 in ??? (bun-10174.exe)
???:?:?: 0x7ffd56de257c in ??? (KERNEL32.DLL)
???:?:?: 0x7ffd5840aa47 in ??? (ntdll.dll)
```

Jarred-Sumner:
many test failures

src/bun.zig
```diff
@@ -768,6 +768,22 @@ pub fn openDirAbsolute(path_: []const u8) !std.fs.Dir {
...
+                is_enabled = false;
+                if (getenvZ(flag_)) |val| {
+                    is_enabled = strings.eqlComptime(val, "1") or strings.eqlComptime(val, "true");
+                }
+
+                break :brk is_enabled.?;
+            };
+        }
+    }.get();
+}
```
gvilums: this has a data race in multi-threaded code, `is_enabled` should be atomic

src/shell/interpreter.zig
```diff
@@ -10105,6 +10118,616 @@ pub const Interpreter = struct {
...
+                fn enqueueToEventLoop(this: *ShellCpTask) void {
+                    if (this.event_loop == .js) {
+                        this.event_loop.js.enqueueTaskConcurrent(this.concurrent_task.js.from(this, .manual_deinit));
+                    } else {
+                        this.event_loop.mini.enqueueTaskConcurrent(this.concurrent_task.mini.from(this, "runFromMainThreadMini"));
+                    }
+                }
+
+                pub fn runFromMainThread(this: *ShellCpTask) void {
+                    print("runFromMainThread", .{});
```
gvilums: is this print supposed to be here?

gvilums (commented):
Looks good overall. See minor feedback related to flags processing

src/bun.zig
Jarred-Sumner: its harmless if it's checked multiple times

gvilums: if it's null and two threads try to initialize it at the same time you get undefined behavior because the two writes do not synchronize

zackradisic: In the case that two threads try to initialize it, would it matter since they would both compute the same value sincee env is constant? I guess it would result in a slight performance degradation

I'll make it atomic just to be safe anyway

src/shell/interpreter.zig
zackradisic: this is the bun output scoped thing, not std.debug.print

src/bun.zig
gvilums: yes it would matter, two threads writing to the same memory without synchronization is undefined behavior, no matter if they're writing the same value or not

Jarred-Sumner:
@zackradisic there are some conflicts now due to the other PR

Jarred-Sumner:
but its very close to mergeable

Jarred-Sumner:
@zackradisic the next step for this PR is to fix the failing cp test on Windows

src/bun.js/node/node_fs.zig
```diff
@@ -242,111 +242,486 @@ pub const Async = struct {
...
+                    .tracker = JSC.AsyncTaskTracker{ .id = 0 },
+                    .arena = arena,
+                    .subtask_count = .{ .raw = 1 },
+                    .shelltask = shelltask,
+                },
+            );
+            task.ref.ref(mini);
+            task.args.src.toThreadSafe();
+            task.args.dest.toThreadSafe();
+            // task.tracker.didSchedule(globalObject);
```
```suggestion

```

src/bun.js/node/node_fs.zig
```diff
@@ -3432,11 +3741,14 @@ pub const Arguments = struct {
             recursive: bool,
             errorOnExist: bool,
             force: bool,
+            deinit_paths: bool = true,
         };
 
         fn deinit(this: Cp) void {
```
```suggestion
        fn deinit(this: *Cp) void {
```

src/bun.js/node/node_fs.zig
```diff
@@ -3432,11 +3741,14 @@ pub const Arguments = struct {
...
-            this.src.deinit();
-            this.dest.deinit();
+            if (this.flags.deinit_paths) {
```
Jarred-Sumner, line 3747: i don't feel good about this

src/shell/interpreter.zig
```diff
@@ -10101,6 +10114,734 @@ pub const Interpreter = struct {
...
+        pub const Cp = struct {
+            bltn: *Builtin,
+            opts: Opts = .{},
+            state: union(enum) {
+                idle,
+                exec: struct {
+                    target_path: [:0]const u8,
+                    paths_to_copy: []const [*:0]const u8,
+                    started: bool = false,
+                    tasks_count: i32 = 0,
```
Jarred-Sumner: is this thread safe? does it need to be thread safe? why is it an i32 instead of a u32?

src/shell/interpreter.zig
```diff
@@ -10101,6 +10114,734 @@ pub const Interpreter = struct {
...
+                        .waiting_write_err => return,
+                        .done => unreachable,
+                    }
+                }
+
+                this.bltn.done(0);
+            }
+
+            pub fn deinit(this: *Cp) void {
+                _ = this; // autofix
```
Jarred-Sumner: Is this intentionally missing a deinit method?

test/js/bun/shell/commands/cp.test.ts
```diff
@@ -0,0 +1,177 @@
...
+const p = process.platform === "win32" ? (s: string) => s.replaceAll("/", "\\") : (s: string) => s;
+
+$.nothrow();
+
+describe.if(!builtinDisabled("cp"))("bunshell cp", async () => {
+  TestBuilder.command`cat ${import.meta.filename} > lmao.txt; cp -v lmao.txt lmao2.txt`
+    .stdout(p("$TEMP_DIR/lmao.txt -> $TEMP_DIR/lmao2.txt\n"))
+    .ensureTempDir()
+    .testMini()
+    .fileEquals("lmao2.txt", await $`cat ${import.meta.filename}`.text())
```
Jarred-Sumner: If this throws, the rest of the tests in this file will not run and we will never know

Please move all of usages of top-level await in this test to be inside closures.

test/js/bun/shell/commands/dirname.test.ts
```diff
@@ -3,6 +3,7 @@ import { describe, test, expect } from "bun:test";
 import { createTestBuilder } from "../test_builder";
 const TestBuilder = createTestBuilder(import.meta.path);
 
+$.nothrow();
```
Jarred-Sumner: why?

test/js/bun/shell/commands/exit.test.ts
```diff
@@ -5,6 +5,7 @@ const TestBuilder = createTestBuilder(import.meta.path);
 import { sortedShellOutput } from "../util";
 import { join } from "path";
 
+$.nothrow();
```
Jarred-Sumner: why? shouldn't the TestBuilder suppress the exception if it's expected to have a non-zero exit code?

src/shell/interpreter.zig
```diff
@@ -10101,6 +10114,734 @@ pub const Interpreter = struct {
...
+                const exit_code = this.state.ebusy.main_exit_code;
+                this.state = .done;
+                this.bltn.done(exit_code);
+            }
+
+            pub fn next(this: *Cp) void {
+                while (this.state != .done) {
+                    switch (this.state) {
+                        .idle => @panic("Invalid state for \"Cp\": idle, this indicates a bug in Bun. Please file a GitHub issue"),
+                        .exec => {
```
Jarred-Sumner, line 10258: maybe there should be a shell panic that includes additional information like what command was run?

src/shell/interpreter.zig
```diff
@@ -10101,6 +10114,734 @@ pub const Interpreter = struct {
...
+                cwd_path: [:0]const u8,
+                verbose_output_lock: std.Thread.Mutex = .{},
+                verbose_output: ArrayList(u8) = ArrayList(u8).init(bun.default_allocator),
+
+                task: JSC.WorkPoolTask = .{ .callback = &runFromThreadPool },
+                event_loop: JSC.EventLoopHandle,
+                concurrent_task: JSC.EventLoopTask,
+                err: ?bun.shell.ShellErr = null,
+
+                const print = bun.Output.scoped(.ShellCpTask, false);
```
Jarred-Sumner: we don't use `print` anywhere else in the codebase for these

can this be called `debug` or `log`?

src/shell/interpreter.zig
```diff
@@ -10101,6 +10114,734 @@ pub const Interpreter = struct {
...
+                    return ResolvePath.Platform.auto.isSeparator(path[path.len - 1]);
+                }
+
+                const Kind = enum {
+                    file,
+                    dir,
+                };
+
+                pub fn isDir(this: *ShellCpTask, path: [:0]const u8) Maybe(bool) {
+                    _ = this;
```
```suggestion

```

src/shell/interpreter.zig
```diff
@@ -10101,6 +10114,734 @@ pub const Interpreter = struct {
...
+                    };
+                    this.onCopyImpl(src, dest);
+                }
+
+                pub fn cpOnFinish(this: *ShellCpTask, result: Maybe(void)) void {
+                    this.onSubtaskFinish(result);
+                }
+            };
+
+            const Opts = struct {
```
```suggestion
            const Opts = packed struct {
```

Jarred-Sumner (changes requested):
A few comments

src/shell/interpreter.zig
zackradisic: its thread safe as it is only read and mutated by the main thread, changed it to a u32

Jarred-Sumner:
@zackradisic 
> [test/js/bun/shell/commands/cp.test.ts](https://github.com/oven-sh/bun/pull/10174#testjsbunshellcommandscptestts) 1 failing

test/js/bun/shell/test_builder.ts
```diff
@@ -17,32 +18,42 @@ export function createTestBuilder(path: string) {
   });
 
   class TestBuilder {
-    promise: { type: "ok"; val: ShellPromise } | { type: "err"; val: Error };
+    // promise: { type: "ok"; val: ShellPromise } | { type: "err"; val: Error };
```
gvilums: can we delete this?

test/js/bun/shell/test_builder.ts
```diff
@@ -260,63 +304,104 @@ export function createTestBuilder(path: string) {
...
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+
+  return result;
+}
+
+export function bunExe() {
+  if (process.platform === "win32") return process.execPath.replaceAll("\\", "/");
+  return process.execPath;
+}
```
gvilums, line 406: don't we already have a `bunExe` function in `harness.ts`?

test/js/bun/shell/bunshell-default.test.ts
```diff
@@ -1,25 +1,54 @@
...
-    await $`echo hi; ls oogabooga`.quiet();
-    expect.unreachable();
-  } catch (e: any) {
-    expect(e).toBeInstanceOf(Error);
-    expect(e.exitCode).toBe(1);
-    expect(e.message).toBe("Failed with exit code 1");
-    expect(e.stdout.toString("utf-8")).toBe("hi\n");
-    expect(e.stderr.toString("utf-8")).toBe("ls: oogabooga: No such file or directory\n");
-  }
+  // Run in a subproc because other tests may change the value of $.throws
```
gvilums, line 7: A thought: Maybe it would make sense to allow users to create a shell instance that is configurable and not global, e.g. `const $ = Bun.$.withOptions(...);`. This would alleviate the issues here. (obviously this is out of scope for this PR)

src/shell/subproc.zig
```diff
@@ -505,7 +505,7 @@ pub const ShellSubprocess = struct {
         // }
 
         // if (!this.hasCalledGetter(.stderr)) {
-        this.stdout.unref();
+        this.stderr.unref();
         // }
```
gvilums: can we get rid of the commented-out stuff?

src/shell/interpreter.zig
```diff
@@ -10107,6 +10120,744 @@ pub const Interpreter = struct {
...
+            /// executed, to avoid complicated synchronization on multiple
+            /// threads, because the precise src or dest for each argument is
+            /// not known until its corresponding ShellCpTask is executed by the
+            /// threadpool.
+            const EbusyState = struct {
+                tasks: std.ArrayListUnmanaged(*ShellCpTask) = .{},
+                absolute_targets: std.StringArrayHashMapUnmanaged(void) = .{},
+                absolute_srcs: std.StringArrayHashMapUnmanaged(void) = .{},
+
+                pub fn deinit(this: *EbusyState) void {
```
gvilums, line 10178: do we need to free the tasks themselves?

src/shell/interpreter.zig
```diff
@@ -10107,6 +10120,744 @@ pub const Interpreter = struct {
...
+
+            pub fn ignoreEbusyErrorIfPossible(this: *Cp) void {
+                if (!bun.Environment.isWindows) @compileError("dont call this plz");
+
+                if (this.state.ebusy.idx < this.state.ebusy.state.tasks.items.len) {
+                    outer_loop: for (this.state.ebusy.state.tasks.items[this.state.ebusy.idx..], 0..) |task_, i| {
+                        const task: *ShellCpTask = task_;
+                        assert(task.tgt_absolute != null);
+                        const failure_src = task.src_absolute.?;
+                        const failure_tgt = task.tgt_absolute.?;
```
gvilums: why do we assert the target but not the source? do we need the assert at all if it's only going to be in debug builds anyway, which check `.?` ?

src/shell/interpreter.zig
```diff
@@ -10107,6 +10120,744 @@ pub const Interpreter = struct {
...
+                        task.src_absolute = null;
+                        if (src_absolute) |tgt| this.state.exec.ebusy.absolute_srcs.put(bun.default_allocator, tgt, {}) catch bun.outOfMemory();
+                    }
+                }
+
+                this.printShellCpTask(task);
+            }
+
+            pub fn printShellCpTask(this: *Cp, task: *ShellCpTask) void {
+                defer task.deinit();
```
gvilums, line 10369: is this `defer` correct? why are we deinitializing the task if we're starting it in all codepaths?

gvilums (approved):

test/js/bun/shell/test_builder.ts
zackradisic: the shell leak tests do some hacky things like taking the `test_builder.ts` file and copying it, its easier to just rewrite the function here

test/js/bun/shell/bunshell-default.test.ts
zackradisic: Yeah we should do this

src/shell/interpreter.zig
zackradisic: added a comment they are freed in ignoreEbusyErrorIfPossible

src/bun.js/node/node_fs.zig
```diff
@@ -242,111 +242,491 @@ pub const Async = struct {
...
+                }
+
+                this.deinit();
+            }
+
+            pub fn deinit(this: *ThisSingleTask) void {
+                // There is only one path buffer for both paths. 2 extra bytes are the nulls at the end of each
+                bun.default_allocator.free(this.src.ptr[0 .. this.src.len + this.dest.len + 2]);
+
+                bun.destroy(this);
```
Jarred-Sumner, line 341: not sure why this is using `bun.destroy` / `bun.new` instead of `usingnamespace bun.New` and ThisSingleTask.new / this.destroy

src/bun.js/node/node_fs.zig
```diff
@@ -242,111 +242,491 @@ pub const Async = struct {
...
+            this: *ThisAsyncCpTask,
+        ) void {
+            const args = this.args;
+            var src_buf: bun.OSPathBuffer = undefined;
+            var dest_buf: bun.OSPathBuffer = undefined;
+            const src = args.src.osPath(@ptrCast(&src_buf));
+            const dest = args.dest.osPath(@ptrCast(&dest_buf));
+
+            if (Environment.isWindows) {
+                const attributes = windows.GetFileAttributesW(src);
```
Jarred-Sumner, line 518: I'm surprised we don't have a helper for this

PR #10179, opened 2024-04-11
fix path resolution for writeFile in nodefs

### LINKED ISSUE #10150: fs.writeFileSync and relative path incorrect on windows

1.1.3+2615dc742

Using latest version of bun for windows (v1.1.3)
**fs.writeFileSync** and relative paths do not work as expected (does not work like nodejs)

`fs.mkdirSync('../../config')`
=> OK

`fs.writeFileSync('../../test.txt', data)`
=> file will be created in **wrong directory**, will always be `../test.txt` regardless of `../../../..`

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

gvilums:
### What does this PR do?
Fixes #10150

Add a new test that checks this case

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -5336,58 +5336,12 @@ pub const NodeFS = struct {
     }
 
     pub fn writeFileWithPathBuffer(pathbuf: *[bun.MAX_PATH_BYTES]u8, args: Arguments.WriteFile) Maybe(Return.WriteFile) {
-        var path: [:0]const u8 = undefined;
-        var pathbuf2: [bun.MAX_PATH_BYTES]u8 = undefined;
-
         const fd = switch (args.file) {
             .path => brk: {
-                // On Windows, we potentially mutate the path in posixToPlatformInPlace
-                // We cannot mutate JavaScript strings in-place. That will break many things.
-                // So we must always copy the path string on Windows.
-                path = path: {
-                    const temp_path = args.file.path.sliceZWithForceCopy(pathbuf, Environment.isWindows);
-                    if (Environment.isWindows) {
-                        bun.path.posixToPlatformInPlace(u8, temp_path);
-                    }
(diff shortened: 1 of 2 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -5336,58 +5336,12 @@ pub const NodeFS = struct {
...
-                if (Environment.isWindows) {
-                    // windows openat does not support path traversal, fix it here.
-                    // use pathbuf2 here since without it 'panic: @memcpy arguments alias' triggers
-                    if (std.mem.indexOf(u8, path, "\\.\\") != null or std.mem.indexOf(u8, path, "\\..\\") != null) {
-                        const fixed_path = bun.path.normalizeStringWindows(path, &pathbuf2, false, false);
-                        pathbuf2[fixed_path.len] = 0;
-                        path = pathbuf2[0..fixed_path.len :0];
-                    }
-                }
+                const path = args.file.path.sliceZ(pathbuf);
```
dylan-conway: Before this was using `sliceZWithForceCopy`. If the path isn't copied into the buffer and an error is returned after `openat`, will the path memory in the error still be alive everywhere it's used?

Jarred-Sumner: yeah i think this makes tests flakier

gvilums: updated to use force copy

Jarred-Sumner:
@gvilums test is failing
```js
2986 | it("using writeFile on an fd does not truncate it", () => {
2987 | const temp = tmpdir();
2988 | const fd = fs.openSync(join(temp, "file.txt"), "w+");
^
EPERM: Operation not permitted
errno: -1
syscall: "open"
path: "C:\Users\RUNNER~1\AppData\Local\Temp\bun-test-tmp-7l0d_-734694376_61zg\file.txt"
```

Jarred-Sumner:
Thank you

PR #10199, opened 2024-04-12
Do not run shell tests outside test scope

Jarred-Sumner:
### What does this PR do?

@zackradisic when one of these shell tests fail, it prevents other tests from running

I'll let you take it from here to get the shell tests passing once again


Jarred-Sumner:
@zackradisic tests still fail

PR #10203, opened 2024-04-12
feat: overhaul the crash handler

paperclover:
This deletes the existing `crash_handler.zig`, `panic_handler.zig`, `report.zig`, and replaces it with a new crash handler which runs in all release builds. The primary goal of this is to allow all users to report useful crash reports to us using a "Trace String", a url-safe encoding of a stack trace or error return trace which can be demangled after the fact.

The tracestrings are all URLs from the base `https://bun.report`, where visiting the URL in a browser will redirect to a prompt to submit a GitHub issue, pre-filled with the remapped stack trace. Going to the homepage will show a frontend tool to just do the remapping.

Bun also adds `--verbose-error-trace`, which in platforms with [error return traces](https://ziglang.org/documentation/master/#Error-Return-Traces) enabled, these trace strings will be printed for any place we catch an error. This will help places we see users run into bun install errors but we don't know where it is at; We can ask them to run with the flag and share trace strings.

**TODOs**:

- [x] Implement tracestrings on all platforms
  - [x] Windows
    - [x] Encode
    - [x] Parse
    - [x] Remap
  - [x] macOS
    - [x] Encode
    - [x] Parse
    - [x] Remap
  - [x] Linux
    - [x] Encode (basically done, need to double-check it entirely works)
    - [x] Parse
    - [x] Remap
- [x] Deploy `bun.report`
  - [X] Write remapping service
  - [x] Fetching of debug information from bucket
  - [x] GitHub redirect
  - [x] View redirect
  - [X] Frontend tool
  - [x] Smart caching of already computed data
- [ ] Verify that `--watch` reload-on-crash behavior works
  - [x] On macOS
  - [ ] On Linux
  - [ ] On Windows
- [X] Make `main` returning an error a better experience.
- [X] Implement `bun.handleErrorReturnTrace` and `--verbose-error-trace`
- [x] Configure CI to store all debug symbols for all builds of bun.
- [X] Delete old dead implementations
- [x] Final pass over messages for trace strings. Currently uses a placeholder.
- [x] ~~Revert changes into `Bun.unsafe` / change `Bun.unsafe.segfault` into `Bun.unsafe.crash` / remove that API entirely.~~ Make Bun.unsafe.segfault private.
- [ ] Verify multi-thread crashes work as expected
  - [ ] On macOS
  - [ ] On macOS --watch
  - [ ] On Linux
  - [ ] On Linux --watch
  - [ ] On Windows
  - [ ] On Windows --watch

**Considering TODOs**

- [ ] Dump crash reports to file. This behavior was removed as it's previous form was not useful.
- [ ] Get JS-stack trace information to identify non-user code. This wont be able to get source mapping though.

I might do some more breaking changes to the format of trace-strings before the main merge. That is what I'm most worried about for now.


src/cli/run_command.zig
```diff
@@ -557,9 +556,6 @@ pub const RunCommand = struct {
                     }
 
                     Output.prettyErrorln("<r><red>error<r>: Failed to run \"<b>{s}<r>\" due to <r><red>{s}<r>", .{ basenameOrBun(executable), @errorName(err) });
-                    if (@errorReturnTrace()) |trace| {
```
dylan-conway, line 560: Are all of these meant to be replaced with `bun.handleErrorReturnTrace(err, @errorReturnTrace());`? These four seem to be missing if so

src/bun.zig
```diff
@@ -1591,14 +1601,23 @@ pub fn reloadProcess(
         const rc = bun.windows.TerminateProcess(@ptrFromInt(std.math.maxInt(usize)), win32.watcher_reload_exit);
```
dylan-conway:
```suggestion
        const rc = bun.windows.TerminateProcess(bun.windows.GetCurrentProcess(), win32.watcher_reload_exit);
```

src/bun.zig
```diff
@@ -1591,14 +1601,23 @@ pub fn reloadProcess(
...
         if (rc == 0) {
             const err = bun.windows.GetLastError();
+            if (may_return) {
+                Output.errGeneric("Failed to reload process: {s}", .{@tagName(err)});
+                return;
+            }
             Output.panic("Error while reloading process: {s}", .{@tagName(err)});
         } else {
+            if (may_return) {
+                Output.errGeneric("Failed to reload process", .{});
```
dylan-conway, line 1611:
```suggestion
```

src/bun.zig
```diff
@@ -3071,13 +3096,16 @@ pub fn SliceIterator(comptime T: type) type {
 
 pub const Futex = @import("./futex.zig");
 
+pub const crash_handler = @import("crash_handler.zig");
```
dylan-conway, line 3099: nit: feels like this should be `CrashHandler`, similar to `bun.Global`

src/cli.zig
```diff
@@ -41,34 +42,31 @@ const Router = @import("./router.zig");
...
+    pub fn start(allocator: std.mem.Allocator) void {
+        is_main_thread = true;
         start_time = std.time.nanoTimestamp();
         log_ = logger.Log.init(allocator);
 
         var log = &log_;
 
-        var panicker = MainPanicHandler.init(log);
-        MainPanicHandler.Singleton = &panicker;
+        // var panicker = MainPanicHandler.init(log);
```
dylan-conway, line 59: delete

build.zig
```diff
@@ -146,7 +146,7 @@ const fs = std.fs;
 pub fn build(b: *Build) !void {
     build_(b) catch |err| {
         if (@errorReturnTrace()) |trace| {
-            std.debug.dumpStackTrace(trace.*);
+            (std.debug).dumpStackTrace(trace.*);
```
dylan-conway, line 149: zig bug? lol

dylan-conway (commented):
A few comments. Mostly wondering if there are four missing `bun.handleErrorReturnTrace(err, @errorReturnTrace());` in `run_command.zig`

src/cli/run_command.zig
paperclover: they are all at line 533

src/bun.zig
paperclover: i used snake case because the crash_handler does not contain any fields, similar to namespaces like `windows`, `path`, `sys` (and, the zig standard library does this naming practice too). but we have many exceptions, like Global and Output.

build.zig
paperclover: no, to suppress our linter

Issue #9243, opened 2024-03-04
Implement a lightweight version of test isolation


Jarred-Sumner:
This will be a breaking change

We're going to add a lightweight version of test isolation to the test runner

This will include:
- Reset every mock/spy after each test
- Automatically clear timers via a generation number

mangs:
What's the performance impact of this vs. doing it manually in JS/TS? I would imagine it might be better if it's done in Zig somehow, but I'm just speculating.

Issue #7823, opened 2023-12-25
Restore mock.module using mock.restore not work as expect


yukikwi:
### What version of Bun is running?

1.0.20+09d51486e

### What platform is your computer?

Linux 6.5.0-14-generic x86_64 x86_64

### What steps can reproduce the bug?

I create test file with this code inside
```javascript
import axios from "axios"
import { describe, expect, it, mock } from "bun:test"

describe("topic", () => {
  it("test case 1", async () => {
    // before mock
    expect((await axios("https://google.com/", { method: "GET" })).status).toBe(200)

    // mock
    mock.module("axios", () => ({
      default: () => {
        return {
          status: 500
        } 
      }
    }));

    // after mock
    expect((await axios("https://google.com/", { method: "GET" })).status).toBe(500)

    // restore mock & test again
    mock.restore()
    expect((await axios("https://google.com/", { method: "GET" })).status).toBe(200)
  })
})
```

### What is the expected behavior?

I expect this test case must pass

### What do you see instead?

// before mock
expect((await axios("https://google.com/", { method: "GET" })).status).toBe(200) -> this pass

// after mock
expect((await axios("https://google.com/", { method: "GET" })).status).toBe(500) -> this pass

// restore mock & test again
expect((await axios("https://google.com/", { method: "GET" })).status).toBe(200) -> this fail (value is 500)

### Additional information

This one work like what I expected but I think it is not good solution

```javascript
import axios from "axios"
import path from "path"
import { describe, expect, it, mock } from "bun:test"

describe("topic", () => {
  it("test case 1", async () => {
    // before mock
    expect((await axios("https://google.com/", { method: "GET" })).status).toBe(200)

    // mock
    const axiosBak = axios  <-- add this
    mock.module("axios", () => ({
      default: () => {
        return {
          status: 500
        } 
      }
    }));

    // after mock
    expect((await axios("https://google.com/", { method: "GET" })).status).toBe(500)

    // restore mock & test again
    mock.restore()
    mock.module("axios", () => axiosBak) <-- add this
    expect((await axios("https://google.com/", { method: "GET" })).status).toBe(200)
  })
})
```

chlorophant:
Not only is the restore not working, modules that are mocked will conflict with other testfiles that also need to mock the given module. Not sure what the path forward is for this but its a major blocker

jpasquers:
This is the workaround I have been using in the meantime:

```
import { mock } from "bun:test";

export type MockResult = {
[...]

Issue #7376, opened 2023-11-30
Implement the Ability to Revert the Mocking of Modules


constraintAutomaton:
### What is the problem this feature would solve?

In the current Bun test package, it is possible to mock a whole modules using the function `mock.module` (I don't think this feature is documented...). But it doesn't seem like it is possible to revert to the original implementation. I think in multiple situations this feature is needed and popular frameworks like Jest, who seem to be an inspiration for the test package of Bun, implement it.

### What is the feature you are proposing to solve the problem?

A `mock.revertAllMock` function and an object return from the `mock.module` function with a method to revert to the specific module to its original implementation.

### What alternatives have you considered?

I don't have an other idea.

chlorophant:
Additionally mocked modules will conflict with other testfiles that need to mock the same module

Jarred-Sumner:
For now, try the following:
```js
import {jest} from 'bun:test';

jest.restoreAllMocks();
```

grant-heath:
@Jarred-Sumner jest.restoreAllMocks() does not work for me.

cwqt:
This was also affecting me, as a work around I settled on getting all the tests files & running them in separate `bun test` runs.

```
find . -path "*test.ts" | xargs -L 1 bun test
```

coodoo:
Found a working solution by using `delete require.cache[]` to reset mocked modules after each test case, confirmed working across multiple test files and cases. Detailed steps below.

```js

// 1. placeholder to persist mocked modules
globalThis.mocked = []

// 2. do the mocking
const spied = spyOn(require(`foo_module`), `bar_fn`)
spied.mockReturnValueOnce = `response 1`
spied.mockReturnValue = `response default`

// 2.1. save the mocked module name so that it can be restored later
globalThis.mocked.push(`foo_module`)

// 3. after each test case, restore the mocked module
globalThis.mocked.forEach(name => {
  delete require.cache[require.resolve(name)]
})
```

@Jarred-Sumner might as well consider implementing something similar to `delete require.cache[]` in bun as a quick solution to this issue? 😉

PR #10210, opened 2024-04-12
Implement generational test isolation in `bun:test`

Jarred-Sumner:
### What does this PR do?

WIP. Towards #9243 

Adds `--isolate=lite` to `bun test`

After each `test()` ends:
- `kill()` all `Subprocess` and `Bun.$` created during the test
- Close all `Bun.serve()`, `Bun.connect()`, `Bun.listen()` created during the test

After each `.test.*` file finishes running:
- `kill()` all `Subprocess` and `Bun.$` created since the module started 
- Close all `Bun.serve()`, `Bun.connect()`, `Bun.listen()` created since the module started 

Before `bun test`'s process exits:
- `kill()` all `Subprocess` and `Bun.$`
- Close all `Bun.serve()`, `Bun.connect()`, `Bun.listen()`

##### How it works

Each test scope gets an (incremented) generation ID.  I/O resources add themselves to the list of active resources and add the current generation ID. Each time we finish a test, a module, or a test scope we run a cleanup function for each I/O resource that has a generation ID >= the current generation ID. 

TODO:
- [ ] Write tests for each type
- [ ] Store a pointer to `ActiveSpySet` in `DescribeScope` and `TestScope`, associate all mocks with these and restore all these mocks created per scope

### DIFF
--- a/src/bun.js/api/BunObject.zig
+++ b/src/bun.js/api/BunObject.zig
@@ -2814,12 +2814,13 @@ pub fn serve(
     globalObject: *JSC.JSGlobalObject,
     callframe: *JSC.CallFrame,
 ) callconv(.C) JSC.JSValue {
+    const vm = globalObject.bunVM();
     const arguments = callframe.arguments(2).slice();
     var config: JSC.API.ServerConfig = brk: {
         var exception_ = [1]JSC.JSValueRef{null};
         const exception = &exception_;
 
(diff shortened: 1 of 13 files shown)


MonsterDeveloper:
This would be awesome. Currently mock restoration doesn't work with `mock.module` (see #7823 and #7376), and this is a big blocker for using Bun as a test runner.

Hopefully test isolation fixes this issue.

Jarred-Sumner:
Closing due to bitrot. But would like to revisit in the future.

Issue #10177, opened 2024-04-11
Regression: response.write with non-ascii latin1 causes duplicated character or segfault


L3P3:
### What version of Bun is running?

1.1.1 .. 1.1.3 (baseline)

### What platform is your computer?

Linux 6.6.15-amd64 x86_64 unknown

### What steps can reproduce the bug?

```js
require('http').createServer((_,response) => {
	response.write('á' + '-'.repeat(254) + 'x');
	response.write('');
	response.end();
}).listen(80);
```

### What is the expected behavior?

á--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------x

### What do you see instead?

á--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------xx

### Additional information

Works in v1.1.0.
The written string containing the unicode character needs to be at least 256 characters long and there needs to be at least one other `.write('')` call or a string passed to `.end()`.

L3P3:
And while playing with it, I got this once:

```
kernel: Bun Pool 5[3417145]: segfault at 6e78 ip 0000561eef93e605 sp 00007f514dd82690 error 4 in bun-1.1.3[561eeed0f000+2dcc000] likely on CPU 6 (core 3, socket 0)
kernel: Code: 77 48 b9 f8 ff ff ff ff ff 01 00 49 21 ce 48 b9 00 00 00 00 00 00 f2 07 4c 09 f1 48 89 08 48 c7 40 08 00 00 00 00 c6 40 10 01 <49> 8b 8f 78 6e 00 00 f0 48 ff 41 58 48 c7 40 08 00 00 00 00 48 89
```

PR #10221, opened 2024-04-12
fix(stream) fix http body-stream sending duplicate data

cirospaciari:
### What does this PR do?
Fix: #10177 

This PR supersedes https://github.com/oven-sh/bun/pull/10205

- [x] Code changes

Added a test

### DIFF
--- a/test/js/node/http/node-http.test.ts
+++ b/test/js/node/http/node-http.test.ts
@@ -1746,3 +1746,37 @@ if (process.platform !== "win32") {
     expect([joinPath(import.meta.dir, "node-http-ref-fixture.js")]).toRun();
   });
 }
+
+it("#10177 response.write with non-ascii latin1 should not cause duplicated character or segfault", async () => {
+  const expected = "á" + "-".repeat(254) + "x";
+  const server = require("http")
+    .createServer((_, response) => {
+      response.write("á" + "-".repeat(254) + "x");
+      response.write("");
+      response.end();
+    })
+    .listen(0, "localhost", async (err, hostname, port) => {
+      try {
+        expect(err).toBeFalsy();
+        expect(port).toBeGreaterThan(0);
+        const url = `http://${hostname}:${port}`;
+        const count = 1000;
+        const all = [];
+        const batchSize = 20;
+        while (all.length < count) {
+          const batch = Array.from({ length: batchSize }, () => fetch(url).then(a => a.text()));
+
+          all.push(...(await Promise.all(batch)));
+        }
+
+        for (const result of all) {
+          expect(result).toBe(expected);
+        }
+      } finally {
+        server.close();
+        Bun.gc(true);
+      }
+    });
+  await Bun.sleep(1000);
+  Bun.gc(true);
+});
--- a/test/js/node/http/node-http.test.ts
+++ b/test/js/node/http/node-http.test.ts
@@ -1758,8 +1758,10 @@ it("#10177 response.write with non-ascii latin1 should not cause duplicated char
   // 128 = small than waterMark, 256 = waterMark, 1024 = large than waterMark
   // 8Kb = small than cork buffer
   // 16Kb = cork buffer
-  // 64Kb = large than cork buffer
-  const sizes = [128, 256, 1024, 8 * 1024, 16 * 1024, 64 * 1024];
+  // 32Kb = large than cork buffer
+  const start_size = 128;
+  const increment_step = 1024;
+  const end_size = 32 * 1024;
   let expected = "";
 
   function finish(err) {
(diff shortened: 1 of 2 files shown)


test/js/node/http/node-http.test.ts
```diff
@@ -1746,3 +1746,37 @@ if (process.platform !== "win32") {
     expect([joinPath(import.meta.dir, "node-http-ref-fixture.js")]).toRun();
   });
 }
+
+it("#10177 response.write with non-ascii latin1 should not cause duplicated character or segfault", async () => {
+  const expected = "á" + "-".repeat(254) + "x";
```
Jarred-Sumner: Please add a test that checks for N variations of sizes
```suggestion
for (let encodedChar of ["á", "📙", "👍🏽"] {
for (let repeatCount = 0; repeatCount < 512; repeatCount++) {
  const expected = encodedChar + "-".repeat(repeatCount) + "x";
}
...
```

test/js/node/http/node-http.test.ts
```diff
@@ -1746,3 +1746,37 @@ if (process.platform !== "win32") {
...
+
+        for (const result of all) {
+          expect(result).toBe(expected);
+        }
+      } finally {
+        server.closeAllConnections();
+        Bun.gc(true);
+      }
+    });
+  await Bun.sleep(1000);
```
Jarred-Sumner: This sleep should not be necessary

test/js/node/http/node-http.test.ts
```diff
@@ -1746,3 +1746,37 @@ if (process.platform !== "win32") {
...
+      response.write(expected);
+      response.write("");
+      response.end();
+    })
+    .listen(0, "localhost", async (err, hostname, port) => {
+      try {
+        expect(err).toBeFalsy();
+        expect(port).toBeGreaterThan(0);
+        const url = `http://${hostname}:${port}`;
+        const count = 1000;
```
Jarred-Sumner: I think we can safely reduce the count if we add iterations above, probably 20 is fine

test/js/node/http/node-http.test.ts
Jarred-Sumner: The reason for 👍🏽 is its a grapheme of 👍 🟤  

The reason for `á` is its a latin1 supplemental character

cirospaciari: variations with some relevant sizes https://github.com/oven-sh/bun/pull/10221/commits/c7afd02ad7c4fe70669c3b07fc899aa4a5281990

test/js/node/http/node-http.test.ts
```diff
@@ -1746,3 +1746,54 @@ if (process.platform !== "win32") {
...
+  // 8Kb = small than cork buffer
+  // 16Kb = cork buffer
+  // 64Kb = large than cork buffer
+  const sizes = [128, 256, 1024, 8 * 1024, 16 * 1024, 64 * 1024];
+  for (const char of chars) {
+    for (const size of sizes) {
+      const expected = char + "-".repeat(size) + "x";
+      const { promise, resolve } = Promise.withResolvers();
+      const server = require("http")
+        .createServer((_, response) => {
```
Jarred-Sumner: don't need to create a new server each time

test/js/node/http/node-http.test.ts
```diff
@@ -1746,3 +1746,54 @@ if (process.platform !== "win32") {
...
+              all.push(...(await Promise.all(batch)));
+            }
+
+            for (const result of all) {
+              expect(result).toBe(expected);
+            }
+          } finally {
+            server.closeAllConnections();
+            Bun.gc(true);
+            resolve();
```
Jarred-Sumner: this test will never fail on error

it should reject

test/js/node/http/node-http.test.ts
```diff
@@ -1746,3 +1746,54 @@ if (process.platform !== "win32") {
...
+  // 📙 = emoji
+  // 👍🏽 = its a grapheme of 👍 🟤
+  // "\u{1F600}" = utf16
+  const chars = ["x", "á", "📙", "👍🏽", "\u{1F600}"];
+
+  // 128 = small than waterMark, 256 = waterMark, 1024 = large than waterMark
+  // 8Kb = small than cork buffer
+  // 16Kb = cork buffer
+  // 64Kb = large than cork buffer
+  const sizes = [128, 256, 1024, 8 * 1024, 16 * 1024, 64 * 1024];
```
Jarred-Sumner: maybe count + 1024 or so for each size

test/js/node/http/node-http.test.ts
L3P3: Do we even need to repeat same tests? It seems like fuzzing to me. If we test for coincidences, errors should be caught by repeating such tests in total or by them just being run hundreds of times in future, I think.

Jarred-Sumner: We want to check that repeatedly allocating memory of slightly different sizes and types produces the correct results every time. GC + network adds a lot of unpredictability so it helps to run the same or similar code repeatedly

PR #10231, opened 2024-04-12
fix create with github URL on windows

gvilums:
### What does this PR do?
Replaces some uses of zig stdlib `accessat` (which were running into unreachable) with `bun.sys.existsAt`.
Fixes #10183 

### How did you verify your code works?
Manual testing

PR #10236, opened 2024-04-13
Revert "fix create with github URL on windows"

Jarred-Sumner:
Reverts oven-sh/bun#10231

windows test failed, should not have merged this


Jarred-Sumner:
cc @gvilums

Issue #2537, opened 2023-04-03
Database.prepare without .finalize results in left-over wal file


betamos:
### What version of Bun is running?

0.5.8

### What platform is your computer?

Linux 5.15.0-69-generic x86_64 x86_64

### What steps can reproduce the bug?

Run this bun program at least twice:

```ts
import { Database } from "bun:sqlite";
const db = new Database("dirty.sqlite", { create:true })
const prep = db.prepare(`SELECT "hello"`)
db.exec("pragma journal_mode = wal")

// NOTE: If this line is uncommented, the `shm` and `wal` files are removed on exit
// prep.finalize()

db.close()
```

### What is the expected behavior?

Normally, the `shm` and `wal` files are cleaned up on close. I suppose that's what's expected, but.. if manual `finalize` is required then it should probably be documented.

### What do you see instead?

```sh
$ ls | grep dirty
dirty.sqlite
dirty.sqlite-shm
dirty.sqlite-wal
```

sayalovesu:
it looks like calling `.finalize()` on `prep` doesn't work if there happened any writes:

```js
import { Database } from 'bun:sqlite'

const db = new Database('test.db')
db.exec('PRAGMA journal_mode = WAL')

// or simply db.exec('CREATE TABLE test (id INTEGER PRIMARY KEY)')
const prep = db.prepare('CREATE TABLE test (id INTEGER PRIMARY KEY)')
prep.run()
prep.finalize()

db.close()
```

```sh
❱ bun test.js
❱ ls test.db*                                                            
test.db     test.db-shm test.db-wal
```

and in the code in the op, the `.finalize()` call doesn't seem to matter either anymore, btw. 
has something changed in the last year?

doroved:
> and in the code in the op, the `.finalize()` call doesn't seem to matter either anymore, btw.
> has something changed in the last year?

I too have encountered this problem and wal_checkpoint doesn't help.

```ts
process.on('SIGINT', (code) => {
  db.exec('PRAGMA wal_checkpoint(TRUNCATE);')
  db.close()
  process.exit(1)
})
```

doroved:
It might be worth checking this information out
![image](https://github.com/oven-sh/bun/assets/30116541/eee48cfc-97f9-44b9-b3dd-b8f492b74a65)

PR #10262, opened 2024-04-14
[bun:sqlite] Support `sqlite3_file_control`, better closing behavior, implement Disposable

Jarred-Sumner:
### What does this PR do?

You can now manage the lifetime of a `Database` instance in `bun:sqlite` via the `using` keyword:

```ts

{
  using db = new Database("mydb.sqlite");
  using query = db.query("select 'Hello world' as message;");
  console.log(query.get()); // => { message: "Hello world" }
}
```

`Database#fileControl` lets you use the `sqlite3_file_control` API

For example:

```ts

const db = new Database();
// Ensure WAL mode is NOT persistent
// this prevents wal files from lingering after the database is closed
db.fileControl(SQL.constants.SQLITE_FCNTL_PERSIST_WAL, 0);
```

Automatically closing databases when the `Database` object GC'd was not implemented correctly before (bun would only automatically close the database before process exit). This has now been fixed

`database.close(true)` will now throw an error if the database has any pending resources associated with it.

Fixes https://github.com/oven-sh/bun/issues/2537

Tests

### DIFF
--- a/docs/api/sqlite.md
+++ b/docs/api/sqlite.md
@@ -417,12 +417,12 @@ db.loadExtension("myext");
 To use the advanced `sqlite3_file_control` API, call `.fileControl(cmd, value)` on your `Database` instance.
 
 ```ts
-import { Database } from "bun:sqlite";
+import { Database, constants } from "bun:sqlite";
 
 const db = new Database();
 // Ensure WAL mode is NOT persistent
 // this prevents wal files from lingering after the database is closed
-db.fileControl(SQL.constants.SQLITE_FCNTL_PERSIST_WAL, 0);
+db.fileControl(constants.SQLITE_FCNTL_PERSIST_WAL, 0);
 ```
 
 `value` can be:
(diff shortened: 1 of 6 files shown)


gvilums (approved):

pekeler:
This doesn't work well for me in 1.1.4. Typescript doesn't know about the SQLITE_FCNTL_* constants, the fileControl() function, the close() parameter, and is complaining about the db not having a dispose() method when using 'using'.
image

gvilums:
@pekeler this might be an issue with our type declarations - have you tried running the code and seeing if it works?

pekeler:
@gvilums the code works for simple use cases (ignoring the type errors). In scenarios where the DB is used for a bit more than one simple query, `db.close(true);` produces a `database is locked` error. I do keep track of all my prepared statements and make sure they're finalized before closing the DB, so something doesn't feel right.

gvilums:
Hm, that does indeed sound like something is going wrong in bun. Could you open an issue for this?

pekeler:
@gvilums I have so far failed to create a simple project to reproduce this issue. However, I also noticed that `db.close(true)` doesn't work after running migrations and created an issue with a simple project that reproduces the issue here https://github.com/oven-sh/bun/issues/11418. Hopefully the underlying cause is the same.

PR #10266, opened 2024-04-14
fix(HTTP Server): can't set custom statusText

coratgerl:
### What does this PR do?

On node we can change the statusText in the Response. On Bun, ```Bun.serve``` and ```createServer``` from node:http only returns the default statusText define here https://github.com/oven-sh/bun/blob/main/src/bun.js/api/server.zig#L1039.

So actually the two following code doesn't change the statusText

```js
Bun.serve({
  fetch(request) {
    return new Response(request.url, {
      statusText: "customStatusText",
    });
  },
  port: 0,
});

```

```js
import { createServer } from 'node:http'

const server = createServer((req, res) => {
	res.writeHead(200, 'statusMessage')
	res.end('Hello World\n')
})

server.listen(3000)

```

- [X] Code changes

I wrote automated tests


coratgerl:
@Jarred-Sumner This issue is problematic for my use case, can you review this one if you have the time ? 🙏 

Thanks for all the work done !

lanezhao:
I have the same issue. Please please fix it. It's a basically need to be when create http server. @Jarred-Sumner

`
const port = 4444;

bun.serve({
  port,
  fetch() {
    return new Response('', {
      statusText: 'customStatusText',
    });
  },
});

const host = `http://localhost:${port}`;

await fetch(`${host}`)
  .then((res) => {
    console.debug({ res, statusText: res.statusText });
  });
`

lanezhao:
@Jarred-Sumner I think this code change has not been merged correctly. Problems have persisted in the last few versions. I need to use "custom statusText" in bun.js, for my open source framework. Please fix it.

lanezhao:
Can sameone respond? This is a major issue when creating web servers with Bun.js. I think it's not so difficult to fix. @Jarred-Sumner @dylan-conway @coratgerl Please!!!

Issue #10265, opened 2024-04-14
🚰🚰 Bun.serve: Memory leak HTTP POST with Body [Bun 1.0.31+ version]


nmdaniil:
### What version of Bun is running?

from 1.0.31 to latest (1.1.3)

### What platform is your computer?

MacOs Darwin 23.4.0 x86_64 i386

### What steps can reproduce the bug?

here is a script with a test where the leak is caught

```ts
import { heapStats } from "bun:jsc"

// Log Memory utils:
let lastMemoryLogAt = 0
let lastMemoryMb = 0
let lastReqs = 0
let reqs = 0
const logMemoryEverySeconds = (opt?: { force?: boolean }) => {
  reqs++
  const now = new Date().getTime()
  if (now - lastMemoryLogAt > 1000 || opt?.force) {
    const memoryMb = (process.memoryUsage.rss() / 1024 / 1024).toFixed(3)
    const stat = heapStats()
    const myStat = {
      heapSize: (stat.heapSize / 1024 / 1024).toFixed(3),
      heapCapacity: (stat.heapCapacity / 1024 / 1024).toFixed(3),
      extraMemorySize: (stat.extraMemorySize / 1024 / 1024).toFixed(3),
      objectCount: (stat.objectCount / 1000).toFixed(1)
    }
    const _reqsDelta = reqs - lastReqs
    const reqsDelta = _reqsDelta >= 1000 ? (_reqsDelta / 1000).toFixed(1) + 'k' : _reqsDelta
    const delta = +memoryMb - lastMemoryMb
    const plusSign = delta > 0 ? '+' : ''
    console.log(`MemUse ${memoryMb} mb (${plusSign}${delta.toFixed(3)} mb)` +
      ` | heap ${myStat.heapSize} mb | heapCap ${myStat.heapCapacity} mb` +
      ` | extMem ${myStat.extraMemorySize} mb | objs ${myStat.objectCount}k` +
      ` | reqs ${reqsDelta}`)
    lastMemoryLogAt = now
    lastMemoryMb = +memoryMb
    lastReqs = reqs
  }
}


//
// Test utils:
//
let startAt = new Date().getTime()
const hasPassedSeconds = (sec: number) => {
  const now = new Date().getTime()
  if ((now - startAt) < (1000 * sec)) return true
  else {
    startAt = now
    return false
  }
}


//
// Start Bun server:
//
const bunPort = 3001
Bun.serve({
  fetch(req: Request) {
[...]

PR #10268, opened 2024-04-14
fix(server) handle ignored body and node:http destroy

cirospaciari:
### What does this PR do?

Fix: #10265 
Fix: #10665

- [x] Code changes


src/bun.js/webcore/body.zig
```diff
@@ -842,7 +843,16 @@ pub const Body = struct {
...
+                    if (readable.ptr == .Bytes) {
+                        // we are streaming so we inform the error to it
+                        readable.ptr.Bytes.onData(
+                            .{
+                                .err = .{ .JSValue = error_instance },
+                            },
+                            bun.default_allocator,
+                        );
+                    }
+                    readable.abort(global);
```
Jarred-Sumner, line 846: I think just calling `abort` should be enough

cirospaciari: abort/cancel actually was only canceling the stream not the source, so I fixed that now.
onCancel actually only mark as done, not actually emits a error so I changed the test that I add to match this expectations.

test/js/node/http/node-http.test.ts
```diff
@@ -1904,3 +1904,43 @@ it.skipIf(!process.env.TEST_INFO_STRIPE)("should be able to connect to stripe",
...
+        chunks++;
+      });
+    });
+    req.end();
+    // wait for the first chunk
+    await promise;
+    // should stop the download
+    req.destroy();
+    await Bun.sleep(100);
+    expect(chunks).toBeLessThan(10);
```
oguimbal: Hi ! Not sure if you take external reviews, but I'll comment anyway hoping that you do :)

Given that the server emits less than 1 payload per 10ms, emitting less than 10 chunks in 100ms seems very likely to me, even if the download doesnt stop.

... meaning that this test is likely to give a false positive or to be flaky in case of a regression. Or am I missing something ?

cirospaciari: indeed, I edited it when testing and forgot to edit back, will update

PR #10287, opened 2024-04-15
Don't error on comments & trailing commas in package.json

Jarred-Sumner:
### What does this PR do?

This makes `bun install`, `bun run <script>`, `bun ./<file>` , `bun build <file>` no longer error when package.json contains trailing commas, single-line comments or multi-line comments.

We do not recommend this feature, as it will break usage in other tools (as well as `JSON.parse`). But enough people have asked for it and it's very easy to add. We trust people can make their own decision whether or not it's worth the tradeoffs.

Fixes #9036 

There are tests


chribjel:
Maybe consider  adding a warning in the console, since the feature is not recommended and is not spec compliant.

`WARN: You are using comments or trailing commas in your package.json. This is not spec-compliant and might break other tooling.`

Lms24:
Have you considered tooling that reads a user's `package.json` file? At Sentry, we maintain a [wizard](https://github.com/getsentry/sentry-wizard) that installs and configures our SDKs in users' projects (basically users just run `npx @sentry/wizard`). The wizard `JSON.parse`s the user's `package.json` to check if our dependencies were already installed. Of course we catch if parsing throws and warn users about it but it'll lead to problems/early returns in our setup flow. 

It's probably reasonable to move to a dynamic `import` instead of `JSON.parse` but I'd argue we're by far not the only tooling that relies on parsing `package.json` this way. All these tools can break or turn to erroneous code paths if they suddenly start encountering comments in `package.json`.

src/json_parser.zig
```diff
@@ -794,11 +802,47 @@ pub fn ParseJSONUTF8AlwaysDecode(
...
+    const len = source.contents.len;
+
+    switch (len) {
+        // This is to be consisntent with how disabled JS files are handled
+        0 => {
+            return Expr{ .loc = logger.Loc{ .start = 0 }, .data = empty_object_data };
+        },
+        // This is a fast pass I guess
+        2 => {
+            if (strings.eqlComptime(source.contents[0..1], "\"\"") or strings.eqlComptime(source.contents[0..1], "''")) {
```
dylan-conway, line 832: `[0..2]` for all of these

test/bundler/esbuild/packagejson.test.ts
```diff
@@ -90,10 +69,55 @@ describe("bundler", () => {
         }
       `,
     },
-    bundleErrors: {
-      "/Users/user/project/node_modules/demo-pkg/package.json": ["JSON does not support trailing commas"],
+    run: {
+      stdout: "123",
     },
   });
+  // itBundled("packagejson/SyntaxErrorComment", {
```
dylan-conway, line 76: should we just delete these?

dylan-conway (approved):
LGTM

miscellaneo:
Consider the following package.json.
```json
{
  "devDependencies": {
    "@company-repo/prettier-config": "1.0.0",
    // We can't use the new version until the prettier config supports it.
    "prettier": "2.8.8"
  },
  "name": "@company-repo/my-package",
  "prettier": "@company-repo/prettier-config",
  "scripts": {
    "format": "prettier --write ."
  }
}
```
As it is now, `bun run format` would be an error, because there's a comment in the JSON.  With this PR, Bun can run the script, but it would still error out, because Prettier will try to read this package.json, but fail because there's a comment in the JSON.  Now, most users will probably figure "okay, Bun allows this, but Prettier doesn't, so I can't use this feature", erase the comment, and move on.  But the problem is that, for a tool that has seen as much adoption as Bun has, there will still be a fair number of users who will see the error and think "okay, Bun allows this, but Prettier doesn't, so I should file an issue with Prettier".

You can probably remember something like this playing out when Bun hit 1.0, where something about Bun's implementation meant certain packages didn't work, and users would file issues with the repo for those packages.  It wasn't what Bun intended, users were instructed to file issues with Bun instead, but it still happened because that's the conclusion some users are going to come to.

The difference here is that it's not a bug, it's a feature.  The behavior that causes the script to crash will continue, because there's nothing for Bun to fix or implement that will make it stop.  So users will keep running into this, not just with Prettier but whenever they try to use this feature along with a package that doesn't parse JSON with comments, and some of them will continue to come to the conclusion that it is an issue with those packages.

You could say that Prettier could avoid this by allowing package.json files to have comments, which would probably be easier for it to do that most packages since, like Bun, it already has code to handle JSON with comments, but that just sort of spreads the problem. Because now it's "okay, Bun and Prettier allow this, but such-and-such doesn't, so I should file an issue with such-and-such", and now such-and-such has to either put effort into getting config to read from JSON with comments or just put up with the extra noise on the issue tracker.

You can say

> But enough people have asked for it and it's very easy to add.

but please remember the words of Tony Hoare, "I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement."  The trouble this change will cause will be nowhere near the magnitude of issues stemming from that, but it is not nothing, either.  It will add friction, going from Node to Bun or back again.  It will add tension between Bun and package maintainers.

test/bundler/esbuild/packagejson.test.ts
Jarred-Sumner: yeah, will do in a follow-up

Issue #10080, opened 2024-04-08
stdin redirection hangs on macOS


alexkornitzer:
Since Bun v1.0.31 which had the major IO rework, talking to Bun via an Erlang Port (https://www.erlang.org/doc/tutorial/c_port) has resulted in hangs on macOS. After investigation the issue appears to come from assumptions that Bun makes in its `FileReader` (https://github.com/oven-sh/bun/blob/main/src/bun.js/webcore/streams.zig#L3441), resulting in it setting the reader as non blocking when the file descriptor is not. This then results in a hang as the `PosixBufferedReader` handles reads in a non blocking way while the read syscall is blocking, thus `EAGAIN` is never raised.

I have used a rather forceful approach in local testing to fix the issue which just ensures the file descriptor is non blocking (using the already present but what appears to be unused function).

```patch
diff --git a/src/bun.js/webcore/streams.zig b/src/bun.js/webcore/streams.zig
index 7703efa21..201d21a83 100644
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -3525,6 +3525,11 @@ pub const FileReader = struct {
                 if (this.nonblocking and this.file_type == .pipe) {
                     this.file_type = .nonblocking_pipe;
                 }
+
+                // ensure non blocking otherwise hangs will occur
+                if (this.nonblocking) {
+                    bun.ensureNonBlocking(fd.int());
+                }
             }

             this.fd = fd;
```

I have not raised a PR for this as I understand the fix might be to handle the broken assumptions for macOS builds rather than forcing the descriptor to be non blocking.

The following was used to reproduce the error:

#### test.js
```javascript
for await (const line of console) {
  console.log(`line: ${line}`);
}
```

#### wrapper.sh
```bash
#!/usr/bin/env bash

exec &> >(./build/bun-debug ./test.js)

echo "I printed...!"

while read -r line; do echo $line; done
```

postnerd:
I can also reproduce this behavior using Bun as the runtime for my chess engine, [postbot](https://github.com/postnerd/postbot/tree/bunjs).

Since Bun v1.0.31, the engine has stopped working with BanksiaGUI or the [lichess-bot wrapper](https://github.com/lichess-bot-devs/lichess-bot). According to the logs, it appears that stdin stops working after the first input.

HuakunShen:
I have a similar issue with Tauri.
[...]

PR #10303, opened 2024-04-16
fix(posix): ensure nonblocking file descriptor is nonblocking (#10080)

alexkornitzer:
### What does this PR do?

The FileReader is responsible for deciding what sort of file descriptor has been passed to it. On MacOS, when stdin is provided to it the FileReader assumes that it is nonblocking even if the descriptor is not, this resulted in hangs as the socket would never raise the `EAGAIN` error. There are a couple of ways to fix this but the one that seemed to make the most sense was to set the fd to be nonblocking when the FileReader deemed it so.

- [X] Code changes

I created a new regression test for the specific edge case.

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -516,6 +516,9 @@ pub fn ensureNonBlocking(fd: anytype) void {
 }
 
 pub fn isNonBlocking(fd: anytype) bool {
+    if (Environment.isWindows) {
+        @compileError("isNonBlocking() is not supported on Windows");
+    }
     return (std.os.fcntl(fd, std.os.F.GETFL, 0) catch 0 & std.os.O.NONBLOCK) == std.os.O.NONBLOCK;
 }
 
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -515,6 +515,10 @@ pub fn ensureNonBlocking(fd: anytype) void {
     _ = std.os.fcntl(fd, std.os.F.SETFL, current | std.os.O.NONBLOCK) catch 0;
 }
 
+pub fn isNonBlocking(fd: anytype) bool {
+    return (std.os.fcntl(fd, std.os.F.GETFL, 0) catch 0 & std.os.O.NONBLOCK) == std.os.O.NONBLOCK;
+}
+
 const global_scope_log = sys.syslog;
 pub fn isReadable(fd: FileDescriptor) PollFlag {
     if (comptime Environment.isWindows) {
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -3562,7 +3562,7 @@ pub const FileReader = struct {
                     this.file_type = .nonblocking_pipe;
                 }
 
-                this.nonblocking = is_nonblocking_tty or (this.pollable and !(file.is_atty orelse false));
+                this.nonblocking = is_nonblocking_tty or bun.isNonBlocking(fd.int());
 
                 if (this.nonblocking and this.file_type == .pipe) {
                     this.file_type = .nonblocking_pipe;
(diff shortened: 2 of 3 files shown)


Jarred-Sumner (changes requested):
Thank you for this

Unfortunately, we cannot change the stdin file descriptor to be non-blocking or it will break a process which is piped into Bun (like `cat foo | bun bar.js`) when enough input has been written to cause `EAGAIN`

What we do instead:
1. If stdin is a TTY, we use [`ttyname_r`](https://github.com/oven-sh/bun/blob/jarred/packagejson/src/bun.js/bindings/c-bindings.cpp#L580-L593) to re-open stdin as non-blocking without being observable to the other process
2. If stdin is a pipe, it should be treated as blocking and handled accordingly
3. If stdin is a regular file, we do not poll it since that is not supported anyway
4. If stdin is a socket, we use pass `MSG_NOWAIT` to `recv` and `send`
5. If stdin is a pipe on Linux and `pwritev2` is supported, we use the flag to only write if it can do so non-blockingly

Furthermore, when spawning processes, we use a socket instead of a blocking pipe which avoids this case

That leaves 2 as the likely scenario where this is happening. So the next step here is to figure out why `nonblocking` is being set to `true` despite this happening.

alexkornitzer:
Hey @Jarred-Sumner,

Sorry to force your hand to reply by creating a PR, was hoping for this feedback in the original issue but am aware that you have a flood of them with no labels for prioritisation and/or triaged status.

If `2` is the intended behaviour then it will be an incorrect assumption in that function so am happy to dig that out.

I spawned the process with a pipe as I needed to replicate the regressive behaviour I was experiencing when using Bun `1.0.31`+ via Erlang.

alexkornitzer:
Did some lazy debugging and the issue here is the assumption that a pollable non atty is nonblocking, but the point of truth here is on the descriptor via the O_NONBLOCK flag. 

`[SYS] is_nonblocking_tty: false, pollable: true, file.is_atty: false, O_NONBLOCK: false`

So if mutating the descriptor is unsafe the only other solution I can think of here with my very basic understanding is to just check the flag as gospel. Which was going to be my other approach to fix this but I had assumed the descriptor was being forced to nonblocking for a reason hence my original approach.

src/bun.zig
```diff
@@ -514,6 +514,10 @@ pub fn ensureNonBlocking(fd: anytype) void {
     _ = std.os.fcntl(fd, std.os.F.SETFL, current | std.os.O.NONBLOCK) catch 0;
 }
 
+pub fn isNonBlocking(fd: anytype) bool {
+    return (std.os.fcntl(fd, std.os.F.GETFL, 0) catch 0 & std.os.O.NONBLOCK) == std.os.O.NONBLOCK;
```
gvilums, line 522:
```suggestion
    if (Environment.isWindows) {
        @compileError("isNonBlocking is not supported on windows");
    }
    return (std.os.fcntl(fd, std.os.F.GETFL, 0) catch 0 & std.os.O.NONBLOCK) == std.os.O.NONBLOCK;
```

alexkornitzer: Thanks for looking at this, do you need any action from me? Would you like me to rebase against master and apply the above suggestion?

gvilums: yeah that would be great 👍

alexkornitzer: Done, I tried to align the formatting to that of other `@compileError` functions within the file. I also squashed the formatting commit into the original commit to keep the git history clean.

alexkornitzer:
Hey team, I understand that this bug may not be very high priority due to the number of people that it is or maybe more accurately is not affecting, but would it be possible for a decision to be made regarding this PR?

Jarred-Sumner:
I believe the intent of this PR was fixed in https://github.com/oven-sh/bun/pull/20152/files

PR #10310, opened 2024-04-17
node:*: fix some typescript errors

nektro:
notable fixes

- missing validators
- missing primordials
- primordials defined wrong
- missing identifiers
- missing builtins
- missing type definitions
- duplicate definitions clobbering names
- `ArrayIsArray` being called wrong
- passing arguments to functions that don't use them
- misc parameter and array type safety
- error builtins being called wrong


src/js/node/child_process.ts
```diff
@@ -1621,6 +1621,10 @@ function validateString(value, name) {
   if (typeof value !== "string") throw new ERR_INVALID_ARG_TYPE(name, "string", value);
 }
 
+function isInt32(value) {
```
Jarred-Sumner, line 1623: is this a builtin? I think it's a builtin

src/js/node/diagnostics_channel.ts
```diff
@@ -3,10 +3,10 @@
 const SafeMap = Map;
 const SafeFinalizationRegistry = FinalizationRegistry;
 
-const ArrayPrototypeAt = (array, index) => array[index];
-const ArrayPrototypeIndexOf = (array, value) => array.indexOf(value);
-const ArrayPrototypePush = (array, value) => array.push(value);
-const ArrayPrototypeSplice = (array, start, deleteCount) => array.splice(start, deleteCount);
+const ArrayPrototypeAt = Array.prototype.at;
+const ArrayPrototypeIndexOf = Array.prototype.indexOf;
+const ArrayPrototypePush = Array.prototype.push;
```
Jarred-Sumner: most of these should actually be `$arrayPush(array, value)`

src/js/node/readline.ts
```diff
@@ -382,7 +386,7 @@ function validateObject(value, name, options = null) {
   var nullable = options?.nullable ?? false;
   if (
     (!nullable && value === null) ||
-    (!allowArray && ArrayIsArray.$call(value)) ||
+    (!allowArray && ArrayIsArray.$call(null, value)) ||
```
Jarred-Sumner:
```suggestion
    (!allowArray && $isJSArray(value)) ||
```

src/js/node/child_process.ts
nektro: node src: https://github.com/nodejs/node/blob/c1d9f426a11eda97/lib/internal/validators.js#L39-L45

PR #10313, opened 2024-04-17
Update platform.ts to fix isWindowsAVX2 implementation.

liudonghua123:
The isWindowsAVX2 function is not working as expected due to the stdout endwith `\r\n`. So the simple `stdout == "True"` will never true.

### What does this PR do?

- [x] Code changes

I tested the following code on my windows, and it works for me.

```ts
// windows-avx2-test.mts
import child_process from "child_process";
import { debug } from "console";

export function spawn(
  cmd: string,
  args: string[],
  options: child_process.SpawnOptions = {},
): {
  exitCode: number;
  stdout: string;
  stderr: string;
} {
  debug("spawn", [cmd, ...args].join(" "));
  const { status, stdout, stderr } = child_process.spawnSync(cmd, args, {
    stdio: "pipe",
    encoding: "utf-8",
    ...options,
  });
  return {
    exitCode: status ?? 1,
    stdout,
    stderr,
  };
}

function isWindowsAVX2(): boolean {
  try {
    const stdout = spawn("powershell", [
        "-c",
        `(Add-Type -MemberDefinition '[DllImport("kernel32.dll")] public static extern bool IsProcessorFeaturePresent(int ProcessorFeature);' -Name 'Kernel32' -Namespace 'Win32' -PassThru)::IsProcessorFeaturePresent(40);`,
    ]).stdout;
    const encoder = new TextEncoder(); 
    console.info(`stdout bytes: ${encoder.encode(stdout)}, True bytes: ${encoder.encode("True")}`)
    return stdout.trim() === "True"
  } catch (error) {
    debug("isWindowsAVX2 failed", error);
    return false;
  }
}

console.info(isWindowsAVX2())
```

```
Liu.D.H  playground   5.389s  09:59 > tsx windows-avx2-test.mts
spawn powershell -c (Add-Type -MemberDefinition '[DllImport("kernel32.dll")] public static extern bool IsProcessorFeaturePresent(int ProcessorFeature);' -Name 'Kernel32' -Namespace 'Win32' -PassThru)::IsProcessorFeaturePresent(40);
stdout bytes: 84,114,117,101,13,10, True bytes: 84,114,117,101
true

Liu.D.H  playground   5.729s  10:03 >
```

### DIFF
--- a/packages/bun-release/src/platform.ts
+++ b/packages/bun-release/src/platform.ts
@@ -111,7 +111,7 @@ function isWindowsAVX2(): boolean {
       spawn("powershell", [
         "-c",
         `(Add-Type -MemberDefinition '[DllImport("kernel32.dll")] public static extern bool IsProcessorFeaturePresent(int ProcessorFeature);' -Name 'Kernel32' -Namespace 'Win32' -PassThru)::IsProcessorFeaturePresent(40);`,
-      ]).stdout == "True"
+      ]).stdout.trim() === "True"
     );
   } catch (error) {
     debug("isWindowsAVX2 failed", error);


liudonghua123:
Thanks for the quick merge, the correct bun binary should be downloaded in the next version via npm install. :smile:

PR #10330, opened 2024-04-17
docs: Remove duplicated content from guides/install/trusted

josephjclark:
### What does this PR do?

There's a duplicated paragraph of text in the trusted docs page, presumably a merge issue somewhere down the line.

I've removed the duplication.

I've also taken the liberty of dropping the note into a callout box, which seems appropriate.

- [ ] Code changes


paperclover:
thank you

PR #10345, opened 2024-04-18
Fixes #10344

### LINKED ISSUE #10344: `bun build --compile` crashes with embedded binary files on Windows

1.1.4

The `dll` file comes from https://github.com/tr1ckydev/webview-bun

```ts
import path from './libwebview.dll' with { type: 'file' };
console.log(path);
```

Things are fine when compiling with `bun build --compile .\test.ts --outfile test`

image

However, bun always crashes when compiling with `bun build --compile --minify --sourcemap .\test.ts --outfile test`

image

Compile and output executable without crashing

### What do you see instead?

Crash when enabling minifier/sourcemap while compiling

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #10344

Test


packages/bun-internal-test/src/runner.node.mjs
```diff
@@ -558,7 +558,7 @@ function emojiTag(tag) {
     emojiText += "";
   }
 
-  if (tag.includes("x86") || tag.includes("_64") || tag.includes("amd64")) {
+  if (tag.includes("x86") || tag.includes("x64") || tag.includes("_64") || tag.includes("amd64")) {
```
paperclover, line 561: why has _64 been a part of this check? that is a very dubious way to check for x86_64.

src/cli/build_command.zig
```diff
@@ -35,6 +35,26 @@ const BundleV2 = @import("../bundler/bundle_v2.zig").BundleV2;
...
+    const compile_define_keys = &.{
+        "process.platform",
+        "process.arch",
+    };
+    const compile_define_values = &.{
+        switch (Environment.os) {
+            .linux => "'linux'",
+            .mac => "'darwin'",
+            .windows => "'win32'",
+            else => "unknown",
```
paperclover: this is a syntax error

paperclover: could consider just using `Environment.os.nameString()` to avoid repeating yourself here. then surround it with quotes

src/cli/build_command.zig
```diff
@@ -35,6 +35,26 @@ const BundleV2 = @import("../bundler/bundle_v2.zig").BundleV2;
...
+        },
+        switch (Environment.isX64) {
```
paperclover: this switch looks weird. i would've just used `switch (builtin.target.cpu.arch)`

make the else a compile error

Issue #9753, opened 2024-03-31
Bun.serve throws an error when using a custom hostname in TLS/HTTPS mode


mangs:
### What version of Bun is running?

1.0.36

### What platform is your computer?

MacOS 14.4.1 (23E224)

### What steps can reproduce the bug?

Here is the relevant portion of code to start the server:

```ts
const server = Bun.serve({
  development: true,
  async fetch(request: Request): Promise<Response> {
    // code here...
  },
  hostname: 'get.babbel.test',
  lowMemoryMode: false,
  port: 443,
  tls: {
    cert: Bun.file('./config/bun/localhost.pem'),
    key: Bun.file('./config/bun/localhost-key.pem'),
  },
});
```

Whether I generate an SSL certificate for `localhost` or a custom domain using `mkcert`, I get the same error.

### What is the expected behavior?

If I remove `hostname` from the `Bun.serve` configuration, the server starts correctly. For example:

```bash
egoldstein@MacCWQXVG6G0V get.apigateway.lambda $ bun --env-file=config/bun/environmentConfig --hot scripts/startDevelopmentServer.mts
Development server listening on https://localhost/
```

### What do you see instead?

Confirmed no other software running on port 443:

```bash
egoldstein@MacCWQXVG6G0V get.apigateway.lambda $ ps
  PID TTY           TIME CMD
10052 ttys001    0:00.66 -bash
egoldstein@MacCWQXVG6G0V get.apigateway.lambda $ bun --env-file=config/bun/environmentConfig --hot scripts/startDevelopmentServer.mts
42 |   console.log(separator);
43 |
44 |   globalThis.hotReloadCount += 1;
45 | }
46 |
47 | const server = Bun.serve({
                    ^
EADDRINUSE: Failed to start server. Is port 443 in use?
 syscall: "listen"
```

Issue #6529, opened 2023-10-16
Multiple TLS certs do not work: KEY_VALUES_MISMATCH


TrackZone1:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Linux 5.10.0-8-amd64 x86_64 unknown

### What steps can reproduce the bug?

```
Bun.serve({
    port: 443,
    async fetch() {
        return new Response("TLS");
    },
    tls: {
        key: [
            Bun.file("/etc/letsencrypt/live/domain1/privkey.pem"),
            Bun.file("/etc/letsencrypt/live/domain2/privkey.pem"),
        ],
        cert: [
            Bun.file("/etc/letsencrypt/live/domain1/cert.pem"),
            Bun.file("/etc/letsencrypt/live/domain2/cert.pem"),
        ],
    },
});
```

### What is the expected behavior?

My server is open on port 443 with 2 domain names. I want to be able to set https on both.

### What do you see instead?

![image](https://github.com/oven-sh/bun/assets/48413426/6715409b-6427-4c47-8873-8ffad0adf431)

Inplex-sys:
Same issue ...

cirospaciari:
@Inplex-sys  Does it work if you concatenate both files into one .pem or in one Buffer? will take a look into it

Inplex-sys:
it doesn't work
```
OpenSSL BAD_END_LINE via OPENSSL_internal PEM routines
```

I mean is it possible to make it work by the way it should work like @TrackZone1 said ?

cirospaciari:
For multiple domains, we would need to implement `SNICallback` option,  nodejs `tls` module throws similar error with using multiple cert/keys with different domains:

```js
node:internal/tls/secure-context:93
  context.setKey(key, passphrase);
          ^

Error: error:05800074:x509 certificate routines::key values mismatch
    at setKey (node:internal/tls/secure-context:93:11)
    at configSecureContext (node:internal/tls/secure-context:172:9)
    at Object.createSecureContext (node:_tls_common:116:3)
    at Server.setSecureContext (node:_tls_wrap:1372:27)
    at new Server (node:_tls_wrap:1236:8)
    at Object.createServer (node:_tls_wrap:1271:10)
    at Object.<anonymous> (/home/cirospaciari/Repos/bun/test.js:492:20)
    at Module._compile (node:internal/modules/cjs/loader:1267:14)
[...]

PR #10364, opened 2024-04-18
feat(SNI) Bun.serve/tls

cirospaciari:
### What does this PR do?
Fix https://github.com/oven-sh/bun/issues/9753 when using serverName:

```ts
const server = Bun.serve({
  development: true,
  async fetch(request: Request): Promise<Response> {
    // code here...
  },
  hostname: '0.0.0.0', //accessible hostname or ip to bind
  lowMemoryMode: false,
  port: 443,
  tls: {
    cert: Bun.file('./config/bun/localhost.pem'),
    key: Bun.file('./config/bun/localhost-key.pem'),
    serverName: 'get.babbel.test' // server name
  },
});
```
Fix https://github.com/oven-sh/bun/issues/6529 by adding support to multiple servernames:
 ```ts
 const server = Bun.serve({
  async fetch(request: Request): Promise<Response> {
    // code here...
  },
  hostname: '0.0.0.0', //accessible hostname or ip to bind
  port: 443,
  tls: [
   {
      cert: Bun.file('./config/bun/example.pem'),
      key: Bun.file('./config/bun/example-key.pem'),
      serverName: '*.example.com' // server name
    },
   {
      cert: Bun.file('./config/bun/test.pem'),
      key: Bun.file('./config/bun/test-key.pem'),
      serverName: '*.test.com' // server name
    }
  ],
});
```
And also fix https://github.com/oven-sh/bun/issues/8910 adding support to tls.Server.addContext https://nodejs.org/api/tls.html#serveraddcontexthostname-context

- [x] Code changes

I wrote automated tests

### CODE REVIEW

File: src/bun.js/api/bun/socket.zig [hunk 1]
```diff
@@ -814,18 +814,40 @@ pub const Listener = struct {
...
+    pub fn addServerName(this: *Listener, global: *JSC.JSGlobalObject, hostname: JSValue, tls: JSValue) JSValue {
+        if (!this.ssl) {
+            global.throwInvalidArguments("addServerName requires SSL support", .{});
+            return .zero;
+        }
+        if (!hostname.isString()) {
+            global.throwInvalidArguments("hostname pattern expects a string", .{});
+            return .zero;
+        }
+        const host_str = hostname.toSliceZ(
```
Jarred-Sumner: The poorly-named `toSliceZ`does not actually make it `[:0]const u8`

https://github.com/oven-sh/bun/blob/c5270f812105133975f343c4618c2411f5d899f3/src/bun.js/bindings/bindings.zig#L804-L821

src/bun.js/bindings/bindings.zig
```diff
@@ -804,8 +804,8 @@ pub const ZigString = extern struct {
     pub fn toSliceZ(this: ZigString, allocator: std.mem.Allocator) Slice {
         if (this.len == 0)
             return Slice.empty;
-
-        if (is16Bit(&this)) {
+        // if is 16 bit or dont has a null terminator
+        if (is16Bit(&this) or this._unsafe_ptr_do_not_use[this.len] != 0) {
```
Jarred-Sumner: We cannot do this. It is an index out of bounds.

src/js/node/tls.ts
```diff
@@ -1,7 +1,10 @@
 // Hardcoded module "node:tls"
 const { isArrayBufferView, isTypedArray } = require("node:util/types");
 const net = require("node:net");
-const { Server: NetServer, [Symbol.for("::bunternal::")]: InternalTCPSocket } = net;
+const {
+  Server: NetServer,
+  [Symbol.for("::bunternal::")]: [InternalTCPSocket, addServerName],
```
Jarred-Sumner: We can use a shared export for these now. No need to use this symbol anymore.

Jarred-Sumner (changes requested):
A few comments

src/js/node/tls.ts
cirospaciari: interesting, do you have any example of this? did not find about shared exports

Jarred-Sumner: you could add `internal/net.ts` which would export `Symbol("kSharedInternal")`

cirospaciari: nice add $zig addServerName into internals and will do a followup with a refactor removing existing unnecessary symbols

Jarred-Sumner:
yay

PR #10365, opened 2024-04-19
feat: crash reporter followup

paperclover:
- [x] store what subcommand that was run
- [x] store features list
- [x] prepare for the future better (when 7 chars isnt enough for the git hash)
- [x] update bun.report with the above changes in mind
- [x] automatically report crashes on windows and macos


paperclover:
Merge after builds succeed

PR #10374, opened 2024-04-19
filter: return with nonzero code if any script fails

### LINKED ISSUE #10359: bun `--filter` with test even if the tests fail, pass the GitHub Action

1.1.4+fbe2fe0c3

Darwin 23.4.0 arm64 arm

`bun run --filter 'my-package-1' --filter 'my-package-2' test` the GitHub action does not detect when it fails and goes green even if tests fail

To reproduce it create `github/workflows/test.yml` with:

```yml
name: Test

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]

jobs:
  build:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest]

    steps:
      - uses: actions/checkout@v4
      - name: Bunjs
        uses: oven-sh/setup-bun@v1
        with:
          bun-version: 1.1.4
      - name: Install dependencies
        run: bun install
      - run: bun run test
```

And add the test script in the global `package.json`:

```json
"scripts": {
  "test": "bun run --filter 'my-package-1' --filter 'my-package-2' test"
}
```

Add some failing test and push to main or do a PR to main and see the GH Action behavior.

the GH Action should crash since you have set the failing test

### What do you see instead?

GH Action passes correctly but if you look at the logs the tests do not pass

### Additional information

_No response_

gvilums:
### What does this PR do?
Makes `bun run --filter` return with exit code 1 if any script returns a non-zero exit code.
Fixes #10359.

Added a test


Jarred-Sumner:
@gvilums could we preserve / re-raise it instead of hardcoding it to 1?

gvilums:
> @gvilums could we preserve / re-raise it instead of hardcoding it to 1?

Problem is that when multiple subprocesses fail we would have to choose an arbitrary one. IMO always returning 1 on failure is more consistent

src/cli/filter_run.zig
```diff
@@ -356,10 +356,20 @@ const State = struct {
...
-    pub fn finalize(this: *This) void {
+    pub fn finalize(this: *This) u8 {
         if (this.aborted) {
             this.redraw(true) catch {};
         }
+        for (this.handles) |handle| {
+            if (handle.process) |proc| {
+                switch (proc.status) {
+                    .exited => |exited| if (exited.code != 0) return exited.code,
+                    .signaled => |signal| return @intFromEnum(signal),
```
Jarred-Sumner:
```suggestion
                    .signaled => |signal| return signal.toExitCode() orelse 1,
```

PR #10377, opened 2024-04-19
Run tests from Node.js and improve CI

Electroid:
### What does this PR do?

- Fixes various bugs with new CI
- Adds new job to run Node.js tests (only fails for regressions)

CI


.github/workflows/ci.yml
```diff
@@ -15,9 +15,29 @@ on:
         type: string
         description: The workflow ID to download artifacts (skips the build step)
   pull_request:
+    paths:
```
Jarred-Sumner: This is brittle

it should be any folder outside of docs

.github/workflows/run-test.yml
```diff
@@ -129,3 +129,77 @@ jobs:
...
+        shell: bash
+        run: |
+          bun install --cwd test/node
+      - name: Run Tests
+        timeout-minutes: 10 # Increase when more tests are added
+        shell: bash
+        working-directory: test/node
+        env:
+          TMPDIR: ${{ runner.temp }}
+          BUN_FEATURE_FLAG_INTERNAL_FOR_TESTING: "true"
```
Jarred-Sumner, line 197:
```suggestion
          BUN_GARBAGE_COLLECTOR_LEVEL: "0"
          BUN_FEATURE_FLAG_INTERNAL_FOR_TESTING: "true"
```

PR #10378, opened 2024-04-19
Fixes #8697

### LINKED ISSUE #8697: `--compile` flag doesn't respect `--outfile` when bundling `.node` files

1.0.25+a8ff7be64

Linux 6.2.16 x86_64 unknown

The `--compile` flag seems to suppress certain compilation errors. Create a project that relies on napi and notice that `bun build` on the file will raise the right error but `bun build --compile` seems to say it compiled successfully

Raise the same errors with `--compile` turned on

### What do you see instead?

image

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #8697

There is a test


paperclover (approved):
pretty reasonable

Issue #4862, opened 2023-09-10
TypeError: Unsupported algorithm sha3-256


Khalid-Nowaf:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

### What steps can reproduce the bug?

```ts
import crypto from "node:crypto"

crypto
            .createHash('sha3-256')
            .update("sdfsdfsdfksjdflskfjslkfjsldkfj")
            .update(process.version)
            .digest('hex');
```

```log
1051 |     }, module.exports = CipherBase;
1052 |   }
1053 | }), require_browser2 = __commonJS({
1054 |   "node_modules/create-hash/browser.js"(exports, module) {
1055 | 
1056 |     const LazyHash = function Hash(algorithm, options) {
                                                                       ^
TypeError: Unsupported algorithm sha3-256
 code: "ERR_INVALID_ARG_TYPE"

      at new Hash (node:crypto:1056:68)
      at createHash (node:crypto:1175:80)
      at /Users/xxxx/projects/xxxx/src/main.ts:14:9
      at main (/Users/xxxx/projects/xxxx/src/main.ts:4:19)
      at /Users/xxxx/projects/xxxx/src/main.ts:22:0
```

### What is the expected behavior?

it should run without error
```
TypeError: Unsupported algorithm sha3-256
```

### What do you see instead?

```log
TypeError: Unsupported algorithm sha3-256
 code: "ERR_INVALID_ARG_TYPE"

      at new Hash (node:crypto:1056:68)
      at createHash (node:crypto:1175:80)
```

### Additional information

This is my first attempt to port my code to Bun, and I noticed one of my dependencies uses `node:crypto`, which is causing this error.

simylein:
This could be missing in the implementation currently because the types suggest it is not supported. 
<img width="365" alt="Screenshot 2023-09-11 at 07 11 50" src="https://github.com/oven-sh/bun/assets/67859356/6283fd06-1128-4ce2-b8ca-f29f60e1b8c1">

Jarred-Sumner:
It's missing from our build of BoringSSL. We could probably get it from zig stdlib though

libre-man:
To add to this, the algorithm `blake2b512` is also missing.

PR #10383, opened 2024-04-20
node:crypto: add blake2b512, sha512-224, sha3-*

nektro:
Closes https://github.com/oven-sh/bun/issues/4862


src/bun.zig
```diff
@@ -2754,6 +2754,30 @@ pub inline fn new(comptime T: type, t: T) *T {
     return ptr;
 }
 
+pub inline fn newCatchable(comptime T: type, t: T) error{OutOfMemory}!*T {
```
Jarred-Sumner: i don't think we should do this for statically known sizes unless you want to write tests for it

packages/bun-types/bun.d.ts
```diff
@@ -3299,6 +3306,13 @@ declare module "bun" {
      */
     static readonly byteLength: 32;
   }
+  class SHA512_224 extends CryptoHashInterface<SHA512_224> {
```
Jarred-Sumner: we could just not add this version, and let someone complain about the lack of it

src/bun.js/api/BunObject.zig
```diff
@@ -2304,8 +2320,12 @@ pub const Crypto = struct {
         }
     };
 
-    pub const CryptoHasher = struct {
-        evp: EVP = undefined,
+    pub const CryptoHasher = union(enum) {
+        evp: EVP,
+        zig: struct {
```
Jarred-Sumner: This can be a bun.TaggedPointerUnion of the supported list of types

src/bun.js/api/BunObject.zig
```diff
@@ -2304,8 +2320,12 @@ pub const Crypto = struct {
         }
     };
 
-    pub const CryptoHasher = struct {
-        evp: EVP = undefined,
+    pub const CryptoHasher = union(enum) {
```
Jarred-Sumner, line 2321: This should go through `bun.New`:

```
usingnamespace bun.New(@This())
```

src/bun.js/api/BunObject.zig
```diff
@@ -2453,12 +2536,31 @@ pub const Crypto = struct {
                 return null;
             }
 
+            const this = bun.default_allocator.create(CryptoHasher) catch return null;
```
Jarred-Sumner: We do not need to allocate until the data is validated successfully. Ideally, error conditions wouldn't allocate

src/bun.js/api/BunObject.zig
```diff
@@ -2453,12 +2536,31 @@ pub const Crypto = struct {
...
+                    return this;
+                }
+
+                bun.default_allocator.destroy(this);
                 globalThis.throwInvalidArguments("Unsupported algorithm {any}", .{algorithm});
                 return null;
             };
-            var this = bun.default_allocator.create(CryptoHasher) catch return null;
-            this.evp = evp;
+            this.* = .{ .evp = evp };
```
Jarred-Sumner: After making the above changes, this can be:

```zig
return CryptoHasher.new(this);
```

src/bun.js/api/BunObject.zig
```diff
@@ -2502,12 +2618,25 @@ pub const Crypto = struct {
             _: *JSC.CallFrame,
         ) callconv(.C) JSC.JSValue {
             const new = bun.default_allocator.create(CryptoHasher) catch @panic("Out of memory");
-            new.evp = this.evp.copy(globalObject.bunVM().rareData().boringEngine()) catch @panic("Out of memory");
+            switch (this.*) {
+                .evp => {
+                    new.* = .{ .evp = this.evp.copy(globalObject.bunVM().rareData().boringEngine()) catch @panic("Out of memory") };
+                },
+                .zig => |*inner| {
```
Jarred-Sumner, line 2536:
```suggestion
                .zig => |*inner| inner.copy()
```

src/bun.js/api/BunObject.zig
```diff
@@ -2563,15 +2692,48 @@ pub const Crypto = struct {
 
             const output_digest_slice: []u8 = &output_digest_buf;
 
+            switch (this.*) {
+                .evp => {},
+                .zig => |*inner| {
+                    switch (inner.algorithm) {
```
Jarred-Sumner:
```suggestion
                    inner.final(BoringSSL.EVP_MAX_MD_SIZE, output_digest_slice)
```

Jarred-Sumner:
The methods using `std.crypto` should go through its own struct because as-is this is going to be difficult to maintain when people ask for more crypto algorithms to be supported or when there's a breaking change in std.crypto

Jarred-Sumner (changes requested):
A few comments

src/bun.js/api/BunObject.zig
```diff
@@ -2563,16 +2597,154 @@ pub const Crypto = struct {
...
+                },
+            }
+            this.destroy();
+        }
+    };
 
-            bun.default_allocator.destroy(this);
+    const CryptoHasherZig = struct {
+        algorithm: EVP.Algorithm,
+        state: *anyopaque,
```
Jarred-Sumner, line 2629: I think using `*anyopaque` here is more unsafe than necessary.

if the BoringSSL structs are large enough, we could instead use a `union(enum)` here of each supported zig crypto hasher type. From there, we can avoid the extra allocation for the state and the lifetime of this is simpler.

If the BoringSSL types are not already large, then we could use a `bun.TaggedPointerUnion`  and that will store the expected type in the same size as a regular pointer

src/bun.js/api/BunObject.zig
```diff
@@ -2563,16 +2597,154 @@ pub const Crypto = struct {
...
+                .@"sha3-256" => std.crypto.hash.sha3.Sha3_256.digest_length,
+                .@"sha3-384" => std.crypto.hash.sha3.Sha3_384.digest_length,
+                .@"sha3-512" => std.crypto.hash.sha3.Sha3_512.digest_length,
+            };
+        }
+
+        fn deinit(self: *CryptoHasherZig) void {
+            switch (self.algorithm) {
+                else => @panic("unreachable"),
+                .@"sha3-224" => bun.destroy(@as(*std.crypto.hash.sha3.Sha3_224, @ptrCast(@alignCast(self.state)))),
```
Jarred-Sumner: all these pointer casts can go away once the above is addressed

Jarred-Sumner (changes requested):
1 more comment

src/bun.js/api/BunObject.zig
nektro:
```zig
pub const CryptoHasher = union(enum) {
    evp: EVP,
    zig: CryptoHasherZig,
```

`evp` is `Algorithm + ptr + (ptr + ptr + ptr + ptr)`
`zig` is `Algorithm + ptr` as-is

`@sizeOf(std.crypto.hash.sha3.Sha3_*)` is `352`, `344`, `312`, `280`.

Jarred-Sumner: ah thats kind of large

can we do TaggedPointerUnion?

src/bun.js/api/BunObject.zig
```diff
@@ -2563,16 +2597,157 @@ pub const Crypto = struct {
 
             const output_digest_slice: []u8 = &output_digest_buf;
 
+            switch (this.*) {
+                .evp => {},
+                .zig => |*inner| {
+                    inner.final(output_digest_slice);
+                    return encoding.encodeWithMaxSize(globalThis, BoringSSL.EVP_MAX_MD_SIZE, output_digest_slice[0..inner.digest_length]);
```
Jarred-Sumner, line 2604: `BoringSSL.EVP_MAX_MD_SIZE` should be a different value here right?

src/bun.js/api/BunObject.zig
```diff
@@ -2563,16 +2597,157 @@ pub const Crypto = struct {
...
+                },
+            }
+
             const out = this.evp.final(globalThis.bunVM().rareData().boringEngine(), output_digest_slice);
```
Jarred-Sumner, line 2608: can this be moved into the .evp branch?:

src/bun.js/api/BunObject.zig
```diff
@@ -2484,6 +2503,14 @@ pub const Crypto = struct {
...
 
+            switch (this.*) {
+                .evp => {},
+                .zig => |*inner| {
+                    inner.update(buffer.slice());
+                    return thisValue;
+                },
+            }
+
             this.evp.update(buffer.slice());
```
Jarred-Sumner, line 2514: can you move this to .evp =>

src/bun.js/api/BunObject.zig
nektro: wanted to keep the diff small for the first iteration to make easier to review and test

src/bun.js/api/BunObject.zig
nektro: wanted to keep the diff small for the first iteration to make easier to review and test

src/bun.js/api/BunObject.zig
nektro: `BoringSSL.EVP_MAX_MD_SIZE` is the max size of all digests, `output_digest_slice` is the size of the current digest

PR #10384, opened 2024-04-20
Fixes #10322

### LINKED ISSUE #10322: `bun run --filter '*'` does not support scoped package name

1.1.4

With following monorepo structure:

```
├─package.json
└─packages
  ├─a
  │ └─package.json
  └─b
    └─package.json
```

Root `package.json` is:

```json
{
  "workspaces": ["packages/*"]
}
```

`packages/a/package.json` is:

```json
{
  "name": "@my/a",
  "private": true,
  "scripts": {
    "test": "echo a"
  }
}
```

`packages/b/package/json` is:

```json
{
  "name": "@my/b",
  "private": true,
  "scripts": {
    "test": "echo b"
  }
}
```

Type `bun run --filter '*' test` under root working directory.

If we "unscope" those subpackages name (like `a` `b` without `@my/` part), then the behavior is correct:

```
>  bun run --filter '*' test
a test $ echo a                                                                                                                                                        
│ a                                                                                                                                                                    
└─ Done in 82 ms                                                                                                                                                       
b test $ echo b                                                                                                                                                        
│ b                                                                                                                                                                    
└─ Done in 79 ms        
```

Or we use `bun run --filter '@my/*'` instead of `'*'`:

```
>  bun run --filter '@my/*' test
@my/a test $ echo a                                                                                                                                                    
│ a                                                                                                                                                                    
└─ Done in 78 ms                                                                                                                                                       
@my/b test $ echo b                                                                                                                                                    
│ b                                                                                                                                                                    
└─ Done in 70 ms      
```

### What do you see instead?

Currently bun doesn't support using '*' filter to find those scoped-name packages:

```
>  bun run --filter '*' test
error: No packages matched the filter
```

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #10322

There is a test


gvilums (approved):

DaniGuardiola:
Thanks for the fix. Has this made it to production? I'm on latest (1.1.17) and still seeing this related bug #10322, I also filed my own #12300. I haven't tried the workaround yet.

PR #10401, opened 2024-04-20
check for nested define value identifiers

### LINKED ISSUE #10399: `--define` doesn't accept nested identifiers for value

1.1.4

Darwin 23.4.0 arm64 arm

run `bun build --define` with a nested value identifier, e.g. 
```
bun build --define import.meta.url=location.origin
```

where you had `import.meta.url` you'll instead have the nested identifier `location.origin`

### What do you see instead?

where you had `import.meta.url` you'll instead have the string `"location.origin"`

### Additional information

It seems that this line is the culprit:

https://github.com/oven-sh/bun/blob/5bdc5bebb14dcca43d656529ddd33c7461e0920a/src/defines.zig#L84

It checks if the entire value is an identifier, instead of doing the nested check that's done for the key a few lines earlier:

https://github.com/oven-sh/bun/blob/5bdc5bebb14dcca43d656529ddd33c7461e0920a/src/defines.zig#L72-L82

The fix should be as simple as adopting that nested check.

Note: this is appropriately handled by esbuild

erikbrinkman:
### What does this PR do?

Makes it so if the value in a define is a nested identifier e.g. `location.origin` it is still properly substituted,  instead of casting it to a string.

fixes #10399

- [x] Code changes

Additionally

- I added tests that I think are correct, but `zig test src/defines.zig` is broken on main so I don't know if the added tests pass
- I tested the debug build manually and it worked


Jarred-Sumner:
Thank you for this

Please include a test that starts a `bun` process from JavaScript. I suggest putting it in`test/transpiler/defines.test.ts` and using `Bun.spawnSync` to spawn a process with the `--define` flag passed.

The zig test modified there won't actually run. I'll go ahead and delete all of those to avoid confusion in the future.

erikbrinkman:
@Jarred-Sumner It seems like there were already tests for this in `test/transpiler/transpiler.test.js`, that used `Bun.Transpiler` instead, so I deleted the zig tests and added one more assert there.

This test fails under `bun test` but succeeds under `build/bun-debug test`, which I assume is the expected behavior.

Jarred-Sumner:
Note that our CI isn't green right now

src/defines.zig
```diff
@@ -81,7 +81,16 @@ pub const DefineData = struct {
                 }
             }
 
-            if (js_lexer.isIdentifier(entry.value_ptr.*) and !js_lexer.Keywords.has(entry.value_ptr.*)) {
+            // check for nested identifiers
+            var valueSplitter = std.mem.split(u8, entry.value_ptr.*, ".");
+            var isIdent = true;
+            while (valueSplitter.next()) |part| {
+                if (!js_lexer.isIdentifier(part) or js_lexer.Keywords.has(part)) {
```
Jarred-Sumner, line 88: What if the value is `foo.class`? `class` is a keyword and its a valid property name, but an invalid identifier.

The first part cannot be a keyword (`import.bar`), but the 2nd one can (`foo.import`)

1. We need to handle the first part differently from the rest.
2. There's probably some edgecases involving spaces or empty parts?

erikbrinkman: For right now, it'll keep the same behavior, which is to cast it as a string. I think it's probably worth considering what the ideal behavior for this is, but it seems safer to increase the scope gradually, rather than allow too much and have to claw it back.

For that reason I decided to compare to what esbuild does. `define`s are validated here:

https://github.com/evanw/esbuild/blob/70893e4d89bab1dbf2d512593f9e4d5bcb1f76bf/pkg/api/api_impl.go#L539

which then calls this function:

https://github.com/evanw/esbuild/blob/70893e4d89bab1dbf2d512593f9e4d5bcb1f76bf/internal/js_parser/js_parser.go#L17348-L17353

which splits on periods and checks that each component is an identifier, without the keyword check.

I think it's probably safe to stick close to what esbuild does, but I don't have strong preferences. I suggest one of the following:

1. check valid identifiers and no keywords - the current code
2. only check for valid identifiers - what esbuild does
3. check all for valid identifiers, and that the first is not a keyword - a split between these, close to what you discuss

do you have a preference among these? Or do you have a case for it being something else?

Jarred-Sumner:
On 2nd thought, I think it's fine. I'll merge after tests run

PR #10420, opened 2024-04-21
Update `Bun.serve()` types documentation to match the exact values from `bun.d.ts`

mangs:
### What does this PR do?

I've been spending a lot of time looking at `Bun.serve()` and its types lately, and I noticed the website documentation documenting the types are out of date. This PR attempts to bring them in line with the current published types in `bun.d.ts`.

I matched the types exactly, added missing keys, and reordered the keys in each object to be in alphabetical order so it's a bit easier to find what you're looking for.

Formatted with Prettier.

- [ ] Code changes


docs/api/http.md
```diff
@@ -293,37 +293,58 @@ The `Bun.serve` server can handle roughly 2.5x more requests per second than Nod
...
-    error?: (error: Error) => Response | Promise<Response>;
+    error?: (request: ErrorLike) => Response | Promise<Response> | undefined | Promise<undefined>;
+    fetch(request: Request, server: Server): Response | Promise<Response>;
+    hostname?: string;
+    id?: string | null;
+    lowMemoryMode?: boolean;
+    maxRequestBodySize?: number;
+    port?: string | number;
+    reusePort?: boolean;
+    serverNames?: Record<string, TLSOptions>; // Server Name Indication (SNI) / TLS virtual hosts
```
Jarred-Sumner: `serverNames` is incorrect, it does not exist in the codebase

`tls` can optionally be an array of the object type or just the object type

mangs: Hmm, this must have changed since I filed this; this is how it was yesterday. Thanks for the heads up. I just pulled `main` now and the types appear to be different. I'll do a once over on all of them to make sure they're right.

mangs: Actually come to think of it I was checking types from `bun-types` version `1.1.4` not `main`. I'm checking `main` now.

mangs: All set

mangs:
Bump. Pretty sure this is ready to go.

Issue #10376, opened 2024-04-19
Infinite loop problems on bun install


eye-wave:
### What version of Bun is running?

1.1.3

### What platform is your computer?

Linux 6.8.0-76060800daily20240311-generic x86_64 x86_64

### What steps can reproduce the bug?

create following package.json
```json
{
  "scripts": {
    "install": "bun install"
  }
}
```

### What is the expected behavior?

with other package managers, they all seem to download dependencies first, and then fall into the infinite loop of doing nothing basically

### What do you see instead?

with bun it's the infinite loop from the beginning. 

### Additional information

in other circumstances this leads to a massive memory spike that crashes entire operating system - when working with workspaces in monorepo with turbopack

PR #10421, opened 2024-04-21
Fix issue #10376

Uziniii:
### What does this PR do?

Fix issue #10376
This pr avoid infinite install loop when putting bun install in post install script in package.json by exiting

- [x] Code changes


paperclover (changes requested):
- please use `bun.strings.eqlComptime()`. you can do this all inline like `if (bun.strings.eqlComptime(combined_script, "bun install")) {`
- i don't think `bun-debug` needs to be considered here
- it should print an error message as to why this is going to be a recursive call. use `Output.errGeneric(fmt, args)` and/or `Output.note` for this.

src/install/lifecycle_script_runner.zig
```diff
@@ -123,6 +123,12 @@ pub const LifecycleScriptSubprocess = struct {
 
         const combined_script: [:0]u8 = copy_script.items[0 .. copy_script.items.len - 1 :0];
 
+        const install_script = "bun install";
+        if (combined_script.len >= install_script.len and bun.strings.eqlComptime(combined_script[0..install_script.len], install_script)) {
```
paperclover: i didnt realize this was a starts with check before. it can just be `bun.strings.hasPrefixComptime(combined_script, "bun install")`

src/install/lifecycle_script_runner.zig
```diff
@@ -123,6 +123,12 @@ pub const LifecycleScriptSubprocess = struct {
...
+            Output.errGeneric("post install script has been canceled to prevent an infinite loop", .{});
```
paperclover:
```suggestion
            Output.errGeneric("\"{s}\" package.json script calling \"bun install\" causes an infinite loop.", .{this.scriptName()});
            
```

src/install/lifecycle_script_runner.zig
```diff
@@ -123,6 +123,12 @@ pub const LifecycleScriptSubprocess = struct {
...
+            Global.exit(0);
```
paperclover:
```suggestion
            Global.exit(1);
```

paperclover (changes requested):

dylan-conway (approved):
Going to merge this. I don't think this is perfect solution (scripts could be `foo=bun; $foo i`, `bun add ...`, ...), but `bun install` will cover most situations. If we had the bun shell enabled for lifecycle scripts on all platforms, we could probably detect `bun install/add` inside the interpreter accurately

dylan-conway:
Needed to revert this pr. It's not uncommon for projects to have `postinstall: "bun install --cwd <dir>"`. This won't trigger an  infinite loop because \<dir\> is not a workspace

Issue #10293, opened 2024-04-16
Completion: Support Tab completion for arguments in powershell


musafiroon:
### What is the problem this feature would solve?

Support tab completion for powershell. 

### What is the feature you are proposing to solve the problem?

Using [Register-ArgumentCompleter
](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/register-argumentcompleter?view=powershell-7.4) , we can register arguments for powershell tab completion

paperclover:
Duplicate of #8939

PR #10424, opened 2024-04-21
Add powershell completer (#8939)

ShaunLawrie:
### What does this PR do?

This adds completions for PowerShell. It should work back to PowerShell v5 and cross platform.  
Rather than add hard coded completions it pulls available args from the bun help.

- [x] Code changes

Built and installed on Windows for now, the idea needs fleshing out which is why this is a draft.

### DIFF
--- a/src/sys.zig
+++ b/src/sys.zig
@@ -2233,27 +2233,14 @@ pub fn isExecutableFileOSPath(path: bun.OSPathSliceZ) bool {
         // Rationale: `GetBinaryTypeW` does not work on .cmd files.
         // Windows does not have executable permission like posix does, instead we
         // can just look at the file extension to determine executable status.
-        @compileError("Do not use isExecutableFilePath on Windows");
-
-        // var out: windows.DWORD = 0;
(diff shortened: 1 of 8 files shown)


src/sys.zig
```diff
@@ -2233,27 +2233,14 @@ pub fn isExecutableFileOSPath(path: bun.OSPathSliceZ) bool {
...
-        //     kernel32.SCS_PIF_BINARY,
-        //     kernel32.SCS_POSIX_BINARY,
-        //     => true,
-        //     else => false,
-        // };
-
-        // log("GetBinaryTypeW({}) = {d}. isExecutable={}", .{ bun.fmt.utf16(path), out, result });
-
-        // return result;
+        const stringPath = bun.String.createFromOSPath(path);
```
ShaunLawrie, line 2236: It's my first time writing zig. This works but I'm sure there's a better way to do this. Windows 16bit-ish paths that couldn't be squished to utf8 would not match any executable extensions.
image

ShaunLawrie:
Also for duplicate issue #10293

PR #10445, opened 2024-04-22
fix(install): handle transitive folder dependencies

### LINKED ISSUE #10324: bun install cannot resolve local file dependencies

1.1.4+fbe2fe0c3

macosx

We have a module we publish which bundles some eslint rules, and it does so with a file:./ reference in the package.json

Screenshot 2024-04-17 at 11 59 27

Subsequently this fails to install in a downstream app with:

```
error: eslint-plugin-webserver-custom-rules@file:./webserver-custom-rules failed to resolve
error: eslint-plugin-custom-rules@file:./custom-rules failed to resolve
```

File dependencies to work

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #11519: `bun install` fails to install `@roxi/routify@3.0.0-next.228`

latest

Linux 6.8.0-31-generic x86_64 x86_64, Linux 6.5.0-1021-azure x86_64 x86_64 (codespaces)

bun install

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

dylan-conway:
### What does this PR do?

Sometimes packages are published with `file:` dependencies. The pr installs them with symlinks, without erroring if the target does not exist

fixes #10324 
fixes #11519 

added tests


src/install/dependency.zig
```diff
@@ -967,18 +990,81 @@ pub fn parseWithTag(
...
+                            };
+                        }
+
+                        break :folder dependency[protocol + 1 ..];
+                    };
+
+                    // from npm:
+                    //
+                    // turn /C:/blah info just C:/blah on windows
+                    // https://github.com/npm/cli/blob/fc6e291e9c2154c2e76636cb7ebf0a17be307585/node_modules/npm-package-arg/lib/npa.js#L277
```
Jarred-Sumner: thanks for including the link to the source code

src/install/install.zig
```diff
@@ -988,10 +988,38 @@ pub const PackageInstall = struct {
...
+                }
+                return this.verifyPackageJSONNameAndVersion(root_node_modules_dir);
+            },
         };
     }
 
+    fn verifySymlinkedFolder(this: *PackageInstall, root_node_modules_dir: std.fs.Dir) bool {
+        var destination_dir = this.node_modules.openDir(root_node_modules_dir) catch return false;
+        defer {
+            if (std.fs.cwd().fd != destination_dir.fd) destination_dir.close();
```
Jarred-Sumner: this will never be true

src/install/install.zig
```diff
@@ -988,10 +988,38 @@ pub const PackageInstall = struct {
...
+        }
+
+        var dir: std.fs.Dir = bun.openDir(destination_dir, this.destination_dir_subpath) catch return false;
+        defer dir.close();
+
+        var iter = bun.DirIterator.iterate(dir, .u8);
```
Jarred-Sumner: doing all this work of verifying costs just as much as just re-symlinking it

src/install/install.zig
```diff
@@ -988,10 +988,38 @@ pub const PackageInstall = struct {
...
+
+        while (switch (iter.next()) {
+            .result => |entry| entry,
+            .err => return false,
+        }) |entry| {
+            if (entry.kind == .file) {
```
Jarred-Sumner: ah i wish i'd seen this sooner

Jarred-Sumner (commented):
Thank you for writing these tests

I think everything in the PR fine except for the verify symlink part. That part should be changed to either:

-  do no checks at all, just verify package.json version like we do elsewhere
-  verify each symlink by re-creating it. it's just as expensive to create a bunch of symlinks as it is to check if those symlinks are valid. and, checking if symlinks are valid is a [TOCTOU race condition](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)

src/install/lockfile.zig
```diff
@@ -821,15 +824,51 @@ pub fn clean(
...
+    return id == 0 or this.buffers.dependencies.items[this.buffers.trees.items[id].dependency_id].behavior.isWorkspaceOnly();
+}
+
+pub fn getWorkspacePackageID(this: *const Lockfile, workspace_name_hash: ?PackageNameHash) PackageID {
+    return if (workspace_name_hash) |workspace_name_hash_| brk: {
+        const packages = this.packages.slice();
+        const name_hashes = packages.items(.name_hash);
+        const resolutions = packages.items(.resolution);
+        const metas = packages.items(.meta);
+        for (resolutions, name_hashes, metas) |res, name_hash, meta| {
```
Jarred-Sumner: We can avoid looping through `metas`

Package IDs are sequential, so the index of the current element is also the package id in meta.

We shouldn't really store the package IDs in meta, but we do for historical reasons.

src/install/install.zig
```diff
@@ -11391,7 +11403,25 @@ pub const PackageManager = struct {
...
                     .symlink, .workspace => installer.installFromLink(this.skip_delete, destination_dir),
-                    else => installer.install(this.skip_delete, destination_dir),
+                    else => result: {
+                        if (resolution.tag == .folder and !this.lockfile.isWorkspaceTreeId(this.current_tree_id)) {
+                            // This is a transitive folder dependency. It is installed with a single symlink to the target folder/file,
+                            // and is not hoisted.
+                            const dirname = std.fs.path.dirname(this.node_modules.path.items) orelse this.node_modules.path.items;
+
+                            installer.cache_dir = this.root_node_modules_folder.openDir(dirname, .{ .iterate = true, .access_sub_paths = true }) catch
+                                break :result PackageInstall.Result.fail(error.FileNotFound, .opening_cache_dir);
```
Jarred-Sumner: Did you mean to hardcode FileNotFound here? There could be other errors.

Issue #10356, opened 2024-04-18
Behaviour of fs.existsSync, when using long file paths, deviates from Node


torjue:
### What version of Bun is running?

1.1.4

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

```typescript
import fs from "node:fs";
fs.existsSync("<string with more than 1025 characters>");
```

### What is the expected behavior?

Not to throw, as Node.js. Also, max file length on most Linux systems are often set to 4096, not 1024.

### What do you see instead?

```
1 | import fs from "node:fs";
2 | fs.existsSync(
    ^
ENAMETOOLONG: File name too long
 errno: -63
```

### Additional information

Some packages uses fs.existsSync as a way to check if a parameter contains information or points to a file. Probably not the best idea, but it's still used. See [these lines](https://github.com/hashicorp/terraform-cdk/blob/153a4b2bb7312abf9ca5ce1b501d724f3da153c4/packages/cdktf/lib/testing/matchers.ts#L146-L148) in CDKTF from Hashicorp.

Jarred-Sumner:
ah yes, 

Reading node's code, looks like `fs.existsSync` is supposed to never throw, including on invalid input:
https://github.com/nodejs/node/blob/main/lib/fs.js#L275-L295

```js
// fs.existsSync never throws, it only returns true or false.
// Since fs.existsSync never throws, users have established
// the expectation that passing invalid arguments to it, even like
// fs.existsSync(), would only get a false in return, so we cannot signal
// validation errors to users properly out of compatibility concerns.
// TODO(joyeecheung): deprecate the never-throw-on-invalid-arguments behavior
/**
 * Synchronously tests whether or not the given path exists.
 * @param {string | Buffer | URL} path
 * @returns {boolean}
 */
function existsSync(path) {
  try {
    path = getValidatedPath(path);
  } catch {
    return false;
  }

  return binding.existsSync(pathModule.toNamespacedPath(path));
}
```

We need to make ours behave similarly

PR #10446, opened 2024-04-22
existsSync should never throw ENAMETOOLONG

nektro:
Fixes https://github.com/oven-sh/bun/issues/10356

### DIFF
--- a/src/js/node/fs.promises.ts
+++ b/src/js/node/fs.promises.ts
@@ -161,7 +161,7 @@ const exports = {
   cp,
   exists: async function exists() {
     try {
-      return fs.exists.$apply(fs, arguments);
+      return await fs.exists.$apply(fs, arguments);
     } catch (e) {
       return false;
     }
--- a/src/js/node/fs.promises.ts
+++ b/src/js/node/fs.promises.ts
@@ -159,7 +159,13 @@ const exports = {
   close: fs.close.bind(fs),
   copyFile: fs.copyFile.bind(fs),
   cp,
-  exists: fs.exists.bind(fs),
+  exists: async function exists() {
+    try {
+      return fs.exists.$apply(fs, arguments);
+    } catch (e) {
+      return false;
+    }
+  },
   chown: fs.chown.bind(fs),
   chmod: fs.chmod.bind(fs),
   fchmod: fs.fchmod.bind(fs),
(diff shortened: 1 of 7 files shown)


Jarred-Sumner (changes requested):
1) This de-optimizes `fs.exists` and `fs.promises.exists` because `PathLike` has better code paths for handling file paths that may be used across threads.

2) This specifically handles `ENAMETOOLONG`, but a future change may add more errors for us to throw, and it will throw in those cases (leading the original issue to regress again)

I think a better fix here would be to define `existsSync` as something like
```ts
(path: string) => { 
  try {
     return fs.existsSync.$apply(fs, arguments);
  } catch (e) {
     return false;
  } 
}
```


Which is closer to what Node does as well

src/js/node/fs.promises.ts
```diff
@@ -159,7 +159,13 @@ const exports = {
   close: fs.close.bind(fs),
   copyFile: fs.copyFile.bind(fs),
   cp,
-  exists: fs.exists.bind(fs),
+  exists: async function exists() {
+    try {
+      return fs.exists.$apply(fs, arguments);
```
Jarred-Sumner: We have to await this value or a rejected Promise will throw

```suggestion
      return await fs.exists.$apply(fs, arguments);
```

src/js/node/fs.ts
```diff
@@ -309,7 +309,13 @@ var access = function access(...args) {
   appendFileSync = fs.appendFileSync.bind(fs),
   closeSync = fs.closeSync.bind(fs),
   copyFileSync = fs.copyFileSync.bind(fs),
-  existsSync = fs.existsSync.bind(fs),
+  existsSync = function existsSync() {
```
Jarred-Sumner, line 318: Don't forget about `fs.exists`, the asynchronous callback-calling kind

https://github.com/nodejs/node/blob/c82f3c9e80f0eeec4ae5b7aedd1183127abda4ad/lib/fs.js#L248-L266

Jarred-Sumner (changes requested):
2 comments

src/js/node/fs.ts
nektro: that one's already try/catch'd but I'll add a test

nektro: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts#L3468 maybe I shouldn't?

Jarred-Sumner:
Thank you

Issue #7716, opened 2023-12-18
IncomingRequest fails to emit `close` event when client disconnects


davidstevens37:
### What version of Bun is running?

1.1.7+b0b7db5c0

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

Creating a Server and attempting to use the `req.on('close')` to manage cleanup for for when a request closes does not work when using the Bun runtime. 

The example below simply requires running the `example.js` file using `bun example.js` and navigating to http://localhost:3000, the page will load, connect to the /event route and begin handling Server Sent Events. After 3 server sent events the client will close the connection, but the `close` event is never emitted. This is true for all requests, not just in the Sever Sent Events use case.

Navigating to the `/slow-response` route, which sets a timeout for 10 seconds before responding to the client, and cancelling the request before it fulfills also produces the same results. The `close` event is not fired, which doesn't allow the cleanup to prevent the timeout (or interval in the SSE use case), of course leading to memory leaks and undeterminable connection state.

```javascript
// example.js
import { createServer } from 'node:http';

const BASIC_HTML = `<html>
  <head> 
    <title>Server Sent Events</title> 
  </head>
  <body>
    <h1>Server Sent Events</h1>
    <div id="events"></div>
    <script>
      function addEvent(data) {
        const events = document.getElementById('events');
        const newItem = document.createElement('div');
        newItem.innerHTML = new Date() + ' ' + JSON.stringify(data);
        events.appendChild(newItem);
      }

      let eventCount = 0;
      const eventSource = new EventSource('/events');

      function closeConnection() {
        eventSource.close();
        addEvent({ type: 'closing_connection' });
      }

      eventSource.onerror = function(err) {
          addEvent({ type: 'error', error: err });
          closeConnection();
      }
      
      eventSource.addEventListener('action', function(event) {
          addEvent({ type: 'received_action', data: event.data});
          eventCount++;
          if (eventCount >= 3) {
            closeConnection();
          }
      });
      window.onbeforeunload = closeConnection;
    </script>
  </body>
`;

function sendHomePage(req, res) {
  console.log('HomeRoute')
  res.writeHead(200, { 'Content-Type': 'text/html' });
  res.end(BASIC_HTML);
}

function handleSlowRoute(req, res) {
    console.log('SlowRoute')

    const timeout = setTimeout(() => {
      console.log('sending response from inside timeout');
      res.writeHead(200, { 'Content-Type': 'text/html' });
[...]

Issue #5470, opened 2023-09-15
ClickHouse DB not connecting


stalingino:
### What version of Bun is running?

1.0.1

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

```js
import { createClient } from "@clickhouse/client";

console.log('connecting ClickHouse DB ...')
const CHDB = createClient({
    host: config.database.chdb.host,
    username: config.database.chdb.user,
    password: config.database.chdb.password,
    database: config.database.chdb.database
})
console.log(await CHDB.ping())
```

### What is the expected behavior?
```
connecting ClickHouse DB ...
{ success: true }
```
### What do you see instead?
```
connecting ClickHouse DB ...
```
### Additional information

It just hangs and laptop fan starts to make high noise.

Before bun 1.0, it used to crash with only "Segmentation Fault" in console.

pavanbadugu:
the same is observed on m1 as well when using bun

PaganMuffin:
After changing ``@clickhouse/client`` to ``@clickhouse/client-web`` it started working for me in bun v1.0.5.

abhi-slash-git:
No luck for me, not even showing error logs

SoprachevAK:
For `v1.0.23` on macos reproduce. No errors, it just doesn't work

yanislav-igonin:
Same, macOS, not working, no errors.

calebeaires:
Hey, I have the same issue in my application. When using @clickhouse/client-web it works, but this "secondary official package" does not have support to insert using stream.

Have you guys found a solution to this matter?

aquinary:
Well, i also ran into this problem

royalbhati:
+1
same issue on mac m1 and bun version v1.0.30

skoshx:
Same issue, `@clickhouse/client` just exits with no stdout, `@clickhouse/client-web` almost works, it inserts data, but then exists with this:

```typescript
1 | (function (controller) {"use strict";
                       ^
TypeError: undefined is not an object
      at readableStreamDefaultControllerCanCloseOrEnqueue (:1:20)
      at readableStreamDefaultControllerCallPullIfNeeded (:1:20)

```

(So catching doesn't work, it just crashes)

Issue #4357, opened 2023-08-26
http server - add 'connection' event


mikeTWC1984:
### What version of Bun is running?

0.8

### What platform is your computer?

linux

### What steps can reproduce the bug?

This is nodejs compatibility issue. HTTP Server has on 'connection' event that handles socket. It doesn't seem to work in bun. If you run below snippet in nodejs it will print out socket object (after curl http://localhost:9000). In bun "connection" event is not handled at all.

```js

const http = require("http");
const url = require("url");

const server = http.createServer((req, res) => {    
    const reqUrl = url.parse(req.url).pathname
    if(reqUrl == "/") {
        res.write("hello world")
        res.end()
    }
})

server.on('connection', function(socket) {
    
    console.log(socket)
})

server.listen(9000)
```

### What is the expected behavior?

need to get socket object on connection

### What do you see instead?

connection event is not handled

mikeTWC1984:
FYI still not working on 1.0.0.  So what I'm really looking for is to get access to socket object on connection. Maybe there is some other ways to achieve it now?

PR #10447, opened 2024-04-22
node:http.request should emit events in the right order

nektro:
Fixes https://github.com/oven-sh/bun/issues/4357
Fixes https://github.com/oven-sh/bun/issues/5470
Fixes https://github.com/oven-sh/bun/issues/7716


test/js/node/http/node-http.test.ts
```diff
@@ -1804,4 +1806,33 @@ it("#10177 response.write with non-ascii latin1 should not cause duplicated char
       }
       finish();
     });
+}, 20_000);
+
+it("should emit events in the right order", async () => {
+  const { stdout, stderr, exited } = Bun.spawn({
+    cmd: [bunExe(), "run", path.join(import.meta.dir, "fixtures/log-events.mjs")],
+    stdout: "pipe",
+    stdin: "pipe",
```
Jarred-Sumner:
```suggestion
    stdin: "ignore",
```

test/js/node/http/node-http.test.ts
```diff
@@ -1804,4 +1806,33 @@ it("#10177 response.write with non-ascii latin1 should not cause duplicated char
...
+    stderr: "pipe",
+    env: bunEnv,
+  });
+  expect(stderr).toBeDefined();
```
Jarred-Sumner: we shouldn't need to check that stderr/stdout is defined, that would be a bug in Bun.spawn if it weren't

```suggestion
  expect(await exited).toBe(0);
```

test/js/node/http/node-http.test.ts
```diff
@@ -1804,4 +1806,33 @@ it("#10177 response.write with non-ascii latin1 should not cause duplicated char
...
+  const err = await new Response(stderr).text();
+  expect(err).toBeEmpty();
+  expect(stdout).toBeDefined();
```
Jarred-Sumner:
```suggestion
```

src/js/node/http.ts
```diff
@@ -1268,7 +1269,7 @@ ServerResponse.prototype.assignSocket = function (socket) {
     throw ERR_HTTP_SOCKET_ASSIGNED();
   }
   socket._httpMessage = this;
-  socket.on("close", onServerResponseClose);
+  socket.on("close", () => onServerResponseClose.$call(socket));
```
Jarred-Sumner, line 1271: theres never an error event or anything passed along as arguments to socket.on("close"), right?

nektro: correct

Jarred-Sumner:
The code looks fine, but there are several failing tests

src/js/node/stream.ts
```diff
@@ -1890,7 +1888,7 @@ var require_destroy = __commonJS({
           r.errored = err;
         }
         if (sync) {
-          runOnNextTick(emitErrorNT, stream, err);
```
Jarred-Sumner, line 1893: This adds an extra property lookup in a very hot code path

davidstevens37:
@nektro @Jarred-Sumner  Thank you for your hard work and contributions on this 🙏 !  It appears this has been included since v1.1.5 (please correct me if i'm wrong). 

I've just retested #7716 in both v.1.1.6 and v1.1.6-canary. However, the behavior presents as it does in the issue: the IncomingRequest events are not emitted when the client disconnects.

Is there something I might be misunderstanding about the expected behavior? If not, we might need to re-open #7716 as it's still prone to memory leaks caused by the server being unaware when a client connection is closed.

Thanks again and apologies if I've misunderstood something.

nektro:
apologies, will take a look and confirm first thing in the morning

Issue #7959, opened 2024-01-03
Invalid stdio option: "[object ReadStream]"


kkhnifes:
### What version of Bun is running?

1.0.21

### What platform is your computer?

Linux 6.6.8-200.fc39.x86_64 x86_64 unknown

### What steps can reproduce the bug?

I'm encountering an issue with Astro Starlight — pagefind is not able to build.

If try to run `bunx --bun  astro build`, I get this error:
```bash
➜  documentation git:(1-add-astro-starlight) ✗ bunx --bun  astro build     

16:22:34 Types generated 297ms
16:22:34 [build] output: "static"
16:22:34 [build] directory: /home/kkhnifes/work/firestart/documentation/public/
16:22:34 [build] Collecting build info...
16:22:34 [build] ✓ Completed in 487ms.
16:22:34 [build] Building static entrypoints...

warn - As of Tailwind CSS v2.2, `lightBlue` has been renamed to `sky`.
warn - Update your configuration file to silence this warning.

warn - As of Tailwind CSS v3.0, `warmGray` has been renamed to `stone`.
warn - Update your configuration file to silence this warning.

warn - As of Tailwind CSS v3.0, `trueGray` has been renamed to `neutral`.
warn - Update your configuration file to silence this warning.

warn - As of Tailwind CSS v3.0, `coolGray` has been renamed to `gray`.
warn - Update your configuration file to silence this warning.

warn - As of Tailwind CSS v3.0, `blueGray` has been renamed to `slate`.
warn - Update your configuration file to silence this warning.
16:22:36 [build] ✓ Completed in 2.15s.

 building client (vite) 
16:22:36 [vite] ✓ 15 modules transformed.
16:22:36 [vite] public/_astro/Tabs.astro_astro_type_script_index_0_lang.YYQMiWOA.js  0.97 kB │ gzip: 0.48 kB
16:22:36 [vite] public/_astro/page.y8mHBBe5.js                                       2.25 kB │ gzip: 1.03 kB
16:22:36 [vite] public/_astro/hoisted.hL1V9QxT.js                                    4.97 kB │ gzip: 1.85 kB
16:22:36 [vite] ✓ built in 141ms

 generating static routes 
16:22:36 λ astro-expressive-code/routes/styles.ts
16:22:36   └─ /_astro/ec.0epgx.css (+5ms)
16:22:36 λ astro-expressive-code/routes/scripts.ts
16:22:36   └─ /_astro/ec.sgewm.js (+1ms)
[...]

Issue #4402, opened 2023-08-30
Support npm-run-all


hyuzea:
### What version of Bun is running?

0.8.2+a846852818278641cf33413ce784adf2fc0e2e52

### What platform is your computer?

Linux 5.15.0-67-generic x86_64 x86_64 - ubuntu 22.04

### What steps can reproduce the bug?

make a project with only has `npm-run-all` in deps
```
{
  "name": "npm-run-all-test",
  "version": "1.0.0",
  "type": "module",
  "scripts": {
    "all": "run-p echo-hello echo-world",
    "echo-hello": "echo hello",
    "echo-world": "echo world"
  },
  "devDependencies": {
    "npm-run-all": "^4"
  }
}
```

Run
`bun install &&  bun --bun run all`

### What is the expected behavior?

```
$ run-p echo-hello echo-world
$ echo hello
hello
$ echo world
world
```

### What do you see instead?

```
error: Invalid stdio option
      at node:child_process:356:46
      at map (:1:20)
      at spawn (node:child_process:611:14)
      at node:child_process:2:40
      at spawn (/data/www/00-cup/bun-play/npm-run-all-test/node_modules/cross-spawn/index.js:12:20)
      at spawn (/data/www/00-cup/bun-play/npm-run-all-test/node_modules/npm-run-all/lib/spawn-posix.js:60:18)
      at /data/www/00-cup/bun-play/npm-run-all-test/node_modules/npm-run-all/lib/run-task.js:174:13
      at new Promise (:1:20)
      at runTask (/data/www/00-cup/bun-play/npm-run-all-test/node_modules/npm-run-all/lib/run-task.js:138:20)
      at next (/data/www/00-cup/bun-play/npm-run-all-test/node_modules/npm-run-all/lib/run-tasks.js:122:28)

error: script "all" exited with code 1 (SIGHUP)
```

silvenon:
I just realized that that `npm-run-all` has last been published 5 years ago 😅 But I use it all the time!

`concurrently` works on Bun apparently so at least we have an alternative.

hyuzea:
> `concurrently` works on Bun apparently so at least we have an alternative.

I dont know about `concurently`. 
Mainly I use `run-p` for spawning php dev server. But finally, instead I replaced run-p with my own vite plugin using Bun.spawn() to start php-cgi

Electroid:
Newest error as of Bun v1.0.7.

```js
❯ bun --bun run all
[...]

PR #10451, opened 2024-04-23
node:child_process: allow passing stdin,stdout,stderr to spawn

nektro:
Fixes https://github.com/oven-sh/bun/issues/4402
Fixes #7959

### DIFF
--- a/src/js/builtins/ProcessObjectInternals.ts
+++ b/src/js/builtins/ProcessObjectInternals.ts
@@ -119,8 +119,7 @@ export function getStdinStream(fd) {
   }
 
   const tty = require("node:tty");
-  const fs = require("node:fs");
-  const ReadStream = tty.isatty(fd) ? tty.ReadStream : fs.ReadStream;
+  const ReadStream = tty.isatty(fd) ? tty.ReadStream : require("node:fs").ReadStream;
   const stream = new ReadStream(fd);
 
   const originalOn = stream.on;
--- a/src/js/builtins/ProcessObjectInternals.ts
+++ b/src/js/builtins/ProcessObjectInternals.ts
@@ -26,7 +26,6 @@
 
 export function getStdioWriteStream(fd) {
   const tty = require("node:tty");
-  const fs = require("node:fs");
 
   let stream;
   if (tty.isatty(fd)) {
@@ -36,6 +35,7 @@ export function getStdioWriteStream(fd) {
     });
     stream._type = "tty";
   } else {
+    const fs = require("node:fs");
     stream = new fs.WriteStream(fd, { autoClose: false, fd });
     stream._type = "fs";
   }
@@ -61,9 +61,6 @@ export function getStdioWriteStream(fd) {
 }
 
 export function getStdinStream(fd) {
-  const tty = require("node:tty");
-  const fs = require("node:fs");
-
   // Ideally we could use this:
   // return require("node:stream")[Symbol.for("::bunternal::")]._ReadableFromWeb(Bun.stdin.stream());
   // but we need to extend TTY/FS ReadStream
@@ -121,6 +118,8 @@ export function getStdinStream(fd) {
     }
   }
 
+  const tty = require("node:tty");
+  const fs = require("node:fs");
   const ReadStream = tty.isatty(fd) ? tty.ReadStream : fs.ReadStream;
   const stream = new ReadStream(fd);
 
--- a/src/js/builtins/ProcessObjectInternals.ts
+++ b/src/js/builtins/ProcessObjectInternals.ts
@@ -26,6 +26,7 @@
 
 export function getStdioWriteStream(fd) {
   const tty = require("node:tty");
+  const fs = require("node:fs");
 
   let stream;
   if (tty.isatty(fd)) {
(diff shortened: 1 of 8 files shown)


src/js/builtins/ProcessObjectInternals.ts
```diff
@@ -26,6 +26,7 @@
 
 export function getStdioWriteStream(fd) {
   const tty = require("node:tty");
+  const fs = require("node:fs");
```
Jarred-Sumner: `require("node:fs")` is one of the most expensive imports we can do 

let's defer it as late as possible

src/js/builtins/ProcessObjectInternals.ts
```diff
@@ -117,9 +121,7 @@ export function getStdinStream(fd) {
     }
   }
 
-  const tty = require("node:tty");
-
-  const ReadStream = tty.isatty(fd) ? tty.ReadStream : require("node:fs").ReadStream;
+  const ReadStream = tty.isatty(fd) ? tty.ReadStream : fs.ReadStream;
```
Jarred-Sumner:
```suggestion
  const ReadStream = require("node:tty").isatty(fd) ? require("node:tty").ReadStream : require("node:fs").ReadStream;
```

src/js/node/child_process.ts
```diff
@@ -2,6 +2,7 @@
 const EventEmitter = require("node:events");
 const StreamModule = require("node:stream");
 const OsModule = require("node:os");
+const FsModule = require("node:fs");
```
Jarred-Sumner: extremely expesnive, lets avoid loading this when not 100% necessary

we should honestly do the same for node:os

src/js/node/child_process.ts
```diff
@@ -1359,9 +1360,17 @@ function nodeToBun(item, index) {
   if (typeof item === "number") {
     return item;
   }
+  if (item instanceof StreamModule.Readable) {
+    if (Object.hasOwn(item, "fd") && typeof item.fd === "number" && item.fd >= 0 && item.fd <= 2) return item.fd;
+    throw new Error(`TODO: stream.Readable stdio @ ${index}`);
```
Jarred-Sumner, line 1364: can we just silently ignore for now and have a debug warning? this might break something which wasn't necessarily previously broken

src/js/node/child_process.ts
```diff
@@ -1359,9 +1360,17 @@ function nodeToBun(item, index) {
   if (typeof item === "number") {
     return item;
   }
+  if (item instanceof StreamModule.Readable) {
```
Jarred-Sumner: instanceof can throw, lets not use it

https://github.com/oven-sh/bun/blob/e6954c440e03d27c3b56d13a7834c2b7dfbb3123/src/js/node/stream.ts#L822-L824

We can check for the existence of a `pipe` function and an `on` function

src/js/node/child_process.ts
```diff
@@ -1359,9 +1360,17 @@ function nodeToBun(item, index) {
...
+    if (Object.hasOwn(item, "fd") && typeof item.fd === "number" && item.fd >= 0 && item.fd <= 2) return item.fd;
+    throw new Error(`TODO: stream.Readable stdio @ ${index}`);
+  }
+  if (item instanceof StreamModule.Writable) {
```
Jarred-Sumner: Looks like we can check for `write` and `on` functions

https://github.com/oven-sh/bun/blob/e6954c440e03d27c3b56d13a7834c2b7dfbb3123/src/js/node/stream.ts#L836-L838

src/js/node/child_process.ts
nektro: I can prevent the throw with a `typeof item === 'object'` guard too

Jarred-Sumner: not quite, instanceof is overridable

src/js/node/child_process.ts
nektro: this path was already throwing with `Invalid stdio option: "[object ReadStream]"`. these throws make the error more descriptive

src/js/builtins/ProcessObjectInternals.ts
Jarred-Sumner: This was not resolved, it is eagerly loading `node:fs` when it isn't always necessary.

src/js/node/child_process.ts
```diff
@@ -1359,9 +1359,17 @@ function nodeToBun(item, index) {
   if (typeof item === "number") {
     return item;
   }
+  if (typeof item === "object" && item && typeof item.pipe === "function" && typeof item.on === "function") {
```
Jarred-Sumner: can we move this into a function so it's clear what it's doing? 

```
function isReadableNodeStream(stream) {

}
```

src/js/node/child_process.ts
```diff
@@ -1359,9 +1359,17 @@ function nodeToBun(item, index) {
...
+    if (Object.hasOwn(item, "fd") && typeof item.fd === "number") return item.fd;
+    throw new Error(`TODO: stream.Readable stdio @ ${index}`);
+  }
+  if (typeof item === "object" && item && typeof item.write === "function" && typeof item.on === "function") {
```
Jarred-Sumner: can we move this into a function so it's clear what it's doing? 

```
function isWritableNodeStream(stream) {

}
```

Jarred-Sumner (changes requested):
3 comments

Jarred-Sumner:
Let's merge once the tests run assuming the aarch64 failing test count goes down

Issue #10206, opened 2024-04-12
Incorrect encoding when loading text file as module


mattdw:
### What version of Bun is running?

1.1.3+2615dc742

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

Create the following `test_file.txt`:

```
Hello, 🐹
Hi, ∆∫å
Hey, 世界
```

And this script `encoding_test.js`:

```javascript
import text from "./test_file.txt";

console.log("The imported module content is:");
console.log(text);

const manual = await Bun.file("./test_file.txt").text();

console.log("The manually read content is:");
console.log(manual);
```

When run with `bun encoding_test.js` I get the following output:

```
The imported module content is:
Hello, ð
Hi, ââ«Ã¥
Hey, ä¸ç

The manually read content is:
Hello, 🐹
Hi, ∆∫å
Hey, 世界
```

### What is the expected behavior?

The expected behaviour is that both the imported module content and the manually read content from the text script are identical, as they are both read from the same file.

### What do you see instead?

Instead, the text from using `import` on a text file has garbled encoding, and any unicode characters are broken.

### Additional information

I'd love if there was a way to provide hints to the import syntax - ideally for both loader and encoding. For instance, it'd be great to be able to do:

```javascript
import text from "./handlebars_template.html" with {type: "text", encoding: "utf-8"};
```

(Currently it's also not possible to bring in a `.html` file as text content.)

mattdw:
The same corruption of unicode/utf-8 characters appears to happen in `TOML` modules as well. In this instance loaded with:

```javascript
const data = require("./data.toml");
```

Issue #5710, opened 2023-09-19
Custom extension for Text Loader


robobun:
I would like to be able to import `.env.example` files with the standard module import syntax:
```ts
import whatever from './.env.example';
```

I tried to use a **custom loader**:
```ts
import { plugin } from "bun";

plugin({
  name: "Env",
  async setup(build) {
    build.onLoad({ filter: /\.env\.example/ }, async (args) => {
      return {
        contents: await Bun.file(args.path).arrayBuffer(),
        loader: 'text',
      };
    })
  },
});
```
Which resulted in the following error:
> error: Expected loader to be one of "js", "jsx", "object", "ts", "tsx", "toml", or "json"
> error: script "dev:build" exited with code 1 (SIGHUP)
This error seems to come from the Module Loader itself: https://github.com/oven-sh/bun/blob/main/src/bun.js/bindings/ModuleLoader.cpp#L225
The `Loader` class for bun-types lists `text` as an option: https://github.com/oven-sh/bun/blob/main/packages/bun-types/bun.d.ts#L3073
It may also look as `text` isn't listed on the runtime plugin documentation: https://github.com/oven-sh/bun/blob/main/docs/runtime/plugins.md?plain=1#L269

_Originally reported on Discord: [`Custom extension for Text Loader`](https://discord.com/channels/876711213126520882/1153210632242470983/1153210632242470983)_

joshkolenko:
@robobun You can achieve this by returning an object with an `exports` property and the loader set to `object`. The export will then be treated as an ES module.

For example:
```js
import { plugin } from "bun";

plugin({
  name: "Env",
  async setup(build) {
    build.onLoad({ filter: /\.env\.example/ }, async (args) => {
      return {
        exports: { default: await Bun.file(args.path).arrayBuffer() },
        loader: 'object',
      };
    })
  },
});
```

Jarred-Sumner:
You can do `bun --loader:.html:text ./input.js`

PR #10456, opened 2024-04-23
Support `type` import attribute with `"text"`, `"json"`, `"toml"`, and `"file"`

### LINKED ISSUE #3449: text loader parses backslashes

Jarred-Sumner:
### What does this PR do?

This implements bundler and runtime support for the `"type"` Import Attribute for the following loaders
- `json`
- `toml`
- `text`

This also fixes encoding issues in the `"text"` loader at both runtime and bundletime.

This makes the following syntax work at both runtime and bundletime:
```js
import sourceCode from "./index.js" with { type: "text" };
import manifest from "./app.webmanifest" with { type: "json" };
const sourceCode2 = await import("./index.js", { with: { type: "text" } });
const manifest2 = await import("./app.webmanifest", { with: { type: "json" } });
```

Other runtimes support `type: "json"`, but it was previously not supported in bun. For consistency between runtime & bundler, we also support `type: 'toml'`, `type: 'text'` and `type: 'file'`.  Currently, we do not support importing the same file with different types but that can be added in a future PR when someone files an issue who ran into that.

Fixes #3449
Fixes https://github.com/oven-sh/bun/issues/10206
Fixes https://github.com/oven-sh/bun/issues/5710

There are runtime and bundler tests


src/bundler.zig
```diff
@@ -1597,7 +1597,7 @@ pub const Bundler = struct {
             },
             // TODO: use lazy export AST
             .text => {
```
dylan-conway, line 1599: should this be `.text, .file => {`?

src/js_printer.zig
```diff
@@ -206,12 +218,28 @@ pub fn quoteForJSONBuffer(text: []const u8, bytes: *MutableString, comptime asci
...
+                3 => text[i..][0..3].* ++ .{0},
+                4 => text[i..][0..4].*,
+                else => unreachable,
+            },
+            width,
+            i32,
+            0,
+        );
         if (canPrintWithoutEscape(i32, c, ascii_only)) {
             const remain = text[i + @as(usize, width) ..];
```
dylan-conway: Is it possible this and line 238 can also go out of bounds? I think `width` should be `clamped_width`

src/bundler.zig
Jarred-Sumner: No, file is handled separately

PR #10491, opened 2024-04-24
Ban `// autofix` in linter

Jarred-Sumner:
### What does this PR do?

Ban `// autofix` in linter

It is very lazy to commit `// autofix` comments

### DIFF
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -334,10 +334,9 @@ pub const ReadableStream = struct {
 
     pub fn fromPipe(
         globalThis: *JSGlobalObject,
-        parent: anytype,
+        _: anytype,
         buffered_reader: anytype,
     ) JSC.JSValue {
-        _ = parent; // autofix
         JSC.markBinding(@src());
         var source = FileReader.Source.new(.{
             .globalThis = globalThis,
@@ -1468,9 +1467,8 @@ pub const SinkDestructor = struct {
 
     pub export fn Bun__onSinkDestroyed(
         ptr_value: ?*anyopaque,
-        sink_ptr: ?*anyopaque,
+        _: ?*anyopaque,
     ) callconv(.C) void {
-        _ = sink_ptr; // autofix
         const ptr = Ptr.from(ptr_value);
 
         if (ptr.isNull()) {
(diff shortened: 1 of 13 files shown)


paperclover (approved):
great. autofix is an indication of forgetting something, be it using the variable or intentionally ignoring it.

src/shell/interpreter.zig
```diff
@@ -5796,8 +5777,8 @@ pub const Interpreter = struct {
                 }
             }
 
-            pub fn deinit(this: *Cat) void {
-                _ = this; // autofix
+            pub fn deinit(_: *Cat) void {
+                // TODO: @zackoverflow is this leaking memory? is there anything to free here?
```
Jarred-Sumner, line 5781: @zackradisic

src/shell/interpreter.zig
```diff
@@ -7258,8 +7231,8 @@ pub const Interpreter = struct {
                 return;
             }
 
-            pub fn deinit(this: *Ls) void {
-                _ = this; // autofix
+            pub fn deinit(_: *Ls) void {
+                // todo @zackoverflow: dos this leak any memory? do we need to free anything here?
```
Jarred-Sumner, line 7235: @zackradisic

src/bun.js/webcore/streams.zig
```diff
@@ -3184,8 +3172,7 @@ pub const FileSink = struct {
         return .{ .result = {} };
     }
 
-    pub fn flushFromJS(this: *FileSink, globalThis: *JSGlobalObject, wait: bool) JSC.Maybe(JSValue) {
-        _ = wait; // autofix
+    pub fn flushFromJS(this: *FileSink, globalThis: *JSGlobalObject, _: bool) JSC.Maybe(JSValue) {
```
dylan-conway, line 3175: I think unused parameters should keep the variable name if comptime logic requires them. Makes it a little more clear what it would be used for. Otherwise the parameter should be completely removed from the function signature

dylan-conway: also interesting 
Screenshot 2024-04-24 at 4 17 20 PM

Jarred-Sumner: yeah

PR #10492, opened 2024-04-24
watcher: major refactor to use `Maybe` types for better error reporting

paperclover:
Previously:

image

This was because:

1. Watchpack is watching files it honestly doesnt need to watch.
2. They check `.code` and ignore some error codes

This PR implements .code for these errors.


paperclover:
Draft initially to see if anything obvious is caught / improved in CI. will mark as reviewable once I think it is reviewable

src/bun.js/node/node_fs_watcher.zig
```diff
@@ -240,76 +266,56 @@ pub const FSWatcher = struct {
...
         if (!this.refTask()) {
             return;
         }
 
-        const encoding = this.encoding;
-        switch (event_type) {
-            inline .rename, .change => |eventType| {
-                const task = FSWatchTaskWindows.new(.{
-                    .ctx = this,
-                    .file_path = switch (encoding) {
```
Jarred-Sumner, line 294: add back optimization

Jarred-Sumner:
does not compile

paperclover:
regresses fs watch tests on windows

PR #10509, opened 2024-04-25
Fix : Enhance directory error handling

### LINKED ISSUE #10472: `bunx bun` in "access denied" folder crashes but should give better error

1. `cd "C:\Documents and Settings"`
2. `bunx bun`

It makes sense that it errors, but @paperdave said there should be a better message then just crash.
Also in a way as I'm using `bunx` it may not need to have access to the folder?

### Relevant log output

_No response_

### Stack Trace (bun.report)

Bun v1.1.5 ([#10383](https://github.com/oven-sh/bun/pull/10383)) on windows x86_64 [BunxCommand]

error: CouldntReadCurrentDirectory

- [`run_command.zig:862`](https://github.com/oven-sh/bun/blob/e3689e7e83507ae9c63dac7d785e41884daead66/src/cli/run_command.zig#L862): `configureEnvForRun`
- [`Allocator.zig:98`](https://github.com/oven-sh/bun/blob/e3689e7e83507ae9c63dac7d785e41884daead66/src/deps/zig/lib/std/mem/Allocator.zig#L98): `rawFree`
- [`cli.zig:1752`](https://github.com/oven-sh/bun/blob/e3689e7e83507ae9c63dac7d785e41884daead66/src/cli.zig#L1752): `start`

Features:

sossost:
### What does this PR do?

Fixes #10472 

Improve error messaging in configureEnvForRun for directory access failures
Detailed the error messages for CouldntReadCurrentDirectory in configureEnvForRun to provide clearer guidance on missing directory or insufficient permissions. This enhancement aims to help users better understand and resolve issues related to directory access failures.

Simplify handling of CouldntReadCurrentDirectory in handleRootError
Omitted verbose error logging for CouldntReadCurrentDirectory in handleRootError to reduce noise and enhance user experience. Since configureEnvForRun already provides comprehensive error information, this change minimizes redundancy and focuses on clean exit strategies in production environments.

- [ ] Code changes

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -312,14 +312,14 @@ pub const RunCommand = struct {
                 if (Environment.isDebug) {
                     Output.prettyError("[bun shell] ", .{});
                 }
-                Output.prettyErrorln("<r><d><magenta>$<r> <d><b>{s}<r>", .{combined_script});
+                Output.errGeneric("<r><d><magenta>$<r> <d><b>{s}<r>", .{combined_script});
                 Output.flush();
             }
 
             const mini = bun.JSC.MiniEventLoop.initGlobal(env);
             const code = bun.shell.Interpreter.initAndRunFromSource(ctx, mini, name, combined_script) catch |err| {
                 if (!silent) {
-                    Output.prettyErrorln("<r><red>error<r>: Failed to run script <b>{s}<r> due to error <b>{s}<r>", .{ name, @errorName(err) });
+                    Output.errGeneric("<r><red>error<r>: Failed to run script <b>{s}<r> due to error <b>{s}<r>", .{ name, @errorName(err) });
                 }
 
                 Global.exit(1);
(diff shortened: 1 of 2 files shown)


src/cli/run_command.zig
```diff
@@ -857,7 +857,7 @@ pub const RunCommand = struct {
             } else {
                 ctx.log.printForLogLevelWithEnableAnsiColors(Output.errorWriter(), false) catch {};
             }
-            Output.prettyErrorln("error loading current directory", .{});
+            Output.prettyErrorln(("Could not read the current directory. Please check if the directory exists and your application has the necessary permissions.", .{}));
```
paperclover: i don't think "Could not read the current directory. Please check [...]" is a better error message, as it doesn't tell the user *what the error is*. other programs that try to read this directory, such as `cmd.exe`'s dir builtin, `readdir()`, and `bun exec ls`. these all error with a specific error code, but it is being masked in this case with. that is the point of the issue.

image

paperclover (changes requested):
thanks for the contribution, but code should be tested before sent for review. by reading the diff i can tell that this code will not build. see https://bun.sh/docs/project/building-windows for building and testing bun on windows.

in [my comment in the issue](https://github.com/oven-sh/bun/issues/10472#issuecomment-2074038391), i state that both paths of this should get better error messages, this only touches one.

also, the call should be changed to either use `Output.err` or `Output.errGeneric` (many places need to be changed to use this)

sossost:
@paperdave Oh, thank you for the quick and kind review. As you said, I will further improve the code and make additional commits.

PR #10527, opened 2024-04-26
node:fs: fix arg parsing of readSync

nektro:
found this while locally setting up the nodejs test runner

https://nodejs.org/docs/latest/api/fs.html#fsreadsyncfd-buffer-offset-length-position

### DIFF
--- a/test/js/node/fs/fs.test.ts
+++ b/test/js/node/fs/fs.test.ts
@@ -1044,6 +1044,19 @@ describe("readSync", () => {
     }
     closeSync(fd);
   });
+
+  it("works with offset + length passed but not position", () => {
+    const fd = openSync(import.meta.dir + "/readFileSync.txt", "r");
+    const four = new Uint8Array(4);
+    {
+      const count = readSync(fd, four, 0, 4);
+      const u32 = new Uint32Array(four.buffer)[0];
+      expect(u32).toBe(firstFourBytes);
+      expect(count).toBe(4);
+    }
+    closeSync(fd);
+  });
+
   it("works without position set", () => {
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
where are the tests?

test/js/node/fs/fs.test.ts
```diff
@@ -1044,6 +1044,19 @@ describe("readSync", () => {
...
     closeSync(fd);
   });
+
+  it("works with offset + length passed but not position", () => {
+    const fd = openSync(import.meta.dir + "/readFileSync.txt", "r");
+    const four = new Uint8Array(4);
+    {
+      const count = readSync(fd, four, 0, 4);
+      const u32 = new Uint32Array(four.buffer)[0];
+      expect(u32).toBe(firstFourBytes);
```
Jarred-Sumner, line 1054: where is `firstFourBytes` defined

test/js/node/fs/fs.test.ts
```diff
@@ -1044,6 +1044,19 @@ describe("readSync", () => {
     }
     closeSync(fd);
   });
+
+  it("works with offset + length passed but not position", () => {
+    const fd = openSync(import.meta.dir + "/readFileSync.txt", "r");
+    const four = new Uint8Array(4);
+    {
+      const count = readSync(fd, four, 0, 4);
```
Jarred-Sumner, line 1052: this tests length, but not offset. can this test both?
```suggestion
      const count = readSync(fd, four, 1, 5);
```

test/js/node/fs/fs.test.ts
nektro: at the top of this describe block

PR #10531, opened 2024-04-26
Support comments & trailing commas in require/import `package.json`

Jarred-Sumner:
### What does this PR do?

Supports `require("./package.json")` and `await import("./package.json")` when the json file has comments or trailing commas

There is a test


paperclover (approved):

silverwind:
Which JSONC parser is in use here and is it available as an npm module?

If the ecosystem is to standardize on JSONC for package.json, it ought to use the same parser.

PR #10543, opened 2024-04-26
Fixes #4718

### LINKED ISSUE #4718: SvelteKit fetch issue

Darwin 22.6.0 arm64 arm

When using SvelteKit fetch, it appears that it is clashing with the native/runtime fetch implementation of Bun. Clashing being defined that it seems to be missing request params like headers and such. For example. Take a `PageLoad` call. 

```ts
import type { PageLoad } from './$types';

export const load: PageLoad = async ({ parent, fetch }) => {
    const { token } = await parent();
    const res = await fetch(`http://localhost:3000/v1/approvals`, {
        headers: {
            Authorization: `Bearer ${token}`
        }
    });

    const d = await res.json();

    return { approvals: d.data };
};
```

When sending to the server the request headers seen are:

```json
{
  "accept": "*/*",
  "accept-encoding": "gzip, deflate",
  "connection": "keep-alive",
  "user-agent": "Bun/1.0.0"
}
```

A simple change to remove the PageLoad fetch, will result in proper headers being sent.

```ts
import type { PageLoad } from './$types';

export const load: PageLoad = async ({ parent }) => {
    const { token } = await parent();
    const res = await fetch(`http://localhost:3000/v1/approvals`, {
        headers: {
            Authorization: `Bearer ${token}`
        }
    });

    const d = await res.json();

    return { approvals: d.data };
};
```

request headers from the server:

```json
{
  "accept": "*/*",
  "accept-encoding": "gzip, deflate",
  "authorization": "***",
  "connection": "keep-alive",
  "content-type": "application/json",
  "user-agent": "Bun/1.0.0"
}
```

Not sure what the expected behavior should be here. Presumably the SvelteKit fetch call would be used as it is part of the function scope?

### What do you see instead?

_No response_

### Additional information

Outside the unexpected behavior I think the major drawback of using the native/runtime fetch call is a change in behavior for SvelteKit apps coming from other JS runtimes. This will also prevent the use of SvelteKit [`HandleFetch`](https://kit.svelte.dev/docs/hooks#server-hooks-handlefetch) hook. At the very least it might be helpful as documentation?

Jarred-Sumner:
### What does this PR do?

`fetch()`, `new Response`, `new Request` are supposed to call getters for the request & response objects. Previously, we would skip calling getters if we knew it was a `Request` or `Response` object passed. But, this is incorrect. We have to call getters if it's not exactly the same type as the native type - for example, if they have a subclass or if they override `method` or `url`. We probably also need to do this for `signal`.

To make this work, this PR adds `JSValue.asDirect` to our bindings generator which checks if the `JSValue` has the same StructureID of the corresponding native type. If it does, we can skip the getters. If it does not, we cannot skip the getters.

Fixes #4718

There is a test

### DIFF
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -3805,6 +3805,16 @@ pub const JSValue = enum(JSValueReprInt) {
         cppFn("push", .{ value, globalObject, out });
     }
 
+    /// Return the pointer to the wrapped object only if it is a direct instance of the type.
+    /// If the object does not match the type, return null.
+    /// If the object is a subclass of the type or has mutated the structure, return null.
+    /// Note: this may return null for direct instances of the type if the user adds properties to the object.
+    pub fn asDirect(value: JSValue, comptime ZigType: type) ?*ZigType {
+        bun.assert(value.isCell()); // you must have already checked this.
+
+        return ZigType.fromJSDirect(value);
+    }
+
     pub fn as(value: JSValue, comptime ZigType: type) ?*ZigType {
         if (value.isEmptyOrUndefinedOrNull())
             return null;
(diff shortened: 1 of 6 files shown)


dylan-conway (approved):

Bewinxed:
Tried canary and 1.1.6

const request = {…,
method: ‘post’}
 await fetch(request)

still has the same issue @Jarred-Sumner 

but directly fetching like:

await fetch(‘url’, {…, method: ‘post’}) works 

so half fixed? Unless it’s a third party library

PR #10590, opened 2024-04-27
fix: seq, dirname, basename should exit with 0 even when they don't have stdout

### LINKED ISSUE #10589: Bun shell hangs when using seq, dirname, or basename without stdout

1.1.5

Run `bun exec 'echo $(basename a)'` in your console

It should print `a` end exit with code 0.

### What do you see instead?

the bun shell interpreter hangs

### Additional information

_No response_

RanolP:
### What does this PR do?

Fixes #10589 
This makes seq, dirname, and basename commands exit with code 0 when they do the job successfully and don't have stdout in bun shell.

- [x] Code changes

Locally tested with `bun-debug exec 'echo $(basename a)'`.
I wrote automated tests ~~but have not yet tested them (due to WSL crash on bun-debug test).~~
**UPDATE (04/29):** I've locally verified the tests on my macOS 14 Sonoma.

- [x] JSValue used outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed


zackradisic (approved):
Thanks!

PR #10620, opened 2024-04-28
fix `node-module-module.test.js` on windows

dylan-conway:
### What does this PR do?
uses `path.dirname` for `_compile` and `_resolveLookupPaths`

manually and added more filename paths


Jarred-Sumner:
Thank you

PR #10625, opened 2024-04-29
Fixes #10624

### LINKED ISSUE #10624: error when use bun build express.ts --compile --minify --sourcemap --outfile dist/express

_No response_

### Relevant log output

```shell
Bun v1.1.6 (e58d67b4) Windows x64
Args: "C:\Program Files\nodejs\node_modules\bun\bin\bun.exe", "build", "express.ts", "--compile", "--minify", "--sourcemap", "--outfile", "dist/express"
Features: bunfig define dotenv tsconfig(9) 
Elapsed: 6791ms | User: 6125ms | Sys: 0ms
RSS: 120.21MB | Peak: 120.29MB | Commit: 0.22GB | Faults: 30391

panic(thread 3616): integer overflow
```

### Stack Trace (bun.report)

Bun v1.1.6 ([#10596](https://github.com/oven-sh/bun/pull/10596)) on windows x86_64 [BuildCommand]

**panic**: integer overflow

- [`sourcemap.zig:443`](https://github.com/oven-sh/bun/blob/e58d67b46875280bc9002add4c3fae45472cb346/src/sourcemap/sourcemap.zig#L443): `advance`
- [`sourcemap.zig:418`](https://github.com/oven-sh/bun/blob/e58d67b46875280bc9002add4c3fae45472cb346/src/sourcemap/sourcemap.zig#L418): `advance`
- [`bundle_v2.zig:6365`](https://github.com/oven-sh/bun/blob/e58d67b46875280bc9002add4c3fae45472cb346/src/bundler/bundle_v2.zig#L6365): `generateChunkJS`
- [`output.zig:402`](https://github.com/oven-sh/bun/blob/e58d67b46875280bc9002add4c3fae45472cb346/src/output.zig#L402): `flush`
- ??? at `0x1257d` in /KERNEL32.DLL
- ??? at `0x5aa48` in /ntdll.dll

Jarred-Sumner:
### What does this PR do?

Fixes #10624

There is an integration test

### DIFF
--- a/src/sourcemap/sourcemap.zig
+++ b/src/sourcemap/sourcemap.zig
@@ -449,6 +449,16 @@ pub const LineColumnOffset = struct {
             var iter = strings.CodepointIterator.initOffset(input, i);
             var cursor = strings.CodepointIterator.Cursor{ .i = @as(u32, @truncate(iter.i)) };
             _ = iter.next(&cursor);
+
+            // Given a null byte, cursor.width becomes 0
+            // This can lead to integer overflow, crashes, or hangs.
+            // https://github.com/oven-sh/bun/issues/10624
+            if (cursor.width == 0) {
+                columns += 1;
+                offset = i + 1;
+                continue;
+            }
+
             offset = i + cursor.width;
 
             switch (cursor.c) {
--- a/test/js/third_party/body-parser/express-bun-build-compile.test.ts
+++ b/test/js/third_party/body-parser/express-bun-build-compile.test.ts
@@ -1,0 +1,18 @@
+import { expect, test } from "bun:test";
+import { join } from "path";
+import { $ } from "bun";
+import "harness";
+import { bunExe, tempDirWithFiles } from "harness";
+
+$.throws(true);
+
+// https://github.com/oven-sh/bun/issues/10624
+test("Express hello world app supports bun build --compile --minify --sourcemap", async () => {
+  const dir = tempDirWithFiles("express-bun-build-compile", {
+    "out.exe": "",
+  });
+
+  const file = join(dir, "out.exe");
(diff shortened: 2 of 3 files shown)


dylan-conway (approved):
Looks good. We might want to change `CodepointIterator.next()` to fill the cursor with width 1 in this situation because a null byte isn't zero width

src/sourcemap/sourcemap.zig
```diff
@@ -449,6 +449,16 @@ pub const LineColumnOffset = struct {
...
             _ = iter.next(&cursor);
+
+            // Given a null byte, cursor.width becomes 0
+            // This can lead to integer overflow, crashes, or hangs.
+            // https://github.com/oven-sh/bun/issues/10624
+            if (cursor.width == 0) {
+                columns += 1;
+                offset = i + 1;
+                continue;
+            }
```
gvilums, line 460: I think we should use the return value of `iter.next` (which returns `false` on a byte sequence length of 0) to detect the invalid case, e.g.
```suggestion
            // Given a null byte, cursor.width becomes 0
            // This can lead to integer overflow, crashes, or hangs.
            // https://github.com/oven-sh/bun/issues/10624
            if (!iter.next(&cursor)) {
                columns += 1;
                offset = i + 1;
                continue;
             }            
```
This would also make the code less likely to break if we change the default `width` value in `CodepointIterator.Cursor`

Jarred-Sumner: .next() returns false when it reaches the end

It shouldn't advance columns by 1 when the end is reached

gvilums: makes sense 👍

PR #10628, opened 2024-04-29
Speculative fix for alignment crash in directory iterator on Windows

Jarred-Sumner:
### What does this PR do?

We've received about 16 crash reports for an alignment issue in the directory iterator code

So let's loosen the alignment requirements here and see if it goes away

> This structure must be aligned on a LONGLONG (8-byte) boundary. If a buffer contains two or more of these structures, the NextEntryOffset value in each entry, except the last, falls on an 8-byte boundary.

See also: https://github.com/ziglang/zig/pull/17958

### DIFF
--- a/src/bun.js/node/dir_iterator.zig
+++ b/src/bun.js/node/dir_iterator.zig
@@ -218,7 +218,7 @@ pub fn NewIterator(comptime use_windows_ospath: bool) type {
                         var io: w.IO_STATUS_BLOCK = undefined;
                         if (self.first) {
                             // > Any bytes inserted for alignment SHOULD be set to zero, and the receiver MUST ignore them
-                            @memset(self.buf, 0);
+                            @memset(&self.buf, 0);
                         }
 
                         const rc = w.ntdll.NtQueryDirectoryFile(
--- a/src/bun.js/node/dir_iterator.zig
+++ b/src/bun.js/node/dir_iterator.zig
@@ -185,6 +185,8 @@ pub fn NewIterator(comptime use_windows_ospath: bool) type {
         },
         .windows => struct {
             const FILE_DIRECTORY_INFORMATION = std.os.windows.FILE_DIRECTORY_INFORMATION;
+            // While the official api docs guarantee FILE_BOTH_DIR_INFORMATION to be aligned properly
+            // this may not always be the case (e.g. due to faulty VM/Sandboxing tools)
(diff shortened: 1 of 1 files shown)


src/bun.js/node/dir_iterator.zig
```diff
@@ -270,7 +282,7 @@ pub fn NewIterator(comptime use_windows_ospath: bool) type {
                         bun.sys.syslog("NtQueryDirectoryFile({}) = {d}", .{ bun.toFD(self.dir.fd), self.end_index });
                     }
 
-                    const dir_info: *w.FILE_DIRECTORY_INFORMATION = @ptrCast(@alignCast(&self.buf[self.index]));
+                    const dir_info: FILE_DIRECTORY_INFORMATION_PTR = @ptrCast(@alignCast(&self.buf[self.index]));
                     if (dir_info.NextEntryOffset != 0) {
```
zackradisic, line 288: If the struct is unaligned, even if we cast alignment, do we have to worry about unaligned reads? Technically that is undefined behavior right?

src/bun.js/node/dir_iterator.zig
```diff
@@ -184,8 +184,16 @@ pub fn NewIterator(comptime use_windows_ospath: bool) type {
             }
         },
         .windows => struct {
+            const FILE_DIRECTORY_INFORMATION = std.os.windows.FILE_DIRECTORY_INFORMATION;
+            const FILE_DIRECTORY_INFORMATION_PTR = *align(2) FILE_DIRECTORY_INFORMATION;
```
zackradisic, line 190: I think a more descriptive comment here would be nice, or a link to the Zig PR

I like the one from the Zig PR:
```
// While the official api docs guarantee FILE_BOTH_DIR_INFORMATION to be aligned properly
// this may not always be the case (e.g. due to faulty VM/Sandboxing tools)
```

zackradisic (commented):
this is an insane bug

i left some ocmments

src/bun.js/node/dir_iterator.zig
Jarred-Sumner: if the expected alignment is 2 but the alignment we're seeing is 8, then it is also 2 so that should be okay

src/bun.js/node/dir_iterator.zig
Jarred-Sumner: added

Issue #5091, opened 2023-09-12
Get Sentry-like error reporting working


Jarred-Sumner:
https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor

- [x] `process.on("unhandledRejection", cb)`
- [x] `process.on("unhandledException", cb)`
- [x] `process.on("uncaughtExceptionMonitor", cb)` 
- [x] `process.setUncaughtExceptionCaptureCallback(cb)`

This lets Sentry get error reporting working. 

Ideally, the fix would go here:

https://github.com/oven-sh/bun/blob/7ce7f08a87079ff4827fc633f8c2d1a46c768633/src/bun.js/javascript.zig#L629-L640

We would add an unhandled rejection scope for either rejections or exceptions and have some way to distinguish between the two.  This one would call into JS potentially for multiple handlers. The part that gets a little tricky is in unit tests where we rely on this behavior for `expect().toThrow()`

cc @dcramer

HazAT:
We will release the Sentry SDK this week - Most things work (manually) once this lands we'll shoot out an update supporting it. Few users already asked for Sentry Bun SDK 😅

AbhiPrasad:
We just shipped https://www.npmjs.com/package/@sentry/bun as part of https://github.com/getsentry/sentry-javascript/releases/tag/7.70.0.

```
bun add @sentry/bun
```

HazAT:
I don't want to spam here 😅  but with version `7.71.0` of the Sentry Bun SDK we wrap `Bun.serve` automatically https://github.com/getsentry/sentry-javascript/pull/9080 capturing Errors and Transactions that happen during requests.

colindotfun:
#7472 still seems to be an issue with: 
```
- `@sentry/bun version 7.93.0`
- `bun version 1.0.20`
```

GingerAdonis:
I'm aware that the team has lots of things to work on, but I'd like to say that this is likely a blocker for many to use Bun in production.

ohbob:
^^ please fix

thecrabcode:
I agree with @GingerAdonis, this is very important for many devs. Do you guys have any estimates on when this could potentially be fixed?

AbhiPrasad:
[...]

PR #10633, opened 2024-04-29
Use error handling callback in more places, emit 1015 when WSS TLS handshaking fails, micro-optimize ServerWebSocket, fix bug in util.inspect exceptions

### LINKED ISSUE #10619: windows client websocket on error event causes crash

1.1.6

windows 11

Upon error handling, when using the value of event in any kind, bun crashes,

Code used for testing:
```ts
const socket = new WebSocket("wss://127.0.0.1:1000", "json")

socket.addEventListener("open", ev => {
    console.log("websocket open", ev)
})  

socket.addEventListener("message", ev => {
    console.log("messaged recieved", ev)
})

socket.addEventListener("close", ev => {
    console.log("websocket closed", ev)
})

socket.addEventListener("error", ev => {
    console.log("websocket errored!")
    console.log(ev)
});
```

Not crashing. 

### What do you see instead?

The following error message:

```
websocket errored!
Bun v1.1.6 (e58d67b4) Windows x64
Args: "C:\Users\Main1\.bun\bin\bun.exe", "run", "test.ts"
Builtins: "bun:main" 
Elapsed: 2130ms | User: 62ms | Sys: 15ms
RSS: 96.36MB | Peak: 96.36MB | Commit: 0.14GB | Faults: 23735

panic(main thread): attempt to use null value

...
```
### Additional information

Only occurs on windows, everything is working fine on linux based systems.

Jarred-Sumner:
### What does this PR do?

- Prepare for #5091 by swapping almost all usages of `vm.runErrorHandler` to `vm.onError` which calls the currently registered error handler instead of the global one. This may cause test failures, we will see.
- Shrink `ServerWebSocket` struct size from 28 bytes -> 24 bytes, which should slightly improve performance (rounding to alignment it was probably 32 bytes before)
- Micro-optimize console.log on objects which implement util.inspect.custom by caching the structure. This was barely measurable, might not be worth it
- Fixes a bug where throwing while logging a custom inspect would print `[native code]` because it was not correctly checking for exceptions
- This adds a test for `onConnectError` and fixes a double free there. Fixes #10619. 
- `WebSocket` now emits close code 1015 for TLS handshaking failure (cc @cirospaciari). Browsers do not do this. But on a server, this is useful information (image from the websocket spec).
image

Along the way, noticed a bug with unhandled errors. We seem to continue executing in cases where we should not continue executing. I think it's simple to fix, but it is very much a breaking change until we do #5091

There are tests

### DIFF
--- a/src/bun.js/bindings/webcore/WebSocket.cpp
+++ b/src/bun.js/bindings/webcore/WebSocket.cpp
@@ -260,19 +260,6 @@ ExceptionOr<void> WebSocket::connect(const String& url, const String& protocol)
     return connect(url, Vector<String> { 1, protocol }, std::nullopt);
 }
 
-void WebSocket::failAsynchronously()
-{
-    // queueTaskKeepingObjectAlive(*this, TaskSource::WebSocket, [this] {
-    // We must block this connection. Instead of throwing an exception, we indicate this
-    // using the error event. But since this code executes as part of the WebSocket's
-    // constructor, we have to wait until the constructor has completed before firing the
-    // event; otherwise, users can't connect to the event.
-    this->dispatchErrorEventIfNeeded();
-    // this->();
-    // this->stop();
-    // });
-}
-
 static String resourceName(const URL& url)
 {
     auto path = url.path();
@@ -304,7 +291,6 @@ ExceptionOr<void> WebSocket::connect(const String& url, const Vector<String>& pr
     m_url = URL { url };
 
     ASSERT(scriptExecutionContext());
-    auto& context = *scriptExecutionContext();
 
     if (!m_url.isValid()) {
         // context.addConsoleMessage(MessageSource::JS, MessageLevel::Error, );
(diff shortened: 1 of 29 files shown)


cirospaciari (approved):

src/bun.js/bindings/webcore/WebSocket.cpp
```diff
@@ -1158,15 +1143,19 @@ void WebSocket::didReceiveBinaryData(const AtomString& eventName, const std::spa
...
 }
 
-void WebSocket::didReceiveClose(CleanStatus wasClean, unsigned short code, WTF::String reason)
+void WebSocket::didReceiveClose(CleanStatus wasClean, unsigned short code, WTF::String reason, bool isConnectionError)
 {
     // LOG(Network, "WebSocket %p didReceiveErrorMessage()", this);
     // queueTaskKeepingObjectAlive(*this, TaskSource::WebSocket, [this, reason = WTFMove(reason)] {
     if (m_state == CLOSED)
         return;
+    const bool wasConnecting = m_state == CONNECTING;
```
dylan-conway, line 1152: can't this be `isConnectionError` when the function is called instead?
```
 didReceiveClose(CleanStatus::NotClean, 1006, message, m_state == CONNECTING);
 ```

Jarred-Sumner: no, see the next line

PR #10661, opened 2024-04-29
Make Windows command line argument parsing more robust

zackradisic:
### What does this PR do?

This fixes a failing test in bunshell.test.ts related to some discrepancies in Windows command line argument parsing

Specifically the problem is that `std.process.argsAlloc()` was incorrectly splitting the command line string into argv

The fix is just to use the Windows API functions that do this

Note that Zig 0.12 shipped with command line arg parsing fixes, which _may_ fix this, so this serves more as a temporary
work around

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -2091,8 +2091,46 @@ pub const Stat = if (Environment.isWindows) windows.libuv.uv_stat_t else std.os.
 
 pub var argv: [][:0]const u8 = &[_][:0]const u8{};
 
+extern "c" fn CommandLineToArgvW([*:0]u16, *c_int) ?[*][*:0]u16;
+
 pub fn initArgv(allocator: std.mem.Allocator) !void {
-    argv = try std.process.argsAlloc(allocator);
+    if (comptime !Environment.isWindows) {
+        argv = try std.process.argsAlloc(allocator);
+    } else {
+        // Zig's implementation of `std.process.argsAlloc()`on Windows platforms
+        // is not reliable, specifically the way it splits the command line string.
+        //
+        // For example, an arg like "foo\nbar" will be
+        // erroneously split into two arguments: "foo" and "bar".
+        //
+        // To work around this, we can simply call the Windows API functions
+        // that do this for us.
+        //
+        // Updates in Zig v0.12 related to Windows cmd line parsing may fix this,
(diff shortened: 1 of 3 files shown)


src/bun.zig
```diff
@@ -2091,8 +2091,46 @@ pub const Stat = if (Environment.isWindows) windows.libuv.uv_stat_t else std.os.
 
 pub var argv: [][:0]const u8 = &[_][:0]const u8{};
 
+extern "c" fn CommandLineToArgvW([*:0]u16, *c_int) ?[*][*:0]u16;
```
gvilums: This should go in `windows.zig`

gvilums: also it should use the winapi calling convention
```suggestion
extern "kernel32" fn CommandLineToArgvW([*:0]u16, *c_int) callconv(windows.WINAPI) ?[*][*:0]u16;
```

gvilums (changes requested):
Logic looks good, see suggestion

src/bun.zig
```diff
@@ -2091,8 +2091,46 @@ pub const Stat = if (Environment.isWindows) windows.libuv.uv_stat_t else std.os.
...
+        //
+        // To work around this, we can simply call the Windows API functions
+        // that do this for us.
+        //
+        // Updates in Zig v0.12 related to Windows cmd line parsing may fix this,
+        // see (here: https://ziglang.org/download/0.12.0/release-notes.html#Windows-Command-Line-Argument-Parsing),
+        // so this may only need to be a temporary workaround.
+        const cmdline_ptr = std.os.windows.kernel32.GetCommandLineW();
+        var length: c_int = 0;
+        const argvu16_ptr = CommandLineToArgvW(cmdline_ptr, &length) orelse {
```
paperclover: outofmemory is handled here, which implies allocation, but i dont see this being freed. i dont think its a big issue since its a small allocation with a max upper bound of 64k.

paperclover (approved):
nice. i'll look in the zig upgrade later to see if this is fixed.

src/bun.zig
zackradisic: According to the GetCommandLineW [docs](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinew#remarks) there's no need to free it:
> The lifetime of the returned value is managed by the system, applications should not free or modify this value.

Added a comment noting that

Jarred-Sumner: it should never be freed

PR #10662, opened 2024-04-29
fix flaky `bun-create.test.ts` with git templates

dylan-conway:
### What does this PR do?
checks for both `GITHUB_TOKEN` and `GITHUB_ACCESS_TOKEN` in `bun create` and `bun upgrade`

CI

### DIFF
--- a/src/cli/upgrade_command.zig
+++ b/src/cli/upgrade_command.zig
@@ -213,17 +213,17 @@ pub const UpgradeCommand = struct {
 
         if (env_loader.map.get("GITHUB_TOKEN") orelse env_loader.map.get("GITHUB_ACCESS_TOKEN")) |access_token| {
             if (access_token.len > 0) {
-                headers_buf = try std.fmt.allocPrint(allocator, default_github_headers ++ "Access-TokenBearer {s}", .{access_token});
+                headers_buf = try std.fmt.allocPrint(allocator, default_github_headers ++ "AuthorizationBearer {s}", .{access_token});
                 try header_entries.append(
                     allocator,
                     Headers.Kv{
                         .name = Api.StringPointer{
-                            .offset = accept.value.length + accept.value.offset,
-                            .length = @as(u32, @intCast("Access-Token".len)),
+                            .offset = accept.value.offset + accept.value.length,
+                            .length = @as(u32, @intCast("Authorization".len)),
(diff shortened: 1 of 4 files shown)


src/cli/upgrade_command.zig
```diff
@@ -211,7 +211,7 @@ pub const UpgradeCommand = struct {
             ),
         );
 
-        if (env_loader.map.get("GITHUB_ACCESS_TOKEN")) |access_token| {
+        if (env_loader.map.get("GITHUB_TOKEN") orelse env_loader.map.get("GITHUB_ACCESS_TOKEN")) |access_token| {
             if (access_token.len > 0) {
                 headers_buf = try std.fmt.allocPrint(allocator, default_github_headers ++ "Access-TokenBearer {s}", .{access_token});
```
Electroid: Where did `Access-TokenBearer` come from? Doesn't appear here: https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28

dylan-conway: updated

src/cli/upgrade_command.zig
```diff
@@ -211,19 +211,19 @@ pub const UpgradeCommand = struct {
             ),
         );
 
-        if (env_loader.map.get("GITHUB_ACCESS_TOKEN")) |access_token| {
+        if (env_loader.map.get("GITHUB_TOKEN") orelse env_loader.map.get("GITHUB_ACCESS_TOKEN")) |access_token| {
             if (access_token.len > 0) {
-                headers_buf = try std.fmt.allocPrint(allocator, default_github_headers ++ "Access-TokenBearer {s}", .{access_token});
+                headers_buf = try std.fmt.allocPrint(allocator, default_github_headers ++ "AuthorizationBearer {s}", .{access_token});
```
Electroid, line 216: I think this should just be `Bearer`?

Electroid: `Authorization` is the header name.
`Bearer {s}` is the header value.

dylan-conway: The names and values are in the same buffer, this is just allocating enough space for both. The entry stores the offsets and lengths for `Authorization` and `Bearer <token>`

PR #10667, opened 2024-04-29
no auth token passed as data in the example

DaleSeo:
### What does this PR do?

`<{ authToken: string }>` would be misleading when there is no auth token passed as data. 

- [ ] Code changes


DaleSeo:
@Electroid, could you please review this simple docs change?

Electroid:
This isn't nessecary.

PR #10670, opened 2024-04-30
Make verdaccio available for all tests in CI

dylan-conway:
### What does this PR do?
testing

- [ ] Code changes


test/cli/install/registry/bun-install-registry.test.ts
```diff
@@ -5,40 +5,16 @@ import { mkdtempSync, realpathSync } from "fs";
...
   toBeValidBin,
   toHaveBins,
 });
 
-var verdaccioServer: ChildProcess;
 var testCounter: number = 0;
-var port: number = 4873;
 var packageDir: string;
 
-beforeAll(async () => {
```
Jarred-Sumner, line 21: @dylan-conway this won't quite work. We need to check if it's CI and then use the CI version, but in development we still want to do this. I suggest having an `isCI`

PR #10674, opened 2024-04-30
Fix crash in Windows PipeReader

### LINKED ISSUE #10660: Fetch in next.js with --bun flag

_No response_

### Relevant log output

_No response_

### Stack Trace (bun.report)

Bun v1.1.6 ([#10627](https://github.com/oven-sh/bun/pull/10627)) on windows x86_64 [RunAsNodeCommand]

**panic**: Internal assertion failure

- [`debug.zig:373`](https://github.com/oven-sh/bun/blob/7af0ed164a84411ae952760ab494ecab3ef06c7b/src/deps/zig/lib/std/debug.zig#L373): `panicExtra`
- [`debug.zig:348`](https://github.com/oven-sh/bun/blob/7af0ed164a84411ae952760ab494ecab3ef06c7b/src/deps/zig/lib/std/debug.zig#L348): `panic`
- [`output.zig:356`](https://github.com/oven-sh/bun/blob/7af0ed164a84411ae952760ab494ecab3ef06c7b/src/output.zig#L356): `panic`
- [`bun.zig:3119`](https://github.com/oven-sh/bun/blob/7af0ed164a84411ae952760ab494ecab3ef06c7b/src/bun.zig#L3119): `assertionFailure`
- [`PipeReader.zig:420`](https://github.com/oven-sh/bun/blob/7af0ed164a84411ae952760ab494ecab3ef06c7b/src/io/PipeReader.zig#L420): `onFileRead`
- `uv__work_done`
- `??`
- `uv_run`
- [`uws.zig:2619`](https://github.com/oven-sh/bun/blob/7af0ed164a84411ae952760ab494ecab3ef06c7b/src/deps/uws.zig#L2619): `run`
- `JSC__VM__holdAPILock`

Features: Bun.stderr, Bun.stdin, Bun.stdout, abort_signal, dotenv, fetch, http_server, spawn, transpiler_cache, tsconfig_paths, tsconfig

paperclover:
### What does this PR do?

Fixes #10660, a crash impacting Next.js dev server while reading files.

To make better panics, I'm introducing two new functions:

- `unexpectedTag`, for use in a switch `else => |t| unexpectedTag`
- `releaseNonNull`, named after the method of the same name in WebKit's `Ref`. This is a release-mode safety checked version of `.?` and should be used if one is not confident 

This fixes the test locally for me, but I don't think it will fix CI as CI was running into a different test failure.

### DIFF
--- a/packages/bun-internal-test/src/runner.node.mjs
+++ b/packages/bun-internal-test/src/runner.node.mjs
@@ -502,7 +502,8 @@ if (failing_tests.length) {
       .replace(/^::(group|endgroup|error|warning|set-output|add-matcher|remove-matcher).*$/gm, "");
 
     if (failing_output.length > 1024 * 64) {
-      failing_output = failing_output.slice(0, 1024 * 64) + `\n\n[truncated output (length: ${failing_output.length})]`;
+      failing_output =
+        failing_output.slice(0, 1024 * 64) + `\n\n[truncated output (length: ${failing_output.length})]\n`;
     }
 
     report += failing_output;
--- a/src/Global.zig
+++ b/src/Global.zig
@@ -100,12 +100,16 @@ pub fn exit(code: u8) noreturn {
 
 pub fn exitWide(code: u32) noreturn {
     if (bun.crash_handler.panicking.load(.SeqCst) != 0) {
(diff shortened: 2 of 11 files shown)


paperclover:
Does not fix the crash

PR #10678, opened 2024-04-30
node:fs.fdatasync: include fd in Error when thrown

nektro:
many others do, but this one did not


Jarred-Sumner:
Where is the test?

PR #10679, opened 2024-04-30
node:fs.readdir: should not fail even if folder contains self-referential symlinks

nektro:
discovered this when trying to run our nodejs test runner with bun


Jarred-Sumner:
Where is the test?

Jarred-Sumner:
@nektro test failures on linux aarch64

image

PR #10728, opened 2024-05-01
Move some dns code into bun.dns

Jarred-Sumner:
### What does this PR do?

In preparation for DNS fixes, move some code from bun's JS DNS API to an internal `bun.dns` namespace


paperclover (approved):
assuming this only moves code and not editing any of it.

PR #10734, opened 2024-05-01
Re-enable support_jsxs_in_jsx_transform flag

### LINKED ISSUE #10733: Re-enable jsxs feature flag

Use case: 

I want to publish this component to NPM:

```tsx
// file.ts
export const MyComp: React.FC = (
  <div>
    {"Hello"}
    {"World"}
  </div>
);
```

I compile this using `NODE_ENV=production bun build file.tsx --external=react`:

```js
// file.tsx
import {
jsx
} from "react/jsx-runtime";
var MyComp = jsx("div", {
  children: [
    "Hello",
    "World"
  ]
});
export {
  MyComp
};
```

And then publish the component to NPM.
If a user installs it and imports it in Next.js, they get

```
Warning: Each child in a list should have a unique “key” prop.
```

This error does not appear with `tsc` which generates:

```js
import { jsxs as _jsxs } from "react/jsx-runtime";
export const MyComp = (_jsxs("div", { children: ["Hello", "World"] }));
```

The difference is that tsc uses `jsxs` and bun uses `jsx` 

Enabling the `support_jsxs_in_jsx_transform ` feature flag again https://github.com/oven-sh/bun/pull/10734

I considered find and replacing `jsx` with `jsxs` in the output.  
However, using `jsxs` for everything will also cause React to throw errors.

JonnyBurger:
### What does this PR do?

Re-enables support for using the `jsxs` function if React children are static

- [x] Code changes

I built Bun from source and verified that it has the desired behavior:

**Input**
```tsx
// file.ts
export const MyComp: React.FC = (
  <div>
    {"Hello"}
    {"World"}
  </div>
);
```
This should use `jsxs`, otherwise if imported in development, React will give a "duplicated key" error. 

**Running `./build-release/bun build file.tsx --external=react`:**

```tsx
// file.tsx
import {
jsxDEV
} from "react/jsx-dev-runtime";
var MyComp = jsxDEV("div", {
  children: [
    "Hello",
    "World"
  ]
}, undefined, true, undefined, this);
export {
  MyComp
};
```

This seems right, jsxs is not supported with the dev runtime

**Running `NODE_ENV=production ./build-release/bun build file.tsx --external=react`**

```tsx
// file.tsx
import {
jsxs
} from "react/jsx-runtime";
var MyComp = jsxs("div", {
  children: [
    "Hello",
    "World"
  ]
});
export {
  MyComp
};
```

This seems right now, the duplicate key errors are gone.  
At Remotion, we have shipped `jsxs` directives in production always ([output by rollup](https://www.unpkg.com/@remotion/player@4.0.100/dist/esm/index.mjs)) and it never led to problems.  

I'm not sure where the feature flag and it's comment are coming from. Therefore I edited the description of the feature flag.

I also used the binary with the feature flag enabled to build the whole Remotion monorepo whose ESM modules depend on each other heavily and at least in our codebase, everything is green and duplicate key errors disappeared.

Fixes #10733


Jarred-Sumner:
These test failures are flaky CI + permissions issues I think

dylan-conway:
Thank you, going to merge this. Originally this was disabled due to bugs with `NODE_ENV` detection

dylan-conway (approved):

PR #10746, opened 2024-05-01
Remove `util.styleText` type overrides because `@types/node` has them now

mangs:
### What does this PR do?

A [previous PR](https://github.com/oven-sh/bun/pull/9945) added types for `util.styleText` because `@types/node` was missing it. `@types/node` [just added it](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69212) so we can remove the maintenance burden and rely on those types again.

- [ ] Code changes


Jarred-Sumner:
Thank you

PR #10753, opened 2024-05-02
support onread in net

### LINKED ISSUE #10719: net's onread doesn't work as intended

1.1.6+e58d67b46

Linux 6.8.1-arch1-1-surface x86_64 unknown

Run the following code:

```
const net = require("net");
const opts = {
	host: "localhost",
	port: 143,
	onread: {
		buffer: Buffer.allocUnsafe(4096),
		callback: (size, buf) => {
			console.log(size);
		}
	}
};

net.connect(opts, () => console.log("Connected"));
```

onread's callback should be called when the socket receives data.

### What do you see instead?

onread's callback is never called.

### Additional information

The code works fine if node.js is used to run it. The documentation for this behavior can be found [here](https://nodejs.org/api/net.html#socketconnectoptions-connectlistener).

gvilums:
### What does this PR do?
Implements support for the `onread` option in `node:net`
Fixes #10719

Added a test


src/js/node/net.ts
```diff
@@ -100,6 +100,11 @@ const Socket = (function (InternalSocket) {
       data({ data: self }, buffer) {
         if (!self) return;
 
+        if (self.#onreadCallback) {
```
```suggestion
        const onReadCallback = self.#onreadCallback;
        if (onReadCallback) {
            try {
              onReadCallback(buffer.length, buffer);
            } catch(e) {
              self.emit("error", e);
           }
            return;
        }
```

Jarred-Sumner: Ideally, we'd have a separate handlers object and not do this check here on every single call to onData

gvilums: Changed to use a separate handlers object

uasan:
The behavior is different from Node.js, you create a copy of the buffer and pass this copy as the second parameter, but the whole idea of ​​this interface is to use one buffer, there is no point in this interface if you make a copy of the buffer.

P.S.
I tested on buffer resized, in bun 1.2

PR #10761, opened 2024-05-02
fix(bundler): naming of dynamic import chunks

### LINKED ISSUE #10406: `@aws-sdk/client-cloudfront` causes build error

1.1.4+fbe2fe0c3

Darwin 23.4.0 arm64 arm

The application I'm developing relies on NPM package `@aws-sdk/client-cloudfront`. The build configuration and a simplified version of the build script is as follows:

```ts
import { build } from 'bun';
import type { BuildConfig } from 'bun';

const buildConfiguration = {
  define: {},
  entrypoints: ['./src/index.mts'],
  // external: ['@aws-sdk/*'],
  format: 'esm',
  loader: {},
  minify: true,
  naming: {
    chunk: '[dir]/[name].[hash].mjs',
    entry: '[name].mjs',
  },
  outdir: './dist',
  plugins: [],
  publicPath: '',
  root: '.',
  sourcemap: 'external',
  splitting: true,
  target: 'bun',
} satisfies BuildConfig;

async function main() {
  const buildOutput = await build(buildConfiguration);
  if (!buildOutput.success) {
    process.exitCode = 1;
    for (const error of buildOutput.logs) {
      console.error(error);
    }
  }
}

// Build the Application
await main();
```

**NOTE: the `external` key in the build configuration is commented out**

The application builds correctly, similar to having the `external` key uncommented but with `@aws-sdk/client-cloudfront` being included in the bundle.

### What do you see instead?

The following error occurs when building:
```
error: Multiple files share the same output path: ./index.mjs
```

### Additional information

The application uses these imports from the AWS package:

```ts
const { CloudFrontClient, CreateInvalidationCommand } = await import('@aws-sdk/client-cloudfront');
```

paperclover:
### What does this PR do?

Fixes #10406

fix this error message:

image

make the error go away in the case it did

and then fix some regressions it caused

### DIFF
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -11651,7 +11651,11 @@ const ContentHasher = struct {
 // this is just being nice
 fn cheapPrefixNormalizer(prefix: []const u8, suffix: []const u8) [2]string {
     if (prefix.len == 0) {
-        return .{ "./", bun.strings.removeLeadingDotSlash(suffix) };
+        const suffix_no_slash = bun.strings.removeLeadingDotSlash(suffix);
+        return .{
+            if (strings.hasPrefixComptime(suffix_no_slash, "../")) "" else "./",
+            suffix_no_slash,
+        };
     }
 
     // There are a few cases here we want to handle:
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -4205,7 +4205,6 @@ const LinkerContext = struct {
                     chunk.template.data = this.resolver.opts.entry_naming;
             } else {
                 chunk.template = PathTemplate.chunk;
-                std.debug.print("cnhufna: {s}\n", .{this.resolver.opts.chunk_naming});
                 if (this.resolver.opts.chunk_naming.len > 0)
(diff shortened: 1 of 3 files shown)


src/bundler/bundle_v2.zig
```diff
@@ -4187,40 +4187,44 @@ const LinkerContext = struct {
...
-                    try this.log.addErrorFmt(null, Logger.Loc.Empty, bun.default_allocator, "{s}: failed to open entry point directory: {s}", .{ @errorName(err), pathname.dir });
-                    return error.FailedToOpenEntryPointDirectory;
-                };
-                defer dir.close();
-
-                var real_path_buf: [bun.MAX_PATH_BYTES]u8 = undefined;
-                chunk.template.placeholder.dir = try resolve_path.relativeAlloc(this.allocator, this.resolver.opts.root_dir, try bun.getFdPath(bun.toFD(dir.fd), &real_path_buf));
             } else {
                 chunk.template = PathTemplate.chunk;
+                std.debug.print("cnhufna: {s}\n", .{this.resolver.opts.chunk_naming});
```
```suggestion

```

mangs:
Thank youuuuuuuu

Issue #9302, opened 2024-03-07
Source maps and error locations are broken


paperclover:
The following is an overview over bugs within Bun's source map implementation. This bug is the root cause behind the following:

- bun run
  - Error locations are not pointing at the right character, or line.
    - Error locations can be off by hundreds of lines in the worst case.
  - Bun's VCSode extension (and the work JetBrains is putting to get debugging to work in their products) is unable to break on many lines.
- bun build
  - Browser debuggers are unable to break on many lines
  - Sourcemap analysis tools are unable to open some of Bun's generated source maps.

I've given some efforts to fix these, but I've realized none of this is a quick fix, and will need a much larger audit over every bit of sourcemap code.

## Incorrect Sourcemaps
Sourcemaps are generated incorrectly. A full pass over sourcemap generation in js_parser.zig or bundle_v2.zig should be done to figure out what is going wrong.

Figure A *(there are two spaces after the `-`)*
```ts
;(hello -  world);
```
Figure B
```ts
hello-
world
```

esbuild produces different sourcemaps despite identical JS output.
This is significant because remapping sources does not map the same, some maps are off, some do not map at all.

`demo.sh`
```ts
bun build '/Users/dave/hello/min.js' --target=browser --outdir=bun --sourcemap=external
bunx esbuild '/Users/dave/hello/min.js' --platform=browser --outdir=esbuild --sourcemap=external
bun dump_maps.js
diff --color bun.map esbuild.map
```
`dump_maps.js`
```ts
import { SourceMapConsumer} from 'source-map';
import { readFileSync, writeFileSync } from 'fs';

async function dump(from, to, off = 0) {
    const data = JSON.parse(readFileSync(from, 'utf8'));
    await SourceMapConsumer.with(data, null, async(map) => {
        let out = '';
        map.eachMapping((m) => {
            out += `${m.source} ${m.originalLine}:${m.originalColumn} -> ${m.generatedLine + off}:${m.generatedColumn}\n`;
        });
        await writeFileSync(to, out);
    });
}

await dump('bun/min.js.map', 'bun.map', -1);
await dump('esbuild/min.js.map', 'esbuild.map', );
```

This is more than just adding brace close locations (the initial assumption), but a more fundamental issue in the sourcemaps.

## JSC vs V8 Stack Differences
This seems to only be visible when accessing error.stack. No idea why; More investigation is needed.

[...]

PR #10778, opened 2024-05-02
Fix bad stack traces

zackradisic:
### What does this PR do?

This fixes stack traces pointing to the wrong line on Windows (the cause of the remaining bunshell.test.ts failure)

The problem was caused by C++ variable being declared but not initialized causing undefined memory, and also an off-by-one error because of `.zeroBasedInt()`/`.oneBasedInt()`

- [x] Code changes

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -4326,14 +4326,15 @@ static void fromErrorInstance(ZigException* except, JSC::JSGlobalObject* global,
                 iterator.forEachFrame([&](const V8StackTraceIterator::StackFrame& frame, bool& stop) -> void {
                     ASSERT(except->stack.frames_len < frame_count);
                     auto& current = except->stack.frames_ptr[except->stack.frames_len];
+                    current = {};
 
                     String functionName = frame.functionName.toString();
                     String sourceURL = frame.sourceURL.toString();
                     current.function_name = Bun::toStringRef(functionName);
                     current.source_url = Bun::toStringRef(sourceURL);
-                    current.position.line = frame.lineNumber.zeroBasedInt();
-                    current.position.column_start = frame.columnNumber.zeroBasedInt();
-                    current.position.column_stop = frame.columnNumber.zeroBasedInt();
+                    current.position.line = frame.lineNumber.oneBasedInt();
+                    current.position.column_start = frame.columnNumber.oneBasedInt();
+                    current.position.column_stop = frame.columnNumber.oneBasedInt();
 
                     current.remapped = true;
 
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -4332,9 +4332,9 @@ static void fromErrorInstance(ZigException* except, JSC::JSGlobalObject* global,
                     String sourceURL = frame.sourceURL.toString();
                     current.function_name = Bun::toStringRef(functionName);
                     current.source_url = Bun::toStringRef(sourceURL);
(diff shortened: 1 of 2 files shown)


src/bun.js/bindings/bindings.cpp
```diff
@@ -4326,14 +4326,15 @@ static void fromErrorInstance(ZigException* except, JSC::JSGlobalObject* global,
                 iterator.forEachFrame([&](const V8StackTraceIterator::StackFrame& frame, bool& stop) -> void {
                     ASSERT(except->stack.frames_len < frame_count);
                     auto& current = except->stack.frames_ptr[except->stack.frames_len];
+                    current = {};
```
Jarred-Sumner, line 4329: this one should've already be zero'd

src/bun.js/bindings/bindings.cpp
```diff
@@ -4326,14 +4326,15 @@ static void fromErrorInstance(ZigException* except, JSC::JSGlobalObject* global,
...
 
                     String functionName = frame.functionName.toString();
                     String sourceURL = frame.sourceURL.toString();
                     current.function_name = Bun::toStringRef(functionName);
                     current.source_url = Bun::toStringRef(sourceURL);
-                    current.position.line = frame.lineNumber.zeroBasedInt();
-                    current.position.column_start = frame.columnNumber.zeroBasedInt();
-                    current.position.column_stop = frame.columnNumber.zeroBasedInt();
+                    current.position.line = frame.lineNumber.oneBasedInt();
```
Jarred-Sumner: I don't feel confident these changes are correct

dylan-conway: running a script with `// @bun` and also with a parser error will be off-by-one for line number and column number

paperclover: it would be nice to have some unit tests for this, however there are source-map related bugs that would not make these tests reliable

src/bun.js/bindings/bindings.cpp
paperclover: i cant find a reference for where this is done. and i dont think it hurts. not all fields are being assigned below

paperclover (commented):
i approve but only on the condition that you've verified this makes all stack traces more correct that the status quo. ~~can you show an example where this impacts something?~~ oh, that test. alright, this seems reasonable.

I detail many places we are wrong in https://github.com/oven-sh/bun/issues/9302, which you should look at if you havent. though, that issue is mostly about source maps.

paperclover (approved):

src/bun.js/bindings/bindings.cpp
dylan-conway: We can't use `oneBasedInt()` because we expected them to be zero based when they get to `SourceURLFormatter.format()`: https://github.com/oven-sh/bun/blob/94bf404c417f2b6ecb716679d86ed47e26fb157a/src/bun.js/bindings/exports.zig#L621-L640

dylan-conway (changes requested):
Need to fix https://github.com/oven-sh/bun/pull/10778#discussion_r1588536121

src/bun.js/bindings/bindings.cpp
zackradisic: I think this looks like somewhere we are mixing up one/zero based ints when going to and from JSC, in Bun::formatStackTrace we are converting it into a oneBasedInt: https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/bun.js/bindings/ZigGlobalObject.cpp#L383-L393

PR #10802, opened 2024-05-03
Fix confirm is never true in windows

### LINKED ISSUE #10643: confirm() is never true on Windows

1.1.6+e58d67b46

Ran into same thing I reported previously on Linux... https://github.com/oven-sh/bun/issues/5985

returns true for `y`

### What do you see instead?

is always false, regardless of value

### Additional information

It does work (was fixed) on Linux

rcaselles:
### What does this PR do?
It fixes confirm dialog was always false in bundows

With the new test added

this fixes #10643

### DIFF
--- a/src/bun.js/webcore.zig
+++ b/src/bun.js/webcore.zig
@@ -129,8 +129,6 @@ fn confirm(globalObject: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
             };
             if(next_byte == '\n'){
                 return .false;
-            }else{
-                continue;
             }
         },
         'y', 'Y' => {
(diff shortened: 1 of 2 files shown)


src/bun.js/webcore.zig
```diff
@@ -121,7 +121,7 @@ fn confirm(globalObject: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
     // *  Not relevant in a server context.
 
     switch (first_byte) {
-        '\n' => return .false,
+        '\n', '\r' => return .false,
```
paperclover: when this \r is read, we should ignore it and read the next character.

i think this would cause a bug if you called confirm twice and press return

paperclover (changes requested):
nice, can you double check my note.

src/bun.js/webcore.zig
rcaselles: Changed

rcaselles:
Done! Please check now

paperclover (approved):

PR #10804, opened 2024-05-03
Update documentation for the `[dir]` option when customizing bundler `naming`

mangs:
### What does this PR do?

After working through some issues with Dave and the bundler, I realized the documentation around `naming`'s `[dir]` option was confusing. So I'm attempting to correct that while my thoughts are still fresh.

- [ ] Code changes


paperclover (approved):
i like this better

PR #10840, opened 2024-05-05
Fix passing stdout/stderr from Bun.spawn -> Bun.serve()'s Response

Jarred-Sumner:
### What does this PR do?

This makes it so Bun.spawn().stdout can be used in `new Response(stdout)` within Bun.serve()

```ts
import { spawn, serve } from "bun";

const server = serve({
  port: 0,
  async fetch(req) {
    const { stdout } = spawn({
      cmd: [process.execPath, "--eval", 'console.write("hello world")'],
      env: process.env,
      stdout: "pipe",
      stderr: "inherit",
      stdin: "ignore",
    });

    return new Response(stdout);
  },
});

const response = await fetch(server.url);
console.write(await response.text());
server.stop(true);
```

We assumed `toBlobIfPossible` would always be able to convert pipes into Blob, but that may not always be the case

There is a test

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -190,7 +190,7 @@ $ iex "& {$(irm https://bun.sh/install.ps1)} -Version 1.1.6"
 
 To download Bun binaries directly, you can visit the [releases page](https://github.com/oven-sh/bun/releases) page on GitHub.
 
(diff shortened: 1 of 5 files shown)


cirospaciari (commented):
LGTM

PR #10842, opened 2024-05-05
Speculative fix for napi_set_property crash

Jarred-Sumner:
### What does this PR do?

We have received many crash reports for `napi_set_property`.

The crash implies that `target` is set to undefined (`08`) and we are calling `.getObject()` on it, without properly validating that it is an object

Node.js' implementation of `napi_set_property` checks target is an object via [`CHECK_TO_OBJECT(env, context, obj, object);`](https://github.com/nodejs/node/blob/c82f3c9e80f0eeec4ae5b7aedd1183127abda4ad/src/js_native_api_v8.cc#L1158). 

This PR makes `napi_set_property` check that the property is an object first. 

We also disable strict mode when putting the property. Since Node doesn't default to strict mode, it's a good bet napi_set_property shouldn't either. 

No tests

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -494,24 +494,35 @@ extern "C" napi_status napi_set_property(napi_env env, napi_value target,
         return napi_invalid_arg;
     }
 
-    auto globalObject = toJS(env);
-    auto& vm = globalObject->vm();
-    auto* object = toJS(target).getObject();
-    if (!object) {
+    JSValue targetValue = toJS(target);
+    if (!targetValue.isObject()) {
         return napi_object_expected;
     }
+
+    auto globalObject = toJS(env);
+    auto& vm = globalObject->vm();
+    auto* object = targetValue.getObject();
 
     auto keyProp = toJS(key);
 
     auto scope = DECLARE_CATCH_SCOPE(vm);
-    PutPropertySlot slot(object, true);
-    Identifier identifier = keyProp.toPropertyKey(globalObject);
-    JSValue jsValue = toJS(value);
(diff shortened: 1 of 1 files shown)


src/bun.js/bindings/napi.cpp
```diff
@@ -494,24 +494,35 @@ extern "C" napi_status napi_set_property(napi_env env, napi_value target,
...
+    auto* object = targetValue.getObject();
+
     auto keyProp = toJS(key);
 
     auto scope = DECLARE_CATCH_SCOPE(vm);
-    PutPropertySlot slot(object, true);
+    PutPropertySlot slot(object, false);
+
     Identifier identifier = keyProp.toPropertyKey(globalObject);
+    RETURN_IF_EXCEPTION(scope, napi_generic_failure);
```
dylan-conway, line 512: I think we also need this exception check in `napi_get_property` before calling `getIfPropertyExists`

src/bun.js/bindings/napi.cpp
```diff
@@ -494,24 +494,35 @@ extern "C" napi_status napi_set_property(napi_env env, napi_value target,
...
+
     JSValue jsValue = toJS(value);
 
-    object->put(object, globalObject, identifier, jsValue, slot);
-    RETURN_IF_EXCEPTION(scope, napi_generic_failure);
+    if (!object->put(object, globalObject, identifier, jsValue, slot)) {
```
dylan-conway, line 516: Why was this check added? How is `put` returning false different from an exception thrown?

dylan-conway (commented):
Two comments. Looks good, just confused why we have a check for `false` from `obj->put`

PR #10883, opened 2024-05-07
Fix fd offset handling in ReadStream

### LINKED ISSUE #9907: `fs.createReadStream` and `fs.createWriteStream` from `fd` behaviour differs from Node

1.1.0+5903a6141

Darwin 21.6.0 arm64 arm

```js
const fs = require('fs');
const path = require('path');
const os = require('os');

const tempFile = path.join(os.tmpdir(), 'temp.txt');
fs.open(tempFile, 'w+', (err, fd) => {
  if (err) {
    console.error('Error opening file:', err);
    return;
  }

  const writeStream = fs.createWriteStream('', { fd: fd });
  const readStream = fs.createReadStream('', { fd: fd, start: 0 });

  readStream.on('data', (chunk) => {
    console.log('Got chunk', chunk.toString())
  });

  const data = 'Hello, World!';
  writeStream.write(data, (err) => {
    if (err) {
      console.error('Error writing to file:', err);
      writeStream.close(() => { });
      return;
    }

    console.log('Data written to file.');
  });
});
```

```
~/projects/bun-fd-repro
❯ node test.js
Data written to file.
Got chunk Hello, World!
```

### What do you see instead?

```
~/projects/bun-fd-repro
❯ bun test.js
Data written to file.
```

### Additional information

Originally ran into this bug bumping Bun while testing https://github.com/replit/ruspty

gvilums:
### What does this PR do?
Fixes #9907

Added a test

### DIFF
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -348,6 +348,37 @@ pub const ReadableStream = struct {
         }
     }
 
+    pub fn fromFileBlobWithOffset(
+        globalThis: *JSGlobalObject,
+        blob: *const Blob,
+        offset: usize,
+    ) JSC.JSValue {
+        JSC.markBinding(@src());
+        var store = blob.store orelse {
+            return ReadableStream.empty(globalThis);
+        };
+        switch (store.data) {
+            .file => {
+                var reader = FileReader.Source.new(.{
+                    .globalThis = globalThis,
+                    .context = .{
+                        .event_loop = JSC.EventLoopHandle.init(globalThis.bunVM().eventLoop()),
+                        .start_offset = offset,
+                        .lazy = .{
+                            .blob = store,
+                        },
+                    },
+                });
+                store.ref();
+
+                return reader.toReadableStream(globalThis);
+            },
+            else => {
+                return .zero;
+            },
+        }
+    }
+
     pub fn fromPipe(
         globalThis: *JSGlobalObject,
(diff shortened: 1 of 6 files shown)


src/io/PipeReader.zig
```diff
@@ -948,6 +981,7 @@ pub const WindowsBufferedReader = struct {
 
         is_paused: bool = true,
         has_inflight_read: bool = false,
+        use_pread: bool = false,
```
Jarred-Sumner, line 984: 8 bits -> 9 bits 😢

Jarred-Sumner: idk what to do about that, seems fine

gvilums: because the entire struct is 8-byte aligned this doesn't matter anyway

src/bun.js/webcore/streams.zig
```diff
@@ -348,6 +348,37 @@ pub const ReadableStream = struct {
...
+                            .blob = store,
+                        },
+                    },
+                });
+                store.ref();
+
+                return reader.toReadableStream(globalThis);
+            },
+            else => {
+                return .zero;
```
Jarred-Sumner, line 378: This will cause a crash at runtime if it ever is reached.

```suggestion
                globalThis.throw("Expected FileBlob", .{});
                return .zero;
```

Jarred-Sumner (approved):
LGTM

One comment

PR #10905, opened 2024-05-08
allow user-defined `Content-Length` passed to `fetch` to override default one

nektro:
another case caught by running axios tests

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -50,12 +50,14 @@ const MAX_REDIRECT_URL_LENGTH = 128 * 1024;
 const print_every = 0;
 var print_every_i: usize = 0;
 
+const request_headers_buf_len = 256;
+
 // we always rewrite the entire HTTP request when write() returns EAGAIN
 // so we can reuse this buffer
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
undici-fetch, in Node.js:


```
> fetch("https://www.toptal.com/developers/postbin/1715151383763-3950115528423", { body: "123", headers: { "Content-Length": "2048" }, method: "POST" })
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 148,
  [Symbol(trigger_async_id_symbol)]: 121
}
> Uncaught [TypeError: fetch failed] {
  [cause]: RequestContentLengthMismatchError: Request body length does not match content-length header
      at AsyncWriter.end (node:internal/deps/undici/undici:8088:19)
      at writeIterable (node:internal/deps/undici/undici:7992:16) {
    code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
  }
}
```

I don't think we should make this change in the HTTP client itself. I think if we want to allow users to override Content-Length via node:http, we should have some other mechanism for doing so which does not percolate up to the global fetch API

PR #10906, opened 2024-05-08
update docker support ps commad

albasyir:
### What does this PR do?

Some framework (like nestjs) need PS command to watch changes, so needed to add this to make sure watch still available when user wanna develop it with this docker image

- [ ] Code changes

i have test using nestjs `start:dev` was not working, after I add that, nestjs can watch normally


Electroid:
This guide is intended to be generic for any Bun application. This sounds like a specific dependency needed for a particular one. Because of that, I don't think we'll add this, but appreciate your PR!

PR #10922, opened 2024-05-08
added a few tests for #7716

davidstevens37:
### What does this PR do?

Add a few tests for - `node:http` and #7716 

Tests

### Node.js Equivalent 

```
import { describe, it } from 'node:test';
import assert from 'node:assert';
import { createServer } from 'node:http';

function expect(value) {
  return {
    toBe: function (expected) {
      assert.equal(value, expected);
    }
  }
}

describe('node:http', async () => {
  describe('IncomingRequest', async () => {
    it("emits `close` event when client disconnects", async () => {
      try {
        let close_event_emitted = false;
        var server = createServer((req, res) => {
          res.writeHead(200, { 'Content-Type': 'text/html' });
          const responseInterval = setInterval(() => {
            res.write('hello world');
            clearInterval(responseInterval);
            throw new Error('timeout not cleaned up -- memmory leak')
          }, 100);
          req.on('close', () => {
            close_event_emitted = true;
            clearInterval(responseInterval);
          });
        });
        await new Promise(resolve => {
          server.listen(3000, resolve);
        })
        const controller = new AbortController();
        setTimeout(() => controller.abort(), 50);
        const res = await fetch("http://localhost:3000", { signal: controller.signal }).catch(e => e.name);
        await new Promise(resolve => setTimeout(resolve, 0)); // allow the server a chance to handle the event and set the flag
        expect(res).toBe("AbortError");
        expect(close_event_emitted).toBe(true);
      } catch (e) {
        throw e;
      } finally {
        server.close();
      }
    });
  });
});
```

**Output:**
```
λ node tests.js                                     
▶ node:http
  ▶ IncomingRequest
    ✔ emits `close` event when client disconnects (57.005708ms)
  ▶ IncomingRequest (58.12275ms)

▶ node:http (58.792375ms)

ℹ tests 1
ℹ suites 2
ℹ pass 1
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 62.561542
```

### DIFF
--- a/test/js/node/http/node-http.test.ts
+++ b/test/js/node/http/node-http.test.ts
@@ -847,7 +847,7 @@ describe("node:http", () => {
       }
     });
 
-    it.todo("should emit `close` event on IncomingRequest when client disconnects", async done => {
+    it("should emit `close` event on IncomingRequest when client disconnects", async done => {
       try {
         const response_interval_duration_ms = 100;
         const client_request_abort_duration_ms = 50;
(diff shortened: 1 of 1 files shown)


davidstevens37:
@nektro Thank you again for your support on 7716. Sorry, I'm unable to contribute any more, but the least I could do is add some tests. I hope it's helpful!

gvilums:
Do I understand correctly that these tests currently fail in the latest version of bun? If so, they should be changed from `it(...` to `it.todo(...`, as they're known-failing

davidstevens37:
@gvilums the first test ```should emit `close` event on IncomingRequest when request is destroyed``` is passing. it correctly emits the _close_ event on the `req` when `req.destroy()` is called.

The second test ```should emit `close` event on IncomingRequest when client disconnects``` is failing as it does not emit the _close_ event on the `req` when the client disconnects.

I've updated the second test to `.todo()` thanks!

nektro:
we shouldn't merge failing tests but leaving them as comments on the original issue will be the best place to leave them for whoever takes up the issue next (whether that be me or someone else). appreciate the reductions! :)

davidstevens37:
okay, thanks @nektro! I'll close this PR out and move the tests to original issue.

Issue #9020, opened 2024-02-21
packages with names over 1024 crash the resolver


paperclover:
### What steps can reproduce the bug?

`bun run boo`

[package.json](https://github.com/oven-sh/bun/files/14352427/package.json)

### What is the expected behavior?

to not crash

### What do you see instead?

```
thread 2355986 panic: reached unreachable code
/Users/dave/code/bun/.cache/zig/lib/std/debug.zig:342:14: 0x100329abf in assert (bun-zig)
    if (!ok) unreachable; // assertion failure
             ^
/Users/dave/code/bun/src/bun.zig:364:60: 0x1003844a3 in copy__anon_46322 (bun-zig)
    if (comptime Environment.allow_assert) std.debug.assert(dest.len >= src.len);
                                                           ^
/Users/dave/code/bun/src/resolver/package_json.zig:64:17: 0x100be4faf in generateHash (bun-zig)
        bun.copy(u8, &hashy, package_json.name);
                ^
/Users/dave/code/bun/src/resolver/package_json.zig:1022:42: 0x1007727af in parse__anon_87999 (bun-zig)
                package_json.generateHash();
```
<img width="793" alt="image" src="https://github.com/oven-sh/bun/assets/24465214/9d10b25f-67f8-42d5-8083-a69839dc070e">

### Additional information

issue found by @ryanccn, thanks for this reproduction

RandoomWalks:
@paperdave 
2 thoughts:
allocate a buffer dynamically based on the length of the package name. 
or
 iteratively hash smaller chunks

paperclover:
this should hash smaller chunks if it cannot entirely fit into 1024 bytes.

sequencerr:
@paperdave I'm sorry if the problem is not about matching other's standards, but
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#name
> The name must be less than or equal to 214 characters.

sequencerr:
Maybe display message about invalid name?

PR #10924, opened 2024-05-08
Fix resolver crash and handling name limits

rcaselles:
### What does this PR do?
This PR addresses issue #9020. It attempts to resolve the problem by dividing the data into smaller chunks and then combining them using a XOR operation. However, I am open to suggestions if there is a more optimal method to accomplish this.

Additionally, I have standardized the maximum name length across both the parser and the init command to align with npm's requirements. According to the npm documentation (https://docs.npmjs.com/cli/v10/configuring-npm/package-json#name), the maximum length for a package name is 214 characters. This standardization was necessary but had not been previously included in our implementation (in both package_json.zig and init_command.zig)

To further enhance usability, I've implemented retries for cases where the name length does not meet this limit. This should help ensure that users are able to correct the input without having to restart the process entirely.

I tested with the provided package.json, and different new project configs


Jarred-Sumner:
Thank you for this, but this PR is much complicated for what seems like a very unusual edgecase. 
- The comments feel very much like prompting an LLM
- This adds errors which shouldn't error
- This adds an unnecessary temporary memory allocation
- Uses `page_allocator` which would should not be used outside of tests

PR #10933, opened 2024-05-08
fix(fetch) allow Response to be GC'd before all the request body is received

cirospaciari:
### What does this PR do?

- [x] Code changes

Added tests


src/bun.js/webcore/body.zig
```diff
@@ -308,6 +292,36 @@ pub const Body = struct {
         Error: JSValue,
         Null: void,
 
+        pub fn initRef(value: Value) *BodyValueRef {
```
Jarred-Sumner: Don't we only need this for requests which have bodies instead of allocating one of these for requests which do not have bodies? And we don't need to do this for Response at all right?

test/js/web/fetch/fetch-leak-test-fixture-4.js
```diff
@@ -0,0 +1,36 @@
...
+  for (let i = 0; i < iterations; i++) {
+    {
+      const promises = [];
+      for (let j = 0; j < batch; j++) {
+        promises.push(fetch(server));
+      }
+      await Promise.all(promises);
+    }
+    {
+      await new Promise(r => setTimeout(r, delay));
```
Jarred-Sumner:
```suggestion
      await Bun.sleep(delay);
```

PR #10935, opened 2024-05-08
feat: add canary to crash handler message

paperclover:
### What does this PR do?

make it easier to identify people's screenshots of Bun crashing to determine if it's from a canary build or not.

image

### DIFF
--- a/src/crash_handler.zig
+++ b/src/crash_handler.zig
@@ -616,8 +616,9 @@ else
     "x64";
 
 const metadata_version_line = std.fmt.comptimePrint(
-    "Bun v{s} {s} {s}{s}\n",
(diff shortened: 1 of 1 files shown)


Electroid:
What about `(canary-{sha})` instead? Or just the same string from `bun --revision`?

paperclover:
ah, from that angle it may make sense to format that exactly like --revision prints, aka
```
Bun v1.1.8-canary+41293812 (macOS Silicon)
```

I think while im at it I want to conver this from "silicon" to "aarch64", which is the technical name of this architecture

Jarred-Sumner:
looks nicer imo with `Bun Canary`

PR #10940, opened 2024-05-09
Fixes missing argument in signal events

Jarred-Sumner:
### What does this PR do?

Fixes https://twitter.com/sahithyandev/status/1788243782011367893

`process.on("SIG123", cb)`  is supposed to emit `[signal name, signal number]`. We were only emitting `[ signal number ]`.

> SIGWINCH 28

There is a test

### DIFF
--- a/src/bun.js/bindings/BunProcess.cpp
+++ b/src/bun.js/bindings/BunProcess.cpp
@@ -773,6 +773,7 @@ void signalHandler(uv_signal_t* signal, int signalNumber)
         String signalName = signalNumberToNameMap->get(signalNumber);
         Identifier signalNameIdentifier = Identifier::fromString(globalObject->vm(), signalName);
         MarkedArgumentBuffer args;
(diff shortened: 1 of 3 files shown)


src/bun.js/bindings/BunProcess.cpp
```diff
@@ -773,6 +773,7 @@ void signalHandler(uv_signal_t* signal, int signalNumber)
         String signalName = signalNumberToNameMap->get(signalNumber);
         Identifier signalNameIdentifier = Identifier::fromString(globalObject->vm(), signalName);
         MarkedArgumentBuffer args;
+        args.append(jsString(globalObject->vm(), signalNameIdentifier.string()));
         args.append(jsNumber(signalNumber));
         // TODO(@paperdave): add an ASSERT(isMainThread());
```
paperclover, line 778: aside: this todo comment makes no sense; it is wrapped in "post task concurrently to main thread"

paperclover (approved):

PR #10976, opened 2024-05-10
Fix sloppily written test

Jarred-Sumner:
### What does this PR do?

Fix sloppily written test (cc @nektro):
- It did not set the current working directory, so it installed `"stripe"` into the bun repository instead of the temporary test folder
- It did not check the exit code of `bun install`
- It only happened to work due to auto-install picking a globally-installed version of stripe.
- `String.raw` usage is unnecessary
- It unnecessarily used the node test harness instead of `bun:test`

This also:
- Make running `bun install` in `/package.json` or `test/package.json` from inside a test fail so that we don't run `bun install` in the wrong directory again in a test

Ran the stripe test locally

### DIFF
--- a/test/js/third_party/stripe.test.ts
+++ b/test/js/third_party/stripe.test.ts
@@ -1,10 +1,9 @@
 import { bunExe } from "bun:harness";
-import { bunEnv, tmpdirSync } from "harness";
+import { expect, it } from "bun:test";
+import { bunEnv, isBunCI, tmpdirSync } from "harness";
 import * as path from "node:path";
-import { createTest } from "node-harness";
-const { describe, expect, it, beforeAll, afterAll, createDoneDotAll } = createTest(import.meta.path);
 
-it.skipIf(!process.env.TEST_INFO_STRIPE)("should be able to query a charge", async () => {
+it.skipIf(!isBunCI && !process.env.TEST_INFO_STRIPE)("should be able to query a charge", async () => {
   const package_dir = tmpdirSync("bun-test-");
 
   await Bun.write(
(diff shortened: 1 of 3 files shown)


test/js/third_party/stripe.test.ts
```diff
@@ -1,10 +1,9 @@
 import { bunExe } from "bun:harness";
-import { bunEnv, tmpdirSync } from "harness";
+import { expect, it } from "bun:test";
+import { bunEnv, isBunCI, tmpdirSync } from "harness";
 import * as path from "node:path";
-import { createTest } from "node-harness";
-const { describe, expect, it, beforeAll, afterAll, createDoneDotAll } = createTest(import.meta.path);
 
-it.skipIf(!process.env.TEST_INFO_STRIPE)("should be able to query a charge", async () => {
+it.skipIf(!isBunCI && !process.env.TEST_INFO_STRIPE)("should be able to query a charge", async () => {
```
nektro, line 6: what issues did you have running it locally?

test/js/third_party/stripe.test.ts
```diff
@@ -16,26 +15,23 @@ it.skipIf(!process.env.TEST_INFO_STRIPE)("should be able to query a charge", asy
...
-    stderr: "pipe",
+    stderr: "inherit",
     env: bunEnv,
   });
-  let err = await new Response(stderr).text();
-  expect(err).not.toContain("panic:");
-  expect(err).not.toContain("error:");
-  expect(err).not.toContain("warn:");
-  let out = await new Response(stdout).text();
+  expect(await exited).toBe(0);
```
nektro, line 26: it was done the other way so that you can see what the error was when it failed and nothing when it succeeds

PR #10978, opened 2024-05-10
Avoid loading `libm.so.6` as it breaks Vercel

Jarred-Sumner:
### What does this PR do?

Fixes this error:
```sh
Archive:  bun-linux-x64.zip
--
04:35:54.020 | creating: bun-linux-x64/
04:35:54.782 | inflating: bun-linux-x64/bun
04:35:55.084 | bun: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by bun)

```

Adds a wrapper for `expf` to load it via `dlsym`, which avoids loading `libm.so.6` (un-breaking Vercel deployments)

```js
❯ llvm-objdump -T (which bun-1.1.6) | rg "GLIBC_2.2([0-9]+)"
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.24) quick_exit
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.25) getrandom
❯ llvm-objdump -T (which bun) | rg "GLIBC_2.2([0-9]+)"
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.24) quick_exit
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.25) getrandom
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.27) expf
```

Began in https://github.com/WebKit/WebKit/pull/28160 because `simde` uses `expf`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1141,6 +1141,7 @@ if(UNIX AND NOT APPLE)
         "-Wl,--wrap=stat64"
         "-Wl,--wrap=pow"
         "-Wl,--wrap=exp"
+        "-Wl,--wrap=expf"
         "-Wl,--wrap=log"
         "-Wl,--wrap=log2"
         "-Wl,--wrap=lstat"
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
```
❯ llvm-objdump -T (which bun-10978) | rg "GLIBC_2.2([0-9]+)"
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.24) quick_exit
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.25) getrandom
```

PR #10979, opened 2024-05-10
Set -ffdata-sections -ffunction-sections

Jarred-Sumner:
### What does this PR do?

Set -ffdata-sections -ffunction-sections

See if bun gets smaller


Jarred-Sumner:
This made it bigger

PR #10981, opened 2024-05-10
Tilde expansion

zackradisic:
### What does this PR do?

This adds basic support for [tilde expansion](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_06_01) in Bun shell. 

"Basic" means that tilde is expanded in two scenarios:

```bash
# lone tilde expands to $HOME
echo ~

# tilde + path expands to $HOME/$PATH
echo ~/Documents
```

Also, this changes how JS strings/objects are substituted into the shell script. It now uses a non-printable char as a special token to denote a substituted JS string/object (it was previously using a tilde).

- [x] Code changes

### DIFF
--- a/src/shell/interpreter.zig
+++ b/src/shell/interpreter.zig
@@ -885,18 +885,11 @@ pub const Interpreter = struct {
         }
 
         pub fn getHomedir(self: *ShellState) EnvStr {
-            if (comptime bun.Environment.isWindows) {
-                if (self.export_env.get(EnvStr.initSlice("USERPROFILE"))) |env| {
-                    env.ref();
-                    return env;
-                }
-            } else {
-                if (self.export_env.get(EnvStr.initSlice("HOME"))) |env| {
-                    env.ref();
-                    return env;
-                }
-            }
-            return EnvStr.initSlice("unknown");
+            const env_var: ?EnvStr = brk: {
+                const static_str = if (comptime bun.Environment.isWindows) EnvStr.initSlice("USERPROFILE") else EnvStr.initSlice("HOME");
+                break :brk self.shell_env.get(static_str) orelse self.export_env.get(static_str);
+            };
+            return env_var orelse EnvStr.initSlice("unknown");
         }
 
         pub fn writeFailingErrorFmt(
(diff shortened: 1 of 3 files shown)


src/shell/shell.zig
```diff
@@ -2292,7 +2309,8 @@ pub fn NewLexer(comptime encoding: StringEncoding) type {
                 const escaped = input.escaped;
 
                 // Special token to denote substituted JS variables
-                if (char == '~') {
+                // we use 8 or \b which is a non printable char
+                if (char == 8) {
```
Jarred-Sumner:
```suggestion
                if (char == SPECIAL_JS_CHAR) {
```

src/shell/shell.zig
```diff
@@ -2152,8 +2169,8 @@ pub const LexError = struct {
     /// Allocated with lexer arena
     msg: []const u8,
 };
-pub const LEX_JS_OBJREF_PREFIX = "~__bun_";
-pub const LEX_JS_STRING_PREFIX = "~__bunstr_";
+pub const LEX_JS_OBJREF_PREFIX = &[_]u8{8} ++ "__bun_";
```
Jarred-Sumner:
```suggestion
const SPECIAL_JS_CHAR = 8;
pub const LEX_JS_OBJREF_PREFIX = &[_]u8{SPECIAL_JS_CHAR} ++ "__bun_";
```

src/shell/shell.zig
```diff
@@ -2152,8 +2169,8 @@ pub const LexError = struct {
...
+pub const LEX_JS_STRING_PREFIX = &[_]u8{8} ++ "__bunstr_";
```
Jarred-Sumner:
```suggestion
pub const LEX_JS_STRING_PREFIX = &[_]u8{SPECIAL_JS_CHAR} ++ "__bunstr_";
```

Jarred-Sumner:
instead of `8` we could use a named constant

otherwise LGTM

src/shell/interpreter.zig
```diff
@@ -885,18 +885,11 @@ pub const Interpreter = struct {
...
-                    env.ref();
-                    return env;
-                }
-            }
-            return EnvStr.initSlice("unknown");
+            const env_var: ?EnvStr = brk: {
+                const static_str = if (comptime bun.Environment.isWindows) EnvStr.initSlice("USERPROFILE") else EnvStr.initSlice("HOME");
+                break :brk self.shell_env.get(static_str) orelse self.export_env.get(static_str);
+            };
+            return env_var orelse EnvStr.initSlice("unknown");
```
gvilums: does it make sense to return the string `unknown`? If we expand `~` to `unknown` it's just going to lead to really confusing errors

gvilums (commented):
Looks good except for one detail

Jarred-Sumner:
`root` is usually `/root` on Linux

Do we need to do anything for that to work?

PR #10993, opened 2024-05-10
emit newListener event before new listener is added

gvilums:
### What does this PR do?
Emits the `newListener` event before the new listener is added to the EventEmitter. This was already correct in our `EventEmitter` implementation in `events.ts`, but wrong in `JSEventEmitter.cpp`, upon which our implementation of `node:stream` relies.

This fixes the current issue in #3150. However, the `got` library still doesn't work after this change (although it gets much further along in making a request).

A new test


Jarred-Sumner:
@nektro `st` test is flaky

test/js/node/stream/node-stream.test.js
```diff
@@ -588,3 +588,11 @@ it("should send Readable events in the right order", async () => {
     ``,
   ]);
 });
+
+it("emits newListener event _before_ adding the listener", () => {
+  const stream = new Stream();
```
Jarred-Sumner: This test would not fail if `newListener` is never called

We need to assert that the event handler has been called

```suggestion
  const cb = jest.fn((event) => {
        expect(stream.listenerCount(event)).toBe(0);
  });
  const stream = new Stream();
  stream.on("newListener", cb);
  stream.on("foo", () => {});
  expect(cb).toHaveBeenCalled();
```

Jarred-Sumner (commented):
1 comment on the test

Jarred-Sumner (approved):

PR #10994, opened 2024-05-10
Snapshot lockfiles in some tests

dylan-conway:
### What does this PR do?
Also changes the format buffer for semver groups. `input` is a slice of `string_bytes`

- [ ] Code changes

CI


test/cli/install/__snapshots__/bun-workspaces.test.ts.snap
```diff
@@ -1,1337 +1,2123 @@
...
+      "package_id": 1,
+      "workspace": "packages/mono",
+    },
+    {
+      "behavior": {
+        "normal": true,
+        "workspace": true,
+      },
+      "id": 2,
+      "literal": "=*",
```
Jarred-Sumner, line 280: `=*`

test/cli/install/__snapshots__/bun-workspaces.test.ts.snap
```diff
@@ -1,1337 +1,2123 @@
...
+        "version": ">=0.0.0",
+      },
+      "package_id": 1,
+    },
+  ],
   "format": "v2",
   "meta_hash": "1e2d5fa6591f007aa6674495d1022868fc3b60325c4a1555315ca0e16ef31c4e",
   "package_index": {
-    "lodash": 1,
+    "bar": 2,
```
Jarred-Sumner, line 540: should this order have flipped everywhere?

dylan-conway: Taking a snapshot of the object vs string. I think I will switch back to returning a string from zig, then parse it as json in js

Jarred-Sumner: i kinda like it tbh

but also i think it would catch determinism issues potentially

PR #10997, opened 2024-05-11
Make `EINTR` less common

Jarred-Sumner:
### What does this PR do?

This makes `EINTR` less likely to occur on macOS by ensuring we only use the `$NOCANCEL` versions of symbols where possible. This also should very slightly improve start performance since there are ~20 fewer symbols to load

CI

### DIFF
--- a/.github/workflows/build-darwin.yml
+++ b/.github/workflows/build-darwin.yml
@@ -9,7 +9,7 @@ on:
     inputs:
       runs-on:
         type: string
-        default: macos-12-large
+        default: macos-13-large
       tag:
         type: string
         required: true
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -1121,6 +1121,10 @@ if(APPLE)
     target_link_options(${bun} PUBLIC "-dead_strip_dylibs")
     target_link_options(${bun} PUBLIC "-Wl,-stack_size,0x1200000")
     target_link_options(${bun} PUBLIC "-exported_symbols_list" "${BUN_SRC}/symbols.txt")
+
+    # Must be anbled after macOS 14
```
staabm:
```suggestion
    # Must be enabled after macOS 14
```

.github/workflows/build-darwin.yml
```diff
@@ -9,7 +9,7 @@ on:
     inputs:
       runs-on:
         type: string
-        default: macos-12-large
+        default: macos-13-large
```
paperclover, line 12: are we dropping MacOS 12 support?

PR #10998, opened 2024-05-11
feat: load sourcemaps at runtime when using a `bun build --target=bun` bundle

paperclover:
This makes it so when using sourcemaps with `bun build --target=bun`, those maps actually get loaded and used for errors:
image

The main api change to our sourcemap functions is you can now specify an intent for resolving mappings, one that includes source code information, and one that does not.

```diff
 if (this.source_mappings.resolveMapping(
     source_url.slice(),
     @max(frame.position.line, 0),
     @max(frame.position.column_start, 0),
+    .no_source_contents,
 )) |lookup| {
```

This was not needed before because we assumed the source is always the untranspiled code. But with --target=bun and it's `// @bun` annotation, the untranspiled code *IS* the transpiled code, and the source is stored in the sourcemap.

We are not ready yet:

- [x] Why is filename still not remapped.
- [x] Verify `error.stack` works
- [x] Support loading external sourcemaps by checking for a `.map` file
- [x] Resolve TODO in `ZigSourceProvider__getSourceSlice`
- [x] Resolve TODO `getSourceCode()` returns null
- [x] Resolve TODO in `getSourceCode`, this is when you load a sourcemap without file contents but then later need them.
- [x] Resolve TODOs in `toParsedSourceMap`.


src/sourcemap/sourcemap.zig
```diff
@@ -39,85 +39,312 @@ sources_content: []string,
...
+/// `source` must be in UTF-8 and can be freed after this call.
+/// The mappings are owned by the `alloc` allocator.
+/// Temporary allocations are made to the `arena` allocator, which
+/// should be an arena allocator (caller is assumed to call `deinit`).
+pub fn parseJSON(
+    alloc: std.mem.Allocator,
+    arena: std.mem.Allocator,
+    source: []const u8,
+    result: ParseUrlResultType,
+) Mapping.ParseResult {
```
paperclover: please rewrite this to use a zig error type so `errdefer` can be used

src/bun.js/javascript.zig
```diff
@@ -3825,4 +3890,12 @@ pub fn NewHotReloader(comptime Ctx: type, comptime EventLoopType: type, comptime
     };
 }
 
+export fn Bun__addSourceProviderSourceMap(bun_vm: *anyopaque, opaque_source_provider: *anyopaque, specifier: *bun.String) void {
```
paperclover: *anyopaque -> VirtualMachine

Jarred-Sumner:
Needs a test that does hot reloading with source maps enabled and makes sure it doesn't crash due to ref/unref of SourceProvider

Jarred-Sumner:
Also needs a test for Bun.inspect on `// @bun`'d file

src/bun.js/bindings/BunString.cpp
```diff
@@ -196,6 +196,22 @@ BunString toStringRef(WTF::StringImpl* wtfString)
...
+        BunStringTag::StaticZigString,
+        { .zig = {
+            .ptr = (const LChar*)(view.is8Bit()
+                ? (size_t)view.rawCharacters()
+                : ((size_t)view.rawCharacters() + (static_cast<uint64_t>(1) << 63))
+            ),
+            .len = view.length()
+        } }
+    };
+}
```
paperclover: i very much want someone's opinion on this. it is hard to tell if StringView.toString is a copy or a reference. this function just takes character reference.

src/bun.js/bindings/BunString.cpp
```diff
@@ -196,6 +196,22 @@ BunString toStringRef(WTF::StringImpl* wtfString)
     return { BunStringTag::WTFStringImpl, { .wtf = wtfString } };
 }
 
+BunString toStringView(StringView view) {
+    view.is8Bit();
```
Jarred-Sumner:
```suggestion

```

src/bun.js/bindings/BunString.cpp
Jarred-Sumner: If `StringView` does not contain a `WTF::StringImpl` inside, then it should be `ZigString` so that we do not attempt to create a string which is never cloned/freed

src/bun.js/bindings/bindings.zig
```diff
@@ -421,6 +421,16 @@ pub const ZigString = extern struct {
             };
         }
 
+        /// Creates a `Slice` from bytes without an allocator, aka a string view.
+        /// This slice is only valid for the lifetime of the given input, and
+        /// will never be freed by this slice.
+        pub fn initStatic(input: []const u8) Slice {
```
Jarred-Sumner: `fromUTF8NeverFree`?

Jarred-Sumner: Otherwise, if this a latin1/utf16 string, it's invalid memory. ZigString.Slice is supposed to always be UTF8.

src/bun.js/javascript.zig
```diff
@@ -211,29 +232,34 @@ pub const SavedSourceMap = struct {
...
+            if (old_value.get(ParsedSourceMap)) |parsed_source_map| {
+                var source_map: *ParsedSourceMap = parsed_source_map;
                 source_map.deinit(default_allocator);
-            } else if (value.get(SavedMappings)) |saved_mappings| {
+            } else if (old_value.get(SavedMappings)) |saved_mappings| {
                 var saved = SavedMappings{ .data = @as([*]u8, @ptrCast(saved_mappings)) };
-
                 saved.deinit();
+            } else if (old_value.get(SourceProviderMap)) |provider| {
+                _ = provider; // do nothing
```
Jarred-Sumner: This will need a test

src/bun.js/bindings/bindings.zig
paperclover: thanks

src/bun.js/bindings/BunString.cpp
paperclover: i chose `StaticZigString` as the tag to specifically avoid any possibility of freeing the data. not sure if this is copied. going on the single usage, i dont think it is.

A StringView is a string view. It doesn't have a destructor or a ref to the underlying string in release builds.

image

src/bun.js/javascript.zig
```diff
@@ -138,50 +143,87 @@ pub const SavedSourceMap = struct {
...
+        }
+    };
+
+    pub fn putZigSourceProvider(this: *SavedSourceMap, opaque_source_provider: *anyopaque, path: []const u8) void {
+        const source_provider: *SourceProviderMap = @ptrCast(opaque_source_provider);
+        this.putValue(path, Value.init(source_provider)) catch bun.outOfMemory();
+    }
+
+    pub fn removeZigSourceProvider(this: *SavedSourceMap, opaque_source_provider: *anyopaque, path: []const u8) void {
+        const entry = this.map.getEntry(bun.hash(path)) orelse return;
```
Jarred-Sumner, line 214: where is the mutex?

src/bun.js/javascript.zig
```diff
@@ -138,50 +143,87 @@ pub const SavedSourceMap = struct {
...
+                    .{note},
+                );
+                Output.note("consider bundling with '--sourcemap' to get an unminified traces", .{});
+            }
+        }
+    };
+
+    pub fn putZigSourceProvider(this: *SavedSourceMap, opaque_source_provider: *anyopaque, path: []const u8) void {
+        const source_provider: *SourceProviderMap = @ptrCast(opaque_source_provider);
+        this.putValue(path, Value.init(source_provider)) catch bun.outOfMemory();
```
Jarred-Sumner, line 207: where is the mutex?

src/string.zig
```diff
@@ -266,10 +266,22 @@ pub const StringImplAllocator = struct {
 };
 
 pub const Tag = enum(u8) {
+    /// String is not valid. Observed on some failed operations.
+    /// To prevent crashes, this value acts similarly to .Empty (such as length = 0)
     Dead = 0,
+    /// String is backed by a WTF::StringImpl from JavaScriptCore.
+    /// Can be in either `latin1` or `utf16le` encodings.
     WTFStringImpl = 1,
+    /// Memory is backed by an allocation in Bun's Zig codebase. When converted to JSValue,
```
Jarred-Sumner:
```suggestion
    /// Memory has an unknown owner, likely in Bun's Zig codebase. If `isGloballyAllocated` is set, then it is owned by mimalloc. When converted to JSValue,
```

paperclover:
Followup TODO:

- [x] Prevent source contents from staying in memory forever

Jarred-Sumner:
@paperdave the windows tests are timing out?

PR #11017, opened 2024-05-12
fix(install.sh): support windows mingw platform

ahaoboy:
### What does this PR do?

Added support for installing bun using bash under windows mingw platform
 
It runs correctly locally


paperclover (approved):

paperclover:
note that the website sync isn't instant and might take some time.

PR #11027, opened 2024-05-13
[bun:test] Implement `beforeEachFile` and `afterEachFile` hooks

Jarred-Sumner:
### What does this PR do?

This adds support for adding a function to call before each test file runs and after each test file runs:

In a preload script:
```js
import {beforeEachFile, afterEachFile} from 'bun:test';

beforeEachFile(() => {
  resetDatabase();
});
```

In a test:
```ts
import {test, expect} from 'bun:test';

test("signup user in database", async () => {
  doSomething();
});
```

When switching to`bun:test` from vitest or jest, there currently is no equivalent to [`setupFiles`](https://vitest.dev/config/#setupfiles). This adds an alternative .

WIP. Needs automated tests.

### DIFF
--- a/test/js/bun/test/test-test.test.ts
+++ b/test/js/bun/test/test-test.test.ts
@@ -3,7 +3,7 @@ import { spawn, spawnSync } from "bun";
 import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it, test } from "bun:test";
 import { mkdirSync, realpathSync, rmSync, writeFileSync, copyFileSync } from "fs";
 import { mkdtemp, rm, writeFile } from "fs/promises";
-import { bunEnv, bunExe } from "harness";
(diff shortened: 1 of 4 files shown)


test/js/bun/test/test-test.test.ts
```diff
@@ -654,3 +654,104 @@ describe("empty", () => {
...
+
+  const lines = stdout
+    .toString()
+    .trim()
+    .split("\n")
+    .map(a => a.trim());
+
+  expect(lines).toStrictEqual([
+    "beforeEachFile",
+    "beforeAll",
```
gvilums, line 738: Is it intended that `beforeEachFile` runs before `beforeAll`?

morgante:
This is still needed to be able to port many test suites.

PR #11050, opened 2024-05-13
fix windows next tests in ci

dylan-conway:
### What does this PR do?
testing using `GetLongPathName` in our resolver

- [ ] Code changes


src/resolver/resolver.zig
```diff
@@ -1160,16 +1160,19 @@ pub const Resolver = struct {
         // Treating these paths as absolute paths on all platforms means Windows
         // users will not be able to accidentally make use of these paths.
         if (std.fs.path.isAbsolute(import_path)) {
+            var normalizer = ResolvePath.PosixToWinNormalizer{};
+            const abs_import_path = normalizer.resolve(source_dir, import_path);
+
             if (r.debug_logs) |*debug| {
-                debug.addNoteFmt("The import \"{s}\" is being treated as an absolute path", .{import_path});
+                debug.addNoteFmt("The import \"{s}\" is being treated as an absolute path", .{abs_import_path});
```
paperclover, line 1167: this should print the user-provided path. though, we dont really show this anywhere. (i could never get the debug messages to work)

src/resolver/resolve_path.zig
```diff
@@ -1910,15 +1920,47 @@ pub const PosixToWinNormalizer = struct {
...
         @memcpy(buf.ptr, maybe_posix_path);
         buf[maybe_posix_path.len] = 0;
         return buf[0..maybe_posix_path.len :0];
     }
+
+    /// `res` needs to exist in `buf` and needs to start at the beginning
+    pub fn ensureLongPath(buf: anytype, res: [:0]const u8) [:0]u8 {
+        assert(buf.ptr == res.ptr);
+        if (bun.Environment.isWindows) {
+            const long_path_len = bun.windows.GetLongPathNameA(
```
Jarred-Sumner, line 1941: Let's only do this if the file path contains `~{0-9}`

Jarred-Sumner: And we should never use the `A` version of functions

PR #11057, opened 2024-05-14
test: unify how all files create a temp directory

nektro:
this makes clearing the bun folders in the system temp directory much easier as they all use the same naming scheme with this patch. as well as ensures all created folders are unique, rather than leaving it up to an impromptu randomness algorithm

### DIFF
--- a/test/harness.ts
+++ b/test/harness.ts
@@ -761,5 +761,5 @@ export function mergeWindowEnvs(envs: Record<string, string | undefined>[]) {
 }
 
 export function tmpdirSync(pattern: string = "bun.test.") {
-  return fs.mkdtempSync(join(fs.realpathSync(os.tmpdir()), "bun.test."));
+  return fs.mkdtempSync(join(fs.realpathSync(os.tmpdir()), pattern));
 }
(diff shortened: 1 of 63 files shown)


test/harness.ts
```diff
@@ -760,6 +760,6 @@ export function mergeWindowEnvs(envs: Record<string, string | undefined>[]) {
   return flat;
 }
 
-export function tmpdirSync(pattern: string) {
-  return fs.mkdtempSync(join(fs.realpathSync(os.tmpdir()), pattern));
+export function tmpdirSync() {
```
Jarred-Sumner:
```suggestion
export function tmpdirSync(pattern: string = "bun.test.") {
```

test/harness.ts
```diff
@@ -760,6 +760,6 @@ export function mergeWindowEnvs(envs: Record<string, string | undefined>[]) {
...
+  return fs.mkdtempSync(join(fs.realpathSync(os.tmpdir()), "bun.test."));
```
Jarred-Sumner:
```suggestion
  return fs.mkdtempSync(join(fs.realpathSync(os.tmpdir()), pattern));
```

Jarred-Sumner:
test failures

Jarred-Sumner:
Thank you

Issue #2297, opened 2023-03-04
Support `msw`


xxleyi:
### What is the problem this feature would solve?

use `msw` in bun has some problem, after some exploration, find `node:http` is not fully implemented in Bun.

with this code
```
import axios from 'axios';

import { setupServer } from 'msw/node'

const server = setupServer(...[])
server.listen({
  onUnhandledRequest: 'warn',
});

axios.get('https://swapi.dev/api/people/?page=2')
  .then(function (response) {
    // handle success
    console.log(response.data.results.length);
  })
  .catch(function (error) {
    // handle error
    console.log(error?.message);
  });
```
and dependencies:
```
{
  "dependencies": {
    "axios": "^1.4.0",
    "msw": "^1.2.3"
  }
}
```

output in Node:
```
[MSW] Warning: captured a request without a matching request handler:

  • GET https://swapi.dev:8001/api/people/

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
Request failed with status code 400
```
 
output in Bun
```
Cannot call a class constructor without |new|
```

and first two of error stack is
```
TypeError: Cannot call a class constructor without |new|
      at ClientRequest (node:http:884:26)
```

### What is the feature you are proposing to solve the problem?

fully implement `node:http` to support msw

ThatOneBro:
Thank you for reporting this.

Our `node:http` module is basically complete but it looks like some exports are missing (but implemented) and aside from that we have been getting a few bug reports with it lately. 

I'm currently working on porting all of the `node:http` tests from Node and working through the bugs now and hopefully we can get full compatibility with `node:http` and `node:https` (or close to it) within the next few days or so.

xxleyi:
> Thank you for reporting this.
> 
> Our `node:http` module is basically complete but it looks like some exports are missing (but implemented) and aside from that we have been getting a few bug reports with it lately.
> 
> I'm currently working on porting all of the `node:http` tests from Node and working through the bugs now and hopefully we can get full compatibility with `node:http` and `node:https` (or close to it) within the next few days or so.

Sorry to bother, do we have trouble in porting node:http?

xxleyi:
@Jarred-Sumner  sorry to bother, can you have a look at this problem, which broke `msw` in bun:test

birkskyum:
Current output in bun:

```zig
➜ bun index.ts
🚀 ~ http: {
  Agent: [class Agent],
[...]

PR #11058, opened 2024-05-14
node:http: allow setting response.statusCode and statusMessage

nektro:
Closes https://github.com/oven-sh/bun/issues/2297


Jarred-Sumner:
`msw` test is flaky

Jarred-Sumner:
conflicts

PR #2833, opened 2023-05-09
bundler tests and improve `Bun.build` return type

paperclover:

PR #11059, opened 2024-05-14
fix(test): use a real png image instead of Symbolic Link

ahaoboy:
### What does this PR do?

Added a png image in [#2833](https://github.com/oven-sh/bun/pull/2833) with path `/data/archive/capture/2023/05/2023-05-10_13-18-34i.png`, I'm not sure if this is intentional or not, it's problematic on Win platforms, perhaps a relative path that exists within the project should be used, I replaced it with a very small black png


paperclover (approved):
`/data/archive/capture/2023/05/2023-05-10_13-18-34i.png` is a path on my linux computer from my screenshot tool. it should not have been a symlink committed. this change is good

Issue #10725, opened 2024-05-01
BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD doesn't work


mnpenner:
### What version of Bun is running?

1.1.6+e58d67b46

### What platform is your computer?

Linux 4.4.0-19041-Microsoft x86_64 x86_64

### What steps can reproduce the bug?

Run

```sh
BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD=1 bun run --watch server/index.ts
```

Then edit some code.

### What is the expected behavior?

Bun does not clear the terminal on reload.

### What do you see instead?

Bun clears the terminal on reload.

### Additional information

https://bun.sh/docs/runtime/env says it should work.

mnpenner:
I just figured it out by looking at the source code.

https://github.com/oven-sh/bun/blob/4970ffc3a4e69827c0a411a1be20634af6ae7f74/src/bun.js/javascript.zig#L868

The docs say `1` should work though:

![image](https://github.com/oven-sh/bun/assets/1112998/27991d82-6ccc-44fc-afab-41d3416a1c0c)

I also just found https://github.com/oven-sh/bun/pull/9647 to fix the docs.

PR #11078, opened 2024-05-14
Fix `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD` env var

Jarred-Sumner:
### What does this PR do?

Fix `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD` env var so that it handles more ways to say false. Any value which is not a falsy value makes it set to true, if it is set

For example:
- `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD=123 bun --watch src/server.ts` will disable clearing the screen
- `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD=0 bun --watch src/server.ts` will enable clearing the screen
- `BUN_CONFIG_NO_CLEAR_TERMINAL_ON_RELOAD= bun --watch src/server.ts` will enable clearing the screen

Fixes https://github.com/oven-sh/bun/issues/10725

Did not test it.

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -890,7 +890,7 @@ pub const VirtualMachine = struct {
 
     pub fn reload(this: *VirtualMachine) void {
         Output.debug("Reloading...", .{});
-        const should_clear_terminal = !this.bundler.env.hasSetNoClearTerminalOnReload() and Output.enable_ansi_colors;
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
manually tested

image

paperclover (approved):

PR #11092, opened 2024-05-15
shell: Fix some `cd` returning the wrong exit code

### LINKED ISSUE #11084: Shell execution continues after failed command

1.1.8+89d25807f

Darwin 23.3.0 arm64 arm

in Bun shell, I believe there's an issue with some commands, where even if they fail, they allow execution of subsequent instructions. Example:
```
import {$} from "bun";
try {
    await $`cd non-existent-directory && echo " This should not be printed?"`
}catch (e) {
    console.log("error");
}
```
the output is: `cd: not a directory: non-existent-directory This should not be printed?`

However this does not happen with all commands, for example
```
import {$} from "bun";
try {
    await $`cat non-existent-directory && echo " This should not be printed?"`
}catch (e) {
    console.log("error");
}
```
prints:
```
cat: non-existent-directory: No such file or directory
error
```

Finally, if the instructions are on 2 lines, then the previous return code is always ignored (which is also the case in bash):
```
import {$} from "bun";
try {
    await $`cat non-existent-directory 
    echo " This should not be printed?"`
}catch (e) {
    console.log("error");
}
```
but it would be nice to have a behavior similar to `set -euo pipefail` (maybe by default or at least opt-in).

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

zackradisic:
### What does this PR do?

Fixes #11084

`cd` was not returning the correct exit code when it failed with an error

This also adds tests for other builtins to make sure they fail with non-zero exit code


test/js/bun/shell/bunshell.test.ts
```diff
@@ -37,6 +37,40 @@ afterAll(async () => {
...
+      // "seq",
+      "dirname",
+      "basename",
+      "cp ksdjflksjdfks lkjsdflksjdfl",
+    ];
+
+    failing_cmds.forEach(cmdstr =>
+      !!cmdstr
+        ? TestBuilder.command`${{ raw: cmdstr }}`
+            .exitCode(c => c !== 0)
```
Jarred-Sumner, line 66: Can we test for a specific exit code here, to avoid unintentionally changing the exit code in the future?
```suggestion
            .exitCode(c => c === 1)
```

PR #11097, opened 2024-05-15
Asynchronous DNS for sockets

gvilums:
### What does this PR do?
Changes usockets to resolve DNS queries asynchronously through a global resolver, instead of blocking.

Existing tests.


packages/bun-usockets/src/loop.c
```diff
@@ -165,25 +169,42 @@ void us_internal_handle_low_priority_sockets(struct us_loop_t *loop) {
     }
 }
 
+void us_internal_handle_dns_results(struct us_loop_t *loop) {
+    pthread_mutex_lock(&loop->data.mutex);
+    struct us_connecting_socket_t *s = loop->data.dns_ready_head;
```
Jarred-Sumner: the mutex shouldn't be necessary, we should be able to do an atomic swap

packages/bun-usockets/src/internal/networking/bsd.h
```diff
@@ -40,6 +40,7 @@
 #endif
 /* For socklen_t */
 #include <sys/socket.h>
+#include <netdb.h>
```
Jarred-Sumner, line 43: was this include intentional

packages/bun-usockets/src/internal/loop_data.h
```diff
@@ -18,6 +18,9 @@
 #ifndef LOOP_DATA_H
 #define LOOP_DATA_H
 
+#include <pthread.h>
```
Jarred-Sumner: if we use atomics we can avoid the mutex and this include (which is probably causing the windows build failure)

packages/bun-usockets/src/internal/internal.h
```diff
@@ -122,6 +126,20 @@ struct us_socket_t {
...
+    alignas(LIBUS_EXT_ALIGNMENT) void *addrinfo_req;
+    struct us_socket_context_t *context;
+    struct us_connecting_socket_t *next;
+    struct us_socket_t *socket;
+    int options;
+    int socket_ext_size;
+    unsigned int closed : 1, shutdown : 1, ssl : 1, shutdown_read : 1;
+    unsigned char timeout;
+    unsigned char long_timeout;
+    int error;
```
Jarred-Sumner, line 155: does the error need to be stored?

packages/bun-usockets/src/context.c
```diff
@@ -343,33 +345,88 @@ struct us_listen_socket_t *us_socket_context_listen_unix(int ssl, struct us_sock
...
 #ifndef LIBUS_NO_SSL
     if (ssl) {
-        return (struct us_socket_t *) us_internal_ssl_socket_context_connect((struct us_internal_ssl_socket_context_t *) context, host, port, source_host, options, socket_ext_size);
+        return us_internal_ssl_socket_context_connect((struct us_internal_ssl_socket_context_t *) context, host, port, options, socket_ext_size);
     }
 #endif
 
-    LIBUS_SOCKET_DESCRIPTOR connect_socket_fd = bsd_create_connect_socket(host, port, source_host, options);
+    struct us_connecting_socket_t *c = malloc(sizeof(struct us_connecting_socket_t) + socket_ext_size);
+    memset(c, 0, sizeof(struct us_connecting_socket_t));
```
Jarred-Sumner: instead of malloc and then memset 0, we can use calloc

packages/bun-usockets/src/context.c
```diff
@@ -343,33 +345,88 @@ struct us_listen_socket_t *us_socket_context_listen_unix(int ssl, struct us_sock
...
+struct us_connecting_socket_t *us_socket_context_connect(int ssl, struct us_socket_context_t *context, const char *host, int port, int options, int socket_ext_size) {
 #ifndef LIBUS_NO_SSL
     if (ssl) {
-        return (struct us_socket_t *) us_internal_ssl_socket_context_connect((struct us_internal_ssl_socket_context_t *) context, host, port, source_host, options, socket_ext_size);
+        return us_internal_ssl_socket_context_connect((struct us_internal_ssl_socket_context_t *) context, host, port, options, socket_ext_size);
     }
 #endif
 
-    LIBUS_SOCKET_DESCRIPTOR connect_socket_fd = bsd_create_connect_socket(host, port, source_host, options);
+    struct us_connecting_socket_t *c = malloc(sizeof(struct us_connecting_socket_t) + socket_ext_size);
```
Jarred-Sumner:
```suggestion
    struct us_connecting_socket_t *c = us_calloc(1, sizeof(struct us_connecting_socket_t) + socket_ext_size);
```

packages/bun-usockets/src/context.c
```diff
@@ -19,6 +19,8 @@
 #include "internal/internal.h"
 #include <stdlib.h>
 #include <string.h>
+#include <pthread.h>
+#include <stdio.h>
```
Jarred-Sumner: did you mean to include stdio.h?

packages/bun-usockets/src/bsd.c
```diff
@@ -905,7 +905,7 @@ static int bsd_do_connect(struct addrinfo *rp, int *fd)
             return LIBUS_SOCKET_ERROR;
         }
 
-        int resultFd = bsd_create_socket(rp->ai_family, rp->ai_socktype, rp->ai_protocol);
+        int resultFd = bsd_create_socket(rp->ai_family, SOCK_STREAM, 0);
```
Jarred-Sumner: why hardcode it?

gvilums: these are the only valid values for a TCP socket, and makes `bsd_do_connect` independent of the values returned by `getaddrinfo` beyond the resolved addresses

packages/bun-usockets/src/internal/internal.h
gvilums: it's necessary so `us_connecting_socket_get_error` can return the right error if `getaddrinfo` fails, as `getaddrinfo` doesn't set `errno`

src/http.zig
```diff
@@ -742,13 +742,14 @@ pub const HTTPThread = struct {
...
-        const loop = bun.uws.Loop.create(struct {
-            pub fn wakeup(_: *uws.Loop) callconv(.C) void {
-                http_thread.drainEvents();
-            }
-            pub fn pre(_: *uws.Loop) callconv(.C) void {}
-            pub fn post(_: *uws.Loop) callconv(.C) void {}
-        });
+        // const loop = bun.uws.Loop.create(struct {
+        //     pub fn wakeup(_: *uws.Loop) callconv(.C) void {
+        //         http_thread.drainEvents();
```
Jarred-Sumner: Where is `drainEvents` called?

Jarred-Sumner:
Looks like WebSocket (websocket_http_client) needs to be updated

src/http.zig
gvilums: Before every iteration of the event loop (`processEvents`)

Issue #11014, opened 2024-05-11
Segmentation fault at address 0x0


e3dio:
### How can we reproduce the crash?

serving http requests, I am looking for minimal reproducer

### Relevant log output

```shell
Bun v1.1.8 (89d25807) Linux x64
Features: jsc fetch(2) http_server transpiler_cache(2) tsconfig 
Builtins: "bun:main" "node:assert" "node:buffer" "node:crypto" "node:dns" "node:events" "node:fs" "node:fs/promises" "node:http" "node:net" "node:os" "node:querystring" "node:stream" "node:string_decoder" "node:timers" "node:tls" "node:url" "node:util" "node:util/types" "node:zlib"
```

### Stack Trace (bun.report)

Bun v1.1.8 ([`89d2580`](https://github.com/oven-sh/bun/tree/89d25807fbf4488d3bfcb116484d0c2cf824f98c)) on linux x86_64 [AutoCommand]

Segmentation fault at address 0x00000000

- [`ld-temp.o:0`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/ld-temp.o#L0): `char32_t WTF::Unicode::next<(...)0, char8_t>`
- `??`
- [`StringImpl.h:262`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c//build/bun/bun-webkit/include/wtf/text/StringImpl.h#L262): `Bun::ProcessBindingUV::jsErrname`
- [`bindings.zig:3733`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/bun.js/bindings/bindings.zig#L3733): `src.bun.js.event_loop.EventLoop.runCallback`
- [`socket.zig:215`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/bun.js/api/bun/socket.zig#L215): `src.bun.js.api.bun.socket.NewSocket`
- [`javascript.zig:845`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/bun.js/javascript.zig#L845): `src.bun.js.api.bun.socket.Handlers.callErrorHandler`
[...]

PR #11103, opened 2024-05-16
Fix bug with errname not allowing double numbers as ints

Jarred-Sumner:
### What does this PR do?

`errname` is in the stack trace here: https://github.com/oven-sh/bun/issues/11014

I suspect the stacktrace is wrong due to some bug in our stack trace code (cc @paperdave) but noticed a couple things in `process.binding("uv")` nonetheless

Test

### DIFF
--- a/src/bun.js/bindings/ProcessBindingUV.cpp
+++ b/src/bun.js/bindings/ProcessBindingUV.cpp
@@ -1,4 +1,7 @@
 #include "ProcessBindingUV.h"
+#include "JavaScriptCore/ArrayAllocationProfile.h"
+#include "JavaScriptCore/JSCJSValue.h"
+#include "JavaScriptCore/ThrowScope.h"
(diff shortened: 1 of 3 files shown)


src/bun.js/bindings/ProcessBindingUV.cpp
```diff
@@ -98,37 +101,43 @@ namespace ProcessBindingUV {
...
 
         BUN_UV_ERRNO_MAP(CASE)
 #undef CASE
+    default: {
+        break;
+    }
     }
 
-    return JSValue::encode(jsString(vm, makeString("Unknown system error "_s, String::number(err))));
+    return JSValue::encode(jsString(vm, String("Unknown system error"_s)));
```
paperclover: any reason to not print the number here?

paperclover (approved):

Issue #11049, opened 2024-05-13
Implement Request/Response `.bytes()` method


bakkot:
See https://github.com/whatwg/fetch/issues/1732.

Tests at https://github.com/web-platform-tests/wpt/pull/46198.

The above spec PR adds a `bytes()` method to get a Uint8Array to Body, instead of requiring users to get an ArrayBuffer with `arrayBuffer()` and then wrap it.

There are also PRs to do the same for Blob and PushMessageData, which copy relevant parts of the Body interface:
- https://github.com/w3c/FileAPI/pull/197
- https://github.com/w3c/push-api/pull/369

I don't think Bun has PushMessageData but it does have Blob.

Jarred-Sumner:
Sounds great. We will add this.

bakkot:
You may be interested in [the WebKit PR](https://github.com/WebKit/WebKit/pull/28530) for the same.

PR #11104, opened 2024-05-16
add .bytes() method to various readables

nektro:
Closes https://github.com/oven-sh/bun/issues/11049


src/bun.js/base.zig
```diff
@@ -394,6 +394,13 @@ pub const ArrayBuffer = extern struct {
         };
     }
 
+    pub fn create2(globalThis: *JSC.JSGlobalObject, bytes: []const u8, comptime kind: JSValue.JSType) JSValue {
```
Jarred-Sumner: what is the `2` for?

test/js/web/streams/streams.test.js
```diff
@@ -748,6 +791,41 @@ it("new Response(stream).arrayBuffer() (default)", async () => {
   expect(new TextDecoder().decode(new Uint8Array(buffer))).toBe("abdefgh");
 });
 
+it("new Response(stream).bytes() (bytes)", async () => {
```
Jarred-Sumner, line 809: how about one for `type: "direct"`?

src/bun.js/bindings/BunObject.cpp
```diff
@@ -120,6 +123,7 @@ static inline JSC::EncodedJSValue flattenArrayOfBuffersIntoArrayBuffer(JSGlobalO
             return JSValue::encode(jsUndefined());
         }
     }
+    byteLength = MIN(byteLength, maxLength);
```
```suggestion
    byteLength = std::min(byteLength, maxLength);
```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -22,6 +22,9 @@
 #include "BunObject.h"
 #include "WebCoreJSBuiltins.h"
 #include <JavaScriptCore/JSObject.h>
+#include <cmath>
```
```suggestion

```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -22,6 +22,9 @@
...
+#include <cstddef>
```
```suggestion

```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -22,6 +22,9 @@
...
+#include <sys/param.h>
```
Jarred-Sumner: - Doesn't exist on Windows
- using `std::min` is more idiomatic than the `MIN` macro

```suggestion

```

packages/bun-polyfills/src/modules/bun.ts
```diff
@@ -410,6 +410,21 @@ export const readableStreamToArrayBuffer = ((stream: ReadableStream<ArrayBufferV
...
+    return (async () => {
+        const sink = new ArrayBufferSink();
+        sink.start({ asUint8Array: true });
+        const reader = stream.getReader();
+        while (true) {
+            const { done, value } = await reader.read();
+            if (done) break;
+            sink.write(value);
+        }
+        return sink.end() as ArrayBuffer;
```
Jarred-Sumner:
```suggestion
        return sink.end() as Uint8Array;
```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -188,7 +197,19 @@ JSC_DEFINE_HOST_FUNCTION(functionConcatTypedArrays, (JSGlobalObject * globalObje
 
     auto arrayValue = callFrame->uncheckedArgument(0);
 
-    return flattenArrayOfBuffersIntoArrayBuffer(globalObject, arrayValue);
+    size_t maxLength = SIZE_MAX;
```
Jarred-Sumner: `std::numeric_limits<size_t>` is more idiomatic

```suggestion
    size_t maxLength = std::numeric_limits<size_t>::max();
```

src/bun.js/bindings/BunObject.cpp
```diff
@@ -188,7 +197,19 @@ JSC_DEFINE_HOST_FUNCTION(functionConcatTypedArrays, (JSGlobalObject * globalObje
...
+    auto arg1 = callFrame->argument(1);
+    if (!arg1.isUndefinedOrNull() && arg1.toNumber(globalObject) != INFINITY) {
```
Jarred-Sumner: `toNumber` calls `[Symbol.toPrimitive]`, which means it can throw.

Here's what we do in Buffer:

https://github.com/oven-sh/bun/blob/9a352b7d1202c0ef2e7746f887e9b6c7e15d34c3/src/bun.js/bindings/JSBuffer.cpp#L1290

I suggest something like this:

```suggestion
    if (!arg1.isUndefined() && arg1.isNumber()) {
       double number = arg1.toNumber(globalObject);
       if (std::isnan(number) || number < 0) {
          throwRangeError(lexicalGlobalObject, scope, "Maximum length must be >= 0"_s);
          return {};
       }
       maxLength = std::min(std::numeric_limits<uint32_t>::max(), static_cast<size_t>(maxLength));
```

src/bun.js/bindings/BunObject.cpp
Jarred-Sumner: Also, this should actually be `uint32_t` because the maximum size of a typed array is max uint32.

PR #11106, opened 2024-05-16
test: use resource disposal to ensure servers shutdown even if tests fail

nektro:
`expect()` throws if given a non-passing value so many tests forget to cleanup or are wrapped in `try`/`finally`. with `using` they don't have to.

### DIFF
--- a/test/js/web/fetch/fetch.test.ts
+++ b/test/js/web/fetch/fetch.test.ts
@@ -286,9 +286,8 @@ describe("AbortSignal", () => {
   it("TimeoutError", async () => {
     const signal = AbortSignal.timeout(10);
 
-    let server: Server | null = null;
     try {
-      server = Bun.serve({
+      using server = Bun.serve({
         port: 0,
         async fetch() {
           await Bun.sleep(100);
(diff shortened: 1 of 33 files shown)


Jarred-Sumner (changes requested):
The tests which start a server inside a `beforeAll` / `afterAll` need to continue doing that

It's really easy to use all the CI machine's open ports and then cause other tests to fail

nektro:
doing it this way still only creates one server per test but also has the added benefit that if/once we run tests in parallel they wont clobber the same global variable

test/js/web/fetch/fetch.test.ts
```diff
@@ -1676,25 +1665,9 @@ it("same-origin status code 302 should not strip headers", async () => {
   expect(headers.get("Authorization")).toEqual("yes");
   expect(url).toEndWith("/redirected");
   expect(redirected).toBe(true);
-  server.stop(true);
 });
 
 describe("should handle relative location in the redirect, issue#5635", () => {
-  var server: Server;
```
Jarred-Sumner, line 1683: can we bring this back

Jarred-Sumner:
Thank you

PR #11135, opened 2024-05-17
node:fs: add Dirent.path and .parentPath fields

nektro:
Closes https://github.com/oven-sh/bun/issues/7358

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -5176,6 +5176,7 @@ pub const NodeFS = struct {
 
         var iterator = DirIterator.iterate(fd.asDir(), .u8);
         var entry = iterator.next();
+        var dirent_path_prev: ?bun.String = null;
 
         while (switch (entry) {
             .err => |err| {
@@ -5228,9 +5229,15 @@ pub const NodeFS = struct {
 
             switch (comptime ExpectedType) {
                 Dirent => {
+                    const path_u8 = bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto);
+                    if (dirent_path_prev != null and bun.strings.eql(dirent_path_prev.?.byteSlice(), path_u8)) {
+                        dirent_path_prev.?.ref();
+                    } else {
+                        dirent_path_prev = bun.String.createUTF8(path_u8);
+                    }
                     entries.append(.{
                         .name = bun.String.createUTF8(utf8_name),
-                        .path = bun.String.createUTF8(bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto)),
+                        .path = dirent_path_prev.?,
                         .kind = current.kind,
                     }) catch bun.outOfMemory();
                 },
@@ -5320,6 +5327,7 @@ pub const NodeFS = struct {
 
             var iterator = DirIterator.iterate(fd.asDir(), .u8);
             var entry = iterator.next();
+            var dirent_path_prev: ?bun.String = null;
 
             while (switch (entry) {
                 .err => |err| {
@@ -5358,9 +5366,15 @@ pub const NodeFS = struct {
 
                 switch (comptime ExpectedType) {
                     Dirent => {
+                        const path_u8 = bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto);
+                        if (dirent_path_prev != null and bun.strings.eql(dirent_path_prev.?.byteSlice(), path_u8)) {
+                            dirent_path_prev.?.ref();
+                        } else {
+                            dirent_path_prev = bun.String.createUTF8(path_u8);
+                        }
                         entries.append(.{
                             .name = bun.String.createUTF8(utf8_name),
-                            .path = bun.String.createUTF8(bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto)),
+                            .path = dirent_path_prev.?,
                             .kind = current.kind,
                         }) catch bun.outOfMemory();
                     },
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -5230,11 +5230,10 @@ pub const NodeFS = struct {
             switch (comptime ExpectedType) {
                 Dirent => {
                     const path_u8 = bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto);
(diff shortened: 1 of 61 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -5217,6 +5220,7 @@ pub const NodeFS = struct {
                 Dirent => {
                     entries.append(.{
                         .name = bun.String.createUTF8(name_to_copy),
+                        .path = bun.String.createUTF8(basename),
```
paperclover: note: this is going to clone the same string value per file in the directory when it doesn't really need to since theyre all the same. `bun.String.createUTF8` creates a WTF::String that can just have a ref-count.

src/bun.js/node/node_fs.zig
```diff
@@ -5119,6 +5129,8 @@ pub const NodeFS = struct {
         comptime is_root: bool,
     ) Maybe(void) {
         const flags = os.O.DIRECTORY | os.O.RDONLY;
+        const dirent_path = bun.String.createUTF8(basename);
```
Jarred-Sumner: Do we need to create the `dirent_path` if `ExpectedType` != `Dirent`?

Is it possible to reuse the input path the user passed instead of converting to UTF-8, then creating a duplicate copy of that?

Jarred-Sumner (commented):
Thanks for this

A couple comments, but it's close to mergeable.

Needs:
- A test that calls the `path` getter on a `recursive: true` `fs.promises.readdir` about 200 times so if there are race conditions/incorrect memory handling when used concurrently it will crash
- A test that calls the `path` getter on `recursive: false` `fs.promises.readdir`
- A test that calls the `path` getter on `recursive: true` `readdirSync` which checks that the path matches the output of node (or at least a fixture of some kind)

src/bun.js/node/node_fs.zig
```diff
@@ -5036,15 +5036,17 @@ pub const NodeFS = struct {
...
         entries: *std.ArrayList(ExpectedType),
     ) Maybe(void) {
         const dir = fd.asDir();
         const is_u16 = comptime Environment.isWindows and (ExpectedType == bun.String or ExpectedType == Dirent);
-        var iterator = DirIterator.iterate(
-            dir,
-            comptime if (is_u16) .u16 else .u8,
-        );
+
+        const dirent_path = bun.String.createUTF8(basename);
```
Jarred-Sumner: Let's only create this String when `ExpectedType` == `Dirent`

src/bun.js/node/node_fs.zig
```diff
@@ -5216,7 +5222,8 @@ pub const NodeFS = struct {
             switch (comptime ExpectedType) {
                 Dirent => {
                     entries.append(.{
-                        .name = bun.String.createUTF8(name_to_copy),
+                        .name = bun.String.createUTF8(utf8_name),
+                        .path = bun.String.createUTF8(std.fs.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto)).?),
```
Jarred-Sumner: is `.?` guaranteed? What if you did `readdir("/")`?

src/bun.js/node/node_fs.zig
```diff
@@ -5345,7 +5352,8 @@ pub const NodeFS = struct {
                 switch (comptime ExpectedType) {
                     Dirent => {
                         entries.append(.{
-                            .name = bun.String.createUTF8(name_to_copy),
+                            .name = bun.String.createUTF8(utf8_name),
+                            .path = bun.String.createUTF8(std.fs.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto)).?),
```
Jarred-Sumner: Will this be the same `path` string when there are multiple entries in a directory? For a directory structure like

- / 📁 foo
- / 📁 foo / 📄  bar.txt
- / 📁 foo / 📄  baz.js

is this duplicating `/foo` when setting `.path`?

src/bun.js/node/node_fs.zig
Jarred-Sumner: and we can do it lazily. if there are no entries, we should not create this String at all.

Jarred-Sumner (changes requested):
A couple comments

src/bun.js/node/node_fs.zig
nektro: good point, I should use `bun.path.dirname` that handles that

src/bun.js/node/node_fs.zig
nektro: in that case yes

Jarred-Sumner: Can we avoid that?

src/bun.js/node/node_fs.zig
```diff
@@ -5163,11 +5176,12 @@ pub const NodeFS = struct {
 
         var iterator = DirIterator.iterate(fd.asDir(), .u8);
         var entry = iterator.next();
+        var dirent_path_prev: ?bun.String = null;
```
Jarred-Sumner, line 5179: Where is this deref'd?

src/bun.js/node/types.zig
```diff
@@ -1769,6 +1770,10 @@ pub const Dirent = struct {
         return this.name.toJS(globalObject);
     }
 
+    pub fn getPath(this: *Dirent, globalObject: *JSC.JSGlobalObject) callconv(.C) JSC.JSValue {
+        return this.path.toJS(globalObject);
```
```suggestion
        return this.path.transferToJS(globalObject);
```

src/bun.js/node/types.zig
```diff
@@ -1769,6 +1770,10 @@ pub const Dirent = struct {
         return this.name.toJS(globalObject);
```
Jarred-Sumner, line 1770:
```suggestion
        return this.name.transferToJS(globalObject);
```

Jarred-Sumner (commented):
1 comment

Jarred-Sumner:
Thank you

src/bun.js/node/node_fs.zig
```diff
@@ -5215,8 +5229,14 @@ pub const NodeFS = struct {
 
             switch (comptime ExpectedType) {
                 Dirent => {
+                    const path_u8 = bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto);
+                    if (dirent_path_prev == null or bun.strings.eql(dirent_path_prev.?.byteSlice(), path_u8)) {
```
Jarred-Sumner, line 5233: @nektro was this line supposed to be?

```suggestion
                    if (dirent_path_prev == null or !bun.strings.eql(dirent_path_prev.?.byteSlice(), path_u8)) {
```

PR #11140, opened 2024-05-17
Use simdutf to validate utf8 in websocket server

Jarred-Sumner:
### What does this PR do?

Use simdutf to validate utf8 in websocket server

Should be slightly faster

Existing tests

### DIFF
--- a/packages/bun-uws/src/WebSocketProtocol.h
+++ b/packages/bun-uws/src/WebSocketProtocol.h
@@ -25,6 +25,9 @@
 #include <cstdlib>
 #include <string_view>
 
+// bun-specific
+#include "simdutf.h"
+
 namespace uWS {
 
 /* We should not overcomplicate these */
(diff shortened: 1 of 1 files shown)


packages/bun-uws/src/WebSocketProtocol.h
```diff
@@ -109,50 +112,9 @@ T cond_byte_swap(T value) {
...
-                    (s[0] == 0xf0 && (s[1] & 0xf0) == 0x80) || (s[0] == 0xf4 && s[1] > 0x8f) || s[0] > 0xf4) {
-                return false;
-            }
-            s += 4;
-        } else {
-            return false;
-        }
-    }
-    return true;
+    return simdutf::validate_utf8(reinterpret_cast<const char *>(s), length);
```
zackradisic, line 117: Maybe we should fallback to the previous code when the string is small like < 64 bytes. Or does simdutf do this already?

zackradisic (approved):
looks good

packages/bun-uws/src/WebSocketProtocol.h
Jarred-Sumner: i think they use stride lengths specific to the supported CPU instructions and then fallback to a scalar implementation

PR #11176, opened 2024-05-19
Introduce `bun.dns.prefetch` API

Jarred-Sumner:
### What does this PR do?

This adds the `bun.dns.prefetch` API.

```ts
import {dns} from "bun";

dns.prefetch("bun.sh", 443);
//
// ... sometime later ...
await fetch("https://bun.sh");
```

Like `<link rel="dns-prefetch">` in browsers, the intended usecase is to amortize the cost of DNS lookups you know you'll make soon.

This also sets the DNS cache to have a TTL of 30 seconds. For reference, the Java Virtual Machine has no TTL set and [AWS recommends 5 seconds](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/jvm-ttl-dns.html).

This also introduces a `dns.getCacheStats()` API which exposes information about the DNS cache.

This also adds documentation.

This also makes `bun install` prefetch the DNS query. 

There is a test

### DIFF
--- a/docs/api/dns.md
+++ b/docs/api/dns.md
@@ -8,34 +8,83 @@ console.log(addrs);
 // => [{ address: "172.67.161.226", family: 4, ttl: 0 }, ...]
 ```
 
-<!--
-## `Bun.dns` - lookup a domain
-`Bun.dns` includes utilities to make DNS requests, similar to `node:dns`. As of Bun v0.5.0, the only implemented function is `dns.lookup`, though more will be implemented soon.
-You can lookup the IP addresses of a hostname by using `dns.lookup`.
-```ts
-import { dns } from "bun";
-const [{ address }] = await dns.lookup("example.com");
-console.log(address); // "93.184.216.34"
-```
-If you need to limit IP addresses to either IPv4 or IPv6, you can specify the `family` as an option.
-```ts
-import { dns } from "bun";
-const [{ address }] = await dns.lookup("example.com", { family: 6 });
-console.log(address); // "2606:2800:220:1:248:1893:25c8:1946"
-```
-Bun supports three backends for DNS resolution:
-- `c-ares` - This is the default on Linux, and it uses the [c-ares](https://c-ares.org/) library to perform DNS resolution.
-- `system` - Uses the system's non-blocking DNS resolver, if available. Otherwise, falls back to `getaddrinfo`. This is the default on macOS, and the same as `getaddrinfo` on Linux.
-- `getaddrinfo` - Uses the POSIX standard `getaddrinfo` function, which may cause performance issues under concurrent load.
+## DNS caching in Bun
 
-You can choose a particular backend by specifying `backend` as an option.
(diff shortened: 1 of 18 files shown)


src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -1263,28 +1310,47 @@ pub const InternalDNS = struct {
...
             key: Request.Key,
+            timestamp_to_store: *u32,
         ) ?*Request {
-            for (this.cache[0..this.len]) |entry| {
+            const len = this.len;
+            for (this.cache[0..len], 0..) |entry, i| {
                 if (entry.key.hash == key.hash and entry.key.port == key.port and entry.valid) {
+                    if (entry.isExpired(timestamp_to_store)) {
+                        log("get: expired entry", .{});
+                        this.deleteEntryAt(len, i);
```
zackradisic: This places the last cache entry at `i`, but then increments `i` so I think the newly swapped value will get skipped and never checked

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -1263,28 +1310,47 @@ pub const InternalDNS = struct {
...
 
         fn remove(this: *This, entry: *Request) void {
             const len = this.len;
             // equivalent of swapRemove
             for (0..len) |i| {
                 if (this.cache[i] == entry) {
-                    this.cache[i] = this.cache[len - 1];
-                    this.len -= 1;
-                    dns_cache_size = len - 1;
+                    this.deleteEntryAt(len, i);
```
zackradisic: same thing here

cirospaciari (approved):

docs/api/dns.md
```diff
@@ -8,34 +8,106 @@ console.log(addrs);
...
+
+An example where you might want to use this is a database driver. When your application first starts up, you can prefetch the DNS entry for the database host so that by the time it finishes loading everything, the DNS query to resolve the database host may already be completed.
+
+### `dns.prefetch`
+
+{% callout %}
+**🚧** — This API is experimental and may change in the future.
+{% /callout %}
+
+To prefetch a DNS entry, you can use the `dns.prefetch` API. This API is useful when you know you'll need to connect to a host in the soon and want to avoid the initial DNS lookup.
```
mangs:
```suggestion
To prefetch a DNS entry, you can use the `dns.prefetch` API. This API is useful when you know you'll need to connect to a host soon and want to avoid the initial DNS lookup.
```

docs/api/dns.md
```diff
@@ -8,34 +8,106 @@ console.log(addrs);
...
+To get the current cache stats, you can use the `dns.getCacheStats` API. 
+
+This API returns an object with the following properties:
+
+```ts
+{
+  // Cache hits 
+  cacheHitsCompleted: number;
+  cacheHitsInflight: number;
+  cacheMisses: number;
```
mangs, line 78: would be useful to see these values per hostname as well to see where lookup activity is concentrated. could generate some useful graphs from this too.

PR #11185, opened 2024-05-20
Optimize fish completions

ridiculousfish:
### What does this PR do?

See [this Jarred tweet](https://twitter.com/jarredsumner/status/1790603279044661513) for context.

It was observed that the fish shell completions for bun invokes bun too frequently. The primary reason for the completion to invoke bun, is to discover the installed package binaries and scripts, with `bun getcompletes`.

Jarred estimated ~30 invocations of bun after `bun <tab>`; with these changes there are only 2 invocations, one for `bun getcompletes b` and one for `bun getcompletes z`.

The optimizations here are:

1. Emit completions for both package binaries and scripts from a single function. This allows us to coalesce these `bun getcompletes` calls.
2. Simplify the conditionals (when completions are offered). Most completions tried to suppress themself if there were a package script or binary but I don't think this is generally necessary; most completions are just subcommands and should be offered if there is not yet a subcommand.

I performed manual testing to validate that commands generally made sense. But I am not a heavy bun user so it's possible I've introduced a regression or missed an improvement, so I would appreciate more testing.

The fish completions may be loaded by running (in fish of course):

```sh
> source /path/to/bun/completions/fish.complete
```

then inside a package, try various `bun <tab>` or `bun subcommand <tab>` scenarios.

### DIFF
--- a/completions/bun.fish
+++ b/completions/bun.fish
@@ -53,7 +53,6 @@ end
 set -l bun_install_boolean_flags yarn production optional development no-save dry-run force no-cache silent verbose global
 set -l bun_install_boolean_flags_descriptions "Write a yarn.lock file (yarn v1)" "Don't install devDependencies" "Add dependency to optionalDependencies" "Add dependency to devDependencies" "Don't install devDependencies" "Don't install anything" "Always request the latest versions from the registry & reinstall all dependencies" "Ignore manifest cache entirely" "Don't output anything" "Excessively verbose logging" "Use global folder"
 
-set -l bun_builtin_cmds dev create help bun upgrade discord run install remove add init link unlink pm x
 set -l bun_builtin_cmds_without_run dev create help bun upgrade discord install remove add init pm x
 set -l bun_builtin_cmds_without_bun dev create help upgrade run discord install remove add init pm x
 set -l bun_builtin_cmds_without_create dev help bun upgrade discord run install remove add init pm x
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Two notes for review:

1. I have deliberately broken this up into multiple commits to ease bisecting if a regression is later found.
2. Please double check the `bun_builtin_cmds_accepting_flags` change. It looks like the original intent was that the options `--origin`, `--port`, `--define`, `--external`, `--use`, and `--hot` would apply to the subcommands `create help bun upgrade discord run init link unlink pm x`; I attempted to propagate that but I don't know if it's desired.

Jarred-Sumner:
I played around with it for a bit locally. It feels faster. Thank you for helping.

> Please double check the bun_builtin_cmds_accepting_flags change. It looks like the original intent was that the options --origin, --port, --define, --external, --use, and --hot would apply to the subcommands create help bun upgrade discord run init link unlink pm x; I attempted to propagate that but I don't know if it's desired.

Some of those don't make sense anymore but I'll go ahead and update that in a follow-up PR. We only want to propagate `--port` for both `bun run <foo>` and `bun <foo>`. `--external` is `bun build --external` only (though at the time this was initially written, it was not).

Most of why this is complicated is because `bun run <foo>` and `bun <foo>` are very similar but subtly different in preference order, and there are shared flags between all.

PR #11206, opened 2024-05-20
happy eyeballs

gvilums:
### What does this PR do?
When connecting a socket we initiate simultaneous connections to all addresses returned by `getaddrinfo`, and keep the first one.

Existing tests.

### DIFF
--- a/packages/bun-usockets/src/context.c
+++ b/packages/bun-usockets/src/context.c
@@ -394,8 +394,11 @@ struct us_connecting_socket_t *us_socket_context_connect(int ssl, struct us_sock
     if (Bun__addrinfo_get(loop, host, port, &ptr) == 0) {
         // Fast-path: it's already cached.
         // Avoid the connection logic.
-        *is_connecting = 1;
-        return (struct us_connecting_socket_t *) us_socket_context_connect_resolved_dns(context, ptr, options, socket_ext_size);
+
+        // with happy eyeballs we can't use the fast path since we need to initiate multiple connections
+
+        // *is_connecting = 1;
+        // return (struct us_connecting_socket_t *) us_socket_context_connect_resolved_dns(context, ptr, options, socket_ext_size);
     }
 
     struct us_connecting_socket_t *c = us_calloc(1, sizeof(struct us_connecting_socket_t) + socket_ext_size);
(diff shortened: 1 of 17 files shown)


packages/bun-usockets/src/context.c
```diff
@@ -394,8 +394,11 @@ struct us_connecting_socket_t *us_socket_context_connect(int ssl, struct us_sock
     if (Bun__addrinfo_get(loop, host, port, &ptr) == 0) {
         // Fast-path: it's already cached.
         // Avoid the connection logic.
-        *is_connecting = 1;
-        return (struct us_connecting_socket_t *) us_socket_context_connect_resolved_dns(context, ptr, options, socket_ext_size);
+
+        // with happy eyeballs we can't use the fast path since we need to initiate multiple connections
```
Jarred-Sumner: this is not quite accurate

we only should go through the happy eyeballs case when there are multiple results from addrinfo

gvilums: makes sense, I updated the code to use the fast path when there is a cached addrinfo result that has only one entry

gvilums:
One of the test failures ("request body and signal life cycle" in `serve.test.ts` is not caused by this PR, it can be reproduced on stable bun as shown in #11208

Jarred-Sumner:
- Fails to compile on Linux
- https://discord.com/channels/1021987251930005594/1022006147751882782/1242283397502603375 windows test failures

src/deps/uws.zig
```diff
@@ -682,22 +690,20 @@ pub fn NewSocketHandler(comptime is_ssl: bool) type {
...
 
-            const host: ?[*:0]u8 = brk: {
-                // getaddrinfo expects `node` to be null if localhost
-                if (raw_host.len < 6 and (bun.strings.eqlComptime(raw_host, "[::1]") or bun.strings.eqlComptime(raw_host, "[::]"))) {
-                    break :brk null;
-                }
-
-                break :brk allocator.dupeZ(u8, raw_host) catch bun.outOfMemory();
-            };
+            // remove brackets from IPv6 addresses, as getaddrinfo doesn't understand them
```
Jarred-Sumner, line 693: i think there are probably like 10 more cases here to handle btw

PR #11218, opened 2024-05-21
fix `which` when bin or PATH is rlly long

### LINKED ISSUE #11203: Crash report

0. Ensure you have Docker installed in Windows with WSL integration on.
1. Open WSL in VS Code
2. Open a folder in VS Code
3. put this file in folder: [2024-05-20-23-12-main-seelie.json](https://github.com/oven-sh/bun/files/15380414/2024-05-20-23-12-main-seelie.json) & my code
4. Open folder in dev container (Used this [template](https://github.com/marcosgomesneto/bun-devcontainers/tree/main/src/basic-bun) from [here](https://containers.dev/templates)
5. Run the code
6. Code crashed

### JavaScript/TypeScript code that reproduces the crash?

```shell
const seelie = require('./2024-05-20-23-12-main-seelie.json')
import {$} from 'bun';
seelie.goals.forEach(goal => {
    if (goal.type != 'Talent') {
        return;
    }

    goal.normal.goal = goal.normal.goal < 9 ? 9 : goal.normal.goal;
    goal.skill.goal = goal.skill.goal < 9 ? 9 : goal.skill.goal;
    goal.burst.goal = goal.burst.goal < 9 ? 9 : goal.burst.goal;
})

await $`${JSON.stringify(seelie, null, 2)} > modified.json`
```

### Relevant log output

```shell
const seelie = require('./2024-05-20-23-12-main-seelie.json')
import {$} from 'bun';
seelie.goals.forEach(goal => {
    if (goal.type != 'Talent') {
        return;
    }

    goal.normal.goal = goal.normal.goal < 9 ? 9 : goal.normal.goal;
    goal.skill.goal = goal.skill.goal < 9 ? 9 : goal.skill.goal;
    goal.burst.goal = goal.burst.goal < 9 ? 9 : goal.burst.goal;
})

await $`${JSON.stringify(seelie, null, 2)} > modified.json`
```

### Stack Trace (bun.report)

Bun v1.1.8 ([`89d2580`](https://github.com/oven-sh/bun/tree/89d25807fbf4488d3bfcb116484d0c2cf824f98c)) on linux x86_64_baseline [AutoCommand]

Segmentation fault at address 0x00000000

- [`ld-temp.o:0`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/ld-temp.o#L0): `void WTF::URLParser::parse<char16_t>`
- `??`

zackradisic:
### What does this PR do?

We were using `bun.copy` to copy bin or path segment into a path buffer, but not checking that it would actually fit inside the path buffer, causing a crash on windows or undefined behavior elsewhere

An example:
```typescript
const rlly_long_str = "lol".repeat(10000000)
await Bun.$`${rlly_long_str}` // shell will use which to see if rlly_long_str is bin in PATH
```

I made the `bun.which` function return null in this case, maybe it is more correct to change the function signature and return `bun.C.E.NAMETOOLONG`?

Fixes #11203

added tests

### DIFF
--- a/src/which.zig
+++ b/src/which.zig
@@ -3,6 +3,7 @@ const bun = @import("root").bun;
 const PosixToWinNormalizer = bun.path.PosixToWinNormalizer;
 
 fn isValid(buf: *bun.PathBuffer, segment: []const u8, bin: []const u8) ?u16 {
+    if (segment.len + 1 + bin.len > bun.MAX_PATH_BYTES) return null;
     bun.copy(u8, buf, segment);
     buf[segment.len] = std.fs.path.sep;
     bun.copy(u8, buf[segment.len + 1 ..], bin);
@@ -15,6 +16,7 @@ fn isValid(buf: *bun.PathBuffer, segment: []const u8, bin: []const u8) ?u16 {
 // Like /usr/bin/which but without needing to exec a child process
 // Remember to resolve the symlink if necessary
 pub fn which(buf: *bun.PathBuffer, path: []const u8, cwd: []const u8, bin: []const u8) ?[:0]const u8 {
+    if (bin.len > bun.MAX_PATH_BYTES) return null;
     bun.Output.scoped(.which, true)("path={s} cwd={s} bin={s}", .{ path, cwd, bin });
 
     if (bun.Environment.os == .windows) {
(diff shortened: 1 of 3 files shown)


Jarred-Sumner:
I don't like the flaky socket tests (cc @gvilums)

Jarred-Sumner:
> I made the bun.which function return null in this case

I think that's fine. A file path which is too long is not executable

PR #11226, opened 2024-05-21
fix(http2) fix aborted behavior

cirospaciari:
### What does this PR do?
`streamError` event is not a thing, so we should not emit it, when aborted should not emit `error`.

- [x] Code changes

Updated tests

### DIFF
--- a/src/js/node/http2.ts
+++ b/src/js/node/http2.ts
@@ -654,28 +654,30 @@ function emitStreamNT(self, streams, streamId) {
 
 function emitStreamErrorNT(self, streams, streamId, error, destroy) {
   const stream = streams.get(streamId);
-  const error_instance = streamErrorFromCode(error);
   if (stream) {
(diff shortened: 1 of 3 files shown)


src/js/node/http2.ts
```diff
@@ -834,15 +837,15 @@ class ClientHttp2Session extends Http2Session {
...
-        const error_instance = streamErrorFromCode(constants.NGHTTP2_CANCEL);
         if (!stream[bunHTTP2Closed]) {
           stream[bunHTTP2Closed] = true;
         }
 
         stream.rstCode = constants.NGHTTP2_CANCEL;
-        stream.emit("aborted", error);
-        stream.emit("error", error_instance);
-        self.emit("streamError", error_instance);
+        stream.emit("aborted");
```
Jarred-Sumner, line 845: i don't think this is right

i think it should be done inside .._destroy

PR #11233, opened 2024-05-21
Fix crash in crypto digest

Jarred-Sumner:
### What does this PR do?

Fix crash in crypto digest

There is a test


nektro (commented):
failing tests, subsumed into https://github.com/oven-sh/bun/pull/11234

PR #11238, opened 2024-05-21
bypass getaddrinfo for ip addresses

gvilums:
### What does this PR do?
This PR makes it so that when we try to open a new TCP connection, we attempt to parse `host` with `inet_pton`. If we succeed, we use the resulting ip address and bypass `getaddrinfo` entirely.

A test (it's hard to test for this specifically)

### DIFF
--- a/test/js/bun/net/socket-leak-fixture.js
+++ b/test/js/bun/net/socket-leak-fixture.js
@@ -3,7 +3,7 @@ import { expect } from "bun:test";
 
 const server = Bun.listen({
   port: 0,
-  hostname: "localhost",
+  hostname: "127.0.0.1",
   socket: {
     open(socket) {
       socket.end();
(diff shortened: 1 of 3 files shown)


test/js/bun/net/socket-leak-fixture.js
```diff
@@ -3,7 +3,7 @@ import { expect } from "bun:test";
 
 const server = Bun.listen({
   port: 0,
-  hostname: "localhost",
+  hostname: "127.0.0.1",
```
Electroid: Is there a reason we changed this?

Jarred-Sumner (approved):

test/js/bun/net/socket-leak-fixture.js
gvilums: this change was unintended

PR #11240, opened 2024-05-21
fix(windows spawn): use Job Object to manage subprocesses of subprocesses

paperclover:
### What does this PR do?

See test case, note how --watch spawns two processes, but before, the `kill` in the outermost process would keep the innermost subprocess alive. Now, calling `kill` in the outermost process kills the whole tree. This was an issue if the caller of this process tree depended on standard io, which would not be closed before. Now, since the processes are dead, the standard io is closed.


src/bun.js/api/bun/process.zig
```diff
@@ -1716,6 +1716,13 @@ pub fn spawnProcessWindows(
         .extra_pipes = try std.ArrayList(WindowsSpawnResult.StdioResult).initCapacity(bun.default_allocator, options.extra_fds.len),
     };
 
+    // By putting this process in a Job Object, any child process will be caught
+    // so that killing this process will kill all children processes.
+    //
+    // It seems that just associating the process does the work neccecary.
+    if (bun.windows.CreateJobObjectA(null, null)) |job|
+        _ = bun.windows.AssignProcessToJobObject(job, process.poller.uv.process_handle);
```
gvilums: What happens if the process has already spawned a child process and then exited by the time it's being assigned to the job object? See [this blog](https://devblogs.microsoft.com/oldnewthing/20230209-00/?p=107812) for context. Not sure if this is an edge case we should care about, but might be relevant.

To do it the "right" way we might need to mess with libuv internals for process spawning, not sure if we want to do that.

gvilums (commented):
Looks good in general, see my comment about the race condition

src/bun.js/api/bun/process.zig
paperclover: i moved this to just the code for --watch (libuv already does this and actually did it more correct than what i have here), and i followed the linked guide so that we spawn the process without this race. thank you.

gvilums:
Changes look good now, however there seems to be a test failure in `watch.test.ts` on windows - I'm not sure if it was there before, but I think it wasn't

src/bun.js/api/bun/process.zig
gvilums: Looks good now 👍

gvilums (approved):

paperdave:
these tests are definitely regressions. its a panic from failing to create the job object 

image

PR #11243, opened 2024-05-22
ci: make test/js/bun/shell/commands/rm.test.ts less flaky

nektro:
typical times for the node_modules test range from 2713.78ms to 9518.02ms

slightly increasing the timeout length and lowering the parallelism should make this much less flaky to network conditions

### DIFF
--- a/src/bun.js/module_loader.zig
+++ b/src/bun.js/module_loader.zig
@@ -991,7 +991,7 @@ pub const ModuleLoader = struct {
 
             pub fn pollModules(this: *Queue) void {
                 var pm = this.vm().packageManager();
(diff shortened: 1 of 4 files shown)


nektro:
if this fails with this remedy I'll move it to a separate file and do the install outside the test

Jarred-Sumner (changes requested):
The cause of flakiness is `bun install` making external network requests that take potentially a long time.

This test isn't meant to test bun install. It's meant to test `rm`. So let's just not run `bun install`, and instead create a big folder with a bunch of files and then verify that recursively deleting the folder works.

PR #11248, opened 2024-05-22
docs: Add guide for deploying a Bun app to Render

jess-render:
### What does this PR do?

Adds a guide that shows how to deploy a Bun app on Render.

- [ ] Code changes

Manually tested by following the instructions

### DIFF
--- a/docs/guides/ecosystem/render.md
+++ b/docs/guides/ecosystem/render.md
@@ -4,7 +4,7 @@ name: Deploy a Bun application on Render
 
 [Render](https://render.com/) is a cloud platform that lets you flexibly build, deploy, and scale your apps.
 
-It offers features like auto deploys from GitHub, a global CDN, private networks, automatic HTTPS setup, and managed PostgreSQL and Redis instances.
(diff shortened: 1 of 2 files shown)


jess-render:
cc @Jarred-Sumner (since I don't seem to have permission to add a Reviewer or Assignee)

docs/guides/ecosystem/render.md
```diff
@@ -0,0 +1,77 @@
...
+
+In your [Render Dashboard](https://dashboard.render.com/), click `New` > `Web Service` and connect your `myapp` repo.
+
+---
+
+In the Render UI, provide the following values during web service creation:
+
+|             |           |
+| ----------- | --------- |
+| **Runtime** | `Node` |
```
gvilums, line 69: Is `Node` the right value for `Runtime`?

jess-render: Yes -- we don't have a separate one for Bun currently. Both Bun and Node are available in the "Node" runtime. 

We considered naming it "Node / Bun" but some teammates thought that might be confusing at the moment.

gvilums: Makes sense 👍

gvilums (approved):

PR #11286, opened 2024-05-23
windows: fix bunx ci test

nektro:
I have noticed it can sometimes rarely still fail which im looking into. but it used to fail every time consistently so I wanted to get the progress in a branch

ci logs used to be inundated with errors like:

```
error: moving "lru-cache" to cache dir failed
ENOTEMPTY: Directory not empty (NtSetInformationFile())
  From: .f7f5ffdec4e5b3bf-00000002.lru-cache
    To: lru-cache@6.0.0
```

### DIFF
--- a/src/install/extract_tarball.zig
+++ b/src/install/extract_tarball.zig
@@ -158,6 +158,9 @@ threadlocal var folder_name_buf: bun.PathBuffer = undefined;
 threadlocal var json_path_buf: bun.PathBuffer = undefined;
 
 fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractData {
+    return this.extractN(tgz_bytes, 0);
+}
+fn extractN(this: *const ExtractTarball, tgz_bytes: []const u8, n: u32) !Install.ExtractData {
     const tmpdir = this.temp_dir;
     var tmpname_buf: if (Environment.isWindows) bun.WPathBuffer else bun.PathBuffer = undefined;
     const name = this.name.slice();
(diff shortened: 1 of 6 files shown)


src/install/extract_tarball.zig
```diff
@@ -359,6 +365,19 @@ fn extract(this: *const ExtractTarball, tgz_bytes: []const u8) !Install.ExtractD
...
                             else => {},
                         }
+                    } else {
+                        switch (err.getErrno()) {
+                            .NOTEMPTY => {
+                                _ = bun.sys.close(dir_to_move);
+
+                                // another bun install is writing this same package.
+                                // try again in just a bit.
+                                std.time.sleep(std.time.ns_per_s);
```
Jarred-Sumner, line 375: i don't think we should do this, it's not going to work very well

Jarred-Sumner (changes requested):
There are two issues with this PR.

1) Extracting the tarball all-over-again doesn't work and would cause significant performance issues (and consume unnecessary disk space). The gzip'd contents are freed in this function after they're used (potentially dozens of megabytes of memory), so it is either a use-after-free that just happened to work when you ran it locally or it works 0% of the time. Since we are extracting to a temporary directory, we can assume the files are still unmodified  (which is why we did it this way). So if we did want to retry after a delay, the right place to do that would be _after_ extracting and not re-extracting the entire tarball.

2) Retrying with a delay doesn't work reliably (as shown by the `bunx` test continuing to fail in CI). On POSIX, this test isn't failing because we use `renameat` followed by `renameat2` to atomically swap the files as a fallback. I do not think there is an equivalent on Windows. 

I think it'd be better to skip the test on Windows than to attempt to fix it this way, unless you find a better system API to use to handle this edgecase.

PR #11304, opened 2024-05-23
Save to package manifest cache async, only check disk once

Jarred-Sumner:
### What does this PR do?

Save these .npm cache files without blocking the main thread

Avoid reading from disk multiple times

Move the selected lines to the thread pool:
image

Existing tests. It should maybe make the bun-install-registry test less flaky

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -2350,10 +2350,25 @@ const PackageManifestMap = struct {
     }
 
     pub fn byNameHash(this: *PackageManifestMap, name_hash: PackageNameHash) ?*Npm.PackageManifest {
+        return byNameHashAllowExpired(this, name_hash, null);
+    }
+
+    pub fn byNameAllowExpired(this: *PackageManifestMap, name: string, is_expired: ?*bool) ?*Npm.PackageManifest {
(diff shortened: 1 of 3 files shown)


src/install/install.zig
```diff
@@ -4503,13 +4564,9 @@ pub const PackageManager = struct {
...
                                     if (this.options.enable.manifest_cache) {
-                                        if (Npm.PackageManifest.Serializer.load(this.allocator, this.getCacheDirectory(), name_str) catch null) |manifest_| {
-                                            const manifest: Npm.PackageManifest = manifest_;
-                                            loaded_manifest = manifest;
-
-                                            if (this.options.enable.manifest_cache_control and manifest.pkg.public_max_age > this.timestamp_for_manifest_cache_control) {
-                                                try this.manifests.put(this.allocator, manifest.pkg.name.hash, manifest);
-                                            }
+                                        var expired = false;
+                                        if (this.manifests.byNameAllowExpired(name_str, &expired)) |manifest| {
```
dylan-conway:
```suggestion
                                        if (this.manifests.byNameHashAllowExpired(name_hash, &expired)) |manifest| {
```

PR #11322, opened 2024-05-24
feat(https/fetch): Support custom ca/cert/key in fetch

LudvigHz:
### What does this PR do?

This is a rebase of #6621 in an attempt to get this feature merged.

- [x] Code changes

I wrote automated tests

- [x] JSValue use outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed


src/bun.js/webcore/response.zig
```diff
@@ -1610,6 +1612,14 @@ pub const Fetch = struct {
                 fetch_tasklet.http.?.client.remaining_redirect_count = 0;
             }
 
+            fetch_tasklet.http.?.client.disable_timeout = fetch_options.disable_timeout;
```
Jarred-Sumner: These values are copied in `Options` above

src/http.zig
```diff
@@ -364,6 +369,34 @@ fn NewHTTPContext(comptime ssl: bool) type {
             return @as(*BoringSSL.SSL_CTX, @ptrCast(this.us_socket_context.getNativeHandle(true)));
         }
 
+        pub fn deinit(this: *@This()) void {
+            this.us_socket_context.deinit(ssl);
+            uws.us_socket_context_free(@as(c_int, @intFromBool(ssl)), this.us_socket_context);
```
Jarred-Sumner, line 386: I think there's likely a UAF lurking here. We need to make sure that there are no actively open sockets that belong to this context at the time this is called. The easiest way to do that would be to disable HTTP keep-alive when a custom ca/cert/key is passed.

src/bun.js/webcore/response.zig
```diff
@@ -2015,14 +2028,22 @@ pub const Fetch = struct {
 
                         if (options.get(ctx, "tls")) |tls| {
                             if (!tls.isEmptyOrUndefinedOrNull() and tls.isObject()) {
+                                if (SSLConfig.inJS(globalThis, tls, exception)) |config| {
+                                    if (ssl_config) |existing_conf| {
+                                        existing_conf.deinit();
+                                        bun.default_allocator.destroy(existing_conf);
+                                        ssl_config = null;
+                                    }
+                                    ssl_config = bun.default_allocator.create(SSLConfig) catch @panic("out of memory!");
```
cirospaciari:
```suggestion
                                    ssl_config = bun.default_allocator.create(SSLConfig)  catch bun.outOfMemory();
```

src/bun.js/webcore/response.zig
```diff
@@ -2015,14 +2028,22 @@ pub const Fetch = struct {
...
+                                    ssl_config.?.* = config;
+                                }
                                 if (tls.get(ctx, "rejectUnauthorized")) |reject| {
```
cirospaciari, line 2097: we also have `rejectUnauthorized` inside SSLConfig, probably dont need this anymore

src/bun.js/webcore/response.zig
```diff
@@ -2291,6 +2324,15 @@ pub const Fetch = struct {
 
                         if (options.get(ctx, "tls")) |tls| {
                             if (!tls.isEmptyOrUndefinedOrNull() and tls.isObject()) {
+                                if (ssl_config) |conf| {
+                                    conf.deinit();
```
cirospaciari: is there anycase that this would not be null?

src/bun.js/webcore/response.zig
```diff
@@ -2015,14 +2028,22 @@ pub const Fetch = struct {
 
                         if (options.get(ctx, "tls")) |tls| {
                             if (!tls.isEmptyOrUndefinedOrNull() and tls.isObject()) {
+                                if (SSLConfig.inJS(globalThis, tls, exception)) |config| {
+                                    if (ssl_config) |existing_conf| {
```
cirospaciari, line 2089: is there anycase that this would not be null?

src/bun.js/api/server.zig
```diff
@@ -422,6 +491,7 @@ pub const ServerConfig = struct {
                         result.key = native_array;
                         result.key_count = 1;
                         any = true;
+                        result.requires_custom_request_ctx = true;
```
cirospaciari, line 494: is there a case that `any == false` and `requires_custom_request_ctx == true`? dont looks like `requires_custom_request_ctx` will be needed because will return null when any is false

cirospaciari (changes requested):
a few comments, we should also add more tests using other SSLConfig options, and checking the socket activity in the context before deinit/free

src/bun.js/api/server.zig
LudvigHz: No but there are the cases where `any = true` and `requires_custom_context = false`. So it's still needed to determine if the request requires a custom ctx?

I might have misunderstood you question though 😅

src/bun.js/webcore/response.zig
LudvigHz: nope, you're right. I'll remove this 👍

src/http.zig
LudvigHz: hmm, true. Although disabling keep-alive will probably hurt performance substantially. I thought about `deinit`ing the socket context but that would also close any other open sockets and I assume (I'm guessing here) it would abruptly close any open requests on those sockets, although it would remove the UAF. Not sure what the best option is here while maintaining perf.

I also can't find anywhere `deinit` is actually called, which is probably also an issue

src/bun.js/webcore/response.zig
LudvigHz: Nope you're right. I'll remove this

src/js/node/http.ts
```diff
@@ -1722,7 +1769,6 @@ class ClientRequest extends OutgoingMessage {
     this.#reusedSocket = false;
     this.#host = host;
     this.#protocol = protocol;
-    this.#tls = options.tls;
```
LudvigHz, line 1726: @cirospaciari I removed this since I assumed it to be a mistake. I see that you added this option in 74d91f6b. I couldn't see any references to it in the docs, but I saw that some tests where relying on it (which I also changed to match the node API). However if we aim for Node.js compatibility, `tls` is not an option that exists, and options like `rejectUnauthorized` are expected to live directly on the `Options` object.

cirospaciari: Yeah, you are right this is a mistake, we should follow the docs for Node.js compatibility and pass to fetch/Bun.serve through our internal `tls` option.

src/http.zig
LudvigHz: I've disabled keepalive in the client when using custom sockets. I don't have a huge amount of spare time to work on this so I'll focus on getting this working atm

cirospaciari (approved):

cirospaciari:
Can you revert the deps changes? now LGTM

LudvigHz:
> @LudvigHz Can you revert the deps changes? now LGTM

whoops, see I messed up the rebase a bit. I fixed it now :)

robobun:
The mkcert leaf certificate added here (test/js/node/http/fixtures/openssl_localhost.crt) expired on 2026-08-25 and broke test-http-should-accept-custom-certs-when-provided.ts on main. Renewed in #40467.

PR #11332, opened 2024-05-24
Replace Streams.Readable with a JS implementation

gvilums:
### What does this PR do?
Previously, Streams.Readable and its corresponding event emitter were implemented in native code, which has been the source of many bugs.

These changes switch it over to use a JS implementation, based on the `nodejs/readable-stream` package.

Existing tests.


paperdave:
ive tried this multiple times before and got extremely weird test failures (not strictly related to streams). i really hope it is no longer going to cause that, or that we can figure out what causes them.

glad to see the native streams code gone.

test/js/node/crypto/crypto.test.ts
```diff
@@ -261,14 +261,16 @@ it("should send cipher events in the right order", async () => {
...
-    // `[ "cipher", "readable" ]`,
     `[ "cipher", "prefinish" ]`,
+    `[ "cipher", "readable" ]`,
     `[ "cipher", "data" ]`,
     `[ 1, "dfb6b7e029be3ad6b090349ed75931f28f991b52ca9a89f5bf6f82fa1c87aa2d624bd77701dcddfcceaf3add7d66ce06ced17aebca4cb35feffc4b8b9008b3c4"`,
     `]`,
-    // `[ "decipher", "readable" ]`,
     `[ "decipher", "prefinish" ]`,
+    `[ "decipher", "readable" ]`,
     `[ "decipher", "data" ]`,
```
nektro, line 274: there should be a comment this was moved

test/js/node/crypto/crypto.test.ts
```diff
@@ -261,14 +261,16 @@ it("should send cipher events in the right order", async () => {
   const err = await new Response(stderr).text();
   expect(err).toBeEmpty();
   const out = await new Response(stdout).text();
+  // TODO: prefinish and readable (on both cipher and decipher) should be flipped
+  // This seems like a bug in our crypto code, which
   expect(out.split("\n")).toEqual([
-    // `[ "cipher", "readable" ]`,
     `[ "cipher", "prefinish" ]`,
+    `[ "cipher", "readable" ]`,
     `[ "cipher", "data" ]`,
```
nektro, line 269: here too

src/js/node/http.ts
```diff
@@ -1,7 +1,7 @@
 // Hardcoded module "node:http"
```
nektro, line 1: does this fix the test in `test/js/node/http/node-http.test.ts` too?

PR #11337, opened 2024-05-25
Speculative crash fix

Jarred-Sumner:
### What does this PR do?

Speculative crash fix

Fixes BUN-GY


paperclover (approved):
very speculative

Issue #5692, opened 2023-09-18
GET request with body throws wrong Error type


jimmywarting:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

```js
fetch('http://localhost', { body: '123' })
  .catch(err => {
    console.log(err instanceof TypeError)
  })
```

### What is the expected behavior?

error should be of type `TypeError`

### What do you see instead?

the constructor is Error, and the example code log: false

PR #11341, opened 2024-05-25
bindings: fix createTypeError and createRangeError inheritance

nektro:
previously these were returning an instance of `Error` with the `.name` field changed rather than actual `TypeError` and `RangeError` instances.

https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-native-error-types-used-in-this-standard-typeerror
https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-native-error-types-used-in-this-standard-rangeerror

Closes https://github.com/oven-sh/bun/issues/5692


Jarred-Sumner:
Where are the tests?

PR #11355, opened 2024-05-25
fix(windows spawn): use cmd to call batch scripts

### LINKED ISSUE #11328: [Bun Shell]: Error calling cmd/bat scripts when both args and path contain spaces

1.1.9+bb13798d9

Microsoft Windows NT 10.0.22635.0 x64

run
```ts
import { $ } from "bun";
const fields = "name scripts"
await $`npm pkg get ${fields}`;
```

I expect that code to show the package name and scripts like this code does:
```ts
import { $ } from "bun";
const fields = ["name", "scripts"];
await $`npm pkg get ${fields}`;
```

### What do you see instead?

```
ShellError: Failed with exit code 1
 info: {
  "exitCode": 1,
  "stderr": "'C:\\Program' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n",
  "stdout": ""
}
```

### Additional information

related: https://github.com/nodejs/node/issues/7367

possible fixes:
https://github.com/SunsetTechuila/bun/commit/d963370908aa5f1fb9be92f64b91c1922b647c98
https://github.com/SunsetTechuila/bun/commit/1c05351037334a33336c6fa96a62508ef7b72a15

wrapping path in quotes doesn't seem to work

SunsetTechuila:
### What does this PR do?

fixes #11328

- [x] Code changes

I wrote automated tests 

based on https://github.com/nodejs/node/issues/7367#issuecomment-1277495127

not 100% sure about this, so count on your expertise

### DIFF
--- a/test/js/bun/shell/bunshell.test.ts
+++ b/test/js/bun/shell/bunshell.test.ts
@@ -833,6 +833,11 @@ ${temp_dir}`
       .stdout("complex > command; $(execute)\n")
       .runAsTest("complex_mixed_special_chars");
   });
+
+  test("spaces in path and args", async () => {
+    const fields = "name scripts";
+    expect(async () => { await $`npm pkg get ${fields}` }).not.toThrow();
+  });
 });
 
 describe("deno_task", () => {
--- a/src/bun.js/api/bun/process.zig
+++ b/src/bun.js/api/bun/process.zig
@@ -1521,7 +1521,7 @@ pub fn spawnProcessWindows(
         const args_slice = std.mem.span(args);
         args = try allocator.allocSentinel(?[*:0]const u8, args_slice.len + 2, null);
         args[0..2].* = .{ file, "/c" };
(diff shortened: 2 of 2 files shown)


test/js/bun/shell/bunshell.test.ts
```diff
@@ -833,6 +833,11 @@ ${temp_dir}`
       .stdout("complex > command; $(execute)\n")
       .runAsTest("complex_mixed_special_chars");
   });
+
+  test("spaces in path and args", async () => {
+    const fields = "name scripts";
+    expect(async () => { await $`npm pkg get ${fields}` }).not.toThrow();
```
paperclover, line 839: note that this test case does not reproduce the original issue, since `bun install` does not use `.cmd` files.

can you create a cmd script using `tempDirWithFiles` and running the a script with that directory in `$PATH`

cirospaciari:
Closing due to bitrot

PR #11363, opened 2024-05-26
Removed ambiguous quotes in utils.md

Ptitet:
### What does this PR do?

image

The double quotes are ambiguous and redundant imo: it’s not clear if they’re included in the return value of the `escapeHTML` function or if they are only used for delimiting the return value.
Additionally, the inline codeblock is already used for delimiting the return value.

Because of this, I removed the double quotes `"`.


Jarred-Sumner:
thanks

Issue #9998, opened 2024-04-05
Vite with vite-plugin-checker plugin don't exit the process after finishing bun run build


adrifer:
### What version of Bun is running?

1.1.1+ca1dbb4eb

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Repro here: https://github.com/adrifer/BunBuildRepro

These are the steps to follow on that repo:
```
bun install
bun run build
```

### What is the expected behavior?

`bun run build ` should finish when everything is complete

### What do you see instead?

Most of the times, after succeeding one of the two build steps, it gets stuck and nothing happen. I think the process doesn't exit. This works fine with yarn and npm.
![image](https://github.com/oven-sh/bun/assets/812246/1cc69061-e4a8-441f-adb6-8f05de16bb3b)
It gets stuck and process doesn't exit after finishing the build step. 

### Additional information

When using the Dockerfile to build the image I can repro it every time.

adrifer:
I've just tested with bun v1.1.2 and the issue is still there.

adrifer:
After some more debugging I found the issue is related to this vite pluging: vite-plugin-checker

Basically, when adding the "checker" plugin to vite.config I can repro it:
```
import { defineConfig } from "vite";
import react from "@vitejs/plugin-react-swc";
import checker from "vite-plugin-checker";

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    react(),
    checker({
      typescript: true,
    })
  ],
});
```
Removing the checker line in the vite.config plugins, solves the issue.

I've updated my repro to simplify it and only add this plugin to a clean `bun create vite`: https://github.com/adrifer/BunBuildRepro

alexgleason:
I am still experiencing this problem on the latest version of bun (v1.2.5) with vite-plugin-checker (v0.8.0, v0.9.0)

PR #11376, opened 2024-05-26
Attempt to fix unresolved terminate() promise

### LINKED ISSUE #11335: Process is stuck after attempting to terminate already-exited worker

1.1.9-debug+bb13798d9

While trying a few things in debug-build of bun I stumbled upon this bug, which is likely related to #9998 (the reason I got into debugging in the first place)

setTimeout can be arbitrarily large number - it only need to wait for worker to exit before we can call `terminate()`

```js
import { Worker, isMainThread } from "node:worker_threads"

if (isMainThread) {
  const worker = new Worker(__filename);

  setTimeout(() => worker.terminate(), 60);
} else {
  // worker does not need to do anything
}
```

Process does not hang and exits correctly

### What do you see instead?

Process hangs

### Additional information

I could not build bun in WSL (while using ManjaroWSL bun-debug crashes every single time with SIGABRT), so I was able to test Windows version only, but it is likely to reproduce on Linux too

Debug logs: https://gist.github.com/zawodskoj/d6c2569db71ac23e131502266d8e6769

zawodskoj:
### What does this PR do?

Worker.terminate() on JS-side expects "close" event to happen in order to complete promise, but this event will never happen, if worker is already terminated. The "solution" is to pass "already exited" flag to JS-side, so it can resolve the promise without any events.

More work is still required - there's some hacks to allow `boolean` to pass through incompatible Worker API

This PR may fix #11335 and #9998

- [x] Code changes

None of the automated tests are provided, but the issues are now fixed and there are no big changes - just passing `bool` around


zawodskoj:
The best error of all times

image

zawodskoj:
"test timeouts when expected" breaks now (does not look to be breaking on last commit in `main`). Looks like timeouts working, but console.log with bad message is still executed somehow

```
bun test v1.1.10-canary.1 (a0a71c22)

test-timeout.test.js:
error: Test "test timeouts when expected" timed out after 10ms
(fail) test timeouts when expected [68.13ms]
error: Test "test timeouts when expected 2" timed out after 10ms
(fail) test timeouts when expected 2 [66.73ms]
unreachable code
error: Test "process doesn't hang on test with ref'd value" timed out after 10ms
(fail) process doesn't hang on test with ref'd value [46.74ms]

 0 pass
 3 fail
Ran 3 tests across 1 files. [193.00ms]
```

PR #11379, opened 2024-05-26
Run deinit() on CallbackJob regardless of result type

### LINKED ISSUE #9998: Vite with vite-plugin-checker plugin don't exit the process after finishing bun run build

### LINKED ISSUE #11377: CallbackJob prevents VM to exit when result of callback is an unresolved promise

1.1.10+5102a9443

Any platform, reproduces everywhere

This simple code will result in frozen process

```js
setTimeout(() => new Promise(() => 0), 0);
```

Process exits immediately after setTimeout (which aligns with Node behavior)

### What do you see instead?

Process is stuck. For whatever reason if CallbackJob sees promise, it will subscribe and wait until it succeeds, and this is not always the case.

https://github.com/oven-sh/bun/blob/af4e844b621e8c889d2d73c2cd12a7426516ebbf/src/bun.js/api/BunObject.zig#L3811

Potential solution is to deinit CallbackJob regardless of promise status

### Additional information

This bug affects Worker.terminate() - terminate() returns promise which will never complete if Worker is already terminated. #9998 is affected by this bug.

The fact that Worker.terminate() is not fulfilling promise is separate issue. Node fulfills promise every time, but with undefined status code in this case. In attempt to fix original issue's problem I fixed that in #11376

There's also (absolutely irrelevant to this issue, but still notable) difference in status codes - Worker.terminate() yields status code 0 when thread is actually terminated, whereas Node returns 1

zawodskoj:
### What does this PR do?

This PR removes matching of callback result for Promises and forces CallbackJob to deinit regardless

- [x] Code changes

Fixes #11377
Fixes #9998

probably some more

The only thing I could verify is repro from #11377 - this now works correctly. No tests were added (I don't know how to write proper tests for this change), but existing test suite seems fine (except some failed tests because of Windows)


Jarred-Sumner:
Thank you.

I think this is probably correct. I think we manually called `.then` on the Promise before because it predated Bun being able to catch unhandled promise rejections correctly

zawodskoj:
Are these failing tests relevant to this PR?

Jarred-Sumner:
Our CI isn't green right now but we are working on fixing that

zawodskoj:
Ok, just wanted to make sure, thanks for quick reply!

PR #11387, opened 2024-05-27
fix(install): semver tag parsing bugfix

dylan-conway:
### What does this PR do?
When deciding how long a pre/build tag is, we were stopping at a space character or the end of the input. Instead, it should stop at the first invalid character or input end.

For example, the pre tag previously for `^3.0.0-next.0||^3.0.0` would be `-next.0||^3.0.0`. Instead it should be `-next.0`.

Removed two tests from `bun-install.test.ts` that used `_` in pre/build tags. `_` is not valid in loose or strict mode.

added a test


Jarred-Sumner:
insane that `||^3.` is a valid pre tag

test/cli/install/bun-install.test.ts
```diff
@@ -2027,92 +2027,6 @@ it("should handle ^0.0.2rc1 in dependencies", async () => {
...
-  expect(await readdirSorted(join(package_dir, "node_modules"))).toEqual([".cache", "bar"]);
-  expect(await readdirSorted(join(package_dir, "node_modules", "bar"))).toEqual(["package.json"]);
-  expect(await file(join(package_dir, "node_modules", "bar", "package.json")).json()).toEqual({
-    name: "bar",
-    version: "0.0.2",
-  });
-  await access(join(package_dir, "bun.lockb"));
-});
-
-it("should handle ^0.0.2b_4+cafe_b0ba in dependencies", async () => {
```
Jarred-Sumner, line 2073: This looks like it was right before?

image

image

https://jubianchi.github.io/semver-check/#/%3E%3D0.0.2%20%3C0.0.3/0.0.2b_4%2Bcafe_b0ba

Jarred-Sumner: oh i see, it's range versus version

PR #11398, opened 2024-05-27
Make `actual` in `expect(actual)` required

dhoulb:
Changed `expect(actual?: T)` to `expect(actual: T)`

1. Allowing `actual` to be omitted in a test is likely to be a mistake

2. Creates strange behaviour with inference for functions that return `undefined`, for example...

```js
function convertToNumberOrUndefined(num: unknown): number | undefined {
  return typeof num === "number" ? num : undefined;
}

expect(convertToNumberOrUndefined("notnumber")).toBe(undefined); // Type error in `toBe()` because `T` is inferred as `number` not `number | undefined`
expect<number | undefined>(convertToNumberOrUndefined("notnumber")).toBe(undefined); // This works but typing `expect()` manually is annoying.
```

### What does this PR do?

- [ ] Code changes

### DIFF
--- a/packages/bun-types/test.d.ts
+++ b/packages/bun-types/test.d.ts
@@ -490,7 +490,7 @@ declare module "bun:test" {
 
   export interface Expect extends AsymmetricMatchers {
     // the `expect()` callable signature
-    <T = unknown>(actual?: T): Matchers<T>;
+    <T = unknown>(actual: T): Matchers<T>;
 
     /**
      * Access to negated asymmetric matchers.


tobycm:
You should tick the `Documentation or TypeScript types` box in your comment

dhoulb:
> You should tick the `Documentation or TypeScript types` box in your comment

Ah, thanks!

gvilums:
Hey, thanks for the contribution! There are cases where the value is actually not required, mostly when writing code like this that asserts that a statement should be unreachable:
```
expect().fail("should be unreachable");
```
Changing this now would kind of be a breaking change. I'll close the PR for now, but feel free to reopen if you feel there is a need for discussion.

Issue #7826, opened 2023-12-25
make the remaining and total size of `bun upgrade` more human-readable?


huseeiin:
### What is the problem this feature would solve?

![image](https://github.com/oven-sh/bun/assets/122984423/f6f3c0d1-7e29-45e1-96a7-aa5971a95213)

### What is the feature you are proposing to solve the problem?

like `15.7 MB/34.8 MB` instead of what's showing in the picture

ShatilKhan:
@Electroid can you tell me which files I can look into for this edit?

Jarred-Sumner:
Sorry, this was incorrectly marked as "good first issue". To do this, we need to change the ProgressBar implementation in Zig and that is non-trivial since it means we'll need to copy paste

vitalspace:
> @Electroid can you tell me which files I can look into for this edit?

upgrade_command.zig

```bash
var progress = refresher.start("Downloading", version.size);
```

birkskyum:
closed by #11402

paperdave:
landed in #11402

paperdave:
> any info on why it was reverted btw

the way it was editing std.Progress' text label was not thread-safe on windows and caused segfaults when upgrading sometimes.

since that approach doesn't really work, and std.Progress is kind of limited, it might be better to make a custom progress bar (this way it can also be an actual progress bar and not text)

Chubercik:
After seeing [this Tweet](https://x.com/mmmmaharshi/status/1882364647917096995) I came here to file an issue asap, only to find out this was already discussed a year ago. Nevertheless, here's an infodump for whoever attempts to implement this feature:

- bun uses [a snapshot of the Zig `std.Progress`](https://github.com/oven-sh/bun/blob/bun-v1.2.1/src/Progress.zig) from 0.12 for its progress "bar"
- progress is refreshed here:

https://github.com/oven-sh/bun/blob/ce532901ceb43e94f9aa0f5b19a19b8b8b404ae1/src/Progress.zig#L335

- in the case of `bun upgrade` the relevant code can be found here:

https://github.com/oven-sh/bun/blob/ce532901ceb43e94f9aa0f5b19a19b8b8b404ae1/src/cli/upgrade_command.zig#L508

---

Possible solutions:

1. Refactor `src/cli/upgrade_command.zig` - not the best idea, because the need for human-readable sizes exists in other places too, and might show up again in the future,
[...]

PR #11402, opened 2024-05-27
feat: :art: Show better `bun upgrade` progress

### LINKED ISSUE #10214: Show better bun upgrade progress

bun upgrade progress make it more readable

Before:

```bash
Bun 1.1.3 is out! You're on 1.0.33
Downloading [3590209/17497656]
```

After:

```bash
Bun 1.1.3 is out! You're on 1.0.33
Downloading [3.4M/16.6M]
```

image

_No response_

tobycm:
Closes #10214, #7826

### What does this PR do?

This PR formats the download progress

- [x] Code changes

I compiled and ran the command on my laptop with no errors and the output formats mostly properly. I still need help with formatting the progress current value to 2 decimal places instead of 0.

This is my first PR, please tell me if I am missing anything. Thank you

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -1506,7 +1506,7 @@ redirect_type: FetchRedirect = FetchRedirect.follow,
 redirect: []u8 = &.{},
 timeout: usize = 0,
 progress_node: ?*std.Progress.Node = null,
-progress_string_buf: [512]u8 = undefined,
+progress_string_buf: []u8 = undefined,
 downloaded_bytes: u64 = 0,
 estimated_content_length: u64 = 0,
 
@@ -2714,6 +2714,15 @@ pub fn onData(this: *HTTPClient, comptime is_ssl: bool, incoming_data: []const u
             // we save the successful parsed response
             this.state.pending_response = response;
 
+            // For requests that have progress nodes, check for content-length header to provide better download estimates.
+            if (this.progress_node != null) {
+                if (response.getHeader("content-length")) |length_str| {
+                    if (std.fmt.parseInt(u64, length_str, 10)) |estimated_length| {
+                        this.estimated_content_length = estimated_length;
+                    } else |_| {}
+                }
+            }
+
             const body_buf = to_read[@min(@as(usize, @intCast(response.bytes_read)), to_read.len)..];
             // handle the case where we have a 100 Continue
             if (response.status_code == 100) {
@@ -3138,9 +3147,17 @@ pub fn toResult(this: *HTTPClient) HTTPClientResult {
 }
 
 fn updateProgress(this: *HTTPClient) void {
+    // Zig 0.12's progress API does not support floating point or percentage progresses,
+    // only integers. To get around this, the items' name is adjusted with the desired
+    // format message.
     if (this.progress_node) |progress| {
         progress.activate();
-        progress.setName(std.fmt.bufPrint(&this.progress_string_buf, "Downloading [{s:.2}/{s:.2}]", .{ bun.fmt.size(this.downloaded_bytes), bun.fmt.size(this.estimated_content_length) }) catch bun.outOfMemory());
+        progress.setName(
+            std.fmt.bufPrint(this.progress_string_buf, "Downloading [{s:.2} / {s:.2}]", .{
+                bun.fmt.size(this.downloaded_bytes),
+                bun.fmt.size(this.estimated_content_length),
+            }) catch "Downloading",
+        );
         progress.context.maybeRefresh();
     }
 }
@@ -3171,7 +3188,7 @@ fn handleResponseBodyFromSinglePacket(this: *HTTPClient, incoming_data: []const 
         this.downloaded_bytes += incoming_data.len;
 
         if (this.progress_node != null)
-            updateProgress(this);
+            this.updateProgress();
     }
     // we can ignore the body data in redirects
     if (this.state.is_redirect_pending) return;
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
These are flaky test failures, CI isn't currently green

phatdev12 (commented):
image
What is the effect of setUnit MB? :D?

src/http.zig
```diff
@@ -1506,6 +1506,7 @@ redirect_type: FetchRedirect = FetchRedirect.follow,
 redirect: []u8 = &.{},
 timeout: usize = 0,
 progress_node: ?*std.Progress.Node = null,
+progress_formatter: ?*const fn (size: usize) usize = null,
```
paperclover: is using a function pointer for this necessary? more specifically, do we actually use progress_node anywhere else where we want a different formatter.

i think we would want all uses of this to be formatted nicely. i think it would be Ok to always format to MB if there is a progress_node set, and remove this function pointer.

paperclover (changes requested):
thank you for this; i like this a lot better than the status quo. but as you said, it would be nice to make this have at least one or two decimal precisions.

A recording of the PR build for reference:
https://github.com/oven-sh/bun/assets/24465214/d84fc342-b88d-4598-927f-fb6cd82e8632

Since `std.Progress.Node` only contains a `[]const u8` name and integer numbers for the number of items.

one route to achieving this is to use `std.Progress.Node.setName` to update the Node's display text, doing formatting ourself. Keep in mind that the display text needs some memory location to live in, i would say for this case placing it on the stack next to the progress would make sense

```diff
 var refresher = std.Progress{};
-var progress = refresher.start("Downloading", version.size);
+var progress = refresher.start("Downloading", 0);
+var progress_text_buf: [512]u8 = undefined;
...
+async_http.client.progress_text_buf = &progress_text_buf;
```

A pointer to this buffer can be given to the http file and it can use `std.fmt.bufPrint` to format the text `Downloading [11.2MB / 32.5MB]` for example.

Another, more overkill, approach could be to just overhaul the progress code with an implementation that had a nice visual progress bar, but I don't think that is necessary.

src/http.zig
tobycm: I think I will remove this and always format to a larger unit as you suggested

src/http.zig
```diff
@@ -3158,7 +3161,8 @@ fn handleResponseBodyFromSinglePacket(this: *HTTPClient, incoming_data: []const
     defer {
         if (this.progress_node) |progress| {
             progress.activate();
-            progress.setCompletedItems(incoming_data.len);
+            this.downloaded_bytes += incoming_data.len;
+            progress.setName(std.fmt.allocPrint(std.heap.page_allocator, "Downloading [{s:.2}/{s:.2}]", .{ bun.fmt.size(this.downloaded_bytes), bun.fmt.size(this.estimated_content_length) }) catch bun.outOfMemory());
```
paperclover: this memory is never freed. an allocation is not really needed for this use case, and it definitely should not re-allocate the message per frame. consider using a stack-owned buffer as described in [this comment](https://github.com/oven-sh/bun/pull/11402#pullrequestreview-2083549656)

paperclover (changes requested):

src/http.zig
```diff
@@ -3156,11 +3168,10 @@ fn handleResponseBodyFromSinglePacket(this: *HTTPClient, incoming_data: []const
...
     defer {
-        if (this.progress_node) |progress| {
-            progress.activate();
-            progress.setCompletedItems(incoming_data.len);
-            progress.context.maybeRefresh();
-        }
+        this.downloaded_bytes += incoming_data.len;
+
+        if (this.progress_node != null)
+            updateProgress(this);
```
paperclover:
```suggestion
            this.updateProgress();
```

tip: method call syntax is equivalent of a function who's first argument is the base

src/http.zig
```diff
@@ -3133,6 +3137,14 @@ pub fn toResult(this: *HTTPClient) HTTPClientResult {
     };
 }
 
+fn updateProgress(this: *HTTPClient) void {
+    if (this.progress_node) |progress| {
+        progress.activate();
+        progress.setName(std.fmt.bufPrint(&this.progress_string_buf, "Downloading [{s:.2}/{s:.2}]", .{ bun.fmt.size(this.downloaded_bytes), bun.fmt.size(this.estimated_content_length) }) catch bun.outOfMemory());
```
paperclover: this function failing (the `catch ...`) is not actually an out of memory, but actually that the formatting requires more than the 512 bytes allotted. we can handle this more smoothly with `catch "Downloading"` to do *something* besides crashing, since that is possible.

paperclover (changes requested):
looks great. some nitpicks before merging.

paperdave:
In the interest of getting this merged now for Bun 1.1.11, I'm just going to go ahead and finish this for you. Thank you for your work getting it this far.

One major note I have:

You had increased the size of the http client by 512 bytes by storing the buffer on the client. This isn't great because this means every usage of this http client, progress or not, requires that 512 bytes exist (`bun install` uses this, `fetch` uses this, etc). Instead, this is changed to use a pointer, which is 8 bytes of storage instead. In systems programming, keeping data structures small when possible is [extremely ideal](https://vimeo.com/649009599).
```diff
-progress_string_buf: [512]u8 = undefined,
+progress_string_buf: ?*[512]u8 = null,
```

Though i noticed that referring to the stack memory across threads this way is not actually allowed (segfault). So following my instructions as stated would not have been possible. Instead, these bytes are now owned by the global allocator: one allocation of 512 bytes before the download starts, re-used across messages.

I also had to go in and add some extra handling in the http client to account for the estimated size being an *estimate*, and sometimes being zero. You might find the diff interesting there.

paperclover (approved):

src/http.zig
tobycm: right I forgot about that 😅

PR #11419, opened 2024-05-28
rewrite timers for setTimeout, setInterval, setImmediate

Jarred-Sumner:
### What does this PR do?

WIP. Rewrites timers. Makes them 2x - 5x lower overhead. It's faster than Node now (where previously we were slower - https://x.com/jarredsumner/status/1736741811039899871).

Instead of having timers, we use kqueue / epoll's builtin `timeout` argument. We tell it how long to go to sleep for instead of updating a wrapper timer.

More importantly, we can use these timers in more places. For the garbage collector stuff, for jest, for AbortSignal.timeout, maybe we could add `fetch(urll, {timeout: 123456})`. Since it's the same thread we don't need to worry about pointer lifetimes of the timer as much. 

On Windows, from libuv's perspective there's only one timer which is great because we can more easily control whether or not the process should be kept alive. This should fix the various event loop bugs related to setTimeout not keeping the event loop alive (specifically `process.test.ts`)

Plus:
- Fixes the Node compatibility issue where `this` inside of `setTimeout`, `setInterval`, and `setImmediate` is supposed to be a `Timeout` object but instead it's `undefined`.
- Uses `JSImmutableButterfly` to store > 1 arguments, which should be very slightly faster than creating an array
- Fixes a bug where the DNS cache resolution on macOS was too short because we were using the timer wrong
- Makes it so you can call .refresh() on a timer that was already called. Node supports this. We previously did not.

TODOs:
- [x] Fix Linux build
- [x] Fix Windows build
- [x] Fix .duration() on `timespec`
- [x] Fix the bug with `.less()` or `.order()` causing the event loop to go to sleep for too long
- [ ] Verify that setImmedate().refresh() works as expected
- [ ] Make the tests pass

The main TODO remaining here (besides fixing the Linux & Windows build issues) is that calculating the timeout duration is wrong in the event loop.


test/js/web/timers/setTimeout-cpu-fixture.js
```diff
@@ -1 +1,3 @@
-setTimeout(() => {}, 3_000); // we use 3s so we can reproduce better the issue #7790
+setTimeout(() => {
```
Electroid, line 1: Why are these files labeled as generated on Github?

staabm: Because the filename matches the pattern in https://github.com/oven-sh/bun/blob/912bccb6243bbc169c0d92e23dc1be7f01e44fba/.gitattributes#L26

test/js/web/timers/setImmediate.test.js
```diff
@@ -35,7 +35,6 @@ it("clearImmediate", async () => {
   var called = false;
   const id = setImmediate(() => {
     called = true;
-    expect.unreachable();
```
Jarred-Sumner, line 38: what was broken about this? it shouldn't be called right?

gvilums: The expect would fail outside the test. Intent was to change it so that the failure is reported inside the test. I changed it to use `Promise.withResolvers` instead

src/bun.js/javascript.zig
```diff
@@ -2540,8 +2536,6 @@ pub const VirtualMachine = struct {
             });
         }
 
-        this.eventLoop().autoTick();
```
Jarred-Sumner, line 2543: interesting change

Issue #11380, opened 2024-05-26
Brotli support for node:zlib does not respect params (and the default quality setting is unusable for on-the-fly compression)


infrahead:
### What version of Bun is running?

1.1.10+5102a9443

### What platform is your computer?

Tested on both Ubuntu and Windows

### What steps can reproduce the bug?

Using a simple example:
```js
import { createReadStream, createWriteStream } from 'node:fs'
import { createBrotliCompress, brotliCompress, constants as ZLIB } from 'node:zlib'
import { pipeline } from 'node:stream/promises'

const opts = {
	params : {
		[ZLIB.BROTLI_PARAM_MODE] : ZLIB.BROTLI_MODE_TEXT,
		[ZLIB.BROTLI_PARAM_QUALITY] : 4,
	},
}

async function run () {
	await pipeline(
		createReadStream('example.pdf'),
		createBrotliCompress(opts),
		createWriteStream('example.pdf.br'),
	)
	console.timeEnd('pipeline')
}

console.time('pipeline')
run().catch(console.error)
```

### What is the expected behavior?

The params, especially quality (compression level), are applied to compression.

### What do you see instead?

Regardless of params, the size/time ratio is always the same for Bun.

Furthermore, passing any bogus object for the options is accepted, with no error at runtime (Although Bun's TypeScript types are correct, so assuming is stubbed out)

### Additional information

Because `BROTLI_DEFAULT_QUALITY` is 11 (the maximum), this is a serious issue, especially for on-the-fly scenarios where is unusable for me.  E.g. for a 3MB file...

Bun v1.1.10: `[3.83s] pipeline`
v.s.
Node.js v22.1.0: `pipeline: 37.505ms`

(Bun does save slightly smaller file (500KB) to Node's 600KB (where quality level 4 is applied) but that is negligible for **100x slower**!  I've tested without streams and async and is same relative difference.

PR #11429, opened 2024-05-29
node:zlib: allow passing brotli params in encode+decode options

nektro:
Closes https://github.com/oven-sh/bun/issues/11380

```
❯ bun-debug-before index.js 
[771.99ms] pipeline

❯ bun-debug-after index.js
[120.84ms] pipeline
```

```
❯ bun-1.1.10 index.js 
[228.10ms] pipeline

❯ ~/src/bun/build-release/bun index.js 
[7.07ms] pipeline
```

### DIFF
--- a/src/brotli.zig
+++ b/src/brotli.zig
@@ -1,6 +1,6 @@
 const bun = @import("root").bun;
 const std = @import("std");
-const c = struct {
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
can you add a test

Issue #11260, opened 2024-05-22
Bun throws error when copying file to a folder that does not exist


birkskyum:
### What version of Bun is running?

1.1.10-canary.1+cc0bce62e

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

I experience that I can make a clean `basic` example (bun create solid), dump a few files in the public folder, like .glb 3d models, and then the `bun --bun run build` command will break with:

```sh
 ERROR  { [ENOENT: No such file or directory]                                                                                                            9:06:21 AM
  code: 'ENOENT',
  path:
   '/Users/admin/repos/bun-kitchensink/solid/wefwef/public/3dmodels/giantherb/giantherb.glb',
  syscall: 'copyfile',
  errno: -2 } 



 ERROR  No such file or directory                                                                                                                        9:06:21 AM

error: script "build" exited with code 1
```

It's worth noting that it's not a requirement to use these files anywhere in the code for this to break

### What is the expected behavior?

When running `pnpm run build` or npm, then the build completes

### Additional information

Related to
- #3736

posterhoudt:
Confirmed, @birkskyum. I have the same issue (used Bun v1.1.9 on Windows 11):

![image](https://github.com/oven-sh/bun/assets/163450497/71b1778e-7db6-4230-8412-8ea34cba0845)

zoto-ff:
the error occurs when working with nitropack (vinxi/nitro component). something is wrong with fs.promises.cp(). its occurs when performing all copies in parallel (await Promise.all(...)) and in queue (for (...) { await ... })

![IMG_6463](https://github.com/oven-sh/bun/assets/142039751/94348ac0-e239-4d03-9dca-b84b6e24805a)

https://github.com/unjs/nitro/blob/61f2079a605fca81b578894edfdbb0c328db12a3/src/build.ts#L87

sirenkovladd:
Smaller reproduce

```js
import { promises } from 'node:fs';

const src = 'newFile.txt';
const folder = 'folder-not-exist'
const dst = `${folder}/${src}`;

await promises.writeFile(src, 'A'.repeat(131073))

await promises.rm(folder, { recursive: true, force: true })
await promises.cp(src, dst);
console.log('Success')
```

```sh
❯ bun test.js 
ENOENT: No such file or directory
   errno: -2
 syscall: "copyfile"
   path: "newFile.txt"


Bun v1.1.10 (macOS arm64)

[...]

PR #11433, opened 2024-05-29
node:fs/promises.cp should make path if dest does not exist

nektro:
Closes https://github.com/oven-sh/bun/issues/11260

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -6779,30 +6779,30 @@ pub const NodeFS = struct {
         }
 
         if (Environment.isWindows) {
+            const src_enoent_maybe = ret.initErrWithP(.ENOENT, .copyfile, this.osPathIntoSyncErrorBuf(src));
+            const dst_enoent_maybe = ret.initErrWithP(.ENOENT, .copyfile, this.osPathIntoSyncErrorBuf(dest));
             const stat_ = reuse_stat orelse switch (windows.GetFileAttributesW(src)) {
-                windows.INVALID_FILE_ATTRIBUTES => return .{ .err = .{
-                    .errno = @intFromEnum(C.SystemErrno.ENOENT),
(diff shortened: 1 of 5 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -6627,7 +6627,12 @@ pub const NodeFS = struct {
                 mode_ |= C.darwin.COPYFILE_EXCL;
             }
 
-            return ret.errnoSysP(C.copyfile(src, dest, null, mode_), .copyfile, src) orelse ret.success;
+            const first_try = ret.errnoSysP(C.copyfile(src, dest, null, mode_), .copyfile, src) orelse return ret.success;
```
Jarred-Sumner, line 6630: What about linux and Windows?

test/js/node/fs/fs.test.ts
```diff
@@ -3216,3 +3216,18 @@ it("promises.fdatasync with a bad fd should include that in the error thrown", a
...
+  let folder = "folder-not-exist";
+  let dst = join(folder, src);
+
+  src = join(x_dir, src);
+  folder = join(x_dir, folder);
+  dst = join(x_dir, dst);
+
+  await promises.writeFile(src, "A".repeat(131073));
+  await promises.rm(folder, { recursive: true, force: true });
+  await promises.cp(src, dst);
```
Jarred-Sumner, line 3233: Where does this test check that the `dst` is equal to `"A".repeat(131073)`?

src/bun.js/node/node_fs.zig
nektro: linux is handled; will push an update with windows

test/js/node/fs/fs.test.ts
nektro: will add this

Jarred-Sumner:
Thank you

Issue #4687, opened 2023-09-09
Missing git in docker image


ToshY:
### What version of Bun is running?

v1.0.0

### What platform is your computer?

Linux 5.15.90.3-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

Currently there seems to be no `git` in the docker image, and whenever I perform a `bun install` it fails as it's unable to perform a `git clone`:

```txt
docker compose run --rm --no-deps bun bun install
[0.15ms] ".env"
bun install v1.0.0 (822a00c4)
  🚚 unicode-property-aliases-ecmascript [737/897]


error: "git clone" for "webcomponents.js" failed
```

The  `webcomponents.js` is used by `@silvermine/videojs-chromecast` which is in my `package.json`.

Verified that `git` is not in the container:

```shell
$ docker compose run --rm --no-deps bun /bin/sh
$ git
/bin/sh: 1: git: not found
```

---

Background info:

Tried to switch out [timbru31/node-alpine-git](https://hub.docker.com/r/timbru31/node-alpine-git) for [oven/bun](https://hub.docker.com/r/oven/bun) while keeping `package.json` the same. The official node alpine image also gives issues when installing dependencies, which is the reason for originally switching to `timbru31/node-alpine-git`. Was hoping bun included git in its docker image.

### What is the expected behavior?

Here's a working example when adding `git` to it.

`docker-compose.yml`
```yml
version: '3.9'

services:
  bun:
    build:
      context: .docker/node
      dockerfile: Dockerfile
    user: "1000:1000"
    working_dir: /app
    security_opt:
      - no-new-privileges:true
    profiles:
      - dev
    restart: "no"
    volumes:
      - .:/app
```

`Dockerfile`
```Dockerfile
FROM oven/bun:1.0

RUN apt-get update  \
    && apt-get install -y \
      git \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*
```

Run `bun install`.

```shell
$ docker compose run --rm --no-deps --build bun bun install

[+] Building 0.0s (6/6) FINISHED                                                                                                                                             docker:default
 => [bun internal] load build definition from Dockerfile                                                                                                                               0.0s
 => => transferring dockerfile: 179B                                                                                                                                                   0.0s
 => [bun internal] load .dockerignore                                                                                                                                                  0.0s
 => => transferring context: 2B                                                                                                                                                        0.0s
[...]

PR #11435, opened 2024-05-29
docker: add git to the images

nektro:
Closes https://github.com/oven-sh/bun/issues/4687

### DIFF
--- a/dockerhub/debian-slim/Dockerfile
+++ b/dockerhub/debian-slim/Dockerfile
@@ -80,6 +80,7 @@ RUN groupadd bun \
     && apt update \
     && apt-get install -qq --no-install-recommends \
       git \
+      python3 \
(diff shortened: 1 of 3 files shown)


dockerhub/debian-slim/Dockerfile
```diff
@@ -77,6 +77,8 @@ RUN groupadd bun \
       --shell /bin/sh \
       --create-home \
     && ln -s /usr/local/bin/bun /usr/local/bin/bunx \
+    && apt-get install -qq --no-install-recommends \
```
Jarred-Sumner, line 81: did you manually test this works? i remember needing to do `apt-get update` in certain cases

nektro: did now

PR #11455, opened 2024-05-30
fix high cpu usage when using bun.build

### LINKED ISSUE #10967: Bun.build: Very high ongoing CPU usage on Windows, despite completing successfully

1.1.7

Command line:

```sh
bun run --watch ./source/server.ts
```

`source/server.ts`:

```ts
import Path from 'node:path';

  entrypoints: [Path.resolve('./source/client.ts')],
});

console.debug(`result.success:`, result.success);
```

`source/client.ts`:

```ts
console.log('Hello, world');
```

In watch mode, running the above results in a constant 13% CPU usage figure until the process is terminated. The problem goes away when I comment out the call to `Bun.build`, suggesting the bundler is the issue, not watch mode. But this is just speculation on my part.

[Here is a zip file](https://github.com/oven-sh/bun/files/15269741/server.zip) containing the project setup.

CPU usage should be negligible, as the process shouldn't be doing any real work other than waiting for a change to the file to trigger a restart.

### What do you see instead?

image

### Additional information

_No response_

paperclover:
### What does this PR do?

Calling `bun --watch` on a file that calls bun.build will pin the bundler's thread to max.

~~This PR calls .inc() in that thread, but that doesnt fix it on it's own. To fix it, we need to make libuv actually sleep. This is done through a shared timer instead of mutating libuv's internal counters. doing this before caused subtle bugs like this.~~ Doing this caused more problems than it solved. I am reducing the scope of this PR to only fix the bundler issues.

Result:

image

Fixes #10967

- [ ] Code changes

Code for measuring resource usage does not work on windows. It is not trivial to write a test for this.

### DIFF
--- a/src/bun.js/api/Timer.zig
+++ b/src/bun.js/api/Timer.zig
@@ -30,8 +30,8 @@ pub const All = struct {
         .context = {},
     },
     active_timer_count: i32 = 0,
-    uv_timer: if (Environment.isWindows) uv.uv_timer_t else void =
-        if (Environment.isWindows) std.mem.zeroes(uv.uv_timer_t) else {},
+    uv_timer: if (Environment.isWindows) uv.Timer else void =
+        if (Environment.isWindows) std.mem.zeroes(uv.Timer) else {},
 
     // We split up the map here to avoid storing an extra "repeat" boolean
     maps: struct {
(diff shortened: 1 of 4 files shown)


paperdave:
All of the SIGKILL failures are timeouts due to the event loop locking up:

image

PR #11456, opened 2024-05-30
Set CLOEXEC more

Jarred-Sumner:
### What does this PR do?

Set CLOEXEC more

Let's see what tests say

CI


src/bun.js/bindings/bun-spawn.cpp
```diff
@@ -164,10 +164,11 @@ extern "C" ssize_t posix_spawn_bun(
         sigprocmask(SIG_SETMASK, &childmask, 0);
         if (!envp)
             envp = environ;
+        
+        // If CLOSE_RANGE_CLOEXEC is unsupported, we risk leaking some file descriptors
+        // This is better than potentially causing a different thread (like the watcher) to crash.
+        bun_close_range(current_max_fd + 1, ~0U, CLOSE_RANGE_CLOEXEC);
```
gvilums, line 170: Maybe `CLOSE_RANGE_UNSHARE` could be useful here - it unshares the fds before closing them, which should avoid issues with other threads

gvilums: This might even let us skip the `CLOSE_RANGE_CLOEXEC` entirely

gvilums: And also `CLOSE_RANGE_UNSHARE` is supported in all kernels

PR #11470, opened 2024-05-30
feat: `bun patch`

### LINKED ISSUE #2336: support patching dependencies, similar to `pnpm patch`/`patch-package`

I am currently working with a few dependencies that will cause errors at runtime due to their imports, and I'd also like to apply general changes without suggesting changes to the projects.

Today I'm using `pnpm` to manage dependencies and `bun` to run the project, which is fine for the most part. Unfortunately `patch-package` will not work as it is looking for npm & yarn lockfiles.

`bun patch`

- behaves similar to [`pnpm patch`](https://pnpm.io/cli/patch)
- executing `bun patch <dependency-name>` allows the developer to modify dependencies
- executing `bun patch --apply <dependency-name>` applies the patch
	- pnpm creates a copy of the dependency into `/private/var/folders` and requires the developer to pass the full directory path when applying a patch with `pnpm patch-commit <directory-path>`

_No response_

zackradisic:
### What does this PR do?

Fixes #2336 

WIP but opening this up now for reviewal because it somehow grew to 5k changes

This adds the following:
- implements a `.patch` file parser and executor
- modifies the package manager to read `patchedDependencies` from package.json and apply them to packages
- adds the following commands:
  - `bun patch <pkg>`: copy a package's contents to a temp folder so changes can be made to it
  - `bun patch-commit <pkg-temp-folder>`: generate the .patch file from the temp folder with changes

Notes:
- `src/patch.zig` contains the logic for parsing and applying patch files
- patch files are created using git diff command, similar to how pnpm does it
- `bun patch <pkg>` needs to use the same clonefile/copyfile logic from `PackageInstall`, so I turned that into a comptime function so it could be used during regular install and for the command too, moving this around caused a bunch of additions
- Patched packages are stored in the cache with a folder name like `${cache_dir_subpath}_patch_hash=${hash_of_patchfile_contents}` (e.g. `is-even@1.0.0_patch_hash=aab1923f`)
  - this is so the install logic doesn't have to be changed too much
  - this also means I had to change all the functions which get the cache folder of a package (e.g. `cachedNPMPackageFolderName()`) to accept a `patch_hash: ?u64` argument
- Patched packages also need a `.bun-tag-<patch-hash>` file. This allows us to know if a package in node_modules was patched or not

Added tests


src/install/install.zig
```diff
@@ -207,6 +220,488 @@ pub const Aligner = struct {
...
+        apply: ApplyPatch,
+    },
+    task: ThreadPool.Task = .{
+        .callback = runFromThreadPool,
+    },
+    next: ?*PatchTask = null,
+
+    const debug = bun.Output.scoped(.InstallPatch, false);
+
+    const Error = struct {
```
Jarred-Sumner: We can use `Syscall.Error` here instead of creating a new one for patch

src/install/install.zig
```diff
@@ -207,6 +220,488 @@ pub const Aligner = struct {
...
+        return bun.JSC.Node.Maybe(T, Error);
+    }
+
+    const CalcPatchHash = struct {
+        patchfile_path: []const u8,
+        name_and_version_hash: u64,
+
+        state: ?EnqueueAfterState = null,
+
+        result: ?Maybe(u64) = null,
```
Jarred-Sumner: What do these optionals mean? Are you sure we need them to exist?

src/io/io.zig
```diff
@@ -13,6 +13,68 @@ const assert = bun.assert;
 
 pub const Source = @import("./source.zig").Source;
 
+pub fn readFileAt(
+    allocator: std.mem.Allocator,
+    dirfd: bun.FileDescriptor,
+    path: [:0]const u8,
+    max_bytes: usize,
+) bun.sys.Maybe([]u8) {
+    const fd = switch (bun.sys.openat(dirfd, path, std.os.O.RDONLY, 0)) {
```
Jarred-Sumner: Replace these with `bun.sys.File.readFrom`

src/io/io.zig
```diff
@@ -13,6 +13,68 @@ const assert = bun.assert;
...
+        .err => |e| return .{ .err = e.withPath(path) },
+        .result => |fd| fd,
+    };
+    defer _ = bun.sys.close(fd);
+    return readFile(allocator, fd, max_bytes);
+}
+
+pub fn readFile(
```
Jarred-Sumner: ✂️

src/io/io.zig
```diff
@@ -13,6 +13,68 @@ const assert = bun.assert;
...
+    if (size > max_bytes) return .{ .err = bun.sys.Error.fromCode(bun.C.E.@"2BIG", .read).withFd(fd) };
+    const buf = allocator.alloc(u8, size) catch bun.outOfMemory();
+    const read = switch (readIntoBuf(fd, buf)) {
+        .err => |e| return .{ .err = e },
+        .result => |r| r,
+    };
+    return .{ .result = read };
+}
+
+pub fn writeAll(fd: bun.FileDescriptor, buf: []const u8) bun.sys.Maybe(void) {
```
Jarred-Sumner: ✂️

src/io/io.zig
```diff
@@ -13,6 +13,68 @@ const assert = bun.assert;
...
+            .result => |v| v,
+            .err => |e| if (e.getErrno() == bun.C.E.INTR) 0 else return .{ .err = e },
+        };
+    }
+    return bun.sys.Maybe(void).success;
+}
+
+/// Read all data from `fd` into `buf` until `buf` is filled or EOF is returned.
+///
+pub fn readIntoBuf(fd: bun.FileDescriptor, buf: anytype) bun.sys.Maybe([]u8) {
```
Jarred-Sumner: ✂️

src/shell/interpreter.zig
```diff
@@ -6814,16 +6814,22 @@ pub const Interpreter = struct {
             pub fn start(this: *Echo) Maybe(void) {
                 const args = this.bltn.argsSlice();
 
+                var has_leading_newline: bool = false;
```
Jarred-Sumner, line 6866: What are these shell changes? Should these maybe go in a different PR?

src/string_immutable.zig
```diff
@@ -4037,6 +4037,19 @@ pub fn encodeBytesToHex(destination: []u8, source: []const u8) usize {
     return to_read * 2;
 }
 
+/// Leave a single leading char
+/// ```zig
+/// trimSubsequentLeadingChars("foo\n\n\n\n", '\n') -> "foo\n"
+/// ```
+pub fn trimSubsequentLeadingChars(slice: []const u8, char: u8) []const u8 {
+    var end = slice.len - 1;
```
Jarred-Sumner, line 4046: integer underflow when slice.len is 0
```suggestion
    var end = slice.len -| 1;
```

src/sys.zig
```diff
@@ -194,6 +196,47 @@ pub const Error = struct {
         return copy;
     }
 
+    pub fn fromZigErr(err: anyerror, syscall: Syscall.Tag) Error {
+        return .{
+            .errno = @as(Int, @intCast(@intFromEnum(switch (err) {
```
Jarred-Sumner: ✂️

src/bun.js/bindings/bindings.zig
```diff
@@ -1616,6 +1616,10 @@ pub const SystemError = extern struct {
         return @enumFromInt(this.errno * -1);
     }
 
+    pub fn toAnyhowError(this: SystemError) bun.anyhow.Error {
```
Jarred-Sumner, line 1619: As discussed, let's not do this

src/bun.js/event_loop.zig
```diff
@@ -915,6 +918,10 @@ pub const EventLoop = struct {
         while (@field(this, queue_name).readItem()) |task| {
             defer counter += 1;
             switch (task.tag()) {
+                @field(Task.Tag, typeBaseName(@typeName(AsyncPatchApplyTask))) => {
+                    var shell_ls_task: *AsyncPatchApplyTask = task.get(AsyncPatchApplyTask).?;
```
Jarred-Sumner: ✂️

src/bun.js/module_loader.zig
```diff
@@ -2444,6 +2446,7 @@ pub const ModuleLoader = struct {
                 .@"node:constants" => return jsSyntheticModule(.@"node:constants", specifier),
                 .@"bun:jsc" => return jsSyntheticModule(.@"bun:jsc", specifier),
                 .@"bun:test" => return jsSyntheticModule(.@"bun:test", specifier),
+                .@"bun:patch" => return jsSyntheticModule(.@"bun:patch", specifier),
```
Jarred-Sumner: ✂️

packages/bun-types/patch.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "bun:patch" {
```
Jarred-Sumner: ✂️

src/shell/interpreter.zig
```diff
@@ -6814,16 +6814,22 @@ pub const Interpreter = struct {
...
+                    } else {
+                        if (thearg.len > 0 and thearg[thearg.len - 1] == '\n') {
+                            has_leading_newline = true;
+                        }
+                        this.output.appendSlice(bun.strings.trimSubsequentLeadingChars(thearg, '\n')) catch bun.outOfMemory();
                     }
                 }
 
-                this.output.append('\n') catch bun.outOfMemory();
+                if (!has_leading_newline) this.output.append('\n') catch bun.outOfMemory();
```
Jarred-Sumner, line 6881: I bet this is related to the test failures

src/install/install.zig
```diff
@@ -6569,6 +7589,8 @@ pub const PackageManager = struct {
                     this.positionals = cli.positionals;
                 }
 
+                // if (sub
```
```suggestion

```

src/install/install.zig
```diff
@@ -6587,6 +7609,16 @@ pub const PackageManager = struct {
 
                 this.update.development = cli.development;
                 if (!this.update.development) this.update.optional = cli.optional;
+
+                if (subcommand == .patch) {
+                    // TODO
```
Jarred-Sumner: what is this supposed ot do?

src/install/install.zig
```diff
@@ -9202,6 +10410,488 @@ pub const PackageManager = struct {
...
+        )) {
+            .result => |v| v,
+            .err => |e| {
+                Output.prettyError("<r><red>error<r>: failed to get bun patch tag: {}<r>\n", .{e.toSystemError()});
+                Output.flush();
+                Global.crash();
+            },
+        };
+
+        const package_json_source_str = brk: {
```
Jarred-Sumner: `bun.sys.File.toSource`

src/patch.zig
```diff
@@ -0,0 +1,1483 @@
...
+        this: *PatchLinesParser,
+        file_: []const u8,
+        opts: struct { support_legacy_diffs: bool = false },
+    ) ParseErr!void {
+        const end = brk: {
+            var iter = std.mem.splitBackwardsScalar(u8, file_, '\n');
+            var prev: usize = file_.len;
+            if (iter.next()) |last_line| {
+                if (last_line.len == 0) {
+                    prev = iter.index.?;
```
dylan-conway, line 764:
```suggestion
                    prev = iter.index orelse return;
```
Hit this with an empty patch file. Maybe a check for len == 0 earlier would be better

src/patch.zig
```diff
@@ -0,0 +1,1485 @@
...
+            written += switch (bun.sys.write(file_fd, contents[written..])) {
+                .result => |w| w,
+                .err => |e| return .{ .err = e.withPath(file_path) },
+            };
+        }
+
+        return JSC.Maybe(void).success;
+    }
+
+    fn applyPatch2(patch: *const FilePatch) !void {
```
Jarred-Sumner: what is all this?

src/patch.zig
```diff
@@ -0,0 +1,1485 @@
...
+        _ = last; // autofix
+        for (patch.hunks.items, 0..) |*hunk, i| {
+            _ = hunk; // autofix
+            _ = i; // autofix
+
+            // if (result.) {}
+        }
+    }
+
+    // const Chunk = union(enum) {
```
Jarred-Sumner: please delete dead code

src/patch.zig
```diff
@@ -0,0 +1,1485 @@
...
+
+        return .{ a_part, b_part };
+    }
+
+    fn parseDiffLinePaths(line: []const u8) ?struct { []const u8, []const u8 } {
+        // From @pnpm/patch-package the regex is this:
+        // const match = line.match(/^diff --git a\/(.*?) b\/(.*?)\s*$/)
+
+        const prefix = "diff --git a/";
+        if (!std.mem.startsWith(u8, line, prefix)) return null;
```
Jarred-Sumner: Replace all usages of `std.mem.startsWith` with `strings.hasPrefix`

gruckion:
Thank you

Issue #2428, opened 2023-03-19
Implement `node:cluster` module


Tushar-Chavan14:
### What version of Bun is running?

0.5.7

### What platform is your computer?

Linux 6.2.0-76060200-generic x86_64 x86_64

### What steps can reproduce the bug?

bun i rate-limiter-flexible

import { RateLimiterMemory } from "rate-limiter-flexible";

const rateLimiter = new RateLimiterMemory({
  points: 5,
  duration: 120,
});

Jarred-Sumner:
We haven't implemented the Node.js `cluster` module yet.

ThatOneCalculator:
Is there currently any other way to do multiprocessing/multithreading with bun? This seems like it should be a very high priority item to support.

xb1itz:
+1 on this. The lack of `node:cluster` implementation is probably the main reason we are not switching to Bun right now.

CharnSrinivas:
> +1 on this. The lack of `node:cluster` implementation is probably the main reason we are not switching to Bun right now.
Same here i am shifting back to node since i get know that `node:cluster` is not available yet

melroy89:
I got indeed:

```
error: Could not resolve Node.js builtin: "cluster". To use Node.js builtins, set target to 'node' or 'bun'

const cluster = require('cluster')
```

I was very surprised to see this error. EDIT: I know `cluster` is not yet implemented. But then don't call it bun 1.0 yet.

joshholly:
I ran bun --bun run dev to run my nextJS app. And I got this exact error.

ThatOneCalculator:
Well, `node:cluster` isn't implemented yet, so that's why you're seeing the error.

ElYaiko:
+1

masfahru:
For now (Bun v.1.0.3), I can imitate the behavior of the `node:cluster` in Bun to spawn as many server instances as the number of CPU cores, using this script.

spawn.ts
```typescript
import os from "node:os";

const numCPUs = os.cpus().length;
for (let i = 0; i < numCPUs; i++) {
  Bun.spawn(["bun", "index.ts"], {
    stdio: ["inherit", "inherit", "inherit"],
    env: { ...process.env },
  });
}
```

index.ts
```typescript
Bun.serve({
  port: 8080,
  reusePort: true, // allow Bun server instances to use same port 
  fetch(req: Request) {
    // ...
```

run `bun spawn.ts`

calebeaires:
[...]

Issue #928, opened 2022-07-31
Bun does not throw when call stack is exceeded


max:
### Version

0.1.5

### Platform

Darwin maxbookpro.local 22.0.0 Darwin Kernel Version 22.0.0: Tue Jun 28 20:46:36 PDT 2022; root:xnu-8792.0.134.131.2~1/RELEASE_ARM64_T6000 arm64

### What steps will reproduce the bug?

`bun run` the following code:

```javascript
const foo = () => bar();
const bar = () => foo();
foo();
```

### How often does it reproduce? Is there a required condition?

Every time

### What is the expected behavior?

`bun` should throw similar to `jsc`:

```
Exception: RangeError: Maximum call stack size exceeded.
bar@scratch.js:2:22
foo@scratch.js:1:22
...
```

### What do you see instead?

The call stack grows until `bun` consumes 100% of the CPU.

evanwashere:
call stack isn't actually growing because of tail call optimization, the question is do we want to limit how much you can take advantage of it

pepper-lees:
Test case can be simplified.

```js
const foo = () => foo();
foo(); // does not throw
```

```js
const foo = function () {
  return foo();
}
foo(); // does throw
```

@evanwashere In my opinion, it should be limited, because infinite loops should be avoided.

vrmiguel:
A lot of current JS is being written in a more functional style, making TCO quite favorable

ethanqm:
@JL102 That second example no longer throws, the tail call is being optimised per ES6

booniepepper:
I recommend to resolve this as "no issue." This is working as intended by bun, and also implemented [as specified](https://exploringjs.com/es6/ch_tail-calls.html).

Node.js does not implement TCO (See: https://node.green and [this blog](https://v8.dev/blog/modern-javascript) and [their proposal](https://github.com/tc39/proposal-ptc-syntax)) but there is no rule that bun or other JS runtimes (system, browser, embedded, etc) must follow Node.js instead of the language specification. Straying from the spec makes code less portable and less valuable.

booniepepper:
BTW if anyone legitimately needs a workaround to force a stack overflow, and you _own the code_ then you can simply make your recursive call not in the tail position.

Example with one extra line for a variable bind:

```javascript
const noTco = () => {
  const res = noTco();
  return res;
}
```

Demo with bun-repl:

```
$ bunx bun-repl
Welcome to Bun v1.0.1
Type ".help" for more information.
[...]

PR #11492, opened 2024-05-31
implement `node:cluster`

nektro:
Closes https://github.com/oven-sh/bun/issues/2428


dienstbereit:
> Closes #2428
> 
> close but not quite.

Hey @nektro, are you currently still working on the `node:cluster` implementation? I think many of us are eagerly awaiting the implementation so that they can go into production with their bun project.

nektro:
I am still working on this yes

Zer0xxxx:
> I am still working on this yes

All the best, I'm with you all the way!

almoatamed:
good luck

mavyfaby:
Thank you, @nektro, for implementing `node:cluster` module. This is truly a masterpiece addition to the Bun framework and will definitely improve backend performance by leveraging multiprocessing. Will this feature be included in the next version `1.1.18`?

src/bun.js/api/BunObject.classes.ts
```diff
@@ -91,20 +90,24 @@ export default [
         fn: "doSend",
         length: 1,
       },
-
       kill: {
         fn: "kill",
         length: 1,
       },
+      disconnect: {
```
Jarred-Sumner, line 97: Why was this API added to the Bun object?

src/bun.js/api/BunObject.classes.ts
```diff
@@ -91,20 +90,24 @@ export default [
...
+        fn: "disconnect",
+        length: 0,
+      },
+      connected: {
```
Jarred-Sumner, line 101: Why was this API added to the Bun object?

src/bun.js/api/bun/subprocess.zig
```diff
@@ -717,7 +715,8 @@ pub const Subprocess = struct {
         }
     }
 
-    pub fn doSend(this: *Subprocess, global: *JSC.JSGlobalObject, callFrame: *JSC.CallFrame) JSValue {
+    pub fn doSend(this: *Subprocess, global: *JSC.JSGlobalObject, callFrame: *JSC.CallFrame) callconv(.C) JSValue {
+        Output.scoped(.IPC, false)("Subprocess#doSend\n", .{});
```
Jarred-Sumner: Please don't declare the IPC scope multiple times. It's going to cause weird bugs because it relies on comptime memoization to dedupe itself

Also, `#doSend\n` shouldn't be necessary, it should always add a newline at the end

src/bun.js/api/BunObject.classes.ts
Jarred-Sumner: oh this is subprocess

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2279,13 +2298,25 @@ pub const Subprocess = struct {
     }
 
     pub fn handleIPCClose(this: *Subprocess) void {
-        this.ipc_data = null;
+        Output.scoped(.IPC, false)("Subprocess#handleIPCClose\n", .{});
```
Jarred-Sumner: Ditto. Don't create a new Output.scoped here

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2279,13 +2298,25 @@ pub const Subprocess = struct {
...
         this.updateHasPendingActivity();
+
+        const this_jsvalue = this.this_jsvalue;
+        this_jsvalue.ensureStillAlive();
+        if (this.on_disconnect_callback.trySwap()) |callback| {
```
Jarred-Sumner, line 2328: Do we not need to emit an optional error, like if it was closed due to the process sending invalid data?

src/bun.js/bindings/bindings.zig
```diff
@@ -261,6 +261,25 @@ pub const ZigString = extern struct {
         return this.slice()[0] == char;
     }
 
+    pub fn hasPrefix(this: ZigString, needle: []const u8) bool {
```
Jarred-Sumner: It looks like this only has one usage? If so, you can use `hasPrefixComptime`. This current implementation does not handle non-ASCII characters, which is fine since it's only using ASCII but that also means we can rely on the ASCII specific functions for this.

src/bun.js/bindings/bindings.zig
```diff
@@ -6017,8 +6032,8 @@ pub const CallFrame = opaque {
 
     fn Arguments(comptime max: usize) type {
         return struct {
-            ptr: [max]JSC.JSValue,
-            len: usize,
+            ptr: [max]JSC.JSValue = .{.undefined} ** max,
```
Jarred-Sumner: This initializes it twice. It's already zeroed in `init`.

src/bun.js/bindings/bindings.zig
```diff
@@ -6131,6 +6146,9 @@ pub fn toJSHostFunction(comptime Function: anytype) JSC.JSHostFunctionType {
 
 const DeinitFunction = *const fn (ctx: *anyopaque, buffer: [*]u8, len: usize) callconv(.C) void;
 
+pub const JSBuiltinFunctionType = fn (*JSGlobalObject) callconv(JSC.conv) JSValue;
```
Jarred-Sumner: This name is confusing. "JS Builtins" refer to the files like `ReadableStreamInternals.ts`. Is this related to the `JS2Native` generator? If so, maybe it should say `JS2NativeFunctionType` instead.

src/bun.js/ipc.zig
```diff
@@ -475,7 +479,7 @@ fn NewSocketIPCHandler(comptime Context: type) type {
             _: ?*anyopaque,
         ) void {
             // Note: uSockets has already freed the underlying socket, so calling Socket.close() can segfault
-            log("onClose\n", .{});
+            log("NewSocketIPCHandler#onClose\n", .{});
```
Jarred-Sumner, line 658: It sounds like you want to add a new `Output.scoped` for this, instead of adding a namespace inside of a namespace.

src/bun.js/javascript.zig
```diff
@@ -3616,13 +3607,24 @@ pub const VirtualMachine = struct {
                 .data => |data| {
                     IPC.log("Received IPC message from parent", .{});
                     if (this.globalThis) |global| {
+                        if (data.isObject()) {
+                            if (data.get(global, "cmd")) |prop| {
```
Jarred-Sumner: This is an expensive function call inside hot code. Do we need to do this? Can we instead make it a separate message type in the encoding?

src/bun.js/javascript.zig
```diff
@@ -3616,13 +3607,24 @@ pub const VirtualMachine = struct {
...
+                                if (prop.isString()) {
+                                    if (prop.toString(global).getZigString(global).hasPrefix("NODE_")) {
```
Jarred-Sumner: Why are we doing this? We control the API. We don't need to put ourselves in a position where we need to check if a user made observable changes.

src/bun.js/javascript.zig
```diff
@@ -3616,13 +3607,24 @@ pub const VirtualMachine = struct {
...
+                            }
+                        }
                         Process__emitMessageEvent(global, data);
                     }
                 },
             }
         }
 
         pub fn handleIPCClose(this: *IPCInstance) void {
+            Output.scoped(.IPC, false)("IPCInstance#handleIPCClose\n", .{});
```
Jarred-Sumner: Ditto re: Output.scoped inline + scope inside of a scope. This can be a `ipc_instance("handleIPCClose", .{})` or something along those lines.

src/bun.js/node/node_error_binding.zig
```diff
@@ -0,0 +1,77 @@
...
+
+pub fn ERR_INVALID_ARG_TYPE(global: *JSC.JSGlobalObject) callconv(JSC.conv) JSC.JSValue {
+    const S = struct {
+        fn cb(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
+            const arguments = callframe.arguments(3);
+            if (arguments.len < 2) {
+                globalThis.throwNotEnoughArguments("ERR_INVALID_ARG_TYPE", 2, arguments.len);
+                return .zero;
+            }
+            const arg0 = arguments.ptr[0].toString(globalThis).getZigString(globalThis).slice();
```
Jarred-Sumner: This will segfault if something which fails to coerce to a string (such a Symbol) is passed. Each of these need an exception check

src/bun.js/node/node_error_binding.zig
```diff
@@ -0,0 +1,77 @@
...
+                    return createFn(globalThis, code, message, .{});
+                }
+            };
+            return JSC.JSFunction.create(global, @tagName(code), S.cb, 0, .{});
+        }
+    };
+    return R.cbb;
+}
+
+pub fn ERR_INVALID_ARG_TYPE(global: *JSC.JSGlobalObject) callconv(JSC.conv) JSC.JSValue {
```
Jarred-Sumner: Use `$newZigFunction` instead of this wrapper struct

src/bun.js/node/node_error_binding.zig
```diff
@@ -0,0 +1,77 @@
...
+const createTypeError = JSC.JSGlobalObject.createTypeErrorInstanceWithCode;
+const createError = JSC.JSGlobalObject.createErrorInstanceWithCode;
+
+pub const ERR_SOCKET_BAD_TYPE = createSimpleError(createTypeError, .ERR_SOCKET_BAD_TYPE, "Bad socket type specified. Valid types are: udp4, udp6");
+pub const ERR_IPC_CHANNEL_CLOSED = createSimpleError(createError, .ERR_IPC_CHANNEL_CLOSED, "Channel closed");
+pub const ERR_INVALID_HANDLE_TYPE = createSimpleError(createTypeError, .ERR_INVALID_HANDLE_TYPE, "This handle type cannot be sent");
+pub const ERR_IPC_DISCONNECTED = createSimpleError(createError, .ERR_IPC_DISCONNECTED, "IPC channel is already disconnected");
+pub const ERR_CHILD_CLOSED_BEFORE_REPLY = createSimpleError(createError, .ERR_CHILD_CLOSED_BEFORE_REPLY, "Child closed before reply received");
+
+fn createSimpleError(comptime createFn: anytype, comptime code: JSC.Node.ErrorCode, comptime message: string) JSC.JSBuiltinFunctionType {
```
Jarred-Sumner: Use `$newZigFunction` instead of this wrapper struct

src/bun.js/node/node_error_binding.zig
```diff
@@ -0,0 +1,77 @@
...
+            const arg0 = arguments.ptr[0].toString(globalThis).getZigString(globalThis).slice();
+            const arg1 = arguments.ptr[1].toString(globalThis).getZigString(globalThis).slice();
+            const arg2 = arguments.ptr[2].jsTypeString(globalThis).getZigString(globalThis).slice();
+            return createTypeError(globalThis, .ERR_INVALID_ARG_TYPE, "The \"{s}\" argument must be of type {s}. Received {s}", .{ arg0, arg1, arg2 });
+        }
+    };
+    return JSC.JSFunction.create(global, "ERR_INVALID_ARG_TYPE", S.cb, 3, .{});
+}
+
+pub fn ERR_MISSING_ARGS(global: *JSC.JSGlobalObject) callconv(JSC.conv) JSC.JSValue {
```
Jarred-Sumner: All of the toString calls in this function need to check for exceptions after and return .zero if a pending exception occurs

src/bun.js/node/node_fs.zig
```diff
@@ -1136,7 +1136,7 @@ pub const Arguments = struct {
         }
 
         pub fn fromJS(ctx: JSC.C.JSContextRef, arguments: *ArgumentsSlice, exception: JSC.C.ExceptionRef) ?Truncate {
-            const path = PathOrFileDescriptor.fromJS(ctx, arguments, bun.default_allocator, exception) orelse {
+            const path = PathOrFileDescriptor.fromJS(ctx, arguments, bun.default_allocator, .truncate, exception) orelse {
```
Jarred-Sumner: We can't pass the syscall to the error? We need to pass it to each of these call sites?

src/js/internal/cluster/RoundRobinHandle.ts
```diff
@@ -0,0 +1,137 @@
+const net = require("node:net");
```
Jarred-Sumner, line 1: Is this code used? Can we delete it?

src/js/internal/cluster/SharedHandle.ts
```diff
@@ -0,0 +1,42 @@
+const dgram = require("internal/dgram");
```
Jarred-Sumner: Is this code used? Can we delete it?

src/output.zig
```diff
@@ -562,6 +562,8 @@ pub noinline fn print(comptime fmt: string, args: anytype) callconv(std.builtin.
     }
 }
 
+var lock = std.Thread.Mutex{};
```
Jarred-Sumner: whats the story behind this change? where is this used?

src/js/node/util.ts
```diff
@@ -3,6 +3,11 @@ const types = require("node:util/types");
 /** @type {import('node-inspect-extracted')} */
 const utl = require("internal/util/inspect");
 
+const ERR_INVALID_ARG_TYPE = $zig("node_error_binding.zig", "ERR_INVALID_ARG_TYPE");
```
Jarred-Sumner:
```suggestion
const ERR_INVALID_ARG_TYPE = $newZigFunction("node_error_binding.zig", "ERR_INVALID_ARG_TYPE");
```

src/js/node/net.ts
```diff
@@ -1024,6 +1027,78 @@ function toNumber(x) {
   return (x = Number(x)) >= 0 ? x : false;
 }
 
+// Returns handle if it can be created, or error code if it can't
+function _createServerHandle(address, port, addressType?, fd?, flags?) {
```
Jarred-Sumner: Delete this? I don't think this code is used or needs to be used

src/js/node/http.ts
```diff
@@ -527,6 +523,7 @@ Server.prototype.listen = function (port, host, backlog, onListen) {
     this[serverSymbol] = Bun.serve<any>({
       tls,
       port,
+      reusePort: !cluster.isPrimary,
```
Jarred-Sumner: This would break SO_REUSEPORT in node:http. It is an option the developer can pass to node:http. We should support that option.

src/js/node/http.ts
```diff
@@ -2,13 +2,16 @@
 const EventEmitter = require("node:events");
 const { isTypedArray } = require("node:util/types");
 const { Duplex, Readable, Writable, ERR_STREAM_WRITE_AFTER_END, ERR_STREAM_ALREADY_FINISHED } = require("node:stream");
+const cluster = require("node:cluster");
```
Jarred-Sumner: This adds overhead to the start time. When do we need to do this? Let's call require as late as possible. if it's only for the below usage in `reusePort`, then we should delete it entirely

src/js/node/cluster.ts
```diff
@@ -1,40 +1,42 @@
...
-  };
-
-  setupPrimary() {
-    throwNotImplemented("node:cluster", 2428);
-  }
 
-  setupMaster() {
-    throwNotImplemented("node:cluster", 2428);
-  }
+const child_process = require("node:child_process");
```
Jarred-Sumner: Let's move this require to as late as possible. People may `require` `node:cluster` without actually using it.

src/js/node/net.ts
```diff
@@ -1024,6 +1027,78 @@ function toNumber(x) {
...
+
+  if (err) {
+    handle.close();
+    return err;
+  }
+
+  return handle;
+}
+
+function createHandle(fd, is_server) {
```
Jarred-Sumner: Can we delete this?

src/js/internal/linkedlist.ts
```diff
@@ -0,0 +1,45 @@
+export function init(list) {
```
Jarred-Sumner, line 1: Can we delete this file?

src/js/node/child_process.ts
```diff
@@ -2,12 +2,17 @@
 const EventEmitter = require("node:events");
 const StreamModule = require("node:stream");
 const OsModule = require("node:os");
+const FsModule = require("node:fs");
```
Jarred-Sumner: Why is this loading the fs module here?

src/js/node/child_process.ts
```diff
@@ -1226,21 +1224,29 @@ class ChildProcess extends EventEmitter {
...
 
-    // TODO: better ipc support
-    const ipc = $isJSArray(stdio) && stdio[3] === "ipc";
-    var env = options.envPairs || undefined;
+    const has_ipc = $isJSArray(stdio) && stdio[3] === "ipc";
+    var env = options.envPairs || process.env;
+
+    if (has_ipc) {
+      // Let child process know about opened IPC channel
+      if (env === undefined) env = {};
```
Jarred-Sumner: This branch is not reachable

src/js/node/child_process.ts
```diff
@@ -1258,7 +1264,8 @@ class ChildProcess extends EventEmitter {
         );
       },
       lazy: true,
-      ipc: ipc ? this.#emitIpcMessage.bind(this) : undefined,
+      ipc: has_ipc ? this.#emitIpcMessage.bind(this) : undefined,
+      onDisconnect: () => this.#disconnect(),
```
Jarred-Sumner: Can we do the same check here? `has_ipc`

src/js/node/child_process.ts
```diff
@@ -1283,6 +1291,10 @@ class ChildProcess extends EventEmitter {
   }
 
   #emitIpcMessage(message) {
+    if (message?.cmd?.startsWith("NODE_")) {
```
Jarred-Sumner: This can be moved into a native binding

src/js/node/child_process.ts
```diff
@@ -1777,6 +1791,113 @@ function genericNodeError(message, options) {
   return err;
 }
 
+function _forkChild(fd, serializationMode) {
```
Jarred-Sumner: Does this need to be in JS?

src/js/node/tty.ts
```diff
@@ -4,6 +4,8 @@ const {
   getWindowSize: _getWindowSize,
 } = $cpp("ProcessBindingTTYWrap.cpp", "createBunTTYFunctions");
 
+const ERR_INVALID_ARG_TYPE = $zig("node_error_binding.zig", "ERR_INVALID_ARG_TYPE");
```
Jarred-Sumner: We are creating this function several times. Might as well put it in a module and load from the cached copy.

src/js/node/util.ts
```diff
@@ -280,6 +285,13 @@ function styleText(format, text) {
   return `\u001b[${formatCodes[0]}m${text}\u001b[${formatCodes[1]}m`;
 }
 
+function getSystemErrorName(err: any) {
```
Jarred-Sumner: Can't we do this in native code and validate the arguments there?

src/bun.js/node/types.zig
```diff
@@ -865,9 +865,13 @@ pub const PathLike = union(enum) {
 };
 
 pub const Valid = struct {
-    pub fn fileDescriptor(fd: i64, ctx: JSC.C.JSContextRef, exception: JSC.C.ExceptionRef) bool {
+    pub fn fileDescriptor(fd: i64, ctx: JSC.C.JSContextRef, syscall: ?Syscall.Tag, exception: JSC.C.ExceptionRef) bool {
```
Jarred-Sumner: It would be better to avoid using JSC.C.ExceptionRef instead of passing the syscall through

src/bun.js/node/types.zig
```diff
@@ -865,9 +865,13 @@ pub const PathLike = union(enum) {
...
         if (fd < 0) {
-            JSC.throwInvalidArguments("Invalid file descriptor, must not be negative number", .{}, ctx, exception);
+            const te = JSC.toTypeError(JSC.Node.ErrorCode.ERR_INVALID_ARG_TYPE, "Invalid file descriptor, must not be negative number", .{}, ctx);
```
Jarred-Sumner: This sets `code` twice. We shouldn't do that

src/bun.js/node/types.zig
```diff
@@ -865,9 +865,13 @@ pub const PathLike = union(enum) {
...
+            te.put(ctx, JSC.ZigString.static("errno"), JSC.JSValue.jsNumber(-9));
+            te.put(ctx, JSC.ZigString.static("code"), JSC.ZigString.static("EBADF").toJS(ctx));
```
Jarred-Sumner: We should use struct for all these

src/bun.js/node/node_util_binding.zig
```diff
@@ -0,0 +1,102 @@
...
+        fn cb(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
+            const arguments = callframe.arguments(1).slice();
+            if (arguments.len < 1) {
+                globalThis.throwNotEnoughArguments("internalErrorName", 1, arguments.len);
+                return .zero;
+            }
+            const err_value = arguments[0];
+            const err_int = err_value.toInt32();
+            const err_i: isize = err_int;
+            const err_e: std.c.E = @enumFromInt(-err_i);
```
Jarred-Sumner: This is not a safe cast. It needs to hceck if it's in the range of the int.

src/bun.js/node/node_util_binding.zig
```diff
@@ -0,0 +1,102 @@
...
+                .named_pipe => 4,
+                .unknown => 5,
+                else => @panic("unreachable"),
+            }));
+        }
+    };
+    return JSC.JSFunction.create(global, "guessHandleType", S.cb, 1, .{});
+}
+
+fn uv_guess_handle(file: uv.uv_file) uv.uv_handle_type {
```
Jarred-Sumner: Let's delete this code

src/bun.js/node/node_util_binding.zig
```diff
@@ -0,0 +1,102 @@
+const std = @import("std");
+const bun = @import("root").bun;
+const Environment = bun.Environment;
+const JSC = bun.JSC;
+const string = bun.string;
+const Output = bun.Output;
+const ZigString = JSC.ZigString;
+const uv = bun.windows.libuv;
+
+pub fn guessHandleType(global: *JSC.JSGlobalObject) callconv(JSC.conv) JSC.JSValue {
```
Jarred-Sumner: ✂️

src/bun.js/javascript.zig
```diff
@@ -924,19 +914,19 @@ pub const VirtualMachine = struct {
             this.hide_bun_stackframes = false;
         }
 
-        if (map.map.fetchSwapRemove("NODE_CHANNEL_FD")) |kv| {
-            const mode = if (map.map.fetchSwapRemove("NODE_CHANNEL_SERIALIZATION_MODE")) |mode_kv|
-                IPC.Mode.fromString(mode_kv.value.value) orelse .json
+        if (map.map.get("NODE_CHANNEL_FD")) |kv| {
+            const mode = if (map.map.get("NODE_CHANNEL_SERIALIZATION_MODE")) |mode_kv|
```
Jarred-Sumner: We should preserve this environment variable? What happens if you use worker_threads inside of an IPC child.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2791,6 +2781,19 @@ extern "C" void Process__emitMessageEvent(Zig::GlobalObject* global, EncodedJSVa
     }
 }
 
+extern "C" void Process__emitInternalMessageEvent(Zig::GlobalObject* global, EncodedJSValue value)
+{
+    auto* process = static_cast<Process*>(global->processObject());
+    auto& vm = global->vm();
+
+    auto ident = Identifier::fromString(vm, "internalMessage"_s);
```
Jarred-Sumner: ✂️ lets use a non-event emitter binding for this

src/bun.js/api/bun/subprocess.zig
```diff
@@ -1810,6 +1815,18 @@ pub const Subprocess = struct {
                     }
                 }
 
+                if (args.getTruthy(globalThis, "onDisconnect")) |onDisconnect_| {
```
Jarred-Sumner, line 1810: maybe instead we let you pass an object to ipc?

```ts
Bun.spawn({
  cmd: process.argv,
  ipc: {
    message(data, subprocess) {},
    error(error) {},
    close(subprocess) {},
  },
});
```

docs/runtime/web-apis.md
```diff
@@ -7,22 +7,37 @@ The following Web APIs are partially or completely supported.
...
 - URLs
-- [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
+- [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL)
+  [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
 
 ---
 
 - Web Workers
-- [`Worker`](https://developer.mozilla.org/en-US/docs/Web/API/Worker) [`self.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/postMessage) [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) [`MessagePort`](https://developer.mozilla.org/en-US/docs/Web/API/MessagePort) [`MessageChannel`](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel), [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel).
+- [`Worker`](https://developer.mozilla.org/en-US/docs/Web/API/Worker)
```
Jarred-Sumner: The formatting of these was intentional. Prefer reading on bun.sh/docs instead of GitHub.

test/js/node/cluster/cluster.test.ts
```diff
@@ -0,0 +1,15 @@
...
+  test(name, () => {
+    expect([path.join(import.meta.dir, "fixtures", name)]).toRun();
+  });
+}
+
+new_test("cwd.ts");
+new_test("call-and-destroy.ts");
+new_test("worker-constructor.ts");
+new_test("worker-death.ts");
+new_test("worker-events.ts");
```
Jarred-Sumner: Where are tests that check that responding to HTTP requests works?

src/js/internal/cluster/child.ts
```diff
@@ -0,0 +1,279 @@
...
+    removeIndexesKey(indexesKey, index);
+    return close.$apply(handle, arguments);
+  };
+  $assert(handles.has(key) === false);
+  handles.set(key, handle);
+  cb(message.errno, handle);
+}
+
+// Round-robin. Master distributes handles across workers.
+function rr(message, { indexesKey, index }, cb) {
```
Jarred-Sumner, line 142: What is this distributing?

src/js/internal/cluster/primary.ts
```diff
@@ -0,0 +1,345 @@
...
+else if (schedulingPolicyEnv === "none") schedulingPolicy = SCHED_NONE;
+else if (process.platform === "win32") {
+  // Round-robin doesn't perform well on
+  // Windows due to the way IOCP is wired up.
+  schedulingPolicy = SCHED_NONE;
+} else schedulingPolicy = SCHED_RR;
+
+cluster.schedulingPolicy = schedulingPolicy;
+
+cluster.setupPrimary = function (options) {
```
Jarred-Sumner, line 49: Most of this should be unnecessary?

src/js/internal/cluster/utils.ts
```diff
@@ -0,0 +1,36 @@
+const callbacks = new Map();
+let seq = 0;
+
+export function sendHelper(proc, message, handle?, cb?) {
+  if (!proc.connected) return false;
+
+  // Mark message as internal. See INTERNAL_PREFIX
+  // in lib/internal/child_process.js
+  message = { cmd: "NODE_CLUSTER", ...message, seq };
```
Jarred-Sumner: We should use a different codepath for this that is not observable to users

Jarred-Sumner (changes requested):
Comments

docs/runtime/web-apis.md
nektro: this change does not affect the display of the website

Zer0xxxx:
I'm very pleased to see the seriousness with which this feature is progressing and has been reviewed! Do you think it will be a "Preview" feature or really stable once the PR will be merged? Either way, I'd be happy to test it!

src/js/node/util.ts
nektro: this is public api https://nodejs.org/docs/latest/api/util.html#utilgetsystemerrornameerr

src/bun.js/bindings/bindings.zig
nektro: i changed this impl so the elements default to `.undefined` instead of `.zero`

src/output.zig
nektro: all `Output` instances use their own lock so it was very easy for them to clobber each other when many logs are printing at the same time on different threads

test/js/node/cluster/cluster.test.ts
nektro: added now

CsVeryLoveXieWenLi:
**祝你们好运，加油。**
**good luck!**

src/bun.js/api/BunObject.zig
```diff
@@ -3436,6 +3438,26 @@ pub fn serve(
     unreachable;
 }
 
+extern fn Bun__setChannelRef(*JSC.JSGlobalObject, bool) void;
+pub fn setRef(globalObject: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) JSC.JSValue {
```
nektro: This should be in a file related to IPC/node:cluster and it should not be Bun.setRef.

src/bun.js/api/bun/socket.zig
```diff
@@ -179,8 +179,7 @@ const Handlers = struct {
         socket_context: ?*uws.SocketContext,
 
         pub fn exit(this: *Scope, ssl: bool, wrapped: WrappedType) void {
-            var vm = this.handlers.vm;
-            defer vm.eventLoop().exit();
+            this.handlers.vm.eventLoop().exit();
```
nektro:
```suggestion
     const vm = this.handlers.vm;
     defer vm.eventLoop().exit();
     // markInactive may free handlers, and we exit the event loop only after mark inactive is called.
```

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2254,10 +2273,13 @@ pub const Subprocess = struct {
         return sync_value;
     }
 
+    const node_cluster_binding = @import("./../../node/node_cluster_binding.zig");
+
     pub fn handleIPCMessage(
         this: *Subprocess,
         message: IPC.DecodedIPCMessage,
     ) void {
+        IPClog("Subprocess#handleIPCMessage\n", .{});
```
nektro:
```suggestion
        IPClog("Subprocess#handleIPCMessage", .{});
```

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2275,17 +2297,36 @@ pub const Subprocess = struct {
...
+        IPClog("Subprocess#handleIPCClose\n", .{});
         this.updateHasPendingActivity();
+        const ok = this.ipc_data != null;
+        if (ok) this.ipc().iimh.deinit();
+        this.ipc_data = null;
+
+        const this_jsvalue = this.this_jsvalue;
+        this_jsvalue.ensureStillAlive();
+        if (this.on_disconnect_callback.trySwap()) |callback| {
+            _ = callback.call(this.globalThis, this_jsvalue, &.{
```
nektro: Does this need to be wrapped in a event_loop.enter() and event_loop.exit()? What happens if a Promise is awaited inside the disconnect callback?

src/js/node/net.ts
```diff
@@ -922,45 +943,75 @@ class Server extends EventEmitter {
...
+  }
+
+  get _handle() {
     return this;
   }
+  set _handle(new_handle) {
+    //nothing
+  }
+
+  getsockname(out) {
```
nektro, line 1042: This function is unnecessary

```js
out.port = this.address().port
```

src/js/node/net.ts
```diff
@@ -1016,6 +1111,23 @@ function normalizeArgs(args) {
   return arr;
 }
 
+function checkBindError(err, port, handle) {
```
nektro, line 1145: Can you add a test for this or delete it?

src/sys.zig
```diff
@@ -253,6 +253,7 @@ pub const Tag = enum(u8) {
 
     // Below this line are Windows API calls only.
 
+    ReadFile,
```
nektro: ✂️

test/harness.ts
```diff
@@ -362,6 +362,7 @@ expect.extend({
       cmd: [bunExe(), ...cmds],
       env: bunEnv,
       stdio: ["inherit", "pipe", "inherit"],
+      cwd: process.cwd(),
```
nektro: This shouldn't be necessary, this is the default value. Unless there's code which is overriding the cwd function.

aralroca:
I hope it will merge soon! 🎉

src/bun.js/api/bun/subprocess.zig
```diff
@@ -717,7 +716,8 @@ pub const Subprocess = struct {
         }
     }
 
-    pub fn doSend(this: *Subprocess, global: *JSC.JSGlobalObject, callFrame: *JSC.CallFrame) JSValue {
+    pub fn doSend(this: *Subprocess, global: *JSC.JSGlobalObject, callFrame: *JSC.CallFrame) callconv(.C) JSValue {
```
Jarred-Sumner: The callconv here is unnecessary

src/bun.js/api/bun/subprocess.zig
```diff
@@ -740,10 +740,18 @@ pub const Subprocess = struct {
         return JSC.JSValue.jsUndefined();
     }
 
-    pub fn disconnect(this: *Subprocess) void {
-        const ipc_data = this.ipc_data orelse return;
-        ipc_data.socket.close(.normal);
+    pub fn disconnect(this: *Subprocess, globalThis: *JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSValue {
```
Jarred-Sumner: The callconv here is unnecessary

src/bun.js/api/bun/subprocess.zig
```diff
@@ -740,10 +740,18 @@ pub const Subprocess = struct {
...
+        _ = globalThis;
+        _ = callframe;
+        const ipc_data = this.ipc_maybe() orelse return .undefined;
+        ipc_data.close();
         this.ipc_data = null;
+        return .undefined;
+    }
+
+    pub fn getConnected(this: *Subprocess, globalThis: *JSGlobalObject) callconv(.C) JSValue {
```
Jarred-Sumner: The callconv here is unnecessary

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,239 @@
...
+
+pub fn onInternalMessageChild(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
+    const arguments = callframe.arguments(2).ptr;
+    InternalMsgHolder.worker = JSC.Strong.create(arguments[0], globalThis);
+    InternalMsgHolder.cb = JSC.Strong.create(arguments[1], globalThis);
+    return .undefined;
+}
+
+pub fn handleInternalMessageChild(globalThis: *JSC.JSGlobalObject, message: JSC.JSValue) void {
+    if (message.get(globalThis, "ack")) |p| {
```
Jarred-Sumner: Why is this an object?

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,239 @@
...
+        .null, // handle
+    });
+    return;
+}
+
+//
+//
+//
+
+pub const IInternalMsgHolder = struct {
```
Jarred-Sumner:
```suggestion
pub const InternalMsgHolder = struct {
```

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,239 @@
...
+            ZigString.static("message").toJS(globalThis),
+            ZigString.static("object").toJS(globalThis),
+            message,
+        ));
+    }
+    if (callback.isFunction()) {
+        ipc_data.iimh.callbacks.put(bun.default_allocator, ipc_data.iimh.seq, JSC.Strong.create(callback, globalThis)) catch bun.outOfMemory();
+    }
+
+    message.put(globalThis, ZigString.static("seq"), JSC.JSValue.jsNumber(ipc_data.iimh.seq));
```
Jarred-Sumner: Why is this an object?

src/bun_js.zig
```diff
@@ -418,7 +418,7 @@ pub const Run = struct {
 
         vm.onUnhandledRejection = &onUnhandledRejectionBeforeClose;
         vm.global.handleRejectedPromises();
-        if (this.any_unhandled and this.vm.exit_handler.exit_code == 0) {
+        if (this.any_unhandled and this.vm.exit_handler.exit_code == null) {
```
Jarred-Sumner: This is a bug. I think it's better to revert it to the previous behavior.

Exit code 0:
```js
❯ bun-11492 /tmp/a.js
1 | process.exitCode = 0;
2 | queueMicrotask(() => {
3 |   throw new Error('oopsie');
            ^
error: oopsie
      at /private/tmp/a.js:3:9
```

Exit code 1:

```js
❯ bun /tmp/a.js
1 | process.exitCode = 0;
2 | queueMicrotask(() => {
3 |   throw new Error('oopsie');
            ^
error: oopsie
      at /private/tmp/a.js:3:9

Bun v1.1.21-canary.1+79ddf0e47 (macOS arm64)
```

Exit code 1:

```js
❯ node /tmp/a.js
/private/tmp/a.js:3
  throw new Error('oopsie');
        ^

Error: oopsie
    at /private/tmp/a.js:3:9
    at node:internal/process/task_queues:140:7
    at AsyncResource.runInAsyncScope (node:async_hooks:206:9)
    at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8)

Node.js v22.4.1
```

src/io/PipeWriter.zig
```diff
@@ -1301,7 +1304,7 @@ pub fn WindowsStreamingWriter(
             // clean both buffers if needed
             this.outgoing.deinit();
             this.current_payload.deinit();
-            this.close();
```
Jarred-Sumner, line 1309: This could explain the test failures.

src/io/PipeWriter.zig
```diff
@@ -1150,7 +1152,8 @@ pub fn WindowsStreamingWriter(
 
         fn onCloseSource(this: *WindowsWriter) void {
             this.source = null;
-            if (!this.closed_without_reporting) {
+            if (this.closed_without_reporting) {
```
Jarred-Sumner, line 1158: This could also explain the test failures

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3919,6 +3919,20 @@ JSC::JSValue GlobalObject::moduleLoaderEvaluate(JSGlobalObject* lexicalGlobalObj
     return result;
 }
 
+JSC::JSValue GlobalObject::createErrorWithCode(ASCIILiteral code, ASCIILiteral message)
```
Jarred-Sumner: ASCIILiteral needs to be `inline`, it's supposed to be a `constexpr` always

robobun:
@nektro, your commit 278a32343b46a606e01e93a28c1a77947fbd6e1b is building: [`#928`](https://buildkite.com/bun/bun/builds/928)

scripts/runner.node.mjs
```diff
@@ -810,6 +811,12 @@ function isJavaScript(path) {
  * @returns {boolean}
  */
 function isTest(path) {
+  if (path.includes("/test-cluster-") && path.endsWith(".js")) return true;
```
Electroid: It would be nice to avoid this, because that means running `bun test` vs. this script will cause different list of tests to run.

What if we just rename the upstream files to have the `.test` extension?

aralroca: And would it make sense to change the `/` to `path.sep` or join it with `path.join` to make the test pass on Windows as well?

nektro: > It would be nice to avoid this, because that means running `bun test` vs. this script will cause different list of tests to run.
> What if we just rename the upstream files to have the `.test` extension?

the files do not work correctly unless run directly with `bun <file>`. getting the files to run in CI is much more important than a global `bun test` since during development most files are run individually anyway. the main issue with this approach is that RunCommand doesnt respond to `--timeout`. does `Bun.spawn` accept an abort signal or is there a CI env var we can use to communicate the timeout?

> And would it make sense to change the `/` to `path.sep` or join it with `path.join` to make the test pass on Windows as well?

yes, and i have this change locally

nektro: > does `Bun.spawn` accept an abort signal or is there a CI env var we can use to communicate the timeout?

oh the runner is using `node:child_process.spawn()` which does so i can look into this

nektro: excellent, we do

nektro: I see there's also a `timeout` option which we already set at 3 minutes, which feels a bit high, even outside this

nektro: hm timeout doesnt throw an error or set the exit code so i might change this to use AbortController anyway. dont want timing out tests to show up as mistakenly passing

nektro: hm i see now runner.node.mjs already has a bunch of extra logic to smooth over this so i wont change it for now

nektro:
https://github.com/oven-sh/bun/pull/11492/#issuecomment-2257181089 buildkite comment

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -892,22 +898,33 @@ extern "C" int Bun__handleUnhandledRejection(JSC::JSGlobalObject* lexicalGlobalO
...
                     && eventEmitter.listenerCount(eventName) == 1) {
                     Bun__ensureProcessIPCInitialized(global);
-                    eventEmitter.scriptExecutionContext()->refEventLoop();
-                    eventEmitter.m_hasIPCRef = true;
+                    Bun__setChannelRef(global, true);
                 }
             } else {
-                if (eventEmitter.listenerCount(eventName) == 0 && eventEmitter.m_hasIPCRef) {
-                    eventEmitter.scriptExecutionContext()->unrefEventLoop();
+                if (eventEmitter.listenerCount(eventName) == 0) {
```
Jarred-Sumner, line 926: process.off("message") and process.off("disconnect") should both cause channel.unref()?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -1916,6 +1933,29 @@ static JSValue constructProcessSend(VM& vm, JSObject* processObject)
     }
 }
 
+JSC_DEFINE_HOST_FUNCTION(processDisonnectFinish, (JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
+{
+    Bun__closeChildIPC(globalObject);
+    return {};
```
Jarred-Sumner: This function will always crash. I don't see a test failure

Can you add a test?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2871,6 +3005,7 @@ extern "C" void Process__emitDisconnectEvent(Zig::GlobalObject* global)
   binding                          Process_functionBinding                             Function 1
   browser                          constructBrowser                                    PropertyCallback
   chdir                            Process_functionChdir                               Function 1
+  channel                          constructProcessChannel                             PropertyCallback
```
Jarred-Sumner, line 3007: Should this be `DontEnum`?

src/bun.js/bindings/BunProcess.h
```diff
@@ -2,6 +2,7 @@
 
 #include "root.h"
 
+#include "ZigGlobalObject.h"
```
Jarred-Sumner:
```suggestion
namespace Zig {
  class GlobalObject;
}
```

src/bun.js/ipc.zig
```diff
@@ -197,12 +255,35 @@ const json = struct {
...
+
+        // TODO: it would be cool to have a 'toUTF8Into' which can write directly into 'ipc_data.outgoing.list'
+        const str = out.toUTF8(bun.default_allocator);
+        defer str.deinit();
+
+        const slice = str.slice();
+
+        try writer.ensureUnusedCapacity(1 + slice.len + 1);
 
+        writer.writeAssumeCapacity(&.{2});
```
Jarred-Sumner, line 287: What is 2?

src/bun.js/ipc.zig
```diff
@@ -237,8 +326,8 @@ const SocketIPCData = struct {
 
     incoming: bun.ByteList = .{}, // Maybe we should use StreamBuffer here as well
     outgoing: bun.io.StreamBuffer = .{},
-
     has_written_version: if (Environment.allow_assert) u1 else u0 = 0,
+    iimh: node_cluster_binding.InternalMsgHolder = .{},
```
Jarred-Sumner: can we use the full name instead of an acronym? who will know what an "iimh" is?

src/bun.js/ipc.zig
```diff
@@ -724,18 +888,19 @@ fn NewNamedPipeIPCHandler(comptime Context: type) type {
                     return;
                 },
                 .result => {
+                    client.setBlocking(false);
```
Jarred-Sumner: Was this intended to still be here? Or is this from before? cc @cirospaciari

src/bun.js/javascript.zig
```diff
@@ -426,39 +426,99 @@ pub export fn Bun__GlobalObject__hasIPC(global: *JSC.JSGlobalObject) bool {
...
-    if (vm.getIPCInstance()) |ipc_instance| {
-        const success = ipc_instance.data.serializeAndSend(globalObject, callFrame.argument(0));
-        return if (success) .undefined else .zero;
+    const zigGlobal: *JSC.ZigGlobalObject = @ptrCast(globalObject);
+    const ipc_instance = vm.getIPCInstance() orelse {
+        const ex = globalObject.ERR_IPC_CHANNEL_CLOSED("Channel closed.", .{}).toJS();
+        if (callback.isFunction()) {
+            process_queueNextTick1(zigGlobal, callback, ex);
+        } else {
+            if (S.value == null) {
```
Jarred-Sumner: What is this doing? Why is it doing this?

src/bun.js/javascript.zig
```diff
@@ -426,39 +426,99 @@ pub export fn Bun__GlobalObject__hasIPC(global: *JSC.JSGlobalObject) bool {
...
+        options_ = .undefined;
+    } else if (options_.isFunction()) {
+        callback = options_;
+        options_ = .undefined;
+    } else if (!options_.isUndefined()) {
+        if (!globalObject.validateObject("options", options_, .{})) return .zero;
+    }
+
+    const S = struct {
+        fn impl(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
```
Jarred-Sumner, line 455: This creates a temporary function on every call to process.send missing a callback

Do we need to do that?

src/bun.js/javascript.zig
```diff
@@ -3758,25 +3825,45 @@ pub const VirtualMachine = struct {
...
+            vm.ipc = null;
+            const event_loop = vm.eventLoop();
+            node_cluster_binding.child_singleton.deinit();
+            event_loop.enter();
+            Process__emitDisconnectEvent(vm.global);
+            event_loop.exit();
             if (Environment.isPosix) {
                 uws.us_socket_context_free(0, this.context);
             }
             this.destroy();
```
Jarred-Sumner, line 3842: This seems dangerous

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,275 @@
...
+
+    // similar code as Bun__Process__send
+    var formatter = JSC.ConsoleObject.Formatter{ .globalThis = globalThis };
+    if (Environment.isDebug) log("child: {}", .{message.toFmt(&formatter)});
+
+    const ipc_instance = vm.getIPCInstance().?;
+    const process_queueNextTick1 = Bun__Process__queueNextTick1;
+    const zigGlobal: *JSC.ZigGlobalObject = @ptrCast(globalThis);
+
+    const S = struct {
```
Jarred-Sumner, line 56: Same thing, we shouldn't be creating a temporary method.

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,275 @@
...
+    message.put(globalThis, ZigString.static("seq"), JSC.JSValue.jsNumber(child_singleton.seq));
+    child_singleton.seq +%= 1;
+
+    // similar code as Bun__Process__send
+    var formatter = JSC.ConsoleObject.Formatter{ .globalThis = globalThis };
+    if (Environment.isDebug) log("child: {}", .{message.toFmt(&formatter)});
+
+    const ipc_instance = vm.getIPCInstance().?;
+    const process_queueNextTick1 = Bun__Process__queueNextTick1;
+    const zigGlobal: *JSC.ZigGlobalObject = @ptrCast(globalThis);
```
Jarred-Sumner: This pointer cast is unnecessary

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,275 @@
...
+            ZigString.static("message"),
+            ZigString.static("object"),
+            message,
+        ));
+    }
+    if (callback.isFunction()) {
+        child_singleton.callbacks.put(bun.default_allocator, child_singleton.seq, JSC.Strong.create(callback, globalThis)) catch bun.outOfMemory();
+    }
+
+    message.put(globalThis, ZigString.static("seq"), JSC.JSValue.jsNumber(child_singleton.seq));
```
Jarred-Sumner, line 47: Why are we doing this? Can you leave a comment explaining?

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,275 @@
...
+    const zigGlobal: *JSC.ZigGlobalObject = @ptrCast(globalThis);
+
+    const S = struct {
+        fn impl(globalThis_: *JSC.JSGlobalObject, callframe_: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
+            const arguments_ = callframe_.arguments(1).slice();
+            const ex = arguments_[0];
+            Process__emitErrorEvent(globalThis_, ex);
+            return .undefined;
+        }
+        var value: ?JSC.JSValue = null;
```
Jarred-Sumner: We cannot hold a JSValue this way. The GC won't see it. It's a global variable for something which is inherently thread-local.

src/deps/libuv.zig
```diff
@@ -2848,6 +2848,10 @@ fn StreamMixin(comptime Type: type) type {
             }
             return .{ .result = {} };
         }
+        /// Be aware that this only works with pipes on Windows.
+        pub fn setBlocking(this: *Type, blocking: bool) void {
+            _ = uv_stream_set_blocking(@ptrCast(this), @intFromBool(blocking));
```
Jarred-Sumner: Is this still necessary?

src/js/internal/cluster/primary.ts
```diff
@@ -0,0 +1,320 @@
...
+    ...options,
+  };
+
+  cluster.settings = settings;
+
+  if (initialized === true) return process.nextTick(setupSettingsNT, settings);
+
+  initialized = true;
+  schedulingPolicy = cluster.schedulingPolicy; // Freeze policy.
+  console.assert(
```
Jarred-Sumner: `console.assert`?

src/js/node/http.ts
```diff
@@ -138,7 +145,8 @@ var FakeSocket = class Socket extends Duplex {
   address() {
     // Call server.requestIP() without doing any propety getter twice.
     var internalData;
-    return (this.#address ??= (internalData = this[kInternalSocketData])?.[0]?.requestIP(internalData[2]) ?? {});
+    return (this.#address ??=
+      (internalData = this[kInternalSocketData])?.[0]?.[serverSymbol].requestIP(internalData[2]) ?? {});
```
Jarred-Sumner, line 150: Why is this an array now?

src/js/node/http.ts
```diff
@@ -591,8 +677,12 @@ Server.prototype = {
             return pendingResponse;
           }
 
+          server[connectionsSymbol]++;
           var { promise, resolve: resolveFunction, reject: rejectFunction } = $newPromiseCapability(GlobalPromise);
-          return promise;
+          return promise.finally(() => {
```
Jarred-Sumner: This adds an extra microtask tick to every node:http response.

Is that necessary?

src/js/node/http.ts
```diff
@@ -1411,10 +1497,12 @@ class ClientRequest extends OutgoingMessage {
     this.#bodyChunks.push(...chunks);
     callback();
   }
+
   _destroy(err, callback) {
     this.destroyed = true;
     // If request is destroyed we abort the current response
     this[kAbortController]?.abort?.();
+    this.socket.destroy();
```
Jarred-Sumner, line 1489: Is there always going to be a socket here? And should it be destroyed on every call?

Jarred-Sumner (changes requested):
At the time of writing, this PR makes `node:http` 15% slower on Linux x64. We cannot merge it until that is fixed. There are also some comments.

On Linux x64, `bench/snippets/http-hello.node.mjs`

| Version | Requests per second | 
| --------|---------------------|
| ab08c590d (this PR)  |  53,920 |
| fe7f5fa73 (canary) | 62,726 |

cmd:
```sh
oha http://localhost:3000 -n 200000
```

src/bun.js/bindings/BunProcess.cpp
nektro: yep
https://github.com/nodejs/node/blob/v22.6.0/lib/child_process.js#L180-L185
https://github.com/nodejs/node/blob/v22.6.0/lib/internal/child_process.js#L551-L561

src/bun.js/bindings/BunProcess.cpp
nektro: in what condition do you see this as being able to cause a crash?

src/bun.js/bindings/BunProcess.cpp
nektro: it could be. i dont feel strongly in either direction

src/bun.js/ipc.zig
nektro: a header for the serialized messages. 1 is regular json, 2 is internal

src/bun.js/javascript.zig
nektro: no it creates one and memoizes it

src/bun.js/node/node_cluster_binding.zig
nektro: it only creates once

src/bun.js/javascript.zig
nektro: doing the same error handling node does if someone calls `process.send()` when ipc isnt active

src/bun.js/node/node_cluster_binding.zig
nektro: done

src/deps/libuv.zig
nektro: cc @cirospaciari

cirospaciari: not any more, sorry I should have removed this when changed the implementation

src/bun.js/ipc.zig
cirospaciari: This is from before we dont need to setBlocking anymore, I belive the main branch dont call it anymore, ` onNewClientConnect`  is dead code and is will not be called

src/deps/libuv.zig
nektro: roger

src/bun.js/ipc.zig
nektro: roger

src/js/node/http.ts
nektro: its not an array, `kInternalSocketData` was changed from holding a `bun.Server` to an `http.Server` so there's a `serverSymbol` access

src/bun.js/bindings/BunProcess.cpp
nektro: clarified offline, `return {}` is problematic if an exception hasnt been thrown

src/js/node/http.ts
nektro: FakeSocket yes, this is only called if explicitly i believe

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2287,17 +2309,42 @@ pub const Subprocess = struct {
...
+        event_loop.enter();
+        defer event_loop.exit();
+
+        const this_jsvalue = this.this_jsvalue;
+        this_jsvalue.ensureStillAlive();
+        if (this.on_disconnect_callback.trySwap()) |callback| {
+            _ = callback.call(this.globalThis, this_jsvalue, &.{
+                JSValue.jsBoolean(ok),
+            });
+            vm.drainMicrotasks();
```
Jarred-Sumner: event_loop.exit will drain microtasks if necessary
```suggestion

```

src/bun.js/api/server.zig
```diff
@@ -5965,6 +5967,26 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
 
         pub fn deinitIfWeCan(this: *ThisServer) void {
             httplog("deinitIfWeCan", .{});
+
+            const vm = this.globalThis.bunVM();
+
+            if (this.pending_requests == 0 and this.listener == null and !this.hasActiveWebSockets() and !this.flags.has_handled_all_closed_promise) {
+                const event_loop = vm.eventLoop();
+                event_loop.enter();
```
Jarred-Sumner: why are we entering the event loop and then enqueuing a task?

src/bun.js/api/server.zig
```diff
@@ -6035,6 +6058,7 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
                 .base_url_string_for_joining = bun.default_allocator.dupe(u8, strings.trim(config.base_url.href, "/")) catch unreachable,
                 .vm = JSC.VirtualMachine.get(),
                 .allocator = Arena.getThreadlocalDefault(),
+                .all_closed_promise = JSC.JSPromise.Strong.init(globalThis),
```
Jarred-Sumner: This should not always be created. We only support this feature from node:http.

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,280 @@
...
+            ZigString.static("object"),
+            message,
+        ));
+    }
+    if (callback.isFunction()) {
+        child_singleton.callbacks.put(bun.default_allocator, child_singleton.seq, JSC.Strong.create(callback, globalThis)) catch bun.outOfMemory();
+    }
+
+    // sequence number for InternalMsgHolder
+    message.put(globalThis, ZigString.static("seq"), JSC.JSValue.jsNumber(child_singleton.seq));
```
Jarred-Sumner, line 47: i still don't like that we do this

these messages should be 2 int32's and nothing else, there's no variable-length or type data in these internal messages afaict

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,280 @@
...
+            const ex = arguments_[0];
+            Process__emitErrorEvent(globalThis_, ex);
+            return .undefined;
+        }
+    };
+
+    const good = ipc_instance.data.serializeAndSendInternal(globalThis, message);
+
+    if (!good) {
+        const ex = globalThis.createTypeErrorInstance("sendInternal() failed", .{});
```
Jarred-Sumner, line 68: this is also not great

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,280 @@
...
+        return .false;
+    }
+
+    return .true;
+}
+
+pub fn onInternalMessageChild(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
+    log("onInternalMessageChild", .{});
+    const arguments = callframe.arguments(2).ptr;
+    child_singleton.worker = JSC.Strong.create(arguments[0], globalThis);
```
Jarred-Sumner, line 81: does the callback truly need to be dynamic?

src/bun.js/node/node_cluster_binding.zig
```diff
@@ -0,0 +1,280 @@
...
+        this.dispatchUnsafe(message, globalThis);
+    }
+
+    fn dispatchUnsafe(this: *InternalMsgHolder, message: JSC.JSValue, globalThis: *JSC.JSGlobalObject) void {
+        const cb = this.cb.get().?;
+        const worker = this.worker.get().?;
+
+        const event_loop = globalThis.bunVM().eventLoop();
+
+        if (message.get(globalThis, "ack")) |p| {
```
Jarred-Sumner, line 132: this still seems not great, perf-wise.

these messages should be a binary encoding, two integers. all the callbacks should be passed in ahead of time and have no dynamic arguments, and kept alive by the module loader continuing to be alive. if we need to keep the individual worker instances alive then that would be the only necessary strong value to hold, and we only need one of them in total.

src/js/internal/cluster/Worker.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.process = options.process;
+    this.process.on("error", (code, signal) => this.emit("error", code, signal));
+    this.process.on("message", (message, handle) => this.emit("message", message, handle));
+  }
+}
+Worker.prototype = {};
+ObjectSetPrototypeOf(Worker.prototype, EventEmitter.prototype);
+
+Worker.prototype.kill = function () {
+  this.destroy.$apply(this, arguments);
```
Jarred-Sumner, line 30: Where is `destroy` defined?

src/js/internal/cluster/primary.ts
```diff
@@ -0,0 +1,318 @@
...
+// Extend generic Worker with methods specific to the primary process.
+Worker.prototype.disconnect = function () {
+  this.exitedAfterDisconnect = true;
+  send(this, { act: "disconnect" });
+  removeHandlesForWorker(this);
+  removeWorker(this);
+  return this;
+};
+
+Worker.prototype.destroy = function (signo) {
```
Jarred-Sumner, line 313: oh, in another file

src/output.zig
Jarred-Sumner: this change should still be reverted.

test/js/bun/http/serve.test.ts
```diff
@@ -1607,7 +1607,8 @@ it("should work with dispose keyword", async () => {
   expect(fetch(url)).rejects.toThrow();
 });
 
-it("should be able to stop in the middle of a file response", async () => {
+// prettier-ignore
+it.todoIf(isWindows)("should be able to stop in the middle of a file response", async () => {
```
Jarred-Sumner: revert this change please

src/io/PipeWriter.zig
```diff
@@ -1385,11 +1386,9 @@ pub fn WindowsStreamingWriter(
 
             this.is_done = true;
             this.closed_without_reporting = false;
-            // if we are done we can call close if not we wait all the data to be flushed
-            if (this.isDone()) {
-                if (!this.owns_fd) return;
-                this.close();
-            }
+            if (this.hasPendingData()) _ = this.flush();
```
Jarred-Sumner: `this.flush()` checks `is_done`. `is_done` is set to true 2 lines up.

therefore, the call to `.flush` does nothing

Jarred-Sumner: close() also sets is_done to true

src/js/node/http.ts
```diff
@@ -500,15 +511,73 @@ Server.prototype = {
       port = 0;
     }
 
+    if (typeof port === "string" && !Number.isNaN(parseInt(port))) {
```
Jarred-Sumner: we should not be parsing the port number twice

src/bun.js/bindings/bindings.zig
```diff
@@ -4615,6 +4639,12 @@ pub const JSValue = enum(JSValueReprInt) {
     pub inline fn isObject(this: JSValue) bool {
         return this.isCell() and this.jsType().isObject();
     }
+    pub inline fn isArray(this: JSValue) bool {
+        return this.isCell() and this.jsType().isArray();
+    }
+    pub inline fn isFunction(this: JSValue) bool {
+        return this.isCell() and this.jsType().isFunction();
```
Jarred-Sumner, line 4646: note that there is a distinction between is it a function, and is it callable

a proxy is not a function

but it might be callable.

src/js/node/http.ts
```diff
@@ -595,6 +668,13 @@ Server.prototype = {
           return promise;
         },
       });
+      // XXX: should this be turned into a bun.Server callback?
+      // node:http.Server#close only disables new connections, and the 'close' event is only emitted once all existing connections have closed.
+      getBunServerAllClosedPromise(this[serverSymbol]).then(() => {
```
Jarred-Sumner: shouldn't be a dynamically created function

Jarred-Sumner:
I've confirmed this doesn't cause a performance regression in node:http. 

There is a Windows-only test failure in node:http that does need to be addressed before the next version of Bun is released. 

But I think this is good enough to merge now.

PR #11499, opened 2024-05-31
fix(Bun.Socket) onDrain will only be called after handshake + docs update

cirospaciari:
…ome docs update

### What does this PR do?

- [x] Code changes

added a test

### DIFF
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -1225,8 +1225,6 @@ fn NewSocket(comptime ssl: bool) type {
             const handlers = this.handlers;
             const callback = handlers.onWritable;
             if (callback == .zero) return;
-            // if we are a SSL Socket and handshake is not done, we should not call the callback
-            if (ssl and this.authorized == false) return;
             var vm = handlers.vm;
             vm.eventLoop().enter();
             defer vm.eventLoop().exit();
--- a/src/bun.js/api/bun/socket.zig
+++ b/src/bun.js/api/bun/socket.zig
@@ -1225,6 +1225,8 @@ fn NewSocket(comptime ssl: bool) type {
             const handlers = this.handlers;
             const callback = handlers.onWritable;
             if (callback == .zero) return;
+            // if we SSL and handshake is not done, we should not call the callback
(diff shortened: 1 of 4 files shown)


Jarred-Sumner:
This should be at the uSockets layer right? If SSL handshaking is in progress the socket isn’t truly writable

cirospaciari:
> This should be at the uSockets layer right? If SSL handshaking is in progress the socket isn’t truly writable

you are right, done in usockets layer now

src/bun.js/api/bun/socket.zig
```diff
@@ -1519,6 +1519,9 @@ fn NewSocket(comptime ssl: bool) type {
             // you should use getAuthorizationError and authorized getter to get those values in this case
             if (is_open) {
                 result = callback.callWithThis(globalObject, this_value, &[_]JSValue{this_value});
+                // clean onOpen callback so only called in the first handshake and not in every renegotiation
+                this.handlers.onOpen.unprotect();
```
Jarred-Sumner: This will break for servers

cirospaciari: fixed

test/js/bun/net/socket.test.ts
```diff
@@ -477,3 +477,24 @@ it("should connect directly when using an ip address", async () => {
...
+          reject(new Error("Socket not authorized"));
+        }
+      },
+      handshake() {
+        resolve();
+      },
+    },
+  });
+  await promise;
+  expect(socket.authorized).toBe(true);
```
Jarred-Sumner, line 499:
```suggestion
  expect(socket.authorized).toBe(true);
```

test/js/bun/net/socket.test.ts
```diff
@@ -477,3 +477,24 @@ it("should connect directly when using an ip address", async () => {
   await promise;
   expect(opened).toBe(true);
 });
+
+it("should not call drain before handshake", async () => {
+  const { promise, resolve, reject } = Promise.withResolvers();
+  const socket = await Bun.connect({
```
Jarred-Sumner:
```suggestion
  using socket = await Bun.connect({
```

PR #11503, opened 2024-05-31
fix `bun create` test `should create template from local folder`

dylan-conway:
### What does this PR do?
Directory should have `.iterate = true`.

Adds code for copying local folders on windows.

manually and updated test


src/cli/create_command.zig
```diff
@@ -456,7 +456,7 @@ pub const CreateCommand = struct {
                 node.name = "Copying files";
                 progress.refresh();
 
-                const template_dir = std.fs.openDirAbsolute(filesystem.abs(&template_parts), .{}) catch |err| {
```
Jarred-Sumner, line 459: I don't think `.iterate` does anything on macOS

I think this is unlikely to fix the issue

src/cli/create_command.zig
```diff
@@ -271,8 +271,8 @@ pub const CreateCommand = struct {
         const destination = try filesystem.dirname_store.append([]const u8, resolve_path.joinAbs(filesystem.top_level_dir, .auto, dirname));
 
         var progress = std.Progress{};
-        var node = progress.start(try ProgressBuf.print("Loading {s}", .{template}), 0);
```
dylan-conway, line 274: might need to do this everywhere, not sure yet

src/cli/create_command.zig
dylan-conway: > I don't think `.iterate` does anything on macOS
> 
> I think this is unlikely to fix the issue

You can repro this issue on linux and windows, but it does not fix the flakiness on macos

paperclover: the directory is iterated. iterate should be passed

PR #11509, opened 2024-06-01
fix sourcemap printing with multiple chunks sharing the same file

paperclover:
Fixes another sourcemap issue.
```
const a = new TextEncoder();
console.log('hey!')
const d = new TextEncoder();

// comment the line out to fix
const b = { hello: 'world' };

const c = new Set([
]);
console.log('hey!')
console.log('hey!')
console.log('hey!')
console.log('hey!')
```

Something with tree-shaking makes it so when `const b` is stripped, the file is split into two chunks, these two chunks mess up the sourcemap as the code was plain wrong. I am a little surprised the mistake didn't get caught in earlier testing.

Additionally, debug builds now dump runtime sourcemaps for all files in /tmp/bun-debug-src, alongside the transpiled files. I will be using this to resolve at-runtime related sourcemapping issues.

### DIFF
--- a/src/bun.js/module_loader.zig
+++ b/src/bun.js/module_loader.zig
@@ -154,11 +154,17 @@ inline fn jsSyntheticModule(comptime name: ResolvedSource.Tag, specifier: String
 ///
 /// This can technically fail if concurrent access across processes happens, or permission issues.
 /// Errors here should always be ignored.
-fn dumpSource(specifier: string, printer: anytype) void {
-    dumpSourceString(specifier, printer.ctx.getWritten());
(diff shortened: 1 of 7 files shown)


src/bun.js/module_loader.zig
```diff
@@ -195,15 +198,42 @@ fn dumpSourceString(specifier: string, written: []const u8) void {
...
+            defer file.close();
+
+            const source_file = parent.readFileAlloc(
+                bun.default_allocator,
+                specifier,
+                std.math.maxInt(u64),
+            ) catch "";
+
+            const w = file.writer();
+            try w.print(
```
Jarred-Sumner, line 221: I know it's debug but a buffered writer here would still be good

PR #11512, opened 2024-06-01
Fix port exhaustion issue

Jarred-Sumner:
### What does this PR do?

WIP.

### DIFF
--- a/packages/bun-usockets/src/bsd.c
+++ b/packages/bun-usockets/src/bsd.c
@@ -490,28 +490,22 @@ inline __attribute__((always_inline)) LIBUS_SOCKET_DESCRIPTOR bsd_bind_listen_fd
     int options
 ) {
 
-    if (port != 0) {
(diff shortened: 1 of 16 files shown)


packages/bun-usockets/src/bsd.c
```diff
@@ -511,6 +511,11 @@ inline __attribute__((always_inline)) LIBUS_SOCKET_DESCRIPTOR bsd_bind_listen_fd
         setsockopt(listenFd, SOL_SOCKET, SO_REUSEADDR, (void *) &enabled, sizeof(enabled));
 #endif
 
+    } else {
+        #if !defined(_WIN32)
```
cirospaciari: I think this part can be simplified a lot optval or enabled are the both `int`, the only variation is `SO_EXCLUSIVEADDRUSE`, `SO_REUSEADDR` and `SO_REUSEPORT`,
also is not clear why windows is skipped on the else condition.

Issue #11427, opened 2024-05-28
Missing Date header in Bun.file response


7f8ddd:
### What version of Bun is running?

1.1.10-canary.1+34b02ada1

### What platform is your computer?

Linux 6.8.0-31-generic x86_64 x86_64

### What steps can reproduce the bug?

```ts
Bun.serve({
  fetch(req: Request) {
    const file = Bun.file("test.css");

    // The response *will* include the Date header.
    return new Response(file, {
      headers: {
        "hello-world": "yes",
      },
    });

    // The response *will not* include the Date header.
    return new Response(file);
  },

  port: 8080,
});

```

### Additional information

The RFC 7231 for HTTP/1.1 states that the Date header should always be included except for in these cases:
> If the response status code is 100 (Continue) or 101 (Switching Protocols).
> If the response has no content-body and the request method was HEAD."

Where it's being added:
https://github.com/oven-sh/bun/blob/912bccb6243bbc169c0d92e23dc1be7f01e44fba/packages/bun-uws/src/HttpResponse.h#L77

It seems like the Bun.file path with no additional headers is stripping/bypassing that code.

bogeychan:
### Additional information

On **Windows** (`bun --version 1.1.10`), both responses *will* contain the `Date` header

PR #11542, opened 2024-06-02
Make HTTP headers consistently cased

7f8ddd:
### What does this PR do?

Makes the casing of the HTTP headers consistent with other applications and code in Bun. It feels wrong, especially when some of them, such as the uWS Date header, are cased differently. From a security perspective, the casing could also be used as a side-channel to determine if a server is running Bun or not.

- [x] Code changes

Unknown, may fail tests that care about capitalization, but I'll fix that.

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -1500,16 +1500,16 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                         resp.writeStatus("200 OK");
                         ctx.flags.has_written_status = true;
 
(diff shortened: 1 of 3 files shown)


mangs:
in the past ive noticed that all headers get lowercased in Bun. i think there may be code that is lowercasing in addition to the changes here.

7f8ddd:
> in the past ive noticed that all headers get lowercased in Bun. i think there may be code that is lowercasing in addition to the changes here.

I'll make it a draft and maybe fix https://github.com/oven-sh/bun/issues/11427 as well.

PR #11549, opened 2024-06-03
fix(install): manifest package name mismatch

dylan-conway:
### What does this PR do?
If the registry encodes the `name` field of the package manifest, we need to check for `%40` and `%2f` in place of `@` and `/`.

- [ ] Code changes

added a test with package manifest names: `@url%2fencoding` and `%40url%2fencoding`


Jarred-Sumner (approved):
Looks good. Hoping we don’t need to support this in the package.json name as well.

PR #11562, opened 2024-06-03
Improve the error message when Bun.serve() is passed a non-Response object

Jarred-Sumner:
### What does this PR do?

Before:
```
error: Expected a Response object
```

After:
```
error: Expected a Response object, but received '{}'
```

Sometimes it will also print a stack trace but it's kind of hard to do that since it's usually the top of the call stack at this point.


src/bun.js/bindings/bindings.zig
```diff
@@ -5595,15 +5595,11 @@ pub const JSValue = enum(JSValueReprInt) {
     extern fn JSC__JSValue__getClassInfoName(value: JSValue, out: *bun.String) bool;
 
     /// Returned memory is read-only memory of the s_info assigned to a JSCell
```
paperclover: this comment needs to be updated because it no longer makes any sense

PR #11567, opened 2024-06-03
fix: create_command telling to 'cd' when it should be omitted or `cd .`

### LINKED ISSUE #11566: `bun create` telling wrong path to cd into

1.1.10+5102a9443

Linux 6.9.2-arch1-1 x86_64 unknown

Create a new bun project using the `bun create` command for example do `bun create elysia .`

When we use `bun create`, it shows a `# To get started, run:` section, something like this :

- Here i created a new `elysia` project usign `bun create elysia app`

image

- but instead if i use `bun create elysia .`, telling it to create the project in the current directory and not a new directory then it shows 

image

which is wrong, doing `cd` will take us to the `$HOME` dir

### What do you see instead?

Wrong path to cd into

image

### Additional information

_No response_

Sushants-Git:
### What does this PR do?

Fixes #11566

Omits the `cd` line from the `# To get started, run:` section, when using the `bun create` command for creating a project in the current directory, instead of telling the user to do `cd` and sending them to the `$HOME` dir.

- [x] Code changes

it's a simple condition change, so i did not run any tests on it.

### DIFF
--- a/src/cli/create_command.zig
+++ b/src/cli/create_command.zig
@@ -1548,18 +1548,33 @@ pub const CreateCommand = struct {
             , .{create_react_app_entry_point_path});
         }
 
-        Output.pretty(
-            \\
-            \\<d>#<r><b> To get started, run:<r>
-            \\
-            \\  <b><cyan>cd {s}<r>
-            \\  <b><cyan>{s}<r>
-            \\
-            \\
-        , .{
-            filesystem.relativeTo(destination),
-            start_command,
-        });
+        const isEmptyDestination = std.mem.eql(u8, filesystem.relativeTo(destination), "");
+
+        if (isEmptyDestination) {
+            Output.pretty(
+                \\
+                \\<d>#<r><b> To get started, run:<r>
+                \\
+                \\  <b><cyan>{s}<r>
+                \\
+                \\
+            , .{
+                start_command,
+            });
+        } else {
+            Output.pretty(
+                \\
+                \\<d>#<r><b> To get started, run:<r>
+                \\
+                \\  <b><cyan>cd {s}<r>
+                \\  <b><cyan>{s}<r>
+                \\
+                \\
+            , .{
+                filesystem.relativeTo(destination),
+                start_command,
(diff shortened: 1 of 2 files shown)


dylan-conway (commented):
Could you add a test for this in `bun-create.test.ts`? It could be a slightly modified version of `"should create template from local folder"`.

Also the failure on darwin-x64 is unrelated. It's an existing failure

src/cli/create_command.zig
```diff
@@ -1548,18 +1548,33 @@ pub const CreateCommand = struct {
...
-            \\
-            \\  <b><cyan>cd {s}<r>
-            \\  <b><cyan>{s}<r>
-            \\
-            \\
-        , .{
-            filesystem.relativeTo(destination),
-            start_command,
-        });
+        const isEmptyDestination = std.mem.eql(u8, filesystem.relativeTo(destination), "");
```
dylan-conway:
```suggestion
        const rel_destination = filesystem.relativeTo(destination);
        const is_empty_destination = rel_destination.len == 0;
```
then use `rel_destination` in `Output.pretty` below.

Sushants-Git:
> Could you add a test for this in `bun-create.test.ts`? It could be a slightly modified version of `"should create template from local folder"`.
> 
> Also the failure on darwin-x64 is unrelated. It's an existing failure

yep got it, will do!

test/cli/install/bun-create.test.ts
```diff
@@ -107,6 +107,24 @@ it("should create template from local folder", async () => {
   expect(dirStat.isDirectory()).toBe(true);
 });
 
+it("should not mention cd prompt when created in current directory", async () => {
```
Sushants-Git, line 110: @dylan-conway wrote this test, for some reason `test-template` wasn't giving any output so i decided to go with creating a `create-test` same as `should create and install github template from`, which worked.

lemme know what changes i should make to this

dylan-conway (approved):
Looks great, thank you!

PR #11575, opened 2024-06-03
fix(install): `bun update` keeps pinned versions

dylan-conway:
### What does this PR do?
`bun update` will keep `~` or `^` if it's used in the version. Previously only `^` would stay if used.

Also fixes a bug where `--latest` would not apply to packages passed to `bun update`.

added tests

### DIFF
--- a/test/cli/install/bun-update.test.ts
+++ b/test/cli/install/bun-update.test.ts
@@ -26,250 +26,268 @@ expect.extend({
   toHaveBins,
 });
 
-it("should update to latest version of dependency", async () => {
-  const urls: string[] = [];
-  const registry = {
-    "0.0.3": {
+for (const { input } of [{ input: { baz: "~0.0.3", moo: "~0.1.0" } }, { input: { baz: "^0.0.3", moo: "^0.1.0" } }]) {
+  it(`should update to latest version of dependency (${input.baz[0]})`, async () => {
+    const urls: string[] = [];
+    const tilde = input.baz[0] === "~";
+    const registry = {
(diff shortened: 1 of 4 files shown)


dylan-conway:
`overrides.test.ts` failure:
Screenshot 2024-06-03 at 3 11 22 PM

test/cli/install/bun-update.test.ts
```diff
@@ -265,7 +265,7 @@ it("should update to latest versions of dependencies", async () => {
     name: "foo",
     dependencies: {
       "@barn/moo": "^0.1.0",
-      baz: "^0.0.5",
+      baz: "~0.0.5",
```
Jarred-Sumner: Can we run this test for both `~0.0.5` and `^0.0.5`?

test/cli/install/bun-update.test.ts
```diff
@@ -133,7 +133,7 @@ it("should update to latest version of dependency", async () => {
   expect(await file(join(package_dir, "package.json")).json()).toEqual({
     name: "foo",
     dependencies: {
-      baz: "^0.0.5",
+      baz: "~0.0.5",
```
Jarred-Sumner: Can we run this test for both `~0.0.5` and `^0.0.5`?

dylan-conway: 👍

test/cli/install/bun-update.test.ts
dylan-conway: 👍

PR #11579, opened 2024-06-03
Fix speculative race in fetch

Jarred-Sumner:
### What does this PR do?

Fix speculative race in fetch


cirospaciari (approved):

cirospaciari:
lets merge after tests run

PR #11581, opened 2024-06-04
Fix runtime stack trace computation

### LINKED ISSUE #8880: assert points to the wrong line

1.0.26+c75e768a6

Darwin 23.2.0 arm64 arm

This file:

```ts
import assert from 'assert'

function calc() {
  return 1
}

assert.strictEqual(1, 2)
```

assertion error points to the right line

### What do you see instead?

Screenshot 2024-02-13 at 04 50 19

### Additional information

_No response_

### LINKED ISSUE #9302: Source maps and error locations are broken

paperclover:
Fixes #9302
Fixes #8880

After the previous adventure of source-mapping related issues, it seems the big issue blocking sourcemaps from working is simply our code reading line+column numbers from JavaScriptCore.

The original code
```diff
-   /* Get the "raw" position info.
-    * Note that we're using m_codeBlock->unlinkedCodeBlock()->expressionRangeForBytecodeOffset
-    * rather than m_codeBlock->expressionRangeForBytecodeOffset in order get the "raw" offsets and
-    * avoid the CodeBlock's expressionRangeForBytecodeOffset modifications to the line and column
-    * numbers, (we don't need the column number from it, and we'll calculate the line "fixes"
-    * ourselves). */
-   ExpressionInfo::Entry info = m_codeBlock->unlinkedCodeBlock()->expressionInfoForBytecodeIndex(bytecodeOffset);
```

Is an alright idea, but the code that handles fixing this raw info does not work, with cases as simple as:

```
// @bun
var l = Object.create; (() => { throw new Error() })();
```

Instead we will simply use `expressionInfoForBytecodeIndex` on the linked code block to retrive all of the computed offsets.

```diff
+auto expr = m_codeBlock->expressionInfoForBytecodeIndex(bytecodeOffset);
```

And to prevent zero-based vs one-based, I am using `OrdinalNumber` practically everywhere to ensure these numbers are interpretted correctly.

Draft PR because there are surely test failures which will need to be corrected with proper location info, but I am noticing some potential mistakes in regards to oneBasedInt and zeroBasedInt

### DIFF
--- a/src/bun.js/bindings/ErrorStackFrame.cpp
+++ b/src/bun.js/bindings/ErrorStackFrame.cpp
@@ -68,9 +68,9 @@ ZigStackFramePosition getAdjustedPositionForBytecode(JSC::CodeBlock* code, JSC::
 
     /// JavaScriptCore places error divots at different places than v8
     // Uncomment to debug this:
-    printf("lc = %d : %d (byte = %d)\n", pos.line.oneBasedInt(), pos.column.oneBasedInt(), expr.divot);
-    printf("off = %d : %d\n", expr.startOffset, expr.endOffset);
-    printf("name = %s\n", inst->name());
+    // printf("lc = %d : %d (byte = %d)\n", pos.line.oneBasedInt(), pos.column.oneBasedInt(), expr.divot);
+    // printf("off = %d : %d\n", expr.startOffset, expr.endOffset);
+    // printf("name = %s\n", inst->name());
 
     switch (inst->opcodeID()) {
     case op_construct:
--- a/src/bun.js/bindings/ErrorStackFrame.cpp
+++ b/src/bun.js/bindings/ErrorStackFrame.cpp
@@ -1,0 +1,91 @@
+#include "root.h"
+#include "JavaScriptCore/CodeBlock.h"
+#include "headers-handwritten.h"
+#include "JavaScriptCore/BytecodeIndex.h"
+#include "wtf/Assertions.h"
+#include "wtf/text/OrdinalNumber.h"
+
+namespace Bun {
+using namespace JSC;
+
+/// Adjust a `ZigStackFramePosition` by a number of bytes. This accounts for when the adjustment
+/// crosses line boundaries, and thus requires the source code in order to properly compute
+/// the result.
+void adjustPositionBackwards(ZigStackFramePosition& pos, int amount, CodeBlock* code)
+{
+    if (pos.byte_position - amount < 0) {
+        pos.line = OrdinalNumber::fromZeroBasedInt(0);
+        pos.column = OrdinalNumber::fromZeroBasedInt(0);
+        pos.byte_position = 0;
+        return;
+    }
+
+    pos.column = OrdinalNumber::fromZeroBasedInt(pos.column.zeroBasedInt() - amount);
+    if (pos.column.zeroBasedInt() < 0) {
+        auto source = code->source().provider()->source();
+        if (!source.is8Bit()) {
+            // Debug-only assertion
+            // Bun does not yet use 16-bit sources anywhere. The transpiler ensures everything
+            // fit's into latin1 / 8-bit strings for on-average lower memory usage.
+            ASSERT_NOT_REACHED("16-bit source re-mapping is not implemented here.");
+
+            pos.line = OrdinalNumber::fromZeroBasedInt(0);
+            pos.column = OrdinalNumber::fromZeroBasedInt(0);
+            pos.byte_position = 0;
+            return;
+        }
+
+        for (int i = 0; i < amount; i++) {
+            if (source[pos.byte_position - i] == '\n') {
+                pos.line = OrdinalNumber::fromZeroBasedInt(pos.line.zeroBasedInt() - 1);
+            }
+        }
+
+        int columns = 0;
+        // Initial -1 to skip the newline that gets counted.
+        int i = pos.byte_position - amount - 1;
+        while (i > 0 && source[i] != '\n') {
+            columns += 1;
+            i -= 1;
+        }
+        pos.column = OrdinalNumber::fromZeroBasedInt(columns);
+    }
+
+    pos.byte_position -= amount;
+}
+
+ZigStackFramePosition getAdjustedPositionForBytecode(JSC::CodeBlock* code, JSC::BytecodeIndex bc)
+{
+    auto expr = code->expressionInfoForBytecodeIndex(bc);
+
+    ZigStackFramePosition pos {
+        .line = OrdinalNumber::fromOneBasedInt(expr.lineColumn.line),
+        .column = OrdinalNumber::fromOneBasedInt(expr.lineColumn.column),
+        .byte_position = (int)expr.divot,
+    };
+
+    auto inst = code->instructionAt(bc);
+
+    /// JavaScriptCore places error divots at different places than v8
+    // Uncomment to debug this:
+    printf("lc = %d : %d (byte = %d)\n", pos.line.oneBasedInt(), pos.column.oneBasedInt(), expr.divot);
+    printf("off = %d : %d\n", expr.startOffset, expr.endOffset);
+    printf("name = %s\n", inst->name());
+
+    switch (inst->opcodeID()) {
+    case op_construct:
+    case op_construct_varargs:
+        // The divot by default is pointing at the `(` or the end of the class name.
+        // We want to point at the `new` keyword, which is conveniently at the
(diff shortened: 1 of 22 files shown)


paperdave:
TODO: this regresses this behavior
image

paperdave:
This seems to move the column number from the `new` keyword to the `Error` function name.

image

paperdave:
> This seems to move the column number from the `new` keyword to the `Error` function name.

This one is actually insane.
image
Look at this sourcemap, if the `Error(` is one source mapping token, then it means a divot pointing at the ( will remap to point at the `E` in Error, which is what we observe.

The old logic was either
- bugged in just the right way to work
- we had an off-by-one token issue

I'm going to work on getting the sourcemap columns on function calls to just *work*. Matching v8 is so close i can taste it.

src/bun.js/bindings/headers-handwritten.h
```diff
@@ -138,14 +139,9 @@ const ZigStackFrameCode ZigStackFrameCodeWasm = 5;
...
 typedef struct ZigStackFramePosition {
-    int32_t source_offset;
-    int32_t line;
-    int32_t line_start;
-    int32_t line_stop;
-    int32_t column_start;
-    int32_t column_stop;
-    int32_t expression_start;
-    int32_t expression_stop;
+    WTF::OrdinalNumber line;
```
Jarred-Sumner: Is it safe to use `WTF::OrdinalNumber` here? It won't be initialized as an C++ `WTF::OrdinalNumber` in Zig. I think this should be stored as an int32_t and then a method could be added to convert it to a WTF::OrdinalNumber.

src/bun.js/bindings/headers-handwritten.h
```diff
@@ -138,14 +139,9 @@ const ZigStackFrameCode ZigStackFrameCodeWasm = 5;
...
+    WTF::OrdinalNumber column;
+    int byte_position;
```
Jarred-Sumner: Let's be more specific so there's no ambiguity:
```suggestion
    int32_t byte_position;
```

src/bun.js/bindings/headers-handwritten.h
```diff
@@ -158,7 +154,7 @@ typedef struct ZigStackFrame {
 
 typedef struct ZigStackTrace {
     BunString* source_lines_ptr;
-    int32_t* source_lines_numbers;
+    OrdinalNumber* source_lines_numbers;
```
Jarred-Sumner, line 167: We can't share C++ classes in Zig
```suggestion
    int32_t* source_lines_numbers;
```

Jarred-Sumner (commented):
A couple nitpicky comments but this is a good improvement and once the tests pass & the comments are addressed, we should merge

src/bun.js/bindings/headers-handwritten.h
paperclover: yeah lets do this

src/bun.js/bindings/ErrorStackFrame.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if (pos.byte_position - amount < 0) {
+        pos.line_zero_based = 0;
+        pos.column_zero_based = 0;
+        pos.byte_position = 0;
+        return;
+    }
+
+    pos.column_zero_based = pos.column_zero_based - amount;
+    if (pos.column_zero_based < 0) {
+        auto source = code->source().provider()->source();
```
Jarred-Sumner, line 25:
```suggestion
        const auto *provider = code->source().provider();
        if (UNLIKELY(!provider)) {
           return;
        }
        const auto& source = provider->source();
```

PR #11585, opened 2024-06-04
Fixes #7263

### LINKED ISSUE #7263: ReferenceError: Cannot access uninitialized variable.

1.0.13+f5bf67bd1

Darwin 22.6.0 arm64 arm

Create the file `foo.ts` with contents:

```
const Infinity = 1/0;
```
then run `bun foo.ts` and it will exit and print this error:

```
ReferenceError: Cannot access uninitialized variable.
```

No crash

### What do you see instead?

Crash

### Additional information

The line `const Infinity = 1/0;` is used in the `@ltd/j-toml` package and when that package is imported it crashes when run with bun.

Jarred-Sumner:
### What does this PR do?

This makes Bun print `Infinity` as `1 / 0` when either (1) we are not renaming symbols, like at runtime or (2) when bundling with --minify-syntax.

Fixes #7263

Loosely based on https://github.com/evanw/esbuild/pull/1385

Since this is an observable output change, this bumps the runtime transpiler cache version.

There are some tests


dylan-conway (approved):
Nice, code and tests look good

dylan-conway:
Looks like expected test failures from output change

Issue #5029, opened 2023-09-12
bun install incorrectly parses custom registries


zacowan:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. Create a `bunfig.toml` file in a project directory
2. Provide a custom registry URL:

```
[install]

registry = "https://www.custom-registry.com/one/two/three/four"
```

3. Attempt to install a package: `bun install react`
4. Note that bun interprets the installation URL in error messages as `www.custom-registry.com/one/two/three/react`, which omits the final part of the URL `four`

### What is the expected behavior?

In the same scenario as above, bun should interpret the installation URL as `www.custom-registry.com/one/two/three/four/react`.

### What do you see instead?

`www.custom-registry.com/one/two/three/react`

zacowan:
Note: adding a trailing slash to the registry URL seems to solve this issue for the specific case outlined above.

Jarred-Sumner:
This was fixed some time ago. We normalize the registry URLs nowadays.

If you're still running into issues like this, please leave a comment and we'll re-open.

PR #11606, opened 2024-06-04
fix(install): invalidate manifest cache on registry change

### LINKED ISSUE #11582: Not being able to use bun install in large companies

1.1.12

Darwin 23.5.0 x86_64 i386

1. You need to set the registry setting in bunfig.toml pointing to your own registry (for all packages, no exceptions). In our case it is artifactory, but it doesn't seem to matter much.
2. On a relatively large project execute `bun install`

There are no entries in bun.lockdb with registry=registry.npmjs.org. Only with the one specified in bunfig.toml in the install.registry section.

### What do you see instead?

With 100% probability there will be several entries in bun.lockdb, which will contain registry=registry.npmjs.org. And this is a blocker for switching to bun in local development and at the pipelines level, taking advantage of its speed advantage as a package manager.

### Additional information

I would like to highlight the importance of one issue that prevents bun from being used as a package manager in all large companies and corporations.

The reason is that large companies are security-conscious and use internal package repositories, and not infrequently, they also use network-level blocking of public repositories.

The problem is as follows: 
1. You need to set the registry setting in bunfig.toml pointing to your own registry (for all packages, no exceptions). In our case it is artifactory, but it doesn't seem to matter much.
2. On a relatively large project execute `bun install`
3. With 100% probability there will be several entries in bun.lockdb, which will contain registry=registry.npmjs.org. And it's a blocker to switch to bun, taking advantage of its speed advantage as a package manager.

The nature of the error seems to lie in some tricky race condition, as most entries are created correctly, but a few are bound to end up with the wrong entry, causing those packages to be downloaded from the wrong source, which will simply be unavailable on corporate networks.

Related issues:
- #11434
- #9804
- #7601
- #6347
- #5722
- #5721
- #5370
- #5029 (only in the aspect of parsing the registry setting)
- #4901

dylan-conway:
### What does this PR do?
We need to make sure manifests in the cache are invalidated when the registry for the package changes. This pr bumps our serialized manifest version to 3, and writes the registry hash + length to end of the file.

fixes #11582 

Added a test that installs the same package from two registries.


Jarred-Sumner:
I think we should store the length also

Jarred-Sumner:
Like hash + length

As another sanity check

Jarred-Sumner:
Let’s merge once tests run

anru:
Thank you very much for this fix, looking forward to the next release to adopt bun as a package manager in our day 2 day development and CI/CD pipelines.

It is also highly likely that the nature of all the related issues listed in #11582 (perhaps except the #5029) lies in the behaviour that was fixed in current mr.

PR #11619, opened 2024-06-05
More Worker Fixes

Jarred-Sumner:
### What does this PR do?

- Prevent calling into JavaScript in `atexit` handlers 
- Add some tests for worker destruction in fetch
- Fix bug where we may unnecessarily enter the event loop to handle `beforeExit` when `beforeExit` was never listened to


src/bun.js/webcore/response.zig
```diff
@@ -1520,6 +1528,10 @@ pub const Fetch = struct {
                 // Note: We cannot call .get() on the ReadableStreamRef. This is called inside a finalizer.
                 if (body.value != .Locked or this.readable_stream_ref.held.has()) {
                     // Scenario 1 or 3.
+
+                    if (vm.isShuttingDown()) {
+                        this.ignoreRemainingResponseBody();
```
cirospaciari, line 1533: we should add `scenario 4 here`, resolving the promise when `vm.isShuttingDown() == true` is ok? because we would be touching JS again in that case, if is not we should add this verification at the top

test/js/node/worker_threads/worker_thread_check.ts
```diff
@@ -81,9 +98,17 @@ if (isMainThread) {
       break;
     }
     case "fetch": {
+      await fetch("http://localhost:" + port);
```
cirospaciari, line 101: we should add cases using `fetch(url).then(res=> res.arrayBuffer())` so we know that closes clean if we are waiting some promise/buffering in the fetch

cirospaciari (changes requested):

src/bun.js/javascript.zig
```diff
@@ -515,7 +515,7 @@ pub const ExitHandler = struct {
         vm.exit_handler.exit_code = code;
     }
 
-    extern fn Process__dispatchOnBeforeExit(*JSC.JSGlobalObject, code: u8) void;
+    extern fn Process__dispatchOnBeforeExit(*JSC.JSGlobalObject, code: u8) c_int;
```
paperclover: note: int32_t in c++ is not zig c_int

test/js/node/worker_threads/worker_thread_check.ts
```diff
@@ -81,9 +98,30 @@ if (isMainThread) {
...
+      break;
+    }
+    case "fetch+blob": {
       const resp = await fetch("http://localhost:" + port);
       await resp.blob();
       break;
     }
+    case "fetch+blob-early-exit": {
+      const resp = await fetch("http://localhost:" + port);
+      await resp.blob();
```
Jarred-Sumner, line 115:
```suggestion
      resp.blob();
```

test/js/node/worker_threads/worker_thread_check.ts
```diff
@@ -81,9 +98,30 @@ if (isMainThread) {
...
+      break;
+    }
+    case "readFile": {
+      await Bun.file(import.meta.path).text();
+      break;
+    }
+    case "readFile-early-exit": {
+      await Bun.file(import.meta.path).text();
```
Jarred-Sumner, line 123:
```suggestion
      Bun.file(import.meta.path).text();
```

Issue #11388, opened 2024-05-27
Fix `fs.watchFile` and `process` tests on Windows


Jarred-Sumner:
I believe the cause for both is related to uv ref/unref counts. libuv thinks there are no active handles, so it doesn't check the list of active handles. Since it doesn't check the list of active handles, the callbacks are never called. Since the callbacks are never called, the tests hang or they flake. It's a similar situation as in #11385

PR #11669, opened 2024-06-07
fix(fs.watchFile)

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/11388

- [x] Code changes

Existing tests


src/bun.js/node/node_fs_stat_watcher.zig
```diff
@@ -36,45 +36,102 @@ pub const StatWatcherScheduler = struct {
     timer: ?*uws.Timer = null,
 
     head: std.atomic.Value(?*StatWatcher) = .{ .raw = null },
```
Jarred-Sumner: `UnboundedQueue` should handle the queue of stat watchers and that will simplify this code a lot

PR #11671, opened 2024-06-07
add `memrchr` to `strings.lastIndexOfChar` on linux

dylan-conway:
CI


Jarred-Sumner:
nice, looks like glibc does implement this in assembly so it should be a perf improvement https://elixir.bootlin.com/glibc/latest/A/ident/memrchr

Jarred-Sumner (approved):

PR #11672, opened 2024-06-07
fix(Bun.serve) keep JS signal alive when not using async fetch

cirospaciari:
### What does this PR do?

- [x] Code changes

Added tests

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -1407,7 +1407,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         /// this prevents an extra pthread_getspecific() call which shows up in profiling
         allocator: std.mem.Allocator,
         req: *uws.Request,
-        signal: ?*JSC.WebCore.AbortSignal = null,
+        signal: JSC.Strong = .{},
         method: HTTP.Method,
 
         flags: NewFlags(debug_mode) = .{},
@@ -1445,6 +1445,13 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
 
         // TODO: support builtin compression
         const can_sendfile = !ssl_enabled and !Environment.isWindows;
+
+        pub fn getSignal(this: *const RequestContext) ?*JSC.WebCore.AbortSignal {
+            if (this.signal.get()) |js_signal| {
+                return js_signal.as(JSC.WebCore.AbortSignal);
+            }
+            return null;
+        }
 
         pub inline fn isAsync(this: *const RequestContext) bool {
             return this.defer_deinit_until_callback_completes == null;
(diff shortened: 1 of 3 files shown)


src/bun.js/api/server.zig
```diff
@@ -6165,6 +6173,11 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
             defer {
                 // uWS request will not live longer than this function
                 request_object.request_context = JSC.API.AnyRequestContext.Null;
+                // JS signal is lazy created when needed and we need to keep it alive so we can call the events
+                const js_signal = RequestPrototype__signalGetCachedValue(request_value);
```
Jarred-Sumner: this will crash if there's a pending exception

```suggestion
                const js_signal = if (!globalThis.hasPendingException()) RequestPrototype__signalGetCachedValue(request_value) else JSValue.empty;
```

cirospaciari: fixed here: https://github.com/oven-sh/bun/pull/11672/commits/3732a78555ae6780871cdf6ff6b23ddf89c49adb

cirospaciari: Handled here:
https://github.com/oven-sh/bun/pull/11672/files#diff-e9484fda309708f80db81a125281dd0b9fdadaf3d2400ae8e78a2b8fc14d85ebR6173-R6180

src/bun.js/api/server.zig
```diff
@@ -5119,7 +5127,7 @@ pub const ServerWebSocket = struct {
         return ZigString.init(text).toValueGC(globalThis);
     }
 };
-
+extern fn RequestPrototype__signalGetCachedValue(JSC.JSValue) JSC.JSValue;
```
Jarred-Sumner: Can you add a method to `Request` instead of duplicating this method in `server.zig`?

src/bun.js/api/server.zig
```diff
@@ -6162,6 +6170,14 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
             request_value.ensureStillAlive();
 
             const response_value = this.config.onRequest.callWithThis(this.globalThis, this.thisObject, &args);
+            response_value.ensureStillAlive();
+            if (!response_value.isAnyError()) {
+                // JS signal is lazy created when needed and we need to keep it alive so we can call the events
+                const js_signal = RequestPrototype__signalGetCachedValue(request_value);
```
Jarred-Sumner: `
```suggestion
                const js_signal = Request.signalGetCachedValue(request_value);
```

src/bun.js/api/server.zig
```diff
@@ -6162,6 +6170,14 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
...
+                if (!js_signal.isEmptyOrUndefinedOrNull()) {
+                    ctx.js_signal = JSC.Strong.create(js_signal, this.globalThis);
```
Jarred-Sumner: What do browsers do here? This feels strange to me

cirospaciari: `Request` owns signal (strong ref), in our case we ref the internal struct but not the JS counter part.
Another way of doing this is always strong ref inside Request when we create the signal, and in the server.zig side we ref count the request it self.

Jarred-Sumner:
I feel like it'd be better to use opaque roots for this. The RequestContext is an opaque root, or in fetch() which may have the same problem, the FetchTasklet is an opaque root.

PR #11679, opened 2024-06-07
Update install.sh to try ~/.bash_profile first for PATH modification

btcbobby:
I suggest swapping lines 254 and 255 for modifying PATH under Bash so that .bash_profile is tried first, before .bashrc, as .bash_profile is where PATH is normally set.

### What does this PR do?

Initializes $bash_configs with ~/.bash_profile as the first element instead, bumping ~/.bashrc from first to second. This is important because the following loop uses this order to search for files to modify and stops after the first one found.

- [ ] Code changes

I didn't run any tests for such a small change.

Only a single line change in install.sh, no Zig or Typescript changes

### DIFF
--- a/src/cli/install.sh
+++ b/src/cli/install.sh
@@ -263,8 +263,8 @@ bash)
     )
 
     bash_configs=(
-        "$HOME/.bashrc"
         "$HOME/.bash_profile"
+        "$HOME/.bashrc"
     )
 
     if [[ ${XDG_CONFIG_HOME:-} ]]; then


markovejnovic:
I am comfortable with this change since it is a net improvement over `.bashrc` as it enables non-interactive mode to have `bun` in its `PATH`.

We should, however, amend this script to use `.profile`, since that is most likely to be the correct location which will propagate to **all** shells.

I'm not expecting you to do this though @btcbobby and thank you very much for your PR.

markovejnovic:
This seems like a good PR to me. Overall net improvement. Tested locally and it LGTM.

@Jarred-Sumner, why don't we patch `PATH` in `.profile` so it is propagated to all shells? That would be less code and would make `install.sh` cleaner.

PR #11687, opened 2024-06-07
Add warning for users who haven't updated to required minimum bun version that includes $ utility 

aadilmallick:
I don't know exactly which version bun starts having the $ shell utility, but it starts at some specific version and otherwise you get a cryptic error. Include this warning so that bun users that haven't updated can update immediately to use the shell utility. Also help me get my first contribution to open source :)


7f8ddd:
I feel like doing this for one feature rather than all of them is wrong. I would ask in the Discord contributing channel if you want to do that.

Jarred-Sumner:
Thank you for this

We should add some kind of marker to the docs site that clarifies when the feature first became available. I don't think a warning/error is the right message.

PR #11694, opened 2024-06-08
Support cancelling pending transpilations

Jarred-Sumner:
### What does this PR do?

When you use `bun --hot`, it reloads every file.

So if you make a small change to a large file repeatedly, it's going to potentially block transpilation/reloading from continuing.

This reduces hot reloading memory usage and makes it faster by "cancelling" async transpilation when the promise to fetch the ESM is garbage collected.


robobun:
Hello!

Issue #11316, opened 2024-05-24
Crash when using setHeader method of http.ServerResponse


Nivdc:
### How can we reproduce the crash?

Running the following code and accessing the website will cause it to crash.

This happens when `fileSuffix` comes from the match method of a string containing CJK characters.

### JavaScript/TypeScript code that reproduces the crash?

```shell
const http = require('http')

const server = http.createServer((req,res) => {
    let fileSuffix = '测试.html'.match(/\.([a-z0-9]*)$/i)[1]

    res.statusCode = 200
    res.setHeader('Content-Type',('text/' + fileSuffix))
    res.end('Hello World!\n')
})

server.listen(3000,'127.0.0.1',()=>{
    console.log('Server running at http://localhost:3000')
})
```

### Relevant log output

```shell
============================================================
Bun v1.1.9 (bb13798d) Windows x64
Args: "C:\Users\nivdc\.bun\bin\bun.exe", "run", "--hot", ".\test.js"
Features: jsc http_server
Builtins: "bun:main" "node:http" "node:string_decoder" "node:util/types"
Elapsed: 3244ms | User: 78ms | Sys: 15ms
RSS: 114.62MB | Peak: 114.63MB | Commit: 0.17GB | Faults: 28215

panic(main thread): ZigString.slice() called on a UTF-16 string.
Please use .toSlice() instead or carefully check that .is16Bit() is false first.
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.9/wr1bb13798AiggB884hR6krkYwqkwWq151Uu2k0zC+moq2Csj9p2Co+hq2Ck13p2Cmgs32CA0eNotyzEOgzAMBdC9p/hjGRqJhQMwMFeiXbpZxgGrViLFZuD2VQX7ex9d52ha1uSmLPcOTGayoBYQ3q/p0Q/wU9yeJuSC3QUp6nwFLR5CC2oDU5O8mx3gTfiL2CiQ1Pth1PhTRyZzQdbmkX51LCp8
```

### Stack Trace (bun.report)

Bun v1.1.9 ([`bb13798`](<https://github.com/oven-sh/bun/tree/bb13798d9802d26770d8467a200022a2b91f7afd>)) on windows x86_64 [RunCommand]

**panic**: ZigString.slice() called on a UTF-16 string.
[...]

PR #11695, opened 2024-06-08
fix(Bun.serve) fix mimetype with utf16

cirospaciari:
### What does this PR do?

Fix: https://github.com/oven-sh/bun/issues/11316

- [x] Code changes

Added a test

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3223,7 +3223,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                         var content_slice = content.toSlice(this.allocator);
                         defer content_slice.deinit();
 
-                        const content_type_allocator = if (content_slice.isNull()) null else this.allocator;
+                        const content_type_allocator = if (content_slice.allocator.isNull()) null else this.allocator;
                         break :brk MimeType.init(content_slice.slice(), content_type_allocator, &content_type_needs_free);
                     }
                 }
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3223,7 +3223,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                         var content_slice = content.toSlice(this.allocator);
                         defer content_slice.deinit();
 
-                        break :brk MimeType.init(content_slice.slice(), this.allocator, &content_type_needs_free);
+                        break :brk MimeType.init(content_slice.slice(), content_slice.allocator, &content_type_needs_free);
                     }
                 }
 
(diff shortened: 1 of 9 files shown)


src/bun.js/api/server.zig
```diff
@@ -3217,7 +3217,11 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                 if (response.init.headers) |headers_| {
                     if (headers_.fastGet(.ContentType)) |content| {
                         needs_content_type = false;
-                        break :brk MimeType.byName(content.slice());
+
+                        var content_slice = content.toSlice(this.allocator);
+                        defer content_slice.deinit();
+
+                        break :brk MimeType.byName(content_slice.slice());
```
Jarred-Sumner: this is a use-after-free isn't it?

cirospaciari: yea, I was just pushing from the macOS to finish on window, we need to MimeType.init and free if it is allocated

src/bun.js/api/server.zig
```diff
@@ -3213,11 +3213,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
             const content_type: MimeType = brk: {
                 if (response.init.headers) |headers_| {
                     if (headers_.fastGet(.ContentType)) |content| {
                         needs_content_type = false;
-                        break :brk MimeType.byName(content.slice());
+
+                        var content_slice = content.toSlice(this.allocator);
+                        defer content_slice.deinit();
+
+                        break :brk MimeType.init(content_slice.slice(), content_slice.allocator.get(), &content_type_needs_free);
```
Jarred-Sumner: This is not the correct allocator:

```suggestion
                        break :brk MimeType.init(content_slice.slice(), this.allocator, &content_type_needs_free);
```

cirospaciari: If I pass `this.allocator` will not return the same allocator I pass on?
https://github.com/oven-sh/bun/blob/main/src/bun.js/bindings/bindings.zig#L766-L791
Looks like it would do that
https://github.com/oven-sh/bun/blob/main/src/NullableAllocator.zig
The ideia was only allocate if necessary, if i pass `this.allocator` will always allocate.
I can also do like `if (content_slice.isNull()) null else this.allocator` but dont look necessary.

Jarred-Sumner: that makes sense, but we can't use the StringImpl allocator to allocate any memory. It only increments the stringImpl's reference count.

cirospaciari: fixed in https://github.com/oven-sh/bun/pull/11695/commits/c6b26e1bdedf5242fa14bb4a5bbabc6163cd6aec

Issue #8986, opened 2024-02-19
WebSocket server: Await within async fetch breaks WebSocket handshake when client provides a sub-protocol


Drejerdk:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157
1.0.27+c34bbb2e3

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

Whenever the fetch function is async and contains an `await` within the block, the WebSocket handshake will fail, if the client requested a sub-protocol.

```
const server = Bun.serve<{ authToken: string }>({
  port: 3000,
  async fetch(req, server) {
    // The async await below causes the handshake to fail 
    await new Promise((resolve) => setTimeout(resolve, 100))

    const success = server.upgrade(req)
    if (success) {
      return undefined
    }

    return new Response("Hello world!")
  },
  websocket: {
    message: (ws, msg) => console.log(msg)
  },
})

console.log(`Listening on ${server.port}`);
```

### What is the expected behavior?

The WebSocket handshake should be able to handle async functions, even if the client provided a sub-protocol.

Drejerdk:
Updated the description and title.

Originally, I blamed a call to global `fetch` to break the handshake, but in reality it seems to be any `await` within the `async fetch`

PaddeK:
Probably the same issue as mine from a while ago https://github.com/oven-sh/bun/issues/7514

wakeupinagear:
Getting this exact same issue (1006 error code and all) when awaiting `Bun.password.verify`. Removing fixes the error.

```ts
const server = Bun.serve({
    port,
    async fetch(req, server) {
        const url = new URL(req.url);

        if (url.pathname.startsWith('/clusters')) {
            console.log(url.pathname);
            const passwordHash = req.headers.get('authorization') || '';
            if (clusterPassword) {
                const valid = Bun.password.verify(
                    clusterPassword,
                    passwordHash
                );
                if (!valid) {
                    console.log('Stopped unauthorized request');
                    return new Response('Unauthorized', { status: 401 });
                }
            }
        }

        if (server.upgrade(req))
            return;
       
        // rest of http + ws implementation follows
}
```

wakeupinagear:
Update: Bun 1.1.0 fixed this issue for me

Drejerdk:
The async fetch does indeed work now, as long as the client doesn't provide a sub-protocol, ex. 
[...]

Issue #7514, opened 2023-12-07
`server.upgrade` with WebSocket subprotocol does not work if `fetch` handler is async


PaddeK:
### What version of Bun is running?

1.0.15 and 1.0.16

### What platform is your computer?

Darwin 23.1.0 x86_64 i386

### What steps can reproduce the bug?

```ts
import {ServerWebSocket} from 'bun';

const delay = (ms: number) => new Promise(ok => setTimeout(ok, ms));

Bun.serve<{}>({
    fetch: async (req, server) => {
        await delay(1);

        if (server.upgrade(req)) return;
    },
    websocket: {
        open(ws: ServerWebSocket<{}>): void | Promise<void> {
            console.log('open');
        },
        close(ws: ServerWebSocket<{}>, code: number, reason: string): void | Promise<void> {
            console.log('closed', code);
        }
    }
});

new WebSocket('http://localhost:3000', 'ocpp1.6');
```

### What is the expected behavior?

```
open
```

### What do you see instead?

```
open
closed 1006
```

### Additional information

If i remove the line:
```ts
await delay(1);
```
i get the expected output. Additionally if i omit sending the protocol it works with and without the `await delay(1);`.

PR #11707, opened 2024-06-08
fix(Bun.serve) fix async upgrade using custom protocols

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/7514
Fix: https://github.com/oven-sh/bun/issues/8986

- [X] Code changes

Added a test

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -2523,8 +2523,8 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             };
 
             // we have to clone the request headers here since they will soon belong to a different request
-            if (request_object.headers == null) {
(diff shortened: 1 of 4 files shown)


src/bun.js/api/server.zig
```diff
@@ -6230,6 +6241,7 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
             request_value.ensureStillAlive();
             const response_value = this.config.onRequest.callWithThis(this.globalThis, this.thisObject, &args);
             defer {
+                if (!ctx.didUpgradeWebSocket()) {}
```
Jarred-Sumner, line 6244: did you mean to do something here?

cirospaciari: I did not, probably was some log that I add and removed sorry my mistake

PR #11724, opened 2024-06-09
Support `NODE_EXTRA_CA_CERTS`

### LINKED ISSUE #271: CA support for SSL connection to NPM registries

### LINKED ISSUE #7124: error: SELF_SIGNED_CERT_IN_CHAIN downloading package manifest

1.0.11+f7f6233ea

Darwin 22.6.0 arm64 arm

I am on a corporate laptop that uses a company self-signed cert.

I ran the command: `bun install -g aws-cdk` and got the following output:
```
bun add v1.0.11 (f7f6233e)
  🔍 aws-cdk [6/6]
error: SELF_SIGNED_CERT_IN_CHAIN downloading package manifest aws-cdk
error: aws-cdk@ failed to resolve
```

I tried following Node's solution for this issue, and I set the following environment variable:
```bash
export NODE_EXTRA_CA_CERTS="path_to_my_cert.pem"
```

This didn't seem to work. I also tried `BUN_EXTRA_CA_CERTS`, but that didn't work either.

Ideally, it would be possible to set an environment variable that points to my CA file, in precisely the same way it works for Node. (Or some other user-friendly method that is clearly documented.)

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #7200: Error verifying https certificates

1.0.12

Linux 5.10.0-23-amd64 x86_64 unknown

* [Install kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installing-from-release-binaries)
* Create a kubernetes cluster using kind: `kind create cluster`
* Create a bun project: `bun init -y`
* Install kubernetes client library: `bun install @kubernetes/client-node`
* Create the following `index.js` file:

  ```typescript
  import k8s from '@kubernetes/client-node';

  const kc = new k8s.KubeConfig();

  kc.loadFromDefault();

  const k8sApi = kc.makeApiClient(k8s.CoreV1Api);

  const main = async () => {
      try {
          const podsRes = await k8sApi.listNamespacedPod('default');
          console.log(podsRes.body);
      } catch (err) {
          console.error(err);
      }
  };
  
  main();

  ```
  
* Execute it with bun: `bun index.js` 

It should list the pods in the kubernetes cluster, below is the same program file executed with `ts-node`:

```javascript
V1PodList {
  apiVersion: 'v1',
  items: [],
  kind: 'PodList',
  metadata: V1ListMeta {
    _continue: undefined,
    remainingItemCount: undefined,
    resourceVersion: '120519',
    selfLink: undefined
  }
}
```

### What do you see instead?

Instead it throws the following error validating the certificates:

```
UNABLE_TO_VERIFY_LEAF_SIGNATURE: unable to verify the first certificate
 path: "https://127.0.0.1:35781/api/v1/namespaces/default/pods"
```

### Additional information

Other software versions that may be of interest:

* ts-node: v10.9.1
* kind: 0.20.0

LudvigHz:
### What does this PR do?

- [x] Code changes

Load ca certificates from NODE_EXTRA_CA_CERTS

Respects the Node.js API and loads the CA file pointed to by the environment variable NODE_EXTRA_CA_CERTS into the global SSL context.

This works for fetch, https.request and (should also cover) requests made by the bun CLI, f.ex `bun install`.

I haven't added tests for `bun install` since there isn't really existing test infra for bun install network activity.

I wrote automated tests

If Zig files changed:

___

A couple notes on this:
- I looked into using the envLoader for getting the env var, but in (almost) all instances the `HTTPThread` is initialized before the envLoader, and it would require quite a bit of rebasing to get the envMap context into the `HTTPContext` initialization. The way it's done now, it requires minimal code changes while still being compatible with Node.js.
- I passed the ca file into the socket context initialization, since the option is already there, however the code for loading ca files to the certificate store will crash when it fails (see the last test I added). I looked into exposing the API for adding additional CAs to the global socket context, so I could do that as well, if the current behavior is unwanted. I just went with the current method to minimize changes and keeping it simple.

Fixes #7124 
Fixes #7200 
Fixes #271

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -379,15 +379,57 @@ fn NewHTTPContext(comptime ssl: bool) type {
 
         pub fn init(this: *@This()) !void {
             if (comptime ssl) {
-                const opts: uws.us_bun_socket_context_options_t = .{
+                var opts: uws.us_bun_socket_context_options_t = .{
                     // we request the cert so we load root certs and can verify it
                     .request_cert = 1,
                     // we manually abort the connection if the hostname doesn't match
                     .reject_unauthorized = 0,
                 };
+
+                var extra_cas: ?std.ArrayList([*c]const u8) = null;
+
+                blk: {
+                    if (std.process.getEnvVarOwned(default_allocator, "NODE_EXTRA_CA_CERTS")) |extra_ca_certs_path| {
+                        const file = bun.openFile(extra_ca_certs_path, .{ .mode = .read_only }) catch |err| {
+                            Output.warn("Skipping loading of extra ca file at {s}: {}", .{ extra_ca_certs_path, err });
+                            Output.flush();
+                            break :blk;
+                        };
+
+                        defer file.close();
+                        defer default_allocator.free(extra_ca_certs_path);
+
+                        const file_size = file.getEndPos() catch |err| {
+                            Output.warn("Skipping loading of extra ca file {s}: {}", .{ extra_ca_certs_path, err });
+                            Output.flush();
+                            break :blk;
+                        };
+
+                        const contents = file.reader().readAllAlloc(default_allocator, file_size) catch |err| {
+                            Output.warn("Skipping loading of extra ca file {s}: {}", .{ extra_ca_certs_path, err });
+                            Output.flush();
+                            break :blk;
+                        };
+
+                        var cas = std.ArrayList([*c]const u8).init(default_allocator);
+                        try cas.append(@ptrCast(contents));
+
+                        opts.ca = @ptrCast(try cas.toOwnedSlice());
+                        opts.ca_count = 1;
+                        Output.debug("Loading extra ca cert from {s}", .{extra_ca_certs_path});
+                        extra_cas = cas;
+                    } else |err| {
+                        log("WARN: Failed to read NODE_EXTRA_CA_CERTS {}", .{err});
+                    }
+                }
+
                 this.us_socket_context = uws.us_create_bun_socket_context(ssl_int, http_thread.loop.loop, @sizeOf(usize), opts).?;
 
                 this.sslCtx().setup();
+
+                if (extra_cas) |ca_list| {
+                    ca_list.deinit();
+                }
             } else {
                 const opts: uws.us_socket_context_options_t = .{};
                 this.us_socket_context = uws.us_create_socket_context(ssl_int, http_thread.loop.loop, @sizeOf(usize), opts).?;
--- a/test/js/node/http/fixtures/node-extra-ca-certs.js
+++ b/test/js/node/http/fixtures/node-extra-ca-certs.js
@@ -1,0 +1,17 @@
+(async () => {
+  const port = process.argv[2] ? parseInt(process.argv[2]) : null;
+
+  try {
+    const res = await fetch(`https://localhost:${port}`, {
+      tls: {
+        rejectUnauthorized: true,
+      },
+    });
+    const t = await res.text();
+    console.log(`res: ${t}`);
+    process.exit(0);
+  } catch (error) {
+    console.log(error);
+    process.exit(1);
+  }
+})();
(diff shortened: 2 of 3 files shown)


src/http.zig
```diff
@@ -379,15 +379,57 @@ fn NewHTTPContext(comptime ssl: bool) type {
...
                     // we request the cert so we load root certs and can verify it
                     .request_cert = 1,
                     // we manually abort the connection if the hostname doesn't match
                     .reject_unauthorized = 0,
                 };
+
+                var extra_cas: ?std.ArrayList([*c]const u8) = null;
+
+                blk: {
+                    if (std.process.getEnvVarOwned(default_allocator, "NODE_EXTRA_CA_CERTS")) |extra_ca_certs_path| {
```
Jarred-Sumner, line 392: This should use something in `DotEnv.Loader` and be cached there. We will need to add support for this in several places instead of only in the http client.

src/http.zig
```diff
@@ -379,15 +379,57 @@ fn NewHTTPContext(comptime ssl: bool) type {
...
+                        const file = bun.openFile(extra_ca_certs_path, .{ .mode = .read_only }) catch |err| {
```
Jarred-Sumner, line 393: There are way too many ways to read files in Bun's codebase which is very confusing (not your fault at all)

Instead of this, let's do:
```zig
const contents = bun.sys.File.readFromUserInput(std.fs.cwd(), extra_ca_certs_path, bun.default_allocator)
```

And not in here, instead in `DotEnv.Loader`

test/js/node/http/fixtures/node-extra-ca-certs.js
```diff
@@ -0,0 +1,17 @@
+(async () => {
```
Jarred-Sumner, line 1: IIFEs like this are unnecessary in Bun. Bun will use ESM by default.

test/js/node/http/node-http.test.ts
```diff
@@ -829,6 +829,90 @@ describe("node:http", () => {
     });
   });
 
+  describe("NODE_EXTRA_CA_CERTS", () => {
+    const run = async (done, env, func: (proc: Subprocess<"pipe">) => Promise<void>) => {
+      const server = https.createServer(
+        {
+          key: nodefs.readFileSync(path.join(import.meta.dir, "fixtures", "openssl_localhost.key")),
```
Jarred-Sumner, line 836: Looks like these files were not committed?

LudvigHz: Whoops, I initially based this off of #11322. I'll rebase now that that's merged and we can reuse the same certs.

src/http.zig
LudvigHz: Okey, that was the main thing I was unsure about. I initially looked at going that route, but it ended up being kind of complicated with the current architecture of how/when the HTTPThread is initialized.

But then I'll move this there instead and I will have to change the initialization of the `HTTPThread` to depend on `DotEnv.Loader`. I do agree it's also better option devex wise.

If you have any ideas on how to best get the ca file contents into the context of the https context initialization, I'd happily take some pointers as well :sweat_smile: (the first/only way I saw how to do this involved passing the dotenv context around alot)

test/js/node/http/node-http.test.ts
mirismaili: @LudvigHz 
Is there any update for the status of this PR?
🙏

LudvigHz: I haven't had the time to work on it lately. And as I'll need to refactor alot of the http initialization it's going to require quite a bit more work than what this PR proposes currently.

But hopefully I'll have some free time soon to look at it some more :)

Jarred-Sumner:
Closing due to inactivity, but feel free to send another if you get the time

PR #11766, opened 2024-06-10
Fix cloning `File` with `structuredClone`

### LINKED ISSUE #11696: `structuredClone` does not properly clone `File`

1.1.12

macOS 14.5

```js
const f1 = new File(['test'], 'foo.txt');
f1.size; // -> 4
f1.name; // -> 'foo.txt'
f1.lastModified; // -> 0
const f2 = structuredClone(f1);
f2.size; // -> 4
f2.name; // -> undefined
f2.lastModified; // -> 4503599627370495
```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

dylan-conway:
### What does this PR do?
Fixes #11696 

- bumps `Blob` serialization version to 2.
- now serializes a byte for whether it's a dom file, the length and bytes for `stored_name`, and a `f64` for `lastModified`.

Also sets `lastModified` to the current time in milliseconds if it's not in options when creating a `File`. Previously we set it to zero.

added tests

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -117,8 +117,8 @@ pub const Blob = struct {
     pub const SizeType = u52;
     pub const max_size = std.math.maxInt(SizeType);
 
-    const serialization_version: u8 = 1;
-    const reserved_space_for_serialization: u32 = 128;
+    const serialization_version: u8 = 2;
+    const reserved_space_for_serialization: u32 = 129;
 
     pub fn getFormDataEncoding(this: *Blob) ?*bun.FormData.AsyncFormData {
         var content_type_slice: ZigString.Slice = this.getContentType() orelse return null;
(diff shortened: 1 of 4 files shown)


src/bun.js/webcore/blob.zig
```diff
@@ -1410,11 +1448,18 @@ pub const Blob = struct {
...
+                    set_last_modified = true;
                     blob.last_modified = last_modified.coerce(f64, globalThis);
                 }
             }
         }
 
+        if (!set_last_modified) {
+            // `lastModified` should be the current date in milliseconds if unspecified.
+            // https://developer.mozilla.org/en-US/docs/Web/API/File/lastModified
+            blob.last_modified = @floatFromInt(std.time.milliTimestamp());
```
Jarred-Sumner, line 1460:
```suggestion
            blob.last_modified = @floatFromInt(bun.timespec.now().ms());
```

Jarred-Sumner: actually im wrong!!

Issue #11720, opened 2024-06-09
Investigate nonsensical stack traces in crash reporter


Jarred-Sumner:
Relevant issues:
- https://github.com/oven-sh/bun/issues/11709
- https://github.com/oven-sh/bun/issues/11708
- https://github.com/oven-sh/bun/issues/11665

paperdave:
it was solved by this PR

PR #11769, opened 2024-06-11
fix(ci): make it so ci doesnt overwrite the release builds with canary's debug symbols

paperclover:
i could write a 10 page rant about this whole situation but instead of that i am just going to fix the bug

see #11720. I will not mark this issue as fixed until i see with my own eyes that this attempt at writing a piece of reliable software works.


Jarred-Sumner:
did you mean to commit .zig-cache

image

Issue #11722, opened 2024-06-09
Memory leak on <ws>.WebSocket()


Yonle:
### What version of Bun is running?

Linux 4.19.191-g96eac46b6306-dirty aarch64 unknown

### What platform is your computer?

Linux

### What steps can reproduce the bug?

```js
const WS = require("ws");
const server = new WS.WebSocketServer({ port: 1111 }); // dummy server

function c() {
  const sock = new WS.WebSocket("ws://localhost:1111");
  sock.on('open', _ => {
    sock.terminate();
    c();
  });
}

c();
```

### What is the expected behavior?

Memory usage should not go above than 60 MB (?) and shrinks.

### What do you see instead?

Memory usage go above than 60 MB and never shrinks.

### Additional information

Same effect occurs even with `WebSocket().close()`. Removing listeners also does not shrink.

Jarred-Sumner:
An example which more conclusively demonstrates a memory leak where the `WebSocket` object seems to never be finalized:
```js
const WS = require("ws");

const server = Bun.serve({
   port: "1111",
   fetch(req, server) {
     return server.upgrade(req);
   },
   websocket: {
     open() {},
     data() {},
     message() {}, 
     drain() {},
  } 
});

var queue = [];
function c() {
  const {promise, resolve} = Promise.withResolvers();
  const sock = new WebSocket("ws://localhost:1111");
  sock.addEventListener('open', _ => {
    sock.addEventListener("close", () => {
       resolve();
    });
    sock.close();
  })

  return promise;
}

setInterval(() => 
{
  Bun.gc(true);
  
  const rss = process.memoryUsage.rss() / 1024 / 1024 | 0;
  const objectTypeCounts = require("bun:jsc").heapStats().objectTypeCounts;
  console.log(require("bun:jsc").heapStats().protectedObjectTypeCounts);
  console.log(Object.fromEntries(Object.entries(objectTypeCounts).sort()));
}, 1000);

while (true) {
  await c();
 await Bun.sleep(0);
}
```

NextGenOP:
this issue also occur while using ws npm package
server: 
```
const WebSocket = require('ws');
const wss = new WebSocket.Server({ port: 8080 });
wss.on('connection', (ws) => {
	ws.close()
});
```
but fixed with Bun.gc

closing connection from server also create issue in client

Yonle:
The memory usage is still exceeding 70 MB of RAM.

Bun:
[...]

PR #11788, opened 2024-06-11
fix(WebSocket) fix ref count so finalize is called

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/11722

- [x] Code changes

Added test

### DIFF
--- a/src/bun.js/bindings/webcore/WebSocket.cpp
+++ b/src/bun.js/bindings/webcore/WebSocket.cpp
@@ -1211,8 +1211,17 @@ void WebSocket::didClose(unsigned unhandledBufferedAmount, unsigned short code, 
 
     if (this->hasEventListeners("close"_s)) {
         this->dispatchEvent(CloseEvent::create(wasClean, code, reason));
-        this->decPendingActivityCount();
 
+        // we deinit if possible in the next tick
+        if (auto* context = scriptExecutionContext()) {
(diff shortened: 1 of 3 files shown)


src/bun.js/bindings/webcore/WebSocket.cpp
```diff
@@ -1206,25 +1206,35 @@ void WebSocket::didClose(unsigned unhandledBufferedAmount, unsigned short code,
     this->m_connectedWebSocketKind = ConnectedWebSocketKind::None;
     this->m_upgradeClient = nullptr;
 
+    // since we are open and closing now we know that we have at least one pending activity
+    // so we just call decPendingActivityCount() after dispatching the event
```
Jarred-Sumner, line 1210: Can we `ASSERT` this?

test/js/web/websocket/websocket.test.js
```diff
@@ -2,7 +2,7 @@ import { describe, it, expect } from "bun:test";
 import { bunExe, bunEnv, gc, tls } from "harness";
 import { readFileSync } from "fs";
 import { join, resolve } from "path";
-import process from "process";
+import process, { nextTick } from "process";
```
Jarred-Sumner, line 5: unnecessary import

Jarred-Sumner: VSCode CMD-Shift + P -> "Organize Imports"

Jarred-Sumner:
Looks good, merging. But maybe add the ASSERT in a follow-up PR?

PR #11790, opened 2024-06-11
In AbortSignal, use Ref explicitly to ensure its alive

Jarred-Sumner:
### What does this PR do?

I pressed the merge button without first checking WebKit's implementation

@cirospaciari did you try this instead?


cirospaciari (approved):
Good catch,  did not tried this first. This will work.

Issue #10178, opened 2024-04-11
class extending blob cant set `name` prop


juliusmarminge:
### What version of Bun is running?

1.1.2+c8d072c2a

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

Repo: https://github.com/juliusmarminge/bun-blob

To test, run `bun run index.js` - see error.
Run `node index.js` - works as expected.

![CleanShot 2024-04-11 at 21 31 03](https://github.com/juliusmarminge/bun-blob/assets/51714798/318eb53f-c8fc-41c5-be3b-6fcde67f9565)

### What is the expected behavior?

```sh
➜  bun-blob node index.js
blob name undefined
Blob { size: 13, type: '' }
hello.txt
1712863862801
undefined
```

### What do you see instead?

```sh
➜  bun-blob bun index.js
blob name undefined
 8 |         ...options,
 9 |         type: options?.type ?? undefined,
10 |         lastModified: options?.lastModified ?? Date.now(),
11 |       };
12 |       super(parts, optionsWithDefaults);
13 |       this.name = name;
           ^
TypeError: Attempted to assign to readonly property.
      at new UTFile (/Users/julius/dev/sandbox/bun-blob/index.js:13:7)
      at /Users/julius/dev/sandbox/bun-blob/index.js:19:16
```

juliusmarminge:
Interesting side note, extending e.g. `Buffer` works fine:

![CleanShot 2024-04-11 at 21 36 00](https://github.com/oven-sh/bun/assets/51714798/dc23a067-098f-4eb3-9144-f82ed7d005df)

fahidsarker:
This issue still exists in bun 1.2.2. why is it closed?

MathiasGruber:
I'm still running into this on latest bun and latest uploadthing

juliusmarminge:
> I'm still running into this on latest bun and latest uploadthing, any hints on what to do [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/51714798?s=64&amp;u=7010559c831ddae241a5c1ace13e5a78048f9219&amp;v=4">@juliusmarminge](https://github.com/juliusmarminge) ?

nothing we can do unfortunately :/ This is an issue in Bun...

Issue #10072, opened 2024-04-08
Blob.type is not working


juliomuhlbauer:
### What version of Bun is running?

1.1.2+c8d072c2a

### What platform is your computer?

Linux 5.13.0-1022-aws x86_64 x86_64

### What steps can reproduce the bug?

1. Fetch a url
2. Get the blob
3. Try to get the blob.type

```ts
const blob = await fetch("https://bun.sh/logo.svg").then((res) => res.blob());

console.log("Blob type is: ",blob.type)
  ```

### What is the expected behavior?

Blob.type appearing on the console

### What do you see instead?

Nothing

### Additional information

Playground: https://codedamn.com/playground/J2cXvOIyG72SDI_lM3Zkp

**Comment by Electroid** [comment]
It looks like there's some sort of race condition going on.

```sh
bun --print 'for (let i = 0; i < 10_000; i++) await fetch("https://bun.sh/logo.svg").then((res) => res.blob()).then(blob => console.log(blob.type))'
```

You'll occasionally see the `blob.type` is undefined.

juliomuhlbauer:
This type of bug makes bun unreliable to me. Sad

Jarred-Sumner:
Confirmed this is still reproducing as of Bun v1.1.13:

```js
~/Desktop
❯ bun hm.js
Blob type is:  image/svg+xml

~/Desktop
❯ bun hm.js
Blob type is:  image/svg+xml

~/Desktop
❯ bun hm.js
Blob type is:  image/svg+xml

~/Desktop
❯ bun hm.js
Blob type is:

~/Desktop
❯ bun hm.js
Blob type is:  image/svg+xml

~/Desktop
❯ bun hm.js
Blob type is:  image/svg+xml

~/Desktop
❯ bun hm.js
Blob type is:  image/svg+xml
```

PR #11815, opened 2024-06-12
fix(Blob/stream) blob from fetch now reliable returns type, blob name can be set

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/10178
Fix: https://github.com/oven-sh/bun/issues/10072

also refactor some `catch undefined` and fix a stream bug

- [X] Code changes

Added tests

### DIFF
--- a/src/bun.js/webcore/blob.zig
+++ b/src/bun.js/webcore/blob.zig
@@ -109,6 +109,10 @@ pub const Blob = struct {
     globalThis: *JSGlobalObject = undefined,
 
     last_modified: f64 = 0.0,
+    /// Blob name will lazy initialize when getName is called, but
+    /// we must be able to set the name, and we need to keep the value alive
+    /// https://github.com/oven-sh/bun/issues/10178
+    name: JSC.Strong = .{},
 
     /// Max int of double precision
     /// 9 petabytes is probably enough for awhile
@@ -127,7 +131,7 @@ pub const Blob = struct {
         var content_type_slice: ZigString.Slice = this.getContentType() orelse return null;
         defer content_type_slice.deinit();
         const encoding = bun.FormData.Encoding.get(content_type_slice.slice()) orelse return null;
-        return bun.FormData.AsyncFormData.init(this.allocator orelse bun.default_allocator, encoding) catch unreachable;
+        return bun.FormData.AsyncFormData.init(this.allocator orelse bun.default_allocator, encoding) catch bun.outOfMemory();
     }
 
     pub fn hasContentTypeFromUser(this: *const Blob) bool {
@@ -171,8 +175,8 @@ pub const Blob = struct {
             *Handler,
             handler,
             Handler.run,
-        ) catch unreachable;
-        var read_file_task = ReadFile.ReadFileTask.createOnJSThread(bun.default_allocator, global, file_read) catch unreachable;
+        );
+        var read_file_task = ReadFile.ReadFileTask.createOnJSThread(bun.default_allocator, global, file_read);
 
         // Create the Promise only after the store has been ref()'d.
         // The garbage collector runs on memory allocations
@@ -209,8 +213,8 @@ pub const Blob = struct {
             NewInternalReadFileHandler(Handler, Function).run,
             this.offset,
             this.size,
-        ) catch unreachable;
-        var read_file_task = ReadFile.ReadFileTask.createOnJSThread(bun.default_allocator, global, file_read) catch unreachable;
+        );
+        var read_file_task = ReadFile.ReadFileTask.createOnJSThread(bun.default_allocator, global, file_read);
         read_file_task.schedule();
     }
 
@@ -524,7 +528,7 @@ pub const Blob = struct {
         buf: []u8 = "",
         globalThis: *JSC.JSGlobalObject,
         pub fn convert(this: *URLSearchParamsConverter, str: ZigString) void {
-            var out = str.toSlice(this.allocator).cloneIfNeeded(this.allocator) catch unreachable;
+            var out = str.toSlice(this.allocator).cloneIfNeeded(this.allocator) catch bun.outOfMemory();
             this.buf = @constCast(out.slice());
         }
     };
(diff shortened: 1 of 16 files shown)


src/bun.js/webcore/blob.zig
```diff
@@ -109,6 +109,10 @@ pub const Blob = struct {
     globalThis: *JSGlobalObject = undefined,
 
     last_modified: f64 = 0.0,
+    /// Blob name will lazy initialize when getName is called, but
+    /// we must be able to set the name, and we need to keep the value alive
+    /// https://github.com/oven-sh/bun/issues/10178
+    name: JSC.Strong = .{},
```
Jarred-Sumner: If we ensure the name is only allowed to be a string, can't we get away with a `bun.String` here instead? Otherwise, couldn't we use a `mutable WriteBarrier` (nameGetCachedValue)?

src/bun.js/webcore/response.classes.ts
```diff
@@ -153,7 +153,7 @@ export default [
       // This is *not* spec-compliant.
       name: {
         getter: "getName",
-        cache: true,
```
Jarred-Sumner, line 156: We can have both `cache: true` and `setter: "setName"` and then the cached value will be kept alive but they can still overwrite it if necessary

Jarred-Sumner: and then in `toJS` we can set the name on the Blob instance there

Jarred-Sumner: That would probably simplify this and avoid worrying about leaking the name string

cirospaciari: We would need a strong ref for the JS `this` value so we can set/get the cached values, this would probably just be more complicated and we would still need to deinit one Strong ref, I can change to be a `bun.String` instead of a `JSC.Strong`, this would limit to be only a string but chrome dont limit the type on blob name to just string.
image

Jarred-Sumner: I think it'd be okay to limit it to String until someone complains about that, and making it `configurable` so they can overwrite it with Object.defineProperty if they really want.

cirospaciari: sg will change to `bun.String` and make it `configurable`

src/bun.js/webcore/blob.zig
```diff
@@ -109,6 +109,10 @@ pub const Blob = struct {
     globalThis: *JSGlobalObject = undefined,
 
     last_modified: f64 = 0.0,
+    /// Blob name will lazy initialize when getName is called, but
+    /// we must be able to set the name, and we need to keep the value alive
+    /// https://github.com/oven-sh/bun/issues/10178
+    name: ?bun.String = null,
```
Jarred-Sumner: this doesn't need to be optional, check if .tag == .Dead

cirospaciari: fixed: https://github.com/oven-sh/bun/commit/7689a6289d8025c234c571e59b50245b03b363b6

src/bun.js/webcore/body.zig
```diff
@@ -683,9 +688,29 @@ pub const Body = struct {
...
+                                    blob.content_type = mimeType.value;
+                                    blob.content_type_allocated = allocated;
+                                    blob.content_type_was_set = true;
+                                    if (blob.store != null) {
+                                        blob.store.?.mime_type = mimeType;
+                                    }
+                                }
+                            }
+                            if (!blob.content_type_was_set and blob.store != null) {
+                                blob.content_type = MimeType.text.value;
```
Jarred-Sumner, line 708: why default to text? is that what browsers do? can you leave a comment if so?

src/bun.js/webcore/body.zig
```diff
@@ -1162,26 +1187,36 @@ pub fn BodyMixin(comptime Type: type) type {
...
+                        blob.content_type = mimeType.value;
+                        blob.content_type_allocated = allocated;
+                        blob.content_type_was_set = true;
+                        if (blob.store != null) {
+                            blob.store.?.mime_type = mimeType;
+                        }
                     }
-                } else {
+                }
+                if (!blob.content_type_was_set and blob.store != null) {
```
Jarred-Sumner, line 1213: can the store already have a content_type set? should we be changing it if that is the case?

Jarred-Sumner: i guess that was the previous behavior

src/io/PipeReader.zig
```diff
@@ -399,6 +399,7 @@ pub fn WindowsPipeReader(
             }
             var this: *This = bun.cast(*This, fs.data);
             fs.deinit();
+            if (this.flags.is_done) return;
```
Jarred-Sumner, line 402: is this an unrelated bug?

src/bun.js/webcore/body.zig
cirospaciari: I just keep the previous behavior

src/io/PipeReader.zig
cirospaciari: unrelated, mentioned in the PR, also added a test for it. just 1 line fix

src/bun.js/webcore/body.zig
cirospaciari: not sure why we did that before, should I remove?

src/bun.js/webcore/response.classes.ts
Jarred-Sumner, line 156: we can still cache it right?

Jarred-Sumner: that way it only creates the name once

Jarred-Sumner:
lets merge after CI

src/bun.js/webcore/response.classes.ts
cirospaciari: fixed: https://github.com/oven-sh/bun/pull/11815/commits/781cebddf33c2d4674719508676a2de505c5569d

Issue #11114, opened 2024-05-16
missing npm_config_user_agent  env in bunx command at windows


kravetsone:
### What version of Bun is running?

1.1.8+89d25807f

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

```typescript 
export function detectPackageManager() {
	const userAgent = process.env.npm_config_user_agent;

	if (!userAgent)
		throw new Error(
			`Package manager was not detected. Please specify template with "--template bun"`,
		);

	return userAgent.split(" ")[0].split("/")[0] as PackageManager;
}
```
I use this function in my [create-elysiajs](https://github.com/kravetsone/create-elysiajs) and [create-gramio](https://github.com/gramiojs/create-gramio).

And `process.env.npm_config_user_agent` is undefined on windows!

If i try to use it via `npx tsx` it works but when i switch to `bunx` it was broken:

```bash
PS Z:\PROJECTS\node-ts\GRAMIO\create-gramio> npx tsx 
Need to install the following packages:
  tsx@4.10.3
Ok to proceed? (y)
Welcome to Node.js v20.9.0.
Type ".help" for more information.
> process.env.npm_config_user_agent
'npm/9.8.0 node/v20.9.0 win32 x64 workspaces/false'
>
(To exit, press Ctrl+C again or Ctrl+D or type .exit)
>
PS Z:\PROJECTS\node-ts\GRAMIO\create-gramio> bunx tsx
Welcome to Node.js v20.9.0.
Type ".help" for more information.
> process.env.npm_config_user_agent
undefined
>
(To exit, press Ctrl+C again or Ctrl+D or type .exit)
>
```

On WSL all works fine

```bash
kravets@KravetsBook:/mnt/z/PROJECTS/node-ts/GRAMIO/create-gramio$ bunx tsx
Welcome to Node.js v20.10.0.
Type ".help" for more information.
> process.env.npm_config_user_agent
'bun/1.1.8 npm/? node/v21.6.0 linux x64'
>
```

Issue #11016, opened 2024-05-11
Add `npm_config_user_agent` when running `bun create`


C0RR1T:
### What is the problem this feature would solve?

Other package managers (like pnpm, yarn and the like) inject `npm_config_user_agent` into `process.env` when running scripts. this makes it really easy for packages to differentiate which package manager was used to execute the script. Bun also seems to do this in some cases, but not in `bun create` (didn't test any command besides `bun run`, I've heard that it works with `bunx` aswell.

### What is the feature you are proposing to solve the problem?

Add the bun user agent via the `process.env.npm_config_user_agent` variable when running `bun create`

Issue #9984, opened 2024-04-05
windows: (bun: v1.1.1) bun create & bunx using npm to install dependencies on windows.


vgseven:
### What version of Bun is running?

1.1.1+ca1dbb4eb

### What platform is your computer?

Microsoft Windows NT 10.0.22621.0 x64 (Windows 11)

### What steps can reproduce the bug?

creating an next.js app using bun (1.1.1) on windows with `bun create next-app@latest` or `bunx create-next-app@latest` uses `npm` to install dependencies which consumes too much time for installations.

following screenshots shows the process while installing `next` using `bun`: 

![bun-npm](https://github.com/oven-sh/bun/assets/81641747/61d70d3f-3f25-41d9-bc17-150464afe551)
![bun-i](https://github.com/oven-sh/bun/assets/81641747/e035cbc6-78fc-420b-8a85-da1549676a52)

### What is the expected behavior?

I have tried `bun` last month when canary version of `bun` (1.0.35 - 36) rolling, the app created using `bun create next-app@latest` uses `bun` not `npm` to install dependencies and init next app due to un-function state of `bunx`.

bun (1.0.35) is fast enough to install and init next app using bun, but currently after `bundows` release to process of installing any app takes too much time and I think it's because of use of `npm` behind the scene.

bun needs to use bun again for all functions of bun...

### What do you see instead?

As shown in screenshot bun is using npm instead of bun for installation process which consumes too much time..

### Additional information

`bunx create-next-app@latest` uses `bun` to install next app which takes almost 94.44 sec.

![bun-i-linux](https://github.com/oven-sh/bun/assets/81641747/bffcba5d-cd02-4b6d-8dd4-81498ad48cc5)

cosigyn:
You can force it with `--use-bun` while waiting for a fix

Electroid:
It looks like `create-next-app` is using `process.env.npm_config_user_agent` to determine whether to use Bun or not:

https://github.com/vercel/next.js/blob/f9ad49fc6f57f3bcf4f508b59deab695f06c74b1/packages/create-next-app/helpers/get-pkg-manager.ts#L4

It's possible we're not setting this properly on Windows.

vgseven:
`npm` is the default pkg manager if none of any (`yarn`, `pnpm`, `bun`) detected. `create-next-app` works with bun if flagged with --bun as @cosigyn mentioned above.

[...]

Issue #9810, opened 2024-04-01
Error when creating astro project with bun


DAVIDS2405:
### What version of Bun is running?

1.1.0

### What platform is your computer?

Windows 11 

### What steps can reproduce the bug?

```
bunx create-astro@latest
bun create astro
```
Whether using bunx or bun create to create the project with astro, there is a small problem that although I skip the installation of packages or not with npm and not with bun with previous versions there was not this error directly I installed everything with bun.
 
![image](https://github.com/oven-sh/bun/assets/85196670/03604002-3209-4c42-a052-905f27bdbe4d)

### What is the expected behavior?

It is expected to directly put me that it is going to occupy bun and not npm

Jarred-Sumner:
<img width="438" alt="image" src="https://github.com/oven-sh/bun/assets/709451/61927372-dc61-48a3-90c5-e136cbf12350">
Fixed in Bun v1.1.14 by @SunsetTechuila

PR #11817, opened 2024-06-12
fix(bun_shim_impl): pass env to `CreateProcessW`

SunsetTechuila:
### What does this PR do?

fixes https://github.com/oven-sh/bun/issues/11114

fixes https://github.com/oven-sh/bun/issues/11016

fixes https://github.com/oven-sh/bun/issues/9984

fixes https://github.com/oven-sh/bun/issues/9810

fixes https://github.com/oven-sh/bun/issues/9791

- [x] Code changes

- [x] I ran my tests locally and they pass


src/install/windows-shim/bun_shim_impl.zig
```diff
@@ -795,8 +795,8 @@ fn launcher(comptime mode: LauncherMode, bun_ctx: anytype) mode.RetType() {
             null,
             null,
             1, // true
-            0,
-            null,
+            w.CREATE_UNICODE_ENVIRONMENT,
```
paperclover:
```suggestion
            if (is_standalone) 0 else w.CREATE_UNICODE_ENVIRONMENT,
```

SunsetTechuila:
I guess a new test should be added for this, but idk what it should be

Jarred-Sumner:
@SunsetTechuila a bun install test that installs a local package bin which where all the script does is `process.stdout.write(process.env.npm_config_user_agent)`

src/install/windows-shim/bun_shim_impl.zig
```diff
@@ -795,8 +795,8 @@ fn launcher(comptime mode: LauncherMode, bun_ctx: anytype) mode.RetType() {
...
+            @constCast(bun_ctx.environment),
```
paperclover:
```suggestion
            if (is_standalone) null else @constCast(bun_ctx.environment),
```

In addition, I think `@constCast` might be a subtle mistake.

paperclover (changes requested):
See review comments. This file is built in two modes, so it has to handle one case where a specific env block is provided, but also pass the default (inherit) in standalone.

Otherwise, this is great. Thanks for finding this mistake.

src/install/windows-shim/bun_shim_impl.zig
SunsetTechuila: this check confused me, for some reason i thought it was above the code that i changed

https://github.com/oven-sh/bun/blob/fab96a74ea13da04459ea7f62663c4d2fd421778/src/install/windows-shim/bun_shim_impl.zig#L817

SunsetTechuila:
@Jarred-Sumner like this?

paperclover (approved):

Jarred-Sumner:
Great. I bet there are more issues this fixes.

SunsetTechuila:
> Great. I bet there are more issues this fixes

I did my best to find all

Jarred-Sumner:
the ones related to `Path` or `PATH` i think would be fixed since we're now propagating environment variables correctly

SunsetTechuila:
more issues fixed by this pr: 

- https://github.com/oven-sh/bun/issues/9810
- https://github.com/oven-sh/bun/issues/9791
- https://github.com/oven-sh/bun/issues/10188
- https://github.com/oven-sh/bun/issues/5995

i will try to find as much as i can later and leave a comment with all the relevant issues i found

staabm:
@Jarred-Sumner just a suggestion. at phpstan/phpstan-src we have a so called "issue-bot" which checks after a PR is merged, whether the newly merged stuff changes outcome/output of open issues.

the bot can do such things, because [we have a playground](https://phpstan.org/try) which people can replicate bugs in. they can then store their repro examples and get a sharable unique url for it, e.g.: https://phpstan.org/r/d488550e-4296-4292-8b86-b707b7628044

the issue bot scans open issues which are tagged as "bug" or "feature require" and finds reproduction links in this issues (or comments of this issues). as soon as the outcome/output of a reproduction changes, the issue [bot will leave a comment](https://github.com/phpstan/phpstan/issues/10297#issuecomment-2034488493) with the output diff in the open issue, and tags the initial author of the issue.
that way you can easily see which issues are affected/fixed by merging a PR.

additionally we run this bot in PRs as a separate job, so we can even see before merging which issues will be affected by a given change. 

---

this might be something helpful for your team to get on top of the huge open issue list.

SunsetTechuila:
> i will try to find as much as i can later and leave a comment with all the relevant issues i found

can't find any

Issue #6292, opened 2023-10-04
default port for `https.Agent` is not correct


FrancescoMasaia:
### What version of Bun is running?

1.0.4+745b6b94ee56cad24d475799690cc9a89957d15b

### What platform is your computer?

Linux 5.15.0-84-generic x86_64 x86_64

### What steps can reproduce the bug?

```ts
import https from "node:https";
const clientCertAndKey = fs.readFileSync(clientCert); //p12 file
const agent = new https.Agent({
  pfx: clientCertAndKey,
    passphrase: "*************",
});
```

### What is the expected behavior?

Node JS 20.8.0 Generation:
```
Agent {
  _events: [Object: null prototype] {
    free: [Function (anonymous)],
    newListener: [Function: maybeEnableKeylog]
  },
  _eventsCount: 2,
  _maxListeners: undefined,
  defaultPort: 443,
  protocol: 'https:',
  options: [Object: null prototype] {
    pfx: <Buffer 30 80 02 01 03 30 80 06 09 2a 86 48 86 f7 0d 01 07 01 a0 80 24 80 04 82 03 e8 30 80 30 80 06 09 2a 86 48 86 f7 0d 01 07 01 a0 80 24 80 04 82 03 e8 30 ... 4029 more bytes>,
    passphrase: "*************",
    noDelay: true,
    path: null
  },
  requests: [Object: null prototype] {},
  sockets: [Object: null prototype] {},
  freeSockets: [Object: null prototype] {},
  keepAliveMsecs: 1000,
  keepAlive: false,
  maxSockets: Infinity,
  maxFreeSockets: 256,
  scheduling: 'lifo',
  maxTotalSockets: Infinity,
  totalSocketCount: 0,
  maxCachedSessions: 100,
  _sessionCache: { map: {}, list: [] },
  [Symbol(kCapture)]: false
}
```

### What do you see instead?

```
{
  _events: {},
  _eventsCount: 0,
  _maxListeners: undefined,
  [Symbol(kCapture)]: false,
  defaultPort: 80,
  protocol: "http:",
  options: {
[...]

PR #2061, opened 2023-02-13
feat(napi): add `napi_get_value_bigint_words`

ThatOneBro:
Closes #2056 , tested with `node-addon-api`. 

We should probably add tests for N-API stuff soon

PR #11826, opened 2024-06-13
node:https: provide proper Agent definition

nektro:
Closes https://github.com/oven-sh/bun/issues/6292


Jarred-Sumner:
can you add a test that sends an https request using the custom agent and verifies it works as expected

src/js/node/http.ts
```diff
@@ -1591,39 +1586,18 @@ class ClientRequest extends OutgoingMessage {
       options = ObjectAssign(input || {}, options);
     }
 
-    var defaultAgent = options._defaultAgent || Agent.globalAgent;
+    this.#agent = options.agent || Agent.globalAgent;
```
Jarred-Sumner: Let's keep it in a local variable to avoid unnecessary property lookups

```suggestion
    const agent = (this.#agent = options.agent || Agent.globalAgent);
```

src/js/node/http.ts
```diff
@@ -1591,39 +1586,18 @@ class ClientRequest extends OutgoingMessage {
...
 
     let protocol = options.protocol;
     if (!protocol) {
       if (options.port === 443) {
         protocol = "https:";
       } else {
-        protocol = defaultAgent.protocol || "http:";
+        protocol = this.#agent.protocol;
```
Jarred-Sumner:
```suggestion
        protocol = agent.protocol;
```

src/js/node/http.ts
```diff
@@ -1635,14 +1609,14 @@ class ClientRequest extends OutgoingMessage {
 
     // Since we don't implement Agent, we don't need this
     if (protocol !== "http:" && protocol !== "https:" && protocol) {
-      const expectedProtocol = defaultAgent?.protocol ?? "http:";
+      const expectedProtocol = this.#agent.protocol ?? "http:";
```
Jarred-Sumner:
```suggestion
      const expectedProtocol = agent.protocol ?? "http:";
```

src/js/node/http.ts
```diff
@@ -1635,14 +1609,14 @@ class ClientRequest extends OutgoingMessage {
...
       throw new Error(`Protocol mismatch. Expected: ${expectedProtocol}. Got: ${protocol}`);
       // throw new ERR_INVALID_PROTOCOL(protocol, expectedProtocol);
     }
 
     const defaultPort = protocol === "https:" ? 443 : 80;
 
-    this.#port = options.port || options.defaultPort || this.#agent?.defaultPort || defaultPort;
+    this.#port = options.port || options.defaultPort || defaultPort || this.#agent.defaultPort;
```
Jarred-Sumner:
```suggestion
    this.#port = options.port || options.defaultPort || defaultPort || agent.defaultPort;
```

Jarred-Sumner: This code is unreachable actually. Note how defaultPort is set on the line above.

src/js/node/http.ts
Jarred-Sumner: https://github.com/nodejs/node/blob/78a326e2eb236874c45a7937d51e5d3ab3c91a3e/lib/_http_client.js#L154-L167

src/js/node/https.ts
```diff
@@ -1,11 +1,13 @@
 // Hardcoded module "node:https"
 const http = require("node:http");
 
+const ObjectSetPrototypeOf = Object.setPrototypeOf;
+
 function request(input, options, cb) {
   if (input && typeof input === "object" && !(input instanceof URL)) {
-    input.protocol ??= "https:";
```
Jarred-Sumner, line 6: Is this intentional?

src/js/node/http.ts
```diff
@@ -1591,39 +1586,19 @@ class ClientRequest extends OutgoingMessage {
...
+    this.#agent = agent;
 
-    let protocol = options.protocol;
+    let protocol = options.protocol || agent.protocol;
     if (!protocol) {
       if (options.port === 443) {
         protocol = "https:";
       } else {
-        protocol = defaultAgent.protocol || "http:";
+        protocol = agent.protocol;
```
Jarred-Sumner: This code does nothing since line 1592 already checks || agent.protocol, right?

VityaSchel:
we need self signed certificates 🙏

src/js/node/https.ts
nektro: yes

robobun:
@nektro, your commit c4f1f8e10dc4ccf29636198631f4d11e0587c3f7 is building: [`#2061`](https://buildkite.com/bun/bun/builds/2061)

src/js/node/https.ts
```diff
@@ -1,14 +1,30 @@
...
-  if (input && typeof input === "object" && !(input instanceof URL)) {
-    input.protocol ??= "https:";
-  } else if (typeof options === "object") {
-    options.protocol ??= "https:";
+const ObjectSetPrototypeOf = Object.setPrototypeOf;
+const ArrayPrototypeShift = Array.prototype.shift;
+const ObjectAssign = Object.assign;
+const ArrayPrototypeUnshift = Array.prototype.unshift;
+
+function request(...args) {
```
Jarred-Sumner, line 10: Why is this code duplicated in http and https?

nektro: this one assigns `options._defaultAgent`

src/js/node/https.ts
```diff
@@ -17,8 +33,25 @@ function get(input, options, cb) {
...
+function Agent(options) {
+  if (!(this instanceof Agent)) return new Agent(options);
+
+  http.Agent.$apply(this, [options]);
+  this.defaultPort = 443;
+  this.protocol = "https:";
+  this.maxCachedSessions = this.options.maxCachedSessions;
+  if (this.maxCachedSessions === undefined) this.maxCachedSessions = 100;
+}
+Agent.prototype = {};
```
Jarred-Sumner:
```suggestion
Agent.prototype = Object.create(http.Agent.prototype);
```

src/js/node/http.ts
```diff
@@ -2,9 +2,10 @@
 const EventEmitter = require("node:events");
 const { isTypedArray } = require("node:util/types");
 const { Duplex, Readable, Writable } = require("node:stream");
-const { ERR_INVALID_ARG_TYPE } = require("internal/errors");
+const { ERR_INVALID_ARG_TYPE, ERR_INVALID_PROTOCOL } = require("internal/errors");
 const { isPrimary } = require("internal/cluster/isPrimary");
 const { kAutoDestroyed } = require("internal/shared");
+const { urlToHttpOptions } = require("node:url");
```
Jarred-Sumner: if we need to re-use this in multiple modules, let's move it to a separate file so that we don't load all of node:url just for this one function

PR #11830, opened 2024-06-13
shell: Fix memory leak, lazily create `ShellInterpreter` object

### LINKED ISSUE #11816: Memory Leak in `Bun.$` (ShellInterpreter)

1.1.12+43f0913c3

Linux 6.8.4-3-pve x86_64

```ts
import { gc, $ } from "bun";
import { heapStats } from "bun:jsc";
for (let j = 0; j < 10; j++) {
  const promises = [];
    promises.push(
      $`/opt/homebrew/opt/coreutils/libexec/gnubin/cat /tmp/over/input.txt`
    );
  }

  await Promise.all(promises);
  gc(true);
}

console.error("ShellInterpreter count", heapStats().objectTypeCounts.ShellInterpreter);
console.error("RSS", process.memoryUsage.rss() / 1024 / 1024 | 0, "MB");
```
where `/tmp/over/input` is `10 MB` from `/dev/urandom`.
kindly provided by @Jarred-Sumner 

Finished ShellInterpreter should be garbage collected without any side effects.

### What do you see instead?

When the ShellInterpreter finishes the execution it retains the memory and does not release it. Even explicit call to the `Bun.gc` does not collect the retained memory. Depending on frequency of use it can grow up to `+1 GB` per 15-20 mins.

### Additional information

I have Callgrind and DHAT data available for a bigger project for indepth analysis. Please request fom me if it's necessary.

zackradisic:
### What does this PR do?

Fixes #11816

**Problem**
- ShellPromise's `resolve/reject` functions closes over `ShellInterpreter` class
- the `ShellInterpreter` class holds strong references to `resolve/reject` functions
- this makes it possible for the `ShellInterpreter` class to never be deinitialized
- was not caught in shell leak tests because of immediately immediately invoked function syntax: `(async function() { /* shell stuff */ })()`

**Fix**
- lazily create `ShellInterpreter` when the `ShellPromise` is actually awaited
- instead of giving `resolve/reject` to `ShellInterpreter`, give it to new `ParsedShellScript` class which holds script AST and other options
- this breaks cycle: now `resolve/reject` don't refer to `ShellInterpreter`
- put `values: ["resolve", "reject"]` in the `ShellInterpreter` class generator
- give `ShellInterpreter` `resolve/reject` when it gets created

### DIFF
--- a/src/bun.js/api/Shell.classes.ts
+++ b/src/bun.js/api/Shell.classes.ts
@@ -9,7 +9,6 @@ export default [
     hasPendingActivity: true,
     configurable: false,
     klass: {},
-    JSType: "0b11101110",
     values: ["resolve", "reject"],
     proto: {
       run: {
--- a/src/bun.js/api/ShellArgs.classes.ts
+++ b/src/bun.js/api/ShellArgs.classes.ts
@@ -2,27 +2,31 @@ import { define } from "../../codegen/class-definitions";
 
 export default [
   define({
-    name: "ShellInterpreter",
+    name: "ParsedShellScript",
     construct: true,
     noConstructor: true,
     finalize: true,
-    hasPendingActivity: true,
+    hasPendingActivity: false,
     configurable: false,
     klass: {},
     JSType: "0b11101110",
     values: ["resolve", "reject"],
     proto: {
-      run: {
-        fn: "runFromJS",
+      setCwd: {
+        fn: "setCwd",
+        length: 1,
+      },
+      setEnv: {
+        fn: "setEnv",
+        length: 1,
+      },
+      setQuiet: {
+        fn: "setQuiet",
         length: 0,
       },
-      isRunning: {
-        fn: "isRunning",
-        length: 0,
-      },
-      started: {
-        fn: "getStarted",
-        length: 0,
+      setResolveAndReject: {
+        fn: "setResolveAndReject",
+        length: 2,
       },
     },
   }),
--- a/src/bun.js/api/ShellArgs.classes.ts
+++ b/src/bun.js/api/ShellArgs.classes.ts
@@ -10,7 +10,6 @@ export default [
     configurable: false,
     klass: {},
     JSType: "0b11101110",
-    values: ["resolve", "reject"],
     proto: {
       setCwd: {
         fn: "setCwd",
@@ -23,10 +22,6 @@ export default [
       setQuiet: {
         fn: "setQuiet",
         length: 0,
-      },
-      setResolveAndReject: {
-        fn: "setResolveAndReject",
-        length: 2,
       },
     },
   }),
--- a/src/bun.js/api/ShellArgs.classes.ts
+++ b/src/bun.js/api/ShellArgs.classes.ts
@@ -9,7 +9,6 @@ export default [
     hasPendingActivity: false,
     configurable: false,
     klass: {},
-    JSType: "0b11101110",
     proto: {
       setCwd: {
         fn: "setCwd",
--- a/src/shell/interpreter.zig
+++ b/src/shell/interpreter.zig
@@ -1082,7 +1082,9 @@ pub const Interpreter = struct {
                     return Maybe(void).initErr(err);
                 },
             };
-            _ = Syscall.close2(this.cwd_fd);
+            if (comptime !in_init) {
+                _ = Syscall.close2(this.cwd_fd);
+            }
 
             this.__prev_cwd.clearRetainingCapacity();
             this.__prev_cwd.appendSlice(this.__cwd.items[0..]) catch bun.outOfMemory();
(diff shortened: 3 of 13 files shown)


src/shell/interpreter.zig
```diff
@@ -602,11 +602,27 @@ pub const EnvMap = struct {
...
+pub const ShellArgs = struct {
+    /// This is the arena used to allocate the input shell script's AST nodes,
+    /// tokens, and a string pool used to store all strings.
+    arena: bun.ArenaAllocator,
+    /// Root ast node
+    script_ast: ast.Script = .{ .stmts = &[_]ast.Stmt{} },
+
+    /// create new ShellArgs
+    /// the ShellArgs struct itself is allocated in the arena
+    pub fn init(allocator: Allocator) *ShellArgs {
```
```zig
arena: *bun.ArenaAllocator,

pub usingnamespace bun.New(@This());

pub fn allocator(this: *ShellArgs) std.mem.Allocator {
  return this.arena.allocator();
}

pub fn deinit(this: *ShellArgs) void {
  this.arena.deinit();
  this.destroy();
}

pub fn init() *ShellArgs {
  const arena = bun.new(bun.ArenaAllocator);
  arena.* = bun.ArenaAllocator.init(bun.default_allocator);
  return ShellArgs.new(.{
    .arena = arena,
   });
});
```

Jarred-Sumner (changes requested):
Test failures

Zamion101:
Looks like some tests are failing but they're not associated with this PR. Is it possible to get a Canary for this PR or must all the fails be adressed before it's possible.

Jarred-Sumner:
@zackradisic bunshell tests are failing on Windows. I think that is a blocker here.

src/shell/interpreter.zig
```diff
@@ -602,19 +602,233 @@ pub const EnvMap = struct {
...
+
+    pub usingnamespace bun.New(@This());
+
+    pub fn arena_allocator(this: *ShellArgs) std.mem.Allocator {
+        return this.__arena.allocator();
+    }
+
+    pub fn deinit(this: *ShellArgs) void {
+        this.__arena.deinit();
+        this.destroy();
```
Jarred-Sumner, line 621: This leaks the bun.ArenaAllocator pointer.


```suggestion
        bun.destroy(this.__arena);
        this.destroy();
```

src/shell/interpreter.zig
```diff
@@ -602,19 +602,234 @@ pub const EnvMap = struct {
...
+    pub fn createParsedShellScript(
+        globalThis: *JSC.JSGlobalObject,
+        callframe: *JSC.CallFrame,
+    ) callconv(.C) JSValue {
+        var shargs = ShellArgs.init();
+
+        const arguments_ = callframe.arguments(2);
+        var arguments = JSC.Node.ArgumentsSlice.init(globalThis.bunVM(), arguments_.slice());
+        const string_args = arguments.nextEat() orelse {
+            globalThis.throw("shell: expected 2 arguments, got 0", .{});
```
Jarred-Sumner: Memory leak:

```suggestion
            globalThis.throw("shell: expected 2 arguments, got 0", .{});
            shargs.deinit();
```

src/shell/interpreter.zig
```diff
@@ -602,19 +602,234 @@ pub const EnvMap = struct {
...
+    }
+
+    pub fn createParsedShellScript(
+        globalThis: *JSC.JSGlobalObject,
+        callframe: *JSC.CallFrame,
+    ) callconv(.C) JSValue {
+        var shargs = ShellArgs.init();
+
+        const arguments_ = callframe.arguments(2);
+        var arguments = JSC.Node.ArgumentsSlice.init(globalThis.bunVM(), arguments_.slice());
```
Jarred-Sumner: This leaks memory on errors.

Let's just not use ArgumentsSlice here - it's unnecessary. And, let's create the shell args after validating the arguments. 

```suggestion
        const arguments = arguments_.slice();
        
       if (arguments.len < 2) {
            globalThis.throwNotEnoughArguments("Bun.$", 2, arguments.len);
            return .zero;
        }
        const string_args = arguments[0];
        const template_args_jst = arguments[1];
```

src/shell/interpreter.zig
```diff
@@ -936,100 +1158,87 @@ pub const Interpreter = struct {
...
         };
 
-        script_heap.* = script_ast;
+        var shargs: *ShellArgs = undefined;
+        var jsobjs: std.ArrayList(JSValue) = std.ArrayList(JSValue).init(allocator);
+        var quiet: bool = false;
+        var cwd: ?bun.String = null;
+        var export_env: ?EnvMap = null;
+
+        parsed_shell_script.take(
```
Jarred-Sumner, line 1198: What verifies that the shell args in `parsed_shell_scripts` is not null? If it was null, wouldn't this be undefined memory and cause a segfault?

src/bun.js/api/ShellArgs.classes.ts
```diff
@@ -0,0 +1,28 @@
...
+export default [
+  define({
+    name: "ParsedShellScript",
+    construct: true,
+    noConstructor: true,
+    finalize: true,
+    hasPendingActivity: false,
+    configurable: false,
+    klass: {},
+    JSType: "0b11101110",
```
Jarred-Sumner:
```suggestion

```

src/bun.js/api/Shell.classes.ts
```diff
@@ -4,11 +4,13 @@ export default [
   define({
     name: "ShellInterpreter",
     construct: true,
+    noConstructor: true,
     finalize: true,
     hasPendingActivity: true,
     configurable: false,
     klass: {},
     JSType: "0b11101110",
```
Jarred-Sumner:
```suggestion

```

PR #11833, opened 2024-06-13
Support `NODE_TLS_REJECT_UNAUTHORIZED=0` at runtime and implement `BUN_CONFIG_VERBOSE_FETCH`

Jarred-Sumner:
### What does this PR do?

- Fixes a bug with HTTPS keep-alive when `NODE_TLS_REJECT_UNAUTHORIZED=0` is set.
- Implements support for setting `process.env.NODE_TLS_REJECT_UNAUTHORIZED=0` at runtime
- Implements `BUN_CONFIG_VERBOSE_FETCH=1`

image

There is a test for `NODE_TLS_REJECT_UNAUTHORIZED=0` and `NODE_TLS_REJECT_UNAUTHORIZED=1` and the keep-alive fix.

### DIFF
--- a/src/bun.js/bindings/JSEnvironmentVariableMap.cpp
+++ b/src/bun.js/bindings/JSEnvironmentVariableMap.cpp
@@ -12,6 +12,7 @@
 
 #include "BunClientData.h"
 #include "wtf/Compiler.h"
+#include "wtf/Forward.h"
 
 using namespace JSC;
 
@@ -233,14 +234,13 @@ JSC_DEFINE_CUSTOM_SETTER(jsBunConfigVerboseFetchSetter, (JSGlobalObject * global
         return false;
 
     JSValue decodedValue = JSValue::decode(value);
-    if (decodedValue.isString()) {
-        WTF::String str = decodedValue.toWTFString(globalObject);
-        RETURN_IF_EXCEPTION(scope, {});
-        if (str == "1"_s || str == "true"_s) {
-            Bun__setVerboseFetchValue(1);
-        } else {
-            Bun__setVerboseFetchValue(0);
-        }
+    WTF::String str = decodedValue.toWTFString(globalObject);
+    RETURN_IF_EXCEPTION(scope, false);
+
+    if (str == "1"_s || str == "true"_s) {
+        Bun__setVerboseFetchValue(1);
+    } else {
+        Bun__setVerboseFetchValue(0);
     }
 
     const auto& privateName = BUN_CONFIG_VERBOSE_FETCH_PRIVATE_PROPERTY(vm);
--- a/src/bun.js/bindings/JSEnvironmentVariableMap.cpp
+++ b/src/bun.js/bindings/JSEnvironmentVariableMap.cpp
@@ -11,6 +11,7 @@
 #include <JavaScriptCore/JSStringInlines.h>
 
 #include "BunClientData.h"
+#include "wtf/Compiler.h"
 
 using namespace JSC;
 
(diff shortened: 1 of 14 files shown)


src/bun.js/bindings/JSEnvironmentVariableMap.cpp
```diff
@@ -121,6 +122,135 @@ JSC_DEFINE_CUSTOM_SETTER(jsTimeZoneEnvironmentVariableSetter, (JSGlobalObject *
...
+JSC_DEFINE_CUSTOM_SETTER(jsNodeTLSRejectUnauthorizedSetter, (JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::EncodedJSValue value, PropertyName propertyName))
+{
+    VM& vm = globalObject->vm();
+    JSC::JSObject* object = JSValue::decode(thisValue).getObject();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+    if (!object)
+        return false;
+
+    JSValue decodedValue = JSValue::decode(value);
+    if (decodedValue.isString()) {
```
cirospaciari: how does node handle if the user pass like `process.env.NODE_TLS_REJECT_UNAUTHORIZED=0` (boolean or number) we probably should convert/coerce that to string here

src/bun.js/bindings/JSEnvironmentVariableMap.cpp
```diff
@@ -121,6 +122,135 @@ JSC_DEFINE_CUSTOM_SETTER(jsTimeZoneEnvironmentVariableSetter, (JSGlobalObject *
...
+JSC_DEFINE_CUSTOM_SETTER(jsBunConfigVerboseFetchSetter, (JSGlobalObject * globalObject, JSC::EncodedJSValue thisValue, JSC::EncodedJSValue value, PropertyName propertyName))
+{
+    VM& vm = globalObject->vm();
+    JSC::JSObject* object = JSValue::decode(thisValue).getObject();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+    if (!object)
+        return false;
+
+    JSValue decodedValue = JSValue::decode(value);
+    if (decodedValue.isString()) {
```
cirospaciari: same here should we check for `number` and `boolean` too?

test/js/bun/http/bun-server.test.ts
```diff
@@ -407,6 +407,25 @@ describe("Server", () => {
       const result = await fetch(`https://${url}`, { tls: { rejectUnauthorized: false } }).then(res => res.text());
       expect(result).toBe("Hello");
     }
+
+    // Test that HTTPS keep-alive doesn't cause it to re-use the connection on
+    // the next attempt, when the next attempt has reject unauthorized enabled
+    {
+      expect(
+        async () => await fetch(`https://${url}`, { tls: { rejectUnauthorized: true } }).then(res => res.text()),
```
cirospaciari:
```suggestion
        async () => await fetch(server.url, { tls: { rejectUnauthorized: true } }).then(res => res.text()),
```

cirospaciari: we probably should refactor to `const url = server.url.href;` that already includes the protocol (the same for the other places in this test)

cirospaciari (changes requested):
few comments also missing tests for `WebSocket` client and `node:http` but if we already have tests for this and its working the fetch test for the runtime should be enough

src/bun.js/bindings/JSEnvironmentVariableMap.cpp
Jarred-Sumner: https://github.com/nodejs/node/blob/78a326e2eb236874c45a7937d51e5d3ab3c91a3e/lib/internal/options.js#L50

It looks like they stringify and then check if the result is "0"

cirospaciari (approved):

PR #11841, opened 2024-06-13
State the min. version for cross-compilation to Linux, macOS etc

folushooladipo:
According to the below, the ability to cross-compile for custom targets like `--target=bun-linux-x64` was added in v1.1.5. I'm just updating the docs to state that minimum version requirement. https://bun.sh/blog/bun-v1.1.5

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
This warning is too prominent. We need to add a way to say "this feature was added in Bun x.x.x"

PR #11858, opened 2024-06-13
Patch

### LINKED ISSUE #11719: Fix bugs and DX issues with `bun patch`
TODOs for @zackradisic:
- [x] Replace `bun patch-commit` with `bun patch --commit package-name`
- [x] Don't copy it into a temporary directory, instead overwrite each file individually in `node_modules/package-name` with a copy of itself (excluding node_modules, like `patch-package`). This lets developers test their patches locally before committing them, and avoids the issue where hardlinks can potentially overwrite the cached package. This does not handle postinstall scripts 
- [ ] Optional, but nice: `bun patch --revert package-name` 
- [x] `bun patch --commit` needs tests. I am surprised that we merged `bun patch` when `bun patch-commit` had zero test coverage.
- [x] `bun patch --commit` should handle when the patch already exists
- [x] `bun patch --commit` should work with scoped packages
- [x] `bun patch --commit` should never cause "Resolving..." to happen indefinitely on install
- [x] `bun patch` should not be missing from the help menu
- [x] Documentation page

zackradisic:
### What does this PR do?

Fixes #11719

- added docs
- added patch/patch-commit in help
- added `bun patch --commit`
- fix various bugs/issues with patchedDependencies
  - have to calculate patchfile hashes everytime because they can be changed
  - fix infinite loop (stuck on Resolving...)
  - fix package not being updated when patch is removed
  - make it work when a package is already patched
  - fix a bunch of things with patchfile postprocessing
- make patching work directly inside of node_modules, no need for temp dir


Jarred-Sumner:
image

src/cli.zig
```diff
@@ -1005,6 +1005,8 @@ pub const HelpCommand = struct {
         \\  <b><blue>update<r>    <d>{s:<16}<r>     Update outdated dependencies
         \\  <b><blue>link<r>      <d>[\<package\>]<r>          Register or link a local npm package
         \\  <b><blue>unlink<r>                         Unregister a local npm package
+        \\  <b><blue>patch <d>\<pkg\><r>                     Prepare a package for patching
+        \\  <b><blue>patch-commit <d>\<folder\><r>          Install a modified package
```
Jarred-Sumner: ✂️ 
```suggestion

```

src/install/install.zig
```diff
@@ -6727,11 +6725,12 @@ pub const PackageManager = struct {
             .workspaces = true,
         },
         patch_features: union(enum) {
+            NOTHING: struct {},
```
Jarred-Sumner: why is this all caps?

src/install/install.zig
```diff
@@ -10030,161 +10061,433 @@ pub const PackageManager = struct {
...
+            Global.crash();
+        };
+
+        Output.pretty("\nTo patch <b>{s}<r>, edit the following folder:\n\n  <cyan>{s}<r>\n", .{ pkg_name, module_folder });
+        Output.pretty("\nOnce you're done with your changes, run:\n\n  <cyan>bun patch --commit '{s}'<r>\n", .{module_folder});
 
         return;
     }
 
+    fn overwriteNodeModulesFolder(
```
```suggestion
    fn overwritePackageInNodeModulesFolder(
```

src/install/install.zig
```diff
@@ -7252,15 +7251,28 @@ pub const PackageManager = struct {
...
+                    },
+                    .commit => {
+                        this.patch_features = .{
+                            .commit = .{
+                                .patches_dir = cli.patch.commit.patches_dir,
+                            },
+                        };
+                    },
                 }
+                // if (subcommand == .patch) {
```
Jarred-Sumner: can you delete this
```suggestion

```

src/install/install.zig
```diff
@@ -9904,7 +9916,26 @@ pub const PackageManager = struct {
                     );
                 }
             },
-            else => {},
+            else => {
+                if (manager.options.patch_features == .commit) {
+                    // _ = manager.lockfile.loadFromDisk(
```
Jarred-Sumner: did you mean to commit this

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
+    fn nodeModulesFolderForDependencyID(iterator: *Lockfile.Tree.Iterator, dependency_id: DependencyID) !?Lockfile.Tree.NodeModulesFolder {
+        while (iterator.nextNodeModulesFolder(null)) |node_modules| {
+            _ = std.mem.indexOfScalar(DependencyID, node_modules.dependencies, dependency_id) orelse continue;
+            return node_modules;
+        }
+
+        return null;
+    }
+
+    fn pkg_dep_id_for_name_and_version(
```
Jarred-Sumner: can you make this camel case

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
-            .package_name = name,
-            .package_version = resolution_label,
-            // dummy value
-            .node_modules = &dummy_node_modules,
-            .lockfile = manager.lockfile,
-        };
+                const existing_patchfile_hash = existing_patchfile_hash: {
+                    var sfb = std.heap.stackFallback(1024, manager.allocator);
+                    const name_and_version = std.fmt.allocPrint(sfb.get(), "{s}@{}", .{ name, actual_package.resolution.fmt(strbuf, .posix) }) catch unreachable;
+                    defer sfb.get().free(name_and_version);
```
Jarred-Sumner: this leaks memory

.get() resets the buffer

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
-                    .{
-                        e.toSystemError(),
-                    },
+                const pkg = manager.lockfile.packages.get(pkg_id);
+                const pkg_name = pkg.name.slice(strbuf);
+
+                const existing_patchfile_hash = existing_patchfile_hash: {
+                    var sfb = std.heap.stackFallback(1024, manager.allocator);
+                    const name_and_version = std.fmt.allocPrint(sfb.get(), "{s}@{}", .{ name, pkg.resolution.fmt(strbuf, .posix) }) catch unreachable;
+                    defer sfb.get().free(name_and_version);
```
Jarred-Sumner: this will never free

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
+                        defer in_file.close();
+
+                        const mode = in_file.mode() catch @panic("OH NO");
+                        var outfile = createFile(destination_dir_, entrypath, .{ .mode = mode }) catch @panic("OH NO");
+                        defer outfile.close();
+
+                        var buf1: bun.windows.WPathBuffer = undefined;
+                        var buf2: bun.windows.WPathBuffer = undefined;
+
+                        const infile_path_len = bun.windows.kernel32.GetFinalPathNameByHandleW(in_file.handle, &buf1, buf1.len, 0);
```
Jarred-Sumner: We only need to do this once, for the directory and then join it. This would avoid the extra 2 system calls for every file we install

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
+                        }
+
+                        var in_file = openFile(entry.dir, basename, .{ .mode = .read_only }) catch |e| {
+                            Output.prettyError("<r><red>error<r>: opening node_modules file {s} {s} ()", .{ @errorName(e), entrypath });
+                            Global.crash();
+                        };
+                        defer in_file.close();
+
+                        const mode = in_file.mode() catch @panic("OH NO");
+                        var outfile = createFile(destination_dir_, entrypath, .{ .mode = mode }) catch @panic("OH NO");
```
Jarred-Sumner: This will cause AV & other software currently accessing the file to error

Instead, we should close the file immediately before we begin copying.

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
+                            bun.toFD(destination_dir_.fd),
+                            entrypathZ,
+                            bun.toFD(destination_dir_.fd),
+                            tmpname,
+                        ).asErr()) |e| {
+                            Output.prettyError("<r><red>error<r>: copying file {}", .{e});
+                            Global.crash();
+                        }
+
+                        var in_file = openFile(entry.dir, basename, .{ .mode = .read_only }) catch |e| {
```
Jarred-Sumner: Same for `in_file`

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
+                        pathbuf[0..entry.path.len :0],
+                    ).asErr()) |e| {
+                        Output.prettyError("<r><red>error<r>: copying file {}", .{e.withPath(entry.path)});
+                        Global.crash();
+                    }
+
+                    var outfile = try createFile(destination_dir_, entry.path, .{});
+                    defer outfile.close();
+
+                    // debug("createFile {} {s}\n", .{ destination_dir_.fd, entry.path });
```
Jarred-Sumner: This is dead code and should be deleted
```suggestion

```

src/install/install.zig
```diff
@@ -10089,162 +10120,606 @@ pub const PackageManager = struct {
...
+        defer pkg_in_cache_dir.close();
+        var walker = Walker.walk(pkg_in_cache_dir, manager.allocator, &.{}, IGNORED_PATHS) catch bun.outOfMemory();
+        defer walker.deinit();
+        _ = try FileCopier.copy(
+            node_modules_folder,
+            &walker,
+        );
+    }
+
+    // fn overwriteNodeModulesFolder(
```
Jarred-Sumner: delete

src/install/install.zig
```diff
@@ -6048,34 +6044,36 @@ pub const PackageManager = struct {
...
-                                false,
-                                false,
-                            );
+                if (ptask.callback.apply.logger.errors == 0) {
+                    if (comptime @TypeOf(callbacks.onExtract) != void) {
+                        if (ptask.callback.apply.task_id) |task_id| {
+                            _ = task_id; // autofix
+
+                            // const name = manager.lockfile.packages.items(.name)[ptask.callback.apply.pkg_id].slice(manager.lockfile.buffers.string_bytes.items);
+                            // if (!callbacks.onPatch(extract_ctx, name, task_id, log_level)) {
```
```suggestion

```

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+                        out_file_close = false;
+                        in_file.close();
+                        outfile.close();
+                        bun.copyFileWithState(infile_path, outfile_path, &copy_file_state) catch |err| {
+                            Output.prettyError("<r><red>{s}<r>: copying file {}", .{ @errorName(err), bun.fmt.fmtOSPath(entry.path, .{}) });
+                            Global.crash();
+                        };
+
+                        continue;
+                    }
```
Jarred-Sumner: Can you move the rest of this into an `else if (comptime Environment.isPosix)`? It's harder to tell where windows ends and where posix ends

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+            _ = try FileCopier.copy(
+                node_modules_folder,
+                &walker,
+                in_dir,
+                out_dir,
+                &buf1,
+                &buf2,
+            );
+            return;
+        }
```
Jarred-Sumner, line 10529:
```suggestion
        } else if (Environment.isPosix) {
```

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+                        Output.flush();
+                        Global.crash();
+                    },
+                };
 
-                            Output.prettyError("  {s}@<blue>{}<r>\n", .{ pkg.name.slice(strbuf), pkg.resolution.fmt(strbuf, .posix) });
-                        } else break;
+                const existing_patchfile_hash = existing_patchfile_hash: {
+                    var __sfb = std.heap.stackFallback(1024, manager.allocator);
+                    const sfballoc = __sfb.get();
```
```suggestion
                    const allocator = __sfb.get();
```

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+            var buf: [1024]u8 = undefined;
+            const dependencies = lockfile.buffers.dependencies.items;
+
+            var matches_found: u32 = 0;
+            var maybe_first_match: ?struct { DependencyID, PackageID } = null;
+            for (dependencies, 0..) |dep, dep_id| {
+                if (dep.name_hash != name_hash) continue;
+                matches_found += 1;
+                const pkg_id = lockfile.buffers.resolutions.items[dep_id];
+                const pkg = lockfile.packages.get(pkg_id);
```
Jarred-Sumner, line 10136: `pkg_id` can be `invalid_package_id` if it was not resolved, which would cause a segfault here.

This should check that `pkg_id` != `invalid_package_id` and if so continue

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+                if (version) |v| {
+                    const label = std.fmt.bufPrint(buf[0..], "{}", .{pkg.resolution.fmt(strbuf, .posix)}) catch @panic("Resolution name too long");
+                    if (std.mem.eql(u8, label, v)) break :brk .{ @intCast(dep_id), pkg_id };
```
Jarred-Sumner, line 10139: This feels like not a great way to do it but I'm not sure 

@dylan-conway wdyt

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
-        const name_hash = String.Builder.stringHash(name);
+        return .{ pkg_id, dependency_id };
+    }
+
+    const PatchArgKind = enum {
+        path,
+        name_and_version,
 
+        pub fn fromArg(argument: []const u8) PatchArgKind {
+            if (bun.strings.hasPrefix(argument, "node_modules/")) return .path;
```
```suggestion
            if (bun.strings.hasPrefixComptime(argument, "node_modules/")) return .path;
```

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+            if (bun.path.Platform.auto.isAbsolute(argument) and bun.strings.contains(argument, "node_modules/")) return .path;
+            if (comptime bun.Environment.isWindows) {
+                if (bun.strings.hasPrefix(argument, "node_modules\\")) return .path;
```
Jarred-Sumner, line 10184:
```suggestion
                if (bun.strings.hasPrefixComptime(argument, "node_modules\\")) return .path;
```

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+        // The package may be installed using the hard link method,
+        // meaning that changes to the folder will also change the package in the cache.
+        //
+        // So we will overwrite the folder by directly copying the package in cache into it
+        manager.overwritePackageInNodeModulesFolder(cache_dir, cache_dir_subpath, module_folder) catch |e| {
+            Output.prettyError(
+                "<r><red>error<r>: error overwriting folder in node_modules: {s}\n<r>",
+                .{@errorName(e)},
             );
             Output.flush();
```
Jarred-Sumner: crash() flushes

```suggestion

```

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+                        pathbuf[0..entry.path.len :0],
+                    ).asErr()) |e| {
+                        Output.prettyError("<r><red>error<r>: copying file {}", .{e.withPath(entry.path)});
+                        Global.crash();
+                    }
+
+                    var outfile = try createFile(destination_dir_, entry.path, .{});
+                    defer outfile.close();
+
+                    if (comptime Environment.isPosix) {
```
Jarred-Sumner: This code path is posix only, lets remove the `if` to avoid confusing the reader

src/install/install.zig
```diff
@@ -10089,162 +10104,466 @@ pub const PackageManager = struct {
...
+                        }
+
+                        var in_file = openFile(entry.dir, basename, .{ .mode = .read_only }) catch |e| {
+                            Output.prettyError("<r><red>error<r>: opening node_modules file {s} {s} ()", .{ @errorName(e), entrypath });
+                            Global.crash();
+                        };
+                        var in_file_close = true;
+                        defer if (in_file_close) in_file.close();
+
+                        const mode = in_file.mode() catch |e| {
```
Jarred-Sumner: Have you checked that mode() does anything on Windows?

src/install/install.zig
```diff
@@ -10290,85 +10611,159 @@ pub const PackageManager = struct {
...
+                var package = Lockfile.Package{};
+                try package.parseWithJSON(lockfile, manager.allocator, manager.log, package_json_source, json, void, {}, Features.folder);
+
+                const name = lockfile.str(&package.name);
+                const actual_package = switch (lockfile.package_index.get(package.name_hash) orelse {
+                    Output.prettyError(
+                        "<r><red>error<r>: failed to find package in lockfile package index, this is a bug in Bun. Please file a GitHub issue.<r>\n",
+                        .{},
+                    );
+                    Output.flush();
```
```suggestion

```

src/install/install.zig
```diff
@@ -10290,85 +10611,159 @@ pub const PackageManager = struct {
...
+                    const package_json_path = bun.path.joinZ(&[_][]const u8{ argument, "package.json" }, .auto);
+
+                    switch (bun.sys.File.toSource(package_json_path, manager.allocator)) {
+                        .result => |s| break :brk s,
+                        .err => |e| {
+                            Output.prettyError(
+                                "<r><red>error<r>: failed to read package.json: {}<r>\n",
+                                .{e.withPath(package_json_path).toSystemError()},
+                            );
+                            Output.flush();
```
```suggestion

```

src/install/patch_install.zig
```diff
@@ -157,7 +157,7 @@ pub const PatchTask = struct {
         _ = manager; // autofix
         if (this.callback.apply.logger.errors > 0) {
             defer this.callback.apply.logger.deinit();
-            // this.log.addErrorFmt(null, logger.Loc.Empty, bun.default_allocator, "failed to apply patch: {}", .{e}) catch unreachable;
+            Output.printErrorln("failed to apply patchfile ({s})", .{this.callback.apply.patchfilepath});
```
dylan-conway:
```suggestion
            Output.errGeneric("failed to apply patchfile ({s})", .{this.callback.apply.patchfilepath});
```

src/install/patch_install.zig
```diff
@@ -170,42 +170,23 @@ pub const PatchTask = struct {
...
+                Output.prettyErrorln(
+                    fmt,
+                    args,
+                );
+            }
+            if (calc_hash.logger.errors > 0) {
+                Output.prettyErrorln("\n\n", .{});
+                calc_hash.logger.printForLogLevel(Output.writer()) catch {};
+            }
+            Output.flush();
```
dylan-conway, line 188: `flush()` before crash can be deleted, we flush stdout/stderr in our `Bun_atexit` handler

src/install/install.zig
dylan-conway: hmm yeah this won't work if version is a range like `^4.0.0` because the `resolution` will have the exact version. I'm not sure if version ranges are allowed with `pnpm patch` so it might be fine

src/install/install.zig
```diff
@@ -10093,162 +10103,452 @@ pub const PackageManager = struct {
...
-                    Output.flush();
-                    Global.crash();
-                    return;
+            if (matches_found > 1) {
+                Output.prettyErrorln(
+                    "\n<r><red>error<r>: Found multiple versions of <b>{s}<r>, please specify a precise version from the following list:<r>\n",
+                    .{name},
+                );
+                var i: usize = 0;
+                const pkg_hashes = lockfile.packages.items(.name_hash);
```
dylan-conway, line 10155: in the future, this is where you would get the `package_id` for the current workspace and only loop through it's packages, instead of all packages

src/install/install.zig
```diff
@@ -10093,162 +10103,452 @@ pub const PackageManager = struct {
...
+                        in_file_close = false;
+                        out_file_close = false;
+                        in_file.close();
+                        outfile.close();
+                        bun.copyFileWithState(infile_path, outfile_path, &copy_file_state) catch |err| {
+                            Output.prettyError("<r><red>{s}<r>: copying file {}", .{ @errorName(err), bun.fmt.fmtOSPath(entry.path, .{}) });
+                            Global.crash();
+                        };
+
+                        continue;
```
```suggestion

```

PR #11865, opened 2024-06-14
Print like curl

Jarred-Sumner:
### What does this PR do?

image

### DIFF
--- a/src/bun.js/webcore/response.zig
+++ b/src/bun.js/webcore/response.zig
@@ -1686,7 +1686,7 @@ pub const Fetch = struct {
             disable_decompression: bool,
             reject_unauthorized: bool,
             url: ZigURL,
-            verbose: bool = false,
+            verbose: http.HTTPVerboseLevel = .none,
             redirect_type: FetchRedirect = FetchRedirect.follow,
             proxy: ?ZigURL = null,
             url_proxy_buffer: []const u8 = "",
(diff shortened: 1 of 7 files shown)


src/bun.js/webcore/response.zig
```diff
@@ -2069,8 +2069,17 @@ pub const Fetch = struct {
                             }
                         }
 
-                        if (options.get(globalThis, "verbose")) |verb| {
-                            verbose = verb.toBoolean();
+                        if (options.get(globalThis, "verbose")) |verb| verbose: {
+                            if (verb.isString()) {
+                                if (verb.getZigString(globalThis).eqlComptime("curl")) {
```
dylan-conway, line 2074: do we also need to for `"1"` and `"true"`?

src/deps/picohttp.zig
```diff
@@ -60,6 +77,61 @@ pub const Request = struct {
...
+
+            for (request.headers) |*header| {
+                _ = try writer.writeAll(" ");
+                if (content_type.len == 0) {
+                    if (bun.strings.eqlCaseInsensitiveASCII("content-type", header.name, true)) {
+                        content_type = header.value;
+                    }
+                }
+
+                try header.curl().format(comptime "", .{}, writer);
```
Jarred-Sumner:
```suggestion
                try header.curl().format("", .{}, writer);
```

src/bun.js/webcore/response.zig
Jarred-Sumner: nah

Issue #4015, opened 2023-08-06
Implement an `lcov` coverage reporter for `bun test`


Jarred-Sumner:
[Format](https://ltp.sourceforge.net/coverage/lcov/geninfo.1.php)

The only thing we are missing for lcov to work well is function names. JSC doesn't include them. 

Otherwise, I think we can just write the file

hrgdavor:
I switched a small project to `bun test` because jsx just works, and I do not think nodejs will allow that in their native test runner anytime soon (or if added would be slow likely).

current output in console is usable, but `lcov ` will be great, for integrating and generating pretty html reports.

jbergstroem:
> current output in console is usable, but lcov  will be great, for integrating and generating pretty html reports.

..and uploading reports to tools such as codecov and coveralls.

Haratsu:
Any progress on uploadable coverage reports?
Alternatively has anyone _successfully_ integrated jest cli commands into bun as a work around already? So that we could call `bun  jest ... ` my inital attempt doing so failed complaining:

```
Test suite failed to run

    TypeError: undefined is not an object (evaluating '_global.hasOwnProperty("queueMicrotask")')
        at hasOwnProperty (native)

      at withGlobal (node_modules/@sinonjs/fake-timers/src/fake-timers-src.js:183:23)
      at new FakeTimers (node_modules/@jest/fake-timers/build/modernFakeTimers.js:27:50)
      at processTicksAndRejections (:12:39)
      at processTicksAndRejections (:12:39)
```

Before digging into deep analysis I wanted to recheck

nemanjamitric:
> Any progress on uploadable coverage reports? Alternatively has anyone _successfully_ integrated jest cli commands into bun as a work around already? So that we could call `bun jest ... ` my inital attempt doing so failed complaining:
> 
> ```
> Test suite failed to run
> 
>     TypeError: undefined is not an object (evaluating '_global.hasOwnProperty("queueMicrotask")')
>         at hasOwnProperty (native)
> 
>       at withGlobal (node_modules/@sinonjs/fake-timers/src/fake-timers-src.js:183:23)
>       at new FakeTimers (node_modules/@jest/fake-timers/build/modernFakeTimers.js:27:50)
>       at processTicksAndRejections (:12:39)
>       at processTicksAndRejections (:12:39)
> ```
> 
> Before digging into deep analysis I wanted to recheck

[...]

PR #11883, opened 2024-06-15
Implement initial LCOV reporter (no function names support)

exoego:
### What does this PR do?

Closes https://github.com/oven-sh/bun/issues/4015

This adds new config to bunfig which allows users to use `lcov` coverage reporter and also disable `console` reporter.

- [x] Code changes

AFAIK, function names in LCOV is optional, and tools like CodeCov accept LCOV file without function names.
So I think LCOV without function names is still valuable to users.

I run the locally-built `bun-debug` over `https://github.com/honojs/hono` and obtained LCOV file like below:

```lcov
...
TN:
SF:src/adapter/bun/index.ts
FNF:1
FNH:1
DA:6,45
DA:7,51
DA:8,49
DA:9,40
LF:10
LH:4
end_of_record
...
```

Then I uploaded the generated LCOV to CodeCov and confirmed files are succesfully processed by CodeCov:
image
image

Sorry, I couldn't figure out how to write test for this.
Happy to add test if guidance provided 🙇 


exoego:
I am happy to finish this PR if I get advised what is missing.

src/cli/test_command.zig
```diff
@@ -271,6 +271,49 @@ pub const CommandLineReporter = struct {
...
+            byte_ranges.appendAssumeCapacity(value);
+        }
+
+        if (byte_ranges.items.len == 0) {
+            return;
+        }
+
+        std.sort.pdq(bun.sourcemap.ByteRangeMapping, byte_ranges.items, void{}, bun.sourcemap.ByteRangeMapping.isLessThan);
+
+        iter = map.valueIterator();
```
paperclover: ~~is this used?~~ clarification: this assignment is not used

src/cli/test_command.zig
```diff
@@ -271,6 +271,49 @@ pub const CommandLineReporter = struct {
...
+
+        var map = bun.sourcemap.ByteRangeMapping.map orelse return;
+        var iter = map.valueIterator();
+        const relative_dir = vm.bundler.fs.top_level_dir;
+
+        var byte_ranges = try std.ArrayList(bun.sourcemap.ByteRangeMapping).initCapacity(bun.default_allocator, map.count());
+
+        while (iter.next()) |entry| {
+            const value: bun.sourcemap.ByteRangeMapping = entry.*;
+            byte_ranges.appendAssumeCapacity(value);
```
paperclover: cant this just be written as appendAssumeCapacity(value.*)

src/cli/test_command.zig
```diff
@@ -572,11 +615,21 @@ pub const TestCommand = struct {
     pub const name = "test";
     pub const CodeCoverageOptions = struct {
         skip_test_files: bool = !Environment.allow_assert,
+        reporters: []const Reporter = &.{ Reporter.console },
```
paperclover: I think instead of a slice, we should just have a struct of all of the coverage reporters.

```
const Reporters = struct {
  console: bool,
  lcov: bool,
};
```

The current way you have makes it possible to specify a coverage reporter more than once, which duplicates work and is likely unintentional. This also would reduce allocations.

src/bunfig.zig
```diff
@@ -252,6 +253,26 @@ pub const Bunfig = struct {
                         this.ctx.test_options.coverage.enabled = expr.data.e_boolean.value;
                     }
 
+                    if (test_.get("coverageReporters")) |expr| {
+                        try this.expect(expr, .e_array);
+                        const items = expr.data.e_array.items.slice();
+                        const reporters = try this.allocator.alloc(TestCommand.Reporter, items.len);
+                        for (items, 0..) |item, i| {
+                            try this.expectString(item);
+                            if (TestCommand.ReporterMap.get(item.asString(bun.default_allocator) orelse "")) |reporter| {
```
paperclover: so this instead of using a map here, it could just be a simple chain of if else if else and `bun.strings.eqlComptime(x, "lcov")` etc

src/bunfig.zig
```diff
@@ -252,6 +253,26 @@ pub const Bunfig = struct {
...
+                                reporters[i] = reporter;
+                            } else {
+                                try this.addError(item.loc, "Invalid reporter.");
```
paperclover: can this say `Invalid coverage reporter "{s}"`?

paperclover (changes requested):
overall this looks good. some notes attached.

in addition

- kind of sad that lines with comments are marked as not hit. though i dont think that is an issue in your implementation, but the coverage stuff overall.
- should we get `--coverage-reporter=` command line arguments?
- for tests, I think it would be reasonable to snapshot test the samples. look for usages of `tempDirWithFiles` to setup directory trees to test.

src/cli/test_command.zig
exoego: No, it doesn't compile on my end:

```
src/cli/test_command.zig:292:51: error: cannot dereference non-pointer type 'src.sourcemap.CodeCoverage.ByteRangeMapping'
            byte_ranges.appendAssumeCapacity(value.*);
                                             ~~~~~^~
src/sourcemap/CodeCoverage.zig:307:30: note: struct declared here
pub const ByteRangeMapping = struct {
                             ^~~~~~
referenced by:
    exec: src/cli/test_command.zig:920:37
```

src/cli/test_command.zig
exoego: Removed https://github.com/oven-sh/bun/pull/11883/commits/b3f7ae8d4b606639d7fc714a29f2f95f78ebda39

src/cli/test_command.zig
exoego: Did so in https://github.com/oven-sh/bun/pull/11883/commits/9ab1a772844f6e02f285a68b01f783556c974d7c

src/bunfig.zig
exoego: Did so in https://github.com/oven-sh/bun/pull/11883/commits/a426f3e45eff522f65aa4c6596d15726d37398dc

src/bunfig.zig
exoego: Did so in https://github.com/oven-sh/bun/pull/11883/commits/19db90c71eb87ba7c7a998208c8bed28befc6065

exoego:
@paperdave Thanks for your review. I think all comments are addressed

src/cli/test_command.zig
paperclover: oh wait sorry, i meant entry.*

the point being these two lines can be one line (i dont think the type annotation is of much value)

paperdave:
@exoego see the two bullet points in the main 'requested changes' comment. to repeat:

- this at least needs tests to verify it works on all platforms, and to make sure it doesnt just stop working one day.
- consider (but not mandatory) adding a CLI flag like `--coverage-reporter`

but otherwise we're good

src/cli/test_command.zig
exoego: Oh, it worked https://github.com/oven-sh/bun/pull/11883/commits/54d12d3abb078c0ec6f208619e1d804a07a937a7
(Sorry, I am a super newbie to Ziglang)

exoego:
Added snapshot test https://github.com/oven-sh/bun/pull/11883/commits/75476db45481c751b714acf72ccc968fcbb88b33
But `bun test --update-snapshots` does not even create `coverage/` directory...
I miss some steps to test `bun` changes?

paperclover (approved):
im going to add a commit that renames --coverage-reports-dir to just --coverage-dir but yeah this is good.

it could use more tests, but i want to get this in for todays release.

paperclover (changes requested):
wait, nevermind, coverage.test.ts is failing

exoego:
Hmm, do you have an advice to fix coverage.test.ts?

src/sourcemap/CodeCoverage.zig
```diff
@@ -687,174 +709,6 @@ pub const ByteRangeMapping = struct {
...
-
-        // ** Branch coverage not supported yet, since JSC does not support those yet. ** //
-        // BRDA: line, block, (expressions,count)+
-        // BRF: branches found
-        // BRH: branches hit
-
-        while (iter.next()) |line| {
-            // DA: line number, hit count
-            // FIXME: Remove the magic number that is used to convert the hit count to the correct value
-            try writer.print("DA:{d},{d}\n", .{ line + 1, report.lines_hits.ptr[line] - 2863311530 });
```
Jarred-Sumner: @exoego the reason this magic number made it seemingly work: undefined memory. In safety-checked builds of  Zig (debug, ReleaseSafe), `0xa` is the default-initialized value for memory.

This magic number was turning `0xa` into 0. In a release build, this is undefined behavior and will produce a result which will either crash the program, infinite loop, or something unexpected. This is why it did not work in CI.

src/sourcemap/CodeCoverage.zig
```diff
@@ -388,8 +446,12 @@ pub const ByteRangeMapping = struct {
             line_count = @truncate(line_starts.len);
             executable_lines = try Bitset.initEmpty(allocator, line_count);
             lines_which_have_executed = try Bitset.initEmpty(allocator, line_count);
-            lines_hits = try LinesHits.initCapacity(allocator, line_count);
-            errdefer lines_hits.deinitWithAllocator(allocator);
+            line_hits = try LinesHits.initCapacity(allocator, line_count);
+            line_hits.len = line_count;
+            const line_hits_slice = line_hits.slice();
+            @memset(line_hits_slice, 0);
```
Jarred-Sumner, line 452: @exoego note how we now zero-initialize the `line_hits` array to ensure that we can access values which have never been written to without worrying about undefined behavior.

src/sourcemap/CodeCoverage.zig
```diff
@@ -415,7 +477,7 @@ pub const ByteRangeMapping = struct {
                     executable_lines.set(line);
                     if (has_executed) {
                         lines_which_have_executed.set(line);
-                        lines_hits.ptr[line] += 1;
+                        line_hits_slice[line] += 1;
```
Jarred-Sumner, line 480: @exoego writing to a many-item pointer (`[*]MyType`) directly is very much a footgun. You want to use a slice (`ptr[0..len]`), which has bounds checks.

exoego:
Thanks!!

src/cli/test_command.zig
```diff
@@ -296,76 +290,215 @@ pub const CommandLineReporter = struct {
...
+                },
+                .sync,
+            );
+
+            // Write the lcov.info file to a temporary file we atomically rename to the final name after it succeeds
+            var base64_bytes: [8]u8 = undefined;
+            var shortname_buf: [512]u8 = undefined;
+            bun.rand(&base64_bytes);
+            const tmpname = std.fmt.bufPrintZ(&shortname_buf, ".lcov.info.{s}.tmp", .{bun.fmt.fmtSliceHexLower(&base64_bytes)}) catch unreachable;
+            const path = bun.path.joinAbsStringBufZ(relative_dir, &lcov_name_buf, &.{ opts.reports_directory, tmpname }, .auto);
```
afonsojramos, line 383: Running multiple times is leaving the folder with multiple tml files, is this expected?
image

Issue #6591, opened 2023-10-19
Allow bun:sqlite bind parameters to a prepared statement with object


drag0n-app:
### What is the problem this feature would solve?

the docs example (https://bun.sh/docs/api/sqlite)
```ts
const query = db.query("SELECT * FROM foo WHERE bar = $bar");
const results = query.all({
  $bar: "bar",
});
```
Binding javascript object directly
```ts
const query = db.query("SELECT * FROM foo WHERE bar = $bar");
const results = query.all({
  bar: "bar",
});
```

### What is the feature you are proposing to solve the problem?

Binding javascript object directly is easier, No need to add key prefixes (: or ? or @) yourself.

### What alternatives have you considered?

refer to https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#binding-parameters

drag0n-app:
Anyone help make it happen?

oliverjam:
Yeah it's awkward having to manually write out every key with a `$` prefix, especially when you're inserting a larger object. It's especially annoying with TypeScript since you end up with identical duplicate types; one for select with normal keys and one for insert with `$` prefixes. Then you have to make sure you keep them in sync...

E.g. here's something I just had to write:

```ts
type Article = {
  slug: string;
  title: string;
  time: number;
  date: string;
  intro: string;
  content: string;
  draft: 0 | 1;
};

type ArticleIn = {
  $slug: string;
  $title: string;
  $time: number;
  $date: string;
  $intro: string;
  $content: string;
  $draft: 0 | 1;
};

function create(article: ArticleIn): Article {
  // ...
}

// ...elsewhere
create({
  $slug,
  $title,
  $time: Number(time),
  $date,
  $intro,
  $content,
  $draft: draft ? 1 : 0,
});
```

Edit: I came up with a silly helper to avoid duplicating types at least 🙃 

```ts
type Dollarize<T> = {
    [K in keyof T as `$${string & K}`]: T[K]
}

type ArticleIn = Dollarize<Article>
//   ^? type ArticleIn = { $slug: string ... }
```

Issue #5661, opened 2023-09-18
bun:sqlite truncates integers that cannot fit in a double


JackTYM:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Linux 5.4.0-153-generic x86_64 x86_64

### What steps can reproduce the bug?

Create a SQLite row with an INT or DOUBLE value over 53 bits
Select the row using .get() or .values() and print it
Resulting numbers are changed

### What is the expected behavior?

The unchanged version of the number should be given

### What do you see instead?

INT provided with value 990760989492400188
Object given with value 990760989492400100

### Additional information

Seemingly happens from numbers over 53 bits

JackTYM:
Current workaround is to just store text, but it is preferable if I could use normal integers or some sort of number data structure

Prinzhorn:
https://github.com/oven-sh/bun/issues/1536 https://github.com/oven-sh/bun/issues/5256

nektro:
53 bits is the max of `Number.MAX_SAFE_INTEGER` for baseline javascript numbers which use the same data type as DOUBLE. (`9007199254740991` for reference).

for clarification:
is the expectation that the number would then instead return in a BigInt instead? was the input already a BigInt?

Issue #5261, opened 2023-09-13
bun:sqlite merges columns inconsistently from better-sqlite3 and javascript in general


HarbingerNight:
### What version of Bun is running?

1.0.0

### What platform is your computer?

Linux 5.10.102.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

bun:sqlite seems to merge sqlite columns with the same name the opposite way of better-sqlite3 and { ,,,x, ...y } would.

it keeps the first column of that name and discards the rest

```javascript
import { Database } from 'bun:sqlite'
const ds = new Database
ds.run( 'create table x ( i integer primary key )' )

ds.query( 'insert into x values ( ?1 )' )
  .run( [ 0x7fffffffffffffffn ] )

ds.query( 'select *, cast ( i as text ) as i from x' ).get( )
ds.query( 'select cast ( i as text ) as i, * from x' ).get( )
```

### What is the expected behavior?

when specifying a query that reuses the column name such as using the sql all columns operator `*` followed by a cast of one of the columns the columns should be merged so that the last specified version of the column is in the results.

This is consistent with both better-sqlite and { ...x, ...y }.

```javascript
> ds.query(  'select *, cast ( i as text ) as i from x' ).get( )
{ i: '9223372036854775807' }
> ds.query( 'select *, cast ( i as text ) as i from x' ).get( )
{ i: 9223372036854776000 }
```

### What do you see instead?

I get the opposite result:

```javascript
> ds.query(  'select *, cast ( i as text ) as i from x' ).get( )
{ i: 9223372036854776000 }
> ds.query( 'select *, cast ( i as text ) as i from x' ).get( )
{ i: '9223372036854775807' }
```

### Additional information

This is helpful to do when you need to cast one of the columns and want to get the rest of the columns as well. It should be consistent w/ better-sqlite3 for code portability.

PR #11887, opened 2024-06-15
[bun:sqlite] Support unprefixed bindings, safe integers / BigInt, `as(Class)`

Jarred-Sumner:
## What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/5661
Fixes https://github.com/oven-sh/bun/issues/5261
Fixes https://github.com/oven-sh/bun/issues/6591
Fixes https://github.com/oven-sh/bun/issues/5256 (bigint)
Fixes https://github.com/oven-sh/bun/issues/1536
Fixes https://github.com/oven-sh/bun/issues/8284
Fixes https://github.com/oven-sh/bun/issues/3284

This adds a new `strict?: boolean` to the `Database` constructor which lets you omit the `$`, `@`, or `:` prefix when binding values to prepared statements.

After:
```ts

const db = new Database(":memory:", {
  // bind values without prefixes
  // throw if a field is missing in the parameters
  strict: true,
});

const query = db.query(`select $message;`);

query.all({ message: "Hello world" });
```

Before:
```ts

const db = new Database(":memory:", {
  // bind values without prefixes
  strict: false,
});

const query = db.query(`select $message;`);

query.all({ $message: "Hello world" });
```

### `safeIntegers: true`

When `safeIntegers` is `true`, `bun:sqlite` will return integers as `bigint` types:

```ts

const db = new Database(":memory:", { safeIntegers: true });
const query = db.query(
  `SELECT ${BigInt(Number.MAX_SAFE_INTEGER) + 102n} as max_int`,
);
const result = query.get();
console.log(result.max_int); // => 9007199254741093n
```

When `safeIntegers` is `true`, `bun:sqlite` will throw an error if a `bigint` value in a bound parameter exceeds 64 bits:

```ts

const db = new Database(":memory:", { safeIntegers: true });
db.run("CREATE TABLE test (id INTEGER PRIMARY KEY, value INTEGER)");

const query = db.query("INSERT INTO test (value) VALUES ($value)");

try {
  query.run({ $value: BigInt(Number.MAX_SAFE_INTEGER) ** 2n });
} catch (e) {
  console.log(e.message); // => BigInt value '81129638414606663681390495662081' is out of range
}
```

### `.as(Class)` - attach methods & getters/setters to results

Use `.as(Class)` to run a query and get back the results as instances of a class.

```ts
class Movie {
  title: string;
  year: number;

  get isMarvel() {
    return this.title.includes("Marvel");
  }
}

const query = db.query("SELECT title, year FROM movies").as(Movie);
const movies = query.all();
const first = query.get();
console.log(movies[0].isMarvel); // => true
console.log(first.isMarvel); // => true
```

As a performance optimization, the class constructor is not called, default initializers are not run, and private fields are not accessible. This is more like using `Object.create` than `new`. The class's prototype is assigned to the object, methods are attached, and getters/setters are set up, but the constructor is not called.

The database columns are set as properties on the class instance.

There are tests


test/js/bun/sqlite/sqlite.test.js
```diff
@@ -437,6 +437,93 @@ it("db.query()", () => {
   db.close();
 });
 
+it("db.run()", () => {
+  const db = Database.open(":memory:");
+
+  db.exec("CREATE TABLE cats (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT UNIQUE NOT NULL, age INTEGER NOT NULL)");
+
+  const insert = db.query("INSERT INTO cats (name, age) VALUES (@name, @age) RETURNING name").all({
```
dylan-conway, line 613: is this test complete?

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -273,6 +285,37 @@ static JSValue createSQLiteError(JSC::JSGlobalObject* globalObject, sqlite3* db)
...
+    SQLiteBindingsMap(JSC::VM& vm, sqlite3_stmt* stmt, bool trimLeadingPrefix = false, size_t count = 0)
+    {
+        bindingNames.reserveInitialCapacity(count);
+        hasLoaded = true;
+        size_t prefixOffset = trimLeadingPrefix ? 1 : 0;
+
+        for (size_t i = 0; i < count; i++) {
+            const unsigned char* name = reinterpret_cast<const unsigned char*>(sqlite3_bind_parameter_name(stmt, i + 1));
+            if (name == nullptr) {
+                break;
```
dylan-conway: should this be `continue` instead of `break`. docs say `NULL` is returned if out of range or if it's nameless

Jarred-Sumner:
The failing sql-raw test should probably be deleted. It is now difficult to use the bindings directly that way, and we shouldn't really expose it anyway.

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -671,63 +806,174 @@ static inline bool rebindValue(JSC::JSGlobalObject* lexicalGlobalObject, sqlite3
...
+    // Does it only contain indexed properties?
+    //
+    // { 0: "foo", 1: "bar", "2": "baz" }
+    //
+    else if (UNLIKELY(bindings.isOnlyIndexed)) {
+        for (size_t i = 0; i < size; i++) {
+            JSValue value = target->getDirectIndex(globalObject, i);
+            if (!value && !scope.exception()) {
+                if (throwOnMissing) {
+                    throwException(globalObject, scope, createError(globalObject, makeString("Missing parameter \""_s, i + 1, "\""_s)));
```
cirospaciari, line 890: If we throw or have an exception should we return? should we add  `RETURN_IF_EXCEPTION(scope, JSValue());` after the loop (in case we throw and skip all other bindings)

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -671,63 +806,174 @@ static inline bool rebindValue(JSC::JSGlobalObject* lexicalGlobalObject, sqlite3
...
+    //
+    // { foo: "bar", baz: "qux" }
+    //
+    else if (target->canUseFastGetOwnProperty(structure)) {
+        for (size_t i = 0; i < size; i++) {
+            const auto& property = bindingNames[i];
+            JSValue value = property.isEmpty() ? target->getDirectIndex(globalObject, i) : target->fastGetOwnProperty(vm, structure, bindingNames[i]);
+            if (!value && !scope.exception()) {
+                if (throwOnMissing) {
+                    throwException(globalObject, scope, createError(globalObject, makeString("Missing parameter \""_s, property.isEmpty() ? String::number(i) : property.string(), "\""_s)));
```
cirospaciari, line 916: same here

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -671,63 +806,174 @@ static inline bool rebindValue(JSC::JSGlobalObject* lexicalGlobalObject, sqlite3
...
+            count++;
         }
+    } else {
+        for (size_t i = 0; i < size; i++) {
+            PropertySlot slot(target, PropertySlot::InternalMethodType::GetOwnProperty);
+            const auto& property = bindingNames[i];
+            bool hasProperty = property.isEmpty() ? target->methodTable()->getOwnPropertySlotByIndex(target, globalObject, i, slot) : target->methodTable()->getOwnPropertySlot(target, globalObject, property, slot);
+            if (!hasProperty && !scope.exception()) {
+                if (throwOnMissing) {
+                    throwException(globalObject, scope, createError(globalObject, makeString("Missing parameter \""_s, property.isEmpty() ? String::number(i) : property.string(), "\""_s)));
```
cirospaciari, line 938: and here

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -742,15 +988,16 @@ static JSC::JSValue rebindStatement(JSC::JSGlobalObject* lexicalGlobalObject, JS
         return jsNumber(0);
     }
 
-    if (count != max) {
-        throwException(lexicalGlobalObject, scope, createError(lexicalGlobalObject, "Expected " + String::number(max) + " values, got " + String::number(count)));
+    int required = bindings.count;
+    if (count != required) {
+        throwException(lexicalGlobalObject, scope, createError(lexicalGlobalObject, makeString("SQLite query expected "_s, required, " values, received "_s, count)));
```
cirospaciari, line 993: after throw should we return empty/zero instead of undefined?

cirospaciari (changes requested):

packages/bun-types/sqlite.d.ts
```diff
@@ -82,6 +82,40 @@ declare module "bun:sqlite" {
...
+
+            /**
+             * When set to `true`, integers are returned as `bigint` types.
+             *
+             * When set to `false`, integers are returned as `number` types and truncated to 52 bits.
+             *
+             * @default false
+             * @since v1.1.14
+             */
+            safeInteger?: boolean;
```
Electroid, line 94: Name is confusing, just name it `bigint: boolean`

packages/bun-types/sqlite.d.ts
```diff
@@ -1040,4 +1112,21 @@ declare module "bun:sqlite" {
...
+  interface Changes {
+    /**
+     * The number of rows changed by the last `run` or `exec` call.
+     */
+    changes: number;
+
+    /**
+     * If `safeIntegers` is `true`, this is a `bigint`. Otherwise, it is a `number`.
+     */
+    lastInsertRowId: number | bigint;
```
Electroid: `better-sqlite3` uses `id` instead of `Id`. Do we want to at least support an alias?

Jarred-Sumner: yeah

packages/bun-types/sqlite.d.ts
Jarred-Sumner: that's what i had originally, but this is consistent with the rest. It also implies that bigint is not supported - which is not true.

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
Jarred-Sumner: It checks two lines below

https://github.com/oven-sh/bun/blob/0ba290c5395b6d9393a1069b68d4d1aae64ae259/src/bun.js/bindings/sqlite/JSSQLStatement.cpp#L922

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
Jarred-Sumner: Checks two lines below

https://github.com/oven-sh/bun/blob/0ba290c5395b6d9393a1069b68d4d1aae64ae259/src/bun.js/bindings/sqlite/JSSQLStatement.cpp#L936-L944

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
Jarred-Sumner: Yes, it does that a few lines below

https://github.com/oven-sh/bun/blob/0ba290c5395b6d9393a1069b68d4d1aae64ae259/src/bun.js/bindings/sqlite/JSSQLStatement.cpp#L896

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
Jarred-Sumner: Both work. We aren't consistent right now

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
cirospaciari: But this is inside the loop right?

Jarred-Sumner: https://github.com/oven-sh/bun/blob/c0ab3ef88b73d7e6805bc22ee98d8159e38d972e/src/bun.js/bindings/sqlite/JSSQLStatement.cpp#L977

cirospaciari (approved):

PR #11906, opened 2024-06-16
[internal] Add `getRuntimeFeatureFlagState`

Jarred-Sumner:
### What does this PR do?

Small refactor in preparation for auto-updating canary builds

### DIFF
--- a/src/env_loader.zig
+++ b/src/env_loader.zig
@@ -80,12 +80,20 @@ pub const Loader = struct {
         return null;
     }
 
+    pub const CI_ENVS = .{
+        "CI",
+        "TDDIUM",
+        "GITHUB_ACTIONS",
+        "JENKINS_URL",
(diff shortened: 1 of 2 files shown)


src/env_loader.zig
```diff
@@ -80,12 +80,20 @@ pub const Loader = struct {
...
+    };
+
     pub fn isCI(this: *const Loader) bool {
-        return (this.get("CI") orelse
-            this.get("TDDIUM") orelse
-            this.get("GITHUB_ACTIONS") orelse
-            this.get("JENKINS_URL") orelse
-            this.get("bamboo.buildKey")) != null;
+        inline for (CI_ENVS) |env| {
+            if (this.has(env)) return true;
```
Electroid, line 93: We should check that the env var is not:
* empty
* `0`
* `false`

PR #11909, opened 2024-06-16
Change "what's new" link to bun blog instead of release page

ShrootBuck:
Instead of linking to the release, the "what's new" link shown on `bun upgrade` links to the relevant blog post.


Electroid (approved):
Thanks, this is a good change.

paperdave:
For a version like `1.1` the trailing `.0` needs to be removed, for example: https://bun.sh/blog/bun-v1.1

PR #11916, opened 2024-06-17
docs(guides): add example how to parse query params for Bun.serve

BitOfUniverse:
### What does this PR do?

Guide describes all basic routing logic, but it doesn't tell anything about query params.
I've added a small snippet to make this guide complete.


Araxeus:
1. Why the convoluted `Object.fromEntries(searchParams.entries())`?
   It should just show something basic like `searchParams.get('token')` 

3. The example now creates a URL twice for no reason (`new URL(req.url)`)

Issue #10641, opened 2024-04-29
Git ssh dependency not resolving as ssh


Blechlawine:
### What version of Bun is running?

1.1.6+e58d67b46

### What platform is your computer?

Linux 6.5.0-28-generic x86_64 x86_64

### What steps can reproduce the bug?

when trying to install a git dependency with ssh (git+ssh://...) from a private github repository it asks for username and password for https://github.com, github deprecated password authentication though, so that doesn't work. I already have an ssh key for github with which cloning the private repo works.

### What is the expected behavior?

I would expect it to simply install the dependency from the github repository over ssh

### What do you see instead?

```
bun install v1.1.6 (e58d67b4)
Username for 'https://github.com':   🔍 Resolving...

```
The `🔍 Resolving...` slowly moves to the left

### Additional information

Is there a reason why bun doesn't want to download git dependencies over ssh?

Issue #6927, opened 2023-11-06
Unable to install packages from private bitbucket repo


LucyEgan:
### What version of Bun is running?

1.0.9+98f20170a

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

Create a private repo on bitbucket
try and install the package from `bitbucket:YOUR_ACCOUNT/YOUR_REPO`

### What is the expected behavior?

Install the package from the private repo using current ssh key 

### What do you see instead?

It attempts to install using https instead of ssh 
<img width="593" alt="image" src="https://github.com/oven-sh/bun/assets/5268652/31ffa7c7-5b39-4983-ab67-02381ed76448">

### Additional information

Seems related to https://github.com/oven-sh/bun/issues/5027 but thats marked as completed.

ngHblanchard:
Same issue here. Trying to install packages in a circleci pipeline: 
the script stops on the resolve step, asking for a password, even if the package is included like this : `git+ssh://git@bitbucket.org/YOUR_ACCOUNT/YOUR_REPO#version`.

Issue #2464, opened 2023-03-23
`bun install` fails to git clone private Enterprise GitHub dependency


privatenumber:
### What version of Bun is running?

0.5.8

### What platform is your computer?

Darwin 22.3.0 x86_64 i386

### What steps can reproduce the bug?

Install with the following `package.json`  where "@org/website-i18n" points to a private GitHub repo.

```json
{
  "dependencies": {
    "@org/website-i18n": "github:org/website-i18n#branch-name"
  }
}
```

### What is the expected behavior?

For it to install with the appropriate Git credentials.

### What do you see instead?

It fails.

Also note, the first error message truncates the package name for some reason.

```
$ bun i --verbose
bun install v0.5.8 (1a25af5e)
  🔍 Resolving [1/1]
error: InstallFailed cloning repository for @org/website-i1

error: "git clone" for "@org/website-i18n" failed


error: "git clone" for "@org/website-i18n" failed
```

### Additional information

- Passing in `--verbose` doesn't provide me with more information. Here, I'm interested in seeing the `git clone` command & error.
- I'm not certain this error is as a result of the repository being private. The org is an enterprise org, so it might be that the installation URL doesn't follow the typical GitHub git URL.

	The git clone command looks like this:
	```
	git clone org-12345678@github.com:org/website-i18n.git
	```

Jarred-Sumner:
github: syntax is only supported for public repositories in Bun. We use the GitHub tarball api for these (I think npm does as well) 

We should make the error message better though, and suggest to try with a git+ssh URL

privatenumber:
Since the error message says "git clone" failed I think it's happening here:
https://github.com/oven-sh/bun/blob/732c5e7fa9b0bb89938ca001749a13501386485e/src/install/repository.zig#L180

----

I was just about to file a bug for the GitHub tarball API url. It doesn't seem to work for private repositories either:

```
bun install v0.5.8 (1a25af5e)
  🔍 Resolving [1/1]
GET https://api.github.com/repos/org/analytics-js/tarball/v0.8.7 - 404
error: analytics-js@github:org/analytics-js#v0.8.7 failed to resolve
```

I think an API token will need to be passed in to access private repositories via GitHub API. It works with npm and I haven't configured a token so I think they're using the Git+SSH method.

PR #11917, opened 2024-06-17
fix(install): use ssh keys for private git repos

Eckhardt-D:
### What does this PR do?

Fixes #11993
Fixes #10641
Fixes #6927
Fixes #2464
Fixes #4068

Do not convert SCPLike and ssh:// urls to https URLs so that the system can use ssh keys when the resolution prefix is ssh:// - check if the host is in known hosts after git@. NB, not 100% sure if the skipping of verifyResolutions for .git tag is the way to go, might need some help there.

- [x] Code changes

I ran bun install `<private-repo-url>` on a few private repo packages I have access to on different platforms.

and also had a package.json test as follows:

```json
{
  "name": "test-git-install-bun",
  "module": "index.ts",
  "type": "module",
  "dependencies": {
    "gw-client": "ssh://git@gitlab.com/<org>/gateway-client.git",
    "gw-client-2": "git@gitlab.com:<org>/gateway-client.git",
    "gw-client-3": "git+ssh://git@gitlab.com/<org>/gateway-client.git"
  }
}
```

☝️ No, because it's dependent on ssh keys.

### Caveat

This is the first time I ever wrote Zig so it may be a good idea to double check my work 🤣 especially what I did in `src/install/lockfile.zig`, I could not for the life of me figure out why verifyResolutions is failing so I silenced it for git repos basically, don't think that's the way to go.

### DIFF
--- a/src/env_loader.zig
+++ b/src/env_loader.zig
@@ -1270,7 +1270,7 @@ pub const Map = struct {
     }
 
     pub fn remove(this: *Map, key: string) void {
-        this.map.remove(key);
+        _ = this.map.swapRemove(key);
     }
 
     pub fn cloneWithAllocator(this: *const Map, new_allocator: std.mem.Allocator) !Map {
--- a/src/install/dependency.zig
+++ b/src/install/dependency.zig
@@ -577,7 +577,9 @@ pub const Version = struct {
                                 }
                             }
 
+
                             if (strings.indexOfChar(url, '.')) |dot| {
+                                if (Repository.Hosts.has(url["git@".len..dot])) return .git;
                                 if (Repository.Hosts.has(url[0..dot])) return .git;
                             }
                         }
--- a/src/install/dependency.zig
+++ b/src/install/dependency.zig
@@ -578,8 +578,11 @@ pub const Version = struct {
                             }
 
 
+                            if (url.len > 4 and strings.eqlComptime(url[0.."git@".len], "git@")) {
+                                url = url["git@".len..];
+                            }
+
                             if (strings.indexOfChar(url, '.')) |dot| {
-                                if (Repository.Hosts.has(url["git@".len..dot])) return .git;
                                 if (Repository.Hosts.has(url[0..dot])) return .git;
                             }
                         }
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -1988,6 +1988,53 @@ pub fn verifyData(this: *const Lockfile) !void {
     }
 }
 
+pub fn verifyResolutions(this: *Lockfile, local_features: Features, remote_features: Features, comptime log_level: PackageManager.Options.LogLevel) void {
+    const resolutions_lists: []const DependencyIDSlice = this.packages.items(.resolutions);
+    const dependency_lists: []const DependencySlice = this.packages.items(.dependencies);
+    const dependencies_buffer = this.buffers.dependencies.items;
+    const resolutions_buffer = this.buffers.resolutions.items;
+    const end = @as(PackageID, @truncate(this.packages.len));
+
+    var any_failed = false;
+    const string_buf = this.buffers.string_bytes.items;
+
+    const root_list = resolutions_lists[0];
+    for (resolutions_lists, dependency_lists, 0..) |resolution_list, dependency_list, parent_id| {
+        for (resolution_list.get(resolutions_buffer), dependency_list.get(dependencies_buffer)) |package_id, failed_dep| {
+            if (package_id < end) continue;
+            if (failed_dep.behavior.isPeer() or !failed_dep.behavior.isEnabled(
+                if (root_list.contains(@truncate(parent_id)))
+                    local_features
+                else
+                    remote_features,
+            )) continue;
+
+            if (log_level != .silent) {
+                if (failed_dep.name.isEmpty() or strings.eqlLong(failed_dep.name.slice(string_buf), failed_dep.version.literal.slice(string_buf), true)) {
+                    Output.prettyErrorln(
+                        "<r><red>error<r><d>:<r> <b>{}<r><d> failed to resolve<r>\n",
+                        .{
+                            failed_dep.version.literal.fmt(string_buf),
+                        },
+                    );
+                } else {
+                    Output.prettyErrorln(
+                        "<r><red>error<r><d>:<r> <b>{s}<r><d>@<b>{}<r><d> failed to resolve<r>\n",
+                        .{
+                            failed_dep.name.slice(string_buf),
+                            failed_dep.version.literal.fmt(string_buf),
+                        },
+                    );
+                }
+            }
+            // track this so we can log each failure instead of just the first
+            any_failed = true;
+        }
+    }
+
+    if (any_failed) Global.crash();
+}
+
 pub fn saveToDisk(this: *Lockfile, filename: stringZ) void {
     if (comptime Environment.allow_assert) {
         this.verifyData() catch |err| {
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -1988,53 +1988,6 @@ pub fn verifyData(this: *const Lockfile) !void {
     }
 }
 
-pub fn verifyResolutions(this: *Lockfile, local_features: Features, remote_features: Features, comptime log_level: PackageManager.Options.LogLevel) void {
-    const resolutions_lists: []const DependencyIDSlice = this.packages.items(.resolutions);
-    const dependency_lists: []const DependencySlice = this.packages.items(.dependencies);
-    const dependencies_buffer = this.buffers.dependencies.items;
-    const resolutions_buffer = this.buffers.resolutions.items;
-    const end = @as(PackageID, @truncate(this.packages.len));
-
-    var any_failed = false;
-    const string_buf = this.buffers.string_bytes.items;
-
-    const root_list = resolutions_lists[0];
-    for (resolutions_lists, dependency_lists, 0..) |resolution_list, dependency_list, parent_id| {
-        for (resolution_list.get(resolutions_buffer), dependency_list.get(dependencies_buffer)) |package_id, failed_dep| {
-            if (package_id < end) continue;
-            if (failed_dep.behavior.isPeer() or !failed_dep.behavior.isEnabled(
-                if (root_list.contains(@truncate(parent_id)))
-                    local_features
-                else
-                    remote_features,
-            )) continue;
-
-            if (log_level != .silent) {
-                if (failed_dep.name.isEmpty() or strings.eqlLong(failed_dep.name.slice(string_buf), failed_dep.version.literal.slice(string_buf), true)) {
-                    Output.prettyErrorln(
-                        "<r><red>error<r><d>:<r> <b>{}<r><d> failed to resolve<r>\n",
-                        .{
-                            failed_dep.version.literal.fmt(string_buf),
-                        },
-                    );
-                } else {
-                    Output.prettyErrorln(
-                        "<r><red>error<r><d>:<r> <b>{s}<r><d>@<b>{}<r><d> failed to resolve<r>\n",
-                        .{
-                            failed_dep.name.slice(string_buf),
-                            failed_dep.version.literal.fmt(string_buf),
-                        },
-                    );
-                }
-            }
-            // track this so we can log each failure instead of just the first
-            any_failed = true;
-        }
-    }
-
-    if (any_failed) Global.crash();
-}
-
 pub fn saveToDisk(this: *Lockfile, filename: stringZ) void {
     if (comptime Environment.allow_assert) {
         this.verifyData() catch |err| {
--- a/src/install/repository.zig
+++ b/src/install/repository.zig
@@ -159,19 +159,29 @@ pub const Repository = extern struct {
         // A value can still be entered, but we need to find a workaround
         // so the user can see what is being prompted. By default the settings
         // below will cause no prompt and throw instead.
-        const originalAskPass = env.map.get("GIT_ASKPASS");
-        const originalSSHCommand = env.map.get("GIT_SSH_COMMAND");
+        const askpass_entry = env.map.getOrPutWithoutValue("GIT_ASKPASS") catch bun.outOfMemory();
+        if (!askpass_entry.found_existing) {
+            askpass_entry.key_ptr.* = allocator.dupe(u8, "GIT_ASKPASS") catch bun.outOfMemory();
+            askpass_entry.value_ptr.* = .{
+                .value = allocator.dupe(u8, "echo") catch bun.outOfMemory(),
+                .conditional = false,
+            };
+        }
 
-        try env.map.putAllocKeyAndValue(allocator, "GIT_ASKPASS", "echo");
-        try env.map.putAllocKeyAndValue(allocator, "GIT_SSH_COMMAND", "ssh -oStrictHostKeyChecking=accept-new");
+        const ssh_command_entry = env.map.getOrPutWithoutValue("GIT_SSH_COMMAND") catch bun.outOfMemory();
+        if (!ssh_command_entry.found_existing) {
+            ssh_command_entry.key_ptr.* = allocator.dupe(u8, "GIT_SSH_COMMAND") catch bun.outOfMemory();
+            ssh_command_entry.value_ptr.* = .{
+                .value = allocator.dupe(u8, "ssh -oStrictHostKeyChecking=accept-new") catch bun.outOfMemory(),
+                .conditional = false,
+            };
+        }
 
         var std_map = try env.map.stdEnvMap(allocator);
 
         defer {
-            _ = if (originalAskPass) |value| env.map.put(allocator, "GIT_ASKPASS", value) catch void else env.map.remove("GIT_ASKPASS");
-
-            _ = if (originalSSHCommand) |value| env.map.put(allocator, "GIT_SSH_COMMAND", value) catch void else env.map.remove("GIT_SSH_COMMAND");
-
+            if (!askpass_entry.found_existing) env.map.remove("GIT_ASKPASS");
+            if (!ssh_command_entry.found_existing) env.map.remove("GIT_SSH_COMMAND");
             std_map.deinit();
         }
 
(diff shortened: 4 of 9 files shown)


Eckhardt-D:
Another note @dylan-conway @Jarred-Sumner I also made sure that the private repo has another private repo dependency in it's package.json to confirm it works recursively - and that is that point where `verifyResolutions` fails, not on the initial install but on the package's deps then it prints the url like:

```
error: pkg@git@gitlab.com:org/repo.git failed to resolve
```
Which does not look right.

test/cli/install/migration/complex-workspace/package.json
```diff
@@ -5,7 +5,6 @@
     "bar": "https://github.com/oven-sh/bun/raw/f7e4eb83694aa007a492ef66c28ffbe6a2dae791/test/cli/install/bar-0.0.2.tgz",
     "bun-types": "file:bun-types",
     "hello": "file:hello-0.3.2.tgz",
-    "install-test": "bitbucket:dylan-conway/public-install-test",
```
Eckhardt-D: @dylan-conway Interesting finding while debugging the test failures. Bitbucket is the only one that still does an ssh key check even if the repo is public. Github and Gitlab both clone with e.g.

```
git clone ssh://git@gitlab.com/org/repo.git
```

if the repo is public, even if the system does not have a linked ssh key. Bitbucket does a key check regardless.

src/install/repository.zig
```diff
@@ -178,14 +178,16 @@ pub const Repository = extern struct {
...
+            and strings.hasSuffixComptime(url, ".git")) {
             return null;
         }
 
         if (Dependency.isSCPLikePath(url)) {
             // Use git@<host>:<repo> urls as-is so that the git clone can
             // use the system's ssh keys for private repos
-            if (strings.hasPrefixComptime(url, "git@")) {
+            if (strings.hasPrefixComptime(url, "git@")
+                and strings.hasSuffixComptime(url, ".git")) {
```
Eckhardt-D: Technically a .git extension is not necessary for the git clone, but it breaks some tests if `git@gitlab.com:some/repo` is specified without it

src/install/dependency.zig
```diff
@@ -562,7 +562,9 @@ pub const Version = struct {
                                 }
                             }
 
+
                             if (strings.indexOfChar(url, '.')) |dot| {
+                                if (Repository.Hosts.has(url["git@".len..dot])) return .git;
```
dylan-conway: The start index can be larger than the end index with input like `ssh://.`, which is not valid in zig. Instead I would remove this line and add this before the check for `'.'`:
```zig
if (url.len > 4 and strings.eqlComptime(url[0.."git@".len], "git@")) {
    url = url["git@".len..];
}
```

test/cli/install/migration/complex-workspace/package.json
dylan-conway: I see what you mean, not sure what's causing this yet

src/install/dependency.zig
Eckhardt-D: Done ✔️ Makes 100% sense and thank you

test/cli/install/migration/complex-workspace/package.json
Eckhardt-D: @dylan-conway Found it! The package-lock.json has the bitbucket git dep listed as:

```
git+ssh://git@bitbucket.org/dylan-conway/public-install-test.git#......
```

This causes git clone to fail because of the `.git`. That causes BitBucket to check for an SSH key for some reason (unlike others) and prompts for the password. With a fresh package, this does not happen, because the `.git` extension is ommitted in the lockfile version when installing those scp-like items. Is this just an unavoidable thing due to bitbucket?

For this to happen, a user would need:

1. An old npm lockfile
2. A bitbucket dependency listed in that lockfile the way npm prints it

Bun install can't work around it by .e.g. removing `.git` for bitbucket, because then it would skip using ssh keys on private repos.

The only thing I can think here is to force the https try to fail and not simply prompt for credentials. When https fails, fall back to ssh. If ssh fails, it's a complete failure.

How to stop git from prompting username & password:
```
GIT_TERMINAL_PROMPT=0 GIT_ASKPASS=echo SSH_ASKPASS=echo GCM_INTERACTIVE=never git clone ...
```

But this does not seem like the right approach imo and think the bitbucket thing should just be a 'caveat'

Eckhardt-D:
@dylan-conway This commit [be70b4d](https://github.com/oven-sh/bun/pull/11917/commits/be70b4de11fbef1e67dcd90e855891093e31d511) introduces quite a bit of changes for caveats I found in the initial implementation:

1. Users expect e.g. `gitlab:owner/repo` to work whether it's private or public. The previous implementation didn't address that.

Again, the code is probably not the best (first time Zigging) - but it does what I am trying to convey:

1. Tries cloning with HTTP URL
2. Tries SSH if HTTP fails

The biggest thing is that previously the `catch` block of the tryHTTPS clone would not through, just prompt. By setting the process env vars to make Git non-interactive it forces it to throw then SSH is tried.

One edge case this caused and that is addressed is that if an https URL is not a valid git repo, it would also not report the error - currently my fix is to read from `stderr` of the Git process when exit != 0. That seems flaky, but the stderr message has not changed in many years and can't think of another way to exit early on HTTP if the repo does not exist.

Eckhardt-D:
@Jarred-Sumner Would it be possible to run the action again please? I would like to test with `bun-pr` in a docker container with one of our organizations large monorepos with plenty of private + public deps.

Eckhardt-D:
Tests failing on just a single dep in complex-workspace 💪 I will look into it and fix today somewhere.

src/install/repository.zig
```diff
@@ -238,7 +238,7 @@ pub const Repository = extern struct {
 
         if (strings.hasPrefixComptime(url, "ssh://")) {
             final_path_buf[0.."https".len].* = "https".*;
-            bun.copy(u8, final_path_buf["https://".len..], url["ssh".len..]);
```
Eckhardt-D: 2.5 hours 🫠 @dylan-conway @Jarred-Sumner this should fix the complex workspace tests.

src/install/lockfile.zig
```diff
@@ -1997,10 +1997,6 @@ pub fn verifyResolutions(this: *Lockfile, local_features: Features, remote_featu
                     remote_features,
             )) continue;
 
-            // TODO: this is temporary to skip resolution check of Git ssh urls
-            // it currently fails to verify, not sure why yet
-            if (failed_dep.version.tag == .git) continue;
-
```
Eckhardt-D: This is no longer necessary.

test/cli/install/migration/complex-workspace/package.json
```diff
@@ -5,6 +5,7 @@
     "bar": "https://github.com/oven-sh/bun/raw/f7e4eb83694aa007a492ef66c28ffbe6a2dae791/test/cli/install/bar-0.0.2.tgz",
     "bun-types": "file:bun-types",
     "hello": "file:hello-0.3.2.tgz",
+    "install-test": "bitbucket:dylan-conway/public-install-test",
```
Eckhardt-D: Added back to ensure 0 regressions.

src/install/repository.zig
```diff
@@ -238,7 +238,7 @@ pub const Repository = extern struct {
...
+            bun.copy(u8, final_path_buf["https".len..], url["ssh".len..]);
```
Eckhardt-D: 2.5 hours 🫠 @dylan-conway @Jarred-Sumner This fixes the test failures in complex-workspace

src/install/repository.zig
```diff
@@ -152,6 +154,14 @@ pub const Repository = extern struct {
         argv: []const string,
     ) !string {
         var std_map = try env.map.stdEnvMap(allocator);
+        // Run git clone in non-interactive mode and
+        // error if ssh or https asks for a password
+        // so that we can fallback to ssh mode
+        try std_map.unsafe_map.put("GIT_TERMINAL_PROMPT", "0");
```
dylan-conway: I think we might want to copy the behavior here https://github.com/npm/cli/blob/22731831e22011e32fa0ca12178e242c2ee2b33d/node_modules/%40npmcli/git/lib/opts.js#L1-L6. Looks like we should use the existing values for `GIT_ASKPASS` and `GIT_SSH_COMMAND`, and if they don't exist use the default values `'echo'` and `'ssh -oStrictHostKeyChecking=accept-new'`

dylan-conway: Also these keys will need to be placed in `env` before calling `stdEnvMap()`. It might be easiest to use `putAllocKeyAndValue`. And the original values need to be restored, I would put that in the `defer { env.map.deinit }` block

dylan-conway:
> Again, the code is probably not the best (first time Zigging) - but it does what I am trying to convey:
> 
> 1. Tries cloning with HTTP URL
> 2. Tries SSH if HTTP fails

This seems like the correct approach, npm does the same thing https://github.com/npm/cli/blob/22731831e22011e32fa0ca12178e242c2ee2b33d/node_modules/pacote/lib/git.js#L82-L84

> One edge case this caused and that is addressed is that if an https URL is not a valid git repo, it would also not report the error - currently my fix is to read from stderr of the Git process when exit != 0. That seems flaky, but the stderr message has not changed in many years and can't think of another way to exit early on HTTP if the repo does not exist.

There might be something we can find in npm's codebase for how they choose to fallback to ssh

src/install/repository.zig
Eckhardt-D: Thanks! Good catch for the Host Checking, hadn't thought about that. In the case that the user's system has the existing values to ask for password, we would need to suspend the stdout output of the installer so that the user has a chance to enter the credentials. Otherwise Bun prints over the prompt, do you have advice on how to achieve that?

Eckhardt-D: @dylan-conway for the clone, in NPM for Windows they also use the `core.longpaths=true` cli option. Because on Windows, git is limited to 260 chars for paths to install to. Shall Bun install do the same? See [here](https://github.com/npm/cli/blob/22731831e22011e32fa0ca12178e242c2ee2b33d/node_modules/%40npmcli/git/lib/clone.js#L97-L99).

dylan-conway: Good idea, we should include it as an option in `checkout/clone` like `git clone -c core.longpaths=true <repo-url>` to avoid spawning another process.

> Otherwise Bun prints over the prompt, do you have advice on how to achieve that?

I'm not sure what we can do, for now I guess we hope the vscode prompt is chosen over terminal. Need to check how npm/yarn/pnpm handles this

src/env_loader.zig
```diff
@@ -1270,7 +1270,7 @@ pub const Map = struct {
     }
 
     pub fn remove(this: *Map, key: string) void {
-        this.map.remove(key);
+        _ = this.map.swapRemove(key);
```
Eckhardt-D, line 1273: This was a doozy, seems that nowhere `map.remove` was being called so this error never surfaced. Not sure why the static analyzer does not pick it up, but Zig's ArrayHashMap does not have a `remove` method, only a few others and I chose [swapRemove](https://ziglang.org/documentation/master/std/#std.array_hash_map.ArrayHashMap.swapRemove) - do you agree @dylan-conway

src/install/repository.zig
```diff
@@ -151,8 +153,30 @@ pub const Repository = extern struct {
...
+        defer {
+            _ = if (originalAskPass) |value| env.map.putAllocKeyAndValue(allocator, "GIT_ASKPASS", value) catch void
+            else env.map.remove("GIT_ASKPASS");
+
+            _ = if (originalSSHCommand) |value| env.map.putAllocKeyAndValue(allocator, "GIT_SSH_COMMAND", value) catch void
+            else env.map.remove("GIT_SSH_COMMAND");
+
+            // Remove the previous environment variables if they were set
+            std_map.deinit();
+        }
```
Eckhardt-D: @dylan-conway is this how you meant? :)

Eckhardt-D:
> There might be something we can find in npm's codebase for how they choose to fallback to ssh

NPM Also Looks for a specific message in the stderr of the response and converts it to a different error, however - they use `ls-remote` first to check the repo. [source1](https://github.com/npm/cli/blob/22731831e22011e32fa0ca12178e242c2ee2b33d/node_modules/%40npmcli/git/lib/revs.js#L21) [source2](https://github.com/npm/cli/blob/22731831e22011e32fa0ca12178e242c2ee2b33d/node_modules/%40npmcli/git/lib/make-error.js#L18)

This adds yet another network call to the process 🐌 given that ultimately there will be an error in the ssh call after the https try, if the stderr output ever changes, the only side effect would be that Bun unnecessarily tries ssh after https, which I think is better than ALWAYS running ls-remote?

Eckhardt-D:
> [🪟💻 4 failing tests](https://github.com/oven-sh/bun/actions/runs/9586962026) Windows x64 baseline

Looking at the crash report, it looks like the exception is coming from `SYSTEM32/ntdll.dll` which hopefully is because of `NtCreateFile` not working because of file path lengths 🙏 could be fixed by latest update.

src/env_loader.zig
paperclover: static analysis is lazy, it does not visit unused things. swapRemove should be fine here unless the ordering of the env keys matters (i dont think the order matters, so this is ok)

Eckhardt-D: Thank you @paperdave

src/install/repository.zig
```diff
@@ -151,8 +153,29 @@ pub const Repository = extern struct {
...
+        const originalSSHCommand = env.map.get("GIT_SSH_COMMAND");
+
+        try env.map.putAllocKeyAndValue(allocator, "GIT_ASKPASS", "echo");
+        try env.map.putAllocKeyAndValue(allocator, "GIT_SSH_COMMAND", "ssh -oStrictHostKeyChecking=accept-new");
+
         var std_map = try env.map.stdEnvMap(allocator);
-        defer std_map.deinit();
+
+        defer {
+            _ = if (originalAskPass) |value| env.map.putAllocKeyAndValue(allocator, "GIT_ASKPASS", value) catch void
```
Eckhardt-D: Oh right, this is wrong. It won't update the key, forgot that this 'gets or updates'

src/install/lockfile.zig
```diff
@@ -1988,6 +1988,53 @@ pub fn verifyData(this: *const Lockfile) !void {
     }
 }
 
+pub fn verifyResolutions(this: *Lockfile, local_features: Features, remote_features: Features, comptime log_level: PackageManager.Options.LogLevel) void {
```
dylan-conway: this function can be deleted, it was moved to `install.zig` recently

src/install/repository.zig
```diff
@@ -168,17 +191,69 @@ pub const Repository = extern struct {
             });
 
         switch (result.term) {
-            .Exited => |sig| if (sig == 0) return result.stdout,
+            .Exited => |sig| if (sig == 0) return result.stdout else if (
+                // remote: The page could not be found <-- for non git
+                // remote: Repository not found. <-- for git
```
dylan-conway: Another variation we can check for:
```
git clone invalidurl
fatal: repository 'invalidurl' does not exist
```

dylan-conway: although, I don't think that will make it through dependency parsing as a git dependency, so it's probably fine

src/install/repository.zig
Eckhardt-D: Updated. Am I right with my assumption here @dylan-conway that if the vars were replaced with putAllocKeyAndValue in the previous call - this won't revert it? I converted it to `.put` now.

dylan-conway: It will revert them, but `putAllocKeyAndValue` will also overwrite the existing vars on lines 165 and 166. We want to keep the existing ones, so instead `getOrPutWithoutValue` works better here. You could do something like:
```zig
const askpass_entry = env.map.getOrPutWithoutValue("GIT_ASKPASS") catch bun.outOfMemory();
if (!askpass_entry.found_existing) {
    askpass_entry.key_ptr.* = allocator.dupe(u8, "GIT_ASKPASS") catch bun.outOfMemory();
    askpass_entry.value_ptr.* = .{
        .value = allocator.dupe(u8, "echo") catch bun.outOfMemory(),
        .conditional = false,
    };
}
```
then in the `defer` block:
```zig
if (!askpass_entry.found_existing) {
    env.map.remove("GIT_ASKPASS");
}
```

test/cli/install/bun-install.test.ts
```diff
@@ -4519,6 +4519,42 @@ it("should fail on invalid Git URL", async () => {
...
+      },
+    }),
+  );
+  const { stdout, stderr, exited } = spawn({
+    cmd: [bunExe(), "install"],
+    cwd: package_dir,
+    stdout: "pipe",
+    stdin: "pipe",
+    stderr: "pipe",
+    env,
```
dylan-conway:
```suggestion
    stdin: "ignore",
    stderr: "pipe",
    env: { ...env, "GIT_ASKPASS": "echo" },
```
to prevent prompting for a password when running the test

dylan-conway:
After https://github.com/oven-sh/bun/pull/11917#discussion_r1648393512 and https://github.com/oven-sh/bun/pull/11917#discussion_r1648392453 are addressed this pr should be almost ready to go. There is one new CI failure in `complex-workspace.test.ts` on windows baseline, but I believe https://github.com/oven-sh/bun/pull/11917#discussion_r1648392453 will fix it

dylan-conway:
> There is one new CI failure in `complex-workspace.test.ts` on windows baseline, but I believe [#11917 (comment)](https://github.com/oven-sh/bun/pull/11917#discussion_r1648392453) will fix it

Actually it's not new, seeing it here as well https://github.com/oven-sh/bun/pull/12022#issuecomment-2181802862

dylan-conway (approved):
Looks good, thank you!

UmeshMohan-Dozee:
`bun add git+ssh://github.com/lodash/lodash.git#4.17.21` works if lodash is a public repository. If lodash is a private repo, it fails with the following messages:
```
bun add v1.1.24 (85a32991)
error: "git fetch" for "git+ssh://github.com/lodash/lodash.git#4.17.21" failed

error: "git fetch" for "git+ssh://github.com/lodash/lodash.git#4.17.21" failed

error: InstallFailed cloning repository for git+ssh://github.com/lodash/lodash.git#4.17.21
error: git+ssh://github.com/lodash/lodash.git#4.17.21 failed to resolve
```

PR #6531, opened 2023-10-16
`bun:sql` is a fast builtin SQL client, starting with PostgreSQL support

Jarred-Sumner:
### What does this PR do?

This adds a fast builtin postgres client to bun, under a new `bun:sql` module.

```zig

❯ bun bun-sql.ts
cpu: Apple M1 Max
runtime: bun 1.0.6 (arm64-darwin)

benchmark                                         time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------------------ -----------------------------
SELECT "Id" FROM "OrderDetail" LIMIT 10 x 10  137.02 µs/iter (101.38 µs … 736.17 µs) 137.58 µs 396.83 µs 566.54 µs
SELECT * FROM "OrderDetail" LIMIT 10 x 10     161.61 µs/iter    (128.04 µs … 1.5 ms) 162.46 µs 255.75 µs 598.13 µs

❯ node postgresjs-sql.mjs
cpu: Apple M1 Max
runtime: node v20.8.0 (arm64-darwin)

benchmark                                                         time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------------------------------------------------- -----------------------------
SELECT "Id" FROM "OrderDetail" LIMIT 10 x 10 - "postgres.js"  235.55 µs/iter      (148.17 µs … 1 ms) 239.67 µs 585.38 µs 706.42 µs
SELECT "Id" FROM "OrderDetail" LIMIT 10 x 10 - "pg"           427.25 µs/iter   (349.88 µs … 3.33 ms) 440.92 µs 632.79 µs   1.32 ms
SELECT * FROM "OrderDetail" LIMIT 10 x 10 - "postgres.js"     261.75 µs/iter     (205.5 µs … 1.7 ms) 270.71 µs 338.46 µs 480.88 µs
SELECT * FROM "OrderDetail" LIMIT 10 x 10 - "pg"              511.38 µs/iter   (435.42 µs … 2.69 ms) 510.75 µs   1.02 ms   1.77 ms
```

<details>

<summary>bun-sql.ts</summary>

```ts
import { sql, Sql } from "bun:sql";
import { bench, run } from "mitata";
process.env.DATABASE_URL = "postgres://jarred@localhost:5432/northwind";

bench(`SELECT "Id" FROM "OrderDetail" LIMIT 10 x 10`, async () => {
  const promises = new Array(10);
  for (let i = 0; i < 10; i++) {
[...]

PR #11920, opened 2024-06-17
Add unusable postgres client behind a canary-or-debug-only feature flag

Jarred-Sumner:
### What does this PR do?

Bring back #6531 

This isn't ready yet, but I don't want it to bitrot - so what if we merge this behind a canary/debug-only feature flag? 

The main issues are:
1. GC isn't being handled correctly when creating results, which leads to a crash after enough of the tests run. This makes it unusable for more than a brief period
2. TLS is not at all implemented yet
3. Only prepared statements are supported
4. No helper for transactions 
5. Only `SCRAM-SHA-256` password auth is implemented. MD5 digest, GSSPI, TLS, etc are not implemented yet.
6. Error messages from Postgres are almost unreadable
7. Various polish details are not implemented. `sql.array`, `command` property, number of changes, no way to transform values etc. I think we could get away with shipping a v1 that lacks most polish details though


src/bun.js/bindings/structure.cpp
```diff
@@ -0,0 +1,301 @@
...
+static JSC::JSValue toJS(JSC::Structure* structure, DataCell* cells, unsigned count, JSC::JSGlobalObject* globalObject)
+{
+    auto& vm = globalObject->vm();
+    auto* object = JSC::constructEmptyObject(vm, structure);
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    for (unsigned i = 0; i < count; i++) {
+        auto& cell = cells[i];
+        JSValue value = toJS(vm, globalObject, cell);
+        RETURN_IF_EXCEPTION(scope, value);
```
cirospaciari: if we throw we return the cell value, should this be empty, undefined or the object?

src/bun.js/bindings/structure.cpp
```diff
@@ -0,0 +1,301 @@
...
+    if (array) {
+        array->push(globalObject, toJS(structure, cells, count, globalObject));
+        return array;
+    }
+
+    auto* newArray = JSC::constructEmptyArray(globalObject, nullptr);
+    if (!newArray)
+        return {};
+
+    newArray->putDirectIndex(globalObject, 0, toJS(structure, cells, count, globalObject));
```
cirospaciari: should we check exceptions here?

src/sql/postgres.zig
```diff
@@ -0,0 +1,3960 @@
...
+            for (this.messages.items) |msg| {
+                var str = switch (msg) {
+                    inline else => |m| m.toUTF8(bun.default_allocator),
+                };
+                defer str.deinit();
+                _ = b.append(str.slice());
+                _ = b.append("\n");
+            }
+
+            return globalObject.createSyntaxErrorInstance("Postgres error occurred\n{s}", .{b.allocatedSlice()[0..b.len]});
```
cirospaciari, line 795: Its ok to just return the error instance instead of throwing?

src/sql/postgres.zig
```diff
@@ -0,0 +1,3960 @@
...
+    }
+
+    pub fn finalize(this: *@This()) callconv(.C) void {
+        debug("PostgresSQLQuery finalize", .{});
+        this.thisValue = .zero;
+        this.deref();
+    }
+
+    pub fn deref(this: *@This()) void {
+        const ref_count = this.ref_count.fetchSub(1, .Monotonic);
```
cirospaciari:
```suggestion
        const ref_count = this.ref_count.fetchSub(1, .Monotonic);
        bun.assert(ref_count > 0);
```

Jarred-Sumner: this will compile away to nothing because ref_count is an unsigned integer

src/bun.js/bindings/structure.cpp
Jarred-Sumner: toJS is necessary, but not in the putDirectIndex part

src/sql/postgres.zig
Jarred-Sumner: in this case, we throw a SyntaxError which becomes a promise rejection later

Issue #11115, opened 2024-05-16
node:http => IncomingMessage: complete=false forever


44kia244:
### What version of Bun is running?

1.1.8+89d25807f

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

1. Create a bun project with `index.js` as code below
2. Run it using bun
3. Using postman, create a POST request with non-empty JSON body to this server
4. Try 2-3 again, but using Node.js

```js
const http = require('http');

// Create a server instance
const server = http.createServer(async (req, res) => {
  process.stdout.write('Waiting for request to complete..');
  while (!req.complete) {
    process.stdout.write('.');
    await new Promise(resolve => setTimeout(resolve, 1000));
  }
  process.stdout.write('\n');

  // Log the request method and URL
  console.log(`Received ${req.method} request for ${req.url}`);

  // Set response header and status code
  res.writeHead(200, {'Content-Type': 'text/plain'});

  // Write response body
  res.end(JSON.stringify({complete: req.complete}, null, 2));
});

// Define the port the server will listen on
const PORT = 9745;

// Start the server and listen for incoming requests
server.listen(PORT, () => {
  console.log(`Server running at http://localhost:${PORT}/`);
});
```

### What is the expected behavior?

There should be "Waiting for request to complete..." for a while, then respond normally. The same on both Bun and Node.js

### What do you see instead?

(On bun) The message "Waiting for request to complete..." is appended with "." forever (`req.complete` is still false)

### Additional information

- I've tested this on Windows, macOS, and Linux. All systems works the same way.
- This bug cause express.js issue from #8926
- After a little bit of digging into code, I think [the request body stream data provider](https://github.com/oven-sh/bun/blob/d55ac0fa434d31503147faefef5340919735c2ad/src/js/node/http.ts#L770) may be the issue comparing to the [no-body case from Line 782](https://github.com/oven-sh/bun/blob/d55ac0fa434d31503147faefef5340919735c2ad/src/js/node/http.ts#L782) and [740](https://github.com/oven-sh/bun/blob/d55ac0fa434d31503147faefef5340919735c2ad/src/js/node/http.ts#L740).
- From previous finding, I tried setting `complete` to `true` after line 770 and 786. It makes express.js work correctly (resolves #8926), but leaving my PoC above unresolved.
[...]

Issue #8926, opened 2024-02-15
Infinite loading on non existing paths on a express server.


rayanfer32:
### What version of Bun is running?

1.0.26

### What platform is your computer?

Linux 5.15.133.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

```js
import express from 'express';

const app = express();
app.use(express.json());

app.get('/api/hotels', (req, res) => {
  res.json({
    success: true
  })
})

app.listen(3000, () => console.log('App is running on port 3000!'));
```

```bash
curl  -X POST \
  'localhost:3000/api/hotel' \
  --header 'Accept: */*' \
  --header 'User-Agent: Thunder Client (https://www.thunderclient.com)' \
  --header 'Content-Type: application/json' \
  --data-raw '{
  "name": "Bellisima",
  "address": "North Hampton, near lake bloomdale"
}'
```

Bun with Express Server doesn't handle routes / paths very well; if you have a typo in the route, usually a missing 's' for plural API endpoints. In my case, there is no endpoint defined for `api/hotel` and I'm seeing an infinite loader because of that. If I do not pass anything in the body, then i get 

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Error</title>
  </head>
  <body>
    <pre>Cannot POST /api/hotel</pre>
  </body>
</html>
``` 
This is correct. But it should be the same when the body is provided too right.

![image](https://github.com/oven-sh/bun/assets/37145078/aed137cd-0392-462b-8f94-e3b27b7b0069)
 

### What is the expected behavior?

Should return path/route does not  exist

### What do you see instead?

Infinite loading

### Additional information

Node js works fine, only facing the issue with bun.

44kia244:
Same issue here. It seems like something in express.js logic. The `express.json()` middleware already call `next()` but express does not handle the not-found case.

44kia244:
After some investigation, it seems to be the `node:http` that bugged. I have opened issue #11115 for it.

PR #11926, opened 2024-06-17
fix(http) mark completed true

cirospaciari:
### What does this PR do?

Fix: https://github.com/oven-sh/bun/issues/8926
Fix: https://github.com/oven-sh/bun/issues/11115

- [X] Code changes

Test added

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -766,6 +766,7 @@ async function consumeStream(self, reader: ReadableStreamDefaultReader) {
     if (self[abortedSymbol]) return;
     if (done) {
       self.push(null);
+      self.complete = true;
       break;
     }
     for (var v of value) {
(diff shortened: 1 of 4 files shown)


src/js/node/http.ts
```diff
@@ -766,6 +766,7 @@ async function consumeStream(self, reader: ReadableStreamDefaultReader) {
     if (self[abortedSymbol]) return;
     if (done) {
       self.push(null);
+      self.complete = true;
```
Jarred-Sumner, line 768: This needs to be before `push(null)` to be consistent with Node.js:

https://github.com/nodejs/node/blob/78a326e2eb236874c45a7937d51e5d3ab3c91a3e/lib/_http_common.js#L141-L153

Jarred-Sumner (changes requested):
1 comment

src/js/node/http.ts
cirospaciari: fixed in https://github.com/oven-sh/bun/pull/11926/commits/5205985edadd0b7ed62f2feb17b54accda755916

Jarred-Sumner (approved):

PR #11927, opened 2024-06-17
fix(windows): dont call SetConsoleMode on stdin

paperclover:
### What does this PR do?

Running `bun` with no arguments on Windows would leave the terminal in an invalid state, which powershell would not expect and exit.

image

I am unaware of any place we actually depend on setting `ENABLE_VIRTUAL_TERMINAL_PROCESSING`. Node.js only does this within libuv's tty stuff, so we should not need to touch it at all.

From MS Docs:

> Setting this flag directs the Virtual Terminal processing engine to convert user input received by the console window into [Console Virtual Terminal Sequences](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) that can be retrieved by a supporting application through [ReadFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile) or [ReadConsole](https://learn.microsoft.com/en-us/windows/console/readconsole) functions.
>
> The typical usage of this flag is intended in conjunction with ENABLE_VIRTUAL_TERMINAL_PROCESSING on the output handle to connect to an application that communicates exclusively via virtual terminal sequences.

We do not use it for this or anything else.

Manually testing a set of prompt libraries, `bun init`, etc

### DIFF
--- a/src/output.zig
+++ b/src/output.zig
@@ -138,6 +138,9 @@ pub const Source = struct {
     const WindowsStdio = struct {
         const w = bun.windows;
 
+        /// At program start, we snapshot the console modes of standard in, out, and err
+        /// so that we can restore them at program exit if they change. Restoration is
+        /// best-effort, and may not be applied if the process is killed abruptly.
         pub var console_mode = [3]?u32{ null, null, null };
         pub var console_codepage = @as(u32, 0);
         pub var console_output_codepage = @as(u32, 0);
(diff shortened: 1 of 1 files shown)


src/output.zig
```diff
@@ -196,7 +196,6 @@ pub const Source = struct {
             if (w.kernel32.GetConsoleMode(stdin, &mode) != 0) {
                 console_mode[0] = mode;
                 bun_stdio_tty[0] = 1;
-                _ = w.SetConsoleMode(stdin, mode | w.ENABLE_VIRTUAL_TERMINAL_INPUT);
```
Jarred-Sumner, line 199: Can you add a comment explaining why we aren't setting it here?

PR #11948, opened 2024-06-18
Experiment: AsyncPromise, a Promise that captures 1 stack frame

Jarred-Sumner:
### What does this PR do?

Stack traces for async functions are often missing

This adds an internal Promise wrapper for async native functions which captures 1 stack frame and propagates it on promise rejection.

After:
```ts
 9 | }
10 |
11 | async function hey() {
12 |   await 42;
13 |
14 |   readFile("fd-3.mjs").catch((e) => {
       ^
   errno: -2
 syscall: "open"
   path: "fd-3.mjs"

      at fd-2.mjs:14:3
```

Before:

```ts
   errno: -2
 syscall: "open"
   path: "fd-3.mjs"
```

Worth reading [what v8 does](https://docs.google.com/document/d/13Sy_kBIJGP0XT34V1CV3nkWya4TwYx9L3Yv45LdGB6Q/edit)

I don't think this PR's approach is quite right. This is only 1 stack frame, and only when using this special AsyncPromise in Zig/C++. It doesn't chain.

I think it'd be better to reuse the extra `context` argument in the await-specific code to have a reference to the StackFrame (though ideally without creating an additional JSCell)

cc @dcrousso incase you find this interesting

I did a little benchmarking and couldn't find a measurable performance impact to this


dcrousso:
support for capturing and linking stack traces across async boundaries was added to Web Inspector in [252543@main](https://commits.webkit.org/252543@main) by instrumenting `queueMicrotask` (and `JSMicrotask::run`).  perhaps yall could use/duplicate that logic?

PR #11950, opened 2024-06-18
Bun.$ should keep the Bun process alive while a command is running

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/12679

The following two scripts exit with exit code 1 but they should exit with exit code 0.

```js
process.exitCode = 1;

(async () => {
  await Bun.$`ls .`;
  process.exit(0);
})();
```

```js
process.exitCode = 1;

(async () => {
  await Bun.$`${process.execPath} -e "console.log('hi')"`;
  process.exit(0);
})();
```

There is a test


zackradisic (dismissed):
lgtm

Issue #11949, opened 2024-06-18
napi_call_threadsafe_function does nothing if nullptr is passed as data (Node-API)


eilvelia:
### What version of Bun is running?

1.1.13

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

### What steps can reproduce the bug?

The issue: if `napi_call_threadsafe_function` is called with `nullptr` as `data`, the `napi_threadsafe_function_call_js` function won't be called at all, which afaik is wrong and is different from the Node.js' behavior.

Create `addon.c`:

```c
#include <assert.h>
#include <stdio.h>
#include <node_api.h>

static void callback(napi_env env, napi_value js_callback, void *context, void *data) {
  printf("hi!! callback called!! data: %p\n", data);
}

static napi_value Example(napi_env env, napi_callback_info info) {
  napi_status status;
  napi_threadsafe_function tsfn;
  napi_value tsfn_name;
  status = napi_create_string_utf8(env, "TSFN", 4, &tsfn_name);
  assert(status == napi_ok);
  status = napi_create_threadsafe_function(env, NULL, NULL, tsfn_name, 0, 1, NULL, NULL, NULL, &callback, &tsfn);
  assert(status == napi_ok);
  status = napi_call_threadsafe_function(tsfn, NULL, napi_tsfn_nonblocking);
  assert(status == napi_ok);
  napi_value result;
  status = napi_get_undefined(env, &result);
  assert(status == napi_ok);
  return result;
}

#define DECLARE_NAPI_METHOD(name, func) { name, 0, func, 0, 0, 0, napi_default, 0 }

napi_value Init(napi_env env, napi_value exports) {
  napi_property_descriptor desc = DECLARE_NAPI_METHOD("example", Example);
  napi_status status = napi_define_properties(env, exports, 1, &desc);
  assert(status == napi_ok);
  return exports;
}

NAPI_MODULE(NODE_GYP_MODULE_NAME, Init)
```

Create `binding.gyp`:

```python
{
  "targets": [
    {
      "target_name": "addon",
      "sources": [ "addon.c" ]
    }
  ]
}
```

Create `test.js`:

```javascript
const addon = require('./build/Release/addon.node')
addon.example()
```

Run: `npm install -D node-gyp && npx node-gyp rebuild`.

[...]

PR #11952, opened 2024-06-18
fix(napi) `napi_call_threadsafe_function` should work with null data and  `napi_create_threadsafe_function` should keep the process alive by default

cirospaciari:
### What does this PR do?

Fix: #11949

- [X] Code changes

Added a test

### DIFF
--- a/src/sync.zig
+++ b/src/sync.zig
@@ -496,15 +496,17 @@ pub fn Channel(
             var popped: usize = 0;
             while (popped < items.len) {
                 const new_item = blk: {
-                    if (self.buffer.readItem()) |item| {
-                        self.putters.signal();
-                        break :blk item;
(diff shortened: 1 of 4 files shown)


src/sync.zig
```diff
@@ -496,15 +496,17 @@ pub fn Channel(
             var popped: usize = 0;
             while (popped < items.len) {
                 const new_item = blk: {
-                    if (self.buffer.readItem()) |item| {
-                        self.putters.signal();
-                        break :blk item;
+                    // Buffer can contain null items but readItem will return null if the buffer is empty.
+                    // we need to check if the buffer is empty before trying to read an item.
+                    if (self.buffer.count == 0) {
```
Jarred-Sumner, line 501: oh we should not use this, we should use the UnboundedQueue thing

Jarred-Sumner: nvm

Issue #424, opened 2022-07-08
Module._nodeModulePaths readonly property


Burak-Esen:
Hi, I try to run project includes module-alias npm package. I got that error
``` ~/.bun/bin/bun run index.js 
12 | var modulePaths = []
13 | var moduleAliases = {}
14 | var moduleAliasNames = []
15 | 
16 | var oldNodeModulePaths = Module._nodeModulePaths
17 | Module._nodeModulePaths = function (from) {
   ^
 TypeError: Attempted to assign to readonly property.
      at /project/node_modules/module-alias/index.js:17:0
      at bun:wrap:1:16354
      at /project/node_modules/module-alias/register.js:1:8
      at /project/index.js:1:8```

Jarred-Sumner:
If you want to alias modules, you can use [tsconfig.json's "paths"](https://www.typescriptlang.org/tsconfig#paths) syntax (if not using tsconfig.json, you can make the filename "jsconfig.json")

Burak-Esen:
I tried with jsconfig and without using module-alias package. It works !

PR #11979, opened 2024-06-19
Support reading from .npmrc

### LINKED ISSUE #643: Support reading from .npmrc

npm is configured by a `.npmrc`, but currently, bun seems to require the configuration to be duplicated to a `bunfig.toml` file.

For `bun install` to support reading from the `.npmrc` file.

_No response_

zackradisic:
### What does this PR do?
WIP

This PR implements a .npmrc parser and loads it with Bun's package manager

Fixes #643


src/ini.zig
```diff
@@ -0,0 +1,569 @@
...
+
+        // starts with anything other than whitespace then don't skip
+        if (switch (line[0]) {
+            ' ',
+            '\t',
+            '\n',
+            '\r',
+            => false,
+            else => true,
+        }) return false;
```
sirenkovladd: Why can't we just delete these lines, the loop below does the same thing?

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+            break :brk count;
+        };
+
+        if (count == 0) break :out;
+
+        const default_registry_url: bun.URL = brk: {
+            if (install.default_registry) |dr|
+                break :brk bun.URL.parse(dr.url);
+
+            break :brk bun.URL.parse("https://registry.npmjs.org/");
```
Jarred-Sumner:
```suggestion
            break :brk bun.URL.parse(Registry.default_url);
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                //
+                // - localhost:4873/
+                // - somewhere-else.com/myorg/
+                //
+                // Scoped registries are set like this:
+                // - @myorg:registry=https://somewhere-else.com/myorg
+                const conf_item: bun.ini.ConfigIterator.Item = conf_item_;
+                switch (conf_item.optname) {
+                    ._auth, .email, .certfile, .keyfile => {
+                        Output.warn(
```
Jarred-Sumner:
```suggestion
                        log.addWarningFmt(
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                const conf_item_url = bun.URL.parse(conf_item.registry_url);
+
+                if (std.mem.eql(u8, bun.strings.withoutTrailingSlash(default_registry_url.host), bun.strings.withoutTrailingSlash(conf_item_url.host))) {
+                    const v: *bun.Schema.Api.NpmRegistry = brk: {
+                        if (install.default_registry) |*r| break :brk r;
+                        install.default_registry = bun.Schema.Api.NpmRegistry{
+                            .password = "",
+                            .token = "",
+                            .username = "",
+                            .url = "https://registry.npmjs.org/",
```
Jarred-Sumner:
```suggestion
                            .url =  Registry.default_url,
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                            ._password => v.password = allocator.dupe(u8, conf_item.value) catch bun.outOfMemory(),
+                            ._auth, .email, .certfile, .keyfile => unreachable,
+                        }
+                        // We have to keep going as it could match multiple scopes
+                        continue;
+                    }
+                }
+
+                if (!matched_at_least_one) {
+                    Output.warn(
```
Jarred-Sumner: Use the logger so the output is consistent

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+    }
+
+    if (!bun.Environment.isDebug) {
+        if (!@import("./bun.js/module_loader.zig").ModuleLoader.is_allowed_to_use_internal_testing_apis)
+            return;
+    }
+
+    if (bun.getenvTruthy("BUN_TEST_LOG_DEFAULT_REGISTRY")) {
+        if (install.default_registry) |reg| {
+            Output.print("Default registry url: {s}\n", .{reg.url});
```
Jarred-Sumner: Per discussion IRL, let's add a test-only wrapper function that can be called from JS

Then if there are any log messages, it calls `.toJS` on it and returns or throws that.

Separately, lets enable `bun install --verbose` and check the output to verify that the expected URL is requested

test/cli/install/bun-patch.test.ts
```diff
@@ -90,10 +90,21 @@ describe("bun patch <pkg>", async () => {
...
+        [
+          "@types/ws@8.5.4",
+          "node_modules/@repo/eslint-config/node_modules/@types/ws",
+          "@types/ws@8.5.4",
+          "patches/@types%2Fws@8.5.4.patch",
+        ],
+        ["@types/ws@7.4.7", "node_modules/@types/ws", "@types/ws@7.4.7", "patches/@types%2Fws@7.4.7.patch"],
       ];
-      for (const [arg, path] of args) {
+      for (const [arg, path, ver, patch_path] of args) {
```
Jarred-Sumner: The reader might not know what `ver` means, the shorthand isn't much shorter

```suggestion
      for (const [arg, path, version, patch_path] of args) {
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                ' ', '\t', '\n', '\r' => {},
+                '#', ';' => return true,
+                else => return false,
+            }
+        }
+        return true;
+    }
+
+    fn parseImpl(this: *Parser) !void {
+        var iter = std.mem.splitScalar(u8, this.src, '\n');
```
Jarred-Sumner, line 64: what about `\r\n`? do we need to handle the more interesting non-ASCII newline characters?

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+        var skip_until_next_section: bool = false;
+
+        while (iter.next()) |line| {
+            if (shouldSkipLine(line)) continue;
+
+            // Section
+            // [foo]
+            if (line[0] == '[') {
+                skip_until_next_section = false;
+                const close_bracket_idx = std.mem.indexOfScalar(u8, line[0..], ']') orelse continue;
```
Jarred-Sumner, line 83: Is an escape character possible here? Can you do `[foo\]]`?

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+    }
+
+    fn parseImpl(this: *Parser) !void {
+        var iter = std.mem.splitScalar(u8, this.src, '\n');
+        var head: *E.Object = this.out.data.e_object;
+
+        // var duplicates = std.StringArrayHashMapUnmanaged(u32){};
+        // defer duplicates.deinit(allocator);
+
+        var stack = std.heap.stackFallback(@sizeOf(Rope) * 6, this.arena.allocator());
```
Jarred-Sumner: Can we make it more clear this is exclusively for use by section names?
```suggestion
        var rope_only_stack_allocator = std.heap.stackFallback(@sizeOf(Rope) * 6, this.arena.allocator());
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+            // [foo]
+            if (line[0] == '[') {
+                skip_until_next_section = false;
+                const close_bracket_idx = std.mem.indexOfScalar(u8, line[0..], ']') orelse continue;
+                // Make sure the rest is just whitespace
+                if (close_bracket_idx + 1 < line.len) {
+                    for (line[close_bracket_idx + 1 ..]) |c| if (switch (c) {
+                        ' ', '\t' => false,
+                        else => true,
+                    }) continue;
```
Jarred-Sumner: Is the `continue` pointing to the `for` loop or to the outer `while` loop? Can you add a label here to make that very clear to the reader?

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                }
+                const section: *Rope = try this.prepareStr(ropealloc, line[1..close_bracket_idx], @as(i32, @intCast(@intFromPtr(line.ptr) - @intFromPtr(this.src.ptr))) + 1, .section);
+                defer stack.fixed_buffer_allocator.reset();
+                const parent_object = this.out.data.e_object.getOrPutObject(section, this.arena.allocator()) catch |e| switch (e) {
```
Jarred-Sumner:
```suggestion
                const parent_object = this.out.data.e_object.getOrPutObject(section, allocator) catch |e| switch (e) {
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+            const is_array: bool = brk: {
+                if (this.opts.bracked_array) {
+                    break :brk key_raw.len > 2 and bun.strings.endsWith(key_raw, "[]");
+                } else {
+                    // const gop = try duplicates.getOrPut(allocator, key_raw);
+                    // if (gop.found_existing) {
+                    //     gop.value_ptr.* = 1;
+                    // } else gop.value_ptr.* += 1;
+                    // break :brk gop.value_ptr.* > 1;
+                    @panic("We don't support this right now");
```
Jarred-Sumner: imagine you're using bun

bun crashes with this message

how will you know what `"this"` is referring to?

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                key_raw;
+
+            if (bun.strings.eql(key, "__proto__")) continue;
+
+            const value_raw: Expr = brk: {
+                if (maybe_eq_sign_idx) |eq_sign_idx| {
+                    if (eq_sign_idx + 1 < line.len) break :brk try this.prepareStr(
+                        ropealloc,
+                        line[eq_sign_idx + 1 ..],
+                        @intCast(line_offset + @as(i32, @intCast(eq_sign_idx)) + 1),
```
Jarred-Sumner, line 172: I don't feel good about all these intcasts and adding up numbers without using slices

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                else
+                    value_raw,
+                else => value_raw,
+            };
+
+            if (is_array) {
+                if (head.get(key)) |val| {
+                    if (val.data != .e_array) {
+                        var arr = E.Array{};
+                        arr.push(this.arena.allocator(), val) catch bun.outOfMemory();
```
Jarred-Sumner:
```suggestion
                        arr.push(allocator, val) catch bun.outOfMemory();
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                        head.put(this.arena.allocator(), key, Expr.init(E.Array, arr, Loc.Empty)) catch bun.outOfMemory();
```
Jarred-Sumner:
```suggestion
                        head.put(allocator, key, Expr.init(E.Array, arr, Loc.Empty)) catch bun.outOfMemory();
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                    }
+                } else {
+                    head.put(this.arena.allocator(), key, Expr.init(E.Array, E.Array{}, Loc.Empty)) catch bun.outOfMemory();
```
Jarred-Sumner:
```suggestion
                    head.put(allocator, key, Expr.init(E.Array, E.Array{}, Loc.Empty)) catch bun.outOfMemory();
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                }
+            }
+
+            // safeguard against resetting a previously defined
+            // array by accidentally forgetting the brackets
+            var was_already_array = false;
+            if (head.get(key)) |val| {
+                if (val.data == .e_array) {
+                    was_already_array = true;
+                    val.data.e_array.push(this.arena.allocator(), value) catch bun.outOfMemory();
```
Jarred-Sumner:
```suggestion
                    val.data.e_array.push(allocator, value) catch bun.outOfMemory();
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                    head.put(this.arena.allocator(), key, val) catch bun.outOfMemory();
```
Jarred-Sumner:
```suggestion
                    head.put(allocator, key, val) catch bun.outOfMemory();
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                }
+            }
+            if (!was_already_array) {
+                head.put(this.arena.allocator(), key, value) catch bun.outOfMemory();
```
Jarred-Sumner:
```suggestion
                head.put(allocator, key, value) catch bun.outOfMemory();
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+            }
+            const src = bun.logger.Source.initPathString("<unknown>", val);
+            var log = bun.logger.Log.init(this.arena.allocator());
+            defer log.deinit();
+            // Try to parse it and it if fails will just treat it as a string
+            const json_val: Expr = bun.JSON.ParseJSONUTF8Impl(&src, &log, this.arena.allocator(), true) catch {
+                break :out;
+            };
+
+            if (json_val.asString(this.arena.allocator())) |str| {
```
Jarred-Sumner:
```suggestion
            if (json_val.asString(allocator)) |str| {
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+        var val = std.mem.trim(u8, val_, " \n\r\t");
+
+        if (isQuoted(val)) out: {
+            // remove single quotes before calling JSON.parse
+            if (val.len > 0 and val[0] == '\'') {
+                val = if (val.len > 1) val[1 .. val.len - 1] else val[1..];
+                offset += 1;
+            }
+            const src = bun.logger.Source.initPathString("<unknown>", val);
+            var log = bun.logger.Log.init(this.arena.allocator());
```
Jarred-Sumner:
```suggestion
            var log = bun.logger.Log.init(allocator);
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                        3 => brk: {
+                            defer i += 2;
+                            break :brk &[_]u8{ c, val[i + 1], val[i + 2] };
+                        },
+                        4 => brk: {
+                            defer i += 3;
+                            break :brk &[_]u8{ c, val[i + 1], val[i + 2], val[i + 3] };
+                        },
+                        // this means invalid utf8
+                        else => unreachable,
```
Jarred-Sumner, line 373: Invalid UTF-8 is not unreachable, this should error

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+                                3 => brk: {
+                                    defer i += 2;
+                                    break :brk &[_]u8{ '\\', c, val[i + 1], val[i + 2] };
+                                },
+                                4 => brk: {
+                                    defer i += 3;
+                                    break :brk &[_]u8{ '\\', c, val[i + 1], val[i + 2], val[i + 3] };
+                                },
+                                // this means invalid utf8
+                                else => unreachable,
```
Jarred-Sumner, line 322: Invalid UTF is not unreachable, this should error

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+        var offset = offset_;
+        var val = std.mem.trim(u8, val_, " \n\r\t");
+
+        if (isQuoted(val)) out: {
+            // remove single quotes before calling JSON.parse
+            if (val.len > 0 and val[0] == '\'') {
+                val = if (val.len > 1) val[1 .. val.len - 1] else val[1..];
+                offset += 1;
+            }
+            const src = bun.logger.Source.initPathString("<unknown>", val);
```
Jarred-Sumner: we could keep the original filename if there's invalid JSON embedded within right? imagine if you're seeing an error here and the error is in `<unknown>`

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+
+    const Options = struct {
+        bracked_array: bool = true,
+    };
+
+    pub fn init(allocator: Allocator, path: []const u8, src: []const u8, env: *bun.DotEnv.Loader) Parser {
+        return .{
+            .logger = bun.logger.Log.init(allocator),
+            .src = src,
+            .out = Expr.init(E.Object, E.Object{}, Loc.Empty),
```
Jarred-Sumner, line 30: This allocates btw.

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+        if (comptime usage == .value) return Expr.init(E.String, E.String.init(val[0..]), Loc{ .start = offset });
+        if (comptime usage == .key) return val[0..];
+        return strToRope(ropealloc, val[0..]);
+    }
+
+    /// Returns index to skip or null if not an env substitution
+    /// Invariants:
+    /// - `i` must be an index into `val` that points to a '$' char
+    fn parseEnvSubstitution(this: *Parser, val: []const u8, start: usize, i: usize, unesc: *std.ArrayList(u8)) ?usize {
+        if (i + 1 < val.len and val[i + 1] == '{') {
```
Jarred-Sumner: We know that if it's not `"${}".length` long then it cannot be an env substition, so we can check that once instead of checking the length an extra time.
```suggestion
        if (i + "{}".len >= val.len or val[0] != '{')
          return null;
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+        }
+        if (comptime usage == .value) return Expr.init(E.String, E.String.init(val[0..]), Loc{ .start = offset });
+        if (comptime usage == .key) return val[0..];
+        return strToRope(ropealloc, val[0..]);
+    }
+
+    /// Returns index to skip or null if not an env substitution
+    /// Invariants:
+    /// - `i` must be an index into `val` that points to a '$' char
+    fn parseEnvSubstitution(this: *Parser, val: []const u8, start: usize, i: usize, unesc: *std.ArrayList(u8)) ?usize {
```
Jarred-Sumner, line 415: If `val[i]` must be a `$` character, we might as well debug assert that.
```suggestion
    fn parseEnvSubstitution(this: *Parser, val: []const u8, start: usize, i: usize, unesc: *std.ArrayList(u8)) ?usize {
      bun.debugAssert(val[i] == '$'); 
```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+        if (i + 1 < val.len and val[i + 1] == '{') {
+            if (i + 2 >= val.len) return null;
```
Jarred-Sumner:
```suggestion

```

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+
+            var found_closing = false;
+            var j = i + 2;
+            while (j < val.len) : (j += 1) {
+                switch (val[j]) {
+                    '$' => return this.parseEnvSubstitution(val, start, j, unesc),
+                    '{' => return null,
+                    '}' => {
+                        found_closing = true;
```
Jarred-Sumner, line 427: Are escape characters supported?

src/ini.zig
```diff
@@ -0,0 +1,1076 @@
...
+    /// - `i` must be an index into `val` that points to a '$' char
+    fn parseEnvSubstitution(this: *Parser, val: []const u8, start: usize, i: usize, unesc: *std.ArrayList(u8)) ?usize {
+        if (i + 1 < val.len and val[i + 1] == '{') {
+            if (i + 2 >= val.len) return null;
+
+            var found_closing = false;
+            var j = i + 2;
+            while (j < val.len) : (j += 1) {
+                switch (val[j]) {
+                    '$' => return this.parseEnvSubstitution(val, start, j, unesc),
```
Jarred-Sumner: Can you add a comment with the line of code from npm's ini parser where they do nested environment subsutitions like this?

robobun:
@zackradisic, your commit e017ee692be4a72afe71490e470e940a403ff713 is building: [`#424`](https://buildkite.com/bun/bun/builds/424)

src/js_ast.zig
```diff
@@ -3360,6 +3365,11 @@ pub const Expr = struct {
         return ArrayIterator{ .array = array, .index = 0 };
     }
 
+    pub inline fn asUtf8StringLiteral(expr: *const Expr) ?string {
+        if (expr.data == .e_string) return expr.data.e_string.data;
```
dylan-conway: do we need to resolve rope strings here, or assert they're already resolved?

src/bun.js/bindings/JSPropertyIterator.cpp
```diff
@@ -37,6 +38,7 @@ extern "C" JSPropertyIterator* Bun__JSPropertyIterator__create(JSC::JSGlobalObje
     JSC::VM& vm = globalObject->vm();
     JSC::JSValue value = JSValue::decode(encodedValue);
     JSC::JSObject* object = value.getObject();
+    ASSERT(object != NULL);
```
Jarred-Sumner, line 41: This assertion doesn't do anything, getObject() will crash before this line is reached if that assertion wasn't true

src/bun.js/bindings/helpers.h
```diff
@@ -335,6 +326,16 @@ static JSC::JSValue getRangeErrorInstance(const ZigString* str, JSC__JSGlobalObj
     return JSC::JSValue(result);
 }
 
+static const JSC::Identifier toIdentifier(ZigString str, JSC::JSGlobalObject* global)
+{
+    if (str.len == 0 || str.ptr == nullptr) {
+        return JSC::Identifier::EmptyIdentifier;
+    }
+    WTF::String wtfstr = Zig::isTaggedExternalPtr(str.ptr) ? toString(str) : Zig::toStringCopy(str);
```
Jarred-Sumner, line 334: What was the reason for this change?

zackradisic: Previously it was just turning the raw bytes into JSC::Identifier. Which meant that any utf8 string would get turned into latin1

src/bun.js/bindings/JSPropertyIterator.cpp
zackradisic: The implementation of getObject() is this which returns a null pointer
```cpp
inline JSObject* JSValue::getObject() const
{
    return isCell() ? asCell()->getObject() : nullptr;
}
```

Jarred-Sumner: isCell() when given a null pointer returns true (yes, surprising)

asCell()->getObject() then de-references the null pointer and 💥

src/ini.zig
```diff
@@ -0,0 +1,1204 @@
...
+        }
+    }
+
+    const had_errors = log.hasErrors();
+    log.printForLogLevel(Output.errorWriter()) catch bun.outOfMemory();
+    if (had_errors) {
+        return error.ParserError;
+    }
+
+    // if (!bun.Environment.isDebug) {
```
Jarred-Sumner, line 1190: is there a reason this is not deleted?

src/ini.zig
```diff
@@ -0,0 +1,1204 @@
...
+                            conf_item.registry_url,
+                        },
+                    ) catch bun.outOfMemory();
+                }
+            }
+        }
+    }
+
+    const had_errors = log.hasErrors();
+    log.printForLogLevel(Output.errorWriter()) catch bun.outOfMemory();
```
Jarred-Sumner, line 1185: why are we printing here instead of in the caller?

test/cli/install/registry/bun-install-registry.test.ts
```diff
@@ -72,6 +97,287 @@ registry = "http://localhost:${port}/"
...
+
+      await Bun.$`echo ${ini} > ${packageDir}/.npmrc`;
+
+      const result = loadNpmrc(ini);
+
+      check(result);
+    });
+  }
+
+  // await makeTest([["_authToken", "skibidi"]], result => {
```
Jarred-Sumner, line 251: Why is this commented out? Should it be deleted or should it be uncommented?

src/ini.zig
```diff
@@ -0,0 +1,1204 @@
...
+        .err => |err| {
+            // TODO: should this exit(1)?
+            Output.prettyErrorln("{}\nwhile reading .npmrc \"{s}\"", .{
+                err,
+                ".npmrc",
+            });
+            Global.exit(1);
+        },
+    }
+    const source = bun.logger.Source.initPathString(".npmrc", npmrc_contents.items[0..]);
```
Jarred-Sumner, line 914: Is there a reason we have another implementation of reading a file from disk, instead of using `bun.sys.File.toSource` or `bun.sys.File.toSourceAt`?

src/ini.zig
```diff
@@ -0,0 +1,1204 @@
...
+                                                .source = this.source,
+                                                .allocator = this.allocator,
+                                            };
+                                            break :brk parser.parseRegistryURLStringImpl(str) catch |e| {
+                                                if (e == error.OutOfMemory) bun.outOfMemory();
+                                                return error.ParserError;
+                                            };
+                                        }
+                                    }
+                                    return .none;
```
Jarred-Sumner, line 866: the only return thats necessary is the last one right?

Jarred-Sumner:
@zackradisic looks like the test failures in bun-install-registry are real

Mouvedia:
[This](https://github.com/oven-sh/bun/blob/main/docs/install/registries.md?plain=1#L28-L30) should be updated.

docs/install/npmrc.md
Mouvedia, line 1: Are [`public-hoist-pattern`](https://pnpm.io/npmrc#public-hoist-pattern), `legacy-peer-deps` planned?

`legacy-peer-deps` would roughly correspond to `peer = false` with this caveat:

> [_`legacy-peer-deps`_ differs from `--omit=peer`, in that `--omit=peer` will avoid unpacking `peerDependencies` on disk, but will still design a tree such that `peerDependencies` could be unpacked in a correct place.](https://docs.npmjs.com/cli/v10/using-npm/config#legacy-peer-deps)

PR #11983, opened 2024-06-19
feat(bundler): Add `--sourcemap=linked` for `//# sourceMappingURL` comments

### LINKED ISSUE #3325: Sourcemaps should contain `sourceMappingURL`

0.6.10

macos

Currently no source maps are working when using `sourcemaps: 'external'`. `sourceMappingURL` is not appended.

It also breaks `stacktrace-gps` package which requires this.

Confirm via `Developer Resources` in DevTools: https://developer.chrome.com/docs/devtools/developer-resources/

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #9314: bun build linked sourcemap option

source map external does not link to the source map preventing browser from automatically loading them

add an option to the source map flag named linked the will add a source mapping comment like this
`//# sourceMappingURL=index.js.map`

I made consider internal but I don't serve source maps to users that don't have a enough permission and I cant limit the internal sourcemap

paperclover:
### What does this PR do?

This implements --sourcemap=linked exactly as esbuild does so. ~~This also changes the default behavior of --sourcemap to linked~~ nvm, it was inline and will stay inline. When using external sourcemaps, it is almost always a better idea to use linked sourcemaps, as that is what browsers and most tooling supports. Having unlinked external sourcemaps (`--sourcemap=external`) can be useful for proprietary codebases where the source map and the bundled code are meant to be isolated.

For more context, see https://esbuild.github.io/api/#sourcemap

also fixes this disabled test because some of these chars are hex lowercase which is silly.
image

Fixes #3325
Fixes #9314


paperdave:
wuh image

src/bundler/bundle_v2.zig
```diff
@@ -9336,12 +9336,27 @@ const LinkerContext = struct {
                 );
 
                 switch (c.options.source_maps) {
-                    .external => {
+                    inline .external, .linked => |tag| {
```
Jarred-Sumner: Does this need the `inline`?
```suggestion
                    .external, .linked => |tag| {
```

src/bundler/bundle_v2.zig
```diff
@@ -9520,13 +9535,26 @@ const LinkerContext = struct {
             );
 
             switch (c.options.source_maps) {
-                .external => {
+                inline .external, .linked => |tag| {
```
Jarred-Sumner: Does this need the `inline`?
```suggestion
                    .external, .linked => |tag| {
```

Jarred-Sumner:
What about `public_path`?

Jarred-Sumner:
We're probably also going to want to allow a different base url as an option for sourcemaps. I remember at Stripe they put the sourcemaps on a corporate-only subdomain so that any employee connected to the VPN could access them easily, but it remained internal otherwise

Jarred-Sumner:
IMO, the behavior should be:
- If --outdir is specified, `--sourcemap` means linked
- If --outdir is not specified, `--sourcemap` means inline

src/bundler/bundle_v2.zig
paperclover: this makes `tag` comptime, which is what it is being used in the `if`. i guess that isn't strictly needed for this use-case. i can remove.

paperdave:
> What about `public_path`?

ah, yeah that should be handled. i overlooked that.

> I remember at Stripe they ...

that makes sense, but i would recommend for cases like this to use `--sourcemap=external`, which would not put the URL in at all outside of debug builds. This seems to be what [stripe.com does](https://b.stripecdn.com/mkt-statics-srv/assets/PricingIcon-YFA54ZVS.js) (sourceMappingURL does not appear in their JS bundles)

> IMO, the behavior should be: [behavior]

Ok. It will be a bit extra to document but we can do that. I was hesitant to change defaults much. I think this will "just work" for most people, but it does feel a bit odd to have it vary in this way (i like what esbuild does, but that would be a breaking api change for us).

paperdave:
> We're probably also going to want to allow a different base url as an option for sourcemaps, like `--sourcemap-url`

I don't think that is necessary, since esbuild doesn't have this. We maybe want to add `--source-root`, but anything else we should just wait for users to feature-request.

Issue #11954, opened 2024-06-18
crash: Failed to enable File Watcher (bun --hot)


huseeiin:
### Stack Trace (bun.report)

Bun v1.1.14-canary ([`9621721`](<https://github.com/oven-sh/bun/tree/9621721e3df4f618cc1f975b043dbefd7fd61c07>)) on linux x86_64 [AutoCommand]

**panic**: Failed to enable File Watcher

- [`javascript.zig:3836`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/bun.js/javascript.zig#L3836>): `src.bun.js.javascript.NewHotReloader`
- [`bun_js.zig:267`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/bun_js.zig#L267>): `src.bun.js.javascript.OpaqueWrap__anon_53574__struct_287229.callback`
- [`bindings.cpp:4633`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/bun.js/bindings/bindings.cpp#L4633>): `JSC__VM__holdAPILock`
- [`shimmer.zig:186`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/bun.js/bindings/shimmer.zig#L186>): `src.bun_js.Run.boot`
- [`cli.zig:2016`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/cli.zig#L2016>): `src.cli.Command.maybeOpenWithBunJS`
- [`cli.zig:1883`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/cli.zig#L1883>): `src.cli.Cli.start`
- [`main.zig:50`](<https://github.com/oven-sh/bun/blob/9621721e3df4f618cc1f975b043dbefd7fd61c07/src/main.zig#L50>): `main`

Features: jsc, tsconfig

paperdave:
When my linked PR is merged, can you re-try your reproduction? otherwise i will close as duplicate of #6856 / has no reproduction.

huseeiin:
> When my linked PR is merged, can you re-try your reproduction? otherwise i will close as duplicate of #6856 / has no reproduction.

i lost my code :)) you can close as duplicate

Issue #6856, opened 2023-11-02
Failed to enable File Watcher


sparseinference:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Linux 6.5.0-10-generic x86_64 x86_64

### What steps can reproduce the bug?

Run Bun with the simple demo server from the website:

> bun --hot bun_server.tsx

```js
const server = Bun.serve({
    port: 3000,
    fetch(request) {
      return new Response("Welcome to Bun!");
    },
  });
  
  console.log(`Listening on localhost:${server.port}`);
```

### What is the expected behavior?

I expected a running server ...

### What do you see instead?

uh-oh: Failed to enable File Watcher
bun will crash now 😭😭😭
----- bun meta -----
Bun v1.0.7 (b0393fba) Linux x64 #10-Ubuntu SMP PREEMPT_DYNAMIC Fri Oct 13 13:49:38 UTC 2023
AutoCommand: 
Elapsed: 8ms | User: 4ms | Sys: 8ms
RSS: 33.55MB | Peak: 29.15MB | Commit: 33.55MB | Faults: 0
----- bun meta -----

0   0x55b6e19179bb
1   ???
2   ???
3   ???
4   ???
5   ???
6   ???
7   ???
8   ???
9   ???
10  ???
11  ???

Crash report saved to:
  ~/.bun/.bun-crash/v1.0.7-1698951924684.crash

### Additional information

Sometimes instead of crashing and exiting the process, hundreds of crash report files are saved before you use CTRL-C  on it.

Also, this system was just updated to Kubuntu 23.10 from Kubuntu 23.04 today. There are a lot of processes running and many UI windows open (mostly Chrome tabs).

The file watching features of other web servers are also failing - for example with deno:

> Task start deno run -A --watch=static/,routes/ dev.ts
> Watcher Process started.
> error: Too many open files (os error 24)

PR #11986, opened 2024-06-20
log the error when file watcher fails to start

paperclover:
### What does this PR do?

Note for writing code, please do not use `thingThatCanFailForManyReasons() catch @panic("Failed to <not specific action>")`. Vauge user errors are not useful. Note that specific panic messages can be fine, but this one is too broad. 

We now have `bun.handleErrorReturnTrace(err, @errorReturnTrace());`, which when error return traces get enabled on all platforms, will let us debug these in production more easily, and `@errorName(err)` can print the error code.

The following issues
- #11954
- #6856

Do not have reproductions, and the first step would be to identify WHAT the error is. And then knowing where makes it even easier to debug.

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -3833,13 +3833,19 @@ pub fn NewHotReloader(comptime Ctx: type, comptime EventLoopType: type, comptime
                         reloader,
                         this.bundler.fs,
                         bun.default_allocator,
-                    ) catch @panic("Failed to enable File Watcher") }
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
`fs.watchFile` is 100% unrelated btw. fs.watchFile does not use the filesystem watcher. It polls.

paperdave:
that issue then has a wrong title. it is a crash report for "Failed to enable File Watcher"

PR #11987, opened 2024-06-20
fix: execArgv sometimes could include user args

### LINKED ISSUE #11673: process.execArgv may incorrectly include user options

1.1.12

Darwin 23.5.0 arm64 arm

Bun is sometimes duplicating all the script options into `process.execArgv`.

`process.execArgv` should only include options for the node application itself. Options coming after the script name are for the script being run and should not appear in `process.execArgv`. 

https://nodejs.org/api/process.html#processexecargv

> The `process.execArgv` property returns the set of Node.js-specific command-line options passed when the Node.js process was launched. These options do not appear in the array returned by the [process.argv](https://nodejs.org/api/process.html#processargv) property, and do not include the Node.js executable, the name of the script, or any options following the script name. These options are useful in order to spawn child processes with the same execution environment as the parent.The process.execArgv property returns the set of Node.js-specific command-line options passed when the Node.js process was launched. These options do not appear in the array returned by the [process.argv](https://nodejs.org/api/process.html#processargv) property, and do not include the Node.js executable, the name of the script, or any options following the script name. These options are useful in order to spawn child processes with the same execution environment as the parent.

Example program:
```js
console.log({ argv: process.argv, execArgv: process.execArgv });
```

```console
% bun run args.js -a -b -c    
{
  argv: [ "/opt/homebrew/Cellar/bun/1.1.12/bin/bun", "/Users/john/Documents/Sandpits/commander/issues/2205/args.js",
    "-a", "-b", "-c"
  ],
  execArgv: [],
}


% bun run args.js foo -a -b -c
{
  argv: [ "/opt/homebrew/Cellar/bun/1.1.12/bin/bun", "/Users/john/Documents/Sandpits/commander/issues/2205/args.js",
    "foo", "-a", "-b", "-c"
  ],
  execArgv: [],
}
```

### What do you see instead?

Depending on the mixture of script arguments, the script options may incorrectly appear in `process.execArgv`.

```console
% bun run args.js -a -b -c    
{
  argv: [ "/opt/homebrew/Cellar/bun/1.1.12/bin/bun", "/Users/john/Documents/Sandpits/commander/issues/2205/args.js",
    "-a", "-b", "-c"
  ],
  execArgv: [],
}


% bun run args.js foo -a -b -c
{
  argv: [ "/opt/homebrew/Cellar/bun/1.1.12/bin/bun", "/Users/john/Documents/Sandpits/commander/issues/2205/args.js",
    "foo", "-a", "-b", "-c"
  ],
  execArgv: [ "-a", "-b", "-c" ],
}
```

### Additional information

I am reporting this because of an issue opened against Commander by a Bun user. When they upgraded Commander, their program broke because Commander has added auto-detection support for the node mode options like `-e` and `--print`. The auto-detection misfires because the script `-e` option is also appearing in `process.execArgv`.

- https://github.com/tj/commander.js/issues/2205
- https://github.com/tj/commander.js/pull/2164

paperclover:
### What does this PR do?

Fixes #11673 (this issue is a good write up)

### DIFF
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -4912,7 +4912,7 @@ pub const Process = struct {
 
         // we re-parse the process argv to extract execArgv, since this is a very uncommon operation
         // it isn't worth doing this as a part of the CLI
-        for (bun.argv[1..]) |arg| {
+        for (bun.argv[@min(1, bun.argv.len)..]) |arg| {
             defer prev = arg;
 
             if (arg.len >= 1 and arg[0] == '-') {
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -4889,7 +4889,8 @@ pub const Process = struct {
     }
 
     pub fn getExecArgv(globalObject: *JSC.JSGlobalObject) callconv(.C) JSC.JSValue {
-        const allocator = globalObject.allocator();
+        var sfb = std.heap.stackFallback(4096, globalObject.allocator());
+        const temp_alloc = sfb.get();
         const vm = globalObject.bunVM();
 
         if (vm.worker) |worker| {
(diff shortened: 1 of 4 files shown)


src/bun.js/node/types.zig
```diff
@@ -4903,32 +4904,62 @@ pub const Process = struct {
...
+        defer args.deinit();
+
+        var seen_run = false;
+        var prev: ?[]const u8 = null;
 
-        for (bun.argv[@min(bun.argv.len, offset)..]) |arg| {
-            if (arg.len == 0)
+        // we re-parse the process argv to extract execArgv, since this is a very uncommon operation
+        // it isn't worth doing this as a part of the CLI
+        for (bun.argv[1..]) |arg| {
```
```suggestion
        for (bun.argv[@min(bun.argv.len, 1)..]) |arg| {
```

paperclover: is it possible to start the runtime with *zero* arguments?

Jarred-Sumner: it should be possible, yes

PR #12006, opened 2024-06-20
[FIX]: Made the LICENCE a markdown file to be previewable and minor fixes in markdown syntax

### LINKED ISSUE #11985: LICENSE file should probably have .md / .MD file extension

Something else

### What is the issue?

The `LICENSE` file is formatted like a Markdown file but does not have a file extension.
That means GitHub does not parse it as a Markdown but simply as plain text.

First, you can notice that due to the Markdown syntax being visible like `## JavaScriptCore` on [line 3](https://github.com/oven-sh/bun/commit/d55ac0fa434d31503147faefef5340919735c2ad#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7R3), or ``[`libarchive`](https://github.com/libarchive/libarchive)`` on [line 40](https://github.com/oven-sh/bun/commit/d55ac0fa434d31503147faefef5340919735c2ad#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7R40).

Secondly, it is indeed meant to be a Markdown file, since it was [moved from `docs/project/licensing.md` to `LICENSE`](https://github.com/oven-sh/bun/commit/d55ac0fa434d31503147faefef5340919735c2ad) file in the root directory. And as you can see, `licensing.md` does indeed have the markdown file extension.

So, the `LICENSE` file should be renamed to `LICENSE.md` or `LICENSE.MD`.

In the end, it's not an important issue, but yet it makes the file hard to read and not accessible.

### Where did you find it?

https://github.com/oven-sh/bun/blob/main/LICENSE

mohiwalla:
Fixes #11985 

image

### DIFF
--- a/test/js/web/websocket/websocket.test.js
+++ b/test/js/web/websocket/websocket.test.js
@@ -529,8 +529,8 @@ describe("WebSocket", () => {
       await openAndCloseWS();
       if (i % 100 === 0) {
         current_websocket_count = getWebSocketCount();
-        // if we have more than 10 websockets open, we have a problem
-        expect(current_websocket_count).toBeLessThanOrEqual(10);
+        // if we have more than 20 websockets open, we have a problem
+        expect(current_websocket_count).toBeLessThanOrEqual(20);
         if (initial_websocket_count === 0) {
           initial_websocket_count = current_websocket_count;
         }


paperclover (changes requested):
LICENCE -> LICEN**S**E

image

[ref](https://www.grammarly.com/blog/licence-license)

mohiwalla:
Sorry @paperdave, I didn't knew how to made changes in same PR, created a new one instead.

PR #12020, opened 2024-06-20
Fixes #12012

### LINKED ISSUE #12012: The returned changes property from statement.run() method is always zero (bun:sqlite)

1.1.15

Linux and Windows

```ts
import { Database } from 'bun:sqlite'

const db = new Database(':memory:')
db.exec('CREATE TABLE cats (id INTEGER PRIMARY KEY, name TEXT)')

const sql = "INSERT INTO cats (name) VALUES ('Fluffy'), ('Furry')"

console.log(db.run(sql)) // yep
console.log(db.prepare(sql).run()) // nope
console.log(db.query(sql).run()) // nope
```

Number of changes is correct, as it is for db.run()

### What do you see instead?

While lastInsertRowid is correct, changes is always `0` for prepared statements it seems.

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #12012

Test

### DIFF
--- a/src/bun.js/bindings/sqlite/JSSQLStatement.cpp
+++ b/src/bun.js/bindings/sqlite/JSSQLStatement.cpp
@@ -2219,6 +2219,8 @@ JSC_DEFINE_HOST_FUNCTION(jsSQLStatementExecuteStatementFunctionRun, (JSC::JSGlob
         DO_REBIND(arg0);
     }
 
+    int total_changes_before = sqlite3_total_changes(castedThis->version_db->db);
+
     int status = sqlite3_step(stmt);
     if (!sqlite3_stmt_readonly(stmt)) {
         castedThis->version_db->version++;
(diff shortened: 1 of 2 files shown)


src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -2219,6 +2219,8 @@ JSC_DEFINE_HOST_FUNCTION(jsSQLStatementExecuteStatementFunctionRun, (JSC::JSGlob
         DO_REBIND(arg0);
     }
 
+    int total_changes_before = sqlite3_total_changes(castedThis->version_db->db);
```
dylan-conway, line 2222: I'm not familiar with the sqlite api, was `sqlite3_step` or `sqlite3_stmt_readonly` resetting the total change count?

Jarred-Sumner: If step only runs once then it would not report any changes

dylan-conway (approved):

PR #12022, opened 2024-06-21
Fix `bun patch` with workspaces and scoped packages

### LINKED ISSUE #11971: `bun patch --commit` fails for scoped packages

1.1.15-canary.1+3ff29955a

Darwin 23.5.0 arm64 arm

1. `bun patch @sanity/rich-date-input`
2. Make changes to some files in `node_modules/@sanity/rich-date-input`
3. `bun patch --commit 'node_modules/@sanity/rich-date-input'`
4. Error

`bun patch --commit` finishes without error and correctly creates the patch file.

### What do you see instead?

`bun patch --commit` fails:

image

But the stated file does exist:

image

Nevertheless, the patch file is created (albeit encoding `/` into `%2F`):

image

But the patch file, in addition to containing the applied changes, also has the rest of the untouched files:

image

image

If I edit the patch file to only contain the actual changes, then `bun install` correctly applies the patch and everything works. So the problem is in `bun patch --commit` for —I think— scoped packages. I can even edit the patch file name to use a `+` instead of `%2F` (like `patch-package` does) and Bun accepts that:

image

### Additional information

_No response_

zackradisic:
### What does this PR do?

Fixes #11971

Two issues:
- `bun patch` commands not executing correctly because package manager changes working directory to workspace root
- bug in `bun patch --commit` causing nested node_modules to be diffed in certain cases

Also fixed:
- bug in `bun.strings.containsComptime`


test/cli/install/bun-patch.test.ts
```diff
@@ -318,7 +468,7 @@ Once you're done with your changes, run:
     15 * 1000,
   );
 
-  ["is-even@1.0.0", "node_modules/is-even"].map(patchArg =>
+  [("is-even@1.0.0", "node_modules/is-even")].map(patchArg =>
```
dylan-conway:
```suggestion
  ["is-even@1.0.0", "node_modules/is-even"].map(patchArg =>
```

dylan-conway (approved):

PR #12023, opened 2024-06-21
fix mock function crash

paperclover:


Jarred-Sumner (changes requested):
Can you add a test?

Jarred-Sumner:
@paperdave you added `base64` as a git submodule again

we cannot merge this until that's addressed 


image

PR #12030, opened 2024-06-21
fix(install): fix potential flakiness with git dependencies

dylan-conway:
### What does this PR do?
`manager.git_dependencies` was used without a lock in a thread callback. Moved it to main thread. This should fix flaky tests we're seeing in `complex-workspace.test.ts` in CI.

CI


Jarred-Sumner:
Nice. But we should later just use bun.spawn for this so that it always runs the spawn step on the main thread. No need to thread this

dylan-conway:
> Nice. But we should later just use bun.spawn for this so that it always runs the spawn step on the main thread. No need to thread this

Agreed

PR #12046, opened 2024-06-21
Fixed proxies getting ignored by http

### LINKED ISSUE #12005: Bunjs axios does not support proxy

1.1.15+b23ba1fe1

Linux 6.5.0-35-generic x86_64 x86_64

The error occurs when using the HttpsProxyAgent library. When axios request is sent with proxy, bunjs ignores the proxy information and sends the request with the computer ip address. But Nodejs sends it with proxy information.
You can use any proxy for testing. The code was executed unchanged. 

Code:

```
const axios = require('axios');
const { HttpsProxyAgent } = require('https-proxy-agent');
const httpsAgent = new HttpsProxyAgent(`http://<proxy-username>:<proxy-password>@<proxy-host>:<proxy-port>`)
axios.get('https://ipapi.co/json', {
    httpsAgent,
}).then(res => {
    console.log(res.data);

}).catch(err => {
    console.log(err.message);
})
```

Node.js Result
image
Bun Result
image

If we use axios' local proxy support instead, it doesn't pass auth and returns a 407 error.

Code:
```
const axios = require('axios');
axios.get('https://ipapi.co/json', {
    proxy: {
        protocol: 'http',
        host: "proxy-host",
        port: "proxy-port",
        auth: {
            username: "proxy-username",
            password: "proxy-password"
        }
    }
}).then(res => {
    console.log(res.data);

}).catch(err => {
    console.log(err.message);
})
```
Node.js Result:
image

Bun Result:
image

The expected behavior is that bun sends the request with a proxy like Node.js. 

### What do you see instead?

Bun HttpsProxyAgent library ignores proxy information. Axios' native proxy support also doesn't send username and password information. The same code works fine with Node.js but Bunjs doesn't use proxy.

### Additional information

_No response_

LeoWiniecki-PowerSystems:
### What does this PR do?

Fixes #12005 . The axios adapter passes the proxy url through the agents in the http options.

Confirmed that proxies are now successfully passed to fetch.

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -1411,6 +1411,7 @@ class ClientRequest extends OutgoingMessage {
       proxy = `${protocol}//${this.#host}${this.#useDefaultPort ? "" : ":" + this.#port}`;
     } else {
       url = `${protocol}//${this.#host}${this.#useDefaultPort ? "" : ":" + this.#port}${path}`;
+      proxy = this.#options.agents?.http?.proxy?.href ?? this.#options.agents?.https?.proxy?.href;
     }
     const tls = protocol === "https:" && this.#tls ? { ...this.#tls, serverName: this.#tls.servername } : undefined;
     try {


Jarred-Sumner:
Closing due to test failures + inactivity. If you're interested in working on this, please add tests

PR #12048, opened 2024-06-21
Use slow move-based fallback for `renameatConcurrently`

zackradisic:
### What does this PR do?

This PR fixes `bun patch` not working when the temp/install dir are not on the same filesystem as the user's project.

This updates the `bun.sys.renameatConcurrently` function to conditionally fallback to a slow move-based implementation when it hits `EXDEV`


src/copy_file.zig
```diff
@@ -26,6 +26,35 @@ const CopyFileError = error{SystemResources} || CopyFileRangeError || posix.Send
 
 const InputType = if (Environment.isWindows) bun.OSPathSliceZ else posix.fd_t;
 
+pub fn copyFileErrorConvert(e: CopyFileError) bun.C.E {
```
Jarred-Sumner: nooo this should be the other way around

PR #12065, opened 2024-06-22
Remove extraneous `Bun.ArrayBufferSink` mention.

bomberstudios:
### What does this PR do?

This removes a mention to `Bun.ArrayBufferSink` that appears to be a leftover on the [Bun.openInEditor()](https://bun.sh/docs/api/utils#bun-openineditor) method documentation.


Jarred-Sumner:
Thanks

PR #12104, opened 2024-06-24
Make reported node.js version a build option

Jarred-Sumner:
### What does this PR do?

Make reported node.js version a build option, so it's consistent everywhere and when we change it later we only need to update one place

### DIFF
--- a/src/bun.js/bindings/BunProcess.cpp
+++ b/src/bun.js/bindings/BunProcess.cpp
@@ -77,9 +77,6 @@ namespace Bun {
 
 using namespace JSC;
 
-#ifndef REPORTED_NODEJS_VERSION
-#define REPORTED_NODEJS_VERSION "22.3.0"
-#endif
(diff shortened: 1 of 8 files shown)


src/bun.js/bindings/BunProcess.cpp
```diff
@@ -77,7 +77,9 @@ namespace Bun {
 
 using namespace JSC;
 
-#define REPORTED_NODE_VERSION "22.2.0"
+#ifndef REPORTED_NODEJS_VERSION
+#define REPORTED_NODEJS_VERSION "22.3.0"
```
paperclover: if the build system is going to define this, we shouldnt need a fallback here. these lines can be removed entirely.

it feels weird to have a (ideally unused) fallback value.

similar comment can be said about the build.zig but that default is more tolerable because `zig build check` is convenient to have no arguments on.

paperclover (changes requested):

PR #12126, opened 2024-06-24
install: Detect cross-FS cache directory and find best one

zackradisic:
### What does this PR do?

WIP need tests

Right now in the package manager, on initialization, we select an install cache directory and a temp directory that are on the same filesystem. This is because copy/moving files between filesystems is slow.

However, we don't ensure that the cache directory and the user's project directory (directory where `package.json` is) are on the same file system. If this is the case, then `bun install` becomes very slow.

This PR detects if the cache directory is on another filesystem from the project directory. If that is the case, then we either:
1. warn the user (if they explicitly set the cache directory with an env variable)
2. attempt to find the "best" cache directory on the same FS as the project dir


src/sys.zig
```diff
@@ -443,6 +443,90 @@ pub fn Maybe(comptime ReturnTypeT: type) type {
...
+/// This function does this test by trying to rename a file named from `dir_with_file_to_rename` -> `dir_to_rename_file_to`
+///
+/// The file must have the name `tmpname`
+pub fn testSameFileSystem(
+    dir_with_file_to_rename: std.fs.Dir,
+    dir_to_rename_file_to: std.fs.Dir,
+    tmpname: [:0]const u8,
+) bool {
+    // Make sure cachedir and cwd are in the same filesystem
+    std.posix.renameatZ(dir_with_file_to_rename.fd, tmpname, dir_to_rename_file_to.fd, tmpname) catch return false;
```
Jarred-Sumner, line 457: None of these should be using `std.posix`

Jarred-Sumner: and this doesn't test they're in the same filesystem. This tests it doesn't throw an error.

Jarred-Sumner: and it shouldn't be in bun.sys unless it's reusable elsewhere

src/sys.zig
```diff
@@ -443,6 +443,90 @@ pub fn Maybe(comptime ReturnTypeT: type) type {
...
+        }
+        this.start = slash_idx + 1;
+        return this.path_to_iterate[0..slash_idx];
+    }
+};
+
+/// Invariants:
+///
+/// - node_modules_folder_path is absolute path with no relative syntax
+pub fn findBestDirectoryInSameFileSystem(
```
Jarred-Sumner, line 495: This can go in the package manager code

we don't need to do this elsewhere afaict

PR #427, opened 2022-07-08
fix(templates/react): Add a SVG type definition

FinnRG:
Closes #315.

PR #12144, opened 2024-06-25
feat(bundler): implement enum inlining / more constant folding

### LINKED ISSUE #2889: `{ __proto__: null }.__proto__` gets inlined incorrectly by bun.

0.6.0

_No response_

Run this file:
```ts
eval('if ({ __proto__: null }.__proto__ !== void 0) throw "fail in eval";');
      if ({ __proto__: null }.__proto__ !== void 0) throw "fail at runtime";
```

the first line runs the code without transpilation, the second line gets transpiled

the proto access should be `undefined`, not `null`. neither line should throw.

### What do you see instead?

`error: fail at runtime`

### Additional information

similar situations
- `({ __proto__: 5 }.__proto__)` and other primatives should inline to `{}`
- `({ __proto__: { x: 1 } }.__proto__)` already inline correctly to `{ x: 1 }`

### LINKED ISSUE #2945: Inline `const enum` in `bun build`

0.6.2

Linux 5.19.0-41-generic x86_64 x86_64

1. Create a file called `bug.ts`.
2. Fill with
```ts
const enum ConstantEnum {
    First,
    Second,
}
console.log(ConstantEnum.First, ConstantEnum.Second)
```
3. Run `bun build ./bug.ts`

The const enum should be have instances of its values replaced, instead of just referenced (like in [this TypeScript playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBApmArgWxgYXNAhmKBRJVAbwCgZyYAxASwCdoAaMigZTlDABMmBfEjiCAA2cAHRCQAcwAUGSFBz5ComvSgN0mBbgIpRbDpwCUQA)).

### What do you see instead?

The const enum acts the exact same as a normal enum, simply adding an IIFE to create an object.

### Additional information

_No response_

### LINKED ISSUE #11764: Self-referencing enums are mangled

1.1.12+43f0913c3

`self-ref.ts` - a simple self-referencing enum:

```typescript
enum E {
  One = "foo",
  Two = One,
}

console.log(E);
```

`self-ref-same-value.ts` - a self-ref enum with a referenced entry which its value is the same string as the name:

```typescript
enum E {
  Common = "Common",
  Alt1 = Common,
  Alt2 = Common,
}

console.log(E);
```

The console output:

```console
> bun self-ref.ts
{
  One: "foo",
  Two: "foo",
}

> bun self-ref-same-value.ts
{
  Common: "Common",
  Alt1: "Common",
  Alt2: "Common",
}
```

### What do you see instead?

Values are correct but there's an extra key:

```console
> bun self-ref.ts
{
  One: "foo",
  Two: "foo",
  foo: "Two",
}
```

Values are completely wrong:

```console
> bun self-ref-same-value.ts
{
  Common: "Alt1",
  Alt1: "Alt2",
  Alt2: "Alt1",
}
```

### Additional information

Code generated by `bun build`:

`self-ref.ts`

```javascript
// self-ref.ts
var E;
(function(E2) {
  E2["One"] = "foo";
  E2[E2["Two"] = E2.One] = "Two";
})(E || (E = {}));
console.log(E);
```

`self-ref-same-value.ts`

```javascript
// self-ref-same-value.ts
var E;
(function(E2) {
  E2["Common"] = "Common";
  E2[E2["Alt1"] = E2.Common] = "Alt1";
  E2[E2["Alt2"] = E2.Common] = "Alt2";
})(E || (E = {}));
console.log(E);
```

---

Code generated by `tsc` (5.4.3), for reference:

`self-ref.ts`

```javascript
var E;
(function (E) {
    E["One"] = "foo";
    E["Two"] = "foo";
})(E || (E = {}));
console.log(E);
```

`self-ref-same-value.ts`

```javascript
var E;
(function (E) {
    E["Common"] = "Common";
    E["Alt1"] = "Common";
    E["Alt2"] = "Common";
})(E || (E = {}));
console.log(E);
```

### LINKED ISSUE #12285: Reduce the size of JS code when using `--minify-syntax` and enum

Reduce the size of JS code when using `--minify-syntax` and enum

In the current version of Bun (1.1.17), this TS code 

```ts
enum VeryLongEnumName {
  A = 'a',
  B = 'b',
  C = 'c',
  D = 'd',
}
```

is being transformed into this JS code (same output when using `--minify-syntax`)

```js
var VeryLongEnumName;
(function(VeryLongEnumName2) {
  VeryLongEnumName2["A"] = 'a';
  VeryLongEnumName2["B"] = 'b';
  VeryLongEnumName2["C"] = 'c';
  VeryLongEnumName2["D"] = 'd';
})(VeryLongEnumName || (VeryLongEnumName = {}));
```

It'd be nice to have the `VeryLongEnumName2` variable replaced with a single-character variable name to reduce the size when using `--minify-syntax`. And use dot-notations to reduce the size even more.

```js
// Replace "VeryLongEnumName2" with "e"
(function(e) {
  e["A"] = 'a';
  e["B"] = 'b';
  e["C"] = 'c';
  e["D"] = 'd';
})(VeryLongEnumName || (VeryLongEnumName = {}));

// Even better: use dot-notations
(function(e) {
  e.A = 'a';
  e.B = 'b';
  e.C = 'c';
  e.D = 'd';
})(VeryLongEnumName || (VeryLongEnumName = {}));

// Or this
(function(e) {
  e.A = 'a', e.B = 'b', e.C = 'c', e.D = 'd';
})(VeryLongEnumName || (VeryLongEnumName = {}));

```

Thanks for your consideration.

_No response_

paperclover:
Initially was a PR to fix an enum bug, but it turned out the bug was due to a lack of a feature: enum inlining. So I decided to do it, since it was a feature I personally have a big use for (For example, the `bun.report` project has a regex hack workaround to force enum inlining, because the codegen otherwise is an abomination).

edit: As time went, this PR turned into not just that, but a lot of changes to constant folding. i am much happier with how the bundler handles this now.

Fixes #11764
Fixes #2945
Fixes #2889 
Fixes #12285
Fixes 
Draft TODO:

- [X] e_inlined_enum
- [X] inline enum as well as in simplify functions
- [X] Enum use-before-define behavior
- [x] Implement required changes in TS namespaces
- [x] Fix sibling namespace scopes
- [x] Cross-file enums while bundling
- [x] Tests
- Note: late-stage numeric folding (js printer) is not implemented, and will not be for now

extras:
- constnat folding now happens with all numeric operators
- constant folding now happens with `n + "baa"`, where n is any constant number. this uses webkit's code to properly adhere to JS spec
- constant folding now happens with `+"1.2"`. this uses webkit's code to properly adhere to JS spec
- constant folding requires `--minify-syntax`, where previously it seemed to **always** be applied. Though keep in mind the JS runtime transpiles all files with `--minify-syntax`
- constant folding is now enabled in the context of `require`, `import`, and macro calls; even if minify-syntax is disabled. This was previously not an issue because it was always enabled.
- string and template string folding is much more advanced now. this may need more tests

---

Simple Sample

```ts
enum A {
  Hello = 1,
  World,
}

console.log(A.Hello);
console.log(A.World);
console.log(A.Hello + A.World);
```
After:
```
// index.ts
console.log(1 /* Hello */);
console.log(2 /* World */);
console.log(3);
```
Before:
```js
// index.ts
var A;
(function(A2) {
  A2[A2["Hello"] = 1] = "Hello";
  A2[A2["World"] = 2] = "World";
})(A || (A = {}));
console.log(A.Hello);
console.log(A.World);
console.log(A.Hello + A.World);
```
---
If you reference the enum directly, you can get the two-way map. This even works with `const enum`, as it is TSC's job to prohibit such usage.

```ts
enum A {
  Hello = 1,
  World,
}

console.log(A);
```

The codegen of these enums have been improved. Projects with heavy enum usage should see smaller bundle sizes. Now:
```
var A;
((A2) => {
  A2[A2.Hello = 1] = "Hello";
  A2[A2.World = 2] = "World";
})(A ||= {});
```
Before:
```
// index.ts
var A;
(function(A2) {
  A2[A2["Hello"] = 1] = "Hello";
  A2[A2["World"] = 2] = "World";
})(A || (A = {}));
console.log(A);
```

It would be interesting as a future project to see if the size of these enums could be shrunk further, but `esbuild` doesn't do that, and I think personally it's fine to leave `enum` in the dust as `const enum` and inlining all values becomes the new norm.


robobun:
@paperdave, your commit 3aec7f1180787b5f68522e81103ef953b175ee28 is building: [`#427`](https://buildkite.com/bun/bun/builds/427)

src/js_ast.zig
```diff
@@ -1786,28 +1786,28 @@ pub const E = struct {
...
-                        return if (int_value < 0)
-                            neg_double_digit[abs]
-                        else
-                            double_digit[abs];
-                    }
+        /// "constant expression" handling in enums. We can match the behavior of a JS VM
+        /// by calling out to the APIs in WebKit which are responsible for this operation.
+        ///
+        /// This can return `null` in wasm builds to avoid linking JSC
+        pub fn toString(this: Number, allocator: std.mem.Allocator) ?string {
```
paperclover, line 1796: note: this is renamed from `toStringSafely` to `toString`, not because it is now less safe, but because the original intent of "we are taking a safer approach" does not apply.

build.zig
```diff
@@ -123,7 +123,7 @@ pub fn getOSGlibCVersion(os: OperatingSystem) ?Version {
 }
 
 pub fn build(b: *Build) !void {
-    std.debug.print("zig build v{s}\n", .{builtin.zig_version_string});
+    std.log.info("zig compiler v{s}", .{builtin.zig_version_string});
```
paperclover, line 126: apologies for doing a ton of other random things in this pr. i fixed the banned.json linter which resulted in needing to apply a ton of small transformations. i did this because i was sick of `std.debug.print` showing at all in my cmd+shift+f menu. this also resulted in the deletion of some old files, which some did not even past astgen.

additionally, i have been using a client side spell-checker, so it just becomes annoying to see a ton of yellow underlines. maybe i will upstream a ci check to keep spelling somewhat intact (i have been very impressed with the `typos` cli and lsp, it is very low false positive).

src/bun.zig
```diff
@@ -1955,10 +1955,10 @@ pub const ArenaAllocator = @import("./ArenaAllocator.zig").ArenaAllocator;
...
 
 pub const RegularExpression = @import("./bun.js/bindings/RegularExpression.zig").RegularExpression;
+
 pub inline fn assertComptime() void {
-    if (comptime !@inComptime()) {
-        @compileError("This function can only be called in comptime.");
-    }
+    var x = 0; // if you hit an error on this line, you are not in a comptime context
+    _ = &x;
 }
```
paperclover, line 1968: `comptime @inComptime` is a compiler error in 0.14 and in 0.13 does not do what we want.

src/feature_flags.zig
```diff
@@ -46,8 +46,6 @@ pub const allow_json_single_quotes = true;
 
 pub const react_specific_warnings = true;
 
-pub const log_allocations = false;
```
paperclover, line 49: this behavior is now provided via BUN_DEBUG_mimalloc=1

src/bundler/bundle_v2.zig
```diff
@@ -5758,8 +5783,49 @@ const LinkerContext = struct {
...
+            // accesses off of imported symbols are inlined enum values and
+            // which ones aren't
+            for (
+                part.import_symbol_property_uses.keys(),
+                part.import_symbol_property_uses.values(),
+            ) |ref, properties| {
+                const use = part.symbol_uses.getPtr(ref).?;
+
+                // Rare path: this import is a TypeScript enum
+                if (c.graph.meta.items(.imports_to_bind)[id].get(ref)) |import_data| {
```
Jarred-Sumner: Please hoist this `.items[id]` call above the for loop

src/js_ast.zig
```diff
@@ -6655,6 +6842,176 @@ pub const Span = struct {
...
+    ///   })(x || (x = {}));
+    ///   (function(x3) {
+    ///     x3.z = x3.y;
+    ///   })(x || (x = {}));
+    ///
+    /// When we try to find the symbol "y", we instead return one of these lazily
+    /// generated proxy symbols that represent the property access "x3.y". This
+    /// map is unique per namespace block because "x3" is the argument symbol that
+    /// is specific to that particular namespace block.
+    property_accesses: std.StringArrayHashMapUnmanaged(Ref) = .{},
```
Jarred-Sumner:
```suggestion
    property_accesses: bun.StringArrayHashMapUnmanaged(Ref) = .{},
```

src/js_printer.zig
```diff
@@ -2549,13 +2550,17 @@ fn NewPrinter(
...
-                    const isOptionalChain = (e.optional_chain orelse js_ast.OptionalChain.ccontinue) == js_ast.OptionalChain.start;
+                    const isOptionalChain = e.optional_chain == .start;
 
                     var wrap = false;
                     if (e.optional_chain == null) {
                         flags.insert(.has_non_optional_chain_parent);
+
+                        // Inline cross-module TypeScript enum references here
+                        if (p.tryToGetImportedEnumValue(e.target, e.name)) |inlined| {
+                            p.printInlinedEnum(inlined, e.name, level);
```
dylan-conway, line 2561: does this callsite need `p.addSourceMapping`?

src/js_parser.zig
```diff
@@ -23189,3 +23998,18 @@ const CommonJSWrapper = union(enum) {
...
+    // Special-case non-finite numbers
+    if (!std.math.isFinite(f))
+        return 0;
+
+    const uint: u32 = @intFromFloat(@mod(@abs(f), 4294967296));
+    const int: i32 = @bitCast(uint);
+    return if (f < 0) -int else int;
+}
+
+fn floatToUInt32(f: f64) u32 {
```
dylan-conway, line 24013: delete

src/js_parser.zig
```diff
@@ -23189,3 +23998,18 @@ const CommonJSWrapper = union(enum) {
...
+
+/// Equivalent of esbuild's js_ast_helpers.ToInt32
+fn floatToInt32(f: f64) i32 {
+    // Special-case non-finite numbers
+    if (!std.math.isFinite(f))
+        return 0;
+
+    const uint: u32 = @intFromFloat(@mod(@abs(f), 4294967296));
+    const int: i32 = @bitCast(uint);
+    return if (f < 0) @as(i32, 0) -% int else int;
```
dylan-conway, line 24010: if `int` is max and `f` is negative, go will actually return `min(i32) - 1` instead of wrapping

src/js_parser.zig
```diff
@@ -23189,3 +23998,18 @@ const CommonJSWrapper = union(enum) {
...
 };
+
+/// Equivalent of esbuild's js_ast_helpers.ToInt32
+fn floatToInt32(f: f64) i32 {
+    // Special-case non-finite numbers
+    if (!std.math.isFinite(f))
+        return 0;
+
+    const uint: u32 = @intFromFloat(@mod(@abs(f), 4294967296));
+    const int: i32 = @bitCast(uint);
```
dylan-conway, line 24009:
```suggestion
    const uint: u32 = @intFromFloat(@mod(@abs(f), std.math.maxInt(u32) + 1));
```

src/js_parser.zig
paperclover: resolved

src/js_parser.zig
```diff
@@ -5552,29 +5852,50 @@ fn NewParser_(
...
+                                if (!gop.found_existing) {
+                                    const ref = try p.newSymbol(.other, name);
+                                    gop.value_ptr.* = ref;
+                                    p.symbols.items[ref.inner_index].namespace_alias = .{
+                                        .namespace_ref = ts_namespace.arg_ref,
+                                        .alias = name,
+                                    };
+                                    declare_loc = member.loc;
+                                    break :brk ref;
+                                }
```
dylan-conway:
```suggestion
                                    };
                                }
                                declare_loc = member.loc;
                                break :brk gop.value_ptr.*;
```

src/js_parser.zig
```diff
@@ -5733,18 +6054,51 @@ fn NewParser_(
...
+                                            .name = ns_alias.alias,
+                                            .name_loc = loc,
+                                        }, loc);
+                                    },
+
+                                    else => {},
+                                }
+                            }
+                        }
+                    }
```
dylan-conway, line 6099: esbuild will return if `namespace_alias != null`, we probably need a default case here

paperdave:
two bugs remain
- utf16 identifiers
- template rope thing

src/js_parser.zig
```diff
@@ -6819,7 +7200,7 @@ fn NewParser_(
                 }
             }
 
-            if (comptime !Environment.isRelease) {
+            if (comptime Environment.allow_assert) {
```
Jarred-Sumner:
```suggestion
            if (comptime Environment.isDebug) {
```

src/js_parser.zig
```diff
@@ -8286,11 +8650,7 @@ fn NewParser_(
                             else => {
                                 if (comptime get_metadata) {
                                     const find_result = p.findSymbol(logger.Loc.Empty, p.lexer.identifier) catch unreachable;
-                                    if (p.known_enum_values.contains(find_result.ref)) {
```
Jarred-Sumner, line 8289: What was this change?

src/js_parser.zig
```diff
@@ -10601,6 +10970,74 @@ fn NewParser_(
...
+                                bun.assert(member_data == .namespace);
+                                try exported_members.put(
+                                    p.allocator,
+                                    p.symbols.items[ns.name.ref.?.inner_index].original_name,
+                                    .{
+                                        .data = member_data,
+                                        .loc = ns.name.loc,
+                                    },
+                                );
+                                // try p.ref_to_ts_namespace_member.put(
```
Jarred-Sumner, line 11019: Was this deliberately commented out?

src/js_parser.zig
paperclover: known_enum_values was removed in favor of real enum inlining.

src/js_parser.zig
paperclover: this second put is seen in esbuild's code but i cannot think of a situation where the value is not already in the map

PR #12164, opened 2024-06-26
24x faster Bun.hash.crc32?

Jarred-Sumner:
### What does this PR do?

After:
```zig
crc32         1.18 µs/iter     (1.16 µs … 1.33 µs)   1.17 µs   1.33 µs   1.33 µs
```

Before:
```zig
crc32        23.89 µs/iter    (23.63 µs … 52.5 µs)  23.75 µs  31.08 µs  32.92 µs
```

Tests


Jarred-Sumner:
This has a test failure and it's not worth fixing right now

But maybe we'll revisit later

PR #12189, opened 2024-06-27
Nest test results under describe scopes

### LINKED ISSUE #3342: Nest `test`/`it` calls under a single `describe` label instead of repeating the `describe` label for each test

`bun test` currently repeats the `describe` label for each `test`/`it` call. Take the following test file as an example:

```ts
describe("test group 1", () => {
  it("does thing 0", () => { expect(0).toBe(0); });
  it("does thing 1", () => { expect(1).toBe(1); });
  it("does thing 2", () => { expect(2).toBe(2); });
  it("does thing 3", () => { expect(3).toBe(3); });
  it("does thing 4", () => { expect(4).toBe(4); });
});

describe("test group 2", () => {
  it("does thing 5", () => { expect(5).toBe(5); });
  it("does thing 6", () => { expect(6).toBe(6); });
  it("does thing 7", () => { expect(7).toBe(7); });
  it("does thing 8", () => { expect(8).toBe(8); });
  it("does thing 9", () => { expect(9).toBe(9); });
});
```

This produces the following output:

image

I would prefer a more concise output like Jest, where the `describe` label is only printed once and the `test`/`it` labels are nested beneath it. I believe this is not just cleaner, but also clearer visually.

Here is Jest's output for the same test file:

image

_No response_

jakeboone02:
### What does this PR do?

Updates the output of `bun test` so that `describe` scope names are only printed once and `test`/`it` results are nested and indented beneath their respective scopes.

- Fixes #3342

- [x] Code changes

There were no existing automated tests for the output format, so manual inspection has been the only test so far. I can write some tests if necessary.

### Comparison of results

```ts

describe("our first test group", () => {
  describe("sub group 1", () => {
    it("knows zero is zero", () => { expect(0).toBe(0); });
    it("1 is the loneliest number", () => { expect(1).toBe(1); });
  });
  describe("nested group 2", () => {
    it("likes the number 2", () => { expect(2).toBe(2); });
    it.todo("thing 3 is still to do", () => { expect(3).toBe(0); });
  });
  it("here's thing 4", () => { expect(4).toBe(4); });
});

describe("some more tests", () => {
  it("does thing cinco", () => { expect(5).toBe(5); });
  it.skip("skips thing 6", () => { expect(6).toBe(6); });
  it("does thing {# days of the week}", () => { expect(7).toBe(7); });
});
```

image


mangs:
@jakeboone02 This is so nice! Great job! One thing that I noticed while looking this over was the "new" example shows as being over 6x slower than before. Is that repeatable?

jakeboone02:
> @jakeboone02 This is so nice! Great job! One thing that I noticed while looking this over was the "new" example shows as being over 6x slower than before. Is that repeatable/consistent?

Thanks! I think the speed difference is just because the new one is a debug build. I'll try a release build later tonight.

jakeboone02:
@mangs I tried the release build from this PR with `bunx bun-pr 12189` and the performance was basically equivalent (a tiny bit faster for some reason??):

```
$ hyperfine --warmup 50 'bun test ./bun-debug-test.test.ts' 'bun-12189 test ./bun-debug-test.test.ts'
Benchmark 1: bun test ./bun-debug-test.test.ts
  Time (mean ± σ):      12.5 ms ±   2.2 ms    [User: 8.3 ms, System: 8.3 ms]
  Range (min … max):     9.9 ms …  36.4 ms    234 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: bun-12189 test ./bun-debug-test.test.ts
  Time (mean ± σ):      11.8 ms ±   1.6 ms    [User: 7.5 ms, System: 8.5 ms]
  Range (min … max):     9.5 ms …  23.4 ms    209 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  'bun-12189 test ./bun-debug-test.test.ts' ran
    1.06 ± 0.23 times faster than 'bun test ./bun-debug-test.test.ts'
```

@Jarred-Sumner I have a feeling this speed difference is insignificant and/or inconsistent, but I'd love to know if I accidentally made it faster somehow.

jakeboone02 (commented):
I added commit c2c667a70859d38d9186d20c644ae714fa50b2e3 that better aligns the message for when a `.todo` test passes and the `--todo` flag has been set. I didn't really know what I was doing so I copied similar code from here:

https://github.com/oven-sh/bun/blob/d356e27a4d602f9614b9c6e614488a1dfa16e41f/src/cli/test_command.zig#L113-L122

That seems like a _lot_ of code just to get the number of items in a linked list, so please let me know if there's a more efficient way.

src/bun.js/test/jest.zig
```diff
@@ -1539,6 +1540,15 @@ pub const TestRunnerTask = struct {
...
             .fail_because_todo_passed => |count| {
+                var scopes_stack = std.BoundedArray(*DescribeScope, 64).init(0) catch unreachable;
+                var parent_: ?*DescribeScope = describe;
+                while (parent_) |scope| {
+                    scopes_stack.append(scope) catch break;
+                    parent_ = scope.parent;
+                }
+                const scopes: []*DescribeScope = scopes_stack.slice();
+                for (1..scopes.len) |_|
+                    Output.prettyError("  ", .{});
```
dylan-conway: I think you can remove the `BoundedArray` and just go up through the parents, skipping the first, something like:
```suggestion
                var parent_: ?*DescribeScope = describe;
                while (parent_) |scope| {
                    if (describe != scope) {
                      Output.prettyError("  ", .{});
                    }
                    parent_ = scope.parent;
                }
```

dylan-conway (commented):
Nice, I like this.

Doesn't need to happen in this pr, but I think printing a status indicator for describe scopes would also be an improvement. Vitest does this, it makes the scopes look a little less out of place

Screenshot 2024-06-28 at 2 01 32 PM

jest for comparison:

Screenshot 2024-06-28 at 2 11 16 PM

jakeboone02:
Yeah, I had the same thought. Needs statusing at the file level, too. I think `bun test` kind of prints the test results as it goes through them, so it might require some refactoring to get the overall status of each file/describe and then print everything.

Other than that, is this ready after I commit that suggestion about the `BoundedArray`?

dylan-conway:
> Yeah, I had the same thought. Needs statusing at the file level, too. I think `bun test` kind of prints the test results as it goes through them, so it might require some refactoring to get the overall status of each file/describe and then print everything.

Yeah, we will probably need to store the results and print either after a describe scope finishes or maybe after the file finishes.

> Other than that, is this ready after I commit that suggestion about the `BoundedArray`?

Yep I'd say it's good to go after that

dylan-conway (approved):
Thanks!

dylan-conway:
We will be reverting this pr before releasing v1.1.19. It's a step in the right direction, but we want to spend more time refining the output. Currently there is too little distinction between file name and describe scope name

jakeboone02:
@dylan-conway I noticed that too, which is one reason I wanted to work on #12378. I'd still like to work on that and get the aesthetics right.

Is that something you guys want to do in-house, or could I maybe propose a design? Probably a mixture of the Vitest and Jest output styles.

PR #12201, opened 2024-06-27
Fixes #12182 - update default port when server is created

### LINKED ISSUE #12182: http.Server.listen(undefined) uses 3000

1.1.12

Darwin 23.1.0 x86_64 i386

```js
import {createServer} from 'http';
let h = createServer(() => {});
h.listen(undefined, () => {
	console.log(h.address());
});
```

random open port

### What do you see instead?

port 3000

### Additional information

_No response_

davidstevens37:
### What does this PR do?

This PR sets the default port to 0, which will allow the OS to assign an available ephemeral port when the server is not instructed to use a specific port resolving 

Fixes #12182 

- [x] Code changes

- [x] I wrote automated tests 
- [x] I ran my tests locally and they pass (`build/bun-debug test test/js/node/http/node-http.test.ts`)

```
✓ node:http > createServer > should use the provided port [2.02ms]
✓ node:http > createServer > should assign a random port when undefined [1.47ms]
```


Jarred-Sumner (changes requested):
We cannot change this default in server.zig (nor should we), as that is a breaking change. If you’d like to change this to match node behavior, then it should be changed in node:http

davidstevens37:
@Jarred-Sumner got it, thanks! Sorry, about that. I wasn't aware that was the default node behavior, just saw an open issue and wanted to support. Love what you're doing here! Just updated it in node:http instead.

Also, any chance you're able to point me in the right direction for #7716?  I don't know a thing about Zig, but could be something fun to dig into with when i've got some downtime.

test/js/node/http/node-http.test.ts
```diff
@@ -158,6 +158,23 @@ describe("node:http", () => {
       server.close();
     });
 
+    it("should use the provided port", async () => {
+      const server = http.createServer(() => {});
+      server.listen(3000);
```
Electroid: To help prevent flaky tests, instead of a hard-coded port, could you import `randomPort()` from `harness`?

test/js/node/http/node-http.test.ts
```diff
@@ -158,6 +158,23 @@ describe("node:http", () => {
...
+      const { port } = server.address();
+      expect(port).toEqual(3000);
+      server.close();
+    });
+
+    it("should assign a random port when undefined", async () => {
+      const server = http.createServer(() => {});
+      server.listen(undefined);
+      const { port } = server.address();
+      expect(port).not.toBe(3000);
```
Electroid: This can be flaky, because a random port could end up being `3000`! Instead, a way to check this is by creating 2 servers, and validating that the ports are not the same.

Electroid (changes requested):
Thanks for submitting a PR! I leave some minor comments on the tests to make them less flaky.

test/js/node/http/node-http.test.ts
davidstevens37: sure thing!

test/js/node/http/node-http.test.ts
davidstevens37: Yeah, makes sense, thanks!

Electroid (approved):
Thanks!

PR #12219, opened 2024-06-27
base64 decode the request body instead of encoding it a second time

mangs:
### What does this PR do?

The Bun lambda layer is correctly identifying when a request body is base64 encoded. However, instead of base64 decoding it, it encodes it a second time. This PR corrects the mistake.

- [x] Code changes

Tested on an AWS environment. This package didn't have any tests before so I just performed manual testing.


mangs:
I just verified this on our AWS staging environment using a real lambda layer. This change appears to work correctly.

PR #12220, opened 2024-06-28
Update fetch.md env var info

metaspartan:
`BUN_CONFIG_MAX_HTTP_REQUESTS=65535` is a required environment variable for NodeJS like functionality of parallel large volume requests. Limited by default to `256`

### What does this PR do?

- [ ] Code changes

Its documentation.

### DIFF
--- a/docs/guides/http/fetch.md
+++ b/docs/guides/http/fetch.md
@@ -22,3 +22,5 @@ const response = await fetch("https://bun.sh/api", {
 
 const body = await response.json();
 ```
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
I think it's worth mentioning this as an environment variable that can be tweaked and explained, but with different copy

metaspartan:
Cool, yea, it just needs to be mentioned or the default changed to behave the same as NodeJS.

PR #12221, opened 2024-06-28
Update http.zig `max_simultaneous_requests` from 256 to 65535 for default NodeJS like request behavior.

metaspartan:
Update http.zig `max_simultaneous_requests` from 256 to 65535 for default NodeJS like request behavior.

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -1763,7 +1763,7 @@ pub const AsyncHTTP = struct {
     signals: Signals = .{},
 
     pub var active_requests_count = std.atomic.Value(usize).init(0);
-    pub var max_simultaneous_requests = std.atomic.Value(usize).init(256);
+    pub var max_simultaneous_requests = std.atomic.Value(usize).init(65535);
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
I'm not sure this is the right default for `fetch`. Your operating system has an upper limit of something like 16,000 - 64,000 maximum open sockets unless you manually configure linux to behave otherwise.

If any of the requests are short-lived, it also prevents HTTP Keep Alive from kicking in

`https://example.com`:

| Elapsed | Limit |
| ----   | ----- | 
| 5.27s | 9999 |
| 2.26s | 256 (default) |

`http://example.com`:

| Elapsed | Limit |
| ----   | ----- | 
| 5.20s | 9999 |
| 2.26s | 256 (default) |

```ts
const p = []
  p.push(fetch('https://example.com/').then(re => console.count("Completed")));
 }
 await Promise.all(p);
```

PR #12238, opened 2024-06-28
Add `test.coveragePathIgnorePatterns` option

### LINKED ISSUE #4021: Add `bun:test` option similar to Jest's `coveragePathIgnorePatterns`

Some files are executed during tests but are irrelevant when it comes to coverage reporting and percentage calculations. `bun:test` needs a way to execute the files while ignoring them for the purposes of coverage.

A configuration option similar (or equivalent) to Jest's `coveragePathIgnorePatterns`. [Documentation here](https://jestjs.io/docs/configuration/#coveragepathignorepatterns-arraystring).

_No response_

jakeboone02:
### What does this PR do?

- Fixes #4021.

Adds a `coveragePathIgnorePatterns` option to `bunfig.toml` similar to Jest's option of the same name.

Per https://github.com/oven-sh/bun/issues/4021#issuecomment-1667141114, the value can be a glob pattern or an array of glob patterns. File paths matching any of the globs will be ignored for the purpose of coverage % calculation and reporting.

I do have a couple of questions/concerns:

- If all tested file paths are ignored (e.g., `"**"`) and the coverage reporter is "lcov", the `lcov.info` file is created but left empty. Seems _maybe_ technically correct, but LCOV processing tools might error (`genhtml` definitely errors; I'm not sure about others).
- The glob matcher seems strict to me, at least in this context. For example, `"*.js"` won't match anything but `"**/*.js"` will. Probably not a big deal, but we might consider relaxing it a little. I don't have an opinion either way.

- [x] Code changes

I wrote automated tests.

- [ ] _(I'm pretty sure this is N/A)_ JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed


cuquo:
I've [asked for an option](https://discord.com/channels/876711213126520882/876711213126520885/1247339989168033832) in Bun Test to ignore certain files (like Jest or Vitest) in bun's discord before. I've even [made jokes about that](https://discord.com/channels/876711213126520882/876711213126520885/1248578315510611989). Thanks @jakeboone02, instead of complaining like I did, you fixed it :)

jakeboone02:
> I've [asked for an option](https://discord.com/channels/876711213126520882/876711213126520885/1247339989168033832) in Bun Test to ignore certain files (like Jest or Vitest) in bun's discord before. I've even [made jokes about that](https://discord.com/channels/876711213126520882/876711213126520885/1248578315510611989). Thanks @jakeboone02, instead of complaining like I did, you fixed it :)

To be fair, I did [complain about it before I did anything about it](https://github.com/oven-sh/bun/issues/4021). :)

Now if I could only get those CI tests passing...  I'm not sure what's going on. They pass on my machine!

PR #12242, opened 2024-06-29
Refactor ZigString -> toJS

Jarred-Sumner:
### What does this PR do?

Previously, we had:
- ZigString.toValueGC
- ZigString.toValue
- ZigString.toValueAuto
- ZigString.toJS

Each of these were subtly different. Exposing a way to do a static ZigString actually work, it leaks small amounts of memory. The `StaticStringImpl*` would never be re-used. So while the string contents weren't copied, the class holding the StringImpl would never be freed which would cause a permanent and very small memory leak

The other issue is it's very confusing which one to use. The distinction wasn't very important and we already check each case anyway.


paperclover (approved):
approved but the number of changed files really says something (considering bun.String)

PR #12278, opened 2024-07-01
fix: add Symbol.toStringTag to KeyObject instances

panva:
### What does this PR do?

As per https://github.com/nodejs/node/pull/46043, this adds Symbol.toStringTag getter to KeyObject.


Jarred-Sumner:
I think these are all flaky tests

PR #12289, opened 2024-07-02
Refactor color formatter

Jarred-Sumner:
### What does this PR do?

Instead of parsing every color format string twice (once with colors enabled and once without), we parse it once and store the version with the colors enabled and with the colors disabled in the same buffer

I was hoping this would make the build faster, but the impact is less than a second and therefore probably not meaningful

Manual

### DIFF
--- a/src/output.zig
+++ b/src/output.zig
@@ -700,86 +700,142 @@ pub const color_map = ComptimeStringMap(string, .{
     &.{ "yellow", ED ++ "33m" },
 });
 const RESET: string = "\x1b[0m";
+const ColorFormatter = struct {
+    enabled: []const u8,
+    disabled: []const u8,
+
+    pub fn parse(comptime fmt: string) *const ColorFormatter {
+        return comptime value: {
+            var fmts: [2][fmt.len * 4]u8 = undefined;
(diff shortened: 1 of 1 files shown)


src/output.zig
```diff
@@ -700,86 +700,142 @@ pub const color_map = ComptimeStringMap(string, .{
...
-                if (i < fmt.len) {
-                    switch (fmt[i]) {
-                        '<', '>' => {
-                            new_fmt[new_fmt_i] = fmt[i];
-                            new_fmt_i += 1;
+const ColorFormatter = struct {
+    enabled: []const u8,
+    disabled: []const u8,
+
+    pub fn parse(comptime fmt: string) *const ColorFormatter {
```
paperclover, line 707: this does not need to return a pointer

paperclover (commented):
i shouldve brought up that i tried this before. or rather, i tried disabling the output formatter entirely.

the project build time is similar to status quo if you make prettyFmt `return fmt;`, doing no work.

so i don't think this change is worth making, especially considering that the code in `ColorFormatter` is still quite a messy parser.

Issue #12322, opened 2024-07-03
Missing `resolveObjectURL` method in `node:buffer` module


zAlweNy26:
### What version of Bun is running?

1.1.17

### What platform is your computer?

Linux 6.5.0-1024-oem x86_64 x86_64

### What steps can reproduce the bug?

I tried this:
```ts
import { resolveObjectURL } from 'node:buffer'
// ...
const blob = resolveObjectURL('blob:xxx') // Error: Method not implemented
```

### What is the expected behavior?

It would be nice to have that method implemented as well, because the other methods usually used in combine with it (`URL.createObjectURL` and `URL.revokeObjectURL`) are already implemented in Bun.

### What do you see instead?

I get the error `Method not implemented` when using that method in my code, and as I can see here:
https://github.com/oven-sh/bun/blob/5a0b93523134ca0043ee7ae84f01003aa4b7ff69/src/bun.js/modules/NodeBufferModule.h#L195

PR #12324, opened 2024-07-03
Implement `buffer.resolveObjectURL`

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/12322

Links:
- https://github.com/nodejs/node/blob/2eff28fb7a93d3f672f80b582f664a7c701569fb/lib/internal/blob.js#L441
- https://nodejs.org/api/buffer.html#bufferresolveobjecturlid
- https://github.com/nodejs/node/blob/2eff28fb7a93d3f672f80b582f664a7c701569fb/test/parallel/test-blob-createobjecturl.js#L35

Copied some tests from Node

### DIFF
--- a/src/bun.js/modules/NodeBufferModule.h
+++ b/src/bun.js/modules/NodeBufferModule.h
@@ -1,3 +1,7 @@
+#pragma once
+
+#include "root.h"
+
 #include "../bindings/JSBuffer.h"
 #include "_NativeModule.h"
 #include "simdutf.h"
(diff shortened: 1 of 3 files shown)


src/bun.js/webcore/ObjectURLRegistry.zig
```diff
@@ -126,9 +132,39 @@ export fn Bun__revokeObjectURL(globalObject: *JSC.JSGlobalObject, callframe: *JS
...
 }
 
+export fn jsFunctionResolveObjectURL(globalObject: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(JSC.conv) JSC.JSValue {
+    const arguments = callframe.arguments(1);
+
+    // Errors are ignored.
+    // Not thrown.
+    // https://github.com/nodejs/node/blob/2eff28fb7a93d3f672f80b582f664a7c701569fb/lib/internal/blob.js#L441
+    if (arguments.len < 1) {
+        return JSC.JSValue.undefined;
```
paperclover, line 142: reminder that these can be written as just return .undefined;

src/bun.js/modules/NodeBufferModule.h
```diff
@@ -1,3 +1,7 @@
+#pragma once
```
paperclover, line 1: i think these files were used in a weird way where this root.h isnt needed but i think this is reasonable to have nonetheless since it compiles

paperclover (approved):
looks nice. my nitpick is not blocking

Issue #9271, opened 2024-03-06
performance.timerify is not a function


maxuai:
### What version of Bun is running?

1.0.30

### What platform is your computer?

Linux 6.5.0-21-generic x86_64 x86_64

### What steps can reproduce the bug?

I am not able to use the `timerify` function from node.js perf_hooks

https://nodejs.org/api/perf_hooks.html#performancetimerifyfn-options

content of index.js file:
```
import { performance } from 'node:perf_hooks';

const fn = () => {
  console.log('works');
};
export const timerifyFn = performance.timerify(fn);
```

### What is the expected behavior?

`timerify` should be supported

### What do you see instead?

when running the app via `bun index.js` I get the following error message:

`TypeError: performance.timerify is not a function. (In 'performance.timerify(fn)', 'performance.timerify' is undefined)`

brettearle:
G'day can I give this a shot?

erik-dunteman:
I'm now working on this too, @maxuai are you accessing performance via the histogram option
```javascript
const { performance, createHistogram } = require("perf_hooks");
function someFunction() {
  console.log('hello world');
}
let histogram = createHistogram();
const wrapped = performance.timerify(someFunction, { histogram: histogram });
wrapped();
console.log(h);
```

or via subscribing a PerformanceObserver to functions entryTypes?

```javascript
const {
  performance,
  PerformanceObserver,
} = require('node:perf_hooks');

function someFunction() {
  console.log('hello world');
}

const wrapped = performance.timerify(someFunction);

const obs = new PerformanceObserver((list) => {
  console.log(list.getEntries()[0].duration);

  performance.clearMarks();
  performance.clearMeasures();
  obs.disconnect();
});
obs.observe({ entryTypes: ['function'] });

// A performance timeline entry will be created
wrapped();
```

maxuai:
> I'm now working on this too, @maxuai are you accessing performance via the histogram option
> 
> ```js
> const { performance, createHistogram } = require("perf_hooks");
> function someFunction() {
>   console.log('hello world');
> }
> let histogram = createHistogram();
> const wrapped = performance.timerify(someFunction, { histogram: histogram });
> wrapped();
> console.log(histogram);
> ```
> 
> or via subscribing a PerformanceObserver to functions entryTypes?
> 
> ```js
> const {
>   performance,
>   PerformanceObserver,
> } = require('node:perf_hooks');
> 
> function someFunction() {
>   console.log('hello world');
> }
> 
[...]

PR #12334, opened 2024-07-04
Erik/timerify

erik-dunteman:
### What does this PR do?
- Adds performance.timerify() https://nodejs.org/api/perf_hooks.html#performancetimerifyfn-options to partially address #9271 
- Adds HDR Histogram algorithm for use by users in timerify.

Note that timerify records durations in two ways, through a Histogram and through sending events to PerformanceObserver, and this PR only addresses Histogram use. 

The specific issue #9271 should remain open until PerformanceObserver is implemented.

- [x] Code changes

Histogram: added plenty of tests in `hdr_histogram.zig`, all pass
Timerify: local sanity checks, below:
```javascript
const { performance, createHistogram } = require("perf_hooks");

const fn = () => {
  console.log("hello world");
};

// test manual recording
let h = createHistogram();
h.record(100, 1);
h.record(200, 100);
h.record(1000, 1000);
h.record(2000, 1000);
h.record(3000, 1000);
h.record(5000, 1000);
h.record(1000000, 1);

// test add
let otherH = createHistogram();
otherH.record(1, 300);
h.add(otherH);

// test reset
h.reset();

// test timerify
let wrapped = performance.timerify(fn, { histogram: h });
  wrapped();
}
console.log(h);
```
all behave as expected

"I wrote automated tests", though curious if all zig tests are ran in CI. we'll see.

Zig changes

Types
- [ ] I added TypeScript types for the new methods, getters, or setters -> matches node type interface


src/bun.js/bindings/JSMapObject.cpp
```diff
@@ -0,0 +1,29 @@
...
+extern "C" JSC::EncodedJSValue Bun__createMapFromDoubleUint64TupleArray(Zig::GlobalObject* globalObject, const double* doubles, size_t length)
+{
+    JSC::JSMap* map
+        = JSC::JSMap::create(globalObject->vm(), globalObject->mapStructure());
+
+    for (size_t i = 0; i < length; i += 2) {
+        // we passed doubles in from Zig, with this doubles.appendSlice(&.{ percentile, @bitCast(val) });
+        // where percentile is a f64 and val is a u64
+        uint64_t value_as_u64;
+        std::memcpy(&value_as_u64, &doubles[i + 1], sizeof(double)); // cast double to u64
```
paperclover: this is really suspicious. is there a reason we cannot just use a c++ cast for this, instead of `memcpy`.

src/js/builtins/Performance.ts
```diff
@@ -0,0 +1,22 @@
+export function timerify(fn: Function, options) {
```
paperclover: is this dead code?

src/js/node/perf_hooks.ts
```diff
@@ -159,7 +162,8 @@ export default {
     throwNotImplemented("perf_hooks.monitorEventLoopDelay");
   },
   createHistogram() {
-    throwNotImplemented("perf_hooks.createHistogram");
+    const { createHistogram } = $zig("node_perf_hooks_histogram_binding.zig", "createPerfHooksHistogramBinding");
```
paperclover: make this lazy load once, as `$zig` does not provide any cache

```
let createHistogram;

function etc() {
  createHistogram ??= $zig("node_perf_hooks_histogram_binding.zig", "createPerfHooksHistogramBinding");
  return createHistogram();
}
```

An alternative is it may make sense to just hoist this $zig call upwards, and pass the native function directly to the user instead of through a wrapper function that does nothing.

src/bun.js/node/node_perf_hooks_histogram_binding.zig
```diff
@@ -0,0 +1,234 @@
...
+    pub usingnamespace JSC.Codegen.JSRecordableHistogram;
+    hdrHist: HDRHistogram = undefined,
+
+    // RecordableHistogram specific internals
+    delta_start: ?bun.timespec = null,
+
+    const This = @This();
+    const PropertyGetter = fn (this: *This, globalThis: *JSC.JSGlobalObject) callconv(.C) JSC.JSValue;
+
+    pub const min_fn = struct {
```
paperclover: im confused what this wrapper struct is for, the `@as()` calls with PropertyGetter. You can write these as simple functions

```
pub fn min(this: *This, globalThis: *JSC.JSGlobalObject) callconv(.C) JSValue {
  ...
}
const minBigInt = getterAsFn(min);
```

In the future, `const a = @as(T, x);` is universally better written as `const a: T = x;`

src/bun.js/node/node_perf_hooks_histogram_binding.zig
```diff
@@ -0,0 +1,234 @@
...
+            // first get 100th percentile, and loop 0, 50, 75, 82.5, ... until we find the highest percentile
+            const maxPercentileValue = this.hdrHist.value_at_percentile(100) orelse return .undefined;
+            var percent: f64 = 0;
+            var stack_allocator = std.heap.stackFallback(4096, bun.default_allocator);
+            var doubles = std.ArrayList(f64).init(stack_allocator.get());
+            defer doubles.deinit();
+
+            while (true) {
+                if (this.hdrHist.value_at_percentile(percent)) |val| {
+                    doubles.appendSlice(&.{ percent, @bitCast(val) }) catch |err| {
```
paperclover: this cast, and the further handling of memory of everything as a `f64` even though that is not the actual type of the data will make this code incredibly hard to debug in the future.

there is a tool for what you are attempting to use, `extern struct`s. these are c abi compatible.

Consider an approach like:

```zig
// zig
const DoubleToIntMapKV = extern struct {
  key: f64,
  value: u64,
};
```

```c++
struct DoubleToIntMapKV {
  double key;
  uint64_t value;
}
```

You can then use `std.ArrayList(DoubleToIntMapKV)`, and it will be much more clear what is what. `append(.{ .key = 1, .value = 2 })`

Then, write a small wrapper around `Bun__createMapFromDoubleUint64TupleArray` so that we can use a slice instead of a pointer and length. We can put this in bindings.zig, though i doubt that this hyper-specific function is going to be used much; it is a good practice to keep bindings out of these sorts of functions.

```zig
extern fn Bun__createMapFromDoubleUint64TupleArray(*JSGlobalObject, [*] DoubleToIntMapKV, usize) JSValue;

pub fn createMapFromDoubleUint64TupleArray(global: *JSGlobalObject, kvs: []DoubleToIntMapKV) JSValue {
    return Bun__createMapFromDoubleUint64TupleArray(global, kvs.ptr, kvs.len);
}
```

src/bun.js/node/node_perf_hooks_histogram_binding.zig
```diff
@@ -0,0 +1,234 @@
...
+                        globalObject.throwError(err, "failed to append to array");
```
paperclover: use throwOutOfMemory() / bun.outOfMemory(). it will make the error message consistant with the other times we remember to use these functions.

I wish this was easier to deal with.

src/bun.js/node/hdr_histogram.zig
```diff
@@ -0,0 +1,503 @@
...
+    significant_figures: u64,
+    sub_bucket_count: u64,
+    sub_bucket_half_count: u64,
+    sub_bucket_half_count_magnitude: u6,
+    unit_magnitude: u8,
+    sub_bucket_mask: u64,
+    bucket_count: u64,
+    counts: []u64,
+
+    const This = @This();
```
paperclover, line 37: nitpick: this `This` is kind of silly to have. you can just refer to `HDRHistogram`, and in init you can use `.{` instead of `This{`

src/bun.js/node/hdr_histogram.zig
```diff
@@ -0,0 +1,503 @@
...
+            .significant_figures = options.significant_figures,
+            .sub_bucket_count = sub_bucket_count,
+            .sub_bucket_half_count = sub_bucket_half_count,
+            .sub_bucket_half_count_magnitude = sub_bucket_half_count_magnitude,
+            .unit_magnitude = unit_magnitude,
+            .sub_bucket_mask = sub_bucket_mask,
+            .bucket_count = bucket_count,
+            .counts = counts,
+            .total_count = 0,
+            .min = 9223372036854776000,
```
paperclover: should this start as `highest_trackable_value`? or `std.math.maxInt(u64)`?

src/bun.js/node/hdr_histogram.zig
```diff
@@ -0,0 +1,503 @@
...
+            if (running_total >= target_count) {
+                // we found the index that corresponds to the percentile
+                return self.value_from_index(index);
+            }
+        }
+        return null;
+    }
+};
+
+test "value_at_percentile" {
```
paperclover, line 252: i enjoy seeing thorough testing. thanks for adding them.

i wish we had a mechanism for running Zig based tests in our codebase.

src/bun.js/node/node_perf_hooks_histogram_binding.zig
```diff
@@ -0,0 +1,234 @@
...
+const bun = @import("root").bun;
+const HDRHistogram = @import("hdr_histogram.zig").HDRHistogram;
+const meta = bun.meta;
+const JSC = bun.JSC;
+const JSValue = JSC.JSValue;
+
+// Wrapper around HRD Histogram
+pub const RecordableHistogram = struct {
+    pub usingnamespace JSC.Codegen.JSRecordableHistogram;
+    hdrHist: HDRHistogram = undefined,
```
paperclover: `undefined` as the default value in a field is almost always a footgun. i suggest removing the default, and then in the `bun.new` call, you can pass the histogram there.

paperclover (changes requested):
overall alright. most of my worry is in the messy binding for creating the JSMap, and then a handful of simple to fix zig nitpicks. nice work on HDRHistogram.

also if youre writing documentation comments in zig code, that can use `///`, [ref](https://ziglang.org/documentation/0.13.0/#Comments)

src/bun.js/node/node_perf_hooks_histogram_binding.zig
Jarred-Sumner: This will cause the alignment to go up, potentially doubling memory usage

That's why I suggested a single flat array of tuples where the first element is a double and the second element is the uint64

README.md
```diff
@@ -1,382 +1,3 @@
-<p align="center">
-  <a href="https://bun.sh"><img src="https://user-images.githubusercontent.com/709451/182802334-d9c42afe-f35d-4a7b-86ea-9985f73f20c3.png" alt="Logo" height=170></a>
-</p>
-<h1 align="center">Bun</h1>
+# HdrHistogram_c
```
paperclover: lol the readme got nuked. please revert

src/bun.js/node/node_perf_hooks_histogram_binding.zig
paperclover: This is not how [alignment](https://en.wikipedia.org/wiki/Data_structure_alignment) works. and even if it was, this would not double memory usage.

`@sizeOf(DoubleToIntMapKV)` is `16`
`@alignOf(DoubleToIntMapKV)` is `8`

<details>

```zig
const DoubleToIntMapKV = extern struct {
    key: f64,
    value: u64,
};

comptime {
    @compileLog(@sizeOf(DoubleToIntMapKV));
    @compileLog(@alignOf(DoubleToIntMapKV));
}
```

</details>

But even if the alignment was 16, the memory layout of a slice would still have no padding, it just would be have 16 bit alignment.

Jarred-Sumner: edit: i'm wrong

src/bun.js/node/node_perf_hooks_histogram_binding.zig
```diff
@@ -0,0 +1,234 @@
...
+    pub fn reset(this: *This, globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSValue {
+        _ = globalThis;
+        _ = callframe;
+        this.hdrHist.reset();
+        return .undefined;
+    }
+
+    pub fn add(this: *This, globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callconv(.C) JSValue {
+        const args = callframe.arguments(1).slice();
+        if (args.len != 1) {
```
Jarred-Sumner: Generally, in JS you ignore it when there are too many arguments

```suggestion
        if (args.len < 1) {
```

src/bun.js/node/node_perf_hooks_histogram_binding.zig
erik-dunteman: you're completely right, I went down an idea maze on setter/getter abstraction and just ended up back at writing the getter functions outright but forgot to remove the PropertyGetter madness

src/bun.js/node/node_perf_hooks_histogram_binding.zig
erik-dunteman: i'll do extern structs then!

src/bun.js/node/hdr_histogram.zig
erik-dunteman: std.math.maxInt is probably better choice, whole point is to guarantee it gets overridden on next record

src/js/builtins/Performance.ts
erik-dunteman: no, that's the timerify half of this pr

src/bun.js/bindings/JSMapObject.cpp
erik-dunteman: moved to extern struct, pushing changes shortly

src/bun.js/node/node_perf_hooks_histogram_binding.zig
erik-dunteman: fixed in upcoming push

src/bun.js/node/node_perf_hooks_histogram_binding.zig
erik-dunteman: done

src/bun.js/node/hdr_histogram.zig
erik-dunteman: TBH did this because unit testing zig test `just_my_file.zig` is so much faster than doing a full build and integration test

And I'm happy to. Tests good.

src/bun.js/node/node_perf_hooks_histogram_binding.zig
erik-dunteman: fixed in upcoming push

erik-dunteman:
Still need to
- rebase to current main
- look at Callconv stuff
- implement bigInt getters

erik-dunteman:
Working on PerformanceObserver stuff now, but this PR should be ready to go whenever CI passes and code review approved @Jarred-Sumner @paperdave

Jarred-Sumner:
> test/js/node/perf_hooks/perf_hooks.test.ts STATUS_INTEGER_DIVIDE_BY_ZERO
> test/js/bun/util/fuzzy-wuzzy.test.ts STATUS_SEVERITY_ERROR

These are new test failures

PR #12350, opened 2024-07-04
Print list of CPU features in crash reports

Jarred-Sumner:
### What does this PR do?

Print list of CPU features in crash reports

image


src/bun.js/bindings/CPUFeatures.zig
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        pub fn get() T {
+            return @bitCast(bun_cpu_features());
+        }
+    };
+}
+
+const X86CPUFeatures = packed struct(u8) {
+    none: bool = false,
```
paperclover, line 43: none should not be a flag here. the numeric values in c++ are off by one (you can start at zero, `1 << 0` is 1). then the padding bit increases by one

src/bun.js/bindings/CPUFeatures.zig
```diff
@@ -0,0 +1,81 @@
...
+else if (bun.Environment.isAarch64)
+    AArch64CPUFeatures
+else
+    struct {
+        pub fn get() @This() {
+            return .{};
+        }
+
+        pub fn format(self: @This(), comptime _: []const u8, _: anytype, writer: anytype) !void {
+            _ = self; // autofix
```
paperclover: no autofix comments. replace all arguments with `_`

paperclover (changes requested):
nitpicks. looks good otherwise.

src/bun.js/bindings/CPUFeatures.zig
Jarred-Sumner: this was deliberate, as a kind of sanity check

src/bun.js/bindings/CPUFeatures.zig
Jarred-Sumner: this should be enforced by linter instead of by human

src/bun.js/bindings/CPUFeatures.zig
paperclover: add `assert(self.none == false and self.padding == 0);` somewhere to document you sanity check

paperclover (changes requested):

paperclover (approved):
thanks

PR #12351, opened 2024-07-05
Use `bun.ComptimeStringMap` instead of `std.StaticStringMap`

Jarred-Sumner:
### What does this PR do?

Use `bun.ComptimeStringMap` instead of `std.StaticStringMap`

Ours is faster


Jarred-Sumner:
@paperdave was it intentional you changed this during the Zig upgrade?

paperclover (changes requested):
> was it intentional you changed this during the Zig upgrade?

yes. but you can change it back.

request changes on can you add this to banned.json. if not i will and merge later

PR #12353, opened 2024-07-05
Handle OOM in `btoa`

Jarred-Sumner:
### What does this PR do?

Handle OOM in `btoa`

I don't think this actually happens, but we check nonetheless


paperclover (approved):
this is better than nothing so approved, however `Bun__encoding__toString(` does not seem to handle OOM

Issue #12325, opened 2024-07-03
`.set` method with `File` from `FormData` differs in behavior from Node.js (file name is lost)


kravetsone:
### What version of Bun is running?

1.1.17+bb66bba1b

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

I came across this when [I was uploading files to Telegram using GramIO](https://gramio.dev/files/overview). I am sure that this is reproduced in other cases.

The fact is that when I fill in the `FormData` using the `.set` method, the **filename** is lost when uploading to Telegram, although in `Node.js` everything works as expected...

```ts
async function sendDocument(body) {
    const response = await fetch(
        `https://api.telegram.org/bot${process.env.BOT_TOKEN}/sendDocument`,
        {
            method: "POST",
            body: body
        }
    );
    
    const json = await response.json();
    
    console.log(json);
}

const formDataSet = new FormData();

formDataSet.set("chat_id", "617580375");
formDataSet.set("document", new File(["some text notes"], "text-notes.txt"));

await sendDocument(formDataSet)

const formDataAppend = new FormData();

formDataAppend.append("chat_id", "617580375");
formDataAppend.append("document", new File(["some text notes"], "text-notes.txt"));

await sendDocument(formDataAppend)
```

If you try to run this option through `Node.js`, both options will send a file with the specified name, but if you run it through `Bun`, then there will be only second file with the correct name (where it was `FormData.append` used)

![image](https://github.com/oven-sh/bun/assets/57632712/94c8da49-537f-4f41-8d29-18bab98d3697)

Verbose logs where `.set` used (wrong filename)

```bash
[fetch] > HTTP/1.1 POST https://api.telegram.org/bot[REDACTED]/sendDocument
[fetch] > Content-Type: multipart/form-data; boundary="-WebkitFormBoundary980f0dfe1eed40e094766072fdf6ca05"
[fetch] > Connection: keep-alive
[fetch] > User-Agent: Bun/1.1.17
[fetch] > Accept: */*
[fetch] > Host: api.telegram.org
[fetch] > Accept-Encoding: gzip, deflate, br
[fetch] > Content-Length: 349
```

Verbose logs where `.append` used

```bash
[fetch] > HTTP/1.1 POST https://api.telegram.org/bot[REDACTED]/sendDocument
[...]

PR #12379, opened 2024-07-06
fix(formdata): handle file names correctly when setting on formdata

billywhizz:
### What does this PR do?

fixes https://github.com/oven-sh/bun/issues/12325

- [x] a new test was added and all http tests including new one are passing

### DIFF
--- a/src/bun.js/bindings/webcore/JSDOMFormData.cpp
+++ b/src/bun.js/bindings/webcore/JSDOMFormData.cpp
@@ -462,10 +462,6 @@ static inline JSC::EncodedJSValue jsDOMFormDataPrototypeFunction_set2Body(JSC::J
     RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
     EnsureStillAliveScope argument1 = callFrame->uncheckedArgument(1);
 
-    EnsureStillAliveScope argument2 = callFrame->argument(2);
-    auto filename = argument2.value().isUndefined() ? String() : convert<IDLUSVString>(*lexicalGlobalObject, argument2.value());
-    RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
-
     RefPtr<Blob> blobValue = nullptr;
     if (argument1.value().inherits<JSBlob>()) {
         blobValue = Blob::create(argument1.value());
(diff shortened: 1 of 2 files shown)


src/bun.js/bindings/webcore/JSDOMFormData.cpp
```diff
@@ -462,19 +462,19 @@ static inline JSC::EncodedJSValue jsDOMFormDataPrototypeFunction_set2Body(JSC::J
...
         blobValue = Blob::create(argument1.value());
     }
 
     if (!blobValue) {
         throwTypeError(lexicalGlobalObject, throwScope, "Expected argument to be a Blob."_s);
-        RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
     }
+    RETURN_IF_EXCEPTION(throwScope, encodedJSValue());
+
+    EnsureStillAliveScope argument2 = callFrame->argument(2);
```
Jarred-Sumner: We should keep this line in the original spot it was in because in certain cases the callFrame can be moved which leads to the GC not realizing the value still needs to be kept alive

test/js/bun/http/form-data-set-append.test.js
```diff
@@ -0,0 +1,65 @@
+import { expect, test } from "bun:test";
+
+// https://github.com/oven-sh/bun/issues/12325
+
+test("formdata set with File works as expected", async () => {
+  const expected = ["617580375", "text-notes1.txt"];
+
+  const server = Bun.serve({
```
Jarred-Sumner: nit: we can do `using` and that will automatically close the server without needing to manually call `.stop`
```suggestion
  using server = Bun.serve({
```

test/js/bun/http/form-data-set-append.test.js
```diff
@@ -0,0 +1,65 @@
...
+  formDataSet.set("chat_id", expected[0]);
+  formDataSet.set("document", new File(["some text notes 1"], expected[1]));
+  await sendDocument(formDataSet);
+  server.stop();
+});
+
+test("formdata apppend with File works as expected", async () => {
+  const expected = ["617580376", "text-notes2.txt"];
+
+  const server = Bun.serve({
```
Jarred-Sumner: nit: we can do `using` and that will automatically close the server without needing to manually call `.stop`
```suggestion
  using server = Bun.serve({
```

src/bun.js/bindings/webcore/JSDOMFormData.cpp
billywhizz: ah cool. i wasn't sure about that. will update.

test/js/bun/http/form-data-set-append.test.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+    const text = await response.text();
+    return text;
+  }
+
+  const formDataSet = new FormData();
+  formDataSet.append("chat_id", expected[0]);
+  formDataSet.append("document", new File(["some text notes 2"], expected[1]));
+  await sendDocument(formDataSet);
+  server.stop();
```
Jarred-Sumner:
```suggestion

```

test/js/bun/http/form-data-set-append.test.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+    const text = await response.text();
+    return text;
+  }
+
+  const formDataSet = new FormData();
+  formDataSet.set("chat_id", expected[0]);
+  formDataSet.set("document", new File(["some text notes 1"], expected[1]));
+  await sendDocument(formDataSet);
+  server.stop();
```
Jarred-Sumner:
```suggestion

```

Jarred-Sumner (approved):

PR #12380, opened 2024-07-06
use `comptime` in wherever possible

anonrig:
I realized that several environment calls lack `comptime`

### DIFF
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -1042,7 +1042,7 @@ pub const EventLoop = struct {
                     this.virtual_machine.modules.onPoll();
                 },
                 @field(Task.Tag, typeBaseName(@typeName(GetAddrInfoRequestTask))) => {
-                    if (Environment.os == .windows) @panic("This should not be reachable on Windows");
+                    if (comptime Environment.isWindows) @panic("This should not be reachable on Windows");
 
                     var any: *GetAddrInfoRequestTask = task.get(GetAddrInfoRequestTask).?;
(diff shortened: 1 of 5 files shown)


Jarred-Sumner (dismissed):

paperdave:
note that `comptime <constant>` does not actually do anything since the operand is alraedy known at compile time. it is nice to have these sometimes as a sanity check that the value is known, but for the `Environment.<some constant>`. i would rather see all uses of `comptime Environment.<some constant>` have the `comptime` operand dropped.

i do like seeing `bun.Environment` moved to `Environment`. that is very nice. but some files you edited dont have that declared as a top level symbol (need `const Environment = bun.Environment;`)

paperdave:
with a development environment setup, you can run `bun run zig-check-all` in the repository to quickly check all platforms for a successful compilation, instead of digging through the verbose CI logs.

anonrig:
> with a development environment setup, you can run `bun run zig-check-all` in the repository to quickly check all platforms for a successful compilation, instead of digging through the verbose CI logs.

`bun run zig-check-all` just fails for me:

```
thread 3701555 panic: ZigGeneratedClasses.zig does not exist in generated code directory!
```

Jarred-Sumner:
@anonrig that command needs to have a `build` directory successfully created

paperdave:
> bun run zig-check-all just fails for me

oh, sorry. i forgot that it still needs code generators. i've never run that command on a blank setup. once you do one setup it'll let you use that command.

PR #12382, opened 2024-07-06
Check that ccache points to a regular file instead of a non-empty string

Jarred-Sumner:
### What does this PR do?

Check that ccache points to a regular file instead of a non-empty string

This doesn't print hi:
```sh
if [ -f /opt/homebrew/bin/node123 ]; then
  echo 'hi'
fi
```

This does print hi:
```sh
if [ -f /opt/homebrew/bin/node ]; then
  echo 'hi'
fi
```


paperclover (approved):

anonrig (commented):
Nice

Issue #5398, opened 2023-09-14
Transpiler stack overflow when loading `aws-csdk-lib`


soulman-is-good:
### What version of Bun is running?

- 1.0.1
- 1.0.2
- 1.0.3

### What platform is your computer?

- Linux 5.4.254-175.358.amzn2int.x86_64 x86_64 x86_64
- Linux 5.10.192-182.736.amzn2.x86_64 x86_64 x86_64 (last Amazon AL2 EC2 AMI)

NOTE: This bug is NOT reproducible on:
- Linux 4.15.0-20-generic x86_64 x86_64
- Linux 5.4.0-1103-aws x86_64 x86_64 (ami-0d7fd770997502624)
- Linux 5.14.21-150500.53-default x86_64 x86_64
- Linux 5.15.0-58-generic x86_64 x86_64
- Linux 6.2.14-300.fc38.aarch64 aarch64 aarch64

### What steps can reproduce the bug?
```
mkdir test-bun-aws-cdk-lib-bug && \
cd test-bun-aws-cdk-lib-bug && \
bun add aws-cdk-lib constructs && \
echo 'import cdk from "aws-cdk-lib"; cdk;' > index.ts && \
bun run .
```

### What is the expected behavior?

exit with status 0

### What do you see instead?

[1]    30502 segmentation fault  bun run .

### Additional information

Would be great to understand what causes segmentation fault

Electroid:
Thanks for reporting this, we're taking a look.

bitfrost:
I am seeing this as well in github action runner's ubuntu-latest images as well.

soulman-is-good:
Tested on
- 1.0.2
- 1.0.3

Still does not work. 
Disclaimer: I'm not tracking changes in the code base. Just tested and reported.

itsyoboieltr:
this issue still persists

weklund:
Is the root cause known for this?  Have experienced this issue as well.

Electroid:
Could you provide an exact EC2 instance type and AMI for us to try? I tried and it works successfully:
* t3.xlarge / [ami-0830c9faf0efc29ff](https://us-west-1.console.aws.amazon.com/ec2/home?region=us-west-1#ImageDetails:imageId=ami-0830c9faf0efc29ff)
[...]

PR #12355, opened 2024-07-05
Make `simplifyUnusedExpr` non-recursive

paperclover:
### What does this PR do?

Fixes #5398
Fixes #8021

`simplifyUnusedExpr` is called on any expression which is not dead code, but has a value that is unused. For example

```ts
a; // a is unused
a = (b, c); // b is unused
a = [d] ? e : f; // [d] is unused because it truthy. f is dead code
```

It is not safe to make this a naive recursive function because some minifiers will build very large comma expressions, with thousands of parts. Bun represents this as a nested binary expression, and when it attempts to simplify the parts, it stack overflows.

TODO:
- [ ] Explain the method used here (maybe draw a diagram), it is kind of confusing and I will forget how this works in the future.
- [ ] Fix E.If handling, it has a subtle flaw.
- [ ] Performance testing. This is hot code as it is run on every expression statement. I have a feeling ArrayList is faster for this use-case. But I wanted to try SegmentedList because the idea of not re-allocating old items was cool.
- [ ] Tests!

PR #12384, opened 2024-07-06
simpler version of `simplifyUnusedExpr` rewrite

paperclover:
### What does this PR do?

See #12355 for related info. This supersedes that.

Instead of making the whole function recursive, make just the comma expression folding non-recursive. This is much more performant.

Fixes https://github.com/oven-sh/bun/issues/5398
Fixes https://github.com/oven-sh/bun/issues/8021


src/js_parser.zig
```diff
@@ -2106,6 +2103,56 @@ pub const SideEffects = enum(u1) {
         return expr;
     }
 
+    const BinaryExpressionSimplifyVisitor = struct {
+        bin: *E.Binary,
+    };
+
+    ///
```
Jarred-Sumner, line 2110: comment

PR #12393, opened 2024-07-06
Fix memory leak in `withFileTypes: true` in fs.readdir

Jarred-Sumner:
### What does this PR do?

The `path` / `parentPath` string in `fs.DirEnt` was being re-created for every file or folder in the directory, and then leaked. This memory leak started in #11135. Thanks to @billywhizz for uncovering this leak.

A test will be added to #12391 

| Runtime | RSS |
|----------|-----|
| `bun --smol` (this PR) |  37 MB |
| `node` |  59 MB |
| `bun` (this PR) |  99 MB |
| `bun` stable | 250 MB |

```ts
import { readdirSync, promises as fs } from "fs";
const readdir = readdirSync;
// const { readdir } = fs;

await (async function () {
  for (let i = 0; i < 10000; i++) {
    for (let { parentPath, path, name } of await readdir(
      "/Users/jarred/Code/bun/test/js/node",
      { recursive: true, withFileTypes: true }
    )) {
    }
  }
})();

if (globalThis.Bun) {
  Bun.gc(true);
  Bun.unsafe.mimallocDump();
}
console.log("RSS", (process.memoryUsage.rss() / 1024 / 1024) | 0, "MB");
```

| Runtime | RSS |
|----------|-----|
| `bun --smol` (this PR) | 52 MB |
| `node` |  60 MB |
| `bun` (this PR) | 68 MB |
| `bun` stable | 265 MB |

```ts
import { readdirSync, promises as fs } from "fs";
const { readdir } = fs;

await (async function () {
  for (let i = 0; i < 10000; i++) {
    for (let { parentPath, path, name } of await readdir(
      "/Users/jarred/Code/bun/test/js/node",
      { recursive: true, withFileTypes: true }
    )) {
    }
  }
})();

if (globalThis.Bun) {
  Bun.gc(true);
  Bun.unsafe.mimallocDump();
}
console.log("RSS", (process.memoryUsage.rss() / 1024 / 1024) | 0, "MB");
```

Note that there may still be a leak in `fs.promises.readdir` with `recursive: true`. This specifically addresses `withFileTypes`.

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -189,7 +189,7 @@ pub const Async = struct {
                     std.mem.doNotOptimizeAway(&node_fs);
                 }
 
-                this.globalObject.bunVMConcurrently().eventLoop().enqueueTaskConcurrent(JSC.ConcurrentTask.create(JSC.Task.init(this)));
+                this.globalObject.bunVMConcurrently().eventLoop().enqueueTaskConcurrent(JSC.ConcurrentTask.createFrom(this));
             }
 
             pub fn runFromJSThread(this: *Task) void {
@@ -5044,7 +5044,10 @@ pub const NodeFS = struct {
         const dir = fd.asDir();
         const is_u16 = comptime Environment.isWindows and (ExpectedType == bun.String or ExpectedType == Dirent);
 
-        var dirent_path: ?bun.String = null;
+        var dirent_path: bun.String = bun.String.dead;
+        defer {
+            dirent_path.deref();
+        }
 
         var iterator = DirIterator.iterate(dir, comptime if (is_u16) .u16 else .u8);
         var entry = iterator.next();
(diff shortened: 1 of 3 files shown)


src/bun.js/node/node_fs.zig
```diff
@@ -5231,13 +5234,15 @@ pub const NodeFS = struct {
             switch (comptime ExpectedType) {
                 Dirent => {
                     const path_u8 = bun.path.dirname(bun.path.join(&[_]string{ root_basename, name_to_copy }, .auto), .auto);
-                    if (dirent_path_prev == null or bun.strings.eql(dirent_path_prev.?.byteSlice(), path_u8)) {
+                    if (dirent_path_prev.isEmpty() or !bun.strings.eql(dirent_path_prev.byteSlice(), path_u8)) {
```
Jarred-Sumner, line 5237: @nektro the main cause of the leak was here. `bun.strings.eql` should've been `!bun.strings.eql`

src/bun.js/node/node_fs.zig
```diff
@@ -5389,9 +5395,6 @@ pub const NodeFS = struct {
                     else => @compileError("Impossible"),
                 }
             }
-            if (dirent_path_prev) |*p| {
```
Jarred-Sumner, line 5392: @nektro it was also leaking on errors because it never called deref if it failed in the middle of visiting directories.

nektro (approved):

PR #12399, opened 2024-07-06
Fixed some documentation to conform to markdownlint rules

okineadev:
### What does this PR do?

Makes documentation compliant with **markdownlint** rules

- [ ] Code changes

- [x] I belive


paperdave:
we should have a ci check for this so it stays in this condition. all of the changes i've seen i am ok with

okineadev:
> This pull request is stale and may be closed due to inactivity.

This PR really does not cover all files, because I fixed them manually, and not with `markdownlint-cli`

If I have time (definitely not in the next 2 weeks), then I will make a new PR.

Issue #12134, opened 2024-06-25
Writing file to stdout causes Bun to crash unexpectedly.


ejini6969:
### How can we reproduce the crash?

Code that causes failure

```js
// write file incrementally
const file2 = Bun.file('./test.txt');
const writer = file2.writer();
writer.write("A");
writer.write("B");
writer.write("C");
writer.flush(); // flush the buffer
writer.end() // flush buffer and close file
// write Blob to file (transfer content from 1 file to another)
await Bun.write('./test2.txt', Bun.file('./test.txt'));
// write to stdout
await Bun.write(Bun.stdout, file2)
```

### Relevant log output

```shell
panic(main thread): reached unreachable code
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.15/wr1b23ba1fAyggggEqkglC48nMkwhlT6j8+Jwz99J_A1

error: script "js" exited with code 3
```

### Stack Trace (bun.report)

Bun v1.1.15 ([`b23ba1f`](<https://github.com/oven-sh/bun/tree/b23ba1fe1805891cdf4566331f0c8e87d0a5b758>)) on windows x86_64 [AutoCommand]

**panic**: reached unreachable code

- [`mem.zig:197`](<https://github.com/oven-sh/bun/blob/b23ba1fe1805891cdf4566331f0c8e87d0a5b758/src/deps/zig/lib/std/mem.zig#L197>): `copyForwards`
- [`windows.zig:3402`](<https://github.com/oven-sh/bun/blob/b23ba1fe1805891cdf4566331f0c8e87d0a5b758/src/windows.zig#L3402>): `GetFinalPathNameByHandle`
- [`blob.zig:2020`](<https://github.com/oven-sh/bun/blob/b23ba1fe1805891cdf4566331f0c8e87d0a5b758/src/bun.js/webcore/blob.zig#L2020>): `copyfile`
- [`blob.zig:881`](<https://github.com/oven-sh/bun/blob/b23ba1fe1805891cdf4566331f0c8e87d0a5b758/src/bun.js/webcore/blob.zig#L881>): `writeFileWithSourceDestination`
- [`blob.zig:1201`](<https://github.com/oven-sh/bun/blob/b23ba1fe1805891cdf4566331f0c8e87d0a5b758/src/bun.js/webcore/blob.zig#L1201>): `writeFile`
[...]

Issue #11352, opened 2024-05-25
Bunx crash on windows in the v1.1.10


Lan-Hekary:
### How can we reproduce the crash?

I just update to the latest version on windows.
I tested it with `bunx cowsay hello`
It crashed immediately.

### JavaScript/TypeScript code that reproduces the crash?

```shell
bunx cowsay hello
```

### Relevant log output

```shell
Bun v1.1.10 (5102a944) Windows x64
Args: "C:\Users\Lan\scoop\apps\bun\1.1.10\bun.exe" "add" "cowsay@latest" "--no-summary" "--no-cache" "--force"
Elapsed: 5ms | User: 0ms | Sys: 15ms
RSS: 25.37MB | Peak: 25.37MB | Commit: 71.75MB | Faults: 6383

panic(main thread): attempt to unwrap error
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.10/wI15102a94AA86glCwvqpComgqEu56rEu15xFqn4BiiJo3ht+CCa\KERNEL32.DLLo05FCU\ntdll.dllirzUA0eNpLLClJzS0oUSjJVyjNKy9KLFBILSrKLwIAbHQJCg
```

### Stack Trace (bun.report)

Bun v1.1.10 ([`5102a94`](<https://github.com/oven-sh/bun/tree/5102a944307d56933a6bb039b99ad76a758b46e4>)) on windows x86_64 [AddCommand]

**panic**: attempt to unwrap error

- [`mem.zig:197`](<https://github.com/oven-sh/bun/blob/5102a944307d56933a6bb039b99ad76a758b46e4/src/deps/zig/lib/std/mem.zig#L197>): `copyForwards`
- [`windows.zig:3402`](<https://github.com/oven-sh/bun/blob/5102a944307d56933a6bb039b99ad76a758b46e4/src/windows.zig#L3402>): `GetFinalPathNameByHandle`
- [`install.zig:7401`](<https://github.com/oven-sh/bun/blob/5102a944307d56933a6bb039b99ad76a758b46e4/src/install/install.zig#L7401>): `initWithCLI`
- [`install.zig:8528`](<https://github.com/oven-sh/bun/blob/5102a944307d56933a6bb039b99ad76a758b46e4/src/install/install.zig#L8528>): `updatePackageJSONAndInstall`
[...]

Issue #10914, opened 2024-05-08
bun crashed


ShadyShenAli:
### How can we reproduce the crash?

in cmd.exe, execute bun repl  and it crashes (see crash report)

The same thing happens when I execute bun install after bun init -y.

### Relevant log output

```shell
D:\>bun repl
============================================================
Bun v1.1.7 (b0b7db5c) Windows x64
Args: "bun", "bun-repl"
Elapsed: 11ms | User: 0ms | Sys: 0ms
RSS: 20.50MB | Peak: 20.50MB | Commit: 62.50MB | Faults: 5153

panic(main thread): attempt to unwrap error
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.7/wB1b0b7db5AA8wzkCk1mlCqkpqEm05rE44krE27gvFq4kBiiJoiyu2CCa\KERNEL32.DLL63yEA0eNpLLClJzS0oUSjJVyjNKy9KLFBILSrKLwIAbHQJCg

D:\p>dir
 Volume in drive D has no label.
 Volume Serial Number is 56A5-A7D0

 Directory of D:\p

05/08/2024  02:34 PM    <DIR>          .
05/08/2024  02:34 PM             2,233 .gitignore
05/08/2024  02:34 PM                30 index.ts
05/08/2024  02:34 PM               172 package.json
05/08/2024  02:34 PM               215 README.md
05/08/2024  02:34 PM               635 tsconfig.json
               5 File(s)          3,285 bytes
               1 Dir(s)   1,750,319,104 bytes free

D:\p>bun install
============================================================
Bun v1.1.7 (b0b7db5c) Windows x64
Args: "bun", "install"
Elapsed: 6ms | User: 15ms | Sys: 0ms
RSS: 24.11MB | Peak: 24.11MB | Commit: 71.60MB | Faults: 6040

panic(main thread): attempt to unwrap error
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

[...]

PR #12411, opened 2024-07-07
On Windows, support `Bun.stdin`, `Bun.stdout`, `Bun.stderr` in `Bun.write` when the other argument is a file

### LINKED ISSUE #12404: Panic "unreachable code" on Windows when writing from stdin to stdout

```js
await Bun.write(Bun.stdout, Bun.stdin);
```

That's all. From a cursory look at the code, it seems that you're calling GetFinalPathNameByHandle, and I guess stdin/out don't have names so it fails.

### Relevant log output

_No response_

### Stack Trace (bun.report)

Bun v1.1.18 ([`5a0b935`](<https://github.com/oven-sh/bun/tree/5a0b93523134ca0043ee7ae84f01003aa4b7ff69>)) on windows x86_64 [AutoCommand]

**panic**: reached unreachable code

- [`windows.zig:1269`](<https://github.com/oven-sh/bun/blob/5a0b93523134ca0043ee7ae84f01003aa4b7ff69/src/deps/zig/lib/std/os/windows.zig#L1269>): `GetFinalPathNameByHandle`
- [`windows.zig:3401`](<https://github.com/oven-sh/bun/blob/5a0b93523134ca0043ee7ae84f01003aa4b7ff69/src/windows.zig#L3401>): `GetFinalPathNameByHandle`
- [`bindings.zig:372`](<https://github.com/oven-sh/bun/blob/5a0b93523134ca0043ee7ae84f01003aa4b7ff69/src/bun.js/bindings/bindings.zig#L372>): `eqlComptime`
- [`bun.zig:3040`](<https://github.com/oven-sh/bun/blob/5a0b93523134ca0043ee7ae84f01003aa4b7ff69/src/bun.zig#L3040>): `new`
- [`blob.zig:1201`](<https://github.com/oven-sh/bun/blob/5a0b93523134ca0043ee7ae84f01003aa4b7ff69/src/bun.js/webcore/blob.zig#L1201>): `writeFile`
- *2 unknown/js code*
- `llint_entry`

Jarred-Sumner:
### What does this PR do?

The following now work as expected on Windows:

```ts
await Bun.write(Bun.stdout, Bun.file(path.join(import.meta.dir, "hello-world.txt")));
await Bun.write(Bun.stderr, Bun.file(path.join(import.meta.dir, "hello-world.txt")));
```

This also addresses a bug where we would have an assertion failure when passing a `NUL` file path on Windows. In most cases, we will now use the Windows `GetFinalPathByHandleW` function instead of Zig's implementation (https://github.com/ziglang/zig/issues/18756).

Fixes #12404 
Fixes https://github.com/oven-sh/bun/issues/12134
Fixes https://github.com/oven-sh/bun/issues/11352 (most likely)
Fixes https://github.com/oven-sh/bun/issues/10914 (most likely)

The `BUN_FEATURE_FLAG_DISABLE_UV_FS_COPYFILE` environment variable forces the code path which does a read-write loop in `Bun.write`. We re-run the tests for `Bun.write` with this environment variable set on Windows. Also, added a couple more tests.


src/bun.js/webcore/blob.zig
```diff
@@ -1979,6 +1977,204 @@ pub const Blob = struct {
...
+            read_write_loop: ReadWriteLoop = .{},
+
+            pub const ReadWriteLoop = struct {
+                source_fd: bun.FileDescriptor = invalid_fd,
+                must_close_source_fd: bool = false,
+                destination_fd: bun.FileDescriptor = invalid_fd,
+                must_close_destination_fd: bool = false,
+                written: usize = 0,
+                read_buf: std.ArrayList(u8) = std.ArrayList(u8).init(default_allocator),
+                uv_buf: libuv.uv_buf_t = libuv.uv_buf_t{ .base = undefined, .len = 0 },
```
paperclover: reminder that `x: Y = Y{` can always be written as `x: Y = .{`

```suggestion
                uv_buf: libuv.uv_buf_t = .{ .base = undefined, .len = 0 },
```

src/bun.js/webcore/blob.zig
```diff
@@ -1979,6 +1977,204 @@ pub const Blob = struct {
...
+                            this.err = err;
+                            this.onReadWriteLoopComplete();
+                        },
+                        .result => {},
+                    }
+                }
+
+                pub fn close(this: *ReadWriteLoop) void {
+                    if (this.must_close_source_fd) {
+                        if (bun.toLibUVOwnedFD(this.source_fd)) |fd| {
```
paperclover, line 2137: is there any case this is not already a libuv owned file descriptor?

src/bun.js/webcore/blob.zig
```diff
@@ -1979,6 +1977,204 @@ pub const Blob = struct {
...
+                                bun.Async.Loop.get(),
+                            );
+                        } else |_| {
+                            _ = bun.sys.close(this.source_fd);
+                        }
+                        this.must_close_source_fd = false;
+                        this.source_fd = invalid_fd;
+                    }
+
+                    if (this.must_close_destination_fd) {
```
paperclover, line 2149: is there any case this is not already a libuv owned file descriptor?

paperclover (changes requested):

paperclover (approved):

src/bun.js/webcore/blob.zig
Jarred-Sumner: yeah

PR #12412, opened 2024-07-07
Fix argument validation with callbacks in node:fs

Jarred-Sumner:
### What does this PR do?

Four things:
1. Instead of always assuming the final arugment is a callback, we copy the behavior node does for validating a callback
2. Preserve the function names in node:fs for the exports. So it says "write" in the stack trace instead of "write2"
3. Fix a debug-only crash when running `next build` (heap breakdown allocator wasn't used)
4. Apparently, you can pass `fs.truncate` a file descriptor and it's supposed to Just Work and become ftruncate

image

I went through Node's code and made sure the specific behavior matches (in some cases, they do `||=`, in other cases, they allow the 2nd-to-last argument, in other cases, they always require it)


Jarred-Sumner:
I'm pretty sure this is right so i'm going to merge it but @dylan-conway please review when you get a chance

dylan-conway (commented):
LGTM

PR #12445, opened 2024-07-09
heavy revision on heap_breakdown's safety

paperclover:
### What does this PR do?

I wish we had some internal docs on heap breakdown. It is very cool. This PR makes it more safe for us to use.

Rename `bun.HeapBreakdown` to `bun.heap_breakdown`. Rename `bun.is_heap_breakdown_enabled` to `bun.heap_breakdown.enabled`.

To use the heap breakdown allocator functions directly, recommend using `heap_breakdown.getZone(T).create(T, t)`. But it is better to simply use the `bun.New` mixin, which opts you into heap breakdown automatically.

Refactors the following: `bun.new`, `bun.destroy`, `bun.create`, `bun.New`, `bun.NewRefCounted`. They are much easier to read now.

Main motivation for this is to introduce a new assertion which ensures types that forcefully opt into heap breakdown (From `bun.New`) are impossible to create via standard library allocators.

image

### DIFF
--- a/src/StandaloneModuleGraph.zig
+++ b/src/StandaloneModuleGraph.zig
@@ -74,31 +74,30 @@ pub const StandaloneModuleGraph = struct {
         loader: bun.options.Loader,
         contents: []const u8 = "",
         sourcemap: LazySourceMap,
-        blob_: ?*bun.JSC.WebCore.Blob = null,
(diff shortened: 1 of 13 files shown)


paperdave:
re: rename `HeapBreakdown` -> `heap_breakdown`

I want to slowly move parts of our codebase into proper casing.

PR #12454, opened 2024-07-09
Warn when installing a global binary and global bin is not in path

Jarred-Sumner:
### What does this PR do?

When you run `bun add -g typescript` and `$HOME/.bun/bin` is not in $PATH, bun install will now print a warning like the below:

```sh
bun add v1.1.19 (91600990)

installed typescript@5.5.3 with binaries:
 - tsc
 - tsserver

1 package installed [449.00ms]

warn: To run "tsc", add the global bin folder to $PATH:

fish_add_path "/Users/jarred/.bun/bin"
```

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -2586,6 +2586,10 @@ pub const PackageManager = struct {
     progress_name_buf_dynamic: []u8 = &[_]u8{},
     cpu_count: u32 = 0,
 
+    track_installed_bin: TrackInstalledBin = .{
+        .none = {},
+    },
+
     // progress bar stuff when not stack allocated
     root_progress_node: *Progress.Node = undefined,
 
@@ -2709,6 +2713,12 @@ pub const PackageManager = struct {
         }
         Global.crash();
     }
+
+    const TrackInstalledBin = union(enum) {
+        none: void,
+        pending: void,
+        basename: []const u8,
+    };
 
     // maybe rename to `PackageJSONCache` if we cache more than workspaces
     pub const WorkspacePackageJSONCache = struct {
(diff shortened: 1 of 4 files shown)


src/install/install.zig
```diff
@@ -9739,6 +9749,14 @@ pub const PackageManager = struct {
             Output.flush();
         }
 
+        // When you run `bun add -g <pkg>` or `bun install -g <pkg>` and the global bin dir is not in $PATH
+        // We should tell the user to add it to $PATH so they don't get confused.
+        if (subcommand == .add or subcommand == .install) {
```
dylan-conway: I think `update` should be included

src/install/lockfile.zig
```diff
@@ -1717,24 +1717,36 @@ pub const Printer = struct {
...
                         }
+
+                        {
+                            const fmt = comptime Output.prettyFmt("<r> <d>- <r><b>{s}<r>\n", enable_ansi_colors);
+                            var manager = &bun.PackageManager.instance;
+
+                            if (manager.track_installed_bin == .pending) {
+                                if (iterator.next() catch null) |bin_name| {
+                                    manager.track_installed_bin = .{
+                                        .basename = bun.default_allocator.dupe(u8, bin_name) catch bun.outOfMemory(),
```
dylan-conway, line 1739: I think it's probably fine, but this will ignore all but the first, e.g. `bun install -g esbuild tsc node-gyp`

Jarred-Sumner: that's intentional, you only want this message to appear for the first package. this is just for the label in the message

PR #12472, opened 2024-07-09
fix(server) fix abrupt stop

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/12489

- [X] Code changes

Added a test


src/bun.js/api/server.zig
```diff
@@ -2632,6 +2632,11 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             ctx.setAbortHandler();
         }
 
+        fn isAborted(this: *RequestContext) bool {
```
```suggestion
        fn isAborted(this: *const RequestContext) bool {
```

src/bun.js/api/server.zig
```diff
@@ -1561,22 +1561,19 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
+
             const resp = ctx.resp.?;
+            resp.clearAborted();
             const has_responded = resp.hasResponded();
             if (!has_responded)
                 ctx.runErrorHandler(
                     value,
                 );
 
-            if (ctx.isAborted()) {
```
Jarred-Sumner: What if the request aborts after the error handler runs?

cirospaciari: fixed https://github.com/oven-sh/bun/pull/12472/commits/c7bcbe8d6835f06b122ae22f7081098d84064a6d

packages/bun-uws/src/HttpResponse.h
```diff
@@ -581,6 +581,12 @@ struct HttpResponse : public AsyncSocket<SSL> {
         return this;
     }
 
+    HttpResponse* clearOnAborted() {
+        HttpResponseData<SSL> *httpResponseData = getHttpResponseData();
+
+        httpResponseData->onAborted = nullptr;
```
Jarred-Sumner, line 593:
```suggestion
        httpResponseData->onAborted = nullptr;
        httpResponseData->onWritable = nullptr;        
```

cirospaciari: fixed in https://github.com/oven-sh/bun/pull/12472/commits/6094340309d0fddecd377aeabb706d04d516afed we clean both only in some cases (onWritable being set or not changes some internal behaviors)

Issue #10789, opened 2024-05-03
TypeError: FormData parse error missing final boundary in SolidStart


huseeiin:
### What version of Bun is running?

1.1.6+e58d67b46

### What platform is your computer?

Linux 6.5.0-26-generic x86_64 x86_64

### What steps can reproduce the bug?

https://github.com/huseeiin/bun-sld-pwd
you have to submit any of the forms in the index page to see the error.

### What is the expected behavior?

for the FormData to be parsed correctly.

### What do you see instead?

![image](https://github.com/oven-sh/bun/assets/122984423/59bb86cf-1726-4078-98aa-d676f7130b96)

### Additional information

**important detail**: it works in prod for some reason
```bash
bun --bun run build
bun .output/server/index.mjs
```

huseeiin:
you have to submit any of the forms in the index page to see the error.

Electroid:
Duplicate of #2644

Issue #6348, opened 2023-10-06
Response.clone does not clone in-progress streaming body


robere2:
### What version of Bun is running?

1.0.4

### What platform is your computer?

Linux 5.10.60.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

[Calling the `clone()` method on a `Response` object should return an identical `Response` object, including the body](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone). Instead, it returns a `Response` that's missing the body, but otherwise identical in every other way (that I've found).

Minimal reproducible example:
```js
const res = await fetch("https://github.com/");
console.log("Clone Response length:", (await res.clone().text()).length)
console.log("Main Response length:", (await res.text()).length)
```

### What is the expected behavior?

The program should output two lines to the console, and each should contain the same integer number.

### What do you see instead?

The first line will contain the number "0" while the second line contains the expected value.

Hanaasagi:
Related to https://github.com/oven-sh/bun/issues/4765

Electroid:
Duplicate of #4765

dangeredwolf:
Regression of #913, related to #6969. Also seems to be a duplicate of #1381 and #2477.  This is a pretty big issue breaking a lot of things, I hope someone will finally pay attention to it because of how many things this bug effects. There's already an open PR for it, #1693, oh, and in #6468 as well!

liz3:
Theres a open pr for ages but the maintainers don't prio such fixes over other things:
https://github.com/oven-sh/bun/pull/6468

BoltDoggy:
Maybe Polyfill

```
Request.prototype.clone = function (this: Request) {
  const { body } = this;
  const teedX = body?.tee();
  Object.defineProperty(this, "body", {
    get() {
      return teedX?.[0] || null;
    },
  });  return new Request(this, {
    body: teedX?.[1] || null,
  });
};
```

More Idea

```
Request.prototype.breakClone = function (this: Request) {
  const {
    url,
    method,
    headers,
    cache,
    credentials,
    destination,
    integrity,
    keepalive,
    mode,
    redirect,
    referrer,
    referrerPolicy,
    signal,
    bodyUsed,
    body,
    clone,
  } = this;
  const teedX = body?.tee();
  const arrayBufferPromise = new Response(teedX?.[0]).arrayBuffer();
  const teedY = teedX?.[1].tee();
  Object.defineProperty(this, "body", {
    get() {
[...]

PR #12473, opened 2024-07-09
feat(body-stream-clone): initial commit with base case working for Re…

billywhizz:
### What does this PR do?

This PR should cover the following changes:

- ```request.clone()``` and ```response.clone()``` will handle cloning the request and response bodies, which they currently do not
- ```new Request(existing_request)``` will also clone the request body - it currently doesn't

this should resolve a number of issues and incompatibilities with existing frameworks when complete:

- https://github.com/oven-sh/bun/issues/6348
- https://github.com/oven-sh/bun/issues/1381
- https://github.com/oven-sh/bun/issues/2477
- https://github.com/oven-sh/bun/issues/2644
- https://github.com/oven-sh/bun/issues/10789
- https://github.com/oven-sh/bun/issues/4161

this initiall draft should resolve the issue for ```new Request(existing_request)```. other fixes will be added as they are ready.

## Todo

- [ ] tests for request.clone() with text/buffer/stream bodies
- [ ] tests for response.clone() with text/buffer/stream bodies
- [ ] tests for ```new Request(existing_request)``` with text/buffer/stream bodies

- tests to come...


src/bun.js/webcore/request.zig
```diff
@@ -818,11 +840,60 @@ pub const Request = struct {
...
+                        }
+
+                        break :stream null;
+                    };
+
+                    if (stream) |s| {
+                        if (s.tee(globalThis)) |teed| {
+                            Request.bodySetCached(req_jsvalue, globalThis, teed[0].value);
+
+                            teed[1].value.protect();
```
Jarred-Sumner: This `protect` likely causes a memory leak when used because the Strong below does the job of the protect

Jarred-Sumner:
Note that `1.1.19-canary.1+9941f7b47` does not address #1381,  though it does change the output

```js
export default {
  port: 4002,
  async fetch(request) {
    console.log(".clone().json()", await request.clone().json());
    console.log(".json()", await request.json());
    return new Response("Ok");
  },
};
```

Run via: ` curl http://localhost:4002 -d '{"hey": "123"}'`

This console.log's the incoming request JSON successfully, and then hangs after the second call to request.clone().

PR #12487, opened 2024-07-10
fix: add error messages for using readable stream to etc methods

paperclover:
### What does this PR do?

Fixes this report on discord:

image


Jarred-Sumner:
`instanceof` throws and we shouldn't do multiple instanceof checks since it's (1) observable and (2) expensive

paperdave:
ah yes

PR #12562, opened 2024-07-14
Add packages option to remove all dependencies from bundle

### LINKED ISSUE #12293: Implement esbuild `packages` option in bun bundler

sometimes, we want to bundle all of our local imports (i.e. `./utils.js`) but not our `node_modules` (i.e. `svelte`)

Implement esbuild `packages: "external"` option
https://esbuild.github.io/api/#packages

```ts
import pkg from "./package.json";

Bun.build({
  external: Object.keys(pkg.dependencies),
});
```

zpix1:
### What does this PR do?

Closes #12293 

Changes add `--packages` option to bundler similar to [esbuild option](https://esbuild.github.io/api/#packages). This option removes all packages (marks them as external) (any imports not starting with '.', '..' or '/') from bundle if set to `external`. It defaults to `bundle` and does nothing in that case.

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
This looks very close to complete, impressed that you made this much progress without asking for help

Jarred-Sumner:
These tests are known to be flaky on Windows. This looks good, I'll merge it. Thank you for this

PR #12575, opened 2024-07-15
Replace some Identifier::fromString usages with vm->propertyNames

Jarred-Sumner:
### What does this PR do?

Replace some Identifier::fromString usages with vm->propertyNames

Small performance thing


paperdave:
we should add `Identifier::fromString(vm, "name"_s)` as a banned.json phrase lol

Jarred-Sumner:
It's in a lot of generated code from WebKit, so that'd be difficult. I kind of think we should have a formatter do a find-and-replace for things like this.

PR #12576, opened 2024-07-15
feat(bundler): implement non-isolated hashes

paperclover:
### What does this PR do?

When you use code splitting, you end up with cross chunk imports:

```ts
// entry.ts
const other = await import('./second');
```
```ts
// second.ts
export const y = 1;
```

Currently, the hashes generated are "isolated" so they can be computed in parallel. But this has the problem of changing `second.ts` will not affect the hash of `entry.ts`, despite changing the import url in it.

This fixes this, making sure that hashes actually produce unique bundles. This follows exactly what esbuild does.

In addition, the hashes are now shorter using a base32 encoding instead of a hex. Part of this is also because it is truncating to 8 characters (same as esbuild)

image


src/bundler/bundle_v2.zig
```diff
@@ -9122,12 +9146,17 @@ const LinkerContext = struct {
             };
             var duplicates_map: bun.StringArrayHashMapUnmanaged(DuplicateEntry) = .{};
 
-            // Compute the final hashes of each chunk. This can technically be done in
-            // parallel but it probably doesn't matter so much because we're not hashing
-            // that much data.
-            for (chunks) |*chunk| {
-                // TODO: non-isolated-hash
-                chunk.template.placeholder.hash = chunk.isolated_hash;
+            var chunk_visit_map = try BitSet.initEmpty(c.allocator, chunks.len);
```
```suggestion
            var chunk_visit_map = try AutoBitSet.initEmpty(c.allocator, chunks.len);
```

PR #12578, opened 2024-07-15
fix: make raiseIgnoringPanicHandler ignore the panic handler

paperclover:
### What does this PR do?

I guess the new panic handler is so good at handling crashes that the method used before in `raiseIgnoringPanicHandler` was not actually ignoring it. This definitely happened on windows, considering there was no code at all to reset signal handlers.

We have a function for this: `bun.crash_handler.resetSegfaultHandler();`. We just invoke that now!


paperdave:
will need to have a test

PR #12581, opened 2024-07-15
fix(serve) fix abrupt close when downloading data

cirospaciari:
### What does this PR do?

- [X] Code changes

Added a test


src/bun.js/api/server.zig
```diff
@@ -1538,10 +1536,60 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
+
+            const server = this.server;
+            server.request_pool_allocator.put(this);
+            server.onRequestComplete();
+        }
+
+        pub fn deref(this: *RequestContext) void {
+            streamLog("deref", .{});
+            bun.assert(this.ref_count > 0);
+            this.ref_count -|= 1;
```
```suggestion
            const ref_count = this.ref_count;
            this.ref_count -= 1;
            if (ref_count == 1)
```

PR #12585, opened 2024-07-16
Prevent unref from hanging on uninitialized dgram socket 

### LINKED ISSUE #12561: node:dgram UDPSocket unref doesn't apply on unitialized socket

1.1.18+5a0b93523

Linux 6.6.36 x86_64 unknown

Calling unref on a UDP socket should prevent it from blocking the process from closing, however when using the node compatible dgram socket unref does not prevent the blocking.

```ts
import dgram from "node:dgram";

const socket = dgram.createSocket({ type: "udp4" });

// Should prevent the socket from blocking process end
socket.unref();

send("test", 1337, "127.0.0.1", (error, bytes) => {
  console.log(error, bytes);
});

// Process hangs after send is finished
```

Calling unref on a dgram socket should prevent it from blocking closing the process; the same behavior as the Deno.udpSocket:

```ts
const socket = await Bun.udpSocket({});

socket.unref();

socket.send("test", 1337, "127.0.0.1");
```

### What do you see instead?

_No response_

### Additional information

I believe this is because the node UdpSocket.unref impl only passes the unref to the inner socket if it exists, hence if you call unref before any connections are made it has no effect.
https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/src/js/node/dgram.ts#L934-L940

This seems to be confirmed by this example not hanging:
```ts
import dgram from "node:dgram";

const socket = dgram.createSocket({ type: "udp4" });

send("test", 1337, "127.0.0.1", (error, bytes) => {
  console.log(error, bytes);
  socket.unref();
});
```

190n:
### What does this PR do?

Make calling `unref` on a `Socket` from `node:dgram` work even if the socket has not been set up yet. Fixes #12561.

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
Should `bind` set `unrefOnBind` to `false`?

190n:
> Should bind set unrefOnBind to false?

Probably, to be safe, although I don't think you can call `bind` twice without errors.

Issue #12597, opened 2024-07-16
Bun panic by `Bun.CryptoHasher.update`


vdata1:
### How can we reproduce the crash?

Hi, 

I want to report a possible bug, it can reproduced by running the following code: 

```javascript
const hasher = new Bun.CryptoHasher("sha1");
hasher.update();
```

### Relevant log output

```shell
============================================================
Bun v1.1.18 (5a0b9352) macOS x64
macOS v14.5
Args: "bun" "run" "bun.js"
Features: jsc 
Builtins: "bun:main" 
Elapsed: 30ms | User: 15ms | Sys: 11ms
RSS: 16.10MB | Peak: 16.10MB | Commit: 1.07GB | Faults: 88

panic(main thread): Segmentation fault at address 0x5
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
```

### Stack Trace (bun.report)

Bun v1.1.18 ([`5a0b935`](<https://github.com/oven-sh/bun/tree/5a0b93523134ca0043ee7ae84f01003aa4b7ff69>)) on macos x86_64 [RunCommand]

Segmentation fault at address 0x00000005

- *2 unknown/js code*
- `src.bun.js.node.types.StringOrBuffer.fromJSMaybeAsync`
- `src.bun.js.node.types.StringOrBuffer.fromJSWithEncodingMaybeAsync`
- `src.bun.js.node.types.BlobOrStringOrBuffer.fromJSWithEncodingValueMaybeAsync`
- `CryptoHasherPrototype__update`
- `WebCore::CryptoHasherPrototype__updateCallback`
- *1 unknown/js code*
- `llint_entry`
- `vmEntryToJavaScript`

vdata1:
Yes, it works on v1.1.20 too.

PR #12607, opened 2024-07-16
fix(CryptoHasher) check of .empty/null/undefined in update

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/12597
Fix: https://github.com/oven-sh/bun/issues/12598

- [X] Code changes

Added a test

### DIFF
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -394,9 +394,6 @@ pub const StringOrBuffer = union(enum) {
     }
 
     pub fn fromJSMaybeAsync(global: *JSC.JSGlobalObject, allocator: std.mem.Allocator, value: JSC.JSValue, is_async: bool) ?StringOrBuffer {
-        if (value.isEmptyOrUndefinedOrNull()) {
-            return null;
-        }
         return switch (value.jsType()) {
             JSC.JSValue.JSType.String, JSC.JSValue.JSType.StringObject, JSC.JSValue.JSType.DerivedStringObject, JSC.JSValue.JSType.Object => {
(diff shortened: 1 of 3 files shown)


src/bun.js/node/types.zig
```diff
@@ -394,6 +394,9 @@ pub const StringOrBuffer = union(enum) {
     }
 
     pub fn fromJSMaybeAsync(global: *JSC.JSGlobalObject, allocator: std.mem.Allocator, value: JSC.JSValue, is_async: bool) ?StringOrBuffer {
+        if (value.isEmptyOrUndefinedOrNull()) {
```
```suggestion
        if (value.isEmpty()) {
```

src/bun.js/node/types.zig
```diff
@@ -443,6 +446,9 @@ pub const StringOrBuffer = union(enum) {
     }
 
     pub fn fromJSWithEncodingMaybeAsync(global: *JSC.JSGlobalObject, allocator: std.mem.Allocator, value: JSC.JSValue, encoding: Encoding, is_async: bool) ?StringOrBuffer {
+        if (value.isEmptyOrUndefinedOrNull()) {
```
```suggestion
        if (value.isEmpty()) {
```

Jarred-Sumner:
Usually the caller should be checking if the JSValue is empty/undefined/null and not each invocation

Jarred-Sumner:
The bug here more is that we're missing a check that `input` is undefined

https://github.com/oven-sh/bun/blob/bd39c53fba5ba088c813b9fdee18ca3da7d70af9/src/bun.js/api/BunObject.zig#L2774-L2775

PR #12631, opened 2024-07-17
bun-sqlite: add sqlite3_update_hook

jecquas:
### What does this PR do?

Add some methods to `Database` to listen for Data Change Notification Callbacks provided by `sqlite3_update_hook` https://www.sqlite.org/c3ref/update_hook.html

- [x] Code changes

 I wrote automated tests

### DIFF
--- a/packages/bun-types/sqlite.d.ts
+++ b/packages/bun-types/sqlite.d.ts
@@ -487,6 +487,46 @@ declare module "bun:sqlite" {
      * @link https://www.sqlite.org/c3ref/file_control.html
      */
     fileControl(zDbName: string, op: number, arg?: ArrayBufferView | number): number;
+
+    /**
+     * Registers a listener for the specified update type.
+     * @param {("insert" | "update" | "delete")} updateType - The type of update to listen for.
+     * @param {(dbName: string, tableName: string, rowId: number) => void} listener - The callback function to be invoked when the event occurs.
+     */
+    on(
+      updateType: "insert" | "update" | "delete",
+      listener: (dbName: string, tableName: string, rowId: number) => void,
+    ): void;
+
+    /**
+     * Unregisters a listener for the specified update type.
+     * @param {("insert" | "update" | "delete")} updateType - The type of update the listener was registered for.
+     * @param {(dbName: string, tableName: string, rowId: number) => void} listener - The callback function to be removed.
+     */
+    off(
+      updateType: "insert" | "update" | "delete",
+      listener: (dbName: string, tableName: string, rowId: number) => void,
+    ): void;
+
+    /**
+     * Registers a listener for the specified update type.
+     * @param {("insert" | "update" | "delete")} updateType - The type of update to listen for.
+     * @param {(dbName: string, tableName: string, rowId: number) => void} listener - The callback function to be invoked when the event occurs.
+     */
+    addEventListener(
+      updateType: "insert" | "update" | "delete",
+      listener: (dbName: string, tableName: string, rowId: number) => void,
+    ): void;
+
+    /**
+     * Unregisters a listener for the specified update type.
+     * @param {("insert" | "update" | "delete")} updateType - The type of update the listener was registered for.
+     * @param {(dbName: string, tableName: string, rowId: number) => void} listener - The callback function to be removed.
+     */
+    removeEventListener(
+      updateType: "insert" | "update" | "delete",
+      listener: (dbName: string, tableName: string, rowId: number) => void,
+    ): void;
   }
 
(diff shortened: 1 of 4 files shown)


packages/bun-types/sqlite.d.ts
```diff
@@ -487,6 +487,46 @@ declare module "bun:sqlite" {
      * @link https://www.sqlite.org/c3ref/file_control.html
      */
     fileControl(zDbName: string, op: number, arg?: ArrayBufferView | number): number;
+
+    /**
```
Jarred-Sumner, line 491: `Database` isn't currently an `EventEmitter` or an `EventTarget`. I think it will be confusing to have these methods which look like one of those, but aren't

Instead, maybe we could have some getters/setters like

```
database.onUpdate = (database, tableName, rowId) => {}
database.onInsert = (database, tableName, rowId) => {}
database.onDelete = (database, tableName, rowId) => {}
```

Then, in sqlite.ts we could create these functions via .bind, so that the first argument is always the `database` object and then the `tableName` and `rowId` are passed along

These would be implemented as getter/setters on the `Database` instance itself, and unregistered when the finalization callback occurs in the JSSQLDatabase probably.

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -1764,6 +1764,110 @@ JSC_DEFINE_HOST_FUNCTION(jsSQLStatementFcntlFunction, (JSC::JSGlobalObject * lex
...
+
+void on_update_hook(void * dataPtr, int type, const char * dbName, const char * tableName, sqlite_int64 rowId) {
+    UpdateHookData* data = (UpdateHookData *)dataPtr;
+
+    auto globalObject = data->globalObject.get();
+    auto callback = data->callback.get();
+
+    JSValue typeValue = jsNumber(type);
+    JSValue dbNameValue = jsString(globalObject->vm(), makeString(ASCIILiteral::fromLiteralUnsafe(dbName)));
+    JSValue tableNameValue = jsString(globalObject->vm(), makeString(ASCIILiteral::fromLiteralUnsafe(tableName)));
```
Jarred-Sumner: When creating a string from a UTF-8 data source, we need to use `String::fromUTF8` instead. `ASCIILiteral` can only be used for constant strings like when you do `"foo"_s` .

packages/bun-types/sqlite.d.ts
Jarred-Sumner: The other option is we make `Database` an `EventEmitter` and that would also be an answer to how it would handle exceptions and promise rejections in these callbacks.

src/bun.js/bindings/sqlite/JSSQLStatement.cpp
```diff
@@ -1764,6 +1764,110 @@ JSC_DEFINE_HOST_FUNCTION(jsSQLStatementFcntlFunction, (JSC::JSGlobalObject * lex
...
+    Strong<JSC::JSGlobalObject> globalObject;
+    Strong<JSC::JSFunction> callback;
+} UpdateHookData;
+
+UpdateHookData * data = nullptr;
+
+void on_update_hook(void * dataPtr, int type, const char * dbName, const char * tableName, sqlite_int64 rowId) {
+    UpdateHookData* data = (UpdateHookData *)dataPtr;
+
+    auto globalObject = data->globalObject.get();
```
Jarred-Sumner, line 1777: This will need an exception check

```suggestion
    auto globalObject = data->globalObject.get();
    auto &vm = globalObject->vm();
    auto scope = DECLARE_THROW_SCOPE(vm);
    RETURN_IF_EXCEPTION(scope, void());
```

src/js/bun/sqlite.ts
```diff
@@ -91,6 +92,13 @@ const constants = {
   SQLITE_FCNTL_RESET_CACHE: 42,
 };
 
+// https://www.sqlite.org/c3ref/c_alter_table.html
+const sqliteActionCodes = {
```
Jarred-Sumner: We can use `const enum` and avoid paying the cost of the property access

```suggestion
const enum sqliteActionCodes {
```

packages/bun-types/sqlite.d.ts
jecquas: > Instead, maybe we could have some getters/setters like

This approach would prevent us from setting multiple listeners, wouldn't it?
Maybe this whole logic should not be part of sqlite.ts and we can just expose the `updateHook` function with documentation and let people build their event emitter on top of it?

PR #12643, opened 2024-07-18
Fix truncating in BigIntStats

Jarred-Sumner:
### What does this PR do?

- [ ] Code changes

### DIFF
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -1390,8 +1390,8 @@ pub fn StatType(comptime Big: bool) type {
 
         // Stats stores these as i32, but BigIntStats stores all of these as i64
         // On windows, these two need to be u64 as the numbers are often very large.
-        dev: if (Environment.isWindows) u64 else Int,
-        ino: if (Environment.isWindows) u64 else Int,
+        dev: u64,
+        ino: u64,
         mode: Int,
         nlink: Int,
         uid: Int,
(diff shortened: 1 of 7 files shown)


src/bun.js/node/types.zig
```diff
@@ -1441,10 +1441,16 @@ pub fn StatType(comptime Big: bool) type {
             return struct {
                 pub fn callback(this: *This, globalObject: *JSC.JSGlobalObject) JSC.JSValue {
                     const value = @field(this, @tagName(field));
-                    if (comptime (Big and @typeInfo(@TypeOf(value)) == .Int)) {
+                    const Type = @TypeOf(value);
+                    if (comptime Big and @typeInfo(Type) == .Int) {
+                        if (Type == u64) {
+                            return JSC.JSValue.fromUInt64NoTruncate(globalObject, @intCast(value));
```
paperclover: why is intCast used here?

src/bun.js/node/types.zig
```diff
@@ -1441,10 +1441,16 @@ pub fn StatType(comptime Big: bool) type {
...
+                        }
+
                         return JSC.JSValue.fromInt64NoTruncate(globalObject, @intCast(value));
```
paperclover: similarly, i dont think intCast here makes sense either

paperclover (changes requested):

Issue #4361, opened 2023-08-26
setTimeout doesn't have promisify.custom


sirenkovladd:
### What version of Bun is running?

0.8.1+16b4bf341acc0f4804f0b6bdf5298c180cd00366

### What platform is your computer?

Linux 6.2.0-26-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
require('util').promisify(setTimeout)(100, 'ok').then(console.log);
```

### What is the expected behavior?

```sh
❯ node bug.js
ok
```

### What do you see instead?

```sh
❯ bun bug.js 
374 |       writable: !1,
375 |       configurable: !0
376 |     }), fn;
377 |   }
378 | 
379 |   function fn() {
                             ^
TypeError: setTimeout expects a function
      at fn (node:util:379:26)
      at <path>/bug.js:17:0
      at globalThis (<path>/bug.js:17:65)
```

### Additional information

```js
console.log(setTimeout[require('util').promisify.custom]);
```

```sh
❯ bun bug.js
undefined

❯ node bug.js 
[Function: setTimeout]
```

https://github.com/oven-sh/bun/blob/bun-v0.8.1/src/js/node/util.js#L478

sirenkovladd:
It blocks the [close-with-grace](https://www.npmjs.com/package/close-with-grace?activeTab=dependents) (dependency in [fastify-cli](https://www.npmjs.com/package/fastify-cli))

evmd:
Any update on this ?

sirenkovladd:
Fixed in https://github.com/oven-sh/bun/pull/9467

PR #12660, opened 2024-07-19
Defer finalization for napi callbacks to the immediate task queue

Jarred-Sumner:
### What does this PR do?

In node, when napi finalizers are run, they are enqueued to the immediate task queue. They're not run immediately. This behavior difference can cause crashes if the napi addon creates any JS objects or does anything which causes GC inside of a finalizer.

https://github.com/nodejs/node/blob/2eff28fb7a93d3f672f80b582f664a7c701569fb/src/js_native_api_v8.cc#L71-L84


robobun:
@190n, your commit e66ec2a10bc6fe1d79e10d46cff38387eb3bdadc is building: [`#4361`](https://buildkite.com/bun/bun/builds/4361)

190n:
Seems to fix #14295, but needs more thorough testing (I haven't had a Pulsar server for it to connect to yet so it was just crashing creating a timeout error) + automated tests

190n:
Superseded by #14501

Issue #12650, opened 2024-07-18
`expect.toThrow()` fails with `expect.any()`


MattBDev:
### What version of Bun is running?

1.1.20+ae1948925

### What platform is your computer?

Linux 5.15.146.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

In Jest the following test passes:
```JavaScript
describe('Test expect.toThrow(expect.any())', () => {
  it('should throw an error', () => {
    expect(() => {
      throw new Error('error')
    }).toThrow(expect.any(Error))
  })
})
```
However in Bun the test fails with the following error:
```
4 | describe('Test expect.toThrow(expect.any())', () => {
5 |   it('should throw an error', () => {
6 |     expect(() => {
7 |       throw new Error('error')
8 |     }).toThrow(expect.any(Error))
           ^
error: expect(received).toThrow(expected)

Expected constructor: ExpectAny
Received constructor: Error

Received message: "error"
```

### What is the expected behavior?

The expected behavior is that `expect.toThrow()` should be using the return value of the function `expect.any()`.

### Additional information

I've tested this multiple times and created a project without any libraries to test it in Jest and in Bun. The issue is specifically with the behavior of `expect.toThrow(expect.any())`.

nektro:
fixed in bun@1.1.21

PR #12670, opened 2024-07-19
Fix expect.toThrow(expect.any()) matcher to correctly handle ExpectAny objects

ippsav:
### What does this PR do?

This PR fixes an issue with the `expect.toThrow(expect.any())` matcher in Bun's testing framework. Previously, this matcher was not correctly handling `ExpectAny` objects created by `expect.any()`, leading to test failures that would pass in Jest.
Link to the issue: https://github.com/oven-sh/bun/issues/12650

Specifically, this PR:

- Modifies the `toThrow` matcher to correctly unwrap `ExpectAny` objects
- Ensures that the constructor comparison is done with the actual constructor passed to `expect.any()`, rather than the `ExpectAny` object itself

- [x] Code changes

I wrote automated tests to verify the fix:
```ts
import {expect, describe, it} from "bun:test";

// Custom class for testing
class CustomException extends Error {
  constructor(message: string) {
    super(message);
    this.name = "CustomException";
  }
}

describe('Test expect.toThrow(expect.any())', () => {
  it('should throw an error', () => {
    expect(() => {
      throw new CustomException("Custom error message")
    }).toThrow(expect.any(Error))
  })

  it('should throw a CustomException', () => {
    expect(() => {
      throw new CustomException("Custom error message")
    }).toThrow(expect.any(CustomException))
  })
})
```
This test now passes in Bun, matching the behavior in Jest.

- [x] I ran my tests locally and they pass (`bun-debug test test-file-name.test.ts`)

### Additional Notes

This fix brings Bun's `expect.toThrow(expect.any())` behavior in line with Jest's implementation, Though I'm not familiar with bun's codebase, is there a way to know if value is an instance of ExpectAny before trying to get the cached value ?

### DIFF
--- a/src/bun.js/test/expect.zig
+++ b/src/bun.js/test/expect.zig
@@ -2285,7 +2285,11 @@ pub const Expect = struct {
                 globalThis.throw("Expected value must be string or Error: {any}", .{value.toFmt(globalThis, &fmt)});
                 return .zero;
             } else if (value.isObject()) {
-                if (ExpectAny.constructorValueGetCached(value)) |v| break :brk v;
+                if (ExpectAny.fromJSDirect(value)) |_| {
+                    if (ExpectAny.constructorValueGetCached(value)) |innerConstructorValue| {
+                        break :brk innerConstructorValue;
+                    }
+                }
             }
             break :brk value;
         } else .zero;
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
I think this is still not 100% right but the edgecases are very much edgecases (`expect.any(String)` likely won't work because `"abc" instanceof String` is false)  and it is better than status quo

ippsav:
> I think this is still not 100% right but the edgecases are very much edgecases (`expect.any(String)` likely won't work because `"abc" instanceof String` is false) and it is better than status quo

Will take a look at that soon thanks for pointing that out !

PR #12683, opened 2024-07-20
Fix assertion failure in bun build when entry point is a file loader

Jarred-Sumner:
### What does this PR do?

Fix assertion failure in bun build when entry point is a file loader

manual this time


paperdave:
i will add a very trivial test to this

Issue #12161, opened 2024-06-25
Support Node's process.getBuiltinModule


jakebailey:
### What is the problem this feature would solve?

This function was added in Node 22.3.0 as part of the work to support `require(ESM)`, as the API allows libraries to access builtin modules without top-level await. It would be good if bun also supported this on the process object so that code written for Node with this function (likely, TypeScript in the future) can work.

(I have a workaround for now in my TypeScript ESM branch where I just see if `require` is defined, which it is in Bun thanks to its mixed-format loading.)

### What is the feature you are proposing to solve the problem?

[https://github.com/nodejs/node/pull/52762](https://redirect.github.com/nodejs/node/pull/52762)

paperdave:
This should be moderately easy to implement by creating a function in `BunProcess.cpp`, and then making it wrap `require` (with validation that it is indeed a built in module), or `InternalModuleRegistry`. adding the 'good first issue' tag but this isnt the most trival 'good first issue'.

ippsav:
i ll try and solve this issue

snowfluke:
Any update on this?

lgarron:
I'd also love to see https://github.com/oven-sh/bun/pull/12689 land.

I'm relying on https://github.com/lgarron/getbuiltinmodule-ponyfill for now, but it involves hardcoding types in an unsustainable way: https://github.com/lgarron/getbuiltinmodule-ponyfill/blob/main/index.d.ts

snowfluke:
the PR https://github.com/oven-sh/bun/pull/12689 has been changed to draft

snowfluke:
any workaround for pdfjs?
I got:

```sh
6452 | class NodeFilterFactory extends BaseFilterFactory {}
6453 | class NodeCanvasFactory extends BaseCanvasFactory {
6454 |   _createCanvas(width, height) {
6455 |     const require = process.getBuiltinModule("module").createRequire(import.meta.url);
                                   ^
TypeError: process.getBuiltinModule is not a function. (In 'process.getBuiltinModule("module")', 'process.getBuiltinModule' is undefined)
```

paperclover:
if you want a very loose workaround,
```
process.getBuiltinModule = require;
```
this will be fixed soon

lgarron:
> this will be fixed soon

oooh! 😯

Is there a PR/commit we can track for this?

PR #12689, opened 2024-07-20
Add getBuiltinModule function to Process object

ippsav:
# Add `process.getBuiltinModule` function

- Implement `Process_functionLoadBuiltinModule` to load built-in modules
- Add `getBuiltinModule` to the `processObjectTable`
- Return the exports object of the loaded module
- TODO

Related issue: https://github.com/oven-sh/bun/issues/12161

### What does this PR do?

This PR adds `getBuiltinModule` to the `process` object in Bun. This function allows direct access to built-in modules without using `require`.

The main changes include:
- Adding `Process_functionLoadBuiltinModule` that handles the loading of built-in modules
- Registering this function in the `processObjectTable` to make it available as `process.getBuiltinModule`
- Implementing the logic to fetch and return the exports of the requested built-in module

This feature solves the problem of accessing built-in modules in environments where top-level `await` is not available or desirable, and provides a consistent API with recent versions of Node.js.

- [x] Code changes

## Additional Notes
This implementation aims to match the behavior of Node.js 22.3.0+.

### DIFF
--- a/src/bun.js/bindings/BunProcess.cpp
+++ b/src/bun.js/bindings/BunProcess.cpp
@@ -4,12 +4,18 @@
 #include <JavaScriptCore/ObjectConstructor.h>
 #include <JavaScriptCore/NumberPrototype.h>
 #include "CommonJSModuleRecord.h"
+#include "JavaScriptCore/CallData.h"
 #include "JavaScriptCore/CatchScope.h"
+#include "JavaScriptCore/EnumerationMode.h"
+#include "JavaScriptCore/ExceptionScope.h"
 #include "JavaScriptCore/JSCJSValue.h"
 #include "JavaScriptCore/JSCast.h"
+#include "JavaScriptCore/JSMap.h"
 #include "JavaScriptCore/JSString.h"
 #include "JavaScriptCore/Protect.h"
 #include "JavaScriptCore/PutPropertySlot.h"
+#include "ModuleLoader.h"
+#include "PathInlines.h"
 #include "ScriptExecutionContext.h"
 #include "headers-handwritten.h"
 #include "node_api.h"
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
I think these are all flaky tests

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2778,6 +2788,70 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionKill,
...
+        throwTypeError(zigGlobalObject, scope, "The module must be a built-in module name"_s);
+        return JSValue::encode(jsUndefined());
+    }
+
+    JSString* requireMapKey = JSC::jsString(vm, moduleNameString);
+
+    JSString* dirname = jsEmptyString(vm);
+    JSString* filename = requireMapKey;
+
+    JSCommonJSModule* moduleObject = JSCommonJSModule::create(
```
Jarred-Sumner, line 2820: We shouldn't create a new object here on each call. It shouldn't need to create one of these at all, in fact.

Jarred-Sumner: Oh, we would still need to for the wrapper object version hm

ippsav: Where/how the required module should be loaded when `Bun::fetchCommonJSModule` expects a target to put the module in (and it returns the target itself at the end)?

ippsav: Oh github just loaded this message for some reason.

> Oh, we would still need to for the wrapper object version hm

Does that mean that even though the object creation here seems to be redundant, but needed to wrap the module loaded is that correct ?

lgarron:
Does this need any substantial changes to land?

Even if it could use some improvements, the functionality would be super valuable for those of us who write cross-compatible libraries and need to load `node` code conditionally. 😻

lgarron:
*poke*?

This is the kind of feature that is not so much *difficult* to work around as it is *impossible* to work around without making assumptions that limit where the code can be run.

This function is extremely useful for those of us who need to write portable code without hacks.

Issue #12573, opened 2024-07-15
Bun test coverage includes node_modules package tests on Windows


antlanc7:
### What version of Bun is running?

1.1.20+ae1948925

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Run `bun test --coverage` in  a project with some dependencies, for example Elysia:
- `mkdir test-coverage-bug`
- `cd test-coverage-bug`
- `bun init -y`
- `bun add elysia`
- create file `index.test.ts` with following content: 
```ts
import { expect, it } from "bun:test";
import Elysia from "elysia";
const app = new Elysia();
it("index", () => expect(app).toBeTruthy());
```
- `bun test --coverage`

### What is the expected behavior?

On Linux (WSL) it results in:
```
bun test v1.1.20 (ae194892)

index.test.ts:
✓ index [0.02ms]

 1 pass
 0 fail
 1 expect() calls
Ran 1 tests across 1 files. [234.00ms]
```

### What do you see instead?

On Windows it results in:

```
bun : bun test v1.1.20 (ae194892)

index.test.ts:
(pass) index
-------------------------------------------------------------------------------------------------|---------|---------|-------------------
File                                                                                             | % Funcs | % Lines | Uncovered Line #s
-------------------------------------------------------------------------------------------------|---------|---------|-------------------
All files                                                                                        |   47.35 |   67.41 |
 index.test.ts                                                                                   |  100.00 |  100.00 | 
 node_modules\@sinclair\typebox\build\esm\compiler\compiler.mjs                                  |    8.33 |   23.84 | 24-28,32,36,40,44-46,50-53,62,66,70,74,84-86,
89-97,100,103,113-123,133,142-143,148-149,158,162,166-168,172,176,187,193,196-217,220,223-233,236,239,242-252,255,258-268,271-283,287,290-294,298,301-302,305,308-31
8,321-353,356,359-369,372-377,380-386,389-400,403,406-408,411-412,415-422,425,428-429,432-436,439,442,445-447,450-532,549,552,555-557,560-564,567-568,571,577-586,59
0-607,612-632
 node_modules\@sinclair\typebox\build\esm\compiler\index.mjs                                     |  100.00 |  100.00 | 
 node_modules\@sinclair\typebox\build\esm\errors\errors.mjs                                      |    2.38 |   18.62 | 90-91,98,104,111-114,118-119,126,133-171,174-
[...]

Issue #11613, opened 2024-06-05
bun:test exclude node_modules when use '--coverage'


tialki:
### What is the problem this feature would solve?

When run test, exclude node_modules when use '--coverage'

### What is the feature you are proposing to solve the problem?

Provider a config to exclude paths from coverage reports
```toml
[test]
coverageReportsExclude = ['node_modules']
```

vtcaregorodtcev:
Is there a possibility to specify files to collect coverage from? Like `--collectCoverageFrom` from jest?

dariushalipour:
fwiw, I think node_modules is an obvious directory to be ignored and this has been already considered in Bun codebase; the only problem is that it has been missing the proper check for Windows machines, since Windows machines use "\" in paths instead of "/".

Jarred-Sumner:
Fixed in #12691

PR #12691, opened 2024-07-20
Fix: test coverage node_modules exclusion in Windows

dariushalipour:
### What does this PR do?

Fixes the issue where test coverage was not excluding node_modules directory files from the report in Windows.

This PR addresses the following issues:
- Fixes https://github.com/oven-sh/bun/issues/12573
- Fixes  https://github.com/oven-sh/bun/issues/11613

**Root Cause**
The root cause of the issue is that Windows uses backslashes in paths rather than forward slashes, and the bun code was assuming it is always forward slashes regardless of the environment.

**Fix Approach**
Although it's ideal to abstract away the path formation differences (e.g., arrays of path pieces held in a path value object), for the sake of simplicity and to avoid a huge refactor just to fix this issue, I have taken the simplest approach to address the problem at hand.

- [x] Code changes

- I have built the changed bun in my Windows machine and ran `bun-debug.exe` against a repository facing the issue.
- Added automated tests, too:
- [x] I ran my tests locally and they pass (`bun-debug test coverage`)

### DIFF
--- a/src/cli/test_command.zig
+++ b/src/cli/test_command.zig
@@ -598,7 +598,7 @@ const Scanner = struct {
         };
 
         // always ignore node_modules.
-        if (strings.contains(slice, "/" ++ "node_modules" ++ "/")) {
+        if (strings.contains(slice, "/node_modules/") or strings.contains(slice, "\\node_modules\\")) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Can you add a test? Code without tests tends to break eventually

dariushalipour:
@Jarred-Sumner I've added the test. Glad you didn't let me skip the test. Ready for review again.

Jarred-Sumner:
Thank you

PR #12698, opened 2024-07-21
Introduce `bun --fetch-preconnect <url> ./my-script.ts`

Jarred-Sumner:
### What does this PR do?

This adds a `--fetch-preconnect` CLI argument which lets you "warm up" a fetch request before any javascript code is loaded. It performs the DNS lookup, establishes a TCP connection, and runs TLS handshaking eagerly.

Also: fixes https://github.com/oven-sh/bun/issues/12701

This is useful for environments where you know your server or script will immediately make a request after it loads, but it takes some time for everything to load.

```ts
$ bun --fetch-preconnect=https://example.com ./hey.mjs
[8.37ms] fetch(https://example.com)
$ bun hey.mjs # Without preconnect
[51.92ms] fetch(https://example.com)
$ node hey.mjs
fetch(https://example.com): 68.417ms
```

```ts
   1   │ if (globalThis.Bun) {
   2   │   await Bun.sleep(1000);
   3   │ } else {
   4   │   await new Promise(resolve => setTimeout(resolve, 1000));
   5   │ }
   6   │
   7   │ console.time("fetch(https://example.com)");
   8   │ await fetch("https://example.com");
   9   │ console.timeEnd("fetch(https://example.com)");
```

This also adds a `fetch.preconnect(url: string): void` function, which does the same thing programmatically.

This is a lot like `<link rel="preconnect" href="<url>">` in HTML.

There are tests. And docs.

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -1858,6 +1858,14 @@ pub const AsyncHTTP = struct {
         url: URL,
         is_url_owned: bool,
     ) void {
+        if (!FeatureFlags.is_fetch_preconnect_supported) {
+            if (is_url_owned) {
+                bun.default_allocator.free(url.href);
+            }
+
+            return;
+        }
+
         var this = Preconnect.new(.{
             .async_http = undefined,
             .response_buffer = MutableString{ .allocator = default_allocator, .list = .{} },
@@ -2443,9 +2451,11 @@ pub fn onWritable(this: *HTTPClient, comptime is_first_call: bool, comptime is_s
         return;
     }
 
-    if (this.is_preconnect_only) {
-        this.onPreconnect(is_ssl, socket);
-        return;
+    if (comptime FeatureFlags.is_fetch_preconnect_supported) {
+        if (this.is_preconnect_only) {
+            this.onPreconnect(is_ssl, socket);
+            return;
+        }
     }
 
     switch (this.state.request_stage) {
--- a/src/http.zig
+++ b/src/http.zig
@@ -1591,8 +1591,8 @@ signals: Signals = .{},
 async_http_id: u32 = 0,
 hostname: ?[]u8 = null,
 reject_unauthorized: bool = true,
-
 unix_socket_path: JSC.ZigString.Slice = JSC.ZigString.Slice.empty,
+is_preconnect_only: bool = false,
 
 pub fn deinit(this: *HTTPClient) void {
     if (this.redirect.len > 0) {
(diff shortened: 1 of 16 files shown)


src/http.zig
```diff
@@ -1408,6 +1408,7 @@ pub const InternalState = struct {
     received_last_chunk: bool = false,
```
cirospaciari, line 1408: We probably should have `flags` instead of 5 booleans in a followup/refactor

src/url.zig
```diff
@@ -130,7 +130,7 @@ pub const URL = struct {
     }
 
     pub fn hasValidPort(this: *const URL) bool {
-        return (this.getPort() orelse 0) > 1;
+        return (this.getPort() orelse 0) > 0;
```
cirospaciari, line 133: nice catch

src/http.zig
```diff
@@ -2368,12 +2417,37 @@ fn printResponse(response: picohttp.Response) void {
     Output.flush();
 }
 
+pub fn onPreconnect(this: *HTTPClient, comptime is_ssl: bool, socket: NewHTTPContext(is_ssl).HTTPSocket) void {
+    log("onPreconnect({})", .{this.url});
+    _ = socket_async_http_abort_tracker.swapRemove(this.async_http_id);
+    const ctx = if (comptime is_ssl) &http_thread.https_context else &http_thread.http_context;
+
+    ctx.releaseSocket(
```
cirospaciari, line 2433: Can we add a test when we preconnect, but we close the connection on the server part before we do a fetch request?

src/http.zig
```diff
@@ -2368,12 +2417,37 @@ fn printResponse(response: picohttp.Response) void {
...
+}
+
 pub fn onWritable(this: *HTTPClient, comptime is_first_call: bool, comptime is_ssl: bool, socket: NewHTTPContext(is_ssl).HTTPSocket) void {
     if (this.signals.get(.aborted)) {
         this.closeAndAbort(is_ssl, socket);
         return;
     }
 
+    if (this.is_preconnect_only) {
+        this.onPreconnect(is_ssl, socket);
```
cirospaciari: should we call onPreconnect inside `onOpen` instead? we are sending a GET request or just connecting? What happens if the GET request have a big body? and whats happens if the server is not writable before we send the first request content? If is pre-flight should it be a HEAD request?
We need tests for this scenarios.
If we are doing a pre-flight we should wait until the request is complete (aka headers, body etc), if we send another request when still receiving body when reusing the socket we will get a mal formed request and async pipelining is not really supported by a lot of servers.

cirospaciari (changes requested):
some comments

src/http.zig
Jarred-Sumner: yeah

src/http.zig
Jarred-Sumner: preconnect does not write any data. it just keeps the socket open. we don't know what the data they're sending will be, so we can't make any assumptions.

src/http.zig
Jarred-Sumner: yeah, though the rest of the existing keep-alive logic is the same

cirospaciari (approved):

PR #12728, opened 2024-07-22
fix(http) timeout

cirospaciari:
### What does this PR do?

- [x] Code changes

Existing tests


Jarred-Sumner:
Using `markSocketAsDead` in each call site will cause it to never use HTTP keep alive

PR #12729, opened 2024-07-23
fix(bundler): become smarter with `__esm` wrappers

### LINKED ISSUE #4212: Fully minify creation of functions

Code like
```js
function func() {}
```
currently minifies to
```js
var f=function(){}
```

Function creation should minify to
```js
function f(){}
```

This minified code is 4 characters shorter and more closely follows the original code. As an additional optimization, `var f = function () {}` should probably minify to `function f(){}`.

N/A

### LINKED ISSUE #6261: bun build with puppeteer dependency fails

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

Linux 5.4.0-113-generic x86_64 x86_64

Trying to `bun build --target bun` a simple script that uses puppeteer 21.3.6 and then `build run` fails with `TypeError: The superclass is not a constructor.`.

Directly running `bun run index.js` works as expected.
Repro case: https://github.com/EmielM/bun-build-repro

This is probably some issue with puppeteer: if you look at `node_modules/puppeteer-core/lib/esm/puppeteer/api/JSHandle.js`, you see it's pretty convoluted/generated javascript.

But flagging here as
- I think puppeteer is not some marginal library :), so will probably impact others
- shouldn't `bun build --target bun --outfile out.js index.js && bun run out.js` always be behaviourally  identical to `bun run index.js` in all cases?

A successful run (generated chrome.png)

### What do you see instead?

```
52593 |   init_Serializer();
52594 | });
52595 | 
52596 | // node_modules/puppeteer-core/lib/esm/puppeteer/bidi/JSHandle.js
52597 | 
52598 | class BidiJSHandle extends JSHandle {
   ^
TypeError: The superclass is not a constructor.
      at /home/emiel/bun-repro/out.js:52598:0
```

### Additional information

_No response_

### LINKED ISSUE #9387: cannot bundle `frog` with --target=bun

_No response_

_No response_

`bun build --target=bun demo.ts`

1. with require
```
const x = require('./node_modules/frog/protobufs/generated/username_proof_pb.ts');
console.log(x);
```

1. with import star
```
import * as x from './node_modules/frog/protobufs/generated/username_proof_pb.ts';
console.log(x);
```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #11880: Bun's bundler hoists classes before variables they `extends` from in dynamic imports, causing runtime errors

1.1.12

Darwin 23.5.0 arm64 arm

When Bun's bundler tries to inline a dynamic import, it appears to hoist class definitions to the top of the generated file.

A problem arises when classes inherit from values which are not hoisted. In that case, the bundled code will error at runtime due to not finding the thing it needs to inherit from.

Here is an example:

```js
const MyReassignedSuper = class MySuper {
  greet() {
    return 'Hello, world!';
  }
};

class MyInherited extends MyReassignedSuper {};

export default MyInherited;
```

Which if imported like so:

```js
async function hi() {
  const { default: MyInherited } = await import('./other.js');
  const myInstance = new MyInherited();
  console.log(myInstance.greet())
}

hi();
```

Then builds into this:

```js
var __defProp = Object.defineProperty;
var __export = (target, all) => {
  for (var name in all)
    __defProp(target, name, {
      get: all[name],
      enumerable: true,
      configurable: true,
      set: (newValue) => all[name] = () => newValue
    });
};
var __esm = (fn, res) => () => (fn && (res = fn(fn = 0)), res);

// other.js
var exports_other = {};
__export(exports_other, {
  default: () => {
    {
      return other_default;
    }
  }
});

class MyInherited extends MyReassignedSuper {
}
var MyReassignedSuper, other_default;
var init_other = __esm(() => {
  MyReassignedSuper = class MySuper {
    greet() {
      return "Hello, world!";
    }
  };
  other_default = MyInherited;
});

// index.js
async function hi() {
  const { default: MyInherited2 } = await Promise.resolve().then(() => (init_other(), exports_other));
  const myInstance = new MyInherited2;
  console.log(myInstance.greet());
}
hi();
```

When run (in Bun or elsewhere) we get an error like this:

```
23 | class MyInherited extends MyReassignedSuper {
     ^
TypeError: The superclass is not a constructor.
```

While this is an odd thing to do, it does occur in libraries and `esbuild` generates working code. So ideally Bun would do the same.

When `esbuild` builds this project, it generates the following:

```js
(() => {
  var __defProp = Object.defineProperty;
  var __getOwnPropNames = Object.getOwnPropertyNames;
  var __esm = (fn, res) => function __init() {
    return fn && (res = (0, fn[__getOwnPropNames(fn)[0]])(fn = 0)), res;
  };
  var __export = (target, all) => {
    for (var name in all)
      __defProp(target, name, { get: all[name], enumerable: true });
  };

  // other.js
  var other_exports = {};
  __export(other_exports, {
    default: () => other_default
  });
  var MyReassignedSuper, MyInherited, other_default;
  var init_other = __esm({
    "other.js"() {
      MyReassignedSuper = class MySuper {
        greet() {
          return "Hello, world!";
        }
      };
      MyInherited = class extends MyReassignedSuper {
      };
      other_default = MyInherited;
    }
  });

  // index.js
  async function hi() {
    const { default: MyInherited2 } = await Promise.resolve().then(() => (init_other(), other_exports));
    const myInstance = new MyInherited2();
    console.log(myInstance.greet());
  }
  hi();
})();
```

This does run successfully.

### What do you see instead?

_No response_

### Additional information

I have prepared a minimal reproduction here: https://github.com/neoeno/bun-dynamic-import-bug-repro (you can run `./build.sh` to see how Bun handles it and how esbuild handles it).

I narrowed this bug down when trying to use Bun to bundle https://github.com/vidstack/player

My code was something like this:

```js
import { MediaPlayer, MediaProvider } from '@vidstack/react';

export function StoredVideo(props: Readonly<React.VideoHTMLAttributes<HTMLVideoElement> & { ref: React.Ref<HTMLVideoElement> }>) {
  return <MediaPlayer title="Sprite Fight" src="https://files.vidstack.io/sprite-fight/720p.mp4">
    <MediaProvider />
  </MediaPlayer>
}
```

And the code in the library that triggered the problem was this:

```js
// ...

const EVENT = IS_SERVER ? class Event2 {
} : Event, DOM_EVENT = Symbol("DOM_EVENT");

class DOMEvent extends EVENT {
  constructor(type, ...init) {
    super(type, init[0]);
    this[_a$1] = true;
    this.triggers = new EventTriggers();
    this.detail = init[0]?.detail;
    const trigger = init[0]?.trigger;
    if (trigger)
      this.triggers.add(trigger);

// ...
```

The `EVENT` there, since it was a constant which resolved at runtime, ended up quite far down in the bundle whereas the class definition was quite high up.

paperclover:
### What does this PR do?

This PR widens the amount of things we can pull out of the __esm wrapper, as well as fixing extending classes from being pulled out (causing runtime errors due to incorrect evaluation order). It also makes it so that if all items from an esm wrapper can be removed, the wrapper itself is not produced.

Fixes #4212
Fixes #11880
Fixes #9387
Fixes #6261


src/js_parser.zig
```diff
@@ -22063,13 +22070,14 @@ fn NewParser_(
...
                                 if (
                                 // Hoist module-level functions when
-                                ((FeatureFlags.unwrap_commonjs_to_esm and p.current_scope == p.module_scope and !data.func.flags.contains(.is_export)) or
-
-                                    // Manually hoist block-level function declarations to preserve semantics.
-                                    // This is only done for function declarations that are not generators
-                                    // or async functions, since this is a backwards-compatibility hack from
-                                    // Annex B of the JavaScript standard.
-                                    !p.current_scope.kindStopsHoisting()) and p.symbols.items[data.func.name.?.ref.?.innerIndex()].kind == .hoisted_function)
+                                // ((FeatureFlags.unwrap_commonjs_to_esm and p.current_scope == p.module_scope and !data.func.flags.contains(.is_export)) or
```
dylan-conway, line 22073: this is meant to be commented?

dylan-conway:
I think tests in `transpiler.test.js` and `snapshot.test.ts` just need to be updated

src/bundler/bundle_v2.zig
```diff
@@ -8806,127 +8794,138 @@ const LinkerContext = struct {
                     // isn't async because then calling "require()" on that module would
                     // swallow any exceptions thrown during module initialization.
                     const is_async = flags.is_async_or_has_async_dependency;
-                    const Hoisty = struct {
-                        decls: std.ArrayList(G.Decl),
+
+                    std.debug.print("aaa", .{});
```
dylan-conway:
```suggestion
```

src/bundler/bundle_v2.zig
```diff
@@ -8806,127 +8794,138 @@ const LinkerContext = struct {
...
+
+                    const ExportHoist = struct {
+                        decls: std.ArrayListUnmanaged(G.Decl),
                         allocator: std.mem.Allocator,
```
dylan-conway, line 8790: can we remove allocator and pass through `wrapIdentifier`?

src/js_parser.zig
paperclover: no, i meant to completely remove it.

PR #12741, opened 2024-07-23
libdeflate 

Jarred-Sumner:
### What does this PR do?

- [ ] Code changes


src/bun.js/rare_data.zig
```diff
@@ -392,3 +392,14 @@ pub fn nodeFSStatWatcherScheduler(rare: *RareData, vm: *JSC.VirtualMachine) *Sta
         return rare.node_fs_stat_watcher_scheduler.?;
     };
 }
+
+pub fn deinit(this: *RareData) void {
```
Jarred-Sumner, line 396: small memory leak fix for worker

packages/bun-types/bun.d.ts
```diff
@@ -3481,6 +3481,13 @@ declare module "bun" {
      * Filtered data consists mostly of small values with a somewhat random distribution.
      */
     strategy?: number;
+
+    library?: "zlib";
```
dylan-conway, line 3485: is `"libdeflate"` the other option here? we should add a comment if so

Jarred-Sumner: typescript autocomplete shows that

src/http.zig
```diff
@@ -1520,40 +1540,93 @@ pub const InternalState = struct {
...
-            if (this.isDone() or error.ShortRead != err) {
-                Output.prettyErrorln("<r><red>Decompression error: {s}<r>", .{bun.asByteSlice(@errorName(err))});
-                Output.flush();
-                return err;
+        var still_needs_to_decompress = true;
+
+        if (FeatureFlags.isLibdeflateEnabled()) {
+            // Fast-path: use libdeflate
+            if (is_final_chunk and !this.is_libdeflate_fast_path_disabled and this.encoding.canUseLibDeflate() and this.isDone()) libdeflate: {
+                this.is_libdeflate_fast_path_disabled = true;
```
dylan-conway, line 1555: why is the fast path disabled here?

src/http.zig
```diff
@@ -3458,6 +3538,7 @@ fn handleResponseBodyChunkedEncodingFromMultiplePackets(
             this.state.received_last_chunk = true;
             _ = try this.state.processBodyBuffer(
                 buffer,
+                true,
```
dylan-conway, line 3541: should this be
```suggestion
                this.state.encoding.isCompressed()
```

dylan-conway: nvm, encoding is checked in `processBodyBuffer`

src/http.zig
Jarred-Sumner: libdeflate doesn't support streaming, so we can only use it once

Issue #12742, opened 2024-07-23
[crash] fs readdir crash


vinhjaxt:
### How can we reproduce the crash?

```
// bun --hot main.ts
import fs from 'node:fs/promises'

const folder = 'does-exists-folder' // yes the folder exists, can list by ls -al does-exists-folder
await fs.readdir(folder, { recursive: true })
```

### Relevant log output

```shell
[preindex] processFile: ENOENT: No such file or directory
   errno: -2
 syscall: "stat"



============================================================
Bun v1.1.20 (ae194892) Linux x64 (baseline)
Linux Kernel v6.1.0 | glibc v2.39
CPU: sse42 popcnt avx avx2
Args: "bun" "--hot" "/preindex.ts"
Features: jsc fetch(6) shell(2) spawn 
Builtins: "bun:main" "node:fs/promises" 
Elapsed: 14385ms | User: 76ms | Sys: 45ms
RSS: 1.07GB | Peak: 94.53MB | Commit: 1.07GB | Faults: 3

panic(main thread): Segmentation fault at address 0x38
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.20/Ba1ae19489AigigG2q06mE+xpRsmml/C6g609C0k6piE4pm48Cwh998C6wsl3Cgt4j3Cy8wKA2AwD
```

### Stack Trace (bun.report)

Bun v1.1.20 ([`ae19489`](<https://github.com/oven-sh/bun/tree/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66>)) on linux x86_64_baseline [AutoCommand]

Segmentation fault at address 0x00000038

- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/ld-temp.o#L0>): `WTF::Unicode::convertReplacingInvalidSequences`
- `??`
- [`unbounded_queue.zig:106`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/src/bun.js/unbounded_queue.zig#L106>): `src.bun.js.event_loop.EventLoop.tick`
[...]

PR #12748, opened 2024-07-23
Fix alignment calculation in Zone.create function

ippsav:
### What does this PR do?

This PR fixes an issue in the `Zone.create` function where the alignment value was incorrectly passed to `rawAlloc`. The changes include:

- Adding explicit calculation of alignment as a usize
- Calculating log2 of alignment using `@ctz`
- Passing log2 of alignment to rawAlloc instead of raw alignment value

- [x] Code changes

### DIFF
--- a/src/heap_breakdown.zig
+++ b/src/heap_breakdown.zig
@@ -97,8 +97,10 @@ pub const Zone = opaque {
 
     /// Create a single-item pointer with initialized data.
     pub inline fn create(zone: *Zone, comptime T: type, data: T) *T {
+        const align_of_t: usize = @alignOf(T);
+        const log2_align_of_t = @ctz(align_of_t);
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you for this. Note that heap breakdown allocator is only used in debug builds on macOS, so it cannot be the fix for this Linux crash happening in the release build

ippsav:
@Jarred-Sumner thanks for pointing that out, I'll spin up a VM and check the issue closely for Linux

ippsav:
@Jarred-Sumner can't seem to reproduce the issue happening in https://github.com/oven-sh/bun/issues/12742 using the same version of bun on Linux, so for now i guess this fix can't get merged, and once i manage to reproduce it i'll make a new PR
Screenshot 2024-07-24 at 03 41 14

PR #12780, opened 2024-07-24
`napi_threadsafe_function` async tracker

dylan-conway:
### What does this PR do?
also:
- lldb helpers for pretty printing JSC and zig types
- fix for index out of bounds when using `%%` with `console.log`
- fix segfault when using `Bun.$` in macros

manually


.lldbinit
dylan-conway, line 1: This isn't in `./scripts` because `../` wasn't working

PR #12792, opened 2024-07-25
fixes relationship between process.kill and process._kill

nektro:
- kill should call _kill
- throw should happen in kill not _kill
- on success kill should return true
- _kill should not throw unless there's not enough arguments
- _kill should behave the same on both posix and windows

for most this change will be invisible but it allows kill to be overridden in the way that the node apis expect
i should add a test before merge

### DIFF
--- a/src/bun.js/bindings/BunProcess.cpp
+++ b/src/bun.js/bindings/BunProcess.cpp
@@ -2715,6 +2715,11 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionReallyKill,
 {
     auto scope = DECLARE_THROW_SCOPE(globalObject->vm());
 
+    if (callFrame->argumentCount() < 2) {
+        throwVMError(globalObject, scope, "Not enough arguments"_s);
+        return {};
+    }
+
     int pid = callFrame->argument(0).toInt32(globalObject);
     RETURN_IF_EXCEPTION(scope, {});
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Sounds like process.kill() should be implemented in JS

Jarred-Sumner:
Since `kill` calls `_kill`, `kill` should be in JS or `kill` should be a property on the ProcessObject and we check if the kill WriteBarrier is set and go from JS instead of native.

Jarred-Sumner (changes requested):
Since `kill` calls `_kill`, `kill` should be in JS or `kill` should be a property on the ProcessObject and we check if the kill WriteBarrier is set and go from JS instead of native.

nektro:
`process.kill` also accepts a string signal name and needs to call `loadSignalNumberMap()` etc. `process._kill` being replaced is an accepted use case and the function as-is in this patch already checks for exceptions

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2766,15 +2770,32 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionKill,
...
     }
 
-#if OS(WINDOWS)
-    int result = uv_kill(pid, signal);
-#else
-    int result = kill(pid, signal);
-#endif
+    auto global = jsCast<Zig::GlobalObject*>(globalObject);
+    auto& vm = global->vm();
+    JSValue _killFn = global->processObject()->get(globalObject, Identifier::fromString(vm, "_kill"_s));
```
Jarred-Sumner, line 2775:
```suggestion
    JSValue _killFn = global->processObject()->get(globalObject, Identifier::fromString(vm, "_kill"_s));
    RETURN_IF_EXCEPTION(scope, {});
```

PR #12798, opened 2024-07-25
test: toRun matcher should check output if optionalStdout is empty string

nektro:

just discovered this when writing some new tests. running on CI on its own to see if it catches anything


nektro:
tests passed, supplanted by https://github.com/oven-sh/bun/pull/12799

PR #12800, opened 2024-07-25
fix(brotli): protect buffer jsvalues

dylan-conway:
TODO: add a test that doesn't take a minute to run

### DIFF
--- a/src/bun.js/api/js_brotli.zig
+++ b/src/bun.js/api/js_brotli.zig
@@ -186,6 +186,7 @@ pub const BrotliEncoder = struct {
                             this.encoder.write_failed = true;
                             return;
                         };
+                        input.deinitAndUnprotect();
                     }
 
                     any = any or pending.len > 0;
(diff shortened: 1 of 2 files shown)


src/bun.js/api/js_brotli.zig
```diff
@@ -16,7 +16,8 @@ pub const BrotliEncoder = struct {
 
     globalThis: *JSC.JSGlobalObject,
 
-    input: Queue = Queue.init(bun.default_allocator),
+    // input: Queue = Queue.init(bun.default_allocator),
+    input: std.ArrayListUnmanaged(JSC.Node.BlobOrStringOrBuffer) = .{},
```
Jarred-Sumner: This should continue to be a FIFO so that we can incrementally read/write

src/bun.js/api/js_brotli.zig
```diff
@@ -179,13 +179,14 @@ pub const BrotliEncoder = struct {
...
                     }
-                    for (pending) |input| {
+                    for (pending) |*input| {
                         var writer = this.encoder.stream.writer(Writer{ .encoder = this.encoder });
                         writer.writeAll(input.slice()) catch {
                             _ = this.encoder.pending_encode_job_count.fetchSub(1, .monotonic);
                             this.encoder.write_failed = true;
                             return;
                         };
+                        input.deinitAndUnprotect();
```
Jarred-Sumner: `deinitAndUnprotect` can only be called from the JavaScript thread.

kelvinwop:
:0 nice

Issue #922, opened 2022-07-30
bun install from symlinked folder


guybarnard:
### Version

0.1.5

### Platform

21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:22 PDT 2022; root:xnu-8020.140.41~1/RELEASE_ARM64_T6000 arm64

### What steps will reproduce the bug?

``` bash
cd ~/projects # a symlinked folder that is actually on another logical drive 
bun create hono ./bun-hono
cd bun-hono
bun install
```

```
Failed to install 2 packages
[9.00ms] done
error: Unexpected installing hono
error: Unexpected installing bun-types
```

### How often does it reproduce? Is there a required condition?

If the packages being installed have not been installed before the error happens every time. 

If the packages have been installed previously (say via the workaround), no error occurs

### What is the expected behavior?

```
bun install v0.1.5
  🔍 Resolving [1/2] 
[49.00ms] git
 + bun-types@0.1.5
 + hono@2.0.5
```

### What do you see instead?

```
Failed to install 2 packages
[9.00ms] done
error: Unexpected installing hono
error: Unexpected installing bun-types
```

## Workaround 1

``` bash
bun install --backend=copyfile
```

## Workaround 2 

Run directly from a folder on the os volume, not a symlinked folder

``` bash
cd ~
mkdir bun-projects
cd bun-projects
bun create hono bun-hono
cd bun-hono
bun install
```

Electroid:
Yes, we need to fix this. Tracking the issue here: #362

PR #12808, opened 2024-07-25
bundler: Add `--ignore-dce-annotations`, and other DCE annotation related stuff

paperclover:
### What does this PR do?

- Add `--ignore-dce-annotations`, one step close to esbuild compatibility
- Print `/* @__PURE */` when not minifying whitespace. This is helpful if the output is consumed by another build tool.
- Add `--emit-dce-annotations` to allow emitting annotations even when minifying.

The second and third cases are things I specifically want for future changes to our node_fallbacks code. Right now we use esbuild for that, when we really should use Bun. An issue with using Bun is that DCE comments will get stripped, resulting in no tree shaking when Bun re-bundles them.

Esbuild also has this issue since we are passing --minify to compress the size. In the future, we can pass `--minify` and `--emit-annotations` and get both.


src/cli.zig
```diff
@@ -185,6 +185,7 @@ pub const Arguments = struct {
         clap.parseParam("--jsx-fragment <STR>              Changes the function called when compiling JSX fragments") catch unreachable,
         clap.parseParam("--jsx-import-source <STR>         Declares the module specifier to be used for importing the jsx and jsxs factory functions. Default: \"react\"") catch unreachable,
         clap.parseParam("--jsx-runtime <STR>               \"automatic\" (default) or \"classic\"") catch unreachable,
+        clap.parseParam("--ignore-annotations             Ignore tree-shaking annotations such as @__PURE__") catch unreachable,
```
Jarred-Sumner: > --ignore-annotation

To someone who doesn't know much about bundlers, this  CLI flag isn't clear.  What is an annotation? A TypeScript type? A decorator? A JSDoc comment? A license comment?

Why not `--ignore-dce-annotation`? Or maybe `--ignore-pure-comments`

Electroid: Bikeshed the new name, should still alias to support what esbuild does: https://esbuild.github.io/api/#ignore-annotations

paperclover: I think it is clear if you read the description of the option: `Ignore tree-shaking annotations such as @__PURE__`

The key words being `tree-shaking` and `@__PURE__`. If that isn't enough, a single google search will take you to a page explaining it.

image

I did not add anything to our docs in this PR explaining these flags.

I would be extremely surprised if someone reading a line talking about `tree-shaking annotations` to misinterpret it as anything related to type definitions (bun already ignores/does not check types), jsdoc (comments are also ignored), and decorators (not sure what ignoring decorators would even do).

> Why not --ignore-dce-annotation?

The flag was chosen to be the same as esbuild. I would be okay with changing it to `--ignore-dce-annotations`.

> `--ignore-pure-comments`

Do not recommend this naming since `@__PURE__` is not the only form of annotations. One day I want to support `@__NO_SIDE_EFFECTS__`. "dce annotations" (or simply annotations) is a more correct term.

Jarred-Sumner: esbuild CLI compatibility is not a goal imo

Jarred-Sumner (changes requested):
tests failing on windows

robobun:
@paperdave, your commit b6a528f8d32a5eb855966dc66972bb5c967f3807 is building: [`#922`](https://buildkite.com/bun/bun/builds/922)

src/js_printer.zig
```diff
@@ -544,7 +546,7 @@ pub const Options = struct {
 
     module_type: options.OutputFormat = .preserve,
 
-    /// Used for cross-module inlining of import items when bundling
+    // /// Used for cross-module inlining of import items when bundling
```
Jarred-Sumner, line 549:
```suggestion
    // Used for cross-module inlining of import items when bundling
```

PR #12809, opened 2024-07-25
node: more process.exitCode fixes

nektro:

adds more tests, fixes them, and fixes ci failures from https://github.com/oven-sh/bun/pull/12799

### DIFF
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -2088,7 +2088,7 @@ pub const Process = struct {
 
         vm.exit_handler.exit_code = code;
         vm.onExit();
-        bun.Global.exit(vm.exit_handler.exit_code);
+        vm.globalExit();
     }
 
     pub export const Bun__version: [*:0]const u8 = "v" ++ bun.Global.package_json_version;
--- a/src/bun.js/node/types.zig
+++ b/src/bun.js/node/types.zig
@@ -2086,8 +2086,9 @@ pub const Process = struct {
             return;
         }
 
+        vm.exit_handler.exit_code = code;
         vm.onExit();
-        bun.Global.exit(code);
+        bun.Global.exit(vm.exit_handler.exit_code);
     }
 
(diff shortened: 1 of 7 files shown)


Jarred-Sumner:
tests failing on Windows

src/bun.js/node/types.zig
```diff
@@ -2086,8 +2086,9 @@ pub const Process = struct {
             return;
         }
 
+        vm.exit_handler.exit_code = code;
         vm.onExit();
-        bun.Global.exit(code);
+        bun.Global.exit(vm.exit_handler.exit_code);
```
Jarred-Sumner: What was the purpose of adding the extra pointer lookup here? If it's to account for the user changing the exit code inside of onExit (assuming Node supports that), then maybe this should be moved into a method on vm?

src/bun_js.zig
```diff
@@ -504,5 +503,5 @@ noinline fn dumpBuildError(vm: *JSC.VirtualMachine) void {
 noinline fn failWithBuildError(vm: *JSC.VirtualMachine) noreturn {
     @setCold(true);
     dumpBuildError(vm);
-    Global.exit(1);
+    Global.exit(vm.exit_handler.exit_code);
```
Jarred-Sumner: Is this right? Do any tests verify that a build error does not exit with code 0?

Jarred-Sumner: I'm not sure how to trigger the errors here actually

src/bun_js.zig
```diff
@@ -458,12 +458,10 @@ pub const Run = struct {
                 .{Global.unhandled_error_bun_version_string},
             );
         }
-        const exit_code = this.vm.exit_handler.exit_code;
 
         vm.onExit();
```
Jarred-Sumner, line 462: calling vm.onExit twice.

Jarred-Sumner: nevermind

Issue #12807, opened 2024-07-25
bun crash


VinayTech3d:
### Stack Trace (bun.report)

Bun v1.1.20 ([`ae19489`](<https://github.com/oven-sh/bun/tree/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66>)) on windows x86_64 [AutoCommand]

**panic**: attempt to use null value

- [`javascript.zig:847`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/src/bun.js/javascript.zig#L847>): `get`
- [`ZigGlobalObject.cpp:564`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/src/bun.js/bindings/ZigGlobalObject.cpp#L564>): `computeErrorInfoWithoutPrepareStackTrace`
- [`Function.h:53`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/C:/a/bun/bun/build/bun-webkit/include/wtf/Function.h#L53>): `WTF::Detail::CallableWrapper<WTF::String (...)(...),WTF::String,JSC::VM &,WTF::Vector<JSC::StackFrame,0,WTF::CrashOnOverflow,16,WTF::FastMalloc> &,unsigned int &,unsigned int &,WTF::String &,JSC::JSObject *>::call`
- [`utility:756`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.40.33807/include/utility#L756>): `std::exchange`
- [`Heap.cpp:713`](<https://github.com/oven-sh/bun/blob/ae19489250ab74ee3ea7d5b50fca5d7d2b4e6d66/C:/a/WebKit/WebKit/Source/JavaScriptCore/heap/Heap.cpp#L713>): `JSC::Heap::finalizeMarkedUnconditionalFinalizers<JSC::ErrorInstance,JSC::IsoSubspace>::<lambda_1>::operator()`
[...]

PR #12810, opened 2024-07-25
Fix crash when creating a new Worker with a numeric environment varia…

Jarred-Sumner:
### What does this PR do?

Probably fixes https://github.com/oven-sh/bun/issues/12807

Also should make Bun start about 0.5ms faster due to not creating a JSGlobalObject for a prototype.

Added a test

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -72,3 +72,7 @@ pub const overrides = struct {
         }
     };
 };
+
+pub export fn Bun__panic(msg: [*]const u8, len: usize) noreturn {
+    Output.panic("{s}\n", .{msg[0..len]});
+}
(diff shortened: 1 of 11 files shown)


src/main.zig
```diff
@@ -72,3 +72,7 @@ pub const overrides = struct {
         }
     };
 };
+
+pub export fn Bun__panic(msg: [*]const u8, len: usize) noreturn {
+    Output.panic("{s}\n", .{msg[0..len]});
```
dylan-conway: We don't include a newline with other uses of `Output.panic`. Is this one needed?

Jarred-Sumner: no

Jarred-Sumner:
```suggestion
    Output.panic("{s}", .{msg[0..len]});
```

dylan-conway (approved):

Issue #1002, opened 2022-08-06
cannot run next app when upgrade to 0.1.7


dawnswwwww:
### Version

0.1.7

### Platform

Darwin C02G86LCMD6R 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64

### What steps will reproduce the bug?

```shell
bun create next ./next-app

cd next-app

bun dev
```

### What do you see instead?

![image](https://user-images.githubusercontent.com/49851272/183248306-5e7fae5e-0a35-474c-95a8-6831c0d188d4.png)

Jarred-Sumner:
Working on a fix

xhyrom:
Already fixed, install canary.

PR #12821, opened 2024-07-25
Start implementing internal V8 APIs

190n:
### What does this PR do?

- [x] Code changes

Implements some of the V8 APIs that native modules written especially for Node.js use. Some have real implementations while some have stubs that just crash, but even those will let us see which APIs are missing because it will print which method failed instead of the generic `dyld: missing symbol called` on macOS.

Most Node.js modules still won't fully load because `node_module_register` isn't fully implemented.

TODOs before we can reasonably expect some modules to work (at least based on `cpu-features`'s usage):

- [ ] Implement module lifecycle properly
- [ ] Stub `HandleScope` and `EscapableHandleScope` without aborting
- [ ] Implement `ObjectTemplate` and `FunctionTemplate`
- [ ] Implement internal fields on objects

I wrote automated tests


src/bun.js/bindings/v8/EscapableHandleScopeBase.cpp
```diff
@@ -0,0 +1,16 @@
...
+namespace v8 {
+
+EscapableHandleScopeBase::EscapableHandleScopeBase(Isolate* isolate)
+    : HandleScope(isolate)
+{
+}
+
+uintptr_t* EscapableHandleScopeBase::EscapeSlot(uintptr_t* escape_value)
+{
+    assert("EscapableHandleScopeBase::EscapeSlot" && 0);
```
Jarred-Sumner:
```suggestion
    ASSERT_NOT_REACHED();
```

src/bun.js/bindings/v8/Object.cpp
```diff
@@ -0,0 +1,58 @@
...
+    JSFinalObject* object = JSC::constructEmptyObject(isolate->globalObject());
+    JSValue jsv(object);
+    return Local<Object>(jsv);
+}
+
+Maybe<bool> Object::Set(Local<Context> context, Local<Value> key, Local<Value> value)
+{
+    JSGlobalObject* globalObject = *context;
+    JSObject* object = toJSValue().getObject();
+    assert(object);
```
Jarred-Sumner: Note that `getObject()` does this assertion internally. It will crash at runtime if it's not an object.

src/bun.js/bindings/v8/String.cpp
```diff
@@ -0,0 +1,80 @@
+#include "v8/String.h"
+
+using JSC::JSString;
+using JSC::JSValue;
+
+namespace v8 {
+
+MaybeLocal<String> String::NewFromUtf8(Isolate* isolate, char const* data, NewStringType type, int signed_length)
+{
+    (void)type;
```
Jarred-Sumner: This may need to be used

The JSC equivalent of "Intern" is `AtomString`.

src/bun.js/bindings/v8/String.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return MaybeLocal<String>(local);
+}
+
+int String::WriteUtf8(Isolate* isolate, char* buffer, int length, int* nchars_ref, int options) const
+{
+    assert(options == 0);
+    auto jsValue = toJSValue();
+    WTF::String string = jsValue.getString(isolate->globalObject());
+
+    if (!string.is8Bit()) {
```
Jarred-Sumner: I assume this is throw away debug code

src/bun.js/bindings/v8/String.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // ReplacingInvalidSequences matches how v8 behaves here
+    auto string = WTF::String::fromUTF8ReplacingInvalidSequences(span);
+    assert(!string.isNull());
+    auto jsString = JSC::jsString(isolate->vm(), string);
+    JSValue jsValue(jsString);
+    Local<String> local(jsValue);
+    return MaybeLocal<String>(local);
+}
+
+int String::WriteUtf8(Isolate* isolate, char* buffer, int length, int* nchars_ref, int options) const
```
Jarred-Sumner: This is one method we will want to optimize carefully

src/bun.js/bindings/ZigGlobalObject.h
```diff
@@ -237,6 +237,8 @@ class GlobalObject : public JSC::JSGlobalObject {
     JSC::JSObject* NodeVMScript() const { return m_NodeVMScriptClassStructure.constructorInitializedOnMainThread(this); }
     JSC::JSValue NodeVMScriptPrototype() const { return m_NodeVMScriptClassStructure.prototypeInitializedOnMainThread(this); }
 
+    JSC::Structure* ObjectTemplateStructure() const { return m_ObjectTemplateStructure.getInitializedOnMainThread(this); }
```
Jarred-Sumner: This should go on a V8 JSCell* object instead of the Zig::GlobalObject directly.

src/bun.js/bindings/v8/InternalFieldObject.h
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        InternalField(void* raw)
+            : data({ .raw = raw })
+            , is_js_value(false)
+        {
+        }
+    };
+
+    WTF::Vector<InternalField>* internalFields() { return &fields; }
```
Jarred-Sumner: This will need it's own isosubspace. As-is, this probably doesn't work without creating a new JSC::Structure every call

src/bun.js/bindings/v8/Object.cpp
```diff
@@ -0,0 +1,81 @@
...
+namespace v8 {
+
+static WTF::Vector<InternalFieldObject::InternalField>* getInternalFieldsContainer(Object* object)
+{
+    JSObject* js_object = object->toJSValue().getObject();
+    const JSC::ClassInfo* class_info = js_object->classInfo();
+
+    // TODO(@190n): do we need to unwrap proxies like node-jsc did?
+
+    if (class_info->isSubClassOf(InternalFieldObject::info())) {
```
Jarred-Sumner:
```suggestion
    if (object->inherits<InternalFieldObject>()) {
```

src/bun.js/bindings/v8/Object.cpp
```diff
@@ -0,0 +1,81 @@
...
+using JSC::JSObject;
+using JSC::JSValue;
+using JSC::PutPropertySlot;
+
+namespace v8 {
+
+static WTF::Vector<InternalFieldObject::InternalField>* getInternalFieldsContainer(Object* object)
+{
+    JSObject* js_object = object->toJSValue().getObject();
+    const JSC::ClassInfo* class_info = js_object->classInfo();
```
Jarred-Sumner: ✂️ 
```suggestion

```

robobun:
@190n, your commit bd84bd7a41abbde3539183b91ef9391f69d52111 is building: [`#1002`](https://buildkite.com/bun/bun/builds/1002)

src/bun.js/bindings/v8/v8.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include "ZigGlobalObject.h"
+
+#if defined(WIN32) || defined(_WIN32)
+#define BUN_EXPORT __declspec(dllexport)
+#else
+#define BUN_EXPORT JS_EXPORT
+#endif
+
+#define V8_UNIMPLEMENTED() BUN_PANIC("You're using a module which calls a V8 function that Bun does not yet implement. Track progress at issue 4290.")
```
Electroid: Let's include the issue URLs for any panics like these, so it's easier

Jarred-Sumner:
Could the filenames for V8 things be prefixed with V8? Like "V8Function.cpp" instead of "Function.cpp" 

I think it would get confusing for other parts of the codebase that when you type "Object" in cmd + p it shows the `V8::Object` class

src/bun.js/bindings/v8/V8String.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+int String::WriteUtf8(Isolate* isolate, char* buffer, int length, int* nchars_ref, int options) const
+{
+    RELEASE_ASSERT(options == 0);
+    auto jsString = localToObjectPointer<JSString>();
+    WTF::String string = jsString->getString(isolate->globalObject());
+
+    // TODO(@190n) handle 16 bit strings
+    RELEASE_ASSERT(string.is8Bit());
```
Jarred-Sumner: I don't think its shippable without UTF-16 strings, but I can help with this

src/bun.js/bindings/napi.cpp
```diff
@@ -341,6 +342,7 @@ class NAPIFunction : public JSC::JSFunction {
         NAPICallFrame frame(JSC::ArgList(args), function->m_dataPtr);
 
         auto scope = DECLARE_THROW_SCOPE(vm);
+        v8::HandleScope handleScope(v8::Isolate::fromGlobalObject(static_cast<Zig::GlobalObject*>(globalObject)));
```
Jarred-Sumner, line 345: I think this can technically be a node:vm global object but I don't think we handle this correctly elsewhere either

src/bun.js/bindings/v8/V8Context.h
```diff
@@ -0,0 +1,34 @@
...
+// Context is always a reinterpret pointer to V8::Roots, so that inlined V8 functions can find
+// values they expect to find at fixed offsets
+class Context : public Data {
+public:
+    JSC::VM& vm() const
+    {
+        return globalObject()->vm();
+    }
+
+    const Zig::GlobalObject* globalObject() const
```
Jarred-Sumner, line 18: We should probably use `Bun::GlobalScope` for these, honestly

Jarred-Sumner: but it's not a blocker

src/bun.js/bindings/v8/V8Data.h
```diff
@@ -0,0 +1,102 @@
...
+    JSC::JSValue localToJSValue(GlobalInternals* globalInternals) const
+    {
+        TaggedPointer root = *reinterpret_cast<const TaggedPointer*>(this);
+        if (root.type() == TaggedPointer::Type::Smi) {
+            return JSC::jsNumber(root.getSmiUnchecked());
+        } else {
+            void* raw_ptr = root.getPtr<void>();
+            // check if this pointer is identical to the fixed locations where these primitive
+            // values are stored
+            if (raw_ptr == globalInternals->undefinedSlot()->getPtr<void>()) {
```
Jarred-Sumner, line 48: interesting that literally everything is a pointer

src/bun.js/bindings/v8/V8External.cpp
```diff
@@ -0,0 +1,26 @@
+#include "V8External.h"
+#include "V8HandleScope.h"
+
+#include "napi_external.h"
+
+namespace v8 {
+
+Local<External> External::New(Isolate* isolate, void* value)
```
Jarred-Sumner, line 8: lol v8 external wraps napi external 

backwards

src/bun.js/bindings/v8/V8Function.cpp
```diff
@@ -0,0 +1,64 @@
...
+    CREATE_METHOD_TABLE(Function)
+};
+
+Structure* Function::createStructure(JSC::VM& vm, JSC::JSGlobalObject* globalObject)
+{
+    return Structure::create(
+        vm,
+        globalObject,
+        globalObject->functionPrototype(),
+        JSC::TypeInfo(JSC::InternalFunctionType, StructureFlags),
```
Jarred-Sumner, line 29: Is this meant to be a constructor? I think InternalFunctionType is fine. But if we can use JSFunction it is generally faster than InternalFunction (i am semi-mindlessly repeating something I saw in a PR review on webkit's github)

src/bun.js/bindings/v8/V8FunctionTemplate.h
```diff
@@ -0,0 +1,155 @@
...
+public:
+    FunctionCallbackInfo(ImplicitArgs* implicit_args_, TaggedPointer* values_, int length_)
+        : implicit_args(implicit_args_)
+        , values(values_)
+        , length(length_)
+    {
+    }
+};
+
+using FunctionCallback = void (*)(const FunctionCallbackInfo<Value>&);
```
Jarred-Sumner, line 45: Does this need a specified calling convention?

src/bun.js/bindings/v8/V8HandleScopeBuffer.h
```diff
@@ -0,0 +1,59 @@
...
+#include "v8.h"
+#include "V8TaggedPointer.h"
+#include "V8Map.h"
+#include "V8Handle.h"
+
+namespace v8 {
+
+// An array used by HandleScope to store the items. Must keep pointer stability when resized, since
+// v8::Locals point inside this array.
+class HandleScopeBuffer : public JSC::JSNonFinalObject {
```
Jarred-Sumner, line 12: Can this be a JSC::JSCell instead?

src/bun.js/bindings/v8/V8Function.cpp
190n: I think it can be a constructor but I don't support that yet.

src/bun.js/bindings/v8/V8Roots.h
```diff
@@ -0,0 +1,32 @@
...
+namespace v8 {
+
+class GlobalInternals;
+
+// Container for some data that V8 expects to find at certain offsets. Isolate and Context pointers
+// actually point to this object. It is a separate struct so that we can use offsetof() to make sure
+// the layout is correct.
+struct Roots {
+    // v8-internal.h:775
+    static const int kUndefinedValueRootIndex = 4;
```
Jarred-Sumner, line 14: Maybe we could use `constexpr` for these?

src/bun.js/bindings/v8/V8String.cpp
```diff
@@ -0,0 +1,76 @@
...
+        to_copy = span.size();
+    } else if (to_copy > span.size()) {
+        to_copy = span.size();
+    } else if (length < span.size()) {
+        to_copy = length;
+        terminate = false;
+    }
+    // TODO(@190n) span.data() is latin1 not utf8, but this is okay as long as the only way to make
+    // a v8 string is NewFromUtf8. that's because NewFromUtf8 will use either all ASCII or all UTF-16.
+    memcpy(buffer, span.data(), to_copy);
```
Jarred-Sumner, line 58: Can `to_copy` be zero?

src/bun.js/bindings/v8/V8FunctionTemplate.h
190n: I don't think so. It's only called by native code (inside `FunctionTemplate::functionCall`) not JS, and V8's definition doesn't specify a calling convention. On windows, the compiler should adapt from SysV to MSVC calling convention for us inside `FunctionTemplate::functionCall`.

src/bun.js/bindings/v8/V8String.cpp
190n: Yes. I'm working on UTF-16 right now in this function so this path is probably gonna change a bit

src/bun.js/bindings/v8/V8HandleScopeBuffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+template<typename Visitor>
+void HandleScopeBuffer::visitChildrenImpl(JSCell* cell, Visitor& visitor)
+{
+    HandleScopeBuffer* thisObject = jsCast<HandleScopeBuffer*>(cell);
+    ASSERT_GC_OBJECT_INHERITS(thisObject, info());
+    Base::visitChildren(thisObject, visitor);
+
+    for (int i = 0; i < thisObject->size; i++) {
```
Jarred-Sumner, line 30: this should be `unsigned` or `size_t`

src/bun.js/bindings/v8/V8HandleScopeBuffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+}
+
+DEFINE_VISIT_CHILDREN(HandleScopeBuffer);
+
+Handle& HandleScopeBuffer::createUninitializedHandle()
+{
+    RELEASE_ASSERT(size < capacity - 1);
+    int index = size;
```
Jarred-Sumner, line 43:
```suggestion
    unsigned index = size;
```

src/bun.js/bindings/v8/V8HandleScopeBuffer.cpp
```diff
@@ -0,0 +1,63 @@
...
+            JSCell::visitChildren(reinterpret_cast<JSCell*>(handle.ptr), visitor);
+        }
+    }
+}
+
+DEFINE_VISIT_CHILDREN(HandleScopeBuffer);
+
+Handle& HandleScopeBuffer::createUninitializedHandle()
+{
+    RELEASE_ASSERT(size < capacity - 1);
```
Jarred-Sumner, line 42:
```suggestion
    ASSERT(size <= capacity);
```

src/bun.js/bindings/v8/V8HandleScopeBuffer.cpp
Jarred-Sumner: and we should copy the size into a local variable to avoid the extra pointer lookup

src/bun.js/bindings/v8/V8HandleScopeBuffer.h
```diff
@@ -0,0 +1,59 @@
...
+    DECLARE_VISIT_CHILDREN;
+
+    friend class EscapableHandleScopeBase;
+
+private:
+    // TODO make resizable
+    static constexpr int capacity = 64;
+
+    Handle storage[capacity];
+    int size = 0;
```
Jarred-Sumner, line 49:
```suggestion
    unsigned size = 0;
```

Jarred-Sumner (approved):

PR #12837, opened 2024-07-26
implement node:util.getSystemErrorName()

nektro:

includes new test coverage that should all pass


Jarred-Sumner:
Serializing IRL discussion:
- The error methods should be done in C++ to ensure we don't run into situations where the JSString is collected before the Error instance is finished being created. Also because the current implementation would involve several Latin1 -> UTF-8  -> Latin1 string conversions which are unnecessary. Also because the usage of ZigString.slice() is incorrect, it would produce invalid data when the String is a UTF-16 string
- The system errno map should follow the convention we have in darwin_c.zig, windows_c.zig, and linux_c.zig as those already have mostly the same data.

Issue #1128, opened 2022-08-22
bun.sh/install should send `Content-Type: text/plain`, but sends `Content-Type: text/html`


mortie:
### Affected URL(s)

https://bun.sh/install

### Description of the problem

The "Show script source" button on the website links to https://bun.sh/install. This file is served with the HTTP header `Content-Type: text/html`. This means that web browsers try to render it as HTML, and you end up with an unreadable mess where all whitespace is collapsed into a single space. It should be served with `Content-Type: text/plain` to make browsers render it as plain text.

andrewbrey:
I want to chime in here expand on _why this is important_ and why, in my opinion, this should be given some priority to resolve.

Today, the top-line recommended way to install `bun` from `*nix` systems is to curl this install script and pipe it to your shell. You don't have to look around too much to find advice that doing this type of install is ill-advised if you don't implicitly trust the script author and the integrity of their servers.

The very minimum due diligence for a user intending to install via this script is to open up the script and read it prior to running in the shell, and since you're serving this file from the same domain as the website, and the link on the website opens up a tab with the script source to aid the user in trusting the author, the resulting page **should be _easily_ human readable** (for any human that speaks some `bash` anyway).

The lack of typical script formatting on this webpage is a barrier to basic security due diligence, and that is in my opinion, a bad look for `bun`.

I don't know anything about the server setup for the Bun website in the slightest, so it's perhaps completely out of band to comment on the level of effort at all, but I would be kind of shocked to learn that changing the response content type for this route was a big lift.

For what it's worth, the _yes, very simple_ workarounds for those interested in quickly seeing the formatted source of this script are to either:

1. just `curl` https://bun.sh/install instead of visiting it in the browser
2. read the source of the script at https://github.com/oven-sh/bun/blob/681f5a521f6bb8e3a5d1b0f7bff96cdb55984381/src/cli/install.sh

Again, this is just my two cents, and those two cents mostly speak to the optics of this bug, but I figured I would expand on the issue a bit with those two cents anyway :smile:

Electroid:
Yes, we recognize it's important that the source is readable. Unfortunately, this is a limitation of GitHub pages, where we are currently hosting the docs. This will change soon and this will be fixed.

Electroid:
The headers have been fixed, browsers will download it as an attachment, which is in-line with other install scripts.

```
< content-type: application/x-sh
< cf-ray: 81b5b5bf7faa679a-SJC
[...]

PR #12842, opened 2024-07-26
bindings: init Arguments members to .undefined instead of .zero

nektro:

this makes it more closely mirror js' behavior at runtime such that the values accessed through `.ptr` beyond what are passed are real but `undefined`


robobun:
@nektro, your commit f82d37355fce597d3cd4c9d7d2456f2077239ddc is building: [`#1128`](https://buildkite.com/bun/bun/builds/1128)

nektro:
superseded by new `.argumentsUndef()` method

PR #12843, opened 2024-07-26
ZigString.static() doesnt need to make an explicit global variable

nektro:


Jarred-Sumner:
discussed IRL but this code requires more thinking than the previous and more thinking == more mistakes

Issue #12846, opened 2024-07-26
Bun Memory Leak when `fetch` used with `proxy`


Zamion101:
### What version of Bun is running?

1.1.20+ae1948925

### What platform is your computer?

Linux 5.15.153.1-microsoft-standard-WSL2 x86_64 unknown

### What steps can reproduce the bug?

Here is a Reproduction of the bug in simple steps with useful informations.

```ts
const USER_AGENT = 'Mozilla/5.0 (Linux; Android 9; SAMSUNG SM-G975F Build/PPR1.180610.011) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/9.4 Chrome/67.0.3396.87 Mobile Safari/537.36'
const PROXY_URL = '<HTTP_PROXY_URL>'

export async function proxyRequest(url: string) {

  const options = {
      proxy: PROXY_URL
  }
  const headers: Record<string, string> = {}
  headers['user-agent'] = USER_AGENT
  headers['accept'] = 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7'
  headers['accept-language'] = 'tr-TR,tr;q=0.9,en-US;q=0.8,en;q=0.7'
  headers['cache-control'] = 'no-cache'
  headers['pragma'] = 'no-cache'

  return await fetch(url, {
    method: 'GET',
    headers: headers,
    // proxy: options.proxy, // Comment out to see expected behavior
    keepalive: false,
  })

}

let requestMade = 0
let peakMemoryUsage = 0
const startMemory = (process.memoryUsage.rss() / 1024 / 1024) | 0
for (let a = 0; a < 10; a++) {
  const startIteration = (process.memoryUsage.rss() / 1024 / 1024) | 0
  const requestCountStart = requestMade
  for (let b = 0; b < 10; b++) {
    //542KB https://media.4-paws.org/5/b/8/b/5b8bca3f74342210ccca652c651f2d7e23288753/VIER%20PFOTEN_2019-12-13_209-2890x2000.jpg
    //1.2MB https://images.pexels.com/photos/774731/pexels-photo-774731.jpeg?cs=srgb&dl=pexels-kerber-774731.jpg&fm=jpg
    let beforeRequest = (process.memoryUsage.rss() / 1024 / 1024) | 0
[...]

Issue #6645, opened 2023-10-21
Memory usage keeps growing when using fetch with proxy


tkhg:
### What version of Bun is running?

1.0.7

### What platform is your computer?

arm64

### What steps can reproduce the bug?

run this code:
```typescript
async function main() {
  while (true) {
    const res = await fetch('https://bun.sh', {
      proxy: 'http://username:password@hostname:port'
    })

    await res.arrayBuffer() // consume the body to avoid leak (another known issue)

    console.log(
      new Date().toLocaleTimeString('en-GB'),
      'Memory usage: ',
      Math.trunc(process.memoryUsage.rss() / 1024 / 1024),
      'MB'
    )

    await Bun.sleep(500)
  }
}

await main()
```

### What is the expected behavior?

memory usage should be stabilize after some time

### What do you see instead?

memory usage keeps growing
![image](https://github.com/oven-sh/bun/assets/28704185/54cd93ad-89ee-4eef-b232-ea4f78c7ff76)

guxuanye:
I have encountered the same problem as you. Although I saw on the official blog of Bun that the fetch function has been fixed for memory leakage due to gzip response issues in version 1.0.5, I still encountered this issue in the latest version: as the number of requests continues to increase (in the thousands), memory usage is getting larger and larger.
[Key point] When I delete "Accept Encoding: gzip/deflate" in the request header, there will be no memory leak, which means I suspect that the memory leak caused by gzip response has not been fixed. @Jarred-Sumner

crooowned:
bump this, i have the same problem

Jarred-Sumner:
Please let us know if you're still running into this after [Bun v1.2.14](https://bun.sh/blog/bun-v1.2.14) or later

PR #12850, opened 2024-07-26
Expose fewer details of the HTTP client outside the HTTP thread

Jarred-Sumner:
### What does this PR do?

This is a WIP PR to make the HTTP client more reliable by exposing less of the implementation details to code using it.  Instead of exposing the `AsyncHTTP` struct we only expose a ref-counted `*PendingHTTPRequest` pointer

### DIFF
--- a/misctools/http_bench.zig
+++ b/misctools/http_bench.zig
@@ -192,7 +192,7 @@ pub fn main() anyerror!void {
     try channel.buffer.ensureTotalCapacity(args.count);
 
     try NetworkThread.init();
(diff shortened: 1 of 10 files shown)


Zamion101:
Is this in any way related to issue #12846 or #6645 ?

Jarred-Sumner:
yes

Zamion101:
Is it possible to get an Estimated Time for merge of this PR? I know this PR may not be priority and there could be more important bugs/crashes but I assume memory leak issues are also big factor of the stability of Bun. Thank you all for the Hard work you do to make Bun a efficient, faster and more robust competator for Node/Deno and sorry for making it sound like I'm impatient.

Issue #934, opened 2022-07-31
Deployment platform 


h44mkk:
Hey, is there any platform which support Bun runtime for deployment?

theotterlord:
You can deploy it to any platform supporting docker containers. The official containers aren't building rn (see #855) but I've created a repository that builds images with github actions [here](https://github.com/TheOtterlord/docker-bun).

Apart from that, there's no simple way to do it yet. Remember, bun is still very new and has a lot of bugs. In addition, could have breaking changes down the line, so I don't see anyone wanting to implement first-class support for bun yet. So for now, I'd stick with a docker container, or deploying manually. 

There are also some community templates on [repl.it](https://replit.com/@bddy/Bun-Template?v=1) for bun.

PR #12867, opened 2024-07-27
feat(bundler): inlining/dead-code-elimination for `import.meta.main` (and --compile)

### LINKED ISSUE #6009: Set `import.meta.main` for compiled executables

`import.meta.main` lets you create a JS library and a script you can run directly in one file. E.g.

```js
export function run() {
  console.log("running");
}

if (import.meta.main) {
  run();
}
```
This can be imported into JS code or run directly with `bun run`. However if you compile the file into an executable with `bun build --compile` then `import.meta.main` is always `false`, so the code doesn't run.

Set `import.meta.main` to `true` in the entrypoint to the compiled executable.

You can work around this by defining the CLI entrypoint in a separate file that doesn't need to use `import.meta.main`, so this isn't super high priority.

### LINKED ISSUE #9366: Make `require.main === module` an AST node
The following code:
```js
import "fs";

if (typeof require !== "undefined" && require.main === module) {
  console.log("y u do dis");
}
```

Throws this error:

```js
❯ bun stupid.js
1 | import "fs";
           ^
error: Cannot use import statement with CommonJS-only features
    at /Users/jarred/Desktop/stupid.js:1:8

note: Try require("fs") instead
note: This file is CommonJS because 'module' was used
```

We should just make `require.main === module` an AST node that conditionally emits one of:
1) `require.main === module`
2) `import.meta.main`
3) `true` (when bundling)
4) `false` (when bundling)

cc @dcramer

paperclover:
### What does this PR do?

Fixes #6009
Fixes #9366

1. The value for `import.meta.main` is inlined to be `true` for bundler entry-points.

```ts
export function run() {
  console.log("running");
}

if (import.meta.main) {
  run();
}
```

2. The expression `require.main == module` is now special cased to not not mark a module as CommonJS. This allows it to participate in the inlining.

```ts
import "fs";

if (typeof require !== "undefined" && require.main === module) {
  console.log("y u do dis");
}
```

Both of the above also support future --format=cjs (needed for js runtime that uses real commonjs modules) where `import.meta.main` when not known at bundle-time will emit `require.main == module` (one equal because it is shorter). Inverting this condition will correctly print `require.main != module`

3. Support `import.meta.main` in `--target=node`

Node.js does not support `import.meta.main`, so this rewrites both forms into something that will actually work at runtime, as previously neither would.

```
console.log(import.meta.main);
console.log(require.main == module);
```
Will bundle to
```
import {createRequire} from "node:module";
var __require = createRequire(import.meta.url);

// import_meta_main.ts
console.log(__require.main == __require.module);
console.log(__require.main == __require.module);
```

TODO: This needs a two at-runtime tests, one for commonjs, one for esm. then one file for for true and false


robobun:
@paperdave, your commit e867bdca8edd572ead2c91b3e32961835f0067e8 is building: [`#934`](https://buildkite.com/bun/bun/builds/934)

Issue #1075, opened 2022-08-14
Easy way to handle http request


lau1944:
### What is the problem this feature will solve?

1. Easy to maintain rest api
2. Easy to write api
3. Easy to learn if people know nodejs and express.

### What is the feature you are proposing to solve the problem?

I writed a package called [bunrest](https://github.com/lau1944/bunrest), it's designed to solve this problem. It's an ExpressJs-like API for bun http server. It was built on top of bun. 

If the team likes this solution, I am able to write a sample on bun example dir and pr.

Jarred-Sumner:
You're welcome to submit a PR with an example!

PR #12870, opened 2024-07-27
darwin: implement node:os.freemem()

nektro:
Closes https://github.com/oven-sh/bun/issues/3944, linux and windows are already implemented

```
❯ node index.js
Free memory: 1839923200
Free memory: 1754MB
```

```
❯ bun-debug index.js
Free memory: 1633452032
Free memory: 1557MB
```


src/bun.js/bindings/OsBinding.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <mach/mach_host.h>
+#include <mach/mach_init.h>
+#include <mach/message.h>
+#include <mach/vm_statistics.h>
+#include <unistd.h>
+
+// Adapted from libuv darwin uv_get_free_memory, MIT
+extern "C" uint64_t Bun__Os__getFreeMemory(void)
+{
+    vm_statistics_data_t info;
```
Jarred-Sumner, line 14: We don't need to worry about zero-initializing this?

Jarred-Sumner:
can you add a test that checks the number is != `1024 * 1024` and != 0?

robobun:
@nektro, your commit 83c549da65904dc875aa1d742b40a21d67cb85f7 is building: [`#1075`](https://buildkite.com/bun/bun/builds/1075)

src/bun.js/bindings/OsBinding.cpp
nektro: no, it's initialized by `host_statistics()`

PR #12871, opened 2024-07-27
zig: remove bun.FD.int() method

nektro:
finishes off a long standing TODO

### DIFF
--- a/src/bun.js/api/bun/process.zig
+++ b/src/bun.js/api/bun/process.zig
@@ -139,11 +139,12 @@ pub const ProcessExitHandler = struct {
         }
     }
 };
-pub const PidFDType = if (Environment.isLinux) fd_t else u0;
+pub const PidFDType = if (Environment.isLinux) bun.FD else u0;
+const pidfd_invalid = if (Environment.isLinux) bun.invalid_fd else 0;
 
 pub const Process = struct {
     pid: pid_t = 0,
-    pidfd: PidFDType = 0,
+    pidfd: PidFDType = pidfd_invalid,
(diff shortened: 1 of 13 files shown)


src/bun.js/api/bun/process.zig
```diff
@@ -454,9 +455,9 @@ pub const Process = struct {
         }
 
         if (comptime Environment.isLinux) {
-            if (this.pidfd != bun.invalid_fd.int() and this.pidfd > 0) {
+            if (this.pidfd != bun.invalid_fd) {
```
Jarred-Sumner, line 458: This is a behavior difference

src/bun.js/api/bun/process.zig
```diff
@@ -174,7 +175,7 @@ pub const Process = struct {
     ) *Process {
         return Process.new(.{
             .pid = posix.pid,
-            .pidfd = posix.pidfd orelse 0,
+            .pidfd = posix.pidfd orelse pidfd_invalid,
```
Jarred-Sumner, line 178: This is also a behavior difference

src/bun.zig
```diff
@@ -161,7 +149,7 @@ pub const FileDescriptor = enum(FileDescriptorInt) {
     }
 
     pub fn eq(this: FileDescriptor, that: FileDescriptor) bool {
-        if (Environment.isPosix) return this.int() == that.int();
+        if (Environment.isPosix) return this.cast() == that.cast();
```
paperclover, line 152: wait this cant just be this == that?

nektro: nope, good point; it can be

PR #12874, opened 2024-07-27
fix: Throw if `readableStreamTo*` stream is not `ReadableStream`

frobinsonj:
### What does this PR do?

This PR adds type checks for the `readableStreamTo*` function's `stream`.

Existing draft PR: https://github.com/oven-sh/bun/pull/12487

@paperdave I hope you don't mind, I assume you have more pressing issues to focus on. I believe this PR addresses the requested changes, if not, feel free to discard!

- [x] Code changes

- [x] I ran my tests locally and they pass (`bun-debug test web/streams`)


paperclover (dismissed):

src/js/builtins/ReadableStream.ts
```diff
@@ -168,18 +176,22 @@ export function readableStreamToFormData(
   stream: ReadableStream<ArrayBuffer>,
   contentType: string | ArrayBuffer | ArrayBufferView,
 ): Promise<FormData> {
-  return Bun.readableStreamToBlob(stream).then(blob => {
-    return FormData.from(blob, contentType);
-  });
+  if (!$isReadableStream(stream)) throw new TypeError("First argument must be a ReadableStream");
+
+  return Bun.readableStreamToBlob(stream).then(blob => FormData.from(blob, contentType));
```
Jarred-Sumner: What was the purpose of this?

frobinsonj: Misread as an internal function call. Reverted.

EDIT: I think I actually just added this to provide a more accurate error message since in this case, there are two arguments. I've reverted it now but happy to re-add this.

PR #12897, opened 2024-07-28
Remove some dynamic memory allocations in uWebSockets

Jarred-Sumner:
### What does this PR do?

The `MovableFunction` in uWebSockets is a mostly unnecessary memory allocation. We can instead use C function pointers, since that's what we're using anyway.

image

Also removed the c-api folder, because it's confusing to have two different versions of the same thing

Ran it


7f8ddd:
https://bun.report/1.1.22/lr255d1c8bAioihwCuhi/iE+xpR+yxq3Dsm/z3D04t03Doq226C2iu87C06t76Dq5i+6Cm11+6CA2AyW

cirospaciari (approved):

PR #12909, opened 2024-07-29
Use typed allocator in streams

Jarred-Sumner:
### What does this PR do?

Use typed allocator in streams

So we can track it in heap stats

```js
import { ArrayBufferSink } from "bun";
import { heapStats } from "bun:jsc";

const huge = Buffer.alloc(1024 * 32, "a");
function run() {
  const sink = new ArrayBufferSink();
  for (let i = 0; i < 100; i++) {
    sink.write(huge);
  }
  return sink.end();
}

function batch() {
  for (let i = 0; i < 20; i++) {
    run();
  }
  Bun.gc(true);
}
batch();
console.log(heapStats().zones);
console.log("RSS", (process.memoryUsage().rss / 1024 / 1024) | 0);

function iter() {
    batch();
  }
  console.log(heapStats().zones);
  console.log("RSS", (process.memoryUsage().rss / 1024 / 1024) | 0);
}

iter();
iter();
iter();
iter();
iter();
iter();
iter();
iter();
```

### DIFF
--- a/src/bun.js/base.zig
+++ b/src/bun.js/base.zig
@@ -472,6 +472,25 @@ pub const ArrayBuffer = extern struct {
 
     const log = Output.scoped(.ArrayBuffer, false);
 
+    pub export fn Bun__ArrayBuffer__deallocate(bytes: ?[*]u8, byte_length: usize, allocatorPtr: ?*anyopaque, fntable: *const std.mem.Allocator.VTable) void {
+        const allocator = std.mem.Allocator{
+            // We use an "undefined" pointer for default_allocator
+            // So this value can genuinely be anything.
+            .ptr = allocatorPtr orelse undefined,
+            .vtable = fntable,
+        };
+
+        if (bytes) |ptr| {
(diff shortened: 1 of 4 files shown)


src/bun.js/base.zig
```diff
@@ -472,14 +472,33 @@ pub const ArrayBuffer = extern struct {
 
     const log = Output.scoped(.ArrayBuffer, false);
 
+    pub export fn Bun__ArrayBuffer__deallocate(bytes: ?[*]u8, byte_length: usize, allocatorPtr: ?*anyopaque, fntable: *const std.mem.Allocator.VTable) void {
+        const allocator = std.mem.Allocator{
+            // We use an "undefined" pointer for default_allocator
+            // So this value can genuinely be anything.
+            .ptr = allocatorPtr orelse undefined,
```
paperclover, line 479: should just change allocatorPtr to be a non-null pointer and let the caller pass in undefined memory, instead of having a runtime branch that willingly puts in undefined memory.

PR #12923, opened 2024-07-29
docs: Add hint for memory timeline in debugger

m1212e:
It took me a good while to setup snapshot logging and importing them into Epiphany while I just could have used the debugger which has the memory tracking included. Didn't know about that, could save someone some time in the future mentioning this in the section about memory snapshots!

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
Ah yeah sorry. This doc was written before we had a debugger.

Issue #12644, opened 2024-07-18
Websocket client receiving 1006 after sending 4998 close code


bsansouci:
### What version of Bun is running?

1.1.20

### What platform is your computer?

Linux 6.1.0-20-cloud-amd64 x86_64 unknown

### What steps can reproduce the bug?

First off, I am not sure this is a bug due to Bun. I only noticed it after switching over but it might've been around for a while under node. The behavior is that on the client I close the websocket with a specific error code 4998, on the bun server I receive 4998 in the onclose callback, but then on the client when the onclose gets finally called, I get 1006 which implies something happened and chrome decided to nuke the connection.
It's possible that there are a few edge cases where chrome or firefox throw 1006 and so I shouldn't rely on the error code, but I don't want to ignore this problem.

It's hard to repro because this only happens with our specific GCP setup. We have a load balancer in front of a compute engine instance. In the compute engine instance we just run bun that listens on a port. All of this is with TLS enabled. This doesn't repro when running bun locally without TLS.

cirospaciari:
Thank for reporting, can you send me a example of the close payload/reason? If the reason/payload contains invalid UTF-8 this can happen. Maybe it can be an encoding bug or utf8 validation bug.

cirospaciari:
In some case we are doing a fast shutdown on TLS that may is the cause of this, working in a fix and also working in a repro for this bug.

PR #12857, opened 2024-07-26
fix(tls/websocket) TLS shutdown and close frame improvements

cirospaciari:
### What does this PR do?

Fix: https://github.com/oven-sh/bun/issues/12644

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

Existing tests

PR #12326, opened 2024-07-03
fix(sockets) ref count usockets context free

cirospaciari:
### What does this PR do?
Fixes https://github.com/oven-sh/bun/issues/10832
Fixes https://github.com/oven-sh/bun/issues/10991
Fixes https://github.com/oven-sh/bun/issues/10797
Fixes https://github.com/oven-sh/bun/issues/10772
Fixes https://github.com/oven-sh/bun/issues/10695
Fixes https://github.com/oven-sh/bun/issues/10539
Fixes https://github.com/oven-sh/bun/issues/10436

context is now ref counted, actual free is delayed until next tick

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [X] Code changes

### How did you verify your code works?

Existing tests

Issue #903, opened 2022-07-28
Implement `process.dlopen`


sno2:
### What is the problem this feature will solve?

This will help build the bridge of interoperability between Bun and Node. I haven't found anything that is blocked by this as it was added in Node 15 so it is most likely low priority.

### What is the feature you are proposing to solve the problem?

Implement https://nodejs.org/dist/latest-v18.x/docs/api/process.html#processdlopenmodule-filename-flags

sno2:
Hmm it was actually implemented just not in the types, I will fix.

PR #12925, opened 2024-07-29
fix(tls/socket/fetch) shutdown fix + ref counted context

cirospaciari:
### What does this PR do?
This PR supersedes https://github.com/oven-sh/bun/pull/12857 and https://github.com/oven-sh/bun/pull/12326
Do not use fast shutdown anymore, close on TLS should try a full shutdown first unless close(.failure), added context ref count so existing sockets keep it alive until we receive close_notify.

HTTP thread instead of shutdown will call close which will close the connection immediately on TCP and do a fast shutdown on TLS.

Fix: https://github.com/oven-sh/bun/issues/12644

- [X] Code changes

Existing tests

### DIFF
--- a/packages/bun-usockets/src/bsd.c
+++ b/packages/bun-usockets/src/bsd.c
@@ -26,6 +26,9 @@
 #include <stdlib.h>
 
 #ifndef _WIN32
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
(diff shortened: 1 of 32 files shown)


robobun:
@cirospaciari, your commit 531bd5223d69600a3b1a8377e19c931785215aec is building: [`#903`](https://buildkite.com/bun/bun/builds/903)

packages/bun-usockets/src/bsd.c
```diff
@@ -26,6 +26,9 @@
 #include <stdlib.h>
 
 #ifndef _WIN32
+#ifndef _GNU_SOURCE
```
Jarred-Sumner, line 30:
```suggestion
// Necessary for the stdint include
#ifndef _GNU_SOURCE
```

PR #12926, opened 2024-07-29
fix: child_process test with specified shell for windows

dariushalipour:
### What does this PR do?

There are couple of constantly failing tests in CI for windows-x64. This addresses the one on child_process.

The issue is that in powershell, to get the currently running shell executable path, one can read it from `$PSHOME`, and `$env:SHELL` would not hold that value as far as I could investigate and try.

- [x] Code changes (sort of, it's only tests)


Jarred-Sumner:
Thank you

Issue #949, opened 2022-08-01
Throws an error if other dependency wants to require http module


h44mkk:
### Version

0.1.5

### Platform

Darwin 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; RELEASE_X86_64 x86_64

### What steps will reproduce the bug?

- Add express dependency
- Add a get method from express
- Add a listener
- Command : bun server.ts
Error shows in Terminal : SyntaxError: Importing binding name 'EventEmitter' is not found.

### What do you see instead?

10 | /**
11 |  * Module dependencies.
12 |  * @private
13 |  */
14 | 
15 | var http = require('http');
                      ^
SyntaxError: Importing binding name 'EventEmitter' is not found.
      at /Users/hamza/Desktop/bun.js/node_modules/methods/index.js:15:19
      at bun:wrap:1:16430
      at /Users/hamza/Desktop/bun.js/node_modules/express/lib/router/route.js:19:22
      at bun:wrap:1:16430
      at /Users/hamza/Desktop/bun.js/node_modules/express/lib/router/index.js:16:20
      at bun:wrap:1:16430
      at /Users/hamza/Desktop/bun.js/node_modules/express/lib/application.js:17:21
      at bun:wrap:1:16430
      at /Users/hamza/Desktop/bun.js/node_modules/express/lib/express.js:18:20
      at bun:wrap:1:16430
      at /Users/hamza/Desktop/bun.js/node_modules/express/index.js:11:25
      at bun:wrap:1:16430
      at /Users/hamza/Desktop/bun.js/http.ts:2:0

robogeek:
I get the same on this:

```
Linux davidpc 5.15.0-41-generic #44-Ubuntu SMP Wed Jun 22 14:20:53 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

The code in my case is rather complex - Using Chokidar to get the list of files in a directory structure, then store the data in an in-memory database (ForerunnerDB).  But, when run under Bun the test case immediately throws same error as in the report above.

In this case - Forerunner is automatically launching an Express server in the background.  The stack trace is part of `require` of `express` which in turn requires `http` as shown above.

Therefore, fixing this will get a step towards being able to run Express

robogeek:
https://github.com/oven-sh/bun/pull/572

PR #12929, opened 2024-07-30
NodeError: add more and use them in child_process and dgram

nektro:


robobun:
@nektro, your commit 4f16a632388b102e8ca8754880bd36d7290ab287 is building: [`#949`](https://buildkite.com/bun/bun/builds/949)

PR #12931, opened 2024-07-30
test: rewrite "should call close" to use promise instead of done

nektro:


robobun:
@nektro, your commit 3e3b6e9efbfb1001b77b8be441d11f7458fb50dc is building: [`#951`](https://buildkite.com/bun/bun/builds/951)

PR #964, opened 2022-08-02
Fix u32 jsNumber cast

zhuzilin:
Before this PR, `bun-debug run examples/hashing.js` will raise `integer cast truncated bits` for crc32.

Thank you for your time on reviewing this PR:)

PR #12932, opened 2024-07-30
fix crash handler test failures

### LINKED ISSUE #12747: bun.report link includes windows path separators

1.1.21-canary.1+5a5f3d6b3

Windows NT 10.0.19045.0 x64

Crash on windows

I get a link that i can click to open, i.e `https://bun.report/1.1.21/wa25a5f3d6AiogggDCqB/node.napi.glibc.nodesjZCqB/node.napi.glibc.node+jjCCqB/node.napi.glibc.nodeo/7Bw6sk0C0q+psB__A2AwK`

### What do you see instead?

A link with `\` in the middle instead of `/`, thus i can't click the full link because the terminal doesn't understand it
`https://bun.report/1.1.21/wa25a5f3d6AiogggDCqB\node.napi.glibc.nodesjZCqB\node.napi.glibc.node+jjCCqB\node.napi.glibc.nodeo/7Bw6sk0C0q+psB__A2AwK`

image

### Additional information

_No response_

paperclover:
### What does this PR do?

Also closes #12747 but i honestly don't want to write a test to orchestrate a crash that would cause this bug.

### DIFF
--- a/test/cli/run/run-crash-handler.test.ts
+++ b/test/cli/run/run-crash-handler.test.ts
@@ -13,11 +13,13 @@ test.if(process.platform === "darwin")("macOS has the assumed image offset", () 
 
 test("raise ignoring panic handler does not trigger the panic handler", async () => {
   let sent = false;
+  const resolve_handler = Promise.withResolvers();
 
   using server = Bun.serve({
     port: 0,
     fetch(request, server) {
       sent = true;
+      resolve_handler.resolve();
       return new Response("OK");
     },
   });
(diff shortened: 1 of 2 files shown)


robobun:
@paperdave, your commit 5226d19fd5ee888a2c8ddb92505db58bf04a406d is building: [`#964`](https://buildkite.com/bun/bun/builds/964)

test/cli/run/run-crash-handler.test.ts
```diff
@@ -33,6 +35,11 @@ test("raise ignoring panic handler does not trigger the panic handler", async ()
     ]),
   });
 
+  await proc.exited;
+
+  /// Wait two seconds for a slow http request, or continue immediatly once the request is heard.
+  await Promise.race([resolve_handler.promise, Bun.sleep(2000)]);
```
Electroid, line 41: 2 seconds is not sufficient on IO busy CI machines. Can we just remove the manual timeout and let the test itself timeout?

paperclover: please re-read this test, it is waiting 2 seconds of threashold, and **passing** after that timer. the other promise is used to end the test (and fail) as soon as it hears a response.

if this test is still failing then there is actually broken underlying behavior i cannot reproduce locally.

PR #12975, opened 2024-07-31
codegen: uninline more callframes in debug mode with bun.callmod_inline

nektro:
first introduced in https://github.com/oven-sh/bun/pull/9001, this decl is used in conjunction with `@call` to change the call modifier being used to not use inline only when in Debug mode. this does not change the higher level behavior but does share a small speed hit for making the affected stack frames behave more consistently inside a debugger when investigating crashes


robobun:
@nektro, your commit 44ee31ebcb873f3d4f28079723d79b4a48e755f2 is building: [`#1037`](https://buildkite.com/bun/bun/builds/1037)

Jarred-Sumner:
The usage of inline is not a performance thing in this case. Callframe arguments can move when there are enough function calls in between. So we have to not add more

Issue #1044, opened 2022-08-10
Does not run on AMD FX-4100


Lambdac0re:
### Version

0.1.7

### Platform

Linux Vulcan 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### What steps will reproduce the bug?

```
$ curl https://bun.sh/install | bash
$ bun
Illegal instruction
```

### How often does it reproduce? Is there a required condition?

Always fails

### What is the expected behavior?

Bun runs

### What do you see instead?

`Illegal instruction`

xhyrom:
Already fixed, use canary build:
Check https://github.com/oven-sh/bun/issues/1011

xhyrom:
Duplicate of https://github.com/oven-sh/bun/issues/1011

AsuToki360:
Still crashing on FX-6300, can't run, just pure SEGILL, even tried as root, CachyOS (Arch Linux based). Tried all versions with "downgrade" package from Cachy, no success.

<img width="1148" height="605" alt="Image" src="https://github.com/user-attachments/assets/5a94c4be-3e7f-468a-927f-1657aa132ca9" />

```
[zsh] > bun --verbose
[1]    24685 illegal hardware instruction (core dumped)  bun --verbose

```

```
[root@calliope-cachyos calli]# bun
Illegal instruction        (core dumped) bun
[root@calliope-cachyos calli]# bun --verbose
Illegal instruction        (core dumped) bun --verbose
[root@calliope-cachyos calli]# 

```

PR #12979, opened 2024-07-31
fix(server/tls/streams) fix onReadFile, streams, avoid shutdown on fatal errors, ensure ssl loop data and server ref count refactor

cirospaciari:
### What does this PR do?

- [X] Code changes

Existing tests

### DIFF
--- a/packages/bun-usockets/src/context.c
+++ b/packages/bun-usockets/src/context.c
@@ -63,9 +63,8 @@ void us_socket_context_close(int ssl, struct us_socket_context_t *context) {
     struct us_listen_socket_t *ls = context->head_listen_sockets;
     while (ls) {
         struct us_listen_socket_t *nextLS = (struct us_listen_socket_t *) ls->s.next;\
-        if(!us_socket_is_closed(ssl, ls)) {
-            us_listen_socket_close(ssl, ls);
-        }
+        us_listen_socket_close(ssl, ls);
+        
         ls = nextLS;
     }
 
@@ -73,7 +72,7 @@ void us_socket_context_close(int ssl, struct us_socket_context_t *context) {
     struct us_socket_t *s = context->head_sockets;
     while (s) {
         struct us_socket_t *nextS = s->next;
-        if(!us_socket_is_closed(ssl, ls)) {
+        if(!us_socket_is_closed(ssl, s)) {
             us_socket_close(ssl, s, LIBUS_SOCKET_CLOSE_CODE_CLEAN_SHUTDOWN, 0);
         }
         s = nextS;
--- a/packages/bun-usockets/src/context.c
+++ b/packages/bun-usockets/src/context.c
@@ -20,7 +20,6 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
-#include <errno.h>
 #ifndef _WIN32
 #include <arpa/inet.h>
 #endif
@@ -63,8 +62,10 @@ void us_socket_context_close(int ssl, struct us_socket_context_t *context) {
     /* Begin by closing all listen sockets */
     struct us_listen_socket_t *ls = context->head_listen_sockets;
     while (ls) {
-        struct us_listen_socket_t *nextLS = (struct us_listen_socket_t *) ls->s.next;
-        us_listen_socket_close(ssl, ls);
+        struct us_listen_socket_t *nextLS = (struct us_listen_socket_t *) ls->s.next;\
+        if(!us_socket_is_closed(ssl, ls)) {
+            us_listen_socket_close(ssl, ls);
+        }
         ls = nextLS;
     }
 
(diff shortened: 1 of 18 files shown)


robobun:
@cirospaciari, your commit 7b243d2feb87b8b5399f36e71cc5986e828f60d6 is building: [`#1044`](https://buildkite.com/bun/bun/builds/1044)

packages/bun-usockets/src/context.c
```diff
@@ -63,16 +62,19 @@ void us_socket_context_close(int ssl, struct us_socket_context_t *context) {
...
         us_listen_socket_close(ssl, ls);
+        
         ls = nextLS;
     }
 
     /* Then close all regular sockets */
     struct us_socket_t *s = context->head_sockets;
     while (s) {
         struct us_socket_t *nextS = s->next;
-        us_socket_close(ssl, s, LIBUS_SOCKET_CLOSE_CODE_CLEAN_SHUTDOWN, 0);
```
Jarred-Sumner: We only need this check when the context is an SSL context, right?

cirospaciari: we need to check internally anyway, so I removed from this place, we have to refactor a lot more before making this an assertion

packages/bun-usockets/src/context.c
```diff
@@ -63,8 +62,9 @@ void us_socket_context_close(int ssl, struct us_socket_context_t *context) {
     /* Begin by closing all listen sockets */
     struct us_listen_socket_t *ls = context->head_listen_sockets;
     while (ls) {
-        struct us_listen_socket_t *nextLS = (struct us_listen_socket_t *) ls->s.next;
+        struct us_listen_socket_t *nextLS = (struct us_listen_socket_t *) ls->s.next;\
```
Jarred-Sumner: C is one of the only languages where you can have an accidental `\` and nothing breaks

Jarred-Sumner:
```suggestion
        struct us_listen_socket_t *nextLS = (struct us_listen_socket_t *) ls->s.next;
```

packages/bun-usockets/src/socket.c
```diff
@@ -348,8 +351,10 @@ int us_socket_write(int ssl, struct us_socket_t *s, const char *data, int length
 
     int written = bsd_send(us_poll_fd(&s->p), data, length, msg_more);
     if (written != length) {
-        s->context->loop->data.last_write_failed = 1;
-        us_poll_change(&s->p, s->context->loop, LIBUS_SOCKET_READABLE | LIBUS_SOCKET_WRITABLE);
+        if(s->context->loop) {
```
Jarred-Sumner: Is this a thing that happens? It has to be undefined memory right?

src/bun.js/api/server.zig
```diff
@@ -1976,7 +1978,7 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             if (this.signal) |signal| {
                 this.signal = null;
                 if (this.flags.aborted and !signal.aborted()) {
-                    const reason = JSC.WebCore.AbortSignal.createAbortError(JSC.ZigString.static("The user aborted a request"), &JSC.ZigString.Empty, this.server.globalThis);
+                    const reason = JSC.WebCore.AbortSignal.createAbortError(JSC.ZigString.static("The user aborted a request"), &JSC.ZigString.Empty, globalThis);
```
Jarred-Sumner, line 1976: As a follow-up, instead of creating this AbortError, we could probably add an enum to AbortSignal for known types of signals. Then, we can defer creating the JSValue here if it's not used.

Jarred-Sumner (changes requested):
2 comments but mostly LGTM

packages/bun-usockets/src/socket.c
cirospaciari: forgot to undo this actually, this will never happen

packages/bun-usockets/src/context.c
cirospaciari: surprised that it compiled

packages/bun-usockets/src/context.c
```diff
@@ -757,6 +758,7 @@ struct us_socket_t *us_socket_context_adopt_socket(int ssl, struct us_socket_con
         if (new_s->next) new_s->next->prev = new_s;
     } else {
         us_internal_socket_context_link_socket(context, new_s);
+        us_socket_context_unref(ssl, context);
```
Jarred-Sumner, line 762: can the socket outlive `new_s`? hard to see how this works

cirospaciari: yeah it can I just need to ref/unref to make sure its alive when unlink/link is called, so we dont free the context if is the last ref to it.

packages/bun-usockets/src/socket.c
```diff
@@ -21,6 +21,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdint.h>
+#include <stdio.h>
```
Jarred-Sumner, line 24:
```suggestion

```

src/bun.js/api/server.zig
```diff
@@ -1586,14 +1588,16 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                 this.request_body = null;
             }
 
-            const server = this.server;
-            server.request_pool_allocator.put(this);
-            server.onRequestComplete();
+            if(this.server) |server| {
+                this.server = null;
+                server.request_pool_allocator.put(this);
```
Jarred-Sumner, line 1593: maybe we shouldn't make server nullable, since it means we don't have a good mechanism for freeing it since the allocator is server-relative.

Issue #1045, opened 2022-08-10
[Bug] Bun crashing on `import React = require("react")`


coffeeispower:
This bug is happening on version 1.8.0 of bun
```ts
import React = require("react"); // CRASH
```
```
SegmentationFault at 0x3700008003000000


----- bun meta -----
Bun v0.1.8 Linux x64 #1 SMP PREEMPT_DYNAMIC Sat, 02 Jul 2022 21:03:06 +0000
RunCommand: 
Elapsed: 88ms | User: 12ms | Sys: 27ms
RSS: 67.11MB | Peak: 26.57MB | Commit: 67.11MB | Faults: 111
----- bun meta -----

Ask for #help in https://bun.sh/discord or go to https://bun.sh/issues
```

xhyrom:
Next time, use bug template

that-ambuj:
I think the commonJS syntax is :
`const DefaultExport = require('package')`

Or more commonly used ESmodule one in React :
`Import React from 'react'`

coffeeispower:
it's an old typescript thing. and visual studio code still does that, but even if won't be implemented, it shouldn't EVER segfault. Segfaults are bugs.

coffeeispower:
Great!

that-ambuj:
So is that unexpected behaviour or wrong syntax? I only have experience with javascript.

PR #12983, opened 2024-07-31
Subprocess: fix crash when stdin outlives it

nektro:
Pulled out of changes for https://github.com/oven-sh/bun/pull/11492, making separate patch to isolate against other test results, will rebase once merged. fixes crash reproducible on that branch on windows in `test/js/bun/spawn/spawn-streaming-stdin.test.ts` that occurred when Subprocess destroys itself before stdin does.


robobun:
@nektro, your commit 66d71fde0ec84b1e103917f8d62248214beffa22 is building: [`#1045`](https://buildkite.com/bun/bun/builds/1045)

src/bun.js/api/bun/subprocess.zig
```diff
@@ -2075,21 +2078,14 @@ pub const Subprocess = struct {
         }
 
         const loop = jsc_vm.eventLoop();
+        var island = bun.new(Island, .{});
```
Jarred-Sumner: Why is this using `bun.new` instead of `Island.new`?

Why is this called `Island` instead of something like `WeakPtr`? What is an island?

src/bun.js/webcore/streams.zig
```diff
@@ -1533,7 +1533,7 @@ pub const SinkDestructor = struct {
     const Subprocess = JSC.API.Bun.Subprocess;
     pub const Ptr = bun.TaggedPointerUnion(.{
         Detached,
-        Subprocess,
+        Subprocess.Island,
```
Jarred-Sumner: Since this is now a separate pointer, we can avoid going through this code path and instead directly store it on the value. We don't need a tagged pointer union for this anymore.

src/bun.js/api/bun/subprocess.zig
```diff
@@ -1434,7 +1436,7 @@ pub const Subprocess = struct {
         }
 
         if (stdin) |pipe| {
-            this.weak_file_sink_stdin_ptr = null;
+            this.island.clearStdin();
             this.flags.has_stdin_destructor_called = true;
```
Jarred-Sumner: Do we still need `has_stdin_destructor_called`?

src/bun.js/api/bun/subprocess.zig
```diff
@@ -706,10 +707,11 @@ pub const Subprocess = struct {
 
     pub fn onStdinDestroyed(this: *Subprocess) void {
         this.flags.has_stdin_destructor_called = true;
-        this.weak_file_sink_stdin_ptr = null;
+        this.island.clearStdin();
 
         if (this.flags.finalized) {
             // if the process has already been garbage collected, we can free the memory now
```
Jarred-Sumner: We don't need `state.finalized` anymore. This function should not be called when the subprocess has been collected.

Jarred-Sumner:
Which test failure is this fixing?

Jarred-Sumner (changes requested):
A few comments

src/bun.js/api/bun/subprocess.zig
nektro: good catch

src/bun.js/api/bun/subprocess.zig
nektro: good catch

src/bun.js/api/bun/subprocess.zig
nektro: doesnt that come from `NewRefCounted` ?

`WeakPtr` seemed like it would infer pointing to only one of them or generic and i didnt want to be misleading

Jarred-Sumner: > doesnt that come from NewRefCounted ?

You can use `usingnamespace bun.New`, it adds it as well.

src/bun.js/webcore/streams.zig
nektro: can i do this refactor in a followup?

Jarred-Sumner (changes requested):
While the crash is fixed, the Subprocess no longer uses the FileSink pointer, and the FileSink pointer no longer uses the Subprocess pointer

What is the purpose of having this shared pointer now?

PR #12993, opened 2024-08-01
bundler: better node builtin polyfills for browser

paperclover:
At a later point I want to switch node_fallbacks to use `bun build`. That is blocked on a release with #12808. At that point I will probably rename this to `src/js/bundler_polyfills`.

For now this just does other work to make the polyfills better.

Also, a new transform in the parser happens specifically when bundling for browser. This comment in the code explains it well:

```
If this is one of a few specific built-ins, make the default value refer to
the *namespace* instead of the default export. This allows dot properties to
note property accesses on this to make tree shaking more effective. We can
only enable it on built-in modules that have a default export that is

Bun's browser polyfills are specially designed to take advantage of this
optimization to make bundles smaller.

    import path from 'node:path'
   -->
    import * as path from 'node:path'

For example, bundling this code only includes a dependency on `function join`
and not on the entire module.

    import path from 'node:path'
    console.log(path.join);
```


robobun:
@paperdave, your commit 0e665b7dae9d858c2b841c413cb45215f63319d1 is building: [`#1072`](https://buildkite.com/bun/bun/builds/1072)

src/js_parser.zig
```diff
@@ -502,6 +501,51 @@ pub fn locAfterOp(e: E.Binary) logger.Loc {
...
+// For example, bundling this code only includes a dependency on `function join`
+// and not on the entire module.
+//
+//     import path from 'node:path'
+//     console.log(path.join);
+//
+const rewrite_default_to_star_map = bun.ComptimeStringMap(void, kvs: {
+    const node_builtins = .{
+        "buffer",
+        "constants",
```
Jarred-Sumner, line 525:
```suggestion

```

src/js_parser.zig
```diff
@@ -3100,6 +3144,8 @@ pub const Parser = struct {
 
         transform_only: bool = false,
 
+        target: options.Target = .bun,
```
Jarred-Sumner, line 3147: What is this doing? We have been going feature (supports runtime require) rather than target (bun) . Should this be in Runtime.Features?

PR #12997, opened 2024-08-01
Fix memory leak when requiring or importing modules that get GC'd later

Jarred-Sumner:
### What does this PR do?

This properly decrements the reference count for output source code.

Requiring `@mui/icons-material` 30 times uses about 30% less memory

image

There are tests

### DIFF
--- a/src/bun.js/bindings/CommonJSModuleRecord.cpp
+++ b/src/bun.js/bindings/CommonJSModuleRecord.cpp
@@ -30,6 +30,8 @@
  */
 
 #include "headers.h"
+
+#include "JavaScriptCore/Synchronousness.h"
 #include "JavaScriptCore/JSCast.h"
 #include <JavaScriptCore/JSMapInlines.h>
 #include "root.h"
@@ -72,6 +74,7 @@
 #include "PathInlines.h"
 #include "wtf/NakedPtr.h"
 #include "wtf/URL.h"
+#include "wtf/text/StringImpl.h"
 
 extern "C" bool Bun__isBunMain(JSC::JSGlobalObject* global, const BunString*);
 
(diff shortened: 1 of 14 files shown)


src/bun.js/bindings/CommonJSModuleRecord.cpp
```diff
@@ -171,6 +173,13 @@ static bool evaluateCommonJSModuleOnce(JSC::VM& vm, Zig::GlobalObject* globalObj
         args.append(Zig::ImportMetaObject::create(globalObject, filename));
     }
 
+    // Clear the source code as early as possible.
+    code = {};
+
+    // Call the CommonJS module wrapper function.
+    //
+    //    fn(exports, require, module, __filename, __dirname) { /* code */ }(exports, require, module, __filename, __dirname)
```
paperclover, line 181: note: it is not an iife expression

src/bun.js/bindings/ModuleLoader.cpp
```diff
@@ -289,13 +307,7 @@ static JSValue handleVirtualModuleResult(
...
     JSC::VM& vm = globalObject->vm();
     auto scope = DECLARE_THROW_SCOPE(vm);
-    WTF::String sourceCodeStringForDeref;
-    const auto getSourceCodeStringForDeref = [&]() {
-        if (res->success && res->result.value.needsDeref && res->result.value.source_code.tag == BunStringTag::WTFStringImpl) {
-            res->result.value.needsDeref = false;
-            sourceCodeStringForDeref = String(res->result.value.source_code.impl.wtf);
-        }
-    };
+    ResolvedSourceCodeHolder sourceCodeHolder(res);
```
paperclover, line 310: this is a nice use of a c++ destructor

test/harness.ts
```diff
@@ -48,6 +48,12 @@ for (let key in bunEnv) {
 
 delete bunEnv.NODE_ENV;
 
+if (isDebug) {
+  // This makes debug build memory leak tests more reliable.
+  // The code for dumping out the debug build transpiled source code has leaks.
+  bunEnv.BUN_DEBUG_NO_DUMP = "1";
```
paperclover, line 54: i think we should have this but can we also fix the memory leak in the source code dump.

line 222: add `defer bun.default_allocator.free(source_file);`

paperclover (approved):

PR #13000, opened 2024-08-01
Include wrangler in default-trusted-dependencies.txt

threepointone:
`@cloudflare/wrangler` is already in this list, but that was the v1 version, long since deprecated. The new one (v2 onwards) is just `wrangler`, this patch adds it to `default-trusted-dependencies.txt`. `wrangler` itself does NOT have a postinstall script right now, but may soon.

### What does this PR do?

- [x] Code changes

 I have not :(

### DIFF
--- a/src/install/default-trusted-dependencies.txt
+++ b/src/install/default-trusted-dependencies.txt
@@ -357,6 +357,7 @@ windows-build-tools
 wix-style-react
 wordpos
 workerd
+wrangler
 wrtc
 xxhash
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Happy to add it once you have a postinstall script

nektro:
no postinstall script as of wrangler@4.22.0 so closing for now.
statement/offer from jarred still stands, happy to add should one be added in the future.

PR #13013, opened 2024-08-02
ci: disable BUN_ENABLE_CRASH_REPORTING

nektro:
bun.report crash urls in our CI as-is will never be valid in our own CI since they will be development versions that bun.report doesn't know about. set this flag to disable in the runner so that bun attempts to print a stack trace to stderr instead


robobun:
@nektro, your commit 6d07913729ef782c18dcedad75406847cb7adc71 is building: [`#1120`](https://buildkite.com/bun/bun/builds/1120)

Issue #1150, opened 2022-08-26
Issues resolving Promise within `setInterval()` scope


ctjlewis:
### Version

0.1.10

### Platform

Darwin Lewiss-MacBook-Air.local 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:35 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T8101 arm64

### What steps will reproduce the bug?

An attempt to resolve a Promise from within a `const interval = setInterval(...)` call that clears itself using `clearInterval(interval)` results in unexpected behavior where the process never exits.

Interestingly, the Promise *does* resolve, and a statement after `await new Promise(...)` will execute, but Bun simply never closes the process. Similar behavior is not reproduced in Node.

#### Repro

This repro is very simple, but it was distilled down (significantly) from a real-world bug I found writing a testing library for Bun. Basically just clarifying that this isn't some trivial edge case but seemingly a noteworthy issue related to resolving a Promise in a setInterval scope.

https://replit.com/@ctjlewis/bun-interval-promise

### How often does it reproduce? Is there a required condition?

No related issues with `setTimeout`, seems `setInterval`-specific.

### What is the expected behavior?

The process should close when the program is finished.

### What do you see instead?

A process that never closes. In the repro program, a Promise is awaited and the last expression is a `console.log` call that does execute, but the `bun` process never terminates.

#### Unsuccessful workarounds

  - ❌  Placing `interval` identifier in parent scope

#### Successful workarounds

  -  ✅  Refactoring interval as recursive `setTimeout`

Jarred-Sumner:
Can you try this in Bun v0.1.11 and see if it still happens?

To upgrade:
```bash
bun upgrade --canary
```

ctjlewis:
@Jarred-Sumner Didn't know about `--canary`. It does not repro! Based.

PR #13017, opened 2024-08-02
fix(bundler): printing e_commonjs_export_identifier when it got deoptimized.

### LINKED ISSUE #13006: Firestore fails with 'undefined is not a constructor' when running bun in compile mode

1.1.21+70ca2b76c

Linux 6.6.36.3-microsoft-standard-WSL2 x86_64 x86_64

here is the code that i run:
```ts
import { initializeApp } from 'firebase-admin/app';
import { getFirestore } from 'firebase-admin/firestore';

const db = getFirestore(initializeApp());

const docRef = db.collection('test').doc('test');
const docSnap = await docRef.get();

if (docSnap.exists) {
  console.log(docSnap.data())
}
```

repo for minimal issue repro is here: 

https://github.com/Gorthog/bun-issue

you will need to:
1. create a json key for firebase. example: `gcloud iam service-accounts keys create service-account-key.json --iam-account=$project_number-compute@developer.gserviceaccount.com`
2. create a firebase collection named "test" and a document named "test" with a field in it.
3. run `bun start:prod`

when compiling bun and then running it, i expect the code to return the data in firestore. just like bun is able to when the app is not compiled. example:

image

### What do you see instead?

image

### Additional information

_No response_

paperclover:
### What does this PR do?

Fixes #13006

Consider this code snippet

```ts
let y = () => module.exports.xyz; // A
module.exports = { xyz: 123 }; // B
let z = () => module.exports.xyz; // C
console.log(y(), z());
```

- A visits and converts `module.exports` to `exports` and creates a CJS Export Identifier. this node is for `<export id>.xyz` in one node, for when this module gets converted CJS->ESM
- reassigning `module.exports` says 
- C visits and does not perform conversion since we are de-optimized

see the issue, we never update A, so we have to handle printing it back as `module.exports`, since the `exports` variable was not updated.

```diff
 var __commonJS = (cb, mod) => () => (mod || cb((mod = { exports: {} }).exports, mod), mod.exports);

 // index.ts
 var require_index = __commonJS((exports, module) => {
-  var y = () => exports.xyz;
+  var y = () => module.exports.xyz;
   module.exports = { xyz: 123 };
   var z = () => module.exports.xyz;
   console.log(y(), z());
 });
 export default require_index();
```

EDIT: the approach taken was slightly altered as just changing the printer would cause react to get a bit larger. the new approach does some parser changes to ensure this change is only applied when it is needed

### DIFF
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -9034,6 +9034,7 @@ pub const LinkerContext = struct {
             .indent = .{},
             .commonjs_named_exports = ast.commonjs_named_exports,
             .commonjs_named_exports_ref = ast.exports_ref,
+            .commonjs_module_ref = if (ast.flags.uses_module_ref) ast.module_ref else Ref.None,
             .commonjs_named_exports_deoptimized = flags.wrap == .cjs,
             // .const_values = c.graph.const_values,
             .ts_enums = c.graph.ts_enums,
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -9036,6 +9036,7 @@ pub const LinkerContext = struct {
             .commonjs_named_exports_ref = ast.exports_ref,
             .commonjs_module_ref = if (ast.flags.uses_module_ref) ast.module_ref else Ref.None,
             .commonjs_named_exports_deoptimized = flags.wrap == .cjs,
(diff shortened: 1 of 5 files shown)


robobun:
@paperdave, your commit d8d6235cdf12e10ba96202fc86b6af862e868150 is building: [`#1150`](https://buildkite.com/bun/bun/builds/1150)

PR #1151, opened 2022-08-26
fix mkdtemp

zhuzilin:
fix #1144

PR #13018, opened 2024-08-02
ci: debug node-http.test.ts windows segfault not having a stack trace

nektro:
https://github.com/oven-sh/bun/pull/13013 should have fixed this, so add some more metadata to shine a light on its decision making

### DIFF
--- a/src/crash_handler.zig
+++ b/src/crash_handler.zig
@@ -884,10 +884,12 @@ pub fn printMetadata(writer: anytype) !void {
             &peak_commit,
             &page_faults,
         );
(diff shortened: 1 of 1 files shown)


robobun:
@nektro, your commit 9275cb16056eb09fad4b2e951f49a4f3d061dfcf is building: [`#1151`](https://buildkite.com/bun/bun/builds/1151)

nektro:
before

image

nektro:
the CI runs with release mode binaries so this may not be workable unless we generate pdb files. im not sure atm if we do.

PR #12936, opened 2024-07-30
Make bun install faster in China

Jarred-Sumner:
## This PR will not be merged

See **[this comment](https://github.com/oven-sh/bun/pull/12936#issuecomment-2265147603)**

----

Original PR:

### What does this PR do?

image

When the locale is the country code `CN` or `cn`:
- Default to registry.npmmirror.com

Default these environment variables for lifecycle scripts:
```bash
npm_config_disturl="https://npmmirror.com/dist"
npm_config_sharp_binary_host="https://npmmirror.com/mirrors/sharp/"
npm_config_sharp_libvips_binary_host="https://npmmirror.com/mirrors/sharp-libvips"
npm_config_profiler_binary_host_mirror="https://npmmirror.com/mirrors/node-inspector/"
npm_config_fse_binary_host_mirror="https://npmmirror.com/mirrors/fsevents"
npm_config_node_sqlite3_binary_host_mirror="https://npmmirror.com/mirrors"
npm_config_sqlite3_binary_host_mirror="https://npmmirror.com/mirrors"
npm_config_sqlite3_binary_site="https://npmmirror.com/mirrors/sqlite3"
npm_config_electron_mirror="https://npmmirror.com/mirrors/electron/v"
PUPPETEER_DOWNLOAD_BASE_URL="https://npmmirror.com/mirrors"
npm_config_chromedriver_cdnurl="https://npmmirror.com/mirrors/chromedriver"
npm_config_operadriver_cdnurl="https://npmmirror.com/mirrors/operadriver"
npm_config_phantomjs_cdnurl="https://npmmirror.com/mirrors/phantomjs"
npm_config_python_mirror="https://npmmirror.com/mirrors/python"
npm_config_sass_binary_site="https://npmmirror.com/mirrors/node-sass"
```

PR #13020, opened 2024-08-02
Add documentation for changing mirror registries in Bun

arvinxx:
based on  #12936 , most users perfer to have a user guide to update registries. So I propose this PR.

Add documentation for changing the mirror registries in Bun.

* Add a new section in `docs/install/registries.md` explaining how to change the mirror source in Bun.
* Include examples and detailed steps for configuring the mirror source in `bunfig.toml`.
* Provide actual mirror sources based on the discussion in #12936.
* Add a Chinese version of the configuration guide for users in mainland China.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/oven-sh/bun?shareId=442d88d2-7528-4b33-978c-bcc9c92acffe).

### DIFF
--- a/docs/install/registries.md
+++ b/docs/install/registries.md
@@ -25,42 +25,43 @@ To configure a private registry scoped to a particular organization:
 "@myorg3" = { token = "$npm_token", url = "https://registry.myorg.com/" }
 ```
 
-### Changing the Mirror Source in Bun
+### Changing to Mirror Registry in Bun
 
-To change the mirror source in Bun, you can configure the `bunfig.toml` file. Here are the steps:
+If you cannot access the official registry, you can change to a mirror registry. 
+
+To change the mirror registry in Bun, you can configure the `bunfig.toml` file. Here are the steps:
 
 1. Open the `bunfig.toml` file in your project directory. If it doesn't exist, create a new file named `bunfig.toml`.
 
-2. Add the following configuration to set the mirror source:
+2. Add the following configuration to set the mirror registry:
 
 ```toml
 [install]
-# set the mirror source URL
-registry = "https://your-mirror-source-url"
+# set the mirror registry URL
+registry = "https://registry.npmmirror.com"
 ```
 
-3. If your mirror source requires authentication, you can include the token or username/password:
+3. If your mirror registry requires authentication, you can include the token:
 
 ```toml
 [install]
-# set the mirror source URL with token
-registry = { url = "https://your-mirror-source-url", token = "your-token" }
-
-# set the mirror source URL with username/password
-registry = "https://username:password@your-mirror-source-url"
+# set the mirror registry URL with token
+registry = { url = "https://registry.npmmirror.com", token = "your-token" }
 ```
 
 4. Save the `bunfig.toml` file.
 
-For more detailed information on configuring the mirror source and other registry settings, refer to the [custom registry guide](../guides/install/custom-registry.md).
+For more detailed information on configuring the mirror registry and other registry settings, refer to the [custom registry guide](../guides/install/custom-registry.md).
 
-### Available Mirror Registry
+### Recommended Mirror Registries
 
-Based on the discussion in #12936, here are some actual mirror sources that you can use:
+Based on the discussion in #12936, here are some recommended mirror registries that you can use:
 
 - `https://registry.npmmirror.com`
 - `https://mirrors.cloud.tencent.com/npm/`
 - `https://repo.huaweicloud.com/repository/npm/`
+
+Please note that our recommendations do not guarantee availability.
 
 ### `.npmrc`
 
@@ -87,30 +88,29 @@ Bun does not currently read `.npmrc` files. For private registries, migrate your
 ```toml
 [install]
 # 设置镜像源 URL
-registry = "https://your-mirror-source-url"
+registry = "https://registry.npmmirror.com"
 ```
 
-3. 如果您的镜像源需要身份验证，可以包含令牌或用户名/密码：
+3. 如果您的镜像源需要身份验证，可以包含令牌：
 
 ```toml
 [install]
 # 设置带有令牌的镜像源 URL
-registry = { url = "https://your-mirror-source-url", token = "your-token" }
-
-# 设置带有用户名/密码的镜像源 URL
-registry = "https://username:password@your-mirror-source-url"
+registry = { url = "https://registry.npmmirror.com", token = "your-token" }
 ```
 
 4. 保存 `bunfig.toml` 文件。
 
 有关配置镜像源和其他注册表设置的详细信息，请参阅[自定义注册表指南](../guides/install/custom-registry.md)。
 
-### 可用的镜像源
+### 推荐的镜像源
 
-根据 #12936 中的讨论，以下是一些实际可用的镜像源：
+根据 #12936 中的讨论，以下是一些推荐的镜像源：
 
 - `https://registry.npmmirror.com`
 - `https://mirrors.cloud.tencent.com/npm/`
 - `https://repo.huaweicloud.com/repository/npm/`
+
+请注意，我们的推荐不保证可用性。
 
 </details>
--- a/docs/install/registries.md
+++ b/docs/install/registries.md
@@ -25,6 +25,92 @@ To configure a private registry scoped to a particular organization:
 "@myorg3" = { token = "$npm_token", url = "https://registry.myorg.com/" }
 ```
 
+### Changing the Mirror Source in Bun
+
+To change the mirror source in Bun, you can configure the `bunfig.toml` file. Here are the steps:
+
+1. Open the `bunfig.toml` file in your project directory. If it doesn't exist, create a new file named `bunfig.toml`.
+
+2. Add the following configuration to set the mirror source:
+
+```toml
+[install]
+# set the mirror source URL
+registry = "https://your-mirror-source-url"
+```
+
+3. If your mirror source requires authentication, you can include the token or username/password:
+
+```toml
+[install]
+# set the mirror source URL with token
+registry = { url = "https://your-mirror-source-url", token = "your-token" }
+
+# set the mirror source URL with username/password
+registry = "https://username:password@your-mirror-source-url"
+```
+
+4. Save the `bunfig.toml` file.
+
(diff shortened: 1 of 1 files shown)


docs/install/registries.md
```diff
@@ -25,6 +25,92 @@ To configure a private registry scoped to a particular organization:
 "@myorg3" = { token = "$npm_token", url = "https://registry.myorg.com/" }
 ```
 
+### Changing the Mirror Source in Bun
```
xhyrom: Why creating another section when there's already one?
https://github.com/oven-sh/bun/blob/214b3ccca052203211e18f9d55796549c135edf3/docs/install/registries.md?plain=1#L1-L11

docs/install/registries.md
```diff
@@ -25,6 +25,92 @@ To configure a private registry scoped to a particular organization:
...
+
+# set the mirror source URL with username/password
+registry = "https://username:password@your-mirror-source-url"
+```
+
+4. Save the `bunfig.toml` file.
+
+For more detailed information on configuring the mirror source and other registry settings, refer to the [custom registry guide](../guides/install/custom-registry.md).
+
+### Available Mirror Registry
```
xhyrom: I don't see a reason why we should maintain a list of available npm mirrors. We're recommending them but have no guarantee that they are/will be OK.

docs/install/registries.md
arvinxx: or maybe we can update the instructment with real registry like `npmmirror` ?

docs/install/registries.md
arvinxx: So maybe we can just update the word of the title to `Recommended Mirror Registry` ?

xhyrom: I don't think that will solve the problem. We shouldn't provide any mirror where we cannot guarantee safe operation.

docs/install/registries.md
xhyrom: npmjs is not a real registry? Or what do you mean?

docs/install/registries.md
arvinxx: We should add recommend mirror registery. Or this PR is invalid

Jarred-Sumner:
@arvinxx this is a good start, thank you for this. I will make some changes tomorrow but I think we definitely at minimum need better instructions on how to change the registry settings because people are still confused by this

docs/install/registries.md
arvinxx: plz check the update. It's what I mean

arvinxx:
@Jarred-Sumner I have updated a little, feel free to give your advice.

xhyrom:
I don't see any reason why this pull request should be merged. There's already a section that explains how to change default registry https://bun.sh/docs/install/registries at the top of this page. Why we're recommending mirrors that we cannot guarantee safe operation? That's hilarious. And why we're providing Chinese documentation? I guess every Chinese programmer knows at least A1 English (or know what translator is) and knows how to use Google to find Chinese npm mirrors.

arvinxx:
> There's already a section that explains how to change the default registry at the top of the https://bun.sh/docs/install/registries page. We can improve the existing section rather than adding another.

I think you may not know how to write a good document. A good document for the newbie user follows a structure of Basic description and then an Example , finally Pro Tips. The top part as I think is the Basic , and what I add is the exmaple part to make it more clear with a real mirror case, which will be useful for developers to just copy/paste to solve their problems.

>  why do we want to recommend mirrors that we cannot guarantee safe operation?

I think it's opinionated. Why bun support ts/tsx natively while node not support? Why bun support CJS & ESM import natively? It's all opinionated for the better developer experience. I think recommanding mirrors can be more friendly for the new bun user because they can solve problems just in the official document other than search solution in Google. Actually I was confused when I started to config the registry, I figure out with several try to solve the mirror issue, and the experience is not such smooth. I propose this PR only because @Jarred-Sumner 's #12936 , it's his thought to make bun more smoothly and I really admire it. This PR is the embodiment of his thought.

> why do we want to maintain Chinese documentation for one section? I guess every Chinese programmer knows at least A1 English (or knows what a translator is) and knows how to use Google to find Chinese npm mirrors. If we want localized documentation we should do it properly however that's out of scope.

It's also the embodiment of #12936 . Due the there is no i18n for document so the only way is to add a Chinese part at the end and folded by default. 

And your guess is wrong. Chinese programmer who don't know English well even don't use Google. And they use baidu like this: https://www.baidu.com/s?wd=bun%20%E9%95%9C%E5%83%8F%E6%BA%90&rsv_spt=1&rsv_iqid=0x97ee74cd00005291&issp=1&f=8&rsv_bp=1&rsv_idx=2&ie=utf-8&tn=baiduhome_pg&rsv_dl=tb&rsv_enter=1&rsv_sug3=1&rsv_sug1=1&rsv_sug7=100&rsv_sug2=0&rsv_btype=i&prefixsug=bun%2520%25E9%2595%259C%25E5%2583%258F%25E6%25BA%2590&rsp=6&inputT=512&rsv_sug4=512

xhyrom:
> I think you may not know how to write a good document. A good document for the newbie user follows a structure of Basic description and then an Example , finally Pro Tips. The top part as I think is the Basic , and what I add is the exmaple part to make it more clear with a real mirror case, which will be useful for developers to just copy/paste to solve their problems.

The documentation is not a step-by-step guide. You also won't become a programmer if you just copy and paste. There is nothing wrong with the current documentation.

> I think it's opinionated. Why bun support ts/tsx natively while node not support? Why bun support CJS & ESM import natively? It's all opinionated for the better developer experience. I think recommanding mirrors can be more friendly for the new bun user because they can solve problems just in the official document other than search solution in Google. Actually I was confused when I started to config the registry, I figure out with several try to solve the mirror issue, and the experience is not such smooth.

I think you don't understand what I mean by "guarantee safe operation". You're recommending third-party mirrors from which you download packages. They can infect millions of computers like a recent issue with polyfill.io. Those third-party mirrors are not under our control thus we shouldn't recommend them. I'm sure you don't want to be responsible if something goes wrong with any of these mirrors because you're recommending them.

> I propose this PR only because @Jarred-Sumner 's #12936 , it's his thought to make bun more smoothly and I really admire it. This PR is the embodiment of his thought.

> It's also the embodiment of #12936 . Due the there is no i18n for document so the only way is to add a Chinese part at the end and folded by default.

Then we should make i18n rather than adding sections under a page with random languages.

> And your guess is wrong. Chinese programmer who don't know English well even don't use Google. And they use baidu like this: https://www.baidu.com/s?wd=bun%20%E9%95%9C%E5%83%8F%E6%BA%90&rsv_spt=1&rsv_iqid=0x97ee74cd00005291&issp=1&f=8&rsv_bp=1&rsv_idx=2&ie=utf-8&tn=baiduhome_pg&rsv_dl=tb&rsv_enter=1&rsv_sug3=1&rsv_sug1=1&rsv_sug7=100&rsv_sug2=0&rsv_btype=i&prefixsug=bun%2520%25E9%2595%259C%25E5%2583%258F%25E6%25BA%2590&rsp=6&inputT=512&rsv_sug4=512

You have to learn along the way. Everybody starts somehow.

arvinxx:
> Documentation is not a step-by-step guide. 

A good documentation can be a step-by-step guide. You can't make good product if you don't polish document for developers.  Good documentation make the product stand out.

> You also won't become a programmer if you just copy and paste.

If you and @Jarred-Sumner make the same product, I will use the product jarred make just because of he allowing me copy and paste, while you just want to teach me how to became a programmer. Everyone is busy, no one want to figure out something tiny with much struggle, they just give up,and you lost your users. Try to think about the DX firstly , not the life experience please.

> Then we should make i18n rather than adding sections under a page with random languages.

If you decide to make i18n, I agree to remove this part. If not , I still think add Chinese part will be really useful.

xhyrom:
> A good documentation can be a step-by-step guide.

That's guide, not documentation. Documentation should be straightforward.

> If you and @Jarred-Sumner make the same product, I will use the product jarred make just because of he allowing me copy and paste, while you just want to teach me how to became a programmer. Everyone is busy, no one want to figure out something tiny with much struggle, they just give up,and you lost your users. Try to think about the DX firstly , not the life experience please.

Sorry but if someone struggles with copying two lines and replacing one line with another url then he's not capable of being programmer.

docs/install/registries.md
```diff
@@ -25,6 +25,92 @@ To configure a private registry scoped to a particular organization:
...
+
+---
+
+<details>
+
+<summary>
+针对中国大陆的用户的配置指南
+</summary>
+
+由于中国特定的网络审查环境，npm 官方的注册表可能无法正常使用。以下是配置文档
```
7086cmd, line 78: You might modify it to "由于中国大陆特定的网络环境，`npm` 官方源可能在特定情况下无法正常访问。您可以按照如下方式手动切换镜像源。" to avoid some troubles.

docs/install/registries.md
```diff
@@ -25,6 +25,92 @@ To configure a private registry scoped to a particular organization:
...
+
+```toml
+[install]
+# 设置带有令牌的镜像源 URL
+registry = { url = "https://registry.npmmirror.com", token = "your-token" }
+```
+
+4. 保存 `bunfig.toml` 文件。
+
+有关配置镜像源和其他注册表设置的详细信息，请参阅[自定义注册表指南](../guides/install/custom-registry.md)。
```
7086cmd, line 104: Using "注册表" is a bit improper (it is what `regedit` in Windows edits).

docs/install/registries.md
```diff
@@ -25,6 +25,92 @@ To configure a private registry scoped to a particular organization:
...
+
+### 推荐的镜像源
+
+根据 #12936 中的讨论，以下是一些推荐的镜像源：
+
+- `https://registry.npmmirror.com`
+- `https://mirrors.cloud.tencent.com/npm/`
+- `https://repo.huaweicloud.com/repository/npm/`
+
+请注意，我们的推荐不保证可用性。
```
7086cmd, line 114: 请注意，我们并不保证以上地址能够一直使用。
Or a better translated version.

7086cmd (commented):
Based on the native language habits, I have provided some suggestions for enhancing the translation and mitigating "potential" issues.

Issue #1257, opened 2022-09-20
Segfault caused by macro oddities


Validark:
### Version

0.1.13

### Platform
```
Linux 5be22774b6dc 5.19.0-76051900-generic #202207312230~1660780566~22.04~9d60db1 SMP PREEMPT_DYNAMIC Thu A x86_64 x86_64 x86_64 GNU/Linux
```
### What steps will reproduce the bug?

This only happens on some runs, so one has to run `bun dev` and spam F5 to get the segfault. 

```bash
bun create blank ./_new_test
```

src/index.js:
```js
// import { graphql } from 'react-relay';

// graphql(['a', 'b']);
console.log("Hello via Bun!");

const a = { a: 1, b: 2, c: 3, d: 4, e: 5 }.b;
// console.log(a.b)

const date = new Date(1996, 8, 1, 0, 0, 0, 0).toUTCString();
const realDate = new Date(date).toISOString();

console.log(date)

console.log(typeof process.env.QUIET3, `"${process.env.QUIET3}"`);
console.log(typeof !process.env.QUIET3, !process.env.QUIET3);
console.log(typeof !!!!process.env.QUIET, !!!!process.env.QUIET);
console.log(typeof process.env.QUIET, `"${process.env.QUIET}"`);
console.log(typeof process.env.bagel, process.env.bagel);
console.log(typeof process.env.donut, process.env.donut);
console.log(typeof process.env.pie, process.env.pie);
console.log(typeof process.env.pie2, process.env.pie2);
console.log(typeof process.env.pie3, process.env.pie3);
console.log(typeof process.env.pie4, process.env.pie4);
// console.log(typeof process.env.pie5, process.env.pie5);
const k = 0.1.toString();
console.log(k)
const { bagel: bagelboi } = process.env;
const { env: { QUIET } } = process
console.log(QUIET)
const b = JSON.parse('{"a":1,"b":2,"c":3,"d":4}');
console.log(b.c);
process.env.pie5.b;
```

Make a `bunfig.toml` file:

```toml
# Set a default framework to use
# By default, bun will look for an npm package like `bun-framework-${framework}`, followed by `${framework}`

macros.react-relay.graphql = "bun-macro-relay"
macros.relay-runtime.graphql = "bun-macro-relay"

[define]
# Replace any usage of "process.env.bagel" with the string `lox`.
# The values are parsed as JSON, except single-quoted strings are supported and `'undefined'` becomes `undefined` in JS.
[...]

PR #13045, opened 2024-08-03
pass unhandled rejections to uncaughtException when no unhandledRejection handler exists

nektro:
also added many more tests for uncaughtException and unhandledRejection handling


robobun:
@nektro, your commit c728e5ed2408e85655656d961957764740169864 is building: [`#1257`](https://buildkite.com/bun/bun/builds/1257)

PR #1258, opened 2022-09-20
Add more tags to be usable with typeof

Validark:

PR #13047, opened 2024-08-03
feat: `bun build --compile --sourcemap`

### LINKED ISSUE #12804: `bun build --compile` sourcemaps don't work
Reproduction from #12571, should still reproduce if instead of the genuine bundler error, you place a manual `throw new Error` into the code

paperclover:
Fixes #12804 

Seems we had advertised this as working on the docs, but the code was never fully wired up.

Though, instead of storing JSON into the final binary, a custom binary format is used to avoid having to decode string literals at runtime. This makes the bundling process take a minuscule amount of extra time, but is worth it considering there is nearly zero sourcemap decoding step (the main step is to just allocate and fill the `ParsedSourceMap`, all the data is ready).

Before merging, I think this needs some more tests, and some more internal doc-comments are needed.

Also, I deleted Lock.init in favor of `.{}`


robobun:
@paperdave, your commit c3096f8f76e50205289bea59d367fae35a8901d5 is building: [`#1258`](https://buildkite.com/bun/bun/builds/1258)

Jarred-Sumner (changes requested):
Can we use StringPointer instead of making up another type for pointing to a string from a buffer?

Can we make all this `align(1)` so we don't worry about under-aligning things?

paperdave:
> Can we make all this align(1) so we don't worry about under-aligning things?

We can do this for String Pointer but if we do this for the multi-array-list payload, it would not be valid for the pointer in `MultiArrayList(Mapping)` as it must be `align(@alignOf(Mapping))`. This is why I am careful with alignment in this PR, as not having it aligned would require the data to be cloned into an aligned portion of memory.

When memory is assumed to be aligned when it is not, Bun will crash. It is not hard to align data.

src/sourcemap/sourcemap.zig
```diff
@@ -267,6 +272,10 @@ pub const Mapping = struct {
 
             const name = source_map.external_source_names[@intCast(lookup.mapping.source_index)];
 
+            if (source_map.is_standalone_module_graph) {
+                return bun.String.createUTF8(bun.path.joinAbs("/", .auto, name));
```
Jarred-Sumner: looks suspicious for Windows

src/StandaloneModuleGraph.zig
```diff
@@ -270,12 +320,23 @@ pub const StandaloneModuleGraph = struct {
...
+                defer source_map_header_list.clearRetainingCapacity();
+                defer source_map_string_list.clearRetainingCapacity();
+                _ = source_map_arena.reset(.retain_capacity);
+                try serializeJsonSourceMapForStandalone(
+                    &source_map_header_list,
+                    &source_map_string_list,
+                    source_map_arena.allocator(),
+                    output_files[output_file.source_map_index].value.buffer.bytes,
+                );
+                string_builder.ensureNextAligned(@alignOf(u32));
```
Jarred-Sumner: can we zero the whole thing instead of zeroing it 4 bytes at a time?

src/StandaloneModuleGraph.zig
```diff
@@ -270,12 +320,23 @@ pub const StandaloneModuleGraph = struct {
...
+                var bytes = string_builder.allocatedSlice()[string_builder.len..];
```
Jarred-Sumner: can you use the string builder instead of doing this? this is technical debt

src/StandaloneModuleGraph.zig
```diff
@@ -824,4 +889,180 @@ pub const StandaloneModuleGraph = struct {
...
+            };
+            try out.writeInt(u32, slice.offset, .little);
+            try out.writeInt(u32, slice.length, .little);
+        }
+
+        for (sources_content.items.slice()) |item| {
+            if (item.data != .e_string)
+                return error.InvalidSourceMap;
+
+            const utf16_decode = try bun.js_lexer.decodeStringLiteralEscapeSequencesToUTF16(item.data.e_string.string(arena) catch bun.outOfMemory(), arena);
```
Jarred-Sumner: this should be a method on E.String

like

.decodedUTF8(allocator)

src/StandaloneModuleGraph.zig
```diff
@@ -824,4 +889,180 @@ pub const StandaloneModuleGraph = struct {
...
+        const string_payload_start_location = @sizeOf(u32) +
+            @sizeOf(u32) +
+            @sizeOf(bun.StringPointer) * sources_content.items.len * 2 + // path + source
+            map_vlq.len;
+
+        for (sources_paths.items.slice()) |item| {
+            if (item.data != .e_string)
+                return error.InvalidSourceMap;
+
+            const utf16_decode = try bun.js_lexer.decodeStringLiteralEscapeSequencesToUTF16(item.data.e_string.string(arena) catch bun.outOfMemory(), arena);
```
Jarred-Sumner: ditto

src/StandaloneModuleGraph.zig
```diff
@@ -824,4 +885,175 @@ pub const StandaloneModuleGraph = struct {
...
+                const size = bun.zstd.getDecompressedSize(compressed_file);
+
+                const bytes = bun.default_allocator.alloc(u8, size) catch bun.outOfMemory();
+                const result = bun.zstd.decompress(bytes, compressed_file);
+
+                if (result == .err) {
+                    bun.Output.warn("Source map decompression error: {s}", .{result.err});
+                    bun.default_allocator.free(bytes);
+                    this.decompressed_files[index] = "";
+                    return "";
```
Jarred-Sumner: This should return null so that the parent is aware the source map contents was null instead of an empty string

src/StandaloneModuleGraph.zig
```diff
@@ -824,4 +885,175 @@ pub const StandaloneModuleGraph = struct {
...
+                }
+
+                const data = bytes[0..result.success];
+                if (data.len != bytes.len) {
+                    _ = bun.default_allocator.resize(bytes, data.len);
```
Jarred-Sumner: resize won't reduce memory usage here, seems like a footgun

src/js_ast.zig
```diff
@@ -2479,6 +2479,12 @@ pub const E = struct {
             }
         }
 
+        pub fn stringDecodedUTF8(s: *const String, allocator: std.mem.Allocator) !bun.string {
+            const utf16_decode = try bun.js_lexer.decodeStringLiteralEscapeSequencesToUTF16(try s.string(allocator), allocator);
```
Jarred-Sumner, line 2498: worst case this clones the string 3 times

Jonatthu:
@Jarred-Sumner bun build src/index.ts --target bun --compile --minify --sourcemap --outfile ./build/apps/server/app
With this command on sentry I still get 
image

Jarred-Sumner:
@Jonatthu prior to this PR, that likely would've shown a callstack pointing to the internal compiled file, or just "processTicksAndRejections"

what you're seeing looks like an issue where we're not preserving async stack frames from napi, which would be unrelated to this PR. 

can you file an issue?

Jonatthu:
I am using latest bun 1.1.24 on azure linux container 
@Jarred-Sumner should I file to where?
Another example:
image

Issue #1259, opened 2022-09-20
'0A' in bunfig.toml [define] macro crashes bun


Validark:
### Version

0.1.13

### Platform

```Linux 5be22774b6dc 5.19.0-76051900-generic #202207312230~1660780566~22.04~9d60db1 SMP PREEMPT_DYNAMIC Thu A x86_64 x86_64 x86_64 GNU/Linux```

### What steps will reproduce the bug?

```bash
bun create blank ./_new_test
cd _new_test
touch bunfig.toml
echo "[define]\n'process.env.pie5' = '0A'" > bunfig.toml
bun dev
```

Then open localhost:3000/ (refresh the page if applicable)

### How often does it reproduce? Is there a required condition?

It crashes every time

### What do you see instead?

```
std.hash_map.HashMapUnmanaged([]const u8,[]src.defines.DotDefine,std.hash_map.StringContext,80).getIndex (/build/zig/lib/std/hash_map.zig:1131)
std.hash_map.HashMapUnmanaged([]const u8,[]src.defines.DotDefine,std.hash_map.StringContext,80).getAdapted (/build/zig/lib/std/hash_map.zig:1261)
std.hash_map.HashMapUnmanaged([]const u8,[]src.defines.DotDefine,std.hash_map.StringContext,80).getContext (/build/zig/lib/std/hash_map.zig:1258)
std.hash_map.HashMap([]const u8,[]src.defines.DotDefine,std.hash_map.StringContext,80).get (/build/zig/lib/std/hash_map.zig:589)
src.js_parser.NewParser_(false,src.js_parser.JSXTransformType.react,false).visitExprInOut (/build/bun/src/js_parser.zig:14608)
src.js_parser.NewParser_(false,src.js_parser.JSXTransformType.react,false).visitExprInOut (/build/bun/src/js_parser.zig:14862)
src.js_parser.NewParser_(false,src.js_parser.JSXTransformType.react,false).visitExpr (/build/bun/src/js_parser.zig:12713)
src.js_parser.NewParser_(false,src.js_parser.JSXTransformType.react,false).visitAndAppendStmt (/build/bun/src/js_parser.zig:16297)
src.js_parser.NewParser_(false,src.js_parser.JSXTransformType.react,false).visitStmts (/build/bun/src/js_parser.zig:17891)
[...]

PR #13048, opened 2024-08-03
Add nullability annotations

Jarred-Sumner:
### What does this PR do?

Add nullability annotations to usockets headers

In release builds, it uses `__attribute__((nonnull))`, which allows the compiler to optimize it.

In debug builds, it uses `_Nonnull` which does not enable compiler optimizations for nonnull. In debug builds, also enabled `fsanitize=null`.

Ran some tests. Merging this is mostly dependent on what CI says about it.


robobun:
@Jarred-Sumner, your commit efce5760728b50463d3302b37e6516cd5c0bec5f is building: [`#1259`](https://buildkite.com/bun/bun/builds/1259)

PR #13058, opened 2024-08-03
properly propogate exit code if process onexit handler throws

nektro:
found this while investigating running nodejs test suite

### DIFF
--- a/test/js/node/process/process.test.js
+++ b/test/js/node/process/process.test.js
@@ -377,6 +377,30 @@ describe("process.onBeforeExit", () => {
     expect(exitCode).toBe(0);
     expect(stdout.toString().trim()).toBe("beforeExit: 0\nbeforeExit: 1\nexit: 2");
   });
+
+  it("throwing inside preserves exit code", async () => {
+    const proc = Bun.spawnSync({
+      cmd: [bunExe(), "-e", `process.on("beforeExit", () => {throw new Error("boom")});`],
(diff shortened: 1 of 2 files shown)


robobun:
@nektro, your commit 12b40fd3dd98a8e0898d837f39d1c9c23c9fe29c is building: [`#1265`](https://buildkite.com/bun/bun/builds/1265)

Jarred-Sumner:
Thank you

test/js/node/process/process.test.js
```diff
@@ -377,6 +377,30 @@ describe("process.onBeforeExit", () => {
     expect(exitCode).toBe(0);
     expect(stdout.toString().trim()).toBe("beforeExit: 0\nbeforeExit: 1\nexit: 2");
   });
+
+  it("throwing inside preserves exit code", async () => {
```
Electroid, line 381: * what if the exit code is non-0 and non-1? does it become 1 or keep the non-1?
* what if the error is thrown in a signal handler?

nektro: this exits with 4 in both node and bun, i believe this path is only for when the exit is otherwise clean

```js
process.exitCode = 4;
process.on("exit", code => {
  console.log("process", "exit", code, process.exitCode);
  throw new Error("boom");
});
```

if u find a repro where the exit code is clobbered and differs from node im happy to look into it

Issue #1266, opened 2022-09-21
macOS catalina unsupported in bun 0.1.13


kt3k:
### Version

0.1.13

### Platform

Darwin kt3k-mac.local 19.6.0 Darwin Kernel Version 19.6.0: Thu May  6 00:48:39 PDT 2021; root:xnu-6153.141.33~1/RELEASE_X86_64 x86_64

### What steps will reproduce the bug?

run any script with `bun run script.js`

### How often does it reproduce? Is there a required condition?

100%

### What is the expected behavior?

run script

### What do you see instead?

crash

### Additional information

```
$ cat a.js 
console.log("A");
$ bun run a.js
dyld: lazy symbol binding failed: Symbol not found: _pthread_self_is_exiting_np
  Referenced from: /Users/kt3k/.bun/bin/bun (which was built for Mac OS X 11.0)
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _pthread_self_is_exiting_np
  Referenced from: /Users/kt3k/.bun/bin/bun (which was built for Mac OS X 11.0)
  Expected in: /usr/lib/libSystem.B.dylib


Crash at 0xDE5EFF0E01000000


----- bun meta -----
Bun v0.1.13 (55bdf268) macOS x64 19.6.0
RunCommand: 
Elapsed: 2ms | User: 2ms | Sys: 3ms
RSS: 3.81MB | Peak: 3.81MB | Commit: 67.11MB | Faults: 3
----- bun meta -----
```

Jarred-Sumner:
Looks like we are missing support for macOS Catalina due to this missing symbol

Jarred-Sumner:
webkit has dropped support for their memory allocator on macOS 10.xx as of the most recent webkit upgrade

need to think on what to do about this

https://github.com/oven-sh/WebKit/commit/677b453daf6855bb2601d776532503e1c07ddc90

EliteCheng:
+1

chrisAXZA:
Same issue here on Mojave (0.1.13) =>
System Version: macOS 10.14.6 Kernel Version: Darwin 18.7.0

baih-npu:
how to change version of bun?

Mouvedia:
This crash is still a blocker for me.

Bun: 0.5.1
macOS: 10.13.6

chrisAXZA:
Due to webkit issues, bun is so far running only on Mac OS `11.x.x` versions and above.

[...]

PR #13061, opened 2024-08-03
fix(build): retry webkit download on failure, resume download with curl

billywhizz:
### What does this PR do?

When building, the Webkit precompiled library is quite large, especially for debug build. If it fails for whatever reason, it will try to download from scratch when re-running the build. This is painful if you are on a slow/unreliable internet connection.

There is also a bug where build will continue on next run when it finds a partially downloaded webkit library and will subsequently fail due to the library being broken.

This change will do the following:

- tell curl to resume downloading from where it left off if it finds a partial download from a previous attempt
- write the webkit library to a temporary file until it is completely downloaded, then rename to the expected name for the rest of the build process
- retry the curl download up to 5 times before failing
- uses http/1.1 for curl, which tends to be more reliable ime

Note: Would have to investigate if same thing is possible on Windows - this works on Linux and Mac.

### DIFF
--- a/scripts/download-webkit.sh
+++ b/scripts/download-webkit.sh
@@ -43,11 +43,34 @@ fi
 
 rm -rf "$OUTDIR"
 
+download () {
+  local command="$1"
+  local retries="$2"
+  local options="$-"
+  if [[ $options == *e* ]]; then
+    set +e
+  fi
+  $command
+  local exit_code=$?
+  if [[ $options == *e* ]]; then
+    set -e
+  fi
+  if [[ $exit_code -ne 0 && $retries -gt 0 ]]; then
+    download "$command" $(($retries - 1)) 
+  else
+    return $exit_code
+  fi
+}
+
+# this is a big download so we will retry 5 times and ask curl to resume
+# download from where failure occurred if it fails and is rerun
 if [ ! -f "$tar" ]; then
   echo "-- Downloading WebKit"
-  if ! curl -o "$tar" -L "$url"; then
+  if ! download "curl -C - --http1.1 -o $tar.tmp -L $url" 5; then
     echo "Failed to download $url"
     exit 1
+  else
+    mv $tar.tmp $tar
   fi
 fi
 


robobun:
Your commit e8103df0956fa0daf4fa0c681f8786a24627b1f6 is building: [`#1266`](https://buildkite.com/bun/bun/builds/1266)

scripts/download-webkit.sh
```diff
@@ -43,11 +43,34 @@ fi
...
+    return $exit_code
+  fi
+}
+
+# this is a big download so we will retry 5 times and ask curl to resume
+# download from where failure occurred if it fails and is rerun
 if [ ! -f "$tar" ]; then
   echo "-- Downloading WebKit"
-  if ! curl -o "$tar" -L "$url"; then
+  if ! download "curl -C - --http1.1 -o $tar.tmp -L $url" 5; then
```
Jarred-Sumner, line 69: This is nice, but I wonder if we could use the `--retry` option in curl instead of doing this manually?

```sh
curl --retry 5 --retry-all-errors -o $tar.tmp -L $url
```

billywhizz: lol. yes. i'll change it to use that.

Jarred-Sumner: tbh I didn't know about it before googling

billywhizz: will have to test it out to see if it works when combined with the resuming downloads flag. this suggests it will start from scratch on each retry.
https://github.com/curl/curl/issues/1084

billywhizz: i did a few experiments with curl using tcpkill to kill the tcp connection for an active download. this is what i found, and it seems to agree with what the manpages for curl say. upshot i think is the method in this PR is only way to resume partial downloads from where they left off reliably.

1. ```--retry N``` only deals with http request status/timeout errors, not tcp errors for an in-flight download
2. ```--retry-all-errors``` will always throw away existing partial download when it gets a tcp/network error mid-download (log below)

```shell
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  7 57.8M    7 4455k    0     0   305k      0  0:03:13  0:00:14  0:02:59  288k
curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104
Warning: Problem (retrying all errors). Will retry in 4 seconds. 3 retries 
Warning: left.
Throwing away 4562597 bytes
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  4 57.8M    4 2701k    0     0   557k      0  0:01:46  0:00:04  0:01:42  659k
```

billywhizz: the curl manpages actually say this re. ```--retry-all-errors```:

https://curl.se/docs/manpage.html


This option is the "sledgehammer" of retrying. Do not use this option by default (for example in your curlrc), there may be unintended consequences such as sending or receiving duplicate data. Do not use with redirected input or output. **You might be better off handling your unique problems in a shell script**. Please read the example below.

Issue #13034, opened 2024-08-02
Worker from node:worker_threads constructed with eval option evaluates filename as code


joas8211:
### What version of Bun is running?

1.1.21+70ca2b76c

### What platform is your computer?

Linux 6.6.40-2-lts x86_64 unknown

### What steps can reproduce the bug?

Running following code:
```typescript
import { Worker } from "node:worker_threads";

new Worker("console.log('this should not get printed')", { eval: false });
``` 

### What is the expected behavior?

Following error is expected:
```
TypeError: BuildMessage: ModuleNotFound resolving "console.log('this should not get printed')" (entry point)
```

### What do you see instead?

Following is printed to console instead:
```
this should not get printed
```

### Additional information

This happens only with `eval` option is set for `Worker` with any value. I tried with `true`, `false`, `undefined`.

billywhizz:
i have taken an initial look at this and can reproduce it. working on a fix.

Issue #11801, opened 2024-06-12
Running eslint hangs


sebdanielsson:
### What version of Bun is running?

1.1.13+bd6a60512

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

packages:

```json
"devDependencies": {
    "@typescript-eslint/parser": "^7.13.0",
    "eslint": "^8.57.0",
    "eslint-plugin-astro": "^1.2.2",
    "postcss-html": "^1.7.0",
    "prettier": "^3.3.2",
    "prettier-plugin-astro": "^0.14.0",
    "prettier-plugin-tailwindcss": "^0.6.3",
    "stylelint": "^16.6.1",
    "stylelint-config-html": "^1.1.0",
    "stylelint-config-standard": "^36.0.0",
    "typescript": "^5.4.5",
    "typescript-eslint": "^7.13.0"
  }
```

.eslintrc.cjs:

```js
/** @type {import("@types/eslint").Linter.Config} */
module.exports = {
  extends: ['plugin:astro/recommended', 'plugin:@typescript-eslint/strict'],
  overrides: [
    {
      files: ['*.astro'],
      parser: 'astro-eslint-parser',
      parserOptions: {
        parser: '@typescript-eslint/parser',
        extraFileExtensions: ['.astro'],
      },
      rules: {},
    },
  ],
};
```

Command:

```shell
bunx --bun eslint "src/**/*.{astro,mjs,cjs,js,ts}"
```

### What is the expected behavior?

Linting should run.

### What do you see instead?

Hangs at line:

```
2024-06-12T08:57:16.484Z eslint:linter Apply the processor: 'astro/.astro'
```

Command: `DEBUG=eslint:* bunx --bun eslint "src/**/*.{astro,mjs,cjs,js,ts}"`

<details>
  <summary>Show logs</summary>
  
```
2024-06-12T08:57:16.088Z eslint:cli CLI args: [ 'src/**/*.{astro,mjs,cjs,js,ts}' ]
2024-06-12T08:57:16.096Z eslint:cli Using flat config? false
2024-06-12T08:57:16.099Z eslint:cli Running on files
2024-06-12T08:57:16.108Z eslint:file-enumerator Start to iterate files: [ 'src/**/*.{astro,mjs,cjs,js,ts}' ]
2024-06-12T08:57:16.108Z eslint:file-enumerator Glob: src/**/*.{astro,mjs,cjs,js,ts}
[...]

PR #11537, opened 2024-06-02
Implement `blob:` URLs in `fetch`, `import`, and `Worker`

Jarred-Sumner:
### What does this PR do?

This lets you use `fetch`, `import`, and `new Worker` on `blob:` URLs.

Worker:
```ts
const code = `
    const foo: number = 123;
    postMessage(foo);
`;
const blob = new File([code], "hey.ts");
const url = URL.createObjectURL(blob);
const worker = new Worker(url);
worker.onmessage = ({data}) => {
   console.log("<", data);
};
```

fetch:

```ts
const response = await fetch("https://");
const blob = await response.blob();
const url = URL.createObjectURL(blob);
const again = await fetch(url);
if (await again.text() === await blob.text()) {
   console.log("yep, it worked");
}
```

Fixes #3925 
Fixes #4165 
Fixes #9816 

### How did you verify your code works?

Tests

PR #13062, opened 2024-08-03
fix (worker-eval): fail worker with source when eval = false

billywhizz:
### What does this PR do?

ensures that creating a Worker with source code and options.eval = false fails with expected error.

Fixes: #13034
Fixes: #11801 (Need to Confirm)

This issue seems to have been introduced [here](https://github.com/oven-sh/bun/pull/11537) and was released in [1.1.13](https://github.com/oven-sh/bun/releases/tag/bun-v1.1.13), which tallies with the original report of the issue in #11801.

All tests which were introduced in the [Blob URL PR](https://github.com/oven-sh/bun/pull/11537) are passing, and new tests have been added to verify this change works as expected.

### DIFF
--- a/src/js/node/worker_threads.ts
+++ b/src/js/node/worker_threads.ts
@@ -216,9 +216,15 @@ class Worker extends EventEmitter {
 
     const builtinsGeneratorHatesEval = "ev" + "a" + "l"[0];
     if (options && builtinsGeneratorHatesEval in options) {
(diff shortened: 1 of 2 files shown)


robobun:
Your commit 4798d73415d63993eb4c3da082f2e095d463821e is building: [`#1268`](https://buildkite.com/bun/bun/builds/1268)

Issue #11739, opened 2024-06-10
After compiling to an executable file using bun and importing a JSON file, the Chinese value will be displayed as garbled, but the Chinese key of an object will not become garbled


Azzellz:
### What version of Bun is running?

1.1.12

### What platform is your computer?

Linux 5.15.0-86-generic x86_64 x86_64

### What steps can reproduce the bug?

Code:
![image](https://github.com/oven-sh/bun/assets/117060720/e2b200e1-8094-4f51-986d-58a7a156402f)
The JSON file reads as follows:
![image](https://github.com/oven-sh/bun/assets/117060720/ec79dacd-94f4-4e1d-a975-12a11ebe423a)
Output:
![image](https://github.com/oven-sh/bun/assets/117060720/4b782b3d-8273-41b4-94d5-5a7bdab8b134)

### Additional information

Could it be related to this [issue](https://github.com/oven-sh/bun/issues/9559#)?

Issue #1272, opened 2022-09-23
bun completion: do not overide my dotfiles


undg:
### What is the problem this feature will solve?

When I'm generating completions by bun, .zshrc is changed. It's my little .zshrc and strangers should not touch him. He can grow up on pervert or something.

I'm using separate folders for completions: generated, written manually and pulled from the zsh-completion community. I'm sourcing them in my way. I don't want bun to touch my dotfiles to source one completion.

### What is the feature you are proposing to solve the problem?

1. Add an extra flag that will prevent overriding personal files.
2. Add an extra flag that will override fish config or .zshrc and don't touch it by default. (I prefer this one)

### What alternatives have you considered?

IDK what I can do. 
Not updating bun? Not really a solution. 
chmoding zshrc? Bit overkill.
As far as I know, completion is regenerated after an update. I'm installing it from AUR, not sure if that issue is present when updating it by YAY. I'll look at this closer after the next update.

Jarred-Sumner:
What if I make it look for this string in `.zshrc` and skip saving it if so?

> \# bun: don't auto-source completions

undg:
Yes, that will be perfect. Whatever is sensible and doesn't involve much effort. Thank you.

undg:
That comment flag can be auto added next to source(before, after 👐). It will require no documentation to figure out what's what by simply seeing it.

If I'm not wrong, you don't need to change the path every time an update is made. Unless I'm wrong and the path for the autocompletion file is not stable yet.

jason0x43:
> What if I make it look for this string in .zshrc and skip saving it if so?

Other tools I use (deno, 1password-cli, rtx) that generate completions simply output them to stdout and don't modify any files. That's simpler for the tool, and it's easy for the user to put the completions in the proper place on the local system (such as a zsh/site-functions directory). It'd be a nice option.

iloveitaly:
Big +1 on this one. Would be great to have a zsh plugin that does something like:

```
bun completions # generate but don't mutate zshrc
fpath+=~/.bun/
```

pesterhazy:
This bit me hard because it creates a conflict with [zsh-autocomplete](https://github.com/marlonrichert/zsh-autocomplete)

With bun completions enabled:

```
~$ % arch(eval):1: command not found: _autocomplete.command
(eval):1: command not found: _autocomplete.command
(eval):1: command not found: _autocomplete.command
(eval):1: command not found: _autocomplete.command
[...]

PR #13064, opened 2024-08-04
Make signal.abort() from native code fast

Jarred-Sumner:
### What does this PR do?

Make signal.abort() from native code fast.

Also:
- Redo how we create errors internally. Error objects with a unique `code` property use a cached prototype.
- We use `WTF::ASCIILiteral` from strings created in Zig now. This should reduce memory usage due to less copying. However, it means we have to add a null-terminated sentinel 0 for each usage. This requires extra care in our code.
- Uses an `ExternalStringImpl` for source contents from `bun build --compile`, which fixes https://github.com/oven-sh/bun/issues/11739 because previously our usage of String.init assumed the contents was UTF-8, and then re-encoded latin1 as UTF-8 since it had non-ascii characters in it.
- Add exception checks to some code that was missing them. 

Ran some tests locally.

Extended the existing express body parser --compile test to also validate JSON & text imports.


robobun:
@Jarred-Sumner, your commit a9ea41f433123899321536ec98e98d1ca395e3cc is building: [`#1272`](https://buildkite.com/bun/bun/builds/1272)

src/bun.js/webcore/streams.zig
```diff
@@ -682,9 +682,32 @@ pub const StreamResult = union(Tag) {
...
+            Strong,
+            Weak,
+        };
+
+        pub fn toJSWeak(this: *const @This(), globalObject: *JSC.JSGlobalObject) struct { JSC.JSValue, WasStrong } {
+            return switch (this.*) {
+                .Error => |err| {
+                    return .{ err.toJSC(globalObject), WasStrong.Weak };
+                },
+                .JSValue => .{ this.JSValue, WasStrong.Strong },
```
Jarred-Sumner, line 703: @cirospaciari flagging this for your attention

src/bun.js/webcore/streams.zig
```diff
@@ -943,13 +966,12 @@ pub const StreamResult = union(Tag) {
         defer loop.exit();
 
         switch (result.*) {
-            .err => |err| {
+            .err => |*err| {
```
Jarred-Sumner, line 969: @cirospaciari flagging this for your attention

src/codegen/generate-node-errors.ts
```diff
@@ -0,0 +1,128 @@
+import NodeErrors from "../bun.js/bindings/ErrorCode.ts";
```
Jarred-Sumner, line 1: @paperdave flagging this for your attention

src/bun.js/api/server.zig
```diff
@@ -5316,7 +5309,7 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
             return if (request.request_context.getRemoteSocketInfo()) |info|
                 JSSocketAddress__create(
                     this.globalThis,
-                    bun.String.static(info.ip).toJS(this.globalThis),
+                    bun.String.init(info.ip).toJS(this.globalThis),
```
Jarred-Sumner, line 5312: @cirospaciari this was a bug

CMakeLists.txt
```diff
@@ -704,6 +704,32 @@ add_custom_command(
 )
 list(APPEND BUN_RAW_SOURCES "${BUN_WORKDIR}/codegen/ZigGeneratedClasses.cpp")
 
+if(NOT NO_CODEGEN)
```
Jarred-Sumner, line 707: @paperdave flagging this for your attention

src/bun.js/api/server.zig
cirospaciari: surprised that we did not catch it earlier, I really did not remember when we added requestIP

src/bun.js/webcore/streams.zig
cirospaciari: nice, way better to follow, but could this be called more than one time? I think we only call this only one time so we don't double unprotect it in the strong case

CMakeLists.txt
Electroid: Is this needed for building Zig object, C++ objects, or both?

Jarred-Sumner: both

cirospaciari (approved):

src/codegen/generate-node-errors.ts
```diff
@@ -0,0 +1,128 @@
...
+static constexpr ErrorCodeData errors[${NodeErrors.length}] = {
+`;
+
+zig = `
+// Generated by: src/codegen/generate-node-errors.ts
+const std = @import("std");
+const bun = @import("root").bun;
+const JSC = bun.JSC;
+
+fn ErrorBuilder(comptime code_: Error, comptime fmt_: [:0]const u8, Args: type) type {
```
paperclover, line 42: these variables do not need to be named `code_` and `fmt_`, they can just be named `code` and `fmt`. not a big deal because generated code.

paperclover (approved):
re: `src/codegen/generate-node-errors.ts` and that error system: this is great.

PR #13077, opened 2024-08-04
Try to fix formatting in rendered guides document

guest271314:
### What does this PR do?

- [X ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

I didn't. Not sure what is causing the rendered text and code to be offset. I just added 3 dashes. Somebody should probably check that this does what it is inteded to do before merging.

Screenshot_2024-08-04_07-51-41

### DIFF
--- a/docs/guides/runtime/import-json.md
+++ b/docs/guides/runtime/import-json.md
@@ -27,6 +27,16 @@ data.version; // => "1.0.0"
 data.author.name; // => "John Dough"
 ```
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Yeah this should probably fix it. Thanks for the follow up

guest271314:
Unfortunately that doesn't fix the offset. Not sure why. 
Screenshot_2024-08-04_09-01-58

Issue #1297, opened 2022-10-06
No stdin/stdout read/write documentation or tests


guest271314:
### Description of the problem

No stdin/stdout read/write documentation or tests.

I am writing a Native Messaging host using bun.

I can't locate any tests nor documentations specific to use of I/O; reading from stdin and writing to stdout.

Using 

```
// this is a ReadableStream
var reader = Bun.stdin.stream().getReader();

Bun.write(
  Bun.stdout,
  "text" // or Bun.file() or ArrayBufferView
)
```

results in the Native Messaging host exiting in error.

Working QuickJS https://github.com/guest271314/native-messaging-quickjs, and Deno https://github.com/guest271314/native-messaging-deno implementations using the same algorithm.

guest271314:
From README.md

```
interface Bun {
  write(
    destination: string | number | FileBlob,
    input: string | FileBlob | Blob | ArrayBufferView
  ): Promise<number>;
}
```

This print to terminal

```
#!bun
Bun.write(Bun.stdout, 'test');
```

this doens't

```
#!bun
const output = new Uint8Array([1]);
Bun.write(Bun.stdout, output.buffer);
```

Jarred-Sumner:
Yes, this needs very explicit & clear docs. For most usecases, `console.log` is the best way to write to the terminal.

```
#!bun
const output = new Uint8Array([1]);
Bun.write(Bun.stdout, output.buffer);
```

Are you sure this one doesn't print to the terminal? `1` would print the non-ascii 1

guest271314:
I am sure nothing prints to the terminal on *nix.

I encountered this trying to implement a Native Messaging host using Bun, which exits early. So I tested each paty of the code and found `write()` doesn't print anything for a `TypedArray`.

I use the same algorthim for QuickJS and Deno implementations, just changing the read/write methods.

Test for yourself as a Native Messaging host.

```
#!bun
// Bun Native Messaging host
// guest271314, 10-5-2022

async function getMessage() {
  const header = new Uint32Array(1);
  const reader = Bun.stdin.stream().getReader();
  const {value, done} = await reader.read();
  header.set(value.subarray(0, 4));
  const output = new Uint8Array(header[0]);
  output.set(value.subarray(4));
  return output;
}

async function sendMessage(message) {
  const header = Uint32Array.from(
    {
      length: 4,
    },
    (_, index) => (message.length >> (index * 8)) & 0xff
  );
  const output = new Uint8Array(header.length + message.length);
[...]

PR #13078, opened 2024-08-04
fix(server)

cirospaciari:
We should only ref/deref when:

ref():
-   Promise pending
-   Reading file

deref():
-   Finishing reading a file
-   Finishing a pending promise (reject/resolved)
-   End the resquest response or Abort the request 

We should also not touch the request context after we responded.
### What does this PR do?

- [X] Code changes

Manually running debug build to check assertions


robobun:
@cirospaciari, your commit 352c351e57dd76fcf36945731c7c24955618684d is building: [`#1297`](https://buildkite.com/bun/bun/builds/1297)

Issue #13071, opened 2024-08-04
Failed to start debugger. Program could not be started.


mkdirJava:
### What version of Bun is running?

1.1.21

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Env Setup 

Visual Studio Code ( Version: 1.92.0 )
Bun for Visual Studio Code Extension, Oven ( Version: v0.0.12,  extension Id oven.bun-vscode)
Code found in 
https://github.com/mkdirJava/BunIssueReplication

Steps

Clone the above repo
Open in Vscode
Whilst main panel has index.ts open run Vscode debug "Debug File"
Open "Debug Console" in the bottom panel ( might have to open a terminal and find the tab )
"Debug Console" reports
"Failed to start debugger.
Program must be a JavaScript or TypeScript file."

Looked further into the issue via logs found in 
C:\Users\<user>\AppData\Roaming\Code\logs\20240804T094251\window1\exthost\exthost.log

Looks like the extension is trying to open a reserved port

2024-08-04 09:46:15.465 [error] Error: listen EACCES: permission denied C:\Users\<user>\AppData\Local\Temp\0vyubmzqvx4g.sock
	at Server.setupListenHandle [as _listen2] (node:net:1881:21)

![image](https://github.com/user-attachments/assets/e162b680-81e1-4dc2-b10a-a2f73a59bd1d)

### What is the expected behavior?

Given I have the repo cloned
https://github.com/mkdirJava/BunIssueReplication
And the Bun Extension installed
When I have the index.ts file open
And I have a break point set
And I click debug in Vscode
Then I should hit the debug line in debug mode

### What do you see instead?

Failed to start debugger.
Program must be a JavaScript or TypeScript file.

### Additional information

Looked further into the issue via logs found in 
C:\Users\<user>\AppData\Roaming\Code\logs\20240804T094251\window1\exthost\exthost.log

Looks like the extension is trying to open a reserved port

2024-08-04 09:46:15.465 [error] Error: listen EACCES: permission denied C:\Users\<user>\AppData\Local\Temp\0vyubmzqvx4g.sock

liudonghua123:
I got the similar logs in `%appdata%\Code\logs\20240805T003523\window2\exthost\exthost.log`

```shell
[...]

PR #13099, opened 2024-08-05
13071 ensure unix path when running in windows

mkdirJava:
### What does this PR do?

Fix for 
https://github.com/oven-sh/bun/issues/13071

method:
randomUnixPath()
found in 
packages\bun-debug-adapter-protocol\src\debugger\signal.ts

makes use of 
import { tmpdir } from "node:os";
which returns a temp path, since the method is called "randomUnixPath" it can create a Windows path, which causes issues in debugging

- [ ] Code changes

### DIFF
--- a/packages/bun-debug-adapter-protocol/package.json
+++ b/packages/bun-debug-adapter-protocol/package.json
@@ -4,5 +4,8 @@
   "dependencies": {
     "semver": "^7.5.4",
     "source-map-js": "^1.0.2"
+  },
+  "devDependencies": {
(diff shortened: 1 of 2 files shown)


mkdirJava:
Hey @Electroid,
I figured it would be complicated. I noticed the source comes from a lib that mentioned it produces path with respect to OS platform. 

Therefore it makes me think. Is the debugging feature of Bun being built in a two steps? Witht he first fixed to an Unix system?

Happy to be wrong but wanted to understand.

Issue #1343, opened 2022-10-17
Deduplicate / dedupe command for bun install


ArnaudBarre:
### What is the problem this feature will solve?

Reduce the total number of installed dependencies after a adding/bumping a dependency.

Example:
First installed: `module-a` request `esbuild ^0.15.7`, `module-b` request `esbuild ^0.15.8`
`esbuild v0.15.10` is installed

Few days later, bumping `module-a`, still request `esbuild ^0.15.7`
Both `esbuild v0.15.10`  and `esbuild v0.15.11` are installed

### What is the feature you are proposing to solve the problem?

Builtin version of https://github.com/scinos/yarn-deduplicate

We could also have installation algorithms that prioritise installed version over last available version (related to https://github.com/oven-sh/bun/issues/118)

### What alternatives have you considered?

On small projects, I delete the lock and start from a fresh install.
On larger project with tens of dependencies used in production, I keep using yarn so that I can safely bump a dev dependency and then deduplicate the install of esbuild for example without touching the exact version of my UI dependencies.

Edit: I built https://github.com/ArnaudBarre/bun-dedupe in the meantime.

steinybot:
I can't seem to upgrade React because of this. I want to upgrade to 18.3.1 but something else is pulling in 18.2.0.

Having 2 reacts fails rather spectacularly with things like:
![Screenshot 2024-08-06 at 2 45 02 PM](https://github.com/user-attachments/assets/c89fbcca-54d6-4dcb-bb2a-c57209626d16)

Writing the dependencies back out to yarn format has:
```
react@*, "react@0.14 - 18", react@>=16.0.0, react@>=16.11.0, react@>=16.13.1, react@>=16.8, react@>=16.8.4, react@>=16.9.0, "react@^16.13.1 || ^17.0.0 || ^18.0.0", "react@^16.8 || ^17 || ^18", "react@^16.8.0 || 17.x", "react@^16.8.0 || ^17.0.0 || ^18.0.0", "react@^16.9.0 || ^17", react@^18.2.0:
  version "18.2.0"
  resolved "https://registry.npmjs.org/react/-/react-18.2.0.tgz"
  integrity sha512-/3IjMdb2L9QbBdWiW5e3P2/npwMBaU9mHCSCUzNln0ZCYbcfTsGbTJrU/kGemdH2IWmB2ioZ+zkxtmq6g09fGQ==
  dependencies:
    loose-envify "^1.1.0"

[...]

PR #13106, opened 2024-08-06
Fix assigning to process.default

Jarred-Sumner:
### What does this PR do?

Fix assigning to process.default

Test

### DIFF
--- a/test/js/node/process/process.test.js
+++ b/test/js/node/process/process.test.js
@@ -724,6 +724,20 @@ it("process.execArgv", async () => {
   }
 });
 
+// node does this in tests.
+it("process.default can be set", () => {
+  expect([
+    `
+      process.default = 58;
+      import("node:process").then(a => {
+        Bun.inspect(a);
+      });
+    `,
+    "",
+    0,
+  ]).toRunInlineFixture();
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit d0207aa3ddce75f079811c66a28f8952bad50dc7 is building: [`#1343`](https://buildkite.com/bun/bun/builds/1343)

test/js/node/process/process.test.js
```diff
@@ -724,6 +724,20 @@ it("process.execArgv", async () => {
   }
 });
 
+// node does this in tests.
+it("process.default can be set", () => {
+  expect([
+    `
+      process.default = 58;
+      import("node:process").then(a => {
+        Bun.inspect(a);
```
nektro, line 733:
```suggestion
        require('node:assert').equal(a.default.default, 58);
```

Jarred-Sumner: Is there a specific reason to use node assert here?

nektro: it was more about ensuring we do the check that `a.default.default === 58`

PR #1347, opened 2022-10-19
Improved readme readability and project first impression

scarletquasar:
## Summary

- Improved readme readability with easier access to information like quick view of the features, discord server and table of contents. Also re-assembled the first part to give a better "first impression" of the project

PR #13107, opened 2024-08-06
fix ref count of source provider when printing error contents

paperclover:
### What does this PR do?

before, for every error remapped, bun would gather the source lines from JSC in case it needed to print it to console. now, that is only done when the error is actually getting printed and there is no disk source.

additionally, the old code always put a ref on the source provider, and i dont think all cases handled this properly, resulting in a ref-count leak. cannot reproduce a test that leaked memory on older version.


robobun:
@paperdave, your commit f0ff0f267d18ed54a29e57bea669acdc2814b687 is building: [`#1347`](https://buildkite.com/bun/bun/builds/1347)

src/bun.js/bindings/bindings.cpp
```diff
@@ -4256,78 +4256,102 @@ static void populateStackFramePosition(const JSC::StackFrame* stackFrame, BunStr
...
-        if (*referenced_source_provider != nullptr) {
-            (*referenced_source_provider)->deref();
-        }
-        *referenced_source_provider = provider;
-        source_lines[0] = Bun::toStringView(sourceString.substring(lineStart, lineEnd - lineStart));
-        source_line_numbers[0] = location.line();
+    JSC::StackFrame& frame = stack->first();
+    JSC::SourceProvider* provider = frame.codeBlock()->source().provider();
+    WTF::StringView view = provider->source();
+    ;
```
dylan-conway, line 4286:
```suggestion
```
stray semicolon

dylan-conway (approved):
I think this is fine

paperclover:
appears to be covered in a later PR, just not as cleanly

Issue #1348, opened 2022-10-19
Fix WebSocket client stability issues


Jarred-Sumner:
Bun's `WebSocket` client does not seem perform as well under load on Linux as it does on macOS. We need to dig into why.

Note that this is unrelated to the WebSocket server because the code is entirely different.

Jarred-Sumner:
Running this file on Linux with a server running: https://github.com/oven-sh/bun/blob/main/bench/websocket-server/chat-client.mjs

reproduces the bug

Electroid:
Duplicate of #6686

PR #13108, opened 2024-08-06
node:worker_threads: fix assertion when require is used with 'eval:true'

nektro:

added new test coverage

### DIFF
--- a/src/bun.js/module_loader.zig
+++ b/src/bun.js/module_loader.zig
@@ -1687,7 +1687,7 @@ pub const ModuleLoader = struct {
                     }
                 }
 
-                var parse_result = switch (disable_transpilying or
+                var parse_result: ParseResult = switch (disable_transpilying or
                     (loader == .json and !path.isJSONCFile())) {
                     inline else => |return_file_only| brk: {
                         break :brk jsc_vm.bundler.parseMaybeReturnFileOnly(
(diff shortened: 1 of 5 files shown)


robobun:
@nektro, your commit 151da52eee3e90e0e08575dd1898a4fa7d8b44b2 is building: [`#1348`](https://buildkite.com/bun/bun/builds/1348)

src/bun.js/module_loader.zig
```diff
@@ -1939,7 +1939,7 @@ pub const ModuleLoader = struct {
                     if (parse_result.ast.exports_kind == .cjs and parse_result.source.path.isFile()) {
                         const actual_package_json: *PackageJSON = package_json orelse brk2: {
                             // this should already be cached virtually always so it's fine to do this
-                            const dir_info = (jsc_vm.bundler.resolver.readDirInfo(parse_result.source.path.name.dir) catch null) orelse
+                            const dir_info = (jsc_vm.bundler.resolver.readDirInfo(parse_result.source.path.name.dirOrDot()) catch null) orelse
```
Jarred-Sumner, line 1942: Why does this work? This becomes a `blob:` URL, right?

Jarred-Sumner (changes requested):
Can you add a test that requires a file that exists on disk and a test that throws an error? `"process"` will always work

PR #13109, opened 2024-08-06
node:buffer: fix assertion when unexpected types are passed to constructor functions

nektro:

added new test coverage

### DIFF
--- a/src/bun.js/bindings/JSBuffer.cpp
+++ b/src/bun.js/bindings/JSBuffer.cpp
@@ -2415,6 +2415,7 @@ JSC_DEFINE_HOST_FUNCTION(constructJSBuffer, (JSC::JSGlobalObject * lexicalGlobal
             RELEASE_AND_RETURN(throwScope, JSC::JSValue::encode(uint8Array));
         }
         case ArrayType:
+        case DerivedArrayType:
             break;
         default: {
             auto arg_string = distinguishingArg.toWTFString(globalObject);
--- a/src/bun.js/bindings/JSBuffer.cpp
+++ b/src/bun.js/bindings/JSBuffer.cpp
@@ -2283,6 +2283,20 @@ JSC_DEFINE_HOST_FUNCTION(constructJSBuffer, (JSC::JSGlobalObject * lexicalGlobal
     if (distinguishingArg.isAnyInt()) {
         throwScope.release();
         return JSBuffer__bufferFromLength(lexicalGlobalObject, distinguishingArg.asAnyInt());
+    } else if (distinguishingArg.isDouble()) {
+        double lengthDouble = distinguishingArg.toIntegerWithTruncation(lexicalGlobalObject);
+        size_t length = static_cast<size_t>(lengthDouble);
+
+        if (UNLIKELY(length < 0 || length > 9007199254740991)) {
+            throwNodeRangeError(lexicalGlobalObject, throwScope, "Buffer size must be 0...9007199254740991"_s);
+            return {};
+        }
+        return JSBuffer__bufferFromLength(lexicalGlobalObject, length);
+    } else if (distinguishingArg.isUndefinedOrNull()) {
+        auto arg_string = distinguishingArg.toWTFString(globalObject);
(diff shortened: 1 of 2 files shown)


robobun:
@nektro, your commit b8e388e573d0784a2b0c3a919a2496001161ec01 is building: [`#1349`](https://buildkite.com/bun/bun/builds/1349)

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2283,6 +2283,20 @@ JSC_DEFINE_HOST_FUNCTION(constructJSBuffer, (JSC::JSGlobalObject * lexicalGlobal
     if (distinguishingArg.isAnyInt()) {
         throwScope.release();
         return JSBuffer__bufferFromLength(lexicalGlobalObject, distinguishingArg.asAnyInt());
+    } else if (distinguishingArg.isDouble()) {
```
```suggestion
    } else if (distinguishingArg.isNumber()) {
```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2400,8 +2414,14 @@ JSC_DEFINE_HOST_FUNCTION(constructJSBuffer, (JSC::JSGlobalObject * lexicalGlobal
 
             RELEASE_AND_RETURN(throwScope, JSC::JSValue::encode(uint8Array));
         }
-        default: {
+        case ArrayType:
```
Jarred-Sumner, line 2417: Is that the only type we need to worry about here? What about DerivedArrayType? There are likely 10 more

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2283,6 +2283,20 @@ JSC_DEFINE_HOST_FUNCTION(constructJSBuffer, (JSC::JSGlobalObject * lexicalGlobal
...
+    } else if (distinguishingArg.isDouble()) {
+        double lengthDouble = distinguishingArg.toIntegerWithTruncation(lexicalGlobalObject);
+        size_t length = static_cast<size_t>(lengthDouble);
+
+        if (UNLIKELY(length < 0 || length > 9007199254740991)) {
+            throwNodeRangeError(lexicalGlobalObject, throwScope, "Buffer size must be 0...9007199254740991"_s);
+            return {};
+        }
+        return JSBuffer__bufferFromLength(lexicalGlobalObject, length);
+    } else if (distinguishingArg.isUndefinedOrNull()) {
```
Jarred-Sumner: Are there other primitives we need to concern ourselves with? What about booleans?

src/bun.js/bindings/JSBuffer.cpp
nektro: good catch, will add DerivedArrayType

image

Jarred-Sumner: there are more, it should really be a "if we don't have any other things to match against" instead of assuming

src/bun.js/bindings/JSBuffer.cpp
nektro: wouldnt any int thats not double be caught by .isAnyInt() ?

Jarred-Sumner:
```c++
inline bool JSValue::isDouble() const
{
    return isNumber() && !isInt32();
}
```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2400,8 +2414,15 @@ JSC_DEFINE_HOST_FUNCTION(constructJSBuffer, (JSC::JSGlobalObject * lexicalGlobal
 
             RELEASE_AND_RETURN(throwScope, JSC::JSValue::encode(uint8Array));
         }
-        default: {
+        case ArrayType:
+        case DerivedArrayType:
             break;
+        default: {
```
Jarred-Sumner, line 2420: What happens if it's an iterable? What happens if its an object which implements arraySpecies?

nektro:
going to pull anything worthwhile into new pr

Issue #1350, opened 2022-10-19
adapter error in Axios


faizalom:
`adapter is not a function. (In 'adapter(config)', 'adapter' is undefined)`

I'm getting this error when i tries to run axios function

Electroid:
Thanks for reporting @faizalom, could you provide a quick code sample so we can reproduce?

faizalom:
> Thanks for reporting @faizalom, could you provide a quick code sample so we can reproduce?

@Electroid Please check this code snippet

```
const axios = require('axios');

axios.post('https://swapi.dev/api/')
    .then(function (response) {
        // handle success
        console.log(response);
    })
    .catch(function (error) {
        // handle error
        console.log(error);
    });
```
Error which I am getting
![Please check this](https://user-images.githubusercontent.com/15194288/197034248-a301d274-2f23-49c0-8a30-63b9c7774f3c.PNG)

laffed:
bump

xhyrom:
I know this is something we have to solve in time but I recommend to use the built-in `fetch` and also axios is one of the slowest http clients. @faizalom

zero41120:
`OpenAIApi` happens to use `Axios`. 

Here is sample code: 
```javascript
import { Configuration, OpenAIApi } from "openai";
import dotenv from "dotenv";

dotenv.config();

const configuration = new Configuration({
  apiKey: process.env.OPENAI_API_KEY,
});

const openai = new OpenAIApi(configuration);

const response = await openai.createCompletion({
  model: "text-davinci-003",
  prompt: "Convert movie titles into emoji.\n\nBack to the Future: 👨👴🚗🕒 \nBatman: 🤵🦇 \nTransformers: 🚗🤖 \nStar Wars:",
  temperature: 0.8,
  max_tokens: 60,
  top_p: 1.0,
  frequency_penalty: 0.0,
  presence_penalty: 0.0,
  stop: ["\n"],
});
console.log(response.data.choices.map(c => c.text));
```

Node runs this without problem
```
$ node -v
v19.6.0
$ node index.js 
[ '🚀⭐🤴👸' ]
```

Bun has this adaptor issue
```
$ bun -v
0.5.6
$ bun index.js 
[0.01ms] ".env"
53 |     }
54 |   );
55 | 
56 |   var adapter = config.adapter || defaults.adapter;
57 | 
58 |   return adapter(config).then(function onAdapterResolution(response) {
             ^
[...]

PR #13110, opened 2024-08-06
js: dont leak as many globals as enumerable

nektro:

test coverage for this will come in a followup soon when i get the suite running in our CI again

### DIFF
--- a/src/bun.js/bindings/ZigGlobalObject.cpp
+++ b/src/bun.js/bindings/ZigGlobalObject.cpp
@@ -3433,8 +3433,8 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
         PropertyAttribute::Accessor | 0);
 
     // TODO: this should be usable on the lookup table. it crashed las time i tried it
-    putDirectCustomAccessor(vm, JSC::Identifier::fromString(vm, "onmessage"_s), JSC::CustomGetterSetter::create(vm, globalOnMessage, setGlobalOnMessage), 0);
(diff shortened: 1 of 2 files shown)


robobun:
@nektro, your commit d8442bd48f80e88f4828e137f50a93e6d6b3a421 is building: [`#1350`](https://buildkite.com/bun/bun/builds/1350)

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3433,8 +3433,8 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
         PropertyAttribute::Accessor | 0);
 
     // TODO: this should be usable on the lookup table. it crashed las time i tried it
-    putDirectCustomAccessor(vm, JSC::Identifier::fromString(vm, "onmessage"_s), JSC::CustomGetterSetter::create(vm, globalOnMessage, setGlobalOnMessage), 0);
-    putDirectCustomAccessor(vm, JSC::Identifier::fromString(vm, "onerror"_s), JSC::CustomGetterSetter::create(vm, globalOnError, setGlobalOnError), 0);
+    putDirectCustomAccessor(vm, JSC::Identifier::fromString(vm, "onmessage"_s), JSC::CustomGetterSetter::create(vm, globalOnMessage, setGlobalOnMessage), 1 << 2);
```
Jarred-Sumner, line 3436: What are these bit fields? These should be dontEnum

Jarred-Sumner:
I think its nice that when you do console.log(globalThis) it shows a bunch of stuff Bun has built in

I don't think this is a good idea to change

PR #1351, opened 2022-10-19
Cache dir loader: Prefer `$BUN_INSTALL` and `$XDG_CACHE_HOME` to `$HOME`.

lgarron:
This partially addresses https://github.com/oven-sh/bun/issues/696 , by using `$XDG_CACHE_HOME` for those of us who already have that env var set.

PR #13111, opened 2024-08-06
node:fs: fix assertion when Readable/Writable are passed non-numbers for start/end

nektro:

status quo throws an error in this situation but this should fix the debug mode assertion and make the error much more clear

### DIFF
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -97,21 +97,6 @@ pub const JSObject = extern struct {
         });
     }
 
-    extern fn Bun__ERR_INVALID_ARG_TYPE(*JSGlobalObject, JSValue, JSValue, JSValue) JSValue;
-    pub fn ERR_INVALID_ARG_TYPE(this: *JSGlobalObject, arg_name: JSValue, etype: JSValue, atype: JSValue) JSValue {
-        return Bun__ERR_INVALID_ARG_TYPE(this, arg_name, etype, atype);
-    }
-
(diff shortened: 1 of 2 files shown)


robobun:
@nektro, your commit c21904fe334f9c223600e5a5ae3b8db73fb20910 is building: [`#1351`](https://buildkite.com/bun/bun/builds/1351)

src/bun.js/bindings/bindings.zig
```diff
@@ -3250,6 +3235,21 @@ pub const JSGlobalObject = extern struct {
         if (bun.Environment.allow_assert) this.bunVM().assertOnJSThread();
     }
 
+    extern fn Bun__ERR_INVALID_ARG_TYPE(*JSGlobalObject, JSValue, JSValue, JSValue) JSValue;
+    pub fn ERR_INVALID_ARG_TYPE(this: *JSGlobalObject, arg_name: JSValue, etype: JSValue, atype: JSValue) JSValue {
```
Jarred-Sumner, line 3239: We should not be creating temporary JSValue just for strings.

src/js/node/fs.ts
```diff
@@ -836,6 +837,8 @@ function ReadStream(this: typeof ReadStream, pathOrFd, options) {
 
   // Get the stream controller
   // We need the pointer to the underlying stream controller for the NativeReadable
+  if (start && typeof start !== "number") throw ERR_INVALID_ARG_TYPE("options.start", "number", start);
+  if (end && typeof end !== "number") throw ERR_INVALID_ARG_TYPE("options.end", "number", end);
```
Jarred-Sumner, line 841: Can you fix this in `.stream(offset)`? The code here isn't causing an assertion failure.

PR #13113, opened 2024-08-06
node:fs: fix assertion when chown is passed non-numbers

nektro:

status quo throws an error in this situation but this should fix the debug mode assertion and make the error much more clear


robobun:
@nektro, your commit ebe3e06497b322124da50e39cd106f94cff80c13 is building: [`#1353`](https://buildkite.com/bun/bun/builds/1353)

src/bun.js/node/node_fs.zig
```diff
@@ -1444,6 +1444,14 @@ pub const Arguments = struct {
                 };
 
                 arguments.eat();
+                if (!uid_value.isNumber()) {
+                    ctx.throwValue(ctx.ERR_INVALID_ARG_TYPE(
+                        JSC.ZigString.static("uid").toJS(ctx),
```
Jarred-Sumner: We should not be creating JSString just to throw the values

src/bun.js/bindings/bindings.zig
```diff
@@ -3196,6 +3181,31 @@ pub const JSGlobalObject = opaque {
         if (bun.Environment.allow_assert) this.bunVM().assertOnJSThread();
     }
 
+    extern fn Bun__ERR_INVALID_ARG_TYPE_static(*JSGlobalObject, *const ZigString, *const ZigString, JSValue) JSValue;
+    /// Caller asserts 'arg_name' and 'etype' are utf-8 literals.
+    pub fn ERR_INVALID_ARG_TYPE_static(this: *JSGlobalObject, arg_name: *const ZigString, etype: *const ZigString, atype: JSValue) JSValue {
+        return Bun__ERR_INVALID_ARG_TYPE_static(this, arg_name, etype, atype);
+    }
+
+    pub const Extern = [_][]const u8{
```
Jarred-Sumner: this was deleted intentionally

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -224,6 +225,20 @@ extern "C" JSC::EncodedJSValue Bun__ERR_INVALID_ARG_TYPE(JSC::JSGlobalObject* gl
     auto message = makeString("The \""_s, arg_name, "\" argument must be of type "_s, expected_type, ". Received "_s, actual_value);
     return JSValue::encode(createError(globalObject, ErrorCode::ERR_INVALID_ARG_TYPE, message));
 }
+extern "C" JSC::EncodedJSValue Bun__ERR_INVALID_ARG_TYPE_static(JSC::JSGlobalObject* globalObject, const ZigString* val_arg_name, const ZigString* val_expected_type, JSC::EncodedJSValue val_actual_value)
+{
+    JSC::VM& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+
+    auto arg_name = std::span<const unsigned char>(val_arg_name->ptr, val_arg_name->len);
```
Jarred-Sumner, line 233: ptr is a tagged pointer. It could be UTF-16. It could be UTF-8. It could be Latin1.

I suggest using the helper for converting to a WTF::String or a WTF::StringView instead of this because it will crash if we send it any user-defined data

Jarred-Sumner: unless we want static to mean "ascii only", in which case we should add an assertion

Issue #1367, opened 2022-10-22
Segfault when getting `err.position` on an error thrown from resolving a specific dynamic import setup.


paperclover:
### What version of Bun is running?

0.2.1 canary as of 2022-10-22 01:15 EST

### What steps can reproduce the bug?

dont ask how i figured this out but there's a segfault here

```ts
// a.ts
try {
  await import(`./b`);
} catch (err) {
  console.log(err.position); // segfault!
}
```
```ts
// b.ts
import "./nonexistant";
```

### What do you see instead?

```
SegmentationFault at 0x0800000000000000


----- bun meta -----
Bun v0.2.1 (c940f00e) Linux x64 #1 SMP PREEMPT_DYNAMIC Tue, 04 Oct 2022 14:36:58 +0000
AutoCommand: 
Elapsed: 15ms | User: 12ms | Sys: 4ms
RSS: 67.11MB | Peak: 73.70MB | Commit: 67.11MB | Faults: 0
----- bun meta -----
```

paperclover:
fixed in 8d67870da45175956cb742ae805b11a63ae097ca

PR #13119, opened 2024-08-06
fix(bundler): handle assigning to `exports`

paperclover:
### What does this PR do?

This is a correction for #13017, which does not handle all cases.

I did not want to add an ast node for this but I don't think it's possible any other way. `e_module_dot_exports` represents `module.exports`, which is either lowered into `exports` or `module.exporst`.

Consider

```
let w = () => [module.exports, module.exports.xyz];
let x = () => [exports, exports.xyz];
// exports = { xyz: 123 };
// module.exporst = { xyz: 456 };
let y = () => [module.exports, module.exports.xyz];
let z = () => [exports, exports.xyz];
```

if line 3 and 4 stay commented out, all module.exports rewrite into `exports`. but once we see an assignment to **either one** of these fields, we must retroactively fix all instances of it, as there are behavioral differences if we substitute module.exports.

### DIFF
--- a/src/bun.zig
+++ b/src/bun.zig
@@ -3366,6 +3366,13 @@ pub fn assertWithLocation(value: bool, src: std.builtin.SourceLocation) callconv
 
 /// This has no effect on the real code but capturing 'a' and 'b' into parameters makes assertion failures much easier inspect in a debugger.
 pub inline fn assert_eql(a: anytype, b: anytype) void {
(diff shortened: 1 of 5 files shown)


robobun:
@paperdave, your commit b78168d38099f856a8dba659a59a5f9e1f474800 is building: [`#1367`](https://buildkite.com/bun/bun/builds/1367)

PR #13132, opened 2024-08-06
fix: Fix bun build for programs that depend on global require

### LINKED ISSUE #5809: Bundling (with target: 'bun') outputs code that is incompatible with bun for certain libraries

Linux 6.2.0-32-generic x86_64 x86_64

1. Use a library such as `gluegun` or `enquirer`
2. Bundle with `target: 'bun'`
3. bun run the output file

```
55345 |     }
55346 |     path3 = nodePath.normalize(path3);
55347 |     if (appModulePaths.indexOf(path3) === -1) {
55348 |       appModulePaths.push(path3);
55349 | 
55350 |       if (require.main) {
             ^
ReferenceError: Can't find variable: require
      at addPath (/home/karl/git/teamsykmelding-cli/tsm-cli/bin/tsm:55350:10)
      at /home/karl/git/teamsykmelding-cli/tsm-cli/bin/tsm:55456:2
      at /home/karl/git/teamsykmelding-cli/tsm-cli/bin/tsm:38:74
      at /home/karl/git/teamsykmelding-cli/tsm-cli/bin/tsm:70651:29

```

Running code directly with bun run, and running it bundled with `bun build` should both work.

### What do you see instead?

It crashes on run, see stack trace above.

### Additional information

Code that bundles wrong:
https://github.com/navikt/teamsykmelding-cli/blob/feat/sync-file/src/actions/sync-file.ts#L2

Bun.build with target as bun:
https://github.com/navikt/teamsykmelding-cli/blob/feat/sync-file/src/build.ts#L4

Serdnad:
### What does this PR do?

This fixes #5809 by adding `var {require} = import.meta;` to the beginning of the generated js file as suggested in the thread, immediately after any hashbang and `// @bun` if applicable.

- [X] Code changes

Verified that
1. Existing codebase that worked with `bun run` failed to run (pre PR changes) after `bun build`, both with and without `--compile`.
2. Same codebase ran correctly after the change and `bun build`, both with and without `--compile`.
3. Output js when built without `--compile` is identical to before the changes except for added line at top.

TODO: Update existing tests

### DIFF
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -6923,6 +6923,11 @@ pub const LinkerContext = struct {
                 j.pushStatic("// @bun\n");
                 line_offset.advance("// @bun\n");
             }
+
+            // Inject `require` definition to support the common use case of
+            // modules expecting it to be defined in the global scope.
+            j.pushStatic("var { require } = import.meta;\n");
+            line_offset.advance("var { require } = import.meta;\n");
         }
 
         // TODO: banner


Serdnad:
Question for reviewers: this is a very literal implementation of Jared's [suggestion](https://github.com/oven-sh/bun/issues/5809#issuecomment-1938056991) - should this be restricted to certain conditions, e.g. based on the selected target?

I plan to update existing tests affected by this addition, but wanted to first sanity check that I'm not overlooking anything major.

paperdave:
thanks for the initiative, but this is a workaround instead of an actual fix. blindly inserting this as suggested is not really the correct thing to do here.

normally `require` is converted to `__require`, which has dedicated codegen depending on --target

```zig
        // When the `require` identifier is visited, it is replaced with e_require_call_target
        // and then that is either replaced with the module itself, or an import to the
        // runtime here.
        const runtime_require = switch (target) {
            // __require is intentionally not implemented here, as we
            // always inline 'import.meta.require' and 'import.meta.require.resolve'
            // Omitting it here acts as an extra assertion.
            .bun, .bun_macro => "",

            .node =>
            \\import { createRequire } from "node:module";
            \\export var __require = /* @__PURE__ */ createRequire(import.meta.url);
            \\
            ,
            
            // ...
            .browser => ...
```

`require` when seen is rewritten as the `e_require_call_target` ast node, which will either:

- point to `__require` (rewritten as `import.meta.require` for `--target=bun`)
- if wrapped (see the `__commonJS` function in runtime.js / how it gets bundled), one of the arguments to this function is to be named `require`.

i believe the bug here is bun is we are not emitting the argument for `require` because we think its not used. could be wrong; i have not looked exactly at the reproduction.

the first step for this bug is to minimize a specific reproduction (it should be 1 or 2 files / about 10 lines), then we can add an exact test case for this, and make sure the usage of require bundles correctly not just for --target=bun but for all targets.

paperdave:
i have a suspicion the emitted `require` is this branch:

image

Serdnad:
firstly, thank you for the very thorough response.

> i believe the bug here is bun is we are not emitting the argument for require because we think its not used.

this is my theory as well. unfortunately, in my particular case, it's due to a dependency that's trying to use `require` within an eval, as in `const lib = eval('require("crypto")');`. short of attempting to parse the contents of eval expressions, I'm not quite sure how bun could've noticed that a definition for `require` needed to be added.

that said, I noticed something potentially interesting while creating the following repro:
```javascript
import {} from "js-sha256" // version "0.9.0"

console.log("hello world!")
```

with the above:
- `node repro.js` -> outputs 'hello world!'
- `bun run repro.js` -> outputs 'hello world!'
- `bun build ...` -> 
```
1 | require('crypto')
           ^
ReferenceError: Can't find variable: require
```

looking at the unminified output from `bun build repro.ts --outdir repro --target bun`, bun is including the entirety of the js-sha256 dependency, but I would've expected the bundler to tree shake out the dependency entirely. could this be a separate bug?

paperdave:
ah, direct eval. i remember something about this.

```zig
                                // Pessimistically assume that if this looks like a CommonJS module
                                // (e.g. no "export" keywords), a direct call to "eval" means that
                                // code could potentially access "module" or "exports".
                                if (p.options.bundle and !p.is_file_considered_to_have_esm_exports) {
                                    p.recordUsage(p.module_ref);
                                    p.recordUsage(p.exports_ref);
                                }
```

maybe we should make this consider `require` too. i'm not sure what implications this has, but theres a chance adding p.require_ref here might just do the whole thing, but it might need some more glue. mainly to check how it plays with minify identifiers (the eval call causes all reachable identifiers to prevent renaming).

> I would've expected the bundler to tree shake out the dependency entirely. would this be a separate bug worth filing?

only in typescript files, importing values that are only used in types will tree-shake the import statement away blindly. i did notice that bun seems to treat `import {} from '...'` is treated like `import '...'` which is supposed to include the file.

a separate part is after a file is included, parts of the file without side effects can be removed, if you look at node_modules/js-sha256/src/sha256.js, you'll see it is a top level IIFE and does module.exports inside, which currently has no special handling

---

i played a bit with bundling eval with require. my suggestion is part of it, but bundling as is

```
console.log(require);
console.log(eval('require("crypto")'));
```

will produce (this is --target=browser)

```
var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
  get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
}) : x)(function(x) {
  if (typeof require !== "undefined")
    return require.apply(this, arguments);
  throw Error('Dynamic require of "' + x + '" is not supported');
});

// index.js
var require_scratchpad_20240808T214956 = __commonJS((exports, module) => {
  console.log(__require);
  console.log(eval('require("crypto")'));
});
export default require_scratchpad_20240808T214956();
```

i was first going to suggest renaming __require to require, but the browser implementation of __require checks for a globally defined `require` to use (this is taken from esbuild and there are many reasons to have it function this way in the browser). to make `require` always a known symbol in a commonjs file with eval.

perhaps in the parser (only when bundling) it can add a statement at the top like `var require = __require` so that the implementation it uses is correct, but i think that has a bunch more implications.

PR #13134, opened 2024-08-07
Make getIfPropertyExists binding safer

Jarred-Sumner:
### What does this PR do?

Make getIfPropertyExists binding safer

### DIFF
--- a/src/bun.js/bindings/bindings.cpp
+++ b/src/bun.js/bindings/bindings.cpp
@@ -3715,13 +3715,20 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsImpl(JSC__JSValue JSValue0,
 {
 
     JSValue value = JSC::JSValue::decode(JSValue0);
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit 6c490fba63104c1e72fe50cfeb2ac295e2962e63 is building: [`#1413`](https://buildkite.com/bun/bun/builds/1413)

src/bun.js/bindings/bindings.cpp
```diff
@@ -3715,13 +3715,20 @@ JSC__JSValue JSC__JSValue__getIfPropertyExistsImpl(JSC__JSValue JSValue0,
 {
 
     JSValue value = JSC::JSValue::decode(JSValue0);
-    if (UNLIKELY(!value.isObject()))
+    if (UNLIKELY(!value)) {
```
dylan-conway: I think this should be a debug assertion instead of an extra branch. We control usage of this function, it should never receive zero

PR #13044, opened 2024-08-03
Support async iterators in fs.promises.writeFile

Jarred-Sumner:
### What does this PR do?

This is an undocumented feature of `fs.promises.writeFile`. Source: https://github.com/nodejs/node/blob/2eff28fb7a93d3f672f80b582f664a7c701569fb/lib/internal/fs/promises.js#L481-L493

Fixes https://github.com/oven-sh/bun/issues/11633

### How did you verify your code works?

There are a couple tests

Issue #1434, opened 2022-11-01
Problem installing npm package croner


Hexagon:
### What version of Bun is running?

0.2.2

### What platform is your computer?

Linux 5.4.0-1094-azure #100~18.04.1-Ubuntu SMP Mon Oct 17 11:44:30 UTC 2022 x86_64 unknown

### What steps can reproduce the bug?

1.   Create an empty folder, enter it
2.   Install `croner` by `bun add croner`
     - An incomplete folder structure without files is created under node_modules/croner
4.   Create a script `test.js`
      ```javascript
      import { Cron } from "croner";
      Cron("* * * * * *", () => console.log("Every sec"));
5.    Run `bun run test.js`
      ```
      error: Could not resolve: "croner". Maybe you need to "bun install"?
      import { Cron } from "croner";
      ```
6.   Rerunning add/install does not help
7.   Running `npm install croner` using regular npm helps

### How often does it reproduce? Is there a required condition?

Every time, no

### What is the expected behavior?

bun add croner installes the npm package properly under node_modules

### What do you see instead?

An incomplete folder structure without files under node_modules/croner

### Additional information

Running in a github Codespace, so should be fairly simple to reproduce.

> Note: When passing `--backend copyfile` croner is successfully installed.

Hexagon:
Verbose output of `bun add croner --verbose`

Also tried `--verbose --no-cache` to no avail

```
@Hexagon ➜ /workspaces/laboratory/buncronertest (main ✗) $ bun add croner --verbose
bun add v0.2.2

 [31.00ms] [cache hit] croner
Enqueue package manifest for download: croner
  🔍 Resolving [1/1] Request: GET /croner
        If-None-Match: "35bf22990db294105b77789a90cb8af1"
        Accept: application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*
        Connection: keep-alive
        User-Agent: Bun/0.2.2
        Host: registry.npmjs.org
        Accept-Encoding: deflate, gzip

Request: GET /croner
        If-None-Match: "35bf22990db294105b77789a90cb8af1"
        Accept: application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*
        Connection: keep-alive
        User-Agent: Bun/0.2.2
        Host: registry.npmjs.org
        Accept-Encoding: deflate, gzip

Request: GET /croner
        If-None-Match: "35bf22990db294105b77789a90cb8af1"
[...]

PR #13136, opened 2024-08-07
windows: fix fs-promises-writeFile-async-iterator.test.ts

nektro:
fixes the mentioned test for windows. was failing when it was introduced in https://github.com/oven-sh/bun/pull/13044. manually tested on windows and macos that `fs-promises-writeFile-async-iterator.test.ts` and `fs.test.ts` still worked after this patch.

### DIFF
--- a/src/js/node/fs.promises.ts
+++ b/src/js/node/fs.promises.ts
@@ -586,22 +586,17 @@ async function writeFileAsyncIteratorInner(fd, iterable, encoding) {
   const mustRencode = !(encoding === "utf8" || encoding === "utf-8" || encoding === "binary" || encoding === "buffer");
   let totalBytesWritten = 0;
 
-  try {
-    for await (let chunk of iterable) {
-      if (mustRencode && typeof chunk === "string") {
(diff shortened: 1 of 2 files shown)


robobun:
@nektro, your commit 7fdcd8e262345ce29d314fd3befca9a5d784785a is building: [`#1434`](https://buildkite.com/bun/bun/builds/1434)

src/js/node/fs.promises.ts
```diff
@@ -586,22 +586,17 @@ async function writeFileAsyncIteratorInner(fd, iterable, encoding) {
...
-        totalBytesWritten += await prom;
-      } else {
-        totalBytesWritten += prom;
-      }
+    const prom = writer.write(chunk);
+    if (prom && $isPromise(prom)) {
+      await prom;
     }
-  } finally {
-    await writer.end();
```
Jarred-Sumner, line 604: `end` should always be called when the writer is done

this would be masking a bug that only occurs in `end`

Jarred-Sumner (changes requested):
A comment

PR #13144, opened 2024-08-07
Better MDN reference

huseeiin:
provide a more correct MDN reference to setting a request body using fetch()

### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
Thanks

Issue #13146, opened 2024-08-07
`fs.promises.filehandle.read()` throws when bytes to read equals 0


Araxeus:
### What version of Bun is running?

1.1.21

### What platform is your computer?

Windows NT 10.0.19045.0 x64

### What steps can reproduce the bug?

use `fs.promises.filehandle.read()` https://nodejs.org/api/fs.html#filehandlereadbuffer-offset-length-position
with 0 bytes to read (`length = 0`)

```js
const { open } = require('node:fs/promises');

open('anyfile.txt', 'r').then(filehandle => {
    filehandle.read(
        Buffer.alloc(1),
        0, // offset
        0, // length
    ).then(console.log).catch(console.error);
});
});
```

### What is the expected behavior?

NodeJS accepts a read operation with length 0 and prints:
```
$ node read.js
[Object: null prototype] { bytesRead: 0, buffer: <Buffer 00> }
```

### What do you see instead?

```
$ bun read.js
1 | const { open } = require('node:fs/promises');
2 | open('anyfile.txt', 'r').then(filehandle => {
3 |     filehandle.read(
                   ^
TypeError: length must be greater than 0
 code: "ERR_INVALID_ARG_TYPE"

      at node:fs/promises:263:50
      at read (node:fs/promises:263:15)
      at C:\Git\test\read.js:3:16
```

### Related Issue

https://github.com/Borewit/strtok3/pull/1141 - this is a dependency of [file-type](https://www.npmjs.com/package/file-type) on npm, which has 20m weekly downloads

### Additional information

Check that throw:
https://github.com/oven-sh/bun/blob/9f7c6e34cb9eab2b634fff284088b27db36b6857/src/bun.js/node/node_fs.zig#L2916-L2919

Function which should be called after the check: (I don't know if it will actually handle `args.length=0`)
https://github.com/oven-sh/bun/blob/9f7c6e34cb9eab2b634fff284088b27db36b6857/src/bun.js/node/node_fs.zig#L4888-L4904

## Possible fix:

https://github.com/oven-sh/bun/blob/9f7c6e34cb9eab2b634fff284088b27db36b6857/src/js/node/fs.promises.ts#L364-L374

rewrite to something equivalent to
```patch
    async read(buffer, offset, length, position) {
+      if (length === 0) {
+        return { buffer, bytesRead: 0 };
+      }
      const fd = this[kFd];
      throwEBADFIfNecessary(read, fd);

      try {
        this[kRef]();
[...]

PR #13149, opened 2024-08-07
fix node/fs.promises.FileHandle.read() not behaving like NodeJS

Araxeus:
### What does this PR do?

fix `fs.promises.filehandle.read()` throws when bytes to read (`args.length`) equals 0 (fix #13146)

[NodeJS doesn't check if args.length equals 0](https://github.com/nodejs/node/blob/a4f609faf5a7fd42a3886e1273fe2af6ad5cb84d/src/node_file.cc#L2595-L2597) and handles it gracefully

I have chosen to fix it on the js side because I think calling the bindings is redundant in this case - and also I'm not sure if there is a reason for the Zig check since [`Syscall.read`](https://github.com/oven-sh/bun/blob/9f7c6e34cb9eab2b634fff284088b27db36b6857/src/bun.js/node/node_fs.zig#L4894C24-L4894C36) should handle 0 length reads ([Windows _read](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/read) and [Linux read(2)](https://man7.org/linux/man-pages/man2/read.2.html) both are ok with it?)

- [x] Code changes

> [!NOTE]  
> This check is kinda redundant now, but i guess can be kept just to be sure?
> https://github.com/oven-sh/bun/blob/9f7c6e34cb9eab2b634fff284088b27db36b6857/src/bun.js/node/node_fs.zig#L2916-L2919

### DIFF
--- a/src/js/node/fs.promises.ts
+++ b/src/js/node/fs.promises.ts
@@ -362,6 +362,9 @@ export default exports;
     }
 
     async read(buffer, offset, length, position) {
+      if (length === 0) {
+        return { buffer, bytesRead: 0 };
+      }
       const fd = this[kFd];
       throwEBADFIfNecessary(read, fd);
 
--- a/test/js/node/fs/fs.test.ts
+++ b/test/js/node/fs/fs.test.ts
@@ -124,6 +124,12 @@ describe("FileHandle", () => {
     expect(await fd.read(buf, 0, 10, 0)).toEqual({ bytesRead: 10, buffer: buf });
   });
 
+  if("FileHandle#read handles 0 length reads", async () => {
+    await using fd = await fs.promises.open(__filename);
+    const buf = Buffer.alloc(1);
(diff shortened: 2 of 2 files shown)


src/js/node/fs.promises.ts
```diff
@@ -362,6 +362,9 @@ export default exports;
     }
 
     async read(buffer, offset, length, position) {
+      if (length === 0) {
```
Jarred-Sumner, line 365: This is semantically different

This does not do a read() system call

If the file descriptor is invalid, a 0-length read would be EBADF, or if something else was wrong with the file, it would error rather than return 0

Araxeus: True, for best results this fix should be in the zig code, but I dont feel comfortable making changes in the zig codebase, and like I said I don't know why this check exists in the first place

Preferably, someone could make another PR that would close the issue 🙏

PR #13115, opened 2024-08-06
Add `TextDecoderStream`, `TextEncoderStream`, and `TextDecoder.decode("", { stream: true})`

dylan-conway:
### What does this PR do?
Adds `TextDecoderStream` and `TextEncoderStream` to globals. Part of implementing this required the `stream: true` option for `TextDecoder`.

Also adds support for `new TextDecoder("utf-16be")`.

closes #5648 
closes #5542 

### How did you verify your code works?
A bunch of tests from webkit and node

PR #1479, opened 2022-11-10
fix(child_process): fix execFileSync options.input

ThatOneBro:

PR #13151, opened 2024-08-07
Revert `TextDecoderStream`

Electroid:
We're temporarily reverting this commit so we can get the Bun v1.1.22 release out. We found out there was a crash with `TextDecoderStream` and Next.js.

Reverts oven-sh/bun#13115


robobun:
@Electroid, your commit fa56f03d0bdf28a4f97b94af28e01257c89b2d5d is building: [`#1479`](https://buildkite.com/bun/bun/builds/1479)

PR #1499, opened 2022-11-13
Add linker script to remove unwanted exports

froody:
fixes crashes when handling c++ exceptions in a shared library loaded by bunffi

PR #13154, opened 2024-08-07
test: run `next build` with node and bun separately

nektro:


robobun:
@nektro, your commit a324fff16c9739d7605d7f81b2f31a215a9124a4 is building: [`#1499`](https://buildkite.com/bun/bun/builds/1499)

nektro:
there's still things to look into but the ci runner's root disk being a network drive and our handling of that is likely the root cause of that mixed with that the two commands were being run in parallel. this now only still failed on baseline.

nektro:
this is worth looking into as we increase our nodejs test coverage but its not load-bearing on the test passing; it must get most of its work done before the crash is raised

```
Compiler server unexpectedly exited with code: null and signal: SIGTERM
Compiler edge-server unexpectedly exited with code: null and signal: SIGTERM
Compiler client unexpectedly exited with code: null and signal: SIGTERM
```

test/integration/next-pages/test/next-build.test.ts
```diff
@@ -105,41 +105,27 @@ test(
...
-      stdio: ["ignore", "pipe", "inherit"],
+      stdio: ["ignore", "inherit", "inherit"],
       env: {
         ...bunEnv,
         NODE_ENV: "production",
       },
     });
+    await bunBuild.exited;
+    console.timeEnd("[bun] next build");
+    expect(bunBuild.exitCode).toBe(0);
```
Jarred-Sumner, line 118: Ideally, we'd assert on the `exited` value because exitCode will return null if it segfaulted

Jarred-Sumner (changes requested):
next-build test is still failing

how does this make it better?

test/integration/next-pages/test/next-build.test.ts
nektro: if it was `null` because of a segfault that would still be caught by `.toBe(0)`

PR #13158, opened 2024-08-08
ci: set duration when subprocess.on(error)

nektro:
this should make it so test files that timeout or fail dont show up in the logs as `0s`


robobun:
@nektro, your commit edc8cf26c0aa042e50e0a1236226520197714b42 is building: [`#1510`](https://buildkite.com/bun/bun/builds/1510)

nektro:
this didn't do it, need to give it a closer look again later

Issue #13152, opened 2024-08-07
windows: FileSink#end closes even when backed by a file descriptor


nektro:
### What version of Bun is running?

1.1.22-debug+d74a19234

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

```js
import fs from "node:fs";
import util from "node:util";
const fd = await util.promisify(fs.open)("test.txt", "w");
const chunk = Buffer.from("1 Hello, world!");
const writer = Bun.file(fd).writer();
const written = await writer.write(chunk);
await writer.end();
await util.promisify(fs.ftruncate)(fd, written);
await util.promisify(fs.close)(fd);
```

### What is the expected behavior?

no error. `test.txt` is created with the 15 byte content

### What do you see instead?

```
EBADF: Bad file descriptor
   errno: -9
 syscall: "ftruncate"
      fd: 4

Bun v1.1.22-debug (Windows x64)
```

### Additional information

n.a

nektro:
this is the root cause of `fs-promises-writeFile-async-iterator.test.ts` currently failing on windows CI

Issue #1518, opened 2022-11-17
TypeError: undefined is not an object (evaluating 'process.stderr.fd')


fernandolguevara:
### What version of Bun is running?

0.2.2

### What platform is your computer?

macos m1 max arm64

### What steps can reproduce the bug?

```sh
mkdir buntest
cd buntest
echo -e "console.log(process.stderr.fd)" > index.ts
bun index.ts
```

### How often does it reproduce? Is there a required condition?

always

### What do you see instead?

1 | console.log(process.stderr.fd);
               ^
TypeError: undefined is not an object (evaluating 'process.stderr.fd')

Electroid:
We're working on this right now! Tracking the issue in #894, and should be fixed when #1495 is merged.

subhanahmed047:
Came across the same issue. Looking forward to the fix. 

```
bun index.ts 
153 |  */
154 | 
155 | function useColors() {
156 |   return 'colors' in exports.inspectOpts ?
157 |           Boolean(exports.inspectOpts.colors) :
158 |           tty.isatty(process.stderr.fd);
                ^
TypeError: undefined is not an object (evaluating 'process.stderr.fd')
```

For context, I get this when I try to require `sequalize` in an express app.

PR #13164, opened 2024-08-08
windows: fix fs-promises-writeFile-async-iterator.test.ts [v2]

nektro:
Closes https://github.com/oven-sh/bun/issues/13152

adds independent tests too


robobun:
@nektro, your commit 2e73560a0a9f12b505a8932c20fbba12e140061c is building: [`#1518`](https://buildkite.com/bun/bun/builds/1518)

src/io/PipeWriter.zig
```diff
@@ -1003,7 +1003,12 @@ pub fn WindowsBufferedWriter(
             this.is_done = true;
             if (this.pending_payload_size == 0) {
                 // will auto close when pending stuff get written
-                this.close();
+                if (this.source) |source| {
+                    switch (source) {
+                        .sync_file, .file => {},
```
Jarred-Sumner: this is confusing to me

Jarred-Sumner: can you add a comment explaining why it should only call close for pipes and ttys?

src/io/PipeWriter.zig
```diff
@@ -1382,7 +1387,12 @@ pub fn WindowsStreamingWriter(
             this.closed_without_reporting = false;
             // if we are done we can call close if not we wait all the data to be flushed
             if (this.isDone()) {
-                this.close();
+                if (this.source) |source| {
+                    switch (source) {
+                        .sync_file, .file => {},
```
Jarred-Sumner: Please add a comment explaining this

test/js/bun/util/filesink.test.ts
```diff
@@ -146,3 +146,31 @@ describe("FileSink", () => {
     });
   }
 });
+
+import fs from "node:fs";
+import util from "node:util";
+import path from "node:path";
+
+it("end doesn't close when backed by a file descriptor", async () => {
```
Jarred-Sumner, line 173: Where does it check the opposite - that `Bun.file(path).writer().end()` does close the file descriptor?

nektro: this fails on posix as well. should it pass?
```ts
it("end does close when not backed by a file descriptor", async () => {
  const x = tmpdirSync();
  const file = Bun.file(path.join(x, "test.txt"));
  const writer = file.writer();
  await writer.write(Buffer.from("1 Hello, world!"));
  await writer.end();
  expect(async () => await writer.write(Buffer.from("2 Hello, world!"))).toThrow();
});
```

nektro: okay, using `getMaxFD()` instead and `expect(fd_after).toBe(fd_before);` does pass on posix

nektro: and does not on windows, i'll use that check instead

src/io/PipeWriter.zig
```diff
@@ -891,6 +891,7 @@ pub fn WindowsBufferedWriter(
 ) type {
     return struct {
         source: ?Source = null,
+        is_fd: ?bool = null,
```
Jarred-Sumner: this variable name doesn't say what it is used for

it's also an optional bool which takes up extra space

instead, couldn't this be an enum key on `Source`? like `owned_file` to say "a file which should be closed"

PR #13176, opened 2024-08-08
fix(sockets) add socket wrapper and refactor context ownership handling in socket.zig

cirospaciari:
### What does this PR do?

- [X] Code changes

Existing tests


src/bun.js/api/bun/socket.zig
```diff
@@ -1354,35 +1354,36 @@ fn NewSocket(comptime ssl: bool) type {
...
-            this.socket = socket;
+            // This socket must be replaced because the previous one is a connecting socket not a uSockets socket
+            this.socket = .{ .socket = socket };
             JSC.markBinding(@src());
             log("onOpen ssl: {}", .{comptime ssl});
 
             // Add SNI support for TLS (mongodb and others requires this)
             if (comptime ssl) {
-                var ssl_ptr = this.socket.ssl();
+                var ssl_ptr = this.socket.ssl() orelse unreachable;
```
Jarred-Sumner: this is not accurate right? don't we set ->ssl = nullptr somewhere?

src/deps/uws.zig
```diff
@@ -1040,7 +1040,215 @@ pub fn NewSocketHandler(comptime is_ssl: bool) type {
         }
     };
 }
+pub fn NewWrappedSocketHandler(comptime is_ssl: bool) type {
```
Jarred-Sumner: Maybe instead of adding another wrapper here, we could do this on the inner NewSocketHandler struct?

Jarred-Sumner (changes requested):
2 comments but almost mergeable

src/bun.js/api/bun/socket.zig
cirospaciari: If is detached/closed, inside onOpen is unreachable but I can check it just in case

src/deps/uws.zig
cirospaciari: Yeah but we would possible touch more places than just socket.zig so I was just cautious, I can refactor so everyone using NewSocketHandler also uses the detachable socket, I think http.zig and probably WebSocket client also uses it

Issue #1579, opened 2022-12-04
throwIfNoEntry isn't acknowledged in fs.statSync


miklosme:
### What version of Bun is running?

0.2.2

### What platform is your computer?

Darwin 21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:28 PDT 2022; root:xnu-8020.140.41~1/RELEASE_ARM64_T8110 arm64 arm

### What steps can reproduce the bug?

```js
import fs from 'fs';

fs.statSync('file_that_does_not_exist', {
  throwIfNoEntry: false,
});

console.log('done');
```

The above code reaches the console.log in Node, but throws in Bun.

Jarred-Sumner:
This has been fixed in Bun v0.6.14

PR #13182, opened 2024-08-09
`throw("Out of memory")` -> throwOutOfMemory()

Jarred-Sumner:
### What does this PR do?

- Use JSC::throwOutOfMemory instead of an ErrorInstance with the message "Out of memory"
- Use bun.outOfMemory() instead of a slightly different panic string for the same thing


robobun:
@Jarred-Sumner, your commit 1c664251f1a25fac5e46bd972bff72c97781307c is building: [`#1579`](https://buildkite.com/bun/bun/builds/1579)

dylan-conway (approved):

nektro (approved):

PR #1606, opened 2022-12-11
fix(stream): Fix Readable.pipe()

ThatOneBro:
It seems .pipe has been broken due to some debug logs referencing variables that don't exist, even when no-op'd. Obviously it's not ideal that debug statements are included in the output, and we should probably address that either via:
1. Commenting them out before commiting
2. Stripping them via transpiler before embedding the polyfills (probably should do this with comments too tbh)

But this PR does fix .pipe in the meantime and adds a test to make sure it's working.

PR #13189, opened 2024-08-09
js: more process exit code fixes

nektro:
add 3 new tests that were previously failing


robobun:
@nektro, your commit 2ee860f5e236291769b8f7feef72468f4b1128bf is building: [`#1606`](https://buildkite.com/bun/bun/builds/1606)

src/bun.js/node/types.zig
```diff
@@ -2117,6 +2118,7 @@ pub const Process = struct {
 
         vm.exit_handler.exit_code = code;
         vm.onExit();
+        if (bun_js.anyUnhandled() and vm.exit_handler.exit_code == 0) vm.exit_handler.exit_code = 1;
```
Jarred-Sumner: What happens if you run `process.exit()` in `bun test` or in macros? It's uninitialized memory

src/bun_js.zig
```diff
@@ -469,6 +469,10 @@ pub const Run = struct {
     }
 };
 
+pub fn anyUnhandled() bool {
```
Jarred-Sumner: `src/bun_js.zig` shouldn't be used outside of `bun run <file>` or `bun <file>`

There can be many other unhandled exception trackers.

Jarred-Sumner (changes requested):
comments

src/bun.js/node/types.zig
nektro: do we allow that?

Jarred-Sumner: we don't prevent it

src/bun.js/node/types.zig
```diff
@@ -23,6 +23,7 @@ const Shimmer = @import("../bindings/shimmer.zig").Shimmer;
 const Syscall = bun.sys;
 const URL = @import("../../url.zig").URL;
 const Value = std.json.Value;
+const bun_js = @import("../../bun_js.zig");
```
```suggestion

```

src/bun.js/node/types.zig
```diff
@@ -2107,16 +2107,18 @@ pub const Process = struct {
...
+        Bun__setExitCode(vm, code, explicit);
+
         if (vm.worker) |worker| {
-            vm.exit_handler.exit_code = code;
             worker.requestTerminate();
             return;
         }
 
-        vm.exit_handler.exit_code = code;
         vm.onExit();
```
Jarred-Sumner, line 2121: Instead of relying on vm.unhandled_error_counter,  let's create an unhandled rejection scope, or just track the count of unhandled errors
```
const scope = vm.unhandledRejectionScope();

Issue #1609, opened 2022-12-13
Update version when a version tag is specified in `bun install`


colinhacks:
### What version of Bun is running?

0.3.0

### What platform is your computer?

Darwin 22.1.0 arm64 arm

### What steps can reproduce the bug?

When the a version tag is specified in `bun install` Bun should do a network call to pull down the latest version with that tag.

1. `bun install pkg@latest` - installs version `x`
2. `pkg` publishes version `x + 1`
3. `bun install pkg@latest` re-installs `x` - no network call

npm allows arbitrary version tag names, though `latest` and `next` are common. If a version tag is specified in the install command Bun should always re-check the registry

colinhacks:
Follow up: even `-f` doesn't seem to refresh the registry cache.

Shortly after publishing a new version of `zod@canary`:

```bash
$ npm add zod@canary                          
up to date, audited 3 packages in 165ms
1 package is looking for funding
  run `npm fund` for details
found 0 vulnerabilities
$ (cd node_modules/zod && npm pkg get version)
"3.22.0-canary.20230305T082229"

$ yarn add zod@canary
yarn add v1.22.19
warning package.json: No license field
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
warning No license field
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...
success Saved lockfile.
warning No license field
success Saved 1 new dependency.
info Direct dependencies
└─ zod@3.22.0-canary.20230305T082229
info All dependencies
└─ zod@3.22.0-canary.20230305T082229
✨  Done in 0.14s.
$ (cd node_modules/zod && npm pkg get version)
"3.22.0-canary.20230305T082229"

$ bun add zod@canary                      
bun add v0.5.8 (c0a49e40)
  installed zod@1.10.2-canary
  1 packages installed [10.00ms]
$ (cd node_modules/zod && npm pkg get version)
"1.10.2-canary"

$ bun add -f zod@canary
bun add v0.5.8 (c0a49e40)
  + typescript@4.9.5
  installed zod@1.10.2-canary
  2 packages installed [30.00ms]
$ (cd node_modules/zod && npm pkg get version)
"1.10.2-canary"
```

[...]

PR #13190, opened 2024-08-09
Set -ffile-prefix-map

Jarred-Sumner:
### What does this PR do?

More reproducible builds + easier debugging in some cases


robobun:
@Jarred-Sumner, your commit 5a77687a8eb40c76dfb8369ac1d46a4985d3324d is building: [`#1609`](https://buildkite.com/bun/bun/builds/1609)

Issue #13184, opened 2024-08-09
Bun does not exit if `net.Socket` end (FIN) from the other side


steelywing:
### What version of Bun is running?

1.1.22

### What platform is your computer?

Microsoft Windows NT 10.0.19045.0 x64

## Description

- Client `net.connect()` to server
- Server `end()` (send FIN) the socket
- Client after the recieve FIN, still hold the connection, make bun does not exit

## Reproduce

- Run `server.js`
- Run `client.js`

```js
// server.js

const net = require('node:net');
const server = net.createServer((c) => {
    console.log('client connected');
    c.on('end', () => {
        console.log('client disconnected');
    });
    c.on('data', (data) => {
        console.log(`Received: ${data}`);
    });
    c.write('Hello client');
    setTimeout(() => {
        c.end();
    }, 1000);
});
server.on('error', (err) => {
    throw err;
});
server.listen(1337, () => {
    console.log('server bound');
});
```

```js
// client.js

var net = require('net');

var client = new net.Socket();
client.connect(1337, '127.0.0.1', function() {
	console.log('Connected');
	client.write('Hello, server');
});

client.on('data', function(data) {
	console.log('Received: ' + data);
});

client.on('end', function() {
	console.log('Connection end');
});

client.on('close', function() {
	console.log('Connection closed');
});
```

### What is the expected behavior?

Client process should exit, the process exit normally in node.js

### What do you see instead?

Client does not exit

### Additional information

Relate: #11414 https://github.com/sidorares/node-mysql2/issues/2884

Issue #11414, opened 2024-05-28
Bun does not end after closing mysql2 pool


Dovonun:
### What version of Bun is running?

1.1.10+5102a9443

### What platform is your computer?

Linux 6.8.0-31-generic x86_64 x86_64

### What steps can reproduce the bug?

This is the simplest example of where I saw this behavior.
Without using the pool(no query), Bun exits as expected.
```javascript
import mysql from "mysql2/promise";

const pool = mysql.createPool({
  uri: process.env.DATABASE_URL,
  database: 'tower',
  waitForConnections: true,
  connectionLimit: 10,
  maxIdle: 10,
  idleTimeout: 60000,
  queueLimit: 0,
  enableKeepAlive: true,
  keepAliveInitialDelay: 0,
});

const test = await pool.query('select 1 from dual where 1=1');
console.log(test);

pool.end();
```

### What is the expected behavior?

Exit after everything is executed.

### What do you see instead?

Bun does not exit.
![image](https://github.com/oven-sh/bun/assets/31408384/08059451-7bcb-47c4-b484-fcfc67c4f619)

wellwelwel:
If you think there's something that could be done from **MySQL2** side, please ping me.

Currently, all **MySQL2** test suite runs _against_ **Node.js** and **Deno**, but due to this reported behavior, we only perform a basic `SELECT` test with **Bun** using a workaround to force the process exit:

```js
connection.end((err) => { 
  assert.ifError(err); 
  process.exit(0); 
}); 
```

---

In addition, 100% of the test suite passes with **Bun** if we ignore this problem.

Tanavya:
Might be related to this issue: https://github.com/oven-sh/bun/issues/12155

PR #13212, opened 2024-08-09
fix(net/tls) we need to call end when we got FIN if allowHalfOpen is false

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/13184
Fix: https://github.com/oven-sh/bun/issues/11414

- [X] Code changes

Added a test


robobun:
@cirospaciari, your commit ad221745515a32fe707485936b45756abf321c6a is building: [`#1651`](https://buildkite.com/bun/bun/builds/1651)

src/js/node/net.ts
```diff
@@ -187,17 +187,35 @@ const Socket = (function (InternalSocket) {
...
     static #Close(socket) {
       const self = socket.data;
       if (!self || self.#closed) return;
       self.#closed = true;
       //socket cannot be used after close
       self[bunSocketInternal] = null;
-      const queue = self.#readQueue;
-      if (queue.isEmpty()) {
-        if (self.push(null)) return;
+      if (self.#final_callback) {
```
```suggestion
     const finalCallback = self.#final_callback;
      if (finalCallback) {
        self.#final_callback = null;
        finalCallback()
```

test/js/node/net/node-net.test.ts
```diff
@@ -490,3 +490,36 @@ it("socket should keep process alive if unref is not called", async () => {
   });
   expect(await process.exited).toBe(1);
 });
+
+it.only("should not hang after FIN", async () => {
```
Jarred-Sumner:
```suggestion
it("should not hang after FIN", async () => {
```

wellwelwel:
Hi @cirospaciari, thanks for this fix 🤝

I have this simple helper:

```ts
import { createConnection } from 'net';

const checkPort = (port: number, host: string): Promise<boolean> =>
  new Promise((resolve) => {
    const client = createConnection(port, host);

    client.on('connect', () => {
      client.destroy(); // it worked normally before this PR
      resolve(true);
    });

    client.on('error', () => {
      resolve(false);
    });
  });
```

---

After this _PR_, the process doesn't finish after a successful port check.

- If I change it to `client.end()` instead of `client.destroy()`, the behavior works as expected.

I also checked the regression changes _(aka `canary`)_, but they don't seem to "visit" the `destroy` method.

---

Do you think it would be interesting to open an issue for this? 🙋🏻‍♂️

cirospaciari:
> ```ts
> import { createConnection } from 'net';
> 
> const checkPort = (port: number, host: string): Promise<boolean> =>
>   new Promise((resolve) => {
>     const client = createConnection(port, host);
> 
>     client.on('connect', () => {
>       client.destroy(); // it worked normally before this PR
>       resolve(true);
>     });
> 
>     client.on('error', () => {
>       resolve(false);
>     });
>   });
> ```

Thanks for reporting and giving the reproduction, I just opened a PR with a fix for that.

Issue #9609, opened 2024-03-25
Add /json Endpoint to Bun Inspector Server to Support External Debugging Tools


june07:
### What is the problem this feature would solve?

Currently debugging with Chromium DevTools is possible but a manual copy paste of the debug socket is needed as you can see from the first screenshot where I copy/pasted "**localhost:4000/7u1bqmpvhg**" into DevTools. https://github.com/june07/nim is a tool I wrote for Node.js (works for Deno as well) some years back for this and it would be nice to use it for Bun as well, to streamline things.

![image](https://github.com/oven-sh/bun/assets/11353590/26fdfd40-c10f-4c1d-bc98-152b67ec6bbd)

Node/Deno both implement inspect servers that have a json endpoint which contains data about the debug websocket:

For example, http://localhost:9229/json will display this output which allows local clients to determine the debugger web socket details so as to connect: `ws=localhost:9229/abcdbdb5-f00c-4554-88e5-d740ba881234"` 

```
[
    {
        "description": "node.js instance",
        "devtoolsFrontendUrl": "devtools://devtools/bundled/js_app.html?experiments=true&v8only=true&ws=localhost:9229/abcdbdb5-f00c-4554-88e5-d740ba881234",
        "devtoolsFrontendUrlCompat": "devtools://devtools/bundled/inspector.html?experiments=true&v8only=true&ws=localhost:9229/abcdbdb5-f00c-4554-88e5-d740ba881234",
        "faviconUrl": "https://nodejs.org/static/images/favicons/favicon.ico",
        "id": "565ebdb5-f00c-4554-88e5-d740ba88bf28",
        "title": "src/index.js",
        "type": "node",
        "url": "file:///usr/src/node-app/src/index.js",
        "webSocketDebuggerUrl": "ws://localhost:9247/565ebdb5-f00c-4554-88e5-d740ba88bf28"
    }
]
```

I'm sure those reading this are familar with this in Node... but in case not: https://nodejs.org/en/learn/getting-started/debugging
![image](https://github.com/oven-sh/bun/assets/11353590/aea37219-43e0-4d70-ba6d-58cb98be4362)

### What is the feature you are proposing to solve the problem?

Implement the same inspector sever /json endpoint

june07:
Ping

zastlx:
+1

june07:
I was referring to the devtools browser address per the original post...

[...]

PR #13219, opened 2024-08-10
feature: add debugger info endpoint (/json) for external tool discovery...

june07:
including browser DevTools similar to Node.js

### What does this PR do?

Fixes this issue https://github.com/oven-sh/bun/issues/9609 by adding a /json endpoint which lists the debugger metadata similar to Node.js

image

This is useful to allow Bun to be debugged with browser DevTools (i.e. Chrome, Edge, etc) in addition to the Bun web debugger.

### DIFF
--- a/src/js/internal/debugger.ts
+++ b/src/js/internal/debugger.ts
@@ -1,5 +1,4 @@
 import type { Server as WebSocketServer, WebSocketHandler, ServerWebSocket, SocketHandler, Socket } from "bun";
-
 export default function (
   executionContextId: string,
   url: string,
@@ -129,7 +128,7 @@ class Debugger {
         return Response.json(versionInfo());
       case "/json":
       case "/json/list":
-      // TODO?
+        return Response.json(debuggerInfo(this));
     }
 
     if (!this.#url.protocol.includes("unix") && this.#url.pathname !== pathname) {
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you for this

We tried doing this early on, but unfortunately the V8 inspector protocol and the WebKit debugger protocol has diverged quite a bit and much of the debugger when used this way doesn't work. It needs a bunch of code to translate between the two protocols.

june07:
Ah! My fault I guess for not digging deep enough... I didn't know that. Thanks for the fast reply! And best of luck with the project moving forward. Cheers!

june07:
After reflecting on this, I realize I may have closed the discussion prematurely. The JSON endpoint would still be valuable for tools like [NiM](https://github.com/june07/nimv3) and potentially others that don’t have direct access to the execution context (e.g., Bun’s CLI entry-point for parsing data).

Currently, NiM supports Deno and Node.js, and adding Bun to this list would be beneficial since all these runtimes are based on the V8 engine. Given that Bun’s debugging isn’t integrated into the browser like Node.js with Chrome DevTools, Bun developers could greatly benefit from NiM’s features for debug session management and more.

Here’s a demonstration of debugging across all three runtimes, including NiM auto-connecting to the Bun debug process:

v8Trifecta

I’ve updated [this commit](https://github.com/oven-sh/bun/commit/697330e2c207d5d3b4130717b9cb3a366dda6a24) to set the `devtoolsFrontendUrl` to Bun’s WebKit URL and removed the Chrome DevTools entries.

I’d love to get your feedback.

PR #13220, opened 2024-08-10
console: implement cutoff for large arrays

nektro:
added new test coverage as well


robobun:
@nektro, your commit 27337245c7a0c0c25da3ce1ed519f95a991d8290 is building: [`#1680`](https://buildkite.com/bun/bun/builds/1680)

src/bun.js/ConsoleObject.zig
```diff
@@ -2195,6 +2197,13 @@ pub const Formatter = struct {
                             }
                             continue;
                         }
+                        if (j >= 100) {
+                            writer.writeAll(",\n");
```
Jarred-Sumner: this should be goodTimeForANewLine() and use printComma

Jarred-Sumner: i think

src/bun.js/ConsoleObject.zig
```diff
@@ -2186,6 +2187,7 @@ pub const Formatter = struct {
                     }
 
                     var i: u32 = 1;
+                    var j: u32 = 1;
```
Jarred-Sumner:
```suggestion
                    var non_holey_array_items_count: u32 = 1;
```

Issue #1704, opened 2023-01-03
 Cannot find package "tls"


wmelton:
### What version of Bun is running?

0.4.0

### What platform is your computer?

Darwin 21.3.0 x86_64 i386

### What steps can reproduce the bug?

Trying out bun + [nats](https://github.com/nats-io/nats.js) - looks like there could be an issue with resolving node-native packages when `require` is used in a dependency, or a more generic dependency resolution error. 

To reproduce: 

```bash
docker pull nats

docker run -d --name nats-main -p 4222:4222 -p 6222:6222 -p 8222:8222 nats
```

Create a file called `publish.js` (for example) and set its contents thusly: 

```js
import { connect, StringCodec } from "nats";

(async () => {
// to create a connection to a nats-server:
const nc = await connect({ servers: "localhost:4222" });

// create a codec
const sc = StringCodec();


nc.publish("hello", sc.encode("world"));
nc.publish("hello", sc.encode("again"));
})();
```

Then run: 

```bash
bun publish.js
```

### What is the expected behavior?

Script file would run same as `node publish.js` does

### What do you see instead?

`Cannot find package "tls" from ".../node_modules/nats/lib/src/node_transport.js"`

### Additional information

If you run `bun install tls` the error is resolved, but then it can't find the `dns` package used in nats. Given that nats package.json file doesn't include `tls` or `dns` as a defined dependency, my gut says this is an issue related to bun not knowing these are native node modules that need to be pulled in that a node developer would likely never name explicitly in a project's package.json file.

That said, if you run `bun install <dep>` to fix it, the resolve logic then throws a new error that you can't get past:

```bash
44 |   function workaround_require(path) {
45 |     if(/^\.\//.test(path) || /^\.\.\//.test(path))
46 |       path = path_lib.resolve(mod_dir, path);
47 | 
48 |     try       {
49 |       return real_require(path)
                 ^
TypeError: undefined is not an object
```

Jarred-Sumner:
`tls` isn't implemented yet

danscan:
I encountered the same when using `nats` on Bun@0.40.0, and after running on canary (at 7fa023b), I now run into #1794.

Jarred-Sumner:
@danscan can you give it another try?

danscan:
Worked like a charm on canary! Incredible support and response time – all around.

During the time I spent looking into workarounds and considering temporarily using Node to work around this, it was resolved. It's hard to state how awesome that is!

[...]

PR #13231, opened 2024-08-11
faster `TextEncoderStream` part 2

dylan-conway:
### What does this PR do?
- Excludes pending surrogates from simdutf conversion, avoiding errors with valid input
- Does not get the length. First over allocates, then resizes

Existing tests

### DIFF
--- a/src/bun.js/rare_data.zig
+++ b/src/bun.js/rare_data.zig
@@ -49,12 +49,23 @@ listening_sockets_for_watch_mode_lock: bun.Lock = .{},
 
 temp_pipe_read_buffer: ?*PipeReadBuffer = null,
 
+text_encoder_stream_buffer: ?*TextEncoderStreamBuffer = undefined,
+
+pub const TextEncoderStreamBuffer = [256 * 1024]u8;
+
 const PipeReadBuffer = [256 * 1024]u8;
 
 pub fn pipeReadBuffer(this: *RareData) *PipeReadBuffer {
     return this.temp_pipe_read_buffer orelse {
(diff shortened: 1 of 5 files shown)


robobun:
@dylan-conway, your commit a2adeb188875369b35851c0a0835b5a2d7a25e55 is building: [`#1704`](https://buildkite.com/bun/bun/builds/1704)

src/bun.js/webcore/encoding.zig
```diff
@@ -549,46 +550,96 @@ pub const TextEncoderStreamEncoder = struct {
...
-        ) catch {
+        const buf = bun.default_allocator.alloc(u8, max_length) catch {
             globalObject.throwOutOfMemory();
             return .zero;
         };
 
-        if (prepend) |*pre| {
-            buf.appendSliceAssumeCapacity(pre.bytes[0..pre.len]);
-        }
+        const count = bun.simdutf.convert.utf16.to.utf8.le(remain, buf);
```
Jarred-Sumner, line 600: I don't think we can use the non-validating version of this function

dylan-conway: Updated to use validating version

src/bun.js/rare_data.zig
```diff
@@ -58,6 +62,13 @@ pub fn pipeReadBuffer(this: *RareData) *PipeReadBuffer {
     };
 }
 
+pub fn textEncoderStreamBuffer(this: *RareData) *PipeReadBuffer {
```
Jarred-Sumner, line 65:
```suggestion
pub fn textEncoderStreamBuffer(this: *RareData) *TextEncoderStreamBuffer {
```

Issue #13088, opened 2024-08-05
Memory leak with NestJS and NestJS-graphql


rskaar:
### What version of Bun is running?

1.1.21

### What platform is your computer?

Darwin 23.5.0 arm64 arm / Docker with base image oven/bun:1.1.21

### What steps can reproduce the bug?

Minimal reproduction repo, with code and documented steps to reproduce:
https://github.com/rskaar/bun-nestjs-graphql-mem-issues

### What is the expected behavior?

The problem occurs when the NestJS app receives requests to any endpoint, as long as the NestJS project has the NestJS-graphQL module imported. For each request, the memory consumption goes up. After 100-300.000 requests, we're at over 1 gb of memory consumption from an initial <100mb.

The same app/code running with NodeJS does not have the same memory leak.

Removing (commenting out) the imported NestJS-GraphQL module gives no memory issues while running.

Note that the requests leading to the memory issues does not need to hit the GraphQL-endpoint. Any traffic to any endpoint in the NestJS app seems to generate a memory leak. I suspect that the NestJS-GraphQL (or ApolloServer) module adds some kind of middleware that each request goes through.

### What do you see instead?

Memory usage goes up for each request and is not garbage collected.

### Additional information

We use Bun in production, and this is a major issue for us - leading to our docker crashing multiple times per day due to memory usage.

Jarred-Sumner:
If these are incoming requests with a request body, can you run `bun upgrade --canary` and see if it’s better?

rskaar:
Tried upgrading to canary, but the issues persist. The requests used for testing are just GET http://locahost:3000/ without any body, special headers or similar.

billywhizz:
i've taken an initial look at this and can repro on ubuntu 22.04 following the excellent guide from @rskaar! :pray:

i can see JSC heap usage growing consistently and not shrinking despite explicit calls to flush any garbage. i will try to figure out what is causing it. 

![image](https://github.com/user-attachments/assets/d4925678-e1a8-499e-8af9-067a5757cdeb)

this is code i am using to capture the heap stats.

```TypeScript
import { NestFactory } from '@nestjs/core';
import { AppModule } from './app.module';
import {
  gcAndSweep,
  heapSize,
  heapStats,
} from "bun:jsc";

const AD = '\u001b[0m' // ANSI Default
const AG = '\u001b[32m' // ANSI Green
const AY = '\u001b[33m' // ANSI Yellow

function to_size_string (bytes) {
[...]

Issue #1722, opened 2023-01-04
bun: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by bun)


Electroid:
This is a regression from 0.4 to canary, we likely added a new symbol.

```
bun: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by bun)
```

PR #13254, opened 2024-08-12
fix(node:http): use a unique FakeSocket per socketId. only fire conne…

billywhizz:
### What does this PR do?

- Partially Fixes https://github.com/oven-sh/bun/issues/13088

### Notes/Investigation

https://gist.github.com/billywhizz/ff5382889152d8f9d357a6de0c10fae1

### Description

- This PR, instead of creating a ```FakeSocket``` for every request, will surface the socketIdentifier to node:http module which will now only create a new ```FakeSocket``` for each socketIdentifier that doesn't exist in it's ```Map``` of active sockets. 
- The overhead is minimal for a local release build against current Bun release - ~1.5% on macos. this overhead may go away with a CI build?

### Questions/Issues

- we seem to have always created a new ```FakeSocket``` instance for **every** request. was this by design - for a particular reason - or just something we did to get over an issue? it obv. adds unnecessary overhead but also, in case of this memory leak, it is a big problem when the middleware is using each socket instance as the key into it's Map. apart from the memory leak, it means all their counts for requests per socket will be 1 maximum. 

```JavaScript
server.on('connection', (socket) => {
    console.log(`server.onconnection`);
    this.requestCountPerSocket.set(socket, 0);
    console.log(socket.constructor.name)
    socket.once('close', () => {
        console.log('socket.once.close')
        this.requestCountPerSocket.delete(socket)
    });
});
server.on('request', (req, res) => {
    this.requestCountPerSocket.set(req.socket, (this.requestCountPerSocket.get(req.socket) ?? 0) + 1);
    res.once('finish', () => {
        const pending = (this.requestCountPerSocket.get(req.socket) ?? 0) - 1;
        console.log(`response.onfinish ${pending}`)
        this.requestCountPerSocket.set(req.socket, pending);
        if (this.stopped && pending === 0) {
            console.log(`req.socket.end`)
            req.socket.end();
        }
    });
});
```

- In my PR i am creating a single ```FakeSocket``` per socketidentifier, which i have surfaced to JS, but i don't know if that will cause problems elsewhere - needs further testing.
- I am not sure how best to surface the close event for underlying sockets to the node:http module. any advice appreciated.

### TODO

- run existing tests
- add specific tests for this issue to ensure memleak is not happening
- handle and test with other kinds of sockets where ```FakeSocket``` is used
- handle the misidentification of the connection as a ```secureConnection```
- Surface/handle the socket close event in node:http module and fire the 'close' event to any listeners when it is received. without this, the middleware in question will keep existing sockets in it's internal map after they are closed, which is a memory leak, but obv. one that will happen at a much slower rate than currently given we don't now create a ```FakeSocket``` on every request.

### Other Notes

- after this change, we don't see the explosive memory growth and memory usage on macos is ~150MB compared to ~230 MB for node.js. RPS is ~68k RPS versus 23k RPS for node.js.

### DIFF
--- a/src/bun.js/api/server.classes.ts
+++ b/src/bun.js/api/server.classes.ts
@@ -32,6 +32,10 @@ function generate(name) {
         fn: "doRequestIP",
         length: 1,
       },
+      requestSocketId: {
+        fn: "doRequestSocketId",
+        length: 1,
+      },
       port: {
         getter: "getPort",
       },
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -1324,6 +1324,13 @@ pub const AnyRequestContext = struct {
         return .{ .tagged_pointer = Pointer.init(request_ctx) };
     }
 
+    pub fn getRemoteSocketId(self: AnyRequestContext) i32 {
+        if (self.tagged_pointer.as(HTTPServer.RequestContext).getNativeHandle()) |handle| {
+            return handle.cast();
+        }
+        return 0;
+    }
+
     pub fn getRemoteSocketInfo(self: AnyRequestContext) ?uws.SocketAddress {
         if (self.tagged_pointer.isNull()) {
             return null;
@@ -3584,6 +3591,10 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             return onStartStreamingRequestBody(bun.cast(*RequestContext, this));
         }
 
+        pub fn getNativeHandle(this: *RequestContext) ?bun.FileDescriptor {
+            return (this.resp orelse return null).getNativeHandle();
+        }
+
         pub fn getRemoteSocketInfo(this: *RequestContext) ?uws.SocketAddress {
             return (this.resp orelse return null).getRemoteSocketInfo();
         }
@@ -5321,6 +5332,7 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
         pub const doReload = onReload;
         pub const doFetch = onFetch;
         pub const doRequestIP = JSC.wrapInstanceMethod(ThisServer, "requestIP", false);
+        pub const doRequestSocketId = JSC.wrapInstanceMethod(ThisServer, "socketId", false);
 
         pub usingnamespace NamespaceType;
         pub usingnamespace bun.New(@This());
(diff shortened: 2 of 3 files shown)


robobun:
Your commit 2467f60e78163d28b3d075d247fbd285888c608e is building: [`#1722`](https://buildkite.com/bun/bun/builds/1722)

src/bun.js/api/server.zig
```diff
@@ -1324,6 +1324,13 @@ pub const AnyRequestContext = struct {
         return .{ .tagged_pointer = Pointer.init(request_ctx) };
     }
 
+    pub fn getRemoteSocketId(self: AnyRequestContext) i32 {
+        if (self.tagged_pointer.as(HTTPServer.RequestContext).getNativeHandle()) |handle| {
```
Jarred-Sumner, line 1328: This would be needed for each type matching the tagged pointer, not just the one type here.

src/bun.js/api/server.zig
```diff
@@ -5332,6 +5344,10 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
 
         extern fn JSSocketAddress__create(global: *JSC.JSGlobalObject, ip: JSValue, port: i32, is_ipv6: bool) JSValue;
 
+        pub fn socketId(_: *ThisServer, request: *JSC.WebCore.Request) JSC.JSValue {
```
Jarred-Sumner, line 5347: This doesn't need to be a public, user-facing method.

This can instead be a private function only used in node:http.

src/js/node/http.ts
```diff
@@ -149,6 +151,13 @@ var FakeSocket = class Socket extends Duplex {
     return this;
   }
 
+  end() {
+    if (socketMap.has(this.#id)) {
```
Jarred-Sumner, line 155: We can skip the `.has` call, `delete` returns true if the key matched
```suggestion
    if (socketMap.delete(this.#id)) {
```

src/bun.js/api/server.classes.ts
```diff
@@ -32,6 +32,10 @@ function generate(name) {
         fn: "doRequestIP",
         length: 1,
       },
+      requestSocketId: {
```
Jarred-Sumner, line 35: No need for a public API here

```suggestion

```

src/js/node/http.ts
```diff
@@ -94,8 +94,8 @@ var _defaultHTTPSAgent;
 var kInternalRequest = Symbol("kInternalRequest");
 const kInternalSocketData = Symbol.for("::bunternal::");
 const kfakeSocket = Symbol("kfakeSocket");
-
 const kEmptyBuffer = Buffer.alloc(0);
+const socketMap = new Map(); // keep track of active sockets so we can handle socket events correctly
```
Jarred-Sumner, line 98:
```suggestion
const socketMap = new Map(); // keep track of active sockets so we can handle socket events correctly
const socketIdFromRequest = $newZigFunction("server.zig", "socketIdFromRequest", 2);
```

Jarred-Sumner (changes requested):
- We don't need to expose a user-facing API for this
- It might be better to do this in uWebSockets, it could involve a change to `AsyncSocket.h`

src/bun.js/api/server.zig
billywhizz: thanks - yeah, i had deleted those while testing. i just put this up as a draft to get some feedback on the approach first.

PR #13259, opened 2024-08-13
Make the error when udpSocket creation fails better

Jarred-Sumner:


robobun:
@Jarred-Sumner, your commit 01d5d1707be903be6ce93965cf011c2c28cce5cc is building: [`#1731`](https://buildkite.com/bun/bun/builds/1731)

Issue #1768, opened 2023-01-12
Cannot set response `status` to `101` 


Electroid:
@robobun

```js
console.log(new Response(null, { status: 101 }));
```

robobun:
@Electroid here you go!
```tsx
Response (0 KB) {
  ok: true,
  url: "",
  statusText: "",
  redirected: false,
  bodyUsed: false,
  status: 200
}
```
<details>
  <summary>Code</summary>
  
```tsx
console.log(new Response(null, { status: 101 }));
```
</details>

_Ran using the latest build of Bun, an all-in-one JavaScript runtime._

PR #13278, opened 2024-08-14
node:fs: use libuv callbacks instead of custom workpool for some operations on windows

nektro:
rather than using libuv's blocking calls on a new thread this uses libuv's callback mechanism to let it decide when to do the work instead. this should be more robust.

this initial PR covers open, close, read, write, readv, writev


robobun:
@nektro, your commit 8ae60f31459589271954b444314bc42ea3b73679 is building: [`#1768`](https://buildkite.com/bun/bun/builds/1768)

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+                const this: *Task = @ptrCast(@alignCast(req.data.?));
+                var node_fs = NodeFS{};
+                this.result = @field(NodeFS, "uv_" ++ @tagName(FunctionEnum))(&node_fs, this.args, @intFromEnum(req.result));
+
+                if (this.result == .err) {
+                    this.result.err.path = bun.default_allocator.dupe(u8, this.result.err.path) catch "";
+                    std.mem.doNotOptimizeAway(&node_fs);
+                }
+
+                this.globalObject.bunVMConcurrently().eventLoop().enqueueTaskConcurrent(JSC.ConcurrentTask.createFrom(this));
```
Jarred-Sumner: Since `uv_callback` is run on the same thread, we don't need to use enqueueTaskConcurrent.

```suggestion
                this.globalObject.bunVM().eventLoop().enqueueTask(JSC.Task.init(this))
```

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+                    bun.default_allocator.free(this.result.err.path);
+                }
+
+                this.ref.unref(this.globalObject.bunVM());
+                if (@hasDecl(ArgumentType, "deinitAndUnprotect")) {
+                    this.args.deinitAndUnprotect();
+                } else {
+                    this.args.deinit();
+                }
+                this.promise.strong.deinit();
```
```suggestion
                this.promise.deinit();
```

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+                bun.destroy(this);
```
Jarred-Sumner: Use `this.destroy` instead of `bun.destroy`

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+            args: ArgumentType,
+            globalObject: *JSC.JSGlobalObject,
+            req: uv.fs_t = std.mem.zeroes(uv.fs_t),
+            result: JSC.Maybe(ReturnType),
+            ref: bun.Async.KeepAlive = .{},
+            tracker: JSC.AsyncTaskTracker,
+
+            pub const Task = @This();
+
+            pub const heap_label = "Async" ++ bun.meta.typeBaseName(@typeName(ArgumentType)) ++ "UvTask";
```
Jarred-Sumner, line 173:
```suggestion
            pub const heap_label = "Async" ++ bun.meta.typeBaseName(@typeName(ArgumentType)) ++ "UvTask";
            pub usingnamespace bun.New(@This());
```

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+
+            pub fn create(globalObject: *JSC.JSGlobalObject, args: ArgumentType, vm: *JSC.VirtualMachine) JSC.JSValue {
+                var task = bun.new(Task, Task{
```
```suggestion
                var task = Task.new(.{
```

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+                });
+                task.ref.ref(vm);
+                task.args.toThreadSafe();
+
+                task.tracker.didSchedule(globalObject);
+
+                const log = bun.sys.syslog;
+                const loop = uv.Loop.get();
+                task.req.data = task;
+                switch (FunctionEnum) {
```
Jarred-Sumner: Let's make sure it doesn't increase the stack size for this function:
```suggestion
                switch (comptime FunctionEnum) {
```

src/bun.js/node/node_fs.zig
```diff
@@ -141,6 +143,185 @@ pub const Async = struct {
...
+                        var buf: bun.PathBuffer = undefined; // uv copies path so its okay to use stack memory for this
+                        const path = args.path.sliceZ(&buf);
+
+                        var flags: c_int = @intFromEnum(args.flags);
+                        flags = uv.O.fromBunO(flags);
+
+                        var mode: c_int = args.mode;
+                        if (mode == 0) mode = 0o644;
+
+                        bun.debugAssert(uv.uv_fs_open(loop, &task.req, path.ptr, flags, mode, &uv_callback) == .zero);
```
Jarred-Sumner: Don't use `debugAssert` for side-effectful things. Let's copy into a local variable. I'm also not sure if these should be debugAssert, as users may provide invalid input and that's not an assertion error.

```
const rc = uv.uv_fs_open(loop, &task.req, path.ptr, flags, mode, &uv_callback);
bun.debugAssert(rc == .zero);

Jarred-Sumner:
@nektro the spawn test timeouts are new

Issue #1775, opened 2023-01-12
Improve test coverage for bunx


Jarred-Sumner:
The following don't work but definitely should:

- `bunx webpack-cli`
- `bunx @babel/cli`

cometkim:
`bun x` works fine in all cases. this should verify the difference between alias `bunx` and `bun x`.

cometkim:
alias version of bunx works correctly only for packages currently installed from node_modules .

dylan-conway:
These packages should work with `bunx`/`npx` after installing their dependencies. `babel-cli` requires `@babel/core` and `@babel/cli`, `webpack` requires `webpack-cli`.
https://babeljs.io/docs/en/babel-cli#usage
https://webpack.js.org/api/cli/

PR #13280, opened 2024-08-14
rework src/js_lexer/identifier_data.zig

nektro:
this would appear to hang running in CI in sometimes. this should prevent that by moving the work it was doing from comptime to runtime


robobun:
@nektro, your commit 5a11baa30a63def151d72c9ae2ac339bd1ebd5bf is building: [`#1775`](https://buildkite.com/bun/bun/builds/1775)

PR #13290, opened 2024-08-14
node:net.Server listen handler should be bound to the server

nektro:
found while running the node tests

node:net version of https://github.com/oven-sh/bun/pull/10533


src/js/node/net.ts
```diff
@@ -984,7 +984,7 @@ class Server extends EventEmitter {
       // That leads to all sorts of confusion.
       //
       // process.nextTick() is not sufficient because it will run before the IO queue.
-      setTimeout(emitListeningNextTick, 1, this, onListen);
+      setTimeout(emitListeningNextTick, 1, this, onListen.bind(this));
```
Jarred-Sumner: Instead of allocating the bound function, we can use `apply`.

PR #13291, opened 2024-08-14
node:net.Socket#{ref,unref} are supposed to return this

nektro:
found while running the node tests


robobun:
@nektro, your commit 6ac332405e40c5ea3ad1b6e9f294bacc817cae39 is building: [`#1825`](https://buildkite.com/bun/bun/builds/1825)

src/js/node/net.ts
```diff
@@ -686,6 +686,7 @@ const Socket = (function (InternalSocket) {
         return;
```
```suggestion
        return this;
```

src/js/node/net.ts
```diff
@@ -724,6 +725,7 @@ const Socket = (function (InternalSocket) {
...
```
```suggestion
        return this;
```

Issue #13001, opened 2024-08-01
Segmentation Fault using Nuxt Nitro Server


yschroe:
### How can we reproduce the crash?

Edit: Managed to reproduce the crash reliably on my system:

Repro repo: https://github.com/yschroe/bun-segfault-nuxt

1. Start the dev server via `bun --bun run dev`.
2. Navigate to `server/api/test.ts`.
3. Just hit CTRL+S a few times to save the file over and over again.
4. Each save should trigger `✔ Nuxt Nitro server built in xx ms` in the console (I am using the prettier Plugin for VS Code - not sure if it is a prerequisite to actually trigger a rebuild due to reformatting).
5. After a few times (10-30) bun crashes on my system.

### Relevant log output

```shell
✔ Nuxt Nitro server built in 88 ms                                                                                                                                                       nitro 14:42:39
✔ Nuxt Nitro server built in 89 ms                                                                                                                                                       nitro 14:42:39
✔ Nuxt Nitro server built in 96 ms                                                                                                                                                       nitro 14:42:39
============================================================
Bun v1.1.21 (70ca2b76) Windows x64
CPU: sse42 avx avx2
Args: "node" "C:\Users\xxx\Documents\repos\xxx\node_modules\nuxi\bin\nuxi.mjs" "dev"
Features: jsc Bun.stderr(4) Bun.stdin(26) Bun.stdout(3) extracted_packages(853) fetch(836) http_server(13) spawn(4) transpiler_cache(63) tsconfig_paths(40) tsconfig(106) 
Builtins: "bun:jsc" "bun:main" "detect-libc" "node:assert" "node:async_hooks" "node:buffer" "node:child_process" "node:constants" "node:crypto" "node:dns" "node:events" "node:fs" "node:fs/promises" "node:http" "node:https" "node:module" "node:net" "node:os" "node:path" "node:perf_hooks" "node:process" "node:querystring" "node:readline" "node:stream" "node:stream/promises" "node:string_decoder" "node:tls" "node:tty" "node:url" "node:util" "node:util/types" "node:vm" "node:zlib" "node:worker_threads" "ws" "node:v8" 
Elapsed: 4159168ms | User: 28625ms | Sys: 5359ms
RSS: 0.94GB | Peak: 2.11GB | Commit: 7.14GB | Faults: 1350488

panic(thread 22756): Segmentation fault at address 0x474BE410278
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

[...]

Issue #1865, opened 2023-01-20
segmentation fault


wmelton:
### What version of Bun is running?

0.5.1

### What platform is your computer?

Darwin 21.6.0 x86_64 i386

### What steps can reproduce the bug?

Create this js file named 'subscribe.js' and try to run it with bun after installing nats: 

```js
import { connect, StringCodec } from "nats";

const nc = await connect({ servers: "demo.nats.io:4222" });
const sc = StringCodec();

// subscriptions can have wildcard subjects
// the '*' matches any string in the specified token position
const s1 = nc.subscribe("help.*.system");
const s2 = nc.subscribe("help.me.*");
// the '>' matches any tokens in that position or following
// '>' can only be specified at the end of the subject
const s3 = nc.subscribe("help.>");

async function printMsgs(s) {
  let subj = s.getSubject();
  console.log(`listening for ${subj}`);
  const c = (13 - subj.length);
  const pad = "".padEnd(c);
  for await (const m of s) {
    console.log(
      `[${subj}]${pad} #${s.getProcessed()} - ${m.subject} ${
        m.data ? " " + sc.decode(m.data) : ""
      }`,
    );
  }
}

printMsgs(s1);
printMsgs(s2);
printMsgs(s3);

// don't exit until the client closes
await nc.closed();
```

### What is the expected behavior?

Runs correctly like node.js

### What do you see instead?

`zsh: segmentation fault  bun subscribe.js`

### Additional information

I wasn't able to trigger a more detailed failure message. `zsh: segmentation fault  bun subscribe.js` is all that's given.

Jarred-Sumner:
Can you go into Console.app and paste the stack trace? If that doesn't have much info, can you download `bun-profile` for x64 from the releases page?

On my M1, it reports this stack trace:
```js
❯ bun index.ts
88 |      * @constructor
89 |      *
90 |      * @api private
91 |      */
92 |     constructor(message, code, chainedError) {
93 |         super(message);
            ^
NatsError: TLS
 code: "TLS"

      at new NatsError (/Users/jarred/Build/bun-repros/1865/node_modules/nats/lib/nats-base-client/error.js:93:8)
      at errorForCode (/Users/jarred/Build/bun-repros/1865/node_modules/nats/lib/nats-base-client/error.js:101:15)
      at /Users/jarred/Build/bun-repros/1865/node_modules/nats/lib/src/node_transport.js:258:25
[...]

PR #13313, opened 2024-08-14
Fixes https://github.com/oven-sh/bun/issues/13001

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/13001

Manually. We do need to write a test for this, somehow. Really we should have an integration test for Nuxt.


robobun:
@Jarred-Sumner, your commit e01dd931f21df5de86c106e0f25c897db831b407 is building: [`#1865`](https://buildkite.com/bun/bun/builds/1865)

PR #13318, opened 2024-08-15
Fix "bun exec" examples

vktrl:
### What does this PR do?

Not much, just fixed examples in `bun exec` output when run without args.

- [ ] Code changes


Jarred-Sumner:
Thanks

PR #1895, opened 2023-01-25
parse package-spec from CLI correctly

alexlamsl:
fixes #861

PR #13320, opened 2024-08-15
fix(TextEncoder): domjit crash in `encode`

### LINKED ISSUE #13298: Next.js standalone with edge - Segmentation fault

My example may be more weird then the average, but on a page with edge runtime specified and that uses cookies it crashes upon request - also note on node it doesn't have the cookies error.

fwiw the standalone artifact is built on github actions: https://github.com/RiskyMH/EmailThing/actions/runs/10387115545 (can download the zip here too) and page is /home

### Relevant log output

```shell
▲ Next.js 14.2.5
  - Local:        http://localhost:3000
  - Network:      http://0.0.0.0:3000

 ✓ Starting...
 ✓ Ready in 162ms
 [Error: `cookies` was called outside a request scope. Read more: https://nextjs.org/docs/messages/next-dynamic-api-wrong-context] {
  digest: '3823075302'
}
 [Error: `cookies` was called outside a request scope. Read more: https://nextjs.org/docs/messages/next-dynamic-api-wrong-context] {
  digest: '88751037'
}
Bun v1.1.24 (85a32991) Linux x64
Linux Kernel v6.8.0 | glibc v2.39                                                                                                                           CPU: sse42 popcnt avx avx2
Args: "bun" "data/RiskyMH_EmailThing/latest/server.js"
Features: jsc abort_signal(2) dotenv fetch(53) http_server transpiler_cache(33) tsconfig
Builtins: "bun:jsc" "bun:main" "detect-libc" "node:assert" "node:async_hooks" "node:buffer" "node:child_process" "node:constants" "node:crypto" "node:events" "node:fs" "node:fs/promises" "node:http" "node:https" "node:module" "node:net" "node:os" "node:path" "node:perf_hooks" "node:querystring" "node:stream" "node:stream/web" "node:string_decoder" "node:tls" "node:tty" "node:url" "node:util" "node:util/types" "node:vm" "node:zlib" "node:worker_threads" "node:punycode" "ws" "node:v8" "node:diagnostics_channel"
Elapsed: 22363ms | User: 6797ms | Sys: 763ms
RSS: 0.99GB | Peak: 0.30GB | Commit: 0.99GB | Faults: 0

panic(main thread): Segmentation fault at address 0x0

 https://bun.report/1.1.24/la185a3299AipihwCu1o/jE+xpR00oy7D__8ivm1D___091m1DA2AA

Illegal instruction (core dumped)
```

### Stack Trace (bun.report)

Bun v1.1.24 ([`85a3299`](<https://github.com/oven-sh/bun/tree/85a32991152871b06becd6b411eb663008369769>)) on linux x86_64 [AutoCommand]

Segmentation fault at address 0x00000000

- [`JSTextEncoder.cpp:265`](<https://github.com/oven-sh/bun/blob/85a32991152871b06becd6b411eb663008369769/src/bun.js/bindings/webcore/JSTextEncoder.cpp#L265>): `jsTextEncoderEncodeWithoutTypeCheck`
- *2 unknown/js code*
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/85a32991152871b06becd6b411eb663008369769/ld-temp.o#L0>): `llint_op_call`
- *3 unknown/js code*
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/85a32991152871b06becd6b411eb663008369769/ld-temp.o#L0>): `llint_op_call_ignore_result`

dylan-conway:
### What does this PR do?
Inside a NodeVM context, bun could crash in a domjit function when encoding a 8-bit string longer than 1024

fixes #13298 

Also fixes a bug where `TextEncoderStream` would return `Buffer` chunks instead of `Uint8Array` chunks

added test for both bugs


robobun:
@dylan-conway, your commit cc907b1be58ef016b3111c334550de9871181983 is building: [`#1895`](https://buildkite.com/bun/bun/builds/1895)

src/codegen/generate-classes.ts
```diff
@@ -1275,6 +1316,22 @@ function generateClassHeader(typeName, obj: ClassDefinition) {
   `.trim();
 }
 
+function domJITTypeCheckFields(proto, klass) {
+  var output = "";
```
Jarred-Sumner, line 1320:
```suggestion
  var output = "#if ASSERT_ENABLED" + "\n";
```

src/codegen/generate-classes.ts
```diff
@@ -1275,6 +1316,22 @@ function generateClassHeader(typeName, obj: ClassDefinition) {
...
+    if (!DOMJIT) continue;
+    output += `std::optional<std::optional<JSC::JSType>> m_${fn}_expectedResultType = std::nullopt;\n`;
+  }
+
+  for (const name in klass) {
+    const { DOMJIT, fn } = klass[name];
+    if (!DOMJIT) continue;
+    output += `std::optional<std::optional<JSC::JSType>> m_${fn}_expectedResultType = std::nullopt;\n`;
+  }
+  return output;
```
Jarred-Sumner, line 1334:
```suggestion
  return output ? ("#if ASSERT_ENABLED" + "\n" + output + "\n" + "#endif" + "\n") : output
```

test/js/web/encoding/text-encoder-stream.test.ts
```diff
@@ -135,6 +135,7 @@ for (const { input, output, description } of testCases) {
     const chunkArray = await Bun.readableStreamToArray(outputStream);
     expect(chunkArray.length, "number of chunks should match").toBe(output.length);
     for (let i = 0; i < output.length; ++i) {
+      expect(chunkArray[i]).toBeInstanceOf(Uint8Array);
```
Electroid: If the bug was chunks were `Buffer` instead of `Uint8Array`, wouldn't this test pass with the bug?

Jarred-Sumner: the bug was in the differing JSType and Buffer and Uint8Array have the same JSType

dylan-conway: Will update test, it won't catch it

Jarred-Sumner:
merge conflicts

Issue #1939, opened 2023-01-30
fs.createWriteStream quits before the end of a large(ish) file


controversial:
### What version of Bun is running?

0.5.1 / 0.5.2

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

First find a reasonably large file (mine is ~322 KB). Then:
```js
import fs from 'fs';

let len = 0;
fs.createReadStream(new URL('./path/to/largeish/file', import.meta.url))
  .on('data', (chunk) => {
    len += chunk.toString().length;
    console.log(len);
  })
  .on('end', () => console.log('done!'));
```

Under `node script.js`, this produces:
```
65526
131054
196578
262110
327644
329912
done!
```
Under `bun script.js`, this produces:
```
65526
131054
196578
done!
```

### What is the expected behavior?

Bun should read and stream the whole file.

### What do you see instead?

Bun only reads and streams a few chunks of the file.

### Additional information

Simple reproduction in one screenshot:
<img width="723" alt="Screenshot 2023-01-30 at 4 05 08 AM" src="https://user-images.githubusercontent.com/10377391/215433413-1c3a9649-4f41-4424-8fba-2fc2537d78d0.png">

Jarred-Sumner:
Failing test:

```js

it("Bun.file().stream() read text from large file", async () => {
  var hugely = "HELLO!".repeat(1024 * 1024 * 10);
  const tmpfile = resolve(tmpdir() + "/bun-streams-test.txt");
  writeFileSync(tmpfile, hugely);
  var chunks = [];
  for await (const chunk of Bun.file(tmpfile).stream()) {
    chunks.push(chunk);
  }
  const output = Buffer.concat(chunks).toString();
  expect(output).toHaveLength(hugely.length);
  expect(output).toBe(hugely);
});
```

PR #13335, opened 2024-08-16
Support Unicode strings in V8 APIs

190n:
### What does this PR do?

Adds support in v8::String::WriteUtf8 for converting UTF-16 strings into UTF-8.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/bun.js/bindings/v8/V8Object.cpp
+++ b/src/bun.js/bindings/v8/V8Object.cpp
@@ -62,10 +62,10 @@ Maybe<bool> Object::Set(Local<Context> context, Local<Value> key, Local<Value> v
 
 void Object::SetInternalField(int index, Local<Data> data)
 {
-    auto fields = getInternalFieldsContainer(this);
-    if (fields && index >= 0 && index < fields->size()) {
-        fields->at(index) = InternalFieldObject::InternalField(data->localToJSValue(Isolate::GetCurrent()->globalInternals()));
-    }
+    auto* fields = getInternalFieldsContainer(this);
+    RELEASE_ASSERT(fields, "object has no internal fields");
+    RELEASE_ASSERT(index >= 0 && index < fields->size(), "internal field index is out of bounds");
+    fields->at(index) = InternalFieldObject::InternalField(data->localToJSValue(Isolate::GetCurrent()->globalInternals()));
 }
 
 Local<Data> Object::GetInternalField(int index)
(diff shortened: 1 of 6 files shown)


robobun:
@190n, your commit 9d34cd7ddcb6db2d07fa4ff1537beac89ca768a2 is building: [`#1939`](https://buildkite.com/bun/bun/builds/1939)

src/bun.js/bindings/v8/V8Object.cpp
```diff
@@ -62,10 +62,10 @@ Maybe<bool> Object::Set(Local<Context> context, Local<Value> key, Local<Value> v
 
 void Object::SetInternalField(int index, Local<Data> data)
 {
-    auto fields = getInternalFieldsContainer(this);
-    if (fields && index >= 0 && index < fields->size()) {
-        fields->at(index) = InternalFieldObject::InternalField(data->localToJSValue(Isolate::GetCurrent()->globalInternals()));
-    }
+    auto* fields = getInternalFieldsContainer(this);
+    RELEASE_ASSERT(fields, "object has no internal fields");
```
Jarred-Sumner, line 66: Can these be `ASSERT`? This will be expensive in a release build

src/bun.js/bindings/v8/V8Object.cpp
```diff
@@ -62,10 +62,10 @@ Maybe<bool> Object::Set(Local<Context> context, Local<Value> key, Local<Value> v
...
+    RELEASE_ASSERT(index >= 0 && index < fields->size(), "internal field index is out of bounds");
+    fields->at(index) = data->localToJSValue(Isolate::GetCurrent()->globalInternals());
```
Jarred-Sumner, line 68: Do we need a lock for this?

src/bun.js/bindings/v8/V8Object.cpp
190n: This can be triggered by modules that misuse the V8 API. I think it's better in those cases to have a more descriptive error message rather than segfaulting. Also, segfault will show the "this is an error in Bun not your code" message which isn't really true.

src/bun.js/bindings/v8/V8Object.cpp
190n: I don't think so. [It doesn't seem like V8's API is thread-safe either](https://stackoverflow.com/a/7921673).

PR #13341, opened 2024-08-16
Update documentation about .stringMatching

mathiasrw:
### What does this PR do?

This will update the documentation of expect().stringMatching  

`.stringMatching` was documented to be supported, but bun v1.1.24 gives the message `.stringMatching is not a function`. 

(I recommend using the supported `.toMatch` instead.)

- [ ] Code changes

### DIFF
--- a/docs/test/writing.md
+++ b/docs/test/writing.md
@@ -394,7 +394,7 @@ Bun implements the following matchers. Full Jest compatibility is on the roadmap
 
 ---
 
-- ✅
+- ❌
 - [`.stringMatching()`](https://jestjs.io/docs/expect#expectstringmatchingstring--regexp)
 
 ---


Jarred-Sumner:
Can you paste the place where you run into this error?

expect.stringMatching is implemented. We use it in our own tests:

https://github.com/oven-sh/bun/blob/4c56d8d778efcc76943174df65e455abe9b39cdd/test/js/bun/test/expect.test.js#L3256-L3260

mathiasrw:
Interesting. 

Here you can see the output when using `stringMatching` instead of `toMatch`

image

You can recreate it via 

```js
// @ts-ignore
import {expect, test, describe} from 'bun:test';

describe('Demo of expeect', () => {
	test('Using toMatch', () => {
		expect('foobar').toMatch('foo');
	});

	test('Using toMatch', () => {
		expect('foobar').stringMatching('foo');
	});
});
```

Would you like me to close the PR and create an issue instead?

Jarred-Sumner:
[`expect.stringMatching`](https://jestjs.io/docs/expect#expectstringmatchingstring--regexp) is a Jest matcher on the `expect` object (static), not on the `expect` instance unless I'm confused

mathiasrw:
Oh! - got it. 

Thanks.

PR #1970, opened 2023-02-02
resolve duplicate npm dependencies correctly

alexlamsl:
fixes #1952

PR #13348, opened 2024-08-16
fix(ws test) deflaky

cirospaciari:
### What does this PR do?

- [X] Code changes

Tests

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -4166,7 +4166,7 @@ pub const ServerWebSocket = struct {
             return;
 
         // we should not call drain if we dont have this_value
-        if (handler.onDrain != .zero and this.this_value.isEmptyOrUndefinedOrNull()) {
+        if (handler.onDrain != .zero and !this.this_value.isEmptyOrUndefinedOrNull()) {
             const globalObject = handler.globalObject;
 
             var corker = Corker{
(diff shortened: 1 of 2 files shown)


robobun:
@cirospaciari, your commit 14c204ab8f91d7bcbbdc968aac0661038da27091 is building: [`#1970`](https://buildkite.com/bun/bun/builds/1970)

src/bun.js/api/server.zig
```diff
@@ -4165,7 +4165,8 @@ pub const ServerWebSocket = struct {
         if (this.isClosed() or vm.isShuttingDown())
             return;
 
-        if (handler.onDrain != .zero) {
+        // we should not call drain if we dont have this_value
+        if (handler.onDrain != .zero and !this.this_value.isEmptyOrUndefinedOrNull()) {
```
Jarred-Sumner: Seems kind of sketchy. Maybe we should call `getThisValue()` instead?

cirospaciari: Actually Im debugging to see where is the issue but `drain` event is called passing a null value, probably we are unprotecting to soon?

cirospaciari: Yeah `getThisValue` was not called before because I did not add a open callback, so getThisValue fix this issue

PR #1977, opened 2023-02-03
Implement os.cpus() for Linux

jwhear:
Partially fixes #1888 

`os.cpus()` currently returns an empty array for all platforms.  This PR implements full functionality for Linux and has been tested on x86-64.  Other OSes will continue to return an empty array.

Note that Linux on Arm64 may report the CPU model differently; if this is the case the CPU model will currently be reported as "unknown".  As I do not have Arm64 hardware to verify and develop against, a todo has been left in the code.

PR #13351, opened 2024-08-16
Experiment: remove visitOutputConstraints

Jarred-Sumner:
### What does this PR do?

We are defining visitOutputConstraints when we don't need to

Maybe Bun will free things earlier if we don't do this

image

CI


robobun:
@Jarred-Sumner, your commit 5337f1d29dc64c1cab58a54d8e2398a696d1c657 is building: [`#1977`](https://buildkite.com/bun/bun/builds/1977)

Issue #1980, opened 2023-02-03
After `bun install`, package.json has weird formatting


Electroid:
Before
```json
{
  "dependencies": {},
}
```

After
```json
{
  "dependencies": { "@actions/core": "^1.10.0" },
}
```

It should look like this instead:
```json
{
  "dependencies": {
    "@actions/core": "^1.10.0"
  },
}
```

Jarred-Sumner:
Let's not prioritize this for awhile

R4R3-M37:
any news? Tabs auto convert to 2 spaces

orbiteleven:
Yeah this is actually quite annoying; if I have Prettier set to tabs but use `bun add|install` via the command line, it'll convert my tabs to spaces, requiring a re-format of `package.json`. It's trivial, but not very ergonomic.

PR #13352, opened 2024-08-16
Use bun instead of npm even for Node.js build in V8 tests

190n:
### What does this PR do?

Previously `v8.test.ts` ran `npm install` to build the V8 test module against Node.js. Now it uses `bun install`, which still runs `node-gyp` with Node.js internally (the desired behavior) but is faster to install dependencies than `npm install`. This hopefully speeds up CI.

- [x] Code changes

I re-ran all the V8 tests on macOS and Windows.


robobun:
@190n, your commit 1badac7c1d5150afae3d3e2bf75bff3ba413fe75 is building: [`#1980`](https://buildkite.com/bun/bun/builds/1980)

PR #13358, opened 2024-08-17
fix(sockets) always uncork when closing

cirospaciari:
### What does this PR do?
Make sure that sockets always uncorked when closed

- [x] Code changes

Existing tests


robobun:
@cirospaciari, your commit 2396c319a2d2fffd9f78ae118628fc31005579e3 is building: [`#1997`](https://buildkite.com/bun/bun/builds/1997)

packages/bun-uws/src/AsyncSocket.h
```diff
@@ -166,33 +166,34 @@ struct AsyncSocket {
...
+                ourCorkOffset = loopData->getCorkOffset();
+                loopData->setCorkOffset(0);
             }
 
             /* Fallback is to use the backpressure as buffer */
             backPressure.resize(ourCorkOffset + existingBackpressure + size);
 
             /* And copy corkbuffer in front */
-            memcpy((char *) backPressure.data() + existingBackpressure, loopData->corkBuffer, ourCorkOffset);
+            memcpy((char *) backPressure.data() + existingBackpressure, loopData->getCorkBuffer(), ourCorkOffset);
```
Jarred-Sumner: `ourCorkOffset` will never be zero right?

cirospaciari: I think it can be in edge cases, still worth checking

packages/bun-uws/src/LoopData.h
```diff
@@ -44,7 +44,11 @@ struct alignas(16) LoopData {
 
     /* Map from void ptr to handler */
     std::map<void *, MoveOnlyFunction<void(Loop *)>> postHandlers, preHandlers;
-
+    /* Cork data */
+    char *corkBuffer = new char[CORK_BUFFER_SIZE];
```
Jarred-Sumner, line 48: We would need to free this

If we instead make this statically allocated, we wouldn't need to worry about freeing it:
```suggestion
    char corkBuffer[CORK_BUFFER_SIZE];
```

Jarred-Sumner: oh this already existed

PR #2039, opened 2023-02-10
[install] fix duplicate check on `peerDependencies`

alexlamsl:
fixes #2037

PR #13381, opened 2024-08-18
Enable reusePort in Bun.serve() by default when using node:cluster

Jarred-Sumner:
### What does this PR do?

Enable reusePort in Bun.serve() by default when using node:cluster

This also makes it so frameworks like Hono which currently don't seem to expose this option can just work when cluster is enabled

CI

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -849,7 +849,7 @@ pub const ServerConfig = struct {
             },
             .development = true,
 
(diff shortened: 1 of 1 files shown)


robobun:
@Jarred-Sumner, your commit 789ee02865635e18daf78a1810b4dc364e4168f5 is building: [`#2039`](https://buildkite.com/bun/bun/builds/2039)

Issue #2102, opened 2023-02-18
Response doesn't stream when `Transfer-Encoding` is set to `chunked` in express.js


noway:
### What version of Bun is running?

0.5.6

### What platform is your computer?

Darwin 22.2.0 arm64 arm

### What steps can reproduce the bug?

1. Run repro script:

```bash
curl https://gist.githubusercontent.com/noway/7097e86dc5650a762b156a63e43ff6d7/raw/da6c5a8cf31bb4b33ed314dfd71c15bf9eb920e9/bun-express-streaming-response-bug.sh | bash
```
   - Alternatively, go to https://gist.github.com/noway/7097e86dc5650a762b156a63e43ff6d7 and copy-paste the script into Terminal.

2. Run `curl localhost:3001` in Terminal
3. Run `curl localhost:3002` in Terminal
4. See the difference between step 2 and step 3. Port `3002` (node) is streaming. Port `3001` (bun) is not streaming.

### What is the expected behavior?

The expected behaviour is for bun (port `3001`) to stream (display the output line-by-line, waiting 300 ms in-between every line) just like node (port `3002`) does.

### What do you see instead?

Bun doesn't stream, it spits out the response as a whole when `res.end()` is called.

### Additional information

N/A

noway:
This is no longer reproducible in bun v1.0.2 🎉

PR #13427, opened 2024-08-20
Everything is cmake

Electroid:
### What does this PR do?

This is a work-in-progress PR that will consolidate all of our build scripts to `cmake`. The goal is that you only need to run `cmake` to build Bun from scratch.

- [ ] Code changes


robobun:
@Electroid, your commit 2e26b66f9ef8362b7c702155564d38d406618060 is building: [`#2102`](https://buildkite.com/bun/bun/builds/2102)

cmake/tools/SetupWebKit.cmake
```diff
@@ -0,0 +1,134 @@
...
+file(REMOVE ${CACHE_PATH}/${WEBKIT_FILENAME})
+file(REMOVE_RECURSE ${WEBKIT_PATH})
+file(RENAME ${CACHE_PATH}/bun-webkit ${WEBKIT_PATH})
+
+if(APPLE)
+  file(REMOVE_RECURSE ${WEBKIT_INCLUDE_PATH}/unicode)
+endif()
+
+# --- WebKit ---
+# WebKit is either prebuilt and distributed via NPM, or you can pass WEBKIT_PATH to use a local build.
```
Jarred-Sumner, line 83: we don't distribute via npm

Jarred-Sumner: its too large

cmake/analysis/RunIWYU.cmake
```diff
@@ -0,0 +1,69 @@
+# IWYU = "Include What You Use"
+# https://include-what-you-use.org/
```
Jarred-Sumner, line 2: are we using this?

cmake/analysis/RunCppLint.cmake
```diff
@@ -0,0 +1,24 @@
+include(Macros)
```
Jarred-Sumner, line 1: we aren't using cpp lint right? just clang-tidy

cpplint looked deprecated when i last checked

Issue #2187, opened 2023-02-25
Match Node's `events.on` (`node:events`) implementation


ThatOneBro:
Currently our implementation is copy-pasta of `EventEmitter.on`, which seems logical. However, calling `events.on` on an existing `EventEmitter` is supposed to return an async iterator which can be iterated as events come in

I've already started an implementation for this, but we should really get this in as it fixes the async iterator for `node:readline`

paperclover:
since then, events was rewritten in js but it still never included events.on so this is still an issue.

closing, see #2679 (the only reason i'm closing this one instead of the newer one, is that the `NotImplementedError` thrown in this function at runtime points to the newer issue; my bad for not seeing this one initially)

PR #13460, opened 2024-08-22
Fix crash in buffer found in node tests

Jarred-Sumner:
### What does this PR do?

Fix crash in buffer found in node tests

The bug was in calling the constructor as a function. We cannot look at the `newTarget` of a JSC::CallFrame unless it is a construct and not a call.

Tests


robobun:
@Jarred-Sumner, your commit f3ccf5c7db4984b2a10ea87378153926047da4be is building: [`#2187`](https://buildkite.com/bun/bun/builds/2187)

dylan-conway (approved):

PR #2191, opened 2023-02-25
Bundler uses incorrect port when dev server is listening on non-default port #1885

xjmdoo:
When starting the dev server,  there is a logic bug when the default / user provided port is already in use and it tries to pick a new random port. In this case although, the server will listen on the new correct port, the bundler's origin is not updated correctly.

PR #13463, opened 2024-08-22
Fix expect.assertions() and done callback

Jarred-Sumner:
### What does this PR do?

`expect.assertions()` and `expect.hasAssertions()` was only working properly in synchronous code, and we had no tests for non-synchronous code.

`done` callback in a test would hang the test runner forever in cases where that was not strictly necessary.

Tests


robobun:
@Jarred-Sumner, your commit 5cee7b1a7607f560e246f84ef3853bd872c3f9af is building: [`#2191`](https://buildkite.com/bun/bun/builds/2191)

src/bun.js/test/jest.zig
```diff
@@ -1459,20 +1459,33 @@ pub const TestRunnerTask = struct {
     };
 
     pub fn handleResult(this: *TestRunnerTask, result: Result, from: ResultType) void {
+        var result_copy = result;
+        this.handleResultPtr(&result_copy, from);
```
dylan-conway, line 1463: Is it important some calls to `handleResult` do not modify `result`? If not, I think `handleResultPtr` should be deleted and `handleResult` should take a pointer

Issue #10850, opened 2024-05-05
test runner runs all tests above the one which is marked as .only


Zombobot1:
### What version of Bun is running?

1.1.7+b0b7db5c0

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

```ts
// test.ts
import { describe, it, expect } from 'bun:test'

describe('describe', () => { // without describe it doesn't work as expected either
  // same for test instead of it
  it('test 1', () => {
    expect(1).toBe(1) // also runs
  })

  it.only('test 2', () => {
    expect(2).toBe(2)
  })
})
```

`bun test ./test.ts`

```
✓ describe > test 1 [0.03ms]
✓ describe > test 2 [0.01ms]
```

### What is the expected behavior?

test runner doesn't run tests which are not marked as only (to match jest and vitest behavior)

### What do you see instead?

test runner runs all tests above the one which is marked as only (but tests below do not run 🫤)

ilbertt:
The same bug is happening to me.

Bun revision: `1.1.12+43f0913c3`
Computer platform: `Darwin 23.5.0 arm64 arm`

ilbertt:
@Zombobot1 I think the solution is to use the `--only` flag: https://github.com/oven-sh/bun/issues/5893#issuecomment-1731810341

EvHaus:
Looks like this is fixed for me in bun v1.1.21

Jarred-Sumner:
This will be fixed in Bun v1.1.26

Issue #9495, opened 2024-03-18
bun test --watch --only has unexpected behavior


Zombobot1:
### What version of Bun is running?

1.0.32

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

Technically this is not a bug but it breaks workflows which were possible with jest and vitest. This is a serious argument in my team against migrating from jest and vitest to bun because it will make development slower.

For example, during development the tests are run in watch mode and it's possible to isolate specific tests using `.only`. When the isolated tests are fixed it is easy to re-run everything by simply removing `.only`. However `bun test --watch --only` in this case runs nothing.

### What is the expected behavior?

There is no need to use `--only` with vitest or jest. Bun makes DX much worse by introducing this flag. Probably in watch mode it should not have any effect.

Jarred-Sumner:
Yeah the —only flag is bad

Issue #7701, opened 2023-12-16
`.only` tests unexpected behavior when running `bun test` without -only flag


bhflm:
### What version of Bun is running?

1.0.15+b3bdf22eb

### What platform is your computer?

Darwin 22.5.0 x86_64 i386

### What steps can reproduce the bug?

Within a file that has more than 3 tests :

- Place the .only on the 2nd or 3rd test
- Run `bun test`
- Watch tests above the .only get executed, until it reaches the `.only` set one. 

![testbun](https://github.com/oven-sh/bun/assets/23338712/49b702a0-df03-4a00-9253-e0b251a59c8f)

### What is the expected behavior?

It should run all of the tests if no flag `-only` is explicitly set. 

### What do you see instead?

Tests being run sequentially from top to bottom until it reaches one test set as `.only`.

### Additional information

I've reproduced the issue on [this repository](https://github.com/bhflm/bun-bug-test-only/tree/master)

mangs:
Thank you for filing this. I've noticed this too. Tests from different files execute as well.

EvHaus:
Some subjective feedback. I find Bun's behavior here very frustrating to use. The main benefit of using `.only()` for me is to quickly isolate tests while in watch mode. There are currently two ways of running tests in Bun, and both are painful to use. You can:

1. Run `bun test --watch`. But when you focus a test with `.only()` it runs all tests above it too (like this bug describes), or you can
2. Run `bun test --only --watch`, which fixes the issue above but now when you remove all `.only()` tests, Bun won't run anything at all.

I would prefer Bun copied the behavior that Jest and Vitest follow, which is:

- No need for an extra CLI `--only` option. Remove it.
- When running `bun test`, if any `.only()` tests are detected, only run those. If none are detected, run everything

Issue #2194, opened 2023-02-26
Segmentation Fault with `bun i`


u9g:
### What version of Bun is running?

0.5.8

### What platform is your computer?

Linux 5.17.0-1021-oem x86_64 x86_64

### What steps can reproduce the bug?

1. clone https://github.com/eugenehp/icalts
2. bun i
3. see core dumped

### What is the expected behavior?

Not to core dump

### What do you see instead?

```
$ bun i
Segmentation fault (core dumped)
```

PR #13465, opened 2024-08-22
You shouldn't need `--only` to only run `test.only` tests

Jarred-Sumner:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/10850
Fixes https://github.com/oven-sh/bun/issues/7701
Fixes https://github.com/oven-sh/bun/issues/9495

Currently, for the below test to pass, you need to start bun with `--only`:
```sh
bun test --only
```

This is a bad product decision. You shouldn't have to pass `--only`. It should just detect you called `.only` and skip the tests that didn't.

```ts
test("this one should not run #1", () => {
  expect.unreachable();
});

test.only("only this one should run", () => {
  expect().pass();
});

test("this one should not run #2", () => {
  expect.unreachable();
});
```

Originally, it was because we didn't enqueue each test and instead ran them immediately, but it was impossible to be compatible with Jest and work that way so we stopped doing that a long time ago (before the v0.1 public release)

This also resets the state of `test.only` at the beginning of each module. That means having one `test.only` in a later file no longer causes all subsequent files to be `.only`. To make all files run with `.only`, you can prefix `--only`. Since the filesystem sorting order was never deterministic, this shouldn't be a breaking change.

Tests


robobun:
@Jarred-Sumner, your commit 5e7cf676a28c3c6588b92aa81f9700a51845d0c8 is building: [`#2194`](https://buildkite.com/bun/bun/builds/2194)

test/js/bun/test/only-fixture-1.ts
```diff
@@ -0,0 +1,11 @@
+test.only("only this one should run", () => {
+  expect().pass();
```
Electroid, line 2: This doesn't actually verify the code ran, it would be better to `console.log` and check the output.

dylan-conway (changes requested):
The new tests are failing on linux and macos

test/js/bun/test/only-fixture-1.ts
Jarred-Sumner: We test the number of tests ran by the output

Jarred-Sumner:
merging per IRL discussion

PR #13481, opened 2024-08-23
docs: updated `sourcemap` option to reflect the real options

iamkoby:
it had the option of "linked" which isn't available any more. options are
 --sourcemap                     Build with sourcemaps - 'inline', 'external', or 'none'

### What does this PR do?

updated docs that had a deprecated option

this is just docs


Jarred-Sumner:
`linked` is not deprecated? Where did you see that?

iamkoby:
ran the cli with `linked` failed. then ran the cli `help` function and it didn't list `linked`

PR #13488, opened 2024-08-23
Expose `maxLifetime` from uWebSockets for controlling maximum number of minutes a ws connection stays open

Jarred-Sumner:
### What does this PR do?

This lets you always disconnect a ServerWebSocket connection after some amount of time without needing to roll your own `setTimeout` code on each websocket connection to do this

Needs some kind of a test before we can merge this

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -2215,6 +2215,17 @@ declare module "bun" {
     idleTimeout?: number;
 
     /**
+     * Total minutes of activity before a client is automatically disconnected.
+     *
+     * `0` disables this feature.
+     * Must be a number between 0 and 240.
+     *
+     * @default 0
+     * @since Bun v1.1.26
+     */
+    maxLifetime?: number;
+
+    /**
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit c4332419f6fe34b21f00dea2fbf80a6c680c25a9 is building: [`#2284`](https://buildkite.com/bun/bun/builds/2284)

packages/bun-types/bun.d.ts
```diff
@@ -2214,6 +2214,17 @@ declare module "bun" {
...
+    /**
+     * Total minutes of activity before a client is automatically disconnected.
+     *
+     * `0` disables this feature.
+     * Must be a number between 0 and 240.
+     *
+     * @default 0
+     * @since Bun v1.1.26
+     */
+    maxLifetime?: number;
```
Electroid, line 2226: Minutes is obscure, can we do seconds or millis?

pnodet: I believe this is the current uWebSockets behavior.

https://unetworking.github.io/uWebSockets.js/generated/interfaces/WebSocketBehavior.html#maxLifetime

PR #13500, opened 2024-08-24
fix fs.test.ts

nektro:
```
2410 |       expect(maxFD).toBe(newMaxFD); // assert we do not leak file descriptors
                           ^
error: expect(received).toBe(expected)
 
Expected: 695
Received: 696
 
      at /var/lib/buildkite-agent/builds/ip-172-31-25-245-yjkoZT/bun/bun/test/js/node/fs/fs.test.ts:2410:21
✗ fs/promises > readdir(path, {recursive: true} should work x 100 [785.72ms]
```

~~this error was reproducible locally and this fixes it testing locally as well. previously we were measuring the fd count after the warmup instead of at the beginning of the test. due to this some of the warmup objects may not have cleaned themself up yet and introducec flakyness depending on gc aggressiveness~~

that diagnosis was wrong

### DIFF
--- a/test/js/node/fs/fs.test.ts
+++ b/test/js/node/fs/fs.test.ts
@@ -2381,11 +2381,11 @@ describe("fs/promises", () => {
     const full = resolve(import.meta.dir, "../");
 
     const doIt = async () => {
+      const maxFD = getMaxFD();
       for (let i = 0; i < warmup; i++) {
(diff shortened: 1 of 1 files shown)


robobun:
@nektro, your commit 978443fa41586b7250fa56979f3fac262d55e221 is building: [`#2315`](https://buildkite.com/bun/bun/builds/2315)

PR #13505, opened 2024-08-24
Added rounded borders for `bun outdated` table

okineadev:
### What does this PR do?

Makes rounded corners in the table

- [x] Code changes

### Description

In my opinion, the rounded corners are more pleasing to the eye and more modern

### DIFF
--- a/src/cli/outdated_command.zig
+++ b/src/cli/outdated_command.zig
@@ -24,20 +24,20 @@ fn Table(
         column_inside_lengths: [num_columns]usize,
 
         pub fn topLeftSep(_: *const @This()) string {
-            return if (enable_ansi_colors) "┌" else "|";
+            return if (enable_ansi_colors) "╭" else "|";
         }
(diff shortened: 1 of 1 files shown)


xhyrom (approved):

okineadev:
@Jarred-Sumner please review

evanwashere (changes requested):
This further restricts what fonts and text renderers can be used to see/hear/describe bun output. I would only consider this change if it required an additional flag/environment variable that the user had to set manually.

And some fonts don't have equal border width or layout for rounded corners compared to the current character set in use.
image

okineadev:
> This further restricts what fonts and text renderers can be used to see/hear/describe bun output. I would only consider this change if it required an additional flag/environment variable that the user had to set manually.
> 
> And some fonts don't have equal border width or layout for rounded corners compared to the current character set in use. image

Making a separate flag or environment variable to get rounded corners is probably too much

Issue #2329, opened 2023-03-07
Remove remaining usages of JSC C API classes


Jarred-Sumner:
We should migrate the C API classes to the generated bindings code to make the rest of Bun more maintainable, faster to compile (`JSC.NewClass` is around 10% of compile time I think) and faster at runtime. 

The following global classes use JavaScriptCore's C API (and there are more in `HTMLRewriter`)
https://github.com/oven-sh/bun/blob/f527cbabeb7c7586fb3d1ff493e5dc6a6078ce62/src/bun.js/javascript.zig#L91-L110

This is an example of a JSC C API function:

https://github.com/oven-sh/bun/blob/f527cbabeb7c7586fb3d1ff493e5dc6a6078ce62/src/bun.js/api/bun.zig#L288-L298

This is an example of a JSC C++ API function (using our bindings):

https://github.com/oven-sh/bun/blob/f527cbabeb7c7586fb3d1ff493e5dc6a6078ce62/src/bun.js/test/jest.zig#L530-L534

Jarred-Sumner:
completed long ago

PR #13518, opened 2024-08-25
Introduce a way to filter tests to run by byte range in file

Jarred-Sumner:
### What does this PR do?

This introduces a way to precisely filter tests to run by source byte range in files at the call site of the `test` or `describe` function.

Unlike `--test-filter`, this is intended for tooling to allow running tests located at a specific position in a file without worrying about running unrelated tests with an identical label.

We use byte ranges instead of line-and-column because it's cheaper to compute a byte range than a line:column pair, and this is intended for tooling rather than humans so being more difficult for a human to type doesn't matter as much.

The syntax is:
```ts
bun test ${file}::${start}:${length-relative-to-start}
```

Multiple can be specified:
```ts
bun test ./file.test.ts::123:16::456::18 ./file.test.ts::789:10
```

To make this work, we introduce two new AST node's to bun's JS parser:
- `location_identifier`
- `location_dot`

These hint to bun's JS parser to append the source byte ranges to the call site of the test or describe functions at print time. We could rely on sourcemaps to do this instead, but we want it to be 100% accurate all the time, even if it costs a little bit more at parse time.

Also, `test.only` is ignored when byte ranges are passed. This is intended to make in-editor UI for only running specific tests work as expected.

There are tests.


robobun:
@Jarred-Sumner, your commit ad638a6beac7cf337dbd0037f2a744644284c97c is building: [`#2329`](https://buildkite.com/bun/bun/builds/2329)

Issue #11756, opened 2024-06-10
`requestIP` returns `null` after async call


adhusson:
### What version of Bun is running?

1.1.12+43f0913c3

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

```javascript
Bun.serve({
  port: 3001,
  async fetch(req,server) {
    console.log(server.requestIP(req));
    await req.formData();
    console.log(server.requestIP(req));
  }
});
```

Then `curl --data "k=v" http://127.0.0.1:3001`

### What is the expected behavior?

```
{
  address: "::ffff:127.0.0.1",
  family: "IPv6",
  port: 54636,
}
{
  address: "::ffff:127.0.0.1",
  family: "IPv6",
  port: 54636,
}
```

### What do you see instead?

```
{
  address: "::ffff:127.0.0.1",
  family: "IPv6",
  port: 54728,
}
null
```

kevinmingtarja:
I'm able to reproduce this too on Bun `1.1.22`, and `Darwin 21.4.0 arm64`.

luksch42:
same with 1.1.26

KenjiGinjo:
🥺

PR #13532, opened 2024-08-25
fix(Server) handle `requestIP` after async call

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/11756

- [X] Code changes

Existing tests + new test

### DIFF
--- a/src/bun.js/Weak.zig
+++ b/src/bun.js/Weak.zig
@@ -4,6 +4,8 @@ const JSC = bun.JSC;
 pub const WeakRefType = enum(u32) {
     None = 0,
     FetchResponse = 1,
+    PostgreSQLQueryClient = 2,
+    HTTPRequest = 3,
 };
 const WeakImpl = opaque {
     pub fn init(globalThis: *JSC.JSGlobalObject, value: JSC.JSValue, refType: WeakRefType, ctx: ?*anyopaque) *WeakImpl {
--- a/src/bun.js/Weak.zig
+++ b/src/bun.js/Weak.zig
@@ -5,7 +5,6 @@ pub const WeakRefType = enum(u32) {
     None = 0,
     FetchResponse = 1,
     PostgreSQLQueryClient = 2,
-    HTTPRequest = 3,
 };
 const WeakImpl = opaque {
     pub fn init(globalThis: *JSC.JSGlobalObject, value: JSC.JSValue, refType: WeakRefType, ctx: ?*anyopaque) *WeakImpl {
--- a/src/bun.js/webcore/request.zig
+++ b/src/bun.js/webcore/request.zig
@@ -79,9 +79,9 @@ pub const Request = struct {
     pub const getBlobWithoutCallFrame = RequestMixin.getBlobWithoutCallFrame;
     pub const WeakRef = struct {
         value: ?*Request = null,
-        pub fn create(this: *Request) WeakRef {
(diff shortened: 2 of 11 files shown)


src/bun.js/Weak.zig
```diff
@@ -4,6 +4,8 @@ const JSC = bun.JSC;
 pub const WeakRefType = enum(u32) {
     None = 0,
     FetchResponse = 1,
+    PostgreSQLQueryClient = 2,
+    HTTPRequest = 3,
```
Jarred-Sumner: I don't think we can use weak

it's very slow

cirospaciari: ~~I see will replace with strong~~  will replace with another aproach dont wanna to keep the JS alive more than it should

cirospaciari: Replaced with native WeakRef so JS side does not need to be alive longer

Jarred-Sumner:
- I expect the test failures are new
- I don't think it's safe to continue using `Request` after finalize is called. Accessing the body is a UAF which would be very easy to do.

src/bun.js/webcore/request.zig
```diff
@@ -59,8 +59,8 @@ pub const Request = struct {
     method: Method = Method.GET,
     request_context: JSC.API.AnyRequestContext = JSC.API.AnyRequestContext.Null,
     https: bool = false,
-    upgrader: ?*anyopaque = null,
-
+    finalized: bool = false,
```
Jarred-Sumner: What makes it weak vs strong? The reference count is keeping the Request from being deinitialized

Couldn’t we do the same thing except clearer using bun.NewRefCounted?

src/bun.js/webcore/request.zig
```diff
@@ -77,6 +77,33 @@ pub const Request = struct {
     pub const getBlob = RequestMixin.getBlob;
     pub const getFormData = RequestMixin.getFormData;
     pub const getBlobWithoutCallFrame = RequestMixin.getBlobWithoutCallFrame;
+    pub const WeakRef = struct {
```
Jarred-Sumner: This should probably be called Optional or something like that 

And then it would still increment the reference count

src/bun.js/webcore/request.zig
cirospaciari: Weak values are not usable after being finalized and can be partially freed (everything just not the main pointer), so this uses less memory overall and avoid ref-counting cycles for example, same technique is used by Rust/C++. BunRefCounted is used for strong ref not weak the principle is different, would keep more memory alive.

Normally implementations for ref counting uses 2 counts and check both to look if is finalized instead of the finalized flag, but the principle is the same. Also, in C++ you have the option to put it in a separated memory block if we want but it costs one extra allocation, recommended normally is `std::make_shared` with dont create this extra allocation, Rust keep it in the same block like I did or like make_shared.  

Implementation notes from C++ https://en.cppreference.com/w/cpp/memory/shared_ptr#Implementation_notes
Rust https://github.com/rust-lang/rust/blob/0d634185dfddefe09047881175f35c65d68dcff1/library/alloc/src/rc.rs#L285-L289

src/bun.js/webcore/request.zig
cirospaciari: Optional would make it confusing since it's a WeakRef not just something nullable.

Issue #2367, opened 2023-03-11
`Body.json()` does not throw when body is empty


Electroid:
```ts
import { test, expect } from "bun:test";

test("should not be able to parse json from empty body", () => {
  expect(async () => await new Response().json()).toThrow(SyntaxError);
  expect(async () => await new Request("http://example.com/").json()).toThrow(SyntaxError);
});
```

```console
error: expect(received).toThrow(expected)

Expected constructor: SyntaxError
Received function did not throw
```

RuyiLi:
Isn't this the expected result? Looking at the [MDN docs for the response object](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response), there doesn't seem to be a copy constructor. Unless this is intended to be a Bun-specific feature/there's something I missed 😅

Jarred-Sumner:
Fixed recently

PR #13540, opened 2024-08-26
Introduce `static` option in Bun.serve()

Jarred-Sumner:
### What does this PR do?

NOTE: this might not ship. Need to prove it is a meaningful performance improvement first.

This introduces a `static` option in Bun.serve() which will serve `Response` objects from input routes without entering in to JavaScript.

```ts
import { serve } from "bun";

serve({
  static: {
    "/api/health-check": new Response("🟢"),
    "/old-link": Response.redirect("/new-link", 301),
    "/api/version": Response.json(
      {
        app: require("./package.json").version,
        bun: Bun.version,
      },
      {
        headers: {
          "X-Powered-By": "bun",
        },
      }
    ),
  },

  async fetch(req) {
    return new Response("Dynamic!");
  },
});
```

The main goal here is as a performance optimization. But it might make some things a bit simpler too.

It doesn't yet support Bun.file(path) in these `Response` objects. That probably needds to be fixed before shipping, since static file serving is probably the first thing people think of when they see "static".

I want to benchmark this before we move much further with this.


robobun:
@Jarred-Sumner, your commit cff98df6c6785752be29f082fcdf6ec60f9cf6a9 is building: [`#2367`](https://buildkite.com/bun/bun/builds/2367)

zoto-ff:
what about serving static directory? 
maybe smth like `'/assets/*': serve.static(pathToDir)`?

src/deps/uws.zig
```diff
@@ -1894,6 +1894,153 @@ pub const SocketAddress = struct {
     is_ipv6: bool,
 };
 
+pub const AnyResponse = union(enum) {
+    SSL: *NewApp(true).Response,
+    TCP: *NewApp(false).Response,
```
cirospaciari, line 1899: would be good to add a .detached option here, and detach after we call end/close/abort to avoid UAF, server.zig actually just set the resp/req to null, We should detach in the same places, even if is less likely to be a Issue in static.

src/bun.js/api/server.zig
```diff
@@ -119,6 +119,318 @@ const BlobFileContentResult = struct {
...
+
+            const blob: AnyBlob = brk: {
+                switch (response.body.value) {
+                    .Used => {
+                        globalThis.throwInvalidArguments("Response body has already been used", .{});
+                        return null;
+                    },
+
+                    else => {
+                        globalThis.throwInvalidArguments("Body must be fully buffered before it can be used in a static route. Consider calling new Response(await response.blob()) to buffer the body.", .{});
```
cirospaciari, line 217: Should we lazy load this in the first call?

src/deps/uws.zig
Jarred-Sumner: There is no risk of UAF because it doesn't keep a reference to the Response pointer outside the stack

Issue #2386, opened 2023-03-14
fetch() failed


Powma:
### What version of Bun is running?

0.5.5

### What platform is your computer?

Linux 5.10.0-20-rt-amd64 x86_64 unknown

### What steps can reproduce the bug?

```typescript
fetch(
	'https://foo.toto:9443',
	{
		method: 'GET',
		headers: {
			'Content-Type': 'application/json',
			Authorization: `Bearer xxxxxxx`,
		},
	}
)
	.then((res: any) => res.json())
	.catch((err) => {
		console.error(err);
	});
```
output : 
```
ConnectionClosed: fetch() failed
 path: "https://foo.toto:9443"
```
 
With curl : 
```
curl -vvv --location 'https://foo.toto' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer XXXXXXX'
*   Trying 10.0.202.96:8443...
* Connected to foo.toto(10.0.0.0) port 9443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: O=XXXX; CN=foo.titi; emailAddress=foo@foo.com
*  start date: Jun 15 12:40:46 2021 GMT
*  expire date: Jun 16 12:40:46 2023 GMT
*  subjectAltName: host "foo.toto" matched cert's "foo.toto"
*  issuer: C=XX; O=XXX; OU=XXXXX; CN=XXXXX
*  SSL certificate verify ok.
> GET / HTTP/1.1
> Host: foo.toto:9443
[...]

PR #13557, opened 2024-08-27
add `--filter` and package pattern arguments to `bun outdated`

dylan-conway:
### What does this PR do?
- Adds workspace filtering to `bun outdated` through `--filter`. You can glob for workspace package names or workspace paths. Examples:
```bash
# Path patterns:
# Matches all workspaces in `./packages`
bun outdated --filter="./packages/*"

# Name patterns:
# All workspaces with names starting with `pkg`
bun outdated --filter="pkg*"

# All workspaces excluding `pkg1`
bun outdated --filter="*" --filter="!pkg1"
```
- Adds dependency name filtering. Works the same as workspace name patterns, and can be used in combination with `--filter`. Example:
```bash
# All dependencies starting with `is-` (could display `is-even`, `is-odd`, ...)
bun outdated "is-*"

# With `--filter`. Matches all outdated `jquery` dependencies in all workspaces.
bun outdated jquery --filter="*"
```
- fixes a bug that would occur when printing versions with tags in the `bun outdated` table

Added tests for `--filter` and args


robobun:
@dylan-conway, your commit fcbc645ffe9da0bcbc572a0718fb051cd0aaf9fc is building: [`#2386`](https://buildkite.com/bun/bun/builds/2386)

src/cli/outdated_command.zig
```diff
@@ -180,92 +186,282 @@ pub const OutdatedCommand = struct {
...
+
+            const patterns_buf = bun.default_allocator.alloc([]const u32, args.len) catch bun.outOfMemory();
+            for (args, patterns_buf) |arg, *converted| {
+                if (arg.len == 0) {
+                    converted.* = &.{};
+                    continue;
+                }
+
+                const length = bun.simdutf.length.utf32.from.utf8.le(arg);
+                const convert_buf = bun.default_allocator.alloc(u32, length) catch bun.outOfMemory();
```
Jarred-Sumner, line 338: seems crazy to me that we have to drop down to simdutf here

src/cli/outdated_command.zig
```diff
@@ -180,92 +186,282 @@ pub const OutdatedCommand = struct {
...
+                        if (filter.len == 0) continue;
+                        const res = pkg_resolutions[workspace_pkg_id];
+
+                        const res_path = switch (res.tag) {
+                            .workspace => res.value.workspace.slice(string_buf),
+                            .root => FileSystem.instance.top_level_dir,
+                            else => unreachable,
+                        };
+
+                        const abs_res_path = path.joinAbsString(FileSystem.instance.top_level_dir, &[_]string{res_path}, .posix);
```
Jarred-Sumner, line 289: probably should use joinAbsStringBuf incase something else is making use of the shared string buffer?

src/cli/outdated_command.zig
```diff
@@ -180,92 +186,282 @@ pub const OutdatedCommand = struct {
...
+        defer {
+            for (converted_filters) |converted| {
+                const filter, _ = converted;
+                allocator.free(filter);
+            }
+            allocator.free(converted_filters);
+        }
+
+        // move all matched workspaces to front of array
+        var i: usize = 0;
```
Jarred-Sumner, line 271:
```suggestion
        var i: usize = 0;
        var abs_res_path_buf: bun.PathBuffer = undefined;
```

src/cli/outdated_command.zig
Jarred-Sumner, line 289:
```suggestion
                        const abs_res_path = path.joinAbsStringBuf(&abs_res_path_buf, FileSystem.instance.top_level_dir, &[_]string{res_path}, .posix);
```

Jarred-Sumner:
This is mergeable so I'm going to merge it

I do think that we should have some canonical way to "get a validated UTF-8 string from bytes" that doesn't directly involve simdutf

PR #2412, opened 2023-03-17
feat(tls.Server) basic support (cert, key, rejectUnauthorized, requestCert, ca)

cirospaciari:
This depends on https://github.com/Jarred-Sumner/uSockets/pull/1
To run the tests, rebuild usockets with the PR source is needed.

Closes https://github.com/oven-sh/bun/issues/2504

PR #13575, opened 2024-08-27
fix #13563 and #12440

### LINKED ISSUE #12440: error: Failed to link vite: EBUSY

 v1.1.18 (5a0b9352)

windows 11

when vite app is running then do bun update in another cmd , i got Failed to link vite: busy, i should be able to update/add packages during vite dev 

 should be able to update/add packages during vite dev 

1. bun dev in my svelte project
2. bun update/ bun install

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #13563: Bun reinstalls dependencies with `11.0.0-rc.485` style semver on every install

1.1.26+0a37423ba

Darwin 23.5.0 arm64 arm

Add these dependencies to your `package.json`.

```
{
  "name": "bun-test",
  "module": "index.ts",
  "type": "module",
  "dependencies": {
    "@trpc/client": "11.0.0-rc.485",
    "@trpc/react-query": "11.0.0-rc.485",
    "@trpc/server": "11.0.0-rc.485"
  },
  "devDependencies": {
    "@types/bun": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  }
}

```

Run `bun i`. Notice that it will install the dependencies.

Run `bun i` again. Notice it will install them again.

It should only install once and then when running `bun i` again it will not re-install them.

### What do you see instead?

It continuously re-installs those 3 dependencies.

```
~/projects/bun-test
❯ bun i
bun install v1.1.26 (0a37423b)

+ @trpc/client@11.0.0-rc.485
+ @trpc/react-query@11.0.0-rc.485
+ @trpc/server@11.0.0-rc.485

10 packages installed [237.00ms]

~/projects/bun-test
❯ bun i
bun install v1.1.26 (0a37423b)

+ @trpc/client@11.0.0-rc.485
+ @trpc/react-query@11.0.0-rc.485
+ @trpc/server@11.0.0-rc.485

3 packages installed [31.00ms]

~/projects/bun-test
❯ bun i
bun install v1.1.26 (0a37423b)

+ @trpc/client@11.0.0-rc.485
+ @trpc/react-query@11.0.0-rc.485
+ @trpc/server@11.0.0-rc.485

3 packages installed [29.00ms]
```

### Additional information

_No response_

dylan-conway:
### What does this PR do?
fixes #13563 
fixes #12440

### DIFF
--- a/src/install/bin.zig
+++ b/src/install/bin.zig
@@ -483,6 +483,11 @@ pub const Bin = extern struct {
             const abs_exe_file: [:0]const u16 = dest_buf[0 .. abs_dest_w.len + ".exe".len :0];
 
             bun.sys.File.writeFile(bun.invalid_fd, abs_exe_file, WinBinLinkingShim.embedded_executable_data).unwrap() catch |err| {
+                if (err == error.EBUSY) {
+                    // exe is most likely running. bunx file has already been updated, ignore error
+                    return;
+                }
+
                 this.err = err;
                 return;
             };
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -1207,8 +1207,13 @@ pub fn NewPackageInstall(comptime kind: PkgInstallKind) type {
             if (!package_json_checker.has_found_version and resolution_tag != .workspace) return false;
 
             const found_version = package_json_checker.found_version;
+
+            // exclude build tags from comparsion
+            // https://github.com/oven-sh/bun/issues/13563
+            const found_version_end = strings.lastIndexOfChar(found_version, '+') orelse found_version.len;
+            const expected_version_end = strings.lastIndexOfChar(this.package_version, '+') orelse this.package_version.len;
             // Check if the version matches
-            if (!strings.eql(found_version, this.package_version)) {
+            if (!strings.eql(found_version[0..found_version_end], this.package_version[0..expected_version_end])) {
                 const offset = brk: {
                     // ASCII only.
                     for (0..found_version.len) |c| {


robobun:
@dylan-conway, your commit ea7cf06d1aeb840b200648e5c6b8894a61eb2fb9 is building: [`#2412`](https://buildkite.com/bun/bun/builds/2412)

src/install/bin.zig
```diff
@@ -483,6 +483,11 @@ pub const Bin = extern struct {
             const abs_exe_file: [:0]const u16 = dest_buf[0 .. abs_dest_w.len + ".exe".len :0];
 
             bun.sys.File.writeFile(bun.invalid_fd, abs_exe_file, WinBinLinkingShim.embedded_executable_data).unwrap() catch |err| {
+                if (err == error.EBUSY) {
```
Jarred-Sumner, line 486: @paperdave does this look right

src/install/bin.zig
```diff
@@ -483,6 +483,11 @@ pub const Bin = extern struct {
...
+                    // exe is most likely running. bunx file has already been updated, ignore error
```
paperclover, line 487: this will break in the edge case exe file is an older copy (which would be the primary case that this code path, the other being the package updated to move the script), in which executing the old binary will reject the bunx file.

the solution is to, in addition to handling this error, write the bunx file *after* the exe has been successfully written.

edit: this case i mentioned is actually not the common case, as i thought that we don't overwrite the shim if it is already valid; we always overwrite it. this would mean updating any package while the bin is running will cause this, even if it's identical.

Issue #12786, opened 2024-07-25
SSH2 throwing weird error since upgrading to Bun v.1.1.20


Kulimeya:
### What version of Bun is running?

1.1.20

### What platform is your computer?

Linux 6.1.0-15-amd64 x86_64 unknown

### What steps can reproduce the bug?

`import { Client } from "ssh2";

  // SSH connection configuration
  const sshConfig = {
    host: "",
    port: 22, // Default SSH port is 22
    username: "",
    password: "",
  };

  const remotePath = ${rootDir}/wp-content/plugins;
  const downloadUrl = "https://"; 
  const rootDir = "";


const conn = new Client();

  conn
    .on("ready", () => {
      console.log("Client :: ready");

      // Download the file using wget
      const wgetCommand = wget -P ${remotePath} ${downloadUrl};
      conn.exec(wgetCommand, (err: any, stream: any) => {
        if (err) throw err;

        stream
          .on("close", (code: any, signal: any) => {
            if (code === 0) {
              console.log("File downloaded successfully.");

              // Extract the downloaded file
              const fileName = path.basename(downloadUrl);
              const extractCommand = unzip -o ${remotePath}/${fileName} -d ${remotePath} && rm ${remotePath}/${fileName};
              conn.exec(extractCommand, (err: any, stream: any) => {
                if (err) throw err;

                stream
                  .on("close", (code: any, signal: any) => {
                    if (code === 0) {
                      console.log("File extracted and cleaned up successfully.");
                    } else {
                      console.log(`Extraction failed with code: ${code}`);
                    }
                    conn.end();
                  })
                  .on("data", (data: any) => {
                    console.log("STDOUT: " + data);
                  })
                  .stderr.on("data", (data: any) => {
                    console.error("STDERR: " + data);
                  });
              });
            } else {
              console.log(`Download failed with code: ${code}`);
              conn.end();
            }
          })
          .on("data", (data: any) => {
            console.log("STDOUT: " + data);
          })
          .stderr.on("data", (data: any) => {
            console.error("STDERR: " + data);
          });
      });
    })
    .connect(sshConfig);`

### What is the expected behavior?

Should normally connect to the ssh server like it did fine in v1.1.7

### What do you see instead?

337 |             if (!desc) {                                                                                                                                                        
338 |               desc = DISCONNECT_REASON_BY_VALUE[reason];                                                                                                                        
339 |               if (desc === undefined)                                                                                                                                           
[...]

Issue #2442, opened 2023-03-20
Porting a Jest test to Bun fails because of some node issue


nikivdev:
### What version of Bun is running?

0.5.8

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

I have this [repo](https://github.com/nikitavoloboev/isomorphic-git/tree/sparse-checkout).

Inside it has some jest tests, if I run one I get: 

```
~/clones/isomorphic-git sparse-checkout
❯ jest __tests__/test-pull.js
  console.log
    /var/folders/bn/br2v7mf14zb2yqhsyd4xd7nw0000gn/T/jest-fixture-AC7TDe

      at log (src/api/pull.js:95:13)

  console.log
    /var/folders/bn/br2v7mf14zb2yqhsyd4xd7nw0000gn/T/jest-fixture-SN5gFA

      at log (src/api/pull.js:95:13)

  console.log
    /var/folders/bn/br2v7mf14zb2yqhsyd4xd7nw0000gn/T/jest-fixture-utX6f8

      at log (src/api/pull.js:95:13)

 PASS  __tests__/test-pull.js
  pull
    ✓ pull (217 ms)
    ✓ pull fast-forward only (55 ms)
    ✓ pull no fast-forward (50 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        0.939 s, estimated 2 s
Ran all test suites matching /__tests__\/test-pull.js/i.
```

However jest is slow and I want to run it in bun. So I created this file:

https://github.com/nikitavoloboev/isomorphic-git/blob/sparse-checkout/__tests__/pull.test.ts

It should in theory work out of the box however there is now this issue:

```
❯ bun test
bun test v0.5.8 (5d296f62)

pull.test.ts:
SyntaxError: Missing 'default' import in module '/Users/nikiv/clones/isomorphic-git/http/node/index.d.ts'.

 0 pass
 1 fail
Ran 1 tests across 1 files [37.00ms]
```

For some reason, `import http from 'isomorphic-git/http'` is okay in node/jest bun in bun it fails.

Would love to solve this issue and move away from jest.

### What is the expected behavior?

bun test should succeed.

### What do you see instead?

It fails to import when in node it works fine.

Electroid:
Closing since the reproduction repo is no longer active. If you are still running into this issue, please re-open this issue and we'll investigate.

PR #13583, opened 2024-08-28
Fixes #13581

### LINKED ISSUE #13581: buffer.utf8Write problem (within ssh2 module)

1.1.27-canary.14+11d7a9d5e

Linux 6.8.0-1008-raspi aarch64 aarch64

We have identified that the buffer created with the ssh public key has a corruption, which is causing the key exchange to fail on initial server connection.  It appears that the buffer.utf8Write method is being called to pass in 'ssh-rsa' and this fails when running with bun.  Expect that this is an internal nodejs buffer method (as it's not detailed in the nodejs docs) and may just be returning random content.  We do see it defined in JSBuffer.cpp though.
This can be tested with the following:
```
// check buffer.utf8Write functionality

const n = Buffer.from([0,203,137,87,196,186,197,81,205,203,66,27,11,76,39,3,17,222,80,161,79,165,213,221,116,67,17,228,122,247,188,125,142,171,65,44,52,202,112,48,112,200,206,89,186,57,209,175,49,51,180,240,211,128,99,94,93,9,239,199,196,187,160,119,67,116,13,69,82,209,99,147,197,104,103,201,15,27,133,184,54,90,93,212,168,74,3,247,204,94,65,153,99,73,251,3,17,172,95,20,213,142,40,99,39,145,175,209,146,57,210,113,108,149,71,171,67,252,180,254,189,18,221,185,165,119,237,247,173,100,78,3,65,137,165,242,91,99,205,185,123,68,180,65,170,139,1,213,225,69,20,108,39,221,107,105,103,171,7,53,31,227,78,39,10,114,71,126,248,182,62,121,78,206,36,127,50,224,125,173,159,207,61,70,61,38,16,27,118,247,237,214,229,29,53,117,84,237,190,165,237,217,191,29,52,220,225,16,137,218,138,12,83,53,114,85,249,43,233,50,131,164,176,146,230,119,206,142,84,136,197,193,140,75,94,138,128,73,3,180,91,76,40,6,192,15,68,187,222,255,206,68,84,251,163,177,243]);
const e = Buffer.from([1,0,1]);

function writeUInt32BE(buf, value, offset) {
  buf[offset++] = (value >>> 24);
  buf[offset++] = (value >>> 16);
  buf[offset++] = (value >>> 8);
  buf[offset++] = value;
  return offset;
}

function genOpenSSHRSAPub(n, e) {
  const publicKey = Buffer.allocUnsafe(4 + 7 + 4 + e.length + 4 + n.length);

  writeUInt32BE(publicKey, 7, 0);
  publicKey.utf8Write('ssh-rsa', 4, 7);

  let i = 4 + 7;
  writeUInt32BE(publicKey, e.length, i);
  publicKey.set(e, i += 4);

  writeUInt32BE(publicKey, n.length, i += e.length);
  publicKey.set(n, i + 4);

  return publicKey;
}

const publicKey = genOpenSSHRSAPub(n, e);

console.log('publicKey: ' + JSON.stringify(publicKey));
```

The desired behavior is that the phrase 'ssh-rsa' is written into the buffer from position 4 (7 characters long).  This should result in buffer contents of [115,115,104,45,114,115,97] for those 7 characters.  In this specific case, this identifies the required ssh key format.

### What do you see instead?

The buffer contents from position 4 for the 7 characters is [115,115,104,0,0,0,0] (and we have also seen [115,115,104,0,253,0,19]).

### Additional information

Happy to help debug if someone can point me to where in the bun code this gets implemented.

Jarred-Sumner:
### What does this PR do?

Fixes #13581 

In Node.js, Buffer.prototype.write calls into `utf8Write`, `asciiWrite`, `latin1Write`, etc but they do not have many tests that check these methods specifically.

Since most of our tests for Buffer are copied or dervied from Node, our tests also weren't checking these methods. However, our implementation is slightly different enough that this bug crept through.

So now we run the buffer tests twice. Once with the real implementation of Buffer.prototype.write, and once again with the implementation that calls out to the corresponding encoded write function. 

I have verified that previously some tests were failing when using the encoded write function instead of the generic write function. Those tests now pass.

### DIFF
--- a/src/bun.js/bindings/JSBuffer.cpp
+++ b/src/bun.js/bindings/JSBuffer.cpp
@@ -1707,7 +1707,7 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunction_writeEncodingBody(JS
         return {};
     }
 
-    size_t max_length = std::min(max - offset, max);
+    size_t max_length = std::min(length - offset, max);
 
     RELEASE_AND_RETURN(scope, writeToBuffer(lexicalGlobalObject, castedThis, str, offset, max_length, encoding));
 }
--- a/test/js/node/buffer.test.js
+++ b/test/js/node/buffer.test.js
@@ -1,1323 +1,2908 @@
-import { Buffer, SlowBuffer, isAscii, isUtf8 } from "buffer";
-import { describe, it, expect, beforeEach, afterEach } from "bun:test";
+import { Buffer, SlowBuffer, isAscii, isUtf8, kMaxLength } from "buffer";
+import { describe, it, expect, beforeEach, afterEach, beforeAll, afterAll } from "bun:test";
 import { gc } from "harness";
 
 const BufferModule = await import("buffer");
+
 beforeEach(() => gc());
 afterEach(() => gc());
 
-it("#9120 fill", () => {
-  let abBuf = Buffer.alloc(2, "ab");
-  let x = Buffer.alloc(1);
-  x.fill(abBuf);
-  expect(x.toString()).toBe("a");
-
-  for (let count = 2; count < 10; count += 2) {
-    const full = Buffer.from("a".repeat(count) + "b".repeat(count));
-    const x = Buffer.alloc(count);
-    x.fill(full);
-    expect(x.toString()).toBe("a".repeat(count));
+/***
+ *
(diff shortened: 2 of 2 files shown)


robobun:
@Jarred-Sumner, your commit ad6aad809d566751bd86b81b7bbf7b8c78f2bbd1 is building: [`#2442`](https://buildkite.com/bun/bun/builds/2442)

KilianB:
Fixes: #12786

cirospaciari (approved):

PR #13618, opened 2024-08-30
Fix hypothetical race condition in `Bun.build` API

Jarred-Sumner:
### What does this PR do?

Fix hypothetical race condition in `Bun.build` where the `waiter`  is null on first use, causing it to never wake up the thread and wait forever


robobun:
@Jarred-Sumner, your commit acf4bf4a403cff3e46f4b43ae4562feda5bd639e is building: [`#2537`](https://buildkite.com/bun/bun/builds/2537)

paperclover (approved):

PR #13612, opened 2024-08-30
fix overriding `sshCommand` and `askpass` from gitconfig

dylan-conway:
### What does this PR do?
fixes #12172 
fixes #12257 

### How did you verify your code works?
TODO: tests

PR #13620, opened 2024-08-30
Revert "fix overriding `sshCommand` and `askpass` from gitconfig"

Jarred-Sumner:
Reverts oven-sh/bun#13612

### DIFF
--- a/src/install/repository.zig
+++ b/src/install/repository.zig
@@ -16,118 +16,11 @@ const string = @import("../string_types.zig").string;
 const strings = @import("../string_immutable.zig");
 const GitSHA = String;
 const Path = bun.path;
-const File = bun.sys.File;
 
 threadlocal var final_path_buf: bun.PathBuffer = undefined;
 threadlocal var ssh_path_buf: bun.PathBuffer = undefined;
 threadlocal var folder_name_buf: bun.PathBuffer = undefined;
 threadlocal var json_path_buf: bun.PathBuffer = undefined;
-
-const SloppyGlobalGitConfig = struct {
-    has_askpass: bool = false,
-    has_ssh_command: bool = false,
-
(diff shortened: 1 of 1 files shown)


jonblatho:
If I can ask, why was this reverted? I tried the canary build from #13612 and can confirm it partially fixed the issue I was encountering — it seems Bun still doesn't respect my SSH config, so I now have to specify `git@` in the repository URL in `package.json`, but when I do so it no longer attempts to resolve over HTTPS instead of SSH, so the install succeeds. Did that break something else?

Jarred-Sumner:
@jonblatho un-reverted since, please run `bun upgrade --canary`

jonblatho:
@Jarred-Sumner my apologies, looks like my successful install earlier was a fluke or I failed to clear some cache somewhere. Just retried with the canary build I mentioned above (`v1.1.27-canary.38 (9faaa9b9)`) and it looks like there was no change in behavior whatsoever from the original issue — `RepositoryNotFound` when `git@` isn't present in the URL, prompted for HTTPS auth when it is. The same holds true with the latest canary build, `v1.1.27-canary.58 (03de99af)`.

jonblatho:
I guess I should also point out that the above was with no changes to my `.gitconfig` file (which does not have any specific references to our GitLab instance, as it's never been necessary) or SSH config. The only thing that needs to be enforced is that packages are resolved over SSH, not HTTPS, and therefore that the username for the SSH connection is `git`, not my own username.

For what it's worth, I did also attempt again after running `git config --global url.ssh://git@my.gitlab.instance.com/.insteadOf https://git@my.gitlab.instance.com/` to seemingly no avail. However, if any specific configuration changes are needed, please specify and I'm happy to retry.

PR #13633, opened 2024-08-30
fix: lcov non-hit executable lines reporting

fmorency:
### What does this PR do?

This fixes the reporting of non-hit executable lines in `lcov` reports.

Before this fix, the reported code coverage would be 100% because non-hit executable lines were not marked as such.

After this fix, I get the correct code coverage

The HTML reports above were generated using `genhtml` tool from the `lcov` project.

- [x] Code changes

I fixed existing automated tests.

I am not a Zig coder so I could use your help to determine if this PR is okay and how to address the following checklist.


fmorency:
CC @Jarred-Sumner

Jarred-Sumner:
Thanks

PR #2577, opened 2023-04-06
chore: simplify prettier command

privatenumber:

PR #13640, opened 2024-08-31
Use -mmacos-version-min instead of -mmacosx-version-min

Jarred-Sumner:
### What does this PR do?

The flag was renamed to `-mmacos-version-min` and remains an alias, but might have slightly different behavior judging by various issues

https://github.com/llvm/llvm-project/pull/95374/files#diff-7f7ebf48869598cf549ff15b340062c9ed163d9dff436b61d9f26b83038ade86

CI


robobun:
@Jarred-Sumner, your commit ec354afbf22d0428339ee349d9d5b9fee045f600 is building: [`#2577`](https://buildkite.com/bun/bun/builds/2577)

Issue #1466, opened 2022-11-07
AES-GCM fails with empty data


panva:
### What steps can reproduce the bug?

```ts
const name = "AES-GCM";
const key = await crypto.subtle.generateKey({ name, length: 128 }, false, [
  "encrypt",
  "decrypt",
]);
const data = new Uint8Array();
const iv = crypto.getRandomValues(new Uint8Array(16));
const algorithm = { name, iv, tagLength: 128 };

const ciphertext = await crypto.subtle.encrypt(algorithm, key, data);

await crypto.subtle.decrypt(algorithm, key, ciphertext);
```

### What do you see instead?

`OperationError: The operation failed for an operation-specific reason`

Jarred-Sumner:
I think line 96 is returning an error:

https://github.com/oven-sh/bun/blob/be9bab5c297756f3548345b8b9e78e809ddbcfca/src/bun.js/bindings/webcrypto/CryptoAlgorithmAES_GCMOpenSSL.cpp#L95-L98

From `openssl/cipher.h`:

```c
// EVP_EncryptFinal_ex writes at most a block of ciphertext to |out| and sets
// |*out_len| to the number of bytes written. If padding is enabled (the
// default) then standard padding is applied to create the final block. If
// padding is disabled (with |EVP_CIPHER_CTX_set_padding|) then any partial
// block remaining will cause an error. The function returns one on success and
// zero otherwise.
OPENSSL_EXPORT int EVP_EncryptFinal_ex(EVP_CIPHER_CTX *ctx, uint8_t *out,
                                       int *out_len);
```

So I'm guessing that "any partial block remaining" is the problem.

<img width="1896" alt="image" src="https://user-images.githubusercontent.com/709451/200479074-65cf0f2b-8fd9-44a8-8d5e-4c9a6e9b93a0.png">

Jarred-Sumner:
@robobun 

test repro bot

```jsx
const name = "AES-GCM";
const key = await crypto.subtle.generateKey({ name, length: 128 }, false, [
  "encrypt",
  "decrypt",
]);
const data = new Uint8Array();
const iv = crypto.getRandomValues(new Uint8Array(16));
const algorithm = { name, iv, tagLength: 128 };

const ciphertext = await crypto.subtle.encrypt(algorithm, key, data);

console.log(await crypto.subtle.decrypt(algorithm, key, ciphertext));
```

Jarred-Sumner:
@robobun 

test #2

```js
[...]

PR #13646, opened 2024-08-31
Fix AES-GCM encryption of empty messages

wpaulino:
### What does this PR do?

AES-GCM is not block based, unlike AES-CBC, so we're not required to pad the ciphertext. Instead, a tag is written at the end to authenticate the plaintext after decryption. `EVP_EncryptFinal_Ex` within this context is used as a way to signal that we've concluded writing our data and should proceed to compute the authentication tag. In the case of a zero message, this logic would essentially get triggered via the earlier call to `EVP_EncryptUpdate`, resulting in an error when `EVP_EncryptFinal_Ex` is then called because it is not idempotent. To fix, we simply check that we have some plaintext to write before writing it.

Fixes https://github.com/oven-sh/bun/issues/1466.

### DIFF
--- a/src/bun.js/bindings/webcrypto/CryptoAlgorithmAES_GCMOpenSSL.cpp
+++ b/src/bun.js/bindings/webcrypto/CryptoAlgorithmAES_GCMOpenSSL.cpp
@@ -56,7 +56,7 @@ static std::optional<Vector<uint8_t>> cryptEncrypt(const Vector<uint8_t>& key, c
         return std::nullopt;
 
     EvpCipherCtxPtr ctx;
-    int len;
+    int len = 0;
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you for this. Let's let the tests run (there will be some unrelated Windows test failures) and then I'll merge after tests pass

 > It looks like the bug is also present [upstream](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/crypto/openssl/CryptoAlgorithmAESGCMOpenSSL.cpp#L89). Do we try to keep the webcrypto bindings up to date against it?

We occasionally sync their changes into ours, but we haven't done a good job of upstreaming things.

PR #13658, opened 2024-09-01
Add snapshot tests for dependency/version parsing

Jarred-Sumner:
### What does this PR do?

This exposes an internal APi like [`npm-package-arg`](https://www.npmjs.com/package/npm-package-arg) which returns details about a dependency/version that is parsed

I think we could use some more tests for various parsing edgecases like this

It is a test


robobun:
@Jarred-Sumner, your commit 1bec6c3810ad49272e0f7ad385c1dd3efbfd7d19 is building: [`#2636`](https://buildkite.com/bun/bun/builds/2636)

dylan-conway (approved):
Nice, this is useful

Issue #13622, opened 2024-08-30
Bun crash with ref-napi


Altepp:
### How can we reproduce the crash?

Using ref-napi and refType function crash it, havent explored more into the bug

```ts
import ref from 'ref-napi';

const reference= ref.refType(null as any); // Crash
const reference= ref.refType(ref.types.void); // Crash
const reference= ref.refType(ref.types.bool); // Crash

```

### Stack Trace (bun.report)

Bun v1.1.24 ([`85a3299`](<https://github.com/oven-sh/bun/tree/85a32991152871b06becd6b411eb663008369769>)) on windows x86_64 [AutoCommand]

**panic**: napi: napi_create_reference
  Error::Error

- [`output.zig:357`](<https://github.com/oven-sh/bun/blob/85a32991152871b06becd6b411eb663008369769/src/output.zig#L357>): `panic`
- [`bun.zig:1408`](<https://github.com/oven-sh/bun/blob/85a32991152871b06becd6b411eb663008369769/src/bun.zig#L1408>): `lenSliceTo`
- ??? at `0x8bd8` in node.napi.node
- ??? at `0xd5e8` in node.napi.node
- ??? at `0x7503` in node.napi.node
- ??? at `0xbf9d` in node.napi.node
- ??? at `0x810f` in node.napi.node
- ??? at `0x7141` in node.napi.node
- [`napi.cpp:907`](<https://github.com/oven-sh/bun/blob/85a32991152871b06becd6b411eb663008369769/src/bun.js/bindings/napi.cpp#L907>): `napi_module_register`

Altepp:
> @Altepp, the latest version of Bun is v1.1.26, but this crash was reported on Bun v1.1.24.
> 
> Are you able to reproduce this crash on the latest version of Bun?
> 
> ```shell
> bun upgrade
> ```

yes

PR #2642, opened 2023-04-13
import file path and fix disabled browser require

dylan-conway:

PR #13664, opened 2024-09-01
Fix crash when throwing an exception from napi

Jarred-Sumner:
### What does this PR do?

Fix crash when throwing an exception from napi

This fixes the crash in https://github.com/oven-sh/bun/issues/13622, but does not fix the underlying cause of the crash.

After this PR, it throws an exception like the below:
```js
17 | var uv = (process.versions.uv || '').split('.')[0]
18 | 
19 | module.exports = load
20 | 
21 | function load (dir) {
22 |   return runtimeRequire(load.resolve(dir))
               ^
TypeError: Underlying ArrayBuffer has been detached from the view or out-of-bounds
      at load (/Users/jarred/Desktop/node_modules/node-gyp-build/node-gyp-build.js:22:10)
      at /Users/jarred/Desktop/node_modules/ref-napi/lib/ref.js:8:18
      at require (1:11)
      at /Users/jarred/Desktop/ref.js:1:5
```

The next step here is to implement NapiHandleScope and wire it up to the various places where objects are created.

Manual :(

### DIFF
--- a/src/bun.js/bindings/napi.cpp
+++ b/src/bun.js/bindings/napi.cpp
@@ -45,7 +45,6 @@
 #include <JavaScriptCore/JSArrayBuffer.h>
 #include "JSFFIFunction.h"
 #include <JavaScriptCore/JavaScript.h>
-#include <JavaScriptCore/JSWeakValue.h>
 #include "napi.h"
 #include <JavaScriptCore/GetterSetter.h>
 #include <JavaScriptCore/JSSourceCode.h>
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit 3b34d8eba66f0cd7fae28f953e04e433d3a793e1 is building: [`#2642`](https://buildkite.com/bun/bun/builds/2642)

Jarred-Sumner:
I'm pretty confident this is better than status quo, but @190n when you get a chance please review

Issue #13652, opened 2024-08-31
--watch/--hot set file extensions


tizu69:
### What is the problem this feature would solve?

I use Bun with Elysia and Tailwind for an internal project of mine. I use tailwind so that the website it serves doesn't look that sh*t. Whenever I do have to modify the raw .css file tho, I do not get the benefits of --watch, as it seems to only listen for js/ts/jsx/tsx files, not anything else.

### What is the feature you are proposing to solve the problem?

--watch=./\*\*/\*.\*

### What alternatives have you considered?

giving up on the project

PR #13665, opened 2024-09-01
Support hot reloading when .css or any other imported file changes

Jarred-Sumner:
### What does this PR do?

Support hot reloading when .css or any other imported file changes

To make a file eligible for hot reloading, import the file
```ts
import './tailwind.css'

// ... rest of my app
```

Previously, we only supported hot reloading known files like .js, .ts. This makes it work for arbitrary non-virtual files.

Fixes https://github.com/oven-sh/bun/issues/13652

There is a test


robobun:
@Jarred-Sumner, your commit 25b4912c3567282c5f3b142340a1b9317cf6d603 is building: [`#2643`](https://buildkite.com/bun/bun/builds/2643)

paperclover (approved):
thank you for the comments in module_loader.zig

it's a bit weird to see this in `import` instead of some more specific api, such as a cli flag or specific `addWatchFile` function, but i think this should be done anyways.

i had an old project (commit in jan 2024) where i found myself trying to use import to register a file watcher

```ts
	if (typeof Bun !== "undefined") {
		// This allows bun --watch to work.
		// it would be nicer if there was `addWatchFile` or something
		import.meta.require?.(resolve(path));
	}
```

to which it didnt work at the time, now it would (assuming watch works with require)

PR #2652, opened 2023-04-13
use a lazyily initialized stream for `node:crypto` `createHash`

paperclover:
```ts
createHash(algo).update(string).digest(format)
```

Will be over 3x faster, just shy of what our native `Bun.CryptoHasher` does. This is done by lazily initializing the `stream.Transform` only if the user decides to use the `Hash` as a stream, meaning the above case only calls CryptoHasher, giving it that performance boost.
 
image

The exact way this is done is by extending `Transform` but not calling ""`super`"" in the constructor (it's an es5 class so this is allowed). Then there are a ton of getters and if any are triggered, we call the `Transform` constructor on our current instance and replace the prototype.

Addresses #2190 but not the issue it's talking about, but it does solve performance issues for what it's testing for.

PR #13677, opened 2024-09-02
Add missing OOM exception check in Bun.escapeHTML

Jarred-Sumner:
### What does this PR do?

Add missing OOM exception check in Bun.escapeHTML

There were a couple crash reports on Windows related to this

Repro code:
```js
var base = "a".repeat(1024).repeat(1024).repeat(1024);
globalThis.strings = [base];
for (let i = 0; i < 1024; i++) {
  const out = base + i + "😊";
  globalThis.strings.push(out);
  Bun.escapeHTML(out);
}
```

Only reproduces on Windows after about 80 seconds.

### DIFF
--- a/src/bun.js/bindings/BunObject.cpp
+++ b/src/bun.js/bindings/BunObject.cpp
@@ -392,23 +392,24 @@ JSC_DEFINE_HOST_FUNCTION(functionBunEscapeHTML, (JSC::JSGlobalObject * lexicalGl
     JSC::JSValue argument = callFrame->argument(0);
     if (argument.isEmpty())
         return JSValue::encode(jsEmptyString(vm));
-    if (argument.isNumber() || argument.isBoolean())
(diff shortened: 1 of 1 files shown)


robobun:
@Jarred-Sumner, your commit e9b9f79ca413a44b520be8d864b49cd691658f47 is building: [`#2652`](https://buildkite.com/bun/bun/builds/2652)

Jarred-Sumner:
Manually verified this fixes it, which I think is good enough in this case
```
1 | var base = "a".repeat(1024).repeat(1024).repeat(1024);    
2 | globalThis.strings = [base];
3 | for (let i = 0; i < 1024; i++) {
4 |       const out = base + i + "ÎõÎõ";
5 |       globalThis.strings.push(out);
6 |       Bun.escapeHTML(out);
              ^
RangeError: Out of memory
      at C:\bun\oomie.js:6:11

Bun v1.1.27-canary.62+e9b9f79ca (Windows x64)
```

Issue #2658, opened 2023-04-14
why does the C++ mac build take 45 minutes?


Jarred-Sumner:

Jarred-Sumner:
Output from [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) 

```c++
Analyzing build trace from 'out.file'...
**** Time summary:
Compilation (470 times):
  Parsing (frontend):         1025.6 s
  Codegen & opts (backend):    156.6 s

**** Files that took longest to parse (compiler frontend):
  5629 ms: src/bun.js/debug-bindings-obj//JSWebSocket.o
  5539 ms: src/bun.js/debug-bindings-obj//JSReadableStreamSource.o
  5522 ms: src/bun.js/debug-bindings-obj//JSWritableStream.o
  5513 ms: src/bun.js/debug-bindings-obj//JSURLSearchParams.o
  5478 ms: src/bun.js/debug-bindings-obj//ReadableStreamSource.o
  5414 ms: src/bun.js/debug-bindings-obj//ReadableStreamSink.o
  5410 ms: src/bun.js/debug-bindings-obj//ZigGlobalObject.o
  5361 ms: src/bun.js/debug-bindings-obj//JSTransformStreamDefaultController.o
  5341 ms: src/bun.js/debug-bindings-obj//ScriptWrappable.o
  5325 ms: src/bun.js/debug-bindings-obj//JSTextEncoder.o
  5299 ms: src/bun.js/debug-bindings-obj//bindings.o
  5248 ms: src/bun.js/debug-bindings-obj//StructuredClone.o
  5218 ms: src/bun.js/debug-bindings-obj//WebCoreTypedArrayController.o
  5211 ms: src/bun.js/debug-bindings-obj//JSReadableStreamSourceCustom.o
  5130 ms: src/bun.js/debug-bindings-obj//TextEncoder.o
  5130 ms: src/bun.js/debug-bindings-obj//JSTransformStream.o
  5112 ms: src/bun.js/debug-bindings-obj//SharedBuffer.o
  4952 ms: src/bun.js/debug-bindings-obj//WebSocket.o
  4946 ms: src/bun.js/debug-bindings-obj//JSReadableStreamSink.o
  4931 ms: src/bun.js/debug-bindings-obj//ReadableStreamDefaultController.o
  4830 ms: src/bun.js/debug-bindings-obj//JSFetchHeaders.o
[...]

PR #13687, opened 2024-09-02
Call cancel on ReadableStream when Bun.serve() response is aborted

### LINKED ISSUE #4517: [HTTP server] req.signal.onabort not called when HTTP request is closed

### LINKED ISSUE #6758: ReadableStream cancel is not called when client disconnects

1.0.7

Debian Bullseye x64

Running Bun with Svelte, in a API GET function:
```
    let keepAlive = true;
    const stream = new ReadableStream({
        start: async controller =>{
            while(keepAlive){           
                await new Promise(r => setTimeout(r, 1000));
            }
        },
        cancel: () =>{
            console.log("cancelled");
            keepAlive = false;
        }
    });
    return new Response(stream, {
        headers: {
            "Cache-Control": "no-store",
            "Content-Type": "text/event-stream",
            "Connection": "keep-alive",
        }
    });
```

Everytime the page is closed or the connection is closed by the client we should see a log on the server "cancelled".

### What do you see instead?

Nothing is logged and the loop keeps running even if the page is fully closed.

### Additional information

It works just fine with Node

Jarred-Sumner:
### What does this PR do?

Fixes #4517
Fixes #6758

There were several different bugs causing #4517 and #6758.
- AbortSignal was being GC'd too early. While the C++ reference-counted object was kept alive, the JavaScript wrapper object was not. This meant there were situations that caused the `"abort"` event to not be emitted
- When given a ReadableStream (or async iterator), we were only starting the response after ReadableStream's `pull` or `start` async function resolved. This means if the request was aborted before the callback was run, it would never call the `cancel` function on the ReadableStream.
- For the async iterator ReadableStream case, we were throwing an uncatchable exception on abort. 

For GC, we add a pending activity counter to `AbortSignal`. 

For the AbortSignal changes, there are memory leak tests

For the ReadableStream changes, there are new tests in serve.test.ts

For the readable stream async iterator changes, previous tests were failing after we started calling cancel/close on the ReadableStream.

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -2351,7 +2351,10 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             // if signal is not aborted, abort the signal
             if (this.signal) |signal| {
                 this.signal = null;
-                defer signal.unref();
+                defer {
+                    signal.pendingActivityUnref();
+                    signal.unref();
+                }
                 if (!signal.aborted()) {
                     signal.signal(globalThis, .ConnectionClosed);
                     any_js_calls = true;
@@ -2416,7 +2419,10 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             // if signal is not aborted, abort the signal
             if (this.signal) |signal| {
                 this.signal = null;
-                defer signal.unref();
+                defer {
+                    signal.pendingActivityUnref();
+                    signal.unref();
+                }
                 if (this.flags.aborted and !signal.aborted()) {
                     signal.signal(globalThis, .ConnectionClosed);
                 }
@@ -6732,6 +6738,7 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
             ctx.request_body = body;
             var signal = JSC.WebCore.AbortSignal.new(this.globalThis);
             ctx.signal = signal;
+            signal.pendingActivityRef();
 
             const request_object = Request.new(.{
                 .method = ctx.method,
--- a/src/bun.js/bindings/AsyncContextFrame.h
+++ b/src/bun.js/bindings/AsyncContextFrame.h
@@ -36,7 +36,7 @@ public:
     {
         if constexpr (mode == JSC::SubspaceAccess::Concurrently)
             return nullptr;
-        return WebCore::subspaceForImpl<AsyncContextFrame, Bun::UseCustomHeapCellType::No>(
+        return WebCore::subspaceForImpl<AsyncContextFrame, WebCore::UseCustomHeapCellType::No>(
             vm,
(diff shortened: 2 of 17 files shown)


robobun:
@Jarred-Sumner, your commit 169a5709249a519708a0a59b3d32d4aeac56c767 is building: [`#2658`](https://buildkite.com/bun/bun/builds/2658)

cirospaciari (approved):
Thanks looks good, appreciate the comments

kaiykay (commented):
What free balance ton

kaiykay:
What free balance ton

PR #13694, opened 2024-09-02
fix(transpiler): remove react element inlining

### LINKED ISSUE #11025: React 19 SSR Fails In Production Mode, But Works In Development

1.1.8

Darwin 23.4.0 arm64 arm

1. Build a react 19 application and  statically server the files
2. Set node env to development mode
3. Use render to renderToReadableStream to server the application
```tsx
import { renderToReadableStream } from "react-dom/server";

await renderToReadableStream(element, {
			signal: request.signal,
			bootstrapScriptContent: [`_INIT_PATH=${JSON.stringify(pathname)}`, `_ROUTES=${JSON.stringify(routes)}`, `_PROPS=${JSON.stringify(props)}`].filter(Boolean).join(";"),
			bootstrapModules: [`${routes["/hydrate"]}`],
			identifierPrefix: "app",
		});
```

4. set up the entry point to the application for the client
```tsx
const root = hydrateRoot(document, <App />);
```

5. start the application and visit the page
6. At this point in development mode everything works as normal

---
How to recreate the error

1. Set NODE_ENV to production and follow steps 1-6 and you'll receive 

```
{"name":"Error","message":"Objects are not valid as a React child (found: object with keys {$$typeof, type, key, ref, props, _owner}). If you meant to render a collection of children, use an array instead."}
```

The expected behavior is that in production mode , the react 19 application would successfully render with bun renderToReadableStream

### What do you see instead?

Instead in production the hydration will fail

image

### Additional information

One way to get the code working is to  moneky patch  the following code out 

react-dom-server.bun.production.js
```tsx

	throw Error(
				"Objects are not valid as a React child (found: " +
					("[object Object]" === childIndex ? "object with keys {" + Object.keys(node$jscomp$0).join(", ") + "}" : childIndex) +
					"). If you meant to render a collection of children, use an array instead.",
			);
		}
```
however by doing so styles will not load correctly when passing a style from the server

```tsx
export const app = new Elysia({ prefix: "" })
	.get("/home", async (ctx) => {
		const Module = await import("../app/home");
		const css = await generate(ctx.path, config.app.dir);
		const props = {
			meta: {
				title: "",
				description: "",
			},
			style: css,
		};

		const stream = await renderToStream(
			<Shell {...props}>
				<script src="/loader.js" type="module" />
				<Module.default {...props} />
				{/* <style precedence="high">{css}</style>  */}
			</Shell>,
			props,
			ctx.request,
		);
		return new Response(stream, {
			headers: { "Content-Type": "text/html; charset=utf-8" },
		});
	})
```

I got around this by passing the styles as a prop and using the 

<style></style> tag in the component itself.
Typically 
```
{/* <style precedence="high">{css}</style>  */}
```
would work in development mode

paperclover:
### What does this PR do?

Fixes #11025

React 19 changed their element format in https://github.com/facebook/react/pull/28813 to use a different symbol, as a warning that

- the underlying element format has changed
- the underlying format is going to receive more changes

It seems that inlining as we do it will not be a viable strategy with going forward, especially seeing Sebastian's comments on the subject.

> I mean those are either broken or causing the page to deopt so you probably want to change them ASAP.

> In addition to changing the format, we plan on JSX in RSC to have side-effects so the inlining strategy won't work in RSC regardless. So just upgrading an inlining plugin to use the new format might not be a good idea.

This changeset is tested by bundler_jsx ensuring that the symbol used in `react/jsx-runtime` is used instead of `react.element`. This is something I had already written a test for, but previously it was special cased to the inlined `react.element`.


robobun:
@paperdave, your commit ab8158496f97eb4403f03ea32a53e8eacc265b9d is building: [`#2662`](https://buildkite.com/bun/bun/builds/2662)

Jarred-Sumner:
This was a 30% performance boost to React SSR. I do think we should expect to bring this back in some form.

PR #13697, opened 2024-09-02
build: use generic x86_64 microarchitecture levels

getchoo:
### What does this PR do?

Currently the Intel CPU families Nehalem and Haswell are used to denote "baseline" and "modern" builds. This creates a bit of a misnomer in the former, as Nehalem is not actually the baseline microarchitecture level for x86_64, but x86_64-v2. Likewise, Haswell is not the modern x86_64-v3 target but a superset of it with instructions that may or may not be present on equivalent x86_64-v3 (non-Haswell) processors. See https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels for a general overview as well as the differences between these targets in zig's stdlib here https://ziglang.org/documentation/master/std/#std.Target.x86.cpu

This is unlikely to cause issues in practice, but is still probably a good idea to prevent any edge cases in the future. It also makes the actual intention behind setting the microarchitecture level here more obvious as you no longer need to remember the vague Intel CPU family names and which level they correspond to

- [ ] Code changes


190n:
Using microarchitecture levels instead of specific generations is a good idea in general, but I don't think Bun will work right now on v1 without more changes because we use Cloudflare's fork of zlib which requires SSE4.2.

No matter what we do here, I think we should also start testing the baseline builds under userspace QEMU with specific CPU features enabled and disabled, to make sure that the baseline builds will actually run on the CPUs they are intended to run on.

Issue #13690, opened 2024-09-02
Implement `bun pm version`


aryzing:
### What is the problem this feature would solve?

Getting and modifying a local package's version info, as is currently done with [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version)

### What is the feature you are proposing to solve the problem?

`bun pm version` or similar

PR #13703, opened 2024-09-03
`pm version` command for updating package version

versecafe:
### What does this PR do?

This implements #13690 allowing for example `buin pm version minor` to bump the package version by 0.1.0, based off the [`npm version`](https://docs.npmjs.com/cli/v10/commands/npm-version) command

> Still in process and marked as draft


versecafe:
@dylan-conway sorry for the ping but I see you've done the most recent edits on the CLI parsing with clap, I'm trying to find where the match for the parsers is being passed in for `<POS>`, `<STR>`, `<dir>` etc to add a new parse input for a enum following the zig-clap docs but I can't seem to find the matcher anywhere

for reference it should look something like this

```zig
const parsers = comptime .{
    .STR = clap.parsers.string,
    .FILE = clap.parsers.string,
    .INT = clap.parsers.int(usize, 10),
    .ANSWER = clap.parsers.enumeration(YesNo),
};
 ```
 
 If you could help me out when you have some time that would be great.

Issue #2672, opened 2023-04-17
Use bun's bundler on runtime JS builtin modules


Jarred-Sumner:
Instead of our current pattern of `src/bun.js/*.exports.js`

What if we did a directory structure like:
- 📁  builtins
- 📁  builtins/bun
- 📂  builtins/bun/ffi.js 
- 📂  builtins/bun/sqlite.js
- 📁  builtins/node/
- 📂  builtins/node/fs.js
- 📂  builtins/node/dns.js
- 📂  builtins/node/fs/promises.js
- 📂  builtins/internal/...

From there, we would bundle each top-level module individually (`node:fs` as one bundle). Note that the build system here will be a little bit complicated. For good reasons, we have to concatenate some code to the top of `node:fs` and `bun:ffi` at Bun's compile-time, which means that those files appear to have unbound global identifiers. Some magic required.

In debug builds of Bun, we can configure it to run the bundler on import of one of these modules.

In release builds of Bun, we would ahead of time build each file and commit built files into the repository so that people can find it online.

### TypeScript?

I think it's probably worthwhile to start writing new modules in TypeScript, but the priority must be (1) correctness and (2) performance.

If we find ourselves writing code in a way that prioritizes making `tsc` not show an error in-editor instead of performance, the TS error should be ignored or the tsconfig.json rules should be changed to permit the behavior. JavaScript builtins code is extremely pedantic, sometimes weird, and that's okay

Our users matter more than our internal developer experience of building Bun.

It might sound convenient for bun-types to be generated from these files that are used, but that won't work:
- Most of the actual implementations of things are not in JavaScript/TypeScript. `node:fs` in Bun mostly just calls .bind() on a bunch of functions
- The internal types should probably be different than the public types. We need to be extremely careful about runtime type checking. `instanceof` is not good enough.

Electroid:
We've started to do some of this, going to close since this is a chore.

PR #13704, opened 2024-09-03
Introduce fast path for buffered ReadableStream

### LINKED ISSUE #13678: Memory leak if accessing body of fetch() response before consuming the stream

1.1.26+0a37423ba

Linux 6.6.16-linuxkit aarch64

Access 'body' of fetch() response before consuming the stream causes memory leak.

Code snippet: https://github.com/mshameti/bun-memory-accumulation-formdata-fetch-docker/blob/main/sender/script.js#L48-L72

The focus of this issue is LEAK 2. 

NOTE: This is the same repo used to report https://github.com/oven-sh/bun/issues/13657, but LEAK 2 is a different but closely related to LEAK 1. 

```
git clone https://github.com/mshameti/bun-memory-accumulation-formdata-fetch-docker
cd bun-memory-accumulation-formdata-fetch-docker
LEAK=2 docker-compose up --build
```

No memory leak.

### What do you see instead?

Memory accumulation until container crashes.

### Additional information

_No response_

Jarred-Sumner:
### What does this PR do?

Fixes #13678

This introduces a fast path for consuming streaming request bodies where previously the developer accessed the ReadableStream. This also fixes a memory leak when doing that.

Previously, the following code would be unnecessarily slow:
```js
Bun.serve({
  async fetch(req) {
    if (req.body) { 
       await req.json();
    }

    // previously, the fast path only worked if you didn't access req.body.
  }
});
```

This would cause us to:
1) Buffer the contents of the stream in JavaScript (slow)
2) High water mark of around 16 KB because we didn't implement the drain function for it.  

TODO:
- [x] Fix FormData issue
- [x] Fix other test failures
- [x] benchmark once CI finishes

There are some new tests, and also I modified some existing tests to also cover this case. 

There's a lot of room for mistakes here. So let's be very confident this works properly when we merge it.


robobun:
@Jarred-Sumner, your commit 84d07716fc04d48cef91d795900f2ad47079bc19 is building: [`#2672`](https://buildkite.com/bun/bun/builds/2672)

test/js/bun/http/bun-serve-static.test.ts
```diff
@@ -143,7 +145,7 @@ describe("static", () => {
         Bun.gc(true);
 
         const rss = (process.memoryUsage.rss() / 1024 / 1024) | 0;
-        expect(rss).toBeLessThan(baseline * 4);
```
Jarred-Sumner, line 145: This test was not failing when it should've been failing. The test itself shouldn't use a ton of memory so I think a hardcoded value is fine here.

src/codegen/buildTypeFlag.ts
```diff
@@ -0,0 +1,16 @@
+const buildTypeFlag = process.argv.find(argv => {
```
Jarred-Sumner, line 1: didn't mean to commit this

src/bun.js/webcore/response.zig
```diff
@@ -1033,9 +1033,18 @@ pub const Fetch = struct {
...
+                        this.scheduled_response_buffer = .{
+                            .allocator = bun.default_allocator,
+                            .list = .{
+                                .items = &.{},
+                                .capacity = 0,
+                            },
+                        };
                         readable.ptr.Bytes.onData(
                             .{
-                                .temporary_and_done = bun.ByteList.initConst(chunk),
```
Jarred-Sumner, line 1038: subtle change

src/bun.js/bindings/Uint8Array.cpp
```diff
@@ -25,4 +22,23 @@ extern "C" JSC::EncodedJSValue JSUint8Array__fromDefaultAllocator(JSC::JSGlobalO
...
     return JSC::JSValue::encode(uint8Array);
 }
+
+extern "C" JSC::EncodedJSValue JSArrayBuffer__fromDefaultAllocator(JSC::JSGlobalObject* lexicalGlobalObject, uint8_t* ptr, size_t length)
+{
+
+    JSC::JSArrayBuffer* arrayBuffer;
+
+    if (LIKELY(length > 0)) {
+        RefPtr<ArrayBuffer> buffer = ArrayBuffer::createFromBytes({ ptr, length }, createSharedTask<void(void*)>([](void* p) {
```
cirospaciari, line 32: should we handle exceptions here?

Jarred-Sumner: this line won't throw, it's a C++ ref counted ptr

src/codegen/buildTypeFlag.ts
cirospaciari: can this be removed so do not cause confusion in the future?

cirospaciari (approved):
LGTM just the uneeded file that I think we should remove so we not associate with this PR in the future

cirospaciari (changes requested):
Failing leak memory test.

mshameti:
thank you for your hard work on this @Jarred-Sumner! the incredibly low turnaround time of issue to fix and enhancement is understated

Issue #13705, opened 2024-09-03
Using pino-pretty as transport with fastify leads to crash


binamralamsal:
### What version of Bun is running?

1.1.27

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

This is my code
```ts
import Fastify from "fastify";

const fastify = Fastify({
  logger: {
    // Bug in bun
    transport: {
      target: "pino-pretty",
    },
  },
});

fastify.get("/", (request, reply) => reply.send("Hello"));

try {
  await fastify.listen({ port: 3001 });
  console.log("Server listening on port 3001");
} catch (err) {
  fastify.log.error(err);
  process.exit(1);
}
```
It only happens when you include this line:
```ts
transport: {
  target: "pino-pretty",
},
````
If I comment it, it works fine.

### What is the expected behavior?

The expected behavior is for fastify to use pino-pretty when logging into the console.

### What do you see instead?

```sh
$ bun src/index.ts
Server listening on port 3001
============================================================
Bun Canary v1.1.27-canary.66 (c50f8d82) Windows x64
Windows v.win10_fe
CPU: sse42 avx avx2
Args: "C:\Users\ASUS\.bun\bin\bun.exe" "src/index.ts"
Features: jsc dotenv(2) http_server(2) tsconfig(10) 
Builtins: "bun:main" "node:assert" "node:async_hooks" "node:buffer" "node:dns" "node:events" "node:fs" "node:http" "node:https" "node:module" "node:os" "node:path" "node:stream" "node:string_decoder" "node:tty" "node:url" "node:util" "node:util/types" "node:worker_threads" "node:diagnostics_channel" 
Elapsed: 365ms | User: 46ms | Sys: 31ms
RSS: 0.22GB | Peak: 0.22GB | Commit: 0.29GB | Faults: 53448

panic(thread 19264): Cast 512[handle] -> FDImpl.UV makes closing impossible!

The supplier of this FileDescriptor should call 'bun.toLibUVOwnedFD'
or 'FDImpl.makeLibUVOwned', probably where open() was called.
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

[...]

Issue #13702, opened 2024-09-03
Cast 104[handle] -> FDImpl.UV makes closing impossible!


malooski:
### How can we reproduce the crash?

When using the Bluesky API library, seems to crash when setting up credential session.

```
import { CredentialSession } from "@atproto/api";

const session = new CredentialSession(new URL(ENV.BSKY_SERVICE_URL));
```

### Relevant log output

```shell
The supplier of this FileDescriptor should call 'bun.toLibUVOwnedFD'
or 'FDImpl.makeLibUVOwned', probably where open() was called.
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

https://bun.report/1.1.26/wa10a37423AioigwD0q9lQq3ol3C__24xsyBA0eNo1zbEKwjAURuG9T/E7VUGLgrOLpSAILtZFHJLmaoK3uSG3pfj2ouh84Hx7owM26+3Vm+iYbljt0NSHPnHVXtCbJyk6Fg3xgdAnUQ2WaVYUZ0/QMSUOlCF3DD4omsBUk3Y5pEEy1MvIDp1hRmnHWA1yDLa9nKZIrqnLQjLKn/ah/pFcuUTKYo3lFyZPmSCJ4nyByej3R656A5dYQK0
```

### Stack Trace (bun.report)

Bun v1.1.26 ([`0a37423`](<https://github.com/oven-sh/bun/tree/0a37423bafab6aebe803e5f1594cffa9796f2ae6>)) on windows x86_64 [AutoCommand]

**panic**: Cast 488[handle] -> FDImpl.UV makes closing impossible!

The supplier of this FileDescriptor should call 'bun.toLibUVOwnedFD'
or 'FDImpl.makeLibUVOwned', probably where open() was called.

- [`js_ast.zig:3211`](<https://github.com/oven-sh/bun/blob/0a37423bafab6aebe803e5f1594cffa9796f2ae6/src/js_ast.zig#L3211>): `assert`
- [`JSCJSValue.h:752`](<https://github.com/oven-sh/bun/blob/0a37423bafab6aebe803e5f1594cffa9796f2ae6/C:/buildkite-agent/builds/windows-x64-hetzner-6/bun/bun/build/bun-webkit/include/JavaScriptCore/JSCJSValue.h#L752>): `JSC::ensureStillAliveHere`
- *2 unknown/js code*
- `llint_entry`

Jarred-Sumner:
[...]

Issue #11401, opened 2024-05-27
Bun crash when enable `pino-pretty`


joeynq:
### JavaScript/TypeScript code that reproduces the crash?

```ts
Bun.serve({
  websocket: {
    message: (ws, data) => {
      const sid = uuid();
      return asyncLocalStorage.run({ sid }, {
        try {
          throw new Error("it crash after caught this");
        } catch (err) {
          ws.publish(err.message);
        }
      });
    }
  }
});
```

### Relevant log output

```shell
Bun v1.1.10 (5102a944) Linux x64
Args: "bun" "src/index.ts"
Features: jsc Bun.stdin(2) bunfig dotenv(8) fetch http_server transpiler_cache(4) tsconfig(47)
Builtins: "bun:main" "bun:sqlite" "node:assert" "node:async_hooks" "node:buffer" "node:events" "node:fs" "node:module" "node:os" "node:path" "node:process" "node:stream" "node:string_decoder" "node:tty" "node:url" "node:util" "node:util/types" "node:worker_threads"
Elapsed: 150928ms | User: 1480ms | Sys: 1398ms
RSS: 0.69GB | Peak: 0.35GB | Commit: 0.69GB | Faults: 0

panic: Segmentation fault at address 0x6EA0
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.10/la15102a94AqqihgF280ujE+xyQksllBA2Agq3B

Illegal instruction
```

### Stack Trace (bun.report)

Bun v1.1.10 ([`5102a94`](<https://github.com/oven-sh/bun/tree/5102a944307d56933a6bb039b99ad76a758b46e4>)) on linux x86_64 [AutoCommand]

Segmentation fault at address 0x00006EA0

- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/5102a944307d56933a6bb039b99ad76a758b46e4/ld-temp.o#L0>): `WTF::jscSignalHandler`
- `??`
- `??`

Electroid:
Could you share some code to reproduce this issue? It looks like the stack trace doesn't give us enough information.

joeynq:
@Electroid I added example. I think the problem come from `AsyncLocalStorage`. When I remove als, it run without any crashes.

joeynq:
@Electroid I think I found the root cause. Turns out `pino-pretty` is the main problem.

[...]

Issue #2704, opened 2023-04-20
Some async code doesn't produce stack trace on error


dhbaird:
### What version of Bun is running?

0.5.9

### What platform is your computer?

Linux

### What steps can reproduce the bug?

```ts
async function aBroken() {
}

async function bBroken() {
    await aBroken();
    throw new Error("Oops"); // <-- no stack trace for this throw
}

async function cBroken() {
    await bBroken();
}

async function aOkay() {
}

async function bOkay() {
    throw new Error("Oops");
    await aOkay();
}

async function cOkay() {
    await bOkay();
}

async function main() {
    try {
        await cBroken();
    }
    catch (e) {
        console.log(e);

        // >>>(bun)
        // 1 | async function aBroken() {
        // 2 | }
        // 3 | 
        // 4 | async function bBroken() {
        // 5 |     await aBroken();
        // 6 |     throw new Error("Oops");
        //               ^
        // error: Oops
        //       at /.../throwy.ts:6:10

        // >>>(node)
        // Error: Oops
        //     at bBroken (/.../throwy.js:6:11)
        //     at async cBroken (/.../throwy.js:10:5)
        //     at async main (/.../throwy.js:27:9)
    }
    try {
        await cOkay();
    }
    catch (e) {
        console.log(e);
        // 
        // >>>(bun)
        // 12 | 
        // 13 | async function aOkay() {
        // 14 | }
        // 15 | 
        // 16 | async function bOkay() {
        // 17 |     throw new Error("Oops");
        //               ^
        // error: Oops
        //       at /.../throwy.ts:17:10
        //       at bOkay (/.../throwy.ts:16:20)
        //       at /.../throwy.ts:22:10
        //       at cOkay (/.../throwy.ts:21:20)
        //       at /.../throwy.ts:50:14

        // >>>(node)
        // Error: Oops
        //     at bOkay (/.../throwy.js:17:11)
        //     at cOkay (/.../throwy.js:22:11)
        //     at main (/.../throwy.js:50:15)
    }
}

main();
```

### What is the expected behavior?

Expect stack trace for the first case.

### What do you see instead?

Stack trace is missing.
```ts
        // error: Oops
        //       at /.../throwy.ts:6:10
```

### Additional information

Thanks!! :) (Thought that maybe this should be a feature - but the bug reporting template was more useful, so I picked that.)

ayonli:
It seems this is a JavaScriptCode bug, if the function is async, JSC cannot produce the full stack trace.

[...]

PR #13719, opened 2024-09-04
windows: fix sometimes crash when FDImpl.uv() is called on stdio

nektro:
Closes https://github.com/oven-sh/bun/issues/13653
Closes https://github.com/oven-sh/bun/issues/13705
Closes https://github.com/oven-sh/bun/issues/13702
Closes https://github.com/oven-sh/bun/issues/11401
Closes https://github.com/oven-sh/bun/issues/13750


robobun:
@nektro, your commit 7ff3680cf97d43fa786472234100e8c86fd1955b is building: [`#2704`](https://buildkite.com/bun/bun/builds/2704)

Jarred-Sumner:
can you add a test

Jarred-Sumner:
The test segfaulted

src/fd.zig
```diff
@@ -170,14 +170,29 @@ pub const FDImpl = packed struct {
...
-                    \\The supplier of this FileDescriptor should call 'bun.toLibUVOwnedFD'
-                    \\or 'FDImpl.makeLibUVOwned', probably where open() was called.
-                ,
-                    .{this},
-                ),
+                .system => {
+                    const w = std.os.windows;
+
+                    const S = struct {
+                        fn is_stdio_handle(id: w.DWORD, handle: w.HANDLE) bool {
```
paperclover, line 177: instead of declaring a struct with a function here, why not just move the function above `pub fn uv`?

paperclover (approved):
unfortunate that system tagged fds are introduced somewhere for the standard io handles, but this fix makes sense.

Jarred-Sumner (changes requested):
Can’t merge until the test passes on windows

Issue #13373, opened 2024-08-17
node:http request events not firing


Lyall-A:
### What version of Bun is running?

1.1.24+85a329911

### What platform is your computer?

Linux 6.10.5-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

```js
const http = require("http");
const server = http.createServer();

server.on("request", (req, res) => {
    console.log("opened");
    req.on("close", () => console.log("closed")); // doesnt fire with bun
    res.end("hi");
});

server.listen(5001);
```

### What is the expected behavior?

$ node ./test.js
$ curl http://localhost:5001
opened
closed

### What do you see instead?

$ bun ./test.js
$ curl http://localhost:5001
opened

kevinmingtarja:
i  was able to repro this too on `1.1.25`, and `Darwin 23.6.0 arm64`.

i dug a bit deeper, and was able to see that with node, if i pass in a `"Connection: close"` header, the http connection does get closed, as seen by the verbose output of curl:

```bash
% curl 127.0.0.1:5001 -H "Connection: close" -v
* processing: 127.0.0.1:5001
*   Trying 127.0.0.1:5001...
* Connected to 127.0.0.1 (127.0.0.1) port 5001
> GET / HTTP/1.1
> Host: 127.0.0.1:5001
> User-Agent: curl/8.2.1
> Accept: */*
> Connection: close
> 
< HTTP/1.1 200 OK
< Date: Sun, 18 Aug 2024 03:54:19 GMT
< Connection: close
< Content-Length: 2
< 
* Closing connection
hi
```

But with Bun, it does not:
```bash
% curl 127.0.0.1:5001 -H "Connection: close" -v
* processing: 127.0.0.1:5001
*   Trying 127.0.0.1:5001...
* Connected to 127.0.0.1 (127.0.0.1) port 5001
> GET / HTTP/1.1
> Host: 127.0.0.1:5001
> User-Agent: curl/8.2.1
> Accept: */*
> Connection: close
> 
< HTTP/1.1 200 OK
< content-type: text/plain;charset=utf-8
< Date: Sun, 18 Aug 2024 03:55:17 GMT
< Content-Length: 2
< 
* Connection #0 to host 127.0.0.1 left intact
hi
```

I also noticed that in Bun, `closeIdleConnections` is not implemented, maybe this is related?

[...]

PR #13738, opened 2024-09-04
fix(node:http) emit request `close` event and mark `complete` only after close

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/13373
Fix: https://github.com/oven-sh/bun/issues/7716
Fix: https://github.com/oven-sh/bun/issues/13745 (remove default timeout on server)

- [X] Code changes

Tests

### DIFF
--- a/src/js/node/http.ts
+++ b/src/js/node/http.ts
@@ -20,7 +20,7 @@ const {
   getHeader: (headers: Headers, name: string) => string | undefined;
   setHeader: (headers: Headers, name: string, value: string) => void;
   assignHeaders: (object: any, req: Request, headersTuple: any) => boolean;
-  assignAbortCallback: (req: Request, callback: () => void, ctx: any) => void;
+  assignAbortCallback: (req: Request, callback: () => void) => void;
   Response: (typeof globalThis)["Response"];
   Request: (typeof globalThis)["Request"];
   Headers: (typeof globalThis)["Headers"];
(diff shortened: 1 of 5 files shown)


robobun:
@cirospaciari, your commit b61114b305a567b4ff1f0a30512419d926787329 is building: [`#2751`](https://buildkite.com/bun/bun/builds/2751)

src/js/node/http.ts
```diff
@@ -637,6 +647,7 @@ Server.prototype = {
           const prevIsNextIncomingMessageHTTPS = isNextIncomingMessageHTTPS;
           isNextIncomingMessageHTTPS = isHTTPS;
           const http_req = new RequestClass(req);
+          req.signal.addEventListener("abort", onRequestAborted.bind(http_req));
```
Jarred-Sumner: I think we can be a little smarter about this

cirospaciari: Should I add a internal method (`$zig`)  that just add the JSValue/JSC.Strong of the onRequestedAborted function on the `Request`? We already have the request weak ptr to trigger this function with should be enough because RequestClass holds a strong ref to the request. We would increase the size of all `Request` objects by a little bit, but would avoid creating the abort signal it self.

A further optimization would be avoiding the `.bind` by having a `http_req` ref to pass onRequestAborted but would cost another field to the `Request` object.

In the perfect scenario we should just rewrote http module to use directly the C api and avoid all this wrapping.

cirospaciari: Something like this: https://github.com/oven-sh/bun/pull/13738/commits/89aaed958953bbe688803fefa374f9c3807a9706

PR #13741, opened 2024-09-04
add hkdfSync + hkdf to node:crypto

e253:
### What does this PR do?

~~Note: `KeyObject` input for the `ikm` arg is not supported, although it is allowed in `@types/node`. There aren't Zig bindings for `KeyObject`.~~
Implemented KeyObject processing in JS.

This adds to new functions to `node:crypto` namespace that were missing. This brings bun closer to node drop in compatibility.

- [x] Code changes

I wrote automated tests.

Not necessary because these functions are already defined in `@types/node`

### DIFF
--- a/src/js/node/crypto.ts
+++ b/src/js/node/crypto.ts
@@ -24,7 +24,12 @@ const {
   randomInt: _randomInt,
   pbkdf2: pbkdf2_,
   pbkdf2Sync: pbkdf2Sync_,
-} = $zig("node_crypto_binding.zig", "createNodeCryptoBindingZig");
+  hkdfSync: _hkdfSync,
+} = $zig("node_crypto_binding.zig", "NodeCrypto.create");
+
+function hkdfSync(digest, ikm, salt, info, keylen) {
+  return _hkdfSync(digest, ikm, salt, info, keylen);
+}
 
 function randomInt(min, max, callback) {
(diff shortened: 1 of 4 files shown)


src/js/node/crypto.ts
```diff
@@ -24,7 +24,26 @@ const {
...
+    ikm = ikm.export();
+  }
+
+  return _hkdfSync(digest, ikm, salt, info, keylen).buffer;
+}
+
+function hkdf(digest, ikm, salt, info, keylen, callback) {
+  if (callback != null) {
+    // Causes weird test failures
+    // process.nextTick(() => callback(null, _hkdfSync(digest, ikm, salt, info, keylen)));
```
Jarred-Sumner: For the callback version, we probably do need use nextTick, but I would suggest doing the work first and then enqueueing that if we do want to keep it synchronous for now

e253: Fixed. The work is done up front and then the callback is called with the result on the next tick. A more faithful async could be implemented like `pbkdf2`, but the current patch should allow packages dependent on `hkdf`/`hkdfSync` to run as they would on Node.

Apollon77:
For running matter.js with bun, like @JimBuzbee tries currently, we would also need hkdf support

Issue #12007, opened 2024-06-20
Malformed_HTTP_Response error intermittently when using bun fetch proxy


mdervisaygan:
### What version of Bun is running?

1.1.15+b23ba1fe1

### What platform is your computer?

Linux 6.5.0-35-generic x86_64 x86_64

### What steps can reproduce the bug?

This error does not occur when no proxy is used. You need a proxy and a website that gives a 403 error for the error to occur. Just enter an ipv4 proxy in the code below and run it a few times.

[Node.js] Tested code:
```
import fetch from 'node-fetch';
import { HttpsProxyAgent } from 'https-proxy-agent';
fetch('https://nopecha.com/demo/cloudflare', {
    agent: new HttpsProxyAgent(`http://<proxy-username>:<proxy-password>@<proxy-ip>:<proxy-port>`)
})
    .then(res => res.text())
    .then(res => {
        console.log('Success Response');
    })
    .catch(console.error)
```

[Bun] Tested code:
```
import fetch from 'node-fetch';
fetch(`https://nopecha.com/demo/cloudflare`, {
    proxy: 'http://<proxy-username>:<proxy-password>@<proxy-ip>:<proxy-port>'
})
    .then(res => res.text())
    .then(res => {
        console.log('Success Response')
    })
    .catch(console.error)
```
node.js output:
Always returns the page source

https://github.com/oven-sh/bun/assets/123484092/ff515bff-b032-46b6-95fb-6c68bbee5797

Bun output:
Sometimes it returns the page source code, sometimes it returns the following error.

```
Malformed_HTTP_Response: Malformed_HTTP_Response fetching "https://nopecha.com/demo/cloudflare". For more information, pass `verbose: true` in the second argument to fetch()
 path: "https://nopecha.com/demo/cloudflare"

```

https://github.com/oven-sh/bun/assets/123484092/79c82179-2194-4149-b631-158ce1945cd1

### What is the expected behavior?

When using bun fetch and proxy, requests returning a 403 response should return the page source code.

### What do you see instead?

Sometimes it returns the page source code, sometimes it returns the following error.
Malformed_HTTP_Response: Malformed_HTTP_Response fetching "https://nopecha.com/demo/cloudflare". For more information, pass `verbose: true` in the second argument to fetch()
 path: "https://nopecha.com/demo/cloudflare"

### Additional information

[...]

PR #13752, opened 2024-09-05
fix(fetch) fix lifecycle of SSL Proxy, fix lifecycle of tls_props, fix handling chunked encoded redirects when proxing.

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/12007

- [X] Code changes

Added tests

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -293,16 +293,13 @@ const ProxyTunnel = struct {
                         return;
                     }
                 },
-                .proxy_decoded_headers, .proxy_headers => {
-                    this.flags.proxy_tunneling = false;
-                    this.state.response_stage = .proxy_decoded_headers;
-
+                .proxy_headers => {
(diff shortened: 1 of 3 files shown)


robobun:
@cirospaciari, your commit 78185dbc629665044b9f5e8460a2f3af6375aa06 is building: [`#2841`](https://buildkite.com/bun/bun/builds/2841)

src/http.zig
```diff
@@ -210,13 +210,18 @@ const ProxyTunnel = struct {
         none: void,
     } = .{ .none = {} },
     write_buffer: bun.io.StreamBuffer = .{},
+    ref_count: u32 = 1,
```
Jarred-Sumner, line 213: Does it need to be threadsafe?

cirospaciari: Don't need only http thread uses it

PR #2870, opened 2023-05-13
Implement `expect.toBeCloseTo`

blackmann:
Progress for #1825

PR #13756, opened 2024-09-05
Implement napi_handle_scope and napi_escapable_handle_scope

190n:
Draft because I have a hunch something will break when handle scopes are nested.

### What does this PR do?

We thought we wouldn't need these due to JSC's stack scanning. Turns out, NAPI modules can put `napi_value`s on the heap where the stack scanner won't see them. So we need a `napi_handle_scope` to keep them alive.

`NapiHandleScopeImpl` is a heap-allocated wrapper around `WTF::Vector<JSC::WriteBarrier<JSC::Unknown>>`. `NapiHandleScope` is an RAII helper class that opens a handle scope and closes it at the end of the C++ scope. `napi_handle_scope` and `napi_escapable_handle_scope` are both pointers to `NapiHandleScopeImpl` (which can be escapable or not).

A `NapiHandleScope` must be created whenever you start running a native function from a NAPI module:

https://github.com/oven-sh/bun/blob/7b1b0e3de8241dfc860baed64cefd3c10b8250d6/src/bun.js/bindings/BunProcess.cpp#L411-L414

https://github.com/oven-sh/bun/blob/7b1b0e3de8241dfc860baed64cefd3c10b8250d6/src/bun.js/bindings/napi.cpp#L441-L444

https://github.com/oven-sh/bun/blob/7b1b0e3de8241dfc860baed64cefd3c10b8250d6/src/napi/napi.zig#L1577-L1579

**Whenever you return a new `napi_value` to native code you must ensure it is stored in the current handle scope**. Otherwise it could get collected if the native function moves it off the stack. Use `toNapi` in C++ code or `napi_value.set`/`napi_value.create` in Zig code.

I wrote automated tests

    - JSValues in handle scopes are stored in write barriers under a GC-visible object stored on the global object.

### DIFF
--- a/src/bun.js/bindings/BunProcess.cpp
+++ b/src/bun.js/bindings/BunProcess.cpp
@@ -33,6 +33,8 @@
 
 #include "AsyncContextFrame.h"
 
+#include "napi_handle_scope.h"
+
 #ifndef WIN32
 #include <errno.h>
 #include <dlfcn.h>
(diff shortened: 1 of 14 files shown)


robobun:
@190n, your commit 7b1b0e3de8241dfc860baed64cefd3c10b8250d6 is building: [`#2870`](https://buildkite.com/bun/bun/builds/2870)

Jarred-Sumner:
Looks great

Issue #2873, opened 2023-05-13
incorrect DCE of unreferenced object literal


kzc:
### What version of Bun is running?

0.6.0 canary

### What platform is your computer?

macos 10.14

### What steps can reproduce the bug?

given:
```
$ cat obj-lit-dce.js

    ({
        0: 1,
        2: console.log("EFFECT1"),
    });
    ({
        0: 1,
        [console.log("EFFECT2")]: 2,
    });
```
expected:
```
$ node obj-lit-dce.js
EFFECT1
EFFECT2
```
actual:
```
$ bun obj-lit-dce.js
EFFECT2
```
```
$ bun run obj-lit-dce.js
EFFECT2
```

### Additional information

This `bun run` bug also exists in `bun build` and `bun build --minify`.

Jarred-Sumner:
Fixed in https://github.com/oven-sh/bun/commit/4d751db27a65c800c1c4cd6e6a241e360a9faa54

PR #13757, opened 2024-09-05
various node:buffer fixes

nektro:
individual commits contain fixes and new tests


robobun:
@nektro, your commit 0494ba2830546d101b5f627f23c9d54ad3bdccb5 is building: [`#2873`](https://buildkite.com/bun/bun/builds/2873)

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -487,6 +499,10 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_allocBody(JSC::JSG
 
     double lengthDouble = lengthValue.toIntegerWithTruncation(lexicalGlobalObject);
 
+    if (UNLIKELY(lengthDouble != lengthDouble)) {
+        throwNodeRangeError(lexicalGlobalObject, scope, "Buffer size must be 0...9007199254740991, got NaN"_s);
```
Jarred-Sumner: can just make throwRangeError do this instead of adding another

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -930,16 +914,13 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunction_compareBody(JSC::JSG
...
-    }
 
-    JSC::JSUint8Array* view = JSC::jsDynamicCast<JSC::JSUint8Array*>(callFrame->uncheckedArgument(0));
+    auto arg0 = callFrame->argument(0);
+    JSC::JSUint8Array* view = JSC::jsDynamicCast<JSC::JSUint8Array*>(arg0);
 
     if (UNLIKELY(!view)) {
-        throwVMTypeError(lexicalGlobalObject, throwScope, "Expected Uint8Array"_s);
-        return JSValue::encode(jsUndefined());
+        throwScope.throwException(lexicalGlobalObject, Bun::Bun__ERR_INVALID_ARG_TYPE_static3(lexicalGlobalObject, "target"_s, "Buffer or Uint8Array"_s, arg0));
```
Jarred-Sumner: we can use template <...Args pattern for this

Jarred-Sumner: e.g. from webkit:
```
    template <typename ...Args>
    NEVER_INLINE UnexpectedResult WARN_UNUSED_RETURN fail(Args... args) const
    {
        using namespace FailureHelper; // See ADL comment in namespace above.
        return UnexpectedResult(makeString("WebAssembly.Module doesn't parse at byte "_s, m_offset, ": "_s, makeString(args)...));
    }
    ```

File: src/bun.js/modules/NodeBufferModule.h [hunk 3]
```diff
@@ -169,15 +169,15 @@ DEFINE_NATIVE_MODULE(NodeBuffer) {
       JSC::jsNumber(MAX_ARRAY_BUFFER_SIZE));
 
   put(JSC::Identifier::fromString(vm, "kStringMaxLength"_s),
-      JSC::jsNumber(std::numeric_limits<unsigned>().max()));
```
Jarred-Sumner, line 172: in JSC, the max string length is unsigned int not int

src/bun.js/modules/NodeBufferModule.h
```diff
@@ -169,15 +169,15 @@ DEFINE_NATIVE_MODULE(NodeBuffer) {
...
+      JSC::jsNumber(std::numeric_limits<int>().max()));
 
   JSC::JSObject *constants = JSC::constructEmptyObject(
       lexicalGlobalObject, globalObject->objectPrototype(), 2);
   constants->putDirect(vm, JSC::Identifier::fromString(vm, "MAX_LENGTH"_s),
                        JSC::jsNumber(MAX_ARRAY_BUFFER_SIZE));
   constants->putDirect(vm,
                        JSC::Identifier::fromString(vm, "MAX_STRING_LENGTH"_s),
-                       JSC::jsNumber(std::numeric_limits<unsigned>().max()));
```
Jarred-Sumner, line 180: in JSC, the max string length is unsigned int not int

test/js/node/buffer.test.js
```diff
@@ -2003,9 +2003,9 @@ for (let withOverridenBufferWrite of [false, true]) {
 
       it("constants", () => {
         expect(BufferModule.constants.MAX_LENGTH).toBe(4294967296);
-        expect(BufferModule.constants.MAX_STRING_LENGTH).toBe(4294967295);
```
Jarred-Sumner, line 2006: in JSC, the max string length is unsigned int not int

test/js/node/buffer.test.js
```diff
@@ -2003,9 +2003,9 @@ for (let withOverridenBufferWrite of [false, true]) {
...
+        expect(BufferModule.constants.MAX_STRING_LENGTH).toBe(2147483647);
         expect(BufferModule.default.constants.MAX_LENGTH).toBe(4294967296);
-        expect(BufferModule.default.constants.MAX_STRING_LENGTH).toBe(4294967295);
```
Jarred-Sumner, line 2008: in JSC, the max string length is unsigned int not int

src/bun.js/bindings/JSBuffer.cpp
nektro: these two variants have the same exact signature but a slightly different message so its much simpler to keep them separate for now

src/bun.js/bindings/JSBuffer.cpp
nektro: this is calling the same method with a different message, not a new function

src/bun.js/modules/NodeBufferModule.h
```diff
@@ -46,8 +46,10 @@ JSC_DEFINE_HOST_FUNCTION(jsBufferConstructorFunction_isUtf8,
     }
 
     if (UNLIKELY(impl->isDetached())) {
-      throwTypeError(lexicalGlobalObject, throwScope,
-                     "ArrayBuffer is detached"_s);
+      throwScope.throwException(
+          lexicalGlobalObject,
+          Bun::new_ERR_INVALID_STATE(lexicalGlobalObject,
```
Jarred-Sumner: Why not name it something more like
```
throwException(throwScope, Bun::ERR_INVALID_STATE(lexicalGlobalScope, "Cannot validate detached buffer"_s))
```

Jarred-Sumner: or

```c++
Bun::throwInvalidState(throwScope, lexicalGlobalObject,  "Cannot validate on a detached buffer"_s)
```

src/bun.js/bindings/headers-handwritten.h
```diff
@@ -263,6 +263,8 @@ typedef void WebSocketHTTPSClient;
 typedef void WebSocketClient;
 typedef void WebSocketClientTLS;
 
+#define MAX_STRING_LENGTH std::numeric_limits<unsigned>().max()
```
Jarred-Sumner:
```suggestion
#define MAX_STRING_LENGTH WTF::String::MaxLength
```

src/js/builtins/JSBufferConstructor.ts
```diff
@@ -29,6 +29,11 @@ export function from(items) {
       }
     }
   }
+  if (typeof items === "object") {
+    if (items.type === "Buffer" && Array.isArray(items.data)) {
+      return new $Buffer(Uint8Array.from(items.data).buffer);
```
Jarred-Sumner: This creates two Uint8Array.

We don't have to pay that performance overhead.

```suggestion
      return new $Buffer(items.data);
```

src/js/builtins/JSBufferPrototype.ts
```diff
@@ -317,13 +425,15 @@ export function writeUIntLE(this: BufferExt, value, offset, byteLength) {
...
-    default: {
-      throw new RangeError("byteLength must be >= 1 and <= 6");
-    }
   }
   return offset + byteLength;
 }
+
 export function writeUIntBE(this: BufferExt, value, offset, byteLength) {
+  require("internal/validators").validateIntRange(byteLength, "byteLength", 1, 6);
+  require("internal/validators").validateIntRange(offset, "offset", 0, this.byteLength - byteLength);
```
Jarred-Sumner: `DataView` does it's own validation

Do we need to validate it twice?

src/js/builtins/JSBufferPrototype.ts
```diff
@@ -353,49 +463,66 @@ export function writeUIntBE(this: BufferExt, value, offset, byteLength) {
...
-      throw new RangeError("byteLength must be >= 1 and <= 6");
-    }
   }
   return offset + byteLength;
 }
 
 export function writeFloatLE(this: BufferExt, value, offset) {
+  if (offset === undefined) offset = 0;
+  require("internal/validators").validateInteger(offset, "offset", 0, this.byteLength - this.byteOffset - 4);
+  if (this.byteOffset + offset + 4 > this.byteLength) throw require("internal/errors").ERR_BUFFER_OUT_OF_BOUNDS();
```
Jarred-Sumner: Ditto here in all of these re: validation

Also, when you see us access the same property more than once, it's better to assign it to a local variable.

src/js/internal/validators.ts
```diff
@@ -0,0 +1,41 @@
+const { ERR_INVALID_ARG_TYPE, ERR_OUT_OF_RANGE, ERR_BUFFER_OUT_OF_BOUNDS } = require("internal/errors");
+
+export default {
+  validateIntRange: function (value, name, min, max) {
```
Jarred-Sumner: Isn't there an implementation in native code already?

src/js/builtins/JSBufferPrototype.ts
nektro: some but not all and none of the errors have codes

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -990,39 +960,44 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunction_compareBody(JSC::JSG
     switch (callFrame->argumentCount()) {
     default:
```
Jarred-Sumner, line 961: does this always go to the `default` value?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2223,25 +2186,43 @@ static const HashTableValue JSBufferPrototypeTableValues[]
...
-          { "writeUint32BE"_s, static_cast<unsigned>(JSC::PropertyAttribute::Builtin), NoIntrinsic, { HashTableValue::BuiltinGeneratorType, jsBufferPrototypeWriteUInt32BECodeGenerator, 1 } },
-          { "writeUint32LE"_s, static_cast<unsigned>(JSC::PropertyAttribute::Builtin), NoIntrinsic, { HashTableValue::BuiltinGeneratorType, jsBufferPrototypeWriteUInt32LECodeGenerator, 1 } },
-          { "writeBigUint64BE"_s, static_cast<unsigned>(JSC::PropertyAttribute::Builtin), NoIntrinsic, { HashTableValue::BuiltinGeneratorType, jsBufferPrototypeWriteBigUInt64BECodeGenerator, 1 } },
-          { "writeBigUint64LE"_s, static_cast<unsigned>(JSC::PropertyAttribute::Builtin), NoIntrinsic, { HashTableValue::BuiltinGeneratorType, jsBufferPrototypeWriteBigUInt64LECodeGenerator, 1 } },
       };
 
+#define ALIAS(to, from)                                                                          \
+    do {                                                                                         \
+        auto original_ident = Identifier::fromString(vm, ASCIILiteral::fromLiteralUnsafe(from)); \
```
Jarred-Sumner, line 2193:
```suggestion
        const auto original_ident = Identifier::fromString(vm, from ""_s); \
```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2223,25 +2186,43 @@ static const HashTableValue JSBufferPrototypeTableValues[]
...
+        auto original = this->getDirect(vm, original_ident);                                     \
+        auto alias_ident = Identifier::fromString(vm, ASCIILiteral::fromLiteralUnsafe(to));      \
```
Jarred-Sumner, line 2195:
```suggestion
        const auto alias_ident = Identifier::fromString(vm, to ""_s);      \
```

src/js/builtins/JSBufferConstructor.ts
```diff
@@ -29,6 +29,11 @@ export function from(items) {
       }
     }
   }
+  if (typeof items === "object") {
+    if (items.type === "Buffer" && Array.isArray(items.data)) {
```
Jarred-Sumner, line 33:
```suggestion
    const data = items.data;
    if (items.type === "Buffer" && Array.isArray(data)) {
```

src/js/builtins/JSBufferConstructor.ts
```diff
@@ -29,6 +29,11 @@ export function from(items) {
...
+      return new $Buffer(items.data);
```
Jarred-Sumner, line 34:
```suggestion
      return new $Buffer(data);
```

src/js/builtins/JSBufferPrototype.ts
```diff
@@ -9,45 +9,79 @@ interface BufferExt extends Buffer {
 }
 
 export function setBigUint64(this: BufferExt, offset, value, le) {
+  if (offset === undefined) offset = 0;
```
Jarred-Sumner, line 12:
```suggestion
  offset ??= 0;
```

src/js/builtins/JSBufferPrototype.ts
```diff
@@ -9,45 +9,79 @@ interface BufferExt extends Buffer {
...
   return (this.$dataView ||= new DataView(this.buffer, this.byteOffset, this.byteLength)).setBigUint64(
     offset,
     value,
     le,
   );
 }
+
 export function readInt8(this: BufferExt, offset) {
+  if (offset === undefined) offset = 0;
```
Jarred-Sumner, line 21:
```suggestion
  offset ??= 0;
```

src/js/builtins/JSBufferPrototype.ts
```diff
@@ -9,45 +9,79 @@ interface BufferExt extends Buffer {
...
+  require("internal/validators").validateInteger(offset, "offset", 0);
   return (this.$dataView ||= new DataView(this.buffer, this.byteOffset, this.byteLength)).getInt8(offset);
 }
+
 export function readUInt8(this: BufferExt, offset) {
+  if (offset === undefined) offset = 0;
```
Jarred-Sumner, line 27:
```suggestion
  offset ??= 0;
```

src/js/builtins/JSBufferPrototype.ts
nektro: `a ??= b` also assigns if `a` is `null` which i do not want to do in this case (its emulating setting a default parameter value)

src/js/builtins/JSBufferPrototype.ts
nektro: see above

src/js/builtins/JSBufferPrototype.ts
nektro: see above

src/bun.js/bindings/JSBuffer.cpp
nektro: only then there's 5 or more arguments

src/bun.js/bindings/JSBuffer.cpp
nektro: `_s` calls `ASCIILiteral::fromLiteralUnsafe` though so doesn't this add a step?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2223,25 +2186,43 @@ static const HashTableValue JSBufferPrototypeTableValues[]
...
+        this->putDirect(vm, alias_ident, original, PropertyAttribute::Builtin | 0);              \
+    } while (false);
+
 void JSBufferPrototype::finishCreation(VM& vm, JSC::JSGlobalObject* globalThis)
 {
     Base::finishCreation(vm);
     JSC_TO_STRING_TAG_WITHOUT_TRANSITION();
     reifyStaticProperties(vm, JSBuffer::info(), JSBufferPrototypeTableValues, *this);
+
+    ALIAS("readUintBE", "readUIntBE");
```
Jarred-Sumner, line 2205: We should update the static hashtables to support this as a PropertyAttribute or something

Issue #2915, opened 2023-05-17
Support self signed certificate


rvitorgomes:
### What is the problem this feature would solve?

I'm trying to use edgedb javascript client (edgeql-js) but it fails to connect to my local edge server.

error: network error: DEPTH_ZERO_SELF_SIGNED_CERT: self signed certificate
 code: "4278256384"
 
 Version: 
 "edgedb": "^1.2.2",
 "bun": "0.6.1",
 
 My code:
 ```typescript
 import { Elysia } from 'elysia'
import { createClient } from "edgedb"

const client = createClient({
    // dsn: "edgedb://edgedb@localhost:5656/edgedb"
})
await client.ensureConnected()

const app = new Elysia()
    .get('/', () => 'Hello Elysia 2')
    .get('/e', async () => {
        return await client.query("select 2 + 2");
    })
    .listen(3000)

console.log(`🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`)

 ```
 
 The error:
 
![image](https://github.com/oven-sh/bun/assets/14111893/30991dd9-ff26-4aa9-ad7f-cedfc28c963a)

### What is the feature you are proposing to solve the problem?

Seems to be a network package missing feature.

t0xicCode:
This seems more like an EdgeDB issue: https://www.edgedb.com/docs/reference/connection#granular-parameters
It provides a method to disable certificate validation, which would allow it to accept self-signed certificates.

rvitorgomes:
Update:
I ran node with the same code and worked
Also tried disable all security, but it still shows an error.
_The server doesn't support the edgedb-binary protocol_
```typescript
export const client = createClient({
   tlsSecurity: "insecure" 
});
```
What I'm trying to reason now is about their "binary protocol" has something that is missing....
I run node with the same code and worked

https://www.edgedb.com/docs/reference/protocol/index

cirospaciari:
@rvitorgomes Boa tarde, poderia fazer um teste versão `v0.6.3`? foram feitas algumas melhorias, talvez funcione agora.

rvitorgomes:
Tried with this example:
```typescript
// example.ts
const e = require('edgedb');

(async () => {
    const client = e.createClient({ 
    });
    const res = await client.query("select 2 + 2");
    process.stdout.write(res.toString());
    process.stdout.write('\n');
})();

In node v17:.
ts-node example.ts

```
[...]

PR #13763, opened 2024-09-06
Add if workers spawned / terminated to crash reports

Jarred-Sumner:
### What does this PR do?

Add if workers spawned / terminated to crash reports


robobun:
@Jarred-Sumner, your commit c6dc45195f7db5959baf4603d6427df374dce54e is building: [`#2915`](https://buildkite.com/bun/bun/builds/2915)

PR #13765, opened 2024-09-06
fix bad translation in child-process-exec-timeout-kill.test.js

nektro:
this block was unneeded and created false negative errors in CI


robobun:
@nektro, your commit 996b1b7239b06066e9c8a955cf5aa26311d08450 is building: [`#2920`](https://buildkite.com/bun/bun/builds/2920)

PR #11770, opened 2024-06-11
node:zlib: move deflate and gzip into native code

nektro:
deleted a bunch of javascript, add a bunch of new test coverage inherited from node, added much more rigorous error handling, wrote new native zlib usage wrappers in order to allow the compression and decompression to be done while streaming

buildkite comment: https://github.com/oven-sh/bun/pull/11770#issuecomment-2194218473

----

performance impact test via https://github.com/timotejroiko/zlib-benchmark

left group is macos arm64, right group is linux x86_64

| Benchmark / MB/s | :node: 22.5.1 | :bun: v1.1.26 | 73324ec |     | 73324ec |
| ---------------- | ------------- | ------------- | ------- | --- | ------- |
| DeflateSync      | 28.809        | 9.887         | 36.422  |     | 14.662  |
| InflateSync      | 78.703        | 38.075        | 86.362  |     | 96.499  |
| DeflateRawSync   | 28.500        | 9.882         | 36.160  |     | 14.583  |
| InflateRawSync   | 77.665        | 40.182        | 83.726  |     | 94.135  |
| GzipSync         | 28.630        | 9.583         | 34.694  |     | 14.620  |
| GunzipSync       | 82.306        | 38.002        | 86.125  |     | 100.105 |
| BrotliCompress   | 0.564         | 0.697         | 0.708   |     | 0.723   |
| BrotliDecompress | 49.612        | 48.868        | 47.763  |     | 42.910  |
| DeflateStream    | 13.926        | 14.737        | 30.240  |     | 33.115  |
| InflateStream    | 20.993        | 49.081        | 73.105  |     | 90.202  |
| DeflateRawStream | 13.991        | 15.032        | 30.467  |     | 33.169  |
| InflateRawStream | 20.907        | 50.720        | 72.245  |     | 90.010  |
| GzipStream       | 14.663        | 14.617        | 30.395  |     | 32.986  |
| GunzipStream     | 24.630        | 42.511        | 70.998  |     | 86.706  |

PR #13787, opened 2024-09-07
Fix 2 memory leaks in zlib from #11770

Jarred-Sumner:
### What does this PR do?

Fix 2 memory leaks in zlib from #11770


robobun:
@Jarred-Sumner, your commit f6296780304446dc0d10c6d6d26335ef95747d03 is building: [`#3092`](https://buildkite.com/bun/bun/builds/3092)

PR #13800, opened 2024-09-08
Fix default value for zlib options

Jarred-Sumner:
### What does this PR do?

- Fix default value for zlib options so that `undefined` is not treated as invalid
- Fix incorrect validation logic for integer ranges. Previously, in certain cases it would only look at numbers like Infinity or -Infinity instead of numbers outside the bounds of the expected number.
- Fix an integer overflow when keeping the event loop alive for brotli streams

This also adds an `bun.fmt.outOfRange(value, .{ .field_name = "", .min = 123, .max = 456 })` formatter and makes the formatters for the out of range in JS stuff a little less repetitive

A couple new tests

### DIFF
--- a/src/bun.js/ConsoleObject.zig
+++ b/src/bun.js/ConsoleObject.zig
@@ -3157,7 +3157,7 @@ pub const Formatter = struct {
             "<r><d>, ... {d} more<r>";
 
         writer.print(comptime Output.prettyFmt(fmt_, enable_ansi_colors), .{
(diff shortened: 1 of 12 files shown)


robobun:
@Jarred-Sumner, your commit 1a9f3e509d920b244ac56149518fd3a8d4eecd7a is building: [`#3123`](https://buildkite.com/bun/bun/builds/3123)

nektro (approved):

PR #13807, opened 2024-09-08
feat(node:util): support array type in format for the styleText util

pavi2410:
### What does this PR do?

This adds support for format array type in the `styleText` function from the `node:util` module. This feature is already [https://nodejs.org/api/util.html#utilstyletextformat-text-options](available) in Node.js

- [x] Code changes

I wrote automated tests matching those from Node.js


paperclover (approved):

paperdave:
implementation looks reasonable, but the test is failing.

image

pavi2410:
I have now removed a useless test case that was not supposed to fail.

PR #13835, opened 2024-09-09
workerData should default to null instead of undefined

Jarred-Sumner:
### What does this PR do?

workerData should default to null instead of undefined

Updated test


robobun:
@Jarred-Sumner, your commit c69ff83cc7b2ad1926adbd7bf5085512b59314e1 is building: [`#3156`](https://buildkite.com/bun/bun/builds/3156)

Issue #3195, opened 2023-06-03
mongoose logs are too verbose


emastho:
### What version of Bun is running?

0.6.7

### What platform is your computer?

Linux 6.3.5-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

Here's a recreation
https://www.youtube.com/watch?v=w93qIzdeJWc

<br />

Too verbose
```ts
model.create().then(console.log)
```

What it should be
```ts
model.create().then(x => console.log(`${res}`))
```
<br />
Writing res into a file also gives expected output

### What is the expected behavior?

```ts
{
  name: 'John',
  age: 54,
  _id: new ObjectId("647b2b30afd486ff87df1424"),
  __v: 0
}
```

### What do you see instead?

<details>
<summary>It's very long, so click to open it</summary>

```ts
                  ],
                  connectionCheckedOut: [
                    [Function]
                  ],
                  connectionCheckedIn: [
                    [Function]
                  ],
                  descriptionReceived: [
                    [Function]
                  ]
                },
                _eventsCount: 18,
                _maxListeners: undefined,
                [Symbol(kCapture)]: false,
                serverApi: undefined,
                topology: [Circular],
                pool: {
                  _events: [Object ...],
                  _eventsCount: 15,
                  _maxListeners: undefined,
                  [Symbol(kCapture)]: false,
                  options: [Object ...],
                  [Symbol(poolState)]: "ready",
                  [Symbol(server)]: [Circular],
                  [Symbol(connections)]: [List ...],
                  [Symbol(pending)]: 0,
                  [Symbol(checkedOut)]:                   Set {},
                  [Symbol(minPoolSizeTimer)]: undefined,
                  [Symbol(generation)]: 0,
                  [Symbol(serviceGenerations)]: Map {},
                  [Symbol(connectionCounter)]: {},
                  [Symbol(cancellationToken)]: [Object ...],
                  [Symbol(waitQueue)]: [List ...],
                  [Symbol(metrics)]: [Object ...],
                  [Symbol(processingWaitQueue)]: false,
                  mongoLogger: [Object ...],
                  component: "connection",
                  address: [Getter],
                  closed: [Getter],
                  generation: [Getter],
                  totalConnectionCount: [Getter],
                  availableConnectionCount: [Getter],
                  pendingConnectionCount: [Getter],
                  currentCheckedOutCount: [Getter],
                  waitQueueSize: [Getter],
                  loadBalanced: [Getter],
[...]

PR #13843, opened 2024-09-10
Add generator for $ERR_* as fake private globals

Jarred-Sumner:
### What does this PR do?

Add generator for $ERR_* as fake private globals

This lets you do anywhere in builtins/modules:
```ts
throw $ERR_STRING_TOO_LONG("The string is too long!!");
```

This becomes
```ts
throw @makeErrorWithCode(11, "The string is too long!!")
```

And we can customize the specific message like so:

https://github.com/oven-sh/bun/blob/46e9a69b4087ee9c6f52ddfff61c291fb46f49a8/src/bun.js/bindings/ErrorCode.cpp#L526-L538

It will also add these `$ERR_*`  to the typescript declarations so it's easier to figure out they exist

I replaced the usages of `new ERR_INVALID_ARG_TYPE` in node:readline with `$ERR_INVALID_ARG_TYPE`. I had to update the tests since we were setting class names for these.


robobun:
@Jarred-Sumner, your commit 46e9a69b4087ee9c6f52ddfff61c291fb46f49a8 is building: [`#3195`](https://buildkite.com/bun/bun/builds/3195)

paperclover (approved):

PR #13844, opened 2024-09-10
Enable SIMD multiline comment optimization

Jarred-Sumner:
### What does this PR do?

Enable SIMD multiline comment optimization from a couple years ago now that we have more tests

First let's run it with assertions enabled on Windows

Then if the tests look right, then let's measure if there's any performance difference


robobun:
@Jarred-Sumner, your commit 46f723fdfba86d9212ba117eb7dd424e9b922d52 is building: [`#3198`](https://buildkite.com/bun/bun/builds/3198)

Jarred-Sumner:
If we copypaste the contents of `bun.d.ts` approximately 77 times (20 MB of source code), we see a very measurable speed improvement:
```
❯ hyperfine "bun build --no-bundle packages/bun-types/bun.d.ts" "bun-13844 build --no-bundle packages/bun-types/bun.d.ts" --warmup=20 -N
Benchmark 1: bun build --no-bundle packages/bun-types/bun.d.ts
  Time (mean ± σ):      78.6 ms ±   4.4 ms    [User: 70.4 ms, System: 5.6 ms]
  Range (min … max):    75.8 ms …  99.0 ms    39 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: bun-13844 build --no-bundle packages/bun-types/bun.d.ts
  Time (mean ± σ):      42.3 ms ±   0.8 ms    [User: 35.5 ms, System: 5.0 ms]
  Range (min … max):    40.9 ms …  44.2 ms    69 runs

Summary
  bun-13844 build --no-bundle packages/bun-types/bun.d.ts ran
    1.86 ± 0.11 times faster than bun build --no-bundle packages/bun-types/bun.d.ts
```

We do not see a slowdown

PR #13848, opened 2024-09-10
Fix edgecase with "os" and "cpu" in bun install

Jarred-Sumner:
### What does this PR do?

A `"cpu"` field in package.json like `["!ia32", "!arm"]` was treated as equivalent to specifying no CPU is ever supported, which means the package would never be installed. This is incorrect.

We were also not handling `"any"` correctly.

References:
- https://github.com/pnpm/pnpm/blob/1f228b0aeec2ef9a2c8577df1d17186ac83790f9/config/package-is-installable/src/checkPlatform.ts#L56-L86
- https://github.com/npm/cli/blob/fefd509992a05c2dfddbe7bc46931c42f1da69d7/node_modules/npm-install-checks/lib/index.js#L2-L96

Now we get all the fields that were added and all the fields that were removed, and only combine the two if both were present

This bumps the manifest cache version because the data needs to be regenerated since we were storing incorrect values for `"os"`, `"cpu"`. Technically, we need to also bump the lockfiles to handle this correctly...which is not great. I'm not doing that in this PR.

There are unit tests

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -12924,7 +12924,7 @@ pub const PackageManager = struct {
                 if (comptime log_level.isVerbose()) {
                     const name = this.lockfile.str(&this.names[package_id]);
                     if (!meta.os.isMatch() and !meta.arch.isMatch()) {
-                        Output.prettyErrorln("<d>Skip installing '<b>{s}<r><d>' os & cpu mismatch", .{name});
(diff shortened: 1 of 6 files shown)


robobun:
@Jarred-Sumner, your commit 26c3cdd138f4fe3234ea7b2c365ce9243df2565d is building: [`#3204`](https://buildkite.com/bun/bun/builds/3204)

dylan-conway (approved):

Issue #3250, opened 2023-06-08
memory leak using crypto sha256 hashing


add1ed:
### What version of Bun is running?

0.6.7

### What platform is your computer?

Darwin 22.4.0 arm64 arm

### What steps can reproduce the bug?

Run the following code:
```
const crypto = require('crypto');

function sha256(buf) {
  return crypto.createHash('sha256').update(buf).digest();
};

async function main() {
  for (var i = 1000000; i >= 0; i--) {
    const buf = Buffer.alloc(2046);
    const hash = sha256(buf);
    if (i % 1000 === 0) {
      await new Promise(r => setTimeout(r, 20));
      global.gc ? global.gc() : Bun?.gc(true);
    }
  }
}

main();
```
using `bun index.js` and watch the memory usage, as reported by the OS (in my case looking at Activity Monitor).

(And repeat using `node --expose-gc index.js` to observe the equivalent behaviour in nodejs.)

### What is the expected behavior?

Memory usage stays flat.

### What do you see instead?

Memory usage grows.

### Additional information

Outputting `heapStats()` data from `bun:jsc` at various points during the looping indicates heap memory stays flat.

Memory seems to approx grow at ~130 bytes per hash call.

I have observed, in my own application, that over the course of an hour, in which ~7.2M hashes are performed, memory usage grows by ~1GB.

Jarred-Sumner:
Thank you for the report.

After:
<img width="451" alt="image" src="https://github.com/oven-sh/bun/assets/709451/d111274a-0a4a-4075-82f2-39699dbf8521">

Before:
<img width="444" alt="image" src="https://github.com/oven-sh/bun/assets/709451/231f7e38-0717-4610-a8dd-aebac2aa70d7">

Node.js, for comparison:
<img width="454" alt="image" src="https://github.com/oven-sh/bun/assets/709451/22e4466e-883e-4659-9f5c-f0310865515b">

BinaryFox83:
> Thank you for the report.
> 
> After: <img alt="image" width="451" src="https://user-images.githubusercontent.com/709451/244854940-d111274a-0a4a-4075-82f2-39699dbf8521.png">
> 
> Before: <img alt="image" width="444" src="https://user-images.githubusercontent.com/709451/244854950-231f7e38-0717-4610-a8dd-aebac2aa70d7.png">
> 
[...]

PR #13870, opened 2024-09-11
Avoid creating a Napi handle scope within a finalizer

Jarred-Sumner:
### What does this PR do?

This makes napi handle scopes optional, and we ignore them when they're not set.

Fixes https://discord.com/channels/876711213126520882/1282959007589601280/1282959007589601280

This mitigates a crash where a napi package creates a handle scope within a finalizer. Any object allocation within that finalizer would still fail, but the act of creating a handle scope at all inside a finalizer fails since it is an allocation. 

I verified the crash is fixed manually.

This needs a test. The test would do something like:
- Create a napi ref wrapping an object
- The napi ref's finalizer function should open a handle scope
- Allocate repeatedly until the napi ref's finalizer is called


robobun:
@Jarred-Sumner, your commit 7cb2bad7999529e20e9ce18af8545321720d80d0 is building: [`#3250`](https://buildkite.com/bun/bun/builds/3250)

190n:
It looks like nodejs/node-addon-api#1409 will (once it lands in a release) also fix this, as they now [intentionally defer the execution of finalizers instead of relying on Node.js to do it](https://github.com/nodejs/node-addon-api/pull/1409/files#diff-cd0b229bcaf532206661e2ee1a603a7f60ac22ecfcb239920f6b42a34d909f34R34-R50):

```cpp
#if (defined(NAPI_EXPERIMENTAL) &&                                             \
     defined(NODE_API_EXPERIMENTAL_HAS_POST_FINALIZER))
template <napi_finalize finalizer>
inline void PostFinalizerWrapper(node_api_nogc_env nogc_env,
                                 void* data,
                                 void* hint) {
  napi_status status = node_api_post_finalizer(nogc_env, finalizer, data, hint);
  NAPI_FATAL_IF_FAILED(
      status, "PostFinalizerWrapper", "node_api_post_finalizer failed");
}
#else
template <napi_finalize finalizer>
inline void PostFinalizerWrapper(napi_env env, void* data, void* hint) {
  finalizer(env, data, hint);
}
#endif
```

But this solution will do for now. I'll write a test -- I might see if we can pass a JS callback into the NAPI functions that calls `Bun.gc` instead of trying to trigger it via a bunch of allocations, as that will make the other GC tests faster too. But I'll have to make sure that that does actually reproduce GC bugs.

190n (changes requested):
needs a test

190n (approved):

190n (changes requested):
the tests crash node

Issue #3274, opened 2023-06-10
bun install nativescript does not work


scalpel4k:
### What version of Bun is running?

v0.6.8

### What platform is your computer?

Linux 6.3.4-1-default x86_64 x86_64

### What steps can reproduce the bug?

bun install -g nativescript

### What do you see instead?

error: package "cli-table" with tag "https://github.com/telerik-boneyard/cli-table/tarball/v0.3.1.2" not found, but package exists

error: package "marked-terminal" with tag "https://github.com/NativeScript/marked-terminal/tarball/v3.1.1n" not found, but package exists

error: package "bplist-parser" with tag "https://github.com/telerik/node-bplist-parser/tarball/master" not found, but package exists

error: package "tabtab" with tag "https://github.com/Icenium/node-tabtab/tarball/master" not found, but package exists

error: package "zipstream" with tag "https://github.com/Icenium/node-zipstream/tarball/master" not found, but package exists

### Additional information

I suspect bun is parsing some dependencies incorrectly ...

thx for looking into it

TiranexDev:
Yep, I can confirm this actually fails on my pc and some vms.
![image](https://github.com/oven-sh/bun/assets/94789999/7e0e300f-8809-41b6-bd93-6f6fdac5d078)

birkskyum:
still breaks.

keisanng:
Same problem here on Linux

birkskyum:
I now see this working:

```sh
➜ bun install -g nativescript


bun add v1.0.18 (36c316a2)
 + @oven/zig@0.12.0-dev.163

 installed nativescript@8.6.3 with binaries:
  - tns
  - nativescript
  - nsc
  - ns


 607 packages installed [6.06s]
```

@keisanng , @scalpel4k , @TiranexDev  - do you see a difference to the better as well?

scalpel4k:
I can confirm, this is working now

PR #13875, opened 2024-09-11
Move EventLoopTask into a separate file because it causes clangd to crash

Jarred-Sumner:
### What does this PR do?

Move EventLoopTask into a separate file because it causes clangd to crash according to the stacktrace


robobun:
@Jarred-Sumner, your commit f648fd835f28b176c3d512986127bcec0d461247 is building: [`#3274`](https://buildkite.com/bun/bun/builds/3274)

PR #13890, opened 2024-09-11
Add --profile option to bun-lambda package

farhan-helmy:
### What does this PR do?

This adds a new flag --profile to bun-lambda package. This is useful for people that has multiple AWS CLI profile on their environment

cd packages/bun-lambda
bun run publish-layer --profile <profile-name>
verify the error if token expired or profile does not exists


farhan-helmy:
@Jarred-Sumner requesting for review, tq

packages/bun-lambda/scripts/build-layer.ts
```diff
@@ -42,6 +42,11 @@ export class BuildCommand extends Command {
       description: "If the layer should be public.",
       default: false,
     }),
+    profile: Flags.string({
+      description: "The AWS CLI profile to use.",
+      multiple: false,
+      default: "default",
```
hiett, line 48: I don't think this should be hardcoded as default, because another tool or the developer is likely setting the AWS default profile env var to something else. It's unlikely someone's actual default profile is named default.

Instead, if the value is not provided, can we omit the --profile from the ran child command? That way, the AWS CLI falls back to its internal default logic which other tools rely on.

PR #13898, opened 2024-09-11
Make NAPI garbage collection tests faster

190n:
### What does this PR do?

Some of the NAPI tests need to check that an object is not freed after garbage collection runs. Previously, they would perform lots of allocations to try to force GC to run. Now, they all take a callback from JS which in turn calls `Bun.gc` to deterministically run the collector. This makes those tests much faster.

I ran the new tests while reintroducing several GC-related bugs one at a time and each bug was consistently caught by all the tests that were intended to catch it.

### DIFF
--- a/test/napi/napi-app/main.cpp
+++ b/test/napi/napi-app/main.cpp
@@ -30,6 +30,10 @@ napi_value ok(napi_env env) {
   return result;
 }
 
+static void run_gc(const Napi::CallbackInfo &info) {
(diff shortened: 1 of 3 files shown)


robobun:
@190n, your commit 79cef0fe528ab5f75af151529f3244d89cadf74a is building: [`#3310`](https://buildkite.com/bun/bun/builds/3310)

Issue #7500, opened 2023-12-07
Bun.stdin.text() does not read complete input


obedm503:
### What version of Bun is running?

1.0.15

### What platform is your computer?

Linux 6.5.6-76060506-generic x86_64 x86_64

### What steps can reproduce the bug?

```ts
const lines: string[] = (await Bun.stdin.text()).split("\n");

console.log(lines.length);
```

then pipe in a [large file](https://github.com/obedm503/bun-stdin-bug-repro/blob/master/big.txt)

```sh
cat big.txt | bun run bug.ts
```

### What is the expected behavior?

The expected output is `141` 

### What do you see instead?

The real output is `30` 

### Additional information

Reproduction repo https://github.com/obedm503/bun-stdin-bug-repro

Jarred-Sumner:
Can you try again in the canary build of Bun?

obedm503:
The issue persists after `bun upgrade --canary`.

obedm503:
Happy to report that the bug is indeed fixed. Thanks @Jarred-Sumner and team.

PR #13900, opened 2024-09-11
Fix ReadFileUV not reading to the end of a non-regular file on Windows

190n:
### What does this PR do?

Previously, ReadFileUV would sometimes read only one chunk of data when reading a non-regular file (like stdin), rather than reading the whole file. Now, it does read the whole file, and the test for it is more consistent.

I made changes to the regression test for #7500 so that it consistently fails using the old version of Bun on a Windows test machine. Previously, that test would sometimes fail in CI but I couldn't get it to fail on my own machine. The new test consistently passes on the new ReadFile.zig code.

### DIFF
--- a/src/bun.js/webcore/blob/ReadFile.zig
+++ b/src/bun.js/webcore/blob/ReadFile.zig
@@ -730,7 +730,9 @@ pub const ReadFileUV = struct {
     }
 
     pub fn queueRead(this: *ReadFileUV) void {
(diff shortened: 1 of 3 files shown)


robobun:
@190n, your commit 636cdc828a6551f9335ee93e33b7276f63fee464 is building: [`#3320`](https://buildkite.com/bun/bun/builds/3320)

Issue #12545, opened 2024-07-13
Allow `--conditions=browser` for `bun test`


mProjectsCode:
### What is the problem this feature would solve?

When running test using Svelte 5, bun will always import Sveltes server code since it prefers the export condition `default` over `brower`.

### What is the feature you are proposing to solve the problem?

Allow the use of `--conditions=browser`, like it already works for `bun run` to prefer other export conditions.

sameer-dudeja:
Hey, I would like to work on this but lack the knowledge know where to or how to start. Can you guide me a little in the right direction, if that's okay

mProjectsCode:
I currently have to patch the Svelte 5 `package.json` every time to remove the export conditions for Svelte SSR to be able to test my client-side code with HappyDOM.

Issue #10036, opened 2024-04-07
support --conditions in `bun test`


subtleGradient:
### What is the problem this feature would solve?

Working on code that requires `--conditions=react-server` to run.
Currently there is no way to use `bun test` to test this code.
Adding support for `--conditions` to `bun test` would allow me to write tests

### What is the feature you are proposing to solve the problem?

add support for `--conditions` to `bun test`

### What alternatives have you considered?

Currently looking into having my test spin up a separate instance of bun (using `--conditions=react-server` in the separate process) and then communicating between the process.
I'm not excited about doing this seemingly unnecessary work

subtleGradient:
related to #8990

Jarred-Sumner:
Fixed by @snoglobe in #13902

PR #13902, opened 2024-09-11
fixes --conditions for bun test

snoglobe:
### What does this PR do?

this PR fixes https://github.com/oven-sh/bun/issues/12545 and https://github.com/oven-sh/bun/issues/10036 - both of which described an issue where the `--conditions` flag did not have any effect when running`bun test`.

- [x] Code changes

I wrote automated tests

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -567,7 +567,7 @@ pub const Arguments = struct {
 
         ctx.passthrough = args.remaining();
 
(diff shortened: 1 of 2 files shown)


paperclover (approved):

Jarred-Sumner:
Maybe you could add a specific test for the `"browser"` condition? We have some speical-casing for it in `bun build` so it would be good to make sure it's handled properly when passed as a custom condition

PR #13909, opened 2024-09-12
fix(watch): use case insensitive path comparison

dylan-conway:
### What does this PR do?
Hopefully will fix `watch.test.ts` and `hot.test.ts` windows CI test failures.

- fixes an off by one in `windowsFilesystemRootT` when receiving absolute UNC paths
- handles converting absolute UNC paths to nt paths
- case insensitive paths comparison windows file/directory watcher

existing tests


src/windows.zig
```diff
@@ -3420,8 +3421,12 @@ pub fn GetFinalPathNameByHandle(
 
     bun.sys.syslog("GetFinalPathNameByHandleW({*p}) = {}", .{ hFile, bun.fmt.utf16(ret) });
 
-    if (ret.len > 4 and std.mem.eql(u16, ret[0..4], &.{ '\\', '\\', '?', '\\' })) {
+    if (bun.strings.hasPrefixComptimeUTF16(ret, "\\\\?\\")) {
```
paperclover: `- \\\\?\\`
`+ nt_maxpath_prefix`
?

dylan-conway: updated, also added comment

src/resolver/resolve_path.zig
```diff
@@ -74,7 +74,13 @@ pub fn isParentOrEqual(parent_: []const u8, child: []const u8) ParentEqual {
     while (parent.len > 0 and isSepAny(parent[parent.len - 1])) {
         parent = parent[0 .. parent.len - 1];
     }
-    if (std.mem.indexOf(u8, child, parent) != 0) return .unrelated;
+
+    const contains = if (comptime !bun.Environment.isLinux)
```
Jarred-Sumner, line 78: this is not technically correct

but it's probably okay enough for now.

case-sensitivity differs by filesystem - not by platform. it's just that on linux, most filesystems are case sensitive. on macos, the two most popular ones default to case insensitive, but HFS+ can be case sensitive if chosen. I don't know if Windows has an option for this.

src/fs.zig
```diff
@@ -617,7 +614,7 @@ pub const FileSystem = struct {
             var existing = this.entries.atIndex(index) orelse return null;
             if (existing.* == .entries) {
                 if (existing.entries.generation < generation) {
-                    var handle = bun.openDirA(std.fs.cwd(), existing.entries.dir) catch |err| {
+                    var handle = bun.openDirForIteration(std.fs.cwd(), existing.entries.dir) catch |err| {
```
Jarred-Sumner, line 617: I feel like there was some reasno for this

Jarred-Sumner:
merging since this reduces the number of test failures

PR #13929, opened 2024-09-12
Make setInterval leak test not flaky

190n:
### What does this PR do?

Changes `setInterval-leak-fixture.js` to not be flaky. Previously, it was calling `Bun.gc()` in one place instead of `Bun.gc(true)`. Since `Bun.gc()` with no argument runs GC asynchronously, this let the test perform more allocations in between forced GC runs, which caused RSS to grow after the warmup period even though nothing was actually leaking.

The test has also been changed to make control flow clearer. Each call to `batch()` is now a full run; previously, since the global `promise` and `resolve` were reused, subsequent calls to `batch()` would `await` on an already-resolved promise and clear all intervals immediately before any of them started running. Now each call creates 1,000 timers and waits for a total of 10,000 invocations. The timer callback also allocates more memory to make leaks more obvious.

The RSS delta is now typically 0 or 1 megabyte, significantly below the threshold of 20 to be counted as a leak. I verified that the delta does not grow when increasing the number of iterations even more.

### DIFF
--- a/test/js/web/timers/setInterval-leak-fixture.js
+++ b/test/js/web/timers/setInterval-leak-fixture.js
@@ -1,7 +1,10 @@
 const delta = 1;
-const initialRuns = 5_000_000;
+const initialRuns = 10_000;
 let runs = initialRuns;
-var initial = 0;
+
+function usage() {
+  return process.memoryUsage.rss();
+}
 
 Promise.withResolvers ??= () => {
   let promise, resolve, reject;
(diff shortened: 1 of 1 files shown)


robobun:
@190n, your commit ca16ed6da96947cd90cb29788c7759028aa0d09a is building: [`#3406`](https://buildkite.com/bun/bun/builds/3406)

Issue #3418, opened 2023-06-26
Cannot use dynamic imports in bun repl


renhiyama:
### What version of Bun is running?

0.6.9

### What platform is your computer?

Linux 5.15.74.2-microsoft-standard-WSL2 #1 SMP Wed Nov 2 19:50:29 UTC 2022 x86_64 x86_64

### What steps can reproduce the bug?

Write `await import("node:process")` in bun repl

### What is the expected behavior?

It should return back the export

### What do you see instead?

```
> await import("node:process")
SyntaxError: Unexpected keyword 'import'
      at /home/ren/coding/frontend/$bun$repl.ts:4:106
```

Electroid:
`bun repl` is not an official command yet, you must be using the unofficial [bun-repl](https://github.com/jhmaster2000/bun-repl)

PR #13931, opened 2024-09-12
update webkit

### LINKED ISSUE #13896: JIT bug

1.1.27+267afa293

Darwin 23.6.0 arm64 arm

Extracted this one from [a larger project](https://github.com/wincent/masochist/tree/next).

`Queue.ts`:

```ts
type Entry<T extends {}> = {
  next: Entry<T> | null;
  value: T;
};

export default class Queue<T extends {}> {
  _head: Entry<T> | null;
  _tail: Entry<T> | null;
  _length: number;

  constructor(items?: Iterable<T>) {
    this._head = null;
    this._tail = null;
    this._length = 0;

    if (items) {
      for (const item of items) {
        this.enqueue(item);
      }
    }
  }

  enqueue(item: T) {
    const entry = {
      next: null,
      value: item,
    };
    if (this._tail) {
      this._tail.next = entry;
      this._tail = entry;
    } else {
      this._head = entry;
      this._tail = entry;
    }
    this._length++;
  }

  dequeue(): T | null {
    const entry = this._head;
    if (entry) {
      this._head = entry.next;
      this._length--;
      if (this._head === null) {
        this._tail = null;
      }
      return entry.value;
    } else {
      return null;
    }
  }
}
```

`__tests__/Queue.test.ts`:

```ts
import {describe, expect, it} from 'bun:test';

import Queue from '../Queue';

describe('Queue', () => {
  it('optionally takes a non-array iterable of items with which to seed the queue', () => {
    const queue = new Queue(new Set(['foo', 'bar', 'baz']));
    expect(queue.dequeue()).toBe('foo');
    expect(queue.dequeue()).toBe('bar');
    expect(queue.dequeue()).toBe('baz');
  });
});
```

Test should pass no matter how many times you run it.

### What do you see instead?

Test passes if you run it 4000 times:

```
$ bun test --rerun-each 4000
 3999 pass
 0 fail
 11997 expect() calls
Ran 3999 tests across 4000 files. [1.83s]
```

But not 5000 times:

```
$ bun test --rerun-each 5000
 4870 pass
 129 fail
 14739 expect() calls
Ran 4999 tests across 5000 files. [2.39s]
```

(ie. it passes _almost_ 5000 times, but fails the last few hundred times)

### Additional information

I put "Memory corruption" in the title but this might also be "JIT gone wrong". No idea 🤷🏻... In the real project from which this example was taken, I'm not running the test case 5000 times; I'm just running the test suite (which has a few hundred tests) once, and seeing a handful of tests that rely on the `Queue` class failing ([sometimes](https://github.com/wincent/masochist/actions/runs/10790964145/job/29927324793), but [not always](https://github.com/wincent/masochist/actions/runs/10791048459/job/29927611775)). If I add `console.log` statements in some places (but not in others), this is enough to make the bug go away, even if I log something unrelated (like `console.log('hi')`). Likewise, I can put an innocuous `else` clause or make some other change to the code, and that is (sometimes) enough to stop the bug from manifesting. Presumably it's changing the way things get JIT-ed or not-JIT-ed, or introducing a timing effect that somehow masks the problem.

Suspecting a regression in Bun, I went back and checked previous versions back as far as v1.18. Bisecting reveals v1.22 is the last good version (and prior versions were good too). v1.23 is the first bad one (and subsequent versions are bad too). Nothing stands out to me in [the changelog](https://bun.sh/blog/bun-v1.1.23), but [here's the diff from v1.22 to v1.23.](https://github.com/oven-sh/bun/compare/bun-v1.1.22..bun-v1.1.23)

dylan-conway:
### What does this PR do?
also adds debug and release build configurations for local webkit builds

closes #13896


robobun:
@dylan-conway, your commit e4a22f685d7cc27f0c38641b755c2734b5a2ee38 is building: [`#3418`](https://buildkite.com/bun/bun/builds/3418)

PR #13935, opened 2024-09-13
cpp: always return empty JSValue value after throwing exception

nektro:
all tests should still pass.


robobun:
@nektro, your commit 3ce6fccb241aea9b6743f17867fd472b00998915 is building: [`#3447`](https://buildkite.com/bun/bun/builds/3447)

src/bun.js/bindings/bindings.cpp
```diff
@@ -3147,7 +3147,7 @@ void JSC__JSPromise__rejectOnNextTickWithHandled(JSC__JSPromise* promise, JSC__J
             globalObject->m_asyncContextData.get()->getInternalField(0),
             promise,
             value);
-        RETURN_IF_EXCEPTION(scope, void());
+        RETURN_IF_EXCEPTION(scope, );
```
dylan-conway, line 3150: this doesn't syntax error?

src/bun.js/bindings/bindings.cpp
```diff
@@ -3389,7 +3389,7 @@ extern "C" void JSC__JSValue__putMayBeIndex(JSC__JSValue target, JSC__JSGlobalOb
 
     JSC::JSObject* object = JSC::JSValue::decode(target).asCell()->getObject();
     object->putDirectMayBeIndex(globalObject, JSC::PropertyName(identifier), JSC::JSValue::decode(value));
-    RETURN_IF_EXCEPTION(scope, void());
+    RETURN_IF_EXCEPTION(scope, );
```
dylan-conway, line 3392: same here

src/bun.js/bindings/bindings.cpp
```diff
@@ -5224,7 +5224,7 @@ static void JSC__JSValue__forEachPropertyImpl(JSC__JSValue JSValue0, JSC__JSGlob
                 iterating->methodTable()->getOwnPropertyNames(iterating, globalObject, properties, DontEnumPropertiesMode::Include);
             }
 
-            RETURN_IF_EXCEPTION(scope, void());
+            RETURN_IF_EXCEPTION(scope, );
```
dylan-conway, line 5227: and here

dylan-conway (approved):

src/bun.js/bindings/bindings.cpp
nektro: it does not
the space appeases the preprocessor for a second value
empty value makes `return value__;` emit `return;` which satisfies a void return

PR #13939, opened 2024-09-13
Support reading from user's .npmrc

robertshuford:
### What does this PR do?

This PR expands upon PR https://github.com/oven-sh/bun/pull/11979 to add support for reading [per-user](https://docs.npmjs.com/cli/v10/configuring-npm/npmrc#files) .npmrc files in addition to `per-project` .npmrc files.

Configuration in project .npmrc files will be combined with configuration in user .npmrc files.  If a specific configuration is defined in both, the project configuration is used.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+; registry=https://registry.npmjs.org/
+registry=https://npm.pkg.github.com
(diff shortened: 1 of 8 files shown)


robertshuford:
I see the test I added for this failing in CI.  I'll work to get that fixed before taking this PR out of draft.

robertshuford:
I'm going to work on the tests some more. I'll open a new PR later.

robertshuford:
I opened a new PR for this here: https://github.com/oven-sh/bun/pull/13990

Issue #3478, opened 2023-07-01
Macros on function arguments - built in macros


DanielMazurkiewicz:
### What is the problem this feature would solve?

I want a simple reference concept known from other programming languages as this will dramatically simplify two way bindings or simplify immediate mode GUIs for games etc...

### What is the feature you are proposing to solve the problem?

```ts
import { Ref } from 'bun:macros'

const htmlInput = (text: Ref<string>) => {
   const input = document.createElement("input")
    if (isRef(text)) {
        input.value = text.value
        input.onchange = (event) => {
             text.value = event.target.value
        }
    } else {
        input.value = text
    }
    return input;
}

// and then usage

let editable_title = ""
const person = {
    name: "Mr"
    surname: "X"
}

const html_title = htmlInput(editable_title);
const html_name = htmlInput(person.name);
const html_surname = htmlInput(person.surname);

```

The simplest transpilation approach would probably be supplying a setter and getter method like that:
```js
const html_title = htmlInput({
   set value (v) { editable_title = v }, 
   get value () { return editable_title },
   name: "editable_title"
});
const html_name = htmlInput({
   set value (v) { person.name = v}, 
   get value () { return person.name }, 
   parent: person
   name: "name"
});
const html_surname = htmlInput({
   set value (v) { person.surname = v }, 
   get value: () { person.surname }, 
   parent: person
   name: "surname"
});

```

To make it typescript compliant in bun:macros code it would look like that: 

```ts
interface Reference<TValue, TParent = any> {
    value: TValue
    parent?: TParent
    name: string
}

function isRef(value: unknown): value is Reference {
    // ...built in macro
}
export type Ref<T> = Reference<T> | T
```

Lastly it could throw errors when anything that can't be reference will be supplied (eg. const x = 1, inline values ...)

In similar way we could add other macro types to gather more information about passed argument eg. ArgInfo

NOTE: some of code above is not fully functional, it is just to show the idea

PR #13951, opened 2024-09-13
fix(sockets) upgradeTLS server sockets is not supported

cirospaciari:
Upgrading a server socket is not support for now, so disable until a full solution is in place.

### What does this PR do?

- [X] Code changes


robobun:
@cirospaciari, your commit 308c682703461772655e484a32865ceaafb2bafb is building: [`#3478`](https://buildkite.com/bun/bun/builds/3478)

Issue #3494, opened 2023-07-02
`afterAll` isn't triggered if there is no `test` at the same level


riywo:
### What version of Bun is running?

0.6.12

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 unknown

### What steps can reproduce the bug?

Run `bun test` with the test code below:

```javascript
describe("without test at the same level", () => {
  afterAll(() => console.log("afterAll without test at the same level"));

  describe("a", () => {
    test("b", () => {
      expect(true).toBe(true);
    });
  });
});

describe("with test at the same level", () => {
  afterAll(() => console.log("afterAll with test at the same level"));

  describe("a", () => {
    test("b", () => {
      expect(true).toBe(true);
    });
  });

  test("c", () => {
    expect(true).toBe(true);
  });
});
```

### What is the expected behavior?

Both `afterAll` are executed:

```
bun test v0.6.12 (039bbc68)

test/index.test.ts:
✓ without test at the same level > a > b [0.05ms]
✓ with test at the same level > a > b [0.01ms]
✓ with test at the same level > c [0.02ms]
afterAll without test at the same level
afterAll with test at the same level
```

Note: This is the Jest's behavior:

```
  console.log
    afterAll without test at the same level

      at Object.<anonymous> (test/index.test.ts:24:26)

  console.log
    afterAll with test at the same level

      at Object.<anonymous> (test/index.test.ts:34:26)

 PASS  test/index.test.ts
  without test at the same level
    a
      ✓ b (1 ms)
  with test at the same level
    ✓ c
    a
      ✓ b
```

### What do you see instead?

No `afterAll` without test at the same level:

```
bun test v0.6.12 (039bbc68)

test/index.test.ts:
✓ without test at the same level > a > b [0.05ms]
✓ with test at the same level > a > b [0.01ms]
✓ with test at the same level > c [0.02ms]
afterAll with test at the same level
```

### Additional information

The workaround is adding a `test` at the same level:

```javascript
  afterAll(() => console.log("afterAll without test at the same level"));

  // Without any `test` on the same level of `afterAll`, it isn't triggered.
  // Workaround to trigger it. https://github.com/oven-sh/bun/issues/3494
  test("Workaround to run afterAll", () => undefined);
```

TiBianMod:
yep, the afterAll is not working as top level lifecycle hook, it works only as scoped lifecycle hook 🥲 

```ts
[...]

PR #13955, opened 2024-09-14
Fix memory corruption in napi_open_escapable_handle_scope

190n:
### What does this PR do?

Fixes a bug in `napi_open_escapable_handle_scope` that would corrupt memory after the location of the handle scope pointer. In:

```cpp
napi_escapable_handle_scope ehs;
auto *ehs_ptr = &ehs;
napi_open_escapable_handle_scope(env, ehs_ptr);
```

`ehs_ptr[1]`, or the memory after `ehs`, would be overwritten.

This was failing the NAPI test in CI on Windows because Windows detected heap corruption (the test stores a `napi_escapable_handle_scope` on the heap). Running locally, the corruption is not detected anymore.


robobun:
@190n, your commit d30cb7ab9249651d45305f78882d3489ed7e3126 is building: [`#3494`](https://buildkite.com/bun/bun/builds/3494)

Jarred-Sumner:
Very strange build errors.

PR #13784, opened 2024-09-07
misc node:zlib followup

nektro:
see commits for individual changes

PR #13958, opened 2024-09-14
misc node:zlib fixes [v2]

nektro:
rebase of https://github.com/oven-sh/bun/pull/13784
more to come but there's a few bug fixes in here i wanted to get up in the meantime

### DIFF
--- a/src/js/node/zlib.ts
+++ b/src/js/node/zlib.ts
@@ -29,7 +29,7 @@ const kFlushBuffers: Buffer[] = [];
 
 //
 
-function ZlibBase(method, options) {
+function ZlibBase(options, method) {
   if (options == null) options = {};
   if ($isObject(options)) {
     options.maxOutputLength ??= maxOutputLengthDefault;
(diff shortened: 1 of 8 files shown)


robobun:
@nektro, your commit 6b2e2527a9e44ad3f05afb7fbb45247a3b03b896 is building: [`#3497`](https://buildkite.com/bun/bun/builds/3497)

src/js/node/zlib.ts
```diff
@@ -45,34 +45,59 @@ function Base(method, options) {
...
+  get: function () {
+    return this[kHandle].finishFlush;
+  },
+});
+ObjectDefineProperty(ZlibBase.prototype, "_defaultFullFlushFlag", {
+  get: function () {
+    return this[kHandle].fullFlush;
+  },
+});
+ObjectDefineProperty(ZlibBase.prototype, "_maxOutputLength", {
```
Jarred-Sumner: Does this match node? Are these fields read-only and non-enumerable and non-configurable in node?

Jarred-Sumner:
Are there any new node tests that pass?

src/js/node/zlib.ts
nektro: not exactly but left a doc comment and a TODO to come back as well as made sure they were enumerable

nektro:
didnt check new tests but this does fix a number of bugs already:

- fixes a memory leak in crc32
- fixes a crash when brotli params are undefined
- fixes an error when zlib dictionary is undefined
- adds observable values for internal fields packages may be checking
- fixes internal class names packages may be checking

PR #13987, opened 2024-09-16
Update `@types/node` dependency version

vitch:
When I upgraded my project to use TypeScript@5.6 I ran into the issue described in [this SO post](https://stackoverflow.com/a/78790944/223931).

The solution is to use a newer version of `@types/node` which has been updated to deal with the changes in the latest TypeScript.

In my project I've introduced an explicit dependency on `@types/node@20.16.15` (the newest version currently available in the `^20.0.0` range) since the fix doesn't exist in `20.12.14` (the newest version available in the `~20.12.8` range which was currently defined) which seems to solve the problem. But I thought it would be helpful to others if the change was made here so they didn't need to debug the problem.

DISCLAIMER: I don't know how these types are used and if there is any chance I've introduced other problems with this change...

### What does this PR do?

Updates the `@types/node` dependency in `bun-types` from `~20.12.8` to `~20.16.5`

- [ ] Code changes


vitch:
I was a little surprised that these were the changes in the lockfile included in this PR:

Screenshot 2024-09-16 at 16 34 37

I'm not sure why `@types/node@*` wasn't also updated from `20.12.14`?

packages/bun-types/package.json
```diff
@@ -14,7 +14,7 @@
   ],
   "homepage": "https://bun.sh",
   "dependencies": {
-    "@types/node": "~20.12.8",
+    "@types/node": "~20.16.5",
```
mangs: The `22.x` branch is a closer match to Bun, so is there a reason to not adopt that?

```suggestion
    "@types/node": "~22.5.5",
```

vitch: I was being conservative because I wasn't really sure how the version was chosen or if `bun` was aiming for compatibility with a particular version of node.
The version was last updated in 2a401486d9fc38b3795571f6a69b3e17afba91a9 and I didn't see why that specific version was chosen (I don't _think_ it was the latest on May 1st?). A specific version of `@types/node` (rather than `*`) was added in #8195 but looking through that I didn't see any reasoning behind how the version was chosen.
TL;DR - I'm happy to make this change (and the corresponding change in the lockfile) but would love to understand the more about the why of this version!

mangs: Very reasonable. I'm curious as well. Thanks for pushing this along!

Jarred-Sumner:
Thanks for this.

We probably should update it to match the current reported version of Node,  which I think is the very latest release

vitch:
> Thanks for this.
> 
> We probably should update it to match the current reported version of Node, which I think is the very latest release

No worries :)

> the current reported version of Node

Sorry for being dense but not sure what you mean here... [This page](https://bun.sh/docs/runtime/nodejs-apis) "reflects Bun's compatibility with Node.js v20". 

The [node website](https://nodejs.org/en) offers `v20.17.0` as LTS or `v22.8.0` as current:
Screenshot 2024-09-17 at 11 09 56

vitch:
> the current reported version of Node

Ah - I poked around a bit more and saw that `process.version` is probably what you're talking about.

I'm using `bun@1.1.27` and `process.version` returns `v22.6.0`. There is no `22.6.x` released for `@types/node` so I'll update to the latest available (`22.5.5`)...

vitch:
I'm still a little confused by the lockfile changes:

Screenshot 2024-09-17 at 11 27 29

I would expect `@types/node@*` to have been also pointed at the new version since it's in range? I tried with `bun update` as well as by manually editing the `package.json` and running `bun install` but the effect was the same?

rtritto:
Why not use `"@types/node": "^22.5.5"` or `"@types/node": "^22"`?

vitch:
> Why not use `"@types/node": "^22.5.5"` or `"@types/node": "^22"`? 

I just tried to follow the pattern that was already there since I don't have full context on why it was like that previously,]
]

> By this way all _22 versions_ should be deduped to once.

Is this about my comment about the state of the lockfile? It seems to me that `20.12.14` _should_ have been removed from `@types/node@*` but it hasn't been... I don't see any other _22 version_ currently?

rtritto:
Bun haven't the dedupe feature (#1343).
So, when you change the version of a dependency, a new dependency will be added (if not exists) in the lock tree and then installed.

> I'm still a little confused by the lockfile changes:
> 
> Screenshot 2024-09-17 at 11 27 29
> 
> I would expect `@types/node@*` to have been also pointed at the new version since it's in range? I tried with `bun update` as well as by manually editing the `package.json` and running `bun install` but the effect was the same?

Previously your change, you have `@types/node@*, @types/node@~20.12.8` where they are deduped with `20.12.14` version.
In package.json of `types-bun` you changed `@types/node@~20.12.8` with `@types/node@~22.5.5`, so in _bun.lockb_:
- the `@types/node@*` version remains with `20.12.14` (there are at least 1 package.json with `@types/node@*`)
- the `@types/node@~20.12.8` version is removed (because there are no package.json with that version)
- the `@types/node@~22.5.5` version is added (new dependency)

rtritto:
Probably the `@types/node@*` was not removed (deduped) with #8195 (used only in package.json of bun-types) and can be removed.

vitch:
> Probably, as workaround, you can try to dedupe manually the `bun.lockb` and then check with run `bun i`.

How would I do that with the binary lockfile? I definitely would have done that if we were using `yarn` or something with an editable lockfile but afaik the bun one is a black box?

vitch:
I've pushed a change so we're just pointing at `^22` for the version of `@types/node`.

I'm still not sure why the lockfile has:

```
"@types/node@^16.18.61":
  version "16.18.102"
  resolved "https://registry.npmjs.org/@types/node/-/node-16.18.102.tgz"
  integrity sha512-eSe2YwGCcRjqPidxfm20IAq02krERWcIIJW4FNPkU0zQLbc4L9pvhsmB0p6UJecjEf0j/E2ERHsKq7madvthKw==

"@types/node@*":
  version "20.12.14"
  resolved "https://registry.npmjs.org/@types/node/-/node-20.12.14.tgz"
  integrity sha512-scnD59RpYD91xngrQQLGkE+6UrHUPzeKZWhhjBSa3HSkwjbQc38+q3RoIVEwxQGRw3M+j5hpNAM+lgV3cVormg==
  dependencies:
    undici-types "~5.26.4"

"@types/node@^22":
  version "22.7.4"
  resolved "https://registry.npmjs.org/@types/node/-/node-22.7.4.tgz"
  integrity sha512-y+NPi1rFzDs1NdQHHToqeiX2TIS79SWEAw9GYhkkx8bD0ChpfqC+n2j5OXOCpzfojBEBt6DnEnnG9MY0zk1XLg==
  dependencies:
    undici-types "~6.19.2"
```

It looks like only the root `package.json` and the one in `packages/bun-types` are included in the workspaces so I would expect to see:

```
"@types/node@*", "@types/node@^22":
  version "22.7.4"
  resolved "https://registry.npmjs.org/@types/node/-/node-22.7.4.tgz"
  integrity sha512-y+NPi1rFzDs1NdQHHToqeiX2TIS79SWEAw9GYhkkx8bD0ChpfqC+n2j5OXOCpzfojBEBt6DnEnnG9MY0zk1XLg==
  dependencies:
    undici-types "~6.19.2"
```

But I guess that's a different issue.

Is there anything that I can do to help this along?

mangs:
Adding an `overrides` field to `package.json` will force the package to have a specific version: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#overrides

You're likely seeing those different versions due to them being included from other packages.

rtritto:
@vitch you can remove all `@types/node@*` in all packages.json (except in package.json of `bun-types`) with Bun (`@types/node@*` should be removed from bun.lockb).
Then add `@types/node@*` in all packages.json (except in package.json of `bun-types`) with Bun.

vitch:
In the latest commit used [`overrides`](https://bun.sh/docs/install/overrides) (thanks @mangs) temporarily to clean up the lock file - the relevant items now look like this:

```
"@types/node@^16.18.61":
  version "16.18.112"
  resolved "https://registry.npmjs.org/@types/node/-/node-16.18.112.tgz"
  integrity sha512-EKrbKUGJROm17+dY/gMi31aJlGLJ75e1IkTojt9n6u+hnaTBDs+M1bIdOawpk2m6YUAXq/R2W0SxCng1tndHCg==

"@types/node@*", "@types/node@^22":
  version "22.7.4"
  resolved "https://registry.npmjs.org/@types/node/-/node-22.7.4.tgz"
  integrity sha512-y+NPi1rFzDs1NdQHHToqeiX2TIS79SWEAw9GYhkkx8bD0ChpfqC+n2j5OXOCpzfojBEBt6DnEnnG9MY0zk1XLg==
  dependencies:
    undici-types "~6.19.2"
```

> You're likely seeing those different versions due to them being included from other packages.

`^16.18.61` was included by other packages and I was expecting it to be there. `@types/node@*` was included by other packages but I was expecting it to automatically "fast forward" to the newest matching version specified in the lockfile. I guess it doesn't for stability reasons but it would be nice to have a way to tell `bun` to bump it.

rtritto:
Yes, `overrides` worked like my workaround.
LGTM, thanks!

djdjz7:
Is there any progress with this pr?

The latest version of bun (1.1.34) still seems to use `"@types/node": "~20.12.8"` with `bun init` in `bun-types`, and needs to be manually updated.

rtritto:
Any update?

Issue #7789, opened 2023-12-22
config gloabl test timeout in bunfig.toml


xlc:
### What is the problem this feature would solve?

so that I don't need to set it for every single tests

### What is the feature you are proposing to solve the problem?

have bunfig.toml supports gloabl timeout

For example, vitest supports config [testTimeout](https://vitest.dev/config/#testtimeout) in `vitest.config.ts`

code-yeongyu:
Reproduced on 1.4.0 (also 1.4.1), and adding a data point on why this matters more than a convenience.

`setDefaultTimeout` is documented as file-scoped, and a preload only runs once per process — so a preload that calls it raises the budget for the **first** test file only; every later file in a sequential `bun test` run falls back to 5000ms. Minimal repro:

```
mkdir -p repro/probe && cd repro
printf '[test]\npreload = ["./setup.ts"]\n' > bunfig.toml
printf 'import { setDefaultTimeout } from "bun:test"\nsetDefaultTimeout(20_000)\n' > setup.ts
for n in a b c; do printf 'import { test, expect } from "bun:test"\ntest("%s", async () => { await new Promise((r) => setTimeout(r, 6500)); expect(true).toBe(true) })\n' $n > probe/$n.test.ts; done
bun test probe            # 1 pass / 2 fail  (b,c: "timed out after 5000ms")
bun test --parallel probe # 3 pass  (each worker re-runs the preload)
bun test --timeout 20000 probe  # 3 pass
```

The natural place to set a repo-wide default is `[test] timeout` in bunfig.toml — but as reported here it is silently ignored: with `timeout = 20000` in `[test]` the run above is still 1 pass / 2 fail on 1.4.0. So today the only mechanism that reaches every file is the CLI flag, which has to be repeated at every `bun test` call site in CI.

Real-world impact: a monorepo preload raising the CI floor to 30s on windows-latest silently protected one file per job; a 9.3s test died at 5000ms and was chased as a Windows flake for weeks. A parsed `[test] timeout` (or a documented note that a preload's `setDefaultTimeout` is first-file-only) would have made that a one-line fix.

PR #13988, opened 2024-09-16
feat: support default test timeout in bunfig.toml

treyturner:
Hello there! I'm new to the project and Zig and thought I would try to help implement a small feature for which I voted.

### What does this PR do?

This PR fixes https://github.com/oven-sh/bun/issues/7789 by allowing default test timeout to be configured in `bunfig.toml`.

Thanks to @Jarred-Sumner for guidance 🙌 

- [x] Code changes

I wrote automated tests [here](https://github.com/treyturner/bun/blob/feat/set_test_timeout_in_bunfig/test/cli/test/bun-test-bunfig.test.ts).

A Zig file changed.
- [ ] ~~JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed~~


Jarred-Sumner (changes requested):
Thank you for this. Would you be able to add a test? I suggest putting it in `test/cli/test` and use `tempDirWithFiles` from `"harness"` to create a temp folder programmatically with the bunfig.toml. Probably set the default test timeout to something like 1 or 5 and then verify the process prints a timed out message

Jarred-Sumner:
> The caveat this creates is that specifying the default timeout via CLI will cause a non-standard value from bunfig.toml to be used instead. I'm not sure how to address this without more extensive changes that seem like overkill for such a corner case, but this is almost surely due to my inexperience with the project and Zig. Any assistance addressing this concern or improving this aspect would be greatly appreciated.

I think this block of code could be moved to run after the bunfig.toml is read:
https://github.com/oven-sh/bun/blob/ab07cf444d02971c6dbf6ade0f3f016eb515f75f/src/cli.zig#L456-L528

treyturner:
> > The caveat this creates is that specifying the default timeout via CLI will cause a non-standard value from bunfig.toml to be used instead. I'm not sure how to address this without more extensive changes that seem like overkill for such a corner case, but this is almost surely due to my inexperience with the project and Zig. Any assistance addressing this concern or improving this aspect would be greatly appreciated.
> 
> I think this block of code could be moved to run after the bunfig.toml is read:
> 
> https://github.com/oven-sh/bun/blob/ab07cf444d02971c6dbf6ade0f3f016eb515f75f/src/cli.zig#L456-L528

10-4, thanks! Will likely be a few days before I can look at this again in detail, but it's on my TODO list.

treyturner:
Ok, I've got the tests for this pushed, including an extra test that I plan to delete before merge. Will be looking at the implementation next chance I get.

treyturner:
Ok @Jarred-Sumner, ready for another look when you get time!

treyturner:
This is good to go from my perspective, but I'll pop in every now and again to rebase from main and would be happy to address any further concerns or change requests.

treyturner:
Hi @Jarred-Sumner, would you be willing to give this another look? Would love to get this over the line.

PR #13989, opened 2024-09-17
zig: error dead code removal

nektro:


robobun:
@nektro, your commit d0a8eb21f21d4a3361f0efa05bbb08eece4205f4 is building: [`#3557`](https://buildkite.com/bun/bun/builds/3557)

src/bun.js/bindings/bindings.zig
```diff
@@ -2927,6 +2927,17 @@ pub const JSGlobalObject = opaque {
         this.vm().throwError(this, err);
     }
 
+    pub fn throwMissingArgumentsValue(this: *JSGlobalObject, arg_names: []const []const u8) JSValue {
```
dylan-conway: I think `arg_names` should be comptime so we don't accidentally panic at runtime

nektro: was worried about about generic instantiation bloat but since the function is so small i can make it `inline` too. the call it makes is non-inline so it wont explode by doing so

dylan-conway (approved):

PR #13997, opened 2024-09-17
fix: use Bun's Fetch interface for fetch

### LINKED ISSUE #13827: TypeScript: 'proxy' property not recognized in fetch RequestInit in Bun.js

1.1.27+267afa293

1. Create a new project using Bun.js and TypeScript.
2. Write an asynchronous function using `fetch` to make a request, including a `proxy` property in the `RequestInit` object.
3. Open the project in **Visual Studio Code** with TypeScript support enabled (e.g., using the TypeScript extension).
4. Add the following code in a TypeScript file:
```typescript
const request = async () => {
  const response = await fetch(
    "https://example-web.com/",
    {
      method: "GET",
      proxy: "http://your-proxy-url:port", // <-- VSCode display an error here
    }
  );
  return response;
};

request().then(console.log).catch(console.error);
```
5. Notice that **VSCode** displays a TypeScript error saying the `proxy` property does not exist in the `RequestInit` type.
6. Despite the error in VSCode, the code still runs without issues when compiled and executed with **Bun.js**.

The TypeScript type definitions in **VSCode** should recognize the `proxy` property as valid within the `RequestInit` type when using `fetch` in **Bun.js**. No TypeScript error or warning should appear in the editor, and the property should be fully supported with proper type hinting and autocomplete in the IDE.

### What do you see instead?

In **Visual Studio Code**, a TypeScript error is displayed in the editor, stating that the `proxy` property does not exist in the `RequestInit` type. The error message is:
```typescript
No overload matches this call.
Overload 1 of 2, '(input: RequestInfo, init?: RequestInit | undefined): Promise<Response>', gave the following error.
Object literal may only specify known properties, and 'proxy' does not exist in type 'RequestInit'.
```
Despite this, the code runs without any issues when compiled and executed with Bun.js.

### Additional information

_No response_

ragavpr:
Allows TypeScript to use Bun's FetchRequestInit instead of Node's RequestInit in fetch request options.

### What does this PR do?

- [ ] Code changes

I modified the `.d.ts` file directly in my own project's `node_modules` and verified the type is recognized correctly in VSCode.

### Additional Info

This previously existing similar type probably supports, `fetch` to be of `Fetch`, hence made this change
https://github.com/oven-sh/bun/blob/3d68a9483fbb69b7ac4dd9c686658ec14a6674e4/packages/bun-types/globals.d.ts#L945

Fixes #13827

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -3170,7 +3170,7 @@ declare module "bun" {
    */
   function openInEditor(path: string, options?: EditorOptions): void;
 
-  const fetch: typeof globalThis.fetch & {
+  const fetch: Fetch & {
     preconnect(url: string): void;
   };
 


dannote:
@Jarred-Sumner, is there anything holding up this PR from being merged? I’d be happy to assist if needed.

alii:
This is already resolved in a PR I'm working on at the moment — #18024

Thank you for the PR, but closing as it will be fixed once 18024 merges

PR #14007, opened 2024-09-18
fix zig path in GH action

robertshuford:
### What does this PR do?

This fixes the zig path problem in the GitHub Actions `.github/workflows/run-format.yml` and `.github/workflows/format.yml`.

I ran the action on my fork for this branch: https://github.com/robertshuford/bun/actions/runs/10914651113.


robertshuford:
Is this worth fixing? 
Or does the removal of `"zig": "vendor/zig/zig.exe"` cause downstream issues I'm unaware of?

robertshuford:
Fixed in https://github.com/oven-sh/bun/pull/14152

PR #14023, opened 2024-09-19
Fixes #14014

### LINKED ISSUE #14014: C Blog example results in "./hello.c:1: error: include file 'stdio.h' not found"
TL;DR: fix is coming in the next stable release, for now do:

```
bun upgrade --canary
```

Example code is not working

### What is the issue?

I'm trying [to follow the first C example](https://bun.sh/blog/compile-and-run-c-in-js) but find this error:

image

Being in MacOS 14.5 (Apple M1) with Bun 1.1.28, I've tried the "basic" things to no avail:
- `xcode-select --install` -> "Command line tools are already installed."
- Double-check files are exactly the same as the first example.
- Changing (temporarily) `void hello() {` to `int main() {` allows me to compile and run the C program as usual:

image

I thought it'd work just like that from reading @Jarred-Sumner's Tweets:
- "No complicated build system. No runtime api."
- "Bun has embedded TinyCC since we introduced FFI support. That means we didn't add any new dependencies to add C compilation support."

So I thought it'd work by default just doing that. Am I missing something perhaps? Do I need to run `bun` in some special way? I don't see any extra instructions in the blog post like "add `--experimental` flag" or anything like that.

### Where did you find it?

First example of https://bun.sh/blog/compile-and-run-c-in-js

Here are my files (they are copy/pasted from the example):

image

Edit: changing the file from `hello.js` to `hello.ts` also didn't help.

Jarred-Sumner:
### What does this PR do?

Fixes #14014 on macOS

I'm still not sure about Linux but let's see

To test this PR once it builds:

```sh
bunx bun-pr jarred/fixes-14014
```

### DIFF
--- a/src/bun.js/api/ffi.zig
+++ b/src/bun.js/api/ffi.zig
@@ -271,19 +271,19 @@ pub const FFI = struct {
                 }
             } else if (Environment.isLinux) {
                 if (Environment.isX64) {
-                    if (bun.sys.directoryExistsAt(std.fs.cwd(), "/usr/include/x86_64-linux-gnu")) {
+                    if (bun.sys.directoryExistsAt(std.fs.cwd(), "/usr/include/x86_64-linux-gnu").isTrue()) {
                         cached_default_system_include_dir = "/usr/include/x86_64-linux-gnu";
                     }
 
-                    if (bun.sys.directoryExistsAt(std.fs.cwd(), "/usr/lib/x86_64-linux-gnu")) {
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit 33d83ba022013921ccf89239a5e170c74926b5b1 is building: [`#3590`](https://buildkite.com/bun/bun/builds/3590)

dylan-conway (approved):

Issue #14027, opened 2024-09-19
bun build cause Elysia to panic on cluster mode


SaltyAom:
### What version of Bun is running?

1.1.28

### What platform is your computer?

Darwin 24.0.0 arm64 arm

### What steps can reproduce the bug?

1. Clone repro
```bash
git clone http://github.com/saltyaom/elysia-cluster-demo
```

2. Build
```bash
cd elysia-cluster-demo && bun run build
```

3. Run
```bash
bun ./dist/index.js
```

### What is the expected behavior?

It should not crash

### What do you see instead?

TypeError: Y of p1 is not a function. (In 'Y of p1(H)', 'Y of p1' is undefined)

### Additional information

Bundled code using Bun build cause panic on run.

This error only happens on Bun 1.1.28 and not on previous versions.

nektro:
PR linked above should have fixed this but did not link it. if the issue persists please let us know

PR #14033, opened 2024-09-19
fix(bundler): disable moving identifiers

paperclover:
Temporary fix for #14027 so elysia can be unblocked

The actual bug is bundle_v2.zig:9218 and has existed for a much longer time.


robobun:
@paperdave, your commit 895615c053b3a5f5f111a7363c88032c522e100f is building: [`#3600`](https://buildkite.com/bun/bun/builds/3600)

Issue #3603, opened 2023-07-11
Standalone minify function


tipiirai:
### What is the problem this feature would solve?

Bun has a fantastic JS minifier, but it's not available as an API method.

### What is the feature you are proposing to solve the problem?

`Bun.minify(js)` or something

### What alternatives have you considered?

`esbuild.transformSync` method or [Terser](https://github.com/terser/terser)

Electroid:
```js
new Bun.Transpiler({ minify: true }).transformSync(code);
```

tipiirai:
Wow! Thanks. Worth mentioning on your docs

PR #14036, opened 2024-09-19
Experiment: Add `buffer` type and inline pointer

Jarred-Sumner:
### What does this PR do?

- [ ] Code changes

### DIFF
--- a/src/bun.js/api/ffi.zig
+++ b/src/bun.js/api/ffi.zig
@@ -76,6 +76,24 @@ const IOTask = JSC.IOTask;
 const TCC = @import("../../tcc.zig");
 extern fn pthread_jit_write_protect_np(enable: bool) callconv(.C) void;
 
+const Offsets = extern struct {
+    JSArrayBufferView__offsetOfLength: u32,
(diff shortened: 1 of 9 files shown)


robobun:
@Jarred-Sumner, your commit 02540ac152e48a23cec8a270d01d948abf135c5c is building: [`#3603`](https://buildkite.com/bun/bun/builds/3603)

src/bun.js/api/ffi.zig
```diff
@@ -2064,7 +2080,7 @@ pub const FFI = struct {
         function = 17,
         napi_env = 18,
         napi_value = 19,
-
+        buffer = 20,
         pub const max = @intFromEnum(ABIType.napi_value);
```
190n, line 2084: should at least change to `ABIType.buffer`, or we can calculate the max:

```suggestion
        pub const max = max_abi_type: {
            var current_max = std.math.minInt(i32);
            for (@typeInfo(ABIType).Enum.fields) |f| {
                current_max = @max(current_max, f.value);
            }
            break :max_abi_type current_max;
        };
```

or, could we `@cImport` these?

PR #14044, opened 2024-09-19
docs: Add new methods to binary-data.md

bakkot:
A couple of my proposals ([x](https://github.com/tc39/proposal-arraybuffer-base64), [x](https://github.com/whatwg/fetch/pull/1753)) have made some thing a bit easier for converting between various representations of binary data: now there's `body.bytes()` for getting a `Uint8Array`, and `uint8Array.toBase64()` and friends for converting to/from base64/hex.

These are available in Bun as of https://github.com/oven-sh/bun/pull/11104 and https://bun.sh/blog/bun-v1.1.22#uint8array-tobase64, so the docs can be updated to refer to them.

Also a couple of drive-by fixes while I was here.


Jarred-Sumner:
Thank you!

PR #14048, opened 2024-09-19
fix for windows debug support

snoglobe:
### What does this PR do?

this fixes https://github.com/oven-sh/bun/issues/8851, which pertains to the bun debugger not working on windows.

for windows platform, we now use websockets to communicate to the debugger and inspector. on unix platforms, we keep the existing unix sockets implementation.

- [x] Code changes

tested it by debugging programs under vscode in windows


robobun:
@snoglobe, your commit bc666970122eb45c7d090b462828547b7bb3f8f6 is building: [`#3614`](https://buildkite.com/bun/bun/builds/3614)

snoglobe:
note: failed builds are due to buildkite

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -2,12 +2,19 @@ import type { InspectorEventMap } from "../../../bun-inspector-protocol/src/insp
...
-import { spawn } from "node:child_process";
+import { spawn, ChildProcess } from "node:child_process";
 import { EventEmitter } from "node:events";
 import { WebSocketInspector, remoteObjectToString } from "../../../bun-inspector-protocol/index";
-import { UnixSignal, randomUnixPath } from "./signal";
+import { randomUnixPath, UnixSignal, WebSocketSignal } from "./signal";
 import { Location, SourceMap } from "./sourcemap";
+import { createServer, AddressInfo } from "node:net";
+
+export function getAvailablePort(): number {
```
Jarred-Sumner: Do we need to wait for the callback to be called?

packages/bun-vscode/scripts/build.mjs
```diff
@@ -9,7 +9,7 @@ buildSync({
   entryPoints: ["src/extension.ts", "src/web-extension.ts"],
   outdir: "dist",
   bundle: true,
-  external: ["vscode"],
+  external: ["vscode", "ws"],
```
Jarred-Sumner: Is this intentional?

snoglobe:
yes; build fails otherwise

Jarred-Sumner:
What happens when you build the installed extension and the ws package
isn’t installed?

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
snoglobe: it works fine without waiting for callback

Jarred-Sumner:
> errors w/ the bundler unable to resolve ws

did you run `bun install` (or npm install)?

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
Jarred-Sumner: does node gurantee that the address will always be set before the listen callback is invoked?

src/js/internal/debugger.ts
```diff
@@ -73,7 +83,7 @@ class Debugger {
   #listen(): void {
     const { protocol, hostname, port, pathname } = this.#url;
 
-    if (protocol === "ws:" || protocol === "ws+tcp:") {
```
Jarred-Sumner: If we're allowing wss, where is the TLS configuration specified?

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
Jarred-Sumner: > server.address() returns null before the 'listening' event has been emitted or after calling server.close().

Jarred-Sumner (changes requested):
Not mergeable until:
1) the `listen()` callback is addressed (it will be randomly flaky, in it's current version even if it appears to work sometimes when you run it locally)
2) The `"ws"` module should not be external. We bundle the VSCode extension instead of expecting a node_modules folder to be present (it won't be, except for development) 

Separately, the `"wss"` validation is a little confusing unless the idea is you have a reverse proxy between Bun.serve() and the websocket connection? This is uncommon, as people usually don't proxy websockets but not impossible I suppose.

src/js/internal/debugger.ts
```diff
@@ -321,6 +331,18 @@ function reset(): string {
   return "";
 }
 
+function notifyWebSocket(url: string): void {
```
Jarred-Sumner: Oh, it's a whole websocket server for this? We can just make a TCP connection.

Jarred-Sumner:
I want to get the Bun changes into the release, so going to merge that separately from this PR since the extension can be updated separately after

snoglobe:
> Not mergeable until:
> 
> 1. the `listen()` callback is addressed (it will be randomly flaky, in it's current version even if it appears to work sometimes when you run it locally)
> 2. The `"ws"` module should not be external. We bundle the VSCode extension instead of expecting a node_modules folder to be present (it won't be, except for development)
> 
> Separately, the `"wss"` validation is a little confusing unless the idea is you have a reverse proxy between Bun.serve() and the websocket connection? This is uncommon, as people usually don't proxy websockets but not impossible I suppose.

will remove wss validation then; yeah not sure why i put it

snoglobe:
> > errors w/ the bundler unable to resolve ws
> 
> did you run `bun install` (or npm install)?

yes

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -2149,10 +2207,10 @@ function titleize(name: string): string {
 }
 
 function sourcePresentationHint(url?: string): DAP.Source["presentationHint"] {
-  if (!url || !url.startsWith("/")) {
+  if (!url || path.isAbsolute(url)) {
```
Jarred-Sumner: !path.isAbsolute

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -2567,3 +2628,11 @@ let sequence = 1;
 function nextId(): number {
   return sequence++;
 }
+
+export function getRandomId() {
+  return Math.random().toString(36).slice(2);
+}
+
+export function normalizeWindowsPath(path: string): string {
+  return (path.charAt(0).toUpperCase() + path.slice(1)).replaceAll("\\\\", "\\");
```
Jarred-Sumner:
```suggestion
  path = normalize(path);
  if (path[1] === ':' && (path[2] === "\\" || path[2] === "/")) { 
    return (path.charAt(0).toUpperCase() + path.slice(1)).replaceAll("\\\\", "\\");
  }
   return path;
```

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -2163,6 +2221,9 @@ function sourceName(url?: string): string {
     return "unknown.js";
   }
   if (isJavaScript(url)) {
+    if (process.platform === "win32") {
+      return url.split("\\").pop() || url;
```
Jarred-Sumner:
```suggestion
      url = url.replaceAll("\\", "/");
```

packages/bun-vscode/src/features/debug.ts
```diff
@@ -211,18 +215,18 @@ class TerminalDebugSession extends FileDebugSession {
...
-      this.signal.on("Signal.received", () => {
-        vscode.debug.startDebugging(undefined, {
-          ...ATTACH_CONFIGURATION,
-          url: this.adapter.url,
-        });
+      });
+    } else {
+      this.signal = new UnixSignal();
+    }
+    this.signal.on("Signal.received", () => {
```
Jarred-Sumner, line 234: signal is undefined here

packages/bun-vscode/src/features/debug.ts
```diff
@@ -176,19 +176,23 @@ class FileDebugSession extends DebugSession {
...
+    let url;
+    if (process.platform === "win32") {
+      getAvailablePort().then(port => {
+        // the signal is only first used well after this resolves
+        url = `ws://127.0.0.1:${port}/${getRandomId()}`;
+      });
+    } else {
+      url = `ws+unix://${tmpdir()}/${uniqueId}.sock`;
+    }
+    this.adapter = new DebugAdapter(url);
```
Jarred-Sumner, line 195: url is undefined here

packages/bun-debug-adapter-protocol/src/debugger/signal.ts
```diff
@@ -85,3 +85,72 @@ function parseUnixPath(path: string | URL): string {
...
+  #server: ReturnType<typeof createServer>;
+  #ready: Promise<void>;
+
+  constructor(port: number) {
+    super();
+    this.#port = port;
+
+    this.#server = createServer((socket: Socket) => {
+      socket.on('data', (data) => {
+        console.error('received', data);
```
Jarred-Sumner:
```suggestion

```

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -489,36 +503,73 @@ export class DebugAdapter extends EventEmitter<DebugAdapterEventMap> implements
...
 
-    if (!started) {
-      throw new Error("Program could not be started.");
+      this.once("Adapter.terminated", () => {
+        signal.close();
+      });
+
+      const query = stopOnEntry ? "break=1" : "wait=1";
+      processEnv["BUN_INSPECT"] = `${url}?${query}`;
+      processEnv["BUN_INSPECT_NOTIFY"] = `tcp://127.0.0.1:${signal.port}`; // 127.0.0.1 so it resolves correctly on windows
```
Jarred-Sumner:
```suggestion
      processEnv["BUN_INSPECT_NOTIFY"] = signal.url
```

liudonghua123:
I see this pr changed unix domain socket to websocket through network ipc. It just like bun inspect cli debugging doing.

I tested some simple demos of unix domain socket on windows and it works. How about to fix existing unix domain socket for windows, it's more high performance in my opinion. It's just my own thought. :smile:

Jarred-Sumner:
@liudonghua123 you're welcome to try getting unix domain sockets working when used in VSCode on Windows

PR #14068, opened 2024-09-20
Keep event loop alive when refConcurrently has been called

### LINKED ISSUE #13938: multiple await prisma query inside for of crash on first iterate

1.1.27

Windows 10

inside a for of loop, there are two await one of sleep and one of query prisma. With the same source code. run in nodejs works as expected but bun is crash on the first iterate

Reproduce repo: https://github.com/alexewsv/bun-prisma-weird

works as epxected with nodejs

### What do you see instead?

stopped (crash) on the first iterate

### Additional information

If commented one await (sleep) will work as expected

190n:
### What does this PR do?

`refConcurrently` and `unrefConcurrently` increment an atomic counter which is later used to ref or unref the event loop on the main thread. Previously, if `refConcurrently` had been called but the ref had not yet been applied by `updateCounts`, the event loop could exit early. Now we check if there are any pending refs when determining whether to exit.

Fixes #13938.

Extended the NAPI tests to cover `napi_threadsafe_function` used without `napi_async_work`, which is what Prisma does that triggered this bug.

### DIFF
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -1288,8 +1288,14 @@ pub const EventLoop = struct {
         _ = this.tickConcurrentWithCount();
     }
 
+    /// Check whether refConcurrently has been called but the change has not yet been applied to the
+    /// underlying event loop's `active` counter
+    pub fn hasPendingRefs(this: *const EventLoop) bool {
+        return this.concurrent_ref.load(.seq_cst) > 0;
+    }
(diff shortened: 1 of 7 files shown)


robobun:
@190n, your commit c1970046f2002a8e34d4b6f96a01d08b6bda0284 is building: [`#3644`](https://buildkite.com/bun/bun/builds/3644)

Issue #3676, opened 2023-07-19
Print commit hash with `bun --revision`


lgarron:
### What is the problem this feature would solve?

While testing https://github.com/oven-sh/bun/issues/3669 I've been using `bun upgrade --canary`, which is much more convenient than building `bun` from scratch (especially if you haven't pulled `main` since a bunch of dependencies changed).

However, I don't see an obvious way to find the version of a currently installed canary.

### What is the feature you are proposing to solve the problem?

I think it would be really handy if `bun --version` printed enough information to identify an exact project commit.

> git describe --tags
bun-v0.6.14-53-g8bd2b784a

One option might be to modify https://github.com/oven-sh/bun/blob/3ec6c9e4fb03a374bb5a1d856bf804483246589c/.github/workflows/bun-release.yml#L68C1-L71 to use:

```shell
> git describe --tags
bun-v0.6.14-53-g8bd2b784a
```

Since this runs after `actions/checkout@v3`, I think it shouldn't be more expensive than what it's currently doing.

I'm on macOS Ventura 13.4.1, if it helps to know.

lgarron:
> Since this runs after `actions/checkout@v3`, I think it shouldn't be more expensive than what it's currently doing.

Hmm, it looks like the relevant code might actually be trying to do something like this:

https://github.com/oven-sh/bun/blob/d265ed80d20267bc5252c31022af36d139bc283e/.github/workflows/bun-release-types-canary.yml#L38-L42

Nevertheless, it doesn't show up for me when I run `bun --version` (and I'm definitely seeing behaviour right now that is only fixed in the latest canary).

Electroid:
We can't change `bun --version`, because old versions of Bun depend on its format for upgrades. But we can introduce something new like `bun --revision` that includes the git commit.

```
> bun --version
0.7.2
> bun --revision
0.7.2+g8bd2b78
```

hyuzea:
currently on linux i can use 
```
echo 'console.info(Bun.revision)' | bunx bun-repl 
[0.07ms] ".env"
Welcome to Bun.js v0.7.2
Type ".help" for more information.
928816bef0945e9655c115767c719d6865ff138c
```
or 
```
echo 'console.info(Bun)' | bunx bun-repl | grep revision
[0.04ms] ".env"
  revision: "928816bef0945e9655c115767c719d6865ff138c",
```

lgarron:
@yus-ham Thanks, that's super handy!

[...]

PR #14079, opened 2024-09-21
rework node:zlib to match internal js api and properly support async writes

nektro:
added many more tests

buildkite comment: https://github.com/oven-sh/bun/pull/14079#issuecomment-2365033092


robobun:
@nektro, your commit abfd3d40030fcb9941eed1e366e4fbadb15404b1 is building: [`#3676`](https://buildkite.com/bun/bun/builds/3676)

Jarred-Sumner:
Definitely a memory leak:
``sh
❯ mem bun-14079 server.js
  ▲ Next.js 14.2.8
  - Local:        http://localhost:3000
  - Network:      http://0.0.0.0:3000

 ✓ Starting...
 ✓ Ready in 47ms
Peak memory usage: 848 MB
```

```sh
❯ mem bun server.js
  ▲ Next.js 14.2.8
  - Local:        http://localhost:3000
  - Network:      http://0.0.0.0:3000

 ✓ Starting...
 ✓ Ready in 39ms
Peak memory usage: 185 MB
```

Jarred-Sumner:
I think it'd be interesting to approach processChunkSync differently, with an internal version of writeSync that batches writes instead of creating a bunch of tiny buffers that immediately get concatenated

src/bun.js/bindings/Constants.h
```diff
@@ -0,0 +1,5 @@
+#include "root.h"
+#include "JavaScriptCore/JSCJSValue.h"
+#include "ZigGlobalObject.h"
+
+JSC::JSValue ZlibConstants(Zig::GlobalObject* globalObject);
```
Jarred-Sumner: should this just be in process binding constants?

https://github.com/oven-sh/bun/blob/fc6f8937c00cf4ab55a0cea408770f7fea5d64b3/src/bun.js/bindings/ProcessBindingConstants.cpp#L977-L1095

src/bun.js/bindings/bindings.zig
```diff
@@ -4241,6 +4241,13 @@ pub const JSValue = enum(JSValueReprInt) {
         return JSBuffer__isBuffer(global, value);
     }
 
+    pub fn isArrayBufferView(val: JSC.JSValue, globalThis: *JSC.JSGlobalObject) bool {
```
Jarred-Sumner: Is this method necesary?

`.asArrayBuffer()` returns null if the value is not an ArrayBufferView.

src/bun.js/bindings/bindings.zig
```diff
@@ -6264,6 +6271,10 @@ pub const CallFrame = opaque {
 
     pub const name = "JSC::CallFrame";
 
+    fn argsDirect(self: *const CallFrame) [*]const JSC.JSValue {
```
Jarred-Sumner: It's not arguments, let's not allow it to be confused

```suggestion
    inline fn asUnsafeJSValueArray(self: *const CallFrame) [*]const JSC.JSValue {
```

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -68,3 +61,785 @@ pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
...
+            else => unreachable,
+        }
+    }
+
+    pub fn reset(this: *BrotliContext) Error {
+        return this.init();
+    }
+
+    pub fn setBuffers(this: *BrotliContext, in: ?[]const u8, out: ?[]u8) void {
+        this.next_in = if (in) |p| p.ptr else null;
```
Jarred-Sumner, line 826: can this be a stale pointer?

src/deps/brotli_decoder.zig
```diff
@@ -152,6 +154,38 @@ pub const BrotliDecoderErrorCode = enum(c_int) {
     ALLOC_BLOCK_TYPE_TREES = -30,
     UNREACHABLE = -31,
 };
+pub const BrotliDecoderErrorCode2 = enum(c_int) {
```
Jarred-Sumner, line 157: Why is it called `BrotliDecoderErrorCode2`?

src/js/internal/errors.ts
```diff
@@ -1,9 +1,20 @@
+function genericNodeError(message, errorProperties) {
```
Jarred-Sumner: We should not encourage this pattern. Is this necessary?

src/js/node/zlib.ts
```diff
@@ -1,78 +1,277 @@
...
+  }
+  buffer = processChunkSync(engine, buffer, engine._finishFlushFlag);
+  if (engine._info) return { buffer, engine };
+  return buffer;
+}
+
+function zlibOnError(message, errno, code) {
+  const self = this[owner_symbol];
+  // There is no way to cleanly recover. Continuing only obscures problems.
+  const error = genericNodeError(message, { errno, code });
```
Jarred-Sumner: our handle can create an error instance, right?

src/js/node/zlib.ts
```diff
@@ -448,19 +748,51 @@ const zlib = {
...
+  codes: {
+    enumerable: true,
+    value: ObjectFreeze(codes),
+  },
 });
 
+// These should be considered deprecated
+// expose all the zlib constants
+for (const bkey of ObjectKeys(constants)) {
+  if (StringPrototypeStartsWith.$call(bkey, "BROTLI")) continue;
```
Jarred-Sumner: we could copy paste it just once instead of dynamically checking this every time someone runs `require("zlib")` for the first time?

src/js/node/zlib.ts
```diff
@@ -86,358 +285,459 @@ ZlibBase.prototype.flush = function (kind, callback) {
...
+      opts = {};
+    }
+    return zlibBuffer(new ctor(opts), buffer, callback);
+  };
 }
-Gunzip.prototype = Object.create(Zlib.prototype);
 
-//
+const kMaxBrotliParam = MathMax(
+  ...ArrayPrototypeMap.$call(ObjectKeys(constants), key =>
```
Jarred-Sumner: can you copy paste the constant once instead of doing this every time bun starts?

src/bun.js/base.zig
```diff
@@ -553,7 +553,7 @@ pub const ArrayBuffer = extern struct {
     ///    new ArrayBuffer(view.buffer, view.byteOffset, view.byteLength)
     /// ```
     pub inline fn byteSlice(this: *const @This()) []u8 {
-        return this.ptr[this.offset .. this.offset + this.byte_len];
+        return this.ptr[this.offset..][0..this.byte_len];
```
Jarred-Sumner, line 556: What was the reason for this change? Zig was/is missing safety checks in `[offset...][0..len]`

src/bun.js/bindings/Constants.cpp
```diff
@@ -0,0 +1,162 @@
...
+    GZIP,
+    GUNZIP,
+    DEFLATERAW,
+    INFLATERAW,
+    UNZIP,
+    BROTLI_DECODE,
+    BROTLI_ENCODE
+};
+
+#define NODE_DEFINE_CONSTANT(obj, C) obj->putDirect(vm, JSC::Identifier::fromString(vm, WTF::ASCIILiteral::fromLiteralUnsafe(#C)), JSC::jsNumber(C))
```
Jarred-Sumner: We can move this whole thing into ProcessBindingConstants

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -68,3 +61,785 @@ pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
...
+                in = null;
+                in_len = 0;
+                in_off = 0;
+            } else {
+                bun.assert(arguments[1].isArrayBufferView(globalThis));
+                const in_buf = arguments[1].asArrayBuffer(globalThis).?;
+                in_off = arguments[2].toU32();
+                in_len = arguments[3].toU32();
+                bun.assert(in_buf.byte_len >= in_off + in_len);
+                in = in_buf.byteSlice()[in_off..][0..in_len];
```
Jarred-Sumner, line 199: missing bounds checks

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -68,3 +61,785 @@ pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
...
+                // just a flush
+                in = null;
+                in_len = 0;
+                in_off = 0;
+            } else {
+                bun.assert(arguments[1].isArrayBufferView(globalThis));
+                const in_buf = arguments[1].asArrayBuffer(globalThis).?;
+                in_off = arguments[2].toU32();
+                in_len = arguments[3].toU32();
+                bun.assert(in_buf.byte_len >= in_off + in_len);
```
Jarred-Sumner, line 198: assert can't really be considered a bounds check

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -68,3 +61,785 @@ pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
...
+            flush = arguments[0].toU32();
+            _ = std.meta.intToEnum(bun.zlib.FlushValue, flush) catch bun.assert(false); // Invalid flush value
+
+            if (arguments[1].isNull()) {
+                // just a flush
+                in = null;
+                in_len = 0;
+                in_off = 0;
+            } else {
+                bun.assert(arguments[1].isArrayBufferView(globalThis));
```
Jarred-Sumner: `.?`  asserts this
```suggestion

```

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -68,3 +61,785 @@ pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
...
+                const in_buf = arguments[1].asArrayBuffer(globalThis).?;
+                in_off = arguments[2].toU32();
+                in_len = arguments[3].toU32();
+                bun.assert(in_buf.byte_len >= in_off + in_len);
+                in = in_buf.byteSlice()[in_off..][0..in_len];
+            }
+
+            bun.assert(arguments[4].isArrayBufferView(globalThis));
```
Jarred-Sumner: Already asserting this
```suggestion
```

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -68,3 +61,785 @@ pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) callcon
...
+            const out_buf = arguments[4].asArrayBuffer(globalThis).?;
```
Jarred-Sumner, line 202: I still don't quite know what is keeping these buffers alive.

src/codegen/generate-classes.ts
```diff
@@ -361,6 +361,12 @@ JSC_DECLARE_CUSTOM_GETTER(js${typeName}Constructor);
 `;
   }
 
+  if (obj.wantsThis) {
+    externs += `
+extern JSC_CALLCONV void* JSC_HOST_CALL_ATTRIBUTES ${classSymbolName(typeName, "_setThis")}(void*, JSC::EncodedJSValue);
```
Jarred-Sumner: What is the purpose of this?

test/harness.ts
```diff
@@ -393,9 +393,9 @@ expect.extend({
       message: () => `Expected ${cmds.join(" ")} to fail`,
     };
   },
-  toThrowWithCode(fn: CallableFunction, cls: CallableFunction, code: string) {
+  async toThrowWithCode(fn: CallableFunction, cls: CallableFunction, code: string) {
```
Jarred-Sumner: is this necessary? and if it is necessary, is it always necessary? I don't trust we'll remember to `await` these expect calls.

test/js/node/test/parallel/zlib-unused-weak.test.js
```diff
@@ -0,0 +1,48 @@
+//#FILE: test-zlib-unused-weak.js
+//#SHA1: cb228e175738cb88a93415ee046c655cb65d26a9
+//-----------------
+'use strict';
+const zlib = require('zlib');
+
+let memoryUsageState = 'initial';
+process.memoryUsage = jest.fn(() => {
```
Jarred-Sumner: can you delete this file

this test is not testing anything

Jarred-Sumner (changes requested):
Several nitpicky comments, but a handful of important ones

What prevents the GC from collecting the input and output buffers?

Can we avoid doing a bunch of setup work on statically-known constant data when node:zlib initializes?

src/deps/brotli_decoder.zig
nektro: uses different field names, to cover the possible codes more exhaustively. didn't want to break existing code in this branch

src/bun.js/base.zig
nektro: fixed in https://github.com/ziglang/zig/pull/19374 which made it into both 0.12 and 0.13

src/bun.js/node/node_zlib_binding.zig
nektro: see above

PR #14086, opened 2024-09-21
✏️ fix: Correct typo "messag:" to "message:"

Sayrix:
### What does this PR do?

Fixed a typo

- [ ] Code changes


jbampton (commented):
Seems this is not a mistake.  

They are showing what happens when there is a typo for both "strict" and "nonStrict"

Issue #3704, opened 2023-07-20
Got Array size error when run bunx prisma init


UnkAtreus:
### What version of Bun is running?

0.6.14

### What platform is your computer?

Microsoft Windows NT 10.0.22621.0 x64

### What steps can reproduce the bug?

create new project
bun add -d prisma
bunx prisma init <-- error issue

### What is the expected behavior?

should run prisma init and migrate

### What do you see instead?

Error: Array size is not a small enough positive integer.
error: "prisma" exited with code 1 (SIGHUP)

### Additional information
prisma version 5.0.0
_No response_

Electroid:
Could you try again after running `bun upgrade --canary`? We likely fixed this issue here: #3567

UnkAtreus:
error message is missing but still not working. Do not have any file created.

telesvar:
On my side when I run the command, I don't get an error 🤔

UnkAtreus:
Can I ask you, What a prisma and bun version that you used?

bytespider:
Maybe relevant or not, but I was getting no output from bunx with the Prisma CLI until I installed node.
It seems likely to me that Prisma is explicitly calling node for some parts of its CLI.

Jarred-Sumner:
Pretty confident this has since been fixed

If you're still running into this issue, please let us know

PR #14088, opened 2024-09-22
Don't re-create the FIFO in streams every time

Jarred-Sumner:
### What does this PR do?

Don't re-create the FIFO in streams every time


robobun:
@Jarred-Sumner, your commit 16038aed50c9a56bdac7499a9e7219cd79bdac9e is building: [`#3704`](https://buildkite.com/bun/bun/builds/3704)

PR #13414, opened 2024-08-19
fix(ipc/subprocess)  

cirospaciari:
### What does this PR do?

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?
Existing tests

PR #14092, opened 2024-09-22
Fix regression from #13414

Jarred-Sumner:
### What does this PR do?

In #13414, a double deref() was added when the process has already exited by the time the stdin getter was called.

Regression test


robobun:
@Jarred-Sumner, your commit a8fa716d409410bd3a1743a1b2b79945e7068c35 is building: [`#3711`](https://buildkite.com/bun/bun/builds/3711)

cirospaciari (approved):

robobun:
The 80 ms sleep in spawn-stdin-destroy.test.ts added here races on slow Windows runners (EPIPE in build 110853). #41522 replaces it with await child.exited.

Issue #3726, opened 2023-07-21
Passing `import.meta.path` to a macro segfaults


vjpr:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

```
import {foo} from './macro-test.ts' with {type: 'macro'}
console.log('foo:', foo(import.meta.path))
```

```
export function foo(file) {
  // TODO: Extract AST and return it.
  return file
}
```

### What do you see instead?

Segfaults

### Additional information

I want to use a macro to serialize an AST of the current file to expose to it at runtime.

Seems like `import.meta.path` should be statically analyzable...

Regardless, shouldn't segfault.

vjpr:
Actually, even passing a string to this macro causes a segfault.

---

EDIT: So when its in the module scope, it segfaults. But when I move it inside a function it works.

But `import.meta.path` still segfaults...

Jarred-Sumner:
This was fixed in Bun v1.0.1

PR #14107, opened 2024-09-23
Fix unbalanced ref count involving file descriptors passed to Bun.connect

Jarred-Sumner:
### What does this PR do?

This does four things:
1) Fix unbalanced ref count with file descriptors passed to Bun.connect leading to a null-pointer dereference (discovered via undefined behavior sanitizer https://github.com/ziglang/zig/issues/21495)
2) When passing an invalid file descriptor to Bun.connect, we now throw an error appropriately instead of opening a socket that will never receive an event.
3) Fix a potential UAF in connectError when markInactive is called after deref() 
4) Avoid allocating a Subprocess until after it has spawned successfully, and remove a few usages of `undefined`

There are a few new tests, along with an internal binding for `socketpair` to make it easier to test the file descriptor scenarios

### DIFF
--- a/packages/bun-usockets/src/eventing/epoll_kqueue.c
+++ b/packages/bun-usockets/src/eventing/epoll_kqueue.c
@@ -368,7 +368,7 @@ struct us_poll_t *us_poll_resize(struct us_poll_t *p, struct us_loop_t *loop, un
     return new_p;
 }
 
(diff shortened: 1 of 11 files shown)


robobun:
@Jarred-Sumner, your commit 384f90994b221c473df90169dbf2c6d1ca5f94d3 is building: [`#3726`](https://buildkite.com/bun/bun/builds/3726)

cirospaciari (approved):

PR #14109, opened 2024-09-23
Use callbacks instead of a promise that calls a callback for node:fs functions that expect callbacks

Jarred-Sumner:
### What does this PR do?

This fixes a hypothetical execution-order bug in node:fs

Callbacks run before promises.

Since all the node:fs callback functions are currently implemented as wrappers around promises, that means we're delaying execution to the end of the current task. Since these are top-level tasks it should normally not make a difference, but in theory - maybe it could? 

This works by making the API for callback-taking functions the same as promise-taking functions and then we don't need to care which is which most of the time

Untested, but wrote most of the code. Probably doesn't compile.


robobun:
@Jarred-Sumner, your commit e977421eb58ef651831f9f79f2a8be841319840b is building: [`#3734`](https://buildkite.com/bun/bun/builds/3734)

PR #14111, opened 2024-09-23
[install.sh] Try fallback to busybox's unzip

sequencerr:
### What does this PR do?

Try use busybox's unzip instead of infozip

https://discord.com/channels/876711213126520882/876711213126520885/1287567969496469617
> Jarred:
IIRC, we use the default installed unzip
macOS comes with unzip as well

But with Debian doesn't by default. WSL comes with Debian by default also

- [x] Code changes

### DIFF
--- a/src/cli/install.sh
+++ b/src/cli/install.sh
@@ -53,8 +53,14 @@ success() {
     echo -e "${Green}$@ ${Color_Off}"
 }
 
-command -v unzip >/dev/null ||
+uzip=
+if command -v unzip >/dev/null; then
+    uzip=unzip
+elif command -v busybox >/dev/null; then
+    unzip="busybox unzip"
+else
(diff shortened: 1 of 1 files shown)


robobun:
Thanks for this, and sorry that it got no review. #43643 now implements the same fallback: when `unzip` is missing, `install.sh` extracts with `busybox unzip`. It first checks that the BusyBox build has the `unzip` applet. If the applet is missing, it tries `7z`, `bsdtar` and `python3`. `bun upgrade` gets the same fallbacks. The PR is written against the current code and has tests. I ran it with a real BusyBox 1.36.1 build.

This branch conflicts because the script moved to `src/runtime/cli/install.sh`. Closing in favor of #43643, which lists this PR as prior work. Please comment there or reopen if it misses a case you need.

Issue #3768, opened 2023-07-23
Bun.build always assume "react-jsxdev" even when set "react-jsx"


rizrmd:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

run:
```
  await Bun.build({
    entrypoints: ["./index.tsx"],
    outdir: "./build",
    define: { "process.env.NODE_ENV": '"production"' },
  });
```
  make sure to set `{jsx: "react-jsx" } `in your tsconfig
  

### What is the expected behavior?

  You will see the output always use:  
  `jsx_dev_runtime.jsxDEV("div", {` 

### What do you see instead?

instead of 
  `jsx_runtime.jsx("div", {`

### Additional information

Why bun always use "react-jsxdev" instead of following tsconfig ?

mnpenner:
This seems to affect `bun ./somefile.tsx` too (i.e. running it directly)

Ricki-BumbleDev:
Same here. I use a custom JSX implementation (`@nanoweb/jsx`) that doesn't come with a dev runtime, in which case the issue becomes particularly obvious. Only way to make it work currently is to *always* set `BUN_ENV=production` or `NODE_ENV=production`.

Enalmada:
Unfortunately I couldn't even get it to work by setting NODE_ENV to production.  Just before giving up completely on bundling react libraries with Bun, I happened to find that changing external from `['*']` to a manual list of packages in my dependencies works.  I hope this helps someone with a workaround and possibly helps narrow down the bug.

```
  import getExternalDependenciesFromPackageJson from '@enalmada/bun-externals'
  const externalDeps = await getExternalDependenciesFromPackageJson();

  await Bun.build({
    entrypoints: ['./src/client/urql/UrqlWrapper.tsx'],
    outdir: './dist/client',
    target: 'node',
    external: [...externalDeps, './src/client/urql/UrqlWrapper'],
    root: './src/client',
  });
```

tonai:
`NODE_ENV` should be equal to `production` when building. It should be the default.
We should not have to add `NODE_ENV=production` when running `bun build`.

jordanbtucker:
Just ran into this myself. Looks like it was an [intentional design decision](https://github.com/oven-sh/bun/blob/fa952b163cd54d631c0a24f1b3afcec755fb18cb/src/resolver/tsconfig_json.zig#L173-L176).

[...]

PR #14119, opened 2024-09-24
Make prototype pollution attacks harder in most Bun APIs that accept objects

Jarred-Sumner:
### What does this PR do?

Fixes GHSA-5jqh-3v22-8jgj

This makes it so that in most natively-implemented APIs, we ignore prototype chains when calling getters on objects. We deliberately continue to consult prototype chains for Web APIs and node APIs because that would be a clearly breaking change.

This _might_ be a breaking change as-is.

There are probably a few more spots in Bun APIs where we should do this that we aren't, like the shell (which has a JS wrapper).

Let's see if this causes any non-flaky test failures

There is a test for GHSA-5jqh-3v22-8jgj


robobun:
@Jarred-Sumner, your commit 68a5c5eacbb28eaa1c36a877ed69e3f2105b8208 is building: [`#3768`](https://buildkite.com/bun/bun/builds/3768)

Issue #6076, opened 2023-09-26
Poor performance for crypto's privateEncrypt/publicDecrypt


naijun0403:
### What version of Bun is running?

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

You just need to run this code with node and bun.

```javascript
const crypto = require('crypto');
const fs = require('fs');

const data = 'Hello, world!';
const iterations = 1000;

const privateKeyPath = './private_key.pem';
const publicKeyPath = './public_key.pem';

function benchmarkCrypto() {
  console.log(`Benchmarking crypto performance using key files...`);
  const start = Date.now();

  for (let i = 0; i < iterations; i++) {
    const privateKey = fs.readFileSync(privateKeyPath, 'utf8');
    const publicKey = fs.readFileSync(publicKeyPath, 'utf8');

    const encrypted = crypto.privateEncrypt(privateKey, Buffer.from(data));

    const decrypted = crypto.publicDecrypt(publicKey, encrypted);
  }

  const end = Date.now();
  const duration = end - start;
  console.log(`Time taken: ${duration} ms`);
}

benchmarkCrypto();
```

### What is the expected behavior?

At least the speed should be similar to node.

### What do you see instead?

Bun is much slower than node in the RSA encryption and decryption part.

```
Benchmark 1: node test
  Time (mean ± σ):      1.013 s ±  0.015 s    [User: 0.964 s, System: 0.061 s]
  Range (min … max):    0.986 s …  1.036 s    10 runs

Benchmark 2: bun test.js
  Time (mean ± σ):     14.536 s ±  0.247 s    [User: 14.638 s, System: 0.296 s]
  Range (min … max):   14.147 s … 15.020 s    10 runs

Summary
  node test ran
   14.35 ± 0.32 times faster than bun test.js
```

### Additional information

Obviously, I also evaluated the file I/O process when benchmarking. however. This does not affect performance. Bun is slightly faster than node.

I used RSA 2048.

Jarred-Sumner:
These are currently from the browserify polyfills which means they aren't implemented in native code and are much slower as a result. We need to make them use BoringSSL instead (as we have done for other parts of `node:crypto`)

naijun0403:
Okay, I'll always be waiting for news

gornostay25:
@naijun0403 @Jarred-Sumner Any updates?

[...]

PR #14125, opened 2024-09-24
Switch RSA asymmetric sign implementation to BoringSSL

wpaulino:
### What does this PR do?

We port over the remaining `crypto.privateEncrypt` and `crypto.publicDecrypt`, allowing us to get rid of the browserify `public-encrypt` dependency implementing them. All relevant test coverage from Node.js was already ported over, which we still adhere to.

I had the following benchmarks on an Apple M1 Pro:

This branch:

```
benchmark                                  time (avg)             (min … max)       p75       p99      p999
----------------------------------------------------------------------------- -----------------------------
RSA sign RSA_PKCS1_PADDING round-trip     832 µs/iter       (816 µs … 915 µs)    834 µs    893 µs    915 µs
```

bun/main:

```
benchmark                                  time (avg)             (min … max)       p75       p99      p999
----------------------------------------------------------------------------- -----------------------------
RSA sign RSA_PKCS1_PADDING round-trip  26'997 µs/iter (26'657 µs … 27'551 µs) 27'119 µs 27'551 µs 27'551 µs
```

Fixes https://github.com/oven-sh/bun/issues/6076.

### DIFF
--- a/bench/crypto/asymmetricSign.js
+++ b/bench/crypto/asymmetricSign.js
@@ -1,0 +1,24 @@
+import { bench, run } from "mitata";
+const crypto = require("node:crypto");
+
+const keyPair = crypto.generateKeyPairSync("rsa", {
+  modulusLength: 2048,
+  publicKeyEncoding: {
+    type: "spki",
+    format: "pem",
+  },
+  privateKeyEncoding: {
+    type: "pkcs8",
+    format: "pem",
+  },
+});
+
+// Max message size for 2048-bit RSA keys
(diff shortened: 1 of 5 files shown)


Jarred-Sumner:
Looks great. We could remove the getIfPropertyExists calls and make the internal function accept the key object and padding more directly. That would probably end up a little faster because it avoids creating the Identifier for `"padding"` and `"key"` each time (creating an identifier involves hashing it, reading it in a threadlocal hash table, and potentially inserting it into that hash table) 

Is there more code we could delete?

Issue #3789, opened 2023-07-24
Bun install stuck while installing nextjs project dependencies


divyam234:
### What version of Bun is running?

0.7.0

### What platform is your computer?

Linux 5.15.0-76-generic x86_64 x86_64

### What is the expected behavior?

Bun install should succeed

### What do you see instead?

Bun  stuck with 100% cpu usage

Electroid:
Hi, @divyam234 was this an actual issue? If so, we'd want to fix it.

divyam234:
@Electroid It was issue due to peerDependencies in packages which was targeting react 17 or lower. For npm  using  legacy-peer-deps bypasses  this but in Bun installation was stuck .Although bun ignores peerDependencies by default don't know why this was happening.For now I have removed the package which was causing issue.

PR #14128, opened 2024-09-24
no need to cache protocol in NewServer when its statically known

nektro:


robobun:
@nektro, your commit f177f7c1a02146883d0ef31687e31db79626e6e8 is building: [`#3789`](https://buildkite.com/bun/bun/builds/3789)

src/bun.js/api/server.zig
```diff
@@ -6467,11 +6466,8 @@ pub fn NewServer(comptime NamespaceType: type, comptime ssl_enabled_: bool, comp
         }
 
         pub fn getProtocol(this: *ThisServer, globalThis: *JSGlobalObject) JSC.JSValue {
-            if (this.cached_protocol.isEmpty()) {
-                this.cached_protocol = bun.String.createUTF8(if (ssl_enabled) "https" else "http");
-            }
-
-            return this.cached_protocol.toJS(globalThis);
+            _ = this;
+            return ZigString.static(if (ssl_enabled) "https" else "http").toJS(globalThis);
```
cirospaciari:
```suggestion
            return String.static(if (ssl_enabled) "https" else "http").toJS(globalThis);
```

cirospaciari (changes requested):
We should prefer bun.String.static over ZigString

cirospaciari (approved):

Issue #3790, opened 2023-07-24
Implement performance.mark(), performance.measure()


birkskyum:
These api's [aren't implemented in Bun](https://github.com/oven-sh/bun/issues/250#issuecomment-1240393129). 

> This is partially implemented, performance.now() exists but not performance.mark(), performance.measure(), nor the other Node-specific functions (they throw a not implemented error)

birkskyum:
Apparently Jest rely on this as well

Jordan-Hall:
@paperdave  this is the last major barrier for NX, Any chance we can stub this. I'm not good with bun internal so wouldn't know where to start :( Stubbing this will help with a few libraries i believe

paperclover:
can you retry on canary. these methods already were stubbed I think.

Jordan-Hall:
@paperdave  THANK YOU. It's worked no more patching NX, with a custom runner  #3786  is now supported

ImLunaHey:
next returns `NaN` because of this 😅 

```
➜  watcher git:(main) bun --revision
1.0.18+a648ed9e6
➜  watcher git:(main) bun --bun next dev
   ▲ Next.js 14.0.4
   - Local:        http://localhost:3000

 ✓ Ready in NaNms
 ○ Compiling / ...
 ✓ Compiled / in 1022ms (491 modules)
 ✓ Compiled in 752ms (235 modules)
^C
                                                                                                                                                                      
➜  watcher git:(main) bun --node next dev
   ▲ Next.js 14.0.4
   - Local:        http://localhost:3000

 ✓ Ready in 2.2s
 ○ Compiling / ...
 ✓ Compiled / in 619ms (491 modules)
 ✓ Compiled in 109ms (235 modules)
^C
```

birkskyum:
Believe this is closed by
- https://github.com/oven-sh/bun/pull/8011

Duplicate issue
- https://github.com/oven-sh/bun/issues/4708

PR #14129, opened 2024-09-24
add missing defers to JSBundler.Plugin.hasAnyMatches

nektro:
this one i wasnt 100% confident on but the three functions below it all follow a similar pattern and do the defers on their strings


robobun:
@nektro, your commit 032fe7bdd7e12abb648116422148493c05c95b38 is building: [`#3790`](https://buildkite.com/bun/bun/builds/3790)

Issue #3791, opened 2023-07-24
Support VitePress


birkskyum:
VitePress - [Website](https://vitepress.dev/) - [GitHub](https://github.com/vuejs/vitepress)

### What version of Bun is running?

0.7.0

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

* `bun x vitepress init`
* Step through wizard with defaults
* `bun add -d vitepress`
* `bun --bun run docs:dev`

### What is the expected behavior?

```

➜ bun --bun run docs:dev        

> docs:dev
> vitepress dev

  vitepress v1.0.0-beta.6

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h to show help

```

### What do you see instead?

```
➜ bun --bun run docs:dev
$ vitepress dev
failed to load config from /Users/admin/repos/vitepress-test-2/.vitepress/config.ts
failed to start server. error:
Error: config must export or return an object.
    at <anonymous> (/Users/admin/repos/vitepress-test-2/node_modules/vite/dist/node/chunks/dep-a8e37fae.js:66014:15)
error: script "docs:dev" exited with code 1 (SIGHUP)
```

<details>
<summary>Last part of debug log</summary>

```zig
Loader [resolve] node:net
[JSC] resolve (src/bun.js/bindings/exports.zig:69)
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
[JSC] ref (src/string.zig:107)
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
[JSC] ref (src/string.zig:107)
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
[JSC] deref (src/string.zig:95)
[JSC] deref (src/string.zig:95)
Loader [resolve] node:zlib
[JSC] resolve (src/bun.js/bindings/exports.zig:69)
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
[JSC] ref (src/string.zig:107)
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
[JSC] ref (src/string.zig:107)
[JSC] ascii (src/bun.js/bindings/bun-simdutf.zig:105)
[JSC] deref (src/string.zig:95)
[JSC] deref (src/string.zig:95)
Loader [resolve] node:http
[JSC] resolve (src/bun.js/bindings/exports.zig:69)
[...]

PR #14130, opened 2024-09-24
fix memory leak in Bun.shellEscape return value

nektro:


robobun:
@nektro, your commit 39c85e05db4b178ab514a2c7e30a58969c5423a9 is building: [`#3791`](https://buildkite.com/bun/bun/builds/3791)

PR #14150, opened 2024-09-25
chore(test): use local builded bun to test

AielloChan:
### What does this PR do?

By read the **scripts/runner.node.mjs** code, I find that, the script in **package.json** not just use current builded bun executable, will always call the global installed one. So I just try to fix it.

Caution: Need check if this change may broken some logic in CI/CD?

- [X] I just read the code and run passed in my local environment.


AielloChan:
Seems @Electroid already fixed this one 👍, I'll close this PR.

Issue #3893, opened 2023-07-30
Plugin not able to import namespaced URLs


wobsoriano:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. Clone [this](https://github.com/wobsoriano/https-plugin-repro) reproduction repo
2. Run `bun run play`

### What is the expected behavior?

The console should run the function imported from lodash.

### What do you see instead?

```bash
error: FileNotFound reading "https://unpkg.com/lodash-es@4.17.15/lodash.js"

error: script "play" exited with code 1 (SIGHUP)
```

### Additional information

The plugin is based on esbuild's [http plugin example](https://esbuild.github.io/plugins/#http-plugin) 

```ts
import { BunPlugin } from "bun"

export function httpsPlugin(): BunPlugin {
  return {
    name: 'https-plugin',
    async setup(build) {

      build.onResolve({ filter: /^https?:\/\// }, args => {
        return {
          path: args.path,
          namespace: 'http-url',
        }
      })

      build.onResolve({ filter: /.*/, namespace: 'http-url' }, args => {
        return {
          path: new URL(args.path, args.importer).toString(),
          namespace: 'http-url',
        }
      })

      build.onLoad({ filter: /.*/, namespace: 'http-url' }, async (args) => {
        const response = await fetch(args.path)
        const contents = await response.text()

        return {
          contents,
          loader: 'js'
        }
      })
    }
  }
}
```

wobsoriano:
Duplicate of #3894

PR #14154, opened 2024-09-25
fix small leak in node:process.execArgv getter

nektro:
this getter is cached so the blast radius is small but this patch still frees up the stray `bun.String` instances created during the initial call


robobun:
@nektro, your commit 04abc7a36405c9b36af3ae83464379317586d41e is building: [`#3893`](https://buildkite.com/bun/bun/builds/3893)

Issue #3925, opened 2023-08-02
Implement `URL.createObjectURL`


truongan07:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Linux 6.3.6-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

const blob = new Blob(['inline worker'], {type: 'text/plain'});

console.log(URL.createObjectURL(blob)); // undefined

### What is the expected behavior?

A reference url like blob:null/dd56aa40-18c1-4b56-a12b-f4346251b814

Jarred-Sumner:
Confirming this isn’t implemented yet

estubmo:
Should URL be marked as fully implemented when this is missing?
https://bun.sh/docs/runtime/nodejs-apis#url

takinabradley:
I would have preferred it wasn't. 

When I needed this, I think I got around it by using `Bun.pathToFileURL` to nab me a file URL. May work for someone else who needs file URLs.

truongan07:
> I would have preferred it wasn't.
> 
> When I needed this, I think I got around it by using `Bun.pathToFileURL` to nab me a file URL. May work for someone else who needs file URLs.

Bun.pathToFileURL is so cool and i use more tiny scripts in worker at anywhere on module and more modules. I can't create more files to only use worker.  I create a function in library and call it. So, needed **inline worker**

cyfung1031:
Any roadmap for implementing this? This is somehow important for using WebWorker.

truongan07:
This method is very important. priority implement please

truongan07:
@Jarred-Sumner @Electroid high priority implementation, please

PR #14166, opened 2024-09-25
cmake: Use toolchain with dependencies

Electroid:
### What does this PR do?

This changes how dependencies are built to use a toolchain file, instead of the previous method, which was error-prone.


robobun:
@Electroid, your commit dad8ef91f3ecfe0e3939a62a5f391e008a6487a9 is building: [`#3925`](https://buildkite.com/bun/bun/builds/3925)

PR #14168, opened 2024-09-25
fix utf8 handling when importing json

### LINKED ISSUE #10851: Importing .json with non-ascii characters within `bun build --target=bun` does not encode to disk as latin1

1.1.7+b0b7db5c0 (issue since v1.1.5)

```js
// @bun
console.log("测试");
```

p/s: running without `// @bun`
```bash
$ bun test.js
测试
```

### What do you see instead?

```bash
$ bun test.js
æµè¯
```

### Additional information

this issue exists since v1.1.5 (i just confirmed the cause recently)

snoglobe:
### What does this PR do?

this closes #10851 - when importing JSON files, strings were encoded as utf8 which caused problems when evaluated in js core

- [x] Code changes

### DIFF
--- a/src/StandaloneModuleGraph.zig
+++ b/src/StandaloneModuleGraph.zig
@@ -999,7 +999,7 @@ pub const StandaloneModuleGraph = struct {
             bun.JSAst.Expr.Data.Store.reset();
             bun.JSAst.Stmt.Data.Store.reset();
         }
-        var json = bun.JSON.ParseJSON(&json_src, &log, arena) catch
+        var json = bun.JSON.ParseJSON(&json_src, &log, arena, false) catch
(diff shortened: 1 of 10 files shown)


paperclover (approved):
im realizing an existing problem with this code, which is that there are 5 or so of these JSON parse functions which are all identical but have slightly different things. the perfect world makes this one single function, which can be configured as needed.

but that has no runtime difference, and neither does this. lgtm

dylan-conway:
> im realizing an existing problem with this code, which is that there are 5 or so of these JSON parse functions which are all identical but have slightly different things. the perfect world makes this one single function, which can be configured as needed.

I agree, json_parser.zig needs a cleanup

Issue #3931, opened 2023-08-02
subsequent runs cause seg fault or bad memory pointer


danthegoodman1:
### What version of Bun is running?

0.7.1

### What platform is your computer?

Darwin 22.5.0 x86_64 i386

### What steps can reproduce the bug?

```ts
import { appendFile, readFile, readdir, unlink } from "fs/promises"
import { Node } from "./types"

const channel = "techsmith314"

const files = (await readdir(`${channel}/raw_data`)).filter((e) => e.endsWith(".json"))

let outFile = `${channel}/prepared_data.jsonl`

try {
  await unlink(outFile)
} catch (error) {}

let buffer = ""
for (const file of files) {
  const content = await readFile(`${channel}/raw_data/${file}`, "utf-8")
  const lines = content.split("\n").filter((l) => l !== "")
  for (let i = 0; i < lines.length; i++) {
    const line = lines[i]
    const node: Node = JSON.parse(line)
    const msg = `${node.commenter.displayName}: ${node.message.messageFragments.map((m) => m.text)}\n`
    buffer += msg
    if (buffer.length > 12_000) {
      await appendFile(outFile, JSON.stringify({t: buffer}) + "\n")
    }
    // pop out the first one
    buffer = buffer.split("\n").slice(1).join("\n")
  }
}

console.log("done!")
```

### What is the expected behavior?

Should work every time

### What do you see instead?

```
➜  dangoodman: ~/tangiaCode/ChatLLMs/chat bun run prepare.ts                                                  1:07PM
1 | import { appendFile, readFile, readdir, unlink } from "fs/promises"
2 | import { Node } from "./types"
3 |
4 | const channel = "techsmith314"
5 |
6 | const files = (await readdir(`${channel}/raw_data`)).filter((e) => e.endsWith(".json"))
                                                                                         ^
TypeError: e.endsWith is not a function. (In 'e.endsWith(".json")', 'e.endsWith' is 1.0028852895063658e-303)
      at /Users/dangoodman/tangiaCode/ChatLLMs/chat/prepare.ts:6:86
      at /Users/dangoodman/tangiaCode/ChatLLMs/chat/prepare.ts:6:6
➜  dangoodman: ~/tangiaCode/ChatLLMs/chat curl -fsSL https://bun.sh/install | bash                            1:07PM
######################################################################## 100.0%
bun was installed successfully to ~/.bun/bin/bun
Run 'bun --help' to get started
[...]

PR #14169, opened 2024-09-25
use bun.String.static instead of ZigString when immediately converting to JSValue

nektro:
thankfully this wasnt used as much as i was expecting but this should save all of these instances from having an unnecessary `JSC.ZigString` instance in `.rodata` as well as a `ZigString.static` generic instantiation in `.text`.

found with `ZigString.static\("([^"]+)"\).toJS` and `ZigString.init\("([^"]+)"\).toJS`
being replaced with `bun.String.static("$1").toJS`

future work: add support for this in the linter so new instances dont make their way in


robobun:
@nektro, your commit b5e61dd58bd20802cc218d8e952464a8f3baa7c9 is building: [`#3931`](https://buildkite.com/bun/bun/builds/3931)

cirospaciari (approved):
LGTM lets merge after CI

PR #14175, opened 2024-09-26
Use _exit on error, but restore stdio first

Jarred-Sumner:
### What does this PR do?

Use _exit on error, but restore stdio first

When a DLL file fails to load on Windows, it usually ends up causing a crash during the exit handler

Let's see what CI says and do a little manual testing


robobun:
@Jarred-Sumner, your commit c3decd4e1a3cb9e784ac68d22a13bd2182b3a8e3 is building: [`#3958`](https://buildkite.com/bun/bun/builds/3958)

PR #3994, opened 2023-08-05
[install] handle `workspace:*` correctly

alexlamsl:
- parse as path so it works on unversioned workspaces
- fix missed storage of workspace version

fixes #3985

### What does this PR do?

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

If Zig files changed:

- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [x] I or my editor ran `zig fmt` on the changed files
- [x] I included a test for the new code, or an existing test covers it
- [ ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed

PR #14198, opened 2024-09-27
fix Buffer method aliases not being equal to each other

nektro:
pulled out of https://github.com/oven-sh/bun/pull/13757 to merge independently


robobun:
@nektro, your commit c0cb37f64aee792bf7e7756cfa5d1c82033b7b08 is building: [`#3994`](https://buildkite.com/bun/bun/builds/3994)

PR #14199, opened 2024-09-27
fix Buffer.fill with a non-null empty fill including uninitialized bytes

nektro:
pulled out of https://github.com/oven-sh/bun/pull/13757 to merge independently

added new test


robobun:
@nektro, your commit 22f194241195f105751f88b7b981313fac24c240 is building: [`#3995`](https://buildkite.com/bun/bun/builds/3995)

Issue #4044, opened 2023-08-07
Bun equivalent of Node's `--preserve-symlinks`


sstur:
### What version of Bun is running?

0.7.3

### What platform is your computer?

Darwin 22.2.0 arm64 arm

### What steps can reproduce the bug?

Symlink a file that imports from another module using relative paths, e.g.
```ts
// src/foo/file.ts (symlink'd from src/bar/file.ts)
import { value } from './file2'
console.log(value)
```

This will import from `/src/bar/file2.ts` however TS thinks it is loading `/src/foo/file2.ts` leading to a mismatch between what my editor thinks `value` is vs what it actually is at runtime.

### What is the expected behavior?

Bun should support some way to opt-in to the TypeScript behavior, the equivalent of Node's `--preserve-symlinks` flag.

### What do you see instead?

Mismatch between runtime value and what TS thinks at edit/compile time.

### Additional information

_Note: The following is what I posted on Discord [here](https://discord.com/channels/876711213126520882/1138128619768647680/1138128619768647680), and was asked to create this issue._

By default Node, and evidently also Bun, resolves imports inside symlink'd files/folders differently than TypeScript's default.

TypeScript:
- resolve imports inside symlink'd files/folders relative to the path of the symlink

Node (by default) and Bun:
- resolve imports inside symlink'd files/folders relative to the path of the original file

However, Node has two ways to switch from the default behavior:
1. `node --preserve-symlinks myfile.ts`
2. `NODE_PRESERVE_SYMLINKS=1 node myfile.ts`

Is there any equivalent in Bun, and if not is there a workaround or is this on the roadmap?

Reference:
https://nodejs.org/docs/latest-v20.x/api/cli.html#--preserve-symlinks

sirenkovladd:
I think it might be related https://github.com/oven-sh/bun/issues/3748

Gakai:
I encountered this problem as well (with Bun v1.1.30). I have multiple packages/workspaces inside my repo and various .test.ts files that apply to multiple packages. I also want to keep these files in sync, so I tried to symlink them into the packages from a central location.

Unfortunately because of this issue, I can't do that, because every import statement fails with
```
error: Cannot find module "./index.ts" from "~/my-monorepo/shared/index.test.ts
```
even though the symlink lies in `~/my-monorepo/packages/package-a/index.test.ts`.

[...]

PR #14221, opened 2024-09-28
stub out WebAssembly.{compile|instantiate}Streaming

nektro:


robobun:
@nektro, your commit d22ab432f928e55cec3ddae45d74c9df5699f970 is building: [`#4044`](https://buildkite.com/bun/bun/builds/4044)

Jarred-Sumner:
We should not do this. Many libraries will attempt to use it if it's defined, and then they will fail instead of using a fallback implementation (which many do implement).

Issue #6227, opened 2023-10-02
`require('node:module')._preloadModules` is not a function.


imcm7:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Linux 6.2.0-33-generic x86_64 x86_64

### What steps can reproduce the bug?

https://github.com/imcm7/nuxt-db/tree/main/server
install mikro-orm and `bun -b db:update`

### What do you see instead?

@Jarred-Sumner with -b also not run, some days ago all was ok

```
process[exports.REGISTER_INSTANCE] = service;
132 |     // Register the extensions.
133 |     registerExtensions(service.options.preferTsExts, service.extensions.compiled, service, originalJsHandler);
134 |     (0, cjs_resolve_hooks_1.installCommonjsResolveHooksIfNecessary)(service);
135 |     // Require specified modules before start-up.
136 |     module_1.Module._preloadModules(service.options.require);
        ^
TypeError: module_1.Module._preloadModules is not a function. (In 'module_1.Module._preloadModules(service.options.require)', 'module_1.Module._preloadModules' is undefined)
      at register (/home/node/nuxt/node_modules/ts-node/dist/index.js:136:4)
      at /home/node/nuxt/node_modules/@mikro-orm/core/utils/ConfigurationLoader.js:115:28
      at registerTsNode (/home/node/nuxt/node_modules/@mikro-orm/core/utils/ConfigurationLoader.js:104:32)
      at /home/node/nuxt/node_modules/@mikro-orm/cli/CLIConfigurator.js:28:33
      at processTicksAndRejections (:55:76)
error: script "db:update" exited with code 1 (SIGHUP)

``

### Additional information

_No response_

imcm7:
you can pass all you want but it not work https://github.com/oven-sh/bun/issues/6226

imcm7:
Some think changed and broke

B4nan:
I think you should drop the [`useTsNode` option](https://github.com/imcm7/nuxt-db/blob/main/package.json#L55) from your package.json, that option triggers the `ts-node.register()` which you don't need with bun. Instead, you need the new `alwaysAllowTs` option, which will make the ORM respect the TS files, but as opposed to the `useTsNode` it won't trigger the `ts-node.register()` call.

imcm7:
@B4nan but I use bun node anlways, I mean without --bun.
[...]

Issue #4060, opened 2023-08-08
Support Angular CLI


birkskyum:
Angular CLI - [Website](https://angular.io/cli) - [GitHub](https://github.com/angular/angular-cli)

### What version of Bun is running?

0.7.3

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

Execute the following
```
bun --bun x ng --help
```

### What is the expected behavior?

The below works if `--bun` is removed

```zig

➜ bun --bun x ng --help      
ng <command>

Commands:
  ng add <collection>            Adds support for an external library to your project.
  ng analytics                   Configures the gathering of Angular CLI usage metrics.
  ng build [project]             Compiles an Angular application or library into an output directory named dist/ at the given output path.                                                            [aliases: b]
  ng cache                       Configure persistent disk cache and retrieve cache statistics.
  ng completion                  Set up Angular CLI autocompletion for your terminal.
  ng config [json-path] [value]  Retrieves or sets Angular configuration values in the angular.json file for the workspace.
  ng deploy [project]            Invokes the deploy builder for a specified project or for the default project in the workspace.
  ng doc <keyword>               Opens the official Angular documentation (angular.io) in a browser, and searches for a given keyword.                                                                [aliases: d]
  ng e2e [project]               Builds and serves an Angular application, then runs end-to-end tests.                                                                                                [aliases: e]
  ng extract-i18n [project]      Extracts i18n messages from source code.
  ng generate                    Generates and/or modifies files based on a schematic.                                                                                                                [aliases: g]
  ng lint [project]              Runs linting tools on Angular application code in a given project folder.
  ng new [name]                  Creates a new Angular workspace.                                                                                                                                     [aliases: n]
  ng run <target>                Runs an Architect target with an optional custom builder configuration defined in your project.
  ng serve [project]             Builds and serves your application, rebuilding on file changes.                                                                                                      [aliases: s]
  ng test [project]              Runs unit tests in a project.                                                                                                                                        [aliases: t]
  ng update [packages..]         Updates your workspace and its dependencies. See https://update.angular.io/.
  ng version                     Outputs Angular CLI version.                                                                                                                                         [aliases: v]

Options:
  --help  Shows a help message for this command in the console.                                                                                                                                          [boolean]

For more information, see https://angular.io/cli/.


```

### What do you see instead?

```sh
55 |     if (process.stdout instanceof tty_1.WriteStream) {
           ^
[...]

PR #14227, opened 2024-09-28
Refactor node:module 

### LINKED ISSUE #8974: require.resolve unable to find node module

1.0.27+c34bbb2e3

Darwin 23.2.0 arm64 arm

1. bun init
2. bun install @tevm/contract 
3. Add following script and run it with bun and node. Node will return the path while bun will fail to return the correct path

```typescript
import { createRequire } from 'module';

const require = createRequire(
  `${process.cwd()}/`
);

console.log(
  require.resolve('@tevm/contract')
)
```

This should be able to resolve a node module relative to process.cwd() like node does

### What do you see instead?

Node is able to resolve this while bun fails

### Additional information

telegram-cloud-photo-size-1-5084681159350266970-y

Jarred-Sumner:
### What does this PR do?

Refactor node:module to add missing properties and be a lot closer to what Node does

Main goal of this PR is to unblock optimizing _resolveFilename, which is used heavily by Next.js and shows up near the top when profiling.

Fixes https://github.com/oven-sh/bun/issues/6227
Fixes #8974

Draft. Something is wrong currently with the static hash table causing it to never materialize the properties.


robobun:
@Jarred-Sumner, your commit a2444f3e7eb920f1860605ca0a2c98d33a186d9a is building: [`#4060`](https://buildkite.com/bun/bun/builds/4060)

src/bun.js/modules/NodeModuleModule.h
dylan-conway, line 1: this file does not seem to be formatted correctly

dylan-conway (approved):

PR #14232, opened 2024-09-28
Introduce bytecode caching, polish `"cjs"` bundler output format

Jarred-Sumner:
### What does this PR do?

This adds a new `--bytecode` flag, and `// @bun @bun-cjs` comment that marks a file as already transpiled with CommonJS. 

```
bun build --compile --bytecode <script>
bun build --target=bun --bytecode --outdir=out <script>
```

2x faster `tsc --help`

image

When you use `--bytecode` without `--compile`, it adds `.jsc` files which contain the compiled bytecode. When the bytecode is unable to be loaded, it uses the source code. Bytecode compilation currently has no impact on async functions and no impact on ESM.

Added tests.

### DIFF
--- a/src/bun.js/api/JSBundler.zig
+++ b/src/bun.js/api/JSBundler.zig
@@ -70,6 +70,8 @@ pub const JSBundler = struct {
         public_path: OwnedString = OwnedString.initEmpty(bun.default_allocator),
         conditions: bun.StringSet = bun.StringSet.init(bun.default_allocator),
         packages: options.PackagesOption = .bundle,
+        format: options.Format = .esm,
+        bytecode: bool = false,
 
         pub const List = bun.StringArrayHashMapUnmanaged(Config);
 
@@ -151,8 +153,23 @@ pub const JSBundler = struct {
                 }
             }
 
+            if (try config.getOwnOptional(globalThis, "bytecode", bool)) |bytecode| {
+                this.bytecode = bytecode;
+
+                if (bytecode) {
+                    // Default to CJS for bytecode, since esm doesn't really work yet.
+                    this.format = .cjs;
+                    this.target = .bun;
+                }
+            }
+
             if (try config.getOwnOptionalEnum(globalThis, "target", options.Target)) |target| {
                 this.target = target;
+
+                if (this.target != .bun and this.bytecode) {
+                    globalThis.throwInvalidArguments("target must be 'bun' when bytecode is true", .{});
+                    return error.JSError;
+                }
             }
 
             var has_out_dir = false;
@@ -184,12 +201,11 @@ pub const JSBundler = struct {
             }
 
             if (try config.getOwnOptionalEnum(globalThis, "format", options.Format)) |format| {
-                switch (format) {
-                    .esm => {},
-                    else => {
-                        globalThis.throwInvalidArguments("Formats besides 'esm' are not implemented", .{});
-                        return error.JSError;
-                    },
+                this.format = format;
+
+                if (this.bytecode and !(this.format == .esm or this.format == .cjs)) {
+                    globalThis.throwInvalidArguments("format must be 'esm' or 'cjs' when bytecode is true", .{});
+                    return error.JSError;
                 }
             }
 
@@ -1042,6 +1058,11 @@ pub const BuildArtifact = struct {
         @"use client",
         @"use server",
         sourcemap,
+        bytecode,
+
+        pub fn isFileInStandaloneMode(this: OutputKind) bool {
+            return this != .sourcemap and this != .bytecode;
+        }
     };
 
     pub fn deinit(this: *BuildArtifact) void {
--- a/src/bun.js/api/JSBundler.zig
+++ b/src/bun.js/api/JSBundler.zig
@@ -166,7 +166,7 @@ pub const JSBundler = struct {
             if (try config.getOwnOptionalEnum(globalThis, "target", options.Target)) |target| {
                 this.target = target;
 
-                if (this.target != .bun and this.bytecode) {
+                if (target != .bun and this.bytecode) {
                     globalThis.throwInvalidArguments("target must be 'bun' when bytecode is true", .{});
                     return error.JSError;
                 }
@@ -203,8 +203,8 @@ pub const JSBundler = struct {
             if (try config.getOwnOptionalEnum(globalThis, "format", options.Format)) |format| {
                 this.format = format;
 
-                if (this.bytecode and !(this.format == .esm or this.format == .cjs)) {
-                    globalThis.throwInvalidArguments("format must be 'esm' or 'cjs' when bytecode is true", .{});
+                if (this.bytecode and format != .cjs) {
+                    globalThis.throwInvalidArguments("format must be 'cjs' when bytecode is true. Eventually we'll add esm support as well.", .{});
                     return error.JSError;
                 }
             }
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -2747,12 +2747,12 @@ pub const VirtualMachine = struct {
                 return promise;
             }
 
-            const promise = JSModuleLoader.loadAndEvaluateModule(this.global, &String.init(main_file_name)) orelse return error.JSError;
+            const promise = JSModuleLoader.loadAndEvaluateModule(this.global, &String.fromBytes(this.main)) orelse return error.JSError;
             this.pending_internal_promise = promise;
             JSC.JSValue.fromCell(promise).ensureStillAlive();
             return promise;
         } else {
-            const promise = JSModuleLoader.loadAndEvaluateModule(this.global, &String.init(this.main)) orelse return error.JSError;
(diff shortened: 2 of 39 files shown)


robobun:
@Jarred-Sumner, your commit b67a063b0dcf1dc00f6944a57b0c97b5f55d37ac is building: [`#4068`](https://buildkite.com/bun/bun/builds/4068)

test/bundler/esbuild/default.test.ts
```diff
@@ -2797,14 +2799,16 @@ describe("bundler", () => {
...
+      `,
     },
     format: "cjs",
-    bundleWarnings: {
-      "/entry.js": [`"import.meta" is not available with the "cjs" output format and will be empty`],
-    },
+    target: "node",
     run: {
-      stdout: "undefined undefined",
+      stdout: "true true",
```
Jarred-Sumner, line 2811: @paperdave

test/bundler/esbuild/default.test.ts
```diff
@@ -3784,6 +3790,13 @@ describe("bundler", () => {
     },
     target: "node",
     format: "cjs",
+    bundleErrors: {
```
Jarred-Sumner, line 3793: @paperdave

test/bundler/esbuild/default.test.ts
```diff
@@ -4630,20 +4643,18 @@ describe("bundler", () => {
...
-  //       console.log('is main:', require.main === module)
-  //       console.log(require('./is-main'))
-  //       console.log('cache:', require.cache);
-  //     `,
-  //     "/is-main.js": `module.exports = require.main === module`,
-  //   },
-  //   format: "cjs",
-  //   platform: "node",
-  // });
+  itBundled("default/RequireMainCacheCommonJS", {
```
Jarred-Sumner, line 4646: @paperdave we fixed this since that test was commented out, pretty sure

test/bundler/esbuild/importstar.test.ts
```diff
@@ -561,6 +561,7 @@ describe("bundler", () => {
     format: "iife",
   });
   itBundled("importstar/ExportSelfIIFEWithName", {
+    todo: true,
```
Jarred-Sumner, line 564: i'm scoping out support for IIFE in this PR

src/js/builtins/Module.ts
```diff
@@ -76,9 +76,7 @@ export function overridableRequire(this: CommonJSModuleRecord, id: string) {
     // If we can pull out a ModuleNamespaceObject, let's do it.
     if (esm?.evaluated && (esm.state ?? 0) >= $ModuleReady) {
       const namespace = Loader.getModuleNamespaceObject(esm!.module);
-      return (mod.exports =
```
Jarred-Sumner, line 79: cc @paperdave @dylan-conway please take a look at this

src/bun.js/api/JSBundler.zig
```diff
@@ -184,12 +201,11 @@ pub const JSBundler = struct {
...
-                switch (format) {
-                    .esm => {},
-                    else => {
-                        globalThis.throwInvalidArguments("Formats besides 'esm' are not implemented", .{});
-                        return error.JSError;
-                    },
+                this.format = format;
+
+                if (this.bytecode and !(this.format == .esm or this.format == .cjs)) {
+                    globalThis.throwInvalidArguments("format must be 'esm' or 'cjs' when bytecode is true", .{});
```
paperclover: `since esm doesn't really work yet.` this should not allow esm if it does not work.

src/js/builtins/Module.ts
paperclover: what is the reasoning behind this change. it is breaking.

src/bundler/bundle_v2.zig
```diff
@@ -3748,7 +3759,7 @@ pub const JSMeta = struct {
 };
 
 pub const Graph = struct {
-    // TODO: consider removing references to this in favor of bundler.options.code_splitting
```
paperclover, line 3751: removing this comment is incorrect. i had put off deciding if removing code_splitting from this structure was possible, but i just checked again and i realize it can be removed. please don't remove TODO tasks without actually performing them.

test/bundler/esbuild/default.test.ts
paperclover: ooh i like this. `import.meta` *should* just work, aliasing to `__dirname` and `__filename`.

this test however should

- run in node to ensure `import.meta` is not actually emitted. you can do this by setting `runtime: "node"` in the run object.
- fs.existsSync is not a good enough behavior test. what if the path is `/`. it should be a string that is checked for exactly. if you turn the config object passed to itBundled into a function with signature `(({ root }) => ({ files: ...` you can use that path to write exact expectations for the absolute paths. see `tsconfig/Path` for more details on this api.

though after irl mention of this, it seems we *are* emitting import.meta, so maybe we just make this behavior error since that is likely easier.

test/bundler/esbuild/default.test.ts
paperclover: amazing

src/bun.js/javascript.zig
```diff
@@ -2752,7 +2752,7 @@ pub const VirtualMachine = struct {
             JSC.JSValue.fromCell(promise).ensureStillAlive();
             return promise;
         } else {
-            const promise = JSModuleLoader.loadAndEvaluateModule(this.global, &String.init(this.main)) orelse return error.JSError;
+            const promise = JSModuleLoader.loadAndEvaluateModule(this.global, &String.fromBytes(this.main)) orelse return error.JSError;
```
dylan-conway, line 2755: This change looks identical to the original

src/bun.js/module_loader.zig
```diff
@@ -1760,14 +1766,18 @@ pub const ModuleLoader = struct {
...
+                    const bytecode_slice = parse_result.already_bundled.bytecodeSlice();
                     return ResolvedSource{
                         .allocator = null,
                         .source_code = bun.String.createLatin1(parse_result.source.contents),
                         .specifier = input_specifier,
                         .source_url = input_specifier.createIfDifferent(path.text),
                         .already_bundled = true,
                         .hash = 0,
+                        .bytecode_cache = if (bytecode_slice.len > 0) bytecode_slice.ptr else null,
+                        .bytecode_cache_size = if (bytecode_slice.len > 0) bytecode_slice.len else 0,
```
dylan-conway, line 1779:
```suggestion
                        .bytecode_cache_size = bytecode_slice.len,
```

src/bun.js/module_loader.zig
```diff
@@ -573,15 +574,19 @@ pub const RuntimeTranspilerStore = struct {
...
+                const bytecode_slice = parse_result.already_bundled.bytecodeSlice();
                 this.resolved_source = ResolvedSource{
                     .allocator = null,
                     .source_code = bun.String.createLatin1(parse_result.source.contents),
                     .specifier = duped,
                     .source_url = duped.createIfDifferent(path.text),
                     .already_bundled = true,
                     .hash = 0,
+                    .bytecode_cache = if (bytecode_slice.len > 0) bytecode_slice.ptr else null,
+                    .bytecode_cache_size = if (bytecode_slice.len > 0) bytecode_slice.len else 0,
```
dylan-conway:
```suggestion
                    .bytecode_cache_size = bytecode_slice.len,
```

test/bundler/esbuild/default.test.ts
paperclover: great

test/bundler/esbuild/importstar.test.ts
paperclover: reasonable.

test/bundler/expectBundled.ts
```diff
@@ -1584,6 +1598,17 @@ export function itBundled(
   }
   return ref;
 }
+itBundled.only = (id: string, opts: BundlerTestInput) => {
```
paperclover, line 1602: oh wow i didn't know people use only instead of `-t` (code looks fine)

paperclover (changes requested):

src/runtime.zig
```diff
@@ -240,6 +240,7 @@ pub const Runtime = struct {
         /// Use `import.meta.require()` instead of require()?
         /// This is only supported with --target=bun
         use_import_meta_require: bool = false,
+        auto_polyfill_require: bool = false,
```
dylan-conway, line 245: comment explaining what this does would be nice

src/bun.js/api/JSBundler.zig
dylan-conway: I agree, it makes sense to include esm in with this error

Jarred-Sumner: It does work - it just has no impact.

src/js/builtins/Module.ts
Jarred-Sumner: Discussed IRL. The goal of this, originally, was to let you mutate values on the ES module namespace object. But, that doesn't work and has not worked.

Instead, this has the effect of making console.log and (...) not work as well

Issue #11083, opened 2024-05-15
`bun --hot` always leaks memory


paperclover:
### What steps can reproduce the bug?

```
import { spawn } from "bun";
import { writeFileSync } from "node:fs";

let i = 0;
function next() {
  i++;
  writeFileSync(
    "hello.mjs",
    `// This is a medium sized testing file
// ${"contents".repeat(100)}
Bun.gc(true);
console.log('RSS in run %s is %d MB', ${i}, process.memoryUsage().rss / 1_000_000);
console.err('[done]');
`,
  );
}

next();

await using runner = spawn({
  cmd: [
    //
    process.execPath,
    "--hot",
    "run",
    "hello.mjs",
  ],
  stdio: ["ignore", "inherit", "pipe"],
  env: {
    ...process.env,
    NO_COLOR: true,
  },
});

for await (const chunk of runner.stderr) {
  if (new TextDecoder().decode(chunk).includes("[done]")) {
    next();
  }
}
```

### What is the expected behavior?

to not leak memory

### What do you see instead?

<img width="386" alt="image" src="https://github.com/oven-sh/bun/assets/24465214/c9e8f63e-1154-49c4-8205-bb6acc548c12">

Jarred-Sumner:
```
RSS in run 57454 is 127 MB
RSS in run 57455 is 127 MB
RSS in run 57456 is 127 MB
RSS in run 57457 is 127 MB
RSS in run 57458 is 127 MB
RSS in run 57459 is 127 MB
RSS in run 57460 is 127 MB
RSS in run 57461 is 127 MB
RSS in run 57462 is 127 MB
RSS in run 57463 is 127 MB
RSS in run 57464 is 127 MB
RSS in run 57465 is 127 MB
RSS in run 57466 is 127 MB
RSS in run 57467 is 127 MB
RSS in run 57468 is 127 MB
RSS in run 57469 is 127 MB
RSS in run 57470 is 127 MB
RSS in run 57471 is 127 MB
RSS in run 57472 is 127 MB
RSS in run 57473 is 127 MB
RSS in run 57474 is 127 MB
RSS in run 57475 is 127 MB
RSS in run 57476 is 127 MB
RSS in run 57477 is 127 MB
RSS in run 57478 is 127 MB
RSS in run 57479 is 127 MB
RSS in run 57480 is 127 MB
RSS in run 57481 is 127 MB
RSS in run 57482 is 127 MB
RSS in run 57483 is 127 MB
[...]

PR #14260, opened 2024-09-30
'let' statements before using statements are now properly converted into 'var' statements

### LINKED ISSUE #12407: `using` hoisting bug with generator functions in module scope
In Bun v1.1.18

The following input emits a `ReferenceError`:
```ts
const env = {}
using server = Bun.serve({
  port: 0,
  fetch(req) {},
});
for (let k of run()) {

}
function* run() {
  console.log(env)
}
```

Error:
```ts
2 | const env = {}
3 | using server = Bun.serve({
4 |   port: 0,
5 |   fetch(req) {},
6 | });
7 | for (let k of run()) {
                  ^
TypeError: undefined is not an object (evaluating 'run()')
      at /private/tmp/b.mjs:7:15

Bun v1.1.18-canary.1+da1b3d200 (macOS arm64)
```

Output:
```ts
var __dispose = Symbol.dispose || Symbol.for("Symbol.dispose");
var __asyncDispose = Symbol.dispose || Symbol.for("Symbol.dispose");
var __using = (stack, value, async) => {
  if (value != null) {
    if (typeof value !== "object" && typeof value !== "function")
      throw TypeError('Object expected to be assigned to "using" declaration');
    var dispose;
    if (async)
      dispose = value[__asyncDispose];
    if (dispose === undefined)
      dispose = value[__dispose];
    if (typeof dispose !== "function")
      throw TypeError("Object not disposable");
    stack.push([async, dispose, value]);
  } else if (async) {
    stack.push([async]);
  }
  return value;
};
var __callDispose = (stack, error, hasError) => {
  var E = typeof SuppressedError === "function" ? SuppressedError : function(e, s, m, _) {
    return _ = Error(m), _.name = "SuppressedError", _.error = e, _.suppressed = s, _;
  }, fail = (e) => error = hasError ? new E(e, error, "An error was suppressed during disposal") : (hasError = true, e), next = (it) => {
    while (it = stack.pop()) {
      try {
        var result = it[1] && it[1].call(it[2]);
        if (it[0])
          return Promise.resolve(result).then(next, (e) => (fail(e), next()));
      } catch (e) {
        fail(e);
      }
    }
    if (hasError)
      throw error;
  };
  return next();
};

// /tmp/b.mjs
var __stack = [];
try {
  let run = function() {
    console.log(env);
  };
  const env = {};
  var server = __using(__stack, Bun.serve({
    port: 0,
    fetch(req) {
    }
  }), 0);
  for (let k of run()) {
  }
} catch (_catch) {
  var _err = _catch, _hasErr = 1;
} finally {
  __callDispose(__stack, _err, _hasErr);
}
```

We are hoisting generator and async generator functions outside of the scope of the using declaration we wrap it in, causing a ReferenceError

snoglobe:
### What does this PR do?

fixes #12407 and the regression demonstrated in #11083 where a `let` statement before a `using` would not be converted into a `var` so anything referencing it outside of the generated try/catch would raise a `ReferenceError`

- [x] Code changes


robobun:
@snoglobe, your commit 4077e3c9a351f46388abe640b0a31a19fb2032f6 is building: [`#4092`](https://buildkite.com/bun/bun/builds/4092)

Jarred-Sumner (changes requested):
Several test failures

src/js_parser.zig
```diff
@@ -22197,29 +22198,37 @@ fn NewParser_(
...
                                     var end: usize = 0;
+                                    var any_decl_in_const_values = local.kind == .k_const;
                                     for (decls) |decl| {
                                         if (decl.binding.data == .b_identifier) {
-                                            const symbol = p.symbols.items[decl.binding.data.b_identifier.ref.innerIndex()];
-                                            if (p.const_values.contains(decl.binding.data.b_identifier.ref) and symbol.use_count_estimate == 0) {
-                                                continue;
+                                            if (p.const_values.contains(decl.binding.data.b_identifier.ref)) {
+                                                any_decl_in_const_values = true;
+                                                const symbol = p.symbols.items[decl.binding.data.b_identifier.ref.innerIndex()];
```
Jarred-Sumner, line 22209:
```suggestion
                                                const symbol = &p.symbols.items[decl.binding.data.b_identifier.ref.innerIndex()];
```

src/js_parser.zig
```diff
@@ -22197,29 +22198,37 @@ fn NewParser_(
...
+                                                }
                                             }
                                         }
                                         decls[end] = decl;
                                         end += 1;
                                     }
                                     local.decls.len = @as(u32, @truncate(end));
-                                    if (end == 0) {
-                                        stmt.* = stmt.*.toEmpty();
+                                    if (any_decl_in_const_values) {
```
Jarred-Sumner: should this if be there? i'm confused

snoglobe: it's the only side effect that happens if it being const/in const_values is true. so we don't want it to happen if the check is false

snoglobe: (the locals.decls.len assignment is a no-op if it's false because end will be the len value. it should probably be moved into the if statement too though then)

Jarred-Sumner (changes requested):
merge conflict

snoglobe:
fixed

PR #14261, opened 2024-10-01
fix many files missing ending newlines

nektro:


robobun:
@nektro, your commit f39e61b52bc3a8565f7682bb45feec298034acc7 is building: [`#4110`](https://buildkite.com/bun/bun/builds/4110)

Jarred-Sumner:
Can we leave this to formatters instead of humans? In this case, clang-format shoudl be doing this .

PR #14218, opened 2024-09-27
allow importing of files with top level awaits in bundler

snoglobe:
### What does this PR do?

fixes #12508 - we now implement validateTLA from esbuild and properly flag `is_async_or_has_async_dependency`. 

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

PR #14025, opened 2024-09-19
kit: implement server components dev server

paperclover:
bun

PR #14262, opened 2024-10-01
ci: fix build error

nektro:
https://github.com/oven-sh/bun/pull/14218 was not rebased on https://github.com/oven-sh/bun/pull/14025 causing the building of the zig code to fail once both had reached `main`


robobun:
@nektro, your commit a32458e7510a6a1aabb1e23aa43a381f7bd6ec8d is building: [`#4113`](https://buildkite.com/bun/bun/builds/4113)

PR #14263, opened 2024-10-01
Experiment: statically override malloc on Windows & Linux

Jarred-Sumner:
### What does this PR do?

On Linux, we are importing the libc free and libc malloc symbols

What if we don't do that


robobun:
@Jarred-Sumner, your commit cee37c7d7d40a547719fca6c8a325ef0c6abeb55 is building: [`#4114`](https://buildkite.com/bun/bun/builds/4114)

Jarred-Sumner:
In microbenchmarks, this increased memory usage across the board

Issue #1130, opened 2022-08-22
Bun Install prints to stderr


DiefBell:
### Version

- 0.1.10
- 0.1.13

### Platform

Microsoft Windows NT 10.0.19044.0 x64

### What steps will reproduce the bug?

`bun install` prints to `stderr` i.e. `bun install 1>/dev/null` still prints `bun install v0.1.10`.

### How often does it reproduce? Is there a required condition?

Always

### What is the expected behaviour?

Print to stout

### What do you see instead?

Prints to stderr

PR #14266, opened 2024-10-01
pm: print command name to stdout

nektro:
Closes https://github.com/oven-sh/bun/issues/1130


robobun:
@nektro, your commit 7e45e71460bd665f20f7f719ece1041a181a5252 is building: [`#4130`](https://buildkite.com/bun/bun/builds/4130)

test/cli/install/bun-workspaces.test.ts
```diff
@@ -212,12 +224,13 @@ test("adding workspace in workspace edits package.json with correct version (wor
     cmd: [bunExe(), "add", "pkg2@workspace:*"],
     cwd: join(packageDir, "packages", "pkg1"),
     stdout: "pipe",
-    stderr: "inherit",
```
Jarred-Sumner: Why change this? it will only make it harder to debug test failures.

dylan-conway:
While we're at it, we should probably change these ones too
Screenshot 2024-10-01 at 1 57 16 PM

PR #4131, opened 2023-08-12
Deprecate loading `node_modules.bun`

Jarred-Sumner:
### What does this PR do?

Bun stopped generating `node_modules.bun` as of Bun v0.6. We continued to have code that loads `node_modules.bun`, but there are no tests. This deprecates `node_modules.bun` entirely, deleting the code that makes use of it.

PR #14267, opened 2024-10-01
clang-format: add InsertNewlineAtEOF

nektro:
proper fix over https://github.com/oven-sh/bun/pull/14261

adds new option to .clang-format
also makes ci run it on all files
also makes ci use the version we install, was previously using clang-format 14


robobun:
@nektro, your commit ffc7b5d21f9112028592772ce345893280358a03 is building: [`#4131`](https://buildkite.com/bun/bun/builds/4131)

PR #4162, opened 2023-08-15
chore: fix method name typo

Hanaasagi:
### What does this PR do?

The latest main branch cannot be built currently. Commit sha 1a6a52314f88946e66ac0f36d00d624ec1a5296f. Strangely, the CI status is still green 🤔. There is no `latin1` method in `WTFStringImplStruct`, only a `latin1Slice` method.

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

PR #14277, opened 2024-10-01
fix 14248

### LINKED ISSUE #14248: Test expression has bad behavior for explicit code units

1.1.29+6d43b3662

Darwin 24.1.0 arm64 arm

Execute the following code using the `bun` command:

```js
if (+'\u0030\u002e\u0031') console.log('a'); // called (only node?)
if (+'\x30\x2e\x31') console.log('b');       // called (only node?)
if (+'0.1') console.log('c');                // called (on node & bun)

console.log('\u0030\u002e\u0031' === '0.1'); // true! (on node & bun)
console.log('\x30\x2e\x31' === '0.1');       // true! (on node & bun)

console.log(!!+'\u0030\u002e\u0031');        // should print true, doesn't on bun
console.log(!!+'\x30\x2e\x31');              // should print true, doesn't on bun
console.log(!!+'0.1');                       // true on node & bun
```

The Bun repl does NOT have this bug, it is only running a script using the `bun` command.

Bun should recognize the string literals coerced by unary plus as truthy values, matching the behavior of the repl and node.

### What do you see instead?

See reproduction steps.

### Additional information

_No response_

paperclover:
Fixes #14248

Escaped UTF-16 data was being interpreted as unescaped UTF-8

This bug reveals that `E.String` is too unsafe, specifically `data`. I have left a TODO comment for when I reduce `E.String`'s memory footprint to also make it harder to mis-interpret the underlying bytes.

### DIFF
--- a/src/js_ast.zig
+++ b/src/js_ast.zig
@@ -2251,6 +2251,7 @@ pub const E = struct {
     pub const String = struct {
         // A version of this where `utf8` and `value` are stored in a packed union, with len as a single u32 was attempted.
         // It did not improve benchmarks. Neither did converting this from a heap-allocated type to a stack-allocated type.
+        // TODO: change this to *const anyopaque and change all uses to either .slice8() or .slice16()
         data: []const u8 = "",
         prefer_template: bool = false,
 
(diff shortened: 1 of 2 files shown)


robobun:
@paperdave, your commit 6013dbc1c7850c999fe03355239d8339f79ae641 is building: [`#4162`](https://buildkite.com/bun/bun/builds/4162)

PR #4179, opened 2023-08-16
tty `ReadStream`, `WriteStream`, and readline rawmode

dylan-conway:
### What does this PR do?
Adds `WriteStream` and `ReadStream` to tty module. Also adds rawmode to readline enabling libraries like inquirer and enquirer.

closes #4153 , closes #4061 , closes #2081 , closes #1787 

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

### How did you verify your code works?
Wrote tests for tty streams and process.stdout/stdin

PR #14280, opened 2024-10-01
cli:test: skip printing successes when in debug mode and BUN_DEBUG_jest=1

nektro:
this greatly reduces the noise when developing by reducing the amount of redundant information printed when developing bun itself

below is the result shown in the debugger output

image

and run manually in the terminal

image

been using this as a local patch for quite some time and thought to finally upstream

### DIFF
--- a/src/cli/test_command.zig
+++ b/src/cli/test_command.zig
@@ -175,9 +175,18 @@ pub const CommandLineReporter = struct {
 
         var this: *CommandLineReporter = @fieldParentPtr("callback", cb);
 
-        writeTestStatusLine(.pass, &writer);
-
-        printTestLine(label, elapsed_ns, parent, false, writer);
+        const debug_jest = blk: {
+            if (bun.getenvZ("BUN_DEBUG_jest")) |env| {
+                if (strings.eqlComptime(env, "1")) {
+                    break :blk true;
+                }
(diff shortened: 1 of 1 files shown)


robobun:
@nektro, your commit 823692c3535b2808150fb14fe7784438aef6f8c3 is building: [`#4179`](https://buildkite.com/bun/bun/builds/4179)

src/cli/test_command.zig
```diff
@@ -175,9 +175,18 @@ pub const CommandLineReporter = struct {
 
         var this: *CommandLineReporter = @fieldParentPtr("callback", cb);
 
-        writeTestStatusLine(.pass, &writer);
-
-        printTestLine(label, elapsed_ns, parent, false, writer);
+        const debug_jest = blk: {
```
paperclover: why not bun.Output.Scoped(.jest, false).isVisible()

(captial `S` to get the namespace instead of the log function)

nektro: didnt know about that function, good callout

Jarred-Sumner (changes requested):
I think it's a product issue if we have to modify the test reporter like this for our own internal debug usage.

Instead, we could add a --style=compact flag or similar that prints less information

PR #14177, opened 2024-09-26
node: implement more validators in native code

nektro:
reduces the overhead of these functions and the amount of duplicate js implementations across the builtin modules

pulled out of https://github.com/oven-sh/bun/pull/13757 to merge independently

PR #14285, opened 2024-10-02
BunProcess: dont use for-loop for isSignalName

nektro:
small followup to https://github.com/oven-sh/bun/pull/14177


robobun:
@nektro, your commit 25e6e69c423a4e311d8fad7bbda255e975609b9f is building: [`#4208`](https://buildkite.com/bun/bun/builds/4208)

PR #14287, opened 2024-10-02
add more logs to fs-watch-recursive-add-folder.test.js

nektro:
i also increased the timeout a bit in case the watcher wasnt ready yet by the time the directory was made. would be nice if there was an event for this


robobun:
@nektro, your commit 1997a306b28829796bdc98c117fa85c9b1b31275 is building: [`#4210`](https://buildkite.com/bun/bun/builds/4210)

PR #14288, opened 2024-10-02
windows: watcher: that line wasnt meant to go to stdout

nektro:
okay this didnt fix this issue, but its back to timing out instead of throwing this stdout matcher error


robobun:
@nektro, your commit 3d3000ed8e432f0d21a209743a754535df7aa5ee is building: [`#4212`](https://buildkite.com/bun/bun/builds/4212)

PR #14291, opened 2024-10-02
lazy load these requires so that node:cluster imports faster

nektro:
`bun-1.1.29`: >5ms
this-branch: <1ms


robobun:
@nektro, your commit 92090d7c26359c655c30c8658fce24b03afb9a0b is building: [`#4220`](https://buildkite.com/bun/bun/builds/4220)

Issue #10887, opened 2024-05-07
Segfault when decorating computed property


lucasvanhalst:
### How can we reproduce the crash?

Decorate any computed property and it triggers a segfault while transpiling the code.

### JavaScript/TypeScript code that reproduces the crash?

```shell
import {IsString} from 'class-validator';

export enum ExampleEnum {
    ONE = '1',
    TWO = '2',
}

export class Example {
    @IsString()
    [ExampleEnum.ONE]: string;
}
```

### Stack Trace (bun.report)

Bun v1.1.6 ([`e58d67b`](https://github.com/oven-sh/bun/tree/e58d67b46875280bc9002add4c3fae45472cb346)) on macos aarch64 [AutoCommand]

Segmentation fault at address 0x00000008

- *2 unknown/js code*
- `src.js_printer.NewPrinter`
- `src.js_printer.NewPrinter`
- `src.bundler.Bundler.printWithSourceMap__anon_301037`
- `src.bun.js.module_loader.ModuleLoader.transpileSourceCode__anon_300818`
- `Bun__transpileFile`
- `Bun::fetchESMSourceCodeAsync`
- `Zig::GlobalObject::moduleLoaderFetch`
- `JSC::moduleLoaderFetch`

PR #14304, opened 2024-10-02
fix segfault when decorating computed property #10887

matubu:
### What does this PR do?

Previously, `lowerClass` wasn't allowing all types of expressions to be decorated. If it wasn't allowed, it would reach an `unreachable` statement and cause a segfault. With this change, it should be possible to decorate any type of expression (eg, `class Example { @deco() [CONSTANT]: number; }`).

This fix: #10887 

- [x] Code changes


dylan-conway (approved):
Thank you

PR #14305, opened 2024-10-02
node:stream: fix setDefaultHighWaterMark

nektro:
localized function implementation was interfering with functions of the same name in global scope

global scoped functions exist at b/L2001-2026


robobun:
@nektro, your commit 89374b30e5126e924fe87419e6cd25a01acbee6e is building: [`#4249`](https://buildkite.com/bun/bun/builds/4249)

PR #14307, opened 2024-10-02
fix `getOwnTruthy`, fix #14299

### LINKED ISSUE #14299: SvelteKit Node adapter not working

latest from nightly Docker

Linux 6.1.94_1 x86_64 unknown

1. Create SvelteKit project
2. Install `@sveltejs/adapter-node`
3. Enable Node adapter in `svelte.config.js`
4. `bun run build`
5. `bun run ./build/`

Server starts (Node can run the same build output properly).

### What do you see instead?

```
$ bun run ./build/
176 | 				(req, _, next) => {
177 | 					if (typeof base === 'string') {
178 | 						let len = base.length;
179 | 						base.startsWith('/') || len++;
180 | 						req.url = req.url.substring(len) || '/';
181 | 						req.path = req.path.substring(len) || '/';
      ^
TypeError: Cannot specify both hostname and unix
 code: "ERR_INVALID_ARG_TYPE"

      at serve (native:1:1)
      at node:http:540:16
      at listen (node:http:497:9)
      at listen (/home/bun/js/mydeals/build/index.js:181:29)
      at module code (/home/bun/js/mydeals/build/index.js:270:9)

Bun v1.1.30-canary.71+af82a446d (Linux x64 baseline)
```

### Additional information

**Why not use SvelteKit with Bun adapter?**

There are multiple Bun adapters for SvelteKit, but those are not officially supported by the Svelte team, and have serious, deployment-blocking bugs.

**Why not run the server with Node instead?**

I'm using Bun-specific APIs.

matubu:
### What does this PR do?

`getOwnTruthy` only checked if the value was `undefined`. Now it checks if the value is truthy like `getTruthy`. Fix #14299

- [x] Code changes


GingerAdonis:
@Jarred-Sumner Sorry for tagging you. But it would be great to include this PR for v1.1.31 as this fixes a regression introduced in v1.1.21. Thank you

Issue #14098, opened 2024-09-22
Top-level `using` statement break `export` statement


codehz:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

```js
using a = {}
export default {}
```

### What is the expected behavior?

no syntax error (a runtime error is expected)

### What do you see instead?

```
2 | using a = {}
    ^
SyntaxError: Unexpected keyword 'export'
      at <parse>
```

### Additional information

the proposal [^1] say
> When a using declaration is parsed with BindingIdentifier Initializer, the bindings created in the declaration are tracked for disposal at the end of the containing Block or Module (a using declaration cannot be used at the top level of a Script)

so it should be able to use in the top level of a **Module**

[^1]: https://github.com/tc39/proposal-explicit-resource-management?tab=readme-ov-file#using-declarations-with-explicit-local-bindings

codehz:
As reference, this is how typescript handle top level using statement:

```js
var __addDisposableResource = (this && this.__addDisposableResource) || function (env, value, async) {
    if (value !== null && value !== void 0) {
        if (typeof value !== "object" && typeof value !== "function") throw new TypeError("Object expected.");
        var dispose, inner;
        if (async) {
            if (!Symbol.asyncDispose) throw new TypeError("Symbol.asyncDispose is not defined.");
            dispose = value[Symbol.asyncDispose];
        }
        if (dispose === void 0) {
            if (!Symbol.dispose) throw new TypeError("Symbol.dispose is not defined.");
            dispose = value[Symbol.dispose];
            if (async) inner = dispose;
        }
        if (typeof dispose !== "function") throw new TypeError("Object not disposable.");
        if (inner) dispose = function() { try { inner.call(this); } catch (e) { return Promise.reject(e); } };
        env.stack.push({ value: value, dispose: dispose, async: async });
    }
    else if (async) {
        env.stack.push({ async: true });
    }
    return value;
};
var __disposeResources = (this && this.__disposeResources) || (function (SuppressedError) {
    return function (env) {
        function fail(e) {
            env.error = env.hasError ? new SuppressedError(e, env.error, "An error was suppressed during disposal.") : e;
            env.hasError = true;
        }
        var r, s = 0;
        function next() {
            while (r = env.stack.pop()) {
                try {
                    if (!r.async && s === 1) return s = 0, env.stack.push(r), Promise.resolve().then(next);
[...]

PR #14313, opened 2024-10-03
hoisting of exports when there is top level using

### LINKED ISSUE #13734: `export` and `using` keywords cannot be used in the same file

1.1.26+0a37423ba

```ts
export class State{
    constructor() {
        console.info("Creating");
    }
    [Symbol.dispose](){
        console.info("Disposing");
    }
}

using state = new State();
```

It should work

### What do you see instead?

```
SyntaxError: Unexpected keyword 'export'
      at <parse> (/path/to/file.ts)
```

### Additional information

_No response_

snoglobe:
### What does this PR do?

fixes #13734 (and #14098 which actually was its own fix) - exported classes and export default were not hoisted properly and instead ended up inside the generated try/catch block.

- [ ] Code changes

### DIFF
--- a/src/ast/base.zig
+++ b/src/ast/base.zig
@@ -114,7 +114,7 @@ pub const Ref = packed struct(u64) {
         allocated_name,
         source_contents_slice,
         symbol,
-    } = .invalid,
+    },
 
     source_index: Int = 0,
 
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -9640,7 +9640,6 @@ fn NewParser_(
                             if (p.lexer.isContextualKeyword("async")) {
                                 const async_range = p.lexer.range();
                                 try p.lexer.next();
-                                var defaultName: js_ast.LocRef = undefined;
                                 if (p.lexer.token == T.t_function and !p.lexer.has_newline_before) {
                                     try p.lexer.next();
                                     var stmtOpts = ParseStatementOptions{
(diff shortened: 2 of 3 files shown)


robobun:
@snoglobe, your commit ace0f1224554840ffc40478d6b688ac4f099d1dd is building: [`#4281`](https://buildkite.com/bun/bun/builds/4281)

paperclover (changes requested):
what if we have more tests for `using` stuff? this is in part my fault for not adding it when i implemented the using lowering.

the one on my mind is doing `export default somethingThatThrows()` should call the dispose function but also throw when importing, so you'll need to import that module via `await import()` (in a file without the using)

then one doing like `export default { x: 123 }` and not throwing, it should verify that all this `var default_export; try ... ; export default default_export;` code ends up actually exporting the default value. (while still calling dispose)

i dont think we have this tested but while you are at it, maybe worth ensuring that stuff like throwing in a dispose function doesn't mask calling the next dispose function.

and then anything else you can think of that isn't tested. these last two will prevent future regressions as we'll catch them before the branch ships.

as for the actual code, great work solving the bug. if you want to clarify what `continue;` in the `.s_export_default` block with a comment saying it has already been hoisted to a variable, that might help future readers.

paperclover (approved):
i apologize for my terrible implementation of `using`

paperdave:
great

PR #14314, opened 2024-10-03
ci: simplify build target labels so the name doesnt get cutoff

nektro:
before

image

after

image

### DIFF
--- a/.buildkite/ci.yml
+++ b/.buildkite/ci.yml
@@ -13,7 +13,7 @@ steps:
     group: ":darwin: aarch64"
     steps:
       - key: "darwin-aarch64-build-deps"
-        label: ":darwin: aarch64 - build-deps"
+        label: "build-deps"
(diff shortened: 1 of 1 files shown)


robobun:
@nektro, your commit 56f7665eefa8723128d5576cdf432102dbedd407 is building: [`#4282`](https://buildkite.com/bun/bun/builds/4282)

Electroid:
The problem with removing those labels, is it makes the annotations ambigious for baseline/non-baseline

Screenshot 2024-10-03 at 8 46 22 AM

nektro:
oh good eye, yeah thats not ideal
that seems like windows is using the wrong label then, because the macos failures look correct:

image

nektro:
oh, no, i see the issue

nektro:
fixed it

Electroid (approved):

Issue #4326, opened 2023-08-25
Bun --hot reload port clash, port already in use


nikhil-swamix:
### What version of Bun is running?

0.7.4+9cf868fef4ed5ed7ebcd2e6e561a1e83b35aaa43

### What platform is your computer?

Linux 6.1.34-58.102.amzn2023.aarch64 aarch64 aarch64

### What steps can reproduce the bug?

this is server code:

```
import cors from "cors";
import dotenv from "dotenv";
import express, { json, urlencoded } from "express";
import { connect } from "mongoose";
import auth from "./middlewares/auth.js";
import { router } from "./routes/index.js";



// INIT APP
const app = express();
dotenv.config();
app.use(cors());
app.use(json());
app.use(urlencoded({ extended: true }));

app.get("/", (req, res) => {
	res.send("Hello World- main branch - 04");
});

app.get("/env", (req, res) => {
	res.json(process.env);
});
app.use(auth);
app.use("/api", router);
// add route to log env variables


// app listen after mongoose connect
connect(Bun.env.MONGO_URI, { useNewUrlParser: true, useUnifiedTopology: true }).then(() => {
	console.log("Database.Connect: Success");
	app.listen(3000);
});

export default app;
``` 

### What is the expected behavior?

should hot reload like vite, webpack (frontend) but expecting for backend
i have read the docs, --watch is working but --hot or watch+hot is breaking
was expecting if i change small string like "apple" to "ball" 
and that route is available at "/"
then refresh browser should show ball. and server reload must not trigger.

### What do you see instead?

330 |       family: isIPv6(address) ? "IPv6" : "IPv4",
331 |       port: this.#server.port
332 |     };
333 |   }
334 | 
335 |   listen(port, host, backlog, onListen) {
                   ^
error: Failed to start server. Is port 3000 in use?
      at listen (node:http:335:16)
      at listen (/home/ec2-user/superrtruck.com/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/application.js:635:9)
      at /home/ec2-user/superrtruck.com/app.js:32:1

henpa:
I can also reproduce this error with the example from https://bun.sh/guides/http/hot

After running "bun --hot run index.ts" the app loads fine, but crashes with "error: Failed to start server. Is port 3000 in use?" as soon as I edit and save the index.ts file

I've just installed bun 0.8.1 and I'm running this on Debian 10.13

[...]

PR #14322, opened 2024-10-03
Make .get() ignore Object.prototype instead of using getOwn

Jarred-Sumner:
### What does this PR do?

This makes it so we never pay attention to `Object.prototype` when calling getters from Zig. This is a less restrictive mitigation for prototype pollution attacks compared to using `getOwn`.

The existing test for glob prototype pollution covers this


robobun:
@Jarred-Sumner, your commit a33f95ed54fb48927ff83cf9d925aded53e88dec is building: [`#4326`](https://buildkite.com/bun/bun/builds/4326)

PR #14340, opened 2024-10-04
Fix FFI bugs and let the test run in CI

190n:
### What does this PR do?

- [ ] Make `ffi.test.js` compile itself so it can run in CI (regressed now that the tests are being rewritten in Zig)
- [x] Make the tests more stringent
- [x] Fix the return values for our implementations of `memcpy` and `memset` that ffi/CC code call
- [x] Add a test for generating a compile error with cc
- [x] Add a `__builtin_debugtrap` function (matches Clang's name) to trigger a breakpoint from `FFI.h`, which helps debugging the conversion functions
- [ ] Fix remaining FFI bugs exposed by the tests
- [ ] Make cc test more stringent (necessary?)
- [ ] Ensure tests can run in CI on every platform (rewrite `ffi-test.c` in Zig) 
- [ ] Make createBuffer/createArrayBuffer actually tested (re-enable something like [this test](https://github.com/oven-sh/bun/blob/3dca3baaec4babf1521f57d46feff6d7c69229aa/test/bun.js/ffi.test.js#L456-L472))
- [ ] Make threadsafe functions actually tested (invoke them from multiple threads in the FFI module, should not crash or deadlock)


robobun:
@190n, your commit ae514ce0b569d8fd528d6661c6ae2816a906998f is building: [`#4357`](https://buildkite.com/bun/bun/builds/4357)

PR #14341, opened 2024-10-04
test: add missing vm.runInContext stubs

nektro:
these being supported block bun from running the test suite of https://github.com/facebook/lexical
(its jest with jsdom)


robobun:
@nektro, your commit c4bd7bdadd0a50efd604c1a8b08e6004b29ff738 is building: [`#4358`](https://buildkite.com/bun/bun/builds/4358)

Issue #4365, opened 2023-08-27
implement `expect.toHaveBeenCalledWith()`


simylein:
### What is the problem this feature would solve?

Instead of writing tests like

```ts
import { describe, expect, spyOn, test } from 'bun:test';

describe('i mock your mocks', () => {
	test('mock mock who is there', () => {
		const log = spyOn(console, 'log');
		log('hello world');
		expect(log.mock.calls[0]).toEqual(['hello world']);
	});
});
```

### What is the feature you are proposing to solve the problem?

We could do this

```ts
import { describe, expect, spyOn, test } from 'bun:test';

describe('i mock your mocks', () => {
	test('mock mock who is there', () => {
		const log = spyOn(console, 'log');
		log('hello world');
		expect(log).toHaveBeenCalledWith('hello world');
	});
});
```

### What alternatives have you considered?

Continuing to use `expect(log.mock.calls[0]).toEqual(['hello world']);`

ImBIOS:
@testing-library/react renderHook  combined with bun:test expect need toHaveBeenCalledWith, as example:

```ts
import { renderHook } from '@testing-library/react'
import { describe, expect, it } from 'bun:test'
import { usePathname, useRouter, useSearchParams } from 'next/navigation'

import { useUrlHelper } from '@/hooks/urlHelper'

jest.mock('next/navigation', () => ({
  // useRouter: jest.fn(),
  usePathname: jest.fn(),
  useSearchParams: jest.fn(),
}))

describe('useUrlHelper', () => {
  beforeEach(() => {
    ;(useRouter as jest.Mock).mockReturnValue({
      push: jest.fn(),
      refresh: jest.fn(),
    })
    ;(usePathname as jest.Mock).mockReturnValue('/test-path')
    ;(useSearchParams as jest.Mock).mockReturnValue(new URLSearchParams())
  })
  
// ... other tests

it('should remove specified query parameters', () => {
    ;(useSearchParams as jest.Mock).mockReturnValue(
      new URLSearchParams('foo=bar&baz=qux'),
    )

    const { result } = renderHook(() => useUrlHelper())

    result.current.appendQuery({ fizz: 'buzz' }, ['baz'])

    // Line bellow will be error, even though it's required and I don't have any alternative for this
    expect(result.current.router.push).toHaveBeenCalledWith(
      '/test-path?foo=bar&fizz=buzz',
    )
  })
```

[...]

PR #14343, opened 2024-10-04
fix jsx symbol collisions when importing own variables with same names

### LINKED ISSUE #14021: Bun 1.1.28 -> ReferenceError: Can't find variable: Fragment

1.1.28+cf4e9cb69

Darwin 23.6.0 arm64 arm

In the new version of Bun, it doesn't detect the Fragment variable even if I have it in an import directly from JSX-runtime.

That it uses correctly the direct fragment of the import. It crashes in the test and also in build who uses this component that has this import, for something the build is not done well.

```tsx
import { Fragment } from '@/jsx-runtime';

// This now shows an error:
return <Fragment>foo</Fragment>
```

### What do you see instead?

I see this error:

```sh
 97 |         >
 98 |           {content}
 99 |           {style.length > 0 && <style>{toInline(style)}</style>}
100 |         </template>
101 |       )}
102 |       <Fragment slot="">{props.children}</Fragment>
                                                ^
ReferenceError: Can't find variable: Fragment
```

### Additional information

https://github.com/user-attachments/assets/e4ad8643-bfbf-4f94-b414-166ffa3bbd4f

In this video is better to see the error.

snoglobe:
### What does this PR do?

fixes #14021 - importing and using something that shares a name with a JSX symbol caused an import collision and a strange reference error. underlying cause was the symbol collision was not properly handled (plus some legacy code that needed to be removed). 

- [ ] Code changes


robobun:
@snoglobe, your commit 8a8f7c1f86ea7ca2ccc4e7a81970b06b81514980 is building: [`#4365`](https://buildkite.com/bun/bun/builds/4365)

Jarred-Sumner (changes requested):
test failures

snoglobe:
fixed

src/js_parser.zig
```diff
@@ -16374,7 +16269,8 @@ fn NewParser_(
                                         break :tagger p.jsxStringsToMemberExpression(expr.loc, p.options.jsx.fragment) catch unreachable;
                                     }
 
-                                    break :tagger p.jsxImport(.Fragment, expr.loc);
+                                    const I_FUCKING_HATE_LLDB = p.jsxImport(.Fragment, expr.loc);
```
paperclover: inline constant into single usage. tbh the label can alos just be named tag to match

paperclover (changes requested):

paperdave:
reasonable otherwise.

src/js_parser.zig
Jarred-Sumner: Please don't name variables like this in Bun's codebase.

snoglobe:
forgot to remove it 😓

Jarred-Sumner:
merge conflicts again

snoglobe:
fixed

paperclover (approved):

Jarred-Sumner (changes requested):
bundler npm test is failing.
image

fel1x-developer:
Any updates? This bug is blocking some projects to upgrading bun from <1.1.28 , such as honojs.

Jarred-Sumner:
This also needs to update the runtime transpiler cache version before it can be merged.

snoglobe:
fixed

PR #14351, opened 2024-10-04
ci: start running some of the citgm projects

nektro:
more to go but these are the ones that are already fully passing

should i make a status tracking issue for the rest upon merge?

### DIFF
--- a/.buildkite/ci.yml
+++ b/.buildkite/ci.yml
@@ -62,12 +62,6 @@ steps:
           - exit_status: 2
         retry:
           automatic:
-            - exit_status: 1
-              limit: 1
-            - exit_status: -1
-              limit: 3
-            - exit_status: 255
-              limit: 3
             - signal_reason: agent_stop
               limit: 3
             - signal: SIGTERM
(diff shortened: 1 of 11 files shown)


robobun:
@nektro, your commit d6a924db92e7b6b6a48e1a79ab1a47f022cf6fbc is building: [`#4377`](https://buildkite.com/bun/bun/builds/4377)

Electroid (changes requested):
It's going to be better if this does not use another runner and is part of the existing test suite. A lot of stuff will not work, such as the error annotations or log previews.

nektro:
> A lot of stuff will not work, such as the error annotations or log previews.

that's fine because the logs are short enough on their own and we still very easily see the binary pass/fail

nektro:
the main benefit this initiative will provide is when we run into a situation where one of the projects have a test failure and we don't. then we can investigate and if there's a real bug there we then fix it and add it to our own suite

nektro:
talked to @Electroid gonna rework this

Issue #10925, opened 2024-05-08
Add SQLite cursor API


broccolihighkicks:
### What is the problem this feature would solve?

Queries must load all of the data into RAM. But SQLite has a cursor API which would allow row at a time processing. Combined with read transactions this is useful for larger than RAM datasets.

### What is the feature you are proposing to solve the problem?

A JS API mapped to the C cursor API

### What alternatives have you considered?

Buy more RAM or delete some data.

SoundOfTheSky:
Duplicate of #8507

Issue #8507, opened 2024-01-26
SQLite Iterators


Mwni:
### What is the problem this feature would solve?

Iterate over large sets of data, without having the overhead of performing a `WHERE` for every retrieval step.

### What is the feature you are proposing to solve the problem?

An additional method called `iterate` added to the `Statement` class.
This has already been done on better-sqlite3 [(reference here)](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#iteratebindparameters---iterator).

Issue #2603, opened 2023-04-10
Make sqlite Statement iterable


tezlm:
### What is the problem this feature would solve?

The only way to get all rows from a statement (as far as I know) is with `.all()`. For large queries, this could use a lot of memory.

### What is the feature you are proposing to solve the problem?

Instead of loading every row at once, Statement could be iterable to stream rows from the database.

Jarred-Sumner:
I like this idea

PR #14361, opened 2024-10-04
Attempt to add support for iterate() with SQLite statements

Skywalker13:
### What does this PR do?

- [x] Code changes

It's an attempt to add support for iterate() with SQLite statements
(like better-sqlite3 https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#iteratebindparameters---iterator)

Related issues:
- Fixes https://github.com/oven-sh/bun/issues/2603
- Fixes https://github.com/oven-sh/bun/issues/8507
- Fixes https://github.com/oven-sh/bun/issues/10925

These changes are based on the `get` and `all` implementations. It's the first time that I see JavaScriptCore stuff. For this feature, I use javascript generators in the wrapper `bun:sqlite`. I suppose that it should be better to implement in JavaScriptCore but honestly I'm not comfortable enough with it.

I hope that is a good start because IMHO it's an essential feature when working with very large databases.

```
bun-debug test test/js/bun/sqlite/sqlite.test.js
```


src/js/bun/sqlite.ts
```diff
@@ -154,6 +157,12 @@ class Statement {
     return this.#raw.all();
   }
 
+  *#iterateNoArgs() {
```
Jarred-Sumner, line 160: Maybe worth also implementing `*[Symbol.iterator]()` version of this?

test/js/bun/sqlite/sqlite.test.js
```diff
@@ -587,6 +587,24 @@ it("db.query()", () => {
...
+  }
+  expect(i).toBe(2);
+
+  // iterate (args)
+  i = 0;
+  for (const row of db.query("SELECT * FROM test WHERE name = $name").iterate({ $name: "World" })) {
+    i === 0 && expect(JSON.stringify(row)).toBe(JSON.stringify({ id: 2, name: "World" }));
+    i++;
+  }
+  expect(i).toBe(1);
```
Jarred-Sumner, line 615: Can you add a test that checks what happens if you call iterate, stop partly through, and then call .all()?

Jarred-Sumner:
Thank you for this! Sorry it took a bit of time to merge.

Mwni:
Very nice

PR #14362, opened 2024-10-04
fix #14338: WebSocket doesn't emit error before close

### LINKED ISSUE #14338: WebSocket doesn't emit `error` before `close` on handshake error

1.1.30-canary.122+adc86c773

Darwin 24.0.0 arm64 arm

run the following code under Bun:

```js
const ws = new WebSocket('ws://venus.web.telegram.org/apiws_test')

ws.onopen = () => {
    console.log('open')
}

ws.onerror = (err) => {
    console.log('error', err)
}

ws.onclose = () => {
    console.log('close')
}
```

`error` event is emitted first, followed by a `close` event (i.e. `error`, then `close` lines are printed)

### What do you see instead?

only `close` event is emitted (i.e. only `close` line is printed)

### Additional information

Telegram backend responds with a `302 Found` if the handshake request does not contain a `Sec-WebSocket-Protocol: binary` header (which i specifically omitted in the repro). this can also be reproduced with non-websocket urls (e.g. `ws://example.com`)

in the browser (tested with firefox and chrome) as well as other runtimes (deno and `npm:ws` under node) this results in an `error` event followed by `close`, which is not the case under bun:

Image

matubu:
### What does this PR do?

This should fix: #14338

- [x] Code changes

### DIFF
--- a/src/bun.js/bindings/webcore/WebSocket.cpp
+++ b/src/bun.js/bindings/webcore/WebSocket.cpp
@@ -1144,7 +1144,7 @@ void WebSocket::didReceiveBinaryData(const AtomString& eventName, const std::spa
     // });
 }
 
-void WebSocket::didReceiveClose(CleanStatus wasClean, unsigned short code, WTF::String reason, bool isConnectionError)
+void WebSocket::didReceiveClose(CleanStatus wasClean, unsigned short code, WTF::String reason, bool wasError)
 {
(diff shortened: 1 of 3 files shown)


Electroid:
Thanks for working on this, to accept this PR we'll also need to add tests for this. You can look in the `test/web/websocket` folder and add tests there.

matubu:
I've added a testcase covering the bug fix in the `test/web/websocket` folder - let me know if you need any changes or additional tests.

Issue #14324, opened 2024-10-03
bake dev server tracking issue


paperclover:
With #14258 merged, Bake can finally be tried out in canary builds (usage guide at end of issue). However, it is missing many features and testing that make it currently unsuitible for production-grade applications (See #14763).

Remember this list is not in order, and that not every item on this list is strictly a blocker. I will edit it as the plan changes.

Contributions are appreciated. Please join [#bake in the Bun Discord](http://bun.sh/discord) for questions relating to the architecture of the Dev Server, as well as if you are trying it out as a user or framework author.

- Features
    - [x] CSS bundling. ~~This should have a custom path that does not invoke the bundler.~~ Waiting on #14281 
        - [x] Hot reloading
        - [x] CSR swapping styles correctly
    - [ ] Server Actions/Server Functions with `"use server"`.
        - (TODO: write out the steps for this)
    - [x] Source maps
    - [ ] Expose bundler options. Not all options will be exposed (for example, disabling `minifySyntax` is not allowed as affects DCE in an observable way)
    - [X] `Request` input (currently, the `Request` object given to the framework entry point is fake)
        - [x] Body streaming in 
    - [x] Response streaming (Response object is converted into a string)
    - [x] Support the existing Plugin API
    - [x] Bug: Mark route root as a SCB
    - [ ] Client side routing. Not as a complete built-in, but there needs to be a way for client-side framework code to request, bundle, and load routes.
    - [x] Prerendering API
        - [x] It should be possible for a framework to report that a route is always a static response.
            - [x] Should support different content-types for the `Accept` header. Concrete use case: picking between serialized RSC payload or HTML.
        - [ ] It also should be possible to report that a route is partially pre-rendered, to store a serialized blob.
        - [X] ~~While the above two are not really dev-server features, it should be possible to enable them in DevServer mode by a flag.~~
    - [x] Top-level await
    - [X] Macros
    - [x] `separateSSRGraph: false`
    - [ ] Framework should be able to specify a set of "built-in routes". For example, one may want to register `POST /render` to render a non-route server component.
        - [x] Routes with custom entry points (right now all go through `framework.serverEntryPoint`)
    - [ ] Server-side Debugger*
        - [ ] In VSCode extension
    - (more)
- API Design
    - [x] Move from `Bun.wipDevServerExpectHugeBreakingChanges` to an integration in `Bun.serve`
    - [x] Syntax of Route patterns (currently it is passthru to UWS)
[...]

Issue #4429, opened 2023-08-31
compatibility with node: console.dir


janwilmake:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Darwin 21.6.0 arm64 arm

### What steps can reproduce the bug?

I think this difference should be documented somewhere as it's not the same as Node.js

This is what I encountered:

`console.dir([/*long array*/], { maxArrayLength: 999 })`

it doesn't care about the options it seems, it just prints them.

In node.js this was typesafe and did wat you asked it to. In bun, it seems to adopt something else and just prints the last argument

This is what I was expecting:

https://techozu.com/show-all-items-console-log-node/

### What is the expected behavior?

It parses the options and doesn't print them

### What do you see instead?

It prints the options as log

zoodogood:
Bun v1.0.0
`console.dir(logs, {showHidden: true, depth: 0})` ignores the second argument as a options set and outputs it into the console as an object

bompus:
I came here with the same issue. For now, I am using this workaround:

```
import { inspect } from 'node:util';
const obj = { a: { b: 1 } };
console.log(inspect({ obj }, { colors: true, depth: 1 }));
```

Jarred-Sumner:
Fixed between Bun v1.0.0 and Bun v1.1.0

PR #14368, opened 2024-10-05
feat(bake): add dependencies to IncrementalGraph

paperclover:
Progress for #14324

Bake now properly tracks dependencies between files. When updating a file, the graph is traversed upwards for dependencies to the route files.

Updating a file now properly indicates which routes were updated (having a tab open on an unaffected route will not yield an update).

Updating a client component will, despite re-bundling SSR code, not invoke a server-side rerender. This is because the client code is sent over and used.

This does not handle an edge case with server files that reside in both the Server and SSR graphs. The first file's imports are not tracked because the second one overrides them. I will fix this later.

To make debugging easier, this also includes a visualizer for the incremental graph. I intend to keep this even in release builds so that if someone messes up their state, it can be explored or dumped.

image


robobun:
@paperdave, your commit 1aabb730c68f9f70e979693b0fe9b71288ca218b is building: [`#4429`](https://buildkite.com/bun/bun/builds/4429)

src/bun.zig
```diff
@@ -3899,3 +3918,13 @@ comptime {
     // Must be nominal
     assert(GenericIndex(u32, opaque {}) != GenericIndex(u32, opaque {}));
 }
+
+/// Reverse of the slice index operator.
+/// Given `&slice[index] == item`, returns the `index` needed.
+/// The item must be in the slice.
+pub fn indexOfPointerInSlice(comptime T: type, slice: []const T, item: *const T) usize {
```
Jarred-Sumner, line 3925: does not account for alignment.

src/js_ast.zig
```diff
@@ -8677,6 +8677,14 @@ pub const ServerComponentBoundary = struct {
                 bun.unsafeAssert(l.list.capacity > 0); // optimize MultiArrayList.Slice.items
                 return l.list.items(.reference_source_index)[i];
             }
+
+            pub fn bitSet(scbs: Slice, alloc: std.mem.Allocator, input_file_count: usize) !bun.bit_set.DynamicBitSetUnmanaged {
+                var scb_bitset = try bun.bit_set.DynamicBitSetUnmanaged.initEmpty(alloc, input_file_count);
```
Jarred-Sumner, line 8682: It would be better to use AutoBitSet here if possible. That will hold the memory inline if it fits.

src/bun.zig
paperclover: doesnt sizeOf return the size of the item in the slice?

if so, this function asserts the item is in the slice; regardless of how the slice is aligned, the item must be at some multiple of the size after the pointers are subtracted.

src/js_ast.zig
paperclover: i dont think autobitset is useful when you can just use stack-fallback allocator, which will hold the memory on the stack and avoid extra branches when reading into the bitset

PR #14371, opened 2024-10-05
more passing node buffer tests

nektro:
adds the passing buffer tests that were a culmination of some past prs and fixes some new ones too


robobun:
@nektro, your commit 7f29b0399c29fe0c426a89dc7b10638d80a759fb is building: [`#4450`](https://buildkite.com/bun/bun/builds/4450)

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -876,13 +888,75 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
...
+    auto throwScope = DECLARE_THROW_SCOPE(vm);
+
+    auto viewValue = callFrame->argument(0);
+    auto view = jsDynamicCast<JSArrayBufferView*>(viewValue);
+    if (!view) {
+        return Bun::ERR::INVALID_ARG_TYPE(throwScope, lexicalGlobalObject, "view"_s, "TypedArray"_s, viewValue);
+    }
+
+    auto ty = JSC::typedArrayType(view->type());
+    if (!JSC::isTypedView(ty)) {
```
Jarred-Sumner: `jsDynamicCast` checks this already, so this check is redundant.
```suggestion

```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -876,13 +888,75 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
...
+        return Bun::ERR::INVALID_ARG_TYPE(throwScope, lexicalGlobalObject, "view"_s, "TypedArray"_s, viewValue);
+    }
+
+    auto viewLength = view->length();
+    if (viewLength == 0) {
+        return JSValue::encode(createEmptyBuffer(lexicalGlobalObject));
+    }
+
+    auto offsetValue = callFrame->argument(1);
```
Jarred-Sumner: Copying from arguments should happen at the top of the function incase the registers move around

src/bun.js/bindings/JSBuffer.cpp
nektro: was guarding against `DataView` but on second thought that might work

nektro:
not seeing atm why test/js/node/test/parallel/buffer-tostring-range.test.js is inexplicably failing on all platforms but other than that this is ready for review. passes locally in debug, in release, when run individually, and when run `bun test buffer`

Jarred-Sumner:
```
error: Cannot create a string longer than 2147483647 characters
  | code: "ERR_STRING_TOO_LONG"
 ```

File: src/bun.js/bindings/JSBuffer.cpp [hunk 3]
```diff
@@ -250,25 +252,29 @@ const unsigned U32_MAX = std::numeric_limits<unsigned>().max();
 
 static inline uint32_t parseIndex(JSC::JSGlobalObject* lexicalGlobalObject, JSC::ThrowScope& scope, ASCIILiteral name, JSValue arg, size_t upperBound)
 {
-    if (!arg.isNumber()) return Bun::ERR::INVALID_ARG_TYPE(scope, lexicalGlobalObject, name, "number"_s, arg);
+    if (!arg.isNumber()) {
+        auto num = arg.toNumber(lexicalGlobalObject);
```
Jarred-Sumner: missing exception check.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -250,25 +252,29 @@ const unsigned U32_MAX = std::numeric_limits<unsigned>().max();
 
 static inline uint32_t parseIndex(JSC::JSGlobalObject* lexicalGlobalObject, JSC::ThrowScope& scope, ASCIILiteral name, JSValue arg, size_t upperBound)
 {
-    if (!arg.isNumber()) return Bun::ERR::INVALID_ARG_TYPE(scope, lexicalGlobalObject, name, "number"_s, arg);
+    if (!arg.isNumber()) {
```
Jarred-Sumner: can we check if it's an Int32 and avoid the cost of converting to a double for the most common case?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -876,13 +887,72 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
...
+    if (viewLength == 0) {
+        return JSValue::encode(createEmptyBuffer(lexicalGlobalObject));
+    }
+
+    double offset;
+    double length;
+
+    if (!offsetValue.isUndefined() || !lengthValue.isUndefined()) {
+        if (!offsetValue.isUndefined()) {
+            Bun::V::validateInteger(throwScope, lexicalGlobalObject, offsetValue, jsString(vm, String("offset"_s)), jsNumber(0), jsUndefined());
```
Jarred-Sumner, line 883: `V`? `Bun::validate::integer` is a bit clearer imo

Jarred-Sumner (changes requested):
These integer conversions should include a permalink to the relevant Node.js source code so we can be confident of the behavior it's trying to match

Almost all of these should be using the existing JSC methods to do these conversions instead of copy pasting slightly different implementations:
- `toLength` if you want `@max(number, 0)`
- `toIntegerWIthTruncation` if you want to preserve the sign

There are many subtleties here with how the Node implementation truncates to a signed 32 bit integer in some cases and then allows an int52 in other cases

src/bun.js/modules/NodeBufferModule.h
```diff
@@ -160,8 +176,15 @@ DEFINE_NATIVE_MODULE(NodeBuffer) {
...
       globalObject->JSDOMFileConstructor());
 
-  put(JSC::Identifier::fromString(vm, "INSPECT_MAX_BYTES"_s),
-      JSC::jsNumber(50));
+  {
+    PropertyDescriptor descriptor;
+    descriptor.setConfigurable(true);
+    descriptor.setEnumerable(true);
+    descriptor.setWritable(false);
+    descriptor.setGetter(JSFunction::create(vm, lexicalGlobalObject, 0, String("get"_s), jsFunction_INSPECT_MAX_BYTES_getter, {}, {}, {}, {}));
```
Jarred-Sumner: Please use custom getter/setters for these. JSFunctions are slower and much more code than they need to be

src/bun.js/bindings/ErrorCode.h
```diff
@@ -75,17 +75,17 @@ enum Bound {
 
 namespace ERR {
 
-JSC::EncodedJSValue INVALID_ARG_TYPE(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject, ASCIILiteral val_arg_name, ASCIILiteral val_expected_type, JSC::JSValue val_actual_value);
-JSC::EncodedJSValue INVALID_ARG_TYPE(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject, JSC::JSValue val_arg_name, ASCIILiteral val_expected_type, JSC::JSValue val_actual_value);
-JSC::EncodedJSValue OUT_OF_RANGE(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject, ASCIILiteral arg_name, size_t lower, size_t upper, JSC::JSValue actual);
+JSC::EncodedJSValue INVALID_ARG_TYPE(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject, WTF::String val_arg_name, WTF::String val_expected_type, JSC::JSValue val_actual_value);
```
Jarred-Sumner: Each usage of `WTF::String` this way makes an extra copy of the string. We can instead of use const `WTF::String&`

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -250,25 +252,38 @@ const unsigned U32_MAX = std::numeric_limits<unsigned>().max();
 
 static inline uint32_t parseIndex(JSC::JSGlobalObject* lexicalGlobalObject, JSC::ThrowScope& scope, ASCIILiteral name, JSValue arg, size_t upperBound)
 {
-    if (!arg.isNumber()) return Bun::ERR::INVALID_ARG_TYPE(scope, lexicalGlobalObject, name, "number"_s, arg);
+    if (arg.isUInt32()) {
```
Jarred-Sumner: Can we replace this function with `JSValue::toIntegerWIthTruncation`?  Or a similar JSC function to handle this? I worry about subtle integer conversion semantics in C++ leading to bugs in our user's code.

src/js/internal/util/inspect.js
```diff
@@ -2071,6 +2072,11 @@ function formatArray(ctx, value, recurseTimes) {
 }
 
 function formatTypedArray(value, length, ctx, ignored, recurseTimes) {
+  if (value instanceof Buffer) {
```
Jarred-Sumner:
```suggestion
  if (Buffer.isBuffer(value)) {
```

src/bun.js/modules/NodeBufferModule.h
nektro: would that match https://github.com/nodejs/node/blob/v22.9.0/lib/buffer.js#L1344-L1353 ?

Jarred-Sumner: yes

Issue #13820, opened 2024-09-09
Discrepancy in `https.request` implementation


etcd:
### What version of Bun is running?

1.1.27+267afa293

### What platform is your computer?

Darwin 23.6.0 arm64 arm

### What steps can reproduce the bug?

The following code reproduces the error:

```js
// test.js
import { request } from "https";

const req = request("https://google.com", (res) => {
  console.log(`Response URL: "${res.url}"`);
});

req.end();
```

Compare the outputs of the following commands:
```
bun run test.js
node test.js
```

### What is the expected behavior?

The output of `node test.js`:

```
Request URL: ""
```

### What do you see instead?

The output of `bun run test.js`:

```
Request URL: "/"
```

### Additional information

To add some context:

I discovered this bug when a Playwright script consistently crashed when run with Bun, but not with Node. This minimal Playwright script reproduces the crash (caused by the bug outlined above):

```js
import { chromium } from "playwright";

const testIntercept = async (url: string) => {
  const browser = await chromium.launch();
  const context = await browser.newContext();
  const page = await context.newPage();

  await page.route("**/*", async (route) => {
    const response = await route.fetch(); // crashes here due to bug
    await route.fulfill({ response });
  });

  await page.goto(url);
  await browser.close();
};

testIntercept("https://google.com");
```

etcd:
Just following up with some additional testing. The code below produces different behaviors when run with Bun vs Node.

`test.js`:
```js
import { request } from "https";

const makeRequest = (url) => {
  let responseUrl;

  request(url, (response) => {
    responseUrl = response.url;
    console.log("Response url inside request callback:", String(responseUrl));

    // consume data so this exits when run with node
    response.on("data", () => {});
  }).end(() => {
    console.log("Response url from end callback:", String(responseUrl));
  });
};

const urls = [
  "https://google.com",
  "https://bun.sh/blog",
  "https://en.wikipedia.org/wiki/1884",
];

urls.forEach((url) => makeRequest(url));
```

Discrepancy can be observed with the following commands:
```
bun run test.js
node test.js
```

Some observations:
* Inside the `.end(...)` callback, Bun and Node both print `undefined` for the response URL.
* Inside the `request(...)` callback, Bun prints `<path><query>`, while Node prints `""`.
* Unrelated issue, but Node requires `response.on("data", () => {})` to not hang. Bun does not require this.

I'm running this with Bun `1.1.43+76800b049` on `Darwin 24.1.0 arm64 arm`.

Issue #4462, opened 2023-09-03
Way to distinguish args for runner from args to script when running package.json scripts


rictic:
### What is the problem this feature would solve?

When running a package.json script, we'd like the ability to use `--` to unambituously distinguish arguments to the script runner from arguments to the script. This came up as part of evaluating bun support in [wireit](https://github.com/google/wireit].

Consider three invocations, where `--foo` is an option to the script runner, and `--bar` is an option to the `server` script. bun v0.8.1 populates `process.argv` like so:

```bash
bun run server --foo -- --bar
# process.argv is [bun, server, "--foo", "--", "--bar"]
```

```bash
bun run server --foo
# process.argv is [bun, server, "--foo"]
```

```bash
bun run server -- --bar
# process.argv is [bun, server, "--bar"]
```

The issue is that there's no way for the `server` npm run command to distinguish between the second and third cases.

### What is the feature you are proposing to solve the problem?

How other package.json script runners behave:

## npm

npm runs the `server` command with the `npm_config_foo` environment variable set when `--foo` is passed before a `--`, and with process.argv set to `[node, server, ...argsAfterStandaloneDoubleHyphen]`

## yarn v1.22.18

Sets the `npm_config_argv` to a JSON string with format:

```json
{
  "remain": [],
  "cooked": ["run", "server"],
  "original": ["run", "server", "--foo", "--", "--bar"],
}
```

The `original` field always has the exact arguments, sufficient to determine which were before the `--` and which after.

## yarn >v3.2.1 and pnpm >v7.1.7

Always passes all arguments to process.argv, including the `"--"` and those arguments before and after it.

### What alternatives have you considered?

In `wireit` we're able to tell that we're being run by bun, and we already have a switch statement to determine how we'll determine arguments. All we need is some way to reliably determine whether an argument was before or after the `--`.

I've looked at `process.env` and `process.argv` and don't see anything with this information. Without it, we'll have to guess, which we can't do reliably. For example, `wireit` takes a `--watch` parameter, but so do many of the scripts that we often run. So we can't tell the difference between `npm run server --watch` and `npm run server -- --watch`, which are both plausible invocations.

PR #14384, opened 2024-10-06
Rewrite node:http to not use Request & Response

Jarred-Sumner:
### What does this PR do?

Extremely WIP.

Rewrite node:http to not use Request & Response for Bun.serve

Introduce a new internal API that currently looks like this and may change at any point:
```js
import { serve } from "bun";

serve({
  port: 3002,

  onNodeHTTPRequest(server, url: string, method: string, headers: Record<string, string>, rawHeaders: string[], handle, hasBody: boolean) {
    handle.writeHead(200, { "Content-Type": "text/plain" });
    handle.end("Hello World");
  },
});
```

**This  _does not_  replace `fetch` in Bun.serve()**

it doesn't really work yet. We aren't emitting `socket`.


robobun:
@Jarred-Sumner, your commit a660692743ed7640bf32c27254be72ff8b417eaf is building: [`#4462`](https://buildkite.com/bun/bun/builds/4462)

kravetsone:
will it be available for the first time through the flag? it sounds dangerous if suddenly the insides change so much

Jarred-Sumner:
@kravetsone i agree dangerous, but we have far better test coverage than we used to.

kravetsone:
> @kravetsone i agree dangerous, but we have far better test coverage than we used to. 

It's great
But it maybe more painful for developers which don't pin bun docker image version 

Test coverage can't find all bugs

huseeiin:
super cool

huseeiin:
there was a trick in bun to get the Request object (from a symbol) in node:http:
```ts
import { createServer } from "http";

createServer(async (req, res) => {
    const request: Request = req[Object.getOwnPropertySymbols(req)[3]];

    const json = await request.json().catch(() => null);

    res.end(json);
}).listen(3000);
```
does this PR mean this is no longer gonna be possible?

etcd:
Is this likely to fix the following confirmed issue (in `https`)?
* https://github.com/oven-sh/bun/issues/13820

The issue is that the `response` object of a `request` does not always contain the right URL.

huseeiin:
hm i wonder how to use a framework with this, like express

mtsewrs:
As a library author this functional approach would be great to have imo since it has less overhead and also more compatible with node and i guess uws. Hope you reconsider ❤️

PR #14400, opened 2024-10-08
move .clang-format up a folder so it affects all our c/cpp files

nektro:
kept to src/ though so it doesnt affect the vendored forks in packages/


robobun:
@nektro, your commit d8ef33481b8312db96fdb33778faac39058b11f3 is building: [`#4513`](https://buildkite.com/bun/bun/builds/4513)

Issue #14397, opened 2024-10-08
Implement `bun run-script`


nektro:
### What is the problem this feature would solve?

`bun run <foo>` will give the filesystem higher precedence than the `package.json[scripts]` object when selecting what trying to run. so if there is a `<foo>` script but also a file and/or folder named `<foo>` as well, the file will be ran instead. this subcommand would allow the user to override this behavior and always prefer running a script, if available.

### What is the feature you are proposing to solve the problem?

https://docs.npmjs.com/cli/v10/commands/npm-run-script

### What alternatives have you considered?

`bun pm run <foo>` but upon realizing this npm command already exists, i feel this might be the better route

Issue #4553, opened 2023-09-07
Global configuration from a config file does not seem to apply


jmcantrell:
### What version of Bun is running?

0.8.1

### What platform is your computer?

Darwin 21.6.0 x86_64 i386
Linux 6.4.12-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

Create a config file with something like the following:

```toml
[install]
globalDir = "~/.bun-alt/install/global"
globalBinDir = "~/.bun-alt/bin"
```

Install a global package like: `bun --config=/path/to/config.toml add -g prettier`

### What is the expected behavior?

Bun should install the package to `~/.bun-alt`.

### What do you see instead?

Bun installs the package to the default directory, `~/.bun`.

stuartc:
I can confirm this is the case on 1.0.0 as well.

qzxst:
system-environment

qzxst:
windows11  
system environment  

BUN_INSTALL_CACHE_DIR
BUN_INSTALL_GLOBAL_BIN_DIR
BUN_INSTALL_GLOBAL_DIR

PR #14423, opened 2024-10-09
Implement `bun run-script`

nektro:
Closes https://github.com/oven-sh/bun/issues/14397


robobun:
@nektro, your commit 6525282cf245c807fd2cf0deba427a338878a412 is building: [`#4553`](https://buildkite.com/bun/bun/builds/4553)

nektro:
instead going to do some research to nail-down and fix inconsistencies in the `bun run` behavior instead

Issue #4597, opened 2023-09-08
Incompatible with lru-native2 dependency


beetlebum:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

- Attempt to run a large NodeJS project that contains the `lru-native2` dependency
- `bun src/index.ts`

### What is the expected behavior?

The application to run

### What do you see instead?

```
error: Cannot find module "./build/Release/native" from "/project/node_modules/lru-native2/index.js"
libc++abi: Pure virtual function called!
```

### Additional information

https://github.com/kibae/node-lru-native/blob/master/index.js
Not too familiar with Bun or NodeAPI but I believe this library is a simple C++ wrapper.

artberri:
The reason for that is that `bun` does not run lifecycle scripts by default and `lru-native2` makes use of them to compile some C++ code after installation.

You need to add  `lru-native2` to the [`trustedDependencies`](https://bun.sh/docs/cli/install#trusted-dependencies) on your `package.json` and the code will be compiled.

The problem in this case is that after doing it an error still happens:

```
TypeError: dlopen(/xxxxxxx/node_modules/lru-native2/build/Release/native.node, 0x0001): tried: '/xxxxxxx/node_modules/lru-native2/build/Release/native.node' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/xxxxxxx/node_modules/lru-native2/build/Release/native.node' (no such file), '/xxxxxxx/node_modules/lru-native2/build/Release/native.node' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
      at internalRequire (:1:20)
      at /xxxxxxx/node_modules/lru-native2/index.js:2:7
      at globalThis (/xxxxxxx/node_modules/lru-native2/index.js:2:51)
      at processTicksAndRejections (:1:2602)
```

### What version of Bun is running?

1.0.1


### What platform is your computer?

Darwin 22.6.0 arm64 arm

Electroid:
Duplicate of #4959

Jarred-Sumner:
This was not fixed in #4959 because `lru-native2` did not make the top 500 allowlist

PR #14452, opened 2024-10-10
ci: start running some of the citgm projects [v2]

nektro:
supersedes https://github.com/oven-sh/bun/pull/14351


robobun:
@nektro, your commit b040b454ac02645f303bb378f8706233b7e180c0 is building: [`#4597`](https://buildkite.com/bun/bun/builds/4597)

src/cli/pm_trusted_command.zig
```diff
@@ -169,16 +169,17 @@ pub const TrustCommand = struct {
...
-            Output.errGeneric("0 scripts ran. This means all dependencies are already trusted or none have scripts.", .{});
+            Output.warn("0 scripts ran. This means all dependencies are already trusted or none have scripts.", .{});
         } else {
-            Output.errGeneric("0 scripts ran. The following packages are already trusted, don't have scripts to run, or don't exist:\n\n", .{});
+            Output.warn("0 scripts ran. The following packages are already trusted, don't have scripts to run, or don't exist:\n\n", .{});
             for (packages_to_trust) |arg| {
                 Output.prettyError(" <d>-<r> {s}\n", .{arg});
             }
         }
+        Global.exit(0);
```
Jarred-Sumner: Whats the rationale? This is a breaking change

Jarred-Sumner:
Where are these run and how can the results be viewed? I don't see it in the log

src/cli/pm_trusted_command.zig
nektro: its exactly the same behavior as before, i just moved this call from the callers into this function

nektro:
image

src/cli/pm_trusted_command.zig
nektro: oh i see; it not being an error. it didnt seem like something that should be an error condition and i consulted with dylan

nektro: the runner script runs this command and having this be an error makes it indistinguishable from having scripts that failed/crashed vs there were none in the first place

nektro: https://github.com/search?q=%22bun+pm+trust+--all%22&type=code looks like there's not a ton of in-the-wild usage and there's even multiple counts of users doing `|| true` potentially to obtain this behavior on their own

nektro:
superseded

Issue #14467, opened 2024-10-10
Bun install ensure all files has at least read permission


AielloChan:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

Given this protobufjs npm package at version 6.11.4, download and extract it.

https://registry.npmjs.org/protobufjs/-/protobufjs-6.11.4.tgz

then just

```bash
> ls -la package/minimal.js
-rw-r-----  1 aiello  staff    96B Oct 26  1985 package/minimal.js
```

You'll find that, **minimal.js** has no permission for **Other**. This may cause some problems. When use yarn/npm to install this package, it will ensure at least read permission for all type of users.

### What is the expected behavior?

bun installed files should has at least read permission for every user.

### What do you see instead?

some file has no permission for **Other** type user.

PR #14468, opened 2024-10-10
Bun install files permission

AielloChan:
### What does this PR do?

Try fix https://github.com/oven-sh/bun/issues/14467

- [X] Code changes

I just test it on my local machine. I'll write a test for it.

### DIFF
--- a/src/install/extract_tarball.zig
+++ b/src/install/extract_tarball.zig
@@ -153,6 +153,26 @@ pub fn buildURLWithPrinter(
     }
 }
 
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
Thank you for highlighting that this is an issue - but we the fix is going to cause a meaningful performance regression by adding an extra loop over all the files. We can make this work without regressing on performance by setting the mode during extraction without adding an extra pass over the input. I'll defer to @dylan-conway on this for what the behavior should be to align with npm/pnpm/yarn

Issue #4650, opened 2023-09-09
Bun install does not install dependencies between packages in the same monorepo


nbbaier:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Darwin 22.5.0 arm64 armw

### What steps can reproduce the bug?

Following the guide for setting up a monorepo [here](https://bun.sh/guides/install/workspaces):

Project structure:
```
.
├── package.json
├── node_modules
└── packages
    ├── stuff-a
    │   └── package.json
    └── stuff-b
        └── package.json
```

Root `package.json`
```
{
  "name": "my-monorepo",
  "private": true,
  "workspaces": ["packages/*"]
}
```

Stuff-b `package.json`
```
{
  "name": "stuff-b",
  "dependencies": {
   "stuff-a": "*"
  }
}
```

### What is the expected behavior?

Running `bun install` from the project root to install dependencies for all workspaces.

### What do you see instead?

I get the following error:

```
bun install v1.0.0 (822a00c4)
  🔍 stuff-a [1/1] 
error: package "stuff-a" not found registry.npmjs.org/stuff-a 404
error: stuff-a@* failed to resolve
```

divmgl:
You need to provide a version for `stuff-a`. `*` won't work.

nbbaier:
Thanks, but even with an explicit version, it still fails.

Revised `packages/stuff-b/package.json`
```
{
  "name": "stuff-b",
  "dependencies": {
    "stuff-a": "1.0.0"
  }
}
```

`packages/stuff-a/package.json`

```
{
    "name": "stuff-a",
    "version": "1.0.0"
}
```
Error:
```
bun install v1.0.0 (822a00c4)
  🔍 stuff-a [1/1] 
error: package "stuff-a" not found registry.npmjs.org/stuff-a 404
error: stuff-a@1.0.0 failed to resolve
```

nil1511:
Faced a similar issue. explicitly specifying as workspace dependency fixes it
```{
  "name": "stuff-b",
  "dependencies": {
   "stuff-a": "workspace:*"
  }
}
```

ramiel:
This didn't fix the problem for me

noClaps:
I've found that you don't need to declare packages in a monorepo as dependencies of each other. They're automatically added to `node_modules`, so you can use them directly.

```diff
"dependencies": {
// other dependencies
-  "stuff-a": "*"
}
```

[...]

PR #14479, opened 2024-10-11
Fix napi_value<=>integer conversions and napi_create_empty_array

190n:
### What does this PR do?

Conversions:
- `napi_get_value_int32` and `napi_get_value_uint32` use the same conversion as JS bitwise ops (truncate then wrap)
- `napi_get_value_int64`:
    - if value is not a number, return error
    - if value is Infinity or NaN, return 0
    - truncate value and _clamp_ to `int64_t` range

`napi_create_empty_array`:
- create array with empty slots instead of filling with JS `undefined`
- change too-long lengths to 0 instead of crashing (Node does this)


robobun:
@190n, your commit 33b2501b691923c073d520d1b0fc31e608e71c49 is building: [`#4650`](https://buildkite.com/bun/bun/builds/4650)

src/napi/napi.zig
```diff
@@ -296,29 +296,17 @@ pub export fn napi_create_array(env: napi_env, result_: ?*napi_value) napi_statu
     result.set(env, JSValue.createEmptyArray(env, 0));
     return .ok;
 }
-const prefilled_undefined_args_array: [128]JSC.JSValue = brk: {
```
Jarred-Sumner, line 299: nice

Jarred-Sumner (approved):
Merge conflicts but otherwise LGTM once tests pass

PR #14484, opened 2024-10-11
Fix console.table for numeric keys

### LINKED ISSUE #14415: `console.table()` does not display values for numeric keys

1.1.30+7996d06b8

Darwin 23.6.0 x86_64 i386

having the object with numeric keys:

```ts
console.table({test: {"10": 123, "100": 154}});
```

should work with numeric keys like it does in Node.js 

```sh
┌─────────┬─────┬─────┐
│ (index) │ 10  │ 100 │
├─────────┼─────┼─────┤
│ test    │ 123 │ 154 │
└─────────┴─────┴─────┘
```

### What do you see instead?

v1.1.30 displays no values:

```sh
┌──────┬─────┬─────┐
│      │ 10  │ 100 │
├──────┼─────┼─────┤
│ test │     │     │
└──────┴─────┴─────┘
```

### Additional information

- used to work fine in 1.1.29
- adding space character is a workaround (it makes it work):

```ts
console.table({test: {" 10": 123, " 100": 154}});
```

```sh
┌──────┬─────┬──────┐
│      │  10 │  100 │
├──────┼─────┼──────┤
│ test │ 123 │ 154  │
└──────┴─────┴──────┘
```

pfgithub:
### What does this PR do?

Fixes #14415

Previously, `getWithString` was called for each property key, but getWithString calls getIfPropertyExistsPrototypePollutionMitigation which asserts the key is not an index key. Presumably, getWithString used to work here before the prototype pollution mitigation in 1.1.20.

With this change, properties that aren't an own property are no longer included in `console.table`

```js
class A {
    constructor(cond) {
        this.a = 1;
        this.b = 2;
    }
}
class B {
    constructor(cond) {
        this.a = 1;
    }
    get b() { return 128; }
}

console.table({a: new A(), b: new B()});
```

Previously, this logged the value of b for both classes, but the one with the getter had the table printed wrong:

```
┌───┬───┬───┐
│   │ a │ b │
├───┼───┼───┤
│ a │ 1 │ 2 │
│ b │ 1 │ 128 │
└───┴───┴───┘
```

Now, it does not log the b with the getter:

```
┌───┬───┬───┐
│   │ a │ b │
├───┼───┼───┤
│ a │ 1 │ 2 │
│ b │ 1 │   │
└───┴───┴───┘
```

This also removes the bindings for getWithString as this was the only place they were used.

- [x] Code changes

- [ ] N/A I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be

### DIFF
--- a/src/bun.js/ConsoleObject.zig
+++ b/src/bun.js/ConsoleObject.zig
@@ -342,7 +342,7 @@ const TablePrinter = struct {
             //  - otherwise: iterate the object properties, and create the columns on-demand
             if (!this.properties.isUndefined()) {
                 for (columns.items[1..]) |*column| {
-                    if (row_value.getWithString(this.globalObject, column.name)) |value| {
+                    if (row_value.getOwn(this.globalObject, column.name)) |value| {
                         column.width = @max(column.width, this.getWidthForValue(value));
                     }
                 }
@@ -436,7 +436,7 @@ const TablePrinter = struct {
                     value = row_value;
                 }
             } else if (row_value.isObject()) {
-                value = row_value.getWithString(this.globalObject, col.name) orelse JSValue.zero;
+                value = row_value.getOwn(this.globalObject, col.name) orelse JSValue.zero;
             }
 
             if (value.isEmpty()) {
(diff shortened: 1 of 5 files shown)


Jarred-Sumner (approved):
Thank you for this

Issue #6136, opened 2023-09-28
Implement `promiseHooks` from `node:v8`


bkniffler:
### What is the problem this feature would solve?

When trying to use temporal sdk (https://github.com/temporalio/sdk-typescript), getting error
```
node:v8 createHook is not yet implemented in Bun
```

Its used right here:
https://github.com/temporalio/sdk-typescript/blob/3ce098be042e86d26ea5b433508a480503cb09b2/packages/worker/src/workflow/vm-shared.ts#L198

### What is the feature you are proposing to solve the problem?

Implement createHook.

### What alternatives have you considered?

Ask temporal to not use createHook, but probably out of reach since its needed to isolate the VM scope.

Electroid:
We'd have to think about if this is worth implementing, since it could have a performance impact. (it tracks _all_ promises) However, even if it is feasible, it's not a high priority.

alirezabonab:
@Electroid  Is there any update on this?

danieldunderfelt:
Sadly the Temporal Typescript SDK does not work without createHook 😩

I'd love to have it at least as an opt-in if the performance impact is big.

Ping @lorensr

loicnestler:
What's the status on this? Is an implementation of `createHook` planned?

bergundy:
I'm the author of the code you linked in the Temporal TypeScript SDK.
Promise hooks are optional, they allow us to capture the stack trace of all running functions in a workflow.
I'm not sure if there's more of a gap for running our SDK on top of Bun but we can look into it.
Apart from promise hooks, we use Node's `vm` and `worker_threads` modules, `AsyncLocalStorage`, and rely heavily on Rust Neon bindings.

paradoxloop:
@bergundy if Promise Hooks are optional is there a way to disable them so that we can use Bun. There is a [linked issue](https://github.com/temporalio/sdk-typescript/issues/1334) on the Temporal project.

axe-me:
Can bun simply keep silent about this instead of throw an error, if there is no plan to implement this in short time. Just like what deno did in this PR: https://github.com/denoland/deno/pull/25979/files

bergundy:
#14485 seems like the way to go. If it ends up not getting merged, we may be able to do something on the Temporal SDK side, it should be easy to edit the SDK files post install to remove the use of promise hooks and confirm.
I’m subscribed to this issue, wondering if this is the last hurdle for running the Temporal SDK on bun.

cortopy:
[...]

PR #14485, opened 2024-10-11
remove promiseHooks stub;

axe-me:
project using this optionally could at least run, trying to resolve this issue: https://github.com/oven-sh/bun/issues/6136

### What does this PR do?

- [x] Code changes

### DIFF
--- a/src/js/node/v8.ts
+++ b/src/js/node/v8.ts
@@ -98,29 +98,12 @@ function writeHeapSnapshot() {
 function setHeapSnapshotNearHeapLimit() {
   notimpl("setHeapSnapshotNearHeapLimit");
 }
-const promiseHooks = {
-    createHook: () => {
-      notimpl("createHook");
-    },
-    onInit: () => {
-      notimpl("onInit");
-    },
-    onBefore: () => {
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you for this, but this would break libraries which test for the existence of promise hooks in other ways

Instead, we can create a function that masquerades as undefined and then assign the promise hook functions onto that function as well. Then, code which does `if (promiseHooks)` would return `false`, but if they were to do `promiseHooks.createHook`, it will warn

gusfune:
> Thank you for this, but this would break libraries which test for the existence of promise hooks in other ways
> 
> Instead, we can create a function that masquerades as undefined and then assign the promise hook functions onto that function as well. Then, code which does `if (promiseHooks)` would return `false`, but if they were to do `promiseHooks.createHook`, it will warn

Is there any plan for this soon? Would really like to use temporal with bun and this is the main blocker for it to work

JakeCooper:
Bumping here as well. This ended up being the roadblock I hit [here](https://x.com/JustJake/status/1934405671728140346) when I tried to have Claude move us from Node -> Bun

PR #14492, opened 2024-10-11
Implement `--drop`

Jarred-Sumner:
### What does this PR do?

Implement `--drop=console` and `--drop=anyproperty.or.identifier` which removes function calls to that identifier or property access

There are tests

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -1612,6 +1612,11 @@ declare module "bun" {
      * Enable CSS support.
      */
     experimentalCss?: boolean;
+
+    /**
+     * Drop function calls to matching property accesses.
+     */
+    drop?: string[];
   }
 
   namespace Password {
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -1449,6 +1449,7 @@ pub const BundleV2 = struct {
                     .env_files = &.{},
                     .conditions = config.conditions.map.keys(),
                     .ignore_dce_annotations = bundler.options.ignore_dce_annotations,
+                    .drop = config.drop.map.keys(),
                 },
                 completion.env,
             );
@@ -1478,8 +1479,8 @@ pub const BundleV2 = struct {
             bundler.options.emit_dce_annotations = config.emit_dce_annotations orelse !config.minify.whitespace;
             bundler.options.ignore_dce_annotations = config.ignore_dce_annotations;
             bundler.options.experimental_css = config.experimental_css;
-            bundler.options.banner = config.banner.toOwnedSlice();
-            bundler.options.footer = config.footer.toOwnedSlice();
+            bundler.options.banner = config.banner.toOwnedSliceLeaky();
+            bundler.options.footer = config.footer.toOwnedSliceLeaky();
 
             bundler.configureLinker();
             try bundler.configureDefines();
--- a/src/bundler/bundle_v2.zig
+++ b/src/bundler/bundle_v2.zig
@@ -1439,7 +1439,7 @@ pub const BundleV2 = struct {
                     .entry_points = config.entry_points.keys(),
                     .target = config.target.toAPI(),
                     .absolute_working_dir = if (config.dir.list.items.len > 0)
-                        config.dir.toOwnedSliceLeaky()
+                        config.dir.slice()
                     else
                         null,
                     .inject = &.{},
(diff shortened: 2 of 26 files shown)


test/bundler/bundler_drop.test.ts
```diff
@@ -0,0 +1,81 @@
+import { itBundled } from "./expectBundled";
+
+describe("bundler", () => {
+  itBundled("js/drop", {
```
paperclover: the convention for these names has been filename/TestName, so this would be drop/Console, then drop/ReassignKeepsConsole

src/bundler/bundle_v2.zig
```diff
@@ -1478,8 +1479,8 @@ pub const BundleV2 = struct {
             bundler.options.emit_dce_annotations = config.emit_dce_annotations orelse !config.minify.whitespace;
             bundler.options.ignore_dce_annotations = config.ignore_dce_annotations;
             bundler.options.experimental_css = config.experimental_css;
-            bundler.options.banner = config.banner.toOwnedSlice();
-            bundler.options.footer = config.footer.toOwnedSlice();
+            bundler.options.banner = config.banner.toOwnedSliceLeaky();
+            bundler.options.footer = config.footer.toOwnedSliceLeaky();
```
paperclover: whats the reasoning here? i forget what leaky does (reviewing on phone) but it sounds worse. what frees this memory

paperclover (changes requested):
overall

- the variable names internally are so long and verbose
- another use of drop (esbuild lets you do this) is --drop=debugger to drop debugger statements. we can add this later

these nits and the comments are not the most important

the requested change is just:
- can you add to the docs. one in the bundler help page and twice in the "vs esbuild" document

src/bundler/bundle_v2.zig
Jarred-Sumner: previously, it leaked

now it doesn't

paperclover: after reading surrounding code i see why this works but it is extremely error-prone for memory management

paperclover: oh i see

packages/bun-types/bun.d.ts
```diff
@@ -1612,6 +1612,11 @@ declare module "bun" {
      * Enable CSS support.
      */
     experimentalCss?: boolean;
+
+    /**
+     * Drop function calls to matching property accesses.
```
dylan-conway, line 1617: probably should mention dropping debugger statements too

src/defines.zig
```diff
@@ -70,75 +72,87 @@ pub const DefineData = struct {
...
-                    }
-                    break;
+    pub fn fromMergeableInputEntry(user_defines: *UserDefines, key: []const u8, value_str: []const u8, value_is_undefined: bool, method_call_must_be_replaced_with_undefined: bool, log: *logger.Log, allocator: std.mem.Allocator) !void {
+        var keySplitter = std.mem.split(u8, key, ".");
+        while (keySplitter.next()) |part| {
+            if (!js_lexer.isIdentifier(part)) {
+                if (strings.eql(part, key)) {
+                    try log.addErrorFmt(null, logger.Loc{}, allocator, "define key \"{s}\" must be a valid identifier", .{key});
+                } else {
+                    try log.addErrorFmt(null, logger.Loc{}, allocator, "define key \"{s}\" contains invalid identifier \"{s}\"", .{ part, value_str });
```
dylan-conway, line 87: is this supposed to be `..., .{ key, part })`?

src/defines.zig
```diff
@@ -70,75 +72,87 @@ pub const DefineData = struct {
...
-            var keySplitter = std.mem.split(u8, entry.key_ptr.*, ".");
-            while (keySplitter.next()) |part| {
-                if (!js_lexer.isIdentifier(part)) {
-                    if (strings.eql(part, entry.key_ptr)) {
-                        try log.addErrorFmt(null, logger.Loc{}, allocator, "define key \"{s}\" must be a valid identifier", .{entry.key_ptr.*});
-                    } else {
-                        try log.addErrorFmt(null, logger.Loc{}, allocator, "define key \"{s}\" contains invalid identifier \"{s}\"", .{ part, entry.value_ptr.* });
-                    }
-                    break;
+    pub fn fromMergeableInputEntry(user_defines: *UserDefines, key: []const u8, value_str: []const u8, value_is_undefined: bool, method_call_must_be_replaced_with_undefined: bool, log: *logger.Log, allocator: std.mem.Allocator) !void {
```
dylan-conway, line 80: `value_is_undefined` and `method_call_must_be_replaced_with_undefined` could probably become one boolean, but not important

dylan-conway (approved):

Issue #4694, opened 2023-09-09
TypeError: process.binding("tty_wrap") is not implemented in Bun.


legzdev:
### What version of Bun is running?

1.0.0+822a00c4d508b54f650933a73ca5f4a3af9a7983

### What platform is your computer?

Linux 6.4.12-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

Execute a script (`bun script.js`) with:

```js
const input = require("readline-sync").question

let text = input("Input: ")
```

### What is the expected behavior?

Wait for a text input from the user

### What do you see instead?

```
1 | (function (bindingName){"use strict";if(bindingName==="constants")return @processBindingConstants;const issue={fs:3546,buffer:2020,natives:2254,uv:2891}[bindingName];if(issue)throw new Error(`process.binding("${bindingName}") is not implemented in Bun. Track the status & thumbs up the issue: https://github.com/oven-sh/bun/issues/${issue}`);@throwTypeError(`process.binding("${bindingName}") is not implemented in Bun. If that breaks something, please file an issue and include a reproducible code sample.`)})
                       ^
TypeError: process.binding("tty_wrap") is not implemented in Bun. If that breaks something, please file an issue and include a reproducible code sample.
      at binding (:1:20)
      at /home/ghost/node_modules/readline-sync/lib/readline-sync.js:19:8
      at globalThis (/home/ghost/node_modules/readline-sync/lib/readline-sync.js:1329:78)
      at require (:1:20)
      at /home/ghost/bing.js:2:6
      at globalThis (/home/ghost/bing.js:208:5)
```

mifopen:
same with importing `ioredis-mock` which is quite popular.
`ioredis-mock` -> `fengari` -> `readline-sync`

peterabrahamdev:
I had the same error and it's probably due to the "readline-sync" import. When I took it out of my project, the error went away.

Jarred-Sumner:
This will ship in Bun v1.0.23

PR #14501, opened 2024-10-12
Start fixing bugs discovered by Node.js's Node-API tests

### LINKED ISSUE #14336: Segmentation fault | RakNet Module

Download RAGE:MP server, start it and try to run this sample code.

```ts
import { Client } from 'raknet-native'

(async () => {
    const client = new Client('127.0.0.1', 22005, { protocolVersion: 9 })

    await client.connect()

    client.on('encapsulated', (buffer) => {
        console.assert(buffer.toString() == '\xA0 Hello world')
    })
})()
```

### Relevant log output

```
#### PS C:\Users\lol\WebstormProjects\raknet-ragemp> bun run .\index.ts
Bun v1.1.25 (fe62a614) Windows x64
Windows v.win10_fe
CPU: sse42 avx avx2 avx512
Args: "C:\Users\Sasha\.bun\bin\bun.exe" "run" ".\index.ts"
Features: jsc tsconfig(5) 
Builtins: "bun:main" "node:events" "node:fs" "node:os" "node:path" "node:string_decoder" "node:tty" "node:util" "node:util/types" 
Elapsed: 398ms | User: 46ms | Sys: 31ms
RSS: 0.15GB | Peak: 0.15GB | Commit: 0.21GB | Faults: 37318

panic(main thread): Segmentation fault at address 0x5
```

### Stack Trace (bun.report)

Bun v1.1.25 ([`fe62a61`](<https://github.com/oven-sh/bun/tree/fe62a614046948ebba260bed87db96287e67921f>)) on windows x86_64 [RunCommand]

Segmentation fault at address 0x00000005

- `llint_entry`

### LINKED ISSUE #15383: Wrap finalizer for PromiseRaw failed (@napi-rs/canvas)

1.1.36+ededc168c

Linux 5.15.153.1-microsoft-standard-WSL2 x86_64 x86_64

Install `@napi-rs/canvas` and make a JS file with:

```ts
import { Image } from "@napi-rs/canvas/js-binding.js"

const image = new Image();
image.src = "./anythinghere"; // doesnt need to be valid
```

no error or crash, like node

### What do you see instead?

```ts
$ bun index.mjs
2 | import { Image } from "@napi-rs/canvas/js-binding.js"
3 |
4 | const image = new Image();
    ^
error: Wrap finalizer for PromiseRaw failed
 code: "ObjectExpected"

      at /index.mjs:4:1
```

### Additional information

Reported on discord: https://discord.com/channels/876711213126520882/1310131351533654016

### LINKED ISSUE #15429: Read-only instance accessors defined through Node-API incorrectly have setters

1.1.37+8ca0eb831

Linux 6.8.0-1017-azure x86_64 x86_64

Due to the nature of the bug, it is difficult to provide a minimal example.

I noticed the bug while trying to use [`node-pqclean`](https://github.com/tniessen/node-pqclean) from within Bun. I will fix the version to 0.6.0 and disable WebAssembly to ensure that both Node.js and Bun will use the native backend.

```console
npm install pqclean@0.6.0 --pqclean-backend=native
```

Here is a `test.js` file that demonstrates the bug using the native addon that was compiled by `node-pqclean` during its installation:

```js
const { KEM } = require('pqclean');

console.log(Object.getOwnPropertyDescriptor(KEM.prototype, 'keySize'));
console.log(Object.getOwnPropertyDescriptor(KEM.prototype, 'keySize').set?.apply({}));

const kem = new KEM('mceliece8192128');
Object.assign(kem, { keySize: 'error' });
```

Note the use of `Object.assign()` to make sure that the behavior does not depend on strict mode.

Node.js 22 prints the following:

```console
$ node test.js 
{
  get: [Function (anonymous)],
  set: undefined,
  enumerable: false,
  configurable: false
}
undefined
/home/tniessen/bun-bug/test.js:7
Object.assign(kem, { keySize: 'error' });
       ^

TypeError: Cannot set property keySize of #<KEM> which has only a getter
    at Function.assign (<anonymous>)
    at Object.<anonymous> (/home/tniessen/bun-bug/test.js:7:8)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object..js (node:internal/modules/cjs/loader:1689:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:170:5)
    at node:internal/main/run_main_module:36:49

Node.js v22.11.0
```

Note that the property descriptor does not have a setter, and that `Object.assign()` throws an error.

### What do you see instead?

The most recent version of Bun prints the following:

```console
$ bun run test.js 
{
  get: [class get keySize],
  set: [Function],
  enumerable: true,
  configurable: true,
}
true
```

Note that, according to Bun, the property descriptor _does_ have a setter, that calling the setter returns `true`, and that `Object.assign()` does not throw an error.

### Additional information

The property is defined by `node-pqclean` [here](https://github.com/tniessen/node-pqclean/blame/7aa6117c00ee29802b641a31adeff5bb40baf5ed/native/node_pqclean.cc#L674):

```cpp
  Napi::Function func = KEM::DefineClass(env, "KEM", {
    // ...
    Napi::ObjectWrap<KEM>::InstanceAccessor("keySize", &KEM::GetKeySize, nullptr),
    // ...
  });
```

Note that the setter callback is set to `nullptr`.

It seems to me that Bun's implementation of Node-API will always create a no-op setter in that case, instead of omitting the setter, which would explain the behavior I described above:

https://github.com/oven-sh/bun/blob/f21fffd1bf8bee80698efea0938436e53a30c9a2/src/bun.js/bindings/napi.cpp#L562-L570

### LINKED ISSUE #17503: napi crash using tree-sitter + huggingface

Minimal repro:

```ts
import Parser from 'tree-sitter';
import JavaScript from 'tree-sitter-javascript';
import { pipeline } from "@huggingface/transformers";

// ---- THIS CAUSES THE CRASH ----
await pipeline(
  "feature-extraction",
  "Xenova/all-MiniLM-L6-v2"
)
// ---- END OF CAUSING THE CRASH ----

const parser = new Parser();
parser.setLanguage(JavaScript as unknown as Parser.Language);

const code = `
  function hello() {
      return "world";
  }
`;
console.log('Fails here 🤷‍♂️ But only when using the huggingface pipeline');
const tree = parser.parse(code);

console.log(tree.rootNode.toString());
```

Godspeed! Oh, and thanks. Loving bun a lot.

### Relevant log output

```shell
Bun v1.2.2 (c1708ea6) macOS Silicon
macOS v14.4.1
Args: "bun" "test" "Code"
Features: jsc transpiler_cache(5) tsconfig napi_module_register process_dlopen(4) 
Builtins: "bun:sqlite" "detect-libc" "node:buffer" "node:child_process" "node:crypto" "node:events" "node:fs" "node:fs/promises" "node:os" "node:path" "node:stream" "node:string_decoder" "node:url" "node:util" "node:util/types" 
Elapsed: 416ms | User: 253ms | Sys: 134ms
RSS: 0.34GB | Peak: 0.34GB | Commit: 3.36GB | Faults: 886

panic(main thread): Segmentation fault at address 0xCA
```

### Stack Trace (bun.report)

Bun v1.2.2 ([`c1708ea`](<https://github.com/oven-sh/bun/tree/c1708ea6ab529a4c747a8282a24d125dc20b0a63>)) on macos aarch64 [TestCommand]

Segmentation fault at address 0x000000CA

- *2 unknown/js code*
- `napi_open_escapable_handle_scope`

Features: napi\_module\_register, process\_dlopen, jsc, transpiler\_cache, tsconfig, tsconfig

<sub>Sentry Issue: <strong><a href="https://bun-p9.sentry.io/issues/6305232949/">BUN-CQB</a></strong></sub>

190n:
### What does this PR do?

TODO before merging:
- [x] **Get these tests running in CI**
- [x] get other new tests to run from napi.test.ts
- [ ] fix leak test
- [x] ~~remove dependency on JSC change for `napi_get_property_names` (@heimskr)~~ merged JSC change instead

Fixes #14336
Fixes #15383
Fixes #15429
Fixes #17503

<details><summary><h4>Tests in <code>js-native-api</code></h4></summary>

- [x] 2_function_arguments/test.js
- [x] 3_callbacks/test.js
- [x] 4_object_factory/test.js
- [x] 5_function_factory/test.js
- [x] 6_object_wrap/test-object-wrap-ref.js
- [x] 6_object_wrap/test.js
- [x] 7_factory_wrap/test.js
- [x] 8_passing_wrapped/test.js
- [x] test_array/test.js
- [x] test_bigint/test.js
- [x] test_cannot_run_js/test.js
- [x] test_constructor/test.js
- [x] test_constructor/test2.js
- [x] test_constructor/test_null.js
- [x] test_conversions/test.js
- [x] test_dataview/test.js
- [x] test_date/test.js
- [x] test_error/test.js
- [x] test_exception/test.js
- [x] test_exception/testFinalizerException.js
- [x] test_finalizer/test.js
- [x] test_finalizer/test_fatal_finalize.js
- [x] test_function/test.js
- [x] test_general/test.js
- [x] test_general/testEnvCleanup.js
- [x] test_general/testFinalizer.js
- [x] test_general/testGlobals.js
- [x] test_general/testInstanceOf.js
- [x] test_general/testNapiRun.js
- [x] test_general/testNapiStatus.js
- [x] test_handle_scope/test.js
- [x] test_instance_data/test.js
- [x] test_new_target/test.js
- [x] test_number/test.js
- [x] test_number/test_null.js
- [x] test_object/test.js
- [x] test_object/test_exceptions.js
- [x] test_object/test_null.js
- [x] test_promise/test.js
- [x] test_properties/test.js
- [x] test_reference/test.js
- [x] test_reference/test_finalizer.js
- [x] test_reference_double_free/test.js
- [x] test_reference_double_free/test_wrap.js
- [ ] test_string/test.js
- [ ] test_string/test_null.js
- [x] test_symbol/test1.js
- [x] test_symbol/test2.js
- [x] test_symbol/test3.js
- [x] test_typedarray/test.js
</details>

Running Node's tests


robobun:
@190n, your commit 6169f1053a6b0304f781489e5f84baa0310f50b2 is building: [`#4694`](https://buildkite.com/bun/bun/builds/4694)

src/bun.js/api/FFI.h
```diff
@@ -62,6 +62,7 @@ typedef enum {
 } napi_status;
 void* NapiHandleScope__open(void* jsGlobalObject, bool detached);
 void NapiHandleScope__close(void* jsGlobalObject, void* handleScope);
+extern void *napiEnv;
```
190n: rename to something clearer and less likely to collide, perhaps
```suggestion
extern void *Bun__thisFFIModuleNapiEnv;
```

src/bun.js/api/ffi.zig
```diff
@@ -432,6 +434,8 @@ pub const FFI = struct {
                 return error.DeferredErrors;
             }
 
+            _ = TCC.tcc_add_symbol(state, "napiEnv", globalThis.makeNapiEnvForFFI());
```
190n: update to whatever the name in `FFI.h` changes to

src/bun.js/api/ffi.zig
```diff
@@ -1800,7 +1803,7 @@ pub const FFI = struct {
 
             if (this.needsHandleScope()) {
                 try writer.writeAll(
-                    \\  void* handleScope = NapiHandleScope__open(JS_GLOBAL_OBJECT, false);
+                    \\  void* handleScope = NapiHandleScope__open(&napiEnv, false);
```
190n: why do we pass it a `void **`?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -273,6 +276,53 @@ extern "C" bool Bun__resolveEmbeddedNodeFile(void*, BunString*);
 extern "C" HMODULE Bun__LoadLibraryBunString(BunString*);
 #endif
 
+/// Returns a pointer that needs to be freed with `delete[]`.
+static char* toFileURI(std::string_view path)
```
190n, line 301: does this return the right string on Windows?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -389,18 +439,17 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionDlopen,
...
-        JSC::EncodedJSValue exports);
 #if OS(WINDOWS)
 #define dlsym GetProcAddress
 #endif
 
     // TODO(@190n) look for node_register_module_vXYZ according to BuildOptions.reported_nodejs_version
     // (bun/src/env.zig:36) and the table at https://github.com/nodejs/node/blob/main/doc/abi_version_registry.json
-    napi_register_module_v1 = reinterpret_cast<JSC::EncodedJSValue (*)(JSC::JSGlobalObject*,
-        JSC::EncodedJSValue)>(
+    auto napi_register_module_v1 = reinterpret_cast<JSC::EncodedJSValue (*)(napi_env, napi_value)>(
```
190n:
```suggestion
    auto napi_register_module_v1 = reinterpret_cast<napi_value (*)(napi_env, napi_value)>(
```

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -411,14 +460,39 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionDlopen,
 #else
         dlclose(handle);
 #endif
+
         JSC::throwTypeError(globalObject, scope, "symbol 'napi_register_module_v1' not found in native module. Is this a Node API (napi) module?"_s);
         return {};
     }
 
+    // TODO(@heimskr): get the API version without node_api_module_get_api_version_v1 a different way
```
190n, line 505: what's the other way? curious

src/bun.js/bindings/ZigGlobalObject.h
```diff
@@ -287,7 +288,6 @@ class GlobalObject : public Bun::GlobalScope {
 
     Structure* NapiExternalStructure() const { return m_NapiExternalStructure.getInitializedOnMainThread(this); }
     Structure* NapiPrototypeStructure() const { return m_NapiPrototypeStructure.getInitializedOnMainThread(this); }
-    Structure* NAPIFunctionStructure() const { return m_NAPIFunctionStructure.getInitializedOnMainThread(this); }
```
190n, line 271: Also delete the `m_NAPIFunctionStructure` member on line 585

src/bun.js/bindings/napi.h
```diff
@@ -7,14 +7,144 @@
...
 #include "ZigGlobalObject.h"
 #include "napi_handle_scope.h"
 #include "napi_finalizer.h"
 
+#include <unordered_set>
+
+extern "C" void napi_internal_register_cleanup_zig(napi_env env);
+extern "C" void napi_internal_crash_in_gc(napi_env);
+
+struct napi_env__ {
```
190n, line 64:
```suggestion
// Named that so we can manipulate napi_env values directly (since napi_env is defined as a pointer to struct napi_env__)
struct napi_env__ {
```

src/bun.js/bindings/napi.h
```diff
@@ -7,14 +7,144 @@
...
+            napi_internal_enqueue_finalizer(this, finalize_cb, data, finalize_hint);
+        } else {
+            finalize_cb(this, data, finalize_hint);
+        }
+    }
+
+    inline Zig::GlobalObject* globalObject() const { return m_globalObject; }
+    inline const napi_module& napiModule() const { return m_napiModule; }
+
+    inline bool mustAlwaysDefer() const { return m_napiModule.nm_version <= 8; }
```
190n:
```suggestion
    // Returns true if finalizers from this module need to be scheduled for the next tick after garbage collection, instead of running during garbage collection
    inline bool mustDeferFinalizers() const { return m_napiModule.nm_version <= 8; }
```

Aren't finalizers deferred even in version 9? [Node's docs list `node_api_basic_env` as stability experimental](https://nodejs.org/api/n-api.html#node_api_basic_env).

src/bun.js/bindings/napi.h
```diff
@@ -7,14 +7,144 @@
...
+        // Not currently used by Bun -- always nullptr
+        .engine_reserved = nullptr,
+        // Not currently used by Bun -- always zero
+        .engine_error_code = 0,
+        .error_code = napi_ok,
+    };
+
+    void* instanceData = nullptr;
+    WTF::RefPtr<Bun::NapiFinalizer> instanceDataFinalizer;
+    char* filename = nullptr;
```
190n: Use `m_` prefix

src/bun.js/bindings/napi.h
```diff
@@ -141,13 +265,13 @@ class NapiRef {
     void unref();
     void clear();
 
-    NapiRef(JSC::JSGlobalObject* global, uint32_t count, Bun::NapiFinalizer finalizer)
-        : finalizer(finalizer)
+    NapiRef(napi_env env, uint32_t count, WTF::RefPtr<Bun::NapiFinalizer> finalizer, bool defer)
+        : env(env)
+        , globalObject(JSC::Weak<JSC::JSGlobalObject>(env->globalObject()))
+        , finalizer(std::move(finalizer))
```
190n:
```suggestion
        , finalizer(WTFMove(finalizer))
```

src/bun.js/bindings/napi.h
```diff
@@ -141,13 +265,13 @@ class NapiRef {
     void unref();
     void clear();
 
-    NapiRef(JSC::JSGlobalObject* global, uint32_t count, Bun::NapiFinalizer finalizer)
-        : finalizer(finalizer)
+    NapiRef(napi_env env, uint32_t count, WTF::RefPtr<Bun::NapiFinalizer> finalizer, bool defer)
```
190n: Is `defer` here ever different than `env->mustAlwaysDefer()`?

src/bun.js/bindings/napi.h
```diff
@@ -167,12 +291,15 @@ class NapiRef {
...
+    napi_env env = nullptr;
     JSC::Weak<JSC::JSGlobalObject> globalObject;
     NapiWeakValue weakValueRef;
     JSC::Strong<JSC::Unknown> strongRef;
-    Bun::NapiFinalizer finalizer;
+    WTF::RefPtr<Bun::NapiFinalizer> finalizer;
     void* data = nullptr;
     uint32_t refCount = 0;
+    bool isOwnedByRuntime = false;
+    bool defer = false;
```
190n: Use `m_` prefix

src/bun.js/bindings/napi_finalizer.h
```diff
@@ -3,10 +3,24 @@
 #include "root.h"
 #include "js_native_api.h"
 
+extern "C" void napi_internal_enqueue_finalizer(napi_env env, napi_finalize finalize_cb, void* data, void* hint);
+
 namespace Bun {
 
-class NapiFinalizer {
+class NapiFinalizer : public WTF::RefCounted<NapiFinalizer> {
```
190n: Why are these ref-counted now?

src/bun.js/event_loop.zig
```diff
@@ -768,6 +768,7 @@ pub const EventLoop = struct {
     ///
     /// Having two queues avoids infinite loops creating by calling `setImmediate` in a `setImmediate` callback.
     immediate_tasks: Queue = undefined,
+    napi_finalizer_queue: JSC.napi.Finalizer.Queue = undefined,
```
190n: Move this below `next_immediate_tasks`, as the above comment "Having two queues" refers to `immediate_tasks` and `next_immediate_tasks` so they should be immediately after the comment

src/napi/napi.zig
```diff
@@ -15,12 +15,8 @@ const Async = bun.Async;
 
 /// Actually a JSGlobalObject
```
190n:
```suggestion
/// This is `struct napi_env__` from napi.h
```

src/napi/napi.zig
```diff
@@ -52,6 +48,17 @@ pub const NapiEnv = opaque {
         }
         return self.setLastError(.generic_failure);
     }
+
+    pub fn checkGC(self: *NapiEnv) void {
```
190n, line 51:
```suggestion
    /// Assert that we are not currently performing garbage collection
    pub fn checkGC(self: *NapiEnv) void {
```

src/napi/napi.zig
```diff
@@ -52,6 +48,17 @@ pub const NapiEnv = opaque {
...
+        napi_internal_check_gc(self);
+    }
+
+    pub fn getVersion(self: *NapiEnv) u32 {
```
190n, line 56:
```suggestion
    /// Return the Node-API version number declared by the module we are running code from
    pub fn getVersion(self: *NapiEnv) u32 {
```

src/napi/napi.zig
```diff
@@ -72,28 +79,28 @@ pub const Ref = opaque {};
...
     /// Create a new handle scope in the given environment, or return null if creating one now is
     /// unsafe (i.e. inside a finalizer)
     pub fn open(env: *NapiEnv, escapable: bool) ?*NapiHandleScope {
-        return NapiHandleScope__open(env.toJS(), escapable);
+        return NapiHandleScope__open(env, escapable);
     }
 
     /// Closes the given handle scope, releasing all values inside it, if it is safe to do so.
     /// Asserts that self is the current handle scope in env.
     pub fn close(self: ?*NapiHandleScope, env: *NapiEnv) void {
```
190n, line 95: I don't think handle scopes are nullable anymore but I need to check

src/napi/napi.zig
```diff
@@ -1248,7 +1233,7 @@ pub export fn napi_fatal_error(location_ptr: ?[*:0]const u8, location_len: usize
 
     const location = napiSpan(location_ptr, location_len);
     if (location.len > 0) {
-        bun.Output.panic("napi: {s}\n  {s}", .{ message, location });
+        bun.Output.panic("FATAL ERROR: {s} {s}", .{ location, message });
```
190n: Is Node searching for a substring in its tests? It'd be nice to include `napi` or `Node-API` or something more descriptive in the output

src/napi/napi.zig
```diff
@@ -1462,9 +1459,37 @@ pub export fn napi_remove_env_cleanup_hook(env_: napi_env, fun: ?*const fn (?*an
...
+            }
+            napi_internal_remove_finalizer(env, finalizer.fun, finalizer.hint, finalizer.data);
+            if (env.toJS().tryTakeException()) |exception| {
+                _ = env.toJS().bunVM().uncaughtException(env.toJS(), exception, false);
+            }
+        }
+    }
+
+    /// Node defers finalizers to the immediate task queue.
+    /// This is most likely to account for napi addons which cause GC inside of the finalizer.
```
190n:
```suggestion
    /// For Node-API modules not built with NAPI_EXPERIMENTAL, finalizers should be deferred to the
    /// immediate task queue instead of run immediately. This lets finalizers perform allocations,
    /// which they couldn't if they ran immediately while the garbage collector is still running.
```

src/bun.js/bindings/napi.cpp
```diff
@@ -55,24 +56,23 @@
 #include <JavaScriptCore/JSNativeStdFunction.h>
 #include <JavaScriptCore/BigIntObject.h>
 #include <JavaScriptCore/JSWeakMapInlines.h>
+#include <JavaScriptCore/JSMapInlines.h>
```
190n:
```suggestion
```

src/bun.js/bindings/napi.cpp
```diff
@@ -105,6 +105,11 @@ using namespace Zig;
         }                                                       \
     } while (0)
 
+#define NAPI_CHECK_ENV_NOT_IN_GC(_env) \
```
190n, line 124:
```suggestion
// Assert that the environment is not performing garbage collection
#define NAPI_CHECK_ENV_NOT_IN_GC(_env) \
```

src/bun.js/javascript.zig
```diff
@@ -1318,6 +1319,12 @@ pub const VirtualMachine = struct {
     }
 
     pub fn onBeforeExit(this: *VirtualMachine) void {
+        Bun__queueFinishNapiFinalizers(this.global);
+        while (this.isEventLoopAlive()) {
```
Jarred-Sumner: This should only be called if there were napi finalizers, right?

src/bun.js/api/ffi.zig
Jarred-Sumner: We should only be creating the napi env for FFI if napi is in use. Not all the time. Right?

test/js/bun/ffi/ffi.test.fixture.receiver.c
```diff
@@ -153,6 +192,25 @@ static bool JSVALUE_IS_NUMBER(EncodedJSValue val) {
   return val.asInt64 & NumberTag;
 }
 
+static uint8_t GET_JSTYPE(EncodedJSValue val) {
```
Jarred-Sumner: Can we keep the always_inline attributes? I benchmarked this at one point and it did improve the performance some nanoseconds

src/bun.js/bindings/napi.cpp
```diff
@@ -1056,18 +1072,20 @@ extern "C" napi_status napi_remove_wrap(napi_env env, napi_value js_object,
...
+    if (ref->isOwnedByRuntime) {
+        delete ref;
+    } else if (ref->finalizer) {
+        ref->finalizer->clear();
+    }
+
+    external->m_value = nullptr;
     globalObject->napiWraps()->remove(value.asCell());
 
     // don't delete the ref: if weak, it'll delete itself when the JS object is deleted;
```
190n, line 1007: Update this comment to reflect the above logic

src/bun.js/bindings/napi.cpp
```diff
@@ -1628,6 +1697,94 @@ extern "C" napi_status napi_create_dataview(napi_env env, size_t length,
...
+    case napi_uint32_array:
+        return JSC::JSUint32Array::create(globalObject, structure, std::move(arrayBuffer), byteOffset, length);
+    case napi_float32_array:
+        return JSC::JSFloat32Array::create(globalObject, structure, std::move(arrayBuffer), byteOffset, length);
+    case napi_float64_array:
+        return JSC::JSFloat64Array::create(globalObject, structure, std::move(arrayBuffer), byteOffset, length);
+    case napi_bigint64_array:
+        return JSC::JSBigInt64Array::create(globalObject, structure, std::move(arrayBuffer), byteOffset, length);
+    case napi_biguint64_array:
+        return JSC::JSBigUint64Array::create(globalObject, structure, std::move(arrayBuffer), byteOffset, length);
```
190n: Use `WTFMove`

src/bun.js/bindings/napi.cpp
```diff
@@ -1628,6 +1697,94 @@ extern "C" napi_status napi_create_dataview(napi_env env, size_t length,
...
+
+extern "C" napi_status napi_create_typedarray(
+    napi_env env,
+    napi_typedarray_type type,
+    size_t length,
+    napi_value arraybuffer,
+    size_t byte_offset,
+    napi_value* result)
+{
+    NAPI_PREAMBLE_NO_THROW_SCOPE(env);
```
190n: why does this function have to declare its own throw scope? if it's only checking for an already pending exception it can use the regular `NAPI_PREAMBLE`

src/bun.js/bindings/napi.h
```diff
@@ -7,14 +7,144 @@
...
+    }
+
+    void addFinalizer(napi_finalize callback, void* hint, void* data)
+    {
+        m_finalizers.emplace(callback, hint, data);
+    }
+
+    void checkGC()
+    {
+        if (UNLIKELY(!mustAlwaysDefer() && m_globalObject->vm().heap.mutatorState() == JSC::MutatorState::Sweeping)) {
```
190n: Why do we check `mustAlwaysDefer()`? If finalizers are supposed to be deferred, then this check should still succeed, because we shouldn't be invoking the finalizer during GC in the first place.

test/js/bun/ffi/ffi.test.fixture.receiver.c
190n: They're declared always_inline, we just don't repeat the attribute in the definition: https://github.com/oven-sh/bun/blob/657f5b9f6a91136d6a4e1bd1df8bd416d41d62ab/src/bun.js/api/FFI.h#L175

src/bun.js/api/ffi.zig
heimskr: Fixed in [`kai/fix-node-napi-tests`](https://github.com/oven-sh/bun/commit/0bee1c9b5d2f2959b681fee897de4560e983b4c6). Appears to work in `test/js/bun/ffi`; napi envs were produced only when at least one of the functions has a napi_env and/or napi_value argument, and never produced where there were no napi-related functions.

src/bun.js/api/FFI.h
heimskr: [Implemented](https://github.com/oven-sh/bun/commit/9ea9925e9c38d00798ab6a8479b4e03ebc7d98f3)

src/bun.js/bindings/BunProcess.cpp
heimskr: Probably not. It expects ASCII.

src/bun.js/bindings/BunProcess.cpp
heimskr: [Fixed](https://github.com/oven-sh/bun/commit/c17e05c191c014f8f85d47df06856e247d3b4d14)

src/bun.js/bindings/BunProcess.cpp
heimskr: I think in some cases there's a global variable of type `napi_module` that's pre-initialized.

src/bun.js/api/ffi.zig
190n: addressed in c17e05c191c014f8f85d47df06856e247d3b4d14

src/bun.js/bindings/napi.h
heimskr: [Fixed](https://github.com/oven-sh/bun/commit/7993f4fa11b12c759cce880628d97acf19694830)

src/bun.js/bindings/napi.h
heimskr: No. Will fix

src/bun.js/bindings/napi_finalizer.h
heimskr: Good question. Seems to work fine [without refcounting](https://github.com/oven-sh/bun/commit/6e7240b6e79b625cd87afe73d7f59417cb1b4765).

src/napi/napi.zig
heimskr: It is. I'll compromise by setting the message prefix to `NAPI FATAL ERROR: `.

src/bun.js/bindings/napi.cpp
heimskr: I saw `RETURN_IF_EXCEPTION` but neglected to notice `NAPI_RETURN_IF_EXCEPTION`, whoops

src/bun.js/bindings/napi.h
heimskr: Good catch. I was missing an exception check too. There was a case where I was setting `*result` and returning `napi_ok` even with a RangeError pending.

heimskr: Seems I commented on the wrong feedback

src/bun.js/bindings/napi.cpp
heimskr: I was missing an exception check too. There was a case where I was setting `*result` and returning `napi_ok` even with a RangeError pending.

src/bun.js/bindings/BunProcess.cpp
heimskr: UTF8, actually, which it's receiving from the WTF::String. And [the NAPI docs](https://nodejs.org/api/n-api.html#node_api_get_module_file_name) don't specify anything about a required encoding. The `const char **` type suggests to me that it's not expecting the output to be UTF16/UCS2.

src/bun.js/bindings/napi.h
heimskr: Yes, the check should be `!= NAPI_VERSION_EXPERIMENTAL`.

src/bun.js/bindings/napi.h
heimskr: Decided that we should use the `m_` prefix only for private members.

src/bun.js/javascript.zig
heimskr: I changed it to execute the while loop only if there are any pending finalizers.

src/bun.js/api/ffi.zig
```diff
@@ -2523,3 +2552,13 @@ const CompilerRT = struct {
 };
 
 pub const Bun__FFI__cc = FFI.Bun__FFI__cc;
+
+fn makeNapiEnvIfNeeded(functions: []FFI.Function, globalThis: *JSGlobalObject) ?*napi.NapiEnv {
```
190n: nit:
```suggestion
fn makeNapiEnvIfNeeded(functions: []const FFI.Function, globalThis: *JSGlobalObject) ?*napi.NapiEnv {
```

190n (commented):
LGTM @heimskr. Now blocked on testing.

src/bun.js/javascript.zig
heimskr: Actually, that causes test failures in test/js/node/process/process.test.js. It seems we have to tick the event loop a bit regardless of whether the NAPI finalizer queue is empty. This is probably because of the call to `vm.heap.addFinalizer` in `napi_add_finalizer`.

heimskr: Never mind. I [found](https://github.com/oven-sh/bun/commit/a8fa566101e3ec17eb6b9aea7fc00dab9ba2c15e) the underlying issue. Now it will tick the event loop only if necessary while also not dispatching onBeforeExit multiple times.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2792,12 +2792,16 @@ extern "C" void Bun__Process__queueNextTick1(GlobalObject* globalObject, Encoded
...
+extern "C" bool Bun__queueFinishNapiFinalizers(JSGlobalObject* jsGlobalObject)
 {
     Zig::GlobalObject* globalObject = jsCast<Zig::GlobalObject*>(jsGlobalObject);
-    globalObject->queueTask(new EventLoopTask([globalObject](ScriptExecutionContext&) {
-        globalObject->finishNapiFinalizers();
-    }));
+    if (globalObject->hasNapiFinalizers()) {
+        globalObject->queueTask(new EventLoopTask([](ScriptExecutionContext&) {
+            // TODO(@heimskr): do something more sensible
+        }));
```
190n: Comment why this is needed too

src/bun.js/bindings/napi.h
```diff
@@ -13,12 +13,50 @@
...
+    } while (0)
+
+#define NAPI_RELEASE_ASSERT(assertion, ...)                                                                         \
+    do {                                                                                                            \
+        if (UNLIKELY(!(assertion))) {                                                                               \
+            WTFReportAssertionFailureWithMessage(__FILE__, __LINE__, __PRETTY_FUNCTION__, #assertion, __VA_ARGS__); \
+            WTFReportBacktrace();                                                                                   \
+            WTFCrash();                                                                                             \
+        }                                                                                                           \
+    } while (0)
```
190n: I'd prefer to figure out why the WTF assertions don't print in release instead

src/bun.js/bindings/napi.h
```diff
@@ -34,37 +72,127 @@ struct napi_env__ {
         delete[] filename;
     }
 
+    void finishFinalizers()
+    {
+        m_isFinishingFinalizers = true;
+        m_isFinishingFinalizers = false;
```
190n: Why is this needed?

src/bun.js/bindings/napi_external.cpp
```diff
@@ -6,7 +6,7 @@ namespace Bun {
 NapiExternal::~NapiExternal()
 {
     ASSERT(m_env);
-    m_finalizer.call(m_env, m_value);
+    m_finalizer.call(m_env, m_value, !m_env->mustDeferFinalizers());
```
190n, line 9: should this also have `|| !env->inGC()` like napi.h:500?

src/bun.js/bindings/napi.h
heimskr: oops

src/crash_handler.zig
```diff
@@ -1425,7 +1425,9 @@ fn report(url: []const u8) void {
 fn crash() noreturn {
     switch (bun.Environment.os) {
         .windows => {
-            std.posix.abort();
+            // This exit code is what Node.js uses when it calls
+            // abort. This is relied on by their Node-API tests.
+            bun.C.quick_exit(134);
         },
```
190n, line 1468: Can we change Node's test so it accepts whatever exit code `std.posix.abort` results in?

src/crash_handler.zig
```diff
@@ -1772,3 +1774,11 @@ pub fn removePreCrashHandler(ptr: *anyopaque) void {
...
     _ = before_crash_handlers.orderedRemove(index);
 }
+
+export fn Bun__crashHandler(message_ptr: [*]u8, message_len: usize) noreturn {
+    crashHandler(.{ .panic = message_ptr[0..message_len] }, null, @returnAddress());
+}
+
+comptime {
+    _ = &Bun__crashHandler;
+}
```
190n, line 1840: This should be unnecessary if we can get the WTF release assertions to work.

test/napi/node-napi-tests/test/js-native-api/test_instance_data/test.js
```diff
@@ -36,5 +36,6 @@ if (module !== require.main) {
   checkOutput(requireAs(__filename, ['--expose-gc'], runOptions, 'child'));
 
   // Run tests in a worker thread in a child process.
-  checkOutput(requireAs(__filename, ['--expose-gc'], runOptions, 'worker'));
+  // TODO(@heimskr): get finalizers in workers working again
+  // checkOutput(requireAs(__filename, ['--expose-gc'], runOptions, 'worker'));
```
190n: Let's see if this test still fails

test/napi/node-napi-tests/test/node-api/test_cleanup_hook/binding.c
```diff
@@ -8,6 +8,7 @@ static int cleanup_hook_count = 0;
 static void cleanup(void* arg) {
   cleanup_hook_count++;
   printf("cleanup(%d)\n", *(int*)(arg));
+  fflush(stdout);
```
190n, line 11: Commit this change in 190n/node

test/napi/node-napi.test.ts
```diff
@@ -34,6 +34,7 @@ const failingNodeApiTests = [
   "test_threadsafe_function/test.js",
   "test_threadsafe_function/test_legacy_uncaught_exception.js",
   "test_worker_buffer_callback/test.js",
+  "test_worker_buffer_callback/test-free-called.js", // TODO(@heimskr)
```
190n, line 35: Let's see if we can unskip this

test/napi/node-napi.test.ts
```diff
@@ -98,14 +104,14 @@ beforeAll(async () => {
     }
   }
 
-  const parallelism = Math.min(8, os.cpus().length);
+  const parallelism = Math.min(8, os.cpus().length, 1 /* TODO(@heimskr): remove */);
```
190n, line 107: remove

test/napi/node-napi.test.ts
```diff
@@ -67,7 +70,10 @@ beforeAll(async () => {
     .map(t => dirname(t));
   const uniqueDirectories = Array.from(new Set(directories));
 
+  const start = Date.now();
+
   async function buildOne(dir: string) {
+    console.log(`Building in ${dir.replace(/.+node-napi-tests.test./, "")}.`, (Date.now() - start) / 1000, "s elapsed");
```
190n: remove

src/bun.js/event_loop.zig
```diff
@@ -770,6 +770,8 @@ pub const EventLoop = struct {
     immediate_tasks: Queue = undefined,
     next_immediate_tasks: Queue = undefined,
 
+    napi_finalizer_queue: JSC.napi.Finalizer.Queue = undefined,
```
Jarred-Sumner: definitely shouldn't be undefined

src/bun.js/event_loop.zig
```diff
@@ -1275,6 +1277,7 @@ pub const EventLoop = struct {
 
     pub fn tickImmediateTasks(this: *EventLoop, virtual_machine: *VirtualMachine) void {
         _ = this.tickQueueWithCount(virtual_machine, "immediate_tasks");
+        JSC.napi.Finalizer.drain(&this.napi_finalizer_queue);
```
Jarred-Sumner: This should not be here. Instead, it should enqueue a task to drain the napi finalizers when there are any finalizers to drain.

src/bun.js/javascript.zig
```diff
@@ -1322,6 +1327,15 @@ pub const VirtualMachine = struct {
     }
 
     pub fn onBeforeExit(this: *VirtualMachine) void {
+        if (Bun__queueFinishNapiFinalizers(this.global) or this.eventLoop().napi_finalizer_queue.count > 0) {
```
Jarred-Sumner: This should also not be here. Enqueuing the napi finalizers to run should make this unnecessary

heimskr: Replaced the napi finalizer queue with a new task type, which made this unnecessary

190n (commented):
I'm happy with the merge now. Should be OK to merge if CI passes.

heimskr:
Thanks for cleaning up the mess after I merged main into this branch :D

190n:
NVM I broke up bigint tests. Will fix

heimskr (approved):
yay

Issue #4695, opened 2023-09-09
Bun migration guides


mporracindie:
### What is the type of issue?

Documentation is missing

### What is the issue?

We already have great ways to initialize a Nextjs (and other frameworks) app with Bun however I think we should have migrations guides to port existing projects.

PR #14502, opened 2024-10-12
fix label in 3-typescript-bug-report.yml

nektro:
`typescript` label is being deprecated and meant for parser/transpiler bugs but we already have `transpiler`

`types` is the correct label for this template


robobun:
@nektro, your commit c28b183f1fa02acd7389603960dfed531c642ee8 is building: [`#4695`](https://buildkite.com/bun/bun/builds/4695)

Issue #14554, opened 2024-10-14
Error when using tinyexec with Bun runtime:


WooWan:
### What version of Bun is running?

1.1.30+7996d06b8

### What platform is your computer?

Darwin 23.4.0 arm64 arm

## Description
When using the `tinyexec` library with Bun runtime, a TypeError occurs. The error suggests that the `for await...of` syntax is not functioning as expected, possibly due to incompatibility between Bun and Node.js stream APIs. 

## Environment
- Runtime: Bun
- Library: tinyexec

## Code
```javascript
import { x } from "tinyexec";
const result = await x("ls", ["./"], {
  nodeOptions: {
    stdio: "inherit",
    shell: true,
  },
});
```

```
Error Message
512 |     let e = this._process;
513 |     if (!e)
514 |       throw new Error("No process was started");
515 |     let n = "", r = "";
516 |     if (this._streamErr)
517 |       for await (let o of this._streamErr)
                           ^
TypeError: undefined is not a function (near '...o of this._streamErr...')
```

* Expected Behavior
The code should execute without errors and successfully run the ls command.
* Actual Behavior
A TypeError is thrown, indicating that this._streamErr is undefined or not iterable.
Possible Cause
This issue may be due to differences in how Bun implements Node.js APIs, particularly regarding stream handling and the for await...of syntax.

PR #14576, opened 2024-10-15
node:child_process: 'ineherit' stdio should make getters be null

nektro:
Closes https://github.com/oven-sh/bun/issues/14554

image

will leave as draft until i add a test and CI passes

----

```js
import child_process from "node:child_process";

const res = child_process.spawn("ls", {
  stdio: "inherit",
});
console.log(res.stderr);
```

image


robobun:
@nektro, your commit 3b95862fd0cf36a5f51c10e77806a19bac99dd84 is building: [`#4840`](https://buildkite.com/bun/bun/builds/4840)

PR #14578, opened 2024-10-15
tsconfig.json: update excludes

nektro:
these are generated folders that should not be searched


robobun:
@nektro, your commit 361bb2fac587a6bae474d4640cdb33c24f6a4934 is building: [`#4851`](https://buildkite.com/bun/bun/builds/4851)

PR #14602, opened 2024-10-16
Force the c-ares libdir to always be 'lib'

refi64:
Otherwise, on systems whose libdir is 'lib64', libcares.a can't be located. This was already fixed a few years ago in the Makefile (9a7333dd5e9d34d26aa94a771f7d3c142809f054) but never got added to the CMake scripts.

Manual testing (i.e. Bun won't build for me without this).


Electroid:
Thanks!

PR #14626, opened 2024-10-17
fix(install): dedupe and no-verify nested git dependencies

### LINKED ISSUE #12036: Bun install with nested private git repos does not resolve

1.1.16-canary.1+087b83c56

Darwin 23.5.0 arm64 arm

Quite a complicated setup, but:

1. You have a private git dependency in your package.json
2. That remote dependency has more private git dependencies
3. One of the remote's private git dependencies has the same name / url as another local private git dependency (think of `logger: 'git@pathtologger'`)
4. Run bun install

Bun installs the remote git dependency and all its dependencies.

### What do you see instead?

Screenshot 2024-06-21 at 16 35 37

### Additional information

This happens when the `resolutions` and `dependencies` ArrayLists aren't in sync at the end of the install. And likely due to the remote's dependency ID not being set in the resolutions somehow - but it is being installed correctly.

Eckhardt-D:
closes #12036 

Nested git dependencies with an invalid id are most likely due to duplicates (in a nested git dep) of deps in the root package. This fix is probably more of a workaround than a fix - the issue seems to be related to the dynamics of how git dependencies are tracked in the resolution list.

E.g. a root package JSON may have something like:

```json
{
   "logger": "git@gitlab.com:org/logger.git",
   "core-entities": "git@gitlab.com:org/entities.git"
}
```

And then the `core-entities` package something like:

```json
{
  "logger": "gitlab:org/logger"
}
```

The root package will correctly resolve and assign an ID to the package cloned from git. But nested deps within that package that reference the same package always gets an `invalid` package ID of `4294967295` which automatically qualifies it to be a failed dependency in `verifyResolutions`

### What does this PR do?

- [x] Code changes

- [x] I ran my tests locally and they pass (`bun-debug install`)

> Install ran on a project with nested private git dependencies


dylan-conway (changes requested):
> This fix is probably more of a workaround than a fix - the issue seems to be related to the dynamics of how git dependencies are tracked in the resolution list.

What I believe is happening is in `enqueueDependencyWithMainAndSuccessFn` we check `hasCreatedNetworkTask` for git dependencies, and if it returns true `successFn` isn't called with the dependency and package id once the task has finished.

I think a possible solution to fix this would be appending to a list of dependency ids everytime `hasCreatedNetworkTask` returns early for git dependencies, and draining the list when `successFn` is called on line 5320.

For now, we can't merge this pr because some git dependencies with `invalid_package_id` in `verifyResolutions` might be a real failure.

Issue #7274, opened 2023-11-22
`bun install` installs optional dependency with optional=false in bunfig.toml


Osmose:
### What version of Bun is running?

1.0.13+d8ca6ca69

### What platform is your computer?

Darwin 22.5.0 x86_64 i386

### What steps can reproduce the bug?

Create a directory with two files:

### `package.json`
```json
{
  "name": "test",
  "dependencies": {
    "chokidar": "^3.5.3"
  }
}
```

### `bunfig.toml`
```toml
[install]
optional = false
```

Then run `bun install` and `ls node_modules/fsevents` to check if `fsevents` (which is an optional dependency for `chokidar`) was installed.

### What is the expected behavior?

`node_modules/fsevents` wasn't installed and doesn't exist.

### What do you see instead?

`fsevents` is installed under `node_modules/fsevents`.

### Additional information

Ran into this while attempting to make a standalone binary with `bun build` that uses chokidar—#3903 prevents this from working since `fsevents` is a native module, so I'd like to avoid installing it since it's an optional dependency to chokidar, but bun seems to install it regardless of my `bunfig.toml`, `--production` flags, etc.

I also checked that it wasn't a cache issue by running `rm -rf ~/.bun/install/cache`, it still installs fsevents.

joakimbeng:
Having the same issue with a dependency that has `typescript` as a `peerOptional` dependency, but it's installed no matter what I try.

infrahead:
Same on Linux.  Thought I was losing my mind.  And this is not just optional dependences, but none of the dependency types documented under https://bun.sh/docs/runtime/bunfig#package-manager follow those settings.  I've tried setting dev and peer to false and are always installed.

Also documented here: https://bun.sh/docs/cli/install#configuration

EDIT: unless I'm misunderstanding how this is expected to work.  I can now see that the settings apply to my own project dependencies, but NOT to transitive dependencies (where optional still installed)

joakimbeng:
For some reason it seems to skip the optional dependencies correctly when I install dependencies as part of a `docker build` when I have the dependency's optional dependency as a dev dependency and only install production dependencies though.

This is my Dockerfile:

```Dockerfile
FROM oven/bun:1.0.26-alpine
ENV CI true
ENV NODE_ENV production

COPY package.json package.json
COPY bun.lockb bun.lockb

RUN bun install --production --frozen-lockfile
```

[...]

PR #14629, opened 2024-10-17
fix(install): Skip optional dependencies if false in bunfig.toml

### LINKED ISSUE #14619: Optional dependencies cause package installations to fail

1.1.30+7996d06b8

Darwin 24.0.0 arm64 arm

Installing [pdf.js](https://github.com/mozilla/pdf.js) always pulls in the optional `canvas` dependency, ignoring bunfig settings. On a M1 Mac, the canvas installation fails, which also causes `bun install` to fail in the process. 

## Steps to Reproduce

Try installing the following package.json on a M1 mac without Cairo. The same behavior can also be reproduced where any library pulls in an optional dependency, which fails to install.

### package.json
```typescript
{
  "name": "test",
  "private": true,
  "version": "0.0.1",
  "main": "dist/index.js",
  "types": "dist/index.d.ts",
  "files": [
    "dist"
  ],
  "dependencies": {
        "pdf.js": "4.7.76"
  },
  "scripts": {
    "build": "bun run build.js"
  }
}
```

Run: `bun install` 

Dependency Installation proceeds without installing the optional dependency, or the installation succeeds even if the optional depdency fails to install 

### What do you see instead?

Dependency Installation fails. 

### Additional information

Linked Issues: 

- https://github.com/oven-sh/bun/issues/7274

Eckhardt-D:
### What does this PR do?

Closes #14619
Closes #7274

Checks the optional dependency feature in `bunfig.toml` before creating any network tasks and installing the dependency, but still lists it as optional in the lockfile.

- [x] Code changes
- [x] Ran `bun-debug install` with the reproduction example in https://github.com/oven-sh/bun/issues/14619

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -4201,6 +4201,8 @@ pub const PackageManager = struct {
             };
         } else if (behavior.isPeer() and !install_peer) {
             return null;
+        } else if (behavior.isOptional() and !this.options.remote_package_features.optional_dependencies) {
+            return null;
         }
 
         // appendPackage sets the PackageID on the package
--- a/test/cli/install/registry/bun-install-registry.test.ts
+++ b/test/cli/install/registry/bun-install-registry.test.ts
@@ -1675,6 +1675,64 @@ describe("optionalDependencies", () => {
   }
 });
 
+describe("optionalDependencies", () => {
+  test("should not install optional deps if false in bunfig", async () => {
+    await writeFile(
+      join(packageDir, "bunfig.toml"),
+      `
+  [install]
+  cache = "${join(packageDir, ".bun-cache")}"
+  optional = false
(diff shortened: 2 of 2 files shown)


dylan-conway:
This looks good. I don't think we have an existing test for this functionality, could you add one to `bun-install-registry.test.ts`?

Eckhardt-D:
@dylan-conway I added a passing test by reusing some of the existing packages - which I'm not sure is the right move? Noticed a bunch of new generations in `test/cli/install/registry/packages/...` after doing so. And an updated bun.lockb which I left out of the commit for now? Advice?

dylan-conway (approved):

dylan-conway:
I'll fix bun.lockb separately, and the package changes are expected. Thanks!

Issue #4977, opened 2023-09-11
Empty Response with 0 KB Body When Using 'Authorization' Header in Bun Fetch


prstance:
### What version of Bun is running?

1.0.0+ea56182c5a45be7978de8521f6cdaee0e0772434

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

Steps to Reproduce the Bug:

1. Create a JavaScript script to perform an HTTP fetch request with an 'Authorization' header.
2. Use a specific URL that requires authentication, such as an API endpoint.
3. Include the 'Authorization' header in the request with valid credentials.
`fetch("xxx", {
      "headers": {
        "Authorization": "xxx",
        "User-Agent": "xxx"
      }
    }).then(response => response.text()).then(json => console.log(json))`
6. Check the response received from the fetch on Bun, and see the difference with NodeJs.
![image](https://github.com/oven-sh/bun/assets/142694305/dfd95b0e-ffb0-42f4-a810-59c135bc58ac)

Bug Explanation:
When performing the fetch request with the 'Authorization' header included, Bun returns a response with an empty body, which has a size of 0 KB. This behavior is unexpected and not in line with the normal behavior of other HTTP client libraries and tools like Node.js fetch, Python, PHP, and curl.

The bug appears to be specific to Bun when handling requests with the 'Authorization' header. It results in an incomplete response, making it impossible to retrieve the expected data from the server when authentication is required.

### What is the expected behavior?

The expected behavior is to receive a response from the server with the correct body data. The response body should contain the data or content that corresponds to the requested URL and authentication credentials, allowing the script to process and utilize the received information as needed.
![image](https://github.com/oven-sh/bun/assets/142694305/dfd95b0e-ffb0-42f4-a810-59c135bc58ac)

### What do you see instead?

Instead of receiving the expected response with the relevant data, when using Bun to make a fetch request with the 'Authorization' header included, the observed behavior is as follows:

1. The response received from the server contains an empty body.
2. The size of the response body is 0 KB, indicating that no data or content has been successfully retrieved.
3. This behavior is inconsistent with other HTTP client libraries and tools, such as Node.js fetch, Python, PHP, and curl, which return the correct response data when the 'Authorization' header is provided.

In summary, the issue with Bun is that it returns an empty response body (0 KB) when the 'Authorization' header is used, which is not the expected behavior for a functioning HTTP client.

### Additional information

[...]

PR #14643, opened 2024-10-17
vscode: set the launch configs' cwd to the root

nektro:
will align it with the ci runners and manually running from the terminal


robobun:
@nektro, your commit 07f178b320162b95b0857228f2baf9ff7e3c307a is building: [`#4977`](https://buildkite.com/bun/bun/builds/4977)

PR #14645, opened 2024-10-17
fix one of the failure checks in expectBundled.ts

nektro:
`.exitCode` is `BundlerTestRunOptions.exitCode?: number | undefined`
this if statement looks to be wanting to check for undefined but `0` is also a falsey value

this fixes a logic bug in that if statement
and will additionally help us debug a windows ci failure in `test/bundler/bundler_compile.test.ts`


robobun:
@nektro, your commit c622cc54100e0796470230aae085d998e994dba8 is building: [`#4984`](https://buildkite.com/bun/bun/builds/4984)

Issue #5085, opened 2023-09-12
Bun not running in termux


Armanidrisi:
I've installed the bun in termux using `curl -fsSL https://bun.sh/install | bash` and added ```export BUN_INSTALL="$HOME/.bun"
export PATH=$BUN_INSTALL/bin:$PATH``` on .bash_profile file at home directory but when is try to run bun i found error `bash: /data/data/com.termux/files/home/.bun/bin/bun: cannot execute: required file not found`

romanovj:
bun linked against glibc, so you should use termux glibc enviroment

pkg i pacman
pacman-key --init
pacman-key --populate
pacman -Syu
pacman -S gpkg-dev
grun --set bun
grun --shell
./bun install --backend=copyfile glob

--backend=copyfile shouldbe used because hardlinks are not allowed by selinux

Sourdface:
When I try to run `pacman -S gpkg-dev` I get `error: target not found: gpkg-dev`

romanovj:
> When I try to run `pacman -S gpkg-dev` I get `error: target not found: gpkg-dev`

now it's

```
yes | pkg upg
pkg i pacman patchelf -y
pacman-key --init
pacman-key --populate
pacman -Syu
pacman -Sy glibc-runner --assume-installed bash,patchelf
```

Sourdface:
While I was attempting to solve this I tried adding a line to `$PREFIX/etc/pacman.d/serverlist`

```
Server = https://service.termux-pacman.dev/gpkg-dev/$arch
```

and then tried your original instructions. It told me I file conflicts, so used `--overwrite "*"` to try to resolve them. Now I think my termux might be corrupted. If I try your updated instructions then I get warnings that local versions are newer than gpkg versions, and when I run `grun` it says `Error: binary not found`.

Any idea how to undo the damage?

hoangpq:
You can install the Debian distribution on Termux, working for me!

https://github.com/denoland/deno/issues/4862#issuecomment-1049502951

<img width="465" alt="image" src="https://github.com/oven-sh/bun/assets/5939997/f00ebbec-f34f-4d16-9a75-cc4d49b5f199">

Armanidrisi:
How?

hoangpq:
@Armanidrisi check the comment https://github.com/denoland/deno/issues/4862#issuecomment-1049502951

```
# If you already have a PRoot distro set up you can just login to it and skip to the 3rd command.
apt update && apt install proot-distro -y
[...]

PR #14692, opened 2024-10-19
Fix ctrl+c sometimes not actually exiting

Jarred-Sumner:
### What does this PR do?

Fix ctrl+c sometimes not actually exiting. Also, don't print the signal message when you receive SIGINT or SIGTERM from the child process - no need to raise it twice.

Before:

image

After:

image

Manually

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -574,27 +574,33 @@ pub const RunCommand = struct {
                     },
 
                     .signaled => |signal| {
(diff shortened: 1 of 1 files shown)


robobun:
@Jarred-Sumner, your commit 3d3410c4ee417aa67dce9e5b94b6a3fbd5f830d7 is building: [`#5085`](https://buildkite.com/bun/bun/builds/5085)

PR #14695, opened 2024-10-20
tls(Server) fix connectionListener and make alpnProtocol more compatible with node.js

cirospaciari:
### What does this PR do?

- [X] Code changes

Test added.

### DIFF
--- a/src/js/node/net.ts
+++ b/src/js/node/net.ts
@@ -164,7 +164,7 @@ const Socket = (function (InternalSocket) {
         self._secureEstablished = !!success;
 
         self.emit("secure", self);
-
+        self.alpnProtocol = socket.alpnProtocol;
         const { checkServerIdentity } = self[bunTLSConnectOptions];
         if (!verifyError && typeof checkServerIdentity === "function" && self.servername) {
           const cert = self.getPeerCertificate(true);
@@ -291,10 +291,7 @@ const Socket = (function (InternalSocket) {
 
         if (typeof connectionListener == "function") {
           this.pauseOnConnect = pauseOnConnect;
-          if (isTLS) {
-            // add secureConnection event handler
-            self.once("secureConnection", () => connectionListener.$call(self, _socket));
-          } else {
+          if (!isTLS) {
             connectionListener.$call(self, _socket);
           }
         }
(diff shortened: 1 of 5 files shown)


robobun:
@cirospaciari, your commit 4e0ffb5eccab585cfb0658f01ac6ce640775da2e is building: [`#5087`](https://buildkite.com/bun/bun/builds/5087)

src/js/node/tls.ts
```diff
@@ -330,6 +330,7 @@ const TLSSocket = (function (InternalTLSSocket) {
     #socket;
     #checkServerIdentity;
     #session;
+    alpnProtocol = null;
```
Jarred-Sumner, line 333: Does node default to null here? If you want to make sure the property is declared, it is better to use undefined I think.

Jarred-Sumner: Answer: yes, `null`:

https://github.com/nodejs/node/blob/2eff28fb7a93d3f672f80b582f664a7c701569fb/lib/_tls_wrap.js#L533-L546

src/js/node/net.ts
```diff
@@ -705,7 +718,6 @@ const Socket = (function (InternalSocket) {
         this.#final_callback = callback;
       } else {
         // emit FIN not allowing half open
-        this[bunSocketInternal] = null;
```
cirospaciari, line 709: lets only detach on _destroy so we can get certs and more info after this

src/js/node/net.ts
```diff
@@ -684,14 +688,23 @@ const Socket = (function (InternalSocket) {
...
+
+      if (this.writableFinished) {
+        // closed we can detach the socket
         this[bunSocketInternal] = null;
-        // we still have a socket, call end before destroy
-        process.nextTick(endNT, socket, callback, err);
-        return;
+      } else {
+        // lets wait for the finish event before detaching the socket
+        this.once("finish", detachAfterFinish);
```
cirospaciari, line 709: lets make sure to only detach after finish and to close both sides so "close" will always be emitted this same code is used on http2, the test added uses the "close" event if you comment the alpnProtocol part will hang in older versions.

src/js/node/net.ts
```diff
@@ -329,7 +331,8 @@ const Socket = (function (InternalSocket) {
         } else {
           self.authorized = true;
         }
-        self.server.emit("secureConnection", self);
+        server[bunSocketServerOptions]?.connectionListener?.$call(server, self);
```
Jarred-Sumner: If connectionListener is undefined, this will error

is that intentional?

cirospaciari: was not intentional fixed in https://github.com/oven-sh/bun/pull/14695/commits/bf3f328a659189400894deca043032a91a66f71f

PR #14704, opened 2024-10-20
Update instructions to run a `bun` application via `pm2` with the use…

sourcegr:
… of `pm2.config.js`

pm2 will not run unless the bun executable is in the path

### What does this PR do?

This PR updates the pm2.config.js file to make sure pm2 can find bun in the PATH.


sourcegr:
I was getting

```
[PM2][ERROR] Error: Interpreter ~/.bun/bin/bun is NOT AVAILABLE in PATH. (type 'which ~/.bun/bin/bun' to double check.)
```

when I run 

```
pm2 start pm2.config.cjs
```

pm2 version: `5.3.1`
bun version: `1.1.31`
OS: Ubuntu 22.04.2 LTS

PR #14706, opened 2024-10-20
feat: install specific version of Bun by `bun upgrade <version>`

mchoo7:
### What does this PR do?

A specific version of Bun can be installed by the command `bun upgrade
<version>`. Installing canary and stable builds is now done by
`bun upgrade <canary|stable>.

By the way, what does `bun upgrade --profile` do? The `--profile` option should be documented.

- [x] Code changes

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -134,8 +134,16 @@ Save the file. You'll need to open a new shell/terminal window for the changes t
 
 Once installed, the binary can upgrade itself.
 
+To upgrade to the latest version:
+
 ```sh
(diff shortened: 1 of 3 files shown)


fel1x-developer:
ping

Nanome203:
ping

nektro:
removing `--canary` etc is a breaking change and a specific version may already be installed with the installation script via:

```sh
$ curl -fsSL https://bun.sh/install | bash -s bun-v1.2.5
```

PR #14717, opened 2024-10-21
Fixed compilation issues with no outdir

### LINKED ISSUE #8280: error: cannot write multiple output files without an output directory

BjornTheProgrammer:
### What does this PR do?

This PR fixes #8280, by changing
```zig
if (root_path.len == 0 and c.parse_graph.additional_output_files.items.len > 0 and !c.resolver.opts.compile) {
    try c.log.addError(null, Logger.Loc.Empty, "cannot write multiple output files without an output directory");
    return error.MultipleOutputFilesWithoutOutputDir;
}

```

```zig
const more_than_one_output = c.parse_graph.additional_output_files.items.len > 0 or c.parse_graph.generate_bytecode_cache or (has_css_chunk and has_js_chunk);

if (!c.resolver.opts.compile and more_than_one_output and (c.resolver.opts.outfile or c.resolver.opts.output_to_stdout)) {
    try c.log.addError(null, Logger.Loc.Empty, "cannot write multiple output files without an output directory");
    return error.MultipleOutputFilesWithoutOutputDir;
}
```

- [x] Code changes

I ran `./build/debug/bun-debug test bundle`

I wrote two automated bundler tests and extended the expectBundled to enable coverage of the issue this pr seeks to solve.


test/bundler/bundler_edgecase.test.ts
```diff
@@ -2031,6 +2031,41 @@ describe("bundler", () => {
...
+      `,
+      "/index.html": `
+        <head></head>
+      `,
+    },
+    generateOutput: false,
+    backend: "api",
+    onAfterApiBundle: build => {
+      expect(build.success).toEqual(true);
+      expect(build.outputs).toBeArrayOfSize(2);
```
paperclover, line 2048: can you verify that the build.outputs items are correct? mainly if you read the outputted `index.html`, you get the text. maybe also check if it's content type is right.

src/options.zig
```diff
@@ -1529,6 +1529,9 @@ pub const BundleOptions = struct {
     /// So we have a list of packages which we know are safe to do this with.
     unwrap_commonjs_packages: []const string = &default_unwrap_commonjs_packages,
 
+    outfile: bool = false,
```
paperclover: outfile being a boolean is kind of silly, can we do without this?

src/options.zig
```diff
@@ -1529,6 +1529,9 @@ pub const BundleOptions = struct {
...
+    output_to_stdout: bool = false,
```
paperclover: then this could be changed to `supports_multiple_outputs`

paperclover (changes requested):
this seems mostly reasonable. have a request to change how the information is passed from build_command to bundle_v2 and one nit on the test

test/bundler/bundler_edgecase.test.ts
BjornTheProgrammer: Makes sense, have done so in newest commit!

src/options.zig
BjornTheProgrammer: You're totally right, folded this into the variable supports_multiple_outputs

paperclover (approved):

BjornTheProgrammer:
I've rebased to the latest, and tested to ensure everything still works.

BjornTheProgrammer:
I'm really confused by these build tests. Are the tests just failing sporadically, also why does it seem like some just don't run? Is it an issue with my code?

Issue #5828, opened 2023-09-20
BrokenPipe error when running a lockfile then breaking the pipe


paperclover:
### What platform is your computer?

mac

### What steps can reproduce the bug?

run `bun bun.lockb | head`

![image](https://github.com/oven-sh/bun/assets/24465214/be950914-3de6-4502-8efe-bf7b37e81895)

### What is the expected behavior?

the error not to be shown. head is supposed to break the pipe to only show the first few lines.

Jarred-Sumner:
confirming this is still a bug

```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1
# bun ./bun.lockb --hash: 0FFF2CA5F690EC90-85b0f98a287b3db9-81166094E0D0E3FE-a4c5192a9478a6cd


"@biomejs/biome@^1.5.3":
  version "1.7.3"
  resolved "https://registry.npmjs.org/@biomejs/biome/-/biome-1.7.3.tgz"
  integrity sha512-ogFQI+fpXftr+tiahA6bIXwZ7CSikygASdqMtH07J2cUzrpjyTMVc9Y97v23c7/tL1xCZhM+W9k4hYIBm7Q6cQ==
  optionalDependencies:
error: An internal error ocurred (EPIPE)
```

PR #5146, opened 2023-09-12
docs: fix typo in import.meta.resolve

jonathantneal:
### What does this PR do?

This fixes a small punctuation typo in the docs for `import.meta.resolve`. It adds a missing quotation.

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

### How did you verify your code works?

- [x] The missing quotation is now present.

PR #14728, opened 2024-10-22
pm: BrokenPipe is not fatal when printing the lockfile

nektro:
Closes https://github.com/oven-sh/bun/issues/5828


robobun:
@nektro, your commit 344b23976cccd6402ee3b195c9a6ef1689ab0d6e is building: [`#5146`](https://buildkite.com/bun/bun/builds/5146)

Jarred-Sumner:
can you add a test? it should use a `|` in a shell

Issue #14182, opened 2024-09-26
undefined symbol: uv_hrtime


SinBirb:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Linux 6.10.10-zen1-1-zen x86_64 unknown

### What steps can reproduce the bug?

We have a large monorepo with dd-trace. I'm trying to run our tests with bun using `bun test <filename>.spec.ts`.

### What is the expected behavior?

The test runs

### What do you see instead?

The following output:

bun: symbol lookup error: /home/<redacted-project-root-path>/node_modules/@datadog/native-metrics/prebuilds/linux-x64/node-napi.node: undefined symbol: uv_hrtime

zackradisic:
Closing this as we now use [this issue](https://github.com/oven-sh/bun/issues/18546) for tracking progress on libuv functions.

Issue #5175, opened 2023-09-13
`for await` from console only works once


robobun:
I'm using the following code mentioned in the docs to read lines from the console:
```ts
async function readInput(): void {
  for await (const line of console) {
    console.log('Reading input.');
    if (line == "break") {
      break;
    }
  }
}
```
This works well if I run it once, but if I try to read the console input multiple times (by calling `readInput` in a loop for example), the input isn't read any more after the first iteration and the `for await (const line of console)` loop isn't called anymore.
Is this a bug or am I doing something wrong?

Example that doesn't work:
```ts
async function readInputMultipleTimes(): void {
  for (int i = 0; i < 10; i++) {
    // Only the first readInput() will work properly
    await readInput();
  }
}
```

_Originally reported on Discord: [`Reading from console using `for await (const line of console)` only works once`](https://discord.com/channels/876711213126520882/1151155593323761734/1151155593323761734)_

PR #14747, opened 2024-10-22
Add uv_hrtime()

Jarred-Sumner:
### What does this PR do?

Add uv_hrtime()

Not tested

### DIFF
--- a/src/bun.js/bindings/uv-posix-polyfills.cpp
+++ b/src/bun.js/bindings/uv-posix-polyfills.cpp
@@ -12,4 +12,22 @@ uv_pid_t uv_os_getppid()
     return getppid();
 }
 
+#if OS(LINUX)
+#include <time.h>
+
+uint64_t uv_hrtime()
+{
+    timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return ts.tv_sec * 1000000000ULL + ts.tv_nsec;
+}
+
+#elif OS(DARWIN)
+#include <mach/mach_time.h>
+uint64_t uv_hrtime()
+{
+    return mach_absolute_time();
+}
+#endif
+
 #endif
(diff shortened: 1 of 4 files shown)


robobun:
@Jarred-Sumner, your commit e6af5d6d164ad059e8ec70e2676ad5a1e5ed64c8 is building: [`#5175`](https://buildkite.com/bun/bun/builds/5175)

190n:
This addresses the direct cause of #14182, but it won't make that package work as [it does a bunch more libuv stuff](https://github.com/DataDog/dd-native-metrics-js/blob/main/src/metrics/EventLoop.hpp).

src/bun.js/bindings/uv-posix-polyfills.cpp
```diff
@@ -12,4 +12,22 @@ uv_pid_t uv_os_getppid()
...
 }
 
+#if OS(LINUX)
+#include <time.h>
+
+uint64_t uv_hrtime()
+{
+    timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return ts.tv_sec * 1000000000ULL + ts.tv_nsec;
```
190n, line 22:
```suggestion
    return ts.tv_sec * 1'000'000'000ULL + ts.tv_nsec;
```

190n (changes requested):
LGTM if there's a test

src/bun.js/bindings/uv-posix-polyfills.cpp
```diff
@@ -12,4 +12,22 @@ uv_pid_t uv_os_getppid()
...
+}
+
+#elif OS(DARWIN)
+#include <mach/mach_time.h>
+uint64_t uv_hrtime()
+{
+    return mach_absolute_time();
```
dylan-conway, line 29: libuv uses `mach_continuous_time` and uses `mach_timebase_info` with the result.
https://github.com/libuv/libuv/blob/v1.x/src/unix/darwin.c#L62

190n: `continuous` [increments](https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time) while the system is asleep and `absolute` [doesn't](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time)

Issue #5177, opened 2023-09-13
`process.send is not a function` using pm2 and nodemon


robobun:
Hello! I'm creating a Discord bot manager for all my processes. However, it seems that pm2 and nodemon aren't able to be installed/used with Bun. For example, nodemon throws an error when using `nodemon.start()` (see screenshot for errors): https://sourceb.in/i0lOjDWqTB
And for `pm2.start()`, I get an error saying that pm2 is unable to connect (likely an issue with node:net/node:dgram not being fully implemented) (see screenshot for errors): https://sourceb.in/RTH2H0WHrt

![image.png](https://cdn.discordapp.com/attachments/1151218222075486269/1151218222402654228/image.png)

![image.png](https://cdn.discordapp.com/attachments/1151218222075486269/1151218222687850608/image.png)

_Originally reported on Discord: [`Using pm2/nodemon`](https://discord.com/channels/876711213126520882/1151218222075486269/1151218222075486269)_

Eltik:
Hello! I'm the creator of this issue. I've created a repository with the bare minimum code to show what the issue is.
https://github.com/Eltik/Bun-pm2-nodemon

paperclover:
the commit that i marked as fixes this issue only fixes the first half of it. the failed to connect thing is something ive noticed in a few other spots

PR #14749, opened 2024-10-23
Don't wrap fcntl on arm64

Jarred-Sumner:
### What does this PR do?

Don't wrap fcntl on arm64

### DIFF
--- a/cmake/targets/BuildBun.cmake
+++ b/cmake/targets/BuildBun.cmake
@@ -846,6 +846,12 @@ else()
     set(LLD_NAME lld-${LLVM_VERSION_MAJOR})
   endif()
 
+  if(ARCH STREQUAL "x86_64")
+    set(x86_64_WRAP_FCNTL "-Wl,--wrap=fcntl -Wl,--wrap=fcntl64")
+  else()
+    set(x86_64_WRAP_FCNTL "")
+  endif()
+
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit e73ba34777a6b9a7ce8e42e0ffc22f9e151604cf is building: [`#5177`](https://buildkite.com/bun/bun/builds/5177)

cmake/targets/BuildBun.cmake
```diff
@@ -846,6 +846,12 @@ else()
     set(LLD_NAME lld-${LLVM_VERSION_MAJOR})
   endif()
 
+  if(ARCH STREQUAL "x86_64")
```
Electroid: This check always fails.

```suggestion
  if(ARCH STREQUAL "x64")
```

cmake/targets/BuildBun.cmake
```diff
@@ -846,6 +846,12 @@ else()
...
+    set(x86_64_WRAP_FCNTL "-Wl,--wrap=fcntl -Wl,--wrap=fcntl64")
```
Electroid, line 850: Please add a comment why we are doing this or what this fixes

Electroid (changes requested):

Issue #5189, opened 2023-09-13
case insensitive scopes - private registry packages


robobun:
```toml
[install]
registry = "https://registry.npmjs.org"

[install.scopes]

"@randomscope" = { token = "$NPM_TOKEN", url = 'https://npm.pkg.github.com' }

```
I have an edge case when an internal package package.json name property pascalcase or all lowercase and when defining the scope name in the toml config I get this error

```bash
uh-oh: internal: package name mismatch expected "@randomscope/lib" but received "@randomScope/lib"
bun will crash now 😭😭😭
```

the mixture of names on my side is my fault and will correct but i couldn't find a raised issue (maybe i didn't try hard enough) where scopes are case insensitive or is this a known issue?

```bash
----- bun meta -----
Bun v1.0.0 (822a00c4) macOS Silicon 22.6.0
InstallCommand: dotenv bunfig 
HTTP: 155
Elapsed: 897ms | User: 100ms | Sys: 238ms
RSS: 86.93MB | Peak: 86.93MB | Commit: 109.08MB | Faults: 379
----- bun meta -----

0   0x102322624 WTFGetBacktrace
1   ??? Bun__crashReportDumpStackTrace
```

_Originally reported on Discord: [`case insensitive scopes - private registry packages`](https://discord.com/channels/876711213126520882/1150879457922531432/1150879457922531432)_

PR #14755, opened 2024-10-23
Reduce default max network connection limit from 256 to 48 in bun install

Jarred-Sumner:
### What does this PR do?

This reduces the default max HTTP/HTTPS connection limit from 256 to 48 in bun install.

256 simultaneous HTTP requests is more than enough to saturate networks

When using a proxy, the default limit is 16 instead of 48.

Fixes https://github.com/oven-sh/bun/issues/4066

We already automatically adjust when there's a network error

There are tests

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -1,10 +1,10 @@
-// Default to a maximum of 48 simultaneous HTTP requests for bun install if no proxy is specified
+// Default to a maximum of 64 simultaneous HTTP requests for bun install if no proxy is specified
 // if a proxy IS specified, default to 16
 // https://github.com/npm/cli/issues/7072
 // https://pnpm.io/npmrc#network-concurrency (pnpm defaults to 16)
 // https://yarnpkg.com/configuration/yarnrc#networkConcurrency (defaults to 50)
 const default_max_simultaneous_requests_for_bun_install = 48;
-const default_max_simultaneous_requests_for_bun_install_for_proxies = 16;
+const default_max_simultaneous_requests_for_bun_install_for_proxies = 64;
 
 const bun = @import("root").bun;
 const FeatureFlags = bun.FeatureFlags;
@@ -8769,7 +8769,7 @@ pub const PackageManager = struct {
                 break :brk @max(network_concurrency, 1);
             }
 
-            // If any HTTP proxy is set, reduce the default limit to 16 from 48.
+            // If any HTTP proxy is set, use a diferent limit
             if (env.has("http_proxy") or env.has("https_proxy") or env.has("HTTPS_PROXY") or env.has("HTTP_PROXY")) {
                 break :brk default_max_simultaneous_requests_for_bun_install_for_proxies;
             }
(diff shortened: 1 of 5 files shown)


robobun:
@Jarred-Sumner, your commit a0f283b8b7cb8be83d0508885c4a99c1350ca627 is building: [`#5189`](https://buildkite.com/bun/bun/builds/5189)

test/cli/install/bun-install.test.ts
```diff
@@ -67,20 +67,149 @@ afterAll(dummyAfterAll);
...
+});
+
 it("should not error when package.json has comments and trailing commas", async () => {
   const urls: string[] = [];
   setHandler(dummyRegistry(urls));
   await writeFile(
     join(package_dir, "package.json"),
     `
     {
-      // such comment!
```
alex-demchenko, line 77: deleting the comment makes this test meaningless

Jarred-Sumner: Thanks, didn’t mean to commit deleting this comment

src/install/install.zig
```diff
@@ -9290,6 +9326,7 @@ pub const PackageManager = struct {
         clap.parseParam("--backend <STR>                       Platform-specific optimizations for installing dependencies. " ++ platform_specific_backend_label) catch unreachable,
         clap.parseParam("--registry <STR>                      Use a specific registry by default, overriding .npmrc, bunfig.toml and environment variables") catch unreachable,
         clap.parseParam("--concurrent-scripts <NUM>            Maximum number of concurrent jobs for lifecycle scripts (default 5)") catch unreachable,
+        clap.parseParam("--network-concurrency <NUM>           Maximum number of concurrent network requests (default 48)") catch unreachable,
```
Electroid, line 9329: `--concurrent-requests` would go nicely with the existing flags

Jarred-Sumner: `--network-concurrency` is what pnpm does

src/install/install.zig
```diff
@@ -1,3 +1,11 @@
+// Default to a maximum of 64 simultaneous HTTP requests for bun install if no proxy is specified
+// if a proxy IS specified, default to 16
+// https://github.com/npm/cli/issues/7072
+// https://pnpm.io/npmrc#network-concurrency (pnpm defaults to 16)
+// https://yarnpkg.com/configuration/yarnrc#networkConcurrency (defaults to 50)
+const default_max_simultaneous_requests_for_bun_install = 48;
+const default_max_simultaneous_requests_for_bun_install_for_proxies = 64;
```
KieranP, line 7: Comment says 16, code sets to 64, which higher than non-proxy.

src/install/install.zig
```diff
@@ -1,3 +1,11 @@
+// Default to a maximum of 64 simultaneous HTTP requests for bun install if no proxy is specified
```
KieranP, line 1: Comment says 64, code sets 48.

Issue #13799, opened 2024-09-08
\x01 byte written to child process 


mattfysh:
### What version of Bun is running?

1.1.27+267afa293

### What platform is your computer?

Darwin 23.6.0 arm64 arm

### What steps can reproduce the bug?

**index.ts**
```js
import { spawn } from 'bun'

const proc = spawn({
  cmd: ['python3', 'child.py'],
  serialization: 'json',
  stdio: ['inherit', 'inherit', 'inherit'],
  ipc(msg) {},
})

proc.send({ test: true })
```

**child.py**
```py
import os

fd = 3
read_pipe = os.fdopen(fd, "r")
line = read_pipe.readline().strip()
print("recv: ", line.encode())
```

### What is the expected behavior?

```
recv:  b'{"test":true}'
```

### What do you see instead?

_Note the `\x01` byte_

```
recv:  b'\x01{"test":true}'
```

### Additional information

The following code run with node produces the correct result:

```js
import { spawn } from 'node:child_process'

const proc = spawn('python3', ['child.py'], {
  stdio: ['inherit', 'inherit', 'inherit', 'ipc'],
})

proc.send({ test: true })
```

kumanoko24:
Not sure if it is related:

OS:
```
System Version: macOS 14.6.1 (23G93)
Kernel Version: Darwin 23.6.0
```

Bun version: 
`1.1.29`

Files:
`bun ./src/index.ts`:

```typescript
import { resolve } from "node:path";

const nodeSubProcess = Bun.spawn({
  cmd: ["node", "./src/node/hello.js"],
  cwd: resolve(import.meta.dir, ".."),
  ipc(message) {
    console.log("bun", message, "from node");
  },
  stdio: ["inherit", "inherit", "inherit"],
  serialization: "json",
});

setInterval(() => {
  nodeSubProcess.send("123");
}, 2000);

process.on("exit", () => {
  nodeSubProcess.kill();
});
```

`./src/node/hello.js`:

```javascript
process.on("message", (msg) => {
  console.log("i am node:", msg, "from bun");

  process.send("i got " + msg);
});

process.on("uncaughtException", (a, b) => {
  console.error(a, b);
});

// keep it running
setInterval(() => {}, 1000);
```

The output looks like:

```
SyntaxError: Unexpected token '', ""123"" is not valid JSON
    at parse (<anonymous>)
    at parseChannelMessages (node:internal/child_process/serialization:152:15)
[...]

Issue #5190, opened 2023-09-13
Running typescript jest tests with bun


robobun:
Hi there!

Trying to migrate our codebase to bun. Sadly we can't use the built in bun test runner as we need to use mocks, which are currently unsupported. Our current test setup is jest, using the `ts-jest` preset in order to run them

As part of the initial configuration with bun, we updated our tsconfig, from:
```json
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2022",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "allowSyntheticDefaultImports": true,
    "outDir": "dist",
    "moduleResolution": "node",
    "sourceMap": true
  },
  "compileOnSave": true,
  "include": ["src/**/*.ts"],
  "exclude": ["node_modules"],
}
```
to
```json
{
  "compilerOptions": {
    "lib": ["ESNext"],
    "module": "esnext",
    "target": "esnext",
    "moduleResolution": "bundler",
    "moduleDetection": "force",
    "allowImportingTsExtensions": true,
    "noEmit": true,
    "composite": true,
    "strict": true,
    "downlevelIteration": true,
    "skipLibCheck": true,
    "jsx": "preserve",
    "allowSyntheticDefaultImports": true,
    "forceConsistentCasingInFileNames": true,
    "allowJs": true,
    "types": [
      "bun-types" // add Bun global
    ]
  }
}
```

However, with this updated tsconfig, both `bun run test` and `npm run test` now fail, as every import resolves to `undefined`, so code like this in test files fails
```ts
import { jsdom } from 'jsdom'
jsdom.JSDOM
```
as jsdom is undefined

_Originally reported on Discord: [`Running typescript jest tests with bun`](https://discord.com/channels/876711213126520882/1150793867893944330/1150793867893944330)_

Wazbat:
Created example repo:
https://github.com/Wazbat/bun-ts-jest-example

PR #14756, opened 2024-10-23
make sure ipc with json serialization still works when bun is parent and not the child

nektro:
Closes https://github.com/oven-sh/bun/issues/13799

----

before

```
SyntaxError: Unexpected token '', ""I am your father"" is not valid JSON
    at parse (<anonymous>)
    at parseChannelMessages (node:internal/child_process/serialization:152:15)
    at parseChannelMessages.next (<anonymous>)
    at channel.onread (node:internal/child_process:623:18)
```

after

passes


robobun:
@nektro, your commit 6634f590bf2152609808fd6262ac9f3a0ec4aab5 is building: [`#5190`](https://buildkite.com/bun/bun/builds/5190)

src/bun.js/ipc.zig
```diff
@@ -194,21 +194,29 @@ const json = struct {
...
+                '0'...'9', //number
+                't', //true
+                'f', //false
+                'n', //null
+                => {
                     kind = 1;
-                    json_data = data[0..idx];
+                },
+                else => {
+                    std.log.warn("{d} {d} {d}", .{ kind, json_data.len, idx });
```
Jarred-Sumner: We shouldn't use `std.log.warn` in Bun's codebase outside of development

src/bun.js/ipc.zig
```diff
@@ -194,21 +194,29 @@ const json = struct {
...
+                    std.log.warn("{d}", .{json_data});
+                    @panic("ipc: invalid json");
```
Jarred-Sumner: This shouldn't hard crash, it should do what node does (emit an exception event, close the socket, allow the process to exit due to the unhandled exception if necessary)

src/bun.js/ipc.zig
```diff
@@ -230,6 +238,7 @@ const json = struct {
             }
 
             const deserialized = str.toJSByParseJSON(globalThis);
+            // if (deserialized == .zero) return error.JSError;
```
Jarred-Sumner: this line probably shouldn't be commented out because toJSByParseJSON potentially throws a syntax error and it should not crash in that case.

dylan-conway (approved):

Jarred-Sumner (changes requested):
Please add a test that checks node -> bun IPC works as well.

Please also add a test that runs the invalid JSON.parse case matches the behavior of Node when it receives invalid JSON.

src/bun.js/ipc.zig
```diff
@@ -194,21 +194,29 @@ const json = struct {
...
+                2 => {
+                    json_data = data[1..idx];
+                },
+                '{', //object
+                '[', //array
+                '"', //string
+                '0'...'9', //number
+                't', //true
+                'f', //false
+                'n', //null
```
Jarred-Sumner: What is the purpose of checking this? Can we treat all values that are not 2 as kind 1? There are other valid first bytes to JSON, such as `-` and since the branches are no longer any different, we don't need to do this work here anymore

src/bun.js/ipc.zig
```diff
@@ -194,21 +194,19 @@ const json = struct {
...
+    pub fn decodeIPCMessage(data: []const u8, globalThis: *JSC.JSGlobalObject) IPCDecodeError!DecodeIPCMessageResult {
         if (bun.strings.indexOfChar(data, '\n')) |idx| {
             var kind = data[0];
-            var json_data = data[1..idx];
+            var json_data = data[0..idx];
 
             switch (kind) {
-                1, 2 => {},
+                2 => {
+                    json_data = data[1..idx];
```
Jarred-Sumner, line 204: won't this go out of bounds if it only received 1 byte

nektro: no because the only case there can be one byte is if that byte is `\n`

Jarred-Sumner (changes requested):
looks like it's printing garbage to stdout on Windows
image

PR #14798, opened 2024-10-24
ci: merge clang-format and clang-tidy into single pipeline

DonIsaac:
### What does this PR do?

Merges `clang-format` and `clang-tidy` into a single workflow/job called `clang`

### Why?

`clang-format` and `clang-tidy` take roughly a minute each. Most of that time is spent on setup, primarily on LLVM installation. Additionally, both use the same exact environment variables and steps (except for the format vs tidy part). Merging these together will reduce GH actions CI time spent. Not by much mind you, but if this runs on each PR it could be worth it.

### Risks & Alternatives
If we merge this PR, we need to update `main`'s branch protections to use `clang`, then remove `clang-format` and `clang-tidy` as required workflows.

If we like having these separate for viewing/comprehension reasons, we could combine these workflows into one but use separate jobs. This negates CI waste benefits and is more of a style preference.


Electroid:
Thanks for contributing this! Looks good, I’ll update the branch protection settings

Issue #5268, opened 2023-09-13
It would be great if Bun support linting and formatting out of the box.


adeel0x01:
### What is the problem this feature would solve?

Currently, users who work with Bun need to rely on external tools and configurations to ensure code quality and consistency, which can be cumbersome and time-consuming. This feature would streamline the development process and enhance the overall developer experience by integrating linting and formatting directly into Bun.

### What is the feature you are proposing to solve the problem?

The proposed feature is to incorporate built-in linting and formatting support into Bun. Users would have the flexibility to customize linting and formatting rules to suit their specific project requirements.

sirenkovladd:
Duplicate of https://github.com/oven-sh/bun/discussions/712

PR #14804, opened 2024-10-25
bindings: make throwInvalidArgumentTypeValue print the value like the real ERR_INVALID_ARG_TYPE

nektro:
with this patch it'll match the cpp one and the impl used by `internal/errors`


robobun:
@nektro, your commit 66aab38428c828a77bba03f6b5af1be0a6e2bab8 is building: [`#5268`](https://buildkite.com/bun/bun/builds/5268)

Issue #5305, opened 2023-09-13
`fs.readSync` behaves unusually with a running readline interface


jhmaster2000:
### What version of Bun is running?

1.0.1+c3455c0ceee6bbe399781819a42fff6cf24792e2

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

```js
$ bun run file.mjs
import fs from 'fs';
import readline from 'readline';
const rl = readline.createInterface({
        input: process.stdin,
        output: process.stdout,
        terminal: true,
        prompt: '> '
});
console.log(fs.readSync(0, Buffer.alloc(64)));
```

### What is the expected behavior?

In Node, the provided file throws an `EAGAIN` read syscall error from `fs.readSync`, which at least makes sense.

### What do you see instead?

In Bun, `fs.readSync` is somehow able to read stdout in parallel from readline, however it only does a single character and stops, the result of running the provided file and typing "abcdef":
```
$ bun run file.mjs
1
bcdef
```
Notice the missing `a` and the 1 (the return of readSync is number of bytes it read)

### Additional information

`fs.readSync` on stdout behaves normally with no readline interface running.

PR #14828, opened 2024-10-25
bindings: assert JSValue.call is not called on zero value

nektro:
found this while working on another branch but putting this patch in its own in case it happens to breaks anything


robobun:
@nektro, your commit 1033ffde667b33f7b7977993e5fec571f6783f8d is building: [`#5305`](https://buildkite.com/bun/bun/builds/5305)

paperdave:
the assertion already exists on this branch

Issue #5307, opened 2023-09-13
Incorrect lines covered when bun test --coverage


eye-wave:
### What version of Bun is running?

1.0.0

### What platform is your computer?

Linux 6.4.6-76060406-generic x86_64 x86_64

### frequencytable.ts
```ts
export function generateMIDITable(tuning = 440) {
  const notes = ["C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B"]

  const frequencies: number[] = []
  const notations: string[] = []

  return {
    getFreq(note: number) {
      const result = frequencies.at(note)
      if (!result) {
        const frequency = tuning * 2 ** ((note - 69) / 12)
        frequencies[note] = frequency
        return frequency
      }
      return result
    },

    getNote(note: number) {
      const result = notations.at(note)
      if (!result) {
        const noteIndex = note % 12
        const octave = Math.floor(note / 12) - 1
        const notation = notes[noteIndex] + octave

        notations[note] = notation
        return notation
      }
      return result
    },
  }
}

export const midiTable = generateMIDITable()

```
---

### frequencytable.test.ts
```ts
import { generateMIDITable, midiTable } from "./frequencytable"
import { describe, it, expect } from "bun:test"

describe("generateMIDITable", () => {
  it("should generate a valid MIDI table within the specified range", () => {
    expect(midiTable.getFreq(60)).toBeCloseTo(261.63)
    expect(midiTable.getFreq(69)).toBeCloseTo(440)
    expect(midiTable.getFreq(84)).toBeCloseTo(1046.5)

    expect(midiTable.getNote(51)).toBe("D#3")
    expect(midiTable.getNote(21)).toBe("A0")
    expect(midiTable.getNote(5)).toBe("F-1")
  })
})
```

### stdout
`$ bun test --coverage`
```
----------------------------------|---------|---------|-------------------
File                              | % Funcs | % Lines | Uncovered Line #s
----------------------------------|---------|---------|-------------------
All files                         |  100.00 |  100.00 |
 frequencytable.ts                |  100.00 |  100.00 |
----------------------------------|---------|---------|-------------------
```

### stdout
`$ bun test --coverage`
```
----------------------------------|---------|---------|-------------------
File                              | % Funcs | % Lines | Uncovered Line #s
----------------------------------|---------|---------|-------------------
All files                         |  100.00 |   92.00 |
 frequencytable.ts                |  100.00 |   92.00 | 14
[...]

PR #14829, opened 2024-10-26
bun:test: implement `test.failing` and the `retry` option

### LINKED ISSUE #15887: test: support `test.failing`

Add `test.failing` to `bun:test` with the same behavior as [Jest](https://jestjs.io/docs/api#testfailingname-fn-timeout).

```ts
// foo.test.ts
describe("foo", () => {
  function foo() {
    return 2
  }

  test.failing("passes if any expectation fails or an error is thrown", () => {
    expect(foo()).toBe(1);
  })

  // this test fails
  test.failing("fails if the test 'passes': that is, no expectations fail", () => {
    expect(foo()).toBe(2)
  })

  it.failing("is also on 'it'", () => {
    expect(foo()).toBe(1)
  })
})
```

Helps developers keep track of known bugs. This is better than `.todo` since it forces the `.failing` modifier to be removed once the bug is fixed.

This also brings us closer to feature parity with Jest.

_No response_

nektro:
for tests that are flaky instead of broken, using this should be much better than disabling them outright

related: https://github.com/oven-sh/bun/issues/1825
fixes #15887

### DIFF
--- a/test/cli/test/bun-test.test.ts
+++ b/test/cli/test/bun-test.test.ts
@@ -846,6 +846,92 @@ describe("bun test", () => {
     });
     test.todo("check formatting for %p", () => {});
   });
+  describe(".failing", () => {
+    test("expect good fail", () => {
+      const stderr = runTest({
+        args: [],
+        input: [
+          `
+          import { test, expect } from "bun:test";
+          test.failing("the", () => {
+            expect(6).toBe(6);
+          });
+          `,
+        ],
+      });
+      expect(stderr).toContain(` 0 pass\n 1 fail\n 1 expect() calls\nRan 1 tests across 1 files. `);
+    });
+    test("expect bad pass", () => {
+      const stderr = runTest({
+        args: [],
+        input: [
+          `
+          import { test, expect } from "bun:test";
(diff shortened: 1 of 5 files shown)


robobun:
@nektro, your commit 1e682dec6a156360c7e8044f57d60882b5b1fb69 is building: [`#5307`](https://buildkite.com/bun/bun/builds/5307)

test/cli/test/bun-test.test.ts
```diff
@@ -846,6 +846,139 @@ describe("bun test", () => {
...
     test.todo("check formatting for %p", () => {});
   });
+  describe(".failing", () => {
+    test("expect good fail", () => {
+      const stderr = runTest({
+        args: [],
+        input: [
+          `
+          import { test, expect } from "bun:test";
+          test.failing("the", () => {
```
Electroid, line 856: [`test.failing`](https://jestjs.io/docs/api#testfailingname-fn-timeout) is a Jest API, but it does not work like this and it will probably confuse people.

How it works in Jest is that it is expected that the test fails, so the test is actually marked as passed.

Issue #5362, opened 2023-09-14
Cross-runtime compatible shebangs


wintercounter:
### What is the problem this feature would solve?

I am currently developing an executable library that will support `npx`, `bunx`, and global install as well. Currently, bun spins up a Node process by default which will immediately fail if I don't transpile my TS files.

If I use `bunx` or install globally using `bun add -g`, I always want my lib to be executed by bun. For `bunx` there's `--bun` flag, but I don't want my users always need to use this flag.

The outcome I'd like to have:
- `bun link` => `myLib` => executed in bun
- `bunx myLib` => executed in bun, without the `--bun` flag
- `bun add myLib -g` => `myLib` => executed in bun

### What is the feature you are proposing to solve the problem?

I tried setting `bun = true` in `bunfig.toml`, but that's not working. So to overcome the `bunx --bun` problem, it'd be nice if I could configure this for my package and bun would respect this upon execution.

However, the above wouldn't solve the problem with global installation. Is there a possibility for bun to automatically replace `#!/usr/bin/env node` with `#!/usr/bin/env bun` upon installing (and linking) a package globally when install happened using bun?

### What alternatives have you considered?

Separate bin entry points for node/bun, where I provide proper shebang. This would also need support from bun.

typedarray:
+1, running into the same problem.

templarundead:
+100

OleksandrKucherenko:
create file: `node-wrapper.sh` with following code and make it executable

```bash
#!/bin/bash
executors=(ts-node bun yarn yarn)
paramaters=('--' '--' 'ts-node' 'node')

for i in "${!executors[@]}"; do
    if /usr/bin/env "${executors[$i]}" "${paramaters[$i]}" "$(pwd)/$1" 2>/dev/null; then
        exit $?
    else
        if [[ $? -eq 127 ]]; then
            echo "Failed to run ${executors[$i]} ${paramaters[$i]} $1 - error code: tool not found" >&2
        else
            exit $? # Exit with the exit code from the failed command
        fi
    fi
done
# We didn't find any of them.
exit 1
```

now customize the Ts script: 


```ts
#!./node-wrapper.sh

console.log('Hello world')

throw new Error('This is an error')

export {}
```

make typescript file executable, and try to run it. first script will try to find the executable for the script: try ts-node, then BUN, `yarn ts-node`, `yarn node`

PR #14858, opened 2024-10-27
Introduce `Bun.randomUUIDv7`

Jarred-Sumner:
### What does this PR do?

Introduce `Bun.randomUUIDv7`

```ts
import { randomUUIDv7 } from "bun";

// "0192ce11-26d5-7dc3-9305-1426de888c5a"
const first = randomUUIDv7();

// "0192ce11-26d5-7dc4-8894-bc88d506d6ee"
const second = randomUUIDv7();

// Unlike UUID v4, UUID v7 is monotonic.
// This makes it good for sorting and databases.
const isMonotonic = second > first;
// > true

// Sorting it produces the same order
const ids = Array.from({ length: 100 }, () => randomUUIDv7());
if (!Bun.deepEquals(ids.toSorted(), ids)) {
  throw new Error("expected sorted to be equal to ids");
}

console.log(first, second, isMonotonic);
```

Tests

### DIFF
--- a/docs/api/utils.md
+++ b/docs/api/utils.md
@@ -106,6 +106,57 @@ const ls = Bun.which("ls", {
 console.log(ls); // null
 ```
 
+You can think of this as a builtin alternative to the [`which`](https://www.npmjs.com/package/which) npm package.
+
+## `Bun.randomUUIDv7()`
+
+`Bun.randomUUIDv7()` returns a [UUID v7](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-01.html#name-uuidv7-layout-and-bit-order), which is monotonic and suitable for sorting and databases.
+
+```ts
+import { randomUUIDv7 } from "bun";
+
+const id = randomUUIDv7();
+// => "0192ce11-26d5-7dc3-9305-1426de888c5a"
+```
+
+A UUID v7 is a 128-bit value that encodes the current timestamp, a random value, and a counter. The timestamp is encoded using the lowest 48 bits, and the random value and counter are encoded using the remaining bits.
+
+The `timestamp` parameter defaults to the current time in milliseconds. When the timestamp changes, the counter is reset to a psuedo-random integer wrapped to 4096. This counter is atomic and threadsafe, meaning that using `Bun.randomUUIDv7()` in many Workers within the same process running at the same timestamp will not have colliding counter values.
+
+The final 8 bytes of the UUID are a cryptographically secure random value. It uses the same random number generator used by `crypto.randomUUID()` (which comes from BoringSSL, which in turn comes from the platform-specific system random number generator usually provided by the underlying hardware).
+
+```ts
+namespace Bun {
+  function randomUUIDv7(
+    encoding?: "hex" | "base64" | "base64url" = "hex",
+    timestamp?: number = Date.now(),
+  ): string;
+  /**
+   * If you pass "buffer", you get a 16-byte buffer instead of a string.
+   */
+  function randomUUIDv7(
+    encoding: "buffer",
+    timestamp?: number = Date.now(),
+  ): Buffer;
+
+  // If you only pass a timestamp, you get a hex string
+  function randomUUIDv7(timestamp?: number = Date.now()): string;
+}
+```
+
+You can optionally set encoding to `"buffer"` to get a 16-byte buffer instead of a string. This can sometimes avoid string conversion overhead.
+
+```ts#buffer.ts
+const buffer = Bun.randomUUIDv7("buffer");
+```
+
+`base64` and `base64url` encodings are also supported when you want a slightly shorter string.
+
+```ts#base64.ts
+const base64 = Bun.randomUUIDv7("base64");
+const base64url = Bun.randomUUIDv7("base64url");
+```
+
 ## `Bun.peek()`
 
 `Bun.peek(prom: Promise)`
--- a/src/bun.js/uuid.zig
+++ b/src/bun.js/uuid.zig
@@ -160,18 +160,18 @@ pub const UUID7 = struct {
         var bytes: [16]u8 = undefined;
 
         // First 6 bytes: timestamp in big-endian
-        bytes[0] = @as(u8, @truncate((timestamp >> 40) & 0xFF));
-        bytes[1] = @as(u8, @truncate((timestamp >> 32) & 0xFF));
-        bytes[2] = @as(u8, @truncate((timestamp >> 24) & 0xFF));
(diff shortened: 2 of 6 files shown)


robobun:
@Jarred-Sumner, your commit 4ce4e6d08bc6749c536708a3789a65c62cb20fdb is building: [`#5362`](https://buildkite.com/bun/bun/builds/5362)

docs/api/utils.md
```diff
@@ -106,6 +106,57 @@ const ls = Bun.which("ls", {
...
+`Bun.randomUUIDv7()` returns a [UUID v7](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-01.html#name-uuidv7-layout-and-bit-order), which is monotonic and suitable for sorting and databases.
+
+```ts
+import { randomUUIDv7 } from "bun";
+
+const id = randomUUIDv7();
+// => "0192ce11-26d5-7dc3-9305-1426de888c5a"
+```
+
+A UUID v7 is a 128-bit value that encodes the current timestamp, a random value, and a counter. The timestamp is encoded using the lowest 48 bits, and the random value and counter are encoded using the remaining bits.
```
dylan-conway, line 122: isn't it the upper 48 bits?

dylan-conway: oh, nvm

docs/api/utils.md
```diff
@@ -106,6 +106,57 @@ const ls = Bun.which("ls", {
...
+
```
dylan-conway, line 123: maybe also add something like: the counter ensures UUIDs generated with the same timestamp have the expected sequencing

src/bun.js/uuid.zig
```diff
@@ -129,3 +134,76 @@ pub const zero: UUID = .{ .bytes = .{0} ** 16 };
...
+
+    var uuid_v7_lock = bun.Lock{};
+    var uuid_v7_last_timestamp: std.atomic.Value(u64) = .{ .raw = 0 };
+    var uuid_v7_counter: std.atomic.Value(u32) = .{ .raw = 0 };
+
+    fn getCount(timestamp: u64) u32 {
+        uuid_v7_lock.lock();
+        defer uuid_v7_lock.unlock();
+        if (uuid_v7_last_timestamp.swap(timestamp, .monotonic) != timestamp) {
+            uuid_v7_counter.store(@intCast(bun.fastRandom() % std.math.maxInt(u32)), .monotonic);
```
dylan-conway: this might be a problem. the spec says it MUST reset to 0. we should look at other implementations

src/bun.js/webcore.zig
```diff
@@ -664,6 +665,59 @@ pub const Crypto = struct {
...
+            }
+
+            break :brk JSC.Node.Encoding.hex;
+        };
+
+        const timestamp: u64 = brk: {
+            const timestamp_value: JSC.JSValue = if (encoding_value != .undefined and arguments.len > 1) arguments[1] else if (arguments.len == 1 and encoding_value == .undefined) arguments[0] else .undefined;
+
+            if (timestamp_value != .undefined) {
+                if (globalThis.validateIntegerRange(timestamp_value, i64, 0, .{
```
dylan-conway, line 706: if we're going to accept a timestamp arg, I think it would be reasonable to also accept `Date` objects

src/bun.js/uuid.zig
```diff
@@ -129,3 +134,76 @@ pub const zero: UUID = .{ .bytes = .{0} ** 16 };
...
+        }
+
+        return uuid_v7_counter.fetchAdd(1, .monotonic) % 4096;
```
dylan-conway, line 154: Would it be possible to make `4096` configurable? Thinking about code like:
```js
const specificDate = Date.now();
for (let i = 0; i < 10_000; i ++) {
    Bun.randomUUIDv7(specificDate);
}
```

src/bun.js/uuid.zig
```diff
@@ -129,3 +134,76 @@ pub const zero: UUID = .{ .bytes = .{0} ** 16 };
...
+        bytes[5] = @as(u8, @truncate(timestamp & 0xFF));
+
+        // Byte 6: Version 7 in high nibble, top 4 bits of counter in low nibble
+        bytes[6] = (@as(u8, 7) << 4) | @as(u8, @truncate((count >> 8) & 0x0F));
+
+        // Byte 7: Lower 8 bits of counter
+        bytes[7] = @as(u8, @truncate(count & 0xFF));
+
+        // Byte 8: Variant in top 2 bits, 6 bits of random
+        bytes[8] = (0x80 & 0xFF) | (random[0] & 0x3F);
```
dylan-conway:
```suggestion
        bytes[8] = 0x80 | (random[0] & 0x3F);
```

src/bun.js/uuid.zig
```diff
@@ -129,3 +134,76 @@ pub const zero: UUID = .{ .bytes = .{0} ** 16 };
...
+        bytes[2] = @as(u8, @truncate((timestamp >> 24) & 0xFF));
+        bytes[3] = @as(u8, @truncate((timestamp >> 16) & 0xFF));
+        bytes[4] = @as(u8, @truncate((timestamp >> 8) & 0xFF));
+        bytes[5] = @as(u8, @truncate(timestamp & 0xFF));
+
+        // Byte 6: Version 7 in high nibble, top 4 bits of counter in low nibble
+        bytes[6] = (@as(u8, 7) << 4) | @as(u8, @truncate((count >> 8) & 0x0F));
+
+        // Byte 7: Lower 8 bits of counter
+        bytes[7] = @as(u8, @truncate(count & 0xFF));
```
dylan-conway:
```suggestion
        bytes[0] = @truncate(timestamp >> 40);
        bytes[1] = @truncate(timestamp >> 32);
        bytes[2] = @truncate(timestamp >> 24);
        bytes[3] = @truncate(timestamp >> 16);
        bytes[4] = @truncate(timestamp >> 8);
        bytes[5] = @truncate(timestamp);

        // Byte 6: Version 7 in high nibble, top 4 bits of counter in low nibble
        bytes[6] = (@as(u8, 7) << 4) | @as(u8, @truncate((count >> 8) & 0x0F));

        // Byte 7: Lower 8 bits of counter
        bytes[7] = @truncate(count);
```

src/bun.js/uuid.zig
Jarred-Sumner: We can reset it to 0

src/bun.js/webcore.zig
Jarred-Sumner: A good future TODO

src/bun.js/uuid.zig
Jarred-Sumner: No, it's the number of bits

Issue #14870, opened 2024-10-28
bun remove non-existent-package doesn't throw any error


vijaybritto:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Run `bun remove any-invalid-package` in any workspace to see that it says `done` without mentioning that the package is not found in package.json.

This suppresses typo errors in the package names and we dont realize that we didnt remove the package.

### What is the expected behavior?

It should throw an error similar to yarn or other package managers. For ex, in yarn v1 we get this:
![Image](https://github.com/user-attachments/assets/c6a75fd2-3cd0-485f-a559-8d9e7507e6ef)

### What do you see instead?

![Image](https://github.com/user-attachments/assets/123873f0-5ebb-4845-b3ad-72ccb6b029a7)
No errors

kaioduarte:
It used to behave like that but it was changed a while ago (see https://github.com/oven-sh/bun/issues/8032) to be closer to what npm does.

Issue #5383, opened 2023-09-14
node:child_process exec returns void in err, stdout,stderr


kerambit:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Make file test.js
Insert code below

```javascript
import {exec} from "child_process";
exec("ls", (err, stdout, stderr) => {
  console.log(`stdout: ${stdout}`);
  console.error(`stderr: ${stderr}`);
  console.log(`err: ${err}`);
});
```

Run
`$ bun test.js`

### What is the expected behavior?

Returns 

```
stdout: package.json
test.js

stderr: 
err: null
```

### What do you see instead?

Returns void

### Additional information

I have tried same cmd with spawn - it works fine 

```javascript
import {spawn} from "child_process";

const ls = spawn('ls');

ls.stdout.on('data', (data) => {
  console.log(`stdout: ${data}`);
});

ls.stderr.on('data', (data) => {
  console.error(`stderr: ${data}`);
});

ls.on('close', (code) => {
  console.log(`child process exited with code ${code}`);
}); 
```

Also looks like exec cmd under the hood works fine, but returns void in stdout/stderr. For example you could use `exec('mkdir hello')` and directory creating.

kerambit:
I've checked #5324 but looks like my cmd could't exceed max buffer size

Hanaasagi:
bun 1.0.3 fixes the issue in https://github.com/oven-sh/bun/pull/5481. Please upgrade.

kerambit:
Great! Works fine. 
Closing the issue as fixed

PR #14882, opened 2024-10-28
pm: error when removing package that does not exist

nektro:
Closes https://github.com/oven-sh/bun/issues/14870

before

image

after

image


robobun:
@nektro, your commit d6282ce5c6dc4ad0d34165a26434d0eb9e23cce3 is building: [`#5383`](https://buildkite.com/bun/bun/builds/5383)

test/cli/install/bun-remove.test.ts
```diff
@@ -191,7 +191,7 @@ it("should not reject missing package", async () => {
     stderr: "pipe",
     env,
   });
-  expect(await rmExited).toBe(0);
+  expect(await rmExited).toBe(1);
```
Electroid, line 194: Does `npm`/`pnpm`/`yarn` error as well? Why did this test exist in the first place

nektro: `pnpm`/`yarn` do, `npm` does not
if users want the `npm` behavior they can do `bun remove <pkg> || true`

Issue #5388, opened 2023-09-14
bun fails at module resolution with bun workspaces


nikivdev:
### What version of Bun is running?

1.0.1+92e95c86dd100f167fb4cf8da1db202b5211d2c1

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

On this repo:

https://github.com/learn-anything/learn-anything.xyz

I tried to migrate from pnpm to bun. Everything worked perfectly with pnpm before.

After [replacing all pnpm things with bun](https://github.com/learn-anything/learn-anything.xyz/commit/732aa2270a2a40b5afcff49795ae55681c02e7f7).

I get this error trying to run `bun web:dev`

<img width="760" alt="image" src="https://github.com/oven-sh/bun/assets/6391776/be6d2d1a-37f8-46fc-83c2-88215a966a8a">

Which is a bug because its there in the package.json dep

<img width="615" alt="image" src="https://github.com/oven-sh/bun/assets/6391776/fa9b8405-e176-42f4-9454-9b052610c0ac">

And I ran `bun i` at root and have these workspaces setup:

https://github.com/learn-anything/learn-anything.xyz/blob/main/package.json#L81

What am I doing wrong?

Would be grateful for helping me migrate away from pnpm to bun now that it reached 1.0

### What is the expected behavior?

Everything works. I can run my scripts without failure.

### What do you see instead?

```
  VITE v4.4.9  ready in 730 ms

  ➜  Local:   http://localhost:3000/
  ➜  Network: use --host to expose
  ➜  Inspect: http://localhost:3000/__inspect/
  ➜  press h to show help

  ➜  Page Routes:
     ┌─ http://localhost:3000/auth
     ├─ http://localhost:3000/
     ├─ http://localhost:3000/pricing
     ├─ http://localhost:3000/@(username)/
     ├─ http://localhost:3000/@(username)/
     ├─ http://localhost:3000/:topic/edit
     └─ http://localhost:3000/:topic/

  ➜  API Routes:
     None! 👻

🔥 GET http://localhost:3000/3d-printing/edit
Failed to resolve import "graphql-mobius" from "src/root.tsx". Does the file exist?
🔥 GET http://localhost:3000/3d-printing/edit
6:01:15 PM [vite] Internal server error: Failed to resolve import "graphql-mobius" from "src/root.tsx". Does the file exist?
  Plugin: vite:import-analysis
[...]

PR #14884, opened 2024-10-29
Fix #14187

### LINKED ISSUE #14187: Memory leak in `on` method from `node:events`

1.1.29+6d43b3662

Darwin 23.4.0 arm64 arm

- Repro: https://github.com/sebastiansandqvist/trpc-ee-repro
- (Setup instructions in the readme)

The `node:events` `on` function takes a `{ signal }` [option](https://nodejs.org/api/events.html#eventsonemitter-eventname-options) which is an `AbortSignal` that should cause the the event listener to deregister. 

### What do you see instead?

When using the `signal` from a request sent over SSE, the event listener is not canceled even after the signal should fire.

### Additional information

https://github.com/trpc/trpc/issues/6045

I've also tried adding the following code to the subscription, which never fires, so perhaps there's something wrong with the request's `AbortController`:

```ts
addAbortListener(signal, () => {
  console.log('aborting');
});
```

pfgithub:
### What does this PR do?

Fixes #14187

- Replace bun's `on` implementation with node's implementation
- Add node's tests for `once` and `on`
- Add listenerCount() support for EventTargets. Node does not support this.
- Replaces `eventTargetAgnosticAddListener` and `eventTargetAgnosticRemoveListener` with node's implementation that throws the right errors.
- Update validators.ts's implementation of validateObject to throw on null, like node's does when no options parameter is provided.
- Removes fixed queue test because it tests nothing.

- [x] Code changes


robobun:
@pfgithub, your commit 8437baf4282d7c5a5bc1ac6ff62ca61fd8801370 is building: [`#5388`](https://buildkite.com/bun/bun/builds/5388)

PR #14895, opened 2024-10-29
Fix additional arguments when running a package.json script

### LINKED ISSUE #14872: [BUG]: Failed to run script: Unexpected token: `(` (when running "bun run <cmd>")

1.1.33+247456b67

1. `bun init`
2. Any JavaScript/TypeScript will create the error ; Create a file and set anything, the contents are not relevant here.
3. In `package.json`, add a script to run the file, like `scratch: bun run scratch.mjs`
4. run `bun run scratch "(hello)"

No error, scripts run as expected.

### What do you see instead?

```
error: Failed to run script scratch due to error Unexpected token: `(`
```

### Additional information

/

pfgithub:
### What does this PR do?

Fixes #14872 

- [x] Code changes

`bun run script arguments` previously copied arguments directly into a command for `/bin/sh` to run, without any escaping. This meant with a script `"dev": "echo"`, running `bun run dev "\$HOME"` would print your home directory, rather than the literal text `$HOME`. Fixes regular and filtered script running to escape arguments.

This also removes `escapeWTF8` because it seemingly never worked? It would re-encode `\$` as `\�` and probably would have converted "æ" to "Ã¦". This also updates shellEscape to use the previously unused `needsEscapeBunstr` / `escapeBunStr` functions rather than calling needsEscapeUTF16/needsEscapeUtf8AsciiLatin1 directly.

  - It's not freed and it wasn't before either


robobun:
@pfgithub, your commit 11e70d8ba36d7f59dbdac2054bd261e6811f74ae is building: [`#5412`](https://buildkite.com/bun/bun/builds/5412)

zackradisic (approved):
Lgtm

Issue #5418, opened 2023-09-14
SvelteKit build with static adapter and `--bun` producing a TypeError


DeedleFake:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Linux 6.5.3-1-MANJARO x86_64 unknown

### What steps can reproduce the bug?

1. Run `bun create svelte@latest test` and choose skeleton project and no TypeScript.
2. Run
```bash
$ cd test
$ bun add -D @sveltejs/adapter-static
```
3. Replace `svelte.config.js`'s contents with
```javascript
import adapter from '@sveltejs/adapter-static';
export default {
	kit: {
		adapter: adapter({fallback: 'index.html'})
	}
};
```
4. Run `bunx --bun vite build`.

### What is the expected behavior?

It should compile correctly.

### What do you see instead?

It _looks_ like it compiles fine, and then at the very end produces

```
error: 30 |  * - `Request`
31 |  * - `Response`
32 |  */
33 | export function installPolyfills() {
34 | 	for (const name in globals) {
35 | 		Object.defineProperty(globalThis, name, {
     ^
TypeError: Attempting to change configurable attribute of unconfigurable property.
      at installPolyfills (/tmp/tmp.wiU9RmCWzZ/test/node_modules/@sveltejs/kit/src/exports/node/polyfills.js:35:2)
      at /tmp/tmp.wiU9RmCWzZ/test/node_modules/@sveltejs/kit/src/core/postbuild/fallback.js:20:1
```

before just hanging until I hit `^C`.

### Additional information

Running `bunx vite build` _without_ the `--bun` flag works correctly. Also, trying to run `bun run build` works fine but once the `build` directory exists, `bun --bun run build` fails because it tries to run the directory instead of the script. That's a different bug, though, I suppose.

The `package.json` that I wound up with while putting together a minimal reproduction is

```json
{
  "name": "test",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "dev": "vite dev",
    "build": "vite build",
    "preview": "vite preview",
    "lint": "prettier --plugin-search-dir . --check .",
    "format": "prettier --plugin-search-dir . --write ."
  },
  "devDependencies": {
    "@sveltejs/adapter-auto": "^1.0.0",
    "@sveltejs/adapter-static": "^2.0.3",
    "@sveltejs/kit": "^1.0.0",
    "prettier": "^2.8.0",
[...]

PR #14897, opened 2024-10-29
Clean up some code in node validators

Jarred-Sumner:
### What does this PR do?

Clean up some code in node validators

### DIFF
--- a/src/bun.js/bindings/NodeValidator.cpp
+++ b/src/bun.js/bindings/NodeValidator.cpp
@@ -12,6 +12,7 @@
 #include <cmath>
 #include <limits>
 
+#include "JSAbortSignal.h"
 #include "JSBufferEncodingType.h"
 #include "BunProcess.h"
 #include "ErrorCode.h"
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit 7bd837dc057eb3ee9672728a116ec7bf711c7334 is building: [`#5418`](https://buildkite.com/bun/bun/builds/5418)

nektro (approved):

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -378,19 +391,32 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_validateEncoding, (JSC::JSGlobalObject * glo
...
+    if (normalized == BufferEncodingType::hex) {
+        auto data = callFrame->argument(0);
+
+        size_t length = 0;
+        if (data.isString()) {
+            length = data.toString(globalObject)->length();
+        } else if (auto* view = JSC::jsDynamicCast<JSC::JSArrayBufferView*>(data)) {
+            length = view->length();
+        } else if (auto* buffer = JSC::jsDynamicCast<JSC::JSArrayBuffer*>(data)) {
+            if (auto* impl = buffer->impl()) {
```
nektro, line 406: when would the first condition be true but this one be false?

Jarred-Sumner: `ArrayBuffer` != `ArrayBufferView`, `Uint8Array` would fit `ArrayBufferView` but not `ArrayBuffer`.

nektro: yes, i meant specifically `if (auto* buffer =` and `if (auto* impl =`

Issue #5424, opened 2023-09-15
OS Module information functions do not work with template strings


ArkinSolomon:
### What version of Bun is running?

1.0.1+31aec4ebe325982fc0ef27498984b0ad9969162b

### What platform is your computer?

Darwin 23.0.0 x86_64 i386

### What steps can reproduce the bug?

In Node.Js, even though `os.platform` (and other functions like it) are technically functions [according to the docs](https://nodejs.org/api/os.html#osplatform), when using template strings, `os.platform` is converted to a string, which is [intended behavior](https://github.com/nodejs/node/blob/dac5f296914503629097b283cdced2df85878833/test/parallel/test-os.js#L239). 

Running the following code in Bun:

```js
const os = require('os')

console.log(typeof os.platform)
console.log(os.platform())
console.log(`${os.platform}`)
```

Results in: 

```
function
darwin
function platform() {
    [native code]
}
```

### What is the expected behavior?

In Node.Js, running the same code results in:

```
function
darwin
darwin
```

### Additional information

In Node.Js

```js
console.log(os.platform[Symbol.toPrimitive]()) // => 'darwin'
```

But in Bun, the code produces the following output:

```
TypeError: os.platform[Symbol.toPrimitive] is not a function. (In 'os.platform[Symbol.toPrimitive]()', 'os.platform[Symbol.toPrimitive]' is undefined)
      at /path/to/test.js:2:12
      at globalThis (/path/to/test.js:2:45)
```

RiskyMH:
We've invested a ton of time into making Bun more reliable over the last year. I'm having a really hard time reproducing this. If you're still running into this issue, please let me know and I'll re-open this

PR #14899, opened 2024-10-29
fix(install): check cached `package.json`s

dylan-conway:
### What does this PR do?
Two things:
- Instead of checking for a cached package directory, bun will now check for the existence of `package.json` in cache. This solves confusing situations where bun could install a package without it's package.json, and reinstall it each time.
- Invalidates cached manifests if they contain 0 versions. It's not possible for a published package to have 0 versions. I don't have a repro for the cause of this bug.

Manually and added a test


robobun:
@dylan-conway, your commit 129f93f3c1c9984354552baee3e4a39437c9a1d1 is building: [`#5424`](https://buildkite.com/bun/bun/builds/5424)

Issue #2865, opened 2023-05-12
Bundling `target=node` should define __dirname, __filename, import.meta.dir, etc


paperclover:
### What is the problem this feature would solve?

it solves annoyances 

### What is the feature you are proposing to solve the problem?

when we compile to target node, we should replace these statements with things like:

```ts
import { dirname, basename } from "path"
import { pathToFileURL } from "url"

const import_meta_path = pathToFileURL(import.meta.url);
const import_meta_dir = dirname(import_meta_path);
const import_meta_file = basename(import_meta_path);

const __filename = import_meta_path
const __dirname = import_meta_dir
```

when format=cjs is a thing this transform would look like

```ts
import { basename } from "path"

const import_meta_path = __filename;
const import_meta_dir = __dirname
const import_meta_file = basename(__filename)
```

PR #14917, opened 2024-10-30
Draft: Fix 12509: Match webpack/babel - evaluate dirname/filename at runtime when doing a build

### LINKED ISSUE #4216: Bundling with --target=bun changes __dirname references to hardcoded values

0.7.3

Bundle any Node.js code with `__dirname` usage, using `bun build example.js --target=bun --outfile=bundle.mjs`

example.js:
```
console.log(__dirname)
```

The bundled code would work from any runtime location

### What do you see instead?

__dirname references are hardcoded to where bundling occurred...

```
var __dirname = "/tmp/example";
console.log(__dirname);
```

### Additional information

I believe __dirname references should be transpiled to import.meta.dir.

Actually, I've learned Bun seems to support __dirname directly in ESM too, so it could simply be left alone?

Somewhat related to #2865

### LINKED ISSUE #12509: bun build inlines __dirname in --target bun and node

1.1.18+5a0b93523

Linux 6.5.0-18-generic x86_64 x86_64

```js
// file.js

console.log(__dirname);
```
bun build file.js --target bun

for `__dirname` to be inlined only in `--target browser`

### What do you see instead?

```js
// @bun
// file.js
var __dirname = "/home/hussein/Desktop/bun-bug";
console.log(__dirname);
```

### Additional information

_No response_

### LINKED ISSUE #13921: `__dirname` is hardcoded with local path value in bundled code when source is `CommonJS`

1.1.27+267afa293

Darwin 23.5.0 arm64 arm

# Problem

The way `Bun` convert `__dirname` for CommonJS source is not portable because the value is hardcoded in output file.

## CommonJS Behavior

**Source file**
```js
const { readFile } = require('fs').promises;
const { resolve } = require('path');

async function main() {
  const [template] = await Promise.all([readFile(resolve(__dirname, './templates/template.hbs'), 'utf-8')]);

  return { template };
}

main().then(console.log).catch(console.error);
```

**Output file**
```js
import {createRequire} from "node:module";
var __require = /* @__PURE__ */ createRequire(import.meta.url);

// src/index.ts
async function main() {
  const [template] = await Promise.all([readFile(resolve(__dirname, "./templates/template.hbs"), "utf-8")]);
  return { template };
}
var __dirname = "/Users/gregory.houllier/Workspace/bun-app/src";
var { readFile } = __require("fs").promises;
var { resolve } = __require("path");
main().then(console.log).catch(console.error);
```

## ESM Behavior

**Source file**
```js
import { readFile } from 'fs/promises';
import { resolve } from 'path';
import { fileURLToPath } from 'url';

const dirname = fileURLToPath(new URL('.', import.meta.url));

async function main() {
  const [template] = await Promise.all([readFile(resolve(dirname, './templates/template.hbs'), 'utf-8')]);

  return { template };
}

main().then(console.log).catch(console.error);
```

**Output file**
```js
// src/index.ts
import {readFile} from "fs/promises";
import {resolve} from "path";
import {fileURLToPath} from "url";
async function main() {
  const [template] = await Promise.all([readFile(resolve(dirname, "./templates/template.hbs"), "utf-8")]);
  return { template };
}
var dirname = fileURLToPath(new URL(".", import.meta.url));
main().then(console.log).catch(console.error);
```

`__dirname` and `fileURLToPath(new URL('.', import.meta.url))` should have the same behavior.

### What do you see instead?

`__dirname` is interpotaled by the path of the user who run the build command.

### Additional information

_No response_

EliLichtblau:
### What does this PR do?

Fixes #12509
Fixes #13921
Fixes #4216

This PR tries to correctly evaluate ```__dirname``` and ```__filename``` whenever possible.

Rather than almost always inlining them to the current source path. This PR changes the behavior to only do that on "run" not build since it doesn't match the behavior of webpack or babel and was bugged in the above issue.

Webpack:
	target browser: inline __dirname
	target esm: convert to esm compat

In webpack you can control the values or preserve them but this is the default.

Babel: 
	always preserve no matter the module target
	

I wrote test cases

- [ X ] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be
- [ X ] I included a test for the new code, or an existing test covers it
- [ X ] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test-file-name.test`)

### DIFF
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -14,6 +14,7 @@ const _runtime = @import("./runtime.zig");
 pub const RuntimeImports = _runtime.Runtime.Imports;
 pub const RuntimeFeatures = _runtime.Runtime.Features;
 pub const RuntimeNames = _runtime.Runtime.Names;
+const Target = options.Target;
 pub const fs = @import("./fs.zig");
 const bun = @import("root").bun;
 const string = bun.string;
@@ -2873,7 +2874,7 @@ pub const Parser = struct {
         suppress_warnings_about_weird_code: bool = true,
         filepath_hash_for_hmr: u32 = 0,
         features: RuntimeFeatures = .{},
-
+        target: Target = Target.browser,
         tree_shaking: bool = false,
         bundle: bool = false,
         package_version: string = "",
@@ -3507,7 +3508,7 @@ pub const Parser = struct {
         //    var __dirname = "foo/bar"
         //    var __filename = "foo/bar/baz.js"
         //
-        if (p.options.bundle or !p.options.features.commonjs_at_runtime) {
+        if (p.options.target == Target.browse) {
             if (uses_dirname or uses_filename) {
                 const count = @as(usize, @intFromBool(uses_dirname)) + @as(usize, @intFromBool(uses_filename));
                 var declared_symbols = DeclaredSymbol.List.initCapacity(p.allocator, count) catch unreachable;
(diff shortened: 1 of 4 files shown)


paperdave:
updated the pr description to link to more issues it would close. see also #2865

test/bundler/dirname-inline.test.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it, beforeAll } from "bun:test";
```
paperclover, line 1: instead of manually testing, can you move these to using the `itBundled` function. you can use `onAfterBundle` to use `expect` on the output code.

test/bundler/dirname-inline.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  it("target browser", async ()=>{
+    const out = Bun.spawnSync({
+      cmd: [bunExe(), "build", "--target=browser", join(temp_dir, "index.ts")],
+      cwd: import.meta.dir,
+      env: bunEnv,
+      stderr: "inherit",
+      stdin: Bun.file("/dev/null"),
+    });
+    const text = out.stdout.toString()
+    expect(text.includes("__dirname =")).toBe(true)
```
paperclover, line 50: we should attempt to ensure that the value assigned to `__dirname` is the correct value.

src/js_parser.zig
```diff
@@ -3550,6 +3551,53 @@ pub const Parser = struct {
                 uses_dirname = false;
                 uses_filename = false;
             }
+        } else if (p.options.output_format == options.Format.esm) {
+            if (uses_dirname or uses_filename) {
+                const count = @as(usize, @intFromBool(uses_dirname)) + @as(usize, @intFromBool(uses_filename));
+                var declared_symbols = DeclaredSymbol.List.initCapacity(p.allocator, count) catch unreachable;
```
paperclover, line 3557: tiny nit, can all newly added `catch unreachable` on memory allocation failures be rewritten to `catch bun.outOfMemory()`

src/js_parser.zig
```diff
@@ -3507,7 +3508,7 @@ pub const Parser = struct {
         //    var __dirname = "foo/bar"
         //    var __filename = "foo/bar/baz.js"
         //
-        if (p.options.bundle or !p.options.features.commonjs_at_runtime) {
+        if (p.options.target == Target.browser or !p.options.transform_only) {
```
paperclover, line 3511: why is target being passed for this? was the original version of this incorrect? i think the main issue is lowering `__dirname` to a string when bundling.

edit: oh, i didnt realize there was more string emitting code (i saw it deleted). i still find this code extremely odd because it emits absolute paths from the builder, which surely dont have any use in a browser.

paperclover (changes requested):
nice! thanks for working on this.

in addition to this, can you add tests for when --format=cjs is used?

src/js_parser.zig
EliLichtblau: Maybe I should look more closely to when commonjs_at_runtime is true. To exactly match current tooling it seemed best to do this only when target was browser. Also I think in node later I use .dirname while in bun it was set to use "dir". 

So I think the cases are 

* Bun run ... - always inline
* Bun build --target="node" -> import.meta.dirname / path.basename(pathToFileURL(import.meta.url))
* Bun build --target="bun" -> import.meta.dir
* Bun build --target="browser" -> souce_path
* Bun build format=cjs -> do nothing

PR #14927, opened 2024-10-31
ci: Reuse build artifacts from previous build if only tests change

Electroid:
### What does this PR do?

When the CI pipeline is being generated, it will detect if only tests were changed. It will then look for previous builds from that branch. Instead of re-building Bun, it will downloads the artifacts from the previous build and run the tests.

### DIFF
--- a/.buildkite/ci.mjs
+++ b/.buildkite/ci.mjs
@@ -73,6 +73,52 @@ async function getChangedFiles() {
   }
 }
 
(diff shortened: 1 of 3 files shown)


robobun:
@Electroid, your commit 8f74a67794529948cea4e819ead4187a648d464a is building: [`#5496`](https://buildkite.com/bun/bun/builds/5496)

PR #14915, opened 2024-10-30
Implement NAPI type tagging

190n:
### What does this PR do?

Copies only the type-tagging parts of #14501 into an independent PR

### How did you verify your code works?

- [x] I included a test for the new code, or existing tests cover it
- [x] I ran my tests locally and they pass (`bun-debug test test-file-name.test`)

Issue #5544, opened 2023-09-16
Repeated: "stdin: Input/output error"


herbievine:
### What version of Bun is running?

1.0.0

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

bun run ed

### What is the expected behavior?

error: missing script "ed"

### What do you see instead?

Like a read on std 0, the terminal waits for input. When CMD+C, this happens:

```
stdin: Input/output error
stdin: Input/output error
stdin: Input/output error
stdin: Input/output error
stdin: Input/output error
```

Electroid:
Are you running Bun in any special way? What shell are you using?

PR #14935, opened 2024-11-01
Fix napi property methods on non-objects

190n:
### What does this PR do?

All Node-API functions that access properties would previously return an error if the parameter was not an object. Now, they convert the parameter to an object first. This matches Node.js's behavior, and enables stuff like [getting the `slice` method on a string](
https://github.com/tree-sitter/node-tree-sitter/blob/9743116fd3368d8130778f555a9763ed6f362325/src/parser.cc#L131-L133). This and #14915 make tree-sitter work.

Tested along with the existing Node-API tests. I tested this works on numbers and strings.


robobun:
@190n, your commit 84cbb8de013ebeba6f4d49048f0759aee580bca3 is building: [`#5544`](https://buildkite.com/bun/bun/builds/5544)

Issue #5545, opened 2023-09-16
`Failed to find package "@oven/bun-linux-x64"` error on `npm i -g bun` docker step


taranek:
### What version of Bun is running?

1.0.1

### What platform is your computer?

node:lts-alpine - docker image

### What steps can reproduce the bug?

Dockerfile to reproduce the issue
```
FROM node:lts-alpine as build-stage
RUN npm install -g bun
```
`npm install -g bun` fails with an error:
```
+ npm install -g bun
npm ERR! code 1
npm ERR! path /usr/local/lib/node_modules/bun
npm ERR! command failed
npm ERR! command sh -c node install.js
npm ERR! Failed to find package "@oven/bun-linux-x64". You may have used the "--no-optional" flag when running "npm install".
npm ERR! Failed to find package "@oven/bun-linux-x64-baseline". You may have used the "--no-optional" flag when running "npm install".
npm ERR! Error: Failed to install package "bun"
npm ERR!     at /usr/local/lib/node_modules/bun/install.js:311:11
npm ERR!     at Generator.throw (<anonymous>)
npm ERR!     at rejected (/usr/local/lib/node_modules/bun/install.js:35:27)
npm ERR!     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### What is the expected behavior?

bun package is installed successfully

### What do you see instead?

Docker build step fails with an installation error

BOTKooper:
Can confirm, same for 1.0.0 and 1.0.2 (tried both basing from `node:18-alpine`), `*-slim` images work as expected

Electroid:
I wonder if this is because alpine node does not install optional dependencies by default? (I don't know this to be true, but that's my guess)

BOTKooper:
> I wonder if this is because alpine node does not install optional dependencies by default? (I don't know this to be true, but that's my guess)

doesn't look like that, it fails even with explicit `--include=optional`

squishyu:
Seeing some strange behavior. Bun itself installs, the postinstall script failed.
When installing with `--ignore-scripts` it properly installs the optional dependencides: `@oven/bun-linux-x64` and `@oven/bun-linux-x64-baseline`

![image](https://github.com/oven-sh/bun/assets/28161629/069ce63e-1267-4160-8116-fa7b839faa92)

When I then manually run the postinstall script (`npm run postinstall`), it throws the original error.

Upon digging further, on the **alpine** image the binary file shows up in the bin directory.

[...]

PR #14936, opened 2024-11-01
Changes expect().resolves.toThrow behaviour to more closely match jest

### LINKED ISSUE #9687: [bun:test] expect resolves not toThrow fails

1.0.35+940448d6b

Linux 5.15.146.1-microsoft-standard-WSL2 x86_64 x86_64

1. Create a file named `test.spec.ts` with the following content:
```ts
import { expect, it } from 'bun:test';

it('should not throw', async () => {
  async function nop() { }
  await expect(nop()).resolves.not.toThrow();
});
```

2. Run `bun test`.
3. See that it fails.

I'd expect the test to succeed, since the tested function did not throw.

### What do you see instead?

error: expect(received).not.toThrow()

Thrown value: undefined

      at /home/diegosogari/test/test.spec.ts:5:9
      at /home/diegosogari/test/test.spec.ts:3:24
✗ should not throw [0.50ms]

 0 pass
 1 fail
 1 expect() calls

### Additional information

_No response_

### LINKED ISSUE #14076: false-negative test failure when using expect().resolves.not.toThrow()

1.1.27+267afa293

Darwin 23.5.0 arm64 arm

```ts
test("foo", async () => {
  await expect(Promise.resolve("hello world")).resolves.not.toThrow();
});
```

test passes

### What do you see instead?

```
test/x.test.js:
1 | test("foo", async () => {
2 |   await expect(Promise.resolve("hello world")).resolves.not.toThrow();
                                                                ^
error: expect(received).not.toThrow()

Thrown value: "hello world"

      at /Users/meghandenny/src/bun3/test/x.test.js:2:61
      at /Users/meghandenny/src/bun3/test/x.test.js:1:13
✗ foo [3.07ms]

 0 pass
 1 fail
 1 expect() calls
Ran 1 tests across 1 files. [30.00ms]
```

### Additional information

looks like `.resolves.not` together is causing the test failure and `.toThrow()` is overriding the final message

pfgithub:
### What does this PR do?

Fixes #14076, Fixes #9687

toThrow is for functions, like `expect(() => {throw ...}).toThrow()`, but jest also allows it to be used with `resolves`/`rejects` if the value is an error: `expect(Promise.resolve(new Error("abc"))).resolves.toThrow("abc")`.

Previously, bun's implementation allowed any result to count as throwing, for example `expect(Promise.resolve("abc")).resolves.toThrow("abc")`. This updates bun:test to match jest and only count an Error result throwing.

- [x] Code changes


robobun:
@pfgithub, your commit aef5c4573938b205b6030c661aa927bc3a6444ce is building: [`#5545`](https://buildkite.com/bun/bun/builds/5545)

robobun:
Closing this draft as superseded.

The file it changes, `src/bun.js/test/expect.zig`, was removed in the Rust rewrite (no Zig sources remain on `main`; this logic now lives in `src/runtime/test_runner/expect.rs`), so the branch can no longer be rebased.

The same change is now in #32952, which gates the settled value on Jest's `isError` check for `.resolves` and also for `.rejects`. The `.resolves` cases from the two regression tests here are covered by the test added in that PR. The one intentional difference is `.rejects.toThrow()` on a non-Error rejection value such as `null`: this branch kept that passing, while #32952 makes it fail with "did not throw", as Jest does. If you think `.rejects` should stay unconditional, #32952 is the place to raise it.

#9687 stays open to track the bug until a fix lands (#14076 was closed as a duplicate of it). Thanks for the original fix and the test cases, they were useful for cross-checking the replacement.

Issue #5548, opened 2023-09-16
`bun install` slow/freezes on ConnectionRefused error


ivankhm:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 21.2.0 x86_64 i386

### What steps can reproduce the bug?

- On Mac OS go to Settings > Network > Advanced > TCP/IP > Configure IPv6 and select Link-local Only, to disable ipv6.
- Create a clean directory, `cd` into it
- Run `bun add nuxt` or `bun add nuxt@2.17.0`

### What is the expected behavior?

Dependencies are installed faster then with `npm` on same machine.

### What do you see instead?

It stops on one random dependency, like  `@netlify/node-cookies [686/687]` or `@csstools/postcss-text-decoration-shorthand [125/140]`. I tried it a couple of times, sometimes its the same dependency, sometimes its not.

Sometimes it moves a little bit, like to `@netlify/node-cookies [690/691]` and then stops again.

Then, some time latter it's done, with 2 errors:
```bash
  🚚 @netlify/node-cookies [690/691] 
error: ConnectionRefused downloading tarball @trysound/sax@0.2.0
  🚚 @netlify/node-cookies [691/691] 
error: ConnectionRefused downloading tarball vscode-languageserver-types@3.16.0

 installed nuxt@3.7.3 with binaries:
  - nuxi
  - nuxt

 549 packages installed [464.57s]
```

Installation time was 7.7 mins.

### Additional information

My initial thought was, that it may be related to #4938. So I turned of ipv6, and it didn't help – so I'm making an issue.

I tested my network connection via `speedtest`, and my download speed is 248 Mbps, so I don't think its a slow network.

My best guess, that it may be system/network related issue, but I can't figure it out :(

---

For comparison, `npm init` and `npm i nuxt` in clean directory takes about `52s`. 

---

Just in case, its just a simplification, my pet-project with 21 `dependencies` and 11 `devDependencies` suffered the same issue.

PR #14937, opened 2024-11-01
Align `"encoding"` option in node fs with node

Jarred-Sumner:
### What does this PR do?

Align `"encoding"` option in node fs with node

Copied a test from node

### DIFF
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -5726,7 +5726,7 @@ pub const JSValue = enum(JSValueReprInt) {
     }
 
     pub fn toBoolean(this: JSValue) bool {
-        return cppFn("toBoolean", .{this});
+        return this != .zero and cppFn("toBoolean", .{this});
     }
 
     pub fn asBoolean(this: JSValue) bool {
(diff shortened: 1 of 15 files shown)


robobun:
@Jarred-Sumner, your commit 0d6d6c6d4d41238cef193f51191bb0c738119491 is building: [`#5548`](https://buildkite.com/bun/bun/builds/5548)

src/bun.js/bindings/bindings.zig
```diff
@@ -4012,6 +4012,32 @@ pub const JSValue = enum(JSValueReprInt) {
         return JSC__JSValue__getDirectIndex(this, globalThis, i);
     }
 
+    pub fn isFalsey(this: JSValue) bool {
```
paperclover, line 4015: use toBooleanSlow. when bindings like this are rewritten, care should be taken to match exactly how JSC or the spec does them. this implementation does not consider BigInt

paperclover (changes requested):
one insane nit but we shouldnt add these two new functions

otherwise fine

src/bun.js/bindings/bindings.zig
paperclover: image

image

paperclover (approved):

PR #5555, opened 2023-09-16
Follow-up for workspace docs #5379 and #5229

bdenham:
### What does this PR do?

This PR adds to the doc fixes in #5379 and #5229 by updating another workspace example to use `"workspace:*"` instead of `"*"`.

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #14940, opened 2024-11-01
Inline `process.versions.bun` in `bun build --compile`

Jarred-Sumner:
### What does this PR do?

Inline `process.versions.bun` in `bun build --compile`

This makes detecting if the compiled executable is Bun more statically analyzable.

For example, tree-sitter has this line:

```js
const binding = require("node-gyp-build")(__dirname);
```

This code is very difficult for `bun build --compile` to support.

Instead, they could do:
```js
const binding = typeof process.versions.bun === "string" ?  
  require(`./prebuilds/${process.platform}-${process.arch}.node`) : 
  require("node-gyp-build")(__dirname);
```

This allows `bun build --compile` to work with it.

There is a test

### DIFF
--- a/src/cli/build_command.zig
+++ b/src/cli/build_command.zig
@@ -36,6 +36,7 @@ pub const BuildCommand = struct {
     const compile_define_keys = &.{
         "process.platform",
         "process.arch",
+        "process.versions.bun",
     };
 
     pub fn exec(ctx: Command.Context) !void {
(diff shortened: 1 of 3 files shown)


robobun:
@Jarred-Sumner, your commit 41969fe5dba4658bb0df0e77dc98cd63c5138cea is building: [`#5555`](https://buildkite.com/bun/bun/builds/5555)

Issue #14687, opened 2024-10-19
Bun minified build produces invalid javascript when importing negative const enum values and using as object keys


zimmed:
### What version of Bun is running?

1.1.31

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

With the following files:

```typescript
// enum.ts
export const enum Foo {
  foo = -1,
  bar = 1,
}
```

```typescript
// index.ts
import { Foo } from './enum';

export const FooRecord = {
  [Foo.foo]: 'foo',
  [Foo.bar]: 'bar',
}
```

```typescript
// build.ts
await Bun.build({ entrypoints: ['index.ts'], minify: true, outdir: '.' });
```

Run

```bash
bun run build.ts
```

### What is the expected behavior?

The output should be:

```javascript
var a={[-1]:"foo",[1]:"bar"};export{a as FooRecord};
```

### What do you see instead?

The output is:

```javascript
var a={-1:"foo",1:"bar"};export{a as FooRecord};
```

Which produces syntax errors due to the unwrapped `-1` property key.

### Additional information

It is important to note that this does not happen for either of the following conditions:

- The build is not minified
- The enum is in the same file as the object

The same issue exists in v1.1.30, so it doesn't appear to be a regression.

PR #14947, opened 2024-11-01
Fix 14687: minification bug Only remove arrays symbols from number if >= 0

EliLichtblau:
### What does this PR do?

Fix [14687](https://github.com/oven-sh/bun/issues/14687). Negative numbers can't be propertyKey's without extra syntax
```ts
// enum.ts
export const enum Foo {
  foo = -1,
  bar = 1,
}
```
```ts
// index.ts
import { Foo } from './enum';
export const FooRecord = {
  [Foo.foo]: 'foo',
  [Foo.bar]: 'bar',
}
```
```ts
// build.ts
await Bun.build({ entrypoints: ['index.ts'], minify: true, outdir: '.' });
```
Old 
```ts
var a={-1:"foo",1:"bar"};export{a as FooRecord};

```

New 
```ts
var f={[-1]:"foo",1:"bar"};export{f as FooRecord};
```


paperclover (changes requested):
can you add tests in test/bundler/bundler_edgecase.test.ts

- cross file enum inlining
- same file enum inlining
- literal -1

RiskyMH:
@paperclover i added tests for them, however there does seem to be some interesting minification (see the lines I commented)

for most of the `const enum` it already was fine as all are `[]` but for **same file enum** it wasn't but now all numbers also get that treatment (not just negative). 

Even if no negative in enum it is putting square brakes so there probably needs to be some other fixes here, but this PR indeed still does fix one case of invalid JS.

test/bundler/bundler_edgecase.test.ts
```diff
@@ -2263,6 +2263,159 @@ describe("bundler", () => {
...
+    run: {
+      stdout: '{"1":"bar","-1":"foo"}',
+    },
+    onAfterBundle(api) {
+      const content = api.readFile("/out.js");
+      // Should use proper negative number syntax in property key
+      expect(content).toContain("[-1]");
+      expect(content).not.toContain("-1:");
+      // expect(content).toContain("1:");
+    },
```
RiskyMH, line 2325: this is the test that's behavior changes in the pr, all others stay the same

nektro:
merge conflicts and lack of response from OP. happy to followup review if they're resolved

Issue #5588, opened 2023-09-16
websocket - potential bug, weird behavior with ping/pong, maybe I'm missing something


robobun:
minimal reproducible
```js
const server = Bun.serve({
    fetch(req) {
        console.log('SERVER incoming request');
        server.upgrade(req);
    },
    websocket: {
        open(ws) {
            console.log(`SERVER open`);
            ws.ping(); // removing this fixes the issue - this is probably the buggy part
        },
        close(ws, code, reason) {
            console.log(`SERVER close: code=${code} reason=${reason}`);
        },
        message(ws, message) {
            console.log(`SERVER message: ${Bun.inspect(message)}`);
        }
    },
    port: 1337,
});

const ws = new WebSocket('ws://127.0.0.1:1337');

ws.onopen = async () => { // same behavior with addEventHandler
    ws.send('this message sends just fine');

    for await (const _ of console) { // removing this fixes the issue
        ws.send('hi'); // causes closure upon first line
    }
};
```

output with code shown above 
```
% bun run index.js
SERVER incoming request
SERVER open
SERVER message: "this message sends just fine"
```

output with server `ws.ping()` *or* client `for await (...)` excluded 
```
% bun run index.js
SERVER incoming request
SERVER open
SERVER message: "this message sends just fine"
  <-- ENTER (for await over console)
SERVER close: code=1006 reason=
```

I haven't a clue what's actually causing this, but the presence or absence of a ping doesn't seem like it should cause any problems. I previously had handlers for ping/pong messages, I was just testing them out (no intervals yet), but the issue was there before as well - I tried to reduce the fluff as much as possible to get a simple reproducible

_Originally reported on Discord: [`websocket - potential bug, weird behavior with ping/pong, maybe I'm missing something`](https://discord.com/channels/876711213126520882/1152579977192153158/1152579977192153158)_

DeveloperHarris:
Encountering this issue as well. The bun websocket client seems to have a bug in regards to receiving pings that causes the connection to close?

I also have a simple ServerWebSocket setup and a WebSocket like you, and am able to get some intervaled pings to work, but not others. 

ping("42") for example works, but ping() or ping("hello") does not.

Very strange.

```typescript
// Server
let timer;

Bun.serve({
    fetch(req, server) {
        // upgrade the request to a WebSocket
        if (server.upgrade(req)) {
            return; // do not return a Response
        }
        return new Response("Upgrade failed :(", { status: 500 });
    },
    websocket: {
[...]

PR #14949, opened 2024-11-02
fix(net) fix bytesWritten drain

cirospaciari:
Makes sure that the drain callback is only called after internal buffer is empty, and account for the written bytes inside flush.

### What does this PR do?

- [X] Code changes

Manually


robobun:
@cirospaciari, your commit 1b1742952d5b66d1a79676b5244ae7df5b43e150 is building: [`#5588`](https://buildkite.com/bun/bun/builds/5588)

PR #14952, opened 2024-11-02
make tickWithoutIdle loop no longer use 100% cpu

pfgithub:
### What does this PR do?

- [ ] Code changes


robobun:
@pfgithub, your commit 4c439e2f19ebb6da50555b15a8a0e487ac025f88 is building: [`#5600`](https://buildkite.com/bun/bun/builds/5600)

Jarred-Sumner (changes requested):
we can't really ship this, we need to fix the underlying cause

PR #14956, opened 2024-11-02
chore(docs): improve readme.md markdown structure

okineadev:
### What does this PR do?

Improved the Markdown structure in `README.md`, replaced HTML tags with native Markdown, which greatly improved the readability of Markdown

- [ ] Code changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,23 @@
-<p align="center">
-  <a href="https://bun.sh"><img src="https://user-images.githubusercontent.com/709451/182802334-d9c42afe-f35d-4a7b-86ea-9985f73f20c3.png" alt="Logo" height=170></a>
-</p>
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Really appreciate your enthusiasm, though using regular HTML there was intentional to be more precise than what the markdown renderer does

okineadev:
You will regret it...

okineadev:
> HTML there was intentional to be more precise than what the markdown renderer does

How will you prove that it is more accurate and better?

Can you show some articles or other materials that confirm this?

After my changes, at least something shifted by one pixel ??


I don't think this is the right decision

sequencerr:
Don't make such a fuss over a markdown change and adjust your tone. This only makes your pr less likely to be merged

PR #14959, opened 2024-11-02
Fix importing of ES modules from data URIs

### LINKED ISSUE #3566: Bun can't Dynamically Import with Data URI

0.6.13

Linux 4.4.0-18362-Microsoft x86_64 x86_64

```js
import("data:text/javascript;base64," + btoa("export function sayHi() { return 'Hi!'; };")).then((script)=>{
    console.log(script.sayHi());
});
```

```
Hi!
```

### What do you see instead?

```
error: FileNotFound reading "data:text/javascript;base64,ZXhwb3J0IGZ1bmN0aW9uIHNheUhpKCkgeyByZXR1cm4gJ0hpISc7IH07"
```

### Additional information

_No response_

yamalight:
### What does this PR do?

This PR makes Bun correctly load module body when importing modules from data URIs.

Fixes #3566

This is my first attempt at zig / contibuting to Bun. 
I'm almost certain I might've changed things in the wrong places / wrong way - but tests pass, so here it is. 
Please do let me know if there are any issues - would be happy to fix / adjust!

- [x] Code changes

### DIFF
--- a/src/bundler.zig
+++ b/src/bundler.zig
@@ -51,6 +51,7 @@ const Resolver = _resolver.Resolver;
 const TOML = @import("./toml/toml_parser.zig").TOML;
 const JSC = bun.JSC;
 const PackageManager = @import("./install/install.zig").PackageManager;
+const DataURL = @import("./resolver/data_url.zig").DataURL;
 
 pub fn MacroJSValueType_() type {
     if (comptime JSC.is_bindgen) {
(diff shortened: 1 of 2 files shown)


src/bundler.zig
```diff
@@ -1300,6 +1301,18 @@ pub const Bundler = struct {
                 break :brk logger.Source.initPathString(path.text, "");
             }
 
+            if (strings.startsWith(path.text, "data:")) {
+                const data_url = DataURL.parseWithoutCheck(path.text) catch |err| {
+                    bundler.log.addErrorFmt(null, logger.Loc.Empty, bundler.allocator, "{s} parsing data url \"{s}\"", .{ @errorName(err), path.text }) catch {};
+                    return null;
+                };
+                const body = data_url.decodeData(bun.default_allocator) catch |err| {
```
Jarred-Sumner: `bun.default_allocator` will leak memory here, instead lets use `this_parse.allocator`
```suggestion
                const body = data_url.decodeData(this_parse.allocator) catch |err| {
```

Jarred-Sumner:
Thank you for this. I'm honestly surprised this is all it took to make it work.

test/js/node/string-module.test.js
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from "bun:test";
+
+test("should import and execute ES module from string", async () => {
```
Jarred-Sumner, line 3: Can you add tests for when parsing the module fails that it throws a `BuildMessage` and has a source url line?

yamalight:
@Jarred-Sumner done and done. thanks for your guidance!

Jarred-Sumner:
Thank you

http-teapot:
I am experiencing this issue on the latest version 1.2.5 (and the version before that). I can't figure out where this fix landed, was it shipped?

PR #14965, opened 2024-11-03
Ignore non-integers in process.reallyExit()

Jarred-Sumner:
### What does this PR do?

Ignore non-integers in process.reallyExit() to match Node.js behavior

Intent was for https://github.com/oven-sh/bun/issues/14955,but it does not fix that


robobun:
@Jarred-Sumner, your commit 2dcb2b08c8b0dfc824968168ffccc0438a8422a2 is building: [`#5617`](https://buildkite.com/bun/bun/builds/5617)

Issue #5622, opened 2023-09-17
express-fileupload stuck while file upload


Ja4V8s28Ck:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Linux 6.4.0-kali3-amd64 x86_64 unknown

### What steps can reproduce the bug?

The bug is when I try to use express-fileupload to upload an "image", it's stuck on uploading. I believe the problem is that busboy is not parsing or can't parse the data buffer sent.

To replicate the issue, I used the example files given by express-fileupload
- [Download](https://github.com/richardgirges/express-fileupload/tree/master/example) index.js and index.html and the use app.listen to host index.html.
- Install the dependencies `express and express-fileupload`, add the `type:"module"` in package.json.
- Then `bun run index.js`
- Get to the page ==> Choose an image ==> Click upload
- Page stuck on loading. (No file uploaded)

- And, I ran it with nodemon as `nodemon index.js` and the file uploads successfully.

### What is the expected behavior?

I enabled the debug option in express-fileupload to troubleshoot and know what went wrong. I ran index.js with nodemon again and saw the following.

![correct](https://github.com/oven-sh/bun/assets/66721411/edb72127-9b3d-44b0-89dc-8997ef873a49)

### What do you see instead?

While using bun, this happened

![wrong](https://github.com/oven-sh/bun/assets/66721411/c953e940-a4b8-4ff2-99df-22f30aee0368)

### Additional information

I think something went wrong with the busboy parsing.

andres-linares:
I am going through the same issue. Somehow using Postman and adding the file directly works but when sending images from reactnative creating a FormData it gets stuck

Ja4V8s28Ck:
I believe the bug has been fixed in the newest version of Bun (1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a).

andres-linares-int:
I'm still having the same issue

Ja4V8s28Ck:
> I'm still having the same issue

I've re-opened the issue

PsiKai:
Commenting for reach and updates.  I am experiencing the exact same issue on `1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a`

imegres:
Same here, i tried to run my app with ts-node-dev, and work perfectly, but when i run with bun, only upload of smallest files (20kb) works.

kartojal:
[...]

PR #14980, opened 2024-11-04
Never try using epoll_pwait2 on old linux kernels

Jarred-Sumner:
### What does this PR do?

Never try using epoll_pwait2 on old linux kernels

This fixes an issue impacting RHEL 8


robobun:
@Jarred-Sumner, your commit a7f9cfb252b4fbab83b43d8f103e45b8b10c5016 is building: [`#5622`](https://buildkite.com/bun/bun/builds/5622)

Electroid:
This isn't exclusive to RHEL 8, right? We probably want to runs CI tests on both newer and older Linux kernels?

PR #14990, opened 2024-11-04
remove checks for self closing tags

### LINKED ISSUE #14515: <input> and <img> complain and crash if they are given children in JSX code, but shouldn't

1.1.30+7996d06b8

Linux 6.8.0-45-generic x86_64 x86_64

I do not use react, but another library that uses JSX.

In this library, it is perfectly valid to give those items children. I feel this is not the place of a JSX transpiler to emit that warning, since even typescript does not do so. In the following example, you can see that the `$on` function is used as a child to bind an event to the element as is the way in this library.

Also, it seems to confuse the scope handling since at some places, it actually crashes the build.

Here is a minimal example that crashes bun 1.1.30 :

```tsx
// test.tsx

export function Input(a: InlineInputAttrs, ch: DocumentFragment) {
  const o_model = a.model
  const nullable = (a.type||'').indexOf('null') > -1

  return <input>
      {$on('input', (ev) => {
        var v = ev.currentTarget.value
        if (nullable && v === '') {
          o_model.set(null!)
        } else {
          // @ts-ignore typescript is confused by the type of o_model, rightly so.
          o_model.set(to_obs(v))
        }
      })}

    </input>

}

function _pad(n: number) {
  return (n < 10 ? ('0' + n) : n)
}

function _iso_date(d: Date) {
  return `${d.getFullYear()}-${_pad(d.getMonth()+1)}-${_pad(d.getDate())}`
}
```

This is used alongside a simple `tsconfig.json`

```json
{
  "compilerOptions": {
    "target": "es6",
    "module": "commonjs",
    "moduleResolution": "node",
    "declaration": true,
    "sourceMap": true,
    "strict": true,
    "jsx": "react",
    "reactNamespace": "E",
  }
}
```

It should just transpile, not crash, and not output *warnings* as the code is perfectly valid.

### What do you see instead?

Running `bun build test.tsx` produces the following crash :

```
===========================================================
Bun v1.1.30 (7996d06b) Linux x64
Linux Kernel v6.8.0 | glibc v2.39
CPU: sse42 popcnt avx avx2 avx512
Args: "bun" "build" "test.tsx"
Features: tsconfig
Elapsed: 2ms | User: 0ms | Sys: 3ms
RSS: 1.07GB | Peak: 17.17MB | Commit: 1.07GB | Faults: 0

panic: Scope mismatch while visiting
17 |     </input>
           ^
warn: <input / is a void element and must not have "children"
   at /home/cey/projects/test/e/test.tsx:17:7

25 | function _iso_date(d: Date) {
                       ^
debug: Expected this scope (.function_args)
    at /home/cey/projects/test/e/test.tsx:25:19

9 |         if (nullable && v === '') {
                                      ^
debug: Found this scope (.block)
    at /home/cey/projects/test/e/test.tsx:9:35

Crashed while visiting /home/cey/projects/test/e/test.tsx

 https://bun.report/1.1.30/lb17996d06AggggggCw58gwD8wtnuDw0vwsD4ynkqDginhwD08ihuDi6/qsDg6g8kD2t56kD0mk0mDA0eNqtks1uwjAMx+99BS5/ceDjQkcRQyDYadsL7Ig0FFJDszVJ1biFaXv4KVE70EDTDuQQf8R2nF/8Im1B0MppwTLDIVM5oVZOsTL7qLMe6/EMX+is7zSAIJd+i5UpKn6IcFqddaJ9QitffVzkrYMojTe8vkCr+ZhlKIMYykGgtioF5aTJMIRJoSvHMJaRiZrQlZnK05JMN1RurhTcVJzc6zizmmJJH3FR2jeS7GImxzEFMWJ3/GmjyZno8ezSF1xR6D2ZNs9vDqd6VxnJyoYXYaOc3aSCaZAu8CiYhvg8h4J/AUppW+2vE3o6FiSZUnCmHFz4rsGo7WEjyr0btjhuwiOZXmE0PwGZn41DOxKBi9oF98BUeS62OaHXQ43VatWGJLrf9+ofiG5B7NlW5heubW7l+20xzS9dk0Au+gan//AO
```

### Additional information

If the two functions at the end are removed, it builds, but erroneously :

```js
function Input(a, ch) {
  const o_model = a.model;
  const nullable = (a.type || "").indexOf("null") > -1;
  return /* @__PURE__ */ React.createElement("input", null);
}
export {
  Input
};
```

Here, the children were removed !
Renaming input to input2 in the source code does produce the correct result :

```js
function Input(a, ch) {
  const o_model = a.model;
  const nullable = (a.type || "").indexOf("null") > -1;
  return /* @__PURE__ */ React.createElement("input2", null, $on("input", (ev) => {
    var v = ev.currentTarget.value;
    if (nullable && v === "") {
      o_model.set(null);
    } else {
      o_model.set(to_obs(v));
    }
  }));
}
export {
  Input
};
```

ceymard:
### What does this PR do?

This fixes #14515 by simply stopping the checks with self-closing tags in jsx, as it should not be bun's responsibility to do so. I can update the tests, but I'm not familiar with the structure, can I just drop a jsx or tsx file in `tests/snippets` ?

This is just code removal and very little is touched in the .zig. The use case I give in #14515 passes.

- [x] Code changes

### DIFF
--- a/src/js_lexer.zig
+++ b/src/js_lexer.zig
@@ -29,7 +29,6 @@ pub const StrictModeReservedWords = tables.StrictModeReservedWords;
 pub const PropertyModifierKeyword = tables.PropertyModifierKeyword;
 pub const TypescriptStmtKeyword = tables.TypescriptStmtKeyword;
 pub const TypeScriptAccessibilityModifier = tables.TypeScriptAccessibilityModifier;
-pub const ChildlessJSXTags = tables.ChildlessJSXTags;
 
 fn notimpl() noreturn {
     Output.panic("not implemented yet!", .{});
--- a/src/js_lexer_tables.zig
+++ b/src/js_lexer_tables.zig
@@ -552,26 +552,6 @@ pub const TypescriptStmtKeyword = enum {
     });
 };
 
-//  Error: meta is a void element tag and must neither have `children` nor use `dangerouslySetInnerHTML`.
-pub const ChildlessJSXTags = ComptimeStringMap(void, .{
-    .{ "area", void },
-    .{ "base", void },
-    .{ "br", void },
-    .{ "col", void },
-    .{ "embed", void },
-    .{ "hr", void },
-    .{ "img", void },
-    .{ "input", void },
-    .{ "keygen", void },
-    .{ "link", void },
-    .{ "menuitem", void },
-    .{ "meta", void },
-    .{ "param", void },
-    .{ "source", void },
-    .{ "track", void },
-    .{ "wbr", void },
-});
-
 // In a microbenchmark, this outperforms
 pub const jsxEntity = ComptimeStringMap(CodePoint, .{
     .{ "Aacute", @as(CodePoint, 0x00C1) },
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -16345,24 +16345,7 @@ fn NewParser_(
 
                             const runtime = if (p.options.jsx.runtime == .automatic) options.JSX.Runtime.automatic else options.JSX.Runtime.classic;
                             const is_key_after_spread = e_.flags.contains(.is_key_after_spread);
-                            var children_count = e_.children.len;
-
-                            const is_childless_tag = FeatureFlags.react_specific_warnings and children_count > 0 and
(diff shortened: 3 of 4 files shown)


Jarred-Sumner:
Thank you.

Issue #918, opened 2022-07-29
Musl LibC Support


theoparis:
### What is the problem this feature will solve?

It will allow bun to be used in linux distributions such as Alpine Linux or BunOS. This could also potentially solve #255.

### What is the feature you are proposing to solve the problem?

I have already *attempted* to implement this by creating a [patched dockerfile, build.zig and makefile for building bun under alpine linux](https://github.com/theo-os/bun-musl-build). However, it is missing proper FFI support, which seems to be an issue with the internals of bun (as FFI and static linking seems to work with [deno](https://github.com/theoparis/deno-static)).

benatkin:
Found this closed issue that suggests JavaScript Core may run on musl https://bugs.webkit.org/show_bug.cgi?id=187485

brynne8:
I'm also using Alpine Linux and being unable to run `bun`.

dargmuesli:
Please add a :+1: to the first post instead of adding "me too" comments :pray:

notramo:
WebKit is shipped on some `musl` Linux distros, which implies support for JavaScriptCore.

Nyaa97:
Easy to compile against musl libc.
I need motivation to write a "normal" patch, but its possible.
This requires:
add musl flag to tcc build in Makefile
-fPIC in build.zig
turn off checking GLIBC version in build.zig
turn off force GLIBC in workaround-missing-symbols.cpp
rebuild JSCore

If you are trying to build using only clang, some additional patches are required.

Tested and works well for musl+clang
![image](https://github.com/oven-sh/bun/assets/39170352/bd3ade57-55c5-4a87-b2bb-0ee0a225a9b2)

theoparis:
Why was the above comment marked as "off topic"? It is showing a temporary solution to the issue and it would help contribute torwards a permanent fix in the future.

brynne8:
> Easy to compile against musl libc. I need motivation to write a "normal" patch, but its possible. This requires: add musl flag to tcc build in Makefile -fPIC in build.zig turn off checking GLIBC version in build.zig turn off force GLIBC in workaround-missing-symbols.cpp rebuild JSCore
> 
> If you are trying to build using only clang, some additional patches are required.
> 
> Tested and works well for musl+clang ![image](https://user-images.githubusercontent.com/39170352/279450296-bd3ade57-55c5-4a87-b2bb-0ee0a225a9b2.png)

Great. Could you please upload the binary so that I can temporarily use?

[...]

Issue #5642, opened 2023-09-17
axios/fetch stalls program


callumok2004:
### What version of Bun is running?

Tried on 1.0.2 and 1.0.3+25aa51dfc4976d1477d2714e5742393f190dc39f

### What platform is your computer?

Linux 5.4.0-155-generic x86_64 x86_64

### What steps can reproduce the bug?

On WSL, everything seems to work just fine, but when I use my linux server, this does not work with bun, with with node everything runs just fine.

```ts
export const DefaultOptions = {
  http: {
    api: "https://discord.com/api",
    version: "10"
  }
};

export const Endpoints = {
  botGateway: "/gateway/bot"
};
//
export async function fetchGatewayInformation(token: string, guildsPerShard = 1000) {
  if (!token) throw new Error("DISCORD_TOKEN_MISSING");

  console.log("Fetching gateway information..."); // This logs
  const data = await fetch(`${DefaultOptions.http.api}/v${DefaultOptions.http.version}${Endpoints.botGateway}`, {
    headers: { Authorization: `Bot ${token.replace(/^Bot\s*/i, "")}`, "Accept-Encoding": "gzip,deflate,compress" }
  }).then(res => res.json());
  console.log("Fetched gateway information."); // This does not log.

  return {
    shards: Math.ceil((data.shards * 1000) / guildsPerShard),
    session_start_limit: data.session_start_limit
  };
}
```

I am also seeing this in my console which suggests the request was successful.
![image](https://github.com/oven-sh/bun/assets/35779365/b0aea52f-7807-453a-9c15-f7d863aa2d4f)

While it halts on the request, memory usage quickly increases until it eventually reaches the memory limit
![image](https://github.com/oven-sh/bun/assets/35779365/1e5d793b-e661-46da-8adc-423509cf7d77)

### What is the expected behavior?

Expecting a response, and for the code to continue as normal

### What do you see instead?

The code halts when making fetch/axios requests

### Additional information

I am unsure if this is something I am doing wrong myself, or an issue with bun, but I believe it could be bun as running the same code on node works as expected. It's also weird how it works fine on WSL but not on the server

Electroid:
Sorry that is happening, we'll take a look.

callumok2004:
Adding this screenshot to show what happens between WSL and linux server, tested on latest release
![image](https://github.com/oven-sh/bun/assets/35779365/0829aeec-4b08-4a68-aa50-0677f59ec9b3)

[...]

PR #14994, opened 2024-11-05
patches to allow linux-musl to bootstrap

nektro:

this patchset allows a user to build bun while on an alpine linux system so long as they have a prebootstrap bun binary
the goal of this pr is to get the changes in the main branch and verify they do not break any existing builds


robobun:
@nektro, your commit a6eec0fc105f9aaa0edb496e925b0579d19f9c09 is building: [`#5642`](https://buildkite.com/bun/bun/builds/5642)

.buildkite/ci.mjs
```diff
@@ -259,12 +270,15 @@ function getPipeline(buildId) {
   };
 
   const getBuildZigStep = platform => {
-    const { os, arch, baseline } = platform;
-    const toolchain = baseline ? `${os}-${arch}-baseline` : `${os}-${arch}`;
+    const { os, arch, abi, baseline } = platform;
+
+    let toolchain = `${os}-${arch}`;
+    if (abi) toolchain += `-${abi}`;
+    if (baseline) toolchain += `-baseline`;
```
nektro:
```suggestion
    const toolchain = getKey(platform);
```

src/sys.zig
```diff
@@ -1579,7 +1579,7 @@ else if (builtin.os.tag.isDarwin())
 else
     system.writev;
 
-const pread_sym = if (builtin.os.tag == .linux and builtin.link_libc)
+const pread_sym = if (builtin.os.tag == .linux and builtin.link_libc and !bun.Environment.isMusl)
```
Jarred-Sumner, line 1582: These use system calls

we should continue to use *64 in that case

Issue #5643, opened 2023-09-17
Inconsistent axios error behavior in bun


ShlomoCode:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. Install the [axios](https://www.npmjs.com/package/axios) library
2. run this code:
```
const axios = require('axios');

(async () => {
    const urls = [
        'https://www.microsoft.com/404', // server returns *html* with 404 status code. error is `AxiosError`
        'https://eastus.stt.speech.microsoft.com' // server returns only 404 status code. error is `ERR_TLS_CERT_ALTNAME_INVALID`
    ];

    for (const url of urls) {
        try {
            await axios.get(url);
        } catch (error) {
            console.error(url, error.name, error.message);
        }
    }
})();
```

### What is the expected behavior?

The error should come from axios when the request fails

### What do you see instead?

In node.js both errors come from axios and they are "AxiosError Request faile with status code 404"
**In bun**, the error in the second address, which does not return an answer but only an error code status, an error is received "`ERR_TLS_CERT_ALTNAME_INVALID fetch() failed. For more information, pass `verbose: true` in the second argument to fetch()`"

### Additional information

https://github.com/oven-sh/bun/assets/78599753/6e89bc74-8c42-4bd7-87fb-3d5e76accc44

I would be happy to provide any additional details needed

Zikama:
Facing same issue ```fetch() failed. For more information, pass `verbose: true` in the second argument to fetch()```

Jarred-Sumner:
i'm unable to reproduce this

<img width="618" alt="image" src="https://github.com/oven-sh/bun/assets/709451/62072fb4-8670-4481-a1ce-25fc6795544a">

sonu628:
@Jarred-Sumner I got the same, seems like the issue was either resolved in a newer version or was due to user's specifications

devHaitham481:
did this get resolved ? i'm getting the same error

grmkris:
same here

nospher3x2:
same here

1cedsoda:
I also can't reproduce in `1.1.0+5903a6141`

kyjus25:
+1

devcaeg:
+1

JRoy:
For people who need a workaround (do not use this in production)

```typescript
import { Agent } from "node:https"

[...]

PR #14995, opened 2024-11-05
fix vscode debugger

RiskyMH:
### What does this PR do?

The vscode debugger wasn't being enabled correctly unless using `--inspect`, as such things weren't working such as showing scoped variables.

manual testing


robobun:
@RiskyMH, your commit 1a9fd07797676fce90b94ea0d444164e56d129e5 is building: [`#5643`](https://buildkite.com/bun/bun/builds/5643)

src/bun.js/javascript.zig
```diff
@@ -1854,7 +1854,7 @@ pub const VirtualMachine = struct {
             },
         }
 
-        if (debugger != .unspecified) {
+        if (this.debugger != null) {
```
Jarred-Sumner, line 1857: Can you verify this doesn't cause us to start with the debugger enabled when it's not actively in use?

RiskyMH: I haven't managed to get any indication this is started unless using `--inspect` or the `BUN_INSPECT` env var.

It is hard to work out however because its the `this.debugger = Debugger` stage that starts the connection (with print).

Edit: after more testing, I still couldn't find any issues.

PR #15009, opened 2024-11-06
Fix unicode imports, unicode-escaped variable names, and printClauseAlias not working for utf-8

### LINKED ISSUE #4217: Length of a new line sequence differs in node and bun

### LINKED ISSUE #14784: [pm] package.json parser SyntaxError in workspaces

1.1.32+fe8d0079e

Linux 6.10.11+bpo-amd64 x86_64 unknown

Originally found with https://github.com/nestjs/typescript-starter/blob/fbe788c6f4161c62bcfd8e18508081dad1496af6/package.json#L66

See minimal reproduction repository at https://github.com/sequencerr/bun-repro-jsonparse
and run `bun i --cwd packages/foo pg` (actually any package) 

_No response_

### What do you see instead?

package installed but package.json not written/updated
```bash
bun add v1.1.32 (fe8d0079)

installed pg@8.13.0

[2.00ms] done
package.json failed to parse due to error SyntaxError
```

### Additional information

tracking progress in https://github.com/oven-sh/bun/pull/14779

### LINKED ISSUE #14976: Bun not recognizing Unicode characters in Import statements

1.1.34

macOS ARM 15.1

Tested against `1.1.30` and `1.1.34`

Bun doesn't recognize unicode characters, like `𐃘` in `mile𐃘nautical` in import statements. For example, if I have defined and exported `mile𐃘nautical` like so:

```ts
// ./lib/math.ts
export const mile𐃘nautical = (int: number) => Math.round(int * 1.150779448)
```

Then in another file, importing it directly fails:

```ts
// ./src/env.ts
import { atoi, mile𐃘nautical } from "./lib/math"

export const SEARCH_RADIUS = mile𐃘nautical(20)
```

Running bun `./src/env.ts`, I get the following error in my console:

```bash
1 | import { mile𐃘nautical } from "./lib/math"
    ^
SyntaxError: Unexpected token '}'. Expected 'as' after the module export name string.
      at <parse> (./src/env.ts:1:1)
      at native:11:43
```

---
However, importing the file aliassed, then accessing the exported constant works with the same `./lib/math.ts` definition. 

```ts
// ./src/env.ts
import * as math from "./lib/math"

export const SEARCH_RADIUS = math.mile𐃘nautical(20)
```

I expect the unicode character to be recognized as an import statement. I have tried this with node and the first example I showed earlier works.

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #15049: [bun build --compile] Compile-time inlining breaks system paths

1.1.34+5e5e7c60f

```ts
const APPDATA = process.env.APPDATA;
console.log(APPDATA);
```
Built using `bun build --compile`

`C:\Users\<user>\AppData\Roaming` is logged to the console (as per `bun run test.ts`)

### What do you see instead?

`C:Users<user>AppDataRoaming` is logged instead

### Additional information

Temporary workaround:
```ts
const { APPDATA } = process.env;
console.log(APPDATA); // works as expected
```

Similar:
#13063

pfgithub:
### What does this PR do?

Fixes #14976, Fixes #15049, Fixes #4217, Fixes #14784

Also fixes `let mile\u{100d8}value = 36;` not working, and `import { mile𐃘add1 as m } from "./import_target";` syntax error.

Also combines quoteForJSONBuffer & writeJSONString into one function because they were both almost identical

  - No new allocated memory


robobun:
@pfgithub, your commit 04417fa0bf510806e8fb294aeeabeddcc19c02b9 is building: [`#5690`](https://buildkite.com/bun/bun/builds/5690)

src/js_ast.zig
```diff
@@ -1443,9 +1443,7 @@ pub const OptionalChain = enum(u1) {
 };
 
 pub const E = struct {
-    pub const ToJsOpts = struct {
-        decode_escape_sequences: bool = true,
-    };
+    pub const ToJsOpts = struct {};
```
paperclover: every usage of this can be removed, along with its definition

src/js_ast.zig
```diff
@@ -2539,33 +2529,31 @@ pub const E = struct {
...
+            s.resolveRopeIfNeeded(allocator);
             if (!s.isPresent()) {
                 var emp = bun.String.empty;
                 return emp.toJS(globalObject);
             }
 
-            if (s.is_utf16) {
-                var out, const chars = bun.String.createUninitialized(.utf16, s.len());
+            if (s.isUTF8()) {
+                if (try strings.toUTF16Alloc(allocator, s.slice8(), false, false)) |utf16| {
```
paperclover, line 2534: leak of `utf16`, but that is ok because `allocator` is always an arena

dylan-conway, line 2534: We might need a test with invalid utf8

dylan-conway: Actually, should this replace invalid sequences?

pfgithub: `"abc"[0]` constant folding expects non-utf16 E.Strings to be ascii-only, and there is probably more code that does also. I think it would make sense to enforce E.String be either ascii or utf-16 rather than utf-8 or utf-16

dylan-conway: I agree. That makes sense

paperclover (commented):
i have not read all the code yet, but i'm very glad we're reviving the .length constant folding. will review the rest during compiles.

src/js_printer.zig
```diff
@@ -986,6 +933,9 @@ fn NewPrinter(
                     p.writer.print(@TypeOf(span), span);
                 },
                 else => {
+                    if (std.debug.runtime_safety and ascii_only) {
```
dylan-conway: we usually use `bun.Environment.allow_assert`

src/js_lexer.zig
```diff
@@ -693,20 +640,15 @@ fn NewLexer_(
             }
         }
 
-        pub const InnerStringLiteral = packed struct { suffix_len: u3, needs_slow_path: bool };
+        pub const InnerStringLiteral = packed struct { suffix_len: u3, needs_decode: bool };
 
-        fn parseStringLiteralInnter(lexer: *LexerType, comptime quote: CodePoint) !InnerStringLiteral {
-            const check_for_backslash = comptime is_json and json_options.always_decode_escape_sequences;
```
dylan-conway, line 699: pretty sure `always_decode_escape_sequences` can be removed, not used anywhere

PR #5711, opened 2023-09-19
fix(run): interpret extensionless files as typescript

paperclover:
### What does this PR do?

also fixes `require('yargs/yargs')`. im surprised we did not have an open issue about this.

Fixes #5173
Furthers #4207 

i wrote automated tests

PR #15015, opened 2024-11-06
Fix prototypical inheritance issue in node:zlib and add a `$toClass` helper

Jarred-Sumner:
### What does this PR do?

- We were missing Object.setPrototypeOf calls on all the Zlib classes (cc @nektro), which can cause bugs when things are subclassed.
- This ensures the function names are set for the various zlib exports

There is a test for the prototypical inheritance stuff. I manually verified it matches Node.


robobun:
@Jarred-Sumner, your commit 69beb329caf5e06f8257f7371f4f7acd20dc041e is building: [`#5711`](https://buildkite.com/bun/bun/builds/5711)

nektro (approved):

Issue #5712, opened 2023-09-19
--dump-environment-variables doesn't work, bun run env ignores files


Thinkscape:
### What version of Bun is running?

1.0.2

### What platform is your computer?

Darwin 22.6.0 arm64 arm

## The --dump-environment-variables does nothing

```
bun --help
[...]
            --prefer-latest                     Use the latest matching versions of packages in the Bun runtime, always checking npm
            --silent                            Don't repeat the command for bun run
            --dump-environment-variables        Dump environment variables from .env and process as JSON and quit. Useful for debugging
            --dump-limits                       Dump system limits. Useful for debugging
```

```
bun --dump-environment-variables
❯ bun --dump-environment-variables
Bun: a fast JavaScript runtime, package manager, bundler and test runner. (1.0.2)

  run       ./my-script.ts       Run JavaScript with Bun, a package.json script, or a bin
[...]
```
```
bun run --dump-environment-variables someScriptName.js
[0.08ms] ".env.development", ".env"
[...normal script output...]
```

```
bun --dump-environment-variables run someScriptName.js
[0.08ms] ".env.development", ".env"
[...normal script output...]
```

## `bun run env` doesn't load env files

```
❯ bun run env
PATH=...
ZPLUG_HOME=/opt/homebrew/opt/zplug
[...]
```

### What is the expected behavior?

1. When `--dump-environment-variables` is used with `run`, it should load the appropriate .env files, dump the variables and then run the script.
2. When running `bun run env` it should first load the `*.env` files, then show contents of `process.env`

### What do you see instead?

`--dump-environment-variables` does nothing
`bun run env` is equivalent to the posix `env` command, ignoring `*.env` files.

RiskyMH:
This flag appears to have been removed, so I don't think this intends to work anymore.

https://github.com/oven-sh/bun/pull/6395/files#diff-0213d803b050248119f629ea8f3e9dc3a9dd5fc1c78cd3176ccdabebb6a1b4e7L594

> `bun run env` not working

I think this answer is related https://github.com/oven-sh/bun/issues/9635#issuecomment-2021350123

PR #15016, opened 2024-11-06
[VSCode] Show more sources in debugger

Jarred-Sumner:
### What does this PR do?

If there's a source mapping url, we could show it


robobun:
@Jarred-Sumner, your commit 1fb34e57be5793f4a5f5e764186a9690f8472008 is building: [`#5712`](https://buildkite.com/bun/bun/builds/5712)

Issue #5714, opened 2023-09-19
Run `bun add -d <pkg>` to move an existing dep to devDependencies


jasongitmail:
### What is the problem this feature would solve?

It's common to forget the `-d` flag when installing a package.
It'd be nice to simply run `bun add -d <pkg>` to move it into devDependecies, like npm does. But Bun requires removing the package and adding it again using the desired flag.

### What is the feature you are proposing to solve the problem?

If a package is already installed as a non-dev dependency, running `bun add -d <pkg>` would move it in package.json into devDependencies.

### What alternatives have you considered?

`bun remove <pkg> && bun add -d <pkg>`

SnailBones:
Almost two years later, I'd still really appreciate this feature!

PR #15017, opened 2024-11-06
Fix issue with sveltekit-adapter-node

Jarred-Sumner:
### What does this PR do?

Fix issue with sveltekit-adapter-node

Treat `unix: ""` like `if ("")`  in Bun.serve()


robobun:
@Jarred-Sumner, your commit dcf6154b841bff1487b25e354ec1c2040e891a16 is building: [`#5714`](https://buildkite.com/bun/bun/builds/5714)

Issue #5748, opened 2023-09-19
bug: Cannot find module "@tiptap/pm/state" with bun run


rubenfiszel:
### What version of Bun is running?

v1.0.2

### What steps can reproduce the bug?

```
import * as tiptap from "@tiptap/core";

console.log("@tiptap/pm", tiptap);
```

bun run main2.ts

### What is the expected behavior?

bun run works

### What do you see instead?

```
error: Cannot find module "@tiptap/pm" from "/git/buntest2/main2.ts"
```

PR #15029, opened 2024-11-07
Bench updates

RiskyMH:
Some fixes to bench scripts.

Adds the bench site's webhook to the buildkite script.


robobun:
@RiskyMH, your commit 885b498b29d71deef1f7d4cc9bb5e58843269538 is building: [`#5748`](https://buildkite.com/bun/bun/builds/5748)

.buildkite/scripts/upload-release.sh
```diff
@@ -206,6 +221,7 @@ function create_release() {
 
   update_github_release "$tag"
   create_sentry_release "$tag"
+  send_bench_webhook "$tag"
```
RiskyMH, line 228: @Electroid this is untested and I'm assuming things here, let me know if its wrongly done or if better way.

also this needs the `BENCHMARK_URL` to make it more hidden from public.

Jarred-Sumner:
This is fine and we should merge it, but it shouldn't error if the benchmark_url is unset

RiskyMH:
> This is fine and we should merge it, but it shouldn't error if the benchmark_url is unset

Done, I also updated the environment variable to `$BENCHMARK_URL` for clarity.

Issue #15044, opened 2024-11-08
`--preload` with `new Worker`


SlexAxton:
### What version of Bun is running?

1.1.33+247456b67

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

```bash
bun --preload preload.ts ./app.ts
```

```tsx
# preload.ts

globalThis.EXPANDO = 'EXPANDO';
```

```tsx
# app.ts
console.log('From app', globalThis.EXPANDO);

const worker = new Worker('./worker.ts');
```

```tsx
# worker.ts
console.log('From worker', globalThis.EXPANDO);
```

### What is the expected behavior?

```
From app EXPANDO
From worker EXPANDO
```

### What do you see instead?

```
From app EXPANDO
From worker undefined
```

### Additional information

I'm not super sure how it should work for sure! But! For something like opentelemetry, it means that all of my telemetry is silently ignored inside of my workers.

I'd hope there's *some* way to make it to where I could reliably load the `telemetry.ts` file for everything.

It's possible there's a better way than `--preload` to add the telemetry file, but that's how @Jarred-Sumner is suggesting it in this issue: https://github.com/oven-sh/bun/issues/3775#issuecomment-1892843091

SlexAxton:
It's possible workers should have their own `--worker-preload` or simply take the preload argument as an option on `new Worker(path, {preload: './file/from/root.ts'})`

Jarred-Sumner:
--worker-preload sounds good + adding a `preload` argument

truongan07:
i may transfer data to worker by globalThis variable without postMessage method ?

Jarred-Sumner:
> i may transfer data to worker by globalThis variable without postMessage method ?

i wish

SlexAxton:
(It’s a different `globalThis` instance is why)

Issue #5787, opened 2023-09-20
yan.lock(v1) is invalid yaml; Support Serialization to: Yaml & JSON


hsjobeki:
### What is the problem this feature would solve?

I'd like to parse the yarn lockfile.
Currently this is very complicated because yarn.lockfile (V1) is invalid yaml.

### What is the feature you are proposing to solve the problem?

I'd like to just parse the lockfile in a simple and reliable way.  For me this could be solved by:

- Allow users to specify the serialization format "yarnv1", "yaml", "json"

### What alternatives have you considered?

Parsing yarn.lock with additional npm packages (annoying / feels flakey)
Converting yarn.lock v1 to yarn.lock v2/v3 (annoying / feels flakey)

Electroid:
Can you elaborate a bit? Are you using `bun install --yarn`?

hsjobeki:
Yes i'm using bun install --yarn. Problem beeing that yarn lockfile version 1 is not valid yaml. (By Yarn specification)
Therefore i cannot use a yaml parser to deserialize the yarn.lock (v1)
This makes the yarn.lock (v1) pretty hard to reuse for third party tools. 
I'd need a standardized format, where i can get an of-the-shelf parser such as yaml and/or json.
Otherwise this yarn.lock feature is only nice for git diff and nothing else.

hsjobeki:
A little further context i evaluted bun as a node alternative. But i need to integrate it with my build processes. This requires pre calculating the dependency graph. Which is not possible because i cannot parse the lockfile(s) of bun

hsjobeki:
@Electroid, I think the npm label you assigned doesn't match the problem.
I'd suggest a label like `lockfile` and/or `ecosystem`

RiskyMH:
The yarn.lock output file isn't as useful now that we have `bun.lock` a JSON*C* based lockfile. Only caveat is its also not fully standard (as it has trailing comma to make diffs better).

hsjobeki:
> The yarn.lock output file isn't as useful now that we have `bun.lock` a JSON_C_ based lockfile. Only caveat is its also not fully standard (as it has trailing comma to make diffs better).

Nice. Thanks for the fix. I think i can pipe the lockfile through prettier or another tolerant formatter to get rid of the trailing commas and get something jsonc-compliant.

PR #15045, opened 2024-11-08
Support `preload` option in `Worker`

Jarred-Sumner:
### What does this PR do?

Support `preload` option in `Worker`

Fixes https://github.com/oven-sh/bun/issues/15044

Tests

### DIFF
--- a/src/bun.js/bindings/webcore/Worker.cpp
+++ b/src/bun.js/bindings/webcore/Worker.cpp
@@ -122,7 +122,7 @@ extern "C" void* WebWorker__create(
     uint32_t argvLen,
     StringImpl* execArgvPtr,
     uint32_t execArgvLen,
-    StringImpl** preloadModulesPtr,
+    BunString* preloadModulesPtr,
     uint32_t preloadModulesLen);
 extern "C" void WebWorker__setRef(
     void* worker,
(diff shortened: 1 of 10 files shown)


robobun:
@Jarred-Sumner, your commit 96e717082a79fce8cfc010e759a5eecc833bb572 is building: [`#5787`](https://buildkite.com/bun/bun/builds/5787)

src/bun.js/bindings/webcore/Worker.cpp
```diff
@@ -160,6 +162,16 @@ ExceptionOr<Ref<Worker>> Worker::create(ScriptExecutionContext& context, const S
 
     Vector<String>* argv = worker->m_options.bun.argv.get();
     Vector<String>* execArgv = worker->m_options.bun.execArgv.get();
+    Vector<String>* preloadModuleStrings = &worker->m_options.bun.preloadModules;
+    Vector<BunString> preloadModules;
+    preloadModules.reserveInitialCapacity(preloadModuleStrings->size());
+    for (auto& str : *preloadModuleStrings) {
+        if (str.startsWith("file://"_s)) {
+            str = WTF::URL(str).fileSystemPath();
```
paperclover: url could start with file:// but be invalid
```
WTF::URL url(str);
if (url.isValid()) str = url.fileSystemPath()
```

src/bun.js/web_worker.zig
```diff
@@ -90,6 +179,8 @@ pub const WebWorker = struct {
         argv_len: u32,
         execArgv_ptr: ?[*]WTFStringImpl,
         execArgv_len: u32,
+        preload_modules_ptr: ?[*]bun.String,
```
paperclover, line 182: c doesn't have a way to pass a non-null garbage pointer? yikes.

paperclover (changes requested):
one nitpick

Jarred-Sumner:
Fixed

PR #15066, opened 2024-11-09
musl patches [v4]

nektro:


robobun:
@nektro, your commit 9513a389a9b5c0b9dd8df4831a6e6d7446de5d21 is building: [`#5838`](https://buildkite.com/bun/bun/builds/5838)

src/js/thirdparty/detect-libc.linux.js
```diff
@@ -1,10 +1,13 @@
 // Hardcoded module "detect-libc" for linux
+
+const is_glibc = process.report.getReport().header.glibcVersionCompiler !== undefined;
```
Jarred-Sumner: process.report.getReport() is kind of expensive, we could have a constant here.

PR #15071, opened 2024-11-09
Add scanner for yarn lockfile v1

unexge:
### What does this PR do?

This PR adds a scanner for yarn lockfile v1 format. The next steps would be adding a parser on top of the scanner to have a structured output, and then migrating it to Bun's lockfile similar to `package-lock.json`.

I'm unfamiliar with Bun's codebase, it might be missing some coding conventions. Please let me know if so.

- [x] Code changes

I added unit tests for the scanner, and ran them with `zig test ./src/install/migration/yarn.zig`.

- [] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be - _there is no allocation in the scanner_
- [ ] JSValue used outside outside of the stack is either wrapped in a JSC.Strong or is JSValueProtect'ed - _it's not exposed to JS at the moment_
- [ ] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test-file-name.test`) - _it's not exposed to JS at the moment_


Jarred-Sumner:
Thank you for this. If we merge this as-is, it will be dead code though which often ends up no longer compiling

It's interesting that you're using zig test here - we haven't really adopted zig test in bun's codebase because it's difficult to use for more than a handful of files, particularly when there are native libraries to link and code necessary to initialize files. We end up using `bun test`, but it forces us to expose many functions to JS which we wouldn't otherwise intend to expose (see `bun:internal-for-testing`).

unexge:
Closing for now. Will work on the remaining parts and converting Zig tests to JS tests.

sasial-dev:
Doesn't bun already understand yarn v1 lockfiles? 
`bun ./bun.lockb` returns
```yml
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1

...
```

Jarred-Sumner:
> Doesn't bun already understand yarn v1 lockfiles? `bun ./bun.lockb` returns
> 
> ```yaml
> # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
> # yarn lockfile v1
> 
> ...
> ```

Bun writes them, but doesn’t read them

unexge:
I added the parser and also moved tests to JS with https://github.com/unexge/bun/commit/f1c533790759efc34f0e56174bbb8b4d61b67587. I tested the parser with lock files from some projects, such as React, Angular, Webpack, Flow, etc., and it seems to be working fine.

Not sure if it would be useful with its current state, if so I can open a PR to get some reviews on the parser. I'll continue working on migrating Yarn lockfiles to Bun lockfiles once I get some free time.

PR #15076, opened 2024-11-10
Add micro-optimization to fs.readFile

Jarred-Sumner:
### What does this PR do?

This does two things:
1) Remove an `fstat` call in `fs.readFile` when the file is < 256 KB
2) Remove a temporary allocation when transcoding strings from input < 256 KB 

Existing tests do cover this. Particularly the fs-oom tests but the others too.

### DIFF
--- a/src/bun.js/node/node_fs.zig
+++ b/src/bun.js/node/node_fs.zig
@@ -3964,6 +3964,7 @@ const Return = struct {
     pub const ReadFile = StringOrBuffer;
     pub const ReadFileWithOptions = union(enum) {
         string: string,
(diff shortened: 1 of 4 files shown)


robobun:
@Jarred-Sumner, your commit abe1996736e86272962afb4ad8de7452a35725a1 is building: [`#5848`](https://buildkite.com/bun/bun/builds/5848)

cirospaciari (approved):
LGTM,  I notice the grpc tests failing because of certs and will be fixed here https://github.com/oven-sh/bun/pull/15090

PR #15077, opened 2024-11-10
Chore(docs): Remove unreachable code

erickarugu:
### What does this PR do?
- Removes unreachable code due to having multiple return Statements.


malezjaa:
image
This is used to show a diff

Issue #5857, opened 2023-09-21
Failure running NextJS with iron session example app.


GitHM:
### What version of Bun is running?

1.0.2+37edd5a6e389265738e89265bcbdf2999cb81a49

### What platform is your computer?

Linux 5.15.0-75-generic x86_64 x86_64

### What steps can reproduce the bug?

bunx create-next-app --example with-iron-session with-iron-session-app
cd with-iron-session-app
bun run dev

### What is the expected behavior?

Run the example app in dev mode

### What do you see instead?

several lines of warning messages and a syntax error :
SyntaxError: Unexpected token '??='

GitHM:
Sorry this is not a bun bug, the problem is in the NextJS example.

PR #15086, opened 2024-11-11
Postgres client - more progress

Jarred-Sumner:
### What does this PR do?

- Binary format for `Date` (`timestamptz`, `timestamp` OIDs)
- Fixed a couple bugs that could cause hangs
- More passing tests
- Only keep the process open while there is an active query in-flight

The following syntax works:
```js
import {sql} from 'bun';
const users = [
  { name: "Boop" },
  { name: "Beep" },
];
await sql`insert users into ${sql(users)} returning *`
```

As well as:
```ts
import {sql} from 'bun';
const users = [
  { name: "Boop", age: 123 },
  { name: "Beep", age: 456 },
];
await sql`insert users into ${sql(users, "name")} returning *`
```

Tests


robobun:
@Jarred-Sumner, your commit 98342eed41f80adbbfed7eb55c0402ee91e28122 is building: [`#5857`](https://buildkite.com/bun/bun/builds/5857)

cirospaciari (approved):

PR #15091, opened 2024-11-11
bustDirCache on FileSystemRouter.reload & fix of dir_cache keys in windows 

### LINKED ISSUE #4522: FileSystemRouter reload method not working

0.8.1

Darwin 22.5.0 x86_64 i386

Hello !

I'm trying to used FIleSystemRouter with watch function from fs but when i renamed, moved, deleted a file in the watched directory, the router keep the last route state (same routes, no change)

i tried many thing, like removing watch function and create a setInterval and reload router and console the router, but still same

i prepared for you a little zip containing minimal reproduction code

repro:
   add a new page in pages directory and remove it after

[FileSystemRouter_issue.zip](https://github.com/oven-sh/bun/files/12538806/FileSystemRouter_issue.zip)

thank you ! 

refreshing the router instance to see the new discovered routes

### What do you see instead?

_No response_

### Additional information

_No response_

Kapsonfire:
### What does this PR do?

When the structure of the routes directory for the FileSystemRouter changes, the reload function cant load it, because readDirInfo is cached. So I changed the code to recursively bust the dirCache.

- [X] Code changes

I dont know how to write tests for this. Ive tested it locally.

Fixes #4522


src/bun.js/api/filesystem_router.zig
```diff
@@ -211,6 +211,38 @@ pub const FileSystemRouter = struct {
...
+            return;
+        };
+
+        if (root_dir_info) |dir| {
+            if (dir.getEntriesConst()) |entries| {
+                var iter = entries.data.iterator();
+                outer: while (iter.next()) |entry_ptr| {
+                    const entry = entry_ptr.value_ptr.*;
+
+                    if (entry.base()[0] == '.') {
```
paperclover, line 216: this should also ignore `node_modules`, see `router.zig`'s `banned_dirs`

paperclover (changes requested):
nice. at a glance this looks reasonable, but you should add a test case to cover changing directories

Jarred-Sumner:
Can’t we bump the version number and then we don’t need to eagerly call bustDirCache?

src/bun.js/api/filesystem_router.zig
Kapsonfire: Done! I've exported the banned_dirs from router.zig and using it in the filesystemrouter

Kapsonfire: Also there's a test included now

paperclover (approved):

Kapsonfire-DE:
The key generation for dir_caches had a bug on windows.
Trailing slashes were removed but it was hardcoded to "/" - since Windows uses "\\" the bustDirCache function in Windows did not work properly. I've changed it by moving the hardcoded to a constant defined before. the constant is set on comptime. 
\\ for windows
else
/

src/allocators.zig
```diff
@@ -469,6 +469,11 @@ pub fn BSSMap(comptime ValueType: type, comptime count: anytype, comptime store_
         const Self = @This();
         const Overflow = OverflowList(ValueType, count / 4);
 
+        const TRAILING_SLASH_CHAR = if (Environment.isWindows)
```
paperclover: instead of this, can't `std.fs.path.sep_str` can be used?

(additionally, this variable should be lowercase, but for this situation it should just be removed)

paperclover (changes requested):
will run tests but one nitpick

src/allocators.zig
Kapsonfire: you are right - replaced it

PR #15108, opened 2024-11-12
Fixed Responses to OPTIONS Requests ignore Body

### LINKED ISSUE #15082: fetch with `method: "OPTIONS"` always returns ""

1.1.34+5e5e7c60f

Linux 5.15.153.1-microsoft-standard-WSL2 x86_64 x86_64

Run this code:
```ts
const result = await fetch("https://google.com", {
  method: "OPTIONS",
});
const data = await result.text();

console.log({ data });
```

This is the output I get using deno or doing the same request in Bruno:
```
{
  data: "<!DOCTYPE html>\n" +
    "<html lang=en>\n" +
    "  <meta charset=utf-8>\n" +
    '  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">\n' +
    "  <title>Error 405 (Method Not Allowed)!!1</title>\n" +
    "  <style>\n" +
    "    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}\n" +
    "  </style>\n" +
    "  <a href=//www.google.com/><span id=logo aria-label=Google></span></a>\n" +
    "  <p><b>405.</b> <ins>That’s an error.</ins>\n" +
    "  <p>The request method <code>OPTIONS</code> is inappropriate for the URL <code>/</code>.  <ins>That’s all we know.</ins>\n"
}
```

### What do you see instead?

the output is:
```
{
  data: "",
}
```

### Additional information

Same result using `axios`, `https`, `unirest`, `node-fetch`

Workaround for now
```ts
exec(`curl --request OPTIONS --url "https://google.com"`);
```

Kapsonfire:
Fixes #15082

Rules for containing body didnt align to Docs


src/http/method.zig
```diff
@@ -52,18 +52,19 @@ pub const Method = enum {
...
+        values.remove(.CONNECT);
         values.remove(.HEAD);
         values.remove(.TRACE);
-        values.remove(.OPTIONS);
         break :brk values;
     };
 
     const with_request_body: std.enums.EnumSet(Method) = brk: {
         var values = std.enums.EnumSet(Method).initFull();
+        values.remove(.CONNECT);
```
cirospaciari: this changes affects:
- [ ] Bun.serve
- [ ] node:http (client and server)
- [ ] fetch

We need tests for all this cases

src/http/method.zig
```diff
@@ -52,18 +52,19 @@ pub const Method = enum {
...
+        values.remove(.GET);
```
cirospaciari, line 62: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET says 
Screenshot 2024-11-13 at 12 45 34

cirospaciari (changes requested):
We should follow the spec and add tests for it if anything is changed

https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1

```
   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.
   ```
   
   for CONNECT we need more work:
   ```
     CONNECT is intended only for use in requests to a proxy.  An origin
   server that receives a CONNECT request for itself MAY respond with a
   2xx (Successful) status code to indicate that a connection is
   established.  However, most origin servers do not implement CONNECT.
   ```
   ```
    A server MUST NOT send any Transfer-Encoding or Content-Length header
   fields in a 2xx (Successful) response to CONNECT.  A client MUST
   ignore any Content-Length or Transfer-Encoding header fields received
   in a successful response to CONNECT.

   A payload within a CONNECT request message has no defined semantics;
   sending a payload body on a CONNECT request might cause some existing
   implementations to reject the request.
   ```
   
   for OPTIONS we also need more checks:
   
   ```
     A client that generates an OPTIONS request containing a payload body
   MUST send a valid Content-Type header field describing the
   representation media type.  Although this specification does not
   define any use for such a payload, future extensions to HTTP might
   use the OPTIONS body to make more detailed queries about the target
   resource.
   ```

src/http/method.zig
Kapsonfire: Whats your point here? GET is removed to have a request body - so thats already correct in my PR? or are you reffering to 
>  A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.

how shall we handle this then?

Kapsonfire-DE:
shall I remove the .CONNECT Change for simpleness of this PR to fix the existing bug and address the OPTIONS request body later?

src/http/method.zig
cirospaciari: for `node:http` and `node:https` we should have the same behavior that node.js has, for Bun.serve and fetch we should respect the https://fetch.spec.whatwg.org/ (you can test the behavior with chrome for example) we will need tests that pass in both (node and bun) to avoid regressions.

cirospaciari:
> shall I remove the .CONNECT Change for simpleness of this PR to fix the existing bug and address the OPTIONS request body later?

You can reduce the context to only change the CONNECT behavior but we will still need tests for Bun.serve, fetch and node:http and node:https

Kapsonfire-DE:
The current bug is, that Responses to OPTIONS have no body in bun, even though the server sends it. So I would reduce the changes, to allow OPTIONS Responses to have body and add a test for this.

So we can close this bug.


I'd address the OPTIONS Requests with Body in a later PR.
For CONNECT is more work to be done, so I'd remove it completely out of scope for this for now

cirospaciari:
> The current bug is, that Responses to OPTIONS have no body in bun, even though the server sends it. So I would reduce the changes, to allow OPTIONS Responses to have body and add a test for this.
> 
> So we can close this bug.
> 
> I'd address the OPTIONS Requests with Body in a later PR. For CONNECT is more work to be done, so I'd remove it completely out of scope for this for now

Sure I meant OPTIONS sorry, you can reduce the context for only 1 change at a time sure, we still will need the tests.

src/http/method.zig
Kapsonfire: Chrome: 
> Promise {<rejected>: TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.
  …}

Kapsonfire: node:http // node:https is ignoring the body in the request and not sending it without warnings/errors

cirospaciari (approved):

PR #15117, opened 2024-11-13
Fix bun run folder

### LINKED ISSUE #13819: bun --bun run build -> "Module not found"

1.1.27+267afa293

Darwin 23.6.0 arm64 arm

If there is both a package.json script entry called "build" and a root folder called "build" (which there often is - populated by said script), then running:

`bun --bun run build` will fail with:
error: Module not found "repo/build"

but `bun run build`, `npm run build` and Node 22 `node --run build` will succeed in running the package.json script.

I'd expect the package.json script to be executed, or have a way of making it explicit, similar to `node --run`, also so bun doesn't have to crawl the file-tree.

deno similarly has the `deno run build` which will also complain about the directory, but then there's `deno task build` to make it explicit.

### LINKED ISSUE #15743: Support to run scope command without `run` command

So I have a `package.json` like this:

```json
{
   "scripts": { "pack": "bun build", "test:todo": "bun test --todo" }
}
```

For `pack` command, I can simply run this `bun pack` and it works nicely. But for command with scope `test:todo`, `bun test:todo` is not working only this work `bun run test:todo`. I think it doesn't make sense.

Could we support `bun test:todo`? (and other characters such as `.`, `!`, `[`, `]`, `+`)

Or we can create a shortcut binary for `bun run` like `bunr` that means for running only package script command?

> actually I did this in machine, quite good - recommend this:
> 
> ```bash
> #!/usr/bin/env bash                                                                                           
> bun run $@
> ```
> 
> with `sudo chmod 751 bunr`

### LINKED ISSUE #16169: Scripts are not run from the root of the project

1.1.42+50eec0025

Linux 6.12.6-arch1-1 x86_64 unknown

- bun init
- add script s to package.json:

```json
{
  "name": "bunbug",
  "module": "index.ts",
  "type": "module",
  "devDependencies": {
    "@types/bun": "latest"
  },
  "peerDependencies": {
    "typescript": "^5.0.0"
  },
	"scripts": {
		"s": "bun index.ts"
	}
}
```
- cd to node_modules
- try and run `bun run s`

The script should be runnable from any subdirectory of the project.

From the npm docs:

_Scripts are run from the root of the package folder, regardless of what the current working directory is when npm run is called._

https://docs.npmjs.com/cli/v11/commands/npm-run-script

### What do you see instead?

from a subdir of the project:

```
$ pwd
/home/lewis/code/bunbug/node_modules
$ bun run s
$ bun index.ts
error: Module not found "index.ts"
error: script "s" exited with code 1
```

### Additional information

_No response_

### LINKED ISSUE #16482: cli: `bun --bun run <script>` sometimes does not look for scripts

1.1.44-canary.74+68089a099

Darwin 24.1.0 arm64 arm

```sh
git clone git@github.com:expressjs/express
cd express
bun install
bun --bun run test
```

Tests run

### What do you see instead?

```
error: Module not found '<omitted>/express/test'

Bun v1.1.44-canary.74+68089a099 (macOS arm64)
```

### Additional information

- `bun run test` works fine
- I cannot repro on a newly initialized codebase. Steps I took
  -  `bun init`
  - add this script: `"test": "bun index.ts"` -- works
  - then installing typescript and trying `"test": "tsc --noEmit"` -- also works
- adding/removing `tsconfig.json` has no effect

pfgithub:
This may be a breaking change, so it might wait for 1.2. When using `bun <script>` or `bun run <script>`, it now always checks in this order:

1. Is there a script in package.json with the name? Execute it and exit.
2. Try loading a module with that name. Can it be done? Execute it and exit.
3. Is there a binary in `node_modules/.bin` with that name? Execute it and exit.
4. (`bun run` only) Is there a binary in `$PATH` with that name? Execute it and exit.
5. Error.

Previously, it would try guessing if it looked like a script name and sometimes it would execute before checking scripts in package.json.

Fixes #13819, Fixes #15743, Fixes #16169, Fixes #16482

New features: If there is a script called `myscript.js` you can now run it with `bun myscript` just like node allows

TODO:

- [x] Docs in `cli/run` about resolution order
  - [x] Update the outdated line "Bun executes the script command in a subshell. It checks for the following shells in order, using the first one it finds: bash, sh, zsh."
    - On windows, it uses bun shell rather than a system shell. This should be mentioned.
  - [x] Add a new section at the bottom for resolution order
- [x] Remove entries added to launch.json
- [x] Remove irrelevant js_printer changes
- [x] Merge main
- [x] Pass tests:
  - [x] prisma.test.ts
  - [x] import-custom-condition.test.ts
  - [x] if-present.test.ts
  - [x] self-reference.test.ts

Test failures:

- [ ] bundler-plugin.test.ts (?)
- [ ] 08757.test.ts (windows, absolute path override code)
- [ ] if-present.test.ts (file not found -> module not found)
- [x] bun-run.test.ts
- [ ] wasi.test.js (windows, absolute path override code)

Consider:

- [ ] Make Run.boot skip generating `bun:main` and run directly instead. `bun:main` is generated in `entry_points.zig`.

Performance:
~0.5ms slower :/ even for `./` which is strange


robobun:
@pfgithub, your commit d661d6e2a28b025593ff693933344ba2cf5238ea is building: [`#5937`](https://buildkite.com/bun/bun/builds/5937)

RiskyMH:
I think you should also update the docs with this new order to be more explicit :) https://bun.sh/docs/cli/run

nektro:
very excited to see this land; of the top ~1000 packages, this affects 402

Jarred-Sumner:
is there a chance we could avoid doing this when a file path with an extension is passed? just so we don't read package.json files up to the root unnecessarily

pfgithub:
@Jarred-Sumner I'm not seeing any performance impact from this PR with `hyperfine "bun-45cbc92e81391b46a16c241043ea38c26310e931 run ./a.js" "bun-287f1628e78e2ecee0dfd937d78e3fe3e307187c run ./a.js" --shell=none --warmup=50`, although `bun-1.1.43` is ~0.2ms faster than latest main

src/cli/run_command.zig
```diff
@@ -1420,41 +1313,80 @@ pub const RunCommand = struct {
...
+
+        if (target_name.len == 1 and target_name[0] == '-') {
+            log("Executing from stdin", .{});
+
+            // read from stdin
+            var stack_fallback = std.heap.stackFallback(2048, bun.default_allocator);
+            var list = std.ArrayList(u8).init(stack_fallback.get());
+            errdefer list.deinit();
+
+            std.io.getStdIn().reader().readAllArrayList(&list, 1024 * 1024 * 1024) catch return false;
```
Jarred-Sumner, line 1435: let's avoid using zig std for this, they use too much unreachable for this to not cause issues for us.

instead, use bun.sys.File

Jarred-Sumner:
This isn't really done, but I'm going to merge it. 

Please add a follow-up PR that:
- [ ] Doesn't make entry-point-build and entry-point-run observable
- [ ] Avoids the use of std.fs for reading from stdin and instead uses bun.sys. Zig's use of unreachable will cause crashes for Bun.

There are 2 additional system calls to bun run from this PR. I think that's okay given the context.

Issue #159, opened 2022-05-06
Bun's Roadmap


Jarred-Sumner:
# Bun's Roadmap

The Bun team updates this issue to share our current roadmap, priorities, and goals. You can see our current update below, and previous updates at the bottom.

### Node.js compatibility

For every change we make to Bun, we run the Node.js test suite to ensure that Bun is compatible with Node.js. We also started a [bounty program](https://x.com/jarredsumner/status/1914830430811177181) within our team, to accelerate our progress.

<p align="center">
<img width="591" alt="Image" src="https://github.com/user-attachments/assets/e6efa3b3-6aa7-4d84-8d71-ac4556c3b1ca" />
</p>

Our goal is to be 90% compatible with Node.js. We'll be providing [updates](https://x.com/bunjavascript) as our progress nears completion.

### Features

While we continue to work on Node.js compatibility and bug fixes, we’re also going to be working on new features in upcoming releases of Bun, including:

- [x] [`#18812`](https://github.com/oven-sh/bun/pull/18812) [Redis](https://bun.sh/docs/api/redis) client
- [x] [`#19699`](https://github.com/oven-sh/bun/issues/19699) MySQL support with [`Bun.sql`](https://bun.sh/docs/api/sql)
- [x] [`#19701`](https://github.com/oven-sh/bun/issues/19701) SQLite support with `Bun.sql` (in addition to the existing [`bun:sqlite`](https://bun.sh/docs/api/sqlite) API)
- [x] [`#4824`](https://github.com/oven-sh/bun/issues/4824) Test runner support in VSCode
- [x] [`#7589`](https://github.com/oven-sh/bun/issues/7589) Running scripts with filters, and in parallel, using `bun run`
- [x] [`#947`](https://github.com/oven-sh/bun/issues/947) REPL support (replacement for existing, 3rd-party package: `bun repl`)
- [ ] Improve usage of `bun install` at work, including:
  - [x] [`#7157`](https://github.com/oven-sh/bun/issues/7157) Support migration from `pnpm-lock.yaml` to `bun.lock`
  - [x] [`#4844`](https://github.com/oven-sh/bun/issues/4844) Support [`catalogs`](https://pnpm.io/catalogs)
  - [ ] [`#6608`](https://github.com/oven-sh/bun/issues/6608) Support nested `resolutions` and `overrides`
  - [ ] Fix issues with workspaces, private registries, and corporate proxies

### Reliability

[...]

Issue #5947, opened 2023-09-23
Ubuntu/Linux Bun.serve throwing Segmentation error/INVALID_ARG_TYPE


Eltik:
### What version of Bun is running?

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

### What platform is your computer?

Linux 5.15.0-78-generic x86_64 x86_64

### What steps can reproduce the bug?

1. Run `bun init`.
2. Copy the following code into `index.ts`
```ts
Bun.serve({
    port: 3000,
    async fetch(req: Request) {
        return new Response("Hello World");
    }
})
```
3. On Linux/Ubuntu v22, run `bun run index.ts`

### What is the expected behavior?

The server should listen to the port given. I've tested on my other Ubuntu machine and on other ports. The issue does not persist on my MacBook Pro (Darwin 22.4.0 arm64 arm).

### What do you see instead?

<img width="1094" alt="image" src="https://github.com/oven-sh/bun/assets/76538547/5578c9f9-70af-4a5d-ad85-a2979fac3b96">

```
1 | Bun.serve({
  port: 3009,
  async fetch(req) {
    return new Response("Hello World");
  }
});
   ^
TypeError: 湕捩摯⁥潨瑳慮敭⁳畭瑳愠牬慥祤戠⁥湥潣敤⁤潦⁲潮⹷渊睥唠䱒椨灮瑵⸩潨瑳慮敭猠潨汵⁤潤琠敨琠楲正.湉整湲污挠湯楳瑳湥祣攠牲牯›慷捴敨⁲畭整⁸獩氠捯敫⁤桷湥椠⁴桳畯摬渠瑯戠⹥䤀呎㐶呟彏半䅖啌⡅半
 code: "ERR_INVALID_ARG_TYPE"

      at À Ùsting/index.ts:1:0
```

Error does not persist on MacOS! Mainly happens on Ubuntu/Linux.

### Additional information

Issue likely is thrown [here](https://github.com/oven-sh/bun/blob/ee93f1b88c141f1cc5d917a60ddb4a3e2706f1ff/src/bun.js/api/server.zig#L803).

Eltik:
If it helps, I've tested on version 1.0.0 as well as the canary build and version 1.0.3. The error exists on all versions.

Hanaasagi:
Related to https://github.com/oven-sh/bun/issues/5457 and https://github.com/oven-sh/bun/issues/3312#issuecomment-1704486463

[...]

PR #15120, opened 2024-11-13
allow zig js host functions to return JSError

nektro:
progress towards zig JSValue stability and safety in https://github.com/oven-sh/bun/issues/159
more followups to come


robobun:
@nektro, your commit 4b7dafc2637b957bd165d412a407901bcd7a6ada is building: [`#5947`](https://buildkite.com/bun/bun/builds/5947)

paperclover (commented):
should we actually be annotating every function with `bun.JSError!`? i think my solution to this in dave/more-bake is better, since it doesnt include the `catch` control flow on functions that don't return an error

nektro:
nearly every function annotated will be taking advantage of it eventually and unless there's a noticeable perf effect then the consistency makes the function's purpose very clear at a glance

paperdave:
if we want all of these functions to have the explicit annotation, it should be `bun.OOM || bun.JSError`, which i have in my branch defined as `bun.JSOOM` (can be named something else). the wrapper checks and handles both of these errors, if specified.

this is why i asked to wait a day for my branch to finish, as i'm nearly close to merging it. was this blocking something else?

nektro:
in a couple more stages i think it would make more sense for `OutOfMemory` to be a part of `bun.JSError`
pr is only blocking itself

paperclover (approved):

src/bun.zig
```diff
@@ -91,20 +91,11 @@ pub inline fn namedAllocator(comptime name: [:0]const u8) std.mem.Allocator {
...
-    ///
-    /// Prefer `global.jsErrorFromCPP()` over constructing this error manually,
-    /// that way the exception is asserted to actually exist.
-    ///
-    /// To pass this to C++ code as JSC::JSValue, use `global.errorUnionToCPP(...)`
+    /// There is an active exception on the global object.
+    /// You should almost never have to construct this manually.
     JSError,
+    // XXX: This is temporary! meghan will remove this soon
+    OutOfMemory,
```
dylan-conway, line 98: I think we should consider keeping this here and rename `error.JSError` to `error.ThrownException` or something similar

nektro: yeah there's many followups ahead of this

dylan-conway (approved):

PR #5972, opened 2023-09-23
add Socket.prototype & Readable.prototype

Geczy:
### What does this PR do?

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #15127, opened 2024-11-13
cpp: Bun::toStringRef: return dead when exception has been thrown

nektro:
tag is compared against here https://github.com/oven-sh/bun/blob/bun-v1.1.34/src/bun.js/bindings/BunString.cpp#L64
`::Empty` is `""`
`::Dead` means a JS exception has been thrown and we need to detect that

`Bun::toString` already does this correctly

### DIFF
--- a/src/bun.js/bindings/BunString.cpp
+++ b/src/bun.js/bindings/BunString.cpp
@@ -149,7 +149,7 @@ BunString toStringRef(JSC::JSGlobalObject* globalObject, JSValue value)
 {
     auto str = value.toWTFString(globalObject);
     if (str.isEmpty()) {
-        return { BunStringTag::Empty };
+        return { BunStringTag::Dead };
     }
 
     StringImpl* impl = str.impl();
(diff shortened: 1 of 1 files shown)


robobun:
@nektro, your commit 761833d40b44f0fad0b8be8b822efe6f54b01a81 is building: [`#5972`](https://buildkite.com/bun/bun/builds/5972)

src/bun.js/bindings/BunString.cpp
```diff
@@ -149,7 +149,7 @@ BunString toStringRef(JSC::JSGlobalObject* globalObject, JSValue value)
 {
     auto str = value.toWTFString(globalObject);
     if (str.isEmpty()) {
```
paperclover: this will break for genuinely empty (length == 0) strings

you should add a separate case before this with `str.isNull()`

example, the non-ref version of this function:

```
BunString fromJS(JSC::JSGlobalObject* globalObject, JSValue value)
{
    WTF::String str = value.toWTFString(globalObject);

    if (UNLIKELY(str.isNull())) {
        return { BunStringTag::Dead };
    }

    if (UNLIKELY(str.length() == 0)) {
        return { BunStringTag::Empty };
    }

    auto impl = str.releaseImpl();

    return { BunStringTag::WTFStringImpl, { .wtf = impl.leakRef() } };
}
```

dylan-conway (approved):

dylan-conway:
`fs.test.ts` test failures are suspicious. Will sync with main and let tests run again before merging

PR #15129, opened 2024-11-13
ci: upload-release.sh: add missing error message in send_bench_webhook

nektro:
noticed because main branch ci is running into it
now if this fails again we'll know exactly why


RiskyMH (approved):
woops i should have tested bash and error codes more, thanks for fix!

Electroid (approved):

Issue #15111, opened 2024-11-12
child_process.spawnSync does not provide signal


190n:
### What version of Bun is running?

1.1.34+5e5e7c60f

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

Create a simple executable that exits with SIGTRAP:

```zig
// trap.zig
pub fn main() void {
    @trap();
}
```
```sh
$ zig build-exe trap.zig
$ ./trap
zsh: trace trap  ./trap
```

Run a script to compare the results of `spawnSync` and `spawn`:

```js
// trap.mjs
import { spawn, spawnSync } from "node:child_process";

const sync = spawnSync("./trap", []);
console.log("sync: status", sync.status, "signal", sync.signal);

spawn("./trap", []).on("exit", (status, signal) => console.log("async: status", status, "signal", signal));
```

### What is the expected behavior?

`spawn` and `spawnSync` both provide signal information:

```sh
$ node trap.mjs
sync: status null signal SIGTRAP
async: status null signal SIGTRAP
```

### What do you see instead?

Only `spawn` provides the signal. In `spawnSync` the signal is always null:

```sh
$ bun trap.mjs
sync: status null signal null
async: status null signal SIGTRAP
```

### Additional information

`Bun.spawnSync` seems to work correctly.

cirospaciari:
Fixed by https://github.com/oven-sh/bun/pull/15137

PR #15137, opened 2024-11-13
Pass missing signal code for `child_process.spawnSync`

ippsav:
### What does this PR do?

Fixes(https://github.com/oven-sh/bun/issues/15111)

### DIFF
--- a/src/js/node/child_process.ts
+++ b/src/js/node/child_process.ts
@@ -563,6 +563,7 @@ function spawnSync(file, args, options) {
     stderr = null,
     success,
     exitCode,
+    signalCode,
   } = Bun.spawnSync({
     cmd: options.args,
     env: options.env || undefined,
@@ -573,7 +574,7 @@ function spawnSync(file, args, options) {
   });
 
   const result = {
-    signal: null,
+    signal: signalCode,
     status: exitCode,
     // TODO: Need to expose extra pipes from Bun.spawnSync to child_process
     output: [null, stdout, stderr],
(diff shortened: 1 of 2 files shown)


cirospaciari (changes requested):
We need tests so we know the issue is fixed on CI in all platforms.

cirospaciari (approved):

ippsav:
> > @cirospaciari I've added a simple test to replicate the issue, let me know if there are any more changes required
> 
> Thank you lets run the CI, running locally looks good.

I think i should wrap the code with an if to check if it's not Windows, since Windows doesn't support Posix signals

src/js/node/child_process.ts
```diff
@@ -573,7 +574,7 @@ function spawnSync(file, args, options) {
   });
 
   const result = {
-    signal: null,
+    signal: signalCode,
```
cirospaciari: Looking the tests and testing locally against node.js we should return null here if signalCode is undefined, you can it.if to not run on windows, we already have the test `test/js/node/test/parallel/unhandled-exception-with-worker-inuse.test.js` that will check for the null case

cirospaciari (changes requested):

src/js/node/child_process.ts
ippsav: good catch it should be null now in the case of no `signalCode` returned, and thanks for the tip didn't know `it.if(condition)` is a thing that's awesome

cirospaciari: Running the CI again lets see if the tests get something, looks good now.

cirospaciari (approved):

cirospaciari:
Thank you

PR #15144, opened 2024-11-14
Upgrade to SQLite v3.47.0

mchoo7:
### What does this PR do?

 - Upgrade to SQLite v3.47.0
 - Update `LICENSE.md` (state that SQLite is public domain)
 - Add `DEPENDENCIES.md` to track dependencies used by Bun.

If there are any other libraries that needs to be frequently updated from upstream, we should add them `DEPENDENCIES.md` better tracking.

- [x] Code changes

- [x] I ran my tests locally and they pass (`bun-debug test` in `test/js/bun/sqlite`)

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -1,0 +1,10 @@
+# Dependencies used by Bun
+
+The following list shows libraries obtained or forked from other projects.
+New versions of these depencies should be frequently checked for bug and security fixes.
+
+Read `LICENSE.md` for the licenses used by following dependencies.
+
+| Library                             | Location                     | Upgrading notes |
+| ----------------------------------- | ---------------------------- | --------------- |
+| [`sqlite`](https://www.sqlite.org/) | `src/bun.js/bindings/sqlite` | `UPGRADING.md`  |
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -18,54 +18,55 @@ This compiles JavaScriptCore, compiles Bun’s `.cpp` bindings for JavaScriptCor
 
 Bun statically links these libraries:
 
-| Library | License |
-|---------|---------|
-| [`boringssl`](https://boringssl.googlesource.com/boringssl/) | [several licenses](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/LICENSE) |
-| [`brotli`](https://github.com/google/brotli) | MIT |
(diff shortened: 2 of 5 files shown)


DEPENDENCIES.md
```diff
@@ -0,0 +1,10 @@
+# Dependcies used by Bun
```
ghiscoding:
```suggestion
# Dependencies used by Bun
```

infrahead:
Thought I'd piggyback on this instead of opening new issue:

1) 3.47.1 is now released

2) would it be possible to auto-update SQLite in the future?

similar to what I saw for https://github.com/oven-sh/bun/pull/15442

3) can sqlite be added to process.versions like other dependencies?

instead of having to do something like ` bun -e "import { Database } from 'bun:sqlite';const db = new Database();console.log(db.query('SELECT sqlite_version() AS version').get());"`


Also, I'm not familiar with how Bun is vendoring SQLite, but saw recently that the project itself will be changing build system as of 3.48
https://news.ycombinator.com/item?id=41921673

infrahead:
My highlights from SQLite major and patch releases since 3.45.0 (current Bun version).  As fodder for a future Bun blog update...

* **FTS5 improvements**
	* **can now create contentless FTS5 tables**
	* improved robustness while parsing the tokenize= arguments in FTS5
	* FTS5 tables can now be dropped even if they use a non-standard tokenizer that has not been registered.
	* fix the PRAGMA integrity_check command so that it works on read-only databases that contain FTS3 and FTS5 tables
* **JSON improvements**
	* restored legacy behavior of interpreting BLOBs as text JSON ( https://sqlite.org/json1.html#jblobbug )
	* handle corrupt JSONB inputs
	* new `json_pretty()` function
	* allow ASCII control characters within JSON5 string literals
	* allow negative values to RHS of the `->>` operator (to index from end)
	* fixed the `->` and `->>` operators so that when the right-hand side operand is a string that looks like an integer it is still treated as a string
* New capabilities in the date and time functions
	* new `ceiling` and `floor` modifiers  
	* `strftime()` now supports %G, %g, %U, and %V
* Allow `_` separators in numeric literals, for human readability.
* Numerous **performance optimizations** and query planner improvements
	* enhancements to the `PRAGMA optimize` command
* A multitude of bug fixes

fel1x-developer:
Closed by 6f8c5959d0826f22de8c5c9f9b020fddf7fea20a. Now sqlite is updated automatically by github actions.

Issue #14184, opened 2024-09-26
HTTP related regression in v1.1.27


Mudbill:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Darwin 23.6.0 arm64 arm

### What steps can reproduce the bug?

I noticed a strange issue while using the [contentful](https://github.com/contentful/contentful.js) package after upgrading Bun. After some testing, I discovered the issue only occurs on Bun versions after 1.1.26 (last working version).

**The issue is that the Contentful API client seemingly attempts to connect to their API URL using a combination of `https://` and port 80. This causes the connection to fail.**

I originally reported this in the Contentful Discord server before I found the relation to Bun. Therefore, I suspect this may be a Node-compatibility related issue.

Contentful uses Axios under the hood for communication. 

I haven't yet found the exact piece of code causing this, but I would assume it has something to do with the detection of which protocol to use and how to match up port 80 or 443 to that protocol. I will pop back with anything I find, but maybe someone more familiar with the related changes in v1.1.27 already has an inkling? 😅

## Reproduction

While using an affected Bun version (such as latest, v1.1.29 atm), create a new project and `bun add contentful`

In the `index.ts` file, add the following:

```ts
import { createClient } from "contentful";

const client = createClient({
  space: "example",
  accessToken: "nothing",
  retryLimit: 1, // just to skip retries for 30 seconds
});

await client.getEntries();
```

Run `bun .` on the file. The output will look something like this:

```
% bun .
[warning] Connection error occurred. Waiting for 2035 ms before retrying...
ConnectionRefused: Unable to connect. Is the computer able to access the url?

ConnectionRefused: Unable to connect. Is the computer able to access the url?
 path: "https://cdn.contentful.com:80/spaces/example/environments/master/entries"

Bun v1.1.29 (macOS arm64)
```

### What is the expected behavior?

I expect the API client to connect as intended, and as it did previously in version 1.1.26 or prior. The protocol should be `https://` and the port should be `443`.

### What do you see instead?

The API client seemingly attempts to connect to their API URL using a combination of `https://` and port `80`. This causes the connection to fail.

### Additional information

I tested several versions of Contenful's package. Both older and newer versions worked before the Bun upgrade, and both stopped working after.

This also likely has nothing to do with OS, as it also occurs on Linux/ARM (although I suppose I haven't tried an x86 OS)

Jarred-Sumner:
@nektro this is probably a regression from the node:http agent changes

[...]

PR #6149, opened 2023-09-28
docs: update image astro.md

Dawntraoz:
### What does this PR do?

Fix my other PR 😅
The link wasn't the right one, my bad, now is right 🥲 So sorry, I owe you a coffee if you are in Amsterdam one day!

- [X] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #15160, opened 2024-11-15
node:https: fix prototype chain of Agent

nektro:
fixes https://github.com/oven-sh/bun/issues/14184
fixed a few others while i was here

code

```js
import { request } from "node:https";

request("https://example.com/", { agent: false }).end();
```

before

```
❯ bun index.js 
ConnectionRefused: Unable to connect. Is the computer able to access the url?
 path: "https://example.com:80/"

Bun v1.1.33 (macOS arm64)
```

after

```
❯ ~/src/bun2/build/debug/bun-debug index.js 
```

### DIFF
--- a/test/js/node/http/node-http.test.ts
+++ b/test/js/node/http/node-http.test.ts
@@ -2390,3 +2390,7 @@ it("must set headersSent to true after headers are sent when using chunk encoded
     server.close();
   }
 });
+
+it("should work when sending https.request with agent:false", () => {
+  https.request("https://example.com/", { agent: false }).end();
+});
--- a/test/js/node/http/node-http.test.ts
+++ b/test/js/node/http/node-http.test.ts
@@ -2391,6 +2391,12 @@ it("must set headersSent to true after headers are sent when using chunk encoded
   }
 });
 
-it("should work when sending https.request with agent:false", () => {
-  https.request("https://example.com/", { agent: false }).end();
+it("should work when sending https.request with agent:false", async () => {
+  const { promise, resolve, reject } = Promise.withResolvers();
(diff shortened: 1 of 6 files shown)


robobun:
@nektro, your commit bce5f5838b5058b5d3ab1cfcfda2aa31e089a508 is building: [`#6149`](https://buildkite.com/bun/bun/builds/6149)

test/js/node/http/node-http.test.ts
```diff
@@ -2390,3 +2390,7 @@ it("must set headersSent to true after headers are sent when using chunk encoded
     server.close();
   }
 });
+
+it("should work when sending https.request with agent:false", () => {
+  https.request("https://example.com/", { agent: false }).end();
```
cirospaciari, line 2395: this test dont fail in revision `1.1.34+5e5e7c60f`

cirospaciari (changes requested):
Looks good but we should tweak the test to fail looks like is passing when I run it.

Screenshot 2024-11-14 at 18 12 41

test/js/node/http/node-http.test.ts
```diff
@@ -2390,3 +2390,7 @@ it("must set headersSent to true after headers are sent when using chunk encoded
...
+});
```
cirospaciari:
```suggestion
it("should work when sending https.request with agent:false", async () => {
  const { promise, resolve, reject } = Promise.withResolvers();
  const client = https.request("https://example.com/", { agent: false });
  client.on("error", reject);
  client.on("close", resolve);
  client.end();
  await promise;
});

```

cirospaciari (approved):
lets merge after CI

Issue #14213, opened 2024-09-27
Syntax Error in String.raw


lmachens:
### What version of Bun is running?

1.1.29

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Run this code:

```js
console.log(String.raw`Hello\user`);
```

### What is the expected behavior?

If you run it with Node.js, it logs `Hello\user`  The `u` is not escaped and you see the `\`.

### What do you see instead?

```
1 | console.log(String.raw`Hello\user`);
                          ^
error: Syntax Error!!
    at C:\dev\the-hidden-gaming-lair\test.js:1:23
    ```

A syntax error, because bun tries to escape Unicode sequence

**Comment by Jarred-Sumner** [comment]
This is a lexer bug. We are validating the input too much in the lexer.

**Comment by nektro** [comment]
this blocks running eslint's own test suite

<details>
<summary> debug log </summary>

```
error: exec: 
 Exception during run: 200 |             errors: [{ messageId: "nested", data: { bref: String.raw`/1`, group: String.raw`(/1)`, otherGroups: "" }, type: "NewExpression" }]
                                                                          ^
error: Syntax Error!!
    at //tests/lib/rules/no-useless-backreference.js:200:69

      at error (//node_modules/shelljs/src/common.js:110:27)
      at execSync (//node_modules/shelljs/src/exec.js:120:12)
      at _exec (//node_modules/shelljs/src/exec.js:223:12)
      at //node_modules/shelljs/src/common.js:335:23
      at //Makefile.js:582:18
      at //node_modules/shelljs/make.js:36:40
      at //Makefile.js:609:12
      at //node_modules/shelljs/make.js:36:40
      at //node_modules/shelljs/make.js:48:23
      at forEach (1:11)

Bun v1.1.30-debug+cf6304222 (macOS arm64)
error: script "test" exited with code 1
```

</details>

RiskyMH:
#15009 seems to have fixed this!

Issue #6179, opened 2023-09-29
Issue with building native dependencies when installing `canvas`


sonigeez:
### What version of Bun is running?

1.0.0

### What platform is your computer?

Darwin 22.5.0 arm64 arm

### What steps can reproduce the bug?

1. Run `bun install canvas`
2. Observe that the installation completes without any errors
3. Attempt to use the canvas module in a Node.js script
4. Observe the failure due to missing native dependencies

### What is the expected behavior?

The necessary build step to compile the native addon modules should be executed during installation, and the canvas module should work as expected.

### What do you see instead?

The build step is not executed, leading to missing native dependencies and the canvas module not working as expected.

### Additional information

i.g bun probably ignoring `prebenchmark` script in `package.json` file

Electroid:
Duplicate of #4959 
Related to #5835

Jarred-Sumner:
Fixed by @dylan-conway and @paperdave in #7132

This will be part of the Bun v1.0.17 release, which ships in a few hours from the time of writing.

In the meantime, you can run `bun upgrade --canary`

PR #15163, opened 2024-11-15
Make E.String always ascii or utf-8, never utf-16

pfgithub:
### What does this PR do?

- Fixes a bug where `("0" + "1") == 0` was true
- Supports folding `"\x00" + "1"` to `"\x001"`
- Supports folding `+"§"` to `NaN`
- Supports folding unicode strings:
   - `"æ" + "™"` to `"æ™"`
   - `"\uD800" + "\uDF34"` to `"𐌴"`.
- Fixes invalid utf-8 in a string did not correctly convert to the replacement character
- Fixes #14213 
- Fixes `let abc\u0070 = 5` would define a variable called "p" rather than "abcp"
- Disallows `` `\8` `` because numeric escapes are not allowed in untagged template literals
- Fixes a bug where the utf-8 sequence (4)(cont)(cont)(cont)(ascii) would parse incorrectly, eg 
  - 0b11110010 0b10101100 0b10110111 0b00101011 would parse as "\u{ACDEB}" rather than "���+" (or "�+")

Reworks E.String

Previously:

- For javascript, E.String was ascii or wtf-16, and constant folding `"a"[0]` and `"a".charCodeAt(0)` depended on not having utf-8.
- For JSON, E.String was always utf-8 with no rope segments. Lots of code accessed `string.data` directly.
- At least one call site in javascript forgot to collapse the rope string before using it (why `("0" + "1") == 0` gave the wrong answer)

Now:

- In javascript, E.String is wtf-8 or a wtf-8 rope string.
- In json, E.String is always wtf-8 and never a rope string
- It is now more difficult to access an uncollapsed rope segment by accident (`.e_string.data` is now `.e_string._value_or_segment_value`)

TODO:

benchmark is in path `~/Dev/Node/temp/generated/fdae05571e21265e7f76681435cbed48/tmp`

performance:
 - [x] must remove strings.isValidUTF8 check. this can be done by making unicode special and then after seeing a unicode character once, next
        itertation it doesn't have to be special anymore and we can go back to strings.isValidUTF8
 - [ ] text_.len < 64 doesn't have a significant effect
 - [ ] try arraylist.toOwnedSlice() over allocator.dupe() in parseStringCharacters

code:
- [x] remove `fn decodeEscapeSequences`

 correctness:
- [x] fix is_all_ascii is only set if a non-ascii byte is found but it also needs to be set if a non-acii codepoint is added in escape parsing
 - [x] implement the looser string parsing for tagged template literals
 - [ ] add more tests
 - [x] compare benchnark bundle output before and after string parsing changes, make sure it is identical
 - [ ] make sure json parsing works right and disallows things it shouldn't allow
 - [x] run parser tests (https://github.com/tc39/test262-parser-tests) and add them to automated testing
   - [x] pass, todo, or skip all tests
   - [x] use test.skip rather than not adding the test for skip tests
   - [x] add so that CI runs it
   - [x] make sure CI runs with `--todo` flag
 - [x] run json tests (https://github.com/nst/JSONTestSuite). re-emit and run again and make sure the result is identical.
   - [x] pass, todo, or skip all tests
   - [x] add so that CI runs it
   - [x] make sure CI runs with `--todo` flag
 - [x] eliminate lexer.peek(); it's used exactly once and implemented wrong
 - [x] nextCodepointSlice() doesn't validate the codepoint - an invalid codepoint should return a 1-len slice rather than eg 3-len (ie `c3 c3` should nextCodepointSlice to `[c3]` not [c3, c3]`)
 - [ ] Should octal escapes in strings be allowed? In 1.1.34 they half-worked (`\32` syntax error but `\324` ok)
 - [x] make sure bundler_string_2.test.ts runs with `--todo` flag
- [x] js_printer.zig writePreQuotedString `\u{{{d}}}` -> `\u{{{X}}}`
- [x] remove "decodeWTF8RuneT", "wtf8ByteSequenceLength", "codepointSize"
- [ ] audit "NewCodePointIterator", "CodepointIterator", "UnsignedCodepointIterator" - these have been changed to not stop on a null byte, make sure the call sites expect that. consider removing these iterators and replacing them with a much simpler one.
  - If keeping, nextCodepointSlice has an incorrect implementation, nextCodepoint asserts utf-8 (not wtf-8), peek is wrong, and the api seems overcomplicated?
- [ ] compare "PackedCodepointIterator" vs regular codepoint iterator
- [ ] remove "std.unicode."
- [ ] compare `strings.wtf8Sequence()`, `strings.encodeWTF8RuneT()`, `strings.unicode.encodeWtf8WithInvalid()` and deduplicate
- [x] identify and fix performance regression in 7dae15b90c9d26c26d7719b12cfd04e1d3babbb0 and 65a8dcc5ea59ce3a9fb9cf786d63266513f80217
- [ ] See if there were any bugs around U+FFFD handling in previous bun versions and if they have been fixed (there is probably a case of ? -> ???)
  - Maybe in toml
- [ ] See if there were any bugs around null byte handling in previous bun versions (there is probably a case of null -> ?)
- [x] Add a test for 0b11110010 0b10101100 0b10110111 0b00101011
- [ ] Compare performance on windows because it is probably worse due to the added assertions
- [x] Pass tests
  - [  ] `esbuild/default.test.ts`, specifically `default/RequireShimSubstitutionBrowser`
- [x] E.String size optimizations:
  - [x] Make sure the union tag and is_from_template_string can share the same 8 bytes, rather than each taking 8 bytes
- [ ] Check if this fixes a bug: 
   ```
   import {$} from "bun";
   const enc = t => new TextEncoder().encode(t);
   
   await Bun.write("./emitted.js", new Uint8Array([
   	...enc("console.log('"),
   	0, 0b1111_0001, 0b10_000000,
   	...enc("');\n\n\n\nthrow new Error('here')\n"),
   ]));
   
   await $`${Bun.argv[0]} ./emitted.js`; // before: (lines stop getting split after null byte)
   ```
- [ ] esbuild/default.test.ts `RequireShimSubstitutionBrowser` seems to be broken? In node 23, it outputs default and module.exports. But shouldn't it be converting the esm import() to require()?
- [x] pass tests:
  - [x] test262-parser-tests.with_todo.test.ts
  - [x] JSONTestSuite.with_todo.test.ts
  - [x] bundler_string_2.with_todo.test.ts on windows
- [ ] rename latin1 to `windows_1252`. jsc 8 bit strings are just raw codepoints but max 255


robobun:
@pfgithub, your commit e5955261e75983db318a37942f1aeaa7227831f4 is building: [`#6179`](https://buildkite.com/bun/bun/builds/6179)

src/string_immutable.zig
```diff
@@ -6746,29 +6448,245 @@ pub fn splitFirstWithExpected(self: string, comptime expected: u8) ?[]const u8 {
...
+                    if (s0 >= 0x80) break :failure;
+                    return .{ .codepoint = s0, .advance = 1 };
+                },
+                .latin1 => {
+                    if (s0 < 0x80) return .{ .codepoint = s0, .advance = 1 };
+                    return .{ .codepoint = latin1ToCodepointAssumeNotASCII(s0, i32), .advance = 1 };
+                },
+                .utf8_assert_no_invalid, .utf8_replace_invalid, .wtf8_assert_no_invalid, .wtf8_replace_invalid => {
+                    const T = i32;
+                    const len: u3 = switch (s0) {
```
Jarred-Sumner: Zig's non^2 integers are super slow. Let's do u8 or even usize.

src/string_immutable.zig
```diff
@@ -6746,29 +6448,245 @@ pub fn splitFirstWithExpected(self: string, comptime expected: u8) ?[]const u8 {
...
+    pub fn codepointIterator(comptime encoding: unicode.Encoding, slice: []const encoding.Unit()) unicode.CodepointIterator(encoding) {
+        if (Environment.allow_assert) bun.assert(isValid(encoding, slice));
+        return .{ .str = slice, .i = 0, .c = -2, .width = 0 };
+    }
+    pub fn CodepointIterator(comptime encoding: unicode.Encoding) type {
+        return struct {
+            const Iterator = @This();
+            str: []const encoding.Unit(),
+            i: usize,
+            width: u3,
```
Jarred-Sumner:
```suggestion
            width: usize,
```

src/string_immutable.zig
```diff
@@ -4293,88 +4226,6 @@ pub fn indexOfNewlineOrNonASCIICheckStart(slice_: []const u8, offset: u32, compt
...
-        if (char > 127 or char < 0x20 or char == '\n' or char == '\r' or char == '"') {
-            return true;
-        }
-    }
-
-    return false;
-}
-
-/// JSON escape
-pub fn indexOfNeedsEscape(slice: []const u8, comptime quote_char: u8) ?u32 {
```
Jarred-Sumner, line 4334: There's a lot of optimizations here being deleted...

pfgithub: These are replaced by `js_lexer.indexOfInterestingCharacterInString`

PR #15167, opened 2024-11-15
Added returning of original path from Bun.build

### LINKED ISSUE #15033: `Bun.build` return the original path of entry point

With the following build,
```ts
    entrypoints: ['index.js', 'second.js'],
    naming: '[hash].[ext]'
})
```

This can return the following object 
```
result: {
  outputs: [
    BuildArtifact (entry-point) {
      path: "./s0d6tbms.js",
      loader: "ts",
      kind: "entry-point",
      hash: "s0d6tbms",
      Blob (86 bytes) {
        type: "text/javascript;charset=utf-8"
      },
      sourcemap: null
    }, BuildArtifact (entry-point) {
      path: "./x2hwc40c.js",
      loader: "ts",
      kind: "entry-point",
      hash: "x2hwc40c",
      Blob (86 bytes) {
        type: "text/javascript;charset=utf-8"
      },
      sourcemap: null
    }
  ],
  success: true,
  logs: [],
}
```

Because the order is not respected, if I added `naming` for the input, it would be impossible to know which one came in first, and from where.  

The feature I am proposing to solve the indeterminate path problem is adding the initially provided entrypoint, if the output is of kind `entry-point`

```
BuildArtifact (entry-point) {
    path: "./x2hwc40c.js",
    loader: "ts",
    kind: "entry-point",
    hash: "x2hwc40c",
    Blob (86 bytes) {
        type: "text/javascript;charset=utf-8"
    },
    originalPath: "./index.js"
    sourcemap: null
}
```

The alternative is to run Bun.build on each entry point separately. But this might cause code duplication and is slower. Additionally it makes it so the paths aren't resolved relative to one another.

BjornTheProgrammer:
### What does this PR do?

This PR fixes #15033, by adding the `sourcefile` attribute on BuildArtificat. It remains as `null` if the output is not an `entry-point`

- [x] Code changes

I ran `bun-debug test bundle`
I added a test to check for the `sourcefile` attribute, and added onto some tests to do the same.


BjornTheProgrammer:
Some things maybe should be considered, such as whether chunks should also have a `sourcefile` if they existed. Additionally, this code makes it so that `sourcefile` doesn't display if the type isn't an `entry-point`. Maybe that should be changed.

src/options.zig
```diff
@@ -2196,14 +2196,8 @@ pub const OutputFile = struct {
...
-
-                var build_output = bun.new(JSC.API.BuildArtifact, .{
-                    .blob = JSC.WebCore.Blob.initWithStore(file_blob, globalObject),
-                    .hash = this.hash,
-                    .loader = this.input_loader,
-                    .output_kind = this.output_kind,
-                    .path = bun.default_allocator.dupe(u8, copy.pathname) catch @panic("Failed to allocate path"),
-                });
+                const source_file = if (this.output_kind != .@"entry-point") "" else bun.default_allocator.dupe(u8, this.src_path.text) catch @panic("Failed to allocate source_path");
+                var build_output = bun.new(JSC.API.BuildArtifact, .{ .blob = JSC.WebCore.Blob.initWithStore(file_blob, globalObject), .hash = this.hash, .loader = this.input_loader, .output_kind = this.output_kind, .source_file = source_file, .path = bun.default_allocator.dupe(u8, copy.pathname) catch @panic("Failed to allocate path") });
```
paperclover: please add trailing comma to take this off of one line

paperclover (commented):
implementation looks reasonable. i want to get other people's opinions on the field name `sourcefile`. maybe it should be named `entrypoint`, since it is the pre-resolved text and not the actual path to the file that it built.

src/options.zig
BjornTheProgrammer: Done!

BjornTheProgrammer:
I've rebased and reran `bun-debug test bundler`. All tests seem to be passing on my machine.

gwer:
This is much needed PR, thank you. But there is one important case that is not reflected in the tests, so I am not sure if it will work correctly.

I mean something like this:
```
{
  "style.module.css": `
    .button {
      width: 100px;
    }
  `,
  "index.ts": `
    import styles from './style.module.css';

    console.log(styles);
  `,
}
```

With default entry-naming `bun build ./index.ts` will generate `index.js` and `index.css`. But with `--entry-naming [name]-[hash].[ext]` they transforms into `index-fwveabcx.js` and `index-gf5q0hss.css` for example, and there is no way to determine dependency between generated css and its entrypoint.

Does your solution solve this problem too?

nektro:
needs to be rebased again, happy to followup review

BjornTheProgrammer:
> With default entry-naming `bun build ./index.ts` will generate `index.js` and `index.css`. But with `--entry-naming [name]-[hash].[ext]` they transforms into `index-fwveabcx.js` and `index-gf5q0hss.css` for example, and there is no way to determine dependency between generated css and its entrypoint.
> 
> 
> 
> Does your solution solve this problem too?

I believe it does, but I will have to verify.

BjornTheProgrammer:
> needs to be rebased again, happy to followup review

Thank you for the reply! I'm willing to rebase, but it will be in a few weeks, as I'm very busy in the current moment.

PR #15178, opened 2024-11-16
musl: fix 'bun upgrade'

nektro:

before

```
error: 'bun upgrade' is unsupported on systems without ld

You are likely on an immutable system such as NixOS, where dynamic
libraries are stored in a global cache.

Please use your system's package manager to properly upgrade bun.
```

after

```
Downloading... error: Canary builds are not available for this platform yet

   Release: https://github.com/oven-sh/bun/releases/tag/canary
  Filename: bun-linux-x64-musl.zip
```

^ will work once there is a first canary release of musl


robobun:
@nektro, your commit 306237d5c1289fe3ecdfb99f6434a23a12cb98c7 is building: [`#6244`](https://buildkite.com/bun/bun/builds/6244)

PR #15184, opened 2024-11-16
ci: bootstrap: raise fd limit

nektro:

this should fix the musl segfault in `test/js/bun/http/serve.test.ts`

`us_create_timer` is allowed to return a null pointer and our code doesn't currently handle this
will send real fix in followup


robobun:
@nektro, your commit 835e5ea8b25e42dbb578b8e4a6d3d2dd7c9cdb9b is building: [`#6268`](https://buildkite.com/bun/bun/builds/6268)

Issue #6273, opened 2023-10-03
Support `metafile` or equivalent in Bun.build()


robobun:
in esbuild, there is a metafile option to get access to stuff about the build in plugins, is there an equivalent in bun?

https://esbuild.github.io/api/#metafile

_Originally reported on Discord: [`bun build metafile`](https://discord.com/channels/876711213126520882/1158501976699584673/1158501976699584673)_

paulzakin:
Agreed - this would be awesome to have - and I feel like a vital feature in making the Bun builder a viable alternative to esbuild!

adamscybot:
I see this is now in Bun since [1.3.6](https://bun.com/blog/bun-v1.3.6) and is documented [here](https://bun.com/docs/bundler#markdown-metafile). Though I note its on the build output object, not inside plugins (?). But you can just post process the `metafile` property to probably achieve what you want.

If so perhaps it can be closed @cirospaciari

Note some TS types are missing on the JS build API for the string/object options. I have raised a PR to fix that over here https://github.com/oven-sh/bun/pull/27055.

PR #15187, opened 2024-11-16
ci: bootstrap: define shell for buildkite user

nektro:

this should fix the musl failure in `test/js/node/child_process/child_process.test.ts`
as it's being created with `machine.mjs`/`bootstrap.sh`


robobun:
@nektro, your commit e8f13fa06653f81b1c06dc9a513d8d9ce2a0fe79 is building: [`#6273`](https://buildkite.com/bun/bun/builds/6273)

nektro:
logs confirm `$USER` is `"buildkite-agent"` but `$SHELL` is still `undefined`

PR #15188, opened 2024-11-16
zig: always set a reasonable NOFILE rlimit max

nektro:

this should fix the musl segfault in `test/js/bun/http/serve.test.ts`

`us_create_timer` is allowed to return a null pointer and our code doesn't currently handle this
will send real fix in followup

additionally,

closes https://github.com/oven-sh/bun/pull/15184

setting the ulimit manually should never be necessary but it appeared to be so because empirically the NOFILE rlimit was coming back as 1024/4096 on alpine and we were prematurely getting EMFILE in the course of normal operation


robobun:
@nektro, your commit eca4c9720cf787ce0d99badb74b24d7c9f6fd5cb is building: [`#6278`](https://buildkite.com/bun/bun/builds/6278)

Issue #6288, opened 2023-10-04
`Request` in `bun test` is not allowing `headers` as RequestInit in bun 1.0.4


aralroca:
### What version of Bun is running?

1.0.4+745b6b94ee56cad24d475799690cc9a89957d15b

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

Execute this test:

```ts
it('SHOULD FAIL', () => {
  const req = new Request("https://example.com/ua", {
    headers: {
      "Accept-Language": "es-ES,es;q=0.9,en-US;q=0.8,en;q=0.7",
      Cookie: "DEFAULT_LOCALE=ru",
    },
  })

  expect(req.headers.get("Cookie")).toBe("DEFAULT_LOCALE=ru")
  expect(req.headers.get("Accept-Language")).toBe("es-ES,es;q=0.9,en-US;q=0.8,en;q=0.7")
})
```

In Bun 1.0.3 pass ✅.

In Bun 1.0.4 ❌ :

```sh
 expect(req.headers.get("Cookie")).toBe("DEFAULT_LOCALE=ru")
          ^
error: expect(received).toBe(expected)

Expected: "DEFAULT_LOCALE=ru"
Received: null
```

### What is the expected behavior?

To pass the test as before.

### Additional information

Is possible to workaround to pass the test changing to:

```ts
it('SHOULD WORK AS A WORKAROUND', () => {
  const req = new Request("https://example.com/ua")

  req.headers.set("Cookie", "DEFAULT_LOCALE=ru")
  req.headers.set("Accept-Language", "es-ES,es;q=0.9,en-US;q=0.8,en;q=0.7")

  expect(req.headers.get("Cookie")).toBe("DEFAULT_LOCALE=ru")
  expect(req.headers.get("Accept-Language")).toBe("es-ES,es;q=0.9,en-US;q=0.8,en;q=0.7")
})
```

aralroca:
I close it because I see that it is really how the Request works in the Web API, probably the weird thing was that the old one worked.

PR #15194, opened 2024-11-17
Introduce InspectorLifecycleAgent and InspectorTestReporterAgent

Jarred-Sumner:
### What does this PR do?

This adds two new APIs to the WebKit Inspector Protocol when used in Bun.

This lets you implement a test reporter on top of the WebKit inspector protocol. This also unlocks reporting runtime errors in editor diagnostic APIs without the debugger attached.

The API will change a little more before this merges. The diagnostics need to already be stacktraced since the client may not have the debugger connected.

This will need some tests before it merges.

### DIFF
--- a/src/bun.js/bindings/InspectorLifecycleAgent.cpp
+++ b/src/bun.js/bindings/InspectorLifecycleAgent.cpp
@@ -1,0 +1,144 @@
+#include "InspectorLifecycleAgent.h"
+
+#include <JavaScriptCore/InspectorFrontendRouter.h>
+#include <JavaScriptCore/InspectorBackendDispatcher.h>
+#include <JavaScriptCore/JSGlobalObject.h>
+#include <wtf/text/WTFString.h>
+#include <JavaScriptCore/ScriptCallStackFactory.h>
+#include <JavaScriptCore/ScriptArguments.h>
+#include <JavaScriptCore/ConsoleMessage.h>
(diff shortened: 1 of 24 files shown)


robobun:
@Jarred-Sumner, your commit 7ca95b1ed853efdda0e138bdb7981cffb444a290 is building: [`#6288`](https://buildkite.com/bun/bun/builds/6288)

src/bun.js/bindings/InspectorLifecycleAgent.cpp
```diff
@@ -0,0 +1,130 @@
...
+    Ref<JSON::ArrayOf<String>> sourceLines = JSON::ArrayOf<String>::create();
+
+    for (size_t i = 0; i < exception.stack.source_lines_len; i++) {
+        sourceLines->addItem(exception.stack.source_lines_ptr[i].toWTFString());
+    }
+
+    for (size_t i = 0; i < exception.stack.frames_len; i++) {
+        ZigStackFrame* frame = &exception.stack.frames_ptr[i];
+        lineColumns->addItem(frame->position.line_zero_based + 1);
+        lineColumns->addItem(frame->position.column_zero_based + 1);
```
alii, line 111:
```suggestion
        lineColumns->addItem(frame->position.column_zero_based + 1);
        urls->addItem(frame->source_url.toWTFString());
```

PR #15203, opened 2024-11-17
promisify implementation

### LINKED ISSUE #15201: setTimeout hasn't promisify implementation

1.1.34+5e5e7c60f

Darwin 24.0.0 arm64 arm

require('util').promisify(setTimeout)(100, 'ok').then(console.log);

node test.js
>ok

### What do you see instead?

Image

### Additional information

I've revised  https://github.com/oven-sh/bun/issues/4361 and https://github.com/oven-sh/bun/issues/7148. But it seems in the current `bun` version it was not fixed.

ippsav:
### What does this PR do?

Fixes #15201

### DIFF
--- a/src/js/internal/promisify.ts
+++ b/src/js/internal/promisify.ts
@@ -1,76 +1,87 @@
 const kCustomPromisifiedSymbol = Symbol.for("nodejs.util.promisify.custom");
 const kCustomPromisifyArgsSymbol = Symbol("customPromisifyArgs");
 
-function defineCustomPromisify(target, callback) {
+function defineCustomPromisify(target: Function, callback: Function) {
   Object.defineProperty(target, kCustomPromisifiedSymbol, {
     value: callback,
-    __proto__: null,
+    writable: true,
     configurable: true,
   });
-
   return callback;
 }
 
-function defineCustomPromisifyArgs(target, args) {
+function defineCustomPromisifyArgs(target: Function, args: string[]) {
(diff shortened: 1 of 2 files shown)


Jarred-Sumner:
This makes the util.promisify function a little bit slower on every call unrelated to setTimeout or setInterval or timers.

Instead of doing that, can we make this work in ZigGlobalObject.cpp when we define the setTimeout and setInterval functions?

Jarred-Sumner (changes requested):
This makes the util.promisify function a little bit slower on every call unrelated to setTimeout or setInterval or timers.

Instead of doing that, can we make this work in ZigGlobalObject.cpp when we define the setTimeout and setInterval functions?

ippsav:
@Jarred-Sumner, I'll need some time to properly understand the internals of JSC and how all the components interact with each other before implementing this change. Would it be possible to merge this current PR to address the immediate issue, and then I can create a follow-up PR with your suggested optimization once I've familiarized myself better with all what's JSC related ?

tannal:
Hi, I think all we need to do is set `setTimeout[Symbol.for('nodejs.util.promisify.custom')]` to `setTimeoutPromise` when we define the setTimeout function.

Same for setInterval.
```js
console.log(setTimeout[Symbol.for('nodejs.util.promisify.custom')])
```
Run this script with node and bun, and you can see the difference.

Jarred-Sumner:
Closing in favor of #15230

Issue #6300, opened 2023-10-04
File not found when running `bun --config bunfig.toml`


FlorianKoerner:
*Maintainer note:* This is currently an issue in the way bun parses optional arguments. For example you can `bun --config` and it will default to `bunfig.toml` as value, this makes it hard to know if the arg after is file to run or config.

### What version of Bun is running?

1.0.3+25e69c71e70ac8a0a88f9cf15b4057bd7b2a633a

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. `bun init`
2. Create a `bunfig.toml` file. The content does not seem to matter, but for testing purposes: `telemetry = false`
3. Run `bun -c bunfig.toml .`

---

Example output on my machine:

```
$ pwd                                                           
/Users/f.koerner/Development/dicebear/api

$ test -e bunfig.toml && echo file exists || echo file not found
file exists

$ bun -c bunfig.toml . 
error: file not found "bunfig.toml"

$ bun -c /Users/f.koerner/Development/dicebear/api/bunfig.toml .
error: file not found "/Users/f.koerner/Development/dicebear/api/bunfig.toml"

$ sw_vers
ProductName:            macOS
ProductVersion:         13.5.2
BuildVersion:           22G91
```

### What is the expected behavior?

-

### What do you see instead?

error: file not found "/Users/f.koerner/Development/dicebear/api/bunfig.toml"

Hanaasagi:
I think subcommand might need to come before the options.

```Bash
bun . -c bunfig.toml
```

FlorianKoerner:
In this order no more error, but it still does not load the configuration.

Hanaasagi:
I think there are something wrong in parsing command line. Temporary workaround is `bun --config=custom-bunfig.toml .`

----

Some debugging information

```Bash
#!/bin/bash

echo '[loader]
".toml" = "text"' \
    > custom-bunfig.toml
echo 'import data from "./custom-bunfig.toml"; console.log(data);' > index.ts

set -x
bun . -c custom-bunfig.toml  # not work
bun . -c=custom-bunfig.toml  # not work
bun -c=custom-bunfig.toml . # not work
bun --config=custom-bunfig.toml . # work
bun --config custom-bunfig.toml . # not work, file not found error
```

expect output is 
```
[loader]
".toml" = "text"
```

not 

```
{
  loader: {
    ".toml": "text"
  }
}

```

WingLim:
[...]

PR #15210, opened 2024-11-18
Introduce `deathSig` option in Bun.spawn

Jarred-Sumner:
### What does this PR do?

Introduce `deathSig` option in `Bun.spawn`

This is a Linux-only option that lets you automatically make spawned processes get killed when Bun exits (as long as they don't set detached: true, or call setsid()).

This enables this option in `bun run <package.json script>`, `bun install` for lifecycle scripts, and almost everywhere else we spawn processes. 

WIP.

### DIFF
--- a/src/bun.js/api/bun/process.zig
+++ b/src/bun.js/api/bun/process.zig
@@ -970,6 +970,11 @@ pub const PosixSpawnOptions = struct {
     argv0: ?[*:0]const u8 = null,
     stream: bool = true,
 
+    /// Linux-only.
+    ///
+    /// Send a signal to the parent process when the child process exits.
+    deathsig: bun.SignalCode = @enumFromInt(0),
+
     /// Apple Extension: If this bit is set, rather
     /// than returning to the caller, posix_spawn(2)
     /// and posix_spawnp(2) will behave as a more
@@ -1040,6 +1045,9 @@ pub const WindowsSpawnOptions = struct {
     argv0: ?[*:0]const u8 = null,
     stream: bool = true,
     use_execve_on_macos: bool = false,
+
+    /// Linux-only. Does nothing on Windows.
+    deathsig: bun.SignalCode = @enumFromInt(0),
 
     pub const WindowsOptions = struct {
         verbatim_arguments: bool = false,
@@ -1226,6 +1234,10 @@ pub fn spawnProcessPosix(
 
     if (options.detached) {
         flags |= bun.C.POSIX_SPAWN_SETSID;
+    } else if (Environment.isLinux) {
+        if (@intFromEnum(options.deathsig) > 0) {
+            actions.deathsig = @intFromEnum(options.deathsig);
+        }
     }
 
     if (options.cwd.len > 0) {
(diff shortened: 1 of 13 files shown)


robobun:
@Jarred-Sumner, your commit 9f73c4d5753372e1b7cfb4de682bcbaeacc3bfdb is building: [`#6300`](https://buildkite.com/bun/bun/builds/6300)

src/bun.js/api/bun/process.zig
```diff
@@ -970,6 +970,11 @@ pub const PosixSpawnOptions = struct {
     argv0: ?[*:0]const u8 = null,
     stream: bool = true,
 
+    /// Linux-only.
+    ///
+    /// Send a signal to the parent process when the child process exits.
```
paperclover, line 975: this comment is reversed from the typescript documentation

src/bun.js/api/bun/process.zig
```diff
@@ -970,6 +970,11 @@ pub const PosixSpawnOptions = struct {
...
+    deathsig: bun.SignalCode = @enumFromInt(0),
```
paperclover, line 976: bad practice doing hiding an invalid state in a valid enum. should zero be added to the enum as `none`?

src/bun.js/api/bun/process.zig
```diff
@@ -1226,6 +1234,10 @@ pub fn spawnProcessPosix(
 
     if (options.detached) {
         flags |= bun.C.POSIX_SPAWN_SETSID;
+    } else if (Environment.isLinux) {
+        if (@intFromEnum(options.deathsig) > 0) {
```
paperclover, line 1238: `options.deathsig.valid()`?

src/bun.js/api/bun/spawn.zig
```diff
@@ -299,6 +300,7 @@ pub const PosixSpawn = struct {
     const BunSpawnRequest = extern struct {
         chdir_buf: ?[*:0]u8 = null,
         detached: bool = false,
+        deathsig: i32 = 0,
```
paperclover, line 303: `c_int`

src/cli/bunx_command.zig
```diff
@@ -637,6 +637,7 @@ pub const BunxCommand = struct {
             .stderr = .inherit,
             .stdout = .inherit,
             .stdin = .inherit,
+            .deathsig = .SIGTERM,
```
paperclover, line 640: should our default be .SIGTERM?

paperclover (changes requested):
in addition to review nitpicks, this feature, _as is_, will worsen the footgun because it only solves this problem on linux, and not on windows or macOS. someone will see this, briefly check the first paragaph, see it works on linux, and declare the problem solved. in reality, the problem persists on the other operating systems.

that being said, that shouldn't block this pr. the other platforms can be done in another pr.

alternate name suggestion: `killOnParentExit: true | string` (i think having an easy `true` aliasing for KILL would be helpful, it would also be the only possible way to emulate this on windows as it does not have kill signals)

Issue #15204, opened 2024-11-17
Running bun amd64 in docker under platform rosetta2 translation breaks Bun.serve


lil5:
### What version of Bun is running?

1.1.34+5e5e7c60f

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

1. create a bun project and add the following to index.ts

command: `bun init; bun i`

file: `index.ts`
```ts
Bun.serve({
  fetch(req) {
    return new Response("Bun!");
  },
});
```
2. create a docker file 

file: `Dockerfile`
```dockerfile
FROM oven/bun AS base
WORKDIR /app

COPY bun.lockb . 
COPY package.json . 

# Install dependencies
RUN bun install --frozen-lockfile

COPY . . 

EXPOSE 3000/tcp
ENTRYPOINT [ "bun", "run", "./index.ts" ]
```

file: `.dockerignore`
```
/node_modules
```

3. create a docker-compose file
file: `docker-compose.yml`
```yml
services:
  app:
    platform: linux/amd64
    build:
      context: .
    ports:
      - 127.0.0.1:3000:3000
    restart: unless-stopped
    environment:
      - PORT=3000
```

4. Build and run `docker compose up -d --build`
5. Create a request to bun `wget --spider http://localhost:3000/`, run this multiple times and it will stop working after a short while.

### What is the expected behavior?

200 responses returning the text "Bun!"

### What do you see instead?

Request without a response, endlessly stuck with no response, no console errors.

### Additional information

Running on M2 15.1 (24B83) with Docker Desktop 4.35.1 (173168)

Kapsonfire-DE:
Is this really bun related? Seems more like an issue for the apple team emulating amd64 on arm64 (m2)

lil5:
As running an amd64 GoLang or NodeJS server via Rosetta2 works just fine in docker, yes I'd say it's most likely to be an underlying problem in Bun rather than that of Rosetta2. 

To be fair I can test this out on a simple Zig Zap server if it's a zig incompatibility.

lil5:
This is effectively a duplicate of https://github.com/oven-sh/bun/issues/15139

Jarred-Sumner:
We no longer provide a macOS x64 binary that supports non-avx CPUs.

On macOS arm64, we recommend using arm64 builds of Bun. If you install via the `curl https://bun./sh/install | bash`  install script, it will automatically do this even in Rosetta.

[...]

PR #15213, opened 2024-11-18
Add notice about not running amd64 under MacOS Silicon

lil5:
Update installation.md

### What does this PR do?

Add a notice in the https://bun.sh/docs/installation#installing page to warn MacOS Silicon to only use arm builds (see #15204 )

- [x] Documentation
- [ ] Code changes


Jarred-Sumner (changes requested):
Thank you for this, but this does not impact enough people to justify a prominent place in the docs.

Somewhere towards the bottom of that page would be fine

Issue #6313, opened 2023-10-05
zeromq module does not work


anttikissa:
### What version of Bun is running?

1.0.4

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Even requiring the zeromq module doesn't work.

To reproduce, run in an empty directory:
```
bun init
bun add zeromq@6.0.0-beta.17
echo "require('zeromq')" > index.ts
bun index.ts
```

### What is the expected behavior?

That the require would work

### What do you see instead?

```
% bun index.ts
1 | (function (id) {"use strict";
                        ^
TypeError: dlopen(/Users/antti/sync/zeromq-test/node_modules/zeromq/prebuilds/darwin-arm64/node.napi.glibc.node, 0x0001): Library not loaded: /Users/runner/arm-target/opt/libsodium/lib/libsodium.23.dylib
  Referenced from: <1DC056E1-92FA-379B-8526-E6E3738194E7> /Users/antti/sync/zeromq-test/node_modules/zeromq/prebuilds/darwin-arm64/node.napi.glibc.node
  Reason: tried: '/Users/runner/arm-target/opt/libsodium/lib/libsodium.23.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/runner/arm-target/opt/libsodium/lib/libsodium.23.dylib' (no such file), '/Users/runner/arm-target/opt/libsodium/lib/libsodium.23.dylib' (no such file), '/usr/lib/libsodium.23.dylib' (no such file, not in dyld cache)
      at internalRequire (:1:20)
      at load (/Users/antti/sync/zeromq-test/node_modules/@aminya/node-gyp-build/index.js:22:9)
      at /Users/antti/sync/zeromq-test/node_modules/zeromq/lib/native.js:6:7
      at globalThis (/Users/antti/sync/zeromq-test/node_modules/zeromq/lib/native.js:6:77)
      at overridableRequire (:1:20)
      at /Users/antti/sync/zeromq-test/node_modules/zeromq/lib/index.js:5:4
      at globalThis (/Users/antti/sync/zeromq-test/node_modules/zeromq/lib/index.js:595:75)
```

### Additional information

First I noticed that Bun tries to load libsodium from non-existing path `/Users/runner/arm-target/opt/libsodium/lib`.

I can get a step further by creating that and linking a homebrew-installed libsodium.23.dylib there:

```
[...]

PR #15217, opened 2024-11-18
[Bun.sql] Support TLS

Jarred-Sumner:
### What does this PR do?

The following works:
```ts
import { getSecret } from "harness";
import { sql as SQL } from "bun";

const TLS_POSTGRES_DATABASE_URL = getSecret("TLS_POSTGRES_DATABASE_URL");

test("tls (explicit)", async () => {
  const sql = new SQL({
    url: TLS_POSTGRES_DATABASE_URL!,
    tls: true,
    adapter: "postgresql",
  });

  const [{ one, two }] = await sql`SELECT 1 as one, '2' as two`;
  expect(one).toBe(1);
  expect(two).toBe("2");
});

test("tls (implicit)", async () => {
  const [{ one, two }] = await SQL`SELECT 1 as one, '2' as two`;
  expect(one).toBe(1);
  expect(two).toBe("2");
});
```

It successfully connected to Neon and with TLS.

There is a test

### DIFF
--- a/packages/bun-usockets/src/crypto/openssl.c
+++ b/packages/bun-usockets/src/crypto/openssl.c
@@ -201,7 +201,7 @@ struct loop_ssl_data * us_internal_set_loop_ssl_data(struct us_internal_ssl_sock
 
 struct us_internal_ssl_socket_t *ssl_on_open(struct us_internal_ssl_socket_t *s,
                                              int is_client, char *ip,
-                                             int ip_length) {
+                                             int ip_length, const char* sni) {
 
   struct us_internal_ssl_socket_context_t *context =
       (struct us_internal_ssl_socket_context_t *)us_socket_context(0, &s->s);
@@ -231,6 +231,10 @@ struct us_internal_ssl_socket_t *ssl_on_open(struct us_internal_ssl_socket_t *s,
   if (is_client) {
     SSL_set_renegotiate_mode(s->ssl, ssl_renegotiate_explicit);
     SSL_set_connect_state(s->ssl);
+
+    if (sni) {
+      SSL_set_tlsext_host_name(s->ssl, sni);
+    }
   } else {
     SSL_set_accept_state(s->ssl);
     // we do not allow renegotiation on the server side (should be the default for BoringSSL, but we set to make openssl compatible)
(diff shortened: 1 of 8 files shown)


robobun:
@Jarred-Sumner, your commit 04295f90123192b9fa58c8f4433a82936e92f9d5 is building: [`#6313`](https://buildkite.com/bun/bun/builds/6313)

packages/bun-usockets/src/crypto/openssl.c
```diff
@@ -231,6 +231,10 @@ struct us_internal_ssl_socket_t *ssl_on_open(struct us_internal_ssl_socket_t *s,
   if (is_client) {
     SSL_set_renegotiate_mode(s->ssl, ssl_renegotiate_explicit);
     SSL_set_connect_state(s->ssl);
+
+    if (sni) {
+      SSL_set_tlsext_host_name(s->ssl, sni);
```
cirospaciari, line 236: normally we do this inside on_open (that happens before the on_handshake) but I think is a OK change, should we do something on server side?

src/js/bun/sql.ts
```diff
@@ -507,6 +569,10 @@ function SQL(o) {
 
 var lazyDefaultSQL;
 var defaultSQLObject = function sql(strings, ...values) {
+  if (new.target) {
```
cirospaciari, line 572: TIL https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target

src/sql/postgres.zig
```diff
@@ -1059,48 +1134,97 @@ pub const PostgresSQLConnection = struct {
...
-        event_loop.enter();
-        defer event_loop.exit();
-        this.flushData();
+        this.start();
+    }
+
+    pub fn onHandshake(this: *PostgresSQLConnection, success: i32, ssl_error: uws.us_bun_verify_error_t) void {
+        debug("onHandshake: {d} {d}", .{ success, ssl_error.error_no });
+        if (success != 1) {
+            if (this.tls_config.reject_unauthorized == 1) {
```
cirospaciari: should we use `BoringSSL.checkServerIdentity` to match fetch and other clients behavior? (checking altname and IP)

cirospaciari: fixed in https://github.com/oven-sh/bun/pull/15217/commits/2fa15c9492854b8fb2b00415f2732bfe15df7f14

cirospaciari (approved):

Issue #6338, opened 2023-10-06
Bun's .env reading causes issues with Vite's .env reading


jorismak:
### What version of Bun is running?

1.0.4+745b6b94ee56cad24d475799690cc9a89957d15b

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

I have a `.env.production` file which overrides `PUBLICPATH`.

So in `.env` I have:
```
PUBLICPATH=/
```

and in a `.env.production` I have:
```
PUBLICPATH=/app
```

But when I run `bun -b run build` to build with vite, it's building for 'production' mode but I still end up with a PUBLICPATH=/.

In a vite.config.js I have something like this:
```
import { defineConfig, loadEnv } from "vite";

export default ({ mode }) => {    
    const loadedEnv = loadEnv(mode, process.cwd(), "");

    return defineConfig({
        base: loadedEnv.PUBLICPATH,
    });
};
```

This was working with node.

### What is the expected behavior?

PUBLICPATH=/app in the list of variables from `loadEnv()`.

### What do you see instead?

PUBLICPATH=/ in the list of variables from `loadEnv()`.

### Additional information

What happens - I think - is that bun reads the .env file beforehand, and thus already sets 'PUBLICPATH' in the 'real' environment.
Then Vite loads the dotEnv files, but doesn't override anything that is already set in process.env.

I have to copy/move `.env.production` to `.env.local` during CI/CD to build it with the correct variable,
or I have to explicitly call the build command with `NODE_ENV=production bun -b run build` which was never needed before 
(NODE_ENV is set in / during Vite somewhere).

ikhedkar-swi:
Same here. As a workaround, explicitly had to pass `NODE_ENV=production` to the build command.

HugoGresse:
> Same here. As a workaround, explicitly had to pass `NODE_ENV=production` to the build command.

setting NODE_ENV=production where not working on my side: `"build": "NODE_ENV=production bunx --bun vite build",`

LoveMeWithoutAll:
I also have a same issue.
But there is a workaround.
This issue have occurred `Only --mode development` , and bun works well if there is only one .env file.

Sollution is below
1. Move all .env file to a specific directory like .env
2. add a copy script to package.json

example is below
`"build:production": "cp ./.env/.env.production . && tsc && bunx --bun vite build --mode production"`

IlyaCherkasov:
[...]

PR #15231, opened 2024-11-19
Fix bundler crash with onLoad plugins on copy-file loaders used on entrypoints

zackradisic:
### What does this PR do?

This fixes a crash in the bundler when:
  - the entry point is a file with a type which should be copied by the bundler (such as `.wasm` or `.css` with `experimentalCss: false`) 
  - an onLoad plugin matches this file

Two problems occurred in this case:
- the entry point was not appending to `additional_files`
- `onLoadAsync` callback was incorrectly appending string to `free_list`


robobun:
@zackradisic, your commit 3ecff6d7bcb361739010d8cb4204b4da025563bc is building: [`#6338`](https://buildkite.com/bun/bun/builds/6338)

Issue #6355, opened 2023-10-07
`pino` is slower in Bun than in Node


floydjones1:
### What version of Bun is running?

1.0.4

### What platform is your computer?

Linux 6.2.0-33-generic x86_64 x86_64

### What steps can reproduce the bug?

Code snippet
```javascript
import pino from 'pino'
const logger = pino()
console.log("Hello via Bun!");

const MAX = 1000000;
const now = Date.now()
for (let index = 0; index <= MAX; index++) {
    const mem = process.memoryUsage.rss()/1024/1024
    if (index % 1000 === 0) {
        logger.info(` memory ${mem}`, index)
    }
}


await new Promise(resolve => setTimeout(resolve, 2_000))

const mem = process.memoryUsage.rss()/1024/1024
console.log(` final memory ${mem} took ${(Date.now() - now)/1000} seconds`) 
```

### What is the expected behavior?

Bun faster than node

### What do you see instead?

Bun slower than node

### Additional information

node final output
```sh
final memory 50.14453125 took 6.024 seconds
```

bun final output
```sh
final memory 60.234375 took 7.928 seconds
```

Electroid:
We'd have to take a look at the `pino` package to see what APIs its using.

floydjones1:
@Electroid appreciate it. I hope this is enough information. 
I use Pino in my production apps and think it's important if it causes some slow down.

Appreciate all the effort from the bun team! 💪🏼

floydjones1:
Just an update with updated versions. 

node 20.10.0
![image](https://github.com/oven-sh/bun/assets/43974214/62097410-93da-4632-840d-0a889351eb87)

bun 1.0.20
![image](https://github.com/oven-sh/bun/assets/43974214/9992a94a-c4d0-4606-815d-92f2d833b069)

floydjones1:
I also ran the basic benchmark that pino has in their repo. https://github.com/pinojs/pino/blob/master/benchmarks/basic.bench.js and just kept `benchPino` `benchPinoMinLength` and `benchPinoNodeStream` with these results: (other logger comparisons were similar)

bun 1.0.20
<img width="373" alt="Screenshot 2023-12-25 at 11 16 31" src="https://github.com/oven-sh/bun/assets/43974214/6a28feee-da70-4db7-b234-b8aabace1fce">

node v18.17.1
[...]

PR #15233, opened 2024-11-19
Fix docs on todo tests

### LINKED ISSUE #8478: `test.todo` not failing when test is already passing

1.0.25+a8ff7be64

Darwin 23.3.0 arm64 arm

The description states that `test.todo` passes as long as the test fails because it is still todo, but that it fails as soon as the test starts to pass. 
Screenshot 2024-01-25 at 13 02 57

```ts
import { expect, test } from 'bun:test';

test.todo('this should not pass because it is no longer todo', () => {
	expect(true).toEqual(true);
});
```

```log
bun test v1.0.25 (a8ff7be6)

src/main.spec.ts:
✗ this should not pass because it is no longer todo

 0 pass
 1 fail
Ran 1 tests across 1 files. [5.00ms]
```

### What do you see instead?

```log
bun test v1.0.25 (a8ff7be6)

src/main.spec.ts:
✎ this should not pass because it is no longer todo

 0 pass
 1 todo
 0 fail
Ran 1 tests across 1 files. [5.00ms]
```

### Additional information

_No response_

pfgithub:
### What does this PR do?

Fixes #8478

Documentation has been out of date for a while about todo tests.

### DIFF
--- a/docs/guides/test/todo-tests.md
+++ b/docs/guides/test/todo-tests.md
@@ -44,10 +44,17 @@ test.todo("unimplemented feature", () => {
 
 ---
 
-If an implementation is provided, it will be executed and _expected to fail_ by test runner! If a todo test passes, the `bun test` run will return a non-zero exit code to signal the failure.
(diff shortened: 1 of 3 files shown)


robobun:
@pfgithub, your commit 17ce4a3619d353649e6eed995563f08a32dcb5fc is building: [`#6355`](https://buildkite.com/bun/bun/builds/6355)

PR #6400, opened 2023-10-10
Documentation for the IPC of Bun.spawn

RaresAil:
### What does this PR do?

Adding the missing documentation for the IPC feature on Bun.spawn

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #15252, opened 2024-11-19
chore: delete some unused functions in meta.zig

190n:
### What does this PR do?

Removes these functions from `src/meta.zig`:

- `OptionalChild`: only used in one place, unclear name (it required a pointer to an optional, not just an optional). Removed and its user was replaced with `@TypeOf`
- `ReturnOfMaybe`, `MaybeResult`: not used
- `ConcatArgs{1,2,4}`: only the `2` variant was used in two places. Those uses have been replaced by tuple concatenation.
- `CreateUniqueTuple`: not used
- `hasStableMemoryLayout`: not used, and wrong (auto-layout structs and unions never have stable layout; not all ints have stable layout ziglang/zig#19755)
- `isScalar`: not used, and wrong (8- and 16-bit integers are scalars)

Checked that it still compiles


robobun:
@190n, your commit d0c454920102534b250f4e5ee531c03b80aaf830 is building: [`#6400`](https://buildkite.com/bun/bun/builds/6400)

Issue #6409, opened 2023-10-11
Support `yarn.lock` (v1) in `bun pm migrate` & `bun install`


Jarred-Sumner:
Tracking issue following up from https://github.com/oven-sh/bun/issues/1751

wojtekmaj:
Note Yarn Classic (v1) and Yarn Berry (v2+) uses completely different lockfile formats. Is this tracking both?

john-griffin:
Looking forward to this. We tried to do yarn.lock -> package-lock.json -> bun. But we lost too much context and couldn't build the apps.

yharaskrik:
This would be amazing and help us incrementally adopt bun

Jarred-Sumner:
> Note Yarn Classic (v1) and Yarn Berry (v2+) uses completely different lockfile formats. Is this tracking both?

This is tracking Yarn v1 because npm download stats suggest more people use it.

wojtekmaj:
@Jarred-Sumner, you cannot rely on npm download stats. Here's why:

https://twitter.com/wojtekmaj91/status/1679944016941527040

> This is an extremely bad way of comparing package managers popularity.
> - Package managers updated more often will get downloaded more often by the same users.
> - Yarn Berry used to sit inside the repo in .yarn/releases. This resulted in one download per repo, and it wasn't from npm.
> - Node Corepack doesn't use npm for all Yarn versions - see source code: https://github.com/nodejs/corepack/blob/main/config.json
> 
> My suggestion would be to count lockfiles and packageManager entries in package.json in OSS repositories instead. Of course, easier said than done.

wojtekmaj:
Actually, this inspired me to measure package managers popularity it as accurately as possible. Here are the results.

https://dev.to/wojtekmaj/package-manager-wars-the-real-picture-e9p

muzuiget:
I hope bun can add support for loading and writing Yarn v1 lock file directly, so we can use bun as a drop-in replacement for Yarn v1, because current workflow and keyboard muscle memory are not easily to change.

```
alias yarn='bun yarn-v1'
yarn install # CRUD the yarn.lock file
```

I think Yarn v1 stills has large users, see the Yarn v4 document [Migrating from 1.x / npm](https://yarnpkg.com/migration/overview), and Alpine Linux only [pack Yarn v1](https://pkgs.alpinelinux.org/package/v3.18/community/x86_64/yarn).

[...]

PR #15260, opened 2024-11-19
assert package manager is allocated once before use

pfgithub:


robobun:
@pfgithub, your commit f53321a525e9e4ea338340f8ee5d345eb22b2270 is building: [`#6409`](https://buildkite.com/bun/bun/builds/6409)

PR #13668, opened 2024-09-01
feat(vscode-extension): test support

samuelgja:
### What does this PR do?

**Bun Vscode extension test support**
It is a little bit inspired by rust vs code extension, where the user can run tests directly in the editor.

I created this functionality a few weeks ago and found it helpful in my projects, so maybe it can be added to the extension, or maybe not :D

- ability to run test directly from vscode opened editor:
Screenshot 2024-09-01 at 12 38 56

- run test for whole file via vscode settings command:
Screenshot 2024-09-01 at 12 41 04

- user defined test file pattern matcher:
Screenshot 2024-09-01 at 12 41 35

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes
### Code changes

- added new test provider [packages/bun-vscode/src/features/tests/index.ts](https://github.com/oven-sh/bun/compare/main...samuelgja:bun:main#diff-a0ef2cf3fa6ec77288d3cf68151458a0b3c787683cc7071cadc8e71607725d87) - handling run (watch) test command + code lens
- code-lens - add `run test` `watch test` buttons for each test in currently opened file
- update vscode extension package.json to support `run test` and `watch test` for the current open file
- update vscode extension package.json to support new `bun.test.filePattern` extension setting (pattern for matching the test files)
### How did you verify your code works?
Unfortunately, there is no test method in bun vs-code extension code base. So just manually :(

PR #15261, opened 2024-11-19
feat(vscode-extension) error reporting, qol

alii:
### What does this PR do?

A few goodies implemented into the VSCode extension.
- A basic test runner that uses the TypeScript AST for finding tests. Most of this came from #13668
- Bring in the LifecycleAgent & co from #15194 
- Show errors inside of VSCode as they happen at runtime
- Implements a new DebugAdapter for communicating via a `node:net` Socket

Ran extension and tested each feature

### Notes:

- Async stack traces are NOT supported. This is because JSC does not actually record them unless the debugger is open. The debugger is big overhead, so it's not open in this case.
- Ideally we would show the function name of `.relatedInformation` for each diagnostic. Right now we just show the error message again. This isn't _amazing_...
- Disabling this right now would be done by setting the `BUN_INSPECT_NOTIFY` variable to `''`. A use case for disabling would be like when running a benchmark file (where this would actually have a very very slight performance overhead)
- The socket framer on the extension side could be rewritten


packages/bun-vscode/package.json
```diff
@@ -328,5 +357,8 @@
         }
       }
     ]
+  },
+  "dependencies": {
+    "strip-ansi": "^7.1.0"
```
Electroid: Let's remove this dependency, we don't need it. Can be 1 line utility somewhere

alii: Sure, removed.

src/bun.js/bindings/InspectorLifecycleAgent.cpp
```diff
@@ -0,0 +1,131 @@
...
+    Ref<JSON::ArrayOf<String>> urls = JSON::ArrayOf<String>::create();
+    Ref<JSON::ArrayOf<int>> lineColumns = JSON::ArrayOf<int>::create();
+    Ref<JSON::ArrayOf<String>> sourceLines = JSON::ArrayOf<String>::create();
+
+    for (size_t i = 0; i < exception.stack.source_lines_len; i++) {
+        sourceLines->addItem(exception.stack.source_lines_ptr[i].toWTFString());
+    }
+
+    for (size_t i = 0; i < exception.stack.frames_len; i++) {
+        ZigStackFrame* frame = &exception.stack.frames_ptr[i];
```
dylan-conway, line 109: In the future, I think we can get the function name from `frame` here

alii: Interesting, we'd need to add this to the inspector event. Probably can't guarantee that we'll always have it (@Jarred-Sumner ?)

dylan-conway (approved):

alii:
cc @Jarred-Sumner @dylan-conway 
CleanShot 2024-11-21 at 17 28 38@2x
Could this actually be relevant? Or expected that some/certain tests are failing here

dylan-conway:
Yup, this is a new test failure, looking into it. I believe the cause is from https://github.com/oven-sh/bun/pull/15261/files#diff-36302f74ed5420c7bee6a5460115be29fd82720720381c0d2c8bb7c934e69591L382

bench/snippets/INTERNAL_DEBUGGER
Electroid, line 1: Remove empty file

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -274,26 +290,24 @@ export class DebugAdapter extends EventEmitter<DebugAdapterEventMap> implements
...
   /**
-   * Gets the inspector url.
+   * Gets the inspector url. This is deprecated and exists for compat.
+   * @deprecated You should get the inspector directly, and if it's a WebSocketInspector you can access `.url` direclty.
    */
   get url(): string {
-    return this.#inspector.url;
-  }
+    // This code has been migrated from a time when the inspector was always a WebSocketInspector.
+    // So t
```
Electroid: Remove

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
```diff
@@ -437,18 +451,36 @@ export class DebugAdapter extends EventEmitter<DebugAdapterEventMap> implements
...
+    if (request.enableDebugger !== false) {
+      this.send("Debugger.enable").catch(error => {
+        const { message } = unknownToError(error);
+        if (message !== "Debugger domain already enabled") {
+          throw error;
+        }
+      });
+    }
+
     this.send("Debugger.setAsyncStackTraceDepth", { depth: 200 });
```
Electroid: Won't this command fail is `Debugger.enable` is not run above? Should it be moved into the new `if` block?

alii:
This code (or at least something similar iirc) was removed previously, I think because it was printing errors twice. Did you test this locally and if so does it seem to be working? I can try it out as well to test

packages/bun-debug-adapter-protocol/src/debugger/adapter.ts
alii: Should be, but interestingly it hasn't been failing from my testing. Will move it for the sake of being more correct though. nice spot thanks

Issue #6421, opened 2023-10-11
Running `bun test` in GitHub action causes SIGABRT since 1.0.5


RPDeshaies:
### What version of Bun is running?

1.0.5+1bf28e0d77a8b2261befbdb708cefd03e0126960

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. Configure a github action workflow with the following

```yaml
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Use Node.js "18"
        uses: actions/setup-node@v3
        with:
          node-version: '18'
      - name: Install bun
        uses: oven-sh/setup-bun@v1
        with:
          bun-version: v1.0.5
      - name: 'Install Dependencies'
        run: |
          bun install
      - name: 'Test'
        run: |
          bun test **/*.spec.ts --coverage

```

2. Create a folder `app`
3. Create a file `bun.spec.ts`  and add the following content
```tsx
import { describe, expect, it } from 'bun:test';
describe('Bun test', () => {
  it('should work', () => {
    expect(true).toEqual(true);
  });
});

```
4. Push the content on Github and let the action run
 

### What is the expected behavior?

Expected: Running `bun run test` should both work locally and on GitHub Action

### What do you see instead?

Actual: Error code 6

```sh
$ bun test **/*.spec.ts --coverage
bun test v1.0.5 (1bf28e0d)
error: script exited with code 6 (SIGABRT)
Error: Process completed with exit code 1.
```

### Additional information

This was working fine in v1.0.4 and if I specify 1.0.4 in the config of the `oven-sh/setup-bun@v1` workflow, it still works. But if I point to latest it crashes with the error code. It also works fine locally so I'm not exactly sure what's going on.

jamonholmgren:
Can confirm that it also fails on CircleCI:

https://app.circleci.com/pipelines/github/jamonholmgren/bluebun/88/workflows/cc81b3bb-81b3-4b96-b4e7-3d6d0469d361/jobs/101/parallel-runs/0/steps/0-105

jamonholmgren:
As a workaround, you can install bun via this command on CI:

```
curl -fsSL https://bun.sh/install | bash -s -- bun-v1.0.4
```

However, @Jarred-Sumner is discouraging pinning `bun` to anything but `latest` in this comment: https://github.com/oven-sh/bun/pull/848#issuecomment-1193697983

[...]

PR #15265, opened 2024-11-19
ci: bootstrap.sh: musl download of bun no longer has to be special-cased

nektro:
bun 1.1.35 included builds of musl so it can use the install script like the other targets


robobun:
@nektro, your commit 487bb4924360e70e18904ecf458d4bc8ae863e6e is building: [`#6421`](https://buildkite.com/bun/bun/builds/6421)

Issue #15141, opened 2024-11-14
tracking: JSValue safety in our Zig and JS bindings


nektro:
- better code generator
- https://github.com/oven-sh/bun/pull/15100
- https://github.com/oven-sh/bun/pull/15120
- https://github.com/oven-sh/bun/pull/15127
- https://github.com/oven-sh/bun/pull/15128
- replace `orelse return null` with `try`
- replace `catch return null` with `try` (done in 15120)
- make `JSValue.get()` return a possible JSError
- make `JSValue.toZigString()` return a possible JSError
- make `globalThis.throw*` return a JSError
- when done, audit all `globalThis.hasException()` calls for redundancy 
- audit all `fromJS()` functions to return possible JSError
- make `globalThis.ERR_*().throw()` return a JSError

more to come

DonIsaac:
to add to this list: `bun.string.toJSByParseJSON` could throw and should return `bun.JSError!JSC.JSValue`

PR #15270, opened 2024-11-20
zig: JSValue: make .get and .toSliceOrNull use JSError

nektro:

### DIFF
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -5446,7 +5446,7 @@ pub const JSValue = enum(i64) {
     /// Equivalent to `obj.property` in JavaScript.
     /// Reminder: `undefined` is a value!
     ///
-    /// Prefer `get2` in new code, as this function is incapable of returning an exception
+    /// Prefer `get` in new code, as this function is incapable of returning an exception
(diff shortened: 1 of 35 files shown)


src/bun.js/bindings/bindings.zig
```diff
@@ -5474,7 +5467,7 @@ pub const JSValue = enum(i64) {
     /// per invocation.
     ///
     /// This function will eventually replace `get`.
```
paperclover: this comment can be updated, now that this function has replaced `get`. additionally the line in get_unsafe can be edited to mention `get` instead of `get2`

paperclover (approved):
awesome. did not actually think get was going to be easy to replace all uses of but i am glad it is.

later we can also remove the public fast get function since `get` covers this use case

src/bun.js/bindings/bindings.zig
nektro:
```suggestion
```

src/bun.js/bindings/bindings.zig
```diff
@@ -5449,19 +5447,14 @@ pub const JSValue = enum(i64) {
     /// Reminder: `undefined` is a value!
     ///
     /// Prefer `get2` in new code, as this function is incapable of returning an exception
```
nektro:
```suggestion
    /// Prefer `get` in new code, as this function is incapable of returning an exception
```

PR #15284, opened 2024-11-20
feat(resolver): add support for self-referencing

### LINKED ISSUE #1137: ES Modules: Cannot resolve own specifier
### Version

0.1.10

### Platform

Darwin Lewiss-MacBook-Air.local 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:35 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T8101 arm64

### Repro

https://replit.com/@ctjlewis/bun-self-resolve

### What steps will reproduce the bug?

An ES Module attempting to resolve its own module specifier will fail to do so. You can see in the repro above that the standard TSC + Node flow does resolve the module's self-reference successfully.

I was initially unsure whether to file this as a bug or a feature request, but ultimately since Bun reuses the package.json standard, [it is expected behavior](https://twitter.com/oven_sh/status/1562248120611885056?s=20&t=6n9WT1t9KiLua2Q48UMN3g) that it supports Node-like ESM resolution and not throw in this case.

Per the specification, self-resolution only happens for ES modules with [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) enabled via the `exports` field in their package.json.  This is very common for supporting backwards compatibility for CJS and is also [encouraged for pure ESM](https://github.com/tsmodule/tsmodule).

An ES Module importing a named module that is equal to its own name (the `name` in its package.json), then it should resolve to itself. 

---

See: https://nodejs.org/api/esm.html#resolver-algorithm-specification

**PACKAGE_SELF_RESOLVE**(_packageName_, _packageSubpath_, _parentURL_)

  1. Let _packageURL_ be the result of **LOOKUP_PACKAGE_SCOPE**(_parentURL_).
  
  2. If _packageURL_ is **null**, then
    1. Return **undefined**.
  
  3. Let _pjson_ be the result of **READ_PACKAGE_JSON**(_packageURL_).
  
  4. If _pjson_ is **null** or if _pjson.exports_ is **null** or **undefined**, then
    1. Return **undefined**.
  
  5. If _pjson.name_ is equal to _packageName_, then
    1. Return the result of **PACKAGE_EXPORTS_RESOLVE**(_packageURL_, _packageSubpath_, _pjson.exports_, _defaultConditions_).
  
  6. Otherwise, return **undefined**.

### What do you see instead?

```
error: Could not resolve: "bun-self-resolve". Maybe you need to "bun install"?
import { test } from "bun-self-resolve";
```

### Additional information

_No response_

### LINKED ISSUE #7644: Support for self-referencing imports

1.0.15+b3bdf22eb

Darwin 21.6.0 arm64 arm

Simple steps (thanks @paperdave):

```
// index.js 
console.log(require("self/package.json"));

// package.json
{
  "name": "self",
  "exports": {
    "./package.json": "./package.json"
  }
}
```

Convoluted steps:

```
git clone https://github.com/prettier/prettier-cli.git
cd prettier-cli
bun i
bun run compile
bun ./dist/bin.js --version
```

Basically a package should be able to import one of its own exports, even if the package is not listed as a dependency to itself.

### What do you see instead?

_No response_

### Additional information

_No response_

SunsetTechuila:
### What does this PR do?

fixes #1137

fixes #7644

- [x] Code changes

I wrote automated tests


Jarred-Sumner:
Thank you

SunsetTechuila:
https://x.com/bunjavascript/status/1861203434931921402

> "why would you do this?"
you tell me

for tests: https://github.com/SunsetTechuila/is-bun-module/commit/d8030017d7cc4043a155f16d50c44d9f94dfa2ae

Issue #14664, opened 2024-10-18
Memory leak when using Prisma


boomNDS:
### How can we reproduce the crash?

I’m using Elysia.js to run an API server that handles approximately 30 requests per second. Over time, I observed an increase in CPU usage, which eventually affects the server’s performance. This issue seems to worsen after making several requests, leading to degraded system performance. Restarting the server temporarily resolves the issue, but this is not a viable long-term solution.

```Dockerfile
FROM oven/bun:1.1.29 AS build
WORKDIR /app
ENV HUSKY=0
COPY package.json bun.lockb ./
COPY ./prisma ./prisma
COPY ./src ./src

# Install dependencies
RUN bun install

# Generate Prisma client with the correct schema location
RUN bunx prisma generate --schema prisma/schema.prisma

# Compile the server to a binary
RUN bun build \
  --compile \
  --minify-whitespace \
  --minify-syntax \
  --target bun \
  --outfile server \
  ./src/index.ts


FROM oven/bun:1.1.29-slim AS runner

WORKDIR /app

# Copy the compiled binary and the Prisma schema from the build stage
COPY --from=build /app/server /app/server
COPY --from=build /app/node_modules/.prisma/client /app/node_modules/.prisma/client
COPY --from=build /app/prisma ./prisma

RUN apt-get update && apt-get install -y openssl

COPY start.sh /app/start.sh

ENV NODE_ENV=production

RUN chmod +x /app/start.sh

CMD ["/app/start.sh"]
```

## Shell Script (start.sh)
Here is the start.sh script used to start the server:
```
#!/bin/sh

# Run Prisma migrations
echo "Running Prisma migrations..."
bunx prisma migrate deploy --schema prisma/schema.prisma

# Check if the migrations succeeded
if [ $? -eq 0 ]; then
    echo "Migrations completed successfully!"
    
    # Start the server
    echo "Starting the server..."
    ./server
else
    echo "Migrations failed. Server will not start."
    exit 1
fi
```

### Stack Trace (bun.report)

```
{
  objectTypeCounts: {
    string: 37367,
    Object: 36919,
    Function: 31116,
    JSLexicalEnvironment: 20342,
    Structure: 12721,
    FunctionExecutable: 11238,
    UnlinkedFunctionExecutable: 9653,
    Promise: 9313,
    Generator: 4865,
    Array: 3660,
    SymbolTable: 2620,
    GetterSetter: 2335,
    FunctionCodeBlock: 2329,
[...]

Issue #12977, opened 2024-07-31
Potentiel memory leak on simple request (bun + fastify + prisma)


goldo:
### What version of Bun is running?

1.1.22-canary.42+bec04c734

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

I made a small reproduction project with bun + fastify + prisma (sqlite):
https://github.com/goldo/bun-memory

```sh
bun install 
bun start
# then start requesting it and check the heapstats logs on the server
curl http://localhost:3010/ 
autocannon -c 30 -d 30 http://localhost:3010
```

### What is the expected behavior?

Memory should be going down,  after a while, at approx. the level of the beginning of the server (first request)

### What do you see instead?

memory is going down just a little (GC), but stays really high

### Additional information

This is a following issue of https://github.com/oven-sh/bun/issues/7377

### Investigation of potential bun memory leak 

heapStats().objectTypeCounts check: 

```ts
    heapStats().objectTypeCounts["Array"],
    heapStats().objectTypeCounts["Object"],
    heapStats().objectTypeCounts["Structure"]
```

at the beginning:

```bash
$ bun start
# curl http://localhost:3010
708 2252 7658
```

after `autocannon -c 30 -d 30 http://localhost:3010` :

```bash
18905 54499 15811
```

After a few minutes (waiting for GC), the objectTypeCounts seems still high, and the memory usage is still high (101mb) vs initial (71mb):
```bash
# curl http://localhost:3010
17515 51800 15549
```

The problem seems to be the same with or without the fastify plugin.

Jarred-Sumner:
A memory leak is indefinite growth in memory usage. 70 MB -> 100 MB is not clearly a leak. If you can make it go from 70 MB -> 1 GB, and then continue rising after receiving the same request volume then that is a leak.

goldo:
On the full project, I got this: 
- before 05/08: bun (without build)
- after 05/08: bun + build target node + run in node 22
![Screenshot 2024-08-05 at 16 11 51](https://github.com/user-attachments/assets/8ec0e106-bdb0-430b-9d06-2b4ff70ad50d)

Jarred-Sumner:
Fixed in https://github.com/oven-sh/bun/pull/15309

This will be part of the Bun v1.1.36 release, which is tomorrow.

Issue #6599, opened 2023-10-19
Docker image breaks when using prisma


alexdotpink:
### What version of Bun is running?

1.0.6+969da088f5db3258a803ec186012e30f992829b4

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

Run some code in bun's linux/arm64 docker image.

### What is the expected behavior?

It should exit the process successfully.

### What do you see instead?

Process hangs, memory leaks and CPU usage skyrockets.

Electroid:
Thanks for reporting, do you have an example that we can run to try to reproduce?

alexdotpink:
I had made this repository to originally report the issue in Sentry's Issues, but then I figured it wasn't Sentry itself that was the problem but bun, because the issue also occurs if I use other libraries like Prisma or my own web framework (https://github.com/grace-js/grace).

The example is: https://github.com/Fumaz/sentry-prisma-memory-leak

alexdotpink:
Here's a screenshot of `ctop` showing the issue:
 
<img width="584" alt="image" src="https://github.com/oven-sh/bun/assets/45318608/911c9031-bde5-436f-8a06-f833aa9046bb">

alexdotpink:
It appears to me this issue is Prisma related. I also noticed this issue not only affects the linux/arm64 image, but also the linux/amd64 image, but in a different way. Instead of leaking memory, it simply crashes (SIGSEGV).

wwelder:
I can confirm the same behavior, memory leak on linux/arm64 and fail on amd64 with prisma

CPlusPatch:
Getting the same behaviour here, I have Bun and Prisma inside a Docker container, as soon as I do anything with Prisma the process gets killed with code 139 (I think this means it got killed by the kernel for using too many resources)

Jarred-Sumner:
Fixed in https://github.com/oven-sh/bun/pull/15309

This will be part of the Bun v1.1.36 release, which is tomorrow.

kravetsone:
> Fixed in [#15309](https://github.com/oven-sh/bun/pull/15309)
> 
> This will be part of the Bun v1.1.36 release, which is tomorrow.

`v1.1.37`)

Jarred-Sumner:
thanks i forgot

PR #15289, opened 2024-11-20
Fix memory leak in napi_threadsafe_function

190n:
### What does this PR do?

Previously napi_threadsafe_function wouldn't free the memory for its channel when it was freed. Now it does. This manifested, for instance, as a leak of 8 bytes per Prisma query.

Fixes #14664 
Fixes #12977 
Fixes #6599

I added a memory leak test to `prisma.test.ts`.


robobun:
@190n, your commit 7478de533e3c037bbabc138b5ba344adbdeee5b5 is building: [`#6475`](https://buildkite.com/bun/bun/builds/6475)

Issue #6510, opened 2023-10-15
Bun.listen never connects, never errors


runspired:
### What version of Bun is running?

1.0.6

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

```ts
function debug(msg) { console.log(msg); }
const port = 7537;

await new Promise((resolve, reject) => {
  Bun.listen({
    port,
    hostname: 'localhost',
    socket: {
      data() {
        debug(`Port ${port} unexpectedly received data 🙈`);
      },
      open() {
        debug(`Port ${port} is available`);
        server.stop(true);
      },
      close() {
        debug(`Releasing ${port} for server use`);
        resolve();
      },
      connectError(socket, error) {
        debug(`Port ${port} received connectError event`);
        reject(error);
      },
      timeout() {
        debug(`Timeout while checking port ${port}`);
        reject(new Error(`Timeout while checking port ${port}`));
      },
      error(socket, e) {
        debug(`Port ${port} received error event`);
        reject(e);
      },
      end(socket) {
        debug(`Port ${port} received end event`);
        reject(new Error(`Port ${port} received end event`));
      }
    }
  });
})
```

### What is the expected behavior?

handlers fire once socket is available or socket connection has failed

### What do you see instead?

socket connection never established, error never thrown

Jarred-Sumner:
`Bun.listen` returns a Promise

have you tried `await`'ing it?

runspired:
@Jarred-Sumner no because the docs do not document it as having a return at all. I was wondering. If it were a promise though I'd still expect a global error to be thrown for the case where I run this and know the port is busy; however, even in that case it appears to just hang.

runspired:
I can confirm I can await success. It still appears to not throw for the case where the port is in fact busy.

liz3:
the question here is very implementation dependent, first of all what do you define as a error which should throw?
Id argue everything should theoretically throw in the current scope no matter what its relateed to socket creation/client/data.

But if you have a defined error handler that changes a lot because then do you `throw` the error or invoke the error handler to deal with the error instead of letting the callee(the socket api) throw?

runspired:
@liz3 I'm specifically trying to detect if the port was already in use, and would expect an `EADDRINUSE` error to be thrown

runspired:
[...]

PR #15298, opened 2024-11-21
remove optional ? from cli.zig

### LINKED ISSUE #15293: `--bail` flag behavior to support space-separated arguments in `bun test`

1.1.37-canary.10+61a3f0859

_No response_

Meanwhile with `bun test` all the tests pass ok, with `bun test --bail 10` is looking for files with 10, so is not working.

```sh
> bun test --bail 10

bun test v1.1.37-canary.10 (61a3f085)
The following filters did not match any test files:
  10
2323 files were searched [62.00ms]

note: Tests need ".test", "_test_", ".spec" or "_spec_" in the filename (ex: "MyApp.test.ts")

Learn more about the test runner: https://bun.sh/docs/cli/test
```

Just run `bun test --bail 10`

### What do you see instead?

_No response_

### Additional information

_No response_

RiskyMH:
### What does this PR do?

Speculative fix for cases like `bun test --bail 10` where you actually need the `=` in between whitespace. I'm not sure I fully understand the reason for the `?` option as there are many other flags that are optional but don't have it.

may fix #15293 

- [ ] Code changes

CI will tell me


robobun:
@RiskyMH, your commit 2a5f8c5eba14653dbabc909cea939d6ddeceb5c9 is building: [`#6510`](https://buildkite.com/bun/bun/builds/6510)

Electroid:
If we remove the `?`, it would be a behaviour change. For example, `--inspect` with no arguments would not work

RiskyMH:
this is very bad change after consideration

PR #15302, opened 2024-11-21
zig: make throwInvalidArgumentTypeValue use JSError

nektro:


robobun:
@nektro, your commit 038ce1519d388aa855501003a60aa43a351bdb89 is building: [`#6526`](https://buildkite.com/bun/bun/builds/6526)

Jarred-Sumner (changes requested):
We cannot merge this without tests that verify these bindings work.

cirospaciari (changes requested):
Need update before merging
Screenshot 2024-11-26 at 15 00 31

cirospaciari (approved):

nektro:
binding tests were added and fixed in https://github.com/oven-sh/bun/pull/15362, then pulled in during an earlier `main` merge

PR #15305, opened 2024-11-21
zig: make throwInvalidArguments use JSError

nektro:


robobun:
@nektro, your commit e4915e43a383bba0a62aeb4347285099fc4e2661 is building: [`#6547`](https://buildkite.com/bun/bun/builds/6547)

Jarred-Sumner:
I think the test failures that are not alpine are real

nektro:
i believe theyre real but not caused by this pr, ive seen them fail on a handful of others
will investigate

paperclover (approved):

Issue #15320, opened 2024-11-21
Bun serve with a Response with empty ReadableStream won't redirect.


adrifer:
### What version of Bun is running?

1.1.36+ededc168c

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Run this code:
```ts
Bun.serve({
    port: process.env.NITRO_PORT || process.env.PORT || 3e3,
    websocket: void 0,
    async fetch(req, server2) {
      const url = new URL(req.url);
      if(url.pathname === "/redirect") {
          const emptyStream = new ReadableStream({
            start(controller) {
              // Immediately close the stream to make it empty
              controller.close();
            }
          });
          
          return new Response(emptyStream, {
            status: 307,
            headers: {
                location: "/"
            }
          });
    }
    
    return new Response("Hello world");
    }
  });
```

1. navigate to localhost:3000/redirect

Nothing will happen, the response the browser receives is empty (no headers or body) and no redirect.

### What is the expected behavior?

It should redirect to / and show Hello world in the browser.

### What do you see instead?

White screen, because the response contains no body or headers.

adrifer:
As context, this breaks new TanStack Starts framework, basically when using Bun all redirects will do nothing and an empty page will be rendered.  https://github.com/TanStack/router/issues/2819

adrifer:
Any updates on this? I'm still blocked on using bun on Production due to this. Any TanStack Start project which does a redirect will be impacted by this.

adrifer:
Thanks a lot @cirospaciari for taking the time to try to fix this!! I really appreciate it!

Issue #6594, opened 2023-10-19
happy-DOM error this._getWindow is not a function


croqaz:
### What version of Bun is running?

1.0.6

### What platform is your computer?

Void Linux 6.5.7_1 x86_64 x86_64

### What steps can reproduce the bug?

Trying to run a simple happy-dom script from:
https://github.com/capricorn86/happy-dom/wiki/Getting-started#usage

From package.json - "happy-dom": "12.9.1"

```js
const { Window } = require('happy-dom');

const window = new Window({
   url: 'https://localhost:8080',
   width: 1024,
   height: 768
});
const document = window.document;

document.body.innerHTML = '<div class="container"></div>';

const container = document.querySelector('.container');
const button = document.createElement('button');

container.appendChild(button);

// Outputs "<div class="container"><button></button></div>"
console.log(document.body.innerHTML);
```

### What is the expected behavior?

It should print some HTML, not crash.

### What do you see instead?

bun happy.js

```
<div class="container"><button></button></div>
75 |      * @see https://www.quirksmode.org/js/events_order.html#link4
76 |      * @param event Event.
77 |      * @returns The return value is false if event is cancelable and at least one of the event handlers which handled this event called Event.preventDefault().
78 |      */
79 |     dispatchEvent(event) {
80 |         const window = this._getWindow();
                           ^
TypeError: this._getWindow is not a function. (In 'this._getWindow()', 'this._getWindow' is undefined)
      at dispatchEvent (/root/PWD/node_modules/happy-dom/cjs/event/EventTarget.cjs:80:23)
      at dispatchEvent (/root/PWD/node_modules/happy-dom/cjs/event/EventTarget.cjs:102:20)
      at /root/PWD/node_modules/happy-dom/cjs/nodes/document/Document.cjs:877:12
```

### Additional information

The same script works in Node.js v18.18.1.

croqaz:
Also similar bug in happy-DOM: https://github.com/oven-sh/bun/issues/5847

I am only trying it because it's mentioned in the ecosystem test runner: https://bun.sh/guides/test/happy-dom

croqaz:
The issue may be fixed in https://github.com/capricorn86/happy-dom/issues/1085
It's not released yet.

PR #15322, opened 2024-11-22
fix(server) Fix empty response 

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/15320

This fix responses when the stream is early closed without sending any data.

- [X] Code changes

New tests + existing tests


robobun:
@cirospaciari, your commit 21ac094038f6a4f133819107ca9c2bc32423aaeb is building: [`#6594`](https://buildkite.com/bun/bun/builds/6594)

src/bun.js/webcore/response.zig
```diff
@@ -1785,12 +1797,15 @@ pub const Fetch = struct {
...
-            defer task.mutex.unlock();
+            // at this point only this thread is accessing result to is no race condition
             const is_done = !result.has_more;
             // we are done with the http client so we can deref our side
-            defer if (is_done) task.deref();
+            // this is a atomic operation and will enqueue a task to deinit on the main thread
+            defer if (is_done) task.derefFromThread();
 
+            task.mutex.lock();
+            // we need to unlock before task.deref();
```
Jarred-Sumner: Are you sure? The reference count is atomic.

cirospaciari: yes the ref count is atomic and dont need the mutex, if task is freed accessing the mutex after deref could be a UAF (with was happen if you use TLS in some tests and run a lot of times), now we are enqueuing the deinit to main thread so is less likely to happen too, but still make sense to avoid the race condition and just unlock before we call derefFromThread with is a thread safe function.

Jarred-Sumner: which tests are no longer flaky due to this change?

cirospaciari: Should be the `test/js/bun/http/serve.test.ts > should be able to stop in the middle of a file response` but dispite fixing a crash it still flaky on macOS and musl, but not failing anymore on ubuntu, with means we have more issues

cirospaciari: I broke down this PR in two https://github.com/oven-sh/bun/pull/15432

cirospaciari:
superseded by https://github.com/oven-sh/bun/pull/18707

Issue #6639, opened 2023-10-21
http.createServer - Content-Type differences between Bun and NodeJS


alexsch01:
### What version of Bun is running?

1.0.25+a8ff7be64

### What platform is your computer?

Linux 6.1.64-09009-gcb08142c2ff8 x86_64 unknown

### What steps can reproduce the bug?

mkdir test
cd test
bun install https://github.com/alexsch01/bunBug

### What is the expected behavior?

**node -e "require('bunBug')()"**

`Hello World`

Response Headers

```
HTTP/1.1 200 OK
Date: Sat, 21 Oct 2023 15:40:28 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Content-Length: 24
```

### What do you see instead?

**bun -e "require('bunBug')()"**

`<body>Hello World</body>`

Response Headers

```
HTTP/1.1 200 OK
***content-type: text/plain;charset=utf-8***
Date: Sat, 21 Oct 2023 15:39:48 GMT
Content-Length: 24
```

### Additional information

I would expect there to not have a `content-type` by default, to align with the behavior of NodeJS

alexsch01:
Updated the description

PR #15333, opened 2024-11-22
crash_handler: use WTFReportBacktrace in canary

nektro:
canary release debug info is not uploaded to https://bun.report so this will help us and users have much more info about where release-mode crashes may be happening


robobun:
@nektro, your commit b9e1c8d949d51c40dc5037418c527fa1f5db05a1 is building: [`#6639`](https://buildkite.com/bun/bun/builds/6639)

PR #15342, opened 2024-11-22
rename `CallFrame.arguments_old` to `argumentsDeprecated`

Jarred-Sumner:
### What does this PR do?

@nektro when you want to say something is deprecated, don't add "old" or "2" to the name. That doesn't communicate useful information. It encourages adding "3" or "oldold" to the name. Instead, say "deprecated".

### DIFF
--- a/src/bun.js/BuildMessage.zig
+++ b/src/bun.js/BuildMessage.zig
@@ -86,7 +86,7 @@ pub const BuildMessage = struct {
         globalThis: *JSC.JSGlobalObject,
         callframe: *JSC.CallFrame,
     ) bun.JSError!JSC.JSValue {
-        const args_ = callframe.arguments_old(1);
+        const args_ = callframe.deprecatedArguments(1);
         const args = args_.ptr[0..args_.len];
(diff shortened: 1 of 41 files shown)


robobun:
@Jarred-Sumner, your commit 27d171ca7a2427e26181954aa72c1c08a405e277 is building: [`#6679`](https://buildkite.com/bun/bun/builds/6679)

paperdave:
this pr doesnt build because of changes on the main branch. instead of renaming a field, we should just delete arguments_old.

instead of renaming something to `_old`, just delete it. patches like this just eat CI and maintainer time. and when theyre large like this one, its really easy to get the code into a state where it doesnt compile.

imo prefer developing replacement api and then give that api a new name, or replace all old uses to the new api.

nektro:
the `_old` pr was to free up the `.arguments` name for the new pattern partially described in the doc comment of CallFrame. there will be no oldold because the plan it to migrate all usages of arguments_old to a much less error prone way of parsing function arguments.

nektro:
the new generator is now being worked on

Issue #6738, opened 2023-10-26
Vite Rollup fails on "bun:sqlite" import even when it is marked as external


joshdchang:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

When importing `bun:sqlite` in a Vite/Rollup project, it causes an error on build.
```ts
import { Database } from "bun:sqlite";
```

I marked `bun:sqlite` as an external dependency (`external: ["bun:sqlite"]` in Rollup config), so it should be ignored but it isn't.

### What is the expected behavior?

I expected the build to compile while allowing `bun:sqlite` to be external. The Vite dev server works as expected, this is just on build.

### What do you see instead?

I'm getting the following error: `Error [PLUGIN_ERROR]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. Received protocol 'bun:'`

### Additional information

I am submitting this as an issue to Bun because I think it may be useful for cases like this to have a way to reference the `bun:sqlite` module without using the `:`, such as having `bun/sqlite` as an alias or just including it as part of the `Bun` object.

Electroid:
We'll need to submit PRs to Rollup and Vite to consider fixing this. /cc @colinhacks

joshdchang:
Should I submit an issue to Rollup?

SlavenIvanov:
Is there a workaround available for this?

keisanng:
Is there any solution @bun team?

zackify:
The only workaround I found, is using better-sqlite3 in dev. And marking it external works but only for production builds. 

This sucks but works in the meantime.

Jarred-Sumner:
If marking `bun:sqlite` as external in Rollup doesn't work in Rollup, then it's hard for us (Bun) to do anything about that.

Can you file an issue in Rollup's repository and link to it here?

zackify:
Oh yeah, I know it’s on them. I actually have it on my list tomorrow to find where the problem is, vite or esbuild and patch it. I’ll respond here with what I find.

zackify:
Alright, I found out the issue. Hopefully this helps someone.

Vite will work, bun you have to be running fully under bun.

I had remix setup like this:

`remix vite:dev`

I had to change this to 


same with the build command:

`bunx --bun remix vite:build`

This will resolve bun deps without extra config options in vite.


[...]

PR #15357, opened 2024-11-23
Fix setRawMode return value on Windows

heimskr:
### What does this PR do?

Fixes the return value of setRawMode, which until now on Windows was directly returning a C++ int as an EncodedJSValue without actually encoding it. This fixes some `node:readline` crashes on Windows.

- [X] Code changes

Spammed a script that calls `readline.createInterface` a lot.


robobun:
@heimskr, your commit 54ef3315f0ccfe979df99eb8a1a778fb2691a934 is building: [`#6738`](https://buildkite.com/bun/bun/builds/6738)

Issue #15396, opened 2024-11-25
Bun.serve's built-in error page does not render correctly


luavixen:
### What version of Bun is running?

1.1.36+ededc168c

### What platform is your computer?

Linux 6.12.1-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

```js
Bun.serve({
  development: true,
  fetch(req) {
    throw new Error("woops!")
  }
})
```

### What is the expected behavior?

![](https://bun.sh/images/exception_page.png)

### What do you see instead?

![](https://dl.vixen.link/b4mycg/Screenshot%20from%202024-11-24%2019-09-31.png)

### Additional information

Example renders error page template which includes this snippet: https://github.com/oven-sh/bun/blob/2d2e329ee3d6a05740aa8bb8adfab72f7a4cecec/src/fallback-backend.html#L20-L24
Which uses variable `bun_error` provided here by `errorJS()`: https://github.com/oven-sh/bun/blob/2d2e329ee3d6a05740aa8bb8adfab72f7a4cecec/src/runtime.zig#L141-L147
Which is incorrectly defined: https://github.com/oven-sh/bun/blob/2d2e329ee3d6a05740aa8bb8adfab72f7a4cecec/src/runtime.zig#L72-L84
Note that `errorJS()` returns `@embedFile("bun-error/bun-error.css")` which is obviously incorrect.

PR #15397, opened 2024-11-25
Bun.serve incorrect file for error page template

luavixen:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/15396

- [x] Code changes

Ran example provided in https://github.com/oven-sh/bun/issues/15396 which now works correctly:


Jarred-Sumner:
Thank you

Issue #6806, opened 2023-10-30
Basic "resolution" fails in a workspace.


jpasquers:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

Basic workspace setup. Package.json in the root looks like:

```
{
  "private": true,
  "workspaces": {
    "packages": [
      "packages/*"
    ]
  }
}
```

package.json in `packages/pkg1` looks like:

```
{
  "name": "x",
  "version": "1.0.0",
  "license": "UNLICENSED",
  "private": true,
  "devDependencies": {
    "uncontrollable": "^7.2.1"
  },
  "resolutions": {
    "@types/react": "^17.0.0"
  }
}
```

Within `packages/pkg1`, run:

`bun install`
`bun pm ls --all`

The output is:

```
├── @babel/runtime@7.23.2
**├── @types/prop-types@15.7.9**
├── @types/react@18.2.33
├── @types/scheduler@0.16.5
├── csstype@3.1.2
├── invariant@2.2.4
├── js-tokens@4.0.0
├── loose-envify@1.4.0
├── react@18.2.0
├── react-lifecycles-compat@3.0.4
├── regenerator-runtime@0.14.0
├── uncontrollable@7.2.1
└── x@workspace:packages/pkg1
```

### What is the expected behavior?

The expected behavior is that following the resolution Bun should install `17..0.59` for `@types/react`.

### What do you see instead?

Version `18.2.33` is installed.

### Additional information

The same behavior works fine if done without workspaces. Am I running workspaces incorrectly? From my reading, the docs say:

> The root package.json should not contain any "dependencies", "devDependencies", etc. 

So I'm assuming I shouldn't be moving the resolution up to the root or anything right?

PR #15400, opened 2024-11-25
Fix cross-compiling zig on alpine

Electroid:
### What does this PR do?

The ordering of options in CMake is important. Before `project()` is called, variables such as `LINUX` and `WIN32` correspond to the _host_ platform, but after `project()` is called, it corresponds to the _target_ platform.

- `Globals.cmake` is evaluated before `project()`
- `Options.cmake` is evaluated after `project()`

Previously, `ABI` was calculated in `Globals`, but this would mean that if you tried to cross-compile Zig on an Alpine host, it would think that the target should also have the `musl` ABI.


robobun:
@Electroid, your commit 953818e1cb99695fc0336ebd4bf7bac3710e52b9 is building: [`#6806`](https://buildkite.com/bun/bun/builds/6806)

Issue #6878, opened 2023-11-03
Improve JSX transform using tsconfig jsx precompile for server side rendering


neves:
### What is the problem this feature would solve?

It's all explained here https://deno.com/blog/v1.38#fastest-jsx-transform but the idea is to implement this faster and memory efficient jsx transform. Example:
```json
{
  "compilerOptions": {
    "jsx": "precompile",
    "jsxImportSource": "custom"
  }
}
```

```javascript
// input
const a = (
  <div className="greeting">
    <a href={link}>
      Hello <b>{name}!</b>
    </a>
  </div>
);

// output
import { jsxAttr, jsxEscape, jsxTemplate } from "custom/jsx-runtime";

const tpl = [
  '<div class="greeting"><a ',
  ">hello <b>",
  "!</b></a></div>"
];
const a = jsxTemplate(tpl, jsxAttr("href", link), jsxEscape(name));
```

`custom/jsx-runtime` can easily be implemented as:
```javascript
export const jsxTemplate = String.raw
export const jsxEscape = Bun.escapeHTML
export const jsxAttr = (name, value) => ` ${name}="${value}"`
```

### What is the feature you are proposing to solve the problem?

It's all explained here https://deno.com/blog/v1.38#fastest-jsx-transform

PR #15413, opened 2024-11-26
zig: make throwOutOfMemory use JSError

nektro:

### DIFF
--- a/src/bun.js/ConsoleObject.zig
+++ b/src/bun.js/ConsoleObject.zig
@@ -91,7 +91,7 @@ pub fn messageWithTypeAndLevel(
 ) callconv(JSC.conv) void {
     messageWithTypeAndLevel_(ctype, message_type, level, global, vals, len) catch |err| switch (err) {
         error.JSError => {},
-        error.OutOfMemory => global.throwOutOfMemory(),
+        error.OutOfMemory => global.throwOutOfMemory() catch {},
     };
 }
 fn messageWithTypeAndLevel_(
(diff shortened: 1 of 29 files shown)


robobun:
@nektro, your commit fe2ba09c6e6d651b4afb1eb4c5c4527652cf218c is building: [`#6878`](https://buildkite.com/bun/bun/builds/6878)

src/bun.js/ConsoleObject.zig
```diff
@@ -91,7 +91,7 @@ pub fn messageWithTypeAndLevel(
 ) callconv(JSC.conv) void {
     messageWithTypeAndLevel_(ctype, message_type, level, global, vals, len) catch |err| switch (err) {
         error.JSError => {},
-        error.OutOfMemory => global.throwOutOfMemory(),
+        error.OutOfMemory => global.throwOutOfMemory() catch {},
```
paperclover: can we add // TODO: properly propagate exception upwards.

i checked why this returns void and discovered that our main usage of this clears the exception immediately.

src/bun.js/api/glob.zig
```diff
@@ -252,14 +243,11 @@ fn makeGlobWalker(
     const only_files = matchOpts.only_files;
 
     if (cwd != null) {
-        var globWalker = alloc.create(GlobWalker) catch {
-            globalThis.throwOutOfMemory();
-            return error.JSError;
-        };
+        var globWalker = try alloc.create(GlobWalker);
```
paperclover: according to zack this allocation is not by an arena. i feel like theres alot of these patterns in the codebase, but we can now use errdefer to prevent memory leaks in error paths.

paperclover (approved):

PR #15426, opened 2024-11-26
ci typescript, improve build time, remove in-memory result buffering 

atoko:
### What does this PR do?

This PR introduces typescript to the code that runs during Bun's CI process. It fixes some outstanding bugs and offers improved test execution speed and memory usage.

This set of changes does not introduce any behavior changes to the existing process, except for additional runtime validation by use of the builder pattern for operating on the CI domain model.

The runtime continues to be NodeJs for these scripts, although it requires an experiment flag to correctly run. 

---

## Motivations

This change is motivated by three factors:
1. Safety
2. Maintainability
3. Efficiency

### Safety:

This change aims to reduce the time needed to add features to CI by providing a more strict type system, that will alert developers whenever a change they are making is invalid. 
*references:
https://github.com/oven-sh/bun/commit/a468d09/
https://github.com/oven-sh/bun/commit/fc94db1/

This change also introduces some very light memoization of expensive system calls to the shell. These are used to get the current git branch, commit, etc.

In the process of testing the changes, I noticed that running the mjs pipeline generation on my local machine generates all targets. I'm not sure if this is the desired behavior. The new code correctly checks the existing architecture and generates the correct ci.yml
*references: https://github.com/oven-sh/bun/blob/f21fffd1bf8bee80698efea0938436e53a30c9a2/.buildkite/ci.mjs#L618

Finally, the current CI script incorrectly calls `process.exit` when tests are successful. I suspect this _fixes a bug_ where the output was not being flushed entirely onto the build log. 

### Maintainability:
This change organizes the CI system's imports from the "utils" script into code files that are more relevant and discoverable for new contributors. This decreases the barrier to entry and also improves retention for developers. In addition, it is known that smaller functions are more easily tested and understood.  

One other point i'd like to make regarding Maintainability is the reduction of dead code. For example, `Vendor[]`. This class isn't actually in the code, but is referenced by at least two functions in the test runner. With proper typing, this should never happen.

*references:
https://github.com/oven-sh/bun/blob/main/scripts/runner.node.mjs#L843

#### Additional notes:

I also added a very simple retry mechanism to the unit tests. 

### Efficiency

The current iteration of the Test runner buffers all tests executions in memory. This forces the test runners to page memory constantly. My computer capped the memory the test process was using at 5gb. In addition, this memory buffer was being read multiple times for the same calculations (verifying if all tests passed)

I rewrote some of the test execution so that test results are buffered in /tmp before being read back to calculate totals.  

We now have more memory to work with, and we can let the OS schedule each process and provide inter-process isolation. This directly leads to less flaky tests. However, this might need additional scrutiny with regards to the inspector/debugger tests, as those do rely on shared state.

All in all, memory usage decreased *90%*. build time optimized by *10%*. Vendor tests remain to be optimized under this model. Note that this is inclusive of the new retry logic, so the total test time may be similar, although with more passing tests

🐰 

-->

#### ./ci/generate-buildkite-pipeline.ts
- Execution log of `./buildkite/ci.mjs` and resulting `.buildkite/ci.yml`: https://gist.github.com/atoko/0271472351253bbc5d288464989221a1 (4s)
- Execution log of `./ci/generate-buildkite-pipeline.node.ts` and resulting `.buildkite/ci.yml`: https://gist.github.com/atoko/690732276cb2bedf1bba8a51cd89ca41 (1s)

#### ./ci/runner.node.ts
- Execution log of existing test runner: https://gist.github.com/atoko/42e0d4bffb7ffc8f221a527973103842 `*./scripts/runner.node.mjs > original.tests.log  1479.90s user 627.06s system 54% cpu 1:21:19.03 total*`

- Execution log of new test runner:
https://gist.github.com/atoko/e45c5ee9b3e1cbcecfee9f31abb8827e
·./ci/runner.node.ts > typescript.tests.log  2851.70s user 1132.35s system 77% cpu 1:25:53.15 total·

Updated runtime:
Just for future reference, this PR enabled me to reduce the runtime down to 36 minutes.

image

I'm deeply saddened by the fact that the Bun team doesn't seem to care about the execution speed or maintainability of their tests.

I see the team struggling with Promise.all and Promise.allSettled and it tells me this is not a serious project.


Jarred-Sumner:
Thank you for the effort

I worry this PR increases complexity and that will make it harder for us to iterate on CI quickly

Issue #6930, opened 2023-11-06
Support for save-prefix in bun add


mixedCase:
### What is the problem this feature would solve?

Npm offers the [save-prefix](https://docs.npmjs.com/cli/v10/using-npm/config#save-prefix) configuration option in order to provide a consistent way of configuring newly added package versions when using `npm i --save` and `npm i --save-dev` without version information.

`bun add` currently offers no equivalent feature.

### What is the feature you are proposing to solve the problem?

Adding an equivalent configuration option to bunfig.toml that is respected by `bun add` when adding the new package into package.json.

### What alternatives have you considered?

- Manual oversight when integrating changes
- Adding a custom validation step as part of each project's CI pipeline
- Praying that our dependencies don't break on a minor version

spacedawwwg:
+1 for this!

oviirup:
I do not know zig language, but after looking through the codebase, I have a probable solution.

The code relevant to save-prefix might be from [this section of code](https://github.com/oven-sh/bun/blob/main/src/install/install.zig#L8332-L8340), which is as follows.

```zig
const new_version = try switch (options.exact_versions) {
    inline else => |exact_versions| std.fmt.allocPrint(
        allocator,
        if (comptime exact_versions) "{}" else "^{}",
        .{
            resolutions[request.package_id].value.npm.version.fmt(request.version_buf),
        },
    ),
};
```

#### Explanation (as per my understanding)

1. It checks if the "exact" flag is set (-E or --exact),
2. Formats the version with either no prefix or "^" as prefix\
   `.npm.version.fmt` function is responsible for formatting, so this might be the point of interest
3. assigns to new_version

#### Solution
1. Check if `--exact` flag is set,\
   if "true" return the version as is
2. Check for `save-prefix` option in `.npmrc` file\
   If not set use "^" as the prefix
3. format the version with prefix
4. assign to new_version

> [!NOTE]
> Once again I do not know the zig language. I've suggested the solution as per my basic understanding of other languages like "rust" or "python". It is just a suggestion, this might be, or might not be the solution for this issue.

mrThomasTeller:
Any progress on this issue?

PR #15431, opened 2024-11-26
FFI: provide napi_env explicitly

heimskr:
### What does this PR do?

As the first step in the process of splitting the NAPI fixes draft PR into smaller pieces, this PR changes how FFI code obtains a napi_env. Instead of casting the global object, we make the napi_env available through an extern symbol.

- [x] Code changes

Ran ffi/cc tests

### DIFF
--- a/src/bun.js/api/ffi.zig
+++ b/src/bun.js/api/ffi.zig
@@ -444,7 +444,7 @@ pub const FFI = struct {
                 return error.DeferredErrors;
             }
 
-            for (this.symbols.map.values()) |symbol| {
+            for (this.symbols.map.values()) |*symbol| {
                 if (symbol.needsNapiEnv()) {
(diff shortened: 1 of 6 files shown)


robobun:
@heimskr, your commit c50dbf3d41cfce8a8dff10c5e6e72bd6aa8557d9 is building: [`#6930`](https://buildkite.com/bun/bun/builds/6930)

src/bun.js/api/ffi.zig
```diff
@@ -444,6 +444,13 @@ pub const FFI = struct {
                 return error.DeferredErrors;
             }
 
+            for (this.symbols.map.values()) |symbol| {
```
Jarred-Sumner: Let's make this a pointer to avoid an extra unnecessary copy
```suggestion
            for (this.symbols.map.values()) |*symbol| {
```

PR #15444, opened 2024-11-27
zig: make throw use JSError

nektro:


robobun:
@nektro, your commit 0b246deb3b72745fe182df1fdb37de1bc3affff8 is building: [`#6976`](https://buildkite.com/bun/bun/builds/6976)

Jarred-Sumner:
Per IRL discussion, this should be done via bindings generator instead of manually changing each call site to use `try`

Jarred-Sumner:
Jsut the ones involving .get() and the like

paperclover (approved):

PR #15362, opened 2024-11-23
zig: assert there is an exception when .zero is returned

nektro:
cc https://github.com/oven-sh/bun/issues/15141

adds some assertions and fixes the trips

Issue #6977, opened 2023-11-08
Can't find variable: module


jld-adriano:
### What version of Bun is running?

1.0.10+d85bd49d8

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

Made a repro: https://github.com/jld-adriano/bun-repro-4818
Relevant files are index.ts, repro.ts and repro2.ts

just bun index.ts

Basically have a file, import another one that uses the module global variable, and that file imports from another one

### What is the expected behavior?

Shouldn't error

### What do you see instead?

An error

Jarred-Sumner:
Use `import.meta.main`

This is a quirk of our ESM<>CJS interop -- you can use `import` and `require` in the same file, but not `export`

PR #15445, opened 2024-11-27
zig: add JSValue assertion to cppFn

nektro:
followup/similar to https://github.com/oven-sh/bun/pull/15362

### DIFF
--- a/src/bun.js/bindings/bindings.zig
+++ b/src/bun.js/bindings/bindings.zig
@@ -6703,6 +6703,15 @@ pub const EncodedJSValue = extern union {
     asPtr: ?*anyopaque,
     asDouble: f64,
 };
+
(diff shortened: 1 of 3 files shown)


robobun:
@nektro, your commit dff4ae9601f0552afc0e6bdd1e8ab52cd97f137e is building: [`#6977`](https://buildkite.com/bun/bun/builds/6977)

Jarred-Sumner:
Fancy, but this adds many duplicate exception checks. Can we avoid duplicate exception checks?

nektro:
per-irl discussion
`.get` uses a manual `extern fn` so this doesn't overlap there
callsite checks are 100% due for audit/removal/rewrite
previous pr adds check at different boundary so there's no overlap there

nektro:
behavior got in in other prs

PR #15447, opened 2024-11-27
Fix `allowHalfOpen` & `exclusive` flag always being set to true

Jarred-Sumner:
### What does this PR do?

Fix `allowHalfOpen` & `exclusive` flag always being set to true

I think we need better tests for upgrading TLS & allowHalfOpen


robobun:
@Jarred-Sumner, your commit 57cf244dd709ec0575a305ef9a994b7959d54d0f is building: [`#6980`](https://buildkite.com/bun/bun/builds/6980)

PR #15454, opened 2024-11-27
Add "DOM.Iterable" lib to tsconfig-for-init.json

inad9300:
Would this make sense for `bun init`? I see it as certainly convenient, and in alignment with the existing "ESNext" plus "DOM" combo.


inad9300:
Any chance of this making it?

PR #15456, opened 2024-11-27
Fix `fish` completion description for `bun install --no-save`

lgarron:
This doesn't fix other details that are out of date, but makes sure this description is consistent with `bun install --help`, at least.

### What does this PR do?

Fix `fish` completion description for `bun install --no-save`

- [x] Code changes

Ran `bun completions` and sourced to test.


RiskyMH (approved):

RiskyMH:
Thanks!

PR #15458, opened 2024-11-28
fix: Connect with 1 socket to new env var but still work with js debug terminal

alii:
### What does this PR do?

Fixes the connection by using a new environment variable, so the logic stays the same from previous versions of bun with BUN_INSPECT and _NOTIFY

Test with bun 1.1.37, as well as a local release build in both the normal terminal and the debug terminal.

It's possible I've missed a case for testing here, and in an effort to not have another botched release, it would be great if someone else could test this as well locally.


robobun:
@alii, your commit aa46c29bf5dcb1bb06738154152586da32b6f3bd is building: [`#7043`](https://buildkite.com/bun/bun/builds/7043)

Issue #7045, opened 2023-11-11
bun can't install  @~3/svelte_mount in cli


3tieto:
### What version of Bun is running?

1.0.11

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

bun i @~3/svelte_mount

```
bun add v1.0.11 (f7f6233e)
  🔍 Resolving [1/1]
error: InstallFailed cloning repository for @~3/svelte_mou

error: "git clone" for "@~3/svelte_mount" failed


error: "git clone" for "@~3/svelte_mount" failed
````

### What is the expected behavior?

![image](https://github.com/oven-sh/bun/assets/147470591/6e39bac6-193a-417d-baa6-1293c05014bc)

PR #15459, opened 2024-11-28
fix: swap to using child process & output channel over terminal

alii:
### What does this PR do?

Change the test runner to print to an output channel instead of terminal

Test it with the codelens buttons and see it prints to an output channel

### DIFF
--- a/packages/bun-vscode/src/features/tests/index.ts
+++ b/packages/bun-vscode/src/features/tests/index.ts
@@ -1,3 +1,4 @@
+import * as cp from "node:child_process";
 import ts from "typescript";
 import * as vscode from "vscode";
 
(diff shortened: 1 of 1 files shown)


robobun:
@alii, your commit 1485e8068db7bc360bd22d4947469edc63c3b0a0 is building: [`#7045`](https://buildkite.com/bun/bun/builds/7045)

packages/bun-vscode/src/features/tests/index.ts
```diff
@@ -101,9 +102,20 @@ export function registerTestCodeLens(context: vscode.ExtensionContext) {
   );
 }
 
+const outputChannel = vscode.window.createOutputChannel("Bun - Test Results", {
+  log: true,
+});
+
+const appendLines = (chunk: { toString(): string }) => {
+  const lines = chunk.toString().split("\n");
```
Jarred-Sumner, line 110: this will cause some issues with non-ascii characters

instead we should set encoding to utf8 on the stderr/stdout and not call .toString

alii: The reason for the `toString()` was so we can split by lines, as vscode prepends a timestamp to every log. Splitting them meant we could have every log lined horizontally up

Jarred-Sumner: Splitting is fine. The issue is these are Buffer objects which means it can be an incomplete unicode sequence. Using string_decoder will prevent partially written unicode from appearing in terminal output. Settings the encoding on the stdout & stderr will fix it by only returning already decoded unicode

Issue #7069, opened 2023-11-12
[webSocket server] : Access getTopics on websocket


Vincz:
### What is the problem this feature would solve?

The feature would solve the problem of accessing the list of topics a web socket has subscribed to.
The method already exists in `uWebsocket`.

### What is the feature you are proposing to solve the problem?

Just give access to the underlying method `getTopics` of `uWebsocket`

### What alternatives have you considered?

Using `uWebsocket.js` directly.

ShaharIlany:
Does anyone have a solution/workaround for this issue?
I would love to see the underlying `getTopics` being exposed on the API

PR #15467, opened 2024-11-28
bake(dev): plugins in dev server, with other fixes

paperclover:
- load plugins in dev mode
- pass `resolveDir` to onResolve plugins
- onLoad plugin with css loader now works
- hmr instrumentation for `export { x } from 'y';`
- fixed `export { x as y }`
- throw SyntaxError `The requested module '${id}' does not provide an export named '${key}'`
- stub `import.meta.hot`, implement `import.meta.hot.dispose`

example project: https://github.com/paperdave/bake-svelte-plugin-example


robobun:
@paperdave, your commit 5c84c760819a0c0922aacc051e919950f6aade0d is building: [`#7069`](https://buildkite.com/bun/bun/builds/7069)

src/bake/DevServer.zig
```diff
@@ -2034,15 +2045,16 @@ pub fn IncrementalGraph(side: bake.Side) type {
             const dev = g.owner();
             dev.graph_safety_lock.assertLocked();
 
-            const abs_path = ctx.sources[index.get()].path.text;
+            const path = ctx.sources[index.get()].path;
```
Jarred-Sumner, line 2048: Is this copy intentional? Can it be a pointer instead? `Path` is not a small struct
```suggestion
            const path = &ctx.sources[index.get()].path;
```

src/bake/bake.zig
```diff
@@ -86,11 +91,62 @@ const StringRefList = struct {
...
+
+            const function = try plugin_config.getFunction(global, "setup") orelse {
+                return global.throwInvalidArguments("Expected plugin to have a setup() function", .{});
+            };
+            const plugin_result = try plugin.addPlugin(function, empty_object, .null, false, true);
+            if (plugin_result.asAnyPromise()) |promise| {
+                promise.setHandled(global.vm());
+                // TODO: remove this call, replace with a promise list that must
+                // be resolved before the first bundle task can begin.
+                global.bunVM().waitForPromise(promise);
```
Jarred-Sumner, line 137: sad

src/bake/bake.zig
```diff
@@ -521,6 +583,10 @@ pub const Framework = struct {
         if (renderer == .server and framework.server_components != null) {
             try out.options.conditions.appendSlice(&.{"react-server"});
         }
+        if (mode == .development) {
+            // Support `esm-env` package using this condition.
+            try out.options.conditions.appendSlice(&.{"development"});
```
Jarred-Sumner, line 588: i've never heard of this

src/bake/hmr-module.ts
```diff
@@ -76,7 +91,79 @@ if (side === "server") {
 }
 
 function initImportMeta(m: HotModule): ImportMeta {
```
Jarred-Sumner, line 93: This is client-side, for browsers I assume?

src/bun.js/api/JSBundler.zig
```diff
@@ -873,22 +820,18 @@ pub const JSBundler = struct {
...
-                this.deinit();
-                return;
-            };
             if (source_code_value.isEmptyOrUndefinedOrNull() or loader_as_int.isEmptyOrUndefinedOrNull()) {
                 this.value = .{ .no_match = {} };
 
                 if (this.was_file) {
                     // Faster path: skip the extra threadpool dispatch
-                    completion.bundler.graph.pool.pool.schedule(bun.ThreadPool.Batch.from(&this.parse_task.task));
+                    this.bv2.graph.pool.pool.schedule(bun.ThreadPool.Batch.from(&this.parse_task.task));
```
Jarred-Sumner, line 828: bv2 implies the existence of bv3 and bv1

is there a better name?

src/bundler/bundle_v2.zig
```diff
@@ -1783,13 +1835,15 @@ pub const BundleV2 = struct {
                 this.graph.heap.gc(true);
             }
         }
-        var log = &load.completion.?.log;
+        const log = this.bundler.log;
```
Jarred-Sumner, line 1838: Why was this changed? This will crash if the bundler thread writes to `logger.Log` at the same time as the main thread. When using `Bun.build` that can happen.

src/bundler/bundle_v2.zig
```diff
@@ -1846,7 +1922,7 @@ pub const BundleV2 = struct {
                 this.graph.heap.gc(true);
             }
         }
-        var log = &resolve.completion.?.log;
+        const log = this.bundler.log;
```
Jarred-Sumner, line 1925: Ditto. We shouldn't change this unless we're confident there's no thread-safety issue.

src/bundler/bundle_v2.zig
```diff
@@ -2116,8 +2195,8 @@ pub const BundleV2 = struct {
 
                         // Discover all CSS roots.
                         for (import_records.slice()) |*record| {
-                            if (record.tag != .css) continue;
                             if (!record.source_index.isValid()) continue;
+                            if (loaders[record.source_index.get()] != .css) continue;
```
Jarred-Sumner, line 2199: This is an extra pointer lookup, which slightly slows things down. Was the `import_record.tag` inaccurate?

src/bundler/bundle_v2.zig
```diff
@@ -2595,7 +2666,7 @@ pub const BundleV2 = struct {
                     import_record.path.pretty = rel;
                     import_record.path = this.pathWithPrettyInitialized(path.*, target) catch bun.outOfMemory();
                     if (entry.kind == .css) {
-                        import_record.tag = .css;
+                        import_record.path.is_disabled = true;
```
Jarred-Sumner, line 2669: Can you leave a comment explaining this?

src/bundler/bundle_v2.zig
```diff
@@ -6241,11 +6313,12 @@ pub const LinkerContext = struct {
                         visit(
                             c,
                             import_records,
+                            loaders,
                             temp,
                             visits,
                             o,
                             record.source_index,
-                            record.tag == .css,
+                            loaders[record.source_index.get()] == .css,
```
Jarred-Sumner, line 6321: If we can continue to use record.tag that's one less pointer lookup

src/bundler/bundle_v2.zig
```diff
@@ -10948,6 +11022,21 @@ pub const LinkerContext = struct {
             .{ .binding = Binding.alloc(allocator, B.Identifier{ .ref = ast.module_ref }, Logger.Loc.Empty) },
         });
         const module_id = Expr.initIdentifier(ast.module_ref, Logger.Loc.Empty);
+
+        // add a marker for the client runtime to tell that this is an ES module
+        if (ast.exports_kind == .esm) {
+            try stmts.inside_wrapper_prefix.append(Stmt.alloc(S.SExpr, .{
```
Jarred-Sumner, line 11028: This would ideally use the lazy export AST and then it can choose whichever output format is best.

src/js/builtins/BundlerPlugin.ts
```diff
@@ -334,12 +335,12 @@ export function runOnResolvePlugins(this: BundlerPlugin, specifier, inputNamespa
   }
 }
 
-export function runOnLoadPlugins(this: BundlerPlugin, internalID, path, namespace, defaultLoaderId) {
+export function runOnLoadPlugins(this: BundlerPlugin, internalID, path, namespace, defaultLoaderId, isServerSide: boolean) {
```
Jarred-Sumner, line 338: Maybe `isServerSide` could be an enum to clarify whether it means:
- RSC
- SSR
- !browser

src/js_parser.zig
```diff
@@ -19093,11 +19080,10 @@ fn NewParser_(
                             return;
                         }
                     } else {
-
                         // This is a re-export and the symbols created here are used to reference
                         for (data.items) |*item| {
                             const _name = p.loadNameFromRef(item.name.ref.?);
-                            const ref = try p.newSymbol(.other, _name);
+                            const ref = try p.newSymbol(.import, _name);
```
Jarred-Sumner, line 19086: ooh i wonder what bugs this causes

src/logger.zig
```diff
@@ -96,13 +96,24 @@ pub const Loc = struct {
...
+    /// Line <= 0 means there is no line and column information.
+    // TODO: move to `bun.Ordinal`
+    line: i32,
+    // TODO: figure out how this is interpreted, convert to `bun.Ordinal`
+    // original docs: 0-based, in bytes.
+    // but there is a place where this is emitted in output, implying one based character offset
+    column: i32,
+    /// Number of bytes this location should highlight.
+    /// 0 to just point at a single character
+    length: usize = 0,
```
Jarred-Sumner, line 111: yeah this struct wasn't thoughtfully created, it was copypasta from esbuild

src/sourcemap/sourcemap.zig
```diff
@@ -607,6 +607,9 @@ pub const ParseResult = union(enum) {
                 .location = Logger.Location{
                     .file = path,
                     .offset = this.loc.toUsize(),
+                    // TODO: populate correct line and column information
+                    .line = -1,
```
Jarred-Sumner, line 611: We need to port the LineColumnTracker abstraction esbuild has

Jarred-Sumner (changes requested):
Some comments. Only blocking comment is the change to reference `bundler.log`, as I worry it will cause crashes in `Bun.build`.

Jarred-Sumner:
I’m going to merge this to unblock further progress, but please take a look at the comments when you get a chance.

PR #15478, opened 2024-11-28
Add request object to Bun.serve error Handler

### LINKED ISSUE #15475: get data from request in Bun.serve Error handler

https://discord.com/channels/876711213126520882/1310569549338247178

> Sancho Godinho
OP
 — 25.11.2024 12:35
I am making a certain site in which I need error handling requiring access to header data. How can I achieve this?

pass the original header to the error handler as second argument

> Jarred 
careful with the PR, it might be difficult to do consistently because it’s easy to free the request object too early
That was the main reason we didn’t do this. Making sure we don’t cause a either a use after free or a memory leak might be difficult to do in each scenario
But I don’t think we actually tried
so might be imagining it to be harder than it is

I'm gonna considering to pass a less populated object only holding URI and Headers if the original object is too complicated to pass

Assign me to this topic please

Kapsonfire:
### What does this PR do?

Fixes #15475 

This PR adds the js request object to the error handler for Bun.serve

- [x] Code changes

[ ] Write Tests


Kapsonfire-DE:
sorry, something is mismatching here

PR #15496, opened 2024-11-29
docs: shorten install commands

okineadev:
### What does this PR do?

Shortened links from `https://bun.sh/install` to `bun.sh/install`


okineadev:
@Serdnad What's wrong here?

Serdnad:
@okineadev curl defaults to HTTP if no scheme (HTTPS vs HTTP) is explicitly specified. Worst case scenario, you run this command and suffer from a MITM attack, which gets directly piped into `bash`. More likely, you hit a different URL than actually intended, and get a useless response - which is what happens here, compare the output from `curl bun.sh/install` and `curl https://bun.sh/install`.

okineadev:
> @okineadev curl defaults to HTTP if no scheme (HTTPS vs HTTP) is explicitly specified. Worst case scenario, you run this command and suffer from a MITM attack, which gets directly piped into `bash`. More likely, you hit a different URL than actually intended, and get a useless response - which is what happens here, compare the output from `curl bun.sh/install` and `curl https://bun.sh/install`.

You are right

`curl -L bun.sh/install`

okineadev:
Closing due to the threat of MITM attacks.

PR #15498, opened 2024-11-29
fix 14540

### LINKED ISSUE #14540: `bun build` crash

I have no clue what is going on here. I've tried to pull out as much code as possible but it seems to work in random situations and fail in others, hopefully this is enough for you guys to go on.

```ts
// repro.ts
import { file } from "bun";

// prettier-ignore
const expectHeader = [] as const

type DNARecord = Record<(typeof expectHeader)[number], string>;

async function read_dna_request_csv(csvData: string) {
  const rows = await read_as_array(csvData);

    if (rows[0][i] !== expectHeader[i]) {
      throw new Error(
        `Invalid header; expect ${expectHeader[i]}, got ${rows[0][i]}`
      );
    }
  }

  const recs = [];
  for (let i = 1; i < rows.length; i++) {
    const rec = {};
    for (let j = 0; j < expectHeader.length; j++) {
      if (rows[i][j]) rec[expectHeader[j]] = rows[i][j];
    }
    recs.push(rec);
  }
  return recs;
}

const jobData = await file(process.env.DATA!).json();
if (!jobData.dnaRequests) throw new Error("No requests found in job data");
const recs: DNARecord[] = [];
const filename: string = jobData.filename;
const fileDataBin = atob(jobData.file);
const fileDataUint8Arr = Uint8Array.from(
  { length: fileDataBin.length },
  (_, i) => fileDataBin.charCodeAt(i)
);
if (filename.toLowerCase().endsWith(".zip")) {
  // Read as ZIP
  const r = new ZipReader(new Uint8ArrayReader(fileDataUint8Arr));
  for (const e of await r.getEntries()) {
    if (e.getData) {
      const text = await e.getData(new TextWriter());
      recs.push(...(await read_dna_request_csv(text)));
    }
  }
} else {
  // Read as CSV
  const fileData = new TextDecoder().decode(fileDataUint8Arr);
  recs.push(...(await read_dna_request_csv(fileData)));
}

// Set up API client
let url = process.env.URL;
if (!url) throw new Error("not set");
if (!url.endsWith("/api/v1")) url += "/api/v1";
const token = process.env.TOKEN;
if (!token) throw new Error("TOKEN not set");

const byBatch = new Map<string, DNARecord[]>();
for (const rec of recs) {
  const batch = rec["BATCH-NO"];
  if (!byBatch.has(batch)) byBatch.set(batch, []);
  byBatch.get(batch)!.push(rec);
}

const uploaded: string[] = [];
for (const [batch, recs] of byBatch.entries()) {
  // Sanity check: every record should have the same herd
  if (!recs.every((r) => r.HERD === recs[0].HERD)) {
    const herds = new Set(recs.map((r) => r.HERD));
    throw new Error(
      `Batch ${batch} has multiple herds (${Array.from(herds).join(", ")})`
    );
  }

  await client.testrequests.upsertTestRequest({
    tmp: {
      name: batch,
      entityID: entity.internalID,
      status: "accepted",
      type: "1", // some comment
      samples: recs.map((r): TestRequestUploadSample => {
        let bundle = "3";
        const sex = parse_sex(r.SEX, cfg.term);
        if (
          (sex === "MALE" || sex === "CASTRATEDMALE") &&
          age < 425 /* 14 months */
        ) {
          bundle = "1"; // Registered male up to 14 months
        } else if (sex === "FEMALE" && age < 730 /* 24 months */) {
          bundle = "2"; // Registered female up to 24 months
        }

        return {};
      }),
    },
  });
}
```

given this script (you shouldn't need to install deps) if you run `bun build  --compile --minify --sourcemap ./repro.ts --outfile repro` it crashes with:
```sh
Bun v1.1.30 (7996d06b) macOS Silicon
macOS v15.0.1
Args: "bun" "build" "--compile" "--minify" "--sourcemap" "./repro.ts" "--outfile" "repro"
Features: define tsconfig tsconfig_paths 
Elapsed: 3ms | User: 5ms | Sys: 4ms
RSS: 13.03MB | Peak: 13.03MB | Commit: 1.07GB | Faults: 38

panic: Segmentation fault at address 0x400000001

 https://bun.report/1.1.30/Mb17996d06Ah4////B_myjsBmyjsBmj8vJ__A2IC

[1]    15989 trace trap  bun build --compile --minify --sourcemap ./repro.ts --outfile repro
```

Note that as I've changed things (pulled more out and so on) the actual error has changed but I imagine it's all the same underlying issue..

### Relevant log output

_No response_

### Stack Trace (bun.report)

Bun v1.1.30 ([`7996d06`](<https://github.com/oven-sh/bun/tree/7996d06b8f5869ccc9ad9a421cd06469da4273fc>)) on macos aarch64 [BuildCommand]

Segmentation fault at address 0x400000001

- *1 unknown/js code*
- `src.thread_pool.Do__anon_71003.RunnerTask.call`
- `src.thread_pool.Do__anon_71003.RunnerTask.call`
- `Thread.PosixThreadImpl.spawn__anon_112062.Instance.entryFn`

Features: define, tsconfig, tsconfig\_paths, tsconfig\_paths

dylan-conway:
### What does this PR do?
fixes #14540 

Fixes crashes caused by invalid pointers copied from old lexer state.

Added a test that relies on `std.ArrayList` capacity growth


robobun:
@dylan-conway, your commit 39695ae93d401ad42d9e2aac104dffcfd895cedf is building: [`#7113`](https://buildkite.com/bun/bun/builds/7113)

src/js_lexer.zig
```diff
@@ -273,6 +273,27 @@ fn NewLexer_(
             // }
         }
 
+        pub fn restore(this: *LexerType, original: *const LexerType) void {
```
Jarred-Sumner, line 276: The length is not the main problem - the main problem is on array resize, the previous pointer is now invalid memory. This wouldn't fix it, as it reuses the memory. 

I think we would instead need to lazily clone it

dylan-conway: it copies the possibly new pointer from the new lexer, and sets the length from the original lexer. This way we can ensure the pointers are valid, new values aren't used, and memory isn't wasted.

doing this for `temp_buffer_u16` is unnecessary though, it's length is set to 0 after each use so we can ignore it. Will update

Jarred-Sumner: the pointer can become invalid memory though when its resized later. even if the ptr and length match up.

PR #15510, opened 2024-11-30
Introduce experimental flag: `bun build --auto-install <entry points...>`

Jarred-Sumner:
### What does this PR do?

This implements [auto install](https://bun.sh/docs/runtime/autoimport) for `bun build`. 

When using bun's bundler, you can now automatically install dependencies -- and only the dependencies the bundled code is using. This means you can skip running `bun install` (or npm/yarn/pnpm install).

This can lead to smaller Docker containers and faster CI times than using something like `turbo prune` or `"devDependencies"` in package.json because dependencies which are not used by any code will not be installed.

To use it:
```sh
bun build --auto-install <entry-points...>
bun build -i <entry-points>
```

If a `bun.lockb` or a package-json.lock is present, it will continue to use the resolved versions from the lockfile (and will still save space because it only installs those which are referenced). If no lockfile is present, it will prefer the versions from package.json.

Tradeoffs:
- This doesn't run any postinstall scripts
- This doesn't support patch (though that could be fixed in the future)
- If dependencies are not imported or required in any code visible to `bun build`, it will not be installed

This uses bun's global cache, which conserves disk space by avoiding a per-project node_modules folder.

TODO before merging:
- [ ] Handle manifest download errors
- [ ] Handle package download errors
- [ ] Investigate difficulty of supporting Bun.build()
- [ ] Be a little smarter about resolving package versions
- [ ] Some tests

### DIFF
--- a/src/bake/production.zig
+++ b/src/bake/production.zig
@@ -47,14 +47,7 @@ pub fn buildCommand(ctx: bun.CLI.Command.Context) !void {
     vm.argv = ctx.passthrough;
     vm.arena = &arena;
     vm.allocator = arena.allocator();
-    b.options.install = ctx.install;
-    b.resolver.opts.install = ctx.install;
-    b.resolver.opts.global_cache = ctx.debug.global_cache;
-    b.resolver.opts.prefer_offline_install = (ctx.debug.offline_mode_setting orelse .online) == .offline;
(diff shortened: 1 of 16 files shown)


robobun:
@Jarred-Sumner, your commit def017a4b403cd65098ecd78204306da75e5895e is building: [`#7126`](https://buildkite.com/bun/bun/builds/7126)

joelmoss:
What's the status of this? I would love to be able to use auto-install with `build` 🤩

Issue #11191, opened 2024-05-20
bun build --target=bun inlines environment variable and .env file values


GeoffWilliams:
### What version of Bun is running?

1.1.8+89d25807f

### What platform is your computer?

Linux 6.1.0-17-amd64 x86_64 unknown

### What steps can reproduce the bug?

`bun build --target=bun` inlines the value of environment variables set in the environment and `.env`, eg:

**.env**

```env
MYVAR=gotcha
```

**test.ts**

```typescript
console.log(process.env.MYVAR);
console.log(process.env.MYOTHERVAR);
```

**bundle**

```shell
export MYOTHERVAR=hmmm
bun build --outdir bundled test.ts --target=bun
```

### What is the expected behavior?

Expect to see directly specified environment variables used when supplied so that I can use environment variables for configuration/externalization of secrets. Running typescript directly with `bun` works:

```shell
MYVAR=hello MYOTHERVAR=world bun test.ts 
hello
world
```

### What do you see instead?

Values from `.env` **and the environment** are inlined into the bundle/compiled executable when `--target=bun` used. Referenced environment variables that had no value during build work as expected:

```shell
MYVAR=hello MYOTHERVAR=world bun ./bundled/test.js 
gotcha
hmmm
``````

### Additional information

Unwanted inlining will embed secrets into bundles/executables. I already did this without realising:

```
grep $TEST_BINANCE_FUTURES_API_KEY build/app 
grep: build/app: binary file matches
```

The behaviour of `target` is documented at https://bun.sh/docs/bundler#target but does not mention anything about inlining environment variables and `.env` files

I guess the workaround would be to run `bun compile` in a clean environment and move any `.env` files out of the way first

**Comment by GeoffWilliams** [comment]
Workaround:

```shell
mv .env .env.nouse
env -i ~/.bun/bin/bun build --compile --sourcemap src/app.ts --outfile build/app 
mv .env.nouse .env
```

benpsnyder:
I am also experiencing this issue.
Here is the environment variable
![image](https://github.com/oven-sh/bun/assets/707567/dbc7df04-03e9-4fe9-b88c-05064205a8da)

but it in-lined the .env from a development build
![image](https://github.com/oven-sh/bun/assets/707567/d0cca269-6b13-429a-8d00-238af45aba50)

the `.env` should never be included in builds, imho, or we need a CLI flag

McGaelen:
This also happens when creating a single-file exectuable. Would love to see this fixed.

[...]

Issue #12695, opened 2024-07-20
`bun build` inlines `process.env` when `--target` is set to `bun` is this intentional?


huseeiin:
### What version of Bun is running?

1.1.20+ae1948925

### What platform is your computer?

Linux 6.5.0-18-generic x86_64 x86_64

### What steps can reproduce the bug?

```js
// file.js

console.log(process.env.HOME);
```
```
bun build file.js --target bun --outdir dist
```

### What is the expected behavior?

This can be extremely harmful if you have something like `const IS_PROD = process.env.NODE_ENV === 'production'` because its going to be transpiled to `const IS_PROD = false` by default

### Additional information

1. This doesn't happen when using `Bun.env` instead
1. This doesn't happen when you set `--target` to `node` or `browser` (or leave `--target` empty)

RiskyMH:
Duplicate of #11191

BadMachine:
When runnung
bun --env-file=.env build --env=inline --minify --target=bun src/main.ts --outfile dist/main.js

no env variables inlined in the built file

Issue #8125, opened 2024-01-12
Bun build "--target bun" should not embed process.env vars


onionhammer:
### What version of Bun is running?

1.0.22

### What platform is your computer?

x86, Ubuntu 22.04.3 LTS

### What steps can reproduce the bug?

run `bun build --target bun index.ts` on an index.ts file which sets a variable from process.env.MY_ENV_VAR

### What is the expected behavior?

process.env.MY_ENV_VAR should be in the output, not the value being swapped

### What do you see instead?

I see the value of MY_ENV_VAR. This makes this code less deployable. The target is `bun` runtime, so it should be able to pull the env var at runtime

Electroid:
We made a change in the runtime to not do this, we might consider doing that in the bundler too.

bitfrost:
Seems super dangerous and will most likely cause security issues / sensitive environment leaks for folks who walk into this behavior.   Please consider making inline'ing opt in or via some sort of documented prefix.

RiskyMH:
This should be better now with the new `--env=disable` flag (will be default in soon Bun 1.2)

https://bun.sh/docs/bundler#env

RiskyMH:
v1.2 is released, so this shouldn't occur by default anymore.

Issue #7434, opened 2023-12-03
Global env vars are hardcoded to release build if building to a single executable


kerambit:
### What version of Bun is running?

1.0.15+b3bdf22eb

### What platform is your computer?

Darwin 23.0.0 arm64 arm

### What steps can reproduce the bug?

Create index.js file with code `console.log(process.env)`

Update .bash_profile or .zshrc to add new global vars. For example `export DEV_EMAIL=kerambit@mymail.com`

`$ bun build ./index.js --compile outfile myfile`

`$ ./myfile`

Update .bash_profile or .zshrc and change our recently added env var for example to `export DEV_EMAIL=kerambit+123456@mymail.com` and save it

Open new terminal window

`$ ./myfile`

### What is the expected behavior?

See an object that contains field DEV_EMAIL: kerambit@mymail.com after first execution and see an object that contains field DEV_EMAIL: kerambit+123456@mymail.com after second execution

### What do you see instead?

See an object that contains field DEV_EMAIL: kerambit@mymail.com after first and second execution

### Additional information

Looks  like this is not usual behaviour. Let's check up another langs and their builds:

Create main.rs and put this code here

```rust
use std::env;

fn main() {
    for (key, value) in env::vars() {
        println!("{}: {}", key, value);
    }
}

```

Update .bash_profile or .zshrc to add new global vars. For example `export DEV_EMAIL=kerambit@mymail.com`

`$ rustc main.rs`

`$ ./main`

Update .bash_profile or .zshrc and change our recently added env var for example to `export DEV_EMAIL=kerambit+123456@mymail.com` and save it

Open new terminal window

`$ ./main`

kerambit:
@Electroid hello.
this is still actual ?
looks like there are no such bug in 1.0.29. revision is 1.0.29+a146856d1

Electroid:
This is still an issue.

```js
// index.js
let thing = process.env.COOL_THING;
console.log(thing);
```

```sh
COOL_THING=1 bun build index.js --compile
```

```sh
./index # actual=1, expected=undefined 
COOL_THING=2 ./index # actual=1, expected=2
```

RiskyMH:
We now have [`--env=disable`](https://bun.sh/docs/bundler#env) as of [`v1.1.39`](https://bun.sh/blog/bun-v1.1.39#inject-environment-variables-in-bun-build). Also as of `1.2` it shouldn't inline by default.

PR #15513, opened 2024-11-30
fix(parser): rewrite process.env to bun.env when compiling

kmate19:
fixes: #11191 , #7434 , #8125 , #12695

### What does this PR do?

Fixes issue where process.env becomes inlined after compiling to an executable, not actually being read from the environment

- [x] Code changes

# Tested manually with:

// index.js
```
const a = process.env.TEST_ENV_VAR;
const b = Bun.env.TEST_ENV_VAR;

a === b ? console.log('process and bun env are equal') : console.log('process and bun env are not equal');
```

Then ran:
```
export TEST_ENV_VAR=test
bun-debug build --compile --target=bun --outfile test ./index.js
./test
```
^ output: process and bun env are equal
then:
```
export TEST_ENV_VAR=test2
./test
```
^ output: process and bun env are equal

Doing the same thing on the release build of bun, after changing the variable the two are not equal.

### DIFF
--- a/src/bun.js/RuntimeTranspilerCache.zig
+++ b/src/bun.js/RuntimeTranspilerCache.zig
@@ -8,7 +8,8 @@
 /// Version 9: String printing changes
 /// Version 10: Constant folding for ''.charCodeAt(n)
 /// Version 11: Fix \uFFFF printing regression
-const expected_version = 11;
+/// Version 12: Fix process.env not behaving correctly when compiling
+const expected_version = 12;
 
 const bun = @import("root").bun;
 const std = @import("std");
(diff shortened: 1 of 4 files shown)


XavierGeerinck:
Would love to get this merged! Any update on this? Got the same issue, when working with `process.env` it compiles the environment variable defaults into my application, which I don't want.

paperclover (changes requested):
thank you for the pr, but this is not the correct solution. it shouldnt need to make `js_lexer`'s `peek` function public

i think the actual solution is to add a CLI argument for `env: 'disable'`, since env inlining is an intended feature, but i don't know if it was intended to run on --target=bun by default
https://bun.sh/docs/bundler#env

RiskyMH:
We have [`--env=disable`](https://bun.sh/docs/bundler#env) as of [`v1.1.39`](https://bun.sh/blog/bun-v1.1.39#inject-environment-variables-in-bun-build) which does what Paper was saying. 

Also as of `1.2` it shouldn't inline by default now.

Issue #7144, opened 2023-11-15
`expect(...)toContain("")` handles `""` incorrectly


CharString:
### What version of Bun is running?

1.0.11+f7f6233ea

### What platform is your computer?

Linux 6.1.0-0.deb11.11-amd64 x86_64 unknown

### What steps can reproduce the bug?

```console
$ bun add -d fast-check
```
Create `string.test.ts` with this contents

```ts
import { describe, it, expect, test } from "bun:test";
import fc from "fast-check";

describe("The empty string", () => {
  const emptyString = "";
  it("has a 0 length", () => expect(emptyString).toHaveLength(0));
  it("is found in all strings", () => {
    fc.assert(
      fc.property(fc.string(), (string: string) => {
        expect(string.indexOf(emptyString)).toBe(0);
        expect(string).toContain(emptyString);
      })
    );
  });

  describe('forms the (strings, concatenation, "") monoid', () => {
    const cat = (a: string, b: string) => a + b;

    test("concatenation is associative on string", () => {
      fc.assert(
        fc.property(
          fc.string(),
          fc.string(),
          fc.string(),
          (a: string, b: string, c: string) =>
            expect(cat(cat(a, b), c)).toBe(cat(a, cat(b, c)))
        )
      );
    });

    test('concatenation with "" is identity', () => {
      fc.assert(
        fc.property(fc.string(), (a: string) => {
          expect(cat(a, emptyString)).toBe(a);
          expect(cat(emptyString, a)).toBe(a);
        })
      );
    });
  });
}
```

### What is the expected behavior?

These tests should pass.

### What do you see instead?

```console
bun test v1.0.11 (f7f6233e)

string.test.ts:
121 |     }
122 |     return defaultReportMessageInternal(out, stringifySecond);
123 | }
124 | function buildError(errorMessage, out) {
125 |     if (!out.runConfiguration.errorWithCause) {
126 |         throw new Error(errorMessage);
                  ^
error: Property failed after 1 tests
{ seed: 2018390894, path: "0:0:0:0", endOnFailure: true }
Counterexample: ["","",""]
Shrunk 3 time(s)
Got error: Property failed by returning false

Hint: Enable verbose mode in order to have the list of all failing values encountered during the run
      at buildError (/home/chars/tmp/bun_playground/node_modules/fast-check/lib/esm/check/runner/utils/RunDetailsFormatter.js:126:14)
[...]

PR #15521, opened 2024-12-01
Preserve authorization headers in bun install when redirects are same hostname

### LINKED ISSUE #15516: Issue with Authorization Token Missing on Redirect for Private Registry in bun install

1.1.38+bf2f153f5

Linux 6.11.8-200.fc40.x86_64 x86_64

1. Set up a private npm registry (e.g., Nexus) that requires an authorization token for access.
2. Add a package from the private registry to your project dependencies (e.g., a private package in `package.json`).
3. Configure the private registry in `bunfig.toml` or .`npmrc` with the correct token.
4. Run `bun install` to install dependencies.

`bun install` successfully follows the redirect (HTTP 302) while preserving the Authorization header and installs the package from the private registry.

### What do you see instead?

The Authorization header is not preserved after the redirect, resulting in a 401 Unauthorized error when accessing the final URL.

### Logs:

**Here is an example of the request/response flow from the logs:**

```
[fetch] > HTTP/1.1 GET http://your-private-registry/repository/group-npm/package-name/-/package-name.tgz
[fetch] > Authorization: Bearer NpmToken.xxxxxx
...

[fetch] < 302 Found
[fetch] < location: https://your-private-registry/repository/group-npm/package-name/-/package-name.tgz

[fetch] > HTTP/1.1 GET https://your-private-registry/repository/group-npm/package-name/-/package-name.tgz
[fetch] < 401 Unauthorized
```

### Possible Cause:

Bun does not retain the Authorization header when following redirects, causing the token to be omitted in the final request to the private registry.

### Why This Is Critical:
This issue prevents users from installing dependencies from private registries that use redirects, which is common for services like Nexus.

Jarred-Sumner:
### What does this PR do?

Preserve authorization headers in bun install when redirects are same hostname

From [make-fetch-happen](https://github.com/npm/cli/blob/fefd509992a05c2dfddbe7bc46931c42f1da69d7/node_modules/make-fetch-happen/lib/fetch.js#L64-L71):
```js
  // Remove authorization if changing hostnames (but not if just
  // changing ports or protocols).  This matches the behavior of request:
  // https://github.com/request/request/blob/b12a6245/lib/redirect.js#L134-L138
  if (new url.URL(request.url).hostname !== redirectUrl.hostname) {
    request.headers.delete('authorization')
    request.headers.delete('cookie')
  }
```
From that comment:
```js
    if (request.uri.hostname !== request.originalHost.split(':')[0]) {
        // Remove authorization if changing hostnames (but not if just
        // changing ports or protocols).  This matches the behavior of curl:
        // https://github.com/bagder/curl/blob/6beb0eee/lib/http.c#L710
        request.removeHeader('authorization')
      }
```

Sadly, the perma link there is not so perma

Fixes #15516 

Untested. Needs a test

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -1990,6 +1990,25 @@ pub const HTTPVerboseLevel = enum {
     curl,
 };
 
+pub const RedirectHeadersBehavior = enum {
+    /// fetch() spec treats origin as ${protocol}://${hostname}:${port}
+    fetch,
+
+    /// curl treats origin as ${hostname}
+    curl,
+
+    /// always preserve headers
+    always,
+
+    pub fn isSameOrigin(this: RedirectHeadersBehavior, url: *const URL, new_url: *const URL) bool {
+        return switch (this) {
+            .fetch => strings.eqlCaseInsensitiveASCII(strings.withoutTrailingSlash(new_url.origin), strings.withoutTrailingSlash(url.origin), true),
+            .curl => strings.eqlCaseInsensitiveASCII(url.hostname, new_url.hostname, true),
+            .always => true,
+        };
+    }
+};
+
 pub const Flags = packed struct {
     disable_timeout: bool = false,
     disable_keepalive: bool = false,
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit c989e4780c476790e912f4bda9d10241e371e241 is building: [`#7144`](https://buildkite.com/bun/bun/builds/7144)

PR #15525, opened 2024-12-01
Support codesigning macOS executables in `bun build --compile`

Jarred-Sumner:
### What does this PR do?

This does two things:
1. Support codesigning macOS executables in `bun build --compile`
2. Deletes the hack where we read the last 4096 bytes of the currently running executables itself to see if it ends with "-- Bun --\n" . We now add a `__BUN,__bun` section in the macho executable so all we need to do is a single pointer lookup to find out if the current executable is a standalone executable and to get the data. This will also reduce memory usage because we no longer need to do another memory allocation to read through the standalone executable's code. 

Must TODO before merging:
- Handle > 16 KB right now. It's something to do with the macho section offsetting/rewriting. I'm too sleepy to fix it right now.


robobun:
@Jarred-Sumner, your commit 7c034acc5dacca1605b9931c6c6e49d2c9a94ddb is building: [`#7145`](https://buildkite.com/bun/bun/builds/7145)

Jarred-Sumner:
> @Jarred-Sumner is there anything I can do to make progress here? My understanding is that there's the case of the macho section not being in the expected place? 🤔
> 
> If there's any chance you can share a bit more on where this is stuck, perhaps I can try to move this along (not super experienced with Zig, more of a Rust guy)

It’s something with codesigning the executable on arm64 macOS. It works on small executables. But when the sections need to be resized the offsets in the sections aren’t being updated correctly

Pranav2612000:
Hey @Jarred-Sumner ! I'll like to give this a shot. Here's my current progress -
- I've been able to compile and build bun, get this up to date with the main branch ( [Ref](https://github.com/Pranav2612000/bun/tree/jarred/macho) ) and have been able to play around a bit with the `StandaloneModuleGraph` to get an idea of what it works. ( Would love some more hints around this if you have them )
- As we know, the Bun binaries generated by this version of bun are also not codesignable ( and very huge 400Mb for hello world ). 
- From your comment above, I understand that the offsets need to be resized along with the sections. Does that mean that in this case, the `12af8000` section offset is the one calculated incorrectly? image


- To debug which section has problems, I also tried creating another binary with an empty bun section i.e the `macho` stuff in `inject` commented out ( thought this could give some insights even if the actual script doesn't run ). ( [commit](https://github.com/Pranav2612000/bun/commit/b51b0573a99203a343c874425c8e2ae99604e6d4) ). This one seems to have generated a code-signable version. image





Also how were you able to generate a 16Kb executable to test out everything was working? Any tips or suggestions on how I can proceed further?

Pranav2612000:
@Jarred-Sumner I think I've fixed the issue now. This [commit](https://github.com/Pranav2612000/bun/commit/e4bab9cc5027bb484e17612a7544282425244f2f) has a fix for the problem with the offsets. ( You can try cherrypicking the commit to test things out ).

The reason was that we were previously writing the signatures from the `aligned_signature_offset` location. This resulted in the data overflowing outside the address space. Especially the signature data overflows outside the `length` set up in the `CODE_SIGNATURE` segment and `Super_Blob` section. This causes the MachO file to be corrupt resulting in `main executable failed strict validation` error.

Since this branch is a lot out of date with the `main` branch, I'll create separate PRs for the fixes and we can decide how to proceed further. I'll love to put in more effort to ensure that these fixes get merged.

Pranav2612000:
Here are the PRs to make it easier to view the diffs and test things out
Main branch: https://github.com/oven-sh/bun/pull/17207 
Fix for this PR: https://github.com/oven-sh/bun/pull/17210

Pranav2612000:
Hey @Jarred-Sumner ! I've updated the PRs to add fix for larger files.
Some things I had to do
- Add the hashes for the updated .bun section size in the CODE_SIGNATURE section, LINKEDIT segment and the final file
- Update the offsets and sizes of the LINKEDIT segment with newer sizes
- Update the datasize of CODE_SIGNATURE section.


You can view these changes here -> https://github.com/oven-sh/bun/pull/17210/commits/6747b3865e884b58b32249d2fbb90692a9312d59

Jarred-Sumner:
This was shipped in Bun v1.2.4

PR #15539, opened 2024-12-02
Fixes #14433 - global .npmrc not using auth

### LINKED ISSUE #14433: Global .npmrc not using authentication for custom registry

1.1.30+7996d06b8

Darwin 23.6.0 arm64 arm

Use a global config like this:

```
@slug:registry=https://my-registry.tld/
//my-registry.tld/:_authToken="<token>"
```

and a project config like this:

```
@slug:registry=https://my-registry.tld/

https://my-registry.tld/:always-auth=true

auto-install-peers=true
engine-strict=true
prefer-frozen-lockfile=true
resolution-mode=highest
save-exact=true
save-prefix=''
strict-peer-dependencies=true
update-notifier=true
```

It should install the dependencies correctly 

### What do you see instead?

```
error: GET https://my-registry.tld/@slug%2fpackage-1 - 401

error: GET https://my-registry.tld/@slug%2fpackage-2 - 401
```

### Additional information

https://github.com/oven-sh/bun/pull/13990

robertshuford:
### What does this PR do?

Fixes #14433

Auth configuration was only applied per `.npmrc` file and then discarded.  Therefore, auth defined in the global `.npmrc` wouldn't apply to registries in the project `.npmrc`.

So with a global like this:
```
@slug:registry=https://my-registry.tld/
//my-registry.tld/:_authToken="<token>"
```
And a project like this:
```
@slug:registry=https://my-registry.tld/
```
When put into the `NpmRegistryMap`, the scoped `NpmRegistry` `@slug` at the project level would replace the global `@slug` scoped `NpmRegistry`, and the auth configuration would not be reapplied.

The fix I added was to collect the `.npmrc` configuration items so `NpmRegistry` objects can be rebuilt and discard the configuration items only after all `.npmrc` files are processed.

- [x] Code changes

Added a test and tested locally.

If Zig files changed:


Jarred-Sumner:
Thank you

PR #7166, opened 2023-11-17
ensure `.bin` folder is created before linking binaries

dylan-conway:
### What does this PR do?
fixes #7161 

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

### How did you verify your code works?
added and modified tests

PR #15540, opened 2024-12-02
package.json: put :local builds into their own folder

nektro:
they have their own cmake settings and overlapping the folders causes weird issues


robobun:
@nektro, your commit ba67940664fede0e2902a50d12017d100c3054c7 is building: [`#7166`](https://buildkite.com/bun/bun/builds/7166)

PR #15542, opened 2024-12-02
docs: add .env.test to guides/runtime/set-env for consistency with do…

hex2f:
### What does this PR do?

Fixed a small inconsistency that tripped me up where [guides/runtime/set-env](https://bun.sh/guides/runtime/set-env) didn't mention the existence of `.env.test`.

- [ ] Code changes


Jarred-Sumner:
thanks

Issue #7179, opened 2023-11-17
linux baseline build has illegal instruction on `bun --version`


GHNewbiee:
### What platform is your computer?

Linux 5.15.0-88-generic x86_64 x86_64

### What steps can reproduce the bug?

```bash
curl -fsSL https://bun.sh/install | bash
```

### What is the expected behavior?

Bun to get installed.

### What do you see instead?

```bash
bun was installed successfully to ~/.bun/bin/bun 
bash: line 159: 27737 Illegal instruction     (core dumped) IS_BUN_AUTO_UPDATE=true $exe completions &> /dev/null
Run 'bun --help' to get started
```

### Additional information

If I run
```bash
bun [--help]

# Then I get

Illegal instruction (core dumped)
```

paperclover:
what exact CPU are you using?

GHNewbiee:
@paperdave Tfyh.

```
$ cat /proc/cpuinfo
processor	: 0
vendor_id	: AuthenticAMD
cpu family	: 18
model		: 1
model name	: AMD A6-3410MX APU with Radeon(tm) HD Graphics
stepping	: 0
microcode	: 0x3000027
cpu MHz		: 800.000
cache size	: 1024 KB
physical id	: 0
siblings	: 4
core id		: 0
cpu cores	: 4
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 6
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm 3dnowext 3dnow constant_tsc rep_good nopl nonstop_tsc cpuid extd_apicid aperfmperf pni monitor cx16 popcnt lahf_lm cmp_legacy svm extapic cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw ibs skinit wdt cpb hw_pstate vmmcall arat npt lbrv svm_lock nrip_save pausefilter
bugs		: fxsave_leak sysret_ss_attrs null_seg spectre_v1 spectre_v2
bogomips	: 3194.05
TLB size	: 1536 4K pages
clflush size	: 64
cache_alignment	: 64
address sizes	: 40 bits physical, 48 bits virtual
power management: ts ttp tm stc 100mhzsteps hwpstate cpb

processor	: 1
vendor_id	: AuthenticAMD
[...]

PR #15544, opened 2024-12-03
bake: fix the big regressions

paperclover:
this also does some work to make multi-threaded SSR, but does not actually start the threads to do so.


robobun:
@paperdave, your commit ebaefcf4e0327a9575c1a0947ac7d787887493b8 is building: [`#7179`](https://buildkite.com/bun/bun/builds/7179)

Issue #7209, opened 2023-11-18
SIGSEGV 11 when terminating a Worker


SukkaW:
### What version of Bun is running?

1.0.13+f5bf67bd1

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

- Upgrade bun to the latest `canary` by running `bun upgrade --canary`
  - This is to confirm the bun hasn't fixed this issue
- `git clone https://github.com/sukka-reproductions/bun-crash-repro --depth=1 && cd bun-crash-repro && bun i`
- `bun run build`

### What do you see instead?

```
error: script "build" exited with code 11 (SIGSEGV)
```

### Additional information

I can't obtain stack traces, see https://github.com/oven-sh/bun/issues/7184

SukkaW:
Updated to the latest bun canary (`1.0.13+d8ca6ca69`), the SIGSEV 11 persists.

SukkaW:
@Jarred-Sumner So I've built a `bun-debug` binary (even created a PR #7348 during the process 😆). How can I obtain the stack trace of crashing scripts?

Hanaasagi:
I can reproduce a `SIGSEGV` using release version 1.0.14, but strangely, the issue cannot be reproduced in the debug version under the 1.0.14 tag 🙀. My platform is Linux.

About debugging a coredump

1. `lldb ~/zig/bun/build/bun-debug`
2. `r run index.ts` (P.S. `bun run build` will spawn a
 new process, we need attach the subprocess. So just `run index.ts`, it would be easiser)
3. If it's not a `SIGSEGV` in the main thread, you'll need to attach to the subthread. Then, wait for the crash and execute `bt` to get the call stack.

In linux, just `coredumpctl debug --debugger=lldb`

SukkaW:
> but strangely, the issue cannot be reproduced in the debug version under the 1.0.14 tag 🙀.

Same here. I can't reproduce this with `bun-debug` built from the `main` branch. But the canary version of `bun` (which was also built from the latest commit) still has the segmentation fault.

Hanaasagi:
I found that it might be related to the size of the data. When I modify the return value of the `foo` to be an array with 512 elements, no coredump occurs. However, if I increase it to over 600 elements, a coredump occurs.

https://github.com/sukka-reproductions/bun-crash-repro/blob/main/lib.ts

```JavaScript
export const foo = async () => {
  // no coredump
[...]

PR #15555, opened 2024-12-03
Remove a duplicate if statement

heimskr:
### What does this PR do?

Removes a duplicate `if (cond) { return true; }` statement. Chances are the compiler removes it as dead code anyway.

- [x] Code changes

I didn't, but this is a trivial change that shouldn't have any effect other than clarifying the source code.


robobun:
@heimskr, your commit efca5062f71aeddd4ad9c1634de45b61d0bb20d2 is building: [`#7209`](https://buildkite.com/bun/bun/builds/7209)

PR #15328, opened 2024-11-22
bump webkit

dylan-conway:

Issue #14982, opened 2024-11-04
Process freezed after upgrade from 1.1.33 to 1.1.34


esciullo:
### What version of Bun is running?

1.1.34+5e5e7c60f

### What platform is your computer?

Linux 5.15.153.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

Use the following `index.ts` and `test-test.ts` files and run `bun index.ts test`

index.ts
```typescript
import { program } from "commander";

program.name("test").command("test", "Test command").parse();
```

test-test.ts
```typescript
import { Command } from "commander";

new Command("test")
  .action(() => {
    console.log("Test command");
    process.exit(0);
  })
  .parse();
```

### What is the expected behavior?

The process exit successfully.

### What do you see instead?

The process remains freezed.

### Additional information

The same code works fine with bun 1.1.33

Jarred-Sumner:
Reproduces on Linux. Does not reproduce on macOS.

Jarred-Sumner:
It is a tick without idle infinite loop.

Leaving a systrace open for more than a minute produces 350 MB of sched_yield() logs.

Jarred-Sumner:
main thread:
```cpp
___lldb_unnamed_symbol3635 (@___lldb_unnamed_symbol3635:56)
pthread_cond_wait (@pthread_cond_wait:150)
WTF::ThreadCondition::wait(WTF::Mutex&) (@WTF::ThreadCondition::wait(WTF::Mutex&):15)
WTF::ThreadCondition::timedWait(WTF::Mutex&, WTF::WallTime) (@WTF::ThreadCondition::timedWait(WTF::Mutex&, WTF::WallTime):36)
WTF::ParkingLot::parkConditionallyImpl(void const*, WTF::ScopedLambda<bool ()> const&, WTF::ScopedLambda<void ()> const&, WTF::TimeWithDynamicClockType const&) (@WTF::ParkingLot::parkConditionallyImpl(void const*, WTF::ScopedLambda<bool ()> const&, WTF::ScopedLambda<void ()> const&, WTF::TimeWithDynamicClockType const&):95)
[...]

PR #14885, opened 2024-10-29
Reduce memory usage in long-running processes

Jarred-Sumner:
### What does this PR do?

New:
```sh
┌─────────────────────────┬──────────┐
│                         │ Values   │
├─────────────────────────┼──────────┤
│   Memory usage MB (RSS) │ 650      │
│ AbortSignal alive count │ 200002   │
│ new AbortSignal() calls │ 55500000 │
│        Uptime (seconds) │ 630      │
└─────────────────────────┴──────────┘
```

Previously:
```
┌─────────────────────────┬──────────┐
│                         │ Values   │
├─────────────────────────┼──────────┤
│   Memory usage MB (RSS) │ 678      │
│ AbortSignal alive count │ 495616   │
│ new AbortSignal() calls │ 55500000 │
│        Uptime (seconds) │ 630      │
└─────────────────────────┴──────────┘
```

For the following script:
```ts
console.clear();
console.log("Starting...");
const iterationCount = 100000;
globalThis.signals = new Array(iterationCount);
globalThis.requests = new Array(iterationCount);
var total = 0;
setInterval(() => {
  const rss = (process.memoryUsage.rss() / 1024 / 1024) | 0;
  const objectTypeCounts = require("bun:jsc").heapStats().objectTypeCounts;
  console.table({
    "Memory usage MB (RSS)": rss,
    "AbortSignal alive count": objectTypeCounts.AbortSignal || 0,
    "new AbortSignal() calls": total,
    "Uptime (seconds)": Math.floor(process.uptime()),
  });
}, 10000);
while (true) {
  for (let i = 0; i < iterationCount; i++) {
    const req = new Request("https://example.com/");
    requests[i] = req;
    signals[i] = AbortSignal.any([req.signal, new AbortController().signal]);
  }
  total += iterationCount;
  await Bun.sleep(1000);
}
```

### How did you verify your code works?

CI

PR #15557, opened 2024-12-03
Allow WTF timers to participate in the event loop

190n:
### What does this PR do?

WebKit has `WTF::RunLoop` and `WTF::RunLoop::Timer`. The latter is created in order to run GC at certain intervals. Since Bun's event loop is really in charge of timers, rather than the RunLoop, the GC timers would previously never run. This PR enables Bun to include those timers in the event loop and fire them at the right times.

This also includes the change from #14885 to release heap access while waiting for events.

TODO:

- [x] fix todos in the code
- [ ] performance testing
- [x] make the test suite reasonable (probably just check #14982 does not regress)
- [ ] see if we can get rid of `GarbageCollectionController` from `event_loop.zig`

Requires oven-sh/WebKit#71 and #15328

CI


robobun:
@190n, your commit 62a26a6f85dee12d1629d88f0bfed5280c35f80f is building: [`#7217`](https://buildkite.com/bun/bun/builds/7217)

PR #15558, opened 2024-12-03
Fix #15556 - Buffer.write__ broken without explicit offset

### LINKED ISSUE #15556: Buffer.write__ slow/incorrect without explicit offset

1.1.38+bf2f153f5

Darwin 23.5.0 arm64 arm

All of Bun's buffer write commands (e.g. `writeInt8`, `writeInt16LE`) experience significant performance degradation when only one value argument, as opposed to a value and an explicit offset argument, is provided. Additionally, not providing an offset will cause the function to return `NaN` instead of the length of the value written (i.e. the new offset when initial offset is 0).

Notably, this does not occur when using Node - performance stays the same, and the same value is returned regardless of whether a second argument is provided.

This performance deterioration can be reproduced by modifying the existing benchmark at `bench/snippets/buffer.js`, by removing the second argument from the write commands:
```javascript
buf.writeInt8(0, 0) // -> buf.writeInt8(0)
```

The NaN issue can be reproduced via the snippet:
```javascript
var buf = Buffer.alloc(4)
console.log(buf.writeInt32LE(0, 0)) // -> 4 in Bun and Node
console.log(buf.writeInt32LE(0)) // -> NaN in Bun, 4 in Node
```

The performance with and without the aforementioned edit to `buffer.js` is comparable, and the output of the two calls in the other snippet are equivalent.

### What do you see instead?

With no explicit offset provided, the performance of Bun deteriorates ~5x, and instead of returning the correct new offset, `NaN` is returned.

### Additional information

_No response_

Serdnad:
### What does this PR do?

This PR brings the behavior of Buffers `.write__` functions (e.g. `writeInt32LE`) to better parity with the behavior in Node, by having the `offset` argument default to 0. Currently, calling these functions without supplying an offset argument will write to the buffer at offset 0, but they will return a new offset of `NaN` and suffer a major performance degradation.

- [X] Code changes

Comparing the performance of `bench/snippets/buffer.js` before and after this change with the slight modification specified in #15556 shows that performance is improved, and becomes essentially equivalent, as might be expected. Additionally, the code snippet also in that issue was used to verify the return value is now as expected.

I have not written additional tests for these changes - I'm more than happy to, but it was not quite clear to me exactly where they would best go, so I thought to reach out and ask first. Maybe in a new `test/js/bun/buffer`?

### DIFF
--- a/src/js/builtins/JSBufferPrototype.ts
+++ b/src/js/builtins/JSBufferPrototype.ts
@@ -174,43 +174,43 @@ export function readBigUInt64BE(this: BufferExt, offset) {
   return (this.$dataView ||= new DataView(this.buffer, this.byteOffset, this.byteLength)).getBigUint64(offset, false);
 }
 
-export function writeInt8(this: BufferExt, value, offset) {
+export function writeInt8(this: BufferExt, value, offset = 0) {
   (this.$dataView ||= new DataView(this.buffer, this.byteOffset, this.byteLength)).setInt8(offset, value);
   return offset + 1;
 }
-export function writeUInt8(this: BufferExt, value, offset) {
+export function writeUInt8(this: BufferExt, value, offset = 0) {
   (this.$dataView ||= new DataView(this.buffer, this.byteOffset, this.byteLength)).setUint8(offset, value);
   return offset + 1;
 }
-export function writeInt16LE(this: BufferExt, value, offset) {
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thanks for this. There is no tests, and this code doesn't work because default arguments are not supported in the builtins generator.

PR #7242, opened 2023-11-20
Enhancement: Exposing additional column details for SQLite.

bhardwajakshit:
### What does this PR do?

- This commit improves the SQLite integration by adding support for retrieving additional column details such as column name, origin name, table name, database name, and type.

- **Issue**: #7134 

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

PR #15565, opened 2024-12-04
ci: run node tests directly instead of translated files

nektro:
- deleted vendored files from node-jester
- add passing node files
- fixes two bugs
- update ci runner


robobun:
@nektro, your commit 5ce5576b5d54585124ff58b4dcd6435412a8a794 is building: [`#7242`](https://buildkite.com/bun/bun/builds/7242)

nektro:
spoke irl

Issue #7264, opened 2023-11-22
Segmentation fault on bun --inspect when the module variable is accessed


baptistepillot:
### What version of Bun is running?

1.0.11+f7f6233ea

### What platform is your computer?

Linux 6.2.0-36-generic x86_64 x86_64 (Linux Mint 21.2)

### What steps can reproduce the bug?

- Create a script named **test.ts** and containing :
```ts
console.log(module)
Bun.serve({
	fetch(request: Request): Promise<Response>|Response {
		return new Response(request.url)
	}
})
```

- Inspect it :
`bun --inspect test.ts`

- Open the "Inspect in browser" link into your favorite browser (tested using Chrome and Firefox)

- Wait 1 second

### What is the expected behavior?

In browser: Should list source files
In console: Should not crash

### What do you see instead?

In browser: The source file list stays empty
In console: Segmentation fault

### Additional information

The problem happens only when there is an access to the module global variable.
If not: no crash.

baptistepillot:
Probably linked to #6082

baptistepillot:
Aaarg, sorry : updated to bun version/revision `1.0.13+f5bf67bd1` and the crash does not occur anymore.
I guess it is ok now.

PR #15571, opened 2024-12-04
ci: fix v8.test.ts on musl

nektro:
this should fix `test/v8/v8.test.ts` for this target


robobun:
@nektro, your commit deab3a6029aed64aa998d1a2975ba4002af27561 is building: [`#7264`](https://buildkite.com/bun/bun/builds/7264)

nektro:
`TypeError: Error relocating /tmp/buntmp-iphHmB/bun.test.OODLhN/build/Release/v8tests.node: _ZN2v812api_internal17FromJustIsNothingEv: symbol not found`

> `v8::api_internal::FromJustIsNothing()`

nektro:
`TypeError: Error relocating /tmp/buntmp-pnCekb/bun.test.jNOmGm/build/Release/v8tests.node: _ZNK2v85Array6LengthEv: symbol not found`

> `v8::Array::Length() const`

nektro:
appears that musl doesn't support lazy symbols when using `dlopen`. in that case we can make weak stubs ourselves.
the list of symbols can be gotten from

```sh
$ docker run -it node:22.6.0 sh
# nm `which node` | grep ' T ' | grep _ZNK | grep v8
```

nektro:
notes
https://nodejs.org/api/n-api.html
https://github.com/nodejs/node/blob/v22.6.0/src/node.h
https://github.com/nodejs/node/blob/v22.6.0/src/node_buffer.h
https://github.com/nodejs/node/blob/v22.6.0/src/node_object_wrap.h
https://github.com/nodejs/node/blob/v22.6.0/src/node_version.h

Issue #15022, opened 2024-11-06
Fetch body doesn't work properly with fs.ReadStream


benardayim:
### What version of Bun is running?

1.1.34+5e5e7c60f

### What platform is your computer?

Microsoft Windows NT 10.0.19044.0 x64

### What steps can reproduce the bug?

Create a request with fetch and use fs.ReadStream (from fs.createReadStream()) as body.

Save this code to a file then run it with both bun and node.

```javascript
import http from "node:http";
import fs from "node:fs";

const TEST_FILE_PATH = "test.txt";

let server = http.createServer((req, res) => {
    res.writeHead(200);

    let data = "";
    req.on("data", e => data += e);
    req.on("end", () => {
        res.end("body is " + data.length + " bytes.");
    })
});

server.listen(3131, () => console.log("Listening"));

let response = await fetch("http://localhost:3131/", {
    method: "POST",
    body: fs.createReadStream(TEST_FILE_PATH)
});
let body = await response.text();

console.log(body);
```

### What is the expected behavior?

It should use the data from the stream as body but it doesn't.

### What do you see instead?

It sends 0 bytes as body.

guest271314:
See [Streaming requests with the fetch API ](https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests) and how the current implementation of Bun's `fetch()` does not handle HTTP/2 full-duplex streaming https://github.com/oven-sh/bun/issues/7206.

Issue #11938, opened 2024-06-18
node:http module doesn't send chunked response


ayonli:
### What version of Bun is running?

1.1.10

### What platform is your computer?

Linux 5.15.0-107-generic x86_64 x86_64

### What steps can reproduce the bug?

Start the server with the following code:

```ts
import * as http from "node:http";

const server = http.createServer((req, res) => {
    res.setHeader("Content-Type", "text/event-stream");
    res.setHeader("Cache-Control", "no-cache");
    res.setHeader("Connection", "keep-alive");
    res.setHeader("Transfer-Encoding", "chunked");
    res.setHeader("Access-Control-Allow-Origin", "*");

    res.writeHead(200, "OK");
    res.write("");
});

server.listen(4000); 
```

Then use `EventSource` or `fetch` to connect to the server, for example:

```ts
const es = new EventSource("http://localhost:4000");

// or
const res = await fetch("http://localhost:4000");
```

### What is the expected behavior?

The server should send a `200 OK` response.

### What do you see instead?

The server doesn't send any response, the connection got stuck forever.

### Additional information

This code works perfectly in Node.js and the client can receive a `200 OK` response.

Issue #11621, opened 2024-06-05
Requests with Transfer-Encoding: chunked seems to be broken in Bun but not in Node


shadone:
### What version of Bun is running?

1.1.12

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

1. create `index.js`
```js
const express = require('express')
const bodyParser = require('body-parser')
const multer = require('multer')

const app = express()
app.use(bodyParser.json())

const upload = multer({ storage: multer.memoryStorage() })

function handleRegistrations(req, res, next) {
	const product_id = req.body.product_id
	console.log(`### handleRegistrations: product_id=${product_id}`)
	res.status(200).json({ product_id })
}

app.post(
  "/registrations",
  upload.single("image"),
  handleRegistrations,
)

app.post("*", (req, res, next) => {
	console.log("Unhandled endpoint", req)
	res.status(404).send()
})

console.log("Listening on port 8080")
app.listen(8080)
```

2. run the server with `bun ./index.js`

3. make the request with `Transfer-Encoding: chunked` like this: `curl -XPOST http://localhost:8080/registrations -F product_id=1234foobar567 -H 'Transfer-Encoding: chunked'`

The server spits the following error:
```
Error: Unexpected end of form
    at _final (/Users/denis/tmp/node-multipart-form-data/node_modules/busboy/lib/types/multipart.js:588:13)
    at callFinal (node:stream:2778:23)
    at prefinish (node:stream:2801:23)
    at finishMaybe (node:stream:2808:25)
    at <anonymous> (node:stream:2741:76)
    at onend (node:stream:2088:21)
    at emit (native)
    at endReadableNT (node:stream:2396:27)
    at processTicksAndRejections (:12:39)
```

### What is the expected behavior?

Run the same server with node instead of bun, no errors occur.

shadone:
I believe this could be the same or related issue as the following:
https://github.com/oven-sh/bun/issues/4139
https://github.com/oven-sh/bun/issues/9100
https://github.com/oven-sh/bun/issues/4234
https://github.com/oven-sh/bun/issues/8638

guest271314:
What happens when `Bun.listen()` is used to handle `Transfer-Encoding: chunked`?

shadone:
> What happens when `Bun.listen()` is used to handle `Transfer-Encoding: chunked`?

[...]

Issue #8638, opened 2024-02-02
Express post request with transfer-encoding chuncked has no body


co-sic:
### What version of Bun is running?

1.0.25, 1.1.8

### What platform is your computer?

Linux 6.5.0-15-generic x86_64 x86_64

### What steps can reproduce the bug?

I'm running express and apollo-server (latest versions) and i am having the problem that the body of an incomming post request is empty in certain situations. I narrowed it down to the ` "transfer-encoding": "chuncked"` header beeing present and/or the 'content-lenght' header beeing not present. If tried the same failing request and send it through a proxy first (ngrok) and then it worked. You can see the diff of the requests here https://www.diffchecker.com/0AdUpa8R/ (Left side is failing, right side is working).

Electroid:
Thanks for reporting, mind sharing some code to reproduce this issue?

co-sic:
Hey @Electroid i added a example here to reproduce this: https://github.com/co-sic/bun-playground/tree/cosmo-bun-apollo-server

ambergristle:
Hi! I've potentially run into a similar issue.

I use Docker to build a Next.js (`14.0.3`) app running on Bun (`oven/bun:1.0-slim`). I use App Router to `GET` some data from a third party (Marketstack), and return it to the client.

The third party service responds with chunk-encoded data. When deployed (on Railway, via Docker) _large (e.g., ~400 rows)_ responses fail on json parse (`SyntaxError: Unexpected end of JSON`). _Small (~30 rows)_ responses work ok both locally and in deployment. Locally (via Docker), the app mostly has no issue; I've been able to repro a couple times, but I'm not sure how. Deploying with Node instead of Bun seems to work regardless of response size.

I added a log of the response before json parse, and those (larger) responses that fail log a size of 0kb, while responses that are (ultimately) parsed ok have a non-zero size.

Source code from the [jshttp](https://github.com/jshttp) lib seems to make its way into the error logs (if left uncaught).

I'm wondering whether this has anything to do with:
- https://github.com/oven-sh/bun/issues/4139
- https://github.com/oven-sh/bun/pull/4127

Here's a permalink to the [failing point in my repo](https://github.com/ambergristle/yolofail/blob/11cafcc5b6db352b9075a4b00a774773b4d4eb1e/app/src/lib/marketstack.ts#L34). I'm currently working on a more contained/easily-reproducible example.

ambergristle:
[...]

PR #15579, opened 2024-12-04
fix(node:http) fix node:http chunked encoding on server and add chunked encoding support on the client

cirospaciari:
### What does this PR do?
This depends on or replaces: https://github.com/oven-sh/bun/pull/15576

Fix: https://github.com/oven-sh/bun/issues/11621
Fix: https://github.com/oven-sh/bun/issues/8638
Fix: https://github.com/oven-sh/bun/issues/11938
Fix: https://github.com/oven-sh/bun/issues/7890
And of course the same ones as #15576:

Fix: https://github.com/oven-sh/bun/issues/6017
Fix: https://github.com/oven-sh/bun/issues/7206
Fix: https://github.com/oven-sh/bun/issues/15022

Improve on: https://github.com/oven-sh/bun/issues/9100

~~Probably https://github.com/oven-sh/bun/pull/13778 should be merged after this if tests pass.~~

- [x] Code changes

Tests

### DIFF
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -743,7 +743,7 @@ pub const DeferredTaskQueue = struct {
             };
 
             if (!this.map.values()[i](key)) {
-                this.map.swapRemoveAt(i);
+                _ = this.map.swapRemove(key);
                 last = this.map.count();
             } else {
                 i += 1;
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -743,7 +743,7 @@ pub const DeferredTaskQueue = struct {
             };
 
             if (!this.map.values()[i](key)) {
-                _ = this.map.swapRemove(key);
+                this.map.swapRemoveAt(i);
                 last = this.map.count();
             } else {
                 i += 1;
--- a/src/bun.js/webcore/body.zig
+++ b/src/bun.js/webcore/body.zig
@@ -158,6 +158,18 @@ pub const Body = struct {
 
             return false;
         }
+
+        pub fn isDisturbed2(this: *const PendingValue, globalObject: *JSC.JSGlobalObject) bool {
+            if (this.promise != null) {
+                return true;
+            }
+
+            if (this.readable.get()) |readable| {
+                return readable.isDisturbed(globalObject);
+            }
+
+            return false;
+        }
         pub fn isStreamingOrBuffering(this: *PendingValue) bool {
             return this.readable.held.has() or (this.promise != null and !this.promise.?.isEmptyOrUndefinedOrNull());
         }
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -2612,6 +2612,8 @@ pub const FetchTaskletChunkedRequestSink = struct {
     done: bool = false,
     auto_flusher: AutoFlusher = AutoFlusher{},
 
+    pub usingnamespace bun.New(FetchTaskletChunkedRequestSink);
+
     fn unregisterAutoFlusher(this: *@This()) void {
         if (this.auto_flusher.registered)
             AutoFlusher.unregisterDeferredMicrotaskWithTypeUnchecked(@This(), this, this.globalThis.bunVM());
@@ -2654,6 +2656,9 @@ pub const FetchTaskletChunkedRequestSink = struct {
     }
     pub fn sink(this: *@This()) Sink {
         return Sink.init(this);
+    }
+    pub fn toSink(this: *@This()) *@This().JSSink {
+        return @ptrCast(this);
     }
     pub fn finalize(this: *@This()) void {
         var buffer = this.buffer;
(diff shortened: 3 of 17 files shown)


robobun:
@cirospaciari, your commit e3fa8aea7f18953a00c369d7448cd64f170b3065 is building: [`#7276`](https://buildkite.com/bun/bun/builds/7276)

test/js/web/fetch/fetch-leak-test-fixture-5.js
```diff
@@ -66,6 +66,19 @@ function getBody() {
...
       body = getURLSearchParams();
       break;
+    case "iterator":
+      body = async function* iter() {
+        yield (cachedBody ??= getString());
+      };
+      break;
+    case "stream":
+      body = new ReadableStream({
+        start(c) {
```
Jarred-Sumner: This should use pull instead of start and have a slight delay in pull. We may detect if the ReadableStream has already ended with a single chunk and do something else

test/js/node/http/node-http.test.ts
```diff
@@ -2435,3 +2435,118 @@ it("should work when sending https.request with agent:false", async () => {
   await promise;
 });
 
+it("client should use chunked encoded if more than one write is called", async () => {
+  using server = Bun.serve({
```
Jarred-Sumner, line 2443: Can you write this test in a way that runs in Node too? This file has had many tests which did not actually run in Node

src/js/node/http.ts
```diff
@@ -838,12 +841,15 @@ IncomingMessage.prototype = {
...
     }
 
-    const contentLength = this.headers["content-length"];
-    const length = contentLength ? parseInt(contentLength, 10) : 0;
-    if (length === 0) {
-      this[noBodySymbol] = true;
-      callback();
-      return;
+    const encoding = this.headers["transfer-encoding"];
+    if (encoding?.indexOf("chunked") === -1) {
```
Jarred-Sumner:
```suggestion
    if (encoding?.indexOf?.("chunked") === -1) {
```

src/js/node/http.ts
```diff
@@ -777,6 +777,9 @@ function requestHasNoBody(method, req) {
   if ("GET" === method || "HEAD" === method || "TRACE" === method || "CONNECT" === method || "OPTIONS" === method)
     return true;
   const headers = req?.headers;
+  const encoding = headers?.["transfer-encoding"];
+  if (encoding?.indexOf("chunked") !== -1) return false;
```
Jarred-Sumner:
```suggestion
  if (encoding?.indexOf?.("chunked") !== -1) return false;
```

test/js/node/http/node-http.test.ts
cirospaciari: You mean using node:http server instead of Bun?

cirospaciari: done

Jarred-Sumner:
Are there node tests we can copy?

src/bun.js/webcore/body.zig
```diff
@@ -158,6 +158,18 @@ pub const Body = struct {
 
             return false;
         }
+
+        pub fn isDisturbed2(this: *const PendingValue, globalObject: *JSC.JSGlobalObject) bool {
```
Jarred-Sumner, line 162: why can't we use isDisturbed?

src/bun.js/event_loop.zig
```diff
@@ -743,7 +743,7 @@ pub const DeferredTaskQueue = struct {
             };
 
             if (!this.map.values()[i](key)) {
-                this.map.swapRemoveAt(i);
+                _ = this.map.swapRemove(key);
```
Jarred-Sumner: Was this change intentional?

src/bun.js/webcore/streams.zig
```diff
@@ -2600,7 +2602,266 @@ pub fn HTTPServerWritable(comptime ssl: bool) type {
...
+    }
+
+    pub fn flush(this: *@This()) JSC.Maybe(void) {
+        _ = this.internalFlush() catch 0;
+        return .{ .result = {} };
+    }
+    pub fn flushFromJS(this: *@This(), globalThis: *JSGlobalObject, _: bool) JSC.Maybe(JSValue) {
+        return .{ .result = JSC.JSPromise.resolvedPromiseValue(globalThis, JSValue.jsNumber(this.internalFlush() catch 0)) };
+    }
+    pub fn destroy(this: *@This()) void {
```
Jarred-Sumner: use `pub usingnamespace bun.New`

We want that memory leak tracking

src/bun.js/event_loop.zig
cirospaciari: Yes, was crashing without it

src/bun.js/webcore/body.zig
cirospaciari: Because can or not be an Request instance or an unknow instance, we only care about the readable stream. (We still use isDisturbed when we know that is a Request instance for example)

src/bun.js/event_loop.zig
cirospaciari: fixed

Issue #13070, opened 2024-08-04
File not found when running `bun --config bunfig.toml` 


joas8211:
### What version of Bun is running?

1.1.21+70ca2b76c

### What platform is your computer?

Linux 6.6.40-2-lts x86_64 unknown

### What steps can reproduce the bug?

1. Open an project folder with an empty bunfig.toml file.
2. Run `bun --config bunfig.toml`

### What is the expected behavior?

I'm expected to see list of possible commands outputted like when `--config` is not specified. Like so:

```
Bun is a fast JavaScript runtime, package manager, bundler, and test runner. (1.1.21+70ca2b76c)

Usage: bun <command> [...flags] [...args]

Commands:
  run       ./my-script.ts       Execute a file with Bun
            lint                 Run a package.json script
  test                           Run unit tests with Bun
  x         eslint               Execute a package binary (CLI), installing if needed (bunx)
  repl                           Start a REPL session with Bun
  exec                           Run a shell script directly with Bun

  install                        Install dependencies for a package.json (bun i)
  add       tailwindcss          Add a dependency to package.json (bun a)
  remove    babel-core           Remove a dependency from package.json (bun rm)
  update    elysia               Update outdated dependencies
  link      [<package>]          Register or link a local npm package
  unlink                         Unregister a local npm package
  patch <pkg>                    Prepare a package for patching
  pm <subcommand>                Additional package management utilities

  build     ./a.ts ./b.jsx       Bundle TypeScript & JavaScript into a single file

  init                           Start an empty Bun project from a blank template
  create    next-app             Create a new project from a template (bun c)
  upgrade                        Upgrade to latest version of Bun.
  <command> --help               Print help text for command.

Learn more about Bun:            https://bun.sh/docs
Join our Discord community:      https://bun.sh/discord
```

### What do you see instead?

```
error: File not found: "bunfig.toml"
```

### Additional information

I suspect this is the reason why project configuration options from bunfig.toml don't apply also when `--config` is not specified, like in these issues: #12990, #11445

Issue #12753 suggests also that some validation is happening before not found error is outputted.

Following issue could also be related to this: #12987.

vktrl:
You need to use this syntax: `bun --config="bunfig.toml"`, otherwise the last arg is interpreted as the entrypoint.

osdiab:
[...]

Issue #12753, opened 2024-07-23
Config file can not be found when "run.bun = false" in bunfig.toml


ptrxyz:
### What version of Bun is running?

1.1.20

### What platform is your computer?

Linux 6.9.10-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

- Create an empty directory, but in a `bunfig.toml` with this content:
```toml
[run]
bun = false
```

- in your shell type `bun -c $(pwd)/bunfig.toml run which node`
- See this:
```
error: File not found: "bunfig.toml"
```
- Change `bun = true` in your `bunfig.toml` and run the same command again. See nothing happen (no output, no error).
- Change `bun = whatever` in your `bunfig.toml` and run the same command again. See an error that a `boolean is expeted`

### What is the expected behavior?

I would expect this to run bun as if i specified `bun --bun`. No error messages at all except for the on in the last case of course.

### What do you see instead?

Some error messages, see above.

### Additional information

I have no other bunfigs, nothing like `$HOME/.bunfig.toml` or `$XDG_CONFIG_DIR/bunfig.toml`.

PR #7361, opened 2023-11-29
Add ecosystem tests

Electroid:
### What does this PR do?

We will now be running the test suites of popular npm packages to make sure that: 1) said packages work, and 2) we do not introduce regressions that break said packages. 

Not all the tests are running, some are skipped due to failures, which we will fix.

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

Added tests.

PR #15596, opened 2024-12-05
read `bunfig.toml` automatically for more commands

### LINKED ISSUE #11445: bunfig's [run] section not working

1.1.10

Darwin 23.1.0 arm64 arm

`bunfig.toml`:
```toml
[run]
silent = true
```
Run command
```
$ bun run dev
```

`bun run dev` should not print the command being run, as explained in https://bun.sh/docs/runtime/bunfig#bun-run

I should only see the info logged/emitted by the dev script itself — same as if I do `bun run --silent dev`

### What do you see instead?

```
$ next dev
  ...dev server output...
```

### Additional information

Using the [bun setting](https://bun.sh/docs/runtime/bunfig#run-bun-auto-alias-node-to-bun) doesn't work either.

```toml
[run]
bun = true
```

(This is the setting I'm most interested in.)

### LINKED ISSUE #15483: `-c`/`--config` should be a default option

1.1.37+8ca0eb831

Debian GNU/Linux 12 (bookworm)

Bug: `bunfig.toml` does not load by default.
Steps to reproduce:
1. Create `bunfig.toml` with these contents:
```toml
[run]
bun = true
```
2. Run `bun run which node`

Running `bun run which node` should output something similar to this:
```
/tmp/bun-node-8ca0eb831/node
```
According to the [documentation for bunfig.toml](https://bun.sh/docs/runtime/bunfig), placing the `bunfig.toml` file in the same directory as the `package.json` should load the config file by default.

### What do you see instead?

Running `bun run which node` outputs this:
```
/home/user/.nvm/versions/node/v22.11.0/bin/node
```

### Additional information

This is a potential fix to solve this issue (implementing this will also close #11445):
Make the argument for `-c` or `--config` to be included by default.
Running `bun run -c which node` outputs the expected alias:
```
/tmp/bun-node-8ca0eb831/node
```

### LINKED ISSUE #15484: `-c`/`--config` should be a default option for `bun run`

When executing `bun run` in the shell, configuration specified in `bunfig.toml` does not load automatically.
The [bunfig.toml documentation](https://bun.sh/docs/runtime/bunfig) states that placing the `bunfig.toml` file in the same directory as the `package.json` should load the configuration automatically.

Steps to reproduce:
1. Create `bunfig.toml` in the same directory as your `package.json` with these contents:
```toml
[run]
bun = true
```
2. Run `bun run which node`. The shell returns this (or wherever the executable for `node` is located on your system):
```bash
/home/user/.nvm/versions/node/v22.11.0/bin/node
```

Instead, this command is supposed to return something like this:
```bash
/tmp/bun-node-8ca0eb831/node
```

Make `-c`/`--config` a default argument for `bun run`.

Running `bun run -c which node` loads the configuration correctly, and outputs the expected alias alias to bun:
```
/tmp/bun-node-8ca0eb831/node
```
Implementing this solution will close #11445, #12753, and #12538.

Adding `-c`/`--config` as an argument to every `bun run` command.

RiskyMH:
allows `bun run` to load `bunfig.toml` and allows for `bunx` to load root config too.

fixes #15484, fixes #15483, fixes #11445, 

(mainly so you can make bun `--bun` more often)

---

~~the following are fixed because i removed `--config` with no value being allowed (now you can have to pass it a path), as it now auto tries local like install which is more expected (so point of empty `-c` is useless). **lots of people may expect this behavior, so I can revert it**:~~

~~fixes *not anymore* #13070, fixes *not anymore* #6300, fixes *not anymore* #12753~~

undid above change as created way more breaking change, will consider better work around in future pr


robobun:
@RiskyMH, your commit 7bace058440f766d96960463e2dafbdfee48a57c is building: [`#7361`](https://buildkite.com/bun/bun/builds/7361)

src/cli.zig
```diff
@@ -1054,17 +1054,15 @@ pub const Arguments = struct {
                 ctx.debug.silent = true;
             }
 
-            if (opts.define) |define| {
-                if (define.keys.len > 0)
-                    bun.JSC.RuntimeTranspilerCache.is_disabled = true;
-            }
-        }
-
-        if (cmd == .RunCommand or cmd == .AutoCommand or cmd == .BunxCommand) {
```
RiskyMH, line 1159: this is because `BunxCommand` doesn't run `Arguments.parse`, so may as well simplify.

src/cli.zig
```diff
@@ -2664,21 +2669,6 @@ pub const Command = struct {
...
-            .RemoveCommand = true,
-            .UpdateCommand = true,
-            .PatchCommand = true,
-            .PatchCommitCommand = true,
-            .PackageManagerCommand = true,
-            .BunxCommand = true,
-            .OutdatedCommand = true,
-            .PublishCommand = true,
-        });
-
```
RiskyMH, line 2641: once i added the `bun run` ones, this list wasn't any different to the base `loads_config`

src/cli/bunx_command.zig
```diff
@@ -315,6 +315,10 @@ pub const BunxCommand = struct {
             true,
         );
 
+        if (!ctx.debug.loaded_bunfig) {
+            try bun.CLI.Arguments.loadConfig(ctx.allocator, null, ctx, comptime .BunxCommand);
+        }
+
```
RiskyMH, line 360: right now you cant `--config=./path/bunfig.toml`, but I don't think many people using `bunx` would need that - its more for global `--bun`.

src/cli.zig
```diff
@@ -186,7 +186,7 @@ pub const Arguments = struct {
     const base_params_ = (if (Environment.isDebug) debug_params else [_]ParamType{}) ++ [_]ParamType{
         clap.parseParam("--env-file <STR>...               Load environment variables from the specified file(s)") catch unreachable,
         clap.parseParam("--cwd <STR>                       Absolute path to resolve files & entry points from. This just changes the process' cwd.") catch unreachable,
-        clap.parseParam("-c, --config <PATH>?              Specify path to Bun config file. Default <d>$cwd<r>/bunfig.toml") catch unreachable,
+        clap.parseParam("-c, --config <PATH>               Specify path to Bun config file. Default <d>$cwd<r>/bunfig.toml") catch unreachable,
```
RiskyMH: This is a breaking change as people may have `bun -c ./index.ts` in their scripts to use local `bunfig.toml`. However in this PR it already does that by default (as its what most expect now), so this could cause problems...

RiskyMH: done because technically its not needed, but this appears to create more problems rn, will consider this change in followup (or allowing `-c` itself at very least)

docs/runtime/bunfig.md
```diff
@@ -457,8 +457,6 @@ editor = "code"
 
 The `bun run` command can be configured under the `[run]` section. These apply to the `bun run` command and the `bun` command when running a file or executable or script.
 
-Currently, `bunfig.toml` isn't always automatically loaded for `bun run` in a local project (it does check for a global `bunfig.toml`), so you might still need to pass `-c` or `-c=bunfig.toml` to use these settings.
-
```
RiskyMH, line 461: this pr does also change this to default import local project `bunfig.toml` as its more expected and inline with `bun install`

did temporally include change that removed `-c`  by itself as a feature, but decided to not pursue as really messed some things up.

PR #15613, opened 2024-12-06
Update ssr-react.md

kjjd84:
### What does this PR do?

- [ ] Code changes

### DIFF
--- a/docs/guides/ecosystem/ssr-react.md
+++ b/docs/guides/ecosystem/ssr-react.md
@@ -2,11 +2,17 @@
 name: Server-side render (SSR) a React component
 ---
 
-To get started, install the canary version of `react` & `react-dom`:
+To get started, install `react` & `react-dom`:
 
 ```sh
 # Any package manager can be used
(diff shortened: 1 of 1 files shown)


versecafe (commented):
@kjjd84 maybe specify react 19 or above so people don't know all their libraries need to be 19.0.0 or 19-rc2 compatible

kjjd84:
> @kjjd84 maybe specify react 19 or above so people don't know all their libraries need to be 19.0.0 or 19-rc2 compatible

react 19 is already out which is why i made this PR

versecafe:
> > @kjjd84 maybe specify react 19 or above so people don't know all their libraries need to be 19.0.0 or 19-rc2 compatible
> 
> react 19 is already out which is why i made this PR

Yes but many existing libraries are not react 19 compatible yet or are only compatible on rcs and canaries, which is why I recommended specifying the minimum version just as the prior version did

RiskyMH:
superseded by #16350

PR #7481, opened 2023-12-06
test: init sharp integration test

ImBIOS:
### What does this PR do?

- #5852 

All 5 basic tests passing

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

### How did you verify your code works?

I ran `bun test` locally

PR #15626, opened 2024-12-06
Fix importing empty .txt file

### LINKED ISSUE #15536: Bun.build Incorrect type | 不正确的类型

_No response_

_No response_

When the imported text is empty, it will return a value of type Object, which is incorrect and should return an empty string.
当导入的文本为空时，将返回Object类型的值，这是错误的，应该返回空字符串。
Image
Image
Image
Image
❌
```js
var a_default = {};
```
✅
```js
var a_default = "";
```

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

pfgithub:
Fixes #15536 

Also removes 'buildWithResolveResult' because it is never used


robobun:
@pfgithub, your commit 7cd823feeb60cc6d3d4af47b7fd373b75a5df6fd is building: [`#7481`](https://buildkite.com/bun/bun/builds/7481)

Jarred-Sumner (changes requested):
Test failures

PR #7505, opened 2023-12-07
fix(resolver): allow builtins to be imported via subpath imports

paperclover:
### What does this PR do?

Fixes #4972 
Fixes https://github.com/oven-sh/bun/issues/7488

PR #15629, opened 2024-12-07
Fix missing "readable" events

heimskr:
### What does this PR do?

Fixes an issue where depending on the size of a file opened with `fs.createReadStream`, a final `readable` event (during which `.read()` returns `null`) can be mistakenly not emitted. This causes, for example, `prisma init` to exit without doing anything.

- [x] Code changes

I added some tests to `/test/js/node/stream/node-stream.test.js` and I also ran `bun --bun x prisma init`.

### DIFF
--- a/src/js/node/stream.ts
+++ b/src/js/node/stream.ts
@@ -5371,14 +5371,18 @@ function createNativeStreamReadable(Readable) {
       }
 
       if (isClosed) {
-        nativeReadable.push(null);
+        ProcessNextTick(() => {
+          nativeReadable.push(null);
+        });
       }
 
       return remainder.byteLength > 0 ? remainder : undefined;
     }
 
     if (isClosed) {
-      nativeReadable.push(null);
+      ProcessNextTick(() => {
+        nativeReadable.push(null);
+      });
     }
 
     return view;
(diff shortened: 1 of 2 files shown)


robobun:
@heimskr, your commit 0cab72634faa43d432e87f7e6ea60da595234acb is building: [`#7505`](https://buildkite.com/bun/bun/builds/7505)

Jarred-Sumner (changes requested):
Tests with hardcoded delays are either optimistically flaky or pessimistically slow by design. We should avoid adding hardcoded delays as much as we can.

When you write a test with a hardcoded amount of time to delay, ask yourself "is there a more specific source of truth we could use?"

We aren't trying to wait 20ms. We are trying to wait until both the `"close"` and the `"readable"` events are emitted. So, let's do that.

heimskr:
> We aren't trying to wait 20ms. We are trying to wait until both the `"close"` and the `"readable"` events are emitted. So, let's do that.

My fear is that eventually some future regression might cause either event to not be emitted, in which case the test would never complete. How should I handle this?

Jarred-Sumner:
> My fear is that eventually some future regression might cause either event to not be emitted, in which case the test would never complete. 

The test will timeout and we will see it in CI

Issue #7515, opened 2023-12-07
bun build --compile fails with EACCESS trying to write output


humphd:
### What version of Bun is running?

1.0.15

### What platform is your computer?

Linux 6.4.16-linuxkit aarch64 unknown (Linux Dev Contianer on MacOS))

### What steps can reproduce the bug?

1. Create a file, `file.ts`: `console.log('hello world');`
2. Create an executable: `bun build file.ts --compile --outfile ./file`

### What is the expected behavior?

I'd expect to have an executable created

### What do you see instead?

```
 [13ms]  bundle  1 modules
error: failed to open temporary file to copy bun into
EACCES: Permission denied (open())
```

### Additional information

The two calls that are failing in this `strace` are interesting, specifically:

```
[pid 19234] openat(AT_FDCWD, ".179e939cefeff7f3-00000000.bun-build", O_RDWR|O_CREAT|O_CLOEXEC, 000) = -1 EACCES (Permission denied)
[pid 19234] openat(AT_FDCWD, "/tmp.179e939cefeff7f3-00000000.bun-build", O_RDWR|O_CREAT|O_CLOEXEC, 000) = -1 EACCES (Permission denied)
```

I'm not clear why it can't write to `.` (I have permissions to do it), but should the second not be `/tmp/.179e939cefeff7f3-00000000.bun-build` (trying to write to the root)?

Full trace:

```sh
strace -e open,openat -f bun build parge.ts --compile --outfile ./parge2
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/lib/aarch64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/lib/aarch64-linux-gnu/libpthread.so.0", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/lib/aarch64-linux-gnu/libdl.so.2", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/lib/aarch64-linux-gnu/libm.so.6", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/proc/sys/vm/overcommit_memory", O_RDONLY) = 3
openat(AT_FDCWD, "/workspaces/DeepStructure/ds/util/bunfig.toml", O_RDONLY) = -1 ENOENT (No such file or directory)
[...]

PR #15632, opened 2024-12-07
Native plugin follow up

zackradisic:
### What does this PR do?

- Only create `NapiModuleMeta` if the native module exports `BUN_PLUGIN_NAME`
- Add tests to make sure the passed in napiModule is actually a napiModule and has the `BUN_PLUGIN_NAME` symbol
- Add a macro wrapper to the Rust bindings to make it nicer to write plugins
- Use Rust's [`panic::catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) function to catch panics so they can be gracefully reported to the user
- Fix a stupid bug where the Rust function `define_bun_plugin!(...)` was not null terminating the string


robobun:
@zackradisic, your commit 5a208023ea88871bb61551082a4aa679ecdfec0e is building: [`#7515`](https://buildkite.com/bun/bun/builds/7515)

src/bun.js/bindings/JSBundlerPlugin.cpp
```diff
@@ -243,18 +236,16 @@ void BundlerPlugin::NativePluginList::append(JSC::VM& vm, JSC::RegExp* filter, S
...
         if (nsGroup == nullptr) {
             namespaces.append(namespaceString);
-            groups.append(Vector<NativeFilterRegexp> {});
+            groups.append(Vector<FilterRegExp> {});
             nsGroup = &groups.last();
             index = namespaces.size() - 1;
         }
 
-        Yarr::RegularExpression regex(
+        Yarr::RegularExpression&& regex = Yarr::RegularExpression(
```
190n:
```suggestion
        Yarr::RegularExpression regex = Yarr::RegularExpression(
```

From a bit of local testing (happy to be proven wrong) I don't believe the `&&` here affects anything. To avoid copying `regex` (assuming that is even worthwhile) you need explicit `WTFMove` calls and/or rvalue-reference constructors, and in those cases making this local variable `&&` isn't necessary.

packages/bun-native-plugin-rs/README.md
```diff
@@ -119,19 +142,14 @@ const result = await Bun.build({
...
-    // Bun knows this plugin failed.
-    handle.log_error("Failed to fetch source code!");
-    return;
-  }
-};
-```
+It is highly advised that you return all errors and avoid `.unwrap()`'ing or `.expecting()`'ing results.
+
+The `#[bun]` wrapper macro actually runs your code inside of a [`panic::catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html),
+which may catch _some_ panics but **not guaranteed to catch all panics**.
```
190n, line 150:
```suggestion
which may catch _some_ panics but **is not guaranteed to catch all panics**. Uncaught panics will cause the entire Bun process to crash.
```

src/bun.js/bindings/JSBundlerPlugin.cpp
```diff
@@ -271,45 +262,50 @@ void BundlerPlugin::NativePluginList::append(JSC::VM& vm, JSC::RegExp* filter, S
...
-    if (group == nullptr) {
+    auto* groupPtr = this->group(namespaceStr->toWTFString(BunString::ZeroCopy), index);
+    if (groupPtr == nullptr) {
         return -1;
     }
+    auto& filters = *groupPtr;
 
     const auto& callbacks = index == std::numeric_limits<unsigned>::max() ? this->fileCallbacks : this->namespaceCallbacks[index];
-    ASSERT_WITH_MESSAGE(callbacks.size() == group->size(), "Number of callbacks and filters must match");
+    ASSERT_WITH_MESSAGE(callbacks.size() == filters.size(), "Number of callbacks and filters must match");
```
190n, line 287: What determines the sizes of these arrays? Can user code/configuration cause them to have mismatched sizes, or are they built by Bun in a way that should always match?

src/bun.js/bindings/JSBundlerPlugin.cpp
```diff
@@ -271,45 +262,50 @@ void BundlerPlugin::NativePluginList::append(JSC::VM& vm, JSC::RegExp* filter, S
...
-                CrashHandler__setInsideNativePlugin(name);
-                callback(onBeforeParseArgs, onBeforeParseResult);
-                CrashHandler__setInsideNativePlugin(nullptr);
-
-                count++;
+
+        if (filters[i].match(vm, path)) {
+            Bun::NapiExternal* external = callbacks[i].external;
+            if (external) {
+                ((OnBeforeParseArguments*)(onBeforeParseArgs))->external = external->value();
```
190n: I think the field name `external` is a misnomer, because it doesn't contain a pointer to the external itself, it directly contains the `void*` that was stored by the external.

Do native plugins actually work with NAPI externals directly? Or do they only ever see the `void*`?

src/bun.js/bindings/JSBundlerPlugin.cpp
Jarred-Sumner: It should always match because they're not appended to by userland code. It's basically a MultiArrayList

PR #7523, opened 2023-12-07
test: add web worker tests for .ts and .tsx worker file

jerome-benoit:

PR #15633, opened 2024-12-07
Adds support to mark a test file as --todo with {todo: "true"}

pfgithub:
### What does this PR do?

`test.todo()` is basically just `test.skip()` with a different name unless you use the `--todo` flag. This adds the ability to mark specific files to always run todo tests and ensure they fail, so that specific files can opt-in to running todo tests.

```ts
// a.test.ts
import { test, expect } from "bun:test" with { todo: "true" };

test.todo("fixme", () => expect(value()).toBe(1));
```


robobun:
@pfgithub, your commit e93b0c4646353432a2996d5e31b16ec80e913a2f is building: [`#7523`](https://buildkite.com/bun/bun/builds/7523)

pfgithub:
#17864 is better

Issue #7544, opened 2023-12-09
Nock disableNetConnect() does not work in Bun


vlucas:
### What version of Bun is running?

1.0.15+b3bdf22eb

### What platform is your computer?

Darwin 22.3.0 arm64 arm

### What steps can reproduce the bug?

1. Start a new project
2. Install `nock` (`bun add nock`)
3. Add the following contents in a new file called `test.test.ts`:

```
import nock from "nock";
import { test, expect } from "bun:test";

test("it should disable real HTTP calls", () => {
  // Should disable all network calls in tests - currently does NOT in Bun
  nock.disableNetConnect();

  expect(async () => {
    await fetch("https://oven.sh");
  }).toThrow();
});
```

4. Run the file with `bun test`
5. The test fails when it should pass - live network calls are still happening in Bun

### What is the expected behavior?

Bun test should pass, because the live network call should be blocked, which will cause `fetch` to throw:

```
bun test
bun test v1.0.15 (b3bdf22e)

test.test.ts:
✓ it should disable real HTTP calls [152.43ms]

 1 pass
 0 fail
 1 expect() calls
Ran 1 tests across 1 files. [192.00ms]
```

### What do you see instead?

Bun fails, because live network calls are NOT disabled:

```
> bun test
bun test v1.0.15 (b3bdf22e)

test.test.ts:
 5 |   // Should disable all network calls in tests - currently does NOT in Bun
 6 |   nock.disableNetConnect();
 7 |
 8 |   expect(async () => {
 9 |     await fetch("https://oven.sh");
10 |   }).toThrow();
                  ^
error: expect(received).toThrow()

Received function did not throw

      at /Users/vlucas/Code/test-bun-nock/test.test.ts:10:14
✗ it should disable real HTTP calls [148.44ms]

 0 pass
 1 fail
 1 expect() calls
Ran 1 tests across 1 files. [188.00ms]
```

vlucas:
Closing this, as this is not a bun issue. This is an issue with nock and built-in `fetch` in Node 18+.

PR #15638, opened 2024-12-07
new binding generator

paperclover:
pasted below is the docs as of opening.

# New Bindings Generator

The new bindings generator, introduced to the codebase in Dec 2024, scans for
`*.bind.ts` to find function and class definition, and generates glue code to
interop between JavaScript and native code.

There are currently other code generators and systems that achieve similar
purposes. The following will all eventually be completely phased out in favor of
this one:

- "Classes generator", converting `*.classes.ts` for custom classes.
- "JS2Native", allowing ad-hoc calls from `src/js` to native code.

## Creating JS Functions in Zig

Given a file implementing a simple function, such as `add`

```zig#src/bun.js/math.zig
pub fn add(global: *JSC.JSGlobalObject, a: i32, b: i32) !i32 {
    return std.math.add(i32, a, b) catch {
        // Binding functions can return `error.OutOfMemory` and `error.JSError`.
        // Others like `error.Overflow` from `std.math.add` must be converted.
        // Remember to be descriptive.
        return global.throwPretty("Integer overflow while adding", .{});
    };
}

const gen = bun.gen.math; // "math" being this file's basename

const std = @import("std");
const bun = @import("root").bun;
const JSC = bun.JSC;
```

Then describe the API schema using a `.bind.ts` function. The binding file goes next to the Zig file.

```ts#src/bun.js/math.bind.ts
import { t, fn } from 'bindgen';

fn({
  name: 'add',
  args: {
    global: t.globalObject,
    a: t.i32,
    b: t.i32.default(1),
  },
  ret: t.i32,
});
```

This function declaration is equivalent to:

```ts
/**
 * Throws if zero arguments are provided.
 * Wraps out of range numbers using modulo.
 */
declare function add(a: number, b: number = 1): number;
```

The code generator will provide `bun.gen.math.jsAdd`, which is the native function implementation. To pass to JavaScript, use `bun.gen.math.createAddCallback(global)`

## Strings

The type for receiving strings is one of [`t.DOMString`](https://webidl.spec.whatwg.org/#idl-DOMString), [`t.ByteString`](https://webidl.spec.whatwg.org/#idl-ByteString), and [`t.USVString`](https://webidl.spec.whatwg.org/#idl-USVString). These map directly to their WebIDL counterparts, and have slightly different conversion logic. Bindgen will pass BunString to native code in all cases.

When in doubt, use DOMString.

`t.UTF8String` can be used in place of `t.DOMString`, but will call `bun.String.toUTF8`. The native callback gets `[]const u8` (WTF-8 data) passed to native code, freeing it after the function returns.

TLDRs from WebIDL spec:

- ByteString can only contain valid latin1 characters. It is not safe to assume bun.String is already in 8-bit format, but it is extremely likely.
- USVString will not contain invalid surrogate pairs, aka text that can be represented correctly in UTF-8.
- DOMString is the loosest but also most recommended strategy.

## Function Variants

A `variants` can specify multiple variants (also known as overloads).

```ts#src/bun.js/math.bind.ts
import { t, fn } from 'bindgen';

fn({
  name: 'action',
  variants: [
    {
      args: {
        a: t.i32,
      },
      ret: t.i32,
    },
    {
      args: {
        a: t.DOMString,
      },
      ret: t.DOMString,
    },
  ]
});
```

In Zig, each variant gets a number, based on the order the schema defines.

```
fn action1(a: i32) i32 {
  return a;
}

fn action2(a: bun.String) bun.String {
  return a;
}
```

## `t.dictionary`

A `dictionary` is a definition for a JavaScript object, typically as a function inputs. For function outputs, it is usually a smarter idea to declare a class type to add functions and destructuring.

## Enumerations

To use [WebIDL's enumeration](https://webidl.spec.whatwg.org/#idl-enums) type, use either:

- `t.stringEnum`: Create and codegen a new enum type.
- `t.zigEnum`: Derive a bindgen type off of an existing enum in the codebase.

An example of `stringEnum` as used in `fmt.zig` / `bun:internal-for-testing`

```ts
export const Formatter = t.stringEnum(
  "highlight-javascript",
  "escape-powershell",
);

fn({
  name: "fmtString",
  args: {
    global: t.globalObject,
    code: t.UTF8String,
    formatter: Formatter,
  },
  ret: t.DOMString,
});
```

WebIDL strongly encourages using kebab case for enumeration values, to be consistent with existing Web APIs.

### Deriving enums from Zig code

TODO: zigEnum

## `t.oneOf`

A `oneOf` is a union between two or more types. It is represented by `union(enum)` in Zig.

TODO:

## Attributes

There are set of attributes that can be chained onto `t.*` types. On all types there are:

- `.required`, in dictionary parameters only
- `.optional`, in function arguments only
- `.default(T)`

When a value is optional, it is lowered to a Zig optional.

Depending on the type, there are more attributes available. See the type definitions in auto-complete for more details. Note that one of the above three can only be applied, and they must be applied at the end.

### Integer Attributes

Integer types allow customizing the overflow behavior with `clamp` or `enforceRange`

```ts
import { t, fn } from "bindgen";

fn({
  name: "add",
  args: {
    global: t.globalObject,
    // enforce in i32 range
    a: t.i32.enforceRange(),
    // clamp to u16 range
    c: t.u16,
    // enforce in arbitrary range, with a default if not provided
    b: t.i32.enforceRange(0, 1000).default(5),
    // clamp to arbitrary range, or null
    d: t.u16.clamp(0, 10).optional,
  },
  ret: t.i32,
});
```

## Callbacks

TODO

## Classes

TODO


robobun:
@paperdave, your commit 0c717ca6fe087245f6065960ff0edc4114dec474 is building: [`#7544`](https://buildkite.com/bun/bun/builds/7544)

src/codegen/bindgen-lib-internal.ts
```diff
@@ -0,0 +1,1023 @@
...
+    const cAbiType = this.canDirectlyMapToCAbi();
+    const namespace = typeHashToNamespace.get(this.hash());
+    if (cAbiType && typeof cAbiType === "string" && this.kind !== "zigEnum" && this.kind !== "stringEnum") {
+      return cAbiTypeName(cAbiType);
+    }
+    return namespace ? `${namespace}::${cap(name)}` : name;
+  }
+
+  #generateName() {
+    return `anon_${this.ownerFile}_${this.hash()}`;
```
Jarred-Sumner:
```suggestion
    return `gen_${this.ownerFile}_${this.hash()}`;
```

src/codegen/bindgen.ts
```diff
@@ -0,0 +1,1262 @@
...
+  } else {
+    switch (type.kind) {
+      case "any":
+        cpp.line(`${storageLocation} = JSC::JSValue::encode(${jsValueRef});`);
+        break;
+      case "USVString":
+      case "DOMString":
+      case "ByteString":
+        cpp.line(
+          `${storageLocation} = Bun::toString(WebCore::convert<WebCore::IDL${type.kind}>(*global, ${jsValueRef}));`,
```
Jarred-Sumner: This needs to check for exceptions before calling Bun::toString

src/codegen/bindgen.ts
```diff
@@ -0,0 +1,1262 @@
...
+        );
+        cpp.line(`RETURN_IF_EXCEPTION(throwScope, {});`);
+        break;
+      case "UTF8String":
+        cpp.line(
+          `${storageLocation} = Bun::toString(WebCore::convert<WebCore::IDLDOMString>(*global, ${jsValueRef}));`,
```
Jarred-Sumner: This also needs to check for exceptions first

src/codegen/bindgen.ts
```diff
@@ -0,0 +1,1262 @@
...
+) {
+  const [kind, min, max] = range;
+  if (kind === "clamp") {
+    cpp.line(`if (${storageLocation} < ${min}) ${storageLocation} = ${min};`);
+    cpp.line(`else if (${storageLocation} > ${max}) ${storageLocation} = ${max};`);
+  } else if (kind === "enforce") {
+    cpp.line(`if (${storageLocation} < ${min} || ${storageLocation} > ${max}) {`);
+    cpp.indent();
+    cpp.line(
+      `throwTypeError(global, throwScope, rangeErrorString<${cAbiTypeName(cAbiType)}>(${storageLocation}, ${min}, ${max}));`,
```
Jarred-Sumner, line 490: We're probably going to want to return the Node.js error code for this.

src/codegen/bindgen.ts
```diff
@@ -0,0 +1,1262 @@
...
+  cpp.line(`}`);
+  cpp.line(`JSC::JSValue propValue;`);
+
+  for (const field of fields) {
+    const { key, type: fieldType } = field;
+    cpp.line("// " + key);
+    cpp.line(`if (isNullOrUndefined) {`);
+    cpp.line(`    propValue = JSC::jsUndefined();`);
+    cpp.line(`} else {`);
+    cpp.line(`    propValue = object->get(global, JSC::Identifier::fromString(vm, ${str(key)}_s));`);
```
Jarred-Sumner: Use our prototype pollution mitigation
```suggestion
    cpp.line(`    propValue = Bun::getIfPropertyExistsPrototypePollutionMitigation(global, JSC::Identifier::fromString(vm, ${str(key)}_s));`);
```

PR #15651, opened 2024-12-08
Fix rare crash in bun install

Jarred-Sumner:
### What does this PR do?

When a workspace package contains an invalid package.json, we write to a hash table without initializing one of the pointers.

This would ideally have a test. Not sure yet how to reproduce the crash exactly.

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -2919,7 +2919,10 @@ pub const PackageManager = struct {
                 },
             );
 
-            const json = json_result catch |err| return .{ .parse_err = err };
+            const json = json_result catch |err| {
+                _ = this.map.remove(path);
+                return .{ .parse_err = err };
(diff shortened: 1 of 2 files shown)


robobun:
@Jarred-Sumner, your commit bdbf65862f66a5d8a7d40ec231159c43ffd34d1b is building: [`#7592`](https://buildkite.com/bun/bun/builds/7592)

src/install/install.zig
```diff
@@ -2859,6 +2864,9 @@ pub const PackageManager = struct {
                     .guess_indentation = opts.guess_indentation,
                 },
             ) catch |err| {
+                _ = this.map.remove(key);
```
dylan-conway, line 2867: will need the same on `getWithSource`

Issue #7619, opened 2023-12-13
Error: Cannot find module 'detect-libc'


ollecoffee:
### What version of Bun is running?

v1.0.17 (5e60861c)

### What platform is your computer?

Github Pages

### What steps can reproduce the bug?

```
bun install v1.0.17 (5e60861c)
node:internal/modules/cjs/loader:1147
  throw err;
  ^

Error: Cannot find module 'detect-libc'
Require stack:
- /home/runner/work/olle.coffee/olle.coffee/node_modules/astro/node_modules/sharp/install/libvips.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at Object.<anonymous> (/home/runner/work/olle.coffee/olle.coffee/node_modules/astro/node_modules/sharp/install/libvips.js:13:20)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/runner/work/olle.coffee/olle.coffee/node_modules/astro/node_modules/sharp/install/libvips.js'
  ]
}

Node.js v21.4.0
node:internal/modules/cjs/loader:1147
  throw err;
  ^

Error: Cannot find module 'semver/functions/coerce'
Require stack:
- /home/runner/work/olle.coffee/olle.coffee/node_modules/astro/node_modules/sharp/lib/libvips.js
- /home/runner/work/olle.coffee/olle.coffee/node_modules/astro/node_modules/sharp/install/can-compile.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at Object.<anonymous> (/home/runner/work/olle.coffee/olle.coffee/node_modules/astro/node_modules/sharp/lib/libvips.js:10:22)
[...]

PR #15657, opened 2024-12-08
docs: add comments to JSNextTickQueue

DonIsaac:
### What does this PR do?
Adds comments describing `JSNextTickQueue`, providing important context to it's internal slots and what they do.

No logic changes.

### DIFF
--- a/src/bun.js/bindings/JSNextTickQueue.cpp
+++ b/src/bun.js/bindings/JSNextTickQueue.cpp
@@ -37,6 +37,7 @@ JSNextTickQueue* JSNextTickQueue::create(VM& vm, Structure* structure)
     mod->finishCreation(vm);
     return mod;
 }
+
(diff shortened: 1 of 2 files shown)


robobun:
Your commit 09a79946556d09884c586592d5fb4a63d634ea14 is building: [`#7619`](https://buildkite.com/bun/bun/builds/7619)

src/bun.js/bindings/JSNextTickQueue.cpp
```diff
@@ -69,9 +67,14 @@ JSNextTickQueue* JSNextTickQueue::create(JSC::JSGlobalObject* globalObject)
     return obj;
 }
 
+WriteBarrier<Unknown>& JSNextTickQueue::queueStatus() { return internalField(0); }
```
Jarred-Sumner, line 70: If you want to document what each field means, you could use an `class enum Field : uint8_t` instead of adding dead code

We should avoid returning WriteBarrier as well, since it's not useful in of itself (you want the value, not the barrier)

Issue #7633, opened 2023-12-13
ResolveError: Cannot find module "typeorm/metadata/ColumnMetadata"


alumni:
### What version of Bun is running?

1.0.17+5e60861c3

### What platform is your computer?

Darwin 22.6.0 arm64 arm

### What steps can reproduce the bug?

1. `bun init && bun add typeorm@0.3.17`.

2. Either in a script (`index.ts`) or in REPL:
```js
import { ColumnMetadata } from 'typeorm/metadata/ColumnMetadata';
```
or just:
```js
require('typeorm/metadata/ColumnMetadata');
```

### What is the expected behavior?

I would expect be able to import the `ColumnMetadata` class.

### What do you see instead?

REPL:
```
Uncaught ResolveError: Cannot find module "typeorm/metadata/ColumnMetadata" from <repo>/$bun$repl.ts
```

`bun run index.ts`:
```
error: Cannot find module "typeorm/metadata/ColumnMetadata" from "<repo>/index.ts"
```

### Additional information

Seems `bun` expects the extension `.js` to be in the import (even though it's not an ESM module?)

Jarred-Sumner:
TypeORM has disabled implicit file extensions in their package.json. This is, unfortunately, not a bug in Bun and  it is how package.json `"exports"` works - package authors can choose what and how their packages are imported.

Adding the `.js` file extension to the end works:
```js
import { ColumnMetadata } from 'typeorm/metadata/ColumnMetadata.js';
```

as does `require`:
```js
require('typeorm/metadata/ColumnMetadata');
```


Specifically, these lines are relevant: https://github.com/typeorm/typeorm/blob/master/package.json#L35-L36

If you continue to see issues with `require`, let us know (I was unable to reproduce that)

alumni:
Ah probably I simplified too much and only tried the `import` without trying CJS. That's right, I don't get the error with `require` in the minimal example I mentioned above. But it does happen for `require` in the original repo which I tried to switch to bun.

I actually discovered this while trying to run the production build of an existing app (compiled with webpack + tsc running under node). The production build is pure CJS and only has a simple `package.json`. In the source repo for that app, I get:

```
➜  <repo> bun repl
Welcome to Bun v1.0.18
Type ".help" for more information.
[!] Please note that the REPL implementation is still experimental!
    Don't consider it to be representative of the stability or behavior of Bun overall.

> require('typeorm/metadata/ColumnMetadata')
[...]

PR #15670, opened 2024-12-09
test(ws): do not create temporary `.sock` files in root repo directory

DonIsaac:
### What does this PR do?
Refactors `inspect.test.ts` so that temporary socket files are created in a temp dir instead of bun's root directory.

Only test files are affected.


robobun:
Your commit b910cddef6407b9b177fe019b629f6971da98737 is building: [`#7633`](https://buildkite.com/bun/bun/builds/7633)

Jarred-Sumner:
This may cause test failures because of the file path length limits. That’s why we do this currently.

DonIsaac:
@Jarred-Sumner changed it to create `.sock` files in `<repo-root>/.tmp/` instead of the OS's temp dir. `.tmp` gets removed after all tests run.

PR #15678, opened 2024-12-10
Introduce `env` option in Bun.build() and `bun build` to let you inject `FOO_PUBLIC_*`-style env vars

Jarred-Sumner:
### What does this PR do?

This makes it simplier to inject environment variables get inlined into the bundle.

For example, to inject all NEXT_PUBLIC_* environment variables set `env: "NEXT_PUBLIC_*"` or `--env=NEXT_PUBLIC_*`

The main caveat with this implementation is that it relies on the system-provided environment variables instead of process.env - meaning that modifications to process.env are ignored. In the future, we probably should add a way to pass an object to inline instead, but we can wait for someone to ask for that before building it.

There are tests.

### DIFF
--- a/docs/bundler/index.md
+++ b/docs/bundler/index.md
@@ -570,20 +570,19 @@ $ FOO=bar BAZ=123 bun build ./index.tsx --outdir ./out --env inline
 
 {% /codetabs %}
 
-For example, given the following environment variables:
+For the input below:
 
-```bash
-$ FOO=bar BAZ=123
+```js#input.js
+console.log(process.env.FOO);
+console.log(process.env.BAZ);
 ```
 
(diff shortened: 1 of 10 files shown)


robobun:
@Jarred-Sumner, your commit c60bc3f325ba067ff38f0e53ccaee6d79140240a is building: [`#7690`](https://buildkite.com/bun/bun/builds/7690)

paperclover (approved):

nektro:
`test/bundler/bundler_edgecase.test.ts` fails are real

nektro (changes requested):
`test/bundler/bundler_edgecase.test.ts` fails are real

PR #15687, opened 2024-12-10
fix(codegen): better error messages for internals using module.exports

DonIsaac:
### What does this PR do?

Ensures internal modules have a default export. Before this, codegen would pass but compilation would fail further along in the process.

This is a codegen change that does not affect output code. If this builds, it works.


robobun:
Your commit f8896a4dde841072370b9446ffb4fc7f37b3ecb9 is building: [`#7716`](https://buildkite.com/bun/bun/builds/7716)

paperclover (approved):
nice

Issue #7721, opened 2023-12-18
"unexpected end of file" error with function and generic


samhuk:
### What version of Bun is running?

1.0.18+36c316a24

### What platform is your computer?

Darwin 23.2.0 x86_64 i386

### What steps can reproduce the bug?

```ts
const a = <T = any>(): T => null as T
const b = a<string>
```

Runs fine in node and browsers ([tsplay](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAPAFSfMBPAfACgEoAuGDRXGMAVwBtb4IyAoUSWAI0zhWgCcAlmADmuZkA)), but fails with Bun with "unexpected end of file" error.

Adding semi-colon on the end of L2 is a workaround.

### What is the expected behavior?

Executes successfully

### What do you see instead?

"unexpected end of file" error.

PR #15690, opened 2024-12-10
allow importing internals when --expose-internals

paperclover:
This allows node tests running bun with --expose-internals to import our internal files. This means that for the node tests checking internals. This flag is implied in debug builds.


robobun:
@paperdave, your commit f47277988f7e9dc1ab8d60c8ddc2f69bddbcb9ab is building: [`#7721`](https://buildkite.com/bun/bun/builds/7721)

paperdave:
meghan has convinced me we shouldn't do this, and instead just not test any internals. instead, we can copy any relevant tests and test them through `bun:internal-for-testing`

PR #15697, opened 2024-12-11
build: trigger cpp build when generated ErrorCode files change

DonIsaac:
### What does this PR do?

Fixes `cmake` so that cpp code in `build/codegen` re-compiles when
`ErrorCode.{h,cpp,ts}` files change.

### How did you verify your code works?
I ran build.

Upstack from #15696

PR #15698, opened 2024-12-11
test(node): get `test-assert.js` working

DonIsaac:
### What does this PR do?

- Overhauls `node:assert` to better match node's behavior
- Adds `test-assert.js` to the list of passing node test cases

Upstack from #15697

I ported node test cases.


robobun:
Your commit 710cbf7b338c2bd266ed459f4128e07a3bb1d5e6 is building: [`#7763`](https://buildkite.com/bun/bun/builds/7763)

src/bun.js/base.zig
```diff
@@ -94,6 +101,11 @@ pub fn toJS(globalObject: *JSC.JSGlobalObject, comptime ValueType: type, value:
                 return value.toJS(globalObject);
             }
 
+            // must come after toJS check in case this enum implements its own serializer.
```
paperclover, line 104: this default is suspicious to me

paperclover (commented):
overall seems reasonable

src/bun.js/base.zig
DonIsaac: What do you suggest it should be instead, if anything?

paperclover: an error asking the caller to specify what they meant, instead of implicitly choosing for them.

DonIsaac: Erroring here prevents comptime marshaling of structs with enum fields

This should work out of the box. It will not work without automatic enum serialization:
```zig
const Foo = enum { a, b };
const Bar = struct {
  foo: Foo,
  bar: []const u8,
};
const zig: Foo = .{ ... };
const js = JSC.toJS(global, Bar, zig, .allocated);
```

I've also added doc comments on `toJS` communicating this behavior to prevent ambiguity/usage mistakes.

paperclover: imo its ambiguous if it means `{ bar: "...", foo: "a" }` or `{ bar: "...", foo: 0 }`. it's good the comment describes this default, but my issue is the fact that default is chosen. it will probably cause a bug down the line.

is this used right now? if so i wont block the pr on it, but if it isn't please remove it.

DonIsaac: It's being used to serialize `Diff` structs.

src/js/internal/util.ts
```diff
@@ -0,0 +1,7 @@
+const { isNativeError } = require("node:util/types");
+
+/**
+ * @fixme errors thrown in a different VM context are neither `isNativeError`
+ * nor `instanceof Error`.
+ */
+export const isError = (err: unknown): err is Error => isNativeError(err) || err instanceof Error || (typeof err === 'object' && err !== null && 'name' in err && 'message' in err && 'stack' in err);
```
Jarred-Sumner: use Error.isError instead

src/js/node/assert.ts
```diff
@@ -1,1271 +1,998 @@
...
-      },
-      TypeError,
-    );
-    module2.exports.codes = codes;
-  },
-});
+    if (argsLen === 2) operator = "!=";
+  }
+
+  if (message instanceof Error) throw message;
```
Jarred-Sumner, line 142: instanceof can throw

can we do `Error.isError(message)` instead

src/js/internal/util.ts
DonIsaac: done

src/js/node/assert.ts
DonIsaac: Done

src/js/internal/primordials.js
```diff
@@ -110,9 +114,17 @@ export default {
...
   DatePrototypeToString: uncurryThis(Date.prototype.toString),
+  Error,
   ErrorCaptureStackTrace,
   ErrorPrototypeToString: uncurryThis(Error.prototype.toString),
+  FunctionPrototypeBind: uncurryThis(Function.prototype.bind),
+  FunctionPrototypeCall: uncurryThis(Function.prototype["call"]),
+  FunctionPrototypeToString: uncurryThis(Function.prototype.toString),
   JSONStringify: JSON.stringify,
+  MapPrototypeDelete: uncurryThis(Map.prototype.delete),
+  MapPrototypeGet: uncurryThis(Map.prototype.get),
```
Jarred-Sumner: each of these are unnecessary, `.$get`, `.$set`, are defined

src/js/internal/primordials.js
```diff
@@ -144,10 +156,13 @@ export default {
   ObjectIs: Object.is,
   ObjectKeys: Object.keys,
   ObjectPrototypeHasOwnProperty: uncurryThis(Object.prototype.hasOwnProperty),
+  ObjectPrototypeIsPrototypeOf: uncurryThis(Object.prototype.isPrototypeOf),
   ObjectPrototypePropertyIsEnumerable: uncurryThis(Object.prototype.propertyIsEnumerable),
   ObjectPrototypeToString: uncurryThis(Object.prototype.toString),
   ObjectSeal: Object.seal,
   ObjectSetPrototypeOf: Object.setPrototypeOf,
+  ReflectApply: Reflect["apply"],
```
Jarred-Sumner: `.$apply`

src/js/node/assert.ts
```diff
@@ -1,1257 +1,1016 @@
...
+  SafeSet,
+  SafeWeakSet,
+  StringPrototypeIndexOf,
+  StringPrototypeSlice,
+  StringPrototypeSplit,
+  SymbolIterator,
+} = require("internal/primordials");
+
+const { Buffer } = require("node:buffer");
+const { deprecate } = require("node:util");
```
Jarred-Sumner: did we need to import node:util at the top of this file (i.e. in all cases)?

src/js/node/assert.ts
```diff
@@ -1,1257 +1,1016 @@
...
+ * @param {any} actual - The actual value to compare.
+ * @param {any} expected - The expected value to compare.
+ * @param {Set} [comparedObjects=new Set()] - Set to track compared objects for handling circular references.
+ * @returns {boolean} - Returns `true` if the actual value matches the expected value, otherwise `false`.
+ * @example
+ * compareBranch({a: 1, b: 2, c: 3}, {a: 1, b: 2}); // true
+ */
+function compareBranch(actual, expected, comparedObjects) {
+  // Check for Map object equality
+  if (isMap(actual) && isMap(expected)) {
```
Jarred-Sumner, line 390: Bun.deepEquals supports Map. There is a `strict` argument

src/js/internal/primordials.js
DonIsaac: Done. $get, $set, $has, and $delete have been swapped in.

src/js/node/assert.ts
DonIsaac: Refactored to be lazy loaded

src/js/internal/primordials.js
DonIsaac: image

DonIsaac: nevermind it's just `fn.$apply`

src/js/node/assert.ts
DonIsaac: Fixed

DonIsaac:
note: I had to move the node test `bunfig.toml` to the root directory. See #16289

Issue #7778, opened 2023-12-22
Node compat: Add `import.meta.dirname` and `import.meta.filename`


risu729:
### What version of Bun is running?

1.0.18+36c316a24

### What platform is your computer?

Linux 5.15.0-1048-oracle aarch64 aarch64

### What steps can reproduce the bug?

```js                                                                                    
// test.mjs
console.log("dir:", import.meta.dir);
console.log("dirname:", import.meta.dirname);
console.log("file:", import.meta.file);
console.log("filename:", import.meta.filename);
```

### What is the expected behavior?

```
$ node test.mjs
dir: undefined
dirname: /home/ubuntu
file: undefined
filename: /home/ubuntu/test.mjs
```

### What do you see instead?

```
$ bun run test.mjs
dir: /home/ubuntu
dirname: undefined
file: test.mjs
filename: undefined
```

### Additional information

Since Node.js v21.2.0, `import.meta.dirname` and `import.meta.filename` are added. https://github.com/nodejs/node/pull/48740

`import.meta.dirname` can be added to Bun as an alias for `import.meta.dir`.

However, while `import.meta.filename` in Node.js includes the directory but `import.meta.file` in Bun does not, `import.meta.filename` would not be just an alias.
`import.meta.filename` equals `${import.meta.dir}/${import.meta.file}`.

I think the behavior might cause some confusion, but I believe that adding these properties would improve compatibility with Node.js.

Node.js docs: https://nodejs.org/api/esm.html#importmeta
Bun docs: https://bun.sh/docs/api/import-meta

Jarred-Sumner:
Fixed in #8127

PR #15705, opened 2024-12-11
Implement text-based lockfile

### LINKED ISSUE #5486: Generate text lockfile along with binary lockfile (bun.lockb) and add "bun.lockb" to .gitignore

Bun generates binary lockfile (`bun.lockb`) by default and is not added to `.gitignore` file. This means `bun.lockb` mostly gets added to git repository (for example, https://github.com/elysiajs/elysia has `bun.lockb`).

The problem arises when merging branches that have changes in `bun.lockb` file. Since, `bun.lockb` file cannot be inspected, developer will most likely remove the `bun.lockb` file and generate a new one. This can be disastrous as the new locked version of a package is different from the previous one.

- Bun should generate text lockfile along with binary lockfile by default.
- Bun should generate binary lockfile based on text lockfile (not the other way around).
- Bun should add `bun.lockb` to `.gitignore` file by default to prevent merge conflict. Developer can force add `bun.lockb` to git repo if needed.

_No response_

### LINKED ISSUE #11863: Implement a text-based lockfile format
This is a tracking issue for implementing a text-based lockfile format for `bun install` and making it the default format going forward. There will be a smooth migration path from `bun.lockb` → `bun.lock`.  

### Why?

When first working on the package manager, the flamegraph showed parsing JSON was the biggest number. So instead of a JSON lockfile, we designed two efficient binary formats: one for the registry manifest cache and one for the lockfile.

The binary lockfile format has served us well, but it isn't worth the cost in developer experience, particularly for larger teams working together. We suggest workarounds today for various things, but none of them are great.

- Merge conflicts are hard. Do you pick `bun.lockb.1` or `bun.lockb.2`? What about both?
- How do you inspect the lockfile from a PR on GitHub/GitLab/etc? 
- How do you diff the lockfile? You can [print as a yarn.lock](https://bun.sh/guides/install/yarnlock) and configure git to diff that way, but that's a whole lot more complicated than it not being a problem in the first place
- When there are lockfile changes, it's not clear enough why. Sometimes, you run `bun install` and the lockfile has changes due to hashes changing, package.json `scripts` changing, or other reasons and this is hard to understand right now because it's hard to read in a text editor.
- People too frequently add `bun.lockb` to the `.gitignore` file 

### Will bun install still be fast?

Yes.

### Will it make bun install slower?

Based on what we've seen, about 1-20 milliseconds.

### What will the new format be?

Probably [JSON with Trailing Commas](https://code.visualstudio.com/docs/languages/json#_json-with-comments), like `tsconfig.json`.

Bun already supports this schema format for the runtime (and package.json). TOML is another option and it's what we use for `bunfig.toml`, though I kind of think that was a mistake and it should've been JSON with Trailing Commas. TOML's editor tooling support is not as mature as JSON with Trailing Commas.

#### Why [JSON with Trailing Commas](https://code.visualstudio.com/docs/languages/json#_json-with-comments) instead of JSON?

Too many collective human lifetimes have been spent fixing merge conflicts from diffs caused by adding or removing trailing commas at the end of lists in JSON

```diff
-    }
+    },
```

#### Why JSON with Trailing Commas instead of YAML?

YAML is fine. We don't have a YAML parser in Bun yet, and the indentation gets really confusing sometimes. Also, YAML parsers tend to be slower than JSON parsers (and YAML parsers are also JSON parsers). My favorite YAML fact is that the two-digit [country code for Norway in YAML](https://langdev.stackexchange.com/questions/1123/what-design-trade-offs-led-to-the-norway-problem-in-yaml-and-when-are-they-wo) is parsed as `false`, though this isn't relevant to a lockfile

#### Why JSON with Trailing Commas instead of JSON5?

We don't have a JSON5 parser in Bun, and would like to avoid formats that're slower to parse than JSON

### When will `bun.lock` be released?

Q3.

### What will the migration plan be?

`bun install` will support both `bun.lockb` and `bun.lock` for awhile, but once released, new features will only be supported via `bun.lock`.

#### Do people really edit lockfiles manually?

Yes, for small tweaks it can be important

#### Why not fix merge conflicts via `bun install` instead?

We will support that as well. But, it shouldn't be impossible to do manually.

dylan-conway:
### What does this PR do?

Adds support for `bun.lock`. With the following `package.json`:

```json
{
  "name": "pkg",
  "version": "2.2.2",
  "dependencies": {
    "is-number": "1.0.0",
    "is-odd": "^3.0.1"
  }
}
```

A JSON with trailing commas `bun.lock` file is created:

```json
{
  "lockfileVersion": 0,
  "workspaces": {
    "": {
      "dependencies": {
        "is-number": "1.0.0",
        "is-odd": "^3.0.1",
      }
    }
  },
  "packages": {
    "is-number": ["is-number@1.0.0", "", {}, "sha512-chlxkgJp4PZIiff6kUe/MWLp5+soELWNYA2IsOTus1YwKj8d9JZS6QsU7Ryqwhb1f4i0nQ5SsoUj6d5kGgq0KQ=="],
    "is-odd": ["is-odd@3.0.1", "", { "dependencies": { "is-number": "^6.0.0" } }, "sha512-CQpnWPrDwmP1+SMHXZhtLtJv90yiyVfluGsX5iNCVkrhQtU3TQHsUWPG9wkdk9Lgd5yNpAg9jQEo90CBaXgWMA=="],
    "is-odd/is-number": ["is-number@6.0.0", "", {}, "sha512-Wu1VHeILBK8KAWJUAiSZQX94GmOE45Rg6/538fKwiloUu21KncEkYGPqob2oSZ5mUT73vLGrHQjKw3KMPwfDzg=="],
  }
}
```

closes #11863
closes #5486 

Currently, if `bun.lockb` exists and `bun.lock` does not, or no lockfile exists, bun will still use the binary format. Running `bun install --save-text-lockfile` will produce `bun.lock`. Bun will choose `bun.lock` over `bun.lockb` if both exist.

Testing is ongoing. Currently there is code for converting `bun.lockb` to `bun.lock` in memory then re-parsing it before installing.


robobun:
@dylan-conway, your commit e36a455aead174265af82afd02519a3ad478f67c is building: [`#7778`](https://buildkite.com/bun/bun/builds/7778)

benpsnyder:
Failing `bun install --frozen-lockfile` after switching

Gobd:
Also failing frozen install. Also the file not valid JSON but maybe it's not intended to be. If it is supposed to be JSON it'd be nice if the extension was json. Has issues like ` ",eslint-config-prettier ",` that shouldn't have that trailing comma, and the `",` and trailing space are weird. All of the JSON issues for me appear in the `optionalPeers` section of the file.

tuyen-at-work:
In real project where each installed package have a long list of dependencies, the single line of dependencies approach would make it hard for versioning, as when we update the packages there will be a lot of change in single line and it'll feel clutered for reviewing.

I suggest use multiple line, or adopt or format like YAML so we can achieve both consise and versioning-friendly purpose.

Sherefedinagri:
Me

PR #15716, opened 2024-12-11
test(events): 66% -> 94%

pfgithub:
not done:
- test-events-add-abort-listener.mjs: waiting on node:test

partial:
- test-events-once.js: one test skipped because bun's on() doesn't resist stopPropagation. node
  implements this using kResistStopPropagation.
- tests-events-customevent.js: one test skipped because it expects `util.inspect(new CustomEvent('$'), {depth: -1})` to return `CustomEvent` but it returns `[CustomEvent]` instead

done:
- test-events-on-async-iterator.js: one test skipped because it tests the internal NodeEventTarget
- test-events-static-geteventlistener.js: test skipped because it tests the internal kWeakHandler
- test-events-listener-count.js


robobun:
@pfgithub, your commit 09b0584d9a88933ee9e8008f6de2b28c0f177ed4 is building: [`#7823`](https://buildkite.com/bun/bun/builds/7823)

src/bun.js/bindings/webcore/JSCustomEvent.cpp
```diff
@@ -47,6 +47,7 @@
 #include <wtf/GetPtr.h>
 #include <wtf/PointerPreparations.h>
 #include <wtf/URL.h>
+#include "../ErrorCode.h"
```
Jarred-Sumner, line 50: the ../ is unnecessary I think

PR #15720, opened 2024-12-12
ci: repair lint setup and run it in CI

DonIsaac:
### What does this PR do?
- Replace `eslint` with `oxlint`, which was not configured correctly. Before this PR, `bun lint` simply crashed.
- Configure `oxlint` to only look for a set of strictly incorrect code
- Add a lint job to CI. While `bun lint` will display warnings locally, this job will not.
- Fix error-level lint rule violations found. These are strictly (minor) bug fixes.

Build + test locally


robobun:
Your commit 9f5b6f256288d78825d83ee699dc03283907f7fd is building: [`#7841`](https://buildkite.com/bun/bun/builds/7841)

Jarred-Sumner:
I think we're going to need to be very specific of what files have which linter rules apply. We necessarily do weird things in various tests. Things like `ReferenceError` should definitely be caught, but most stylistic opinions should be enforced by formatter and not by linter

DonIsaac:
@Jarred-Sumner I've greatly narrowed the scope of files subject to linting. You'll find the full list in the `ignorePatterns` property in `oxlint.json`. I'm also only checking for correctness issues. I've already found a precedence bug in `node:stream`.

test/js/bun/ffi/ffi.test.js
```diff
@@ -927,14 +927,6 @@ const libSymbols = {
     returns: "int",
     args: ["ptr", "ptr", "usize"],
   },
-  pthread_attr_getguardsize: {
```
Electroid, line 930: ?

DonIsaac: Duplicate property. It's declared above (line 866)

test/snippets/react-context-value-func.tsx
```diff
@@ -10,7 +10,7 @@ const ContextProvider = ({ children }) => {
   return <Context.Provider value={cb}>{children(foo)}</Context.Provider>;
 };
```
Jarred-Sumner, line 12: this snippets folder shouldn't be linted. We don't actively run it in CI, we could potentially delete it but it was previously "weird code" that was useful to run sometimes

Issue #7854, opened 2023-12-26
Wrong stacktrace when using assert


ArnaudBarre:
### What version of Bun is running?

1.0.20+09d51486e

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

Run
```
import assert from "node:assert";

if (true) {
  assert(false);
}
```

### What is the expected behavior?

Error pointer to line 4

### What do you see instead?

Error point to line 3

Jarred-Sumner:
our sourcemaps are not in a good place right now

it's on our TODO list to address

paperdave:
This will be resolved in Bun 1.1.13

PR #15722, opened 2024-12-12
fix the rest of node:buffer

nektro:
missing some cosmetic tests but all the functionality is there now


robobun:
@nektro, your commit 3b3c29e0d26e51a561af4662f484aeadee2699d0 is building: [`#7854`](https://buildkite.com/bun/bun/builds/7854)

test/js/node/buffer.test.js
```diff
@@ -2188,7 +2188,7 @@ for (let withOverridenBufferWrite of [false, true]) {
           // If byteOffset is not numeric, it defaults to 0.
           const ab = new ArrayBuffer(10);
           const expected = Buffer.from(ab, 0);
-          expect(Buffer.from(ab, "fhqwhgads")).toStrictEqual(expected);
+          // expect(Buffer.from(ab, "fhqwhgads")).toStrictEqual(expected);
```
Jarred-Sumner: was this commenting out intentional

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2339,6 +2373,127 @@ JSC::JSObject* createBufferConstructor(JSC::VM& vm, JSC::JSGlobalObject* globalO
...
+    }
+
+    RELEASE_AND_RETURN(throwScope, JSC::JSValue::encode(uint8Array));
+}
+JSC_DEFINE_HOST_FUNCTION(jsFunction_BufferFrom_Size, (JSGlobalObject * lexicalGlobalObject, CallFrame* callFrame))
+{
+    auto& vm = lexicalGlobalObject->vm();
+    auto throwScope = DECLARE_THROW_SCOPE(vm);
+    auto lengthValue = callFrame->argument(0);
+    Bun::V::validateNumber(throwScope, lexicalGlobalObject, lengthValue, jsString(vm, String("size"_s)), jsNumber(0), jsNumber(Bun::Buffer::kMaxLength));
```
Jarred-Sumner: Please avoid creating `jsString` for code that doesn't reach JavaScript.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2339,6 +2373,127 @@ JSC::JSObject* createBufferConstructor(JSC::VM& vm, JSC::JSGlobalObject* globalO
...
+}
+EncodedJSValue constructBufferFromArray(JSC::ThrowScope& throwScope, JSGlobalObject* lexicalGlobalObject, JSValue arrayValue)
+{
+    auto* globalObject = reinterpret_cast<Zig::GlobalObject*>(lexicalGlobalObject);
+
+    auto* constructor = lexicalGlobalObject->m_typedArrayUint8.constructor(lexicalGlobalObject);
+    MarkedArgumentBuffer argsBuffer;
+    argsBuffer.append(arrayValue);
+    JSValue target = globalObject->JSBufferConstructor();
+    auto* object = JSC::construct(lexicalGlobalObject, constructor, target, argsBuffer, "Buffer failed to construct"_s);
```
Jarred-Sumner, line 2436: I wish we could avoid this - it adds ~30ns of overhead just using JSC::construct.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1794,93 +1856,64 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunctionWriteWithEncoding(JSC
...
-    if (UNLIKELY(!offsetValue.isNumber())) {
-        throwTypeError(lexicalGlobalObject, scope, "Invalid offset"_s);
-        return {};
+        if (lengthValue.isUndefined()) {
+            lengthValue = jsNumber(remaining);
+        } else if (lengthValue.isString()) {
+            encodingValue = lengthValue;
+            lengthValue = jsNumber(remaining);
+        } else {
+            validateOffset(scope, lexicalGlobalObject, lengthValue, jsString(vm, String("length"_s)), jsNumber(0), jsNumber(length));
```
Jarred-Sumner: Please avoid creating `jsString` for code that doesn't reach JavaScript.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1794,93 +1856,64 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunctionWriteWithEncoding(JSC
...
-
     if (lengthValue.isUndefined() && offsetValue.isString()) {
-        // https://github.com/nodejs/node/blob/e676942f814915b2d24fc899bb42dc71ae6c8226/lib/buffer.js#L1056
         encodingValue = offsetValue;
-        setEncoding();
+        lengthValue = jsNumber(castedThis->byteLength());
+        offsetValue = jsNumber(0);
+    } else {
+        uint32_t length = castedThis->byteLength();
+        validateOffset(scope, lexicalGlobalObject, offsetValue, jsString(vm, String("offset"_s)), jsNumber(0), jsNumber(length));
```
Jarred-Sumner: Here as well.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1794,93 +1856,64 @@ static inline JSC::EncodedJSValue jsBufferPrototypeFunctionWriteWithEncoding(JSC
...
-    };
+    auto stringValue = callFrame->argument(0);
+    auto offsetValue = callFrame->argument(1);
+    auto lengthValue = callFrame->argument(2);
+    auto encodingValue = callFrame->argument(3);
 
     if (offsetValue.isUndefined()) {
-        // https://github.com/nodejs/node/blob/e676942f814915b2d24fc899bb42dc71ae6c8226/lib/buffer.js#L1053
-        RELEASE_AND_RETURN(scope, writeToBuffer(lexicalGlobalObject, castedThis, str, offset, length, encoding));
+        Bun::V::validateString(scope, lexicalGlobalObject, stringValue, "string"_s);
```
Jarred-Sumner, line 1911: These validator functions would be better if they returned the intended type instead of validating them and then coercing to them after. It would both be more performant and less likely to cause bugs.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1311,108 +1346,130 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
+            const void* offset = memchr(reinterpret_cast<const void*>(typedVector + byteOffset), byteValue, byteLength - byteOffset);
+            if (offset != NULL) return static_cast<const uint8_t*>(offset) - typedVector;
         }
+        return -1;
     }
 
-    if (value.isString()) {
-        auto* str = value.toStringOrNull(lexicalGlobalObject);
-        RETURN_IF_EXCEPTION(scope, -1);
+    JSString* encodingValue_jstr;
```
Jarred-Sumner: This is a footgun - C++ pointers/primitives are uninitialized, and this JSValue isn't accessed other than to get the WTF::String. Therefore, we only need to keep the WTF::String alive.
```suggestion

```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1311,108 +1346,130 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
+    WTF::String encodingValue_wstr;
+    if (!encodingValue.isUndefined()) {
+        encodingValue_jstr = encodingValue.toStringOrNull(lexicalGlobalObject);
```
```suggestion
        encodingString = encodingValue.toWTFString(lexicalGlobalObject);
```

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -454,6 +459,13 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_validateBuffer, (JSC::JSGlobalObject * globa
 
     auto buffer = callFrame->argument(0);
     auto name = callFrame->argument(1);
+    return V::validateBuffer(scope, globalObject, buffer, name);
+}
+JSC::EncodedJSValue V::validateBuffer(JSC::ThrowScope& scope, JSC::JSGlobalObject* globalObject, JSValue buffer, JSValue name)
+{
+    JSC::VM& vm = globalObject->vm();
+
+    if (name.isUndefined()) name = jsString(vm, String("buffer"_s));
```
Jarred-Sumner: Please avoid creating JSString for code that doesn't reach JavaScript
```suggestion

```

src/js/builtins/JSBufferConstructor.ts
```diff
@@ -1,78 +1,37 @@
...
+    if (valueOf != null && valueOf !== value && (typeof valueOf === "string" || typeof valueOf === "object")) {
+      return Buffer.from(valueOf, encodingOrOffset, length);
+    }
 
-    if (toPrimitive) {
-      const primitive = toPrimitive.$call(items, "string");
+    const b = fromObject(value);
+    if (b) return b;
 
+    const toPrimitive = $tryGetByIdWithWellKnownSymbol(value, "toPrimitive");
```
Jarred-Sumner, line 21: You want to do this instead of `value + ""`?

Jarred-Sumner (changes requested):
Some comments but getting close

Jarred-Sumner (changes requested):
There's a lot of unnecessary work here. Bun doesn't stay fast by doing unnecessary work.

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -51,6 +51,24 @@ JSC::EncodedJSValue V::validateInteger(JSC::ThrowScope& scope, JSC::JSGlobalObje
 
     return JSValue::encode(jsUndefined());
 }
+JSC::EncodedJSValue V::validateInteger(JSC::ThrowScope& scope, JSC::JSGlobalObject* globalObject, JSC::JSValue value, const WTF::String& name, JSC::JSValue min, JSC::JSValue max)
+{
+    if (!value.isNumber()) return Bun::ERR::INVALID_ARG_TYPE(scope, globalObject, name, "number"_s, value);
+    if (min.isUndefined()) min = jsDoubleNumber(JSC::minSafeInteger());
```
Jarred-Sumner: When these values are coming from C++ (not JavaScript), we shouldn't be converting from C++ -> JS types -> back.

Especially when it's compile-time known.

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -297,6 +335,30 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_validateArray, (JSC::JSGlobalObject * global
...
+}
+JSC::EncodedJSValue V::validateArray(JSC::ThrowScope& scope, JSC::JSGlobalObject* globalObject, JSValue value, JSValue name, JSValue minLength)
+{
+    JSC::VM& vm = globalObject->vm();
+
+    if (minLength.isUndefined()) minLength = jsNumber(0);
+
+    if (!JSC::isArray(globalObject, value)) return Bun::ERR::INVALID_ARG_TYPE(scope, globalObject, name, "Array"_s, value);
+
+    auto length = value.get(globalObject, Identifier::fromString(vm, "length"_s));
```
Jarred-Sumner: This getter call, and the length coercion below is unnecessary and likely will cause bugs.

Plus, the identifier `"length"` is in the special list of already available properties in the JSC::VM. `vm.propertyNames.length`.

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -297,6 +335,30 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_validateArray, (JSC::JSGlobalObject * global
...
     auto minLength = callFrame->argument(2);
+    return V::validateArray(scope, globalObject, value, name, minLength);
+}
+JSC::EncodedJSValue V::validateArray(JSC::ThrowScope& scope, JSC::JSGlobalObject* globalObject, JSValue value, JSValue name, JSValue minLength)
+{
+    JSC::VM& vm = globalObject->vm();
+
+    if (minLength.isUndefined()) minLength = jsNumber(0);
+
+    if (!JSC::isArray(globalObject, value)) return Bun::ERR::INVALID_ARG_TYPE(scope, globalObject, name, "Array"_s, value);
```
Jarred-Sumner: Instead of using `JSC::isArray`, and then otherwise treating it as a normal object, we can cast it to a `JSArray*` and then call `length` on that `JSArray`. It is stored as an unsigned integer.


```suggestion
   auto *array = jsDynamicCast<JSC::JSArray*>(value);
  if (!array)
  return Bun::ERR::INVALID_ARG_TYPE(scope, globalObject, name, "Array"_s, value);
```

Jarred-Sumner:
This PR makes Buffer-related methods 3x - 5x slower. We are not merging this until that performance regression is fixed.

image

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -201,14 +204,14 @@ WTF::String JSValueToStringSafe(JSC::JSGlobalObject* globalObject, JSValue arg)
         auto name = JSC::getCalculatedDisplayName(vm, cell->getObject());
         if (catchScope.exception()) {
             catchScope.clearException();
-            name = "Function"_s;
+            name = ""_s;
```
cirospaciari:
```suggestion
         return "[Function: (anonymous)]"_s;
```

cirospaciari: I assuming that is fine to ignore this exception

cirospaciari (changes requested):

nektro:
this project is now complete

PR #15740, opened 2024-12-13
feat(s3) s3 client 

cirospaciari:
### What does this PR do?

- [x] Code changes

Tests


src/env_loader.zig
```diff
@@ -112,6 +115,40 @@ pub const Loader = struct {
...
+    pub fn getAWSCredentials(this: *Loader) s3.AWSCredentials {
+        if (this.aws_credentials) |credentials| {
+            return credentials;
+        }
+
+        var accessKeyId: []const u8 = "";
+        var secretAccessKey: []const u8 = "";
+        var region: []const u8 = "";
+
+        if (this.get("AWS_ACCESS_KEY_ID")) |access_key| {
```
Electroid: `S3` should take priority over `AWS`

src/env_loader.zig
```diff
@@ -112,6 +115,40 @@ pub const Loader = struct {
         }
     }
 
+    pub fn getAWSCredentials(this: *Loader) s3.AWSCredentials {
+        if (this.aws_credentials) |credentials| {
+            return credentials;
+        }
+
+        var accessKeyId: []const u8 = "";
```
Electroid, line 123: Also needs `{S3/AWS}_ENDPOINT` to allow for non-AWS buckets like R2

src/bun.js/webcore/blob.zig
```diff
@@ -3423,12 +3431,192 @@ pub const Blob = struct {
...
+            this.destroy();
+        }
+    };
+
+    pub fn getPresignUrl(this: *Blob, globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) bun.JSError!JSValue {
+        if (this.isS3()) {
+            const args = callframe.arguments_old(1);
+            var method: bun.http.Method = .GET;
+            var expires: usize = 86400; // 1 day
+            if (args.len == 1) {
```
Jarred-Sumner:
```suggestion
            if (args.len > 0) {
```

src/bun.js/webcore/blob.zig
```diff
@@ -3423,12 +3431,192 @@ pub const Blob = struct {
...
+                const options = args.ptr[0];
+                if (options.isObject()) {
+                    if (try options.getTruthyComptime(globalThis, "method")) |method_| {
+                        method = Method.fromJS(globalThis, method_) orelse {
+                            return globalThis.throwInvalidArguments("method must be GET, PUT, DELETE or HEAD when using s3 protocol", .{});
+                        };
+                    }
+                    if (try options.getTruthyComptime(globalThis, "expiresIn")) |expires_| {
```
Jarred-Sumner:
```suggestion
                    if (try options.getOptional(i32, globalThis, "expiresIn")) |expires_| {
```

src/bun.js/webcore/blob.zig
```diff
@@ -3423,12 +3431,192 @@ pub const Blob = struct {
         return JSValue.jsBoolean(bun.isRegularFile(store.data.file.mode) or bun.C.S.ISFIFO(store.data.file.mode));
     }
 
+    fn isS3(this: *Blob) bool {
+        if (this.store) |store| {
+            if (store.data == .file) {
```
Jarred-Sumner: Maybe we could add a `s3` or `remote` to the Store instead?

src/s3.zig
```diff
@@ -0,0 +1,572 @@
...
+                day,
+                hours,
+                minutes,
+                seconds,
+            }) catch bun.outOfMemory(),
+        };
+    }
+
+    const DIGESTED_HMAC_256_LEN = 32;
+    threadlocal var SIGNATURE_CACHE: bun.StringArrayHashMap([DIGESTED_HMAC_256_LEN]u8) = undefined;
```
Jarred-Sumner: this will leak memory  and the key will be invalid, we can probably get away with not doing this

src/s3.zig
```diff
@@ -0,0 +1,572 @@
...
+            .headers = .{
+                .{ .name = "x-amz-content-sha256", .value = aws_content_hash },
+                .{ .name = "x-amz-date", .value = amz_date },
+                .{ .name = "Authorization", .value = authorization[0..] },
+                .{ .name = "Host", .value = host },
+            },
+        };
+    }
+
+    pub const S3StatResult = union(enum) {
```
Jarred-Sumner, line 591: feels like we should have a generic for this like OptionalResult

src/url.zig
```diff
@@ -105,6 +109,12 @@ pub const URL = struct {
         return "localhost";
     }
 
+    pub fn s3Path(this: *const URL) string {
+        // we need to remove protocol if exists and ignore searchParams, should be host + pathname
+        const href = if (this.protocol.len > 0) this.href[this.protocol.len + 2 ..] else this.href;
```
Jarred-Sumner:
```suggestion
        const href = if (this.protocol.len > 0 and this.href.len > this.protocol.len + 2) this.href[this.protocol.len + 2 ..] else this.href;
```

src/s3.zig
cirospaciari: I still think we need signature cache but we can do it later in a better way

src/bun.js/webcore/blob.zig
cirospaciari: i32 was not implemented in getOptional added one but is basically the same thing I was doing before.

src/bun.js/webcore/blob.zig
cirospaciari: `stora.data== .s3` is what I was planing for today, because we also need to store the options passed in the constructor.

src/s3.zig
cirospaciari: not sure about that, failure is present on each result but not_found don't fit uploads.

src/bun.js/rare_data.zig
```diff
@@ -48,7 +48,58 @@ listening_sockets_for_watch_mode_lock: bun.Lock = .{},
 
 temp_pipe_read_buffer: ?*PipeReadBuffer = null,
 
+aws_signature_cache: AWSSignatureCache = .{},
+
 const PipeReadBuffer = [256 * 1024]u8;
+const DIGESTED_HMAC_256_LEN = 32;
+pub const AWSSignatureCache = struct {
+    cache: bun.StringArrayHashMap([DIGESTED_HMAC_256_LEN]u8) = undefined,
```
Jarred-Sumner:
```suggestion
    cache: bun.StringArrayHashMap([DIGESTED_HMAC_256_LEN]u8) = bun.StringArrayHashMap([DIGESTED_HMAC_256_LEN]u8).init(bun.default_allocator),
```

src/bun.js/webcore/response.zig
```diff
@@ -911,6 +955,13 @@ pub const Fetch = struct {
                 };
             }
 
+            pub fn isS3(this: *HTTPRequestBody) bool {
```
Jarred-Sumner:
```suggestion
            pub fn isS3(this: *const HTTPRequestBody) bool {
```

src/bun.js/webcore/response.zig
```diff
@@ -911,6 +955,13 @@ pub const Fetch = struct {
...
+                return switch (this.*) {
+                    .AnyBlob => |blob| blob.isS3(),
```
Jarred-Sumner:
```suggestion
                    .AnyBlob => |*blob| blob.isS3(),
```

src/bun.js/webcore/response.zig
```diff
@@ -3141,36 +3207,228 @@ pub const Fetch = struct {
...
-                        .offset = body.AnyBlob.Blob.offset,
-                        .max_size = body.AnyBlob.Blob.size,
-                    },
-                    .sync,
-                );
+                // is a file we can use chunked here
 
-                if (body.store().?.data.file.pathlike == .path) {
-                    _ = bun.sys.close(opened_fd);
+                if (JSC.WebCore.ReadableStream.fromJS(JSC.WebCore.ReadableStream.fromBlob(globalThis, &body.AnyBlob.Blob, s3.MultiPartUpload.DefaultPartSize), globalThis)) |stream| {
```
Jarred-Sumner: I don't think we can call this on the JS value twice?

src/bun.js/webcore/response.zig
```diff
@@ -3141,36 +3207,228 @@ pub const Fetch = struct {
...
+                method = .PUT;
+            }
+
+            var result = credentialsWithOptions.credentials.signRequest(.{
+                .path = url.s3Path(),
+                .method = method,
+            }, null) catch |sign_err| {
+                switch (sign_err) {
+                    error.MissingCredentials => {
+                        const err = JSC.toTypeError(.ERR_INVALID_ARG_VALUE, "missing s3 credentials", .{}, ctx);
```
Jarred-Sumner: Can we make each of these a unique error code? Can we avoid JSC.toTypeError and instead use globalObject.ERR_


```suggestion
                        const err = globalObject.ERR_S3_MISSING_CREDENTIALS("Missing S3 credentials", .{}).throw();
```

src/bun.js/webcore/response.zig
```diff
@@ -3141,36 +3207,228 @@ pub const Fetch = struct {
...
-                        body.detach();
-                        body.AnyBlob.from(std.ArrayList(u8).fromOwnedSlice(allocator, @constCast(result.slice())));
-                        http_body = .{ .AnyBlob = body.AnyBlob };
+                    error.InvalidMethod => {
+                        const err = JSC.toTypeError(.ERR_INVALID_ARG_VALUE, "method must be GET, PUT, DELETE or HEAD when using s3 protocol", .{}, ctx);
+                        is_error = true;
+                        return JSPromise.rejectedPromiseValue(globalThis, err);
+                    },
+                    error.InvalidPath => {
+                        const err = JSC.toTypeError(.ERR_INVALID_ARG_VALUE, "invalid s3 bucket, key combination", .{}, ctx);
```
Jarred-Sumner:
```suggestion
                        const err = globalObject.ERR_S3_INVALID_BUCKET("The S3 bucket & key makes an invalid URL", .{}).throw();
```

src/s3.zig
```diff
@@ -0,0 +1,2110 @@
...
+        };
+
+        var result = this.signRequest(.{
+            .path = path,
+            .method = .GET,
+        }, null) catch |sign_err| {
+            if (range) |range_| bun.default_allocator.free(range_);
+
+            return switch (sign_err) {
+                error.MissingCredentials => callback(.{ .allocator = bun.default_allocator, .list = .{} }, false, .{ .code = "MissingCredentials", .message = "missing s3 credentials" }, callback_context),
```
Jarred-Sumner: please use some wrapper struct that generates consistent messages from the same error codes instead of duplicating errors and messages in multiple places.

src/s3.zig
```diff
@@ -0,0 +1,2110 @@
...
+                        } else {
+                            readable.ptr.Bytes.onData(
+                                .{
+                                    .temporary_and_done = bun.ByteList.initConst(chunk.list.items),
+                                },
+                                bun.default_allocator,
+                            );
+                        }
+                    }
+                } else {
```
Jarred-Sumner: can we instead make the scope above return, and then only put this code in one place instead of two?

test/js/bun/s3/s3.test.ts
```diff
@@ -0,0 +1,443 @@
...
+const s3Options: S3FileOptions = {
+  accessKeyId: getSecret("S3_R2_ACCESS_KEY"),
+  secretAccessKey: getSecret("S3_R2_SECRET_KEY"),
+  endpoint: getSecret("S3_R2_ENDPOINT"),
+};
+
+const S3Bucket = getSecret("S3_R2_BUCKET");
+
+// 15 MiB big enough to Multipart upload in more than one part
+const bigPayload = Buffer.alloc(15 * 1024 * 1024, "a");
```
Jarred-Sumner: this should be a longer repeating pattern so that we can verify the data is correct. maybe the contents of this file itself would be a good way to do that.

src/s3.zig
```diff
@@ -0,0 +1,2110 @@
...
+        // explicitly set it to a dead pointer
+        // we use this memory address to disable signals being sent
+        signal.clear();
+        bun.assert(signal.isDead());
+
+        return response_stream.sink.toJS(globalThis);
+    }
+};
+
+pub const MultiPartUpload = struct {
```
Jarred-Sumner, line 1673: These should have more debug logs

src/s3.zig
```diff
@@ -0,0 +1,2110 @@
...
+    pub const OneMiB: usize = 1048576;
+    pub const MAX_SINGLE_UPLOAD_SIZE_IN_MiB: usize = 5120; // we limit to 5 GiB
+    pub const MAX_SINGLE_UPLOAD_SIZE: usize = MAX_SINGLE_UPLOAD_SIZE_IN_MiB * OneMiB; // we limit to 5 GiB
+    pub const MIN_SINGLE_UPLOAD_SIZE_IN_MiB: usize = 5;
+    pub const DefaultPartSize = OneMiB * MIN_SINGLE_UPLOAD_SIZE_IN_MiB;
+    const MAX_QUEUE_SIZE = 64; // dont make sense more than this because we use fetch anything greater will be 64
+    const AWS = AWSCredentials;
+    queue: std.ArrayListUnmanaged(UploadPart) = .{},
+
+    available: std.bit_set.IntegerBitSet(MAX_QUEUE_SIZE) = std.bit_set.IntegerBitSet(MAX_QUEUE_SIZE).initFull(),
```
Jarred-Sumner:
```suggestion

    available: bun.bit_set.IntegerBitSet(MAX_QUEUE_SIZE) = bun.bit_set.IntegerBitSet(MAX_QUEUE_SIZE).initFull(),
```

src/s3.zig
```diff
@@ -0,0 +1,2110 @@
...
+
+    pub const MultiPartUploadOptions = struct {
+        queueSize: u8 = 5, // more than 255 dont make sense http thread cannot handle more than that
+        // in s3 client sdk they set it in bytes but the min is still 5 MiB
+        // var params = {Bucket: 'bucket', Key: 'key', Body: stream};
+        // var options = {partSize: 10 * 1024 * 1024, queueSize: 1};
+        // s3.upload(params, options, function(err, data) {
+        //   console.log(err, data);
+        // });
+        // See. https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property
```
Jarred-Sumner: these should be `///` doc comments

PR #15751, opened 2024-12-13
Fix flaky signal handlers on posix

Jarred-Sumner:
### What does this PR do?

Fix flaky signal handlers on posix. We cannot allocate memory inside of a signal handler. 

The approach here uses a fixed-length ring buffer

There is one test


robobun:
@Jarred-Sumner, your commit 6cef3dc5d841d827063623f42c02053f26d24e06 is building: [`#7946`](https://buildkite.com/bun/bun/builds/7946)

Issue #7956, opened 2024-01-03
Support offline installs via caching dependency tarballs


zachbryant:
### What is the problem this feature would solve?

My CI cannot make external requests. Using yarn, we have an offline cache folder that contains `.tgz` files ready to install without an internet connection.

### What is the feature you are proposing to solve the problem?

Would love to see a configuration option under `install.cache` like maybe `offlineDir` where I could find the tarballs.

### What alternatives have you considered?

This feature is something I can't migrate without. Uploading the global cache is also too slow for any of our pipelines.

zachbryant:
This is something I'm also open to contributing if it's not a high priority on the roadmap

tylergets:
This would be great because it would allow a nice integration with Nix

antonk52:
Another +1. [yarn offline mirror](https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/) is an anchor that makes it impossible to migrate away if there is not internet access in the CI. it would be incredible in bun had support for offline mirror or a similar feature and would allow companies with stricter security requirements to adapt bun.

nyabinary:
> This is something I'm also open to contributing if it's not a high priority on the roadmap

It would be very helpful :3 considering we need it for nixpkgs tooling.

Eveeifyeve:
I might look into how this is possible.

CC @Jarred-Sumner do you know how this could be possible?

nyabinary:
> I might look into how this is possible.
> 
> CC [@Jarred-Sumner](https://github.com/Jarred-Sumner) do you know how this could be possible?

Any updates?

Eveeifyeve:
> > I might look into how this is possible.
> > CC [@Jarred-Sumner](https://github.com/Jarred-Sumner) do you know how this could be possible?
> 
> Any updates?

No unfortunately, since I still don't know how the implementation would work.

nyabinary:
> > > I might look into how this is possible.
> > > CC [@Jarred-Sumner](https://github.com/Jarred-Sumner) do you know how this could be possible?
> > 
> > 
> > Any updates?
> 
> No unfortunately, since I still don't know how the implementation would work.

Could try joining the discord and try to get support there?

bumping: @Jarred-Sumner

Glandos:
I've tried a curious workaround. It uses the following bunfig.toml:
```toml
[install.cache]
# when true, don't load from the global cache.
# Bun may still write to node_modules/.cache
disable = true
```
[...]

PR #15754, opened 2024-12-14
document `npm:<package-name>` in install docs

RiskyMH:
fixes https://github.com/oven-sh/bun/discussions/14938

technically its not "non-npm dependencies", however it is documenting the other ways then normal, so i think it fits and shows capability

(the actual example i chose i have no idea how much it could break though as @types/bun requires bun-types)


robobun:
@RiskyMH, your commit f30d5448ddee3c310e2f87e6a79fd97ef22f7128 is building: [`#7956`](https://buildkite.com/bun/bun/builds/7956)

Issue #7963, opened 2024-01-03
`bun create` doesn't recognize the registry setting in `bunfig.toml`


uiolee:
### What version of Bun is running?

1.0.21

### What platform is your computer?

Microsoft Windows NT 10.0.19044.0 x64 （WSL Debian）

### What steps can reproduce the bug?

1. cd 'project/'
1. set custom/private registry in `project/bunfig.toml`. https://bun.sh/docs/install/registries
2. run `bun create myCustomApp`

### What is the expected behavior?

`bun` should download my `create-myCustomApp` package from my custom/private registry.

### What do you see instead?

`bun` download package from npmjs.org

```log
error: package "create-myCustomApp" not found registry.npmjs.org/create-myCustomApp 404
```

### Additional information

```bash
$ debian@win10-wsl:~/project$ cat bunfig.toml
[install]
registry = "http://private.registry.com/"
```

Electroid:
Duplicate of #617

PR #15756, opened 2024-12-14
Don't call _kill when its not observable

Jarred-Sumner:
### What does this PR do?

Don't call _kill when its not observable. This skips a getter and call into JS until the PropertyCallback is called.

Existing tests


robobun:
@Jarred-Sumner, your commit 76f40520b00d6447d10c1e259eab19c754cc230b is building: [`#7963`](https://buildkite.com/bun/bun/builds/7963)

Jarred-Sumner:
This needs to apply on the setter and not just Object.defineProperty or the getter.

Issue #7384, opened 2023-11-30
export not found when executing typescript file


MRDGH2821:
### What version of Bun is running?

1.0.14+d8be3e51b

### What platform is your computer?

Linux 5.15.133.1-microsoft-standard-WSL2 x86_64 unknown 

### What steps can reproduce the bug?

1. `git clone https://github.com/MRDGH2821/Discord-Ban-Utils-Bot` inside WSL.
2. Switch to `rewrite-sapphire` branch
3. Open the repo inside dev container
4. `bun src/index.ts`

### What is the expected behavior?

Code should run successfully

### What do you see instead?

```
$ bun src/index.ts
56 |         }
57 |       }
58 |       return new FixedQueue;
59 |     }();
60 | 
61 |     function processTicksAndRejections() {
                                                                               ^
SyntaxError: export 'ValueOf' not found in './EventTypes'
      at processTicksAndRejections (:61:76)
```      

`EventTypes.ts` is located in `src/lib`
And `ValueOf` is defined in `src/lib/EventTypes.ts`, re-exported in `src/lib/utils.ts`

### Additional information

My bot requires the following to start:
1. Discord bot token
2. Firebase Cloud Firestore service account key file

And, I'm using Dev container inside WSL

RobertP09:
I'm seeing this too. I have this import inside the the file where the error is occurring but bun doesn't recognize it..
Error:
`SyntaxError: Export named 'Status' not found in module ***l/src/data/types.ts'.
      at processTicksAndRejections (:61:76)`
      
```
import {
  STATUS,
} from '@src/types';
```

fdaciuk:
Same problem here with version `1.0.20`:
![image](https://github.com/oven-sh/bun/assets/487669/a55b3ed7-24cc-40e5-883c-9aaad039c740)

I have a `Context` type exported from this file:
![image](https://github.com/oven-sh/bun/assets/487669/4e8f6320-6e40-469b-affb-a6d01d5d3658)

nikhilro:
+1 

```
export const OPENAI_MODEL_NAMES = ['gpt-4', 'gpt-3.5-turbo'] as const;
export type OpenAIModelName = (typeof OPENAI_MODEL_NAMES)[number];
```

```
SyntaxError: Export named 'OpenAIModelName' not found in module '/Users/nikhil/github/vapi/backend/types/openAITypes.ts'.
      at link (:1:21)
      at link (:1:21)
      at link (:1:21)
[...]

PR #15758, opened 2024-12-14
Skip javascriptcore's first parse step for ES Modules

pfgithub:
First half of #7384. Does not fix that problem, but gets closer to fixing it.

This pr:

- Adds a test to prepare for when 7384 is fixed (marked with {todo: true} (#15633)
- Constructs ModuleInfo when printing with all the information needed to create a JavaScriptCore JSModuleRecord
- Replaces javascriptcore's first parsing step to generate the module record with our own generated module record
- Adds an env var `COMPILE_ERRORS_ONLY=1 bun run build` for building bun that skips codegen. This shows errors ~50% faster and can often reveal an error that was crashing zig previously

TODO:

- [ ] Make sure memory is freed in the case where the module is discarded before it is loaded?

Check:
- [ ] Make sure cached ModuleInfo in RuntimeTranspilerCache is used
- [ ] Test all the types of module imports and exports and re-exports and star exports and make sure they still work
- [ ] Consider whether to run the module_info building as a seperate pass at the end of printAst rather than integrating it into js_printer
- [ ] Consider using Ast.import records, export records, scope, ... to find the needed items rather than finding them in printer. There are a few special cases like the jest thing (but maybe could be solved by moving that transformation to a step before printing?)
- [ ] Does setFromUtf8 need to be used? What happens if it is called on a bad string containing non-utf8 bytes? Is this unsafe? Is it faster to use a setFromAscii fn when possible?
- [ ] remove the `eqlComptime("*default*")` check somehow (either just remove it and it's fine or add special indexes for those)
- [ ] Is ModuleInfo information needed in standalone module graph? see: `module_loader.zig` line 2633

Performance impact:

- [x] large file, with runtime transpiler cache: ~15% ± 1% faster (node is ~2x faster than bun at this)
- [x] large file, without runtime transpiler cache: ~5% ± 1% faster
- [x] large file with only one import: 12% faster with runtime transpiler cache, 5% faster without
- [x] small file, with imports: ~1% ± 2% slower
- [x] small file, without imports: ~1% ± 2% slower
- [ ] (todo) real-world imports

<details><summary>Benchmark</summary>

```js
// genlarge.js
let large_export_top = "";
let large_export_bottom = "export {\n";
let large_import_top = "import {\n";
let large_import_bottom = `function expect(a, b, c) {
    if(b !== a || c !== b) throw new Error("failure");
}\n`;

for(let i = 0; i < 30_000; i++) {
    large_export_top += `export const _${i} = ${i};\n`;
    large_export_bottom += "    _"+i+" as _"+i+"_d,\n";
    large_import_top += `    _${i}, _${i}_d,\n`;
    large_import_bottom += `expect(${i}, _${i}, _${i}_d);\n`;
}
large_export_bottom += "};\n";
large_import_top += "} from './large2.js';\n"

await Bun.write("large2.js", large_export_top + large_export_bottom);
await Bun.write("large.js", large_import_top + large_import_bottom);
```

```js
// small.js
import { b } from "./small2.js";

console.log(b);
```

```js
// small2.js
export const b = "abc";
```

```js
// noimports.js
console.log("wow, no imports");
```

```js
// runbench.js
const cmds = [
    "./bun-15758",
    "./bun-main",
];
const targets = [
    "./large.js woutrtc",
    "./large.js withrtc",
    "./largeni.js woutrtc",
    "./largeni.js withrtc",
    "./small.js",
    "./noimports.js",
];
for(const target of targets) {
    console.log("\n\n\nrunning benchmark:", target);
    await Bun.spawnSync({
        cmd: ["hyperfine", ...cmds.map(cmd => cmd + " " + target), "--warmup=5", "--shell=none"],
        env: {
            BUN_RUNTIME_TRANSPILER_CACHE_PATH: target.includes("withrtc") ? process.env.PWD + "/.rtc" : "0",
        },
        stdio: ["inherit", "inherit", "inherit"],
    });
}
```

```js
// genlargeni.js
let large_export = "export const m = 25;\n";
let large_import = "import { m, n } from './largeni2.js';\n";
large_import += "console.log(m, n);\n";

for(let i = 0; i < 30_000; i++) {
    large_export += "console.log(1);\n";
    large_import += "console.log(1);\n";
}
large_export += "export const n = 25;\n";

await Bun.write("largeni2.js", large_export);
await Bun.write("largeni.js", large_import);
```

Run benchmark:

```
bun genlarge.js
bun genlargeni.js
bun runbench.js
```

</details>


robobun:
@pfgithub, your commit c57fead7aecfee1ab02b304d07980bb20afe4f5d is building: [`#7972`](https://buildkite.com/bun/bun/builds/7972)

Issue #7984, opened 2024-01-04
Tesseract.js crashed while downloading eng.traneddata with bun index.ts


snomiao:
### What version of Bun is running?

1.0.19

### What platform is your computer?

Linux 5.15.0-91-generic x86_64 unknown

### What steps can reproduce the bug?

1. Install tesseract.js with 

`bun install tesseract.js`

2. Create a ts file with code:

```typescript
// index.ts
// just copied from from https://www.npmjs.com/package/tesseract.js
import { createWorker } from "tesseract.js";

(async () => {
  const worker = await createWorker("eng");
  const ret = await worker.recognize(
    "https://tesseract.projectnaptha.com/img/eng_bw.png"
  );
  console.log(ret.data.text);
  await worker.terminate();
})();
```

3. Run `bun index.ts`, got failed result, and program stucked ❌

```log
$ bun index.ts
239 |       rejects[action](data);
240 |       if (action === 'load') workerResReject(data);
241 |       if (errorHandler) {
242 |         errorHandler(data);
243 |       } else {
244 |         throw Error(data);
                    ^
error: TypeError: Not a string or buffer
      at /code/node_modules/tesseract.js/src/createWorker.js:244:15
      at emit (node:events:122:95)
      at #onMessage (node:worker_threads:178:25)
```

4. Run `npx tsx index.ts`,  got expected result ✅

```log
$ npx tsx index.ts
Mild Splendour of the various-vested Night!
Mother of wildly-working visions! hail
I watch thy gliding, while with watery light
Thy weak eye glimmers through a fleecy veil;
And when thou lovest thy pale orb to shroud
Behind the gather’d blackness lost on high;
And when thou dartest from the wind-rent cloud
Thy placid lightning o’er the awaken’d sky.
```

* And it downloads `eng.traineddata` into current directory

5. Run `bun index.ts`, got expected result ✅

```log
$ bun index.ts
Mild Splendour of the various-vested Night!
Mother of wildly-working visions! hail
I watch thy gliding, while with watery light
Thy weak eye glimmers through a fleecy veil;
And when thou lovest thy pale orb to shroud
Behind the gather’d blackness lost on high;
And when thou dartest from the wind-rent cloud
Thy placid lightning o’er the awaken’d sky.
```

### What is the expected behavior?

bun index.ts should success got OCR result at step 3

### What do you see instead?

`bun index.ts` crashed,

and after `npx tsx index.ts`, the `eng.traineddata` was downloaded

[...]

PR #15762, opened 2024-12-14
Don't open node_modules 1,618 times

Jarred-Sumner:
### What does this PR do?

We don't need to open relative to directories when we aren't going to hit ENAMETOOLONG


robobun:
@Jarred-Sumner, your commit 552361e378ccd139336f1682d3847f02964e4667 is building: [`#7984`](https://buildkite.com/bun/bun/builds/7984)

PR #15771, opened 2024-12-15
Make git dependencies faster + further optimize bun install

Jarred-Sumner:
### What does this PR do?

- Avoid opening directories unnecessarily when linking bins
- Avoid opening directories unnecessarily when verifying git packages
- Add an optimization for reading small files that avoids the fstat() call

Our existing tests should cover this, but I did check the syscall count for some git dependencies.


robobun:
@Jarred-Sumner, your commit 36f6044b583a1a768792b056873935f6c03e6bf7 is building: [`#7994`](https://buildkite.com/bun/bun/builds/7994)

PR #15790, opened 2024-12-16
Fix heap snapshots memory usage stats. Introduce `estimateDirectMemoryUsageOf` function in `"bun:jsc"`

Jarred-Sumner:
### What does this PR do?

The bindings generator was not wired up to report memory usage stats to the heap analyzer, leading the memory usage stats of heap snapshots to exclude the cost of native code memory.  

Note this does not impact GC / memory usage reporting to the garbage collector, as the heap analyzer is a different API.

This also exposes the very precisely-named `estimateDirectMemoryUsageOf` function, which given an object it will return the value that it would directly report for a heap snapshot.

```js
import {estimateDirectMemoryUsageOf} from 'bun:jsc';

const request = new Request("https://example.com");
console.log(estimateDirectMemoryUsageOf(request));
```

Internally:

This adds a `memoryCost` option in our bindings generator which expects a function like this:
```zig
pub fn memoryCost(this: *ThisType) usize {
    return  ...
}
```

All classes have a default implementation in the bindings generator, which is the sizeof the Zig struct.

There are tests


src/bun.js/api/bun/subprocess.zig
```diff
@@ -794,6 +802,15 @@ pub const Subprocess = struct {
         array_buffer: JSC.ArrayBuffer.Strong,
         detached: void,
 
+        pub fn memoryCost(this: *const Source) usize {
+            return switch (this.*) {
+                .blob => this.blob.memoryCost(),
+                // ArrayBuffer is owned by GC.
+                .array_buffer => 0,
```
dylan-conway, line 810: Does this still need to report the memory for the struct?

dylan-conway (commented):
Confused by the meaning of direct memory usage. Sometimes struct size is included, and sometimes it's excluded. `ArrayBufferSink` excludes the struct size and only reports bytes used for the buffer.

paperclover (approved):

docs/api/utils.md
```diff
@@ -771,3 +771,28 @@ console.log(obj); // => { foo: "bar" }
 ```
 
 Internally, [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) and [`postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) serialize and deserialize the same way. This exposes the underlying [HTML Structured Clone Algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) to JavaScript as an ArrayBuffer.
+
+## `estimateDirectMemoryUsageOf` in `bun:jsc`
```
paperclover, line 775: rename to estimateShallowMemoryUsageOf

PR #15788, opened 2024-12-16
attempt to fix debugger

RiskyMH:
2 different fixes:
1. make vscode debugger work by removing `BUN_INSPECT_CONNECT_TO` condition to not wait forever (as extension always sets it, bun doesnt wait enough usually)
2. reverts some changes in `BunDebugger.cpp` so it works in debug.bun.sh again (it seems to still work for vscode)

fixes #15439
fixes maybe #15715

PR #15801, opened 2024-12-17
vscode extension: use new debug terminal provider

### LINKED ISSUE #15223: Bun vscode extension breaks "javascript debug terminal" for node

1.1.18

Darwin 24.1.0 arm64 arm

0. Have Bun vscode extension installed
1. Open command palette (cmd+shift+p or F1) and invoke command "debug: javascript debug terminal"
2. In the newly opened javascript debug terminal, run `node` and see how debugger cannot attach to the process 

Javascript debug terminal can attach to a new node process 

### What do you see instead?

Javascript debug terminal cannot attach to a new node process 

### Additional information

This's the cause of the problem - https://github.com/oven-sh/bun/blob/d5a118e25f5cbb980078c52b35a6a9d13f3f85af/packages/bun-vscode/src/features/debug.ts#L95-L132. 

@connor4312 (from the vscode team, who works on debugging experience) would be happy to add some simple metadata to JS debug terminal environment to make Bun debugging work well.

RiskyMH:
fixes #15223 by implementing https://github.com/oven-sh/bun/issues/15223#issuecomment-2506996794

Keeps the previous version existent if older vscode (which can be removed if/when we increase the min requirements of vscode version in package.json).

I have tried this with many bun versions (inc #15788) and node versions and it works well.

---

would appreciate others testing this too as it's a decent change

and @connor4312 was this how you were imagining it being used?


robobun:
<div><sup>Updated 4:20 PM PT - Dec 16th, 2024</sup></div>

@RiskyMH, your commit 84820e579dd9c647c9edfcebe2e326f187ba83dd is building: [`#8056`](https://buildkite.com/bun/bun/builds/8056)

packages/bun-vscode/src/features/debug.ts
```diff
@@ -64,8 +74,15 @@ export function registerDebugger(context: vscode.ExtensionContext, factory?: vsc
       vscode.DebugConfigurationProviderTriggerKind.Dynamic,
     ),
     vscode.debug.registerDebugAdapterDescriptorFactory("bun", factory ?? new InlineDebugAdapterFactory()),
-    vscode.window.onDidOpenTerminal(injectDebugTerminal),
   );
+
+  if (getConfig("debugTerminal.enabled")) {
```
Jarred-Sumner, line 69: Can we feature detect instead of relying on a particular version number? The installed extension may not be the VSCode version

RiskyMH: The version number is vscode and the stable release that included this new feature I use. I'm not 100% sure on if we can feature detect as the `ms-vscode.js-debug` extension has been around for a while (maybe we could just increase the package.json min version instead and not worry)

Jarred-Sumner: this function is new right

RiskyMH: the `getConfig` has been there for some time already (i moved it from the old `injectDebugTerminal` to overarching to be more clear)

packages/bun-vscode/src/features/debug.ts
```diff
@@ -133,6 +148,35 @@ async function injectDebugTerminal(terminal: vscode.Terminal): Promise<void> {
...
   setTimeout(() => terminal.dispose(), 100);
 }
+async function injectDebugTerminal2() {
+  const jsDebugExt = vscode.extensions.getExtension('ms-vscode.js-debug-nightly') || vscode.extensions.getExtension('ms-vscode.js-debug');
+  if (!jsDebugExt) return
+
+  await jsDebugExt.activate()
+  const jsDebug: import('@vscode/js-debug').IExports = jsDebugExt.exports;
+
+  return jsDebug.registerDebugTerminalOptionsProvider({
```
Jarred-Sumner, line 150: `registerDebugTerminalOptionsProvider` is new right

RiskyMH: oh yeah it is

https://github.com/microsoft/vscode-js-debug/pull/2136/files#diff-3d8109446d747816948a48dd717a7477bd2ec4f72669d157fa07a700de7a3e0fR19

packages/bun-vscode/src/features/debug.ts
connor4312: In versions before this was added, `jsDebugExt.exports` will be undefined, so feature detecting that should be pretty easy.

connor4312:
lgtm from the js-debug side :)

packages/bun-vscode/src/features/debug.ts
RiskyMH: oh ok thats awesome, thanks i'll update

Issue #8060, opened 2024-01-09
Incorrect timezone when running within a Docker container.


flysand7:
### What version of Bun is running?

1.0.7+b0393fba6200d8573f3433fb0af258a0e33ac157

### What platform is your computer?

Linux 6.2.0-35-generic x86_64 unknown ---> (the container, not the computer)

### What steps can reproduce the bug?

1. Created a docker container using `oven/bun:1` as the base image, the docker-compose has the following lines added to it:

```diff
+   environment:
+     - TZ=:/etc/localtime
    volumes:
      - ../client:/app
      - /app/node_modules
+     - /etc/timezone:/etc/timezone:ro
+     - /etc/localtime:/etc/localtime:ro
```

2. In the application that is started by bun I print out the following line:

```js
console.log(new Date(Date.now()).toString());
```

### What is the expected behavior?

Should print the time in the same timezone that is set in the container with UTC+11 timezone.

### What do you see instead?

The output is the following:

```
Tue Jan 09 2024 07:02:56 GMT+0000 (Coordinated Universal Time)
```

### Additional information

The local time on my machine and within the container is 11 hours ahead of this (GMT+11). I verified this by attaching a shell to the container and running the `date` command, which displayed the correct timezone.

This issue might be related to #3649, however I believe that bun should be loading the timezone from `/etc/timezone` file regardless of whether the environment variable `TZ` is set. The GNU specification mentioned in the linked issue does mention that it should not be set normally and has a default interpretation of `/etc/timezone`. But I don't know, seems like nodejs doesn't set the timezone too

PR #15802, opened 2024-12-17
pass all upstream node:os tests, all supported node:async_hooks tests

paperclover:
New Node.js test passing:

- `node:os`: pass all of them
- `node:async_hooks` i do not intend to support createHook or async_wrap, however i will support AsyncLocalStorage and AsyncResource. we pass all of these tests.

In addition,

- bindgen gets `validateInteger` and `validateInt32`
- after being bit by a manual c-translation mistake, i've added src/c-headers-for-zig.h, which is run through translate-c and exposed under bun.C.translated. We should move most of our manual definitions to this process eventually (all calls into documented libc, documented windows APIs, all dependencies, and maybe eventually all our own C++ headers)


robobun:
<div><sup>Updated 4:36 PM PT - Dec 16th, 2024</sup></div>

@paperdave, your commit c73168f9248dfc7baff9a900be4c87ff59c9b114 is building: [`#8060`](https://buildkite.com/bun/bun/builds/8060)

src/bun.js/javascript.zig
```diff
@@ -3993,7 +3993,7 @@ pub const VirtualMachine = struct {
 
         if (show.system_code) {
             if (show.syscall) {
-                try writer.writeAll("  ");
+                try writer.writeAll("   ");
```
dylan-conway, line 3996: is this supposed to be two spaces? and do we also need to print spaces before `syscall` in `if (show.syscall)`?

paperclover: it printed with one less in the situation i tried. maybe this whole logic should be redone to just compute the indent level of the : and then be able to compute the # of spaces per line.

dylan-conway (approved):

Issue #15348, opened 2024-11-22
Macros with default import/export


stag-enterprises:
### What version of Bun is running?

1.1.36+ededc168c

### What platform is your computer?

Linux 6.1.91-060191-generic x86_64 x86_64

### What steps can reproduce the bug?

When using default imports/exports, macros do not work.
```ts
// index.ts
import { default as func } from "./func.ts" with { type: "macro" };
// same with alternative:
import func from "./func.ts" with { type: "macro" };

console.log(func());
```

```ts
// func.ts
export default function func() {
  return Math.random();
}
```

### What is the expected behavior?

The macro to work.

### What do you see instead?

```sh
 5 |   Macros = await import('/workspace/template-bun/func.ts');
 6 | } catch (err) {
 7 |    console.error("Error importing macro");
 8 |    throw err;
 9 | }
10 |   throw new Error("Macro 'func' not found in '/workspace/template-bun/func.ts'");
             ^
error: Macro 'func' not found in '/workspace/template-bun/func.ts'
      at macro://20453aec061c25d8.js:10:9
3 | console.log(func());
                ^
error: "MacroLoadError" error in macro
    at /workspace/template-bun/index.tsx:3:13
gitpod /workspace/template-bun (main) $ bun run index.tsx 
 5 |   Macros = await import('/workspace/template-bun/func.ts');
 6 | } catch (err) {
 7 |    console.error("Error importing macro");
 8 |    throw err;
 9 | }
10 |   throw new Error("Macro 'func' not found in '/workspace/template-bun/func.ts'");
             ^
error: Macro 'func' not found in '/workspace/template-bun/func.ts'
      at macro://20453aec061c25d8.js:10:9
      at requestFetch (1:11)
      at requestInstantiate (1:11)
      at requestSatisfyUtil (1:11)
3 | console.log(func());
                ^
error: "MacroLoadError" error in macro
    at /workspace/template-bun/index.tsx:3:13
```

PR #15833, opened 2024-12-18
Fix macro imports

brainkim:
### What does this PR do?

Currently, macros cannot be imported using a namespace import, default imports, or with named import aliases.

```js
// All of the following imports will fail:
import defaultMacro from "macros.js" with {type: "macro"};
import * as macros from "macros.js" with {type: "macro"};
import {macro as myMacro} from "macros.js" with {type: "macro"};
```

Fixes https://github.com/oven-sh/bun/issues/15348

- [x] Code changes

I wrote automated tests

## TODO:

- [ ] increment expected_version in RuntimeTranspilerCache?
- [x] test situations with export aliases or re-exports

### DIFF
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -8972,7 +8972,7 @@ fn NewParser_(
                 return p.s(S.Empty{}, loc);
             }
 
-            const macro_remap = if ((comptime allow_macros))
+            const macro_remap = if (comptime allow_macros)
                 p.options.macro_context.getRemap(path.text)
             else
                 null;
@@ -14653,48 +14653,6 @@ fn NewParser_(
                 }
             }
         }
-
-        pub const MacroVisitor = struct {
-            p: *P,
-
-            loc: logger.Loc,
-
-            pub fn visitImport(this: MacroVisitor, import_data: js_ast.Macro.JSNode.ImportData) void {
-                var p = this.p;
-
-                const record_id = p.addImportRecord(.stmt, this.loc, import_data.path);
-                var record: *ImportRecord = &p.import_records.items[record_id];
-                record.was_injected_by_macro = true;
-                p.macro.imports.ensureUnusedCapacity(import_data.import.items.len) catch unreachable;
-                var import = import_data.import;
-                import.import_record_index = record_id;
-
-                p.is_import_item.ensureUnusedCapacity(
-                    p.allocator,
-                    @as(u32, @intCast(p.is_import_item.count() + import.items.len)),
-                ) catch unreachable;
-
-                for (import.items) |*clause| {
-                    const import_hash_name = clause.original_name;
-
-                    if (strings.eqlComptime(clause.alias, "default")) {
-                        const non_unique_name = record.path.name.nonUniqueNameString(p.allocator) catch unreachable;
-                        clause.original_name = std.fmt.allocPrint(p.allocator, "{s}_default", .{non_unique_name}) catch unreachable;
-                        record.contains_default_alias = true;
-                    }
-                    const name_ref = p.declareSymbol(.import, this.loc, clause.original_name) catch unreachable;
-                    clause.name = LocRef{ .loc = this.loc, .ref = name_ref };
-
-                    p.is_import_item.putAssumeCapacity(name_ref, {});
-
-                    p.macro.imports.putAssumeCapacity(js_ast.Macro.JSNode.SymbolMap.generateImportHash(import_hash_name, import_data.path), name_ref);
-
-                    // Ensure we don't accidentally think this is an export from
-                }
-
-                p.macro.prepend_stmts.append(p.s(import, this.loc)) catch unreachable;
-            }
-        };
 
         pub fn panic(p: *P, comptime fmt: string, args: anytype) noreturn {
             p.panicLoc(fmt, args, null);
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -190,7 +190,9 @@ const exports_string_name: string = "exports";
 
 const MacroRefData = struct {
     import_record_id: u32,
-    name: string,
+    // if name is null the macro is imported as a namespace import
+    // import * as macros from "./macros.js" with {type: "macro"};
+    name: ?string = null,
 };
 
 const MacroRefs = std.AutoArrayHashMap(Ref, MacroRefData);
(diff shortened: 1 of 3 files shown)


src/js_parser.zig
```diff
@@ -16499,12 +16522,15 @@ fn NewParser_(
                         e_.tag = p.visitExpr(tag);
 
                         if (comptime allow_macros) {
```
brainkim: I updated the code for macro template tags but it doesn’t seem like Bun supports template tag macros yet...

src/js_parser.zig
```diff
@@ -17425,8 +17460,13 @@ fn NewParser_(
 
                     if (comptime allow_macros) {
                         if (is_macro_ref and !p.options.features.is_macro_runtime) {
-                            const ref = e_.target.data.e_import_identifier.ref;
-                            const import_record_id = p.macro.refs.get(ref).?;
+                            const ref = switch (e_.target.data) {
+                                .e_import_identifier => |ident| ident.ref,
+                                .e_dot => |dot| dot.target.data.e_identifier.ref,
```
brainkim: I’m not really sure why `.e_dot` expressions have a target which is `.e_identifier` and not `.e_import_identifier` when the target is a namespace identifier, but I’m too afraid to change it.

src/js_parser.zig
```diff
@@ -8969,6 +8990,8 @@ fn NewParser_(
                         .import_record_index = stmt.import_record_index,
                     }) catch unreachable;
                 }
+
+                // TODO: not sure how to handle macro remappings for namespace imports
```
brainkim, line 8994: I’m not too sure about how the package.json macro remapping feature works. I tried to do what I can, but to allow namepspace imports to be remapped would probably be pretty hard as far as I can tell.

Jarred-Sumner:
I think this approach overall makes sense.

src/js_parser.zig
Jarred-Sumner: We remove these imports before they reach the final output

test/bundler/transpiler/macro-test.test.ts
```diff
@@ -1,6 +1,12 @@
 import { escapeHTML } from "bun" assert { type: "macro" };
 import { expect, test } from "bun:test";
 import { addStrings, addStringsUTF16, escape, identity } from "./macro.ts" assert { type: "macro" };
+import defaultMacro, {
+  identity as identity1,
+  identity as identity2,
+} from "./macro.ts" assert { type: "macro" };
+
+import * as macros from "./macro.ts" assert { type: "macro" };
```
Jarred-Sumner, line 10: Can you add one more test that aliases a default import?

```js
import { default as defaultMacro2 } from './macro.ts" assert {type: "macro"};
```

Jarred-Sumner (approved):
Looks good assuming the relevant tests pass and it doesn't cause any regressions (I don't think it will cause any regressions) 

The main thing I'd be curious about is how we generate the wrapper function in the macro to register the callback to call. I haven't looked at that code in awhile.

brainkim:
@Jarred-Sumner The macros are generated using an entrypoint in src/bundler/entry_points.zig

The exports are imported with a top-level await on an `import` expression.

```zig
// This is not very fast. The idea is: we want to generate a unique entry point
// per macro function export that registers the macro Registering the macro
// happens in VirtualMachine We "register" it which just marks the JSValue as
// protected. This is mostly a workaround for being unable to call ESM exported
// functions from C++. When that is resolved, we should remove this.
pub const MacroEntryPoint = struct {
    code_buffer: [bun.MAX_PATH_BYTES * 2 + 500]u8 = undefined,
    output_code_buffer: [bun.MAX_PATH_BYTES * 8 + 500]u8 = undefined,
    source: logger.Source = undefined,

    pub fn generateID(entry_path: string, function_name: string, buf: []u8, len: *u32) i32 {
        var hasher = bun.Wyhash11.init(0);
        hasher.update(js_ast.Macro.namespaceWithColon);
        hasher.update(entry_path);
        hasher.update(function_name);
        const hash = hasher.final();
        const fmt = bun.fmt.hexIntLower(hash);

        const specifier = std.fmt.bufPrint(buf, js_ast.Macro.namespaceWithColon ++ "//{any}.js", .{fmt}) catch unreachable;
        len.* = @as(u32, @truncate(specifier.len));

        return generateIDFromSpecifier(specifier);
    }

    pub fn generateIDFromSpecifier(specifier: string) i32 {
        return @as(i32, @bitCast(@as(u32, @truncate(bun.hash(specifier)))));
    }

    pub fn generate(
        entry: *MacroEntryPoint,
        _: *Bundler,
        import_path: Fs.PathName,
        function_name: string,
        macro_id: i32,
        macro_label_: string,
    ) !void {
        const dir_to_use: string = if (import_path.dir.len == 0) "" else import_path.dirWithTrailingSlash();
        bun.copy(u8, &entry.code_buffer, macro_label_);
        const macro_label = entry.code_buffer[0..macro_label_.len];

        const code = brk: {
            if (strings.eqlComptime(import_path.base, "bun")) {
                break :brk try std.fmt.bufPrint(
                    entry.code_buffer[macro_label.len..],
                    \\//Auto-generated file
                    \\var Macros;
                    \\try {{
                    \\  Macros = globalThis.Bun;
                    \\}} catch (err) {{
                    \\   console.error("Error importing macro");
                    \\   throw err;
                    \\}}
                    \\const macro = Macros['{s}'];
                    \\if (!macro) {{
                    \\  throw new Error("Macro '{s}' not found in 'bun'");
                    \\}}
                    \\
                    \\Bun.registerMacro({d}, macro);
                ,
                    .{
                        function_name,
                        function_name,
                        macro_id,
                    },
                );
            }

            break :brk try std.fmt.bufPrint(
                entry.code_buffer[macro_label.len..],
                \\//Auto-generated file
                \\var Macros;
                \\try {{
                \\  Macros = await import('{s}{s}');
                \\}} catch (err) {{
                \\   console.error("Error importing macro");
                \\   throw err;
                \\}}
                \\if (!('{s}' in Macros)) {{
                \\  throw new Error("Macro '{s}' not found in '{s}{s}'");
                \\}}
                \\
                \\Bun.registerMacro({d}, Macros['{s}']);
            ,
                .{
                    bun.fmt.fmtPath(u8, dir_to_use, .{
                        .escape_backslashes = true,
                    }),
                    bun.fmt.fmtPath(u8, import_path.filename, .{
                        .escape_backslashes = true,
                    }),
                    function_name,
                    function_name,
                    bun.fmt.fmtPath(u8, dir_to_use, .{
                        .escape_backslashes = true,
                    }),
                    bun.fmt.fmtPath(u8, import_path.filename, .{
                        .escape_backslashes = true,
                    }),
                    macro_id,
                    function_name,
                },
            );
        };

        entry.source = logger.Source.initPathString(macro_label, code);
        entry.source.path.text = macro_label;
        entry.source.path.namespace = js_ast.Macro.namespace;
    }
};
```

I thought I might have to do additional work to handle re-exports, but this seems to be robust so long as you have the name of the export.

paperclover (approved):

Issue #12431, opened 2024-07-08
Do not elide lines with `bun run` with workspaces


spiffytech:
### What is the problem this feature would solve?

My DB migration library prints out what's wrong with my SQL, but running `bun run migrate` from my project top-level omits the most important lines from the error: the SQL error message.

```
$ bun run migrate
$ bun run --filter server --env-file .env migrate
server migrate $ bunx pg-migrations apply --directory migrations && bunx @databases/pg-schema-cli --directory __generated__
│ [8 lines elided]
│     at handleError (/home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg/lib/Driver.js:394:29)
│     at executeQueryInternal (/home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg/lib/Driver.js:369:9)
│     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
│     at async PgDriver._executeQuery (/home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg/lib/Driver.js:199:29)
│     at async PgDriver.executeAndReturnLast (/home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg/lib/Driver.js:226:16)
│     at async Transaction.query (/home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/shared/lib/BaseTransaction.js:59:24)
│     at async /home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg-migrations/lib/PostgresDatabaseEngine.js:19:25
│     at async Object.applyMigration (/home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg-migrations/lib/PostgresDatabaseEngine.js:137:21)
│     at async /home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/migrations-base/lib/methods/applyOperations.js:43:29
│     at async /home/spiffytech/Documents/bootspoon/barbellbasics/node_modules/@databases/pg-migrations/lib/PostgresDatabaseEngine.js:54:20
└─ Exited with code 1
error: script "migrate" exited with code 1
```

If I `cd` into `packages/server` and run the same command manually I get the full error:

```
$ bun run --env-file ../../.env migrate
$ bunx pg-migrations apply --directory migrations && bunx @databases/pg-schema-cli --directory __generated__
Applying 0003-lifts-order.sql
[...]

Issue #12216, opened 2024-06-27
Values in `bunfig.toml` take precedence over command line args


jakeboone02:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Linux 5.15.153.1-microsoft-standard-WSL2 x86_64 x86_64

#### bunfig.toml:
```toml
[test]
coverage = false
```
#### index.js:
```js
export const x = () => "x";
export const y = () => "y";
```
#### idx.test.js:
```js
import { x } from ".";
it("works", () => { expect(x()).toBe("x"); });
```
```shell
bun test idx.test.js --coverage
```

### What is the expected behavior?

The `--coverage` flag on the command line should override the `coverage = false` configuration in `bunfig.toml`, and coverage should be collected/reported.

### What do you see instead?

The `coverage` value configured in `bunfig.toml` always takes precedence over the `--coverage` command line flag. In the example above, coverage is not reported due to the `coverage = false` in `bunfig.toml`.

### Additional information

I was starting to address #4346, but ran into this issue.

jakeboone02:
I added this test locally to `/test/cli/test/coverage.test.ts` track it.

```diff
diff --git a/test/cli/test/coverage.test.ts b/test/cli/test/coverage.test.ts
index d5bb7bdb7..1699b5958 100644
--- a/test/cli/test/coverage.test.ts
+++ b/test/cli/test/coverage.test.ts
@@ -54,3 +54,48 @@ export class Y {
   expect(result.signalCode).toBeUndefined();
   expect(readFileSync(path.join(dir, "coverage", "lcov.info"), "utf-8")).toMatchSnapshot();
 });
+
+test("command line flag --coverage overrides bunfig", async () => {
+  const dir = tempDirWithFiles("cmd-line-overrides-bunfig", {
+    "cmd-line-overrides-bunfig.test.ts": `
+import { covered } from "./cmd-line-overrides-bunfig";
+
+test("cmd line overrides bunfig", () => {
+  expect(covered()).toBe(42);
+});
+`,
+    "cmd-line-overrides-bunfig.ts": `
+export function covered() {
+  // this function IS covered
+  return 42;
+}
+
+export function uncovered() {
+  // this function is not covered
+  return 43;
+}
+`,
+    "bunfig.toml": `
+[test]
+coverage = false
+`,
+    "package.json": JSON.stringify(
+      {
+        name: "cmd-line-overrides-bunfig",
+        version: "1.0.0",
+      },
[...]

PR #15837, opened 2024-12-18
add --elide-lines override flag for workspace filtering

martinamps:
### What does this PR do?

This adds a new flag --elide-lines to bun test to allow overriding the default value of 10. Lots of other folks ran into this in https://github.com/oven-sh/bun/issues/12431

Manually and via unit tests

before:

```
✗ bun run  --filter gateway test
gateway test $ bun test
│ [28 lines elided]
│
│ tests/unit/openai/structuredOutput.test.ts:
│ (pass) OpenAI Structured Output > should handle structured output for research paper extraction [0.80ms]
│ (pass) OpenAI Structured Output > should handle structured output for math reasoning [0.28ms]
│ (pass) OpenAI Structured Output > should handle invalid structured output [0.29ms]
│
│  22 pass
│  0 fail
│  155 expect() calls
│ Ran 22 tests across 4 files. [66.00ms]
└─ Done in 77 ms
```

after:

```
✗ /Users/martinamps/code/bun/build/release/bun run --elide-lines 0 --filter gateway test
gateway test $ bun test
│ bun test v1.1.39 (1d485617)
│
│ tests/unit/model-definitions/getDefinition.test.ts:
│ (pass) Model Definitions > should return correct OpenAI chat model definition
│ (pass) Model Definitions > should return correct Anthropic chat model definition [0.14ms]
│ (pass) Model Definitions > should handle unknown model gracefully
│ (pass) Model Definitions > should type check different model types correctly [0.03ms]
│ (pass) Model Definitions > provider fields should be present and correct type [0.04ms]
│
│ tests/unit/openai/functionCalling.test.ts:
│ Cache Middleware is not enabled because caches is not defined.
│ (pass) OpenAI Function Calling > should handle function calling with a single function [3.81ms]
│ (pass) OpenAI Function Calling > should handle function calling with multiple functions [0.67ms]
│ (pass) OpenAI Function Calling > should handle function calling with a specified function [0.53ms]
│ (pass) OpenAI Function Calling > should handle function execution and follow-up [0.53ms]
│ (pass) OpenAI Function Calling > should handle errors in function calling [0.33ms]
│ (pass) OpenAI Function Calling > should handle function calling with invalid arguments [0.46ms]
│ (pass) OpenAI Function Calling > should handle streaming responses [1.75ms]
│
│ tests/unit/openai/chatCompletions.test.ts:
│ (pass) OpenAI Chat Completions > should forward a basic chat completion request [0.49ms]
│ (pass) OpenAI Chat Completions > should handle rate limiting [0.25ms]
│ mamps here3
│ (pass) OpenAI Chat Completions > should handle network errors [0.33ms]
│ (pass) OpenAI Chat Completions > should reject requests with invalid content type [0.21ms]
│ (pass) OpenAI Chat Completions > should use the correct API key [0.28ms]
│ (pass) OpenAI Chat Completions > should handle very large requests [0.32ms]
│ (pass) OpenAI Chat Completions > should handle streaming responses [0.36ms]
│
│ tests/unit/openai/structuredOutput.test.ts:
│ (pass) OpenAI Structured Output > should handle structured output for research paper extraction [0.87ms]
│ (pass) OpenAI Structured Output > should handle structured output for math reasoning [0.28ms]
│ (pass) OpenAI Structured Output > should handle invalid structured output [0.31ms]
│
│  22 pass
│  0 fail
│  155 expect() calls
│ Ran 22 tests across 4 files. [68.00ms]
└─ Done in 77 ms
```

The coupling of elide to pretty output was a little tight so my matching for the output is a little relaxed as a result and I had to make minor tweaks to the test runner to pass in the extra env flags


jakeboone02:
Might be nice to have this as a bunfig option as well, so it doesn't have to be included in every command.

Jarred-Sumner:
Nice. Looks like there's some test failures, but this is a good idea and I would like to merge it once the tests pass.

src/cli.zig
```diff
@@ -243,13 +243,15 @@ pub const Arguments = struct {
...
         clap.parseParam("--silent                          Don't print the script command") catch unreachable,
+        clap.parseParam("--elide-lines <NUMBER>            Number of lines of script output shown (default: 10). Set to 0 to show all lines.") catch unreachable,
         clap.parseParam("-v, --version                     Print version and exit") catch unreachable,
         clap.parseParam("--revision                        Print version with revision and exit") catch unreachable,
     } ++ auto_or_run_params;
     pub const auto_params = auto_only_params ++ runtime_params_ ++ transpiler_params_ ++ base_params_;
 
     const run_only_params = [_]ParamType{
         clap.parseParam("--silent                          Don't print the script command") catch unreachable,
+        clap.parseParam("--elide-lines <NUMBER>            Number of lines of script output shown (default: 10). Set to 0 to show all lines.") catch unreachable,
```
RiskyMH: i think it may help to make sure its clear that `--elide-lines` is only when `--filter` is used (and not just normally only allowing 10 lines for the script run)

martinamps:
@jakeboone02 sounds reasonable! I attempted to add it here https://github.com/oven-sh/bun/blob/main/src/bunfig.zig#L592 and it didn't do anything - happen to know why off the top of your head before I debug?

@Jarred-Sumner awesome 🙏 which tests? looks like they're all passing in CICD afaict

src/cli.zig
martinamps: good point - will do so on my next push

jakeboone02:
> @jakeboone02 sounds reasonable! I attempted to add it here https://github.com/oven-sh/bun/blob/main/src/bunfig.zig#L592 and it didn't do anything - happen to know why off the top of your head before I debug?

I don't know but be aware of #12216 (although that seems like the opposite problem...🤷‍♂️).

RiskyMH:
>  which tests? looks like they're all passing in CICD afaict

I'm not sure why its green status, but if you press "details" it will send you to the page: https://buildkite.com/bun/bun/builds/8169

jakeboone02:
> @jakeboone02 I agree with you it'd be nice to have it in bunfig but unclear to me immediately why that isn't working like --silent.. Ok if we make that a follow-up PR when I have more cycles?

Sure! The sooner the better for the CLI flag IMO. A bunfig option can wait.

Jarred-Sumner:
I woud like to merge this but there are Windows test failures

martinamps:
ah-- I think it's because windowsIsTerminal() does not respect FORCE_COLOR which is how I trigger the code path in the other environments https://github.com/oven-sh/bun/blob/main/src/cli/filter_run.zig#L534 -- now the code will explicitly error if you try use it in a non TTY environment it'll error, and I explicitly test that on windows. Hopefully passes this time 😅 sorry!

Veskel01:
My goodness - I'm waiting for the merge. This is the most annoying thing about Bun-based monorepo

Jarred-Sumner:
Thank you. This will be part of Bun v1.1.43

Issue #8170, opened 2024-01-14
JSX/TSX memory leak


7f8ddd:
### What version of Bun is running?

1.0.23+317998fdd (canary)

### What platform is your computer?

Linux 5.15.133.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

```js
import "@kitajs/html/register";

Bun.serve({
  port: 8080,
  fetch() {
    return new Response(
      (
        <div>
          Test
          <div>
            Test
            <div>
              Test
              <div>
                Test<div>Test</div>
              </div>
            </div>
          </div>
        </div>
      ).toString()
    );
  },
});
```

### What is the expected behavior?

No memory leak.

### What do you see instead?

The memory usage skyrockets depending on the number of elements, or strings, with a single benchmark adding 5% on my actual website.

![image](https://github.com/oven-sh/bun/assets/141754874/45eca7b1-86b8-4e55-8bec-e76cac82e76f)

### Additional information

Issue is not present on stable.

PR #15838, opened 2024-12-18
perf(node:util): fast path for `extractedSplitNewLines`

DonIsaac:
`extractedSplitNewLines` uses a regex with a positive lookahead to spit strings by newlines. Benchmarks show that this causes `inspect` to be ~2-3x slower than node when this codepath is hit. This PR adds a fast path when the object being split is known to be a string.

We cannot run this fast path in all cases since splitting via a regexp respects `Symbol.split`.


robobun:
<div><sup>Updated 12:41 AM PT - Dec 18th, 2024</sup></div>

Your commit 125a2cf920a0e660cf5c4368d06c19bdf1cf27bd is building: [`#8170`](https://buildkite.com/bun/bun/builds/8170)

src/bun.js/node/node_util_binding.zig
```diff
@@ -105,3 +105,70 @@ pub fn internalErrorName(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFr
...
+    var fallback = std.heap.stackFallback(1024, bun.default_allocator);
+    const allocator = fallback.get();
+    bun.assert(callframe.argumentsCount() == 1);
+    const value = callframe.argument(0);
+    bun.assert(value.isString());
+
+    const str = try value.toBunString2(globalThis);
+
+    // if (str.isUTF16()) return JSC.JSValue.jsUndefined();
+    if (str.is8Bit() or bun.strings.isAllASCII(str.byteSlice())) {
```
paperclover: `bun.strings.isAllASCII(str.byteSlice())` is reinterpreting bytes of utf16 data as utf8/ascii. for example, this check is wrong for the character `㈴`, which in UTF-16 would be `0x34 0x32`, interpreted as ascii gives you `42`

(imo, byteSlice should be removed to prevent this category of mistake)

src/bun.js/node/node_util_binding.zig
```diff
@@ -105,3 +105,70 @@ pub fn internalErrorName(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFr
...
+
+/// `extractedSplitNewLines` for ASCII/Latin1 strings. Panics if passed a non-string.
+//Returns `undefined` if param is utf8 or utf16 and not fully ascii.
+///
+/// ```js
+/// // util.js
+/// const extractedNewLineRe = new RegExp("(?<=\\n)");
+/// extractedSplitNewLines = value => RegExpPrototypeSymbolSplit(extractedNewLineRe, value);
+/// ```
+pub fn extractedSplitNewLinesFastPathStringsOnly(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) bun.JSError!JSC.JSValue {
```
paperclover, line 118: this might be a good place to use bindgen, though something we should add to that is a way to turn all validations into assertions (internal APIs like this)

```zig
pub fn extractedSplitNewLinesFastPathStringsOnly(global: *JSC.JSGlobalObject, str: bun.String) !JSC.JSValue {
    // ...
}
```
Create `node_util_binding.bind.ts`
```
import { fn, t } from 'bindgen';
export const extractedSplitNewLinesFastPathStringsOnly = fn({
  args: {
    global: t.globalObject,
    str: t.DOMString,
  },
  ret: t.any,
})
```
in JavaScript builtins
```diff
-$newZigFunction("node_util_binding.zig", "extractedSplitNewLinesFastPathStringsOnly", 1);
+$bindgenFn("node_util_binding.bind.ts", "extractedSplitNewLinesFastPathStringsOnly");
```
I wont block the pr on this point, but i think your interest in bindgen would want to use it here.

src/bun.js/node/node_util_binding.zig
```diff
@@ -105,3 +105,70 @@ pub fn internalErrorName(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFr
...
+            lines.deinit(allocator);
+        }
+
+        var start: usize = 0;
+        const bytes = str.byteSlice();
+
+        while (std.mem.indexOfScalarPos(u8, bytes, start, '\n')) |delim_start| {
+            const end = delim_start + 1;
+            const buf = bytes[start..end];
+            const s = switch (str.encoding()) {
```
paperclover: the string's encoding is already known to be latin1

paperclover (changes requested):
i think on my end, i should make the bindings generator pass `WTFString` instead of `bun.String`, since it will always be of the WTFStringImpl tag. this is significant because the input to this function is always `latin1` or `utf16`, and never `utf8`. checking for the `utf8` encoding in this function shouldn't be done since it wont ever be present.

paperclover (approved):

PR #15840, opened 2024-12-18
Add option to include or exclude files in the testing coverage

abenjaminov:
This PR adds support for filtering files included or excluded from coverage during `bun test`. The feature allows users to specify `coverageInclude` and `coverageExclude` options in their configuration. Both glob patterns and regular expressions are supported.

For example:

```toml
[test]
coverageInclude = ["**/*.service.ts"]
coverageExclude = ["**/*.test.ts"]
```

- [x] I wrote automated tests for the new functionality in `patterns.zig` and verified they pass.
- [x] I ran local tests for the `matchesGlob`, `isGlobPattern`, `matchesRegex`, and `matchesAnyPattern` functions to confirm correctness.
- [x] I validated the integration by ensuring the `coverageInclude` and `coverageExclude` settings work as intended with a sample test configuration.

#### For Zig changes:
- [x] I checked the lifetime of memory allocated to verify it's (1) freed and (2) only freed when it should be.
- [x] I included a test for the new code, ensuring coverage settings are respected.


Jarred-Sumner (changes requested):
Thank you for this PR.

- Instead of linking and including PCRE2, you can use `bun.glob` to match globs
- This will need tests, which you can see examples of in the `test` folder

ayalonmen1:
LGTM!

mladibejn:
kudos

abenjaminov:
> Thank you for this PR.
> 
> * Instead of linking and including PCRE2, you can use `bun.glob` to match globs
> * This will need tests, which you can see examples of in the `test` folder

Thanks for the reply, i have incorperated bun.glob for this and wrote a few testing cases

christian-draeger:
what happend to the PR?
would relly like to see it merged

PR #15862, opened 2024-12-19
don't make inline sourcemap in normal vscode terminal

RiskyMH:
the in-vscode error reporting it effectively creates a debugger, however we don't need everything (in this context sourcemap is unneeded)

I have manually tested and both debugger and inline errors still appear to work


robobun:
<div><sup>Updated 5:02 PM PT - Dec 18th, 2024</sup></div>

@RiskyMH, your commit 36867a02829a8386a2e6f54926e7154ccfba7448 is building: [`#8185`](https://buildkite.com/bun/bun/builds/8185)

PR #15864, opened 2024-12-19
DNS fixes

heimskr:
### What does this PR do?

Adds some DNS tests and the changes that caused them to start passing.

- [x] Code changes

Added tests from Node.


robobun:
<div><sup>Updated 6:03 PM PT - Dec 18th, 2024</sup></div>

@heimskr, your commit 3aa930e038c0123a3cc2e4cd54a9fdf4f59eb0d2 is building: [`#8189`](https://buildkite.com/bun/bun/builds/8189)

Jarred-Sumner (changes requested):
After each request is sent or fulfilled, we should check if there are still any in-flight requests and if there are no longer any, then cancel the timer so that we don't keep the process alive longer than necessary.

A helper method for updating the timer state would be a good idea I think

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2864,13 +3112,192 @@ pub const DNSResolver = struct {
...
+            const port = JSValue.getIndex(triple, globalThis, 2).toInt32();
+
+            if (family != 4 and family != 6) {
+                return globalThis.throwInvalidArguments("Invalid address family", .{});
+            }
+
+            const addressString = try JSValue.getIndex(triple, globalThis, 1).toBunString2(globalThis);
+            defer addressString.deref();
+
+            const addressSlice = addressString.toSlice(bun.default_allocator).slice();
```
Jarred-Sumner: This leaks memory

Jarred-Sumner: Instead, you want to call `.toOwnedSlice`

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2864,13 +3112,192 @@ pub const DNSResolver = struct {
...
+        defer bun.default_allocator.free(entries);
+
+        var i: u32 = 0;
+
+        while (triplesIterator.next()) |triple| : (i += 1) {
+            if (!triple.isArray()) {
+                return globalThis.throwInvalidArgumentType("setServers", "triple", "array");
+            }
+
+            const family = JSValue.getIndex(triple, globalThis, 0).toInt32();
```
Jarred-Sumner, line 3245: This would be a great use-case for the bindings generator @paperdave

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2864,13 +3112,192 @@ pub const DNSResolver = struct {
...
+            const port = JSValue.getIndex(triple, globalThis, 2).toInt32();
+
+            if (family != 4 and family != 6) {
```
Jarred-Sumner, line 3248: Do they allow `"IPv6"` and `"IPv4"` strings?

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2864,13 +3112,192 @@ pub const DNSResolver = struct {
...
+
+        var i: u32 = 0;
+
+        while (triplesIterator.next()) |triple| : (i += 1) {
+            if (!triple.isArray()) {
+                return globalThis.throwInvalidArgumentType("setServers", "triple", "array");
+            }
+
+            const family = JSValue.getIndex(triple, globalThis, 0).toInt32();
+            const port = JSValue.getIndex(triple, globalThis, 2).toInt32();
```
Jarred-Sumner, line 3246: This should probably use `coerce` because what happens if you pass it a Symbol or something that is not a valid integer?

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2886,15 +3313,27 @@ pub const DNSResolver = struct {
         }
 
         const addr_s = addr_str.getZigString(globalThis).slice();
-        const port: u16 = if (port_value.isNumber()) blk: {
-            break :blk port_value.to(u16);
-        } else {
-            return globalThis.throwInvalidArgumentType("lookupService", "port", "invalid port");
+        const port: u16 = blk: {
+            if (port_value.isNumber()) {
+                const double = try port_value.toNumber(globalThis);
```
Jarred-Sumner: We really need a helper method for getting a valid port number

src/bun.js/bindings/bindings.zig
```diff
@@ -3105,6 +3098,15 @@ pub const JSGlobalObject = opaque {
         return this.throwValue(this.createInvalidArgumentType(name_, field, typename));
     }
 
+    pub fn throwInvalidArgumentValue(
+        this: *JSGlobalObject,
+        argname: []const u8,
+        value: JSValue,
+    ) bun.JSError {
+        var formatter = JSC.ConsoleObject.Formatter{ .globalThis = this };
```
Jarred-Sumner, line 2969:
```suggestion
        var formatter = JSC.ConsoleObject.Formatter{ .globalThis = this };
        defer formatter.deinit()
```

src/deps/c_ares.zig
```diff
@@ -200,39 +201,38 @@ pub const struct_hostent = extern struct {
...
-                return array;
-            }
-            return JSC.JSValue.createEmptyArray(globalThis, 0);
-        } else {
-            if (this.h_aliases == null) {
+            // A cname lookup always returns a single record but we follow the common API here.
+            if (this.h_name == null) {
                 return JSC.JSValue.createEmptyArray(globalThis, 0);
             }
+            const array = JSC.JSValue.createEmptyArray(globalThis, 1);
```
Jarred-Sumner: Use String.toJSArray, there are a surprising number of footguns when iteratively creating an array

src/bun.js/api/bun/dns_resolver.zig
heimskr: Our implementation of dns.(promises.)Resolver.setServers in dns.ts always passes either 4, 6 or 0 to this Zig method, and the only way the Zig method could get any other value is if the user uses the undocumented Bun.dns.newResolver function and calls setServers on the return value, but it's unlikely anyone would ever want to do that

src/bun.js/api/bun/dns_resolver.zig
heimskr: That can happen only if the user directly uses Bun.dns.newResolver instead of the node:dns module's Resolver class, which always passes an integer.

src/bun.js/bindings/bindings.zig
heimskr: Formatter doesn't have a deinit function

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -600,6 +600,16 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_ERR_BUFFER_OUT_OF_BOUNDS, (JSC::JSGlobalObje
     return JSC::JSValue::encode(createError(globalObject, ErrorCode::ERR_BUFFER_OUT_OF_BOUNDS, "Attempt to access memory outside buffer bounds"_s));
 }
 
+JSC_DEFINE_HOST_FUNCTION(jsFunction_ERR_INVALID_IP_ADDRESS, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
```
Jarred-Sumner: instead of creating a one-off funciton for htis, modify the function in this file that handels all the cases

src/bun.js/bindings/bindings.zig
Jarred-Sumner: rebase

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2864,13 +3112,192 @@ pub const DNSResolver = struct {
...
+        const slice = str.toSlice(bun.default_allocator).slice();
+        var buffer = bun.default_allocator.alloc(u8, slice.len + 1) catch bun.outOfMemory();
+        defer bun.default_allocator.free(buffer);
+        _ = strings.copy(buffer[0..], slice);
+        buffer[slice.len] = 0;
+
+        var addr: [16]u8 = undefined;
+
+        if (c_ares.ares_inet_pton(c_ares.AF.INET, buffer.ptr, &addr) == 1) {
+            const ip = std.mem.readInt(u32, addr[0..4], std.builtin.Endian.big);
```
Jarred-Sumner:
```suggestion
            const ip = std.mem.readInt(u32, addr[0..4], .big);
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -4173,6 +4179,14 @@ JSC__JSString* JSC__JSValue__toStringOrNull(JSC__JSValue JSValue0, JSC__JSGlobal
     return value.toStringOrNull(arg1);
 }
 
+bool JSC__JSValue__toNumber(JSC::EncodedJSValue value, JSC::JSGlobalObject* global, double* result)
```
Jarred-Sumner: Why was this added? There are several functions for returning a number

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2864,13 +3112,192 @@ pub const DNSResolver = struct {
...
+            return c_ares.AF.INET6;
+        }
+
+        return JSC.Error.ERR_INVALID_IP_ADDRESS.throw(globalThis, "Invalid IP address: \"{s}\"", .{slice});
+    }
+
+    fn setChannelServers(channel: *c_ares.Channel, globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) bun.JSError!JSC.JSValue {
+        // It's okay to call dns.setServers with active queries, but not dns.Resolver.setServers
+        if (channel != try getChannelFromVM(globalThis) and c_ares.ares_queue_active_queries(channel) != 0) {
+            return globalThis.throwValue((JSC.SystemError{
```
Jarred-Sumner: add this to ErrorCode.ts and throw that instead

src/bun.js/api/bun/dns_resolver.zig
heimskr: Replaced Bun.dns.newResolver with $newZigFunction

src/bun.js/api/bun/dns_resolver.zig
heimskr: Replaced Bun.dns.newResolver with $newZigFunction

src/bun.js/api/bun/dns_resolver.zig
```diff
@@ -2212,22 +2434,26 @@ pub const DNSResolver = struct {
         ttl: i32 = 0,
     };
 
-    pub const RecordType = enum(u8) {
+    pub const RecordType = enum(c_int) {
```
Jarred-Sumner, line 2437: Is the c_int to be consistent with CAres or another reason?

heimskr: It's to be consistent with CAres. For example, CAA is 257, which doesn't fit in a u8.

src/js/node/dns.ts
```diff
@@ -60,11 +321,17 @@ function lookup(domain, options, callback) {
 }
 
 function lookupService(address, port, callback) {
-  if (typeof callback != "function") {
-    throw new TypeError("callback must be a function");
+  if (arguments.length !== 3) {
```
Jarred-Sumner:
```suggestion
  if (arguments.length < 3) {
```

docs/install/lockfile.md
```diff
@@ -72,6 +72,24 @@ $ bun install --yarn
 print = "yarn"
 ```
 
+### Text-based lockfile
+
+Bun v1.1.39 introduced `bun.lock`, a JSONC formatted lockfile. `bun.lock` is human-readable and git-diffable without configuration, at [no cost to performance](https://bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster).
```
Jarred-Sumner: Is this change intentional?

heimskr: Might be an artifact of my rebasing. I didn't intend to change any docs

Issue #8207, opened 2024-01-16
`satori-html` / `ultrahtml` packages not working in Bun


JulianCataldo:
### What version of Bun is running?

1.0.21+837cbd60d

### What platform is your computer?

Darwin 22.6.0 x86_64 i386

### What steps can reproduce the bug?

Reproduction repository and cross-issue:

- https://github.com/natemoo-re/ultrahtml/issues/66
- https://github.com/JulianCataldo/satori-html-bun

The issue is located in `parsel`, `satori-html` or `ultrahtml`, not quite sure yet.

Works fine with other runtimes (node with transpiled ts to js and `tsx`)

grmkris:
same here :( following

samdze:
Stumbled onto this too, following as well

Norbiros:
@dylan-conway Could you please take a look at this or point me to a good starting point? Should I try implementing UTF-16 printing, or would it be better to add UTF-8 support in JSC’s parser?

PR #15870, opened 2024-12-19
Fixes #3554

### LINKED ISSUE #3554: Support JSDOM

Currently DOM testing *can* be achieved, albeit through a third party library, while this is fine it'd be great if we can standardise this to some extent similar to Vite / Vitest's approach, by making this an easily configurable option.

For tests, I believe the DOM should automatically be added in where for instance the `tsconfig.json` contains `DOM` in the `lib` field, or alternatively a configuration option in `bunfig.toml`. This could also additionally be used to enable DOM matchers (related to #1825).

JSDOM is the obvious candidate, but other alternatives could be considered here, such as happy-dom. Although from my (limited) understanding, JSDOM remains prevalent with many testing libraries out there.

Related to #198 and #1825 

N/A

Jarred-Sumner:
### What does this PR do?

Fixes #3554

Tests


robobun:
<div><sup>Updated 10:17 PM PT - Dec 18th, 2024</sup></div>

@Jarred-Sumner, your commit 8084dba076ee04081a576cf3b7abbfd8cab848d0 is building: [`#8207`](https://buildkite.com/bun/bun/builds/8207)

PR #8286, opened 2024-01-19
Make process.binding errors more specific to get better idea of which methods to support

jdalton:
### What does this PR do?

Make process.binding errors more specific to get better idea of which methods to support

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

- [x] I included a test for the new code, or existing tests cover it
- [x] I ran my tests locally and they pass (`bun-debug test test-file-name.test`)

PR #15888, opened 2024-12-20
fix(lexer): do not treat '#bun' in a url as a pragma

DonIsaac:
### What does this PR do?

- [x] Code changes

Added test cases

### DIFF
--- a/src/js_lexer.zig
+++ b/src/js_lexer.zig
@@ -1941,7 +1941,6 @@ fn NewLexer_(
                             );
                         }
 
-                        var has_seen_text = false;
                         for (@as([strings.ascii_vector_size]u8, vec), 0..) |c, i| {
                             switch (c) {
                                 '@', '#' => {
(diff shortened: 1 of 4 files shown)


robobun:
<div><sup>Updated 5:44 PM PT - Dec 19th, 2024</sup></div>

Your commit 153d99cdbadc032144c2710eebad678df2155f73 is building: [`#8286`](https://buildkite.com/bun/bun/builds/8286)

src/js_lexer.zig
```diff
@@ -2012,8 +2015,9 @@ fn NewLexer_(
                             }
                         }
 
+                        // @bun must be the first pragma. it may have spaces before it.
                         if (lexer.bun_pragma == .none and strings.hasPrefixWithWordBoundary(chunk, "bun")) {
```
Jarred-Sumner: This must be a bug in hasPrefixWithWordBoundary. That is supposed to only apply when there is a space or tab character between. it does another pass through the input to check.

paperclover (changes requested):
the fix should just remove the bun_pragma field from the lexer entirely. right now bun is checking every comment for `@bun` when it only needs to check the first one in the file.

DonIsaac:
> the fix should just remove the bun_pragma field from the lexer entirely. right now bun is checking every comment for `@bun` when it only needs to check the first one in the file.

Doing this requires us to have some lexing logic in `js_parser.zig`. Are we ok with this?

Jarred-Sumner:
> Doing this requires us to have some lexing logic in js_parser.zig. Are we ok with this?

We could have a 'has_seen_any_comments' boolean and then not check after that

src/js_lexer.zig
```diff
@@ -2012,8 +2007,9 @@ fn NewLexer_(
...
```
paperclover: you should be able to remove `bun_pragma`, now that the value is never being read by any code. after removing the field, the changes in this section that write to `bun_pragma` can then be eliminated. the `has_seen_text` part also becomes dead.

paperclover (changes requested):
one nitpick

src/js_lexer.zig
DonIsaac: I can't: `lexer.start != 0`

paperclover (approved):

Issue #8290, opened 2024-01-20
bun test --coverage off by 2 error


hholst80:
### What version of Bun is running?

1.0.23

### What platform is your computer?

I run arch (and oven/bun:1)

### What steps can reproduce the bug?

```
# bun test --coverage http.test.mjs
bun test v1.0.23 (83f2432d)

http.test.mjs:
✓ GET /api/live [0.82ms]
✓ GET /api/live/site-test [0.41ms]
✓ DELETE /api/live/test-site [1.17ms]
✓ POST /api/live with empty body should fail [1.43ms]
✓ POST /api/live site=site-test [0.89ms]
✓ POST /api/live site=site-test live:=true [0.33ms]
✓ POST /api/live site=site-test description=kallekula [0.40ms]
✓ POST /api/live site=site-test live:=false description=kallekula [0.34ms]
---------------|---------|---------|-------------------
File           | % Funcs | % Lines | Uncovered Line #s
---------------|---------|---------|-------------------
All files      |   86.36 |   80.56 |
 http.mjs      |  100.00 |  100.00 |
 live.mock.mjs |   72.73 |   61.11 | 3-5,22-25
---------------|---------|---------|-------------------

 8 pass
 0 fail
Ran 8 tests across 1 files. [51.00ms]
#
```

```
  1 export class MockLive {
  2
  3     calls = []
  4
  5     async consume(event) {
  6         this.calls.push({ consume: { event } })
  7     }
  8
...
```

### What is the expected behavior?

It should report that the consume method, defined on line 5-7, not on line 3-5, was marked as unchecked. Because it is not checked.

### What do you see instead?

The line numbers are shifted by 2.

PR #15889, opened 2024-12-20
perf(node/util): use string.split

DonIsaac:


robobun:
<div><sup>Updated 5:57 PM PT - Dec 19th, 2024</sup></div>

Your commit 7738d6642017d26289b797fa0b68ebd03156c934 is building: [`#8290`](https://buildkite.com/bun/bun/builds/8290)

src/js/internal/util/inspect.js
```diff
@@ -453,11 +436,14 @@ try {
 }
 
 const extractedSplitNewLines = value => {
-  if (typeof value === "string") {
-    return extractedSplitNewLinesFastPathStringsOnly(value) || extractedSplitNewLinesSlow(value);
+  const lines = typeof value === "string" ? StringPrototypeSplit(value, "\n") : RegExpPrototypeSymbolSplit(/\n/, value);
```
Jarred-Sumner: why continue to use REgExpPrototypeSymbolSplit?

src/js/internal/util/inspect.js
```diff
@@ -453,11 +436,14 @@ try {
...
+  const last = ArrayPrototypePop(lines);
+  const nlLines = ArrayPrototypeMap(lines, line => line + "\n");
```
Jarred-Sumner: Do we need to use map here? Can we for loop and modify the strings in-place?

src/js/internal/util/inspect.js
```diff
@@ -453,11 +436,14 @@ try {
 }
 
 const extractedSplitNewLines = value => {
-  if (typeof value === "string") {
-    return extractedSplitNewLinesFastPathStringsOnly(value) || extractedSplitNewLinesSlow(value);
+  const lines = typeof value === "string" ? StringPrototypeSplit(value, "\n") : RegExpPrototypeSymbolSplit(/\n/, value);
+  const last = ArrayPrototypePop(lines);
```
Jarred-Sumner: Do we need to pop the last element of the array and then push it again? or can we check `lines[lines.length - 1] === '\n'`

src/js/internal/util/inspect.js
```diff
@@ -453,11 +436,16 @@ try {
 }
 
 const extractedSplitNewLines = value => {
-  if (typeof value === "string") {
-    return extractedSplitNewLinesFastPathStringsOnly(value) || extractedSplitNewLinesSlow(value);
+  const lines = StringPrototypeSplit(value, "\n");
```
Jarred-Sumner, line 439: `pop()` is unnecessary and will potentially cause the array to resize. Reading the `lines.length` property is only needed once. Checking if the line is the last line on each iteration is unnecessary. 

```suggestion
  const lines = StringPrototypeSplit(value, "\n");
  let i = 0;
  for (let j = 0, length = lines.length; j < length; j++) {
      const line = lines[j];
     if (line !== "")
          lines[i++] = line + "\n";
  }
  lines.length = i;
  return lines;
```

PR #15896, opened 2024-12-20
build(bindgen): check for corresponding .zig file

DonIsaac:
When generating bindings for `foo.bind.ts`, ensure a `foo.zig` file exists next to it. Builds will fail downstream if this is missing, this error just makes it surface sooner and in a more clear manner.


robobun:
<div><sup>Updated 9:53 PM PT - Dec 19th, 2024</sup></div>

Your commit ff4acb181888c20794d4834102051dc0fec63161 is building: [`#8317`](https://buildkite.com/bun/bun/builds/8317)

src/codegen/bindgen-lib.ts
```diff
@@ -1,5 +1,9 @@
-// This is the public API for `bind.ts` files
-// It is aliased as `import {} from 'bindgen'`
+/**
```
paperclover, line 1: does jsdoc make this a comment for the file or the next declared symbol (Type)? i could be totally wrong and this form is correct.

src/codegen/bindgen-lib.ts
```diff
@@ -196,6 +200,23 @@ export namespace t {
   export const ByteString = builtinType<string>()("ByteString");
   /**
    * DOMString but encoded as `[]const u8`
+   *
+   * @example
```
paperclover: i think simple examples like these aren't worth their line count, especially when my laptop screen can only fit 70 lines of code at once. this is why i try to write extremely concise comments.

src/codegen/bindgen-lib.ts
DonIsaac: for `//`? no. It _should_ make one for `/** */` but I'm not seeing it. Maybe there's a problem with the tsconfig?

paperclover (approved):
thank you for adding the additional assertion

src/codegen/bindgen-lib.ts
DonIsaac: It makes it appear nicely when hovering over the symbol.
image

If this was ts-only, VSCode would still perform syntax highlighting.

DonIsaac: Ah the hover doesn't appear in screenshots :(

PR #15900, opened 2024-12-20
refactor: ♻️ move constants to top in `install.sh`

okineadev:


okineadev:
@robobun

```js
const response = await fetch("https://example.com/");

console.log("Status", response.status);
console.log("Headers", Object.fromEntries(response.headers.entries()));
```

okineadev:
why

PR #15899, opened 2024-12-20
Revert "fix(lexer): do not treat '#bun' in a url as a pragma"

Jarred-Sumner:
Reverts oven-sh/bun#15888

This should error:
```
#!/usr/bin/env bun
// @bun
export const foo: number = 123;
```

It doesn't:
```
❯ bun-15888 pragma-hash.ts
```

Expected:
```
❯ bun pragma-hash.ts
1 | #!/usr/bin/env bun
2 | // @bun
3 | export const foo: number = 123;
    ^
SyntaxError: Unexpected token ':'. const declared variable 'foo' must have an initializer.
      at <parse> (/Users/jarred/Code/bun/pragma-hash.ts:3:1)
      at native:11:43
note: missing sourcemaps for /Users/jarred/Code/bun/pragma-hash.ts
note: consider bundling with '--sourcemap' to get unminified traces

Bun v1.1.40-canary.1+7dac35944 (macOS arm64)
```

Issue #8355, opened 2024-01-22
Worker Segmentation fault's when connected to cassandra using "cassandra-driver"


Darker-Ink:
### What version of Bun is running?

1.0.25

### What platform is your computer?

Linux 5.15.90.1-microsoft-standard-WSL2 x86_64 unknown ( Microsoft Windows NT 10.0.22631.0 x64 )

### What steps can reproduce the bug?

When using workers and connecting to cassandra-driver, if the worker encounters an error and or exists, the parent will "Segmentation fault".

Parent: 

```ts
import { join } from "node:path";

const getUrl = (url: string): string => join(import.meta.dirname, url);
const api = new Worker(getUrl("./worker.ts"), { argv: process.argv });

api.onmessage = (event: MessageEvent) => {
    console.log(event.data);
}

api.onerror = (event: ErrorEvent) => {
    console.log(event);
}

setInterval(() => {}, 1_000) // stay alive
```

Worker:

```ts
import cassandra from "@kastelll/cassandra-driver";

try {
    const cas = new cassandra.Client({
        contactPoints: ["localhost"],
        localDataCenter: "datacenter1",
        credentials: {
            username: "",
            password: "",
        }
    });

    await cas.connect();
} catch {
    console.log("A fatal error occurred before the server could start.");
}

setTimeout(() => {
    throw new Error("Failing"); // or process.exit()
}, 2_000);
```

In the code to reproduce I'm using `@kastelll/cassandra-driver` which is a fork I made of `cassandra-driver` to fix a single issue (so it will work in bun: https://github.com/oven-sh/bun/issues/4033#issuecomment-1749318191 ) It shouldn't affect this issue though.

### What is the expected behavior?

The parent process shouldn't be killed / the worker shouldn't Segmentation fault while using cassandra-driver

### What do you see instead?

The parent process dies from "Segmentation fault" 

![image](https://github.com/oven-sh/bun/assets/45852757/dc41d5c4-db51-43f5-9d16-1dc7d7ac7de0)

PR #15912, opened 2024-12-20
fix(lexer): do not treat '#bun' in a url as a pragma

### LINKED ISSUE #15884: Code comments containing "#bun" causes Bun to not parse TypeScript

1.1.40+b5b51004e

Darwin 24.1.0 arm64 arm

Run this:

```ts
// https://bun.sh/docs/api/http#bun-serve
const a: string = "hello";
console.log(a);
```

Output should be:

```
hello
```

### What do you see instead?

SyntaxError is thrown:

```
1 | // https://bun.sh/docs/api/http#bun-serve
2 | const a: string = "hello";
    ^
SyntaxError: Unexpected token ':'. const declared variable 'a' must have an initializer.
      at <parse> (/Users/dtinth/Projects/elysia-by-example/hello.ts:2:1)
      at native:11:43
note: missing sourcemaps for /Users/dtinth/Projects/elysia-by-example/hello.ts
note: consider bundling with '--sourcemap' to get unminified traces

Bun v1.1.40 (macOS arm64)
```

### Additional information

The repro case can be reduced to:

```ts
// #bun
const a: string = "hello";
console.log(a);
```

Without the comment, it works:

```ts
const a: string = "hello";
console.log(a);
```

With the comment not containing `#bun`, it works:

```ts
// https://bun.sh/docs/api/http#fetch-request-handler
const a: string = "hello";
console.log(a);
```

With the comment containing `#bun` immediately followed by an alphanumeric character, it works:

```ts
// #bunlai
const a: string = "hello";
console.log(a);
```

DonIsaac:
### What does this PR do?
Fixes #15884.

Re-creation of #15888, which was reverted by @Jarred-Sumner in #15899 due to mishandling of hashbangs. This PR fixes that bug and adds many more tests that cover other edge cases.


robobun:
<div><sup>Updated 1:37 PM PT - Dec 20th, 2024</sup></div>

@DonIsaac, your commit f9fe79208c37f7630892e4c0bf3940456374925b is building: [`#8355`](https://buildkite.com/bun/bun/builds/8355)

DonIsaac:
CC @paperdave who should not be reading this right now because they have a flight to pack for.

Issue #8420, opened 2024-01-23
[windows] BroadcastChannel test crashes


Jarred-Sumner:
Seems to be infinite crash looping so I truncated thetrace 

```ts
bun-debug test test\js\web\broadcastchannel\broadcast-channel.test.ts
```

```ts
-------------------------------------------------------------------
You may only use the C/C++ Extension for Visual Studio Code
with Visual Studio Code, Visual Studio or Visual Studio for Mac
software to help you develop and test your applications.
-------------------------------------------------------------------
Loaded 'C:\bun\build\bun-debug.exe'. Symbols loaded.
Loaded 'C:\Windows\System32\ntdll.dll'.
Loaded 'C:\Windows\System32\kernel32.dll'.
Loaded 'C:\Windows\System32\KernelBase.dll'.
Loaded 'C:\Windows\System32\ucrtbase.dll'.
Loaded 'C:\Windows\System32\advapi32.dll'.
Loaded 'C:\Windows\System32\msvcrt.dll'.
Loaded 'C:\Windows\System32\sechost.dll'.
Loaded 'C:\Windows\System32\wsock32.dll'.
Loaded 'C:\Windows\System32\bcrypt.dll'.
Loaded 'C:\Windows\System32\ws2_32.dll'.
Loaded 'C:\Windows\System32\rpcrt4.dll'. 
Loaded 'C:\Windows\System32\rpcrt4.dll'.
Unloaded 'C:\Windows\System32\rpcrt4.dll'.
Loaded 'C:\Windows\System32\shell32.dll'.
Loaded 'C:\Windows\System32\msvcp_win.dll'.
Loaded 'C:\Windows\System32\user32.dll'.
Loaded 'C:\Windows\System32\msvcp140.dll'.
Loaded 'C:\Windows\System32\win32u.dll'.
Loaded 'C:\Windows\System32\gdi32.dll'.
Loaded 'C:\Windows\System32\gdi32full.dll'.
Loaded 'C:\Windows\System32\vcruntime140_1.dll'.
Loaded 'C:\Windows\System32\vcruntime140.dll'.
Loaded 'C:\Windows\System32\ole32.dll'.
Loaded 'C:\Windows\System32\combase.dll'.
Loaded 'C:\Windows\System32\vcruntime140.dll'.
Unloaded 'C:\Windows\System32\vcruntime140.dll'.
Loaded 'C:\Windows\System32\winmm.dll'.
[...]

PR #15944, opened 2024-12-22
+5 passing node:zlib tests

Jarred-Sumner:
### What does this PR do?

+5 passing node:zlib tests

- This fixes a memory leak in node:zlib caused by keeping strong references to the callbacks and the `this` value when it wasn't necessary
- This fixes the callbacks being associated with the wrong class (!)
- This fixes a memory leak that could potentially occur where the stream was initialized but never freed (or never closed)

Ran existing node tests and verified they pass

### DIFF
--- a/src/bun.js/node/node_zlib_binding.zig
+++ b/src/bun.js/node/node_zlib_binding.zig
@@ -239,7 +239,6 @@ pub fn CompressionStream(comptime T: type) type {
                 this.pending_close = true;
                 return;
             }
-            this.poll_ref.unref(JSC.VirtualMachine.get());
             this.pending_close = false;
             this.closed = true;
             this.this_value.deinit();
--- a/src/bun.js/node/node_zlib_binding.zig
+++ b/src/bun.js/node/node_zlib_binding.zig
@@ -6,6 +6,7 @@ const string = bun.string;
 const Output = bun.Output;
 const ZigString = JSC.ZigString;
 const validators = @import("./util/validators.zig");
+const debug = bun.Output.scoped(.zlib, true);
 
 pub fn crc32(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) bun.JSError!JSC.JSValue {
     const arguments = callframe.arguments_old(2).ptr;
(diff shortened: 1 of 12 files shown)


robobun:
<div><sup>Updated 8:22 AM PT - Dec 22nd, 2024</sup></div>

@Jarred-Sumner, your commit a62df5cdf9342c508a7c347d5c0d16156a23d362 is building: [`#8420`](https://buildkite.com/bun/bun/builds/8420)

test/js/node/test/parallel/test-internal-module-require.js
```diff
@@ -1,112 +1,114 @@
...
+// //    is on.
+// // 2. We do not accidentally leak any modules to the public global scope.
+// // 3. Deprecated modules are properly deprecated.
 
 const common = require('../common');
 
-if (!common.isMainThread) {
-  common.skip('Cannot test the existence of --expose-internals from worker');
-}
+common.skip("This test is not going to be implemented in Bun. We do not support --expose-internals.")
```
nektro, line 12: `common.skip` exits the process so the commenting out below here is unnecessary btw

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -662,16 +671,14 @@ pub const NativeBrotli = JSC.Codegen.JSNativeBrotli.getConstructor;
 
 pub const SNativeBrotli = struct {
     pub usingnamespace bun.NewRefCounted(@This(), deinit);
-    pub usingnamespace JSC.Codegen.JSNativeZlib;
+    pub usingnamespace JSC.Codegen.JSNativeBrotli;
```
nektro, line 674: ! nice find

src/bun.js/node/node_zlib_binding.zig
```diff
@@ -759,9 +769,12 @@ pub const SNativeBrotli = struct {
...
 
     pub fn deinit(this: *@This()) void {
-        this.write_callback.deinit();
-        this.onerror_value.deinit();
+        this.this_value.deinit();
         this.poll_ref.deinit();
+        switch (this.stream.mode) {
+            .BROTLI_ENCODE, .BROTLI_DECODE => this.stream.close(),
+            else => {},
+        }
```
nektro, line 777:
```suggestion
        this.stream.close();
```

nektro (approved):
nice

src/bun.js/node/node_zlib_binding.zig
Jarred-Sumner: There's an assertion in there related to .mode so I was doing this to be cautious

Issue #8428, opened 2024-01-23
bun unable to create astro project


ghost:
### What version of Bun is running?

1.0.25+a8ff7be64

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

`bunx create-astro@latest` and `bun create astro`

### What is the expected behavior?

https://bun.sh/guides/ecosystem/astro

### What do you see instead?

`error: could not determine executable to run for package create-astro`

### Additional information

Was working in Bun 1.0.23

ghost:
Also not able to run `bunx @astrojs/upgrade`. got: 
```
node:internal/modules/esm/resolve:264
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/private/tmp/@astrojs/upgrade@latest--bunx/node_modules/@astrojs/upgrade/dist/index.js' imported from /private/tmp/@astrojs/upgrade@latest--bunx/node_modules/@astrojs/upgrade/upgrade.mjs
    at finalizeResolution (node:internal/modules/esm/resolve:264:11)
    at moduleResolve (node:internal/modules/esm/resolve:917:10)
    at defaultResolve (node:internal/modules/esm/resolve:1130:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:396:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:365:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:240:38)
    at ModuleLoader.import (node:internal/modules/esm/loader:328:34)
    at importModuleDynamically (node:internal/modules/esm/translators:158:35)
    at importModuleDynamicallyCallback (node:internal/modules/esm/utils:207:14)
    at file:///private/tmp/@astrojs/upgrade@latest--bunx/node_modules/@astrojs/upgrade/upgrade.mjs:15:1 {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///private/tmp/@astrojs/upgrade@latest--bunx/node_modules/@astrojs/upgrade/dist/index.js'
}
```

PR #15951, opened 2024-12-23
Make originalLine and originalColumn getter calls not observable

Jarred-Sumner:
### What does this PR do?

- We should do our best to avoid calling into JS inside error handlers when possible
- node:util.inspect is displaying originalLine and originalColumn which messes with some node tests

Updated tests


robobun:
<div><sup>Updated 9:46 PM PT - Dec 22nd, 2024</sup></div>

@Jarred-Sumner, your commit 05f7057cc5f9af613c4fba656ee397c5ae7e7858 is building: [`#8428`](https://buildkite.com/bun/bun/builds/8428)

PR #15960, opened 2024-12-23
docs: Fix port number in http.md

MoE1N:
### What does this PR do?

Change to `docs/api/http.md`: Change the port number in one of the code examples where it was different from what is expected in the comment and the rest of the examples.

- [ ] Code changes


DonIsaac:
Thanks for the PR, but I'm not convinced this change is that useful. Can you please explain how this is clearer?

MoE1N:
> Thanks for the PR, but I'm not convinced this change is that useful. Can you please explain how this is clearer?

No problem!
This is not critical, it's just a mismatch of the set default port number in the example code (`8080`) vs what is expected from the comment (`3000`). 

```
  port: 8080, // defaults to $BUN_PORT, $PORT, $NODE_PORT otherwise 3000
```

This change makes sure the set port number is consistent with the comment and the rest of the codes in the documentation.

nektro:
the comment specifies the default, the code is showing you may set it to whichever port you desire

PR #15982, opened 2024-12-24
Add documentation for threadsafe option in FFI JSCallback

### LINKED ISSUE #15925: Bun FFI `JSCallback` fails when called across threads in the same process

This issue arose while consuming the [Webview](https://github.com/webview/webview) foreign library.

Webview opens a native OS browser window (webview), blocking the calling (Worker) thread until terminated. In order to interact with the Webview API after a browser window is opened, calls need to be made from the parent thread.

The [bun minimal reproduction](https://gist.github.com/citkane/98e562ec53c1691912d929ec354da58a) illustrates how Webview API commands using `JSCallback` work from the worker thread, but fail from the parent thread.

Observations:
- The [Deno minimal reproduction](https://gist.github.com/citkane/b25855b2b99679ae0540e5efedb2b5d0) works as expected for all Webview API commands from both threads.
- In Deno, the minimal reproduction works as expected using "node:worker_threads"
- The Bun minimal reproduction fails with native `Worker` and "node:worker_threads"
- Webview FFI commands not involving `JSCallback` work as expected from the parent thread.

I have thus formed an opinion that the issue does not lie with Bun `Worker`, but specifically with Bun FFI `JSCallback`.

In order to run the minimal reproduction (for linux x64) , [some OS dependencies are required](https://github.com/webview/webview?tab=readme-ov-file#packages).
[minimal.repro.zip](https://github.com/user-attachments/files/18218757/minimal.repro.zip) contains all of the minimal reproduction code, as well as "libwebview.so" (the Webview library for linux x64)

### Relevant log output

```shell
Overriding existing handler for signal 10. Set JSC_SIGNAL_FOR_GC if you want WebKit to use a different signal
Bun v1.1.39 (1d485617) Linux x64
Linux Kernel v6.1.0 | glibc v2.36
CPU: sse42 popcnt avx avx2
Args: "bun" "./index.ts"
Features: jsc tsconfig workers_spawned 
Builtins: "bun:ffi" "bun:main" "node:path" 
Elapsed: 1229ms | User: 152ms | Sys: 114ms
RSS: 1.07GB | Peak: 0.16GB | Commit: 1.07GB | Faults: 0

panic(main thread): Segmentation fault at address 0x40

 https://bun.report/1.1.39/la11d48561Ihgg8//DoqqomEk0/pvB__A2AgE

Illegal instruction
```

### Stack Trace (bun.report)

Bun v1.1.39 ([`1d48561`](<https://github.com/oven-sh/bun/tree/1d4856170978abe3aa47960e2fd7d4f400423808>)) on linux x86_64 [AutoCommand]

Segmentation fault at address 0x00000040

- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/1d4856170978abe3aa47960e2fd7d4f400423808/ld-temp.o#L0>): `icudt75_dat`

Features: workers\_spawned, jsc, tsconfig, tsconfig

citkane:
### What does this PR do?

It documents the `threadsafe` option for FFI `JSCallback`.
It resolves #15925 

- [ ] Code changes


DonIsaac (approved):

docs/api/ffi.md
```diff
@@ -280,6 +280,7 @@ const searchIterator = new JSCallback(
   {
     returns: "bool",
     args: ["ptr", "usize"],
+    threadsafe: true // Optional. Defaults to `false`
```
190n: Thanks for addressing missing documentation. I think it'd be better to describe `threadsafe` in a separate section, and explain what it does rather than just demonstrating that it is an option that exists. The `threadsafe` option isn't necessary in this example so it could be confusing to see it set to `true`.

I think we should also state that `threadsafe` is experimental for now since there are cases where it doesn't 100% work.

190n (changes requested):

docs/api/ffi.md
citkane: That may be true, but setting up a new documentation structure for Bun is out of scope for me as a user. I am not a Bun developer.

All I can add is to articulate my user experience of being frustrated for many days over this issue before I found the `threadsafe` flag in code. I have grown to like Bun,  but it was causing a roadblock for my scenario in this instance, and I was relieved that it was only a documentation issue. 

The documentation quickfix I am proposing may save much similar frustration in Bun userland.

190n: > setting up a new documentation structure for Bun is out of scope for me as a user. I am not a Bun developer.

Sorry, I didn't mean add a whole new page or anything. I just meant a new paragraph or subheading under "Callbacks" on the existing `ffi.md`, rather than putting `threadsafe` in the existing example.

> All I can add is to articulate my user experience of being frustrated for many days over this issue before I found the threadsafe flag in code. I have grown to like Bun, but it was causing a roadblock for my scenario in this instance, and I was relieved that it was only a documentation issue.
>
> The documentation quickfix I am proposing may save much similar frustration in Bun userland.

I understand, and I'm sorry we hadn't documented this option before. I still think it's worth ensuring that the documentation is done right when we do add it.

citkane: PR has been updated.

190n (approved):

docs/api/ffi.md
190n: Thank you! Looks good now.

190n:
Sorry @citkane, I don't know why this didn't merge. I'll try running CI again. I don't think you need to be updating this against `main` as there aren't any conflicts. I think there is just some issue with our CI and contributor PRs.

citkane:
It was my initial bad for not updating my fork before committing the change.
Chaos ensued...

I see that all the checks and approvals have now passed, but still no merge. I will leave it to you to push the right buttons.

Issue #14104, opened 2024-09-23
Bun build stucks with async reading file in macro


sirenkovladd:
### What version of Bun is running?

1.1.29+6d43b3662

### What platform is your computer?

Darwin 23.6.0 arm64 arm

### What steps can reproduce the bug?

```ts
Bun.write(
  "randomFile",
  '123'
);

Bun.write(
  "content.ts",
  // works
  // 'export function content() { return require("node:fs").readFileSync("./randomFile", "utf8"); }',

  // stucks
  'export function content() { return Bun.file("./randomFile").text(); }'
);
Bun.write(
  "test.ts",
  'import { content } from "./content" with { type: "macro" };\n console.log(await content())'
);

await Bun.$`bun test.ts`
console.log('\n')
await Bun.$`bun build --minify test.ts`

await Bun.$`rm -f randomFile content.ts test.ts`
```

### What is the expected behavior?

```sh
❯ bun index.ts
123


console.log(await"123");
```

### What do you see instead?

```sh
❯ bun index.ts
123


^C
```

sirenkovladd:
it doesn't work with this code either
```ts
export async function content() { return await require("node:fs/promises").readFile("./randomFile", "utf8"); }
```

so it looks like the problem is with asynchronous file reading

brainkim:
Probably a duplicate of https://github.com/oven-sh/bun/issues/7611

PR #15991, opened 2024-12-26
Fix macros which call Bun.file operations causing build commands to hang

### LINKED ISSUE #7611: Build macros hang on Bun.file(...).text()

v1.0.17

```bash
git clone git@github.com:mckamyk/bun-bundle-macro-bug
cd bun-bundle-macro-bug
bun build ./index.ts
```

As per the `macro.ts` script, should see the following console logs:
```bash
Entered getContent
test.txt exists: true
Content length: 35
```

### What do you see instead?

The bundler seems to hang on the `file.text()` line, printing the follwing
```bash
Entered getContent
test.txt exists: true
# just hangs here forever
```

### Additional information

I also attempted this on the `Linux 5.15.0-1050-aws x86_64 x86_64` kernel, same result.

On my actual production repo, calling
```js
Bun.build({
  entrypoints: ['./my/entrypoint/file/with/macro.ts']
  ...
})
```
also had the same issue, though it seem in the reproduction code, calling `Bun.build({...})` doesn't seem to reproduce the error.

Also, it appears this issue started with `v1.0.16`. Prior to this version, I don't seem to be having an issue.

brainkim:
### What does this PR do?

- [x] Code changes

Users reported in https://github.com/oven-sh/bun/issues/7611 and https://github.com/oven-sh/bun/issues/14104 that calling `Bun.file().text()` in a macro would cause builds to hang. This bug was introduced in v1.0.16, likely when some changes related to how Bun handles multithreading occurred.

I went through a minimal reproduction with a debugger and determined that work would get stuck in ThreadPool threads because the ThreadPool was initialized twice. The double instantiation of ThreadPool, which is assumed to be a global singleton, caused issues in macros because file I/O would get scheduled on a thread, the work would be pushed to the thread local queue/buffer, and then the thread would wait for the promise to resolve, causing a dead lock. There is code which allows for “work stealing” but threads initialized from different thread pools do not steal work each other.

The least invasive fix I found was to consistently call `WorkPool.get()` in the bundle_v2’s initialization code.

Fixes #7611 and fixes #14104.
Happy holidays!

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -1758,7 +1758,7 @@ pub fn NewPackageInstall(comptime kind: PkgInstallKind) type {
             var queue: Queue = undefined;
             pub fn getQueue() *Queue {
                 queue = Queue{
-                    .thread_pool = &PackageManager.get().thread_pool,
+                    .thread_pool = PackageManager.get().thread_pool,
                 };
                 return &queue;
             }
@@ -2687,7 +2687,7 @@ pub const PackageManager = struct {
         }
     } = .{},
 
-    thread_pool: ThreadPool,
+    thread_pool: *ThreadPool,
     task_batch: ThreadPool.Batch = .{},
     task_queue: TaskDependencyQueue = .{},
 
@@ -8982,9 +8982,7 @@ pub const PackageManager = struct {
             .root_dir = entries_option.entries,
             .env = env,
             .cpu_count = cpu_count,
-            .thread_pool = ThreadPool.init(.{
-                .max_threads = cpu_count,
-            }),
+            .thread_pool = JSC.WorkPool.get(),
             .resolve_tasks = .{},
             .lockfile = undefined,
             .root_package_json_file = root_package_json_file,
@@ -9152,9 +9150,7 @@ pub const PackageManager = struct {
             .root_dir = root_dir.entries,
             .env = env,
             .cpu_count = cpu_count,
-            .thread_pool = ThreadPool.init(.{
-                .max_threads = cpu_count,
-            }),
+            .thread_pool = JSC.WorkPool.get(),
             .lockfile = undefined,
             .root_package_json_file = undefined,
             .event_loop = .{
--- a/src/thread_pool.zig
+++ b/src/thread_pool.zig
@@ -440,15 +440,17 @@ pub fn warm(self: *ThreadPool, count: u14) void {
     while (sync.spawned < to_spawn) {
         var new_sync = sync;
         new_sync.spawned += 1;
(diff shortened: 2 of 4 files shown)


brainkim:
I guess things are never this easy...

Jarred-Sumner:
My guess is that by not using the bundler’s threadpool, the threads never call their init function

it sounds like we should have a global variable for “threadpool ID” and only use the global one in that case maybe?

brainkim:
I tried building on ubuntu and the failing test passes locally. I also tested it by hand.

Is it possible that the CI is running the tests against a stale build? I’m not sure how this failure could be platform dependent. Perhaps it is due to the number of threads available?

Running the build process in a container takes a painfully long time, wishing for cross-compilation here.

Screenshot 2024-12-27 at 7 44 59 PM

brainkim:
Sorry about the failures. The additional Windows failure may indicate that this is a race condition of some sort. I searched the codebase for additional direct uses of `ThreadPool.init()` and found one in `installer.zig`, which might be getting called during the test running process. Beyond that, I’m stumped unless I figure out how to reproduce the test failure locally.

Again, sorry about the test failures!

brainkim:
@Jarred-Sumner
Took another look at this and I think I figured out the problem. The platform differences are likely due to the `max_threads` being set to the number of logical cores available for the work_pool, which is probably 2 for the linux environments. The deadlock in the test suite happened when the `max_threads` was set to anything less than 3.

This would have been fine, except there’s a bug in `ThreadPool.warm` function:

https://github.com/oven-sh/bun/blob/7830e156505b3d00397361df0acdfb80bfb1474c/src/thread_pool.zig#L416-L435

In short, `cmpxchngWeak` will return null on success, but this code breaks on null, meaning the `spawned` property is incremented without actually spawning any new threads, causing an off-by-one error when spawning new threads normally. This, combined with the task dependency created by macros, where the parsing thread is blocked until the macro finishes, caused the deadlock at two threads. You can confirm this caused the bug by commenting out the `warm()` call here (https://github.com/oven-sh/bun/blob/7830e156505b3d00397361df0acdfb80bfb1474c/src/bundler/bundle_v2.zig#L169). I assumed that there are performance reasons for warming the thread pool so I fixed the function instead.

brainkim:
Fixed a dumb pointer error which was caught by the Windows build. The test failures on other platforms look like flake (s3, fetch failures).

brainkim:
Rebased against the latest main. May I get another test run? The previous run seems to have failed for things that I think are unrelated to the code. And the tests related to the change pass.

PR #16013, opened 2024-12-27
fix formatting of Set in Bun.inspect()

### LINKED ISSUE #16007: `Bun.inspect()` has strange formatting when showing a `Set` inside an object

1.1.42+50eec0025

Darwin 23.6.0 arm64 arm

The following code produces strange formatting if a set is nested inside an object
```ts
const map = new Map([
  ["foo", "bar"],
  ["baz", "qux"],
]); // for comparison
const set = new Set(["foo", "bar"]);
const bug = Bun.inspect({ map, set, foo: { bar: { set } } });
console.log(bug);
```

Expected output looks like for map. 
```shell
❯ bun min.ts
{
  map: Map(2) {
    "foo": "bar",
    "baz": "qux",
  },
  set:   Set(2) {
    "foo",
    "bar",
},
  foo: {
    bar: {
      set:       Set(2) {
        "foo",
        "bar",
},
    },
  },
}
```

### What do you see instead?

2 extra spaces between the field name and `Set` for each nesting layer of the object.
also the closing } is not indented.

### Additional information

_No response_

laesse:
### What does this PR do?

Fixes #16007

- [x] Code changes

Minimal reproduction from bug now shows expected output: 
```shell
{
  map: Map(2) {
    "foo": "bar",
    "baz": "qux",
  },
  set: Set(2) {
    "foo",
    "bar",
  },
  foo: {
    bar: {
      set: Set(2) {
        "foo",
        "bar",
      },
    },
  },
}
```

I wrote automated tests `test/regression/issue/16007.test.ts`


Jarred-Sumner:
Thank you

Issue #15953, opened 2024-12-23
Unnecessray file permissions on `bun.lock`


LitoMore:
### What version of Bun is running?

1.1.42

### What platform is your computer?

Darwin 24.2.0 arm64 arm

### What steps can reproduce the bug?

```shell
bun init
bun install --save-text-lockfile
```

<img width="451" alt="Image" src="https://github.com/user-attachments/assets/36a5164b-3636-4bec-8d3f-a36e10559c87" />

### What is the expected behavior?

The file should be `-rw-r--r--`

### What do you see instead?

The file mode is `-rwxrwxrwx`.

### Additional information

The `-rwxrwxrwx` permission is unnecessary since it's just a text file.

jbergstroem:
Looks intentional: https://github.com/oven-sh/bun/blob/608101c9751c12d5d1e361c6304584f69e20b9b9/src/install/lockfile.zig#L2380-L2391

Jarred-Sumner:
its a hold over from `bun.lockb`, which is intentionally executable

LitoMore:
Even the the `bun.lockb` has some unnecessary permssions. For the `bun.lockb`, it should be `-rwxr-xr-x` rather than `-rwxrwxrwx`, IMO.

But this is another topic.

PR #16018, opened 2024-12-27
fix: avoid world-writable permissions for lockfiles

jbergstroem:
### What does this PR do?

Use the file mode `0x644` for the plaintext lockfile. Also, switch the permission for the binary lockfile to not default to group/world writable (`0x777` -> `0x755`). Windows has a different behavior, so we test for that.

Finally, add tests for writing plaintext lockfiles and ensure its file stat properties.

Built and tested locally.

Closes: https://github.com/oven-sh/bun/issues/15953

### DIFF
--- a/src/install/lockfile.zig
+++ b/src/install/lockfile.zig
@@ -2423,8 +2423,12 @@ pub fn saveToDisk(this: *Lockfile, save_format: LoadResult.LockfileFormat, verbo
     }
 
     if (comptime Environment.isPosix) {
-        // chmod 777 on posix
-        switch (bun.sys.fchmod(file.handle, 0o777)) {
+        // chmod 755 for binary, 644 for plaintext
+        var filemode: bun.Mode = 0o755;
+        if (save_format == .text) {
+            filemode = 0o644;
+        }
+        switch (bun.sys.fchmod(file.handle, filemode)) {
             .err => |err| {
                 file.close();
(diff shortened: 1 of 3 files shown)


jbergstroem:
The codebase is new to me; can't seem to find tests around the plaintext lockfile. Should I add one that does something similar as the `lockb` test?

dylan-conway:
@jbergstroem thanks for opening this pr. For `bun.lock` tests, could you create a new file in `test/cli/install`, maybe `bun.lock.test.ts` or something similar? Ideally the test installs a local folder dependency, to avoid flakiness from requesting from a registry

jbergstroem:
It seems windows permissions are different than the rest. Need to set up a local environment and figure it out.

dylan-conway:
@jbergstroem I think it would be okay to add:
```ts
import { isWindows } from "harness";
...
if (!isWindows) {
    // check permissions
}
```
Or hardcode it to the received value from CI (33206). We're only changing permissions on posix.

jbergstroem:
> Or hardcode it to the received value from CI (33206). We're only changing permissions on posix.

I kept the test for windows so we pick up if permissions start changing on windows for some reason.

jbergstroem:
Failed tests looks unrelated

dylan-conway (approved):
Agreed, looks good

PR #14621, opened 2024-10-16
misc node:process fixes

nektro:
- fixes a bug where padding multiple `"ipc"` options to `node:child_process.spawn` stdio option did not throw an error
- add definition and implementation for `process.availableMemory()`
- fixes a bug where passing `process.chdir` a non-string non-undefined argument would not throw an error
- add definition and implementation for `process.constrainedMemory()`
- fixes error handling of `process.cpuUsage()`
- fixes error handling of `process.hrtime()`
- fixes a bug where `process.exit()` was not calling `process.reallyExit()`
- fixes fields of `process.release`; removes `.lts`, fills out proper content of `.headersUrl`, removes bad field `.libUrl`
- add definition and implementation for `process.setegid()`, `.seteuid()`, `.setgid()`, `.setuid()`, `.setgroups()`

Issue #8624, opened 2024-02-01
`Function.prototype.toString()` omits function name, makes lighthouse fail


hilja:
### What version of Bun is running?

1.0.25

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

Make a file:

```js
// test.js
function fooBar() {
  return 'bar'
}

function foo(fn) {
  console.log(fn.toString())
}

foo(fooBar)
```

Run it:

```sh
$ node test.js
function fooBar() {
  return 'bar'
}
```

```sh
$ bun test.js
function() {
  return "bar";
}
```

### What is the expected behavior?

String with the function name:

```
function fooBar() {
  return 'bar'
}
```

### What do you see instead?

Function name omitted:

```
function() {
  return "bar";
}
```

### Additional information

Lighthouse performance metric package fails here https://github.com/GoogleChrome/lighthouse/blob/5e18c5a0656b427e59890dc9c125164ef9f276c3/core/lib/page-functions.js#L578

They have a util like this:

```js
/**
 * @param {Function} fn
 * @return {string}
 */
function getRuntimeFunctionName(fn) {
  const match = fn.toString().match(/function ([\w$]+)/);
  if (!match) throw new Error(`could not find function name for: ${fn}`);
  return match[1];
}
```

Bun can't find the match and errors out:

```
576 |  */
577 | function getRuntimeFunctionName(fn) {
578 |   const match = fn.toString().match(/function ([\w$]+)/);
579 |   if (!match) throw new Error(`could not find function name for: ${fn}`);
                          ^
error: could not find function name for: function(string, characterLimit) {
  return Util.truncate(string, characterLimit);
}
      at getRuntimeFunctionName (/Users/bob/web/foo/node_modules/.pnpm/lighthouse@11.5.0/node_modules/lighthouse/core/lib/page-functions.js:581:21)
      at /Users/bob/web/foo/node_modules/.pnpm/lighthouse@11.5.0/node_modules/lighthouse/core/lib/page-functions.js:589:13
```

bjon:
Just stumbled upon this too.

Chrome does the same as Node
![image](https://github.com/oven-sh/bun/assets/280413/1174cf77-a284-4f35-a360-8048dc1288ba)

hilja:
> Since ES2018, the spec requires the return value of toString() to be the exact same source code as it was declared, including any whitespace and/or comments

[...]

PR #16026, opened 2024-12-28
node: fix the rest of test-process

nektro:
supercedes https://github.com/oven-sh/bun/pull/14621
fixes all the tests that don't depend on missing behavior in `node:child_process` or `node:v8`
a few missing for `process.env` that will get a separate followup

not done, pushing now so the robobun comment is at the top


robobun:
<div><sup>Updated 10:26 PM PT - Dec 27th, 2024</sup></div>

@nektro, your commit 2ebba0a1ad438b91e1b92c174a668f1e3db161aa is building: [`#8624`](https://buildkite.com/bun/bun/builds/8624)

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -463,51 +472,56 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionDlopen,
...
-
+    auto& vm = globalObject->vm();
     auto* process = jsCast<Process*>(globalObject->processObject());
     MarkedArgumentBuffer arguments;
     arguments.append(jsNumber(exitCode));
-    process->wrapped().emit(Identifier::fromString(globalObject->vm(), "beforeExit"_s), arguments);
+    Bun__VirtualMachine__exitDuringUncaughtException(bunVM(vm));
+    auto fired = process->wrapped().emit(Identifier::fromString(vm, "beforeExit"_s), arguments);
+    if (fired) {
+        if (auto ntq = globalObject->m_nextTickQueue) {
```
Jarred-Sumner: This copies the `WriteBarrier`, which you didn't mean to do.
```suggestion
        if (globalObject->m_nextTickQueue) {
```

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -463,51 +472,56 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionDlopen,
...
+            auto nextTickQueue = jsDynamicCast<JSNextTickQueue*>(ntq.get());
```
Jarred-Sumner:
```suggestion
            auto nextTickQueue = jsDynamicCast<JSNextTickQueue*>(globalObject->m_nextTickQueue.get());
```

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -522,58 +536,101 @@ extern "C" void Process__dispatchOnExit(Zig::GlobalObject* globalObject, uint8_t
...
+
+    setProcessExitCodeInner(globalObject, process, code);
+    RETURN_IF_EXCEPTION(throwScope, {});
+
+    auto exitCode = Bun__getExitCode(bunVM(zigGlobal));
     Process__dispatchOnExit(zigGlobal, exitCode);
-    Bun__Process__exit(zigGlobal, exitCode);
+
+    // process.reallyExit(exitCode);
+    auto reallyExitVal = process->get(globalObject, Identifier::fromString(vm, "reallyExit"_s));
```
Jarred-Sumner, line 562: Can we track if `reallyExit` was created and skip calling it if it wasn't?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -522,58 +536,101 @@ extern "C" void Process__dispatchOnExit(Zig::GlobalObject* globalObject, uint8_t
...
+    Bun::V::validateString(throwScope, globalObject, id_value, "id"_s);
+    RETURN_IF_EXCEPTION(throwScope, {});
+
+    auto id_str = id_value.toWTFString(globalObject);
+    RETURN_IF_EXCEPTION(throwScope, {});
+
+    auto is_valid_bultin_name = Bun::isBuiltinModule(id_str);
+    if (!is_valid_bultin_name) return JSValue::encode(jsUndefined());
+
+    auto require_key = id_value.toString(globalObject);
```
Jarred-Sumner: We shouldn't call `toString` on the id_value twice.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -522,58 +536,101 @@ extern "C" void Process__dispatchOnExit(Zig::GlobalObject* globalObject, uint8_t
...
+        // but CommonJSModuleRecord.cpp calls it so it seems likely to be necessary, otoh the tests work without it /shrug
+        // globalObject->requireMap()->set(globalObject, require_key, moduleObject);
     }
-    auto* zigGlobal = defaultGlobalObject(globalObject);
-    jsCast<Process*>(zigGlobal->processObject())->setUncaughtExceptionCaptureCallback(arg0);
+
+    auto fn = globalObject->requireFunctionUnbound();
+    MarkedArgumentBuffer args;
+    args.append(id_value);
+    return JSValue::encode(JSC::profiledCall(globalObject, ProfilingReason::API, fn, JSC::getCallData(fn), moduleObject, args));
```
Jarred-Sumner: Calling require on it is unnecessary work, and also means that the user can override the exports.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -522,58 +536,101 @@ extern "C" void Process__dispatchOnExit(Zig::GlobalObject* globalObject, uint8_t
...
+    auto throwScope = DECLARE_THROW_SCOPE(vm);
+
+    auto id_value = callFrame->argument(0);
+    Bun::V::validateString(throwScope, globalObject, id_value, "id"_s);
+    RETURN_IF_EXCEPTION(throwScope, {});
+
+    auto id_str = id_value.toWTFString(globalObject);
+    RETURN_IF_EXCEPTION(throwScope, {});
+
+    auto is_valid_bultin_name = Bun::isBuiltinModule(id_str);
```
Jarred-Sumner: This is a similar problem as in the validators causing performance regressions.

Instead of a validator which returns a boolean and then again is later validated, we should make this return a `std::variant<std::nullopt_t, InternalModuleRegistry::Field, NativeTag>`

```suggestion
    auto is_valid_bultin_name = Bun::isBuiltinModule(id_str);
```

When that `variant` returns `nullopt_t`, we know it's invalid
When that `variant` returns an `InternalModuleRegistry::Field`, we know it's a JS builtin
Otherwise we know it's a function to require and we call require on that.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -1127,40 +1205,89 @@ JSC_DEFINE_HOST_FUNCTION(Process_emitWarning, (JSGlobalObject * lexicalGlobalObj
...
-    if (callFrame->argumentCount() < 1) {
-        throwVMError(globalObject, scope, "Not enough arguments"_s);
-        return {};
+    auto warning = callFrame->argument(0);
+    auto type = callFrame->argument(1);
+    auto code = callFrame->argument(2);
+    auto ctor = callFrame->argument(3);
+    auto detail = jsUndefined();
+
+    auto dep_warning = jsString(vm, String("DeprecationWarning"_s));
```
Jarred-Sumner, line 1180: Please don't create jsString unnecessarily.

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -1127,40 +1205,89 @@ JSC_DEFINE_HOST_FUNCTION(Process_emitWarning, (JSGlobalObject * lexicalGlobalObj
...
-        err->putDirect(vm, vm.propertyNames->name, jsString(vm, String("warn"_s)), JSC::PropertyAttribute::DontEnum | 0);
-        return err;
-    })();
+        type = type.get(globalObject, vm.propertyNames->type);
+        RETURN_IF_EXCEPTION(scope, {});
+        if (!type.toBoolean(globalObject)) type = jsString(vm, String("Warning"_s));
+    } else if (type.isCallable()) {
+        ctor = type;
+        code = jsUndefined();
+        type = jsString(vm, String("Warning"_s));
```
Jarred-Sumner, line 1203: ditto

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2154,6 +2309,167 @@ JSC_DEFINE_HOST_FUNCTION(Process_functiongetgroups, (JSGlobalObject * globalObje
...
+    if (!str.is8Bit()) {
+        auto message = makeString("User identifier does not exist: "_s, str);
+        throwScope.throwException(globalObject, createError(globalObject, ErrorCode::ERR_UNKNOWN_CREDENTIAL, message));
+        return {};
+    }
+
+    auto name = (const char*)(str.span8().data());
+    struct passwd pwd;
+    struct passwd* pp = nullptr;
+    char buf[8192];
```
Jarred-Sumner, line 2292: what is this buffer size based on? 8KB can be a lot

src/bun.js/bindings/BunProcess.cpp
heimskr: I think [`std::monostate`](https://en.cppreference.com/w/cpp/utility/variant/monostate) instead of `std::nullopt_t` is more idiomatic to indicate an empty `std::variant`

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2400,34 +2709,28 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionCpuUsage,
                 userValue = comparator->getDirect(0);
                 systemValue = comparator->getDirect(1);
             } else {
-                userValue = comparator->getIfPropertyExists(globalObject, JSC::Identifier::fromString(vm, "user"_s));
-                RETURN_IF_EXCEPTION(throwScope, JSC::JSValue::encode(JSC::jsUndefined()));
+                userValue = comparator->get(globalObject, JSC::Identifier::fromString(vm, "user"_s));
```
Jarred-Sumner: is it because of proxies? what's the rationale?

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -51,6 +52,24 @@ JSC::EncodedJSValue V::validateInteger(JSC::ThrowScope& scope, JSC::JSGlobalObje
 
     return JSValue::encode(jsUndefined());
 }
+JSC::EncodedJSValue V::validateInteger(JSC::ThrowScope& scope, JSC::JSGlobalObject* globalObject, JSC::JSValue value, ASCIILiteral name, JSC::JSValue min, JSC::JSValue max)
```
Jarred-Sumner, line 54: None of the usages of this warrant using `JSC::JSValue` for `min` or `max`. Let's avoid the extra work and pass an int32_t instead.

And, instead of returning a JSC::EncodedJSValue, we should return `int32_t`

or you could use the IDL converters for this

src/bun.js/bindings/bindings.cpp
```diff
@@ -1947,26 +1947,23 @@ JSC__JSValue SystemError__toErrorInstance(const SystemError* arg0,
 
     JSC::JSValue options = JSC::jsUndefined();
 
-    JSC::JSObject* result = JSC::ErrorInstance::create(globalObject, globalObject->errorStructureWithErrorType<JSC::ErrorType::Error>(), message, options);
+    JSC::JSObject* result = JSC::ErrorInstance::create(globalObject, JSC::ErrorInstance::createStructure(vm, globalObject, globalObject->errorPrototype()), message, options);
```
Jarred-Sumner: Why was this reverted?

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -405,7 +405,7 @@ namespace ERR {
 
 JSC::EncodedJSValue INVALID_ARG_TYPE(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject, const WTF::String& arg_name, const WTF::String& expected_type, JSC::JSValue val_actual_value)
 {
-    auto arg_kind = arg_name.startsWith("options."_s) ? "property"_s : "argument"_s;
+    auto arg_kind = arg_name.contains("."_s) ? "property"_s : "argument"_s;
```
Jarred-Sumner: Use `'.'` instead of `""_s` since it's a single byte
```suggestion
    auto arg_kind = arg_name.contains('.') ? "property"_s : "argument"_s;
```

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -420,7 +420,7 @@ JSC::EncodedJSValue INVALID_ARG_TYPE(JSC::ThrowScope& throwScope, JSC::JSGlobalO
 {
     auto arg_name = val_arg_name.toWTFString(globalObject);
     RETURN_IF_EXCEPTION(throwScope, {});
-    auto arg_kind = arg_name.startsWith("options."_s) ? "property"_s : "argument"_s;
+    auto arg_kind = arg_name.contains("."_s) ? "property"_s : "argument"_s;
```
Jarred-Sumner: Use `'.'` instead of `""_s` since it's a single byte
```suggestion
    auto arg_kind = arg_name.contains('.') ? "property"_s : "argument"_s;
```

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -574,6 +588,28 @@ JSC::EncodedJSValue SOCKET_BAD_PORT(JSC::ThrowScope& throwScope, JSC::JSGlobalOb
...
 }
 
+JSC::EncodedJSValue UNCAUGHT_EXCEPTION_CAPTURE_ALREADY_SET(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject)
+{
+    auto message = makeString("`process.setupUncaughtExceptionCapture()` was called while a capture callback was already active"_s);
+    throwScope.throwException(globalObject, createError(globalObject, ErrorCode::ERR_UNCAUGHT_EXCEPTION_CAPTURE_ALREADY_SET, message));
+    return {};
+}
+
+JSC::EncodedJSValue ASSERTION(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject, JSC::JSValue msg)
```
Jarred-Sumner, line 598: Instead of adding more functions, modify the one at the bottom to throw this message by default when `$ERR_ASSERTION` is passed through JS.

Jarred-Sumner: Assuming this was meant for JS, otherwise ignore me if not

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -725,6 +761,11 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_ERR_UNHANDLED_ERROR, (JSC::JSGlobalObject *
     return JSC::JSValue::encode(createError(globalObject, ErrorCode::ERR_UNHANDLED_ERROR, makeString("Unhandled error. ("_s, err_str, ")"_s)));
 }
 
+JSC_DEFINE_HOST_FUNCTION(jsFunction_ERR_IPC_ONE_PIPE, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
```
Jarred-Sumner: Instead of adding more functions, modify the one at the bottom to throw this message by default when `$ERR_IPC_ONE_PIPE,` is passed through JS.

src/bun.js/bindings/NodeValidator.cpp
```diff
@@ -16,6 +16,7 @@
 #include "JSBufferEncodingType.h"
 #include "BunProcess.h"
 #include "ErrorCode.h"
+#include "wtf/text/ASCIILiteral.h"
```
Jarred-Sumner:
```suggestion

```

src/bun.js/bindings/helpers.cpp
```diff
@@ -1,6 +1,260 @@
 #include "root.h"
 #include "helpers.h"
 #include "BunClientData.h"
+#include <string.h>
+
+const char* errno_string(int errorno)
```
Jarred-Sumner: Don't we already have all these strings in Zig? Wouldn't this be duplicating the strings in the binary? Couldn't we expose a function to get those strings instead?

src/bun.js/bindings/exports.zig
```diff
@@ -362,15 +362,17 @@ pub const Process = extern struct {
...
-        title.* = ZigString.init(_bun);
+        const str = bun.CLI.Bun__Node__ProcessTitle;
+        title.* = ZigString.init(str orelse "bun");
     }
 
     // TODO: https://github.com/nodejs/node/blob/master/deps/uv/src/unix/darwin-proctitle.c
-    pub fn setTitle(globalObject: *JSGlobalObject, _: *ZigString) callconv(.C) JSValue {
-        return ZigString.init(_bun).toJS(globalObject);
+    pub fn setTitle(globalObject: *JSGlobalObject, newvalue: *ZigString) callconv(.C) JSValue {
+        if (bun.CLI.Bun__Node__ProcessTitle) |_| bun.default_allocator.free(bun.CLI.Bun__Node__ProcessTitle.?);
```
Jarred-Sumner, line 378: This needs a mutex guarding it incase process.title is used in a Worker

src/sys.zig
```diff
@@ -505,58 +511,62 @@ pub fn chmod(path: [:0]const u8, mode: bun.Mode) Maybe(void) {
         Maybe(void).success;
 }
 
-pub fn chdirOSPath(destination: bun.OSPathSliceZ) Maybe(void) {
+pub fn chdirOSPath(path: bun.stringZ, destination: if (Environment.isPosix) bun.stringZ else bun.string) Maybe(void) {
```
Jarred-Sumner, line 514: Why was this changed to always be a stringZ? The string conversion is unnecessary on Windows

Jarred-Sumner (changes requested):
Some comments, let me know when ready for another pass

src/bun.js/bindings/BunProcess.cpp
nektro: https://github.com/nodejs/node/blob/v22.9.0/src/node_credentials.cc#L142-L145

src/bun.js/bindings/BunProcess.h
```diff
@@ -36,6 +36,7 @@ class Process : public WebCore::JSEventEmitter {
     }
 
     DECLARE_EXPORT_INFO;
+    bool m_reportOnUncaughtException;
```
Jarred-Sumner: Let's default initialize this to false to avoid undefined memory
```suggestion
    bool m_reportOnUncaughtException = false;
```

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -574,6 +588,28 @@ JSC::EncodedJSValue SOCKET_BAD_PORT(JSC::ThrowScope& throwScope, JSC::JSGlobalOb
     return {};
 }
 
+JSC::EncodedJSValue UNCAUGHT_EXCEPTION_CAPTURE_ALREADY_SET(JSC::ThrowScope& throwScope, JSC::JSGlobalObject* globalObject)
+{
+    auto message = makeString("`process.setupUncaughtExceptionCapture()` was called while a capture callback was already active"_s);
```
Jarred-Sumner: `makeString` allocates a String. We don't need to allocate the string here - just the string literal is fine

src/bun.js/bindings/NodeValidator.cpp
Jarred-Sumner: This comment was marked as resolved but I don't see where it was resolved?

src/bun.js/bindings/BunProcess.cpp
Jarred-Sumner: where was this resolved?

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2154,6 +2273,167 @@ JSC_DEFINE_HOST_FUNCTION(Process_functiongetgroups, (JSGlobalObject * globalObje
...
+
+JSC_DEFINE_HOST_FUNCTION(Process_functionsetuid, (JSGlobalObject * globalObject, CallFrame* callFrame))
+{
+    auto& vm = globalObject->vm();
+    auto scope = DECLARE_THROW_SCOPE(vm);
+    auto value = callFrame->argument(0);
+    auto is_number = value.isNumber();
+    value = maybe_uid_by_name(scope, globalObject, value);
+    RETURN_IF_EXCEPTION(scope, {});
+    if (is_number) Bun::V::validateInteger(scope, globalObject, value, jsString(vm, String("id"_s)), jsNumber(0), jsNumber(std::pow(2, 31) - 1));
```
Jarred-Sumner: we are still creating JSString here unnecessarily

src/bun.js/bindings/BunProcess.cpp
```diff
@@ -2168,13 +2448,18 @@ JSC_DEFINE_HOST_FUNCTION(Process_functionAssert, (JSGlobalObject * globalObject,
...
 
-    JSValue arg1 = callFrame->argument(1);
-    String message = arg1.isUndefined() ? String() : arg1.toWTFString(globalObject);
-    RETURN_IF_EXCEPTION(throwScope, {});
-    auto error = createError(globalObject, makeString("Assertion failed: "_s, message));
-    error->putDirect(vm, Identifier::fromString(vm, "code"_s), jsString(vm, makeString("ERR_ASSERTION"_s)));
-    throwException(globalObject, throwScope, error);
-    return {};
+    auto msg = callFrame->argument(1);
+    auto msgb = msg.toBoolean(globalObject);
```
Jarred-Sumner, line 2451:
```suggestion
    auto msgb = msg.toBoolean(globalObject);
    RETURN_IF_EXCEPTION(throwScope, {});
```

nektro: toBoolean can't throw

src/bun.js/bindings/BunProcess.cpp
nektro: this value is passed to `putDirect` unmodified so it needs to be a JSValue

src/bun.js/bindings/NodeValidator.cpp
nektro: this wasnt modified as part of this pr so it felt more appropriate to change in a followup rather than balloon this one anymore

Jarred-Sumner (changes requested):
We shouldn't make `error:` print as `Error:`
1) There are very likely 3rd-party scripts depending on this behavior
2) It's aesthetically worse

We can make whatever tests that specifically check for the string "Error:" instead check for "error:", or we can make sure it's using util.inspect isntead of Bun.inspect / console.log in the appropriate places.

PR #16031, opened 2024-12-28
Add remove() and isRemoved in HTMLRewriterTypes.Doctype interface

### LINKED ISSUE #16020: Add `remove()` and `isRemoved` in `HTMLRewriterTypes.Doctype` interface

Currently, the argument for the doctype handler in `HTMLRewriterDocumentContentHandlers` is missing `remove()` method and `removed` property. These APIs were added in `lol-html` [v0.3.3](https://github.com/cloudflare/lol-html/commit/93001bf4be7d84aaf30745392ef34fbc75050819#diff-eebe32c5476a191cf32390f7ef174a972274a16d8bfe147508fc9bca66c092e4R321). I assume they were not created as very few people use these APIs but it has niche use cases (like extracting a single node).

Create Zig bindings to `lol_html_doctype_remove` and `lol_html_doctype_is_removed` functions and JavaScript bindings to those Zig APIs. Having type declarations for these APIs will also be very useful.

_No response_

Devanand-Sharma:
### What does this PR do?
Resolves #16020

This adds `remove()` method and `removed` property to the DOCTYPE handler in HTMLRewriter. These APIs were added in `lol-html` v0.3.3 but were not accessible through Bun's HTMLRewriter bindings. 

- [x] Code changes

For Zig files changed:
- [x] I included a test for the new code: `test/js/web/html/html-rewriter-doctype.test.ts`
- [x] I wrote TypeScript/JavaScript tests and they pass locally
- [x] I added TypeScript types for the new methods and getter


Jarred-Sumner (approved):
Looks great. Thank you

PR #16060, opened 2024-12-30
Implement retry and repeat functionality for bun test

### LINKED ISSUE #16051: `bun test` does not respect `repeats` or `retry` properties in test options

1.1.42+50eec0025

Run these tests. Do not expect it to complete in a reasonable amount of time.

```ts
import { it, expect } from "bun:test";

it("fails", () => {
  expect.unreachable();
}, { retry: 1e9 });

it("passes", () => {}, { repeats: 1e9 });
```

Both tests should hang because they have a massive amount of retries or repeats.

### What do you see instead?

Both tests complete immediately.

### Additional information

I'm not sure what the output should be in these cases, which is why the repro is about making the tests hang instead of observing the output.

Devanand-Sharma:
### What does this PR do?
Fixes #16051

This implements retry and repeat functionality for `bun test`. When tests include a `retry` count, failed tests will retry up to the specified number of times. When tests include a `repeats` count, tests will run the specified additional number of times regardless of result. This also fixes async promise state handling in the test runner.

Examples:
```typescript
// Retry up to 3 times on failure
it("should retry on failure", () => {
 expect(false).toBe(true);
}, { retry: 3 });

// Run test 3 additional times
it("should repeat regardless of result", () => {
}, { repeats: 3 });
```

- [x] Code changes

I wrote automated tests in test/js/bun/test/retries-test.test.ts that verifies:
- Retry functionality for both sync and async tests
- Repeat functionality for both sync and async tests
- Tests that eventually succeed after retries
- Edge cases (i.e. negative retry/repeat counts)
- Proper promise state handling during retries
- [x] I ran my tests locally and they pass (`bun-debug test test/js/bun/test/retries-test.test.ts`)

- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test/js/bun/test/retries-test.test.ts`)


goldenstein64:
What behavior should there be when both `retry` and `repeats` are specified together? IMO it wouldn't make sense to do this, so it should error.

```ts
it("should not run this test", () => {}, { retry: 3, repeats: 3 });
```

Devanand-Sharma:
@goldenstein64 Thank you for the feedback. The current expected behaviour is that retry and repeat used together should run. 

My thinking is that retry and repeat should be able to be used together for being able to test stability on cases where an occassional failure can be expected.

retry --> test handle failure until passing (this is useful for when you're expecting occasional fails).
repeat --> used for testing stability

```typescript
// Testing network API endpoint reliability
it("should handle API requests consistently", async () => {
  const response = await fetch("https://api.example.com/data");
  expect(response.status).toBe(200);
}, { 
  // Retry if network request fails initially
  retry: 3,
  // After successful request, repeat to verify endpoint stability
  repeats: 2
});
```

If this doesn't make sense or you disagree, lmk! I'm open to suggested changes.

goldenstein64:
I was mostly concerned that the behavior would be ambiguous, e.g. either `retry` or `repeats` takes precedence, or there was undefined behavior.

From your description/example, I think the general behavior would be mocked up with this code. If this is correct, then it makes sense to me :)

```ts
import type { TestOptions } from "bun:test";
import { it as defineTest, expect } from "bun:test";

export async function it(
  testName: string,
  runBody: () => void,
  { retry = 1, repeats = 1 }: TestOptions = {}
) {
  for (let trial = 0; trial < repeats; trial++) {
    defineTest(`${testName} > [${trial}]`, async () => {
      let succeededOnce = false;
      let errors: unknown[] = [];

      for (let j = 0; j < retry; j++) {
        try {
          await runBody();
          succeededOnce = true;
          break;
        } catch (err) {
          errors.push(err);
        }
      }

      if (!succeededOnce) {
        expect.unreachable(`\n${errors.join("\n")}`);
      }
    });
  }
}

```

Devanand-Sharma:
Yep! lgtm!

PR #16072, opened 2024-12-31
Update nextjs.md docs with starter example

sirmews:
Adds an explicit mention of how to create a new nextjs project with a starter template.

I've been asked multiple times in my team on how to do this and figured it warranted a mention in the official docs, in case it wasn't obvious to newcomers.

- [x] Documentation


Jarred-Sumner:
Thank you

Issue #8699, opened 2024-02-05
Bun.$ (or Bun.spawn) thinks all dependencies are git repositories?


huseeiin:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Linux 6.1.66-1-lts x86_64 unknown

### What steps can reproduce the bug?

https://github.com/huseeiin/create-svelte-bun
see `src/index.ts` line 165

Electroid:
Could you provide a more minimal reproduction? It's a little difficult to cipher through

huseeiin:
> Could you provide a more minimal reproduction? It's a little difficult to cipher through

it appears to be because I'm doing this pattern "bun i dep && bun i dep2 -D" where you run two commands (the same cause of https://github.com/oven-sh/bun/issues/8700 somehow?).

PR #16073, opened 2024-12-31
fix(us): memory leak when getting root certificate

DonIsaac:
### What does this PR do?
Fixes a memory leak when an x509 root certificate is successfully read for an https server.

There are already tests for this. I just ran this command to find+fix it:
```sh
leaks -atExit -- ./build/debug/bun-debug test ./test/js/bun/http/serve-body-leak.test.ts | tee leaks.log
```

There's a bunch of other reported leaks that I'll address in other PRs.

### DIFF
--- a/packages/bun-usockets/src/crypto/root_certs.cpp
+++ b/packages/bun-usockets/src/crypto/root_certs.cpp
@@ -35,6 +35,9 @@ us_ssl_ctx_get_X509_without_callback_from(struct us_cert_string_t content) {
     OPENSSL_PUT_ERROR(SSL, ERR_R_PEM_LIB);
     goto end;
   }
+
+  // NOTE: PEM_read_bio_X509 allocates, so input BIO must be freed.
+  BIO_free(in);
   return x;
 end:
   X509_free(x);


robobun:
<div><sup>Updated 6:12 PM PT - Dec 30th, 2024</sup></div>

@DonIsaac, your commit dd4b248e9bf98f207669982e876500bd1981a604 is building: [`#8699`](https://buildkite.com/bun/bun/builds/8699)

Jarred-Sumner (changes requested):
Either one of two things are true:
- This code causes `x` to be invalid memory
- ~There are two more memory leaks in that code~ edit: i'm wrong

But I do think it's worth finding out if `x` is still valid

DonIsaac:
- `PEM_read_bio_X509` is declared by a `DECLARE_PEM_rw` macro.
- That function is implemented by `IMPLEMENT_PEM_read_bio`. It wraps a call to `PEM_ASN1_read_bio` and casts the returned void pointer.
- `PEM_ASN1_read_bio` calls `PEM_bytes_read_bio` [here](https://github.com/oven-sh/boringssl/blob/914b005ef3ece44159dca0ffad74eb42a9f6679f/crypto/pem/pem_oth.cc#L77)
- `PEM_bytes_read_bio` passes output pointers to [PEM_read_bio](https://github.com/oven-sh/boringssl/blob/914b005ef3ece44159dca0ffad74eb42a9f6679f/crypto/pem/pem_lib.cc#L221), and will [free those pointers on error](https://github.com/oven-sh/boringssl/blob/914b005ef3ece44159dca0ffad74eb42a9f6679f/crypto/pem/pem_lib.cc#L232)
- `PEM_read_bio` then [allocates memory here](https://github.com/oven-sh/boringssl/blob/914b005ef3ece44159dca0ffad74eb42a9f6679f/crypto/pem/pem_lib.cc#L612)

Thus this is a memory leak.

Issue #8707, opened 2024-02-05
pendingWebsockets not decreasing after closing/terminating connections from server


cgjme:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Linux 4.4.0-19041-Microsoft x86_64 x86_64

### What steps can reproduce the bug?

Simply call `ws.close()` or `ws.terminate` from any server handler such as `open` or `message`.

Simple snippet based on the "Simple Websocket server" example from the docs:
```ts
const server = Bun.serve<{ authToken: string }>({
  fetch(req, server) {
    const success = server.upgrade(req);
    if (success) {
      // Bun automatically returns a 101 Switching Protocols
      // if the upgrade succeeds
      return undefined;
    }

    // handle HTTP request normally
    return new Response("Hello world!");
  },
  websocket: {
    // this is called when a message is received
    async message(ws, message) {
      console.log(`Received ${message}`);
      // send back a message
      ws.send(`You said: ${message}`);
      ws.close(4000, "Closing");
    },
  },
});

console.log(`Listening on ${server.hostname}:${server.port}`);

setInterval(() => {
  console.log(`Pending sockets: ${server.pendingWebSockets}`);
}, 2000);
```

### What is the expected behavior?

When a user is disconnected using `ws.close()` or `ws.terminate()`, `pendingWebsockets` should decrease.

### What do you see instead?

`pendingWebsockets` only decreases when the client closes the connection, never when the server decides to close.

When I close clients that don't provide a valid auth token or exceed a rate limitation, `pendingWebsockets` shows hundreds/thousands connections, despite the real number being <5. This persists for hours/days on end after the connections were closed.

### Additional information

Same behaviour observed in both WSL and in Google Kubernetes Engine.

e3dio:
Can confirm `server.pendingWebSockets` is broken, count does not go down when server closes websocket, only when client closes websocket. Also maybe rename `pendingWebSockets` to `activeWebSockets` or `websocketCount` ?

Bernhard-Kodio:
I'm also running into issues because of this.

This issue makes Bun.serve and frameworks like Elysia that depend on bun serve quite difficult to use.

PR #16075, opened 2024-12-31
fix: memory leak when reading chunks from a stream

### LINKED ISSUE #12198: Streams memory leak

1.1.17+bb66bba1b

Darwin 23.1.0 arm64 arm

Execute the below code, and watch memory usage:

```typescript

setInterval(() => {
    Bun.gc(true);
}, 100);

const p = Bun.spawn(['cat'], {
    stdin: 'pipe',
    stdout: 'pipe',
})


const r = p.stdout.getReader()
while (true) {
    const buf = new Uint8Array(1_000_000);
    await p.stdin.write(buf);

    let i = 0;
    while (true) {
        const { value } = await r.read();
        i += value?.length ?? 0;
        if (i >= buf.length) {
            break;
        }
    }
}
```

Each iteration just wait until the buffer is streamed through `cat`, so I'd expect memory not to increase.

### What do you see instead?

CPU 100%, memory increases rapidly.

Interestingly enough, when every byte has been streamed, if I add another `.read()` (which rightfuly returns no data), the memory leak disapears:

```typescript


setInterval(() => {
    Bun.gc(true);
}, 100);

const p = Bun.spawn(['cat'], {
    stdin: 'pipe',
    stdout: 'pipe',
})


const r = p.stdout.getReader()
while (true) {
    const buf = new Uint8Array(1_000_000);
    await p.stdin.write(buf);

    let i = 0;
    while (true) {
        const { value } = await r.read();
        i += value?.length ?? 0;
        if (i >= buf.length) {
           // 👉  this fixes the leak ! (and never throws)
            const { value: value2 } = await r.read();
            if (value2?.length) {
                throw new Error('Expected EOF');
            }
            break;
        }
    }
}
```

### Additional information

This issue is kind of linked to #12194 (which I discovered investigating this leak)

DonIsaac:
### What does this PR do?
Fixes #12198


robobun:
<div><sup>Updated 8:17 PM PT - Dec 30th, 2024</sup></div>

@DonIsaac, your commit 327451429da2ece4e8ce5bc1ee791bf6d8409939 is building: [`#8707`](https://buildkite.com/bun/bun/builds/8707)

src/bun.js/webcore/streams.zig
```diff
@@ -1021,6 +1023,11 @@ pub const StreamResult = union(Tag) {
                 const value = result.toJS(globalThis);
                 value.ensureStillAlive();
 
+                switch (result.*) {
+                    .temporary => |*temporary| temporary.deinitWithAllocator(bun.default_allocator),
+                    .temporary_and_done => |*temporary_and_done| temporary_and_done.deinitWithAllocator(bun.default_allocator),
+                    else => {},
+                }
                 result.* = .{ .temporary = .{} };
```
DonIsaac, line 1030: Will this cause JSValues for `.owned`, etc. variants to point to leaked/undefined memory?

DonIsaac: Does line 1046 move or borrow?

Man I miss Rust's borrow checker...

src/bun.js/webcore/streams.zig
```diff
@@ -1021,6 +1023,11 @@ pub const StreamResult = union(Tag) {
                 const value = result.toJS(globalThis);
                 value.ensureStillAlive();
 
+                switch (result.*) {
+                    .temporary => |*temporary| temporary.deinitWithAllocator(bun.default_allocator),
```
cirospaciari: temporary should mean that the StreamResult don't own the data and should be cloned, the owner (how created the StreamResult) should free it.

Jarred-Sumner:
I'm skeptical this is either the cause or the fix.

More likely, we're not reporting the memory allocated in ArrayBufferSink to the GC, so it has no idea we allocated a bunch of memory.

Issue #8774, opened 2024-02-07
`fetch` function causes a weird error with bun test + react + happy-dom


lumap:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

- Clone https://github.com/lumap/bun-issue-hapydom
- Run `bun i`
- Bun `bun test`
- Observe

### What is the expected behavior?

An error telling me `fetch` is not mocked

### What do you see instead?

```js
bun test v1.0.26 (c75e768a)

src/App.test.js:
312 |             });
313 |             this.nodeRequest.on('error', this.onError.bind(this));
314 |             this.nodeRequest.on('socket', this.onSocket.bind(this));
315 |             this.nodeRequest.on('response', this.onResponse.bind(this));
316 |             if (this.request.body === null) {
317 |                 this.nodeRequest.end();
                      ^
TypeError: undefined is not a function
      at node:http:601:95
      at _final (node:http:865:17)
      at callFinal (node:stream:2763:45)
      at prefinish (node:stream:2786:45)
      at finishMaybe (node:stream:2793:49)
      at node:stream:2726:151
      at /Users/lumap/tmp/node_modules/happy-dom/lib/fetch/Fetch.js:317:17
      at new Promise (:1:21)
      at sendRequest (/Users/lumap/tmp/node_modules/happy-dom/lib/fetch/Fetch.js:268:16)
      at /Users/lumap/tmp/node_modules/happy-dom/lib/fetch/Fetch.js:119:22

32 |   // throw errors w/ suggestions for better queries. Opt in so off by default.
33 |   throwSuggestions: false,
34 |   // called when getBy* queries fail. (message, container) => Error
35 |   getElementError(message, container) {
36 |     const prettifiedDOM = (0, _prettyDom.prettyDOM)(container);
37 |     const error = new Error([message, `Ignored nodes: comments, ${config.defaultIgnore}\n${prettifiedDOM}`].filter(Boolean).join('\n\n'));
                       ^
TestingLibraryElementError: Unable to find an element with the text: Home. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

Ignored nodes: comments, script, style
<body>
  <div>
    <div
      class="App"
    >
      <header
        class="App-header"
      >
        <p>
          Edit
          <code>
            src/App.js
          </code>
           and save to reload.
        </p>
        <a
[...]

PR #16093, opened 2025-01-01
support `bun install --filter <pattern>`

### LINKED ISSUE #15948: Allow restricting `bun install` to specific packages in a workspace

When we have a workspace project with many different packages, `bun` requires all packages to be present when running `bun install` even if we just want to install the dependencies for a single package. 

This is relevant for Docker where if we have many packages, we want to copy just the required packages to keep the Docker image size small. 

For example, in a monorepo with multiple services, each service might only depend on a subset of shared packages:
- `packages/api` depends on `packages/shared`
- `packages/web` depends on `packages/ui` and `packages/shared`
- `packages/mobile` depends on `packages/ui` and `packages/shared`

When building a Docker image for just the API service, we currently have to copy ALL packages (`web`, `mobile`, `ui`, etc.) even though we only need `api` and `shared`. This is because the lockfile contains references to all workspace packages.

Add ability to scope installation to specific packages, similar to pnpm's `--filter` flag:

1. Copy entire workspace (inefficient, larger Docker images)
2. Split into separate repos (loses monorepo benefits)
3. Dynamically update `package.json` in the Dockerfile (loses the lockfile benefits and is just generally janky)

dylan-conway:
### What does this PR do?
This allows `bun install` to install packages only for the matching patterns provided to `--filter`.
```sh
# Install dependencies for all workspaces with names starting with `frontend-`
$ bun install --filter "frontend-*"
```

closes #15948 

tests:
- [x] patterns with `!`
- [x] no matching workspaces
- [x] only root package, no workspaces
- [x] matched workspace pulling in filtered workspace
- [x] workspace path filters


robobun:
<div><sup>Updated 7:08 PM PT - Dec 31st, 2024</sup></div>

@dylan-conway, your commit d20c3f234a1c571de8800d72dbbdaab8d712f544 is building: [`#8774`](https://buildkite.com/bun/bun/builds/8774)

ghiscoding:
> Not sure about this flag name. In the future `bun add <package> --filter` will add the package to each matching workspace, making `bun install <package> --filter` confusing.

seems like an ok flag since that is exactly what pnpm offers via their [`pnpm install --filter`](https://pnpm.io/cli/install#--filter-package_selector). I would think that offering the same kind of filters as pnpm does seems like a good fit

PR #16097, opened 2025-01-01
Update `defines` to `define` in `cc` function documentation

komiya-atsushi:
### What does this PR do?

This option is named `defined` [in the code](https://github.com/oven-sh/bun/blob/be959e111ab6842026a54328e5c06747a8e47f6f/packages/bun-types/ffi.d.ts#L676-L698), but it is written as `defines` [in the documentation](https://bun.sh/docs/api/cc#defines-record-string-string).

- [ ] Code changes


Jarred-Sumner:
Thank you

Issue #15355, opened 2024-11-22
Bun always sets Content-Length to 0 for HEAD response


kensnyder:
### What version of Bun is running?

1.1.36

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

Consider the following code:

```ts
const server = Bun.serve({
  fetch() {
    return new Response(null, {
      status: 200,
      headers: {
        'Content-Type': 'image/jpeg',
        'Content-Length': '1000',
      },
    });
  },
});

const resp = await fetch(server.url, { method: 'HEAD' });

// should be 1000, but is actually 0
const contentLength = resp.headers.get('Content-Length');
console.log(`Content-length=${contentLength}`);

server.stop(true);
```

### What is the expected behavior?

To conform to HTTP Server specs, I'd like to be able to respond to HEAD requests with a Content-Length of the size of the actual resource.

### What do you see instead?

Bun appears to change the Content-Length to 0 before responding.

### Additional information

Also reported in Discord on May 15, 2024 by fry69: https://discord.com/channels/876711213126520882/1240212448552816710/1240212448552816710

Kapsonfire-DE:
>  The HEAD method is identical to GET except that the server MUST NOT
   send a message body in the response

https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

I dont see a bug here

Kapsonfire-DE:
@cirospaciari I'd like to work on this, since I also did the OPTIONS thing.
The same part of code is responsible here. But HEAD is different, since RFC says it must not send a body, but headers should be same as if its a GET Request... so i think its reasonable to make an exception in Code for GET Request, not changing headers, but ommiting the body.

kensnyder:
Funny story... just today I ran into a new place this bug is affecting me. Let me explain and share my workaround code.

When Cloudflare Stream downloads a video you ask it to transcode, it appears recently started performing a HEAD request on the video URL before requesting the whole video. My videos are being served by Bun, so the HEAD responses will always have `Content-Length` of `0`. When Cloudflare Stream sees the `0`, gives the following error, and fails to transcode the video:

```json
messages: [
  {
    code: 10010,
    message: "Performed a HTTP HEAD request, but HEAD request did not include a content-length header. Please make sure your origin returns content length headers."
  }
]
```

I took 3 steps to work around this bug:
 
[...]

PR #16099, opened 2025-01-01
fix(server) HEAD requests

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/15355

This PR fixed HEAD requests respecting the headers provided or the body informed to get the `Content-Length`, and also don't send the body even if is provided because this would be a malformed http request, now we will never send the body respecting the spec https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2

```js
 // now if is a HEAD request this will never send the content only the content-length
 // if the body is a readable stream will send transfer-encoding: chunked
 const server = Bun.serve({
      port: 0,
      fetch(req) {
        // or new Response(file(filename))) or any variation
        return new Response("Hello World");
      },
 });
 ```

- [x] Code changes

Tests

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -5069,10 +5069,17 @@ pub const ServerWebSocket = struct {
             return globalThis.throw("publishText requires a non-empty message", .{});
         }
 
-        var string_slice = message_value.toSlice(globalThis, bun.default_allocator);
-        defer string_slice.deinit();
+        var js_string = message_value.toString(globalThis);
+        if (globalThis.hasException()) {
+            return .zero;
+        }
+        const view = js_string.view(globalThis);
+        const slice = view.toSlice(bun.default_allocator);
+        defer slice.deinit();
 
-        const buffer = string_slice.slice();
+        defer js_string.ensureStillAlive();
+
+        const buffer = slice.slice();
 
         const result = if (!publish_to_self and !this.isClosed())
             this.websocket().publish(topic_slice.slice(), buffer, .text, compress)
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -5323,10 +5323,17 @@ pub const ServerWebSocket = struct {
         }
 
         {
-            var string_slice = message_value.toSlice(globalThis, bun.default_allocator);
-            defer string_slice.deinit();
+            var js_string = message_value.toString(globalThis);
+            if (globalThis.hasException()) {
+                return .zero;
+            }
+            const view = js_string.view(globalThis);
+            const slice = view.toSlice(bun.default_allocator);
+            defer slice.deinit();
 
-            const buffer = string_slice.slice();
+            defer js_string.ensureStillAlive();
+
+            const buffer = slice.slice();
             switch (this.websocket().send(buffer, .text, compress, true)) {
                 .backpressure => {
                     log("send() backpressure ({d} bytes string)", .{buffer.len});
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3272,7 +3272,6 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         }
 
         // Each HTTP request or TCP socket connection is effectively a "task".
-        // Each HTTP request or TCP socket connection is effectively a "task".
         //
         // However, unlike the regular task queue, we don't drain the microtask
         // queue at the end.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by robobun** [comment]
<div><sup>Updated 11:38 AM PT - Jan 1st, 2025</sup></div>

@cirospaciari, your commit a54a8fb4d473461cd20bae4d8a920a16c52e858c is building: [`#8793`](https://buildkite.com/bun/bun/builds/8793)

File: src/bun.js/api/server.zig [hunk 1]
```diff
@@ -3467,17 +3589,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
                                 // we can avoid streaming it and just send it all at once.
                                 if (byte_stream.has_received_last_chunk) {
-                                    this.blob.from(byte_stream.buffer);
+                                    this.blob.from(byte_stream.drain().listManaged(bun.default_allocator));
                                     this.readable_stream_ref.deinit();
                                     this.doRenderBlob();
                                     return;
                                 }
-
+                                this.ref();
```
cirospaciari, line 3596: this was a missing ref when onPipe is attached, onPipe will deref when done

src/bun.js/api/server.zig
```diff
@@ -3467,17 +3589,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
                                 byte_stream.pipe = JSC.WebCore.Pipe.New(@This(), onPipe).init(this);
                                 this.readable_stream_ref = JSC.WebCore.ReadableStream.Strong.init(stream, globalThis);
 
                                 this.byte_stream = byte_stream;
-                                this.response_buf_owned = byte_stream.buffer.moveToUnmanaged();
+                                this.response_buf_owned = byte_stream.drain().list();
```
cirospaciari, line 3601: drain is the right method to use here

src/bun.js/api/server.zig
```diff
@@ -3467,17 +3589,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                                 // If we've received the complete body by the time this function is called
                                 // we can avoid streaming it and just send it all at once.
                                 if (byte_stream.has_received_last_chunk) {
-                                    this.blob.from(byte_stream.buffer);
+                                    this.blob.from(byte_stream.drain().listManaged(bun.default_allocator));
```
cirospaciari, line 3591: drain is the right method to use here too

src/bun.js/api/server.zig
```diff
@@ -4877,10 +5000,17 @@ pub const ServerWebSocket = struct {
         }
 
         {
-            var string_slice = message_value.toSlice(globalThis, bun.default_allocator);
-            defer string_slice.deinit();
+            var js_string = message_value.toString(globalThis);
```
cirospaciari, line 5002: I saw this changes on main branch and replicated in more places lmk if this is right

src/bun.js/api/server.zig
```diff
@@ -3162,6 +3171,105 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
+                // server detached?
+                resp.writeHeaderInt("content-length", 0);
+                this.renderMetadata();
+                this.endWithoutBody(this.shouldCloseConnection());
+                return;
+            };
+            const globalThis = server.globalThis;
+            if (response.getFetchHeaders()) |headers| {
+                // first respect the headers
+                if (headers.get("transfer-encoding", globalThis)) |transfer_encoding| {
```
Jarred-Sumner, line 3218:
```suggestion
                if (headers.fastGet(.TransferEncoding)) |transfer_encoding| {
```

src/bun.js/api/server.zig
```diff
@@ -3162,6 +3171,105 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
...
+                    resp.writeHeader("transfer-encoding", transfer_encoding);
+                } else if (headers.get("content-length", globalThis)) |content_length| {
```
Jarred-Sumner, line 3220:
```suggestion
                } else if (headers.fastGet(.ContentLength, globalThis)) |content_length| {
```

src/bun.js/api/server.zig
Jarred-Sumner: reason is it will avoid cloning sliced strings

src/bun.js/api/server.zig
cirospaciari: fixed on the followup https://github.com/oven-sh/bun/pull/16115

src/bun.js/api/server.zig
cirospaciari: fixed on the followup https://github.com/oven-sh/bun/pull/16115

PR #16109, opened 2025-01-02
Support generating V8 Heap Snapshots

Jarred-Sumner:
### What does this PR do?

This does two things:
- Support generating V8 Heap Snapshots
- Wires up the code generator to make the cached values in getters look like properties in the heap snapshots, so you can see the values even though they're behind a getter

TODO before merging:
- [x] Add a couple tests
- [x] Wire up Bun.captureHeapSnapshot to accept either v8 or jsc instead of only one
- [ ] Investigate moving the JSC::BunV8HeapSnapshot class to Bun's codebase instead of JSC's
- [ ] Figure out a way to make this DX better.

There were three tools I used for manually testing this:
- Chrome DevTools
- [vscode-v8-heap-tools](https://github.com/microsoft/vscode-v8-heap-tools), which would panic when the format was wrong
- V8's implementation https://chromium.googlesource.com/v8/v8/+/master/src/profiler/heap-snapshot-generator.cc


robobun:
<div><sup>Updated 5:13 AM PT - Jan 2nd, 2025</sup></div>

@Jarred-Sumner, your commit 6b2d83274ec69b047c56685f07f71390760b991a is building: [`#8804`](https://buildkite.com/bun/bun/builds/8804)

PR #16112, opened 2025-01-02
feat: run_command start "pnpx " & "pnpm dlx " -> "bun x "

maxktz:
### What does this PR do?

For run commands of `package.json` `"scripts"` bun replaces starting "npx " to "bun x "

I made it replace also the "pnpx" (5 lines of code)

image

I haven't tested this small change, hope you can make it yourself

However I haven't tested it not because It's just 5 lines, but because it gives me Zig errors at the middle of compiling. I think your contributing guide has issues, did exact what it said. (Tried on Mac M1)

### DIFF
--- a/src/cli/run_command.zig
+++ b/src/cli/run_command.zig
@@ -220,6 +220,13 @@ pub const RunCommand = struct {
                             continue;
                         }
 
+                        if (strings.hasPrefixComptime(script[start..], "pnpm dlx ")) {
+                             try copy_script.appendSlice(BUN_BIN_NAME ++ " x ");
+                             entry_i += "pnpm dlx ".len;
+                             delimiter = 0;
+                             continue;
+                         }
+
                         if (strings.hasPrefixComptime(script[start..], "pnpx ")) {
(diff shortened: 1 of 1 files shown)


hubdidub:
I don't know for sure but I'm guessing the 'n' case of this switch-statement is the starting letter of the string that is being matched. So you probably need to move it to the 'p'-case below.

src/cli/run_command.zig
```diff
@@ -219,6 +219,13 @@ pub const RunCommand = struct {
                             delimiter = 0;
                             continue;
                         }
+
```
nektro, line 222:
```suggestion

                        if (strings.hasPrefixComptime(script[start..], "pnpm dlx ")) {
                             try copy_script.appendSlice(BUN_BIN_NAME ++ " x ");
                             entry_i += "pnpm dlx ".len;
                             delimiter = 0;
                             continue;
                         }

```

pfgithub:
What were the errors you got while trying to build bun?

PR #16126, opened 2025-01-03
Use polling-based VM traps on Linux

Jarred-Sumner:
### What does this PR do?

SIGUSR1 triggers GC on Linux

This is confusing and likely leads to some hangs

Let's try enabling polling-based VM traps


robobun:
<div><sup>Updated 11:07 PM PT - Jan 2nd, 2025</sup></div>

@Jarred-Sumner, your commit 6a923026ca7bab70f941823e6e0a57394bc59885 is building: [`#8866`](https://buildkite.com/bun/bun/builds/8866)

Issue #5283, opened 2023-09-13
`bun ci` as an alias of `bun install --frozen-lockfile`


Pandapip1:
### What is the problem this feature would solve?

NPM has a command, `npm ci`, that installs the exact packages in the lockfile for reproducible builds. Bun's command, `bun install --frozen-lockfile`, is much more verbose, is less well documented, requires adjustment of muscle memory, and makes migration slightly harder.

### What is the feature you are proposing to solve the problem?

A new subcommand, `bun ci`, which is just an alias for `bun install --frozen-lockfile`.

### What alternatives have you considered?

`bun install` could do a clean install by default.

madwau:
I believe this already exists in Bun as an option:

```
bun install --frozen-lockfile
```

From the [docs](https://bun.sh/docs/cli/install):

> To install with reproducible dependencies, use --frozen-lockfile. If your package.json disagrees with bun.lockb, Bun will exit with an error. This is useful for production builds and CI environments.

Pandapip1:
Could you make `bun ci` an alias of that, since I wasn't easily able to find that subcommand and it's one that would see really common use?

nrmnqdds:
> I believe this already exists in Bun as an option:
> 
> ```
> bun install --frozen-lockfile
> ```
> 
> From the [docs](https://bun.sh/docs/cli/install):
> 
> > To install with reproducible dependencies, use --frozen-lockfile. If your package.json disagrees with bun.lockb, Bun will exit with an error. This is useful for production builds and CI environments.

im getting error
```
error: lockfile had changes, but lockfile is frozen
```
when trying to execute it in my github actions like this `  - run: bun install --frozen-lockfile && bun run build`

kurucu:
> > I believe this already exists in Bun as an option:
> > ```
> > bun install --frozen-lockfile
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > From the [docs](https://bun.sh/docs/cli/install):
> > > To install with reproducible dependencies, use --frozen-lockfile. If your package.json disagrees with bun.lockb, Bun will exit with an error. This is useful for production builds and CI environments.
> 
> im getting error
> 
> ```
> error: lockfile had changes, but lockfile is frozen
> ```
> 
> when trying to execute it in my github actions like this ` - run: bun install --frozen-lockfile && bun run build`

That would happen with npm ci too - both fail with an error if your lockfile doesn't match package.json.

[...]

PR #16137, opened 2025-01-03
Add a 'ci' command alias equivalent to 'install --frozen-lockfile'

mchv:
### What does this PR do?

This adds a new command `ci` similar to [npm ci](https://docs.npmjs.com/cli/v11/commands/npm-ci) which is an alias for `install --frozen-lockfile`. The associated changes try to limit new code to handle this alias, I have added specific comments below.

fixes #5283

- [X] Code changes

- I wrote one automated test
- I tested locally

.

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -573,6 +573,8 @@ pub const Command = struct {
 
                 break :brk .InstallCommand;
             },
+            RootCommandMatcher.case("ci") => .InstallCommand,
+
             RootCommandMatcher.case("c"), RootCommandMatcher.case("create") => .CreateCommand,
 
             RootCommandMatcher.case("test") => .TestCommand,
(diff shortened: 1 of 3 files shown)


src/install/install.zig
```diff
@@ -9669,6 +9669,25 @@ pub const PackageManager = struct {
...
+                const i = this.i;
+                this.i += 1;
+                return this.buf[i];
+            }
+
+            pub fn skip(this: *ArgsIterator) bool {
+                return this.next() != null;
+            }
+        };
+
```
mchv: This `ArgsIterator` struct comes from [cli file](https://github.com/oven-sh/bun/blob/bun-v1.1.42/src/cli.zig#L1554-L1571) and maybe it should be extracted to be reused instead of copied but it is the first copy and code is pretty simple so I don't know if it is worth yet.

Zamiell:
@mchv It looks like the PR has conflicts, would you mind resolving?

mchv:
@RiskyMH Excellent! thank you

Issue #8910, opened 2024-02-15
TLS(HTTPS) Server don't have addContext function 


rahbari:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

in Bun a secure context can't be added to HTTPS server for SNI like nodes: 
`HTTPS.createServer({key, cert}, this.HandleHttp).addContext(domain, {key, cert})`
https://nodejs.org/api/tls.html#serveraddcontexthostname-context

### What do you see instead?

`TypeError: server.addContext is not a function. (In 'server.addContext(domain, {key, cert})', 'server.addContext' is undefined)`

PR #16149, opened 2025-01-04
Support absolute paths when bundling HTML

Jarred-Sumner:
### What does this PR do?

Ordinarily, importing an absolute filesystem path should resolve to the absolute filesystem path.

In HTML, people expect it to resolve relative to the server root. They often do `<script src="/src/index.js"></script>` instead of `<script src="./src/index.js"></script>`

Instead of erroring, we should adapt this behavior specifically when bundling .html files and make it work. 

There is a test that previously failed and now passes

### DIFF
--- a/src/HTMLScanner.zig
+++ b/src/HTMLScanner.zig
@@ -28,12 +28,20 @@ pub fn deinit(this: *HTMLScanner) void {
     this.import_records.deinitWithAllocator(this.allocator);
 }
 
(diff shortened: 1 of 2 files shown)


robobun:
<div><sup>Updated 5:31 PM PT - Jan 3rd, 2025</sup></div>

@Jarred-Sumner, your commit fea90245340ae528ae9424fa6fe6a3120b54e951 is building: [`#8910`](https://buildkite.com/bun/bun/builds/8910)

dylan-conway (approved):

PR #16158, opened 2025-01-04
fix assertion in h2_frame_parser for StreamPriority

nektro:
discovered because `test/js/node/test/parallel/test-http2-client-set-priority.js` crashes in debug mode
did not add to this pr because there are still remaining failures in the file

per https://httpwg.org/specs/rfc7540.html#PRIORITY

> Weight: An unsigned 8-bit integer representing a priority weight for the stream (see [Section 5.3](https://httpwg.org/specs/rfc7540.html#StreamPriority)). Add one to the value to obtain a weight between 1 and 256.

thus `.weight` is actually supposed to be offset by one

cc @cirospaciari


robobun:
<div><sup>Updated 12:25 AM PT - Jan 4th, 2025</sup></div>

@nektro, your commit 8be47c93c38042666e0c9c8b8462807601199c98 is building: [`#8943`](https://buildkite.com/bun/bun/builds/8943)

src/bun.js/api/bun/h2_frame_parser.zig
```diff
@@ -2660,7 +2660,7 @@ pub const H2FrameParser = struct {
 
             var priority: StreamPriority = .{
                 .streamIdentifier = stream_identifier.toUInt32(),
-                .weight = @truncate(stream.weight),
+                .weight = @truncate(stream.weight - 1),
```
cirospaciari, line 2663: Good catch

cirospaciari (approved):

cirospaciari (changes requested):
Looks like  stream.weight is being set in more places we should add tests for this in specific and check to not being less than zero

Issue #14721, opened 2024-10-21
Building with "--packages external" exclude imports with typescript paths


EMPTYVOID-DEV:
### What version of Bun is running?

1.1.27+267afa293

### What platform is your computer?

Linux 6.11.3-200.fc40.x86_64 x86_64 unknown

#### Create typescript paths (e.g):

`
    "paths": {
      "@shared/*": ["./src/shared/*"],
      "@utils/*": ["./src/utils/*"],
      "@middlewares/*": ["./src/middlewares/*"],
      "@database/*": ["./src/database/*"],
      "@routes/*": ["./src/routes/*"]
    }
`

#### Run bun build and exclude packages imports

`
bun build ./src/index.ts --outdir ./dist --target bun --packages external
`

### Relevant log?

Bun will exclude typescript paths from the bundle.

`
// @bun
// src/index.ts
import {OpenAPIHono} from "@hono/zod-openapi";
import {logger as logger2} from "hono/logger";
import {readStaticRoute} from "@routes/files/files.dal";
import {readStaticHandler} from "@routes/files/files.handlers";
import {apiReference} from "@scalar/hono-api-reference";
import limiter from "@middlewares/rateLimiter";
import filesRoute from "@routes/files/index";
import imgTransformationRoute from "@routes/imgTransformations/index";
import pdfTransformationRoute from "@routes/pdfTransformations/index";
import vidTransformations2 from "@routes/vidTransformations/index";
import {writeToLog} from "@utils/utils.general";
var port = "3000";
var app = new OpenAPIHono;
app.use(limiter);
app.use(logger2());
app.openapi(readStaticRoute, readStaticHandler);
app.route("/files", filesRoute);
app.route("/transformation/images", imgTransformationRoute);
app.route("/transformation/pdf", pdfTransformationRoute);
app.route("/transformation/videos", vidTransformations2);
app.get("/", (c) => {
  return c.redirect("/ui-docs", 303);
});
app.doc("/json-docs", {
  info: {
    version: "1.0.0",
    title: "g-store api",
    description: "External api to handle store management for g-store"
  },
  openapi: "3.0.0"
});
app.get("/ui-docs", apiReference({
  theme: "purple",
  pageTitle: "g-store api",
  layout: "modern",
  defaultHttpClient: {
    targetKey: "node",
    clientKey: "fetch"
  },
  spec: {
    url: "/json-docs"
  }
}));
app.onError((err, c) => {
  writeToLog(err.message, "ERROR");
[...]

PR #16163, opened 2025-01-04
fix(bundler) Typescript paths with packages external fix

zpix1:
### What does this PR do?

This patch fixes issue #14721, where packages=external option of bundler incorrectly excluded packages which were imported with mapped path (using tsconfig path aliases).

- [X] Code changes

I wrote automated tests 


src/resolver/resolver.zig
```diff
@@ -3623,16 +3621,10 @@ pub const Resolver = struct {
...
-                                    },
-                                    .diff_case = _result.diff_case,
-                                    .dirname_fd = _result.dirname_fd,
-                                    .package_json = package_json,
-                                    .file_fd = auto_main_result.file_fd,
-                                };
+                                return MatchResult{ .path_pair = .{
+                                    .primary = _result.path_pair.primary,
+                                    .secondary = auto_main_result.path_pair.primary,
+                                }, .diff_case = _result.diff_case, .dirname_fd = _result.dirname_fd, .package_json = package_json, .file_fd = auto_main_result.file_fd, .is_node_module = is_node_module_folder };
```
paperclover: this should have a trailing comma to tell the formatter to use multiple lines

paperclover (dismissed):
this seems reasonable

test/bundler/bundler_edgecase.test.ts
```diff
@@ -1357,10 +1353,43 @@ describe("bundler", () => {
...
-    run: {
-      stdout: `
-        Hello World
+    packages: "external",
+    target: "bun",
+    onAfterBundle(api) {
+      api.expectFile("/out.js").not.toInclude(`Hello World`);
+    },
+  });
+  itBundled.only("edgecase/PackageExternalRespectTSPathAliases", {
```
Jarred-Sumner:
```suggestion
  itBundled("edgecase/PackageExternalRespectTSPathAliases", {
```

paperclover (approved):

PR #16187, opened 2025-01-06
handle pnpx & pnpm dlx in package.json scripts

pfgithub:
#16112 but with the added cases in the right switch branch & added tests

Also changes all_yarn_commands to be a ComptimeStringMap instead of a slice of hashes


robobun:
<div><sup>Updated 12:54 PM PT - Jan 6th, 2025</sup></div>

@pfgithub, your commit ed55b58bcdc9fccf97c1b55e8fe3564bd610a62c is building: [`#9017`](https://buildkite.com/bun/bun/builds/9017)

PR #16192, opened 2025-01-07
Download correct musl binaries in `bun build --compile`

thecrypticace:
### What does this PR do?

This makes sure, when compiling for a musl target, that the correct `-musl` npm package is referenced when downloading the Bun binary in `bun build --compile`

- [x] Code changes

Manual testing of `bun build --compile` and subsequent running of said binary in an Alpine Linux Docker container. I'd write automated tests except I'm not certain how to set that up for this case. Any pointers would be appreciated!

cc @Jarred-Sumner this fixes the `--compile` issue we discussed

### DIFF
--- a/src/compile_target.zig
+++ b/src/compile_target.zig
@@ -28,6 +28,14 @@ const Libc = enum {
     /// musl libc
     musl,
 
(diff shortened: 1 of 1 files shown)


Jarred-Sumner:
Thank you

Issue #9038, opened 2024-02-21
error: script "dev" was terminated by signal SIGABRT


warcayac:
### What version of Bun is running?

1.0.28

### What platform is your computer?

Linux 6.7.5-zen1-1-zen x86_64 unknown

### What steps can reproduce the bug?

1. start server: `bun run dev`
2. make a GET request for:  `http://localhost:3001/`
3. wait for some seconds

### What is the expected behavior?

running server without any unexpected errors

### What do you see instead?

```
$ bun run dev
$ NODE_ENV=development bun --env-file=.env run --watch src/index.tsx
🦊 Elysia is running at localhost:3001
2024-02-21T20:34:10.720Z | GET / - 200 (7.45 ms) 
2024-02-21T20:34:10.756Z | GET /public/output.css - 200 (3.46 ms) 
error: script "dev" was terminated by signal SIGABRT (Abort)
[1]    12585 IOT instruction (core dumped)  bun run dev
```

### Additional information

This is a new Elysia project which use Typescript, React, Htmx and Tailwind. My previous projects working with Elysia and Typescript run OK, the only difference in this project is I'm trying these packages: tailwindcss, drizzle-orm. I've configured Tailwind according to [its doc](https://tailwindcss.com/docs/installation).

My project is in this repo: [https://github.com/warcayac/elysia-htmx-tailwind_demo1](https://github.com/warcayac/elysia-htmx-tailwind_demo1)

Electroid:
Duplicate of #9031

warcayac:
In my case I found this error appears when using **@elysiajs/static** package.

PR #16193, opened 2025-01-07
Use separate function to make V8 heap snapshots instead of parameter

190n:
### What does this PR do?

Changes `Bun.generateHeapSnapshot("v8")` to `Bun.generateV8HeapSnapshot()`. `Bun.generateHeapSnapshot()` once again takes no parameter and always returns a snapshot in JSC format.

The motivation for this is that `Bun.generateHeapSnapshot()` currently returns an object, and we can't change that (it is a breaking change), but returning a string is easier for us and for users (since most use cases will just immediately pass the object to `JSON.stringify`). Previously, `Bun.generateHeapSnapshot` would return JSC snapshots as objects, but V8 snapshots as strings. It's unclear to have a function that returns different types according to its parameter. Now it always returns an object, like before, but we can make `Bun.generateV8HeapSnapshot()` return a string since it's a different function.

In short:

```diff
-   function generateHeapSnapshot(format?: "jsc"): HeapSnapshot;
-   function generateHeapSnapshot(format: "v8"): string;
+   function generateHeapSnapshot(): HeapSnapshot;
+   function generateV8HeapSnapshot(): string;
```

- [x] Code changes

Ran the V8 heap snapshot tests.

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -4176,9 +4176,11 @@ declare module "bun" {
   /**
    * Show precise statistics about memory usage of your application
    *
-   * Generate a heap snapshot in JavaScriptCore's format that can be viewed with `bun --inspect` or Safari's Web Inspector
+   * Generate a heap snapshot in JavaScriptCore's format that can be viewed with `bun --inspect` or Safari's Web Inspector.
+   *
+   * This returns an object that should be JSON stringified then written to a file.
    */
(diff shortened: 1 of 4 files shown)


robobun:
<div><sup>Updated 4:43 PM PT - Jan 6th, 2025</sup></div>

@190n, your commit 5f4b2115cd2600caef989fb0daa9e3e68e0ea362 is building: [`#9038`](https://buildkite.com/bun/bun/builds/9038)

Issue #9055, opened 2024-02-22
Can't import jsonc as config file


truongan07:
### What version of Bun is running?

1.0.27+c34bbb2e3

### What platform is your computer?

Linux 6.3.6-arch1-1 x86_64 unknown

### What steps can reproduce the bug?

Bun.sh can't read jsonc file from import

```js
const jsonc = await import('./services.config.jsonc');

console.log('jsonc: ', jsonc);
```

### What is the expected behavior?

return an object as :

```js
{
 a: "a",
 b: "b"
}
```

### What do you see instead?

```js
Module {
  default: "./services.config.jsonc",
}

```

### Additional information

I want to use config file in jsonc format for comments

Jarred-Sumner:
hmm this is supposed to work but i don't think we have tests for it

truongan07:
@Jarred-Sumner Hi my friend, jsonc is very useful when used to create a config file. We need more comments to describe keys.

Delapouite:
Here is the relevant part in `fs.zig`: https://github.com/oven-sh/bun/blob/main/src/fs.zig#L1668-L1675

```zig
    pub fn isJSONCFile(this: *const Path) bool {
        const str = this.name.filename;
        if (!(strings.hasPrefixComptime(str, "tsconfig.") or strings.hasPrefixComptime(str, "jsconfig."))) {
            return false;
        }

        return strings.hasSuffixComptime(str, ".json");
    }
```

So

`tsconfig.json`
```jsonc
{
  // comment
  "foo": "bar"
}
```

`index.js`
```js
const jsonc = await import('./tsconfig.json');

console.log('jsonc: ', jsonc);
```

```sh
bun run index.js
jsonc:  Module {
  default: {
    foo: "bar",
  },
  foo: "bar",
}
```

truongan07:
resolved in v.1.1.5

```
import jsonc from './config/services.config.jsonc' with { type: 'text' };
```

Thank you support

truongan07:
resolved in v.1.1.5

```
import jsonc from './config/services.config.jsonc' with { type: 'text' };
```

Thank you support

RiskyMH:
btw I added better support for this in #16244 where it should just be able to import `*.jsonc` by default and parse it as jsonc.

PR #16200, opened 2025-01-07
Don't chmod UNIX sockets to 700

### LINKED ISSUE #15686: listen on unix socket: different file permissions than node

1.1.38+bf2f153f5

Linux 5.15.167.4-microsoft-standard-WSL2 x86_64 x86_64

I noticed the unix domain sockets created by bun have different permissions than the ones created by node, is that deliberate? Could the permissions be changed to be the same as node?

Example hello world server:

```js
// index.js

const http = require('http');
const fs = require('fs');

// Define the Unix socket path
const socketPath = '/tmp/hello-world.sock';

// Ensure the socket file does not already exist
if (fs.existsSync(socketPath)) {
  fs.unlinkSync(socketPath);
}

// Create the server
const server = http.createServer((req, res) => {
  res.writeHead(200, { 'Content-Type': 'text/plain' });
  res.end('Hello, World!\n');
});

// Start listening on the Unix socket
server.listen(socketPath, () => {
  console.log(`Server is listening on ${socketPath}`);
});

// Handle cleanup on exit
process.on('SIGINT', () => {
  fs.unlinkSync(socketPath);
  process.exit();
});
```

The expected behavior would be for the socket file to have the same permissions as when running the server with node.

Running with `node index.js` (`v22.11.0`):

```srwxr-xr-x 1 irons irons 0 2024-12-08 15:04 hello-world.sock=```

### What do you see instead?

Running with `bun index.js`:

```srwx------ 1 irons irons 0 2024-12-08 15:04 hello-world.sock=```

### Additional information

_No response_

heimskr:
### What does this PR do?

Fixes #15686. Node.js doesn't chmod UNIX domain sockets after their creation, but we (by way of µSockets) chmod them to 700. This PR fixes that by removing the call to chmod.

- [x] Code changes

I ran the provided repro in #15686 before and after this PR's changes. On Ubuntu (in a container on a macOS host), the permissions on the UNIX domain socket were 700 before the removal of the chmod call, as reported, and after, they were 755, which matches the expected Node.js behavior. (It's interesting to note that on macOS proper, the permissions were already `srwxr-xr-x` even with the call to chmod in place.)

### DIFF
--- a/packages/bun-usockets/src/bsd.c
+++ b/packages/bun-usockets/src/bsd.c
@@ -843,13 +843,6 @@ static LIBUS_SOCKET_DESCRIPTOR internal_bsd_create_listen_socket_unix(const char
         return LIBUS_SOCKET_ERROR;
     }
 
-#ifndef _WIN32
-    // 700 permission by default
-    fchmod(listenFd, S_IRWXU);
-#else
-    _chmod(path, S_IREAD | S_IWRITE | S_IEXEC);
-#endif
-
 #ifdef _WIN32
     _unlink(path);
 #else


robobun:
<div><sup>Updated 6:59 PM PT - Jan 6th, 2025</sup></div>

@heimskr, your commit f5594c9f208e1913b345d1b16e1f997d458bb515 is building: [`#9055`](https://buildkite.com/bun/bun/builds/9055)

Jarred-Sumner (approved):

Issue #9098, opened 2024-02-25
npm i -g bun gives an error in docker using the image node:alpine


ghost:
### What version of Bun is running?

1.0.29

### What platform is your computer?

Docker node alpine

### What steps can reproduce the bug?

```dockerfile
FROM node:alpine

WORKDIR /app

RUN npm i -g bun
```

### What is the expected behavior?

download it successfully

### What do you see instead?
```
 > [backend  3/10] RUN npm i -g bun:
79.62 npm notice
79.62 npm notice New minor version of npm available! 10.2.4 -> 10.4.0
79.62 npm notice Changelog: <https://github.com/npm/cli/releases/tag/v10.4.0>
79.62 npm notice Run `npm install -g npm@10.4.0` to update!
79.62 npm notice
79.62 npm ERR! code 1
79.62 npm ERR! path /usr/local/lib/node_modules/bun
79.62 npm ERR! command failed
79.62 npm ERR! command sh -c node install.js
79.62 npm ERR! Failed to find package "@oven/bun-linux-x64". You may have used the "--no-optional" flag when running "npm install".
79.62 npm ERR! Failed to find package "@oven/bun-linux-x64-baseline". You may have used the "--no-optional" flag when running "npm install".
79.62 npm ERR! Error: Failed to install package "bun"
79.62 npm ERR!     at /usr/local/lib/node_modules/bun/install.js:311:11
79.62 npm ERR!     at Generator.throw (<anonymous>)
79.62 npm ERR!     at rejected (/usr/local/lib/node_modules/bun/install.js:35:27)
79.62 npm ERR!     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
79.63
79.63 npm ERR! A complete log of this run can be found in: /root/.npm/_logs/2024-02-25T08_40_46_038Z-debug-0.log

failed to solve: process "/bin/sh -c npm i -g bun" did not complete successfully: exit code: 1
```

ChillerDragon:
Duplicate of https://github.com/oven-sh/bun/issues/5545

ghost:
@ChillerDragon I didn't knew, sorry

PR #16221, opened 2025-01-07
feat: expose `globalThis.gc` when `--expose-gc` flag is passed

### LINKED ISSUE #9620: Equivalent for `--expose-gc` in node

1.0.4

Darwin 21.6.0 arm64 arm

`bun --expose-gc --max-old-space-size=4096 benchmark.js` is not supported

`node --expose-gc --max-old-space-size=4096 benchmark.js` with `global.gc()` is callable.

### What do you see instead?

`global.gc()` is not defined in `bun benchmark.js`

### Additional information

_No response_

DonIsaac:
### What does this PR do?
Closes #9620.

Adds support for Node's `--expose-gc` flag. `process.gc` is effectively an alias for `Bun.gc(true)`.

This change is to make Bun more compatable with Node, as well as making it easier to port parallel test cases that use `process.gc()` (e.g. [this test](https://github.com/oven-sh/bun/blob/07287a07f0ea44786b5b27191bdef72cfd0d72e6/test/js/node/test/parallel/test-vm-create-and-run-in-context.js), which fails after making improvements to assert).

I've added tests.

### DIFF
--- a/src/bun.js/bindings/ZigGlobalObject.cpp
+++ b/src/bun.js/bindings/ZigGlobalObject.cpp
@@ -1762,14 +1762,6 @@ JSC_DEFINE_HOST_FUNCTION(functionATOB,
     RELEASE_AND_RETURN(throwScope, JSValue::encode(jsString(vm, result.releaseReturnValue())));
 }
 
-/// `globalThis.gc()` is an alias for `Bun.gc(true)`
-JSC_DEFINE_HOST_FUNCTION(functionJsGc,
-    (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
-{
-    callFrame->setArgument(1, JSC::jsBoolean(true));
-    return Generated::BunObject::jsGc(globalObject, callFrame);
-}
-
 JSC_DEFINE_HOST_FUNCTION(functionReportError,
     (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
 {
@@ -3706,6 +3698,29 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
     consoleObject->putDirectCustomAccessor(vm, Identifier::fromString(vm, "_stdout"_s), CustomGetterSetter::create(vm, getConsoleStdout, nullptr), PropertyAttribute::DontEnum | PropertyAttribute::CustomValue | 0);
     consoleObject->putDirectCustomAccessor(vm, Identifier::fromString(vm, "_stderr"_s), CustomGetterSetter::create(vm, getConsoleStderr, nullptr), PropertyAttribute::DontEnum | PropertyAttribute::CustomValue | 0);
 }
+
+
+// ===================== start conditional builtin globals =====================
+// These functions register globals based on runtime conditions (e.g. CLI flags,
+// environment variables, etc.). See `Run.addConditionalGlobals()` in bun_js.zig
+// for where these are called.
+
+/// `globalThis.gc()` is an alias for `Bun.gc(true)`
+JSC_DEFINE_HOST_FUNCTION(functionJsGc,
+    (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
+{
+    // set `force` to true
+    callFrame->setArgument(1, JSC::jsBoolean(true));
+    return Generated::BunObject::jsGc(globalObject, callFrame);
+}
+
+extern "C" void JSC__JSGlobalObject__addGc(JSC__JSGlobalObject* globalObject) {
+    JSC::VM& vm = globalObject->vm();
+    globalObject->putDirectNativeFunction(vm, globalObject, JSC::Identifier::fromString(vm, "gc"_s), 0, functionJsGc, ImplementationVisibility::Public, JSC::NoIntrinsic, PropertyAttribute::DontEnum | 0);
+}
+
+
+// ====================== end conditional builtin globals ======================
 
 extern "C" bool JSC__JSGlobalObject__startRemoteInspector(JSC__JSGlobalObject* globalObject, unsigned char* host, uint16_t arg1)
 {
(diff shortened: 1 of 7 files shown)


robobun:
<div><sup>Updated 1:44 PM PT - Jan 7th, 2025</sup></div>

@DonIsaac, your commit 04b13d1ddeaf47565e9c6da3a6c25715ddc3e2af is building: [`#9098`](https://buildkite.com/bun/bun/builds/9098)

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3698,6 +3699,28 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
...
+// ===================== start conditional builtin globals =====================
+// These functions register globals based on runtime conditions (e.g. CLI flags,
+// environment variables, etc.). See `Run.addConditionalGlobals()` in bun_js.zig
+// for where these are called.
+
+/// `globalThis.gc()` is an alias for `Bun.gc(true)`
+JSC_DEFINE_HOST_FUNCTION(functionJsGc,
+    (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
+{
+    // set `force` to true
```
paperclover: i don't really like that this is passing a call frame to do this. but its okay i guess

paperclover (commented):
in addition: should this work in `bun test`?

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3698,6 +3699,28 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
...
+    callFrame->setArgument(1, JSC::jsBoolean(true));
```
190n: is `setArgument` safe with an out-of-bounds index?

190n: This sets the argument at index 1 but Bun.gc reads its argument from index 0. So `force` ends up being false.

190n: Can we call `JSC__VM__runGC` instead? That is what `Bun.gc` ends up calling, and it takes `sync` as a C argument so we won't need to mess with a call frame.

paperclover: it also calls a mimalloc function. but yeah that is probably better

DonIsaac: yeah it also [cleans up mimalloc](https://github.com/oven-sh/bun/blob/c22315d399de0169b28976cc1179caf757eedd9e/src/bun.js/javascript.zig#L1267). I'd also prefer having `Bun.gc` and `global.gc` to be completely in sync; this method means all improvements to `Bun.gc` are automatically adopted by `global.gc`

190n: Maybe then we could make an exported wrapper around `BunObject.gc` or `VirtualMachine.garbageCollect` in Zig so that C++ can call them without going through a call frame.

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3708,8 +3708,8 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
 JSC_DEFINE_HOST_FUNCTION(functionJsGc,
     (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
 {
-    // set `force` to true
-    callFrame->setArgument(1, JSC::jsBoolean(true));
+    // set `force` to true. Note that `Bun.gc` reads from the 0th argument, which is normally reserved for `this`.
```
190n: It's true that `this` is stored internally as the 0th argument, but the most commonly-used `JSC::CallFrame` functions add 1 to the offset so that `0` means the first argument, and you use separate functions to access `this`: https://github.com/oven-sh/WebKit/blob/995e299d57a3dd8a7718d2a9d2d978745c46db52/Source/JavaScriptCore/interpreter/CallFrame.h#L270-L277

```cpp
        // In the following (argument() and setArgument()), the 'argument'
        // parameter is the index of the arguments of the target function of
        // this frame. The index starts at 0 for the first arg, 1 for the
        // second, etc.
        //
        // The arguments (in this case) do not include the 'this' value.
        // arguments(0) will not fetch the 'this' value. To get/set 'this',
        // use thisValue() and setThisValue() below.
```

src/bun.js/bindings/ZigGlobalObject.cpp
DonIsaac: Updated to call `fn gc()` directly from CPP code w/o JS passthrough

src/bun_js.zig
```diff
@@ -448,6 +444,23 @@ pub const Run = struct {
         if (!JSC.is_bindgen) JSC.napi.fixDeadCodeElimination();
         vm.globalExit();
     }
+
+    extern fn Bun__ExposeNodeModuleGlobals(*JSC.JSGlobalObject) void;
+    /// add `gc()` to `globalThis`.
+    extern fn JSC__JSGlobalObject__addGc(*JSC.JSGlobalObject) void;
+
+    // fn addConditionalGlobals(vm: *VirtualMachine, ctx: *const Command.Context) void {
```
190n: Dead code

```suggestion
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3698,6 +3699,30 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
...
+// environment variables, etc.). See `Run.addConditionalGlobals()` in bun_js.zig
+// for where these are called.
+
+/// `globalThis.gc()` is an alias for `Bun.gc(true)`
+/// Note that `vm` is a `VirtualMachine*`
+extern "C" size_t Bun__gc(void* vm, bool sync);
+JSC_DEFINE_HOST_FUNCTION(functionJsGc,
+    (JSC::JSGlobalObject * global, JSC::CallFrame* callFrame))
+{
+    Zig::GlobalObject* globalObject = reinterpret_cast<Zig::GlobalObject*>(global);
```
cirospaciari:
```suggestion
    Zig::GlobalObject* globalObject = defaultGlobalObject(global);
```

190n (changes requested):
One nitpick, otherwise LGTM

190n (approved):

PR #16223, opened 2025-01-07
Set error code in Node-API functions

190n:
### What does this PR do?

This is the part of #14501 that made Node-API functions set information about the last error. All functions are supposed to store their error code (including success) in a place where it can be retrieved by `napi_get_last_error_info`.

In Zig code, I made `NapiStatus` an `enum` and `napi_status` an integer type, and it is intended to use `NapiEnv.setLastError` to convert between them, e.g. `return env.setLastError(.string_expected)` will both store the error code on the environment and return the appropriate integer code so the foreign module knows what happened. `@intFromEnum` should only be used in places where you don't have an `env` available or it'd be incorrect to set the error on it for some reason (e.g. I do that in some `napi_threadsafe_function` APIs that may be called from another thread).

In C++ code, we unfortunately have implicit conversions so we have to be more careful to not return an error code without setting it in the environment. Most code should use the `NAPI_` macros at the top of `napi.cpp`, but some cases may need to explicitly call `napi_set_last_error` instead.

I wrote automated tests


robobun:
<div><sup>Updated 1:52 PM PT - Jan 7th, 2025</sup></div>

@190n, your commit f2857c65104f0058891c665cae09bece02ff52ab is building: [`#9101`](https://buildkite.com/bun/bun/builds/9101)

src/bun.js/bindings/napi.cpp
```diff
@@ -2597,99 +2410,80 @@ extern "C" napi_status napi_get_value_bigint_words(napi_env env,
...
-        return napi_invalid_arg;
-    }
 
-    WTF::String code = scriptValue.toWTFString(globalObject);
-    RETURN_IF_EXCEPTION(throwScope, napi_generic_failure);
-
-    if (UNLIKELY(code.isNull())) {
-        return napi_generic_failure;
-    }
+    WTF::String code = scriptValue.getString(globalObject);
```
Jarred-Sumner, line 2493: This needs an exception check since joining rope strings can throw an error

src/bun.js/bindings/napi.cpp
```diff
@@ -2772,109 +2558,84 @@ extern "C" napi_status napi_create_bigint_words(napi_env env,
...
-        return napi_function_expected;
-
     JSC::MarkedArgumentBuffer args;
-    if (argc > 0 && LIKELY(argv != nullptr)) {
-        auto end = argv + argc;
-        for (auto it = argv; it != end; ++it) {
-            args.append(toJS(*it));
-        }
-    }
+    args.fill(vm, argc, [&](JSValue* buffer) {
```
Jarred-Sumner, line 2643: `args.fill` does a slower thing that's only necessary when the size of the value is large enough to exceed the inline stack buffer limit

it will be slightly faster to do the for loop when the length is less than like 6

Jarred-Sumner: (it adds itself to the GC, you can read the code in args.fill in webkit)

src/napi/napi.zig
```diff
@@ -1749,12 +1753,12 @@ pub export fn napi_create_threadsafe_function(
     function.tracker.didSchedule(vm.global);
 
     result.* = function;
-    return .ok;
+    return env.ok();
 }
 pub export fn napi_get_threadsafe_function_context(func: napi_threadsafe_function, result: *?*anyopaque) napi_status {
```
Jarred-Sumner, line 1758: do we need a separate napi_status and NapiStatus?

190n: The intent there is to make it hard to return a status without storing it in the environment, because the functions all return `napi_status` but only `NapiStatus` has the enum values. And the easiest conversion path is `env.setLastError` or the helper functions for very commons statuses like `ok`.

Jarred-Sumner: yeah makes sense

src/bun.js/bindings/napi.cpp
190n: Seems to only interact with GC when inline storage is not used (i.e. there are a lot of arguments): https://github.com/oven-sh/WebKit/blob/3cb28b31a96dcd075b356739d07947ce0ece8773/Source/JavaScriptCore/runtime/ArgList.h#L228-L233

```cpp
        if (!isUsingInlineBuffer()) {
            if (LIKELY(!m_markSet)) {
                m_markSet = &vm.heap.markListSet();
                m_markSet->add(this);
            }
        }
```

Jarred-Sumner (approved):

PR #15437, opened 2024-11-27
Split up napi envs

heimskr:
### What does this PR do?

This PR allows multiple napi envs per global object to exist at once.

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

Ran our napi tests and made sure none of Node's napi tests failed compared to the current stable release of Bun.

Issue #9148, opened 2024-02-28
Bun Shell not able to run `next build` command


RPDeshaies:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

- Run `bunx create-next-app@latest`
- Create a script like `build.ts` and write down this code

```ts
await $`next build`;
```

- update your package.json to include 

```json
"build": "bun build.ts
```

- Run the command `bun run build`

### What is the expected behavior?

The NextJS app is built without any errors

### What do you see instead?

```
 ⨯ TypeError: Cannot read properties of null (reading 'useContext')
Error occurred prerendering page "/_not-found". Read more: https://nextjs.org/docs/messages/prerender-error
Error: <Html> should not be imported outside of pages/_document.
Read more: https://nextjs.org/docs/messages/no-document-import-in-page
Error occurred prerendering page "/404". Read more: https://nextjs.org/docs/messages/prerender-error
```

Electroid:
If you run `next build` outside the shell, does it work?

RPDeshaies:
It does yeah

RPDeshaies:
Here's a repo where I can reproduce the issue.

Download the repo, run `bun install`, and then run `bun run build`

[my-app.zip](https://github.com/oven-sh/bun/files/14438009/my-app.zip)

RPDeshaies:
Just an FYI that I've just tested with Bun `v1.0.30` and the problem persists.

KryptXBSA:
I've just downloaded your code and ran bun run build and the next.js app was built successfully. 
OS: `Linux 6.6.10-76060610-generic x86_64 x86_64`
Bun version: `1.1.3+2615dc742`

Also I think you're encountering that error on another project, because looking at the error it appears that you're using the pages dir but the project in the zip file is using the `app` dir: [https://nextjs.org/docs/messages/no-document-import-in-page](https://nextjs.org/docs/messages/no-document-import-in-page)

The output on my machine:
```
 bun run build
$ bun build.ts
   ▲ Next.js 14.1.0

   Creating an optimized production build ...
request to https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap failed, reason:

Retrying 1/3...
 ✓ Compiled successfully
   Linting and checking validity of types ...
   Collecting page data ...
   Generating static pages (0/5) ...
   Generating static pages (1/5)
[...]

PR #16230, opened 2025-01-08
Fix Node-API bugs (1/n)

190n:
### What does this PR do?

This should only be merged after #15437 and #16223 are merged (because those are broad Node-API refactors, and merging more Node-API stuff into `main` will just make more conflicts for merging those PRs).

This is a few commits from #14501 split into a separate PR. Fixes:

- Properties would not be marked read-only
- Read-only properties would have setters (#15429)
- `napi_define_class` did not check if all its arguments are `NULL`
- All `napi_coerce` functions didn't handle exceptions correctly
- `napi_get_value_bool` didn't handle non-booleans correctly
- Constructor functions would get `NULL` from `napi_get_cb_info` instead of the data pointer provided when the class was created

Current Node-API tests pass.


robobun:
<div><sup>Updated 6:00 PM PT - Jan 7th, 2025</sup></div>

@190n, your commit b3387f48c6ea5d7ac5064e36fddcf056225f0656 is building: [`#9148`](https://buildkite.com/bun/bun/builds/9148)

190n:
Closed: we're going to merge #14501 instead of splitting it into multiple PRs

PR #16236, opened 2025-01-08
bun-types: always depend on the latest version of @types/node

nektro:
`~20.12.8` uses a syntax which uses patch-level compat meaning the highest the package was resolving to was `20.12.14`

discovered with a local run of `bun install --save-text-lockfile`


robobun:
<div><sup>Updated 8:36 PM PT - Jan 7th, 2025</sup></div>

@nektro, your commit d990f4a750fbb469097809d7b51cd110fe67a91d is building: [`#9195`](https://buildkite.com/bun/bun/builds/9195)

PR #16244, opened 2025-01-08
allow importing `bun.lock` (+ types for it)

RiskyMH:
### What does this PR do?
Allows you to `import lockfile from "./bun.lock"` and also provides types (to hopefully help 3rd parties understand it). Also adds extension `.jsonc` that ignores comments too.

I believe the typescript is good, I have tried it myself but maybe I'm being too strict.

Will add tests


robobun:
<div><sup>Updated 3:19 AM PT - Jan 8th, 2025</sup></div>

@RiskyMH, your commit 963d21af5461f342b3bcd471c6df22ff7d26d835 is building: [`#9230`](https://buildkite.com/bun/bun/builds/9230)

packages/bun-types/bun.d.ts
```diff
@@ -6417,6 +6417,77 @@ declare module "bun" {
...
+     * root        -> [ "name@root:", { bin, binDir } ]
+     * git         -> [ "name@git+repo", INFO, .bun-tag string (TODO: remove this) ]
+     * github      -> [ "name@github:user/repo", INFO, .bun-tag string (TODO: remove this) ]
+     * ```
+     * */
+    packages: {
+      [package: string]:
+        | {
+            /** npm */
+            1: string;
```
pfgithub: You can use tuple types here with descriptive keys: `[package: string, registry: string, info: BunLockFilePackageInfo, integrity: string]`

also typescript 5.7.2 doesn't like `[package: string]`, it says "'package' is a reserved word in strict mode"

RiskyMH: I chose to the object method as it was behaving better with strict types for me and merging less... I can try again though.

For package,  I didn't know and will fix.

RiskyMH: Oh wow the tuple types here with descriptive keys work really well, I have implemented!

(before I think I used object so I could add `length:` attribute, but then I forgot and the types got even more different where you cant just rely on length for the type)

src/fs.zig
```diff
@@ -1608,8 +1608,13 @@ pub const PathName = struct {
 
         // Strip off the extension
         if (strings.lastIndexOfChar(base, '.')) |dot| {
-            ext = base[dot..];
-            base = base[0..dot];
+            if (strings.endsWith(base, "bun.lock")) {
```
dylan-conway: should we make sure it's the entire file name? like:
```suggestion
            if (strings.endsWith(base, std.fs.path.sep_str ++ "bun.lock")) {
```

dylan-conway: or actually why are we setting extension to `"bun.lock"`?

RiskyMH: I tried to go around it better, but the way it determines what to do with the import, it seems to everywhere use the extension, so this felt like the simplest and safest method.

But it comes with its own issues so I'm not 100% sure if its good, just couldn't work out any better way.

RiskyMH: btw i found out `base` was just filename, so was easy to ensure entire file name

RiskyMH: > or actually why are we setting extension to `"bun.lock"`?

oops i didn't look hard enough into the source, i have found a way safer option now. Thanks!

packages/bun-types/bun.d.ts
```diff
@@ -6421,6 +6421,66 @@ declare module "bun" {
...
     timestamp?: number | Date,
   ): Buffer;
+
+  /**
+   * Types for `bun.lock`
+   */
+  type BunLockFile = {
+    lockfileVersion: 0;
+    workspaces: {
+      [workspace: string]: {
```
DonIsaac: Consider making this a separate type so others can import it in their own code

RiskyMH: good idea, i just wasn't sure on naming and `BunLockFIle["workspaces"]["anything"]` type thing is something i have used too much sadly.

I'll try to make the types here more abstracted and reusable

RiskyMH: done

Issue #16256, opened 2025-01-08
Spooky function call during makeSafe(Map) in <internal:primordeals> causing catastrophic error.


janbiedermann:
### What version of Bun is running?

1.1.43

### What platform is your computer?

all of them

### What steps can reproduce the bug?

// with a recent ruby installed
git clone git@github.com:opal/opal.git
cd opal
bin/setup
bundle exec rake mspec_ruby_bun

### What is the expected behavior?

Should run specs instead of exiting with catastrophic error.

### What do you see instead?

```
FrozenError: can't modify frozen Hash: {}
    cause: ::NilClass {
  $$id: 4,
  apply: [Function],
  call: [Function],
  $$frozen: true,
  $$comparable: false,
  $$class: [Function: ::NilClass],
  $$meta: [Function: ::NilClass],
  "$!": [Function: $!],
  "$&": [Function: $&],
  "$|": [Function: $|],
  "$^": [Function: $^],
  "$==": [Function: $==],
  $dup: [Function: $dup],
  $clone: [Function: $clone],
  $inspect: [Function: $inspect],
  "$nil?": [Function: $nil?],
  $singleton_class: [Function: $singleton_class],
  $to_a: [Function: $to_a],
  $to_h: [Function: $to_h],
  $to_i: [Function: $to_i],
  $to_s: [Function: $to_s],
  $to_c: [Function: $to_c],
  $rationalize: [Function: $rationalize],
  $to_r: [Function: $to_r],
  $instance_variables: [Function: $instance_variables],
  $to_f: [Function: $to_f],
  toString: [Function],
  "$=~": [Function: $=~],
  "$!~": [Function: $!~],
  "$===": [Function: $===],
  "$<=>": [Function: $<=>],
  $method: [Function: $method],
  $methods: [Function: $methods],
  $public_methods: [Function: $public_methods],
  $Array: [Function: $Array],
  $at_exit: [Function: $at_exit],
  $caller: [Function: $caller],
  $caller_locations: [Function: $caller_locations],
  $class: [Function: $class],
  $copy_instance_variables: [Function: $copy_instance_variables],
  $copy_singleton_methods: [Function: $copy_singleton_methods],
  $initialize_clone: [Function: $initialize_clone],
  $define_singleton_method: [Function: $define_singleton_method],
  $initialize_dup: [Function: $initialize_dup],
  $enum_for: [Function: $enum_for],
  "$equal?": [Function: $equal?],
  $exit: [Function: $exit],
[...]

PR #16259, opened 2025-01-08
Guard makeSafe() against failing speculative calls

janbiedermann:
### What does this PR do?

Fix for #16256

Guards <internal:primordeals> `makeSafe()` against failing speculative function calls, which can happen if the prototypes of Set or Map have been modified, e.g. functions added to the prototypes.

### DIFF
--- a/src/js/internal/primordials.js
+++ b/src/js/internal/primordials.js
@@ -47,20 +47,24 @@ const makeSafe = (unsafe, safe) => {
     let next; // We can reuse the same `next` method.
 
     ArrayPrototypeForEach(Reflect.ownKeys(unsafe.prototype), key => {
-      if (!Reflect.getOwnPropertyDescriptor(safe.prototype, key)) {
-        const desc = Reflect.getOwnPropertyDescriptor(unsafe.prototype, key);
-        if (typeof desc.value === "function" && desc.value.length === 0) {
-          const called = desc.value.$call(dummy) || {};
-          if (Symbol.iterator in (typeof called === "object" ? called : {})) {
-            const createIterator = uncurryThis(desc.value);
-            next ??= uncurryThis(createIterator(dummy).next);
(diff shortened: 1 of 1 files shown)


DonIsaac:
I don't think this is the correct approach.

The point of `makeSafe` is to protect internal js code against prototype pollution. Since JS code is loaded in a _very_ lazy way, it is possible to pollute the prototype before primordials are instantiated by polluting before any dependent internal module `e.g. require("node:assert")`.

In the short term, you should be able to fix your linked bugs by making sure you import the internal modules you need before prototype pollution occurs. On our end, we may need to take a more defensive approach and eagerly instantiate SafeMap and SafeSet on startup.

cc @paperclover

janbiedermann:
Alright, thanks for the quick response. I will experiment with that idea (early import of some modules) and report back. Gimme a few days or so.

paperclover:
js code should just never mutate the global prototype, because it always makes the code more confusing. since that ask is unfortunately not realistic, here is how to fix SafeSet:

I noticed that Bun already has the ability to use Set.prototype and Map.prototype without being affected by prototype pollution, WebKit already snapshots things on startup for their own builtins (see WebKit: `SetPrototype::finishCreation` for example)

for us to use these, we can edit replacements.ts and add `Map` and `Set` to the list `export const globalsToPrefix = [`.

for all `SafeSet`/`SafeMap` usages, replace with `new Set` / `new Map`, and then replace all method calls to prefix with $, so `set.has` -> `set.$has`. this is equivilent to in JSC internals they always use `new @Sap` and `set.@has`. our builtins use $ because it is part of a valid identifier and we are okay with running a light preprocessor.

then SafeSet and SafeMap can be deleted. this is a more robust and performant solution to the primordials.js file.

janbiedermann:
Closing in favor of https://github.com/oven-sh/bun/issues/16300.

Issue #16255, opened 2025-01-08
Bun S3 acl option does not work


Gobd:
### What version of Bun is running?

1.1.43+76800b049

### What platform is your computer?

Darwin 24.2.0 arm64 arm

### What steps can reproduce the bug?

```    
const res = await Bun.write(
        Bun.s3(key, {
            accessKeyId,
            secretAccessKey,
            sessionToken,
            bucket,
            region: 'us-east-1',
            acl: 'public-read',
            type: filetype,
        }),
        body
    );
```

gives `There were headers present in the request which were not signed` comment out the acl bit and it works.

Perhaps this block of code is it? https://github.com/oven-sh/bun/blob/2af12f49a81c3629f54446ac3c4ac84dd58e7016/src/s3/credentials.zig#L522

### What is the expected behavior?

It uploads the file without issues

### What do you see instead?

`There were headers present in the request which were not signed`

PR #16260, opened 2025-01-08
Add missing x-amz-acl header

Gobd:
### What does this PR do?

Closes https://github.com/oven-sh/bun/issues/16255

- [x] Code changes

### DIFF
--- a/src/s3/credentials.zig
+++ b/src/s3/credentials.zig
@@ -528,9 +528,9 @@ pub const S3Credentials = struct {
                     }
                 } else {
                     if (session_token != null) {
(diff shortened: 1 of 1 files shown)


DonIsaac:
Thanks for the PR @Gobd. It would be a big help if you could also add some test cases.

Gobd:
Sorry I'm not sure how I'd write a test for this :(

Gobd:
@cirospaciari using `bunx bun-pr 16260` I can confirm that this fixes the issue

cirospaciari (approved):
Thank you! I missed this in a merge will add someway of testing this so we dont have regretions, your fix is correct in fact.

PR #9260, opened 2024-03-05
Update cache.md

ryands17:
### What does this PR do?

Specify `bunfig.toml` to make it easy to check where this option needs to be added.

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #16263, opened 2025-01-08
fix(js): better error messages when makeSafe fails

DonIsaac:
### What does this PR do?
Partial redress of https://github.com/opal/opal/issues/2706. Related to #16259.

`makeSafe` can fail when creating safe iterables if users modify the prototype of `Set` or `Map` before primordials are loaded. This PR adds a better error message for such cases, but does nothing to address the underlying problem. We still need to decide on the best course of action there.

I've added tests.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -505,6 +505,18 @@
       // Don't pause when the GC runs while the debugger is open.
       "postRunCommands": ["command source '${workspaceFolder}/misctools/lldb/lldb_commands'"],
     },
+    {
+      "type": "bun",
+      "request": "launch",
+      "name": "JS: bun test [file]",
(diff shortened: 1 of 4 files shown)


robobun:
<div><sup>Updated 3:48 PM PT - Jan 8th, 2025</sup></div>

@DonIsaac, your commit 9974369cad011b46ec432a8f1049ef3fc242528c is building: [`#9260`](https://buildkite.com/bun/bun/builds/9260)

Jarred-Sumner:
We should just delete these wrappers and instead make JSC have private symbols for what we need 

these wrappers are a skill issue

paperclover:
see https://github.com/oven-sh/bun/pull/16259#issuecomment-2578853256

paperclover (changes requested):
no

PR #16267, opened 2025-01-09
fix(install): hoisting bugs causing flaky tests in CI

### LINKED ISSUE #1343: Deduplicate / dedupe command for bun install

### LINKED ISSUE #13533: `bun install` routinely accumulates package duplication

1.1.26

Darwin 23.6.0 arm64 arm

When updating package versions, either manually in the `package.json` and running `bun install`, or by running `bun update`, Bun routinely starts duplicating packages in the dependency tree and then continues to accumulate duplicated dependencies over time, entirely un-necessarily.

This happens silently until we realise what has been happening due to obscure runtime bugs or TypeScript type checking errors.

Now, we always need to blast away the entire installation and lockfile and install from scratch every time to avoid these obscure problems.

Reproduction:

Consider then packages
- `@myorg/dep1@1.0.89`
- `@myorg/dep2@1.0.89` depends on `@myorg/dep1@^1.0.89`

When installing, we get a good tree like this:

```
% bun pm ls --all
./myproject/core node_modules
├── @myorg/dep1@1.0.89
├── @myorg/dep2@1.0.89
```

Now, `@myorg/dep1` is updated to `1.0.90`, and we update the dependency, and the tree becomes:

```
% bun pm ls --all
./myproject/core node_modules
├── @myorg/dep1@1.0.90
├── @myorg/dep2@1.0.89
│   └── @myorg/dep1@1.0.89
```

This prolific duplication seems crazy.

I expect Bun should only duplicate the dependency tree when absolutely necessary to maintain semver compatibility. Dependencies should always be deduplicated/hoisted except where the hoisted version cannot meet semver requirements.

TL;DR prioritise having a single version of each dependency.

### What do you see instead?

See above.

### Additional information

[`npm update`](https://docs.npmjs.com/cli/v10/commands/npm-update#subdependencies)

> npm will prioritize having a single version of [dependencies] in your tree rather than two when that single version can satisfy the semver requirements of multiple dependencies in your tree.

[`npm install`](https://docs.npmjs.com/cli/v10/commands/npm-install#install-strategy)

> hoisted (default): Install non-duplicated in top-level, and duplicated as necessary within directory structure

### LINKED ISSUE #16527: Inconsistent `bun.lock` when doing lockfile maintenance

1.1.45+196621f25

Linux 5.15.167.4-microsoft-standard-WSL2 x86_64 x86_64

- Have multiple sub dependencies using different versions of a sub-sub dependency.
- Do a lockfile maintenance : `rm -rf bun.lock && bun install`

Consistent `bun.lock`

### What do you see instead?

I don't know how to explain the bug, but I've 3 sub-dependencies using theirselves a subdependency named `xml-name-validator`.
- 2 of them use `v5.0`
- 1 of them uses `v4.0`

And when I do a lockfile maintenance `rm -rf bun.lock && bun install`, I'm randomly having `bun` switching the versions, this is probably due to some parallelism or cache.

So sometimes it hoists `v5.0` at top level `bun.lock` and hoists `v4.0` to the sub dependency that is using a downgraded version compared to its siblings using `v5.0`.

Image

Image

Image

Sometimes it hoists  `v4.0` at top level `bun.lock`  and hoists `v5.0` to the 2 sub dependencies that is using an upgraded version compared to its sibling using `v4.0`

Image

Image

Image

Image

Maybe I'm skewed by not using `--no-cache` option and that it sometimes read different caches, or skipping it because some package updates, then when redoing lock file maintenance it uses old cache (not saving new cache after package update?).

Still, with the `--no-cache --force`, i'm still having inconsistent updates, so it must be caused by random factor such as parallelism.

dylan-conway:
### What does this PR do?
This fixes a bug causing resolving multiple versions of the same package to be out of order, resulting in an inconsistent amount of versions of the package to appear in the lockfile.

This also changes `bun install` to deduplicate more packages, similar to npm where having one copy of a package is prioritized.

fixes #13533
fixes #16527 
closes #1343


robobun:
<div><sup>Updated 5:09 PM PT - Jan 8th, 2025</sup></div>

@dylan-conway, your commit 9ae235cc0e820f6e269de8be515336cf435a5956 is building: [`#9270`](https://buildkite.com/bun/bun/builds/9270)

kevinmarrec:
@dylan-conway Any update on this potential fix of https://github.com/oven-sh/bun/issues/16527 ?

Thanks !

PR #16276, opened 2025-01-09
Fixes #14553

### LINKED ISSUE #14553: Bun not installing private packages from registry.npmjs.org when using .npmrc in some cases

1.1.30+7996d06b8

Linux 6.9.3-76060903-generic x86_64 x86_64

Create a private package in NPM registry 
ex. "@myorg/mypackage"

setup auth token in .npmrc file
`
//registry.npmjs.org/:_authToken=<AUTH_TOKEN>
`

try to `bun add @myorg/mypackage` 

bun will install packages from a private registry 

### What do you see instead?

[package not found 404](error: GET https://registry.npmjs.org/@myorg%2fmypackage - 404)

### Additional information

_No response_

robertshuford:
### What does this PR do?

Fixes https://github.com/oven-sh/bun/issues/14553

The auth configuration was not applied to a scoped registry if a default registry had the same host as a scoped registry. 

With this change auth configuration will be applied to all registries with a matching host.

- [x] Code changes

### DIFF
--- a/src/ini.zig
+++ b/src/ini.zig
@@ -1168,6 +1168,7 @@ pub fn loadNpmrc(
             const conf_item_url = bun.URL.parse(conf_item.registry_url);
 
             if (std.mem.eql(u8, bun.strings.withoutTrailingSlash(default_registry_url.host), bun.strings.withoutTrailingSlash(conf_item_url.host))) {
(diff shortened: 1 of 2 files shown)


dylan-conway (approved):
Looks good, thank you

Issue #9343, opened 2024-03-10
Do not inline variables when debugging


FreePhoenix888:
### What is the problem this feature would solve?

When debugging I want to use variables but they are inlined therefore they are considered as not declared
![image](https://github.com/oven-sh/bun/assets/66206278/23cd8042-4481-4666-95dd-cfdd35373fe0)
- I cannot use them in Debug Console
- I cannot set breakpoint on inlined lines (it just moves to the line which is not inlined)

### What is the feature you are proposing to solve the problem?

Do not inline variables when debugging

### What alternatives have you considered?

Pass such variables to log to prevent them from inlining

Jarred-Sumner:
We don’t inline variables when debugging — this is probably a VSCode extension bug. Do the variables show up in the web inspector UI?

FreePhoenix888:
> We don’t inline variables when debugging — this is probably a VSCode extension bug. Do the variables show up in the web inspector UI?

Where can I find it? What is web inspector UI?

FreePhoenix888:
Note: maybe the problem is in the web version of bun vscode extension? https://open-vsx.org/extension/oven/bun-vscode

FreePhoenix888:
To reproduce you can go here https://github.com/gitpod-samples/template-bun and click Open in Gitpod button, extension will be installed there and you can debug your code by using web version of bun vscode extension

FreePhoenix888:
I should check whether this is reproducible in non-web version of extension

Jarred-Sumner:
If you run `bun --inspect script.ts` it will print a URL you can go to the web inspector

FreePhoenix888:
> If you run `bun --inspect script.ts` it will print a URL you can go to the web inspector

There is a problem with it: https://discord.com/channels/876711213126520882/1215700688290717767
I would be happy if you check it and help me :)

Electroid:
This is likely a bug with the VSCode debugger.

FreePhoenix888:
As I remember this does not happen when just debugging js. Not sure about ts. Anyway I should check it

RiskyMH:
@FreePhoenix888 I believe I fixed this in #14995 (released in `v1.1.35`), please let me know if disables minification for you too.

FreePhoenix888:
Thank you, @RiskyMH! 👍🏻

PR #16303, opened 2025-01-10
Make fetch() faster at uploading files over http://

Jarred-Sumner:
### What does this PR do?

Make fetch() faster at uploading files over http://

### DIFF
--- a/src/http.zig
+++ b/src/http.zig
@@ -1048,8 +1048,69 @@ pub const HTTPThread = struct {
 
     has_awoken: std.atomic.Value(bool) = std.atomic.Value(bool).init(false),
     timer: std.time.Timer,
-
     lazy_libdeflater: ?*LibdeflateState = null,
+    lazy_request_body_buffer: ?*HeapRequestBodyBuffer = null,
+
+    pub const HeapRequestBodyBuffer = struct {
+        buffer: [512 * 1024]u8 = undefined,
+        fixed_buffer_allocator: std.heap.FixedBufferAllocator,
+
+        pub usingnamespace bun.New(@This());
+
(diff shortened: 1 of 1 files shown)


robobun:
<div><sup>Updated 6:24 PM PT - Jan 9th, 2025</sup></div>

@Jarred-Sumner, your commit 3f19d2aa0c808bc8e95f9a8ab550d41bd27bd5cf is building: [`#9343`](https://buildkite.com/bun/bun/builds/9343)

src/http.zig
```diff
@@ -3077,11 +3185,12 @@ pub fn onWritable(this: *HTTPClient, comptime is_first_call: bool, comptime is_s
...
     switch (this.state.request_stage) {
         .pending, .headers => {
-            var stack_fallback = std.heap.stackFallback(16384, default_allocator);
-            const allocator = stack_fallback.get();
-            var list = std.ArrayList(u8).initCapacity(allocator, stack_fallback.buffer.len) catch unreachable;
-            defer if (list.capacity > stack_fallback.buffer.len) list.deinit();
-            const writer = &list.writer();
+            var request_body_buffer = this.getRequestBodySendBuffer();
+            defer request_body_buffer.deinit();
+            var temporary_send_buffer = request_body_buffer.toArrayList();
```
cirospaciari, line 3190: we have a case that we call this.onWritable(false, is_ssl, socket); and in this case we would call getRequestBodySendBuffer twice before calling deinit on it

https://github.com/oven-sh/bun/blob/3f19d2aa0c808bc8e95f9a8ab550d41bd27bd5cf/src/http.zig#L3310

cirospaciari (approved):

PR #9407, opened 2024-03-14
improve `--print` with promises

dylan-conway:
### What does this PR do?
enables ```bun-debug --print "Bun.$`bat CMakeList.txt`.text()"```

### How did you verify your code works?
tested manually

PR #16324, opened 2025-01-11
Switch bun repo to use bun.lock

pfgithub:
### What does this PR do?

Might require updating the bun version in CI

- [ ] Wait for CI bun version update
- [ ] `bun run prettier:extra` re-added a bun.lockb. update its BUN_VERSION to 1.1.34


robobun:
<div><sup>Updated 6:37 PM PT - Jan 10th, 2025</sup></div>

@pfgithub, your commit 70f56d41b0893b4a45822d2d4eaa2a0ca5b5bf58 is building: [`#9407`](https://buildkite.com/bun/bun/builds/9407)

fel1x-developer:
We need to add following to `bunfig.toml` ([source](https://bun.sh/docs/runtime/bunfig#install-savetextlockfile)) at least until we release v1.2.0

```toml
[install]
saveTextLockfile = true
```

RiskyMH:
> We need to add following to `bunfig.toml` ([source](https://bun.sh/docs/runtime/bunfig#install-savetextlockfile)) at least until we release v1.2.0
> 
> ```toml
> [install]
> saveTextLockfile = true
> ```

I'm not sure why everyone is adding this on their updates, but if bun sees `bun.lock` it will update/use it. Only time this config would be used is if you delete file and want to make again.

fel1x-developer:
> > We need to add following to `bunfig.toml` ([source](https://bun.sh/docs/runtime/bunfig#install-savetextlockfile)) at least until we release v1.2.0
> > ```toml
> > [install]
> > saveTextLockfile = true
> > ```
> 
> I'm not sure why everyone is adding this on their updates, but if bun sees `bun.lock` it will update/use it. Only time this config would be used is if you delete file and want to make again.

This can prevent situations where a developer accidently deletes bun.lock and executes `bun install`. Even though we can correct wrong lockfiles in PR review stage, adding these two lines makes the review simpler and more efficient.

.gitattributes
```diff
@@ -19,6 +19,7 @@
 *.mts text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 
 *.lockb binary diff=lockb
+bun.lock linguist-language=json
```
RiskyMH, line 22: I added as temporary measure as it makes the files look better (wait until https://github.com/github-linguist/linguist/pull/7193 and then _can_ remove)

PR #16344, opened 2025-01-12
Consolidate cli command code into a separate folder

mchoo7:
### What does this PR do?

Currently, commands that are directly and indirectly executed coexist in `src/cli` folder. This has two problems:

1. There are too many files with file name ending `*_command.zig`
2. When adding a new command or importing files under `src/cli`, it is hard to know if a command is directly executed by users. (e.g. run is directly executed by `bun run` while pack does not. There is no `bun pack` command)

When there is `bun <command>`, the code that executes the command is in `src/cli/<command>.zig`, with exception of `bunx`/`bun x` which is under `src/cli/bunx.zig`, not `src/cli/x.zig`.

- [x] Code changes

Build success on Ubuntu 24.04.1 LTS WSL kernel 6.12.8


RiskyMH:
image
there seems to be build fails https://buildkite.com/bun/bun/builds/9461#019457ef-3ea6-4cda-b536-d2c5544b9d39

scripts/install.sh
RiskyMH, line 1: moving these files may cause issues with how the site redirects and other linkings to github raw

mchoo7: I don't have access to the code for https://bun.sh because it is closed source.
For documentation files, I finished following `grep` in bun repository.

```console
$ grep -inr --include \*.md "_command.zig"
$ grep -inr --include \*.md "src/cli/"
```

Issue #9464, opened 2024-03-16
nextjs fails to run in bun


philipheinser:
### What version of Bun is running?

1.0.31

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

nextjs project that fetches json form endpoints:

running in node works every time

running with `bun --bun run dev` it fails every few times when refreshing the times. sometimes work sometimes fails.

### What is the expected behavior?

site to load

### What do you see instead?

```
SyntaxError: Unexpected end of JSON input
    at <parse> (:0)
    at json (native)
    at <anonymous> (:62:32)
    at processTicksAndRejections (:12:39)
```

Electroid:
Thanks for reporting! Could you share your package.json and code so we can reproduce this? If you're able to narrow down to a minimal reproduction that would be great.

Yasamato:
I'm experiencing something similar, the cause being that fetch sometimes (2-3 out of 10 calls) has an empty response body and as such failing the json decode.

Running it via `bun test-fetch.ts` is unable to reproduce the error. I only observed this error in combination of an next.js app, but I am unable to pinpoint which version (nextjs or bun) did have a regression.

Below instructions seem to work... Only if I switch the api endpoint to my server it breaks, maybe something wrong with my server instead of the code/bun/nextjs

-----

```bash
bunx create-t3-app
```
Create a server side function in `/src/lib/util.ts`

```ts
'use server'

export async function fetchApi<T>(
  method: 'GET' | 'POST' | 'DELETE',
  path: string,
  data?: object,
) {
  const url = "https://meowfacts.herokuapp.com/?" + path

  return fetch(url, {
    method: method,
    cache: 'no-store',
    headers: {
      "Content-Type": "application/json"
    },
    body: data ? JSON.stringify(data) : null,
  })
    .then((res) => {
      if (!res.ok) {
        throw new Error(
          `[api]: ${res.status} ${method} ${path} responded with error: ${res.statusText}`,
        )
      }

      return res
        .json()
        .then((resData) => {
          console.log("Response body has", Object.keys(resData), typeof(resData))
          return resData as T
        })
        .catch(() => {
          throw Error(
            `[api]: ${res.status} ${method} ${path} failed to decode JSON response`,
          )
        })
    })
    .catch((e) => {
      throw Error(`[api]: ${method} ${path} fetch failed due to ${e}`)
    })
}
```

Create a client component in `/src/components/SearchDemo.tsx`
```ts
[...]

PR #16346, opened 2025-01-12
Probably reduce memory usage of JSON IPC

Jarred-Sumner:
### What does this PR do?

Previously, we often cloned JSON strings an extra time over IPC. 

Before merging, need to verify this does in fact reduce memory usage because currently this is just an educated guess

Added a test that roundtrip encodes & decodes latin1, ascii and utf-16 for IPC


robobun:
<div><sup>Updated 6:32 PM PT - Jan 11th, 2025</sup></div>

@Jarred-Sumner, your commit 534eeb85e7264b0b85fd083a8b44a08534f8efac is building: [`#9464`](https://buildkite.com/bun/bun/builds/9464)

PR #16349, opened 2025-01-12
Rewrite the internal Web Stream native bindings to use less memory 

### LINKED ISSUE #12198: Streams memory leak

Jarred-Sumner:
### What does this PR do?

Fixes #12198
Closes [#16075](https://github.com/oven-sh/bun/issues/16075)

The usage of `type: "bytes"` in our lazy internal ReadableStream sources led to a memory leak because each read from the stream sometimes returns a larger & different buffer than the one the user provides. 

The version using Node APIs uses the same amount of memory as the version using Bun/Web APIs.

Will follow up with a memory leak test once a release build is generated in CI.

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3056,6 +3056,9 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         }
         fn endRequestStreaming(this: *RequestContext) bool {
             assert(this.server != null);
+
+            this.request_body_buf.clearAndFree(bun.default_allocator);
+
             // if we cannot, we have to reject pending promises
             // first, we reject the request body promise
             if (this.request_body) |body| {
@@ -3069,6 +3072,8 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             return false;
         }
         fn detachResponse(this: *RequestContext) void {
+            this.request_body_buf.clearAndFree(bun.default_allocator);
+
             if (this.resp) |resp| {
                 this.resp = null;
 
--- a/src/bun.js/webcore/streams.zig
+++ b/src/bun.js/webcore/streams.zig
@@ -4233,7 +4233,7 @@ pub const FileReader = struct {
 
     pub fn onReadChunk(this: *@This(), init_buf: []const u8, state: bun.io.ReadState) bool {
         var buf = init_buf;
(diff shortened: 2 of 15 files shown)


robobun:
<div><sup>Updated 2:26 AM PT - Jan 12th, 2025</sup></div>

@Jarred-Sumner, your commit c61b4b6436fa289bb701c85fa4c9db587ce93c0a is building: [`#9471`](https://buildkite.com/bun/bun/builds/9471)

leeoniya:
totally drive-by comment, but a PR title like "Streams: Fix memory leak with `type: "bytes"`" would be much nicer than "Fixes #12198". does the tl;dr without clicking around for context :)

DonIsaac:
Additionally, JSC's GC was not being informed of allocations in `BufferedReader`. Since the GC didn't know memory pressure was increasing, it never knew it needed to sweep old objects out of the heap. This is now fixed.

PR #16350, opened 2025-01-12
Update ssr-react.md to reference normal non-beta react

rgarcia:
### What does this PR do?

Updates this doc to reflect that react 19 is the latest non-beta version of react.

I ran through the steps in the docs on my machine and it worked.


RiskyMH:
Similar to #15613, but this is still a slightly different take.

PR #9480, opened 2024-03-17
keep `SQLITE_OPEN_READWRITE` options is `{}` or `{ create: false }`

camc314:
### What does this PR do?

Closes #9410

Previously, an `SQLiteError` was thrown:
```
SQLiteError: flags must include SQLITE_OPEN_READONLY or SQLITE_OPEN_READWRITE
```

as netheir `SQLITE_OPEN_READONLY` or `SQLITE_OPEN_READWRITE` was set then options was `{}` or `{ create: false }`

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?

- [x] I included a test for the new code, or existing tests cover it
- [x] I ran my tests locally and they pass (`bun-debug test test-file-name.test`)

PR #16355, opened 2025-01-12
Convert request.body into Readable in node:http

Jarred-Sumner:
### What does this PR do?

This makes node:http go through the node:stream code path instead of the ReadableStream code path for request bodies

Need to check the memory & http req/s impact of this change before it merges


robobun:
<div><sup>Updated 9:21 AM PT - Jan 12th, 2025</sup></div>

@Jarred-Sumner, your commit a6b7f49bdd709b26e2c85d1e170216d358656ee9 is building: [`#9480`](https://buildkite.com/bun/bun/builds/9480)

PR #16365, opened 2025-01-12
format: replace `@import("./foo.zig")` with `@import("foo.zig")`

mchoo7:
`@import("./foo.zig")` and `@import("foo.zig")` are equivalent, and ziglang/zig uses `@import("foo.zig")` in their lib/std since `./` is redundant.

Even in C++, this is not the case. `#include "foo.h"` is common, whereas `#include "./foo.h"` is not.

### What does this PR do?

- [x] Code changes


paperclover:
instead of manually fixing this, this type of style thing should be enforced through a linter, or else the problem will just come back.

fel1x-developer:
> instead of manually fixing this, this type of style thing should be enforced through a linter, or else the problem will just come back.

This cannot be fixed by linter yet because neither `zig fmt` and vscode-zig's auto-format support this case. I created an issue in ziglang/vscode-zig(https://github.com/ziglang/vscode-zig/issues/379) to support this feature. Until zig or vscode have a feature to format this, we should check manually if the correct syntax is used when reviewing new PRs.

#16366 suggests making a coding style for Bun project, and I have written a section about this.

fel1x-developer:
@paperclover Could you give a try for CI run?

PR #16366, opened 2025-01-13
Add coding-styles.md

mchoo7:
### What does this PR do?

- [ ] Code changes


paperclover:
thank you for writing this up, but i don't think this document is fully accurate or high quality (there are multiple spelling errors, for example). this is the sort of thing that one of the core team members writes. when this is written, it should probably go alongside https://bun.sh/docs/project/contributing. when i have a free weekend i'll write it up, as there is a lot more to say.

Issue #13346, opened 2024-08-16
`@eslint/eslintrc` `FlatCompat` / `import-fresh` throws an error in Bun


controversial:
### What version of Bun is running?

1.1.24+85a329911

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

module.parent or module.main or module.children properties are always undefined 

---

The `FlatCompat` class exported by `@eslint/eslintrc` is designed to help “legacy” eslint configs (pre-flat config) work with eslint 9+.

The following produces an error in bun:
```ts
import { FlatCompat } from '@eslint/eslintrc';
const compat = new FlatCompat();
const config = compat.extends('eslint-config-airbnb'); // or any “legacy” eslint config
console.log(config);
```

### What is the expected behavior?

```js
// node test.js
[
  {
    rules: {
      'accessor-pairs': 'off',
      'array-callback-return': [Array],
      'block-scoped-var': 'error',
      complexity: [Array],
      'class-methods-use-this': [Array],
      'consistent-return': 'error',
      curly: [Array],
      'default-case': [Array],
...
```

### What do you see instead?

An error from within the [`import-fresh`](https://www.npmjs.com/package/import-fresh) module

```
// bun test.js

27 | 
28 |    delete require.cache[filePath]; // Delete module from cache
29 | 
30 |    const parent = require.cache[parentPath]; // If `filePath` and `parentPath` are the same, cache will already be deleted so we won't get a memory leak in next step
31 | 
32 |    return parent === undefined ? require(filePath) : parent.require(filePath); // In case cache doesn't have parent, fall back to normal require
                                                               ^
TypeError: Cannot read config file: /Users/luke/Developer/misc/bun-eslint-compat/node_modules/eslint-config-airbnb/index.js
Error: parent.require is not a function. (In 'parent.require(filePath)', 'parent.require' is undefined)
Referenced from: 
      at /Users/luke/Developer/misc/bun-eslint-compat/node_modules/import-fresh/index.js:32:59
      at loadJSConfigFile (/Users/luke/Developer/misc/bun-eslint-compat/node_modules/@eslint/eslintrc/lib/config-array-factory.js:237:16)
      at _loadConfigData (/Users/luke/Developer/misc/bun-eslint-compat/node_modules/@eslint/eslintrc/lib/config-array-factory.js:638:42)
[...]

PR #16377, opened 2025-01-13
Update nextjs.md for proper build command

geowarp:
This pull request includes a change to the `docs/guides/ecosystem/nextjs.md` file to add instructions for building a project with Bun.

### Documentation update:

* [`docs/guides/ecosystem/nextjs.md`](diffhunk://#diff-aab0c778a60b8cef1d760a2004128d006915488d93f8b9f072a04ca8a0279e60R30-R38): Added a new section with instructions on how to build a project using Bun by running `bun run build` from the project root.

Given that to start the dev server you needed the `--bun` flag inside the `bun --bun run dev` command to utilize only Bun and not Node for runtime, I naturally thought the build process would follow a similar command pattern, such as `bun --bun run build`. However, that command introduced the following linting error:

```
Linting and checking the validity of types  .. ⨯ ESLint: Cannot read config file: /Users/geowhale/Desktop/dev-archive/miniqtrade/miniqt-fe/node_modules/eslint-config-next/core-web-vitals.js Error: parent.require is not a function. (In 'parent.require(filePath)', 'parent.require' is undefined) Referenced from
```

Therefore, I believe it is necessary to specifically point out that to build a project with Bun it should simply be `bun run build`. Running `bun run build` works great and passes through the linting checks during the build process.

### DIFF
--- a/docs/guides/ecosystem/nextjs.md
+++ b/docs/guides/ecosystem/nextjs.md
@@ -27,6 +27,15 @@ $ bun create next-app --example with-supabase
 
 ---
 
+To build a project with Bun, run `bun run build` from the project root.
+
+```sh
+$ cd my-app
+$ bun run build
+```
+
+---
+
 To start the dev server with Bun, run `bun --bun run dev` from the project root.
 
 ```sh


paperclover:
we should not add this to the documentation, but instead fix #13346

PR #16381, opened 2025-01-13
feat(sql) transactions, savepoints, connection pooling and reserve

cirospaciari:
### What does this PR do?

- [X] Code changes

Add support for transactions:

#### BEGIN / COMMIT `await sql.begin([options = ''], fn) -> fn()`

```js
import { SQL } from "bun:sql";

const sql = new SQL({
  host: "localhost",
  port: 5432,
  user: "ciro",
  password: "bunbunbun",
  database: "bun",
});

const [user, account] = await sql.begin(async sql => {
  const [user] = await sql`
    insert into users (
      name
    ) values (
      'Murray'
    )
    returning *
  `

  const [account] = await sql`
    insert into accounts (
      user_id
    ) values (
      ${ user.user_id }
    )
    returning *
  `

  return [user, account]
})
```
#### SAVEPOINT `await sql.savepoint([name], fn) -> fn()`

```js
sql.begin('read write', async sql => {
  const [user] = await sql`
    insert into users (
      name
    ) values (
      'Murray'
    )
  `

  const [account] = (await sql.savepoint(sql =>
    sql`
      insert into accounts (
        user_id
      ) values (
        ${ user.user_id }
      )
    `
  ).catch(err => {
    // Account could not be created. ROLLBACK SAVEPOINT is called because we caught the rejection.
  })) || []

  return [user, account]
})
.then(([user, account]) => {
  // great success - COMMIT succeeded
})
.catch(() => {
  // not so good - ROLLBACK was called
})
```

`ERR_POSTGRES_UNSAFE_TRANSACTION` error is throw if BEGIN command is used without `.begin()`, `max: 1` or `.reserve()`.

Like `postgres` package if returned an array of Promises on .begin or .savepoint it will `Promise.all` all promises before COMMIT.

```js
 await sql.begin(async sql => [
  sql`insert into test values(1)`,
  sql`insert into test values(2)`,
  sql`insert into test values(3)`
]);
 ```
`sql.transaction` is a alias for `sql.begin`

#### RESERVE `await sql.reserve()`
The reserve method pulls out a connection from the pool, and returns a client that wraps the single connection. This can be used for running queries on an isolated connection.

```js
// compatible with `postgres` example
const reserved = await sql.reserve();
await reserved`select * from users`;
await reserved.release();
// with in a production scenario would be something more like
const reserved = await sql.reserve();
try {
  // ... queries
} finally {
  await reserved.release();
}
```
To make it simpler bun supports`Symbol.dispose` and `Symbol.asyncDispose`
```js
{
// always release after context (safer)
using reserved = await sql.reserve()
await reserved`select * from users`
}
```
Calling `.reserve` in a reserved Sql will return a new reserved connection, not the same connection (behavior matches `postgres` package).

#### DISTRIBUTED TRANSACTIONS `await sql.beginDistributed(name, fn) -> fn()`

Also know as Two-Phase Commit, in a distributed transaction, Phase 1 involves the coordinator preparing nodes by ensuring data is written and ready to commit, while Phase 2 finalizes with nodes committing or rolling back based on the coordinator's decision, ensuring durability and releasing locks.

In PostgreSQL and MySQL distributed transactions persist beyond the original session, allowing privileged users or coordinators to commit/rollback them, ensuring support for distributed transactions, recovery, and administrative tasks.

`beginDistributed` will automatic rollback if any exception are not caught, and you can commit and rollback later if everything goes well.

PostgreSQL natively supports distributed transactions using PREPARE TRANSACTION, while MySQL uses XA Transactions, and MSSQL also supports distributed/XA transactions. However, in MSSQL, distributed transactions are tied to the original session, the DTC coordinator, and the specific connection. These transactions are automatically committed or rolled back following the same rules as regular transactions, with no option for manual intervention from other sessions, in MSSQL distributed transactions are used to coordinate transactions using Linked Servers.

```js
await sql.beginDistributed("numbers", async sql => {
  await sql`create table if not exists numbers (a int)`;
  await sql`insert into numbers values(1)`;
});
// later you can call
await sql.commitDistributed("numbers");
// or await sql.rollbackDistributed("numbers");
```

`sql.distributed` is a alias for `sql.beginDistributed`

### Connection Pool
No connection will be made until a query is made.
```
const sql = Bun.sql() // no connection are created

await sql`...` // pool is started until max is reached (if possible), first available connection is used
await sql`...` // previous connection is reused

// two connections are used now at the same time
await Promise.all([
  sql`...`,
  sql`...`
]);

await sql.close(); // close all connection from the pool no new queries can be made.
```
If a connection is closed or not able to connect it will only retry it if no other connection is currently available to be used, this enables Bun to run even if not all connections are able to established minimizing downtimes. Bun will open as many connections until max number of connections is reached. By default max is 10. This can be changed by setting max in the Bun.sql() call. Example - Bun.sql({ max: 20 }).

Differences from `postgres` package:

1 - Error code `UNSAFE_TRANSACTION` is `ERR_POSTGRES_UNSAFE_TRANSACTION`

2 - `postgres` package will accept 0 in `max` value, we will throw a error, the default value will be the same 10, `postgres` will use 0 with means hanging when trying to do a query.

3 - `postgres` is conservative and will lazy open each connection only if enough concurrency is achieved, until `max` is reached, `Bun.sql` try to keep the pool full (until `max` is reached) when possible after the first query is called or when retries are needed to maximize throughput.

4 - Features present in `postgres` and not in `Bun.sql`:
- `sql.readable()`
-  `sql.writable()`
- `sql.cursor`
- `sql.file`
- `sql.simple`
- `sql.listen`
- `sql.notify`
- `sql.subscribe`
- `sql.unsafe`
- `sql.prepare` in transactions see `sql.beginDistributed`
- `options.types`
- `options.transform`

### DIFF
--- a/src/js/bun/sql.ts
+++ b/src/js/bun/sql.ts
@@ -874,7 +874,6 @@ function SQL(o) {
         throw $ERR_POSTGRES_UNSAFE_TRANSACTION("Only use sql.begin, sql.reserved or max: 1");
       }
     }
-
     return createQuery(sqlString, values, new SQLResultArray(), columns);
   }
 
@@ -1142,7 +1141,7 @@ function SQL(o) {
         throw err;
       }
     };
-    let transaction_started = false;
+    let needs_rollback = false;
     try {
       if (options) {
         //@ts-ignore
@@ -1151,13 +1150,14 @@ function SQL(o) {
         //@ts-ignore
         await transaction_sql("BEGIN");
       }
-      transaction_started = true;
+      needs_rollback = true;
       const transaction_result = await callback(transaction_sql);
       await transaction_sql("COMMIT");
+      needs_rollback = false;
       return resolve(transaction_result);
     } catch (err) {
       try {
-        if (!state.closed && transaction_started) {
+        if (!state.closed && needs_rollback) {
           await transaction_sql("ROLLBACK");
         }
       } catch (err) {
--- a/src/js/bun/sql.ts
+++ b/src/js/bun/sql.ts
@@ -238,8 +238,9 @@ init(
 
 function onQueryFinish(onClose) {
   this.queries.delete(onClose);
+  this.pool.release(this);
 }
-class ConnectionWithState {
+class PooledConnection {
   pool: ConnectionPool;
   connection: ReturnType<typeof createConnection>;
   state: "pending" | "connected" | "closed" = "pending";
@@ -248,7 +249,6 @@ class ConnectionWithState {
   onFinish: ((err: Error | null) => void) | null = null;
   canBeConnected: boolean = false;
   connectionInfo: any;
-
   #onConnected(err, _) {
     const connectionInfo = this.connectionInfo;
     if (connectionInfo?.onconnect) {
@@ -353,8 +353,8 @@ class ConnectionWithState {
 class ConnectionPool {
   connectionInfo: any;
 
-  connections: ConnectionWithState[];
-  readyConnections: Set<ConnectionWithState>;
+  connections: PooledConnection[];
+  readyConnections: Set<PooledConnection>;
   waitingQueue: Array<(err: Error | null, result: any) => void> = [];
   poolStarted: boolean = false;
   closed: boolean = false;
@@ -375,11 +375,10 @@ class ConnectionPool {
     this.readyConnections = new Set();
   }
 
-  release(connection: ConnectionWithState) {
+  release(connection: PooledConnection) {
     if (this.waitingQueue.length > 0) {
       // we have some pending connections, lets connect them with the released connection
       const pending = this.waitingQueue.shift();
-
       pending?.(connection.storedError, connection);
     } else {
       if (connection.state !== "connected") {
@@ -509,7 +508,7 @@ class ConnectionPool {
       this.poolStarted = true;
       const pollSize = this.connections.length;
       for (let i = 0; i < pollSize; i++) {
-        this.connections[i] = new ConnectionWithState(this.connectionInfo, this);
+        this.connections[i] = new PooledConnection(this.connectionInfo, this);
       }
       return;
     }
@@ -859,7 +858,6 @@ function SQL(o) {
       pool.release(pooledConnection); // release the connection back to the pool
       return query.reject($ERR_POSTGRES_QUERY_CANCELLED("Query cancelled"));
     }
-
     // bind close event to the query (will unbind and auto release the connection when the query is finished)
     pooledConnection.bindQuery(query, onQueryDisconnected.bind(query));
     handle.run(pooledConnection.connection, query);
--- a/src/js/bun/sql.ts
+++ b/src/js/bun/sql.ts
@@ -508,6 +508,7 @@ class ConnectionPool {
    * @param {boolean} reserved - Whether the connection is reserved, if is reserved the connection will not be released until release is called, if not release will only decrement the queriesUsing counter
    */
   connect(onConnected: (err: Error | null, result: any) => void, reserved: boolean = false) {
+    // TODO: this always reserve a connection, we should only reserve if reserved is true
     if (this.closed) {
       return onConnected(connectionClosedError(), null);
     }
--- a/src/js/bun/sql.ts
+++ b/src/js/bun/sql.ts
@@ -1091,7 +1091,10 @@ function assertValidTransactionName(name: string) {
     throw Error(`Distributed transaction name cannot contain single quotes.`);
   }
 }
-function SQL(o) {
+function SQL(o, e = {}) {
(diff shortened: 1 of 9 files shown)


src/js/bun/sql.ts
```diff
@@ -33,12 +40,14 @@ const _run = Symbol("run");
...
 const _handler = Symbol("handler");
 const PublicPromise = Promise;
+type TransactionCallback = (sql: (strings: string, ...values: any[]) => Query) => Promise<any>;
 
 const {
   createConnection: _createConnection,
   createQuery,
   PostgresSQLConnection,
   init,
+  setPromiseAsHandled, // maybe we should move this to utils.zig
```
Jarred-Sumner: `$markPromiseAsHandled`

src/js/bun/sql.ts
```diff
@@ -228,6 +241,457 @@ init(
...
+      // analyse type of error to see if we can retry
+      switch (this.storedError?.code) {
+        case "ERR_POSTGRES_UNSUPPORTED_AUTHENTICATION_METHOD":
+        case "ERR_POSTGRES_UNKNOWN_AUTHENTICATION_METHOD":
+        case "ERR_POSTGRES_TLS_NOT_AVAILABLE":
+        case "ERR_POSTGRES_TLS_UPGRADE_FAILED":
+        case "ERR_POSTGRES_INVALID_SERVER_SIGNATURE":
+        case "ERR_POSTGRES_INVALID_SERVER_KEY":
+        case "ERR_POSTGRES_AUTHENTICATION_FAILED_PBKDF2":
+          // we can't retry this are authentication errors
```
Jarred-Sumner:
```suggestion
          // we can't retry, these are authentication errors
```

src/js/bun/sql.ts
```diff
@@ -228,6 +241,457 @@ init(
...
+  pool: ConnectionPool;
+  connection: ReturnType<typeof createConnection>;
+  state: "pending" | "connected" | "closed" = "pending";
+  storedError: Error | null = null;
+  queries: Set<(err: Error) => void> = new Set();
+  onFinish: ((err: Error | null) => void) | null = null;
+  canBeConnected: boolean = false;
+  connectionInfo: any;
+  /// reserved is used to indicate that the connection is currently reserved
+  reserved: boolean = false;
```
Jarred-Sumner: Do we need all these booleans? Can we use a single const enum with a number?

src/js/bun/sql.ts
```diff
@@ -523,108 +1007,598 @@ function loadOptions(o) {
...
+  function onQueryDisconnected(err) {
+    // connection closed mid query this will not be called if the query finishes first
+    const query = this;
     if (err) {
       return query.reject(err);
     }
-
-    if (!connected) {
-      return query.reject(storedErrorForClosedConnection || new Error("Not connected"));
+    // query is cancelled when waiting for a connection from the pool
```
Jarred-Sumner, line 1127:
```suggestion
    // The query was cancelled while waiting for a connection from the pool
```

src/js/bun/sql.ts
cirospaciari: yeah, will change to flags in one field

src/js/bun/sql.ts
```diff
@@ -523,108 +1007,598 @@ function loadOptions(o) {
...
+      }
+      let timeout = options?.timeout;
+      if (timeout) {
+        timeout = Number(timeout);
+        if (timeout > 2 ** 31 || timeout < 0 || timeout !== timeout) {
+          throw $ERR_INVALID_ARG_VALUE("options.timeout", timeout, "must be a non-negative integer less than 2^31");
+        }
+        state.closed = true;
+        // no new queries will be allowed
+        await Bun.sleep(timeout * 1000);
```
Jarred-Sumner: This keeps the process alive. We don't want to do that I don't htink.

cirospaciari: yeah will change this implementation so actually works like it should

src/js/bun/sql.ts
```diff
@@ -228,6 +241,526 @@ init(
...
+  canBeConnected = 1 << 0,
+  /// reserved is used to indicate that the connection is currently reserved
+  reserved = 1 << 1,
+  /// preReserved is used to indicate that the connection will be reserved in the future when queryCount drops to 0
+  preReserved = 1 << 2,
+}
+class PooledConnection {
+  pool: ConnectionPool;
+  connection: ReturnType<typeof createConnection>;
+  state: PooledConnectionState = PooledConnectionState.pending;
```
Jarred-Sumner, line 265: technically the flags and the connection state can be the same property since the connection state is 2 bits and the flags are 3 bits and we can go up to 32 bit int

Jarred-Sumner: but its probably fine

PR #16384, opened 2025-01-13
refactor: isolate root.C and root.JavaScriptCore

mchoo7:
### What does this PR do?

Currently,
```zig
const C = @import("root").C;
```
and
```zig
const C = bun.C;
```
are same. This is because src/bun.zig exposes C as a `pub const`. This should be avoided and only
```zig
const C = @import("root").C;
```
should be accepted.

- [x] Code changes


paperclover:
i think all uses of `@import("root").C` should be moved to `bun.C` and not the other way around, since `bun` is the primary namespace in this codebase, not the root file's struct. if anything, the root file should be empty besides `pub const bun = ...;` and the stuff from `main.zig`.

fel1x-developer:
@paperclover Could you give a try for CI check?

fel1x-developer:
@paperclover , why do we need this?

https://github.com/oven-sh/bun/blob/4e193b0ebd8cc49bab8263327623e1043f6b68ab/src/install/windows-shim/bun_shim_impl.zig#L49-L50

This is causing error since I move JSC to be under Bun namespace (`bun.JSC`)

Issue #9526, opened 2024-03-20
experimentalDecorators problem


jhcpeixoto:
### What version of Bun is running?

1.0.33+9e91e137f

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

bun run build.ts

```ts
/// <reference types='bun-types' />
import { existsSync, rmSync } from 'fs';
import pkg from './package.json';

const root = import.meta.dir;

const libDir = root + '/build';
if (existsSync(libDir)) rmSync(libDir, { recursive: true });

// Bundle all with Bun
Bun.build({
    root,
    format: 'esm',
    target: 'bun',
    outdir: libDir,
    minify: {
        whitespace: true,
    },
    entrypoints: ['./src/index.ts', './src'],
    external: Object.keys((pkg as any).dependencies! ?? {}),
}).then(console.log);

// Build type declarations
Bun.spawn(['bun', 'x', 'tsc', '--outdir', libDir], {
    stdout: 'inherit',
    stderr: 'inherit',
});

```

### What do you see instead?

logs: [
    13 | import { Coordinates } from '@models/types';
              ^
error: No matching export in "src/models/types/index.ts" for import "Coordinates"

### Additional information

in tsconfig.json I have this path: "@models/types": ["./src/models/types"],

PR #16385, opened 2025-01-13
fix more node:stream

nektro:


robobun:
<div><sup>Updated 3:39 PM PT - Jan 13th, 2025</sup></div>

@nektro, your commit b9fd28ece2a8e79948c24763a4c2e118cece0fae is building: [`#9526`](https://buildkite.com/bun/bun/builds/9526)

src/bun.js/node/node_util_binding.zig
```diff
@@ -194,3 +189,13 @@ pub fn SplitNewlineIterator(comptime T: type) type {
         }
     };
 }
+
+pub fn normalizeEncoding(globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) bun.JSError!JSC.JSValue {
```
Jarred-Sumner, line 193: this should probably be done in C++ so we can reuse encoding strings instead of creating a new encoding string every time

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3507,6 +3507,14 @@ JSC_DEFINE_HOST_FUNCTION(jsFunctionToClass, (JSC::JSGlobalObject * globalObject,
     return JSValue::encode(jsUndefined());
 }
 
+JSC_DEFINE_HOST_FUNCTION(jsFunctionisBlob, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
```
Jarred-Sumner: can we create a single function for all of these instead of one-off functions for each one?

src/js/internal/errors.ts
```diff
@@ -1,7 +1,34 @@
+const { SafeArrayIterator } = require("internal/primordials");
+
+const ArrayIsArray = Array.isArray;
+const ArrayPrototypePush = Array.prototype.push;
+
+class AbortError extends Error {
```
Jarred-Sumner: this abort error will rarely be used.

can this be the $ABORT_ERR instead?

src/js/internal/primordials.js
```diff
@@ -5,13 +5,15 @@
...
-const createSafeIterator = (factory, next) => {
+const createSafeIterator = (factory, next_) => {
   class SafeIterator {
     constructor(iterable) {
+      console.log("SafeIterator +", factory, next_);
       this._iterator = factory(iterable);
     }
     next() {
-      return next(this._iterator);
+      console.log("SafeIterator.next +", next_, this._iterator);
```
Jarred-Sumner: you left this here

src/js/internal/primordials.js
```diff
@@ -5,13 +5,15 @@
 const ObjectSetPrototypeOf = Object.setPrototypeOf;
 const ObjectFreeze = Object.freeze;
 
-const createSafeIterator = (factory, next) => {
+const createSafeIterator = (factory, next_) => {
   class SafeIterator {
     constructor(iterable) {
+      console.log("SafeIterator +", factory, next_);
```
Jarred-Sumner: you left this here

src/js/internal/stream.ts
```diff
@@ -0,0 +1,113 @@
+"use strict";
+
+const ObjectKeys = Object.keys;
+const ObjectDefineProperty = Object.defineProperty;
+
+const customPromisify = require("node:util").promisify.custom;
```
Jarred-Sumner: maybe we can move this to a separate file so we don't import all of util in here?

src/bun.js/bindings/ZigGlobalObject.cpp
nektro: yes i deleted this in a local commit, its no longer used

src/js/internal/errors.ts
nektro: will check uses

src/js/internal/stream.ts
nektro: good point

src/js/internal/abort_listener.ts
```diff
@@ -0,0 +1,38 @@
...
+const { kResistStopPropagation } = require("internal/shared");
+
+const SymbolDispose = Symbol.dispose;
+
+/**
+ * @param {AbortSignal} signal
+ * @param {EventListener} listener
+ * @returns {Disposable}
+ */
+function addAbortListener(signal, listener) {
```
DonIsaac: Consider using type annotations instead of JSDoc types

nektro: yep! i deleted that locally

src/codegen/generate-classes.ts
```diff
@@ -2101,6 +2106,9 @@ const GENERATED_CLASSES_HEADER = [
 #include "root.h"
 
 namespace Zig {
+
+JSC_DEFINE_HOST_FUNCTION(jsFunctionInherits, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame));
```
Jarred-Sumner:
```suggestion
JSC_DECLARE_HOST_FUNCTION(jsFunctionInherits);
```

Issue #9547, opened 2024-03-21
Bun doesnt pick changes in monorepos packages


kaminskypavel:
### What version of Bun is running?

1.0.34-canary.20+b24d3ba5b

### What platform is your computer?

windows x64

### What steps can reproduce the bug?

consider the following scenario 
```

- apps
| - myapp
- packages
| - db

```

changes in @db currently are not picked by `--watch` nor by `--hot` flags

### What do you see instead?

a full restart (ctrl+c , up arrow , enter) 🤷

firtoz:
I have the same issue, I think it's not recognizing that the packages are part of the same workspace.

> `is not in the project directory and will not be watched`

firtoz:
I created test repo FYI https://github.com/Lunix-AI/BunTest/ @Electroid I hope it helps!

DobroslavR:
I have the same issue. Anyone got solution? Only happening on Windows

UltimateGG:
Same issue, even `bun --watch . ../Shared/types/*.ts ../ ../../` doesn't fix it

HarjjotSinghh:
Facing the same issue, any workarounds for this?

darklight9811:
Shouldn't bun be recognizing those as dependencies for the project instead of their absolute path?

kaminskypavel:
related? https://github.com/oven-sh/bun/issues/533#issuecomment-2514770195

eliahilse:
Ran into the same issue, you can circumvent the limitation by incurring the bun watch command in the root of the monorepo.

I.e. one can manually just run the command from the monorepo root:
> `bun --watch apps/your-app/src/index.ts` from the monorepo root

Or navigate to the monorepo root in the dev command, e.g. like that:
```json
"scripts": {
    "dev": "cd ../../ && bun --watch apps/your-app/src/index.ts",
}
```
(above would be in the package.json of the respective app, so that `turbo dev` works)

kaminskypavel:
> Ran into the same issue, you can circumvent the limitation by incurring the bun watch command in the root of the monorepo.
> 
> I.e. one can manually just run the command from the monorepo root:
> 
> > `bun --watch apps/your-app/src/index.ts` from the monorepo root
> 
> Or navigate to the monorepo root in the dev command, e.g. like that:
> 
> "scripts": {
>     "dev": "cd ../../ && bun --watch apps/your-app/src/index.ts",
> }
> (above would be in the package.json of the respective app, so that `turbo dev` works)

[...]

PR #16391, opened 2025-01-14
`bun exec --help` now prints its own help menu

### LINKED ISSUE #16359: `bun exec --help` prints help for `bun run`

1.1.43+76800b049

Linux 6.12.8-WSL2-LTS+ x86_64 x86_64

```console
$ bun exec --help
```

```
\\<b>Usage: bun exec <r><cyan>\<script\><r>
\\
\\Execute a shell script directly from Bun.
\\
\\<b><red>Note<r>: If executing this from a shell, make sure to escape the string!
\\
\\<b>Examples<d>:<r>
\\  <b>bun exec "echo hi"<r>
\\  <b>bun exec "echo \"hey friends\"!"<r>
\\
```

as written in https://github.com/oven-sh/bun/blob/36ad2974abed85ddff426e97a596b43a9717483a/src/cli.zig#L2703-L2717

### What do you see instead?

```console
$ bun exec --help
Usage: bun run [flags] <file or script>

Flags:
      --silent                    Don't print the script command
      --elide-lines               Number of lines of script output shown when using --filter (default: 10). Set to 0 to show all lines.
  -F, --filter                    Run a script in all workspace packages matching the pattern
  -b, --bun                       Force a script or package to use Bun's runtime instead of Node.js (via symlinking node)
      --shell                     Control the shell used for package.json scripts. Supports either 'bun' or 'system'
      --watch                     Automatically restart the process on file change
      --hot                       Enable auto reload in the Bun runtime, test runner, or bundler
      --no-clear-screen           Disable clearing the terminal screen on reload when --hot or --watch is enabled
      --smol                      Use less memory, but run garbage collection more often
  -r, --preload                   Import a module before other modules are loaded
      --inspect                   Activate Bun's debugger
      --inspect-wait              Activate Bun's debugger, wait for a connection before executing
      --inspect-brk               Activate Bun's debugger, set breakpoint on first line of code and wait
      --if-present                Exit without an error if the entrypoint does not exist
      --no-install                Disable auto install in the Bun runtime
      --install                   Configure auto-install behavior. One of "auto" (default, auto-installs when no node_modules), "fallback" (missing packages only), "force" (always).
  -i                              Auto-install dependencies during execution. Equivalent to --install=fallback.
  -e, --eval                      Evaluate argument as a script
      --print                     Evaluate argument as a script and print the result
      --prefer-offline            Skip staleness checks for packages in the Bun runtime and resolve from disk
      --prefer-latest             Use the latest matching versions of packages in the Bun runtime, always checking npm
  -p, --port                      Set the default port for Bun.serve
      --conditions                Pass custom conditions to resolve
      --fetch-preconnect          Preconnect to a URL while code is loading
      --max-http-header-size      Set the maximum size of HTTP headers in bytes. Default is 16KiB
      --expose-internals          Expose internals used for testing Bun itself. Usage of these APIs are completely unsupported.
      --expose-gc                 Expose gc() on the global object. Has no effect on Bun.gc().
      --no-deprecation            Suppress all reporting of the custom deprecation.
      --throw-deprecation         Determine whether or not deprecation warnings result in errors.
      --title                     Set the process title
      --main-fields               Main fields to lookup in package.json. Defaults to --target dependent
      --extension-order           Defaults to: .tsx,.ts,.jsx,.js,.json
      --tsconfig-override         Specify custom tsconfig.json. Default <d>$cwd<r>/tsconfig.json
  -d, --define                    Substitute K:V while parsing, e.g. --define process.env.NODE_ENV:"development". Values are parsed as JSON.
      --drop                      Remove function calls, e.g. --drop=console removes all console.* calls.
  -l, --loader                    Parse files with .ext:loader, e.g. --loader .js:jsx. Valid loaders: js, jsx, ts, tsx, json, toml, text, file, wasm, napi
      --no-macros                 Disable macros from being executed in the bundler, transpiler and runtime
      --jsx-factory               Changes the function called when compiling JSX elements using the classic JSX runtime
      --jsx-fragment              Changes the function called when compiling JSX fragments
      --jsx-import-source         Declares the module specifier to be used for importing the jsx and jsxs factory functions. Default: "react"
      --jsx-runtime               "automatic" (default) or "classic"
      --ignore-dce-annotations    Ignore tree-shaking annotations such as @__PURE__
      --env-file                  Load environment variables from the specified file(s)
      --cwd                       Absolute path to resolve files & entry points from. This just changes the process' cwd.
  -c, --config                    Specify path to Bun config file. Default <d>$cwd<r>/bunfig.toml
  -h, --help                      Display this menu and exit

Examples:
  Run a JavaScript or TypeScript file
  bun run ./index.js
  bun run ./index.tsx

  Run a package.json script
  bun run dev
  bun run lint

Full documentation is available at https://bun.sh/docs/cli/run
```

### Additional information

_No response_

RiskyMH:
### What does this PR do?
fixes #16359

- [ ] Code changes

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -2328,7 +2328,7 @@ pub const Command = struct {
                 try HelpCommand.exec(allocator);
             },
             .ExecCommand => {
-                const ctx = try Command.init(allocator, log, .RunCommand);
+                const ctx = try Command.init(allocator, log, .ExecCommand);
 
                 if (ctx.positionals.len > 1) {
                     try ExecCommand.exec(ctx);


robobun:
<div><sup>Updated 6:01 PM PT - Jan 13th, 2025</sup></div>

@RiskyMH, your commit c9a0dc66664179497b7d8ca1aca605629d8b2338 is building: [`#9547`](https://buildkite.com/bun/bun/builds/9547)

paperclover (approved):

Issue #9573, opened 2024-03-22
process.stdout.write outputs nothing, while console.log works


sean256:
### What version of Bun is running?

1.0.35

### What platform is your computer?

Linux 5.4.144-16.el7pie x86_64 unknown

### What steps can reproduce the bug?

test.ts
```ts
console.log('hello'); // works

process.stdout.write('hello'); // nothing
console._stdout.write('hello'); // nothing

throw new Error('hello'); // error outputs correctly
```

```sh
bun ./test.ts
```

### What is the expected behavior?

For `process.stdout.write` to output to console

### What do you see instead?

`stdout.write` outputs nothing

### Additional information

This is running in a proprietary CI environment and details on it are and will be very limited. All I know is the platform information I filled out above. This bug does not occur on my local machine and everything works in NodeJS when run on this platform.

paperclover:
i couldnt reproduce this on linux 6.1. i dont have linux 5.4 right now. did not observe on mac os.

can you check if an older version of bun like maybe ~1.0.20 has this issue? a lot of things were changed recently, but mostly for windows support.

sean256:
Oh snap, `1.0.20` does in fact work.

Jarred-Sumner:
macOS reproduces

sean256:
@paperdave
Is there a workaround for this other than downgrading? `1.0.20` breaks other stuff for us.
We are relying on Winston logger which outputs using `process.stdout.write` so it's no easy for us to switch to `console.log`

Jarred-Sumner:
if you run `bun upgrade --canary` does it by chance work?

sean256:
> Welcome to Bun's latest canary build!

No dice, same as `1.0.35`

PR #16395, opened 2025-01-14
Introduce experimental support for on-demand bundling via HTML imports and Bun.serve() 

Jarred-Sumner:
### What does this PR do?

This adds a zero-configuration frontend  production and frontend dev server to Bun.serve() behind the`--experimental-html` flag.

```ts
import hello from "./index.html";

Bun.serve({
  static: {
    "/": hello,
  },
  development: true,

  port: 3003,

  async fetch(req) {
    return new Response("Hello World: " + req.url);
  },
});
```

When navigating to `/`, it will bundle `./index.html`, reading the `<script>` tags, bundling any JavaScript, CSS, and assets. Then, those assets get served automatically in Bun.serve(). When `development: true` is set, Bun.serve() will rebuild the assets on the next reload of the HTML, log on build error and print the build status on completion. When not set, Bun will minify the assets and only build lazily on the first request.

This is an extremely rough,  simple version of this feature and it will get better soon. Keep in mind:
- no hot reloading / watching yet, still need to wire that up
- no `bun build` support yet, still need to wire that up. The plan there is it will do this build step statically, and return a manifest which can then also be used in Node or CF Workers or elsewhere


robobun:
<div><sup>Updated 5:29 AM PT - Jan 14th, 2025</sup></div>

@Jarred-Sumner, your commit f607a00d9446c49e5713737628fdbd0a92a0b7d1 is building: [`#9573`](https://buildkite.com/bun/bun/builds/9573)

packages/bun-uws/src/HttpRouter.h
```diff
@@ -35,43 +35,7 @@ namespace uWS {
 
 template <class USERDATA>
 struct HttpRouter {
-    /* These are public for now */
-    std::vector<std::string> upperCasedMethods = {
```
paperclover, line 39: this reminded me that bun currently doesnt support exotic method strings like `walawala`, instead defaulting them to get. but this diff is just uws updating so it ok.

paperclover (approved):
fwiw tests are not a blocker for merging as this is an experimental feature. focus on figuring out how these HTML bundles will be configurable since the `with { ... }` argument is kind of lame.

other things i think about:
- taking a directory tree of HTML files, since file system routing is optimal for generated pages
- a way to have parts or the whole html file generated by javascript / easily SSR an initial app shell or use HTML bundler as a SSG.
these two use cases should **not** be worked on yet, but just kept in mind as ways the HTML bundling api will expand over time.

src/bun.js/api/server/StaticRoute.zig
```diff
@@ -0,0 +1,256 @@
...
+    status = if (status == 200 and size == 0 and !this.blob.isDetached())
+        204
+    else
+        status;
+
+    this.doWriteStatus(status, resp);
+    this.doWriteHeaders(resp);
+}
+
+const StaticRoute = @This();
```
paperclover, line 263: nit: make this the first non-comment line. it makes it very obvious to a reader that the file is a struct for that name. not a big deal, one day i'll write a lint to force the ordering of import constants.

Issue #9579, opened 2024-03-23
Dedupe error messages in `logger.Log`


Jarred-Sumner:
This is a bad developer experience:

```
❯ bun install
bun install v1.0.35 (940448d6)
warn: incorrect peer dependency "lodash@4.17.20"

warn: incorrect peer dependency "lodash@4.17.20"
```

Why is it warning twice? 

We need to add a method to `logger.Log` that checks if the same message already existed and skips adding it again.

PR #16399, opened 2025-01-14
fix(codegen): handle generic internal functions

DonIsaac:
### What does this PR do?
`bundle-functions` now strips type parameters from internal function declarations. Before, type parameters caused codegen to crash.


robobun:
<div><sup>Updated 1:01 PM PT - Jan 14th, 2025</sup></div>

@DonIsaac, your commit 5e5c3b8588aa05e0d429e17dc777ca52c42fc2f4 is building: [`#9579`](https://buildkite.com/bun/bun/builds/9579)

paperclover (approved):

PR #16404, opened 2025-01-15
test(node/url): add more `node:test` parallel cases

DonIsaac:
### What does this PR do?
Ports remaining `node:url` parallel test cases.

#### Checklist
- [x] test-url-format-invalid-input
- [x] test-url-parse-invalid-input (_note: blocked by inconsistencies with reporting deprecation codes_)
- [x] test-url-parse-query
- [x] test-url-relative
- [x] ~test-url-is-url-internal~ tests node internals we can safely skip
- [ ] test-url-fileurltopath

These ones require PRs into WebKit and are not part of this PR:
- test-url-pathtofileurl

This pr, shockingly, contains new test cases


DonIsaac:
* **#16404** <a href="https://app.graphite.dev/github/pr/oven-sh/bun/16404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/oven-sh/bun/16404?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

DonIsaac:
Error code formatting fixes should appear in #16385

src/bun.js/bindings/bindings.cpp
```diff
@@ -6308,3 +6308,29 @@ extern "C" EncodedJSValue Bun__JSObject__getCodePropertyVMInquiry(JSC::JSGlobalO
...
+    JSC::StackVisitor::visit(callFrame, vm, [&](JSC::StackVisitor& visitor) -> WTF::IterationStatus {
+        if (Zig::isImplementationVisibilityPrivate(visitor) || visitor->isNativeCalleeFrame()) {
+            return WTF::IterationStatus::Continue;
+        }
+
+        if (visitor->hasLineAndColumnInfo()) {
+            String sourceURL = Zig::sourceURL(visitor);
+            if (sourceURL.startsWith("node:"_s) || sourceURL.startsWith("bun:"_s))
+                return WTF::IterationStatus::Continue;
+            if (sourceURL.startsWith("bun:"_s) || sourceURL.contains("node_modules"_s))
```
Jarred-Sumner: do we need to check for `sourceURL.startsWith("bun:"` twice?

src/js/builtins.d.ts
```diff
@@ -99,6 +99,14 @@ declare function $getMapIteratorInternalField(): TODO;
 declare function $getSetIteratorInternalField(): TODO;
 declare function $getProxyInternalField(): TODO;
 declare function $idWithProfile(): TODO;
+/**
+ * True for `JSCell`s. That is, this is roughly equivalent to this JS code:
```
Jarred-Sumner:
```suggestion
 * True for `JSCell` that are are object-like. That is, this is roughly equivalent to this JS code:
```

src/js/internal/util.ts
```diff
@@ -0,0 +1,5 @@
+const isInsideNodeModules: () => boolean = $newZigFunction("node_util_binding.zig", "isInsideNodeModules", 0);
```
Jarred-Sumner, line 1: ~instead of creating the function on every invocation, lets create it only once~

i misread the typescript type

test/js/bun/plugin/plugins.test.ts
```diff
@@ -476,7 +476,11 @@ describe("errors", () => {
         return new Response(result);
       },
     });
-    const { default: text } = await import(`http://${server.hostname}:${server.port}/hey.txt`);
+    const sleep = ms => new Promise<string>(res => setTimeout(() => res("timeout"), ms));
```
Jarred-Sumner, line 479: This forces the test to take a minimum of 2.5 seconds

src/bun.js/bindings/bindings.cpp
DonIsaac: That's a mistake, thanks for catching

src/js/builtins.d.ts
DonIsaac: note: I found that this was returning `true` for functions as well

test/js/bun/plugin/plugins.test.ts
DonIsaac: image

DonIsaac: If you prefer we can use something like this:
```ts
export function timeout(timeout: number): CancellablePromise {
  const { promise, resolve, reject } = Promise.withResolvers<void>();

  const handle = setTimeout(() => {
    reject(new TimeoutError(timeout));
  });

  const cancellablePromise = Object.assign(promise, {
    cancel() {
      clearTimeout(handle);
      resolve();
    },
  });

  cancellablePromise.cancel = cancellablePromise.cancel.bind(cancellablePromise);
  return cancellablePromise;
}
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3629,6 +3628,7 @@ void GlobalObject::addBuiltinGlobals(JSC::VM& vm)
 
     putDirectBuiltinFunction(vm, this, builtinNames.overridableRequirePrivateName(), moduleOverridableRequireCodeGenerator(vm), 0);
 
+    putDirectNativeFunction(vm, this, builtinNames.decodeURIComponentPrivateName(), 1, JSC::globalFuncDecodeURIComponent, ImplementationVisibility::Public, NoIntrinsic, PropertyAttribute::DontDelete | PropertyAttribute::ReadOnly);
```
Jarred-Sumner: this didn't need to be a global

could've exported it in the object bindings

DonIsaac: I've removed it

PR #16405, opened 2025-01-15
fix: update username

mchoo7:
### What does this PR do?

@paperdave is now @paperclover , so the comment needs to be updated accordingly.

- [ ] Code changes


paperclover (changes requested):
this is not the only comment mentioning my name. i have not had the time to go through things, since its not a hug deal.

though, for the TODO ones, don't alter them. i'd rather go through them when i have time and properly assess if they are still needed, or fix them if they are small.

fel1x-developer:
@paperclover . I replaced paperdave with paperclover except for TODOs using
```console
$ grep -rl paperdave * | xargs sed -i 's/paperdave/paperclover/g'
```
and `git diff`.

paperclover:
thank you

Issue #9617, opened 2024-03-25
Module '"react-hook-form"' has no exported member 'SubmitHandler'.


sh4den:
### What version of Bun is running?

1.0.35+940448d6b

### What platform is your computer?

Linux 6.1.0-13-amd64 x86_64 unknown

### What steps can reproduce the bug?

````sh
$ bun create vite my-vue-app # React, TypeScript + SWC
$ bun i react-hook-from
```

Then try to import useForm
```ts
import { useForm } from "react-hook-form";
```

### What is the expected behavior?

Should not return an error since it's the code that i copied from the react

### What do you see instead?

Module '"react-hook-form"' has no exported member 'useForm'.

### Additional information

Vite -> 5.2.6
React -> 18.2.69
react-dom -> 18.2.7

rsheppard-dev:
I am getting exact same error

pmkod:
Same error

struckchure:
+1

PR #16406, opened 2025-01-15
zig: else branch is implicitly void

nektro:
learned about this recently, still not decided on how i feel about whether its a net positive or negative to readability 
what do you think @paperclover @190n ?

### DIFF
--- a/src/bun.js/api/Timer.zig
+++ b/src/bun.js/api/Timer.zig
@@ -32,8 +32,7 @@ pub const All = struct {
         .context = {},
     },
     active_timer_count: i32 = 0,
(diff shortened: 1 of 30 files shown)


robobun:
<div><sup>Updated 8:56 PM PT - Jan 14th, 2025</sup></div>

@nektro, your commit c4b5d506d1f07c98e072bae627315a25ea694161 is building: [`#9617`](https://buildkite.com/bun/bun/builds/9617)

paperclover (approved):
i use this all the time. @DonIsaac should make this a check in `zlint`

190n (approved):

Issue #9659, opened 2024-03-27
Github Actions annotation does not render `cause` in Errors


Electroid:
```js
throw new Error("Oops!", { cause: new Error("Does not show in annotation") });
```

PR #16420, opened 2025-01-15
Disable IPInt again

190n:
### What does this PR do?

Disables IPInt, WebKit's in-place WASM interpreter, by default. This matches upstream behavior where it is disabled as of WebKit/WebKit#39029 due to bugs.

This code change will be unnecessary once either we update to a WebKit commit that disables it by default, or we update to a WebKit which has fixed and re-enabled IPInt.

We will see if it compiles


robobun:
<div><sup>Updated 12:03 PM PT - Jan 15th, 2025</sup></div>

@190n, your commit 5907c4485990dec2afc379dbde43c9e231bc41de is building: [`#9659`](https://buildkite.com/bun/bun/builds/9659)

PR #15688, opened 2024-12-10
Readline test fixes

heimskr:
### What does this PR do?

Adds new readline tests from Node.js and changes to Bun that cause the tests to pass.

- [ ] Documentation or TypeScript types
- [x] Code changes

### How did you verify your code works?

- [ ] I included a test for the new code, or existing tests cover it
- [x] I ran my tests locally and they pass (`bun-debug test test-file-name.test`)

PR #16425, opened 2025-01-16
Pass more node tests in util and readline

pfgithub:
- More readline tests: https://github.com/oven-sh/bun/pull/15688/files

TODO:

- [x] readline.node.test.ts


robobun:
<div><sup>Updated 5:51 PM PT - Jan 15th, 2025</sup></div>

@pfgithub, your commit f1a538800d2412fed6c3dbcb2a0db992351a99f8 is building: [`#9722`](https://buildkite.com/bun/bun/builds/9722)

src/bun.js/modules/NodeUtilTypesModule2.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include "BunClientData.h"
+
+JSC_DEFINE_HOST_FUNCTION(jsFunctionIsError,
```
Jarred-Sumner: Did you mean to commit this?

src/js/internal/util/inspect.js
```diff
@@ -963,10 +974,30 @@ function isInstanceof(object, proto) {
   }
 }
 
+// Special-case for some builtin prototypes in case their `constructor` property has been tampered.
+const wellKnownPrototypes = new SafeMap();
```
Jarred-Sumner: can we avoid doing this on startup?

src/bun.js/modules/NodeUtilTypesModule2.h
pfgithub: This is for $newCppFn() in util.ts - it can't be done on the existing NodeUtilTypesModule.h because '.h' extension is not allowed, and adding a NodeUtilTypesModule.cpp causes NodeUtilTypesModule.h to get compiled in twice. So this needs to go somewhere, I can put it in a different file though

src/bun.js/modules/NodeUtilTypesModule.h
```diff
@@ -144,6 +147,18 @@ JSC_DEFINE_HOST_FUNCTION(jsFunctionIsNativeError,
 
     return JSValue::encode(jsBoolean(false));
 }
+JSC_DEFINE_HOST_FUNCTION(jsFunctionIsNativeError,
+    (JSC::JSGlobalObject * globalObject,
+        JSC::CallFrame* callframe))
+{
+    GET_FIRST_VALUE
+    if (value.isCell()) {
+        if (value.inherits<JSC::ErrorInstance>() || value.asCell()->type() == ErrorInstanceType)
```
Jarred-Sumner: just the ErrorInstanceType should be it right?

PR #16426, opened 2025-01-16
Fix under-reporting string memory usage to GC

Jarred-Sumner:
### What does this PR do?

Fix under-reporting string memory usage to GC


robobun:
<div><sup>Updated 6:32 PM PT - Jan 15th, 2025</sup></div>

@Jarred-Sumner, your commit a7d58747c3d1ed69097a686d6500ae51dae22356 is building: [`#9726`](https://buildkite.com/bun/bun/builds/9726)

PR #16436, opened 2025-01-16
fixes #16431 HEAD requests returning 200 OK regardless of returned re…

### LINKED ISSUE #16431: HTTP Server HEAD request always responds 200 OK even if returning a different status code, ie: 404

1.1.43

Darwin 23.6.0 arm64 arm

Create a simple Bun.Serve file:

```ts
Bun.serve({
  fetch(req) {
    return new Response(null, {
      status: 404,
    });
  },
});
```

Send a HEAD request:
```bash
curl -vvv -X HEAD http://localhost:3000

* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:3000...
* Connected to localhost (::1) port 3000
> HEAD / HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< content-length: 0
< 
* Connection #0 to host localhost left intact
```

the Response should be `HTTP/1.1 404 Not Found`

### What do you see instead?

`HTTP/1.1 200 OK`

### Additional information

I've traced this to the following locations in the code, but im not really familiar enough with the project to know what the next step would be... totally willing to help though, lemme know!

The branch starts here:
https://github.com/oven-sh/bun/blob/8960a78e1f5fccd6466f0c4c22ed505431f1c72c/src/bun.js/api/server.zig#L3225-L3231

Which calls `writeHeaderInt` inside the `doRenderHeadResponse` function
https://github.com/oven-sh/bun/blob/8960a78e1f5fccd6466f0c4c22ed505431f1c72c/src/bun.js/api/server.zig#L3098-L3110

which jumps through a couple function calls but ultimately ends up here:
https://github.com/oven-sh/bun/blob/8960a78e1f5fccd6466f0c4c22ed505431f1c72c/packages/bun-uws/src/HttpResponse.h#L408-L417

Which we can see calls `writeStatus(HTTP_200_OK);`

Somehow we need to be able to write the status the was returned in our program above.  😊

zachlankton:
### What does this PR do?

Fixes Issue #16431 where HEAD requests would always return 200 OK Status Codes even when a Response was returned with a different status code such as 404.

Writing out Status codes and headers to the response in the server.zig file looks to be handled by the `renderMetaData` function in many cases, but for HEAD requests this function gets called after a header has already been written using a function called `writeHeaderInt`.

`writeHeaderInt` will (ultimately) send a status of 200 OK if no status has been sent yet.  This ensures that headers are NOT sent before status codes which would likely break a lot of things.

This PR adds a new flag named `head_request_status_written` so we can write the status before the call to `writeHeaderInt` and the subsequent call to write the status in `renderMetaData` -> `doWriteStatus` can be skipped.

- [x] Code changes

Added Automated Tests

### DIFF
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3117,7 +3117,6 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                 this.endWithoutBody(this.shouldCloseConnection());
                 return;
             };
-
             const globalThis = server.globalThis;
             var has_content_length_or_transfer_encoding = false;
             if (response.getFetchHeaders()) |headers| {
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3106,11 +3106,12 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             this.response_ptr = response;
             const server = this.server orelse {
                 // server detached?
-                resp.writeHeaderInt("content-length", 0);
                 this.renderMetadata();
+                resp.writeHeaderInt("content-length", 0);
                 this.endWithoutBody(this.shouldCloseConnection());
                 return;
             };
+            this.renderMetadata();
             const globalThis = server.globalThis;
             var has_content_length_or_transfer_encoding = false;
             if (response.getFetchHeaders()) |headers| {
@@ -3173,7 +3174,6 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
                 }
             }
 
-            this.renderMetadata();
             this.endWithoutBody(this.shouldCloseConnection());
         }
 
--- a/src/bun.js/api/server.zig
+++ b/src/bun.js/api/server.zig
@@ -3106,19 +3106,17 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             }
             const resp = this.resp.?;
             this.response_ptr = response;
+
+            this.doWriteStatus(response.statusCode());
+            this.flags.head_request_status_written = true;
+
             const server = this.server orelse {
                 // server detached?
-                this.doWriteStatus(response.statusCode());
-                this.flags.head_request_status_written = true;
-
                 resp.writeHeaderInt("content-length", 0);
                 this.renderMetadata();
                 this.endWithoutBody(this.shouldCloseConnection());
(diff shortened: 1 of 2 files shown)


zachlankton:
existing tests are passing and new tests have been added to cover this fix

src/bun.js/api/server.zig
```diff
@@ -3904,6 +3909,8 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
         }
 
         fn doWriteStatus(this: *RequestContext, status: u16) void {
+            if (this.flags.head_request_status_written) return;
```
cirospaciari, line 3912: this condition makes the assertion bellow unreachable

src/bun.js/api/server.zig
```diff
@@ -3105,6 +3106,10 @@ fn NewRequestContext(comptime ssl_enabled: bool, comptime debug_mode: bool, comp
             }
             const resp = this.resp.?;
             this.response_ptr = response;
+
+            this.doWriteStatus(response.statusCode());
```
cirospaciari, line 3110: we should just call renderMetadata before `resp.writeHeaderInt` / `resp.resp.writeHeader` to centralize the status logic, no need to manually mark ` head_request_status_written`

cirospaciari (changes requested):
Thank you for the PR and for opening the Issue, added some comments, on the case of s3 stats we should not write the status until we resolve the stat and get corked again otherwise we will send 2 packages over the network instead of 1, we also need more tests covering other conditions

src/bun.js/api/server.zig
zachlankton: ok, sweet, I see you've already done all this work, Im more than ok with closing this PR in favor of yours if you'd like

cirospaciari: Yeah I did not saw that you have finished this PR, if you wanna feel free to copy the stuff to your PR and I will approve this in favor of mine since you started first

zachlankton: lol you are too kind, I'm happy to have been a part of the bug report and triage!  I'm gonna close this one, lets get yours merged in!

zachlankton:
closing in favor of: https://github.com/oven-sh/bun/pull/16445

zachlankton:
https://github.com/oven-sh/bun/pull/16445

Issue #9794, opened 2024-04-01
Cannot find module "@panva/hkdf"


0-don:
### What version of Bun is running?

1.1.0+5903a6141

### What platform is your computer?

Linux 6.8.2-arch2-1 x86_64

### What steps can reproduce the bug?

I am using solid-start nextauth and running the bun app in docker container which produces this error

```bash
at processTicksAndRejections (:12:39)
[nitro] [request error] [unhandled] Cannot find module "@panva/hkdf" from "/app/.output/server/node_modules/@auth/core/jwt.js"
  at new H3Error (./.output/server/index.mjs:2086:4)  
  at createError$1 (./.output/server/index.mjs:4043:26)  
  at <anonymous> (./.output/server/index.mjs:4975:34)  
  at processTicksAndRejections (:12:39)
  ```

### What is the expected behavior?

it works in the previous bun version 1.0.36

### What do you see instead?

it works in the previous bun version 1.0.36

shtse8:
I have the same issue. It works on 1.0.36.

```bash
2024-04-14 14:06:25 error: Cannot find module "@panva/hkdf" from "/usr/src/app/.output/server/node_modules/@auth/core/jwt.js"
2024-04-14 14:06:25 error: script "prod" exited with code 1
2024-04-14 14:06:27 $ bun x prisma migrate deploy && bun .output/server/index.mjs
2024-04-14 14:06:27 error: Cannot find module "@panva/hkdf" from "/usr/src/app/.output/server/node_modules/@auth/core/jwt.js"
2024-04-14 14:06:27 error: script "prod" exited with code 1
2024-04-14 14:06:28 $ bun x prisma migrate deploy && bun .output/server/index.mjs
2024-04-14 14:06:29 error: Cannot find module "@panva/hkdf" from "/usr/src/app/.output/server/node_modules/@auth/core/jwt.js"
2024-04-14 14:06:29 error: script "prod" exited with code 1
2024-04-14 14:06:30 $ bun x prisma migrate deploy && bun .output/server/index.mjs
2024-04-14 14:06:31 error: Cannot find module "@panva/hkdf" from "/usr/src/app/.output/server/node_modules/@auth/core/jwt.js"
2024-04-14 14:06:31 error: script "prod" exited with code 1
2024-04-14 14:06:32 $ bun x prisma migrate deploy && bun .output/server/index.mjs
[...]

PR #16446, opened 2025-01-16
node:dgram compatibility improvements

heimskr:
### What does this PR do?

Improves the dgram module to add features and fix bugs. Adds some tests from Node.js.

- [x] Code changes

Added tests.


robobun:
<div><sup>Updated 10:49 AM PT - Jan 16th, 2025</sup></div>

@heimskr, your commit efae2d42dfc6aa7d78c3d51ea715791bad455d72 is building: [`#9794`](https://buildkite.com/bun/bun/builds/9794)

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -90,11 +93,28 @@ fn onData(socket: *uws.udp.Socket, buf: *uws.udp.PacketBuffer, packets: c_int) c
...
+            break :blk bun.String.createFormat("{s}%{d}", .{ span, id }) catch bun.outOfMemory();
+        } else bun.String.init(span);
+        defer hostname_string.deref();
+
         _ = callback.call(globalThis, udpSocket.thisValue, &.{
             udpSocket.thisValue,
             udpSocket.config.binary_type.toJS(slice, globalThis),
             JSC.jsNumber(port),
-            JSC.ZigString.init(std.mem.span(hostname.?)).toJS(globalThis),
+            hostname_string.toJS(globalThis),
```
Jarred-Sumner:
```suggestion
            hostname_string.transferToJS(globalThis),
```

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -153,9 +174,21 @@ pub const UDPSocketConfig = struct {
             }
         };
 
+        const flags: i32 = brk: {
+            if (try options.getTruthy(globalThis, "flags")) |value| {
+                if (!value.isInt32()) {
```
Jarred-Sumner, line 179:
```suggestion
                if (!value.isAnyInt()) {
```

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -153,9 +174,21 @@ pub const UDPSocketConfig = struct {
             }
         };
 
+        const flags: i32 = brk: {
+            if (try options.getTruthy(globalThis, "flags")) |value| {
```
Jarred-Sumner: This should use validateInt from the validators so that it returns the correct error code for each scenario and so that it doesn't break if you do `13.0` instead of `13`
```suggestion
            if (try options.getTruthy(globalThis, "flags")) |value| {
              
```

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -290,19 +323,36 @@ pub const UDPSocket = struct {
...
+                const code = @tagName(bun.C.SystemErrno.init(@as(c_int, @intCast(err))).?);
+                const sys_err = JSC.SystemError{
+                    .errno = err,
+                    .code = bun.String.static(code),
+                    .message = bun.String.createFormat("bind {s} {s}", .{ code, config.hostname }) catch bun.outOfMemory(),
+                };
+                const address = bun.String.createUTF8(config.hostname);
+                defer address.deref();
+                const error_value = sys_err.toErrorInstance(globalThis);
+                error_value.put(globalThis, "address", address.toJS(globalThis));
```
Jarred-Sumner:
```suggestion
                error_value.put(globalThis, "address", bun.String.createUTF8ForJS(globalThis, config.hostname));
```

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -353,6 +403,97 @@ pub const UDPSocket = struct {
...
+        const res = this.socket.setBroadcast(enabled);
+
+        if (getUSError(res, .setsockopt, true)) |err| {
+            return globalThis.throwValue(err.toJS(globalThis));
+        }
+
+        return arguments[0];
+    }
+
+    pub fn setMulticastLoopback(this: *This, globalThis: *JSGlobalObject, callframe: *CallFrame) bun.JSError!JSValue {
```
Jarred-Sumner, line 417: Do these need to be functions or can we make these getters/setters

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -737,6 +737,16 @@ JSC_DEFINE_HOST_FUNCTION(jsFunction_ERR_UNHANDLED_ERROR, (JSC::JSGlobalObject *
     return JSC::JSValue::encode(createError(globalObject, ErrorCode::ERR_UNHANDLED_ERROR, makeString("Unhandled error. ("_s, err_str, ")"_s)));
 }
 
+JSC_DEFINE_HOST_FUNCTION(jsFunction_ERR_INVALID_IP_ADDRESS, (JSC::JSGlobalObject * globalObject, JSC::CallFrame* callFrame))
```
Jarred-Sumner: instead of adding another one of these one-off functions can this be in the createError generic function below? this can be used via

```
throw $ERR_INVALID_IP_ADDRESS(...args)
```

src/cli.zig
```diff
@@ -770,6 +770,10 @@ pub const Arguments = struct {
                 ctx.runtime_options.dns_result_order = order;
             }
 
+            if (args.option("--dns-result-order")) |order| {
```
Jarred-Sumner: is that from this branch?

src/bun.js/api/bun/udp_socket.zig
heimskr: They're functions in node:dgram, but maybe we could make those functions wrappers around getters/setters on the object returned by Bun.udpSockets

src/cli.zig
heimskr: Yeah I merged the DNS branch into this one

heimskr: But it seems I was careless with the merge

src/bun.js/bindings/ErrorCode.cpp
heimskr: It's already handled in jsFunctionMakeErrorWithCode, must be a result of careless merging

src/bun.js/api/bun/udp_socket.zig
heimskr: Doing this means I'd have to do some coercion in JS instead, because there are tests that are depending on the coercion.

heimskr: Actually, that's just for the port number. Should be fine with flags.

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -90,11 +93,28 @@ fn onData(socket: *uws.udp.Socket, buf: *uws.udp.PacketBuffer, packets: c_int) c
...
+
+                var buffer = std.mem.zeroes([net_if.IF_NAMESIZE:0]u8);
+                if (net_if.if_indextoname(id, &buffer) != null) {
+                    break :blk bun.String.createFormat("{s}%{s}", .{ span, std.mem.span(@as([*:0]u8, &buffer)) }) catch bun.outOfMemory();
+                }
+            }
+
+            break :blk bun.String.createFormat("{s}%{d}", .{ span, id }) catch bun.outOfMemory();
+        } else bun.String.init(span);
+        defer hostname_string.deref();
```
Jarred-Sumner:
```suggestion

```

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -290,19 +317,36 @@ pub const UDPSocket = struct {
...
-            this.deinit();
+            defer this.deinit();
+            if (err != 0) {
+                const code = @tagName(bun.C.SystemErrno.init(@as(c_int, @intCast(err))).?);
+                const sys_err = JSC.SystemError{
+                    .errno = err,
+                    .code = bun.String.static(code),
+                    .message = bun.String.createFormat("bind {s} {s}", .{ code, config.hostname }) catch bun.outOfMemory(),
+                };
+                const address = bun.String.createUTF8(config.hostname);
```
Jarred-Sumner: where is address used?

src/bun.js/api/bun/udp_socket.zig
```diff
@@ -510,6 +670,35 @@ pub const UDPSocket = struct {
...
+                                if (std.math.cast(u32, index)) |id| {
+                                    break :blk id;
+                                }
+                            }
+                        }
+                        break :blk null;
+                    };
+
+                    if (iface_id) |id| {
+                        address_slice[percent] = '\x00';
```
Jarred-Sumner: is there always room in the buffer?

src/bun.js/api/bun/udp_socket.zig
heimskr: Oops, nowhere

src/bun.js/api/bun/udp_socket.zig
heimskr: Yes, because `percent` is always before the end of the buffer

heimskr: This is just to trick inet_pton into thinking the string is shorter than it actually is (specifically, that it doesn't contain `%foo` at the end).

src/bun.js/bindings/bindings.cpp
```diff
@@ -4130,6 +4130,14 @@ JSC__JSString* JSC__JSValue__toStringOrNull(JSC__JSValue JSValue0, JSC__JSGlobal
     return value.toStringOrNull(arg1);
 }
 
+bool JSC__JSValue__toNumber(JSC::EncodedJSValue value, JSC::JSGlobalObject* global, double* result)
```
Jarred-Sumner: Was this necessary?

src/js/node/dgram.ts
```diff
@@ -95,9 +114,26 @@ function newHandle(type, lookup) {
...
+
   return handle;
 }
 
+function onMessage(nread, handle, buf, rinfo) {
+  const self = handle[kOwnerSymbol];
+  if (nread < 0) {
+    return self.emit(
+      "error",
+      Object.assign(new Error("recvmsg"), {
```
Jarred-Sumner, line 127: This seems not great but I don't want to block merging on it

src/js/node/dgram.ts
```diff
@@ -690,127 +753,133 @@ Socket.prototype.remoteAddress = function () {
...
 
 Socket.prototype.setBroadcast = function (arg) {
-  throwNotImplemented("setBroadcast", 10381);
-  /*
-  const err = this[kStateSymbol].handle.setBroadcast(arg ? 1 : 0);
-  if (err) {
-    throw new ErrnoException(err, 'setBroadcast');
+  const handle = this[kStateSymbol].handle;
+  if (!handle?.socket) {
+    throw new Error("setBroadcast EBADF");
```
Jarred-Sumner, line 758: We should be using a using errors with codes instead of errors with messages. It sounds like we need a helper for throwing a system error with a syscall name and an errno.

src/js/node/dgram.ts
```diff
@@ -690,127 +753,133 @@ Socket.prototype.remoteAddress = function () {
...
-    throw new ErrnoException(err, 'setMulticastTTL');
+  if (typeof ttl !== "number") {
+    throw $ERR_INVALID_ARG_TYPE("ttl", "number", ttl);
   }
 
-  return ttl;
-  */
+  const handle = this[kStateSymbol].handle;
+  if (!handle?.socket) {
+    throw new Error("setMulticastTTL EBADF");
```
Jarred-Sumner, line 782: Here as well

src/js/node/dgram.ts
```diff
@@ -690,127 +753,133 @@ Socket.prototype.remoteAddress = function () {
...
-    throw new ErrnoException(err, 'setTTL');
+  if (typeof ttl !== "number") {
+    throw $ERR_INVALID_ARG_TYPE("ttl", "number", ttl);
   }
 
-  return ttl;
-  */
+  const handle = this[kStateSymbol].handle;
+  if (!handle?.socket) {
+    throw new Error("setTTL EBADF");
```
Jarred-Sumner, line 770: And here

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -813,26 +813,25 @@ static inline JSC::EncodedJSValue jsBufferConstructorFunction_concatBody(JSC::JS
...
 
-        if (length > 0)
-            args.append(element);
+            if (length > 0)
+                args.append(element);
 
-        byteLength += length;
+            byteLength += length;
+        } else {
+            throwTypeError(lexicalGlobalObject, throwScope, "Buffer.concat expects Buffer or Uint8Array"_s);
```
Jarred-Sumner, line 832: This throwTypeError -- is that throwing the error with the code property that is expected or is it not doing that?

src/bun.js/bindings/bindings.cpp
heimskr: No, removed

PR #16459, opened 2025-01-17
fix: use const

mchoo7:
### What does this PR do?

Detected by vscode-zig.

- [x] Code changes


fel1x-developer:
@Jarred-Sumner can we merge this?

PR #16480, opened 2025-01-17
Strip query parameter in File import

### LINKED ISSUE #16476: Regression: release 1.1.45 broke `?url` imports

1.1.45

Ubuntu 22.04.5 LTS 

```
git clone https://github.com/tmr232/function-graph-overview/
cd function-graph-overview
bun install
bun test
```

I expect all the tests to run successfully, as they do up to and including bun-v1.1.44:

(this is the end of the output. The rest removed for brevity.)
```
 428 pass
 0 fail
 238 snapshots, 257 expect() calls
Ran 428 tests across 4 files. [2.24s]
```

### What do you see instead?

The tests fail with the following error:

```
bun test v1.1.45 (196621f2)

src\test\backlinks.test.ts:
✓ detect-backlinks-0
✓ detect-backlinks-1
✓ detect-backlinks-2
✓ detect-backlinks-3

src\test\commentTest.test.ts:

# Unhandled error between tests
-------------------------------
2949 |               bytes = Promise.resolve(input);
2950 |             } else {
2951 |               const url = input;
2952 |               if (typeof process !== "undefined" && process.versions && process.versions.node) {
2953 |                 const fs = require("fs");
2954 |                 bytes = Promise.resolve(fs.readFileSync(url));
                                                  ^
ENOENT: no such file or directory, open 'C:\Code\sandbox\function-graph-overview\parsers\tree-sitter-c.wasm?url'
    path: "C:\\Code\\sandbox\\function-graph-overview\\parsers\\tree-sitter-c.wasm?url",
 syscall: "open",
   errno: -2,
    code: "ENOENT"

      at load (C:\Code\sandbox\function-graph-overview\node_modules\web-tree-sitter\tree-sitter.js:2954:44)
      at <anonymous> (C:\Code\sandbox\function-graph-overview\src\parser-loader\bun.ts:10:48)
-------------------------------


src\test\ranges_class.test.ts:
✓ Basic Range Test
✓ Add to 0
✓ Add to same start
✓ Invalid range addition

src\test\zip.test.ts:
✓ zip
✓ structZip
✓ pairwise

 11 pass
 1 fail
 1 error
 19 expect() calls
Ran 12 tests across 4 files. [153.00ms]
```

The relevant import is defined in https://github.com/tmr232/function-graph-overview/blob/main/src/parser-loader/wasmMapping.ts

### Additional information

The `?url` queries are added because the same file is currently used both when running directly in Bun, and also when building a website using Vite.

If I remove the `?url` query, Bun-v1.1.45 works but Vite does not, so it's not a good solution in my case.

pfgithub:
### What does this PR do?

Fixes #16476

Something I noticed: query parameters don't work at all in `bun build`

```ts
import text from "./b.txt?1";
import file from "./c.file?2";

console.log(text, file);
```

```ts
$> bun build --target=browser a.ts
1 | import text from "./b.txt?1";
                     ^
error: Could not resolve: "./b.txt?1"
    at a.ts:1:18

2 | import file from "./c.file?2";
                     ^
error: Could not resolve: "./c.file?2"
    at a.ts:2:18
Exited with code [1]
```

They should probably work and with `--target=browser` maybe even preserve the query parameter (but not for bun/node)


robobun:
<div><sup>Updated 1:19 PM PT - Jan 17th, 2025</sup></div>

@pfgithub, your commit 9b3fa336d3a0acf0efe233266a1d38259c5a37ec is building: [`#9941`](https://buildkite.com/bun/bun/builds/9941)

Jarred-Sumner (approved):

Jarred-Sumner:
oh this is silly

we should just create the JSString

PR #16462, opened 2025-01-17
meta: migrate bun to the text lockfile

nektro:
ran `bun install --save-text-lockfile --frozen-lockfile --lockfile-only` in each folder

PR #16487, opened 2025-01-18
fix: add `bun.lock` to `.gitattributes`, `.gitignore`, and docs

mchoo7:
Fixes #16462. Supersedes #16324.

### What does this PR do?

- [ ] Code changes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,7 +19,8 @@
 *.mjs text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 *.mts text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 
-*.lockb binary diff=lockb
+bun.lock linguist-language=json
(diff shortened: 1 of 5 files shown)


robobun:
Thank you for the PR. I am closing it because the changes no longer apply to main.

- #16462 was merged in January 2025. The repo has used the text `bun.lock` since then, and it commits its `bun.lock` files (51 of them are tracked on main today).
- `.gitattributes` already has a `*.lock text eol=lf` rule, which covers `bun.lock`. The `*.lockb binary diff=lockb` rule still needs the glob, because `test/cli/install/fixtures/invalid-optional-peer.lockb` is not named `bun.lockb`.
- `bench/install/bun.lock` is a tracked file now, so `bench/install/.gitignore` must not ignore it.
- `misctools/.gitignore` no longer exists.
- The tests copy the `complex-workspace` and `contoso-test` fixtures to a temporary directory before they run, so no `bun.lock` is written into the fixture directories.

The only part of this diff that main does not have is the `linguist-language` hint for `bun.lock`. If you still see a gap there, or anywhere else, please open a fresh PR against current main.

PR #16490, opened 2025-01-18
docs: improve command line environment variable tips

dy0gu:
### What does this PR do?

- [X] Documentation

Improves the existing docs section for setting variables through the command line. 

The method currently described in the docs is the following:

```sh
$ FOO=helloworld bun run dev
```

The problem is this syntax is only available on POSIX-compliant shells, which mostly means MacOS and Linux. I assumed that bit was a leftover from when Bun did not support Windows and had yet to be updated.

I've recently seen a significant number of Windows users being confused as to why following the docs on this was not working, so I decided to clarify where that syntax is available.

In addition to that I added a cross-platform alternative for setting variables like previously expected, although I'm open to feedback if you think it's not _clean_ enough to be in the official docs. It makes use of a somewhat chained and lengthy `bun exec` syntax.

I believe the presence of the examples added by this PR are important as I found myself and other peers regularly needing specific `package.json` commands to set the `NODE_ENV` variable, for example.


docs/runtime/env.md
```diff
@@ -13,12 +13,26 @@ FOO=hello
...
+```
+
+You can even use it directly on a `package.json` script:
+
+```json
+"scripts": {
+  "dev": "bun exec 'NODE_ENV=development bun --watch  app.ts'",
+},
+```
+
```
RiskyMH: fwiw with bun on windows, it executes package.json scripts with bun shell and as a result you dont need the `bun exec` unless you are expecting to use it with other package managers.

docs/runtime/env.md
```diff
@@ -13,12 +13,26 @@ FOO=hello
 BAR=world
 ```
 
-Variables can also be set via the command line.
+Variables can be set before execution via the command line, on POSIX-compliant shells.
 
 ```sh
 $ FOO=helloworld bun run dev
 ```
```
RiskyMH: i was thinking of this some time ago but wasnt sure how many people were effected.

personally id think something like this doc page (https://bun.sh/guides/ecosystem/edgedb) where it shows how to do it with macos/linux or windows (where `set FOO=helloworld && bun run dev` can set the env)

> <sub>only issue with my example is i believe it changes session like `export` in linux</sub>

docs/runtime/env.md
dy0gu: Oh... you're absolutely right. I just checked and we had been running the `dev` script with NPM this whole time, that's why we were having to use `bun exec`. I'll update the PR with that in mind. And thanks for just clearing up a lot of ugly syntax on some projects here!

docs/runtime/env.md
dy0gu: I think your idea sounds good. Just a question before I get started on the changes, can I use the `{% codetabs %}` syntax for the OS toggle on these documentation pages like it's done in the guides or do they function differently from the `/guides` pages?

RiskyMH: yeah its available on guides, and the example guide I linked https://github.com/oven-sh/bun/blob/main/docs/guides/ecosystem/edgedb.md?plain=1

RiskyMH: image
also it does seem to update for the session which isn't a big deal, just isn't 1:1 with linux and I haven't found an easier method other then `bun exec ''` 

but it doesn't change global which is what really matters.

dy0gu: Nice! Also, I noticed that `set` is Windows Terminal / CMD only, do you think it's worth it to also document the usage with Powershell? Personally I use it way more.
 
 ```sh
 $env:FOO="helloworld"; bun run dev
 ```
 
 This one also sticks with the session, but doesn't change global either.

RiskyMH: ahhh why does powershell have to be so different to cmd... uh to be honest i don't really know because people use either or _(im a cmd fan)_, so maybe have both with comment on top, or just use bun exec and comment saying windows makes env hard 😆

dy0gu: Yeah it also drives me nuts when I'm looking up Windows commands that only work on one or the other.

I made the changes, let me know what you think!

RiskyMH:
image

i did some fixes to fix it rendering on the site, also this is how it looks rn ^

Personally i think it could possibly be more concise as its just the top of docs and maybe in different section/sub-section it could be explored further. Also btw its actually because of [bun shell](https://bun.sh/docs/runtime/shell) that it works on windows (which `bun exec` uses), and i say windows as by default only windows gets this feature.

dy0gu:
I update the mention of `bun exec` on Windows to bun shell instead, for clarity. 

I also shortened the longer bit of text, I think all the needed info is still nicely there.

Side note, it's nice to finally give back to Bun, the experience has been great and the built-in environment utilities are probably one of the biggest reasons for having migrated personally. Also, thanks for your review @RiskyMH, let me know if any more changes are needed!

RiskyMH:
nice glad to hear!

to make it more clear i put it in a dropdown which i believe will catch your eye better also.. let me know if you preferred it the other way though.

| normal | expanded |
| --- | --- |
| image | image |

dy0gu:
Looks good to me 👍

RiskyMH (approved):

PR #16508, opened 2025-01-18
Fix a typo in docs

zigazajc007:
### What does this PR do?

- [ ] Code changes


Jarred-Sumner:
looks like a legit typo but its missing the name

zigazajc007:
Was confused what file even is here. But then delete function without path parameter wouldn't work.

PR #16512, opened 2025-01-19
more(sql) type fixes and tests

cirospaciari:
### What does this PR do?

- [x] Code changes

Tests


robobun:
<div><sup>Updated 4:21 PM PT - Jan 18th, 2025</sup></div>

@cirospaciari, your commit 68a0212df7af462899ae4c782562a0478ae899b6 is building: [`#10071`](https://buildkite.com/bun/bun/builds/10071)

src/bun.js/bindings/SQLClient.cpp
```diff
@@ -102,9 +115,24 @@ static JSC::JSValue toJS(JSC::VM& vm, JSC::JSGlobalObject* globalObject, DataCel
...
+        if (cell.value.raw.length > 0) {
+            memcpy(uint8Array->vector(), reinterpret_cast<void*>(cell.value.raw.ptr), cell.value.raw.length);
+        }
+        return uint8Array;
+    }
     case DataCellTag::String: {
-        return jsString(vm, WTF::String(cell.value.string));
-        break;
+        if (cell.value.string) {
+            return jsString(vm, WTF::String(cell.value.string));
```
Jarred-Sumner, line 133: Should the WTF::StringImpl in the cell be set to nullptr after this?
```suggestion
            return jsString(vm, WTF::String(cell.value.string));
```

src/bun.js/bindings/SQLClient.cpp
```diff
@@ -250,79 +280,103 @@ static JSC::JSValue toJS(JSC::VM& vm, JSC::JSGlobalObject* globalObject, DataCel
...
+            auto& cell = cells[i];
+            JSValue value = toJS(vm, globalObject, cell);
+            RETURN_IF_EXCEPTION(scope, {});
+            array->putDirectIndex(globalObject, i, value);
+        }
+        return array;
+    }
+
+    default:
+        // not a valid result mode
```
Jarred-Sumner, line 373:
```suggestion
        // not a valid result mode
        ASSERT_NOT_REACHED();
```

src/js/bun/sql.ts
```diff
@@ -22,23 +23,31 @@ function connectionClosedError() {
 }
 hideFromStack(connectionClosedError);
 
+enum SQLQueryResultMode {
+  objects = 0,
+  values = 1,
+  raw = 2,
+}
+const escapeIdentifier = function escape(str) {
+  return '"' + str.replace(/"/g, '""').replace(/\./g, '"."') + '"';
```
Jarred-Sumner: Can we use `replaceAll` instead of `replace` with a regex?

src/js/bun/sql.ts
```diff
@@ -468,42 +520,29 @@ class ConnectionPool {
...
-        if (pendingReserved) {
-          connection.flags |= PooledConnectionFlags.reserved;
-          connection.queryCount++;
-          // we have a connection waiting for a reserved connection lets prioritize it
-          pendingReserved(connection.storedError, connection);
-          return;
+
+        // we have no connections available so lets fails
+        let pending;
+        while ((pending = this.waitingQueue.shift())) {
```
Jarred-Sumner: Is it intentional that this is .shift instead of a for loop? Should we swap the arrays and then for loop?

Jarred-Sumner:
lets merge once tests run

src/bun.js/bindings/SQLClient.cpp
cirospaciari: we need to deref it later so we still keepit

Issue #10120, opened 2024-04-09
Playwright does not work on Bunjs environment


sohaieb:
### What version of Bun is running?

1.1.3

### What platform is your computer?

Microsoft Windows NT 10.0.19045.0 x64"Microsoft Windows NT 10.0.19045.0 x64

### What steps can reproduce the bug?

Hi, 
Actually I installed [Playwright](https://playwright.dev/docs/intro) following its documentation, its dependencies and its browsers (using `npx playwright install`).

then I wrote this simple code: 

```
import { chromium } from "playwright";

async function start() {
    const browser = await chromium.launch({
        headless: false
    })
    const page = await browser.newPage();
    await page.goto('https://google.com')
}


start();
```
runing the same code works well on pure **NodeJs environment** (without TypeScript), but running it using **BunJs** does not work.

### What is the expected behavior?

It should display a browser window with a default tab showing https://google.com page

### What do you see instead?

running the code above with `--verbose` flag shows this bunch of errors:
```
1 | import { chromium } from "playwright";
2 | 
                                         ^
TimeoutError: launch: launch: Timeout 180000ms exceeded.
Call log:
  - <launching> C:\Users\[user]\AppData\Local\ms-playwright\chromium-1112\chrome-win\chrome.exe --disable-field-trial-config --disable-background-networking --enable-features=NetworkSe
rvice,NetworkServiceInProcess --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-back-forward-cache --disable-breakpad --disable-client-side-phishi
ng-detection --disable-component-extensions-with-background-pages --disable-component-update --no-default-browser-check --disable-default-apps --disable-dev-shm-usage --disable-extensio
ns --disable-features=ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose,MediaRouter,DialMediaRouteProvider,AcceptCHFrame,AutoExpandDetailsElement,
CertificateTransparencyComponentUpdater,AvoidUnnecessaryBeforeUnloadCheckSync,Translate,HttpsUpgrades,PaintHolding --allow-pre-commit-input --disable-hang-monitor --disable-ipc-flooding
-protection --disable-popup-blocking --disable-prompt-on-repost --disable-renderer-backgrounding --force-color-profile=srgb --metrics-recording-only --no-first-run --enable-automation -
[...]

PR #16540, opened 2025-01-20
Make `bun.lock` the default for new projects

dylan-conway:
### What does this PR do?
Makes `bun.lock` the default lockfile format for new lockfiles. Existing binary lockfiles will continue to be used.

This also changes the format of the lockfile slightly. Workspace resolutions in the `"packages"` object will no longer list their dependencies. The lockfile version is changed from 0 to 1.

Added tests for:
- using binary lockfiles if they already exist.
- creating a binary lockfile if `saveTextLockfile=false` in bunfig.toml.
- `bun.lock` is created when no lockfile exists.
- `bun.lock` is created when migrating from `package-lock.json`.


robobun:
<div><sup>Updated 7:02 PM PT - Jan 19th, 2025</sup></div>

@dylan-conway, your commit 4e96c066ec413b3aa939ac03dbc40df5b83a98c2 is building: [`#10120`](https://buildkite.com/bun/bun/builds/10120)

Issue #10184, opened 2024-04-11
Getting the following error once I'm setting up a @company scope for JetBrains Space registry (error: FailedToOpenSocket downloading package manifest)


zozobalogh0817:
### What version of Bun is running?

1.1.3+2615dc742

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Create a package on JetBrains Space Registry and then define that in bunfig.toml as the following

[install.scopes]
"@company" = "https://username:password@https://npm.pkg.jetbrains.space/company/p/PROJECT_ID_ON_JB_SPACE/registry/"

### What is the expected behavior?

Download the package from the corresponding registry

### What do you see instead?

I'm getting the following error.
error: FailedToOpenSocket downloading package manifest

zozobalogh0817:
I'm not sure what happened, I think it was a missconfiguration from my side, so I'm closing the ticket

This is the working configuration for me in `bunfig.toml`

```toml
[install.scopes]
company = "https://username:password@npm.pkg.jetbrains.space/company/p/PROJECT_ID_AT_SPACE/registry/"
```

zozobalogh0817:
Thanks and happy engineering 🎉📣

PR #16555, opened 2025-01-21
packages/bun-types: remove bun.lockb

nektro:
`packages/bun-types` is a workspace in the root `package.json` is this is unnecessary.
is not regenerated when running `bun install` in the folder so this removal is idempotent


robobun:
<div><sup>Updated 4:45 PM PT - Jan 20th, 2025</sup></div>

@nektro, your commit 711e67a8407f49175e45276218cda5dd5ad2073e is building: [`#10184`](https://buildkite.com/bun/bun/builds/10184)

PR #16556, opened 2025-01-21
bun-types: depend on @types/node@*

nektro:
Closes https://github.com/oven-sh/bun/pull/16236
Closes https://github.com/oven-sh/bun/pull/16237

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
     "autoprefixer": "^10.4.19",
     "@mdn/browser-compat-data": "~5.5.28"
   },
+  "resolutions": {
+    "bun-types": "workspace:packages/bun-types"
+  },
   "scripts": {
     "build": "bun run build:debug",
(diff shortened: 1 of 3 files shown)


robobun:
<div><sup>Updated 5:25 PM PT - Jan 20th, 2025</sup></div>

@nektro, your commit a89132b6d68ffd1e72cbd16787aa731300add13e is building: [`#10186`](https://buildkite.com/bun/bun/builds/10186)

Jarred-Sumner (changes requested):
We should not depend on a particular version of Node in `bun-types`, it should use * and infer from the user's environment.

package.json
```diff
@@ -26,9 +26,6 @@
     "autoprefixer": "^10.4.19",
     "@mdn/browser-compat-data": "~5.5.28"
   },
-  "resolutions": {
-    "bun-types": "workspace:packages/bun-types"
-  },
```
RiskyMH: i think this is needed so @types/bun wont install them from npm and instead use local

image

PR #16558, opened 2025-01-21
CSS tests and bundler plugins in serve

zackradisic:
### What does this PR do?

- Bunch of CSS tests
- Support bundler plugins in serving static routes


src/bun.js/api/server.zig
```diff
@@ -5796,6 +5796,177 @@ pub const ServerWebSocket = struct {
...
+        defer bunstr.deref();
+        const str = bunstr.toUTF8(bun.default_allocator);
+        defer str.deinit();
+        const str_slice = str.slice();
+        for (plugins.value.pending.pending_bundled_routes.items) |route| {
+            route.onPluginsRejected(str_slice);
+            route.deref();
+        }
+        plugins.value.pending.pending_bundled_routes.clearRetainingCapacity();
+        plugins.value = .{ .err = str_slice };
```
Jarred-Sumner: this is a use after free

src/bun.js/api/server.zig
```diff
@@ -5796,6 +5796,177 @@ pub const ServerWebSocket = struct {
...
+        plugins_result.ensureStillAlive();
+
+        handleOnResolve(plugins);
+
+        return JSValue.jsUndefined();
+    }
+
+    pub fn handleOnResolve(this: *ServePlugins) void {
+        this.value.pending.promise.deinit();
+        for (this.value.pending.pending_bundled_routes.items) |route| {
```
Jarred-Sumner: can we swap the values here to make this durable to any kind of recursion


```suggestion
        var pending_bundled_routes = this.value.pending.pending_bundled_routes;
        defer pending_bundled_routes.deinit();
        this.value.pending.pending_bundled_routes = .{};
        for (pending_bundled_routes.items) |route| {
```

src/bun.js/api/server.zig
```diff
@@ -5796,6 +5796,177 @@ pub const ServerWebSocket = struct {
...
+        const bunstr = e.toBunString(globalThis);
+        defer bunstr.deref();
+        const str = bunstr.toUTF8(bun.default_allocator);
+        defer str.deinit();
+        const str_slice = str.slice();
+        for (plugins.value.pending.pending_bundled_routes.items) |route| {
+            route.onPluginsRejected(str_slice);
+            route.deref();
+        }
+        plugins.value.pending.pending_bundled_routes.clearRetainingCapacity();
```
Jarred-Sumner: Ideally we would let the user be able to intercept this and handle it via the `onError` callback in Bun.serve().

IF we cannot do that for timing reasons for now, then we can instead use vm.onUnhandledRejection and pass the JSValue to that. 

We should not be stringifying

test/js/bun/http/bun-serve-html.test.ts
```diff
@@ -241,6 +241,296 @@ button:hover {
...
+
+export default p;
+`,
+    });
+
+    const { subprocess, port, hostname } = await waitForServer(dir, {
+      "/": join(dir, "index.html"),
+    });
+    const response = await fetch(`http://${hostname}:${port}/`);
+    expect(response.status).toBe(500);
```
Jarred-Sumner, line 396: Lets add a test that verifies it still 500s without crashing when failing on another request after.
```ts
{
    const response = await fetch(`http://${hostname}:${port}/`);
    expect(response.status).toBe(500);
}
```

src/bun.js/api/server.zig
```diff
@@ -5796,6 +5796,177 @@ pub const ServerWebSocket = struct {
...
+        return JSValue.jsUndefined();
+    }
+
+    pub fn handleOnResolve(this: *ServePlugins) void {
+        this.value.pending.promise.deinit();
+        for (this.value.pending.pending_bundled_routes.items) |route| {
+            route.onPluginsResolved(this.value.pending.plugins.?);
+            route.deref();
+        }
+        this.value.pending.pending_bundled_routes.clearRetainingCapacity();
```
```suggestion

```

src/bun.js/api/server.zig
```diff
@@ -5796,6 +5796,177 @@ pub const ServerWebSocket = struct {
...
+        this.ref();
+        defer this.deref();
+
+        const plugin = bun.JSC.API.JSBundler.Plugin.create(globalThis, .browser);
+        this.value.pending.plugins = plugin;
+        var sfb = std.heap.stackFallback(@sizeOf(bun.String) * 4, bun.default_allocator);
+        const alloc = sfb.get();
+        const bunstring_array = alloc.alloc(bun.String, this.value.pending.raw_plugins.len) catch bun.outOfMemory();
+        defer alloc.free(bunstring_array);
+        for (this.value.pending.raw_plugins, 0..) |raw_plugin, i| {
```
```suggestion
        for (this.value.pending.raw_plugins, bunstring_array) |specifier, *out|| {
```

src/bun.js/api/server.zig
```diff
@@ -5796,6 +5796,177 @@ pub const ServerWebSocket = struct {
...
+            bunstring_array[i] = bun.String.init(raw_plugin);
```
```suggestion
            out.* = bun.String.init(raw_plugin);
```

docs/bundler/fullstack.md
```diff
@@ -198,6 +198,52 @@ When serving your app in production, set `development: false` in `Bun.serve()`.
 - Enables `Cache-Control` headers and `ETag` headers
 - Minifies JavaScript/TypeScript/TSX/JSX files
 
+## Plugins
+
+Bun's [bundler plugins](https://bun.sh/docs/bundler/plugins) are also supported when bundling static routes.
+
+To configure plugins for `Bun.serve`, add a `plugins` array in the `[serve.static]` section of your `bunfig.toml`.
+
+For example, enable TailwindCSS on your routes by adding add the `bun-plugin-tailwind` plugin:
```
Jarred-Sumner, line 209:
```suggestion

### Using TailwindCSS in HTML routes

For example, enable TailwindCSS on your routes by adding add the `bun-plugin-tailwind` plugin:
```

src/bun.js/api/JSBundler.zig
```diff
@@ -129,6 +129,10 @@ pub const JSBundler = struct {
                 } else false;
             }
 
+            if (try config.getTruthy(globalThis, "experimentalHtml")) |enable_html| {
```
Jarred-Sumner: it's html: true

Issue #10307, opened 2024-04-16
Bun 10x slower while running OCR


derekhearst:
### What version of Bun is running?

1.1.4+fbe2fe0c3

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Clone [this repo](https://github.com/derekhearst/BunSlowdown)
run bun install
run bun index.js 
then run node index.js

### What is the expected behavior?

Bun is over 10x slower than node

### What do you see instead?

I expect bun to be equal or quicker than node

derekhearst:
https://github.com/oven-sh/bun/issues/10306\
Main isue

Jarred-Sumner:
(marking as duplicate for tracking purposes, as it was not completed)

PR #16575, opened 2025-01-21
docs: `bun run` doesn't load any bunfig files rn

RiskyMH:
alternative to #15596 by just fixing docs and not making it actually work


robobun:
<div><sup>Updated 2:06 PM PT - Jan 21st, 2025</sup></div>

@RiskyMH, your commit a1c10dbc337d20f7c25cbabe3ea7a68f0262ee19 is building: [`#10307`](https://buildkite.com/bun/bun/builds/10307)

PR #16606, opened 2025-01-22
Allow workers to continue execution if `onerror` is handled

### LINKED ISSUE #16424: `onerror` in a Worker does not prevent thread exit

1.1.42+50eec0025

Linux 6.6.69 x86_64 unknown

```js
self.onerror = (err) => {
    console.log("onerror", err)
}

setInterval(() => {
    console.log("active")
}, 100)

setTimeout(() => {
    throw new Error("hello")
}, 250)
```

In the browser, `onerror` lets the user swallow errors. Presumably, we want the same behavior.

### What do you see instead?

The thread exits after the `onerror` handler.

### Additional information

This is relatively severe, as `process.on("uncaughtException", ...)` does not function in Workers. Currently, there's no way to throw a global error and have a Worker continue execution.

kylecarbs:
Fixes #16424.

```ts
self.onerror = (event) => {
  // Stops the worker from exiting, just like the browser.
}
```

The way I did this is certainly janky, and _probably_ should not be merged in current state.

I wasn't sure how to mark the event loop alive, since the current exception handler increments unhandled errors. Maybe `onQuietUnhandledRejectionHandlerCaptureValue` should be replaced so this code is less confusing? Open to thoughts.

### DIFF
--- a/bench/install/README.md
+++ b/bench/install/README.md
@@ -23,7 +23,7 @@ $ hyperfine --prepare 'rm -rf node_modules' --runs 1 'bun install' 'pnpm install
 To run the benchmark with offline mode but without lockfiles:
 
 ```sh
-$ hyperfine --prepare 'rm -rf node_modules' --warmup 1 'rm bun.lockb && bun install' 'rm pnpm-lock.yaml && pnpm install --prefer-offline' 'rm yarn.lock && yarn --offline' 'rm package-lock.json && npm install --prefer-offline'
+$ hyperfine --prepare 'rm -rf node_modules' --warmup 1 'rm bun.lock && bun install' 'rm pnpm-lock.yaml && pnpm install --prefer-offline' 'rm yarn.lock && yarn --offline' 'rm package-lock.json && npm install --prefer-offline'
 ```
 
 ##
--- a/docs/cli/bun-install.md
+++ b/docs/cli/bun-install.md
@@ -95,7 +95,7 @@ disableManifest = false
 [install.lockfile]
 
 # Print a yarn v1 lockfile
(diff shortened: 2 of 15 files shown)


src/bun.js/web_worker.zig
```diff
@@ -375,7 +375,11 @@ pub const WebWorker = struct {
             bun.outOfMemory();
         };
         JSC.markBinding(@src());
-        WebWorker__dispatchError(globalObject, worker.cpp_worker, bun.String.createUTF8(array.slice()), error_instance);
+        const handled = WebWorker__dispatchError(globalObject, worker.cpp_worker, bun.String.createUTF8(array.slice()), error_instance);
+        if (handled) {
+            vm.unhandled_error_counter -= 1;
```
Jarred-Sumner: I'm not sure if this should be decremented

Jarred-Sumner:
Had one comment. This does need a test before we can merge it though.

src/bun.js/web_worker.zig
kylecarbs: It seemed suspect to me as well. I'll look at making a new handler instead.

kylecarbs: I'm not actually certain how else to do this without incrementing/decrementing whether it was handled or not. If you have thoughts, please lmk.

kylecarbs:
@Jarred-Sumner force pushed and GitHub auto-closed by accident 🤦

New PR is [here](https://github.com/oven-sh/bun/pull/16661) with a test added. In my testing the decrement is fine, but definitely a bit janky.

PR #16617, opened 2025-01-22
feat(s3Client): add support for AWS S3 Object Storage Class

Inqnuam:
### What does this PR do?

This PR adds support for AWS S3 Object **Storage Class** for `PutObject`, `MultipartUpload` and `PresignUrl` actions.

- [x] Code changes

- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test/js/bun/s3/s3-storage-class.test.ts test/js/bun/s3/s3.test.ts`)
- [x] I tested the implementation with AWS


packages/bun-types/bun.d.ts
```diff
@@ -1435,6 +1435,28 @@ declare module "bun" {
...
+    /**
+     * By default, Amazon S3 uses the STANDARD Storage Class to store newly created objects.
+     *
+     * @example
+     *    // Setting explicit Storage class
+     *     const file = s3("my-file.json", {
+     *       storageClass: "STANDARD_IA"
+     *     });
+     */
+    storageClass?:
```
cirospaciari, line 1447: do R2, Google, Minio support this feature too? are this the same values or service dependent?

cirospaciari:
Thank you for the PR look good lets see the tests

packages/bun-types/bun.d.ts
Inqnuam: Looks like both [Minio](https://min.io/docs/minio/linux/reference/minio-server/settings/storage-class.html) and [R2](https://developers.cloudflare.com/r2/api/s3/api/#implemented-object-level-operations) are supporting it with the same header (`x-amz-storage-class`).

Inqnuam: For Google Cloud Storage we need to replace `x-amz-*` with [`x-goog-*`](https://cloud.google.com/storage/docs/migrating)

Issue #10357, opened 2024-04-18
Possible race condition with --filter


chimericdream:
### What version of Bun is running?

1.1.4+fbe2fe0c3

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

I haven't been able to reproduce this in a trivial repository, so I will break down how our repo is set up. We have a monorepo with workspaces set up in `packages/apps`, `packages/services`, and `packages/libraries`. The various packages in the monorepo are as follows:

Apps
- React app
- Storybook

Services
- GraphQL Server 1
- GraphQL Server 2

Libraries
- Design System
- Lib 1
- Lib 2
- Lib 3
- Lib 4

Here is a quick diagram of the relationships between them:
![Screenshot 2024-04-18 at 2 59 03 PM](https://github.com/oven-sh/bun/assets/198075/d5134d60-9fa9-4175-a01f-7ea4e7c2be30)

A couple of the packages, most notably the React app, have multiple scripts (including pre- and post- scripts) that get triggered when we run `bun run --filter '@ourscope/*' build`. However, the React app fails to build almost every time because it does not properly wait for the other packages (usually the design system) to finish building. I added a shell script with a 30s pause to the beginning of the React app's build, and that caused everything to build successfully. This leads me to believe that there is a bug in the logic where Bun is keeping track of a given script's dependencies.

The storybook package does finish successfully most of the time, but I don't think that is related, as its deps are runtime and not compile-time dependencies.

### What is the expected behavior?

All of the scripts, including pre- and post-, for dependencies of a given package finish executing before any pre- or post- scripts of the package itself.

### What do you see instead?

The React app usually fails to build because it cannot resolve the dependency to the design system, which has not finished building by the time Bun tries to build the app. If I artificially delay the React app's build, it succeeds.

gvilums:
This definitely seems like a bug - it's probably an issue with the dependency detection logic for `--filter`. My hunch is that this is related to scoped packages. Would it be at all possible for you to provide the repository where you encountered this issue?

chimericdream:
Unfortunately not. It's in a private ADO repository for my company. I'll try to see if I can approximate it enough to reproduce the issue. It may or may not be today, though.

chimericdream:
[...]

PR #16624, opened 2025-01-23
Support `import with { type: "json" }` and others

### LINKED ISSUE #12175: [Loaders] Runtime does not respect `import ... with { type: "json" }`

1.1.17+bb66bba1b

Use a test json file with any extension other than "json".
For example, "data.any":
```json
{ "hello": "world" }
```

Then, write `index.ts`:
```ts
import data from "./data.any" with { type: "json" };
console.log(json);
console.log(typeof data);
```

The console should show
```text
{
  hello: "world",
}
object
```

### What do you see instead?
The console shows
```text
/path/to/file/data.any
string
```

### Additional information

It works for other types, such as "toml", "text" or "file". Only "json" (to my knowledge) is affected.

pfgithub:
### What does this PR do?

Fixes:

```ts
import config from "./.prettierrc" with { type: "json" };
console.log(config);
// before: /path/to/.prettierrc
// after: (contents)
```

```ts
await import("./some.file", {with: {type: "js"}});
// before: (crash)
// after: (imports some.file as js)
```

Breaking change: must now use `with {type: "jsonc"}` or change the file path to `.jsonc` when using `bun build` (`bun.lock`/`tsconfig.json`/... are unaffected) 

```ts
// a.ts
import config from "./my_jsonc_file.json"
console.log(config);
// before: `bun build a.ts` : works (different than `bun a.ts`)
// after: `bun build a.ts`: errors (same as `bun a.ts`)
```

fixes #12175

### DIFF
--- a/src/options.zig
+++ b/src/options.zig
@@ -1676,7 +1676,9 @@ pub const BundleOptions = struct {
         this.defines_loaded = true;
     }
 
-    pub fn loader(this: *const BundleOptions, ext: string) Loader {
+    pub const loader = loaderForExtension;
+    /// loader from {type: ""} import attribute must be preferred over the file extension
+    pub fn loaderForExtension(this: *const BundleOptions, ext: string) Loader {
         return this.loaders.get(ext) orelse .file;
     }
 
--- a/src/options.zig
+++ b/src/options.zig
@@ -906,7 +906,136 @@ pub const Loader = enum(u8) {
             else => bun.resolver.SideEffects.has_side_effects,
         };
     }
+
+    pub fn fromMimeType(mime_type: bun.http.MimeType) Loader {
+        if (strings.hasPrefixComptime(mime_type.value, "application/javascript-jsx")) {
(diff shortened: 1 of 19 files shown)


robobun:
<div><sup>Updated 4:37 PM PT - Jan 22nd, 2025</sup></div>

@pfgithub, your commit e901e7f83858e98158e3a6c70c6017b4fb4f9f54 is building: [`#10357`](https://buildkite.com/bun/bun/builds/10357)

src/options.zig
```diff
@@ -896,8 +899,147 @@ pub const Loader = enum(u8) {
...
+                    .contents = blob.sharedView(),
+                };
+                virtual_source = &virtual_source_to_use.*.?;
+            }
+        } else {
+            return error.BlobNotFound;
+        }
+    }
+
+    if (strings.eqlComptime(query, "?raw")) {
```
Jarred-Sumner, line 1025: heh

PR #16641, opened 2025-01-23
fix(node/url): respect `process.noDeprecation` in `url.parse()`

### LINKED ISSUE #16639: Difference between bun run and running the compiled binary - Deprecation warning url.parse

1.2.0+b0c5a7655

Darwin 23.6.0 arm64 arm

If I perform any axios request with a script envoked with `bun run index.ts` I see no deprecation warnings. But if I run the same script with `bun build --compile index.ts --outfile out`, `./out` I see a DeprecationWarning for `url.parse()`.

Here's an example script that shows the behaviour:
```ts
import axios from "axios";

const response = await axios.get('https://example.com');
```

I'd like to not see any unexpected deprecation warnings after compiling the binary. I'd be happy to simply turn off the warnings for the compiled binary.

### What do you see instead?

```ts
DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
 code: "DEP0169"

      at urlParse (node:url:32:45)
      at getProxyForUrl (/$bunfs/root/out:9442:56)
      at setProxy (/$bunfs/root/out:11622:66)
      at <anonymous> (/$bunfs/root/out:11867:15)
      at dispatchHttpRequest (/$bunfs/root/out:11685:57)
      at <anonymous> (/$bunfs/root/out:11671:18)
      at new Promise (1:11)
      at wrapAsync (/$bunfs/root/out:11654:10)
      at dispatchRequest (/$bunfs/root/out:12640:17)
```

### Additional information

I'm using axios 1.7.9 if that helps.

_No response_

DonIsaac:
### What does this PR do?
Closes #16639

Existing `node:url` tests cover most of this behavior. I've added tests for the rest.

### DIFF
--- a/src/js/node/url.ts
+++ b/src/js/node/url.ts
@@ -105,25 +105,11 @@ var protocolPattern = /^([a-z0-9.+-]+:)/i,
     "file:": true,
   };
 
-let urlParseWarned = false;
 function urlParse(
   url: string | URL | Url, // really has unknown type but intellisense is nice
   parseQueryString?: boolean,
   slashesDenoteHost?: boolean,
 ) {
-  const noDeprecation = process.noDeprecation;
-  const showDep = noDeprecation == null ? true : !noDeprecation;
-  if (!urlParseWarned && showDep && !lazyUtil().isInsideNodeModules()) {
-    urlParseWarned = true;
-    process.emitWarning(
-      "`url.parse()` behavior is not standardized and prone to " +
-        "errors that have security implications. Use the WHATWG URL API " +
-        "instead. CVEs are not issued for `url.parse()` vulnerabilities.",
-      "DeprecationWarning",
-      "DEP0169",
-    );
-  }
-
   if ($isObject(url) && url instanceof Url) return url;
 
   var u = new Url();
--- a/src/js/node/url.ts
+++ b/src/js/node/url.ts
@@ -111,7 +111,8 @@ function urlParse(
   parseQueryString?: boolean,
   slashesDenoteHost?: boolean,
 ) {
-  const showDep = process.noDeprecation == null ? true : !process.noDeprecation;
+  const noDeprecation = process.noDeprecation;
+  const showDep = noDeprecation == null ? true : !noDeprecation;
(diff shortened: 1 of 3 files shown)


robobun:
<div><sup>Updated 10:25 PM PT - Jan 22nd, 2025</sup></div>

@DonIsaac, your commit 0fb82ea4e3903d46bcb0c1adfa89a7ba3195f21c is building: [`#10383`](https://buildkite.com/bun/bun/builds/10383)

src/js/node/url.ts
```diff
@@ -111,7 +111,8 @@ function urlParse(
   parseQueryString?: boolean,
   slashesDenoteHost?: boolean,
 ) {
-  if (!urlParseWarned && !lazyUtil().isInsideNodeModules()) {
+  const showDep = process.noDeprecation == null ? true : !process.noDeprecation;
```
Jarred-Sumner: this reads `process.noDeprecation` twice when we don't need to.
```suggestion
  const noDeprecation = process.noDeprecation;
  const showDep = noDeprecation == null ? true : !noDeprecation;
```

test/js/node/url/url.test.ts
```diff
@@ -79,3 +81,18 @@ it("URL constructor throws ERR_MISSING_ARGS", () => {
   // @ts-expect-error
   expect(err?.code).toEqual("ERR_MISSING_ARGS");
 });
+
+describe("url.parse()", () => {
+  it("respects process.noDeprecation", () => {
+    const dir = tempDirWithFiles("url-parse", {
```
cirospaciari: can we add a test for `noDeprecation=false` too? so we do not have regressions by always disallowing it.

Jarred-Sumner:
Can we just not print any of these deprecation warnings and comment out the tests that check they get printed? Or pass some flag to opt in to printing them we pass to node tests? It’s really noisy and offers no value for users most of the time

DonIsaac:
`url.parse()` warning is gone. I've left the other one b/c it warns about correctness.

PR #10385, opened 2024-04-20
fix(bundler): fix --compile with asset files on windows

paperclover:
### What does this PR do?

a regression

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #16642, opened 2025-01-23
node: fix buffer includes+indexof

nektro:


robobun:
<div><sup>Updated 10:55 PM PT - Jan 22nd, 2025</sup></div>

@nektro, your commit 41a3c16b7f341262b3f4637a57feaa31485cca3b is building: [`#10385`](https://buildkite.com/bun/bun/builds/10385)

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1350,108 +1403,97 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
-            RETURN_IF_EXCEPTION(scope, -1);
-
-            if (std::isnan(byteOffset_) || std::isinf(byteOffset_)) {
-                byteOffset = last ? length - 1 : 0;
-            } else if (byteOffset_ < 0) {
-                byteOffset = length + static_cast<int64_t>(byteOffset_);
-            } else {
-                byteOffset = static_cast<int64_t>(byteOffset_);
-            }
+    byteOffsetValue = jsDoubleNumber(byteOffsetValue.toNumber(lexicalGlobalObject));
```
Jarred-Sumner: this call again to jsDoubleNumber is unnecesasry

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1350,108 +1403,97 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
+    if (valueValue.isString()) {
+        if (!encoding.has_value()) {
+            return Bun::ERR::UNKNOWN_ENCODING(scope, lexicalGlobalObject, encodingString);
+        }
+        auto* str = valueValue.toStringOrNull(lexicalGlobalObject);
+        RETURN_IF_EXCEPTION(scope, -1);
+        ssize_t byteOffset = indexOfOffset(byteLength, byteOffsetD, str->length(), dir);
+        if (byteOffset == -1) return -1;
+        if (str->length() == 0) return byteOffset;
+        JSC::EncodedJSValue encodedBuffer = constructFromEncoding(lexicalGlobalObject, str, encoding.value());
```
Jarred-Sumner: why does this return an EncodedJSValue?

nektro: it calls 4 separate functions based on the encoding and two are in zig it seems

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1335,6 +1366,28 @@ static int64_t indexOf(const uint8_t* thisPtr, int64_t thisLength, const uint8_t
     return -1;
 }
 
+static int64_t indexOf16(const uint8_t* thisPtr, int64_t thisLength, const uint8_t* valuePtr, int64_t valueLength, int64_t byteOffset)
+{
+    size_t finalresult = 0;
```
```suggestion
    int64_t finalresult = 0;
```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1346,108 +1399,112 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
 {
-    auto& vm = JSC::getVM(lexicalGlobalObject);
-    auto scope = DECLARE_THROW_SCOPE(vm);
-    if (callFrame->argumentCount() < 1) {
-        throwVMError(lexicalGlobalObject, scope, createNotEnoughArgumentsError(lexicalGlobalObject));
-        return -1;
+    ssize_t byteOffset = indexOfOffset(byteLength, byteOffsetD, 1, !last);
+    if (byteOffset == -1) return -1;
+    if (last) {
+        for (int64_t i = byteOffset; i >= 0; --i) {
```
Jarred-Sumner: let's use `memrchr` if it's glibc linux

https://github.com/lattera/glibc/blob/master/sysdeps/x86_64/memrchr.S

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1346,108 +1399,112 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
-            if (callFrame->argumentCount() > 2) {
-                EnsureStillAliveScope encodingValue = callFrame->uncheckedArgument(2);
-                if (!encodingValue.value().isUndefined()) {
-                    encoding = parseEncoding(lexicalGlobalObject, scope, encodingValue.value());
-                    RETURN_IF_EXCEPTION(scope, -1);
-                }
-            }
-        }
+    if (valueValue.isNumber()) {
+        uint8_t byteValue = (valueValue.toInt32(lexicalGlobalObject)) % 256;
```
Jarred-Sumner: max int32 is intentional? and not int64 or asAnyInt?

Jarred-Sumner (changes requested):
A couple questions and nitpicks

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1346,109 +1398,119 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
-        const uint8_t* typedVectorValue = arrayValue->typedVector();
-        if (last) {
-            return lastIndexOf(typedVector, length, typedVectorValue, lengthValue, byteOffset);
-        } else {
-            return indexOf(typedVector, length, typedVectorValue, lengthValue, byteOffset);
-        }
-    } else if (value.isNumber()) {
-        uint8_t byteValue = static_cast<uint8_t>((value.toInt32(lexicalGlobalObject)) % 256);
+    if (valueValue.isNumber()) {
+        uint8_t byteValue = valueValue.asAnyInt() % 256;
```
Jarred-Sumner: Let's keep the original cast here because it avoids an assertion failure when passing a non-integer value
```suggestion
        uint8_t byteValue = static_cast<uint8_t>((value.toInt32(lexicalGlobalObject)) % 256);
```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1322,19 +1322,71 @@ extern "C" void* zig_memmem(const void* haystack, size_t haystack_len, const voi
...
     auto start = thisPtr + byteOffset;
-
     auto it = static_cast<uint8_t*>(MEMMEM_IMPL(start, static_cast<size_t>(thisLength - byteOffset), valuePtr, static_cast<size_t>(valueLength)));
     if (it != NULL) {
         return it - thisPtr;
     }
     return -1;
 }
 
+static int64_t indexOf16(const uint8_t* thisPtr, int64_t thisLength, const uint8_t* valuePtr, int64_t valueLength, int64_t byteOffset)
```
Jarred-Sumner, line 1362: Something feels off about this function but I can't quite place why

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1319,24 +1319,59 @@ static JSC::EncodedJSValue jsBufferPrototypeFunction_fillBody(JSC::JSGlobalObjec
...
-    }
-    return -1;
+static int64_t indexOf16(const uint8_t* thisPtr, int64_t thisLength, const uint8_t* valuePtr, int64_t valueLength, int64_t byteOffset)
+{
+    if (thisLength == 1) return -1;
+    if (valueLength == 1) return -1;
+    thisLength = thisLength / 2;
+    valueLength = valueLength / 2;
+    byteOffset = byteOffset / 2;
+    auto haystack = std::span<const uint16_t>((const uint16_t*)(thisPtr), thisLength).subspan(byteOffset);
```
heimskr, line 1369: This will be slow if thisPtr isn't 2-byte aligned

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1319,24 +1319,59 @@ static JSC::EncodedJSValue jsBufferPrototypeFunction_fillBody(JSC::JSGlobalObjec
...
+    auto needle = std::span<const uint16_t>((const uint16_t*)(valuePtr), valueLength);
```
heimskr, line 1370: Same here, you should be sure that valuePtr is 2-byte aligned. If you have a guarantee that it's 2-byte aligned then please ignore this comment and the previous.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1319,24 +1319,59 @@ static JSC::EncodedJSValue jsBufferPrototypeFunction_fillBody(JSC::JSGlobalObjec
...
-        return it - thisPtr;
-    }
-    return -1;
+static int64_t indexOf16(const uint8_t* thisPtr, int64_t thisLength, const uint8_t* valuePtr, int64_t valueLength, int64_t byteOffset)
+{
+    if (thisLength == 1) return -1;
+    if (valueLength == 1) return -1;
+    thisLength = thisLength / 2;
+    valueLength = valueLength / 2;
+    byteOffset = byteOffset / 2;
```
heimskr: Use `/= 2` if you feel it would be cleaner

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1319,24 +1319,59 @@ static JSC::EncodedJSValue jsBufferPrototypeFunction_fillBody(JSC::JSGlobalObjec
...
 
-    auto it = static_cast<uint8_t*>(MEMMEM_IMPL(start, static_cast<size_t>(thisLength - byteOffset), valuePtr, static_cast<size_t>(valueLength)));
-    if (it != NULL) {
-        return it - thisPtr;
-    }
-    return -1;
+static int64_t indexOf16(const uint8_t* thisPtr, int64_t thisLength, const uint8_t* valuePtr, int64_t valueLength, int64_t byteOffset)
+{
+    if (thisLength == 1) return -1;
+    if (valueLength == 1) return -1;
```
heimskr, line 1365: Is it possible that zero or negative values will end up here? If so, would `<= 1` be better?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1319,24 +1319,59 @@ static JSC::EncodedJSValue jsBufferPrototypeFunction_fillBody(JSC::JSGlobalObjec
...
 
-#if OS(WINDOWS)
-extern "C" void* zig_memmem(const void* haystack, size_t haystack_len, const void* needle, size_t needle_len);
-#define MEMMEM_IMPL zig_memmem
-#else
-#define MEMMEM_IMPL memmem
-#endif
+static ssize_t indexOfOffset(size_t length, ssize_t offset_i64, ssize_t needle_length, bool is_forward)
+{
+    ssize_t length_i64 = static_cast<ssize_t>(length);
```
heimskr: This is very nitpicky, but a common practice when declaring a variable equal to the result of a cast is to use `auto` on the left-hand side so that it's not possible for you to later change the type on one side and forget to update the other to match. In this case, that would be
```c++
auto length_i64 = static_cast<ssize_t>(length);
```

src/bun.js/bindings/JSBuffer.cpp
heimskr: Instead of mutating byteOffset, I would declare a new variable equal to byteOffset / 2 (perhaps named `u16Offset` or similar) because you later use `byteOffset` for an offset in a span of 2-byte values, which can be confusing.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1350,109 +1385,110 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
-    auto value = callFrame->uncheckedArgument(0);
-    WebCore::BufferEncodingType encoding = WebCore::BufferEncodingType::utf8;
+static int64_t indexOfString(JSC::JSGlobalObject* lexicalGlobalObject, bool last, const uint8_t* typedVector, size_t byteLength, double byteOffsetD, JSString* str, BufferEncodingType encoding)
+{
+    ssize_t byteOffset = indexOfOffset(byteLength, byteOffsetD, str->length(), !last);
+    if (byteOffset == -1) return -1;
+    if (str->length() == 0) return byteOffset;
+    JSC::EncodedJSValue encodedBuffer = constructFromEncoding(lexicalGlobalObject, str, encoding);
+    auto* arrayValue = JSC::jsCast<JSC::JSUint8Array*>(JSC::JSValue::decode(encodedBuffer));
+    int64_t lengthValue = static_cast<int64_t>(arrayValue->byteLength());
```
heimskr: See above comment about using `auto` on the LHS

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1350,109 +1385,110 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
+    const uint8_t* typedVectorValue = arrayValue->typedVector();
+    if (last) {
+        return lastIndexOf(typedVector, byteLength, typedVectorValue, lengthValue, byteOffset);
+    } else {
+        if (encoding == BufferEncodingType::ucs2) return indexOf16(typedVector, byteLength, typedVectorValue, lengthValue, byteOffset);
+        return indexOf(typedVector, byteLength, typedVectorValue, lengthValue, byteOffset);
```
heimskr: Could be flattened to
```c++
if (last) {
    return lastIndexOf(...);
}
if (encoding == ...::ucs2) {
    return indexOf16(...);
}
return indexOf(...);
```

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1350,109 +1385,110 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
+    ssize_t byteOffset = indexOfOffset(byteLength, byteOffsetD, lengthValue, !last);
+    if (byteOffset == -1) return -1;
+    if (lengthValue == 0) return byteOffset;
+    const uint8_t* typedVectorValue = array->typedVector();
+    if (last) {
+        return lastIndexOf(typedVector, byteLength, typedVectorValue, lengthValue, byteOffset);
+    } else {
+        if (encoding == BufferEncodingType::ucs2) return indexOf16(typedVector, byteLength, typedVectorValue, lengthValue, byteOffset);
+        return indexOf(typedVector, byteLength, typedVectorValue, lengthValue, byteOffset);
+    }
```
heimskr: Also a candidate for flattening if you think it would look better

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -1350,109 +1385,110 @@ static int64_t lastIndexOf(const uint8_t* thisPtr, int64_t thisLength, const uin
...
-        const uint8_t* typedVectorValue = arrayValue->typedVector();
-        if (last) {
-            return lastIndexOf(typedVector, length, typedVectorValue, lengthValue, byteOffset);
-        } else {
-            return indexOf(typedVector, length, typedVectorValue, lengthValue, byteOffset);
-        }
-    } else if (value.isNumber()) {
-        uint8_t byteValue = static_cast<uint8_t>((value.toInt32(lexicalGlobalObject)) % 256);
+    if (valueValue.isNumber()) {
+        uint8_t byteValue = static_cast<uint8_t>((valueValue.toInt32(lexicalGlobalObject)) % 256);
```
heimskr: Perhaps `auto` on the LHS

src/bun.js/bindings/JSBuffer.cpp
nektro: this doesn't seem to be an issue in practice

src/bun.js/bindings/JSBuffer.cpp
nektro: this doesn't seem to be an issue in practice

src/bun.js/bindings/JSBuffer.cpp
nektro: it is not

heimskr (approved):

Issue #16655, opened 2025-01-23
Bun aws lambda layer puts query params in request header


lovisschmidt:
### What version of Bun is running?

1.1.43+76800b049

### What platform is your computer?

arm64

### What steps can reproduce the bug?

The bun aws lambda layer puts all query params of a received request into the header params and removes them from the url that gets passed down. This results in query params being not considered by downstream operations.
Example:
I have an url like `https://myhost.com/mypath?input=44`. However, my downstream handler of the request (trpc) only receives as url `https://myhost.com/mypath`.

### What is the expected behavior?

I am expecting to receive the original url in the downstream handling, i.e. `https://myhost.com/mypath?input=44`.

### What do you see instead?

Reading the code, I see that a new `Request` object is being created, where the `queryStringParameters` are put on the header object and the url is build as
```
const request = event.requestContext;
...
const url = new URL(request.path, `${proto}://${hostname}/`);
```
but `request.path` is missing the queryStringParameters.

Not nowing the details, I would for example expect the url to be build as
```
const request = event.requestContext;
...
const url = new URL(request.path, `${proto}://${hostname}/`);

for (const [name, value] of Object.entries(event.queryStringParameters ?? {})) {
    url.searchParams.set(name, value)
 }
```
which might not result in the original url but at least keep the correct information.

lovis-ff:
The solution is as described above for `multiValueQueryStringParameters` in `formatHttpEventV1`, i.e.
```
for (const [name, values] of Object.entries(event.multiValueQueryStringParameters ?? {})) {
    for (const value of values ?? []) {
      url.searchParams.append(name, value);
    }
  }
```
The http event is version 2 and the `formatHttpEventV2` does not handle the queryStringParameters.

lovis-ff:
This got fixed via [#16684](https://github.com/oven-sh/bun/pull/16684)

PR #16657, opened 2025-01-23
fix(bun-lambda): handle queryStringParameters as searchParams for v2 http events

lovisschmidt:
### What does this PR do?

Fix for https://github.com/oven-sh/bun/issues/16655.

- [ ] Code changes


lovis-ff:
Thanks, that looks good!

PR #16661, opened 2025-01-23
Allow workers to continue execution if `onerror` is handled

kylecarbs:
Accidentally closed from force-pushing superseding #16606.


kylecarbs:
@Jarred-Sumner friendly ping. This would unblock my use-case.

PR #16664, opened 2025-01-23
load local `bunfig.toml` for `bun run` earlier (for `run.bun` option)

### LINKED ISSUE #11445: bunfig's [run] section not working

RiskyMH:
Alternative to #15596 where it now only impacts `bun run` for the same cwd dir. This does not effect `bunx` ([even though according to code it should load it](https://github.com/oven-sh/bun/blob/7830e156505b3d00397361df0acdfb80bfb1474c/src/cli.zig#L2597-L2628)), and isnt as fancy as `bun install` where it ensures to check the bunfig in `package.json` dir. 

This shouldn't have any performance issues because its already loading the file, but now its loading earlier so it can use `run.bun` option.

Fixes #11445, (as well as fixes #15484, fixes #15483, fixes #17064)


robobun:
<div><sup>Updated 8:55 AM PT - Jan 23rd, 2025</sup></div>

@RiskyMH, your commit d4ed8f8b60f6d4aea79d49319de94cc6a6304716 is building: [`#10401`](https://buildkite.com/bun/bun/builds/10401)

pfgithub (dismissed):

src/cli/run_command.zig
```diff
@@ -1397,17 +1397,17 @@ pub const RunCommand = struct {
 
         // setup
 
+        if (!ctx.debug.loaded_bunfig) {
+            bun.CLI.Arguments.loadConfigPath(ctx.allocator, true, "bunfig.toml", ctx, .RunCommand) catch {};
+        }
```
pfgithub: I think this can be moved above the fast run code and in `fn maybeOpenWIthBunJS`, remove the part where it loads bunfig.toml

pfgithub:
Should add a test that shows what this affects and makes sure it doesn't regress

src/cli/run_command.zig
RiskyMH: Fair, but I think there is some more checks one there like if its just a file to run, and we don't really need to load `bunfig` for them. I could also load it in main cli part like before PR too, but people seem to just really want it for running pkg scripts.

RiskyMH:
> Should add a test that shows what this affects and makes sure it doesn't regress

Done, I think my 3 cases should be appropriate:
* loads the config on same cwd
* doesn't load it in sub dir, even though its next to package.json
* doesn't load global

The second one could in theory be changed, but I am trying not have any perf change this time (and be minimal changes). `bun install` is an example of command that does read on same dir.

src/cli/run_command.zig
pfgithub: bunfig is currently unconditionally loaded in maybeOpenWithBunJS, so it should have no impact

RiskyMH: Ah ok my bad, i didn't look close enough at the code. Fixed!

KholdStare:
@pfgithub @RiskyMH Any updates on this? Seems worth merging

RiskyMH:
I haven't worked out the right method of doing this yet. There are many assumptions made about how bun is launched, and I don't want to brake sublaunched instances (ie the standalone running another standalone). Just so many hings I have seen and realized that I need to do it properly to be the least confusing.

also the tests have a weird failure for `bunx` that is confusing me

RiskyMH:
Ignore my previous comment I thought this was another pr.

I'm not sure whats blocking this but would be nice to have consistency and run this.

pfgithub (approved):

PR #16684, opened 2025-01-24
bun-lambda: fix http v2 event query parsing

AndrewBarba:
### What does this PR do?

Fixes the bad parsing logic when constructing the V2 event:
1. Properly adds query params to the url and not headers
2. Does not split query values on comma, this is not to spec and breaks assumptions

- [x] Code changes

### How do I know this works?

I have a Bun API I deployed to Lambda with this fix. That API verifies the full request signature on every request, and part of that verification sorts and signs all the query parameters. With this fix I've yet to see a failing request so I'm quite confident it works now.


packages/bun-lambda/runtime.ts
```diff
@@ -364,17 +364,15 @@ function formatHttpEventV2(event: HttpEventV2): Request {
       headers.append(name, value);
     }
   }
-  for (const [name, values] of Object.entries(event.queryStringParameters ?? {})) {
-    for (const value of values.split(",")) {
```
AndrewBarba, line 368: The split on comma is very bad. I have parameters like: `?select=name,age` and this is certainly not meant to be a multi value on select.

packages/bun-lambda/runtime.ts
```diff
@@ -364,17 +364,15 @@ function formatHttpEventV2(event: HttpEventV2): Request {
...
-      headers.append(name, value);
```
AndrewBarba, line 369: This is the primary bug, adding these values to headers instead of url

Issue #10436, opened 2024-04-22
crash 


nicklvh:
### How can we reproduce the crash?

run bun src/index.ts in my project

### Relevant log output

```shell
============================================================
Bun v1.1.5 (ab79940e) Windows x64
Args: "C:\Users\iamni\.bun\bin\bun.exe", "."
Features: Bun.stdin(2) dotenv fetch transpiler_cache(6) tsconfig(4) WebSocket 
Builtins: "bun:main" "node:buffer" "node:crypto" "node:events" "node:fs" "node:fs/promises" "node:http" "node:net" "node:os" "node:path" "node:perf_hooks" "node:stream" "node:string_decoder" "node:timers" "node:timers/promises" "node:tls" "node:url" "node:util" "node:util/types" "node:zlib" "node:worker_threads" "undici" "ws" 
Elapsed: 940ms | User: 281ms | Sys: 109ms
RSS: 0.22GB | Peak: 0.22GB | Commit: 0.26GB | Faults: 54890

panic(main thread): Segmentation fault at address 0xFFFFFFFFFFFFFFFF
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
```

### Stack Trace (bun.report)

Bun v1.1.5 ([#10417](https://github.com/oven-sh/bun/pull/10417)) on windows x86_64 [AutoCommand]

Segmentation fault at address 0xFFFFFFFFFFFFFFFF

- `??`
- `??`
- `uv__process_poll_req`
- `??`
- `uv_run`
- [`uws.zig:2617`](https://github.com/oven-sh/bun/blob/ab79940e6a621b902589016d58bec92e33fdebb6/src/deps/uws.zig#L2617): `run`
- [`event_loop.zig:1464`](https://github.com/oven-sh/bun/blob/ab79940e6a621b902589016d58bec92e33fdebb6/src/bun.js/event_loop.zig#L1464): `waitForPromise`
- [`javascript.zig:1149`](https://github.com/oven-sh/bun/blob/ab79940e6a621b902589016d58bec92e33fdebb6/src/bun.js/javascript.zig#L1149): `waitForPromise`
- `JSC__VM__holdAPILock`
- [`shimmer.zig:186`](https://github.com/oven-sh/bun/blob/ab79940e6a621b902589016d58bec92e33fdebb6/src/bun.js/bindings/shimmer.zig#L186): `cppFn`

Features:

cirospaciari:
[...]

PR #16685, opened 2025-01-24
Allow indented inline snapshots

### LINKED ISSUE #16403: bun test `toMatchInlineSnapshot` fails due to backtick string indentation

1.1.44-canary.57+cc8ec65e3

Darwin 23.6.0 arm64 arm

I wanted to test the compatibility of `@apollo/server` with bun and wanted to try out bun's own jest compatible test runner `bun test`

when i ran the tests, i've noticed tests failing that are using `.toMatchInlineSnapshot` matchers ([which bun lists as implemented](https://bun.sh/docs/test/writing#matchers)), that otherwise succeed in node.js with jest.

Upon closer inspection, i've noticed that the "Expected" part of the failing tests are indented just like the given string in the test file, while the "Received" part shows the value as received with a different indentation. This indentation mismatch causes the test to fail, although it doesn't do that in jest.

for an example see the `errors.test.ts` in the apollo-server repository.

https://github.com/apollographql/apollo-server/blob/a94828d0bca58ba862e9d02b06d076d478b98446/packages/server/src/__tests__/errors.test.ts#L120

To confirm this behavior, i've modified this file locally to put the backtick string at the very start of the line and further modified the indentation (2 spaces instead of 4 originally) to match the result output of `bun test`. The test then passes.

`.toMatchInlineSnapshot` should ignore indentation just like in jest. At very least when the string passed is a serialized JSON object like in the example.

https://github.com/apollographql/apollo-server/blob/a94828d0bca58ba862e9d02b06d076d478b98446/packages/server/src/__tests__/errors.test.ts#L120

### What do you see instead?

_No response_

### Additional information

```
Expected: 
          {
            "extensions": {
              "code": "INTERNAL_SERVER_ERROR",
              "exception": {
                "key": "value",
              },
            },
            "message": "message",
          }
        
Received: 
{
  "extensions": {
    "code": "INTERNAL_SERVER_ERROR",
    "exception": {
      "key": "value",
    },
  },
  "message": "message",
}


      at <anonymous> (/Users/david-ix/Desktop/apollo-server/packages/server/src/__tests__/errors.test.ts:119:23)
```

pfgithub:
### What does this PR do?

Fixes #16403

Indentation in inline snapshots is ignored, and when new inline snapshots are created, they are automatically indented. Existing snapshots preserve indentation when updated.

### DIFF
--- a/src/bun.js/test/expect.zig
+++ b/src/bun.js/test/expect.zig
@@ -2546,14 +2546,17 @@ pub const Expect = struct {
         var hint = hint_string.toSlice(default_allocator);
         defer hint.deinit();
 
-        const value: JSValue = try this.fnToErrStringOrUndefined(globalThis, try this.getValue(globalThis, thisValue, "toThrowErrorMatchingSnapshot", "<green>properties<r><d>, <r>hint"));
+        const value: JSValue = (try this.fnToErrStringOrUndefined(globalThis, try this.getValue(globalThis, thisValue, "toThrowErrorMatchingSnapshot", "<green>properties<r><d>, <r>hint"))) orelse {
+            const signature = comptime getSignature("toThrowErrorMatchingSnapshot", "", false);
+            return this.throw(globalThis, signature, "\n\n<b>Matcher error<r>: Received function did not throw\n", .{});
+        };
 
         return this.snapshot(globalThis, value, null, hint.slice(), "toThrowErrorMatchingSnapshot");
     }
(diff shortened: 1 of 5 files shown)


robobun:
<div><sup>Updated 9:00 PM PT - Jan 23rd, 2025</sup></div>

@pfgithub, your commit 515606b36a24ab81602551a11229d72c7dc953a2 is building: [`#10436`](https://buildkite.com/bun/bun/builds/10436)

PR #16688, opened 2025-01-24
cpp: synchronize on JSC::getVM since its more likely to be forward compatible

nektro:


robobun:
<div><sup>Updated 10:07 PM PT - Jan 23rd, 2025</sup></div>

@nektro, your commit b068c97f6b564f02009a30a65664210d7f21a9e1 is building: [`#10445`](https://buildkite.com/bun/bun/builds/10445)

PR #16700, opened 2025-01-24
Bun.sql: SCRAM-SHA-256-PLUS (channel binding) support

jawj:
### What does this PR do?

Adds support for SCRAM-SHA-256-PLUS auth (i.e. tls-server-end-point channel binding) to Bun.sql, and uses it if the server offers it.

This is my first time reading or writing Zig, so it may be inelegant or even wrong in places.

 I've done some manual testing locally using the [OpenSSL test certs](https://github.com/openssl/openssl/tree/master/test/certs) with this Docker script:

```sh
#!/bin/zsh
docker run --rm --name postgres-tls-cert \
  -p 5432:5432 \
  -e POSTGRES_USER=frodo \
  -e POSTGRES_PASSWORD=friend \
  -v ./openssl/test/certs:/etc/pgssl \
  postgres:17.2-bookworm \
  -c ssl=on \
  -c ssl_cert_file=/etc/pgssl/server-$1-cert.pem \
  -c ssl_key_file=/etc/pgssl/server-$1-key.pem

# e.g. ./docker-pg.sh ecdsa
```

And this command:

```sh
bun run build && echo 'import { sql } from "bun"; console.log(await sql`SELECT now()`)' \
| DATABASE_URL="postgresql://frodo:friend@localhost:5432/frodo?sslmode=require" build/debug/bun-debug run -
```

I'm not sure the channel binding code as it currently stands would deal correctly with RSA-PSS or Ed25519 certificates. But I can't check this, because it seems Bun.sql doesn't support those certificates in the first place? (I get `ERR_POSTGRES_CONNECTION_CLOSED` locally and `LOG:  could not accept SSL connection: no suitable signature algorithm` in the Postgres logs when I use these, even using the unmodified Bun 1.2.0 release).

I haven't written any automated tests so far. If you're otherwise willing to merge this, perhaps you could give a few pointers on these?


Jarred-Sumner:
Wow awesome. 

> I'm not sure the channel binding code as it currently stands would deal correctly with RSA-PSS or Ed25519 certificates. But I can't check this, because it seems Bun.sql doesn't support those certificates in the first place

@cirospaciari knows more than me about this, but I wonder if we need to pass `tls: { ca: ...` in the `Bun.SQL` constructor? It's also possible there's a missing configuration somewhere when creating the SSL_CTX. One other thought is some misconfiguration when we upgrade the socket from TCP -> TLS (since thats how postgres does it, iirc) versus most other socket-related code in Bun where we start with TLS at initialization time

This looks directionally good to me but we would need a test of some kind before we can merge it

jawj:
@cirospaciari I guess this may also be a bug report, then. :)

If it set up a Postgres instance using the script above, with either `./docker-pg.sh ed25519` or  `./docker-pg.sh pss` , I can connect fine with psql, but not with Bun. In both cases it looks like this:

```
Georges-Neon-MBP-M3 george 15:50 ~ % echo 'SELECT * FROM pg_stat_ssl' | psql "postgresql://frodo:friend@localhost:5432/frodo?sslmode=require"
 pid | ssl | version |         cipher         | bits | client_dn | client_serial | issuer_dn 
-----+-----+---------+------------------------+------+-----------+---------------+-----------
 157 | t   | TLSv1.3 | TLS_AES_256_GCM_SHA384 |  256 |           |               | 
(1 row)

Georges-Neon-MBP-M3 george 15:50 ~ % echo 'import { sql } from "bun"; console.log(await sql`SELECT * FROM pg_stat_ssl`)' \
| DATABASE_URL="postgresql://frodo:friend@localhost:5432/frodo?sslmode=require" bun run -
PostgresError: Connection closed
 code: "ERR_POSTGRES_CONNECTION_CLOSED"
```

And in Postgres logs I see:

```
2025-01-24 15:51:08.414 UTC [88] LOG:  could not accept SSL connection: no suitable signature algorithm
```

cirospaciari:
@jawj thank you for the PR, looks like CA is not being load, will investigate

jawj:
Thanks @cirospaciari. Once this is addressed I can continue to work on channel binding support.

cirospaciari:
> Thanks @cirospaciari. Once this is addressed I can continue to work on channel binding support.

I already have some fix in-place but is not merged yet, I'm investigating and fixing another Issue and after this probably tomorrow a PR will be made for it.

cirospaciari:
New update after https://github.com/oven-sh/bun/pull/17129/ arrives we can check SCRAM-256-PLUS again

jawj:
Thanks @cirospaciari. I don't think #17129 changes anything regarding certificate signing algorithms, so I had a bit more of a dig around.

The three signature algorithms that weren't working were Ed25519, Ed448 and RSASSA-PSS.

It seems BoringSSL explicitly doesn't support [either Ed488](https://commondatastorage.googleapis.com/chromium-boringssl-docs/evp.h.html#EVP_PKEY_ED448) [or RSASSA-PSS](https://commondatastorage.googleapis.com/chromium-boringssl-docs/rsa.h.html#RSA_get0_pss_params), so there's nothing to be done about those.

For some reason Ed25519 isn't in BoringSSL's [default set of verify signature algorithms](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/ssl/extensions.cc#296), but it's possible to add it in, and I'll probably open a PR for that shortly. At the same time it might be worth considering dropping RSA_PKCS1_SHA1 (as Node has done).

So — as far as I can tell, this channel binding PR does in fact work with all the signature algorithms BoringSSL supports. So I'll have a look at adding a test or two next.

cirospaciari:
> Thanks @cirospaciari. I don't think #17129 changes anything regarding certificate signing algorithms, so I had a bit more of a dig around.
> 
> The three signature algorithms that weren't working were Ed25519, Ed448 and RSASSA-PSS.
> 
> It seems BoringSSL explicitly doesn't support [either Ed488](https://commondatastorage.googleapis.com/chromium-boringssl-docs/evp.h.html#EVP_PKEY_ED448) [or RSASSA-PSS](https://commondatastorage.googleapis.com/chromium-boringssl-docs/rsa.h.html#RSA_get0_pss_params), so there's nothing to be done about those.
> 
> For some reason Ed25519 isn't in BoringSSL's [default set of verify signature algorithms](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/ssl/extensions.cc#296), but it's possible to add it in, and I'll probably open a PR for that shortly. At the same time it might be worth considering dropping RSA_PKCS1_SHA1 (as Node has done).
> 
> So — as far as I can tell, this channel binding PR does in fact work with all the signature algorithms Bun/BoringSSL supports. So I'll have a look at adding a test or two next.

 #17129 only changed the fact that now we can load the certs properly not actually the algorithms. Thank you for all contributions, IRRC in KeyObject in `node:crypto` module we still don't support `Ed448` too but we have some `RSASSA-PSS` tests in `test/js/node/crypto/crypto.key-objects.test.ts` and some mentions about BoringSSL limitations.

robobun:
@jawj, thank you for this implementation, and for the investigation into Ed25519, Ed448 and RSASSA-PSS server certificates further down the thread. #41703 still cites that investigation.

Since this PR was opened, the Postgres client was rewritten from Zig to Rust. `src/sql/postgres.zig` and `src/sql/postgres/postgres_protocol.zig` no longer exist on `main` (the client now lives in `src/sql/` and `src/sql_jsc/`). So this diff can no longer apply.

#41703 carries SCRAM-SHA-256-PLUS forward on the current client, with the same approach as this PR. The client reads the server's SASL mechanism list and picks `SCRAM-SHA-256-PLUS` when the socket is TLS. It sends `p=tls-server-end-point,,` and puts the RFC 5929 certificate hash in `c=`, with SHA-256 for MD5 and SHA-1 signatures as you did here. It uses the same `ERR_POSTGRES_SASL_NO_KNOWN_MECHANISM` error code. On top of that it adds libpq's `channel_binding=disable|prefer|require` URL parameter and `PGCHANNELBINDING`. When the certificate's signature algorithm has no digest, it falls back to plain `SCRAM-SHA-256` under `prefer` and fails under `require`.

I am closing this PR in favor of #41703. Your review there would be very welcome, in particular on the certificate hash selection. If you would prefer to keep this one open instead, say so and it can be reopened.

PR #16725, opened 2025-01-24
Add x25519 support to generateKeyPair

Jarred-Sumner:
### What does this PR do?

Add x25519 support to generateKeyPair

### How did you verify your code works?

This was mostly claude/cursor, need someone who knows more about this to verify its correct. Also need to add a node test.

PR #16730, opened 2025-01-25
test(crypto): use KeyObjects for Diffie-Hellman computation and valid…

jlucaso1:
fix test on draft PR: https://github.com/oven-sh/bun/pull/16725


cirospaciari:
Obrigado!

PR #16731, opened 2025-01-25
js: fix Buffer constructor and Buffer.from

nektro:


robobun:
<div><sup>Updated 5:12 PM PT - Jan 24th, 2025</sup></div>

@nektro, your commit 703345ba668880e8eb111856cf97c068a8f20022 is building: [`#10482`](https://buildkite.com/bun/bun/builds/10482)

nektro:
from reference https://github.com/nodejs/node/blob/v23.6.0/lib/buffer.js#L298

src/codegen/generate-classes.ts
```diff
@@ -2281,7 +2281,7 @@ JSC_DEFINE_HOST_FUNCTION(Zig::jsFunctionInherits, (JSC::JSGlobalObject * globalO
     switch (id) {
 ${jsclasses
   .map(v => v[0])
-  .map((v, i) => `    case ${i}: return JSValue::encode(jsBoolean(cell->inherits<WebCore::JS${v}>()));`)
+  .map((v, i) => `    case ${i}: return JSValue::encode(jsBoolean(jsDynamicCast<WebCore::JS${v}*>(cell) != nullptr));`)
```
Jarred-Sumner: what was the purpose of this change?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2388,35 +2442,34 @@ static inline JSC::EncodedJSValue createJSBufferFromJS(JSC::JSGlobalObject* lexi
     JSValue distinguishingArg = args.at(0);
     JSValue encodingArg = argsCount > 1 ? args.at(1) : JSValue();
     auto* globalObject = reinterpret_cast<Zig::GlobalObject*>(lexicalGlobalObject);
-
     if (distinguishingArg.isAnyInt()) {
```
Jarred-Sumner, line 2481: this should use the check that only does this operation once instead of twice

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2388,35 +2442,34 @@ static inline JSC::EncodedJSValue createJSBufferFromJS(JSC::JSGlobalObject* lexi
...
         throwScope.release();
         if (args.at(1).isString()) {
```
Jarred-Sumner, line 2483: it does the bounds check for us?

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2377,6 +2377,60 @@ JSC::JSObject* createBufferConstructor(JSC::VM& vm, JSC::JSGlobalObject* globalO
...
+
+    if (!lengthValue.isUndefined()) {
+        double lengthD = lengthValue.toNumber(lexicalGlobalObject);
+        RETURN_IF_EXCEPTION(throwScope, {});
+        if (std::isnan(lengthD)) lengthD = 0;
+        length = lengthD;
+        if (length > byteLength - offset) return Bun::ERR::BUFFER_OUT_OF_BOUNDS(throwScope, lexicalGlobalObject, "length"_s);
+    }
+
+    auto* subclassStructure = globalObject->JSBufferSubclassStructure();
```
Jarred-Sumner, line 2459: Do we need to not include the isResizableOrGrowableShared bool? There was code for this previously. I think it means the Uint8Array structure needs to be different but not 100% sure.

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2429,127 +2482,50 @@ static inline JSC::EncodedJSValue createJSBufferFromJS(JSC::JSGlobalObject* lexi
...
-            auto* uint8Array = JSC::JSUint8Array::create(lexicalGlobalObject, subclassStructure, WTFMove(buffer), offset, length);
-            if (UNLIKELY(!uint8Array)) {
-                throwOutOfMemoryError(globalObject, throwScope);
-                return JSC::JSValue::encode({});
-            }
-
-            RELEASE_AND_RETURN(throwScope, JSC::JSValue::encode(uint8Array));
+            return constructBufferFromArrayBuffer(throwScope, lexicalGlobalObject, args.size(), distinguishingArg, args.at(1), args.at(2));
         }
         default: {
```
Jarred-Sumner, line 2556: what happens if it's a symbol or something weird? it goes through the  super constructor and throws an error appropriately?

src/js/builtins/JSBufferConstructor.ts
```diff
@@ -1,78 +1,43 @@
...
+    const valueOf = value.valueOf && value.valueOf();
+    if (valueOf != null && valueOf !== value && (typeof valueOf === "string" || typeof valueOf === "object")) {
+      return Buffer.from(valueOf, encodingOrOffset, length);
     }
-  }
-  if (typeof items === "object") {
-    const data = items.data;
-    if (items.type === "Buffer" && Array.isArray(data)) {
+
+    if (value.length !== undefined || $inheritsArrayBuffer(value.buffer)) {
```
Jarred-Sumner, line 16: do we need to access .buffer specifically? could we instead check if it's any typed array view and if it isn't, only then check if it inherits arraybuffer? calling .buffer often has a side effect of cloning the contents of the array

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -2377,6 +2377,60 @@ JSC::JSObject* createBufferConstructor(JSC::VM& vm, JSC::JSGlobalObject* globalO
 
 } // namespace WebCore
 
+EncodedJSValue constructBufferFromArray(JSC::ThrowScope& throwScope, JSGlobalObject* lexicalGlobalObject, JSValue arrayValue)
+{
+    auto* globalObject = reinterpret_cast<Zig::GlobalObject*>(lexicalGlobalObject);
```
Jarred-Sumner: I don't think this reinterpret cast is correct. It likely crashes in node:vm.

Jarred-Sumner (changes requested):
A few questions

src/bun.js/bindings/JSBuffer.cpp
nektro: yes

src/js/builtins/JSBufferConstructor.ts
nektro: that happens on line 9

nektro: i wonder if that check supersedes the need for this if block entirely

src/bun.js/bindings/JSBuffer.cpp
nektro: yep
```
~/src/bun2 nektro-patch-29530*                                                                                                                                                                                                                                                     21:54:39
❯ ./build/debug-local/bun-debug -p 'new ArrayBuffer(Symbol())'
1 | new ArrayBuffer(Symbol())
    ^
TypeError: Cannot convert a symbol to a number
      at /Users/meghandenny/src/bun2/[eval]:1:1
```

src/js/builtins/JSBufferConstructor.ts
nektro: wait i have a better idea

src/bun.js/bindings/JSBuffer.cpp
nektro: wait i have a better idea

nektro: nvm that's the same error node has

src/bun.js/bindings/JSBuffer.cpp
nektro: it does now

src/bun.js/bindings/JSBuffer.cpp
nektro: nope

```
❯ node -p 'Buffer.from({ buffer: new ArrayBuffer(8, { maxByteLength: 16 }) }).buffer.resizable'  
false
```

```
❯ node -p 'Buffer.from({ buffer: new ArrayBuffer(8, { maxByteLength: 16 }) })'                 
<Buffer >
```

heimskr:
This causes `Buffer.from(new ArrayBuffer())` to throw a TypeError

PR #16735, opened 2025-01-25
Fix bugs in Workers

190n:
### What does this PR do?

Cherry-picked changes from #15557. Several Worker test failures on that branch's CI runs could also happen on `main` with more aggressive GC:

- Store a Worker's online and closing flags together in an atomic int
- Disambiguate a worker that _has_ terminated from a worker that _has been requested to_ terminate, and store those bits in another atomic bitfield
- Fix a worker's reference count not being incremented when it is referenced by Zig code
- Mark events that can only fire once with `{ once: true }`


robobun:
<div><sup>Updated 7:16 PM PT - Jan 24th, 2025</sup></div>

@190n, your commit 2d9634dfcc66c54b06b9af7596fafb4a1f79f018 is building: [`#10503`](https://buildkite.com/bun/bun/builds/10503)

Issue #10540, opened 2024-04-26
Bun installation doc doesn't work on Alpine


NamesMT:
### What platform is your computer?

Windows 11 WSL2 2.2.2.0 + Alpine 3.19.1

### What steps can reproduce the bug?

Follow installation steps at https://bun.sh/docs/installation or homepage.

### What is the expected behavior?

Bun is installed and executable

### What do you see instead?

Bun is not executable

### Additional information

I have tried investigating the [Dockerfile](https://github.com/oven-sh/bun/blob/main/dockerhub/alpine/Dockerfile) that builds the working alpine version.
After downloading and installing two glibc apks, bun could now be executed.

Jarred-Sumner:
Duplicate of #918

NamesMT:
@Jarred-Sumner Is a better error message for the time being possible?

If the glibc apks wasn't installed, currently the error is: `no such file or directory: bun`, which is really weird/odd and have pushed me away from trying bun for a long time, because there is no information to try and fix it.

Jarred-Sumner:
`no such file or directory` is surprising -- i would expect that specific error to be due to `$PATH`.

We do need to make a better error message here

PR #16745, opened 2025-01-25
ci: set build-zig timeout to 25 minutes

nektro:
the default timeout is 1 hour so this will make it auto-restart much quicker
most of the zig builds finish much quicker but sometimes linux takes a bit longer because of LTO
the ideal fix is figuring out why there's a hang/deadlock but in the meantime this prevents build time from ballooning


robobun:
<div><sup>Updated 11:44 PM PT - Jan 24th, 2025</sup></div>

@nektro, your commit 4276d1a02070e2a18d347f5784a7d15d3936b91d is building: [`#10540`](https://buildkite.com/bun/bun/builds/10540)

Issue #10547, opened 2024-04-26
bun install ERROR


Bader-Idris:
### How can we reproduce the crash?

same cause as in `#10546` and `#10545`

### Stack Trace (bun.report)

Bun v1.1.5 ([#10527](https://github.com/oven-sh/bun/pull/10527)) on windows x86_64_baseline [AddCommand]

Illegal instruction at address 0x7FF67B6B8831

- [`tccgen.c:4947`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/tinycc/tccgen.c#L4947): `post_type`
- [`tccgen.c:5051`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/tinycc/tccgen.c#L5051): `post_type`
- [`Allocator.zig:314`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/zig/lib/std/mem/Allocator.zig#L314): `free`
- [`semver.zig:827`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/install/semver.zig#L827): `order`
- [`writer.zig:23`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/zig/lib/std/io/writer.zig#L23): `writeAll`
- [`multi_array_list.zig:74`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/zig/lib/std/multi_array_list.zig#L74): `items`
- ??? at `0x153e0` in /KERNEL32.DLL
- ??? at `0x485b` in /ntdll.dll

PR #16747, opened 2025-01-25
Don't check if the same files exist 248 extra times

Jarred-Sumner:
### What does this PR do?

This only impacts build.zig (our internal build)

This is part of why `zig build` on Windows is really really slow


robobun:
<div><sup>Updated 1:50 AM PT - Jan 25th, 2025</sup></div>

@Jarred-Sumner, your commit 15b4ac3753ac1163c8137df721d6e20cf768722b is building: [`#10547`](https://buildkite.com/bun/bun/builds/10547)

Issue #10549, opened 2024-04-26
bun 1.1.5 crash on `bun install` (windows)


jexroid:
### Relevant log output

```shell
bun install
[1.01ms] ".env"
bun install v1.1.5 (b257a309)
  🚚 bun-types... ============================================================
Bun v1.1.5 (b257a309) Windows x64 (baseline)
Args: "bun", "install"
Features: dotenv
Elapsed: 12906ms | User: 203ms | Sys: 375ms
RSS: 63.53MB | Peak: 63.53MB | Commit: 117.64MB | Faults: 15716

panic(thread 8232): Illegal instruction at address 0x7FF7ADE08831
oh no: Bun has crashed. This indicates a bug in Bun, not your code.



package.json:

{
  "name": "AiMaker_auth",
  "version": "1.0.50",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "dev": "bun run --watch src/index.ts"
  },
  "dependencies": {
    "@elysiajs/static": "^1.0.3",
    "@prisma/client": "^5.13.0",
    "@types/jsonwebtoken": "^9.0.6",
    "bcrypt": "^5.1.1",
    "elysia": "latest",
    "jsonwebtoken": "^9.0.2",
    "zod": "^3.23.4"
  },
  "devDependencies": {
    "bun-types": "latest"
  },
  "module": "src/index.js"
}
```

### Stack Trace (bun.report)

Bun v1.1.5 ([#10527](https://github.com/oven-sh/bun/pull/10527)) on windows x86_64_baseline [InstallCommand]

Illegal instruction at address 0x7FF7ADE08831

- [`tccgen.c:4947`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/tinycc/tccgen.c#L4947): `post_type`
- [`tccgen.c:5051`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/tinycc/tccgen.c#L5051): `post_type`
- [`Allocator.zig:314`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/deps/zig/lib/std/mem/Allocator.zig#L314): `free`
- [`semver.zig:827`](https://github.com/oven-sh/bun/blob/b257a309772db2e9dfb745f97068aea6a32ca151/src/install/semver.zig#L827): `order`
[...]

PR #16748, opened 2025-01-25
Fix bug causing AbortSignal to sometimes collect too early

Jarred-Sumner:
### What does this PR do?

When passing an AbortSignal to native code, we don't use EventTarget's event listeners. 

This means the JSAbortSignal can be collected too early. This reproduces mostly on Windows.

When the `onabort` callback is commented out, this test sometimes fails on Windows.
```js
 const server_url = `http://${server.hostname}:${server.port}`;
      try {
        const signal = timeout < 0 ? AbortSignal.abort() : AbortSignal.timeout(timeout);
        // signal.onabort = function() {
        //   console.log("Abort ran", new Date());
        // }
        const res = await fetch(server_url, {
          signal: signal,
        });

        const reader = res.body?.getReader();
        let results = [];
        while (true) {
          const { done, data } = await reader?.read();
          if (data) results.push(data);
          if (done) break;
        }
        console.log("Done ran", new Date());
        expect.unreachable();
      } catch (err: any) {
        if (timeout < 0) {
          if (err.name !== "AbortError") throw err;
          expect(err.message).toBe("The operation was aborted.");
        } else {
          if (err.name !== "TimeoutError") throw err;
          expect(err.message).toBe("The operation timed out.");
        }
      }
    });
```

### DIFF
--- a/src/bun.js/bindings/webcore/AbortSignal.cpp
+++ b/src/bun.js/bindings/webcore/AbortSignal.cpp
@@ -195,6 +195,7 @@ void AbortSignal::cleanNativeBindings(void* ref)
     });
 
     std::exchange(m_native_callbacks, WTFMove(callbacks));
+    this->eventListenersDidChange();
 }
 
 // https://dom.spec.whatwg.org/#abortsignal-follow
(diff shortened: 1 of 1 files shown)


robobun:
<div><sup>Updated 2:13 AM PT - Jan 25th, 2025</sup></div>

@Jarred-Sumner, your commit cd91f5e4f671a6297a8ee132e30496dfab354ab6 is building: [`#10549`](https://buildkite.com/bun/bun/builds/10549)

Issue #13867, opened 2024-09-10
NODE_EXTRA_CA_CERTS does not work with certificate bundle pem files (it does work with node)


sean256:
### What version of Bun is running?

1.1.27

### What platform is your computer?

Darwin arm64 arm

### What steps can reproduce the bug?

Get a valid .pem file which has multiple valid certificates, point NODE_EXTRA_CA_CERTS to it, and observe it fails to work.

### What is the expected behavior?

For it to load the cert and function.

### What do you see instead?

`UNABLE_TO_GET_ISSUER_CERT_LOCALLY`

sean256:
Any known workarounds for this?

redbmk:
I was noticing that the test files (e.g. https://github.com/oven-sh/bun/blob/main/test/js/node/tls/fixtures/ca5-cert.pem) only include a single cert rather than a bundle. If you only need to support a single extra CA then you could use that instead of the bundle, but hopefully this is an easy fix

ShlomoCode:
I'm working on this.

PR #16782, opened 2025-01-27
feat: load full certificate bundles from `NODE_EXTRA_CA_CERTS`

ShlomoCode:
### What does this PR do?
Closes https://github.com/oven-sh/bun/issues/13867

- [x] Code changes

I added tests.

### P.S.
Inspired by Node.js' approach (https://github.com/nodejs/node/blob/main/src/crypto/crypto_context.cc#L215-L236 - https://github.com/nodejs/node/blob/main/src/crypto/crypto_context.cc#L565-L577)


ShlomoCode:
@cirospaciari Would you mind taking a look? Thank you!

packages/bun-usockets/src/crypto/root_certs.cpp
```diff
@@ -27,57 +27,81 @@ us_ssl_ctx_get_X509_without_callback_from(struct us_cert_string_t content) {
...
 end:
-  X509_free(x);
   BIO_free(in);
+  if (certs) {
+    sk_X509_pop_free(certs, X509_free);
+  }
+
+  char buf[256];
+  ERR_error_string_n(ERR_peek_last_error(), buf, sizeof(buf));
+  fprintf(stderr, "Warning: Ignoring extra certs from `%s`, load failed: %s\n", filename, buf);
```
cirospaciari: We should use `bun.Output.warn` in `src/output.zig` for this message, for reference you can create a Zig function and export it like we do for `Server__setIdleTimeout` in `src/bun.js/api/server.zig`

cirospaciari (changes requested):
Looks good only 1 comment, and I would ask for tests using `tls` module too

packages/bun-usockets/src/crypto/root_certs.cpp
ShlomoCode: Should I export the `warn` function in `src/output.zig`, or create a wrapper Zig function (where to place it?) and export the wrapper?

I also didn't really understand what I should test with the `tls` module, I would appreciate it if you could elaborate a little more. Thank you!

cirospaciari: You can implement a function on `src/deps/uws.zig` that do the logging calling `bun.Output.warn` since bun.Ouput.warn is in Zig and use features not compatible with C directly.
For the tests I would like to see something close what you did using Bun.serve and fetch but using `tls` module `Server` and `connect` so we can see if the certs are being loaded when using this features. Feel free to DM me on discord.

packages/bun-usockets/src/crypto/root_certs.cpp
```diff
@@ -27,57 +29,83 @@ us_ssl_ctx_get_X509_without_callback_from(struct us_cert_string_t content) {
...
   BIO_free(in);
+  if (certs) {
+    sk_X509_pop_free(certs, X509_free);
+  }
+
+  char buf[256];
+  ERR_error_string_n(ERR_peek_last_error(), buf, sizeof(buf));
+  char msg[512];
+  snprintf(msg, sizeof(msg), "ignoring extra certs from `%s`, load failed: %s", filename, buf);
+  bun_log_warn(msg);
```
cirospaciari: instead of snprintf can you pass filename and buf to bun_log_warn and handle it there?

also can you call ERR_clear_error after logging?

cirospaciari (changes requested):

packages/bun-usockets/src/crypto/root_certs.cpp
ShlomoCode: @cirospaciari Done

cirospaciari (approved):

Issue #10630, opened 2024-04-29
Bun crashes on bun add and bun upgrade


adxthya:
### How can we reproduce the crash?

`bun upgrade` or `bun add`

### Relevant log output

```shell
Bun v1.1.6 (e58d67b4) Linux x64 (baseline)
Args: "bun", "upgrade"
Elapsed: 3ms | User: 6ms | Sys: 0ms
RSS: 1.07GB | Peak: 12.13MB | Commit: 1.07GB | Faults: 0

panic: Illegal instruction at address 0x55F563A8DC1F
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
```

### Stack Trace (bun.report)

Bun v1.1.6 ([#10596](https://github.com/oven-sh/bun/pull/10596)) on linux x86_64_baseline [UpgradeCommand]

Illegal instruction at address 0x55F563A8DC1F

- `??`

RiskyMH:
We've invested a ton of time into making Bun more reliable over the last year. I'm having a really hard time reproducing this. If you're still running into this issue, please let me know and I'll re-open this

PR #16787, opened 2025-01-27
Fix memory leak in certain cases when long URLs are passed to req.url

Jarred-Sumner:
### What does this PR do?

Fix memory leak in certain cases when long URLs are passed to req.url

There is a memory leak test


robobun:
<div><sup>Updated 10:28 PM PT - Jan 26th, 2025</sup></div>

@Jarred-Sumner, your commit 80461fb302ec3e505fe88194e2a34ae4ee450e5b is building: [`#10630`](https://buildkite.com/bun/bun/builds/10630)

PR #16817, opened 2025-01-27
Enable sessionToken to be unset in `S3Client`

### LINKED ISSUE #16810: S3Client.write fails with `S3Error: X-Amz-Security-Token` and code: `InvalidArgument` when run in an AWS Lambda

1.2.0

Darwin 24.3.0 arm64 arm

I'm trying to write to an R2 (cloudflare) bucket from an AWS Lambda function and the call always fails with error `S3Error: X-Amz-Security-Token` (code: `InvalidArgument`).  I'm pretty sure that my credentials are all accurate. 

My gut tells me that the lambda function *always* provides some ENV vars to the runtime (in my case that I do not wish to use) and I'm wondering if those credentials are being picked up anyway and the api call is therefore incorrectly using the wrong credentials?

The client is setup like so: 

```
    const bunS3Client = new Bun.S3Client({
      endpoint:
        `https://${ctx.env.CLOUDFLARE_ACCOUNT}.r2.cloudflarestorage.com`,
      bucket: appId,
      accessKeyId: ctx.env.UPLOAD_AWS_ACCESS_KEY_ID,
      secretAccessKey: ctx.env.UPLOAD_AWS_SECRET_ACCESS_KEY,
    });
```

and the code that fails looks like this:

```
await bunS3Client.write(`${s3Prefix}${bundleCSSPath}`, bundleCSS, {
  type: "text/css",
});
```

I would expect that write call to succeed or at least give me a more helpful error message on what went wrong.

### What do you see instead?

The method fails with a cryptic error

### Additional information

I'm using the Bun lambda layer and bun version `1.2.0`

nicksrandall:
### What does this PR do?

Fixes #16810

This PR enables users to opt out of using the `AWS_SESSION_TOKEN` environment variable when using the S3Client. This is useful for users who are using Bun in an environment where the `AWS_SESSION_TOKEN` environment variable is set, but they do not want to use it (eg AWS Lambda functions, or session token is being used for another purpose).

WIth this change, users can opt out of using the `AWS_SESSION_TOKEN` (and `S3_SESSION_TOKEN`) by passing `null` to the `sessionToken` property in constructor. 

-   [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
-   [x] Code changes

-   I need some guidance on how to test this change.


src/s3/credentials.zig
```diff
@@ -161,6 +161,9 @@ pub const S3Credentials = struct {
                         } else {
                             return globalObject.throwInvalidArgumentTypeValue("bucket", "string", js_value);
                         }
+                    } else if (js_value.isNull()) {
+                        new_credentials.credentials.sessionToken = "";
```
cirospaciari, line 165:
```suggestion
                        new_credentials.credentials.sessionToken = "";
                        new_credentials._sessionTokenSlice = null;
```

cirospaciari (changes requested):
Only one comment, will manually test and check how we can test this using CI

cirospaciari (approved):

cirospaciari (changes requested):
LGTM for the test In `test/js/bun/s3/s3.test.ts` we should be able to start a process passing the credentials and a invalid sessionToken in the env, using Bun.spawn, and write/read to the bucket cleaning the sessionToken parameter, check `bunExe`, `bunEnv` and `allCredentials`, we can exit 1 for a error and exit 0 if everything is right, and check the exitCode

cirospaciari:
> @cirospaciari I tried to write a test but I actually couldn't figure out how to get it to work. I might need some help finishing this PR. 

Sure will help you with that!

PR #16824, opened 2025-01-28
Glob Match Rewrite

### LINKED ISSUE #14934: Recursive globs fail too early on partial path match

1.1.31+e448c4cc3

Linux 6.11.0-400.asahi.fc40.aarch64+16k aarch64 unknown

```typescript
console.log(new Bun.Glob(`**/*abc*`).match(`a/abc`))
console.log(new Bun.Glob(`**/*abc*`).match(`z/abc`))
```

```
true
true
```

### What do you see instead?

```
false
true
```

### Additional information

I'm pretty sure this is because the glob code sees the `a` and tries the parse the full `abc`, then bails out early when that fails?

probably-neb:
### What does this PR do?

Rewrites the glob match implementation (both ascii and non-ascii) by porting [the fast-glob crate](https://github.com/shulaoda/fast-glob)

This fixes #14934 as well as a few other issues I found along the way.
Additionally, based on my local testing it results in a large perf win.

The perf win I am referring to was measured by running each test in `test/js/bun/glob/match.test.ts` 10,000 times each and comparing the total runtime for the current and my new implementation. On my machine I get an average of a little over 200x faster than the current implementation. I'd be happy to upload the code for this comparison to github and share if it is of interest.

I didn't do very much more in depth perf testing because my main focus with the PR was to fix the aforementioned issues with the current implementation.

The high-level overview of the change is instead of keeping a track of a `BraceStack` to deal with braces `{foo,bar}`, it recursively creates new globs for each branch (i.e. `1{foo,bar}2` -> [`1foo2`,`1bar2`]). 
This does mean, however, that `matchImpl` must now take in an allocator parameter in order to create the sub globs. For better performance the implementation basically uses a `std.stackFallback` to stack allocate sub globs less than a predetermined length defined by the `GLOB_STACK_BUF_SIZE` variable in both `src/glob/ascii.zig` and `src/glob/GlobWalker.zig` (currently 64 in both places).

You will notice in the PR that I updated all callsites of `matchImpl` to pass in an instance of `Allocator`. I would appreciate if someone more familiar with the codebase than I would double check each place that I did so to make sure I'm using the allocator that makes the most sense in each context. Please note that in places where an allocator was not available in the context I used `bun.default_allocator` rather than change more function signatures, but I'm happy to change this if desired.

- [x] Code changes

I wrote automated tests 


probably-neb:
Here is the hyperfine output for my perf test.
original: current glob match implementation
rewrite-v0: rewritten implementation before `perf improvements` commit
rewrite-v3: rewritten implementation after `perf improvements` commit

```
❱ hyperfine -N ./original ./rewrite-v3 ./rewrite-v0
Benchmark 1: ./original
  Time (mean ± σ):     117.691 s ±  0.910 s    [User: 38.036 s, System: 79.520 s]
  Range (min … max):   116.630 s … 119.453 s    10 runs

Benchmark 2: ./rewrite-v3
  Time (mean ± σ):     566.2 ms ±  32.4 ms    [User: 565.0 ms, System: 0.0 ms]
  Range (min … max):   540.2 ms … 619.7 ms    10 runs

Benchmark 3: ./rewrite-v0
  Time (mean ± σ):      2.753 s ±  0.087 s    [User: 0.808 s, System: 1.783 s]
  Range (min … max):    2.680 s …  2.916 s    10 runs

Summary
  ./rewrite-v3 ran
    4.86 ± 0.32 times faster than ./rewrite-v0
  207.86 ± 11.99 times faster than ./original
  ```

**Comment by RiskyMH** [comment]
I'm not sure if the tests are flaky, but they do seem related: https://buildkite.com/bun/bun/builds/10675

**Comment by zackradisic** [comment]
Thanks a lot @probably-neb, porting the fast-glob crate is something that I've been meaning to do for a while but never had the time to do so!

I'll take a closer look at this PR when I get the chance, but thanks once again!

**Comment by probably-neb** [comment]
@RiskyMH they do seem related. I'll look into it

**Comment by probably-neb** [comment]
Ok something is very wrong with the Apple x64. Won't show it in buildkite and when I download it it's full of null bytes. Not sure what to do about that one. 
Seems like the glob `./**` is failing to match any paths on the windows tests that are failing. Unfortunately my Windows install decided it is unable to connect to the internet a few weeks ago so I can't test it on Windows ATM.
I'd be happy to join a call on Discord with someone who does and help them debug it tomorrow, otherwise I will just try to get Windows working again this weekend and attempt to debug it myself

File: src/glob/ascii.zig [hunk 1]
```diff
@@ -316,154 +187,230 @@ pub fn match(glob: []const u8, path: []const u8) MatchResult {
...
+    }
 
-        return if (negated) .negate_match else .no_match;
+    const max_sub_len = open_brace_index + max_branch_len + (glob.len - (close_brace_index + 1));
+
+    // PERF: doing the following results in a large performance improvement over using std.heap.stackFallback
+    var buffer = if (max_sub_len <= GLOB_STACK_BUF_SIZE) blk: {
+        var buf: [GLOB_STACK_BUF_SIZE]u8 = undefined;
+        var fixed_buf_alloc = std.heap.FixedBufferAllocator.init(&buf);
+        const buf_alloc = fixed_buf_alloc.allocator();
```
zackradisic: These 3 lines should probably be above and happen unconditionally. Haven't verified but it looks like it will cause undefined behavior as the code is exiting the scope in which they were defined

zackradisic:
Hey @probably-neb thanks once again for the PR. Bun's Glob is in need of some refactoring so I'm gonna take it from here and do some changes like removing the ascii and unicode matchers as only the ascii matcher is necessary, as well as some other changes.

There are two options:
1. you could give me write access to your fork so I can make the changes their and we can use this same PR
2. I push these changes into a branch in the Bun repo and make a PR and reference this one

Which one would you like to proceed with?

probably-neb:
Thanks a bunch, I've really been meaning to get back around to this, but I've been swamped with work and unable to get Windows working again.

I'm fine with either approach. I'd love if my name was attached in the git log somewhere but I don't feel that strongly about it.

I invited you as a collaborator to my fork. Feel free to proceed with whichever approach is easier for you

probably-neb:
Also feel free to ask questions if you have any. Happy to continue helping to the extent I can

src/glob/GlobWalker.zig
```diff
@@ -1641,51 +1580,42 @@ pub fn GlobWalker_(
...
-            var cp_len: u32 = 0;
             var prevIsBackslash = false;
             var saw_special = false;
-            while (iter.next(&cursor)) : (cp_len += 1) {
-                const c = cursor.c;
+            var i: u32 = 0;
+            var width: u32 = 0;
+            while (i < pattern.len) : (i += 1) {
+                const c = pattern[i];
+                width = bun.strings.utf8ByteSequenceLength(c);
```
Jarred-Sumner, line 1595: This should use the CodepointIterator because this code will not handle invalid surrogate pairs correctly on Windows

test/js/bun/glob/match.test.ts
```diff
@@ -24,8 +24,14 @@
 
 import { Glob } from "bun";
 import { describe, expect, test } from "bun:test";
+import { isWindows } from "harness";
 
 describe("Glob.match", () => {
+  test("WTF", () => {
```
Jarred-Sumner, line 30: Can you delete this test

src/cli/outdated_command.zig
```diff
@@ -116,22 +116,22 @@ pub const OutdatedCommand = struct {
...
             };
         }
 
-        pub fn deinit(this: @This(), allocator: std.mem.Allocator) void {
-            switch (this) {
-                .path, .name => |pattern| allocator.free(pattern),
-                else => {},
-            }
+        pub fn deinit(_: @This(), _: std.mem.Allocator) void {
+            // switch (this) {
```
Jarred-Sumner: Can you delete this function so that the next person working on this code doesn't think it's a memory leak?

src/glob/GlobWalker.zig
```diff
@@ -1697,448 +1627,52 @@ pub fn GlobWalker_(
...
                         }
-                        start_cp = cp_len + 1;
-                        start_byte = cursor.i + cursor.width;
+                        start_byte = i + width;
                     },
                     // TODO: Support other escaping glob syntax
                     else => {},
                 }
             }
+            i -|= 1;
```
Jarred-Sumner, line 1655: this seems likely to be incorrect?

bench/glob/rewrite.mjs
```diff
@@ -0,0 +1,27 @@
+import { Glob } from "bun";
```
Jarred-Sumner: can you give this a better filename?

Issue #14459, opened 2024-10-10
`bun in bun` does not run pure bun from compiled binary


thers:
### What version of Bun is running?

1.1.31

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

Make a file `main.ts`:
```ts
import { $ } from "bun";

console.log('Running main.ts entrypoint, PID:', process.pid);

switch (process.argv[2]) {
    case "main": {
        console.log("Hello from main subcommand");
        break;
    }
    case "bun": {
        const args = process.argv.slice(3).map($.escape).join(' ');

        await $`bun ${{ raw: args }}`;
        break;
    }
    default: {
        console.log('Unknown subcommand');
        process.exit(1);
    }
}
```

Build compiled binary:
```
.\bun build --compile main.ts --outfile bnb.exe
```

Run the compiled binary:
```
.\bnb bun --version
```

### What is the expected behavior?

Should print the same out put as `.\bun --version`:
```
1.1.31
```

### What do you see instead?

Error:
```
Running main.ts entrypoint, PID: 21384
Running main.ts entrypoint, PID: 20820
Unknown subcommand
 7 |     console.log("Hello from main subcommand");
 8 |     break;
 9 |   }
10 |   case "bun": {
11 |     const args = process.argv.slice(3).map($.escape).join(" ");
12 |     await $`bun ${{ raw: args }}`;
                    ^
ShellError: Failed with exit code 1
 info: {
  "exitCode": 1,
  "stderr": "",
  "stdout": "Running main.ts entrypoint, PID: 20820\nUnknown subcommand\n"
}

      at new ShellError (12:16)
      at new ShellPromise (72:16)
      at BunShell (187:35)
      at module code (B:/~BUN/root/bnb.exe:12:12)

Bun v1.1.31-canary.9+ff476313a (Windows x64)
```

### Additional information

Given that under the hood [`bun in bun`](https://bun.sh/docs/runtime/shell#builtin-commands) just runs current binary it is no mystery, but having access to pure Bun within the compiled binary is very handy for my use case.

**Use case:**
Using bun's ability to compile into one single binary is helping create a minimal distribution of a tool with assortment of smaller utilities, but at the same time, there's a desire to enable using such compiled tool as a pure bun instead of having tool and bun side-by-side, and `bun in bun` sounds like the most straightforward approach.

[...]

Issue #5414, opened 2023-09-14
Programmatic `install` API


lukeed:
### What is the problem this feature would solve?

Ability to run `bun install` within a compiled executable in host environments that dont have `bun` installed.

Also would be great for CI, task runners, and/or interactive `bun create ...` scaffolders that look to prepare/modify a repository without needing (or having the ability) to spawn a shell command.

### What is the feature you are proposing to solve the problem?

A `Bun.install` function that accepts all the options as the CLI variant. 

I imagine this is also a way to apply `.npmrc`-like modifications (#643, #4904) without relying on filesystem existence.

PuruVJ:
+1 to this, would be great to have this

Issue #10690, opened 2024-04-30
Bun crashes when trying to run OpenGL bindings for Node (NodeGL)


MrMasrozYTLIVE:
### How can we reproduce the crash?

1) Install nodegl & glfw-n-api
2) Use example code from https://github.com/EnzoAlbornoz/nodegl
3) bun index.js (With commonjs)
4) It crashes (Only with bun)

### Stack Trace (bun.report)

Bun v1.1.6 ([`e58d67b`](https://github.com/oven-sh/bun/tree/e58d67b46875280bc9002add4c3fae45472cb346)) on windows x86_64_baseline [AutoCommand]

Segmentation fault at address 0x00000C70

- ??? at `0xb48ad9` in /node.exe
- ??? at `0xb59706` in /node.exe
- ??? at `0xb5e2d8` in /node.exe
- ??? at `0xb485fb` in /node.exe
- ??? at `0xb54c44` in /node.exe
- ??? at `0xcde6c9` in /node.exe
- ??? at `0x2c3455` in /node.exe
- ??? at `0x24041` in /nodegl.node
- ??? at `0x23e9b` in /nodegl.node
- ??? at `0x2e2ab` in /nodegl.node

MrMasrozYTLIVE:
Issue still exists on bun 1.1.7 (Windows)
Bun v1.1.7 ([`b0b7db5`](https://github.com/oven-sh/bun/tree/b0b7db5c060e830801b0273b7ff6e34e3eb5c254)) on windows x86_64_baseline [RunCommand]

Segmentation fault at address 0x00000C70

- ??? at `0xb48ad9` in node.exe
- ??? at `0xb59706` in node.exe
- ??? at `0xb5e2d8` in node.exe
- ??? at `0xb485fb` in node.exe
- ??? at `0xb54c44` in node.exe
- ??? at `0xcde6c9` in node.exe
- ??? at `0x2c3455` in node.exe
- ??? at `0x24041` in nodegl.node
- ??? at `0x23e9b` in nodegl.node
- ??? at `0x2e2ab` in nodegl.node

dylan-conway:
Confirming this still exists in Bun v1.3.3. Seems to only happen on Windows

PR #16829, opened 2025-01-28
alow standalone bun executable to use local bun instance in `Bun.$`

### LINKED ISSUE #16262: Use Bun's package manager from a built single-executable app

Ultimately I would like to use Bun's package manager in a single-executable Bun app without installing Bun globally, so my app manages npm packages automatically without additional downloads. It seems stupid to require node.js/Bun installed to install npm packages when an executable has all the needed APIs somewhere there, but not exported to Bun API. My target audience is only just partially overlaps with web developers so I would avoid making them download Bun as a dependency.

I would like Bun's package manager to be exposed in its `bun` import, maybe as a `pm` namespace (stands for Package Manager), or its CLI be accessible when calling a packaged single-executable app. For JS API, I would like it to have at least these methods that match their CLI versions: `install`, `add`, `remove`, `update`, `pm ls`. I assume this is not a hard tasks as the implementation already exists.

Installing Bun/Node.js in end users' systems (and this sucks)
Having `npm` or an alternative as a package dependency in a Bun project 🤡

RiskyMH:
### What does this PR do?

Ensures you can always do this, even when in standalone and no bun in path:
```js
await Bun.$`bun --revision`
await Bun.$`bunx cowsay hi`

import { fork } from "child_process";
fork("-p", ["1 + 1"]);
```

~~currently it will use same executable if no bun in path, however standalone it just infinite loops because its running itself~~ Now it will always try to use the same bun that started it (including standalone).  fixes #16262, fixes #14459, fixes #5414

uses @pfgithub's [solution](https://github.com/oven-sh/bun/issues/16262#issuecomment-2581181312) and extends it to be able to use in safer Bun shell

**concerns:** implies the whole binary will be available forever, so could be harder to minify and remove unneeded parts

- [ ] Code changes

will need to test with the the CI's release build


robobun:
<div><sup>Updated 12:02 AM PT - Jan 28th, 2025</sup></div>

@RiskyMH, your commit 3f977f4596de62adc918b813f5720700a5b16ea1 is building: [`#10690`](https://buildkite.com/bun/bun/builds/10690)

Jarred-Sumner:
Seems nice

Needs a test though

RiskyMH:
i added tests!

Also should `bun.$` (for "bun") prefer the current executable and fallback to PATH, or be right now and prefer PATH and fallback so there is something?

The issue that this is closing wants it for `bun install` and I was wondering if maybe the user has a Bun but really old, so the standalone now will use it and get all sorts of errors.

CosmoMyzrailGorynych:
> The issue that this is closing wants it for `bun install` and I was wondering if maybe the user has a Bun but really old, so the standalone now will use it and get all sorts of errors.

My two cents,
In case of standalone executable, the main use case of exposing bun CLI to the application is making self-contained, self-sufficient packages — useful for proprietary products or those that aim for non-techie people. So my say is that a packaged app should use itself as a bun CLI to eliminate environment inconsistencies.

RiskyMH:
Makes sense, to simplify I'll just make it always try to use the same executable (otherwise it would just be weird code to check if standalone and do one bit earlier)

and it kinda made sense the way it was written in docs to use same bun

pfgithub:
Node has a child_process.fork() function to call the node executable and pass it args, it would make sense for it to work in standalone executables

```ts
import { fork } from "child_process";

fork("-p", ["1 + 1"]);
```

In node/bun, that prints `2`, but in a standalone executable it loops forever because it's launching the executable

RiskyMH:
Oh nice, I didn't know about it. Might see if it's simple to add env var there too

yus-ham:
for minifying, there should be a build config for what parts to be included into executable

src/StandaloneModuleGraph.zig
```diff
@@ -785,6 +785,12 @@ pub const StandaloneModuleGraph = struct {
         const self_exe = openSelf() catch return null;
         defer _ = Syscall.close(self_exe);
 
+        if (bun.getenvZ("BUN_SKIP_STANDALONE_MODULE_GRAPH")) |v| {
```
Jarred-Sumner:
```suggestion
        if (bun.getRuntimeFeatureFlag("BUN_SKIP_STANDALONE_MODULE_GRAPH"))  return null;
```

RiskyMH: nice, fixed

RiskyMH:
> Node has a child_process.fork() function to call the node executable and pass it args, it would make sense for it to work in standalone executables
> 
> ```ts
> import { fork } from "child_process";
> 
> fork("-p", ["1 + 1"]);
> ```
> 
> In node/bun, that prints `2`, but in a standalone executable it loops forever because it's launching the executable

Added support! I'm surprised this child_processes stuff is all in JS 😅

RiskyMH:
> for minifying, there should be a build config for what parts to be included into executable

Yeah, I don't know the plan for that, but its way out of scope for this pr. 

I'd think at least initially your suggestion of flag makes most sense (but there has to be some way of bun knowing which parts of bundle to delete - which gets harder with more interconnected stuff).

RiskyMH:
i just realized about `bunx`, i think this is also important to support if bun not installed (not as nice as `bun x`)

Issue #10691, opened 2024-04-30
bun install and bunx not work well when use husky and @commitlint/cli


sharh:
### What version of Bun is running?

1.1.6

### What platform is your computer?

Microsoft Windows NT 10.0.19045.0 x64

### What steps can reproduce the bug?

```sh
# bun init
bun init
# install deps
bun install -d @commitlint/config-conventional @commitlint/cli husky
# init git
git init -b dev
# init git hook
bunx husky init
```
`bunx husky init` not works well:
![image](https://github.com/oven-sh/bun/assets/7692602/bd28bd97-e1c3-4a15-a67c-e4ad3612e0d8)

when change to `npx`, it works well as the above picture presents.
```sh
# huasky init
npx husky init
# commit lint
echo npx --no -- commitlint --edit $1 > .husky/commit-msg
```

```sh
# git commit lint
git commit -am 'test'
```
while use `bun install` the deps，the `@commitlint/cli` not works with `husky` git hook `npx --no -- commitlint --edit $1`
![image](https://github.com/oven-sh/bun/assets/7692602/65b2bdcc-0f3a-40f4-a43e-45271469552f)

infact, the command `commitlint` is installed in `node_modules/.bin` diretory：
![image](https://github.com/oven-sh/bun/assets/7692602/990d787a-af8e-4afd-abed-179dfe1de59f)

when I change the git hook to:  `bunx --no -- commitlint --edit $1`  it works:
![image](https://github.com/oven-sh/bun/assets/7692602/df9cf5d3-ae23-42ae-9e43-baefa423627d)

when use `npm install -D @commitlint/config-conventional @commitlint/cli husky`，this can be work well:
```sh
npx --no -- commitlint --edit $1
```
![image](https://github.com/oven-sh/bun/assets/7692602/20ecfc95-69ee-4e03-8bdd-78160839d13a)

```sh
bunx --no -- commitlint --edit $1
```
![image](https://github.com/oven-sh/bun/assets/7692602/14b49e93-b8df-459c-8585-0c18c5a64934)
both `bunx` and `npx` can work well:

### What is the expected behavior?

I hope `bun install` can works well as `npm install`
```sh
bun install -d @commitlint/config-conventional @commitlint/cli husky
bunx husky init
```

```sh
npm install -D @commitlint/config-conventional @commitlint/cli husky
npx husky init
```

`bunx` should works well as `npx`
```sh
bunx --no -- commitlint --edit $1 
```

```sh
[...]

PR #16830, opened 2025-01-28
Introduce `server.abort(request: Request)` to abort an in-flight HTTP/HTTPS request

Jarred-Sumner:
### What does this PR do?

Introduce `server.abort(request: Request)` to abort an in-flight HTTP/HTTPS request.

This immediately closes the TCP socket.

One subtlety here is we have to make sure that we don't call drain microtasks directly in the abort handler. Instead, we have to go through the event loop enter/exit - so that we don't drain microtasks inside of the .abort() function call itself.

There are tests

### DIFF
--- a/test/js/bun/http/bun-serve-abort.test.ts
+++ b/test/js/bun/http/bun-serve-abort.test.ts
@@ -1,0 +1,185 @@
+import { describe, expect, test, mock } from "bun:test";
+import { bunEnv, bunExe, rejectUnauthorizedScope, tempDirWithFiles, tls } from "harness";
+
+describe("server.abort()", async () => {
+  test("after sleep", async () => {
+    using server = Bun.serve({
+      port: 0,
+
+      async fetch(request, server) {
+        await Bun.sleep(0);
+        server.abort(request);
+        return new Response("Hello, world!");
+      },
+    });
+
+    expect(async () => {
+      const response = await fetch(`http://localhost:${server.port}`);
+    }).toThrow("The socket connection was closed");
+  });
+
+  test("before sleep", async () => {
+    using server = Bun.serve({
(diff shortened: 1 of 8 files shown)


robobun:
<div><sup>Updated 12:06 AM PT - Jan 28th, 2025</sup></div>

@Jarred-Sumner, your commit ee239d7159ccd23e685ed77c05e3f18f9474dbfd is building: [`#10691`](https://buildkite.com/bun/bun/builds/10691)

cirospaciari (changes requested):
`this.flags.user_called_abort` should be part of `isAbortedOrEnded` otherwise will be a footgun in future.

Also after calling abort the request_context should be detached from AnyRequestContext to avoid accessing invalid data

I would also duplicate all tests for TLS and non-TLS

test/js/bun/http/bun-serve-abort.test.ts
```diff
@@ -0,0 +1,185 @@
...
+            },
+          }),
+        );
+      },
+    });
+
+    expect(async () => {
+      await fetch(`https://localhost:${server.port}`, {
+        tls: {
+          rejectUnauthorized: false,
```
cirospaciari, line 180:
```suggestion
          ca: tls.cert,
```

Jarred-Sumner:
> `this.flags.user_called_abort` should be part of `isAbortedOrEnded` otherwise will be a footgun in future.
> 
> Also after calling abort the request_context should be detached from AnyRequestContext to avoid accessing invalid data
> 

us_socket_close callback calls the onAbort callback, which triggers the normal request abort flow. I think it's less likely to cause bugs if we can rely on the same code paths continuing to be used without changes

PR #16847, opened 2025-01-28
Feature/s3 multi delete

### LINKED ISSUE #16520: S3 option to delete multiple objects with single request

Currently Bun's implementation of S3 only allows us to delete a single file / object at the time.

Example:
```ts
await client.delete("my-file.txt");
```

delete function should accept array of paths / keys for deletion.

Example:
```ts
await client.delete(["my-file.txt","another-file.txt"]);
```

Using S3 library from AWS, we can delete multiple keys at once.

Example:

Image

Inqnuam:
### What does this PR do?

Closes #16520

This PR adds full support for AWS DeleteObjects operations, with the same Input/Output as `@aws-sdk/s3-client`

```typescript
const client = new S3Client();

try {
  const response = await client.deleteObjects(["pictures/file_1.png", "pictures/file_3.png"]);
} catch (err: any) {
  //:(
}

try {
  const response = await client.deleteObjects(["styles/main.css", "js/footer.js", { Key: "slow.js", VersionId: "8" }]);
} catch (err: any) {
  //:(
}
```

- [x] Code changes

- [x] I ran my tests locally and they pass (`bun-debug test test/js/bun/s3/s3-multi-delete.test.ts`)
- [x] I ran my tests on AWS and they pass

If Zig files changed:

- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test/js/bun/s3/s3-multi-delete.test.ts`)


Inqnuam:
The custom-made XML parser deserves a benchmark; it should be very fast!

test/js/bun/s3/s3-multi-delete.test.ts
```diff
@@ -0,0 +1,782 @@
...
+      ...options,
+      endpoint: server.url.href,
+    });
+
+    try {
+      await client.deleteObjects([
+        // @ts-expect-error not allowed
+        {
+          VersionId: "ok",
+          // key: "wrong prop name"   // NOTE Bun issue ? When uncommented 'key' must not be returned with getTruthyComptime("Key")
```
Inqnuam: @Jarred-Sumner is this a bug or its how `getTruthyComptime` works ?

Jarred-Sumner:
@cirospaciari can you review this?

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+    ETag?: string;
+    /**
+     * Key name of the object.
+     *
+     * Replacement must be made for object keys containing special characters (such as carriage returns).
+     *
+     * For more information
+     * @see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-xml-related-constraints
+     */
+    Key: string;
```
cirospaciari:
```suggestion
    key: string;
```

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+
+  interface S3DeleteObjectsObjectIdentifier {
+    /**
+     * An entity tag (ETag) is an identifier assigned by a web server to a specific version of a resource found at a URL.
+     *
+     * This header field makes the request method conditional on ETags.
+     *
+     * Entity tags (ETags) for S3 Express One Zone are random alphanumeric strings unique to the object.
+     */
+    ETag?: string;
```
cirospaciari:
```suggestion
    etag?: string;
```

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+     * For more information
+     * @see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-xml-related-constraints
+     */
+    Key: string;
+    /**
+     * If present, the objects are deleted only if its modification times matches the provided Timestamp.
+     *
+     * This functionality is only supported for directory buckets.
+     */
+    LastModifiedTime?: string;
```
cirospaciari:
```suggestion
    lastModified?: string;
```

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+    /**
+     * If present, the objects are deleted only if its size matches the provided size in bytes.
+     *
+     * This functionality is only supported for directory buckets.
+     */
+    Size?: number;
```
cirospaciari:
```suggestion
    size?: number;
```

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+    /**
+     * Version ID for the specific version of the object to delete.
+     *
+     * This functionality is not supported for directory buckets.
+     */
+    VersionId?: string;
```
cirospaciari:
```suggestion
    versionID?: string;
```

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+     *
+     * In quiet mode the response includes only keys where the delete operation encountered an error.
+     *
+     * For a successful deletion in a quiet mode, the operation does not return any information about the delete in the response body.
+     */
+    quiet?: true;
+  } & Pick<S3Options, "accessKeyId" | "secretAccessKey" | "sessionToken" | "bucket" | "endpoint">;
+
+  interface S3DeleteObjectsResponse {
+    Deleted?: { Key: string; VersionId?: string; DeleteMarker?: boolean; DeleteMarkerVersionId?: string }[];
```
cirospaciari:
```suggestion
    deleted?: { Key: string; VersionId?: string; DeleteMarker?: boolean; DeleteMarkerVersionId?: string }[];
```

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,64 @@ declare module "bun" {
...
+    Errors?: { Code?: string; Key: string; Message?: string; VersionId?: string }[];
```
cirospaciari:
```suggestion
    errors?: { Code?: string; Key: string; Message?: string; VersionId?: string }[];
```

packages/bun-types/bun.d.ts
```diff
@@ -1946,6 +2004,7 @@ declare module "bun" {
     unlink(path: string, options?: S3Options): Promise<void>;
     delete: S3Client["unlink"];
 
+    deleteObjects(keys: S3DeleteObjectsInput[], options?: S3DeleteObjectsOptions): Promise<S3DeleteObjectsResponse>;
```
cirospaciari: we should also allow an array of S3File for `convenience`, also I think this should be a overload of the `delete` function and not another function

src/bun.js/webcore/S3Client.zig
```diff
@@ -236,6 +236,17 @@ pub const S3Client = struct {
         });
     }
 
+    pub fn deleteObjects(ptr: *@This(), globalThis: *JSC.JSGlobalObject, callframe: *JSC.CallFrame) bun.JSError!JSValue {
+        const args = callframe.argumentsAsArray(2);
+        const object_keys = args[0];
+        const options = args[1];
+
+        var blob = try S3File.constructS3FileWithS3CredentialsAndOptions(globalThis, .{ .string = bun.PathString.empty }, options, ptr.credentials, ptr.options, null, null);
```
cirospaciari: this should accept the credentials inside the options like every other method in S3Client or S3File

src/bun.js/webcore/blob.zig
```diff
@@ -3584,6 +3584,197 @@ pub const Blob = struct {
...
+                        .success => |_success| {
+                            const jsResult = JSValue.createEmptyObject(globalObject, 2);
+
+                            if (_success.deleted) |del| {
+                                defer del.deinit();
+                                const array = JSValue.createEmptyArray(globalObject, del.items.len);
+
+                                for (del.items, 0..) |item, i| {
+                                    const deletedObject = JSValue.createEmptyObject(globalObject, 1);
+                                    deletedObject.put(globalObject, JSC.ZigString.static("Key"), bun.String.init(item.key).toJS(globalObject));
```
cirospaciari: properties should start with lower case
bun.Sring.init(item.key).toJS will leak memory use `bun.String.createUTF8ForJS(globalObject, item.key)` in this case;

src/bun.js/webcore/blob.zig
```diff
@@ -3584,6 +3584,197 @@ pub const Blob = struct {
...
+            }
+
+            const Wrapper = struct {
+                promise: JSC.JSPromise.Strong,
+                store: *Store,
+
+                pub usingnamespace bun.New(@This());
+
+                pub fn resolve(result: S3.S3DeleteObjectsResult, self: *@This()) void {
+                    defer self.deinit();
```
cirospaciari: `req_body_slice` needs to be freed after the resolve

cirospaciari (changes requested):
Thank for the PR added some comments

packages/bun-types/bun.d.ts
Inqnuam: good idea to support `S3File`!

I think using method overloading will cause more confusion than benefit.

src/bun.js/webcore/S3Client.zig
Inqnuam: If im not wrong this implementation accepts all credentials options as second argument (see the tests)
In type definition `acl` `storageClass` and other non `DeleteObjects` options which I think are confusing are intentionally omitted

src/bun.js/webcore/blob.zig
Inqnuam: Thanks!

src/bun.js/webcore/blob.zig
Inqnuam: thanks

src/bun.js/webcore/blob.zig
Inqnuam: lowercasing properties will break drop-in `@aws-sdk/s3-client` compatibility

cirospaciari: Bun S3 API is not intended to be compatible with `@aws-sdk/s3-client` format but with S3 REST API

ibufu:
What is the progress?

azvyae:
Any info for this implementation?

Onassis-dev:
Looking forward on this 👀

Issue #10707, opened 2024-04-30
Unable to install gyp-based packages with Bun based image (llvm_version is not defined) (for unix-dgram)


masterful:
### What version of Bun is running?

1.1.6+e58d67b46

### What platform is your computer?

mostly n/a (docker based issue - using x86_64, though)

### What steps can reproduce the bug?

When attempting to install a package that requires gyp bindings, I understand that I need to provide a proper build environment - however, despite providing these packages, the install step fails with an error about `llvm_version` not being defined. I've done some sleuthing, and have found [some](https://github.com/oven-sh/bun/issues/9807#issuecomment-2035439988) [comments](https://github.com/oven-sh/bun/issues/9807#issuecomment-2041573065) [on](https://github.com/oven-sh/bun/issues/9807#issuecomment-2041849549) this topic - but nothing that actually resolves the issue within the Bun image itself.

While the multi-stage build referenced in one of the above comments does work, relying on a Node image to install my dependencies feels ... off. There's also the added drawback that, unless you can guarantee running on the same instance, your layers in a multi-stage build will not be well cached and so your build times increase substantially. My ultimate goal is to get a small image that is able to build gyp-based packages.

I've been able to narrow it down to demonstrate there's something a bit weird going on with the following dockerfiles:
Fails to install:
```Dockerfile
FROM    --platform=linux/amd64 oven/bun:1.1.6-debian
WORKDIR /app
RUN     echo '{"name": "app", "version": "1.0.0", "dependencies": { "unix-dgram": "^2" } }' > package.json
RUN     apt update; apt install -y build-essential curl python3
RUN     bun install
```
But if I add Node to the image:
```Dockerfile
FROM    --platform=linux/amd64 oven/bun:1.1.6-debian
WORKDIR /app
RUN     echo '{"name": "app", "version": "1.0.0", "dependencies": { "unix-dgram": "^2" } }' > package.json
RUN     apt update; apt install -y build-essential curl python3
RUN     curl -sL https://deb.nodesource.com/setup_22.x | bash - && apt install -y nodejs
RUN     bun install
```
It suddenly works.

I'll keep digging on my end to see if I can narrow down what in the Node installation process is necessary for this to work, but it's clear that _something_ is being configured that allows these packages to be installed...

### What is the expected behavior?

Should be able to successfully install gyp-bindings based packages using Bun (assuming you have all the build-tools available at install time)

### What do you see instead?

A requirement on Node also being present in the container in order for builds to successfully compile.

[...]

PR #16851, opened 2025-01-28
allow resolution to work when the source file does not exist

### LINKED ISSUE #16705: [nuxt] [request error] [unhandled] [500] File URL path must not include encoded / characters

1.2.0+b0c5a7655

Darwin 24.1.0 arm64 arm

After upgrading to Bun 1.2, I tried installing clean Nuxt application using this guide: [https://bun.sh/guides/ecosystem/nuxt](https://bun.sh/guides/ecosystem/nuxt). 

After installing everything, I've ran this command:
```
bun --bun run dev
```
And tried opening the application, but I was given Error 500 and also in the logs was written:
> [nuxt] [request error] [unhandled] [500] File URL path must not include encoded / characters

I tried running this command:
```
bun run dev
```
And it works normally.

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

paperclover:
Fixes #16705

## URL decoding issues

Vite handles percent encoding special characters like / and : in urls, but expects pathToFileURL to encode `%2F` as `%252F` (encoding the %). This change has been made in WebKit since i believe all places we create file URLs out of paths should have this, for example `import.meta.url` in a file named `%2F.js` (which also matches node)

## Resolver changes

Nuxt/Vite uses `createRequire("file://...")` to facilitate `require` in files. Some virtual files get a filepath `@vue/server-renderer` which gets normalized to an absolute path relative to the current working directory (`file:///Users/clo/scratch/my-nuxt-app/@vue/server-renderer`). This path, given to createRequire, produces a function that cant resolve from node_modules in Bun, but can in Node.js. The bug in Bun is when we walk up for node_modules, we use a directory cache; but the cache is not populated for non-existing directories. The change is to walk up until a cache entry exists, then to use it.

This can be reproduced in the node and bun repls, by replacing the path to your project dir.
```
require("module").createRequire("file:///Users/clo/scratch/my-nuxt-app/@vue/server-renderer")("vue")
```

Here is the code path in Node.js that causes this to successfully resolve

- `createRequire` calls `createRequireFromPath` and assigns `m.paths` to the list of all parent paths that contain a node_modules [link](<https://github.com/nodejs/node/blob/e346323109b49fa6b9a4705f4e3816fc3a30c151/lib/internal/modules/cjs/loader.js#L1934>)
- `require` calls `_resolveFilename` with the parent module (above: m) and the request
- since it does not call it with any options, this line [`paths = Module._resolveLookupPaths(request, parent);` ](https://github.com/nodejs/node/blob/e346323109b49fa6b9a4705f4e3816fc3a30c151/lib/internal/modules/cjs/loader.js#L1346)
- the paths value becomes the concatenation of `m.paths` above (in code parent.paths) and the list of NODE_PATH entries. [link](<https://github.com/nodejs/node/blob/e346323109b49fa6b9a4705f4e3816fc3a30c151/lib/internal/modules/cjs/loader.js#L930>. this is hit since the require string "vue" is not a relative file path
- `trySelf` does some native code and finds zero results. this is the directory traversal i mentioned before, which would find the nearest package.json, the import is not a self referential import.
- `_findPath` loops [through paths](https://github.com/nodejs/node/blob/e346323109b49fa6b9a4705f4e3816fc3a30c151/lib/internal/modules/cjs/loader.js#L697) from point 4 
- the module is resolved


robobun:
<div><sup>Updated 3:27 PM PT - Jan 28th, 2025</sup></div>

@paperclover, your commit 1f16852b2db7b35822f11920ee5ab953cc2859df is building: [`#10707`](https://buildkite.com/bun/bun/builds/10707)

src/resolver/resolver.zig
```diff
@@ -1338,7 +1338,16 @@ pub const Resolver = struct {
...
+            // If the source directory doesn't exist, walk up to find the first one that does, if any.
+            var source_dir_info = first_dir_info: {
+                var source_dir_to_use = source_dir;
+                while (true) {
+                    if (r.dirInfoCached(source_dir_to_use) catch return .{ .not_found = {} }) |dir|
+                        break :first_dir_info dir;
+                    source_dir_to_use = std.fs.path.dirname(source_dir) orelse
+                        return .{ .not_found = {} };
+                }
+            };
```
paperclover, line 1376: my worry on this loop is that dirInfoCached does not seem to cache not found entries, which might be worth doing in some places. on the other hand, that type of caching can lead to problems.

Jarred-Sumner: It does cache not found entries. No need to worry.

Issue #10720, opened 2024-05-01
Segfault using Discord.JS and ts3-nodejs-library


gardc:
### How can we reproduce the crash?

```ts
import { TeamSpeak, QueryProtocol } from "ts3-nodejs-library";
import { Client, TextChannel } from "discord.js";

const TS_SERVER = {
  host: "10.0.0.33",
  queryport: 10011, // Default TeamSpeak query port
  username: "username",
  password: "password",
  nickname: "something",
};

const DISCORD_TOKEN =
  "some token";
const CHANNEL_ID = "some channel id";

// Initialize the TeamSpeak client
const tsClient = new TeamSpeak({
  protocol: QueryProtocol.RAW, // This could be RAW or SSH depending on your setup
  host: TS_SERVER.host,
  queryport: TS_SERVER.queryport,
  username: TS_SERVER.username,
  password: TS_SERVER.password,
  nickname: TS_SERVER.nickname,
  serverport: 9987, // Default TeamSpeak server port
});

// Initialize the Discord client
const discordClient = new Client({ intents: ["Guilds", "GuildMessages"] });

discordClient.on("ready", () => {
  if (discordClient.user) {
    console.log(`Logged in as ${discordClient.user.tag}!`);
    setInterval(() => updateChannelNamePeriodically(), 5000); // Update every 5 seconds
  } else throw new Error("Failed to login to Discord");
});

// Function to update Discord channel name
async function updateChannelNamePeriodically() {
  const userCount = await getActiveUsersFromTeamSpeak();

  // Fetch the channel and assert the type after checking if it's text-based
  const channel = await discordClient.channels.fetch(CHANNEL_ID);
  if (channel != null && channel.isTextBased() && "setName" in channel) {
    // Checking if setName method exists
    let newName = "";
    userCount === 0
      ? (newName = "nobody-on-ts")
      : (newName = `${userCount}-on-ts`);
    // We assert channel as TextChannel since isTextBased() confirms it and we have checked for setName
    if ((channel as TextChannel).name !== newName) {
      try {
        await (channel as TextChannel).setName(newName);
        console.log(`Updated channel name to "${newName}"`);
      } catch (error) {
        console.error("Failed to update channel name:", error);
      }
    }
  }
}

// Function to fetch the number of active users on TeamSpeak
async function getActiveUsersFromTeamSpeak() {
  try {
    const server = await tsClient.serverInfo();
    return (
[...]

PR #16857, opened 2025-01-29
Drain microtasks again after deferred tasks run

Jarred-Sumner:
### What does this PR do?

Deferred tasks might call callbacks to JavaScript. We need to drain microtasks when calling into JavaScript. Therefore, we should drain microtasks after draining deferred tasks. 

Whether or not we should recursively do this is another question.

Let's see what CI says about this.

### DIFF
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -743,9 +743,10 @@ pub const DeferredTaskQueue = struct {
         return this.map.swapRemove(ctx);
     }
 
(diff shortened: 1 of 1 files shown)


robobun:
<div><sup>Updated 5:31 PM PT - Jan 28th, 2025</sup></div>

@Jarred-Sumner, your commit aa04bcc984f848b527f50ab83c48735deb7b4292 is building: [`#10720`](https://buildkite.com/bun/bun/builds/10720)

Issue #10735, opened 2024-05-01
Can't run bunx on Windows


juliomuhlbauer:
### What version of Bun is running?

1.1.12+43f0913c3

### What platform is your computer?

Microsoft Windows NT 10.0.19045.0 x64

### What steps can reproduce the bug?

1. Try to run the command: `bunx @inlang/paraglide-next init` in the terminal
2. Get the error: `error: Script not found "@inlang/paraglide-next"`

### What is the expected behavior?

To work like 'npx @inlang/paraglide-next init'

See the docs: @inlang/paraglide-next init

juliomuhlbauer:
Happens with Million too: 
![image](https://github.com/oven-sh/bun/assets/53458125/ba6d2e98-75cf-45d0-87f9-78b08f6f96bc)

Jarred-Sumner:
The bug is that `bunx` can no longer detect it's `bunx` and not `bun`. 

Not sure yet why this happens

@paperdave do you know?

paperclover:
it feels like a weird configuring, possibly involving what happens through `bun upgrade`.

for what its worth, i haven't been able to reproduce this.

i'll do some more investigation on this.

juliomuhlbauer:
I can run bun x but not bunx

PR #16859, opened 2025-01-29
Make passing an signal: AbortSignal.timeout to fetch implicitly set timeout: false

Jarred-Sumner:
### What does this PR do?

This makes passing an signal: AbortSignal.timeout to fetch implicitly set timeout: false

If you do 
```js
fetch(url, {signal: AbortSignal.timeout(60*1024*20)});
```

You expect the fetch call to timeout after 20 minutes. You don't expect it to timeout after 5 minutes because you forgot to set `timeout` to false.


robobun:
<div><sup>Updated 6:42 PM PT - Jan 28th, 2025</sup></div>

@Jarred-Sumner, your commit 7589b86f7d48983374dbf6071902469555fa2a33 is building: [`#10735`](https://buildkite.com/bun/bun/builds/10735)

PR #16863, opened 2025-01-29
Pass more node child process tests

pfgithub:
Breaking changes:

- args[0] and argv0 in Bun.spawn/spawnSync is reversed and now matches node.js
  - previously: spawnSync({cmd: ["true"], argv0: "false"}) // would execute "false"
  - now: spawnSync({cmd: ["true"], argv0: "false"}) // executes "true"

New features:

- Timeout is now supported in child_process execSync/spawnSync and Bun.spawn/spawnSync

BUN-10684


robobun:
<div><sup>Updated 8:46 PM PT - Jan 28th, 2025</sup></div>

@pfgithub, your commit 74dd15ad8e6fbd91b687bd1152e0be2b3e6dcd86 is building: [`#10752`](https://buildkite.com/bun/bun/builds/10752)

src/js/node/child_process.ts
```diff
@@ -770,7 +783,7 @@ function fork(modulePath, args = [], options) {
     // and stderr from the parent if silent isn't set.
     options.stdio = stdioStringToArray(options.silent ? "pipe" : "inherit", "ipc");
   } else if (!ArrayPrototypeIncludes.$call(options.stdio, "ipc")) {
-    throw ERR_CHILD_PROCESS_IPC_REQUIRED("options.stdio");
+    throw $ERR_CHILD_PROCESS_IPC_REQUIRED("options.stdio");
```
DonIsaac, line 783: delete old `ERR_CHILD_PROCESS_IPC_REQUIRED` JS implementation

src/bun.js/ipc.zig
```diff
@@ -210,25 +210,30 @@ const json = struct {
                 },
             }
 
-            const is_ascii = bun.strings.isAllASCII(json_data);
+            const can_be_external = bun.strings.isAllASCII(json_data) and json_data.len > 0;
```
DonIsaac: We should return an error before the `isAllASCII` check if `json_data` is empty. Before, it would cause an assertion to fail. With this change, we'll basically be trying `JSON.parse("")`

```
Welcome to Node.js v23.6.1.
Type ".help" for more information.
> JSON.parse("")
Uncaught SyntaxError: Unexpected end of JSON input
```

pfgithub: The empty string is an invalid IPC message, so it's expected that it would cause an error

src/js/node/child_process.ts
```diff
@@ -273,156 +271,123 @@ function execFile(file, args, options, callback) {
...
       child.kill(options.killSignal);
     } catch (e) {
       ex = e;
-      exitHandler();
+      exithandler();
     }
   }
 
   if (options.timeout > 0) {
     timeoutId = setTimeout(function delayedKill() {
```
Jarred-Sumner: 2 timeouts!

src/js/node/child_process.ts
```diff
@@ -273,156 +271,123 @@ function execFile(file, args, options, callback) {
...
-      "data",
-      maxBuffer === Infinity
-        ? function onUnlimitedSizeBufferedData(chunk) {
-            $arrayPush(_stdout, chunk);
-          }
-        : onData(_stdout, "stdout"),
-    );
+    child.stdout.on("data", function onChildStdout(chunk) {
+      // Do not need to count the length
+      if (options.maxBuffer === Infinity) {
```
Jarred-Sumner: Can we keep it as separate functions?

Jarred-Sumner: Or alternatively, could we move this to native code?

src/js/node/child_process.ts
pfgithub: timeout is only set on Bun.spawnSync()

pfgithub:
#18052

PR #16866, opened 2025-01-29
Add Nix Flake

RossComputerGuy:
### What does this PR do?

Changes the vendoring of sources to allow for building inside of Nix. Adds a nix flake.

- [x] Code changes
- [x] Build system changes

Run `nom build` / `nix build`. I've tried it but it gets stuck half way through bun's build process on Apple M1 Pro.

- [x] I made sure that specific versions of dependencies are used instead of ranged or tagged versions


paperclover:
too high maintenance burden (new files for deps, package-lock.json, etc), ignoring mistakes in the actual build. though if this works on your machine, congrats to being the first to build bun with the nix build system, i am impressed (many have tried and gave up).

most of the team uses macOS / homebrew.

Eveeifyeve:
>too high maintenance burden (new files for deps, package-lock.json, etc), ignoring mistakes in the actual build. though if this works on your machine, congrats to being the first to build bun with the nix build system, i am impressed (many have tried and gave up).
> most of the team uses macOS / homebrew.

Okay I have some disagreements, 1. Updating dependancies can be community ran plus there is a github action by determined to update flake lockm, two you can run checks on a flake and do a devshell if it fails then you can add the dependancy in the flake really easily, three backwards compatiblity with older nixos installations could be done with a shell.nix as the nix/nixos community that use bun is big and has a potential. Homebrew will not work for nixos as it's linux and it doesn't accept dymanic linked libraries it is possible through nix-std but that is what nixos is not made for and in a sandboxed environment homebrew could potentially not work when using nix on darwin because it's a dynamic linked library again.

Eveeifyeve:
Btw for backwards compatibility with shell.nix is through flake-compact.

flake.nix
```diff
@@ -0,0 +1,390 @@
+{
+  description = "Incredibly fast JavaScript runtime, bundler, test runner, and package manager – all in one";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs";
```
robinanterior, line 5: Unless I'm mistaking this uses the default branch on the target repo, which for nixos/nixpkgs is `master`, which isn't necessarily built by hydra. You usually want to default to `github:nixos/nixpkgs/nixpkgs-unstable`, which follows `master` but gets advanced only when hydra has finished building ~all derivations and uploaded them to the binary cache.

(I know the PR is closed and obsolete but could still be worth it for the next person picking this up)

Eveeifyeve:
> too high maintenance burden (new files for deps, package-lock.json, etc), ignoring mistakes in the actual build. though if this works on your machine, congrats to being the first to build bun with the nix build system, i am impressed (many have tried and gave up).
> 
> most of the team uses macOS / homebrew.

for future reference for anyone wants a shell to use since bun, I will be making a unofficial flake for now at https://github.com/eveeifyeve/bun-flake until bun if they do change their decision.

Issue #10765, opened 2024-05-02
postinstall script from "full-icu" exited with 1


baphony:
### What version of Bun is running?

1.1.6+e58d67b46

### What platform is your computer?

Linux 5.15.146.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

bun i

bunfig.toml:
```
[install]
registry = { url = "https://...", token = "..." }
# Avoid patches to be applied on the Bun cache files
backend = "copyfile"

[install.lockfile]
print = "yarn"
```

### What is the expected behavior?

Not sure what's the status of Bun for internationalization but there is what NodeJS is doing:
https://github.com/nodejs/node/pull/3102/files

### What do you see instead?

```shell
bun install v1.1.6 (e58d67b4)
  🚚 jss... 53 |   } else {
54 |     icuver = process.config.variables.icu_ver_major // only get the major
55 |   }
56 |
57 |   if (!icuver) {
58 |     throw Error('Cannot determine Node’s ICU version!')
               ^
error: Cannot determine Node’s ICU version!
      at /home/bap/work/symphony/sfe-rtc/node_modules/full-icu/full-icu.js:58:11
      at /home/bap/work/symphony/sfe-rtc/node_modules/full-icu/postinstall.js:5:7

error: postinstall script from "full-icu" exited with 1
```

### Additional information

I am using Ubuntu in WSL2

node_modules/full-icu/full-icu.js:58:11
```
  if (process.versions.icu) {
    icuver = process.versions.icu // Thanks https://github.com/nodejs/node/issues/3089
  } else {
    icuver = process.config.variables.icu_ver_major // only get the major
  }

  if (!icuver) {
    throw Error('Cannot determine Node’s ICU version!')
  }
```
https://www.npmjs.com/package/full-icu

dylan-conway:
We are missing `icu` (and `unicode`) on `process.versions`. This will be fixed when #10760 merges

PR #16867, opened 2025-01-29
meta: update bun.locks with bun 1.2

nektro:


robobun:
<div><sup>Updated 10:54 PM PT - Jan 28th, 2025</sup></div>

@nektro, your commit 1670ac6ef7d43a12567be067cde309856dfb4dd0 is building: [`#10765`](https://buildkite.com/bun/bun/builds/10765)

nektro:
running `[build images]` wasnt actually necessary since the load bearing part of these runs all took place on the `build-zig` queue which needed an out-of-band run of `bun upgrade` on the machines.

PR #16881, opened 2025-01-29
install.sh: support busybox unzip, 7z, 7zz, 7za, bsdtar, and same in `bun upgrade`

erg:
see https://github.com/denoland/deno_install/pull/321 as well

### What does this PR do?

This makes installing bun easier on systems where you don't have root. We want to unzip a .zip file, so besides unzip there are `busybox unzip`, `7z/7zz/7za`, and `bsdtar` which might be installed.

We need to do the same thing for the `bun upgrade` command. `unzip -o` preserves xattrs, the other commands might not.

- [ x ] Code changes

Test with (updated command):
```sh
curl -fsSL https://raw.githubusercontent.com/erg/bun/c8aa12adb5d94437c431ff24ce80e8206194d01d/src/cli/install.sh | bash
```


paperclover (changes requested):
i have a 24.10 vm via orbstack and it doesnt seem to have busybox or the others, however i still think this is a reasonable change

can the message be changed to 
"unzip is required to install Bun (7z and busybox supported)"
to make it so that someone reading extremely quickly sees that it is some unzipping tool, and then that other options work. excluding 7zz because it appears to be the same. and while were at it we can capitalize Bun in this message.

erg:
Changed the message, it's testing

RiskyMH:
I think `bun update` may need to be updated so people can still update 

https://github.com/oven-sh/bun/blob/29839737dffa92b63dd96619bc9079fe14e9849b/src/cli/upgrade_command.zig#L610

erg:
Alright, I'll do that soon.

erg:
I updated the PR to unzip from the zig side. I believe all of the methods preserve xattrs for the code-signing.

- `unzip`/busybox due to `-o`
- lz/lzz/lza should preserve them (?)
- bsdtar due to `-xattrs`

salonalmas (commented):
https://github.com/denoland/deno_install/issues/328

paperclover (commented):
this branch is stale. i am leaving this review to remove this PR from my "requested reviews" page, since GitHub does not actually have a way to dismiss a review request.

erg:
Still hitting this in 2026. Spend 10s to have claude fix this. 🤖 

```
erg@computer:~$ curl -fsSL https://bun.com/install | bash
error: unzip is required to install bun
erg@computer:~$ bun upgrade
Bun v1.3.9 is out! You're on v1.2.1
error: Failed to locate "unzip" in PATH. bun upgrade needs "unzip" to work.
erg@computer~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 25.10
Release:        25.10
Codename:       questing
erg@computer:~$
```

I have `busybox unzip` and editing the install script by hand worked.

robobun:
Thanks for this, and sorry it sat for so long. #43643 now implements the same fallbacks for both `install.sh` and `bun upgrade`. It probes `busybox unzip`, `7z`, `7zz`, `7za` and `bsdtar`, plus `python3`. It also downloads with `wget` when `curl` is missing. The PR is written against the current code and has tests. I ran each of those extractors against it. That includes a real BusyBox 1.36.1 build, because `busybox unzip` is what you have.

As the review here asked, the error text names the alternatives: `unzip is required to install bun (7z, busybox, bsdtar, python3 supported)`. #43643 also takes one more idea from this PR: when extraction fails, `bun upgrade` names the extractor that ran. That commit credits you as co-author.

This branch no longer applies because the script moved to `src/runtime/cli/` and `bun upgrade` was rewritten in Rust. Closing in favor of #43643, which lists this PR as prior work. Please comment there or reopen if it misses a case you need.

Issue #10795, opened 2024-05-03
bun install crashed


wan54:
### How can we reproduce the crash?

Sorry, I'm not sure if it's reproducible. But looks like I can't use bun on this machine.

`uname -a` outputs

```
Darwin home.local 21.6.0 Darwin Kernel Version 21.6.0: Mon Feb 19 20:24:34 PST 2024; root:xnu-8020.240.18.707.4~1/RELEASE_X86_64 x86_64
```

### Relevant log output

```shell
bun install v1.1.6 (e58d67b4)
  🔍 Resolving [1/1] ============================================================
Bun v1.1.6 (e58d67b4) macOS x64 (baseline)
Args: "bun", "i"
Elapsed: 177ms | User: 45ms | Sys: 22ms
RSS: 18.46MB | Peak: 18.46MB | Commit: 1.07GB | Faults: 0

panic: Illegal instruction at address 0x107CC76FD
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
```

### Stack Trace (bun.report)

Bun v1.1.6 ([`e58d67b`](https://github.com/oven-sh/bun/tree/e58d67b46875280bc9002add4c3fae45472cb346)) on macos x86_64_baseline [InstallCommand]

Illegal instruction at address 0x109D146FD

- *2 unknown/js code*
- `src.install.install.PackageManager.enqueueDependencyToRoot__struct_445492.issue_19586_workaround`
- `src.shell.interpreter.Interpreter.Binary.makeChild`
- `uWS::WebSocketContext<false, true, void*>::handleFragment`
- `auto uWS::WebSocketContext<true, true, void*>::init(...)::__invoke<us_socket_t>`
- `auto uWS::WebSocketContext<true, true, void*>::init(...)::__invoke<us_socket_t>`
- `src.bun.js.javascript.VirtualMachine.printErrorInstance__anon_414869`
- `src.js_printer.NewPrinter`
- *1 unknown/js code*

Features: dotenv

wan54:
Additional info:

`bun init` also hangs for me. No errors.

```
➜  buntest bun init
bun init helps you get started with a minimal project and tries to guess sensible defaults. Press ^C anytime to quit

package name (buntest):
entry point (index.ts):

Done! A package.json file was saved in the current directory.
 + index.ts
 + .gitignore
 + tsconfig.json (for editor auto-complete)
 + README.md

To get started, run:
  bun run index.ts
```

```
➜  buntest bun -v
1.1.7
[...]

PR #16885, opened 2025-01-29
Correctly handle `__esModule` for `loader: "object"`

Jarred-Sumner:
### What does this PR do?

Correctly handle `__esModule` for `loader: "object"`

Unlike CJS normally, we were ignoring the `__esModule` property.

There is a test that fails in previous versions of Bun and passes in this version


robobun:
<div><sup>Updated 3:37 PM PT - Jan 29th, 2025</sup></div>

@Jarred-Sumner, your commit b47a1f879c11b71243a9d2be6303812ebf8d5053 is building: [`#10795`](https://buildkite.com/bun/bun/builds/10795)

PR #16886, opened 2025-01-30
Return default export when virtual module with `__esModule: true` is loaded via `require()`

thecrypticace:
### What does this PR do?

Returns the default export of a virtual module loaded through `require()` when `__esModule: true` is set on the exports object.

- [x] Code changes

- [x] I ran my tests locally and they pass (`bun-debug test plugins.test`)

I also manually verified that these changes work in a separate real-world scenario.

h/t to @Jarred-Sumner for the help with the initial code and direction


thecrypticace:
oops didn't realize you'd already opened a PR to fix this: #16885

Will close this one

Issue #16781, opened 2025-01-26
S3 stream upload crashes Bun


RainyPixel:
### How can we reproduce the crash?

This code crash bun :)

```ts
const s3Client = new S3Client({
  accessKeyId: process.env.S3_REPOSITORY_CACHE_ACCESS_KEY_ID!,
  secretAccessKey: process.env.S3_REPOSITORY_CACHE_SECRET_ACCESS_KEY!,
  endpoint: process.env.S3_REPOSITORY_CACHE_ENDPOINT!,
  bucket: process.env.S3_REPOSITORY_CACHE_BUCKET_NAME!,
});

const uploadToS3 = async (filePath: string, s3Key: string) => {
  const s3File = s3Client.file(s3Key, { type: "application/x-tarzstd" });
  const writer = s3File.writer({
    partSize: 5 * 1024 * 1024,
    queueSize: 10, 
    retry: 3,
  });

  const fileToStream = Bun.file(filePath);
  const fileStream = fileToStream.stream();

  for await (const chunk of fileStream) {
    writer.write(chunk);
  }

  await writer.end();
};
```

### Relevant log output

```shell

```

### Stack Trace (bun.report)

Bun v1.1.45 ([`196621f`](<https://github.com/oven-sh/bun/tree/196621f253cf318b9096d2b8335b74e0b0e98d6c>)) on windows x86_64 [AutoCommand]

Segmentation fault at address 0xFFFFFFFFFFFFFFFF

- [`multipart.zig:100`](<https://github.com/oven-sh/bun/blob/b0c5a7655dac9040655a22754c5eab91046e923b/src/s3/multipart.zig#L100>): `onPartResponse`
- [`simple_request.zig:215`](<https://github.com/oven-sh/bun/blob/b0c5a7655dac9040655a22754c5eab91046e923b/src/s3/simple_request.zig#L215>): `onResponse`
- [`event_loop.zig:1020`](<https://github.com/oven-sh/bun/blob/b0c5a7655dac9040655a22754c5eab91046e923b/src/bun.js/event_loop.zig#L1020>): `tickQueueWithCount`
- [`event_loop.zig:1325`](<https://github.com/oven-sh/bun/blob/b0c5a7655dac9040655a22754c5eab91046e923b/src/bun.js/event_loop.zig#L1325>): `tickConcurrent`
[...]

Issue #10806, opened 2024-05-03
Cannot find module on windows bun


cacacaio:
### What version of Bun is running?

1.1.7+b0b7db5c0

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

I just went into a expo project and used `bun install` and then went ahead and bun start to run the expo start script on package.json and it gave me this error.

### What is the expected behavior?

The project ran as normal.

### What do you see instead?

```
Error: Cannot find module 'ajv/dist/compile/codegen'
Require stack:
- C:\Users\REDACTED\node_modules\ajv-keywords\dist\definitions\typeof.js
- C:\Users\REDACTED\node_modules\ajv-keywords\dist\keywords\typeof.js
- C:\Users\REDACTED\node_modules\ajv-keywords\dist\keywords\index.js
- C:\Users\REDACTED\node_modules\ajv-keywords\dist\index.js
- C:\Users\REDACTED\node_modules\schema-utils\dist\validate.js
- C:\Users\REDACTED\node_modules\schema-utils\dist\index.js
- C:\Users\REDACTED\node_modules\expo-router\plugin\build\index.js
- C:\Users\REDACTED\node_modules\expo-router\app.plugin.js
- C:\Users\REDACTED\node_modules\@expo\config-plugins\build\utils\plugin-resolver.js
- C:\Users\REDACTED\node_modules\@expo\config-plugins\build\plugins\withStaticPlugin.js
- C:\Users\REDACTED\node_modules\@expo\config-plugins\build\plugins\withPlugins.js
- C:\Users\REDACTED\node_modules\@expo\config-plugins\build\index.js
- C:\Users\REDACTED\node_modules\@expo\config\build\plugins\withConfigPlugins.js
- C:\Users\REDACTED\node_modules\@expo\config\build\Config.js
- C:\Users\REDACTED\node_modules\@expo\config\build\index.js
- C:\Users\REDACTED\node_modules\@expo\cli\build\src\utils\analytics\getDevClientProperties.js
- C:\Users\REDACTED\node_modules\@expo\cli\build\src\start\resolveOptions.js
- C:\Users\REDACTED\node_modules\@expo\cli\build\src\start\index.js
- C:\Users\REDACTED\node_modules\@expo\cli\build\bin\cli
- C:\Users\REDACTED\node_modules\expo\bin\cli
Error: Cannot find module 'ajv/dist/compile/codegen'
Require stack:
[...]

PR #16890, opened 2025-01-30
fix(s3) fix queue and multipart flow

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/16781

- [X] Code changes

Tests

### DIFF
--- a/src/s3/multipart.zig
+++ b/src/s3/multipart.zig
@@ -260,7 +260,6 @@ pub const MultiPartUpload = struct {
             switch (state) {
                 .pending => {
                     this.freeAllocatedSlice();
-                    this.ctx.deref();
                 },
                 // if is not pending we will free later or is already freed
                 else => {},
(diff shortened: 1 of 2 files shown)


robobun:
<div><sup>Updated 6:23 PM PT - Jan 29th, 2025</sup></div>

@cirospaciari, your commit f3ed65b97cfe43bd8276a8ec22cdd22a4eea9b6d is building: [`#10806`](https://buildkite.com/bun/bun/builds/10806)

Jarred-Sumner (changes requested):
If you want to pre-allocate, then this should be a slice and not an ArrayList.

Keeping it as an ArrayList is a footgun.

Jarred-Sumner (changes requested):
Needs more tests. There should be tests for different queue sizes, for different amounts of writes, for writes that exceed the queue size, for writes that never exceed the queue size.

PR #16894, opened 2025-01-30
fix: `Bun.deepEquals` on empty objects with the same prototype

DonIsaac:
### What does this PR do?

Fixes this issue that was originally found in https://github.com/nodejs/node/issues/10258. Needed for `node:assert` compatibility.

```js
    function FakeDate() {}
    FakeDate.prototype = Date.prototype;
    const a = new Date("2016");
    const b = new FakeDate();
    expect(deepEquals(a, b)).toBe(false); // yay
    expect(deepEquals(b, a)).toBe(false); // :(
```

### How did you verify your code works?
There are tests

Issue #10820, opened 2024-05-04
Bun crashes when trying to install any package using ``bun install`` command.


YoussefAhmedDev:
### Stack Trace (bun.report)

Bun v1.1.7 ([`b0b7db5`](https://github.com/oven-sh/bun/tree/b0b7db5c060e830801b0273b7ff6e34e3eb5c254)) on windows x86_64_baseline [InstallCommand]

Illegal instruction at address 0x7FF7D672FA41

- [`libtcc.c:188`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/deps/tinycc/libtcc.c#L188): `pstrcpy`
- [`tccgen.c:5018`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/deps/tinycc/tccgen.c#L5018): `post_type`
- [`logger.zig:993`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/logger.zig#L993): `addRangeErrorFmt`
- [`semver.zig:268`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/install/semver.zig#L268): `slice`
- [`math.zig:452`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/deps/zig/lib/std/math.zig#L452): `sub`
- [`bun.zig:1754`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/bun.zig#L1754): `insert`
- ??? at `0x17344` in KERNEL32.DLL
- ??? at `0x526b1` in ntdll.dll

nektro:
if you upgrade to canary with `bun upgrade --canary` does the issue go away? we just recently landed a fix that should address this

PR #16895, opened 2025-01-30
fix(node/assert): port more test cases from node

DonIsaac:
### What does this PR do?
Ports most remaining `node:assert` unit tests from node.
- [x] `test-assert-calltracker-getCalls`
- [x] `test-assert-calltracker-calls`
- [x] `test-assert-calltracker-report`
- [x] `test-assert-calltracker-verify`
- [x] `test-assert-checktag`
- [x] `test-assert-esm-cjs-message-verify`
- [x] `test-assert-fail`
- [x] `test-assert-if-error`
- [x] `test-assert-deep-with-error`

Tests that still need to be ported are:
- `test-assert-async`
- `test-assert-deep`
- `test-assert-fail-deprecation`
- `test-assert-first-line`
- `test-assert-objects`
- `test-assert-typedarray-deepequal`

These require a good deal of refactoring and will be added in another PR.

There are tests.


robobun:
<div><sup>Updated 9:36 PM PT - Jan 29th, 2025</sup></div>

@DonIsaac, your commit 6a4653070fbf2e62b5bd8c5350ed242d5e1f0634 is building: [`#10820`](https://buildkite.com/bun/bun/builds/10820)

DonIsaac:
`test-assert-checktag` needs #16894 to land.

src/bun.js/bindings/bindings.zig
```diff
@@ -5630,38 +5630,49 @@ pub const JSValue = enum(i64) {
...
     /// Object.is()
+    ///
     /// This algorithm differs from the IsStrictlyEqual Algorithm by treating all NaN values as equivalent and by differentiating +0𝔽 from -0𝔽.
     /// https://tc39.es/ecma262/#sec-samevalue
     pub fn isSameValue(this: JSValue, other: JSValue, global: *JSGlobalObject) bool {
         return @intFromEnum(this) == @intFromEnum(other) or cppFn("isSameValue", .{ this, other, global });
     }
 
+    /// NOTE: can throw. Check for exceptions.
     pub fn deepEquals(this: JSValue, other: JSValue, global: *JSGlobalObject) bool {
```
dylan-conway: doesn't have to happen now, but these can be changed to use `bun.JSError!bool`

dylan-conway (approved):

dylan-conway (changes requested):
Inside `Bun__deepEquals` when we are checking `JSMapType` equality we should have a `RETURN_IF_EXCEPTION` check after each `Bun__deepEquals`. Same with `JSSetType`. The other deep equals calls should be okay as they are because they immediately return on false

dylan-conway:
Ah, that's now inside `specialObjectsDequal`

src/bun.js/bindings/bindings.cpp
```diff
@@ -1371,7 +1375,11 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
     compareAsNormalValue:
         break;
     }
-
+    // globalThis is only equal to globalThis
+    // NOTE: Zig::GlobalObject is tagged as GlobalProxyType
+    case GlobalObjectType:
+    case GlobalProxyType:
+        return c1Type == c2Type;
```
Jarred-Sumner: This is not correct. A node:vm JSGlobalObject/globalThis could be compared to here. It should check if the values are strict equal for that case.

src/bun.js/bindings/bindings.cpp
```diff
@@ -2462,52 +2441,37 @@ bool JSC__JSValue__isSameValue(JSC__JSValue JSValue0, JSC__JSValue JSValue1,
     return JSC::sameValue(globalObject, left, right);
 }
 
+#define IMPL_DEEP_EQUALS_WRAPPER(strict, enableAsymmetricMatchers, globalObject, a, b) \
```
Jarred-Sumner: For future reference, you can use a `static inline bool` here and it'll compile to almost the same thing

src/js/internal/primordials.js
```diff
@@ -129,6 +129,14 @@ export default {
       }
     },
   ),
+  SafeWeakMap: makeSafe(
```
Jarred-Sumner, line 132: What is using this?

src/bun.js/bindings/bindings.cpp
DonIsaac: fixed

src/js/internal/primordials.js
DonIsaac: [Calltracker](https://github.com/oven-sh/bun/blob/1ccc13ecf79861eb0b5a697c0b5571772d42a28d/src/js/internal/assert/calltracker.ts#L61)

src/bun.js/bindings/bindings.cpp
DonIsaac: I think I've fixed it, but I'm not sure. Will confirm and update afterwards.

DonIsaac:
It looks like `Buffer` instances use the same structure, class info, and JSType as the typed array they're constructed from. Since changing that is in a fully separate part of the codebase, I'll make that change in a follow-up pr.

src/bun.js/bindings/bindings.cpp
```diff
@@ -1185,16 +1191,117 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
...
+                    return false;
+                }
+            }
+
+            VM& vm = globalObject->vm();
+
+            // `.cause` is non-enumerable, so it must be checked explicitly.
+            // note that an undefined cause is different than a missing cause in
+            // strict mode.
+            const Identifier causeIdent = Identifier::fromLatin1(vm, "cause");
```
dylan-conway:
```suggestion
            const Identifier causeIdent = Identifier::fromString(vm, "cause"_s);
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -1185,16 +1191,117 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
...
+                if (propertyArrayLength1 != propertyArrayLength2) {
+                    return false;
+                }
+            }
+
+            // take a property name from one, try to get it from both
+            size_t i;
+            for (i = 0; i < propertyArrayLength1; i++) {
+                Identifier i1 = a1[i];
+                if (!i1.isEmpty() && !i1.isSymbol() && i1.string() == "stack"_s) continue;
```
dylan-conway:
```suggestion
                if (!i1.isEmpty() && !i1.isSymbol() && i1 == "stack"_s) continue;
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -1185,16 +1191,117 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
...
+            if (!Bun__deepEquals<isStrict, enableAsymmetricMatchers>(globalObject, leftCause, rightCause, gcBuffer, stack, scope, true)) {
+                return false;
+            }
+            RETURN_IF_EXCEPTION(*scope, false);
+
+            // check arbitrary enumerable properties. `.stack` is not checked.
+            left->materializeErrorInfoIfNeeded(vm);
+            right->materializeErrorInfoIfNeeded(vm);
+            // JSObject* o1 = c1->getObject();
+            // JSObject* o2 = c2->getObject();
```
dylan-conway:
```suggestion
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -1185,16 +1191,117 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
...
+                if (!Bun__deepEquals<isStrict, enableAsymmetricMatchers>(globalObject, prop1, prop2, gcBuffer, stack, scope, true)) {
+                    return false;
+                }
+
+                RETURN_IF_EXCEPTION(*scope, false);
+            }
+
+            // for the remaining properties in the other object, make sure they are undefined
+            for (; i < propertyArrayLength2; i++) {
+                Identifier i2 = a2[i];
```
dylan-conway, line 1290: do we also need to check for `stack` here?

src/bun.js/bindings/bindings.cpp
```diff
@@ -1185,16 +1191,117 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
...
+            const Identifier causeIdent = Identifier::fromLatin1(vm, "cause");
+            const PropertyName cause(causeIdent);
+            if constexpr (isStrict) {
+                if (left->hasProperty(globalObject, cause) != right->hasProperty(globalObject, cause)) {
+                    return false;
+                }
+            }
+            auto leftCause = left->get(globalObject, cause);
+            auto rightCause = right->get(globalObject, cause);
+            if (!Bun__deepEquals<isStrict, enableAsymmetricMatchers>(globalObject, leftCause, rightCause, gcBuffer, stack, scope, true)) {
```
dylan-conway:
```suggestion
            auto leftCause = left->get(globalObject, cause);
            RETURN_IF_EXCEPTION(*scope, false);
            auto rightCause = right->get(globalObject, cause);
            RETURN_IF_EXCEPTION(*scope, false);
            if (!Bun__deepEquals<isStrict, enableAsymmetricMatchers>(globalObject, leftCause, rightCause, gcBuffer, stack, scope, true)) {
```

this might be missing above too

src/bun.js/bindings/bindings.cpp
```diff
@@ -1371,7 +1478,20 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
     compareAsNormalValue:
         break;
     }
-
+    // globalThis is only equal to globalThis
+    // NOTE: Zig::GlobalObject is tagged as GlobalProxyType
+    case GlobalObjectType: {
+        if (c1Type != c2Type) return false;
+        auto* g1 = jsDynamicCast<JSC::JSGlobalObject*, JSCell>(c1);
+        auto* g2 = jsDynamicCast<JSC::JSGlobalObject*, JSCell>(c2);
```
dylan-conway, line 1491: not sure if it's possible, but let's add

```suggestion
        auto* g2 = jsDynamicCast<JSC::JSGlobalObject*, JSCell>(c2);
        if (!g1 || !g2) {
            return !!g1 == !!g2;
        }
```

src/bun.js/bindings/bindings.cpp
```diff
@@ -1371,7 +1478,20 @@ std::optional<bool> specialObjectsDequal(JSC__JSGlobalObject* globalObject, Mark
...
+        return g1->m_globalThis == g2->m_globalThis;
+    }
+    case GlobalProxyType: {
+        if (c1Type != c2Type) return false;
+        auto* gp1 = jsDynamicCast<JSC::JSGlobalProxy*, JSCell>(c1);
+        auto* gp2 = jsDynamicCast<JSC::JSGlobalProxy*, JSCell>(c2);
```
dylan-conway, line 1497: same here

dylan-conway (changes requested):
Left a few suggestions.

nit: the object equality code shouldn't be duplicated. Might be best to split it into template function with `bool skipStackProp`, but that can happen in a future pr

src/bun.js/bindings/bindings.cpp
Jarred-Sumner: vm.propertyNames.stack can be used here instead.

src/bun.js/bindings/bindings.cpp
Jarred-Sumner: vm.propertyNames.cause

src/bun.js/bindings/bindings.cpp
DonIsaac: GlobalProxyType may also be cast into a `JSC::JSGlobalObject`, so we still need to check if `JSTypes` are equal

src/bun.js/bindings/bindings.cpp
DonIsaac: reply is above

src/bun.js/bindings/bindings.cpp
DonIsaac: fixed

src/bun.js/bindings/bindings.cpp
DonIsaac: fixed. used `vm.propertyNames->stack`.

src/bun.js/bindings/bindings.cpp
Jarred-Sumner: they're different types

src/bun.js/bindings/bindings.cpp
Jarred-Sumner: if its a JSGlobalProxy, then `target` should be checked to see if it matches the JSGlobalObject

dylan-conway (approved):

Issue #16779, opened 2025-01-26
Astro endpoints cannot import from Bun


spiffytech:
### What version of Bun is running?

1.2.0+b0c5a7655

### What platform is your computer?

Linux 6.12.9-200.fc41.x86_64 x86_64 unknown

### What steps can reproduce the bug?

1. `bun create astro@latest`
2. Write the following into `src/pages/my-endpoint.json.ts`:
```ts
import {$, SQL} from 'bun';

export const GET = () => {
  console.log("$", $);
  console.log("Bun.$", Bun.$);
  console.log("SQL", SQL);
  return new Response("Ok");
}

```
3. Call the endpoint: `curl http://localhost:4322/my-endpoint.json`
4. Observe the server output:
```
$ undefined
Bun.$ [Function: BunShell]
SQL undefined
16:13:04 [200] /my-endpoint.json 1ms
```

### What is the expected behavior?

The imports work.

DonIsaac:
<img width="618" alt="Image" src="https://github.com/user-attachments/assets/83fa4998-4c8c-46dc-a5e4-f95661564502" />

DonIsaac:
@cirospaciari `var SQL: SQL` is an erroneous definition in `@bun/types`

spiffytech:
1. Both `sql` and `SQL` are noted in [the Bun docs](https://bun.sh/docs/api/sql), so if there's a mistake that should be updated too.
2. The Bun + Astro bug affects `SQL`, `sql`, and all other Bun imports I've tried:
```ts
import {SQL, sql, $, s3, Glob} from 'bun';

export const GET = () => {
  console.log(SQL, sql, $, s3, Glob);
  return new Response("Ok");
}
```

```
$ bun --bun run dev 
$ astro dev --port "4321"
08:45:39 [types] Generated 0ms
08:45:39 [content] Syncing content
08:45:39 [content] Synced content

 astro  v5.1.9 ready in 71 ms

┃ Local    http://localhost:4321/
┃ Network  use --host to expose

08:45:39 watching for file changes...
undefined undefined undefined undefined undefined
08:45:40 [200] /my-endpoint.json 4ms
```

Jarred-Sumner:
It’s really hard to see how this is not a bug in Vite and/or Astro. We aren’t removing the Bun global. You’ll note that it works without Astro. 

I guess it could be a node compatibility bug in the module loader _maybe_. Do any of the Bun object properties work? Do other things work in older versions of Bun?

[...]

Issue #16775, opened 2025-01-26
Bun + Vite SSR + React Router producing undefined for S3Client symbol


jdmcd:
### What version of Bun is running?

1.2.0-canary.3+4641b6184

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

Attached is a repo, you can run it with `bun run dev` and see that loading the `/` route produces the error. Or, follow these steps:

1. `bunx create-react-router@latest my-react-router-app`
2. Adjust `package.json` scripts to call `bunx --bun` instead of `react-router` directly (i.e. `react-router dev` -> `bunx --bun react-router dev`)
3. `bun install -d @types/bun`
4. Create `utils.server.ts` and export an `S3Client` instance from it 
5. In the `loader` function in `home.tsx`, call `presign` (or any method) on the exported `S3Client` object 

[bun-vite-repro.zip](https://github.com/user-attachments/files/18550713/bun-vite-repro.zip)

### What is the expected behavior?

Bun + Vite SSR + React Router should probably bundle in dev mode so that bun types + functionality can be used 

### What do you see instead?

```
10:55:07 AM [vite] Error when evaluating SSR module /app/utils.server.ts:
|- TypeError: undefined is not a constructor (evaluating 'new __vite_ssr_import_0__.S3Client({
  region: "us-east-1"
})')
    at /Users/jamesmcdermott/Developer/rubie-bun-vite/app/utils.server.ts:5:17
    at processTicksAndRejections (native)
```

Vite is turning the import from `bun` in `app.utils.ts` to `undefined`/`__vite_ssr_import_0__`

### Additional information

I have tried adding `bun` to Vite's `external` array under `ssr` but the behavior does not change.

ozitrance:
Same is happening with the new SQL module:

```
9:57:18 AM [vite] (ssr) Error when evaluating SSR module app/server/index.ts:
|- TypeError: undefined is not a constructor (evaluating 'new __vite_ssr_import_1__.SQL("postgres://postgres:postgres@localhost/bndb")')
    at /home/custom-mount/app/server/context.ts:5:11
    at processTicksAndRejections (native)
```

kravetsone:
Seems like u run it by node

jdmcd:
@kravetsone no it's running via bun (see step #2)

kravetsone:
> @kravetsone no it's running via bun (see step #2) 

did u try just typeof Bun

[...]

PR #16938, opened 2025-01-31
Fix import("bun") in Vite

Jarred-Sumner:
### What does this PR do?

Previously, if you did something like this:

```js
const Bun = await import(eval('"bun"'));
```

The `Bun` object here would behave completely different than what it would do in a statement. At build time, Bun converts import statements from`"bun"` to `globalThis.Bun`. This allows us to lazily load all the properties on the Bun object.

In Vite 6, they made it skip going through Bun's transpiler and made the imports not statically analyzable. This causes Bun's transpiler to not run, which broke importing from the Bun object.

Fixes https://github.com/oven-sh/bun/issues/16775
Fixes https://github.com/oven-sh/bun/issues/16779

There is a test


robobun:
<div><sup>Updated 5:11 AM PT - Jan 31st, 2025</sup></div>

@Jarred-Sumner, your commit 3c4592b3dd552212e17d6c0f8391c89b26e5c626 is building: [`#10916`](https://buildkite.com/bun/bun/builds/10916)

Issue #8947, opened 2024-02-16
@aws-sdk/client-s3 `ListObjectsV2Command` hangs forever when listing many files (>800)


moreirathomas:
### What version of Bun is running?

1.0.26+c75e768a6

### What platform is your computer?

Darwin 23.3.0 arm64 arm

### What steps can reproduce the bug?

Sending `ListObjectsCommand` (or `ListObjectsV2Command`) via `S3Client.send` hangs forever at runtime.

```ts
const client = new S3Client({
  region,
  credentials: { accessKeyId, secretAccessKey },
});

const command = new ListObjectsCommand({
  Bucket: bucket,
  Prefix: prefix,
  Delimiter: "/",
});

const res = await client.send(command); // <- hangs forever
```

### What is the expected behavior?

Do not hang forever. At least fail and return an error or throw an exception.

### What do you see instead?

When testing: the test fail due to timeout.

At runtime as an http handler and called via curl: the request hangs forever (until ctrl+c).

### Additional information

I have two reproduction repositories available: one running on Bun (where the bug occurs), on running on Deno (no bug reproduced).

- https://github.com/moreirathomas/aws-s3api-hangs-reproduction-bun
- https://github.com/moreirathomas/aws-s3api-hangs-reproduction-deno

Tested with @aws-sdk/client-s3 at 3.449.0 and 3.511.0.

Electroid:
Thanks for reporting, do you happen to know if it works with smaller output sizes? or does it always hang

moreirathomas:
It does not hang with a smaller output size (either fewer objects to be returned by S3 or `pageSize` is set to a smaller value, as shown in the reproduction code).

I have a S3 bucket with 2 folders, the first has 2 objects (does not hang), the second has 884 objects (hangs).

truongan07:
RAM has increased abnormally high ?

tusharlock10:
I am facing the same issue, bun version 1.0.28

```
const command = new ListObjectsV2Command({ Bucket: Env.AWS_S3_BUCKET_NAME });
const response = await client.send(command);
```

However in my case it hangs on the first line itself, it doesn't even reach the client.send command.

tschersach:
For me it's all related to something like #7428 where we have similar reports. For all aws-sdk a lot of hanging is expected.

computerdane:
[...]

PR #16948, opened 2025-01-31
feat(s3Client): add support for ListObjectsV2 action

### LINKED ISSUE #16495: s3 list function

There's a function on the AWS S3 NPM package that allows you to list all of the items in a directory/path. Bun version doesn't have this.

there should be an s3 list function to see a list of objects in a path

_No response_

Inqnuam:
### What does this PR do?

This PR adds support for S3 ListObjectsV2 action.
Same input/output as `@aws-sdk/s3-client`.

Closes #16495
May be helpful for #8947

- [x] Code changes

 I wrote automated tests
- [x] I ran my tests locally and they pass (`bun-debug test test/js/bun/s3/s3-list-objects.test.ts`)
- [x] I ran my tests on AWS and they pass

 If Zig files changed:

- [x] I wrote TypeScript/JavaScript tests and they pass locally (`bun-debug test test/js/bun/s3/s3-list-objects.test.ts`)


zshzebra:
Hopefully a maintainer can get to this soon. I would love to see this added.

packages/bun-types/bun.d.ts
```diff
@@ -1813,6 +1813,106 @@ declare module "bun" {
    *     const url = bucket.presign("file.pdf");
    *     await bucket.unlink("old.txt");
    */
+  interface S3ListObjectsOptions {
+    /** Limits the response to keys that begin with the specified prefix. */
+    Prefix?: string;
```
cirospaciari: The names of fields in the types should start in lower cased.

packages/bun-types/bun.d.ts
```diff
@@ -1946,6 +2046,14 @@ declare module "bun" {
     unlink(path: string, options?: S3Options): Promise<void>;
     delete: S3Client["unlink"];
 
+    /** Returns some or all (up to 1,000) of the objects in a bucket with each request.
+     *
+     * You can  use the request parameters as selection criteria to return a subset of the objects in a bucket.
+     */
+    listObjects(
```
cirospaciari: just `list` would be better instead of `listObjects`

src/s3/client.zig
```diff
@@ -99,6 +103,127 @@ pub fn delete(
...
+    if (listOptions.start_after) |start_after| {
+        var buff: [1024]u8 = undefined;
+        const encoded = S3Credentials.encodeURIComponent(start_after, &buff, true) catch bun.outOfMemory();
+        search_params.appendFmt(bun.default_allocator, "&start-after={s}", .{encoded}) catch bun.outOfMemory();
+    }
+
+    const result = this.signRequest(.{
+        .path = "",
+        .method = .GET,
+        .search_params = search_params.slice(),
```
cirospaciari, line 175: search_params is leaking memory here, we need to free it after calling signRequest

src/s3/credentials.zig
```diff
@@ -517,7 +536,7 @@ pub const S3Credentials = struct {
         }
 
         // if we allow path.len == 0 it will list the bucket for now we disallow
-        if (path.len == 0) return error.InvalidPath;
+        if (!isComplexeRequest and path.len == 0) return error.InvalidPath;
```
cirospaciari: this is not right we cannot allow bucket unless we are listing a bucket you can pass a comptime bool for this case

src/s3/credentials.zig
```diff
@@ -678,128 +801,266 @@ pub const S3Credentials = struct {
...
+                                        break :brk_canonical try std.fmt.bufPrint(&tmp_buffer, "{s}\n{s}\n{s}\ncontent-disposition:{s}\nhost:{s}\nx-amz-content-sha256:{s}\nx-amz-date:{s}\n\n{s}\n{s}", .{ method_name, normalizedPath, if (search_params) |p| p[1..] else "", encoded_content_disposition, host, aws_content_hash, amz_date, signed_headers, aws_content_hash });
+                                    }
                                 } else {
-                                    break :brk_canonical try std.fmt.bufPrint(&tmp_buffer, "{s}\n{s}\n{s}\nhost:{s}\nx-amz-content-sha256:{s}\nx-amz-date:{s}\n\n{s}\n{s}", .{ method_name, normalizedPath, if (search_params) |p| p[1..] else "", host, aws_content_hash, amz_date, signed_headers, aws_content_hash });
+                                    if (session_token) |token| {
+                                        break :brk_canonical try std.fmt.bufPrint(&tmp_buffer, "{s}\n{s}\n{s}\nhost:{s}\nx-amz-content-sha256:{s}\nx-amz-date:{s}\nx-amz-security-token:{s}\n\n{s}\n{s}", .{ method_name, normalizedPath, if (search_params) |p| p[1..] else "", host, aws_content_hash, amz_date, token, signed_headers, aws_content_hash });
```
cirospaciari, line 1111: this conditions are getting too big we should break to a slow path when we actualy sort in the right order instead of if/else stuff

test/js/bun/s3/s3-list-objects.test.ts
```diff
@@ -0,0 +1,1001 @@
+import { describe, it, expect } from "bun:test";
+import { randomUUIDv7, S3Client, S3ListObjectsResponse, S3Options } from "bun";
+
+const options: S3Options = {
```
cirospaciari, line 5: we should add tests using the CI credentials take a look on the s3.tests.ts

src/bun.js/webcore/blob.zig
```diff
@@ -3596,6 +3596,58 @@ pub const Blob = struct {
 
             return value;
         }
+
+        pub fn listObjects(this: *@This(), store: *Store, globalThis: *JSC.JSGlobalObject, listOptions: JSValue, extra_options: ?JSValue) bun.JSError!JSValue {
```
cirospaciari, line 3714: this should be in S3File.zig not in blob.zig

cirospaciari (changes requested):

packages/bun-types/bun.d.ts
Inqnuam: What about if we want to implement list buckets later ?

src/s3/credentials.zig
Inqnuam: Indeed this is unmaintainable and error prone
I wrote an implementation of the entire SigV4 with dynamic values, need deep testing before merging here

src/s3/list_objects.zig
```diff
@@ -0,0 +1,557 @@
...
+            val.toZigString(&zig_val, globalThis);
+
+            listObjectsOptions.prefix = zig_val.slice();
+        }
+    }
+
+    if (try listOptions.getTruthyComptime(globalThis, "startAfter")) |val| {
+        if (val.isString()) {
+            var zig_val: JSC.ZigString = JSC.ZigString.Empty;
+            val.toZigString(&zig_val, globalThis);
```
cirospaciari: instead of `toZigString` use `bun.String.fromJS` we also need to make sure that this strings are alive otherwise if GC passes or if this is re-encoded to UTF8 etc it can be an UAF, I would do something like I did in S3CredentialsWithOptions for reference
```js
const str = bun.String.fromJS(js_val, globalObject);
defer str.deref(); // str will be alive until you deref()
str.slice()
```

test/js/bun/s3/s3-list-objects.test.ts
```diff
@@ -0,0 +1,1111 @@
...
+  const keyPrefix = `${randomUUIDv7()}/`;
+
+  const file_1 = `${keyPrefix}file_1.txt`;
+  const file_2 = `${keyPrefix}file_2.txt`;
+  const file_3 = `${keyPrefix}file_3.txt`;
+
+  it("should list objects with prefix, maxKeys and nextContinuationToken", async () => {
+    await bucket.write(file_1, "content 1");
+    await bucket.write(file_2, "content 2");
+    await bucket.write(file_3, "content 3");
```
cirospaciari, line 1110: Can you make sure that everything is deleted using an afterAll or try/finally? Can you write a test testing `startAfter`, `encodingType` and `fetchOwner`?

cirospaciari (changes requested):

src/bun.js/webcore/blob.zig
cirospaciari: this can be done after in a followup

packages/bun-types/bun.d.ts
cirospaciari: we decided that `client.list()` is better and in the future for listing buckets will be `client.buckets()`

src/s3/credentials.zig
cirospaciari: will make this in a followup so will mark this as resolved

test/js/bun/s3/s3-list-objects.test.ts
Inqnuam: Sure! in AWS storing 9 bytes is cheaper than making delete request to remove it ,that's why I preferred to keep them, but looks like in R2 delete is free

geoffhendrey:
We are trying to work around problems with listObjectsV2 hanging on bun when there are large numbers of objects in the bucket. WOuld love to see this merge

rileymiller:
Agreed, we are also blocked from adopting bun for some s3 workloads we'd love to migrate over until this is merged.

cirospaciari (approved):

cirospaciari:
Thank you!

Issue #10946, opened 2024-05-09
Bun crashes when running Discord bot example


cronJ:
### How can we reproduce the crash?

Build the example project from https://bun.sh/guides/ecosystem/discordjs

### JavaScript/TypeScript code that reproduces the crash?

```js
// import discord.js
import { Client, Events, GatewayIntentBits } from "discord.js";

// create a new Client instance
const client = new Client({ intents: [GatewayIntentBits.Guilds] });

// listen for the client to be ready
client.once(Events.ClientReady, (c) => {
  console.log(`Ready! Logged in as ${c.user.tag}`);
});

// login with the token from .env.local
client.login(process.env.DISCORD_TOKEN);
```

### Relevant log output

```
Bun v1.1.7 (b0b7db5c) Windows x64
Args: "C:\Users\xxxxx\.bun\bin\bun.exe", "run", "bot.ts"
Features: jsc Bun.stdin(2) dotenv fetch transpiler_cache(3) tsconfig(4) WebSocket 
Builtins: "bun:main" "node:buffer" "node:events" "node:fs" "node:fs/promises" "node:http" "node:path" "node:string_decoder" "node:timers" "node:timers/promises" "node:url" "node:util" "node:util/types" "node:zlib" "node:worker_threads" "undici" "ws" 
Elapsed: 795ms | User: 140ms | Sys: 15ms
RSS: 0.16GB | Peak: 0.16GB | Commit: 0.20GB | Faults: 38876

panic(main thread): Segmentation fault at address 0xFFFFFFFFFFFFFFFF
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
```

### Stack Trace (bun.report)

Bun v1.1.7 ([`b0b7db5`](https://github.com/oven-sh/bun/tree/b0b7db5c060e830801b0273b7ff6e34e3eb5c254)) on windows x86_64 [RunCommand]

Segmentation fault at address 0xFFFFFFFFFFFFFFFF

- `??`
- `??`
- `uv__process_poll_req`
- `??`
- `uv_run`
- [`uws.zig:2619`](https://github.com/oven-sh/bun/blob/b0b7db5c060e830801b0273b7ff6e34e3eb5c254/src/deps/uws.zig#L2619): `run`
- `JSC__VM__holdAPILock`
[...]

PR #16953, opened 2025-02-01
Fix default ports being rejected in preconnect

heimskr:
### What does this PR do?

fixes #19626

Fixes an issue where port validation throws an error after default ports get normalized out of URLs passed to `fetch.preconnect`.

- [x] Code changes

I ran this code:
```js
import { fetch } from "bun";
fetch.preconnect("https://example.com");
```

Fails in the current release version but doesn't with this PR.


robobun:
<div><sup>Updated 6:34 PM PT - Jan 31st, 2025</sup></div>

@heimskr, your commit 3ee9af7649e11e38d03d0db623996d716c6276d2 is building: [`#10946`](https://buildkite.com/bun/bun/builds/10946)

heimskr:
Now it fails to reject port 0. I'll fix this.

dylan-conway:
I don't fully understand this change yet, but similar might be needed in `bun_js.zig` in `doPreconnect`

heimskr:
> I don't fully understand this change yet, but similar might be needed in `bun_js.zig` in `doPreconnect`

My latest commit moves the check into hasValidPort, which is used in both `doPreconnect` and `Bun__fetchPreconnect_`.

Jarred-Sumner (changes requested):
Can you add a test

RiskyMH:
> Can you add a test

I added test that fails without the fix

tkhg:
any update on this one?

hckhanh:
any updates on this?

Aidan647:
any updates on this?

robobun:
Thank you for this fix, @heimskr, and @RiskyMH for the test.

This PR patches `src/url.zig`. That file no longer exists, because the source moved from Zig to Rust. The check is now `URL::has_valid_port` in `src/url/lib.rs`.

#43034 makes the same change there and also closes #19626. Its first commit credits both of you as co-authors. Its tests cover the `fetch.preconnect("http://localhost")` case from this PR and the `--fetch-preconnect` flag. The `example.com` test is not in #43034, because a test must not contact the public internet. I ran that sequence manually with the #43034 build and a local listener on port 80: the preconnect opens one connection, and the `fetch()` that follows reuses it and returns 200.

I am closing this PR in favor of #43034.

Issue #16892, opened 2025-01-30
SQL from bun gets timeouts very easly


belizariogr:
### What version of Bun is running?

1.2.1

### What platform is your computer?

Windows x64, Linux x64

### What steps can reproduce the bug?

Using the SQL from "bun", the server hangs and returns:

``Idle timeout reached after 5s``

 changing the driver for the "postgres" (that the native driver is based on) the problem does not happening.

```typescript
import { Elysia } from 'elysia'
import { SQL } from 'bun'
import postgres from 'postgres'

const sql = new SQL({ url: process.env.DATABASE_URL, max: 10, idleTimeout: 5 });

// const sql = postgres(process.env.DATABASE_URL, { max: 10, idle_timeout: 5});


new Elysia({ serve: { idleTimeout: 10 } })
    .get('/', async () => {
        const res = await sql`SELECT * FROM cidades`;
        return { rows: res };
    })
    .listen(3000, () => {
        console.log('Server is running on port 3000');
    });
``` 

Here is a video showing the issue happening.

https://youtu.be/tjbe4-xixUk

### What is the expected behavior?

Return the records all the time.

### What do you see instead?

Hangs and gets timeout.

### Additional information

Tested on Windows x64 and Linux (Ubuntu 24.04), PostgreSQL 17.

belizariogr:
Just to let you know... the issue still persists on 1.2.2.

@RiskyMH @Jarred-Sumner  @cirospaciari

cirospaciari:
> Just to let you know... the issue still persists on 1.2.2.
> 
> [@RiskyMH](https://github.com/RiskyMH) [@Jarred-Sumner](https://github.com/Jarred-Sumner) [@cirospaciari](https://github.com/cirospaciari)

Can you send a repro?
Im trying the following code:

```js

import { Elysia } from "elysia";

const sql = new SQL({
  url: "postgresql://bun:bunbunbun@localhost:5432/bun",
  idleTimeout: 5,
});

new Elysia({ serve: { idleTimeout: 10 } })
  .get("/", async () => {
    const res = await sql`SELECT * FROM "users" limit 1`;
    return { rows: res };
  })
  .listen(3000, () => {
    console.log("Server is running on port 3000");
  });
await fetch("http://localhost:3000/")
  .then(res => res.json())
  .then(console.log);

await Bun.sleep(10000);
await fetch("http://localhost:3000/")
  .then(res => res.json())
  .then(console.log);
```

Output:

[...]

PR #16984, opened 2025-02-02
fix(sql) disable idle timeout when still processing data

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/16892
Fix: https://github.com/oven-sh/bun/issues/16999

Disable idle timeout when processing data, fix short read handling, remove unnecessary/dead code and small refactor

- [X] Code changes

Test

### DIFF
--- a/src/sql/postgres.zig
+++ b/src/sql/postgres.zig
@@ -1306,6 +1306,7 @@ pub const PostgresSQLConnection = struct {
         if (this.timer.state == .ACTIVE) {
             this.globalObject.bunVM().timer.remove(&this.timer);
         }
+        this.timer.state = .CANCELLED;
     }
     pub fn resetConnectionTimeout(this: *PostgresSQLConnection) void {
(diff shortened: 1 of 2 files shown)


robobun:
<div><sup>Updated 10:19 PM PT - Feb 1st, 2025</sup></div>

@cirospaciari, your commit 195fa6aefba13bd930c41fe2cdf6dbfb9488a04d is building: [`#10983`](https://buildkite.com/bun/bun/builds/10983)

Jarred-Sumner:
maybe instead it'd be better to reset the connection timeout at both the start and end of the onData callback?

cirospaciari:
> maybe instead it'd be better to reset the connection timeout at both the start and end of the onData callback?

I just disable the idle timeout when processing the data, if we are processing data we are active.

Jarred-Sumner:
Do you think this only needs one test?

Issue #16973, opened 2025-02-01
Bun build removes script html element if entrypoint html file doesn't have header element


darekzak:
### What version of Bun is running?

1.2.2-canary.3+26d3688e5

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

Run ```$ Bun build ./index.html --outdir=dist```

```html
<!doctype html>
<html>
  <body>
    test
    <script src="./app.ts" type="module"></script>
  </body>
</html>
```

If there is no ```<head></head>``` element inside ./index.html then in dist/index.html there is no 
```html
<script src="./app.ts" type="module"></script>
```
Additionally if there is head element inside ./index.html and script element is in body, then after running ```$ Bun build ./index.html --outdir=dist``` script element is being moved to head element.

### What is the expected behavior?

After running ```$ Bun build ./index.html --outdir=dist``` html script element in dist/index.html should be in the same place where it was in ./index.html. So if it was in header, it should be in header. If it was inside body, it should stay there. 
```html
<script src="./app.ts" type="module"></script>
```

### What do you see instead?

> dist/index.html
```html
<!doctype html>
<html>
  <body>
    test
    
  </body>
</html>

```

PR #16993, opened 2025-02-02
Introduce `bun ./index.html`

Jarred-Sumner:
### What does this PR do?

This lets you start a frontend dev server by directly running `.html` files in Bun.

```sh
> bun index.html
Bun v1.2.3 ready in 0.03 ms

➜ http://localhost:3001/

Press h + Enter to show shortcuts
```

For multi-page apps, you can pass multiple routes explicitly.
image

You can also pass a glob.
image

If you pass only one route, that route is matched as the fallback route so that you can use this with single-page apps that do client-side routing from a single shared entry point.

There are tests

### DIFF
--- a/docs/bundler/html.md
+++ b/docs/bundler/html.md
@@ -284,6 +284,9 @@ All paths are resolved relative to your HTML file, making it easy to organize yo
 - No HMR support yet
 - Need more plugins
 - Need more configuration options for things like asset handling
+- Need a way to configure CORS, headers, etc.
+
+If you want to submit a PR, most of the [code is here](https://github.com/oven-sh/bun/blob/main/src/js/internal/html.ts).
 
 ## How this works
 
--- a/docs/bundler/html.md
+++ b/docs/bundler/html.md
@@ -286,7 +286,7 @@ All paths are resolved relative to your HTML file, making it easy to organize yo
 - Need more configuration options for things like asset handling
 - Need a way to configure CORS, headers, etc.
 
-If you want to submit a PR, most of the [code is here](https://github.com/oven-sh/bun/blob/main/src/js/internal/html.ts).
+If you want to submit a PR, most of the [code is here](https://github.com/oven-sh/bun/blob/main/src/js/internal/html.ts). You could even copy paste that file into your project and use it as a starting point.
 
 ## How this works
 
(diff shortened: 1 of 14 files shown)


robobun:
<div><sup>Updated 7:08 AM PT - Feb 2nd, 2025</sup></div>

@Jarred-Sumner, your commit 2ed24db1541415bd66d827fddf63f05bd9b0c1dd is building: [`#11019`](https://buildkite.com/bun/bun/builds/11019)

jakeg:
@Jarred-Sumner why's it stripping out `<script>` tags?

image

Here you can see the `<script>` tag is gone:

image

Jarred-Sumner:
Looks like a bug. It’s expecting a `<head>` tag to inject the `<script>` into

jakeg:
Yeah, I never bother with `<html>`, `<head>` or `<body>` tags for quick things, as my understanding is that they're optional, and the minimum required for a valid HTML document is a `<!doctype html>` and a `<title>`.

RiskyMH:
btw this bug mentioned above is tracked in https://github.com/oven-sh/bun/issues/16973

docs/bundler/html.md
```diff
@@ -13,45 +13,221 @@ As of Bun v1.1.43, Bun's bundler now has first-class support for HTML. Build sta
...
+
+This generates `./app.css` and `./app.js` in the output directory. All CSS files imported from JavaScript will be bundled into a single CSS file per entry point. If you import the same CSS file from multiple JavaScript files, it will only be included once in the output CSS file.
+
+## Plugins
+
+The dev server supports plugins.
+
+### Tailwind CSS
+
+To use TailwindCSS, add the plugin to your `bunfig.toml`:
```
RiskyMH, line 160: Don't forget to mention you need to install this package (as it was one a decent confusion with people thinking it was inbuilt)

Also why is there so much duplication between pages, are you merging `fullstack.md` into this? Or at very least make some template so they are the same and can't as easily get so different

Jarred-Sumner: duplication in docs is generally fine IMO. People don't read every page.

Jarred-Sumner:
Going to merge this, but @paperclover please take a look over it when you get a chance.

Issue #11052, opened 2024-05-13
please check so i have to install some software's please


shahidiqbal369:
### How can we reproduce the crash?

Bun v1.1.8 (89d25807) Windows x64
Args: "C:\Users\shahi\.bun\bin\bun.exe", "install"
Elapsed: 69ms | User: 0ms | Sys: 46ms
RSS: 51.59MB | Peak: 51.59MB | Commit: 92.16MB | Faults: 12779

panic(main thread): Internal assertion failure
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.8/wi189d2580AAglq4L+g0uIwit1F+swCm/uwH0v53D0k7kE0jnjE8zqvFqg4BA0eNrzzCtJLcpLzFFILC5OLSrJzM9TSEvMzCktSgUAiSkKPg

### Stack Trace (bun.report)

Bun v1.1.8 ([`89d2580`](https://github.com/oven-sh/bun/tree/89d25807fbf4488d3bfcb116484d0c2cf824f98c)) on windows x86_64 [InstallCommand]

**panic**: Internal assertion failure

- [`debug.zig:373`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/deps/zig/lib/std/debug.zig#L373): `panicExtra`
- [`debug.zig:348`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/deps/zig/lib/std/debug.zig#L348): `panic`
- [`output.zig:356`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/output.zig#L356): `panic`
- [`bun.zig:3177`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/bun.zig#L3177): `assertionFailure`
- [`lockfile.zig:4902`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/install/lockfile.zig#L4902): `save`
- [`lockfile.zig:1643`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/install/lockfile.zig#L1643): `saveToDisk`
[...]

PR #17013, opened 2025-02-03
Fix loading react-jsxdev instead of react-jsx

### LINKED ISSUE #3768: Bun.build always assume "react-jsxdev" even when set "react-jsx"

Jarred-Sumner:
### What does this PR do?

Fixes #3768

Fix loading react-jsxdev instead of react-jsx

The main bug: the state of the resolver sometimes got out of sync with the state of the transpiler for tracking if it should use development JSX or non-development JSX. This meant that setting `NODE_ENV` to a value in the parent process that is different than the environment seen from Bun.build would cause this to not work properly.

Separately, I also implemented tracking `react-jsx` and `react-jsxdev` to work as expected.

There is a test that runs each permutation of `NODE_ENV` for runtime and bundler.

There is also a test for Bun.serve() `static`


robobun:
<div><sup>Updated 6:08 AM PT - Feb 3rd, 2025</sup></div>

@Jarred-Sumner, your commit 194e4eafca42d9bd83ae1df350af7d91980db71a is building: [`#11052`](https://buildkite.com/bun/bun/builds/11052)

src/options.zig
```diff
@@ -991,13 +991,18 @@ pub const ESMConditions = struct {
...
+    const RuntimeDevelopmentPair = struct {
+        runtime: JSX.Runtime,
+        development: ?bool,
+    };
+
+    pub const RuntimeMap = bun.ComptimeStringMap(RuntimeDevelopmentPair, .{
+        .{ "classic", RuntimeDevelopmentPair{ .runtime = .classic, .development = null } },
+        .{ "automatic", RuntimeDevelopmentPair{ .runtime = .automatic, .development = true } },
+        .{ "react", RuntimeDevelopmentPair{ .runtime = .classic, .development = null } },
+        .{ "react-jsx", RuntimeDevelopmentPair{ .runtime = .automatic, .development = true } },
```
paperclover, line 1003: why is development true in this case?

src/options.zig
```diff
@@ -1575,13 +1580,17 @@ pub const BundleOptions = struct {
 
     supports_multiple_outputs: bool = true,
 
+    force_development: ?bool = null,
```
paperclover: it is not clear exactly what the difference between `null`, `false`, and `true` are.

paperclover (changes requested):
review comments addressed in person. i will probably fix these last points

- convert ?bool into an enum tristate between unspecified, production, development
- document the behavior

PR #17021, opened 2025-02-03
fix(gitattributes): add github syntax highlighting for `.mdc` files

okineadev:
### What does this PR do?

Add github syntax highlighting for `.mdc` files


okineadev:
It would be better if we generally used some file like `.airules.md` or something like that so that it was neutral and did not promote or mention this or that service

Making such a setting for Copilot is not a problem at all, but for Cursor I don't know

okineadev:
Is there any reason for the closure?

PR #17027, opened 2025-02-03
Fix potential crash when printing errors in bun install

Jarred-Sumner:
### What does this PR do?

Fix potential crash when printing HTTP errors in bun install caused by referencing potentially invalid memory

This should fix the patch test

### DIFF
--- a/src/install/install.zig
+++ b/src/install/install.zig
@@ -255,7 +255,8 @@ pub const Aligner = struct {
 };
 
 const NetworkTask = struct {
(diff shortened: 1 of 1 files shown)


robobun:
<div><sup>Updated 2:47 PM PT - Feb 3rd, 2025</sup></div>

@Jarred-Sumner, your commit 15a3e4d60407a6c6981cb03ff04ff737bb9ae14c is building: [`#11068`](https://buildkite.com/bun/bun/builds/11068)

Issue #11080, opened 2024-05-15
Update fish & zsh completions


Jarred-Sumner:
`fish` and `zsh` completions are extremely out of date. Many flags are missing. We need to update these to include the missing flags.

Flags for `bun build`:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/cli.zig#L213-L237

Flags for `bun test`:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/cli.zig#L241-L251

Flags for `bun run <source-file.js>` and `bun <source-file.js>`:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/cli.zig#L170-L189

Flags for almost all commands:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/cli.zig#L147-L169

Flags for `bun install`, `bun add`, `bun remove`:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/install/install.zig#L7760-L7784

Flags for `bun install` specifically:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/install/install.zig#L7787-L7794

Flags for `bun add <package>` specifically:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/install/install.zig#L7804-L7810

Flags for `bun pm ls`:

https://github.com/oven-sh/bun/blob/d62cd47885505dbcb77e292c0caf3958aad3774e/src/install/install.zig#L7799-L7802

robobun:
Same request as #2503: the shipped fish/zsh (and bash) completion scripts are out of date and need to be updated with the CLI flags added since; bringing the completions up to date (or generating them automatically, as discussed in #2503) resolves both. #2503 is the older issue and already has the discussion, including a maintainer response, and the similar #4016 was previously closed as a duplicate of it.

Duplicate of #2503

Please add a 👍 or any extra details over there so it all stays in one place. If this is actually a different problem, say so here and we'll reopen.

PR #17035, opened 2025-02-04
Add shadcn, tailwind and react detection & templates to bun create. Also: `bun install --analyze <files...>`

### LINKED ISSUE #4133: feature: bun install <script>

It is currently possible to run a script  `bun run script.ts` without a package.json, so bun is capable of doing auto-install and resolution of the dependencies needed to run a script. 

It is however impossible to ask bun to generate a package.json and lockfile from a script.

This is useful in a case where you want to quickly migrate from using `bun run` to using `bun install`. It would be similar to `go mod tidy` where go resolve the dependencies  from the logic and generate the dependencies in go.mod.

For us at windmill: https://github.com/windmill-labs/windmill, it would allow to securely generate a lockfile and package.json from the script directly so that we can benefit from a secure 2-step process of doing `bun install`   in a sandbox that has write access on the global cacce and then `bun run.ts` in a sandbox that doesn't have access to the global cache.

`bun install <script>` would generate a package.json a lockfile from a script and install the dependencies

_No response_

Jarred-Sumner:
### What does this PR do?

`bun install --analyze <files ...>`  scans imported packages from source files using Bun's bundler and adds any missing ones to package.json

fixes #4133

WIP


robobun:
<div><sup>Updated 9:51 PM PT - Feb 3rd, 2025</sup></div>

@Jarred-Sumner, your commit a8180608930e3f9a7f4214995e61a4d56f32678b is building: [`#11080`](https://buildkite.com/bun/bun/builds/11080)

dylan-conway:
Ideas for flag name: `--infer`, `--create`, `--from-source`

src/bundler/bundle_v2.zig
```diff
@@ -1451,6 +1451,64 @@ pub const BundleV2 = struct {
...
+                                    }
+                                }
+                            }
+                        } else if (strings.indexOfChar(record.path.text, '/')) |slash_index| {
+                            if (slash_index > 0) {
+                                package_path = record.path.text[0..slash_index];
+                            }
+                        } else {
+                            package_path = record.path.text;
+                        }
```
dylan-conway:
```suggestion
                        } else if (!strings.containsChar(record.path.text, '/')) {
                            package_path = record.path.text;
                        }
```
Reasoning: the package shouldn't contain `/` if it doesn't start with `@`

dylan-conway: or, this could just be `else { package_path = record.path.text }` and let `strings.isNPMPackageName` handle invalid name

src/bundler/bundle_v2.zig
```diff
@@ -1487,6 +1546,11 @@ pub const BundleV2 = struct {
         const reachable_files = try this.findReachableFiles();
         reachable_files_count.* = reachable_files.len -| 1; // - 1 for the runtime
 
+        if (fetcher) |fetch| {
+            try this.getAllDependencies(reachable_files, fetch);
+            return std.ArrayList(options.OutputFile).init(allocator);
```
dylan-conway:
```suggestion
            .init(allocator);
```
Not important

dylan-conway (changes requested):
Needs tests

RiskyMH:
Not sure how hard it would be, but could be cool to support `bunfig.toml` and its plugins.

Also you probably already know but it installs node builtins from npm (ie `fs` and `path`)

src/create/projects/react-shadcn-spa/components.json
```diff
@@ -0,0 +1,21 @@
+{
```
Jarred-Sumner, line 1: @shadcn let me know if you have any opinions on what we should put in the components.json `bun create` will inject when it detects shacdn

PR #17058, opened 2025-02-05
Enable asan on debug macos aarch64 builds

pfgithub:
### What does this PR do?

Problems:
- Something is wrong with caching because you can't build this branch coming directly from main, you have to delete `build` first

Usage:
- Disable with `bun run build -DENABLE_ASAN=OFF` (must delete `build` dir)
- `ASAN_OPTIONS=detect_leaks=1 bun-debug abcd`
  - not very useful
- `ASAN_OPTIONS=detect_stack_use_after_return=1 bun-debug abcd`
  - only works for codepaths which do not use javascript core

Running node tests locally:
- test-async-hooks-worker-asyncfn-terminate-4.js
- some failures due to mimalloc warning
- `for file in (ls test/js/node/test/parallel); echo $file; bun-debug test
/js/node/test/parallel/$file; end`

Followup:
- Support on linux
- Support on windows
- Maybe support in a ReleaseSafe build and run tests on that?


robobun:
<div><sup>Updated 6:09 PM PT - Feb 4th, 2025</sup></div>

@pfgithub, your commit 7d5a8224701b37f3d3cec3516c7e2aaa74907e0e is building: [`#11105`](https://buildkite.com/bun/bun/builds/11105)

cmake/targets/BuildMimalloc.cmake
```diff
@@ -19,6 +19,10 @@ set(MIMALLOC_CMAKE_ARGS
   -DMI_SKIP_COLLECT_ON_EXIT=ON
 )
 
+if(ENABLE_ASAN)
```
DonIsaac, line 22: you may also need to update `CompilerFlags.cmake` so other libraries get `-fsanitize=address`

Issue #11171, opened 2024-05-18
bignum not working


mgpai22:
### What version of Bun is running?

v1.1.8

### What platform is your computer?

docker

### What steps can reproduce the bug?

Trying to run the [bpp branch](https://github.com/reqlez/ErgoStratumServer/tree/bpp) using `bun run start.js`

Get this error:
```
 107 |         return opts[p] || p;
 108 |       })
 109 |     );
 110 |     tries.push(n);
 111 |     try {
 112 |       b = opts.path ? requireFunc.resolve(n) : requireFunc(n);
                                                      ^
 TypeError: /app/node_modules/bignum/build/Release/bignum.node: undefined symbol: _ZN2v814ObjectTemplate21SetInternalFieldCountEi
       at bindings (/app/node_modules/bindings/bindings.js:112:48)
       at /app/node_modules/bignum/index.js:1:5
       at /app/lib/algoProperties.js:1:5
       at /app/lib/index.js:5:1
       at /app/start.js:1:5

 Bun v1.1.8 (Linux x64 baseline)

```

Electroid:
It looks like [`bignum`](https://www.npmjs.com/package/bignum?activeTab=code) uses the V8 C++ APIs, which Bun does not yet support.

Electroid:
Duplicate of #4290

PR #17098, opened 2025-02-06
cmake: download zig fork from cdn instead of git clone

nektro:


robobun:
<div><sup>Updated 7:19 PM PT - Feb 5th, 2025</sup></div>

@nektro, your commit 97f0d5b6e0f66f069e90784e6daa830065fe204d is building: [`#11171`](https://buildkite.com/bun/bun/builds/11171)

PR #17106, opened 2025-02-06
cmake: don't download WebKit unless we're gonna need it

nektro:
this will save both bandwidth and CI time

for example, running `bun run build --target zstd` with this patch will no longer eagerly download WebKit


robobun:
<div><sup>Updated 10:50 PM PT - Feb 5th, 2025</sup></div>

@nektro, your commit 7da33c1afb353228bf70c2bad6afe259f7d72372 is building: [`#11193`](https://buildkite.com/bun/bun/builds/11193)

Issue #11207, opened 2024-05-20
Segmentation fault at address 0xFFFFFFFFFFFFFFFF


SenorLawyer:
### How can we reproduce the crash?

by running my code (NODE_ENV=development bun --watch run src/index.ts)

### JavaScript/TypeScript code that reproduces the crash?

```shell
erm there are like 50 files
```

### Relevant log output

```shell
============================================================
Bun v1.1.8 (89d25807) Windows x64
Args: "C:\Users\larsm\.bun\bin\bun.exe", "--watch", "run", "src/index.ts"
Features: jsc Bun.stdin(2) dotenv fetch transpiler_cache(3) tsconfig(4) WebSocket 
Builtins: "bun:main" "node:buffer" "node:events" "node:fs" "node:fs/promises" "node:http" "node:path" "node:string_decoder" "node:timers" "node:timers/promises" "node:url" "node:util" "node:util/types" "node:zlib" "node:worker_threads" "undici" "ws" 
Elapsed: 533ms | User: 31ms | Sys: 0ms
RSS: 0.17GB | Peak: 0.17GB | Commit: 0.22GB | Faults: 41007

panic(main thread): Segmentation fault at address 0xFFFFFFFFFFFFFFFF
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.8/wr189d2580AqoiggV+h7u8Cu9ir2C8u0ne4ttkegzqkesq+xG2t9t1C6r+pF8kzqF4+pvFA2DD

--- Bun is auto-restarting due to crash [time: 1716242176859] ---
```

### Stack Trace (bun.report)

Bun v1.1.8 ([`89d2580`](https://github.com/oven-sh/bun/tree/89d25807fbf4488d3bfcb116484d0c2cf824f98c)) on windows x86_64 [RunCommand]

Segmentation fault at address 0xFFFFFFFFFFFFFFFF

- `??`
- `??`
- `uv__process_poll_req`
- `??`
- `uv_run`
- [`uws.zig:2619`](https://github.com/oven-sh/bun/blob/89d25807fbf4488d3bfcb116484d0c2cf824f98c/src/deps/uws.zig#L2619): `run`
- `JSC__VM__holdAPILock`
[...]

PR #17117, opened 2025-02-06
test: bump pglite version

190n:
### What does this PR do?

Bumps the version of PGlite we test from 0.2.15 to 0.2.16. This test was introduced in #16188 because older versions of the IPInt Wasm interpreter had errors with PGlite ("Out of bounds memory access"). But it turns out 0.2.15 had another issue ("Out of bounds call_indirect") that affects (at least on macOS aarch64) all versions of Bun I've tried, regardless of the Wasm backend used (more discussion at electric-sql/pglite#486).

I'm still investigating why this test had not been failing in CI.

Ran the new PGlite test.

### DIFF
--- a/test/bun.lock
+++ b/test/bun.lock
@@ -6,7 +6,7 @@
       "dependencies": {
         "@azure/service-bus": "7.9.4",
         "@duckdb/node-api": "1.1.3-alpha.7",
-        "@electric-sql/pglite": "0.2.15",
+        "@electric-sql/pglite": "0.2.16",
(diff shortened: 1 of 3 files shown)


robobun:
<div><sup>Updated 11:51 AM PT - Feb 6th, 2025</sup></div>

@190n, your commit 15f44c1b8e5681764bf479b613b5cf26dd3e5078 is building: [`#11207`](https://buildkite.com/bun/bun/builds/11207)

PR #16865, opened 2025-01-29
Try bringing release/acquire heap access back

Jarred-Sumner:
### What does this PR do?

Try bringing release/acquire heap access back

Issue #11225, opened 2024-05-21
socket.destroyed not sets to true and socket.write triggers the callback instantly


Mrgaton:
### What version of Bun is running?

1.1.8

### What platform is your computer?

linux - debian

### What steps can reproduce the bug?

download a file from an express server that is over 512mb that is using bun

Mrgaton:
the promise ends instanly before the data arrived to the client

Mrgaton:
Yes, the callback trigers instanly. How i can fix it?

Mrgaton:
And `req.socket.destroyed` seems to not set to true when download is cancelled

Electroid:
This is because Bun's `node:http` implementation uses `fetch()` and we do not allow access to the `request.socket`. This is something we want to change.

Electroid:
Duplicate of #11225

Electroid:
Duplicate of #11230

PR #17126, opened 2025-02-06
 Move GC signal configuration to run earlier

190n:
### What does this PR do?

In #16865 we told JSC to use SIGPWR instead of SIGUSR1 to suspend/resume threads for garbage collection. This is because whatever signal it uses can't be relied on by JS code, and we don't want to break any libraries that use SIGUSR1. But the place where we ran that configuration was too late, and it would sometimes fail to reconfigure the signal because JSC had already been initialized by that point (so it would still use SIGUSR1). The case I discovered was `bun test --test-name-pattern`, because this creates a regex for the test filter, using JSC's regex implementation, and this regex is made before `JSCInitialize` had a chance to run. Now we run this configuration in `main()`, which should be early enough.

I also changed `lldb_commands` to use the right signal to make debugging on Linux more pleasant (see comment in that file).

Tried using test filters and running the `14982.test.ts` that would hang before we initially fixed the signal issue.

### DIFF
--- a/misctools/lldb/lldb_commands
+++ b/misctools/lldb/lldb_commands
@@ -1,4 +1,12 @@
-process handle -p true -s false -n false SIGUSR1
+# Tell LLDB what to do when the debugged process receives SIGPWR:
+# pass it through to the process (-p), but do not stop the process (-s)
+# or notify the user (-n).
+#
+# JSC's garbage collector sends this signal (as configured by Bun at
+# main.zig:71) to the JS thread to suspend or resume it. So stopping
+# the process would just create noise when debugging any
+# long-running script.
+process handle -p true -s false -n false SIGPWR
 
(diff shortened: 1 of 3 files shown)


robobun:
<div><sup>Updated 3:42 PM PT - Feb 6th, 2025</sup></div>

@190n, your commit 21c0cdd93177d7e829658e2dd4c853c647bcff90 is building: [`#11225`](https://buildkite.com/bun/bun/builds/11225)

190n:
Closed in favor of changing the default in our WebKit fork, as that avoids running the configuration function (which may gain unknown side effects in the future) when we don't need to and also avoids running into failures because we didn't know we needed to reconfigure the signal in some edge case (like with `bun test --test-name-pattern`).

PR #17128, opened 2025-02-07
Download zig from oven-sh/zig releases & auto set up zls

pfgithub:
### What does this PR do?

Downloads zig from https://github.com/oven-sh/zig/releases automatically

These releases include zls, so updates vscode settings to use zls from vendor/zig/zls.exe in order to not require manual zls setup

Does not require deleting the `build` folder because the optionx for `ZIG_COMMIT` is changed to `set`. optionx will save its default value into the cache after the first run, and not clear it until the cache is deleted.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,6 +36,7 @@
   // "zig.buildOnSave": true,
   "zig.buildFilePath": "${workspaceFolder}/build.zig",
   "zig.path": "${workspaceFolder}/vendor/zig/zig.exe",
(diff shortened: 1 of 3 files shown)


robobun:
<div><sup>Updated 5:39 PM PT - Feb 6th, 2025</sup></div>

@pfgithub, your commit b25dc4bb841a45e022e5513b27412aa85f6a07ed is building: [`#11235`](https://buildkite.com/bun/bun/builds/11235)

Issue #6037, opened 2023-09-25
Bash autocomplete


manast:
### What is the problem this feature would solve?

Since Bun's early days, I have been annoyed by this missing feature: bash autocomplete does not work! https://sourabhbajaj.com/mac-setup/BashCompletion/
On all Unix commands, I expect bash autocompletion to work, this is specially annoying if you want to use bun to execute a script that is inside a nested directory. Not sure why it is not working on bun, I thought this was a feature of bash that all commands got for free, maybe something in bun is preventing this default behaviour?

### What is the feature you are proposing to solve the problem?

Fix bun so that autocompletion works.

### What alternatives have you considered?

None.

Nekosis:
> I thought this was a feature of bash that all commands got for free, maybe something in bun is preventing this default behaviour?

No. Bash completions need to be written manually, so Bun will need to create their own.

manast:
Ah ok. I think bun does actually have its own, and that's why it is not working as expected. For example, if I type ```bun``` and press TAB, I get something like:
```
-- scripts --
dev                                              start
-- command --
add      -- Add a dependency to package.json
bun      -- Generate a bundle
create   -- Create a new project
dev      -- Start a dev server
help     -- Show command help
install  -- Install packages from package.json
pm       -- Manage local packages
remove   -- Remove a dependency from package.json
run      -- Run a script or package bin
upgrade  -- Upgrade to the latest version of bun
x        -- Run a command from a local or remote NPM package
-- files --
```
But if I type ```bun sr``` and the press TAB it does not autocomplete to the ```src``` directory, whereas if I just type any other unix command or even an unexisting command, it will autocomplete the directory and let me navigate in the files by tabbing.

Nekosis:
I don't think you're supposed to call `bun example.js` to directly run a file. Don't you mean `bun run example.js`?

manast:
Aha. Using "bun run" the autocomplete actually works. But you can use bun without run and it works the same (besides autocomplete), for example:
```bash
bun --watch src/index.ts
```

Nekosis:
I see. I've never used a program that could put keywords/commands and file paths in the same field before, so I wouldn't know if this is a Bun problem or a `bash-completion` problem.

For reference, Bash completions are just shell scripts. If you're at all familiar with Bash scripting, maybe you'd want to take a look at [Bun's completions script](https://github.com/oven-sh/bun/blob/main/completions/bun.bash) to see if you can narrow down the issue.

[...]

PR #17147, opened 2025-02-07
fix: bash completion script

Foo-x:
### What does this PR do?

https://github.com/oven-sh/bun/issues/6037

Fixed the bash completion script to complete scripts in package.json.

- `local scripts="${matched%%\}*}";` didn't work if scripts contained `}` characters.
- Since `re_comp_word_script` was not defined, the condition `( "${COMPREPLY[*]}" =~ ${re_comp_word_script} )` was always true and completions for scripts were removed.

---

- [ ] Code changes

package.json
image

| before | after |
| --- | --- |
| image | image |
| image | image |

### DIFF
--- a/completions/bun.bash
+++ b/completions/bun.bash
@@ -43,22 +43,13 @@ _read_scripts_in_package_json() {
 
     [[ "${package_json}" =~ "\"scripts\""[[:space:]]*":"[[:space:]]*\{(.*)\} ]] && {
         local package_json_compreply;
-        local matched="${BASH_REMATCH[@]:1}";
-        local scripts="${matched%%\}*}";
-        scripts="${scripts//@(\"|\')/}";
-        readarray -td, scripts <<<"${scripts}";
-        for completion in "${scripts[@]}"; do
-            package_json_compreply+=( "${completion%:*}" );
-        done
+        readarray -t package_json_compreply < <(bun -e 'import("package.json").then(p=>Object.keys(p.scripts).forEach(k=>console.log(k)))')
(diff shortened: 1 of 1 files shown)


RiskyMH:
This appears to work. But the whole bash completions feel buggy. I don't think it was this pr, but folders aren't autocompleting for me in `bun run` (it only accepts files or the scripts).

Also `bun build` and some other commands now have the `package.json` scripts which is odd.

RiskyMH:
Also just found out that there are other bun cli commands you can use instead of reading yourself. 

For example `bun getcompletes i` seems to be `package.json` scripts:

https://github.com/oven-sh/bun/blob/c2edbe848fdafd636595a690e6fa6b43d116457d/completions/bun.zsh#L674-L677
https://github.com/oven-sh/bun/blob/c2edbe848fdafd636595a690e6fa6b43d116457d/completions/bun.zsh#L873-L874
https://github.com/oven-sh/bun/blob/c2edbe848fdafd636595a690e6fa6b43d116457d/completions/bun.zsh#L920

RiskyMH:
i think its still a good improvement, only thing that i dont think is wise is `bun build` showing pkg scripts, but otherwise nice :) 

full fix could be later

Foo-x:
OK, I have fixed to read package.json scripts only on `bun run`!

cam-narzt:
Just a small head's up. The shebang on line 1 is wrong, it should be `#!` not just `#`. Also the script has quite a few misuses of array variables which shellcheck will point out to you if you run it.

Issue #11305, opened 2024-05-23
[windows] bunshell stacktrace test is failing


nektro:
in `test/js/bun/shell/bunshell.test.ts`

```ts
  test("stacktrace", async () => {
    // const folder = TestBuilder.tmpdir();
    const code = /* ts */ `import { $ } from 'bun'

    $.throws(true)

    async function someFunction() {
      await $\`somecommandthatdoesnotexist\`
    }

    await someFunction()
    `;

    const [_, lineNr] = code
      .split("\n")
      .map((l, i) => [l, i + 1] as const)
      .find(([line, _]) => line.includes("somecommandthatdoesnotexist"))!;

    if (lineNr === undefined) throw new Error("uh oh");

    await TestBuilder.command`BUN_DEBUG_QUIET_LOGS=1 ${BUN} -e ${code} 2>&1`
      .exitCode(1)
      .stdout(s => expect(s).toInclude(`[eval]:${lineNr}`))
      .run();
  });
```

```
Expected to include: "[eval]:6"
Received: "bun: command not found: somecommandthatdoesnotexist\n1 | import { $ } from 'bun'\n2 | \n3 |     $.throws(true)\n4 | \n5 |     async function someFunction() {\n                   ^\nShellError: Failed with exit code 1\n info: {\n  \"exitCode\": 1,\n  \"stderr\": \"bun: command not found: somecommandthatdoesnotexist\\n\",\n  \"stdout\": \"\"\n}\n\n      at new ShellError (:12:16)\n      at new ShellPromise (:70:16)\n      at BunShell (:194:35)\n      at C:\\bun\\test\\[eval]:5:12\n      at asyncFunctionResume (:1:11)\n      at someFunction (C:\\bun\\test\\[eval]:3:18)\n      at module code (C:\\bun\\test\\[eval]:9:11)\n      at evaluate (native:1:1)\n      at asyncFunctionResume (:1:11)\n\nBun v1.1.10-debug (Windows x64)\n"

      at C:\bun\test\js\bun\shell\test_builder.ts:223:11
      at asyncFunctionResume (:1:21)
      at doChecks (C:\bun\test\js\bun\shell\test_builder.ts:217:20)
      at C:\bun\test\js\bun\shell\test_builder.ts:268:15
      at asyncFunctionResume (:1:21)
      at promiseReactionJobWithoutPromise (:1:21)
✗ deno_task > stacktrace [58.29ms]
```

PR #17169, opened 2025-02-08
fix memory leak in ERR_INVALID_ARG_TYPE

nektro:
dont use ZigString like that on Zig<->Cpp boundary

verified memory leak regression fix with

```js
import dns from "node:dns";
while (true) {
  try {
    dns.resolve(242);
  } catch (e) {
    e.message;
  }
}
```


robobun:
<div><sup>Updated 11:52 PM PT - Feb 7th, 2025</sup></div>

@nektro, your commit 2666142c2f2ec25e05d8117049bf0a99f94e11b6 is building: [`#11305`](https://buildkite.com/bun/bun/builds/11305)

nektro:
this leaks much more consistently but still takes some time to rise, will add a test that stresses it

```ts
import dns from "node:dns";
while (true) {
  try {
    dns.resolve({ __proto__: null });
  } catch (e) {
    e.message;
  }
}
```

Issue #11314, opened 2024-05-24
[Crash]: Bun crashes when trying to install packages in subdirectories


uhidontkno:
### How can we reproduce the crash?

1. Setup Bun (eg. start a new React project)
2. Go into sub folders (eg. `project/nested/folder`)
3. Install a package (eg. `bun i tailwindcss`)

### JavaScript/TypeScript code that reproduces the crash?

```shell
Not applicable
```

### Relevant log output

```shell
[0.04ms] ".env"
bun add v1.1.9 (bb13798d)
============================================================
Bun v1.1.9 (bb13798d) Linux x64
Args: "bun", "i", "@catppuccin/tailwindcss"
Features: dotenv 
Elapsed: 4ms | User: 6ms | Sys: 3ms
RSS: 1.07GB | Peak: 24.47MB | Commit: 1.07GB | Faults: 0

panic(main thread): ENOENT parsing package.json
oh no: Bun has crashed. This indicates a bug in Bun, not your code.

To send a redacted crash report to Bun's team,
please file a GitHub issue using the link below:

 https://bun.report/1.1.9/li1bb13798AgIgh/h+Cw4126Civq73C2vyo4Csu4n4Cgltg0C+g//zCA0eNqTjjbIlY42Nsx19fN39QuRjjbIVShILCrOzEtXKEhMzk5MT9XLKs7PA0kAACt0DjI

Illegal instruction (core dumped)
```
Working directory: `project/server/static`
fyi: `package.json` resides in the root directory.

### Stack Trace (bun.report)
```
Bun v1.1.9 ([`bb13798`](<https://github.com/oven-sh/bun/tree/bb13798d9802d26770d8467a200022a2b91f7afd>)) on linux x86_64 [InstallCommand]

**panic**: [0m[31mENOENT[0m parsing package.json[0m

- [`debug.zig:373`](<https://github.com/oven-sh/bun/blob/bb13798d9802d26770d8467a200022a2b91f7afd/src/deps/zig/lib/std/debug.zig#L373>): `debug.panicExtra__anon_318900`
- [`debug.zig:348`](<https://github.com/oven-sh/bun/blob/bb13798d9802d26770d8467a200022a2b91f7afd/src/deps/zig/lib/std/debug.zig#L348>): `debug.panic__anon_291546`
[...]

PR #17179, opened 2025-02-08
Avoid creating temporary strings when throwing errors

Jarred-Sumner:
### What does this PR do?

- Avoid creating temporary strings when throwing errors. Use `->view` instead of `.toWTFString()`.
- `toWTFStringForConsole` doesn't escape the string. 
- Do not create jsNumber just to call .toString() on it...

Existing node tests will likely fail


robobun:
<div><sup>Updated 4:12 AM PT - Feb 8th, 2025</sup></div>

@Jarred-Sumner, your commit 1445d4b8c2ed404a4b802ac3ca7da2ba7f018202 is building: [`#11314`](https://buildkite.com/bun/bun/builds/11314)

src/bun.js/bindings/ErrorCode.cpp
```diff
@@ -802,8 +982,12 @@ JSC::JSObject* Bun::createInvalidThisError(JSC::JSGlobalObject* globalObject, JS
     }
 
     // Pathological case: the this value returns a string which is extremely long or causes an out of memory error.
-    const auto& typeString = thisValue.isString() ? String("a string"_s) : JSC::errorDescriptionForValue(globalObject, thisValue);
-    return Bun::createError(globalObject, Bun::ErrorCode::ERR_INVALID_THIS, makeString("Expected this to be instanceof "_s, typeName, ", but received "_s, typeString));
+    WTF::StringBuilder builder;
+    builder.append("Expected this to be instanceof "_s);
+    builder.append(typeName);
+    builder.append(", but received "_s);
+    determineSpecificType(JSC::getVM(globalObject), globalObject, builder, thisValue);
```
dylan-conway, line 1059: should we still check for `thisValue.isString()` and use `"a string"` if true?

PR #17207, opened 2025-02-09
feat: Support codesigning macOS executables in `bun build --compile`

### LINKED ISSUE #7208: Cannot code-sign compiled executable on macOS

1.0.13+f5bf67bd1

Darwin 23.1.0 arm64 arm

1. Build a compiled executable by following https://bun.sh/docs/bundler/executables
2. Create a self-signed certificate with code-signing capabilities by following https://support.apple.com/en-in/guide/keychain-access/kyca8916/mac
3. Attempt to code-sign the compiled executable with the command –

	```
	codesign -f -s "Acme Corp" ./mycli --deep
	```
4. See the error –

	```
	./mycli: replacing existing signature
	./mycli: main executable failed strict validation
	```

The executable should be code-signed correctly and runnable on the system without Gatekeeper blocks/prompts.

### What do you see instead?

```
./mycli: replacing existing signature
./mycli: main executable failed strict validation
```

### Additional information

FWIW, Deno's compiled binaries face the same issue (although I'm not sure if the root cause is the same) – see https://github.com/denoland/deno/issues/575 and https://github.com/denoland/deno/issues/17753

Pranav2612000:
### What does this PR do?

Based off #15525, this PR adds support for code signing macos executables. This is done with 2 changes:

1. Support codesigning macOS executables in bun build --compile
2. Deletes the hack where we read the last 4096 bytes of the currently running executables itself to see if it ends with "-- Bun --\n" . We now add a __BUN,__bun section in the macho executable so all we need to do is a single pointer lookup to find out if the current executable is a standalone executable and to get the data. This will also reduce memory usage because we no longer need to do another memory allocation to read through the standalone executable's code.

- Ensured that bun is able to generate binaries which are codesignable.

Fixes #7208

### DIFF
--- a/src/StandaloneModuleGraph.zig
+++ b/src/StandaloneModuleGraph.zig
@@ -793,7 +793,7 @@ pub const StandaloneModuleGraph = struct {
                     Global.exit(1);
                 },
             .{ .windows_hide_console = windows_hide_console },
-            target
+            target,
         );
         fd.assertKind(.system);
 
--- a/src/StandaloneModuleGraph.zig
+++ b/src/StandaloneModuleGraph.zig
@@ -12,6 +12,7 @@ const Syscall = bun.sys;
 const SourceMap = bun.sourcemap;
 const StringPointer = bun.StringPointer;
 
+const macho = bun.macho;
 const w = std.os.windows;
 
 pub const StandaloneModuleGraph = struct {
@@ -111,6 +112,23 @@ pub const StandaloneModuleGraph = struct {
         none = 0,
         esm = 1,
         cjs = 2,
+    };
+
+    const Macho = struct {
+        pub extern "C" fn Bun__getStandaloneModuleGraphMachoLength() ?*align(1) u32;
+
+        pub fn getData() ?[]const u8 {
+            if (Bun__getStandaloneModuleGraphMachoLength()) |length| {
+                if (length.* < 8) {
+                    return null;
+                }
+
+                const slice_ptr: [*]const u8 = @ptrCast(length);
+                return slice_ptr[4..][0..length.*];
+            }
+
+            return null;
+        }
     };
 
     pub const File = struct {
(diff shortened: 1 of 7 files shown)


Jarred-Sumner:
Wow nice find. The state of the Linux version of this was not ready - it had similar off-by-one issues somewhere - so we can't fully delete the original code.

Pranav2612000:
> Wow nice find. The state of the Linux version of this was not ready - it had similar off-by-one issues somewhere - so we can't fully delete the original code.

Oops. Do you have plans to look at this? If not, I can give the linux one a try as well.

Jarred-Sumner:
Judging by the tests it looks like there's some more work to do on the macOS parts before this can move forward

If you want to take the Linux one feel free but I suggest doing this one step at a time to make sure it ships

Pranav2612000:
Makes sense @Jarred-Sumner ! 
I'm working on getting the Mac version working first. 
How flaky are the tests in general? Asking because the same tests are passing locally.

I can see that the errors are probably due to storage full errors. Could a possible reason that the executable created using debug builds in this branch are larger?() . Although I tested out the production builds and both have the same size ~50MB for simple `hello-world`.

src/bun.js/bindings/c-bindings.cpp
```diff
@@ -880,3 +880,45 @@ extern "C" const char* BUN_DEFAULT_PATH_FOR_SPAWN = "C:\\Windows\\System32;C:\\W
 #else
 extern "C" const char* BUN_DEFAULT_PATH_FOR_SPAWN = "/usr/bin:/bin";
 #endif
+
+#if OS(DARWIN)
+
+#if CPU(ARM64)
+#define BLOB_HEADER_ALIGNMENT 16 * 1024
```
Pranav2612000, line 887: @Jarred-Sumner Do you think all MachO files should have the same blob header alignment? I couldn't find references which mentioned different blob header alignment size for intel and apple-chip macs

src/StandaloneModuleGraph.zig
```diff
@@ -597,71 +627,165 @@ pub const StandaloneModuleGraph = struct {
...
-                },
-            }
-        }
+                };
+                if (comptime !Environment.isWindows) {
+                    _ = bun.C.fchmod(cloned_executable_fd.int(), 0o777);
+                }
+                return cloned_executable_fd;
+            },
+            .linux => {
```
Pranav2612000: What are your thoughts on keeping Linux same as the older approach ( because the ELF modification is turning out to be tricky )? 
We can revisit the ELF approach again later.

pfgithub: It makes sense to go back too the old approach for linux for now

Pranav2612000: Done 💯 ! Thank you

Pranav2612000:
@Jarred-Sumner @pfgithub Can I get another review on this PR?
- I've ensured all the pipelines pass ( https://buildkite.com/bun/bun/builds/11974#01952b57-51cc-4cb1-a82d-442101eea844 except for some flaky issues )
- Squashed commits to have a better git history
- Tested both small and large mac executables are codesignable

Pranav2612000:
Thank you @pfgithub for fixing the lint errors. Let me know if there's something else I can help with.

pfgithub:
The PR is great, thanks! We will merge it soon

ctjlewis:
Great job, we had lots of people waiting on this, thanks so much!

PR #17210, opened 2025-02-09
fix [PR-15525]: use correct alignment to enable codesigning

Pranav2612000:
### What does this PR do?

This PR fixes an error with PR #15525 to allow codesigning larger files.
This is done by ensuring that the signature writer `sig_writer` starts from the original code segment location ( instead of with offset ). This way the whole signature fits in the `CODE_SIGNATURE` segment ( `SuperBlob` section ) and does not result in an error.

- Ensured that bun is able to generate binaries which are codesignable.


pfgithub:
This is superseeded by #17207 right?

PR #17218, opened 2025-02-10
feat(bun/test): Allow numbers, functions, and classes (anonymous and named) as first arg to `describe` blocks

### LINKED ISSUE #17100: Support functions as the description for describe

In Jest and Vitest we can write tests like this: 

```js
import { describe, expect, test } from 'vitest'
import { myFunctionUnderTest } from './sum.js'

//                 👇 - direct reference to the function, not a string. 
describe(myFunctionUnderTest, () =>{
  test('adds 1 + 2 to equal 3', () => {
    expect(myFunctionUnderTest(1, 2)).toBe(3)    
  })
})


```

And this will output like this: 

```
 ✓ src/sum.test.js (1 test) 1ms
   ✓ myFunctionUnderTest > adds 1 + 2 to equal 3
```

This is a nice bit of convenience for writing tests - 

1. Saves having to write a string ourselves
2. If we rename the function, the test description will automatically be updated

Support passing functions as description parameter in the same manner. 

_No response_

reillyjodonnell:
### What does this PR do?

Allows user to pass either a number, string, function, or class (anonymous and named) as the first argument to `describe.` fixes #17100, #7849

Before these changes if you pass a number, function, or class:
1. You get a TS error
2. Bun silently stops the test execution beyond the describe block and doesn't fail

- [x] Code changes

I wrote a passing test at `test/js/bun/test/describe.test.ts`


RiskyMH:
nice, i think you should also make a test that runts `bun test` as child process and ensure the function name was used correctly (and not just `[object object]` or smth)

reillyjodonnell:
> nice, i think you should also make a test that runs `bun test` as child process and ensure the function name was used correctly (and not just `[object object]` or smth)

Great suggestion - just added additional testing against the name

reillyjodonnell:
idk why this test is failing. I reverted all of the files I changed, rm  -rf'd`build` and rebuilt via `bun run build` and the test is still failing

test/js/bun/test/describe.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  describe("also here", () => {
+    test("Should also pass", () => {
+      stringBlockRan = true;
+      expect(true).toBeTrue();
+    });
+  });
+
+  test("both blocks should have run", () => {
+    expect(functionBlockRan).toBeTrue();
```
DonIsaac: use `jest.fn` to create a mock, then use `expect(fn).toHaveBeenCalled()`

src/bun.js/test/jest.zig
```diff
@@ -1725,7 +1725,7 @@ inline fn createScope(
     var function = if (args.len > 1) args[1] else .zero;
     var options = if (args.len > 2) args[2] else .zero;
 
-    if (description.isEmptyOrUndefinedOrNull() or !description.isString()) {
+    if (description.isEmptyOrUndefinedOrNull() or (description.isFunction() and description.getName(globalThis).isEmpty())) {
```
DonIsaac: What happens if 
```js
describe(MyClass, function testMyClass() {
  // ...
});
```

test/js/bun/test/describe.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(stringBlockRan).toBeTrue();
+  });
+});
+
+describe("shows function name correctly in test output", () => {
+  test("describe block shows function name correctly in test output", async () => {
+    const test_dir = tmpdirSync();
+    try {
+      await writeFile(
```
DonIsaac: You'll find `tempDirWithFiles` useful here

src/bun.js/test/jest.zig
RiskyMH: Also interested in how far Jest goes with this too

reillyjodonnell: I'm going to do a bit more digging to track how far Jest goes w it. After will update the code & do a little write up. Thanks yall!

reillyjodonnell:
Did some digging and wrote test cases based on my findings. This updates bun to accept functions, classes (both anonymous and named), strings, and numbers - just like jest!

@DonIsaac / @RiskyMH - thank you both for the comments!

reillyjodonnell:
updated `test/js/bun/test/test-test.test.ts` to be more resilient to stack trace changes in the `unhandled errors between tests are reported` describe block. We're matching against text to ensure the errors are still reported, instead of ensuring the stack trace is EXACTLY the same.

test/js/bun/test/describe.test.ts
```diff
@@ -0,0 +1,225 @@
...
+  });
+});
+
+describe("passing arrow function as args", () => {
+  test("passes if sole argument", () => {
+    const test_dir = tempDirWithFiles(".", {
+      "describe-test.test.js": `
+      import { describe, test, expect } from "bun:test";
+
+      describe(() => {
```
paperclover, line 174: im confused on the use case for this. it would make a little sense for named functions i guess. is this pattern supported and used by jest?

reillyjodonnell: I felt the same way. It is not supported by jest. The anonymous function as the sole argument pattern is used multiple times throughout [this other test file](https://github.com/oven-sh/bun/blob/11979f69ebee4ff1dc6c284430f7be045d67c3af/test/js/bun/test/test-test.test.ts#L475C1-L476C1)

Just wanted to keep things consistent. But happy to either fix all instances in this pr to remove it or if we want to tackle it in a separate one I can also do that :)

paperclover: ah, we already support it. i guess we should not remove it then.

paperclover (approved):

reillyjodonnell:
Screenshot 2025-03-06 at 9 23 52 PM
The only thing that failed is the lockfile for windows. This is ready to merge pending 1 more reapproval :D

Issue #17188, opened 2025-02-08
__dirname handled incorrectly when building


ficusk:
### What version of Bun is running?

1.2.0+b0c5a7655

### What platform is your computer?

Darwin 24.3.0 arm64 arm

### What steps can reproduce the bug?

Making references to __dirname from multiple different directories doesn't work correctly when running a js file emitted by `bun build`.

This shell script contains a complete repro:
```bash
#!/bin/bash

if [ -d "dirname-bug" ]; then
    echo "dirname-bug already exists"
    exit 1
fi

mkdir -p dirname-bug
cd dirname-bug

cat << EOF > apple.ts
import { orange } from "./subdir/orange";

console.log(\`\${__dirname}\`);
orange();
EOF

mkdir subdir

cat << EOF > subdir/orange.ts
export function orange() {
    console.log(\`\${__dirname}\`);
}
EOF

echo "bun run apple.ts"
bun run apple.ts

echo "bun build apple.ts --outdir dist --target bun"
bun build apple.ts --outdir dist --target bun

echo "bun run dist/apple.js"
bun run dist/apple.js
```

### What is the expected behavior?

Both the run version and the build version should output the same thing:

```
/Users/ficus/projects/bun-dirname-broken/dirname-bug
/Users/ficus/projects/bun-dirname-broken/dirname-bug/subdir
```

### What do you see instead?

The run version is correct but the built version is wrong:

```
❯ ./repro.sh
bun run apple.ts
/Users/ficus/projects/bun-dirname-broken/dirname-bug
/Users/ficus/projects/bun-dirname-broken/dirname-bug/subdir
bun build apple.ts --outdir dist --target bun

  ./apple.js  0.28 KB

[4ms] bundle 2 modules
bun run dist/apple.js
/Users/ficus/projects/bun-dirname-broken/dirname-bug
/Users/ficus/projects/bun-dirname-broken/dirname-bug
```

### Additional information

The problem is that the compiler is emitting a top level declaration for each reference to __dirname, so all references to __dirname evaluate to the last one in the file. Here's the output of `bun build` from my repro script:

```js
// @bun
// subdir/orange.ts
var __dirname = "/Users/ficus/projects/bun-dirname-broken/dirname-bug/subdir";
function orange() {
  console.log(`${__dirname}`);
}

// apple.ts
var __dirname = "/Users/ficus/projects/bun-dirname-broken/dirname-bug";
console.log(`${__dirname}`);
orange();
```

[...]

PR #17222, opened 2025-02-10
fix: __dirname and __filename collisions

amorriscode:
### What does this PR do?

This PR resolves issues with `__dirname` and `__filename` collisions (https://github.com/oven-sh/bun/issues/17188) by creating unique variables when `__dirname` or `__filename` are used in a bundle.

- [x] Used the debug build to test the fix

### Reproduction

When multiple files utilize `__dirname`/`__filename` Bun hoists the values to the top of the module code. However, since it is a `var` it can be overwritten effectively making it so only `__dirname`/`__filename` values collide.

#### Before

```
// test2/test.ts
var __dirname = "/Users/anthony/dev/test/test2", __filename = "/Users/anthony/dev/test/test2/test.ts";
function test() {
  console.log(__filename);
  console.log(__dirname);
}

// index.ts
var __dirname = "/Users/anthony/dev/test", __filename = "/Users/anthony/dev/test/index.ts";
console.log(__filename);
console.log(__dirname);
test();
```

In this single JS output, `__dirname` ends up evaluating to `/Users/anthony/dev/test` in both places it is called because that is the last value the var is assigned to.

#### After

The collision no longer occurs because usages of `__dirname` and `__filename` get unique values.

```
// test2/test.ts
var __dirname_af38e5cdfa6876c9 = "/Users/anthony/dev/test/test2", __filename_9b03b912cc9c01cd = "/Users/anthony/dev/test/test2/test.ts";
function test() {
  console.log(__filename_9b03b912cc9c01cd);
  console.log(__dirname_af38e5cdfa6876c9);
}

// index.ts
var __dirname_58f1efdc5b67fffe = "/Users/anthony/dev/test", __filename_c508108910e49d6d = "/Users/anthony/dev/test/index.ts";
console.log(__filename_c508108910e49d6d);
console.log(__dirname_58f1efdc5b67fffe);
test();
```

### DIFF
--- a/src/js_parser.zig
+++ b/src/js_parser.zig
@@ -3540,8 +3540,15 @@ pub const Parser = struct {
                 var declared_symbols = DeclaredSymbol.List.initCapacity(p.allocator, count) catch unreachable;
                 var decls = p.allocator.alloc(G.Decl, count) catch unreachable;
                 if (uses_dirname) {
+                    const unique_dir_name = try std.fmt.allocPrint(
+                        p.allocator,
+                        "__dirname_{x}",
+                        .{std.hash.Wyhash.hash(0, p.source.path.name.dir)},
+                    );
+                    const dir_ref = p.declareSymbol(.other, logger.Loc.Empty, unique_dir_name) catch unreachable;
+
                     decls[0] = .{
-                        .binding = p.b(B.Identifier{ .ref = p.dirname_ref }, logger.Loc.Empty),
+                        .binding = p.b(B.Identifier{ .ref = dir_ref }, logger.Loc.Empty),
                         .value = p.newExpr(
                             E.String{
                                 .data = p.source.path.name.dir,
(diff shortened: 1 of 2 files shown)


Jarred-Sumner (changes requested):
Thank you for this - the changes look like they work as expected, but I think we need to add an option to control the behavior of `__dirname` and `__filename` and similar. 

This collision is intentional and important for making things like `.node` addons bundle successfully. The behavior of `__dirname` and `__filename` when bundled is ambiguous.

robobun:
Thank you for this fix, and sorry that it sat for so long.

This branch patches `src/js_parser.zig`. That file no longer exists. The parser now lives in Rust under `src/js_parser/`, so this branch no longer applies to the current code and cannot be rebased.

#39714 addresses the same collision for the Rust parser. It keeps the inlined values and gives each module its own `__dirname` and `__filename` declaration name, which is the same result this PR produced with hashed names. The review note on this PR about an option is quoted on #39714, so the discussion continues there. Closing this one in its favor.

#17188 stays open. #39714 tracks it.

Issue #10437, opened 2024-04-22
Bun test hangs and fails when mocking rejected promises and asserting with Supertest


jacobwheale:
### What version of Bun is running?

1.1.4+fbe2fe0c3

### What platform is your computer?

Darwin 23.4.0 arm64 arm

### What steps can reproduce the bug?

When using an Express app and trying to test it with [supertest](https://github.com/ladjs/supertest), bun test hangs if a function within the tested express route is mocked to error:

```TypeScript
//functions.ts
export const testFunc = async () => {
  setTimeout(() => {}, 200);
  return "Hello from testFunc!";
};
```

```TypeScript
import { test, expect, spyOn } from "bun:test";
import express from "express";
import request from "supertest";
import * as funcs from "./functions";

const app = express();

app.get("/", async (req, res) => {
  try {
    const text = await funcs.testFunc();
    res.status(200);
    return res.send(text);
  } catch (err) {
    res.status(400);
    return res.send("Error");
  }
});

test("supertest", async () => {
  spyOn(funcs, "testFunc").mockRejectedValueOnce(new Error("error"));
  const res = await request(app).get("/").expect(400);

  expect(res.text).toBe("Error");
});

```

### What is the expected behavior?

The test should pass as the function being called inside of the endpoint is being mocked to reject, and the Express route inside of the catch block sets the res.status to 400 and returns "Error".

### What do you see instead?

The test hangs and the process doesn't complete:
![image](https://github.com/oven-sh/bun/assets/150444019/bd746a53-6831-4345-b0d2-d7c1dc7e7d70)

### Additional information

This style of tests works using Jest, I have been trying to convert an existing repo from Jest to bun test. if I change the spyOn to mock a different resolved value, the tests do not hang which makes me think it's related to the error being mocked or handling of errors.

taorepoara:
I have the same problem and it seems to work fine when using the watch mode: `bun test --watch`

mrcaidev:
Same problem here. Temporary workaround:

```ts
spyOn(funcs, "testFunc").mockImplementationOnce(() => {
  throw new Error("error");
});
```

gpfunk:
So it seems like if you don't include an `expect` around that function your mocking (or the one thats calling it internally), then it will create the error your talking about

But if you have an `expect()` around the function than it works as intended

```ts
  spyOn(spy2Fixture, "iSpy3").mockRejectedValue(new Error("42"));
[...]

PR #17258, opened 2025-02-11
fix mockRejectedValue throwing error

gpfunk:
### What does this PR do?

This lets `mockRejectedValue()` and `mockRejectedValueOnce()` not throw errors if they're not fulfilled / resolved immediately after being set. This would match jest's implementation

To do this I switched up the logic so rather than returning a rejected promise, it returns a function that returns a rejected promise

Fixes https://github.com/oven-sh/bun/issues/10437

I'm not that familiar with c++ or this codebase, so if theres a cleaner way to implement this please let me know! Thanks

- [x] Code changes

I wrote automated tests to ensure mocks that weren't called and fullfilled would not throw errors


pfgithub:
Fixed by #21489

PR #17267, opened 2025-02-11
node: fix Buffer.from(arrayBuffer)

nektro:
thanks @heimskr for the discovery!

### DIFF
--- a/test/js/node/buffer.test.js
+++ b/test/js/node/buffer.test.js
@@ -2980,3 +2980,25 @@ it("should not trim utf-8 start bytes at end of string", () => {
   const buf2 = Buffer.from("36e1", "hex");
   expect(buf2.toString("utf-8")).toEqual("6\uFFFD");
 });
+
+it("Buffer.from(arrayBuffer)", () => {
+  const ab = Buffer.from([10, 11, 12, 13, 14, 15, 16, 17, 18, 19]).buffer;
+  const buf = Buffer.from(ab);
+  expect(buf.length).toBe(10);
+  expect(buf.buffer).toBe(ab);
+  expect(buf[Symbol.iterator]().toArray()).toEqual([10, 11, 12, 13, 14, 15, 16, 17, 18, 19]);
+});
+it("Buffer.from(arrayBuffer, byteOffset)", () => {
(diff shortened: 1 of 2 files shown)


robobun:
<div><sup>Updated 3:18 PM PT - Feb 11th, 2025</sup></div>

@nektro, your commit ce5a61cf37f93e1ca3079000a4e95f746f883c80 is building: [`#11455`](https://buildkite.com/bun/bun/builds/11455)

test/js/node/buffer.test.js
```diff
@@ -2980,3 +2980,31 @@ it("should not trim utf-8 start bytes at end of string", () => {
...
+});
+it("Buffer.from(arrayBuffer, byteOffset, length)", () => {
+  const ab = Buffer.from([10, 11, 12, 13, 14, 15, 16, 17, 18, 19]).buffer;
+  const buf = Buffer.from(ab, 3, 5);
+  expect(buf.length).toBe(5);
+  expect(buf.buffer).toBe(ab);
+  expect(buf.byteOffset).toBe(3);
+  expect(buf.byteLength).toBe(5);
+  expect(buf[Symbol.iterator]().toArray()).toEqual([13, 14, 15, 16, 17]);
+});
```
190n, line 3010: Could we also test that, with code like

```js
class AB1 extends ArrayBuffer {}
function AB2() {}
Object.setPrototypeOf(AB2, ArrayBuffer);
Object.setPrototypeOf(AB2.prototype, ArrayBuffer.prototype);
```

`Buffer.from(new AB())` works and `Buffer.from(new AB2())` throws? (This seems to be the case both on Node and your PR, I just want to make sure it stays that way.)

190n (changes requested):
Mostly LGTM, I'd just like a bit more testing

test/js/node/buffer.test.js
nektro: https://github.com/oven-sh/bun/blob/78c1e3cfad192d9fc7e36460/test/js/node/buffer.test.js#L2140-L2145

190n: Oh, sorry, I thought those were still commented out because I was looking at a local clone of Bun from before #16731 which uncommented that test.

190n (approved):

Issue #17178, opened 2025-02-08
SQL: loses connection after timeout


jakeg:
### What version of Bun is running?

1.2.2+c1708ea6a

### What platform is your computer?

Linux 5.15.167.4-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

Sample code:

```js
import { sql } from 'bun'

Bun.serve({
  async fetch (req) {
    await sql`SELECT 'hello world'`
    return new Response('hi')
  }
})
```

Using a `.env` file with a `POSTGRES_URL` that connects to either Supabase's "Session pooler" or their "Transaction pooler".

Works fine until you don't do any requests (and thus SQL queries) for ~30+ seconds, then next time you try a request I presume the pool of connections have been closed due to an idle timeout, but no attempt is made to reconnect. The request hangs then this error appears:

```
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
```

Am I supposed to do something manually to create new connections in the pool? I assumed that would be automatic.

Only way to fix is to restart the server, then of course it will happen again if no queries for 30+ secondds.

jakeg:
Here's some code which shows this with just a timeout and no server needed. It seems like it's 60s+ not 30s that causes the issue:

```js
import { sql } from 'bun'

let timeout = 65
let count = 0
let query = async () => {
  console.log(`Trying ${++count}...`)
  await sql`SELECT 'hello world'`
  console.log('worked.')
}

await query()
console.log(`Trying again in ${timeout}s`)
setTimeout(query, timeout * 1_000)
```

RiskyMH:
@jakeg can you try canary as I believe this has been fixed.

```zig
bun upgrade --canary
```

jakeg:
Just upgraded (`1.2.3-canary.48+584db03a7`) and it's still a bug.

cirospaciari:
The error:
```bash
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
```

In this case`idleTimeout` is from `Bun.serve` take a look in https://bun.sh/docs/api/http#idletimeout and https://bun.sh/docs/api/http#server-timeout-request-seconds-custom-request-timeouts

SQL connection should reconnect if an `idleTimeout` happens but its not the error that showed. 
For the SQL it self you can increase the `maxLifetime` of the connection and `idleTimeout` so the query can be completed. See https://bun.sh/docs/api/sql#connection-options.

```ts
[...]

Issue #16691, opened 2025-01-24
ERR_POSTGRES_CONNECTION_CLOSED


Nabwin:
### What version of Bun is running?

1.2.0

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

```ts

import { SQL } from "bun";

export const db = new SQL({
  url: "postgres://postgres:nabin@localhost:5433/refresher",

  max: 20, // Maximum connections in pool
  idleTimeout: 30, // Close idle connections after 30s

  onconnect: (client) => {
    console.log("Connected to database");
  },
  onclose: (client) => {
    console.log("Connection closed");
  },
});

```

```ts
import { sql } from "bun";
import { db } from "./db";

db.connect()
  .then(() => {
    console.log("Connected to database");
  })
  .catch((err) => {
    console.log("Error connecting to database", err);
  });

const users = await sql`
  SELECT * FROM users
  LIMIT ${10}
`;

console.log("active Users:", users);

```
```
➜  bun-1.2 bun run index.ts
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Connection closed
Error connecting to database PostgresError: Connection closed
 code: "ERR_POSTGRES_CONNECTION_CLOSED"


PostgresError: Connection closed
 code: "ERR_POSTGRES_CONNECTION_CLOSED"
```
i have my postgres container running in port 5433 

## While using the pg package it work success fully

database.ts
```ts

import { Client } from "pg";

const connectionString =
  "postgres://postgres:postgres@localhost:5433/refresher";

const client = new Client({
  connectionString: connectionString,
});

client
  .connect()
  .then(() => console.log("Connected to database"))
  .catch((err: unknown) => console.error("Error connecting to database", err));

export { client };

```

index.ts

```ts
import { sql } from "bun";
import { db } from "./db";
import { client } from "./database";

// db.connect()
//   .then(() => {
//     console.log("Connected to database");
//   })
//   .catch((err) => {
//     console.log("Error connecting to database", err);
//   });

// const users = await sql`
//   SELECT * FROM users
//   LIMIT ${10}
// `;

client.query("SELECT * FROM users LIMIT 10", (err, res) => {
  if (err) {
    console.error("Error executing query", err);
    return;
[...]

Issue #11464, opened 2024-05-30
Dead code elimination strips out code which is actually not dead


sedubois:
### What version of Bun is running?

1.1.10

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

- clone https://github.com/sedubois/bun-dead-code-bug/tree/670cfa650f32429b754bdf846213e368526d3f3e
- `bun run build`
- open index.html in a browser
- open the JS console

### What is the expected behavior?

No JS error.

### What do you see instead?

There is an error `Uncaught TypeError: class heritage EVENT is not an object or null`.

<img width="631" alt="Screenshot 2024-05-30 at 18 29 44" src="https://github.com/oven-sh/bun/assets/4217871/f84d9d30-e3ce-4f0a-8ec0-7a6682965307">

The error is raised in maverick.js, a transitive dependency of Vidstack. An `EVENT` constant is defined just before the place where it is referenced `maverick/packages/maverick/src/std/event.ts`:
https://github.com/maverick-js/maverick/blob/d72709f1e367265ab2465fa153c25d72d0b2c709/packages/maverick/src/std/event.ts#L17

 The EVENT constant referred to in the error should be kept by bun in the bundle.

### Additional information

I first [reported](https://discord.com/channels/742612686679965696/1237413629989621913/1237413629989621913) this to Vidstack, who responded _"This would be an issue with Bun, not Vidstack. It seems like they're too aggressively eliminating dead code and not checking if a variable is used in a class heritage."_

paperdave:
Fixed in #12729

PR #17272, opened 2025-02-12
fix(sql) fix ready for query state

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/17178
Fix: https://github.com/oven-sh/bun/issues/16691

After investigating was discovered that supabase returns an additional ParameterStatus with value "S\x00\x00\x00\x1fapplication_name\x00Supavisor\x00" informing the application name, with makes we mark the connection as `is_ready_for_query = false` when we should not, we only need to mark as `is_ready_for_query = false` when writting .Sync.

In the case of the Issue `16691` the connection was in a broken state until the server closed it, but the cause was `is_ready_for_query` was false when should not be.

- [X] Code changes

Manually

### DIFF
--- a/src/sql/postgres.zig
+++ b/src/sql/postgres.zig
@@ -718,12 +718,14 @@ pub const PostgresSQLQuery = struct {
                         if (!connection.hasQueryRunning()) {
                             this.flags.binary = this.statement.?.fields.len > 0;
                             debug("bindAndExecute", .{});
+
                             // bindAndExecute will bind + execute, it will change to running after binding is complete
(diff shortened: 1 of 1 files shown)


robobun:
<div><sup>Updated 4:41 PM PT - Feb 11th, 2025</sup></div>

@cirospaciari, your commit 9f3925081677cfe89fbe0723c30c8361aa03520d is building: [`#11464`](https://buildkite.com/bun/bun/builds/11464)

PR #17276, opened 2025-02-12
Remove `@intFromPtr(...) == 0`

pfgithub:
A pointer without allowzero can never be null, so this is a no-op


robobun:
<div><sup>Updated 5:38 PM PT - Feb 11th, 2025</sup></div>

@pfgithub, your commit 77b60a3390eec0774c6f6476d569102245b1499e is building: [`#11472`](https://buildkite.com/bun/bun/builds/11472)

Issue #17044, opened 2025-02-04
Add support for PgBouncer transaction pooling mode by allowing to disable prepared statements


incpo:
### What is the problem this feature would solve?

When trying to use Bun's SQL client with PgBouncer in transaction pooling mode, the connection fails because PgBouncer doesn't support prepared statements. Currently, there's no way to disable prepared statements in Bun's SQL client.

### What is the feature you are proposing to solve the problem?

Add configuration options to the Database constructor to disable prepared statements:

### What alternatives have you considered?

1. Using PgBouncer in session pooling mode instead of transaction pooling mode
   - Less performant
   - Doesn't scale
   - Defeats the purpose of using PgBouncer for connection pooling

2. Connecting directly to PostgreSQL (port 5432)
   - Removes the benefits of connection pooling
   - Not suitable for high-scale applications
   - Increases database connection load

3. Using different database clients
   - Loses Bun's performance benefits
   - Requires changing existing codebase
   - Adds unnecessary dependencies

nicksrandall:
This *could* be solved by #17029

Issue #17029, opened 2025-02-03
SQL: support "simple", multi-statement queries.


nicksrandall:
### What version of Bun is running?

1.2.2

### What platform is your computer?

Darwin 24.3.0 arm64 arm

### What steps can reproduce the bug?

Bun's SQL implementation should only convert a query to a prepared statement *if* it has at least one variable (parameter). This is the only situation in where a prepared statement is needed. Otherwise, the client should just use a "simple" query which has the advantage of allowing multiple statements.

### What is the expected behavior?

I would expect to be able to support multiple statements in a single query (very common for migration scripts). 

### What do you see instead?

I get the following error: `cannot insert multiple commands into a prepared statement`

### Additional information

IMO, this is an essential feature for a SQL client. It's very cumbersome to do db migrations without this.

PR #17296, opened 2025-02-13
fix(sql) add `prepare: false` option and sql``.simple()

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/17029
Fix: https://github.com/oven-sh/bun/issues/17044

- [x] Code changes

Tests

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -2054,6 +2054,8 @@ declare module "bun" {
     max?: number;
     /** By default values outside i32 range are returned as strings. If this is true, values outside i32 range are returned as BigInts. */
     bigint?: boolean;
+    /** Automatic creation of prepared statements, defaults to true */
+    prepare?: boolean;
   };
 
   /**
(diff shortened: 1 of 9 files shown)


robobun:
<div><sup>Updated 7:08 PM PT - Feb 12th, 2025</sup></div>

@cirospaciari, your commit c151d167a9c03e1f1eb8282ec8739a87ece97dd4 is building: [`#11532`](https://buildkite.com/bun/bun/builds/11532)

Jarred-Sumner:
are these flaky tests related?
image

cirospaciari:
> are these flaky tests related?
> image
> 

Looks like it is due to transaction pooling behavior, need to see what tests are flaky to fix or skip the test

Jarred-Sumner:
The only part I'm not sure about is if we should make it simple when there are no params.

cirospaciari:
> The only part I'm not sure about is if we should make it simple when there are no params. 

Postgres.js don't make it simple if they have no parameters, but if it's unsafe and has no parameters it's simple, performance is better when it is not a simple query when running the query multiple times.

Issue #11568, opened 2024-06-03
Bun crashes in `Bun::formatStackTrace`


KevinArce:
### How can we reproduce the crash?

I just ran as usual my project with Dockerfile:

```dockerfile
# Use the official Bun Docker image
FROM oven/bun

# Create app directory
WORKDIR /usr/src/app

# Install app dependencies by copying package.json, package-lock.json, and bun.lockb
# and then running bun install. Note that if you don't have package-lock.json, you can omit it.
COPY package*.json bun.lockb ./
RUN bun install --production

# Bundle app source by copying the rest of your app's source code
COPY . .

# Build the project according to the "build" script in package.json
RUN bun run build

# Your app binds to port 8080 so you'll use the EXPOSE instruction to have it mapped by the docker daemon
EXPOSE 8080

# Define environment variable
ENV NODE_ENV production
ENV PORT 8080

# Define the command to run your app using the "start" script from your package.json
CMD [ "bun", "run", "start" ]
```

And this is my package.json:
```json
{
  "name": "bun-bug",
  "version": "1.0",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "dev": "bun run --watch src/index.ts",
    "build": "bun build ./src/*.ts ./src/**/*.ts --outdir ./dist --target bun --sourcemap=external --minify",
    "start": "bun run build && bun run dist/src/index.js --watch"
  },
  "dependencies": {
    "@bogeychan/elysia-logger": "^0.0.22",
    "@elysiajs/cors": "^1.0.2",
    "@elysiajs/swagger": "^1.0.5",
    "@sentry/bun": "^8.7.0",
    "@types/luxon": "^3.4.2",
    "elysia": "^1.0.22",
    "jsonwebtoken": "^9.0.2",
    "luxon": "^3.4.4",
    "mongoose": "^8.4.1",
    "openai": "^4.47.3",
    "winston": "^3.13.0",
    "winston-daily-rotate-file": "^5.0.0"
  },
  "devDependencies": {
    "@types/jsonwebtoken": "^9.0.6",
    "bun-types": "^1.1.12"
  },
  "module": "src/index.js"
}
```

### Relevant log output

```shell
error: script "start" was terminated by signal SIGILL (Illegal instruction)
============================================================
Bun v1.1.12 (43f0913c) Linux x64 (baseline)
Args: "bun" "run" "start"
Features: spawn 
Elapsed: 15749ms | User: 9ms | Sys: 23ms
[...]

PR #17316, opened 2025-02-13
fix(napi): memory leak when creating strings

DonIsaac:
### What does this PR do?
Fixes two bugs in NAPI:
1. `napi_create_string_utf16` not `deref`ing newly created strings, leading to a memory leak
2. `napi_create_string_utf16` not correctly returning when passed a 0-length buffer. This would cause `createUninitialized` to get called, triggering an assertion failure.

I added a test

### DIFF
--- a/src/napi/napi.zig
+++ b/src/napi/napi.zig
@@ -340,15 +340,8 @@ pub export fn napi_create_string_latin1(env: napi_env, str: ?[*]const u8, length
 
     log("napi_create_string_latin1: {s}", .{slice});
 
-    if (slice.len == 0) {
-        result.set(env, bun.String.empty.toJS(env.toJS()));
-        return env.ok();
-    }
-
-    var string, const bytes = bun.String.createUninitialized(.latin1, slice.len);
+    var string = bun.String.createWTF(.latin1, slice) catch bun.outOfMemory();
     defer string.deref();
-
-    @memcpy(bytes, slice);
 
     result.set(env, string.toJS(env.toJS()));
     return env.ok();
@@ -401,12 +394,8 @@ pub export fn napi_create_string_utf16(env: napi_env, str: ?[*]const char16_t, l
     if (comptime bun.Environment.allow_assert)
         log("napi_create_string_utf16: {d} {any}", .{ slice.len, bun.fmt.FormatUTF16{ .buf = slice[0..@min(slice.len, 512)] } });
 
-    if (slice.len == 0) {
-        result.set(env, bun.String.empty.toJS(env.toJS()));
-    }
-
-    var string, const chars = bun.String.createUninitialized(.utf16, slice.len);
-    @memcpy(chars, slice);
+    var string = bun.String.createWTF(.utf16, slice) catch bun.outOfMemory();
+    defer string.deref();
 
     result.set(env, string.transferToJS(env.toJS()));
     return env.ok();
--- a/test/napi/napi-app/main.cpp
+++ b/test/napi/napi-app/main.cpp
@@ -6,6 +6,7 @@
 #include <cassert>
 #include <cinttypes>
 #include <cmath>
+#include <codecvt>
 #include <cstdarg>
 #include <cstdint>
 #include <cstdio>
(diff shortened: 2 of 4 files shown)


robobun:
<div><sup>Updated 1:26 PM PT - Feb 13th, 2025</sup></div>

@DonIsaac, your commit a634127dca22b1b8016a95c7a1dc53f0f63b1e4b is building: [`#11568`](https://buildkite.com/bun/bun/builds/11568)

test/napi/napi-app/main.cpp
```diff
@@ -155,6 +156,53 @@ test_napi_get_value_string_utf8_with_buffer(const Napi::CallbackInfo &info) {
...
+  const size_t BUFSIZE = 256;
+  Napi::Env env = info.Env();
+
+  napi_value *result = nullptr;
+
+  // test with a string that will definitely end up utf16-encoded.
+
+  std::u16string cpp_str_utf16 = u"Bun's napi support is 👌";
+
+  assert(napi_create_string_utf16(env, cpp_str_utf16.data(),
```
190n, line 169: For new Node-API tests use the macro `NODE_API_CALL(env, ...)` rather than `assert(... == napi_ok)`. This throws a JS exception including the error code instead of crashing the process.

test/napi/napi-app/main.cpp
```diff
@@ -155,6 +156,53 @@ test_napi_get_value_string_utf8_with_buffer(const Napi::CallbackInfo &info) {
   return ok(env);
 }
 
+napi_value test_napi_create_string_utf16(const Napi::CallbackInfo &info) {
```
190n, line 159: This function needs to be registered in `InitAll` otherwise it won't be included in the JS object this module exports so `main.js` won't be able to call it.

test/napi/napi-app/main.cpp
```diff
@@ -155,6 +156,53 @@ test_napi_get_value_string_utf8_with_buffer(const Napi::CallbackInfo &info) {
...
+
+  napi_value *result = nullptr;
+
+  // test with a string that will definitely end up utf16-encoded.
+
+  std::u16string cpp_str_utf16 = u"Bun's napi support is 👌";
+
+  assert(napi_create_string_utf16(env, cpp_str_utf16.data(),
+                                  cpp_str_utf16.length(), result) == napi_ok);
+  assert(result != nullptr);
```
190n, line 171: This will fail, `napi_create_string_utf16` takes a pointer parameter by value so it can't change `result` to something other than `nullptr`. The usual pattern is:
```
napi_value result;
NODE_API_CALL(env, napi_whatever(env, &result));
```

190n (changes requested):
The runtime changes seem fine but I have some comments on the tests. More comments below which don't apply to specific changed lines:

I see `napi_create_string_utf8` still does not `deref` the string it creates -- does this function need to be changed, too, or is there a reason `deref` isn't required there?

Can we test that the string-creation functions do not leak?

Finally, we're planning to merge #14501 after the next release. Unless that plan changes I think we should hold off on merging this or any other Node-API changes until after that PR lands. Merging `main` into your branch after that is landed will take a bit of care because that branch splits up `test/napi/napi-app/main.cpp` into a bunch of smaller files -- I can handle doing that.

190n:
Is this fixing a real leak? I wrote a test that loops creating 10 million character UTF-16 strings with Node-API while running GC occasionally and the memory usage is steady on Bun canary.

src/napi/napi.zig
```diff
@@ -401,12 +394,8 @@ pub export fn napi_create_string_utf16(env: napi_env, str: ?[*]const char16_t, l
...
-    if (slice.len == 0) {
-        result.set(env, bun.String.empty.toJS(env.toJS()));
-    }
-
-    var string, const chars = bun.String.createUninitialized(.utf16, slice.len);
-    @memcpy(chars, slice);
+    var string = bun.String.createWTF(.utf16, slice) catch bun.outOfMemory();
+    defer string.deref();
 
     result.set(env, string.transferToJS(env.toJS()));
```
Jarred-Sumner, line 400: This derefs the string, it's not a leak

PR #16855, opened 2025-01-29
node:timers fixes

190n:
### What does this PR do?

Progress on passing `node:timers` tests.

- Make `setImmediate` return a different kind of object than `setTimeout`/`setInterval`
- Handle edge cases in the number of milliseconds for timers the same way Node does
- Make unref'd `setImmediate`s not keep the event loop alive
- Emit warnings for invalid `after` parameters
- `clearTimeout` and `clearInterval` accept stringified IDs
- New data structure to ensure timers run in the same order as Node.js
- Getter for `_destroyed` property

Fixes https://github.com/oven-sh/bun/issues/5767

### How did you verify your code works?

Running Node tests

PR #17346, opened 2025-02-14
Handle rejected promises in `EventLoop.autoTick`

190n:
### What does this PR do?

Adds a call to `this.global.handleRejectedPromises()` inside `EventLoop.autoTick()`. This fixes a failure exposed on #16855. That PR happened to change `Bun.sleep(0)` to sleep for 0ms instead of 1ms, which made `html-rewriter.test.js` fail because if the `element` callback rejects too quickly, the exception would seem to be handled twice (the `catch` block would run, but the exception would also appear globally as an uncaught rejected promise).

I changed the test to use `await setImmediatePromise()` instead of `await Bun.sleep(0)`, which exposes the problem without depending on any timers changes, and I fixed this issue by adding the `handleRejectedPromises()` call.

Ran the `HTMLRewriter` test

### DIFF
--- a/src/bun.js/event_loop.zig
+++ b/src/bun.js/event_loop.zig
@@ -1514,6 +1514,7 @@ pub const EventLoop = struct {
 
         this.flushImmediateQueue();
         ctx.onAfterEventLoop();
+        this.global.handleRejectedPromises();
     }
 
     pub fn flushImmediateQueue(this: *EventLoop) void {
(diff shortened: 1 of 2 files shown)


robobun:
<div><sup>Updated 1:19 PM PT - Feb 14th, 2025</sup></div>

@190n, your commit 8d9ead77654e49f16d33ec3ff566ff9283d3a2ba is building: [`#11621`](https://buildkite.com/bun/bun/builds/11621)

Issue #11651, opened 2024-06-06
Bun Segementation Fault in Production


vlucas:
### How can we reproduce the crash?

Deploy this and refresh a bunch of times: https://github.com/vlucas/hyperspan

### Stack Trace (bun.report)

Bun v1.1.12 ([`43f0913`](<https://github.com/oven-sh/bun/tree/43f0913c3875cbf260c60868268adc5953b09a80>)) on linux x86_64_baseline [RunCommand]

Segmentation fault at address 0x00000000

- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/ld-temp.o#L0>): `WTF::jscSignalHandler`
- `??`
- [`StringImpl.h:1154`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80//build/bun/bun-webkit/include/wtf/text/StringImpl.h#L1154>): `Bun__WTFStringImpl__deref`
- [`string.zig:103`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/string.zig#L103>): `MatchedRouteClass__finalize`
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/ld-temp.o#L0>): `JSC::PreciseAllocation::sweep`
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/ld-temp.o#L0>): `JSC::Heap::handleNeedFinalize`
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/ld-temp.o#L0>): `JSC::Heap::finishChangingPhase`
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/ld-temp.o#L0>): `JSC::Heap::runEndPhase`
- [`ld-temp.o:0`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/ld-temp.o#L0>): `JSC::Heap::runCurrentPhase`
[...]

PR #17352, opened 2025-02-15
asan integration enhancements

nektro:
- improves option parsing for distinguishing between asan and non-asan builds in debug mode
- makes app code use system malloc so asan override is able to detect it more consistently 
- tells mimalloc to override `malloc`/`free` when enabled to work with ^


robobun:
<div><sup>Updated 9:38 PM PT - Feb 14th, 2025</sup></div>

@nektro, your commit 0e60df536fedb08a040065ab7c9e20bbe5acd454 is building: [`#11651`](https://buildkite.com/bun/bun/builds/11651)

Issue #11656, opened 2024-06-06
Write file from stream using thread worker


vitaliy-melnik:
### Stack Trace (bun.report)

Bun v1.1.10 ([`5102a94`](<https://github.com/oven-sh/bun/tree/5102a944307d56933a6bb039b99ad76a758b46e4>)) on macos aarch64 [RunCommand]

Segmentation fault at address 0x100009C5F

- *4 unknown/js code*

Jarred-Sumner:
Likely fixed in Bun v1.1.13 with the other Worker-related fixes

PR #17353, opened 2025-02-15
Chloe/hmr4

### LINKED ISSUE #17329: CSS `@import` and `url()` do not work in HMR
`url()` and `@import` in CSS files with HMR returns `import_record.path.text` instead of the unique_key. This causes it to output the absolute file path on disk in the output css file, which the server cannot serve (nor should it)

Input:
```css
body::before {
  content: "";
  position: fixed;
  inset: 0;
  z-index: -1;
  opacity: 0.05;
  background: url("./logo.svg");
  background-size: 256px;
  transform: rotate(-10deg) scale(1.4);
  animation: slide 120s linear infinite;
  pointer-events: none;
}
```

Output:
```css
body:before {
  content: "";
  position: fixed;
  z-index: -1;
  opacity: .05;
  animation: slide 120s linear infinite;
  pointer-events: none;
  background: url("/private/tmp/ok/src/logo.svg") 0 0 / 256px;
  inset: 0;
  transform: rotate(-10deg)scale(1.4);
}

```

Ideally, it would use source_index like JS. Then, we can defer worrying about what the correct URL is to the end and rewrite at the end all at once. We shouldn't be overwriting `source.path.text` either (unless it's to reuse an existing allocation of the same contents and free that one).

### LINKED ISSUE #17330: Directory watcher does not handle CSS assets in HMR correctly
1. Have a successful build (which includes CSS)
2. Create a new css file in the same source directory as a previously built .css file
3. Add a `@import './another.css'`
4. You get a module resolution error until you press refresh

paperclover:
Fixes #17330
Fixes #17329

note: css printer takes in ImportInfo which contains import records + some bundler information, instead of just import records. this prevents mutating path.text while bundling.

### DIFF
--- a/src/bake/DevServer.zig
+++ b/src/bake/DevServer.zig
@@ -4226,6 +4226,21 @@ pub fn IncrementalGraph(side: bake.Side) type {
                         // knowledge of the boundary between them.
                         .js, .unknown => if (!data.flags.is_hmr_root) {
                             try entry_points.appendJs(alloc, path, .client);
+
+                            // TODO(@paperclover): do not loop here
+                            var it = g.first_dep.items[index].unwrap();
+                            while (it) |edge_index| {
+                                const entry = g.edges.items[edge_index.get()];
+                                const dep = entry.dependency;
+                                g.stale_files.set(dep.get());
+
+                                const dep_file = values[dep.get()];
+                                if (dep_file.flags.is_css_root) {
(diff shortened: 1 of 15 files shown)


robobun:
<div><sup>Updated 12:30 AM PT - Feb 15th, 2025</sup></div>

@paperclover, your commit 3cf8ee761268bcdecfab6b7cb9e4f208bd161b0e is building: [`#11656`](https://buildkite.com/bun/bun/builds/11656)

Issue #11660, opened 2024-06-06
`bun test` ignores tests in subsequent files when using a test factory


Hexagon:
### What version of Bun is running?

1.1.12+43f0913c3

### What platform is your computer?

Linux 5.15.0-107-generic x86_64 x86_64

### What steps can reproduce the bug?

1. Create  **testfactory.js:**

```javascript
import { test } from "bun:test";
export { test }; 
```

2. Create two or more test files **test1.test.js and test2.test.js:**

```javascript
import { test } from "./testfactory.js";

test('add1', async() => { /* ... */ });
test('add2', async() => { /* ... */ });
```

```javascript
import { test } from "./testfactory.js";

test('sub1', async() => { /* ... */ });
test('sub2', async() => { /* ... */ });
```

3. Run `bun test`

### What is the expected behavior?

All tests from all test files should be executed.

### What do you see instead?

Only the tests from the first file (`test1.test.js` in this case) are run. The tests from the second file are not discovered or executed.

```
test2.test.js:
✓ sub1 [0.04ms]
✓ sub2 [0.04ms]

test1.test.js: 

 2 pass
 0 fail
Ran 2 tests across 2 files. [23.00ms]
```

### Additional information

This issue does not occur if I import the `test` function directly from "bun:test" in each test file, instead of using my test factory (which is just a re-exported `test` in this example).

Pinta365:
Same problem for me on Windows.

Bun version:
1.1.13

OS:
Windows 11 Pro

lowlighter:
Can also confirm this still happens on 1.1.37

This seems to be a common issue unfortunately 😕 
- #7392
- #5400
- #3965

However, thanks to @benmerckx's findings about the fact that calling `Bun.jest(path)` creates a new `test()` function that somehow is able to register tests correctly when called, it's possible to exploit the stack trace to find back where your wrapper function was called and make it "transparent" (though it's still hacky)...

Basically you need something like:
```ts
/** Your test runner wrapper */
export function myTest(name: string, fn: Function) {
  return Bun.jest(caller()).test(name, fn)
}

/** Retrieve caller test file. */
function caller() {
  const Trace = Error as unknown as { prepareStackTrace: (error: Error, stack: CallSite[]) => unknown }
  const _ = Trace.prepareStackTrace
  Trace.prepareStackTrace = (_, stack) => stack
  const { stack } = new Error()
  Trace.prepareStackTrace = _
  const caller = (stack as unknown as CallSite[])[2]
[...]

PR #17357, opened 2025-02-15
Add routes to Bun.serve()

Jarred-Sumner:
### What does this PR do?

This adds routing support to Bun.serve(). 

```ts
import { serve, sql, type BunRequest } from "bun";
import App from "./myReactSPA.html";

serve({
  port: 3000,

  routes: {
    "/*": App,
    "/api/users": async () => Response.json(await sql`SELECT * FROM users LIMIT 10`),
    "/api/users/:id": async req => {
      const [user] = await sql`SELECT * FROM users WHERE id = ${req.params.id} LIMIT 1`;
      if (!user) {
        return new Response("User not found", { status: 404 });
      }
      return Response.json(user);
    },

    // req.params is inferred from the route path, but you can still specify them explicitly
    "/api/users/:id/posts": async (req: BunRequest<"/api/users/:id/posts">) =>
      Response.json(await sql`SELECT * FROM posts WHERE user_id = ${req.params.id}`),
  },

  fetch: () => new Response("fallback"),
});
```
Tests


robobun:
<div><sup>Updated 3:31 AM PT - Feb 15th, 2025</sup></div>

@Jarred-Sumner, your commit 726adb0e7bbcf7a9e56a87fa9985b4684b05806b is building: [`#11660`](https://buildkite.com/bun/bun/builds/11660)

jakeg:
I would expect this to work:

```js
Bun.serve({
  routes: {
    '/': new Response('hello world')
  }
})
```

But it gives error `Expected fetch() to be a function`, thus you need:

```js
Bun.serve({
  routes: {
    '/': new Response('hello world')
  },
  fetch: () => new Response('hi')
})
```

jakeg:
Surely with:

```js
routes: {
    "/*": App,
    "/api/users": ...
}
```
... the `/*` would be greedy so must go after the other routes?

jakeg:
2 more questions:

- what about GET vs POST (and other verbs)?
- what about the equivalent of `app.use('/*', serveStatic({ root: './web' }))` in Hono, which would serve the contents of a directory

TiBianMod:
verbs are supported, take a look the tests


```ts
Bun.serve({
  port: 0,
  fetch: () => new Response("fallback"),
  routes: {
    "/api": {
      GET: () => new Response("GET"),
      POST: () => new Response("POST"),
      PUT: () => new Response("PUT"),
      DELETE: () => new Response("DELETE"),
      PATCH: () => new Response("PATCH"),
      OPTIONS: () => new Response("OPTIONS"),
      HEAD: () => new Response("HEAD"),
    },
  },
});
```

yes, will be nice to have the `serve assets from directory`

src/bun.js/bindings/decodeURIComponentSIMD.cpp
```diff
@@ -0,0 +1,315 @@
...
+                if ((byte & 0xC0) != 0x80) {
+                    result.append(replacementChar);
+                    continue;
+                }
+                value = (value << 6) | (byte & 0x3F);
+                cursor += 3;
+
+                // Check for overlong encoding and surrogate range
+                if (value < 0x800 || value > 0xFFFF || (value >= 0xD800 && value <= 0xDFFF) || // Surrogate range check
+                    (byte == 0xE0 && (value & 0x1F00) == 0)) // Overlong check for E0
```
pfgithub, line 159: What is this check intended to be for? byte can never be 0xE0 because above it is checked if `(byte & 0xC0) != 0x80`

PR #17367, opened 2025-02-15
refactor: remove BunString.fromJSRef

DonIsaac:
### What does this PR do?
Removes `bun.String.fromJSRef`. This method is a footgun. I've left `Bun::toStringRef` as is.

- [ ] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [x] Code changes

### How did you verify your code works?
Changes are covered by existing tests.

PR #17369, opened 2025-02-15
refactor: move string-like structs into `string` module

DonIsaac:
### What does this PR do?
- Moves various string data structures into a new `string` module
- Tweak return signatures of some allocating methods to specify errors they may return. That is, `fn foo() !SomeType` -> `fn foo() Allocator.Error!SomeType`
- Includes changes from #17367, which should be merged first

This PR has no business logic changes, so if it builds and existing tests pass, this works.

### Other Notes
- I'd like to move `ZigString` as well, but thats a pretty large refactor on its own and is likely better left to another pull request
- We should consider removing the `bun.string` type completely, and have `bun.string` be this module. Then we can just re-export it wholesale from `bun.zig`.


robobun:
<div><sup>Updated 1:41 PM PT - Feb 15th, 2025</sup></div>

@DonIsaac, your commit 62cb87b915e5a6d33fa39ed59997a929a1bfbfe9 is building: [`#11675`](https://buildkite.com/bun/bun/builds/11675)

PR #17378, opened 2025-02-16
Try mimalloc v3

Jarred-Sumner:
(For internal tracking: fixes ENG-19852)


robobun:
<div><sup>Updated 12:18 AM PT - Feb 16th, 2025</sup></div>

@Jarred-Sumner, your commit 6145469875be6b58901c85160e88fdb938411eb4 is building: [`#11693`](https://buildkite.com/bun/bun/builds/11693)

src/allocators/basic.zig
```diff
@@ -25,22 +24,22 @@ fn mimalloc_free(
...
         if (comptime Environment.enable_logs)
             log("mi_alloc({d}, {d})", .{ len, alignment.toByteUnits() });
 
-        const ptr: ?*anyopaque = if (mimalloc.canUseAlignedAlloc(len, alignment.toByteUnits()))
-            mimalloc.mi_malloc_aligned(len, alignment.toByteUnits())
-        else
-            mimalloc.mi_malloc(len);
+        // The posix_memalign only accepts alignment values that are a
+        // multiple of the pointer size
+        const effective_alignment = @max(alignment.toByteUnits(), @sizeOf(usize));
```
Jarred-Sumner: can we revert this before merging

src/allocators/MimallocArena.zig
```diff
@@ -75,7 +75,7 @@ fn alignedAlloc(heap: *mimalloc.Heap, len: usize, alignment: mem.Alignment) ?[*]
     const ptr: ?*anyopaque = if (mimalloc.canUseAlignedAlloc(len, alignment.toByteUnits()))
         mimalloc.mi_heap_malloc_aligned(heap, len, alignment.toByteUnits())
     else
-        mimalloc.mi_heap_malloc(heap, len);
+        mimalloc.mi_heap_malloc_aligned_at(heap, len, alignment.toByteUnits(), 0);
```
Jarred-Sumner: can we revert this

src/js_printer.zig
```diff
@@ -5400,6 +5380,10 @@ pub fn NewWriter(
             return this.ctx.getMutableBuffer();
         }
 
+        pub fn takeBuffer(this: *Self) MutableString {
```
taylordotfish, line 5383: You added this method but it's never used

src/js_printer.zig
```diff
@@ -6011,9 +6004,12 @@ pub fn printWithWriterAndPlatform(
         break :brk chunk;
     } else null;
 
+    var buffer = printer.writer.ctx.getMutableBuffer();
```
taylordotfish: Did you mean for this to be `takeBuffer()`?

taylordotfish: If so, `buffer.slice()` below is incorrect because it could have a different length than the actual block of memory, so when we free it the length won't be correct.

src/sourcemap/sourcemap.zig
```diff
@@ -1522,15 +1522,9 @@ pub fn appendSourceMapChunk(
...
-            MutableString.initEmpty(allocator),
-            j.lastByte(),
-            prev_end_state,
-            start_state,
-        ).list.items,
-        allocator,
-    );
+    var str = MutableString.initEmpty(allocator);
+    appendMappingToBuffer(&str, j.lastByte(), prev_end_state, start_state);
+    j.push(str.slice(), allocator);
```
taylordotfish, line 1527: `str.slice()` might have a different length than the underlying memory, so when we free it the length won't be correct.

src/ast/Expr.zig
```diff
@@ -96,8 +96,8 @@ pub fn fromBlob(
 
     if (mime_type.category.isTextLike()) {
         var output = MutableString.initEmpty(allocator);
-        output = try JSPrinter.quoteForJSON(bytes, output, true);
-        var list = output.toOwnedSlice();
+        try JSPrinter.quoteForJSON(bytes, &output, true);
+        var list = output.slice();
```
taylordotfish: Same thing here, although it's not clear to me whether this actually ever gets freed

src/bake/DevServer/PackedMap.zig
```diff
@@ -31,24 +32,24 @@ end_state: struct {
...
-pub fn newNonEmpty(source_map: SourceMap.Chunk, quoted_contents: []u8) bun.ptr.RefPtr(PackedMap) {
-    assert(source_map.buffer.list.items.len > 0);
+pub fn newNonEmpty(chunk: SourceMap.Chunk, quoted_contents: []u8) bun.ptr.RefPtr(PackedMap) {
+    assert(chunk.buffer.list.items.len > 0);
     return .new(.{
         .ref_count = .init(),
-        .vlq_ptr = source_map.buffer.list.items.ptr,
-        .vlq_len = @intCast(source_map.buffer.list.items.len),
+        .vlq_ptr = chunk.buffer.list.items.ptr,
+        .vlq_len = @intCast(chunk.buffer.list.items.len),
```
taylordotfish: Same thing here since `vlq_len` is not necessarily the length of the underlying memory

src/js_printer.zig
```diff
@@ -489,28 +483,14 @@ pub const RequireOrImportMeta = struct {
...
-                },
-            },
-            .err => PrintResult{
-                .err = this.err,
-            },
-        };
-    }
+    pub const Success = struct {
+        code: []u8,
+        code_allocator: ?std.mem.Allocator,
```
taylordotfish: Since this is set to null only when `code` is empty, I think we can change this to a non-optional `std.mem.Allocator`, and just use `bun.default_allocator` instead of null. This will save space since Zig won't optimize the optional

PR #17118, opened 2025-02-06
fix(sql) types

cirospaciari:
### What does this PR do?
Fix: https://github.com/oven-sh/bun/issues/16724

- [x] Documentation or TypeScript types (it's okay to leave the rest blank in this case)
- [ ] Code changes

PR #17380, opened 2025-02-16
fix: avoid warning when only a few fields of SQLOptions are provided

nns52k:
VSCode (or VSCodium) will give us this warning when we doesn't provide all fields of `SQLOptions` to the ctor of `SQL`.

### What does this PR do?

Avoid the warning message in VSCodium when we don't provide all fields of `SQLOptions` to create a `SQL`.

I modified the Bun code in my node_modules/ folder, and the red wavy underline is now gone.

### DIFF
--- a/packages/bun-types/bun.d.ts
+++ b/packages/bun-types/bun.d.ts
@@ -2250,12 +2250,12 @@ declare module "bun" {
      * @example
      * const sql = new SQL("postgres://localhost:5432/mydb", { idleTimeout: 1000 });
      */
-    new (connectionString: string | URL, options: SQLOptions): SQL;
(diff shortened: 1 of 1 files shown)


RiskyMH:
Nice, but isn't it already partial?
https://github.com/oven-sh/bun/blob/05530627766d925bd7b6d58e4c685ff3f399edf6/packages/bun-types/bun.d.ts#L2156-L2205

RiskyMH:
Ah it was fixed in #17118, but there hasn't been a stable release yet.

For now you can install `bun-types@canary` (may need to use `"overrides"` with it if using `@types/bun`).

Thanks for the PR though!

Issue #11710, opened 2024-06-08
Crash during install


sockmaster27:
### How can we reproduce the crash?

Only happened one time, but what I did was
- Cloned [knip](https://github.com/webpro-nl/knip).
- Bumped the patch version of a dependency in `packages/knip/package.json`.
- Installed Bun for Windows via the recommended `powershell -c "irm bun.sh/install.ps1 | iex"` in the VS Code terminal.
- Restarted VS Code.
- Ran `bun install`.

Rerunning `bun install` worked though.

### Relevant log output

```shell
bun install v1.1.12 (43f0913c)
  📦 Installing [903/1174] ============================================================
Bun v1.1.12 (43f0913c) Windows x64
Args: "C:\Users\Holger\.bun\bin\bun.exe" "install"
Features: extracted_packages(923) 
Elapsed: 30776ms | User: 4328ms | Sys: 32265ms
RSS: 0.65GB | Peak: 0.65GB | Commit: 0.98GB | Faults: 279045
```

### Stack Trace (bun.report)

Bun v1.1.12 ([`43f0913`](<https://github.com/oven-sh/bun/tree/43f0913c3875cbf260c60868268adc5953b09a80>)) on windows x86_64 [InstallCommand]

**panic**: index out of bounds

- [`install.zig:5156`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/install/install.zig#L5156>): `processDependencyListItem`
- [`logger.zig:850`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/logger.zig#L850>): `allocPrint`
- [`install.zig:5937`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/install/install.zig#L5937>): `runTasks`
- [`install.zig:10031`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/install/install.zig#L10031>): `installPackageWithNameAndResolution`
- [`event_loop.zig:1875`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/bun.js/event_loop.zig#L1875>): `tick`
- [`filter_run.zig:431`](<https://github.com/oven-sh/bun/blob/43f0913c3875cbf260c60868268adc5953b09a80/src/cli/filter_run.zig#L431>): `install`
[...]

PR #17397, opened 2025-02-17
fix(ffi): make using `tinycc` bindings safer

DonIsaac:
### What does this PR do?
Adds a wrapper to all tcc bindings to convert error codes to error unions, forcing better error handling. FFI errors should be much better handled now, though there is still room for improvement.

This PR also fixes a bus error caused by running `library.close()` on any library created by `cc`. Turns out we cannot free relocated memory, even if JIT write protections are disabled.

I've added tests. These changes should still be carefully reviewed.


robobun:
<div><sup>Updated 9:19 PM PT - Feb 16th, 2025</sup></div>

@DonIsaac, your commit 2e087fa2279c1ee78f95770a5ee26ff1068020d8 is building: [`#11710`](https://buildkite.com/bun/bun/builds/11710)

test/js/bun/ffi/cc.test.ts
```diff
@@ -14,3 +16,106 @@ it.todoIf(isWindows)("can run a .c file", () => {
...
+  it("when compiled with a symbol that doesn't exist, throws an error", () => {
+    expect(() => {
+      cc({
+        source: path.join(dir, "add.c"),
+        symbols: { subtract: { args: ["int", "int"], returns: "int" } },
+      });
+    }).toThrow(/"subtract" is missing/);
+  });
+
+  it("when compiled with a `void` return type instead of `int`, returns undefined", () => {
```
Jarred-Sumner: this is undefined behavior

Jarred-Sumner: instead another function should be used, like `addVoid`

but tinycc will probably compile it into nothing

src/bun.js/api/ffi.zig
```diff
@@ -9,72 +8,37 @@ const strings = bun.strings;
...
 const VM = bun.JSC.VM;
-const JSFunction = bun.JSC.JSFunction;
-const Config = @import("../config.zig");
-const URL = @import("../../url.zig").URL;
 const VirtualMachine = JSC.VirtualMachine;
-const IOTask = JSC.IOTask;
 
-const TCC = @import("../../tcc.zig");
+const TCC = @import("../../deps/tcc.zig");
 extern fn pthread_jit_write_protect_np(enable: bool) callconv(.C) void;
```
190n: this function's argument is `int` not `bool`, and `callconv(.C)` is redundant on `extern` declarations

src/deps/tcc.zig
```diff
@@ -0,0 +1,315 @@
...
+            @branchHint(.unlikely);
+            return error.InvalidSymbol;
+        }
+    }
+
+    pub fn addSymbolsComptime(s: *State, symbols: anytype) Error!void {
+        const info = @typeInfo(@TypeOf(symbols));
+        inline for (info.@"struct".fields) |field| {
+            const value = @field(symbols, field.name);
+            std.mem.doNotOptimizeAway(value); // just in case
```
190n: Why is this needed?

test/js/bun/ffi/cc.test.ts
```diff
@@ -14,3 +16,181 @@ it.todoIf(isWindows)("can run a .c file", () => {
...
+  it("when compiled with a symbol that doesn't exist, throws an error", () => {
+    expect(() => {
+      cc({
+        source: path.join(dir, "add.c"),
+        symbols: { subtract: { args: ["int", "int"], returns: "int" } },
+      });
+    }).toThrow(/"subtract" is missing/);
+  });
+
+  it("compiling with an incorrect signature is U.B. but doesn't crash", () => {
```
190n: Is there a reason to test this? It's U.B., like you said, and plenty of incorrect signatures would cause a crash.

190n (changes requested):

src/deps/tcc.zig
DonIsaac: It's cleaner than calling `addSymbol` over and over again. I haven't refactored all the code to use it yet, just some of it.

test/js/bun/ffi/cc.test.ts
DonIsaac: Its more user-friendly to not crash. Ideally we'd throw an error.

src/deps/tcc.zig
190n: I mean why is `doNotOptimizeAway` needed

test/js/bun/ffi/cc.test.ts
190n: I'm just wary because we can never make a guarantee of not crashing when the signature is wrong.

src/bun.js/api/ffi.zig
DonIsaac: fixed

src/deps/tcc.zig
DonIsaac: I've refactored remaining code to use it.

190n: Does that code fail if `doNotOptimizeAway` is removed?

We're passing `value` to an `extern` function (`s.addSymbol` -> `tcc_add_symbol`) so I don't see how it could get optimized away in any case.

DonIsaac: I'm not certain. I added it because of the "workaround" at the bottom, where several added symbols are explicitly marked with `doNotOptimizeAway`. Something happened there a while ago and I don't know what...

test/js/bun/ffi/cc.test.ts
DonIsaac: I've removed it

src/deps/tcc.zig
190n: Let's see if it works without `doNotOptimizeAway`. This seems like the linker issue that sometimes deletes symbols we need that aren't referenced, but it seems like these out to already be referenced by the Zig code so it ought not to be necessary to reference them more explicitly.

DonIsaac: removed

190n (approved):

paperclover (approved):

Issue #11716, opened 2024-06-09
Implement `aborted` in `node:util`


amagee:
### What version of Bun is running?

Bun v1.1.12 (Linux x64)

### What platform is your computer?

Linux 6.5.0-41-generic x86_64 x86_64

### What steps can reproduce the bug?

1. `bun init`
2. `bun install execa` (or `bun execa@9.2.0` for current version as of today)
3. Create a file `abort.js` with a single line - `import { execa } from "execa";`
4. `bun abort.js`

### What is the expected behavior?

Program should execute and do nothing but no errors

### What do you see instead?

```
1 | (function (entry, fetcher)
    ^
SyntaxError: Export named 'aborted' not found in module 'util'.

Bun v1.1.12 (Linux x64)
```

yoursunny:
I have a complicated script that imports execa in one of its dependencies.
Without reading this thread, I would not have guessed the offending import is from execa.

Bun only gives the following error message:

```
$ bun run ./producer.ts 
[bun] Warning: async_hooks.createHook is not implemented in Bun. Hooks can still be created but will never be called.
- | (function (entry, fetcher)
SyntaxError: Export named 'aborted' not found in module 'util'.

Bun v1.1.17 (Linux x64)
```

**Is there any command line flag or environment variable setting that can make Bun print where the offending import is?**
I already tried `--trace-warnings` and `--inspect-wait`, but never could reveal this information.

OskarLebuda:
Same issue here :(

hiddentao:
Seeing this too in v9+. Switched to v8 of `execa` and it works.

Jarred-Sumner:
@hiddentao what version of Bun are you on? we implemented support for this months ago

hiddentao:
> [@hiddentao](https://github.com/hiddentao) what version of Bun are you on? we implemented support for this months ago

My bad, I'm on 1.1.34. Will upgrade!

PR #17399, opened 2025-02-17
fix: memory leak in code coverage

DonIsaac:
### What does this PR do?
`generateReportFromBlocks` was not `deref`-ing its source map when it was done with it, causing a memory leak.

### DIFF
--- a/src/bun.js/javascript.zig
+++ b/src/bun.js/javascript.zig
@@ -293,6 +293,7 @@ pub const SavedSourceMap = struct {
         entry.value_ptr.* = value.ptr();
     }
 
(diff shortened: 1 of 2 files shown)


robobun:
<div><sup>Updated 9:55 PM PT - Feb 16th, 2025</sup></div>

@DonIsaac, your commit d85921f0aaa8a0eed0fdca025ac693ded4cea9c1 is building: [`#11716`](https://buildkite.com/bun/bun/builds/11716)

Jarred-Sumner:
Nice catch. Note that sourcemaps are generally only freed when hot reloading is enabled, so this would impact `bun test --hot` mostly. But it's still good hygiene to deref it.

Issue #11752, opened 2024-06-10
Support ${configDir} in typescript paths


lurebat:
### What is the problem this feature would solve?

The upcoming typescript 5.5 will have a template variable in tsconfig.json called ${configDir}, which lets you specify relative path to the tsconfig.json file.

Currently bun does not support this feature when parsing typescript's paths.

### What is the feature you are proposing to solve the problem?

Add support for this feature.
Currently the version is still in RC, but I didn't see any issue about this yet.

alextbok:
Looks like this still isn't supported in bun v1.1.16 even though typescript v5.5.2 is now released.

ivodolenc:
Hi, I can also confirm that this is not working for me either.

`${configDir}` is super useful and a long awaited feature from `TS`, especially for shared config from npm.

Bun version: `1.1.27`

---

### Current behavior:

**shared npm pkg `tsconfig.json`**

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "baseUrl": "${configDir}",
    "paths": {
      "@": ["${configDir}/src"],
      "@/*": ["${configDir}/src/*"],
      "~": ["${configDir}/src"],
      "~/*": ["${configDir}/src/*"]
    }
  },
  "include": ["${configDir}/src/**/*"],
  "exclude": ["${configDir}/node_modules"]
}
```

**local project `tsconfig.json`**

```json
{
  "extends": "shared-npm-pkg/tsconfig"
}
```

Everything works great, alias and all local dev configs such as eslint etc, but as soon as I run CLI command `bun run ./src/test.ts` it throws error:

```txt
error: Cannot find module "@/utils" from "..."

Bun v1.1.27 (macOS x64)
 ELIFECYCLE  Command failed with exit code 1.
```

### Current Workaround:

The current solution is to manually specify the `paths`, `baseUrl`, `include` and `exclude` options in `tsconfig.json`.

This example below works fine with `bun`:

```json
{
  "extends": "shared-npm-pkg/tsconfig",
  "compilerOptions": {
    "baseUrl": "./",
    "paths": {
      "@": ["./src"],
      "@/*": ["./src/*"]
    }
  },
  "include": ["src/**/*", "*.js", "*.ts"],
  "exclude": ["node_modules"]
}
```

### Expected behavior:

I would expect `bun` to work without additional manual config changes within the `tsconfig.json` file.

ivodolenc:
[...]

PR #17421, opened 2025-02-18
Enable tree shaking for the dev server

### LINKED ISSUE #17409: Implement optimizePackageImports in Bun.build() & HMR
Importing `lucide-react` imports every icon, which makes the default shadcn template have a noticable ~200ms delay when starting up. What should be < 1 MB or so costs 3.8 MB.

Image

We should solve this the same way Next.js does: rewrite imports from a bunch of packages (with user-overridable defaults) to their module-specific equivalents. So `import { World } from 'lucide-react'` becomes something like `import World from 'lucide-react/dist/esm/World.js'`.

While it would be easiest to do this in the transpiler, one nuance here is we can't entirely rely on a specific string aliasing pattern to work. Instead, we should follow the exports for the package. However, following the exports for the package is expensive - because that would ordinarily parse every file. We _probably_ don't need to parse every file. Instead, in onParseTask, we can check if the file came from a barrel package, follow the direct export for that package (one level deep), and then make it so all references point to the new file instead. This will need to handle multiple named imports.

See [this code in Next.js](https://github.com/vercel/next.js/blob/d69f796522cb843b959e6d30d6964873cfd14d23/packages/next/src/server/config.ts#L937-L1067)

```js
const userProvidedModularizeImports = result.modularizeImports
  // Unfortunately these packages end up re-exporting 10600 modules, for example: https://unpkg.com/browse/@mui/icons-material@5.11.16/esm/index.js.
  // Leveraging modularizeImports tremendously reduces compile times for these.
  result.modularizeImports = {
    ...(userProvidedModularizeImports || {}),
    // This is intentionally added after the user-provided modularizeImports config.
    '@mui/icons-material': {
      transform: '@mui/icons-material/{{member}}',
    },
    lodash: {
      transform: 'lodash/{{member}}',
    },
  }

  const userProvidedOptimizePackageImports =
    result.experimental?.optimizePackageImports || []
  if (!result.experimental) {
    result.experimental = {}
  }
  result.experimental.optimizePackageImports = [
    ...new Set([
      ...userProvidedOptimizePackageImports,
      'lucide-react',
      'date-fns',
      'lodash-es',
      'ramda',
      'antd',
      'react-bootstrap',
      'ahooks',
      '@ant-design/icons',
      '@headlessui/react',
      '@headlessui-float/react',
      '@heroicons/react/20/solid',
      '@heroicons/react/24/solid',
      '@heroicons/react/24/outline',
      '@visx/visx',
      '@tremor/react',
      'rxjs',
      '@mui/material',
      '@mui/icons-material',
      'recharts',
      'react-use',
      'effect',
      '@effect/schema',
      '@effect/platform',
      '@effect/platform-node',
      '@effect/platform-browser',
      '@effect/platform-bun',
      '@effect/sql',
      '@effect/sql-mssql',
      '@effect/sql-mysql2',
      '@effect/sql-pg',
      '@effect/sql-squlite-node',
      '@effect/sql-squlite-bun',
      '@effect/sql-squlite-wasm',
      '@effect/sql-squlite-react-native',
      '@effect/sql-squlite-wasm',
      '@effect/rpc',
      '@effect/rpc-http',
      '@effect/typeclass',
      '@effect/experimental',
      '@effect/opentelemetry',
      '@material-ui/core',
      '@material-ui/icons',
      '@tabler/icons-react',
      'mui-core',
      // We don't support wildcard imports for these configs, e.g. `react-icons/*`
      // so we need to add them manually.
      // In the future, we should consider automatically detecting packages that
      // need to be optimized.
      'react-icons/ai',
      'react-icons/bi',
      'react-icons/bs',
      'react-icons/cg',
      'react-icons/ci',
      'react-icons/di',
      'react-icons/fa',
      'react-icons/fa6',
      'react-icons/fc',
      'react-icons/fi',
      'react-icons/gi',
      'react-icons/go',
      'react-icons/gr',
      'react-icons/hi',
      'react-icons/hi2',
      'react-icons/im',
      'react-icons/io',
      'react-icons/io5',
      'react-icons/lia',
      'react-icons/lib',
      'react-icons/lu',
      'react-icons/md',
      'react-icons/pi',
      'react-icons/ri',
      'react-icons/rx',
      'react-icons/si',
      'react-icons/sl',
      'react-icons/tb',
      'react-icons/tfi',
      'react-icons/ti',
      'react-icons/vsc',
      'react-icons/wi',
    ]),
```

Also relevant:
- https://vercel.com/blog/how-we-optimized-package-imports-in-next-js

zackradisic:
Enables tree shaking for the dev server

Fixes #17409

### DIFF
--- a/src/bun.js/api/JSTranspiler.zig
+++ b/src/bun.js/api/JSTranspiler.zig
@@ -64,7 +64,7 @@ const TranspilerOptions = struct {
     tsconfig_buf: []const u8 = "",
     macros_buf: []const u8 = "",
     log: logger.Log,
-    runtime: Runtime.Features = Runtime.Features{ .top_level_await = true },
+    runtime: Runtime.Features,
     tree_shaking: bool = false,
     trim_unused_imports: ?bool = null,
     inlining: bool = false,
@@ -310,8 +310,11 @@ fn exportReplacementValue(value: JSValue, globalThis: *JSGlobalObject) ?JSAst.Ex
 
 fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std.mem.Allocator, args: *JSC.Node.ArgumentsSlice) (bun.JSError || bun.OOM)!TranspilerOptions {
     const globalThis = globalObject;
-    const object = args.next() orelse return TranspilerOptions{ .log = logger.Log.init(temp_allocator) };
-    if (object.isUndefinedOrNull()) return TranspilerOptions{ .log = logger.Log.init(temp_allocator) };
+    const object = args.next() orelse return TranspilerOptions{
+        .log = logger.Log.init(temp_allocator),
+        .runtime = Runtime.Features.default(),
+    };
+    if (object.isUndefinedOrNull()) return TranspilerOptions{ .log = logger.Log.init(temp_allocator), .runtime = Runtime.Features.default() };
 
     args.eat();
     var allocator = args.arena.allocator();
@@ -320,6 +323,7 @@ fn transformOptionsFromJSC(globalObject: JSC.C.JSContextRef, temp_allocator: std
         .default_loader = .jsx,
         .transform = default_transform_options,
         .log = logger.Log.init(allocator),
+        .runtime = Runtime.Features.default(),
     };
 
     if (!object.isObject()) {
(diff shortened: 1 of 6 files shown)


robobun:
<div><sup>Updated 4:53 PM PT - Feb 17th, 2025</sup></div>

@zackradisic, your commit 1ecee6d11dd639a7c53acc5bca833b2ea1ca90c6 is building: [`#11752`](https://buildkite.com/bun/bun/builds/11752)

paperclover:
please test code before asking for a review

paperclover:
Tree shaking is disabled for a reason. See failing tests.

Issue #15481, opened 2024-11-28
@types/bun conflicts with @types/node breaking tsc


CodingMarkus:
### What version of Bun is running?

1.1.37

### What platform is your computer?

Darwin 23.6.0 arm64 arm

### What steps can reproduce the bug?

```js
bun install -d @types/bun
```

is all that is required to make the project incompatible with `tsc`.

### What is the expected behavior?

tsc should still work, after all I just installed the bun types to the project.

### What do you see instead?

`tsc --emitDeclarationOnly` results in:

```js
node_modules/@types/node/globals.d.ts:370:14 - error TS2300: Duplicate identifier 'fetch'.

370     function fetch(
                 ~~~~~

  node_modules/bun-types/globals.d.ts:1037:6
    1037  var fetch: Fetch;
              ~~~~~
    'fetch' was also declared here.

node_modules/@types/node/module.d.ts:283:13 - error TS2687: All declarations of 'dirname' must have identical modifiers.

283             dirname: string;
                ~~~~~~~

node_modules/@types/node/module.d.ts:289:13 - error TS2687: All declarations of 'filename' must have identical modifiers.

289             filename: string;
                ~~~~~~~~

node_modules/bun-types/bun.d.ts:117:8 - error TS2420: Class 'ShellError' incorrectly implements interface 'ShellOutput'.
  Property 'bytes' is missing in type 'ShellError' but required in type 'ShellOutput'.

117  class ShellError extends Error implements ShellOutput {
           ~~~~~~~~~~

  node_modules/bun-types/bun.d.ts:434:3
    434   bytes(): Uint8Array;
          ~~~~~~~~~~~~~~~~~~~~
    'bytes' is declared here.

node_modules/bun-types/globals.d.ts:1037:6 - error TS2300: Duplicate identifier 'fetch'.

1037  var fetch: Fetch;
          ~~~~~

  ../../../../../../../opt/local/lib/node_modules/typescript/lib/lib.dom.d.ts:28148:18
    28148 declare function fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response>;
                           ~~~~~
    'fetch' was also declared here.
  node_modules/@types/node/globals.d.ts:370:14
    370     function fetch(
                     ~~~~~
    and here.

node_modules/bun-types/globals.d.ts:1947:12 - error TS2687: All declarations of 'dirname' must have identical modifiers.

1947   readonly dirname: string;
                ~~~~~~~

[...]

Issue #13864, opened 2024-09-10
Expose `ShellError`


ajitid:
### What is the problem this feature would solve?

Right now when I install `@types/bun` and write:

```ts
import { $, ShellError } from 'bun'

try {
  await $`a-command-that-will-surely-fail`
} catch (err) {
  if (err instanceof ShellError) {
    // handle error
  }
}
```

The code doesn't give TypeScript error but fails at runtime because ShellError is undefined.

### What is the feature you are proposing to solve the problem?

Expose ShellError class, just like `$` is available. This way I would be able to use instanceof on it.

### What alternatives have you considered?

I tried two workarounds neither which I liked:

1. Intentionally make Bun shell throw error once so I could grab an actual instance

```ts
import { $, type ShellError } from 'bun'

let classCtor: Function
try {
  await $`a-command-that-will-surely-fail^#^#`
} catch (err: any) {
  classCtor = err.constructor
}

export const isShellError = (err: any): err is ShellError => {
  return err instanceof classCtor
}
```  

2. Check if properties match, and declare the error as ShellError

```ts
import { type ShellError } from 'bun'

export const isShellError = (err: any): err is ShellError => {
  if (err === null) return false
  if (typeof err !== 'object') return false
  return (
    err.constructor?.name === 'ShellError' &&
    typeof err.json === 'function' &&
    typeof err.exitCode === 'number'
  )
}
```

alii:
This is fixed in #17424. Working on fixing other type declaration issues before that merges

alii:
The PR above just got merged, so will be included in the next release.

It's no longer exported from the root, but you can access the instance on the Bun.$ object

```ts
import {$} from 'bun';

if (e instanceof $.ShellError) {
  e // much wow
}
```

Issue #8761, opened 2024-02-07
`@types/bun` still conflicts with `@types/node`


lgarron:
### What version of Bun is running?

1.0.25+a8ff7be64

### What platform is your computer?

Darwin 23.2.0 arm64 arm

### What steps can reproduce the bug?

```
mkdir /tmp/bun-node-types && cd /tmp/bun-node-types
bun init --yes
bun add @types/node @types/bun typescript
echo "{}" > tsconfig.json # Remove `skipLibCheck`
bun x tsc
```

### What is the expected behavior?

TypeScript runs successfully.

### What do you see instead?

```
node_modules/typescript/lib/lib.dom.d.ts:8138:11 - error TS2430: Interface 'EventSource' incorrectly extends interface 'Bun.EventSource'.
  Types of property 'onerror' are incompatible.
    Type '(this: EventSource, ev: Event) => any' is not assignable to type '(this: Bun.EventSource, ev: Event) => any'.
      The 'this' types of each signature are incompatible.
        Type 'Bun.EventSource' is not assignable to type 'EventSource'.
          Types of property 'onmessage' are incompatible.
            Type '(this: Bun.EventSource, ev: Bun.MessageEvent<any>) => any' is not assignable to type '(this: EventSource, ev: MessageEvent<any>) => any'.
              Types of parameters 'ev' and 'ev' are incompatible.
                Type 'MessageEvent<any>' is not assignable to type 'Bun.MessageEvent<any>'.
                  Types of property 'ports' are incompatible.
                    Type 'readonly MessagePort[]' is not assignable to type 'readonly import("worker_threads").MessagePort[]'.
                      Type 'MessagePort' is missing the following properties from type 'MessagePort': ref, unref, addListener, emit, and 13 more.

8138 interface EventSource extends EventTarget {
               ~~~~~~~~~~~

node_modules/typescript/lib/lib.dom.d.ts:15414:11 - error TS2430: Interface 'MessageEvent<T>' incorrectly extends interface 'Bun.MessageEvent<T>'.
  Types of property 'ports' are incompatible.
    Type 'readonly MessagePort[]' is not assignable to type 'readonly import("worker_threads").MessagePort[]'.

15414 interface MessageEvent<T = any> extends Event {
                ~~~~~~~~~~~~

[...]

PR #17424, opened 2025-02-18
Fix some issues in Bun types

alii:
### What does this PR do?

This PR improves some of the TypeScript issues in @types/bun. Originally it attempted to address conflicts with @types/node, but unfortunately we can't fully resolve that to do limitations with TypeScript. More about that in a comment below.

Addresses #15481 #8761 #13864 #17470 #17598

- [x] Code changes (only minor)

There's now an integration test for TypeScript which uses the local `bun-types` and pretends to be a real test case.


robobun:
<div><sup>Updated 7:32 PM PT - Feb 17th, 2025</sup></div>

@alii, your commit 1d6edb3daefbbfd13bae7e8a1fa78c5f47cd70eb is building: [`#11766`](https://buildkite.com/bun/bun/builds/11766)

kravetsone:
Did u format it using bun's repo formatter?

Looks like there so many style diffs

alii:
Unfortunately this problem is not fully solvable due to a limitation of TypeScript.

The TLDR is that @types/node is almost always transitively loaded in every program, and TypeScript doesn't let us fully merge all declarations without compiler errors in the `.d.ts` files. For this reason we suggest you keep `"skipLibCheck": true` in your tsconfig.

--- 

Breakdown:

- There's no way to force disable a specific `@types/[pkg]` from loading, it's either everything (breaks lib) or manually specifying each individual package (bad experience)
- @types/node looks for `onmessage` to exist on globalThis to set all Request, Response, etc to `{}` to allow for any implementation from another package...
- ...but if we overwrite/redeclare properties like Request/Response, then there's an error in @types/node because TypeScript claims that `Request` was "either direclty indirectly used" in its own declaration, so the type can't be evaluated in @types/node. This means it fails with `"skipLibCheck": false`
- And even if we don't set `onmessage: never`, the declarations still claim to be duplicate anyway

There are a few issues talking about this on the TypeScript repo https://github.com/microsoft/TypeScript/issues/37053, https://github.com/microsoft/TypeScript/issues/33111, https://github.com/microsoft/TypeScript/issues/37775

Some other runtimes that publish TypeScript types suggest developers use the `types: []` tsconfig option, but you must do it for every single `@types/[package]`, and unfortuantely there's no way to say "I don't want this library to load at all, even transitively" and even if there was, it's sort a bandaid solution for the bigger problem. The bigger problem is really like "How can we tell TypeScript that the dependencies we are using will/won't work with a specific runtime (Bun)"

We'd like to get this working in TypeScript eventually. There are certainly valid cases for wanting to typecheck your libs.

src/bun.js/bindings/BunObject.cpp
```diff
@@ -684,7 +670,8 @@ JSC_DEFINE_HOST_FUNCTION(functionFileURLToPath, (JSC::JSGlobalObject * globalObj
 
 /* Source for BunObject.lut.h
 @begin bunObjectTable
-    $                                              constructBunShell                                                   ReadOnly|DontDelete|PropertyCallback
+    $                                              BunObject_getter_wrap_BunShell                                      DontDelete|PropertyCallback
+    ShellError                                     BunObject_getter_wrap_ShellError                                    DontDelete|PropertyCallback
```
Jarred-Sumner: Can we instead do `$.ShellError`?

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3386,6 +3387,31 @@ void GlobalObject::finishCreation(VM& vm)
...
+            JSFunction* createParsedShellScript = JSFunction::create(init.vm, init.owner, 2, "createParsedShellScript"_s, BunObject_callback_createParsedShellScript, ImplementationVisibility::Private, NoIntrinsic);
+            JSFunction* createShellInterpreterFunction = JSFunction::create(init.vm, init.owner, 1, "createShellInterpreter"_s, BunObject_callback_createShellInterpreter, ImplementationVisibility::Private, NoIntrinsic);
+            JSC::JSFunction* createShellFn = JSC::JSFunction::create(init.vm, init.owner, shellCreateBunShellTemplateFunctionCodeGenerator(init.vm), init.owner);
+
+            auto scope = DECLARE_THROW_SCOPE(init.vm);
+            auto args = JSC::MarkedArgumentBuffer();
+            args.append(createShellInterpreterFunction);
+            args.append(createParsedShellScript);
+            JSC::JSValue shell = JSC::call(init.owner, createShellFn, args, "BunShell"_s);
+            RETURN_IF_EXCEPTION(scope, (void)0);
```
Jarred-Sumner:
```suggestion
            RETURN_IF_EXCEPTION(scope, {});
```

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3386,6 +3387,31 @@ void GlobalObject::finishCreation(VM& vm)
             init.setStructure(Zig::JSFFIFunction::createStructure(init.vm, init.global, init.global->functionPrototype()));
         });
 
+    m_BunShell.initLater(
+        [](const JSC::LazyProperty<JSC::JSGlobalObject, JSC::JSObject>::Initializer& init) {
```
Jarred-Sumner: Can this be moved back into BunObject? Zig::GlobalObject is already massive

src/bun.js/bindings/ZigGlobalObject.cpp
```diff
@@ -3386,6 +3387,31 @@ void GlobalObject::finishCreation(VM& vm)
...
+            RETURN_IF_EXCEPTION(scope, (void)0);
+
+            if (UNLIKELY(!shell.isObject())) {
+                throwTypeError(init.owner, scope, "Internal error: BunShell constructor did not return an object"_s);
+                return (void)0;
+            }
+            auto* bunShell = shell.getObject();
+
+            bunShell->putDirectNativeFunction(init.vm, init.owner, Identifier::fromString(init.vm, "braces"_s), 1, Generated::BunObject::jsBraces, ImplementationVisibility::Public, NoIntrinsic, JSC::PropertyAttribute::DontDelete | JSC::PropertyAttribute::ReadOnly | 0);
+            bunShell->putDirectNativeFunction(init.vm, init.owner, Identifier::fromString(init.vm, "escape"_s), 1, BunObject_callback_shellEscape, ImplementationVisibility::Public, NoIntrinsic, JSC::PropertyAttribute::DontDelete | JSC::PropertyAttribute::ReadOnly | 0);
```
Jarred-Sumner: I suggest putting ShellError here

packages/bun-types/ambient.d.ts
```diff
@@ -23,3 +23,9 @@ declare module "*.html" {
   var contents: any;
   export = contents;
 }
+
+declare module "*.svg" {
+  // Bun 1.2.3 added support for frontend dev server
+  var contents: `${string}.svg`;
```
Jarred-Sumner, line 29: nice way to say "this is a file path"

is there a way we could let users override it though? like if they use the text loader to import SVGs or something

packages/bun-types/ambient.d.ts
```diff
@@ -23,3 +23,9 @@ declare module "*.html" {
...
+  export = contents;
+}
```
RiskyMH, line 31: I think for this one it should create a .d.ts file with it for bun init, as otherwise you will have to add every common extension

packages/bun-types/devserver.d.ts
```diff
@@ -0,0 +1,24 @@
...
+       * import.meta.hot.data maintains state between module instances during hot replacement, enabling data transfer from previous to new versions.
+       *
+       * @example
+       * ```ts
+       * import.meta.hot.data = {
+       *   bun: 'is cool',
+       * };
+       * ```
+       */
+      data: any;
```
RiskyMH, line 21: @paperclover said it was `Record<PropertyKey, any>`, but doesn't really matter that much as you can set it to whatever (its just default an object)

packages/bun-types/package.json
```diff
@@ -16,7 +15,6 @@
   ],
   "homepage": "https://bun.sh",
   "dependencies": {
-    "@types/node": "*",
```
RiskyMH: poor node

packages/bun-types/ambient.d.ts
RiskyMH: I think for this one it should create a `.d.ts` file with it for `bun init`, as otherwise you will have to add every common extension

alii: @RiskyMH are you saying similar to like `next-env.d.ts`?

packages/bun-types/package.json
alii: woops, will add back

packages/bun-types/ambient.d.ts
RiskyMH: Yeah it should show people how to add other extensions

but tbh still could be fine to add a bigger default list of extensions (I don't think ts validates path so could give too much false hope though)

alii: Yeah it doesn't validate it exists unfortunately. Our declaration is the contract that it exists.

I think it's ok in this case, and maybe we can write some doc about "If you want to support .myfile you can define bun-env.d.ts or similar"

packages/bun-types/package.json
```diff
@@ -16,7 +15,7 @@
...
+    "@types/node": "^22.13.5",
```
Jarred-Sumner: Can we keep this to a `*`? What if they use 23?

alii: Mistake, updated

packages/bun-types/ambient.d.ts
michaelhays: A doc for this would be helpful -- with `bun-types@1.2.5`, I'm running into the TypeScript issue:

```
Type '`${string}.svg`' is not assignable to type 'ComponentType<PropsWithChildren<SVGProps<SVGElement>>>'.
```

and I'm still not sure how to fix it.

chimericdream: > A doc for this would be helpful -- with `bun-types@1.2.5`, I'm running into the TypeScript issue:
> 
> ```
> Type '`${string}.svg`' is not assignable to type 'ComponentType<PropsWithChildren<SVGProps<SVGElement>>>'.
> ```
> 
> and I'm still not sure how to fix it.

Came here to say exactly this. For now, we haven't upgraded to 1.2.x of Bun, but the types accidentally got updated to the latest, and I wound up chasing this error for a few hours today.

alii: @michaelhays @chimericdream I've opened #18527 to address this

Issue #11821, opened 2024-06-12
[BUG] Bun doesnt Ignore SSL issues with rejectUnauthorized


kravetsone:
### What version of Bun is running?

1.1.12+43f0913c3

### What platform is your computer?

Microsoft Windows NT 10.0.22631.0 x64

### What steps can reproduce the bug?

When we develop in Dev mode, we are needed to use a VPN

```js
import https from "node:https";

// two ways of disabling UNABLE_TO_GET_ISSUER_CERT_LOCALLY errors
process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
https.globalAgent.options.rejectUnauthorized = false;

const response = await fetch(
	process.env.SOME_LINK
);

const data = await response.text();

console.log(data.length);
```

Run with [Bun](https://bun.sh/)

```ts
PS Z:\PROJECTS\node-ts\elycart-monorepo> bun .\1.mjs
UNABLE_TO_GET_ISSUER_CERT_LOCALLY: unable to get local issuer certificate
 path: "https://130.193.66.56/RuRu.FrontEnd.ServiceProvider2/TransactionService.svc"

Bun v1.1.13 (Windows x64)
```

Run with Node.js

```ts
PS Z:\PROJECTS\node-ts\elycart-monorepo> node 1.mjs  
(node:33948) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
(Use `node --trace-warnings ...` to show where the warning was created)
35183
```

With https.Agent too:

```ts
const httpsAgent = new https.Agent({
  rejectUnauthorized: false,
});

axios.get(process.env.SOME_LINK, { httpsAgent });
```

paperdave:
i thought NODE_TLS_REJECT_UNAUTHORIZED=0 works, but it might have to be passed as an environment variable from outside, not being the mutated global. workaround should be:

```
NODE_TLS_REJECT_UNAUTHORIZED=0 bun run ./index.ts
```

since the two options shown do work in node, we should make sure this works in bun too

kravetsone:
> NODE_TLS_REJECT_UNAUTHORIZED=0

Ohh)
Thank you! it works for me

Jarred-Sumner:
This is fixed in #11833

PR #17450, opened 2025-02-19
node: test-buffer-write.js

nektro:


robobun:
<div><sup>Updated 6:31 PM PT - Feb 18th, 2025</sup></div>

@nektro, your commit a65999b3e0734b3523e05867c1f68e98dc20e44c is building: [`#11821`](https://buildkite.com/bun/bun/builds/11821)

nektro:
new commits no longer regress relative to 1.2.2

image

src/bun.js/bindings/JSBuffer.cpp
```diff
@@ -591,6 +614,9 @@ static JSC::EncodedJSValue jsBufferConstructorFunction_allocBody(JSC::JSGlobalOb
     RETURN_IF_EXCEPTION(scope, {});
     size_t length = lengthValue.toLength(lexicalGlobalObject);
```
Jarred-Sumner, line 615:
```suggestion
    size_t length = lengthValue.toLength(lexicalGlobalObject);
    RETURN_IF_EXCEPTION(scope, {});
```

nektro: not sure this'll ever be hit with the `validateNumber` call right above it but ill add just to be sure

PR #17481, opened 2025-02-20
fix: don't save lockfile if --frozen-lockfile or --no-save

### LINKED ISSUE #16646: `bun install --frozen-lockfile` creates bun.lock, even if `package-lock.json` exists

1.2.0+b0c5a7655

Darwin 24.2.0 arm64 arm

in bun 1.1.x running `bun install --frozen-lockfile` did not generate a bun lockfile.
since bun 1.2.x it seems a lockfile is always generated. 

I wonder whether its expected or whether `bun install --frozen-lockfile` should not generate a new lock file when `package-lock.json` exists

### What do you see instead?

_No response_

### Additional information

_No response_

### LINKED ISSUE #16965: `bun install --no-save` saves a lockfile when it shouldn't

1.2.2+c1708ea6a

Darwin 24.2.0 arm64 arm

```shell
git clone https://github.com/cubing/create-cubing-app && cd create-cubing-app
git switch --detach bb841abaab25aa0ee097ce47354d1014ee61811d
cd app-template/
bun install --no-save
git status
```

`git status` shows no changes, since `bun install --no-save` has this to say about the flag:

```
      --no-save                Don't update package.json or save a lockfile
```

### What do you see instead?

`git status` shows that `bun.lock` was added.

### Additional information

Is it possible this is due to a logic issue introduced by the new lockfile format?

pikdum:
### What does this PR do?

When running `bun install --frozen-lockfile` against a project with a `package.json` and `package-lock.json`, this brings back behavior of not creating a `bun.lock` file. (Regression?)

This also gives `bun install --no-save` the same behavior. (Not sure if it ever worked when migrating.)

Closes #16646
Closes #16965

- [x] Code changes

```console
$ bun install --frozen-lockfile &> /dev/null && ls bun.lock && rm -f bun.lock
bun.lock
$ bun-debug install --frozen-lockfile &> /dev/null && ls bun.lock
ls: cannot access 'bun.lock': No such file or directory
```

Added some automated tests, too:

```console
$ bun run test 16646
```

And verified that the tests fail against current `main`.


src/install/install.zig
```diff
@@ -15439,6 +15439,9 @@ pub const PackageManager = struct {
         packages_len_before_install: usize,
         log_level: Options.LogLevel,
     ) OOM!void {
+        if (!this.options.do.save_lockfile) {
```
dylan-conway: We should check for this flag before calling this function. The bug exists in the long chain of or's and and's used to create `should_save_lockfile`. `this.options.do.save_lockfile` is already checked, so likely it's not checked early enough.

dylan-conway (changes requested):
See above

src/install/install.zig
pikdum: Moved there: https://github.com/oven-sh/bun/pull/17481/commits/eb649a4025124418171411f83565c2627e14a51e

pikdum:
@dylan-conway Can you give this another look?

pikdum:
:(

pikdum:
:(

